PR #4708, opened 2025-08-27
D2M Add Lowering into matmul_block LLK and Make it the Default Lowering for ttir.matmul

arminaleTT:
### Ticket
#4648 

### Problem description
The current lowering path for `ttir.matmul` targets the `matmul_tiles` LLK. This LLK performs one tile of matrix multiplication, relying entirely on D2M to manage partial sums and data preparation for the compute.

There exists a block-based LLK, `matmul_block` with built-in source A/B reuse and optimized address generation. This is the LLK used in the TTNN matmul op, and D2M should target it.

It is desirable to keep the lowering of `ttir.matmul` -> a `linalg.generic` with `ttir.tile_matmul` inside it. The affine map for this `linalg.generic` is how we generate datamovement loops for L1<->DEST transfers.

### What's changed
- Added a new TTKernel ops for `matmul_block` and its init functions
- Each call to `matmul_block` only computes one tile of output. I created an experimental LLK to wrap `matmul_block` and compute all tiles of output in one block
-  Modified `InsertDestRegisterAccessPass` to:
  - Match a linalg.generic with `ttir.tile_matmul` in its body, and
  - Replace it with a `ttir.tile_matmul_block` op and remove the compute loops after generating the dest access loops
- Modified `TTIRToTTkernel` to:
  - Lower `ttir.tile_matmul_block` to `ttkernel.experimental_matmul_block`
  - Compute the tile indices for the input blocks
    - The blocked inputs are originally `memref.subview` or `memref.cast`. They must be simplified into the tile indices based on the global/unblocked input mermef
- Added a pass option, `use-tile-matmul` to toggle between keeping old behaviour or using the new block matmul. By default the new block matmul is used
- Added IR and end to end tests with `use-tile-matmul` on and off

### Checklist


lib/Conversion/TTIRToTTKernel/TTIRToTTKernel.cpp
```diff
@@ -307,7 +351,7 @@ class TTIRFPUOpsRewriter : public OpConversionPattern<ConcreteOp> {
   LogicalResult
   matchAndRewrite(ConcreteOp op, typename ConcreteOp::Adaptor adaptor,
                   ConversionPatternRewriter &rewriter) const final {
-    assert(op->hasOneUse());
+    assert(op->hasOneUse() || op->use_empty());
```
arminaleTT, line 373: See `ttir.tile_matmul_block` definition here: https://github.com/tenstorrent/tt-mlir/blob/main/include/ttmlir/Dialect/TTIR/IR/TTIRGenericRegionOps.td#L332

I am open to changing the op definition to restore this assert

lib/Conversion/TTIRToTTKernel/TTIRToTTKernelPass.cpp
```diff
@@ -55,6 +55,8 @@ struct ConvertTTIRToTTKernel
     target.addLegalOp<ttir::StreamLayoutOp>();
     target.addLegalOp<ttir::ViewLayoutOp>();
     target.addLegalOp<ttir::GenericOp>();
+    target.addLegalOp<memref::CastOp>();
```
arminaleTT, line 62: The inputs to matmul_block are casts or subviews. When using tile_matmul, these are folded into the load/store at an earlier pass. Here we need to fold/resolve them manually

lib/Conversion/TTIRToTTKernel/TTIRToTTKernel.cpp
```diff
@@ -70,11 +71,63 @@ static std::pair<Value, Value> getMcastEndCoords(PatternRewriter &rewriter,
...
+  llvm_unreachable("Expected subview or cast op");
+}
+
 static Value getCB(ConversionPatternRewriter &rewriter, Value cb) {
-  memref::LoadOp loadOp = mlir::cast<memref::LoadOp>(cb.getDefiningOp());
-  assert(loadOp.getIndices().size() == 1 &&
-         "Expected single index in load op, failing.");
-  return rewriter.getRemappedValue(loadOp.getMemref());
+  if (mlir::isa<memref::LoadOp>(cb.getDefiningOp())) {
+    memref::LoadOp loadOp = mlir::cast<memref::LoadOp>(cb.getDefiningOp());
```
nsmithtt: These isa/cast pairs could be written in 1 step:
```
if (memref::LoadOp loadOp = mlir::dyn_cast<memref::LoadOp>(cb.getDefiningOp()); loadOp) {
}
```

lib/Conversion/TTIRToTTKernel/TTIRToTTKernel.cpp
```diff
@@ -70,11 +71,63 @@ static std::pair<Value, Value> getMcastEndCoords(PatternRewriter &rewriter,
               index(rewriter, loc, 1))};
 }
 
+static Value getTileIndexFromBlockView(RewriterBase &rewriter, Location loc,
+                                       Value inputView) {
```
nsmithtt, line 74: Is there a way we can lower the memref views via upstream passes to become absolute offsets instead of having to do this here?  Arguably, if no such pass exists, we could do this as part of the linearize memref pass.

Double check passes `--fold-memref-alias-ops`, `--memref-expand`, `--normalize-memrefs`, `--flatten-memref`.

lib/Conversion/TTIRToTTKernel/TTIRToTTKernel.cpp
nsmithtt: Makes sense! Let's leave the op definition for now.

lib/Dialect/TTIR/Transforms/InsertDstRegisterAccess.cpp
```diff
@@ -286,6 +296,59 @@ struct TTIRInsertDstRegisterAccessRewriter final
...
+      PatternRewriter &rewriter, GenericOp op, Region &region,
+      linalg::GenericOp linalgGenericOp, bool &modified) {
+    assert(linalgGenericOp.getInputs().size() == 2 &&
+           "Expected exactly 2 input for tile matmul");
+    assert(linalgGenericOp.getOutputs().size() == 1 &&
+           "Expected exactly 1 output for tile matmul");
+
+    Value inputA_memref = linalgGenericOp.getInputs()[0];
+    Value inputB_memref = linalgGenericOp.getInputs()[1];
+    Value outputC_memref = linalgGenericOp.getOutputs()[0];
```
nsmithtt: Style guide is camelCase.

nsmithtt (commented):
This is looking great! Comments inline

lib/Conversion/TTIRToTTKernel/TTIRToTTKernel.cpp
arminaleTT: Upstream we already have the `--fold-memref-alias-ops` pass. The limitation is that this pass only implements the subview -> absolute offset logic in the context of folding subview into a load or store that consumes it. This works fine for matmul_tiles, but in matmul_block there is no downstream memref.load or memref.store, and the consumers are hidden inside the matmul_block LLK

Other passes aren't applicable here:
- `--memref-expand`
  - Per docs, "legalizes memref operations to be convertible to LLVM". Not applicable here
  - Running it on IR before TTIRToTTKernel does not change the IR
- `--normalize-memrefs`
  - Not applicable here. The subview needs to be folded away, its affine map is fine
- `--flatten-memref`
  - Not applicable here. We don't need to linearize the subview

arminaleTT: I would prefer to not put it in `GenericLinearizeMemref`. That pass is singularly concerned with actually linearizing memrefs for loads and stores. What we are doing here isn't linearization, it's more akin to normalizing or folding and we are doing it because of the semantics of our `*_block` LLKs. 

I think the best path forward is figuring out how to avoid the manual manipulation by only using the `affine::resolveIndicesIntoOpWithOffsetsAndStrides` function used in `--fold-memref-alias-ops` (issue #4717). Then it can be its own pass or a more generic rewrite pattern to be applied around TTIRToTTKernel for `*_block` TTKernel ops

nsmithtt: I was trying to think how we could leverage that pass (separately run, before TTKernel lowering) so that we can get this complicated logic for free.  It seems like we could get it for free if we added support in GenericLinearizeMemref, imo it does seem like during lowering we need a linearized index `in0_tile_index`, `in1_tile_index`:
```
ALWI void matmul_block(uint32_t in0_cb_id, uint32_t in1_cb_id,
                       uint32_t in0_tile_index, uint32_t in1_tile_index,
                       uint32_t idst, const uint32_t transpose, uint32_t ct_dim,
                       uint32_t rt_dim, uint32_t kt_dim, uint32_t nt_dim)
```

Maybe we lose `ct_dim` + `rt_dim` but we could add those as op attributes.

arminaleTT: I'm unclear what transformation you are asking for in `GenericLinearizeMemref`

We currently have:
```
ttir.tile_matmul_block(%subviewA, %subviewB, %subviewOut) 
```
such that `%subviewA`, `%subviewB` and `%subviewOut` each represent one block of `A`, `B`, and `Out` tensors.

Which of these would be your desired output from `GenericLinearizeMemref`?

Option A)
```
ttir.tile_matmul_block(%collapsedSubviewA, %collapsedSubviewB, %collapsedSubviewOut, %block_m, %block_n, %block_k) 
```
where `%collapsedSubviewA`, `%collapsedSubviewB`, and `%collapsedSubviewOut` still refer to one block, just in a linearized form

Option B)
```
ttir.tile_matmul_block(%A, %B, %Out, %A_starting_tile_index, %B_starting_tile_index, %Out_starting_tile_index %block_m, %block_n, %block_k) 
```
where `%A`, `%B`, and `%Out` are the global members and we have entirely eliminated the subviews

arminaleTT: @nsmithtt let me know what you think of this implementation. I proceeded with option A per our discussions. I linearize the subviews in GenericLinearizeMemref and extract the block dims. Index resolution is still done in TTIRToTTKernel since the ttir op has no index parameters but the ttkernel op does

include/ttmlir/Dialect/TTIR/IR/TTIRGenericRegionOps.td
```diff
@@ -338,10 +338,24 @@ def TTIR_TileMatmulBlockOp : TTIR_GenericRegionComputeOp<"tile_matmul_block",
...
+        build($_builder, $_state, a, b, output, /*block_m=*/nullptr,
+              /*block_k=*/nullptr, /*block_n=*/nullptr,
+              /*b_block_stride=*/nullptr);
+      }]>
+    ];
 
     let extraClassDeclaration = [{
       MutableOperandRange getDpsInitsMutable() { return getOutputMutable(); }
+      bool hasAllBlockDims();
+      bool hasNoBlockDims();
```
nsmithtt: Probably just `hasBlockDims()` is enough and in the op verifier we need to check that they are all equally in the same state. And then the rest of the code can just check `if (hasBlockDims())` or `if (!hasBlockDims())`.

lib/Conversion/TTIRToTTKernel/TTIRToTTKernel.cpp
```diff
@@ -70,11 +70,73 @@ static std::pair<Value, Value> getMcastEndCoords(PatternRewriter &rewriter,
...
+          mlir::dyn_cast<memref::SubViewOp>(inputView.getDefiningOp())) {
+    // We have blocked this input. We need to get the indicies for the first
+    // tile in the subview.
+    SmallVector<Value> indices = {index(rewriter, loc, 0),
+                                  index(rewriter, loc, 0)};
+    SmallVector<Value> sourceIndices;
+
+    // TODO(#4717): This call alone should be enough to get the tile indices,
+    // but currently it returns block index instead. Once fixed, we can remove
+    // all the other calculations below.
```
nsmithtt, line 89: Sounds good!

lib/Dialect/TTIR/Transforms/InsertDstRegisterAccess.cpp
```diff
@@ -24,7 +25,10 @@ template <typename GenericOrFuncOp>
 struct TTIRInsertDstRegisterAccessRewriter final
     : public OpRewritePattern<GenericOrFuncOp> {
 public:
-  using OpRewritePattern<GenericOrFuncOp>::OpRewritePattern;
+  TTIRInsertDstRegisterAccessRewriter(mlir::MLIRContext *ctx,
+                                      const bool useTileMatmul)
+      : OpRewritePattern<GenericOrFuncOp>(ctx), useTileMatmul(useTileMatmul) {};
+  bool useTileMatmul;
```
nsmithtt: We've typically been putting member vars at the bottom of the class.

lib/Dialect/TTIR/IR/TTIRGenericRegionOps.cpp
```diff
@@ -61,9 +61,22 @@ ::mlir::LogicalResult mlir::tt::ttir::TileMatmulBlockOp::verify() {
                        "element type");
   }
 
+  if (!(hasAllBlockDims() || hasNoBlockDims())) {
+    return emitOpError(
+        "all or none of the block dim attributes must be present");
+  }
```
nsmithtt, line 69: We might want to assert something to the effect that they're either all set or all not set:
```
int numAttributesSet = getBlockM() + getBlockK() + getBlockN() + getBBlockStride();
if (numAttributesSet != 0 && numAttributesSet != 4)
```

azecevicTT (approved):
TTIR changes LGTM.

PR #4131, opened 2025-07-18
D2M support grid padding

vwellsTT:
### Ticket
https://github.com/tenstorrent/tt-mlir/issues/3810

### Problem description
We want grid shapes to pad s.t. we align up to multiples of 8 iff the grid dim is > 8.  E.g. a 12x12 -> 16x16, but a 2x3 stays 2x3.

### What's changed
This PR implements grid blocking s.t. if a grid dim is > 8, the dimension is aligned up to a multiple of 8 via padding.  It also moves the grid selection logic to construction, instead of passing a 1x1 grid around until we hit OptimizeTensorLayouts.  This allows us to eliminate this pass completely--grid override logic is moved to TTIRToTTIRGeneric + shard construction.

### Checklist
- [X] New/Existing tests provide coverage for changes

Issue #3023, opened 2025-04-21
d2m ttir.broadcast


nsmithtt:

nsmithtt:
I think we can safely close this issue, we can open bcast improvements separately.

PR #4278, opened 2025-07-30
D2M: make dim collapse toggle-able + add uncollapsed 3D tests

vwellsTT:
### Ticket
Link to Github Issue

### Problem description
Provide context for the problem.

### What's changed
Describe the approach used to solve the problem.
Summarize the changes made and its impact.

### Checklist
- [ ] New/Existing tests provide coverage for changes

PR #4709, opened 2025-08-27
D2M: support non-tile-aligned tensor shapes

wenbinlyuTT:
### Ticket
Related-To: #3013 #3014 #3015 #3023

### Problem description
Currently when we pass a tensor from the runtime, its elements are assumed to be densely packed into a flat row-major storage.

This means that elements of a tensor with a non-tile-aligned shape end up in completely wrong locations in the DRAM and thus in the CB.

Additionally, due to the grid & tile alignment logic (ref. #4131), non-tile-aligned tensors will have strictly smaller storage than their device counterparts, overflow/segfault is thus ensured during `Enqueue{Read,Write}Buffer`.

Functionally this is the foundation for tile-level broadcast and some other ops, as zero-padding is required in the last two dimensions if they are not already tile-aligned.

### What's changed

- Create a new `ttmetal-generate-dim-aligned-host-io` pass that creates a bounce buffer with fully-aligned size to do I/O with device buffer, issue a `memref.copy` between the bounce buffer and the tightly-packed host buffer, and relay per-dim alignment info to the runtime to do the correct strided copy.
- Create a new `TTCore_HostLayoutAttr` attribute to pass tensor shape & alignment info to the host runtime, inside the `memref` associated with the bounce buffer.
- The calculation and application of dimensional alignments is revised, now the alignments are precise and we no longer push the alignment to the outermost dim of the collapsed interval.
- A new `dim_alignments` field is added to the `BufferDesc` struct to hold the dimensional alignments extracted from the `HostLayoutAttr` and pass it to the runtime. The same filed is also added to `TensorDesc`.
- `HostAllocCommand`: now allocates zero-filled tensor storage with fully-aligned shapes.
- `ttmetal::memcpy`: now does strided memcpy if any of the two operands is aligned & padded.
- Add golden tests that verify the addition of non-tile-aligned shapes from 2D to 4D.

### Known issues

- The bounce buffers respect both `8x8` grid alignment and `32x32` tile alignment, and thus could lead to extreme waste for shapes like `257x1x1x1`.

### Checklist


include/ttmlir/Dialect/TTCore/IR/TTCoreOpsTypes.td
```diff
@@ -310,7 +341,7 @@ def TTCore_MetalLayoutAttr : TTCore_Attr<"MetalLayout", "metal_layout", [TTCore_
     // The logical tensor shape before distribution.
     ArrayRefParameter<"int64_t">:$logical_shape,
 
-    // Alignment for each logical dim, default is 32x1x...x1x32.
+    // Alignment for each logical dim, default is 1x1x...x32x32.
```
vwellsTT: can you explain why this is needed? in default case where we're collapsing leading dims to get 2D, aligning the first element of collapsed group ought to guarantee we're 32-aligned while maintaining minimum size I think; we collapse from right to left, and apply the alignments as we go, so a rightmost 32 is more efficient

runtime/include/tt/runtime/types.h
```diff
@@ -165,34 +163,55 @@ struct RuntimeCheckedConstObjectImpl {
 } // namespace detail
 
 struct TensorDesc {
-  std::vector<std::uint32_t> shape;
-  std::vector<std::uint32_t> stride;
-  std::uint32_t itemsize;
-  ::tt::target::DataType dataType;
-  std::int64_t alignment = 1;
+  std::vector<uint32_t> shape = {};  // Logical
```
vwellsTT: I wonder if we should just rename these vars to `logical_shape/stride`

runtime/include/tt/runtime/utils.h
```diff
@@ -15,8 +17,25 @@
...
+template <typename T>
+T alignUp(const T val, const T alignment) {
+  if (alignment == 0 || alignment == 1) {
+    return val;
+  }
+  return ((val + alignment - 1) / alignment) * alignment;
+}
+
+inline std::shared_ptr<void> mallocShared(size_t size,
+                                          size_t alignment = sizeof(size_t)) {
```
vwellsTT: hmm, why is this the default alignment value? Seems arbitrary

test/ttmlir/Conversion/TTIRToTTIRGeneric/alignments.mlir
```diff
@@ -6,11 +6,11 @@
 
 // CHECK-DAG: #[[LAYOUT_SMALL:.*]] = #ttcore.metal_layout<logical_shape = 64x64, dim_alignments = 32x32, {{.*}}>
 // CHECK-DAG: #[[LAYOUT_MEDIUM:.*]] = #ttcore.metal_layout<logical_shape = 128x96, dim_alignments = 32x32, {{.*}}>
-// CHECK-DAG: #[[LAYOUT_LARGE:.*]] = #ttcore.metal_layout<logical_shape = 256x256, dim_alignments = 256x256, {{.*}}>
+// CHECK-DAG: #[[LAYOUT_LARGE:.*]] = #ttcore.metal_layout<logical_shape = 256x256, dim_alignments = 32x32, {{.*}}>
```
vwellsTT: I think we want to preserve the grid-padding logic s.t. this testcase doesn't change

lib/Dialect/TTCore/IR/TTCoreOpsTypes.cpp
```diff
@@ -820,44 +863,107 @@ llvm::SmallVector<int64_t> MetalLayoutAttr::getNormalizedIntervals() const {
...
+// - One for the 2nd last logical dim (intervalEnd), and it's always the height
+// of a tile.
+// - One for the outermost logical dim (intervalStart), it should be the minimum
+// value such that:
+// - After appling both alignments to their corresponding dims, the entire
+// collapsed dim is tile/grid-aligned, depending on the need.
+//
+// For example, tensor<4x43x7> + grid<8x8> + tile<32x32> gives:
+// - Tile-aligned logical shape 4x64x32.
+// - Collapsed physical 2D shape 256x32.
```
vwellsTT, line 879: but if you use the old scheme, you can apply the collapse + alignment together, and get a 192x64 tensor instead (assuming alignments 32x1x32, intervals[[0,2], [2,3]] => for first interval we take 43, align to 1, multiple by 4 => 172, align to 32 => 192).  I'm missing why this change is needed, I think old approach was better optimized

lib/Dialect/TTCore/IR/TTCoreOpsTypes.cpp
```diff
@@ -820,44 +863,107 @@ llvm::SmallVector<int64_t> MetalLayoutAttr::getNormalizedIntervals() const {
...
-    // Determine which tile dimension corresponds with this interval.
-    const int64_t tileIdx =
-        (idx == static_cast<int64_t>(deviceGridShape.size()) - 2) ? 0 : 1;
-    const int64_t tileDim = tileShape[tileIdx];
-    const int64_t gridAlignmentThreshold = deviceGridShape[idx] * tileDim;
-
     // Determine alignment based on collapsed size.
     // If size > gridAlignmentThreshold, align to grid boundary, else align to
     // tile boundary.
-    int64_t alignment = (collapsedSize >= gridAlignmentThreshold)
```
vwellsTT, line 862: I guess this is why test results below changed; I think >= is more natural to me (though functionally equivalent)

test/ttmlir/Conversion/TTIRToTTIRGeneric/alignments.mlir
vwellsTT: Ah, maybe this is simple `>` vs `>=` issue

lib/Dialect/TTMetal/Transforms/GenerateDimAlignedHostIO.cpp
```diff
@@ -0,0 +1,120 @@
...
+      TTMetalGenerateDimAlignedHostIO>::TTMetalGenerateDimAlignedHostIOBase;
+
+  void runOnOperation() final {
+    ModuleOp module = getOperation();
+    SmallVector<ttmetal::EnqueueWriteBufferOp> enqueueWrites;
+    SmallVector<ttmetal::EnqueueReadBufferOp> enqueueReads;
+
+    module->walk([&](Operation *op) {
+      if (auto w = mlir::dyn_cast<ttmetal::EnqueueWriteBufferOp>(op)) {
+        enqueueWrites.push_back(w);
```
vwellsTT: do we need to store this to be processed later? I think you could just process each op in the `walk` itself?

vwellsTT (changes requested):
Thanks for taking this on!  I think the logic mostly looks fine, but I'm not sure on changing the collapse/alignment logic, I think I'm missing something on that.
I'll do another pass of nits etc. after we sort that out too.

include/ttmlir/Dialect/TTCore/IR/TTCoreOpsTypes.td
wenbinlyuTT: In this PR I make the alignments 'precise' so that the host knows exactly how to align the inputs & outputs. Aligning to `32x32` in the last two dims is indeed going to cause more wastage but currently it's the easiest way to unblock the bcast & reduce work.

In the existing logic, `tensor<2x33x32>` will have collapsed shape of `66x32`, which gets aligned up to `96x32`.

While this is efficient and works for eltwise ops, it won't work for bcast/reduce because the middle one of the three CB tiles you get from this strategy will have data from both of the `33x32` logical slices, and it will be difficult/impossible for the kernel to proceed from there.

Therefore I choose to align the 2nd dim up to `tensor<2x64x32>` which will have collapsed shape of `128x32`, so we have two `64x32` tile-blocks that holds the two `33x32` logical slices, and things are easy for bcast & reduce.

vwellsTT: Hmm, does that generalize? e.g. suppose we have a 4D tensor WxXxYxZ, and we want to broadcast along X.  Is 1x1x32x32 correct there? I think this a separate concern from the ragged tensor case, and should be solved by not collapsing (I have PR for this here: https://github.com/tenstorrent/tt-mlir/pull/4278 ).

runtime/include/tt/runtime/utils.h
wenbinlyuTT: There're many places that use this API without specifying the proper alignment, so I want to default align to the maximum elementary data type on the current system. I think this is safe enough for 99% of our needs (if someone want to instantiate a class/struct they will use `make_shared`, `mallocShared` is typically used to store numbers).

test/ttmlir/Conversion/TTIRToTTIRGeneric/alignments.mlir
wenbinlyuTT: Yeah, I want to make the alignments 'precise'. A `256x256` tensor will have `dim_alignments = 32x32` because 32 is the minimum required alignment for both dimensions. It's just more consistent in all related calculations.

runtime/include/tt/runtime/utils.h
vwellsTT: are there that many users for `malloc_shared`? On main I only see like 3--imo would be better to avoid default-param entirely and just pass in correct values for each case if that's possible

lib/Dialect/TTCore/IR/TTCoreOpsTypes.cpp
wenbinlyuTT: Yes, the old approach saves memory but breaks when bcast/reduce is involved. There's a '3' in '192' and it's gonna cause trouble :(

lib/Dialect/TTCore/IR/TTCoreOpsTypes.cpp
wenbinlyuTT: Same reasoning as I provided above, now the alignments are the 'minimum required amount'.

test/ttmlir/Conversion/TTIRToTTIRGeneric/alignments.mlir
vwellsTT: I don't care strongly, but I don't think I agree FWIW--imo 256 makes it more clear we've hit the grid boundary in this case.

lib/Dialect/TTMetal/Transforms/GenerateDimAlignedHostIO.cpp
wenbinlyuTT: I was afraid of getting the `walk` too long and too difficult to read. Plus, it's the 4th level of indention inside the `if`, v.s. the 3rd level inside the `for`.

vwellsTT: I would consider a pattern like 
```
 module->walk([&](Operation *op) {
      if (auto w = mlir::dyn_cast<ttmetal::EnqueueWriteBufferOp>(op)) {
        helperFuncToProcessWrite(w);
      } else if (auto r = mlir::dyn_cast<ttmetal::EnqueueReadBufferOp>(op)) {
        helperFuncToProcessRead(r);
      }
```
if you want to minimize indentation, iterating twice is unaesthetic imo

lib/Dialect/TTCore/IR/TTCoreOpsTypes.cpp
vwellsTT: well, for case which perfectly hits boundary anyways, minimum is 1 🙂

include/ttmlir/Dialect/TTCore/IR/TTCoreOpsTypes.td
wenbinlyuTT: Yeah, for bcast in higher dimensions we need to adjust the collapse intervals. My logic here will definitely be updated by then. I think bcast in the last two dims and in higher dims inherently require different approaches because the HW work with tiles.

vwellsTT: I don't think I'm following why we don't just leave this unchanged then, given this isn't really a general fix anyway?  I'd prefer to keep things optimized for 2D case for now, rely on other PRs to support higher dims + broadcasting etc.

wenbinlyuTT: This `1x1x...x32x32` change is pretty much the gist of the entire PR, but I'm totally open to suggestions about another way to achieve tile-alignment, whether logically or physically.

Consider the addition with implicit bcast `tensor<256x1x1> + tensor<256x32x32>`. The kernel API for this requires that the LHS tiles have exactly one number at the NW corner and zeros everywhere else.

The input tensor received by the runtime is just a [256] array, it needs to somehow create 256 32x32 tiles and put a single elements to the NW corner of each of these tiles. We need to give the runtime the correct information so that it knows to do 256 memcpy's with stride `1024 = 32x32`.

Using the old way, we will get `dim_alignments = 256x1x32`. It's not terribly clear to me how do we derive the correct strides for the 256 memcpy's from this, since the only seemingly un-matching dimension is the innermost 1 v.s. 32.

BTW, if my change kept the old `>=` logic and get `dim_alignments = 1x256x32` it's even more unclear how to interpret this info.

With this PR, the runtime sees shape `256x1x1` and `dim_alignments = 1x32x32`, and it's immediately clear how many tiles are needed and what the strides are.

This example is a bit extreme but I think it does show my point about how special the last two dims are. This restriction originates from our HW & kernel API and travels up our software stack.

Granted, I am not very familiar with the data movement generation pass. Maybe there's a way to achieve the said bcast operation on the device, even though we started from a single CB tile that contains all 256 elements?

lib/Dialect/TTMetal/Transforms/GenerateDimAlignedHostIO.cpp
wenbinlyuTT: Fixed

runtime/include/tt/runtime/utils.h
wenbinlyuTT: Fixed

lib/Dialect/TTCore/IR/TTCoreOpsTypes.cpp
```diff
@@ -820,44 +863,109 @@ llvm::SmallVector<int64_t> MetalLayoutAttr::getNormalizedIntervals() const {
...
+// However, this strategy has issues around sharding. Consider the same example
+// as above but with tensor<9x43x7>:
+// - Tile-aligned logical shape 9x64x32.
+// - Collapsed physical 2D shape 576x32.
+// - Dim alignments 4x32x32.
+// - Unsharded device shape [alignUp(9,4)*64]x32=768x32
+// - Grid & shard is 768x32 / 32x32 = 24x1.
+// - Tiled device shape 8x1x3x1x!tile<32x32>.
+// Similar to the example above, both result in 'unnatural' shard shapes like
+// 1x1x!tile<32x32> and 3x1x!tile<32x32>. The 'natual' shard shape is
```
vwellsTT: nit: natural spelling

lib/Dialect/TTCore/IR/TTCoreOpsTypes.cpp
```diff
@@ -820,44 +863,109 @@ llvm::SmallVector<int64_t> MetalLayoutAttr::getNormalizedIntervals() const {
                                       getLogicalShape().size());
 }
 
+// When the height of the physical 2D shape is collapsed from multiple logical
```
vwellsTT: Nice detailed comment 🙂

runtime/include/tt/runtime/utils.h
```diff
@@ -15,8 +17,25 @@
 
 namespace tt::runtime::utils {
 
-inline std::shared_ptr<void> malloc_shared(size_t size) {
-  return std::shared_ptr<void>(std::malloc(size), std::free);
+template <typename T>
```
vwellsTT: nit: debatably this could use same
`template <typename T, typename = std::enable_if_t<std::is_integral_v<T>>>` as below

lib/Dialect/TTMetal/Transforms/GenerateDimAlignedHostIO.cpp
```diff
@@ -0,0 +1,117 @@
...
+    const auto dimAlignments = llvm::to_vector(hostInfo.getDimAlignments());
+    op->removeAttr(hostInfoAttr);
+
+    auto hostLayoutAttr = ttcore::HostLayoutAttr::get(
+        module.getContext(), hostMemref.getShape(), dimAlignments);
+    return MemRefType::get(hostMemref.getShape(), hostMemref.getElementType(),
+                           hostLayoutAttr, hostMemref.getMemorySpace());
+  };
+
+  bool matchingHostDeviceShapes(MemRefType hostMemref,
```
vwellsTT: nit: perhaps this file could use some more func-level comments explaining basic ideas (though I see functionality is documented in tablegen descriptions, so maybe is fine)

include/ttmlir/Dialect/TTCore/IR/TTCoreOpsTypes.td
vwellsTT: Sync'ed offline, I think Wenbin is right and this is required for bcast unless metal supports some kind of striding, unfortunately.

runtime/include/tt/runtime/types.h
wenbinlyuTT: I think I will do this in another PR since it touches lots of places, along with other things..

runtime/lib/ttmetal/runtime.cpp
```diff
@@ -526,6 +499,53 @@ std::vector<Tensor> toHost(Tensor tensor, bool untilize, bool blocking) {
...
+  for (T i = 0; i < dimSize; i++) {
+    const T skip = i * strides[level];
+    for (size_t j = 0; j < prev.size(); j++) {
+      indices[i * prev.size() + j] = prev[j] + skip;
+    }
+  }
+  return indices;
+}
+
+void stridedMemcpy(const TensorDesc &dst, const TensorDesc &src, void *dstData,
```
jnie-TT: Might be nice for clarity to make this function static or put in an unnamed namespace/utils file if it's not going to be user-facing.

runtime/include/tt/runtime/utils.h
```diff
@@ -15,8 +17,25 @@
...
+template <typename T, typename = std::enable_if_t<std::is_integral_v<T>>>
+T alignUp(const T val, const T alignment) {
+  if (alignment == 0 || alignment == 1) {
+    return val;
+  }
+  return ((val + alignment - 1) / alignment) * alignment;
+}
+
+inline std::shared_ptr<void> mallocShared(const size_t size,
+                                          const size_t alignment) {
```
jnie-TT: I feel like this is a little cumbersome for cases where we don't care about alignment, could we maybe set alignment to a default value or overload the function? Also I believe malloc has some minimal alignment guarantee whereas we might override that with aligned_alloc

runtime/include/tt/runtime/utils.h
```diff
@@ -15,8 +17,25 @@
...
+  return std::shared_ptr<void>(
+      std::aligned_alloc(alignment, alignUp(size, alignment)), std::free);
+}
+
+inline std::shared_ptr<void> callocShared(const size_t size,
+                                          const size_t alignment) {
```
jnie-TT: Same here

runtime/include/tt/runtime/types.h
```diff
@@ -165,34 +163,55 @@ struct RuntimeCheckedConstObjectImpl {
...
+
+  TensorDesc(const std::vector<uint32_t> &shape,
+             ::tt::target::DataType dataType,
+             const std::vector<uint32_t> &dimAlignments)
+      : TensorDesc(shape, utils::calculateStride(shape), dataType,
+                   dimAlignments) {}
+
+  TensorDesc(const std::vector<uint32_t> &shape,
+             ::tt::target::DataType dataType)
+      : TensorDesc(shape, dataType, std::vector<uint32_t>(shape.size(), 1)) {}
```
jnie-TT: Might be cleaner to unify these constructor overloads into one by making stride, dataType, itemSize (might not even be needed), and alignment `std::optional`s and set to default values when they're not provided.

jnie-TT (approved):
Overall looks good, left a couple comments but nothing blocking

include/ttmlir/Dialect/TTCore/IR/TTCoreOpsTypes.td
nsmithtt: Hey Wenbin, but this is a function of which dimension you're broadcasting over right?  I think by default we still want collapse to look like 32x1x1x1x32.  We probably need to come up with a set of rules for programming the collapse based on bcast dim.

The following cases for example will still work with 32x1x1x1x32 style collapse (keeping dim -1, bcasting the rest):
- [1, 128], [128, 128] 
- [1, 1, 128], [5, 128, 128] 
- [1, 1, 1, 128], [3, 1, 2, 128] 

Whereas if you need to bcast over any other dim, you need to adjust the collapse:
- [1, 128, 1], [3, 128, 128] -> you must keep dim -2, so collapse intervals [(0, -3), (1, -2), (2, -1)] with alignments 1x1x32x32

So I think the scalar bcast and the column broadcast are special cases (albeit by far the most common) that are compatible with the default collapse strategy.  Once you start involving non-bcast dims outside of that innermost dim, we need to choose the strategy that you have here.

include/ttmlir/Utils.h
```diff
@@ -40,9 +40,19 @@ constexpr inline llvm::StringLiteral g_outputMemoryConfigAttrName =
     "memory_config";
 
 template <typename T>
-T alignUp(T ptr, T alignment) {
-  T distance = ptr % alignment;
-  return ptr + (distance == 0 ? 0 : (alignment - distance));
+T alignUp(const T val, const T alignment) {
+  if (alignment == 0 || alignment == 1) {
```
nsmithtt: maybe we should `assert(alignment > 0)`, that seems like a logical error to allow alignment of 0.

lib/Dialect/TTCore/IR/TTCoreOpsTypes.cpp
```diff
@@ -711,9 +755,8 @@ applyCollapsedIntervalsAndAlignments(llvm::ArrayRef<int64_t> shape,
 
       // Process remaining dimensions from inner to outer w/ multplication.
       for (int64_t j = end - 2; j >= start; --j) {
-        // For outer dimensions, multiply then align the product.
-        collapsedDim =
-            ttmlir::utils::alignUp(shape[j] * collapsedDim, alignments[j]);
+        // For outer dimensions, align and then multiply the product.
+        collapsedDim *= ttmlir::utils::alignUp(shape[j], alignments[j]);
```
nsmithtt: I don't think this is desirable, we need to multiply first, then align.

lib/Dialect/TTMetal/Transforms/GenerateDimAlignedHostIO.cpp
```diff
@@ -0,0 +1,131 @@
...
+namespace mlir::tt::ttmetal {
+#define GEN_PASS_DEF_TTMETALGENERATEDIMALIGNEDHOSTIO
+#include "ttmlir/Dialect/TTMetal/Transforms/Passes.h.inc"
+
+namespace {
+class TTMetalGenerateDimAlignedHostIO
+    : public impl::TTMetalGenerateDimAlignedHostIOBase<
+          TTMetalGenerateDimAlignedHostIO> {
+
+  static constexpr llvm::StringLiteral hostInfoAttr = "host_info";
```
nsmithtt: We might be able to just use the generated string name: `HostInfoAttr::name`, same with above.

lib/Dialect/TTMetal/Transforms/GenerateDimAlignedHostIO.cpp
```diff
@@ -0,0 +1,131 @@
...
+  void runOnOperation() final {
+    ModuleOp module = getOperation();
+    IRRewriter rewriter(&getContext());
+
+    module->walk([&](Operation *op) {
+      if (auto w = mlir::dyn_cast<ttmetal::EnqueueWriteBufferOp>(op)) {
+        rewriteEnqueueWrites(module, rewriter, w);
+      } else if (auto r = mlir::dyn_cast<ttmetal::EnqueueReadBufferOp>(op)) {
+        rewriteEnqueueReads(module, rewriter, r);
+      }
```
nsmithtt: Typically it's prefferred to use a PatternRewriter in conjunction with `walkAndApplyPatterns` which will do the `isa` check automatically and is generally easier to compose with other rewriters in the future.

include/ttmlir/Dialect/TTCore/IR/TTCoreOpsTypes.td
vwellsTT: Yeah, let's discuss in person tomorrow, but I think my understanding after chatting with Wenbin is this is sort of a stop-gap to make broadcasting easier for now, and in the future we can change things s.t. we do this iff we're broadcasting.  I think it's fair to question whether that makes sense or we should keep default and make broadcasting PR change this or what.

runtime/lib/ttmetal/runtime.cpp
wenbinlyuTT: Fixed

runtime/include/tt/runtime/utils.h
wenbinlyuTT: You're right, malloc has pretty generous default alignment guarantee. Switched to plain malloc.

runtime/include/tt/runtime/utils.h
wenbinlyuTT: Same, `calloc` has enough alignment guarantee, switched.

runtime/include/tt/runtime/types.h
wenbinlyuTT: Fixed

include/ttmlir/Utils.h
wenbinlyuTT: Fixed

lib/Dialect/TTMetal/Transforms/GenerateDimAlignedHostIO.cpp
wenbinlyuTT: Fixed

lib/Dialect/TTMetal/Transforms/GenerateDimAlignedHostIO.cpp
wenbinlyuTT: Fixed

lib/Dialect/TTCore/IR/TTCoreOpsTypes.cpp
wenbinlyuTT: Restored.

include/ttmlir/Dialect/TTCore/IR/TTCoreOpsTypes.td
```diff
@@ -263,6 +263,43 @@ def TTCore_ShardLayoutAttr : TTCore_Attr<"ShardLayout", "shard", [TTCore_DeviceL
   }];
 }
 
+def TTCore_HostLayoutAttr : TTCore_Attr<"HostLayout", "host_layout", [TTCore_DeviceLayoutInterface, MemRefLayoutAttrInterface]> {
+  let summary = "Host-side memref layout attribute with padding support.";
+  let description = [{
```
vwellsTT, line 268: I think the first sentence here should mention this is for a memref, not a tensor, similar to how the Shard/View/MetalLayoutAttrs do
(i.e. "Describes a host-side memref layout..." instead of "buffer")

vwellsTT: (Assuming I'm interpreting the code correctly, and this is only applied to memrefs)

include/ttmlir/Dialect/TTMetal/Transforms/Passes.h
```diff
@@ -5,6 +5,8 @@
 #ifndef TTMLIR_DIALECT_TTMETAL_TRANSFORMS_PASSES_H
 #define TTMLIR_DIALECT_TTMETAL_TRANSFORMS_PASSES_H
 
+#include "mlir/Pass/Pass.h"
```
vwellsTT: is this needed?

include/ttmlir/Utils.h
```diff
@@ -36,9 +36,20 @@ constexpr inline llvm::StringLiteral g_decomposedFromAllReduceAttrName =
     "ttir.decomposed_from_all_reduce";
 
 template <typename T>
-T alignUp(T ptr, T alignment) {
-  T distance = ptr % alignment;
-  return ptr + (distance == 0 ? 0 : (alignment - distance));
+T alignUp(const T val, const T alignment) {
+  assert(alignment > 0);
+  if (alignment == 1) {
```
vwellsTT: nit: personally prefer removing the `if` statement, 1 work like any other case below

lib/Dialect/TTCore/IR/TTCoreOpsTypes.cpp
```diff
@@ -577,6 +577,51 @@ ViewLayoutAttr ViewLayoutAttr::compose(ViewLayoutAttr g) const {
   return get(getContext(), getAffineMap().compose(g.getAffineMap()));
 }
 
+mlir::AffineMap HostLayoutAttr::getAffineMap() const {
```
vwellsTT, line 580: Can this be commonized with `ShardLayoutAttr::getAffineMap()`? It seems ~identical

test/ttmlir/Dialect/TTIR/metal_layout_misc.mlir
```diff
@@ -48,7 +47,7 @@ func.func @add_3d_unaligned(%arg0: tensor<2x33x128xf32>, %arg1: tensor<2x33x128x
 //  For 3D tensors with shape 4x64x256:
 //  - Logical shape: 4x64x256
 //  - After collapse [[0,2], [2,3]]: dims 0,1 -> 256, dim 2 -> 256
-//  - With alignment 32x1x32: 256 stays 256 (8*32), 256 stays 256 (8*32)
```
vwellsTT, line 51: it would be nice if we could make some kind of test to confirm that we can still use the old collapse logic, like
`collapsedIntervals =[[0,2],[2,3]], dim_alignments=32x1x32` and a `2x16x32` tensor -> `1x1x1x1x!tile<32x32>`
but I'm not sure if we have a codepath which could do this today or not

wenbinlyuTT: Good point! I guess it's difficult to do this today because the alignment calculation and the sharded shape calculation logic are fixed and tightly coupled in the same rewriter right?
It's still unclear to me how should the ttir-to-ttirgeneric pass select different alignment & collapse strategy.
@nsmithtt wdyt?

vwellsTT: Yeah, the only test-case I can think of is like the `lower_to_layout.mlir`, which should take in hard-coded IR w/ these fields, so might be of some value to put it there.  But it won't execute e2e anyway so I'm not sure how much it buys us.

lib/Dialect/TTCore/IR/TTCoreOpsTypes.cpp
wenbinlyuTT: They're too different to be reused I think.
The host layout one is pretty 'regular', both shape & strides are rank-N and the resulting map is N-to-(N+1).
The shard layout one is a bit odd, it starts from the rank-2 shard strides and produce `(d0, d1, d2, d3) -> (d0, d1, d2 * X + d3 * Y)`, a 4-to-3 map. Note how in the second loop, it does `getAffineDimExpr(rank + i, context)` to go to 4D.

include/ttmlir/Utils.h
wenbinlyuTT: Fixed.

include/ttmlir/Dialect/TTMetal/Transforms/Passes.h
wenbinlyuTT: Leftover from the previous impl.

include/ttmlir/Dialect/TTCore/IR/TTCoreOpsTypes.td
wenbinlyuTT: Fixed.

include/ttmlir/Dialect/TTIR/IR/TTIROps.h
```diff
@@ -46,7 +46,8 @@ inline void getDpsEffects(
   }
 }
 
-MemRefType getBufferType(Type type, bool isView);
+MemRefType getBufferType(Type type, bool isView,
+                         std::optional<ttcore::MetalLayoutAttr> hostInfo);
```
nsmithtt: Should we default to null opt?

lib/Dialect/TTMetal/Pipelines/TTMetalPipelines.cpp
```diff
@@ -85,7 +85,6 @@ void createTTIRToTTMetalFrontendPipeline(
   }
   pm.addPass(tt::createTTIRToTTIRGenericPass(toTTIRGenericOptions));
   pm.addPass(createCanonicalizerPassWithOptions(options));
-  pm.addPass(createCanonicalizerPassWithOptions(options));
```
nsmithtt, line 88: ty

runtime/include/tt/runtime/types.h
```diff
@@ -165,35 +163,37 @@ struct RuntimeCheckedConstObjectImpl {
...
-  std::uint32_t itemsize;
-  ::tt::target::DataType dataType;
-  std::int64_t alignment = 1;
+  std::vector<uint32_t> shape = {}; // Logical
+  ::tt::target::DataType dataType = ::tt::target::DataType::MAX;
+  uint32_t itemsize = 0;
+  std::vector<uint32_t> stride = {}; // Logical
+
+  // Members below are for metal tensors only.
+  std::vector<uint32_t> physicalStrides = {};
```
nsmithtt: Why is this different from `stride`?

test/ttmlir/Dialect/TTIR/metal_layout_misc.mlir
nsmithtt: Yeah using hard coded alignments in an mlir or builder test would be good!

lib/Dialect/TTIR/IR/TTIROps.cpp
```diff
@@ -2811,9 +2845,46 @@ mlir::LogicalResult mlir::tt::ttir::ToLayoutOp::bufferize(
...
+    MemRefType alignedHostMemref = mlir::cast<MemRefType>(
+        *getBufferType(getInput(), options, state, invocationStack));
+
+    if (mlir::cast<ttcore::HostLayoutAttr>(alignedHostMemref.getLayout())
+            .isPadded()) {
+      auto alignedHostTensor =
+          rewriter.create<memref::AllocOp>(getLoc(), alignedHostMemref);
+      rewriter.create<memref::CopyOp>(getLoc(), *maybeInput, alignedHostTensor);
+      maybeInput = alignedHostTensor.getResult();
+    }
```
nsmithtt, line 2859: Maybe this could be made a utility top level function since it's essentially the same logic below just with src/dst flipped.

lib/Dialect/TTCore/IR/TTCoreOpsTypes.cpp
```diff
@@ -974,6 +1074,52 @@ MetalLayoutAttr::getShardStride(RankedTensorType tensorType) const {
   return shardStride;
 }
 
+std::pair<llvm::SmallVector<int64_t>, int64_t>
+MetalLayoutAttr::getHostStrideAndVolume() const {
+  const auto logicalShape = getLogicalShape();
+  const auto alignments = getDimAlignments();
+  const auto normalizedIntervals = getNormalizedIntervals();
+  assert(normalizedIntervals.size() == 4);
```
nsmithtt: Why are we constraining the interval rank?  Before we supported intervals of any rank.

lib/Dialect/TTCore/IR/TTCoreOpsTypes.cpp
nsmithtt: I don't understand this host map tbh, let's say you had a rank 3 tensor, it reads like it'd output a map `(d0, d1, d2) -> (d0, d1, d2, d0 * s1 * s2 + d1 * s2 + d2)`?  Why is the result map rank + 1, why have the identity dims in there?

wenbinlyuTT: OK this was copied & modified from the shard layout one and I didn't know what was I doing. After checking https://mlir.llvm.org/docs/Dialects/Builtin/#layout , indeed `MemRefLayoutAttrInterface::getAffineMap` only needs the linearization affine map. I will change this to return only the last expression.

runtime/include/tt/runtime/types.h
wenbinlyuTT: We have [many](https://github.com/tenstorrent/tt-mlir/blob/cbf538d12fda78a7d2427182ccb2e9489ec2dcb4/runtime/lib/ttnn/runtime.cpp#L449) [different](https://github.com/tenstorrent/tt-mlir/blob/cbf538d12fda78a7d2427182ccb2e9489ec2dcb4/runtime/lib/ttmetal/runtime.cpp#L156) [places](https://github.com/tenstorrent/tt-mlir/blob/cbf538d12fda78a7d2427182ccb2e9489ec2dcb4/runtime/lib/ttmetal/runtime.cpp#L204) that passes/uses this `stride`. I am not feeling very safe to change this and risk getting this PR to snowball into including ttnn parts.
I was planning to do a follow-up PR to do some renaming & cleanup, like I mentioned in https://github.com/tenstorrent/tt-mlir/pull/4709#discussion_r2310817676

lib/Dialect/TTIR/IR/TTIROps.cpp
wenbinlyuTT: The insertion points & order are different, the H2D path reads from `maybeInput`, while the D2H path writes to `maybeOutput` and replaces the dst operand of the `toLayout`.
I feel like the extracted common util function will have too many input args and lots of if-else's, in the end we make this function easier to read but create another convoluted one.

lib/Dialect/TTCore/IR/TTCoreOpsTypes.cpp
wenbinlyuTT: Removed.

include/ttmlir/Dialect/TTIR/IR/TTIROps.h
wenbinlyuTT: Fixed.

runtime/include/tt/runtime/types.h
nsmithtt: I think there's little risk of this leaking into the TTNN side because they will program this field differently and that's OK.  I'd really prefer to use the same member up front, it's one of those things that will become low priority and in 6 months will be way harder to revert back to a single stride.

Try unifying it and kicking a CI run, if there's unexpected TTNN fallout we can revisit.

test/ttmlir/Dialect/TTIR/metal_layout_misc.mlir
wenbinlyuTT: This still turns out to be difficult: if I test tilize/untilize in the lower-to-layout pass, the `SplitCompoundLayoutRewriter` will invoke `getOrCreate{Input,Output}Layout()`, which creates the new `1x32x32` layout for `tensor<2x16x32>`. So then the op becomes a change-of-layout when it encounters my manually specified `32x1x32` alignment and asserts.

I was only able to test reblock like `<1x1x8x1x!tile<32x32>> -> <8x1x1x1x!tile<32x32>>` for `tensor<13x16x32>`, and it's an almost pointless test since there's no alignment-related calculation occur.

Any ideas?

runtime/include/tt/runtime/types.h
wenbinlyuTT: Unified and it seems to be working fine.

test/ttmlir/Dialect/TTIR/metal_layout_misc.mlir
wenbinlyuTT: Opened #4852 to track

nsmithtt (approved):
Awesome work, thank you for trying various approaches, I think we landed on something pretty robust!

PR #4718, opened 2025-08-28
  Add TTNNOperationValidationAndFallback pass with basic validation

### LINKED ISSUE #4623: Add a pass that validates every non-checked op in IR
Today we have hard-coded work arounds for known issues for various ops in TTNN. However, we aim to avoid this hack because it interferes with optimizer. This affects validity of certain ops. This issue proposes to add a pass that checks validity of every (already) non-checked op in the IR. This check should be performed using OpModel interface.

rpavlovicTT:
### Ticket
Closes #4623 

### Description

  This commit introduces a new pass that validates TTNN operations using the OpConstraintValidator. The pass walks through all operations in a module and checks if they satisfy their constraints with the original input layouts. Operations that fail validation are currently marked as "fixed" with a placeholder - actual fallback logic will be added in a subsequent commit.

  Key features:
  - Validates operations against their original input/output configurations
  - Logs validation statistics (total checked, valid, fixed, failed)
  - Uses OpConstraintValidator with fatalOnUnsupported=false
  - Includes extractOpConfigFromIR helper for configuration extraction
  - Placeholder for fallback handling (currently marks failed ops as fixed)


lib/Dialect/TTNN/Transforms/OperationValidationAndFallback.cpp
```diff
@@ -0,0 +1,148 @@
+// SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
```
odjuricicTT: We need to adjust codeowners for this as well. Maybe extract this, prepare_conv2d_weights and optimizer pass into a subdir?

include/ttmlir/Dialect/TTNN/Transforms/Passes.td
```diff
@@ -216,4 +216,14 @@ def TTNNUniqueLocations: Pass<"ttnn-unique-locations", "::mlir::ModuleOp">
   let description = "This pass checks if all operations relevant to the optimizer have unique locations. If not, it will emit an error. This is necessary for the overrides to be applied correctly.";
 }
 
+def TTNNOperationValidationAndFallback: Pass<"ttnn-operation-validation-and-fallback", "::mlir::ModuleOp"> {
```
odjuricicTT, line 219: This is not added to the pipeline in this PR?

lib/Dialect/TTNN/Transforms/OperationValidationAndFallback.cpp
```diff
@@ -0,0 +1,148 @@
...
+                 "operation result type encoding";
+          signalPassFailure();
+          return;
+        }
+
+        // Extract input layouts from the operation
+        std::vector<TTNNLayoutAttr> inputLayouts =
+            utils::extractInputLayouts(operation);
+
+        if (inputLayouts.empty()) {
```
odjuricicTT: Add a comment why this is fine.

lib/Dialect/TTNN/Transforms/OperationValidationAndFallback.cpp
```diff
@@ -0,0 +1,148 @@
...
+    size_t operationsValid = 0;
+    size_t operationsFixed = 0;
+    size_t operationsFailed = 0;
+
+    // Create validator configured not to throw on unsupported ops.
+    OpConstraintValidator validator(
+        OpConstraintValidator::ValidationOptions(/*fatalOnUnsupported=*/false));
+
+    moduleOp->walk([&](func::FuncOp func) {
+      if (ttmlir::utils::isConstEvalFunc(func)) {
```
odjuricicTT: Should we not validate const-eval funcs? It could contains ops that need layout changes?

lib/Dialect/TTNN/Transforms/OperationValidationAndFallback.cpp
```diff
@@ -0,0 +1,148 @@
...
+        return;
+      }
+
+      // llvm::SmallVector<Type> funcResultTypes;
```
odjuricicTT: Remove.

lib/Dialect/TTNN/Transforms/OperationValidationAndFallback.cpp
```diff
@@ -0,0 +1,148 @@
...
+
+      func.walk([&](Operation *operation) {
+        // Skip operations without results
+        if (operation->getNumResults() == 0) {
+          return;
+        }
+
+        // Skip operations that are not OpModel castable (can't be validated)
+        if (!mlir::dyn_cast<OpModel>(operation)) {
+          return;
```
odjuricicTT: Is this what we want to do? After we have constraints for all ops, maybe this should be changed.

odjuricicTT (commented):
Looks good!

include/ttmlir/Dialect/TTNN/Transforms/Passes.td
rpavlovicTT: It is added in the subsequent PR

lib/Dialect/TTNN/Transforms/OperationValidationAndFallback.cpp
rpavlovicTT: I can leave a TODO that we expect at some point to cover everything, but at the moment we can't fail until we have constraints for all the ops.

lib/Dialect/TTNN/Transforms/OperationValidationAndFallback.cpp
rpavlovicTT: Yeah makes sense. Initially I had this pass as analysis within Optimizer where we touch only non-const-eval function, so this condition was left behind. I will enable it for const-eval too.

lib/Dialect/TTNN/Transforms/OperationValidationAndFallback.cpp
rpavlovicTT: ACtually I could remove this

lib/Dialect/TTNN/Transforms/OptimizerPasses/OperationValidationAndFallback.cpp
```diff
@@ -0,0 +1,136 @@
...
+        totalOperationsChecked++;
+
+        // Extract OpConfig from IR
+        OpConfig config = extractOpConfigFromIR(operation);
+        if (!config.outputLayout) {
+          operation->emitError()
+              << "OperationValidationAndFallback: No output layout found in "
+                 "operation result type encoding";
+          signalPassFailure();
+          return;
```
mtopalovicTT: return WalkResult::interrupt() here
and skip/advance on other places.

mtopalovicTT (approved):

lib/Dialect/TTNN/Transforms/OptimizerPasses/OperationValidationAndFallback.cpp
```diff
@@ -0,0 +1,138 @@
...
+#include "llvm/ADT/SmallVector.h"
+#include "llvm/Support/Error.h"
+
+#include <cassert>
+#include <cstddef>
+#include <vector>
+
+using namespace mlir;
+using namespace mlir::tt;
+using namespace mlir::tt::ttnn;
```
azecevic000, line 23: These `using`s don't make sense when all of the code is defined in `mlir::tt::ttnn` ns.

rpavlovicTT: will remove it in subsequent commit

PR #4719, opened 2025-08-28
Add TTNN Hand-tuned Resnet Models

vcanicTT:
These tests and files represent hand-tuned models generated with EmitC and then manually modified. With this type of testing, we aim to achieve two goals: first, to evaluate the EmitC tool in terms of usability and its support for manual optimization; and second, to improve the performance of specific RED models.


vcanicTT:
@azecevicTT The idea is to preserve hand-tuned models, and different versions of them, versions will be based on features introduced and changed during the tuning. For, now, these versions will not be executed on CI. Hand-tuning is something that helps us to figure out patterns during the optimization process, and then to transform those patterns to algorithms, features and passes in the optmizer. For, now, let's think about these models as helpers.

PR #4720, opened 2025-08-28
Add LayoutFallbackHandler with complete fallback implementation and t…

### LINKED ISSUE #4624: Add fallback mechanism for ops that are not valid after Optimizer pass
Some ops may not be valid (fail in runtime) after Optimizer pass. This can happen due to tensor layouts being incompatible with op's implementation in the backend. 
This issue proposes mechanism that brute forces all layouts possible for operand tensors of an operation, and finds a working one. This presents an automatic "workaround". 

- This pass should exist after Optimizer in ttir-to-ttnn pipeline. 
- This pass should aim to add minimal number of extra ops (tensor layout manipulations) to IR to limit negative perf effects. 
- This pass should be as generic as possible, not caring about op specific characteristics.

### LINKED ISSUE #4633: Remove existing workarounds from Optimizer pass
Currently there are workarounds for MaxPool2D and Upsample2D ops in Optimizer, but with automatic validation this should not be necessary anymore.

rpavlovicTT:
This commit implements the handling of fallbacks that provides comprehensive layout transformation capabilities when operations fail validation with their original configurations. 

Key features:
    - Distance-based optimization to try closest transformations first
    - Complete input/output layout transformation with ToLayoutOp insertion
    - Proper output layout reversion for downstream consumer 
     
Includes comprehensive test suite:
    - Layout fallback validation tests for multiple operation types
    - Tests cover argmax, binary ops, concat, embedding, pool2d, reshape,
    slice, sum, tanh dtype, and upsample operations
    - Validates that operations can be successfully transformed when original
    layouts fail validation

### Ticket
Closes #4624 
Closes #4633 

### Checklist


lib/Dialect/TTNN/Transforms/Optimizer.cpp
```diff
@@ -762,253 +760,6 @@ class TTNNOptimizer : public impl::TTNNOptimizerBase<TTNNOptimizer> {
     }
   }
 
-  // Check if the op can be executed with row major layout on the input. Returns
```
odjuricicTT: Nice :)

odjuricicTT (commented):
Did not get to take a deep look into this one. Leaving for vacation :).

odjuricicTT:
@sdjordjevicTT / @mtopalovicTT could take a look once the other PRs that this depends on are merged.

mtopalovicTT:
> @sdjordjevicTT / @mtopalovicTT could take a look once the other PRs that this depends on are merged.

will do

include/ttmlir/Dialect/TTNN/IR/TTNNOpsAttrs.td
```diff
@@ -662,6 +662,7 @@ def TTNN_TTNNLayoutAttr: TTNN_Attr<"TTNNLayout", "ttnn_layout"> {
     TTNNLayoutAttr withBufferType(BufferType bufferType);
     TTNNLayoutAttr withMemoryLayout(TensorMemoryLayoutAttr memLayoutAttr);
     TTNNLayoutAttr withMemoryLayout(TensorMemoryLayout memLayout);
+    TTNNLayoutAttr withLayout(Layout layout, ArrayRef<int64_t> tensorShape);
```
mtopalovicTT, line 666: Since you are touching things in encoding attribute can you please remove this:
```cpp
AffineMap replaceMemoryMapSymbolsWithShardShape(AffineMap physicalMemoryMap) const;
AffineMap getIdentityTileLinearMap() const;
```

lib/Dialect/TTNN/IR/TTNNOpsAttrs.cpp
```diff
@@ -968,3 +968,26 @@ DeviceComputeKernelConfigAttr::withDstFullSyncEn(bool value) const {
...
+                                                    tensorShape);
+  }
+
+  if (layout == Layout::Tile) {
+    Type elementType =
+        ttnn::utils::getElementType(getContext(), Layout::Tile, getDataType());
+    return withElementType(elementType, tensorShape);
+  }
+
+  // For any other layout, return the current layout unchanged
```
mtopalovicTT: llvm_unreachable here

lib/Dialect/TTNN/IR/TTNNOpsAttrs.cpp
```diff
@@ -968,3 +968,26 @@ DeviceComputeKernelConfigAttr::withDstFullSyncEn(bool value) const {
...
+// Transform TTNNLayoutAttr with a different layout while preserving the
+// element type.
+//
+// param layout The target layout (RowMajor or Tile).
+// param tensorShape The shape of the tensor.
+// return The new TTNNLayoutAttr with the specified layout.
+TTNNLayoutAttr TTNNLayoutAttr::withLayout(Layout layout,
+                                          ArrayRef<int64_t> tensorShape) {
+  if (layout == Layout::RowMajor) {
+    return ttnn::utils::convertTTNNLayoutToRowMajor(getContext(), *this,
```
mtopalovicTT: I'm more in favor of removing this utility function now that you added `withLayout` helper.
On sidenote I think you can merge these two if's like this:

```cpp
    if (layout == Layout::RowMajor || layout == Layout::Tile) {
        Type elementType =
            ttnn::utils::getElementType(getContext(), layout, getDataType());
        return withElementType(elementType, tensorShape);
     }
```

lib/Dialect/TTNN/Transforms/OptimizerPasses/OperationValidationAndFallback.cpp
```diff
@@ -135,4 +250,501 @@ class TTNNOperationValidationAndFallback
...
+
+  // Create new tensor type with updated layout
+  // Use scalar element type for RankedTensorType, not tile type
+  Type scalarElementType = mlir::tt::ttcore::dataTypeToElementType(
+      operation->getContext(), newLayoutAttr.getDataType());
+  RankedTensorType newTensorType = RankedTensorType::get(
+      currentTensorType.getShape(), scalarElementType, newLayoutAttr);
+
+  // Determine which layout to use for ToLayoutOp
+  Layout targetLayoutValue =
```
mtopalovicTT: this should be newLayoutAttr.getLayout(). Same for dataType below you get it from newLayoutAttr.getDataType

lib/Dialect/TTNN/Transforms/OptimizerPasses/OperationValidationAndFallback.cpp
```diff
@@ -135,4 +250,501 @@ class TTNNOperationValidationAndFallback
...
+      ttmlir::LogComponent::OpValidation,
+      "Applied input operand change for operation {} operand {}: "
+      "layout {} -> {}, memory layout {} -> {}, buffer type {} -> {}, data "
+      "type {} -> {}",
+      operation->getName(), change.operandIndex,
+      change.originalLayout.getLayout(),
+      change.targetLayout.value_or(change.originalLayout.getLayout()),
+      change.originalLayout.getMemLayout().getValue(),
+      change.targetMemoryLayout.value_or(
+          change.originalLayout.getMemLayout().getValue()),
```
mtopalovicTT: `change.originalLayout.getMemLayout().getValue()` this will throw if original layout doesn't have memLayout set

lib/Dialect/TTNN/Transforms/OptimizerPasses/OperationValidationAndFallback.cpp
```diff
@@ -135,4 +250,501 @@ class TTNNOperationValidationAndFallback
...
+
+bool tryFallbacks(Operation *operation,
+                  const std::vector<TTNNLayoutAttr> &originalInputLayouts,
+                  const OpConfig &config) {
+
+  // Extract tensor shapes for all input operands
+  std::vector<llvm::ArrayRef<int64_t>> tensorShapes;
+  for (Value operand : operation->getOperands()) {
+    if (auto tensorType = mlir::dyn_cast<RankedTensorType>(operand.getType())) {
+      if (auto layoutAttr = mlir::dyn_cast_or_null<TTNNLayoutAttr>(
```
mtopalovicTT: Just cast?

test/ttmlir/Dialect/TTNN/optimizer/op_layout_fallbacks/binary_ops_validation.mlir
```diff
@@ -0,0 +1,36 @@
...
+
+module attributes {} {
+  func.func @subtract_with_row_major_i32(%arg0: tensor<1x1x32x32xsi32, #ttnn_layout_row_major_si32>, %arg1: tensor<1x1x32x32xsi32, #ttnn_layout_row_major_si32>) -> tensor<1x1x32x32xbf16, #ttnn_layout_tile_bf16> {
+    %0 = "ttnn.get_device"() <{mesh_shape = #ttnn<mesh_shape 1x1>}> : () -> !ttnn.device
+
+    // The op validation pass should detect that op works with int32 inputs, but the output layout
+    // doesn't match the expected one, so it inserts a revert to tile layout.
+
+    // CHECK: "ttnn.subtract"
+    // CHECK: "ttnn.to_layout"
```
mtopalovicTT, line 19: I would add more rigorous checks here, since in the future there could be other pass which can transform the graph in a way which this tests doesn't expect which would result in uncaught error.
```mlir
// CHECK: %[[SUB_RES:.*]] = "ttnn.subtract"
// CHECK: "ttnn.to_layout"
// CHECK-SAME: dtype = #ttcore.supportedDataTypes<si32>
// CHECK-SAME: : (%[[SUB_RES]]
```

same for other tests

test/ttmlir/Dialect/TTNN/optimizer/op_layout_fallbacks/binary_ops_validation.mlir
```diff
@@ -0,0 +1,36 @@
...
+
+    %1 = "ttnn.subtract"(%arg0, %arg1) <{operandSegmentSizes = array<i32: 2, 0>, dtype = #ttcore.supportedDataTypes<bf16>}> : (tensor<1x1x32x32xsi32, #ttnn_layout_row_major_si32>, tensor<1x1x32x32xsi32, #ttnn_layout_row_major_si32>) -> tensor<1x1x32x32xbf16, #ttnn_layout_tile_bf16>
+
+    return %1 : tensor<1x1x32x32xbf16, #ttnn_layout_tile_bf16>
+  }
+
+  func.func @add_with_row_major_inputs(%arg0: tensor<1x1x32x32xbf16, #ttnn_layout_row_major_bf16>, %arg1: tensor<1x1x32x32xbf16, #ttnn_layout_row_major_bf16>) -> tensor<1x1x32x32xbf16, #ttnn_layout_tile_bf16> {
+    %0 = "ttnn.get_device"() <{mesh_shape = #ttnn<mesh_shape 1x1>}> : () -> !ttnn.device
+
+    // CHECK: "ttnn.add"
```
mtopalovicTT: Can you add comment on why to_layout is needed on the result of add (i.e adding two row row major tensors outputs row major tensor so we need to revert to tile after the add)

test/ttmlir/Dialect/TTNN/optimizer/op_layout_fallbacks/embedding_validation.mlir
```diff
@@ -0,0 +1,40 @@
...
+
+    // CHECK: "ttnn.to_layout"
+    // CHECK: "ttnn.embedding"
+
+    %1 = "ttnn.embedding"(%arg0, %arg1) : (tensor<32x10xui32, #ttnn_layout_tile_ui32>, tensor<100x32xf32, #ttnn_layout_tile_f32>) -> tensor<32x10x32xbf16, #ttnn_layout_row_major_bf16>
+
+    return %1 : tensor<32x10x32xbf16, #ttnn_layout_row_major_bf16>
+  }
+
+  func.func @embedding_with_tile_output(%arg0: tensor<32x32xui32, #ttnn_layout_tile_input>, %arg1: tensor<1000x32xf32, #ttnn_layout_tile_f32>) -> tensor<32x32x32xbf16, #ttnn_layout_tile_output> {
```
mtopalovicTT: Do you need this one since there is no output changes on the embedding itself?

lib/Dialect/TTNN/Transforms/OptimizerPasses/OperationValidationAndFallback.cpp
```diff
@@ -18,14 +19,86 @@
...
 
-using namespace mlir;
-using namespace mlir::tt;
-using namespace mlir::tt::ttnn;
-
 namespace mlir::tt::ttnn {
 #define GEN_PASS_DEF_TTNNOPERATIONVALIDATIONANDFALLBACK
 #include "ttmlir/Dialect/TTNN/Transforms/Passes.h.inc"
 
+namespace fallbacks {
```
mtopalovicTT, line 28: Can we move all declarations for mlir::tt::ttnn::fallbacks to OpConstaintsValidation.h?

lib/Dialect/TTNN/Transforms/OptimizerPasses/OperationValidationAndFallback.cpp
```diff
@@ -135,4 +250,501 @@ class TTNNOperationValidationAndFallback
...
+  for (size_t fallbackIdx = 0;
+       fallbackIdx < operandFallbacks[operandIdx].size(); ++fallbackIdx) {
+    currentCombination[operandIdx] = operandFallbacks[operandIdx][fallbackIdx];
+
+    double addedDistance = calculateLayoutTransformDistance(
+        originalInputLayouts[operandIdx],
+        operandFallbacks[operandIdx][fallbackIdx]);
+
+    // Go to the next operand, accumulating distance along the way.
+    generateAllCombinations(operandFallbacks, originalInputLayouts,
```
mtopalovicTT, line 517: Did you run maybe resnet or some other bigger model? How frequently do we end up in situations where we need to fixup the inputs on the operation.

lib/Dialect/TTNN/Transforms/OptimizerPasses/OperationValidationAndFallback.cpp
```diff
@@ -73,20 +152,56 @@ class TTNNOperationValidationAndFallback
...
+          llvm::consumeError(originalResult.takeError());
+
+          // Original config failed, try fallback configurations
+          if (fallbacks::tryFallbacks(operation, inputLayouts, config)) {
+            operationsFixed++;
+            TTMLIR_DEBUG(ttmlir::LogComponent::OpValidation,
+                         "Operation {} at {} fixed with fallback configuration",
+                         operation->getName(), operation->getLoc());
+          } else {
+            operationsFailed++;
```
mtopalovicTT: Do we stop here?

mtopalovicTT: We don't use operationsFailed operationsValid

lib/Dialect/TTNN/IR/TTNNOpsAttrs.cpp
```diff
@@ -183,32 +183,54 @@ TTNNLayoutAttr::calculateLogicalShardShapeForSharding(
 // tensor memory layout.
 //
 // All examples assume that the tensor is mapped to a 8x8 grid.
+// Examples for TileType:
 // Example: tensor<1x1024xbf16> ( -> 32 tiles ) -> {1, 1}
```
mtopalovicTT: `( -> 32 tiles ) -> {1, 1}`
Should this be `(-> 32 tiles) -> {32, 32}` since you are returning scalarShape in case of tile type?

lib/Dialect/TTNN/Transforms/OptimizerPasses/OperationValidationAndFallback.cpp
```diff
@@ -135,4 +250,501 @@ class TTNNOperationValidationAndFallback
...
+  if (fromIsFloat && toIsFloat) {
+    if (isUpcast(from, to)) {
+      return MID_COST; // Upcast (BFloat16 -> Float32) is more expensive
+    }
+    return LOW_COST; // Downcast (Float32 -> BFloat16) is cheaper
+  }
+
+  // Cross-category changes are more expensive as they are not safe conversions
+  // wrt. precision
+  return MID_COST;
```
mtopalovicTT: Something like this more clean mb?
```suggestion
double calculateDataTypeDistance(ttcore::DataType from, ttcore::DataType to) {
  if (from == to) {
    return NO_COST;
  }

  enum class TypeCategory { Integer, Float };
  enum class PrecisionLevel { Low, High };

  auto getTypeInfo = [](ttcore::DataType dtype) -> std::pair<TypeCategory, PrecisionLevel> {
    switch (dtype) {
      case ttcore::DataType::Int32:
      case ttcore::DataType::UInt32:
        return {TypeCategory::Integer, PrecisionLevel::High};
      case ttcore::DataType::BFloat16:
        return {TypeCategory::Float, PrecisionLevel::Low};
      case ttcore::DataType::Float32:
        return {TypeCategory::Float, PrecisionLevel::High};
      default:
        // Default to high cost for unknown types
        return {TypeCategory::Float, PrecisionLevel::High};
    }
  };

  auto [fromCategory, fromPrecision] = getTypeInfo(from);
  auto [toCategory, toPrecision] = getTypeInfo(to);

  // Same category conversions
  if (fromCategory == toCategory) {
    if (fromPrecision == toPrecision) {
      return LOW_COST; // Lateral moves within same precision (e.g., Int32 <-> UInt32)
    }
    // Precision changes within same category
    return (fromPrecision < toPrecision) ? MID_COST : LOW_COST; // Upcast vs downcast
  }

  // Cross-category conversions are most expensive (int <-> float)
  return HIGH_COST;
}
```

lib/Dialect/TTNN/Transforms/OptimizerPasses/OperationValidationAndFallback.cpp
```diff
@@ -135,4 +250,501 @@ class TTNNOperationValidationAndFallback
...
+      "Testing fallback combinations for operation {} at {} with {} operands",
+      operation->getName(), operation->getLoc(), originalInputLayouts.size());
+
+  // Create fallback combinations for each operand
+  std::vector<std::vector<TTNNLayoutAttr>> operandFallbacks;
+  for (size_t i = 0; i < originalInputLayouts.size(); ++i) {
+    auto fallbacks =
+        createFallbackTransforms(originalInputLayouts[i], tensorShapes[i]);
+    // Insert original layout at the beginning
+    fallbacks.insert(fallbacks.begin(), originalInputLayouts[i]);
```
mtopalovicTT: Why at the beginning? I don't see that order matters downstream and inserting at the beginning adds to time complexity.

lib/Dialect/TTNN/Transforms/OptimizerPasses/OperationValidationAndFallback.cpp
```diff
@@ -135,4 +250,501 @@ class TTNNOperationValidationAndFallback
...
+  ttcore::DataTypeAttr dtypeAttr = nullptr;
+  if (change.targetDataType.has_value()) {
+    dtypeAttr = ttcore::DataTypeAttr::get(operation->getContext(),
+                                          change.targetDataType.value());
+  } else {
+    dtypeAttr = ttcore::DataTypeAttr::get(operation->getContext(),
+                                          currentLayoutAttr.getDataType());
+  }
+
+  auto toLayoutOp = builder.create<ToLayoutOp>(
```
mtopalovicTT: How we create tolayout op for input is the same as for output, so we should be able to extract it to common function and use it in both functions to remove duplicates.
```cpp
ToLayoutOp createToLayoutOp(OpBuilder &builder, Location loc, Value input,
                            TTNNLayoutAttr targetLayout,
                            llvm::ArrayRef<int64_t> tensorShape) {
  auto inputTensorType = mlir::cast<RankedTensorType>(input.getType());
  
  // Create the target result type
  Type targetElementType = mlir::tt::ttcore::dataTypeToElementType(
      builder.getContext(), targetLayout.getDataType());
  RankedTensorType targetTensorType = RankedTensorType::get(
      inputTensorType.getShape(), targetElementType, targetLayout);

  // Create MemoryConfigAttr for the target layout
  MemoryConfigAttr memoryConfigAttr = MemoryConfigAttr::get(
      builder.getContext(), targetLayout.getMemLayout(),
      BufferTypeAttr::get(builder.getContext(), targetLayout.getBufferType()),
      /*shardSpec=*/std::nullopt);

  // Create DataTypeAttr
  ttcore::DataTypeAttr dtypeAttr = ttcore::DataTypeAttr::get(
      builder.getContext(), targetLayout.getDataType());

  // Create ToLayoutOp
  return builder.create<ToLayoutOp>(
      loc, targetTensorType, input,
      LayoutAttr::get(builder.getContext(), targetLayout.getLayout()),
      dtypeAttr, memoryConfigAttr);
}
```
What do you think?

lib/Dialect/TTNN/Transforms/OptimizerPasses/OperationValidationAndFallback.cpp
```diff
@@ -135,4 +250,501 @@ class TTNNOperationValidationAndFallback
...
+
+    // Step 2: Add revert ToLayoutOp to convert back to expected layout for
+    // consumers
+    applyOutputLayoutRevert(operation, result.actualOutputLayout,
+                            config.outputLayout);
+  }
+}
+
+// Apply a single input operand change by inserting ToLayoutOp
+void applyInputOperandChange(Operation *operation,
```
mtopalovicTT: what if we changed `applyInputOperandChange` signature to accept
`Operation* operation, size_t operandIndex, TTNNLayoutAttr encoding` and just call `toLayoutOp` helper function which would create `ToLayoutOp` given the input value and encoding. That way you could remove this:

```cpp
    // Create InputOperandChange object and populate it
    InputOperandChange change(i, originalLayout);

    // Detect layout changes
    if (originalLayout.getLayout() != transformedLayout.getLayout()) {
      change.targetLayout = transformedLayout.getLayout();
    }

    // Detect memory layout changes
    if (originalLayout.getMemLayout() != transformedLayout.getMemLayout()) {
      change.targetMemoryLayout = transformedLayout.getMemLayout().getValue();
    }

    // Detect buffer type changes
    if (originalLayout.getBufferType() != transformedLayout.getBufferType()) {
      change.targetBufferType = transformedLayout.getBufferType();
    }

    // Detect data type changes
    if (originalLayout.getDataType() != transformedLayout.getDataType()) {
      change.targetDataType = transformedLayout.getDataType();
    }

    if (change.hasChanges()) {
      applyInputOperandChange(operation, change);
    }
```

test/ttmlir/Dialect/TTNN/optimizer/op_layout_fallbacks/embedding_validation.mlir
rpavlovicTT: Updated this test to expect both input operands to be changed to BF16.

lib/Dialect/TTNN/Transforms/OptimizerPasses/OperationValidationAndFallback.cpp
rpavlovicTT: fallbacks are coupled to this pass, while  `OpConstaintsValidation.h` is a separate lib simply for validation of op using getOpConstraints APIs.

lib/Dialect/TTNN/Transforms/OptimizerPasses/OperationValidationAndFallback.cpp
rpavlovicTT: I did run big models, usually we need to update operands in max pool, upsample, embedding ops. Those are most frequent. But they are rare in models, meaning only few such ops occur among thousands of others.

lib/Dialect/TTNN/Transforms/OptimizerPasses/OperationValidationAndFallback.cpp
rpavlovicTT: Yeah I can stop here

lib/Dialect/TTNN/IR/TTNNOpsAttrs.cpp
rpavlovicTT: Yes, fixed it

lib/Dialect/TTNN/Transforms/OptimizerPasses/OperationValidationAndFallback.cpp
rpavlovicTT: Thanks for suggesting, I refactored it in a way like you suggested.

lib/Dialect/TTNN/Transforms/OptimizerPasses/OperationValidationAndFallback.cpp
rpavlovicTT: Yeah this makes it much simpler, thanks.

lib/Dialect/TTNN/Transforms/OptimizerPasses/OperationValidationAndFallback.cpp
rpavlovicTT: I agree, I like when we can modularise code to such functions.

lib/Dialect/TTNN/Transforms/OptimizerPasses/OperationValidationAndFallback.cpp
mtopalovicTT: Sorry I didn't check the filename when I wrote the comment. Meant to say OperationVlidationAndFallback.h. To leave only implementation here. Its just nit picking really, you don't have to follow the suggestion :D

Issue #4721, opened 2025-08-28
Mean and Squeeze TTIRToLinalg


djstefanovicTT:
Lower `ttir.squeeze` and `ttir.mean` to Linalg using Tosa.

PR #4723, opened 2025-08-28
Adding Mean and Squeeze ops to TTIRToLinalg conversion

djstefanovicTT:
### Ticket
https://github.com/tenstorrent/tt-mlir/issues/4721

### Problem description
`ttir.squeeze` and `ttir.mean`  ops were not implemented in conversion from ttir to linalg.

### What's changed
`ttir.squeeze` is lowered using `tosa.reshape`. 
`ttir.mean` is lowered using `tosa.reduce_sum`. This op reduces a tensor along the given axis by computing the sum of the axis. Then `linalg.div` is applied to calculate mean.
Fixed bug in `ttir.reluOp` where it did not support element types such as `bf16`.
Fixed bug in `createReductionOpChain` function where it did not work if keepDim is false.

### Checklist


lib/Conversion/TTIRToLinalg/TTIRToLinalg.cpp
```diff
@@ -1654,6 +1664,114 @@ class ConstantOpConversionPattern
...
+    bool keepDim = getKeepDimFromAttribute(op);
+
+    Value sum = createReductionOpChain<tosa::ReduceSumOp>(
+        input, resultType, dims, keepDim, op.getLoc(), rewriter);
+
+    int64_t numElements = 1;
+    auto inputShape = inputType.getShape();
+    for (int64_t dim : dims) {
+      numElements *= inputShape[dim];
+    }
```
azecevic000, line 1709: Nit: `ttmlir::utils::volume(inputType.getShape())`

lib/Conversion/TTIRToLinalg/TTIRToLinalg.cpp
```diff
@@ -1654,6 +1664,114 @@ class ConstantOpConversionPattern
...
+          resultType, APFloat(cast<FloatType>(elementType).getFloatSemantics(),
+                              numElements));
+    } else if (isa<IntegerType>(elementType)) {
+      divisorAttr = DenseElementsAttr::get(
+          resultType,
+          APInt(cast<IntegerType>(elementType).getWidth(), numElements));
+    } else {
+      return rewriter.notifyMatchFailure(op,
+                                         "Unsupported element type for mean");
+    }
```
azecevic000: This same logic repeats in the relu op, you can write some helper function for both.

lib/Conversion/TTIRToLinalg/TTIRToLinalg.cpp
djstefanovicTT: Updated.

lib/Conversion/TTIRToLinalg/TTIRToLinalg.cpp
djstefanovicTT: Moved into a separate helper function.

lib/Conversion/TTIRToLinalg/TTIRToLinalg.cpp
```diff
@@ -263,36 +263,50 @@ static Value createReductionOpChain(Value input, RankedTensorType resultType,
...
-    }
+    shape[dim] = 1;
+    opResultType = RankedTensorType::get(shape, inputType.getElementType());
 
     // Create the reduction operation
     result = rewriter.create<ReductionOp>(loc, opResultType, result, axisAttr);
   }
-
+  if (!keepDim) {
+    auto newShape = resultType.getShape();
```
vwellsTT: nit: better not to use auto here imo

lib/Conversion/TTIRToLinalg/TTIRToLinalg.cpp
```diff
@@ -263,36 +263,50 @@ static Value createReductionOpChain(Value input, RankedTensorType resultType,
...
+    SmallVector<int64_t> newShapeValues(newShape.begin(), newShape.end());
```
vwellsTT: I assume this is needed because newShape is an arrayRef?  But even so, it's not clear to me why you can't use arrayRef below, does `getIndexTensorAttr` not accept it or something?

lib/Conversion/TTIRToLinalg/TTIRToLinalg.cpp
```diff
@@ -1448,10 +1462,15 @@ class ReluOpConversionPattern : public OpConversionPattern<ttir::ReluOp> {
         this->getTypeConverter()->convertType(op.getType()));
 
     assert(resultType && "Result type must be a ranked tensor type.");
-    DenseElementsAttr zerosAttr =
-        DenseElementsAttr::get(resultType, /*value=*/0.0f);
-    auto zeroes =
-        rewriter.create<arith::ConstantOp>(op.getLoc(), resultType, zerosAttr);
+
+    Attribute zeroAttr = createDenseElementsAttr(resultType, 0);
+    if (!zeroAttr) {
```
vwellsTT, line 1466: nit: I wonder if both of these `if(!attr)` checks could be simplified to an assert in the new `createDenseElementsAttr`, doesn't seem like we'd expect to encounter any types other than int/floats here; but I'm not sure if we have a strong general philosophy on assert vs pass failure() tbh

lib/Conversion/TTIRToLinalg/TTIRToLinalg.cpp
djstefanovicTT: Updated.

lib/Conversion/TTIRToLinalg/TTIRToLinalg.cpp
djstefanovicTT: It can use arrayRef, so it is updated now.

lib/Conversion/TTIRToLinalg/TTIRToLinalg.cpp
djstefanovicTT: I will leave it as is.

vwellsTT:
Btw, as always it'd be ideal to add coverage for the hoisting workflow for the news ops to test_ttir_ops.py too, but not strictly required.

djstefanovicTT:
@vwellsTT Could you please confirm if this is what you meant? Should a parametrized test also be added?

vwellsTT:
@djstefanovicTT yeah, that's exactly what I meant, thanks! No, I think this is fine, these are simple unary ops so they fit this pattern perfectly.  I have 1 minor new nit though

test/python/golden/test_ttir_ops.py
```diff
@@ -759,6 +759,10 @@ def transpose(
     return builder.transpose(in0, unit_attrs=unit_attrs)
 
 
+def squeeze(in0: Operand, builder: TTIRBuilder, unit_attrs: Optional[List[str]] = None):
```
vwellsTT, line 762: nit: here we should deal with the existing nested `def squeeze` too though--ideally, the hoisted + non-hoisted case can share this func instead of having 2 impls of a squeeze op creator func

vwellsTT: see: line 845

vwellsTT:
Actually, I spoke too soon--yeah I think it would be better to use a specialized test-case for squeeze, since the non-hoisted version is not one of the unary ops--I suspect the params won't fit anyway and it'll cause problems there.
mean is good as-is, since it is.

djstefanovicTT:
There were bugs with implementation of squeeze that got caught with specialized test-case for squeeze. Should be fixed now.

PR #4728, opened 2025-08-29
[optimizer] allow passing external device

### LINKED ISSUE #4232: Optimizer Overrides only run one successfully per python file
### Describe the bug

This is the [bug report](https://github.com/tenstorrent/tt-mlir/actions/runs/16452498991/job/46503307336?pr=4017).

Every time any optimizer overrides run in this file, the first one always works, the second one always fails regardless of the order of the overrides. Something is likely going wrong with the device reset. 

```
cmake -G Ninja -B build -DCMAKE_BUILD_TYPE=Debug -DCMAKE_C_COMPILER=clang-17 -DCMAKE_CXX_COMPILER=clang++-17 -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DTTMLIR_ENABLE_RUNTIME=ON -DTT_RUNTIME_ENABLE_PERF_TRACE=ON -DTTMLIR_ENABLE_RUNTIME_TESTS=ON -DTTMLIR_ENABLE_STABLEHLO=ON -DTT_RUNTIME_DEBUG=ON -DCMAKE_EXPORT_COMPILE_COMMANDS=ON -DTTMLIR_ENABLE_OPMODEL=ON -DTTMLIR_ENABLE_PYKERNEL=ON
cmake --build build
cmake --build build -- ttrt
ttrt query --save-artifacts
export SYSTEM_DESC_PATH=$(pwd)/ttrt-artifacts/system_desc.ttsys
pytest test/python/golden/test_ttir_ops.py

/home/jgrim/wh-01-src/tt-mlir/lib/OpModel/TTNN/SingletonDeviceContext.cpp:33: static SingletonDeviceContext &mlir::tt::op_model::ttnn::SingletonDeviceContext::getInstance(): Assertion instance.m_device != nullptr' failed
```

### Expected behavior

_No response_

_No response_

pilkicTT:
Optimizer needs an open device to work - this causes issues in frontends where we need to close the device before running the `ttir-to-ttnn` pipeline and then re-open the device afterwards. New field is added in the pipeline options, allowing us to pass a device pointer to the optimizer. If the frontend doesn't pass the device pointer, the optimizer will (try to) open its own device.

The added field is not an `Option<T>`, since we do not want to parse it - for example when running the pipeline through the CLI. This means that, for now, we will not be able to use this mechanism in `tt-forge-fe` - due to the RTTI we cannot use the pipeline options struct directly. We'll need to figure out how to workaround this.

The `SingletonDeviceContext` class is refactored to allow for this. Additionally, the behaviour of the singleton is changed so that the lifetime of the device under it is managed fully by the users of the singleton. Even before this change, the situation with the singleton was a bit complex, since we allowed users to close the device manually, but opening of the device was done only during construction of the singleton. This caused issues if you wanted to run optimizer multiple times in a single process (#4232).

Asserts are in place to verify that the singleton is used properly. More info on this in the code comments.

Needed For tenstorrent/tt-xla#952
Closes #4232

### DIFF
--- a/include/ttmlir/OpModel/TTNN/SingletonDeviceContext.h
+++ b/include/ttmlir/OpModel/TTNN/SingletonDeviceContext.h
@@ -19,17 +19,22 @@ namespace mlir::tt::ttnn::op_model {
 
 // Singleton class to provide access to the global device context.
 //
-// Since we do not always own the device (it can be provided by frontends) and
-// there currently isn't a clearly defined scope of the device (within a single
-// function), the lifetime of the device is not managed automatically. The users
-// of this class should ensure that the device lifetime fully contains their
-// scope of usage. For now, if the device is not provided externally, we open
-// our device once we start executing the optimizer and we close it after the
-// `TTNNPrepareConv2dWeightsAndBias` pass - this covers the whole span where we
-// might need the device (for now).
+// The singleton exposes methods to open device, provide already opened device
+// and close the current device.
 //
-// Because the lifetime needs to be managed manually, there are asserts in place
-// to verify that the instance is used correctly.
+// Because we need to support both externally provided device
+// and to open device on our own (if needed), the device is not managed in RAII
+// fashion.
+//
+// Due to the above, it is essential that the users of this class ensure that
+// they have a valid device in the context when they need it.
+//
+// Because the lifetime of the device needs to be managed manually, there are
+// asserts in place to verify that the singleton instance is used correctly.
+//
+// NOTE: If we would close active device in the destructor, we would crash. The
+// reason being that the `tt-metal` singletons (needed for managing devices)
+// would be tore down before ours.
 //
 // TODO (mbezulj): enforce mockup/simulation device when it's enabled in
 // tt-metal.


mtopalovicTT (approved):

include/ttmlir/OpModel/TTNN/SingletonDeviceContext.h
```diff
@@ -7,48 +7,89 @@
...
   static void closeInstance();
 
+  // Sets an externally managed device. The device pointer must remain valid
+  // while the instance is using it. The instance will not attempt to close
+  // the device when `closeInstance()` is called - it will just clear the
+  // pointer.
+  //
+  // This is used when the device is passed to us by the frontend.
+  static void
+  setExternalDevice(::tt::tt_metal::distributed::MeshDevice *device);
```
sgholamiTT: This is just a suggestion:
If `SingletonDeviceContext` is "the" owner of `device`, is it possible to change the signature of this function to:
```c++
static void setExternalDevice(std::unique_ptr<::tt::tt_metal::distributed::MeshDevice> device);
```
Based on this comment of yours:
> _The device pointer must remain valid while the instance is using it._


I get this feeling that this code could potentially introduce some bugs. Eg.

```c++
{
  ::tt::tt_metal::distributed::MeshDevice device = initDevice(...);
  SingletonDeviceContext::setExternalDevice(&device);
} // here the `device` goes out of scope and it will be deleted. 
// So `SingletonDeviceContext` class stores a nullptr.
```

pilkicTT: as we've discussed offline - in the case of `setExternalDevice()` we do not own the device; this is used when a device pointer is passed (borrowed) from the frontends...

arminaleTT (approved):

rpavlovicTT (approved):

PR #4733, opened 2025-08-29
Support lowering constants through D2M

vtangTT:
### Ticket
[Link to Github Issue](https://github.com/tenstorrent/tt-mlir/issues/4746)

### Problem description
Support lowering `ttir.constant` through the D2M flow.

### What's changed
- legalize `ttir.full` in `TTIRtoTTIRGenericPass`
- legalize `memref.global_op` and `memref.get_global_op` in TTIRtoTTKernelPass`
- fixed assert in runtime `executor.cpp`
- re-enabled `Silicon/TTMetal/n150/simple_constant.mlir` test to qualify changes

### Checklist

### DIFF
--- a/lib/Conversion/TTIRToTTIRGeneric/TTIRToTTIRGenericPass.cpp
+++ b/lib/Conversion/TTIRToTTIRGeneric/TTIRToTTIRGenericPass.cpp
@@ -69,6 +69,7 @@ struct TTIRToTTIRGenericPass final
       target.addLegalOp<ttir::ViewLayoutOp>();
       target.addLegalOp<ttir::EmptyOp>();
       target.addLegalOp<ttir::ConstantOp>();
+      target.addLegalOp<ttir::FullOp>();
 
       target.addLegalOp<
 #define GET_OP_LIST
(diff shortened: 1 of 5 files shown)


wenbinlyuTT:
I'm wondering if it's worth it to wire this up to the [`fill_tile`](https://github.com/tenstorrent/tt-metal/blob/main/tt_metal/include/compute_kernel_api/eltwise_unary/fill.h) kernel op, we already have it defined in the TTKernel dialect at https://github.com/tenstorrent/tt-mlir/blob/5122b71de25350ae5c6dc8e7ae0e74a37fc28482/include/ttmlir/Dialect/TTKernel/IR/TTKernelOps.td#L1064

The benefits are:
- We don't need to store the entire tensor in the flatbuffer, reducing its size.
- The constant tensor can be created on the device with the required shape, no need to worry about host side alignment and we save the data transfer cost.
- The creation & fill of the constant tensor can potentially happen after our auto-inserted tilization as an optimization, since it doesn't actually need any tilization anyway. If we optimize hard enough, we can even generate a single CB tile right there in the kernel and reuse that tile for the entire op!

nsmithtt (approved):
Curious to see what you think of @wenbinlyuTT's idea.

vtangTT:
> I'm wondering if it's worth it to wire this up to the [`fill_tile`](https://github.com/tenstorrent/tt-metal/blob/main/tt_metal/include/compute_kernel_api/eltwise_unary/fill.h) kernel op, we already have it defined in the TTKernel dialect at
> 

I think this is a great idea and will open an issue to track. Unsure if we want this to block this PR since it seems much more involved. Maybe this could be a stopgap until we wire up `file_tile`? @nsmithtt @wenbinlyuTT

wenbinlyuTT:
Yeah I was just throwing my ideas out there. I think it should be a follow-up work.

Issue #4201, opened 2025-07-24
Add constraint API for Memory Management Operations


sgholamiTT:
Support the following Ops:
```
TTNN_AllocOp
TTNN_DeallocateOp
TTNN_FillCacheOp
TTNN_UpdateCacheOp
```

Issue #1510, opened 2024-12-04
Dealing with ttnn::update_cache update index


LPanosTT:
We currently (or soon will) have a runtime workaround that reads a tensor off device, and plucks a single value from it in order to populate this argument. The reason the compile models `update_idx` with a runtime tensor is because the graph we get from tt-torch models it that way. Plus, for batch > 1 the forward graph will have more than one value in the `update_idx` tensor. 

We need to figure out a nicer way to do this. I have two ideas:

 - Get the ttnn team to modify this op to accept a device tensor
    - A drawback may be that this could just push the workaround into their stack
 - Since the tensor is initially moved to device, if we can flag graph inputs to either remain on host, or be pushed to device we can avoid placing the tensor on the device just to take it out again. We will still have to retrieve the value from the tensor storage in runtime, but perhaps there is/could be a cleaner way to index host tensors.

skhatriTT:
@LPanosTT Can we close this issue?

LPanosTT:
The issue isn't quite resolved. The workaround is in. Being a workaround it is inherently and undesirable implementation and we need a better way of handling the index for cache updates. This will likely not be resolved for some time however.

sgholamiTT:
Just for the record: [this PR](https://github.com/tenstorrent/tt-mlir/pull/4734/files#diff-63aa5b4c29464f31b7a8e2787813778daf6433e2888d0de717277320d9e2b3b7) is also dealing with the same issue for `update_cache`.

rpavlovicTT:
@odjuricicTT @pilkicTT @jameszianxuTT 

Bringing up this issue to our attention. 
I like the second option suggested by @LPanosTT , as we'd really want to avoid host<->device transactions when not needed during inference.

odjuricicTT:
I like the second option better as well. Putting this in a new list of runtime issues to tackle.

LPanosTT:
Just a note on this one, `ttnn.experimental.paged_update_cache` can perform a static cache update as well, and it can take a cache position for each user in a device tensor. I think going forward we should be using this op instead to perform static cache updates as it has a broader scope of functionality, and we don't need a workaround like this to use it.

rpavlovicTT:
[...]

PR #1437, opened 2024-11-28
Bringup cache updates to ttir level

LPanosTT:
Add `ttnn.update_cache`, `ttir.update_cache`, `ttir.fill_cache`, and `ttnn.fill_cache`

- Frameworks store cache position in a tensor, ttnn needs a `uint32_t` update index (not a tensor). So In runtime I currently read this value from the tensor buffer and call the op. Made a runtime workaround and filed an issue: https://github.com/tenstorrent/tt-mlir/issues/1510
- I noticed that `ttir.constant` ops where the value is 0 is lowered to `ttnn.empty`. I've made a test where the update index tensor holds a single zero. In the runtime however, the tensor returned by the `ttnn.empty` did not have 0 in it, it had a random integer. So I've changed the `ttir.constant` lowering to just use `ttnn.full` no matter the value.
    - Needed to change the file checks for some tests with `ttir.constant`

PR #4734, opened 2025-08-29
[Optimizer] Add constraint API for memory management ops

sgholamiTT:
### Ticket
#4201 

### Problem description
Need constraint APIs for the following ops:
```
TTNN_AllocOp
TTNN_DeallocateOp
TTNN_FillCacheOp
TTNN_UpdateCacheOp
TTNN_WriteTensorOp
```

### What's changed
This PR adds constraint APIs to the mentioned ops. It also promotes the op requirement to `TTNN_InplaceOp`.
The following should be noted:
- `DeallocateOp` has no memory footprint as it simply deallocates an allocated tensor. So I decided to return `OpConstraint{}` in the implementation.
   - Motivation: `query_op_constraints` method in [metal](https://github.com/tenstorrent/tt-metal/blob/79a85942a1c8ddae87ddaa9748ac63f39d900db8/ttnn/api/ttnn/graph/graph_query_op_constraints.hpp#L107C13-L107C86) cannot handle void returning invoke method (like the one for [deallocate](https://github.com/tenstorrent/tt-metal/blob/8b4e4c828ea4689136ace86ab9b37b7d22135b94/ttnn/cpp/ttnn/operations/core/core.hpp#L35C1-L36C1)).
- `AllocOp` cannot be implemented rn as there's no equivalent op in `tt-metal`.
- `UpdateCacheOp`: The argument `UpdateIndex` is stored as a tensor in [mlir](https://github.com/tenstorrent/tt-mlir/blob/8bc924d90b997ba440fd80b243c47fcc2ab48894/include/ttmlir/Dialect/TTNN/IR/TTNNOps.td#L964C1-L965C1), but the `ttnn::update_cache` expects a scalar uint32_t in [metal](https://github.com/tenstorrent/tt-metal/blob/79a85942a1c8ddae87ddaa9748ac63f39d900db8/ttnn/cpp/ttnn/operations/kv_cache/kv_cache.hpp#L22C1-L23C1). So we need to extract the scalar value from the tensor which is not possible in compile time (as opposed to the workaround that is implemented in runtime code in [this PR](https://github.com/tenstorrent/tt-mlir/pull/1437/files#diff-1d3d9d44506c59638c3fa3fd315afdb3f04e395a7a735c4f06513ba9993aa1b6) by Lewis). So we use a default value of 0. [This issue](https://github.com/tenstorrent/tt-mlir/issues/1510) tracks this.
- `WriteTensorOp` cannot be implemented as there's no metal definition for this op.
- The tests for `FillCacheOp` and `UpdateCacheOp` in `TestOpModelInterface.cpp` hang in `getOpRuntime` API call unless the device is reset. This is discussed [here](https://github.com/tenstorrent/tt-metal/issues/25772).

### Checklist


lib/Dialect/TTNN/IR/TTNNOpModelInterface.cpp
```diff
@@ -22,6 +22,78 @@
...
+// supported for a specific reason.
+template <typename T>
+static llvm::Expected<T> issueError(mlir::Operation *op,
+                                    ReasonForLackOfSupport reason,
+                                    APIType apiType) {
+  static_assert(std::is_same_v<T, std::size_t> ||
+                std::is_same_v<T, op_model::OpConstraints>);
+  assert((std::is_same_v<T, std::size_t> && apiType == APIType::OpRunTime) ||
+         (std::is_same_v<T, op_model::OpConstraints> &&
+          apiType == APIType::OpConstraint));
```
azecevic000: Do you expect this to change in the future, currently `APIType` seems redundant.

sgholamiTT: Good call. No I don't think `APIType` is necessary. I removed it.

PR #4735, opened 2025-08-29
[D2M] Fix single bank DRAM sharded host interop

bgrady-tt:
### Problem description
It is convenient to be able to be able to read/write row-major tensors to a single bank in device DRAM. This avoids D2M runtime's lack of support for row-major interleaved tensors while still allowing device side tilization/untilization. 

However the ability to do this was broken recently (it is not clear when), _and likely only worked accidentally for certain cases_. This PR adds more robust support for host<->device transfers using single-bank, single-shard tensors till more complete support is ready. 

### What's changed
- Instead of configuring all DRAM buffers to be 12x1 sharded, DRAM buffers now use only the DRAM banks they actually use in D2M
  - This fixes single-shard host<->device transfers
- The ShardedBufferConfig shapes now has the _same total size as actual buffer_, but use dummy shard and page shapes to support grids that overflow/wrap 12x1 DRAM banks

Note: _**host<->device transfers for multi-shard configs are still broken**. D2M DRAM sharding scheme is not supportable via ShardedBufferConfig due to a lack of support for wrapping grid dims.

### Checklist

### DIFF
--- a/lib/Target/TTMetal/TTMetalToFlatbuffer.cpp
+++ b/lib/Target/TTMetal/TTMetalToFlatbuffer.cpp
@@ -364,9 +364,6 @@ memrefTypeToFlatbuffer(FlatbufferObjectCache &cache, MemRefType memref,
     bool isSharded = mlir::isa<ttcore::ShardLayoutAttr>(memref.getLayout());
 
     if (isSharded) {
-
-      llvm::dbgs() << "SHARDED memref: " << memref << " \n";
-
       flatbuffers::Offset<target::metal::ShardedBufferConfig>
           shardedBufferConfig = memrefTypeToShardedBufferConfigFlatbuffer(
               cache, memref, device, elementShape);
@@ -385,7 +382,6 @@ memrefTypeToFlatbuffer(FlatbufferObjectCache &cache, MemRefType memref,
                          shardedBufferConfig.Union(), circularBufferConfig)
                          .Union();
     } else {
-      llvm::dbgs() << "INTERLEAVED memref: " << memref << " \n";
       // must be interleaved if not sharded
       flatbuffers::Offset<target::metal::InterleavedBufferConfig>
           interleavedBufferConfig =
--- a/lib/Target/TTMetal/TTMetalToFlatbuffer.cpp
+++ b/lib/Target/TTMetal/TTMetalToFlatbuffer.cpp
@@ -169,18 +169,30 @@ createShardedBufferConfigForDRAMMemref(FlatbufferObjectCache &cache,
 
   // NOTE: for DRAM, the coreRangeSet is a core range set from (0,0) to
   // (num_banks=12,1)
-  constexpr int32_t numBanks = 12;
+  // constexpr int32_t numBanks = 12;
   std::vector<target::Dim2dRange> coreRangeSet = {
-      target::Dim2dRange(target::Dim2d(0, 0), target::Dim2d(numBanks, 1))};
+      target::Dim2dRange(target::Dim2d(0, 0), target::Dim2d(1, 1))};
+
+  auto shardLayout = mlir::cast<ttcore::ShardLayoutAttr>(memref.getLayout());
+  auto memrefShardShape = shardLayout.getShardShape(memref);
 
   uint64_t pageSize = device.getMemrefSizeBytes(memref);
-  uint64_t shardSize = pageSize;
-  uint64_t size = pageSize * numBanks;
+  // uint64_t shardSize = pageSize;
+  uint64_t size = pageSize;
 
   // shard shape is (num_elems, 1)
-  target::Dim2d shardShape(shardSize, 1);
-  target::Dim2d pageShape(pageSize, 1);
-  target::Dim2d tensorShapeInPages(numBanks, 1);
+  target::Dim2d shardShape(memrefShardShape[0], memrefShardShape[1]);
+  target::Dim2d pageShape(memrefShardShape[0], memrefShardShape[1]);
+  target::Dim2d tensorShapeInPages(1, 1);
+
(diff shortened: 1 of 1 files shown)


lib/Target/TTMetal/TTMetalToFlatbuffer.cpp
```diff
@@ -167,27 +167,48 @@ createShardedBufferConfigForDRAMMemref(FlatbufferObjectCache &cache,
...
+  auto shardLayout = mlir::cast<ttcore::ShardLayoutAttr>(memref.getLayout());
+  auto memrefGridShape = shardLayout.getGridShape(memref);
+  uint64_t actualShardSize = device.getMemrefSizeBytes(memref);
+  uint64_t gridVolume = ttmlir::utils::volume(memrefGridShape);
 
-  uint64_t pageSize = device.getMemrefSizeBytes(memref);
-  uint64_t shardSize = pageSize;
-  uint64_t size = pageSize * numBanks;
+  // determine how many DRAM banks are actually used by D2M
+  constexpr uint64_t numDRAMBanks = 12;
```
nsmithtt: We need to get this from the system desc.

lib/Target/TTMetal/TTMetalToFlatbuffer.cpp
```diff
@@ -167,27 +167,48 @@ createShardedBufferConfigForDRAMMemref(FlatbufferObjectCache &cache,
...
+  uint64_t numDRAMBanksUsed = std::min(numDRAMBanks, gridVolume);
+  std::vector<target::Dim2dRange> coreRangeSet = {target::Dim2dRange(
+      target::Dim2d(0, 0), target::Dim2d(numDRAMBanksUsed, 1))};
 
-  // shard shape is (num_elems, 1)
-  target::Dim2d shardShape(shardSize, 1);
-  target::Dim2d pageShape(pageSize, 1);
-  target::Dim2d tensorShapeInPages(numBanks, 1);
+  uint64_t actualShardsPerBank = std::ceil(float(gridVolume) / numDRAMBanks);
```
nsmithtt: utils::alignUpDiv

lib/Target/TTMetal/TTMetalToFlatbuffer.cpp
```diff
@@ -167,27 +167,48 @@ createShardedBufferConfigForDRAMMemref(FlatbufferObjectCache &cache,
...
+
+  // compute dummy shapes that occupy same space as actual D2M memref
```
nsmithtt: Titlecase + `.`

lib/Target/TTMetal/TTMetalToFlatbuffer.cpp
```diff
@@ -167,27 +167,48 @@ createShardedBufferConfigForDRAMMemref(FlatbufferObjectCache &cache,
...
+  // will have to be properly fixed using BufferDistributionSpec to describe how
+  // sharded D2M DRAM buffers are actually laid out.
+  //
+  // NOTE: For the special case of a single shard mapped to a single DRAM bank,
+  // the shapes and size here are incidentally correct. So host
+  // enqueue_write_buffer and enqeue_read_buffer commands will work correctly.
+
+  auto shardLayout = mlir::cast<ttcore::ShardLayoutAttr>(memref.getLayout());
+  auto memrefGridShape = shardLayout.getGridShape(memref);
+  uint64_t actualShardSize = device.getMemrefSizeBytes(memref);
```
nsmithtt: You know, I think we might want to just update `device.getMemrefSizeBytes` to return the total memref footprint and not necessarily just the shard shape.  We're going to need this total footprint for allocator as well.  This API was intended to mean total footprint, which happens to be equivalent to the shard for worker grid.

lib/Target/TTMetal/TTMetalToFlatbuffer.cpp
bgrady-tt: fixed 👍

lib/Target/TTMetal/TTMetalToFlatbuffer.cpp
bgrady-tt: I refactored latter half of `DeviceAttr::getMemrefSizeBytes()` into a leaf function `DeviceAttr::getShardSizeInBytes()` that is used to compute shard size taking into account # buffers and desired alignment. The scope of `getShardSizeInBytes()` is more clearly restricted to interpreting the size of a single shard.

`getMemrefSizeBytes()` can change in the future to mean the entire memref footprint (which IMO makes much more sense).

lib/Target/TTMetal/TTMetalToFlatbuffer.cpp
```diff
@@ -165,29 +165,54 @@ static bool isMemrefDeviceL1Memspace(MemRefType memref) {
...
+  auto shardLayout = mlir::cast<ttcore::ShardLayoutAttr>(memref.getLayout());
+  auto memrefGridShape = shardLayout.getGridShape(memref);
+  uint64_t actualShardSize = device.getShardSizeInBytes(memref, 1, true);
+  uint64_t gridVolume = ttmlir::utils::volume(memrefGridShape);
+
+  // determine how many DRAM banks are actually used by D2M
+  uint64_t numDramBanks =
+      systemDesc.getChipDescs().front().getNumDramChannels();
+  uint64_t numDRAMBanksUsed = std::min(numDramBanks, gridVolume);
+  assert(numDramBanks == 12);
```
nsmithtt: Why?

bgrady-tt: Good catch; I added that as a temporary check that post-refactor behavior was consistent. That assert should be removed 😊

PR #4739, opened 2025-08-29
TTRT support for TTNN GenericOp

vtangTT:
### Ticket
[Link to Github Issue](https://github.com/tenstorrent/tt-mlir/issues/4634)

### Problem description
Need runtime support for TTNN GenericOp

### What's changed
- Added `generic_op.cpp` with necessary conversions to create `ProgramDescriptor` and invoke the generic op

Follow up issue for E2E test: https://github.com/tenstorrent/tt-mlir/issues/4805

### DIFF
--- a/runtime/lib/ttnn/operations/generic/generic_op.cpp
+++ b/runtime/lib/ttnn/operations/generic/generic_op.cpp
@@ -53,8 +53,9 @@ createCBDescriptor(const ::tt::target::ttnn::KernelCBDescriptor &cbDesc) {
       .total_size = cbDesc.total_size(),
       .core_ranges =
           tt::runtime::ttnn::utils::toTTNNCoreRangeSet(*cbDesc.core_range()),
-      .format_descriptors = {
-          createCBFormatDescriptor(*cbDesc.formats()->Get(0))}};
+      .format_descriptors = {createCBFormatDescriptor(
+          *cbDesc.formats()->Get(0))},
+      .remote_format_descriptors = {}};
   return cbDescriptor;
 }
 
--- a/runtime/lib/ttnn/operations/generic/generic_op.cpp
+++ b/runtime/lib/ttnn/operations/generic/generic_op.cpp
@@ -7,15 +7,35 @@
 #include "tt/runtime/detail/common/common.h"
 #include "tt/runtime/detail/common/logger.h"
 #include "tt/runtime/detail/ttnn/ttnn.h"
-#include "tt/runtime/runtime.h"
 
-#include "tt/runtime/detail/ttnn/operations/utils.h"
 #include "tt/runtime/detail/ttnn/utils.h"
 #include "ttmlir/Target/TTNN/operations/generic_op_generated.h"
 #include "ttmlir/Target/TTNN/program_generated.h"
 #include "ttnn/types.hpp"
 
 namespace tt::runtime::ttnn::operations::generic_op {
+
+inline CoreType convertCoreType(const ::tt::target::ttnn::CoreType &core_type) {
+  switch (core_type) {
+  case ::tt::target::ttnn::CoreType::WORKER: {
+    return CoreType::WORKER;
+  }
+  case ::tt::target::ttnn::CoreType::ETH: {
+    return CoreType::ETH;
+  }
+    LOG_FATAL("Unsupported core type");
+  }
+}
+
+::tt::tt_metal::SemaphoreDescriptor
+createSemaphoreDescriptor(const ::tt::target::ttnn::SemaphoreDescriptor
+                              &kernel_semaphore_descriptor) {
+  return ::tt::tt_metal::SemaphoreDescriptor{
+      .core_type = convertCoreType(kernel_semaphore_descriptor.core_type()),
+      .core_ranges = tt::runtime::ttnn::utils::toTTNNCoreRangeSet(
+          *kernel_semaphore_descriptor.core_ranges()),
+      .initial_value = kernel_semaphore_descriptor.initial_value()};
+}
 
 ::tt::tt_metal::CBFormatDescriptor createCBFormatDescriptor(
     const ::tt::target::ttnn::KernelCBFormat &kernel_cb_format) {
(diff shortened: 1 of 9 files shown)


phizalev-TT:
Looks good! What is the plan for testing?

vtangTT:
> Looks good! What is the plan for testing?

Looking to follow the common testing plan for these ops found in `test/ttmlir/Silicon/TTNN/n150` using lit, where the ttir op is run through `ttmlir-opt` and `ttmlir-translate`. TTRT then runs the resulting flatbuffer on CI.

In this case, will need to wait for the impl of GenericOp in TTNNtoFlatbuffers pipelines.

runtime/lib/ttnn/operations/generic/generic_op.cpp
```diff
@@ -0,0 +1,275 @@
...
+static_assert(
+    static_cast<uint8_t>(::tt::target::ttnn::SourceType::SOURCE_CODE) ==
+    static_cast<uint8_t>(
+        ::tt::tt_metal::KernelDescriptor::SourceType::SOURCE_CODE));
+inline constexpr ::tt::tt_metal::KernelDescriptor::SourceType
+convertSourceType(const tt::target::ttnn::SourceType &source_type) {
+  return static_cast<::tt::tt_metal::KernelDescriptor::SourceType>(source_type);
+}
+
+::tt::tt_metal::KernelDescriptor::ConfigDescriptor createKernelConfigDescriptor(
```
nsmithtt: `static`

runtime/lib/ttnn/operations/generic/generic_op.cpp
```diff
@@ -0,0 +1,275 @@
...
+    LOG_DEBUG("createing writer kernel config");
+    return ::tt::tt_metal::WriterConfigDescriptor();
+  }
+  default: {
+    LOG_FATAL("Unknown or no kernel config type");
+  }
+  }
+}
+
+std::vector<uint32_t> createKernelArgs(
```
nsmithtt: `static`

runtime/lib/ttnn/operations/generic/generic_op.cpp
```diff
@@ -0,0 +1,275 @@
...
+    }
+    default: {
+      LOG_FATAL("Unknown kernel arg type");
+    }
+    }
+  }
+  return core_args;
+}
+
+::tt::tt_metal::KernelDescriptor
```
nsmithtt: `static`

runtime/lib/ttnn/operations/generic/generic_op.cpp
```diff
@@ -0,0 +1,275 @@
...
+      .common_runtime_args = common_runtime_args,
+      .config = createKernelConfigDescriptor(kernel_desc)};
+
+  return kernel_descriptor;
+}
+
+void run(const ::tt::target::ttnn::GenericOp *op, ProgramContext &context) {
+  ProgramTensorPool &tensorPool = context.getTensorPool();
+  auto size = op->io_tensors()->size();
+  std::vector<::ttnn::Tensor> io_tensors;
```
nsmithtt: minor opt `io_tensors.resize(size)`.

Also camelCase

phizalev-TT (approved):

phizalev-TT:
I suggest we land this before e2e test is possible since it has been tested manually with flatbuffer JSON.

Once landed please start on PR that will contain the test and we will coordinate it with my work on TTNNToFlatbuffers pass.

runtime/lib/ttnn/operations/generic/generic_op.cpp
vtangTT: fixed ty

runtime/lib/ttnn/operations/generic/generic_op.cpp
vtangTT: fixed ty

runtime/lib/ttnn/operations/generic/generic_op.cpp
vtangTT: fixed ty

runtime/lib/ttnn/operations/generic/generic_op.cpp
vtangTT: fixed and changed everything to camel oops

runtime/lib/ttnn/operations/generic/generic_op.cpp
```diff
@@ -0,0 +1,268 @@
...
+#include "tt/runtime/detail/common/common.h"
+#include "tt/runtime/detail/common/logger.h"
+#include "tt/runtime/detail/ttnn/ttnn.h"
+
+#include "tt/runtime/detail/ttnn/utils.h"
+#include "ttmlir/Target/TTNN/operations/generic_op_generated.h"
+
+namespace tt::runtime::ttnn::operations::generic_op {
+
+inline CoreType convertCoreType(const ::tt::target::ttnn::CoreType &coreType) {
```
jnie-TT: I think we can add this to `runtime/include/tt/runtime/detail/ttnn/utils.h`

runtime/lib/ttnn/operations/generic/generic_op.cpp
```diff
@@ -0,0 +1,268 @@
...
+}
+
+static_assert(static_cast<uint8_t>(::tt::target::UnpackToDestMode::Fp32) ==
+              static_cast<uint8_t>(UnpackToDestMode::UnpackToDestFp32));
+static_assert(static_cast<uint8_t>(::tt::target::UnpackToDestMode::Default) ==
+              static_cast<uint8_t>(UnpackToDestMode::Default));
+inline constexpr UnpackToDestMode
+convertUnpackToDestMode(const tt::target::UnpackToDestMode &unpackToDestMode) {
+  return static_cast<UnpackToDestMode>(unpackToDestMode);
+}
```
jnie-TT: Same here and below, I don't think we need the static asserts, we have APIs that map flatbuffer types to ttnn types similar to this in `runtime/include/tt/runtime/detail/ttnn/utils.h` defined in `runtime/lib/ttnn/utils/utils.cpp`

jnie-TT (approved):

runtime/lib/ttnn/operations/generic/generic_op.cpp
vtangTT: I think we should keep them for GenericOp only since these are mapping to metal types, like in `executor_utils.h`. I don't think other ops will or should be using these conversions.

jnie-TT: Hmmmm if it's shared between runtimes then it might be worth adding to `runtime/common` directly then? So that we won't have to update multiple places when say there's a new unpack to dest mode

vtangTT: Ok yea I agree. I'll move it into `runtime/include/tt/runtime/utils.h` wdyt?

jnie-TT: Let's move it to`runtime/include/tt/runtime/detail/common/common.h`, `runtime/include/tt/runtime/utils.h` is a public-facing header so it shouldn't contain any ttnn/ttmetal types

runtime/lib/ttnn/operations/generic/generic_op.cpp
vtangTT: moved ty!

runtime/lib/ttnn/operations/generic/generic_op.cpp
vtangTT: moved ty!

Issue #3875, opened 2025-06-19
implement "incremental" version of greedy allocation planner


vroubtsovTT:
Experiments with "large real life model" input suggest an idea for a (possibly) simple approach to L1/DRAM-capable algorithm:

1. compute a greed-by-size allocation plan assuming everything is in L1 and infinite L1 capacity
2. it is likely that memory pressure will exceed the actual L1 1MB watermark at a small fraction of program points (in llama experiments, those were ~5% of all ops)
3. take a step towards a feasible solution by spilling into DRAM only those ops/operands that are live at the above points -- these could be spilled in priority order based on mem size + number of live interval conflicts;
4. recompute the allocation plan

This could be iterated until a feasible solution is found or could be bisected to find a tight feasible solution. 

How well this will works need experimentation; the nice part about the idea is that we re-use the existing greedy-by-size algorithm and no MIP/CP solver is needed.

vroubtsovTT:
Implemented in #4744

Issue #3508, opened 2025-05-22
extend allocator to handle both L1 and DRAM


vroubtsovTT:
Current allocator implementation ("v2") cannot spill alloc requests to DRAM and will assert if the algorithm runs out of L1 address space. This ticket describes the next implementation step that will add ability to use both memory spaces.

We conceptualize each data stream as being in one of two possible (and exclusive) materializations:

1. tensor is entirely in L1 and the "stream" is effectively looping logic to step through the data shard and feed it into the appropriate ttkernel op
   - only one `memref.alloc` is present in the IR, for a CB for the I/O between the reader and compute kernel threads (as in v2)
2. tensor is entirely in DRAM and is understood not to be able to fit into L1: its data needs to be paged into L1 in chunks of size to be determined by the allocator
   - there are **two** `memref.alloc`s present in the IR, one for the tensor residing in DRAM and another one for the I/O CB

The allocator's decision to "promote" a buffer from DRAM to L1 or, conversely, to "spill" a buffer into DRAM is then a decision of which of the above two setups to choose for each operand data stream. The overall decision for a model should optimize a cost function that reflects data movements cost and manycore parallelism.

Note that unlike the traditional register spill, an L1->DRAM spill for us is not a step that relieves resource pressure monotonically: our spill to DRAM removes an L1 request of size scaling with the tensor size but adds back an L1 request for the I/O buffer (whose size is determined by a generic op's operand type).

Additionally, op results propagate into subsequent ops and thus a decision made for one op will impact data costs/spilling choices for some subsequent ops, i.e. the optimization problem has some sequential decision-making nature.

nsmithtt:
Closing

PR #4744, opened 2025-08-30
D2M: v3 of ttir.allocate is "incremental greedy", can spill from L1 to DRAM

vroubtsovTT:
### Ticket
- #3875
- #3508

### Problem description

This offers v3 of `ttir.allocate` that can spill generic operands to DRAM in order to fit a model's L1 memory usage within a given limit. Briefly, it works as follows:

1. attempt to place everything (that's not explicitly allocated in DRAM) into L1 and allocate/defragment L1 addresses using v2 of the algorithm (`greedy-by-size` by default);
2. if everything was able to fit in the previous step, nothing else needs to be done and the outcome should be identical to what's currently in `main`;
3. otherwise, assign heuristic measures of "spill priority" to various alloc sites that are eligible for spilling; one important contribution to such measures is whether a given alloc makes memory requests that are live during high memory pressure (this is defined as those liveness regions where step 1 required memory in excess of the watermark); these allocs are then  spilled into DRAM incrementally until the result just fits into the L1 limit.

### What's changed

A fuller description is required than what I can fit in here, but:

- `ttir.allocate` pass works with a "planner" API that is abstracted from our IR details; this planner has been moved into `include/ttmlir/Dialect/TTIR/Analysis/Allocation/` and `lib/Dialect/TTIR/Analysis/Allocation/`
- to be able to work with "interesting" and "large" problems, a helper API has been created in `include/ttmlir/Dialect/TTIR/Analysis/Allocation/Tools.h`
- by default, `ttir.allocate` will consider generic op outputs to be ineligible for spilling; this can be changed via pass option `allow-output-spilling=1`

##### Streams for both input operands of a matmul (one out of L1, the other out of DRAM), no stream for the output:
```mlir
    %alloc_42 = memref.alloc() {address = 1024 : i64, alignment = 16 : i64} : memref<8x8x12x12x!ttcore.tile<32x32, f32>, #ttcore.shard<49152x4096>, #l1>
    %alloc_43 = memref.alloc() {address = 1184768 : i64, alignment = 16 : i64} : memref<8x8x1x1x!ttcore.tile<32x32, f32>, #ttcore.shard<4096x4096>, #l1>
    %stream_44 = "ttir.stream_layout"(%alloc_0, %alloc_43) : (memref<8x8x12x12x!ttcore.tile<32x32, f32>, #ttcore.shard<49152x4096>, #dram>, memref<8x8x1x1x!ttcore.tile<32x32, f32>, #ttcore.shard<4096x4096>, #l1>) -> memref<8x8x12x12x!ttcore.tile<32x32, f32>, #ttcore.view<map(4)>, #dram>
    %alloc_45 = memref.alloc() {address = 1180672 : i64, alignment = 16 : i64} : memref<8x8x1x1x!ttcore.tile<32x32, f32>, #ttcore.shard<4096x4096>, #l1>
    %stream_46 = "ttir.stream_layout"(%alloc_37, %alloc_45) : (memref<8x8x12x12x!ttcore.tile<32x32, f32>, #ttcore.shard<49152x4096>, #l1>, memref<8x8x1x1x!ttcore.tile<32x32, f32>, #ttcore.shard<4096x4096>, #l1>) -> memref<8x8x12x12x!ttcore.tile<32x32, f32>, #ttcore.view<map(4)>, #l1>
    ttir.generic {block_factors = [1, 1, 1], grid = #ttcore.grid<8x8>, indexing_maps = [#map1, #map2, #map3], iterator_types = [#parallel, #parallel, #reduction], threads = [#ttir.thread<compute>]}
        ins(%stream_44, %stream_46 : memref<8x8x12x12x!ttcore.tile<32x32, f32>, #ttcore.view<map(4)>, #dram>, memref<8x8x12x12x!ttcore.tile<32x32, f32>, #ttcore.view<map(4)>, #l1>)
        outs(%alloc_42 : memref<8x8x12x12x!ttcore.tile<32x32, f32>, #ttcore.shard<49152x4096>, #l1>)  {
    ^compute0(%cb0: memref<12x12x!ttcore.tile<32x32, f32>, #l1>, %cb1: memref<12x12x!ttcore.tile<32x32, f32>, #l1>, %cb2: memref<12x12x!ttcore.tile<32x32, f32>, #l1>):
      linalg.generic {indexing_maps = [#map1, #map2, #map3], iterator_types = ["parallel", "parallel", "reduction"]} ins(%cb0, %cb1 : memref<12x12x!ttcore.tile<32x32, f32>, #l1>, memref<12x12x!ttcore.tile<32x32, f32>, #l1>) outs(%cb2 : memref<12x12x!ttcore.tile<32x32, f32>, #l1>) {
      ^bb0(%in: !ttcore.tile<32x32, f32>, %in_49: !ttcore.tile<32x32, f32>, %out: !ttcore.tile<32x32, f32>):
        %0 = "ttir.tile_matmul"(%in, %in_49, %out) : (!ttcore.tile<32x32, f32>, !ttcore.tile<32x32, f32>, !ttcore.tile<32x32, f32>) -> !ttcore.tile<32x32, f32>
        linalg.yield %0 : !ttcore.tile<32x32, f32>
      }
    }
    memref.dealloc %alloc_45 : memref<8x8x1x1x!ttcore.tile<32x32, f32>, #ttcore.shard<4096x4096>, #l1>
    memref.dealloc %alloc_43 : memref<8x8x1x1x!ttcore.tile<32x32, f32>, #ttcore.shard<4096x4096>, #l1>
    ...
```
##### No stream for the input of non-matmul generic op, stream for the output:
```mlir
    %alloc = memref.alloc() {address = 1024 : i64, alignment = 16 : i64} : memref<8x8x12x12x!ttcore.tile<32x32, f32>, #ttcore.shard<49152x4096>, #l1>
    %alloc_0 = memref.alloc() {address = 2360320 : i64, alignment = 32 : i64} : memref<8x8x12x12x!ttcore.tile<32x32, f32>, #ttcore.shard<49152x4096>, #dram>
    %alloc_1 = memref.alloc() {address = 590848 : i64, alignment = 16 : i64} : memref<8x8x1x1x!ttcore.tile<32x32, f32>, #ttcore.shard<4096x4096>, #l1>
    %stream = "ttir.stream_layout"(%alloc_0, %alloc_1) : (memref<8x8x12x12x!ttcore.tile<32x32, f32>, #ttcore.shard<49152x4096>, #dram>, memref<8x8x1x1x!ttcore.tile<32x32, f32>, #ttcore.shard<4096x4096>, #l1>) -> memref<8x8x12x12x!ttcore.tile<32x32, f32>, #ttcore.view<map(4)>, #dram>
    ttir.generic {block_factors = [1, 1], grid = #ttcore.grid<8x8>, indexing_maps = [#map, #map], iterator_types = [#parallel, #parallel], threads = [#ttir.thread<compute>]}
        ins(%alloc : memref<8x8x12x12x!ttcore.tile<32x32, f32>, #ttcore.shard<49152x4096>, #l1>)
        outs(%stream : memref<8x8x12x12x!ttcore.tile<32x32, f32>, #ttcore.view<map(4)>, #dram>)  {
    ^compute0(%cb0: memref<12x12x!ttcore.tile<32x32, f32>, #l1>, %cb1: memref<12x12x!ttcore.tile<32x32, f32>, #l1>):
      linalg.generic {indexing_maps = [#map, #map], iterator_types = ["parallel", "parallel"]} ins(%cb0 : memref<12x12x!ttcore.tile<32x32, f32>, #l1>) outs(%cb1 : memref<12x12x!ttcore.tile<32x32, f32>, #l1>) {
      ^bb0(%in: !ttcore.tile<32x32, f32>, %out: !ttcore.tile<32x32, f32>):
        %0 = "ttir.tile_cos"(%in) : (!ttcore.tile<32x32, f32>) -> !ttcore.tile<32x32, f32>
        linalg.yield %0 : !ttcore.tile<32x32, f32>
      }
    }
    memref.dealloc %alloc_1 : memref<8x8x1x1x!ttcore.tile<32x32, f32>, #ttcore.shard<4096x4096>, #l1>
    memref.dealloc %alloc : memref<8x8x12x12x!ttcore.tile<32x32, f32>, #ttcore.shard<49152x4096>, #l1>
```

### TODOs

- we've discussed a possible API that the allocator can use to query generic ops for their buffer sizing options; this PR does not attempt to implement it and instead mocks all buffers to have `1x1...` shapes, i.e. smallest possible size.
- there is some scope for tweaking heuristic ranking of spill candidates, currently 
```cpp
  auto byPriority = make_lexicographical_field_comparator<std::greater<>>(
    &SpillPriority::inContentionGroup,   // Contributor to a contention group.
    &SpillPriority::worstSize,           // Largest mem size of a request.
    &SpillPriority::worstMaxConflictSize // Largest conflict set size contributed to.
  );
 ``` 
 but this ought to be experimented on with real model inputs. Only the first term above is "obvious".

### Checklist
- [x] New IR-oblivious unit tests in `test/unittests/Allocation/TestAllocation.cpp` 
- [x] Existing lit allocator tests pass
- [x] New lit allocator tests check spilling to DRAM: `test/ttmlir/Dialect/TTIR/allocate/allocate.spill{_negative}.mlir`

### DIFF
--- a/lib/Dialect/TTIR/Analysis/Allocation/Planner.cpp
+++ b/lib/Dialect/TTIR/Analysis/Allocation/Planner.cpp
@@ -84,8 +84,8 @@ Planner::AllocateStats PlannerImpl::allocateImpl<Planner::Algorithm::Simple>(
             (requests[lhs].last > requests[rhs].last));
   };
 
-  std::priority_queue<IndexT, std::vector<IndexT>, decltype(lexicographic)> pq(
-      lexicographic);
+  std::priority_queue<IndexT, std::vector<IndexT>, decltype(lexicographic)>
+      requestIndices(lexicographic);
   [[maybe_unused]] int32_t varCount = 0;
 
   for (const Planner::Variable &var : problem.variables) {
@@ -93,20 +93,20 @@ Planner::AllocateStats PlannerImpl::allocateImpl<Planner::Algorithm::Simple>(
       continue;
     }
     for (const auto reqIndex : var.domain[ordinal(var.placement)]) {
-      pq.push(reqIndex);
+      requestIndices.push(reqIndex);
     }
     ++varCount;
   }
 
   TT_ALLOC_DEBUG("allocating {} var(s), {} requests(s) ...", varCount,
-                 pq.size());
+                 requestIndices.size());
 
   AllocSizeT memUsage = 0;
   AllocSizeT maxSize = 0;
 
-  while (!pq.empty()) {
-    const IndexT reqIndex = pq.top();
-    pq.pop();
+  while (!requestIndices.empty()) {
+    const IndexT reqIndex = requestIndices.top();
+    requestIndices.pop();
     Request &request = requests[reqIndex];
     TT_ALLOC_TRACE("request #{}: {}", reqIndex, request);
 
@@ -141,7 +141,8 @@ Planner::AllocateStats PlannerImpl::allocateImpl<Planner::Algorithm::Greedy>(
     return (requests[lhs].size < requests[rhs].size);
   };
 
-  std::priority_queue<IndexT, std::vector<IndexT>, decltype(bySize)> pq(bySize);
+  std::priority_queue<IndexT, std::vector<IndexT>, decltype(bySize)>
+      requestIndices(bySize);
   [[maybe_unused]] int32_t varCount = 0;
 
   for (const Planner::Variable &var : problem.variables) {
@@ -149,13 +150,13 @@ Planner::AllocateStats PlannerImpl::allocateImpl<Planner::Algorithm::Greedy>(
       continue;
     }
     for (const auto reqIndex : var.domain[ordinal(var.placement)]) {
-      pq.push(reqIndex);
+      requestIndices.push(reqIndex);
     }
     ++varCount;
   }
 
   TT_ALLOC_DEBUG("allocating {} var(s), {} requests(s) ...", varCount,
-                 pq.size());
+                 requestIndices.size());
 
   // An index of already visited requests in increasing offset order.
   // TODO(vroubtsov) using std::multimap allows some undesired (but not
@@ -166,9 +167,9 @@ Planner::AllocateStats PlannerImpl::allocateImpl<Planner::Algorithm::Greedy>(
   AllocSizeT memUsage = 0;
   AllocSizeT maxSize = 0;
 
-  while (!pq.empty()) {
-    const IndexT reqIndex = pq.top();
-    pq.pop();
+  while (!requestIndices.empty()) {
+    const IndexT reqIndex = requestIndices.top();
+    requestIndices.pop();
     Request &request = requests[reqIndex];
     TT_ALLOC_TRACE("request #{}: {}", reqIndex, request);
     TT_assertv(request.offset < 0,
@@ -232,7 +233,7 @@ Planner::AnalysisStats PlannerImpl::analyze(const Problem &solution,
   using IntervalTree = llvm::IntervalTree<SequenceT, IndexT>;
 
   IntervalTree::Allocator allocator;
-  IntervalTree iv(allocator);
+  IntervalTree intervals(allocator);
 
   for (const Variable &var : solution.variables) {
     if (var.placement == Space::NA) {
@@ -242,11 +243,11 @@ Planner::AnalysisStats PlannerImpl::analyze(const Problem &solution,
       const Request &req = solution.request(reqIndex);
 
       requests.emplace_back(reqIndex);
-      iv.insert(req.first, req.last, reqIndex);
+      intervals.insert(req.first, req.last, reqIndex);
     }
   }
   // NOLINTNEXTLINE
-  iv.create();
+  intervals.create();
 
   analysis.requestMetrics.resize(solution.requests.size());
 
@@ -257,7 +258,7 @@ Planner::AnalysisStats PlannerImpl::analyze(const Problem &solution,
     SequenceT position = req.first;
     AllocSizeT usage = 0;
 
-    const auto conflicts = iv.getContaining(position);
+    const auto conflicts = intervals.getContaining(position);
     for (const auto &entry : conflicts) {
       const auto conflictIndex = entry->value();
       const Request &conflict = solution.request(conflictIndex);
@@ -449,10 +450,10 @@ Planner::AllocateStats Planner::verify(const Problem &solution) {
   using IntervalTree = llvm::IntervalTree<SequenceT, IndexT>;
 
   IntervalTree::Allocator allocator;
-  IntervalTree iv(allocator);
+  IntervalTree intervals(allocator);
 
   // All requests referenced by enabled 'solution.variables'.
-  std::vector<Planner::IndexT> requests;
+  std::vector<Planner::IndexT> requestIndices;
   [[maybe_unused]] int32_t varCount = 0;
 
   for (const Variable &var : solution.variables) {
@@ -462,15 +463,15 @@ Planner::AllocateStats Planner::verify(const Problem &solution) {
     for (const auto reqIndex : var.domain[ordinal(var.placement)]) {
       const Request &req = solution.request(reqIndex);
 
-      requests.emplace_back(reqIndex);
-      iv.insert(req.first, req.last, reqIndex);
+      requestIndices.emplace_back(reqIndex);
+      intervals.insert(req.first, req.last, reqIndex);
     }
     ++varCount;
   }
   // NOLINTNEXTLINE
-  iv.create();
+  intervals.create();
 
-  const int32_t reqCount = requests.size();
+  const int32_t reqCount = requestIndices.size();
   TT_ALLOC_TRACE("verifying {} var(s), {} request(s) ...", varCount, reqCount);
   if (!reqCount) {
     return {0, 0, 0};
@@ -480,13 +481,13 @@ Planner::AllocateStats Planner::verify(const Problem &solution) {
   AllocSizeT maxSize = 0;
   AllocSizeT maxLoad = 0;
 
-  for (const auto reqIndex : requests) {
+  for (const auto reqIndex : requestIndices) {
     const Request &req = solution.request(reqIndex);
     TT_assertv(req.offset >= 0, "unexpected unallocated request {}", req);
 
     AllocSizeT load = req.size;
 
-    const auto conflicts = iv.getContaining(req.first);
+    const auto conflicts = intervals.getContaining(req.first);
     for (const auto &entry : conflicts) {
       if (entry->value() != reqIndex) {
         const Request &conflict = solution.request(entry->value());

### CODE REVIEW

File: include/ttmlir/Dialect/TTIR/Analysis/Allocation/Planner.h [hunk 1]
```diff
@@ -0,0 +1,101 @@
...
+///   that sense, `spillAllocate()` could be seen as "decide which variables to
+///   spill, followed by allocating remaining unspilled variables; do both
+///   'optimally'".
+///
+/// The caller can mark some problem variables as "bound", in which case
+/// the planner will consider them pinned to a particular space but will still
+/// include their memory offsets in the allocatin/defragmentation part of its
+/// task.
+///
+class Planner : public PlannerImpl {
```
vwellsTT, line 37: I suppose the idea is eventually you'll have other planner variants?  Otherwise I don't see why this separate impl is useful

lib/Dialect/TTIR/Analysis/Allocation/Planner.cpp
```diff
@@ -0,0 +1,528 @@
...
+  return true;
+}
+
+llvm::raw_ostream &operator<<(llvm::raw_ostream &os,
+                              const Planner::Problem &obj) {
+  std::stringstream s;
+  Tools::write(obj, s);
+  return os << s.str();
+}
+// ............................................................................
```
vwellsTT: Ugly 😆

lib/Dialect/TTIR/Analysis/Allocation/Planner.cpp
```diff
@@ -0,0 +1,528 @@
...
+
+  auto &requests = problem.requests;
+
+  auto lexicographic = [&](IndexT lhs, IndexT rhs) {
+    return (requests[lhs].first > requests[rhs].first) ||
+           ((requests[lhs].first == requests[rhs].first) &&
+            (requests[lhs].last > requests[rhs].last));
+  };
+
+  std::priority_queue<IndexT, std::vector<IndexT>, decltype(lexicographic)> pq(
```
vwellsTT: I feel like this should be named something more descriptive, I wouldn't name an `std::vector` `v`, but something related to what it's a vector (or priority_queue) of

lib/Dialect/TTIR/Analysis/Allocation/Tools.cpp
```diff
@@ -0,0 +1,451 @@
...
+  TT_assertv(problem.valid(), "expected a valid problem");
+
+  std::ofstream out(file);
+  TT_assertv(out.is_open(), "couldn't open output file [{}]", file);
+
+  return write(problem, out);
+}
+
+} // namespace mlir::tt::ttir::allocation
+// ----------------------------------------------------------------------------
```
vwellsTT: 👎 to line break at EOF

test/python/golden/test_metal_dma.py
```diff
@@ -102,15 +102,13 @@ def tilize(
         unit_attrs: List[str] = None,
     ):
         # derive sharded shapes
-        assert (
-            (shape[0] % start_grid[0] == 0) and (shape[1] % start_grid[1] == 0),
-            "shape must be divisible by start_grid",
-        )
+        assert (shape[0] % start_grid[0] == 0) and (
+            shape[1] % start_grid[1] == 0
+        ), "shape must be divisible by start_grid"
```
vwellsTT, line 107: personally kind of prefer old syntax here

lib/Dialect/TTIR/Analysis/Allocation/Tools.cpp
```diff
@@ -0,0 +1,451 @@
...
+#include <memory>
+#include <random>
+#include <sstream>
+
+// ----------------------------------------------------------------------------
+namespace mlir::tt::ttir::allocation {
+
+namespace {
+
+using std::int32_t;
```
vwellsTT, line 26: why not just omit `std::` without this? Like a normal person 😛

include/ttmlir/Dialect/TTIR/Analysis/Allocation/Planner.h
vroubtsovTT: The idea was that I wanted to separate a relatively compact public API of `Planner` from the many types and enums  nested inside of it. The nesting is necessary in c++17 so I could have useful overloads for operator++, operator<<, etc without making them global[*]. In c++20 some of these scoping problems could be done more elegantly with namespaces and `using` but c++17 enums are more limited. So, I chose to use inheritance which sort of works for type injection into a class scope and allows me to keep nested types out of the way, in another header.

[*] I normally solve such problems by using "smart typesafe enums" implemented via macros. metal uses [magic enum](https://github.com/Neargye/magic_enum) for the same. But we currently don't have either option.

vwellsTT: I get that, but what is the use of the public API? Are there other users planned, or what?  Seems overkill if not.

test/python/golden/test_metal_dma.py
vroubtsovTT: the old syntax was actually broken because `assert (condition, "message")` is testing a python tuple and will never trigger regardless of `condition`. (`assert` is a keyword in Python, not a function call).

The ensuring formatting is from the black formatter which I inferred as our python style standard.

vwellsTT: Oh, 🤯

vwellsTT: yet another reason python is a terrible language 😡

lib/Dialect/TTIR/Analysis/Allocation/Tools.cpp
vroubtsovTT: Because c++ doesn't have `int32_t` as a built-in, it needs to be obtained from `<cstdint>` or elsewhere. And since I would like to use such primitive types without `std::` everywhere, I do so by importing them via `using std::...` in the namespaces I control (and never do so in public headers, etc). I normally do this once per project, by injecting all such types into the project's top-level namespace in a common project header but we don't have such project setup so I only do this in .cpp files as needed. In header files I continue to use the `std::` prefix.

llvm does something like this also, except it's not clear exactly where they get these standard int types from. Like I mentioned in another comment, I have a piece (Planner) that's quasi-llvm-independent.

test/python/golden/test_metal_dma.py
vroubtsovTT: Python is nice, it just has a grammar parser that's not derived from LL/LALR/etc and is rather hand-coded because it wants to be a _nice_ language to use. People feel _good_ using Python... 😄

include/ttmlir/Dialect/TTIR/Analysis/Allocation/Planner.h
vroubtsovTT: I am trying to keep the impl so that there is a piece that is "MLIR-dependent" (Allocate.cpp and friends) and "MLIR-independent" (Tools, Planner, etc). The latter I want because that's how I can create synthetic problems to try different algorithms on, an OR solver, etc, since we don't have functional completeness yet to drive a full mode like LLama all the way into the mid-pipeline. You'll notice that Planner only has llvm ADTs included, nothing related to compiler passes, etc.

lib/Dialect/TTIR/Analysis/Allocation/Tools.cpp
vwellsTT: sure, but in practice `cstdint` adds to global namespace almost always

lib/Dialect/TTIR/Analysis/Allocation/Planner.cpp
vroubtsovTT: These beautiful dots were grandfathered from the previous version that you were ok with 😄.  I was hoping for one more time.

lib/Dialect/TTIR/Analysis/Allocation/Tools.cpp
vroubtsovTT: don't understand this... clang-tidy enforces line feed at EOF, right?

vwellsTT: I don't like horizontal bar at end of line, there's nothing to break up here

include/ttmlir/Dialect/TTIR/Analysis/Allocation/Planner.h
vwellsTT: Seems a bit overcomplicated to me, but I guess if it's needed for testing or whatnot makes sense

lib/Dialect/TTIR/Analysis/Allocation/Tools.cpp
vwellsTT: are you a `#pragma once` denier too? =P

vroubtsovTT: is that part of the language standard? my habit comes from the times where `<cstdint>` was really a boost lib...

vwellsTT: nope, but it's used fairly ubiquitously, somewhat more common than real include guards I think these days-- haha!

lib/Dialect/TTIR/Analysis/Allocation/Planner.cpp
vroubtsovTT: Replaced with llvm's line thingies.

lib/Dialect/TTIR/Analysis/Allocation/Tools.cpp
vroubtsovTT: not good enough for me 😦

lib/Dialect/TTIR/Analysis/Allocation/Tools.cpp
vroubtsovTT: Ok. I am used to two different styles of "lines", one is always there to remind me to have a line feed, etc. I've changed it to be the approved llvm line.

This all comes from ancient genetic memories of coding c++ when there was no syntax coloring and all you had to help your bleeding eyes was a bit of ascii art...

lib/Dialect/TTIR/Analysis/Allocation/Tools.cpp
vroubtsovTT: > are you a `#pragma once` denier too? =P

Missed that comment at first but actually I wanted to use `#pragma once`. And I can prove it: https://github.com/vladium/static_hash/blob/af1ab80ccb2f9e4c595d792c591cf5f527ee1404/src/vr/hashing.h#L1

include/ttmlir/Dialect/TTIR/Analysis/Allocation/Planner.h
vroubtsovTT: A reasonable thought. But the real challenge here is the NP-hard problem of defragmenting or spilling-and-defragmenting in large models. One considered direction was to use an off-the-shelf optimizer engine that would take in [an abstracted statement of the problem](https://arxiv.org/abs/2311.18246), with interval variables, their constraints, etc. Hence the desire to develop a mathematical problem definition that's completely independent of IR -- that's `Planner::Problem` and its previous version, `AllocationPlanner::Context`.  It is somewhat [similar to what you'd see in the minimalloc implementation.](https://github.com/google/minimalloc/blob/18d8590d548daf2574b57265b71307e92a7b559d/src/minimalloc.h#L87).

This may yet prove to be an overkill, especially if we end up never needing to do static allocation over full/large models. Or if we never find ourselves running close to L1 capacity limit, e.g. because we can always do "soft" spilling via stream of varying buffer sizes.

lib/Dialect/TTIR/Analysis/Allocation/Planner.cpp
vroubtsovTT: Renamed several variables `iv -> intervals, pq -> requestIndices` in the very last commit.

PR #4753, opened 2025-09-02
[D2M] TransposeOp

vwellsTT:
### Ticket
https://github.com/tenstorrent/tt-mlir/issues/4755 (Step 1).
### Problem description
D2M need to support lowering transpose ops.

### What's changed
Added support for lowering transpose into 2 components, a reblocking StreamLayoutOp, and a GenericOp which tranposes each tile.  Added TTKernel support for tranpose_wh_tile op for this, a lit test, and a builder golden test.

### Checklist


lib/Conversion/TTIRToTTKernel/TTIRToTTKernel.cpp
```diff
@@ -212,6 +213,8 @@ class MemrefStoreRewriter : public OpConversionPattern<memref::StoreOp> {
   LogicalResult
   matchAndRewrite(memref::StoreOp op, memref::StoreOpAdaptor adaptor,
                   ConversionPatternRewriter &rewriter) const final {
+
+    llvm::errs() << "MemrefStoreRewriter::matchAndRewrite\n";
```
nsmithtt: stray print

lib/Conversion/TTIRToTTIRGeneric/TTIRToTTIRGeneric.cpp
```diff
@@ -720,10 +726,150 @@ class TTIRMatmulRewriter final
...
+      : OpConversionPattern<ConcreteOp>(typeConverter, ctx),
+        TTIRNamedRewriterCommon(defaultInputMemSpace, defaultOutputMemSpace,
+                                targetGridShape) {}
+
+  LogicalResult
+  matchAndRewrite(ttir::PermuteOp op, typename ConcreteOp::Adaptor adaptor,
+                  mlir::ConversionPatternRewriter &rewriter) const final {
+    auto permutation = op.getPermutation();
+    assert(permutation.size() == 2 && permutation[0] == 1 &&
+           permutation[1] == 0 && "Only 2D transpose supported");
```
nsmithtt, line 778: Instead of asserting can we just return failure?  Then it'll gracefully fail with a conversion error and it'll be obvious where someone should update to handle the other permute cases.  On that note, maybe we can even set this up further by moving the body of this function into a separate function, with the eventual goal of having it look like:

```
if (permutation.size() > 2 &&  any_of(permutation, [](int elem) { return elem == -1; })) {
  return splitPermute(...); // Break the permute into 2 permutes, 1 that pulls out a permute on just the inner dims and a separate one that handles the rest of the outer dims.
} else if (permutation.size() == 2 && permutation[0] == 1 && permutation[1] == 0) {
  return permuteInnerDims(...);  // Handles the special case you have in this PR
} else {
  return permuteOuterDims(...); // Generally handles all other dims with just affine map in view_layout op
}
```

We can just stub out these extra unsupported cases with `return failure();` for now.

lib/Conversion/TTIRToTTIRGeneric/TTIRToTTIRGeneric.cpp
```diff
@@ -720,10 +726,150 @@ class TTIRMatmulRewriter final
...
+    for (unsigned i = 0; i < deviceRank; i += 2) {
+      if (i + 1 < deviceRank) {
+        // Swap each pair.
+        results.push_back(rewriter.getAffineDimExpr(i + 1));
+        results.push_back(rewriter.getAffineDimExpr(i));
+      } else {
+        // Odd dimension count - keep last dim as-is.
+        results.push_back(rewriter.getAffineDimExpr(i));
+      }
+    }
```
nsmithtt: Maybe this could be written more succinctly:
```
results.resize(deviceRank);
assert(permutation.size() == deviceRank);
for (auto [i, permuteDim] : llvm::enumerate(permutation)) {
  results[i] = rewriter.getAffineDimExpr(permuteDim);
  results[deviceRank + i] = rewriter.getAffineDimExpr(deviceRank + permuteDim);
}
```

lib/Conversion/TTIRToTTIRGeneric/TTIRToTTIRGeneric.cpp
```diff
@@ -720,10 +726,150 @@ class TTIRMatmulRewriter final
...
+        inputLayout.getCollapsedIntervals(), inputLayout.getOobVal(),
+        inputLayout.getMemorySpace(), composedMap);
+
+    SmallVector<int64_t> resultDeviceShape(inputShape.begin(),
+                                           inputShape.end());
+    // For now, we've asserted we're 2D, so we should have 4D on device (grid +
+    // shard).
+    assert(resultDeviceShape.size() == 4);
+    std::swap(resultDeviceShape[0], resultDeviceShape[1]);
+    std::swap(resultDeviceShape[2], resultDeviceShape[3]);
```
nsmithtt: Same kind of for loop from comment above could be applied here which would support any rank.

lib/Conversion/TTIRToTTIRGeneric/TTIRToTTIRGeneric.cpp
```diff
@@ -720,10 +726,150 @@ class TTIRMatmulRewriter final
...
+
+    // Identity maps since grid shapes now match.
+    auto identityMap = rewriter.getMultiDimIdentityMap(2);
+    SmallVector<mlir::Attribute> iteratorTypes(
+        2, ttcore::IteratorTypeAttr::get(ctx, ttcore::IteratorType::Parallel));
+
+    auto generic = rewriter.create<ttir::GenericOp>(
+        loc, inputs, outputs,
+        rewriter.getAffineMapArrayAttr({identityMap, identityMap}),
+        rewriter.getArrayAttr(iteratorTypes));
```
nsmithtt: There's a different constructor you can use where you don't need to specify the identity maps / iterators, it'll infer them as identity and parallel:

```c++
    auto generic = rewriter.create<ttir::GenericOp>(
        loc, inputs, outputs, [] (OpBuilder& builder, Location loc, ValueRange blockArgs) {
          auto linalgGeneric = rewriter.create<mlir::linalg::GenericOp>(
          loc,
          llvm::to_vector(mlir::ValueRange(blockArgs.take_back(1)).getTypes()),
          blockArgs.take_front(1), blockArgs.take_back(1), linalgIndexingMaps,
          linalgIteratorTypes,
          [&](mlir::OpBuilder &bbBuilder, mlir::Location bbLoc,
              mlir::ValueRange bbArgs) {
            mlir::Value yield = bbBuilder.create<ttir::TileTransposeOp>(
                bbLoc, bbArgs.take_back(1).getTypes(), bbArgs.take_front(1));
            bbBuilder.create<mlir::linalg::YieldOp>(bbLoc, yield);
          });
  });
```

This could replace lines 822 - 861.

lib/Conversion/TTIRToTTKernel/TTIRToTTKernel.cpp
```diff
@@ -505,6 +508,45 @@ class TTIRTilizeUntilizeRewriter
...
+class TTIRTileTransposeRewriter
+    : public OpConversionPattern<ttir::TileTransposeOp> {
+public:
+  using OpConversionPattern<ttir::TileTransposeOp>::OpConversionPattern;
+
+  LogicalResult
+  matchAndRewrite(ttir::TileTransposeOp op,
+                  ttir::TileTransposeOpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const final {
+    llvm::errs() << "TTIRTileTransposeRewriter::matchAndRewrite\n";
```
nsmithtt: Stray print

test/python/golden/test_ttir_ops.py
```diff
@@ -1601,23 +1601,19 @@ def softmax(
     in0: Operand,
     builder: TTIRBuilder,
     dimension: int = -1,
-    numeric_stable: bool = False,
     unit_attrs: Optional[List[str]] = None,
 ):
-    return builder.softmax(
-        in0, dimension=dimension, numeric_stable=numeric_stable, unit_attrs=unit_attrs
-    )
+    return builder.softmax(in0, dimension=dimension, unit_attrs=unit_attrs)
```
nsmithtt: Is this intentional? Seems unrelated.

test/python/golden/test_ttir_ops.py
```diff
@@ -1601,23 +1601,19 @@ def softmax(
...
 @pytest.mark.parametrize("dimension", [-1])
-@pytest.mark.parametrize("numeric_stable", [False, True])
-def test_softmax(shape: Shape, dimension: int, numeric_stable: bool, request):
+def test_softmax(shape: Shape, dimension: int, request):
     # Create a wrapper function that captures dimension
     def softmax_wrapper(
         in0: Operand, builder: TTIRBuilder, unit_attrs: Optional[List[str]] = None
     ):
-        return softmax(in0, builder, dimension, numeric_stable, unit_attrs)
+        return softmax(in0, builder, dimension, unit_attrs)
```
nsmithtt: Is this intentional? Seems unrelated.

vwellsTT: nope again

test/python/golden/test_ttir_ops.py
vwellsTT: nope, was trying to get cursor to fix things last night and it made some spurious changes I didn't catch, reverting...

lib/Conversion/TTIRToTTIRGeneric/TTIRToTTIRGeneric.cpp
vwellsTT: Ah, nice--much less ugly.

lib/Conversion/TTIRToTTIRGeneric/TTIRToTTIRGeneric.cpp
vwellsTT: makes sense

lib/Conversion/TTIRToTTIRGeneric/TTIRToTTIRGeneric.cpp
vwellsTT: This doesn't seem quite equivalent to me, but then I'm not sure current logic is sensible, it seems kind of weird when I'm looking at it now 🤔

lib/Conversion/TTIRToTTIRGeneric/TTIRToTTIRGeneric.cpp
vwellsTT: yeah, I guess this is definitely a bit over-specialized currently.  I think the way I see this:
1. we should apply the permuted mapping to grid, then to shard.
2. if we've already done collapsing on our inputs though, we're in trouble--we might need to avoid collapsing on higher dimensional permute ops?  I think for now asserting 2D + applying mapping makes sense, and as we flesh out bcast cases, permute can be another case where we don't permit collapsing

lib/Conversion/TTIRToTTIRGeneric/TTIRToTTIRGeneric.cpp
nsmithtt: I suppose this assert can still fail if permutation.size() > 2

nsmithtt (approved):

lib/Conversion/TTIRToTTIRGeneric/TTIRToTTIRGeneric.cpp
vwellsTT: Yeah--I think until we support 3D+ tensors w/o collapse, this is desirable, since it's not correct to try to do a transpose post collapse.  (e.g. if input is 3D, permute is [0,2,1], but we've collapse to 2D, I think we're screwed) -- all of this assuming `permutations.size() == logicalShape.size()`.

lib/Dialect/TTKernel/IR/TTKernelOps.cpp
```diff
@@ -155,6 +155,45 @@ ::mlir::LogicalResult ExperimentalUntilizeBlockOp::verify() {
   return success();
 }
 
+::mlir::LogicalResult TransposeInitOp::verify() {
+  if (!insideEnqueueProgramOpRegion(getOperation())) {
+    return emitOpError(
+        "TransposeInitOp must be inside of a EnqueueProgramOp region");
+  }
+
+  // Both input and output should have tile element types for transpose
```
anuragsingh-tt: nit: period after comments

lib/Conversion/TTIRToTTIRGeneric/TTIRToTTIRGeneric.cpp
vwellsTT: I guess is your point you'd prefer this to be a pass failure() instead of an assert? I don't have strong feeling for which is more suitable for this case, happy to change if you think that's better

lib/Dialect/TTIR/IR/TTIROps.cpp
```diff
@@ -2718,8 +2718,11 @@ createDefaultLayout(mlir::MLIRContext *ctx,
   // Create default layout for tensor without encoding
   llvm::SmallVector<int64_t> logicalShape(tensorType.getShape());
 
+  SmallVector<int64_t> squareGridShape =
```
vwellsTT, line 2722: Remark: it's a little ugly we have to update the worker grid like this here and in LowerToLayout.cpp--I think there should be a nicer mechanism to access the same grid consistently, but I'd rather do that in a follow-up PR and just call util anywhere we create MetalLayoutAttr for now--open to pushback though.

lib/Conversion/TTIRToTTIRGeneric/TTIRToTTIRGeneric.cpp
nsmithtt: I'm not super tied either way, was thinking that pass failure might be a little more graceful, but not going to block the PR on it.

lib/Dialect/TTIR/IR/TTIROps.cpp
nsmithtt: I think it might be cleaner to just do this clamp at the TTIRToTTIRGenericPass.cpp level and override the `workerGridShape` that get's passed around there.

nsmithtt: imo this constructor shouldn't be in the business of changing its inputs, it should blindly construct the layout it was asked to construct.  The square grid policy should be one on the pass level.

lib/Dialect/TTIR/Utils/Utils.cpp
```diff
@@ -96,4 +96,14 @@ mlir::LogicalResult broadcastValue(mlir::PatternRewriter &rewriter,
                                           broadcastDims);
   return mlir::success();
 }
+
+llvm::SmallVector<int64_t, 2>
+getSquareTargetGrid(mlir::ArrayRef<int64_t> targetGridShape) {
+  const int64_t minGridValue =
+      *(std::min_element(targetGridShape.begin(), targetGridShape.end()));
```
azecevic000: Nit: `llvm::min_element(targetGridShape)`

lib/Conversion/TTIRToTTIRGeneric/TTIRToTTIRGeneric.cpp
```diff
@@ -208,21 +210,29 @@ class TTIRNamedRewriterCommon {
     return defaultMemSpaceAttr ? defaultMemSpaceAttr.getValue() : dflt;
   }
 
-  static constexpr mlir::ArrayRef<int64_t> expectedInputGridShape() {
-    return s_expectedInputGridShape;
-  }
-
+private:
+protected:
```
azecevic000: Is this by accident?

lib/Conversion/TTIRToTTIRGeneric/TTIRToTTIRGeneric.cpp
vwellsTT: I can change--I could see a loose criteria that valid IR we don't support yet should be pass `failure` instead, and assert should be more programmatic invariants (though there are other places where we don't hold to this ofc).

edit: actually this *is* an invariant here, the calling func has an `if` and returns `failure()` currently, so I think assert is appropriate here.

lib/Conversion/TTIRToTTIRGeneric/TTIRToTTIRGeneric.cpp
vwellsTT: Yes, thanks--missed this!

lib/Dialect/TTIR/IR/TTIROps.cpp
vwellsTT: So I agree, but in practice, this is a bit annoying.

I have my 3D grid PR which simplifies this logic a fair amount s.t. this code goes away and we don't default-construct layout in LowerToLayouts or here in the CompoundComponents.  I think the desirable state here is that the constructor which depends on worker grid should _only ever_ be invoked inside TTIRToTTIRGeneric, and in these other places we should always explicit pass in the `collapseIntervals` and `dimAlignments`--after all, LowerToLayout shouldn't need to create completely new alignments or collapsing when decomposing, AFAICT.

I had thought it made more sense to sort of quick fix here, then deal with the nasty rebase in the 3D layouts PR.  Really after sleeping on it I don't feel strongly, since I'll have a nasty rebase in any case--I can move some of that logic here if you'd prefer, and simplify that PR, or vice-versa.

lib/Dialect/TTIR/Utils/Utils.cpp
vwellsTT: Ah, TIL 👍

lib/Dialect/TTIR/IR/TTIROps.cpp
nsmithtt: I see, if you have a plan to deal with it immediately in your next PR then I'm OK with it landing this way.

vwellsTT: Cool, I'll add a TODO w/ ticket here and handle separately, thanks

Issue #3849, opened 2025-06-18
[Bounty $1000] Fix Hoisted + Non-Hoisted Gather Op Tests for Complex Cases


vwellsTT:
### Is your feature request related to a problem? Please describe

Currently, our ttir_builder tests for gather ops in both hoisted and non-hoisted path works for simple index cases, but not for more complex cases.

### Describe the solution you'd like

Basically what we want to see is:
* enable the complex indices case for the regular test_gather_op in test_ttir_ops.py here https://github.com/tenstorrent/tt-mlir/blob/main/test/python/golden/test_ttir_ops.py#L2677
* fix the golden s.t. the golden matches the actual op behavior
* enable the hoisted testcases for complex indices, fixing implementation as needed
* the end goal is 3-way match between the golden, the actual ttnn op, and the hoisted op

### Describe alternatives you've considered

This ticket seems to cover a mix of improper golden func implementation, and issues with converting TTIR gather op into linalg primitives.  The first step must be to get a golden which maps 1:1 with the actual op--after that, we can see if the hoisted version is incorrect w.r.t. a real golden, and fix it as needed.

### Steps to reproduce issue
1. Wait for https://github.com/tenstorrent/tt-mlir/pull/3573 to merge or checkout its branch
2. Uncomment the "SKIP" directive for complex_indices case(s) for test_gather in test_ttir_ops.py
3. Build ttrt (`cmake --build build -- ttrt`), and run pytests to generate binaries (e.g. `python -m pytest test/python/golden/test_ttir_ops.py --path=output --sys-desc=ttrt-artifacts/system_desc.ttsys -vv --allow-subset-mesh`)
4. Run the generated testcases `ttrt run output/ttnn/test_hoisted_gather\[ttnn-dim_args0-shap e0\]_ttnn.mlir.ttnn` and note golden mismatches.

### Additional context

https://github.com/tenstorrent/tt-mlir/pull/3911 https://github.com/tenstorrent/tt-mlir/pull/3910 and https://github.com/tenstorrent/tt-mlir/pull/3908 are related; the should add faulty impl/tests for this op.

https://github.com/tenstorrent/tt-mlir/issues/3884 This ticket tracks issues w/ golden check we experience in non-hoisted gather op; now this is directly related.  Note the mismatch in output patterns here.

Mountagha:
[...]

PR #4757, opened 2025-09-02
Fix gather op

Mountagha:
### Ticket
#3849 

### Problem description
Currently, our ttir_builder tests for gather ops in both hoisted and non-hoisted path works for simple index cases, but not for more complex cases. This PR adds support for complex cases.

### What's changed
Added a more general golden_gather
Modified the Op lowering so that it takes into account some parameters when permuting dims
make the builder Op takes into account the offset-dim when creating output shape.

### Checklist


test/python/golden/test_ttir_ops.py
```diff
@@ -2678,9 +2678,9 @@ def gather(
             [1],
             # Complex indices - f32.
             [1, 16, 1],
-            marks=pytest.mark.skip(
-                reason="Multi-dimensional gather has known issues, but the builder golden may also be incorrect: https://github.com/tenstorrent/tt-mlir/issues/3884"
-            ),
+            # marks=pytest.mark.skip(
```
vwellsTT: let's remove these

tools/builder/ttir/ttir_builder.py
```diff
@@ -4416,6 +4416,7 @@ def gather(
         indices_shape = self.get_shape(start_indices)
 
         # Batch dimensions: all dims from start_indices except index_vector_dim
+        '''
```
vwellsTT: same here, let's remove commented out code before PR lands

lib/Conversion/TTIRToTTIRDecomposition/TTIRToTTIRDecomposition.cpp
```diff
@@ -844,7 +844,7 @@ struct GatherToEmbeddingConversionPattern
   // Example: expectedOutputShape = [2, 3, 4, 5], offsetDims = [1, 3]
   // -> embeddingOutputShape = [2, 4, 15] -reshape-> [2, 4, 3, 5] -permute-> [2,
   // 3, 4, 5]
-  static ttir::PermuteOp reshapeAndPermuteOutput(
+  static mlir::Value reshapeAndPermuteOutput(
```
azecevic000: I see why you did this change, but with one of my following comments in mind, I do prefer `ttir::PermuteOp` as a return value in this case, it's implicitly convertible to `mlir::Value` if that's what a caller needs.

lib/Conversion/TTIRToTTIRDecomposition/TTIRToTTIRDecomposition.cpp
```diff
@@ -867,17 +867,37 @@ struct GatherToEmbeddingConversionPattern
          offsetDimsIndex++) {
       outputPermutation.push_back(offsetDims[offsetDimsIndex]);
     }
-    auto permutedOutputShape = ttmlir::utils::applyPermutation(
-        expectedOutputType.getShape(), outputPermutation);
+    auto invPerm = ttmlir::utils::inversePermutation(outputPermutation);
```
azecevic000: `invPerm` is not a very descriptive name; it's used in multiple places, so something like `inverseOutputPermutation` might be more suitable.

lib/Conversion/TTIRToTTIRDecomposition/TTIRToTTIRDecomposition.cpp
```diff
@@ -867,17 +867,37 @@ struct GatherToEmbeddingConversionPattern
...
+    bool isIdentity = true;
+    for (int64_t i = 0, e = (int64_t)invPerm.size(); i < e; ++i) {
+      if (invPerm[i] != i) {
+        isIdentity = false;
+        break;
+      }
+    }
+    if (isIdentity) {
+      return reshapedOutput.getResult();
+    }
```
azecevic000: This whole part is unnecessary; there is a folding in place that should deal with this case.

Mountagha: @azecevicTT aight will get rid of it. thanks 


> This whole part is unnecessary; there is a folding in place that should deal with this case.

lib/Conversion/TTIRToTTIRDecomposition/TTIRToTTIRDecomposition.cpp
Mountagha: > I see why you did this change, but with one of my following comments in mind, I do prefer `ttir::PermuteOp` as a return value in this case, it's implicitly convertible to `mlir::Value` if that's what a caller needs.

okay will address it asap

lib/Conversion/TTIRToTTIRDecomposition/TTIRToTTIRDecomposition.cpp
Mountagha: > `invPerm` is not a very descriptive name; it's used in multiple places, so something like `inverseOutputPermutation` might be more suitable.

sounds good

azecevicTT (approved):
TTIR changes LGTM.

vwellsTT:
@tapspatel or @ctodTT can you review this pls?

tapspatel:
builder, golden and test files changes are good

tapspatel (approved):
thank you for adapting to builder golden!

Issue #4196, opened 2025-07-24
Add constraint API for all ElementwiseBinaryCompositeOp


sgholamiTT:
Support the following ElementwiseBinaryCompositeOps
```
TTNN_BitwiseAndOp
TTNN_BitwiseOrOp
TTNN_BitwiseXorOp
TTNN_RemainderOp
TTNN_PowOp
TTNN_Atan2Op
TTNN_ScatterOp
TTNN_RemainderOp
```

PR #4760, opened 2025-09-02
Add constraints API for Atan2 and Remainder

ckaravasilisTT:
### Ticket
#4196

Added constraints for `Atan2Op`, `RemainderOp`.

### DIFF
--- a/test/unittests/OpModel/TTNN/Lib/TestOpModelLib.cpp
+++ b/test/unittests/OpModel/TTNN/Lib/TestOpModelLib.cpp
@@ -1415,7 +1415,8 @@ protected:
       const auto [cbSize, peakSize, outputSize, outputLayoutReadBack] =
           constraintsExp.get();
 
-      bool useGreaterThan = false;
+      bool useGreaterThan =
+          std::is_same_v<OpTy, Atan2Op> || std::is_same_v<OpTy, RemainderOp>;
       EXPECT_EQ_OR_GE(cbSize, expectedCbSize, useGreaterThan);
       EXPECT_EQ_OR_GE(peakSize, expectedPeakSize, useGreaterThan);
       EXPECT_EQ_OR_GE(outputSize, expectedOutputSize, useGreaterThan);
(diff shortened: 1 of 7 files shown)


sgholamiTT (approved):
Thanks for adding this.

azecevicTT (approved):

sgholamiTT:
I rebased this PR after [this](https://github.com/tenstorrent/tt-mlir/pull/4734) got merged.

lib/OpModel/TTNN/SingletonDeviceContext.cpp
```diff
@@ -15,7 +15,7 @@ namespace mlir::tt::ttnn::op_model {
 // This requires the device to be opened with sufficient trace region size. This
 // number is currently set based on manual testing of supported ops to
 // accommodate the highest required trace buffer size (2004992B)
-static constexpr size_t opModelDefaultTraceRegionSize = 5000000;
+static constexpr size_t opModelDefaultTraceRegionSize = 6000000;
```
xanderchin: Which tests fail if we don't bump the region size?

ckaravasilisTT: Some test(s) for `atan2` in `TestOpModelLib.cpp`. I don't think there's anything special with the tests, as they pass individually even with a much smaller region size (the max size of 2004992B mentioned in the comments seems to apply to `atan2` also). 

It seems that the more tests are executed (e.g. the entire file as opposed to individual op tests), the larger the trace buffer gets. I don't know why that happens.

test/unittests/OpModel/TTNN/Lib/TestOpModelLib.cpp
```diff
@@ -1279,7 +1279,8 @@ class OpModelBinaryEltwiseParam
       const auto [cbSize, peakSize, outputSize, outputLayoutReadBack] =
           constraintsExp.get();
 
-      bool useGreaterThan = false;
+      bool useGreaterThan =
+          std::is_same_v<OpTy, Atan2Op> || std::is_same_v<OpTy, RemainderOp>;
       EXPECT_EQ_OR_GE(cbSize, expectedCbSize, useGreaterThan);
       EXPECT_EQ_OR_GE(peakSize, expectedPeakSize, useGreaterThan);
       EXPECT_EQ_OR_GE(outputSize, expectedOutputSize, useGreaterThan);
```
rpavlovicTT, line 1422: Question about the new macro `EXPECT_EQ_OR_GE`. What's the rationale for it? When do we need to use EQ comparison?

ckaravasilisTT: Many ops of the same family are grouped and tested together. Some ops have different memory usage than the rest of their group, and repeating the tests with different expected values is sometimes awkward (i.e. might lead to more boilerplate and/or won't be as elegant). Given that we eventually want to move away from `EXPECT_EQ` (although it's not decided exactly how we'll test), using `EXPECT_EQ_OR_GE` was a quick, and probably temporary way of adding support for more ops.

Issue #4208, opened 2025-07-24
Add constraint API for miscellaneous Ops


sgholamiTT:
Support the following Ops (that don't fit into other categories IMO):
| No | Op | PR
| :---: | :---: | :---: |
| 1 | `TTNN_SliceDynamicOp` |  [PR](https://github.com/tenstorrent/tt-mlir/pull/4761)
| 2 | `TTNN_ConcatenateHeadsOp` |  [PR](https://github.com/tenstorrent/tt-mlir/pull/4722)
| 3 | `TTNN_RandOp ` | [PR](https://github.com/tenstorrent/tt-mlir/pull/4647)
| 4 | `TTNN_RMSNormOp ` | [PR](https://github.com/tenstorrent/tt-mlir/pull/4706)
----

This issue will be updated with other `TTNN` ops that might be added to the dialect.

PR #4761, opened 2025-09-02
[Optimizer] Add constraint APIs for SliceDynamicOp

sgholamiTT:
### Ticket
#4208 

### Problem description
Need constraint API for `SliceDynamicOp`.

### What's changed
This PR adds constraint/runtime APIs for `SliceDynamicOp`.
There's a small caveat though:
It is not possible to use the [dynamic version of slice](https://github.com/tenstorrent/tt-metal/blob/9e9a91bc1e87a3f9cc7fb38998bfdd3a849bfe2c/ttnn/cpp/ttnn/operations/data_movement/slice/slice.hpp#L97) in tt-metal since the validity of the op depends on the actual data that is stored in the `begins/ends/step` tensors (which is not available at compile time). Therefore, we have to approximate the op by using the [static version](https://github.com/tenstorrent/tt-metal/blob/9e9a91bc1e87a3f9cc7fb38998bfdd3a849bfe2c/ttnn/cpp/ttnn/operations/data_movement/slice/slice.hpp#L57) and imitating the worst case scenario for the static version which is:
- slicing from the beginning to the end (Capturing the entire tensor) for `getOpConstraint` which is the most memory intensive pattern for sliceOp.
- slicing from the beginning to the end with a step of 2 (Capturing all possible stripes of data from the input tensor) for `getOpRuntime` which is the most run time intensive pattern for sliceOp.

Note that this is a fairly accurate approximation since the dynamic version in metal also [converts the three tensors](https://github.com/tenstorrent/tt-metal/blob/9e9a91bc1e87a3f9cc7fb38998bfdd3a849bfe2c/ttnn/cpp/ttnn/operations/data_movement/slice/slice.cpp#L251) (`begins, ends, step`) to vectors and then calls the static version.

### Checklist


lib/OpModel/TTNN/TTNNOpModel.cpp
```diff
@@ -1610,6 +1610,122 @@ llvm::Expected<size_t> OpModel<SliceStaticOp>::getOpRuntime(
...
+  // in metal also converts the three tensors (begins, ends, step) to vectors
+  // and then calls the static version.
+  ::ttsl::SmallVector<int> stepVec =
+      ::ttsl::SmallVector<int>(inputShape.size(), 1);
+  ::ttsl::SmallVector<int> beginsVec =
+      ::ttsl::SmallVector<int>(inputShape.size(), 0);
+  ::ttsl::SmallVector<int> endsVec =
+      ::ttsl::SmallVector<int>(inputShape.size());
+  for (size_t i = 0; i < inputShape.size(); ++i) {
+    endsVec[i] = inputShape[i] - 1;
```
gfengTT: I think the end indices are exclusive so it should be `inputShape[i]` here.

lib/OpModel/TTNN/TTNNOpModel.cpp
```diff
@@ -1610,6 +1610,122 @@ llvm::Expected<size_t> OpModel<SliceStaticOp>::getOpRuntime(
...
+  }
+  (void)beginsVec;
+  (void)endsVec;
+  (void)stepVec;
```
gfengTT: Should these (void) statements be removed?

lib/OpModel/TTNN/TTNNOpModel.cpp
```diff
@@ -1610,6 +1610,122 @@ llvm::Expected<size_t> OpModel<SliceStaticOp>::getOpRuntime(
...
+  // in metal also converts the three tensors (begins, ends, step) to vectors
+  // and then calls the static version.
+  ::ttsl::SmallVector<int> stepVec =
+      ::ttsl::SmallVector<int>(inputShape.size(), 2);
+  ::ttsl::SmallVector<int> beginsVec =
+      ::ttsl::SmallVector<int>(inputShape.size(), 0);
+  ::ttsl::SmallVector<int> endsVec =
+      ::ttsl::SmallVector<int>(inputShape.size());
+  for (size_t i = 0; i < inputShape.size(); ++i) {
+    endsVec[i] = inputShape[i] - 1;
```
gfengTT: same as above

azecevicTT (approved):
+1 for @gfengTT comments, dialect changes LGTM

sgholamiTT:
Actually after talking with @sjameelTT, the peak memory usage is what was considered previously, i.e. `begins=[0,...,0]` and `ends=[n-1,...,n-1]`.
According to him:
> `begins=[0,...,0]` and `ends=[n,...,n]` just returns the input tensor so we won't even end up allocating CBs.

I will revert the change for getOpConstraint then.
FYI, @azecevicTT @gfengTT

gfengTT:
I wrote a simple test with input shape {4, 32, 32}, begins {0, 0, 0}, and ends {4, 32, 32}.
- For steps {2, 2, 2}, the output shape is {2, 16, 16}.
- For steps {4, 4, 4}, the output shape is {1, 8, 8}.
Maybe there's a misunderstanding here? Let's clarify with Saad.

mtopalovicTT:
> I wrote a simple test with input shape {4, 32, 32}, begins {0, 0, 0}, and ends {4, 32, 32}.
> 
> * For steps {2, 2, 2}, the output shape is {2, 16, 16}.
> * For steps {4, 4, 4}, the output shape is {1, 8, 8}.
>   Maybe there's a misunderstanding here? Let's clarify with Saad.

I think what @sgholamiTT is referring to is when step is 1,  then we can just return input tensor.

sgholamiTT:
> I wrote a simple test with input shape {4, 32, 32}, begins {0, 0, 0}, and ends {4, 32, 32}.
> 
> * For steps {2, 2, 2}, the output shape is {2, 16, 16}.
> * For steps {4, 4, 4}, the output shape is {1, 8, 8}.
>   Maybe there's a misunderstanding here? Let's clarify with Saad.

Hmm, I'm not talking about `step` and `output shape`.
As @mtopalovicTT mentioned, I'm talking about the circular buffer size.
- With `begins=[0,...,0]`, `ends=[n,...,n]`, `steps[1,...,1]`: `ttnn::slice` just returns the input tensor and the allocated CB size is 0.
- With `begins=[0,...,0]`, `ends=[n-1,...,n-1]`, `steps[1,...,1]`: `ttnn::slice` actually allocates a tensor and excludes the last row. So the allocated CB size is not zero anymore, hence, the worst case in terms of memory usage.

This is also observed in the last two commits in this branch: you can compare [this](https://github.com/tenstorrent/tt-mlir/pull/4761/commits/b0ad7718554ddea19e92a408f59ca862f80658d3) with [this](https://github.com/tenstorrent/tt-mlir/pull/4761/commits/e43f003af80e38a31a6d9cabce17d8833e33076e) and note how the `cbL1PeakSize` increased from 0 to 4096 when I changed `ends` to `n-1`.


(What I described here is just for memory usage, for runtime it's a bit different. Extracting strides of data (with any `step != 1`) gives you the most runtime intensive pattern for `ttnn::slice`).

Issue #4726, opened 2025-08-29
Add fusing pattern for avg_pool for efficientnet


mvasiljevicTT:
There are two ways global average pooling is lowered from torch to stableHLO. The one in vovnet is already covered by `AveragePoolingWithPoolingDenominatorFusionPattern`, but in efficientnet it is lowered as:
```
sum <dim = [2,3], keepdim=False>     full
   |                                  |
  multiply ---------------------------/
   |
  reshape (N,C) -> (N,C,1,1)
```
so there is a need for another pattern that would fuse this into one `avg_pool` too.

Issue #4666, opened 2025-08-26
Add pattern for global_avg_pool2d


mvasiljevicTT:
Efficientnet and vovnet use global average pooling, implemented in torch as adaptive average pooling with output size `(1,1)`. 
Since there is a [`global_avg_pool2d`](https://docs.tenstorrent.com/tt-metal/latest/ttnn/ttnn/api/ttnn.global_avg_pool2d.html#ttnn.global_avg_pool2d) op in metal, this special case of `avg_pool2d` should be folded as `global_avg_pool2d`.

mvasiljevicTT:
Partially solved with [this commit](https://github.com/tenstorrent/tt-mlir/pull/5174) by fusing pattern from [this issue](https://github.com/tenstorrent/tt-mlir/issues/4726) directly as `global_avg_pool2d`. 
There are still some `avg_pool2d` ops created through different fusing pattern that could be folded as `global_avg_pool2d`.

mvasiljevicTT:
Soon(ish) `global_avg_pool2d` will be deprecated as it will be a special case of `avg_pool2d`, working the way it is now. 
At that moment, all occurrences of  `global_avg_pool2d` will be converted back to `avg_pool2d`.

As remaining avg_pools are not creating OOM issues like previous ones, they can just wait for that.

PR #4766, opened 2025-09-03
Average pooling from torch-xla fusing pattern

mvasiljevicTT:
### Ticket
[Link to Github Issue](https://github.com/tenstorrent/tt-mlir/issues/4726)

### Problem description
When global average pooling is used in efficientnet as `nn.AdaptiveAvgPool(1)` it is lowered through torch-xla as:
```
sum <dim=[2,3], keepdim=False>     full <value=1/(H*W)>
   |                                  |
  multiply ---------------------------/
   |
  reshape (N, C) -> (N, C, 1, 1)
```
This pattern represents global average pooling but is not recognized and optimized in the current pipeline

### What's changed
Added a new pattern `GlobalAveragePoolingPattern` that matches `multiply(sum, 1/(H*W)) `and rewrites it as `avg_pool2d` with the right window size. 

When the `sum` does not keep dimensions, we insert a `reshape` after `pooling` to restore the expected 2D shape:

```
sum <dim=[2,3], keepdim=False>     full <value=1/(H*W)>             avg_pool(act, window=[h,w], stride=1, padding=0)
   |                                  |                     ->          |
  multiply ---------------------------/                             reshape (N, C, 1, 1) -> (N, C)
```

 If a `reshape` already exists (as is often the case), It will be folded away by canonicalization:
```
sum <dim=[2,3], keepdim=False>     full <value=1/(H*W)>             avg_pool(act, window=[h,w], stride=1, padding=0)
   |                                  |                                   |                                                     
  multiply ---------------------------/                     ->      reshape (N, C, 1, 1) -> (N, C)                   ->   avg_pool(act, window=[h,w], stride=1, padding=0)
   |                                                                    |
  reshape (N, C) -> (N, C, 1, 1)                                    reshape (N, C) -> (N, C, 1, 1)
```

We use `ttnn.avg_pool2d` since `ttnn.global_avg_pool2d` is not yet supported in mlir dialects ([#4666](https://github.com/tenstorrent/tt-mlir/issues/4666)).
### Checklist


lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
```diff
@@ -1487,6 +1487,165 @@ class AveragePoolingWithPoolingDenominatorFusionPattern
...
+      rewriter.replaceOp(multiplyOp, reshapePoolOp.getResult());
+    } else {
+      rewriter.replaceOp(multiplyOp, poolingOp.getResult(0));
+    }
+
+    return mlir::success();
+  };
+
+private:
+  bool validateFullOp(FullOp fullOp, ArrayRef<int64_t> inputShape) const {
```
mtopalovicTT, line 1371: You can deduce input shape from fullOp

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
```diff
@@ -1487,6 +1487,165 @@ class AveragePoolingWithPoolingDenominatorFusionPattern
...
+
+    return config;
+  }
+
+  RankedTensorType createPoolingOutputType(RankedTensorType inputType) const {
+    SmallVector<int64_t> poolOutputShape(inputType.getShape());
+    poolOutputShape[SPATIAL_HEIGHT_DIM] = 1;
+    poolOutputShape[SPATIAL_WIDTH_DIM] = 1;
+
+    return RankedTensorType::get(poolOutputShape, inputType.getElementType(),
```
mtopalovicTT, line 1492: You need to update encoding too:
`RankedTensorTypeFactory::create(inputType, poolOutputShape);`

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
```diff
@@ -1487,6 +1487,165 @@ class AveragePoolingWithPoolingDenominatorFusionPattern
...
+  }
+
+  PoolingOp createPoolingOp(mlir::PatternRewriter &rewriter,
+                            SumOp sumOp) const {
+
+    auto config = createPoolingConfig(sumOp.getInput().getType().getShape());
+    auto outputType = createPoolingOutputType(sumOp.getInput().getType());
+    auto loc = sumOp.getLoc();
+
+    auto poolingOp = ttir::utils::createDPSOp<ttir::PoolingOp>(
```
mtopalovicTT: Why pooling? Why not avgpool2d?

mtopalovicTT:
>  it will be removed by EraseInverseOps:

It will be folded away by canonicalization.

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
sdjordjevicTT: Yep, better go with avgpool2d directly

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
sdjordjevicTT: Is there encoding in TTIR?

mtopalovicTT: Ufff my bad, forgot this is TTIR...

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
mvasiljevicTT: SumOp has (N, C, H, W) input and (N, C) or (N, C, 1, 1) output shape, so fullOp will have either (N, C) or (N, C, 1, 1) shape and here I want to check that its values are 1/H*W

mtopalovicTT: Oh, my bad. Guess I'm missing stuff left and right on this one :)

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
mvasiljevicTT: changed to using AvgPool2dOp

Issue #4747, opened 2025-09-01
[TTNN Dialect] Merge ToLayout ops into creation ops


sdjordjevicTT:
Canonicalize ToLayout ops into the TTNN creation ops. All creation ops support specifying dtype, layout, and memory config, hence we can merge ttnn.to_layout ops into them.

Issue #4743, opened 2025-08-29
Use interface instead of OutputTensorSpec trait


sdjordjevicTT:
We should consider using interfaces instead of traits in this case:
https://github.com/tenstorrent/tt-mlir/pull/4658#discussion_r2307680037

Issue #4066, opened 2025-07-14
Arange op now supports specifying output layout


sdjordjevicTT:
With this Metal issue fixed:
https://github.com/tenstorrent/tt-metal/issues/20251#event-18549192694

We can remove all special casing from our codebase.

PR #4768, opened 2025-09-03
Merging to layout ops into creation ops

sdjordjevicTT:
### Ticket
Solves https://github.com/tenstorrent/tt-mlir/issues/4747
Solves https://github.com/tenstorrent/tt-mlir/issues/4743
Solves https://github.com/tenstorrent/tt-mlir/issues/4066

### Problem description
TTNN dialect had redundant to_layout operations that were often placed after creation operations (empty, arange, rand, full, constant, zeros, ones). This created unnecessary intermediate operations when the creation ops could directly produce tensors with the desired layout, dtype, and memory configuration. This redundancy led to less efficient IR and suboptimal code generation.

### What's changed
  - Introduced TTNNTensorSpecInterface: Created a new interface (TTNNTensorSpecInterface.h/td) that allows creation ops to specify their output tensor properties (dtype, layout, memory config) directly during creation
  - Moved tensor specification logic: Extracted tensor specification functionality from TTNNTraits into the dedicated interface, reducing code duplication and improving modularity
  - Enhanced creation ops: Updated creation operations (empty, arange, rand, full, constant, zeros, ones) to implement the tensor spec interface, allowing them to merge with subsequent to_layout operations
  - Added canonicalization patterns: Implemented canonicalization that automatically merges to_layout operations into preceding creation ops when possible, eliminating redundant IR
  - Added comprehensive tests: Created extensive test coverage (layout_to_creation_ops_canonicalizer.mlir) to verify the merging behavior works correctly across all creation ops
  
 Example IR before canonicalization:
 ```mlir
%1 = "ttnn.empty"(%0) <{
      dtype = #ttcore.supportedDataTypes<f32>,
      layout = #ttnn.layout<row_major>,
      memory_config = #ttnn.memory_config<#dram, <interleaved>>,
      shape = #ttnn.shape<32x32>
    }> : (!ttnn.device) -> tensor<32x32xf32, #ttnn_layout_dram_f32_rm>

%2 = "ttnn.to_layout"(%1) <{
      dtype = #ttcore.supportedDataTypes<bf16>,
      layout = #ttnn.layout<tile>,
      memory_config = #ttnn.memory_config<#dram, <interleaved>>
    }> : (tensor<32x32xf32, #ttnn_layout_dram_f32_rm>) -> tensor<32x32xbf16, #ttnn_layout_dram_bf16_tile> 
```

Example IR after canonicalization:
```mlir
%1 = "ttnn.empty"(%0) <{
      dtype = #ttcore.supportedDataTypes<bf16>,
      layout = #ttnn.layout<tile>,
      memory_config = #ttnn.memory_config<#dram, <interleaved>>,
      shape = #ttnn.shape<32x32>
    }> : (!ttnn.device) -> tensor<32x32xbf16, #ttnn_layout_dram_bf16_tile>
```

### Checklist


lib/Dialect/TTNN/IR/TTNNOps.cpp
```diff
@@ -1792,6 +1797,139 @@ mlir::OpFoldResult ttnn::ToLayoutOp::fold(FoldAdaptor adaptor) {
...
+                tensorSpecOp->getOperand(i).getType())) {
+          deviceExist = true;
+          break;
+        }
+      }
+
+      if (!deviceExist) {
+        tensorSpecOp->insertOperands(
+            0, {utils::getOrInsertDevice(rewriter, toLayoutOp)});
+      }
```
azecevic000: Device is usually the first operand when exists, but it doesn't have to be, and you already check that in the loop, so you can just do `insertOperand(i, ...)` in the `if` in the previous `for`.

lib/Dialect/TTNN/IR/TTNNOps.cpp
```diff
@@ -1792,6 +1797,139 @@ mlir::OpFoldResult ttnn::ToLayoutOp::fold(FoldAdaptor adaptor) {
...
+        if (mlir::isa<mlir::tt::ttnn::DeviceType>(
+                tensorSpecOp->getOperand(i).getType())) {
+          tensorSpecOp->eraseOperand(i);
+          break;
+        }
+      }
+    }
+
+    // Update the tensor ranked type of creation op with the new layout and new
+    // data type
```
azecevic000: Nit: `.`

lib/Dialect/TTNN/IR/TTNNOps.cpp
```diff
@@ -1792,6 +1797,139 @@ mlir::OpFoldResult ttnn::ToLayoutOp::fold(FoldAdaptor adaptor) {
...
+    tensorSpecOp->getResult(0).setType(
```
azecevic000, line 1873: You should at least assert that there is a one result (or even better, check at the beginning and break out of the pattern if that's not true), there are cases when this isn't necessarily true.

lib/Dialect/TTNN/IR/TTNNOps.cpp
```diff
@@ -1792,6 +1797,139 @@ mlir::OpFoldResult ttnn::ToLayoutOp::fold(FoldAdaptor adaptor) {
...
+    return success();
+  });
+
+  // Merging to layout op into TTNN empty op on host should produce ttnn.zeros
+  // op on host.
+  patterns.add(+[](mlir::tt::ttnn::ToLayoutOp toLayoutOp,
+                   mlir::PatternRewriter &rewriter) {
+    // Check if the toLayoutOp is being applied on a TTNN empty op
+    EmptyOp emptyOp = mlir::dyn_cast_if_present<mlir::tt::ttnn::EmptyOp>(
+        toLayoutOp.getInput().getDefiningOp());
```
azecevic000: Nit: `toLayoutOp.getInput(). getDefiningOp<ttnn::EmptyOp>()`

lib/Dialect/TTNN/IR/TTNNOps.cpp
```diff
@@ -1792,6 +1797,139 @@ mlir::OpFoldResult ttnn::ToLayoutOp::fold(FoldAdaptor adaptor) {
...
+    if (!emptyOp) {
+      return mlir::failure();
+    }
+
+    // Verify that the empty op has a single use.
+    if (!emptyOp->hasOneUse()) {
+      return failure();
+    }
+
+    // Verify that the target buffer type is a system memory
```
azecevic000: Nit: `.`

lib/Dialect/TTNN/IR/TTNNOps.cpp
```diff
@@ -1792,6 +1797,139 @@ mlir::OpFoldResult ttnn::ToLayoutOp::fold(FoldAdaptor adaptor) {
...
+    // data type
+    tensorSpecOp->getResult(0).setType(
+        RankedTensorType::Builder(
+            mlir::cast<RankedTensorType>(creationOp->getResult(0).getType()))
+            .setEncoding(ttnnLayoutAttr)
+            .setElementType(ttnnLayoutAttr.getScalarElementType()));
+
+    rewriter.finalizeOpModification(tensorSpecOp);
+    rewriter.replaceAllOpUsesWith(toLayoutOp, tensorSpecOp);
+    rewriter.eraseOp(toLayoutOp);
```
azecevic000, line 1881: Nit: should be erased by default as a pure op

lib/Dialect/TTNN/IR/TTNNOps.cpp
```diff
@@ -1792,6 +1797,139 @@ mlir::OpFoldResult ttnn::ToLayoutOp::fold(FoldAdaptor adaptor) {
...
+        emptyOp, toLayoutOp.getType(), /*device=*/nullptr, emptyOp.getShape(),
+        toLayoutOp.getDtypeAttr() ? toLayoutOp.getDtypeAttr()
+                                  : emptyOp.getDtypeAttr(),
+        toLayoutOp.getLayoutAttr() ? toLayoutOp.getLayoutAttr()
+                                   : emptyOp.getLayoutAttr(),
+        toLayoutOp.getMemoryConfigAttr() ? toLayoutOp.getMemoryConfigAttr()
+                                         : emptyOp.getMemoryConfigAttr());
+
+    rewriter.replaceAllOpUsesWith(toLayoutOp, zerosOp);
+    rewriter.eraseOp(toLayoutOp);
```
azecevic000, line 1920: Same comment as in the above pattern.

azecevicTT (approved):
LGTM, few minor comments.

lib/Dialect/TTNN/IR/TTNNOps.cpp
sdjordjevicTT: As we agreed offline, I added TTNNDeviceOperandInterface to handle this better.

lib/Dialect/TTNN/IR/TTNNOps.cpp
sdjordjevicTT: Added a check to verify that the pattern only works if the creation op has only 1 result.

lib/Dialect/TTNN/IR/TTNNOps.cpp
sdjordjevicTT: I am not opposed to that, but I prefer to make this clear from the pattern itself. If we are explicit, it is much easier to reason about the pattern itself.

lib/Dialect/TTNN/IR/TTNNOps.cpp
sdjordjevicTT: Same as above.

include/ttmlir/Dialect/TTNN/Interfaces/TTNNDeviceOperandInterface.td
```diff
@@ -0,0 +1,52 @@
...
+      /*desc=*/[{
+        Set the device operand value. For operations with required device operands,
+        this sets the device value. For operations with optional device operands,
+        this assigns the device value if provided, or clears the operand if null.
+      }],
+      /*retTy=*/"void",
+      /*methodName=*/"setDevice",
+      /*args=*/(ins "Value": $device),
+      /*methodBody=*/[{
+        // Use SFINAE to detect if this is an optional operand
```
mtopalovicTT, line 35: Not ideal but I get that there is no other way.
If I may propose just a little enhancement. For ops that require the device and the caller passes `nullptr` can we add explicit assert? I had situations before where I was passing in nullptr to some required parameter and mlir generated some horrendous callstack without error message. 

```cpp
        if (device) {
          $_op.getDeviceMutable().assign(device);
        } else if constexpr (std::is_same_v<decltype($_op.getDeviceMutable()), ::mlir::MutableOperandRange>) {
          $_op.getDeviceMutable().clear();
        } else {
          llvm::report_fatal_error("Attempting to clear a required device operand");
        }
```

include/ttmlir/Utils.h
```diff
@@ -34,7 +34,6 @@ constexpr inline llvm::StringLiteral g_skipQdqCommuteAttrName =
     "ttir.skip_qdq_commute";
 constexpr inline llvm::StringLiteral g_decomposedFromAllReduceAttrName =
     "ttir.decomposed_from_all_reduce";
-constexpr inline llvm::StringLiteral g_outputDtypeAttrName = "dtype";
```
mtopalovicTT, line 37: can we also delete layout/memory config attributes and use checks like isa<TTNNLayoutInterface...> to check if attribute exists and override it?

lib/Dialect/TTNN/Transforms/Workarounds/TTNNWorkaroundsPatterns.cpp
```diff
@@ -190,13 +190,13 @@ workaroundOutputOperand(mlir::TypedValue<RankedTensorType> opResult,
       op->setAttr(ttmlir::utils::g_outputLayoutAttrName, updatedLayoutAttr);
     }
 
-    if (outputWorkaroundResults.tensorDataTypeResult.isModified() &&
-        (op->hasTrait<ttnn::HasDTypeTrait>() ||
-         op->getAttrDictionary().get(ttmlir::utils::g_outputDtypeAttrName))) {
+    TTNNDtypeOpInterface dtypeOp =
+        mlir::dyn_cast<TTNNDtypeOpInterface>(op.getOperation());
+    if (outputWorkaroundResults.tensorDataTypeResult.isModified() && dtypeOp) {
```
mtopalovicTT, line 196: Should we assert in case dtype workaround is set but dtypeOp is nullptr?

mtopalovicTT (approved):

lib/Dialect/TTNN/Transforms/Workarounds/TTNNWorkaroundsPatterns.cpp
```diff
@@ -190,13 +190,13 @@ workaroundOutputOperand(mlir::TypedValue<RankedTensorType> opResult,
       op->setAttr(ttmlir::utils::g_outputLayoutAttrName, updatedLayoutAttr);
     }
 
-    if (outputWorkaroundResults.tensorDataTypeResult.isModified() &&
-        (op->hasTrait<ttnn::HasDTypeTrait>() ||
-         op->getAttrDictionary().get(ttmlir::utils::g_outputDtypeAttrName))) {
+    TTNNDtypeOpInterface dtypeOp =
```
mtopalovicTT, line 194: Do you want to do this for layout also?

include/ttmlir/Dialect/TTNN/Interfaces/TTNNDeviceOperandInterface.td
sdjordjevicTT: Totally agree, will add this enhancement in the next push.

include/ttmlir/Utils.h
sdjordjevicTT: I thought this would require a greater refactor, but it seems quite simple with the newly introduced interfaces. Will remove them.

lib/Dialect/TTNN/Transforms/Workarounds/TTNNWorkaroundsPatterns.cpp
sdjordjevicTT: I don't think that we want to assert in that case. We want to allow for workarounds on other ops as well. These are just special handlings for ops that have these attributes, and that we keep these attributes in sync with their expected output TTNNLayout attrs.

lib/Dialect/TTNN/Transforms/Workarounds/TTNNWorkaroundsPatterns.cpp
sdjordjevicTT: Yep, will fix layout and memory_config as well. :)

Issue #4772, opened 2025-09-03
Testing trace in EmitC


azecevicTT:
Currently, it's not possible to test trace in EmitC (or ttrt in general). There are two main issues:
1. Program cache has to be enabled
2. Upper bound of trace size needs to be known at compile time

Issue #3672, opened 2025-06-04
☔ Add trace support to emitC


jnie-TT:
We want to add metal trace support to emitC as well, currently blocked by #3606

svuckovicTT:
@azecevicTT after the "initial" support has landed, what's the remaining work in terms of making trace functional for "all" models?

azecevic000:
There is an issue of testing that needs to be resolved. Other than that, as far as my modest knowledge goes, the answer would be nothing (if we are talking about parity with the runtime). Current constraints (they are soft constraints that are tied to the current implementation) are:
- there is an expectation that there won't be multiple traces in a single program
- there is an expectation that the model returns a single tensor

If these turn out to be too strong constraints, implementation could be adapted relatively easily to accommodate multiple traces and/or output tensors.

svuckovicTT:
Synced offline, leaving notes.

Need to handle tensors that end up living in system memory, potentially related to how weights/biases are lowered for convs.

Trying to codegen resnet with:
```
tt-alchemist generate-cpp test/ttmlir/models/resnet_hf.mlir --pipeline-options "system-desc-path=/localdev/svuckovic/_workspace/repos/tt-mlir/ttrt-artifacts/system_desc.ttsys enable-trace=true" -o cpp_resnet --standalone
```

errors out with
```
loc("ResNetForImageClassification":0:0): error: 'ttnn.capture_or_execute_trace' op All input arguments of trace function must be on device.<block argument> of type 'tensor<64x3x7x7xbf16, #ttnn.ttnn_layout<(d0, d1, d2, d3) -> (d0 * 21 + d1 * 7 + d2, d3), <1x1>, memref<1344x7xbf16, #ttnn.buffer_type<system_memory>>>>' at index: 99 is not on device.
Failed to run TTIR to EmitC pipeline
Failed to generate standalone solution in: cpp_resnet
```

PR #4773, opened 2025-09-03
Initial support of trace in EmitC

azecevicTT:
### Ticket
https://github.com/tenstorrent/tt-mlir/issues/3672

### Problem description
There isn't a way to codegen trace in EmitC.

### What's changed
Added a lowering for trace-related ops in TTNN to EmitC. Most of them are straightforward lowering for ops that exist in TTNN lib. The only op that required non-trivial lowering logic is `ttnn.capture_or_execute_trace_op`. It's implemented as a call to the generated function, which on first call calls `capture_calle`, and afterwards `execute_calle`. State between calls is stored in global variables. Initially, I planned to track in the state through local variables with static lifetime, but EmitC doesn't support static variables.

Since there isn't `TraceId` as a type in TTNN dialect (it's modeled as a tensor with a single scalar value), I've added a `ttnn.trace_id` attribute that's stored in the encoding of the tensor for the tensor that represents `TraceId`. This is necessary because otherwise, `ttnn::Tensor` would be generated as a type for the trace id.

I've tested in manually through ttnn-standalone, but there isn't an easy way currently to test through ttrt (that's the same issue for trace through the runtime). I'll look into alternatives to enable testing in CI (https://github.com/tenstorrent/tt-mlir/issues/4772).

### Checklist


test/ttmlir/EmitC/TTNN/trace/mnist.mlir
```diff
@@ -0,0 +1,7 @@
+UNSUPPORTED: true
+https://github.com/tenstorrent/tt-mlir/issues/4772
```
svuckovicTT: Does having the link outside of a comment do anything differently than if it were inside a comment?

lib/Target/TTNN/TTNNToFlatbuffer.cpp
```diff
@@ -124,13 +124,32 @@ flatbuffers::Offset<::tt::target::ttnn::TensorDesc>
 tensorTypeToFlatbuffer(FlatbufferObjectCache &cache, Type type,
                        ttcore::DeviceAttr deviceAttr) {
   auto tensorType = mlir::cast<RankedTensorType>(type);
+
+  // Runtime deals with a trace id as a system memory tensor. Appropriate
+  // TTNNLayoutAttr is created for it.
```
svuckovicTT, line 129: Consider creating a tracking issue for this. Maybe @jnie-TT would be able to help with this on the runtime side? That'd really help us avoid the tensor shenanigans.

tools/ttnn-standalone/ttnn-precompiled.hpp
```diff
@@ -67,6 +69,7 @@ namespace ttnn {
 class DeviceGetter {
 public:
   static constexpr std::size_t l1SmallSize = 1 << 15;
+  static constexpr std::size_t traceRegionSize = 1 << 16;
```
svuckovicTT: Feel free to increase this. This region is reserved in DRAM, so we should be good to use more. I can't recall the exact numbers, but I think resnet (a fairly small model by today's standards) was around 50-something or 90-something kB.

include/ttmlir/Conversion/TTNNToEmitC/EmitCConversion.h
```diff
@@ -1704,6 +1711,11 @@ class EmitCTTNNEmitter {
     return emit(memoryConfigAttr);
   }
 
+  mlir::Attribute emitCQ(uint32_t cqId) {
```
svuckovicTT: Can this be done via template specialization so that we keep the `emit` signature?

test/ttmlir/EmitC/TTNN/trace/mnist.mlir
azecevic000: No, I guess it works because once `UNSUPPORTED: true` is parsed, the rest of the content is discarded. I will put it under the comment.

tools/ttnn-standalone/ttnn-precompiled.hpp
azecevic000: Bumped it to `1 << 20` (1MB).

include/ttmlir/Conversion/TTNNToEmitC/EmitCConversion.h
azecevic000: Yeah, good idea, done.

lib/Target/TTNN/TTNNToFlatbuffer.cpp
azecevic000: It depends on the way we want to approach TraceId. We can either make it a strong type with no connection with `RankedTensorType`, but that would require some bigger changes in the runtime as I understand it. Previous solution where `trace_id` was treated like a regular system memory tensor wasn't a viable approach for codegen paths (EmitC, EmitPy etc.). So the middle ground was to create a special attribute that would be attached as an encoding of a tensor that represents `trace_id`. The only user of the flatbuffer currently is the runtime, so I'm not sure if it makes sense to have something like that modeled in the flatbuffer and dealt with in the runtime. Just wanted to give some context, I agree this is not the best approach, but since it's contained only in one place I don't think it will cause much issue, @jnie-TT has more context about the work needed to be done in the runtime, and if it makes sense to model `trace_id` (as a type or encoding of a tensor) in the flatbuffer and to deal with it in the runtime.

Issue #4207, opened 2025-07-24
Add constraint API for Device related Ops


sgholamiTT:
Support the following Device related Ops:
```
TTNN_GetDeviceOp
TTNN_ToDTypeOp
TTNN_ToDeviceOp
TTNN_FromDeviceOp
```

gfengTT:
- GetDevice: TBD, may not need a constraint API.
- ToDevice/FromDevice: TBD, includes memory I/O, may not need a constraint API.
- ToDType: similar to Typecast, except it applies to host tensors. WIP (Btw: https://github.com/tenstorrent/tt-metal/blob/42cddc12d06f50a09c71d8482a0acb518e949377/ttnn/ttnn/operations/core.py#L492)

PR #4774, opened 2025-09-03
add constraint apis to device related ops, no implementation

sgholamiTT:
### Ticket
#4207 

### Problem description
Constraint APIs for device related ops

### What's changed
Constraint APIs are not needed for device related ops, but they block the end goal of adding the APIs to `TTNNOps` class. Therefore, we just add the APIs and issue errors.
- `getDeviceOp`: is used once in each input graph and the optimizer does not need to make modifications to it.
- `toDeviceOp` and `fromDeviceOp`: require memory IO, cannot be implemented.
- `ToDTypeOp `: host only op, cannot be implemented rn.

### Checklist
- [X]Existing tests provide coverage for changes


sgholamiTT:
@gfengTT Please sign off this change and let me know if the descriptions of this PR matches your expectations.

PR #4787, opened 2025-09-03
Add e2e pytest for d2m profiling flow

jdesousa-TT:
### Ticket
#4786 

### Problem description
We don't test the profiler e2e on d2m flatbuffers. This has left the profiler broken a bunch of times without us noticing. 

### What's changed
Add a ttrt pytest that will run the `ttrt perf` on a ttm flatbuffer, and then validate that there is coherent data in the tracy output files.

### Checklist


arminaleTT (approved):
😍

Issue #4143, opened 2025-07-21
[OpModel] Move L1 total usage calculation to tt-metal


rpavlovicTT:
Right now `getOpConstraints` API returns 

```
struct OpConstraints {
  size_t cbL1PeakSize;                   // CB L1 peak allocation in bytes
  size_t tensorL1PeakSize;               // Tensor L1 peak allocation in bytes
  size_t outputL1BufferSize;             // Output L1 buffer allocation in bytes
  tt::ttnn::TTNNLayoutAttr outputLayout; // Layout of the output tensor
```

On the other hand in `ShardSolver` we compute L1 usage as combination of metrics returned by above API and internal layouts 
```
producerL1OutputUsage + tensorUsage + cBUsagePeak
```
where:
- `producerL1OutputUsage` is size of the input tensor in L1
- `tensorUsage` is `tensorL1PeakSize` (this includes output tensor too)

Couple of problems with this approach are:
1. We are relying on peak usage of CBs and Tensors within an op, thus we are pessimistic because this is an upper bound of L1 usage. It's possible, that real usage is lower than sum of these.
2. We calculate input's footprint by inspecting its layout. We should get this from backend too. For example, conv2d op has `deallocate_activation` option which should affect this footprint, which we do not take into account. So, this is not totally accurate calculation in all cases.
3.  For some reason in `query_op_constraints` in tt-metal we have outer graph capture which is ignored, thus we don't track allocation of input tensors in the inner graph capture.

This issue proposes to move L1 usage calculation fully to tt-metal side. It should calculate real peak usage by tracking all CB and Buffer allocations. Also, it should include input tensors as well. Ideally, `getOpConstraints` API will return
```
struct OpConstraints {
  size_t L1PeakUsage;
  tt::ttnn::TTNNLayoutAttr outputLayout;
// optionally, additional metrics that are possible to capture like
  size_t CBSizeAtPeakUsage;
}
```

@xanderchin I am assigning this to you, and will let you reassign to someone from your team. Let me know if you have questions.

gfengTT:
Picking this up. 

Just recorded some trace json files to better understand the issue. Are the events in chronological order? I'm seeing a `circular_buffer_deallocate_all` first, followed by some `circular_buffer_dallocate`s.

[trace_binaryadd.json](https://github.com/user-attachments/files/21553798/trace_binaryadd.json)
[trace_matmul.json](https://github.com/user-attachments/files/21553797/trace_matmul.json)
[...]

PR #4790, opened 2025-09-04
Add peakMemorySize to OpConstraints

gfengTT:
This is the new field returned by query_op_constraints, which is the peak of combined CB+L1 usage.

In test files, expectedTotalPeakSize is added to the struct of expected result.

### Ticket
#4143

### Issues
- If the intention is to replace cbPeak and l1Peak with this combinedPeak, shall we remove cbPeak and l1Peak now?
- Inconsistent names and use of snake_case instead of CamelCase in some tests. I've changed as much as possible.
- Btw, most of the time totalPeakSize is the sum of cbPeakSize and l1PeakSize, not surprisingly.


include/ttmlir/OpModel/TTNN/TTNNOpConstraints.h
```diff
@@ -18,22 +18,24 @@ namespace mlir::tt::ttnn::op_model {
  */
 
 struct OpConstraints {
-  size_t cbL1PeakSize;                   // CB L1 peak allocation in bytes
-  size_t tensorL1PeakSize;               // Tensor L1 peak allocation in bytes
-  size_t outputL1BufferSize;             // Output L1 buffer allocation in bytes
+  size_t cbL1PeakSize;       // CB L1 peak allocation in bytes
+  size_t tensorL1PeakSize;   // Tensor L1 peak allocation in bytes
+  size_t peakMemorySize;     // Peak memory (CB+L1) allocation in bytes
```
rpavlovicTT: Does this include input tensors?

gfengTT: ~Yes, it does.~

rpavlovicTT: I ask because I looked at https://github.com/tenstorrent/tt-metal/pull/27253/files and I see we still have 2 levels of graph capture (outer/inner). So I am not sure if that affects memory allocations with respect to input tensors. 
If you tested this, then I am fine with proceeding with uplifting changes.

rpavlovicTT: With your change in OpConstraints structure, we should be able to replace existing code 
```
producerL1OutputUsage + tensorUsage + cBUsagePeak
```
with the peak usage from the structure

gfengTT: I see. @odjuricicTT mentioned he tested it and it looked good.
cc @bmalesevicTT

rpavlovicTT: @odjuricicTT is OOF. 
@gfengTT can you confirm please? I checked quickly by modifying `tests/ttnn/unit_tests/gtests/test_graph_query_op_constraints.cpp` and it seems to me inputs are not accounted for peak memory usage.

gfengTT: @rpavlovicTT I should correct myself - it doesn't include input tensor size.

The intention is for the new field to replace the sum of existing cb peak and l1 peak (`tensorUsage + cBUsagePeak`). I guess @odjuricicTT 's tests are based on this. I'd prefer this PR to be merged first, given that: 1) the metal uplift has gone through; 2) it's backward compatible, won't affect optimizer  code; 3) any edit to constraint api is very likely to cause a conflict with this PR.

Meanwhile I have another metal branch that would include the input tensor size: please modify third_party/CMakeLists.txt and change `TT_METAL_VERSION` to 
2452356bfea4ccff43da3ee850b0f06381d72b44. Please then try to replace `producerL1OutputUsage + tensorUsage + cBUsagePeak` with `peakMemoryUsage` and let me know if it's what you would like to use.

rpavlovicTT: If you take a look carefully at https://github.com/tenstorrent/tt-mlir/issues/4143 -- it says it should include input tensors as well. So I propose to make things right in tt-metal first. Then wait for uplift of tt-metal and then make this change. This change does not bring much currently, except it updates unit tests, which will again be updated with new API. 

I think it would be better to write some tests on your branch that show input tensors usage is actually accounted for. Then we can move to testing in other repos.

gfengTT: Do you remember why we have outer vs. inner trace in the first place? Sounds like we can just keep one according to your plan. Also, do you have an example of memory footprint reduced by deallocating input tensor? I tried setting `deallocate_activation=true` but it didn't have an effect on memory usage in my case. We need to justify that using outer trace accounts for memory usage more accurately.

I tested some conv2d ops locally. Using outer trace does include input tensor size, but that's the same value that is calculated on the mlir side.

rpavlovicTT: I wasn't involved in designing the initial library for op constraints. So I can not give you a precise answer why we had outer/inner traces. 

As for example, I don't have one, but if I was in your place I would try with conv2d ops, which have sharded input tensor. That one should have different peak L1 usage depending on `deallocate_activation` flag.

gfengTT: I've added some tests to my branch https://github.com/tenstorrent/tt-metal/tree/gfeng/graph-trace-mem-usage-v2. Can you take a look?

The tests can be invoked by running `./build/test/ttnn/unit_tests_ttnn --gtest_filter="Conv2dOpIfTest*"`. You may toggle conv2dconfig between `std::nullopt` and `ttnn::operations::conv::conv2d::Conv2dConfig{.deallocate_activation=true}` (the commented line in my branch). For these tests they all resulted in the same `peak_memory_usage_per_core`.

rpavlovicTT: From the quick peak I had I can see you are making tensors in DRAM or SYS memory but I advised to try with L1. Then you should be able to observe differences. 

Can you please make a PR with sufficient test coverage and then I will take a real look?

gfengTT: I've added tests with l1 sharded input tensor (in https://github.com/tenstorrent/tt-metal/tree/gfeng/graph-trace-mem-usage-v2). And the tests now run with both null conv2dconfig and one with `deallocate_activation=true`. From these tests, it seems that the effects of `deallocate_activation` have been accounted for in the inner trace; using the outer trace only adds the size of input tensor to the usage, which can be computed by the input parameters alone.

Again, these tests are non-exhaustive. While we keep looking into this, according to our conversation at the meeting earlier today, shall we merge this PR now? Once this PR goes through, future updates should require changing the numerical values only. Thanks!

rpavlovicTT: Let's call this accordingly `peakL1MemorySize`

PR #4796, opened 2025-09-04
Enable BatchNorm decomposition only with fusion flag on

mvasiljevicTT:
### Problem description
BatchNorm is currently always decomposed after a `ConvolutionOp`, but fusion with convolution happens only if the `enable-fusing-conv2d-with-multiply-pattern` flag is enabled.
When the flag is disabled, BatchNorm still gets decomposed but not fused, leading to a performance drop.

### What's changed
BatchNorm is now decomposed only if the `enable-fusing-conv2d-with-multiply-pattern` flag is enabled.

### Checklist


jameszianxuTT:
There were some slight PCC drops (~0.01) in around 35 models running through tt-torch as a result of this [commit](https://github.com/tenstorrent/tt-mlir/commit/9da5a78bc733d7793f5acf0df763fc6c54e8e944) which enables batchnorm decomposition after ttir.convolution ops sourced from torch-xla sourced models. 

Testing with this patch resolves those PCC drops. cc @kmabeeTT

mtopalovicTT (approved):

Issue #4799, opened 2025-09-04
Fix slice op in shardy partitioner


tapspatel:
### Describe the bug

```
func.func @main(%arg0: tensor<4x128x16384xbf16> {sdy.sharding = #sdy.sharding<@mesh, [{}, {}, {"batch"}]>}) -> (tensor<4x128x8192xbf16>) {
  %1 = stablehlo.slice %arg0 [0:4, 0:128, 1:16384:2] : (tensor<4x128x16384xbf16>) -> tensor<4x128x8192xbf16>
  return %1 : tensor<4x128x8192xbf16>
}
```

Need to only rely on end indices, and not start indices when updating the slice attributes.

PR #4800, opened 2025-09-04
#4799: Fix bug when updating slice op attribute dictionary

tapspatel:
Issue: https://github.com/tenstorrent/tt-mlir/issues/4799

In shardy partitioner, we incorrectly use the output tensor shardings when updating the slice op attribute dictionary. We need to use the input sharding of this op, as well as update only the limit indices.

### DIFF
--- a/lib/Dialect/StableHLO/Transforms/UpdateGlobalToLocalShapes.cpp
+++ b/lib/Dialect/StableHLO/Transforms/UpdateGlobalToLocalShapes.cpp
@@ -65,39 +65,46 @@ static FailureOr<mlir::OperationState> createNewOperationState(
             return mlir::success();
           })
           .Case<mlir::stablehlo::SliceOp>([&](auto sliceOp) {
+            // 1. Get the sharding for each operand dimension.
+            llvm::ArrayRef<mlir::sdy::DimensionShardingAttr>
+                operandDimShardings =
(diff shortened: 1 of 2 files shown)


lib/Dialect/StableHLO/Transforms/UpdateGlobalToLocalShapes.cpp
```diff
@@ -65,39 +65,44 @@ static FailureOr<mlir::OperationState> createNewOperationState(
...
+            // 2. Copy the current start_indices and limit_indices attributes.
             llvm::SmallVector<int64_t> startIndices(sliceOp.getStartIndices());
             llvm::SmallVector<int64_t> limitIndices(sliceOp.getLimitIndices());
 
-            // Iterate through start and limit indices and update them based on
-            // the sharding annotation for that dimension.
-            for (uint32_t i = 0; i < tensorShardings.size(); i++) {
+            // 3. Iterate through start and limit indices and update them based
+            // on the sharding annotation for that dimension.
+            for (uint32_t i = 0; i < operandDimShardings.size(); i++) {
```
anuragsingh-tt: NIT: can use `llvm::enumerate` here instead

tapspatel: fixed

Issue #4803, opened 2025-09-04
Convert propagation strategy to conservative


tapspatel:
### Is your feature request related to a problem? Please describe

This is to prevent any unintended errors during CCL insertion.

PR #4804, opened 2025-09-04
#4803: Updated sharding propagation strategy to be conservative.

tapspatel:
Issue: https://github.com/tenstorrent/tt-mlir/issues/4803

Convert propagation strategy to be aggressive but disallow split axes and non-divisible sharding axes during propagation. This is critical during CCL insertion so that incorrect shardings don't end up getting inserted.


anuragsingh-tt (approved):

lib/Dialect/StableHLO/Pipelines/StableHLOPipelines.cpp
```diff
@@ -35,7 +36,12 @@ void createStableHLOPipeline(OpPassManager &pm,
       mlir::sdy::createApplyShardingConstraintsPass());
 
   // Propagate tensor shardings through the entire graph.
-  pm.addPass(mlir::sdy::createAggressivePropagationPass());
+  mlir::sdy::PropagationOptions propagationOptions;
+  mlir::sdy::PropagationStrategy propagationStrategy =
+      mlir::sdy::PropagationStrategy::Aggressive;
+  propagationOptions.conservativePropagation = true;
```
mmanzoorTT, line 52: `Aggressive strategy` and `conservativePropagation` seems opposite. Please explain their usage and add a comment to avoid any confusion in future.

tapspatel: fixed

PR #4811, opened 2025-09-05
Add permission needed for tt-forge's perf benchmark workflow

vvukomanTT:
### Ticket
[Issue](https://github.com/tenstorrent/tt-forge/issues/344)

### Problem description
Even though tt-mlir doesn't have the new performance benchmark regression check, the workflow call will fail without the added permission.

### What's changed
Added the needed permission.

### Checklist

### DIFF
--- a/.github/workflows/perf-benchmark.yml
+++ b/.github/workflows/perf-benchmark.yml
@@ -39,6 +39,7 @@ permissions:
   contents: read
   packages: write
   checks: write
+  id-token: write
 
 jobs:
   trigger:


.github/workflows/perf-benchmark.yml
```diff
@@ -39,6 +39,7 @@ permissions:
   contents: read
   packages: write
   checks: write
+  id-token: write
```
nsumrakTT, line 42: tt-mlir triggers FE workflows so this is not needed here.

vvukomanTT: You're right, we don't need the permissions set here then, thanks.

PR #4812, opened 2025-09-05
[L1InterleavedFallbackAnalysis] Improve perf by skipping L1 upgrades for matmul, linear and specific conv2d 

bmalesevicTT:
This PR improves the L1 interleaved fallback analysis by explicitly skipping the upgrades for inputs and output tensors of `ttnn::matmul` and `ttnn::linear` ops, as well as `ttnn::conv2d` ops that are convertible to the other two at lower levels (i.e., 1x1 kernel, stride=1, padding=0, groups=1). 
These operations are inefficient with L1 interleaved layout inputs/outputs and should not be upgraded by the analysis.

### What's changed
- Performance Impact:
  - Enabling L1IFA now provides visible performance improvements for models like YOLO_v8/v9/v10, where previously L1IFA had no effect or caused slight regression.
  - For other ops that previously benefitted from L1IFA, performance is either unchanged or further improved.

- No API changes.

### Checklist

### DIFF
--- a/lib/Dialect/TTNN/Utils/Utils.cpp
+++ b/lib/Dialect/TTNN/Utils/Utils.cpp
@@ -4,6 +4,7 @@
 
 #include "ttmlir/Dialect/TTNN/Utils/Utils.h"
 
+#include "ttmlir/Dialect/TTNN/IR/TTNNOps.h"
 #include "ttmlir/Dialect/TTNN/IR/TTNNOpsAttrs.h"
 #include "ttmlir/Dialect/TTNN/Types/Types.h"
 #include "ttmlir/Utils.h"
@@ -13,6 +14,7 @@
 #include "mlir/IR/Value.h"
 #include "llvm/Support/Casting.h"
 
+#include "llvm/ADT/ArrayRef.h"
 #include <optional>
 
 namespace mlir::tt::ttnn::utils {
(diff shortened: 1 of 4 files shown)


lib/Dialect/TTNN/Utils/Utils.cpp
```diff
@@ -307,4 +309,46 @@ bool producesTiledTensorLayout(Operation *op) {
...
 
+bool isConv2DConvertibleToMatMul(Operation *op) {
+  auto conv2dOp = dyn_cast<ttnn::Conv2dOp>(op);
+  if (!conv2dOp) {
+    return false;
+  }
+
+  // Get weight tensor to check kernel size
+  RankedTensorType weightType =
+      mlir::cast<RankedTensorType>(conv2dOp.getWeight().getType());
```
azecevic000: Type is already a `RankedTensorType`.

lib/Dialect/TTNN/Utils/Utils.cpp
```diff
@@ -11,6 +11,7 @@
 #include "mlir/Dialect/Quant/IR/QuantTypes.h"
 #include "mlir/IR/Location.h"
 #include "mlir/IR/Value.h"
+#include "llvm/ADT/ArrayRef.h"
```
rpavlovicTT: revert this

PR #4826, opened 2025-09-08
Encode producer-consumer dependency during ToLayoutOp lowering

bgrady-tt:
### Problem description
Device ToLayoutOps (those that have pre and post layouts that are both in device memspace) can be pipelined such that one ToLayoutOp produces the input for the next ToLayoutOp. 

```cpp
%1 = ttir.to_layout %0 ....
...
%3 = ttir.to_layout %1 .... // output of first to_layout op is pipelined directly into the second!
```

The lowering process for device ToLayoutOps rewrites the operation as a GenericOp; these GenericOp may have optional ViewLayoutOps applied to its output and input operands. These views are added to encode the required layout changes of the ToLayoutOp.

The addition of views to operands makes this lowering process _order dependent_; it isn't possible to process each to_layout concurrently, **or views may be applied multiple times to the same operand**!

### What's changed
- Added a check that causes rewrite to "fail" if the defining op producing current to_layout's input is itself a to_layout op operating from device->device memspace. 
- The above change forces the `GreedyPatternRewriter` to process only the first ToLayoutOp in a pipeline in its first iteration. _Each lowered Device ToLayoutOp unlocks the next in the pipeline to be rewritten, as the defining op will now appear as a regular GenericOp._

### Checklist


include/ttmlir/Dialect/TTCore/IR/Utils.h
```diff
@@ -105,6 +105,18 @@ ArrayRef<int64_t> getTensorTileShape(RankedTensorType tensorType);
...
+  return s == ttcore::MemorySpace::DeviceL1 ||
+         s == ttcore::MemorySpace::DeviceDRAM;
+}
+
+// returns true if the memory space is a system memory space
+inline bool isSystemMemspace(ttcore::MemorySpace s) {
+  return s == ttcore::MemorySpace::System ||
+         s == ttcore::MemorySpace::SystemMMIO;
+}
+
```
nsmithtt: These helpers already exist in `include/ttmlir/Dialect/TTCore/IR/TTCoreOpsTypes.h`, see `isDeviceMemorySpace` and `isSystemMemorySpace`.

lib/Dialect/TTIR/Transforms/LowerToLayout.cpp
```diff
@@ -59,6 +59,24 @@ class TTIRLowerToLayoutRewriter : public OpRewritePattern<ToLayoutOp> {
...
+    if (auto producer = op.getInput().getDefiningOp<ToLayoutOp>()) {
+      auto inputOperandMemspace =
+          producer.getOrCreateInputLayout().getMemorySpace();
+      auto outputOperandMemspace =
+          producer.getOrCreateOutputLayout().getMemorySpace();
+      if (ttcore::isDeviceMemspace(inputOperandMemspace) &&
+          ttcore::isDeviceMemspace(outputOperandMemspace)) {
+        return failure();
+      }
+    }
```
nsmithtt: Took a minute for this logic to click for me, but it makes sense now!  2 things, is it possible:
1. To move this into a utility func like `bool needsProducerLowering` or something named like that.
2. Not sure, but is it possible to move it up a stack frame?  Organizationally I think it'd be cleaner to handle rewriter early outs at the root level `matchAndRewrite`, but I'm not sure if logically it'll work out.

include/ttmlir/Dialect/TTCore/IR/Utils.h
bgrady-tt: noted, using the helpers in `TTCoreOpsTypes.h` now

lib/Dialect/TTIR/Transforms/LowerToLayout.cpp
bgrady-tt: Refactored this to "fail" early at beginning of `matchAndRewrite` if ToLayout input is detected to be a lowerable device->device memspace op.

PR #4828, opened 2025-09-08
[ttnn.jit] Runtime to dispatch fb through mlir runtime bindings.

vtangTT:
### Ticket
[Link to Github Issue](https://github.com/tenstorrent/tt-mlir/issues/4673)

### Problem description
ttnn.jit runtime

### What's changed
- expose runtime nanobinds by injecting PYTHONPATH in `env/activate`
- exposed `runtime/lib/ttnn/utils/utils.cpp` to nanobinds
  - need to include pybinds in order to resolve pybinded ttnn types in nanobinded helper functions
  - separate object TU for `utils.cpp` to isolate pybind deps
- example runtime flow in `test_unary_composite.py`

I think the main point of contention is whether we want pybinds included back into the project. It's header only right now, so maybe not that bad. Should be ok, assuming tt-metal eventually moves over to nanobind.

Also removed cmake targets `ttnn-jit` and `pykernel` that used to grep and install ttnn pip deps. As of recent main, these pip dependencies are already available in `env/activate` w/ the new docker image.

### Checklist


runtime/lib/ttnn/utils/utils.cpp
```diff
@@ -346,6 +349,27 @@ createRuntimeTensorFromTTNN(const ::ttnn::Tensor &tensor,
...
+
+  // Wrap the the device in the runtime device.
+  auto ttnnTraceCache = std::make_shared<::tt::runtime::ttnn::TraceCache>(
+      unsafeMeshDeviceSharedPtr);
+  auto traceCache = std::make_shared<::tt::runtime::TraceCache>(
+      std::static_pointer_cast<void>(ttnnTraceCache), DeviceRuntime::TTNN);
+
+  return Device(std::static_pointer_cast<void>(unsafeMeshDeviceSharedPtr),
+                traceCache, DeviceRuntime::TTNN);
+}
```
nsmithtt, line 507: We should get @jnie-TT's thoughts on this too.

runtime/python/CMakeLists.txt
```diff
@@ -3,13 +3,17 @@
 # SPDX-License-Identifier: Apache-2.0
 
 find_package(nanobind CONFIG REQUIRED)
+find_package(pybind11 CONFIG REQUIRED)
```
nsmithtt, line 6: Why do we need pybind?

runtime/python/runtime/utils.cpp
```diff
@@ -0,0 +1,61 @@
...
+void registerRuntimeUtilsBindings(nb::module_ &m) {
+  m.def(
+      "create_runtime_tensor_from_ttnn",
+      [](nb::object tensor_obj, bool retain) -> ::tt::runtime::Tensor {
+        py::handle tensor_pybind_obj(tensor_obj.ptr());
+        const ::ttnn::Tensor &tensor =
+            py::cast<const ::ttnn::Tensor &>(tensor_pybind_obj);
+        std::cout << "tensor: " << tensor.is_allocated() << " storage_type: "
+                  << static_cast<int>(tensor.storage_type()) << std::endl;
+        return ::tt::runtime::ttnn::utils::createRuntimeTensorFromTTNN(
```
nsmithtt, line 27: Oh cool this already existed!

runtime/python/CMakeLists.txt
vtangTT: Need pybind to be able to resolve the ttnn python object types. Nanobind can't resolve pybinded ttnn types (`nb::cast` wont work), so have to cast it with pybind to get the underlying c++ types.
`const ::ttnn::Tensor &tensor = py::cast<const ::ttnn::Tensor &>(tensor_pybind_obj);`

nsmithtt: Oh bummer, Ok!  Luckily I think they're planning to migrate to nano bind.

vtangTT: 🙏 🙏 🙏

runtime/lib/ttnn/utils/utils.cpp
jnie-TT: Could we pass in a shared_ptr of mesh device directly instead of a raw pointer? I see currently that we're getting the device from the tensor, would it make sense to pass in the opened device directly?

```
# instead of using the tensor's device
create_runtime_device_from_ttnn(tensor.device())

# use the opened device directly
device = ttnn.open_device(...)
create_runtime_device_from_ttnn(device)
```

I believe under the hood `ttnn.open_device` returns a shared pointer to the device.

runtime/python/CMakeLists.txt
```diff
@@ -44,4 +67,7 @@ if (CMAKE_CXX_COMPILER_ID MATCHES "Clang")
       -Wno-c++98-compat-extra-semi
       -Wno-covered-switch-default
   )
+  target_compile_options(_mlir_runtime_utils PRIVATE
+    -Wno-covered-switch-default
```
jnie-TT: I'm assuming this is for the pybind headers, could we wrap the headers with `#pragma clang diagnostic ignored` instead of ignoring for the entire object?

tools/pykernel/_src/dispatch_op.py
```diff
@@ -0,0 +1,73 @@
+# SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+
+import ttnn
+import torch
+
+from _ttmlir_runtime import runtime as runtime, binary as binary, utils as utils
```
jnie-TT: Nit: I think the `as` clauses are redundant here

runtime/include/tt/runtime/detail/ttnn/utils.h
```diff
@@ -9,6 +9,7 @@
 #include "tt/runtime/detail/ttnn/ttnn.h"
 #include "ttmlir/Target/Common/types_generated.h"
 #include "ttmlir/Target/TTNN/Target.h"
+#include <ttnn/events.hpp>
```
jnie-TT: Nit: quotation marks `#include "ttnn/events.hpp"`

runtime/python/CMakeLists.txt
jnie-TT: I love that we're building utils as a separate TU to isolate PYBIND details

tools/pykernel/_src/dispatch_op.py
vtangTT: 🤦

runtime/lib/ttnn/utils/utils.cpp
vtangTT: I think in terms of the api boundary, it's better to pass the tensor device. Reason being, the device itself would have to be passed all the way through the jit'ed op, which is weird to me as opposed to just passing the tensor. 

See `cosh_jit.py`

Right now, after jit decorating the function, we can just do the following, and `_run_binary` will pull the device pointer from the input tensor
```
output_tensor = cosh_ttnn(input_tensor)
```

Instead we would have to do something like this
```
output_tensor = cosh_ttnn(input_tensor, device)
```
Any thoughts on this?

runtime/python/CMakeLists.txt
vtangTT: wrapped it in the `#pragma` in `runtime/python/runtime/utils.cpp`

I also removed the above `target_compile_options` for `nanobind-static` since it's already covred in `nanobind_headers.h` and seemed redundant.. Unless I am missing something here..

runtime/include/tt/runtime/detail/ttnn/utils.h
vtangTT: fixed ty

tools/ttnn-jit/api.py
```diff
@@ -18,20 +18,21 @@
...
 from ttnn_jit._src.ttir_ast import TTIRCompiler
 from ttnn_jit._src.utils import _cleanup_source_code
+from ttnn_jit._src.dispatch_op import _run_binary
 
 
 def jit(
     backend: Literal["ttnn", "metal"] = "ttnn",
     perf: bool = False,
-    dump_flatbuffer: bool = False,
+    dump_flatbuffer: bool = True,
```
nsmithtt: What's the reason this should default to True?

tools/ttnn-jit/api.py
```diff
@@ -18,20 +18,21 @@
...
     debug: bool = False,
```
nsmithtt, line 28: Maybe we should have a separate mode like, `compile_only` which just compiles and returns the flatbuffer or the IR.

tools/ttnn-jit/api.py
```diff
@@ -72,6 +77,8 @@ def _wrapper(*args, **kwargs):
                     ttmetal_to_flatbuffer_file(
                         ir, os.path.join(out_dir, f.__name__ + ".ttm"), {}, []
                     )
+                # TODO: hook up metal runtime here
+                return ir
```
nsmithtt: If we had a `compile_only` flag we could return the ir here, otherwise we should `throw Unimplemented` or something if the user is expecting something to run through the metal backend path.

tools/ttnn-jit/api.py
```diff
@@ -80,14 +87,18 @@ def _wrapper(*args, **kwargs):
...
-                    out_dir = os.path.join("generated", "pykernels")
-                    os.makedirs(out_dir, exist_ok=True)
-                    ttnn_to_flatbuffer_file(
-                        ir, os.path.join(out_dir, f.__name__ + ".ttnn"), {}, []
-                    )
+                out_dir = os.path.join("generated", "pykernels")
+                os.makedirs(out_dir, exist_ok=True)
+                flatbuffer_bin = os.path.join(out_dir, f.__name__ + ".ttnn")
+
+                # TODO: remove once we can run flatbuffer from memory
```
nsmithtt: Can you file an issue for this?  I actually have this working on my branch `nsmith/pykernel_gen2`.

nsmithtt (approved):
Overall looks good! Few comments inline.

tools/ttnn-jit/api.py
vtangTT: Yea this makes more sense.

tools/ttnn-jit/api.py
vtangTT: https://github.com/tenstorrent/tt-mlir/issues/5055 and linked in comment

tools/ttnn-jit/api.py
vtangTT: Added a `compile_only` flag. I've removed the `dump_flatbuffer` flag. For `compile_only=False`, we'll dump the flatbuffer used in runtime if `debug=True`

tools/ttnn-jit/api.py
vtangTT: see below 🙏

tools/ttnn-jit/api.py
vtangTT: see below 🙏

runtime/lib/ttnn/utils/utils.cpp
jnie-TT: Hmm yeah I see fair point it does make the API cleaner, I guess the downside would be that the user needs to be aware to manage the device lifetime such that it outlives all the tensors that are on that device. I'm not super familiar with the future plans/flow of ttnn jit, but if that's a reasonable and maintainable flow then I'm ok with doing getting the device from the tensor.

runtime/lib/ttnn/utils/utils.cpp
```diff
@@ -488,6 +491,22 @@ createRuntimeTensorFromTTNN(const ::ttnn::Tensor &tensor,
                                /*data=*/nullptr, DeviceRuntime::TTNN);
 }
 
+::tt::runtime::Device
+createRuntimeDeviceFromTTNN(::ttnn::MeshDevice *meshDevice) {
+  // Create a non-owning shared_ptr to the provided MeshDevice with no-op
+  // deleter.
+  std::shared_ptr<::ttnn::MeshDevice> unsafeMeshDeviceSharedPtr(meshDevice,
+                                                                [](auto *) {});
```
jnie-TT: Nit: We have a utility API for this: `::tt::runtime::utils::unsafe_borrow_shared`

runtime/python/__init__.cpp
```diff
@@ -6,8 +6,13 @@
 #define RUNTIME_TEST_ENABLED
 #endif
 
+#if defined(TT_RUNTIME_ENABLE_TTNN) && TT_RUNTIME_ENABLE_TTNN == 1
+#define RUNTIME_UTILS_ENABLED
```
jnie-TT, line 10: Nit: Maybe we should enable this as long as runtime is enabled, I don't think this should be ttnn runtime specific, so that we could add utility APIs that might target metal runtime constructs as well. So essentially we can remove this ifdef check entirely, and have its scope be the same as runtime submodule.

vtangTT: Yep `TT_RUNTIME_ENABLE_TTNN` is only true when runtime is enabled (`TTMLIR_ENABLE_RUNTIME=ON`)

The reason I have this ifdef is because `TTRuntimeTTNNUtils` target doesn't get built unless `TTNN_RUNTIME_ENABLED` which is only set when `TT_RUNTIME_ENABLE_TTNN` is true. 

Checking for `TT_RUNTIME_ENABLE_TTNN` instead of `TTMLIR_ENABLE_RUNTIME` because the latter isn't passed as a preprocessor definition that can get picked up during runtime.

runtime/lib/ttnn/utils/utils.cpp
vtangTT: ty, forgot about it 😓 . Switched it over now.

Issue #4669, opened 2025-08-26
D2M: Explore enabling double buffering by default


jdesousa-TT:
We have a numBuffers attr on our shard descriptors. By default, we use single buffering right now, which could be hiding some easy perf gains. 

Theoretically, we should be able to set the numBuffers = 2 and see double buffering just work. The data movement shouldn't have to change at all, the DM kernels should just run ahead if there is space.

Explore this.

PR #4831, opened 2025-09-08
D2M: Add pipeline option for num-stream-buffers, enable variable buffers in allocator and FB generation

jdesousa-TT:
### Ticket
#4669 

### Problem description
We would like to enable double buffering on input streams. 

### What's changed
Add pipeline option for setting the number of stream buffers. For now this will be a global assignment, but we should probably eventually decide this on an op-by-op / buffer-by-buffer basis. 

### Checklist

### DIFF
--- a/include/ttmlir/Dialect/TTIR/Analysis/CBProducerConsumer.h
+++ b/include/ttmlir/Dialect/TTIR/Analysis/CBProducerConsumer.h
@@ -13,18 +13,22 @@
 
 namespace mlir::tt::ttir {
 
+// Default here is mostly for constructed testing purposes. i.e. Hand written
+// LIT tests that don't have ttir.yield or ttir.await ops.
 enum ThreadCBOrientation {
   Producer,
   Consumer,
   ProducerConsumer,
+  Default,
 };
 struct CBProducerConsumer {
   CBProducerConsumer(Operation *op);
 
   ThreadCBOrientation get(Value cb) const {
     auto match = threadCBOrientationMap.find(cb);
-    assert(match != threadCBOrientationMap.end() &&
-           "CB orientation mapping not found.");
+    if (match == threadCBOrientationMap.end()) {
+      return Default;
+    }
     return match->second;
   }
 
--- a/include/ttmlir/Dialect/TTIR/Analysis/CBProducerConsumer.h
+++ b/include/ttmlir/Dialect/TTIR/Analysis/CBProducerConsumer.h
@@ -1,0 +1,32 @@
+// SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#ifndef TTMLIR_DIALECT_TTIR_ANALYSIS_CBPRODUCERCONSUMER_H
+#define TTMLIR_DIALECT_TTIR_ANALYSIS_CBPRODUCERCONSUMER_H
+
+#include "ttmlir/Dialect/TTIR/IR/TTIROps.h"
+
+#include "mlir/IR/BuiltinAttributes.h"
+#include "mlir/IR/Value.h"
+#include "llvm/ADT/DenseMap.h"
+
+namespace mlir::tt::ttir {
+
+struct CBProducerConsumer {
+  CBProducerConsumer(Operation *op);
+
+  ttir::ThreadCBMapping get(mlir::StringRef threadRef, Value cb) const {
+    auto match = threadCBMappingMap.find(std::make_pair(threadRef, cb));
+    assert(match != threadCBMappingMap.end() &&
+           "CB producer consumer mapping not found.");
+    return match->second;
+  }
+
+  llvm::DenseMap<std::pair<mlir::StringRef, Value>, ttir::ThreadCBMapping>
+      threadCBMappingMap;
+};
+
+} // namespace mlir::tt::ttir
+
+#endif // TTMLIR_DIALECT_TTIR_ANALYSIS_CBPRODUCERCONSUMER_H
--- a/include/ttmlir/Dialect/TTIR/Analysis/CBProducerConsumer.h
+++ b/include/ttmlir/Dialect/TTIR/Analysis/CBProducerConsumer.h
@@ -13,18 +13,22 @@
 
 namespace mlir::tt::ttir {
 
+enum ThreadCBOrientation {
+  Producer,
+  Consumer,
+  ProducerConsumer,
+};
 struct CBProducerConsumer {
   CBProducerConsumer(Operation *op);
 
-  ttir::ThreadCBMapping get(mlir::StringRef threadRef, Value cb) const {
-    auto match = threadCBMappingMap.find(std::make_pair(threadRef, cb));
-    assert(match != threadCBMappingMap.end() &&
-           "CB producer consumer mapping not found.");
+  ThreadCBOrientation get(Value cb) const {
+    auto match = threadCBOrientationMap.find(cb);
+    assert(match != threadCBOrientationMap.end() &&
+           "CB orientation mapping not found.");
     return match->second;
   }
 
-  llvm::DenseMap<std::pair<mlir::StringRef, Value>, ttir::ThreadCBMapping>
-      threadCBMappingMap;
+  llvm::DenseMap<Value, ThreadCBOrientation> threadCBOrientationMap;
 };
 
 } // namespace mlir::tt::ttir
--- a/include/ttmlir/Dialect/TTIR/IR/TTIROpsEnums.td
+++ b/include/ttmlir/Dialect/TTIR/IR/TTIROpsEnums.td
@@ -44,17 +44,4 @@ def TTIR_ThreadType : I32EnumAttr<"ThreadType", "TTIR ThreadType", [
   let cppNamespace = "::mlir::tt::ttir";
 }
 
-def TTIR_ThreadCBMappingProducer : I32EnumAttrCase<"Producer", 0, "producer">;
-def TTIR_ThreadCBMappingConsumer : I32EnumAttrCase<"Consumer", 1, "consumer">;
-def TTIR_ThreadCBMappingProducerConsumer : I32EnumAttrCase<"ProducerConsumer", 2, "producer_consumer">;
-
-def TTIR_ThreadCBMapping : I32EnumAttr<"ThreadCBMapping", "TTIR ThreadCBMapping", [
-                                  TTIR_ThreadCBMappingProducer,
-                                  TTIR_ThreadCBMappingConsumer,
-                                  TTIR_ThreadCBMappingProducerConsumer
-                                ]> {
-  let genSpecializedAttr = 1;
-  let cppNamespace = "::mlir::tt::ttir";
-}
-
 #endif
--- a/include/ttmlir/Dialect/TTIR/IR/TTIROpsEnums.td
+++ b/include/ttmlir/Dialect/TTIR/IR/TTIROpsEnums.td
@@ -44,4 +44,17 @@ def TTIR_ThreadType : I32EnumAttr<"ThreadType", "TTIR ThreadType", [
   let cppNamespace = "::mlir::tt::ttir";
 }
 
+def TTIR_ThreadCBMappingProducer : I32EnumAttrCase<"Producer", 0, "producer">;
+def TTIR_ThreadCBMappingConsumer : I32EnumAttrCase<"Consumer", 1, "consumer">;
+def TTIR_ThreadCBMappingProducerConsumer : I32EnumAttrCase<"ProducerConsumer", 2, "producer_consumer">;
+
+def TTIR_ThreadCBMapping : I32EnumAttr<"ThreadCBMapping", "TTIR ThreadCBMapping", [
+                                  TTIR_ThreadCBMappingProducer,
+                                  TTIR_ThreadCBMappingConsumer,
+                                  TTIR_ThreadCBMappingProducerConsumer
+                                ]> {
+  let genSpecializedAttr = 1;
+  let cppNamespace = "::mlir::tt::ttir";
+}
+
 #endif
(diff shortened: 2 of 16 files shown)


lib/Conversion/TTIRToTTKernel/TTIRToTTKernel.cpp
```diff
@@ -779,12 +779,12 @@ class TTIRDMAWriteRewriter : public OpConversionPattern<ttir::DMAWriteOp> {
...
           // Dests are one less because the sender core is not included
           rewriter.create<ttkernel::NocAsyncWriteMulticastOp>(
-              op.getLoc(), srcL1Start, mcastAddr, transferSize, numDests,
+              op.getLoc(), dstL1Start, mcastAddr, transferSize, numDests,
               nullptr, nullptr, nullptr);
         } else {
           // If src != dst, we loopback mcast
           rewriter.create<ttkernel::NocAsyncWriteMulticastLoopbackSrcOp>(
-              op.getLoc(), srcL1Start, mcastAddr, transferSize, numDests,
+              op.getLoc(), dstL1Start, mcastAddr, transferSize, numDests,
```
nsmithtt: For loopback we should keep it as `src` right?

include/ttmlir/Dialect/TTIR/IR/TTIROpsEnums.td
```diff
@@ -44,4 +44,17 @@ def TTIR_ThreadType : I32EnumAttr<"ThreadType", "TTIR ThreadType", [
...
+
+def TTIR_ThreadCBMapping : I32EnumAttr<"ThreadCBMapping", "TTIR ThreadCBMapping", [
+                                  TTIR_ThreadCBMappingProducer,
+                                  TTIR_ThreadCBMappingConsumer,
+                                  TTIR_ThreadCBMappingProducerConsumer
+                                ]> {
+  let genSpecializedAttr = 1;
+  let cppNamespace = "::mlir::tt::ttir";
+}
+
```
nsmithtt: I think let's just make these enums defined as part of the analysis header.  The only reason to define them in tablegen is if we want them to be part of the IR, which I think we don't in this case.

include/ttmlir/Dialect/TTIR/Analysis/CBProducerConsumer.h
```diff
@@ -0,0 +1,32 @@
...
+
+  ttir::ThreadCBMapping get(mlir::StringRef threadRef, Value cb) const {
+    auto match = threadCBMappingMap.find(std::make_pair(threadRef, cb));
+    assert(match != threadCBMappingMap.end() &&
+           "CB producer consumer mapping not found.");
+    return match->second;
+  }
+
+  llvm::DenseMap<std::pair<mlir::StringRef, Value>, ttir::ThreadCBMapping>
+      threadCBMappingMap;
```
nsmithtt: `Value`'s should be globally unique, so there's probably no reason to also need the kernel name.

include/ttmlir/Dialect/TTIR/IR/TTIROpsEnums.td
```diff
@@ -44,4 +44,17 @@ def TTIR_ThreadType : I32EnumAttr<"ThreadType", "TTIR ThreadType", [
   let cppNamespace = "::mlir::tt::ttir";
 }
 
+def TTIR_ThreadCBMappingProducer : I32EnumAttrCase<"Producer", 0, "producer">;
+def TTIR_ThreadCBMappingConsumer : I32EnumAttrCase<"Consumer", 1, "consumer">;
+def TTIR_ThreadCBMappingProducerConsumer : I32EnumAttrCase<"ProducerConsumer", 2, "producer_consumer">;
+
+def TTIR_ThreadCBMapping : I32EnumAttr<"ThreadCBMapping", "TTIR ThreadCBMapping", [
```
nsmithtt: Maybe `ThreadCBOrientation` or Direction?

lib/Conversion/TTIRToTTKernel/TTIRToTTKernel.cpp
```diff
@@ -752,10 +767,29 @@ class TTIRDMAWriteRewriter : public OpConversionPattern<ttir::DMAWriteOp> {
...
+
+      auto srcCBMapping = cbProducerConsumer->get(
+          ttmlir::utils::getKernelName(op), op.getSrc());
+      Value srcL1Start;
+      Value dstL1Start;
+      if (srcCBMapping == ttir::ThreadCBMapping::Producer) {
+        srcL1Start = rewriter.create<ttkernel::GetWritePtrOp>(op.getLoc(),
+                                                              adaptor.getSrc());
+      } else if (srcCBMapping == ttir::ThreadCBMapping::Consumer ||
+                 srcCBMapping == ttir::ThreadCBMapping::ProducerConsumer) {
```
nsmithtt: Maybe just `else` and `assert` this?

lib/Dialect/TTIR/Analysis/CBProducerConsumer.cpp
```diff
@@ -0,0 +1,50 @@
...
+            ttir::ThreadCBMapping::Consumer);
+      }
+    }
+  }
+}
+
+CBProducerConsumer::CBProducerConsumer(Operation *op) {
+  op->walk([&](func::FuncOp funcOp) {
+    findThreadCBMappings(funcOp, threadCBMappingMap);
+  });
```
nsmithtt, line 24: Potential alternative:
```c++
  op->walk([&](Operation* op) {
    bool isConsumer = mlir::isa<ttir::YieldOp>(op);
    bool isProducer = mlir::isa<ttir::AwaitOp>(op);
    if (!isConsumer && !isProducer) {
      return;
    }
    auto enumVal = isConsumer ? Consumer : Producer;
    for (auto operand : op->getOperands()) {
      auto [iter, inserted] = threadCBMappingMap.insert({operand, enumVal});
      if (!inserted && *iter != enumVal) {
        *iter = ProducerConsumer;
      }
    }
  });
```

include/ttmlir/Dialect/TTIR/Analysis/CBProducerConsumer.h
bgrady-tt: Taking a closer look at `Value` [implementation](https://mlir.llvm.org/doxygen/Value_8h_source.html), it has a specialization for `DenseMapInfo` and a friend function `hash_value()`; this is pretty strong evidence it is enough to create robust hash key.

include/ttmlir/Dialect/TTIR/Transforms/Passes.td
```diff
@@ -439,6 +439,13 @@ def TTIRAllocate: Pass<"ttir-allocate", "::mlir::ModuleOp"> {
     ```
   }];
   let dependentDialects = ["::mlir::tt::ttcore::TTCoreDialect", "::mlir::memref::MemRefDialect"];
+
+  list<Option> options = [
+    Option<"numStreamBuffers",
+           "num-stream-buffers",
+           "unsigned", /*default=*/"1",
+           "Number of backing buffers to allocate per stream storage (>=1)">,
```
bgrady-tt: I think that maybe this should be termed a 'default' or 'max' number of backing buffers. We are definitely going to want to choose between single/double buffering at finer granularity in the future. It is fine right now to have some level of global control though.

lib/Conversion/TTIRToTTKernel/TTIRToTTKernel.cpp
```diff
@@ -704,8 +707,17 @@ class TTIRDMAReadRewriter : public OpConversionPattern<ttir::DMAReadOp> {
     auto srcNocAddr =
         buildNocAddress(rewriter, op.getLoc(), adaptor.getSrc(),
                         op.getSrcIndices(), chipDesc, op.getSrcMemorySpace());
-    auto dstL1Addr = buildL1Address<ttkernel::GetWritePtrOp>(
-        rewriter, op.getLoc(), adaptor.getDst(), op.getDstIndices());
+    auto dstCBMapping =
+        cbProducerConsumer->get(ttmlir::utils::getKernelName(op), op.getDst());
+    Value dstL1Addr;
+    if (dstCBMapping == ttir::ThreadCBMapping::Consumer) {
```
bgrady-tt: For a `DMAReadOp`, the only valid configuration is **remote memref --> local memref**. Any purely local DMA is lowered to a `DMAWriteOp` by convention. 

So it should be an _assert_ here that the dst CB is a  `ttir::ThreadCBMapping::Producer` i.e. pushing remote data _into_ the local CB via write ptr.

lib/Conversion/TTIRToTTKernel/TTIRToTTKernel.cpp
bgrady-tt: I think the only valid cases for `DMAWriteOp` are:

1. src = Consumer, dst = Producer
2. src & dst = ProducerConsumer

In both cases for `DMAWriteOp`, the data is _consumed_ from the src and _produced_ into the dst. The ambiguity of roles for `ProducerConsumer` is resolved by setting src=consumer and dst=producer.

bgrady-tt (changes requested):
Close to LGTM, but needs some minor cleanup in TTIRToTTKernel

The logic for read/write ptr selection in TTIRToTTKernel must always assume src CBs as consumer (read from) and dst CBs as producer (written to).

This might be confusing, but the read and write roles for DMAReadOp|DMAWriteOp refer _to the relationship to remote memrefs (read from remote, write to remote)_. DMAWriteOp unfortunately also represents local->local DMA which muddies this relationship.

include/ttmlir/Dialect/TTIR/Transforms/Passes.td
jdesousa-TT: I think this should be viewed as a global override right now since we don't have the logic to determine when this should be less than a max, or some other smart value. I agree that this should be changed to a "max" once we have some smarter analysis here.

lib/Dialect/TTIR/Analysis/CBProducerConsumer.cpp
jdesousa-TT: I like it, cleaner to not have to track the thread... didn't know that Value was super unique like that.

lib/Conversion/TTIRToTTKernel/TTIRToTTKernel.cpp
jdesousa-TT: Right, yes this pointer selection should only affect writes, ty

lib/Conversion/TTIRToTTKernel/TTIRToTTKernel.cpp
jdesousa-TT: I don't think this hold for mcast, which is the case that spawned this issue...

For example, the matmul reader thread will
1. Gather into cb A (producer) via a DMAReadOp, then
2. Multicast from cb A (not consuming, the compute will consume this) into cb A on a remote core (producer)

I think we can eliminate the case where the dst cb is a consumer, since we cannot consume via writing **to** a destination. 

But src is more tricky because of this case... thoughts?

include/ttmlir/Dialect/TTIR/Transforms/Passes.td
nsmithtt: Maybe the default should be 2?

include/ttmlir/Dialect/TTMetal/Pipelines/TTMetalPipelines.h
```diff
@@ -94,6 +94,13 @@ struct TTIRToTTMetalPipelineOptions
       *this, "insert-profiler-traces",
       llvm::cl::desc("Insert DeviceZone scopes around selected TTKernel ops"),
       llvm::cl::init(false)};
+
+  // Number of backing buffers to allocate per stream storage.
+  Option<unsigned> numStreamBuffers{
+      *this, "num-stream-buffers",
+      llvm::cl::desc(
+          "Number of backing buffers to allocate per stream storage (>=1)"),
+      llvm::cl::init(1)};
```
nsmithtt: Same here

lib/Conversion/TTIRToTTKernel/TTIRToTTKernel.cpp
```diff
@@ -900,29 +896,22 @@ class TTIRDMAWriteRewriter : public OpConversionPattern<ttir::DMAWriteOp> {
...
-      Value dstL1Start;
-      if (srcCBMapping == ttir::ThreadCBMapping::Producer) {
+      auto srcCBMapping = cbProducerConsumer->get(op.getSrc());
+      if (srcCBMapping == ttir::ThreadCBOrientation::Producer) {
         srcL1Start = rewriter.create<ttkernel::GetWritePtrOp>(op.getLoc(),
                                                               adaptor.getSrc());
-      } else if (srcCBMapping == ttir::ThreadCBMapping::Consumer ||
-                 srcCBMapping == ttir::ThreadCBMapping::ProducerConsumer) {
+      } else {
         srcL1Start = rewriter.create<ttkernel::GetReadPtrOp>(op.getLoc(),
```
bgrady-tt, line 907: I still think that if the dst CB is constrained to _always_ be a producer, then src must be a consumer to match.

bgrady-tt (approved):
Small nit that lowering for DMAWriteOp still "supports" both src and dst CBs accessing read ptr in the same thread. But approving because this is easy enough to followup on.

lib/Conversion/TTIRToTTKernel/TTIRToTTKernel.cpp
```diff
@@ -836,8 +840,15 @@ class TTIRDMAReadRewriter : public OpConversionPattern<ttir::DMAReadOp> {
     auto srcNocAddr =
         buildNocAddress(rewriter, op.getLoc(), adaptor.getSrc(),
                         op.getSrcIndices(), chipDesc, op.getSrcMemorySpace());
-    auto dstL1Addr = buildL1Address<ttkernel::GetWritePtrOp>(
+    auto dstCBMapping = cbProducerConsumer->get(op.getDst());
+    assert(dstCBMapping == ttir::ThreadCBOrientation::Producer ||
```
vwellsTT: nit: might be good to use Vlad's new TT_assertv

svuckovicTT (approved):

Issue #4820, opened 2025-09-08
[D2M] Cleanup MetalLayout Default Construction Logic in LowerToLayouts pass.


vwellsTT:
### Is your feature request related to a problem? Please describe

Two of my open PRs:
(https://github.com/tenstorrent/tt-mlir/pull/4753 and https://github.com/tenstorrent/tt-mlir/pull/4278) include changes to the worker grid; these both require workaround in the LowerToLayout pass as ToLayoutOp's are bounced into different forms.  We should have a canonicalized solution to this s.t. the worker grid is only at issue in the TTIRToTTIRGeneric Pass, and elsewhere we use other `MetalLayoutAttr::get` overloads s.t. we can be grid agnostic.

### Describe the solution you'd like

Cleanup the LowerToLayout.cpp + CompoundComponents in TTIROps.cpp to not involve default constructing a MetalLayoutAttr at all--when bouncing between layout forms, preserve information from the original/destination worker grid instead of re-checking working grid shape from IR.

vwellsTT:
closed by https://github.com/tenstorrent/tt-mlir/pull/5349

PR #4832, opened 2025-09-08
[D2M] Clean-up LowerToLayout Pass + MetalLayoutAttr Construction

vwellsTT:
### Ticket
[#4820
](https://github.com/tenstorrent/tt-mlir/issues/4820)

### Problem description
Some clunky default construction logic piled up for various changes to MetalLayoutAttr w.r.t. LowerToLayout pass.  We want a cleaner solution where:
1. the HW worker grid is not being used outside of initial tensor sizing in D2M
2. transitions from system tensors w/o MetalLayoutAttr to device tensors w/ MetalLayoutAttr and vice versa are handled gracefully

### What's changed
Refactored the bounce logic in LowerToLayout to use a dedicated subclass, which I find cleaner.  Avoid default constructing MetalLayoutAttrs for comparison--special case system <=> device transitions as needed.  Remove default contructing code from TTIROps.cpp and add special case logic there as well.

### Checklist
- [X] Existing tests provide coverage for changes

### DIFF
--- a/lib/Dialect/TTIR/IR/TTIROps.cpp
+++ b/lib/Dialect/TTIR/IR/TTIROps.cpp
@@ -4,6 +4,7 @@
 
 #include "ttmlir/Dialect/TTIR/IR/TTIROps.h"
 
+#include "ttmlir/Asserts.h"
 #include "ttmlir/Dialect/TTCore/IR/TTCoreOps.h"
 #include "ttmlir/Dialect/TTCore/IR/TTCoreOpsTypes.h"
 #include "ttmlir/Dialect/TTIR/IR/TTIRGenericRegionOps.h"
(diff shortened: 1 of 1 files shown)


nsmithtt:
This seems like a good change! I think it might conflict with @bgrady-tt's change here:
- https://github.com/tenstorrent/tt-mlir/pull/4826

Maybe let's let his land first since it's a relatively simpler change.

vwellsTT:
> This seems like a good change! I think it might conflict with @bgrady-tt's change here:
> 
> * [Encode producer-consumer dependency during ToLayoutOp lowering #4826](https://github.com/tenstorrent/tt-mlir/pull/4826)
> 
> Maybe let's let his land first since it's a relatively simpler change.

SGTM

lib/Dialect/TTIR/Transforms/LowerToLayout.cpp
```diff
@@ -17,6 +17,46 @@ namespace mlir::tt::ttir {
 #define GEN_PASS_DEF_TTIRLOWERTOLAYOUT
 #include "ttmlir/Dialect/TTIR/Transforms/Passes.h.inc"
 
+// Helper struct to encapsulate tensor info; this allows us to package
+// MetalLayoutAttr as optional gracefully.
+struct TensorInfo {
```
nsmithtt, line 23: Needs inline namespace wrapping this struct.  Seems like we might be missing them around all the other classes in this file too...

vwellsTT: Good point, I'll add 'em all

lib/Dialect/TTIR/IR/TTIROps.cpp
```diff
@@ -2658,47 +2658,64 @@ ::mlir::LogicalResult mlir::tt::ttir::ToLayoutOp::verify() {
...
   CompoundComponents components;
-  if (mlir::isa<mlir::RankedTensorType>(getInput().getType())) {
-    auto inputTensor = mlir::cast<mlir::RankedTensorType>(getInput().getType());
-    auto outputTensor =
-        mlir::cast<mlir::RankedTensorType>(getOutput().getType());
+
+  auto inputType = getInput().getType();
+  auto outputType = getOutput().getType();
+
+  if (mlir::isa<mlir::RankedTensorType>(inputType)) {
```
svuckovicTT: Consider denesting by processing the `else` branch first and then returning

vwellsTT: I see your point, but subjectively I think I prefer currently structure; I find
```
if(tensor) {
  components.a = foo();
  components.b = bar();
} else {
  components.a = memref_foo();
  components.b = memref_bar();
}
return components;
```
feels a little more symmetric to me vs
```
if (tensor) {
  components.a = foo();
  components.b = bar();
  return components;
}
components.a = memref_foo();
components.b = memref_bar();
return components;
```
but open to push back if someone feels strongly

svuckovicTT: The reason I suggested this was because I found it hard to match opening and closing braces while looking at the code - there's 2 levels of nested ifs.

Additionally, in this specific case, we'd be handling "errors" (or rather less expected and more easily solvable cases) first, which makes it easier for the reader of the code to "discard" the thoughts of those corner cases and focus on the meat of the logic, vs having to keep in mind that there's an else branch that handles another case.

Not a strong feeling here, as it is subjective :)

I like sharing this video that convinced me to avoid nesting 😄 
https://www.youtube.com/watch?v=CFRhGnuXG-4

nsmithtt: Normally I agree with you Sasha, it's almost always better to reduce nesting levels / prefer early outs.  That said, early outs signal to the reader that there's some kind of exception to the code path that's different from the rest of the main body.

This particular case though isn't an early out, it's conditionally constructing an object so here I agree with Vincent, it reads much more natural to me to have top level object `Foo` and in this condition populate the fields this way and in the other condition populate it that way and have a _single_ return because we're returning _this_ object.

vwellsTT: Yeah, that's what I meant;`if/else` sort of seems to keep both branches more "equal" for an initialization like this--agree with de-nesting in general, but here it feels less clear to me here.

Of course, in practice we don't exercise the mem-ref path @nsmithtt , so we could just change to an assert + eliminate the `else` branch here.  I assume the original thought was we might use this in the future in memref space, but honestly I'm not sure why we would, right?

vwellsTT: The only potential case I can think of is if we need for TTNN or something; it seems like in the regular d2m path we should decompose`ToLayout`s once in `LowerToLayout` in tensor space, and then never need to deal w/ them again in memref space

nsmithtt: Yeah that's true, this code only runs on tensors, probably safe to just assert that or maybe we should just move this logic out of the op and into the pass instead.  Either approach is ok with me.

This is actually a good thing to keep in mind for the D2M dialect work, we should for now just copy the to_layout op wholesale into the D2M and have it work as is, but eventually I want to add new ops:
- `d2m.to_device`
- `d2m.from_device`

And make `d2m.to_layout` an illegal op in lower parts of the pipeline.  Arguably we could do partial conversions:
- TTIRToD2M -> handles ttir ops to d2m generic conversion and freely uses ttir.to_layout, i.e. ttir.to_layout is the only legal op allowed to persist from ttir dialect.
- LowerToLayout -> runs sometime afterwards and finishes conversion of ttir.to_layout to `d2m.generic` + `d2m.to_device` + `d2m.from_device` and ttir dialect is no longer legal for any subsequent passes.

vwellsTT: Cool, makes sense.  I was originally wondering if we should make this just part of `LowerToLayout` like you said, but wondered if maybe there'd be some marginal value in being able to assert that after `LowerToLayout`, a given `ToLayout` op cannot be compound anymore, or something.  Lean towards leaving as-is for this PR (i.e. going assert + removing else branch route), then in d2m dialect refactor removing this from TTIR ToLayout moving to D2M ToLayout, then perhaps subsequently removing altogether when we add `to/from_device` you propose (which I'd also be in favor of, `ToLayout` feels pretty overloaded currently).
👍

nsmithtt: Sounds good

Issue #4836, opened 2025-09-08
[D2M] SFPU Ops must be reading/writing to DST


dlokeTT:
### Describe the bug

For comparison ops that are being treated as though they were fused, each sfpu op must be reading and writing its inputs/outputs from dst. Currently if we have two sfpu ops back to back such as sub_binary and eqz, eqz tries to get its operands from sub_binary

PR #4845, opened 2025-09-09
[D2M] Added dst stores and loads after chained SFPU tile ops

dlokeTT:
### Ticket
#4836 

### Problem description
Intermediate results in back to back tile ops in a linalg generic weren't storing their results to dst register

### What's changed
Added a check for chains of tile ops in TTIRInsertDstRegisterAccess pass that adds in stores/loads for dst

### Checklist


lib/Dialect/TTIR/Transforms/InsertDstRegisterAccess.cpp
```diff
@@ -140,6 +141,11 @@ struct TTIRInsertDstRegisterAccessRewriter final
 
     // 3. Generate data copy loops to/from dst and output cb.
     dataCopyGenerate(rewriter, loc, dst, copyInfo);
+
+    // 4. Insert dst access for tile operation chains
+    insertDstAccessForTileOpChain(rewriter, loc, region, dst,
```
nsmithtt: Let's call this func `insertDstRegisterAllocation`.

lib/Dialect/TTIR/Transforms/InsertDstRegisterAccess.cpp
```diff
@@ -338,6 +344,92 @@ struct TTIRInsertDstRegisterAccessRewriter final
...
   }
 
+  static void
+  insertDstAccessForTileOpChain(PatternRewriter &rewriter, Location loc,
+                                Region &region, Value dst,
+                                Operation *outermostInnerComputeLoop) {
+    // Collect all tile operations for processing
+    SmallVector<Operation *> tileOps;
+    region.walk([&](Operation *op) {
+      if (op->getName().getStringRef().starts_with("ttir.tile_")) {
```
nsmithtt: I think a better way to filter is based on OpTrait `if (op->hasTrait<TTIRGenericRegionComputeOpTrait>())`.

lib/Dialect/TTIR/Transforms/InsertDstRegisterAccess.cpp
```diff
@@ -338,6 +344,92 @@ struct TTIRInsertDstRegisterAccessRewriter final
...
+            }
+          }
+
+          // Now create a store after the current tile operation to dst
+          rewriter.setInsertionPointAfter(currentTileOp);
+          rewriter.create<affine::AffineStoreOp>(loc, tileResult, dst, loadMap,
+                                                 loadIndices);
+        }
+      }
+    }
```
nsmithtt: Instead of collecting all of the tileOps up front, maybe something like this could work?

```c++
region.walk([&](Operation *op) {
  if (!op->hasTrait<TTIRGenericRegionComputeOpTrait>()) {
    return;
  }

  // We will need a proper dst register allocation scheme to support the following assert.
  assert(op->hasOneUse() && "Currently we do not support multiple users in the same compute dst region");
  assert(op->getNumResults() == 1);
  
  // Early out if all users are already stores.
  if (llvm::all_of(op->getUsers(), [] (Operation* op) { return mlir::isa<memref::StoreOp>(op); })) {
    return;
  }

  // We might want some kind of stateful tracking structure within this routine that asserts that the dst store location we pick isn't clobbering the operands of other ops.  We definitely need to think a bit more about this.
  ...loadMap...loadIndices...

  rewriter.setInsertionPointAfter(op);
  // Bounce through dst.
  auto dstStore = rewriter.create<affine::AffineStoreOp>(loc, op->getResult(0), dst, loadMap, loadIndices);
  auto dstLoad = rewriter.create<affine::AffineLoadOp>(loc, dst, loadMap, loadIndices);
  rewriter.replaceAllUsesWithIf(op->getResult(0), dstLoad->getResult(0), [] (Operation* op) { return !mlir::isa<memref::StoreOp>(op); });
});
```

Another thing to be mindful of is that by default `walk` will do a post-order traversal of the region and is controlled by its first template argument `template<WalkOrder Order = WalkOrder::PostOrder,`.  We might want to choose a PreOrder traversal if it's easier for register tracking.

nsmithtt:
OK hang on, was just revisiting the `collectDstAccesses` existing code as part of this pass and I think we should actually move the analysis bit of the dst register here instead. So something like:

```c++
// Maps each TTIRGenericRegionComputeOpTrait operation result to a dest register offset.
using DstRegisterAllocation = DenseMap<Operation*, int64_t>;

// Return both the copy nest info and dst allocation info
static std::pair<DenseMap<Operation *, CopyInfo>, DstRegisterAllocation>
// Rename the func to be more descriptive
collectCopyNestAndDstAccesses(...) {
  ...
  DstRegisterAllocation dstRegisterAllocation;
  region.walk([&](OperandLoadRegisterOpInterface op) {
    // Collect loads to this op.
    ...

          // Collect stores from this op.
      for (auto *user : op->getUsers()) {
        if (auto potentialStore = mlir::dyn_cast<affine::AffineStoreOp>(user);
            notDstMemspace(potentialStore)) {
          ...
        } else {
          assert(op->hasTrait<TTIRGenericRegionComputeOpTrait>());
          assert(op->hasOneUse() && "Currently we do not support multiple users in the same compute dst region");
          assert(op->getNumResults() == 1);
          assert(!dstRegisterAllocation.contains(op));
          dstRegisterAllocation[op] = dstRegisterAllocationState.allocate();
        }
      }
  });
  return {loopNests, dstRegisterAllocation};
}
```

We currently do not have a way to deallocate dst registers, so this may need more work to make the `DstRegisterAllocationState` more sophisticated.  But then we can just pass `dstRegisterAllocation` into `insertDstRegisterAllocation` where the `loadIndices` are looked up from this map.

lib/Dialect/TTIR/Transforms/InsertDstRegisterAccess.cpp
```diff
@@ -249,9 +263,18 @@ struct TTIRInsertDstRegisterAccessRewriter final
           dstRegisterAllocationState.allocate();
           dstRegisterAllocationState.setStoreToDst();
         }
+        // If the op is a compute op, we need to allocate a dst register for it
```
nsmithtt: Maybe this comment should read "If the user isn't a store, it must be another compute consumer and we need to allocate a dest register intermediate for it".

lib/Dialect/TTIR/Transforms/InsertDstRegisterAccess.cpp
```diff
@@ -249,9 +263,18 @@ struct TTIRInsertDstRegisterAccessRewriter final
...
+        else {
+          assert(op->hasTrait<TTIRGenericRegionComputeOpTrait>());
```
nsmithtt: I may have overlooked this in my suggested code, I think this should actually read:
```c++
assert(user->hasTrait<TTIRGenericRegionComputeOpTrait>());
```

lib/Dialect/TTIR/Transforms/InsertDstRegisterAccess.cpp
```diff
@@ -487,6 +510,61 @@ struct TTIRInsertDstRegisterAccessRewriter final
...
+      Operation *outermostInnerComputeLoop = nullptr) {
+    auto dstType = dyn_cast<MemRefType>(dst.getType());
+    if (!dstType) {
+      return;
+    }
+    unsigned dstRank = dstType.getRank();
+
+    // Iterate directly through dst register allocation entries
+    for (const auto &[op, dstIndex] : dstRegisterAllocation) {
+      if (op->getNumResults() == 0) {
```
nsmithtt: I think we should assert this, did you run into any cases where this is true?

lib/Dialect/TTIR/Transforms/InsertDstRegisterAccess.cpp
```diff
@@ -487,6 +510,61 @@ struct TTIRInsertDstRegisterAccessRewriter final
...
+          Value inductionVar = affineFor.getInductionVar();
+          storeIndices.push_back(inductionVar);
+        }
+      } else {
+        storeIndices.push_back(rewriter.create<arith::ConstantIndexOp>(loc, 0));
+      }
+
+      while (storeIndices.size() < dstRank) {
+        storeIndices.push_back(rewriter.create<arith::ConstantIndexOp>(loc, 0));
+      }
```
nsmithtt, line 564: Is there a reason we have to use `outermostInnerComputeLoop`?  My thinking of the indices calculation should look like:
```c++
SmallVector<Value> storeIndices;
storeIndices.push_back(rewriter.create<arith::ConstantIndexOp>(loc, dstIndex));
while (storeIndices.size() < dstRank) {
  storeIndices.push_back(rewriter.create<arith::ConstantIndexOp>(loc, 0));
}
```

Lmk what you think.

lib/Dialect/TTIR/Transforms/InsertDstRegisterAccess.cpp
```diff
@@ -487,6 +510,61 @@ struct TTIRInsertDstRegisterAccessRewriter final
...
+          loc, result, dst, storeMap, storeIndices);
+
+      auto loadedResult = rewriter.create<affine::AffineLoadOp>(
+          loc, dst, storeMap, storeIndices);
+
+      for (auto &use : result.getUses()) {
+        if (use.getOwner() != storeOp) {
+          use.set(loadedResult.getResult());
+        }
+      }
```
nsmithtt: We should always use the `rewriter` to do IR modification:

```c++
rewriter.replaceAllUsesWithIf(op->getResult(0), loadedResult->getResult(0), [] (Operation* op) { return !mlir::isa<affine::StoreOp>(op); });
```

lib/Dialect/TTIR/Transforms/InsertDstRegisterAccess.cpp
dlokeTT: I initially thought I had but removing it and running tests, they all pass so I can remove it

lib/Dialect/TTIR/Transforms/InsertDstRegisterAccess.cpp
dlokeTT: Yea, this slipped thru the cracks while I was double checking. We already did the calculation for `dstIndex` in `collectDstAccesses`

lib/Dialect/TTIR/Transforms/InsertDstRegisterAccess.cpp
```diff
@@ -240,18 +254,41 @@ struct TTIRInsertDstRegisterAccessRewriter final
...
+                 "Multiple stores from last op to dst not supported");
+
+          auto dstRegInPlace = false;
+          if (auto attr =
+                  op->getAttrOfType<mlir::BoolAttr>("dst_reg_in_place")) {
+            dstRegInPlace = attr.getValue();
+          }
+          int64_t dstIndex;
+          if (dstRegInPlace) {
+            dstIndex = dstRegisterAllocationState.getCurrDstIndex();
```
nsmithtt, line 277: I think for `getCurrDstIndex` to be true we need to assert that `op->getNumOperands() == 2` otherwise for binary ops `getCurrDstIndex` will be pointing to the second index, but I think it's more likely that it'd store back to the first index.

Perhaps we should track currDstIndex as `minDstIndex` inside of this walk statement where we pick the min dst index from the `potentialLoad` condition above.

Then I think we might need to assert the implication if storesInPlace then minDstIndex must be set: `assert(!op.storesInPlace() || minDstIndex != -1)` (if we initialize minDstIndex to -1).

lib/Dialect/TTIR/Transforms/InsertDstRegisterAccess.cpp
```diff
@@ -487,6 +528,51 @@ struct TTIRInsertDstRegisterAccessRewriter final
     }
   }
 
+  // Rewrite stores to use dst register based on allocation map
+  static void insertDstRegisterAllocation(
+      PatternRewriter &rewriter, Location loc, Value dst,
+      const DstRegisterAllocation &dstRegisterAllocation,
+      Operation *outermostInnerComputeLoop) {
```
nsmithtt: I think this is no longer used.

lib/Dialect/TTIR/Transforms/InsertDstRegisterAccess.cpp
```diff
@@ -507,6 +593,7 @@ struct TTIRInsertDstRegisterAccessRewriter final
   }
 
   bool useTileMatmul;
+  static constexpr bool explain = false;
```
nsmithtt: Perhaps unintentional.

lib/Dialect/TTIR/Transforms/InsertDstRegisterAccess.cpp
dlokeTT: Yeah, this makes sense for binary ops. My understanding, was that based on this llk and metal change: https://github.com/tenstorrent/tt-metal/pull/27513, all binary ops had to have their output dest tile specified so this case is only run into for unary ops

lib/Dialect/TTIR/Transforms/InsertDstRegisterAccess.cpp
dlokeTT: This is still used for allocating the intermediate destination registers

lib/Dialect/TTIR/Transforms/InsertDstRegisterAccess.cpp
```diff
@@ -64,6 +66,7 @@ struct TTIRInsertDstRegisterAccessRewriter final
 
     void setStoreToDst() { storedToDst = true; }
     bool didStoreToDst() { return storedToDst; }
+    int64_t getCurrDstIndex() { return nextDstIndex; }
```
vwellsTT: nit: I'd probably prefer naming this `getNextDstIndex`

lib/Dialect/TTIR/Transforms/InsertDstRegisterAccess.cpp
```diff
@@ -208,14 +216,21 @@ struct TTIRInsertDstRegisterAccessRewriter final
...
   // access. Returns a map of loop nest -> copy info, which contains a list of
   // loads and stores to copy into hoisted loop nests.
-  static DenseMap<Operation *, CopyInfo>
+
+  // Maps each TTIRGenericRegionComputeOpTrait operation result to a dest
+  // register offset.
+  using DstRegisterAllocation = DenseMap<Operation *, int64_t>;
+
+  // Return both the copy nest info and dst allocation info
+  static std::tuple<DenseMap<Operation *, CopyInfo>, DstRegisterAllocation>
```
vwellsTT: nit: I think a pair is more appropriate than a tuple if you have two elements like this.
And personally, I don't like either, I prefer POD struct w/ clear variable names, to make it less likely you access fields out of order, etc.

lib/Dialect/TTIR/Transforms/InsertDstRegisterAccess.cpp
```diff
@@ -487,6 +525,51 @@ struct TTIRInsertDstRegisterAccessRewriter final
...
+
+      // Store the result of this operation to dst register
+      rewriter.setInsertionPoint(op);
+
+      SmallVector<Value> storeIndices;
+
+      storeIndices.push_back(
+          rewriter.create<arith::ConstantIndexOp>(loc, dstIndex));
+      while (storeIndices.size() < dstRank) {
+        storeIndices.push_back(rewriter.create<arith::ConstantIndexOp>(loc, 0));
```
vwellsTT, line 563: this logic isn't quite clear to me, can you add a comment maybe?

lib/Dialect/TTIR/Transforms/InsertDstRegisterAccess.cpp
```diff
@@ -487,6 +525,51 @@ struct TTIRInsertDstRegisterAccessRewriter final
...
+  // Rewrite stores to use dst register based on allocation map
+  static void insertDstRegisterAllocation(
+      PatternRewriter &rewriter, Location loc, Value dst,
+      const DstRegisterAllocation &dstRegisterAllocation,
+      Operation *outermostInnerComputeLoop) {
+    auto dstType = dyn_cast<MemRefType>(dst.getType());
+    if (!dstType) {
+      return;
+    }
+    unsigned dstRank = dstType.getRank();
```
vwellsTT: nit: imo this sort of primitive ought to be `const`, though we don't generally adhere to `const`-ing much tbf

lib/Dialect/TTIR/Transforms/InsertDstRegisterAccess.cpp
```diff
@@ -149,6 +152,11 @@ struct TTIRInsertDstRegisterAccessRewriter final
 
     // 3. Generate data copy loops to/from dst and output cb.
     dataCopyGenerate(rewriter, loc, dst, copyInfo);
+
+    // 4. Rewrite stores to use dst register based on allocation
```
vwellsTT: nit: here and throughout PR, pls end comments with `.` or other punctation, per our style guide: https://docs.tenstorrent.com/tt-mlir/coding-guidelines.html#comments

test/ttmlir/Dialect/TTIR/Transforms/insert_dst_register_access_block.mlir
```diff
@@ -112,7 +112,7 @@ module {
       // CHECK-NEXT: affine.for %[[WB_J:.*]] = 0 to 2 {
 
       // Check writeback: load from dst, store to l1
-      // CHECK: %[[FINAL_VAL:.*]] = affine.load %[[DST]][0, %[[WB_I]], %[[WB_J]]] : memref<1x3x2x!ttcore.tile<32x32, f32>, #dst>
+      // CHECK: %[[FINAL_VAL:.*]] = affine.load %[[DST]][1, %[[WB_I]], %[[WB_J]]] : memref<1x3x2x!ttcore.tile<32x32, f32>, #dst>
```
vwellsTT: I'm also not sure I get why this + the other existing test have index changed like this, can you explain?

lib/Dialect/TTIR/Transforms/InsertDstRegisterAccess.cpp
nsmithtt: At the very least we should assert! :) I think I had a typo above, probably should `assert(op->getNumOperands() == 1)`

lib/Dialect/TTIR/Transforms/InsertDstRegisterAccess.cpp
nsmithtt: Maybe my grep is failing me, which line in this func is it referenced?

dlokeTT: 157

dlokeTT: oh wait you're referencing the variable, yes that's not used. ill remove it

lib/Dialect/TTIR/Transforms/InsertDstRegisterAccess.cpp
dlokeTT: This is actually referencing the currentDstIndex. We've been calling this `nextDstIndex` because so far we always allocate the next spot and then increment but for ops that store the destination register in place, we don't want to allocate so we just use the last index.

nsmithtt: I guess to Vincent's point, wouldn't CurrDstIndex be `nextDstIndex-1`?

dlokeTT: Yeah, never mind. Got this confused. It should be -1

test/ttmlir/Dialect/TTIR/Transforms/insert_dst_register_access_block.mlir
dlokeTT: Fixing this, it should still be 0. Tile Matmul is a special case where it loads accumulator into dest tile and we want to write it back in place to that same tile.

lib/Dialect/TTIR/Transforms/InsertDstRegisterAccess.cpp
dlokeTT: added

lib/Dialect/TTIR/Transforms/InsertDstRegisterAccess.cpp
```diff
@@ -208,14 +215,27 @@ struct TTIRInsertDstRegisterAccessRewriter final
   // register allocation for loads and just assumes that stores get exclusive
   // access. Returns a map of loop nest -> copy info, which contains a list of
   // loads and stores to copy into hoisted loop nests.
-  static DenseMap<Operation *, CopyInfo>
+
+  // Maps each TTIRGenericRegionComputeOpTrait operation result to a dest
+  // register offset.
+  using DstRegisterAllocation = DenseMap<Operation *, int64_t>;
+
+  // Struct to hold the results of dst access collection
```
vwellsTT: nit: .

vwellsTT (approved):
Thanks for addressing feedback, 1 last nit otherwise I think LGTM.
You should wait for @nsmithtt 's sign-off too though, since he has better understanding of logical changes than me.

lib/Dialect/TTIR/Transforms/InsertDstRegisterAccess.cpp
```diff
@@ -240,18 +260,44 @@ struct TTIRInsertDstRegisterAccessRewriter final
       for (auto *user : op->getUsers()) {
         if (auto potentialStore = mlir::dyn_cast<affine::AffineStoreOp>(user);
             notDstMemspace(potentialStore)) {
+
           assert(!dstRegisterAllocationState.didStoreToDst() &&
-                 "Multiple stores to dst not supported");
+                 "Multiple stores from last op to dst not supported");
+
+          auto dstRegInPlace = op.getDstRegInPlace();
+          int64_t dstIndex;
```
nsmithtt: Let's initialize to `-1` or something invalid that'd blow up in case this code gets refactored in the future.

nsmithtt (approved):
Awesome looks good!  One minor nit left inline

PR #4897, opened 2025-09-11
Add `unary_bcast_init` and `unary_bcast_tile` ops

navdeepkk-polymagelabs:
### Ticket
Link to Github Issue

### Problem description
`unary_bcast_init` and `unary_bcast` ops are missing from `tt-mlir`.

### What's changed
Add broadcast a single tile from a CB. The tile is read from the CB and the broadcast result is stored in the provided DST register index. Broadcast along `rows` or `columns` or along both is supported.

### Checklist

### DIFF
--- a/include/ttmlir/Dialect/TTKernel/IR/TTKernelOpsEnums.td
+++ b/include/ttmlir/Dialect/TTKernel/IR/TTKernelOpsEnums.td
@@ -105,6 +105,22 @@ def TTKernel_ReduceType : I32EnumAttr<"ReduceType", "TTKernel Reduce Types",
   let cppNamespace = "::mlir::tt::ttkernel";
 }
 
+def TTKernel_BcastTypeNone : I32EnumAttrCase<"None", 0, "bcast_type_none">;
+def TTKernel_BcastTypeCol : I32EnumAttrCase<"Col", 1, "bcast_type_col">;
+def TTKernel_BcastTypeRow : I32EnumAttrCase<"Row", 2, "bcast_type_row">;
+def TTKernel_BcastTypeScalar : I32EnumAttrCase<"Scalar", 3, "bcast_type_scalar">;
+
+def TTKernel_BcastType : I32EnumAttr<"BcastType", "TTKernel broadcast types",
+                         [
+                           TTKernel_BcastTypeNone,
+                           TTKernel_BcastTypeCol,
+                           TTKernel_BcastTypeRow,
+                           TTKernel_BcastTypeScalar
+                         ]> {
+  let genSpecializedAttr = 0;
+  let cppNamespace = "::mlir::tt::ttkernel";
+}
+
 def TTKernel_ReduceDimRow : I32EnumAttrCase<"Row", 0, "reduce_dim_row">;
 def TTKernel_ReduceDimCol : I32EnumAttrCase<"Col", 1, "reduce_dim_col">;
 def TTKernel_ReduceDimScalar : I32EnumAttrCase<"Scalar", 2, "reduce_dim_scalar">;
--- a/lib/Conversion/TTKernelToEmitC/TTKernelToEmitC.cpp
+++ b/lib/Conversion/TTKernelToEmitC/TTKernelToEmitC.cpp
@@ -277,6 +277,19 @@ public:
     return {reduceType, reduceDim};
   }
 
+  StringRef getBroadcastType(ttkernel::BcastType bcastType) const {
+    switch (bcastType) {
+    case ttkernel::BcastType::Row:
+      return "BroadcastType::ROW";
+    case ttkernel::BcastType::Col:
+      return "BroadcastType::COL";
+    case ttkernel::BcastType::Scalar:
+      return "BroadcastType::SCALAR";
+    default:
+      return "BroadcastType::NONE";
+    }
+  }
+
   ArrayAttr getTemplateArgs(Builder &builder, SourceOp op) const {
     if constexpr (std::is_same_v<SourceOp, ttkernel::ReduceInitOp> ||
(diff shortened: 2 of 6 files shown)


include/ttmlir/Dialect/TTKernel/IR/TTKernelOpsEnums.td
```diff
@@ -105,6 +105,22 @@ def TTKernel_ReduceType : I32EnumAttr<"ReduceType", "TTKernel Reduce Types",
   let cppNamespace = "::mlir::tt::ttkernel";
 }
 
+def TTKernel_BcastDimRow : I32EnumAttrCase<"Row", 0, "bcast_dim_row">;
+def TTKernel_BcastDimCol : I32EnumAttrCase<"Col", 1, "bcast_dim_col">;
+def TTKernel_BcastDimScalar : I32EnumAttrCase<"Scalar", 2, "bcast_dim_scalar">;
+def TTKernel_BcastDimNone : I32EnumAttrCase<"None", 3, "bcast_dim_none">;
+
+def TTKernel_BcastDim : I32EnumAttr<"BcastDim", "TTKernel broadcast Dimensions",
```
wenbinlyuTT: I'm not sure which is better:
- Follow our existing naming convention of `ReduceDim` and use `BcastDim`, or
- Follow the compute API definition and call these `BcastType`

It's unfortunate that the compute API decide to use `ReduceDim` & `BcastType` to mean the direction of the op inside the tile. Personally I prefer `BcastType`, because broadcast dimension already have some other meanings (e.g. `tensor<2x1x2x1> + tensor<2x2x2x2>` has bcast dim 1 & 3.)

include/ttmlir/Dialect/TTKernel/IR/TTKernelOpsEnums.td
```diff
@@ -105,6 +105,22 @@ def TTKernel_ReduceType : I32EnumAttr<"ReduceType", "TTKernel Reduce Types",
   let cppNamespace = "::mlir::tt::ttkernel";
 }
 
+def TTKernel_BcastDimRow : I32EnumAttrCase<"Row", 0, "bcast_dim_row">;
+def TTKernel_BcastDimCol : I32EnumAttrCase<"Col", 1, "bcast_dim_col">;
+def TTKernel_BcastDimScalar : I32EnumAttrCase<"Scalar", 2, "bcast_dim_scalar">;
+def TTKernel_BcastDimNone : I32EnumAttrCase<"None", 3, "bcast_dim_none">;
```
wenbinlyuTT: It would be nice if the order of these followed the metalium enum definition:
```
enum BroadcastType {
  NONE   = 0x0,
  COL    = 0x1,
  ROW    = 0x2,
  SCALAR = 0x3,
};
```

lib/Conversion/TTKernelToEmitC/TTKernelToEmitC.cpp
```diff
@@ -293,10 +307,20 @@ class TTKernelToEmitCOpaqueRewriter : public OpConversionPattern<SourceOp> {
...
             getReduceTypeAndDim<ttkernel::ReduceTileOp>(reduceOp);
       }
-      template_args.push_back(
-          emitc::OpaqueAttr::get(op.getContext(), reduceType));
-      template_args.push_back(
-          emitc::OpaqueAttr::get(op.getContext(), reduceDim));
+      return ArrayAttr::get(op.getContext(), template_args);
+    } else if (std::is_same_v<SourceOp, ttkernel::UnaryBcastInitOp>) {
+      auto bcastInitOp = mlir::cast<ttkernel::UnaryBcastInitOp>(op);
+      SmallVector<Attribute, 3> template_args;
```
wenbinlyuTT: I think 1 is enough?

lib/Conversion/TTKernelToEmitC/TTKernelToEmitC.cpp
```diff
@@ -293,10 +307,20 @@ class TTKernelToEmitCOpaqueRewriter : public OpConversionPattern<SourceOp> {
...
+      template_args.push_back(emitc::OpaqueAttr::get(
+          op.getContext(),
+          getBroadcastDim<ttkernel::UnaryBcastInitOp>(bcastInitOp)));
+      return ArrayAttr::get(op.getContext(), template_args);
+    } else if (std::is_same_v<SourceOp, ttkernel::UnaryBcastTileOp>) {
```
wenbinlyuTT: I think these two branches can be unified, something like `template_args.push_back(emitc::OpaqueAttr::get(op.getContext(), getBcastDim(op)));` should be able to handle both.

navdeepkk-polymagelabs:
Thanks for the comments. Will address all of these and push.

lib/Conversion/TTKernelToEmitC/TTKernelToEmitC.cpp
```diff
@@ -293,10 +307,20 @@ class TTKernelToEmitCOpaqueRewriter : public OpConversionPattern<SourceOp> {
         std::tie(reduceType, reduceDim) =
             getReduceTypeAndDim<ttkernel::ReduceTileOp>(reduceOp);
       }
-      template_args.push_back(
-          emitc::OpaqueAttr::get(op.getContext(), reduceType));
-      template_args.push_back(
-          emitc::OpaqueAttr::get(op.getContext(), reduceDim));
```
navdeepkk-polymagelabs: This was accidentally removed and is most like the cause of the CI failures.

include/ttmlir/Dialect/TTKernel/IR/TTKernelOpsEnums.td
navdeepkk-polymagelabs: Using `type` against `dim` here.

PR #4902, opened 2025-09-11
Allow creation of empty borrowed tensors

LPanosTT:
### Ticket
N/A

### Problem description
vLLM internally makes some empty tensors during initialization. Our runtime asserts that the buffers we wish to create tensors for are non null. TTNN can create a borrowed tensor with null data if the volume is zero. 

### What's changed
- Removed the assertion that causes this to fail and let ttnn make the tensor.
### Checklist

### DIFF
--- a/runtime/lib/ttnn/runtime.cpp
+++ b/runtime/lib/ttnn/runtime.cpp
@@ -195,6 +195,12 @@ createBorrowedHostTensor(void *data, const std::vector<std::uint32_t> &shape,
                          const std::vector<std::uint32_t> &stride,
                          std::uint32_t itemsize,
                          ::tt::target::DataType dataType) {
+  LOG_ASSERT(
+      data != nullptr ||
+          (shape.size() == 0 ||
+           std::accumulate(shape.begin(), shape.end(), 1,
+                           std::multiplies<std::uint32_t>()) == 0),
+      "Cannot create borrowed tensor with null data unless the volume is 0.");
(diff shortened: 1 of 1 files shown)


runtime/lib/ttnn/runtime.cpp
```diff
@@ -195,7 +195,6 @@ createBorrowedHostTensor(void *data, const std::vector<std::uint32_t> &shape,
                          const std::vector<std::uint32_t> &stride,
                          std::uint32_t itemsize,
                          ::tt::target::DataType dataType) {
-  LOG_ASSERT(data != nullptr, "Cannot create borrowed tensor with null data");
```
AleksKnezevic, line 198: It can only be a nullptr in case of a size 0 tensor, right? We should assert that.

LPanosTT: That is checked within ttnn:

```
TT_FATAL(
        buffer.size() == volume, "Current buffer size is {} different from shape volume {}", buffer.size(), volume);
```

Should we check here as well then?

AleksKnezevic: I think so. I would assert that we either have data or there shouldn't be any data

jojo-771771:
I am interested in working on this issue. I have experience with MLIR and can deliver quality work. Please assign it to me!

jojo-771771:
I am interested in working on this issue. I have experience with MLIR and can deliver quality work. Please assign it to me!

Issue #4912, opened 2025-09-12
Add support for gpt-oss scatter op


ddilbazTT:
Please add support for 
```
  %1 = "ttir.scatter"(%arg0, %arg1, %arg2, %0) <{index_vector_dim = 2 : i32, indices_are_sorted = false, input_batching_dims = array<i32>, inserted_window_dims = array<i32: 0, 1>, scatter_dims_to_operand_dims = array<i32: 0, 1>, scatter_indices_batching_dims = array<i32>, unique_indices = false, update_window_dims = array<i32>}> : (tensor<71x32xbf16>, tensor<71x4x2xi64>, tensor<71x4xbf16>, tensor<71x32xbf16>) -> tensor<71x32xbf16>
```

AleksKnezevic:
@ddilbazTT can this be closed?

ddilbazTT:
Yes, it can be closed.

Issue #4531, opened 2025-08-18
Support stablehlo.scatter operation


ndrakulicTT:
Here is a fwd/bwd graph generated with torch_xla -> tt-xla path for torch.nn.Embedding:
```
// FWD
module @SyncTensorsGraph.8 attributes {mhlo.cross_program_prefetches = [], mhlo.input_output_alias = [], mhlo.is_dynamic = false, mhlo.use_auto_spmd_partitioning = false, ttcore.meshes = #ttcore.meshes<[<"mesh" = 1x1>]>} {
  func.func @main(%arg0: tensor<2x5xi64> {ttcore.shard_status = #ttcore.shard_status<unsharded>}, %arg1: tensor<1000x32xf32> {ttcore.shard_status = #ttcore.shard_status<unsharded>}) -> (tensor<2x5x32xf32> {ttcore.shard_status = #ttcore.shard_status<unsharded>}) {
    %0 = ttir.empty() : tensor<10xi64>
    %1 = "ttir.reshape"(%arg0, %0) <{shape = [10 : i32]}> : (tensor<2x5xi64>, tensor<10xi64>) -> tensor<10xi64>
    %2 = ttir.empty() : tensor<10xui32>
    %3 = "ttir.typecast"(%1, %2) <{conservative_folding = false}> : (tensor<10xi64>, tensor<10xui32>) -> tensor<10xui32>
    %4 = ttir.empty() : tensor<10x32xf32>
    %5 = "ttir.gather"(%arg1, %3, %4) <{collapsed_slice_dims = array<i64: 0>, index_vector_dim = 1 : si64, indices_are_sorted = false, offset_dims = array<i64: 1>, operand_batching_dims = array<i64>, slice_sizes = array<i64: 1, 32>, start_index_map = array<i64: 0>, start_indices_batching_dims = array<i64>}> : (tensor<1000x32xf32>, tensor<10xui32>, tensor<10x32xf32>) -> tensor<10x32xf32>
    %6 = ttir.empty() : tensor<2x5x32xf32>
    %7 = "ttir.reshape"(%5, %6) <{shape = [2 : i32, 5 : i32, 32 : i32]}> : (tensor<10x32xf32>, tensor<2x5x32xf32>) -> tensor<2x5x32xf32>
    return %7 : tensor<2x5x32xf32>
  }
}
// BWD
[...]

PR #4913, opened 2025-09-12
Uplift ttnn scatter op and add coverage

ddilbazTT:
### Ticket
#4531 
#4912 

### Problem description
Current stablehlo scatter op is not lowered into ttnn scatter op as this was not implemented in tt-metal. It is a binary op and it provides coverage for a subset of scatter operations.

### What's changed
- Uplifted tt-metal scatter op implementation.
- Removed binary op references for scatter op and lowered ttir scatter to ttnn scatter.
- Wrote conversion pattern for scatter op which supports element-wise scatter and multi-dimensional scatter. 
- Added scatter decomposition workaround in ttnn due to the following restriction (third_party/tt-metal/src/tt-metal/ttnn/cpp/ttnn/operations/data_movement/scatter/scatter.cpp)
```
index_shape[dim] < to_layout_int32_scatter_axis_max_length (256)
```
- Added scatter workaround pass in ttnn due to the following restriction (third_party/tt-metal/src/tt-metal/ttnn/cpp/ttnn/operations/data_movement/scatter/scatter.cpp)
```
    TT_FATAL(
        !(input_dtype == DataType::FLOAT32 && input_layout == Layout::TILE),
        "Scatter doesn't work for fp32 tiled tensors yet (see to_layout issue #) - input tensor is {} {}.",
        input_dtype,
        input_layout);
    TT_FATAL(
        !(source_dtype == DataType::FLOAT32 && source_layout == Layout::TILE),
        "Scatter doesn't work for fp32 tiled tensors yet (see to_layout issue #) - source tensor is {} {}.",
        source_dtype,
        source_layout);
```

Conversion logic:
- scatter_dims_to_operand_dims: 
    - if len == 1 --> simple scatter operation. dim = scatter_dims_to_operand_dims[0]. We are expecting dim == 0.
    - if len > 1 --> multi-dimensional scatter operation. Need to flatten all tensors to 1D and use custom logic to calculate linear indices.
 - If simple scatter:
     - indices need to match updates shape --> reshape by adding 1d singleton dimensions at the beginning if needed
     - indices are broadcasted based on `update_window_dims` if needed
- If multi-dimensional scatter
    - inputs and updates are flattened into 1d
    - 1d indices are created using strides

### Checklist
- [ ] TODO: Add scatter op constraints again since they are removed in this PR.
- [ ] TODO: Add window-based scatter support.
- [ ] TODO: Simplify multi-dimensional scatter to use multiple scatters instead of flattened scatter.

### DIFF
--- a/include/ttmlir/Dialect/TTNN/IR/TTNNOps.td
+++ b/include/ttmlir/Dialect/TTNN/IR/TTNNOps.td
@@ -2068,15 +2068,13 @@ def TTNN_ScatterOp : TTNN_Op<"scatter", [DeclareOpInterfaceMethods<TTNN_OpModelI
         - `index` (ttnn.Tensor): Indices where values will be written to.
         - `source` (ttnn.Tensor): The values to scatter into the input tensor.
         - `dim` (int32_t): The dimension along which to scatter.
-        - `cq_id` i32: The command queue to copy the tensor with. Must be 0 or 1.
     }];
 
     let arguments = (ins AnyRankedTensor:$input,
                          AnyRankedTensor:$index,
                          AnyRankedTensor:$source,
                          I32Attr:$dim,
-                         OptionalAttr<TTNN_MemoryConfigAttr>:$memory_config,
-                         DefaultValuedAttr<UI32Attr, "0">:$cq_id);
+                         OptionalAttr<TTNN_MemoryConfigAttr>:$memory_config);
 
     let results = (outs AnyRankedTensor:$result);
 
--- a/test/ttmlir/Dialect/TTNN/simple_scatter.mlir
+++ b/test/ttmlir/Dialect/TTNN/simple_scatter.mlir
@@ -6,7 +6,7 @@ module attributes {} {
     %0 = ttir.empty() : tensor<1x3x320x320xf32>
     %1 = ttir.empty() : tensor<1x1xi32>
     %2 = "ttir.scatter"(%arg0, %1, %arg1, %0) <{index_vector_dim = 1 : i32, indices_are_sorted = false, input_batching_dims = array<i32>, inserted_window_dims = array<i32: 0>, scatter_dims_to_operand_dims = array<i32: 0>, scatter_indices_batching_dims = array<i32>, unique_indices = false, update_window_dims = array<i32: 1, 2, 3>}> : (tensor<1x3x320x320xf32>, tensor<1x1xi32>, tensor<1x3x32x32xf32>, tensor<1x3x320x320xf32>) -> tensor<1x3x320x320xf32>
-    // CHECK: %{{[0-9]+}} = "ttnn.scatter"({{.*}}) <{cq_id = 0 : ui32, dim = 0 : i32}> : (tensor<1x3x320x320xf32, {{.*}}>, tensor<1x3x32x32xsi32, {{.*}}>, tensor<1x3x32x32xf32, {{.*}}>) -> tensor<1x3x320x320xf32, {{.*}}>
+    // CHECK: %{{[0-9]+}} = "ttnn.scatter"({{.*}}) <{dim = 0 : i32}> : (tensor<1x3x320x320xf32, {{.*}}>, tensor<1x3x32x32xsi32, {{.*}}>, tensor<1x3x32x32xf32, {{.*}}>) -> tensor<1x3x320x320xf32, {{.*}}>
     return %2 : tensor<1x3x320x320xf32>
     // CHECK: return %{{[0-9]+}} : tensor<1x3x320x320xf32, {{.*}}>
   }
@@ -15,8 +15,8 @@ module attributes {} {
   func.func @scatter_1(%arg0: tensor<71x32xbf16>, %arg1: tensor<71x4x2xi64>, %arg2: tensor<71x4xbf16>) -> tensor<71x32xbf16> {
     %0 = ttir.empty() : tensor<71x32xbf16>
     %1 = "ttir.scatter"(%arg0, %arg1, %arg2, %0) <{index_vector_dim = 2 : i32, indices_are_sorted = false, input_batching_dims = array<i32>, inserted_window_dims = array<i32: 0, 1>, scatter_dims_to_operand_dims = array<i32: 0, 1>, scatter_indices_batching_dims = array<i32>, unique_indices = false, update_window_dims = array<i32>}> : (tensor<71x32xbf16>, tensor<71x4x2xi64>, tensor<71x4xbf16>, tensor<71x32xbf16>) -> tensor<71x32xbf16>
-    // CHECK: %{{[0-9]+}} = "ttnn.scatter"({{.*}}) <{cq_id = 0 : ui32, dim = 0 : i32}> : (tensor<2272xbf16, {{.*}}>, tensor<256xsi32, {{.*}}>, tensor<256xbf16, {{.*}}>) -> tensor<2272xbf16, {{.*}}>
-    // CHECK: %{{[0-9]+}} = "ttnn.scatter"({{.*}}) <{cq_id = 0 : ui32, dim = 0 : i32}> : (tensor<2272xbf16, {{.*}}>, tensor<28xsi32, {{.*}}>, tensor<28xbf16, {{.*}}>) -> tensor<2272xbf16, {{.*}}>
+    // CHECK: %{{[0-9]+}} = "ttnn.scatter"({{.*}}) <{dim = 0 : i32}> : (tensor<2272xbf16, {{.*}}>, tensor<256xsi32, {{.*}}>, tensor<256xbf16, {{.*}}>) -> tensor<2272xbf16, {{.*}}>
+    // CHECK: %{{[0-9]+}} = "ttnn.scatter"({{.*}}) <{dim = 0 : i32}> : (tensor<2272xbf16, {{.*}}>, tensor<28xsi32, {{.*}}>, tensor<28xbf16, {{.*}}>) -> tensor<2272xbf16, {{.*}}>
     return %1 : tensor<71x32xbf16>
     // CHECK: return %{{[0-9]+}} : tensor<71x32xbf16, {{.*}}>
   }
--- a/test/ttmlir/Silicon/TTNN/n150/data_movement/scatter/scatter.mlir
+++ b/test/ttmlir/Silicon/TTNN/n150/data_movement/scatter/scatter.mlir
@@ -13,7 +13,7 @@ func.func @scatter_simple_1(%arg0: tensor<1x3x320x320xf32>, %arg1: tensor<1x1xi3
   // CHECK: "ttnn.reshape"({{.*}}) <{shape = [1 : i32, 1 : i32, 1 : i32, 1 : i32]}>
   // CHECK-SAME: (tensor<1x1xsi32, {{.*}}>) -> tensor<1x1x1x1xsi32, {{.*}}>
   // CHECK: "ttnn.repeat"({{.*}}) <{repeat_dims = #ttnn.shape<1x3x32x32>}>
-  // CHECK: "ttnn.scatter"({{.*}}) <{cq_id = 0 : ui32, dim = 0 : i32}>
+  // CHECK: "ttnn.scatter"({{.*}}) <{dim = 0 : i32}>
   // CHECK-SAME: (tensor<1x3x320x320xf32, {{.*}}>, tensor<1x3x32x32xsi32, {{.*}}>, tensor<1x3x32x32xf32, {{.*}}>) -> tensor<1x3x320x320xf32, {{.*}}>
   return %1 : tensor<1x3x320x320xf32>
 }
@@ -23,7 +23,7 @@ func.func @scatter_simple_2(%arg0: tensor<32x32xi32>, %arg1: tensor<16x1xi32>, %
   %1 = "ttir.scatter"(%arg0, %arg1, %arg2, %0) <{index_vector_dim = 1 : i32, indices_are_sorted = false, input_batching_dims = array<i32>, inserted_window_dims = array<i32: 0>, scatter_dims_to_operand_dims = array<i32: 0>, scatter_indices_batching_dims = array<i32>, unique_indices = false, update_window_dims = array<i32: 1>}> : (tensor<32x32xi32>, tensor<16x1xi32>, tensor<16x32xi32>, tensor<32x32xi32>) -> tensor<32x32xi32>
   // CHECK-LABEL: func.func @scatter_simple_2
   // CHECK: "ttnn.repeat"({{.*}}) <{repeat_dims = #ttnn.shape<1x32>}>
-  // CHECK: "ttnn.scatter"({{.*}}) <{cq_id = 0 : ui32, dim = 0 : i32}>
+  // CHECK: "ttnn.scatter"({{.*}}) <{dim = 0 : i32}>
   // CHECK-SAME: (tensor<32x32xsi32, {{.*}}>, tensor<16x32xsi32, {{.*}}>, tensor<16x32xsi32, {{.*}}>) -> tensor<32x32xsi32, {{.*}}>
   return %1 : tensor<32x32xi32>
 }
@@ -47,8 +47,8 @@ func.func @scatter_simple_3(%arg0: tensor<71x32xbf16>, %arg1: tensor<71x4x2xi64>
   // flatten updates:
   // CHECK: "ttnn.reshape"({{.*}}) <{shape = [284 : i32]}>
   // Scatter is broken into chunks where index_shape[dim] < 256.
-  // CHECK: "ttnn.scatter"({{.*}}) <{cq_id = 0 : ui32, dim = 0 : i32}> : (tensor<2272xbf16, {{.*}}>, tensor<256xsi32, {{.*}}>, tensor<256xbf16, {{.*}}>) -> tensor<2272xbf16, {{.*}}>
-  // CHECK: "ttnn.scatter"({{.*}}) <{cq_id = 0 : ui32, dim = 0 : i32}> : (tensor<2272xbf16, {{.*}}>, tensor<28xsi32, {{.*}}>, tensor<28xbf16, {{.*}}>) -> tensor<2272xbf16, {{.*}}>
+  // CHECK: "ttnn.scatter"({{.*}}) <{dim = 0 : i32}> : (tensor<2272xbf16, {{.*}}>, tensor<256xsi32, {{.*}}>, tensor<256xbf16, {{.*}}>) -> tensor<2272xbf16, {{.*}}>
+  // CHECK: "ttnn.scatter"({{.*}}) <{dim = 0 : i32}> : (tensor<2272xbf16, {{.*}}>, tensor<28xsi32, {{.*}}>, tensor<28xbf16, {{.*}}>) -> tensor<2272xbf16, {{.*}}>
   // reshape flattened output to expected output shape
   // CHECK: "ttnn.reshape"({{.*}}) <{shape = [71 : i32, 32 : i32]}>
   return %1 : tensor<71x32xbf16>
@@ -61,7 +61,7 @@ func.func @scatter_simple_4(%arg0: tensor<1000x32xf32>, %arg1: tensor<10x1xi64>,
   // CHECK: "ttnn.repeat"({{.*}}) <{repeat_dims = #ttnn.shape<1x32>}>
   // CHECK: "ttnn.to_layout"({{.*}}) <{layout = #ttnn.layout<row_major>}>
   // CHECK: "ttnn.to_layout"({{.*}}) <{layout = #ttnn.layout<row_major>}>
-  // CHECK: "ttnn.scatter"({{.*}}) <{cq_id = 0 : ui32, dim = 0 : i32}> : (tensor<1000x32xf32, {{.*}}>, tensor<10x32xsi32, {{.*}}>, tensor<10x32xf32, {{.*}}>) -> tensor<1000x32xf32, {{.*}}>
+  // CHECK: "ttnn.scatter"({{.*}}) <{dim = 0 : i32}> : (tensor<1000x32xf32, {{.*}}>, tensor<10x32xsi32, {{.*}}>, tensor<10x32xf32, {{.*}}>) -> tensor<1000x32xf32, {{.*}}>
   // CHECK: "ttnn.to_layout"({{.*}}) <{layout = #ttnn.layout<tile>}>
   %1 = "ttir.scatter"(%arg0, %arg1, %arg2, %0) <{index_vector_dim = 1 : i32, indices_are_sorted = false, input_batching_dims = array<i32>, inserted_window_dims = array<i32: 0>, scatter_dims_to_operand_dims = array<i32: 0>, scatter_indices_batching_dims = array<i32>, unique_indices = false, update_window_dims = array<i32: 1>}> : (tensor<1000x32xf32>, tensor<10x1xi64>, tensor<10x32xf32>, tensor<1000x32xf32>) -> tensor<1000x32xf32>
   return %1 : tensor<1000x32xf32>


ddilbazTT:
Note: the tt-metal branch I use in this PR is not from the main branch since tt-mlir is not up to date with tt-metal. I made a branch out of the tt-metal version we were using and cherry-picked Jerry's commit. https://github.com/tenstorrent/tt-metal/compare/main...ddilbaz/scatter https://github.com/tenstorrent/tt-metal/pull/26747 

I will turn this draft PR into an official PR when we uplift tt-metal.

Edit: I turned this into an official PR to have more reviews on it.

ddilbazTT:
Note 2: I tested this PR in tt-torch with some PyTorch scatter unit tests to make sure output values make sense. I used gpt heavily to come up with the conversion functions and I focused on the scatter cases from issues. Regardless, the tests I wrote passed with PCC: 1.0, ATOL: 0.0. I am attaching the output dump and the tt-torch tests I wrote (You can also view: https://github.com/tenstorrent/tt-torch/tree/ddilbaz/scatter_tests) 

```
# SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
#
import torch
from torch import nn
import pytest

from tt_torch.tools.verify import verify_module

# Examples from https://docs.pytorch.org/docs/stable/generated/torch.Tensor.scatter_.html

# Example 1:
# src = torch.arange(1, 11).reshape((2, 5))
# src
# tensor([[ 1,  2,  3,  4,  5],
#         [ 6,  7,  8,  9, 10]])
# index = torch.tensor([[0, 1, 2, 0]])
# torch.zeros(3, 5, dtype=src.dtype).scatter_(0, index, src)
# tensor([[1, 0, 0, 4, 0],
#         [0, 2, 0, 0, 0],
#         [0, 0, 3, 0, 0]])
def test_scatter_dim0():
    class ScatterDim0(nn.Module):
        def forward(self, x, index, src):
            return torch.scatter(x, dim=0, index=index, src=src)

    src = torch.arange(1, 11, dtype=torch.bfloat16).reshape((2, 5))
    index = torch.tensor([[0, 1, 2, 0, 1]], dtype=torch.int64)  # shape (1,5)
    x = torch.zeros((3, 5), dtype=torch.bfloat16)
    verify_module(
        ScatterDim0(),
        inputs=[x, index, src],
        required_atol=1,
    )


# Example 2:
# src = torch.arange(1, 11).reshape((2, 5))
# src
# tensor([[ 1,  2,  3,  4,  5],
#         [ 6,  7,  8,  9, 10]])
# index = torch.tensor([[0, 1, 2], [0, 1, 4]])
# torch.zeros(3, 5, dtype=src.dtype).scatter_(1, index, src)
# tensor([[1, 2, 3, 0, 0],
#         [6, 7, 0, 0, 8],
#         [0, 0, 0, 0, 0]])
def test_scatter_dim1():
    class ScatterDim1(nn.Module):
        def forward(self, x, index, src):
            return torch.scatter(x, dim=1, index=index, src=src)

    src = torch.arange(1, 11, dtype=torch.bfloat16).reshape((2, 5))
    index = torch.tensor([[0, 1, 2], [0, 1, 4]], dtype=torch.int64)  # shape (2,5)
    x = torch.zeros((3, 5), dtype=torch.bfloat16)
    verify_module(
        ScatterDim1(),
        inputs=[x, index, src],
        required_atol=1,
    )


# Example 3:
# index = torch.tensor([[0, 1]])
# value = 2
# torch.zeros(3, 5).scatter_(0, index, value)
# tensor([[2., 0., 0., 0., 0.],
#         [0., 2., 0., 0., 0.],
#         [0., 0., 0., 0., 0.]])
def test_scatter_scalar_value_fixed():
    class ScatterScalarValueFixed(nn.Module):
        def forward(self, x, index, value):
            src = torch.full_like(index, value, dtype=torch.bfloat16)
            return torch.scatter(x, dim=0, index=index, src=src)

    index = torch.tensor([[0, 1]], dtype=torch.int64)  # shape (1,2)
    value = torch.tensor(2, dtype=torch.bfloat16)
    x = torch.zeros((3, 5), dtype=torch.bfloat16)
    verify_module(
        ScatterScalarValueFixed(),
        inputs=[x, index, value],
        required_atol=1,
    )

```
[scatter_test.txt](https://github.com/user-attachments/files/22288141/scatter_test.txt)

ddilbazTT:
Note 3: I talked about this with @sdjukicTT offline and I'll re-iterate. Reduction mode of scatter ops is out-of-scope for this PR. To add support for reduction modes, we could:
- lower stablehlo scatter to ttir scatter and other ops --> this is very messy as most reduction modes end up being cumulative scatter ops. So we would end up creating a cascade of scatter + add/ multiply/ etc. ops.
- wait for reduction mode support in tt-metal: I believe there is already a tt-metal issue for that I'll try to find it.

lib/Conversion/TTIRToTTNN/TTIRToTTNN.cpp
```diff
@@ -1573,38 +1573,436 @@ class RandOpConversionPattern : public OpConversionPattern<ttir::RandOp> {
...
+     * Operations:
+     *   1. Extract indices:
+     *      - ttnn.slice_static: begins=[0,0,0], ends=[1,5,1] ->
+     * tensor<1x5x1xi32>
+     *   2. Expand input tensor (rank 2 -> 3):
+     *      - ttnn.reshape: <205x768xf32> -> tensor<1x2050x768xf32>
+     *   3. Broadcast indices to match updates:
+     *      - ttnn.repeat: repeat_dims=<1x1x768> -> tensor<1x5x768xi32>
+     *   4. Final scatter:
+     *      - ttnn.scatter: dim=0, all tensors rank 3
```
jbbieniekTT: 0, not 1 or 2?

jbbieniekTT:
> Note 3: I talked about this with @sdjukicTT offline and I'll re-iterate. Reduction mode of scatter ops is out-of-scope for this PR. To add support for reduction modes, we could:
> 
> * lower stablehlo scatter to ttir scatter and other ops --> this is very messy as most reduction modes end up being cumulative scatter ops. So we would end up creating a cascade of scatter + add/ multiply/ etc. ops.
> * wait for reduction mode support in tt-metal: I believe there is already a tt-metal issue for that I'll try to find it.

https://github.com/tenstorrent/tt-metal/pull/26757 - reduction is done in the reader kernel, and a benchmark between reader-kernel-row-major-only and compute-tile-based must be carried out - there's also the fact that reader kernel doesn't handle `bfloat16`s accurately

lib/Conversion/TTIRToTTNN/TTIRToTTNN.cpp
ddilbazTT: Oh this is just the example in the comment block. It's not always 0, it's `scatter_dims_to_operand_dims[-1]` which is 0 in this case since `scatter_dims_to_operand_dims=[0]`

ddilbazTT: It's `scatter_dims_to_operand_dims[-1]` when it's a single-dimension scatter btw, not multi-dimensional.

ddilbazTT:
Note 4: I added a tile layout workaround for f32 and tested that it works through tt-torch. Now, we change f32 source or input tensor layout from tile to row.

lib/Conversion/TTNNToEmitC/TTNNToEmitC.cpp
```diff
@@ -2145,6 +2145,35 @@ class ReduceScatterOpConversionPattern
...
+      mlir::tt::ttnn::ScatterOp>::TTNNToEmitCBaseOpConversionPattern;
+
+  LogicalResult
+  matchAndRewrite(mlir::tt::ttnn::ScatterOp srcOp,
+                  mlir::tt::ttnn::ScatterOp::Adaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    ttnn_to_emitc::EmitCTTNNEmitter<mlir::tt::ttnn::ScatterOp> emitter(
+        srcOp, adaptor, rewriter);
+    llvm::SmallVector<mlir::Attribute> args{
+        emitter.emit(std::nullopt), // queue_id
```
jbbieniekTT: remove `queue_id`?

lib/Conversion/TTIRToTTNN/TTIRToTTNN.cpp
```diff
@@ -1573,38 +1573,436 @@ class RandOpConversionPattern : public OpConversionPattern<ttir::RandOp> {
...
+        mlir::cast<RankedTensorType>(scatterIndices.getType());
+    auto scatterIndicesShape = scatterIndicesType.getShape();
+
+    // Slice to extract indices for the target dimension
+    llvm::SmallVector<int32_t> begins(scatterIndicesShape.size(), 0);
+    llvm::SmallVector<int32_t> ends(scatterIndicesShape.begin(),
+                                    scatterIndicesShape.end());
+    llvm::SmallVector<int32_t> steps(scatterIndicesShape.size(), 1);
+
+    begins[indexVectorDim] = dimIndex;
```
sdjukicTT: In case of single-dimension scatter, in StableHLO index_vector_dim can be equal to rank(scatter_indices). Will this cause a problem? In that case calling extractIndices wouldn't be necessary (neither slice op nor reshape op are needed).

lib/Conversion/TTIRToTTNN/TTIRToTTNN.cpp
sdjukicTT: But input dim 0 has size 1 after expanding, the original dim 0 moved to dim 1?

lib/Conversion/TTNNToEmitC/TTNNToEmitC.cpp
ddilbazTT: Do you mean remove the comment or remove the line? For context, I was following:

```
struct ScatterOperation {
    static Tensor invoke(
        const QueueId& queue_id,
        const Tensor& input_tensor,
        const int32_t& dim,
        const Tensor& index_tensor,
        const Tensor& source_tensor,
        const std::optional<MemoryConfig>& opt_out_memory_config,
        const std::optional<scatter::ScatterReductionType>& opt_reduction);
};
```

jbbieniekTT: i mean we don't use `queue_id` anymore, it has been removed from scatter as well

svuckovicTT: fyi, this was a very recent change: https://github.com/tenstorrent/tt-metal/pull/27635/

lib/Conversion/TTIRToTTNN/TTIRToTTNN.cpp
```diff
@@ -1573,38 +1573,436 @@ class RandOpConversionPattern : public OpConversionPattern<ttir::RandOp> {
...
+     * update_window_dims=[2], inserted_window_dims=[0]
+     *
+     * Operations:
+     *   1. Extract indices:
+     *      - ttnn.slice_static: begins=[0,0,0], ends=[1,5,1] ->
+     * tensor<1x5x1xi32>
+     *   2. Expand input tensor (rank 2 -> 3):
+     *      - ttnn.reshape: <205x768xf32> -> tensor<1x2050x768xf32>
+     *   3. Broadcast indices to match updates:
+     *      - ttnn.repeat: repeat_dims=<1x1x768> -> tensor<1x5x768xi32>
```
sdjukicTT: Could we skip this step in the future by lowering to ttnn.tosa_scatter?

lib/Conversion/TTIRToTTNN/TTIRToTTNN.cpp
```diff
@@ -1573,38 +1573,436 @@ class RandOpConversionPattern : public OpConversionPattern<ttir::RandOp> {
...
+    }
+
+    RankedTensorType squeezedType = RankedTensorType::get(
+        squeezedShape, scatterIndicesType.getElementType(),
+        scatterIndicesType.getEncoding());
+
+    llvm::SmallVector<int32_t> squeezedShapeI32;
+    for (int64_t dim : squeezedShape) {
+      squeezedShapeI32.push_back(static_cast<int32_t>(dim));
+    }
```
sdjukicTT: You can avoid this loop by adding static_cast<int32_t> to the loop where squeezedShape is made and making it int32_t from the start.

lib/Conversion/TTNNToEmitC/TTNNToEmitC.cpp
ddilbazTT: I still see it here: https://github.com/tenstorrent/tt-metal/blob/main/ttnn/cpp/ttnn/operations/data_movement/scatter/scatter.hpp I'll keep in queue id for now. Right now I'm using a tt-metal version that's custom because I cherry picked Jerry's changes to the tt-metal version we're currently using. If tt-metal gets uplifted in the meantime I'll make this change!

ddilbazTT: Also seems like they reverted the removal recently: https://github.com/tenstorrent/tt-metal/pull/27635 that might complicate things in the metal uplifts, we'll see

lib/Conversion/TTIRToTTNN/TTIRToTTNN.cpp
```diff
@@ -1573,38 +1573,436 @@ class RandOpConversionPattern : public OpConversionPattern<ttir::RandOp> {
 
 namespace {
 class ScatterOpConversionPattern : public OpConversionPattern<ttir::ScatterOp> {
+  /**
```
sdjordjevicTT: Can we please move this logic to the TTIRToTTIRDecompostion pass and keep TTIRToTTNN conversion logic simple?

lib/Dialect/TTNN/IR/TTNNOps.cpp
```diff
@@ -2512,6 +2512,40 @@ mlir::tt::ttnn::ReduceScatterOp::fold(FoldAdaptor adaptor) {
...
+  llvm::ArrayRef<int64_t> indexShape = indexType.getShape();
+  llvm::ArrayRef<int64_t> sourceShape = sourceType.getShape();
+
+  const size_t inputTypeRank = inputShape.size();
+  const size_t indexTypeRank = indexShape.size();
+  const size_t sourceTypeRank = sourceShape.size();
+
+  if (inputTypeRank != indexTypeRank || inputTypeRank != sourceTypeRank ||
+      indexTypeRank != sourceTypeRank) {
+    return emitOpError() << "Input tensor, index tensor, and source tensor "
```
sdjordjevicTT, line 2605: Is this something that is a limitation of the op implementation, or is it just not implemented?

sdjukicTT (changes requested):
I think the logic of transforming operands to fit `ttnn.scatter` (in `TTIRtoTTNN.cpp`) is flawed/incomplete.

Some problems:
- `extractIndices` is called even in single-dim scatter, which adds an unnecessary slice op
- when  `index_vector_dim` is equal to `rank(scatter_indices)`, it breaks
- logic for how input, index and source are made to have the same rank doesn't make sense to me: why is `inserted_window_dims` used to infer where extra dimensions of size 1 should be added to input, it's seemingly unrelated
- up for discussion whether it is better to use multiply and add ops or matmul op to transform multiple dim scatter indices

The 3 types of scatter op that we get from jax tt-xla inference model tests (below) aren't working when tested with `ttmlir-opt` and `ttrt`, either breaking or giving incorrect results. If some case can't be supported, we should get an error message that states what combination of parameters is unsupported. Also code should be easy to follow and understand.

```
%19 = "stablehlo.scatter"(%16, %17, %18) <{indices_are_sorted = true, scatter_dimension_numbers = #stablehlo.scatter<update_window_dims = [0, 1], inserted_window_dims = [1], scatter_dims_to_operand_dims = [1]>, unique_indices = true}> ({
    ^bb0(%arg207: tensor<i1> loc("scatter"), %arg208: tensor<i1> loc("scatter")):
      stablehlo.return %arg208 : tensor<i1> loc(#loc879)
    }) : (tensor<1x258x1xi1>, tensor<1xi32>, tensor<1x1xi1>) -> tensor<1x258x1xi1> loc(#loc879)

 %5 = "stablehlo.scatter"(%2, %4, %3) <{indices_are_sorted = true, scatter_dimension_numbers = #stablehlo.scatter<update_window_dims = [0, 1], scatter_dims_to_operand_dims = [1]>, unique_indices = true}> ({
    ^bb0(%arg354: tensor<i32> loc("scatter"), %arg355: tensor<i32> loc("scatter")):
      stablehlo.return %arg355 : tensor<i32> loc(#loc1422)
    }) : (tensor<1x128xi32>, tensor<1xi32>, tensor<1x127xi32>) -> tensor<1x128xi32> loc(#loc1422)

 %8 = "stablehlo.scatter"(%5, %6, %7) <{indices_are_sorted = true, scatter_dimension_numbers = #stablehlo.scatter<update_window_dims = [0], inserted_window_dims = [1], scatter_dims_to_operand_dims = [1]>, unique_indices = true}> ({
    ^bb0(%arg354: tensor<i32> loc("scatter"), %arg355: tensor<i32> loc("scatter")):
      stablehlo.return %arg355 : tensor<i32> loc(#loc1424)
    }) : (tensor<1x128xi32>, tensor<1xi32>, tensor<1xi32>) -> tensor<1x128xi32> loc(#loc1424)
```

lib/Conversion/TTNNToEmitC/TTNNToEmitC.cpp
ddilbazTT: Removed it now

lib/Dialect/TTNN/IR/TTNNOps.cpp
ddilbazTT: So yes and no. The torch op requires input, source and index to have the same rank; and that source and index have the same dimensions. In the current conversion, I limited it to only indices expansion if necessary.

lib/Conversion/TTIRToTTNN/TTIRToTTNN.cpp
ddilbazTT: I was doing unnecessary slices and reshaping here. I refactored it to make it simpler. I think the only slices are coming from multi-dimensional scatter indices.

ddilbazTT:
@sdjukicTT @sdjordjevicTT I updated this PR to be much cleaner than how it used to be, and I targeted element-wise scatter in this PR. I want to do window based scatter as a follow-up PR. Our team has been using this branch for a while so I want to divide the scatter op implementation into a few PRs and get simple element-wise scatter in. I tested simple element-wise scatter in tt-xla: https://github.com/tenstorrent/tt-xla/pull/1514 Hope you guys could take a look at this PR.

lib/Conversion/TTIRToTTNN/TTIRToTTNN.cpp
ddilbazTT: Are we talking about indices expansion? Which cases should we lower to tosa_scatter? I think I could do that in a follow-up PR.

lib/Conversion/TTIRToTTNN/TTIRToTTNN.cpp
ddilbazTT: I could but how should I differentiate between names? Both are called scatter so should I use `TTIR_ScatterSHLO` vs `TTIR_ScatterTTNN`?

lib/Conversion/TTNNToEmitC/TTNNToEmitC.cpp
jbbieniekTT: yeah, it's good to avoid it altogether, i was advised by Artem not to use queue ids in my ops

lib/Conversion/TTIRToTTNN/TTIRToTTNN.cpp
ddilbazTT: Yeah I think I was unnecessarily doing slices and reshapes. I am using slices in multi dimensional scatter as I am representing that as a flattened scatter operation. I still use reshape to match indices shape with updates shape.

lib/Conversion/TTIRToTTNN/TTIRToTTNN.cpp
```diff
@@ -1573,38 +1573,402 @@ class RandOpConversionPattern : public OpConversionPattern<ttir::RandOp> {
...
+                "if needed.");
+      }
+
+      // Extract multi-dimensional indices and flatten to 1D
+      Value finalIndexTensor =
+          extractMultiDimensionalScatterIndices(op, adaptor, rewriter);
+
+      // Flatten input tensor to 1D
+      Value flattenedInput =
+          flattenTensor(inputTensor, rewriter, op.getLoc(), "_input_flatten");
```
sdjukicTT: Didn't test, but I think this implementation works only if update_window_dims is empty. Because the whole input is flattened to 1D, but finalIndexTensor strides only take into account indexing dims. This should be either handled or checked.

lib/Conversion/TTIRToTTNN/TTIRToTTNN.cpp
```diff
@@ -1573,38 +1573,402 @@ class RandOpConversionPattern : public OpConversionPattern<ttir::RandOp> {
...
+
+    // Number of dimensions being indexed
+    int64_t numIndexDims = indexShape[indexVectorDim];
+
+    // Calculate strides for each dimension (product of subsequent dimensions)
+    llvm::SmallVector<int64_t> strides(numIndexDims);
+    for (int64_t i = 0; i < numIndexDims; ++i) {
+      int64_t stride = 1;
+      for (int64_t j = i + 1; j < numIndexDims; ++j) {
+        stride *= inputShape[j];
```
sdjukicTT: This assumes that scatter_dims_to_operand_dims are in order ([0,1,2.., numIndexDims - 1]). This should be either handled or checked.

lib/Conversion/TTIRToTTNN/TTIRToTTNN.cpp
```diff
@@ -1573,38 +1573,402 @@ class RandOpConversionPattern : public OpConversionPattern<ttir::RandOp> {
...
+      llvm::SmallVector<int32_t> newShapeI32;
+      for (int64_t dimSize : dimIndexShape) {
+        newShapeI32.push_back(static_cast<int32_t>(dimSize));
+      }
+
+      Value dimensionIndices = rewriter.create<ttnn::ReshapeOp>(
+          ttmlir::utils::appendLocationSuffix(
+              op.getLoc(), "_dim_" + std::to_string(dim) + "_reshape"),
+          dimIndexType, dimensionSlice, rewriter.getI32ArrayAttr(newShapeI32),
+          nullptr);
```
sdjukicTT: Is this reshape redundant if indices are flattened to 1D later?

lib/Conversion/TTIRToTTNN/TTIRToTTNN.cpp
```diff
@@ -1573,38 +1573,402 @@ class RandOpConversionPattern : public OpConversionPattern<ttir::RandOp> {
...
+                "if needed.");
+      }
+
+      Value finalIndexTensor =
+          extractElementWiseScatterIndices(op, adaptor, rewriter);
+
+      auto dimAttr = rewriter.getI32IntegerAttr(dim);
+      RankedTensorType convertedResultType = mlir::cast<RankedTensorType>(
+          typeConverter->convertType(op.getResult().getType()));
+      rewriter.replaceOpWithNewOp<ttnn::ScatterOp>(
```
sdjukicTT: Will this work if scatter_dims_to_operand_dims != inserted_window_dims? Because ttnn.scatter indexes one dim, and other dims are window dims.

lib/Conversion/TTIRToTTNN/TTIRToTTNN.cpp
```diff
@@ -1573,38 +1573,402 @@ class RandOpConversionPattern : public OpConversionPattern<ttir::RandOp> {
...
+    ArrayRef<int32_t> updateWindowDims = op.getUpdateWindowDims();
+    llvm::SmallVector<int64_t> repeatDims(indexShape.size(), 1);
+    bool needsRepeat = false;
+
+    // For each update_window_dim, set repeat factor to match update tensor size
+    for (auto dimAttr : updateWindowDims) {
+      int64_t dim = dimAttr;
+      if (static_cast<size_t>(dim) < indexShape.size() &&
+          indexShape[dim] != updateShape[dim]) {
+        repeatDims[dim] = updateShape[dim];
```
sdjukicTT: Will this work if update_window_dims are not [1,2,..] (if inserted_window_dims and scatter_dims_to_operand_dims aren't exactly [0])? Because if indexShape was original [N, 1] (N > 1), repeat op result wouldn't be same shape as updates.

lib/Conversion/TTIRToTTNN/TTIRToTTNN.cpp
```diff
@@ -1573,38 +1573,402 @@ class RandOpConversionPattern : public OpConversionPattern<ttir::RandOp> {
 
 namespace {
 class ScatterOpConversionPattern : public OpConversionPattern<ttir::ScatterOp> {
-public:
   using OpConversionPattern<ttir::ScatterOp>::OpConversionPattern;
 
+public:
```
sdjordjevicTT, line 1709: > @[sdjordjevicTT](https://github.com/sdjordjevicTT) sdjordjevicTT [3 weeks ago](https://github.com/tenstorrent/tt-mlir/pull/4913#discussion_r2359463297)
> Can we please move this logic to the TTIRToTTIRDecompostion pass and keep TTIRToTTNN conversion logic simple?
> 
> Member
> Author
> @[ddilbazTT](https://github.com/ddilbazTT) ddilbazTT [last week](https://github.com/tenstorrent/tt-mlir/pull/4913#discussion_r2389729642)
> I could but how should I differentiate between names? Both are called scatter so should I use TTIR_ScatterSHLO vs TTIR_ScatterTTNN?

Can we please introduce the PyTorch version to TTIR? To differentiate it from the StableHLO's one, we can call it for example:
`TTIR_ScatterInDimOp`

Then we can do the TTIRToTTIR decomposition pass, where we lower `TTIR_ScatterOp` to `TTIR_ScatterInDimOp`

lib/Conversion/TTIRToTTNN/TTIRToTTNN.cpp
```diff
@@ -1573,38 +1573,402 @@ class RandOpConversionPattern : public OpConversionPattern<ttir::RandOp> {
...
+      // check ranks of input, update, and index - ttnn requires them to be
+      // equal
+      size_t targetRank =
+          std::max({inputShape.size(), updateShape.size(), indexShape.size()});
+
+      if (inputShape.size() != targetRank || updateShape.size() != targetRank) {
+        return rewriter.notifyMatchFailure(
+            op, "TTNN element-wise scatter requires input and update tensors "
+                "to have the same rank. Please add support for rank promotion "
+                "if needed.");
```
sdjukicTT: In checkBasicLegality it is checked that: rank(update) = size(update_window_dims) + size(inserted_window_dims). From stablehlo.scatter constraints (https://openxla.org/stablehlo/spec#constraints_79) and the fact that we support no batching dims: rank(input) = size(update_window_dims) + size(inserted_window_dims). Here we can already know that rank(input) = rank(update).

lib/Conversion/TTIRToTTNN/TTIRToTTNN.cpp
```diff
@@ -1573,38 +1573,402 @@ class RandOpConversionPattern : public OpConversionPattern<ttir::RandOp> {
...
+      for (int64_t dim : newShape) {
+        newShapeI32.push_back(static_cast<int32_t>(dim));
       }
+      indexTensor = rewriter.create<ttnn::ReshapeOp>(
+          op.getLoc(), newIndexType, indexTensor,
+          rewriter.getI32ArrayAttr(newShapeI32), nullptr);
+      indexType = newIndexType;
+      indexShape = newShape;
+    }
+    if (indexShape.size() > updateShape.size()) {
```
sdjukicTT: No need to be checked both here and on line 1618 (right before calling this method).

lib/Conversion/TTIRToTTNN/TTIRToTTNN.cpp
```diff
@@ -1573,38 +1573,402 @@ class RandOpConversionPattern : public OpConversionPattern<ttir::RandOp> {
...
+
+    // Repeat along update_window_dims to match update tensor shape
+    ArrayRef<int32_t> updateWindowDims = op.getUpdateWindowDims();
+    llvm::SmallVector<int64_t> repeatDims(indexShape.size(), 1);
+    bool needsRepeat = false;
+
+    // For each update_window_dim, set repeat factor to match update tensor size
+    for (auto dimAttr : updateWindowDims) {
+      int64_t dim = dimAttr;
+      if (static_cast<size_t>(dim) < indexShape.size() &&
```
sdjukicTT: Indices are reshaped to have the same rank as update, so this will always be true.

lib/Conversion/TTIRToTTNN/TTIRToTTNN.cpp
```diff
@@ -1573,38 +1573,402 @@ class RandOpConversionPattern : public OpConversionPattern<ttir::RandOp> {
...
+      size_t targetRank = std::max({inputShape.size(), updateShape.size()});
+
+      int32_t dim = 0; // Always scatter along dimension 0 for flattened tensors
+
+      if (inputShape.size() != targetRank || updateShape.size() != targetRank) {
+        return rewriter.notifyMatchFailure(
+            op, "TTNN element-wise scatter requires input and update tensors "
+                "to have the same rank. Please add support for rank promotion "
+                "if needed.");
+      }
```
sdjukicTT: Same as in single dim case, redundant check.
In checkBasicLegality it is checked that: rank(update) = size(update_window_dims) + size(inserted_window_dims). From stablehlo.scatter constraints (https://openxla.org/stablehlo/spec#constraints_79) and the fact that we support no batching dims: rank(input) = size(update_window_dims) + size(inserted_window_dims). Here we can already know that rank(input) = rank(update).

lib/Conversion/TTIRToTTNN/TTIRToTTNN.cpp
ddilbazTT: I don't think that's necessary. i think if `update_window_dims` is non-empty, it would fail the following condition in `checkBasicLegality`:
```
    // Check update_window_dims.
    for (auto dim : updateWindowDims) {
      assert(dim >= 0 && dim < updateRank &&
             "update_window_dims contains invalid dimension index");
      dimsCovered[dim] = true;
    }

    // Check inserted_window_dims.
    for (auto dim : insertedWindowDims) {
      assert(dim >= 0 && dim < updateRank &&
             "inserted_window_dims contains invalid dimension index");
      if (dimsCovered[dim]) {
            op, "update_window_dims and inserted_window_dims have overlapping "
                "dimensions");
      }
      dimsCovered[dim] = true;
    }

    // Check that all dimensions are covered.
    for (int64_t i = 0; i < updateRank; ++i) {
      if (!dimsCovered[i]) {
            op, "ScatterInDim does not support window scatter.");
      }
    }
```

>

ddilbazTT: I couldn't think of an example where `update_window_dims` is non-empty, there is a multi-dimensional scatter, and it satisfies `checkBasicLegality`.

lib/Conversion/TTIRToTTNN/TTIRToTTNN.cpp
ddilbazTT: I think this would only happen if it's a window scatter, which we don't cover. So I don't think we need a condition here as it's checked in checkBasicLegality.

lib/Conversion/TTIRToTTNN/TTIRToTTNN.cpp
ddilbazTT: Removed the check before instead of this one.

lib/Conversion/TTIRToTTNN/TTIRToTTNN.cpp
ddilbazTT: I don't think so. This reshape makes it match to the stride value shape. Keeping stride value shape the same as index shape might introduce some logic that's not easy to follow.

ddilbazTT: For example:
`  %1 = "ttir.scatter"(%arg0, %arg1, %arg2, %0) <{index_vector_dim = 2 : i32, indices_are_sorted = false, input_batching_dims = array<i32>, inserted_window_dims = array<i32: 0, 1>, scatter_dims_to_operand_dims = array<i32: 0, 1>, scatter_indices_batching_dims = array<i32>, unique_indices = false, update_window_dims = array<i32>}> : (tensor<71x32xbf16>, tensor<71x4x2xi64>, tensor<71x4xbf16>, tensor<71x32xbf16>) -> tensor<71x32xbf16>
` 

This becomes:

```
  func.func @scatter_simple_3(%arg0: tensor<71x32xbf16>, %arg1: tensor<71x4x2xi64>, %arg2: tensor<71x4xbf16>) -> tensor<71x32xbf16> {
    %0 = ttir.empty() : tensor<71x32xbf16>
    %1 = ttir.empty() : tensor<71x4x1xi64>
    %2 = "ttir.slice_static"(%arg1, %1) <{begins = [0 : i32, 0 : i32, 0 : i32], ends = [71 : i32, 4 : i32, 1 : i32], step = [1 : i32, 1 : i32, 1 : i32]}> : (tensor<71x4x2xi64>, tensor<71x4x1xi64>) -> tensor<71x4x1xi64>
    %3 = ttir.empty() : tensor<71x4xi64>
    %4 = "ttir.reshape"(%2, %3) <{shape = [71 : i32, 4 : i32]}> : (tensor<71x4x1xi64>, tensor<71x4xi64>) -> tensor<71x4xi64>
    %5 = "ttir.full"() <{fill_value = 32 : i32, shape = array<i32: 71, 4>}> : () -> tensor<71x4xi64>
    %6 = ttir.empty() : tensor<71x4xi64>
    %7 = "ttir.multiply"(%4, %5, %6) : (tensor<71x4xi64>, tensor<71x4xi64>, tensor<71x4xi64>) -> tensor<71x4xi64>
    %8 = ttir.empty() : tensor<71x4x1xi64>
    %9 = "ttir.slice_static"(%arg1, %8) <{begins = [0 : i32, 0 : i32, 1 : i32], ends = [71 : i32, 4 : i32, 2 : i32], step = [1 : i32, 1 : i32, 1 : i32]}> : (tensor<71x4x2xi64>, tensor<71x4x1xi64>) -> tensor<71x4x1xi64>
    %10 = ttir.empty() : tensor<71x4xi64>
    %11 = "ttir.reshape"(%9, %10) <{shape = [71 : i32, 4 : i32]}> : (tensor<71x4x1xi64>, tensor<71x4xi64>) -> tensor<71x4xi64>
    %12 = ttir.empty() : tensor<71x4xi64>
    %13 = "ttir.add"(%7, %11, %12) : (tensor<71x4xi64>, tensor<71x4xi64>, tensor<71x4xi64>) -> tensor<71x4xi64>
    %14 = ttir.empty() : tensor<284xi64>
    %15 = "ttir.reshape"(%13, %14) <{shape = [284 : i32]}> : (tensor<71x4xi64>, tensor<284xi64>) -> tensor<284xi64>
    %16 = ttir.empty() : tensor<2272xbf16>
    %17 = "ttir.reshape"(%arg0, %16) <{shape = [2272 : i32]}> : (tensor<71x32xbf16>, tensor<2272xbf16>) -> tensor<2272xbf16>
    %18 = ttir.empty() : tensor<284xbf16>
    %19 = "ttir.reshape"(%arg2, %18) <{shape = [284 : i32]}> : (tensor<71x4xbf16>, tensor<284xbf16>) -> tensor<284xbf16>
    %20 = ttir.empty() : tensor<2272xbf16>
    %21 = "ttir.scatter_in_dim"(%17, %15, %19, %20) <{dim = 0 : i32}> : (tensor<2272xbf16>, tensor<284xi64>, tensor<284xbf16>, tensor<2272xbf16>) -> tensor<2272xbf16>
    %22 = ttir.empty() : tensor<71x32xbf16>
    %23 = "ttir.reshape"(%21, %22) <{shape = [71 : i32, 32 : i32]}> : (tensor<2272xbf16>, tensor<71x32xbf16>) -> tensor<71x32xbf16>
    return %23 : tensor<71x32xbf16>
  }
```

lib/Conversion/TTIRToTTNN/TTIRToTTNN.cpp
ddilbazTT: Edit: It would not work with scatter_dims_to_operand_dims that aren't exactly [0]

ddilbazTT: I'll just add an assert for now

mmanzoorTT:
@ddilbazTT https://github.com/tenstorrent/tt-mlir/pull/5408 fixes the EmitC conversion for const-eval wrapper function with zero arguments.

lib/Conversion/TTIRToTTNN/TTIRToTTNN.cpp
sdjukicTT: I tried two cases where update_window_dims = [2], there is a multi-dimensional scatter, and it satisfies checkBasicLegality:
1. Where updateShape[2] > 1
This really does fail in compilation, in TTIR op verifier, because after flattening and replacing with ScatterInDim, indexShape != sourceShape
2. Where updateShape[2] = 1
This came from valid StableHLO and works on CPU, but gives wrong result on your branch, TTIR graph pasted below.

I think we should check if update_window_dims is empty because even in the first example, the error message would be clearer (the user would know right away that it's unsupported). Getting "indexShape != sourceShape" error for shapes that are flattened is harder to understand.
```
module @jit_scatter attributes {mhlo.num_partitions = 1 : i32, mhlo.num_replicas = 1 : i32, ttcore.meshes = #ttcore.meshes<[<"empty_mesh" = 1x1>]>} {
  func.func public @main(%arg0: tensor<2x3x3xf32> {ttcore.argument_type = #ttcore.argument_type<input>, ttcore.shard_status = #ttcore.shard_status<presharded>}, %arg1: tensor<2x1x2xi32> {ttcore.argument_type = #ttcore.argument_type<input>, ttcore.shard_status = #ttcore.shard_status<presharded>}, %arg2: tensor<2x1x1xf32> {ttcore.argument_type = #ttcore.argument_type<input>, ttcore.shard_status = #ttcore.shard_status<presharded>}) -> (tensor<2x3x3xf32> {jax.result_info = "result", ttcore.shard_status = #ttcore.shard_status<unsharded>}) {
    %0 = ttir.empty() : tensor<2x3x3xf32>
    %1 = "ttir.constant"() <{value = dense<[[[1,2]],[[1,1]]]> : tensor<2x1x2xi32>}> : () -> tensor<2x1x2xi32>
    %2 = "ttir.scatter"(%arg0, %1, %arg2, %0) <{index_vector_dim = 2 : i32, indices_are_sorted = false, input_batching_dims = array<i32>, inserted_window_dims = array<i32: 0, 1>, scatter_dims_to_operand_dims = array<i32: 0, 1>, scatter_indices_batching_dims = array<i32>, unique_indices = false, update_window_dims = array<i32: 2>}> : (tensor<2x3x3xf32>, tensor<2x1x2xi32>, tensor<2x1x1xf32>, tensor<2x3x3xf32>) -> tensor<2x3x3xf32>
    return %2 : tensor<2x3x3xf32>
  }
}
```

lib/Conversion/TTIRToTTNN/TTIRToTTNN.cpp
sdjukicTT: When you make stride tensor, you can make it the right shape in the start, the shape of slice op output.
```
RankedTensorType dimIndexType = RankedTensorType::get(
            dimIndexShape, indexType.getElementType(), indexType.getEncoding());

        Value strideTensor = rewriter.create<ttir::FullOp>(
            ttmlir::utils::appendLocationSuffix(
                op.getLoc(), "_stride_" + std::to_string(dim)),
            dimIndexType, scalarAttr);
```
dimIndexShape can be replaced with sliceShape, with dimIndexShape deleted and extra Reshape of slice output deleted.

include/ttmlir/Dialect/TTNN/IR/TTNNOps.td
```diff
@@ -2035,11 +2035,35 @@ def TTNN_DeallocateOp : TTNN_MemFreeOp<"deallocate"> {
...
+        - `input_tensor` (ttnn.Tensor): The tensor being updated.
+        - `index_tensor` (ttnn.Tensor): Indices where values will be written to.
+        - `source_tensor` (ttnn.Tensor): The values to scatter into the input tensor.
+        - `dim` (int32_t): The dimension along which to scatter.
+        - `cq_id` i32: The command queue to copy the tensor with. Must be 0 or 1.
+    }];
+
+    let arguments = (ins AnyRankedTensor:$input_tensor,
+                         AnyRankedTensor:$index_tensor,
+                         AnyRankedTensor:$source_tensor,
```
mmanzoorTT:
```suggestion
    let arguments = (ins AnyRankedTensor:$input,
                         AnyRankedTensor:$index,
                         AnyRankedTensor:$source,
```

runtime/lib/ttnn/operations/data_movement/scatter.cpp
```diff
@@ -10,32 +10,25 @@
...
-      ::ttnn::pad(onesLikeLhs, rhs.padded_shape().to_array_4D(), startIndex, 0,
-                  false, std::nullopt);
-
-  ::ttnn::Tensor tempA = ::ttnn::pad(lhs, rhs.padded_shape().to_array_4D(),
-                                     startIndex, 0, false, std::nullopt);
-
-  ::ttnn::Tensor out = ::ttnn::where(indexPad, tempA, rhs);
+  //   LOG_ASSERT(::tt::runtime::ttnn::utils::inSystemMemory(op->out()) ||
+  //                  outputMemoryConfig.has_value(),
+  //              "Memory config must exist for device tensors");
```
jnie-TT: Nit: stale comment

include/ttmlir/Target/TTNN/operations/data_movement.fbs
```diff
@@ -10,6 +10,16 @@ table ConcatOp {
...
 }
 
+table ScatterOp {
+  input: tt.target.ttnn.TensorRef;
+  out: tt.target.ttnn.TensorRef;
+  index: tt.target.ttnn.TensorRef;
+  source: tt.target.ttnn.TensorRef;
+  dim: int32;
+  memory_config: tt.target.ttnn.MemoryConfig;
+  cq_id: uint32;
```
jnie-TT: Nit: Doesn't seem like cq_id is used, can we remove it?

lib/Conversion/TTIRToTTIRDecomposition/TTIRToTTIRDecomposition.cpp
```diff
@@ -2342,6 +2343,365 @@ struct RequantizeOpPattern : public OpConversionPattern<ttir::RequantizeOp> {
...
+    // Get update tensor rank.
+    RankedTensorType updateType = op.getUpdate().getType();
+    int64_t updateRank = updateType.getRank();
+
+    // Create array to track which dimensions are covered.
+    llvm::SmallVector<bool> dimsCovered(updateRank, false);
+
+    // Check update_window_dims.
+    for (auto dim : updateWindowDims) {
+      assert(dim >= 0 && dim < updateRank &&
```
vwellsTT: nit: throughout changes in this file, probably better to use `TT_assert` instead of regular `assert`, since you include header + use `TT_assertv` here too

lib/Conversion/TTIRToTTIRDecomposition/TTIRToTTIRDecomposition.cpp
```diff
@@ -2342,6 +2343,365 @@ struct RequantizeOpPattern : public OpConversionPattern<ttir::RequantizeOp> {
...
+      if (!dimsCovered[i]) {
+        return rewriter.notifyMatchFailure(
+            op, "ScatterInDim does not support window scatter.");
+      }
+    }
+
+    // Check that scatter_dims_to_operand_dims is in order.
+    ArrayRef<int32_t> scatterDimsToOperandDims =
+        op.getScatterDimsToOperandDims();
+    if (!std::is_sorted(scatterDimsToOperandDims.begin(),
```
vwellsTT: nit: llvm::is_sorted instead https://llvm.org/doxygen/STLExtras_8h.html

lib/Conversion/TTNNToEmitC/TTNNToEmitC.cpp
```diff
@@ -2022,10 +2022,17 @@ class LoadCachedOpConversionPattern
     // Restore the insertion point to continue with the function
     rewriter.restoreInsertionPoint(currentInsertionPoint);
 
+    bool isZeroArgWrapper = adaptor.getInputs().size() == 0;
```
vwellsTT: nit: I know we don't really use `const` much, since a lot of llvm types don't work well with etc, but imo this looks weird without it :)

lib/Dialect/TTNN/Transforms/Workarounds/Decomposition/ScatterOpRewritePattern.cpp
```diff
@@ -0,0 +1,121 @@
...
+
+#include "mlir/IR/BuiltinTypes.h"
+
+namespace mlir::tt::ttnn::workarounds::decomposition {
+
+LogicalResult
+TTNNScatterWorkarounds::matchAndRewrite(ttnn::ScatterOp op,
+                                        PatternRewriter &rewriter) const {
+
+  /*
```
vwellsTT: nit: this is a great comment! But per out style guide we should use // instead of /* */ here

lib/Dialect/TTNN/Transforms/Workarounds/Decomposition/ScatterOpRewritePattern.cpp
```diff
@@ -0,0 +1,121 @@
...
+    return failure();
+  }
+
+  // Calculate number of chunks needed using round-up (ceiling) division.
+  int64_t numChunks =
+      (scatterAxisSize + MAX_SCATTER_SIZE - 1) / MAX_SCATTER_SIZE;
+
+  Value currentResult = op.getInput();
+
+  // Process each chunk sequentially
```
vwellsTT: nit: a few comments not in proper sentence form here

lib/Conversion/TTNNToEmitC/TTNNToEmitC.cpp
ddilbazTT: This is actually a change I uplifted from another PR since the scatter uplift exposed a consteval bug: https://github.com/tenstorrent/tt-mlir/pull/5408 I'll leave your remark there

vwellsTT: cool, don't really worry about it, pretty nit-picky nit
(I guess approved the other PR without leaving this comment I see, ha!)

lib/Conversion/TTIRToTTIRDecomposition/TTIRToTTIRDecomposition.cpp
ddilbazTT: You're right, will do!

lib/Conversion/TTIRToTTIRDecomposition/TTIRToTTIRDecomposition.cpp
ddilbazTT: Thanks for catching this, my bad 🥲

include/ttmlir/Dialect/TTNN/IR/TTNNOps.td
```diff
@@ -2058,11 +2058,35 @@ def TTNN_DeallocateOp : TTNN_MemFreeOp<"deallocate"> {
...
+        - `dim` (int32_t): The dimension along which to scatter.
+        - `cq_id` i32: The command queue to copy the tensor with. Must be 0 or 1.
+    }];
+
+    let arguments = (ins AnyRankedTensor:$input,
+                         AnyRankedTensor:$index,
+                         AnyRankedTensor:$source,
+                         I32Attr:$dim,
+                         OptionalAttr<TTNN_MemoryConfigAttr>:$memory_config,
+                         DefaultValuedAttr<UI32Attr, "0">:$cq_id);
```
mmanzoorTT: `command queue` is removed from ttnn::scater op. Please remove it.

mmanzoorTT: ttnn::scatter op uses [opt_reduction](https://github.com/tenstorrent/tt-metal/blob/a7004f7e1d62a02f3dba06eee767c6c245f1346a/ttnn/cpp/ttnn/operations/data_movement/scatter/scatter.hpp#L24C9-L24C74) as last argument. Any plan for adding it?

mmanzoorTT: It is optional argument; so you may add it and assign `nulltptr` if you are not planning to use it for now.

mmanzoorTT: opt_reduction is defined as class enum of type uint8_t; so you may use std::optional<uint8_t>.

include/ttmlir/Target/TTNN/operations/data_movement.fbs
```diff
@@ -10,6 +10,15 @@ table ConcatOp {
...
 }
 
+table ScatterOp {
+  input: tt.target.ttnn.TensorRef;
+  out: tt.target.ttnn.TensorRef;
+  index: tt.target.ttnn.TensorRef;
+  source: tt.target.ttnn.TensorRef;
+  dim: int32;
+  memory_config: tt.target.ttnn.MemoryConfig;
+}
```
mmanzoorTT, line 20: Add `opt_reduction` here as well if you decide to add in TTNN dialect.

include/ttmlir/Dialect/TTNN/IR/TTNNOps.td
ddilbazTT: @vladimirjovanovicTT 's team is working on adding opt_reduction support since they have use cases in training models, I don't want to convolute this PR as they are waiting for this one to land.

include/ttmlir/Target/TTNN/operations/data_movement.fbs
ddilbazTT: I think it's best to not include this in this PR since there will be coming work on adding opt reduction

include/ttmlir/Dialect/TTNN/IR/TTNNOps.td
vladimirjovanovicTT: Thanks @ddilbazTT , that is right. It is crucial for forge training team for this to land in the current state soon.
My team will work on adding reduction support and other potential extensions as future PRs.

ddilbazTT: Removed

PR #4914, opened 2025-09-12
Fix library dependencies

azecevicTT:
### Ticket
N/A

### Problem description
Library dependencies were broken in multiple places. One example that exposes it is an `OptimizerTests` target in debug build mode. TBH, it's a miracle that we managed to get this far without catching it, but the two main reasons are:
- in release build `-ffunction-sections -Wl,--gc-sections` is used, so unused functions don't need to be resolved
- tools like `ttmlir-opt` and `ttmlir-translate` link all MLIR libraries

### Checklist

### DIFF
--- a/lib/Dialect/StableHLO/Utils/CMakeLists.txt
+++ b/lib/Dialect/StableHLO/Utils/CMakeLists.txt
@@ -4,4 +4,10 @@ add_mlir_dialect_library(TTMLIRSTABLEHLOUtils
 
   ADDITIONAL_HEADER_DIRS
   ${MLIR_MAIN_INCLUDE_DIR}/mlir/Dialect/StableHLO
+
+  LINK_LIBS
+  PUBLIC
+  SdyDialect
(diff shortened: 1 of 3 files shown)


rpavlovicTT (approved):
Maybe we should have a CI job that just builds every lib (as a target) in release.
This should catch any linkage problem. Also it should be lightweight job if it reuses already compiled archives and just invokes linker on top of. What do you think?

mmanzoorTT (approved):

tapspatel (approved):

azecevic000:
@rpavlovicTT In principle, I'm not against it, but it would be hard to properly maintain such CI job. We would probably have to write our own CMake module with functions like `add_ttmlir_library`, which we would use instead of `add_library` etc. to be able to collect all targets automatically.

Just to clarify, I believe we are already building most targets in the release build in CI, one way or the other. These problems only got exposed in the debug build mode.

PR #4915, opened 2025-09-12
[multidevice] Integrate Llama attention tensor-parallel tests

### LINKED ISSUE #3213: Manually sharded llama
Verify tt-mlir supports forward pass of transformer models over multidevice.
Llama attention block already implemented with TTIRBuilder. So you can start from it. Convert llama attention block TTIRBuilder test into multi device environment.

hkwonTT:
### Ticket
closes #3213

### Problem description
The Llama attention tensor-parallel test was previously split into two parts due to low-PCC errors.
 - The `ttnn.softmax` op produced incorrect results with a wider range of input values.
The `ttnn.softmax` op has been updated to support the `numeric_stable` flag, which improves accuracy.
This update fixes the low-PCC errors.

### What's changed
 - Integrated the two test parts into a single test
 - Fixed a typo bug in the softmax golden library
 - Added a test with a 1x8 mesh

### Checklist


test/python/golden/test_ttir_models_llama_tp.py
```diff
@@ -0,0 +1,399 @@
...
+        input_tensors_torch = [
+            input_tensor.shard_map[0]
+            for input_tensor in get_input_tensors_from_builder(
+                operands,
+                builder,
+            )
+        ]
+
+        # Generate the golden output using a single CPU device
+        golden = golden_llama(*input_tensors_torch)
```
tapspatel, line 371: do we need a custom implementation of golden_llama? isn't it sufficient to just set inputs and have builder calculate the goldens

tapspatel: eg
```
builder.set_goldens(inputs={})
builder.add()
builder.relu()
...
return builder.multiply()
```

This will set golden inputs and enable both op by op and output level checking.

test/python/golden/test_ttir_models_llama_tp.py
```diff
@@ -0,0 +1,399 @@
...
+                operand: input_tensor
+                for operand, input_tensor in zip(operands, input_tensors_torch)
+            },
+            {
+                output116: golden,
+            },
+        )
+        # Clear the golden check list for the op-level golden check.
+        # This is a bit hacky, but it's the only way to get both the op-level and program-level golden checks.
+        builder.set_goldens_to_check([], override=True)
```
tapspatel, line 385: Potential user flows

User doesn't want to do anything but store all goldens in flatbuffer

```
def @some_function(in0, in1, builder)
    ....
```
User wants to set inputs and store all goldens in flatbuffer
```
def @some_function(in0, in1, builder):
    builder.set_goldens({in0: torch.tensor, in1: torch:tensor})
    ....
```
User wants to set inputs and outputs and only store them in flatbuffer
```
def @some_function(in0, in1, builder):
    add0 = builder.add(in0, in1)
    builder.set_goldens({in0: torch.tensor, in1: torch:tensor}, {add0: torch.tensor})
    ....
```
User wants to set inputs (or let builder generate them) and select which goldens to store in the flatbuffer
```
def @some_function(in0, in1, builder):
    builder.set_goldens({in0: torch.tensor, in1: torch:tensor})
    add0 = builder.add(in0, in1)
    add1 = builder.add(add0, in1)
    builder.set_goldens_to_check([add0, add1])
    ....
```
User wants to set inputs and set some intermediate goldens to store in the flatbuffer. Inputs + whatever intermediate goldens user set are stored in flatbuffer.
```
def @some_function(in0, in1, builder):
    builder.set_goldens({in0: torch.tensor, in1: torch:tensor})
    add0 = builder.add(in0, in1)
    builder.set_operand_goldens({add0: torch.tensor})
    ....
```
In your case, we should just set the inputs and the builder infra will store all intermediates + outputs.

test/python/golden/test_ttir_models_llama_tp.py
hkwonTT: I’m just making sure that the manual tensor-parallel results match the single-device results exactly.

test/python/golden/test_ttir_models_llama_tp.py
hkwonTT: So your recommendation is to call `set_operand_goldens()`, right? But in this case, I have more than 100 operands to store. This situation might be quite rare, but it does occur.

tapspatel: in your case, you are setting the inputs after the entire graph is executed, which means all the intermediates were executed on random data. You either have to move your set_goldens to the top of the graph or not check intermediates if you have it in the place you do right now.

tapspatel: oh sorry I see you get the inputs from builder

```
for input_tensor in get_input_tensors_from_builder(
                operands,
                builder,
            )
```

PR #4920, opened 2025-09-12
[TTNN Dialect] rotary embedding llama op definition

mtopalovicTT:
This PR adds `rotary_embedding_llama` op to ttnn dialect. Few notes on this op:

* This op has very long list of requirements which are checked in [ttnn runtime](https://github.com/tenstorrent/tt-metal/blob/main/ttnn/cpp/ttnn/operations/experimental/transformer/rotary_embedding_llama/device/rotary_embedding_llama_device_operation.cpp#L14). Initially I've added the same checks in verifier but I decided to remove some of them since it would be hard to keep them in sync with checks in ttnn runtime.
* Fusing which is the most important part of this will come with subsequent


lib/Dialect/TTNN/IR/TTNNOps.cpp
```diff
@@ -3394,4 +3388,84 @@ ::mlir::LogicalResult mlir::tt::ttnn::GenericOp::verify() {
...
+  };
+
+  auto tileLayoutPredicate = [](mlir::RankedTensorType type) {
+    auto encoding = cast<ttnn::TTNNLayoutAttr>(type.getEncoding());
+    return encoding.isTiled();
+  };
+
+  auto deviceStoragePredicate = [](mlir::RankedTensorType type) {
+    return ttnn::utils::isTensorOnDevice(type);
+  };
```
azecevic000, line 3416: I would rather see this in a workaround. All of these constraints are implementation details, and we could workaround them. Fusing will also need to account for all of these constraints which will make it a painful process all around.

mtopalovicTT: Makes sense. I would prefer if I could do it in separate PR (right after this one) to keep stuff simple in this PR.

azecevic000: I'm okay with that, I would just remove these verification steps now, this op probably won't be in usage until fusing is implemented, so there isn't much of a point to keep these verifications.

sdjordjevicTT (approved):

mtopalovicTT:
@sgholamiTT I've added op model implementation for this op. One thing I didn't add (cause I didn't see any value in it) is integration test. By looking at what integration test is testing it's basically the same test as unit test with subtle difference that integration test doesn't go through OpModelInterface and unit test does. Please tell me if I missed something and I need to add it.

+ @rpavlovicTT FYI

mtopalovicTT:
@jnie-TT can you please review for runtime owners?

sgholamiTT:
> @sgholamiTT I've added op model implementation for this op. One thing I didn't add (cause I didn't see any value in it) is integration test. By looking at what integration test is testing it's basically the same test as unit test with subtle difference that integration test doesn't go through OpModelInterface and unit test does. Please tell me if I missed something and I need to add it.
> 
> * @rpavlovicTT FYI

Thanks for the implementation.
Yes, you're not completely wrong. The test you added has to go through the implementation anyways.

One reason that we have two separate UTs for OpModel is that it's much easier to test an op through `TestOpModelLib`. For example, in your case, you can create a test suite and easily set `isDecodeMode` to both `true` and `false` with all other test params being the same. This is difficult to achieve through `TestOpModelInterface` (unless we create two separate ops or override the attributes of an existing op after testing it). 
This gives us flexibility for reproducing errors in the future if an op fails for some reason. 

Therefore, although it's not absolutely necessary to write tests in both files, we highly recommend to do so.
I'm approving this PR regardless.

rpavlovicTT (approved):
Approving OpModel changes

runtime/lib/ttnn/operations/transformer/rotary_embedding_llama.cpp
```diff
@@ -0,0 +1,42 @@
...
+
+namespace tt::runtime::ttnn::operations::transformer {
+static void
+runRotaryEmbeddingLlama(const ::tt::target::ttnn::RotaryEmbeddingLlamaOp *op,
+                        ProgramTensorPool &tensorPool) {
+  std::optional<::ttnn::MemoryConfig> outputMemoryConfig =
+      ::tt::runtime::ttnn::utils::createMemoryConfigIfNeeded(op->memcfg());
+
+  const ::ttnn::Tensor &in = tensorPool.getTTNNTensorAndValidate(op->input());
+  const ::ttnn::Tensor &cos_cache =
```
jnie-TT: nit: camelCase

runtime/lib/ttnn/operations/transformer/rotary_embedding_llama.cpp
```diff
@@ -0,0 +1,42 @@
...
+      tensorPool.getTTNNTensorAndValidate(op->cos_cache());
+  const ::ttnn::Tensor &sin_cache =
+      tensorPool.getTTNNTensorAndValidate(op->sin_cache());
```
jnie-TT, line 21: Same here and below

Issue #4641, opened 2025-08-25
[D2M] Handle mesh_shard op lowering and multi-device annotation in D2M pipeline


wooseokTT:
This feature depends on other issue in D2M pipeline from Wenbin, so need to wait for the PR to be merged. 

In this ticket, the mesh_shard op lowering in D2M pipeline along with annotating multi-device tensors should be handled.

Issue #2291, opened 2025-02-26
Remove MeshShardOp workarounds from TTNNLayout.cpp


sdjordjevicTT:
Once the folding for ToLayoutOp is implemented in the code, we can remove the workarounds for MeshShardOp in TTNNLayout.cpp:
```C++
// TTNN mesh shard expects host input and output
// TODO(#2102): This can be removed once the workaround pass can correctly
// handle cannonicalization of toLayout ops. Currently the workaround pass
// cannot detect redundant toLayout ops as a result of forcing the output
// layout and removing them.
if (mlir::isa<ttir::MeshShardOp>(op.getOperation())) {
  modified = changeLayoutToHost(op, operand, rewriter, isDPSResult);
  continue;
}
```

Issue #2102, opened 2025-02-04
[Workaround Pass] Redundant toLayout ops are not removed


jnie-TT:
After converting the output layout in the workaround pass, if there was a toLayout op that consumed the old layout, this toLayoutOp may be redundant after the workaround conversion. However this redundant toLayoutOp is not removed. To repro, remove the code block that references this issue in #1744 for mesh shard op:
```
// TTNN mesh shard expects host input and output
// TODO(#2102): This can be removed once the workaround pass can correctly
// handle cannonicalization of toLayout ops. Currently the workaround pass
// cannot detect redundant toLayout ops as a result of forcing the output
// layout and removing them.
if (mlir::isa<ttir::MeshShardOp>(op.getOperation())) {
  modified = changeLayoutToHost(op, operand, rewriter, isDPSResult);
  continue;
}
```

sdjordjevicTT:
Hi @jnie-TT, are there any compiler or silicon tests that fail if I remove this check?

sdjordjevicTT:
I see, I manged to repro the compiler test once I removed the above code, will take a look at it!

sdjordjevicTT:
Hi @jnie-TT, once I removed all the things in TTNNLayout.cpp regarding mesh shard op, and applied system memory input\output workaround, for the following test:
test/ttmlir/Dialect/TTNN/ccl/mesh_shard.mlir

I am getting the following IR after the workaround pass and applied canonicalization:
```mlir
// -----// IR Dump After Canonicalizer (canonicalize) //----- //
#device = #tt.device<workerGrid = #tt.grid<8x8, (d0, d1) -> (0, d0, d1)>, l1Map = (d0, d1)[s0, s1] -> (0, d0 floordiv s0, d1 floordiv s1, (d0 mod s0) * s1 + d1 mod s1), dramMap = (d0, d1)[s0, s1] -> (0, 0, ((((d0 floordiv s0) * 8 + d1 floordiv s1) * (s1 * s0) + (d0 mod s0) * s1 + d1 mod s1) floordiv 8192) mod 12, (((d0 floordiv s0) * 8 + d1 floordiv s1) * (s1 * s0) + (d0 mod s0) * s1 + d1 mod s1) floordiv 98304 + (((d0 floordiv s0) * 8 + d1 floordiv s1) * (s1 * s0) + (d0 mod s0) * s1 + d1 mod s1) mod 8192), meshShape = , chipIds = [0]>
[...]

PR #4925, opened 2025-09-12
Enable MeshShard op in D2M pipeline

wooseokTT:
### Ticket
https://github.com/tenstorrent/tt-mlir/issues/4641

### Problem description
For multi-device support in D2M, we need to enable mesh_shard op in D2M pipeline. 

### What's changed
1. Current mesh_shard op needs to be updated to non-dps, bufferizable op for proper handling in D2M pipeline. 
2. HostLayoutAttr is extended to incorporate TensorMesh info.
3. Annotate multi-device tensors mainly for host buffer allocation in D2M runtime.

### Checklist

### DIFF
--- a/lib/Dialect/TTNN/Transforms/TTNNLayout.cpp
+++ b/lib/Dialect/TTNN/Transforms/TTNNLayout.cpp
@@ -335,25 +335,27 @@ public:
 
     bool modified = false;
     Value input = op.getOperand();
-    Value result = op.getResult();
     Location newLoc = appendInputSuffix(op.getLoc(), 0);
     std::optional<Value> inputLayout = createToLayoutOp(
-        rewriter, newLoc, input, BufferType::SystemMemory, false /* tiled */);
+        rewriter, newLoc, input, BufferType::SystemMemory, /* tiled */ false);
     if (inputLayout.has_value()) {
       rewriter.modifyOpInPlace(op, [&]() { op->setOperand(0, *inputLayout); });
       modified = true;
     }
-    std::optional<Value> resultLayout = createToLayoutOp(
-        rewriter, newLoc, result, BufferType::SystemMemory, false /* tiled */);
-    if (resultLayout.has_value()) {
+
+    RankedTensorType resultType =
+        mlir::cast<RankedTensorType>(op.getResult().getType());
+    TTNNLayoutAttr newLayout =
+        createLayoutAttr(rewriter.getContext(), nullptr, resultType,
+                         BufferType::SystemMemory, /* isTiled */ false);
+    if (newLayout != resultType.getEncoding()) {
+      auto resultSystemMemoryType = RankedTensorType::get(
+          resultType.getShape(), resultType.getElementType(), newLayout);
       rewriter.modifyOpInPlace(
-          op, [&]() { op->getResult(0).setType(resultLayout->getType()); });
+          op, [&]() { op->getResult(0).setType(resultSystemMemoryType); });
       modified = true;
     }
-    if (modified) {
-      return success();
-    }
-    return failure();
+    return success(modified);
   }
 };
 } // namespace


include/ttmlir/Dialect/TTMetal/IR/TTMetalOps.td
```diff
@@ -73,6 +73,23 @@ def TTMetal_DeallocateBufferOp : TTMetal_Op<"deallocate_buffer", [MemoryEffects<
     let arguments = (ins AnyNon0RankedMemRef:$input);
 }
 
+def TTMetal_MeshShardOp : TTMetal_Op<"mesh_shard", [MemoryEffects<[MemFree]>]> {
```
nsmithtt: Should be MemAlloc no?

include/ttmlir/Target/TTMetal/binary.fbs
```diff
@@ -29,10 +29,17 @@ table TensorRef {
 }
 ///////////////////////////////
 
+table DeviceRef {
+  name: string;
+  mesh_shape: [int64];
+  mesh_offset: [int64];
+}
```
nsmithtt: Can we call this `MeshDesc` and put it under `include/ttmlir/Target/TTMetal/types.fbs`?

Note there is also `mesh_shape` on the `table Program` just below, I think we should probably remove that since it seems redundant now.  Or is that meant to represent something else?

lib/Target/TTMetal/TTMetalToFlatbuffer.cpp
```diff
@@ -409,12 +410,16 @@ memrefTypeToFlatbuffer(FlatbufferObjectCache &cache, MemRefType memref,
 
   std::vector<int32_t> hostStrides{};
   uint64_t hostVolume = 0;
+  std::string mesh = "";
```
nsmithtt: Maybe it's better to use `nullptr` for the sentinel case?  Might simplify the check on the runtime side.

lib/Target/TTMetal/TTMetalToFlatbuffer.cpp
```diff
@@ -815,6 +863,16 @@ static std::shared_ptr<void> translateModuleToFlatbuffer(
       }
     });
 
+    // We currently use the first mesh (default_device) for the device program.
+    ttcore::DeviceAttr deviceAttr = ttcore::lookupDevice(entry);
+    auto meshName = (meshes && !meshes.getMeshes().empty())
+                        ? meshes.getMeshes()[0].getName().str()
+                        : "";
```
nsmithtt: Same here, maybe nullptr?

include/ttmlir/Dialect/TTIR/Transforms/Passes.td
```diff
@@ -629,4 +629,14 @@ def TTIRFusing: Pass<"ttir-fusing", "::mlir::ModuleOp">
   ];
 }
 
+def TTIRMultiDeviceTensorAnnotation: Pass<"ttir-multi-device-tensor-annotation", "::mlir::ModuleOp">
+{
+  let summary = "TTIR multi-device tensor annotation pass.";
+  let description = [{
+    This pass walks through the graph with mesh and add multi-device tensor annotations to tensor encoding.
```
nsmithtt, line 646: Can you add mlir example snippet that show before/after this pass is applied?  Can be pseudo code

lib/Dialect/TTIR/Transforms/MultiDeviceTensorAnnotation.cpp
```diff
@@ -0,0 +1,154 @@
...
+            if (mlir::isa<mlir::func::ReturnOp>(op)) {
+              return mlir::WalkResult::skip();
+            }
+            if (auto meshShardOp =
+                    mlir::dyn_cast_if_present<mlir::tt::ttir::MeshShardOp>(
+                        op)) {
+              handleMeshShardOpAnnotation(builder, funcOp, meshShardOp,
+                                          meshAttr);
+            } else {
+              handleRestOpsForMultiDeviceTensorAnnotation(builder, op);
```
nsmithtt, line 132: Do we really need to annotate the rest of the graph?  It seems like this information gets quickly thrown away when we run TTIRToTTIRGeneric conversion.

nsmithtt:
Looks good!! Few comments inline ^^^

include/ttmlir/Dialect/TTMetal/IR/TTMetalOps.td
wooseokTT: Thanks for catching this... would it be better to specify memory effects with "MemRead, MemAlloc, MemWrite" as the mesh_shard op read from host memory, allocate new memory and write back to the new memory?

lib/Dialect/TTIR/Transforms/MultiDeviceTensorAnnotation.cpp
wooseokTT: I agree that the information is gone with the ttir generic ops. However, we cannot just annotate some part of the IRs as they fails to match types, and I was unable to propagate to early (as we iterate back from return op) tensors without the annotated intermediate tensor information. Some unnecessary annotation will be ignored and some necessary information will survive such as 
    %10 = ttir.empty() : tensor<512x512xf32, #ttcore.tensor_mesh<"mesh">>
and this information will be used for appropriate host storage selection.

include/ttmlir/Target/TTMetal/binary.fbs
wooseokTT: It seems that program->mesh_shape is being used to open the device using getProgramMeshShape() runtime API. Our metal programming model is trying to allow multiple mesh per deviceProgram with difference mesh descriptor, so we can see this new MeshDesc per device program as similar to submesh although it is not being used for now.

include/ttmlir/Dialect/TTIR/Transforms/Passes.td
wooseokTT: Added example of simple 2x4 add.

lib/Target/TTMetal/TTMetalToFlatbuffer.cpp
wooseokTT: Fixed.

lib/Target/TTMetal/TTMetalToFlatbuffer.cpp
wooseokTT: Fixed.

include/ttmlir/Target/TTMetal/binary.fbs
wooseokTT: Agreed. We may need to clean up this.

include/ttmlir/Dialect/TTMetal/IR/TTMetalOps.td
nsmithtt: Maybe it's just MemRead, MemAlloc.  MemWrite means that you're writing to an operand

lib/Dialect/TTIR/Transforms/MultiDeviceTensorAnnotation.cpp
nsmithtt: Instead of annotating the type, is it possible to just tag this information on the mesh_shard op?  This way we don't have to modify the whole graph.

include/ttmlir/Dialect/TTMetal/IR/TTMetalOps.td
wooseokTT: Got it. Thanks.

lib/Dialect/TTIR/Transforms/MultiDeviceTensorAnnotation.cpp
wooseokTT: MeshShard op issue was resolved by making mesh_shard op non-dps. We can get rid of the empty op for mesh_shard op and mesh_shard op can create host buffer by itself depending on shardtoFull or fulltoShard. The problem still remains for the simple 2x4 add test case in that toLayout op needs to allocate host buffers, which will be input to mesh_shard op, and we need to determine whether the buffer is single hostBuffer or distributedHostBuffer. For example, %10 will leads to hostAlloc and needed to be marked as multi-device tensor.

`  %10 = ttir.empty() : tensor<512x512xf32, #ttcore.tensor_mesh<"mesh">>
    %11 = ttir.to_layout %9, %10 : tensor<8x8x2x2x!ttcore.tile<32x32, f32>, #ttcore.metal_layout<logical_shape = 512x512, dim_alignments = 256x256, collapsed_intervals = dense<[[0, 1], [1, 2]]> : tensor<2x2xi64>, undef, l1>> into tensor<512x512xf32, #ttcore.tensor_mesh<"mesh">> -> tensor<512x512xf32, #ttcore.tensor_mesh<"mesh">>
    %12 = "ttir.mesh_shard"(%11) <{shard_dims = array<i64: 0, 1>, shard_direction = #ttcore.shard_direction<shard_to_full>, shard_shape = array<i64: 2, 4>, shard_type = #ttcore.shard_type<devices>}> : (tensor<512x512xf32, #ttcore.tensor_mesh<"mesh">>) -> tensor<1024x2048xf32>
    return %12 : tensor<1024x2048xf32>`

Without this, we may need to traverse/guess to determine the correct buffer type or force to use one type of buffer. I believe the former one is error prone while the latter one is restrictive to users. Let's discuss in person. I will try to figure out better approach as well.

include/ttmlir/Target/TTMetal/binary.fbs
wooseokTT: As described, program->mesh_shape is used by ttrt to open the parent mesh and now each device program can open submesh based on the mesh inside device program.

lib/Conversion/StableHLOToTTIR/ShardyToTTIRPatterns.cpp
```diff
@@ -196,9 +196,8 @@ class ShardyToTTIRManualComputationOpConversionPattern
       // Create a new mesh shard op.
       auto outputType = mlir::cast<mlir::RankedTensorType>(
           getTypeConverter()->convertType(localArgType));
-      auto meshShardOp = ttir::utils::createDPSOp<mlir::tt::ttir::MeshShardOp>(
-          rewriter, loc, outputType, globalOperand,
-          shardyMeshSharding->getShardType(),
+      auto meshShardOp = rewriter.create<mlir::tt::ttir::MeshShardOp>(
```
mmanzoorTT, line 199: Any specific reason for not using the utility function?

mmanzoorTT: never mind. got it.

jnie-TT (approved):

lib/Dialect/TTNN/Transforms/TTNNLayout.cpp
```diff
@@ -345,6 +316,48 @@ class TTNNLayoutHoistedFuncCallRewriter
...
+    if (resultLayout.has_value()) {
+      rewriter.modifyOpInPlace(
+          op, [&]() { op->getResult(0).setType(resultLayout->getType()); });
+      modified = true;
+    }
+    if (modified) {
+      return success();
+    }
+    return failure();
+  }
```
azecevic000: Nit: `return success(modified);`

lib/Dialect/TTNN/Transforms/TTNNLayout.cpp
```diff
@@ -345,6 +316,48 @@ class TTNNLayoutHoistedFuncCallRewriter
...
+    if (!shouldMeshShardOpForceSystemMemory(op.getOperation())) {
+      return failure();
+    }
+
+    bool modified = false;
+    Value input = op.getOperand();
+    Value result = op.getResult();
+    Location newLoc = appendInputSuffix(op.getLoc(), 0);
+    std::optional<Value> inputLayout = createToLayoutOp(
+        rewriter, newLoc, input, BufferType::SystemMemory, false /* tiled */);
```
azecevic000: Nit:  `/*tiled=*/false` style is the preferred one https://docs.tenstorrent.com/tt-mlir/coding-guidelines.html#comments

lib/Dialect/TTNN/Transforms/TTNNLayout.cpp
```diff
@@ -345,6 +316,48 @@ class TTNNLayoutHoistedFuncCallRewriter
...
+      rewriter.modifyOpInPlace(op, [&]() { op->setOperand(0, *inputLayout); });
+      modified = true;
+    }
+    std::optional<Value> resultLayout = createToLayoutOp(
+        rewriter, newLoc, result, BufferType::SystemMemory, false /* tiled */);
+    if (resultLayout.has_value()) {
+      rewriter.modifyOpInPlace(
+          op, [&]() { op->getResult(0).setType(resultLayout->getType()); });
+      modified = true;
+    }
```
azecevic000: Is the idea here to force result layout to system memory and relying on `to_layout` folder to remove redundant `to_layout` op? If it is, it should work, but it's a very hacky way of doing it, why wouldn't you just create a system memory TTNNLayout and set result encoding with it?

lib/Dialect/TTNN/Transforms/TTNNLayout.cpp
```diff
@@ -345,6 +316,48 @@ class TTNNLayoutHoistedFuncCallRewriter
...
+  TTNNLayoutMeshShardRewriter(MLIRContext *ctx)
+      : OpRewritePattern<ttir::MeshShardOp>(ctx) {}
+  // Match and rewrite the MeshShardOp.
+  LogicalResult matchAndRewrite(ttir::MeshShardOp op,
+                                PatternRewriter &rewriter) const override {
+    // TTNN mesh shard expects host input and output
+    // TODO(#2291): This can be removed once the workaround pass can correctly
+    // handle canonicalization of toLayout ops (#2102). Currently the
+    // workaround pass cannot detect redundant toLayout ops as a result of
+    // forcing the output layout and removing them.
```
azecevic000, line 331: This issue seems like it's resolved, did you try with a workaround approach?

wooseokTT: I believe this is still open (https://github.com/tenstorrent/tt-mlir/issues/2291), so didn't try them in my side. Hmm, should I explore more for the issue?

lib/Dialect/TTNN/Transforms/TTNNLayout.cpp
wooseokTT: Thanks. It will be fixed.

lib/Dialect/TTNN/Transforms/TTNNLayout.cpp
wooseokTT: Thanks. It will be fixed.

lib/Dialect/TTNN/Transforms/TTNNLayout.cpp
wooseokTT: I wanted to reuse the function such that I can get away with creating TTNNLayout by myself as you pointed. I will try to fix such hacky way.

wooseokTT: Updated. Let me know if anything further needs to be updated.

lib/Dialect/TTNN/Transforms/TTNNLayout.cpp
azecevic000: That one is still open, but the main one (#2102) is merged, so I believe it can be done in the workaround pass now. I'm okay with this staying as-is in this PR, since this was already the case before this PR and it's not a major point of this PR.

azecevicTT (approved):

lib/Dialect/TTIR/Transforms/MultiDeviceTensorAnnotation.cpp
wooseokTT: We will stick to current approach for now and see if there is any issues. Later if there is any issue, we can update it as Nick suggested.

PR #4927, opened 2025-09-12
Adding _force_graph_level_check API to builder

jgrimTT:
### Ticket
Closes [#4926 ](https://github.com/tenstorrent/tt-mlir/issues/4926)

### Problem description
Graph level checks may cause issues in StableHLOBuilder tests.

### What's changed
Disabling graph level checks in StableHLOBuilder and adding an option to enable them.

### Checklist


tools/builder/base/builder.py
```diff
@@ -121,7 +122,10 @@ def golden_map(self) -> Dict[str, Dict[int, GoldenTensor]]:
```
tapspatel, line 122: we should exit early right after adding outputs
```
if force graph level check:
    return golden_map
```

test/python/golden/test_shardy_ops_n300.py
```diff
@@ -0,0 +1,69 @@
...
+    )
+
+    builder.sharding_constraint(in0, tensor_sharding_attr=tensor_sharding_attr)
+    return builder.add(in0, in1, unit_attrs=unit_attrs)
+
+
+@pytest.mark.parametrize("shape", [(128, 128)], ids=shape_str)
+@pytest.mark.parametrize("dtype", [torch.float32], ids=["f32"])
+@pytest.mark.parametrize("mesh_shape", [(1, 2)])
+@pytest.mark.parametrize(
```
tapspatel, line 47: any reason we don't have 
```
@pytest.mark.parametrize("target", ["ttnn", "ttmetal"])
```

do these tests compile into flatbuffer with output goldens and run in ttrt?

jgrimTT: They do compile into flatbuffers with output/graph level goldens and run ttrt. I was getting a `Failed to run pass manager` for the n300 `target="ttmetal"` tests. I was unsuccessful debugging it, I can make an issue or dig deeper.

jgrimTT:
> nothing blocking, just general comment about making sure the tests runs e2e and still does output pcc evaluation

They do, thx!

tools/builder/base/builder.py
jgrimTT: Oh that makes sense! To clarify, did you want this under the outputs or operands for loop?

test/python/golden/test_shardy_ops_n300.py
tapspatel: can we just keep ttnn, I expect metal to fail multichip tests

tools/builder/base/builder.py
tapspatel: right after we add outputs into golden map, it should exit early

Issue #4255, opened 2025-07-30
Add ttnn.experimental.nlp_concat_heads op to TTNN IR


sdjordjevicTT:
## Summary
Add support for `ttnn.experimental.nlp_concat_heads` operation to TTNN IR as part of LLM-specific ops onboarding for Llama3.2 architecture.

## Description
This op concatenates attention heads for NLP operations, which is essential for transformer-based models.

## Implementation Requirements
- [ ] Define the op in TTNN dialect
- [ ] Implement op lowering/conversion
- [ ] Add fusing patterns (investigation needed)
- [ ] Add unit tests
- [ ] Add integration tests
- [ ] Update documentation

## Related Issues
- Parent issue: #4214 (LLM-specific ops onboarding)

## Architecture
- Target: Llama3.2 architecture
- Category: Experimental NLP operations

## Notes
- Fusing patterns will be needed for performance optimization
- Part of broader LLM ops initiative

mtopalovicTT:
Op requirements for workarounds:

* input must be on device
* supported dtypes - f32/bf16/bfp8_b
* supported layout - tile
* non sharded inputs must be interleaved
* sharded input
    * memory layout on input tensor cannot be width sharded
    * memory layout on output cannot be height sharded

PR #4937, opened 2025-09-15
[TTNN Dialect] nlp_concat_heads op

mtopalovicTT:
Adding nlp_concat_heads op in ttnn dialect. This op seems redudant at first as it can be implemented as standard concat_heads + unsqueeze but it's special case (which seems to be frequent) in LLMs where `[B, head_size, seq_len, head_dim]` into `[B, 1, seq_len, head_size * head_dim]`.

There will be followup PR after this which will capture workarounds needed to run this op in ttnn (see https://github.com/tenstorrent/tt-mlir/issues/4255)

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -514,7 +514,7 @@ jobs:
     - name: Download tt-alchemist wheel
       uses: actions/download-artifact@v4
       with:
-        name: tt-alchemist-whl-tracy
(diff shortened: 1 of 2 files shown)


sgholamiTT (approved):
OpModel code looks good to me.
Thanks for doing this.

jnie-TT (approved):

rpavlovicTT (approved):

azecevicTT (approved):

Issue #4316, opened 2025-08-01
Remove resetInstance() from OpModel tests


ckaravasilisTT:
[TestOpModelLib.cpp](https://github.com/tenstorrent/tt-mlir/blob/main/test/unittests/OpModel/TTNN/Lib/TestOpModelLib.cpp) and [TestOpModelInterface.cpp](https://github.com/tenstorrent/tt-mlir/blob/main/test/unittests/OpModel/TTNN/Op/TestOpModelInterface.cpp) contain a number of `SingletonDeviceContext::resetInstance();`. We should gradually remove those and investigate any issues (mostly hangs) that might arise.

xanderchin:
@gfengTT to supply branch with failing test

gfengTT:
metal branch: https://github.com/tenstorrent/tt-metal/tree/gfeng/test-20250721
issue: https://github.com/tenstorrent/tt-metal/issues/25772

PR #4969, opened 2025-09-16
Remove most device resets in OpModel tests

ckaravasilisTT:
### Ticket
#4316 and [issue on metal](https://github.com/tenstorrent/tt-metal/issues/25772)

### Problem description
`getOpRuntime()` (sometimes in combination with `getOpConstraints()` )  hangs. We have used `resetInstance()` as a temporary workaround until issues are fixed on metal. We want to gradually remove these resets when tests can run without hangs.

### What's changed
Removed almost all `resetInstance()` from `TestOpModelLib.cpp` and `TestOpModelInterface.cpp`. Right now it seems only `Reshape` is causing an issue.

### Testing
Ran CI tests 4 times. Offline testing with`--gtest_shuffle --gtest_repeat=2`.


arminaleTT:
Please also include the details of local testing (looping, shuffling, etc.)

PR #4978, opened 2025-09-16
D2M/TTNN Integration: Convert TTNN to TTIR pass 

phizalev-TT:
### Ticket
[Link to Github Issue](https://github.com/tenstorrent/tt-mlir/issues/4763)

### Problem description
Convert TTNN operations targged with `ttnn.hoist_generic_via_d2m` attribute to TTIR. Limit support to unary and binary element-wise operations. See [design doc](https://docs.google.com/document/d/1SyJyK7yYPvMDzm0eBe89vrjhA5ZkIFc2_Lw2gOIarhY/edit?tab=t.0).

### What's changed
- Add ConvertTTNNToTTIR pass;
- Add MLIR test for unary and binary op.

### DIFF
--- a/include/ttmlir/Conversion/Passes.td
+++ b/include/ttmlir/Conversion/Passes.td
@@ -205,5 +205,10 @@ def ConvertTTIRToLinalg: Pass<"convert-ttir-to-linalg", "::mlir::ModuleOp"> {
   let dependentDialects = ["mlir::tt::ttir::TTIRDialect", "mlir::linalg::LinalgDialect", "mlir::tosa::TosaDialect"];
 }
 
+def ConvertTTNNToTTIR : Pass<"convert-ttnn-to-ttir", "::mlir::ModuleOp"> {
+let summary = "Convert TTNN dialect to TTIR dialect.";
+  let constructor = "createConvertTTNNToTTIRPass()";
+  let dependentDialects = ["mlir::tt::ttnn::TTNNDialect", "mlir::tt::ttir::TTIRDialect"];
+}
 
 #endif // TTMLIR_CONVERSION_PASSES
--- a/include/ttmlir/Conversion/TTNNToTTIR/TTNNToTTIR.h
+++ b/include/ttmlir/Conversion/TTNNToTTIR/TTNNToTTIR.h
@@ -1,0 +1,27 @@
+// SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#ifndef TTMLIR_CONVERSION_TTNNTOTTIR_TTNNTOTTIR_H
+#define TTMLIR_CONVERSION_TTNNTOTTIR_TTNNTOTTIR_H
+
+#include "mlir/Pass/Pass.h"
+#include "mlir/Transforms/DialectConversion.h"
+
+namespace mlir::tt {
+
+namespace ttir {
+
+#define GEN_PASS_DECL_CONVERTTTNNTOTTIR
+#include "ttmlir/Conversion/Passes.h.inc"
+
+} // namespace ttir
+
+void populateTTNNToTTIRPatterns(MLIRContext *ctx, RewritePatternSet &patterns,
+                                TypeConverter &typeConverter);
+
+std::unique_ptr<OperationPass<ModuleOp>> createConvertTTNNToTTIRPass();
+
+} // namespace mlir::tt
+
+#endif // TTMLIR_CONVERSION_TTNNTOTTIR_TTNNTOTTIR_H
--- a/include/ttmlir/Conversion/TTNNToTTIR/TTNNToTTIR.h
+++ b/include/ttmlir/Conversion/TTNNToTTIR/TTNNToTTIR.h
@@ -10,12 +10,12 @@
 
 namespace mlir::tt {
 
-namespace ttir {
+namespace ttnn {
 
 #define GEN_PASS_DECL_CONVERTTTNNTOTTIR
 #include "ttmlir/Conversion/Passes.h.inc"
 
-} // namespace ttir
+} // namespace ttnn
 
 void populateTTNNToTTIRPatterns(MLIRContext *ctx, RewritePatternSet &patterns,
                                 TypeConverter &typeConverter);
--- a/lib/Conversion/TTNNToTTIR/CMakeLists.txt
+++ b/lib/Conversion/TTNNToTTIR/CMakeLists.txt
@@ -1,6 +1,3 @@
-include_directories(${TTMLIR_SOURCE_DIR}/include)
-include_directories(${PROJECT_SOURCE_DIR}/include)
-
 add_mlir_conversion_library(TTMLIRTTNNToTTIR
   TTNNToTTIRPass.cpp
   TTNNToTTIRPatterns.cpp
--- a/lib/Conversion/TTNNToTTIR/CMakeLists.txt
+++ b/lib/Conversion/TTNNToTTIR/CMakeLists.txt
@@ -1,0 +1,20 @@
+include_directories(${TTMLIR_SOURCE_DIR}/include)
+include_directories(${PROJECT_SOURCE_DIR}/include)
+
+add_mlir_conversion_library(TTMLIRTTNNToTTIR
+  TTNNToTTIRPass.cpp
+  TTNNToTTIRPatterns.cpp
+
+  ADDITIONAL_HEADER_DIRS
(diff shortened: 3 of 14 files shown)


phizalev-TT:
@nsmithtt do we want to add tag attribute in TTNN dialect that will mark ops for selective conversion?

nsmithtt:
> @nsmithtt do we want to add tag attribute in TTNN dialect that will mark ops for selective conversion?

Yeah that's probably a good idea.  Then optimizer just tags the ones it wants to go through generic.  We might be able to use `target.addDynamicallyLegalOp([](Operation* op) { return op->hasAttr("the tag"); });`, but at first glance addDynamicallyLegalOp is templated so it's unclear without trying if it'll be happy to accept the Operation base class.

vtangTT (commented):
Not blocking, just wondering about the tests: is it redundant to test the conversion for every unary/binary op pattern that's been added.

If not, thoughts on adopting a repo structure similar to StableHLOToTTIR or TosaToTTIR. If we do, I think we should put the `skip_binary_op.mlir` in the same file as `binary_op.mlir` just to reduce # of files and same w/ unary

phizalev-TT:
@vtangTT looks like by testing one of each kind we've achieved code coverage. What would be the value of exhaustive tests in addition to it?

vtangTT:
> @vtangTT looks like by testing one of each kind we've achieved code coverage. What would be the value of exhaustive tests in addition to it?

Oh, didn't take into account the coverage report. In that case yea, seems redundant to me.

lib/Conversion/TTNNToTTIR/TTNNToTTIRPass.cpp
```diff
@@ -0,0 +1,88 @@
...
+
+  ConvertTTNNToTTIRPass() = default;
+
+  void runOnOperation() final {
+    mlir::ConversionTarget target(getContext());
+
+    // Common legal/illegal ops/dialects for both partial and full conversion.
+    target.addLegalDialect<ttir::TTIRDialect>();
+    target.addLegalDialect<ttcore::TTCoreDialect>();
+    target.addLegalOp<mlir::tt::ttir::EmptyOp>();
```
svuckovicTT: Is the empty op needed here?

lib/Conversion/TTNNToTTIR/TTNNToTTIRPatterns.cpp
```diff
@@ -0,0 +1,168 @@
...
+      typeConverter, ctx);
+  patterns.add<TTNNToTTIROpDefaultConversionPattern<mlir::tt::ttnn::TanOp,
+                                                    mlir::tt::ttir::TanOp>>(
+      typeConverter, ctx);
+  patterns.add<TTNNToTTIROpDefaultConversionPattern<mlir::tt::ttnn::TanhOp,
+                                                    mlir::tt::ttir::TanhOp>>(
+      typeConverter, ctx);
+  patterns.add<TTNNToTTIROpDefaultConversionPattern<mlir::tt::ttnn::LogOp,
+                                                    mlir::tt::ttir::LogOp>>(
+      typeConverter, ctx);
```
svuckovicTT: Suggestion:

Should be able to put these together like so, I find that it increases readability:
```
   patterns.add<TTNNToTTIROpDefaultConversionPattern<mlir::tt::ttnn::AbsOp,
                                                    mlir::tt::ttir::AbsOp>,
                TTNNToTTIROpDefaultConversionPattern<mlir::tt::ttnn::CbrtOp,
                                                    mlir::tt::ttir::CbrtOp>,
...
                (typeConverter, ctx);
```

lib/Dialect/TTNN/Utils/Utils.cpp
```diff
@@ -13,6 +13,7 @@
 #include "mlir/IR/Value.h"
 #include "llvm/Support/Casting.h"
 
+#include <mlir/IR/Operation.h>
```
svuckovicTT: formatting

include/ttmlir/Conversion/Passes.td
```diff
@@ -205,5 +205,10 @@ def ConvertTTIRToLinalg: Pass<"convert-ttir-to-linalg", "::mlir::ModuleOp"> {
   let dependentDialects = ["mlir::tt::ttir::TTIRDialect", "mlir::linalg::LinalgDialect", "mlir::tosa::TosaDialect"];
 }
 
+def ConvertTTNNToTTIR : Pass<"convert-ttnn-to-ttir", "::mlir::ModuleOp"> {
+let summary = "Convert TTNN dialect to TTIR dialect.";
+  let constructor = "createConvertTTNNToTTIRPass()";
+  let dependentDialects = ["mlir::tt::ttnn::TTNNDialect", "mlir::tt::ttir::TTIRDialect"];
+}
```
svuckovicTT, line 212: Alt Text

include/ttmlir/Conversion/TTNNToTTIR/TTNNToTTIR.h
```diff
@@ -0,0 +1,27 @@
...
+
+#ifndef TTMLIR_CONVERSION_TTNNTOTTIR_TTNNTOTTIR_H
+#define TTMLIR_CONVERSION_TTNNTOTTIR_TTNNTOTTIR_H
+
+#include "mlir/Pass/Pass.h"
+#include "mlir/Transforms/DialectConversion.h"
+
+namespace mlir::tt {
+
+namespace ttir {
```
azecevic000: IMO it makes more sense to put this in `ttnn` namespace.

lib/Conversion/TTNNToTTIR/CMakeLists.txt
```diff
@@ -0,0 +1,20 @@
+include_directories(${TTMLIR_SOURCE_DIR}/include)
+include_directories(${PROJECT_SOURCE_DIR}/include)
```
azecevic000: Is this necessary?

lib/Conversion/TTNNToTTIR/TTNNToTTIRPass.cpp
```diff
@@ -0,0 +1,88 @@
...
+#include "mlir/IR/BuiltinOps.h"
+#include "mlir/IR/BuiltinTypes.h"
+#include "mlir/IR/Dialect.h"
+#include "mlir/IR/MLIRContext.h"
+#include "mlir/IR/PatternMatch.h"
+#include "mlir/Pass/Pass.h"
+#include "mlir/Transforms/DialectConversion.h"
+
+using namespace mlir;
+using namespace mlir::tt;
```
azecevic000: You can remove this.

lib/Conversion/TTNNToTTIR/TTNNToTTIRPatterns.cpp
```diff
@@ -0,0 +1,168 @@
...
+#include "mlir/IR/Builders.h"
+#include "mlir/IR/BuiltinAttributes.h"
+#include "mlir/IR/BuiltinTypes.h"
+#include "mlir/IR/PatternMatch.h"
+#include "mlir/Support/LLVM.h"
+#include "mlir/Support/LogicalResult.h"
+#include "mlir/Transforms/DialectConversion.h"
+
+using namespace mlir;
+using namespace mlir::tt;
```
azecevic000: Same here.

lib/Conversion/TTNNToTTIR/TTNNToTTIRPatterns.cpp
```diff
@@ -0,0 +1,168 @@
...
+
+namespace {
+template <typename SrcOp, typename DestOp,
+          typename Adaptor = typename SrcOp::Adaptor>
```
azecevic000: Nit: We always want `Adaptor` to be `SrcOp::Adaptor`, there isn't a point for it to be a template parameter and it can result in an error. A better choice would be `using Adaptor = typename SrcOp::Adaptor;` in the class definition.

lib/Conversion/TTNNToTTIR/TTNNToTTIRPatterns.cpp
```diff
@@ -0,0 +1,168 @@
...
+class TTNNToTTIROpDefaultConversionPattern : public OpConversionPattern<SrcOp> {
```
azecevic000: I wouldn't call this `DefaultConversionPattern` for a simple reason that a truly default conversion pattern would also have to take attributes into account. I think `ElementwiseConversionPattern` is more suitable, I don't think we will use it for any other op/class of ops (even if we could do it, creating a separate conversion pattern is probably worth it).

azecevicTT (approved):
Thank you for addressing comments, LGTM.

nsmithtt (approved):
LGTM!

PR #4982, opened 2025-09-17
[L1InterleavedFallbackAnalysis] Improve perf by skipping L1 upgrades for reshape ops that can be optimized in runtime

bmalesevicTT:
This PR enhances the L1 interleaved fallback analysis by skipping reshape operation inputs/outputs that can be optimized to zero-cost view operations at runtime. The analysis now implements TTNN's reshape-to-view optimization logic to predict when reshapes would be converted to views if they stayed in DRAM, avoiding unnecessary L1 memory allocation for operations that didn't benefit from it or even caused perf degradation.

The new `checkReshapeSkip` function evaluates reshape operations using the same criteria as TTNN's runtime optimization:
- First checks if input and output tensors have identical shapes (no-op detection)
- For non-trivial reshapes, validates view optimization conditions:
  - Last dimension preservation for memory layout compatibility  
  - Memory layout consistency (sharded/interleaved matching)
  - Memory buffer consistency (DRAM/L1)
  - At least one of: tile alignment (32x32), second-to-last dimension preservation, or row-major input

This prevents the analysis from upgrading reshape operations to L1 when they will be optimized away at the tt-metal runtime level.

### What's changed
- Performance Impact: 
  - Prevents performance degradation from unnecessary reshapes being kept as regular ops in runtime
  - Reduces unnecessary L1 memory pressure by skipping reshape operations that become zero-cost views
  - With future changes, might show to enable more beneficial operations to utilize available L1 memory for actual performance gains
- No API changes
### Checklist
- [x] Test perf for all models with newest changes, on Forge-FE (when python3.11 upgraded on CI)

### DIFF
--- a/include/ttmlir/Dialect/TTNN/Analysis/L1InterleavedFallbackAnalysis.h
+++ b/include/ttmlir/Dialect/TTNN/Analysis/L1InterleavedFallbackAnalysis.h
@@ -7,6 +7,7 @@
 
 #include "ttmlir/Dialect/TTNN/Analysis/OpConfig.h"
 #include "ttmlir/Dialect/TTNN/Analysis/TTNNAnalysis.h"
+#include "ttmlir/Dialect/TTNN/IR/TTNNOps.h"
 #include "ttmlir/Dialect/TTNN/IR/TTNNOpsAttrs.h"
 
 #include "mlir/Dialect/Func/IR/FuncOps.h"
@@ -124,7 +125,7 @@ private:
   // row-major input
   //
   // Called by handleReshapeOps() if conditions 3&4 are met.
-  bool checkReshapeSkip(Operation *reshapeOperation) const;
+  bool checkReshapeSkip(ReshapeOp reshapeOp) const;
 
   // Try to upgrade an operation to L1 interleaved layout by testing available
   // L1 configurations and selecting the first one that passes validation.
--- a/include/ttmlir/Dialect/TTNN/Analysis/L1InterleavedFallbackAnalysis.h
+++ b/include/ttmlir/Dialect/TTNN/Analysis/L1InterleavedFallbackAnalysis.h
@@ -88,8 +88,9 @@ private:
   // (1x1 kernel, stride=1, padding=0, dilation=1)
   bool isConv2DConvertibleToMatMul(Operation *op);
 
-  // Check if a reshape operation should be skipped based on TTNN optimization
+  // Check if a reshape operation should be skipped based on tt-metal TTNN optimization
   // rules. Returns true if the operation should be skipped, false otherwise.
+  // reference: ttnn::operations::data_movement::ReshapeViewOperation::invoke
   // Parameters:
   // - reshapeOp: The reshape operation to analyze
   // - contextOp: The operation context for logging (producer op for user
--- a/include/ttmlir/Dialect/TTNN/Analysis/L1InterleavedFallbackAnalysis.h
+++ b/include/ttmlir/Dialect/TTNN/Analysis/L1InterleavedFallbackAnalysis.h
@@ -90,13 +90,21 @@ private:
 
   // Check if a reshape operation should be skipped based on tt-metal TTNN
   // optimization rules. Returns true if the operation should be skipped, false
-  // otherwise. reference:
+  // otherwise.
+  //
+  // This check is performed in two contexts:
+  // 1. When checking the reshape operation itself - upgrading its output tensor
+  //    would change buffer type, potentially breaking tt-metal's view
+  //    optimization
+  // 2. Before that, alongside checking reshape's predecessor, producer of its
+  //    input tensor - upgrading the producer would change the
+  //    buffer type of reshape's input tensor, again potentially breaking
+  //    tt-metal's view optimization.
+  //
+  // reference:
   // ttnn::operations::data_movement::ReshapeViewOperation::invoke
-  // Parameters:
-  // - reshapeOp: The reshape operation to analyze
-  // - isUserOp: true if this is a user reshape check, false for direct reshape
   // TODO(bmalesevic,#5086): replace to dynamic check when tt-metal fixed
-  bool checkReshapeSkip(Operation *reshapeOperation, bool isUserOp) const;
+  bool checkReshapeSkip(Operation *reshapeOperation) const;
 
   // Try to upgrade an operation to L1 interleaved layout by testing available
   // L1 configurations and selecting the first one that passes validation.
--- a/include/ttmlir/Dialect/TTNN/Analysis/L1InterleavedFallbackAnalysis.h
+++ b/include/ttmlir/Dialect/TTNN/Analysis/L1InterleavedFallbackAnalysis.h
@@ -91,13 +91,12 @@ private:
   // Check if a reshape operation should be skipped based on tt-metal TTNN
   // optimization rules. Returns true if the operation should be skipped, false
   // otherwise. reference:
-  // ttnn::operations::data_movement::ReshapeViewOperation::invoke Parameters:
+  // ttnn::operations::data_movement::ReshapeViewOperation::invoke
+  // Parameters:
   // - reshapeOp: The reshape operation to analyze
-  // - contextOp: The operation context for logging (producer op for user
-  // reshapes)
   // - isUserOp: true if this is a user reshape check, false for direct reshape
-  bool checkReshapeSkip(Operation *reshapeOperation, Operation *contextOp,
-                        bool isUserOp) const;
+  // TODO(bmalesevic,#5086): replace to dynamic check when tt-metal fixed
+  bool checkReshapeSkip(Operation *reshapeOperation, bool isUserOp) const;
 
   // Try to upgrade an operation to L1 interleaved layout by testing available
   // L1 configurations and selecting the first one that passes validation.
--- a/include/ttmlir/Dialect/TTNN/Analysis/L1InterleavedFallbackAnalysis.h
+++ b/include/ttmlir/Dialect/TTNN/Analysis/L1InterleavedFallbackAnalysis.h
@@ -88,6 +88,16 @@ private:
   // (1x1 kernel, stride=1, padding=0, dilation=1)
   bool isConv2DConvertibleToMatMul(Operation *op);
 
+  // Check if a reshape operation should be skipped based on TTNN optimization
+  // rules. Returns true if the operation should be skipped, false otherwise.
+  // Parameters:
+  // - reshapeOp: The reshape operation to analyze
+  // - contextOp: The operation context for logging (producer op for user
+  // reshapes)
+  // - isUserOp: true if this is a user reshape check, false for direct reshape
+  bool checkReshapeSkip(Operation *reshapeOperation, Operation *contextOp,
+                        bool isUserOp) const;
+
   // Try to upgrade an operation to L1 interleaved layout by testing available
   // L1 configurations and selecting the first one that passes validation.
   void tryUpgradeToL1Interleaved(Operation *op);
--- a/include/ttmlir/Dialect/TTNN/Analysis/L1InterleavedFallbackAnalysis.h
+++ b/include/ttmlir/Dialect/TTNN/Analysis/L1InterleavedFallbackAnalysis.h
@@ -88,10 +88,10 @@ private:
   // (1x1 kernel, stride=1, padding=0, dilation=1)
   bool isConv2DConvertibleToMatMul(Operation *op);
 
-  // Check if a reshape operation should be skipped based on tt-metal TTNN optimization
-  // rules. Returns true if the operation should be skipped, false otherwise.
-  // reference: ttnn::operations::data_movement::ReshapeViewOperation::invoke
-  // Parameters:
+  // Check if a reshape operation should be skipped based on tt-metal TTNN
+  // optimization rules. Returns true if the operation should be skipped, false
+  // otherwise. reference:
+  // ttnn::operations::data_movement::ReshapeViewOperation::invoke Parameters:
   // - reshapeOp: The reshape operation to analyze
   // - contextOp: The operation context for logging (producer op for user
   // reshapes)
--- a/include/ttmlir/Dialect/TTNN/Analysis/L1InterleavedFallbackAnalysis.h
+++ b/include/ttmlir/Dialect/TTNN/Analysis/L1InterleavedFallbackAnalysis.h
@@ -88,22 +88,42 @@ private:
   // (1x1 kernel, stride=1, padding=0, dilation=1)
   bool isConv2DConvertibleToMatMul(Operation *op);
 
-  // Check if a reshape operation should be skipped based on tt-metal TTNN
-  // optimization rules. Returns true if the operation should be skipped, false
-  // otherwise.
+  // Check if a reshape operation should be skipped to
+  // preserve tt-metal's view optimization. Returns true if the operation should
+  // be skipped.
   //
-  // This check is performed in two contexts:
-  // 1. When checking the reshape operation itself - upgrading its output tensor
+  // Checks two scenarios:
+  // 1. Checking the reshape operation itself - upgrading its output tensor
   //    would change buffer type, potentially breaking tt-metal's view
   //    optimization
-  // 2. Before that, alongside checking reshape's predecessor, producer of its
-  //    input tensor - upgrading the producer would change the
-  //    buffer type of reshape's input tensor, again potentially breaking
-  //    tt-metal's view optimization.
+  // 2. Checking if the operation has a reshape consumer - upgrading the
+  //    producer would change the buffer type of reshape's input tensor,
+  //    again potentially breaking tt-metal's view optimization.
+  //
+  // Conditions reshape can be optimized to a view operation (based on tt-metal
+  // logic):
+  // 1. Last dimension must be the same
+  // 2. Either second-to-last dimension is the same OR no tile padding OR
+  // row-major input
+  // 3. Sharded/Interleaved must match for output and input
+  // 4. L1/DRAM must match for output and input
   //
   // reference:
   // ttnn::operations::data_movement::ReshapeViewOperation::invoke
-  // TODO(bmalesevic,#5086): replace to dynamic check when tt-metal fixed
+  // TODO(bmalesevic,#5086): replace with dynamic runtime check when tt-metal
+  // provides API to query view optimization eligibility at compile time
+  bool handleReshapeOps(Operation *op) const;
+
+  // Check if a reshape operation should be skipped based on remaining tt-metal
+  // TTNN optimization rules. Returns true if the operation should be skipped,
+  // false otherwise.
+  //
+  // Tests the geometric conditions for view optimization:
+  // 1. Last dimension must be the same
+  // 2. Either second-to-last dimension is the same OR no tile padding OR
+  // row-major input
+  //
+  // Called by handleReshapeOps() if conditions 3&4 are met.
   bool checkReshapeSkip(Operation *reshapeOperation) const;
 
   // Try to upgrade an operation to L1 interleaved layout by testing available
--- a/lib/Dialect/TTNN/Analysis/L1InterleavedFallbackAnalysis.cpp
+++ b/lib/Dialect/TTNN/Analysis/L1InterleavedFallbackAnalysis.cpp
@@ -249,16 +249,23 @@ bool L1InterleavedFallbackAnalysis::checkReshapeSkip(
     const int64_t tileHeight = 32; // tt::constants::TILE_HEIGHT
     const int64_t tileWidth = 32;  // tt::constants::TILE_WIDTH
 
-    bool inputTensorSharded = false;
-    if (auto ttnnLayout =
-            mlir::dyn_cast<TTNNLayoutAttr>(inputType.getEncoding())) {
-      inputTensorSharded = ttnnLayout.hasShardedL1TensorMemoryLayout();
+    // Checking if the other tensor is sharded/interleaved:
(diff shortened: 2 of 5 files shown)


include/ttmlir/Dialect/TTNN/Analysis/L1InterleavedFallbackAnalysis.h
```diff
@@ -88,6 +88,17 @@ class L1InterleavedFallbackAnalysis
   // (1x1 kernel, stride=1, padding=0, dilation=1)
   bool isConv2DConvertibleToMatMul(Operation *op);
 
+  // Check if a reshape operation should be skipped based on tt-metal TTNN
+  // optimization rules. Returns true if the operation should be skipped, false
+  // otherwise. reference:
+  // ttnn::operations::data_movement::ReshapeViewOperation::invoke Parameters:
+  // - reshapeOp: The reshape operation to analyze
+  // - contextOp: The operation context for logging (producer op for user
```
odjuricicTT: We shouldn't pass params just for logging. You can probably fetch the producer op from `reshapeOp->getOperand(0)->getDefiningOp()` or something similar.

lib/Dialect/TTNN/Analysis/L1InterleavedFallbackAnalysis.cpp
```diff
@@ -184,6 +198,112 @@ bool L1InterleavedFallbackAnalysis::isConv2DConvertibleToMatMul(Operation *op) {
...
+                   "L1InterleavedFallbackAnalysis: Skipping {} - reshape "
+                   "no-op.",
+                   contextOp->getName());
+    }
+    return true;
+  }
+
+  // Check if this reshape can be optimized to a view (TTNN-style check)
+  bool canBeView = false;
+  if (inputShape.size() >= 1 && outputShape.size() >= 1) {
```
odjuricicTT: If possible, always reduce the number of nested if statements.

This can be written as:
```cpp
if (inputShape.size() < 1 || outputShape.size() < 1) {
    return false;
}

// rest of the logic
```
Do we get shapes with size < 1, tho?

lib/Dialect/TTNN/Analysis/L1InterleavedFallbackAnalysis.cpp
```diff
@@ -184,6 +198,112 @@ bool L1InterleavedFallbackAnalysis::isConv2DConvertibleToMatMul(Operation *op) {
...
+
+  // Get input and output shapes
+  auto inputType = mlir::dyn_cast<mlir::RankedTensorType>(
+      reshapeOp->getOperand(0).getType());
+  auto outputType =
+      mlir::dyn_cast<mlir::RankedTensorType>(reshapeOp->getResult(0).getType());
+  auto inputShape = inputType.getShape();
+  auto outputShape = outputType.getShape();
+
+  // Skip if reshape is a no-op (same shape in and out).
```
odjuricicTT: Does this case exist? Do we remove such reshapes in canonicalizer pass?

lib/Dialect/TTNN/Analysis/L1InterleavedFallbackAnalysis.cpp
```diff
@@ -184,6 +198,112 @@ bool L1InterleavedFallbackAnalysis::isConv2DConvertibleToMatMul(Operation *op) {
...
+  if (inputShape == outputShape) {
+    if (isUserOp) {
+      TTMLIR_TRACE(ttmlir::LogComponent::Optimizer,
```
odjuricicTT: To reduce the amount of duplicated code, this can be wirtten as:
```cpp
TTMLIR_TRACE(ttmlir::LogComponent::Optimizer,
            "L1InterleavedFallbackAnalysis: Skipping {} -{} reshape "
            "no-op.",
            contextOp->getName(),
            isUserOp ? " user" : "");
```

lib/Dialect/TTNN/Analysis/L1InterleavedFallbackAnalysis.cpp
```diff
@@ -103,8 +103,22 @@ void L1InterleavedFallbackAnalysis::analysisImplementation() {
                    op->getName());
       return;
     }
+
+    if (isa<ttnn::ReshapeOp>(op)) {
+      if (checkReshapeSkip(op, op, false)) {
```
odjuricicTT: This can be one `if`

lib/Dialect/TTNN/Analysis/L1InterleavedFallbackAnalysis.cpp
```diff
@@ -184,6 +198,112 @@ bool L1InterleavedFallbackAnalysis::isConv2DConvertibleToMatMul(Operation *op) {
...
+    // 4. L1/DRAM must match for output and input
+    const int64_t tileHeight = 32; // tt::constants::TILE_HEIGHT
+    const int64_t tileWidth = 32;  // tt::constants::TILE_WIDTH
+
+    // Checking if the other tensor is sharded/interleaved:
+    // Input if we are checking the output of the op (this reshape is contextOp)
+    // Output if we are checking the input of the op (this reshape is user of
+    // contextOp)
+    bool otherTensorSharded = false;
+    if (isUserOp) {
```
odjuricicTT: It seems cleaner to remove the `isUserOp` argument from the function and have the `!otherTensorSharded` check before calling `checkReshapeSkip`.  It makes the function much simpler and easier to read and results in less code overall.

include/ttmlir/Dialect/TTNN/Analysis/L1InterleavedFallbackAnalysis.h
```diff
@@ -88,6 +88,17 @@ class L1InterleavedFallbackAnalysis
   // (1x1 kernel, stride=1, padding=0, dilation=1)
   bool isConv2DConvertibleToMatMul(Operation *op);
 
+  // Check if a reshape operation should be skipped based on tt-metal TTNN
+  // optimization rules. Returns true if the operation should be skipped, false
+  // otherwise. reference:
```
odjuricicTT: We want to avoid implementing logic that already exists in tt-metal if it is not absolutely necessary. The logic in tt-metal will change over time and that creates bugs with our version is out of sync and maintainance cost.

The proper solution for this problem is to check this based on getOpRuntime results once it can handle this case. Please file an issue to remove this once it can be replaced with getOpRuntime and leave a TODO here referencing the issue.

lib/Dialect/TTNN/Analysis/L1InterleavedFallbackAnalysis.cpp
bmalesevicTT: We do, my bad, this test, for example, even checks such case: [test/ttmlir/Dialect/TTNN/data_movement/reshape/reshape_folding_test_ttnn.mlir](https://github.com/tenstorrent/tt-mlir/blob/57f4afd4/test/ttmlir/Dialect/TTNN/data_movement/reshape/reshape_folding_test_ttnn.mlir)

lib/Dialect/TTNN/Analysis/L1InterleavedFallbackAnalysis.cpp
bmalesevicTT: I didn't come across such case, but to make sure still kept the if statement, rather than make it an assert or let it go under the hood.

include/ttmlir/Dialect/TTNN/Analysis/L1InterleavedFallbackAnalysis.h
bmalesevicTT: I will drive for a fix on this issue, in the meantime it is described here and added as a TODO:
[[L1IFA] Replace hardcoded reshape optimization logic with getOpRuntime once reshape hang is fixed #5086](https://github.com/tenstorrent/tt-mlir/issues/5086)

lib/Dialect/TTNN/Analysis/L1InterleavedFallbackAnalysis.cpp
bmalesevicTT: With removing "isUserOp" parameter, we lose information of whether this was 
called for checking the reshape operation itself vs. checking its producer.
To maintain debugging context, I made both trace messages mention both the 
reshape operation and its input producer, so logs consistently show the 
complete producer→reshape relationship regardless of which context triggered 
the check.

lib/Dialect/TTNN/Analysis/L1InterleavedFallbackAnalysis.cpp
bmalesevicTT: I did that, but slightly different: I realized it is simpler to check the other tensor being in DRAM, as that guarantees it is not sharded, and also, as another tt-metal optimization condition is that both input and output should have same buffer type, the skip of moving from DRAM->L1 should anyway only happen if the other is in DRAM. Everything should be clearer from the code comments.

lib/Dialect/TTNN/Analysis/L1InterleavedFallbackAnalysis.cpp
```diff
@@ -103,8 +103,28 @@ void L1InterleavedFallbackAnalysis::analysisImplementation() {
...
       return;
     }
+
+    if (isa<ttnn::ReshapeOp>(op)) {
+      // Output of reshape op is considered for keeping in DRAM, only if its
+      // input is still in DRAM.
+      auto inputType =
+          mlir::dyn_cast<mlir::RankedTensorType>(op->getOperand(0).getType());
+      auto ttnnLayout = mlir::dyn_cast<TTNNLayoutAttr>(inputType.getEncoding());
+      if (ttnnLayout.hasDRAMBufferType() && checkReshapeSkip(op)) {
```
rpavlovicTT: Use `utils::producesDRAMLayout` here too.

lib/Dialect/TTNN/Analysis/L1InterleavedFallbackAnalysis.cpp
```diff
@@ -184,6 +204,81 @@ bool L1InterleavedFallbackAnalysis::isConv2DConvertibleToMatMul(Operation *op) {
   return true;
 }
 
+bool L1InterleavedFallbackAnalysis::checkReshapeSkip(
+    Operation *reshapeOperation) const {
+
+  auto reshapeOp = cast<ttnn::ReshapeOp>(reshapeOperation);
```
rpavlovicTT: `dyn_cast` to be precise

lib/Dialect/TTNN/Analysis/L1InterleavedFallbackAnalysis.cpp
```diff
@@ -184,6 +204,81 @@ bool L1InterleavedFallbackAnalysis::isConv2DConvertibleToMatMul(Operation *op) {
...
+
```
rpavlovicTT: Also assert reshapeOp after cast to validate calls

lib/Dialect/TTNN/Analysis/L1InterleavedFallbackAnalysis.cpp
```diff
@@ -184,6 +204,81 @@ bool L1InterleavedFallbackAnalysis::isConv2DConvertibleToMatMul(Operation *op) {
...
+  int64_t outputLastDim = outputShape.back();
+
+  // Get second-to-last dimensions (or 1 if rank == 1)
+  int64_t inputSecondLastDim =
+      inputShape.size() >= 2 ? inputShape[inputShape.size() - 2] : 1;
+  int64_t outputSecondLastDim =
+      outputShape.size() >= 2 ? outputShape[outputShape.size() - 2] : 1;
+
+  const int64_t tileHeight = 32; // tt::constants::TILE_HEIGHT
+  const int64_t tileWidth = 32;  // tt::constants::TILE_WIDTH
```
rpavlovicTT: why don't you use constants from the comment?

lib/Dialect/TTNN/Analysis/L1InterleavedFallbackAnalysis.cpp
```diff
@@ -184,6 +204,81 @@ bool L1InterleavedFallbackAnalysis::isConv2DConvertibleToMatMul(Operation *op) {
...
+
+  bool inputTensorTiled = false;
+  if (auto ttnnLayout =
+          mlir::dyn_cast<TTNNLayoutAttr>(inputType.getEncoding())) {
```
rpavlovicTT: I think we are guaranteed to have `TTNNLayout` along with the input here. So this should be an assert.

lib/Dialect/TTNN/Analysis/L1InterleavedFallbackAnalysis.cpp
```diff
@@ -184,6 +204,81 @@ bool L1InterleavedFallbackAnalysis::isConv2DConvertibleToMatMul(Operation *op) {
...
+    inputTensorTiled = ttnnLayout.isTiled();
+  }
+
+  // Conditions reshape can be optimized to a view operation (based on tt-metal
+  // logic):
+  // 1. Last dimension must be the same
+  // 2. Either second-to-last dimension is the same OR no tile padding OR
+  // row-major input
+  // 3. Sharded/Interleaved must match for output and input
+  // 4. L1/DRAM must match for output and input
```
rpavlovicTT: We don't check these here, you can remove them from the comment.

lib/Dialect/TTNN/Analysis/L1InterleavedFallbackAnalysis.cpp
```diff
@@ -103,8 +103,28 @@ void L1InterleavedFallbackAnalysis::analysisImplementation() {
...
+      }
+    }
+    for (auto *user : op->getUsers()) {
+      // Input of reshape user is considered for keeping in DRAM, only if its
+      // output is still in DRAM.
+      if (isa<ttnn::ReshapeOp>(user) && utils::producesDRAMLayout(user) &&
+          checkReshapeSkip(user)) {
+        return;
+      }
+    }
```
rpavlovicTT: As this is a heuristic to align your analysis to tt-metal impl., let's put this block into a function for itself. Call it `handleReshapeOps` or similar. Let it return boolean whether you skip this op or not.

lib/Dialect/TTNN/Analysis/L1InterleavedFallbackAnalysis.cpp
```diff
@@ -184,6 +204,81 @@ bool L1InterleavedFallbackAnalysis::isConv2DConvertibleToMatMul(Operation *op) {
...
+  // 1. Last dimension must be the same
+  // 2. Either second-to-last dimension is the same OR no tile padding OR
+  // row-major input
+  // 3. Sharded/Interleaved must match for output and input
+  // 4. L1/DRAM must match for output and input
+  bool canBeView =
+      // 1. Sharded/Interleaved and L1/DRAM match for output and input is
+      // already checked, as the tensor being considered for upgrade is DRAM
+      // Interleaved, and the other one is checked to be DRAM as well before
+      // this function.
```
rpavlovicTT: When you add a new function to handle reshape ops, you can write down the full comment with all needed conditions at the top or header of the function.

odjuricicTT (approved):
Looks good from my side! Please address Radenko's comments and feel free to merge.

lib/Dialect/TTNN/Analysis/L1InterleavedFallbackAnalysis.cpp
bmalesevicTT: Since here I am checking the input, the only way to use the same function would be getting the producer operation here and then `utils::producesDRAMLayout`, which might make things more confusing in my opinion. Instead I added another utils function for this. Let me know if you think there is still space for improvement

lib/Dialect/TTNN/Analysis/L1InterleavedFallbackAnalysis.cpp
bmalesevicTT: The constants referenced in the comments are the ones from tt-metal, which I couldn't use, but you are right, now I used the same constants that exist in tt-mlir.

lib/Dialect/TTNN/Analysis/L1InterleavedFallbackAnalysis.cpp
```diff
@@ -184,6 +204,81 @@ bool L1InterleavedFallbackAnalysis::isConv2DConvertibleToMatMul(Operation *op) {
...
 
+bool L1InterleavedFallbackAnalysis::checkReshapeSkip(
+    Operation *reshapeOperation) const {
+
+  auto reshapeOp = cast<ttnn::ReshapeOp>(reshapeOperation);
+
+  auto inputType = mlir::dyn_cast<mlir::RankedTensorType>(
+      reshapeOp->getOperand(0).getType());
+  auto outputType =
+      mlir::dyn_cast<mlir::RankedTensorType>(reshapeOp->getResult(0).getType());
```
azecevic000: Once you `cast`, `reshapeOp` is a `ttnn::ReshapeOp` which means you have `getInput()` and `getResult()` accessors with a compile time known types.

lib/Dialect/TTNN/Analysis/L1InterleavedFallbackAnalysis.cpp
```diff
@@ -184,6 +204,81 @@ bool L1InterleavedFallbackAnalysis::isConv2DConvertibleToMatMul(Operation *op) {
...
+  auto inputShape = inputType.getShape();
+  auto outputShape = outputType.getShape();
+
+  assert(inputShape != outputShape &&
+         "Identity reshapes should have been removed by TTIR canonicalization");
```
azecevic000: Technically, it's a TTNN canonicalization because there is still a possibility that a TTNN IR is a source.

lib/Dialect/TTNN/Analysis/L1InterleavedFallbackAnalysis.cpp
```diff
@@ -184,6 +193,88 @@ bool L1InterleavedFallbackAnalysis::isConv2DConvertibleToMatMul(Operation *op) {
...
+  // Get second-to-last dimensions (or 1 if rank == 1)
+  int64_t inputSecondLastDim =
+      inputShape.size() >= 2 ? inputShape[inputShape.size() - 2] : 1;
+  int64_t outputSecondLastDim =
+      outputShape.size() >= 2 ? outputShape[outputShape.size() - 2] : 1;
+
+  bool inputTensorTiled = false;
+  auto ttnnLayout = mlir::dyn_cast<TTNNLayoutAttr>(inputType.getEncoding());
+  assert(ttnnLayout && "Expected input tensor to have TTNN layout attribute");
+  inputTensorTiled = ttnnLayout.isTiled();
```
azecevic000: Nit: you can declare `inputTensorTiled` directly in the place of its definition (`bool inputTensorTiled = ttnnLayout.isTiled();`), it's not used before that.

lib/Dialect/TTNN/Utils/Utils.cpp
```diff
@@ -319,6 +319,26 @@ bool producesTiledTensorLayout(Operation *op) {
...
+
+  auto firstOperand = op->getOperand(0);
+  auto tensorType =
+      mlir::dyn_cast<mlir::RankedTensorType>(firstOperand.getType());
+  if (!tensorType) {
+    return false;
+  }
+
+  if (auto ttnnLayout =
+          mlir::dyn_cast<TTNNLayoutAttr>(tensorType.getEncoding())) {
```
azecevic000: This can be written more tersely as
```suggestion
  if (auto ttnnLayout =
          mlir::dyn_cast_if_present<TTNNLayoutAttr>(tensorType.getEncoding())) {
```

lib/Dialect/TTNN/Analysis/L1InterleavedFallbackAnalysis.cpp
```diff
@@ -184,6 +193,88 @@ bool L1InterleavedFallbackAnalysis::isConv2DConvertibleToMatMul(Operation *op) {
...
+    if (isa<ttnn::ReshapeOp>(user) && utils::producesDRAMLayout(user) &&
+        checkReshapeSkip(user)) {
+      return true;
+    }
+  }
+  return false;
+}
+
+bool L1InterleavedFallbackAnalysis::checkReshapeSkip(
+    Operation *reshapeOperation) const {
```
azecevic000: Consider either renaming this variable to something like `op` because it's easy to mix it up with `reshapeOp`, or actually change the API to `checkReshapeSkip(ReshapeOp reshapeOp)`, which IMO is preferred because it gives you compile-time type verification, instead of runtime, which is currently the case. If `handleReshapeOps` is the only caller, the call site would change to
```cpp
if (auto reshapeOp = dyn_cast<ReshapeOp>(op); reshapeOp && utils::hasFirstOperandInDRAM(reshapeOp) &&
      checkReshapeSkip(reshapeOp)) {
  ...
}
```
which also means that you'll end up with one cast less. On the other hand, `utils::producesDRAMLayout` has a different intention, so `Operation *` should still be preferred.

lib/Dialect/TTNN/Analysis/L1InterleavedFallbackAnalysis.cpp
```diff
@@ -14,6 +14,8 @@
 #include "mlir/IR/BuiltinTypes.h"
 #include "mlir/Support/LLVM.h"
 
+#include "ttmlir/Dialect/TTNN/Types/Types.h"
```
azecevic000: Nit: move this into 
```cpp
#include "ttmlir/Dialect/TTCore/IR/Utils.h"
#include "ttmlir/Dialect/TTNN/IR/TTNNOps.h"
#include "ttmlir/Support/Logger.h"
```
section.

Issue #4992, opened 2025-09-17
Mesh shard Replicate asserts on 1D tensors


dgolubovicTT:
### Describe the bug

I ran into this issue while writing Mnist data parallel multichip test (on n300).

Here is the error in metal:

```

 RuntimeTTNN |    DEBUG | Executing operation: %42 = "ttnn.mesh_shard"(%41, %1) <{shard_dims = array<i64: -1>, shard_direction = #ttcore.shard_direction<shard_to_full>, shard_shape = array<i64: 1>, shard_type = #ttcore.shard_type<replicate>}> : (tensor<512xbf16, #ttnn.ttnn_layout<(d0) -> (0, d0), <1x1>, memref<1x512xbf16, #ttnn.buffer_type<system_memory>>>>, !ttnn.device) -> tensor<512xbf16, #ttnn.ttnn_layout<(d0) -> (0, d0), <1x1>, memref<1x512xbf16, #ttnn.buffer_type<system_memory>>>> loc(unknown)
2025-09-17 12:14:57.127 | critical |          Always | Invalid dimension index; got dims: SmallVector([1]), tensor dimension: 1 (assert.hpp:107) 

```

Seems like the problem is in creating mesh_shard config in `tt-mlir/runtime/lib/ttnn/operations/ccl/mesh_shard.cpp`:

```
    MeshComposerConfig meshComposerConfig;
    if (shardType == ::tt::target::MeshShardType::Replicate) {
      meshComposerConfig.dims.push_back(static_cast<int>(1));
```

This hard coded '1' triggers the assert in `partition.cpp` `normalize_dims`

PR #4993, opened 2025-09-17
Fix bug in concatenating 1d tensor

wooseokTT:
### Ticket
https://github.com/tenstorrent/tt-mlir/issues/4992

### Problem description
When concatenating replicated ttnn tensors, current code assume at least two dimensional tensor. 

### What's changed
Relax such 2d requirement and allow 1d tensor concatenation.

### DIFF
--- a/runtime/lib/ttnn/operations/ccl/mesh_shard.cpp
+++ b/runtime/lib/ttnn/operations/ccl/mesh_shard.cpp
@@ -69,7 +69,11 @@ void run(const ::tt::target::ttnn::MeshShardOp *op, ProgramContext &context) {
     // Nd (partial) Concat
     MeshComposerConfig meshComposerConfig;
     if (shardType == ::tt::target::MeshShardType::Replicate) {
-      meshComposerConfig.dims.push_back(static_cast<int>(1));
+      // All buffers in devices are replicated across devices. Thus, we pick up
+      // the data from the first device by providing {1} mesh shape. By setting
+      // 0 in the dim, we allow all dimensional tensors staring from single
+      // dimensional tensor.
+      meshComposerConfig.dims.push_back(static_cast<int>(0));
       meshComposerConfig.mesh_shape_override = ::ttnn::MeshShape({1});
     } else {
       // meshComposerConfig.dims must be unique, and thus, we need to find


runtime/lib/ttnn/operations/ccl/mesh_shard.cpp
```diff
@@ -69,7 +69,7 @@ void run(const ::tt::target::ttnn::MeshShardOp *op, ProgramContext &context) {
     // Nd (partial) Concat
     MeshComposerConfig meshComposerConfig;
     if (shardType == ::tt::target::MeshShardType::Replicate) {
-      meshComposerConfig.dims.push_back(static_cast<int>(1));
+      meshComposerConfig.dims.push_back(static_cast<int>(0));
```
dgolubovicTT, line 76: Can we add some comment in the code why this 0 now stands here?

wooseokTT: Got it. I added comments for detailed explanation.

jnie-TT: @wooseokTT I thought the dim here specified which dim the tensor replicas will get concatenated on. Should this be configured dynamically in the compiler instead of hardcoded to 0? Or do we always want to concat on dim 0?

wooseokTT: @jnie-TT  I believe that is true for regular concat cases. This "replicate" case, all buffers has same data, and thus, we use the first device buffer and make memcpy, resulting in single buffer. If I recall correctly, I am expecting this code is being used. 

https://github.com/tenstorrent/tt-metal/blob/04bbd54aa3d01fec1925638eb84bc90e8f53fd64/ttnn/core/tensor/xtensor/partition.cpp#L174

concat_ndim() is expected to be called with single entry of expressions = {first_device_data_view}, num_chunks = {1}, dims={0}

wooseokTT: Since num_chunks[0] = 1, the input and result shapes are identical, and thus, which dim we perform concatenation doesn't really matter, I believe. We use the first dim to make sure that we support the one dimensional tensor that caused the issue.

jnie-TT: Ah ok I see that makes sense, thanks for the explanation!

Issue #5008, opened 2025-09-18
EmitPy llama 3.2


sdjordjevicTT:

svuckovicTT:
We were testing prefill only and made that work, reopening as there's another issue for decode graph #5132

svuckovicTT:
The issue above was fixed, closing.

jojo-771771:
I can work on this. Please assign!

jojo-771771:
I can work on this. Please assign!

jojo-771771:
I can work on this. Please assign!

jojo-771771:
I can work on this. Please assign!

jojo-771771:
I can work on this. Please assign!

PR #5005, opened 2025-09-18
Add a ttnn-requirements.txt deps file for python TTNN lib

svuckovicTT:
### Ticket
Closes https://github.com/tenstorrent/tt-mlir/issues/4987

### Problem description
Python codegen'd models would fail due to couple missing python libs, needed by TTNN pylib.

### What's changed
Added a `ttnn-requirements.txt` to venv. All missing packages (4) were added here.

To add these packages to the current env, one can run:
```
cmake --build env/build --verbose -- python-venv
```
> [!NOTE]  
> This was failing for me due to ownership issues, as I was using the ird mlir docker that comes with prebuilt toolchain. To change owner from `root` to your own alias, you can run `sudo chown -R $USER /opt/ttmlir-toolchain`.

@tenstorrent/forge-developers-mlir-devops I've added this new file to docker hash file list. This is a small change to the python env and shouldn't affect anyone's workflow today as they wouldn't be using these packages yet. Is there any follow-up that needs to happen on my side?

### Checklist
- [ ] New/Existing tests provide coverage for changes

PR #5010, opened 2025-09-18
EmitPy codegen for Llama 3.2

sdjordjevicTT:
### Ticket
Closes https://github.com/tenstorrent/tt-mlir/issues/5008

### Problem description
EmitPy has rudimentary support at the moment, hence, llama 3.2 cannot be Python code generated.

### What's changed
Added a bunch of op conversions to support Python codegen for Llama 3.2.

### Checklist


llama.ttir.mlir
svuckovicTT, line 1: remove

tools/tt-alchemist/csrc/lib/generate_python.cpp
```diff
@@ -145,6 +145,24 @@ bool TTAlchemist::generatePython(const std::string &input_file,
...
+    return false;
+  }
+
+  // Add required Python packages
+  requirementsFile << "loguru\n";
+  requirementsFile << "pandas\n";
+  requirementsFile << "seaborn\n";
+  requirementsFile << "graphviz\n";
+
+  requirementsFile.close();
```
svuckovicTT: not needed once https://github.com/tenstorrent/tt-mlir/pull/5005 lands

tools/tt-alchemist/templates/python/local/run
```diff
@@ -12,6 +12,14 @@ if ! command -v python3 &> /dev/null; then
     exit 1
 fi
 
+# Check if VIRTUAL_ENV is set
+if [[ -z "${VIRTUAL_ENV}"
+  ]]; then
+    echo "Error: VIRTUAL_ENV environment variable is not set. Please activate your virtual environment."
+    exit 1
+fi
+pip install -r requirements.txt
```
svuckovicTT: Also should be removed once https://github.com/tenstorrent/tt-mlir/pull/5005 is in.

lib/Dialect/TTNN/IR/TTNNOpsAttrs.cpp
```diff
@@ -431,6 +431,29 @@ TTNNLayoutAttr TTNNLayoutAttr::withElementType(
...
+// param dataType The new data type.
+// return The new TTNNLayoutAttr with the given data type.
+TTNNLayoutAttr TTNNLayoutAttr::withDataType(ttcore::DataType dataType) {
+  Type elementType = getElementType();
+  if (isTiled()) {
+    elementType = mlir::tt::ttcore::TileType::get(
+        ttcore::dataTypeToElementType(getContext(), dataType));
+  } else {
+    elementType = ttcore::dataTypeToElementType(getContext(), dataType);
+  }
```
svuckovicTT:
```suggestion
  Type elementType = ttcore::dataTypeToElementType(getContext(), dataType);
  if (isTiled()) {
    elementType = mlir::tt::ttcore::TileType::get(elementType);
  }
```

include/ttmlir/Dialect/TTNN/Transforms/Passes.td
```diff
@@ -226,4 +226,15 @@ def TTNNOperationValidationAndFallback: Pass<"ttnn-operation-validation-and-fall
...
 }
 
+def TTNNEmitPyWorkarounds: Pass<"ttnn-emitpy-workarounds", "::mlir::ModuleOp"> {
+  let summary = "Apply TTNN workarounds specific to EmitPy pipeline.";
+  let description = [{
+    This pass applies TTNN workarounds that are specific to the EmitPy pipeline.
+    Currently handles:
+    - ConstantOp 4D shape workaround: Rewrites non-4D ConstantOps to 4D with
+      added reshape operations to fix the shape for ttnn.Tensor constructor
+      compatibility.
```
svuckovicTT: Removing this as it is hard to remember to keep up to date - it is already outdated :)

llama.ttir.mlir
svuckovicTT, line 1: Remove before merging

include/ttmlir/Conversion/TTNNToEmitPy/EmitPyConversion.h
```diff
@@ -860,6 +871,82 @@ struct EmitPyTypeConverter<std::array<T, k>> {
...
+  static std::string convert(mlir::FloatAttr attr) {
+    return convert(attr.getValueAsDouble());
+  }
+
+  static std::string convert(float value) { return std::to_string(value); }
+
+  static std::string convert(mlir::APFloat attr) {
+    return convert(static_cast<float>(attr.convertToDouble()));
+  }
+};
```
azecevic000: There is a more general floating-point type converter in `EmitCConversion.h`, the analogous one should be used here. It's also important to handle special FP values.

include/ttmlir/Dialect/TTNN/IR/TTNNBase.td
```diff
@@ -54,6 +54,10 @@ class TTNN_InplaceOp<string mnemonic, list<Trait> traits = []> :
         Op<TTNN_Dialect, mnemonic, [MemoryEffects<[MemWrite]>, TTNN_WorkaroundInterface,
         DeclareOpInterfaceMethods<TTNN_OpModelInterface, ["getOpConstraints", "getOpRuntime"]>] # traits>;
 
+class TTNN_MemFreeOp<string mnemonic, list<Trait> traits = []> :
+        Op<TTNN_Dialect, mnemonic, [MemoryEffects<[MemFree]>, TTNN_WorkaroundInterface,
+        DeclareOpInterfaceMethods<TTNN_OpModelInterface, ["getOpConstraints", "getOpRuntime"]>] # traits>;
+
```
azecevic000, line 60: IMO this class doesn't make sense, is there any other op other than `deallocate` that would be part of this class? If not, we can either create `TTNN_BaseOp` that would have `[TTNN_WorkaroundInterface, DeclareOpInterfaceMethods<TTNN_OpModelInterface, ["getOpConstraints", "getOpRuntime"]>` as traits and then add `[MemoryEffects<[MemFree]>` in deallocate, or just have `deallocate` as an 'instance' of `Op` class with appropriate traits (my preferred choice would be the former).

lib/Conversion/TTNNToEmitPy/TTNNToEmitPy.cpp
```diff
@@ -6,18 +6,100 @@
...
+                              getPrefixSwapPattern());
+  }
+};
+} // namespace
+
+// Default op conversion pattern, used to convert most ops.
+//
+namespace {
+template <typename SourceOp>
+class DefaultOpConversionPattern
```
azecevic000: It's used exactly 0 times at the moment, and I would assume it will stay the same in the future.

include/ttmlir/Conversion/TTNNToEmitPy/EmitPyConversion.h
svuckovicTT: Remove the float converter, doubled down on double converter

include/ttmlir/Dialect/TTNN/IR/TTNNBase.td
svuckovicTT: Our most basic class is `TTNN_Op`, which is already defined as `Pure`, so there's a need to define a new class that doesn't have the `Pure` trait on it. Ideally, the `TTNN_Op` would be clean of any traits, which seems to be a design choice for many other dialects. I'll open up an issue for this since it's a change for itself.

> or just have deallocate as an 'instance' of Op class with appropriate traits

I've tried this but seems it's not possible to get the `traits` list this way, looks like it needs to be defined when inheriting, but I might be missing something...

```
def TTNN_DeallocateOp : Op<TTNN_Dialect, "deallocate", [MemoryEffects<[MemFree]>, TTNN_WorkaroundInterface,
        DeclareOpInterfaceMethods<TTNN_OpModelInterface, ["getOpConstraints", "getOpRuntime"]>] # traits> { ... }
```
```
/localdev/svuckovic/_workspace/repos/tt-mlir/include/ttmlir/Dialect/TTNN/IR/TTNNOps.td:1949:99: error: Variable not defined: 'traits'
        DeclareOpInterfaceMethods<TTNN_OpModelInterface, ["getOpConstraints", "getOpRuntime"]>] # traits> {
```

svuckovicTT: https://github.com/tenstorrent/tt-mlir/issues/5087

include/ttmlir/Conversion/TTNNToEmitPy/EmitPyConversion.h
```diff
@@ -860,6 +872,72 @@ struct EmitPyTypeConverter<std::array<T, k>> {
   }
 };
 
+// Python's builtin float is actually double precision, so we always convert
+// floating point types to double. This disables conversion to single precision
+// float on purpose.
+//
+template <typename T>
+struct EmitPyTypeConverter<T,
+                           std::enable_if_t<std::is_same_v<T, float>, void>> {};
```
azecevic000: There is an inherent difference between semantics of types in C++ and Python, but that has little to do with serialization. We also have a converter for all integral types, despite the fact that none of them is a direct equivalent to Python's `int`.

This approach of converting various types is generally suboptimal for EmitPy, but if we are going with it, there isn't a single issue with having a converter for `float`, we are just making life harder without it. I see there is already an implementation for FP conversion on line 273. It should just be changed to
```cpp
template <typename T>
struct EmitPyTypeConverter<T, std::enable_if_t<std::is_floating_point_v<T>>> {
```
`convert(double)` in that struct can stay, if the argument is any FP type it will be implicitly converted to `double`.

include/ttmlir/Dialect/TTNN/IR/TTNNBase.td
azecevic000: .

svuckovicTT:
@azecevicTT ping

include/ttmlir/Dialect/TTNN/IR/TTNNBase.td
azecevic000: I think you do just
```
def TTNN_DeallocateOp : Op<TTNN_Dialect, "deallocate", [MemoryEffects<[MemFree]>, TTNN_WorkaroundInterface,
        DeclareOpInterfaceMethods<TTNN_OpModelInterface, ["getOpConstraints", "getOpRuntime"]>]>
```
`traits` is there for classes to let op instantiating them to inject per op traits, but here we are doing direct class instantiation.

include/ttmlir/Conversion/TTNNToEmitPy/EmitPyConversion.h
azecevic000: Technically, it's now not part of this PR, so I won't block on it, but the reason for not doing it is fundamentally wrong.

azecevicTT (approved):

jojo-771771:

jojo-771771:

jojo-771771:

jojo-771771:

jojo-771771:

PR #5022, opened 2025-09-18
Amend workaround for update cache op to extract a replicated updateIndex

jameszianxuTT:
Due to #1510, the update_cache op's updateIndex tensor must be returned to host and dissected to provide a uint32_t to TTNN.

This workaround does not work for a distributed tensor - for tensor parallel static cache update, the updateIndex may end up replicated across multiple devices, and ::ttnn::get_host_buffer explicitly asserts against this case.

This change uses the MeshComposer API to coalesce the replicated updateIndex into a single tensor and returns it to host. This continues to work if the updateIndex is not replicated as well.

### DIFF
--- a/runtime/lib/ttnn/operations/kv_cache/update_cache.cpp
+++ b/runtime/lib/ttnn/operations/kv_cache/update_cache.cpp
@@ -7,6 +7,8 @@
 #include "tt/runtime/detail/common/logger.h"
 
 #include "tt/runtime/workarounds.h"
+#include "ttnn/distributed/api.hpp"
+#include "ttnn/distributed/distributed_tensor.hpp"
 
 namespace tt::runtime::ttnn::operations::kv_cache {
 void run(const ::tt::target::ttnn::UpdateCacheOp *op, ProgramContext &context) {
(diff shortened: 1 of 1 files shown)


jameszianxuTT:
Required for https://github.com/tenstorrent/tt-xla/issues/1001 because tensor parallel static cache update involves a replicated updateIndices and we cannot pull a replicated tensor back to host via existing path.

Alternatively, some more complex logic in tt-xla to keep this update_index tensor on host, but I think this is a much simpler solution.

jameszianxuTT:
Assert referenced in commit description: https://github.com/tenstorrent/tt-metal/blob/main/ttnn/core/tensor/host_buffer/functions.cpp#L24

jameszianxuTT:
This also works on a single device / non replicated updateIndex tensor, I ran tt-torch CI against this commit and it passed: https://github.com/tenstorrent/tt-torch/actions/runs/17868355578/job/50819094879#step:11:114

LPanosTT:
I think to avoid needing to do this we should leave update cache indices tensors on host in the first place. That is to never push to device. Could you create an issue for this and add a TODO in the workaround here?

runtime/lib/ttnn/operations/kv_cache/update_cache.cpp
```diff
@@ -20,9 +22,26 @@ void run(const ::tt::target::ttnn::UpdateCacheOp *op, ProgramContext &context) {
...
+
   if (workaround::Env::get().readUpdateIndexFromDeviceForKVCache) {
 
-    const ::ttnn::Tensor indexOnHost = ::ttnn::from_device(updateIndex);
+    // Handle distributed tensors by aggregating them to a single tensor
+    // TODO(jameszianxu): The updateIndex should not be kept on host and not
+    // transferred to device. See
+    // https://github.com/tenstorrent/tt-xla/issues/1437
+
+    auto *mesh_device = updateIndex.device();
```
jnie-TT: Nit: camelCase `meshDevice`

runtime/lib/ttnn/operations/kv_cache/update_cache.cpp
```diff
@@ -20,9 +22,26 @@ void run(const ::tt::target::ttnn::UpdateCacheOp *op, ProgramContext &context) {
...
+
+    auto composer = ::ttnn::distributed::create_mesh_composer(
+        *mesh_device, ::ttnn::distributed::MeshComposerConfig{
+                          .dims = {0},
+                          .mesh_shape_override =
+                              ::ttnn::MeshShape(1), // Collapse to single device
+                      });
+
+    indexOnHost = ::ttnn::distributed::aggregate_tensor(updateIndex, *composer);
```
jnie-TT: Nit: Do we need to initialize an empty tensor outside the workaround block? Could we just do `::ttnn::Tensor indexOnHost = ::ttnn::distributed::aggregate_tensor(updateIndex, *composer);`

jnie-TT (approved):
Couple nits otherwise looks good

runtime/lib/ttnn/operations/kv_cache/update_cache.cpp
jameszianxuTT: Good catch, thanks.

jojo-771771:

jojo-771771:

jojo-771771:

Issue #4801, opened 2025-09-04
implement API for querying stream buffer sizes


vroubtsovTT:
`ttir.allocator` pass v3 currently mocks the choices of which buffer sizes to use so that (a) the choice is binary and (b) it is a single tile. We need to take the next step and implement [the buffer size query API as discussed previously](https://docs.google.com/document/d/1k-QDCAYwezy6azOdgPhzHVZXnQYwsen5_kMArBIgeqo/).

nsmithtt:
Closing

PR #5028, opened 2025-09-19
[D2M] Initial implementation of GenericOpBufferAnalysis

bgrady-tt:
### Ticket
Link to Github Issue

### What's changed

### Checklist

### DIFF
--- a/include/ttmlir/Dialect/D2M/Analysis/GenericOpBufferAnalysis.h
+++ b/include/ttmlir/Dialect/D2M/Analysis/GenericOpBufferAnalysis.h
@@ -32,7 +32,8 @@ public:
 
   /// Buffer settings for a single operand.
   struct BufferSetting {
-    SmallVector<int64_t> bufferShape; // in terms of elements
+    /// Shape is in terms of elements.
+    SmallVector<int64_t> bufferShape;
     std::uint32_t numBuffers;
   };
 
--- a/lib/Dialect/D2M/Analysis/GenericOpBufferAnalysis.cpp
+++ b/lib/Dialect/D2M/Analysis/GenericOpBufferAnalysis.cpp
@@ -18,14 +18,13 @@ GenericOpBufferAnalysis::analyzeGenericOp(const Constraints &constraints,
   SmallVector<SmallVector<int64_t>> operandShardShapes =
       op.getOperandShardShapes(/*convertTileToScalar=*/false);
   for (auto [operandIndex, operand] : llvm::enumerate(op.getOperands())) {
-    size_t numBuffers = (constraints.bufferingStrategy ==
-                         Constraints::BufferStrategy::DoubleBuffered)
-                            ? 2
-                            : 1;
 
     BufferSetting bufferSetting;
     bufferSetting.bufferShape = operandShardShapes[operandIndex];
-    bufferSetting.numBuffers = numBuffers;
+    bufferSetting.numBuffers = (constraints.bufferingStrategy ==
+                                Constraints::BufferStrategy::DoubleBuffered)
+                                   ? static_cast<std::uint32_t>(2)
+                                   : static_cast<std::uint32_t>(1);
     result.operandBufferSettings.push_back(std::move(bufferSetting));
   }
 
--- a/test/unittests/Allocation/TestGenericOpBufferAnalysis.cpp
+++ b/test/unittests/Allocation/TestGenericOpBufferAnalysis.cpp
@@ -145,7 +145,7 @@ TEST_F(GenericOpBufferAnalysisTest, CanCreateGenericOpsWithDifferentGrids) {
   EXPECT_EQ(grid3.getShape()[1], 4);
 }
 
-// Value-parameterized tests for different sharding configurations
+// Value-parameterized tests for different sharding configurations.
 struct GenericOpBufferAnalysisTestParams {
   llvm::SmallVector<int64_t> grid;
   llvm::SmallVector<int64_t> dims;


bgrady-tt:
@vroubtsovTT this is something resembling the BlockFactorAnalysis I discussed in the face-to-face. For now, it basically just allows querying buffer memref shapes for single/double buffer cases.

vroubtsovTT:
Is this going to be an implementation of 

> @vroubtsovTT this is something resembling the BlockFactorAnalysis I discussed in the face-to-face. For now, it basically just allows querying buffer memref shapes for single/double buffer cases.

Is this going to be an implementation of #4801 ?

lib/Dialect/TTIR/Transforms/Allocate.cpp
```diff
@@ -7,6 +7,7 @@
 #include "ttmlir/Asserts.h"
 #include "ttmlir/Dialect/TTCore/IR/TTCore.h"
 #include "ttmlir/Dialect/TTIR/Analysis/AllocationPlanner.h"
+#include "ttmlir/Dialect/TTIR/Analysis/BlockFactorAnalysis.h"
```
vroubtsovTT: there doesn't seem to be any value in adding an unused #include at this stage. `Allocate.cpp` has changed a lot in #4744 that is in flight.

test/unittests/Allocation/TestGenericOpBufferAnalysis.cpp
```diff
@@ -0,0 +1,258 @@
...
+#include "ttmlir/Dialect/TTIR/IR/TTIROps.h"
+
+#include "mlir/Dialect/Func/IR/FuncOps.h"
+#include "mlir/IR/AffineMap.h"
+#include "mlir/IR/Builders.h"
+#include "mlir/IR/BuiltinTypes.h"
+#include "mlir/IR/MLIRContext.h"
+#include "mlir/IR/OwningOpRef.h"
+
+#include "gtest/gtest.h"
```
vroubtsovTT, line 19: consider replacing this with
```cpp
#include "testing/Utils.h"
```
which includes gtest.h along with some unit testing facilities we've added, e.g. for writing randomized tests.

include/ttmlir/Dialect/TTIR/Analysis/BlockFactorAnalysis.h
```diff
@@ -0,0 +1,68 @@
...
+#include "mlir/IR/Operation.h"
+#include "mlir/IR/Value.h"
+#include "mlir/Support/LLVM.h"
+
+#include "llvm/ADT/SmallVector.h"
+
+namespace mlir::tt::ttir {
+
+struct BlockFactorAnalysisConstraints {
+  enum class BufferStrategy { SINGLE_BUFFERED, DOUBLE_BUFFERED };
```
vroubtsovTT: I suggest specifying an explicit underlying base type of `std::int32_t` instead of the `int` default which has no standard width should we ever need to serialized this enum.

vroubtsovTT: Also, although I see conflicting examples within llvm codebase, I think our convention for enumerator naming style is what's in the official guide https://llvm.org/docs/CodingStandards.html#name-types-functions-variables-and-enumerators-properly, i.e. the same style as for types.

lib/Dialect/TTIR/Transforms/Allocate.cpp
bgrady-tt: Will remove 👍

bgrady-tt:
@vroubtsovTT 
> Is this going to be an implementation of #4801 ?

Yes, as-is this amounts to just a query class for stream buffer sizes. 

In the medium term, I see this evolving into a kind of unified performance analysis for different blocking schemes. So the indirect naming reflects that it is intended to not be _just_ for buffer sizing.

test/unittests/Allocation/TestBlockFactorAnalysis.cpp
```diff
@@ -0,0 +1,263 @@
...
+  for (size_t i = 0; i < dims.size(); i++) {
+    shardShape[i] = dims[i] / grid[i];
+  }
+  // full memref shape is grid concat shard shape
+  auto memrefShape = llvm::SmallVector<int64_t>(grid);
+  memrefShape.insert(memrefShape.end(), shardShape.begin(), shardShape.end());
+
+  auto tileSize = tileType.getSizeBytes();
+  auto shardStrides = llvm::SmallVector<int64_t>(shardShape.size(), 0);
+  int64_t stride_accum = tileSize;
```
vroubtsovTT: naming style

test/unittests/Allocation/TestBlockFactorAnalysis.cpp
```diff
@@ -0,0 +1,263 @@
...
+  auto identityMap2D = mlir::AffineMap::getMultiDimIdentityMap(2, &context);
+  auto indexingMaps = builder.getAffineMapArrayAttr(
+      {identityMap2D, identityMap2D, identityMap2D});
+
+  // Create iterator types (parallel for all dimensions)
+  auto parallelType = mlir::tt::ttcore::IteratorTypeAttr::get(
+      &context, mlir::tt::ttcore::IteratorType::Parallel);
+  auto iteratorTypes = builder.getArrayAttr({parallelType, parallelType});
+
+  // use memref alloc to create concrete memref values from types
```
vroubtsovTT: nit: our style suggests comments should always form full sentences ending with '.'

include/ttmlir/Dialect/D2M/Analysis/BlockFactorAnalysis.h
```diff
@@ -0,0 +1,69 @@
...
+/// Multiple buffering options may be returned with different runtime tradeoffs;
+/// each BufferConfig represents an implementable set of buffering decisions for
+/// all operands (hopefully) representing a point on the implementation pareto
+/// frontier.
+class BlockFactorAnalysis {
+public:
+  /// Buffer settings for a single operand.
+  struct BufferSetting {
+    SmallVector<int64_t> bufferShape; // in terms of elements
+    size_t numBuffers;
```
vroubtsovTT: I think in the tablegen we have std::uint32_t for `buffers` values, perhaps this should be done here as well.
```
  let parameters = (ins ArrayRefParameter<"int64_t">:$stride,
                        DefaultValuedParameter<"uint32_t", "1">:$buffers);
```

test/unittests/Allocation/TestBlockFactorAnalysis.cpp
```diff
@@ -0,0 +1,263 @@
...
+TEST_F(BlockFactorAnalysisTest,
+       CanPerformBlockFactorAnalysis1x1ShardedGenericOp) {
+
+  // Test Case 1: Small single-core operation (1x1 grid, 2x4 dims)
+  auto genericOp1 = createGenericOp(builder, context, {1, 1}, {2, 4});
+
+  BlockFactorAnalysis analysis1;
+  auto bufferConfigs1 = analysis1.analyzeGenericOp(genericOp1);
+
+  EXPECT_FALSE(bufferConfigs1.empty());
```
vroubtsovTT: If you use `ASSERT_FALSE()` here instead of EXPECT, you won't need to condition the subsequent test logic on the same condition.

test/unittests/Allocation/TestBlockFactorAnalysis.cpp
```diff
@@ -0,0 +1,263 @@
...
+      EXPECT_FALSE(setting.bufferShape.empty());
+      EXPECT_GT(setting.numBuffers, 0u);
+    }
+    EXPECT_EQ(config2.operandBufferSettings[0].bufferShape,
+              (llvm::SmallVector<int64_t>{2, 4}));
+  }
+}
+
+TEST_F(BlockFactorAnalysisTest,
+       CanPerformBlockFactorAnalysis1x4ShardedGenericOp) {
```
vroubtsovTT: Am I correct in reading CanPerformBlockFactorAnalysis1x1ShardedGenericOp, CanPerformBlockFactorAnalysis2x2ShardedGenericOp, and CanPerformBlockFactorAnalysis1x4ShardedGenericOp as being the same unit test but with different parameters?

gtest supports value- and type-parameterized tests for exactly this situation. Over time, I've come to think that type-parameterized are the most universal type. They are a bit painful to set up but once I understood the macro recipe it became cookie cutter. Have a look at TYPED_TEST in TestAllocation.cpp; the doc is https://google.github.io/googletest/reference/testing.html#TYPED_TEST_SUITE

include/ttmlir/Dialect/D2M/Analysis/BlockFactorAnalysis.h
```diff
@@ -0,0 +1,69 @@
...
+
+  BlockFactorAnalysis() = default;
+  explicit BlockFactorAnalysis(BlockFactorAnalysisConstraints constraints)
+      : constraints(constraints) {}
+  ~BlockFactorAnalysis() = default;
+
+  BlockFactorAnalysis(const BlockFactorAnalysis &) = default;
+  BlockFactorAnalysis &operator=(const BlockFactorAnalysis &) = default;
+  BlockFactorAnalysis(BlockFactorAnalysis &&) = default;
+  BlockFactorAnalysis &operator=(BlockFactorAnalysis &&) = default;
```
vroubtsovTT: Do we expect the caller to want/be able to copy BlockFactorAnalysis around? When in doubt I default such classes to be non-copyable/non-moveable because that's safer and undoing it later is always a non-breaking code change.

include/ttmlir/Dialect/D2M/Analysis/BlockFactorAnalysis.h
```diff
@@ -0,0 +1,69 @@
...
+  /// Buffer configuration options for a single GenericOp.
+  struct BufferConfig {
+    /// Buffer settings indexed by operand index of the associated GenericOp.
+    SmallVector<BufferSetting> operandBufferSettings;
+    float predictedRuntimeCost = 0.0f;
+  };
+
+  BlockFactorAnalysis() = default;
+  explicit BlockFactorAnalysis(BlockFactorAnalysisConstraints constraints)
+      : constraints(constraints) {}
```
vroubtsovTT: Is default-constructing equivalent to a particular choice of `BlockFactorAnalysisConstraints`? The default constructor should be doc'ed to make that clear.

include/ttmlir/Dialect/D2M/Analysis/BlockFactorAnalysis.h
```diff
@@ -0,0 +1,69 @@
...
+  ~BlockFactorAnalysis() = default;
+
+  BlockFactorAnalysis(const BlockFactorAnalysis &) = default;
+  BlockFactorAnalysis &operator=(const BlockFactorAnalysis &) = default;
+  BlockFactorAnalysis(BlockFactorAnalysis &&) = default;
+  BlockFactorAnalysis &operator=(BlockFactorAnalysis &&) = default;
+
+  // Analyze a ttir::GenericOp and return buffering
+  SmallVector<BufferConfig> analyzeGenericOp(GenericOp op);
```
vroubtsovTT: BlockFactorAnalysis looks like it's stateless (except for `constraints` which are `const` conceptually). Therefore, analyzeGenericOp() can be `const`? Otherwise it's unclear what c++ state it might be changing.

lib/Dialect/D2M/Analysis/BlockFactorAnalysis.cpp
```diff
@@ -0,0 +1,36 @@
...
+  SmallVector<BufferConfig> results;
+
+  BufferConfig result;
+  result.operandBufferSettings.reserve(op.getOperands().size());
+  result.predictedRuntimeCost = 0.0f;
+
+  SmallVector<SmallVector<int64_t>> operandShardShapes =
+      op.getOperandShardShapes(/*convertTileToScalar=*/false);
+  for (auto [operandIndex, operand] : llvm::enumerate(op.getOperands())) {
+    size_t num_buffers =
```
vroubtsovTT: naming style

lib/Dialect/D2M/Analysis/BlockFactorAnalysis.cpp
```diff
@@ -0,0 +1,36 @@
...
+        (constraints.bufferingStrategy ==
+         BlockFactorAnalysisConstraints::BufferStrategy::DoubleBuffered)
+            ? 2
+            : 1;
+
+    BufferSetting bufferSetting;
+    bufferSetting.bufferShape = operandShardShapes[operandIndex];
+    bufferSetting.numBuffers = num_buffers;
+    result.operandBufferSettings.push_back(bufferSetting);
```
vroubtsovTT: As a code pattern, I'd normally `emplace_back(std::move(bufferSetting))` if it's the last use of `bufferSetting` in this loop.

test/unittests/Allocation/TestBlockFactorAnalysis.cpp
```diff
@@ -0,0 +1,263 @@
...
+TEST_F(BlockFactorAnalysisTest, CanAnalyzeBufferingStrategies) {
+
+  auto genericOp = createGenericOp(builder, context, {1, 1}, {2, 4});
+
+  BlockFactorAnalysisConstraints singleBufferConstraints{
+      Constraints::BufferStrategy::SingleBuffered};
+  mlir::tt::d2m::BlockFactorAnalysis analysisSingle(singleBufferConstraints);
+  auto bufferConfigsSingle = analysisSingle.analyzeGenericOp(genericOp);
+
+  EXPECT_FALSE(bufferConfigsSingle.empty());
```
vroubtsovTT: like above, this should probably be ASSERT_FALSE, or the next line segfaults

vroubtsovTT (commented):
some more comments

test/unittests/Allocation/TestBlockFactorAnalysis.cpp
bgrady-tt: fixed 👍

include/ttmlir/Dialect/D2M/Analysis/BlockFactorAnalysis.h
bgrady-tt: If/when there is eventually caching, all calls to `analyzeGenericOp` would become technically non-const. So this is preventing painful refactor later. 

I can communicate that here so it is clear that the non-const doesn't imply the results returned are themselves stateful or dependent on prior setup.

vroubtsovTT: Ok with your suggestion but would point out that such "caching" in c++ APIs is often done using state fields marked with `mutable`, since the API remains "semantically" const from the caller's point of view.

lib/Dialect/D2M/Analysis/BlockFactorAnalysis.cpp
vwellsTT: I agree with using std::move, but not emplace_back, FWIW.  Milan sent me a stackoverflow link a while ago which changed my views, let me see if I can find.  But the TL;DR is that `push_back` is safer, and therefore better to use for cases where you don't want to use the `emplace_back` construct-in-place feature

vwellsTT: https://stackoverflow.com/questions/10890653/why-would-i-ever-use-push-back-instead-of-emplace-back I think it was this

vwellsTT: (ofc in this case doesn't really matter, but I still believe in defaulting to push_back nowadays)

vroubtsovTT: Yes, the salient part of my suggestion is doing `std::move()` on the last use of an object.

The push_back vs emplace_back example Vincent has linked is interesting. I think our codebase has already descended the unfortunate path of using more lax constructors (but choosing not to use {}-initialization).... I will think about whether to change my default for push_back/emplace_back.

vroubtsovTT:
Looking forward to the first non-initial version of this, e.g. one that generates >1 config option so it is actually a decision variable for the allocator.

include/ttmlir/Dialect/D2M/Analysis/BlockFactorAnalysis.h
bgrady-tt: fixed

include/ttmlir/Dialect/D2M/Analysis/BlockFactorAnalysis.h
bgrady-tt: Removed these copy and move constructors. Normally I regret this, but thinking about the use case it likely shouldn't be a value in a container anyway.

include/ttmlir/Dialect/D2M/Analysis/BlockFactorAnalysis.h
bgrady-tt: That is a good suggestion; I will make it `const` and insert `mutable` to the cache member var if needed later.

lib/Dialect/D2M/Analysis/BlockFactorAnalysis.cpp
bgrady-tt: fixed

lib/Dialect/D2M/Analysis/BlockFactorAnalysis.cpp
bgrady-tt: I will add a `std::move` there, although since its trivially copyable data on the stack I don't think it makes much difference.

I have seen these arguments before against defaulting to `emplace_back`, and so avoid it unless the code path is expected to be hot or constructing in place is desirable for other reasons.

test/unittests/Allocation/TestBlockFactorAnalysis.cpp
bgrady-tt: fixed.

bgrady-tt:
@vroubtsovTT I finished addressing all of your comments; I believe in all cases I ended up applying changes (except emplace_back). I also renamed the class to `GenericOpBufferAnalysis`, as I think this is more descriptive of what it does right now.

include/ttmlir/Dialect/D2M/Analysis/GenericOpBufferAnalysis.h
```diff
@@ -0,0 +1,57 @@
...
+class GenericOpBufferAnalysis {
+public:
+  struct Constraints {
+    enum class BufferStrategy : uint32_t { SingleBuffered, DoubleBuffered };
+    BufferStrategy bufferingStrategy = BufferStrategy::DoubleBuffered;
+  };
+
+  /// Buffer settings for a single operand.
+  struct BufferSetting {
+    SmallVector<int64_t> bufferShape; // in terms of elements
```
vwellsTT: nit: I'd perhaps consider moving the comment up a line and making it a `///` for consistency (and making it proper sentence)

lib/Dialect/D2M/Analysis/GenericOpBufferAnalysis.cpp
```diff
@@ -0,0 +1,36 @@
...
+  SmallVector<OpConfig> results;
+
+  OpConfig result;
+  result.operandBufferSettings.reserve(op.getOperands().size());
+  result.predictedRuntimeCost = 0.0f;
+
+  SmallVector<SmallVector<int64_t>> operandShardShapes =
+      op.getOperandShardShapes(/*convertTileToScalar=*/false);
+  for (auto [operandIndex, operand] : llvm::enumerate(op.getOperands())) {
+    size_t numBuffers = (constraints.bufferingStrategy ==
```
vwellsTT: nit: personally, would just assign `bufferSetting.numBuffers` to this RHS directly, since it's not used elsewhere.  Else should be `const` imo

test/unittests/Allocation/TestGenericOpBufferAnalysis.cpp
```diff
@@ -0,0 +1,218 @@
...
+  EXPECT_EQ(grid2.getShape()[1], 2);
+
+  auto genericOp3 = createGenericOp(builder, context, {1, 4}, {2, 8});
+  EXPECT_TRUE(genericOp3);
+  auto grid3 = genericOp3.getGrid();
+  EXPECT_EQ(grid3.getShape()[0], 1);
+  EXPECT_EQ(grid3.getShape()[1], 4);
+}
+
+// Value-parameterized tests for different sharding configurations
```
vwellsTT: I guess this should end with `.` though I'm not sure if unit tests have same code-standards as normal code :)

include/ttmlir/Dialect/D2M/Analysis/GenericOpBufferAnalysis.h
vroubtsovTT: For comments like this one that is meant to be an "after member" comment I use `///< ...` syntax, one of the allowed variants, per https://www.doxygen.nl/manual/docblocks.html#memberdoc

bgrady-tt: fixed.

lib/Dialect/D2M/Analysis/GenericOpBufferAnalysis.cpp
```diff
@@ -0,0 +1,36 @@
...
+                            ? 2
+                            : 1;
+
+    BufferSetting bufferSetting;
+    bufferSetting.bufferShape = operandShardShapes[operandIndex];
+    bufferSetting.numBuffers = numBuffers;
+    result.operandBufferSettings.push_back(std::move(bufferSetting));
+  }
+
+  results.push_back(result);
```
vroubtsovTT, line 31: > I believe in all cases I ended up applying changes (except emplace_back).

nit: this is one more case here where you could be pushing std::move()d `result`, analogous reason to `bufferSetting`.

test/unittests/Allocation/TestGenericOpBufferAnalysis.cpp
```diff
@@ -0,0 +1,218 @@
...
+
+INSTANTIATE_TEST_SUITE_P(
+    GenericOpBufferAnalysisTests, GenericOpBufferAnalysisParamTest,
+    ::testing::Values(
+        GenericOpBufferAnalysisTestParams{{1, 1}, {2, 4}, {2, 4}, "1x1Sharded"},
+        GenericOpBufferAnalysisTestParams{{2, 2}, {4, 8}, {2, 4}, "2x2Sharded"},
+        GenericOpBufferAnalysisTestParams{
+            {1, 4}, {2, 8}, {2, 2}, "1x4Sharded"}));
+
+TEST_F(GenericOpBufferAnalysisTest, CanAnalyzeBufferingStrategies) {
```
vroubtsovTT, line 189: nice!

lib/Dialect/D2M/Analysis/GenericOpBufferAnalysis.cpp
vwellsTT: heck, the most optimal thing to do would be to use `emplace_back` with proper fields to construct the `OpConfig` truly in place here 🙂

lib/Dialect/D2M/Analysis/GenericOpBufferAnalysis.cpp
bgrady-tt: fixed!

jojo-771771:

jojo-771771:

jojo-771771:

jojo-771771:

Issue #5003, opened 2025-09-18
Add an option to load tensors in TTNN->EmitC and TTNN->EmitPy


azecevicTT:
There is only a pass that creates inputs as `ttnn.ones` tenors. We need to implement a pass that loads tensors from disk for inputs and add an option in the pipeline as a choice which of these approaches a user wants to take.

jojo-771771:
I can work on this. Please assign!

jojo-771771:
I can work on this. Please assign!

jojo-771771:
I can work on this. Please assign!

jojo-771771:
I can work on this. Please assign!

jojo-771771:
I can work on this. Please assign!

jojo-771771:
I can work on this. Please assign!

PR #5037, opened 2025-09-22
[EmitC/EmitPy] Load input tensors

azecevicTT:
### Ticket
https://github.com/tenstorrent/tt-mlir/issues/5003

### Problem description
With the addition of `ttnn.load_tensor`, it's now possible to load tensors from disks. We can use this in EmitC (and EmitPy) pipelines to load input tensors instead of creating fixed `ttnn.ones`.

### What's changed
Added a pipeline option `load-input-tensors-from-disk` that will generate a function that will load tensors with `ttnn.load_tensor`. Currently, it's a boolean option; if specified, it's expected that all inputs will be loaded and their names are hardcoded to `arg0.tensorbin`, `arg1.tensorbin`,..., `argN.tensorbin`. These restrictions could be weakened in the future if there is a necessity (for example, loading some inputs from defined paths, while generating synthetic tensors for others).
 
### Checklist


test/ttmlir/EmitC/TTNN/load_input/arg0.tensorbin
svuckovicTT, line 1: Instead of checking in binaries, I think it'd be better to have a short script that calls `ttnn.save_tensor` to generate these.

lib/Dialect/TTNN/Pipelines/TTNNPipelines.cpp
```diff
@@ -215,10 +215,15 @@ void createTTNNBackendToEmitCPipeline(
...
+    // In canonical path, run tuplification + input generation/loading.
     //
     pm.addPass(createTTNNTuplifyTensors());
-    pm.addPass(createTTNNCreateInputGenerators());
+
+    if (options.loadInputTensorsFromDisk) {
+      pm.addPass(createTTNNLoadInputTensors());
+    } else {
+      pm.addPass(createTTNNCreateInputGenerators());
+    }
```
svuckovicTT, line 229: C/p for emitpy pipeline below.

lib/Dialect/TTNN/Transforms/Passes.cpp
```diff
@@ -154,17 +155,11 @@ class TTNNDeallocate : public impl::TTNNDeallocateBase<TTNNDeallocate> {
...
-public:
-  using impl::TTNNCreateInputGeneratorsBase<
-      TTNNCreateInputGenerators>::TTNNCreateInputGeneratorsBase;
-
-  void runOnOperation() final {
-    ModuleOp moduleOp = getOperation();
-    IRRewriter rewriter(&getContext());
-
+template <typename Derived>
+class TTNNInputFunctionCreatorBase {
```
svuckovicTT, line 160: Nice solution!

lib/Dialect/TTNN/Transforms/Passes.cpp
```diff
@@ -340,24 +361,66 @@ class TTNNCreateInputGenerators
...
-      ttnn::GetDeviceOp device =
-          ttnn::utils::getOrInsertDevice(rewriter, onesOp);
+    ttnn::TTNNLayoutAttr layoutAttr =
+        mlir::cast<ttnn::TTNNLayoutAttr>(tensorType.getEncoding());
 
-      mlir::Value tensorOnDevice = rewriter.create<ttnn::ToDeviceOp>(
-          loc, tensorType, onesOp, device, nullptr);
+    // Create hardcoded filename: arg0.tensorbin, arg1.tensorbin, etc.
+    //
+    std::string filename = "arg" + std::to_string(argIndex) + ".tensorbin";
```
svuckovicTT, line 411: @sgligorijevicTT can we name tensors like this when dumping them to disk?

Another way to do it would be to read an attribute off the tensor (e.g. location), which you've already tested? If so, we could to that as a follow-up.

.github/workflows/build-and-test.yml
svuckovicTT, line 1: There's 3 reasons I believe we should test this in tt-alchemist instead of ttnn-standalone:
- we want to deprecate ttnn-standalone asap
- models team will be dumping models from tt-xla, which uses tt-alchemist
- emitpy support

.github/workflows/build-and-test.yml
```diff
@@ -352,6 +352,26 @@ jobs:
...
+        # Process the load_input test
+        cd ${{ steps.strings.outputs.build-output-dir }}/test/ttmlir/EmitC/TTNN/load_input
+        ttmlir-opt -ttir-to-emitc-pipeline="load-input-tensors-from-disk=true" ttnn_load_input_tensors.mlir | ttmlir-translate -mlir-to-cpp -o ttnn-standalone.cpp
+
+        # Copy generated code and tensor files to ttnn-standalone directory
+        cp arg0.tensorbin arg1.tensorbin ttnn-standalone.cpp ${{ steps.strings.outputs.install-output-dir }}/tools/ttnn-standalone/
+
+        # Build and run
+        cd ${{ steps.strings.outputs.install-output-dir }}/tools/ttnn-standalone
+        ./run
```
nsumrakTT: I'm not sure why having another step when you can add these lines to previous step. It is executed on the same machine and same type of test anyway...

lib/Dialect/TTNN/Transforms/Passes.cpp
sgligorijevicTT: Yeah, naming inputs according to their name in the function signature is the easiest for me to support. In an internal implementation of dumping I have, I named tensors `arg_{num}_loc_{loc}.tensorbin`. It would be easy to remove the underscores and loc part to conform to this interface.

azecevic000: Discussed this with @sgligorijevicTT offline, since 'forward' function is a public function, there shouldn't be a way to change its signature because it would break of contract between a frontend and tt-mlir. So any convention that involves indicies should be alright.

.github/workflows/build-and-test.yml
azecevic000: Changed the test from `ttnn-standalone` to `tt-alchemist`, but still as a new step (which follows the current approach for `tt-alchemist` tests). Just curious, is there (a significant) overhead of executing multiple steps vs executing the same tasks in one step?

nsumrakTT: No, just the clutter of workflow file.
I'm working on refurbishment of tt-mlir CI, so I have to track changes and incorporate them into new CI.

.github/workflows/build-and-test.yml
```diff
@@ -548,6 +538,19 @@ jobs:
...
+      run: |
+        source testenv/bin/activate
+        OUTPUT_DIR=/tmp/test-load-tensors
+        rm -rf $OUTPUT_DIR
+        tt-alchemist generate-cpp --pipeline-options 'load-input-tensors-from-disk=true' test/ttmlir/EmitC/TTNN/load_input/ttnn_load_input_tensors.mlir --output $OUTPUT_DIR --standalone
+        cp test/ttmlir/EmitC/TTNN/load_input/dump_inputs.py $OUTPUT_DIR
+        cd $OUTPUT_DIR
+        [ -d $OUTPUT_DIR ] || { echo "Directory not found: $OUTPUT_DIR" >&2; exit 1; }
+        python dump_inputs.py
+        ./run
```
svuckovicTT, line 565: Given that we're prioritizing python, let's test that one instead here.

svuckovicTT (approved):
🚀

test/ttmlir/EmitC/TTNN/load_input/arg0.tensorbin
svuckovicTT, line 1: Reminder to remove bins

.github/workflows/build-and-test.yml
svuckovicTT: IMO, separating the tests as @azecevicTT did, divides the test responsibilites clearly and makes them easier to understand, especially in these cases where we're invoking various scripts in shell.

We initially wanted to separate alchemist tests into separate file, but there were some issues that I can't recall right now. After the refactor you're working on, will this be easier to achieve? cc @vvukomanTT who looked into this initially

.github/workflows/build-and-test.yml
azecevic000: Synced offline, because of current limitations of running ttnn, has to revert to serializing tensors AOT and testing with cpp API.

jojo-771771:

jojo-771771:

jojo-771771:

jojo-771771:

Issue #5050, opened 2025-09-22
Create a Pass to Translate TTNN Tensor Layouts to Metal Tensor Layouts


arminaleTT:

jojo-771771:
I can work on this. Please assign!

jojo-771771:
I can work on this. Please assign!

jojo-771771:
I can work on this. Please assign!

jojo-771771:
I can work on this. Please assign!

jojo-771771:
I can work on this. Please assign!

Issue #5019, opened 2025-09-18
Infer Metal Tensors from L1 Block Sharded TTNN Tensors


arminaleTT:

jojo-771771:
I can work on this. Please assign!

jojo-771771:
I can work on this. Please assign!

jojo-771771:
I can work on this. Please assign!

PR #5052, opened 2025-09-22
[D2M/TTNN Integration] Add TTNN Layout Translation Logic in TTIRToTTIRGeneric and Translate Block Sharded TTNN Tensors to Metal Tensors

arminaleTT:
### Ticket
#5050 and #5019 

### Problem description
As part of D2M/TTNN integration, the D2M pipeline will be invoked with ttir ops that have tensor operands with `TTNNLayoutAttr` (i.e. tensors whose layout is set by TTNN). The D2M pipeline needs to: 
1. parse these tensors to run existing D2M passes, and
2. maintain the tensor layouts as set by TTNN to ensure D2M-generated ops and downstream TTNN ops access the tensor in the same way

### What's changed
- Modified the  layout generation logic in the `TTIRToTTIRGeneric` pass. When invoked with the `ttnn-mode=true` option, this pass:
  1. matches ttir ops with TTNN tensor operands and/or results
  2. creates an equivalent metal tensor with fully specified grid/shard shape and `MetalLayout` for each TTNN tensor
  3. inserts a `ttir.ttnn_ttmetal_cast_op` to represent the layout translation
    a. The casts are inserted `ttnn tensor -> metal tensor` for op operands and `metal tensor -> ttnn tensor` for op results
- Added a new `ttnn-mode` option to `TTIRToTTIRGeneric` and connected it to the `ttnn-mode` pipeline option in `ttir-to-ttmetal-pipeline`. When this option is false, the pass asserts when encountering TTNN tensors as the behaviour is undefined
- Currently only block sharded L1 tensors are supported, with more to be added
- Added IR tests, including testing integration with https://github.com/tenstorrent/tt-mlir/pull/4978

### Checklist


include/ttmlir/Dialect/TTIR/Transforms/Passes.td
```diff
@@ -636,4 +636,25 @@ def TTIRFusing: Pass<"ttir-fusing", "::mlir::ModuleOp">
...
 
+def TTNNLayoutTranslation : Pass<"ttnn-layout-translation", "::mlir::ModuleOp"> {
+  let summary = "Translate tensor operands of TTIR ops with TTNNLayoutAttr encoding to equivalent tensors with MetalLayoutAttr";
+  let description = [{
+    This pass translates RankedTensorType values with TTTNLayoutAttr encoding to RankedTensorType values with MetalLayoutAttr encoding if:
+    - the tensor is an operand or resultof a TTIR op, and
+    - the tensor layout can be represented and used in the TTMetalPipeline
+
+    The translation is represented by inserting a ttir.ttnn_metal_layout_cast op.
+    This pass assumes the tensors are device tensors and already allocated. Translated tensors bypass allocation logic in TTMetalPipeline.
```
bgrady-tt: Since tensors don't really have a concept of allocation (this is inserted at bufferization time), I would maybe state instead something like:

> This pass can only cast existing tensors to tensors representing the same underlying data. This pass _does not_ generate new temporaries that would later require allocation and/or layout conversion ala ToLayoutOp.

lib/Dialect/TTIR/Transforms/TTNNLayoutTranslation.cpp
```diff
@@ -0,0 +1,206 @@
...
+        mlir::cast<ttcore::TileType>(ttnnLayout.getElementType()).getHeight() ==
+            ttcore::TileType::getDefaultShape()[0] &&
+        "Only default tile shape is supported");
+    assert(ttnnLayout.hasL1BufferType() &&
+           ttnnLayout.getMemLayout().getValue() ==
+               ttnn::TensorMemoryLayout::BlockSharded &&
+           "Only block sharded L1 tensor memory layout is supported");
+  }
+
+  Type getMetalTensorType(PatternRewriter &rewriter, Value value) const {
```
bgrady-tt: Why not have this return the narrower type `RankedTensorType` ?

bgrady-tt (approved):
LGTM 👍

include/ttmlir/Dialect/TTIR/Transforms/Passes.td
arminaleTT: Thanks for the better wording. Updated

lib/Dialect/TTIR/Transforms/TTNNLayoutTranslation.cpp
arminaleTT: Fixed

lib/Conversion/TTIRToTTIRGeneric/TTIRToTTIRGeneric.cpp
```diff
@@ -71,6 +71,13 @@ class TTIRNamedRewriterCommon {
                               bool tiled,
                               mlir::ConversionPatternRewriter &rewriter) const {
     auto tensorType = mlir::cast<mlir::RankedTensorType>(value.getType());
+    if (tensorType.getEncoding()) {
+      assert(mlir::isa<ttcore::MetalLayoutAttr>(tensorType.getEncoding()) &&
+             "Tensors must have a MetalLayoutAttr or no encoding");
+      // The tensor is already a metal tensor. Maintain the existing layout.
+      return value;
```
nsmithtt: i.e. 
```c++
if (ttnnMode) {
  auto ttnnLayout = mlir::cast<TTNNLayoutAttr>(tensorType.getEncoding())
  return rewriter.create<TTNNMetalLayoutCastOP>(toMetalLayout(ttnnLayout), value)->getResult();
}
```

lib/Conversion/TTIRToTTIRGeneric/TTIRToTTIRGeneric.cpp
```diff
@@ -129,9 +136,17 @@ class TTIRNamedRewriterCommon {
...
-        rewriter.create<tt::ttir::EmptyOp>(fromValue.getLoc(), toResultType);
+                                   Value fromValue, Value toResult) {
+    // If the original op's result is immediately consumed by a
+    // TTNNMetalLayoutCastOp, skip inserting an Empty/ToLayout. This tensor is
+    // already allocated in TTNN, and the layout is set by the cast.
+    for (Operation *user : toResult.getUsers()) {
+      if (mlir::isa<ttir::TTNNMetalLayoutCastOp>(user)) {
+        return fromValue.getDefiningOp();
+      }
+    }
```
nsmithtt, line 206:
```c++
if (ttnnMode) {
  return rewriter.create<TTNNMetalLayoutCastOP>(toTTNNLayout(metalLayout), value)->getResult();
}
```

arminaleTT:
> Thinking a bit more about this, it would seem cleaner in my opinion to just insert the cast during TTIRToTTIRGeneric conversion pass. There we already wrap all ops in to_layout ops, it seems like it'd shim in perfectly to just insert cast ops instead. We could turn TTNNLayoutTranslation.cpp into just a library with the routines for calculating the respective attribute conversions.
> 
> As a standalone pass it feels a bit fragile because there's a lot of bookkeeping we have to do on operands and results that'd likely just work out for free if it were part of the conversion pass instead.
> 
> Let me know what you think.

What you're suggesting is actually how I had originally set this up. That approach certainly works.

During discussions with @xanderchin @bgrady-tt @vtangTT there was a desire to try to isolate this code from core D2M logic and protect for the ability to create a "ttnn-to-d2m" pipeline that would sit in front of the regular D2M pipeline or completely parallel to it. That's why I spun it off into a separate pass. If it's integrated into TTIRToTTIRGeneric, the core D2M flow can never be isolated from the integration/jit flows.

I think this is potentially a bigger conversation. Let me gather some thoughts and I'll bring it up offline

nsmithtt:
> During discussions with @xanderchin @bgrady-tt @vtangTT there was a desire to try to isolate this code from core D2M logic and protect for the ability to create a "ttnn-to-d2m" pipeline that would sit in front of the regular D2M pipeline or completely parallel to it. That's why I spun it off into a separate pass. If it's integrated into TTIRToTTIRGeneric, the core D2M flow can never be isolated from the integration/jit flows.

I agree that isolation is a good rule of thumb, but I'd argue that this isn't isolation because a) with this approach the TTIRToTTIRGeneric conversion pass still had to be modified and taught about the cast op / special cased and b) we require a new pass to run just before to setup the IR for this to work.

I'd propose that if an existing pass is mechanically doing the same transformation, we should use it and teach it about these new ops/config space ideally generalizing the pass.

Not saying we should do this, but in theory we could generalize the the TTIRToTTIRGeneric conversion pass to take a `toLayout` lambda and a `fromLayout` lambda so that users of the conversion could supply their own implementations for materializing the requested layout.

Technically the [TypeConverter](https://mlir.llvm.org/doxygen/classmlir_1_1TypeConverter.html) has exactly this mechanism `addSourceMaterialization` and `addTargetMaterialization` which accepts these lambdas, but I think we weren't able to use it because there's no way for the op to pass desired layout sideband info for constructing the metal layout.

lib/Conversion/TTIRToTTIRGeneric/TTIRToTTIRGenericPass.cpp
```diff
@@ -87,7 +94,7 @@ struct TTIRToTTIRGenericPass final
     mlir::RewritePatternSet patterns{&ctx};
     populateTTIRToTTIRGenericPatterns(
         &ctx, patterns, typeConverter, defaultInputMemSpace,
-        defaultOutputMemSpace, getTargetGridShape());
+        defaultOutputMemSpace, getTargetGridShape(), ttnnMode);
```
nsmithtt, line 97: Minor nit, it might be overkill to pipe the ttnnMode through the rewriters since it's really only used in one place for an assert.  I think the assert is effectively already present because if ttnn mode isn't set conversion will fail since ttnn is not a legal dialect.

nsmithtt (approved):
Minor nit inline, but won't block the PR on it!  Looks great thank you!

lib/Conversion/TTIRToTTIRGeneric/TTIRToTTIRGenericPass.cpp
arminaleTT: I wasn't sure if the dialect being illegal is enough to trigger a failure if the tensor has the ttnn attribute. I'll verify and take this out if that's enough

nsmithtt: Oh yeah just an attribute isn't enough good point, there has to be ops.

arminaleTT: I tried taking out the asserts. We still error out, but it's a vague "failed to legalize ttir.abs". It's totally unclear where the actual failure is, perhaps in the construction of the generic? I think it'll be less error prone to keep the asserts

nsmithtt: Cool, I agree, let's keep them.

jojo-771771:

jojo-771771:

jojo-771771:

Issue #4854, opened 2025-09-10
[Training] Add support for Dropout in Forge


umalesTT:
Dropout is not supported in our Forge compiler, which consumes both Jax and Torch frontends. Compilation for both frontends fails during stablehlo->ttir legalization pass.
For Jax unsupported op is [stablehlo.bitcast_convert](https://github.com/tenstorrent/tt-mlir/issues/979) and for Torch it is [stablehlo.rng_bit_generator](https://github.com/tenstorrent/tt-mlir/issues/4793).

Both of these ops are currently unsupported in metal so one way forward is to fuse dropout operations in stablehlo for both frontends, therefore circumventing those 2 unsupported ops and lowering the fused operation onto [ttnn.experimental.dropout](https://docs.tenstorrent.com/tt-metal/v0.55.0/ttnn/ttnn/api/ttnn.experimental.dropout.html) - which, irregardless of the mentioned ops, should be the way to go efficiency-wise.

So the general workflow would be:

1. Add fusing patterns for dropout in stablehlo for both frontends. In offline talks with @azecevicTT we concluded that for Jax dropout should be treated as composite op and monkey-patched in Jax, much like the [gelu case](https://github.com/tenstorrent/tt-xla/pull/785), while for Torch regular fusing pattern is best until we support composite ops as they are in Jax. Because of this differences, these will be 2 separate issues.
2. Add ```ttir.dropout``` op which will be compatible with both frontend paths from 1.
3. Lower ```ttir.dropout``` to ```ttnn.experimental.dropout```

Concern that remains here is that the experimental metal op should have all [rng algorithms of interest](https://openxla.org/stablehlo/spec#rng_bit_generator) supported as otherwise we won't get same results as CPU.

umalesTT:
After some additional testing and offline discussions, plan for Jax is modified (plan for Torch stays the same).
When dropout in Jax is fused as in initial idea, some occurrences of stablehlo.bitcast_convert still emerge in training and the next largest function we may fuse in order to eliminate mentioned op is jax.random.uniform (https://github.com/tenstorrent/tt-mlir/issues/5066).

Once the stablehlo.bitcast_convert is supported in our stack we will migrate to initial plan for Jax.

sdjordjevicTT:
Dropout op is now fully supported in ttmlir with this PR:
https://github.com/tenstorrent/tt-mlir/pull/6656

PR #5065, opened 2025-09-23
 Add support for stablehlo.rng_bit_generator op

### LINKED ISSUE #4613: Support stablehlo.rng_bit_generator operation
Here is a fwd graph generated with torch_xla -> tt-xla path for this model:

```
class ModelWithDropout(torch.nn.Module):
    def __init__(self):
        super().__init__()
        self.dropout = torch.nn.Dropout(0.5)
    def forward(self, x):
        return self.dropout(x)
```

```
module @SyncTensorsGraph.40 attributes {mhlo.cross_program_prefetches = [], mhlo.input_output_alias = [], mhlo.is_dynamic = false, mhlo.use_auto_spmd_partitioning = false} {
  func.func @main(%arg0: tensor<f32>, %arg1: tensor<i64>, %arg2: tensor<1x3x224x224xf32>) -> tensor<1x3x224x224xf32> {
    %cst = stablehlo.constant dense<1.1920929E-7> : tensor<1x3x224x224xf32>
    %c = stablehlo.constant dense<9> : tensor<1x3x224x224xui32>
    %c_0 = stablehlo.constant dense<0> : tensor<1xui64>
    %c_1 = stablehlo.constant dense<2531011> : tensor<i64>
    %c_2 = stablehlo.constant dense<214013> : tensor<i64>
    %0 = stablehlo.multiply %arg1, %c_2 : tensor<i64>
    %1 = stablehlo.add %0, %c_1 : tensor<i64>
    %2 = stablehlo.convert %1 : (tensor<i64>) -> tensor<ui64>
    %3 = stablehlo.reshape %2 : (tensor<ui64>) -> tensor<1xui64>
    %4 = stablehlo.concatenate %3, %c_0, dim = 0 : (tensor<1xui64>, tensor<1xui64>) -> tensor<2xui64>
    %output_state, %output = stablehlo.rng_bit_generator %4, algorithm =  DEFAULT : (tensor<2xui64>) -> (tensor<2xui64>, tensor<1x3x224x224xui32>)
    %5 = stablehlo.shift_right_logical %output, %c : tensor<1x3x224x224xui32>
    %6 = stablehlo.convert %5 : (tensor<1x3x224x224xui32>) -> tensor<1x3x224x224xf32>
    %7 = stablehlo.multiply %6, %cst : tensor<1x3x224x224xf32>
    %8 = stablehlo.reshape %arg0 : (tensor<f32>) -> tensor<1xf32>
    %9 = stablehlo.broadcast_in_dim %8, dims = [0] : (tensor<1xf32>) -> tensor<1x3x224x224xf32>
    %10 = stablehlo.compare  LT, %7, %9 : (tensor<1x3x224x224xf32>, tensor<1x3x224x224xf32>) -> tensor<1x3x224x224xi1>
    %11 = stablehlo.convert %10 : (tensor<1x3x224x224xi1>) -> tensor<1x3x224x224xf32>
    %12 = stablehlo.broadcast_in_dim %arg0, dims = [] : (tensor<f32>) -> tensor<1x3x224x224xf32>
    %13 = stablehlo.divide %11, %12 : tensor<1x3x224x224xf32>
    %14 = stablehlo.multiply %arg2, %13 : tensor<1x3x224x224xf32>
    return %14 : tensor<1x3x224x224xf32>
  }
}

```

### LINKED ISSUE #4770: stablehlo.rng_bit_generator operation needed for Stable Diffusion XL
### Describe the bug

Here is the last output before SHLO -> MLIR: https://gist.github.com/mstojkovicTT/fba39b4eefcf3d36cd461e272c00346a

I get the following issue: `failed to legalize operation 'stablehlo.rng_bit_generator'`

### Expected behavior

_No response_

_No response_

### LINKED ISSUE #4793: error: failed to legalize operation 'stablehlo.rng_bit_generator'
SHLO:
```
    %output_state, %output = stablehlo.rng_bit_generator %13, algorithm =  DEFAULT : (tensor<2xui64>) -> (tensor<2xui64>, tensor<4x64x12x12xui8>)
```
We encounter similar ir in multiple other places:
```
    %output_state, %output = stablehlo.rng_bit_generator %13, algorithm =  DEFAULT : (tensor<2xui64>) -> (tensor<2xui64>, tensor<4x64x12x12xui8>)
    %output_state_13, %output_14 = stablehlo.rng_bit_generator %18, algorithm =  DEFAULT : (tensor<2xui64>) -> (tensor<2xui64>, tensor<4x128xui8>)
    %output_state, %output = stablehlo.rng_bit_generator %27, algorithm =  DEFAULT : (tensor<2xui64>) -> (tensor<2xui64>, tensor<1x9216xui32>)
    %output_state_14, %output_15 = stablehlo.rng_bit_generator %32, algorithm =  DEFAULT : (tensor<2xui64>) -> (tensor<2xui64>, tensor<1x4096xui32>)
    %output_state, %output = stablehlo.rng_bit_generator %4, algorithm =  DEFAULT : (tensor<2xui64>) -> (tensor<2xui64>, tensor<1x1x1x1xui32>)
    %output_state_63, %output_64 = stablehlo.rng_bit_generator %9, algorithm =  DEFAULT : (tensor<2xui64>) -> (tensor<2xui64>, tensor<1x1x1x1xui32>)
    %output_state_65, %output_66 = stablehlo.rng_bit_generator %14, algorithm =  DEFAULT : (tensor<2xui64>) -> (tensor<2xui64>, tensor<1x1x1x1xui32>)
    %output_state_67, %output_68 = stablehlo.rng_bit_generator %19, algorithm =  DEFAULT : (tensor<2xui64>) -> (tensor<2xui64>, tensor<1x1x1x1xui32>)
    %output_state_69, %output_70 = stablehlo.rng_bit_generator %24, algorithm =  DEFAULT : (tensor<2xui64>) -> (tensor<2xui64>, tensor<1x1x1x1xui32>)
    %output_state_71, %output_72 = stablehlo.rng_bit_generator %29, algorithm =  DEFAULT : (tensor<2xui64>) -> (tensor<2xui64>, tensor<1x1x1x1xui32>)
    %output_state_73, %output_74 = stablehlo.rng_bit_generator %34, algorithm =  DEFAULT : (tensor<2xui64>) -> (tensor<2xui64>, tensor<1x1x1x1xui32>)
    %output_state_75, %output_76 = stablehlo.rng_bit_generator %39, algorithm =  DEFAULT : (tensor<2xui64>) -> (tensor<2xui64>, tensor<1x1x1x1xui32>)
    %output_state_77, %output_78 = stablehlo.rng_bit_generator %44, algorithm =  DEFAULT : (tensor<2xui64>) -> (tensor<2xui64>, tensor<1x1x1x1xui32>)
    %output_state_79, %output_80 = stablehlo.rng_bit_generator %49, algorithm =  DEFAULT : (tensor<2xui64>) -> (tensor<2xui64>, tensor<1x1280xui32>)
```

This is failing in following models (not yet running on tt-xla main):
- test_torch_mnist_cnn_dropout_training
- test_torch_alextnet_training
- test_torch_efficientnet_training

Here are full shlo.mlir files from the models:
[test_torch_alextnet_training_1_4_shlo_after_stablehlo.txt](https://github.com/user-attachments/files/22135972/test_torch_alextnet_training_1_4_shlo_after_stablehlo.txt)
[test_torch_efficientnet_training_1_4_shlo_after_stablehlo.txt](https://github.com/user-attachments/files/22135974/test_torch_efficientnet_training_1_4_shlo_after_stablehlo.txt)
[test_torch_mnist_cnn_dropout_training_1_4_shlo_after_stablehlo.txt](https://github.com/user-attachments/files/22135973/test_torch_mnist_cnn_dropout_training_1_4_shlo_after_stablehlo.txt)

pglusacTT:
### Ticket
Closes #4793, closes #4770, closes #4613
References #4854

### Problem description
Missing conversion for stablehlo.rng_bit_generator op.

### What's changed
Added conversion that decomposes stablehlo.rng_bit_generator to ttir.rand + ttir.typecast.

### Checklist

### DIFF
--- a/lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
+++ b/lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
@@ -2908,6 +2908,8 @@ public:
         mlir::tt::ttir::utils::createDPSOp<mlir::tt::ttir::TypecastOp>(
             rewriter, srcOp.getLoc(), outputType, randOp.getResult());
 
+    // HACK (pglusac): Output state is discarded, initial state is returned as
+    // a result. https://github.com/tenstorrent/tt-mlir/issues/5101
     rewriter.replaceOp(srcOp,
                        {adaptor.getInitialState(), typecastOp.getResult()});
 


lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
```diff
@@ -2867,6 +2867,65 @@ class StableHLOToTTIRRngOpConversionPattern
...
+    auto floatOutputType = mlir::RankedTensorType::get(
+        outputType.getShape(), floatElementType, outputType.getEncoding());
+
+    // Use min and max unsigned int value to cover most of the range of uint32.
+    auto fromFloat = rewriter.getF32FloatAttr(
+        static_cast<float>(std::numeric_limits<unsigned int>::min()));
+    auto toFloat = rewriter.getF32FloatAttr(
+        static_cast<float>(std::numeric_limits<unsigned int>::max()));
+
+    auto seed = rewriter.getUI32IntegerAttr(0);
```
umalesTT, line 2898: nit: maybe leave a comment on why we proceed with default seed = 0

lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
```diff
@@ -2867,6 +2867,65 @@ class StableHLOToTTIRRngOpConversionPattern
...
+
+    rewriter.replaceOp(srcOp,
+                       {adaptor.getInitialState(), typecastOp.getResult()});
+
+    return success();
+  }
+
+private:
+  LogicalResult
+  checkConversionLegality(mlir::stablehlo::RngBitGeneratorOp &srcOp,
```
mmanzoorTT: I suggest to remove `checkConversionLegality` function as you are not verifying anything here.

lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
mmanzoorTT, line 2898: `stablehlo.rng_bit_generator` takes `initial_state` as operand and this should be used as seed value. Always using same seed (seed=0) does not seem right way to go.

lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
```diff
@@ -2867,6 +2867,65 @@ class StableHLOToTTIRRngOpConversionPattern
...
+        mlir::TypeAttr::get(floatElementType), fromFloat, toFloat, seed);
+
+    // TODO (pglusac): Change to bit cast once metal supports it or remove if
+    // rand starts supporting uint32.
+    auto typecastOp =
+        mlir::tt::ttir::utils::createDPSOp<mlir::tt::ttir::TypecastOp>(
+            rewriter, srcOp.getLoc(), outputType, randOp.getResult());
+
+    rewriter.replaceOp(srcOp,
+                       {adaptor.getInitialState(), typecastOp.getResult()});
```
mmanzoorTT, line 2914: Returning `initialState` as `outputState` is bit skeptical specially if we use `outputState` for next `stablehlo.rng_bit_generator` op and it will generate identical random numbers. A possible solution can be that we may perform some operation on `initialState` (e.g. add the `ttir.rand output` with the `initialState` to generate `outputState`. This will resolve the issue of generating identical random number and also add stateful nature as intended with initial/output states).

mmanzoorTT (dismissed):

lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
pglusacTT: TTNN seed=0 is [a special value](https://github.com/tenstorrent/tt-metal/blob/7536da690e0f6442ef8d7561c8fafe32c02ac864/ttnn/cpp/ttnn/operations/rand/device/rand_program_factory.cpp#L110C1-L110C108) that is equivalent to having a random seed.

If we wanted to actually use the seed that was produced by stablehlo, ttnn rand should be taking in a tensor instead of uint32, or we would need to return that value to host, extract the needed value, etc. which is really bad.

I'm happy to improve this if you have ideas on how we could approach this

lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
pglusacTT: This ties in with seed=0 being a random seed, if we find a non-hacky and neat way to actually use the seed produced by the graph then I agree that the "add approach" you mentioned is definitely the way to go and would add reproducibility.

The good news is that the operations we encountered thus far do not actually use the output state. You can find a few examples that have this operation [here](https://github.com/tenstorrent/tt-mlir/issues/4793), and you can see that the output state is not used.

Another idea we had for reproducibility (which could be an extension for this) is to still set the seed on the host side, disregarding the graph computation of the seed on device like we do now, but to set the seed by deterministically computing it on the host side. This could be done by hashing the loc of the input to the op or something to that extent. 
We haven’t yet found a strong use case for this, but I’d be interested in your thoughts.

vladimirjovanovicTT: We will need reproducibility for rand but I believe it is outside of scope for this PR; we can create a separate issue for this and tackle it once the need for reproducible training runs is a high priority (we are still bootstrapping training at this point).

lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
```diff
@@ -2867,6 +2867,52 @@ class StableHLOToTTIRRngOpConversionPattern
...
+    auto toFloat = rewriter.getF32FloatAttr(
+        static_cast<float>(std::numeric_limits<unsigned int>::max()));
+
+    auto seed = rewriter.getUI32IntegerAttr(0);
+
+    auto randOp = rewriter.create<mlir::tt::ttir::RandOp>(
+        srcOp.getLoc(), floatOutputType, rewriter.getI32ArrayAttr(size),
+        mlir::TypeAttr::get(floatElementType), fromFloat, toFloat, seed);
+
+    // TODO (pglusac): Change to bit cast once metal supports it or remove if
```
anuragsingh-tt: [From our coding guidelines:](https://docs.tenstorrent.com/tt-mlir/coding-guidelines.html)

> A TT-MLIR issue should be created and linked inline to track the TODO.

pglusacTT: Done here [eba17c6](https://github.com/tenstorrent/tt-mlir/pull/5065/commits/eba17c6e73aeb6b859d65380b40a9d82fb574b19)

lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
mmanzoorTT: I had seen some examples where the initialState was computed using some constants but it seems that it is not always the case.

lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
mmanzoorTT: PyTorch does not use the output_state but JAX does use output_state for next random generation. We may require it for JAX models bringup.

lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
```diff
@@ -2867,6 +2867,53 @@ class StableHLOToTTIRRngOpConversionPattern
...
+
+    auto floatElementType = rewriter.getF32Type();
+    auto floatOutputType = mlir::RankedTensorType::get(
+        outputType.getShape(), floatElementType, outputType.getEncoding());
+
+    // Use min and max unsigned int value to cover most of the range of uint32.
+    auto fromFloat = rewriter.getF32FloatAttr(
+        static_cast<float>(std::numeric_limits<unsigned int>::min()));
+    auto toFloat = rewriter.getF32FloatAttr(
+        static_cast<float>(std::numeric_limits<unsigned int>::max()));
```
azecevic000, line 2894: TBH don't know what would be the distribution of individual bits in each range, did you try to find something that would yield an acceptable result? For example, float uses one bit for sign, so it's always going to be `0` in this case, might be a better idea to use some other range.

pglusacTT: So for float32->uint32 for typecast we really do not have much of a choice, this is the best we can do.
I ran dropout using this rng_bit_generator and measured how often are the weights masked and I get pretty close to the percentage I set.
I totally agree that some bits will be underrepresented, and this isn't totally random, but this is both good enough and the best we can do.

pglusacTT: To be precise, with the rate set to 0.4, I sampled dropout 1000 weights and 417 of them were dropped (about 41.7%). The null hypothesis here is that dropout is truly random with probability 0.4. Binomial test gave a p-value of 0.38, which means the difference between 41.7% and 40% is well within what you’d expect.

I used the Ljung–Box test to test for correlation between the drops, where the null is that there’s no correlation up to lag 10. The p-value came out to 0.23, so the dropped values aren't clustering in that range, null wasn't rejected.

I think this is enough evidence that this is behaving as expected, but even if something was missed, we'll still replace the typecast with bitcast once we have it, and then it will surely be fine.

lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
pglusacTT: Done here [f72999a](https://github.com/tenstorrent/tt-mlir/pull/5065/commits/f72999a8330cd7796f754a2d222c2042ba737417)

lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
sgligorijevicTT: @mmanzoorTT I have never observed Jax targetting `rng_bit_generator` op. Everything related to randomness I tried was handled by emitting separate functions that implement the exact generator Jax expects via elementwise ops. Do you have an example of Jax using this op?

sgligorijevicTT: With the above being said, @pglusacTT this is a **major** hack, can you please at least leave a comment in the code containing `HACK`(for easier grepping later) and make an issue and link it there.

lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
azecevic000: Sounds good, thank you for the detailed explanation!

lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
pglusacTT: @sgligorijevicTT agreed, done here [7ad3a6f](https://github.com/tenstorrent/tt-mlir/pull/5065/commits/7ad3a6f02c2b27fea7979f43b5df280d5b6a2b93)

jojo-771771:

jojo-771771:

PR #5072, opened 2025-09-23
Change default tensorL1UsageCap to 100% and Pass pipeline option cap to Validation pass

bmalesevicTT:
This PR updates the default percentage of L1 memory considered usable in the Optimizer and related passes. Previously, the default was set to 80%; it is now increased to 100% (i.e., full L1 capacity). Specifically, both analyses in the Optimizer (L1IFA and MLA) now use the updated option value.

Additionally, the newer post-Optimizer validation pass now receives this usage cap as a pipeline option, rather than relying on a hardcoded value. This ensures that the validation logic aligns with the cap used in earlier passes. When the cap was set to 0.8, the validation pass would revert changes that utilized more memory, which significantly degraded performance for the Resnet model in particular. With the cap set to 1.0, such degradation is avoided, and performance improves.

While an intermediate cap value might be considered in the future if "L1 out of memory" runtime error or other issues arise, there is currently no observed need for extra safety.

### Checklist

### DIFF
--- a/lib/Dialect/TTNN/Transforms/OptimizerPasses/OperationValidationAndFallback.cpp
+++ b/lib/Dialect/TTNN/Transforms/OptimizerPasses/OperationValidationAndFallback.cpp
@@ -4,6 +4,7 @@
 
 #include "ttmlir/Dialect/TTNN/IR/TTNNOps.h"
 #include "ttmlir/Dialect/TTNN/IR/TTNNOpsAttrs.h"
+#include "ttmlir/Dialect/TTNN/Transforms/Passes.h"
 #include "ttmlir/Dialect/TTNN/Utils/Utils.h"
 #include "ttmlir/Dialect/TTNN/Validation/OpConstraintValidation.h"
 #include "ttmlir/OpModel/TTNN/SingletonDeviceContext.h"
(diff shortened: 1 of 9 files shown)


azecevicTT (approved):

lib/Dialect/TTNN/Transforms/OptimizerPasses/OperationValidationAndFallback.cpp
```diff
@@ -76,11 +77,12 @@ void generateAllCombinations(
     const std::vector<std::vector<TTNNLayoutAttr>> &operandFallbacks,
     const std::vector<TTNNLayoutAttr> &originalInputLayouts, size_t operandIdx,
     std::vector<TTNNLayoutAttr> &currentCombination, double currentDistance,
-    std::vector<CombinationCandidate> &allCombinations);
+    std::vector<CombinationCandidate> &allCombinations, float tensorL1UsageCap);
```
rpavlovicTT, line 80: I don't think you need it in this function.

bmalesevicTT: Thanks, my bad, but since this was already merged, here is the cleanup: https://github.com/tenstorrent/tt-mlir/pull/5117

jojo-771771:

jojo-771771:

jojo-771771:

PR #4791, opened 2025-09-04
[MultiHost] runtime client implementation

jnie-TT:
### Ticket
Closes #4483 

### Problem description
We need a standalone executable that can be dispatched to different hosts and execute runtime APIs/flatbuffers on the dedicated device

### What's changed

Added RPC infrastructure for server-client communication

#### Socket
Client and server socket classes for communication between server and client.

#### Command
Added `command.fbs` flatbuffer schema file defining various commands that can be sent to and executed on the client. Added `CommandFactory` for mapping user arguments into flatbuffer commands (will be integrated with runtime APIs later).

#### Response
Added `response.fbs` flatbuffer schema file defining various responses sent from client to server. 
Added `ResponseFactory` for mapping command execution outputs to flatbuffer responses. These responses will be handled by the server to update data structures, poll command completion, and retrieve client data results.

#### Command Executor
Main entry point of the runtime client for executing various commands and sending corresponding responses. 

#### Runtime Client
Standalone executable that connects to server socket and continuously polls/executes commands.

#### Testing
Added runtime gtest that creates a subprocess running the runtime client and executes a flatbuffer e2e on the client. The same flatbuffer is executed locally on the main process and results are compared for validation.

### Checklist
- [X] New/Existing tests provide coverage for changes

PR #5076, opened 2025-09-23
[Multihost] Runtime controller implementation and distributed runtime

### LINKED ISSUE #4484: Runtime server implementation and API integration
Implement a runtime server that can fork multiple instances of runtime clients on multiple processes/hosts, and dispatch commands to them. This should be done under the hood and should hook into the existing runtime APIs such that it is transparent to users and porting should be seamless.

jnie-TT:
### Ticket
Closes #4484 

### Problem description
As part of the multi host effort, we need a distributed runtime implementation abstracted behind runtime APIs that can dispatch commands to the runtime worker implemented in #4791 

### What's changed
* Added distributed host runtime as a brand new code path for running runtime APIs. 

* Added a runtime controller that is the backend of the distributed runtime.
  * The controller launches runtime worker processes, connects to them, and dispatch runtime API commands.
  * Currently, the process is launched as a subprocess on the same host node. The next step is to launch on remote hosts mapping to different devices
  * The controller is currently a singleton, but it's easily updatable to be per host cluster as needed when we want to support multi mesh and MPMD.
  * The controller has a multi-threaded infra that can execute commands synchronously/asynchronously. `getCurrentSystemDesc`, `memcpy`, and `shutdown` are synchronously executed; runtime APIs on the critical path like `getLayout`, `toLayout`, and `submit` are asynchronous.
  
* Added runtime pytest that executes a flatbuffer end to end in a loop through distributed host runtime, and verifies the output matches golden. Also manually verified that consteval runs as expected. Would be a good TODO to bring up the runtime debug stat API to distributed runtime so that we can verify consteval within the test script.
  * To not double the CI load, I didn't add a separate test that runs all flatbuffers through the distributed runtime, since as long as we are able to run any flatbuffer, the other flatbuffers should just work because the actual execution on the worker side is the same as if it were local execution. The main thing to test here is the controller/worker communication and RPC rather than the actual execution itself.

### Checklist


vvukomanTT (approved):

nsumrakTT (approved):

runtime/include/tt/runtime/types.h
```diff
@@ -164,6 +183,63 @@ struct RuntimeCheckedConstObjectImpl {
 
 } // namespace detail
 
+struct SharedHandle {
```
nsmithtt: What is the purpose of this class? Hard for me to tell why it couldn't be `using SharedHandle = std::shared_ptr<void>`.

runtime/include/tt/runtime/utils.h
```diff
@@ -156,6 +156,8 @@ void handleBFloat16ToFloat16(const uint16_t *oldBuffer, uint16_t *newBuffer,
 
 } // namespace detail
 
+std::string getMlirHome();
```
nsmithtt, line 159: I didn't see this from last time, I'd really like to avoid env vars / paths.  I'll try and comment on each use with a proposed workaround.

runtime/include/tt/runtime/detail/distributed/server/server.h
```diff
@@ -0,0 +1,190 @@
...
+  Server() = default;
+  ~Server();
+
+  Server(const Server &) = delete;
+  Server &operator=(const Server &) = delete;
+  Server(Server &&) = delete;
+  Server &operator=(Server &&) = delete;
+
+  // Launches a local subprocess that will connect to the server
+  void launchLocalSubprocess(uint16_t serverPort);
```
nsmithtt: Is local subprocess just for testing? It seems like the onus is on some configuration layer to initialize all of the clients before the server launches.

runtime/include/tt/runtime/types.h
jnie-TT: Essentially we need a nested pointer structure to support asynchronous execution where multiple objects share access to data that gets populated later.

`SharedHandle` wraps a nested shared pointer structure to solve this:
  - Handles are often null created before the underlying data is available
  - Multiple objects need to share the same handle
  - When async data arrives, all shared copies should see the update
 
 Using system descriptor as an example here but this could apply to any data structure:
```
// The system descriptor contains a shared handle that's null initialized
// Might not be populated here since the execution is async
// The runtime server will also hold a copy of the shared handle that 
// it will populate once it gets a response from the client
auto systemDesc = getSystemDescAsync();

// systemDesc2 contains a copy of the shared handle, might also not be populated
auto systemDesc2 = systemDesc;

// Once the API finishes execution on the client and is sent back to the server
// the sharedHandle will get populated by the server
// and now both systemDesc and systemDesc2 will see a brand new handle that points to the system 
// descriptor. 
// Outer shared_ptr: shared among all copies for the event/promise
// Inner shared_ptr: gets swapped from null to actual data, and all copies see the change
systemDesc.wait();
```

runtime/include/tt/runtime/utils.h
jnie-TT: Yeah I agree, I don't like it either, currently using it to locate the path to the runtime client executable. I opted to do it this way because:
* This is how we locate flatbuffers in ttrt
* This environment variable gets set in our env/activate script and so is fairly seamless

But if there's a better way I'll definitely update to move away from env vars

runtime/include/tt/runtime/detail/distributed/server/server.h
jnie-TT: Yeah this is mainly for testing. I added a TODO comment below this, the actual way would be to add extra parameters to the `launchDistributedServer` API such that the user can specify MPI arguments and potentially tt-run arguments if we want to use that so that we can launch through an MPI launcher on multiple hosts. This is the next task/milestone on the list - to launch through MPI and actually run a multi-host workload

runtime/include/tt/runtime/utils.h
nsmithtt: Yeah in light of the convo last week, it seems like we should be wrapping our clients in tt-run so we'll have to lookup that script on disk too.  Could be a follow on change, but I'm thinking we just repurpose `tt::runtime::debug::Env` and move that out of debug only, tbh I think most of the features it has currently would be desirable in release builds too.  Then we can add this kind of environment info underneath that singleton.

runtime/include/tt/runtime/types.h
nsmithtt: I see, are these shared handles ever accessed from different threads?

jnie-TT: They shouldn't be accessed across multiple threads, but they are shared between the user thread and the response thread (the thread that will populate the handle). 

I was debating if I should add a mutex to protect the handle, but in reality two threads should never access the handle simultaneously - the user should only be able to extract info/data from the handle through APIs, and the APIs should just wait on the handle and only when the event is populated (in this case the handle will exist only on the user thread) will the API access the handle.

runtime/include/tt/runtime/utils.h
jnie-TT: Yeah that makes sense, I guess we also have a `tt::runtime::RuntimeContext` singleton that tracks global configs like current runtime and fabric config, might be a good place to store the mlir home. We could add a runtime API `setMlirHome` for the user to configure it

runtime/include/tt/runtime/types.h
nsmithtt: Why can't the handles (aka shared_ptr<void>) just be copied like value types into the queue?  It seems like the queue is the shared resource, the data being passed back and forth is also shared in some sense, but it's not clear to me why they couldn't be passed as values.

If there's some truly async api, maybe we can use `std::future`?

jnie-TT: * I guess the use-case I was trying to cover for was that if the handle value type can't be default initialized. In that case, we would need to set it to `nullptr` intitially, and then we would need to create an entirely new shared_ptr once the response arrives, and for all handles to see the update, we would swap the nested pointer (initialized to null) with the newly created response shared_ptr. But I think you're right currently we don't need this nested structure, I'll update to use just `shared_ptr` as the handle, and if we ever need this in the future we can just bring it back. 

* That's exactly what we're doing. Each event is essentially an `std::future`, and the promise will be passed into the response handler thread. Once a response comes back from the worker, the handle will be populated and the promise will be set to signal to everybody waiting on the `std::future` event.

nsmithtt: I think my preference is:
- If we can get away with value semantics, let's just pass std::shared_ptr around.
- Otherwise let's use std::future

runtime/include/tt/runtime/utils.h
nsmithtt: Yeah that works too!

runtime/include/tt/runtime/types.h
jnie-TT: I removed `SharedHandle` and updated to just pass around copies of `std::shared_ptr<void>`

I think we'll still need `std::future` since that'll be our event to wait on if we want to block on any of the async APIs. It'll just be an `std::future<void>` though that'll just signal when a handle is populated. Merging the 2 into just an `std::future<std::shared_ptr<void>>` doesn't work because we have APIs like memcpy where we have a user-provided data pointer to begin with that we need to populate instead of creating a new shared_ptr entirely.

runtime/include/tt/runtime/utils.h
jnie-TT: Created issue #5146 to track this

jojo-771771:

jojo-771771:

jojo-771771:

PR #5081, opened 2025-09-24
multi tile fix

### LINKED ISSUE #5080: [D2M] Add support for multi-tile indexing into dest register
### Describe the bug

Currently all intermediate loads and stores in fused chains don't factor in loops used for multi tile indexing and assumed to use 0,0 indexing.

### Expected behavior

_No response_

_No response_

dlokeTT:
### Ticket
Closes #5080 

### Problem description
Assumed that we were only indexing 0, 0 for the tile so it would fail in multi tile situations.

### What's changed
Use the loop variables to get which tile we should be indexing in multi tile situations.

### Checklist

### DIFF
--- a/lib/Dialect/TTIR/Transforms/InsertDstRegisterAccess.cpp
+++ b/lib/Dialect/TTIR/Transforms/InsertDstRegisterAccess.cpp
@@ -600,6 +600,14 @@ public:
         storeIndices.push_back(rewriter.create<arith::ConstantIndexOp>(loc, 0));
       }
 
+      // Ensure storeIndices matches the destination memref rank.
+      assert(storeIndices.size() == dstRank &&
+             "storeIndices size must match destination memref rank. If it's "
+             "greater, probably need to use getNonParticipatingLoopDims to "
+             "skip loop dimensions: "
+             "https://github.com/tenstorrent/tt-mlir/pull/"
+             "5081#discussion_r2376709558");
+
       auto storeMap =
           AffineMap::getMultiDimIdentityMap(dstRank, rewriter.getContext());
 


lib/Dialect/TTIR/Transforms/InsertDstRegisterAccess.cpp
```diff
@@ -299,7 +303,19 @@ struct TTIRInsertDstRegisterAccessRewriter final
...
-          dstRegisterAllocation[op] = allocatedIndex;
+
+          // Collect induction variables from containing loops
+          SmallVector<Value> loopInductionVars;
+          for (Operation *ancestor = op->getParentOp(); ancestor != nullptr;
+               ancestor = ancestor->getParentOp()) {
+            if (auto affineFor = dyn_cast<affine::AffineForOp>(ancestor)) {
+              loopInductionVars.push_back(affineFor.getBody()->getArgument(0));
+            }
+          }
```
nsmithtt: I think technically we need to use `getNonParticipatingLoopDims` and skip the loop dims that this set returns.  This'll be a requirement for matmul fusion I believe.  That said, I don't have concrete IR to base this on, so I think it'll just fail because the number of loopInductionVars here will exceed the rank of the memref that's being loaded/stored to.

nsmithtt (approved):

lib/Dialect/TTIR/Transforms/InsertDstRegisterAccess.cpp
dlokeTT: Hmm yea that makes sense. Should I include a check in this pr or just be on the lookout for when we get to matmul fusion?

nsmithtt: Maybe it's better to add an assert for now, below where we plug this into the load/store op, we can check that the resulting indices match the rank of the memref.  We could add some comment in the assert or link it to this thread in the message.

xanderchin (approved):

jojo-771771:

jojo-771771:

jojo-771771:

PR #5082, opened 2025-09-24
[D2M] Interleaved DRAM layout support

bgrady-tt:
### Problem description
D2M x TTNN integration requires support for interleaved DRAM layouts (the only DRAM layout supported).

### What's changed
- A new layout attribute called `InterleavedLayoutAttr` conforming to `DeviceLayoutInterface` can be used as the layout attribute on a memref. 
  - This behaves similarly to `ShardLayoutAttr`, but results in an interleaved-compatible affine map being generated by `DeviceAttr::getMemoryMap`
  
- Introduces a new interface in TTCoreAttrInterfaces called `BufferedDeviceLayoutInterface` that only InterleavedLayoutAttr and ShardLayoutAttr conform to
  - Sole function in API is `getBuffers()` for now
  
- `MetalLayoutAttr` has a new enum parameter `TensorMemoryLayout` that closely matches the existing layout categorization used by TTNNLayoutAttr
  - Layout can be either interleaved, width_sharded, height_sharded, or block_sharded
  - Defaults to block sharded
  - Allows casting tensors with TTNNLayoutAttr to MetalLayoutAttr without losing this layout information
  
- A new pair of functions `isPhysical`/`isView` have been added to the DeviceLayoutInterface API
  - Previously the presence of `ShardLayoutAttr` was enough to determine if the memref was to a physical device tensor
  - Allows differentiating between memrefs that are physical versus views

### Checklist

### DIFF
--- a/include/ttmlir/Dialect/TTCore/IR/TTCoreAttrInterfaces.td
+++ b/include/ttmlir/Dialect/TTCore/IR/TTCoreAttrInterfaces.td
@@ -52,6 +52,22 @@ def TTCore_DeviceLayoutInterface : AttrInterface<"DeviceLayoutInterface"> {
         return std::accumulate(shardShape.begin(), shardShape.end(), 1LL, std::multiplies<int64_t>());
       }]
     >,
+    InterfaceMethod<
+      "Return true if the layout is a view on top of a physical layout",
+      "bool", "isView", (ins),
+      /*methodBody=*/"",
+      /*defaultImplementation=*/[{
+        return false;
+      }]
+    >,
+    InterfaceMethod<
+      "Return true if the layout is physical rather than a view",
+      "bool", "isPhysical", (ins),
+      /*methodBody=*/"",
+      /*defaultImplementation=*/[{
+        return !isView();
+      }]
+    >,
   ];
 }
 
--- a/include/ttmlir/Dialect/TTCore/IR/TTCoreOpsEnums.td
+++ b/include/ttmlir/Dialect/TTCore/IR/TTCoreOpsEnums.td
@@ -196,6 +196,18 @@ def TTCore_ArgumentType : I32EnumAttr<"ArgumentType", "Argument Type",
   let symbolToStringFnName = "ArgumentTypeEnumToString";
 }
 
+def TTCore_TensorMemoryLayout_Interleaved : I32EnumAttrCase<"Interleaved", 0, "interleaved">;
+def TTCore_TensorMemoryLayout_Sharded     : I32EnumAttrCase<"Sharded",     1, "sharded">;
+
+def TTCore_TensorMemoryLayout : I32EnumAttr<"TensorMemoryLayout", "TTCore Tensor Memory Layout",
+                           [
+                            TTCore_TensorMemoryLayout_Interleaved,
+                            TTCore_TensorMemoryLayout_Sharded,
+                           ]> {
+  let genSpecializedAttr = 0;
+  let cppNamespace = "::mlir::tt::ttcore";
+}
+
 def TTCore_ShardStatus_Presharded: I32EnumAttrCase<"Presharded", 0, "presharded">;
 def TTCore_ShardStatus_Unsharded: I32EnumAttrCase<"Unsharded", 1, "unsharded">;
 
--- a/include/ttmlir/Dialect/TTCore/IR/TTCoreOpsTypes.td
+++ b/include/ttmlir/Dialect/TTCore/IR/TTCoreOpsTypes.td
@@ -250,6 +250,8 @@ def TTCore_ViewLayoutAttr : TTCore_Attr<"ViewLayout", "view", [TTCore_DeviceLayo
   let extraClassDeclaration = [{
       // Compose two view layouts f(g(x)) where f=this and g=other.
       ViewLayoutAttr compose(ViewLayoutAttr g) const;
+
+      bool isView() { return true; }
   }];
 }
 
@@ -273,6 +275,29 @@ def TTCore_ShardLayoutAttr : TTCore_Attr<"ShardLayout", "shard", [TTCore_DeviceL
     static ShardLayoutAttr get(::mlir::MLIRContext *context, ArrayRef<int64_t> shape, uint64_t elementSize, uint32_t buffers);
     static ShardLayoutAttr get(ArrayRef<int64_t> shape, Type elementType, uint32_t buffers);
     static ShardLayoutAttr get(MemRefType memrefType, uint32_t buffers);
+
+    bool isView() { return false; }
+
+    AffineMap getAffineMap() const;
+  }];
+}
+
+def TTCore_InterleavedLayoutAttr : TTCore_Attr<"InterleavedLayout", "interleaved", [TTCore_DeviceLayoutInterface, MemRefLayoutAttrInterface]> {
+  let summary = "Interleaved layout attribute in TT dialect";
+  let description = [{
+    Describes overall layout of an interleaved memref buffer.
+    - Stride: Stride of each dim in bytes.
+  }];
+  let parameters = (ins ArrayRefParameter<"int64_t">:$stride);
+
+  let assemblyFormat = "`<` custom<DimensionList>($stride) `>`";
+
+  let extraClassDeclaration = [{
+    static InterleavedLayoutAttr get(::mlir::MLIRContext *context, ArrayRef<int64_t> shape, uint64_t elementSize);
+    static InterleavedLayoutAttr get(ArrayRef<int64_t> shape, Type elementType);
+    static InterleavedLayoutAttr get(MemRefType memrefType);
+
+    bool isView() { return false; }
 
     AffineMap getAffineMap() const;
   }];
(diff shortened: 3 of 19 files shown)


lib/Dialect/TTCore/IR/TTCoreOpsTypes.cpp
```diff
@@ -1330,35 +1352,67 @@ DeviceAttr DeviceAttr::get(::mlir::MLIRContext *context,
...
+
+    // interleaved layout is either single tile or stride of outermost dim
+    int64_t elementSizeBytes = getElementSizeBytes(memrefType.getElementType());
+    pageSize = mlir::isa<ttcore::TileType>(memrefType.getElementType())
+                   ? elementSizeBytes
+                   : interleavedLayout.getStride().front();
+
+    // interleaved layout for DRAM is constrained to have unit grid dims,
+    // similar to TTNNLayoutAttr convention
+    assert(ttmlir::utils::volume(interleavedLayout.getGridShape(memrefType)) ==
```
bgrady-tt, line 1397: I would love to push this check up into a verify() function somewhere. This would however have to be verified against each _memref_, as the InterleavedLayoutAttr by itself doesn't have the grid shape. 

I think enforcing this within `GenericOp::verify()` makes the most sense, as each memref should be an input/output of a GenericOp (at least for the foreseeable future).

lib/Dialect/TTCore/IR/TTCoreOpsTypes.cpp
```diff
@@ -555,22 +555,40 @@ ShardLayoutAttr ShardLayoutAttr::get(mlir::MemRefType memrefType,
 }
 
 mlir::AffineMap ShardLayoutAttr::getAffineMap() const {
-  auto *context = getContext();
-  int64_t rank = getStride().size();
-  SmallVector<mlir::AffineExpr> mapExprs(rank + 1);
+  return ttmlir::utils::generateAffineMapFromShardStrides(getStride(),
```
bgrady-tt, line 558: note: this affine map gen code is identical between ShardLayoutAttr and InterleavedLayoutAttr, so I extracted it as a util

lib/Target/TTMetal/TTMetalToFlatbuffer.cpp
```diff
@@ -275,20 +278,43 @@ memrefTypeToInterleavedBufferConfigFlatbuffer(FlatbufferObjectCache &cache,
...
+    int64_t numPages = mlir::isa<ttcore::TileType>(memref.getElementType())
+                           ? memref.getNumElements()
+                           : layout.getShardShape(memref).front();
+    int64_t size = ttmlir::utils::alignUp(numPages * pageSize, pageSize);
+
+    config = target::metal::CreateInterleavedBufferConfig(*cache.fbb, size,
+                                                          pageSize);
+
+  } else {
+    // Infer interleaved buffer config for tiled tensors in absence of an
```
bgrady-tt, line 328: We should sunset support for inferring interleaved layout when layout is absent, now that there is an explicit attribute. However this is probably best to do once this is merged and Polymage has some time to migrate to generating memref with InterleavedLayoutAttr consistently.

include/ttmlir/Dialect/TTCore/IR/TTCoreAttrInterfaces.td
```diff
@@ -52,6 +52,32 @@ def TTCore_DeviceLayoutInterface : AttrInterface<"DeviceLayoutInterface"> {
...
     >,
+    InterfaceMethod<
+      "Return true if the layout is physical rather than a view",
+      "bool", "isPhysical", (ins),
+      /*methodBody=*/""
+    >,
+    InterfaceMethod<
+      "Return true if the layout is a view on top of a physical layout",
+      "bool", "isView", (ins),
+      /*methodBody=*/""
```
nsmithtt: This and above are always mutually exclusive right? We could have a default impl for this `return !isPhysical();`

include/ttmlir/Dialect/TTCore/IR/TTCoreOpsTypes.td
```diff
@@ -259,6 +262,32 @@ def TTCore_ShardLayoutAttr : TTCore_Attr<"ShardLayout", "shard", [TTCore_DeviceL
...
     static ShardLayoutAttr get(MemRefType memrefType, uint32_t buffers);
 
+    bool isView() { return false; }
+    bool isPhysical() { return true; }
+
+    AffineMap getAffineMap() const;
+  }];
+}
+
+def TTCore_InterleavedLayoutAttr : TTCore_Attr<"InterleavedLayout", "interleaved", [TTCore_DeviceLayoutInterface, MemRefLayoutAttrInterface, TTCore_BufferedDeviceLayoutInterface]> {
```
nsmithtt: Can this really conform to `BufferedDeviceLayoutInterface`? I'm thinking this is reserved for CBs which only make sense in sharded mode right?

include/ttmlir/Dialect/TTCore/IR/TTCoreOpsEnums.td
```diff
@@ -196,6 +196,22 @@ def TTCore_ArgumentType : I32EnumAttr<"ArgumentType", "Argument Type",
...
+def TTCore_TensorMemoryLayout_HeightSharded : I32EnumAttrCase<"HeightSharded", 1, "height_sharded">;
+def TTCore_TensorMemoryLayout_WidthSharded : I32EnumAttrCase<"WidthSharded", 2, "width_sharded">;
+def TTCore_TensorMemoryLayout_BlockSharded : I32EnumAttrCase<"BlockSharded", 3, "block_sharded">;
+
+def TTCore_TensorMemoryLayout : I32EnumAttr<"TensorMemoryLayout", "TTCore Tensor Memory Layout",
+                           [
+                            TTCore_TensorMemoryLayout_Interleaved,
+                            TTCore_TensorMemoryLayout_HeightSharded,
+                            TTCore_TensorMemoryLayout_WidthSharded,
+                            TTCore_TensorMemoryLayout_BlockSharded,
```
nsmithtt: Can't we get away with just Interleaved and Sharded?  I thought we were going to use `MetalLayout.index_map` to encode all virtual layouts including height/width sharded.

lib/Dialect/TTCore/IR/TTCoreOpsTypes.cpp
nsmithtt: Man this interleaved layout feels like a special case of our dram sharded layout.  I guess it's OK to be explicit, but the only real diff is the pageSize right?

include/ttmlir/Dialect/TTCore/IR/TTCoreOpsEnums.td
bgrady-tt: I believe we can get away with just Interleaved and Sharded. Although I like the idea of pushing MetalLayoutAttr to be a superset of TTNNLayoutAttr so eventual unification of the two is simpler.

include/ttmlir/Dialect/TTCore/IR/TTCoreOpsTypes.td
bgrady-tt: Right, interleaved layout can't be a CB so there would only ever be a single backing buffer. I will strip this out.

include/ttmlir/Dialect/TTCore/IR/TTCoreAttrInterfaces.td
bgrady-tt: I can change isView to be defined in terms of isPhysical 👍

bgrady-tt:
@nsmithtt Couldn't reply to comment directly:

For DRAM, we could model interleaved layouts as a special case of ShardedLayoutAttr. It is basically just decoupling page size from shard size. This is possible for DRAM because the page->bank mapping in D2M is ultimately page-interleaved, where page normally equals shard.

For **L1 interleaved**, there is no clean way to model distribution correctly for non-unit grids. You can set page size to be smaller, but as-is shards would still be distributed as contiguous on each worker instead of interleaved. This is where having a different Attr will likely help.

nsmithtt:
> @nsmithtt Couldn't reply to comment directly:
> 
> For DRAM, we could model interleaved layouts as a special case of ShardedLayoutAttr. It is basically just decoupling page size from shard size. This is possible for DRAM because the page->bank mapping in D2M is ultimately page-interleaved, where page normally equals shard.
> 
> For **L1 interleaved**, there is no clean way to model distribution correctly for non-unit grids. You can set page size to be smaller, but as-is shards would still be distributed as contiguous on each worker instead of interleaved. This is where having a different Attr will likely help.

Cool ok that makes sense, thanks!

nsmithtt (approved):

lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
```diff
@@ -111,14 +111,20 @@ class TTIRNamedRewriterCommon {
     DenseIntElementsAttr collapsedIntervals =
         DenseIntElementsAttr::get(intervalTy, llvm::ArrayRef<int64_t>({0, -1}));
 
+    ttcore::TensorMemoryLayout memLayout =
```
bgrady-tt, line 123: @arminaleTT This is all that is needed for translating interleaved DRAM layouts. Everything else is identical to existing logic for sharded layouts. This will likely change with L1 interleaved support, but for now its pretty simple

bgrady-tt:
@nsmithtt implemented your suggestions, please take another look!

include/ttmlir/Utils.h
```diff
@@ -121,6 +121,32 @@ replaceAffineMapSymbols(mlir::AffineMap map, mlir::ArrayRef<int64_t> symbols) {
                                    map.getNumDims(), numResultSyms);
 }
 
+// Generates an affine map translating ND grid x ND shard coordinates into ND
+// grid + a linearized offset i.e. collapsing all shard dims into a single
+// byte offset based on provided shard strides.
+//
+// i.e. if strides = [4, 2] --> (g0, g1, s0, s1) -> (g0, g1, 4*s0 + 2*s1)
```
vwellsTT, line 128: Nit: I found this syntax a bit hard to parse, maybe 
"if strides = [4, 2], then (g0, g1, s0, s1) -> (g0, g1, 4*s0 + 2*s1)." is clearer

lib/Dialect/TTCore/IR/TTCoreOpsTypes.cpp
```diff
@@ -1331,35 +1347,67 @@ DeviceAttr DeviceAttr::get(::mlir::MLIRContext *context,
...
+    }
+  } else if (mlir::isa<ttcore::InterleavedLayoutAttr>(memrefType.getLayout())) {
+
+    assert(memorySpace == MemorySpace::DeviceDRAM &&
+           "interleavedLayoutAttr only supported for deviceDRAM memory space");
+
+    auto interleavedLayout =
+        mlir::cast<ttcore::InterleavedLayoutAttr>(memrefType.getLayout());
+
+    // interleaved layout is either single tile or stride of outermost dim
```
vwellsTT, line 1389: nit: pls sentence-ify all comments

test/python/golden/test_metal_dma.py
```diff
@@ -234,3 +234,73 @@ def dram_write(
...
+    def interleaved_dma(
+        in0: Operand,
+        builder: TTIRBuilder,
+        unit_attrs: List[str] = None,
+    ):
+        # derive sharded shapes
+        assert (
+            (shape[0] % end_grid[0] == 0) and (shape[1] % end_grid[1] == 0),
+            "shard_shape must be divisible by end_grid",
+        )
```
ctodTT, line 255: I would reimplement this as a `pytest.skip` so the test doesn't fail, it just isn't run in nonsensical cases

ctodTT (approved):
LGTM save for a small nit

jojo-771771:

jojo-771771:

jojo-771771:

PR #5084, opened 2025-09-24
[wip] Create bypass path for ScatterOp in UpdateGlobalToLocalShapesPass when applied to sharded static cache update

jameszianxuTT:
[wip] Mechanism is to roughly pattern match a scatter op that would eventually be fused in TTIRFusing to a fill_cache/update_cache op, assert some safety-invariants (tldr: don't shard inputs along the scatter axis) and punch a hole through the assert to unblock TP llama bringup.

### DIFF
--- a/lib/Dialect/StableHLO/Transforms/UpdateGlobalToLocalShapes.cpp
+++ b/lib/Dialect/StableHLO/Transforms/UpdateGlobalToLocalShapes.cpp
@@ -27,7 +27,6 @@ static bool isSafeCacheUpdate(mlir::stablehlo::ScatterOp scatterOp,
                               mlir::sdy::MeshOp &globalMeshOp) {
   mlir::stablehlo::ScatterDimensionNumbersAttr scatterDimensionNumbers =
       scatterOp.getScatterDimensionNumbers();
-  auto updateWindowDims = scatterDimensionNumbers.getUpdateWindowDims();
   auto insertedWindowDims = scatterDimensionNumbers.getInsertedWindowDims();
   auto scatterDimsToOperandDims =
       scatterDimensionNumbers.getScatterDimsToOperandDims();
@@ -47,30 +46,6 @@ static bool isSafeCacheUpdate(mlir::stablehlo::ScatterOp scatterOp,
   // Early exit conditions - return false if we can't analyze safely
   if (inputDimShardings.empty() || updateDimShardings.empty()) {
     return false;
-  }
-
-  // Debug output for analysis
-  llvm::errs() << "ScatterOp debug info:\n";
-  llvm::errs() << "  UpdateWindowDims: ";
-  llvm::interleaveComma(updateWindowDims, llvm::errs());
-  llvm::errs() << "\n";
-  llvm::errs() << "  InsertedWindowDims: ";
-  llvm::interleaveComma(insertedWindowDims, llvm::errs());
-  llvm::errs() << "\n";
-  llvm::errs() << "  ScatterDimsToOperandDims: ";
-  llvm::interleaveComma(scatterDimsToOperandDims, llvm::errs());
-  llvm::errs() << "\n";
-
-  llvm::errs() << "  Input sharding dims: ";
-  int ctr = 0;
-  for (mlir::sdy::DimensionShardingAttr sharding : inputDimShardings) {
-    ctr++;
-    llvm::errs() << "inputDimShardings #" << ctr << "\n";
-    llvm::ArrayRef<mlir::sdy::AxisRefAttr> sharding_axes = sharding.getAxes();
-    for (mlir::sdy::AxisRefAttr axisref : sharding_axes) {
-      llvm::errs() << axisref.toString() << ",";
-    }
-    llvm::errs() << "\n";
   }
 
   auto scatterInputs = scatterOp.getInputs();
@@ -100,7 +75,6 @@ static bool isSafeCacheUpdate(mlir::stablehlo::ScatterOp scatterOp,
   }
 
   auto scatterAxis = insertedWindowDims.front();
-  llvm::errs() << "Scatter axis is " << scatterAxis << "\n";
 
   // figure out sharding axis
 
@@ -119,8 +93,6 @@ static bool isSafeCacheUpdate(mlir::stablehlo::ScatterOp scatterOp,
 
   // We expect the sharding axes and scatter axes to be disjoint
   if (llvm::is_contained(shardingAxes, scatterAxis)) {
-    llvm::errs() << "Scatter axis " << scatterAxis
-                 << " conflicts with sharding axes - unsafe\n";
     return false; // Not safe - scatter axis is sharded
   }
 
--- a/lib/Dialect/StableHLO/Transforms/UpdateGlobalToLocalShapes.cpp
+++ b/lib/Dialect/StableHLO/Transforms/UpdateGlobalToLocalShapes.cpp
@@ -16,6 +16,119 @@ namespace mlir::tt::stablehlo {
 #define GEN_PASS_DEF_UPDATEGLOBALTOLOCALSHAPESPASS
 #include "ttmlir/Dialect/StableHLO/Transforms/Passes.h.inc"
 
+// Helper function to determine if a ScatterOp represents a safe cache update
+// that doesn't require the full scatter operation error handling.
+// This requires the scatter to have inputs that are either unsharded,
+//  or jointly sharded along the same axis, which must differ from the
+//  scatter axis
+// i.e. both insertedWindowDims and scatterDimsToOperandDims must be orthogonal
+// to the update/input sharding dims
+static bool isSafeCacheUpdate(mlir::stablehlo::ScatterOp scatterOp,
+                              mlir::sdy::MeshOp &globalMeshOp) {
+  mlir::stablehlo::ScatterDimensionNumbersAttr scatterDimensionNumbers =
+      scatterOp.getScatterDimensionNumbers();
+  auto updateWindowDims = scatterDimensionNumbers.getUpdateWindowDims();
+  auto insertedWindowDims = scatterDimensionNumbers.getInsertedWindowDims();
+  auto scatterDimsToOperandDims =
+      scatterDimensionNumbers.getScatterDimsToOperandDims();
+
+  // Get sharding info for inputs
+  llvm::ArrayRef<mlir::sdy::DimensionShardingAttr> inputDimShardings =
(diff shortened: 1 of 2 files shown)


llama_singlelayer.shlo
```diff
@@ -0,0 +1,239 @@
+// -----// IR Dump Before ApplyArgumentShardStatusPass (apply-argument-shard-status) ('builtin.module' operation: @SyncTensorsGraph.450) //----- //
```
jameszianxuTT: I am working on isolating a smaller unit test to be checked into the appropriate location

lib/Dialect/StableHLO/Transforms/UpdateGlobalToLocalShapes.cpp
```diff
@@ -16,6 +16,89 @@ namespace mlir::tt::stablehlo {
...
+      shardingAxes.push_back(dimIndex);
+    }
+  }
+
+  // We expect the sharding axes and scatter axes to be disjoint
+  if (llvm::is_contained(shardingAxes, scatterAxis)) {
+    return false; // Not safe - scatter axis is sharded
+  }
+
+  llvm::errs() << "Scatter axis " << scatterAxis
```
hkwonTT: You can use `llvm::dbgs()` if this message is meant to be a debug message.

jameszianxuTT: Sure, I'll add it / change to llvm::dbgs(). Thanks for the review!

lib/Dialect/StableHLO/Transforms/UpdateGlobalToLocalShapes.cpp
```diff
@@ -2,20 +2,104 @@
...
+      shardingAxes.push_back(dimIndex);
+    }
+  }
+
+  // We expect the sharding axes and scatter axes to be disjoint
+  if (llvm::is_contained(shardingAxes, scatterAxis)) {
+    return false; // Not safe - scatter axis is sharded
+  }
+
+  llvm::dbgs() << "Scatter axis " << scatterAxis
```
mmanzoorTT: Please remove the debug message before merging.

lib/Dialect/StableHLO/Transforms/UpdateGlobalToLocalShapes.cpp
```diff
@@ -2,20 +2,104 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 
+#include "mlir/Transforms/GreedyPatternRewriteDriver.h"
```
mmanzoorTT: Move it back to original location.

lib/Dialect/StableHLO/Transforms/UpdateGlobalToLocalShapes.cpp
```diff
@@ -2,20 +2,104 @@
...
 #include "ttmlir/Dialect/StableHLO/Transforms/Passes.h"
 #include "ttmlir/Dialect/StableHLO/Transforms/ShardyCCLToStableHLOCCL.h"
 #include "ttmlir/Dialect/StableHLO/Utils/GSPMDUtils.h"
 #include "ttmlir/Dialect/StableHLO/Utils/ShardyUtils.h"
 #include "ttmlir/Dialect/TTIR/IR/TTIR.h"
-
-#include "mlir/Transforms/GreedyPatternRewriteDriver.h"
+#include "llvm/Support/Debug.h"
```
mmanzoorTT: Remove it.

lib/Dialect/StableHLO/Transforms/UpdateGlobalToLocalShapes.cpp
```diff
@@ -2,20 +2,104 @@
...
+  auto scatterUpdates = scatterOp.getUpdates();
+
+  // shlo scatter accepts a variadic number of inputs / updates tensors, we
+  // expect only 1
+  if (scatterInputs.size() != 1 || scatterUpdates.size() != 1) {
+    return false;
+  }
+
+  // expect the inputs and updates to have the same size
+  mlir::RankedTensorType scatterInput =
```
mmanzoorTT: Please rename
```suggestion
  mlir::RankedTensorType scatterInputType =
```

lib/Dialect/StableHLO/Transforms/UpdateGlobalToLocalShapes.cpp
```diff
@@ -2,20 +2,104 @@
...
+      mlir::dyn_cast<mlir::RankedTensorType>(scatterInputs.front().getType());
+  auto scatterUpdate =
```
mmanzoorTT: Please rename
```suggestion
  auto scatterUpdateType =
```

lib/Dialect/StableHLO/Transforms/UpdateGlobalToLocalShapes.cpp
jameszianxuTT: I left it here because this is a workaround, and I thought I should warn the user that this workaround is taking place. Is there an alternative pattern to highlight such warnings, or is this generally discouraged as being overly verbose?

mmanzoorTT: You are use `op->emitWarning()` similar to `op->emitError()`.

jameszianxuTT: Thanks, will do. I think it is a useful warning to retain given the workaround nature of this change, and it should not be extensively triggered as it only matches a narrow case of static cache update.

lib/Dialect/StableHLO/Transforms/UpdateGlobalToLocalShapes.cpp
```diff
@@ -16,6 +16,92 @@ namespace mlir::tt::stablehlo {
 #define GEN_PASS_DEF_UPDATEGLOBALTOLOCALSHAPESPASS
 #include "ttmlir/Dialect/StableHLO/Transforms/Passes.h.inc"
 
+// Helper function to determine if a ScatterOp represents a safe sharded scatter
+//  that doesn't require attr rewriting when running
+//  UpdateGlobalToLocalShapesPass This requires the scatter to have inputs that
+//  are either unsharded, or jointly sharded along the same axis, which must
+//  differ from the scatter axis
+// i.e. both insertedWindowDims and scatterDimsToOperandDims must be orthogonal
+//  to the update/input sharding dims
```
anuragsingh-tt: NIT: period after comments for all comments.

lib/Dialect/StableHLO/Transforms/UpdateGlobalToLocalShapes.cpp
```diff
@@ -16,6 +16,92 @@ namespace mlir::tt::stablehlo {
...
+static bool isSafeShardedScatter(mlir::stablehlo::ScatterOp scatterOp,
+                                 mlir::sdy::MeshOp &globalMeshOp) {
+  mlir::stablehlo::ScatterDimensionNumbersAttr scatterDimensionNumbers =
+      scatterOp.getScatterDimensionNumbers();
+  auto insertedWindowDims = scatterDimensionNumbers.getInsertedWindowDims();
+  auto scatterDimsToOperandDims =
```
anuragsingh-tt: Let's not use auto here as its not clear what the data type is.

lib/Dialect/StableHLO/Transforms/UpdateGlobalToLocalShapes.cpp
```diff
@@ -16,6 +16,93 @@ namespace mlir::tt::stablehlo {
...
+
+  // We expect the sharding axes and scatter axes to be disjoint
+  if (llvm::is_contained(shardingAxes, scatterAxis)) {
+    return false; // Not safe - scatter axis is sharded
+  }
+
+  scatterOp.emitWarning("In scatterOp ")
+      << scatterOp->getName() << ", the scatter axis " << scatterAxis
+      << " is orthogonal to input sharding axes - this sharded scatter "
+         "is safe to lower through UpdateGlobalToLocalShapesPass. \n";
```
mmanzoorTT: No need of newline character for these type of messages.
```suggestion
         "is safe to lower through UpdateGlobalToLocalShapesPass.";
```

lib/Dialect/StableHLO/Transforms/UpdateGlobalToLocalShapes.cpp
```diff
@@ -16,6 +16,93 @@ namespace mlir::tt::stablehlo {
...
+      shardingAxes.push_back(dimIndex);
+    }
+  }
+
+  // We expect the sharding axes and scatter axes to be disjoint
+  if (llvm::is_contained(shardingAxes, scatterAxis)) {
+    return false; // Not safe - scatter axis is sharded
+  }
+
+  scatterOp.emitWarning("In scatterOp ")
```
wooseokTT: NIT: just curious if this warning makes sense...  I believe warning messages make sense when the feature is a kind of experimental and may causes some unknown problems, but this message seems to be informing your decision only. In addition, if the message is discarded by many people, it doesn't make sense to keep it.

lib/Dialect/StableHLO/Transforms/UpdateGlobalToLocalShapes.cpp
mmanzoorTT: I am still unsure whether to call it a workaround. It look likes a legality check; if certain conditions are satisfied then we can perform some conversion/transformation safely (and nothing is removed/added/modified). The debug message tells us the same thing. We don't add messages for legality check so I don't see need of this debug message but it is your choice.

lib/Dialect/StableHLO/Transforms/UpdateGlobalToLocalShapes.cpp
mmanzoorTT: I agree. It look likes a legality check; if certain conditions are satisfied then we can perform some conversion/transformation safely (and nothing is removed/added/modified). The debug message tells us the same thing. We don't add messages for legality check and I don't see a need of it.

jameszianxuTT: Okay, these are fair points. I've removed the warning.

anuragsingh-tt (approved):

jojo-771771:

jojo-771771:

jojo-771771:

jojo-771771:

PR #5085, opened 2025-09-24
D2M/TTNN Integration - Always insert streams for `ttnn_metal_layout_cast` ops.

vtangTT:
### Ticket
Link to Github Issue

### Problem description

### What's changed

### Checklist


lib/Dialect/TTIR/Transforms/InsertStreams.cpp
```diff
@@ -0,0 +1,111 @@
...
+
+  void insertStream(PatternRewriter &rewriter, OpOperand &operand,
+                    ttir::GenericOp op) const {
+    auto memref = mlir::cast<MemRefType>(operand.get().getType());
+    auto streamAttr = rewriter.getAttr<ttcore::ViewLayoutAttr>(
+        rewriter.getMultiDimIdentityMap(memref.getRank()));
+    auto streamMemref =
+        MemRefType::get(memref.getShape(), memref.getElementType(), streamAttr,
+                        memref.getMemorySpace());
+    auto storageAttr = ttcore::ShardLayoutAttr::get(memref, /*buffers=*/1);
```
bgrady-tt: I think we need to select the number of buffers here based on `num-stream-buffers` option. This is passed in to Allocate pass at construction time in TTMetalPipelines.cpp

lib/Dialect/TTIR/Transforms/InsertStreams.cpp
```diff
@@ -0,0 +1,111 @@
...
+namespace mlir::tt::ttir {
+
+#define GEN_PASS_DEF_TTIRALWAYSINSERTSTREAMS
+#include "ttmlir/Dialect/TTIR/Transforms/Passes.h.inc"
+
+namespace {
+class TTIRInsertStreams final : public OpRewritePattern<ttir::GenericOp> {
+public:
+  TTIRInsertStreams(MLIRContext *context)
+      : OpRewritePattern<ttir::GenericOp>(context) {}
```
bgrady-tt: Constructor here should take as an argument num-stream-buffers setting, and use it when inferring stream ShardLayouts

lib/Dialect/TTIR/Transforms/InsertStreams.cpp
vtangTT: Yep forgot about the options. Added the num-stream-buffers now, unsure if we also want the `allow-output-spilling` option?

lib/Dialect/TTIR/Transforms/InsertStreams.cpp
```diff
@@ -0,0 +1,105 @@
...
+    if (auto genericOp = mlir::dyn_cast_or_null<ttir::GenericOp>(definingOp)) {
+      // Lack of a compute thread implies this is in special DMA-only form,
+      // which should not have streams inferred.
+      if (not llvm::any_of(genericOp.getThreads(), [](Attribute attr) {
+            return mlir::cast<ttir::ThreadAttr>(attr).getThreadType() ==
+                   ttir::ThreadType::Compute;
+          })) {
+        return false;
+      }
+    }
```
vroubtsovTT: this check is at the level of the parent `GenericOp`, it should be hoisted out of this method and into `matchAndRewrite()` (it seems silly to nest into an operand only to recover the parent op right away as the defining op; also seems silly to use a `dyn_cast_or_null` on something we already know is a `GenericOp`)

since ttir.allocate also uses the same check, a related thought here is to factor this out into a utility method on GenericOp itself, smth like `isDMAOnly()`.

lib/Dialect/TTIR/Transforms/InsertStreams.cpp
vroubtsovTT: output spilling is a mode modifier for a pass that does L1/DRAM spilling, it isn't needed for this pass since it won't do any memory allocation

lib/Dialect/TTIR/Transforms/InsertStreams.cpp
vtangTT: I've moved this to the `isDMAOnlyForm()` utility under the `GenericOp` itself. Switched it in both this pass and the Allocate pass. 

ptal 🙏

lib/Dialect/D2M/Transforms/InsertStreams.cpp
```diff
@@ -0,0 +1,85 @@
...
+      if (opIsStream || op.isDMAOnlyForm()) {
+        continue;
+      }
+      insertStream(rewriter, operand, op);
+      modified = true;
+    }
+
+    return success(modified);
+  }
+  void insertStream(PatternRewriter &rewriter, OpOperand &operand,
```
nsmithtt, line 48: Could be nice to share this code with the allocator in a common utility header since it's identical.  Won't block the PR on it though.

nsmithtt (approved):

lib/Dialect/D2M/Transforms/InsertStreams.cpp
```diff
@@ -0,0 +1,84 @@
...
```
azecevic000, line 48: Nit: add a new line above this one

lib/Dialect/D2M/Transforms/InsertStreams.cpp
```diff
@@ -0,0 +1,84 @@
...
+    auto storageMemref =
+        MemRefType::get(memref.getShape(), memref.getElementType(), storageAttr,
+                        memref.getMemorySpace());
+    auto storage = rewriter.create<memref::AllocOp>(op.getLoc(), storageMemref);
+    auto streamLayout = rewriter.create<d2m::StreamLayoutOp>(
+        op.getLoc(), streamMemref, operand.get(), storage);
+    rewriter.modifyOpInPlace(
+        op, [&]() { operand.assign(streamLayout.getResult()); });
+  }
+  unsigned numStreamBuffers;
```
azecevic000, line 69: Nit: Should this variable be private? (also new line)

azecevicTT (approved):

lib/Dialect/D2M/Transforms/InsertStreams.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+  LogicalResult matchAndRewrite(d2m::GenericOp op,
+                                PatternRewriter &rewriter) const final {
+    bool modified = false;
+    for (OpOperand &operand : op->getOpOperands()) {
+      // If input is not already a stream, insert one.
+      // For DMA-only form, stream insertion will break semantics.
+      bool opIsStream = mlir::isa_and_nonnull<d2m::StreamLayoutOp>(
+          operand.get().getDefiningOp());
+      if (opIsStream || op.isDMAOnlyForm()) {
```
vroubtsovTT: It is (again) confusing to see the op-level check `op.isDMAOnlyForm()` done inside a loop over operands. `op.isDMAOnlyForm()` condition is a loop invariant. 

But more imporantly, generic ops in DMA-only form are simply ignored by this pass, period. I think that would be more clearly communicated with something like
```
 // For DMA-only form, stream insertion will break semantics.
if (op.isDMAOnlyForm()) {
  return failure();
}
``` 
at the very beginning of matchAndRewrite(). This DMA-only check will probably go away (there is discussion to use streams for DMA ops anyway, apparently) and it would be easy to remove this check here when that happens.

lib/Dialect/D2M/Transforms/InsertStreams.cpp
vtangTT: fixed ty

lib/Dialect/D2M/Transforms/InsertStreams.cpp
vtangTT: fixed ty

lib/Dialect/D2M/Transforms/InsertStreams.cpp
vtangTT: https://github.com/tenstorrent/tt-mlir/issues/5106

jojo-771771:

jojo-771771:

jojo-771771:

jojo-771771:

PR #5097, opened 2025-09-25
adding more info on print_ir flag

anuragsingh-tt:
### Problem description
While debugging a failing op in builder I had set the print_ir parameter to True but was not seeing the IR dump. This is because MLIR's IR printing happens at the C++ level and gets lost when the process crashes. So the only valid path to get IR is to provide `print_ir="/path/to/dir"` which prints IR to files. 

From MLIR docs:

> - IR printing is unsafe, e.g. when using `PassManager` with `PassManager.enable_ir_printing()` which calls thread-unsafe `llvm::raw_ostream`.

Sending the IR to a directory was not actually useful for my needs for debugging as a failing assert is not visible in that directory. Printing the IR to stdout was helpful but not always reproducible. I have added a note to this parameter to indicate its limited use for debugging IR.

### DIFF
--- a/tools/builder/base/builder_utils.py
+++ b/tools/builder/base/builder_utils.py
@@ -374,10 +374,26 @@ def compile_ttir_to_flatbuffer(
     pipeline_options : *Optional[List[str]]*
         Pipeline options to be added to the pass
 
-    print_ir : *Union[bool, str]*, optional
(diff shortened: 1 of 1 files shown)


jojo-771771:

jojo-771771:

jojo-771771:

jojo-771771:

PR #5103, opened 2025-09-26
Update `exp_tile` op to include fast math options

navdeepkk-polymagelabs:
`exp_tile` op was missing `approx`, `fast_approx`, `scale_enable`, and `skip_positive_check` flags. These are  exposed in the tt-metal API. Expose them in `tt-mlir` `exp_tile` op as well.

### Problem description
exp_tile` op was missing `approx`, `fast_approx`, `scale_enable`, and `skip_positive_check` flags.

### What's changed
These are  exposed in the tt-metal API. Expose them in `tt-mlir` `exp_tile` op as well.


wenbinlyuTT:
Thanks for doing support of this! A few suggestions:
* `exp_tile_init` seems to require a similar but different set of template args, I assume they need to be 'in sync' with the ones used on `exp_tile` so it's better to support that as well.
* Since the template args need to be the synchronized for both calls, maybe it's better to model these in the `tile_exp` op, so during lowering we can guarantee we use the same args for them.

nsmithtt (approved):
Thank you!

Oh just saw @wenbinlyuTT's feedback, let's address those points.

navdeepkk-polymagelabs:
Thanks for the feedback:
> * `exp_tile_init` seems to require a similar but different set of template args, I assume they need to be 'in sync' with the ones used on `exp_tile` so it's better to support that as well.

I see [here](https://github.com/tenstorrent/tt-metal/blob/5be2f8ac068a51609e3483edcdd35ec2f7341275/tt_metal/include/compute_kernel_api/eltwise_unary/exp.h#L22) that `exp_tile_init` has only three args. I can update the lowering for it to account for these as well.

> * Since the template args need to be the synchronized for both calls, maybe it's better to model these in the `tile_exp` op, so during lowering we can guarantee we use the same args for them.

I am not aware of the `tile_exp` op. Did you mean `exp_tile` op?

wenbinlyuTT:
@navdeepkk-polymagelabs 
1. Yes, they have similar but different set of template args, some match but others don't.
2. [`tile_exp`](https://github.com/tenstorrent/tt-mlir/blob/ddc871aec156fad024bfac04f0654075511c4da4/include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td#L153) is the higher level op in the D2M dialect, it eventually gets lowered to the TTKernel ops like `exp_tile` and `exp_tile_init`.

navdeepkk-polymagelabs:
> @navdeepkk-polymagelabs
> 
> 1. Yes, they have similar but different set of template args, some match but others don't.
> 2. [`tile_exp`](https://github.com/tenstorrent/tt-mlir/blob/ddc871aec156fad024bfac04f0654075511c4da4/include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td#L153) is the higher level op in the D2M dialect, it eventually gets lowered to the TTKernel ops like `exp_tile` and `exp_tile_init`.

Got it, thanks. Should I expose the same set of options on `tile_exp` op as well, namely `approx`, `fast_and_approx`, `enable_scale` and `scal_value`?

wenbinlyuTT:
@navdeepkk-polymagelabs Ideally on `tile_exp` it should be the union of all the template arguments supported by both `exp_tile` and `exp_tile_init`. Right now the set is:
```
bool approx
bool fast_and_approx
bool scale_en
bool skip_positive_check
int iterations
```
Plus the `scale`, which is both a template parameter and an input argument.
[Reference](https://github.com/tenstorrent/tt-metal/blob/6c306a589056377b826dc4820e45cf8ae8b3d452/tt_metal/include/compute_kernel_api/eltwise_unary/exp.h#L56)

As a first step though, unless required by some application, we can just support `approx` and `fast_and_approx` on `tile_exp`, so the PR is a bit simpler.

PR #5111, opened 2025-09-26
Ops with # threads <= # DMA HW Threads don't need to be merged

nsmithtt:
Previously we were making the assumption that there will always be a 1-1 relationship between the number of operands and the number of datamovement regions.  In the context of manually written generic ops, this is not necessarily true, and the datamovement regions are already manually fused.

Instead we should count the number of datamovement threads which is more accurate.


nsmithtt:
> Although I am unsure it would make a difference if num_threads was < 2 ?

By num_threads you mean on the incoming generic?  I think it'd still be an issue because the number of operands doesn't change, ie. you could have 3 operands and a manually fused single thread and I think the old rewriter would have still tried to do something.

PR #5113, opened 2025-09-26
Custom load/store kernel dir

nsmithtt:
You can now specify somewhere to dump kernels to other than /tmp.

```bash
ttrt run --kernel-source-dir /my/path ...
```

You can also now use the location name (if it exists) as the kernel name.  This makes it possible to tag a particular kernel function with a location name and have the dumped artifact match exactly.


vroubtsovTT (approved):
For the record, I dislike working with file paths without going through a proper lib like `std::filesystem` for the simple reason that I like canonicalizing output paths into their absolute form.

jnie-TT (approved):

PR #5114, opened 2025-09-26
[D2M] ttnn-jit python wheel - first draft

sgholamiTT:
### Ticket
Link to Github Issue

### Problem description

### What's changed

### Checklist

### DIFF
--- a/.github/actions/build-tt-mlir-action/action.yaml
+++ b/.github/actions/build-tt-mlir-action/action.yaml
@@ -59,7 +59,7 @@ runs:
           -DTTMLIR_ENABLE_RUNTIME=ON \
           -DTTMLIR_ENABLE_RUNTIME_TESTS=ON \
           -DTT_RUNTIME_ENABLE_PERF_TRACE=${{ inputs.enable-perf }} \
-          -DTTMLIR_ENABLE_PYKERNEL=${{ inputs.enable-pykernel }} \
+          -DTTMLIR_ENABLE_PYKERNEL=ON \
           -DTTMLIR_ENABLE_STABLEHLO=ON \
           -DTTMLIR_ENABLE_OPMODEL=${{ inputs.enable-op-model }} \
           -DTT_RUNTIME_DEBUG=${{ inputs.enable-runtime-debug }} \
@@ -79,6 +79,17 @@ runs:
         source env/activate
         cmake --build ${{ inputs.build-output-dir }} -- ttrt
         cmake --build ${{ inputs.build-output-dir }} -- compile-ttmlir-tests
+
+    - name: Build ttnn-jit wheel
+      shell: bash
+      run: |
+        source env/activate
+        export CURRENT_DIR=$(pwd)
+        cd ${{ inputs.work-dir }}/tools/ttnn-jit
+        mkdir -p dist
+        pip install build setuptools wheel
+        python3 -m build --wheel
+        cd $CURRENT_DIR
 
     - name: Build tt-alchemist
       shell: bash
@@ -134,6 +145,13 @@ runs:
       with:
         name: ttrt-whl-${{ inputs.build-name }}
         path: build/tools/ttrt/build/ttrt*.whl
+
+    - name: Upload ttnn-jit .whl
+      uses: actions/upload-artifact@v4
+      with:
+        name: ttnn-jit-whl-${{ inputs.build-name }}
+        path: ${{ inputs.work-dir }}/tools/ttnn-jit/dist/ttnn_jit*.whl
+        if-no-files-found: error
 
     - name: Upload tt-alchemist .whl
       uses: actions/upload-artifact@v4
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -285,6 +285,18 @@ jobs:
           pip install pytest
         fi
 
+    - name: Download ttnn-jit whl
+      uses: actions/download-artifact@v4
+      with:
+        name: ttnn-jit-whl-${{ matrix.build.image }}
+
+    - name: Install ttnn-jit whl
+      shell: bash
+      run: |
+        source env/activate
+        pip install ttnn_jit-*.whl --force-reinstall
+        python -c "import ttnn_jit; print('Successfully imported ttnn_jit:', ttnn_jit.__file__)"
+
     - name: Download Build Artifacts
       uses: ./.github/actions/download-artifact
       with:
--- a/tools/ttnn-jit/setup.py
+++ b/tools/ttnn-jit/setup.py
@@ -1,0 +1,321 @@
+# SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+
+import os
+import pathlib
+import shutil
+import subprocess
+import re
+from setuptools import setup
+from datetime import datetime
+
+
+def extract_tt_metal_version() -> str | None:
+    """Extract TT_METAL_VERSION from third_party/CMakeLists.txt"""
+    try:
+        project_root = pathlib.Path(__file__).resolve().parent.parent.parent
+        cmake_file = project_root / "third_party" / "CMakeLists.txt"
+        with open(cmake_file, "r") as f:
+            content = f.read()
+            match = re.search(r'set\(TT_METAL_VERSION\s+"([^"]+)"\)', content)
+            if match:
+                return match.group(1)
+    except Exception:
+        pass
+    return None
+
+
+# Environment and build configuration (similar to ttrt setup.py)
+TTMLIR_VERSION_MAJOR = os.getenv("TTMLIR_VERSION_MAJOR", "0")
+TTMLIR_VERSION_MINOR = os.getenv("TTMLIR_VERSION_MINOR", "0")
+TTMLIR_VERSION_PATCH = os.getenv("TTMLIR_VERSION_PATCH", "0")
+
+src_dir = os.environ.get(
+    "SOURCE_ROOT",
+    os.path.join(os.path.dirname(os.path.abspath(__file__)), "..", ".."),
+)
+# Use 'src_dir/build' as default location if TTMLIR_BINARY_DIR env variable is not available.
+ttmlir_build_dir = os.environ.get(
+    "TTMLIR_BINARY_DIR",
+    os.path.join(src_dir, "build"),
+)
+
+metaldir = f"{src_dir}/third_party/tt-metal/src/tt-metal/build"
+ttmetalhome = os.environ.get("TT_METAL_HOME", "")
+
+# Set RPATH for runtime linking (similar to ttrt)
+os.environ["LDFLAGS"] = "-Wl,-rpath,'$ORIGIN'"
+
+# Feature flags - assume these are enabled for ttnn-jit
+enable_runtime = True  # Always needed for ttnn-jit
+enable_ttnn = True  # Always needed for ttnn-jit
+enable_ttmetal = True  # Always needed for ttnn-jit
+enable_ttnn_jit = True  # Always enabled for this package
+enable_ttrt = True  # Always needed for ttnn-jit
+arch = os.environ.get("CMAKE_SYSTEM_PROCESSOR", "x86_64")
+
+# Runtime libraries that ttnn-jit needs
+runtime_module = f"_ttmlir_runtime.cpython-311-{arch}-linux-gnu.so"
+# ttnn-jit specific module
+ttnn_jit_module = f"_ttnn_jit.cpython-311-{arch}-linux-gnu.so"
+dylibs = []
+runlibs = []
+metallibs = []
+
+# Core ttnn-jit libraries
+dylibs += ["libTTMLIRRuntime.so", runtime_module]
+
+# TTMLIR libraries (required for ttmlir package)
+ttmlir_libs = [
+    "_mlir.cpython-311-x86_64-linux-gnu.so",
+    "_mlirDialectsLinalg.cpython-311-x86_64-linux-gnu.so",
+    "_mlirDialectsQuant.cpython-311-x86_64-linux-gnu.so",
+    "_mlirLinalgPasses.cpython-311-x86_64-linux-gnu.so",
+    "_ttmlir.cpython-311-x86_64-linux-gnu.so",
+    "libTTMLIRPythonCAPI.so",
+]
+
+# TTNN libraries (required for ttnn-jit)
+if enable_ttnn:
+    runlibs += ["_ttnncpp.so"]
+
+# TT-Metal libraries (required for ttnn-jit)
+if enable_ttmetal:
+    runlibs += ["libtt_metal.so"]
+
+# Common libraries for TTNN/TT-Metal
+if enable_ttnn or enable_ttmetal:
+    runlibs += ["libdevice.so"]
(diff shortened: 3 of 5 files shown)


jmcgrathTT (approved):

.github/actions/build-tt-mlir-action/action.yaml
```diff
@@ -80,6 +80,17 @@ runs:
         cmake --build ${{ inputs.build-output-dir }} -- ttrt
         cmake --build ${{ inputs.build-output-dir }} -- compile-ttmlir-tests
 
+    - name: Build ttnn-jit wheel
+      shell: bash
```
nsumrakTT, line 85: If inputs.enable-pykernel != ON
should build ttnn-jit wheel?

.github/workflows/build-and-test.yml
```diff
@@ -285,6 +285,18 @@ jobs:
...
+      with:
+        name: ttnn-jit-whl-${{ matrix.build.image }}
+
+    - name: Install ttnn-jit whl
+      shell: bash
+      run: |
+        source env/activate
+        pip install ttnn_jit-*.whl --force-reinstall
+        python -c "import ttnn_jit; print('Successfully imported ttnn_jit:', ttnn_jit.__file__)"
+
```
nsumrakTT, line 298: why installing the wheel when tests work without it?
If this is just a test of wheel it should be enclosed in separate test and not run on every test machine (currently 26 times).
If it is required for test, please ensure it is downloaded/installed only when it is needed.

tools/ttnn-jit/setup.py
```diff
@@ -0,0 +1,337 @@
...
+    """Get dependencies needed for TTNN JIT functionality"""
+    return [
+        "nanobind",  # Python binding framework
+        "numpy",  # Core numeric computing
+        "torch",  # PyTorch for tensor operations
+        "pyyaml",  # YAML parsing
+        "loguru",  # Logging (used by TTNN)
+        "tqdm",  # Progress bars
+        "psutil",  # System info
+        "graphviz",  # Visualization for JIT graphs
```
vtangTT: idt graphviz is for ttnn_jit, this is a ttnn dep no? same with most of these deps, except for nanobind and torch?

tools/ttnn-jit/setup.py
```diff
@@ -0,0 +1,337 @@
...
+    }
+
+    # Add ttmlir dialect package data
+    for sub in [
+        sub for sub in TTMLIR_SUBPACKAGES if sub != "_mlir_libs" and sub != "extras"
+    ]:
+        package_data[f"ttmlir.{sub}"] = ["*.py"]
+
+    # Add ttnn package data with common file patterns
+    common_ttnn_patterns = ["*.py", "*.so", "*.md", "*.ipynb", "*.txt", "*.sh", "*.svg"]
```
vtangTT: why do we need ipynb, md and svg files?

tools/ttnn-jit/setup.py
```diff
@@ -0,0 +1,337 @@
...
+        packages=packages,
+        package_dir=package_dir,
+        entry_points={
+            "console_scripts": ["ttnn-jit = ttnn_jit:main"],
+        },
+        package_data=package_data,
+        zip_safe=False,
+        long_description=get_readme(),
+        long_description_content_type="text/markdown",
+        python_requires=">=3.7",
```
vtangTT: this should prob be whatever version ttnn uses or ttmlir uses, whichever one is the lowest no? I think we are def at least on 3.10?

tools/ttnn-jit/setup.py
```diff
@@ -0,0 +1,337 @@
...
+    all_runtime_libs = setup_runtime_libraries(config)
+
+    # Generate package configuration
+    packages, package_dir = generate_package_configuration(config)
+
+    # Generate package data
+    package_data = generate_package_data(all_runtime_libs)
+
+    # Call setup with clean configuration
+    setup(
```
vtangTT, line 306: is it possible to have the wheel detect a pre-existing build and only build when it doesn't exist. I think we do smthn like this in metal if you look at their setup.py. They have something like this:

`cmdclass=dict(build_ext=CMakeBuild, editable_wheel=EditableWheel),`

tools/ttnn-jit/setup.py
```diff
@@ -0,0 +1,337 @@
...
+    "dialects",
+    "dialects.linalg",
+    "dialects.linalg.opdsl",
+    "dialects.linalg.passes",
+    "extras",
+    "_mlir_libs",
+]
+
+# Library definitions
+CORE_TTNN_LIBS = ["_ttnncpp.so"]
```
vtangTT, line 36: For all the metal libs we copy, thoughts on using their install targets? 

```
cmake --install third_party/tt-metal/src/tt-metal/build --prefix install_dir --component metalium-runtime
cmake --install third_party/tt-metal/src/tt-metal/build --prefix install_dir --component metalium-dev
cmake --install third_party/tt-metal/src/tt-metal/build --prefix install_dir --component ttnn-runtime
cmake --install third_party/tt-metal/src/tt-metal/build --prefix install_dir --component ttnn-dev```

Comment (vtangTT) [reply tools/ttnn-jit/setup.py:36]: also thoughts on creating install targets for any of mlir runtime libs we need for this wheel insteda of using `shutil.copy`

File: .github/workflows/build-and-test.yml [hunk 8]
```diff
@@ -285,6 +285,18 @@ jobs:
...
+    - name: Download ttnn-jit whl
+      uses: actions/download-artifact@v4
+      with:
+        name: ttnn-jit-whl-${{ matrix.build.image }}
+
+    - name: Install ttnn-jit whl
+      shell: bash
+      run: |
+        source env/activate
+        pip install ttnn_jit-*.whl --force-reinstall
```
vtangTT, line 295: Can we also test `import ttnn`. This wheel is installed in a new env right?

tools/ttnn-jit/_src/dispatch_op.py
```diff
@@ -5,35 +5,45 @@
 import ttnn
 import torch
 
-from _ttmlir_runtime import runtime, binary, utils
+from ttnn_jit.runtime._ttmlir_runtime.runtime import (
```
vtangTT, line 8: Curious if there's a functional reason to change these imports? Is there a problem with doing `utils.create_runtime_device_from_ttnn`

tools/ttnn-jit/setup.py
sgholamiTT: Correct. Removed all except `nanobind` and `torch`.

tools/ttnn-jit/setup.py
sgholamiTT: We don't. It's removed.

tools/ttnn-jit/setup.py
sgholamiTT: Yes, changed it to `python_requires=">=3.11"`.

tools/ttnn-jit/setup.py
sgholamiTT: The line that you've pointed out in your review (I think) has nothing to do with detecting pre-existing wheels.
TBH, I fail to realize the significance/customer value of this mechanism. I know that it might save time for building the wheel for the same git hash of `tt-mlir` + `tt-metal`, but that should be rare, no ? Because even if I change one line in each repo, there's no guarantee that the wheel behaves the same anymore, i.e. it should be built again from scratch. I think it might introduce bugs to the wheel building process. 

Also, I couldn't find a similar mechanism for `ttrt` that also uses copy/pasting of files.

Please LMK what you think.

tools/ttnn-jit/setup.py
sgholamiTT: Following up the discussion on Tuesday's weekly sync, I believe we decided to put this on hold for now.
Please confirm.

.github/workflows/build-and-test.yml
sgholamiTT: Added.
Yes it's tested in a fresh env.

tools/ttnn-jit/_src/dispatch_op.py
sgholamiTT: The only functional reason is that this file should work with the wheel as well. That's why I changed it like this.
But regardless, this change follows Python best practices for explicit imports, so I think it's fine.
LMKWYT.

.github/workflows/build-and-test.yml
sgholamiTT: Can you take a look at [the most recent change](https://github.com/tenstorrent/tt-mlir/pull/5114/commits/b770b69b57cf87e1ef582f996e5f6936641c8b39) list and let me know if that's what you were expecting ?

tools/ttnn-jit/setup.py
```diff
@@ -0,0 +1,339 @@
...
+# Package structure definitions
+TTNN_SUBPACKAGES = [
+    "ttnn",
+    "ttnn.operations",
+    "ttnn.distributed",
+    "ttnn.examples",
+    "ttnn.examples.bert",
+    "ttnn.examples.usage",
+    "ttnn.experimental_loader",
+]
```
nsmithtt, line 24: Do we really need to package the TTNN and TTLIB packages inside our wheel?  What happens if we don't include these.

.github/actions/build-tt-mlir-action/action.yaml
```diff
@@ -59,7 +59,7 @@ runs:
           -DTTMLIR_ENABLE_RUNTIME=ON \
           -DTTMLIR_ENABLE_RUNTIME_TESTS=ON \
           -DTT_RUNTIME_ENABLE_PERF_TRACE=${{ inputs.enable-perf }} \
-          -DTTMLIR_ENABLE_PYKERNEL=${{ inputs.enable-pykernel }} \
+          -DTTMLIR_ENABLE_PYKERNEL=ON \
```
nsumrakTT: This means wheel will be created twice for each image.
Do we need two versions of wheels?

.github/workflows/build-and-test.yml
```diff
@@ -83,11 +83,7 @@ jobs:
           echo "docker-image-harbor=harbor.ci.tenstorrent.net/$DOCKER_CI_IMAGE" >> "$GITHUB_OUTPUT"
           echo "docker-image=$DOCKER_CI_IMAGE" >> "$GITHUB_OUTPUT"
           echo "docker-tag=$DOCKER_TAG" >> "$GITHUB_OUTPUT"
-          if [[ "${{ inputs.builder_build }}" == "true" ]]; then
-            echo "runner=builder" >> "$GITHUB_OUTPUT"
-          else
-            echo "runner=ubuntu-latest" >> "$GITHUB_OUTPUT"
-          fi
+          echo "runner=builder" >> "$GITHUB_OUTPUT"
```
nsumrakTT, line 86: Why this change? How it is related to the wheel?
Please revert.

.github/workflows/build-and-test.yml
```diff
@@ -285,6 +281,21 @@ jobs:
...
+        name: ttnn-jit-whl-${{ matrix.build.image }}
+
+    - name: Install ttnn-jit whl
+      if: matrix.build.type == 'pykernel'
+      shell: bash
+      run: |
+        source env/activate
+        pip install ttnn_jit-*.whl --force-reinstall
+        python -c "import ttnn_jit; print('Successfully imported ttnn_jit:', ttnn_jit.__file__)"
+        python -c "import ttnn; print('Successfully imported ttnn:', ttnn.__file__)"
```
nsumrakTT, line 297: Does this require HW or it can be imported on compute machine?

.github/actions/build-tt-mlir-action/action.yaml
sgholamiTT: Reverted this change. Thanks for pointing this out.

.github/workflows/build-and-test.yml
sgholamiTT: This was suggested by @jmcgrathTT as a workaround for the memory issue we were facing in CI.

.github/workflows/build-and-test.yml
sgholamiTT: Does not require hw, can just be compute.

tools/ttnn-jit/setup.py
sgholamiTT: Thanks for the suggestion. Removed `TTLIB` packages completely. Also removed most of `ttnn` packages. the wheel size is now really small (around 60MBs, from the original 170MB).
For the remaining `ttnn` packages, we need `ttnn` device/tensor creation `ttnn.from_torch`, `device`, etc. So I think we need to keep them.

.github/workflows/build-and-test.yml
jmcgrathTT: Hey @nsumrakTT 

For more context, we didn't end up using the `maximize-space` action here because we are invoking the container above, making it difficult to clean up the underlying host. 

https://github.com/tenstorrent/tt-xla/blob/refs/heads/lpanos/vllm_plugin_tt/.github/workflows/call-build.yml#L100

It can create weird issues like this.
https://github.com/tenstorrent/tt-mlir/actions/runs/18139340890/job/51626231675#step:3:292

If there is a better way to handle this, let me know us. I can handle this in this or a follow-up PR.

tools/ttnn-jit/setup.py
nsmithtt: I suppose I was thinking that ttnn jit is meant to be included from ttnn itself, it feels confusing to package ttnn inside of jit.

.github/workflows/build-and-test.yml
nsumrakTT: We have very limited number of builder machines and these are used for uplitfs.
This will increase PR time drastically. Why not shared builder?

.github/workflows/build-and-test.yml
nsumrakTT: Why running on n150 then?
You could add job for ubuntu-latest like other wheels.

tools/ttnn-jit/setup.py
nsmithtt: @sgholamiTT, we should try to see if it works to _not_ package ttnn to see what happens.

sgholamiTT: @nsmithtt Sure. I will try that. 
However, the point of packaging 'ttnn' inside the `ttnn-jit` wheel is to build a standalone python package, where developers can just `pip install` it and start using it without worrying about the backend explicitly, i.e. they might not have our `ttnn` backend and the wheel should still work.

nsumrakTT: @sgholamiTT It should work when pip install is used, except is doesn't (without our docker image):
https://github.com/tenstorrent/tt-mlir/issues/5232

sgholamiTT: @nsumrakTT I will take a look. Can you add reproduce steps ? I've tried it in a fresh `venv` and it worked.

nsumrakTT: You have to try without our CI image on base docker with freshly created venv (not one with toolchain). It doesn't work.
Check PR linked to the issue, I've moved it to ubuntu-latest and tried to install and run basic import test.

.github/workflows/build-and-test.yml
jmcgrathTT: Shared builders sound good to me! I added the change to https://github.com/tenstorrent/tt-mlir/pull/5241/files

PR #5115, opened 2025-09-26
TTKernel to EmitC ArithFloorDivRewriter

nsmithtt:
Arith FloorDivSIOp doesn't have an emitc lowering, probably because of the spec which says:

>  Signed integer division. Rounds towards negative infinity, i.e. 5 / -2 = -3

However we know our index type will map to size_t which is unsigned, making a negative denominator impossible, so as long as we assert that this floordiv is working on values of `index` type it's safe to map this op to regular divi.

Prerequisite for DSL 2 work.

### DIFF
--- a/lib/Conversion/TTKernelToEmitC/TTKernelToEmitC.cpp
+++ b/lib/Conversion/TTKernelToEmitC/TTKernelToEmitC.cpp
@@ -674,6 +674,34 @@ public:
 } // namespace
 
 namespace {
+// Arith FloorDivSIOp doesn't have an emitc lowering, probably because of the
+// spec which says:
+//   Signed integer division. Rounds towards negative infinity, i.e. 5 / -2 = -3
+//
+// However we know our index type will map to size_t which is unsigned, making a
+// negative denominator impossible, so as long as we assert that this floordiv
+// is working on values of `index` type it's safe to map this op to regular
+// divi.
+class ArithFloorDivRewriter : public OpConversionPattern<arith::FloorDivSIOp> {
+public:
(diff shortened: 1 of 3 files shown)


lib/Conversion/TTKernelToEmitC/TTKernelToEmitC.cpp
```diff
@@ -673,6 +673,34 @@ class TTKernelTensorAccessorOpsRewriter : public OpConversionPattern<SourceOp> {
...
+// However we know our index type will map to size_t which is unsigned, making a
+// negative denominator impossible, so as long as we assert that this floordiv
+// is working on values of `index` type it's safe to map this op to regular
+// divi.
+class ArithFloorDivRewriter : public OpConversionPattern<arith::FloorDivSIOp> {
+public:
+  using OpConversionPattern<arith::FloorDivSIOp>::OpConversionPattern;
+
+  LogicalResult
+  matchAndRewrite(arith::FloorDivSIOp op, arith::FloorDivSIOp::Adaptor adaptor,
```
vroubtsovTT, line 690: I have to ask: this being a conversion pattern, should it not obtain operands types from the `adaptor`?

nsmithtt: The types on the adaptor will be the post-converted types (in this case `!emitc.size_t`), which could also be a valid approach for the early out condition.  But I need to use the pre-converted types in order to legally replace this op with a different arith op which will then be recursively converted.

vroubtsovTT (approved):

PR #5118, opened 2025-09-26
[ttnn.jit] E2E eltwise unary.

vtangTT:
### Ticket
Link to Github Issue

### Problem description

### What's changed

### Checklist

### DIFF
--- a/python/Passes.cpp
+++ b/python/Passes.cpp
@@ -127,6 +127,28 @@ void populatePassesModule(nb::module_ &m) {
       nb::arg("module"), nb::arg("options") = "");
 
   m.def(
+      "ttnn_to_ttmetal_pipeline",
+      [](MlirModule module, std::string options = "") {
+        mlir::Operation *moduleOp = unwrap(mlirModuleGetOperation(module));
+        mlir::PassManager pm(moduleOp->getName());
+
+        pm.addPass(tt::createConvertTTNNToTTIRPass());
+
+        const auto *pipeline =
+            mlir::PassPipelineInfo::lookup("ttir-to-ttmetal-pipeline");
+        std::function<mlir::LogicalResult(const llvm::Twine &)> err_handler =
+            [](const llvm::Twine &) { return mlir::failure(); };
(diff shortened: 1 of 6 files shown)


python/Passes.cpp
```diff
@@ -126,6 +126,28 @@ void populatePassesModule(nb::module_ &m) {
...
+  m.def(
+      "ttnn_to_ttmetal_pipeline",
+      [](MlirModule module, std::string options = "") {
+        mlir::Operation *moduleOp = unwrap(mlirModuleGetOperation(module));
+        mlir::PassManager pm(moduleOp->getName());
+
+        pm.addPass(tt::createConvertTTNNToTTIRPass());
+
+        const auto *pipeline =
+            mlir::PassPipelineInfo::lookup("ttir-to-ttmetal-pipeline");
```
nsmithtt, line 138: Doesn't ttnn mode need to be set on the pipeline?

Somewhat preferred method of running pipelines (that we don't seem to be following closely) is to use the generic python API's directly.  See `tools/builder/base/builder_utils.py`:
```python
pipeline_str = f"builtin.module({','.join([register_device, pipeline])})"
with module.context:
    pm = PassManager.parse(pipeline_str)
    pm.enable_verifier(verify)
    print("Running custom pipeline:", pm)
    if print_ir:
        print_ir_path = print_ir if isinstance(print_ir, str) else None
        pm.enable_ir_printing(tree_printing_dir_path=print_ir_path)
    pm.run(module.operation)
```

It's easier to change / configure for testing if it's just controlled directly in python.

vwellsTT (approved):

python/Passes.cpp
vtangTT: `ttnn-mode` and other options are set in the decorator when we call the pybinded version. 

There isn't a pipeline yet for ttnn integration path, I'll make an issue and update that outside of this PR, will leave a ToDo in the decorator for now.

https://github.com/tenstorrent/tt-mlir/issues/5224

ctodTT (approved):

Issue #3851, opened 2025-06-18
Implement a TTIR DotGeneralOp path for CPU Fallback


vwellsTT:
### Is your feature request related to a problem? Please describe

We would like to support lowering a TTIR DotGeneralOp into something that we can execute on CPU via our fallback path.  This is a particularly tricky op because it isn't DPS format, unlike most of our ops; our lowering pipeline requires DPS currently.

### Describe the solution you'd like

We must find a way to lower a DotGeneralOp into something we can run on CPU through our pipeline; this might involve a decomposition strategy (see below), refactoring the op to make it DPS, or some specialized workarounds.  We must also exercise via suitable tests in https://github.com/tenstorrent/tt-mlir/blob/main/test/python/golden/test_ttir_ops.py

### Describe alternatives you've considered

My initial thought was to run TTIRToTTIRDecomp on this s.t. we can decompose DotGeneral into matmul (which is DPS) for most cases.  However, this has it's own challenges, because if we decomp after hoisting, the func signature will already be incorrect (since it is based on original non-DPS signature).  We could hoist after decomp instead, but that would require propagating the `should_hoist` attrs from the original op to all of the resulting decomposed ops with our current schema.

BenraouaneSoufiane:
Hello, can I work on this?

C-mmon:
Decomposing the DotGeneraOp, only concern area is hoisting which ends up optimizing data flow between different operation.
Trying to decompose DotGeneralOp after hoisiting has occured, will only break things.
Can you assign the issue to me, I will try to take the compiler changes and then try to start with some basic experiment.

vwellsTT:
Hi @C-mmon 
Yeah, we would be happy for you to take a look at this.  Please note that this issue is dependent upon an open PR linked in the ticket; I'd recommend waiting till it lands (should be sometime this week), but if you'd like to begin before that, you work off of that sidebranch.  Please take a look at the testing implemented in this PR, and TTIRToLinalg conversion of matmul op there.
Also note that decomposition already exists for this op, but the path to utilize this with the TTIRToLinalg is less than clear.
Thanks.

C-mmon:
Hey, I have started a deep dive into it.

I will mostly start with some basic changes today on a separate pr, that will help

cguerreroTT:
Hi @C-mmon, I've assigned the bounty to you.

[...]

PR #5119, opened 2025-09-26
Dot general cpufallback

Mountagha:
### Ticket
#3851 

### Problem description
this PR supports lowering a TTIR DotGeneralOp into something that can be executed on CPU via our fallback path using a (permute + reshape + matmul) decomposition strategy.

### What's changed
Since dot_general is not DPS, it get decomposed into permute, reshape, matmul ops which are DPS.
Changed the order of passes execution in the pipeline (making sure we decompose before hoisting)
added a test_hoisted_matmul and make matmul TTIR->Linalg conversion DPS compliant. 

### Checklist


Mountagha:
hellow @vwellsTT. review plz and see if it's align with what you were expecting. Sorry for the late.

vwellsTT:
Hi @Mountagha, I'll do a proper review tomorrow, but at first glance this looks great, thanks!  Not late at all 🙂

lib/RegisterAll.cpp
```diff
@@ -94,6 +95,7 @@ void mlir::tt::registerAllExtensions(mlir::DialectRegistry &registry) {
   arith::registerBufferizableOpInterfaceExternalModels(registry);
   arith::registerBufferDeallocationOpInterfaceExternalModels(registry);
   linalg::registerBufferizableOpInterfaceExternalModels(registry);
+  linalg::registerSubsetOpInterfaceExternalModels(registry); // Add this line
```
vwellsTT: nit: remove comment

vwellsTT (approved):
One small nit, otherwise this look great to me!
@azecevicTT mind reviewing as well, to make sure your team is okay with the TTIRToTTIRDecomp changes etc?

lib/RegisterAll.cpp
```diff
@@ -37,6 +37,7 @@
 #include "mlir/IR/DialectRegistry.h"
 #include "mlir/InitAllDialects.h"
 #include "mlir/InitAllPasses.h"
+#include "mlir/Dialect/Linalg/Transforms/SubsetInsertionOpInterfaceImpl.h"
```
vwellsTT: our pre-commit is rejecting this line because it's not in alphabetical order w.r.t. other headers in same block btw, I think `clang-format` should fix

lib/Conversion/TTIRToTTIRDecomposition/TTIRToTTIRDecomposition.cpp
```diff
@@ -999,6 +1002,11 @@ struct DotGeneralToMatmulConversionPattern
         rewriter, op.getLoc(), matmulDestinationShape, elementType, encoding,
         lhsMatmulInput, rhsMatmulInput);
 
+    // Propagate the hoist attribute to the matmul op.
+    if (shouldHoist) {
+      matmulOp->setAttr("ttir.should_hoist", rewriter.getUnitAttr());
```
azecevic000: Nit: Can you please create a small util function (in TTIR/Utils.h(cpp)) `void addShouldHoistAttr(Operation *)`? It will make adding an attribute error proof (e.g. typo in the name of the attribute), and the name of the attribute would be centralized in one place if we ever decide to change it for some reason.

lib/Conversion/TTIRToTTIRDecomposition/TTIRToTTIRDecomposition.cpp
```diff
@@ -1070,17 +1083,23 @@ struct DotGeneralToMatmulConversionPattern
...
 
     SmallVector<int64_t> destinationShape =
         ttmlir::utils::applyPermutation(inputType.getShape(), permutation);
 
-    return ttir::utils::createDPSOp<ttir::PermuteOp>(
+    auto permuteOp = ttir::utils::createDPSOp<ttir::PermuteOp>(
         rewriter, loc, destinationShape, inputType.getElementType(),
         inputType.getEncoding(), input, permutation);
+
+    // propagate the hoist attribute if needed
```
azecevic000: Nit: write comment as a full sentence

lib/Conversion/TTIRToTTIRDecomposition/TTIRToTTIRDecomposition.cpp
```diff
@@ -1106,17 +1125,23 @@ struct DotGeneralToMatmulConversionPattern
...
+                                    bool shouldHoist = false) const {
 
     llvm::SmallVector<int32_t> finalShapeI32(finalShape.begin(),
                                              finalShape.end());
 
-    return ttir::utils::createDPSOp<ttir::ReshapeOp>(
+    auto reshapeOp = ttir::utils::createDPSOp<ttir::ReshapeOp>(
         rewriter, loc, finalShape, type.getElementType(), type.getEncoding(),
         input, rewriter.getI32ArrayAttr(finalShapeI32));
+    // propagate the hoist attribute if needed
```
azecevic000: Same here.

azecevicTT (approved):
Decomposition changes LGTM, few minor nits inline.

lib/RegisterAll.cpp
Mountagha: will fix it.

lib/Conversion/TTIRToTTIRDecomposition/TTIRToTTIRDecomposition.cpp
Mountagha: yes I can do that.

install_setup.sh
azecevic000, line 1: Please remove this file.

Mountagha: > Please remove this file.

done.

Mountagha:
@vwellsTT @azecevicTT please Re-review (not on the weekend :smile: ). I added more changes to fix the broken CI. Will be happy to answer questions.

include/ttmlir/Conversion/Passes.td
```diff
@@ -56,7 +56,9 @@ def TTIRToTTIRDecomposition: Pass<"ttir-to-ttir-decomposition", "::mlir::ModuleO
              clEnumValN(::mlir::tt::DecompMode::TTMetal, "ttmetal",
                         "Configuration for TTMetal backend"),
              clEnumValN(::mlir::tt::DecompMode::CPUFallback, "cpu-fallback",
-                        "Configuration for CPU fallback path")
+                        "Configuration for CPU fallback path"),
```
vwellsTT: I think we could use `CPUFallback` instead of adding a new option

test/python/golden/test_ttir_ops.py
```diff
@@ -2357,7 +2357,8 @@ def permute_wrapper(
         builder: TTIRBuilder,
         unit_attrs: Optional[List[str]] = None,
     ):
-        return permute(in0, in1, builder, permutation, unit_attrs=["ttir.should_hoist"])
+        _ = in1  # Unused, but required to match shapes
```
vwellsTT: I'm not following this comment, what is this needed for? and same with `_ = out0` below?

vwellsTT:
> @vwellsTT @azecevicTT please Re-review (not on the weekend 😄 ). I added more changes to fix the broken CI. Will be happy to answer questions.

Thanks! I've kicked off CI again, hopefully it works now 😄 
I left a couple comments, but I think this still LGTM for the core logic.

include/ttmlir/Conversion/Passes.td
vwellsTT: (I added it for these sort of CPU hoisted cases as a placeholder originally 🙂 )

Mountagha: hmmm okay then I might need to modify the case: CPUFallback (since hoisted Quant/Dequant want to decompose before hoist and hoisted Dot_general want it to go the other way around hoist -> decompose)
I am not sure my comment is clear. Lemme know if you need me to clarify on discord :smile:

test/python/golden/test_ttir_ops.py
Mountagha: ooh these are supposed to be local tests. Will get rid of them right away

include/ttmlir/Conversion/Passes.td
vwellsTT: Ah, I see.  Hmm, yeah if we need separate versions of this I think this is fine then, perhaps we should just change the descriptions to make each role a bit clearer

Mountagha: aight. Lemme see if they both can cohabitate in the same case. Othewrise I will keep the change and make it more clear.

Mountagha: @vwellsTT as you stated there was no need for a new case mode in the TTIRToDecompositionPass, hence I reverted it back to use the CPUFallback only.

PR #5140, opened 2025-09-29
Support arith and math dialect conversion to d2m tile ops

nsmithtt:
This is a prerequisite pass for kernel dsl work.  It also enables us to have better support for these upstream dialects which could be useful in other compilation contexts.

### DIFF
--- a/include/ttmlir/Conversion/Utils/D2MTileOpRewriter.h
+++ b/include/ttmlir/Conversion/Utils/D2MTileOpRewriter.h
@@ -2,33 +2,43 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 
+#ifndef TTMLIR_CONVERSION_UTILS_D2MTILEOPREWRITER_H
+#define TTMLIR_CONVERSION_UTILS_D2MTILEOPREWRITER_H
+
 #include "mlir/Transforms/DialectConversion.h"
 
 namespace mlir::tt::d2m {
 
-template <typename SourceOp, typename D2MTileOp, int Arity>
-class TileOpRewriter : public OpConversionPattern<SourceOp> {
+template <typename SourceOp, typename D2MTileOp>
+class UnaryTileOpRewriter : public OpConversionPattern<SourceOp> {
 public:
   using OpConversionPattern<SourceOp>::OpConversionPattern;
-
-  static_assert(Arity == 1 || Arity == 2);
 
   LogicalResult
   matchAndRewrite(SourceOp op, typename SourceOp::Adaptor adaptor,
                   ConversionPatternRewriter &rewriter) const final {
     auto operands = adaptor.getOperands();
-    if constexpr (Arity == 1) {
-      rewriter.replaceOpWithNewOp<D2MTileOp>(op, op.getResult().getType(),
-                                             operands[0]);
-      return success();
-    } else if constexpr (Arity == 2) {
-      rewriter.replaceOpWithNewOp<D2MTileOp>(op, op.getResult().getType(),
-                                             operands[0], operands[1]);
-      return success();
-    }
+    rewriter.replaceOpWithNewOp<D2MTileOp>(op, op.getResult().getType(),
+                                           operands[0]);
+    return success();
+  }
+};
 
-    return failure();
+template <typename SourceOp, typename D2MTileOp>
+class BinaryTileOpRewriter : public OpConversionPattern<SourceOp> {
+public:
+  using OpConversionPattern<SourceOp>::OpConversionPattern;
+
+  LogicalResult
+  matchAndRewrite(SourceOp op, typename SourceOp::Adaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const final {
+    auto operands = adaptor.getOperands();
+    rewriter.replaceOpWithNewOp<D2MTileOp>(op, op.getResult().getType(),
+                                           operands[0], operands[1]);
+    return success();
   }
 };
 
 } // namespace mlir::tt::d2m
+
+#endif // TTMLIR_CONVERSION_UTILS_D2MTILEOPREWRITER_H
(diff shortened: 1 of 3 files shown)


lib/Conversion/ArithToD2MTileOps/ArithToD2MTileOps.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+    target.addLegalDialect<mlir::BuiltinDialect>();
+    target.addLegalDialect<mlir::func::FuncDialect>();
+    target.addLegalDialect<mlir::linalg::LinalgDialect>();
+    target.addLegalDialect<mlir::arith::ArithDialect>();
+    target.addLegalDialect<mlir::scf::SCFDialect>();
+    target.addLegalDialect<ttcore::TTCoreDialect>();
+    target.addLegalDialect<d2m::D2MDialect>();
+
+    // Mark arith ops inside of a d2m.generic that operate on tiles as illegal.
```
vwellsTT: Do we expect/want to have arith ops in other places?

include/ttmlir/Conversion/Utils/D2MTileOpRewriter.h
```diff
@@ -0,0 +1,34 @@
+// SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#include "mlir/Transforms/DialectConversion.h"
+
+namespace mlir::tt::d2m {
+
+template <typename SourceOp, typename D2MTileOp, int Arity>
+class TileOpRewriter : public OpConversionPattern<SourceOp> {
```
vwellsTT: I don't feel super strongly, but it seems like in other dialects we always have separate class for unary vs binary--if nothing else I guess it would be trivial to sort of "curry" the Arity in wrapper classes?

vwellsTT (approved):

lib/Conversion/MathToD2MTileOps/MathToD2MTileOps.cpp
```diff
@@ -0,0 +1,82 @@
...
+
+    target.addLegalDialect<mlir::BuiltinDialect>();
+    target.addLegalDialect<mlir::func::FuncDialect>();
+    target.addLegalDialect<mlir::linalg::LinalgDialect>();
+    target.addLegalDialect<mlir::math::MathDialect>();
+    target.addLegalDialect<mlir::scf::SCFDialect>();
+    target.addLegalDialect<ttcore::TTCoreDialect>();
+    target.addLegalDialect<d2m::D2MDialect>();
+
+    // Mark math ops inside of a d2m.generic that operate on tiles as illegal.
```
vwellsTT: And same here, do we want math ops legal outside generics?

test/ttmlir/Conversion/MathToD2MTileOps/math_to_d2m_tile_ops.mlir
```diff
@@ -0,0 +1,10 @@
+// RUN: ttmlir-opt --math-to-d2m-tile-ops -o %t %s
+// RUN: FileCheck %s --input-file=%t
+
+!tile = !ttcore.tile<32x32, f32>
+
+func.func @absf(%arg0: !tile) -> !tile {
+  // CHECK: d2m.tile_abs
```
vwellsTT, line 7: I'm actually quite surprised this is converted--I thought since this op is marked as legal (since it's not in a generic), it should be left alone, no?

vwellsTT: Oh wait, I think your comments are lying and the `dynamicallyLegal` func has nothing to do with generics 😆 
can you update em?

nsmithtt: Yes sorry, the check used to also check for a generic parent, but I relaxed it so I could write tests like this! My b, I'll update the comment.

include/ttmlir/Conversion/Utils/D2MTileOpRewriter.h
nsmithtt: I sort of copied this pattern from the D2MToTTKernel SFPU rewriter which is templated like this.  I could break it out, only did this to share the boilerplate.

vwellsTT: Gotcha--yeah I'm not sure, maybe it's better like this.  I don't really think there's any actual common code between the two cases though, so imo splitting might be better

test/ttmlir/Conversion/MathToD2MTileOps/math_to_d2m_tile_ops.mlir
vwellsTT: makes sense--maybe a moot point if we can just mark entire dialects as illegal instead

nsmithtt: The issue is you can't mark whole dialects as conditionally illegal, hence I used that `GET_OP_LIST` macro to feed the template for `target.addDynamicallyLegalOp`.  We can only legally convert these ops if their argument type is a tile type.

vwellsTT: Hmm, that's only true if we support intermixing `arith` or `math` ops elsewhere in our IR though, right? I guess perhaps there would be some reason to do that in the future, but today I think this would be the only such case

nsmithtt: Yeah we do, all of our kernels use arith to do integer math for loop / index calculations.

vwellsTT: Oh, good point

Issue #5132, opened 2025-09-29
[EmitPy] Handle UpdateCache scalar packed as tensor


svuckovicTT:
TTNN's UpdateCacheOp (dialect) has the update index packaged as a 1-element tensor, but the lib expects a scalar. This was probably done to make it easier to implement in our runtime which can currently only track tensors.

```
  let arguments = (ins Arg<AnyRankedTensor, "cache tensor", [MemWrite]>:$cache,
                       AnyRankedTensor:$input,
                       AnyRankedTensor:$update_index,   <------
                       I32Attr:$batch_offset);
```

Runtime works around this by doing:
```cpp
    const ::ttnn::Tensor indexOnHost = ::ttnn::from_device(updateIndex);
    const ::tt::tt_metal::HostBuffer buffer =
        ::tt::tt_metal::host_buffer::get_host_buffer(indexOnHost);
    const auto &buf = buffer.view_as<uint32_t>();
    uint32_t upIdx = *buf.begin();

    ::ttnn::update_cache(cache, input, upIdx, op->batch_offset(), std::nullopt);
```

This isn't generally an issue, but complicates codegen. To fix this in EmitPy path, we can use the `tensor.item()` on single value tensors to unpack. Might need to do a `from_device` before that.

We currently have no member func op in EmitPy dialect, which is needed to call the `.item()` `ttnn.Tensor` function.

PR #5144, opened 2025-09-29
Add support for `ttnn.update_cache` op in EmitC/EmitPy

azecevicTT:
### Ticket
https://github.com/tenstorrent/tt-mlir/issues/5132

### Problem description
There wasn't support for `ttnn.update_cache` op, mainly because of the linked issue. 

### What's changed
- Added conversions in both EmitC/EmitPy
- `update_index` is obtained by calling a util function that extracts a scalar from a tensor that is effectively a scalar (similar to the runtime approach)

### Checklist


tools/tt-alchemist/templates/cpp/local/ttnn-precompiled.hpp
azecevic000, line 1: This looks like an ugly change because I've just copy-pasted the content from `tools/ttnn-standalone/ttnn-precompiled.hpp` (minus `ANCHOR`).

include/ttmlir/Conversion/TTNNToEmitPy/EmitPyConversion.h
```diff
@@ -1265,15 +1269,27 @@ class EmitPyTTNNEmitter {
...
 
-  mlir::Attribute emit(mlir::Value val, std::string attrName = "") {
+  // The `val` should be either an operand of the current source operation, in
+  // which case `index` should be nullopt, and the index it's found in the
+  // operands list. If `index` is provided, it means that the `val` is not an
+  // operand of the current source operation, and it is added as-is. Note that
+  // providing an `index` for an operand of the current source operation will
+  // result in an error.
+  mlir::Attribute emit(mlir::Value val, std::string attrName = "",
+                       std::optional<uint32_t> index = std::nullopt) {
```
azecevic000, line 1236: Still contemplating if this is the right approach. It's a bit hacky, the alternative is to do the whole conversion without the emitter class.

svuckovicTT: It's pretty ugly, but I don't see a better way :)

lib/Conversion/TTNNToEmitC/TTNNToEmitC.cpp
```diff
@@ -3157,6 +3157,45 @@ class ExecuteTraceOpConversionPattern
...
+  LogicalResult
+  matchAndRewrite(mlir::tt::ttnn::UpdateCacheOp srcOp, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+
+    ttnn_to_emitc::EmitCTTNNEmitter<mlir::tt::ttnn::UpdateCacheOp> emitter(
+        srcOp, adaptor, rewriter);
+
+    // The `update_index` is modeled as a tensor in the IR, but the
+    // `ttnn::update_cache` expects a `uint32_t` scalar.
+    auto updateIndex =
```
svuckovicTT: Easier to see the type this way since we invoke `.getResult(0)`:
```suggestion
    mlir::Value updateIndex =
```

tools/tt-alchemist/templates/cpp/local/ttnn-precompiled.hpp
svuckovicTT: Yeah, my bad on not keeping these in sync properly. When writing the includes in alchemist, I purposefully picked the `#include "ttnn/...` relative pathing so that everything in TTNN has the same root, making it easier to maintain include dirs. This fits with tt-metal repo's install components that we're using in the standalone path (`metalium-runtime`, `metalium-dev`, `ttnn-runtime`, `ttnn-dev`).

I suggest we revert this back and add the missing includes.

svuckovicTT: These are the missing includes in alchemist's hpp:

```cpp
#include "ttnn/operations/data_movement/sort/sort.hpp"
#include "ttnn/operations/experimental/transformer/nlp_concat_heads/nlp_concat_heads.hpp"
#include "ttnn/operations/experimental/transformer/nlp_concat_heads_decode/nlp_concat_heads_decode.hpp"
#include "ttnn/operations/experimental/transformer/rotary_embedding_llama/rotary_embedding_llama.hpp"
#include "ttnn/operations/normalization/rmsnorm/rmsnorm.hpp"
#include "ttnn/operations/rand/rand.hpp"
#include "ttnn/operations/transformer/concatenate_heads/concatenate_heads.hpp"
#include "ttnn/operations/transformer/sdpa/sdpa.hpp"
#include "ttnn/operations/transformer/sdpa_decode/sdpa_decode.hpp"
```

svuckovicTT (approved):

sdjordjevicTT (approved):

test/ttmlir/EmitPy/utils.py
```diff
@@ -0,0 +1,12 @@
+# SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+import ttnn
+
+
+def get_scalar_from_tensor(tensor: ttnn.Tensor) -> int:
```
mtopalovicTT: Why is definition needed both here and in alchemist?

mtopalovicTT (approved):

tools/tt-alchemist/templates/cpp/local/ttnn-precompiled.hpp
azecevic000: Thanks for pointing that out. Changed `operations/` to `ttnn/operations`, so most changes are now reverted. I've opted to keep includes and their order in sync in all three files (up to the prefix) so it's easier to find diff.

test/ttmlir/EmitPy/utils.py
azecevic000: Technically, it's not used anywhere currently, depending on the testing strategy taken for EmitPy it might be used at some point and it can be useful for local testing at this point. Opted to create symlink for this file from `tt-alchemist`.

PR #5149, opened 2025-09-29
Ddilbaz/dot general with bias

ddilbazTT:
### Ticket
None

### Problem description
I am working on adding split_query_key_value_and_split_heads fusion support. To be able to do so, I need to fuse dot_general/matmul + add into linear op otherwise the fusion becomes very messy.

In this draft PR, I am showcasing my demo of merging matmul + add op into a linear op. I would appreciate feedback. 

The default case it fuses is: 
```
%0 = ttir.empty() : tensor<68x1024xf32>
%1 = "ttir.matmul"(%arg0, %arg1, %0) <{transpose_a = false, transpose_b = false}> : (tensor<68x1024xf32>, tensor<1024x1024xf32>, tensor<68x1024xf32>) -> tensor<68x1024xf32>
%2 = ttir.empty() : tensor<68x1024xf32>
%3 = "ttir.add"(%1, %arg2, %2) : (tensor<68x1024xf32>, tensor<68x1024xf32>, tensor<68x1024xf32>) -> tensor<68x1024xf32>
```

This becomes: 
```
%0 = ttir.empty() : tensor<68x1024xf32>
%1 = "ttir.linear"(%arg0, %arg1, %arg2, %0) <{transpose_a = false, transpose_b = false}> : (tensor<68x1024xf32>, tensor<1024x1024xf32>, tensor<68x1024xf32>, tensor<68x1024xf32>) -> tensor<68x1024xf32>
```

I also added a way to fuse if the chain of operations is: matmul -> reshape -> add
For example, 

```
%0 = ttir.empty() : tensor<68x1024xf32>
%1 = "ttir.matmul"(%arg0, %arg1, %0) <{transpose_a = false, transpose_b = false}> : (tensor<68x1024xf32>, tensor<1024x1024xf32>, tensor<68x1024xf32>) -> tensor<68x1024xf32>
%2 = ttir.empty() : tensor<2x34x1024xf32>
%3 = "ttir.reshape"(%1, %2) <{shape = [2 : i32, 34 : i32, 1024 : i32]}> : (tensor<68x1024xf32>, tensor<2x34x1024xf32>) -> tensor<2x34x1024xf32>
%4 = ttir.empty() : tensor<2x34x1024xf32>
%3 = "ttir.add"(%3, %arg2, %2) : (tensor<2x34x1024xf32>, tensor<2x34x1024xf32>, tensor<2x34x1024xf32>) -> tensor<2x34x1024xf32>
```

This becomes:
```
      %0 = ttir.empty() : tensor<68x1024xf32>
      %1 = "ttir.reshape"(%arg2, %0) <{shape = [68 : i32, 1024 : i32]}> : (tensor<2x34x1024xf32>, tensor<68x1024xf32>) -> tensor<68x1024xf32>
      %2 = ttir.empty() : tensor<68x1024xf32>
      %3 = "ttir.linear"(%arg0, %arg1, %1, %2) <{transpose_a = false, transpose_b = false}> : (tensor<68x1024xf32>, tensor<1024x1024xf32>, tensor<68x1024xf32>, tensor<68x1024xf32>) -> tensor<68x1024xf32>
      %4 = ttir.empty() : tensor<2x34x1024xf32>
      %5 = "ttir.reshape"(%3, %4) <{shape = [2 : i32, 34 : i32, 1024 : i32]}> : (tensor<68x1024xf32>, tensor<2x34x1024xf32>) -> tensor<2x34x1024xf32>
```

### What's changed
Add a fusion pass which merges matmul + add into linear op.


lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
```diff
@@ -1055,6 +1055,130 @@ class PadPoolingFusionPattern : public mlir::OpRewritePattern<PoolingOp> {
...
+      Value matmulOpB = matmulOp.getB();
+      llvm::ArrayRef<int64_t> matmulOpShape =
+          matmulOp.getResult().getType().getShape();
+      llvm::ArrayRef<int64_t> addOpShape =
+          addOp.getResult().getType().getShape();
+      SmallVector<int32_t> matmulShapeI32(matmulOpShape.begin(),
+                                          matmulOpShape.end());
+      SmallVector<int32_t> addShapeI32(addOpShape.begin(), addOpShape.end());
+
+      auto biasType = mlir::cast<RankedTensorType>(bias.getType());
```
mmanzoorTT: TTIR dialect uses `RankedTensorType` so there shouldn't be a need of a typecast.
```suggestion
      RankedTensorType biasType = bias.getType();
```

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
```diff
@@ -1055,6 +1055,130 @@ class PadPoolingFusionPattern : public mlir::OpRewritePattern<PoolingOp> {
...
+public:
+  using mlir::OpRewritePattern<AddOp>::OpRewritePattern;
+  mlir::LogicalResult
+  matchAndRewrite(AddOp addOp, mlir::PatternRewriter &rewriter) const final {
+    if (MatmulOp matmulOp = isFusableMatmulOp(addOp); matmulOp) {
+      Value bias = addOp.getLhs() == matmulOp.getResult() ? addOp.getRhs()
+                                                          : addOp.getLhs();
+      Value matmulOpA = matmulOp.getA();
+      Value matmulOpB = matmulOp.getB();
+      LinearOp linearOp = rewriter.create<LinearOp>(
```
mmanzoorTT: AddOp supports implicit broadcast. Does LinearOp also support implicit broadcast for bias? OR we won't need broadcast for bias?

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
```diff
@@ -1055,6 +1055,130 @@ class PadPoolingFusionPattern : public mlir::OpRewritePattern<PoolingOp> {
...
+          rewriter, addOp.getLoc(), matmulOpShape, biasType.getElementType(),
+          biasType.getEncoding(), bias,
+          rewriter.getI32ArrayAttr(matmulShapeI32));
+
+      LinearOp linearOp = ttir::utils::createDPSOp<ttir::LinearOp>(
+          rewriter, addOp.getLoc(), matmulOp.getResult().getType(), matmulOpA,
+          matmulOpB, reshapedBias, matmulOp.getTransposeA(),
+          matmulOp.getTransposeB());
+
+      auto linearOpType = mlir::cast<RankedTensorType>(linearOp.getType());
```
mmanzoorTT:
```suggestion
      auto linearOpType = linearOp.getType();
```

test/ttmlir/Dialect/TTIR/fusing/dot_general_with_bias_fusing.mlir
```diff
@@ -0,0 +1,101 @@
+// RUN: ttmlir-opt -ttir-fusing -ttir-to-ttir-decomposition -ttir-fusing %s
```
mmanzoorTT: Please create separate tests for positive and negative cases.

test/ttmlir/Dialect/TTIR/fusing/matmul_with_bias_positive.mlir
```diff
@@ -0,0 +1,101 @@
+// RUN: ttmlir-opt -ttir-fusing -ttir-to-ttir-decomposition -ttir-fusing %s
+
```
mmanzoorTT, line 3: file name should use `matmul` instead of `dot_general` to match Fusing pattern name.

test/ttmlir/Dialect/TTIR/fusing/dot_general_with_bias_fusing.mlir
mmanzoorTT: You are running 3 passes here. Does first `ttir-fusing` perform any fusion? If not then you can remove the first pass. 
The other possibility can be to use the graph after `ttir-to-ttir-decomposition` as test and only run `ttir-fusing` pass once but the graph will be larger in this case.

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
ddilbazTT: What is implicit broadcast and how can I check it?

test/ttmlir/Dialect/TTIR/fusing/dot_general_with_bias_fusing.mlir
ddilbazTT: Yeah, I was trying to emulate the ttir passes we do when lowering to ttnn. The first one doesn't do anything if I keep this is a matmul fusion pattern. However, it will do the dot general fusion if I change this to a dot general fusion pattern. I think `-ttir-to-ttir-decomposition` is where we convert dot general to matmul.

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
```diff
@@ -1055,6 +1055,130 @@ class PadPoolingFusionPattern : public mlir::OpRewritePattern<PoolingOp> {
...
+
+    if (!matmulOp) {
+      return MatmulOp(nullptr);
+    }
+
+    // Check that the MatmulOp has only one user
+    auto users = matmulOp.getResult().getUsers();
+    if (llvm::range_size(users) != 1) {
+      return MatmulOp(nullptr);
+    }
```
mmanzoorTT:
```suggestion
    if (!matmulOp.getResult().hasOneUse()) {
      return MatmulOp(nullptr);
    }
```

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
```diff
@@ -1055,6 +1055,130 @@ class PadPoolingFusionPattern : public mlir::OpRewritePattern<PoolingOp> {
...
```
mmanzoorTT:
```suggestion
    if (!matmulOp.getResult().hasOneUse()) {
      return MatmulOp(nullptr);
    }
```

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
ddilbazTT: I guess this is a question of how to interpret bias, correct me if I am wrong. For example, we have this extended MLIR:

```
    %108 = "ttir.dot_general"(%94, %arg19) <{batch_dims_lhs = array<i64>, batch_dims_rhs = array<i64>, contract_dims_lhs = array<i64: 1>, contract_dims_rhs = array<i64: 0>}> : (tensor<68x1024xf32>, tensor<1024x1024xf32>) -> tensor<68x1024xf32>
    %109 = ttir.empty() : tensor<1x1024xf32>
    %110 = "ttir.reshape"(%arg18, %109) <{shape = [1 : i32, 1024 : i32]}> : (tensor<1024xf32>, tensor<1x1024xf32>) -> tensor<1x1024xf32>
    %111 = ttir.empty() : tensor<68x1024xf32>
    %112 = "ttir.broadcast"(%110, %111) <{broadcast_dimensions = array<i64: 68, 1>}> : (tensor<1x1024xf32>, tensor<68x1024xf32>) -> tensor<68x1024xf32>
    %113 = ttir.empty() : tensor<68x1024xf32>
    %114 = "ttir.add"(%108, %112, %113) : (tensor<68x1024xf32>, tensor<68x1024xf32>, tensor<68x1024xf32>) -> tensor<68x1024xf32>
    %115 = ttir.empty() : tensor<2x34x16x64xf32>
    %116 = "ttir.reshape"(%114, %115) <{shape = [2 : i32, 34 : i32, 16 : i32, 64 : i32]}> : (tensor<68x1024xf32>, tensor<2x34x16x64xf32>) -> tensor<2x34x16x64xf32>
    %117 = ttir.empty() : tensor<2x16x64x34xf32>
    %118 = "ttir.permute"(%116, %117) <{permutation = array<i64: 0, 2, 3, 1>}> : (tensor<2x34x16x64xf32>, tensor<2x16x64x34xf32>) -> tensor<2x16x64x34xf32>
    %119 = ttir.empty() : tensor<32x64x34xf32>
    %120 = "ttir.reshape"(%118, %119) <{shape = [32 : i32, 64 : i32, 34 : i32]}> : (tensor<2x16x64x34xf32>, tensor<32x64x34xf32>) -> tensor<32x64x34xf32>
```
I assumed the bias into the linear op would be `%112`, not `%110` or `%arg18`.

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
```diff
@@ -1055,6 +1055,130 @@ class PadPoolingFusionPattern : public mlir::OpRewritePattern<PoolingOp> {
...
+          rewriter.getI32ArrayAttr(addShapeI32));
+      rewriter.replaceOp(addOp, finalReshape);
+
+      return mlir::success();
+    }
+    return mlir::failure();
+  }
+
+private:
+  MatmulOp isFusableReshapeMatmulOp(AddOp addOp) const {
```
mmanzoorTT: You can also create another function for commone part of `isFusableReshapeMatmulOp` and `isFusableMatmulOp`. 
```
  if (matmulOpLHS && matmulOpRHS) {
      // Both operands are MatmulOps, cannot fuse
      return MatmulOp(nullptr);
    }
    MatmulOp matmulOp = matmulOpLHS ? matmulOpLHS : matmulOpRHS;

    if (!matmulOp) {
      return MatmulOp(nullptr);
    }

    // Check that the MatmulOp has only one user
    auto users = matmulOp.getResult().getUsers();
    if (llvm::range_size(users) != 1) {
      return MatmulOp(nullptr);
    }

    return matmulOp;
```

mmanzoorTT: and call it as
```
   return commonCode(matmulOpLHS, matmulOpRHS);
```

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
mmanzoorTT: yup. We have broadcast folding pattern which will remove `%112` and use `%110` directly for addOp as tt-metal can do this broadcasting internally. `ImplicitBroadcastFold` pass comes after `TTIRFusion` pass so the broadcast should be here.

lib/Dialect/TTNN/Pipelines/TTNNPipelines.cpp
```diff
@@ -170,6 +170,7 @@ void createTTIRToTTNNBackendPipeline(
   // Run regular TTIR to TTNN pipeline on DeviceModule.
   OpPassManager &devicePm =
       pm.nest<ttcore::DeviceModuleOp>().nest<mlir::ModuleOp>();
+  createTTNNPipelineTTIRImplicitBroadcastFoldPass(devicePm, options);
```
ddilbazTT: I added implicit broadcasting here before ttir passes. @AleksKnezevic @mmanzoorTT

test/ttmlir/Dialect/TTNN/trace/linear_add_consteval.mlir
```diff
@@ -0,0 +1,34 @@
+// RUN: ttmlir-opt --ttir-to-ttnn-backend-pipeline="enable-const-eval=true enable-trace=true" -o %t %s
```
ddilbazTT: I renamed `matmul_add_consteval.mlir` and replaced matmul with a linear op.

test/ttmlir/Dialect/TTNN/trace/linear_add_no_consteval.mlir
```diff
@@ -0,0 +1,30 @@
+// RUN: ttmlir-opt --ttir-to-ttnn-backend-pipeline="enable-const-eval=false enable-trace=true" -o %t %s
+// RUN: FileCheck %s --input-file=%t
```
ddilbazTT: I renamed `matmul_add_no_consteval.mlir` and replaced matmul with a linear op.

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
ddilbazTT: Hm, I still get: 
```
/localdev/ddilbaz/tt-mlir/lib/Dialect/TTIR/Transforms/TTIRFusing.cpp:1104:24: error: no viable conversion from 'Type' to 'RankedTensorType'
 1104 |       RankedTensorType biasType = bias.getType();
      |                        ^          ~~~~~~~~~~~~~~
```

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
ddilbazTT: Done!

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
ddilbazTT: Done

lib/Dialect/TTNN/Pipelines/TTNNPipelines.cpp
LPanosTT: Placing implicit broadcast folding before the TTNNPipelineTTIRPasses will stop EraseInverseOps from being able to commute through broadcasts, thus disallowing a lot of ops to be removed from the graph or at best, end up constevaling broadcasted values taking up more space on device than necessary. We shouldn't do this. The linear fusion pattern should handle cases where broadcasts are in the pattern.

lib/Conversion/TTNNToEmitPy/TTNNToEmitPy.cpp
```diff
@@ -359,6 +359,40 @@ class MatmulOpConversionPattern
 };
 } // namespace
 
+// LinearOp conversion pattern
```
vwellsTT: nit: end with .

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
```diff
@@ -1055,6 +1055,120 @@ class PadPoolingFusionPattern : public mlir::OpRewritePattern<PoolingOp> {
...
+      rewriter.replaceOp(addOp, finalReshape);
+
+      return mlir::success();
+    }
+    return mlir::failure();
+  }
+
+private:
+  // Shared helper function to validate a matmul op for fusion
+  MatmulOp validateMatmulForFusion(MatmulOp matmulOpLHS,
```
vwellsTT: nit: hmm, I found this func a bit confusing, I'd consider naming variables like `maybeMatmulOpLHS` or the like instead to indicate it might be null? idk

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
```diff
@@ -1055,6 +1055,120 @@ class PadPoolingFusionPattern : public mlir::OpRewritePattern<PoolingOp> {
...
+
+    // Check that the MatmulOp has only one user
+    if (!matmulOp.getResult().hasOneUse()) {
+      return MatmulOp(nullptr);
+    }
+
+    return matmulOp;
+  }
+
+  MatmulOp isFusableReshapeMatmulOp(AddOp addOp) const {
```
vwellsTT: for both this and below, I kind of think of `isX` as a boolean, perhaps something like `tryGetFusableReshapeMatmulOp` would express intent more clearly (although certainly a bit of a mouthful haha)

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
```diff
@@ -1055,6 +1055,120 @@ class PadPoolingFusionPattern : public mlir::OpRewritePattern<PoolingOp> {
...
+          rewriter.getI32ArrayAttr(addShapeI32));
+      rewriter.replaceOp(addOp, finalReshape);
+
+      return mlir::success();
+    }
+    return mlir::failure();
+  }
+
+private:
+  // Shared helper function to validate a matmul op for fusion
```
vwellsTT: nit: end with .

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
```diff
@@ -1055,6 +1055,120 @@ class PadPoolingFusionPattern : public mlir::OpRewritePattern<PoolingOp> {
...
+  MatmulOp validateMatmulForFusion(MatmulOp matmulOpLHS,
+                                   MatmulOp matmulOpRHS) const {
+    if (matmulOpLHS && matmulOpRHS) {
+      // Both operands are MatmulOps, cannot fuse
```
vwellsTT: nit: end with .

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
```diff
@@ -1055,6 +1055,120 @@ class PadPoolingFusionPattern : public mlir::OpRewritePattern<PoolingOp> {
...
+      return MatmulOp(nullptr);
+    }
+    MatmulOp matmulOp = matmulOpLHS ? matmulOpLHS : matmulOpRHS;
+
+    if (!matmulOp) {
+      return MatmulOp(nullptr);
+    }
+
+    // Check that the MatmulOp has only one user
```
vwellsTT: nit: end with .

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
```diff
@@ -1055,6 +1055,120 @@ class PadPoolingFusionPattern : public mlir::OpRewritePattern<PoolingOp> {
...
+    MatmulOp matmulOpLHS =
+        reshapeLHS ? reshapeLHS.getInput().getDefiningOp<MatmulOp>() : nullptr;
+    MatmulOp matmulOpRHS =
+        reshapeRHS ? reshapeRHS.getInput().getDefiningOp<MatmulOp>() : nullptr;
+
+    return validateMatmulForFusion(matmulOpLHS, matmulOpRHS);
+  }
+
+  MatmulOp isFusableMatmulOp(AddOp addOp) const {
+    // Check if one operand is a MatmulOp with only this AddOp as its user
```
vwellsTT: nit: end with .

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
```diff
@@ -1055,6 +1055,120 @@ class PadPoolingFusionPattern : public mlir::OpRewritePattern<PoolingOp> {
...
+
+      rewriter.replaceOp(addOp, linearOp);
+
+      return mlir::success();
+    }
+    if (MatmulOp matmulOp = isFusableReshapeMatmulOp(addOp); matmulOp) {
+
+      ReshapeOp reshapeOp =
+          mlir::dyn_cast<ReshapeOp>(*matmulOp.getResult().getUsers().begin());
+      assert(reshapeOp &&
```
vwellsTT: nit: Perhaps using new `TT_assertv` is better now

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
```diff
@@ -1055,6 +1055,120 @@ class PadPoolingFusionPattern : public mlir::OpRewritePattern<PoolingOp> {
...
+             "MatmulWithBiasFusionPattern isFusable logic is broken.");
+
+      Value bias = addOp.getLhs() == reshapeOp.getResult() ? addOp.getRhs()
```
vwellsTT: nit: personally, prefer surrounding condition with (), like 
```
      Value bias = (addOp.getLhs() == reshapeOp.getResult()) ? addOp.getRhs() : addOp.getLhs();

File: lib/Dialect/TTIR/Transforms/TTIRFusing.cpp [hunk 21]
```diff
@@ -1055,6 +1055,120 @@ class PadPoolingFusionPattern : public mlir::OpRewritePattern<PoolingOp> {
...
+    if (MatmulOp matmulOp = isFusableMatmulOp(addOp); matmulOp) {
+      Value bias = addOp.getLhs() == matmulOp.getResult() ? addOp.getRhs()
+                                                          : addOp.getLhs();
+      Value matmulOpA = matmulOp.getA();
+      Value matmulOpB = matmulOp.getB();
+      LinearOp linearOp = ttir::utils::createDPSOp<ttir::LinearOp>(
+          rewriter, addOp.getLoc(), addOp.getResult().getType(), matmulOpA,
+          matmulOpB, bias, matmulOp.getTransposeA(), matmulOp.getTransposeB());
+
+      rewriter.replaceOp(addOp, linearOp);
```
vwellsTT, line 1211: I'm probably dumb, but this should replace both the add and the matmul op, right? where is matmul op removed?

vwellsTT (approved):
Some comments inline, mostly LGTM

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
ddilbazTT: You're not dumb :D Matmul op doesn't need to be removed as it will be removed automatically since there are no users. Replacing add op with linear op meant matmul has no users anymore.

vwellsTT: Ah, makes sense thanks

test/ttmlir/Dialect/TTIR/fusing/dot_general_with_bias_fusing.mlir
mmanzoorTT: You are right. `ttir-to-ttir-decomposition` converts dot_general to matmul but it also adds permute/reshape ops.

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
```diff
@@ -1055,6 +1055,120 @@ class PadPoolingFusionPattern : public mlir::OpRewritePattern<PoolingOp> {
...
+
+      return mlir::success();
+    }
+    if (MatmulOp matmulOp = isFusableReshapeMatmulOp(addOp); matmulOp) {
+
+      ReshapeOp reshapeOp =
```
azecevic000, line 1217: You also need to check if `reshapeOp` has only one use. Otherwise, some other users will get a reshaped result of linear instead of matmul.

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
```diff
@@ -1055,6 +1055,120 @@ class PadPoolingFusionPattern : public mlir::OpRewritePattern<PoolingOp> {
...
+          biasType.getEncoding(), bias,
+          rewriter.getI32ArrayAttr(matmulShapeI32));
+
+      LinearOp linearOp = ttir::utils::createDPSOp<ttir::LinearOp>(
+          rewriter, addOp.getLoc(), matmulOp.getResult().getType(), matmulOpA,
+          matmulOpB, reshapedBias, matmulOp.getTransposeA(),
+          matmulOp.getTransposeB());
+
+      RankedTensorType linearOpType =
+          mlir::cast<RankedTensorType>(linearOp.getType());
```
azecevic000: Redundant cast.

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
```diff
@@ -1055,6 +1055,120 @@ class PadPoolingFusionPattern : public mlir::OpRewritePattern<PoolingOp> {
...
+
+      Value finalReshape = ttir::utils::createDPSOp<ttir::ReshapeOp>(
+          rewriter, addOp.getLoc(), addOpShape, linearOpType.getElementType(),
+          linearOpType.getEncoding(), linearOp.getResult(),
```
azecevic000: It shouldn't make a difference here, but logically it makes more sense to replace `linearOpType.getElementType(), linearOpType.getEncoding()` with `addOp.getElementType(), addOp.getEncoding()`.

test/ttmlir/Dialect/TTNN/trace/matmul_add_consteval.mlir
azecevic000, line 1: Why is this deleted? You can change test to something like `matmul->max` (or whatever binary eltwise op), because it's testing trace, so concrete op types aren't that important, but testing that trace is functioning properly is.

test/ttmlir/Dialect/TTNN/trace/matmul_add_no_consteval.mlir
azecevic000, line 1: Same here.

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
ddilbazTT: There are two places I cast because I get an error. For example:

```
/localdev/ddilbaz/tt-mlir/lib/Dialect/TTIR/Transforms/TTIRFusing.cpp:1104:24: error: no viable conversion from 'Type' to 'RankedTensorType'
 1104 |       RankedTensorType biasType = bias.getType();
      |                        ^          ~~~~~~~~~~~~~~
```

I think I should keep these casts, no?

test/ttmlir/Dialect/TTNN/trace/matmul_add_consteval.mlir
ddilbazTT: I didn't delete it, I renamed it to `linear_with_add` and replaced matmuls with linear ops. The tests are still testing the same functionality.

test/ttmlir/Dialect/TTNN/trace/linear_add_consteval.mlir
ddilbazTT: @azecevicTT

test/ttmlir/Dialect/TTNN/trace/linear_add_no_consteval.mlir
ddilbazTT: @azecevicTT

lib/Dialect/TTNN/Pipelines/TTNNPipelines.cpp
ddilbazTT: I also talked with @sdjordjevicTT I'll handle broadcast folding afterwards. @LPanosTT recommended to fold after fusion patterns but for that, i think I need to declare linear op a broadcastable op in ttir. Regardless, I'll just do that as a separate PR.

lib/Conversion/TTNNToEmitPy/TTNNToEmitPy.cpp
ddilbazTT: I think this is a title so doesn't need a `.` I noticed other titles in `TTNNToEmitPy.cpp` don't end with `.` I will add `.` to the comments you mentioned below though.

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
ddilbazTT: I think `getFusableReshapeMatmulOp` and `getFusdableMatmulOp` are good ideas.

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
ddilbazTT: Also if I didn't declare as RankedTensorType and cast, I get:

```
/localdev/ddilbaz/tt-mlir/lib/Dialect/TTIR/Transforms/TTIRFusing.cpp:1107:61: error: no member named 'getElementType' in 'mlir::Type'
 1107 |           rewriter, addOp.getLoc(), matmulOpShape, biasType.getElementType(),
```

lib/Conversion/TTNNToEmitPy/TTNNToEmitPy.cpp
vwellsTT: I think it's kind of a silly rule, but it is in our coding standards https://docs.tenstorrent.com/tt-mlir/coding-guidelines.html#comments so I think we should follow universally, FWIW

ddilbazTT: I'll just do it for every instance in the file then. Sure thing

vwellsTT: Thanks!

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
azecevic000: Bias is a different story, you declared it with `Value` type, so its type value is going to have `Type` type, that's why you need to cast it. Here `linearOp.getType()` returns a value of `RankedTensorType` type. Btw, you could also delcare bias with `TypedValue<RankedTensorType>`, in which case you wouldn't need an explicit cast for `bias.getType()`.

test/ttmlir/Dialect/TTNN/trace/matmul_add_consteval.mlir
azecevic000: That part looks good; my bad, apologies for the confusion.

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
ddilbazTT: Ok, I have removed redundant cast from linear op type (now add op type) and used `addOp.getElementType(), addOp.getEncoding()`.

tools/explorer/test/run_tests.py
```diff
@@ -199,8 +199,8 @@ def test_execute_mnist_stablehlo():
 
 def test_execute_mnist_with_overrides():
     overrides = {
-        'loc("matmul_1"("MNISTLinear":4294967295:10))__17': {
-            "named_location": "matmul_1",
+        'loc("relu_3"("MNISTLinear":4294967295:6))': {
```
brataTT, line 202: LGTM for explorer
Thanks for updating it to get it passing

To check if we are passing the overrides without errors, this should be good
For execution, it may be better to point to the new fused op. I don't know a lot about that yet but can take a look at that after this PR is merged as a part of working on explorer's error infra

brataTT (approved):

jazpurTT (approved):

lib/Conversion/TTNNToEmitPy/TTNNToEmitPy.cpp
```diff
@@ -53,7 +53,7 @@ class TTNNToEmitPyBaseOpConversionPattern
 };
 } // namespace
 
-// EltwiseUnaryOp conversion pattern
+// EltwiseUnaryOp conversion pattern.
```
mmanzoorTT: This is not a complete sentence (and below as well). We shouldn't add `period` for such cases.

tools/explorer/test/run_tests.py
ddilbazTT: Sounds good! I think that would be great

lib/Conversion/TTNNToEmitPy/TTNNToEmitPy.cpp
ddilbazTT: Vincent asked me to add the `period`, and I added it to all instances so as not to look different for `linear op`.

jnie-TT (approved):

include/ttmlir/Dialect/TTNN/Transforms/Workarounds/Decomposition/LinearOpRewritePattern.h
```diff
@@ -0,0 +1,26 @@
...
+
+/// Rewrite pattern for decomposing batched LinearOp into MatMulOp + AddOp.
+class LinearOpRewritePattern : public mlir::OpRewritePattern<ttnn::LinearOp> {
+public:
+  using OpRewritePattern::OpRewritePattern;
+
+  mlir::LogicalResult
+  matchAndRewrite(ttnn::LinearOp srcOp,
+                  mlir::PatternRewriter &rewriter) const override;
+};
```
azecevic000: Can you provide more context here? Does batched linear fail on TTNN? Also, please add this in the PR description.

lib/Dialect/TTNN/Transforms/Workarounds/Decomposition/LinearOpRewritePattern.cpp
```diff
@@ -0,0 +1,151 @@
...
+#include "mlir/IR/BuiltinTypes.h"
+
+namespace mlir::tt::ttnn::workarounds::decomposition {
+
+// Rewrite Linear op into matmul + add if input B is batched.
+// Follows
+// third_party/tt-metal/src/tt-metal/ttnn/cpp/ttnn/operations/matmul/matmul.cpp.
+static bool isBatchedLinearOp(ttnn::LinearOp linearOp) {
+  RankedTensorType inputBType =
+      mlir::cast<RankedTensorType>(linearOp.getB().getType());
```
azecevic000: Redundant cast

lib/Dialect/TTNN/Transforms/Workarounds/Decomposition/LinearOpRewritePattern.cpp
```diff
@@ -0,0 +1,151 @@
...
+  if (rank < 2) {
+    return false;
+  }
+
+  for (int64_t i = 0; i < rank - 2; ++i) {
+    if (inputBShape[i] > 1) {
+      return true;
+    }
+  }
+  return false;
```
azecevic000:
```suggestion
  // Check if batched: any dimension before the last 2 has size > 1
  return rank < 2 || llvm::any_of(inputBShape.drop_back(2),
                                  [](int64_t dim) { return dim > 1; });
```

lib/Dialect/TTNN/Transforms/Workarounds/Decomposition/LinearOpRewritePattern.cpp
```diff
@@ -0,0 +1,151 @@
...
+}
+
+// Calculate the output shape of a matmul operation following tt-metal's logic.
+// Reference: ttnn/cpp/ttnn/operations/matmul/matmul.cpp
+static SmallVector<int64_t>
+computeMatmulOutputShape(llvm::ArrayRef<int64_t> shapeA,
+                         llvm::ArrayRef<int64_t> shapeB, bool transposeA,
+                         bool transposeB) {
```
azecevic000: `transposeA` and `transposeB` are unused parameters in this function. So, if one of them is set to true, the calculation won't be correct.

lib/Dialect/TTNN/Transforms/Workarounds/Decomposition/LinearOpRewritePattern.cpp
```diff
@@ -0,0 +1,151 @@
...
+      return true;
+    }
+  }
+  return false;
+}
+
+// Calculate the output shape of a matmul operation following tt-metal's logic.
+// Reference: ttnn/cpp/ttnn/operations/matmul/matmul.cpp
+static SmallVector<int64_t>
+computeMatmulOutputShape(llvm::ArrayRef<int64_t> shapeA,
```
azecevic000, line 37: It's really hard to infer the semantics of this calculation based on the code (a lot of state changes, with a lot of special cases). Writing it down in the form of a comment would really help.

IMHO handling special cases should be done at the beginning, not at the end of the function, for a multitude of reasons.

lib/Dialect/TTNN/Transforms/Workarounds/Decomposition/LinearOpRewritePattern.cpp
```diff
@@ -0,0 +1,151 @@
...
+  } else {
+    outputShape.assign(shapeA.begin(), shapeA.end());
+  }
+
+  // Handle batch dimensions for the case where rankB > rankA
+  for (int64_t index = 0; index < rankDifference; ++index) {
+    // tt-metal requires these front dimensions to be 1
+    // For our purposes, we'll just copy them
+    outputShape[index] = shapeB[index];
+  }
```
azecevic000: This isn't doing anything. If `rankB > rankA` then `outputShape` is already equal to `shapeB`. If `rankA >= rankB`, then `rankDifference == 0`.

lib/Dialect/TTNN/Transforms/Workarounds/Decomposition/LinearOpRewritePattern.cpp
```diff
@@ -0,0 +1,151 @@
...
+    SmallVector<int64_t> newShape;
+    newShape.reserve(outputShape.size() - 1);
+    for (size_t index = 0; index < outputShape.size() - 1; ++index) {
+      newShape.push_back(outputShape[index]);
+    }
+    outputShape = std::move(newShape);
+  }
+
+  return outputShape;
+}
```
azecevic000, line 58: One proposal for refactoring that doesn't take `transposeA` and `transposeB` into consideration
```suggestion
static SmallVector<int64_t>
computeMatmulOutputShape(llvm::ArrayRef<int64_t> shapeA,
                         llvm::ArrayRef<int64_t> shapeB) {

  int64_t rankA = shapeA.size();
  int64_t rankB = shapeB.size();

  // If input B is a vector, we drop the last dimension of shapeA.
  // (n_1,..., n_k, p, q) x (q, (implicit 1)) -> (n_1,..., n_k, p)
  if (rankB == 1) {
    return llvm::SmallVector<int64_t>(shapeA.begin(), shapeA.end() - 1);
  }

  // If input A is a vector, we drop the second-to-last dimension of shapeB.
  // Here, we can safely assume that rankB > 1.
  // ((implicit 1), p) x (n_1,..., n_k, p, q) -> (n_1,..., n_k, q)
  if (rankA == 1) {
    llvm::SmallVector<int64_t> outputShape(shapeB);
    outputShape.erase(outputShape.end() - 2);
    return outputShape;
  }

  // This would be a literal refactoring of the tt-metal code. There are a lot of implicit assumptions that take place here for the semantics to make any sense.
  int64_t outRank = std::max(rankA, rankB);
  SmallVector<int64_t> outputShape(outRank);
  outputShape.insert(std::next(outputShape.begin(), outRank - rankB), shapeB.begin(), shapeB.end());
  outputShape.insert(std::next(outputShape.begin(), outRank - rankA), shapeA.begin(), shapeA.end());
  // The last dimension comes from input B.
  // If we ignore batch dimensions, we end up with
  // (p, q) x (q, r) -> (p, r)
  outputShape[outRank - 1] = shapeB[rankB - 1];

  return outputShape;
}
```

lib/Dialect/TTNN/Transforms/Workarounds/Decomposition/LinearOpRewritePattern.cpp
```diff
@@ -0,0 +1,151 @@
...
+  if (!isBatchedLinearOp(srcOp) || !srcOp.getBias()) {
+    return failure();
+  }
+
+  RankedTensorType inputAType =
+      mlir::cast<RankedTensorType>(srcOp.getA().getType());
+  RankedTensorType inputBType =
+      mlir::cast<RankedTensorType>(srcOp.getB().getType());
+  RankedTensorType outputType =
+      mlir::cast<RankedTensorType>(srcOp.getResult().getType());
```
azecevic000: All three casts are redundant.

lib/Dialect/TTNN/Transforms/Workarounds/Decomposition/LinearOpRewritePattern.cpp
```diff
@@ -0,0 +1,151 @@
...
+  // Compute matmul output shape
+  SmallVector<int64_t> matmulShape =
+      computeMatmulOutputShape(inputAType.getShape(), inputBType.getShape(),
+                               srcOp.getTransposeA(), srcOp.getTransposeB());
+
+  // Create matmul output type
+  auto outputEncoding =
+      mlir::cast<ttnn::TTNNLayoutAttr>(outputType.getEncoding());
+  auto matmulOutputType = RankedTensorType::get(
+      matmulShape, outputType.getElementType(), outputType.getEncoding());
```
azecevic000:
```suggestion
  auto matmulOutputType = utils::RankedTensorTypeFactory::create(outputType, matmulShape);
```
It's not just nit, encoding can be dependent on the shape, `RankedTensorTypeFactory` handles it gracefully.

include/ttmlir/Dialect/TTNN/Transforms/Workarounds/Decomposition/LinearOpRewritePattern.h
ddilbazTT: Yes, batched linear op fails. For example, these examples from the dialect test actually can't run on hardware without hitting the error with `is_batched`: 
```
  // linear nd - nd tests
  func.func @linear_nd_nd_bias_broadcast_bias(%arg0: tensor<14x7x32x32xbf16>, %arg1:tensor<14x1x32x64xbf16>, %bias: tensor<64xbf16>) -> tensor<14x7x32x64xbf16> {
    %0 = ttir.empty() : tensor<14x7x32x64xbf16>
    // CHECK: "ttnn.linear"
    // CHECK-SAME: tensor<14x7x32x32xbf16
    // CHECK-SAME: tensor<14x1x32x64xbf16
    // CHECK-SAME: tensor<64xbf16
    // CHECK-SAME: tensor<14x7x32x64xbf16
    %1 = "ttir.linear"(%arg0, %arg1, %bias, %0) : (tensor<14x7x32x32xbf16>, tensor<14x1x32x64xbf16>, tensor<64xbf16>, tensor<14x7x32x64xbf16>) -> tensor<14x7x32x64xbf16>
    return %1 : tensor<14x7x32x64xbf16>
  }

  func.func @linear_nd_nd_bias_broadcast_matmul(%arg0: tensor<3x64x128xbf16>, %arg1: tensor<4x3x128x32xbf16>, %bias: tensor<14x4x3x64x32xbf16>) -> tensor<14x4x3x64x32xbf16> {
    %0 = ttir.empty() : tensor<14x4x3x64x32xbf16>
    // CHECK: "ttnn.linear"
    // CHECK-SAME: tensor<3x64x128xbf16
    // CHECK-SAME: tensor<4x3x128x32xbf16
    // CHECK-SAME: tensor<14x4x3x64x32xbf16
    %1 = "ttir.linear"(%arg0, %arg1, %bias, %0) : (tensor<3x64x128xbf16>, tensor<4x3x128x32xbf16>, tensor<14x4x3x64x32xbf16>, tensor<14x4x3x64x32xbf16>) -> tensor<14x4x3x64x32xbf16>
    return %1 : tensor<14x4x3x64x32xbf16>
  }
```
https://github.com/tenstorrent/tt-mlir/blob/main/test/ttmlir/Dialect/TTNN/linear/linear_tests_positive.mlir

I'll add more context to the PR description. The failure comes from TTNN: https://github.com/tenstorrent/tt-metal/blob/main/ttnn/cpp/ttnn/operations/matmul/matmul.cpp 
```
bool is_input_batched(const ttnn::Shape& shape) {
    if (shape.rank() < 2) [[unlikely]] {
        return false;
    }

    auto is_batched = false;
    for (auto i = 0; i < shape.rank() - 2; ++i) {
        if (shape[i] > 1) {
            is_batched = true;
            break;
        }
    }
    return is_batched;
}
...
    bool b_is_batched = detail::is_input_batched(input_tensor_b.logical_shape());
    TT_FATAL(!(b_is_batched && bias.has_value()), "Batched input not supported when bias exists (linear operation).");
```

lib/Dialect/TTNN/Transforms/Workarounds/Decomposition/LinearOpRewritePattern.cpp
ddilbazTT: I didn't come up with this calculation logic, I am using tt-metal's compute matmul shape logic:

```
ttnn::Shape compute_matmul_output_shape(const Tensor& input_tensor_a, const Tensor& input_tensor_b) {
    const auto& input_shape_a = input_tensor_a.logical_shape();
    const auto& input_shape_b = input_tensor_b.logical_shape();

    const auto a_rank = input_shape_a.rank();
    const auto b_rank = input_shape_b.rank();

    // Rank difference will be used to align batch dimensions
    const int32_t out_rank = std::max<int32_t>(a_rank, b_rank) - (a_rank == 1 || b_rank == 1);
    const int32_t rank_difference = std::max<int32_t>(0, out_rank - a_rank);

    // Initialize output shape based on the tensor with higher rank
    ttnn::Shape output_shape = (b_rank > a_rank) ? input_shape_b : input_shape_a;

    // Handle batch dimensions for the case where b_rank > a_rank
    for (auto index = 0; index < rank_difference; ++index) {
        TT_FATAL(input_shape_b[index] == 1, "When in1 rank greater than in0 rank front dimensions need to be 1");
        output_shape[index] = input_shape_b[index];
    }

    // Copy dimensions from input_shape_a except the last one
    for (auto index = 0; index < a_rank - 1; ++index) {
        output_shape[rank_difference + index] = input_shape_a[index];
    }

    // The last dimension comes from input_tensor_b
    output_shape[-1] = input_shape_b[-1];

    // Handle the vector matmul case: if a_rank == 1, remove the second-to-last dimension
    if (a_rank == 1 && output_shape.rank() > 1) [[unlikely]] {
        ttnn::SmallVector<uint32_t> new_shape(output_shape.rank() - 1);
        // Copy all elements except the second-to-last dimension
        size_t dst_idx = 0;
        for (size_t src_idx = 0; src_idx < output_shape.rank(); ++src_idx) {
            if (src_idx != output_shape.rank() - 2) {
                new_shape[dst_idx++] = output_shape[src_idx];
            }
        }
        output_shape = ttnn::Shape(new_shape);
    }

    // Handle the case where b_rank == 1, remove the last dimension
    if (b_rank == 1) [[unlikely]] {
        ttnn::SmallVector<uint32_t> new_shape(output_shape.rank() - 1);
        for (auto index = 0; index < output_shape.rank() - 1; ++index) {
            new_shape[index] = output_shape[index];
        }
        output_shape = ttnn::Shape(new_shape);
    }

    return output_shape;
}
```
https://github.com/tenstorrent/tt-metal/blob/bdaa4a2d9a7566302fe2f15c1be0da3ce2096886/ttnn/cpp/ttnn/operations/matmul/device/matmul_op.cpp#L1314

lib/Dialect/TTNN/Transforms/Workarounds/Decomposition/LinearOpRewritePattern.cpp
ddilbazTT: This is from tt-metal: https://github.com/tenstorrent/tt-metal/blob/bdaa4a2d9a7566302fe2f15c1be0da3ce2096886/ttnn/cpp/ttnn/operations/matmul/device/matmul_op.cpp#L1314

lib/Dialect/TTNN/Transforms/Workarounds/Decomposition/LinearOpRewritePattern.cpp
ddilbazTT: Given this function is re-implementing tt-metal logic: https://github.com/tenstorrent/tt-metal/blob/bdaa4a2d9a7566302fe2f15c1be0da3ce2096886/ttnn/cpp/ttnn/operations/matmul/device/matmul_op.cpp#L1314 I want to know if you would revise your suggestion.

lib/Dialect/TTNN/Transforms/Workarounds/Decomposition/LinearOpRewritePattern.cpp
ddilbazTT: This returns true when it shouldn't for the following case:
```
  func.func @linear_1d_1d_bias(%arg0: tensor<128xbf16>, %arg1: tensor<128xbf16>, %bias: tensor<1xbf16>) -> tensor<1xbf16> {
    %0 = ttir.empty() : tensor<1xbf16>
    // CHECK: "ttnn.linear"
    // CHECK-SAME: tensor<128xbf16
    // CHECK-SAME: tensor<128xbf16
    // CHECK-SAME: tensor<1xbf16
    // CHECK-SAME: tensor<1xbf16
    %1 = "ttir.linear"(%arg0, %arg1, %bias, %0) : (tensor<128xbf16>, tensor<128xbf16>, tensor<1xbf16>, tensor<1xbf16>) -> tensor<1xbf16>
    return %1 : tensor<1xbf16>
  }
```

ddilbazTT: This is probably reversed, will use the following instead:
```
return rank >= 2 && llvm::any_of(inputBShape.drop_back(2),
                                 [](int64_t dim) { return dim > 1; });
```

lib/Dialect/TTNN/Transforms/Workarounds/Decomposition/LinearOpRewritePattern.cpp
ddilbazTT: The section with 
```
  outputShape.insert(std::next(outputShape.begin(), outRank - rankB), shapeB.begin(), shapeB.end());
  outputShape.insert(std::next(outputShape.begin(), outRank - rankA), shapeA.begin(), shapeA.end());
```
is incorrect as it inserts twice. I will try to find a way to rewrite this.

ddilbazTT: I kept the first half of the suggestion where we have vector as either input. But for general case I changed it:

```
// Calculate the output shape of a matmul operation following tt-metal's logic.
// Reference: ttnn/cpp/ttnn/operations/matmul/matmul.cpp
static SmallVector<int64_t>
computeMatmulOutputShape(llvm::ArrayRef<int64_t> shapeA,
                         llvm::ArrayRef<int64_t> shapeB) {
  int64_t rankA = shapeA.size();
  int64_t rankB = shapeB.size();

  // Input B is a vector -> We drop the last dimension of shapeA.
  // (n_1,..., n_k, p, q) x (q, (implicit 1)) -> (n_1,..., n_k, p)
  if (rankB == 1) {
    return llvm::SmallVector<int64_t>(shapeA.begin(), shapeA.end() - 1);
  }
  // Input A is a vector -> We drop the second-to-last dimension of shapeB.
  int64_t outRank = std::max(rankA, rankB);
  SmallVector<int64_t> outputShape(outRank, 1);

  int64_t aOffset = outRank - rankA;
  int64_t bOffset = outRank - rankB;

  // Broadcast batch dimensions
  for (int64_t i = 0; i < outRank - 2; ++i) {
    int64_t aDim = (i >= aOffset) ? shapeA[i - aOffset] : 1;
    int64_t bDim = (i >= bOffset) ? shapeB[i - bOffset] : 1;
    outputShape[i] = std::max(aDim, bDim);
  }

  // Matmul inner dims: (…, p, q) x (…, q, r) -> (…, p, r)
  outputShape[outRank - 2] = shapeA[rankA - 2];
  outputShape[outRank - 1] = shapeB[rankB - 1];

  return outputShape;
}
```

include/ttmlir/Dialect/TTNN/Transforms/Workarounds/Decomposition/LinearOpRewritePattern.h
```diff
@@ -0,0 +1,26 @@
...
+#define TTMLIR_DIALECT_TTNN_TRANSFORMS_WORKAROUNDS_DECOMPOSITION_LINEAROPREWRITEPATTERN_H
+
+#include "ttmlir/Dialect/TTNN/IR/TTNNOps.h"
+
+#include "mlir/IR/PatternMatch.h"
+#include "mlir/Support/LogicalResult.h"
+
+namespace mlir::tt::ttnn::workarounds::decomposition {
+
+/// Rewrite pattern for decomposing batched LinearOp into MatMulOp + AddOp.
```
mtopalovicTT: remove extra /. There are also couple other places in PR which have extra /

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
```diff
@@ -1185,6 +1186,131 @@ class PadPoolingFusionPattern : public mlir::OpRewritePattern<PoolingOp> {
...
+    }
+    return mlir::failure();
+  }
+
+private:
+  // Shared helper function to validate the matmul ops from an add op.
+  MatmulOp getValidMatmulOp(MatmulOp matmulOpLHS, MatmulOp matmulOpRHS) const {
+    if (matmulOpLHS && matmulOpRHS) {
+      // Both operands are MatmulOps, cannot fuse.
+      return MatmulOp(nullptr);
```
mtopalovicTT: just nullptr? and below?

lib/Dialect/TTNN/Transforms/Workarounds/Decomposition/LinearOpRewritePattern.cpp
```diff
@@ -0,0 +1,112 @@
...
+
+// Rewrite Linear op into matmul + add if input B is batched.
+// Follows
+// third_party/tt-metal/src/tt-metal/ttnn/cpp/ttnn/operations/matmul/matmul.cpp.
+static bool isBatchedLinearOp(ttnn::LinearOp linearOp) {
+  RankedTensorType inputBType = linearOp.getB().getType();
+  auto inputBShape = inputBType.getShape();
+  int64_t rank = inputBShape.size();
+
+  // if rank < 2, it cannot be batched. Return false.
```
mtopalovicTT: If rank <= 2 it cannot be batched.

lib/Dialect/TTNN/Transforms/Workarounds/Decomposition/LinearOpRewritePattern.cpp
```diff
@@ -0,0 +1,112 @@
...
+                         llvm::ArrayRef<int64_t> shapeB) {
+  int64_t rankA = shapeA.size();
+  int64_t rankB = shapeB.size();
+
+  // Input B is a vector -> We drop the last dimension of shapeA.
+  // (n_1,..., n_k, p, q) x (q, (implicit 1)) -> (n_1,..., n_k, p)
+  if (rankB == 1) {
+    return llvm::SmallVector<int64_t>(shapeA.begin(), shapeA.end() - 1);
+  }
+  // Input A is a vector -> We drop the second-to-last dimension of shapeB.
```
mtopalovicTT: should this be followed by:
```
if (rankA == 1) {
       llvm::SmallVector<int64_t> outShape;
       for (size_t i = 0; i < shapeB.size(); ++i) {
               if (i != shape.size() - 2) {
                      result.push_back(shapeB[i]);
                }
        }
}
```

there is probably some nicer way to write this.

test/ttmlir/Dialect/TTNN/Transforms/Workarounds/linear_op_workaround.mlir
```diff
@@ -0,0 +1,34 @@
+// RUN: ttmlir-opt --ttcore-register-device --ttnn-layout --ttnn-workaround -o %t %s
+// RUN: FileCheck %s --input-file=%t
+
+module  {
+  func.func @linear_with_batched_rhs_and_bias(%arg0: tensor<2x33x1024xf32>, %arg1: tensor<2x1024x1024xf32>, %arg2: tensor<2x33x1024xf32>) -> tensor<2x33x1024xf32>{
```
mtopalovicTT, line 5: can you add tests when either A or B are vectors?

lib/Dialect/TTNN/Transforms/Workarounds/Decomposition/LinearOpRewritePattern.cpp
```diff
@@ -0,0 +1,112 @@
...
+// Reference: ttnn/cpp/ttnn/operations/matmul/matmul.cpp
+static SmallVector<int64_t>
+computeMatmulOutputShape(llvm::ArrayRef<int64_t> shapeA,
+                         llvm::ArrayRef<int64_t> shapeB) {
+  int64_t rankA = shapeA.size();
+  int64_t rankB = shapeB.size();
+
+  // Input B is a vector -> We drop the last dimension of shapeA.
+  // (n_1,..., n_k, p, q) x (q, (implicit 1)) -> (n_1,..., n_k, p)
+  if (rankB == 1) {
```
mtopalovicTT: Can we handle case when both a and b are 1?

lib/Dialect/TTNN/Transforms/Workarounds/Decomposition/LinearOpRewritePattern.cpp
```diff
@@ -0,0 +1,112 @@
...
+  // Broadcast batch dimensions
+  for (int64_t i = 0; i < outRank - 2; ++i) {
+    int64_t aDim = (i >= aOffset) ? shapeA[i - aOffset] : 1;
+    int64_t bDim = (i >= bOffset) ? shapeB[i - bOffset] : 1;
+    outputShape[i] = std::max(aDim, bDim);
+  }
+
+  // Matmul inner dims: (…, p, q) x (…, q, r) -> (…, p, r)
+  outputShape[outRank - 2] = shapeA[rankA - 2];
+  outputShape[outRank - 1] = shapeB[rankB - 1];
```
mtopalovicTT:
```suggestion
  SmallVector<int64_t> outputShape;
  SmallVector<int64_t> batchShapeA(shapeA.begin(), shapeA.end() - 2);
  SmallVector<int64_t> batchShapeB(shapeB.begin(), shapeB.end() - 2);

  mlir::OpTrait::util::getBroadcastedShape(
      batchShapeA, batchShapeB, outputShape);

  outputShape.push_back(shapeA[rankA - 2]);
  outputShape.push_back(shapeB[rankB - 1]);
```

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
ddilbazTT: Since it's not a valid matmul op, I am returning nullptr.

ddilbazTT: Basically:
- if both inputs to an add op is a matmul, I cannot fuse into a linear op since bias is unclear.
- if none of the inputs to an add op is a matmul, I cannot fuse since it's not the pattern I am looking for.

lib/Dialect/TTNN/Transforms/Workarounds/Decomposition/LinearOpRewritePattern.cpp
ddilbazTT: It's coming from tt-metal, this is their logic:

```
bool is_input_batched(const ttnn::Shape& shape) {
    if (shape.rank() < 2) [[unlikely]] {
        return false;
    }

    auto is_batched = false;
    for (auto i = 0; i < shape.rank() - 2; ++i) {
        if (shape[i] > 1) {
            is_batched = true;
            break;
        }
    }
    return is_batched;
}
```
But you're right, since the later for loop wouldn't loop when rank = 2. I'll just make it clear on my side. @azecevicTT and I talked offline and I'll ping tt-metal that they should update the linear op definition on their side.

ddilbazTT: I am changing it to 
```
  return rank > 2 && llvm::any_of(inputBShape.drop_back(2),
                                   [](int64_t dim) { return dim > 1; });
```
since it's clearer.

lib/Dialect/TTNN/Transforms/Workarounds/Decomposition/LinearOpRewritePattern.cpp
ddilbazTT: I think they wouldn't hit `is_batched` so we don't need to handle that case.

lib/Dialect/TTNN/Transforms/Workarounds/Decomposition/LinearOpRewritePattern.cpp
ddilbazTT: Sorry I was missing this, thanks for catching:
```
  if (rankA == 1) {
    llvm::SmallVector<int64_t> outputShape(shapeB);
    outputShape.erase(outputShape.end() - 2);
    return outputShape;
  }
```

test/ttmlir/Dialect/TTNN/Transforms/Workarounds/linear_op_workaround.mlir
ddilbazTT: I think if B is a vector, it wouldn't hit the is_batched therefore it wouldn't hit the workaround. I'll add an A is a vector test case though.

ddilbazTT: I actually can't add cases due to tt-metal limitations. When A is a vector, it is expecting the leading dims of B to be 1. However, when leading dims are equal to 1, that is not a batched B anymore. Therefore, it is a linear op that can execute. I'll add A and B are 1d cases to linear op silicon tests, but they are actually not hit in the workaround.

ddilbazTT:
Thanks @azecevicTT and @mtopalovicTT I appreciate your comments and suggestions!

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
mtopalovicTT: Yea yea I know. What i mean is you don't need to wrap nullptr into MatmulOp, you can just return nullptr.

lib/Dialect/TTNN/Transforms/Workarounds/Decomposition/LinearOpRewritePattern.cpp
mtopalovicTT: Okay. Do we need to then check if (rankB == 1) since first thing we check in the rewriter is if bRank > 2 and if not we fail to rewrite?

test/ttmlir/Dialect/TTNN/Transforms/Workarounds/linear_op_workaround.mlir
mtopalovicTT: Yea you are right for B, but you can still add the test case for A right?

ddilbazTT: For both cases, we are not encountering the batched input. When A is a vector, the batching dims of B need to be 1 as per tt-metal. But if 1, they can just execute as a regular linear op and don't need the workaround. I added tests for when A is a vector or B is a vector to the silicon tests. They lower to ttnn linear, they don't go through the workaround so I didn't add them to the workaround test.

lib/Dialect/TTNN/Transforms/Workarounds/Decomposition/LinearOpRewritePattern.cpp
ddilbazTT: I think implicitly assuming we won't hit that case might not be good coding practice. However, I can add an assert saying this condition shouldn't be encountered.

test/ttmlir/Dialect/TTNN/Transforms/Workarounds/linear_op_workaround.mlir
ddilbazTT: As per my message above, maybe I can add an assert for rankA == 1 || rankB == 1 that we shouldn't be encountering these cases.

Issue #4499, opened 2025-08-15
Eltwise fusion


nsmithtt:
We want to support trivial eltwise fusion, via 2 phases:
- ttir.generic fusion
- linalg fusion

For example, given graph:
```mlir
%a = ttir.generic {
  %0 = linalg.generic {
     mul
   }
  yield %0
}
%b = ttir.generic %a {
  %1 = linalg.generic {
     add
   }
  yield %1
}
```

We first run ttir.generic fusion pass:
```mlir
%a = ttir.generic {
  %0 = linalg.generic {
     mul
   }
   %1 = linalg.generic %0 {
     add
   }
  yield %1
}
```

Then we simply run the upstream linalg fusion pass:
```mlir
%a = ttir.generic {
  %0 = linalg.generic {
     mul
     add
   }
  yield %0
}
```

PR #5151, opened 2025-09-30
Mbagherbeik/eltwise fusion pass

mbagherbeikTT:
### Ticket
Link to Github Issue https://github.com/tenstorrent/tt-mlir/issues/4499

### Problem description

### What's changed (in pipeline order)
D2M ElementwiseFusion pass fuses 2 D2M.generic ops that:
- are both in compute only form
- do not contain tile_tilieze or tile_untilize ops
- have pure tensor semantics
- neither op has an input operand that is used by more than 1 op
- - this also checks if the input comes from a arg->to_layout->tilize where the arg is shared between multiple to_layouts
- both have compatible blocking
- both have all parallel iterators
- the number of required tiles post-fusion does not exceed capacity (very simple and conservative check for now)

D2M LinAlgElementwiseFusion pass (followed by canonicalization to remove d2m.empty() and tensor.empty())
- pre-canned pass that fuses lining.generics within the previously fused d2m.generic.compute regions

D2M LinAlgLoopFission Pass
- added after insertDstRegisterAccess pass to split the the double nested affine loops at affine.store op boundaries
- duplicates load ops within new loops where required (unused loads are removed by a canonicalization pass afterwards)

### Checklist
- [ ] cosh passing e2e through builder and ttrt with bfloat16 and various grid shapes before regression (see Issue https://github.com/tenstorrent/tt-mlir/issues/5152)
- [ ] enable and fix all remaining tests once regression is fixed


lib/Dialect/D2M/Transforms/SFPUTileLoopFission.cpp
nsmithtt, line 1: nit: Linalg is doesn't have capital A.

include/ttmlir/Dialect/D2M/Transforms/Passes.td
```diff
@@ -7,6 +7,42 @@
...
+    Option<"maxDstRegisterSizeTiles", "max-dst-register-size-tiles", "unsigned", "0", "Override the max dst size tiles or 0 if unset.">,
+  ];
+
+  let dependentDialects = [
+    "::mlir::tt::d2m::D2MDialect",
+    "::mlir::affine::AffineDialect"
+  ];
+}
+
+def D2MLinAlgLoopFission : Pass<"d2m-linalg-loop-fission", "::mlir::ModuleOp"> {
```
nsmithtt: nit: Linalg is doesn't have capital A.

lib/Dialect/TTMetal/Pipelines/TTMetalPipelines.cpp
```diff
@@ -134,6 +142,9 @@ void createTTIRToTTMetalMiddleendPipeline(
   pm.addPass(
       d2m::createD2MInsertDstRegisterAccess(insertDstRegisterAccessOptions));
 
+  pm.addPass(d2m::createD2MLinAlgLoopFission());
```
nsmithtt: Fission needs to run before `InsertDstRegisterAccess`, once the dst access pass runs the loops need to be set in stone, it won't be scalable to run any transforms like this afterwards.  Ideally this pass runs immediately after the fusion passes w/ primitive model of dst (to be replaced with the real dst analysis).

lib/Dialect/D2M/Transforms/LinAlgLoopFission.cpp
```diff
@@ -0,0 +1,282 @@
...
+  // Create a deep copy of the affine loop nest
+  IRMapping map;
+  rewriter.setInsertionPointAfter(outerFor);
+  auto postNest =
+      cast<affine::AffineForOp>(rewriter.clone(*outerFor.getOperation(), map));
+
+  // Erase operations in original loop nest in a bottom up manner until
+  // we reach the store op. Bottom up order guarantees we don't attempt
+  // to erase ops that have results that have users.
+  // >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> //
```
nsmithtt: Nit: we shouldn't use divider comments like this, especially not ones that can be confused for git merge conflict markers.

lib/Dialect/D2M/Transforms/ElementwiseFusion.cpp
```diff
@@ -0,0 +1,585 @@
...
+    if (maxDstRegisterSizeTiles.getValue() > 0) {
+      dstRegisterSizeTiles =
+          std::min(dstRegisterSizeTiles, maxDstRegisterSizeTiles.getValue());
+    }
+
+    MLIRContext *ctx = &getContext();
+    RewritePatternSet patterns(ctx);
+    patterns.add<FuseD2MElementwiseOpsPattern>(ctx, dstRegisterSizeTiles);
+    GreedyRewriteConfig cfg;
+    (void)applyPatternsGreedily(getOperation(), std::move(patterns), cfg);
```
nsmithtt: If you're using the greedy rewrite driver, please handle the return error:
```c++
    if (failed(applyPatternsGreedily(getOperation(), std::move(patterns)))) {
      signalPassFailure();
    }
```

lib/Dialect/D2M/Transforms/ElementwiseFusion.cpp
```diff
@@ -0,0 +1,585 @@
...
+
+      llvm::errs() << "Consumer to be deleted: " << consumer << "\n";
+      llvm::errs() << "Producer to be deleted: " << producer << "\n\n";
+
+      llvm::errs() << "Fused op: " << fused << "\n";
+
+      llvm::errs() << "Consumer users = "
+                   << llvm::range_size(consumer->getUsers()) << "\n";
+      llvm::errs() << "Producer users = "
+                   << llvm::range_size(producer->getUsers()) << "\n";
```
nsmithtt: Stray prints, please remove.

lib/Dialect/D2M/Transforms/ElementwiseFusion.cpp
```diff
@@ -0,0 +1,585 @@
...
+                                OpOperand *use, unsigned dstRegisterSizeTiles) {
+  // Check if the operand has 16-bit or 32-bit element type
+  auto tensorType = dyn_cast<RankedTensorType>(use->get().getType());
+  assert(tensorType); // use MUST be a tensorType unless something got borked
+
+  Type elementType = tensorType.getElementType();
+  auto shape = tensorType.getShape();
+
+  llvm::errs() << "blockshape " << shape[0] << " " << shape[1] << shape[2]
+               << " " << shape[3] << "\n";
```
nsmithtt: Stray prints.

lib/Dialect/D2M/Transforms/ElementwiseFusion.cpp
```diff
@@ -0,0 +1,585 @@
...
+  }
+  return covered.all();
+}
+
+static int countBinaryOps(Operation *op) {
+  int count = 0;
+
+  for (Region &region : op->getRegions()) {
+    for (Operation &opInner : region.getOps()) {
+      llvm::errs() << "After " << opInner << "\n\n";
```
nsmithtt: Stray prints, please remove.

lib/Dialect/D2M/Transforms/ElementwiseFusion.cpp
```diff
@@ -0,0 +1,585 @@
...
+    elementType = tileType.getElementType();
+  }
+
+  // unsigned bitWidth = elementType.getIntOrFloatBitWidth();
+
+  // TODO(mbagherbeikTT) previous 32bit logic is disabled. Need to eventually
+  // add a check for fp32DstAccumulate.
+  // int dstTilesRemaining = dstRegisterSizeTiles / (bitWidth == 32 ? 2 : 1);
+  int dstTilesRemaining = dstRegisterSizeTiles;
+  llvm::errs() << "Tiles left initial" << dstTilesRemaining << "\n\n";
```
nsmithtt: Stray prints, please remove.

lib/Dialect/D2M/Transforms/ElementwiseFusion.cpp
```diff
@@ -0,0 +1,585 @@
...
+
+  dstTilesRemaining -= blockSize * (consumer->getNumOperands() +
+                                    producer->getNumOperands() - 2 - 1);
+
+  llvm::errs() << "Tiles left after input " << dstTilesRemaining << "\n\n";
```
nsmithtt: Stray prints, please remove.

lib/Dialect/D2M/Transforms/ElementwiseFusion.cpp
```diff
@@ -0,0 +1,585 @@
...
+    }
+  }
+  return false;
+}
+
+/// Check if op should be skipped for elementwise fusion.
+///
+/// If op or any of the operations nested within it's regions have
+/// the D2MSkipOpEltWiseFusionTrait, the op should be skipped.
+static bool shouldBeSkipped(Operation *op) {
```
nsmithtt: Let's call this `hasSkipOpEltwiseFusionTrait`.

include/ttmlir/Dialect/D2M/IR/D2MTraits.h
```diff
@@ -32,6 +32,10 @@ struct D2MGenericRegionDatamovementOpTrait
   }
 };
 
+template <typename ConcreteType>
+struct D2MSkipOpEltWiseFusionTrait
```
nsmithtt: Nit: W typically isn't capitalized.

include/ttmlir/Dialect/D2M/IR/D2MBase.td
```diff
@@ -57,4 +57,8 @@ class D2M_Trait<string name, list<Trait> traits = []> : NativeOpTrait<name, trai
 def D2M_GenericRegionComputeOpTrait : D2M_Trait<"D2MGenericRegionComputeOpTrait">;
 def D2M_GenericRegionDatamovementOpTrait : D2M_Trait<"D2MGenericRegionDatamovementOpTrait">;
 
+// SkipOpEltWiseFusion is a trait that can be used to mark specific compute operations
+// to skip the ElementwiseFusion pass
+def D2M_SkipOpEltWiseFusionTrait : D2M_Trait<"D2MSkipOpEltWiseFusionTrait">;
```
nsmithtt: Nit: W typically isnt' capitalized.

lib/Dialect/D2M/Transforms/ElementwiseFusion.cpp
```diff
@@ -0,0 +1,585 @@
...
+    if (llvm::range_size(input->get().getUsers()) > 1) {
+      return true;
+    }
+
+    auto tilizeGenOp = dyn_cast<GenericOp>(input->get().getDefiningOp());
+
+    // Function arguments go through to_layout->tilize for each unique
+    // non-tilize compute region they appear in. Need to check if operands are
+    // coming from a tilize generic and then trace that back to a to_layout op
+    // and see if the input arg is used multiple times
```
nsmithtt: I don't fully understand this case, aren't we already handling tilize/untilize w/ the skip trait?  Why do we need additional special casing here?

lib/Dialect/D2M/Transforms/ElementwiseFusion.cpp
```diff
@@ -0,0 +1,585 @@
...
+  auto tensorType = dyn_cast<RankedTensorType>(use->get().getType());
+  assert(tensorType); // use MUST be a tensorType unless something got borked
+
+  Type elementType = tensorType.getElementType();
+  auto shape = tensorType.getShape();
+
+  llvm::errs() << "blockshape " << shape[0] << " " << shape[1] << shape[2]
+               << " " << shape[3] << "\n";
+
+  int blockSize = shape[3];
```
nsmithtt: The shape is not guaranteed to be rank 4, nor does [3] map to the blockSize.

lib/Dialect/D2M/Transforms/ElementwiseFusion.cpp
```diff
@@ -0,0 +1,585 @@
...
+        }
+      }
+      count += countBinaryOps(&opInner);
+    }
+  }
+  return count;
+};
+
+static bool fitsInDstPostFusion(GenericOp producer, GenericOp consumer,
+                                OpOperand *use, unsigned dstRegisterSizeTiles) {
```
nsmithtt, line 42: This function needs more work.  For now can we do something simpler and just say that all ops need numOperands worth of dst tiles or something?

lib/Dialect/D2M/Transforms/ElementwiseFusion.cpp
```diff
@@ -0,0 +1,585 @@
...
+  for (Attribute it : iteratorTypes) {
+    auto itAttr = cast<mlir::tt::ttcore::IteratorTypeAttr>(it);
+    if (itAttr.getValue() != mlir::tt::ttcore::IteratorType::Parallel) {
+      return false;
+    }
+  }
+  return true;
+}
+
+static bool coversAllDimsAfterFusion(GenericOp producer, GenericOp consumer,
```
nsmithtt: It think this function could use a comment, what does it mean to cover all dims?

lib/Dialect/D2M/Transforms/ElementwiseFusion.cpp
```diff
@@ -0,0 +1,585 @@
...
+      size_t fusedIdx = use->getOperandNumber();
+      for (size_t i = 0; i < fusedIdx; ++i) {
+        fusedInputs.push_back(inputs[i]);
+        fusedMaps.push_back(consumer.getIndexingMap(i));
+      }
+      // producer inputs (remapped)
+      for (OpOperand *pi : producer.getDpsInputOperands()) {
+        fusedInputs.push_back(pi->get());
+        fusedMaps.push_back(
+            computeFusedArgMap(producer, pi, prodResMap, consMap));
```
nsmithtt: Trying to work through why this should be done, e.g.

```mlir
%r0 = generic ins(%i0, %i1) outs(...)
%r1 = generic ins(%i2, %r0) outs(...)
```

Shouldn't this just become?
```mlir
%r0 = generic ins(%i0, %i1, %i2) outs(...)
```
ie. r0 was absorbed, why do we need to compose with the consumer map?

lib/Dialect/D2M/Transforms/ElementwiseFusion.cpp
```diff
@@ -0,0 +1,585 @@
...
+    // Preserve if used by something other than the consumer
+    bool hasOtherUse = llvm::any_of(it.value().getUsers(), [&](Operation *u) {
+      return u != consumer.getOperation();
+    });
+    if (hasOtherUse) {
+      keep.insert(idx);
+      continue;
+    }
+    // In doubt about dropping, conservatively keep for now.
+    // TODO: Refine like linalg's invertibility concat check.
```
nsmithtt: Can you file an issue?

lib/Dialect/D2M/Transforms/ElementwiseFusion.cpp
```diff
@@ -0,0 +1,585 @@
...
+      auto fused = rewriter.create<GenericOp>(
+          consumer.getLoc(), fusedResultTypes, fusedInputs, fusedOutputs,
+          consumer.getGrid(), consumer.getBlockFactors(),
+          rewriter.getAffineMapArrayAttr(fusedMaps),
+          consumer.getIteratorTypes(), consumer.getThreads(), /*regions=*/1);
+
+      // Build fused region: clone producer then consumer, map fused operand to
+      // producer yield value. Ensure fused block argument types match original
+      // region operand argument types (shard types), not the top-level types.
+      IRMapping map;
```
nsmithtt: Can we call this `irMap`? map is pretty overloaded as it's often the name for affine maps.

lib/Dialect/D2M/Transforms/ElementwiseFusion.cpp
```diff
@@ -0,0 +1,585 @@
...
+
+  return false;
+}
+
+static bool hasCompatibleBlocking(GenericOp a, GenericOp b) {
+  return a.getGrid() == b.getGrid() &&
+         a.getBlockFactors() == b.getBlockFactors();
+}
+
+static bool isAllParallel(ArrayAttr iteratorTypes) {
```
nsmithtt: Might be nice to have this be a generic op method.

lib/Dialect/D2M/Transforms/ElementwiseFusion.cpp
```diff
@@ -0,0 +1,585 @@
...
+
+      // consumer inputs before fused
+      auto inputs = consumer.getInputs();
+      size_t fusedIdx = use->getOperandNumber();
+      for (size_t i = 0; i < fusedIdx; ++i) {
+        fusedInputs.push_back(inputs[i]);
+        fusedMaps.push_back(consumer.getIndexingMap(i));
+      }
+      // producer inputs (remapped)
+      for (OpOperand *pi : producer.getDpsInputOperands()) {
```
nsmithtt: This for loop is a bit long, can we break the middle section into a helper func `GenericOp createFusedGeneric(...)`? That does then whole mechanical part of creating the fused generic.

lib/Dialect/D2M/Transforms/ElementwiseFusion.cpp
mbagherbeikTT: there was an issue where if we had a function arg that was an input operand to multiple, lets say two, d2m.generic ops, the arg would go through two separate to_layout->d2m.generic{tilize}, one per use. 

The fusion pass would then, expectedly, treat the outputs of the two tilize ops as separate and it was possible to fuse the two d2m.generic ops. This crashed the allocator so I added this specific way of handling it for now.  

Will definitely need to revisit this but I'm not sure if I should create the issue now or wait until we start spec'ing the next version.

lib/Dialect/TTMetal/Pipelines/TTMetalPipelines.cpp
mbagherbeikTT: I need to think about this one. Looking at the IR after the linalg fusion as well as after computeLoops, I'm not sure how to manipulate that IR to get the effect that we want as it gets lowered through the downstream passes.

I can create the issue and we can tackle it after the demo or is it critical to address this now?

nsmithtt: I'd prefer to address it now, we can have the demo working on a branch.  I have an example piece of manually fused IR that essentially demonstrates this works.
[fused_eltwise.mlir.zip](https://github.com/user-attachments/files/22641481/fused_eltwise.mlir.zip)

If you take a look at the pass just before insert dst pass, you'll notice each pair of linalg nests are independently copy-nested.

`InsertDstRegisterAccess` is meant to be _the_ pass that transforms the IR to legally allocate from dst and manage all of its access.  It feels fragile for that pass to leave the IR in an illegal state and require an additional pass to further transform the IR to make it legal.

lib/Dialect/D2M/Transforms/ElementwiseFusion.cpp
nsmithtt: > The fusion pass would then, expectedly, treat the outputs of the two tilize ops as separate and it was possible to fuse the two d2m.generic ops

Fuse which 2 generic ops?  Can you provide some pseudo code demonstrating this situation?  I agree that it's not ideal that a single func arg gets uploaded twice, but it's not illegal either.

> This crashed the allocator so I added this specific way of handling it for now.

We should debug and discover why and ideally fix the allocator.

> Will definitely need to revisit this but I'm not sure if I should create the issue now or wait until we start spec'ing the next version.

We should definitely file an issue or raise it with the group on the channel to help diagnose.  Adding a workaround here for a different pass is a last resort fix.

mbagherbeikTT: I tried recreating the issue but it's not popping up anymore after the rebase. Will remove it for testing unless the bug pops up again and I'll create a detailed issue if it does

lib/Dialect/D2M/Transforms/ElementwiseFusion.cpp
mbagherbeikTT: I added comments about what the function does but I also essentially commented out the body for now. 

This was a slightly reformatted variant of how the linalg fusion pass checks that all dimensions are touched by at least one input when we have reduction iterators. Since we only operate on ops that have all parallel iterators, we never reach that part of the code. 

I'll add an issue to re-enable it later if we still need it at that point.

lib/Dialect/D2M/Transforms/ElementwiseFusion.cpp
mbagherbeikTT: would shape.back() be correct or am I completely out in the weeds?

lib/Dialect/D2M/Transforms/ElementwiseFusion.cpp
mbagherbeikTT: with the way insertDst currently works, I think what's there now after some clean-up might be the bare minimum. Based on my understanding and how tests have gone:
- every input operand needs a tile
- every binary op needs it's own output tile, either to store the intermediate or the final output of the fused block; unary ops do everything in place
- blockSize needs to be accounted for as well

lib/Dialect/D2M/Transforms/ElementwiseFusion.cpp
mbagherbeikTT: I'm not sure I'm following. In this case r0 (in the first snippet) is one of the consumer's DPS Inputs. The loop you highlighted composes the producer's DPS Inputs (i0, i1) to the consumer map. r0 isn't touched by the loops surrounding the highlighted code

lib/Dialect/D2M/Transforms/ElementwiseFusion.cpp
mbagherbeikTT: removed since it was dead code, per our discussion

lib/Dialect/D2M/Transforms/ElementwiseFusion.cpp
mbagherbeikTT: dead code, removed

lib/Dialect/D2M/IR/D2MOps.cpp
```diff
@@ -1492,4 +1497,84 @@ d2m::GenericOp::getBufferType(mlir::Value value,
...
+        if (helper(&opInner)) {
+          return true;
+        }
+      }
+    }
+
+    return false;
+  };
+
+  return helper(this->getOperation());
```
nsmithtt: Maybe it's more succinct/performant to write:
```c++
bool skipFusion = false;
walk([&](Operation* op) {
  skipFusion |= op->hasTrait<D2MSkipOpEltwiseFusionTrait>();
  return skipFusion ? WalkResult::interrupt() : WalkResult::advance();
});
return skipFusion;
```

lib/Dialect/D2M/IR/D2MOps.cpp
```diff
@@ -1492,4 +1497,84 @@ d2m::GenericOp::getBufferType(mlir::Value value,
...
+bool d2m::GenericOp::isAllParallel() {
+  ArrayAttr iteratorTypes = getIteratorTypes();
+
+  for (Attribute it : iteratorTypes) {
+    auto itAttr = mlir::cast<mlir::tt::ttcore::IteratorTypeAttr>(it);
+    if (itAttr.getValue() != mlir::tt::ttcore::IteratorType::Parallel) {
+      return false;
+    }
+  }
+  return true;
```
nsmithtt: Nit: could use `llvm::all_of`

lib/Dialect/D2M/IR/D2MOps.cpp
```diff
@@ -1492,4 +1497,84 @@ d2m::GenericOp::getBufferType(mlir::Value value,
...
+}
+
+bool d2m::GenericOp::isValidElementwiseFusionTarget() {
```
nsmithtt: I think this might be better as a static method in the pass, reasoning is that this logic is directly tied to a specific pass and not intrinsic to the generic op itself.

lib/Dialect/D2M/Transforms/ElementwiseFusion.cpp
```diff
@@ -0,0 +1,389 @@
...
+#define GEN_PASS_DEF_D2MELEMENTWISEFUSION
+#include "ttmlir/Dialect/D2M/Transforms/Passes.h.inc"
+
+static int countBinaryOps(Operation *op) {
+  int count = 0;
+
+  for (Region &region : op->getRegions()) {
+    for (Operation &opInner : region.getOps()) {
+      if (opInner.hasTrait<D2MGenericRegionComputeOpTrait>()) {
+        if (opInner.getNumOperands() > 1) {
```
nsmithtt: Should it not be NumOperands == 2?

lib/Dialect/D2M/Transforms/ElementwiseFusion.cpp
```diff
@@ -0,0 +1,389 @@
...
+      count += countBinaryOps(&opInner);
+    }
+  }
+  return count;
+};
+
+static bool fitsInDstPostFusion(GenericOp producer, GenericOp consumer,
+                                OpOperand *use, unsigned dstRegisterSizeTiles) {
+  auto tensorType = dyn_cast<RankedTensorType>(use->get().getType());
+  assert(tensorType); // use MUST be a tensorType unless something got borked
```
nsmithtt: Use `mlir::cast` which will assert for you (instead of dyn_cast).

lib/Dialect/D2M/Transforms/ElementwiseFusion.cpp
```diff
@@ -0,0 +1,389 @@
...
+
+  auto shape = tensorType.getShape();
+  int blockSize = shape.back();
```
nsmithtt, line 46: Shouldn't this be multiples of the full shard volume?  Conservatively we should just pick `1` here imo and let the loop tiling pass derive the biggest legal size afterwards.  Ofc this will be cleaner once we have the analysis.

lib/Dialect/D2M/Transforms/ElementwiseFusion.cpp
```diff
@@ -0,0 +1,389 @@
...
+
+  if (blockSize > static_cast<int>(dstRegisterSizeTiles)) {
+    blockSize = dstRegisterSizeTiles;
+  }
+
+  int dstTilesRemaining = dstRegisterSizeTiles;
+
+  // Account for number of tiles needed to store input operands after fusion
+  dstTilesRemaining -= blockSize * (consumer->getNumOperands() +
+                                    producer->getNumOperands() - 2 - 1);
```
nsmithtt: Why `- 2 - 1`? NumOperands also returns an unsigned number so this expression can overflow.

lib/Dialect/D2M/Transforms/ElementwiseFusion.cpp
```diff
@@ -0,0 +1,389 @@
...
+
+  if (dstTilesRemaining < 0) {
+    return false;
+  }
+
+  // Subtract # of intermediate tiles needed
+  dstTilesRemaining -=
+      blockSize * (countBinaryOps(consumer) + countBinaryOps(producer));
```
nsmithtt, line 73: Can't we reclaim the register space if we're disallowing fork/joins?  This seems overly conservative.

lib/Dialect/D2M/Transforms/ElementwiseFusion.cpp
```diff
@@ -0,0 +1,389 @@
...
+      assert(producer);
+
+      // Build fused op operands, maps, results
+      SmallVector<Value> fusedInputs;
+      SmallVector<Value> fusedOutputs;
+      SmallVector<Type> fusedResultTypes;
+      SmallVector<AffineMap> fusedMaps;
+
+      gatherAndComposeFusedOpIO(use, producer, consumer, fusedInputs,
+                                fusedOutputs, fusedResultTypes, fusedMaps);
```
nsmithtt: Can this return a tuple instead of using out params?  Also `fusedResultTypes` is redundant since we could use `TypeRange(fusedOutputs)`.  Perhaps a more descriptive name is `getFusedOperands`?

lib/Dialect/D2M/Transforms/ElementwiseFusion.cpp
```diff
@@ -0,0 +1,389 @@
...
+
+      IRMapping irMap;
+      Block &fusedBlock = fused.getRegion(0).emplaceBlock();
+      Block &pb = producer.getRegion(0).front();
+      Block &cb = consumer.getRegion(0).front();
+
+      mapFusedOpIo(fused, use, producer, consumer, irMap, fusedBlock, pb, cb);
+
+      populateFusedOpBody(fused, use, producer, consumer, irMap, fusedBlock, pb,
+                          cb, rewriter);
```
nsmithtt: I think from line 330-334 should be merged into a single function w/ `mapFusedOpIo` and `populateFusedOpBody` inlined. Called like `createFusedGeneric` or something, reasoning is that these helpers feel kind of arbitrarily broken up and having the pass multiple intermediate arguments between them.

lib/Dialect/D2M/Transforms/ElementwiseFusion.cpp
```diff
@@ -0,0 +1,389 @@
...
+    if (auto y = dyn_cast<YieldOp>(op)) {
+      consYield = y;
+    }
+  }
+  for (Value y : consYield.getOperands()) {
+    fusedYields.push_back(irMap.lookupOrDefault(y));
+  }
+  rewriter.setInsertionPointToEnd(&fusedBlock);
+  rewriter.create<YieldOp>(fused.getLoc(), fusedYields);
+
```
nsmithtt, line 341: We should return here, if we go with the `createFusedGeneric` op this is where we'd return the newly created op.

lib/Dialect/D2M/Transforms/ElementwiseFusion.cpp
```diff
@@ -0,0 +1,389 @@
...
+  // Replace uses: from producer and consumer results to fused results.
+  int resIdx = 0;
+  // Consumer results
+  for (auto r : consumer->getResults()) {
+    r.replaceAllUsesWith(fused->getResult(resIdx++));
+  }
+
+  rewriter.eraseOp(consumer);
+  rewriter.eraseOp(producer);
```
nsmithtt: From line 284 - 292 should happen in the callers stack frame.

lib/Dialect/D2M/Transforms/ElementwiseFusion.cpp
nsmithtt: High level, if we wrote pseudo code to outline the steps here, it seems like there's 3:
1. Check to see if we can fuse.
2. Create a new generic op that is fused.
3. Replace the existing generics with a fused one.

lib/Dialect/D2M/Transforms/ElementwiseFusion.cpp
```diff
@@ -0,0 +1,389 @@
...
+      Block &fusedBlock = fused.getRegion(0).emplaceBlock();
+      Block &pb = producer.getRegion(0).front();
+      Block &cb = consumer.getRegion(0).front();
+
+      mapFusedOpIo(fused, use, producer, consumer, irMap, fusedBlock, pb, cb);
+
+      populateFusedOpBody(fused, use, producer, consumer, irMap, fusedBlock, pb,
+                          cb, rewriter);
+
+      return success();
```
nsmithtt: It's a little unconventional to return from inside of a for loop.  Perhaps this loop could be reformulated as:
```c++
OpOperand *use = findFusableProducer(consumer.getDpsInputOperands(), ...);
if (!use) {
  return failure();
}
```

It's easier to read because after this point the reader can now assume that the rest of this body is only concerned with mechanically doing the fusing.

lib/Dialect/TTMetal/Pipelines/TTMetalPipelines.cpp
```diff
@@ -134,6 +142,9 @@ void createTTIRToTTMetalMiddleendPipeline(
   pm.addPass(
       d2m::createD2MInsertDstRegisterAccess(insertDstRegisterAccessOptions));
 
+  pm.addPass(d2m::createD2MLinalgLoopFission());
```
nsmithtt: I'd still really prefer to do fission before InsertDstRegisterAccess.

lib/Dialect/D2M/Transforms/ElementwiseFusion.cpp
mbagherbeikTT: forks aren't supported but joins are. There's also cases like 3 inputs into 2 adds: each add requires its own output tile. Most of the binary ops need an output tile except maximum which operates in place on one of the inputs (not special casing for it right now)

lib/Dialect/D2M/IR/D2MOps.cpp
mbagherbeikTT: that is much more elegant, ty!

lib/Dialect/D2M/Transforms/ElementwiseFusion.cpp
mbagherbeikTT: will fix the unsigned issue.

-2 is for removing the input operand on the consumer and the init/output operand on the producer

-1 is removing the consumer's init/output operand --> will be added back if op is binary because it needs an output tile; if it's unary --> in place, no extra tile

lib/Dialect/D2M/Transforms/ElementwiseFusion.cpp
mbagherbeikTT: if it's set to 1, it over-fuses and fails PCC. I changed it to be a product of the last 2 tensor shape dims which should be consistent with how the loops are set up.

lib/Dialect/TTMetal/Pipelines/TTMetalPipelines.cpp
mbagherbeikTT: let me know how you want to proceed now that we cleared up the confusion. the pass has been renamed to SFPUTileLoopFission

xanderchin (approved):

test/python/golden/test_ttir_eltwise_fusion.py
```diff
@@ -0,0 +1,939 @@
...
+        return builder.rsqrt
+    if op_name == "sqrt":
+        return builder.sqrt
+    if op_name == "sigmoid":
+        return builder.sigmoid
+    if op_name == "sin":
+        return builder.sin
+    if op_name == "tan":
+        return builder.tan
+    if op_name == "eqz":
```
dlokeTT: TTIR only has the general versions of these... eq, ne and so on. We lower the generalized versions to a subtract then compare to zero in because that's how it's represented in metal/llk. Just removing the z at the end of all these will work

dlokeTT: wait nvm, those are binary. probably should just get rid of them

mbagherbeikTT: done! thanks :)

PR #5154, opened 2025-09-30
Add support for power(tensor, scalar) op

### LINKED ISSUE #5180: Add support for pow_scalar op
TTNN API supports pow(tensor, tensor) and pow(tensor, scalar) ops. However we only support pow(tensor, tensor) op which requires exponent in tensor form even if the exponent is a scalar value; for example torch.pow(input, 2) will generate a graph like. 

```
main (%arg0: tensor<32768x2560xf32>) -> <32768x2560xf32> {
  %cst = stablehlo.constant dense<2.000000e+00> : tensor<32768x2560xf32> 
  %0 = stablehlo.power %arg0, %cst : tensor<32768x2560xf32>
  return %0 : <32768x2560xf32>
}
```
Now `%cst` is created with `ttnn.full` op and exists till last power op is executed and utilizing memory.
Qwen3-4B model failed to run large token size (greater than 15K) due to `out of memory` error. using scalar version gives us memory back. We successfully executed Qwen3-4B model for 32K token inputs by avoiding creating these constant tensors and saved ~1.3GB memory. 

Objective:
Add support for pow(tensor, scalar) op.

mmanzoorTT:
### Ticket
closes #5180 

### Problem description
TTNN API supports pow(tensor, tensor) and pow(tensor, scalar) ops. 
However we only support pow(tensor, tensor) op which requires
exponent in tensor form even if the exponent is a scalar value.

### What's changed
Add support for pow(tensor, scalar) op end-to-end.

### Checklist

### DIFF
--- a/include/ttmlir/Target/TTNN/operations/eltwise.fbs
+++ b/include/ttmlir/Target/TTNN/operations/eltwise.fbs
@@ -54,10 +54,6 @@ table EltwiseBinaryCompositeOp {
   out: tt.target.ttnn.TensorRef;
 }
 
-table UIntegralType {
-  value: uint32;
-}
-
 union RhsParams {
   FP: FloatingPointType,
   UI32: UIntegralType,
--- a/runtime/lib/ttnn/operations/eltwise/binary/binary_composite.cpp
+++ b/runtime/lib/ttnn/operations/eltwise/binary/binary_composite.cpp
@@ -62,8 +62,8 @@ static void runEltwiseBinaryCompositeOp(
 }
 
 static void
-runPowOp(const ::tt::target::ttnn::EltwiseBinaryCompositeScalarOp *op,
-         auto &&exponent, ProgramContext &context) {
+runPowScalarOp(const ::tt::target::ttnn::EltwiseBinaryCompositeScalarOp *op,
+               auto &&exponent, ProgramContext &context) {
   ProgramTensorPool &tensorPool = context.getTensorPool();
   ::ttnn::Tensor *input = &(tensorPool.getTTNNTensorAndValidate(op->lhs()));
 
@@ -134,10 +134,10 @@ void run(const ::tt::target::ttnn::EltwiseBinaryCompositeScalarOp *op,
   case ::tt::target::ttnn::EltwiseBinaryCompositeScalarOpType::PowScalar: {
     switch (op->rhs_type()) {
     case ::tt::target::ttnn::RhsParams::FP:
-      runPowOp(op, op->rhs_as_FP()->value(), context);
+      runPowScalarOp(op, op->rhs_as_FP()->value(), context);
       break;
     case ::tt::target::ttnn::RhsParams::UI32:
-      runPowOp(op, op->rhs_as_UI32()->value(), context);
+      runPowScalarOp(op, op->rhs_as_UI32()->value(), context);
       break;
     default:
       LOG_FATAL("unknown exponent type");
--- a/include/ttmlir/Target/TTNN/types.fbs
+++ b/include/ttmlir/Target/TTNN/types.fbs
@@ -87,3 +87,7 @@ table TensorRef {
 table FloatingPointType {
   value: float;
 }
+
+table UIntegralType {
+  value: uint32;
+}
--- a/test/ttnn-jit/test_eltwise.py
+++ b/test/ttnn-jit/test_eltwise.py
@@ -247,7 +247,7 @@ def remainder(a, b):
 
 
 def pow(a, b):
-    return ttnn.pow(a, b)
+    return ttnn.pow_tensor(a, b)
 
 
 def atan2(a, b):


include/ttmlir/Dialect/TTNN/IR/TTNNOps.td
```diff
@@ -746,6 +746,31 @@ def TTNN_PowOp : TTNN_ElementwiseBinaryCompositeOp<"pow"> {
     }];
 }
 
+def TTNN_PowScalarOp : TTNN_Op<"pow_scalar"> {
```
sgholamiTT: Is it possible to define a `TTNN_ElementwiseScalarBinaryOp` class that this op inherits from ?
Although `TTNN_PowScalarOp` might be the only op that implements it, this will make our life much easier in the future to add support for other ops (`TTNN_AddScalarOp`, `TTNN_MulScalarOp`, etc.). I know @vtangTT is currently looking into some workarounds for supporting binary scalar in `ttnn-jit`, though ultimately we should be able to support this end to end (then `ttnn-jit` could continue to remain a thin wrapper around our backend). Did you create the 'north start' issue @vtangTT ? :))

For instance:
```
class TTNN_ElementwiseScalarBinaryOp<string mnemonic, list<Trait> traits = []> :
    TTNN_Op<mnemonic, [TTNN_MemoryConfigOpInterface] # traits> {
    let summary = "Eltwise scalar binary op.";
    let description = [{
      Eltwise scalar binary op.
    }];

    let arguments = (ins AnyRankedTensor:$lhs,
                         AnyAttrOf<[F32Attr, I32Attr]>:$rhs,
                         OptionalAttr<TTNN_MemoryConfigAttr>:$memory_config);
    let results = (outs AnyRankedTensor:$result);

      // workarounds ?
    }];

    let builders =
    [
      // builders ?
    ];
}
```

and then:
```
def TTNN_PowTensorOp : TTNN_ElementwiseScalarBinaryOp<"pow_scalar"> {
  let summary = "Eltwise power OP.";
  let description = [{
    Performs element-wise exponentiation of lhs tensor by rhs tensor and produces a
    result tensor. Tensors must be of same shape.

    Example:
    ```
      %result = "ttnn.pow"(%lhs, %rhs) : (tensor<6xf64>, tensor<6xf64>) -> tensor<6xf64>

      %lhs: [-2.0, -0.0, -36.0, 5.0, 3.0, 10000.0]
      %rhs: [2.0, 2.0, 1.1, 2.0, -1.0, 10.0]
      %result: [4.0, 0.0, -nan, 25.0, 0.333333343, inf]
    ```
  }];
}
```

lib/OpModel/TTNN/TTNNOpModel.cpp
```diff
@@ -427,7 +429,9 @@ auto getOpSymbol() {
     return ::ttnn::remainder;
   } else if constexpr (std::is_same_v<OpTy, Atan2Op>) {
     return ::ttnn::atan2;
-  } else if constexpr (std::is_same_v<OpTy, PowOp>) {
+  } else if constexpr (std::is_same_v<OpTy, PowTensorOp>) {
+    return ::ttnn::pow;
+  } else if constexpr (std::is_same_v<OpTy, PowScalarOp>) {
```
sgholamiTT: No need to add this.

include/ttmlir/Dialect/TTIR/IR/TTIROps.td
```diff
@@ -1718,6 +1718,51 @@ def TTIR_PowOp : TTIR_ElementwiseBinaryOp<"pow"> {
...
+    // [8, 9, 5]
+    ```
+
+    Special cases:
+    - 0^0 is typically defined as 1
+    - For integer types, negative bases with non-integer exponents may result in complex numbers,
+      which are typically not supported and may result in undefined behavior
+
+    Mathematical definition: pow(x, y) = x^y
+  }];
```
nsmithtt: Description needs to be updated to reflect the new scalar semantics.

test/unittests/OpModel/TTNN/Lib/TestOpModelLib.cpp
```diff
@@ -1820,6 +1820,71 @@ INSTANTIATE_TEST_SUITE_P(
     Atan2Tests, OpModelAtan2Param,
     generateBinaryEltwiseParamsSameLayout(binaryEltwiseParams));
 
+class OpModelPowScalarParam
+    : public OpModelTest,
+      public testing::WithParamInterface<
+          std::tuple<detail::TestTensor,    // input
+                     detail::TestTensor,    // output
+                     float,                 // exponent
```
sgholamiTT: Maybe we can test with both `int` and `float`.
You can change this to `std::variant<float, int>` to test both.

test/unittests/OpModel/TTNN/Op/TestOpModelInterface.cpp
```diff
@@ -15,6 +15,7 @@
 #include "llvm/ADT/SmallVector.h"
 
 #include <cstdint>
+#include <iostream>
```
sgholamiTT: Please remove.

test/unittests/OpModel/TTNN/Op/TestOpModelInterface.cpp
```diff
@@ -597,6 +598,42 @@ INSTANTIATE_TEST_SUITE_P(
...
+  auto constraintsExp = getOpConstraints(powScalarOp.getOperation());
+  if (!constraintsExp) {
+    FAIL() << "Missing L1 constraints; Error="
+           << llvm::toString(constraintsExp.takeError()) << std::endl;
+  }
+  const auto &[cbSize, l1PeakSize, totalPeakSize, outputSize, outputLayout] =
+      constraintsExp.get();
+  EXPECT_EQ(cbSize, 8192);
+  EXPECT_EQ(l1PeakSize, 2048);
+  EXPECT_EQ(outputSize, 2048);
```
sgholamiTT, line 632: Please assert the value of `totalPeakSize` too.

include/ttmlir/Dialect/TTIR/IR/TTIROps.td
```diff
@@ -1683,7 +1683,7 @@ def TTIR_RemainderOp : TTIR_ElementwiseBinaryOp<"remainder"> {
 
 // TODO We need to add a broadcastable trait to this op.
 // See issue https://github.com/tenstorrent/tt-mlir/issues/2015
-def TTIR_PowOp : TTIR_ElementwiseBinaryOp<"pow"> {
+def TTIR_PowTensorOp : TTIR_ElementwiseBinaryOp<"pow_tensor"> {
```
nsmithtt: Arguably we could have a single PowOp, where the scalar case is:
```mlir
%scalar = ttir.constant {DenseElements{5.f}}
ttir.pow(%lhs, %scalar, %output)
```

We can special case this pattern in lowering to TTNN_PowScalarOp.

include/ttmlir/Dialect/TTIR/IR/TTIROps.td
mmanzoorTT: My bad. I miss pushing description changes.

include/ttmlir/Dialect/TTIR/IR/TTIROps.td
mmanzoorTT: yeah, I agree, it will be better to have single pow op in TTIR and scalar/tensor versions in TTNN.

include/ttmlir/Dialect/TTNN/IR/TTNNOps.td
anuragsingh-tt: Why is there a need for two separate ops in TTNN if this branching can be handled in the lowering from TTNNToFlatbuffer? This class is a preferable approach as there's at least some reusable infra. Why do we need a separate `EltwiseBinaryCompositeParamsOp` in eltwise.fbs?

lib/OpModel/TTNN/TTNNOpModel.cpp
```diff
@@ -1230,6 +1230,97 @@ template struct BinaryCompositeOpModel<LogicalLeftShiftOp>;
...
+  auto powScalarQuery = [=](auto convertedExponent) {
+    return [=]() {
+      return ::ttnn::graph::query_op_runtime(
+          ::ttnn::pow, device, inputSpec, convertedExponent,
+          detail::getNullableMemoryConfig(outputLayout));
+    };
+  };
+
+  // The invoke function of PowScalarOp is templated over the exponent value
+  // type. That's why the following code is aranged in this way.
```
anuragsingh-tt: NIT: arranged.

lib/Conversion/TTNNToEmitC/TTNNToEmitC.cpp
```diff
@@ -496,6 +512,61 @@ class EltwiseBinaryNGCompositeOpConversionPattern
...
+
+private:
+  std::string getPrefixSearchPattern() const override {
+    return "ttnn.pow_scalar";
+  }
+
+  std::string getPrefixSwapPattern() const override { return "ttnn::pow"; }
+
+  template <typename ExponentT,
+            typename = std::void_t<llvm::is_one_of<ExponentT, float, uint32_t>>>
```
anuragsingh-tt: Could we assert using llvm::is_one_of in matchAndRewriteImpl instead of SFINAE in the template parameter list?

include/ttmlir/Dialect/TTIR/IR/TTIROps.td
mmanzoorTT: updated as suggested.

lib/OpModel/TTNN/TTNNOpModel.cpp
mmanzoorTT: Thanks. removed.

test/unittests/OpModel/TTNN/Lib/TestOpModelLib.cpp
mmanzoorTT: Added test for both float and int.

test/unittests/OpModel/TTNN/Op/TestOpModelInterface.cpp
mmanzoorTT: Removed.

test/unittests/OpModel/TTNN/Op/TestOpModelInterface.cpp
mmanzoorTT: Added as suggested.

lib/OpModel/TTNN/TTNNOpModel.cpp
mmanzoorTT: Thanks for pointing it out. fixed.

lib/Conversion/TTNNToEmitC/TTNNToEmitC.cpp
mmanzoorTT: Updated as suggested. thanks for pointing it out.

include/ttmlir/Dialect/TTNN/IR/TTNNOps.td
mmanzoorTT: @sgholamiTT `PowScalarOp` is binary composite op; so I have created `TTNN_ElementwiseBinaryCompositeScalarOp` to keep it aligned with TTNN API and similar to `TTNN_ElementwiseBinaryCompositeOp` which handles the binary composite op for tensor inputs. 

@anuragsingh-tt 
All decompositions and transformations are currently handled within the TTNN dialect. Moving this kind of branched lowering to the TTNNToFlatbuffer stage would introduce unnecessary complexity and inconsistency in the lowering pipeline. Additionally, the TTNN dialect is designed to support multiple backend targets. If we shift this logic to the Flatbuffer lowering, we would need to duplicate the same handling for other targets as well, leading to potential code duplication and increased maintenance overhead in the future.

PowScalarOp (and other similar ops) have different structure than other binary ops. PowScalarOp accepts two operands and perform element-wise operation but it has a scalar value as rhs instead of tensor. In MLIR, a single operand cannot simultaneously be used as both a tensor and a scalar. that's why we need two different classes to define binary_op(tensor, tensor) and binary_op(tensor, scalar) ops. 
Similarly, we need two different definitions for flatbuffer (defined in eltwise.fbs as `EltwiseBinaryCompositeOp` and `EltwiseBinaryCompositeScalarOp`) as runtime uses these definitions in `getOpInputRefs` to create runtime tensors (both type of ops need different number of input tensors).

include/ttmlir/Target/TTNN/operations/eltwise.fbs
```diff
@@ -54,6 +54,27 @@ table EltwiseBinaryCompositeOp {
   out: tt.target.ttnn.TensorRef;
 }
 
+table UIntegralType {
+  value: uint32;
+}
+
+union rhsParams {
```
azecevic000: Nit: use PascalCase

lib/Conversion/TTIRToTTNN/TTIRToTTNN.cpp
```diff
@@ -535,6 +535,52 @@ class ClampOpConversionPattern : public OpConversionPattern<TTIROpTy> {
...
 } // namespace
 
+namespace {
+class PowOpConversionPattern : public OpConversionPattern<ttir::PowOp> {
+private:
+  // Helper function to extract constant value.
+  static mlir::Attribute getConstantAttr(mlir::Value value) {
+    mlir::Operation *op = value.getDefiningOp();
+    while (mlir::isa_and_present<mlir::tt::ttnn::ReshapeOp,
+                                 mlir::tt::ttnn::TypecastOp>(op)) {
```
azecevic000, line 545: Is it safe to just ignore `typecast`? I'm not sure about the semantics of FP->integer conversion (or if it does work at all), but if it does have a truncation semantics it can result in an error.

lib/Conversion/TTIRToTTNN/TTIRToTTNN.cpp
```diff
@@ -535,6 +535,52 @@ class ClampOpConversionPattern : public OpConversionPattern<TTIROpTy> {
...
+    }
+
+    auto fullOp = mlir::dyn_cast_if_present<mlir::tt::ttnn::FullOp>(op);
+    if (!fullOp) {
+      return {};
+    }
+
+    mlir::Attribute fillValueAttr = fullOp.getFillValueAttr();
+
+    if (!isa<FloatAttr>(fillValueAttr) && !isa<IntegerAttr>(fillValueAttr)) {
```
azecevic000: Nit: `!isa<FloatAttr, IntegerAttr>(fillValueAttr)`

lib/Conversion/TTNNToEmitC/TTNNToEmitC.cpp
azecevic000, line 1: Can you please do the same for TTNNToEmitPy?

include/ttmlir/Dialect/TTNN/IR/TTNNOps.td
sgholamiTT: I agree with most of what you shared. Just a small question, you mentioned that:
> In MLIR, a single operand cannot simultaneously be used as both a tensor and a scalar. that's why we need two different classes to define binary_op(tensor, tensor) and binary_op(tensor, scalar) ops.

Do you mean something like this is not possible ?
```
def TensorOrScalar: AnyTypeOf<[AnyRankedTensor, AnyAttrOf<[F32Attr, I32Attr]>]>;
```
And then changing `TTNN_ElementwiseBinaryOp` (okay, or `TTNN_ElementwiseBinaryCompositeOp`, doesn't matter) to accept a `TensorOrScalar` as its rhs ?

mmanzoorTT: MLIR enforces strong typing, so the type of an operand (whether it's a scalar or a tensor) must be unambiguous. For example:
- A scalar is represented as an MLIR Attribute (like FloatAttr, IntegerAttr) or a Value with type f32, i32, etc.
- A tensor is represented as a Value of type tensor<...>, even if it's tensor<> with a single element (e.g., tensor<1xf32>).

I had previously looked into unifying tensor/scalar as one operand but couldn't find a way. I had tried `AnyTypeOf<[F32Attr, AnyRankedTensor]` but it caused build failure.

include/ttmlir/Target/TTNN/operations/eltwise.fbs
mmanzoorTT: Thanks. fixed.

lib/Conversion/TTIRToTTNN/TTIRToTTNN.cpp
mmanzoorTT: updated as suggested.

lib/Conversion/TTNNToEmitC/TTNNToEmitC.cpp
mmanzoorTT: Added the conversion.

include/ttmlir/Target/TTNN/operations/eltwise.fbs
```diff
@@ -54,6 +54,27 @@ table EltwiseBinaryCompositeOp {
...
+union RhsParams {
+  FP: FloatingPointType,
+  UI32: UIntegralType,
+}
+
+enum EltwiseBinaryCompositeScalarOpType: uint32 {
+  PowScalar,
+}
+
+table EltwiseBinaryCompositeScalarOp {
```
jnie-TT, line 66: I don't think we need a separate type for this, could we add something like `EltwiseCompositeParams` and add that to `EltwiseBinaryCompositeOp`? For ops that don't have extra params we can just null the params. As an example please see `EltwiseUnaryOp`.

runtime/lib/ttnn/operations/eltwise/binary/binary_composite_scalar.h
```diff
@@ -0,0 +1,19 @@
+// SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#ifndef RUNTIME_LIB_TTNN_OPERATIONS_ELTWISE_BINARY_COMPOSITE_SCALAR_H
+#define RUNTIME_LIB_TTNN_OPERATIONS_ELTWISE_BINARY_COMPOSITE_SCALAR_H
```
jnie-TT: I think once we unify the params with `EltwiseBinaryCompositeOp` in the flatbuffer schema, we can remove this file and then handle it directly in `runtime/lib/ttnn/operations/eltwise/binary/binary_composite.cpp`

include/ttmlir/Target/TTNN/operations/eltwise.fbs
mmanzoorTT: All Unary ops have single input tensor with additional params. This is bit different scenario. These params are used instead of RHS tensor (no extra params in addition to LHS and RHS tensors) . So we have LHS as tensor and RHS as scalar param. If we add these scalar value as extra params then 1) we will have to make RHS as NULL tensors and params as NULL for ops which don't need them. 2) more importantly, this will deviate the structure from TTNN API (RHS tensor which is not part of TTNN API).
TTNN signature as reference
```
struct ExecutePower {
    static Tensor invoke(
        const Tensor& input_tensor,
        uint32_t exponent,
        const std::optional<MemoryConfig>& memory_config = std::nullopt,
        const std::optional<Tensor>& optional_output_tensor = std::nullopt);

    static Tensor invoke(
        const Tensor& input_tensor,
        float exponent,
        const std::optional<MemoryConfig>& memory_config = std::nullopt,
        const std::optional<Tensor>& optional_output_tensor = std::nullopt);

    static Tensor invoke(
        float input_a,
        const Tensor& exponent,
        const std::optional<const DataType>& dtype = std::nullopt,
        const std::optional<ttnn::MemoryConfig>& memory_config = std::nullopt,
        const std::optional<Tensor>& optional_output_tensor = std::nullopt,
        tt::stl::Span<const unary::EltwiseUnaryWithParam> post_activations = {},
        tt::stl::Span<const unary::EltwiseUnaryWithParam> lhs_activations = {},
        tt::stl::Span<const unary::EltwiseUnaryWithParam> rhs_activations = {},
        std::optional<bool> use_legacy = std::nullopt);
```

lib/Conversion/TTIRToTTNN/TTIRToTTNN.cpp
mmanzoorTT: Previously, we were using `torch-mlir` to generate stablehlo graph from PyTorch model. `torch-mlir` creates a 64-bit scalar (either float or integer depending on the op), downcast the scalar then reshape and broadcast to match the op's shape and type. Now we are using TorchXLA which generates constants with the matched shape and type and doesn't require such modifications (typecast, reshape, broadcast).
Secondly, I am extracting and using the ConstantOp Value as mlir::Attribute directly. This will ensure that the actual scalar value is being used by the op.

runtime/lib/ttnn/operations/eltwise/binary/binary_composite_scalar.h
mmanzoorTT: Discussed it with @jnie-TT offline and it is decided to keep separate type `EltwiseBinaryCompositeScalarOp` in `eltwise.fbs`. However, move the runtime execution to same file `binary_composite.cpp`. All binary composite ops have execution logic at one place.

runtime/lib/ttnn/operations/eltwise/binary/binary_composite.cpp
```diff
@@ -61,6 +61,25 @@ static void runEltwiseBinaryCompositeOp(
   tensorPool.insertTTNNTensorAndValidate(op->out(), out);
 }
 
+static void
+runPowOp(const ::tt::target::ttnn::EltwiseBinaryCompositeScalarOp *op,
+         auto &&exponent, ProgramContext &context) {
```
jnie-TT: Nit: Could we rename to something like `runPowScalar` or something since we have another path that runs `::ttnn::pow` for the tensor input version

runtime/lib/ttnn/operations/eltwise/binary/binary_composite.cpp
```diff
@@ -108,4 +127,24 @@ void run(const ::tt::target::ttnn::EltwiseBinaryCompositeOp *op,
...
+    case ::tt::target::ttnn::RhsParams::FP:
+      runPowOp(op, op->rhs_as_FP()->value(), context);
+      break;
+    case ::tt::target::ttnn::RhsParams::UI32:
+      runPowOp(op, op->rhs_as_UI32()->value(), context);
+      break;
+    default:
+      LOG_FATAL("unknown exponent type");
+    }
+    break;
```
jnie-TT, line 145: Nit: Redundant break

include/ttmlir/Target/TTNN/operations/eltwise.fbs
```diff
@@ -54,6 +54,27 @@ table EltwiseBinaryCompositeOp {
   out: tt.target.ttnn.TensorRef;
 }
 
+table UIntegralType {
```
jnie-TT: Nit: Let's add this to `types.fbs` since floating point type is there as well

runtime/lib/ttnn/operations/eltwise/binary/binary_composite.cpp
mmanzoorTT: updated.

include/ttmlir/Target/TTNN/operations/eltwise.fbs
mmanzoorTT: Moved to `types.fbs`

anuragsingh-tt (approved):

runtime/lib/ttnn/operations/eltwise/binary/binary_composite.cpp
mmanzoorTT: This break is for the outer switch. Currently, it has no impact as we only have one case for the outer switch but will be required when we add more cases/ops here.

PR #5167, opened 2025-10-01
Implement TTNN to TTIR for Matmul op

### LINKED ISSUE #5158: Implement TTNN to TTIR for Matmul op
Extend `ConvertTTNNToTTIR` pass to support converting `TTNN_MatmulOp` to `TTIR_MatmulOp`.

brnorris03:
### Ticket
Closes #5158

### Problem description
The `ttnn.matmul` op is not supported in `ConvertTTNNToTTIR`.

### What's changed
Extends the `ConvertTTNNToTTIR` pass to support converting `TTNN_MatmulOp` to `TTIR_MatmulOp`. The approach is the same as for elementwise ops. TTNN attributes not present in TTIR are _not_ preserved.

The pass is updated to make untagged TTNN operations dynamically legal and do a full conversion.

### Checklist
- Add `TTNN_MatmulOp` to `TTIR_MatmulOp` conversion
- Add Lit tests


lib/Conversion/TTNNToTTIR/TTNNToTTIRPatterns.cpp
```diff
@@ -24,6 +18,24 @@
...
+
+    if (attrName.getValue() == "ttnn.hoist_generic_via_d2m") {
+      continue;
+    }
+
+    if (!destOp->hasAttr(attrName)) {
+      destOp->setAttr(attrName, attr.getValue());
+    }
+  }
+}
```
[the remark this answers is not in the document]
brnorris03: replaced hardcoded string with existing named const

lib/Conversion/TTNNToTTIR/TTNNToTTIRPatterns.cpp
```diff
@@ -45,6 +59,34 @@ class TTNNToTTIRElementwiseConversionPattern
...
+
+      auto newOp = mlir::tt::ttir::utils::replaceOpWithNewDPSOp<
+          mlir::tt::ttir::MatmulOp>(rewriter, srcOp, outputType, adaptor.getA(),
+                                    adaptor.getB(), adaptor.getTransposeA(),
+                                    adaptor.getTransposeB());
+
+      // Preserve TTNN-specific attributes as custom attributes
+      preserveTTNNAttributesAsCustom(srcOp.getOperation(),
+                                     newOp.getOperation());
+    }
```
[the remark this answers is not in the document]
brnorris03: not really applicable here IMO; I haven't really used Co-Pilot before for reviews and was curious

lib/Conversion/TTNNToTTIR/TTNNToTTIRPatterns.cpp
```diff
@@ -45,6 +59,34 @@ class TTNNToTTIRElementwiseConversionPattern
...
+          this->getTypeConverter()->convertType(srcOp.getResult().getType()));
+
+      auto newOp = mlir::tt::ttir::utils::replaceOpWithNewDPSOp<
+          mlir::tt::ttir::MatmulOp>(rewriter, srcOp, outputType, adaptor.getA(),
+                                    adaptor.getB(), adaptor.getTransposeA(),
+                                    adaptor.getTransposeB());
+
+      // Preserve TTNN-specific attributes as custom attributes
+      preserveTTNNAttributesAsCustom(srcOp.getOperation(),
+                                     newOp.getOperation());
```
brnorris03: If this is something we want to do (alternatively these attributes can be simply dropped), I think it should be done for all ops, so I'd add to elementwise, as well.

xanderchin: I think we should be preserving them specifically for cases when we end up translating from ttir back into ttnn.<named op> and not ttnn.generic.

lib/Conversion/TTNNToTTIR/TTNNToTTIRPatterns.cpp
```diff
@@ -45,6 +59,34 @@ class TTNNToTTIRElementwiseConversionPattern
...
+
+  using mlir::OpConversionPattern<
+      mlir::tt::ttnn::MatmulOp>::OpConversionPattern;
+
+public:
+  mlir::LogicalResult
+  matchAndRewrite(mlir::tt::ttnn::MatmulOp srcOp,
+                  mlir::tt::ttnn::MatmulOp::Adaptor adaptor,
+                  mlir::ConversionPatternRewriter &rewriter) const override {
+    if (mlir::tt::ttnn::utils::isTTNNHoistGenericViaD2MOp(srcOp)) {
```
nsmithtt: @phizalev-TT / @vtangTT I missed this the first time around, but I think the way things are currently constructed wouldn't enable recursive rewrites and isn't fully conformant to the "mlir" way of doing pattern rewriting.

We should move this check to the Pass.cpp file by appending the target with:
```c++
    target.addDynamicallyLegalOp<
#define GET_OP_LIST
#include "mlir/Dialect/TTNN/IR/TTNNOps.cpp.inc"
        >([&](Operation *op) {
      return !mlir::tt::ttnn::utils::isTTNNHoistGenericViaD2MOp(op);
    });
```

And then `applyPartialConversion` -> `applyFullConversion`.

lib/Conversion/TTNNToTTIR/TTNNToTTIRPatterns.cpp
nsmithtt: I don't think that TTIRToTTNN conversion will preserve these attributes and in general I don't think it's typical to carry forward op attributes of the dialect you're converting from.

In my opinion we should drop the attributes, it's unclear at this time how we'd support translating TTIR back to TTNN named ops.  For now I think we should treat d2m promotion as a one way street.

lib/Conversion/TTNNToTTIR/TTNNToTTIRPatterns.cpp
nsmithtt: @brnorris03, this is outside the scope of this PR, we can tackle this as a follow on change.

nsmithtt (commented):
This looks great!! Only one comment inline regarding the attribute preservation, imo we shouldn't do this, but I'm open to other's input here.

lib/Conversion/TTNNToTTIR/TTNNToTTIRPatterns.cpp
brnorris03: Definitely can be in scope, will do (if others agree) -- thanks for pointing it out! I was going for minimal changes but I agree this is better (correct).

lib/Conversion/TTNNToTTIR/TTNNToTTIRPatterns.cpp
brnorris03: I am not familiar enough with the different downstream transformations (current and future) to have an opinion. It's trivial enough to preserve them if needed in future, so I could just leave a comment there to mention that they are not if I remove this.

nsmithtt: My thinking is that there's nothing else in the d2m pipeline nor in the conversions back to TTNN that will look for or preserve these attributes, ie, they will just be dropped on the floor elsewhere.  I think it'd be better to tackle this holistically in the future if we ever do find a need to preserve the attributes.

brnorris03: That makes sense. I went ahead and changed to drop them, with a comment mentioning that

lib/Conversion/TTNNToTTIR/TTNNToTTIRPatterns.cpp
brnorris03: Updated as suggested

lib/Conversion/TTNNToTTIR/TTNNToTTIRPass.cpp
```diff
@@ -27,25 +28,33 @@ namespace mlir::tt::ttnn {
...
+    target.addLegalOp<::mlir::ModuleOp>();
 
     TypeConverter typeConverter;
     typeConverter.addConversion([](Type type) { return type; });
 
     RewritePatternSet patterns(&getContext());
 
+    // Mark TTNN dialect as dynamically legal for all ops that do NOT have the
+    // `hoist_generic_via_d2m` attribute
+    target.addDynamicallyLegalDialect<ttnn::TTNNDialect>(
```
nsmithtt, line 52: Oh wow, good find!  I didn't know this method existed!

lib/Conversion/TTNNToTTIR/TTNNToTTIRPass.cpp
```diff
@@ -27,25 +28,33 @@ namespace mlir::tt::ttnn {
 #define GEN_PASS_DEF_CONVERTTTNNTOTTIR
 #include "ttmlir/Conversion/Passes.h.inc"
 
+#define GET_OP_LIST
+#include "ttmlir/Dialect/TTNN/IR/TTNNOps.h.inc"
```
nsmithtt: I think because you found `addDynamicallyLegalDialect`, this GET_OP_LIST include hack isn't needed.

nsmithtt (approved):
One minor nit inline, otherwise this looks great!

lib/Conversion/TTNNToTTIR/TTNNToTTIRPass.cpp
brnorris03: Yep, thanks for noticing!

lib/Conversion/TTNNToTTIR/TTNNToTTIRPass.cpp
brnorris03: Me too, or at least had managed to forget it temporarily

Issue #4094, opened 2025-07-16
Integrate ttrt into builder testing flow


tapspatel:
### Status Quo
As it stands, tests for builder under `test/python/golden` are only there to _generate_ flatbuffers with golden information, that are then to be checked by a second process, i.e. `ttrt`. This leads to a few issues:
- Convoluted test execution; multiple invocations and manual readings of test output are required to get a full picture of our op coverage
- This necessitates a manual marking system to sate the CI, where tests are marked with different failure modes depending on how the flatbuffer they produce behaves when run. Without extremely careful and constant checking, as our codebase updates and more ops begin to work, these marks will desync with the actual behavior of the system.

### Solution
If the `pytest` run itself could instead execute these generated flatbuffers, we could appropriately xfail the ops that we know fail, and then be notified when they come up with an xpass. Furthermore, we would only need a single command to fully test our op coverage, and generate a report for us to analyze. 

### Approach
The suggested approach is to use the nanobinded runtime API within `compile_ttir_to_flatbuffer` (which may be due for a rename as it grows) to follow how the `ttrt` tool executes and checks goldeness of flatbuffers.

sdjordjevicTT:
If I understand the title correctly, this will allow us to write Python builder tests and execute them from the same Python test, correct? This would be excellent, as it would enable us to have xFail Python silicon tests...

tapspatel:
@sdjordjevicTT yeah that's right. we would migrate silicon lit tests into builder ecosystem.

PR #5168, opened 2025-10-01
CI refurbishment Part 2

nsumrakTT:
### Problem description
Test matrix has 26 tests and use 26 different machines.
Some of the tests last just several seconds while preparing to run lasts ~4 minutes.
The test process is very inefficient. 

### What's changed
Test run is prepared by grouping tests from input test matrix into jobs containing test batches based on test duration and target test time.
There is only one test step "Run Tests" which has output of whole batch but test runs are clearly separated and easily searched.
Test durations are collected on successful push to main and automatically used in all subsequent workflows.

These changes will save ~90 minutes of machine time _per run_  and will use 20 HW machines less _per run_ which will have an effect of much lower queue times.

**PLEASE NOTE:**
The way how tests are defined is **significantly** changed!
Documentation on CI is also updated.

### Checklist

### DIFF
--- a/.github/workflows/on-pr.yml
+++ b/.github/workflows/on-pr.yml
@@ -64,18 +64,18 @@ jobs:
           if [[ "${{ env.TARGET_REPO }}" == "tenstorrent/tt-xla" ]]; then
             gh workflow run ${{ env.WORKFLOW_NAME }} \
               --repo ${{ env.TARGET_REPO }} --ref main \
-              --field mlir_override=${{ github.event.pull_request.head.sha }} \
-              --field test_suite=model-test-push.json
+              --raw-field mlir_override=${{ github.event.pull_request.head.sha }} \
+              --raw-field test_suite=basic-test.json
             gh workflow run ${{ env.WORKFLOW_NAME }} \
               --repo ${{ env.TARGET_REPO }} --ref main \
-              --field mlir_override=${{ github.event.pull_request.head.sha }} \
-              --field test_suite=basic-test.json
+              --raw-field mlir_override=${{ github.event.pull_request.head.sha }} \
+              --raw-field test_suite=model-test-push.json
           else
             gh workflow run ${{ env.WORKFLOW_NAME }} \
               --repo ${{ env.TARGET_REPO }} --ref main \
-              --field mlir_override=${{ github.event.pull_request.head.sha }}
+              --raw-field mlir_override=${{ github.event.pull_request.head.sha }}
           fi
-          gh run list --workflow=${{ env.WORKFLOW_NAME }} --repo ${{ env.TARGET_REPO }} --limit 1
+          gh run list --workflow=${{ env.WORKFLOW_NAME }} --repo ${{ env.TARGET_REPO }} --branch main --limit 1
           echo "Triggered ${{ env.TARGET_REPO }} with tt-mlir ${{ github.event.pull_request.head.sha }}"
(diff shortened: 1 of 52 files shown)


nsmithtt (changes requested):
I'm sorry, we should have sync'd before making this change.  We probably should hold off on this rework until our builder test runner supports the runtime.  I want us to change our GH actions to use the matrix as it's intended and let the jobs opt into the configurations that they should run on.

See slide 4 of this presentation, this is the North Star, https://docs.google.com/presentation/d/1KeepEtB_MN7eh4jKGzl8lDVo5-x6PL9AP8n98xeRzNw/edit?slide=id.g357240f26bd_0_0#slide=id.g357240f26bd_0_0

I'm sorry, but I cannot approve this approach as this is going further in the opposite direction from using vanilla GH action matrix syntax.  I think it is achievable once runtime support lands in builder (it's actively being worked on):

- https://github.com/tenstorrent/tt-mlir/issues/4094

nsumrakTT:
@nsmithtt This is the first time I hear about this. I don't think approach in this PR is not in line with what is in the presentation. At least before mentioned ticket is being done, this PR will **significantly** improve tt-mlir workflow.
We should definitely sync about this.

nsmithtt:
We can discuss more in tomorrow's meeting, but I'd really prefer if we didn't introduce a python script to generate the GH build action.  I think it's possible for us to get away with a vanilla GH matrix which would be easier to maintain in the long run.

nsumrakTT:
Failing problem fix: https://github.com/tenstorrent/tt-mlir/pull/5217

.github/test_matrix/pr-push-matrix.json
```diff
@@ -0,0 +1,36 @@
...
+
+  { "runs-on": "n150",   "image": "tracy",  "type": "pytest", "path": "tools/explorer/test/run_tests.py", "flags": "--force-reinstall $BUILD_DIR/bin/wheels/tt_adapter*.whl $BUILD_DIR/bin/wheels/ai_edge_model_explorer*.whl"},
+  { "runs-on": "n150",   "image": "tracy",  "type": "pytest", "path": "runtime/tools/chisel/test", "flags": "-e runtime/tools/chisel/"},
+  { "runs-on": "n300",   "image": "tracy",  "type": "pytest", "path": "runtime/test/ttnn/python/multi_device"},
+
+  { "runs-on": ["n150","llmbox","n300","p150"],
+        "image": "tracy",
+            "type": "builder", "path": "test/python/golden", "args": "not run_error and not fails_golden", "flags": "run-ttrt"},
+
+  { "runs-on": "n150",   "image": "speedy", "type": "builder", "path": "test/python/golden/optimizer", "args": "not run_error and not fails_golden", "flags": "run-ttrt require-opmodel"},
```
nsmithtt: Few additional cleanups, we can save these for follow on PRs:

1. Can we rename `type` to `script`.
2. For the bash scripts can we use the full name, i.e. `builder.sh`, we can chmod them as executable and add `.github/test_scripts` to the `PATH` so that there's a uniform way to invoke the script.
3. Can we unify `path`, `args`, and `flags` under a single key, maybe just `args` that accepts a list?
    - `flags` is sometimes used to represent gh `tags`, we should clean this up and make it explicit. 

Some examples post these changes:
```json
{ "script": "runtime_debug.sh" },
{ "script": "pytest", "args": ["runtime/test/ttnn/python/multi_device"] },
{ "script": "builder.sh", "args": ["test/python/golden/optimizer", "not run_error and not fails_golden", "--require-opmodel"], "tags": ["require-opmodel"] },
```

vvukomanTT (approved):

.github/test_matrix/pr-push-matrix.json
nsumrakTT: Done.
I've introduced `reqs` for requirements as it is used, tags is also possible but we can add it later when it is actually used.

nsmithtt: Awesome, thank you!

PR #5176, opened 2025-10-01
Remove strict consetval check: matching tupleness in input/output

### LINKED ISSUE #4841: ConstEval Fails on EmitC Path if tuplifyInputIfEmpty Is False
### Describe the bug

For any const-eval graph with no inputs, we get an assertion error when running `--ttnn-tuplify-tensors` unless `"tuplify-input-if-empty=true"` is set.

### Steps to reproduce the issue

1. copy the `@creation_ops()` test from `test/ttmlir/Dialect/TTNN/const-eval/const-eval.mlir` into `tmp.mlir`
2. run `ttmlir-opt --ttir-to-ttnn-backend-pipeline tmp.mlir -o tmp.ttnn.mlir`
3. run `ttmlir-opt --ttnn-tuplify-tensors --convert-ttnn-to-emitc --mlir-print-ir-after-all tmp.ttnn.mlir`
4. observe the assertion failure
5. run `ttmlir-opt --ttnn-tuplify-tensors="tuplify-input-if-empty=true" --convert-ttnn-to-emitc --mlir-print-ir-after-all tmp.ttnn.mlir` -- observe the tuplify empty flag resolves error.

### Expected behavior

Somehow we should be more robust to this--not clear to me exactly why this option exists vs defaulting to `true`.  Perhaps need to sync w/ Sasa or Aleksander about this.

_No response_

svuckovicTT:
### Ticket
Closes #4841

### Problem description
Sometimes consteval funcs get created that take no input. A verifier fires an error, requiring that both input/output are tuples, or neither of them are. We should allow for these cases as they seem valid.

### What's changed
Removed the verifier check.

### Checklist

### DIFF
--- a/.github/workflows/wheel-build.yml
+++ b/.github/workflows/wheel-build.yml
@@ -34,7 +34,6 @@ jobs:
           docker-images: true
           swap-storage: true
 
-
       - uses: actions/checkout@v4
         with:
           fetch-depth: 0


vwellsTT (approved):

lib/Dialect/TTCore/IR/TTCoreOps.cpp
```diff
@@ -156,12 +156,6 @@ LogicalResult LoadCachedOp::verify() {
   bool hasTupleResult = fnType.getNumResults() == 1 &&
```
vwellsTT, line 156: there's probably a loose argument to be made that it would be more ideal to change these bools to be like `hasTupleOrEmptyInput/Result`, add a condition like `(fnType.getNumInputs() == 1 && mlir::isa<mlir::TupleType>(fnType.getInput(0)) || fnType.getNumInputs() == 0`, and then leave the check.  But in practice I can't see much actual risk we end up with mismatch of tuples vs non-tuples anyway, so I don't feel strongly either way.

svuckovicTT: Updated the code, kept the variable names as it made more sense to me, but commented the intent - lmk if you disagree with it.

azecevicTT (approved):

lib/Dialect/TTCore/IR/TTCoreOps.cpp
svuckovicTT: This IR fails the proposed check:
```mlir
%6 = "ttcore.load_cached"() <{callee = @arange_const_eval_0}> : () -> tensor<1x32x1x1xf32, #ttnn.ttnn_layout<(d0, d1, d2, d3) -> (d0 * 1024 + d1 * 32 + d2, d3), <1x1>, memref<32x1x!ttcore.tile<32x32, f32>, #ttnn.buffer_type<dram>>, <interleaved>>>
```

`hasTupleInput` is `true`, while ` hasTupleResult` is `false` as it's not a tuple return type but has a tensor - I'll go ahead and remove the check.

sdjordjevicTT (approved):

vvukomanTT (approved):

PR #5185, opened 2025-10-01
D2M to TTKernel SFPU op placement fix

nsmithtt:
Previously we were attempting to hoist the SFPU op to its earliest possible insertion point by analyzing its relative location to its operands.  This breaks down when the operands span blocks, we can only safely apply this movement if we can guarantee we're pushing the op downward in program order and in the same block.

Introduce new boolean `setInsertionPointAfterOperands` called `allowHoisting` which can be set to true for init functions where it's required to hoist them out of loops.

### DIFF
--- a/lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
+++ b/lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
@@ -196,7 +196,8 @@ static Value getOutCB(ConversionPatternRewriter &rewriter, Operation *op) {
 }
 
 static void setInsertionPointAfterOperands(OpBuilder &rewriter,
-                                           llvm::ArrayRef<Value> operands) {
+                                           llvm::ArrayRef<Value> operands,
+                                           bool allowHoisting) {
   Operation *latestDefOp = nullptr;
   for (Value operand : operands) {
     Operation *definingOp = operand.getDefiningOp();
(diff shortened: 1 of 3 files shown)


wenbinlyuTT:
I think this need proper handling of `init_sfpu` and similar, right now if you compile
```
module {
  func.func @neg(%arg0: tensor<32x1x1xf32>) -> tensor<32x1x1xf32> {
    %0 = ttir.empty() : tensor<32x1x1xf32>
    %1 = "ttir.neg"(%arg0, %0) : (tensor<32x1x1xf32>, tensor<32x1x1xf32>) -> tensor<32x1x1xf32>
    return %1 : tensor<32x1x1xf32>
  }
}
```
you will see that `init_sfpu` is stuck inside the loop.

I encountered this exact issue and I did:
```cpp
static void setInsertionPointAfterOperands(OpBuilder &rewriter,
                                           llvm::ArrayRef<Value> operands,
                                           const bool allowLoopEscape) {
  // ...

  // Only move the insertion point if we're pushing it downward in the
  // topological order, or if loop escape is allowed.
  auto currentInsertionPoint = rewriter.getInsertionPoint();

  Block *currentBlock = currentInsertionPoint->getBlock();
  Block *targetBlock = latestDefOp->getBlock();

  if (currentBlock == targetBlock) {
    if (currentInsertionPoint->isBeforeInBlock(latestDefOp)) {
      rewriter.setInsertionPointAfter(latestDefOp);
    }
  } else {
    // Different blocks: walk up the region hierarchy to see if target is in a
    // parent region/loop nest.
    Region *currentRegion = currentBlock->getParent();
    Region *targetRegion = targetBlock->getParent();
    Region *searchRegion = currentRegion;
    bool targetIsParent = false;
    while (searchRegion) {
      if (searchRegion == targetRegion) {
        targetIsParent = true;
        break;
      }
      Operation *parentOp = searchRegion->getParentOp();
      searchRegion = parentOp ? parentOp->getParentRegion() : nullptr;
    }

    // Move the insertion point if the target isn't outside of the loop, or if
    // loop escape is explicitly allowed (e.g. for init calls).
    if (!targetIsParent || allowLoopEscape) {
      rewriter.setInsertionPointAfter(latestDefOp);
    }
  }
}
```
There probably is a better way...

nsmithtt:
> I think this need proper handling of `init_sfpu` and similar, right now if you compile

@wenbinlyuTT and I worked offline and came up with an `allowHoisting` flag for `setInsertionPointAfterOperands` which can be set to true for init functions where it's required to hoist them out of loops.

PR #5187, opened 2025-10-01
Update the stream_layout op's verifier

nsmithtt:
The stream layout previously didn't have a stringent enough verifier, which is now able to validate that shard shapes between the storage and the result type are equal.

There was one bit of fallout that was detected by an existing test `test/ttmlir/Dialect/D2M/allocate/generic_form_streams_matmul.mlir`. The subtest `matmul_multi_core_transpose` uses a view_layout op to transpose the input on the fly.  The allocator needed a small update to the stream_layout's storage memref type which must come from the post view's shape.  This is accomplished by reading the generic's operand type.

### DIFF
--- a/lib/Dialect/D2M/IR/D2MOps.cpp
+++ b/lib/Dialect/D2M/IR/D2MOps.cpp
@@ -210,103 +210,59 @@ struct ToLayoutFoldRedundantPattern : public OpRewritePattern<ToLayoutOp> {
 };
 
 static ::mlir::LogicalResult
-verifyLayoutOp(mlir::Operation *op, mlir::Type inputTensorOrMemrefTy,
-               mlir::Type outputTensorOrMemrefTy, bool allowFormatChange,
-               bool allowMemorySpaceChange, bool checkMemrefRank = false,
-               bool checkMemrefGridShardForm = false,
-               bool checkMemrefShardShape = false) {
-  if (mlir::RankedTensorType inputTy =
-          mlir::dyn_cast<mlir::RankedTensorType>(inputTensorOrMemrefTy)) {
-    mlir::RankedTensorType outputTy =
-        mlir::dyn_cast<mlir::RankedTensorType>(outputTensorOrMemrefTy);
-    if (!outputTy) {
-      return op->emitOpError("Input and output types must be the same");
-    }
-
-    auto inputLayout =
-        mlir::dyn_cast_if_present<mlir::tt::ttcore::MetalLayoutAttr>(
-            inputTy.getEncoding());
-    auto outputLayout =
-        mlir::dyn_cast_if_present<mlir::tt::ttcore::MetalLayoutAttr>(
-            outputTy.getEncoding());
-
-    if (!inputLayout || !outputLayout) {
-      // If the input/output tensor does not have a layout, we can early exit.
-      return mlir::success();
-    }
-
-    const bool isFormatChange =
-        inputTy.getElementType() != outputTy.getElementType();
-    if (isFormatChange && !allowFormatChange) {
-      return op->emitOpError(
-          "Input and output tensor element types must be the same");
-    }
-
-    const bool isMemorySpaceChange =
-        inputLayout.getMemorySpace() != outputLayout.getMemorySpace();
-    if (!allowMemorySpaceChange && isMemorySpaceChange) {
-      return op->emitOpError(
-          "Input and output layout memory spaces must be the same");
-    }
+verifyLayoutOp(mlir::Operation *op, const char *aName, const char *bName,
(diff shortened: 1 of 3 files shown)


phizalev-TT (approved):

lib/Dialect/D2M/IR/D2MOps.cpp
```diff
@@ -210,103 +210,59 @@ struct ToLayoutFoldRedundantPattern : public OpRewritePattern<ToLayoutOp> {
...
-
-    const bool isMemorySpaceChange =
-        inputLayout.getMemorySpace() != outputLayout.getMemorySpace();
-    if (!allowMemorySpaceChange && isMemorySpaceChange) {
-      return op->emitOpError(
-          "Input and output layout memory spaces must be the same");
-    }
+verifyLayoutOp(mlir::Operation *op, const char *aName, const char *bName,
+               mlir::Type aType, mlir::Type bType, bool checkSameElementType,
+               bool checkSameMemorySpace = false, bool checkSameRank = false,
```
vwellsTT: can we remove the default values now, since you're specifying them explicitly?

lib/Dialect/D2M/IR/D2MOps.cpp
```diff
@@ -210,103 +210,59 @@ struct ToLayoutFoldRedundantPattern : public OpRewritePattern<ToLayoutOp> {
...
-            outputTy.getLayout());
-    if (checkMemrefGridShardForm && !outputDeviceLayout) {
-      return op->emitOpError(
-          "output memref must have device layout, i.e. have even rank, grid "
-          "shape followed by shard shape of equal rank, e.g. GxGxSxS");
-    }
+  if (checkSameGridShape &&
+      (aLayout.getGridShape(a) != bLayout.getGridShape(b))) {
+    return op->emitOpError()
+           << aName << " and " << bName << "'s grid shape must be the same";
```
vwellsTT: as a general note, it might be nicer if we could provide what the mismatched grid shapes are here (and same for other related error).  I guess on the other hand we don't usually do such things, perhaps it's not idiomatic for MLIR error/too long/or something

lib/Dialect/D2M/IR/D2MOps.cpp
vwellsTT: I guess we don't for the ToLayout case, but maybe we should.  I personally hate C++ default params, find them very error-prone 🙂

lib/Dialect/D2M/Transforms/Allocate.cpp
```diff
@@ -619,7 +623,10 @@ class D2MAllocate final : public impl::D2MAllocateBase<D2MAllocate> {
                     // In principle, buffer shape/size could depend on whether
                     // the stream is out of L1 or DRAM... but not right now.
                     operandCtx.bufferType = selectStreamBuffer(
-                        rewriter, memrefCtx.type, numStreamBuffers);
+                        rewriter,
+                        mlir::cast<MemRefType>(
+                            user.getOperand(operandCtx.operandIndex).getType()),
+                        numStreamBuffers);
```
vroubtsovTT, line 629: When can the defining memref's type (`memrefCtx.type`) and operand type differ? When some layout aspects are altered by an intervening `layout_stream`?

nsmithtt: In the case where there's a view in between, tbh it feels like there's going to be edge cases here, but the test that caught this (`test/ttmlir/Dialect/D2M/allocate/generic_form_streams_matmul.mlir`) was doing (simplified version):
```mlir
%output_for_some_other_op = memref.alloc() : memref<1x1x8x6x!ttcore.tile<32x32, f32>, #ttcore.shard, #l1_>

// ... output_for_some_other_op is written to ...

%lhs : memref<1x1x4x6x!ttcore.tile<32x32, f32>, #ttcore.shard, #l1_>

// The view transposes the access to memory
%rhs = d2m.view_layout %output_for_some_other_op
  :   memref<1x1x8x6x!ttcore.tile<32x32, f32>, #ttcore.shard, #l1_>
  ->  memref<1x1x6x8x!ttcore.tile<32x32, f32>, #ttcore.view<(d0, d1, d2, d3) -> (d1, d0, d3, d2)>, #l1_>

d2m.generic ins(%lhs, %rhs) ...
```

Note that:
- Before the view, the memref has shape 8x6
- After the view, the memref has shape 6x8

The generic op validation will fail if it doesn't get the post-view shape because it'll check that both the lhs and rhs have the same inner dim of 6.

Transitively, this means that the stream storage must also have shape 6x8, i.e. post-view.

vroubtsovTT: Yes, transpose is a good example.

Issue #5023, opened 2025-09-18
[D2M] Unaligned shapes not working with dest register set to fp32


dlokeTT:
### Describe the bug

Unaligned tests failing now that fp32_dest_accum set to true when using fp32

wenbinlyuTT:
Cause:
* By default we use DST half sync. When fp32DST is enabled at the same time, we only have 4 DST tiles available.
* Both the `GenericTileComputeLoops` pass and the `InsertDstRegisterAccess` pass doesn't properly handle the (dynamic) capacity of the DST, especially for binary SFPU ops that needs 3 tiles at a time.
* These tests use tensors so large that we start to have shards larger than `1x1`, and the results are incorrect since the compiler tries to do more ops than the DST capacity allows in a single iteration.

jojo-771771:
I can work on this. Please assign!

jojo-771771:
I can work on this. Please assign!

jojo-771771:
I can work on this. Please assign!

jojo-771771:
I can work on this. Please assign!

PR #4964, opened 2025-09-16
re enabling DST f32 accumulation mode

dlokeTT:
### Ticket
Link to Github Issue

### Problem description
Provide context for the problem.

### What's changed
Describe the approach used to solve the problem.
Summarize the changes made and its impact.

### Checklist
- [ ] New/Existing tests provide coverage for changes

PR #5206, opened 2025-10-02
D2M: unify DST capacity calculation and take 32b mode into account

wenbinlyuTT:
### Ticket
Closes: #5023
Related-To: #4964 #5108 #5151

### Problem description
By default we set DST to `SyncHalf`, so after `fp32DestAccum` is enabled in #4964, we're only left with 4 tiles to work with if we're doing f32 operations.

Existing code isn't aware of the DST's dynamic capacity (the fusion / loop tiling / insert DST access passes), they assume 8 tiles and go straight to overflow land.

Additionally, we only consider if the output can fit in the DST, ignoring inputs and/or intermediate results, worsening the overflow issue.

We typically only test small f32 tensors so this has rarely caused any trouble in the CI. For the few tests that does trigger overflow, the only symptom is PCC error and we simply masked the off (e.g. the matmul tests and the unaligned shapes tests).

### What's changed
* Make clear distinction between the DST's physical size and logical size in the code base.
* Unified logical DST capacity calculation logic as a helper function under `ttcore::ChipDescAttr`.
* Now the logical capacity of the DST will be cut in half if any of the operands of a GenericOp is 32b.
* Renamed `max-dst-register-size-tiles` to `max-dst-physical-size-tiles` and support this option in the insert DST access pass as well.
* Be explicit about our choice of setting `SyncHalf`, left a note about its implications.
* Switch to conservative loop tiling if any of the compute op uses more than one DST tile, or doesn't use the DST in-place.
* In the insert DST access pass, we now assert if we allocate more DST slices than it can hold. The change above is the defense against this situation.
* Memref linearization is disabled for block matmuls due to the un-collapsible strided `memref.subview` resulted from limited DST capacity 
* Reworked the affected tests: original tests are switched to `f16/bf16` to keep the shapes the same, the f32 variant of these tests are being tested under the override ``max-dst-register-size-tiles=32`.
* Various cleanups, renames and comments to assist understanding.

### Checklist

### DIFF
--- a/lib/Dialect/D2M/Transforms/GenericLinearizeMemref.cpp
+++ b/lib/Dialect/D2M/Transforms/GenericLinearizeMemref.cpp
@@ -110,26 +110,6 @@ public:
       : OpRewritePattern<TileMatmulBlockOp>(context),
         collapseOps(&collapseOps) {}
 
-  static std::optional<memref::CollapseShapeOp>
-  linearize(Value val, PatternRewriter &rewriter,
-            DenseMap<Value, memref::CollapseShapeOp> *cache) {
-    auto memrefTy = mlir::cast<MemRefType>(val.getType());
-    memref::CollapseShapeOp linearized = cache->lookup(val);
-    if (!linearized) {
-      rewriter.setInsertionPointAfterValue(val);
-      SmallVector<ReassociationIndices, 4> collapsedDims = {
-          llvm::to_vector(llvm::seq<int64_t>(0, memrefTy.getRank()))};
-      if (!memref::CollapseShapeOp::isGuaranteedCollapsible(memrefTy,
-                                                            collapsedDims)) {
-        return std::nullopt;
-      }
-      linearized = rewriter.create<memref::CollapseShapeOp>(val.getLoc(), val,
-                                                            collapsedDims);
-      cache->insert({val, linearized});
-    }
-    return linearized;
-  }
-
   static int64_t getNumColumns(Value view) {
     if (auto castOp = mlir::dyn_cast<memref::CastOp>(view.getDefiningOp())) {
       view = castOp.getSource();
@@ -160,24 +140,11 @@ public:
     int64_t ctDim = typeB.getShape()[1];
     int64_t ntDim = getNumColumns(op.getB());
 
-    auto linA = linearize(op.getA(), rewriter, collapseOps);
-    auto linB = linearize(op.getB(), rewriter, collapseOps);
-    auto linO = linearize(op.getOutput(), rewriter, collapseOps);
-
     rewriter.setInsertionPoint(op);
-    // For memrefs that are not collapsible (e.g., strided subviews), set the
-    // block dimensions but keep the original operands.
-    if (!linA || !linB || !linO) {
-      rewriter.replaceOpWithNewOp<TileMatmulBlockOp>(
-          op, op.getA(), op.getB(), op.getOutput(),
-          rewriter.getI64IntegerAttr(rtDim), rewriter.getI64IntegerAttr(ktDim),
-          rewriter.getI64IntegerAttr(ctDim), rewriter.getI64IntegerAttr(ntDim));
-    } else {
-      rewriter.replaceOpWithNewOp<TileMatmulBlockOp>(
-          op, linA->getResult(), linB->getResult(), linO->getResult(),
-          rewriter.getI64IntegerAttr(rtDim), rewriter.getI64IntegerAttr(ktDim),
-          rewriter.getI64IntegerAttr(ctDim), rewriter.getI64IntegerAttr(ntDim));
-    }
+    rewriter.replaceOpWithNewOp<TileMatmulBlockOp>(
+        op, op.getA(), op.getB(), op.getOutput(),
+        rewriter.getI64IntegerAttr(rtDim), rewriter.getI64IntegerAttr(ktDim),
+        rewriter.getI64IntegerAttr(ctDim), rewriter.getI64IntegerAttr(ntDim));
 
     return success();
   }
--- a/lib/Dialect/D2M/Transforms/InsertDstRegisterAccess.cpp
+++ b/lib/Dialect/D2M/Transforms/InsertDstRegisterAccess.cpp
@@ -215,9 +215,9 @@ public:
 
     auto [cbType, maxDstSliceIdx] = inferCbInfoFromAllAccesses(copyInfos);
     // Calculate dst shape as N slices of cb shape.
-    int64_t volume = ttmlir::utils::volume(cbType.getShape());
-    assert(volume <= dstCapacity);
-    int64_t numDstSlices = dstCapacity / volume;
+    const int64_t volume = ttmlir::utils::volume(cbType.getShape());
+    TT_assert(volume <= dstCapacity);
+    const int64_t numDstSlices = dstCapacity / volume;
     TT_assertv(maxDstSliceIdx < numDstSlices,
                "Insufficient DST capacity for all operands.");
     SmallVector<int64_t> dstShape({numDstSlices});
--- a/lib/Dialect/TTCore/IR/TTCoreOpsTypes.cpp
+++ b/lib/Dialect/TTCore/IR/TTCoreOpsTypes.cpp
@@ -537,7 +537,8 @@ ChipDescAttr::getDstLogicalSizeTiles(Type type, bool fullSyncEn,
     nDstTiles /= 2;
   }
   // Will/should this be in-sync with the compute thread's fp32DestAccum?
-  if (type.getIntOrFloatBitWidth() == 32) {
+  const auto dataType = ttcore::elementTypeToDataType(type);
+  if (ttcore::getNumberOfBits(dataType) == 32u) {
     nDstTiles /= 2;
   }
   return nDstTiles;
--- a/lib/Dialect/D2M/Transforms/ElementwiseFusion.cpp
+++ b/lib/Dialect/D2M/Transforms/ElementwiseFusion.cpp
@@ -353,6 +353,7 @@ struct FuseD2MElementwiseOpsPattern : public OpRewritePattern<GenericOp> {
       return failure();
     }
 
+    assert(consumer.getNumRegions() == 1u);
     Type largestDstType =
         utils::getRegionLargestDstElemType(consumer.getRegion(0));
     const unsigned dstCapacity =
--- a/lib/Dialect/D2M/Utils/Utils.cpp
+++ b/lib/Dialect/D2M/Utils/Utils.cpp
@@ -57,6 +57,10 @@ getSquareTargetGrid(mlir::ArrayRef<int64_t> targetGridShape) {
 }
 
 Type getRegionLargestDstElemType(Region &region) {
+  auto getTypeNumberOfBits = [](Type type) {
+    return ttcore::getNumberOfBits(ttcore::elementTypeToDataType(type));
+  };
+
   Type largestType = nullptr;
   region.walk([&](OperandLoadStoreRegisterOpInterface op) {
     // Only the typecast op has different input & output types, but it's a DST
@@ -65,11 +69,11 @@ Type getRegionLargestDstElemType(Region &region) {
       Type t = ttcore::getOperandInnerElementType(v);
 
       if (!largestType ||
-          (t.getIntOrFloatBitWidth() > largestType.getIntOrFloatBitWidth())) {
+          (getTypeNumberOfBits(t) > getTypeNumberOfBits(largestType))) {
         largestType = t;
       }
 
-      if (largestType && largestType.getIntOrFloatBitWidth() == 32) {
+      if (largestType && getTypeNumberOfBits(largestType) >= 32u) {
         return WalkResult::interrupt();
       }
     }
@@ -77,7 +81,7 @@ Type getRegionLargestDstElemType(Region &region) {
   });
 
   assert(largestType);
-  TT_assert(largestType.getIntOrFloatBitWidth() <= 32u);
+  TT_assert(getTypeNumberOfBits(largestType) <= 32u);
   return largestType;
 }
 
--- a/python/TTModule.cpp
+++ b/python/TTModule.cpp
@@ -189,7 +189,7 @@ void populateTTModule(nb::module_ &m) {
                                                overridePhysicalSize);
           },
           nb::arg("type"), nb::arg("full_sync_en") = false,
-          nb::arg("override_physical_size") = std::nullopt)
+          nb::arg("override_physical_size") = 0)
       .def_prop_ro("arch", &tt::ttcore::ChipDescAttr::getArch)
       .def_prop_ro(
           "grid",


lib/Conversion/D2MToTTMetal/D2MToTTMetal.cpp
```diff
@@ -75,15 +60,20 @@ class D2MGenericRewriter : public OpConversionPattern<d2m::GenericOp> {
...
+          auto elemType = ttcore::getOperandInnerElementType(operands[i]);
           if (elemType.getIntOrFloatBitWidth() == 32) {
             fp32DestAccum = true;
           }
         }
+        // This setting is a perfomance optimization closely related to all the
+        // hard-coded dstRegisterSizeTiles/dst_register_size_tiles = 8 scattered
+        // across our code base. Should this becomes a variable, the max DST
+        // size needs to change to 16 and halfed when this setting is false.
+        constexpr bool dstFullSyncEn = false;
```
dlokeTT, line 72: I think if we're going to have references to full sync here then it would be good to change the # of dest tiles to 16 in the system descriptor and then divide by 2 if we do half sync. Would just clear up exactly how much space we have in dest and why especially since it was confusing for me to learn about and probably will clear it up for others in the future. Not sure if this should fit in this pr or a separate one or not needed at all. Lmk what y'all think @nsmithtt @wenbinlyuTT

dlokeTT (approved):
Other than that thing about full sync, LGTM!

lib/Conversion/D2MToTTMetal/D2MToTTMetal.cpp
wenbinlyuTT: It is possible to add an attribute to `d2m.generic` indicating the full/half sync mode and start to reduce DST capacity from 16 [like this](https://github.com/tenstorrent/tt-metal/blob/a3385d13bd89f537072cf4fb52792a913aaf6f07/ttnn/cpp/ttnn/operations/core/compute_kernel/compute_kernel_config.cpp#L208). I'm just unsure if it's the right place, or if it's worth it to clutter the IR further (the attribute will stay `SyncHalf` for quite some time).

nsmithtt: I think it's not a bad idea to encode 16 in the system desc and we could add member a function on the `ChipDescAttr`:
```c++
size_t getDstRegisterSizeTiles(TileType type,
                               bool fullSyncEn = false,
                               std::optional<size_t> dstRegisterSizeOverride = std::nullopt) {
  size_t dstSize = dstRegisterSizeOverride.value_or(getDstRegisterSizeTiles()); // now returns 16
  if (type.is32Bit()) {
    dstSize /= 2;
  }
  if (!fullSyncEn) {
    dstSize /= 2;
  }
  return dstSize;
}

File: include/ttmlir/Dialect/D2M/IR/D2MOps.td [hunk 2]
```diff
@@ -289,6 +289,7 @@ def D2M_GenericOp : D2M_Op<"generic",
       unsigned getNumLoops();
       unsigned getNumDims();
       AffineMap getIndexingMap(int64_t operandIndex);
+      unsigned getDstTileCapacity();
```
nsmithtt: I think it might be better to have member function with something like:
- `TileType getDstModeTileType();`
- `bool needs32BitDstMode();`
- something that just walks the generic regions ops to determine if 32 bit mode is needed.

Something that can be used in conjunction with the ChipDescAttr member function from the other thread, so that the caller can do:
`chipDesc.getDstRegisterSizeTiles(op.getDstMode....())`

I'd prefer to not have member functions that go and lookup stuff from elsewhere in the IR, I'd prefer the onus to be on the caller to stitch these two members together.

include/ttmlir/Dialect/TTMetal/Pipelines/TTMetalPipelines.h
```diff
@@ -44,12 +44,6 @@ struct TTIRToTTMetalPipelineOptions
                                   "Use mock blackhole system desc.")),
       llvm::cl::init(ttcore::Arch::WormholeB0)};
 
-  Option<unsigned> maxDstRegisterSizeTiles{
-      *this, "max-dst-register-size-tiles",
-      llvm::cl::desc("Clamp the maximum destination register size in tiles. 0 "
-                     "means unset."),
-      llvm::cl::init(0)};
```
nsmithtt, line 51: Wait I think this would still be a really nice pipeline option to continue supporting.  This will enable tests to artificially clamp dst to be smaller and is a nice escape hatch (that we recently used) for debugging.

lib/Dialect/D2M/Transforms/GenericLinearizeMemref.cpp
nsmithtt, line 1: It's somewhat concerning that this pass had to change.  Do we understand the IR diff here?

include/ttmlir/Dialect/D2M/IR/D2MOpsInterfaces.td
```diff
@@ -107,29 +107,13 @@ def D2M_GenericParent : OpInterface<"GenericParent"> {
...
-        sourced from the destination register. It is used by the insert-dst-register-access
-        pass to autogenerate copynests to and from dst.}],
-      /*retTy=*/"::mlir::SmallVector<int64_t>",
-      /*methodName=*/"getOperandsLoadFromDstRegister",
-      /*args=*/(ins),
-      /*methodBody=*/"",
-      /*defaultImplementation=*/"return ::mlir::SmallVector<int64_t>{};"
-    >,
-  ];
-}
```
nsmithtt, line 124: Oh weird, was this just spuriously alongside the LoadStore version and unused?

lib/Dialect/D2M/Transforms/GenericTileComputeLoops.cpp
```diff
@@ -109,19 +130,42 @@ static SmallVector<int64_t> calculateOptimalSubblockSizes(
...
         mlir::cast<MemRefType>(op.getOutputs().front().getType());
-    SmallVector<int64_t> subblockSizes = calculateOptimalSubblockSizes(
-        op.getIndexingMapsArray(), op.getInputs(), outputTensor.getShape(),
-        dstRegisterSizeTiles);
+
+    SmallVector<int64_t> subblockSizes(outputTensor.getShape().size(), 1);
+    if (optimizeSubblocking) {
+      auto *region =
+          ttmlir::utils::getRegionWithParentOfType<d2m::GenericOp>(op);
+      auto d2mGenericOp = mlir::cast<d2m::GenericOp>(region->getParentOp());
```
nsmithtt: `op->getParentOfType<d2m::GenericOp>()`

lib/Dialect/D2M/Transforms/InsertDstRegisterAccess.cpp
```diff
@@ -162,34 +158,57 @@ struct D2MInsertDstRegisterAccessRewriter final
...
-    auto device = ttcore::lookupDevice(op);
-    auto systemDesc = ttcore::getCurrentScopeSystemDesc(op);
-    auto chipIds = device.getChipIds();
-    auto chipDescs = systemDesc.getChipDescs();
-    auto chipDescIndices = systemDesc.getChipDescIndices();
-    auto chipDesc = chipDescs[chipDescIndices[chipIds[0]]];
-    return chipDesc.getDstRegisterSizeTiles();
+  static std::pair<MemRefType, int64_t>
+  inferCbInfoFromAllAccesses(const CopyInfoMap &copyInfos) {
+    MemRefType canonicalType;
```
nsmithtt: I think we can default to nullptr instead of having to separately track `first`.

lib/Dialect/D2M/Transforms/InsertDstRegisterAccess.cpp
```diff
@@ -162,34 +158,57 @@ struct D2MInsertDstRegisterAccessRewriter final
...
+          canonicalType = storeOp.getMemRefType();
+          first = false;
+        } else {
+          TT_assertv(storeOp.getMemRefType() == canonicalType,
+                     "Multiple interpretations of DST not supported.");
+        }
+        maxDstSliceIdx = std::max(maxDstSliceIdx, idx);
+      }
+    }
+    TT_assert(!first);
```
nsmithtt: could be good to assert maxDstSliceIdx > 0

include/ttmlir/Dialect/TTMetal/Pipelines/TTMetalPipelines.h
wenbinlyuTT: Oh yeah I meant to remove its usage from the compute loops pass, this is indeed is still useful.

include/ttmlir/Dialect/D2M/IR/D2MOpsInterfaces.td
wenbinlyuTT: Yeah I was confused in the beginning and had to double check to make sure it's unused.

vwellsTT: Hmm, seems to have been moved here: https://github.com/tenstorrent/tt-mlir/pull/4845/files#diff-eaacc0d16a3cb56bf8a2bc24735d78bd72bd9f98370e5afa036b51ac9788a20fL113
and somehow my refactor kept both wrongly

test/ttmlir/Silicon/TTMetal/n150/simple_constant.mlir
```diff
@@ -18,7 +18,7 @@ module {
   }
 
   // CHECK-LABEL: func.func public @add5_unaligned
-  func.func public @add5_unaligned(%arg0: tensor<9x43x7xf32>) -> tensor<9x43x7xf32> {
+  func.func public @add5_unaligned(%arg0: tensor<43x7xf32>) -> tensor<43x7xf32> {
```
vwellsTT: why is this test made "easier"?

test/ttmlir/Silicon/TTMetal/n150/simple_add.mlir
```diff
@@ -1,13 +1,13 @@
-// RUN: ttmlir-opt --ttir-to-ttmetal-pipeline="system-desc-path=%system_desc_path% override-device-shape=1,1" -o %t.mlir %s
+// RUN: ttmlir-opt --ttir-to-ttmetal-pipeline="system-desc-path=%system_desc_path% override-device-shape=2,2" -o %t.mlir %s
```
vwellsTT: I also don't think I get the motive for changing this test

test/ttmlir/Dialect/D2M/Transforms/insert_dst_register_access_matmul_tile.mlir
```diff
@@ -51,17 +51,17 @@ module {
   }
 
   func.func private @generic_matmul(
-    %in0: memref<1x1x3x3x!ttcore.tile<32x32, f32>, #ttcore.shard<12288x4096>, #ttcore.memory_space<l1>>,
-    %in1: memref<1x1x3x2x!ttcore.tile<32x32, f32>, #ttcore.shard<8192x4096>, #ttcore.memory_space<l1>>,
-    %out0: memref<1x1x3x2x!ttcore.tile<32x32, f32>, #ttcore.shard<8192x4096>, #ttcore.memory_space<l1>>
+    %in0: memref<1x1x3x3x!ttcore.tile<32x32, f16>, #ttcore.shard<6144x2048>, #l1_>,
```
vwellsTT, line 30: for this and other insert_dst_register_access test, it's probably good to leave some f32 matmul test coverage--perhaps we should separate into 2 tests or something instead?

lib/Dialect/TTCore/IR/Utils.cpp
```diff
@@ -43,6 +43,16 @@ DeviceAttr lookupDevice(Operation *op, llvm::StringRef deviceName) {
   return deviceOp.getDeviceAttr();
 }
 
+unsigned getDstRegisterSizeTiles(Operation *op) {
+  auto device = ttcore::lookupDevice(op);
+  auto systemDesc = ttcore::getCurrentScopeSystemDesc(op);
+  auto chipIds = device.getChipIds();
```
vwellsTT, line 48: nit: I'd move this up a line and add a line break to get like 
```
device = ...
chipIds = device.getChipIds

sysDesc = ...
<all fields populated from sysDec>

auto chipDesc = chipDescs[chipDescIndices[chipIds[0]]];

return chipDesc.getDstRegisterSizeTiles();
```
to make more visually clear

lib/Dialect/D2M/Transforms/GenericTileComputeLoops.cpp
```diff
@@ -40,6 +47,19 @@ calculateOutputSubblockFactors(ArrayRef<int64_t> outputBlockShape,
   return llvm::to_vector(llvm::reverse(outputBlockFactors));
 }
 
+// For a linalg op with multiple operands (inputs & output) that have different
```
vwellsTT, line 51: Great comment 🙂

lib/Dialect/D2M/Transforms/GenericLinearizeMemref.cpp
vwellsTT: yeah, I also don't think I'm following how this is related?

include/ttmlir/Dialect/TTIR/IR/TTIROpsInterfaces.td
```diff
@@ -70,35 +70,6 @@ def TTIR_TensorManipulation : OpInterface<"TensorManipulation"> {
   let cppNamespace = "::mlir::tt::ttir";
 }
 
-def TTIR_OperandLoadStoreRegisterOpInterface : OpInterface<"OperandLoadStoreRegisterOpInterface"> {
```
vwellsTT, line 73: bleh, good catch here too

lib/Dialect/D2M/Transforms/InsertDstRegisterAccess.cpp
```diff
@@ -162,34 +158,55 @@ struct D2MInsertDstRegisterAccessRewriter final
...
-    auto chipDescs = systemDesc.getChipDescs();
-    auto chipDescIndices = systemDesc.getChipDescIndices();
-    auto chipDesc = chipDescs[chipDescIndices[chipIds[0]]];
-    return chipDesc.getDstRegisterSizeTiles();
+  static std::pair<MemRefType, int64_t>
+  inferCbInfoFromAllAccesses(const CopyInfoMap &copyInfos) {
+    MemRefType canonicalType = nullptr;
+    int64_t maxDstSliceIdx = -1;
+
+    for (auto [loopNest, copyInfo] : copyInfos) {
```
vwellsTT, line 179: nit: I don't think you want to copy the copyInfo structs here, better to use ref/const ref

wenbinlyuTT: Nice suggestion! Unfortunately `getMemrefType()` isn't `const` and so I had to do a copy some where. I've decided to do it in the outer loop.

lib/Dialect/TTCore/IR/Utils.cpp
wenbinlyuTT: The function is a lot simpler now.

test/ttmlir/Silicon/TTMetal/n150/simple_constant.mlir
wenbinlyuTT: Debug leftovers... Nice catch! Reverted

test/ttmlir/Silicon/TTMetal/n150/simple_add.mlir
wenbinlyuTT: ditto

lib/Dialect/TTCore/IR/TTCoreOpsTypes.cpp
```diff
@@ -530,6 +528,23 @@ unsigned SystemDescAttr::getPcieAddressAlignBytes(unsigned chipIndex) const {
   return getChipDesc(chipIndex).getPcieAddressAlignBytes();
 }
 
+unsigned ChipDescAttr::getDstLogicalSizeTiles(
+    Type type, bool fullSyncEn,
+    std::optional<unsigned> overridePhysicalSize) const {
+  auto nDstTiles = getDstPhysicalSizeTiles();
+  if (overridePhysicalSize.value_or(0) > 0) {
```
nsmithtt: If we're using `0` to mean unset then maybe there's no point in using an optional, I know that was my recommendation, but I think we should pick one or the other, either 0 means unset or nullopt means unset.

wenbinlyuTT: Went back to "0 means unset", as it's awkward to convert 0 to `nullopt` and do another check later.

test/ttmlir/Dialect/D2M/Transforms/insert_dst_register_access_matmul_tile.mlir
wenbinlyuTT: I've re-organized the tests and now we have all four combinations of `tile_matmul(true/false) x large_dst_for_f32(true/false)`.

lib/Dialect/D2M/Transforms/GenericLinearizeMemref.cpp
wenbinlyuTT: @nsmithtt @vwellsTT Let's start from a relatively big f32 matmul like
```
module {
  func.func @matmul(%arg0: tensor<512x1024xf32>, %arg1: tensor<1024x2048xf32>) -> tensor<512x2048xf32> {
    %0 = ttir.empty() : tensor<512x2048xf32>
    %1 = "ttir.matmul"(%arg0, %arg1, %0) <{transpose_a = false, transpose_b = false}> : (tensor<512x1024xf32>, tensor<1024x2048xf32>, tensor<512x2048xf32>) -> tensor<512x2048xf32>
    return %1 : tensor<512x2048xf32>
  }
}
```

The `d2m.generic` will have inputs & output like:
```
memref<8x8x2x4x!ttcore.tile<32x32, f32>, ...>, memref<8x8x4x8x!ttcore.tile<32x32, f32>, ...> -> memref<8x8x2x8x!ttcore.tile<32x32, f32>, ...>
```
And so the linalg generic ops will be handing matmul of `2x4 @ 4x8 -> 2x8`.

In `main`, after the loop tiling pass, we get the following subviews for the linalg loops:
```
%subview = memref.subview %cb0[%arg2, %arg4] [1, 4] [1, 1] : memref<2x4x!ttcore.tile<32x32, f32>, #ttcore.memory_space<l1>> to memref<1x4x!ttcore.tile<32x32, f32>, strided<[4, 1], offset: ?>, #ttcore.memory_space<l1>>
%subview_16 = memref.subview %cb1[%arg4, %arg3] [4, 8] [1, 1] : memref<4x8x!ttcore.tile<32x32, f32>, #ttcore.memory_space<l1>> to memref<4x8x!ttcore.tile<32x32, f32>, strided<[8, 1], offset: ?>, #ttcore.memory_space<l1>>
%subview_17 = memref.subview %cb2[%arg2, %arg3] [1, 8] [1, 1] : memref<2x8x!ttcore.tile<32x32, f32>, #ttcore.memory_space<l1>> to memref<1x8x!ttcore.tile<32x32, f32>, strided<[8, 1], offset: ?>, #ttcore.memory_space<l1>>
```
It thought the DST has 8 tiles so it's fine to do the entire `1x4 & 4x8 -> 1x8` in one go, all subviews look contiguous and nice.

With my change, the DST is only 4 tiles, and we end up with subviews like:
```
%subview = memref.subview %cb0[%arg2, %arg4] [1, 4] [1, 1] : memref<2x4x!ttcore.tile<32x32, f32>, #ttcore.memory_space<l1>> to memref<1x4x!ttcore.tile<32x32, f32>, strided<[4, 1], offset: ?>, #ttcore.memory_space<l1>>
%subview_16 = memref.subview %cb1[%arg4, %arg3] [4, 4] [1, 1] : memref<4x8x!ttcore.tile<32x32, f32>, #ttcore.memory_space<l1>> to memref<4x4x!ttcore.tile<32x32, f32>, strided<[8, 1], offset: ?>, #ttcore.memory_space<l1>>
%subview_17 = memref.subview %cb2[%arg2, %arg3] [1, 4] [1, 1] : memref<2x8x!ttcore.tile<32x32, f32>, #ttcore.memory_space<l1>> to memref<1x4x!ttcore.tile<32x32, f32>, strided<[8, 1], offset: ?>, #ttcore.memory_space<l1>>
```
The subview for CB 1&2 has 'gaps' in inner dim and so `memref::CollapseShapeOp::isGuaranteedCollapsible` returns false on it. The LinearizeMemref pass then asserts.

My changes here are suggested by Cursor and they look OK to me, not sure if there're better ways around this.

nsmithtt: @arminaleTT can you also take a look?

nsmithtt: It kinda seems like if we can support _not_ using CollapseShapeOp, perhaps we should just not do the linearization here anymore and just support it in ttkernel. idk, I'd like to get others thoughts.

arminaleTT: On the narrow question of "is it okay to move this code somewhere else, (e.g. D2MToTTKernel)?", yes it is totally fine to move/remove this. In the original draft implementation all of this was done in D2MToTTKernel. That pass originally received 2D subviews and in the end resolved linearized indices for the LLK call.

However, I do need to look at and think about the full IR here. This "gap" seems like blocking on K, leading us to have partial results. I'm not sure off the top of my head if matmul_bloc/our implementation of it handle that correctly. Let me get back to you on this

cc: @jdesousa-TT  @xanderchin

arminaleTT: Okay on further examination and running the matmul tests, this change works with our matmul block without issues.

Regarding the GenericLinearizeMemref changes, incorporating this code back into D2MToTTKernel requires the following:
- Deleting the `D2MLinearizeTileMatmulBlockRewriter`
- Reverting the D2MToTTKernel pass lowering for `TileMatmulBlockOp` to roughly [this](https://github.com/tenstorrent/tt-mlir/pull/4708/files/56cf77dd4e81a227381c9c939fd94cf53266d008#diff-8a13ad901f6fcf8dd44e3f47b6afd3e14eceaf313e05f6bf34f195599eb65990) state
  - Infer block dims and strides in that pass by inspecting the memref dims
  - Accept that 2d subviews will exist for `TileMatmulBlockOp` while other memrefs/subviews are linearized
- Remove the block dim parameters from the `TileMatmulBlockOp` and assorted verification

This seems a bit out of scope for this PR. I don't mind doing it after this is in.

nsmithtt: I guess I'm just confused, the way I read this PR is:
- if we cannot maximally fit the shard row into dst, then we do not perform shape collapse.

What's confusing to me is that this pass can just ignore collapse and it still "just works".  My question is, why collapse at all if we appear support lowering without collapse in main today?

arminaleTT: The op doesn't need collapsing. It only needs the block dims. The collapsing was done to unify the state of memrefs/subviews in the IR. The block index calculation downstream is able to handle 2D memrefs

nsmithtt: I see, then perhaps for this op we should unconditionally _not_ collapse.  I think conditionally doing the collapse is objectively worse than just picking one way over another.

wenbinlyuTT: So what's our plan of action here? Should I delete the `D2MLinearizeTileMatmulBlockRewriter` in this PR and let @arminaleTT do the follow-up PR? Or should I just merge this as-is?

arminaleTT: I think you should merge as is. Deleting the rewriter needs to coincide with some changes in D2MToTTKernel. I have filed an issue to myself and will take care of that: https://github.com/tenstorrent/tt-mlir/issues/5275

nsmithtt: I was thinking that we should just delete the `else` branch, so unconditionally we do:
```c++
      rewriter.replaceOpWithNewOp<TileMatmulBlockOp>(
          op, op.getA(), op.getB(), op.getOutput(),
          rewriter.getI64IntegerAttr(rtDim), rewriter.getI64IntegerAttr(ktDim),
          rewriter.getI64IntegerAttr(ctDim), rewriter.getI64IntegerAttr(ntDim));
```

nsmithtt: Get rid of `linA`/`libB`/`linO` path for `TileMatmulBlockOp`.

nsmithtt: We can still cleanup further w/ #5275, my main gripe was this conditional.

PR #5208, opened 2025-10-02
Enable the generic op block factor free dim check

nsmithtt:
The core change of this commit is enabling the check of free block factors (ones that don't participate in worker grid dims, like matmul inner dim).  This enforces that block factors are always in alignment with the input tensor grid shapes and they can now be directly used for loop nest creation.

This change also:
- Automatically infers block factors in the generic op defualt builder.
- Move `concatInversePermutationMap` to its own utility function since it's used in multiple places.
- Introduces new method `getLegalBlockFactorFactorizations`, which can be used by block factor / allocator analysis to choose legal factors from.


lib/Dialect/D2M/Utils/Utils.cpp
```diff
@@ -52,4 +52,28 @@ getSquareTargetGrid(mlir::ArrayRef<int64_t> targetGridShape) {
...
+  // (d0, d1, d2) -> (d2, d1)
+  // (d0, d1, d2) -> (d0, d1)
+  // Becomes:
+  // (d0, d1, d2) -> (d0, d2, d2, d1, d0, d1)
+  AffineMap concat =
+      concatAffineMaps(affineMaps, affineMaps.front().getContext());
+
+  // Invert the permutation to get a map that we can use to get the loop
+  // bounds. Above example becomes: (d0, d1, d2, d3, d4, d5) -> (d0, d3, d1)
+  return inversePermutation(concat);
```
vroubtsovTT: just reading the method comments where they are available, `inversePermutation()` requires input to be a valid permutation... this wouldn't necessarily be true of `concat` here...

nsmithtt: Where does it say this?  The docs for `inversePermulation` are:
```
/// Returns a map of codomain to domain dimensions such that the first codomain
/// dimension for a particular domain dimension is selected.
/// Returns an empty map if the input map is empty.
/// Returns null map (not empty map) if `map` is not invertible (i.e. `map` does
/// not contain a subset that is a permutation of full domain rank).
///
/// Prerequisites:
///   1. `map` has no symbols.
```

nsmithtt: I _think_ for example, this map `(d0, d1, d2, d3, d4, d5) -> (d0, d3, d1)` is not invertible because the results don't fully cover the domain, i.e. you have no mapping for dims `d2`, `d4` & `d5` so there's no way to generate some map that goes from rank 3 to rank 6.

lib/Dialect/D2M/IR/D2MOps.cpp
```diff
@@ -1236,6 +1339,34 @@ mlir::SmallVector<int64_t> d2m::GenericOp::getBlockFactorsValue() {
   });
 }
 
+mlir::SmallVector<int64_t> d2m::GenericOp::getLegalBlockFactorFactorizations() {
```
vroubtsovTT: a "factorization" of a number is a list of its primary factors... perhaps it's a question of terminology, but I don't understand what this method returns...

nsmithtt: I don't know what to call it, it returns a single number for each block factor, for which you can pick any of _this_ number's factors as a legal option.

e.g. this function might return `[1, 8, 10]`, where the user can then expand the entire solution space to be:
- `[[1], [1, 2, 4, 8], [1, 2, 5, 10]]`

nsmithtt: Open to naming suggestions.

nsmithtt: Maybe `getTotalBlockFactors` or something w/ a better comment? lmk what you think

lib/Dialect/D2M/Utils/Utils.cpp
vroubtsovTT: > Where does it say this? The docs for inversePermulation are:

Is there more than one? It says [right here](https://github.com/tenstorrent/tt-mlir/blob/71ebc63b150954f22f4d17ce8e9e928e8dfd76e2/include/ttmlir/Utils.h#L256):

```
// Returns a vector `inversePermutation`, such that
// inversePermutation[permutation[i]] = i, for all i. Assumes that permutation
// is a valid permutation of a range(0, permutation.size()). Example:
// permutation = [2, 0, 1] -> [1, 2, 0]
```

vroubtsovTT: Ah!, there is a `mlir::inversePermutation(AffineMap)` and `ttmlir::utils::inversePermutation(llvm::ArrayRef<int64_t>)`.  My mistake.

lib/Dialect/D2M/IR/D2MOps.cpp
vroubtsovTT: Naming is always a struggle. Trying to convey the fact that these are block factorizations that are _consistent_ across the entire operand block, i.e. they apply to the whole generic op as a whole...:
1. getAllOperandBlockFactors
2. getOpBlockFactors
3. getFullBlockFactors
4. getLegalBlockFactorizations
5. getLegalBlockFactorComposites // a "composite number" being a product of its prime factors

lib/Dialect/D2M/Utils/Utils.cpp
vroubtsovTT: Not sure if it feels the same to others, but I would prefix both `concatAffineMaps()` above and `inversePermutation()` with `mlir::` for clarity... otherwise it reads to me like calls to other tt utils methods. 

...nm I see this has been done everwhere else with AffineMap ops so no point fixing just this cast here...

lib/Dialect/D2M/IR/D2MOps.cpp
nsmithtt: I like 3 the best `getFullBlockFactors`, so I went with that! Thanks

lib/Dialect/D2M/Utils/Utils.cpp
nsmithtt: Good call, I updated the usages here, I agree using mlir:: is clearer.

Issue #4795, opened 2025-09-04
error: failed to legalize operation 'ttir.gather' that was explicitly marked illegal


abogdanovicTT:
TTIR:

```
%76 = "ttir.gather"(%41, %74, %75) <{collapsed_slice_dims = array<i64>, index_vector_dim = 1 : si64, indices_are_sorted = false, offset_dims = array<i64: 1, 2, 3, 4>, operand_batching_dims = array<i64>, slice_sizes = array<i64: 8, 54, 54, 5>, start_index_map = array<i64: 0, 1, 2, 3>, start_indices_batching_dims = array<i64>}> : (tensor<8x54x54x68xbf16>, tensor<64x4xi32>, tensor<64x8x54x54x5xbf16>) -> tensor<64x8x54x54x5xbf16>
```
Other examples from the same model have the following parameters and shapes:

```
%124 = "ttir.gather"(%27, %122, %123) <{collapsed_slice_dims = array<i64: 3>, index_vector_dim = 0 : si64, indices_are_sorted = false, offset_dims = array<i64: 0, 1, 2>, operand_batching_dims = array<i64>, slice_sizes = array<i64: 8, 54, 54, 1>, start_index_map = array<i64: 3>, start_indices_batching_dims = array<i64>}> : (tensor<8x54x54x64xbf16>, tensor<1xi32>, tensor<8x54x54xbf16>) -> tensor<8x54x54xbf16>
%209 = "ttir.gather"(%174, %207, %208) <{collapsed_slice_dims = array<i64>, index_vector_dim = 1 : si64, indices_are_sorted = false, offset_dims = array<i64: 1, 2, 3, 4>, operand_batching_dims = array<i64>, slice_sizes = array<i64: 8, 26, 26, 5>, start_index_map = array<i64: 0, 1, 2, 3>, start_indices_batching_dims = array<i64>}> : (tensor<8x26x26x196xbf16>, tensor<192x4xi32>, tensor<192x8x26x26x5xbf16>) -> tensor<192x8x26x26x5xbf16>
[...]

PR #5209, opened 2025-10-02
Supporting ttir.gather case from AlexNet model

sdjukicTT:
### Ticket
Closes https://github.com/tenstorrent/tt-mlir/issues/4795

### Problem description
This example has two _ that are not supported:
1. There are dims that are in start_index_map for which slice_size is full dim size, which means they aren't actually being indexed, they are taken whole
2. There is a dim in start_index_map for which slice_size is larger than 1, but not full, which means multiple 'rows' are selected with one index

### What's changed
Changed TTIR to TTIR decomposition to now supports cases where:
- for all but one dim in start_index_map, slice_size is full dim size, and the slice_size for the remaining dim can be anything less than full size. For dims that aren't in start_index_map, slice_size has to be full dim size.

The two _ are supported:
 1. When there are dims from start_index_map that are taken whole, they are removed from start_index_map. Start_indices are sliced (with `static slice op`) to leave only one index which applies to the actually indexed dimension. This is also taken into account when output of `embedding op` is permuted and reshaped, because only the 'real' indexed dim is moved to the front of the operand before `embedding op`.
 2. When there is a sliced indexed dim (dim in start_index_map for which slice_size is larger than 1 and less than full dim size), start_indices are expanded to include the implied indices as well. This is done by broadcasting start_indices from Nx1 to Nxslice_size, and adding a constant Nxslice_size tensor that has [0, 1, 2, ..., sliceSize-1] for every row. This is done after reshaping start_indices to make sure they are Nx1 shape.

### Checklist

### DIFF
--- a/lib/Conversion/TTIRToTTIRDecomposition/TTIRToTTIRDecomposition.cpp
+++ b/lib/Conversion/TTIRToTTIRDecomposition/TTIRToTTIRDecomposition.cpp
@@ -562,8 +562,10 @@ struct GatherToEmbeddingConversionPattern
    * transformation:
    * - start indices tensor isn't 1D when we are indexing multiple dims
    * - operandBatchingDims and startIndicesBatchingDims are none
-   * - startIndexMap = collapsedSliceDims
    * - sliceSizes are fullDim for dimensions we are not indexing
+   * - for dimensions we are indexing, sliceSizes must fit into one of:
+   *   - all sliceSizes are 1
+   *   - all sliceSizes are fullDim except one which can be anything
    */
 
   LogicalResult checkBasicLegality(ttir::GatherOp op,
@@ -576,7 +578,6 @@ struct GatherToEmbeddingConversionPattern
     // Get attributes needed for embedding op pattern matching checks.
     auto sliceSizes = op.getSliceSizes();
     auto startIndexMap = op.getStartIndexMap();
-    auto collapsedSliceDims = op.getCollapsedSliceDims();
 
     // Check if start indices tensor isn't 1D when we are indexing multiple
     // dimensions because of matmul restrictions.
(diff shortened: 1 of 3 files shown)


lib/Conversion/TTIRToTTIRDecomposition/TTIRToTTIRDecomposition.cpp
```diff
@@ -660,33 +659,78 @@ struct GatherToEmbeddingConversionPattern
...
+    // If there is an indexed dim with slice size > 1, but not full, we need to
+    // expand start indices to contain the implied ones.
+    bool needsExpansion = false;
+
+    // Create startIndexMap without dims for which sliceSizes[dim] =
+    // inputShape[dim]
+    llvm::SmallVector<int64_t> startIndexMap;
+    for (size_t i = 0; i < originalStartIndexMap.size(); ++i) {
+      int64_t dim = originalStartIndexMap[i];
+      if (sliceSizes[dim] != inputShape[dim]) {
```
pglusacTT: You can invert the logic here and denest the code

lib/Conversion/TTIRToTTIRDecomposition/TTIRToTTIRDecomposition.cpp
```diff
@@ -844,41 +973,50 @@ struct GatherToEmbeddingConversionPattern
...
-    auto expectedOutputType = expectedOutput.getType();
+  static ttir::PermuteOp
+  reshapeAndPermuteOutput(ConversionPatternRewriter &rewriter,
+                          ::mlir::TypedValue<::mlir::RankedTensorType> output,
+                          int64_t specialDim, ttir::GatherOp op) {
+    auto expectedOutputType = op.getOutput().getType();
     auto expectedOutputShape = expectedOutputType.getShape();
+    auto offsetDims = op.getOffsetDims();
+    // In case there were full indexed dims, we permuted input to move the
+    // actual indexed dims to the front, so is is before other offset dims in
```
pglusacTT: Typo

lib/Conversion/TTIRToTTIRDecomposition/TTIRToTTIRDecomposition.cpp
```diff
@@ -844,41 +973,50 @@ struct GatherToEmbeddingConversionPattern
...
+    // the output, as well.
+    bool specialCase = false;
```
pglusacTT: Can we use better naming than "specialCase" ?

PR #5215, opened 2025-10-03
Break L1 chains

rpavlovicTT:
Break L1 chains if current op is not first operand in the next op. We are not able to shard ops where producer is not first operand in the consumer op yet.


odjuricicTT (approved):
LGTM

PR #5217, opened 2025-10-03
Run ttrt tests with saving and restoring existing ttrt artifacts

nsumrakTT:
### Problem description
When running ttrt tests along with other tests it garbles tart-artifacts and make other tests fail.

### What's changed
Save existing ttrt-artifacts before test and restore after tests are finished.

### Checklist

### DIFF
--- a/tools/ttrt/test/conftest.py
+++ b/tools/ttrt/test/conftest.py
@@ -6,16 +6,24 @@ import os
 import pytest
 
 from util import *
+import shutil
 
 
 @pytest.fixture(scope="session", autouse=True)
 def session_setup():
+    # save previous ttrt-artifacts directory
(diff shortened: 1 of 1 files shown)


nsumrakTT:
@vmilosevic TTRT tests are using ttrt in various configs and destroy existing artifacts, mainly systemdesc used by other tests. So, it is not ttrt issue, just an issue with tests which is not seen when it is run a as single test. However, if you try running it with other tests, some test executed after these ttrt tests will fail.
Instead of opening issue, I've just fixed it so it preserves state as it was before running and cleanup afterward.

jgrimTT (approved):
Looks good! It'll be nice to not have to re-query the system desc every time I run `--clean-artifacts`

Issue #4351, opened 2025-08-05
Fix dst access w/ typecast op


nsmithtt:
We're thinking this can be resolved by introducing a new DstType from which you can take memref views of differing element types.

PR #5222, opened 2025-10-03
Fix typecast op dst access

ckaravasilisTT:
### Ticket
#4351

### Problem description
Dst acquired using original type, typecast result stored back to dst causes type mismatch. The same issue would show during storing back to L1.

### What's changed
`InsertDstRegisterAccess`: Utilized `builtin.unrealized_conversion_cast` to bypass the mismatch. An unrealized conversion cast is inserted after the main typecast operation before storing back to dst, and another cast is inserted before storing back to L1.
`D2MToTTKernel`: Modified load/store copy pattern to see through unrealized casts. Fixed minor bug in the TypecastRewriter that was using the converted input/output types (IndexType) instead of the original ones (TileType).

### DIFF
--- a/lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
+++ b/lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
@@ -287,7 +287,23 @@ public:
   LogicalResult
   matchAndRewrite(memref::StoreOp op, memref::StoreOpAdaptor adaptor,
                   ConversionPatternRewriter &rewriter) const final {
-    auto load = mlir::dyn_cast<memref::LoadOp>(op.getValue().getDefiningOp());
+    // Look for the load operation, potentially through an unrealized conversion
+    // cast
+    Operation *definingOp = op.getValue().getDefiningOp();
+    auto load = mlir::dyn_cast<memref::LoadOp>(definingOp);
+
+    // If not a direct load, check if it's an unrealized cast wrapping a load
+    if (!load && definingOp) {
+      if (auto unrealizedCast =
+              mlir::dyn_cast<UnrealizedConversionCastOp>(definingOp)) {
+        // Look through the unrealized cast to find the actual load
+        if (unrealizedCast.getNumOperands() == 1) {
+          load = mlir::dyn_cast_or_null<memref::LoadOp>(
+              unrealizedCast.getOperand(0).getDefiningOp());
+        }
+      }
+    }
+
     bool storeToDst = ttcore::getMemorySpace(op.getMemRef()) ==
                       ttcore::MemorySpace::RegisterDst;
 
@@ -295,6 +311,10 @@ public:
       // If we are coming from a load, then we are a copy tile. Pattern:
       //    %0 = memref.load %arg0, %c0 : memref<1x!tt.tile, l1>
       //    tt.store %0, %arg1, %c0 : memref<1x!tt.tile, dst>
+      // OR with unrealized cast:
+      //    %0 = memref.load %arg0, %c0 : memref<1x!tt.tile, l1>
+      //    %1 = unrealized_conversion_cast %0 : type1 -> type2
+      //    tt.store %1, %arg1, %c0 : memref<1x!tt.tile, dst>
       return lowerCopyTile(load, op, adaptor, rewriter);
     }
 
@@ -743,13 +763,15 @@ public:
   LogicalResult
   matchAndRewrite(Operation *op, ArrayRef<Value> operands,
                   ConversionPatternRewriter &rewriter) const final {
-    if (mlir::isa<d2m::TileTypecastOp>(op)) {
+    if (auto typecastOp = mlir::dyn_cast<d2m::TileTypecastOp>(op)) {
       rewriter.create<ttkernel::TypecastTileInitOp>(op->getLoc());
 
       auto inDtype =
-          mlir::cast<ttcore::TileType>(operands[0].getType()).getDataType();
-      auto outDtype = mlir::cast<ttcore::TileType>(op->getResult(0).getType())
-                          .getDataType();
+          mlir::cast<ttcore::TileType>(typecastOp.getInput().getType())
+              .getDataType();
+      auto outDtype =
+          mlir::cast<ttcore::TileType>(typecastOp.getResult().getType())
+              .getDataType();
       rewriter.create<ttkernel::TypecastTileOp>(
           op->getLoc(), i32(rewriter, op->getLoc(), 0), inDtype, outDtype);
     } else {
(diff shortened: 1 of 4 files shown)


ckaravasilisTT:
I'm not fully aware of the conventional use cases of `builtin.unrealized_conversion_cast`, but I see your point.
Would a new d2m op be used in the exact same way in this context, but simply communicate the intent better (i.e. 'dst' doesn't really have a 'type')?

nsmithtt:
> I'm not fully aware of the conventional use cases of `builtin.unrealized_conversion_cast`, but I see your point. Would a new d2m op be used in the exact same way in this context, but simply communicate the intent better (i.e. 'dst' doesn't really have a 'type')?

Yes exactly, and then we have more control over things like memory effects or other traits that might change the behavior of how certain passes can transform the IR.

xanderchin (approved):

nsmithtt:
Are we still planning on introducing a new op?

ckaravasilisTT:
Yes, I plan to first finalize the data type override pass I’ve been working on (using typecasts), and then introduce a new op to replace the unrealized cast in a follow-up PR.

nsmithtt:
> Yes, I plan to first finalize the data type override pass I’ve been working on (using typecasts), and then introduce a new op to replace the unrealized cast in a follow-up PR.

imo we should do the new op as part of this PR, should be relatively quick addition, unrealized_conversion_cast is really not intended to be used like this and could cause problems with upstream passes.

nsmithtt:
Relevant snippet of MLIR documentation (esp the last sentence):
> An `unrealized_conversion_cast` operation represents an unrealized
    conversion from one set of types to another, that is used to enable the
    inter-mixing of different type systems. This operation should not be
    attributed any special representational or execution semantics, and is
    generally only intended to be used to satisfy the temporary intermixing of
    type systems during the conversion of one type system to another.

Here we're not converting from a set of types to another, these are legitimate casts within the same dialect.  Think of `unrealized_conversion_cast` kind of like using an "llvm_unreachable" macro in C++, it's useful as a placeholder for a program state that you know is illegal.

nsmithtt (changes requested):
Let's intoduce an explicit dst casting op.

lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
```diff
@@ -290,14 +290,34 @@ class MemrefStoreRewriter : public OpConversionPattern<memref::StoreOp> {
...
+    // cast
+    Operation *definingOp = op.getValue().getDefiningOp();
+    auto load = mlir::dyn_cast<memref::LoadOp>(definingOp);
+
+    // If not a direct load, check if it's an unrealized cast wrapping a load
+    if (!load && definingOp) {
+      if (auto unrealizedCast =
+              mlir::dyn_cast<UnrealizedConversionCastOp>(definingOp)) {
+        // Look through the unrealized cast to find the actual load
+        if (unrealizedCast.getNumOperands() == 1) {
```
bgrady-tt: @ckaravasilisTT might be appropriate to assert here if num operands != 1

lib/Dialect/D2M/Transforms/InsertDstRegisterAccess.cpp
```diff
@@ -628,8 +654,18 @@ struct D2MInsertDstRegisterAccessRewriter final
 
       rewriter.setInsertionPointAfter(op);
 
+      // Insert unrealized conversion cast if compute result type differs from
+      // dst type
+      Value valueToStore = op->getResult(0);
+      if (valueToStore.getType() != dstType.getElementType()) {
+        valueToStore = rewriter
```
zoecarver: When we create this op, we also need to track it so that we can skip it below, otherwise we will replace the use of `op->getResult(0)` inside the cast itself, creating a dominance issue/circular dependency. On line 641/677, just check that owner != store **and** owner != cast. 

Minimal reproducer:
```
  func.func @typecast_chain(%arg0: memref<1x1x!ttcore.tile<32x32, f32>, #l1>, %arg1: memref<1x1x!ttcore.tile<32x32, f32>, #l1>) attributes {d2m.thread = #d2m.thread<compute>} {
    %c0 = arith.constant 0 : index
    %0 = affine.load %arg0[%c0, %c0] : memref<1x1x!ttcore.tile<32x32, f32>, #l1>
    %1 = "d2m.tile_typecast"(%0) : (!ttcore.tile<32x32, f32>) -> !ttcore.tile<32x32, f16>
    %2 = "d2m.tile_typecast"(%1) : (!ttcore.tile<32x32, f16>) -> !ttcore.tile<32x32, f32>
    affine.store %2, %arg1[%c0, %c0] : memref<1x1x!ttcore.tile<32x32, f32>, #l1>
    return
  }
```

lib/Dialect/D2M/Transforms/InsertDstRegisterAccess.cpp
```diff
@@ -628,8 +654,18 @@ struct D2MInsertDstRegisterAccessRewriter final
...
+                           .create<UnrealizedConversionCastOp>(
+                               loc, dstType.getElementType(), valueToStore)
+                           .getResult(0);
+      }
+
       auto storeOp = rewriter.create<affine::AffineStoreOp>(
-          loc, op->getResult(0), dst, storeMap, storeIndices);
+          loc, valueToStore, dst, storeMap, storeIndices);
```
zoecarver, line 673: If we cast this for storage, we also need to cast it back to the original op result type. When we store an f16 value to an f32 dst (for example), we cast it to f32 for storage. But when we load it back and replace uses, downstream ops still expect the original f16 type.

Minimal reproducer: 
```
  func.func @typecast_store_type_mismatch(%arg0: memref<1x1x!ttcore.tile<32x32, f32>, #l1>, %arg1: memref<1x1x!ttcore.tile<32x32, f16>, #l1>, %arg2: memref<1x1x!ttcore.tile<32x32, f32>, #l1>) attributes {d2m.thread = #d2m.thread<compute>} {
    %c0 = arith.constant 0 : index
    %0 = affine.load %arg0[%c0, %c0] : memref<1x1x!ttcore.tile<32x32, f32>, #l1>
    %1 = "d2m.tile_typecast"(%0) : (!ttcore.tile<32x32, f32>) -> !ttcore.tile<32x32, f16>
    affine.store %1, %arg1[%c0, %c0] : memref<1x1x!ttcore.tile<32x32, f16>, #l1>
    %2 = affine.load %arg1[%c0, %c0] : memref<1x1x!ttcore.tile<32x32, f16>, #l1>
    %3 = "d2m.tile_typecast"(%2) : (!ttcore.tile<32x32, f16>) -> !ttcore.tile<32x32, f32>
    affine.store %3, %arg2[%c0, %c0] : memref<1x1x!ttcore.tile<32x32, f32>, #l1>
    return
  }
```

lib/Dialect/D2M/Transforms/InsertDstRegisterAccess.cpp
ckaravasilisTT: Good catch!

zoecarver: BTW I have fixes for both of these here: https://github.com/tenstorrent/tt-mlir/pull/5379/commits/c4dc91b60dee424bc1fe019093cc26a331dc65f8

Issue #4946, opened 2025-09-15
Improve builder location setting


tapspatel:
### Is your feature request related to a problem? Please describe

1. Allow users to set their own locs
2. Improve the current autogenerated locations in builder by migrate the builder location analysis function to be called right after the op definition 

eg
```
def add():
    // get_location_from_calle_row_col
```

tapspatel:
this has been merged in

PR #5225, opened 2025-10-03
Improved builder location setting #4946

ddesaiTT:
### Ticket
https://github.com/tenstorrent/tt-mlir/issues/4946

### Problem description
Builder ops needed better location tracking

### What's changed
- Improved location tracking with accurate caller info (file, line, col)
- Added support for manual location setting via `loc` parameter
- Improved call stack handling to find actual caller location
- Added test coverage for manual location setting

### DIFF
--- a/tools/builder/base/builder.py
+++ b/tools/builder/base/builder.py
@@ -305,19 +305,41 @@ class Builder:
 
     def _get_loc_of_extra_file_callee(self, id: int = 0) -> Location:
         stack = inspect.stack()
-        caller_filename = stack[1].filename
 
-        while len(stack) > 0 and stack[0].filename == caller_filename:
-            stack = stack[1:]
+        # Skip frames from builder files to find the actual caller
+        builder_files = {
+            "builder.py",
+            "ttir_builder.py",
+            "stablehlo_builder.py",
+            "d2m_builder.py",
+        }
+        caller_frame = None
 
-        if len(stack) == 0:
+        for frame in stack:
+            if not any(
+                builder_file in frame.filename for builder_file in builder_files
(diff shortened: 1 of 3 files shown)


tools/builder/ttir/ttir_builder.py
```diff
@@ -84,6 +84,15 @@ def _op_proxy(
         loc: Optional[Union[str, Location]] = None,
         skip_golden: bool = False,
     ) -> Any:
+
+        # Prepare location for the op
+        id = self._get_next_global_id()
+        loc = (
+            self._get_loc_from_str(loc)
+            if loc is not None
+            else self._get_loc_of_extra_file_callee(id=id)
```
tapspatel, line 93: migrate this to call after each ttir op so we get the callee of the ttir op

tools/builder/base/builder.py
```diff
@@ -305,19 +305,41 @@ def _get_next_global_id(self) -> int:
...
     def _get_loc_of_extra_file_callee(self, id: int = 0) -> Location:
         stack = inspect.stack()
-        caller_filename = stack[1].filename
 
-        while len(stack) > 0 and stack[0].filename == caller_filename:
-            stack = stack[1:]
-
-        if len(stack) == 0:
+        # Skip frames from builder files to find the actual caller
+        builder_files = {
```
tapspatel, line 310: lets no hardcode these file names. Instead, we should move this functionality to be called after each builder op, and lookup the stack frame before that op was called, so you don't have nested stack frames you have to parse like this

tapspatel:
@ctodTT can you also review this pr?

PR #5229, opened 2025-10-05
TTNN builder

mtopalovicTT:
This adds minimal changes for ttnn builder. With new LLM ops we are adding in TTNN this builder is needed asap in order to test fusing. I've only added minimal changes for this builder, in order to run eltwise binary op, so I omitted stuff like `dump_module`, `print_ir` and other options which are not used.

Biggest difference between ttnn_builder and ttir_builder is how we create tensors where ttnn tensors require encoding attribute and ttir tensors don't.

### DIFF
--- a/tools/builder/base/builder_utils.py
+++ b/tools/builder/base/builder_utils.py
@@ -10,7 +10,7 @@ from typing import Callable, List, Optional, Tuple, Union, Literal, Dict
 from collections import OrderedDict
 
 from ttmlir.ir import *
-from ttmlir.dialects import func, ttcore
+from ttmlir.dialects import func, ttcore, ttnn
 from ttmlir.passmanager import PassManager
 from ttmlir.passes import (
     tt_populate_argument_types,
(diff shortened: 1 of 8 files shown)


tools/builder/base/builder_utils.py
```diff
@@ -666,6 +667,181 @@ def compile_ttir_to_flatbuffer(
         raise TTBuilderCompileException(e)
 
 
+def build_ttnn_module(
```
mtopalovicTT, line 1144: This method and build_ttir_module can be merged into single method since they are mostly the same except the tensor creation part.

jgrimTT: Yeah you're right, that's true of all of the build module functions. I made an [issue](https://github.com/tenstorrent/tt-mlir/issues/5265) for it

jgrimTT (approved):
Looks great! If the test is CI ready, you can add `ttnn-builder-artifacts` to `.github/test_scripts/builder.sh`

mtopalovicTT:
> Looks great! If the test is CI ready, you can add `ttnn-builder-artifacts` to `.github/test_scripts/builder.sh`

Done

PR #5234, opened 2025-10-06
[ttnn.jit] build rpath fix

vtangTT:
### Ticket
Link to Github Issue

### Problem description
`import ttnn_jit` wouldn't be able to find the metal libs. This problem was hidden bc we always `import ttnn` before `ttnn_jit` so the metal libs were already found

### What's changed
- added a `TT_METAL_LIBRARY_DIR` to `BUILD_RPATH` of `TTMLIRRuntime`.
- should only affect local builds; the wheel build was never broken in this aspect.


runtime/python/CMakeLists.txt
```diff
@@ -22,7 +22,7 @@ nanobind_add_module(
 
 set_target_properties(_ttmlir_runtime PROPERTIES
   INSTALL_RPATH "$ORIGIN"
-  BUILD_RPATH "$ORIGIN;$<TARGET_FILE_DIR:TTMLIRRuntime>"
+  BUILD_RPATH "$ORIGIN;$<TARGET_FILE_DIR:TTMLIRRuntime>:${TTMETAL_LIBRARY_DIR}"
```
jnie-TT: Nit: Should this be semicolon-separated `;${TTMETAL_LIBRARY_DIR}` instead?

vtangTT: yes, I also added the rpath to the wrong lib 🤦

Issue #5031, opened 2025-09-20
D2M/TTNN Integration - Support setting buffer ptr in CBDescriptor


vtangTT:
Need to be able to set [this](https://github.com/tenstorrent/tt-metal/blob/ecaab1191635177c2961b8d0b5367bff8efa3122/tt_metal/api/tt-metalium/program_descriptors.hpp#L62) field when creating ttnn.generic in `TTIRToTTNNGeneric` pass and wire it through the runtime.

Should be an optimization.

vtangTT:
github.com/tenstorrent/tt-mlir/pull/5235 closes

jojo-771771:
I can work on this. Please assign!

jojo-771771:
I can work on this. Please assign!

jojo-771771:
I can work on this. Please assign!

jojo-771771:
I can work on this. Please assign!

PR #5235, opened 2025-10-06
D2M/TTNN Integration - alias buffer in ttnn.generic op

vtangTT:
### Ticket
https://github.com/tenstorrent/tt-mlir/issues/5031

### Problem description
Right now for `ttnn-mode` we are always streaming both inputs and outputs since we can't alias buffers in the ttnn.generic op.

### What's changed
- wire the buffer pointer from `D2MToTTNN` pass to flatbuffer and mlir runtime the same way we pass tensor buffer address: as an index that will choose the correct tensor from `io_tensors`
    - add `TTNN_KernelCBGlobalBufferAddressOfTensorAttr` to `TTNN_KernelCBAttr` as an optional
    - add `KernelGlobalCBIndexOfTensor` to `generic_op.fbs`
    - relevant changes to flatbuffer conversion and mlir runtime
- change `InsertStreams` pass to only insert streams for inputs that need to be streamed from DRAM.
    - never stream output, but wired up `allowOutputSpilling` for the future when we could spill to DRAM.
- modified existing lit tests
### Checklist
- [x] New/Existing tests provide coverage for changes -> e2e jit tests will now never be streamed, qualifying this change.

### DIFF
--- a/include/ttmlir/Target/TTNN/operations/generic_op.fbs
+++ b/include/ttmlir/Target/TTNN/operations/generic_op.fbs
@@ -35,7 +35,7 @@ table KernelCBFormat {
 }
 
 table KernelGlobalCBIndexOfTensor {
-  tensor_index: uint32;
+  tensor_operand_index: uint32;
 }
 
 table KernelCBDescriptor {
--- a/include/ttmlir/Target/TTNN/operations/generic_op.fbs
+++ b/include/ttmlir/Target/TTNN/operations/generic_op.fbs
@@ -35,7 +35,7 @@ table KernelCBFormat {
 }
 
 table KernelGlobalCBIndexOfTensor {
-  buffer_index: uint32;
+  tensor_index: uint32;
 }
 
 table KernelCBDescriptor {
--- a/include/ttmlir/Target/TTNN/operations/generic_op.fbs
+++ b/include/ttmlir/Target/TTNN/operations/generic_op.fbs
@@ -34,10 +34,15 @@ table KernelCBFormat {
   page_size: uint32;
 }
 
+table KernelGlobalCBIndexOfTensor {
+  buffer_index: uint32;
+}
+
 table KernelCBDescriptor {
   total_size: uint32;
   core_range: tt.target.ttnn.CoreRangeSet;
   formats: [KernelCBFormat];
+  buffer: KernelGlobalCBIndexOfTensor;
 }
 
 table ComputeKernelConfig {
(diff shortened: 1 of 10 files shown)


include/ttmlir/Target/TTNN/operations/generic_op.fbs
```diff
@@ -34,10 +34,15 @@ table KernelCBFormat {
   page_size: uint32;
 }
 
+table KernelGlobalCBIndexOfTensor {
+  tensor_index: uint32;
```
nsmithtt: Maybe `operand_index` might be slightly clearer.

nsmithtt (approved):

include/ttmlir/Target/TTNN/operations/generic_op.fbs
vtangTT: I have no strong opinions on this. Any thoughts @phizalev-TT

phizalev-TT: Maybe let's make it super clear: `tensor_operand_index`. Also in `KernelArgBufferAddressOfTensor`.

phizalev-TT (approved):

include/ttmlir/Target/TTNN/operations/generic_op.fbs
vtangTT: changed 👍

jnie-TT (approved):

lib/Conversion/D2MToTTNN/D2MToTTNN.cpp
```diff
@@ -168,8 +172,20 @@ class D2MGenericRewriter : public OpConversionPattern<d2m::GenericOp> {
 
       ttnn::KernelCBFormatAttr cbFormat =
           ttnn::KernelCBFormatAttr::get(ctx, i, dtype, pageSize);
-      ttnn::KernelCBAttr cbDescriptor = ttnn::KernelCBAttr::get(
-          ctx, numPages * pageSize, coreRangeSet, {cbFormat});
+
+      ttnn::KernelCBGlobalBufferAddressOfTensorAttr globalCBIndexOfTensor;
+      if (auto castOp = mlir::dyn_cast_if_present<ttir::TTNNMetalLayoutCastOp>(
+              cb.getDefiningOp())) {
+        // Not streamed, thus buffer is aliased.
```
vwellsTT: nit: I don't think this comment is immediately clear.  I guess our style guide says we should use proper sentences, and in this case it's not clear to me what the "subject" is--what is "not streamed"?  Does this mean: "We know the buffer is not stream, so it must be an alias", or something else?

vwellsTT (approved):

PR #5255, opened 2025-10-07
Use generic LLVM targets

alexbaden:
Updates the `LLVMToDynamicLib` target to use generic LLVM target initializers. The correct target is linked based on the architecture at build time.


alexbaden:
I think the build failures were happening because calling `InitializeAllTargets` with `LLVM_TARGETS_TO_BUILD` property not set expects all targets to be available. I changed the initialization calls to only initialize the native target. We could also add `-DLLVM_TARGETS_TO_BUILD="host"` to the LLVM build in `env/CMakeLists.txt` to avoid building targets that we don't need if you are building LLVM on each deployed host and only need the native host target.

PR #5256, opened 2025-10-07
Make the LLVM build optional during env setup

alexbaden:
Adds a new CMAKE flag `TTMLIR_BUILD_LLVM` which makes the LLVM build step optional. This is useful for overriding with out-of-tree built LLVM (for integration into third party projects or for building against a debug version of LLVM, for example). When building with third party LLVM I was not able to find `flatbuffers.h` - I believe this is because the include path for the toolchain directory is not explicitly included and LLVM includes were ensuring `flatbuffers.h` was found. I explicitly added an include for the toolchain directory and the problem went away.


nsmithtt:
@alexbaden this build-image pipeline failure is something we're working on internally to sort out.  Since this is an environment change it triggers a rebuild of our docker which runs into a weird perm issue w/ external PR's, we're taking a look at a fix!

nsmithtt:
Seems like it's actually intertwined with the way that GitHub managers dockers, unfortunately I think we might have to commandeer this one and repost as an internal PR.

nsmithtt:
This PR has moved here due to pipeline permission issues:
- https://github.com/tenstorrent/tt-mlir/pull/5273

PR #5257, opened 2025-10-07
Use TTMLIR_VENV_DIR when setting venv path

alexbaden:
Previously the PATH was set assuming the venv directory lives inside `TTMLIR_TOOLCHAIN`.

### DIFF
--- a/env/activate
+++ b/env/activate
@@ -22,7 +22,7 @@ elif [ -z "${_ACTIVATE_SUPPRESS_INIT_WARNING}" ]; then
 fi
 
 export TTMLIR_ENV_ACTIVATED=1
-export PATH="$(pwd)/${BUILD_DIR:=build}/bin:$TTMLIR_TOOLCHAIN_DIR/bin:$TTMLIR_TOOLCHAIN_DIR/venv/bin:$PATH"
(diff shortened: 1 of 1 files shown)


env/activate
```diff
@@ -22,8 +22,7 @@ elif [ -z "${_ACTIVATE_SUPPRESS_INIT_WARNING}" ]; then
 fi
 
 export TTMLIR_ENV_ACTIVATED=1
-export PATH="$(pwd)/${BUILD_DIR:=build}/bin:$TTMLIR_TOOLCHAIN_DIR/bin:$TTMLIR_TOOLCHAIN_DIR/venv/bin:$PATH"
-export TT_METAL_HOME="$(pwd)/third_party/tt-metal/src/tt-metal"
+export PATH="$(pwd)/${BUILD_DIR:=build}/bin:$TTMLIR_TOOLCHAIN_DIR/bin:$TTMLIR_VENV_DIR/bin:$PATH"export TT_METAL_HOME="$(pwd)/third_party/tt-metal/src/tt-metal"
```
nsmithtt: I think there might be an accidental join of lines 25 & 26.

alexbaden: Fixed - thanks!

nsmithtt (approved):
Awesome thanks!

PR #5269, opened 2025-10-08
Adding more emitPy conversions

sdjordjevicTT:
### Ticket
Part of the https://github.com/tenstorrent/tt-mlir/issues/4930

### Problem description
Some emitPy op conversions were missing.

### What's changed
Added the following emitpy op conversions:
- Arange and Pad ops
- Relu and Silu ops
- Quantization ops
- LLM ops
- MorehCumSum op
- Prod op
- EmbeddingBackward op
- ToMemoryConfig op

With this PR, we are at 91% completeness for all required emitPy conversions.

### Checklist

### DIFF
--- a/lib/Conversion/TTNNToEmitPy/TTNNToEmitPy.cpp
+++ b/lib/Conversion/TTNNToEmitPy/TTNNToEmitPy.cpp
@@ -2520,6 +2520,7 @@ public:
         mlir::tt::ttnn::ScaledDotProductAttentionDecodeOp>
         emitter(srcOp, adaptor, rewriter);
 
+    // NOLINTBEGIN(clang-analyzer-cplusplus.NewDelete)
     llvm::SmallVector<mlir::Attribute> args{
         emitter.emit(srcOp.getQuery()),
         emitter.emit(srcOp.getKey()),
@@ -2533,6 +2534,7 @@ public:
                          emitter.getMemoryConfig(srcOp.getResult()),
                      "memory_config"),
     };
+    // NOLINTEND(clang-analyzer-cplusplus.NewDelete)
 
     emitter.replaceOp(*this, args);
 


include/ttmlir/Conversion/TTNNToEmitPy/EmitPyConversion.h
```diff
@@ -683,6 +689,14 @@ struct EmitPyContainerTypeConverter {
     return convert(result);
   }
 
+  static std::string convert(const T &container) {
+    llvm::SmallVector<std::string> result;
+    for (const auto &element : container) {
+      result.push_back(EmitPyTypeConverter<value_type>::convert(element));
+    }
+    return convert(result);
+  }
```
svuckovicTT: This seems like a catch-all instead of "catch all container types".

lib/Conversion/TTNNToEmitPy/TTNNToEmitPy.cpp
```diff
@@ -1861,22 +2068,26 @@ class LoadCachedOpConversionPattern
...
-    mlir::Value tensorsInList =
-        rewriter
-            .create<emitpy::CallOpaqueOp>(
-                loadCachedOp.getLoc(),
-                emitpy::OpaqueType::get(rewriter.getContext(), "[ttnn.Tensor]"),
-                ttnn_to_emitpy::kCreateListFunctionName, adaptor.getOperands(),
-                nullptr)
-            ->getResult(0);
+    llvm::SmallVector<Value> operands;
+    if (loadCachedOp.getInputs().size() > 0) {
```
svuckovicTT, line 2080: I wonder why we don't need this conditional in emitc path...

lib/Dialect/TTNN/Transforms/Passes.cpp
```diff
@@ -26,6 +26,7 @@
 #include "llvm/ADT/SmallVector.h"
 #include "llvm/ADT/StringRef.h"
 #include "llvm/Support/Casting.h"
+#include <unordered_map>
```
svuckovicTT:
```suggestion

#include <unordered_map>
```

lib/Dialect/TTNN/Transforms/Passes.cpp
```diff
@@ -190,18 +191,25 @@ class TTNNInputFunctionCreatorBase {
       return mlir::WalkResult::advance();
     });
 
-    // Iterate over all func ops and add input tensor functions.
+    // Iterate over all func ops and add input tensor functions if needed.
     //
-    llvm::SmallVector<mlir::func::FuncOp, 1> inputFuncOps;
+    llvm::SmallVector<std::pair<mlir::func::FuncOp, mlir::func::FuncOp>, 1>
+        forwardAndInputFuncOps;
```
svuckovicTT, line 195: I see that input/fwd funcs are now packed into pairs, meaning we don't rely on corresponding pairs being in the same order in their respective vectors - were there issues with that approach?

lib/Dialect/TTNN/Transforms/Passes.cpp
```diff
@@ -293,18 +302,24 @@ class TTNNInputFunctionCreatorBase {
       rewriter.setInsertionPointToStart(mainFuncOp.addEntryBlock());
     });
 
-    for (auto [forwardFuncOp, inputFuncOp] :
-         llvm::zip_equal(forwardFuncOps, inputFuncOps)) {
+    for (auto forwardAndInputFuncOp : forwardAndInputFuncOps) {
```
svuckovicTT: Prefer unpacking here
```suggestion
    for (auto [forwardFuncOp, inputFuncOp] : forwardAndInputFuncOps) {
```

svuckovicTT (approved):

include/ttmlir/Conversion/TTNNToEmitPy/EmitPyConversion.h
sdjordjevicTT: I need to double check, I think I don't need this one...

lib/Conversion/TTNNToEmitPy/TTNNToEmitPy.cpp
sdjordjevicTT: I think I caught this with Arange op, which isn't supported in EmitC.

I used this dialect test, and you can see that this specific test creates consteval function which doesn't have input args, hence the conversion wasn't working because it expected some tensors to be passed to the consteval func:
```mlir
  func.func @arange_multiple_users() -> (tensor<1x1024x1xi32>, tensor<1x1x1024xi32>) {
    %0 = "ttir.arange"() <{arange_dimension = 0 : i64, end = 1024 : si64, start = 0 : si64, step = 1 : si64}> : () -> tensor<1024xi32>
    %1 = ttir.empty() : tensor<1x1024x1xi32>
    %2 = "ttir.reshape"(%0, %1) <{shape = [1 : i32, 1024 : i32, 1 : i32]}> : (tensor<1024xi32>, tensor<1x1024x1xi32>) -> tensor<1x1024x1xi32>
    %3 = ttir.empty() : tensor<1x1x1024xi32>
    %4 = "ttir.reshape"(%0, %3) <{shape = [1 : i32, 1 : i32, 1024 : i32]}> : (tensor<1024xi32>, tensor<1x1x1024xi32>) -> tensor<1x1x1024xi32>
    // Verify that the arange is op in IR is before both reshape ops
    // CHECK: "ttnn.arange"
    // CHECK: "ttnn.reshape"
    // CHECK: "ttnn.reshape"
    return %2, %4: tensor<1x1024x1xi32>, tensor<1x1x1024xi32>
  }
```

lib/Dialect/TTNN/Transforms/Passes.cpp
sdjordjevicTT: This fix here solves the issue when we have a forward function without arguments. I could keep two arrays, but I opted for a more structured way by keeping pairs.

So here the code got changed into the pairs of:
<ForwardFunc, InputGeneratorFunc>

For example:
<func1, func1g>, <func2, nullptr>, <func3, func3g>

Instead of having this:
ForwardFuncs:              | func1   | func2  | func3   |
InputGeneratorFuncs:  | func1g | nullptr | func3g |

I can revert to having two arrays if you see that as a better approach, not that I am 100% convinced that this is better... :)

include/ttmlir/Conversion/TTNNToEmitPy/EmitPyConversion.h
sdjordjevicTT: Rewrote the conversion; actually didn't need this.

lib/Dialect/TTNN/Transforms/Passes.cpp
svuckovicTT: Ah, cool. No need to revert, this is okay as well :)

PR #5274, opened 2025-10-08
Integrate Flatbuffer Execution Into Builder `pytest`s

### LINKED ISSUE #4094: Integrate ttrt into builder testing flow

### LINKED ISSUE #4945: Add feature to allow user to set their own testing parameters

example
- expected_pcc
- atol
- rtol
- specific golden functions (MAE, root mean square error, cosine similarity) 

_No response_

### Describe alternatives you've considered

_No response_

### Additional context

_No response_

ctodTT:
### Ticket
Closes #4945 
Closes #4094 

### Problem description
The flow before this change was to `pytest test/python/golden` to generate flatbuffers, and then separately run `ttrt run` on the generated artifacts to test runtime and golden. This is cumbersome and is prone to leaving false negative test exclusion from what `ttrt` runs. It also fragments the test report into two completely different artifacts, with completely different schemas

### What's changed
Now, each test under `test/python/golden` will execute its own flatbuffer for runtime and golden correctness checks. The failures will be correctly logged in the `failure_stage` property of each test within the JUnit XML that is produced by the test run.

### DIFF
--- a/test/python/golden/test_metal_matmul.py
+++ b/test/python/golden/test_metal_matmul.py
@@ -162,11 +162,8 @@ def test_matmul_ttnn_shapes_single_buffered(
     )
 
 
-<<<<<<< HEAD
 @pytest.mark.skip_config(["ttmetal", "p150"], reason="See issue #5341")
-=======
 @pytest.mark.skip_config(["p150"], ["p300"])
->>>>>>> 17e091504 (Review and fix some `p150` tests)
 @pytest.mark.xfail(reason="fails golden")
 @pytest.mark.parametrize(
     "shape",


runtime/lib/runtime.cpp
```diff
@@ -144,6 +145,28 @@ void setCurrentDeviceRuntime(const DeviceRuntime &runtime) {
...
+    return RuntimeContext::instance().setCurrentDeviceRuntime(
+        DeviceRuntime::TTNN);
+  }
+#endif
+
+#if defined(TT_RUNTIME_ENABLE_TTMETAL) && (TT_RUNTIME_ENABLE_TTMETAL == 1)
+  if (lowerRuntime == "ttmetal") {
+    return RuntimeContext::instance().setCurrentDeviceRuntime(
+        DeviceRuntime::TTMetal);
+  }
```
nsmithtt: Might be slightly cleaner to convert the string to its enum value and calling `setCurrentDeviceRuntime` enum overload of this func.

tools/builder/base/builder_utils.py
```diff
@@ -165,6 +180,68 @@ def get_metal_tensor_layout(
...
+    """
+    mlir_path = compile_fn(
+        target=target,
+        **compile_kwargs,
+    )
+
+    if skip_exec:
+        raise TTBuilderRuntimeException("Manually skipped execution")
+
+    fb_path = mlir_path + "." + ("ttnn" if target == "ttnn" else "ttm")
```
nsmithtt, line 229: Let's file an issue to enable passing a fbb via shared ptr, I'll have an example on how to do this soon.  This way we don't have to bounce it to disk.

ctodTT: Sounds like a plan, will file!

runtime/lib/runtime.cpp
ctodTT: Ah that's not a bad idea, lemme do that rq

.github/test_matrix/pr-push-matrix.json
```diff
@@ -25,9 +25,9 @@
 
   { "runs-on": ["n150","llmbox","n300","p150"],
         "image": "tracy",
-            "script": "builder.sh", "args": [ "test/python/golden", "not run_error and not fails_golden", "run-ttrt" ] },
+            "script": "builder.sh", "args": [ "test/python/golden", "", "" ] },
```
nsumrakTT: no need for "" or [] if single argument is passed. For example:
`"script": "builder.sh", "args": "test/python/golden"`
is fine.

ctodTT: Done!

Issue #5213, opened 2025-10-02
[ttnn.jit] Translate DRAM Interleaved Tensors


arminaleTT:

arminaleTT:
Accomplished between #5277 #5433 and #5693

PR #5277, opened 2025-10-08
[ttnn.jit] Initial Support for DRAM Interleaved Tensors

arminaleTT:
### Ticket
https://github.com/tenstorrent/tt-mlir/issues/5213

### Problem description
TTNN JIT needs to support tensors with various data layouts, including DRAM interleaved.

### What's changed
This PR adds initial support for translating DRAM tensors in the TTNN JIT frontend. 

Tensors that aren't sharded are assumed to be DRAM interleaved and a corresponding TTNNLayoutAttr is created. The D2M backend already supports ingesting and translating DRAM interleaved tensors. Disabled too-strict checking of memspaces in the D2M generic op (@bgrady-tt). Added tests to verify unary ops on DRAM tensors.

Limitations:
- The optional `memory_config` argument available on many TTNN Ops to select the `memory_config` for the output tensor is not supported.
  - i.e. `ttnn.add(tensor_a, tensor_b, memory_config=ttnn.DRAM_MEMORY_CONFIG)` will still produce an output tensor that matches `tensor_a`. This is a limitation of the AST parsing approach. More details in https://github.com/tenstorrent/tt-mlir/issues/5280
  - This means that we cannot have operations where operands are in one memory space and the output is in another memory space. Operands and outputs are either all in L1 or all in DRAM
- Operations on DRAM interleaved tensors only run on a single core
  - This is a limitation resulting from how layouts and grids are represented in D2M core, as well as how `TTNNLayoutAttr` represents DRAM interleaved layouts. More details in https://github.com/tenstorrent/tt-mlir/issues/5279

### Checklist

### DIFF
--- a/tools/ttnn-jit/_src/ttir_ast.py
+++ b/tools/ttnn-jit/_src/ttir_ast.py
@@ -139,7 +139,9 @@ class TTIRCompiler(ast.NodeVisitor):
             )
             return ttnn_layout
         else:
-            # DRAM interleaved tensors must have 1x1 grid in TTNNLayoutAttr
+            assert (
+                self.max_grid[0] == 0 and self.max_grid[1] == 0
+            ), "The grid for DRAM interleaved tensors is always 1x1"
             buffer_type = ttnn.ir.BufferTypeAttr.get(self.ctx, ttnn.BufferType.DRAM)
             grid = ttcore.ir.GridAttr.get(self.ctx, [1, 1])
             shape = [tensor_arg.shape[0] // 32, tensor_arg.shape[1] // 32]


tools/ttnn-jit/_src/ttir_ast.py
```diff
@@ -104,41 +104,54 @@ def _create_get_device_op(self):
...
+                self.ctx,
+                identity_map,
+                grid,
+                memref,
+                ttnn.TensorMemoryLayout.BlockSharded,
+                None,
+            )
+            return ttnn_layout
+        else:
+            # DRAM interleaved tensors must have 1x1 grid in TTNNLayoutAttr
```
vtangTT: can we assert max_grid here for 1x1, just so ppl don't pass 8x8 as max_grid and not know we are hard coding 1x1 here

Issue #5009, opened 2025-09-18
[EmitPy] Enable emitPy tests generated from builder to run in CI 


tapspatel:
Investigate how we can build ttnn python whls from tt-metal and how to enable emitPy tests generated from builder to run with it in CI.

jojo-771771:
I can work on this. Please assign!

jojo-771771:
I can work on this. Please assign!

jojo-771771:
I can work on this. Please assign!

jojo-771771:
I can work on this. Please assign!

jojo-771771:
I can work on this. Please assign!

PR #5288, opened 2025-10-09
Implementing EmitPy CI tests in ttrt

jgrimTT:
### Ticket
Closes [#5009](https://github.com/tenstorrent/tt-mlir/issues/5009)

### Problem description
EmitPy tests don't run in CI

### What's changed
Added ttnn python package path to `sys.path` in `emitpy.py` to allow for `import ttnn`.
Added `ttrt emitpy` tests for `test/ttmlir/EmitPy`. 
Added `ttrt emitpy` tests to `builder.sh` for builder artifacts.

### Checklist


.github/test_scripts/emitpy.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+# SPDX-FileCopyrightText: (c) 2024 Tenstorrent AI ULC
```
jmcgrathTT: Let's switch this to 2025

jgrimTT:
Regarding the first point @sdjordjevicTT, makes total sense, I'll remove that!

About the output comparison, yes we can do that at the moment, it would take a bit of configuration because of the directory/filenames setup. I was thinking about just leaving it alone for now because when the [runtime integration into builder pytest PR](https://github.com/tenstorrent/tt-mlir/pull/5274) lands, I'll start working on migrating emitc and emitpy builder tests (similarly to how that PR works) to be executed in builder instead of ttrt. Once that is done, we can start playing around with golden checks instead of flatbuffer output comparisons. Lmk if you'd still like that temporarily implemented :)

sdjordjevicTT:
> Regarding the first point @sdjordjevicTT, makes total sense, I'll remove that!
> 
> About the output comparison, yes we can do that at the moment, it would take a bit of configuration because of the directory/filenames setup. I was thinking about just leaving it alone for now because when the [runtime integration into builder pytest PR](https://github.com/tenstorrent/tt-mlir/pull/5274) lands, I'll start working on migrating emitc and emitpy builder tests (similarly to how that PR works) to be executed in builder instead of ttrt. Once that is done, we can start playing around with golden checks instead of flatbuffer output comparisons. Lmk if you'd still like that temporarily implemented :)

Thanks for the explanation, it makes total sense!

After this PR lands, I will start onboarding existing non-emitpy builder tests. If I am not mistaken, I will just need to add the following to become testable in CI, right?
`@pytest.mark.parametrize("target", [...,"emitpy",...])`

jgrimTT:
> > Regarding the first point @sdjordjevicTT, makes total sense, I'll remove that!
> > About the output comparison, yes we can do that at the moment, it would take a bit of configuration because of the directory/filenames setup. I was thinking about just leaving it alone for now because when the [runtime integration into builder pytest PR](https://github.com/tenstorrent/tt-mlir/pull/5274) lands, I'll start working on migrating emitc and emitpy builder tests (similarly to how that PR works) to be executed in builder instead of ttrt. Once that is done, we can start playing around with golden checks instead of flatbuffer output comparisons. Lmk if you'd still like that temporarily implemented :)
> 
> Thanks for the explanation, it makes total sense!
> 
> After this PR lands, I will start onboarding existing non-emitpy builder tests. If I am not mistaken, I will just need to add the following to become testable in CI, right? `@pytest.mark.parametrize("target", [...,"emitpy",...])`

Yep!

.github/test_scripts/emitpy.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+# SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+
```
nsumrakTT: Please add
`set -e -o pipefail`

.github/test_matrix/pr-push-matrix.json
```diff
@@ -34,5 +34,6 @@
   { "runs-on": "n150",   "image": "tracy",  "script": "ttnn_runtime.sh", "args": "runtime/test/ttnn/python/device_agnostic" },
   { "runs-on": "n300",   "image": "tracy",  "script": "runtime_debug.sh" },
   { "runs-on": "n150",   "image": "tracy",  "script": "emitc.sh" },
+  { "runs-on": "n150",   "image": "tracy",  "script": "emitpy.sh" },
```
nsumrakTT: It is compiled for both images, does it really need "tracy" image?

jgrimTT: Hmm perhaps not, I was following `emitc.sh` as a guideline, I'll take another look, thanks!

Issue #5042, opened 2025-09-22
[tt-alchemist][EmitPy] Buffer is not allocated


vcanicTT:
I tried to run generated Resnet TTNN code in Python, and I get the error:
```
2025-09-22 13:02:08.541 | critical |          Always | Buffer is not allocated (assert.hpp:111)
Traceback (most recent call last):
  File "/localdev/vcanic/projects/tt-mlir/resnet50_hf_python_bkp/main.py", line 1330, in <module>
    main()
  File "/localdev/vcanic/projects/tt-mlir/resnet50_hf_python_bkp/main.py", line 1321, in main
    measure(v1, loop_count=4)
  File "/localdev/vcanic/projects/tt-mlir/resnet50_hf_python_bkp/main.py", line 1285, in measure
    outputs = forward(inputs)
              ^^^^^^^^^^^^^^^
  File "/localdev/vcanic/projects/tt-mlir/resnet50_hf_python_bkp/main.py", line 234, in forward
    v165 = forward_const_eval_0(v164)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/localdev/vcanic/projects/tt-mlir/resnet50_hf_python_bkp/main.py", line 6, in forward_const_eval_0
    v3 = ttnn.permute(v2, [0, 2, 3, 1], memory_config=ttnn.MemoryConfig(ttnn.TensorMemoryLayout.INTERLEAVED, ttnn.BufferType.DRAM, None), pad_value=0)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/localdev/vcanic/projects/tt-mlir/third_party/tt-metal/src/tt-metal/ttnn/ttnn/decorators.py", line 334, in __call__
    return self.function(*function_args, **function_kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: TT_THROW @ /localdev/vcanic/projects/tt-mlir/third_party/tt-metal/src/tt-metal/ttnn/core/tensor/storage.cpp:35: tt::exception
info:
Buffer is not allocated
backtrace:
[...]

Issue #3787, opened 2025-06-13
Proposal: Improve tensor deallocation logic in tt-mlir


svuckovicTT:
This issue is a proposal on improvements to the `TTNNDeallocate` pass.

Tensor deallocation (`ttnn::deallocate()`) is used to free up memory when tensors are no longer needed. `TTNNDeallocate` pass is responsible for inserting these tensor deallocation calls.

## Current state

Example IR before `TTNNDeallocate` pass:
```mlir
module @MNISTLinear attributes {tt.system_desc = #tt.system_desc<[{role = host, target_triple = "x86_64-pc-linux-gnu"}], [{arch = <wormhole_b0>, grid = 8x8, coord_translation_offsets = 18x18, l1_size = 1499136, num_dram_channels = 12, dram_channel_size = 1073741824, noc_l1_address_align_bytes = 16, pcie_address_align_bytes = 32, noc_dram_address_align_bytes = 32, l1_unreserved_base = 1024, erisc_l1_unreserved_base = 1024, dram_unreserved_base = 1024, dram_unreserved_end = 1073741824, physical_helper_cores = {dram = [ 8x0,  9x0,  10x0,  8x1,  9x1,  10x1,  8x2,  9x2,  10x2,  8x3,  9x3,  10x3]}, supported_data_types = [<f32>, <f16>, <bf16>, <bfp_f8>, <bfp_bf8>, <bfp_f4>, <bfp_bf4>, <bfp_f2>, <bfp_bf2>, <u32>, <u16>, <u8>, <si32>], supported_tile_sizes = [ 4x16,  16x16,  32x16,  4x32,  16x32,  32x32], dst_register_size_tiles = 8, num_cbs = 32, num_compute_threads = 1, num_datamovement_threads = 2}], [0], [3 : i32], [ 0x0x0x0]>} {
[...]

PR #5289, opened 2025-10-09
Add TTNNAdjustDeallocs pass that deletes parameter tensor deallocations

svuckovicTT:
### Ticket
Closes https://github.com/tenstorrent/tt-mlir/issues/3787
Closes https://github.com/tenstorrent/tt-mlir/issues/5042

### Problem description
TTNNDeallocate pass adds deallocate ops on parameter/constant tensors. In codegen path, this will result in an error if a forward function is called twice.

### What's changed
Removed the deallocate calls on constant/parameter tensors in a new pass TTNNAdjustDeallocs.

The initial idea was to add options to the TTNNDeallocate pass so that more granular selection of which tensors to deallocate. However, that would be tricky as it would make TTIRToTTNNBackendPipeline differ for the canonical runtime path and codegen paths.

Many downstream users of tt-mlir (tt-xla, FFE, tt-explorer) rely on the ability to run the TTIRToTTNNBackendPipeline and then TTNNBackendToEmitPyPipeline to avoid having to rerun the first part of the pipeline when going to codegen (this is relevant in cases where they need to get both the flatbuffer and code for a given model).

TODO:
- [x] Confirm the logic is sound for consteval'd functions.
- [x] Add this to emitc/emitpy pipelines

Marking the PR as draft until these are fixed.

### Checklist

### DIFF
--- a/test/ttmlir/Dialect/TTNN/Transforms/ttnn_adjust_deallocs.mlir
+++ b/test/ttmlir/Dialect/TTNN/Transforms/ttnn_adjust_deallocs.mlir
@@ -1,7 +1,7 @@
-// RUN: ttmlir-opt --ttir-to-ttnn-backend-pipeline="enable-const-eval=false system-desc-path=%system_desc_path%" -o %t %s
-// RUN: ttmlir-opt --ttir-to-ttnn-backend-pipeline="enable-const-eval=false system-desc-path=%system_desc_path%" --ttnn-adjust-deallocs -o %t2 %s
-// RUN: [ "$(cat %t | grep -c ttnn.deallocate)" -eq 4 ]
-// RUN: [ "$(cat %t2 | grep -c ttnn.deallocate)" -eq 2 ]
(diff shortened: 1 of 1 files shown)


lib/Dialect/TTNN/Transforms/CMakeLists.txt
```diff
@@ -7,14 +7,15 @@ add_mlir_dialect_library(MLIRTTNNTransforms
...
+        TTNNEmitPyWorkarounds.cpp
+        TTNNFusing.cpp
+        TTNNLayout.cpp
         TTNNToCpp.cpp
         TTNNToPython.cpp
-        TTNNUniqueLocs.cpp
-        TTNNFusing.cpp
         TTNNTraceHoistTransform.cpp
-        TTNNEmitPyWorkarounds.cpp
+        TTNNUniqueLocs.cpp
```
svuckovicTT, line 18: Sorted alphabetically + added `TTNNAdjustDeallocs.cpp`

Issue #4257, opened 2025-07-30
Add ttnn.experimental.nlp_create_qkv_heads op to TTNN IR


sdjordjevicTT:
## Summary
Add support for `ttnn.experimental.nlp_create_qkv_heads` operation to TTNN IR as part of LLM-specific ops onboarding for Llama3.2 architecture.

## Description
This op creates query, key, and value heads for attention mechanisms in transformer models.

## Implementation Requirements
- [ ] Define the op in TTNN dialect
- [ ] Implement op lowering/conversion
- [ ] Add fusing patterns (investigation needed)
- [ ] Add unit tests
- [ ] Add integration tests
- [ ] Update documentation

## Related Issues
- Parent issue: #4214 (LLM-specific ops onboarding)
- Related: ttnn.experimental.nlp_create_qkv_heads_decode

## Architecture
- Target: Llama3.2 architecture
- Category: Experimental NLP operations

## Notes
- Fusing patterns will be needed for performance optimization
- Essential for multi-head attention implementation
- Part of broader LLM ops initiative

ddilbazTT:
I don't think we need this op supported because `split_query_key_value_and_split_heads` op calls this under the hood anyways. @odjuricicTT I believe we can close this issue.

PR #5276, opened 2025-10-08
Register split query key value and split heads

ddilbazTT:
### Ticket
#4967 

### Problem description
This PR registers split query key value and split heads op. TTIR Fusing for bge-m3 encode is under [ddilbaz/split_heads](https://github.com/tenstorrent/tt-mlir/compare/main...ddilbaz/split_heads), which is currently blocked by https://github.com/tenstorrent/tt-mlir/pull/5149 

### What's changed
Added TTIR & TTNN level definitions of split query key value and split heads op. Added op model interface. Added tests. Thanks @mtopalovicTT for improvements.

### Checklist
- [X] New/Existing tests provide coverage for changes

PR #5295, opened 2025-10-09
[TTNN] Add ttnn.experimental.nlp_create_qkv_heads op

jserbedzijaTT:
### Ticket
[Link to Github Issue](https://github.com/tenstorrent/tt-mlir/issues/4257)

### What's changed
- Added support for `nlp_create_qkv_heads` in the TTNN dialect, along with tests.

### Checklist

### DIFF
--- a/lib/Conversion/TTNNToEmitC/TTNNToEmitC.cpp
+++ b/lib/Conversion/TTNNToEmitC/TTNNToEmitC.cpp
@@ -2823,12 +2823,10 @@ private:
 
 public:
   using TTNNToEmitCBaseOpConversionPattern<
-      mlir::tt::ttnn::NLPCreateQKVHeadsOp>::
-      TTNNToEmitCBaseOpConversionPattern;
+      mlir::tt::ttnn::NLPCreateQKVHeadsOp>::TTNNToEmitCBaseOpConversionPattern;
 
   LogicalResult
-  matchAndRewrite(mlir::tt::ttnn::NLPCreateQKVHeadsOp srcOp,
-                  OpAdaptor adaptor,
+  matchAndRewrite(mlir::tt::ttnn::NLPCreateQKVHeadsOp srcOp, OpAdaptor adaptor,
                   ConversionPatternRewriter &rewriter) const override {
 
     ttnn_to_emitc::EmitCTTNNEmitter<mlir::tt::ttnn::NLPCreateQKVHeadsOp>
--- a/include/ttmlir/OpModel/TTNN/TTNNOpModel.h
+++ b/include/ttmlir/OpModel/TTNN/TTNNOpModel.h
@@ -733,20 +733,21 @@ struct OpModel<RotaryEmbeddingLlamaOp> {
 
 template <>
 struct OpModel<NLPCreateQKVHeadsOp> {
-  static llvm::Expected<OpConstraints> getOpConstraints(
-      ttcore::GridAttr deviceGrid, llvm::ArrayRef<int64_t> inputQShape,
-      TTNNLayoutAttr inputQLayout,
-      std::optional<llvm::ArrayRef<int64_t>> inputKVShape,
-      std::optional<TTNNLayoutAttr> inputKVLayout, uint32_t numQHeads,
-      std::optional<uint32_t> numKVHeads, bool transposeKHeads,
-      TTNNLayoutAttr outputLayout);
+  static llvm::Expected<OpConstraints>
+  getOpConstraints(ttcore::GridAttr deviceGrid,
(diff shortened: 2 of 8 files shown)


lib/Conversion/TTNNToEmitC/TTNNToEmitC.cpp
```diff
@@ -2810,6 +2810,48 @@ class NLPConcatHeadsOpConversionPattern
...
+        emitter.emit(srcOp.getInputQ()),
+        emitter.emit(srcOp.getInputKv()),
+        emitter.emit(srcOp.getNumQHeads()),
+        emitter.emit(srcOp.getNumKvHeads()),
+        emitter.emit(srcOp.getTransposeKHeads()),
+        emitter.emit(srcOp.getMemoryConfig()) |
+            emitter.getMemoryConfig(srcOp.getQuery()),
+    };
+
+    emitter.replaceOp(*this, args);
```
mtopalovicTT, line 2845: This will fail down the road when trying to translate to cpp. Check nlpcreateqkvheadsdecodeop conversion pattern.

include/ttmlir/Dialect/TTNN/IR/TTNNOps.td
```diff
@@ -1082,6 +1082,70 @@ def TTNN_ConcatenateHeadsOp: TTNN_Op<"concatenate_heads", [TTNN_MemoryConfigOpIn
     let hasVerifier = 1;
 }
 
+def TTNN_NLPCreateQKVHeadsOp: TTNN_Op<"nlp_create_qkv_heads", [TTNN_MemoryConfigOpInterface]> {
+    let summary = "Create query, key, value heads for attention layer.";
+    let description = [{
+      Shuffles a fused QKV matrix into separate Q, K, and V head tensors for multi-head attention.
```
mtopalovicTT, line 1088: Interesting. This op looks the same as https://github.com/tenstorrent/tt-mlir/pull/5276 but with exception that this op input is 4D tensor with unsqueezed dim 1. I'm wondering what are the differences between the two and is there real need to have both of them, we should ping someone from metal for this. What are your thoughts @ddilbazTT @jserbedzijaTT ?

include/ttmlir/OpModel/TTNN/MetalHeaders.h
```diff
@@ -61,6 +61,7 @@ extract_output_tensor(const std::tuple<Tensor, Tensor, Tensor> &result) {
 #include "ttnn/operations/embedding_backward/embedding_backward.hpp"
 #include "ttnn/operations/experimental/transformer/nlp_concat_heads/nlp_concat_heads.hpp"
 #include "ttnn/operations/experimental/transformer/nlp_concat_heads_decode/nlp_concat_heads_decode.hpp"
+#include "ttnn/operations/experimental/transformer/nlp_create_qkv_heads/nlp_create_qkv_heads.hpp"
```
mtopalovicTT, line 64: I think you also need to add this header to all `ttnn-precompiled.hpp` files for cpp.

runtime/lib/ttnn/operations/transformer/nlp_create_qkv_heads.cpp
```diff
@@ -0,0 +1,41 @@
...
+      tensorPool.getTTNNTensorAndValidate(op->input_q());
+
+  std::optional<::ttnn::Tensor> inputKV = std::nullopt;
+  if (op->input_kv()) {
+    inputKV = tensorPool.getTTNNTensorAndValidate(op->input_kv());
+  }
+
+  uint32_t numQHeads = op->num_q_heads();
+  std::optional<uint32_t> numKVHeads = std::nullopt;
+  if (op->num_kv_heads() != 0) {
```
mtopalovicTT, line 27: This is a bug, use if (op->num_kv_heads) { ... }

include/ttmlir/Dialect/TTNN/IR/TTNNOps.td
ddilbazTT: Yeah, we should check with metal. I am under the impression that experimental ops are usually targeting specific models rather than being generalistic, but correct me if I'm wrong. This op though seems to be a reimplementation of split heads. I don't know if there's a tradeoff between the two and tt-metal folks could probably help us out with that. For the bge-m3 model I worked with, I think I would have to insert an unsqueeze before the create qkv heads op, whereas I don't need that with the split heads op.

jserbedzijaTT: Yep, looks to me like the only difference is the shape of the input tensors. I’ll write to someone from metal to check if there are any other differences.

PR #5301, opened 2025-10-09
Model ttnn.view

LPanosTT:
### Ticket

Modelling `ttnn.view`: https://github.com/tenstorrent/tt-mlir/issues/5312
- This PR will not close the above issue. In order to remove the runtime workaround introduced in this PR we must also model `ttnn.clone` so that we may convert all cases of `ttnn.reshape` to view, even when we must clone the input.

The following ticket is indirectly related as these issues are blocking vLLM development progress: https://github.com/tenstorrent/tt-xla/issues/1365

### Problem description

- The following problem stems from the fact that we do not model the behavior of `ttnn::reshape` in its entirety.
- `ttnn::reshape` can occasionally be an in-place operation. This occurs when the reshape is equivalent to a _view_. That is to say, no data movement is required to represent the tensor we wish to make with the reshape. The "reshape" can be performed solely by modifying the tensors shape attribute(s), and if that is true then that is what ttnn does. Thus, the output ttnn tensor object is the **exact same object** as the input (in-place).
- This is an issue when we have multiple programs, i.e:
    - Program A takes an input, reshapes it, and returns the reshape as an output. Since the same input is required in Program B as well, we will pass the same runtime tensor used as an input to Program A to Program B as well. 
    - **In the event the reshape inside Program A is performed in-place, the runtime tensor which we now wish to pass as an input to Program B contains a ttnn tensor with a shape which is different than what Program B expects.**
    - In addition to that, tt-xla deallocates output tensors once they have been pushed to host. This is an issue because when we wish to execute Program B, its input has actually been deallocated since the output runtime tensor for Program A has been deallocated, and it held the exact same ttnn tensor as the input we wished to reuse. 

### What's changed

- Add a runtime workaround to **force** reshapes to behave in an out-of-place manner at all times as that is how we model the op.
- This is done by replicating the logic which ttnn uses to determine if a reshape should be done in-place.
- If we determine that `ttnn::reshape` will be performed in-place, we **clone** the input tensor, and allow `ttnn::reshape` to modify that tensor in-place. This way, the output tensor we place inside the `tensorPool` is always a different ttnn tensor object for any reshape.
- Modelled the `ttnn.view` op. When a `ttnn.reshape` is a valid `ttnn.view` AND the reshape is the last user (sequentially) of its operand, then we canonicalize `ttnn.reshape` to `ttnn.view`.
    - Note that in some cases where a reshape will truly perform a view, we do not canonicalize:
        - If the operand is a graph input we do not canonicalize to view because we will end up modifying an input tensor, which can be problematic if that same tensor is used in another program down the line.
        - If the operand of the view is consumed by a descendent of the view. If this happens then we cannot canonicalize to view because the input will be changed as well, breaking the program.
            - I also added a canonicalization pattern to reshape which attempts to move the reshape down in the graph so that it is the last user of its operand. Not that this doesn't change the structure of the graph, only the sequence in which ops are executed.
### Checklist

### DIFF
--- a/lib/Dialect/TTNN/IR/TTNNOps.cpp
+++ b/lib/Dialect/TTNN/IR/TTNNOps.cpp
@@ -1308,25 +1308,9 @@ static mlir::OpFoldResult foldIdentityReshape(mlir::tt::ttnn::ReshapeOp op) {
 static mlir::OpFoldResult foldConsecutiveReshape(mlir::tt::ttnn::ReshapeOp op) {
   auto reshapeOperand =
       op.getInput().getDefiningOp<mlir::tt::ttnn::ReshapeOp>();
-
   if (!reshapeOperand) {
     return nullptr;
   }
-
-  // We do not want to fold the consecutive reshapes if the first reshape's
-  // result is consumed by an op which has memory effects on it BEFORE the
-  // second reshape.
-  auto reshapeOperandUsers = reshapeOperand.getResult().getUsers();
-  for (Operation *user : reshapeOperandUsers) {
-    if (utils::operationHasNonReadMemoryEffectsOnValue(
-            reshapeOperand.getResult(), user) &&
-        user->getBlock() == op->getBlock() && user->isBeforeInBlock(op)) {
-      return nullptr;
-    }
-  }
-
-  op.getOperation()->setOperand(0, reshapeOperand.getInput());
-  return op.getResult();
 
   // Check if any user (except this reshape) writes to the intermediate value
   mlir::Value intermediate = reshapeOperand.getResult();
--- a/include/ttmlir/Dialect/TTNN/Utils/Utils.h
+++ b/include/ttmlir/Dialect/TTNN/Utils/Utils.h
@@ -113,12 +113,6 @@ bool producesTiledTensorLayout(Operation *op);
 bool hasFirstOperandInDRAM(Operation *op);
 
 mlir::RankedTensorType getTraceIdType(MLIRContext *ctx);
-
-// Will return true if the operation has non-read memory effects on the provided
-// Value.
-bool operationHasNonReadMemoryEffectsOnValue(mlir::Value value,
-                                             mlir::Operation *op);
-
 } // namespace mlir::tt::ttnn::utils
 
 #endif // TTMLIR_DIALECT_TTNN_UTILS_UTILS_H
--- a/lib/Dialect/TTNN/Utils/Utils.cpp
+++ b/lib/Dialect/TTNN/Utils/Utils.cpp
@@ -12,7 +12,6 @@
 #include "mlir/IR/Location.h"
 #include "mlir/IR/Operation.h"
 #include "mlir/IR/Value.h"
-#include "mlir/Interfaces/SideEffectInterfaces.h"
 #include "mlir/Support/LLVM.h"
 #include "llvm/Support/Casting.h"
 
@@ -343,23 +342,6 @@ mlir::RankedTensorType getTraceIdType(MLIRContext *ctx) {
       /*shape=*/{},
       ::mlir::IntegerType::get(ctx, /*width=*/32, IntegerType::Unsigned),
       ttnn::TraceIdAttr::get(ctx));
-}
-
-bool operationHasNonReadMemoryEffectsOnValue(mlir::Value value,
-                                             mlir::Operation *op) {
-  mlir::MemoryEffectOpInterface memoryEffectOp =
-      mlir::dyn_cast_or_null<mlir::MemoryEffectOpInterface>(op);
-  if (memoryEffectOp) {
-    SmallVector<SideEffects::EffectInstance<MemoryEffects::Effect>> effects;
-    memoryEffectOp.getEffectsOnValue(value, effects);
-    return llvm::any_of(
-        effects,
-        [](const SideEffects::EffectInstance<MemoryEffects::Effect> &effect) {
-          return !isa<MemoryEffects::Read>(effect.getEffect());
-        });
-  }
-
-  return false;
 }
 
 } // namespace mlir::tt::ttnn::utils


runtime/lib/ttnn/operations/data_movement/reshape.cpp
```diff
@@ -22,7 +24,49 @@ void run(const ::tt::target::ttnn::ReshapeOp *op, ProgramContext &context) {
                 ::tt::runtime::ttnn::utils::getTensorRefMemoryConfig(op->out()))
           : ::tt::runtime::ttnn::utils::createMemoryConfigIfNeeded(
                 op->memory_config());
-  ::ttnn::Tensor out = ::ttnn::reshape(in, shape, memoryConfig);
+  ::ttnn::Tensor out;
+
+  if (::tt::runtime::workaround::Env::get().forceOutOfPlaceReshape) {
```
jnie-TT: Could we move all the `this_is_view` logic into a separate function, so we can update the code to be something like
```
if (::tt::runtime::workaround::Env::get().forceOutOfPlaceReshape && isInplaceReshape(in)) {
       ::ttnn::Tensor clonedInput =
          ::ttnn::clone(in, std::nullopt, std::nullopt, std::nullopt);
      out = ::ttnn::reshape(clonedInput, shape, memoryConfig);
}
else {
      out = ::ttnn::reshape(in, shape, memoryConfig);
}
```
where `isInplaceReshape` will contain all the `this_is_view` checks.

Also minor nit to update to `camelCase`.

jnie-TT (approved):

runtime/lib/ttnn/operations/data_movement/reshape.cpp
LPanosTT: Will do

lib/Dialect/TTNN/Interfaces/TTNNOpModelInterface.cpp
```diff
@@ -1414,6 +1414,41 @@ ReshapeOp::getOpRuntime(const std::vector<TTNNLayoutAttr> &inputs,
...
+ViewOp::getOpConstraints(const std::vector<TTNNLayoutAttr> &inputs,
+                         const OpConfig &opConfig) {
+  ttcore::GridAttr deviceGrid =
+      ttcore::lookupDevice(getOperation()).getWorkerGrid();
+  const auto inputShape = getInput().getType().getShape();
+  SmallVector<int64_t> outputShape;
+
+  for (const Attribute dimAttr : getShape().getValue()) {
+    outputShape.push_back(mlir::cast<mlir::IntegerAttr>(dimAttr).getInt());
+  }
```
sgholamiTT: Can we use `detail::convertArrayAttrToSmallVec` here ?

lib/Dialect/TTNN/Interfaces/TTNNOpModelInterface.cpp
```diff
@@ -1414,6 +1414,41 @@ ReshapeOp::getOpRuntime(const std::vector<TTNNLayoutAttr> &inputs,
...
+
+llvm::Expected<size_t>
+ViewOp::getOpRuntime(const std::vector<TTNNLayoutAttr> &inputs,
+                     const OpConfig &opConfig) {
+  const auto inputShape = getInput().getType().getShape();
+  SmallVector<int64_t> outputShape;
+
+  for (const Attribute dimAttr : getShape().getValue()) {
+    outputShape.push_back(mlir::cast<mlir::IntegerAttr>(dimAttr).getInt());
+  }
```
sgholamiTT: Similar here.

runtime/lib/ttnn/operations/data_movement/reshape.cpp
```diff
@@ -8,7 +8,37 @@
...
+  return (tensorShapeLastDim == shapeLastDim) &&
+         (memConfig.is_sharded() == in.memory_config().is_sharded()) &&
+         (memConfig.is_l1() == in.memory_config().is_l1()) &&
+         ((in.layout() == ::ttnn::ROW_MAJOR_LAYOUT) || // It's row major
+          (tensorShapeSecondLastDim ==
+           shapeSecondLastDim) || // Second last dimension is the same
+          (shapeSecondLastDim % tileSecondDim == 0 &&
+           tensorShapeSecondLastDim % tileFirstDim ==
+               0)); // There is no padding on the second last dimension
+}
```
jserbedzijaTT: Instead of having this utility function here, can you put it in ttnn and just include it from there?

lib/Dialect/TTNN/IR/TTNNOps.cpp
```diff
@@ -1322,9 +1335,199 @@ ::mlir::OpFoldResult mlir::tt::ttnn::ReshapeOp::fold(FoldAdaptor adaptor) {
...
+    if (user == getOperation()) {
+      continue;
+    }
+    if (user->getBlock() == getOperation()->getBlock()) {
+      if (!user->isBeforeInBlock(getOperation())) {
+        return emitOpError(
+            "View op must be the last operation which uses its operand.");
+      }
+    }
+  }
```
jserbedzijaTT, line 1501: What if there is a user in a different block?

lib/Dialect/TTNN/IR/TTNNOps.cpp
```diff
@@ -1322,9 +1335,199 @@ ::mlir::OpFoldResult mlir::tt::ttnn::ReshapeOp::fold(FoldAdaptor adaptor) {
...
+      operandIsFuncArgument = true;
+      break;
+    }
+  }
+
+  if (operandIsFuncArgument) {
+    return emitOpError(
+        "View op must not be used as an operand of a function argument as it "
+        "will modify the input tensors shape in-place.");
+  }
```
jserbedzijaTT:
```suggestion
for (BlockArgument arg : getOperation()->getBlock()->getArguments()) {
    if (arg == getInput()) {
      return emitOpError(
        "View op must not be used as an operand of a function argument as it "
        "will modify the input tensors shape in-place.");
    }
  }
```

lib/Dialect/TTNN/IR/TTNNOps.cpp
```diff
@@ -1322,9 +1335,199 @@ ::mlir::OpFoldResult mlir::tt::ttnn::ReshapeOp::fold(FoldAdaptor adaptor) {
...
+    if (dimValue != outputShape[i]) {
+      return emitOpError() << "Shape attribute " << dimValue
+                           << " must match the output tensor shape "
+                           << outputShape[i] << " at index " << i << ".";
+    }
+  }
+
+  if (!isValidTTNNView(getInput(), getResult())) {
+    return emitOpError("Desired input -> output view is not valid.");
+  }
```
jserbedzijaTT, line 1537: We should give more information to the user about the error because `isValidTTNNView` performs some non-trivial checks.

lib/Dialect/TTNN/IR/TTNNOps.cpp
```diff
@@ -1322,9 +1335,199 @@ ::mlir::OpFoldResult mlir::tt::ttnn::ReshapeOp::fold(FoldAdaptor adaptor) {
...
+        for (BlockArgument arg : block->getArguments()) {
+          if (arg == op.getInput()) {
+            operandIsFuncArgument = true;
+            break;
+          }
+        }
+        if (!isLastUser || !isValidTTNNView(op.getInput(), op.getResult()) ||
+            operandIsFuncArgument) {
+          return failure();
+        }
```
jserbedzijaTT:
```suggestion
        if (llvm::any_of(op.getInput().getUsers(), [&](Operation *user) {
          return user != op && !user->isBeforeInBlock(op);
        })) {
          return failure();
        }

        if (mlir::isa<BlockArgument>(op.getInput())) {
          return failure();
        }

        if (!isValidTTNNView(op.getInput(), op.getResult())) {
          return failure();
        }
```

jserbedzijaTT: The following if statement checks whether the input is an argument of any Block. I’m not sure if that’s the desired behavior, so please double-check
```
if (mlir::isa<BlockArgument>(op.getInput())) {
  return failure();
}
```

lib/Dialect/TTNN/IR/TTNNOps.cpp
```diff
@@ -1322,9 +1335,199 @@ ::mlir::OpFoldResult mlir::tt::ttnn::ReshapeOp::fold(FoldAdaptor adaptor) {
...
+            op.getShape());
+        rewriter.replaceOp(op, viewOp);
+        return success();
+      },
+      0);
+
+  // If a reshape is a valid view, we want to move it after the last user of its
+  // operand so that we can convert this to a valid view.
+  patterns.add(
+      +[](mlir::tt::ttnn::ReshapeOp op, mlir::PatternRewriter &rewriter) {
```
jserbedzijaTT, line 1426: If we had an example like this:
```
%input = some_producer(...)
%reshaped = ttnn.reshape(%input)
%a = op1(%reshaped)
op2(%input)
```

We would not perform any movement of the reshape. To me, it looks like we could’ve rearranged the operations like this:
```
%input = some_producer(...)
op2(%input)
%reshaped = ttnn.reshape(%input)
%a = op1(%reshaped)
```

Am I missing something?

jgrimTT (approved):
ttrt looks good

lib/Dialect/TTNN/IR/TTNNOps.cpp
LPanosTT: Then we may end up with an incorrect graph. What I am trying to achieve here is ensuring that the view is the last op which is executed in the sequential order of the graph. In the case where there is another user of the view's operand in a different block, I am unsure what to do.

lib/Dialect/TTNN/IR/TTNNOps.cpp
LPanosTT: If the `ttnn.reshape` op will actually perform a `ttnn.view` under the hood, then in your first example, `%input` will have it s shape updated in-place. However, we are modelling `ttnn.reshape` to be out-of-place and so in the IR, we are passing `%input` to `op2`, but the shape of `%input` has actually changed. We wish to move the `reshape` to be the last user of `%input` so that there are no down-stream effects if it having changed the shape of `%input` in place.

If the `ttnn.reshape` does not perform a `ttnn.view` under the hood (i.e. data must move) then `ttnn.reshape` will be out of place and we can move it or not move it, it doesn't matter.

LPanosTT: The goal of this movement is so that we can convert all reshapes which are actually views to `ttnn.view` and have the most correct graph possible. We'd like to keep our assumtion that `ttnn.reshape` is out-of-place and not "sometimes in-place sometimes out-of-place". 

The workaround added in runtime ensures that reshape is out of place, and this canonicalization to view exists so that we can remove perf regressions due to forcing in-place reshapes which (luckily) haven't caused issues yet to out-of-place ones.

mtopalovicTT:
> **Problem 2**
> - The following problem stems from the fact that we do not model the behavior of `ttnn::reshape` in its entirety.
> - `ttnn::reshape` can occasionally be an in-place operation. This occurs when the reshape is equivalent to a _view_. That > is to say, no data movement is required to represent the tensor we wish to make with the reshape. The "reshape" can be performed solely by modifying the tensors shape attribute(s), and if that is true then that is what ttnn does. Thus, the output ttnn tensor object is the **exact same object** as the input (in-place).
> - This is an issue when we have multiple programs, i.e:
    - Program A takes an input, reshapes it, and returns the reshape as an output. Since the same input is required in Program B as well, we will pass the same runtime tensor used as an input to Program A to Program B as well. 
    - **In the event the reshape inside Program A is performed in-place, the runtime tensor which we now wish to pass as an input to Program B contains a ttnn tensor with a shape which is different than what Program B expects.**
    - In addition to that, tt-xla deallocates output tensors once they have been pushed to host. This is an issue because when we wish to execute Program B, its input has actually been deallocated since the output runtime tensor for Program A has been deallocated, and it held the exact same ttnn tensor as the input we wished to reuse. 

**Program A takes an input, reshapes it, and returns the reshape as an output. Since the same input is required in Program B as well, we will pass the same runtime tensor used as an input to Program A to Program B as well. **

How is this possible? If view is returned, tensor from graph A is cloned with storage from non reshaped tensor and new shape. That means now we have two tensor objects in memory that have different shapes but same storage.

LPanosTT:
>How is this possible? If view is returned, tensor from graph A is cloned with storage from non reshaped tensor and new shape. That means now we have two tensor objects in memory that have different shapes but same storage.

@mtopalovicTT The tensor is not cloned. The tensors shape attribute is updated in place. We wrap this output in a new **runtime** tensor, but that **runtime** tensor still refers to a ttnn tensor which is identical to the one held by the runtime tensor we passed to the input.

mtopalovicTT:
> > How is this possible? If view is returned, tensor from graph A is cloned with storage from non reshaped tensor and new shape. That means now we have two tensor objects in memory that have different shapes but same storage.
> 
> @mtopalovicTT The tensor is not cloned. The tensors shape attribute is updated in place. We wrap this output in a new **runtime** tensor, but that **runtime** tensor still refers to a ttnn tensor which is identical to the one held by the runtime tensor we passed to the input.

I'm not sure if we messed up something in runtime, but from what I was able to see from [metal](https://github.com/tenstorrent/tt-metal/blob/main/ttnn/cpp/ttnn/operations/experimental/reshape/view.cpp#L36) they create new TensorSpec with new shape and then just create tensor with new tensor spec and storage from original tensor.
When I said cloned I meant class Tensor inside metal is copy constructed with new shape and storage from input tensor.


Edit:

You are hitting this because output is force deallocated from xla. I didn't get into the actual code that you added here (I will check it today), but we can maybe not force deallocate output of graph from xla?

lib/Dialect/TTNN/IR/TTNNOps.cpp
```diff
@@ -1308,10 +1308,26 @@ static mlir::OpFoldResult foldIdentityReshape(mlir::tt::ttnn::ReshapeOp op) {
...
+  for (Operation *user : reshapeOperandUsers) {
+    if (utils::operationHasNonReadMemoryEffectsOnValue(
+            reshapeOperand.getResult(), user) &&
+        user->getBlock() == op->getBlock() && user->isBeforeInBlock(op)) {
+      return nullptr;
+    }
+  }
+
+  op.getOperation()->setOperand(0, reshapeOperand.getInput());
+  return op.getResult();
```
mtopalovicTT: Code below this return is not reachable.

lib/Dialect/TTNN/IR/TTNNOps.cpp
```diff
@@ -1308,10 +1308,26 @@ static mlir::OpFoldResult foldIdentityReshape(mlir::tt::ttnn::ReshapeOp op) {
...
       op.getInput().getDefiningOp<mlir::tt::ttnn::ReshapeOp>();
+
   if (!reshapeOperand) {
     return nullptr;
   }
 
+  // We do not want to fold the consecutive reshapes if the first reshape's
+  // result is consumed by an op which has memory effects on it BEFORE the
+  // second reshape.
+  auto reshapeOperandUsers = reshapeOperand.getResult().getUsers();
```
mtopalovicTT: Why is this not working. Shouldn't we just extend below logic to check if op with memory effect is below?
```cpp
  if (llvm::any_of(intermediate.getUsers(), hasWriteEffect)) {
    return nullptr;
  }
```

LPanosTT:
@mtopalovicTT 
>Do we really need view? From what I could test from xla just removing deallocate force true fixes the issue you were seeing.

That does resolve the issue, however since the result of a reshape which is performed by a view will have shared memory with the operand. I still think it will be useful to model the op differently in case ops which have memory effects are done on the view result, thus changing the data we expect to be in the operand tensor (if it is used elsewhere).

lib/Dialect/TTNN/IR/TTNNOps.cpp
LPanosTT: I created this PR a little bit ago. It seems @pglusacTT has added similar checks here and It didn't come into conflict when I rebased. I'll see if my changes here are needed anymore and remove them if not.

lib/Dialect/TTNN/IR/TTNNOps.cpp
LPanosTT: This return is here because of a rebase on top of similar changes. Mentioned in last comment.

mtopalovicTT:
> @mtopalovicTT
> 
> > Do we really need view? From what I could test from xla just removing deallocate force true fixes the issue you were seeing.
> 
> That does resolve the issue, however since the result of a reshape which is performed by a view will have shared memory with the operand. I still think it will be useful to model the op differently in case ops which have memory effects are done on the view result, thus changing the data we expect to be in the operand tensor (if it is used elsewhere).

@LPanosTT can you confirm this? I looked at metal code yesterday and I seen that Buffer class has own_data member, which is I would assume false in case of the view. Shouldn't be hard to test.

lib/Dialect/TTNN/IR/TTNNOps.cpp
LPanosTT: It seems like @pglusacTT's changes have the exact same effect. I have removed mine.

LPanosTT:
@mtopalovicTT 

Here is a repro in ttnn:
```
import ttnn
import torch

device = ttnn.open_device(device_id=0)

torch_ones = torch.ones((128, 4, 32, 256), dtype=torch.bfloat16)
ttnn_ones = ttnn.from_torch(torch_ones, dtype=ttnn.bfloat16, layout=ttnn.TILE_LAYOUT, device=device)

# This reshape operation performs a view under the hood.
ttnn_ones_reshaped = ttnn.reshape(ttnn_ones, (1, 128, 128, 256))

# Deallocate output tensor. We would think that reshape is out-of-place and so `ttnn_ones` is unnaffected ...
ttnn.deallocate(ttnn_ones_reshaped)

# However, ttnn.from_device fails with: "Buffer must be allocated on device!"
# And so ttnn_ones is also deallocated
host_ones = ttnn.from_device(ttnn_ones)
```

If you change the target shape of the reshape to `(128, 8, 16, 256)` the test will pass as data movement must occur to perform that reshape, and so it is done out of place.

LPanosTT:
@mtopalovicTT 

In this example, `ttnn_zeros` ends up with some values in it even though we never explicitly place values inside: 
```
import ttnn
import torch

device = ttnn.open_device(device_id=0)

torch_zeros = torch.zeros((1, 32, 32, 32), dtype=torch.bfloat16)
ttnn_zeros = ttnn.from_torch(torch_zeros, dtype=ttnn.bfloat16, layout=ttnn.TILE_LAYOUT, device=device)

fill_val = torch.randn((1, 4, 1, 32), dtype=torch.bfloat16)
ttnn_fill_val = ttnn.from_torch(fill_val, dtype=ttnn.bfloat16, layout=ttnn.TILE_LAYOUT, device=device)

ttnn_zeros_reshaped = ttnn.reshape(ttnn_zeros, (8, 4, 32, 32))
ttnn.update_cache(ttnn_zeros_reshaped, ttnn_fill_val, 1)


host_zeros = ttnn.from_device(ttnn_zeros)
host_zeros = ttnn.to_torch(host_zeros)
print(host_zeros)
```

mtopalovicTT:
> @mtopalovicTT 
> 
> 
> 
> In this example, `ttnn_zeros` ends up with some values in it even though we never explicitly place values inside: 
> 
> ```
> 
> import ttnn
> 
> import torch
> 
> 
> 
> device = ttnn.open_device(device_id=0)
> 
> 
> 
> torch_zeros = torch.zeros((1, 32, 32, 32), dtype=torch.bfloat16)
> 
> ttnn_zeros = ttnn.from_torch(torch_zeros, dtype=ttnn.bfloat16, layout=ttnn.TILE_LAYOUT, device=device)
> 
> 
> 
> fill_val = torch.randn((1, 4, 1, 32), dtype=torch.bfloat16)
> 
> ttnn_fill_val = ttnn.from_torch(fill_val, dtype=ttnn.bfloat16, layout=ttnn.TILE_LAYOUT, device=device)
> 
> 
> 
> ttnn_zeros_reshaped = ttnn.reshape(ttnn_zeros, (8, 4, 32, 32))
> 
> ttnn.update_cache(ttnn_zeros_reshaped, ttnn_fill_val, 1)
> 
> 
> 
> 
> 
> host_zeros = ttnn.from_device(ttnn_zeros)
> 
> host_zeros = ttnn.to_torch(host_zeros)
> 
> print(host_zeros)
> 
> ```

This seems like a metal bug. I will check it later.

LPanosTT:
@mtopalovicTT It does make sense to me to perform a view instead of performing the op out of place if the reshape can be down that way. It effectively turns the operation into a no-op. I think this is intentional.

mtopalovicTT:
> @mtopalovicTT It does make sense to me to perform a view instead of performing the op out of place if the reshape can be down that way. It effectively turns the operation into a no-op. I think this is intentional.

Yea i get that. The part that doesnt make sense is that you after deallocate you cannot do an operation on original tensor.

LPanosTT:
> Yea i get that. The part that doesnt make sense is that you after deallocate you cannot do an operation on original tensor.

Both tensors share the same device buffer. There probably should be some warning at least during deallocate if the ref count > 1 or something.

mtopalovicTT:
> > Yea i get that. The part that doesnt make sense is that you after deallocate you cannot do an operation on original tensor.
> 
> Both tensors share the same device buffer. There probably should be some warning at least during deallocate if the ref count > 1 or something.

This is how it works under the hood. There is a bug with ttnn.deallocate when calling it from python from what I was able to test. If you break under the debugger on deallocate_impl when ttnn.deallocate is called you will see that they are setting force = true, this is some issue with how we pybind ttnn.dealloacte in metal we should file issue for that.

```py
import ttnn
import torch

device = ttnn.open_device(device_id=0)

torch_ones = torch.ones((128, 4, 32, 256), dtype=torch.bfloat16)
ttnn_ones = ttnn.from_torch(torch_ones, dtype=ttnn.bfloat16, layout=ttnn.TILE_LAYOUT, device=device)

# This reshape operation performs a view under the hood.
ttnn_ones_reshaped = ttnn.reshape(ttnn_ones, (1, 128, 128, 256))

# Deallocate output tensor. We would think that reshape is out-of-place and so `ttnn_ones` is unnaffected ...
print("Deallocating")
ttnn.deallocate(ttnn_ones_reshaped, False)

# However, ttnn.from_device fails with: "Buffer must be allocated on device!"
# And so ttnn_ones is also deallocated
host_ones = ttnn.from_device(ttnn_ones)
```

This works and only diff is False parameter.

LPanosTT:
> If you break under the debugger on deallocate_impl when ttnn.deallocate is called you will see that they are setting force = tru

In tt-xla we deallocate graph outputs with `force = true`. While we probably should not do this in tt-xla, I do still think we should be modelling `ttnn.view` for the in-place changes case.

This is sort of a separate point I still want to clarify: To remove the reshape workaround introduced in this PR (which clones the input in case of a view), we'll need to model the `ttnn::clone` op. Then we can convert all `ttnn.reshape` ops which are actually `ttnn.view` to `ttnn.view`, not just the ones which are the last users of their operand. In the case where a `ttnn.reshape` is actually a `ttnn.view` and cannot be made to be the last user of its operand, we can insert a clone on the operand and view that instead.

lib/Dialect/TTNN/IR/TTNNOps.cpp
```diff
@@ -1350,9 +1350,195 @@ ::mlir::OpFoldResult mlir::tt::ttnn::ReshapeOp::fold(FoldAdaptor adaptor) {
...
   }
+
   return nullptr;
 }
 
+static bool isValidTTNNView(const TypedValue<RankedTensorType> &input,
+                            const TypedValue<RankedTensorType> &result) {
+  int64_t inputRank = input.getType().getRank();
+  int64_t outputRank = result.getType().getRank();
+  int64_t inputLastDimSize =
```
mtopalovicTT, line 1361: Why do we need to account for rank 0 tensor, I thought we don't have those ever.

lib/Dialect/TTNN/IR/TTNNOps.cpp
```diff
@@ -1350,9 +1350,195 @@ ::mlir::OpFoldResult mlir::tt::ttnn::ReshapeOp::fold(FoldAdaptor adaptor) {
...
+
+  int64_t inputSecondLastDimsSize =
+      inputRank >= 2 ? input.getType().getShape()[inputRank - 2] : 1;
+  int64_t outputSecondLastDimsSize =
+      outputRank >= 2 ? result.getType().getShape()[outputRank - 2] : 1;
+
+  int64_t tileSecondDim = TILE_HEIGHT;
+  int64_t tileFirstDim = TILE_WIDTH;
+
+  TTNNLayoutAttr inputLayout =
```
mtopalovicTT, line 1374: Encoding should always be present in TTNN

lib/Dialect/TTNN/IR/TTNNOps.cpp
```diff
@@ -1350,9 +1350,195 @@ ::mlir::OpFoldResult mlir::tt::ttnn::ReshapeOp::fold(FoldAdaptor adaptor) {
...
+    return false;
+  }
+
+  TTNNLayoutAttr outputLayout =
+      mlir::dyn_cast_or_null<TTNNLayoutAttr>(result.getType().getEncoding());
+  if (!outputLayout) {
+    return false;
+  }
+
+  bool isView = inputLastDimSize == outputLastDimSize;
```
mtopalovicTT, line 1386: Is this requirement also try for row major tensors?

lib/Dialect/TTNN/IR/TTNNOps.cpp
```diff
@@ -1350,9 +1350,195 @@ ::mlir::OpFoldResult mlir::tt::ttnn::ReshapeOp::fold(FoldAdaptor adaptor) {
...
+  int64_t outputLastDimSize =
+      outputRank >= 1 ? result.getType().getShape().back() : 1;
+
+  int64_t inputSecondLastDimsSize =
+      inputRank >= 2 ? input.getType().getShape()[inputRank - 2] : 1;
+  int64_t outputSecondLastDimsSize =
+      outputRank >= 2 ? result.getType().getShape()[outputRank - 2] : 1;
+
+  int64_t tileSecondDim = TILE_HEIGHT;
+  int64_t tileFirstDim = TILE_WIDTH;
```
mtopalovicTT, line 1372: we should add asssert here to confirm that input/output encoding have expected tile width and tile height, since there is option to specify tile width and height in compiler and we might use it in the future. Here you took assumption that it will always be 32/32 but that might not be true in future.

lib/Dialect/TTNN/IR/TTNNOps.cpp
```diff
@@ -1350,9 +1350,195 @@ ::mlir::OpFoldResult mlir::tt::ttnn::ReshapeOp::fold(FoldAdaptor adaptor) {
...
+  if (!outputLayout) {
+    return false;
+  }
+
+  bool isView = inputLastDimSize == outputLastDimSize;
+  isView = isView && inputLayout.getMemLayout().getValue() ==
+                         outputLayout.getMemLayout().getValue();
+  isView = isView && (!inputLayout.isTiled() ||
+                      (inputSecondLastDimsSize == outputSecondLastDimsSize) ||
+                      (outputSecondLastDimsSize % tileSecondDim == 0 &&
```
mtopalovicTT, line 1391: Just curious, why do you mod ouput with tile second dim and input with tile first dim?

lib/Dialect/TTNN/IR/TTNNOps.cpp
```diff
@@ -1350,9 +1350,195 @@ ::mlir::OpFoldResult mlir::tt::ttnn::ReshapeOp::fold(FoldAdaptor adaptor) {
...
+                       inputSecondLastDimsSize % tileFirstDim == 0));
+
+  return isView;
+}
+
+void mlir::tt::ttnn::ReshapeOp::getCanonicalizationPatterns(
+    mlir::RewritePatternSet &patterns, mlir::MLIRContext *context) {
+  patterns.add(
+      +[](mlir::tt::ttnn::ReshapeOp op, mlir::PatternRewriter &rewriter) {
+        if (llvm::any_of(op.getInput().getUsers(), [&](Operation *user) {
```
mtopalovicTT, line 1401: Can you add comment here about what is it that you are testing here

lib/Dialect/TTNN/IR/TTNNOps.cpp
```diff
@@ -1350,9 +1350,195 @@ ::mlir::OpFoldResult mlir::tt::ttnn::ReshapeOp::fold(FoldAdaptor adaptor) {
...
+
+        auto viewOp = rewriter.create<mlir::tt::ttnn::ViewOp>(
+            op.getLoc(), op.getResult().getType(), op.getInput(),
+            op.getShape());
+        rewriter.replaceOp(op, viewOp);
+        return success();
+      },
+      0);
+
+  // If a reshape is a valid view, we want to move it after the last user of its
```
mtopalovicTT, line 1423: Why after last user? If I have this sequence:

%arg;
%view = ttnn.reshape(%arg, ....)
%in_place_op(arg)

if you move it after the last user you will end up with reshape you will see leftover of in_place_op right? I'm just going by what you commented here.

lib/Dialect/TTNN/IR/TTNNOps.cpp
```diff
@@ -1350,9 +1350,195 @@ ::mlir::OpFoldResult mlir::tt::ttnn::ReshapeOp::fold(FoldAdaptor adaptor) {
...
+                         outputLayout.getMemLayout().getValue();
+  isView = isView && (!inputLayout.isTiled() ||
+                      (inputSecondLastDimsSize == outputSecondLastDimsSize) ||
+                      (outputSecondLastDimsSize % tileSecondDim == 0 &&
+                       inputSecondLastDimsSize % tileFirstDim == 0));
+
+  return isView;
+}
+
+void mlir::tt::ttnn::ReshapeOp::getCanonicalizationPatterns(
```
mtopalovicTT, line 1397: can you just dump it down for me. My understanding is that if reshape will result in a view, we should always move the reshape as first operation in graph so that copy will be created. Now we can do some optimizations like, if there are no users of reshape operand which are in place update operations we can leave operation as a view, if not we need to clone the input operand to reshape. You probably went into this more deeper, but it would be good to understand all the cases. Maybe some little doc explaning what are all edgecases would be nice, shouldn't take long to put into words. Also I tried to figure out how other frameworks are dooing it but it seems that noone has architecture like we do where we are basically at the backseat of whatever ttnn is doing.

lib/Dialect/TTNN/IR/TTNNOps.cpp
LPanosTT: in the IR:
```
%arg;
%view = ttnn.reshape(%arg, ....)
%in_place_op(arg)
```

And we move the `view` after `in_place_op`, if there are no other ops between that use the view then it should make no difference as the same data changes will appear in `%view` but just before the call the `ttnn.reshape` instead of after.

lib/Dialect/TTNN/IR/TTNNOps.cpp
LPanosTT: Yes, the current condition(s) for wether to perform a view instead of a reshape in ttnn are here: https://github.com/tenstorrent/tt-metal/blob/8f3cc33a7fc77fd257b14f16cac81c8269691b8f/ttnn/cpp/ttnn/operations/data_movement/reshape_view/reshape.cpp#L300

lib/Dialect/TTNN/IR/TTNNOps.cpp
LPanosTT: I'm pretty sure you're right, however I mimicked how these values were determined by ttnn when they check wether to use view or not: https://github.com/tenstorrent/tt-metal/blob/8f3cc33a7fc77fd257b14f16cac81c8269691b8f/ttnn/cpp/ttnn/operations/data_movement/reshape_view/reshape.cpp#L283

lib/Dialect/TTNN/IR/TTNNOps.cpp
LPanosTT: Copied from the ttnn conditions. Now that you mention it it does seem odd. Its possible this is a mistake and we just haven't hit any issues since the tile sizes have been 32x32 consistently.

lib/Dialect/TTNN/IR/TTNNOps.cpp
LPanosTT: For sure, I have created a doc illustrating the two case we might see in IR: https://docs.google.com/document/d/1eePWgscmHQfFfQ93utRKWYmov9f3O44GMkhEvqhojdc/edit?usp=sharing

PR #5326, opened 2025-10-13
Fix operand fetch in shard solver

rpavlovicTT:
Minor bug fix, ShardSolver would take result type instead of operand's when checking interleaved to sharded case.

Also minor fix in fallback path wrt. ConvTranspose2d.

### DIFF
--- a/lib/Dialect/TTNN/Transforms/OptimizerPasses/OperationValidationAndFallback.cpp
+++ b/lib/Dialect/TTNN/Transforms/OptimizerPasses/OperationValidationAndFallback.cpp
@@ -2,6 +2,7 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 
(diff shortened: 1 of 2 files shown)


lib/Dialect/TTNN/Transforms/OptimizerPasses/OperationValidationAndFallback.cpp
```diff
@@ -280,6 +282,17 @@ class TTNNOperationValidationAndFallback
                                            conv2dOp.getComputeConfigAttr()};
     }
 
+    // TODO(rpavlovicTT): Once
+    // https://github.com/tenstorrent/tt-mlir/issues/3994 is
```
odjuricicTT, line 286: Maybe mention that this should be changed in the issue itself.

Issue #5306, opened 2025-10-10
Force L1Full in conv slice config when optimizer enabled


odjuricicTT:
After this is done the the generality WA (https://github.com/tenstorrent/tt-mlir/pull/5285) should be removed.

Also this might need to get merged before removing the WA https://github.com/tenstorrent/tt-metal/pull/29993

rpavlovicTT:
Remove all hard coded values for slice config throughout tt-mlir repo.

PR #5340, opened 2025-10-14
Fix prepare conv2d weights in OpModelLib

rpavlovicTT:
There was a discrepancy between weights preparation and conv2d invocation which led to compilation failures.

### DIFF
--- a/lib/OpModel/TTNN/TTNNOpModel.cpp
+++ b/lib/OpModel/TTNN/TTNNOpModel.cpp
@@ -569,6 +569,11 @@ getPrepareConv2dWeightsOpOutputTensorSpec(
   std::optional<::ttnn::operations::conv::conv2d::Conv2dConfig>
       conv2dConfigConverted = conversion::getConv2dConfig(conv2dConfig);
 
+  std::optional<::ttnn::operations::conv::conv2d::Conv2dSliceConfig>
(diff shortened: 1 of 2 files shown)


lib/OpModel/TTNN/TTNNOpModel.cpp
```diff
@@ -569,6 +569,11 @@ getPrepareConv2dWeightsOpOutputTensorSpec(
   std::optional<::ttnn::operations::conv::conv2d::Conv2dConfig>
       conv2dConfigConverted = conversion::getConv2dConfig(conv2dConfig);
 
+  std::optional<::ttnn::operations::conv::conv2d::Conv2dSliceConfig>
```
odjuricicTT, line 572: Why hardcode this? Should we not propagate all the way from the op IR?

rpavlovicTT: We already have few places where we hard code it in this file, plus we are missing conversion for it. So I think this is easier temporarily and then I will fix it properly as part of #5306

odjuricicTT: @rpavlovicTT Please file an issue for this, or explicitly mention that this is hardcoded and needs to be fixed in some issue

PR #5346, opened 2025-10-14
Add arith.cmpf conversion to d2m.tile_xxx comparison ops

philei-tt:
### Ticket
[Link to Github Issue](https://github.com/tenstorrent/tt-mlir/issues/5126)

### Problem description
Add conversion from arith.cmpf to one of: 
- TileEqzOp
- TileNezOp
- TileGtzOp
- TileGezOp
- TileLtzOp
- TileLezOp

### What's changed
The logic is trivial. Basically cmpf -> d2.tile_sub + d2.tile_specific_cmp, e.g.
```mlir
%0 = arith.cmpf oeq, %arg0, %arg1 : !ttcore.tile<32x32, f32>
=>
%0 = "d2m.tile_sub"(%arg0, %arg1) : (!ttcore.tile<32x32, f32>, !ttcore.tile<32x32, f32>) -> !ttcore.tile<32x32, f32>
%1 = "d2m.tile_eqz"(%0) : (!ttcore.tile<32x32, f32>) -> !ttcore.tile<32x32, i1>

```

Unfortunately, arith.cmpf has BoolLike output:
```
// Type constraint for bool-like types: bools, vectors of bools, tensors of
// bools.
def BoolLike : TypeOrValueSemanticsContainer<I1, "bool-like">;
```

So I had to make **TTCore_Tile implement ShapedTypeInterface, ValueSemantics to and support i1 type**. 

To run test: 
```
llvm-lit test/ttmlir/Conversion/ArithToD2MTileOps/arith_to_d2m_tile_ops.mlir
```

### Checklist

### DIFF
--- a/.github/Dockerfile.base
+++ b/.github/Dockerfile.base
@@ -4,6 +4,8 @@ SHELL ["/bin/bash", "-c"]
 # Set environment variables
 ENV DEBIAN_FRONTEND=noninteractive
 
+ARG TT_METAL_DEPENDENCIES_COMMIT=7aee9a53eac0023e3660668448e1fbb105038d63
+
 # Install dependencies
 RUN <<EOT
 # ANCHOR: developer_dependencies
@@ -41,7 +43,7 @@ add-apt-repository ppa:deadsnakes/ppa && \
     apt-get install -y python3.11 python3.11-dev python3.11-venv python3.11-distutils
 
 # Setup / install metal dependencies
-wget https://raw.githubusercontent.com/tenstorrent/tt-metal/7aee9a53eac0023e3660668448e1fbb105038d63/{install_dependencies.sh,tt_metal/sfpi-version.sh}
+wget https://raw.githubusercontent.com/tenstorrent/tt-metal/${TT_METAL_DEPENDENCIES_COMMIT}/{install_dependencies.sh,tt_metal/sfpi-version.sh}
 bash install_dependencies.sh --docker
 # ANCHOR_END: developer_dependencies
 EOT
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -52,7 +52,11 @@ jobs:
           cat << 'EOF' > _test_durations
             ${{ vars.TEST_DURATIONS }}
           EOF
-          python .github/scripts/python/generate_test_matrix.py .github/test_matrix/pr-push-matrix.json ${{ vars.TARGET_DURATION }}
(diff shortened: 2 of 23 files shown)


lib/Conversion/TTKernelToEmitC/TTKernelToEmitC.cpp
```diff
@@ -159,6 +159,9 @@ datatypeToDataformatEnumValue(Builder &builder,
   case ::mlir::tt::ttcore::DataType::Int32:
     expression += "Int32";
     break;
+  case ::mlir::tt::ttcore::DataType::Bool:
+    llvm_unreachable("Bool DataType is not supported in TTKernel DataFormat");
```
wenbinlyuTT, line 163: Just curious, if the output of these ops made them non-executable on the HW, where will this functionality be used? Or it's expected that we cast the output to another dataformat?

nsmithtt: Yeah that's correct, in order for our TileType to be compatible with arith / math dialects it needs to be "BoolLike" aka be a ShapedType w/ i1 element type.  This is really only meant to be used early on in the pipeline, as a separate task we'll have to write a pass that promotes these bools to a HW supported dtype.

lib/Conversion/ArithToD2MTileOps/ArithToD2MTileOps.cpp
```diff
@@ -62,6 +62,82 @@ struct ArithToD2MTileOpsPass final
...
+    auto subOp = rewriter.create<d2m::TileSubOp>(loc, tileType, operands[0],
+                                                 operands[1]);
+
+    auto operandTileType = mlir::cast<ttcore::TileType>(operands[0].getType());
+
+    // The expected result type should be a bool tile (tile<i1>) with the same
+    // shape as input tile
+    auto boolTileType = ttcore::TileType::get(operandTileType.getContext(),
+                                              operandTileType.getShape(),
+                                              ttcore::DataType::Bool);
```
nsmithtt, line 93: Re: thread below, not sure if it's too hacky, but we could promote to the operand tile type here during conversion. @wenbinlyuTT thoughts?

wenbinlyuTT: I think it's fine as-is, we can address this when the issue actually arise and we have a better picture of how these bool results are consumed.

PR #5349, opened 2025-10-14
[D2M] Grid selection rework

vwellsTT:
### Problem description
Grid selection is a bit basic at the moment in D2M--we need to make it more op-aware, able to optimize data movement vs compute costs, etc.

### What's changed
As a first step, this PR reverts to grid selection in a standalone pass.  TTIRToD2M now should have no dependencies on the compute grid (both w.r.t. assigning grids and MetalLayoutAttr) -- it will assign 1-filled grids, and tile-aligned layouts, and all grid-specific computations will happen in new subsequent pass.

Follow-on PRs will address special op cases + global optimizations

### Checklist


lib/Dialect/D2M/Transforms/GridSelection.cpp
```diff
@@ -0,0 +1,486 @@
+// SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
```
vwellsTT, line 1: remark: maybe this should be named GenericGridSelection?  Not sure exactly what our philosophy on which passes start w/ "Generic" is tbh

lib/Dialect/TTCore/IR/TTCoreOpsTypes.cpp
```diff
@@ -927,65 +922,25 @@ MetalLayoutAttr::getIndexAffineMapOrIdentity(unsigned rank) const {
 // reduction of 4x43x7 -> 4x1x7 is now worker-local).
 llvm::SmallVector<int64_t>
 MetalLayoutAttr::computeAlignments(ArrayRef<int64_t> logicalShape,
```
wenbinlyuTT: Is it appropriate to rename this function to `computeDefault/MinimalAlignments`? Or it might still become more powerful in the future? Considering you didn't move the comments above to the new pass or simply delete them.

vwellsTT: Not sure on the rename, don't feel strongly either way.  Maybe this could be computeTileAlignments

> Considering you didn't move the comments above to the new pass or simply delete them.

Hmm, yeah I had rewritten the above comment a bit more thoroughly, I think I must've squashed that when rebasing or something.

My vision is that this file should not have any grid specific logic, so I'd like this func to remain in its current form.  I'll modify the above examples to remove any grid-alignments etc. and perhaps move them to the new impl in GridSelection.cpp

wenbinlyuTT: Thanks for the explanation. I was just happy that most of the shape-related calculation logic are grouped into a single place by this PR, and thought that not-so-important logic elsewhere should be clear about their trivialness.

vwellsTT: Yeah, you're probably right; I think `computeTileAlignments` is clearer to me than `computeDefaultAlignments` so I'll change to that

lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
```diff
@@ -408,25 +328,20 @@ class D2MNamedElementwiseRewriter final
...
     const std::size_t numOperands = (numInputs + numOutputs);
-
     assert(numOperands == op->getNumOperands());
 
-    ttcore::GridAttr grid =
-        getOutputGrid(ctx, mlir::cast<ShapedType>(outputs[0].getType()));
-
-    const std::size_t rank = grid.getShape().size();
+    const std::size_t rank =
+        mlir::cast<RankedTensorType>(outputs[0].getType()).getRank() / 2;
```
nsmithtt: `getDeviceLayout(mlir::cast<ShapedType>(outputs[0].getType())).getRank()` (need to add getRank interface method)

We should make an overload that just accepts a Value and does the getType + cast for you, then we could write `getDeviceLayout(outputs[0]).getRank()`.

lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
```diff
@@ -1193,16 +1122,52 @@ class TTIRToD2MPass final
...
 
-    // Get target grid shape from device or override.
-    llvm::SmallVector<int64_t> gridShape = getTargetGridShape();
-
     TypeConverter typeConverter;
     typeConverter.addConversion([](Type t) { return t; });
 
+    // Add source materialization: TTIR tensor -> D2M tensor with layout.
+    // This automatically inserts ToLayout ops when operands need conversion.
+    typeConverter.addSourceMaterialization(
```
nsmithtt: Oh wait I missed how this is working.  It seems like we still have toLayout and unLayout?

lib/Dialect/D2M/Transforms/GridSelection.cpp
nsmithtt: It was originally to help disambiguate generic passes from the rest of ttir, now that we have d2m dialect it might be overkill to be calling everything generic.  I'm fine with this name.

lib/Dialect/TTCore/IR/TTCoreOpsTypes.cpp
nsmithtt: I'm surprised anything had to change in this file, is there a reason we want to remove the ability to pass in a `deviceGridShape`?  It seems like it's still useful + D2M conversion will just pass in its unit grid.

lib/Dialect/TTCore/IR/TTCoreOpsTypes.cpp
```diff
@@ -995,25 +950,26 @@ MetalLayoutAttr::computeAlignments(ArrayRef<int64_t> logicalShape,
 // Getter with no intervals or alignments, we calculate them both.
 MetalLayoutAttr MetalLayoutAttr::get(::mlir::MLIRContext *context,
                                      ArrayRef<int64_t> logicalShape,
-                                     ArrayRef<int64_t> deviceGridShape,
                                      OOBVal oobVal, MemorySpace memorySpace,
                                      TensorMemoryLayout memoryLayout) {
+
+  constexpr size_t kGridRank = 2;
```
nsmithtt, line 943: Don't we expect grid shapes of any rank? The old code seemed more robust.

lib/Dialect/D2M/Transforms/GridSelection.cpp
```diff
@@ -0,0 +1,509 @@
...
+
+  struct ToLayoutUpdateInfo {
+    d2m::ToLayoutOp op;
+    llvm::SmallVector<int64_t> grid;
+  };
+  llvm::SmallVector<ToLayoutUpdateInfo> toLayoutsToUpdate;
+  llvm::SmallVector<std::pair<d2m::StreamLayoutOp, llvm::SmallVector<int64_t>>>
+      streamLayoutsToUpdate;
+
+  // Phase 1: Analyze each operand and compute its optimal grid.
```
nsmithtt: Perhaps we can break each of these phases into subroutines?

lib/Dialect/D2M/Transforms/GridSelection.cpp
```diff
@@ -0,0 +1,509 @@
...
+                                                   llvm::ArrayRef(tileShape));
+    auto newStorageEmpty = builder.create<d2m::EmptyOp>(
+        storageEmpty.getLoc(), newStorageShape, elementType, newStorageLayout);
+
+    builder.setInsertionPoint(streamLayout);
+    auto newStreamLayout = builder.create<d2m::StreamLayoutOp>(
+        streamLayout.getLoc(), newStorageEmpty.getType(),
+        streamLayout.getInput(), newStorageEmpty);
+
+    streamLayout.getResult().replaceAllUsesWith(newStreamLayout.getResult());
```
nsmithtt, line 429: Maybe it's safer to only replace the use for _this_ generic and let folding erase it if needed, or assert that there is exactly one use.  If for some reason the stream was used in multiple places we might be updating other users in an illegal way.

lib/Dialect/TTCore/IR/TTCoreOpsTypes.cpp
vwellsTT: This particular constructor is for computing with default collapse strategy, which precludes grids of rank other than 2 fundamentally (at least for today)

lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
vwellsTT: Good idea

lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
vwellsTT: Oh hmm yeah, I had decided to go back to `toLayout/unLayout` based on our discussions, and missed removing this I guess--surprised it works with both like this 🤔

lib/Dialect/D2M/Transforms/GridSelection.cpp
vwellsTT: makes sense, cool

lib/Dialect/D2M/Transforms/GridSelection.cpp
vwellsTT: 👍 Probably preferable to this 300 line func, ha

lib/Dialect/TTCore/IR/TTCoreOpsTypes.cpp
vwellsTT: This func in its old form is moved to GridSelection.cpp directly.  My thinking is that we could I guess keep this the same and pass {1,1} in TTIRToD2M, and a manually constructed {8,8} in all the python bindings in our tests, etc. but it doesn't feel right to me--the GridSelection should be the sole owner of the grid shape, s.t. it can take the "overrideGrid" param, handle squaring, etc.  I don't like the idea that we need to pass a meaningless grid in on ALL the constructors except the most explicit one (previous state).

I could see a reasonable argument that we should leave this func here, and have 1 ctor which does use this instead.    I'm not 100% sure but I thought maybe in the future when doing some cleanup s.t. we align to something other than full grids, it might be nicer to compute alignments inline vs creating a MetalLayoutAttr, deciding to change it, and creating another one.  But I suppose I could obviate that by making `computeGridAwareAlignments` a static member func of `MetalLayoutAttr` and leaving it here, if you feel strongly.

vwellsTT: So I guess TL;DR I just think it's cleaner to not have the device grid be omni-present in all the `get()` impls when we don't use it most of the time; could specialize a specific `get()` or 2 if you think that's better.

jgrimTT (approved):
Looks good on the `builder` side

lib/Dialect/D2M/Transforms/GridSelection.cpp
```diff
@@ -0,0 +1,585 @@
...
+               llvm::zip(oldBlock.getArguments(), blockArgs)) {
+            mapping.map(oldArg, newArg);
+          }
+          for (Operation &op : oldBlock) {
+            Operation *clonedOp = b.clone(op, mapping);
+
+            // For nested linalg.generic ops, update result types to match the
+            // new output operand types (which have changed due to grid
+            // updates).
+            if (clonedOp->getName().getStringRef() == "linalg.generic") {
```
nsmithtt: Perhaps we should match the linalg generic with isa, instead of a string compare.

nsmithtt: Or maybe even better is to match all `DestinationStyleOpInterface` ops and handle auto-updating those.

vwellsTT: Definitely I agree I should use `isa` over string checks; hmm, are there are other op types we'd expect to handle here than `linalg.generic`, for the second comment?

lib/Dialect/D2M/Transforms/GridSelection.cpp
```diff
@@ -0,0 +1,585 @@
...
+                                      storageLayout.getNormalizedIntervals());
+
+    auto newStorageLayout =
+        (hasIndexMap)
+            ? ttcore::MetalLayoutAttr::get(
+                  builder.getContext(), storageLayout.getLogicalShape(),
+                  storageDimAlignments, storageLayout.getCollapsedIntervals(),
+                  storageLayout.getOobVal(), storageLayout.getMemorySpace(),
+                  storageLayout.getMemoryLayout(),
+                  storageLayout.getIndexAffineMap())
```
nsmithtt: Does `storageLayout.getIndexAffineMap()` not work unconditionally?  Unclear if we need a conditional here.

lib/Dialect/D2M/Transforms/GridSelection.cpp
```diff
@@ -0,0 +1,585 @@
...
+
+    llvm::SmallVector<Value> newInputs(newOperands.begin(),
+                                       newOperands.begin() + numInputs);
+    llvm::SmallVector<Value> newOutputs(newOperands.begin() + numInputs,
+                                        newOperands.end());
+
+    Region &oldRegion = genericOp.getRegion(0);
+
+    builder.setInsertionPoint(genericOp);
+    auto newGenericOp = builder.create<d2m::GenericOp>(
```
nsmithtt, line 461: Doesn't the generic op itself need a new grid shape?

nsmithtt: Oh nvm, I forgot it could be inferred from the operands.

lib/Dialect/D2M/Transforms/GridSelection.cpp
vwellsTT: I had some bugs here I iterated on, but possible I overcomplicated this; I'll try it your way and see if it breaks anything

lib/Dialect/D2M/Transforms/GridSelection.cpp
nsmithtt: We know of at least two ops, scatter/gather that will have loops here instead of linalg.generic, but I think the kernel can still be written in a way that it's robust to the grid changing.

vwellsTT: Gotcha, makes sense, thanks!

lib/Dialect/D2M/Transforms/GridSelection.cpp
vwellsTT: seems like you're right, we can just always pass the map and if it's default that's fine 👍

nsmithtt (approved):

PR #5350, opened 2025-10-14
[explorer] Preload graphs from IR dump

madcampos:
### Problem description
Integrates with https://github.com/tenstorrent/tt-mlir/tree/brata/ir_dump to load the IR dumps into explorer.

Relates to: https://github.com/tenstorrent/model-explorer/pull/72

### What's changed
Added a new command to `tt-explorer` for preloading graphs.

### Checklist

### DIFF
--- a/tools/explorer/tt_adapter/src/tt_adapter/utils.py
+++ b/tools/explorer/tt_adapter/src/tt_adapter/utils.py
@@ -153,7 +153,7 @@ def get_collection_path(model_path: str):
     resolved_model_path = Path(model_path).resolve()
 
     # If the path is adirectory and has an "extension", simply return it.
-    if os.path.is_dir(resolved_model_path) and MODEL_EXTENSIONS.count(resolved_model_path.suffix) > 0:
+    if os.path.isdir(resolved_model_path) and MODEL_EXTENSIONS.count(resolved_model_path.suffix) > 0:
         return resolved_model_path
 
     resolved_ir_dir = Path(IR_DUMPS_DIR).resolve()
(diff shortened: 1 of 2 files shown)


vcanicTT (approved):

tools/explorer/tt_adapter/src/tt_adapter/utils.py
```diff
@@ -130,3 +134,14 @@ def needs_stablehlo_pass(module_path: str) -> bool:
     module_dialect = ModuleDialect.detect(module_str)
 
     return module_dialect == ModuleDialect.STABLE_HLO
+
+def get_collection_label(model_path: str, ir_dir: str):
```
brataTT: This makes sense, and will return `x` if a file exists under `ir_dump/x/file1.txt`
But from the UI, I see this format instead. (for files e.g. `ir_dump/abs/1_Canonicalizer.mlir`)
Is this supposed to be just `abs`?

Image

madcampos: I believe this is the correct behaviour.

If it was `abs.mlir`, then yes, all models would be grouped under `abs.mlir`, but as it "doesn't have an extension", then all files are consideded "loose files" so they are returned alone instead of in a collection.

madcampos: @brataTT we can adjust the behaviour to be different if that makes more sense. Let's discuss this.

IR_DUMPING.md
nsmithtt, line 1: Can you move this to somewhere in `/docs/...` that makes sense?

brataTT: oh, didn't realize the branch was pushing to a PR
we should point this PR back to merge in main 😅

brataTT: @madcampos now that we're aligned on changes and have test data generated, could you separate this PR from the IR dump change and point it back to main?

madcampos: Yes, I will rebase this PR to `main` once @brataTT's is merged 🙂

PR #5358, opened 2025-10-15
Refactor device lifecycle management with OptimizerPassesWrapper

### LINKED ISSUE #5324: Need a way to run cleanup logic after registered pipeline (even on failure)
### Summary

We currently have an MLIR pipeline registered via PassPipelineRegistration, and we need to perform resource management around it — specifically to open a device before the pipeline starts and close it after it finishes.
The main requirement is that cleanup (closeDevice()) must run even if one of the passes fails within the pipeline.
However, PassPipelineRegistration only provides access to an OpPassManager, and pass execution stops immediately after a failure — meaning a final pass added at the end will not execute in failure cases. 
Also, it is not ideal to open/close device within a specific pass like today. Then when we add a pass before/after that needs device we need to move this logic around.

We also don’t control the top-level PassManager (it lives in the client repository), so we can’t attach a PassInstrumentation for before/after hooks.

### Proposed Solution

Wrap the pipeline in a single meta-pass that:
Opens the device before running the inner pipeline.
Runs the inner pipeline via runPipeline().
Closes the device using an RAII guard (llvm::make_scope_exit) to ensure cleanup runs even if the pipeline fails.

```
struct MyPipelineWrapperPass
    : PassWrapper<MyPipelineWrapperPass, OperationPass<ModuleOp>> {

  void runOnOperation() override {
    Operation *op = getOperation();

    // Open device before running sub-pipeline
    openDeviceSingleton();

    // Guarantee cleanup on all exit paths
    auto guard = llvm::make_scope_exit([]() {
      closeDeviceSingleton();
    });

    // Build and run sub-pipeline
    OpPassManager inner(ModuleOp::getOperationName());
    buildMyPipeline(inner);

    if (failed(runPipeline(inner, op))) {
      signalPassFailure();
      return;
    }

    // Device will be closed automatically via guard
  }
};
```

### Potential follow-up

In the future, MLIR could expose a lightweight “pipeline-level instrumentation” API for OpPassManager, allowing pipelines to register before/after hooks without needing a wrapper pass.
Until then, the RAII-style wrapper pattern provides a robust workaround.

### LINKED ISSUE #5357: SingletonDeviceContext is not closed leading to segfault
### Description

Failing in the middle of compilation pipeline would result in assertion

```
python: /localdev/rpavlovic/tt-xla/third_party/tt-mlir/src/tt-mlir/lib/OpModel/TTNN/SingletonDeviceContext.cpp:17: mlir::tt::ttnn::op_model::SingletonDeviceContext::~SingletonDeviceContext(): Assertion `m_device == nullptr && "Device should be null when SingletonDeviceContext is destructed. Call " "closeInstance() once you are done with the device."' failed.
Aborted (core dumped)
```

This is because we don't close device on time. 
We need better device instance management through pipeline.

rpavlovicTT:
This change centralizes device lifecycle management for optimizer passes by introducing a wrapper pass that handles device open/close operations, eliminating the need for individual passes to manage device state.

Key changes:
- Add OptimizerPassesWrapper pass to manage device lifecycle for nested optimizer passes using llvm::make_scope_exit for RAII-style cleanup
- Add OpModelDeviceGuard for scoped device management
- Remove device open/close calls from individual optimizer passes (TTNNOptimizer, TTNNOperationValidationAndFallback, TTNNPrepareConv2dWeightsAndBias)
- Integrate wrapper into TTNNPipeline to wrap all optimizer passes

With this change it's gonna be easier to add new passes in the optimizer group.

Fixes #5357 
Fixes #5324

### DIFF
--- a/.github/workflows/call-build-release.yml
+++ b/.github/workflows/call-build-release.yml
@@ -41,7 +41,7 @@ jobs:
       matrix:
         build: [
           { enable_op_model: "ON", enable_emitc: "ON", name: "speedy" },
-          { enable_op_model: "ON", enable_perf: "ON", enable_runtime_debug: "ON", enable_explorer: "ON", enable_pykernel: "ON", name: "tracy" }
+          { enable_perf: "ON", enable_runtime_debug: "ON", enable_explorer: "ON", enable_pykernel: "ON", name: "tracy" }
         ]
 
     name: Build ${{ matrix.build.name }} Release


include/ttmlir/OpModel/TTNN/OpModelDeviceGuard.h
```diff
@@ -0,0 +1,76 @@
+// SPDX-FileCopyrightText: © 2024 Tenstorrent Inc.
```
rpavlovicTT: 2025

lib/Dialect/TTNN/Transforms/OptimizerPasses/OptimizerPassesWrapper.cpp
```diff
@@ -0,0 +1,83 @@
+// SPDX-FileCopyrightText: © 2025 Tenstorrent Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#include "ttmlir/Support/Logger.h"
+#include <llvm/Support/raw_ostream.h>
```
rpavlovicTT: quotes

lib/Dialect/TTNN/Transforms/OptimizerPasses/OptimizerPassesWrapper.cpp
```diff
@@ -0,0 +1,83 @@
...
+      op_model::SingletonDeviceContext::getInstance().openDevice();
+    }
+
+    // Create nested pass manager and populate it.
+    OpPassManager nestedPm(getOperation()->getName());
+    populatePipeline(nestedPm);
+
+    // Ensure closeInstance() gets called always.
+    auto guard = llvm::make_scope_exit([]() noexcept {
+      llvm::errs() << "Exiting nested pipeline and closing device instance\n";
```
rpavlovicTT: delete print

lib/Dialect/TTNN/Pipelines/TTNNPipelines.cpp
```diff
@@ -77,34 +78,38 @@ void createTTNNPipelineTTIRPasses(
...
+              mlir::tt::ttnn::createTTNNOperationValidationAndFallback(
+                  validationOptions));
+          innerPm.addPass(
+              mlir::tt::ttnn::createTTNNPrepareConv2dWeightsAndBias());
+        },
+        wrapperOptions));
+#else
+    // Non-OpModel path
+    pm.addPass(mlir::tt::ttnn::createTTNNOptimizer(optimizerOptions));
+    pm.addPass(mlir::createCanonicalizerPass());
```
odjuricicTT: Do we need this? Or can it be removed?

lib/Dialect/TTNN/Transforms/OptimizerPasses/OptimizerPassesWrapper.cpp
```diff
@@ -0,0 +1,83 @@
...
+      op_model::SingletonDeviceContext::closeInstance();
```
odjuricicTT: Nice. Does this cover the case when an assert is hit in one of the nested passes as well?

include/ttmlir/OpModel/TTNN/OpModelDeviceGuard.h
```diff
@@ -0,0 +1,76 @@
...
+// when the guard goes out of scope, even if exceptions or early returns occur.
+// Useful for standalone passes that need to ensure device is available.
+//
+// Usage:
+//   void runOnOperation() override {
+//     OpModelDeviceGuard deviceGuard;
+//     // ... do work with device ...
+//   }  // Device automatically closed if opened by this guard
+//
+class OpModelDeviceGuard {
```
odjuricicTT: Is this abstraction needed or can this logic live in SingletonDeviceContext?

lib/Dialect/TTNN/Transforms/OptimizerPasses/OperationValidationAndFallback.cpp
```diff
@@ -126,34 +127,22 @@ class TTNNOperationValidationAndFallback
...
-    if (openedDeviceInPass) {
-      op_model::SingletonDeviceContext::getInstance().closeInstance();
-    }
-#endif
-  }
-
   void runOnOperation() override {
+    // Device lifecycle is managed by OpModelDeviceWrapperPass in the pipeline,
+    // but for standalone pass usage (e.g., in tests), the guard opens/closes it.
+    op_model::OpModelDeviceGuard deviceGuard;
```
odjuricicTT: Do we have a test with single optimizer pass invoked?

lib/Dialect/TTNN/Pipelines/TTNNPipelines.cpp
rpavlovicTT: Replaced non-op model path with llvm fatal error.

lib/Dialect/TTNN/Transforms/OptimizerPasses/OptimizerPassesWrapper.cpp
rpavlovicTT: Covers pass failures, but abortions/asserts are not covered as they would require special signal handling.

include/ttmlir/OpModel/TTNN/OpModelDeviceGuard.h
rpavlovicTT: Moved it to same file as SingletonDeviceContext and renamed to `ScopedSingletonDeviceGuard`.

lib/Dialect/TTNN/Transforms/OptimizerPasses/OperationValidationAndFallback.cpp
rpavlovicTT: We have just this one `test/ttmlir/Dialect/TTNN/optimizer/insert_memreconfig_const_eval.mlir` as far as I can tell.

vvukomanTT (approved):

lib/Dialect/TTNN/Transforms/OptimizerPasses/OperationValidationAndFallback.cpp
odjuricicTT: Can you add at least one more? Maybe on for op validation pass?

rpavlovicTT: For validation pass we have couple of tests here https://github.com/tenstorrent/tt-mlir/tree/main/test/ttmlir/Dialect/TTNN/optimizer/op_layout_fallbacks

tools/explorer/test/run_tests.py
```diff
@@ -176,9 +176,11 @@ def test_execute_model(model_path):
 
 
 def test_execute_mnist_df_sharding():
+    # Optimizer disabled because these tests are run with tracy build in CI, and tracy
```
brataTT, line 179: I'm not familiar with how optimizer passes work, but is enabling optimizer incompatible with tracy build somehow? If so, I'll assign myself an issue to figure out another way to test this.

Also, do you know more about why these tests might start failing after this PR and not before?

rpavlovicTT: Couple of things here lead to this
1. explorer tests require "tracy" build because that build involves building perf infra, so explorer can run `ttrt perf`
2. optimizer component requires `OPMODEL` compile flag to be defined, so it builds OpModelLib which is a requirement for optimizer to be working
3. so far, we had non-op-model path where optimizer would just be skipped, but as of this commit we will throw an error if a user compiles with optimizer in unsuitable build
4. I tried enabling `OPMODEL` in tracy build as well, but this leads to very long testing times in CI. Reason is that llvm-lit tests (which are more than 1000) are run sequentially when `OPMODEL=ON` because of device correspondence (we can't use parallelism)
5. So, I just decided to disable optimizer in these tests -- I don't think we lose much.

brataTT: Thanks for the detailed explanation Radenko
And for trying to enable opmodel in tracy build
Filed [[an issue]](https://github.com/tenstorrent/tt-mlir/issues/5470) to myself to add this test back once we find a workaround

brataTT (approved):

Issue #5137, opened 2025-09-29
Finalize TTIR -> D2M Refactor


vwellsTT:
We've added new dedicated D2M dialect for Generic ops etc, but there a few cleanups left to do.
1. we need to make it a full conversion by adding a MeshShardOp to D2M'
2. we need to remove bufferization from all TTIR Ops
3. we need to decide how to handle TTNNMetalLayoutCastOp--probably should more to ttcore and be unrelated
4. decide if we need constantOp, fullOp, etc. in D2M

Issue #2978, opened 2025-04-16
consider using DstBufferizableOpInterfaceExternalModel for bufferizable impl of DPS ops


vroubtsovTT:
### Is your feature request related to a problem? Please describe

Many (all?) ttir ops get their `BufferizableOpInterface` implementation from `TTIR_BufferizableOp` which injects 5 methods into every op class via `extraClassDefinition` hooks. It might be possible to simplify the implementation by delegating to [a model class provided by MLIR specifically for DPS dialects](https://github.com/llvm/llvm-project/blob/main/mlir/docs/Bufferization.md) which is claimed to reduce the impl burden to a single method:

> Interface implementations of DPS ops (that implement DestinationStyleOpInterface) can derive from DstBufferizableOpInterfaceExternalModel, which provides all necessary method implementations except for bufferize.

I believe this model was added as part of one short bufferization and so seems to be future-proof.

### Describe the solution you'd like

Available docs on using external modes for interface impl:

- https://mlir.llvm.org/docs/Interfaces/#external-models-for-attribute-operation-and-type-interfaces
- mentions on several slides of this 2022 deck: https://mlir.llvm.org/OpenMeetings/2022-01-13-One-Shot-Bufferization.pdf

PR #5362, opened 2025-10-15
[D2M] refactor to cleanup bufferization w.r.t. TTIR vs D2M

vwellsTT:
### Ticket
Closes https://github.com/tenstorrent/tt-mlir/issues/5137

### Problem description
When refactoring D2M ops out of TTIR, there were a number of ops which were duplicated, etc.  Some of these had a bufferization interface which is only needed in D2M, not TTIR.  Others were not duplicated but need to be.

### What's changed
Remove bufferization logic for all TTIR ops--only D2M ops should get bufferized.  Copy any ops missing from D2M which required bufferization + add conversions for them.

### Checklist

### DIFF
--- a/lib/Dialect/TTCore/IR/Utils.cpp
+++ b/lib/Dialect/TTCore/IR/Utils.cpp
@@ -202,7 +202,7 @@ MemRefType getBufferType(Type type, bool isView,
 
   auto layout = mlir::cast<MetalLayoutAttr>(tensorType.getEncoding());
 
-  SmallVector<int64_t> fullMemrefShape = tensorType.getShape();
+  SmallVector<int64_t> fullMemrefShape(tensorType.getShape());
 
   // Create the appropriate layout attribute based on whether this is a view or
   // a materialized buffer. Views use affine maps for flexible indexing, while


lib/Dialect/D2M/IR/D2MOps.cpp
```diff
@@ -186,6 +198,125 @@ d2m::EmptyOp::getBufferType(mlir::Value value,
...
+}
+
+mlir::FailureOr<mlir::BaseMemRefType> d2m::MeshShardOp::getBufferType(
+    mlir::Value value, const mlir::bufferization::BufferizationOptions &,
+    const mlir::bufferization::BufferizationState &,
+    ::llvm::SmallVector<mlir::Value> &) {
+  return d2m::getBufferType(value.getType(), false);
+}
+
+::mlir::OpFoldResult d2m::MeshShardOp::fold(FoldAdaptor adaptor) {
```
vwellsTT, line 233: remark: this func, and below, are direct copy-pasted from `ttir::MeshShardOp`.  This seems like a pretty prevalent pattern in our ops code--a lot of common reuse between TTIR and TTNN, or TTIR and D2M, etc. for the same ops across different dialects.  Currently, we just duplicate the code, but I wonder if we should consider consolidating some shared locations for impls of such funcs shared by multiple dialects, and then here just invoke `shared::MeshShardOpfold(adaptor)`.

include/ttmlir/Dialect/D2M/IR/D2MOps.td
```diff
@@ -290,4 +290,81 @@ def D2M_EmptyOp : D2M_Op<"empty", [
   let assemblyFormat = "`(` `)` attr-dict `:` type($result)";
 }
 
+def D2M_FullOp : D2M_Op<"full", [
```
nsmithtt, line 293: I suppose this is just carried forward w/ current behavior.  But I can't tell if/how the FullOp is actually handled or lowerable.

vwellsTT: Yeah, was thinking adding runtime support would be follow-on work, but maybe that's a bad paradigm idk.  Was not aiming to change any RT functionality with this PR, just finish up refactor

vwellsTT: But I haven't added any lowering to TTKernel/TTMetal as of now

vwellsTT: Actually, I guess maybe it's meant to just bufferize into a `memref` op and that gets supported at RT, come to think of it--I thought we did support `ttir.constant` op before, and I think that's how that works.
I'm not sure if we had any tests which covered this ops in TTIR before.  But in the canonicalizer, I see that ttir.constant gets replaced by ttir.full, so it seemed like I could just support d2m.full here and it should achieve parity w/ old behavior

vwellsTT: I guess `test/ttmlir/Silicon/TTMetal/n150/simple_constant.mlir` should cover this, locally all lit tests worked

lib/Dialect/D2M/IR/D2MOps.cpp
wenbinlyuTT: Just a question: what's the relationship between this function and the other `getBufferType()` in `TTIROps.cpp`? They look exceedingly similar and I'm confused why we have two copies.

vwellsTT: oh, good point--I think I should remove the TTIR one now

vwellsTT: Annoyingly, since `TTNNMetalLayoutCastOp` is still in `ttir`, it's pretty painful to do so today.  I believe the plan is to move that to `ttcore` at some point.  Maybe I can refactor it into a shared util file for now or something.

include/ttmlir/Dialect/D2M/IR/D2MOps.td
nsmithtt: Ok, well we can revisit the full op separately since this PR is just maintaining the status quo.

wenbinlyuTT (approved):
LGTM Thanks!

vroubtsovTT:
I don't know if we're ever in a situation where `DstBufferizableOpInterfaceExternalModel ` would've helped us: https://github.com/tenstorrent/tt-mlir/issues/2978

lib/Dialect/TTCore/IR/Utils.cpp
```diff
@@ -170,4 +171,68 @@ llvm::SmallVector<int64_t, 2> collapseGridTo2D(ArrayRef<int64_t> gridShape) {
...
+                           hostLayout);
+  }
+
+  auto layout = mlir::cast<MetalLayoutAttr>(tensorType.getEncoding());
+
+  auto gridShape = layout.getGridShape(tensorType);
+  auto shardShape = layout.getShardShape(tensorType);
+  SmallVector<int64_t> fullMemrefShape;
+  fullMemrefShape.append(gridShape.begin(), gridShape.end());
+  fullMemrefShape.append(shardShape.begin(), shardShape.end());
```
vroubtsovTT: is this concatenation better (more robust?) than simply using `tensorType.getShape()`?

svuckovicTT (approved):

lib/Dialect/TTCore/IR/Utils.cpp
vwellsTT: hmm, good point.  This is just a copy-paste, but I don't see any reason it wouldn't be 1:1, probably something got removed in the past or something.  I'll try just replacing it w/ tensorType.getShape()

vwellsTT:
> Consider using TT_assert/debug ;)

Ha, this is just refactoring code around, not adding anything new, I'll leave the mass-replacement to you 😛

PR #5364, opened 2025-10-15
D2M: remove fails_golden on reduction tests and constrain inputs to TF32 friendly vals

jdesousa-TT:
### Problem description
Golden checks are disabled for d2m reductions.

### What's changed
Enable these tests and use restricted golden vals and atol parameter to ensure they pass CI. 

### Checklist

### DIFF
--- a/tools/ttrt/common/callback.py
+++ b/tools/ttrt/common/callback.py
@@ -171,7 +171,11 @@ def golden(callback_runtime_config, binary, program_context, op_context):
             return
 
         _, _, cal_pcc, output_str = get_atol_rtol_pcc(
-            golden_tensor_torch, output_tensor_torch, logging
+            golden_tensor_torch,
+            output_tensor_torch,
+            callback_runtime_config.atol,
+            callback_runtime_config.rtol,
+            logging,
         )
 
         # Handle case where tensor has only one element.
--- a/tools/ttrt/common/run.py
+++ b/tools/ttrt/common/run.py
@@ -940,6 +940,8 @@ class Run:
                                     _, _, cal_pcc, _ = get_atol_rtol_pcc(
                                         golden_tensor_torch,
                                         output_tensor_torch,
+                                        self["atol"],
+                                        self["rtol"],
                                         self.logging,
                                     )
                                     pcc_fail = (


test/python/golden/test_metal_reductions.py
```diff
@@ -36,24 +59,16 @@ def test_sum(
...
-        reduce_sum,
+        create_reductions_constrained_inputs(shape, "sum", dim_arg, keep_dim),
         [shape],
         target=target,
         test_base=request.node.name,
         print_ir=True,
         output_root=request.config.getoption("--path"),
         system_desc_path=request.config.getoption("--sys-desc"),
         device=device,
+        atol=5,
```
wenbinlyuTT: This feels like a magic number, maybe we can change it to something like `shape[0] * shape[1] * 0.00005` so for the biggest reduction it's `8x16x1024x0.00005=6.5536`?

jdesousa-TT: This is a really good idea, since the error seems to scale with the amount of accumulation we do. Didn't think of making it dynamic, thanks! 0.00005 seems to be a bit too tight still for some inputs, so I relaxed it a little more, but this is better especially for smaller inputs.

tools/ttrt/common/util.py
```diff
@@ -83,7 +83,7 @@ def mask_torch_inf_nan(tensor):
     return tensor
 
 
-def get_atol_rtol_pcc(golden, calculated, logging):
+def get_atol_rtol_pcc(golden, calculated, atol, rtol, logging):
```
jgrimTT, line 86: Simple fix, but I'm a little concerned that CI tests didn't catch this since it's such a simple error (I made an [issue for it](https://github.com/tenstorrent/tt-mlir/issues/5391) @tapspatel), but this throws a missing positional arguments error when this function is called in `run.py` and `callback.py`.

jdesousa-TT: Hey @jgrimTT, I updated the usage of `get_atol_rtol_pcc` in `run.py` and `callback.py`, ptal :)

jgrimTT: Looks great, thx!

PR #5372, opened 2025-10-16
[uplift] update tt-metal commit hash upon dependency updates

gfengTT:
### Problem description
When `tt_metal/sfpi-version.sh` or `install_dependencies.sh` is changed
in tt-metal, we should update the commit hash in Dockerfile.

### What's changed
Compare the sha sum of the files in the current tt-metal version vs.
uplifted version. If they are different, update the commit hash in the
Dockerfile.

### DIFF
--- a/.github/Dockerfile.base
+++ b/.github/Dockerfile.base
@@ -4,7 +4,7 @@ SHELL ["/bin/bash", "-c"]
 # Set environment variables
 ENV DEBIAN_FRONTEND=noninteractive
 
-ARG TT_METAL_COMMIT=7aee9a53eac0023e3660668448e1fbb105038d63
+ARG TT_METAL_DEPENDENCIES_COMMIT=7aee9a53eac0023e3660668448e1fbb105038d63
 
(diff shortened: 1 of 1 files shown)


gfengTT:
@nsumrakTT would you mind reviewing again, given there's significant change?

.github/Dockerfile.base
```diff
@@ -4,6 +4,8 @@ SHELL ["/bin/bash", "-c"]
 # Set environment variables
 ENV DEBIAN_FRONTEND=noninteractive
 
+ARG TT_METAL_COMMIT=7aee9a53eac0023e3660668448e1fbb105038d63
```
nsumrakTT: Might be nitpicking:
I would prefer to rename this to TT_METAL_DEPENDENCIES_COMMIT just not to mix it up with TT_METAL_VERSION in CMakeFile.txt or metal_override as these might be totally different versions.

gfengTT: Great suggestion. Applied.

PR #5374, opened 2025-10-16
[ttnn.jit] simple JitCache 

vtangTT:
### Ticket
Link to Github Issue

### Problem description

### What's changed
- `JitCache` cpp impl, hashing off tensor metadata using same hash method as ttnn ops infra
- `JitFunction` python class, holds it's own `JitCache` so no need to hash off jit params or function signature 
- decorating with `jit` will now return a `JitFunction` which has a `__call__` impl 
    - each `JitFunction` owns it's own `JitCache`
    - on miss, will run ast traversal, necessary passes and translation to flatbuffer
    - removed dead jit flabs
- in mlir runtime, creation of `program_descriptor` is expensive so we cache it
    -  `Binary` now owns a `ProgramDescCache`

ToDo before merge: 
- [x] commonize `wrapInCapsule` since also used in `Passes.cpp`
- [x] needs https://github.com/tenstorrent/tt-metal/pull/30324 for caching to work e2e, or else lint error.

### Checklist

### DIFF
--- a/runtime/include/tt/runtime/detail/ttnn/types/program_desc_cache.h
+++ b/runtime/include/tt/runtime/detail/ttnn/types/program_desc_cache.h
@@ -21,10 +21,10 @@ public:
   ProgramDescCache(ProgramDescCache &&) = delete;
   ProgramDescCache &operator=(ProgramDescCache &&) = delete;
 
-  void *get(const std::size_t &hash) const {
+  std::shared_ptr<void> get(const std::size_t &hash) const {
     auto it = cache_.find(hash);
     if (it != cache_.end()) {
-      return it->second.get();
+      return it->second;
     }
     return nullptr;
   }
--- a/runtime/lib/ttnn/operations/generic/generic_op.cpp
+++ b/runtime/lib/ttnn/operations/generic/generic_op.cpp
@@ -216,22 +216,24 @@ createKernelDescriptor(const ::tt::target::ttnn::KernelDescriptor &kernelDesc,
   return kernelDescriptor;
 }
 
-static ::tt::tt_metal::ProgramDescriptor createProgramDescriptor(
+static std::shared_ptr<::tt::tt_metal::ProgramDescriptor>
+createProgramDescriptor(
     const ::tt::target::ttnn::ProgramDescriptor *programDesc,
     const std::vector<::ttnn::Tensor> &ioTensors) {
-  ::tt::tt_metal::ProgramDescriptor programDescriptor;
+  auto programDescriptor =
+      std::make_shared<::tt::tt_metal::ProgramDescriptor>();
   for (const tt::target::ttnn::KernelDescriptor *kernelDesc :
        *programDesc->kernels()) {
-    programDescriptor.kernels.push_back(
+    programDescriptor->kernels.push_back(
         createKernelDescriptor(*kernelDesc, ioTensors));
   }
   for (const tt::target::ttnn::KernelCBDescriptor *cbDesc :
        *programDesc->cbs()) {
-    programDescriptor.cbs.push_back(createCBDescriptor(*cbDesc, ioTensors));
+    programDescriptor->cbs.push_back(createCBDescriptor(*cbDesc, ioTensors));
   }
   for (const tt::target::ttnn::SemaphoreDescriptor *semaphoreDesc :
        *programDesc->semaphores()) {
-    programDescriptor.semaphores.push_back(
+    programDescriptor->semaphores.push_back(
         createSemaphoreDescriptor(*semaphoreDesc));
   }
   return programDescriptor;
@@ -247,30 +249,28 @@ void run(const ::tt::target::ttnn::GenericOp *op, ProgramContext &context) {
   }
 
   auto programDescCache = context.getExecutableHandle().getProgramDescCache();
-  std::shared_ptr<tt::runtime::ProgramDescCache> cache = programDescCache;
 
   auto *programDesc = op->program();
   std::size_t hash =
-      ttsl::hash::hash_objects_with_default_seed(programDesc, cache);
-  void *cachedPtr = cache ? cache->get(hash) : nullptr;
+      ttsl::hash::hash_objects_with_default_seed(programDesc, programDescCache);
+  std::shared_ptr<void> cachedPtr = programDescCache->get(hash);
 
-  ::tt::tt_metal::ProgramDescriptor programDescriptor;
+  std::shared_ptr<::tt::tt_metal::ProgramDescriptor> programDescriptor;
   if (cachedPtr) {
+    LOG_DEBUG("Cache hit for program descriptor");
     programDescriptor =
-        *static_cast<::tt::tt_metal::ProgramDescriptor *>(cachedPtr);
+        std::static_pointer_cast<::tt::tt_metal::ProgramDescriptor>(cachedPtr);
   } else {
+    LOG_DEBUG("Cache miss for program descriptor");
     programDescriptor = createProgramDescriptor(programDesc, ioTensors);
-    programDescriptor.custom_program_hash =
+    programDescriptor->custom_program_hash =
         reinterpret_cast<ttsl::hash::hash_t>(hash);
-    if (cache) {
-      auto heapCopy = std::make_shared<::tt::tt_metal::ProgramDescriptor>(
-          programDescriptor);
-      cache->insert(hash, std::move(heapCopy));
-    }
+    programDescCache->insert(hash,
+                             std::static_pointer_cast<void>(programDescriptor));
   }
 
   ::ttnn::Tensor outputTensor =
-      ::ttnn::generic_op(ioTensors, programDescriptor);
+      ::ttnn::generic_op(ioTensors, *programDescriptor);
   tensorPool.insertTTNNTensorAndValidate(op->io_tensors()->Get(size - 1),
                                          outputTensor);
 }


tools/ttnn-jit/csrc/lib/jit_cache.cpp
```diff
@@ -0,0 +1,64 @@
...
+  mlir::tt::registerAllExtensions(registry);
+}
+
+void JitCache::compile(Operation *op, std::string options) {
+  mlir::PassManager pm(op->getName());
+  mlir::MLIRContext *context = op->getContext();
+  context->appendDialectRegistry(registry);
+  context->loadAllAvailableDialects();
+  pm.addPass(tt::createConvertTTNNToTTIRPass());
+  const auto *pipeline =
```
vtangTT, line 29: Is there a better way to run this pipeline, without pipeline lookup or just in general.

If I do something like `tt::createConvertTTIRToTTMetalPipeline(pm, options));`, there will be RTTI issues with `TTIRToTTMetalPipelineOptions`

tools/ttnn-jit/csrc/__init__.cpp
```diff
@@ -0,0 +1,88 @@
...
+#include <pybind11/cast.h>
+#include <pybind11/pybind11.h>
+#include <pybind11/pytypes.h>
+#pragma clang diagnostic pop
+
+namespace nb = nanobind;
+namespace py = pybind11;
+
+namespace mlir::tt::ttnn::jit {
+static nb::capsule wrapInCapsule(std::shared_ptr<void> underlying) {
```
vtangTT: unsure what the best place to commonize this would be, since it's also used in `Passes.cpp` .. open to suggestions.

nsmithtt: Maybe `include/ttmlir/Target/Common/Target.h`?

runtime/lib/ttnn/operations/generic/generic_op.cpp
```diff
@@ -216,6 +216,9 @@ createKernelDescriptor(const ::tt::target::ttnn::KernelDescriptor &kernelDesc,
 }
 
 void run(const ::tt::target::ttnn::GenericOp *op, ProgramContext &context) {
+  static std::unordered_map<const ::tt::target::ttnn::ProgramDescriptor *,
+                            tt::tt_metal::ProgramDescriptor>
+      cachedProgramDescs;
```
nsmithtt: I think we need to put this inside of the device context or something.

vtangTT: Ok yea, I was unsure whether I should so just left it in global scope. I'll make it similar to `TraceCache`

tools/ttnn-jit/csrc/__init__.cpp
vtangTT: added to `include/ttmlir/Bindings/Python/Utils.h`

runtime/lib/ttnn/operations/generic/generic_op.cpp
vtangTT: changed ptal

include/ttmlir/Target/Python/Utils.h
nsmithtt, line 1: I'd prefer this actually go into `include/ttmlir/Target/Utils/Utils.h`, the Target subdirectory is the only area where we allow sharing code between compiler and runtime.

runtime/include/tt/runtime/detail/ttnn/types/program_desc_cache.h
```diff
@@ -0,0 +1,37 @@
...
+
+  void insert(const ::tt::target::ttnn::ProgramDescriptor *programDesc,
+              const ::tt::tt_metal::ProgramDescriptor &programDescriptor);
+
+private:
+  // vector lookup faster for smaller sizes
+  static constexpr size_t MAX_SIZE = 128;
+  std::vector<std::pair<const ::tt::target::ttnn::ProgramDescriptor *,
+                        ::tt::tt_metal::ProgramDescriptor>>
+      cache_;
```
nsmithtt: I hate to say it, but probably for now let's just use an unordered_map because it fits the abstraction better and can grow infinitely.  If we run into perf issues we should write our own DenseMap implementation which can toggle between different underlying datastructures depending on element count.

runtime/include/tt/runtime/types.h
```diff
@@ -374,17 +374,33 @@ struct TraceCache : public detail::RuntimeCheckedObjectImpl {
   using detail::RuntimeCheckedObjectImpl::RuntimeCheckedObjectImpl;
 };
 
+struct ProgramDescCache : public detail::RuntimeCheckedObjectImpl {
```
nsmithtt: Why do we need to inherit from `RuntimeCheckedObjectImpl`?  This inheritance is only meant for public API types, `ProgramDescCache` should be an internal only type.  @jnie-TT do we have a conventional place for internal data structures?

Edit: I see now that this is just copying what TraceCache did, perhaps this is intentional for some reason @jnie-TT?

jnie-TT: We do have places for internal data structures, but this gets a little tricky when the internal data structure needs to be a member of a public API type. 

`TraceCache` was added here because it's a part of the public API type `Device`. `TraceCache` is a type-erased representation of `tt::runtime::ttnn::TraceCache` which holds backend-specific constructs like `ttnn::Tensor` `ttnn::MeshDevice` etc.

I think it makes more sense to move these types into `namespace detail` since they shouldn't be public-facing, but we need to reference them in public-facing types.

vtangTT: Do you mean to move both `TraceCache` and `ProgramDescCache`?

jnie-TT: @vtangTT yeah, just so that it's clear that these types are meant for internal use only and the user needn't be aware of them.

@nsmithtt does this work for you?

nsmithtt: Yeah moving them into a detail ns is exactly what I was thinking too.  Let's do that.

include/ttmlir/Target/Python/Utils.h
```diff
@@ -0,0 +1,37 @@
...
+#pragma clang diagnostic ignored "-Wcovered-switch-default"
+#include <nanobind/nanobind.h>
+#include <nanobind/stl/shared_ptr.h>
+#pragma clang diagnostic pop
+
+namespace nb = nanobind;
+
+namespace mlir::ttmlir::python {
+
+inline nb::capsule wrapInCapsule(std::shared_ptr<void> underlying) {
```
nsmithtt, line 23: FYI @ctodTT, once this PR lands, we can use this same `wrapInCapsule` trick for passing the program binary from the compiler to the runtime without having to bounce through the disk.

.github/test_scripts/pykernel.sh
```diff
@@ -13,5 +13,5 @@ ln -sf $INSTALL_DIR/tt-metal third_party/tt-metal/src/tt-metal
 
 echo "Running PyKernel tests..."
 pytest -v $WORK_DIR/test/pykernel/demo/test.py --junit-xml=$TEST_REPORT_PATH
-pytest -v $WORK_DIR/test/ttnn-jit/test_eltwise.py --junit-xml=${TEST_REPORT_PATH%_*}_ttnn_jit_${TEST_REPORT_PATH##*_}
+pytest -v $WORK_DIR/test/ttnn-jit/ --junit-xml=${TEST_REPORT_PATH%_*}_ttnn_jit_${TEST_REPORT_PATH##*_}
```
nsumrakTT: Why not:
```
pytest -v $WORK_DIR/test/pykernel/demo/test.py $WORK_DIR/test/ttnn-jit/ --junit-xml=$TEST_REPORT_PATH`
```
or
change test script to accept argument and add another test for ttnn-jit:
```
{ "runs-on": "n150", "image": "tracy", "script": "pykernel.sh", arg: "test/pykernel/demo/test.py" },
{ "runs-on": "n150", "image": "tracy", "script": "pykernel.sh", arg: "test/ttnn-jit/" }
```

include/ttmlir/Target/Python/Utils.h
vtangTT: what about a new file in `include/ttmlir/Target/Python/`, as `include/ttmlir/Target/Utils/Utils.h` is used in both `TTMetalToFlatbuffer` and `TTNNToFlatbuffer` and I don't want to expose a nb deps to those files.

nsmithtt: Yeah that sounds good to me!

nsmithtt: Outside of scope for this PR, but arguably we should move runtime structs, `Binary`, `SystemDesc`, `Flatbuffer` to this new include too.  I know we've avoided sharing any code between runtime and compiler, but the flatbuffer technically is shared code so having the wrapper helpers be shared might be an acceptable case.  They would no longer inherit from `ObjectImpl` though. @jnie-TT thoughts?

jnie-TT: @nsmithtt that works for me, though I'm curious if the compiler has any use-cases of these data structures at all? If so we should definitely move to a shared location.

runtime/include/tt/runtime/types.h
vtangTT: I've moved both into the detail namespace above in the same file.. ptal I'm unsure if this is what was meant 😅

runtime/include/tt/runtime/detail/ttnn/types/program_desc_cache.h
vtangTT: changed

include/ttmlir/Target/Python/Utils.h
nsmithtt: The only compelling reason is that they can then share a nanobind serializable type so they we don't have to do this trick of wrapping it inside of a capsule.  That said, it does seem to be working for now, and you're right, there aren't other use cases for the compiler at this point.

runtime/include/tt/runtime/types.h
nsmithtt: Oh I meant to move TraceCache and ProgramDescCache into a detail namespace and they don't inherit from `RuntimeCheckedObjectImpl`.  There's no need for these to inherit from this base-class.

jnie-TT: Since this still needs to be type erased, I think it would make sense to inherit from `detail::ObjectImpl`, and as @nsmithtt said there's no need to inherit from `detail:: RuntimeCheckedObjectImpl` anymore since the runtime check should be done within the parent class, in this case `Device`

tapspatel (approved):
python folder looks good

runtime/include/tt/runtime/types.h
nsmithtt: @jnie-TT why does it need to be type erased?  The intention of `ObjectImpl` is for types that are directly returned to users and are part of API.  For things like cache, imo they should just be in a detail namespace ie. not API, but could be used as private members of API types.  For example:

```c++
namespace detail {
struct ProgramDescCache {
...
};
}

...

struct Device : public detail::RuntimeCheckedObjectImpl {

private:
  detail::ProgramDescCache programDescCache;
};
```

nsmithtt: So all that said, I do see now that TraceCache is API (is it meant to be used by users?).  I don't think ProgramDescCache needs to be API though, correct me if I'm wrong.

jnie-TT: @nsmithtt these need to be type-erased because they contain ttnn/ttmetal types, which we don't want to expose to the users. For example `TraceCache` contains `::ttnn::MeshTraceId`, I believe `ProgramDescCache` contains a metal type as well. `TraceCache` shouldn't be API and so I believe we're moving it into detail and making it private.

nsmithtt: If it's in detail though it's implicitly not API and users shouldn't touch it / ever be vended this object, so it's OK for them to include whatever types they need.  TraceCache however seems to be API because a user can access it via Device method:
```c++
std::shared_ptr<TraceCache> getTraceCache() { return traceCache; }
```

So TraceCache should not be moved to detail.

runtime/include/tt/runtime/types.h
```diff
@@ -370,21 +378,30 @@ struct Binary : public Flatbuffer {
...
+         std::shared_ptr<detail::ProgramDescCache> programDescCache,
          DeviceRuntime runtime)
       : detail::RuntimeCheckedObjectImpl(handle, runtime),
-        globalId(nextDeviceGlobalId()), traceCache(traceCache) {}
+        globalId(nextDeviceGlobalId()), traceCache(traceCache),
+        programDescCache(programDescCache) {}
 
-  std::shared_ptr<TraceCache> getTraceCache() { return traceCache; }
+  std::shared_ptr<detail::TraceCache> getTraceCache() { return traceCache; }
+  std::shared_ptr<detail::ProgramDescCache> getProgramDescCache() {
```
nsmithtt: This should not be an API method, i.e. we do not intend for users to ever retrieve the program desc cache.  Also we should never vend something from detail namespace.  I called this out on the other thread too, but additionally noting it here.

runtime/include/tt/runtime/types.h
nsmithtt: For example:
- `MyInternalType`: Example of an internal detail type, can _only_ be used as a private member of an API type.
- `MyAPIType` / `APIObject`: Example of API types and legal/illegal usages of other types.

```c++
namespace detail {
struct MyInternalType {
  // can use whatever other internal types you want...
};
}

struct MyAPIType {
public:
  // can only use/vend other public / API types here ...

private:
  // can use whatever you want from detail ...
};

struct APIObject : public detail::RuntimeCheckedObjectImpl {
public:
  detail::MyInternalType getMyInternalType(); // NOT OK! A public method means a user can use it! detail has leaked!
  MyAPIType getMyAPIType(); // this is OK!

private:
  detail:: MyInternalType myInternalType;  // this is OK!
  MyAPIType myAPIType; // this is OK!
};
```

include/ttmlir/Target/Python/Utils.h
```diff
@@ -0,0 +1,37 @@
+// SPDX-FileCopyrightText: (c) 2024 Tenstorrent AI ULC
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#ifndef TTMLIR_BINDINGS_PYTHON_UTILS_H
+#define TTMLIR_BINDINGS_PYTHON_UTILS_H
```
vmilosevic: :warning: **llvm\-header\-guard** :warning:
```suggestion
#ifndef TTMLIR_TARGET_PYTHON_UTILS_H
#define TTMLIR_TARGET_PYTHON_UTILS_H
```

include/ttmlir/Target/Python/Utils.h
```diff
@@ -0,0 +1,37 @@
...
+  *binary = underlying;
+  return nb::capsule(
+      static_cast<void *>(
+          binary), // NOLINTNEXTLINE(cppcoreguidelines-owning-memory)
+      +[](void *data) noexcept { std::free(data); });
+}
+
+} // namespace mlir::ttmlir::python
+
+#endif // TTMLIR_BINDINGS_PYTHON_UTILS_H
```
vmilosevic: :warning: **llvm\-header\-guard** :warning:
```suggestion
#endif // TTMLIR_TARGET_PYTHON_UTILS_H
```

include/ttmlir/Target/Python/Utils.h
vtangTT: fixed

include/ttmlir/Target/Python/Utils.h
vtangTT: fixed.

runtime/include/tt/runtime/types.h
vtangTT: as discussed offline, 
- reverted any changes to TraceCache
- ProgramDescCache type erases metal types with void*
- Binary now owns ProgramDescCache

vmilosevic (approved):

runtime/include/tt/runtime/types.h
```diff
@@ -361,13 +365,21 @@ struct Binary : public Flatbuffer {
   // Get the tensor cache associated with this binary
   std::shared_ptr<TensorCache> getConstEvalTensorCache() { return tensorCache; }
 
+  // Get the program descriptor cache associated with this binary
+  std::shared_ptr<tt::runtime::ttnn::ProgramDescCache> getProgramDescCache() {
```
nsmithtt: I think if we eventually want to use this for metal runtime too, this should live in namespace `tt::runtime`

vtangTT: changed, also made it header only.

@jnie-TT should I move `runtime/include/tt/runtime/detail/ttnn/types/program_desc_cache.h` into `runtime/include/tt/runtime/` w/ `tensor_cache.h` or leave it where it is for now and move it in the future once metal runtime needs it.

jnie-TT: @vtangTT If we want to move it to a common location I think `runtime/include/tt/runtime/detail/common` would be a good place, since from what I understand it shouldn't be user-facing. I'm slightly leaning towards moving it now since 1. it shouldn't be ttnn-specific and 2. so that we don't forget later and accidentally implement the same structure for metal runtime

runtime/lib/ttnn/operations/generic/generic_op.cpp
```diff
@@ -241,6 +234,41 @@ void run(const ::tt::target::ttnn::GenericOp *op, ProgramContext &context) {
...
+  ProgramTensorPool &tensorPool = context.getTensorPool();
+  auto size = op->io_tensors()->size();
+  std::vector<::ttnn::Tensor> ioTensors(size);
+  for (unsigned int i = 0; i < size; i++) {
+    ioTensors[i] =
+        tensorPool.getTTNNTensorAndValidate(op->io_tensors()->Get(i));
+  }
+
+  auto programDescCache = context.getExecutableHandle().getProgramDescCache();
+  std::shared_ptr<tt::runtime::ProgramDescCache> cache = programDescCache;
```
jnie-TT: Seems this might be redundant, can we use programDescCache directly?

runtime/lib/ttnn/operations/generic/generic_op.cpp
```diff
@@ -241,6 +234,41 @@ void run(const ::tt::target::ttnn::GenericOp *op, ProgramContext &context) {
...
+
+  auto *programDesc = op->program();
+  std::size_t hash =
+      ttsl::hash::hash_objects_with_default_seed(programDesc, cache);
+  void *cachedPtr = cache ? cache->get(hash) : nullptr;
+
+  ::tt::tt_metal::ProgramDescriptor programDescriptor;
+  if (cachedPtr) {
+    programDescriptor =
+        *static_cast<::tt::tt_metal::ProgramDescriptor *>(cachedPtr);
```
jnie-TT: Seems like `::tt::tt_metal::ProgramDescriptor` is a non-trivial copy. Could we use shared pointers here instead of values to eliminate the copy? So maybe update `cache->get` to return a shared pointer, and update `createProgramDescriptor` to return a shared pointer as well, this way we won't need to make a separate heapCopy when inserting into the cache later either.

runtime/lib/ttnn/operations/generic/generic_op.cpp
vtangTT: fixed ty 🙏

runtime/lib/ttnn/operations/generic/generic_op.cpp
vtangTT: Yep this was sloppy by me. Changed, ptal 🙏

arminaleTT (approved):

PR #5376, opened 2025-10-16
V0 of the dest register analysis pass

### LINKED ISSUE #5108: [d2m] split and refactor TTIRInsertDstRegisterAccess into analysis pass

dlokeTT:
### Ticket
Closes #5108 

### Problem description
Create an analysis pass that analyzes how many destination register slots will be used in a fused kernel. 

### What's changed
- ElementwiseFusion now tries to aggresively fuse as many ops together as possible not focusing on block sizes but still checking that one block will fit in dest
- Analysis pass does the same logic currently done in InsertDstRegister but passes the # of tiles used to GenericTileComputeLoops
- GenericTileComputeLoops then uses that # to identify how to subblock DST
- Currently the heuristic is to just fuse as many ops as possible and then see if we can do loop tiling. Future versions of the analysis pass should provide information on different ways we can schedule the work including spills to L1 along with a cost model to decide what the best option is.

### Checklist


mbagherbeikTT:
looks good to me. Thanks for getting this moving :)

nsmithtt:
@dlokeTT, shouldn't the InsertDstRegister pass be using this analysis too?

dlokeTT:
> @dlokeTT, shouldn't the InsertDstRegister pass be using this analysis too?

This was something that I was a little confused about. Based on our conversation, you had mentioned that insertDst actually does the mechanics around the inserting, I wasn't sure if that meant it also did the computation on which position to store results to or if it should get it from the analysis. Initially I had gone down that route and I can add that back.

nsmithtt:
> > @dlokeTT, shouldn't the InsertDstRegister pass be using this analysis too?
> 
> This was something that I was a little confused about. Based on our conversation, you had mentioned that insertDst actually does the mechanics around the inserting, I wasn't sure if that meant it also did the computation on which position to store results to or if it should get it from the analysis. Initially I had gone down that route and I can add that back.

Yeah sorry for the confusion.  I think one of the goals of this is to move the current analysis that's being done inside of InsertDst pass, out of that pass into a dedicated analysis pass that can be reused with the intention of also being used by the InsertDst pass itself.

dlokeTT:
> > > @dlokeTT, shouldn't the InsertDstRegister pass be using this analysis too?
> > 
> > 
> > This was something that I was a little confused about. Based on our conversation, you had mentioned that insertDst actually does the mechanics around the inserting, I wasn't sure if that meant it also did the computation on which position to store results to or if it should get it from the analysis. Initially I had gone down that route and I can add that back.
> 
> Yeah sorry for the confusion. I think one of the goals of this is to move the current analysis that's being done inside of InsertDst pass, out of that pass into a dedicated analysis pass that can be reused with the intention of also being used by the InsertDst pass itself.

@nsmithtt Just updated it with this

lib/Dialect/D2M/Analysis/DestRegisterAnalysis.cpp
```diff
@@ -0,0 +1,103 @@
...
+    llvm::DenseMap<Value, int> valueToDstIndex;
+
+    // Phase 1: Collect all input values (from block args and non-compute ops).
+    llvm::SmallVector<Value> inputValues;
+    genericOp->walk([&](Operation *innerOp) {
+      if (innerOp->hasTrait<D2MGenericRegionComputeOpTrait>()) {
+        for (Value operand : innerOp->getOperands()) {
+          if (Operation *definingOp = operand.getDefiningOp()) {
+            if (!definingOp->hasTrait<D2MGenericRegionComputeOpTrait>()) {
+              if (valueToDstIndex.find(operand) == valueToDstIndex.end()) {
```
wenbinlyuTT: Will map to be non-empty during this walk? It seems that this `if` and the next one are not necessary.
Then, the three variable definitions above can be moved to just before phase 2, so it's a bit clearer when and why are they used for.

lib/Dialect/D2M/Analysis/DestRegisterAnalysis.cpp
```diff
@@ -0,0 +1,103 @@
...
+        llvm::SmallVector<int64_t> dstOperandIndices;
+        if (auto loadStoreInterface =
+                llvm::dyn_cast<OperandLoadStoreRegisterOpInterface>(innerOp)) {
+          dstOperandIndices =
+              loadStoreInterface.getOperandsLoadFromDstRegister();
+        }
+
+        // Allocate DST indices for operands that need to be loaded from DST
+        for (int64_t operandIdx : dstOperandIndices) {
+          if (operandIdx < static_cast<int64_t>(innerOp->getNumOperands())) {
```
wenbinlyuTT: Will this condition be evaluated to false ever? Feels like this should be an assertion or nothing at all.

lib/Dialect/D2M/Analysis/DestRegisterAnalysis.cpp
```diff
@@ -0,0 +1,103 @@
...
+            }
+          }
+        }
+
+        // Check if this is an in-place operation.
+        bool isInPlace = false;
+        if (auto loadStoreInterface =
+                llvm::dyn_cast<OperandLoadStoreRegisterOpInterface>(innerOp)) {
+          isInPlace = loadStoreInterface.getDstRegInPlace();
+        }
```
wenbinlyuTT: This can be moved up and combined with the other interface check.

lib/Dialect/D2M/Analysis/DestRegisterAnalysis.cpp
```diff
@@ -0,0 +1,103 @@
...
+
+        // Allocate or reuse destination index for the output.
+        int outputIndex;
+        if (isInPlace && innerOp->getNumOperands() > 0) {
```
wenbinlyuTT: Just curious, what ops inside the linalg generic op have 0 operands? We're not in TTKernel land so no init ops yet.

lib/Dialect/D2M/Analysis/DestRegisterAnalysis.cpp
```diff
@@ -0,0 +1,103 @@
...
+          Value operand = innerOp->getOperand(0);
+          if (Operation *definingOp = operand.getDefiningOp()) {
```
wenbinlyuTT: We can extract this out and turn the nested if-else into a simple one, just like the for loop above.

lib/Dialect/D2M/Analysis/DestRegisterAnalysis.cpp
```diff
@@ -0,0 +1,103 @@
...
+        bool isInPlace = false;
+        if (auto loadStoreInterface =
+                llvm::dyn_cast<OperandLoadStoreRegisterOpInterface>(innerOp)) {
+          isInPlace = loadStoreInterface.getDstRegInPlace();
+        }
+
+        // Allocate or reuse destination index for the output.
+        int outputIndex;
+        if (isInPlace && innerOp->getNumOperands() > 0) {
+          Value operand = innerOp->getOperand(0);
```
wenbinlyuTT: I don't really understand this code here, but since the CI is green so it must be something I've missed.

In the loop above, `valueToDstIndex` is populated with all the DST loads, but here we always check operand 0 only? Binary SFPU load from DST slice 0 & 1, while matmul & reduction load from DST slice 2. How are we getting matches in the `valueToDstIndex[operand]` query in these situations?

lib/Dialect/D2M/Analysis/DestRegisterAnalysis.cpp
```diff
@@ -0,0 +1,103 @@
...
+        } else {
+          outputIndex = nextAvailableIndex++;
+        }
+        opToDstIndex[innerOp] = outputIndex;
+        // Store the intermediate DST slice indices for this compute op.
+        info.dstSliceIndices.push_back(outputIndex);
+      }
+    });
+
+    info.dstMaxUsage = std::max(info.dstMaxUsage, nextAvailableIndex);
```
wenbinlyuTT: Why the `max`? `info` seems to be created for each `linalg.generic` and so `dstMaxUsage` will only ever be assigned once.

lib/Dialect/D2M/Analysis/DestRegisterAnalysis.cpp
```diff
@@ -0,0 +1,103 @@
...
+  op->walk([&](linalg::GenericOp genericOp) {
+    DstRegisterInfo info;
+
+    // Track destination register state.
+    int nextAvailableIndex = 0;
+    llvm::DenseMap<Operation *, int> opToDstIndex;
+    llvm::DenseMap<Value, int> valueToDstIndex;
+
+    // Phase 1: Collect all input values (from block args and non-compute ops).
+    llvm::SmallVector<Value> inputValues;
```
wenbinlyuTT: `inputValues` doesn't seem to be used later? Is phase 1 used for anything?

wenbinlyuTT (changes requested):
Thanks for working on this! I know you've done a lot of iterations but there will probably be a few rounds of reviews & revisions.

lib/Dialect/D2M/Analysis/DestRegisterAnalysis.cpp
dlokeTT: Yea this was used early on and forgot to delete in the refactor

lib/Dialect/D2M/Analysis/DestRegisterAnalysis.cpp
dlokeTT: Yeah true, this is an unnecessary check. Just removed it

lib/Dialect/D2M/Analysis/DestRegisterAnalysis.cpp
dlokeTT: Nope, also an unnecessary check. Removed

lib/Dialect/D2M/Analysis/DestRegisterAnalysis.cpp
dlokeTT: Unused, part of old code that didn't get deleted in refactor

lib/Dialect/D2M/Analysis/DestRegisterAnalysis.cpp
dlokeTT: Currently, only unary sfpu ops and matmul/reduction have isInPlace set to true so we don't go down this case if it is a binary sfpu and this matches current logic since binary ops have to store to next available. Matmul and reductions have isInPlace set to true because they are special cases where they do a load/store on DST slice 0 since they accumulate/spill results

dlokeTT: Matmul/reduction inputs are loaded into src a and b so we don't consider their inputs for allocation in dst which is why they load/store to slice 0 in dst

lib/Dialect/D2M/Analysis/DestRegisterAnalysis.cpp
dlokeTT: Good catch, you're right it's only assigned one

lib/Dialect/D2M/Analysis/DestRegisterAnalysis.cpp
wenbinlyuTT: For the binary SFPU ops, your're right and I was just dazzled by all the if's and indentations.
It's convenient that all the DST-in-place ops are using a single tile/slice, but I was just puzzled about why `valueToDstIndex[operand]` works here:
* Both matmul & reduction put operand 2 in DST, as `getOperandsLoadFromDstRegister` would return `{2}`, so in the loop above we record operand 2 into the map.
* Here we always query the map with operand 0, and somehow it worked?

dlokeTT: Oh yeah, I see what you're saying. It should be `Value operand = innerOp->getOperand(dstOperandIndices[0]);` since unary/matmul/reductions only use one tile in dst.  Not actually sure why this worked but everything passes even changing it to the correct thing

include/ttmlir/Dialect/D2M/Analysis/DestRegisterAnalysis.h
```diff
@@ -0,0 +1,34 @@
...
+/// Information about destination register usage for a GenericOp.
+struct DstRegisterInfo {
+  int dstMaxUsage = 0;
+  // Store intermediate DST slice indices instead of a map of operations.
+  // This avoids keeping stale Operation pointers around.
+  llvm::SmallVector<int> dstSliceIndices;
+};
+
+/// Analysis for determining destination register tile usage in GenericOps.
+struct DestRegisterAnalysis {
```
vwellsTT, line 27: nit: I think this should be a class since it's not a POD, per our usual style?

lib/Dialect/D2M/Analysis/DestRegisterAnalysis.cpp
```diff
@@ -0,0 +1,75 @@
...
+        llvm::SmallVector<int64_t> dstOperandIndices;
+        bool isInPlace = false;
+        if (auto loadStoreInterface =
+                llvm::dyn_cast<OperandLoadStoreRegisterOpInterface>(innerOp)) {
+          dstOperandIndices =
+              loadStoreInterface.getOperandsLoadFromDstRegister();
+          isInPlace = loadStoreInterface.getDstRegInPlace();
+        }
+
+        // Allocate DST indices for operands that need to be loaded from DST
```
vwellsTT: nit: .

lib/Dialect/D2M/Transforms/ElementwiseFusion.cpp
```diff
@@ -60,15 +53,14 @@ static bool fitsInDstPostFusion(GenericOp producer, GenericOp consumer,
   int numConsOperands = static_cast<int>(consumer.getNumOperands());
   int numProdOperands = static_cast<int>(producer.getNumOperands());
 
-  dstTilesRemaining -= blockSize * (numConsOperands + numProdOperands - 2 - 1);
+  dstTilesRemaining -= (numConsOperands + numProdOperands - 2 - 1);
```
vwellsTT, line 56: nit: I know this isn't really new to your code, but this line should have a comment explaining where `- 2 - 1` comes from

lib/Dialect/D2M/Transforms/GenericTileComputeLoops.cpp
```diff
@@ -109,47 +109,96 @@ static SmallVector<int64_t> calculateOptimalSubblockSizes(
...
+// Helper function to calculate output block shape based on subblock factor.
+static SmallVector<int64_t>
+calculateOutputBlockShape(ArrayRef<int64_t> outputShape,
+                          int64_t subblockFactor) {
+  SmallVector<int64_t> outputBlockShape;
+
+  if (outputShape[0] % subblockFactor == 0) {
+    // Distribute subblock factor across dimensions.
+    // First dimension is capped at the actual output shape.
+    int64_t dim0 = std::min(subblockFactor, outputShape[0]);
```
vwellsTT: nit: personally would prefer just calling push_back directly for both of these, vs binding to temporary variable which doesn't get used for anything else

lib/Dialect/D2M/Transforms/GenericTileComputeLoops.cpp
```diff
@@ -109,47 +109,96 @@ static SmallVector<int64_t> calculateOptimalSubblockSizes(
...
+    // can subblock by factor 2 (fitting 2 copies of 3 tiles = 6 tiles < 8).
+    bool optimizeSubblocking = false;
+    int subblockFactor = 1;
+    if (maxDstUsage > 0 && dstCapacity / maxDstUsage > 1) {
+      optimizeSubblocking = true;
+      subblockFactor = dstCapacity / maxDstUsage;
+    }
+
     SmallVector<int64_t> subblockSizes(outputTensor.getShape().size(), 1);
+    // Check if the first dimension is less than or divisible by the subblock
```
vwellsTT: nit: I'm not sure, but I kinda think this comment is better attached to `calculateOutputBlockShape` -- I was just scratching my head as to why that func works the way it does w/o the context

lib/Dialect/D2M/Transforms/GenericTileComputeLoops.cpp
```diff
@@ -170,6 +219,8 @@ struct D2MGenericComputeRewriter : public OpRewritePattern<linalg::GenericOp> {
   }
 
   unsigned maxDstPhysicalSizeTiles = 0;
+  const DestRegisterAnalysis *analysis;
+  size_t *genericOpIndexRef;
```
vwellsTT: using raw ptr to primitive is unusual pattern in modern C++, why is this needed? Can it be a `shared_ptr` etc?

lib/Dialect/D2M/Transforms/GenericTileComputeLoops.cpp
```diff
@@ -181,11 +232,19 @@ class D2MGenericTileComputeLoops
...
   void runOnOperation() final {
+
+    // Run the analysis once before any pattern rewriting
+    DestRegisterAnalysis &analysis = getAnalysis<DestRegisterAnalysis>();
+
     MLIRContext *ctx = &getContext();
     RewritePatternSet patterns(ctx);
-    patterns.add<D2MGenericComputeRewriter>(ctx,
-                                            maxDstPhysicalSizeTiles.getValue());
+    size_t genericOpIndex = 0;
```
vwellsTT: more confused now, this is only used once, why would it need to be passed as ptr?

lib/Dialect/D2M/Analysis/DestRegisterAnalysis.cpp
```diff
@@ -0,0 +1,75 @@
...
+                !definingOp->hasTrait<D2MGenericRegionComputeOpTrait>()) {
+              // This is an external input, allocate a DST index.
+              valueToDstIndex[operand] = nextAvailableIndex++;
+              info.dstSliceIndices.push_back(valueToDstIndex[operand]);
+            }
+          }
+        }
+
+        // Allocate or reuse destination index for the output.
+        int outputIndex;
```
vwellsTT: I know this is actually always initialized before use, but I would still give it some default value (or that's feedback others have left on my PRs in the past).
perhaps
```
int outputIndex = nextAvailableIndex;
if (isInPlace) { ... }
else { ++nextAvailableIndex; }
```
or something

lib/Dialect/D2M/Analysis/DestRegisterAnalysis.cpp
```diff
@@ -0,0 +1,75 @@
...
+        if (isInPlace) {
+          Value operand = innerOp->getOperand(dstOperandIndices[0]);
+          Operation *definingOp = operand.getDefiningOp();
```
vwellsTT: nit: prefer moving this decl into the below `if` statement to minimize its scope:
```
if (Operation *definingOp = operand.getDefiningOp(); definingOp && definingOp->hasTrait<D2MGenericRegionComputeOpTrait>())
```
ought to work I think

lib/Dialect/D2M/Transforms/GenericTileComputeLoops.cpp
```diff
@@ -109,47 +109,96 @@ static SmallVector<int64_t> calculateOptimalSubblockSizes(
...
         mlir::cast<MemRefType>(op.getOutputs().front().getType());
 
+    // Calculate DST capacity to determine if we can optimize subblocking.
+    Type largestDstType = utils::getRegionLargestDstElemType(op.getRegion());
+    const unsigned dstCapacity =
+        ttcore::getOpChipDescAttr(op).getDstLogicalSizeTiles(
+            largestDstType, false, maxDstPhysicalSizeTiles);
+
+    // Get max DST usage for this GenericOp from the analysis using the current
+    // index
```
vwellsTT: nit: .

lib/Dialect/D2M/Transforms/GenericTileComputeLoops.cpp
```diff
@@ -109,47 +109,96 @@ static SmallVector<int64_t> calculateOptimalSubblockSizes(
...
+    int maxDstUsage = 0;
+    if (*genericOpIndexRef < analysis->dstRegisterInfoList.size()) {
+      maxDstUsage =
+          analysis->dstRegisterInfoList[*genericOpIndexRef].dstMaxUsage;
+    }
+
+    // Increment index for the next generic op
```
vwellsTT: nit: .

vwellsTT:
Btw, Dhruv points out @wenbinlyuTT isn't in D2M codeowners, seems like he probably should be? 😄  @nsmithtt

lib/Dialect/D2M/Transforms/ElementwiseFusion.cpp
dlokeTT: The comments for this are above, I think the git diff hid it

vwellsTT: Ah, I see

include/ttmlir/Dialect/D2M/Analysis/DestRegisterAnalysis.h
dlokeTT: I just followed the MLIR example for the structure of the analysis pass https://mlir.llvm.org/docs/PassManagement/#analysis-management but

lib/Dialect/D2M/Transforms/GenericTileComputeLoops.cpp
dlokeTT: I use dim0 for calculating dim1. I could call push_back on dim1 but thought this was cleaner.

vwellsTT: Ah, I'm dumb; yeah personally would do 
```
dim0 = ...;
pushback(dim0);
push_back( calculation using dim0);
```
but it's def subjective

include/ttmlir/Dialect/D2M/Analysis/DestRegisterAnalysis.h
vwellsTT: I guess we have mix and match for other analyses today actually.  I don't have strong personal preference, but would be good to agree on a standard here imo

lib/Dialect/D2M/Transforms/GenericTileComputeLoops.cpp
dlokeTT: Yeah, you're right. Didn't move this in the refactor

include/ttmlir/Dialect/D2M/Analysis/DestRegisterAnalysis.h
dlokeTT: Yeah true, maybe @nsmithtt can chime in. From all the analysis in the D2M path it looks like `GenericOpBufferAnalysis` is the only one that uses a class

nsmithtt: Here's what llvm style guide says https://llvm.org/docs/CodingStandards.html#use-of-class-and-struct-keywords.

I'm kind of in agreement with it personally, I think usage of `struct` is fine up until you need to write the `public`/`private` labels, then it should become a class.

lib/Dialect/D2M/Analysis/DestRegisterAnalysis.cpp
```diff
@@ -0,0 +1,81 @@
...
+        }
+
+        // Allocate DST indices for operands that need to be loaded from DST.
+        for (int64_t operandIdx : dstOperandIndices) {
+          Value operand = innerOp->getOperand(operandIdx);
+          if (valueToDstIndex.find(operand) == valueToDstIndex.end()) {
+            if (Operation *definingOp = operand.getDefiningOp();
+                !definingOp ||
+                !definingOp->hasTrait<D2MGenericRegionComputeOpTrait>()) {
+              // Operation *definingOp = operand.getDefiningOp();
```
vwellsTT: remove?

lib/Dialect/D2M/Analysis/DestRegisterAnalysis.cpp
```diff
@@ -0,0 +1,81 @@
...
+        }
+
+        // Allocate or reuse destination index for the output.
+        int outputIndex = nextAvailableIndex;
+        if (isInPlace) {
+          Value operand = innerOp->getOperand(dstOperandIndices[0]);
+          if (Operation *definingOp = operand.getDefiningOp();
+              definingOp &&
+              definingOp->hasTrait<D2MGenericRegionComputeOpTrait>()) {
+            // Operation *definingOp = operand.getDefiningOp();
```
vwellsTT: remove

lib/Dialect/D2M/Transforms/GenericTileComputeLoops.cpp
```diff
@@ -181,11 +228,19 @@ class D2MGenericTileComputeLoops
...
   void runOnOperation() final {
+
+    // Run the analysis once before any pattern rewriting
+    DestRegisterAnalysis &analysis = getAnalysis<DestRegisterAnalysis>();
+
     MLIRContext *ctx = &getContext();
     RewritePatternSet patterns(ctx);
-    patterns.add<D2MGenericComputeRewriter>(ctx,
-                                            maxDstPhysicalSizeTiles.getValue());
+    std::shared_ptr<size_t> genericOpIndex = std::make_shared<size_t>(0);
```
vwellsTT: this still doesn't make sense to me, where's the other user? Why does it need to be passed in at all?

lib/Dialect/D2M/Transforms/InsertDstRegisterAccess.cpp
```diff
@@ -235,10 +241,13 @@ struct D2MInsertDstRegisterAccessRewriter final
   // Return both the copy nest info and dst allocation info.
   static DstAccessCollection
   collectDstAccesses(GenericOp op, Region &region,
-                     Operation *outermostInnerComputeLoop) {
+                     Operation *outermostInnerComputeLoop,
+                     const SmallVector<int> &dstSliceIndices) {
     CopyInfoMap copyInfos;
-    DstSliceAllocationState dstSliceAllocationState;
+    // DstSliceAllocationState dstSliceAllocationState;
```
vwellsTT: remove

lib/Dialect/D2M/Analysis/DestRegisterAnalysis.cpp
```diff
@@ -0,0 +1,81 @@
...
+#include "mlir/Dialect/Linalg/IR/Linalg.h"
+
+namespace mlir::tt::d2m {
+
+DestRegisterAnalysis::DestRegisterAnalysis(Operation *op) {
+  op->walk([&](linalg::GenericOp genericOp) {
+    DstRegisterInfo info;
+
+    // Track destination register state.
+    int nextAvailableIndex = 0;
```
wenbinlyuTT: Let's rename most of the 'index' in the variables' names in this analysis to 'slice', just so someone unfamiliar with with our DST management logic can get a good understanding from the get-go.

lib/Dialect/D2M/Analysis/DestRegisterAnalysis.cpp
```diff
@@ -0,0 +1,81 @@
...
+              valueToDstIndex[operand] = nextAvailableIndex++;
+              info.dstSliceIndices.push_back(valueToDstIndex[operand]);
+            }
+          }
+        }
+
+        // Allocate or reuse destination index for the output.
+        int outputIndex = nextAvailableIndex;
+        if (isInPlace) {
+          Value operand = innerOp->getOperand(dstOperandIndices[0]);
```
wenbinlyuTT, line 58: This probably won't be an actual issue, but let's add an `assert(dstOperandIndices.size() == 1)` in this branch, so the 'contract' is clear to the reader, instead of just a magical 0.

lib/Dialect/D2M/Transforms/InsertDstRegisterAccess.cpp
```diff
@@ -773,8 +776,15 @@ class D2MInsertDstRegisterAccess
...
+    // Retrieve the cached analysis from the pass manager instead of recomputing
+    auto analysisMaybe = getCachedAnalysis<DestRegisterAnalysis>();
+    if (!analysisMaybe.has_value()) {
+      return signalPassFailure();
+    }
+    DestRegisterAnalysis &analysis = analysisMaybe.value().get();
+
     patterns.add<D2MInsertDstRegisterAccessRewriter>(
-        ctx, useTileMatmul, maxDstPhysicalSizeTiles.getValue());
+        ctx, useTileMatmul, maxDstPhysicalSizeTiles.getValue(), &analysis);
```
vwellsTT: another case where I don't think you need a raw ptr.
More broadly, it'd be a better pattern to extract the data you want out of the analysis and pass that directly instead of the entire analysis, I believe.

lib/Dialect/D2M/Transforms/GenericTileComputeLoops.cpp
dlokeTT: Yeah true, I now do what `insertDstRegisterAccess.cpp` does and use a mutable

lib/Dialect/D2M/Transforms/InsertDstRegisterAccess.cpp
dlokeTT: Updated `genericTileComputeLoop.cpp` and `insertDstRegisterAccess.cpp` to follow this pattern.

lib/Dialect/D2M/Transforms/GenericTileComputeLoops.cpp
```diff
@@ -109,47 +109,90 @@ static SmallVector<int64_t> calculateOptimalSubblockSizes(
...
+    // Calculate DST capacity to determine if we can optimize subblocking.
+    Type largestDstType = utils::getRegionLargestDstElemType(op.getRegion());
+    const unsigned dstCapacity =
+        ttcore::getOpChipDescAttr(op).getDstLogicalSizeTiles(
+            largestDstType, false, maxDstPhysicalSizeTiles);
+
+    // Get max DST usage for this GenericOp from the analysis using the current
+    // index.
+    int maxDstUsage = 0;
+    if (genericOpIndex < dstRegisterInfoList.size()) {
```
wenbinlyuTT: When will this be false? We've just ran the analysis pass and so I expect the number of `linalg.generic` should be exact?

lib/Dialect/D2M/Transforms/GenericTileComputeLoops.cpp
```diff
@@ -109,47 +109,90 @@ static SmallVector<int64_t> calculateOptimalSubblockSizes(
...
+    genericOpIndex++;
+
+    // Enable subblocking optimization if DST can hold multiple copies of the
+    // usage pattern. For example, if dstCapacity = 8 and maxDstUsage = 3, we
+    // can subblock by factor 2 (fitting 2 copies of 3 tiles = 6 tiles < 8).
+    bool optimizeSubblocking = false;
+    int subblockFactor = 1;
+    if (maxDstUsage > 0 && dstCapacity / maxDstUsage > 1) {
+      optimizeSubblocking = true;
+      subblockFactor = dstCapacity / maxDstUsage;
```
wenbinlyuTT, line 182: By doing this we basically force 1 tile per DST slice correct? Does this affect the logic in the insert DST access pass?
Also we should leave a comment about this decision if it's intentional.

lib/Dialect/D2M/Transforms/GenericTileComputeLoops.cpp
dlokeTT: True, probably an unnecessary check

lib/Dialect/D2M/Transforms/GenericTileComputeLoops.cpp
```diff
@@ -109,49 +109,144 @@ static SmallVector<int64_t> calculateOptimalSubblockSizes(
...
+    Type largestDstType = utils::getRegionLargestDstElemType(op.getRegion());
+    const unsigned dstCapacity =
+        ttcore::getOpChipDescAttr(op).getDstLogicalSizeTiles(
+            largestDstType, false, maxDstPhysicalSizeTiles);
+
+    // Get max DST usage for this GenericOp from the analysis using the current
+    // index.
+    int maxDstUsage = 0;
+    if (genericOpIndex < dstRegisterInfoList.size()) {
+      maxDstUsage = dstRegisterInfoList[genericOpIndex].dstMaxUsage;
```
vwellsTT: I'm not sure I follow what `dstRegisterInfoList` is passed in from, but in general this doesn't strike me as a safe way to index--I believe `matchAndRewrite` order is not guaranteed, though I could be wrong.  What's the intent here? If you want to map values to specific generics, better to use a DenseMap or similar on the op itself I think.

lib/Dialect/D2M/Transforms/GenericTileComputeLoops.cpp
dlokeTT: Sorry wdym? If this check is true, we enable sub blocking and can do > 1 # of input tiles. If it's false, we force 1 input tile

lib/Dialect/D2M/Transforms/GenericTileComputeLoops.cpp
```diff
@@ -109,49 +109,144 @@ static SmallVector<int64_t> calculateOptimalSubblockSizes(
...
     if (optimizeSubblocking) {
-      Type largestDstType = utils::getRegionLargestDstElemType(op.getRegion());
-      const unsigned dstCapacity =
-          ttcore::getOpChipDescAttr(op).getDstLogicalSizeTiles(
-              largestDstType, false, maxDstPhysicalSizeTiles);
-
-      subblockSizes = calculateOptimalSubblockSizes(
-          op.getIndexingMapsArray(), op.getInputs(), outputTensor.getShape(),
-          dstCapacity);
+      llvm::errs()
```
vwellsTT: remove debug code

lib/Dialect/D2M/Transforms/GenericTileComputeLoops.cpp
```diff
@@ -181,11 +280,18 @@ class D2MGenericTileComputeLoops
       D2MGenericTileComputeLoops>::D2MGenericTileComputeLoopsBase;
 
   void runOnOperation() final {
+
+    // Run the analysis once before any pattern rewriting
```
vwellsTT: nit: .

lib/Dialect/D2M/Transforms/GenericTileComputeLoops.cpp
```diff
@@ -181,11 +280,18 @@ class D2MGenericTileComputeLoops
...
+
     MLIRContext *ctx = &getContext();
     RewritePatternSet patterns(ctx);
-    patterns.add<D2MGenericComputeRewriter>(ctx,
-                                            maxDstPhysicalSizeTiles.getValue());
+    patterns.add<D2MGenericComputeRewriter>(
+        ctx, maxDstPhysicalSizeTiles.getValue(), analysis.dstRegisterInfoList);
     walkAndApplyPatterns(getOperation(), std::move(patterns));
+
+    // Mark the analysis as preserved for use by downstream passes
```
vwellsTT: nit: .

lib/Dialect/D2M/Transforms/InsertDstRegisterAccess.cpp
```diff
@@ -98,11 +92,23 @@ struct D2MInsertDstRegisterAccessRewriter final
           linalgToAffineFailed = true;
           return;
         }
+
+        // Extract maxDstUsage and dstSliceIndices from analysis before erasing
```
vwellsTT: nit: .

lib/Dialect/D2M/Transforms/InsertDstRegisterAccess.cpp
```diff
@@ -297,17 +292,14 @@ struct D2MInsertDstRegisterAccessRewriter final
...
-                  : dstSliceAllocationState.allocate();
-
-          dstRegisterAllocation[computeOp] = {allocatedIndex,
+
+          dstRegisterAllocation[computeOp] = {dstSliceIndex,
                                               outermostInnerComputeLoop};
         }
       }
+
+      // Move to next DST slice index for next compute op
```
vwellsTT: nit: .

lib/Dialect/D2M/Transforms/InsertDstRegisterAccess.cpp
```diff
@@ -386,9 +380,20 @@ struct D2MInsertDstRegisterAccessRewriter final
     if (failed(linalgLoops)) {
       return false;
     }
+
+    // Extract maxDstUsage and dstSliceIndices from analysis before erasing
```
vwellsTT: nit: .

lib/Dialect/D2M/Transforms/InsertDstRegisterAccess.cpp
```diff
@@ -777,8 +784,16 @@ class D2MInsertDstRegisterAccess
     MLIRContext *ctx = &getContext();
     RewritePatternSet patterns(ctx);
 
+    // Retrieve the cached analysis from the pass manager instead of recomputing
```
vwellsTT: nit: .

lib/Dialect/D2M/Transforms/GenericTileComputeLoops.cpp
dlokeTT: Yeah, I'm about to remove this

lib/Dialect/D2M/Transforms/GenericTileComputeLoops.cpp
wenbinlyuTT: If these numbers have units attached, we would get: `dstCapacity tiles / maxDstUsage slices = X tiles/slice`.
But from the comment above it seems that `maxDstUsage` also has the unit of tiles? I thought its the number of slices with unknown/unspecified tiles per slice.
I guess I am just really confused about the DST management logic that spans four passes...

lib/Dialect/D2M/Transforms/GenericTileComputeLoops.cpp
dlokeTT: Yeah, this was one of the weird things I did to have this work. I wasn't sure on matchAndRewrite either but I haven't encountered any issues doing this, and I've run this numerous times in testing. Based on this https://llvm.org/devmtg/2024-10/slides/techtalk/Springer-Pattern-Based-IR-Rewriting-in-MLIR.pdf though it looks like it's not guaranteed, although this might be stale. Yeah, the initial approach was to use a DenseMap but hashing on the operations doesn't work because of the way the rewrite changes op's locations in memory. We could hash on an attribute in each generic, although I wasn't sure if we wanted to actually add anything to the IR so I went with this approach. I do this in 2 places, with the generic ops itself and storing intermediate results from compute ops so i felt like just adding counter's as attributes in the ir was bad. Happy to go down that route if we think that's the better approach though. Currently we assume that the generics are traversed in post order and the computes in pre order but not having this contract explicit and it potentially not being guaranteed are big risks

lib/Dialect/D2M/Transforms/GenericTileComputeLoops.cpp
dlokeTT: Yeah maxDstUsage is basically maximum number of slices of dest used for 1 input tile. That unit math gets kind of tricky with the way the sub blocking factor is consumed but to your original question yes those slices are all used for 1 input tile of data

dlokeTT: And this doesn't affect insertDst access since it already has the loops. As an example: say we have 8 tiles of dest. We are doing an add (any binary op will work). the analysis will tell us we use 3 slices of dest at max at any one point. It will say use slice 0 and slice 1 for the two inputs and slice 2 for the output. Then generic tile will see that we use 3 slices max, we have 8 tiles so it will subblock to factors of [2,1].  insertDst has support for indexing multi tile using the loop nest info to update the real slice to get from in dst. In the example IR below you can see how dest gets sub blocked with the same factors so that the indexing is done correctly.

```
%dst = d2m.acquire_dst() : memref<4x2x1x!ttcore.tile<32x32, bf16>, #ttcore.memory_space<dst>>
affine.for %arg4 = 0 to 2 {
  affine.for %arg5 = 0 to 1 {
    %3 = affine.load %subview[%arg4, %arg5] : memref<2x1x!ttcore.tile<32x32, bf16>, strided<[4, 1], offset: ?>, #ttcore.memory_space<l1>>
    affine.store %3, %dst[0, %arg4, %arg5] : memref<4x2x1x!ttcore.tile<32x32, bf16>, #ttcore.memory_space<dst>>
    %4 = affine.load %subview_6[%arg4, %arg5] : memref<2x1x!ttcore.tile<32x32, bf16>, strided<[4, 1], offset: ?>, #ttcore.memory_space<l1>>
    affine.store %4, %dst[1, %arg4, %arg5] : memref<4x2x1x!ttcore.tile<32x32, bf16>, #ttcore.memory_space<dst>>
  }
}
affine.for %arg4 = 0 to 2 {
  affine.for %arg5 = 0 to 1 {
    %3 = affine.load %dst[0, %arg4, %arg5] : memref<4x2x1x!ttcore.tile<32x32, bf16>, #ttcore.memory_space<dst>>
    %4 = affine.load %dst[1, %arg4, %arg5] : memref<4x2x1x!ttcore.tile<32x32, bf16>, #ttcore.memory_space<dst>>
    %5 = "d2m.tile_add"(%3, %4) : (!ttcore.tile<32x32, bf16>, !ttcore.tile<32x32, bf16>) -> !ttcore.tile<32x32, bf16>
    affine.store %5, %dst[2, %arg4, %arg5] : memref<4x2x1x!ttcore.tile<32x32, bf16>, #ttcore.memory_space<dst>>
  }
}
affine.for %arg4 = 0 to 2 {
  affine.for %arg5 = 0 to 1 {
    %3 = affine.load %dst[2, %arg4, %arg5] : memref<4x2x1x!ttcore.tile<32x32, bf16>, #ttcore.memory_space<dst>>
    affine.store %3, %subview_7[%arg4, %arg5] : memref<2x1x!ttcore.tile<32x32, bf16>, strided<[4, 1], offset: ?>, #ttcore.memory_space<l1>>
  }
}
```

lib/Dialect/D2M/Transforms/GenericTileComputeLoops.cpp
```diff
@@ -109,47 +109,88 @@ static SmallVector<int64_t> calculateOptimalSubblockSizes(
...
+// not divisible, we can subblock by the actual shape.
+// Todo @dloke figure out how to get by this, do we set dst shape in
+// analysis pass or does it have to be done in runtime, or do we pad.
+static SmallVector<int64_t>
+calculateOutputBlockShape(ArrayRef<int64_t> outputShape,
+                          int64_t subblockFactor) {
+  SmallVector<int64_t> outputBlockShape;
+
+  // Only support subblocking for 2D output shapes.
+  if (outputShape.size() == 2 && (outputShape[0] % subblockFactor == 0 ||
```
wenbinlyuTT: Switch to early return: if output's rank isn't 2, return {}; otherwise return {dim0, dim1}.
Then split the case for evenly divisible outputShape[0] and the case for outputShape[0] < subblockFactor.
This makes the logic clearer and can eliminate some `std::min` (surely subblockFactor is smaller if it divides outputShape[0]?).

dlokeTT: This doesn't even work 😢. Have to rework this

wenbinlyuTT: OK, but I think the logic in this function is pretty sound now, if you do the rewrite it's basically
```cpp
  if (outputShape.size() != 2) {
    return {};
  }

  if ((outputShape[0] % subblockFactor == 0)) {
    return {subblockFactor, 1};
  }

  if ((outputShape[0] < subblockFactor)) {
    int64_t dim0 = outputShape[0];
    int64_t dim1 = std::min(subblockFactor / outputShape[0], outputShape[1]);
    return {dim0, dim1};
  }

  return {};
```
So simple it's unlikely to be wrong.

lib/Dialect/D2M/Transforms/GenericTileComputeLoops.cpp
vwellsTT: Yeah, I think this may be technically safe as written because it seems there are some ordering constraints (order is fixed if nothing is rewritten, etc), but I think it's pretty brittle, and reliant on the behavior of this pass adhering to implicit guarantees, etc.  I think this should be rewritten in a more robust form.

> Yeah, the initial approach was to use a DenseMap but hashing on the operations doesn't work because of the way the rewrite changes op's locations in memory.

I would think this is still a workable solution, it'd just require some manual intervention to update keys when a given generic op is rewritten?

> We could hash on an attribute in each generic, although I wasn't sure if we wanted to actually add anything to the IR so I went with this approach

Yeah, the overhead of this seems nontrivial, probably best to avoid.

Taking a step back, what is the source of this list? Are there 2 separate passes running at different stages in the pipeline sharing the data, or what?  If so, I'm not sure I see how this is safe, since intervening passes could remove/add generics, etc.  If not, I'm not sure I see why this is needed.

dlokeTT: > Taking a step back, what is the source of this list? Are there 2 separate passes running at different stages in the pipeline sharing the data, or what? If so, I'm not sure I see how this is safe, since intervening passes could remove/add generics, etc. If not, I'm not sure I see why this is needed.

Yeah. the source of this list is an analysis pass that looks at each linalg generic and determines the max number of slices of dest used at any time and the slice indices that operands and results are stored in. The reason for this is so we do one computation beforehand and then two passes, `genericTileComputeLoop.cpp` and 'insertDstRegisterAccess.cpp' can use their results. The passes are run back to back ensuring that no generics are added/removed.

> I would think this is still a workable solution, it'd just require some manual intervention to update keys when a given generic op is rewritten?

Hmm yeah I guess I could go down this route. Although I would have to update every key every time a `matchAndRewrite` occurs because even though it just modifies it's linalg generic, the whole IR get's changed and all key's get invalidated. I thought this would lead to too much IR parsing so stuck with what I had. I tried looking for a more MLIR native way to solve this problem but I couldn't find one

vwellsTT: Gotcha.  Yeah, okay that makes more sense, but I still think this pattern feels a bit icky to me, though the alternatives don't feel great either.  I think personally I'd vote for some kind of temporary attr-tagging based approach instead; it's a bit ugly for sure but I think that'd be the MLIR-esque thing to do to carry this kind of per-op data between passes.

vwellsTT: Curious to hear what others think though, maybe I'm being too picky

vwellsTT: I also wonder if you key on `location` or not--unsure exactly how that works, but I know there is such a concept and it seems like it ought to uniquely specify each generic

dlokeTT: Yeah, I totally agree on the way I've done it so far being icky. Let me look into the location stuff and yea maybe others can chime in

Issue #5348, opened 2025-10-14
d2m ttir.sign


brnorris03:
### Is your feature request related to a problem? Please describe

Implement `ttir.sign` support for d2m

### Describe the solution you'd like

* Define `TTKernel_SignTileOpInit` and `TTKernel_SignTileOp`. 
* Define `D2M_TileSignOp`
* In TTIRToD2M, add a conversion that first casts to f16 regardless of input type based on a new condition `D2MNamedElementwiseRewriter` similar to https://github.com/tenstorrent/tt-mlir/blob/9f0e66b2169fc833c25a30e73cdfb0d8f462f2c6/lib/Conversion/TTIRToD2M/TTIRToD2M.cpp#L472, that generates the new `D2M_TileSignOp`
* Add conversion to `D2MToTTKernel`.

brnorris03:
Implemented. `TTKernel_SignTileOp`/`SignTileInitOp` (`TTKernelOps.td`), `D2M_TileSignOp` (`D2MGenericRegionOps.td`), the TTIR->D2M rewriter `D2MNamedElementwiseRewriter<ttir::SignOp, d2m::TileSignOp>` (`TTIRToD2M.cpp`), and the D2M->TTKernel conversion all exist, with lit coverage in `test/ttmlir/Conversion/TTIRToD2M/named_to_generic.mlir` (`ttir.sign` -> `d2m.tile_sign`). Closing as done.

PR #5379, opened 2025-10-16
[d2m] Add `ttir.sign` op (#5348)

zoecarver:
### Ticket
[Link to Github Issue](https://github.com/tenstorrent/tt-mlir/issues/5348)

### Problem description
Implement `ttir.sign` support for d2m.

### What's changed
* Add `D2M_TileSignOp`, `TTKernel_SignTileOp`, and `TTKernel_SignTileInitOp`. 
* Add `ttir::SignOp` -> `d2m::TileSignOp` conversion; `d2m::TileSignOp` -> `ttkernel::SignTileOp` conversion; `ttkernel::SignTileInitOp` and `ttkernel::SignTileOp` emit to c rewriter support.
* Add typecast where needed (llk sign kernel only supports f16, so other floats need to be casted). 
* Verification for d2m sign op types (only accept f16 inputs).
* Fixes for typecast lowering on top of #5222 (please ignore for this change).
* Tests for verifier, lit tests for `ttir::SignOp` -> `d2m::TileSignOp` and `d2m::TileSignOp` -> `ttkernel::SignTileOp`, lit test for full pipeline (ttir -> d2m -> ttkernel), runtime pytests for sign (in `ttnn-jit/test_eltwise.py`).
* Bufferization fix (don't crash if there is no HostLayoutAttr). Please let me know if this is correct. 

This PR is "based on" #5222 and requires it to land for the casting logic. Please only review the sign related changes here and focus on the typecast fixes in the other PR. 

Open question: `ttnn-jit/test_eltwise.py` highlights the case where casting rounds to zero and loses precision. This combined with the potential overhead of casting brings into question if this is the correct approach at all. It may be better to lower sign to pure math ops and skip the llk kernel altogether. 

### Checklist
- [X] Tested on wormhole HW


test/ttnn-jit/test_eltwise.py
```diff
@@ -81,6 +81,21 @@ def run_op_test(
         ]
```
wenbinlyuTT, line 102: I think we can also enable the test for this in the normal builder test suite: https://github.com/tenstorrent/tt-mlir/blob/e8d05138b74a0c03b1ef4d5ae0d71b76e0a3ba8a/test/python/golden/test_ttir_ops.py#L2688

test/ttmlir/Conversion/D2MToTTKernel/sign_op.mlir
```diff
@@ -0,0 +1,22 @@
+// RUN: ttmlir-opt --ttcore-register-device --d2m-insert-dst-register-access --lower-affine --d2m-generic-linearize-memref --lower-affine --convert-d2m-to-ttkernel --canonicalize -o %t.mlir %s
```
wenbinlyuTT: It's probably better to put this test inside `single_tile_ops.mlir`, with all the other simple eltwise ops.

test/ttmlir/Silicon/TTMetal/n150/sign_op_full_pipeline.mlir
```diff
@@ -0,0 +1,33 @@
+// RUN: ttmlir-opt --ttcore-register-device --ttir-to-d2m --ttir-bufferization-pipeline --d2m-insert-dst-register-access --lower-affine --d2m-generic-linearize-memref --lower-affine --convert-d2m-to-ttkernel --canonicalize -o %t %s
```
wenbinlyuTT, line 1: Ditto, and we just need one test for each of f16 & f32.

test/ttmlir/Conversion/TTIRToD2M/sign_op.mlir
```diff
@@ -0,0 +1,43 @@
+// RUN: ttmlir-opt --ttcore-register-device --ttir-to-d2m -o %t %s
```
wenbinlyuTT: It's better to put this in `named_to_generic.mlir` I think.

zoecarver:
Thank you @wenbinlyuTT! Will do :)

test/ttnn-jit/test_eltwise.py
```diff
@@ -81,6 +81,21 @@ def run_op_test(
...
         inputs = [create_dram_tensor(device, h, w, dtype) for _ in range(num_inputs)]
+
+    # FIXME: The sign kernel performs f32->f16 typecast which causes small f32 values to underflow to zero.
+    # The sign kernel returns -1 for negative, 0 for zero, and 1 for positive. So small numbers will be
+    # rounded to zero during the typecast. For example: 2.4840892365e-05 produces the wrong answer
+    # (should be 1 but produces 0 after rounding to f16).
+    # The hardware may flush subnormals to zero, so we skip any values below f16's minimum normal value.
+    if op.__name__ == "sign" and dtype == torch.float32:
+        input_torch = inputs[0].cpu().to_torch()
+        min_f16_normal = 6.104e-5  # Minimum normal float16 value
```
brnorris03: an alternative to hard-coding is to use numpy, e.g., `np.finfo(np.float16).eps`

lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
```diff
@@ -281,14 +281,34 @@ class MemrefStoreRewriter : public OpConversionPattern<memref::StoreOp> {
...
-    auto load = mlir::dyn_cast<memref::LoadOp>(op.getValue().getDefiningOp());
+    // Look for the load operation, potentially through an unrealized conversion
+    // cast
+    Operation *definingOp = op.getValue().getDefiningOp();
+    auto load = mlir::dyn_cast<memref::LoadOp>(definingOp);
+
+    // If not a direct load, check if it's an unrealized cast wrapping a load
+    if (!load && definingOp) {
+      if (auto unrealizedCast =
+              mlir::dyn_cast<UnrealizedConversionCastOp>(definingOp)) {
```
brnorris03: What if there is no defining op (or reachable load), e.g., as for func args? Should that be an error?

lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
```diff
@@ -404,6 +404,8 @@ class D2MNamedElementwiseRewriter final
       std::is_same_v<TileOp, d2m::TileLtzOp> ||
       std::is_same_v<TileOp, d2m::TileLezOp>;
 
+  static constexpr bool isSignOp = std::is_same_v<TileOp, d2m::TileSignOp>;
```
brnorris03: If following LLVM naming conventions, this would be called something like `kIsSignOp`

lib/Dialect/D2M/IR/D2MGenericRegionOps.cpp
```diff
@@ -576,6 +576,19 @@ mlir::LogicalResult TileUntilizeBlockOp::verify() {
   return success();
 }
 
+// TileSignOp verification
+::mlir::LogicalResult TileSignOp::verify() {
+  auto tileType = mlir::cast<mlir::tt::ttcore::TileType>(getInput().getType());
```
brnorris03, line 582: what if the type is not a valid TileType? (unless of course that is ensured elsewhere, e.g., in the TileType definition, but otherwise I think here I'd use `dyn_cast`)

lib/Dialect/D2M/Transforms/InsertDstRegisterAccess.cpp
```diff
@@ -464,14 +467,37 @@ struct D2MInsertDstRegisterAccessRewriter final
...
-                loc, dstLoad.getResult(), cb, l1AccessMap, l1AccessIndices);
+                loc, valueToStore, cb, l1AccessMap, l1AccessIndices);
           },
           // Replacement of the original store with one from dst.
           [&](PatternRewriter &rewriter, affine::AffineStoreOp op,
               AffineMap dstAccessMap, ValueRange dstAccessIndices) {
+            Value valueToStore = op.getValue();
+            // Insert unrealized conversion cast if value type differs from dst
+            // type
+            auto dstType = mlir::cast<MemRefType>(dst.getType());
```
brnorris03: is `MemRefType` really required or is `ShapedType` enough here?

lib/Dialect/D2M/Transforms/InsertDstRegisterAccess.cpp
```diff
@@ -464,14 +467,37 @@ struct D2MInsertDstRegisterAccessRewriter final
               AffineMap dstAccessMap, ValueRange dstAccessIndices) {
             auto dstLoad = rewriter.create<affine::AffineLoadOp>(
                 loc, dst, dstAccessMap, dstAccessIndices);
+            Value valueToStore = dstLoad.getResult();
+
+            // Insert unrealized conversion cast if destination CB type differs
+            // from dst type
+            auto cbType = mlir::cast<MemRefType>(cb.getType());
```
brnorris03, line 446: I could be wrong, but I think this could also potentially be a tensor, so `dyn_cast` if so? Or require memref in the op verifier? Same comment about `dstType` below.

lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
zoecarver: Yeah I was actually going to leave this same comment in the PR (#5222). It seems like there should be a more formal way to handle this pattern, especially if it's required for lowering. Maybe we could introduce one op that can encapsulate both if they must appear together like this. Worst case, maybe there's a tool that can cover all MLIR ops in a switch or something to trace through arbitrary instructions (seems like it might be generally useful).

lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
zoecarver: This matches `isComparisonOp` above.

lib/Dialect/D2M/Transforms/InsertDstRegisterAccess.cpp
zoecarver: This is a good question: is it strictly required to run bufferization before insert-dst-register-access? If so, then I think it should always be a memref. 

Regardless also a comment for #5222.

test/ttmlir/Silicon/TTMetal/n150/sign_op_full_pipeline.mlir
zoecarver: I will consolidate the other tests, but this one is testing the full pipeline (ttir -> d2m -> ttkernel), so it needs a different run comment. Is there an existing test that I could move it to (with a similar run comment)? Or want me to omit it entirely?

lib/Dialect/D2M/IR/D2MGenericRegionOps.cpp
zoecarver: The generated verifier actually already covers this. Added a test.

zoecarver:
Review comments addressed! Thank you @brnorris03! Now for the rebase...

test/ttmlir/Silicon/TTMetal/n150/sign_op_full_pipeline.mlir
wenbinlyuTT: It's just that we don't usually do this, probably because this test belongs to the `D2MToTTKernel` directory as much as it belongs to the `TTIRToD2M` directory, so where exactly should we put it?
OTOH, we have tests under `test/ttmlir/Silicon/TTMetal` that is even more complete, they start from ttir and eventually get executed on the silicon (sans the correctness checks). Maybe you can move this test there.

lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
vwellsTT: better to change both to standardized style imo

lib/Dialect/D2M/Transforms/InsertDstRegisterAccess.cpp
vwellsTT: We generally are okay with assuming pipeline with run in order for such things; I think a `cast` is good here (will assert if an unexpected type arises)

test/ttmlir/Silicon/TTMetal/n150/sign_op_full_pipeline.mlir
vwellsTT: Yeah, I agree with Wenbin, I think the `Conversion` dir tests should be strictly for specific, standalone conversion pass unit tests, and e2e pipeline runs belong elsewhere (such as `Silicon/TTMetal`

lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
```diff
@@ -281,14 +281,34 @@ class MemrefStoreRewriter : public OpConversionPattern<memref::StoreOp> {
   LogicalResult
   matchAndRewrite(memref::StoreOp op, memref::StoreOpAdaptor adaptor,
                   ConversionPatternRewriter &rewriter) const final {
-    auto load = mlir::dyn_cast<memref::LoadOp>(op.getValue().getDefiningOp());
+    // Look for the load operation, potentially through an unrealized conversion
+    // cast
+    Operation *definingOp = op.getValue().getDefiningOp();
+    auto load = mlir::dyn_cast<memref::LoadOp>(definingOp);
+
+    // If not a direct load, check if it's an unrealized cast wrapping a load
```
vwellsTT: nit: Please add end punctation to all comments in PR, per our guidlelines https://docs.tenstorrent.com/tt-mlir/coding-guidelines.html#comments (comments should be full sentences, including punctation)

test/ttmlir/Silicon/TTMetal/n150/sign_op_full_pipeline.mlir
zoecarver: Those all seem to be targeted at specific HW. Happy to omit the test if preferred.

zoecarver: Or move it into `Silicon/TTMetal` (no sub directory) or wherever else you think it should go :)

wenbinlyuTT: Moving it to the N150 silicon test suite is good enough for a quick smoke test and code coverage. For more serious tests on different HW we have the builder tests.

lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
```diff
@@ -414,13 +416,31 @@ class D2MNamedElementwiseRewriter final
...
+      if (tileType.getDataType() != ttcore::DataType::Float16) {
+        yield = bbBuilder.create<d2m::TileTypecastOp>(loc, tileType, yield);
+      }
     } else {
       yield = bbBuilder.create<TileOp>(loc, resultTypes, operands);
     }
 
     bbBuilder.create<mlir::linalg::YieldOp>(bbLoc, yield);
   }
-
```
brnorris03, line 423: nit: add blank line back?

test/ttmlir/Conversion/D2MToTTKernel/single_tile_ops.mlir
```diff
@@ -602,6 +603,31 @@ module {
...
+    ^compute0(%cb0: memref<1x1x!ttype_f16, #l1_>, %cb1: memref<1x1x!ttype_f16, #l1_>):
+      linalg.generic {indexing_maps = [#map_, #map_], iterator_types = ["parallel", "parallel"]} ins(%cb0 : memref<1x1x!ttype_f16, #l1_>) outs(%cb1 : memref<1x1x!ttype_f16, #l1_>) {
+      ^bb0(%arg0: !ttype_f16, %arg1: !ttype_f16):
+        // CHECK-NOT: d2m.tile_sign
+        // CHECK: %[[CB_IN:.+]] = ttkernel.cb_reinterpret_shape
+        // CHECK: %[[CB_OUT:.+]] = ttkernel.cb_reinterpret_shape
+        // CHECK: ttkernel.init_sfpu(%[[CB_IN]], %[[CB_OUT]])
+        // CHECK: ttkernel.tile_regs_acquire
+        // CHECK: ttkernel.copy_tile_init(%[[CB_IN]])
+        // CHECK-NEXT: ttkernel.copy_tile(%[[CB_IN]], %{{.+}}, %[[DST_IDX:[^,)]+]])
```
brnorris03: Curious about the `%[[DST_IDX:[^,)]+]]` regular expression -- why is it needed (instead of just `%[[DST_IDX:.+]]`?

test/ttmlir/Conversion/TTIRToD2M/named_to_generic.mlir
```diff
@@ -191,4 +191,39 @@ module {
...
   }
+
+  // CHECK-LABEL: func @test_sign_f32
+  func.func @test_sign_f32(%arg: !ttype, %out: !ttype) -> (!ttype) {
+    // CHECK: linalg.generic
+    // CHECK: ^bb0([[IN:%.*]]: !ttcore.tile<32x32, f32>,
+    // CHECK: [[CAST_IN:%.*]] = "d2m.tile_typecast"([[IN]]) : (!ttcore.tile<32x32, f32>) -> !ttcore.tile<32x32, f16>
+    // CHECK: [[SIGN:%.*]] = "d2m.tile_sign"([[CAST_IN]]) : (!ttcore.tile<32x32, f16>) -> !ttcore.tile<32x32, f16>
+    // CHECK: [[CAST_OUT:%.*]] = "d2m.tile_typecast"([[SIGN]]) : (!ttcore.tile<32x32, f16>) -> !ttcore.tile<32x32, f32>
+    // CHECK: linalg.yield [[CAST_OUT]] : !ttcore.tile<32x32, f32>
```
brnorris03: Nit: Lit var naming -- in some tests, `%[[VAR:.+]]` (the usual for LLVM) is used while in others, `[[VAR:%.*]]`. I think it's better to be consistent and pick one for all tests unless there is some reason that is not a good idea?

test/ttmlir/Conversion/TTIRToD2M/sign_op_verifier.mlir
```diff
@@ -0,0 +1,24 @@
+// RUN: not ttmlir-opt %s 2>&1 | FileCheck %s
+
```
brnorris03: Oh, interesting, I have never done invalid (error) tests that way. How does it compare to using `-verify-diagnostics` and `// expected-error`?

zoecarver: I find it easier to match, but expected error is probably better because it will catch other verifier issues.

test/ttmlir/Conversion/D2MToTTKernel/single_tile_ops.mlir
zoecarver: A confusing ish way to prevent the regex from being to greedy.

zoecarver:
Thanks to all the reviewers! Everything is addressed so hoping to land once CI is green.

test/ttmlir/Silicon/TTMetal/n150/sign_op_full_pipeline.mlir
vwellsTT: yeah, I think our approach is to use n150 for generic testing, and only use other variants for things we specifically need that HW to test (multidevice, etc), FWIW

Issue #5366, opened 2025-10-16
Linear pattern matmul -> reshape -> add not working as intended


mstojkovicTT:
### Describe the bug

When uplifting MLIR to Forge-FE, i got bug on ViT model:
`loc("Add_16"): error: 'ttir.reshape' op Input tensor
    number of elements 768 and output tensor number of elements 1210368 must be the same`

Upon further inspection i got to the code that was problematic:
```
%1601 = "ttir.matmul"(%1599, %arg288, %1600) <{transpose_a = false, transpose_b = false}> : (tensor<1576x768xbf16>, tensor<768x768xbf16>, tensor<1576x768xbf16>) -> tensor<1576x768xbf16>
    %1602 = ttir.empty() : tensor<8x197x768xbf16>
    %1603 = "ttir.reshape"(%1601, %1602) <{shape = [8 : i32, 197 : i32, 768 : i32]}> : (tensor<1576x768xbf16>, tensor<8x197x768xbf16>) -> tensor<8x197x768xbf16>
    %1604 = ttir.empty() : tensor<8x197x768xbf16>
    %1605 = "ttir.add"(%1603, %arg289, %1604) : (tensor<8x197x768xbf16>, tensor<768xbf16>, tensor<8x197x768xbf16>) -> tensor<8x197x768xbf16>
    %1606 = ttir.empty() : tensor<8x197x768xbf16>
```

When bisecting trough MLIR commits, i found the problematic commit: #5149 
The problematic part is the second fusion pattern: 
`matmul -> reshape -> add = reshape -> linear -> reshape`

![Image](https://github.com/user-attachments/assets/099865c0-844d-4dc7-9e81-27bc9bf7efc5)

This bug happens because we are trying to reshape bias tensor to matmul output shape which isnt needed in this case and isnt possible (when bias is implicitly cast to match the other tensor in op, there are no elements for reshape).

### Steps to reproduce the issue

Here is minimal repro you can run on Forge-FE to reproduce the bug: 

1. `git checkout mstojkovic/repro_vit_bug`

2. run minimal repro: `pytest -svv test_matmul_reshape_add_fusion.py`

``` python
import torch
import pytest
from torch import nn
from forge import forge
from forge.verify.verify import verify


class MatmulReshapeAddVIT(nn.Module):
    """
    Reproduces the exact VIT pattern that triggers the reshape issue.

    Pattern from VIT error:
    matmul: A x B -> [1576, 768] (1,210,368 elements)
[...]

PR #5382, opened 2025-10-16
Linear op fusion broadcast + reshape folding

ddilbazTT:
### Ticket
#5366 

### Problem description
I had added a fusion pattern for matmul -> reshape -> add which does not work with 1d bias. 
It originally would take in a graph like the following:
matmul_AB = matmul(A, B)
reshape_AB = reshape(matmul_AB)
output = add (reshape_AB, bias)

and change it to:
reshape_bias = reshape(bias)
linear_output = linear(A, B, reshape_bias)
output = reshape(linear_output) 

However, this would not work in case where:
A: [1576, 2]
B: [2, 768]
bias: [768] 

Therefore, I introduced `FindOneDimensionalBias` function which finds the 1-dimensional bias in a valid matmul -> add or matmul -> reshape -> add sequence.

In matmul -> add: now I use the one dimensional bias as the bias input in linear op if it exists.
In matmul -> reshape -> add: one dimensional bias is required to be able to fuse. 

### What's changed
- Introduced `FindOneDimensionalBias` 
- Removed bias reshape from `matmul -> reshape -> add` 
- Using one dimensional bias in `matmul -> reshape -> add`
- Using one dimensional bias is `matmul -> add` if it exists
- Added test cases 

### Checklist

### DIFF
--- a/lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
+++ b/lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
@@ -1192,27 +1192,33 @@ public:
 // a MatmulOp and the other operand is a bias term. It then replaces the
 // AddOp with a LinearOp that combines the functionality of both operations.
 // The pattern also handles the case where the MatmulOp is followed by a
-// ReshapeOp before the AddOp. In this case, it reshapes the bias term
-// accordingly and creates a LinearOp followed by a ReshapeOp to maintain the
-// original output shape.
+// ReshapeOp before the AddOp. In this case, it tries to find the 1D bias term
+// and creates a LinearOp followed by a ReshapeOp to maintain the original
+// output shape.
 class MatmulWithBiasFusionPattern : public mlir::OpRewritePattern<AddOp> {
 public:
   using mlir::OpRewritePattern<AddOp>::OpRewritePattern;
   mlir::LogicalResult
   matchAndRewrite(AddOp addOp, mlir::PatternRewriter &rewriter) const final {
+    // Case 1: Direct MatmulOp -> AddOp pattern.
     if (MatmulOp matmulOp = getFusableMatmulOp(addOp); matmulOp) {
-      Value bias = addOp.getLhs() == matmulOp.getResult() ? addOp.getRhs()
-                                                          : addOp.getLhs();
+      TypedValue<RankedTensorType> bias = addOp.getLhs() == matmulOp.getResult()
+                                              ? addOp.getRhs()
+                                              : addOp.getLhs();
+      Value oneDimensionalBias = FindOneDimensionalBias(bias);
+      Value linearBias = oneDimensionalBias ? oneDimensionalBias : bias;
       Value matmulOpA = matmulOp.getA();
       Value matmulOpB = matmulOp.getB();
       LinearOp linearOp = ttir::utils::createDPSOp<ttir::LinearOp>(
           rewriter, addOp.getLoc(), addOp.getResult().getType(), matmulOpA,
-          matmulOpB, bias, matmulOp.getTransposeA(), matmulOp.getTransposeB());
+          matmulOpB, linearBias, matmulOp.getTransposeA(),
+          matmulOp.getTransposeB());
 
       rewriter.replaceOp(addOp, linearOp);
 
       return mlir::success();
     }
+    // Case 2: MatmulOp -> ReshapeOp -> AddOp pattern.
     if (MatmulOp matmulOp = getFusableReshapedMatmulOp(addOp); matmulOp) {
 
       ReshapeOp reshapeOp =
@@ -1223,26 +1229,20 @@ public:
       TypedValue<RankedTensorType> bias =
           (addOp.getLhs() == reshapeOp.getResult()) ? addOp.getRhs()
                                                     : addOp.getLhs();
-      RankedTensorType biasType = bias.getType();
+      Value oneDimensionalBias = FindOneDimensionalBias(bias);
+      TT_assertv(oneDimensionalBias,
+                 "MatmulWithBiasFusionPattern FindOneDimensionalBias logic is "
+                 "broken.");
 
       Value matmulOpA = matmulOp.getA();
       Value matmulOpB = matmulOp.getB();
-      llvm::ArrayRef<int64_t> matmulOpShape =
-          matmulOp.getResult().getType().getShape();
       llvm::ArrayRef<int64_t> addOpShape =
           addOp.getResult().getType().getShape();
-      SmallVector<int32_t> matmulShapeI32(matmulOpShape.begin(),
-                                          matmulOpShape.end());
       SmallVector<int32_t> addShapeI32(addOpShape.begin(), addOpShape.end());
-
-      Value reshapedBias = ttir::utils::createDPSOp<ttir::ReshapeOp>(
-          rewriter, addOp.getLoc(), matmulOpShape, biasType.getElementType(),
-          biasType.getEncoding(), bias,
-          rewriter.getI32ArrayAttr(matmulShapeI32));
 
       LinearOp linearOp = ttir::utils::createDPSOp<ttir::LinearOp>(
           rewriter, addOp.getLoc(), matmulOp.getResult().getType(), matmulOpA,
-          matmulOpB, reshapedBias, matmulOp.getTransposeA(),
+          matmulOpB, oneDimensionalBias, matmulOp.getTransposeA(),
           matmulOp.getTransposeB());
 
       RankedTensorType addOpType = addOp.getType();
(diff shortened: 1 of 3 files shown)


lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
```diff
@@ -1309,6 +1318,48 @@ class MatmulWithBiasFusionPattern : public mlir::OpRewritePattern<AddOp> {
...
+      // Check if the input to broadcast is 1D.
+      if (broadcastInputType.getRank() == 1) {
+        return broadcastInput;
+      }
+
+      // Check if broadcast input comes from reshape.
+      // We can absorb the reshape if it is acting like a broadcast.
+      // i.e. 1024 -> (1, 1024) or (1, 1, 1024) etc.
+      // If the shapes are broadcast-compatible and the broadcasted shape
+      // matches the output shape, then this is a broadcast-like reshape.
```
LPanosTT: I wouldn't classify this case as a "broadcast-like reshape" as broadcast isn't typically regarded as an op that changes rank. Rather this is a squeesze/unsqueeze.

LPanosTT (approved):
Other than my nit on the comment, looks good!

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
ddilbazTT: Makes sense, you're right

test/ttmlir/Dialect/TTIR/fusing/matmul_with_bias_positive.mlir
```diff
@@ -43,18 +43,43 @@ module {
...
+    %5 = ttir.empty() : tensor<2x34x1024xf32>
+    %6 = "ttir.reshape"(%4, %5) <{shape = [2 : i32, 34 : i32, 1024 : i32]}> : (tensor<68x1024xf32>, tensor<2x34x1024xf32>) -> tensor<2x34x1024xf32>
+    %7 = ttir.empty() : tensor<2x34x1024xf32>
+    %8 = "ttir.add"(%6, %3, %7) : (tensor<2x34x1024xf32>, tensor<2x34x1024xf32>, tensor<2x34x1024xf32>) -> tensor<2x34x1024xf32>
+    return %8 : tensor<2x34x1024xf32>
+  }
+}
+
+module {
+  func.func @dot_general_with_bias_4(%arg0: tensor<1576x2xf32>, %arg1: tensor<2x768xf32>, %bias: tensor<768xf32>) -> tensor<8x197x768xf32> {
```
mtopalovicTT, line 69: Can you try this test. I think this failed with previous code also and is a bug:
```mlir
module {
  func.func @dot_general_with_bias_4(%arg0: tensor<1x1024xf32>, %arg1: tensor<1024x1024xf32>, %bias: tensor<1024x32x2xf32>) -> tensor<1024x32x2xf32> {
    // CHECK: func.func @dot_general_with_bias_4
    %0 = ttir.empty() : tensor<68x1024xf32>
    %1 = "ttir.dot_general"(%arg0, %arg1) <{batch_dims_lhs = array<i64>, batch_dims_rhs = array<i64>, contract_dims_lhs = array<i64: 1>, contract_dims_rhs = array<i64: 0>}> : (tensor<1x1024xf32>, tensor<1024x1024xf32>) -> tensor<1x1024xf32>
    %2 = ttir.empty() : tensor<1024x1x1xf32>
    %3 = "ttir.reshape"(%1, %2) <{shape = [1024 : i32, 1 : i32, 1 : i32]}> : (tensor<1x1024xf32>, tensor<1024x1x1xf32>) -> tensor<1024x1x1xf32>
    %4 = ttir.empty() : tensor<1024x32x2xf32>
    %5 = "ttir.add"(%3, %bias, %4) : (tensor<1024x1x1xf32>, tensor<1024x32x2xf32>, tensor<1024x32x2xf32>) -> tensor<1024x32x2xf32>
    return %5 : tensor<1024x32x2xf32>
  }
}
```

mtopalovicTT (changes requested):
This is to complicated imo. I think we can leverage ttir-implicit-broadcast-fold pass here which will fold broadcast into add. Then you don't need to special case broadcast, you just take the bias and check if bias is broadcast compatible with matmul result. Check snippet below, lmk what you think:


```cpp

      ReshapeOp reshapeOp =
          mlir::dyn_cast<ReshapeOp>(*matmulOp.getResult().getUsers().begin());
      TT_assertv(reshapeOp,
                 "MatmulWithBiasFusionPattern getFusable logic is broken.");

      TypedValue<RankedTensorType> bias =
          (addOp.getLhs() == reshapeOp.getResult()) ? addOp.getRhs()
                                                    : addOp.getLhs();

      // This can be moved into the validation function.
      // Check if bias and matmulOp result are broadcastable.
      SmallVector<int64_t> linearWithBiasShape;
      if (!OpTrait::util::getBroadcastedShape(matmulOp.getType().getShape(),
          bias.getType().getShape(),
          linearWithBiasShape)) {
        return mlir::failure();
      }

      // This can be moved into the validation function.
      // Check that broadcasted volume of matmul matches addOp result volume.
      // Maybe this check is too defensive, I would need to verify.
      RankedTensorType addOpType = addOp.getType();
      SmallVector<int32_t> addOpShape(
          addOpType.getShape().begin(), addOpType.getShape().end());
      if (std::accumulate(addOpShape.begin(), addOpShape.end(), 1, std::multiplies<int64_t>{}) !=
          std::accumulate(linearWithBiasShape.begin(),
                          linearWithBiasShape.end(), 1, std::multiplies<int64_t>{})) {
        return mlir::failure();
      }

      // Create linear op replace result with broadcasted result.
      LinearOp linearOp = ttir::utils::createDPSOp<ttir::LinearOp>(
          rewriter, addOp.getLoc(),
          linearWithBiasShape, matmulOp.getType().getElementType(), matmulOp.getType().getEncoding(), matmulOp.getA(),
          matmulOp.getB(), bias, matmulOp.getTransposeA(),
          matmulOp.getTransposeB());

      Value finalReshape = ttir::utils::createDPSOp<ttir::ReshapeOp>(
          rewriter, addOp.getLoc(), addOp.getType(), linearOp.getResult(),
          rewriter.getI32ArrayAttr(addOpShape));
      rewriter.replaceOp(addOp, finalReshape);

      return mlir::success();
    }
```

Ps by this:
> I think we can leverage ttir-implicit-broadcast-fold pass here_

I mean add additional ttir-fusing pass after broadcast-fold in ttir-to-ttnn-backend-pipeline, which would remove broadcasts.

ddilbazTT:
@mtopalovicTT You're right that this is too complicated. I should do something similar to what you suggested. However, I think adding an extra broadcast fold is not favourable based on the discussions I am having and the errors I am seeing.

I tried adding broadcast fold after ttir fusion pass, and the following test case failed:
```
// RUN: ttmlir-opt --ttir-to-ttnn-backend-pipeline="system-desc-path=%system_desc_path%" -o %t.mlir %s
// RUN: FileCheck %s --input-file=%t.mlir
// RUN: ttmlir-translate --ttnn-to-flatbuffer -o %t.ttnn %t.mlir
module {
  func.func @main(%arg0: tensor<1x16x32xf32>, %arg1: tensor<1x1x32xf32>) -> tensor<1x16x32xf32> {
    // CHECK-NOT: ttnn.repeat
    // CHECK: %{{[0-9]+}} = "ttnn.add"
    %0 = ttir.empty() : tensor<1x16x32xf32>
    %1 = "ttir.broadcast"(%arg1, %0) <{broadcast_dimensions = array<i64: 1, 16, 1>}> : (tensor<1x1x32xf32>, tensor<1x16x32xf32>) -> tensor<1x16x32xf32>
    %2 = ttir.empty() : tensor<1x16x32xf32>
    %3 = "ttir.add"(%arg0, %1, %2) : (tensor<1x16x32xf32>, tensor<1x16x32xf32>, tensor<1x16x32xf32>) -> tensor<1x16x32xf32>
    return %3 : tensor<1x16x32xf32>
  }
}
```
It ends up returning repeat op:
```
module {
  ttcore.device_module {
    builtin.module attributes {ttcore.system_desc = #system_desc} {
      ttcore.device @default_device = <workerGrid = #ttcore.grid<8x8, (d0, d1) -> (0, d0, d1)>, l1Map = (d0, d1, d2)[s0] -> (0, d0, d1, d2 + s0), dramMap = (d0, d1, d2)[s0, s1, s2, s3, s4, s5, s6] -> (0, 0, (((d0 * s1) * (s2 * (s3 * s6)) + d1 * (s2 * (s3 * s6)) + d2) floordiv s4) mod 12, ((((d0 * s1) * (s2 * (s3 * s6)) + d1 * (s2 * (s3 * s6)) + d2) floordiv s4) floordiv 12) * s4 + ((d0 * s1) * (s2 * (s3 * s6)) + d1 * (s2 * (s3 * s6)) + d2) mod s4 + s5), meshShape = , chipIds = [0]>
      func.func @main(%arg0: tensor<1x16x32xf32, #ttnn_layout>, %arg1: tensor<1x1x32xf32, #ttnn_layout>) -> tensor<1x16x32xf32, #ttnn_layout> {
        %0 = "ttnn.repeat"(%arg1) <{repeat_dims = #ttnn.shape<1x16x1>}> : (tensor<1x1x32xf32, #ttnn_layout>) -> tensor<1x16x32xf32, #ttnn_layout>
        "ttnn.deallocate"(%arg1) <{force = false}> : (tensor<1x1x32xf32, #ttnn_layout>) -> ()
        %1 = "ttnn.add"(%arg0, %0) <{dtype = #ttcore.supportedDataTypes<f32>}> : (tensor<1x16x32xf32, #ttnn_layout>, tensor<1x16x32xf32, #ttnn_layout>) -> tensor<1x16x32xf32, #ttnn_layout>
        "ttnn.deallocate"(%0) <{force = false}> : (tensor<1x16x32xf32, #ttnn_layout>) -> ()
        "ttnn.deallocate"(%arg0) <{force = false}> : (tensor<1x16x32xf32, #ttnn_layout>) -> ()
        return %1 : tensor<1x16x32xf32, #ttnn_layout>
      }
    }
  }
}
```
I also discussed this with @mmanzoorTT who told me we should limit the number of broadcast fold passes we have. 

Regardless, there are some things I could do which are similar to what you suggested:
- make linear op broadcastable -> This would require adding an if statement case to lib/Dialect/TTIR/Transforms/Broadcast.cpp
- move broadcast fold before ttir fusion -> I need to test and see if there are any concerns with this.

Maybe I could try each separately, fixing the tests that fail, and then put PR's up separately to see which method is more favourable and get more opinions.

ddilbazTT:
@mtopalovicTT I put up a PR for it here: https://github.com/tenstorrent/tt-mlir/pull/5525 Moving broadcast fold before ttir fusions did not give me any errors when I tried with `cmake --build build -- check-ttmlir` and I started a discussion thread to see if anybody would have thoughts against moving it. I will close this PR for the time being as it is not the preferred way of going forward for matmul with bias fusion pattern.
Edit: Forgot to thank you for the insightful comment, as I referenced that in the new PR.

mtopalovicTT:
I'm not sure why you quick jumped to open the new PR, we should have fixed stuff on this PR. Let me see what is the issue with what I suggested.

mtopalovicTT:
@ddilbazTT I've send you changes offline via patch. Checkout ddilbaz/broadcast_bias and apply the patch. All tests are passing and I didn't change simple_repeat.mlir test, only small change to multiply_with_bias_positive.mlir test to account for dims changes.

ddilbazTT:
@mtopalovicTT For everyone's reference, there are at least two ways I can go about the changes for matmul with bias fixes:
- add a new ttir fusions pass after broadcast fold
- moving broadcast fold before ttir fusions pass (we would not be adding another pass)

The new PR is doing the second way. The first way is what Milan is recommending changes for.

Issue #5356, opened 2025-10-15
[EmitPy] Add CCL ops support


svuckovicTT:
Initial work in https://github.com/tenstorrent/tt-mlir/compare/main...svuckovic/multihost-emitpy-tests

Add support for CCL ops in EmitPy.

PR #5387, opened 2025-10-17
[EmitPy] Add CCL support

svuckovicTT:
### Ticket
Closes https://github.com/tenstorrent/tt-mlir/issues/5356

### Problem description
No CCL op support in EmitPy path.

### What's changed
Added CCL op support.

- Added CCL dialect conversions from TTNN to EmitPy - thanks @hkwonTT!
- Added `ttnn_supplemental.py` which implements CCL ops in Python
- Monkey-patched `ttnn_supplemental` into `ttnn` module in `utils.py`, so that `main.py` remains clean
- Removed `my_get_device.py` and moved device logic to `utils.py`
- Unlinked `test/ttmlir/EmitPy/utils.py` as the symlink was hardcoded

No tests added (op-level tests being added in ttir-builder atm). Manually tested these implementations on 1x2/1x4/1x8 configs for mnist and alexnet models.

### Checklist

### DIFF
--- a/lib/Conversion/TTNNToEmitPy/TTNNToEmitPy.cpp
+++ b/lib/Conversion/TTNNToEmitPy/TTNNToEmitPy.cpp
@@ -2584,14 +2584,10 @@ public:
     ttnn_to_emitpy::EmitPyTTNNEmitter<mlir::tt::ttnn::AllReduceOp> emitter(
         srcOp, adaptor, rewriter);
 
-    // TODO: complete this
     llvm::SmallVector<mlir::Attribute> args{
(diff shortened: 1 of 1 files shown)


test/ttmlir/EmitPy/ttnn_supplemental.py
svuckovicTT, line 1: Probably remove this file and `utils.py` in the same dir. These files are copies from tt-alchemist templates dir, and aren't used by anything.

Technically, this dir should be used for EmitPy dialect tests, so there's no reason to supply necessary infra to test generated Python code.

jgrimTT (approved):

PR #5390, opened 2025-10-17
Add script ttsysjson for converting system descs to/from json and fbb

nsmithtt:
Usage:

    # Convert to binary my_sysdesc.ttsys
    ./tools/scripts/ttsysjson my_sysdesc.json

    # Convert to json my_sysdesc.json
    ./tools/scripts/ttsysjson my_sysdesc.ttsys

Can be convenient for copy pasting the json dump of the sysdesc from CI and converting it to a binary for local repro.

### DIFF
--- a/tools/scripts/ttsysjson
+++ b/tools/scripts/ttsysjson
@@ -46,10 +46,13 @@ if [[ "$INPUT" == *.json ]]; then
     # JSON to FlatBuffer
     flatc -I "$INCLUDE_DIR" -b --size-prefixed --root-type "$ROOT_TYPE" "$SCHEMA" "$INPUT"
     OUTPUT="$(basename "$INPUT" .json).ttsys"
-else
+elif [[ "$INPUT" == *.ttsys ]]; then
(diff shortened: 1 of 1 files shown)


tools/scripts/ttsysjson
```diff
@@ -0,0 +1,55 @@
...
+    echo "Install from: https://github.com/google/flatbuffers" >&2
+    exit 1
+fi
+
+# Detect conversion mode based on file extension
+if [[ "$INPUT" == *.json ]]; then
+    # JSON to FlatBuffer
+    flatc -I "$INCLUDE_DIR" -b --size-prefixed --root-type "$ROOT_TYPE" "$SCHEMA" "$INPUT"
+    OUTPUT="$(basename "$INPUT" .json).ttsys"
+else
```
ddilbazTT: Instead of else, maybe `elif [[ "$INPUT" == *. ttsys ]]` and an error message in `else` ? I don't think you're checking for valid file extensions before conversion.

PR #5393, opened 2025-10-17
Preliminary python bindings for d2m dialect

nsmithtt:
Prerequisite changes for DSL support.


python/CMakeLists.txt
```diff
@@ -161,6 +164,8 @@ set(TTMLIR_PYTHON_SOURCES
   MLIRPythonSources.Dialects.func
   MLIRPythonSources.Dialects.tensor
   MLIRPythonSources.Dialects.linalg
+  MLIRPythonSources.Dialects.math
+  MLIRPythonSources.Dialects.complex
```
brnorris03, line 168: Just curious -- where is complex used? Or is the goal to include all the builtin dialects?

nsmithtt: For some reason the math dialect includes some stuff from it so it's a transitive dep for math.

PR #5394, opened 2025-10-18
Uplift third_party/tt-metal to 69a5ddb1a1c4ba88c9d0e6f1c60c2887be51c4dc 2025-10-20

vmilosevic:
This PR uplifts the third_party/tt-metal to the 69a5ddb1a1c4ba88c9d0e6f1c60c2887be51c4dc

**Note** Dockerfile.base is modified due to dependency updates

### Checklist
- **Frontend CI passing links**
  - [x] [tt-forge-fe](https://github.com/tenstorrent/tt-forge-fe/actions/workflows/on-pr.yml): https://github.com/tenstorrent/tt-forge-fe/actions/runs/18752075454 (with applied uplift fix)
  - [x] [tt-xla (basic + models)](https://github.com/tenstorrent/tt-xla/actions/workflows/manual-test.yml): https://github.com/tenstorrent/tt-xla/actions/runs/18760073516 (minus ResNet50 PCC regression)
- **Follow-up Actions**
  - [x] **Issues filed** to follow up on incomplete changes (if any):
            -  https://github.com/tenstorrent/tt-xla/issues/1778   
  - [x] **Frontend fix PRs** ready (if needed by this uplift):
             - https://github.com/tenstorrent/tt-forge-fe/pull/3001

### DIFF
--- a/include/ttmlir/Dialect/TTNN/Transforms/Workarounds/Decomposition/Conv2dEnableKernelStrideFoldingRewritePattern.h
+++ b/include/ttmlir/Dialect/TTNN/Transforms/Workarounds/Decomposition/Conv2dEnableKernelStrideFoldingRewritePattern.h
@@ -26,7 +26,8 @@ public:
     auto conv2dConfig = srcOp.getConv2dConfig();
 
     if (conv2dConfig && *conv2dConfig &&
-        !conv2dConfig->getEnableKernelStrideFolding()) {
+        conv2dConfig->getEnableKernelStrideFolding() &&
+        !conv2dConfig->getEnableKernelStrideFolding().getValue()) {
       // Conv2dConfig already has disabled kernel stride folding, no need to
       // apply the workaround.
       return failure();


arminaleTT (approved):
Approved for op model

sgholamiTT:
I'm now handing off the uplift PR to @azecevicTT.

-----
**Note about the [CI failure](https://github.com/tenstorrent/tt-mlir/actions/runs/18687695874): ⚠️**
There are some lit tests failing in our CI ([here](https://github.com/tenstorrent/tt-mlir/actions/runs/18687695874/job/53302995816?pr=5394#step:14:1424)).
I believe they're failing due to [this PR](https://github.com/tenstorrent/tt-metal/pull/29993) of @sankarmanoj-tt (specifically, [this change](https://github.com/tenstorrent/tt-metal/commit/14cb8d40e2b953a9eb2dd0a0287d834248238e56#diff-ee7dd99fe5ee66f5457b1526299df1a2ce973a97db9e1383d5d807b7130e71c2)).
The reason is that the conditions added are triggered for the failing tests (eg. of the [IR for the failing test](https://github.com/tenstorrent/tt-mlir/blob/main/test/ttmlir/Silicon/TTNN/n150/conv/large_conv2d_1.mlir)). So they reach the `validate_fold`  function and fail [this](https://github.com/tenstorrent/tt-metal/blob/574fcb54942b52204616ff114ba2910dffb91183/ttnn/cpp/ttnn/operations/data_movement/fold/device/fold_device_op.cpp#L48) assertion.

I believe there might be a huge gap between how `metal` assumes `tt-mlir`  is calling the APIs and how we're actually calling it. For example, should `metal` assume correct shapes of the `conv2d` inputs ? Because so far, we've been flattening the inputs and permute the dimensions and then pass the modified tensor to metal (we first permute `(N, C, H, W)` to `(N, H, W, C)` and then flatten to `(1, 1, NHW, C)`). This is how we always have been doing `conv2d` as far as I'm concerned (@LPanosTT can comment as he has better understanding about it). But it seems like `metal` is relying on the actual input shapes `(N, C, H, W)` to be received (which is unrealistic).

With [that change](https://github.com/tenstorrent/tt-metal/pull/29993), the folding logic gets triggered in some lit tests and then they fail.
To confirm what would happen if we bypass  [this change](https://github.com/tenstorrent/tt-metal/commit/14cb8d40e2b953a9eb2dd0a0287d834248238e56#diff-ee7dd99fe5ee66f5457b1526299df1a2ce973a97db9e1383d5d807b7130e71c2): I changed the paddings of the `conv2d` ops in the failing tests as follows and then the tests succeeded:
```
- padding = array<i64: 0, 0, 0, 0>
+ padding = array<i64: 1, 1, 1, 1>
```
So if we don't enable kernel folding, there's no issue. But I don't think it's sustainable to just bypass this failure with this change (as the current test might be how the costumer of the compiler will be using it). I leave this decision to @azecevicTT though!


**Why other similar tests (Like [this](https://github.com/tenstorrent/tt-mlir/blob/main/test/ttmlir/Silicon/TTNN/n150/conv/large_conv2d_3.mlir) and [this](https://github.com/tenstorrent/tt-mlir/blob/main/test/ttmlir/Silicon/TTNN/n150/conv/large_conv2d_4.mlir)) don't fail?** ❓
They either have `1,1,1,1` padding or stride and kernel spatial dims are not equal. So in short, we just got lucky that those tests also don't fail since the code that does enable auto folding does not apply for those cases. (see the conditions [here](https://github.com/tenstorrent/tt-metal/blob/62f414cd5cfab94bde5844cc2f6723dd2551f418/ttnn/cpp/ttnn/operations/conv/conv2d/conv2d_utils.cpp#L1558-L1578)).
I am mostly sure the other [failing test](https://github.com/tenstorrent/tt-mlir/actions/runs/18687695874/job/53302995816#step:14:4485) should have the same root cause too as the log shows many instances of the assertion failure.

**Local reproduction:** 🔄
```bash
ttmlir-opt --ttir-to-ttnn-backend-pipeline="system-desc-path=$SYSTEM_DESC_PATH" -o ./test.mlir test/ttmlir/Silicon/TTNN/n150/conv/large_conv2d_1.mlir
ttmlir-translate --ttnn-to-flatbuffer -o test.ttnn test.mlir
ttrt run test.ttnn
```

tapspatel (approved):

lib/Conversion/TTNNToEmitPy/TTNNToEmitPy.cpp
```diff
@@ -1338,6 +1339,16 @@ class Conv2dOpConversionPattern
     ttnn_to_emitpy::EmitPyTTNNEmitter<mlir::tt::ttnn::Conv2dOp> emitter(
         conv2dOp, adaptor, rewriter);
 
+    // TODO (azecevic): Has to be set explicitly to false, otherwise it will
+    // assert for flattened Conv2dOp.
+    // https://github.com/tenstorrent/tt-metal/issues/30985
+    auto conv2dConfig = conv2dOp.getConv2dConfig();
```
sdjordjevicTT: Can we move this to TTNN workarounds? It will be easier to track the referenced work item and its removal, once it is fixed?

azecevic000: Yeah, that's a great idea.

Issue #5307, opened 2025-10-10
error: failed to legalize operation `ttir.convolution`


mmilosevicTT:
### Describe the bug

Models are failing in training mode with JAX implementation in TT-XLA. The error message is as in the title:

```
error: failed to legalize operation `ttir.convolution`
```

The full list is:
- SqueezeBERT

### Steps to reproduce the issue

1. Go to TT-XLA repo
2. Go to the specific model folder and remove @pytest.mark.xfail from the training test
3. Run: `pytest path/to/test`

sdjordjevicTT:
Hey @mmilosevicTT can you please attach the generated stablehlo graph so we can start investigating from there.

mmilosevicTT:
Providing full output of the test, as there are several SHLO graphs present:
[sqeezebert_output.txt](https://github.com/user-attachments/files/22905860/sqeezebert_output.txt)

PR #5401, opened 2025-10-20
TTIR convoution with batch count greater than 1.

djstefanovicTT:
### Ticket
https://github.com/tenstorrent/tt-mlir/issues/5307

### Problem description
`TTIR.convolution` is not able to be translated if batch group count is greater than one.

### What's changed
Added support for this case. Inputs are sliced along batch dimension and multiple convolutions with batch group count of exactly one are created. This was done analogous to [stableHLO description](https://openxla.org/stablehlo/spec#convolution).

### Checklist

### DIFF
--- a/include/ttmlir/Conversion/TTNNToEmitC/EmitCConversion.h
+++ b/include/ttmlir/Conversion/TTNNToEmitC/EmitCConversion.h
@@ -1427,6 +1427,12 @@ struct EmitCTypeConverter<::ttnn::operations::conv::conv2d::Conv2dConfig> {
     if (attr.getInPlace()) {
       rso << (firstElement ? "" : ", ") << ".in_place = "
           << EmitCTypeConverter<bool>::convert(attr.getInPlace());
+      firstElement = false;
+    }
+    if (attr.getEnableKernelStrideFolding()) {
+      rso << (firstElement ? "" : ", ") << ".enable_kernel_stride_folding = "
+          << EmitCTypeConverter<bool>::convert(
+                 attr.getEnableKernelStrideFolding());
     }
     rso << "}";
     return buf;
--- a/include/ttmlir/Conversion/TTNNToEmitPy/EmitPyConversion.h
+++ b/include/ttmlir/Conversion/TTNNToEmitPy/EmitPyConversion.h
@@ -1397,6 +1397,29 @@ struct EmitPyTypeConverter<::ttnn::operations::conv::conv2d::Conv2dConfig> {
       rso << (firstElement ? "" : ", ") << "output_layout="
           << EmitPyTypeConverter<::ttnn::Layout>::convert(
                  *attr.getOutputLayout());
+      firstElement = false;
+    }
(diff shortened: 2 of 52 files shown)


lib/Conversion/TTIRToTTIRDecomposition/TTIRToTTIRDecomposition.cpp
```diff
@@ -139,11 +139,6 @@ struct ConvolutionDecompositionPattern
       return false;
     }
 
-    // Not currently support batch groups
```
sdjordjevicTT, line 143: Can we add here a check that if batch group count is greater than 1, the feature group count must be one?

lib/Conversion/TTIRToTTIRDecomposition/TTIRToTTIRDecomposition.cpp
```diff
@@ -139,11 +139,6 @@ struct ConvolutionDecompositionPattern
...
-    if (op.getBatchGroupCount() != 1) {
```
sdjordjevicTT, line 144: It seems that we don't have support for feature group counts in the code. Can we add it? It seems that it would be easy to extend the code to support it.

test/ttmlir/Dialect/TTIR/Decomposition/convolution_decomposition.mlir
```diff
@@ -33,4 +33,64 @@ module @test_convolution {
     // CHECK-SAME: <{permutation = array<i64: 1, 2, 0, 3>}>
```
sdjordjevicTT, line 33: Can you add a builder test with the new parameters so we can verify the test execution on silicon?

djstefanovicTT: Added silicon tests.

lib/Conversion/TTIRToTTIRDecomposition/TTIRToTTIRDecomposition.cpp
djstefanovicTT: Added assert that at least one of them must be 1.

lib/Conversion/TTIRToTTIRDecomposition/TTIRToTTIRDecomposition.cpp
djstefanovicTT: There was already support for feature group counts.

test/ttmlir/Silicon/TTNN/n150/conv/split_conv.mlir
```diff
@@ -0,0 +1,57 @@
+// RUN: ttmlir-opt --ttir-to-ttnn-backend-pipeline="system-desc-path=%system_desc_path%" -o %t.mlir %s
+// RUN: ttmlir-translate --ttnn-to-flatbuffer -o %t.ttnn %t.mlir
+
+  module {
```
sdjordjevicTT: Nit: whitespace before module keyword

Issue #5386, opened 2025-10-17
Fix fusing patterns for conv that has bias


mvasiljevicTT:
Unet model has batch_norm ops that come after convolutions that already have bias. After batch norm is decomposed as mul and add op, mul is incorrectly fused and add is not fused at all, as:

```
conv<weight=w, bias=null>
|                            conv<weight=w, bias=b>
add(b)                       |                              conv<weight=a_bn*w, bias=b>
|                      ->    mul(a_bn)                ->    |
mul(a_bn)                    |                              add(b_bn)
|                            add(b_bn)
add(b_bn)
```

and it should be fused as:
```
conv<weight=w, bias=null>
|                            
add(b)                              
|                  ->     . . .    ->   conv<weight=a_bn*w, bias=a_bn*b + b_bn>
mul(a_bn)                           
|                            
add(b_bn)
```

PR #5402, opened 2025-10-20
Fix ConvWithMultiply and ConvAddBias patterns to handle pre-existing bias

mvasiljevicTT:
### Ticket
[Link to Github Issue](https://github.com/tenstorrent/tt-mlir/issues/5386)

### Problem description
Some models, like unet, have batch norm after convolution that already has bias, which is currently not well supported.

After the batch norm is decomposed into mul and add ops, the mul op is incorrectly fused into the convolution, while the add op is not fused at all, as:
```
conv2d(w, b)
|                       conv2d(a_bn*w, b)
multiply(a_bn)    ->    |
|                       add(b_bn)
add(b_bn)   
```
where it should be fused as:
```
conv2d(w, b)
|                       
multiply(a_bn)    ->   conv2d(a_bn*w, a_bn*b + b_bn)
|                       
add(b_bn)  
```
### What's changed
- Updated `ConvAddBias`:
    - Previously only supported conv without bias: `conv(x, w) + b -> conv(x, w, b)`
    - Now also supports conv with existing bias: `conv(x, w, b1) + b2 -> conv(x, w, b1+b2)`
- Updated `ConvWithMultiply`:
    - Previously only supported conv without bias: `conv(x, w) * a -> conv(x, a*w)`
    - Now also supports conv with existing bias: `conv(x, w, b) * a -> conv(x, a*w, a*b)`

- Updated tests accordingly 

### Checklist

### DIFF
--- a/test/ttmlir/Dialect/TTIR/fusing/conv2d_bias_fusing.mlir
+++ b/test/ttmlir/Dialect/TTIR/fusing/conv2d_bias_fusing.mlir
@@ -85,31 +85,32 @@ module {
   }
 }
 
-// Check that we can only fuse one add into conv2d. Second add is not fused.
+// Check that we can fuse two adds into conv2d bias.
 module {
-  // CHECK-LABEL: func.func @conv2d_single_add
-  func.func @conv2d_single_add(%arg0: tensor<1x32x32x64xbf16>, %arg1: tensor<64x64x3x3xbf16>, %arg2: tensor<1x1x1x64xbf16>) -> tensor<1x30x30x64xbf16> {
+  // CHECK-LABEL: func.func @conv2d_double_add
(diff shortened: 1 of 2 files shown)


lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
```diff
@@ -622,6 +645,13 @@ class ConvWithMultiplyTemplate : public mlir::OpRewritePattern<MultiplyOp> {
     // Update conv to use scaled weights and replace multiply operation.
     rewriter.modifyOpInPlace(
         convOp, [&]() { convOp.getWeightMutable().assign(scaledWeights); });
+
+    if (auto bias = convOp.getBias()) {
```
mtopalovicTT, line 649: Nice catch!

mtopalovicTT (approved):
Nice!

PR #5411, opened 2025-10-20
Migrating golden mappings into new python library: goldens

jgrimTT:
### Ticket
Closes [#5410](https://github.com/tenstorrent/tt-mlir/issues/5410)

### Problem description
Golden mappings used for `builder` are stored in builder and soon need to be used in other tools. 

### What's changed
Created python library `goldens` in new directory `tools/goldens`
Altered tests, builder, CMake to accommodate 
Very minor docs changes

### Checklist

### DIFF
--- a/runtime/include/tt/runtime/types.h
+++ b/runtime/include/tt/runtime/types.h
@@ -194,7 +194,7 @@ struct MeshDeviceOptions {
   std::vector<uint32_t> meshOffset{0, 0};
   std::vector<int> deviceIds{};
   size_t numHWCQs = 1;
-  bool enableProgramCache = false;
+  bool enableProgramCache = true;
   std::optional<std::vector<uint32_t>> meshShape = std::nullopt;
   std::optional<size_t> l1SmallSize = std::nullopt;
   std::optional<size_t> traceRegionSize = std::nullopt;


tools/goldens/utils.py
```diff
@@ -0,0 +1,361 @@
...
+    ArrayAttr,
+    IntegerAttr,
+    IntegerType,
+    BoolAttr,
+    DenseI32ArrayAttr,
+    DenseI64ArrayAttr,
+)
+
+
+class GoldenFunction:
```
tapspatel: I personally think GoldenTensor might be a better name for this, since this class is technically a representation of "tensors" and a function is applied on this tensor

tools/builder/base/builder_golden.py
```diff
@@ -2,22 +2,11 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-"""
```
tapspatel, line 5: I would prefer to keep all these in a single file for now. I think it's okay since most of them are just op mappings, and we can get the runtime working with a single python script before thinking about how to refactor this for maintainability.

tapspatel (changes requested):
Small comments about code structure and naming.

tools/goldens/utils.py
jgrimTT: I agree, and I'll look into a better name, but I don't think GoldenTensor is an option given the fact that builder already uses a type of that name `from ttmlir.passes import GoldenTensor`

tapspatel: Yeah you're right, I forgot we also had that definition in ttmlir. `GoldenMeshTensor`? chatgpt has some other interesting names

dlokeTT (approved):
Changes look good to me!

PR #5416, opened 2025-10-20
[D2M] Insert ttkernel.init_sfpu for typecast

ckaravasilisTT:
### Problem description
Typecast operations produced incorrect output because unpacker/packer were not configured properly.

### What's changed

Added tile typecast to the list of SFPU ops for which `D2MSFPUOpsRewriter` produces `ttkernel.init_sfpu`, configuring unpacker and packer with the correct input/output CB formats.

### DIFF
--- a/test/python/golden/test_ttir_ops.py
+++ b/test/python/golden/test_ttir_ops.py
@@ -1900,16 +1900,18 @@ def test_arange(
 @pytest.mark.parametrize(
     "from_type,to_type",
     [
-        (torch.int32, torch.float32),
-        (torch.float32, torch.int32),
+        pytest.param(
+            torch.int32, torch.float32, marks=pytest.mark.xfail(reason="Golden failure")
+        ),
+        pytest.param(
+            torch.float32, torch.int32, marks=pytest.mark.xfail(reason="Golden failure")
+        ),
         (torch.bfloat16, torch.float32),
         (torch.float32, torch.bfloat16),
     ],
     ids=["i32-f32", "f32-i32", "bf16-f32", "f32-bf16"],
 )
-@pytest.mark.parametrize(
-    "target", ["ttnn", "ttmetal" | Marks(pytest.mark.xfail(reason="Golden failure"))]
-)
+@pytest.mark.parametrize("target", ["ttnn", "ttmetal"])
 def test_typecast(
     shape: Shape,
     from_type: torch.dtype,


wenbinlyuTT:
Could you try to remove the mark at
https://github.com/tenstorrent/tt-mlir/blob/df8527cdff921e1bcccd110e0b563b5b87b1a710/test/python/golden/test_ttir_ops.py#L1911
and see if it fixes the builder tests as well?

lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
```diff
@@ -789,6 +789,27 @@ class D2MTypecastRewriter : public OpTraitConversionPattern<
   matchAndRewrite(Operation *op, ArrayRef<Value> operands,
```
nsmithtt: Why doesn't this PatternRewriter target the `d2m::TileTypecastOp` directly? i.e.
```c++
class D2MTypecastRewriter : public OpTraitConversionPattern<d2m::TileTypecastOp> {
...
  LogicalResult
  matchAndRewrite(d2m::TileTypecastOp op, d2m::TileTypecastOp::Adaptor adaptor,
```

Then we don't need to check in the body if it is a typecast op.

lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
```diff
@@ -789,6 +789,27 @@ class D2MTypecastRewriter : public OpTraitConversionPattern<
...
+          auto insertPt = rewriter.saveInsertionPoint();
+
+          setInsertionPointAfterOperands(rewriter, {inCB, outCB},
+                                         /*allowHoisting*/ true);
+          rewriter.create<ttkernel::ComputeKernelHWStartupOp>(
+              op->getLoc(), inCB, nullptr, outCB);
+
+          rewriter.restoreInsertionPoint(insertPt);
+        }
+      }
```
nsmithtt: Perhaps we should just add the tile typecast op to the list of SFPU ops which already does all of this same machinery? @wenbinlyuTT thoughts?

wenbinlyuTT: Good point! Yeah I think it's safe and better to do it there (it should even work with the f32->f16->f32 workflow in #5379).

lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
ckaravasilisTT: I agree this is better. I was just following `D2MTilizeUntilizeRewriter` because it was also using `ComputeKernelHWStartupOp`. Still learning best practices with MLIR patterns :)

ckaravasilisTT:
> Could you try to remove the mark at
> 
> https://github.com/tenstorrent/tt-mlir/blob/df8527cdff921e1bcccd110e0b563b5b87b1a710/test/python/golden/test_ttir_ops.py#L1911
> 
> 
> and see if it fixes the builder tests as well?

@wenbinlyuTT I think the failed job was because of an OpModel unit test and should be fixed after rebasing.

wenbinlyuTT:
> @wenbinlyuTT I think the failed job was because of an OpModel unit test and should be fixed after rebasing.

We are probably not talking about the same test. OpModel should not affect this D2M test.

The test I linked here is a simple `ttir.typecast` -> flatbuffer test that compares the casted result with pytorch's output.

Currently the ttmetal variant of this test fails on `main`, and so it is marked as `xfail` in the pytest so the CI stays green.

You can run `pytest -v -k typecast test/python/golden/test_ttir_ops.py` to confirm the failure.

And then run `ttrt run ttir-builder-artifacts/` to see exactly how far off the results are (they're pretty far off in `main`).

I was hoping this PR could fix the ttmetal variant of this test so we can remove the mark here.

wenbinlyuTT:
BTW, we also need a test in `test/ttmlir/Conversion/D2MToTTKernel/single_tile_ops.mlir` so that we're sure `ttkernel.compute_kernel_hw_startup` or `ttkernel.init_sfpu` is emitted, without needing to compile & run the test for silicon.

ckaravasilisTT:
> I was hoping this PR could fix the ttmetal variant of this test so we can remove the mark here.

Ah I thought you were referring to the failed CI. I just tested this and only 1 test fails now (f32 to i32 with pcc = 0.036). I'll spend some time trying to debug this.

I will also add a Conversion test to make sure `ttkernel.compute_kernel_hw_startup` is emitted (or `ttkernel.init_sfpu).`

lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
ckaravasilisTT: I added typecast to the list and created a special case for it in `D2MSFPUOpsRewriter`.

wenbinlyuTT (approved):

PR #5418, opened 2025-10-20
[uplift] update files for tt-metal dependencies

gfengTT:
sfpi-version.sh is replaced by sfpi-info.sh and sfpi-version.

Also changed the order of the two sections so that we are getting the true CURRENT_TT_METAL_VERSION before it is updated by the workflow.


brataTT (approved):
Thanks for staying on top of the changes in metal side.

I'm assuming nothing broke to show you that this change was needed. So, uplift owners should manually check what changed whenever the hash in dockerfile gets updated? Which is a good idea.

nsumrakTT (approved):

gfengTT:
@brataTT Nothing broke yet nothing worked. The earlier change I put in first updated the commit hash, then compared it, so it was essentially comparing against itself.

There happened to be another sfpi change that updated the files. I guess I should finish what I started, so I took a second look and realized the issue. We should see the files automatically updated next time the uplift workflow is triggered.

PR #5428, opened 2025-10-21
CI refurbishment, Part 4

nsumrakTT:
### Problem description
Make build more efficient and versatile
Make CI according to DevEx naming convention.

### What's changed
In this CI refurbish edition:
- Fixed some bugs from previous (eg summary is displayed when jobs fail)
- All workflows renamed according to DevEx naming convention.
- test_matrix is now called `settings` and json file is called `tests.json`
- Build has "build components" defined in build.json
- Optional components are introduced which are built only when component files are changed or on uplift/nightly. Defined in `optional-components.yml` and "if" value in `build.json` and `tests.json`
- Nightly run is introduced, builds macOS, normal build with all optional components and wheels (only built in nightly). Failed nightly will send message to #tt-mlir-ci slack channel.
- Build artifacts are compressed with zstd which has 20% better compression, kudos to @wenbinlyuTT 
- Lower build time, introduction of new builder machines (kudos to @vvukomanTT ) and lower sized artifacts improve overall build time and machine availability 
- Documentation updated

### Checklist

### DIFF
--- a/.github/Dockerfile.base
+++ b/.github/Dockerfile.base
@@ -4,7 +4,7 @@ SHELL ["/bin/bash", "-c"]
 # Set environment variables
 ENV DEBIAN_FRONTEND=noninteractive
 
-ARG TT_METAL_DEPENDENCIES_COMMIT=7aee9a53eac0023e3660668448e1fbb105038d63
+ARG TT_METAL_DEPENDENCIES_COMMIT=4e557ea4385f0c74692fc9fdce7eda039585f129
 
 # Install dependencies
 RUN <<EOT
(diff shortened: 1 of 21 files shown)


.github/workflows/call-prepare-run.yml
```diff
@@ -0,0 +1,173 @@
...
+        with:
+          fetch-depth: 0
+
+      - name: Decide on build runner
+        id: decide-runner
+        run: |
+          sh_builder=false
+          if [[ "${{ inputs.builder_build }}" == "true" ]]; then
+              # Only use builder runner for uplift jobs
+              # REF: https://github.com/tenstorrent/tt-mlir/pull/5114/files/480d8b9216ab7b578d07d500d9fc16f448135f75#r2405470436
```
vvukomanTT, line 96: note: This was true in the past, but in this PR we use "builder" machines as default?

vvukomanTT (approved):

.github/workflows/call-prepare-run.yml
nsumrakTT: Yes, the logic of selecting machine is still there but the main workflows set builder by default.

Issue #5400, opened 2025-10-20
Propagate sharding to inputs


sshonTT:
### Describe the bug

The symptom described in [#1726](https://github.com/tenstorrent/tt-xla/issues/1726)
 shows up as an invalid volume size for the operand of a reshape when running a sharded graph.

I have a workaround here: https://github.com/tenstorrent/tt-mlir/commit/0e596ac3254b760da6cc81289b3be5b260d851c1
. It forces all replicated tensors to unsharded and removes the corresponding limitation in TT-Metal.

This works in practice, but it’s heavy-handed: inputs that are legitimately replicated (and don’t need sharding) still get sent to the device, which hurts performance.

I’d appreciate feedback on two points:

- TT-Metal change: Is it acceptable to remove this limitation in TT-Metal?
- Selective reset after AggressivePropagationPass: Is there a mechanism to persist the original argument sharding state and compare it after AggressivePropagationPass, so that if the pass introduces a mismatch we can reset the argument’s shard status from presharded to unsharded only in those cases?

### Steps to reproduce the issue

In tt-xla, running torch graph below

```
def test_dg_slice():

    class ReproA(nn.Module):
        def __init__(self):
            super().__init__()
            self.fc = nn.Linear(2560, 19456, bias=False)

        def forward(self, x):
            y = self.fc(x)
            y = y[:, :9728]
            return y

    num_devices = xr.global_runtime_device_count()
    mesh_shape = (1, num_devices)
    axis_names = ("batch", "model")
    device_ids = np.array(range(num_devices))
    mesh = Mesh(device_ids=device_ids, mesh_shape=mesh_shape, axis_names=axis_names)

    model = ReproA().to(dtype=torch.bfloat16)
    model_cpu = copy.deepcopy(model)
    model = model.to(device=torch_xla.device(), dtype=torch.bfloat16)

    x = torch.randn(256, 2560, dtype=torch.bfloat16)
    x_cpu = x.clone()
    x = x.to(device=torch_xla.device(), dtype=torch.bfloat16)
    # row sharding for weight which automatically implies column sharding for input
    xs.mark_sharding(model.fc.weight, mesh, (None, "model"))

    y = model(x)

    y_cpu = model_cpu(x_cpu)
    assert torch.allclose(y.to("cpu"), y_cpu, atol=1e-1)
```

[...]

Issue #5290, opened 2025-10-09
Matmul bias that is set to replicate gets sharded in ttir


dgolubovicTT:
I am trying to run simple MNIST with tensor parallel on n300. 
I have 2 matmuls, and I shard first one outer and secon one inner.

Sharding second matmul inner requires its bias to be replicated across devices. I do that, but the ttir tracks that bias as sharded and metal runtime breaks when reshape tries to reshape that same bias:

```
E       RuntimeError: TT_FATAL @ /localdev/dgolubovic/repos/tt-xla/third_party/tt-mlir/src/tt-mlir/third_party/tt-metal/src/tt-metal/ttnn/core/tensor/tensor_utils.cpp:65: new_volume == old_volume
E       info:
E       Invalid arguments to reshape
```

Turns out that reshape expects that input tensor is sharded while it isn't.

runtime print before assert:

```
old_volume: 512, new_volume: 256
Input Shape: (1,512, )
New Shape: (1, 256, )
```

Here is the problematic reshape:

```

 "ttnn.deallocate"(%26) <{force = false}> : (tensor<32x1x256xbf16, #ttnn_layout16>) -> () loc(#loc2)
        %28 = "ttnn.reshape"(%2) <{shape = [1 : i32, 256 : i32]}> : (tensor<256xbf16, #ttnn_layout7>) -> tensor<1x256xbf16, #ttnn_layout13> loc(#loc2)

```
And input %2 is in fact bias of second matmul:

```
 %2 = "ttnn.mesh_shard"(%arg0, %1) <{shard_dims = array<i64: -1, 0>, shard_direction = #ttcore.shard_direction<full_to_shard>, shard_shape = array<i64: 2>, shard_type = #ttcore.shard_type<identity>}> : (tensor<512xbf16, #ttnn_layout2>, !ttnn.device) -> tensor<256xbf16, #ttnn_layout7> loc(#loc)
        "ttnn.deallocate"(%arg0) <{force = false}> : (tensor<512xbf16, #ttnn_layout2>) -> () loc(#loc)
```

I set bias of second matmul to replicate, not shard...

REPRO:

I reduced repro to 2 matmuls with ReLus after each:

```
class MNISTLinear(nn.Module):
    """Simple linear MNIST model for inference testing."""
    
    def __init__(self, input_size: int = 784, hidden_size: int = 512, num_classes: int = 10, bias: bool = True):
        super().__init__()
[...]

PR #5434, opened 2025-10-21
Refine DimensionShardingAttr::get usage in Shardy util

sshonTT:
### Ticket
[1726](https://github.com/tenstorrent/tt-xla/issues/1726)
[1725](https://github.com/tenstorrent/tt-xla/issues/1725)
#5400 
#5290 

### Problem description
`mlir::sdy::DimensionShardingAttr::get(context, {}, false)` currently returns an attribute with "?", which triggers unnecessary or invalid propagation during sharding inference following passes.

### What's changed
 Updated this to use `mlir::sdy::DimensionShardingAttr::get(context, {}, true)` instead, ensuring the value is treated as closed and preventing redundant or invalid propagation. Also, if it contains dimension info, it would follow the input graph's shardy open/close attribute.

### Checklist

### DIFF
--- a/test/ttmlir/Dialect/StableHLO/xla_sdy_to_sdy/open_close_xla_sdy.mlir
+++ b/test/ttmlir/Dialect/StableHLO/xla_sdy_to_sdy/open_close_xla_sdy.mlir
@@ -1,0 +1,29 @@
+// REQUIRES: stablehlo
+// RUN: ttmlir-opt -split-input-file --stablehlo-pipeline -o %t.mlir %s
+// RUN: FileCheck %s --input-file=%t.mlir
+
+// CHECK-LABEL: module @ClosedShardy
+module @ClosedShardy attributes {mhlo.cross_program_prefetches = [], mhlo.frontend_attributes = {xla.sdy.meshes = "{mesh = #sdy.mesh<[\22_axis_0\22=2]>}"}, mhlo.input_output_alias = [], mhlo.is_dynamic = false, mhlo.use_auto_spmd_partitioning = false} {
+  // CHECK: sdy.mesh @mesh = <["_axis_0_updated"=1, "_axis_0"=2]>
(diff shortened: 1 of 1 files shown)


lib/Dialect/StableHLO/Utils/ShardyUtils.cpp
```diff
@@ -684,7 +684,7 @@ parseDimensionShardings(const std::string &dimsContent,
 
     if (dimContent.empty()) {
       dimShardings.push_back(
-          mlir::sdy::DimensionShardingAttr::get(context, {}, false));
+          mlir::sdy::DimensionShardingAttr::get(context, {}, true));
```
tapspatel, line 688: While I think this is okay for all the cases you guys are encountering, I would suggest to update the parsing mechanism to scan for "open" tensor dimensions.

An open tensor dimension is indicated by a `?`.  so if you have a test case that has `xla.sdy.sharding = {?}` it should be translated into `sdy.sharding = {?}`, whereas `xla.sdy.sharding = {}` should become `sdy.sharding = {}`. In this change, you are defaulting to closed.

lib/Dialect/StableHLO/Utils/ShardyUtils.cpp
```diff
@@ -706,7 +706,7 @@ parseDimensionShardings(const std::string &dimsContent,
       }
 
       dimShardings.push_back(
-          mlir::sdy::DimensionShardingAttr::get(context, axisRefs, false));
+          mlir::sdy::DimensionShardingAttr::get(context, axisRefs, true));
```
hshahTT: I don't think we should always set `is_closed` to true here since that would falsely parse a `dimsContent` string like `{"axis_0", ?}` as a closed dim sharding when it's actually open. We should search the string for a "?" and use that to determine whether it's open or closed: https://github.com/tenstorrent/tt-mlir/commit/beaa11f1feb4892474825f12469503eb0293fa3c

lib/Dialect/StableHLO/Utils/ShardyUtils.cpp
hshahTT: +1, I think this is a simple enough fix that we should do it in this PR itself before merging in. We don't want to encounter issues related to this in the future.

hshahTT (changes requested):
Could you fix the parsing to search for a "?" and determine open vs closed shardings based on that?

lib/Dialect/StableHLO/Utils/ShardyUtils.cpp
sshonTT: Fixed it. Thank you for review!

Issue #5351, opened 2025-10-15
Run sdy passes only on non-solved graphs


hshahTT:
### Context
We are trying to uplift Shardy (https://github.com/tenstorrent/tt-mlir/pull/5282). There's a change to Shardy's `insert-explicit-reshard` pass that will insert a reshard between a function argument and the corresponding `sdy.manual_computation` op if the sharding annotations between the function arguments don't match. 

For example:
Before the uplift:
```
func.func public @main(%arg0: tensor<1x1024x128x1024xf32> {ttcore.shard_status = #ttcore.shard_status<unsharded>}) -> (tensor<1x1024x128x1024xf32> {jax.result_info = "", sdy.sharding = #sdy.sharding<@mesh, [{?}, {"x", ?}, {?}, {?}]>, ttcore.shard_status = #ttcore.shard_status<unsharded>}) {
  %0 = sdy.manual_computation(%arg0) in_shardings=[<@mesh, [{}, {"x"}, {}, {}]>] out_shardings=[<@mesh, [{}, {"x"}, {}, {}]>] manual_axes={"x", "y"} (%arg1: tensor<1x512x128x1024xf32>) {
    %1 = stablehlo.negate %arg1 : tensor<1x512x128x1024xf32>
    sdy.return %1 : tensor<1x512x128x1024xf32>
  } : (tensor<1x1024x128x1024xf32>) -> tensor<1x1024x128x1024xf32>
  return %0 : tensor<1x1024x128x1024xf32>
}
```
Notice how `%arg0` in `@main` doesn't have any associated sharding annotations, while `%arg0` in the `manual_computation` op does.

After `insert-explicit-reshard`:
```
func.func public @main(%arg0: tensor<1x1024x128x1024xf32> {ttcore.shard_status = #ttcore.shard_status<unsharded>}) -> (tensor<1x1024x128x1024xf32> {jax.result_info = "", sdy.sharding = #sdy.sharding<@mesh, [{?}, {"x", ?}, {?}, {?}]>, ttcore.shard_status = #ttcore.shard_status<unsharded>}) {
  %0 = sdy.reshard %arg0 <@mesh, [{}, {"x"}, {}, {}]> : tensor<1x1024x128x1024xf32>
[...]

PR #5282, opened 2025-10-08
llvm uplift 2025 October

nsmithtt:
### Problem description
TT-MLIR compiler references the following external projects:
- LLVM
- StableHLO
- Shardy

Last successful uplift was:
https://github.com/tenstorrent/tt-mlir/pull/4375

### What's changed
Integrating the fresh LLVM, StableHLO, and Shardy versions in the tt-mlir compiler.

Shardy: edfd6730ddfc39da5fbea8b6b202357fdf1cdb90 (October 7th)
StableHLO: 0a4440a5c8de45c4f9649bf3eb4913bf3f97da0d (October 3rd)
LLVM: 4efe170d858eb54432f520abb4e7f0086236748b (October 6th)

PR #5435, opened 2025-10-21
Skip shardy's insert-explicit-reshard pass for non-solved graphs

gengelageTT:
### Ticket
https://github.com/tenstorrent/tt-mlir/issues/5351

### Problem description
There's a change to Shardy's insert-explicit-reshard pass (in uplift https://github.com/tenstorrent/tt-mlir/pull/5282 ) that will insert a reshard between a function argument and the corresponding sdy.manual_computation op if the sharding annotations between the function arguments don't match.

This breaks how our compiler handles solved graphs going into shardy's insert-explicit-reshard pass. We want to implement a stop-gap solution to allow the uplift to continue while keeping current functionality. Until the compiler can handle these explicit reshards correctly we should wrap the shardy passes into a tt-mlir specific pass and only run them on non-solved graphs.

### What's changed
- added a template `ConditionalSdyPassWrapper` that will only run a pass if the graph is non-solved
  - eventually we want to wrap all the shardy passes with this 
- wrapped `mlir::sdy::createInsertExplicitReshardsPass()` with custom `createInsertExplicitReshardsPass()` that only runs on non-solved graphs
- added `// CHECK: %0 = sdy.manual_computation` in /Dialect/StableHLO/analyze_mesh/shardy_solved_graph.mlir to check that a reshard is not inserted 

### Checklist

### DIFF
--- a/test/ttmlir/Dialect/StableHLO/analyze_mesh/shardy_solved_graph.mlir
+++ b/test/ttmlir/Dialect/StableHLO/analyze_mesh/shardy_solved_graph.mlir
@@ -4,7 +4,7 @@
 
 module @jit_neg_shardy1 attributes {mhlo.num_partitions = 8 : i32, mhlo.num_replicas = 1 : i32} {
   sdy.mesh @mesh = <["x"=2, "y"=4]>
-  func.func public @main(%arg0: tensor<1x1024x128x1024xf32> {sdy.sharding = #sdy.sharding<@mesh, [{}, {"x"}, {}, {}]>}) -> (tensor<1x1024x128x1024xf32> {jax.result_info = ""}) {
+  func.func public @main(%arg0: tensor<1x1024x128x1024xf32>) -> (tensor<1x1024x128x1024xf32> {jax.result_info = ""}) {
     %0 = sdy.manual_computation(%arg0) in_shardings=[<@mesh, [{}, {"x"}, {}, {}]>] out_shardings=[<@mesh, [{}, {"x"}, {}, {}]>] manual_axes={"x", "y"} (%arg1: tensor<1x512x128x1024xf32>) {
       %1 = stablehlo.negate %arg1 : tensor<1x512x128x1024xf32>
       sdy.return %1 : tensor<1x512x128x1024xf32>


lib/Dialect/StableHLO/Transforms/CMakeLists.txt
```diff
@@ -3,6 +3,7 @@ add_mlir_dialect_library(MLIRStableHLOTransforms
   ApplyArgumentShardStatus.cpp
   ConvertXlaSdyToSdy.cpp
   ShardyCCLToStableHLOCCLPatterns.cpp
+  UnsolvedGraphPasses.cpp
```
tapspatel, line 6: can we keep the file name the same as the pass name for easier lookup

lib/Dialect/StableHLO/Transforms/UnsolvedGraphPasses.cpp
```diff
@@ -0,0 +1,68 @@
...
+// Specialized pass implementations using the template
+//===----------------------------------------------------------------------===//
+
+// Module-level wrapper for sdy::createInsertExplicitReshardsPass
+class InsertExplicitReshardsPass
+    : public ConditionalSdyPassWrapper<
+          impl::InsertExplicitReshardsPassBase<InsertExplicitReshardsPass>> {
+protected:
+  mlir::LogicalResult addSdyPass(mlir::PassManager &pm) override {
+    mlir::sdy::InsertExplicitReshardsPassOptions options;
```
tapspatel, line 60: nit: inline for readability

lib/Dialect/StableHLO/Transforms/UnsolvedGraphPasses.cpp
```diff
@@ -0,0 +1,68 @@
...
+
+#define GEN_PASS_DEF_INSERTEXPLICITRESHARDSPASS
+#include "ttmlir/Dialect/StableHLO/Transforms/Passes.h.inc"
+
+//===----------------------------------------------------------------------===//
+// Template for conditional SDY pass wrappers
+//===----------------------------------------------------------------------===//
+
+template <typename PassBase>
+class ConditionalSdyPassWrapper : public PassBase {
```
tapspatel, line 22: can we follow the pass structure as in the other stablehlo-pipeline passes

eg
```
class UpdateGlobalToLocalShapesPass
    : public impl::UpdateGlobalToLocalShapesPassBase<
          UpdateGlobalToLocalShapesPass> {
public:
  using impl::UpdateGlobalToLocalShapesPassBase<
      UpdateGlobalToLocalShapesPass>::UpdateGlobalToLocalShapesPassBase;

  void runOnOperation() final {
```

lib/Dialect/StableHLO/Transforms/CMakeLists.txt
gengelageTT: My thinking was eventually this would contain all the sdy passes that will be skipped using the template in this file (once some issues are debugged). Would it be preferred for each pass to have its own file?

gengelageTT: Oh or do you mean it should be named `ConditionalSdyPassWrapper`?

tapspatel: After talk offline, this is fine :)

PR #5436, opened 2025-10-21
Introduce Assign Op

### LINKED ISSUE #5452: TTNN Assign Operation Support
TT-MLIR currently lacks support for the `Assign` operation.  
This operation exists in TTNN but has not yet been integrated into the TT-MLIR compiler pipeline.  
Ref:
https://github.com/tenstorrent/tt-metal/blob/main/ttnn/cpp/ttnn/operations/data_movement/copy/copy.hpp

`Assign` is a general-purpose tensor copy operation.  
It returns a new tensor which is a copy of the input tensor, or alternatively, copies the input tensor into an optional output tensor if their shapes and memory layouts match.  
Input tensors can be of any data type, and the output tensor will have the same data type as the input.

Integrating `Assign` into the TT-MLIR pipeline will enable general tensor duplication functionality needed in various decomposition or composition workflows.

*Note:* `Assign` can also be used in scenarios where intra-device tensor replication is required, such as cases that would otherwise use same-device PointToPoint (self-P2P).

hkwonTT:
### Ticket
closes #5452 

### Problem description
The TT-MLIR compiler lacked support for the Assign operation, which is implemented in TTNN but missing from the compiler pipeline.
This prevents models that use tensor copy or in-place duplication patterns from being compiled.
Assign is also required in operation decomposition scenarios where tensor replication within the same device (e.g., same-device PointToPoint) is needed.

### What's changed
Added complete Assign operation support to TT-MLIR compiler including MLIR dialect definition, conversion patterns, runtime execution, flatbuffer serialization, operation modeling and test cases.

### Checklist

### DIFF
--- a/lib/Conversion/TTNNToEmitPy/TTNNToEmitPy.cpp
+++ b/lib/Conversion/TTNNToEmitPy/TTNNToEmitPy.cpp
@@ -3050,6 +3050,35 @@ public:
 };
 } // namespace
 
+// Assign op conversion pattern
+//
+namespace {
+class AssignOpConversionPattern
+    : public TTNNToEmitPyBaseOpConversionPattern<tt::ttnn::AssignOp> {
+public:
+  using TTNNToEmitPyBaseOpConversionPattern<
+      tt::ttnn::AssignOp>::TTNNToEmitPyBaseOpConversionPattern;
+
+  LogicalResult
+  matchAndRewrite(tt::ttnn::AssignOp srcOp, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+
+    ttnn_to_emitpy::EmitPyTTNNEmitter<tt::ttnn::AssignOp> emitter(
+        srcOp, adaptor, rewriter);
+
+    llvm::SmallVector<mlir::Attribute> args{
+        emitter.emit(srcOp.getInput()),
+        emitter.emit(srcOp.getDtype(), "dtype"),
+        emitter.emit(srcOp.getMemoryConfig(), "memory_config"),
+    };
+
+    emitter.replaceOp(*this, args);
+
+    return success();
+  }
+};
+} // namespace
+
 namespace mlir::tt {
 
 void populateTTNNToEmitPyPatterns(MLIRContext *ctx, RewritePatternSet &patterns,
@@ -3177,7 +3206,8 @@ void populateTTNNToEmitPyPatterns(MLIRContext *ctx, RewritePatternSet &patterns,
                SliceDynamicOpConversionPattern,
                SliceStaticOpConversionPattern,
                SortOpConversionPattern,
-               TransposeOpConversionPattern
+               TransposeOpConversionPattern,
+               AssignOpConversionPattern
               >(typeConverter, ctx);
   // clang-format on
 


lib/Dialect/TTNN/Interfaces/TTNNOpModelInterface.cpp
```diff
@@ -3938,4 +3938,56 @@ GlobalAvgPool2dOp::getOpRuntime(const std::vector<TTNNLayoutAttr> &inputs,
...
+  llvm::Expected<bool> check = detail::checkDeviceWorkerGrid(getOperation());
+  if (!check) {
+    return check.takeError();
+  }
+  ttcore::GridAttr deviceGrid =
+      ttcore::lookupDevice(getOperation()).getWorkerGrid();
+
+  const auto inputShape = getInput().getType().getShape();
+  const mlir::tt::ttnn::ShapeAttr shapeAttr =
+      mlir::tt::ttnn::ShapeAttr::get(getContext(), inputShape);
```
sgholamiTT: We can get rid of `shapeAttr`.

lib/Dialect/TTNN/Interfaces/TTNNOpModelInterface.cpp
```diff
@@ -3938,4 +3938,56 @@ GlobalAvgPool2dOp::getOpRuntime(const std::vector<TTNNLayoutAttr> &inputs,
...
+  const std::optional<mlir::tt::ttcore::DataType> outputDtype =
+      getOutputDtype();
+  const std::optional<::mlir::TypedValue<::mlir::RankedTensorType>>
+      optionalOutputTensor = getOptionalOutputTensor()
+                                 ? std::make_optional(getOptionalOutputTensor())
+                                 : std::nullopt;
+
+  return opConstraintsCache().getOrCompute(
+      op_model::OpModel<mlir::tt::ttnn::AssignOp>::getOpConstraints, *this,
+      deviceGrid, shapeAttr, inputs[0], outputMemConfig, outputDtype,
```
sgholamiTT: We can pass `inputShape` here:
```c++,
deviceGrid, inputShape, inputs[0], ...
```

lib/Dialect/TTNN/Interfaces/TTNNOpModelInterface.cpp
```diff
@@ -3938,4 +3938,56 @@ GlobalAvgPool2dOp::getOpRuntime(const std::vector<TTNNLayoutAttr> &inputs,
...
+  const auto inputShape = getInput().getType().getShape();
+  const mlir::tt::ttnn::ShapeAttr shapeAttr =
+      mlir::tt::ttnn::ShapeAttr::get(getContext(), inputShape);
+  const mlir::tt::ttnn::MemoryConfigAttr outputMemConfig = getOutputMemConfig();
+  const std::optional<mlir::tt::ttcore::DataType> outputDtype =
+      getOutputDtype();
+  const std::optional<::mlir::TypedValue<::mlir::RankedTensorType>>
+      optionalOutputTensor = getOptionalOutputTensor()
+                                 ? std::make_optional(getOptionalOutputTensor())
+                                 : std::nullopt;
```
sgholamiTT: Is there a reason to store the return value of `getOptionalOutputTensor()` in a variable ? Can't we just pass that to the function directly ?

lib/Dialect/TTNN/Interfaces/TTNNOpModelInterface.cpp
```diff
@@ -3938,4 +3938,56 @@ GlobalAvgPool2dOp::getOpRuntime(const std::vector<TTNNLayoutAttr> &inputs,
...
+  assert(inputs.size() == 1);
+
+  const auto inputShape = getInput().getType().getShape();
+  const std::optional<mlir::tt::ttcore::DataType> outputDtype =
+      getOutputDtype();
+  const mlir::tt::ttnn::MemoryConfigAttr outputMemConfig = getOutputMemConfig();
+  const std::optional<::mlir::TypedValue<::mlir::RankedTensorType>>
+      optionalOutputTensor = getOptionalOutputTensor()
+                                 ? std::make_optional(getOptionalOutputTensor())
+                                 : std::nullopt;
```
sgholamiTT: Same here. Let's just pass `getOptionalOutputTensor()` to the function:
`opRuntimeCache().getOrCompute(..., getOptionalOutputTensor())`

lib/Dialect/TTNN/Interfaces/TTNNOpModelInterface.cpp
```diff
@@ -3938,4 +3938,56 @@ GlobalAvgPool2dOp::getOpRuntime(const std::vector<TTNNLayoutAttr> &inputs,
...
+  if (!check) {
+    return check.takeError();
+  }
+  ttcore::GridAttr deviceGrid =
+      ttcore::lookupDevice(getOperation()).getWorkerGrid();
+
+  const auto inputShape = getInput().getType().getShape();
+  const mlir::tt::ttnn::ShapeAttr shapeAttr =
+      mlir::tt::ttnn::ShapeAttr::get(getContext(), inputShape);
+  const mlir::tt::ttnn::MemoryConfigAttr outputMemConfig = getOutputMemConfig();
```
sgholamiTT: Shouldn't we use `opConfig.outputLayout` instead of this ? @rpavlovicTT @odjuricicTT 
I'm really not sure since the backend expects a non-optional `const MemoryConfig& output_mem_config`. If we pass `opConfig.outputLayout` we'd be able to retrieve an `std::optional<MemoryConfig>` out of it.

lib/Dialect/TTNN/Interfaces/TTNNOpModelInterface.cpp
```diff
@@ -3938,4 +3938,56 @@ GlobalAvgPool2dOp::getOpRuntime(const std::vector<TTNNLayoutAttr> &inputs,
       inputs[0], getDtype(), opConfig.outputLayout);
 }
 
+//===----------------------------------------------------------------------===//
+// AssignOp - TTNN Op Model Interface
+//===----------------------------------------------------------------------===//
+
+llvm::Expected<op_model::OpConstraints>
+AssignOp::getOpConstraints(const std::vector<TTNNLayoutAttr> &inputs,
+                           const OpConfig &config) {
```
sgholamiTT: Nit: `const OpConfig &opConfig) {` for consistency.

lib/OpModel/TTNN/TTNNOpModel.cpp
```diff
@@ -5946,4 +5946,123 @@ OpModel<ConstantOp>::getOpConstraints(ttcore::GridAttr deviceGrid,
...
+                                           metalMemConfig, metalOutputDtype,
+                                           optionalOutputTensorSpec);
+  };
+
+  return operation::getOpRuntime(assignOpQuery);
+#else
+  return llvm::createStringError("Not Implemented");
+#endif // TTMLIR_ENABLE_OPMODEL
+}
+
```
sgholamiTT, line 6041: Let's add tests in `TestOpModelLib.cpp` and `TestOpModelInterface.cpp` for this op as well as described [here](https://github.com/tenstorrent/tt-mlir/blob/main/docs/src/ttnn-op-constraints.md#step-3-add-unit-tests) and [here](https://github.com/tenstorrent/tt-mlir/blob/main/docs/src/ttnn-op-constraints.md#step-4-add-integration-tests).

sgholamiTT (commented):
Thanks for doing this. OpModel changes LGTM.
Few minor comments inlined.

lib/Dialect/TTNN/Interfaces/TTNNOpModelInterface.cpp
hkwonTT: I originally thought I couldn’t use `::mlir::TypedValue<::mlir::RankedTensorType>` as an `optional`, so I wrapped it in `std::optional`. After checking other ops, I realized they pass `optional<shape>` and `optional<layoutattr>` for attributes of type `optional<RankedTensorType>`. I updated my code to match the pattern used by the other operations.

lib/OpModel/TTNN/TTNNOpModel.cpp
hkwonTT: Some test cases were added to both files.

lib/Dialect/TTNN/Interfaces/TTNNOpModelInterface.cpp
hkwonTT: Some ops (`ToMemoryConfigOp`, `EmptyOp`, `RandOp`) that need a non-optional `memory config` attribute pass the specified memory config, while the others (`PrepareConv2dWeightsOp`, `PrepareConv2dBiasOp`) pass the input layout attribute. 

An inconsistency between the `memory config` on the output (or input) tensor and the specified `memory config` attribute basically means the IR is invalid. Verifying this should be the responsibility of the verifier in `TTNNOps.cpp`. In my view, passing the specified attribute instead of the output layout is appropriate if we have an explicit attribute. If we don’t use the specified attribute and rely on the output layout instead, the unused attribute should be removed from the IR definition. I’m not an expert on this and will follow the guidance of the `forge-developers-mlir-core` group.

lib/Dialect/TTNN/Interfaces/TTNNOpModelInterface.cpp
```diff
@@ -3938,4 +3938,71 @@ GlobalAvgPool2dOp::getOpRuntime(const std::vector<TTNNLayoutAttr> &inputs,
...
+llvm::Expected<op_model::OpConstraints>
+AssignOp::getOpConstraints(const std::vector<TTNNLayoutAttr> &inputs,
+                           const OpConfig &opConfig) {
+  if (getOptionalOutputTensor()) {
+    assert(inputs.size() == 2);
+    // ttnn::assign with optional output tensor must have exactly 2 inputs
+  } else {
+    assert(inputs.size() == 1);
+    // ttnn::assign without optional output tensor must have exactly 1 input
+  }
```
sgholamiTT: Nit: `assert(inputs.size() == (1 + (getOptionalOutputTensor() == nullptr ? 0 : 1)));`

lib/Dialect/TTNN/Interfaces/TTNNOpModelInterface.cpp
```diff
@@ -3938,4 +3938,71 @@ GlobalAvgPool2dOp::getOpRuntime(const std::vector<TTNNLayoutAttr> &inputs,
...
+llvm::Expected<size_t>
+AssignOp::getOpRuntime(const std::vector<TTNNLayoutAttr> &inputs,
+                       const OpConfig &opConfig) {
+  if (getOptionalOutputTensor()) {
+    assert(inputs.size() == 2);
+    // ttnn::assign with optional output tensor must have exactly 2 inputs
+  } else {
+    assert(inputs.size() == 1);
+    // ttnn::assign without optional output tensor must have exactly 1 input
+  }
```
sgholamiTT: Similar as above.

include/ttmlir/Dialect/TTNN/IR/TTNNOps.td
```diff
@@ -2749,4 +2749,24 @@ def TTNN_GlobalAvgPool2dOp: TTNN_Op<"global_avg_pool2d", [
...
+      Alternatively, copies input tensor ``input`` to ``optional_output_tensor``
+      if their shapes and memory layouts match, and returns input_b tensor.
+      Input tensors can be of any data type.
+      Output tensor will be of same data type as Input tensor.
+    }];
+
+    let arguments = (ins AnyRankedTensor:$input,
+                         TTNN_MemoryConfigAttr:$output_mem_config,
+                         OptionalAttr<TTCore_DataTypeAttr>:$output_dtype,
+                         Optional<AnyRankedTensor>:$optional_output_tensor);
```
sdjordjevicTT: Do we really see a need for a DPS variant of this op? Can we remove the optional_output_tensor please.

include/ttmlir/Dialect/TTNN/IR/TTNNOps.td
```diff
@@ -2749,4 +2749,24 @@ def TTNN_GlobalAvgPool2dOp: TTNN_Op<"global_avg_pool2d", [
...
+    let description = [{
+      Returns a new tensor which is a new copy of input tensor.
+      Alternatively, copies input tensor ``input`` to ``optional_output_tensor``
+      if their shapes and memory layouts match, and returns input_b tensor.
+      Input tensors can be of any data type.
+      Output tensor will be of same data type as Input tensor.
+    }];
+
+    let arguments = (ins AnyRankedTensor:$input,
+                         TTNN_MemoryConfigAttr:$output_mem_config,
```
sdjordjevicTT: Can you rename this attribute to $memory_config and attach the TTNN_MemoryConfigOpInterface trait. This will give you a layout and memory config validation for free.

include/ttmlir/Dialect/TTNN/IR/TTNNOps.td
```diff
@@ -2749,4 +2749,24 @@ def TTNN_GlobalAvgPool2dOp: TTNN_Op<"global_avg_pool2d", [
...
+                         OptionalAttr<TTCore_DataTypeAttr>:$output_dtype,
```
sdjordjevicTT: Can you rename this attribute to dtype and attach the TTNN_DtypeOpInterface trait? This will give you a layout and dtype validation for free.

include/ttmlir/Dialect/TTNN/IR/TTNNOps.td
```diff
@@ -2749,4 +2749,24 @@ def TTNN_GlobalAvgPool2dOp: TTNN_Op<"global_avg_pool2d", [
 
   let hasVerifier = 1;
 }
+
+def TTNN_AssignOp : TTNN_Op<"assign"> {
```
sdjordjevicTT: This is like torch equivalent of clone op, right? Do we need the TTIR version of this op?

lib/Dialect/TTNN/IR/TTNNOps.cpp
```diff
@@ -4126,4 +4126,67 @@ ::mlir::LogicalResult mlir::tt::ttnn::GlobalAvgPool2dOp::verify() {
...
+  // Verify shape compatibility.
+  if (inputType.getShape() != outputType.getShape()) {
+    return emitOpError() << "input and output tensor must have the same shape";
+  }
+
+  TTNNLayoutAttr outputLayoutAttr =
+      mlir::cast<TTNNLayoutAttr>(outputType.getEncoding());
+  // MemoryConfig
+
+  if (getOutputMemConfig().getBufferType().getValue() !=
```
sdjordjevicTT: All of this can be removed with the traits that I suggested to be add in the op definition in TD file.

lib/Dialect/TTNN/IR/TTNNOps.cpp
```diff
@@ -4126,4 +4126,67 @@ ::mlir::LogicalResult mlir::tt::ttnn::GlobalAvgPool2dOp::verify() {
...
+        "Buffer type mismatch between op and output memory config");
+  }
+  if (getOutputMemConfig().getTensorMemoryLayout() !=
+      outputLayoutAttr.getMemLayout()) {
+    return emitOpError(
+        "Tensor memory layout mismatch between op and output memory config");
+  }
+
+  // Verify optional output tensor if present.
+  if (getOptionalOutputTensor()) {
```
sdjordjevicTT: We can remove this as well if we decide not to use the DPS op variant, which I don't see a need for it.

include/ttmlir/Dialect/TTNN/IR/TTNNOps.td
hkwonTT: I just tried to align it with the [TTNN op definition](https://github.com/tenstorrent/tt-metal/blob/c9561fffe7cc5e2e5445e70fbcae0c62153fd5e6/ttnn/cpp/ttnn/operations/data_movement/copy/copy.hpp#L23). I don’t see any need for the DPS variant of this op either. If someone needs a DPS variant in the future, they can update the op definition. I’ll remove it. Thanks.

include/ttmlir/Dialect/TTNN/IR/TTNNOps.td
hkwonTT: Thanks for the kind suggestion. Is TTNN_DtypeOpInterface trait applicable to `OptionalAttr<TTCore_DataTypeAttr>`? 
[The verifier](https://github.com/tenstorrent/tt-mlir/blob/04762bc0236d9eb5dd45855db11d31dd74493a6e/include/ttmlir/Dialect/TTNN/Interfaces/TTNNTensorSpecInterface.h#L29) emits an OpError when the attribute does not exist.

include/ttmlir/Dialect/TTNN/IR/TTNNOps.td
hkwonTT: I need the `ttnn.assign` op to decompose certain CCL ops—like `ttir.all_to_all`—into multiple assign (or clone) ops and PointToPoint ops. In my view, the assign op doesn’t need to be added to the ttir op definition. We might need a clone-like op in ttir in the future. Do you think it would be better to add it now?

jnie-TT (approved):

include/ttmlir/Dialect/TTNN/IR/TTNNOps.td
sdjordjevicTT: Great, thanks!

include/ttmlir/Dialect/TTNN/IR/TTNNOps.td
sdjordjevicTT: Yeah, take a look, for example, at:
https://github.com/tenstorrent/tt-mlir/blob/0dc96fbf3e375ad8b529840155dcf704c91a5abb/include/ttmlir/Dialect/TTNN/IR/TTNNOps.td#L1034

You will need something similar to this, as you need to specify this dtype attribute during TTIR to TTNN conversion:
https://github.com/tenstorrent/tt-mlir/blob/0dc96fbf3e375ad8b529840155dcf704c91a5abb/lib/Conversion/TTIRToTTNN/TTIRToTTNN.cpp#L431

include/ttmlir/Dialect/TTNN/IR/TTNNOps.td
sdjordjevicTT: No worries, we can add it later if needed.

include/ttmlir/Dialect/TTNN/IR/TTNNOps.td
sdjordjevicTT: Oh, I see, you don't have a conversion, hence you don't have this place to add. I think we should keep this trait in order to keep in sync this attribute with the output layout. The verifier should fail if that attribute is not in sync with the output layout. The responsibility is on the one who creates this op in pure ttnn dialect to always pass this attribute.

hkwonTT: I'll keep that in mind when I actually use the assign op in another PR. Thank you!

PR #5437, opened 2025-10-22
GenericOp Virtual Grid Support

bgrady-tt:
### Ticket
Link to Github Issue

### Problem description
The new mechanism implementing virtual grids is the **'core virtualization' map** within `ShardLayoutAttr`. Wherever worker core indexes are used (i.e. DMA stream index calculation), they are first translated from their raw physical coordinates to coordinates in virtual grid space. This makes indexing and block factors within the GenericOp are purely in terms of the apparent (viewed) shapes of the operands.

image

### Checklist

### DIFF
--- a/include/ttmlir/Dialect/TTCore/IR/TTCoreOpsTypes.td
+++ b/include/ttmlir/Dialect/TTCore/IR/TTCoreOpsTypes.td
@@ -273,11 +273,11 @@ def TTCore_ShardLayoutAttr : TTCore_Attr<"ShardLayout", "shard", [TTCore_DeviceL
     a grid of cores.
   }];
   let parameters = (ins ArrayRefParameter<"int64_t">:$stride,
-                        DefaultValuedParameter<"uint32_t", "1">:$buffers,
+                        "uint32_t":$buffers,
                         DefaultValuedParameter<"AffineMap", "$_builder.getEmptyAffineMap()">:$coreVirtualizationMap
                         );
 
-  let assemblyFormat = "`<` custom<DimensionList>($stride) (`,` $buffers^)? (`,` $coreVirtualizationMap^)? `>`";
+  let assemblyFormat = "`<` custom<DimensionList>($stride) `,` $buffers (`,` $coreVirtualizationMap^)? `>`";
 
   let extraClassDeclaration = [{
     static ShardLayoutAttr get(::mlir::MLIRContext *context, ArrayRef<int64_t> shape, uint64_t elementSize, uint32_t buffers);
--- a/test/ttmlir/Conversion/D2MToTTKernel/dma_lowering.mlir
+++ b/test/ttmlir/Conversion/D2MToTTKernel/dma_lowering.mlir
@@ -22,13 +22,13 @@ func.func @test_local_to_local_same_core(%arg0: memref<2x2x!ttcore.tile<32x32, f
 // Test 2: Local to local DMA between different cores
 // CHECK-LABEL: func.func @test_local_to_local_remote_core
 func.func @test_local_to_local_remote_core(%arg0: memref<2x2x!ttcore.tile<32x32, f32>, #l1_>, %arg1: memref<2x2x!ttcore.tile<32x32, f32>, #l1_>) attributes {d2m.thread = #d2m.thread<datamovement>} {
-  %dst = d2m.get_global_operand(0) : memref<1x2x2x2x!ttcore.tile<32x32, f32>, #ttcore.shard<16384x4096>, #l1_>
+  %dst = d2m.get_global_operand(0) : memref<1x2x2x2x!ttcore.tile<32x32, f32>, #ttcore.shard<16384x4096, 1>, #l1_>
   %c0 = arith.constant 0 : index
   %c1 = arith.constant 1 : index
   // CHECK: ttkernel.get_read_ptr
   // CHECK: ttkernel.get_noc_addr
   // CHECK: ttkernel.noc_async_write
-  %0 = d2m.dma %arg0[%c0, %c0], %dst[%c0, %c1, %c0, %c0] : (memref<2x2x!ttcore.tile<32x32, f32>, #l1_>, memref<1x2x2x2x!ttcore.tile<32x32, f32>, #ttcore.shard<16384x4096>, #l1_>) -> !d2m.mem_tx
+  %0 = d2m.dma %arg0[%c0, %c0], %dst[%c0, %c1, %c0, %c0] : (memref<2x2x!ttcore.tile<32x32, f32>, #l1_>, memref<1x2x2x2x!ttcore.tile<32x32, f32>, #ttcore.shard<16384x4096, 1>, #l1_>) -> !d2m.mem_tx
   d2m.dma_wait %0
   return
 }
@@ -72,11 +72,11 @@ func.func @test_local_to_remote_multicast_loopback(%arg0: memref<2x2x!ttcore.til
 func.func @test_remote_l1_to_local(%dst: memref<2x2x!ttcore.tile<32x32, f32>, #l1_>) attributes {d2m.thread = #d2m.thread<datamovement>} {
   %c0 = arith.constant 0 : index
   %c1 = arith.constant 1 : index
-  %src = d2m.get_global_operand(0) : memref<1x2x2x2x!ttcore.tile<32x32, f32>, #ttcore.shard<16384x4096>, #l1_>
+  %src = d2m.get_global_operand(0) : memref<1x2x2x2x!ttcore.tile<32x32, f32>, #ttcore.shard<16384x4096, 1>, #l1_>
   // CHECK: ttkernel.get_noc_addr
   // CHECK: ttkernel.get_write_ptr
   // CHECK: ttkernel.noc_async_read
-  %0 = d2m.dma %src[%c0, %c1, %c0, %c0], %dst[%c0, %c0] : (memref<1x2x2x2x!ttcore.tile<32x32, f32>, #ttcore.shard<16384x4096>, #l1_>, memref<2x2x!ttcore.tile<32x32, f32>, #l1_>) -> !d2m.mem_tx
+  %0 = d2m.dma %src[%c0, %c1, %c0, %c0], %dst[%c0, %c0] : (memref<1x2x2x2x!ttcore.tile<32x32, f32>, #ttcore.shard<16384x4096, 1>, #l1_>, memref<2x2x!ttcore.tile<32x32, f32>, #l1_>) -> !d2m.mem_tx
   d2m.dma_wait %0
   return
 }
@@ -86,8 +86,8 @@ func.func @test_remote_l1_to_local(%dst: memref<2x2x!ttcore.tile<32x32, f32>, #l
 func.func @test_dma_wait_read(%dst: memref<2x2x!ttcore.tile<32x32, f32>, #l1_>) attributes {d2m.thread = #d2m.thread<datamovement>} {
   %c0 = arith.constant 0 : index
   %c1 = arith.constant 1 : index
-  %src = d2m.get_global_operand(0) : memref<1x2x2x2x!ttcore.tile<32x32, f32>, #ttcore.shard<16384x4096>, #l1_>
-  %0 = d2m.dma %src[%c0, %c1, %c0, %c0], %dst[%c0, %c0] : (memref<1x2x2x2x!ttcore.tile<32x32, f32>, #ttcore.shard<16384x4096>, #l1_>, memref<2x2x!ttcore.tile<32x32, f32>, #l1_>) -> !d2m.mem_tx
+  %src = d2m.get_global_operand(0) : memref<1x2x2x2x!ttcore.tile<32x32, f32>, #ttcore.shard<16384x4096, 1>, #l1_>
+  %0 = d2m.dma %src[%c0, %c1, %c0, %c0], %dst[%c0, %c0] : (memref<1x2x2x2x!ttcore.tile<32x32, f32>, #ttcore.shard<16384x4096, 1>, #l1_>, memref<2x2x!ttcore.tile<32x32, f32>, #l1_>) -> !d2m.mem_tx
   // CHECK: ttkernel.noc_async_read_barrier
   d2m.dma_wait %0
   return
@@ -108,13 +108,13 @@ func.func @test_dma_wait_write(%arg0: memref<2x2x!ttcore.tile<32x32, f32>, #l1_>
 func.func @test_remote_to_local_partial(%dst: memref<2x2x!ttcore.tile<32x32, f32>, #l1_>) attributes {d2m.thread = #d2m.thread<datamovement>} {
   %c0 = arith.constant 0 : index
   %c1 = arith.constant 1 : index
-  %src = d2m.get_global_operand(0) : memref<1x2x2x2x!ttcore.tile<32x32, f32>, #ttcore.shard<16384x4096>, #l1_>
+  %src = d2m.get_global_operand(0) : memref<1x2x2x2x!ttcore.tile<32x32, f32>, #ttcore.shard<16384x4096, 1>, #l1_>
   // CHECK: scf.for
   // CHECK: scf.for
   // CHECK: ttkernel.get_noc_addr
   // CHECK: ttkernel.get_write_ptr
   // CHECK: ttkernel.noc_async_read
-  %0 = d2m.dma %src[%c0, %c1], %dst : (memref<1x2x2x2x!ttcore.tile<32x32, f32>, #ttcore.shard<16384x4096>, #l1_>, memref<2x2x!ttcore.tile<32x32, f32>, #l1_>) -> !d2m.mem_tx
+  %0 = d2m.dma %src[%c0, %c1], %dst : (memref<1x2x2x2x!ttcore.tile<32x32, f32>, #ttcore.shard<16384x4096, 1>, #l1_>, memref<2x2x!ttcore.tile<32x32, f32>, #l1_>) -> !d2m.mem_tx
   d2m.dma_wait %0
   return
 }
--- a/test/ttmlir/Conversion/D2MToTTKernel/dram_dma_lowering.mlir
+++ b/test/ttmlir/Conversion/D2MToTTKernel/dram_dma_lowering.mlir
@@ -11,7 +11,7 @@ module {
 
 // Test 1: Small DRAM read, only accessing bank 0
 // CHECK-LABEL: func.func private @datamovement_kernel0
-func.func @test_dram_read_onebank(%arg0: memref<1x1x32x32xf32, #ttcore.shard<128x4>, #ttcore.memory_space<dram>>) {
+func.func @test_dram_read_onebank(%arg0: memref<1x1x32x32xf32, #ttcore.shard<128x4, 1>, #ttcore.memory_space<dram>>) {
 
   // Should be lowered to a single noc_async_read() call; NO loops or guards should be generated here
   //   CHECK-NOT: scf.for
@@ -20,11 +20,11 @@ func.func @test_dram_read_onebank(%arg0: memref<1x1x32x32xf32, #ttcore.shard<128
   //   CHECK: ttkernel.get_write_ptr
   //   CHECK: ttkernel.noc_async_read
 
-  %alloc_1 = memref.alloc() : memref<1x1x32x32xf32, #ttcore.shard<128x4>, #ttcore.memory_space<l1>>
-  %view = d2m.view_layout %arg0 : memref<1x1x32x32xf32, #ttcore.shard<128x4>, #ttcore.memory_space<dram>> -> memref<1x1x32x32xf32, #ttcore.view<(d0, d1, d2, d3) -> (d0 + d2 floordiv 32, d1 + d3 floordiv 32, d2 mod 32, d3 mod 32)>, #ttcore.memory_space<dram>>
+  %alloc_1 = memref.alloc() : memref<1x1x32x32xf32, #ttcore.shard<128x4, 1>, #ttcore.memory_space<l1>>
+  %view = d2m.view_layout %arg0 : memref<1x1x32x32xf32, #ttcore.shard<128x4, 1>, #ttcore.memory_space<dram>> -> memref<1x1x32x32xf32, #ttcore.view<(d0, d1, d2, d3) -> (d0 + d2 floordiv 32, d1 + d3 floordiv 32, d2 mod 32, d3 mod 32)>, #ttcore.memory_space<dram>>
   d2m.generic {block_factors = [1, 1], grid = #ttcore.grid<1x1>, indexing_maps = [affine_map<(d0, d1) -> (d0, d1)>, affine_map<(d0, d1) -> (d0, d1)>], iterator_types = [#ttcore.iterator_type<parallel>, #ttcore.iterator_type<parallel>], threads = [#d2m.thread<datamovement>]}
       ins(%view : memref<1x1x32x32xf32, #ttcore.view<(d0, d1, d2, d3) -> (d0 + d2 floordiv 32, d1 + d3 floordiv 32, d2 mod 32, d3 mod 32)>, #ttcore.memory_space<dram>>)
-      outs(%alloc_1 : memref<1x1x32x32xf32, #ttcore.shard<128x4>, #ttcore.memory_space<l1>>)  {
+      outs(%alloc_1 : memref<1x1x32x32xf32, #ttcore.shard<128x4, 1>, #ttcore.memory_space<l1>>)  {
   ^datamovement0(%cb0: !d2m.cb<memref<32x32xf32, #ttcore.memory_space<dram>>>, %cb1: !d2m.cb<memref<32x32xf32, #ttcore.memory_space<l1>>>):
     %buf = d2m.reserve %cb1 : !d2m.cb<memref<32x32xf32, #ttcore.memory_space<l1>>> -> memref<32x32xf32, #ttcore.memory_space<l1>>
     %tx = d2m.dma %view<affine_map<(d0, d1) -> (d0, d1)>>, %buf : (memref<1x1x32x32xf32, #ttcore.view<(d0, d1, d2, d3) -> (d0 + d2 floordiv 32, d1 + d3 floordiv 32, d2 mod 32, d3 mod 32)>, #ttcore.memory_space<dram>>, memref<32x32xf32, #ttcore.memory_space<l1>>) -> !d2m.mem_tx
@@ -35,7 +35,7 @@ func.func @test_dram_read_onebank(%arg0: memref<1x1x32x32xf32, #ttcore.shard<128
 
 // Test 2: DRAM read spanning all DRAM banks
 // CHECK-LABEL: func.func private @datamovement_kernel1
-func.func @test_dram_read_multibank(%arg0: memref<1x1x128x128xf32, #ttcore.shard<128x4>, #ttcore.memory_space<dram>>) {
+func.func @test_dram_read_multibank(%arg0: memref<1x1x128x128xf32, #ttcore.shard<128x4, 1>, #ttcore.memory_space<dram>>) {
 
   // Should be lowered to a guarded, coalesced gather loop
   //   CHECK: scf.for
@@ -45,11 +45,11 @@ func.func @test_dram_read_multibank(%arg0: memref<1x1x128x128xf32, #ttcore.shard
   //   CHECK: ttkernel.get_write_ptr
   //   CHECK: ttkernel.noc_async_read
 
-  %alloc_1 = memref.alloc() : memref<1x1x128x128xf32, #ttcore.shard<128x4>, #ttcore.memory_space<l1>>
-  %view = d2m.view_layout %arg0 : memref<1x1x128x128xf32, #ttcore.shard<128x4>, #ttcore.memory_space<dram>> -> memref<1x1x128x128xf32, #ttcore.view<(d0, d1, d2, d3) -> (d0 + d2 floordiv 32, d1 + d3 floordiv 32, d2 mod 32, d3 mod 32)>, #ttcore.memory_space<dram>>
+  %alloc_1 = memref.alloc() : memref<1x1x128x128xf32, #ttcore.shard<128x4, 1>, #ttcore.memory_space<l1>>
+  %view = d2m.view_layout %arg0 : memref<1x1x128x128xf32, #ttcore.shard<128x4, 1>, #ttcore.memory_space<dram>> -> memref<1x1x128x128xf32, #ttcore.view<(d0, d1, d2, d3) -> (d0 + d2 floordiv 32, d1 + d3 floordiv 32, d2 mod 32, d3 mod 32)>, #ttcore.memory_space<dram>>
   d2m.generic {block_factors = [1, 1], grid = #ttcore.grid<1x1>, indexing_maps = [affine_map<(d0, d1) -> (d0, d1)>, affine_map<(d0, d1) -> (d0, d1)>], iterator_types = [#ttcore.iterator_type<parallel>, #ttcore.iterator_type<parallel>], threads = [#d2m.thread<datamovement>]}
       ins(%view : memref<1x1x128x128xf32, #ttcore.view<(d0, d1, d2, d3) -> (d0 + d2 floordiv 32, d1 + d3 floordiv 32, d2 mod 32, d3 mod 32)>, #ttcore.memory_space<dram>>)
-      outs(%alloc_1 : memref<1x1x128x128xf32, #ttcore.shard<128x4>, #ttcore.memory_space<l1>>)  {
+      outs(%alloc_1 : memref<1x1x128x128xf32, #ttcore.shard<128x4, 1>, #ttcore.memory_space<l1>>)  {
   ^datamovement0(%cb0: !d2m.cb<memref<128x128xf32, #ttcore.memory_space<dram>>>, %cb1: !d2m.cb<memref<128x128xf32, #ttcore.memory_space<l1>>>):
     %buf = d2m.reserve %cb1 : !d2m.cb<memref<128x128xf32, #ttcore.memory_space<l1>>> -> memref<128x128xf32, #ttcore.memory_space<l1>>
     %tx = d2m.dma %view<affine_map<(d0, d1) -> (d0, d1)>>, %buf : (memref<1x1x128x128xf32, #ttcore.view<(d0, d1, d2, d3) -> (d0 + d2 floordiv 32, d1 + d3 floordiv 32, d2 mod 32, d3 mod 32)>, #ttcore.memory_space<dram>>, memref<128x128xf32, #ttcore.memory_space<l1>>) -> !d2m.mem_tx
--- a/test/ttmlir/Conversion/D2MToTTKernel/single_tile_ops.mlir
+++ b/test/ttmlir/Conversion/D2MToTTKernel/single_tile_ops.mlir
@@ -18,12 +18,12 @@ module {
   // TTIR FPU operations
   //===----------------------------------------------------------------------===//
 
-  func.func @test_matmul_lowering(%in0: memref<1x1x1x1x!ttype_f32, #ttcore.shard<4096x4096>, #l1_>,
-                                  %in1: memref<1x1x1x1x!ttype_f32, #ttcore.shard<4096x4096>, #l1_>,
-                                  %out: memref<1x1x1x1x!ttype_f32, #ttcore.shard<4096x4096>, #l1_>) {
+  func.func @test_matmul_lowering(%in0: memref<1x1x1x1x!ttype_f32, #ttcore.shard<4096x4096, 1>, #l1_>,
+                                  %in1: memref<1x1x1x1x!ttype_f32, #ttcore.shard<4096x4096, 1>, #l1_>,
+                                  %out: memref<1x1x1x1x!ttype_f32, #ttcore.shard<4096x4096, 1>, #l1_>) {
     d2m.generic {block_factors = [1, 1, 1], grid = #ttcore.grid<1x1>, indexing_maps = [#map1_, #map2_, #map3_], iterator_types = [#parallel_, #parallel_, #reduction_], threads = [#d2m.thread<compute>]}
-        ins(%in0, %in1 : memref<1x1x1x1x!ttype_f32, #ttcore.shard<4096x4096>, #l1_>, memref<1x1x1x1x!ttype_f32, #ttcore.shard<4096x4096>, #l1_>)
-        outs(%out : memref<1x1x1x1x!ttype_f32, #ttcore.shard<4096x4096>, #l1_>)  {
+        ins(%in0, %in1 : memref<1x1x1x1x!ttype_f32, #ttcore.shard<4096x4096, 1>, #l1_>, memref<1x1x1x1x!ttype_f32, #ttcore.shard<4096x4096, 1>, #l1_>)
+        outs(%out : memref<1x1x1x1x!ttype_f32, #ttcore.shard<4096x4096, 1>, #l1_>)  {
     ^compute0(%arg0_cb: !d2m.cb<memref<1x1x!ttype_f32, #l1_>>, %arg1_cb: !d2m.cb<memref<1x1x!ttype_f32, #l1_>>, %arg2_cb: !d2m.cb<memref<1x1x!ttype_f32, #l1_>>):
       %cb0 = d2m.wait %arg0_cb : !d2m.cb<memref<1x1x!ttype_f32, #l1_>> -> memref<1x1x!ttype_f32, #l1_>
       %cb1 = d2m.wait %arg1_cb : !d2m.cb<memref<1x1x!ttype_f32, #l1_>> -> memref<1x1x!ttype_f32, #l1_>
@@ -42,12 +42,12 @@ module {
     return
   }
 
-  func.func @test_add_lowering(%in0: memref<1x1x1x1x!ttype_f32, #ttcore.shard<4096x4096>, #l1_>,
-                               %in1: memref<1x1x1x1x!ttype_f32, #ttcore.shard<4096x4096>, #l1_>,
-                               %out: memref<1x1x1x1x!ttype_f32, #ttcore.shard<4096x4096>, #l1_>) {
+  func.func @test_add_lowering(%in0: memref<1x1x1x1x!ttype_f32, #ttcore.shard<4096x4096, 1>, #l1_>,
+                               %in1: memref<1x1x1x1x!ttype_f32, #ttcore.shard<4096x4096, 1>, #l1_>,
+                               %out: memref<1x1x1x1x!ttype_f32, #ttcore.shard<4096x4096, 1>, #l1_>) {
     d2m.generic {block_factors = [1, 1], grid = #ttcore.grid<1x1>, indexing_maps = [#map_, #map_, #map_], iterator_types = [#parallel_, #parallel_], threads = [#d2m.thread<compute>]}
-        ins(%in0, %in1 : memref<1x1x1x1x!ttype_f32, #ttcore.shard<4096x4096>, #l1_>, memref<1x1x1x1x!ttype_f32, #ttcore.shard<4096x4096>, #l1_>)
-        outs(%out : memref<1x1x1x1x!ttype_f32, #ttcore.shard<4096x4096>, #l1_>)  {
+        ins(%in0, %in1 : memref<1x1x1x1x!ttype_f32, #ttcore.shard<4096x4096, 1>, #l1_>, memref<1x1x1x1x!ttype_f32, #ttcore.shard<4096x4096, 1>, #l1_>)
+        outs(%out : memref<1x1x1x1x!ttype_f32, #ttcore.shard<4096x4096, 1>, #l1_>)  {
     ^compute0(%arg0_cb: !d2m.cb<memref<1x1x!ttype_f32, #l1_>>, %arg1_cb: !d2m.cb<memref<1x1x!ttype_f32, #l1_>>, %arg2_cb: !d2m.cb<memref<1x1x!ttype_f32, #l1_>>):
       %cb0 = d2m.wait %arg0_cb : !d2m.cb<memref<1x1x!ttype_f32, #l1_>> -> memref<1x1x!ttype_f32, #l1_>
       %cb1 = d2m.wait %arg1_cb : !d2m.cb<memref<1x1x!ttype_f32, #l1_>> -> memref<1x1x!ttype_f32, #l1_>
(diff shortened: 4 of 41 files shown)


include/ttmlir/AffineMapUtils.h
```diff
@@ -0,0 +1,135 @@
...
+#include "ttmlir/Asserts.h"
+#include "ttmlir/Utils.h"
+#include "llvm/ADT/SmallVector.h"
+
+#include <cstdint>
+
+namespace ttmlir::utils {
+
+template <typename Vector>
+llvm::SmallVector<int64_t> evalShape(mlir::AffineMap map, Vector shape) {
```
bgrady-tt, line 22: Note: I moved this util function here, but I don't think it is correct. 

Counterexample:
`(d0) -> ((8 - d0) % 4)` where bounds of `d0` are 0 to 7

Sampling the extremes (0 and 7), the max values we would get are:
(8 - (0)) % 4 = 0
(8 - (7)) % 4 = 1

However the actual input values that maximize the result are `1` and `5`. So the entire grid needs to be sampled in the general case (see `applyMapToGrid`)

include/ttmlir/AffineMapUtils.h
```diff
@@ -0,0 +1,135 @@
...
+  for (unsigned i = 0; i < map.getNumResults(); i++) {
+    results.push_back(builder.create<mlir::affine::AffineApplyOp>(
+        loc, affineMapSelectOneOutput(map, i), inputs));
+  }
+  return results;
+}
+
+// Derive a grid shape by sampling an affine map over a reference grid shape.
+inline llvm::SmallVector<int64_t>
+applyMapToGrid(mlir::ArrayRef<int64_t> gridShape, mlir::AffineMap map) {
```
bgrady-tt, line 117: See https://github.com/tenstorrent/tt-mlir/pull/5437/files#r2452361871 for why `evalShape` isn't used instead.

include/ttmlir/Utils.h
```diff
@@ -87,66 +89,6 @@ inline void sample(const Vector &shape, Fn fn) {
   }
 }
 
-template <typename Vector>
```
bgrady-tt, line 90: code moved to AffineMapUtils.h

lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
```diff
@@ -50,6 +52,67 @@ class D2MNamedRewriterCommon {
...
+        aspectRatio = ratio;
+        maxDimIndex = i;
+      }
+    }
+    return {maxDimIndex, aspectRatio};
+  }
+
+  bool shouldImplementAsVirtualGrid(ArrayRef<int64_t> physicalShape) const {
+    auto [maxRatioIndex, aspectRatio] = findMaxDimAndAspectRatio(physicalShape);
+    constexpr double HIGH_ASPECT_RATIO_THRESHOLD = 8.0;
```
bgrady-tt: This is a simple heuristic; open to ideas about what to do here.

lib/Dialect/D2M/Transforms/LowerToLayout.cpp
```diff
@@ -20,14 +24,15 @@ namespace mlir::tt::d2m {
 // MetalLayoutAttr as optional gracefully.
 namespace {
 struct TensorInfo {
+  Value val;
```
bgrady-tt: NB: The actual tensor `Value` is needed here to derive physical grid shape

include/ttmlir/Dialect/D2M/Utils/Utils.h
```diff
@@ -42,6 +44,25 @@ Type getRegionLargestDstElemType(Region &region);
 AffineMap concatInversePermutationMap(SmallVector<AffineMap> affineMaps,
                                       bool reverse);
 
+// Get grid shape of a tensor or memref using its DeviceLayoutInterface.
+llvm::SmallVector<int64_t> getGridShape(mlir::Value tensorOrMemref);
```
nsmithtt: This should live in `include/ttmlir/Dialect/TTCore/IR/Utils.h`, next to `getDeviceLayout`.

include/ttmlir/Dialect/D2M/Utils/Utils.h
```diff
@@ -42,6 +44,25 @@ Type getRegionLargestDstElemType(Region &region);
...
+
+// Returns true if the tensor/memref has a virtual grid (i.e. non trivial core
+// virtualization map defined in underlying ShardLayoutAttr).
+bool hasVirtualGrid(mlir::Value tensorOrMemref);
```
nsmithtt: Perhaps this is required, I just often find that code is more readable when it doesn't have conditions, i.e. users of the virtual grid should be robust to there being an identity grid here so potentially no need to conditionalize this.

include/ttmlir/Dialect/D2M/Utils/Utils.h
nsmithtt: I think everything below should be a ttcore util since the metal layout and device interface live in ttcore, not d2m.

include/ttmlir/Dialect/D2M/Utils/VirtualGrid.h
```diff
@@ -0,0 +1,80 @@
...
+  ///
+  /// The physical memref/tensor uses the inverse map to perform core
+  /// virtualization, translating raw physical core locations at runtime into
+  /// virtual core locations that are compatible with virtual space. The inverse
+  /// map is restricted to only the grid dimensions; shard dims CANNOT
+  /// participate in virtual grid dim exprs (and vice-versa) or reblocking will
+  /// not work reliably.
+  createCoreVirtMaps(MLIRContext *context,
+                     const SmallVector<int64_t> &virtualGrid,
+                     const SmallVector<int64_t> &targetGrid) {
```
nsmithtt: I feel like this whole routine should be made part of `metal_layout` since it's an intrinsic concept to layout.

lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
nsmithtt: Seems reasonable to me, fyi @vwellsTT who's introducing a grid selection pass, where this should probably be moved to.

lib/Dialect/D2M/Transforms/GenericLowerDMAs.cpp
```diff
@@ -46,6 +48,27 @@ class D2MGenericLowerAffineDMAsRewritePattern : public OpRewritePattern<DMAOp> {
...
+    SmallVector<Value> virtualGridIndices;
+    if (coreVirtualizationMap) {
+      virtualGridIndices = ttmlir::utils::fullyApplyAffineMap(
+          builder, loc, coreVirtualizationMap, physicalCoreIndices);
+    } else {
+      virtualGridIndices = physicalCoreIndices;
+    }
+    TT_assertv(virtualGridIndices.size() == gridIndexingMap.getNumResults(),
+               "Core virtualization map must have the same number of results "
+               "as the grid indexing map");
```
bgrady-tt: The above code implements core virtualization for the purpose of stream indexing, by applying the `coreVirtualizationMap` associated with the GenericOp's output operand to the raw physical CoreIndex. The stream index then generates final virtual indices applying the block factors (`virtDim * blockFactor + iterIndex`).

lib/Dialect/D2M/IR/D2MOps.cpp
```diff
@@ -800,7 +821,8 @@ void d2m::GenericOp::build(mlir::OpBuilder &builder,
     TT_assertv(
         layout,
         "This generic constructor expects operands to be in device layout");
-    grid = builder.getAttr<ttcore::GridAttr>(layout.getGridShape(shapedType));
+    grid = builder.getAttr<ttcore::GridAttr>(
```
nsmithtt: It kind of seems like the generic op should hold the virtual grid shape, at least that will be much easier to read.  The grid attr already has an affine map to express the remapping.

lib/Dialect/D2M/Transforms/LowerToLayout.cpp
```diff
@@ -329,18 +352,78 @@ class D2MSplitCompoundLayoutRewriter : public OpRewritePattern<ToLayoutOp> {
...
+    if (virtualGridShape[0] > deviceGridShape[0]) {
+      int64_t divisor = std::gcd(virtualGridShape[0], deviceGridShape[0]);
+      ret = {divisor, virtualGridShape[1]};
+    } else {
+      int64_t divisor = std::gcd(virtualGridShape[1], deviceGridShape[1]);
+      ret = {virtualGridShape[0], divisor};
+    }
+    assert(ret.size());
+    return ret;
+  }
```
bgrady-tt: This code must be updated to handle ND sharding, but we currently don't support that. In some cases it would be better to bounce off a MxN block sharded grid rather than Nx1 or 1xN to minimize allocator requirements, but this doesn't like it will limit testing/development for now.

lib/Dialect/D2M/Transforms/LowerToLayout.cpp
```diff
@@ -329,18 +352,78 @@ class D2MSplitCompoundLayoutRewriter : public OpRewritePattern<ToLayoutOp> {
...
+      emptyShape = shardedShape;
+
+      // Reset index affine map from layout, preserving everything else.
+      auto map = rewriter.getMultiDimIdentityMap(emptyShape.size());
+      layout = ttcore::MetalLayoutAttr::get(
+          rewriter.getContext(), layout.getLogicalShape(),
+          layout.getDimAlignments(), layout.getCollapsedIntervals(),
+          layout.getOobVal(), layout.getMemorySpace(), layout.getMemoryLayout(),
+          map);
+    }
```
bgrady-tt: Virtual grids cannot use the grid shape of the desired output can be used for the init operand (the "empty"), as the metal host cannot write/read from a virtual gridded tensor. Instead, we must generate a _regular block sharded grid_ shape for the temporary bounce tensor.

include/ttmlir/Dialect/D2M/Utils/VirtualGrid.h
```diff
@@ -0,0 +1,80 @@
...
+#include "ttmlir/Utils.h"
+#include "llvm/ADT/SmallVector.h"
+
+namespace ttmlir::d2m {
+using namespace llvm;
+using namespace mlir;
+
+class VirtualGridUtil {
+public:
+  static std::pair<AffineMap, AffineMap>
```
bgrady-tt: This is the primary function that would need to change to support new kinds of virtual grid collapsing (i.e. for ND sharding).

lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
```diff
@@ -155,6 +218,56 @@ class D2MNamedRewriterCommon {
     return mlir::RankedTensorType::get(shardedShape, elementType, metalLayout);
   }
 
+  Value createVirtualGridToLayoutOp(
```
bgrady-tt: Virtual grid tensors are implemented as pairs of an underlying physical tensor (that can be allocated) and a view that has the desired virtual grid shape. To support this, this function generates a `toLayout` op with a _viewed_ init operand as its 'destination', instead of trying to use a empty physical tensor with an illegal grid shape.

```cpp
%init_phys = ttir.empty()
%init_view = d2m.view_layout %underlying
%out = to_layout %input %init_view
```

test/ttmlir/Dialect/D2M/generic/virtual_grids.mlir
```diff
@@ -0,0 +1,53 @@
...
+!streamT   = memref<1x64x4x4x!ttcore.tile<32x32, f32>, #ttcore.view<(d0,d1,d2,d3) -> (d0,d1,d2,d3)>, #ttcore.memory_space<l1>>
+!cbT2 = memref<4x4x!ttcore.tile<32x32, f32>, #ttcore.memory_space<l1>>
+
+// CHECK: func.func @virtual_grid_blocked
+func.func @virtual_grid_blocked(%arg0 : !virtGridT) -> () {
+
+  // input
+  %In = memref.alloc() : !virtGridT
+  %Storage = memref.alloc() : !virtGridT
+  %StreamIn = "d2m.stream_layout"(%In,%Storage) : (!virtGridT,!virtGridT) -> !streamT
```
vroubtsovTT, line 36: 1. `%In` and `%Storage` look to be two SSA values of the same type and have no other inputs... it is confusing what role might be played by `%Storage`, since that value doesn't inject any new information into `%StreamIt`?
 
2. how does one determine the allocation size of `%Storage` and/or `!cbT2`, is it still the shard shape size of ... `!streamT`?

test/ttmlir/Dialect/D2M/generic/virtual_grids.mlir
```diff
@@ -0,0 +1,53 @@
...
+// ShardLayoutAttr has a new optional param coreVirtualizationMap. This behaves
+// like a built-in view applied *after* any views into this memref, but _before_
+// the shard layout and device maps. A non-empty core virt map indicates that
+// the grid of the memref is virtual.
+#virtShardLayoutAttr = #ttcore.shard<16384x4096, 1, (d0, d1) -> (d1 floordiv 8, d1 mod 8)>
+
+!virtGridT = memref<1x64x4x4x!ttcore.tile<32x32, f32>, #virtShardLayoutAttr, #ttcore.memory_space<l1>>
+
+// NOTE: streaming a virtual grid tensor DOES NOT fold/absorb its affine map
+!streamT   = memref<1x64x4x4x!ttcore.tile<32x32, f32>, #ttcore.view<(d0,d1,d2,d3) -> (d0,d1,d2,d3)>, #ttcore.memory_space<l1>>
```
vroubtsovTT: "DOES NOT fold/absorb" is not clear, please elaborate... for example, how is an expression like `%StreamIt[i0, i1]` interpreted?

include/ttmlir/AffineMapUtils.h
```diff
@@ -0,0 +1,130 @@
...
+  mlir::SmallVector<int64_t> lastIndex;
+  for (auto dim : shape) {
+    lastIndex.push_back(dim - 1);
+  }
+
+  auto result = map.compose(lastIndex);
+  for (auto &dim : result) {
+    dim += 1;
+  }
+  return result;
```
vroubtsovTT, line 32: Are we opening a can of worms here? Unless we restrict the class of affine maps that is allowed for `map` (e.g. those for which  `AffineMap::isProjectedPermutation()` is `true`, which I believe is currently true only of the `indexing_maps` of `GenericOp`), there is a large (infinite?) number of valid mappings for which the output shape is not computable by simply evaluating `map` on the input dim extents - 1.

include/ttmlir/Dialect/D2M/Utils/VirtualGrid.h
```diff
@@ -0,0 +1,85 @@
...
+#include "mlir/IR/BuiltinTypes.h"
+#include "mlir/Transforms/DialectConversion.h"
+#include "ttmlir/Asserts.h"
+#include "ttmlir/Utils.h"
+#include "llvm/ADT/SmallVector.h"
+
+namespace ttmlir::d2m {
+using namespace llvm;
+using namespace mlir;
+
```
vroubtsovTT: since injecting entire namespaces in header files is somewhat fraught and since we haven't done this elsewhere, I believe, it is much preferred to not do `using namespace llvm/mlir` in `.h`s and instead use such short ns prefixes as needed on method calls.

include/ttmlir/AffineMapUtils.h
```diff
@@ -0,0 +1,130 @@
...
+}
+
+/// Returns a new affine map with all symbols replaced with given constant
+/// values.
+inline mlir::AffineMap
+replaceAffineMapSymbols(mlir::AffineMap map, mlir::ArrayRef<int64_t> symbols) {
+  assert(map.getNumSymbols() == symbols.size());
+
```
vroubtsovTT: since you've already started using TT_assert, probably should continue here

include/ttmlir/Dialect/D2M/Utils/VirtualGrid.h
nsmithtt: +1, no `using namespace` in a header.

include/ttmlir/Dialect/D2M/Utils/VirtualGrid.h
```diff
@@ -0,0 +1,85 @@
...
+class VirtualGridUtil {
```
nsmithtt: Maybe this should be a namespace instead of a class? `mlir::tt::d2m::utils::virtual_grid`

include/ttmlir/Dialect/TTCore/IR/TTCoreOpsTypes.td
```diff
@@ -32,19 +32,26 @@ def TTCore_GridAttr : TTCore_Attr<"Grid", "grid"> {
     TT grid attribute
   }];
 
-  let parameters = (ins ArrayRefParameter<"int64_t">:$shape,
+  let parameters = (ins
+                        ArrayRefParameter<"int64_t">:$shape,
+                        ArrayRefParameter<"int64_t">:$physShape,
```
nsmithtt: I'm hesitant to introduce this attribute since it isn't going to work long term, i.e. `1x63` case.  I think we should just have a routine on this attribute that derives it given the shape.

lib/Dialect/D2M/IR/D2MOps.cpp
```diff
@@ -788,19 +799,46 @@ mlir::OpFoldResult d2m::ViewLayoutOp::fold(FoldAdaptor adaptor) {
 void d2m::GenericOp::build(mlir::OpBuilder &builder,
                            mlir::OperationState &state, ValueRange inputs,
                            ValueRange outputs, ArrayAttr indexingMaps,
-                           ArrayAttr iteratorTypes, ThreadType singleThreadType,
-                           ttcore::GridAttr grid,
+                           ArrayAttr iteratorTypes,
+                           ArrayRef<int64_t> targetGridShape,
```
nsmithtt: Per my comment above on the `GridAttr`, I think for now we should be deriving the physical grid shape given a virtual grid shape so that we don't need to pass in a bunch of extra information.  For now, since we limiting things to just height/width sharding it should be trivial to infer the physical extents.

test/ttmlir/Dialect/D2M/generic/virtual_grids.mlir
vroubtsovTT: I am guessing that the intent is to say that the affine map in the return type does not compose with the affine map of the stream's input. I think this is how we have it working right now, but I don't understand the design intent. A user of the return value must examine the stream's input to know the fully composed mapping... why?

test/ttmlir/Dialect/D2M/generic/virtual_grids.mlir
vroubtsovTT: This example would be useful to the allocator work if it showed a case of storage buffer size being different from the input size.

include/ttmlir/Dialect/D2M/Utils/VirtualGrid.h
vroubtsovTT: Thoughts:

- using a static class API vs a namespace can have advantages, e.g. being able to use `private` access levels. Hiding impl details is a little bit harder in a namespace, the usual `details` trick is somewhat iffy.
- if this API is foreseen to only ever need (mostly) public static methods it can be converted to a namespace; if not, the class should be explicitly shown as non-copyable/non-moveable by deleting the appropriate constructor/assignment operators otherwise I would be confused as the end user. The overhead of doing so (we can't use `boost::noncopyable` and I don't see [P2895](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2023/p2895r0.html) on the adoption timeline) might be a sign that the namespace option is a better choice here.

include/ttmlir/AffineMapUtils.h
bgrady-tt: @vroubtsovTT Agree; see my [comments](https://github.com/tenstorrent/tt-mlir/pull/5437/files/c7df8f6f31ebec8de76a2cae0c3980b01aa701f8#r2452361871) above. evalShape existed before this PR; I just moved it here to organize the sprawling helper functions related to affine maps.

I think `evalShape` is _incorrect_ in the general case, so I also wrote `applyMapToGrid` which uses a brute-force sampling approach to check extents.

lib/Dialect/D2M/IR/D2MOps.cpp
bgrady-tt: Regarding inferring the physical shape from (virtual_grid_shape, physicalToVirtualMap), I can do that and remove the physical shape. For now, the code will still verify that the inferred physical grid shape _must be a rectangular grid_. 

I think _non-rectangular_ physical grids are supportable, but more out-of-scope work is needed to make sure core-ranges are generated properly, etc.

include/ttmlir/Dialect/TTCore/IR/TTCoreOpsTypes.td
bgrady-tt: See comment below; `physShape` will be removed.

include/ttmlir/Dialect/D2M/Utils/VirtualGrid.h
bgrady-tt: A namespace here would work; I was on the fence about which way to go. It really doesn't matter much I think.

test/ttmlir/Dialect/D2M/generic/virtual_grids.mlir
bgrady-tt: @vroubtsovTT the meaning here is quite vague. I will update the wording.

**With views,** a stream layout _could_ be folded with a view input, such that the affine maps of both are composed together:

tensor -> view<#someMap> -> stream<#identityMap>
collapsed to
tensor -> stream<#someMap>

In contrast if a stream layout wraps a **virtual grid tensor,** canonicalization should **_not_** attempt to fold in the affine map from the `ShardLayoutAttr` into the stream layout affine map.

test/ttmlir/Dialect/D2M/generic/virtual_grids.mlir
bgrady-tt: This is the same as how stream layout operands work now; the stream storage buffer must match the apparent (viewed) shape of the stream input. I think we discussed this convention before.

To have a stream buffer storage that is smaller than the input tensor requires _reblocking the input via a view to reduce the input shape's apparent shard size_. 

This is true for virtual grid tensors and 'conventional' grid tensors, and the principles are the same. Its still subdividing the shard dims to create a larger 'blocked' grid; the initial unblocked grid shapes are just less constrained.

```cpp
original_input : memref<1x64x1x16xElemType>

// note: shard size is now 1x1 vs original at 4x4
reblocked_viewed_input : memref<1x1024x1x1xElemType>

// storage matches reblocked_viewed_input, and is 16x smaller allocation than original input tensor
storage : memref<1x1024x1x1xElemType>
```

vroubtsovTT: > To have a stream buffer storage that is smaller than the input tensor requires reblocking the input via a view to reduce the input shape's apparent shard size.

That's what I referred to as being a more useful example. We should view buffer size being different from the input's shard size as the _default_ use case and the case of preserving the shard shape as an edge case. My latest allocator PR added some reblocking logic in order to "practice" all these conventions and having a known good IR example would have been helpful at the time.

vroubtsovTT:
I'll mention another point that seems to be missing from our docs/code that deals with grid/shard shaping: the case when a generic op's _element type_ changes from inputs to outputs. 

Here's an example of IR we currently emit:
```
   d2m.generic {block_factors = [1, 1], grid = #ttcore.grid<1x1>, indexing_maps = [affine_map<(d0, d1) -> (d0, d1)>, affine_map<(d0, d1) -> (d0, d1)>], iterator_types = [#ttcore.iterator_type<parallel>, #ttcore.iterator_type<parallel>], threads = [#d2m.thread<compute>]}
        ins(%alloc_0 : memref<1x1x384x384xf32, #ttcore.shard<1536x4>, #ttcore.memory_space<l1>>)
        outs(%alloc : memref<1x1x12x12x!ttcore.tile<32x32, f32>, #ttcore.shard<49152x4096>, #ttcore.memory_space<l1>>)  {
    ^compute0(%cb0: !d2m.cb<memref<384x384xf32, #ttcore.memory_space<l1>>>, %cb1: !d2m.cb<memref<12x12x!ttcore.tile<32x32, f32>, #ttcore.memory_space<l1>>>):
      %0 = d2m.wait %cb0 : <memref<384x384xf32, #ttcore.memory_space<l1>>> -> memref<384x384xf32, #ttcore.memory_space<l1>>
      %1 = d2m.reserve %cb1 : <memref<12x12x!ttcore.tile<32x32, f32>, #ttcore.memory_space<l1>>> -> memref<12x12x!ttcore.tile<32x32, f32>, #ttcore.memory_space<l1>>
      "d2m.tile_tilize_block"(%0, %1) : (memref<384x384xf32, #ttcore.memory_space<l1>>, memref<12x12x!ttcore.tile<32x32, f32>, #ttcore.memory_space<l1>>) -> ()
    }
```
The element type here changes from `f32` to `ttcore.tile<32x32, f32>`. Tiling is in fact another level of "blocking" the calculation and it is not clear how our [full] blocking factor/reblocking utils support this.

lib/Dialect/TTCore/IR/TTCoreOpsTypes.cpp
```diff
@@ -794,6 +799,67 @@ static llvm::SmallVector<int64_t> applyCollapsedIntervalsAndAlignments(
   return resultShape;
 }
 
+SmallVector<int64_t>
+GridAttr::getPhysicalGridShape(ArrayRef<int64_t> deviceGridShape) {
```
bgrady-tt, line 805: @nsmithtt This is the function that does trial-and-error checks to infer a valid physical grid given a virtual grid, device grid, and virtual->physical affine map. This removes the need to store the physical grid shape in ttcore.grid attr; it can be computed on the fly.

lib/Dialect/D2M/IR/D2MOps.cpp
```diff
@@ -1101,6 +1118,84 @@ ::mlir::LogicalResult d2m::GenericOp::verify() {
     }
   }
 
+  if (!getGrid().getMapping().isEmpty()) {
```
bgrady-tt, line 1122: The following are consistency checks that make sure that a GenericOp with a non-empty virtual grid mapping in its ttcore.grid matches its output tensor

include/ttmlir/AffineMapUtils.h
nsmithtt: I don't know how to enforce, but there are probably certain affine constructions that we should deem illegal.  For example, negating an iterator seems like an illegal thing to do.

include/ttmlir/Dialect/D2M/Utils/VirtualGrid.h
nsmithtt: Somewhat prefer this being a namespace since there isn't a useful reason for it to be a class, but won't block the PR on it.

lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
```diff
@@ -41,9 +44,19 @@ class D2MNamedRewriterCommon {
...
+        ttnnMode(ttnnMode), collapseTensors(collapseTensors),
+        targetGridShape(targetGridShape),
+        targetSquareGridShape(
+            d2m::utils::getSquareTargetGrid(targetGridShape)) {
+    assert(!targetGridShape.empty());
+  }
+
+  ArrayRef<int64_t> getTargetSquareGridShape() const {
+    return ArrayRef<int64_t>(targetSquareGridShape);
+  }
```
nsmithtt: Is this a bad rebase? Do we still need the TTIRToD2M pass to change now that grid selection happens in GridSelection.cpp.

lib/Dialect/D2M/IR/D2MOps.cpp
nsmithtt: Bumping this thread, still would prefer not to have to pass the physical grid into the builder here.

lib/Dialect/D2M/IR/D2MOps.cpp
```diff
@@ -1097,6 +1118,84 @@ ::mlir::LogicalResult d2m::GenericOp::verify() {
...
+
+    if (getGrid().getMapping().getNumInputs() != 2ul) {
```
nsmithtt, line 1124: Is there a plan for higher rank support? We've adding a few changes to local branches for the folks working on TM's that require higher rank support to be implemented.

lib/Dialect/D2M/IR/D2MOps.cpp
```diff
@@ -1097,6 +1118,84 @@ ::mlir::LogicalResult d2m::GenericOp::verify() {
...
+    }
+
+    // Generic op with defined physical->virtual mapping in its grid attr should
+    // have output operand(s) with a virtual->physical mapping defined in the
+    // layout attr.
+    for (auto output : getOutputs()) {
+      mlir::ShapedType outputType =
+          mlir::cast<mlir::ShapedType>(output.getType());
+
+      AffineMap fwdMap;
```
nsmithtt: Perhaps we should make the following if conditions part of the DeviceLayoutInterface, then you could just write:
```c++
AffineMap fwdMap = getDeviceLayout(outputType).getCoreVirtualizationMap(/*gridPortion=*/true);
```

lib/Dialect/D2M/Transforms/GenericLowerDMAs.cpp
```diff
@@ -35,7 +37,7 @@ class D2MGenericLowerAffineDMAsRewritePattern : public OpRewritePattern<DMAOp> {
   buildStreamIndex(OpBuilder &builder, Location loc,
                    ArrayRef<int64_t> gridShape, ArrayRef<int64_t> blockFactors,
                    ArrayRef<int64_t> shardShape, AffineMap dmaIndexingMap,
-                   AffineMap gridIndexingMap) {
+                   AffineMap gridIndexingMap, AffineMap coreVirtualizationMap) {
```
nsmithtt: The argument names to this function are starting to get confusing, but I don't have a great suggestion.  Maybe we should rename gridIndexingMap to outputOperandIndexingMap?

lib/Dialect/TTCore/IR/TTCoreOpsTypes.cpp
nsmithtt: Wow, this is more complicated than I thought, can we not just apply the virtual shape to the affine map?  Is this because we're storing the inverse affine map?

lib/Dialect/D2M/Utils/Utils.cpp
```diff
@@ -128,4 +132,48 @@ AffineMap concatInversePermutationMap(mlir::ArrayRef<AffineMap> affineMaps,
...
+ArrayRef<int64_t> getGridShape(Value tensorOrMemref) {
+  TT_assertv((mlir::isa<RankedTensorType>(tensorOrMemref.getType()) ||
+              mlir::isa<MemRefType>(tensorOrMemref.getType())),
+             "Expected a tensor or memref type");
+
+  Attribute layout =
+      (mlir::isa<RankedTensorType>(tensorOrMemref.getType())
+           ? mlir::dyn_cast<RankedTensorType>(tensorOrMemref.getType())
+                 .getEncoding()
+           : mlir::dyn_cast<MemRefType>(tensorOrMemref.getType()).getLayout());
```
nsmithtt: See `include/ttmlir/Dialect/TTCore/IR/Utils.h`, there is `getDeviceLayout(Value value)`.  This could be:
```c++
ArrayRef<int64_t> getGridShape(Value tensorOrMemref) {
  return getDeviceLayout(tensorOrMemref).getGridShape(cast<ShapedType>(tensorOrMemref.getType()));
}
```

lib/Dialect/D2M/IR/D2MOps.cpp
bgrady-tt: I think I missed that there is a straightforward way to derive the physical grid from the virtual grid and core virtualization map. That removes the need to have the target grid passed into the build() here.

lib/Dialect/D2M/IR/D2MOps.cpp
bgrady-tt: Following merging this PR, I plan to immediately start working on higher rank support in a follow on PR. Based on pressure to merge this PR as soon as possible, I want to avoid inflating scope beyond 2D HW sharding support.

lib/Dialect/D2M/IR/D2MOps.cpp
bgrady-tt: Agree, there are a few sections of code that require this awkward logic. I can create a new accessor on DeviceLayoutInterface

test/ttmlir/Silicon/TTMetal/n150/simple_transpose.mlir
```diff
@@ -1,4 +1,4 @@
-// RUN: ttmlir-opt --ttir-to-ttmetal-pipeline="system-desc-path=%system_desc_path%" -o %t.mlir %s
+// RUN: ttmlir-opt --ttir-to-ttmetal-pipeline="system-desc-path=%system_desc_path%" --mlir-print-ir-after-all -o %t.mlir %s
```
anuragsingh-tt: Needed?

lib/Dialect/TTCore/IR/TTCoreOpsTypes.cpp
bgrady-tt: The grid attribute stores the _inverse_ function (physical -> virtual); this mapping is used to 'virtualize' the physical grid coordinates. The GridAttr shape is also the virtual shape now.

So we have to solve for the physical shape the satisfies `invMap(physicalShape) = virtual`. This is why having the physical grid shape computed and stored was helpful; it is non-trivial to derive it later.

include/ttmlir/AffineMapUtils.h
vroubtsovTT: I propose to add another utility method `isSemiAffine(AffineMap)` or `isPureAffine(AffineMap)` and use it inside a TT_assert/TT_debug input guard inside `evalShape()`: https://github.com/tenstorrent/tt-mlir/issues/5869

test/ttmlir/Dialect/D2M/generic/virtual_grids.mlir
vroubtsovTT: Ok, I am resolving this as "fold" referring purely to how MLIR _folders_ would work with these.

include/ttmlir/Dialect/D2M/Utils/VirtualGrid.h
```diff
@@ -0,0 +1,92 @@
...
+#ifndef TTMLIR_DIALECT_D2M_UTILS_VIRTUALGRID_H
+#define TTMLIR_DIALECT_D2M_UTILS_VIRTUALGRID_H
+
+#include "mlir/IR/AffineExpr.h"
+#include "mlir/IR/AffineMap.h"
+#include "mlir/IR/BuiltinTypes.h"
+#include "mlir/Transforms/DialectConversion.h"
+#include "ttmlir/Asserts.h"
+#include "ttmlir/Utils.h"
+#include "llvm/ADT/SmallVector.h"
```
vroubtsovTT, line 14: FYI, #include ordering here does not follow our convention (our project includes should be before llvm/mlir.

include/ttmlir/Dialect/D2M/Utils/VirtualGrid.h
```diff
@@ -0,0 +1,92 @@
...
+      auto forward =
+          mlir::AffineMap::get(2 * rank, 0, forwardMapExprs, context);
+      auto inverse = mlir::AffineMap::get(rank, 0, inverseMapExprs, context);
+      return {forward, inverse};
+    } else {
+      TT_assertv((is2DWidthSharded || is2DHeightSharded),
+                 "Only supporting 2D width or height sharding (actual grid "
+                 "shape = {1})",
+                 ttmlir::utils::formatIterable(virtualGrid, "x"));
+      llvm_unreachable("Only supporting 2D width or height sharding");
```
vroubtsovTT: using `llvm_unreachable` after TT_assert does not seem to make sense to me... If you are going to error out via `llvm_unreachable()` then the assert condition is known to be `false`. If you then use `TT_assertv(false,...)` then the `llvm_unreachable` itself is not going to be reached ever.

Perhaps you wanted to be able to do some formatting of the error message? In which case this should work without a following `llvm_unreachable` (also note that using position-less `{}` placeholders is usually enough):
```
TT_assertv(false,
                 "Only supporting 2D width or height sharding (actual grid "
                 "shape = {})",
                 ttmlir::utils::formatIterable(virtualGrid, "x"));
```

lib/Dialect/TTCore/IR/TTCoreOpsTypes.cpp
nsmithtt: Yeah OK I see.  Maybe we've gone back and forth on this too many times now lol, so I'm happy to leave it as is if this code works.  But maybe it's worth considering having 2 maps on the grid attr:
- physicalToVirtualMap
- virtualToPhysicalMap

include/ttmlir/Dialect/D2M/Utils/VirtualGrid.h
vroubtsovTT: Right now, there is a single util method in this class, which feels like slight overkill. But if you anticipate more util methods in the future but everything will remain stateless and does not need inheritance it seems fine to go with a namespace. I would generalize it to something like `mlir::tt::d2m::utils::grids` then and maybe we'll use this as home for more/most grid-related calc utilities.

test/python/golden/test_metal_virtual_grids.py
```diff
@@ -0,0 +1,80 @@
+# SPDX-FileCopyrightText: (c) 2024 Tenstorrent AI ULC
+#
```
vroubtsovTT: 2025

lib/Dialect/D2M/Transforms/LowerToLayout.cpp
```diff
@@ -258,6 +268,11 @@ class D2MLowerToLayoutRewriter : public OpRewritePattern<ToLayoutOp> {
 
     llvm_unreachable("Unknown compound component");
   }
+
+  ArrayRef<int64_t> getTargetGridShape() const { return targetGridShape; }
+
+private:
+  const llvm::SmallVector<int64_t> targetGridShape;
```
vmilosevic: :warning: **cppcoreguidelines\-avoid\-const\-or\-ref\-data\-members** :warning:
member `` targetGridShape `` of type `` const llvm::SmallVector<int64_t> `` \(aka `` const SmallVector<long> ``\) is const qualified

lib/Dialect/D2M/Transforms/LowerToLayout.cpp
```diff
@@ -272,23 +287,52 @@ class D2MSplitCompoundLayoutRewriter : public OpRewritePattern<ToLayoutOp> {
   public:
     explicit BounceTypeBuilder(MLIRContext *ctx) : ctx(ctx) {}
 
+    // Computes a workable bounce shape grid for a virtual grid.
+    llvm::SmallVector<int64_t>
+    computeVirtualGridBounceShape(ArrayRef<int64_t> virtualGridShape,
+                                  ArrayRef<int64_t> deviceGridShape) const {
+
+      // TODO: Generalize to N dimensions.
```
missing username/bug in TODO
```suggestion
      // TODO(nsmith): Generalize to N dimensions.
```

lib/Dialect/TTNN/IR/TTNNOpsAttrs.cpp
```diff
@@ -7,6 +7,7 @@
 #include "ttmlir/Dialect/TTCore/Utils/CoreRangeSet.h"
 #include "ttmlir/Dialect/TTNN/Utils/Conv2dConfigParams.h"
 #include "ttmlir/Dialect/TTNN/Utils/Utils.h"
+#include "ttmlir/AffineMapUtils.h"
```
vmilosevic: :warning: **llvm\-include\-order** :warning:
```suggestion

```

include/ttmlir/Dialect/D2M/Utils/VirtualGrid.h
```diff
@@ -0,0 +1,92 @@
...
+        forwardMapExprs = {d0 % gridRowStride, d0.floorDiv(gridRowStride), d2,
+                           d3};
+        inverseMapExprs = {zero, d1 * gridColStride + d0, zero};
+      }
+      auto forward =
+          mlir::AffineMap::get(2 * rank, 0, forwardMapExprs, context);
+      auto inverse = mlir::AffineMap::get(rank, 0, inverseMapExprs, context);
+      return {forward, inverse};
+    } else {
+      TT_assertv((is2DWidthSharded || is2DHeightSharded),
```
vmilosevic: :warning: **llvm\-else\-after\-return** :warning:
```suggestion
    }       TT_assertv((is2DWidthSharded || is2DHeightSharded),
```

include/ttmlir/Dialect/D2M/Utils/VirtualGrid.h
```diff
@@ -0,0 +1,92 @@
...
+                 "Only supporting 2D width or height sharding (actual grid "
+                 "shape = {1})",
+                 ttmlir::utils::formatIterable(virtualGrid, "x"));
+      llvm_unreachable("Only supporting 2D width or height sharding");
+    }
```
vmilosevic: :warning: **llvm\-else\-after\-return** :warning:
```suggestion
   
```

lib/Dialect/D2M/Transforms/GenericLinearizeMemref.cpp
```diff
@@ -4,6 +4,7 @@
 
 #include "ttmlir/Dialect/D2M/Transforms/Passes.h"
 #include "ttmlir/Utils.h"
```
vmilosevic, line 7: :warning: **llvm\-include\-order** :warning:
```suggestion
#include "ttmlir/AffineMapUtils.h"
#include "ttmlir/Utils.h"
```

lib/Dialect/D2M/Transforms/GenericLinearizeMemref.cpp
```diff
@@ -4,6 +4,7 @@
...
+#include "ttmlir/AffineMapUtils.h"
```
vmilosevic: :warning: **llvm\-include\-order** :warning:
```suggestion

```

lib/Dialect/D2M/Utils/Utils.cpp
```diff
@@ -128,4 +132,48 @@ AffineMap concatInversePermutationMap(mlir::ArrayRef<AffineMap> affineMaps,
...
+
+  if (auto deviceLayout =
+          mlir::dyn_cast_if_present<ttcore::DeviceLayoutInterface>(layout)) {
+    return deviceLayout.getGridShape(
+        mlir::dyn_cast<ShapedType>(tensorOrMemref.getType()));
+  } else {
+    return {};
+  }
```
vmilosevic: :warning: **llvm\-else\-after\-return** :warning:
```suggestion
  }     return {};
 
```

vmilosevic (commented):

include/ttmlir/Dialect/D2M/Utils/VirtualGrid.h
bgrady-tt: Fixed; I assert first that we are height/width sharding compatible and then eliminate the branch.

test/ttmlir/Silicon/TTMetal/n150/simple_transpose.mlir
bgrady-tt: good catch! I will remove that

lib/Dialect/D2M/Utils/Utils.cpp
bgrady-tt: Fixed.

lib/Dialect/D2M/Transforms/GenericLinearizeMemref.cpp
bgrady-tt: Fixed.

lib/Dialect/D2M/Transforms/GenericLinearizeMemref.cpp
bgrady-tt: Fixed.

include/ttmlir/Dialect/D2M/Utils/VirtualGrid.h
bgrady-tt: Fixed.

include/ttmlir/Dialect/D2M/Utils/VirtualGrid.h
bgrady-tt: Fixed.

lib/Dialect/TTNN/IR/TTNNOpsAttrs.cpp
bgrady-tt: Fixed.

lib/Dialect/D2M/Transforms/LowerToLayout.cpp
bgrady-tt: Fixed.

lib/Dialect/D2M/Utils/Utils.cpp
bgrady-tt: Fixed!

lib/Dialect/D2M/Transforms/GenericLowerDMAs.cpp
bgrady-tt: outputOperandIndexingMap is probably less confusing, so I changed it to that.

lib/Dialect/D2M/IR/D2MOps.cpp
bgrady-tt: Added a new member function on DeviceLayoutInterface `getVirtualizationMapIfExists()` that allows querying the underlying virtualization map for tensor/memref.

lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
bgrady-tt: This was been completely fixed in latest; GenericOp build() no longer requires a target grid shape argument, so all of this grid shape querying is eliminated.

lib/Dialect/D2M/IR/D2MOps.cpp
bgrady-tt: This has been fixed in the latest version; physical grid arg is not required for GenericOp builder anymore.

include/ttmlir/Dialect/TTCore/IR/TTCoreOpsTypes.td
bgrady-tt: physShape parameter in GridAttr has been removed from latest.

include/ttmlir/Dialect/D2M/Utils/VirtualGrid.h
bgrady-tt: This class has been converted to a namespace.

include/ttmlir/AffineMapUtils.h
bgrady-tt: If true, `evalShape` should check this assumption and assert. Otherwise it is incorrect for known cases when a correct (if slower) method exists.

include/ttmlir/Dialect/D2M/Utils/Utils.h
bgrady-tt: Fixed in latest version.

include/ttmlir/Dialect/D2M/Utils/Utils.h
bgrady-tt: This function is removed in latest

lib/Dialect/D2M/IR/D2MOps.cpp
bgrady-tt: Fixed in latest

lib/Dialect/TTNN/IR/TTNNOpsAttrs.cpp
```diff
@@ -3,6 +3,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 #include "ttmlir/Dialect/TTNN/IR/TTNNOpsAttrs.h"
+#include "ttmlir/AffineMapUtils.h"
```
svuckovicTT, line 6:
```suggestion
#include "ttmlir/AffineMapUtils.h"
#include "ttmlir/Dialect/TTNN/IR/TTNNOpsAttrs.h"
```

test/python/golden/test_metal_virtual_grids.py
```diff
@@ -0,0 +1,80 @@
...
+
+        return result
+
+    compile_and_execute_ttir(
+        eltwise_wrapper,
+        [shape],
+        [dtype],
+        device=device,
+        test_base=request.node.name,
+        print_ir="test_logical_not_ir",
```
anuragsingh-tt, line 76: NIT: different test file.

lib/Dialect/D2M/Transforms/GridSelection.cpp
```diff
@@ -124,6 +125,70 @@ computeGridAwareDimAlignments(ArrayRef<int64_t> logicalShape,
...
+    if (ratio > aspectRatio) {
+      aspectRatio = ratio;
+      maxDimIndex = i;
+    }
+  }
+  return {maxDimIndex, aspectRatio};
+}
+
+int64_t getTargetGridVolume(ArrayRef<int64_t> targetSquareGridShape) {
+  return std::accumulate(targetSquareGridShape.begin(),
```
anuragsingh-tt, line 155: NIT: we can use `ttmlir::utils::volume` here.

lib/Dialect/D2M/Transforms/GridSelection.cpp
```diff
@@ -124,6 +125,70 @@ computeGridAwareDimAlignments(ArrayRef<int64_t> logicalShape,
   return alignments;
 }
 
+//--------------------------------------------------------
+// Virtual Grid
+//--------------------------------------------------------
+
+std::pair<unsigned, double>
```
anuragsingh-tt, line 132: NIT: We could use `static` here for functions in the anonymous namespace.

lib/Dialect/D2M/Utils/Utils.cpp
```diff
@@ -3,6 +3,10 @@
 // SPDX-License-Identifier: Apache-2.0
 
 #include "ttmlir/Dialect/D2M/Utils/Utils.h"
+#include "ttmlir/Asserts.h"
```
anuragsingh-tt, line 6: NIT: redundant include here.

include/ttmlir/AffineMapUtils.h
```diff
@@ -0,0 +1,138 @@
...
+#include "mlir/Conversion/AffineToStandard/AffineToStandard.h"
+#include "mlir/Dialect/Affine/IR/AffineOps.h"
+#include "mlir/IR/AffineMap.h"
+#include "mlir/IR/BuiltinAttributes.h"
+#include "ttmlir/Asserts.h"
+#include "ttmlir/Utils.h"
+#include "llvm/ADT/SmallVector.h"
+
+#include <cstdint>
```
anuragsingh-tt, line 16: NIT: `llvm/ADT/SmallVector.h` already has this.

anuragsingh-tt (approved):

include/ttmlir/AffineMapUtils.h
```diff
@@ -0,0 +1,138 @@
...
+#ifndef TTMLIR_AFFINEMAPUTILS_H
+#define TTMLIR_AFFINEMAPUTILS_H
+
+#include "mlir/Conversion/AffineToStandard/AffineToStandard.h"
+#include "mlir/Dialect/Affine/IR/AffineOps.h"
+#include "mlir/IR/AffineMap.h"
+#include "mlir/IR/BuiltinAttributes.h"
+#include "ttmlir/Asserts.h"
+#include "ttmlir/Utils.h"
+#include "llvm/ADT/SmallVector.h"
```
bgrady-tt, line 14: @vroubtsovTT Fixed include ordering in latest PR #6053

PR #5472, opened 2025-10-24
Add Python bindings for d2m::CBType

brnorris03:
### Ticket
Closes https://github.com/tenstorrent/tt-lang/issues/4

### Problem description
Add Python bindings for d2m::CBType

### Checklist

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -2,6 +2,7 @@ add_subdirectory(unittests)
 
 llvm_canonicalize_cmake_booleans(
     MLIR_ENABLE_BINDINGS_PYTHON
+    TTMLIR_ENABLE_BINDINGS_PYTHON
 )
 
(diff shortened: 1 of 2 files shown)


nsmithtt (approved):
Awesome!

test/python/test_d2m_cbtype.py
```diff
@@ -0,0 +1,41 @@
+# SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
```
ctodTT, line 1: Nothing that should block a PR, but this test seems a little out of place/standalone. I don't think we have a place for explicitly testing bindings though, maybe at some point we create a test battery for this, but for now the test itself looks great!

brnorris03: Yeah, I wasn't really sure where to put it myself...

PR #5502, opened 2025-10-24
[D2M] global data type override frontend pass

ckaravasilisTT:
### Ticket
#3049

Added a pass to convert all data types to a target type, following similar logic to the `ElementTypeNormalization` pass in TTNN.

- Function signatures remain the same
- All types inside the function body get converted
- Typecast operations are inserted at function boundaries to respect input/output types
- Currently only supports F32, BF16, BFP8 as a target type. More will be enabled as typecasting is tested more thoroughly 
- Off by default when running the pipeline

Out of scope for this version:

- Special handling of pre-existing typecasts in function body
- Special handling of constant op/hardcoded constants
- Special handling of quantization ops

### DIFF
--- a/lib/Dialect/D2M/Transforms/GlobalDataFormatConversion.cpp
+++ b/lib/Dialect/D2M/Transforms/GlobalDataFormatConversion.cpp
@@ -22,8 +22,8 @@ parseTargetFormat(llvm::StringRef format) {
   if (format == "bf16") {
     return ttcore::DataType::BFloat16;
   }
-  if (format == "bfp_f8") {
-    return ttcore::DataType::BFP_Float8;
+  if (format == "bfp_bf8") {
+    return ttcore::DataType::BFP_BFloat8;
   }
   return std::nullopt;
 }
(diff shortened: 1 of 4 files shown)


lib/Dialect/D2M/Transforms/GlobalDataFormatConversion.cpp
```diff
@@ -0,0 +1,137 @@
...
+#define GEN_PASS_DEF_D2MGLOBALDATAFORMATCONVERSION
+#include "ttmlir/Dialect/D2M/Transforms/Passes.h.inc"
+
+namespace {
+
+static std::optional<ttcore::DataType>
+parseTargetFormat(llvm::StringRef format) {
+  if (format == "f32") {
+    return ttcore::DataType::Float32;
+  } else if (format == "bf16") {
```
vmilosevic: :warning: **llvm\-else\-after\-return** :warning:
```suggestion
  } if (format == "bf16") {
```

ckaravasilisTT: It should now be fixed!

PR #5503, opened 2025-10-24
add pass to duplicate const-fed broadcasts and decouple sharding

sshonTT:
### Ticket
#5290 

### Problem description
[Provide context for the problem.](https://github.com/tenstorrent/tt-mlir/issues/5290#issuecomment-3438999880)

In short: a single stablehlo.broadcast_in_dim fed by a constant is shared by multiple users that end up with incompatible sharding requirements. The propagation logic then couples these paths, forcing shard alignment and introducing unnecessary collectives (e.g., all-reduce / all-slice) downstream.

### What's changed
This patch duplicates the const-fed broadcast per use, so each consumer can carry its own, independent sharding. That breaks the coupling, prevents redundant CCL ops, and preserves the intended sharding (e.g., keeping bias replicated while other paths shard).

Notes:
- Currently targets `broadcast_in_dim`; other const-fed ops can be added later if needed.

### Checklist

### DIFF
--- a/lib/Dialect/StableHLO/Transforms/DecoupleConstFanout.cpp
+++ b/lib/Dialect/StableHLO/Transforms/DecoupleConstFanout.cpp
@@ -1,0 +1,96 @@
+// SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#include "shardy/dialect/sdy/ir/utils.h"
+#include "stablehlo/dialect/StablehloOps.h"
+#include "ttmlir/Dialect/StableHLO/Transforms/Passes.h"
+
+#include "llvm/Support/Error.h"
+
+namespace mlir::tt::stablehlo {
+#define GEN_PASS_DEF_DECOUPLECONSTFANOUTPASS
+#include "ttmlir/Dialect/StableHLO/Transforms/Passes.h.inc"
+
+// Clone definition op of value each of its uses and connect independent results
+static void cloneDefPerUse(Value v, IRRewriter &rewriter) {
+  llvm::SmallVector<OpOperand *> uses = llvm::to_vector(
+      llvm::map_range(v.getUses(), [](OpOperand &use) { return &use; }));
+
+  // Keep first use as is, clone for rest
+  for (OpOperand *use : llvm::drop_begin(uses)) {
+    rewriter.setInsertionPoint(use->getOwner());
+    Operation *clone = rewriter.clone(*v.getDefiningOp());
+    use->set(clone->getResult(0));
+  }
+}
+
+template <typename T>
+static void duplicateConstFedOps(func::FuncOp func, IRRewriter &rewriter) {
+  SmallVector<T> targets;
+
+  func.walk([&](T op) {
+    Value src = op.getOperand();
+
+    // Condition 1) source is constant feed
+    if (!isa_and_nonnull<mlir::stablehlo::ConstantOp>(src.getDefiningOp())) {
+      return;
+    }
+
+    // Condition 2) result has multiple uses
+    if (op.getResult().getNumUses() < 2) {
+      return;
+    }
+    targets.push_back(op);
+  });
+
+  for (auto op : targets) {
+    rewriter.setInsertionPoint(op);
+    cloneDefPerUse(op.getResult(), rewriter);
+  }
+}
+
+// Returns true if any block argument carries explicit sharding (i.e. any dim
+// has non-empty axes), as opposed to being fully replicated.
+static bool hasExplicitlyShardedArguments(func::FuncOp funcOp) {
+  for (BlockArgument arg : funcOp.getArguments()) {
+    if (auto tensorSharding = mlir::sdy::getSharding(arg)) {
+      for (const auto &dimSharding : tensorSharding.getDimShardings()) {
+        // we consider [{?}] as sharded
+        if (!dimSharding.getAxes().empty()) {
+          return true;
+        }
+      }
+    }
+  }
+  return false; // either no sharding attrs or all replicated
+}
+
+struct DecoupleConstFanoutPass
+    : public impl::DecoupleConstFanoutPassBase<DecoupleConstFanoutPass> {
+public:
+  using impl::DecoupleConstFanoutPassBase<
+      DecoupleConstFanoutPass>::DecoupleConstFanoutPassBase;
+
+  void runOnOperation() override {
+    ModuleOp module = getOperation();
+    MLIRContext *ctx = module.getContext();
+    IRRewriter rewriter(ctx);
+
+    module.walk([&](func::FuncOp func) {
+      // Skip functions without explicitly sharded arguments
+      if (!hasExplicitlyShardedArguments(func)) {
+        return;
+      }
+
+      // TODO(sshon): Right now we copy every constant-fed op with multiple
+      // uses. But actually it is only necessary to duplicate those whose uses
+      // are in sharding conflicts. But we don't have a way to detect sharding
+      // conflict as we rely on sharding propagation pass from upstream. Once we
+      // have a way to detect sharding conflict, we can limit the duplication
+      // only to those ops with sharding conflicts.
+      duplicateConstFedOps<::mlir::stablehlo::BroadcastInDimOp>(func, rewriter);
+    });
+  }
+};
+} // namespace mlir::tt::stablehlo
--- a/include/ttmlir/Dialect/StableHLO/Transforms/Passes.td
+++ b/include/ttmlir/Dialect/StableHLO/Transforms/Passes.td
@@ -124,6 +124,27 @@ def AnalyzeMeshPass : Pass<"analyze-mesh", "::mlir::ModuleOp">
   ];
 }
 
+def DecoupleConstFanoutPass : Pass<
+    "decouple-const-fanout", "::mlir::ModuleOp"> {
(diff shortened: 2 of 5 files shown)


lib/Dialect/StableHLO/Transforms/DecoupleConstFanout.cpp
```diff
@@ -0,0 +1,86 @@
...
+  for (OpOperand *use : llvm::drop_begin(uses)) {
+    Operation *def = v.getDefiningOp();
+    rewriter.setInsertionPoint(use->getOwner());
+    Operation *clone = rewriter.clone(*def);
+    use->set(clone->getResult(0));
+  }
+}
+
+template <typename T>
+static void duplicateConstFedOps(func::FuncOp func, IRRewriter &rewriter) {
```
hshahTT, line 29: Could you add a TODO comment specifying that ideally we would duplicate only when we detect mismatches so it remains on our radar (even if we don't prioritize it anytime soon). In the future it would be good to bring this up with the OpenXLA folks to see if we can fix it in their AggressiveProp pass.

sshonTT: Agree, I added TODO comment to [here](https://github.com/tenstorrent/tt-mlir/pull/5503/files#diff-5846ca05b5b74d8488eb99465f8330d6e1b1bab8e979581ce272dc9717d09873R103)

lib/Dialect/StableHLO/Transforms/DecoupleConstFanout.cpp
```diff
@@ -0,0 +1,113 @@
...
+    uses.push_back(&use);
+  }
+
+  // Keep first use as is, clone for rest
+  for (OpOperand *use : llvm::drop_begin(uses)) {
+    Operation *def = v.getDefiningOp();
+    rewriter.setInsertionPoint(use->getOwner());
+    Operation *clone = rewriter.clone(*def);
+    use->set(clone->getResult(0));
+  }
```
mmanzoorTT: Please try this
```suggestion
auto uses = llvm::to_vector(v.getUses());

// Keep first use as is, clone for the rest.
for (auto *use : llvm::drop_begin(uses)) {
  Operation *clone = rewriter.clone(*v.getDefiningOp());
  rewriter.setInsertionPoint(use->getOwner());
  use->set(clone->getResult(0));
}
```

lib/Dialect/StableHLO/Transforms/DecoupleConstFanout.cpp
```diff
@@ -0,0 +1,113 @@
...
+#include "llvm/Support/Error.h"
+
+namespace mlir::tt::stablehlo {
+#define GEN_PASS_DEF_DECOUPLECONSTFANOUTPASS
+#include "ttmlir/Dialect/StableHLO/Transforms/Passes.h.inc"
+
+// Clone definition op of value each of its uses and connect independent results
+static void cloneDefPerUse(Value v, IRRewriter &rewriter) {
+  // Skip if use count is less than 2
+  if (v.getNumUses() < 2) {
```
mmanzoorTT: Redundant check. you are already performing this check in `duplicateConstFedOps` (line 58).

lib/Dialect/StableHLO/Transforms/DecoupleConstFanout.cpp
```diff
@@ -0,0 +1,113 @@
...
+  func.walk([&](T op) {
+    Value src = op.getOperand();
+
+    // Condition 1) source is constant feed or scalar (rank-0) feed
+    bool srcIsConst = false;
+    bool srcIsScalar = false;
+    if (Operation *defOp = src.getDefiningOp()) {
+      srcIsConst = isa<::mlir::stablehlo::ConstantOp>(defOp);
+    }
+    if (auto rankedType = dyn_cast<RankedTensorType>(src.getType())) {
```
mmanzoorTT: Do you have an example for scalar type? Does it mean that if a previous op generated scalar output or a function argument is of type scalar then this op will clone and create multiple copies of broadcast op?

sshonTT: As noted in [the issue](https://github.com/tenstorrent/tt-mlir/issues/5290#issuecomment-3438999880) , broadcast_in_dim can take a scalar operand. For example:

```
%cst = stablehlo.constant dense<0.000000e+00> : tensor<bf16>
%1 = stablehlo.broadcast_in_dim %cst, dims = [] :
     (tensor<bf16>) -> tensor<2048x2048xbf16>
```

In this example, %cst is a stablehlo.constant with type tensor<bf16> (rank-0), so both flags are true: `srcIsConst = true` and `srcIsScalar = true.`
If the broadcast_in_dim operand comes directly from a function argument of type tensor<bf16> (rank-0), then `srcIsConst = false` and `srcIsScalar = true.`

lib/Dialect/StableHLO/Transforms/DecoupleConstFanout.cpp
sshonTT: Right, I’ll remove this check. I originally wanted cloneDefPerUse to be a reusable utility, but even then this check wasn’t necessary.

lib/Dialect/StableHLO/Transforms/DecoupleConstFanout.cpp
mmanzoorTT: Is this correct assumption that if either condition is true (like your second example where `srcIsConst = false` and `srcIsScalar = true`) then we will clone the broadcast ops?

lib/Dialect/StableHLO/Transforms/DecoupleConstFanout.cpp
sshonTT:
```
/localdev/sshon/tt-xla/third_party/tt-mlir/src/tt-mlir/lib/Dialect/StableHLO/Transforms/DecoupleConstFanout.cpp:17:21: note: in instantiation of function template specialization 'llvm::to_vector<llvm::iterator_range<mlir::ValueUseIterator<mlir::OpOperand>>>' requested here
   17 |   auto uses = llvm::to_vector(v.getUses());
      |                     ^
/opt/ttmlir-toolchain/include/mlir/IR/Value.h:257:19: note: copy constructor of 'OpOperand' is implicitly deleted because base class 'IROperand<OpOperand, Value>' has a deleted copy constructor
  257 | class OpOperand : public IROperand<OpOperand, Value> {
      |                   ^
/opt/ttmlir-toolchain/include/mlir/IR/UseDefLists.h:137:3: note: copy constructor is implicitly deleted because 'IROperand<mlir::OpOperand, mlir::Value>' has a user-declared move constructor
  137 |   IROperand(IROperand &&other) : detail::IROperandBase(std::move(other)) {
```

This fails because v.getUses() yields a range of mlir::OpOperand by reference, and OpOperand has a deleted copy constructor. Since `llvm::to_vector` copies elements, it triggers the error. Instead, I am trying to using llvm::map_range to store address of OpOperand.

lib/Dialect/StableHLO/Transforms/DecoupleConstFanout.cpp
sshonTT: Yes—if either condition holds (e.g., srcIsConst = false and srcIsScalar = true) and the result has multiple uses, we clone the broadcast_in_dim per use. The goal is to decouple fanout to avoid sharding conflicts on the consumer side.
I haven’t seen a real case where the operand comes directly from a scalar function argument with sharding conflict, but it would still qualify. My primary concern here is sharding conflicts arising from broadcast_in_dim; once we have conflict detection, we can narrow the cloning accordingly.

lib/Dialect/StableHLO/Transforms/DecoupleConstFanout.cpp
mmanzoorTT: My bad. Thanks for trying.

lib/Dialect/StableHLO/Transforms/DecoupleConstFanout.cpp
mmanzoorTT: My concern here is that we are widening the scope of this pass/transformation by adding `srcIsScalar` condition. This may cause some unintended consequences. I am more inclined to only use `srcIsConst` condition for now and deal with `srcIsScalar` condition later if we find some use case.

sshonTT: Sounds fair. I will check only `srcIsConst` for now, and get back later if needed.

lib/Dialect/StableHLO/Transforms/DecoupleConstFanout.cpp
```diff
@@ -0,0 +1,100 @@
...
+    Value src = op.getOperand();
+
+    // Condition 1) source is constant feed
+    bool srcIsConst = false;
+    if (Operation *defOp = src.getDefiningOp()) {
+      srcIsConst = isa<::mlir::stablehlo::ConstantOp>(defOp);
+    }
+    if (!srcIsConst) {
+      return;
+    }
```
mmanzoorTT: Please try this out
```suggestion
    if (!isa_and_nonnull<mlir::stablehlo::ConstantOp>(src.getDefiningOp())) {
      return;
    }
```

mmanzoorTT (approved):

lib/Dialect/StableHLO/Transforms/DecoupleConstFanout.cpp
sshonTT: It works. Thanks!

PR #5507, opened 2025-10-26
Fix several d2m.generic problems in verifier, canonicalization, and conversions

brnorris03:
### Ticket
Closes https://github.com/tenstorrent/tt-mlir/issues/5618

### Problem/solution descriptions

This PR contains a collection of small fixes for issues encountered during pykernel-related work. 

1. Problem: The `d2m.generic` verifier requires that pure tensor ops have only 1 region, hence can't have a generic with the usual three regions for the dm and compute threads.
   Solution: Remove the `d2m.generic` verifier requiring that pure tensor ops have only 1 region.
   
   Note: Other parts of the D2M dialect and transforms (e.g., `GridSelection`, `GenericGenerateDatamovement`) also assume or directly access `.getRegion(0)`. These accesses would need to be updated to properly handle multi-region d2m.generic operations. These are not in scope of this PR but should be addressed in subsequent PRs to fully support multi-region d2m.generic.
   
   Tests: test/ttmlir/Dialect/D2M/generic/generic_pure_tensor_multi_region.mlir
   
2. Problem: Several legal dialects are missing in the `D2MGenericReplaceGlobals` conversion.

   Solution: Add a few missing legal dialects (linalg.generic, scf.for) to the `D2MGenericReplaceGlobals` conversion.

   Tests: test/ttmlir/Dialect/D2M/Transforms/generic_replace_globals.mlir (extended with additional test cases)

3. Problem: The `d2m.generic` canonicalization pattern incorrectly replaces DPS operands with values that are defined *after* the op in which they are used.

   Solution: Fix the `d2m.generic` canonicalization to replace DPS operands with values (defined with `wait` or `reserve`) whose definition dominates the use. This includes cross-block dominance checking using MLIR's `DominanceInfo` and `FunctionOpInterface` (instead of just `func::FuncOp`). The verifier now asserts that d2m.generic operations must always be nested within a function.

   Tests: test/ttmlir/Dialect/D2M/generic/generic_canonicalize_dps_dominance.mlir

4. Problem: When `StreamLayoutOp::getBufferType` is called during one-shot bufferization, it passes `isView=true` to `ttcore::getBufferType`. This function then tries to get the index affine map from the `MetalLayoutAttr`, but the layout doesn't have an explicit index map set (it defaults to an empty map). 

   Solution: Use `getIndexAffineMapOrIdentity` instead of `getIndexAffineMap` in `ttcore::getBufferType`. The semantics of `index_map` in `MetalLayoutAttr` are defined as "Optional index remapping (e.g., view remap or virtual→physical grid mapping). Defaults to empty, which is treated as identity."

   Tests: test/ttmlir/Dialect/TTCore/Transforms/one_shot_bufferize_stream_layout_empty_index_map.mlir

   Unrelated to the above problem, the rest of the one-shot bufferization tests were modified to use `--ttcore-register-device` instead of defining a `#ttcore.system_desc` attribute to simplify and reduce boilerplate (and be more consistent with other tests in the project).

5. Problem: The verifier for d2m.generic did not validate that indexing_maps must be non-empty when not in explicit datamovement form (in which all three -- block_factors, indexing_maps, and iterator_types -- are empty), leading to invalid IR passing validation.

   Solution: Introduce a new `isExplicitDatamovementForm()` that returns true when all of block_factors, indexing_maps, and iterator_types are empty (user is responsible for managing all data movement). Add verifier logic in d2m.generic to require non-empty indexing_maps unless in explicit datamovement form. Also fix `GenericGenerateLoops` to correctly create d2m.generic operations by explicitly setting block_factors to empty when creating the transformed GenericOp.

   Tests: test/ttmlir/Dialect/D2M/generic/generic_verify_indexing_maps.mlir

6. Problem: The `d2m-insert-dst-register-access` pass crashed when encountering `d2m.generic` operations with empty indexing_maps because it attempted to access indexing map elements and compute DST capacity without checking if the maps existed.

   Solution: Modify the pass to use the `isExplicitDatamovementForm()` method (from 5) to detect explicit data movement form. Always calculate DST capacity and insert DST register access, even when indexing_maps is empty. Only skip tile matmul rewrite when in explicit datamovement form, falling through to regular linalg.generic to affine loop conversion.

   Tests: test/ttmlir/Dialect/D2M/Transforms/insert_dst_register_access_empty_indexing_maps.mlir

7. Problem: The `d2m-allocate` pass crashed when encountering `d2m.generic` operations with empty indexing_maps in `getOperandContexts` because it attempted to analyze reduction/broadcast dimensions without checking if indexing maps existed. Without indexing maps, we cannot determine which dimensions participate in reductions or broadcasts.

   Solution: Add a check in `getOperandContexts` to use `isExplicitDatamovementForm()` (from 5) to skip reduction/broadcast dimension analysis when in explicit datamovement form. Stream requirements will still be determined by other factors (e.g., memory space for DRAM inputs).

   Tests: test/ttmlir/Dialect/D2M/allocate/allocate_empty_indexing_maps.mlir

### Checklist


nsmithtt (approved):
Thank you for the comprehensive tests!!

phizalev-TT (approved):

lib/Dialect/D2M/Transforms/Allocate.cpp
```diff
@@ -397,6 +396,12 @@ class D2MAllocate final : public impl::D2MAllocateBase<D2MAllocate> {
         continue;
       }
 
+      // If indexing maps are not yet generated (pre-loop generation),
+      // we can't analyze reduction/broadcast dimensions, so skip this analysis.
+      if (genericOp.getIndexingMaps().empty()) {
+        continue;
+      }
+
       // A core participating in a reduction dim necessarily requires
```
vroubtsovTT: I am not aware of `d2m.generic` uses with empty indexing maps... `GenericOp::build()` that many other builders route through has
```
  TT_assertv(!indexingMaps.empty(), "expected non-empty indexing maps");
```
which is not being altered by this PR.

vroubtsovTT: As discussed elsewhere, let's condition this early out on some new GenericOp getter that conveys its "form" intent more explicitly, something like `isExplicitDatamovementForm()` or similar name.

brnorris03: Added this and used instead of checking for empty indexing maps in the 2 relevant items. Please take a look!

lib/Dialect/D2M/IR/D2MOps.cpp
```diff
@@ -1332,6 +1352,10 @@ unsigned d2m::GenericOp::getNumDims() {
 }
 
 mlir::AffineMap d2m::GenericOp::getIndexingMap(int64_t operandIndex) {
+  // If indexing_maps is empty, return an empty affine map.
+  if (getIndexingMaps().empty()) {
+    return AffineMap::get(0, 0, {}, getContext());
+  }
```
vroubtsovTT: I think we should error out instead, e.g. by having an input check
```
TT_assert_limit(operandIndex, getIndexingMapsValue().size());
```
 since we can't use exceptions. Because otherwise we would have `getIndexingMap(i)` not equivalent to `getIndexingMapsValue()[i]` and that's a terrible API design.

lib/Dialect/D2M/IR/D2MOps.cpp
```diff
@@ -1455,6 +1479,10 @@ d2m::GenericOp::getParticipatingLoopDims(int64_t operandIndex) {
 mlir::SmallVector<int64_t>
 d2m::GenericOp::getNonParticipatingLoopDims(int64_t operandIndex) {
   AffineMap indexingMap = getIndexingMap(operandIndex);
+  // If indexing_maps is empty, return empty vector (no non-participating dims).
+  if (indexingMap.getNumDims() == 0) {
+    return {};
+  }
```
vroubtsovTT: If you accept the suggestion for `getIndexingMap(operandIndex)` above, this if-check here will not be necessary (I think).

lib/Dialect/D2M/Transforms/Allocate.cpp
```diff
@@ -397,6 +396,12 @@ class D2MAllocate final : public impl::D2MAllocateBase<D2MAllocate> {
         continue;
       }
 
+      // In explicit datamovement form, we can't analyze reduction/broadcast
+      // dimensions, so skip this analysis.
+      if (genericOp.isExplicitDatamovementForm()) {
+        continue;
+      }
```
vroubtsovTT, line 403: the allocator code here has already been modified in a parallel PR but I will make use of `isExplicitDatamovementForm()` for a similar effect (exclude such generic ops from allocator analysis)

lib/Dialect/D2M/IR/D2MOps.cpp
nsmithtt: +1, perhaps we should have an optional return value.

brnorris03: Changed as discussed on Slack.

lib/Dialect/D2M/IR/D2MOps.cpp
brnorris03: I made this more consistent by checking for explicit data movement mode instead.

PR #5513, opened 2025-10-27
Add const-eval hoisting for standalone cacheable creation ops

### LINKED ISSUE #5512: Const eval standalone creation ops
### Describe the bug

Ops in TTIR like
```
    %0 = "ttir.constant"() <{value = dense<1.000980e-01> : tensor<1x256x60x80xbf16>}> : () -> tensor<1x256x60x80xbf16> loc(#loc)
    %1 = "ttir.constant"() <{value = dense<0.000000e+00> : tensor<1x256x60x80xbf16>}> : () -> tensor<1x256x60x80xbf16> 
```
get translated to 
```
        %118 = "ttnn.full"(%117) <{dtype = #ttcore.supportedDataTypes<bf16>, fill_value = 0.100097656 : f32, layout = #ttnn.layout<tile>, shape = #ttnn.shape<1x256x60x80>}> : (!ttnn.device) -> tensor<1x256x60x80xbf16, #ttnn_layout148>
        %119 = "ttnn.full"(%117) <{dtype = #ttcore.supportedDataTypes<bf16>, fill_value = 0.000000e+00 : f32, layout = #ttnn.layout<tile>, shape = #ttnn.shape<1x256x60x80>}> : (!ttnn.device) -> tensor<1x256x60x80xbf16, #ttnn_layout148>
```
Currently these are executed in every inference run, but they can be put into `const eval` functions.

rpavlovicTT:
This commit enhances the const-eval hoisting pass to identify and hoist standalone cacheable creation operations (those with CanExecuteOnHostTrait and no tensor operands) into separate const-eval subgraphs, even when they are consumed by non-const-eval operations.

Previously, creation ops like ttnn.full were only hoisted if they were consumed by other const-eval operations. This meant that constant tensors used directly by non-const-eval operations were recreated on every inference run, missing optimization opportunities for caching across multiple runs.

Fixes #5512

### DIFF
--- a/tools/ttrt/common/emitpy.py
+++ b/tools/ttrt/common/emitpy.py
@@ -355,6 +355,7 @@ class EmitPy:
                         isinstance(node, ast.FunctionDef)
                         and node.name != "main"
                         and node.name[0:18] != "create_inputs_for_"
+                        and not node.name.__contains__("_const_eval_")
                     ):
                         program_names.append(node.name)
 


lib/Transforms/ConstEvalHoist.cpp
```diff
@@ -301,6 +317,26 @@ class ConstEvalAnalyze {
         ->hasTrait<mlir::tt::ttcore::Trait::TTCoreDuplicateConstEvalTrait>();
   }
 
+  bool isCacheableCreationOp(mlir::Operation *op) {
+    assert(op != nullptr);
+    if (!isCreationOp(op)) {
+      return false;
+    }
+
+    if (!op->hasTrait<mlir::tt::ttnn::CanExecuteOnHostTrait>()) {
```
sdjordjevicTT: Can you just comment here and elaborate more on why caching only works for creation ops that can execute on the host?

sdjordjevicTT (approved):

lib/Transforms/ConstEvalHoist.cpp
rpavlovicTT: My bad here,  we don't need this trait for const-eval. Thanks for catching it!

tapspatel (approved):

Issue #5224, opened 2025-10-03
D2M/TTNN Integration -  Create a pipeline for the integration path


vtangTT:
should include ttnn-to-ttir conversion + all of d2m pipeline w/ ttnn-mode=true

nsmithtt:
Is this issue still relevant @vtangTT / @arminaleTT ?

vtangTT:
we can prob still work under this issue for the current TTNN optimizer integration unless there is another more recent one opened.

PR #5522, opened 2025-10-27
D2M/TTNN Integration pipeline

vtangTT:
### Ticket
https://github.com/tenstorrent/tt-mlir/issues/5224

### Problem description
Separate pipeline for the ttnn integration path. Also no longer expose `ttnn-mode` as a pipeline option.

### What's changed
- `TTMetalPipelines`: removed `ttnn-mode` from the pipeline options
    - added new `ttnn-to-ttmetal-backend-pipeline` and `ttir-to-ttnn-generic-pipeline`
    - renamed `TTIRBufferization` to `D2MBufferization`
    - completely separate `TTNND2MBufferization` for the integration path
- `TTIRToD2M`: passing `TTIRToD2MOptions` instead of individual `collapseTensorsTo2D`, `ttnnMode`, etc. options

### DIFF
--- a/lib/Dialect/TTMetal/Pipelines/TTMetalPipelines.cpp
+++ b/lib/Dialect/TTMetal/Pipelines/TTMetalPipelines.cpp
@@ -38,16 +38,10 @@ std::unique_ptr<Pass> createCanonicalizerPassWithOptions(
   return mlir::createCanonicalizerPass({}, disabledPatterns);
 }
 
-void createTTIRBufferizationPipeline(
-    OpPassManager &pm, const TTIRToTTMetalPipelineOptions &options) {
+void createD2MBufferizationPipeline(OpPassManager &pm) {
   bufferization::OneShotBufferizePassOptions bufferizePassOptions;
-  if (options.ttnnMode) {
-    bufferizePassOptions.allowUnknownOps = true;
-    bufferizePassOptions.bufferizeFunctionBoundaries = false;
-  } else {
-    bufferizePassOptions.allowUnknownOps = false;
-    bufferizePassOptions.bufferizeFunctionBoundaries = true;
-  }
+  bufferizePassOptions.allowUnknownOps = false;
+  bufferizePassOptions.bufferizeFunctionBoundaries = true;
   bufferizePassOptions.functionBoundaryTypeConversion =
       bufferization::LayoutMapOption::IdentityLayoutMap;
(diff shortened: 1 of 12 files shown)


lib/Dialect/TTMetal/Pipelines/TTMetalPipelines.cpp
```diff
@@ -214,6 +239,33 @@ void createTTIRToTTMetalPipeline(OpPassManager &pm,
...
+  createTTIRToTTMetalFrontendPipelineImpl(devicePm, options, true);
+  createTTIRToTTMetalMiddleendPipelineImpl(devicePm, options, true);
+  createTTIRToTTMetalBackendPipelineImpl(devicePm, options, true);
+
+  // Run lowering to LLVM pass on hoisted funcs in CPUModule.
+  ttir::LinalgToLLVMPipelineOptions linalgToLLVMOptions;
+  ttir::createTTIRToCPUPipeline(pm, linalgToLLVMOptions);
+}
+
+void createTTNNToTTMetalBackendPipeline(
```
arminaleTT, line 262: I don't like this name because it doesn't line up with the names for the underlying passes/pipelines, but I can't think of a better suggestion

arminaleTT (approved):

nsmithtt (commented):
Not opposed nor for this change.  It's not clear to me that we should do this, does it solve an immediate problem?  The old way was simpler in some respects.

nsmithtt:
imo it might be a better North Star to unify the pipelines further and all the passes are just robust to IR from both contexts.  But I'm not sure yet.  This change though starts the fork which will make it difficult to undo later if we do decide that unification is a better strategy.

vtangTT:
> Not opposed nor for this change. It's not clear to me that we should do this, does it solve an immediate problem? The old way was simpler in some respects.

Doesn't solve an immediate problem. The goal was to hide away the `ttnn-mode` flag from the user, thus have them run a separate pipeline instead if using the integration path. I agree that the old way was simpler in some aspects.

> imo it might be a better North Star to unify the pipelines further and all the passes are just robust to IR from both contexts. But I'm not sure yet. This change though starts the fork which will make it difficult to undo later if we do decide that unification is a better strategy.

Hmm yea, I don't have a compelling argument to do one over the other. However, if we're looking to unify in the future, then this isn't the way to go 🤔

Issue #5495, opened 2025-10-24
CPU-hoisted `ttir.conv2d` fails in runtime with `undefined symbol: memrefCopy`


dmilinkovicTT:
After the problem described in #5165 has been fixed with #5471, new issue emerged for CPU-hoisted `ttir.conv2d` test.

## The issue

The lowered code contains the following declaration in the CPU module:
```
llvm.func @memrefCopy(i64, !llvm.ptr, !llvm.ptr)
```

The implementation for this function isn't available in the lowered code, which leads to the following runtime failure:
```
undefined symbol: memrefCopy
``` 

CPU-hoisted executable should be dynamically linked against `libmlir_c_runner_utils.so`. However -- we can't do this, since `ttrt` doesn't support dynamic linking. 

## Why does `memrefCopy` invocation exist?

When `memref` dialect is being lowered, `memref.copy` lowers to:
- intrinsic `memcpy`, if both operands are identity buffers (with regular strides)
- invocation of `memrefCopy` function, otherwise

(source: [MemRefToLLVM.cpp](https://mlir.llvm.org/doxygen/MemRefToLLVM_8cpp_source.html):1234)

In case of `conv2d`, the latter happens, and thus, we're ending up with the `memrefCopy` invocation.

PR #5529, opened 2025-10-28
Fix CPU-hoisted `ttir.conv2d`

dmilinkovicTT:
### Ticket
#5495

### Problem description
CPU-hoisted `ttir.conv2d` fails because `memref.copy` gets lowered into an invocation of MLIR runtime's `@memrefCopy` function, since it operates on a strided memref.

We want to avoid dynamically linking the MLIR runtime libs, so the alternative is to lower the problematic `memref.copy` to `linalg.generic`, thus preventing it to get lowered to `@memrefCopy` invocation.

Another issue with hoisted variant of this op is the fact that `tensor.extract_slice` op, which can be created when lowering `ttir.conv2d` to `linalg`, isn't a DPS op, so additional `memref.copy` is required in order to make it CPU-hoistable.

### What's changed
- Implemented `NonContiguousMemrefCopyToLinalg` pass
- Enabled copying the `tensor.extract_slice` into the output buffer
- Un-quarantined the test

### Checklist

### DIFF
--- a/lib/Dialect/LLVM/Transforms/NonContiguousMemrefCopyToLinalg.cpp
+++ b/lib/Dialect/LLVM/Transforms/NonContiguousMemrefCopyToLinalg.cpp
@@ -59,7 +59,8 @@ class NonContiguousMemrefCopyToLinalg
                memref::isStaticShapeAndContiguousRowMajor(memrefType)));
     };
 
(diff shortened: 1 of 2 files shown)


vwellsTT (approved):
This looks great to me, thanks!

PR #5531, opened 2025-10-28
Pass through non bfloat16 dtypes in bfp8 conversion pass

### LINKED ISSUE #5532: Ignore non bfloat16 dtypes in bfp8 conversion
Currently, when TTIR graph gets converted from bfloat16 to bfp8, it is required that whole graph is bfloat16.
In many cases we have mixed precision graphs (float32, int32, and bfloat16) coming from frontends.

BFP8 Conversion pass should ignore non bfloat16 dtypes and not assert. It should only cast bfloat16 to bfp8 for now...

dgolubovicTT:
Closes #5532 
### Problem description
Currently, when TTIR graph gets converted from bfloat16 to bfp8, it is required that whole graph is bfloat16.
In many cases we have mixed precision graphs (float32, int32, and bfloat16) coming from frontends.

BFP8 Conversion pass should ignore non bfloat16 dtypes and not assert. It should only cast bfloat16 to bfp8 for now and pass through everything else.

### What's changed
- Edit block format conversion pattern to convert only bfloat16 while passing through other types.
- Add test with float32 operation in the graph test/ttmlir/Dialect/TTNN/block_types/block_type_conv_mixed_precision.mlir


lib/Dialect/TTIR/Transforms/ElementTypeNormalization.cpp
```diff
@@ -151,6 +151,18 @@ class FuncBodyTypeCast : public mlir::ConversionPattern {
   LogicalResult
   matchAndRewrite(mlir::Operation *op, llvm::ArrayRef<mlir::Value> operands,
                   mlir::ConversionPatternRewriter &rewriter) const final {
+
+    // Fire this pattern ONLY if at least one result is bf16.
```
mtopalovicTT: I would just say
`Match this pattern ONLY if at least one result is bf16`

lib/Dialect/TTIR/Transforms/ElementTypeNormalization.cpp
```diff
@@ -181,20 +193,16 @@ class FuncBodyTypeCast : public mlir::ConversionPattern {
...
-        mlir::Type elementType = type.getElementType();
-        if (!mlir::isa<BFloat16Type>(elementType)) {
-          assert(mlir::isa<ttcore::TileType>(elementType) &&
-                 "Expected TileType for non-bfloat16 element type.");
-          assert(
-              mlir::cast<ttcore::TileType>(elementType).getDataType() ==
-                  ttcore::DataType::BFP_BFloat8 &&
-              "Expected BFP_BFloat8 TileType for non-bfloat16 element type.");
-          return type;
+        mlir::Type el_type = type.getElementType();
```
mtopalovicTT: nit: elType

azecevicTT (approved):

Issue #3037, opened 2025-04-22
Support toLayout lowering of metal_layout change


nsmithtt:
Sync with @nsmithtt

nsmithtt:
This was supported w/
- #3167

nsmithtt:
This was incorrectly closed, this issue is meant to track this condition in the lower to layout pass:
```c++
  static LogicalResult lowerLayoutChange(PatternRewriter &rewriter,
                                         ToLayoutOp op) {
    assert(false &&
           "TODO issue https://github.com/tenstorrent/tt-mlir/issues/3037");
  }
```

nsmithtt:
@vwellsTT I think this could be a good task to look at next before we add more features to the grid selection pass.

nsmithtt:
I think this was largely completed, we can open new issues for improvements.

PR #5540, opened 2025-10-28
[D2M] Handle all shape + stride changes as affine map => view + DMA

vwellsTT:
### Ticket
https://github.com/tenstorrent/tt-mlir/issues/3037

### Problem description
We lacked support for lowering what we called a "layout change" in LowerToLayout--any fundamental change to the alignments, collapse dims, etc. could not be lowered.

### What's changed
Consolidate grid changes + stride changes into a new class of change called "mappingChange"-- we inverse the logical -> device mapping on the input, compose this with the logical -> device mapping on the output, and therefore map the input to the output as a purely affine-map transformation.  This allows us to lower a broad range of changes through a single simple mechanism.
We also immediately materialize these views into new tensors via inserting a genericOp with a DMAOp--this was necessary to avoid race conditions when evaluating the views (using StreamLayoutOps instead did not work, due to lack of a folding pattern today).

### Checklist

### DIFF
--- a/include/ttmlir/Dialect/D2M/Utils/Utils.h
+++ b/include/ttmlir/Dialect/D2M/Utils/Utils.h
@@ -16,6 +16,14 @@ getSquareTargetGrid(mlir::ArrayRef<int64_t> targetGridShape);
 // Get the largest destination element type used in a region.
 Type getRegionLargestDstElemType(Region &region);
 
+// Computes dim constraints implied by the indexing maps and shapes. If
+// successful, returns a vector of dim constraints for each dimension; a '0'
+// indicates that the dimension is not constrained. If the shapes imply
+// incompatible constraints, returns std::nullopt.
+std::optional<SmallVector<int64_t>>
+computeDimConstraints(mlir::ArrayRef<mlir::AffineMap> indexingMaps,
+                      mlir::ArrayRef<mlir::SmallVector<int64_t>> shapes);
+
 } // namespace mlir::tt::d2m::utils
 
 #endif
--- a/lib/Dialect/D2M/Utils/Utils.cpp
+++ b/lib/Dialect/D2M/Utils/Utils.cpp
@@ -63,4 +63,33 @@ Type getRegionLargestDstElemType(Region &region) {
   return largestType;
 }
 
+std::optional<SmallVector<int64_t>>
+computeDimConstraints(mlir::ArrayRef<mlir::AffineMap> indexingMaps,
+                      mlir::ArrayRef<mlir::SmallVector<int64_t>> shapes) {
+  TT_assert(!indexingMaps.empty());
+  TT_assert(indexingMaps.size() == shapes.size());
+  auto numDims = indexingMaps.front().getNumDims();
+  SmallVector<int64_t> constrainedDims(numDims, 0);
+  for (auto [shapeIdx, shape] : llvm::enumerate(shapes)) {
+    auto dimProjectionMap =
+        mlir::inverseAndBroadcastProjectedPermutation(indexingMaps[shapeIdx]);
+    auto impliedDimConstraints = dimProjectionMap.compose(shape);
+
+    for (auto [dimIdx, dimConstraint] :
+         llvm::enumerate(impliedDimConstraints)) {
+      if (dimConstraint == 0) {
+        continue;
+      }
+
+      // Early exit if shapes are incompatible.
+      if (constrainedDims[dimIdx] != 0 &&
+          constrainedDims[dimIdx] != dimConstraint) {
+        return std::nullopt;
+      }
+      constrainedDims[dimIdx] = dimConstraint;
+    }
+  }
+  return constrainedDims;
+}
+
 } // namespace mlir::tt::d2m::utils


vwellsTT:
~I want to see if CI passes on this PR, but not sure if it's really ready for review--need to decide how to properly test something like this (facing problems with getting builder tests which actually materialize a view working today), FWIW~
now ready for reviews

include/ttmlir/Dialect/D2M/Utils/Utils.h
nsmithtt, line 1: I wonder if it makes sense to rename this file to be AffineMapUtils.h?  It seems like a good idea to have a dedicated place for all of our affine needs since more are likely to come with TM support.

nsumrakTT:
Seems that you have a problem with rebase

runtime/lib/ttmetal/executor_utils.h
```diff
@@ -307,16 +307,19 @@ inline std::string createKernelFilePath(
   if (debug::Env::get().useLocForKernelName && kernelLoc) {
     path /= kernelLoc;
   } else {
-    prefix /= "ttmlir_";
```
vwellsTT: remark: this logic is somewhat unrelated, but `--dump-kernels` seems to be broken currently, this fixes.  Ran into this during debugging issues with rest of my PR

vwellsTT: I guess @wenbinlyuTT will fix here: https://github.com/tenstorrent/tt-mlir/pull/5776/files#diff-e864c47d212902e40d7a3b956c5fe169dafbb011c3fa1c7b7c462a7dcb2eb4bd so I'll remove

include/ttmlir/Dialect/D2M/Utils/Utils.h
vwellsTT: not quite everything in this file is currently affine map related, but no objections to splitting all of those into what you suggest; I'll do that for now and we can see how it looks

vwellsTT: (I guess only `getSquareTargetGrid & getRegionLargestDstElemType` aren't, ha)

lib/Dialect/D2M/IR/D2MOpsInterfaces.cpp
```diff
@@ -35,11 +35,19 @@ mlir::tt::d2m::applyViews(mlir::Operation *op) {
   Value input = viewOp.getInput();
   auto inputMemref = mlir::cast<mlir::MemRefType>(input.getType());
 
+  // Recursively apply view composition if the input is also a view. This
+  // handles nested view chains by composing affine maps.
+  if (mlir::isa<ttcore::ViewLayoutAttr>(inputMemref.getLayout())) {
+    if (auto inputOp = input.getDefiningOp()) {
```
vmilosevic: :warning: **llvm\-qualified\-auto** :warning:
`` auto inputOp `` can be declared as `` auto *inputOp ``
```suggestion
    if (auto *inputOp = input.getDefiningOp()) {
```

lib/Dialect/D2M/Transforms/LowerToLayout.cpp
```diff
@@ -237,7 +380,8 @@ class D2MLowerToLayoutRewriter : public OpRewritePattern<ToLayoutOp> {
 
   LogicalResult matchAndRewrite(ToLayoutOp op,
                                 PatternRewriter &rewriter) const final {
-    auto components = op.compoundComponents();
+    auto components =
+        CompoundComponents::analyze(op.getInput(), op.getOutput());
```
vmilosevic: :warning: **clang\-diagnostic\-error** :warning:
use of undeclared identifier `` CompoundComponents ``

lib/Dialect/D2M/Transforms/LowerToLayout.cpp
```diff
@@ -425,115 +609,141 @@ class D2MSplitCompoundLayoutRewriter : public OpRewritePattern<ToLayoutOp> {
...
             inputInfo.type, *outputInfo.layout, outputInfo.type,
             ttcore::MemorySpace::DeviceL1, getTargetGridShape());
         bounce(rewriter, op, bounceType);
+        return success();
       } else {
-        // Device -> System: need intermediate in L1 with input's
-        // characteristics.
-        assert(inputInfo.layout && "Input must have layout for device->system");
-        bool reblockVirtualGridShapes = true;
+        // Device → System: need to prepare for transfer
```
vmilosevic: :warning: **llvm\-else\-after\-return** :warning:
```suggestion
      }         // Device → System: need to prepare for transfer
```

lib/Dialect/D2M/Transforms/LowerToLayout.cpp
```diff
@@ -425,115 +609,141 @@ class D2MSplitCompoundLayoutRewriter : public OpRewritePattern<ToLayoutOp> {
...
+              ttcore::MemorySpace::DeviceL1,
+              /*newTensorGrid=*/{}, /*newElementType=*/{},
+              /*newTileShape=*/{}, /*reblockVirtualGridShapes=*/true);
+          bounce(rewriter, op, bounceType);
+          return success();
+        } else {
+          // Ready for system transfer (handled by D2MLowerToLayoutRewriter)
+          return failure();
+        }
+      }
```
vmilosevic, line 730: :warning: **llvm\-else\-after\-return** :warning:
```suggestion
     
```

lib/Dialect/D2M/Transforms/LowerToLayout.cpp
```diff
@@ -425,115 +609,141 @@ class D2MSplitCompoundLayoutRewriter : public OpRewritePattern<ToLayoutOp> {
...
+          // Untilize first
+          Type scalarType = getScalarType(inputInfo.type.getElementType());
+          auto bounceType = typeBuilder.modifyDeviceType(
+              inputInfo.type, *inputInfo.layout, getTargetGridShape(),
+              ttcore::MemorySpace::DeviceL1,
+              /*newTensorGrid=*/{}, scalarType,
+              /*newTileShape=*/std::nullopt);
+          bounce(rewriter, op, bounceType);
+          return success();
+        } else if (inputHasVirtualGrid) {
```
vmilosevic: :warning: **llvm\-else\-after\-return** :warning:

lib/Dialect/D2M/Transforms/LowerToLayout.cpp
```diff
@@ -425,115 +609,141 @@ class D2MSplitCompoundLayoutRewriter : public OpRewritePattern<ToLayoutOp> {
...
         auto bounceType = typeBuilder.modifyDeviceType(
             inputInfo.type, *inputInfo.layout, getTargetGridShape(),
-            ttcore::MemorySpace::DeviceL1,
-            /*newTensorGrid=*/{}, outputInfo.type.getElementType(),
-            /*newTileShape=*/{}, reblockVirtualGridShapes);
+            ttcore::MemorySpace::DeviceL1);
         bounce(rewriter, op, bounceType);
+        return success();
+      } else {
+        // L1 → DRAM (handle other transformations first)
```
vmilosevic: :warning: **llvm\-else\-after\-return** :warning:
```suggestion
      }         // L1 → DRAM (handle other transformations first)
```

lib/Dialect/D2M/Transforms/LowerToLayout.cpp
```diff
@@ -425,115 +609,141 @@ class D2MSplitCompoundLayoutRewriter : public OpRewritePattern<ToLayoutOp> {
...
+        if (formatChange || virtualGridChange) {
+          // These need to be handled before DRAM transfer
+          // Fall through to handle them
+        } else {
+          // Ready for DRAM transfer
+          return failure();
+        }
       }
```
vmilosevic: :warning: **llvm\-else\-after\-return** :warning:
```suggestion
     
```

lib/Dialect/D2M/Transforms/LowerToLayout.cpp
```diff
@@ -425,115 +609,141 @@ class D2MSplitCompoundLayoutRewriter : public OpRewritePattern<ToLayoutOp> {
...
             /*memSpace=*/{},
-            /*newTensorGrid=*/{}, inputInfo.type.getElementType(),
-            ttcore::getTensorTileShape(inputInfo.type));
+            /*newTensorGrid=*/{}, scalarType,
+            /*newTileShape=*/std::nullopt);
         bounce(rewriter, op, bounceType);
+        return success();
       } else {
-        // Scalar -> tilized: use input's layout/grid, change element type.
+        // Scalar → tilize
```
vmilosevic: :warning: **llvm\-else\-after\-return** :warning:
```suggestion
      }         // Scalar → tilize
```

lib/Dialect/D2M/Transforms/LowerToLayout.cpp
```diff
@@ -425,115 +609,141 @@ class D2MSplitCompoundLayoutRewriter : public OpRewritePattern<ToLayoutOp> {
...
         auto bounceType = typeBuilder.modifyDeviceType(
             inputInfo.type, *inputInfo.layout, getTargetGridShape(),
             /*memSpace=*/{},
             /*newTensorGrid=*/{}, outputInfo.type.getElementType(),
             ttcore::getTensorTileShape(outputInfo.type));
         bounce(rewriter, op, bounceType);
+        return success();
       }
```
vmilosevic, line 780: :warning: **llvm\-else\-after\-return** :warning:
```suggestion
     
```

lib/Dialect/D2M/Transforms/LowerToLayout.cpp
```diff
@@ -395,19 +198,309 @@ class D2MSplitCompoundLayoutRewriter : public OpRewritePattern<ToLayoutOp> {
...
+                        .getResult();
+                auto dma = builder.create<d2m::DMAOp>(innerLoc, viewOp,
+                                                      indexingMap, outputCB);
+                builder.create<d2m::DMAWaitOp>(innerLoc, dma);
+                builder.create<YieldOp>(innerLoc, outputCB);
+              },
+              ThreadType::Datamovement, grid)
+          .getResult(0);
+    } else {
+      // DRAM operations use the view directly without immediate
```
vmilosevic: :warning: **llvm\-else\-after\-return** :warning:
```suggestion
    }       // DRAM operations use the view directly without immediate
```

lib/Dialect/D2M/Transforms/LowerToLayout.cpp
```diff
@@ -395,19 +198,309 @@ class D2MSplitCompoundLayoutRewriter : public OpRewritePattern<ToLayoutOp> {
...
+      // materialization.
+      return viewOp;
+    }
```
vmilosevic, line 347: :warning: **llvm\-else\-after\-return** :warning:
```suggestion
   
```

lib/Dialect/D2M/Transforms/LowerToLayout.cpp
```diff
@@ -395,19 +198,309 @@ class D2MSplitCompoundLayoutRewriter : public OpRewritePattern<ToLayoutOp> {
...
+
+    assert(inputInfo.isSystem() != outputInfo.isSystem() &&
+           "one of input or output must be system for now");
+
+    // Use the layout of whichever side has a layout (input or output).
+    auto deviceLayout =
+        inputInfo.isSystem() ? outputInfo.layout : inputInfo.layout;
+    assert(deviceLayout.has_value() && "Device side must have a layout");
+
+    // TODO: If the device side has a virtual grid (non-empty index map),
```
missing username/bug in TODO
```suggestion
    // TODO(nsmith): If the device side has a virtual grid (non-empty index map),
```

vmilosevic (commented):

lib/Target/TTMetal/TTMetalToFlatbuffer.cpp
```diff
@@ -254,16 +254,33 @@ createShardedBufferConfigForDRAMMemref(FlatbufferObjectCache &cache,
...
+// the virtualization map places them.
 static SmallVector<int64_t>
 getPhysicalGridShapeForVirtualGrid(ttcore::ShardLayoutAttr shardLayout,
                                    ttcore::DeviceAttr device,
-                                   ArrayRef<int64_t> memrefGridShape) {
-  auto squareTargetGrid =
-      d2m::utils::getSquareTargetGrid(device.getWorkerGrid().getShape());
-  return squareTargetGrid;
+                                   MemRefType memref) {
+  auto virtMap = shardLayout.getCoreVirtualizationMap();
```
bgrady-tt, line 264: I think the bottom portion of code here could use `ttmlir::utils::applyMapToGrid()`

lib/Dialect/D2M/IR/D2MOps.cpp
```diff
@@ -739,14 +699,15 @@ mlir::LogicalResult d2m::ViewLayoutOp::verify() {
...
-      }
+      // if (inputLayout.getCollapsedIntervals() !=
+      //     resultLayout.getCollapsedIntervals()) {
+      //   return emitOpError("view cannot change collapsed intervals");
+      // }
 
-      if (inputLayout.getDimAlignments() != resultLayout.getDimAlignments()) {
-        return emitOpError("view cannot change dim alignments");
-      }
+      // if (inputLayout.getDimAlignments() != resultLayout.getDimAlignments())
```
bgrady-tt: remove commented code.

lib/Dialect/D2M/IR/D2MOps.cpp
```diff
@@ -739,14 +699,15 @@ mlir::LogicalResult d2m::ViewLayoutOp::verify() {
         return emitOpError("view cannot change memory space");
       }
 
-      if (inputLayout.getCollapsedIntervals() !=
-          resultLayout.getCollapsedIntervals()) {
-        return emitOpError("view cannot change collapsed intervals");
-      }
+      // if (inputLayout.getCollapsedIntervals() !=
```
bgrady-tt: remove commented code

lib/Dialect/D2M/IR/D2MOps.cpp
```diff
@@ -907,18 +868,31 @@ void d2m::GenericOp::build(mlir::OpBuilder &builder,
...
+    if (!outputIsView && metalLayout && hasNonIdentityIndexMap) {
       auto shapedType = mlir::cast<ShapedType>(outputs[0].getType());
       auto physicalGridShape = metalLayout.getPhysicalGridShape(shapedType);
-      auto [_, invMap] = ttmlir::d2m::utils::grids::createCoreVirtMaps(
-          builder.getContext(), gridShape, physicalGridShape);
 
-      grid = builder.getAttr<ttcore::GridAttr>(gridShape, invMap);
+      // Only create virtual grid mapping if the grid actually exceeds the
+      // physical grid (needs virtualization). Non-virtual index_maps (like
+      // reblocking) don't need a grid inverse mapping.
```
bgrady-tt: I would strongly prefer that if an output tensor has a non-identity index map that it always have an inverse map attached to the grid as well. 

1. It is simpler to verify that either the output tensor's core virtualization map or the grid's inverse map on the grid hasn't been accidentally dropped somewhere. 
2. Additionally we don't currently use virtual grid index maps to reorganize grids that already fit into the compute grid, but we may want to in the future.

lib/Dialect/D2M/IR/D2MOps.cpp
```diff
@@ -1630,7 +1629,25 @@ d2m::GenericOp::getOperandGridShapes() {
     } else {
       auto tensorType = mlir::cast<RankedTensorType>(operand.getType());
       ttcore::MetalLayoutAttr layout =
-          mlir::cast<ttcore::MetalLayoutAttr>(tensorType.getEncoding());
+          mlir::dyn_cast_if_present<ttcore::MetalLayoutAttr>(
+              tensorType.getEncoding());
+
+      // If operand has ViewLayoutAttr (from StreamLayoutOp), get layout from
+      // storage.
```
bgrady-tt: I think this should (if possible) be replaced with code that checks that defining op of the operand to retrieve the layout.

lib/Dialect/D2M/IR/D2MOps.cpp
nsmithtt: I'd go a step further, I'd say we should remove the concept of a "needsVirtualization" all together, that is, **all grids are virtual grids**.  Some virtual grids might happen to be identity with the underlying physical grid.  I agree with your points @bgrady-tt, and it aligns with my statement, an identity virtualized grid also has an identity inverse map.

lib/Dialect/D2M/IR/D2MOps.cpp
nsmithtt: I think I introduced this, but I support the removal for collapse impl.

lib/Dialect/D2M/IR/D2MOps.cpp
nsmithtt: I think I introduced this, but I support the removal for collapse impl.

lib/Dialect/D2M/IR/D2MOps.cpp
```diff
@@ -979,8 +953,33 @@ void d2m::GenericOp::build(
   llvm::SmallVector<Type> blockTypes =
       llvm::map_to_vector(TypeRange(state.operands), [&](Type t) -> Type {
         mlir::RankedTensorType tensorType = mlir::cast<RankedTensorType>(t);
-        auto layout =
-            mlir::cast<ttcore::MetalLayoutAttr>(tensorType.getEncoding());
+        auto layout = mlir::dyn_cast_if_present<ttcore::MetalLayoutAttr>(
+            tensorType.getEncoding());
+
+        // If the operand has ViewLayoutAttr (from StreamLayoutOp), get the
+        // layout from storage
```
nsmithtt: The storage shard shape must equal the generic op's operand shard shape so the original code seems more correct to me, i.e. not having to traverse the IR to verify the correct shapes are already encoded on the generic's operands.

lib/Dialect/D2M/IR/D2MOps.cpp
```diff
@@ -1630,7 +1629,25 @@ d2m::GenericOp::getOperandGridShapes() {
     } else {
       auto tensorType = mlir::cast<RankedTensorType>(operand.getType());
       ttcore::MetalLayoutAttr layout =
-          mlir::cast<ttcore::MetalLayoutAttr>(tensorType.getEncoding());
```
nsmithtt, line 1633: Same comment as above, this removed bit seems more correct to me.

lib/Dialect/D2M/IR/D2MOps.cpp
```diff
@@ -1653,8 +1670,25 @@ d2m::GenericOp::getOperandShardShapes(bool convertTileToScalar) {
       elementType = memrefType.getElementType();
     } else {
       auto tensorType = mlir::cast<RankedTensorType>(shapedType);
-      layout = mlir::cast<mlir::tt::ttcore::DeviceLayoutInterface>(
```
nsmithtt, line 1656: Same as above.

lib/Dialect/D2M/IR/D2MOpsInterfaces.cpp
```diff
@@ -35,11 +35,19 @@ mlir::tt::d2m::applyViews(mlir::Operation *op) {
   Value input = viewOp.getInput();
   auto inputMemref = mlir::cast<mlir::MemRefType>(input.getType());
 
+  // Recursively apply view composition if the input is also a view. This
+  // handles nested view chains by composing affine maps.
+  if (mlir::isa<ttcore::ViewLayoutAttr>(inputMemref.getLayout())) {
+    if (auto *inputOp = input.getDefiningOp()) {
```
nsmithtt, line 41: Perhaps this should be an assert.

lib/Dialect/D2M/Transforms/LowerToLayout.cpp
```diff
@@ -395,19 +198,324 @@ class D2MSplitCompoundLayoutRewriter : public OpRewritePattern<ToLayoutOp> {
...
+    bool bothTilized =
+        ttcore::isTiled(inputInfo.type) && ttcore::isTiled(outputInfo.type);
+
+    AffineMap viewMap;
+
+    if (isSimpleReblocking && bothTilized) {
+      // Fast path: pure grid reblocking on tilized tensors.
+      // Use calculateReblockMap which works directly on device shapes without
+      // going through logical space (avoids tile alignment issues).
+      llvm::errs() << "      lowerMappingChange: simple reblocking (tilized)\n";
```
nsmithtt: Stray prints in this file.

lib/Dialect/D2M/Transforms/LowerToLayout.cpp
```diff
@@ -425,114 +533,338 @@ class D2MSplitCompoundLayoutRewriter : public OpRewritePattern<ToLayoutOp> {
...
+    // 8. L1→DRAM
+    if (currentInfo.hasLayout() && !currentInfo.isDRAM() &&
+        targetInfo.hasLayout() && targetInfo.isDRAM()) {
+      llvm::errs() << "  8. L1→DRAM\n";
+      currentValue = lowerDatamovementGeneric(rewriter, currentValue,
+                                              op.getOutput(), op.getLoc());
+      currentInfo = TensorInfo::from(currentValue);
     }
 
+    // 9. VIRTUAL GRID COLLAPSE: If current has virtual grid but target doesn't
```
nsmithtt: Should this happen before L1 -> DRAM?

lib/Dialect/D2M/Transforms/LowerToLayout.cpp
```diff
@@ -425,114 +533,338 @@ class D2MSplitCompoundLayoutRewriter : public OpRewritePattern<ToLayoutOp> {
...
+          currentInfo.getGridShape(), tileShape);
+      auto tiledType = RankedTensorType::get(
+          deviceShape, targetInfo.type.getElementType(), *currentInfo.layout);
+      auto tiledEmpty = createEmpty(tiledType);
+      currentValue = lowerFormatConversionGeneric(rewriter, currentValue,
+                                                  tiledEmpty, op.getLoc());
+      currentInfo = TensorInfo::from(currentValue);
+    }
+
+    // 5. MAPPING CHANGE: Grid/index_map/logical_shape/dim_alignments (after
```
nsmithtt: Seems like this should happen before tilize?  The general case is that you'll need to untilize to do this transformation.

lib/Dialect/D2M/Utils/AffineMapUtils.cpp
```diff
@@ -0,0 +1,597 @@
...
+
+// Build complete layout transformation from one layout to another.
+// Strategy: Use buildDeviceToLogicalMap for both layouts, which already handles
+// existing index maps correctly.
+mlir::AffineMap
+buildLayoutTransformMap(mlir::tt::ttcore::MetalLayoutAttr fromLayout,
+                        mlir::RankedTensorType fromType,
+                        mlir::tt::ttcore::MetalLayoutAttr toLayout,
+                        mlir::RankedTensorType toType) {
+  llvm::errs() << "  fromLayout: " << fromLayout << "\n";
```
nsmithtt: Stray prints.

lib/Dialect/D2M/IR/D2MOps.cpp
vwellsTT: So I think there's some special casing involved w.r.t. the transfer from host to device logic for virtual grids that Brett added, s.t. a 1x64 tensor gets first transferred to a 1x8 core grid, and then from there to an 8x8.  That's what "needsVirtualization" is referencing here (perhaps unclearly).
I guess defer to Brett on the specifics of if we can get rid of that etc, but for this PR, I don't want to make any big changes to virtual grid.
All that said, what would "attaching inverse map" actually imply here; emitting a (potentially unused) view op similar to @nsmithtt 's changes elsewhere? Or adding a new field to the layout or something?

lib/Dialect/D2M/IR/D2MOpsInterfaces.cpp
vwellsTT: Because folding ought to guarantee this isn't possible, or what? At some stage I did need this change, but can experiment w/ undoing if you prefer.  Not 100% sure I see a problem with recursive approach off the top of head though

lib/Dialect/D2M/IR/D2MOps.cpp
vwellsTT: yeah, I think all 3 of these cases were some debugging attempts with streams I should remove, will double check

lib/Dialect/D2M/Transforms/LowerToLayout.cpp
vwellsTT: I'm not really sure tbh, don't think we exercise this case today.  Ideally I'd like to keep the case structure in a symmetric V shape (so if virtual grid expansion comes before dram -> l1 on the way in, virtual grid collapse must come AFTER l1 -> dram on the way out).  Perhaps both are wrong? not clear to me how virtual grid + dram would interplay today, if at all.

lib/Dialect/D2M/Transforms/LowerToLayout.cpp
vwellsTT: so in the complex case we need to untilize, remap, and retilize, we do that inline below (L#649).  This is a bit ugly, but it does obviate need to introduce `bounce`s back into the code, which I was hoping to avoid completely.  Lmk what you think.

lib/Dialect/D2M/Transforms/GenericTileComputeLoops.cpp
```diff
@@ -6,6 +6,7 @@
 #include "ttmlir/Dialect/D2M/Transforms/Passes.h"
 #include "ttmlir/Dialect/D2M/Utils/Utils.h"
 
+#include "ttmlir/Dialect/D2M/Utils/AffineMapUtils.h"
```
vmilosevic: :warning: **llvm\-include\-order** :warning:
```suggestion

```

lib/Dialect/D2M/Transforms/GenericTileComputeLoops.cpp
```diff
@@ -6,6 +6,7 @@
...
 #include "mlir/Dialect/Linalg/Transforms/Transforms.h"
 #include "mlir/Transforms/WalkPatternRewriteDriver.h"
```
vmilosevic, line 12: :warning: **llvm\-include\-order** :warning:
```suggestion
#include "ttmlir/Dialect/D2M/Utils/AffineMapUtils.h"

```

lib/Dialect/D2M/IR/D2MOps.cpp
nsmithtt: I see, OK I misunderstood what that was needed for.

lib/Dialect/D2M/IR/D2MOpsInterfaces.cpp
nsmithtt: I think the only way to get a ViewLayoutAttr is to come from a view_layout or stream_layout op, unless there's a test or something that doesn't do this.

lib/Dialect/D2M/Transforms/LowerToLayout.cpp
nsmithtt: I see, I somewhat prefer the bounce routine personally, but there are different tradeoffs I suppose. Won't block the PR on it.

lib/Dialect/D2M/Transforms/LowerToLayout.cpp
nsmithtt: My thinking is that if we previously had it in L1 from one of the above handlers it'd make sense to keep it in L1 if we plan to continue working with it.  Going to DRAM or host should probably be the last thing you do.

lib/Dialect/D2M/Transforms/LowerToLayout.cpp
vwellsTT: Yeah, I think I see both sides, bouncing has more graceful code patterns in general, but more chances of nasty infinite recursion.  I sort of think in the long run we might want to bounce on some specific cases (this, DRAM), and generally use a decision tree like this for simple, unbounce-able cases?
Will likely leave this as-is for now and slate that for follow-up, if there aren't objections--I find this slightly easier to debug etc. for now

lib/Dialect/D2M/Transforms/LowerToLayout.cpp
vwellsTT: that makes sense, though I think the "virtualizing" is also a pretty final transform (it's clump together on a Nx1 or 1xN s.t. we can easily transfer to host, should not be used in this state for anything else).  I don't have a strong preference on one being "more" final than the other I guess, I don't think we exercise virtual grid + dram in our tests today

lib/Dialect/D2M/IR/D2MOpsInterfaces.cpp
vwellsTT: Tbh, I'm not following you here, will likely merge this week and follow-up next week once we've synced I guess.
>I think the only way to get a ViewLayoutAttr is to come from a view_layout or stream_layout op, unless there's a test or something that doesn't do this.

Right, but what prevents you from coming from a view of a view of a view...? (or view of a stream of a view, to avoid folding)? In such cases, I think you need this logic to recursively trace to the "root" MetalLayoutAttr

jmcgrathTT (approved):

include/ttmlir/Dialect/D2M/Utils/AffineMapUtils.h
```diff
@@ -0,0 +1,110 @@
+// SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
```
vwellsTT: will elim this and move to include/ttmlir/AffineMapUtils.h I think

PR #5542, opened 2025-10-28
[D2M] Fix bug in GridSelection

vwellsTT:
### Problem description
In my GridSelection changes, I ensured that the actual grid selection used a squared grid, but missed passing this to the dim_alignments calculation, which broke `test_metal_matmul.py` on non-square grid chips, apparently.

### What's changed
Quick clean-up to ensure we always use the square-ified grid for grid decisions in GridSelection.cpp (including dim_alignment calculations).

### Checklist
- [X] Manually confirmed this fixes issues on a 8x10 WH locally--not sure how to make sure this is exercised in CI

### DIFF
--- a/test/python/golden/test_metal_matmul.py
+++ b/test/python/golden/test_metal_matmul.py
@@ -110,7 +110,6 @@ def test_matmul_multi_core_8otpc(m: int, k: int, n: int, target: str, request, d
 def test_matmul_p150_grid_selection_compile_only(target: str, request, device):
     from builder.base.builder_utils import compile_ttir_to_flatbuffer
 
-    tile_size = 32
     lhs = (1024, 1024)
     rhs = (1024, 1024)
 
--- a/test/ttmlir/Silicon/TTMetal/n150/simple_transpose.mlir
+++ b/test/ttmlir/Silicon/TTMetal/n150/simple_transpose.mlir
@@ -10,11 +10,3 @@ func.func @transpose(%arg0: tensor<512x256xf32>) -> tensor<256x512xf32> {
   // CHECK: call_opaque "transpose_wh_tile"
   return %1 : tensor<256x512xf32>
 }
-
-func.func @transpose_square(%arg0: tensor<512x512xf32>) -> tensor<512x512xf32> {
-  %0 = ttir.empty() : tensor<512x512xf32>
-  %1 = "ttir.transpose"(%arg0, %0) <{dim0 = 0 : si32, dim1 = 1 : si32}> : (tensor<512x256xf32>, tensor<512x512xf32>) -> tensor<512x512xf32>
-  // CHECK: call_opaque "transpose_wh_init"
-  // CHECK: call_opaque "transpose_wh_tile"
-  return %1 : tensor<512x512xf32>
-}


wenbinlyuTT (commented):
Changes LGTM, but maybe we can add a lit test that overrides the device shape to be non-square and verify we still use the square grid?

vwellsTT:
> Changes LGTM, but maybe we can add a lit test that overrides the device shape to be non-square and verify we still use the square grid?

Yeah, I'm debating how we can add coverage for this.  Apparently we can't run matmul on BH yet, but maybe just a general matmul non-execution lit-test would cover this sort of case? I guess overriding looks like it would work too, it's not 100% clear to me if we ought to apply square-fying on top of an override or not though (could see an argument the override should override the square-ify logic instead--I do think today the square-ify override the override though)

wenbinlyuTT:
Good point on what exactly should override override...
The failure in `test_metal_matmul.py` is value error right? Not compilation or running.

vwellsTT:
I think it's what I'd call a `compilation` error:
```
 builder.base.builder_utils.TTBuilderCompileException: Failure while executing pass pipeline:
E           error: "/localdev/jdesousa/tt-mlir/build/python_packages/builder/base/builder.py:311:id(0)": 'd2m.generic' op shard shape mismatch between operand[0] shard_shape=[160, 128] and operand[1] shard_shape=[160, 128] at affine dim d2
E            note: "/localdev/jdesousa/tt-mlir/build/python_packages/builder/base/builder.py:311:id(0)": see current operation:
E             %7 = "d2m.generic"(%2, %4, %6) <{block_factors = [1, 1, 8], grid = #ttcore.grid<8x8>, indexing_maps = [affine_map<(d0, d1, d2) -> (d0, d2)>, affine_map<(d0, d1, d2) -> (d2, d1)>, affine_map<(d0, d1, d2) -> (d0, d1)>], iterator_types = [#ttcore.iterator_type<parallel>, #ttcore.iterator_type<parallel>, #ttcore.iterator_type<reduction>], operandSegmentSizes = array<i32: 2, 1>, threads = [#d2m.thread<compute>]}> ({
E             ^bb0(%arg2: !d2m.cb<tensor<5x4x!ttcore.tile<32x32, f32>>>, %arg3: !d2m.cb<tensor<5x4x!ttcore.tile<32x32, f32>>>, %arg4: !d2m.cb<tensor<5x4x!ttcore.tile<32x32, f32>>>):
E               %10 = "d2m.wait"(%arg2) : (!d2m.cb<tensor<5x4x!ttcore.tile<32x32, f32>>>) -> tensor<5x4x!ttcore.tile<32x32, f32>>                                                                                                                     E               %11 = "d2m.wait"(%arg3) : (!d2m.cb<tensor<5x4x!ttcore.tile<32x32, f32>>>) -> tensor<5x4x!ttcore.tile<32x32, f32>>
E               %12 = "d2m.reserve"(%arg4) : (!d2m.cb<tensor<5x4x!ttcore.tile<32x32, f32>>>) -> tensor<5x4x!ttcore.tile<32x32, f32>>
E               %13 = "linalg.generic"(%10, %11, %12) <{indexing_maps = [affine_map<(d0, d1, d2) -> (d0, d2)>, affine_map<(d0, d1, d2) -> (d2, d1)>, affine_map<(d0, d1, d2) -> (d0, d1)>], iterator_types = [#linalg.iterator_type<parallel>, #linalg.iterator_type<parallel>, #linalg.iterator_type<reduction>], operandSegmentSizes = array<i32: 2, 1>}> ({
E               ^bb0(%arg5: !ttcore.tile<32x32, f32>, %arg6: !ttcore.tile<32x32, f32>, %arg7: !ttcore.tile<32x32, f32>):                                                                                                                              E                 %14 = "d2m.tile_matmul"(%arg5, %arg6, %arg7) : (!ttcore.tile<32x32, f32>, !ttcore.tile<32x32, f32>, !ttcore.tile<32x32, f32>) -> !ttcore.tile<32x32, f32>
E                 "linalg.yield"(%14) : (!ttcore.tile<32x32, f32>) -> ()
E               }) : (tensor<5x4x!ttcore.tile<32x32, f32>>, tensor<5x4x!ttcore.tile<32x32, f32>>, tensor<5x4x!ttcore.tile<32x32, f32>>) -> tensor<5x4x!ttcore.tile<32x32, f32>>
E               "d2m.yield"(%13) : (tensor<5x4x!ttcore.tile<32x32, f32>>) -> ()
E             }) : (tensor<8x8x5x4x!ttcore.tile<32x32, f32>, #ttcore.metal_layout<logical_shape = 1024x1024, dim_alignments = 320x256, collapsed_intervals = dense<[[0, 1], [1, 2]]> : tensor<2x2xi64>, undef, l1>>, tensor<8x8x5x4x!ttcore.tile<32x32
, f32>, #ttcore.metal_layout<logical_shape = 1024x1024, dim_alignments = 320x256, collapsed_intervals = dense<[[0, 1], [1, 2]]> : tensor<2x2xi64>, undef, l1>>, tensor<8x8x5x4x!ttcore.tile<32x32, f32>, #ttcore.metal_layout<logical_shape = 1024x1024, dim_alignments = 320x256, collapsed_intervals = dense<[[0, 1], [1, 2]]> : tensor<2x2xi64>, undef, l1>>) -> tensor<8x8x5x4x!ttcore.tile<32x32, f32>, #ttcore.metal_layout<logical_shape = 1024x1024, dim_alignments = 320x256, collapsed_intervals =
 dense<[[0, 1], [1, 2]]> : tensor<2x2xi64>, undef, l1>>
 ```
 but not 100% sure what you mean by ValueError vs compilation tbh

wenbinlyuTT:
By value error I mean incorrect results.
Currently the 3rd and the 4th of the matmul tests are marked as skipping the p150 so I guess it's the `test_matmul_multi_core_8otpc` that is failing compilation? `test_matmul_single_core_8otpc` is a single core test.
Let me just grab a BH node and see what's going on.

vwellsTT:
> I think I understood the nature of the original bug... is it possible to add a dedicated pass lit test for this particular edge (?) case ?

This is being discussed above :)

vwellsTT:
>Currently the 3rd and the 4th of the matmul tests are marked as skipping the p150 so I guess it's the test_matmul_multi_core_8otpc that is failing compilation? test_matmul_single_core_8otpc is a single core test.

Ah, actually Jacob flagged this issue on a n150 with an 8x10 grid, and I repro'ed it on the same (after searching quite a bit for such a card on ird...).  I think the smaller testcases don't run into the alignment boundaries => they pass on p150 and n150 regardless of grid.

wenbinlyuTT (approved):
So I ran the test on BH, indeed the two smaller tests are fine, it's the two larger ttnn shapes tests are failing (3rd & 4th), but they're already marked as skipped on BH so CI didn't catch this.

I have some good news and some bad ones:

Good: this PR fixes matmul compilation on BH so the marker at https://github.com/tenstorrent/tt-mlir/blob/93961a5cdf9b4155edd6c55f8c37ea197759175f/test/python/golden/test_metal_matmul.py#L107 and in the next test can probably be removed. By doing this we have coverage of this in the CI.

Bad: the test hangs on BH during flatbuffer execution so we can't remove the markers yet, and currently we lack the ability to tell a pytest to compile but not run on BH

- There is a `skip_exec` argument to all the `compile_and_execute_*` functions, but from inside the test there's no good way to know if we're running on BH (or know if we're running a square grid).

- We can duplicate the 3rd and the 4th tests and replace `compile_and_execute_ttir` with `compile_ttir_to_flatbuffer` to only test compilation, until the `skip_config` mark equivalent for `skip_exec` is added to builder utils.

Neither solution is perfect, because the existence of 8x10 WH...

Overall I think this can be merged as-is.

vwellsTT:
@wenbinlyuTT oh nice, thanks!  Yeah, that makes sense, I guess in the long run it'd be nice to have a more general way to handle "skip exec on X arch" anyway.
Overall, do you think it'd be better if I add something like the 2nd approach to this PR as a stopgap, or just merge as-is? I could see arguments either way

wenbinlyuTT:
I think we can add a single, simple 512x512x512 one that just test the compilation (make sure it fails on main and passes with this fix). Should be quick.

vroubtsovTT (approved):

vwellsTT:
Previously, this sort of thing would've been caught by my `simple_transpose.mlir` test in `Silicon/TTMetal/n150`--it seems we no longer run these (or any other lit) tests on p150 after the refurbishments.
I guess given this, it's probably easiest to just add a trivial case to the `test_metal_matmul.py` itself instead of making a new lit target for p150, so I'll go ahead and do that.

Issue #5547, opened 2025-10-29
Bufferize tensors with MetalLayout attributes


brnorris03:
Currently tensors with MetalLayout attributes are not automatically bufferized to MemRefType with an appropriate attribute; as a result, information (e.g., MemorySpace) is lost.

PR #5548, opened 2025-10-29
Add TTCoreOneShotBufferizePass with MetalLayout support

brnorris03:
### Ticket
Closes https://github.com/tenstorrent/tt-mlir/issues/5547

## Problem description

The default MLIR one-shot bufferization pass doesn't properly handle TT-MLIR's `MetalLayoutAttr` tensor encodings, effectively losing the layout information: sharded vs interleaved, memory space (L1, DRAM), and dimension alignments.

Additionally, `ttcore.global` tensor operations and `d2m.stream_layout` view operations were not being properly converted during bufferization.

## What's changed

We implement `TTCoreOneShotBufferizePass` which extends MLIR's one-shot bufferization framework by providing a specialized type converter (`functionArgTypeConverterFn`) that preserves `MetalLayoutAttr` semantics when converting tensors to memrefs in device modules.

Key changes:
- Converts tensors with `MetalLayoutAttr` to memrefs with appropriate layout attributes (`ShardLayoutAttr`, `InterleavedLayoutAttr`, or `ViewLayoutAttr`) and explicit `MemorySpaceAttr` in (nested) modules inside device modules.
- Properly handles function boundaries and `ttcore.global` operations.
- Fixed `getBufferType` in `Utils.cpp` to use `getIndexAffineMapOrIdentity()` instead of `getIndexAffineMap()`, allowing view-type buffers to use identity affine maps when no explicit transformation is specified.
- Updated the TTMetal pipeline to use the new pass for non-TTNN mode. The base (builtin) one-shot bufferization is configured with the same options as previously used in the TTMetal pipeline.

By extending one-shot bufferization rather than implementing custom bufferization passes we leverage existing alias analysis, avoid unnecessary copies, can integrate seamlessly with other dialects via `BufferizableOpInterface`, and automatically benefit from future MLIR bufferization improvements. Also it's easily extensible by modifying the custom type converter.

### Checklist

### DIFF
--- a/test/ttmlir/Dialect/TTCore/Transforms/one_shot_bufferize_fallback.mlir
+++ b/test/ttmlir/Dialect/TTCore/Transforms/one_shot_bufferize_fallback.mlir
@@ -77,7 +77,7 @@ module attributes {ttcore.system_desc = #system_desc} {
 
 // -----
 // Test 4: Function argument with unranked tensor.
-// This tests the fallback path for unranked tensors (lines 23-26 in BufferizationTypeConverter.h).
+// This tests the fallback path for unranked tensors
 
 #system_desc = #ttcore.system_desc<[{role = host, target_triple = "x86_64-pc-linux"}], [{arch = <wormhole_b0>, grid = 8x8, coord_translation_offsets = 18x18, l1_size = 1499136, num_dram_channels = 12, dram_channel_size = 1073741824, noc_l1_address_align_bytes = 16, pcie_address_align_bytes = 32, noc_dram_address_align_bytes = 32, l1_unreserved_base = 101152, erisc_l1_unreserved_base = 98304, dram_unreserved_base = 32, dram_unreserved_end = 1073177056, supported_data_types = [<f32>, <f16>, <bf16>, <bfp_f8>, <bfp_bf8>, <bfp_f4>, <bfp_bf4>, <bfp_f2>, <bfp_bf2>, <u32>, <u16>, <u8>, <si32>], supported_tile_sizes = [ 4x16,  16x16,  32x16,  4x32,  16x32,  32x32], dst_physical_size_tiles = 16, num_cbs = 32, num_compute_threads = 1, num_datamovement_threads = 2}], [0], [0 : i32], [ 0x0x0x0]>
 


lib/Dialect/D2M/Transforms/Allocate.cpp
```diff
@@ -32,15 +32,8 @@ namespace mlir::tt::d2m {
 // Helper definitions.
 //===----------------------------------------------------------------------===//
 
-inline ttcore::MemorySpace getMemorySpace(MemRefType memref,
-                                          ttcore::MemorySpace dflt) {
-  auto memSpace = memref.getMemorySpace();
-  return memSpace ? mlir::cast<ttcore::MemorySpaceAttr>(memSpace).getValue()
-                  : dflt;
-}
-
```
vroubtsovTT: I have an allocate PR in flight and would like to avoid a non-essential merge conflict here. Could you add the utils method but not make Allocate.cpp changes for the time being?

include/ttmlir/Dialect/TTCore/IR/TTCoreOpsTypes.h
```diff
@@ -379,8 +379,19 @@ inline MemorySpace getMemorySpace(MemorySpaceAttr memorySpaceAttr) {
...
+          mlir::dyn_cast_if_present<MemorySpaceAttr>(memref.getMemorySpace())) {
+    return memSpaceAttr.getValue();
+  }
+  return dflt;
+}
+
+// Convenience overload with System as default.
 inline MemorySpace getMemorySpace(MemRefType memref) {
-  return getMemorySpace(mlir::cast<MemorySpaceAttr>(memref.getMemorySpace()));
+  return getMemorySpace(memref, MemorySpace::System);
```
vroubtsovTT: perhaps this convenience method overload is redundant if we change the first method to have a parameter default a la `getMemorySpace(MemRefType memref, MemorySpace dflt = MemorySpace::System)`?

include/ttmlir/Dialect/TTCore/Transforms/BufferizationTypeConverter.h
```diff
@@ -0,0 +1,44 @@
...
+inline void setTTCoreBufferizationTypeConverter(
+    bufferization::OneShotBufferizationOptions &options) {
+  options.functionArgTypeConverterFn =
+      [](TensorType tensorType, Attribute memorySpace, func::FuncOp funcOp,
+         const bufferization::BufferizationOptions &options) -> BaseMemRefType {
+    auto rankedTensorType = mlir::dyn_cast<RankedTensorType>(tensorType);
+    if (!rankedTensorType) {
+      // Fallback to default conversion for unranked tensors
+      return MemRefType::get(tensorType.getShape(), tensorType.getElementType(),
+                             AffineMap(), memorySpace);
```
vroubtsovTT: if this tensor is unranked, would `UnrankedMemRefType::get()` be the appropriate choice here?

vroubtsovTT (approved):
thanks for adding extra comments in lit tests explaining what the intent of each test is!

However, I am not sure what you refer to in in your description bullet
> Enable D2M allocate pass to handle cases where indexing maps haven't been generated yet.

as the only Allocate.cpp change seems to be to move the helper memspace getter to Utils.h... Is it using `dyn_cast_if_present()` ?

include/ttmlir/Dialect/TTCore/Transforms/BufferizationTypeConverter.h
brnorris03: I think that unranked support may have to be added separately when/if needed. `getBufferType` itself does not support unranked, hance the default fallback where the memory space at least is preserved. So it's not a straightforward mapping from `MetalLayout` (which tbh, I am not really sure how to even define for unranked) to memref layouts, which is all that `getBufferType` supports atm.

brnorris03: Oh, sorry, I think I misunderstood what you said -- yes, absolutely, great catch! Adding a test for that, too (didn't have one).

vroubtsovTT: I am not 100 %sure, tbh, but passing `AffineMap()` to MemRefType::get() looked strange to me... if we wanted a memref type without an explicit/non-identity mapping, would passing `nullptr` have been just as good/better?

brnorris03: I changed it to `return UnrankedMemRefType::get(tensorType.getElementType(), memorySpace);` which I think is the best we can do with what's available in an UnrankedTensorType

brnorris03:
> thanks for adding extra comments in lit tests explaining what the intent of each test is!
> 
> However, I am not sure what you refer to in in your description bullet
> 
> > Enable D2M allocate pass to handle cases where indexing maps haven't been generated yet.
> 
> as the only Allocate.cpp change seems to be to move the helper memspace getter to Utils.h... Is it using `dyn_cast_if_present()` ?

No, initially I also had an early return from `getOperandContexts` when `genericOp.getIndexingMaps().empty()` is true, however, I removed it from this PR since it's not related and forgot to update the description!

include/ttmlir/Dialect/TTCore/IR/TTCoreOpsTypes.h
brnorris03: Yes, will do that instead. Was wondering if that default is okay?

include/ttmlir/Dialect/TTCore/Transforms/BufferizationTypeConverter.h
brnorris03: To answer the question -- I think in that particular case, creating an empty AffineMap would have been the way to go (not nullptr, which would possibly trigger various asserts/exceptions later).

runtime/include/tt/runtime/detail/distributed/controller/command_factory.h
```diff
@@ -63,6 +63,13 @@ class CommandFactory {
       const std::vector<uint32_t> &shape, const std::vector<uint32_t> &stride,
       uint32_t itemSize, ::tt::target::DataType dataType);
 
+  static uint64_t buildCreateMultiDeviceHostTensorFromShardsCommand(
```
jnie-TT, line 66: Hmmm I don't think there are actually any new changes in runtime, this seems like some sort of weird rebase miss. Looking at the new commits on this branch I see many commits that are already in main. @brnorris03 could you please rebase your changes on top of main to remove these duplicate commits.

arminaleTT (approved):
The ttnn-jit test diff is probably a rebase artifact.

@vtangTT how does this affect your PR for the ttnn_mode option?

include/ttmlir/Dialect/TTCore/Transforms/BufferizationTypeConverter.h
vroubtsovTT: I think in a number of places MLIR uses null ptrs (which is basically what many of their default-constructed types are) to indicate absence of attributes. For example, the following factory method
```
static MemRefType get(ArrayRef<int64_t> shape, Type elementType, MemRefLayoutAttrInterface layout = {}, Attribute memorySpace = {});
```
would seem to be a better way to get a MemRefType with an implicit layout and a given memorySpace. The docs say
>in absence of an explicit layout, a memref is considered to have a multi-dimensional identity affine map layout. Identity layout maps do not contribute to the MemRef type identification and are discarded on construction. That is, a type with an explicit identity map is memref<?x?xf32, (i,j)->(i,j)> is strictly the same as the one without a layout, memref<?x?xf32>.

So I would have used `MemRefType::get(tensorType.getShape(), tensorType.getElementType(), {}, memorySpace)`  instead of passing a default-constructed AffineMap which I don't think we see in llvm/mlir codebase often.

include/ttmlir/Dialect/TTCore/IR/TTCoreOpsTypes.h
vroubtsovTT: The default of `System` seems ok to me.

runtime/include/tt/runtime/detail/distributed/controller/command_factory.h
brnorris03: Yeah, and also for CI -- I rebased successfully this time I think! All these are gone. Sorry for the distraction

vroubtsovTT:
> No, initially I also had an early return from `getOperandContexts` when `genericOp.getIndexingMaps().empty()` is true, however, I removed it from this PR since it's not related and forgot to update the description!

Got it. I would appreciate minimal changes to Allocate.cpp (basically, as you have them now) as I have a PR in flight that will have conflicts.

runtime/include/tt/runtime/detail/distributed/controller/command_factory.h
jnie-TT: no worries at all, thanks for the quick turnaround!

brnorris03:
> The ttnn-jit test diff is probably a rebase artifact.
> 
> @vtangTT how does this affect your PR for the ttnn_mode option?

This was very odd, I did a regular rebase and somehow picked up the unrelated commits (not mine); I did it again and they are gone now.

For ttnn_mode, the builtin one-shot-bufferization is used, but it may be good for you to confirm that the defaults are what you still want (I preserved them from before). https://github.com/tenstorrent/tt-mlir/pull/5548/files#diff-a1880c9e46b566423b57497104b058f12a89f5d1bea8be08319ac79a63f7ea87R23-R34

include/ttmlir/Dialect/TTCore/Transforms/BufferizationTypeConverter.h
brnorris03: I don't know. I think returning an unranked memref (as you originially suggested), rather than an incomplete ranked one makes a lot more sense. I think the Unranked to Ranked conversion does not belong in bufferization, but can be done after as a separate pass if that's needed.

vroubtsovTT: I don't know bufferization all that well, tbh, I was just responding to your if-branch being for an unranked tensor. I would be ok with either using UnrankedMemRefType::get as you show above or the MemRefType::get with an empty/null layout.

brnorris03: I do appreciate you catching this, it was a real bug and missing a test, too. Thank you!

zoecarver (approved):

PR #5578, opened 2025-10-29
Signal handler and pretty stack trace for pykernel crashes

zoecarver:
### Ticket
N/A

### Problem description
Before: if the compiler crashes in a pykernel test, you will get a system dialog / crash report. After: clean exit, stack trace, pass that failed. 

### What's changed
* Use LLVM tools to print a nice stack trace
* Instal signal handler to exit cleanly 

### Discussion 

MLIR and LLVM are compiled without rtti, but rtti is required for nanobind, so we have this inconsistency that leads to linker errors. This PR includes a little hack to get around this inconsistency (create new static library with a C function to bridge), but long term we might want to think about not mixing these two compilation modes.

### DIFF
--- a/python/PassTracker.cpp
+++ b/python/PassTracker.cpp
@@ -4,13 +4,13 @@
 //
 // PassTracker - Compiled with -fno-rtti to match MLIR
 
+#include "mlir-c/Pass.h"
+#include "mlir/IR/Operation.h"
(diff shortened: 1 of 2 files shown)


zoecarver:
Closing in favor of the no-fork one.

PR #5588, opened 2025-10-29
 Smattering of pykerenel changes for dsl

zoecarver:
### Ticket
N/A

### Problem description
Refactor pykernel to extract TTCompilerBase class, enabling tt-lang's D2MGenericCompiler to inherit shared compilation infrastructure. Includes bug fixes for MLIR Python bindings and enhanced Python DSL features.

### What's changed

**Core Refactoring:** extract `TTCompilerBase` parent class from `TTKernelCompiler`

**D2M Dialect Support**
  - Add `visit_Tuple` for multi-value returns
  - Support `RankedTensorType` in subscripting

**Python DSL**
  - Support `range()` with 1, 2, or 3 arguments
  - Keyword arguments in function calls
  - Qualified operator dispatch for dialect overloads

**Bug Fixes**
  - Handle missing MLIR context in standalone compilation
  - Cast for loop bounds to index type for `scf.for`
  - Continue to use `i32` for integer constants; adds casting to index where needed

Several features (async operations, RankedTensorType support, qualified operator dispatch) are D2M-specific and tested comprehensively in the new dsl tests where the full D2M pipeline is available. Tests here validate core refactoring and TTKernel backward compatibility.

This PR continues to use `i32` for integer constants to maintain compatibility with existing `TTKernel` ops (e.g., `ttkernel.TensorAccessorArgs` requires `I32` operands per its TableGen definition). However, this creates an inconsistency: D2M dialect ops (`d2m.core_index`, `d2m.iter_index`) require `I64Attr` for their dimension attributes. Long term, we should make these consistent, but that can be a follow on PR. To unblock the dsl, just add ` 64 if as_attr else 32` in `visit_Constant`. 

Thank you to Nick for the base commit!

### DIFF
--- a/tools/pykernel/_src/kernel_ast.py
+++ b/tools/pykernel/_src/kernel_ast.py
@@ -506,11 +506,11 @@ class TTCompilerBase(PyKernelAstBase):
         if isinstance(rhs, OpView):
             rhs = rhs.result
 
-        if hasattr(lhs, 'type') and isinstance(lhs.type, memref.MemRefType):
+        if hasattr(lhs, "type") and isinstance(lhs.type, memref.MemRefType):
             lhs = memref.LoadOp(
                 lhs, arith.ConstantOp(IndexType.get(self.ctx), 0)
             ).result
-        if hasattr(rhs, 'type') and isinstance(rhs.type, memref.MemRefType):
+        if hasattr(rhs, "type") and isinstance(rhs.type, memref.MemRefType):
             rhs = memref.LoadOp(
                 rhs, arith.ConstantOp(IndexType.get(self.ctx), 0)
             ).result
@@ -875,7 +875,7 @@ class TTKernelCompiler(TTCompilerBase):
         if isinstance(tile_id, OpView):
             tile_id = tile_id.result
 
-        if hasattr(tile_id, 'type') and isinstance(tile_id.type, IndexType):
+        if hasattr(tile_id, "type") and isinstance(tile_id.type, IndexType):
             tile_id = arith.index_cast(IntegerType.get_signless(32), tile_id)
 
         return ttkernel.noc_async_read_tile(tile_id, addr_gen, dst_addr)


zoecarver:
This is the other half of https://github.com/tenstorrent/tt-lang/pull/8 which may help for review

tools/pykernel/_src/kernel_ast.py
```diff
@@ -892,15 +834,178 @@ def populate_list(arr, _idx=[]):
...
             return op_constructor(IntegerType.get_signless(1, self.ctx), node.value)
         elif isinstance(node.value, int):
             return op_constructor(IntegerType.get_signless(32, self.ctx), node.value)
         else:
             raise NotImplementedError(
                 f"constant type {type(node.value).__name__} not implemented"
             )
+
+
+class TTKernelCompiler(TTCompilerBase):
```
brnorris03, line 856: I am not looking at this line by line on the premise that it is largely the same -- just relocated; if there are specific parts that changed that should be checked, please point them out.

tools/pykernel/_src/kernel_ast.py
```diff
@@ -79,17 +80,22 @@ def __init__(self, name, kernel_type=None, *args, **kwargs):
...
+            default_context = get_default_loc_context()
+        except ValueError:
+            default_context = None
+        self.ctx = default_context if default_context is not None else Context()
         self.cursor = Location.unknown(self.ctx)
         self.module = Module.create(self.cursor)
         self.insert_point = self.module.body
         self.func_entry = None
         self.symbol_tables = []
+        self.module_symbol_table = None
```
vtangTT, line 98: doens't seem to be used anywhere? delete?

tools/pykernel/_src/kernel_ast.py
```diff
@@ -598,39 +506,55 @@ def visit_BinOp(self, node):
...
             case ast.BitAnd():
-                return arith.andi(lhs, rhs)
+                return qualified_or("__and__", arith.andi, lhs, rhs)
             case ast.BitXor():
-                return arith.xori(lhs, rhs)
-            # case ast.Div(): # only worried about integers right now
-            # return arith.divf(lhs, rhs)
+                return qualified_or("__xor__", arith.xori, lhs, rhs)
             case _:
                 raise NotImplementedError(
```
vtangTT, line 559: maybe can replace this with the `unimplemented` helper?

tools/pykernel/_src/kernel_ast.py
zoecarver: Used by tt-lang:
```
$ rg module_symbol_table
python/ttlang/_src/codegen.py
231:    module_symbol_table: SymbolTable,
239:        module_symbol_table: Module-level symbol table
249:            if sym_name.value in f_params and sym_name.value in ct.module_symbol_table:
254:                module_symbol_table.insert(clone)

python/ttlang/d2m_api.py
265:        module_symbol_table = SymbolTable(module.operation)
267:            copy_symbol_table_globals(module_symbol_table, compiled_threads, f_params)

python/ttlang/_src/d2m_ast.py
155:        self.module_symbol_table = SymbolTable(self.module.operation)
201:                        self.module_symbol_table.insert(globalTensor.operation)
```

tools/pykernel/_src/kernel_ast.py
zoecarver: If the CI passes, I may do this as a follow up since it's not part of this PR

tools/pykernel/_src/kernel_ast.py
zoecarver: I think just the noc adapters.

PR #5592, opened 2025-10-29
[ttnn.jit] change `max_grid` to be from (1, 1) to (8,8)

vtangTT:
### Problem description
`max_grid being from (0, 0) to (7, 7) is confusing.

### What's changed
- changed it to be from (1, 1) to (8, 8)
- updated tests

### Checklist

### DIFF
--- a/test/ttnn-jit/test_eltwise.py
+++ b/test/ttnn-jit/test_eltwise.py
@@ -16,19 +16,19 @@ from utils import (
 )
 
 BLOCK_SHARDED_SHAPE_GRIDS = [
-    (32, 32, (0, 0)),
-    (32, 64, (0, 0)),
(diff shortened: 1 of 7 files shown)


arminaleTT (commented):
I don't think this is an improvement. Right now our `max_grid` matches the `end_coord` of the `coreRangeSet` for a sharded tensor. By changing this, the user will now need to manually subtract 1 from the x and y. E.g. see the `coreRangeSet`s here: https://github.com/tenstorrent/tt-metal/blob/main/tech_reports/tensor_sharding/tensor_sharding.md

vtangTT:
Agreed we should be matching with what metal/ttnn is doing.. closing.

Can reopen if needed

PR #5593, opened 2025-10-29
GridSelection recreates d2m.generic and preserves original thread type

anuragsingh-tt:
### Problem description
When the GridSelection pass recreates the d2m.generic, it uses the builder overload that defaults the thread type to compute, such that datamovement regions get rewritten as compute.

### What's changed
Title above.


vroubtsovTT (approved):
To unblock someone else's issue, this is fine.

PR #5596, opened 2025-10-30
[TTIR BUILDER] preserve bias in conv2d

ddesaiTT:
### Ticket
NA

### Problem description
`ttir.conv2d` was dropping the optional bias operand

### What's changed
- updated `ttir.conv2d`, it now preserves the bias operand
- updated `conv2d_golden`
- added `test_conv2d_no_bias`

### Checklist

### DIFF
--- a/tools/builder/ttir/ttir_builder.py
+++ b/tools/builder/ttir/ttir_builder.py
@@ -2941,9 +2941,10 @@ class TTIRBuilder(Builder):
         """
         if not bias:
             bias = None
+        golden_bias = self._get_golden_tensor(bias) if bias is not None else None
         return self._op_proxy(
             ttir.Conv2dOp,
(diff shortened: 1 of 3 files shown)


tools/builder/ttir/ttir_builder.py
```diff
@@ -2941,9 +2941,10 @@ def conv2d(
         """
         if not bias:
             bias = None
+        golden_bias = self._get_golden_tensor(bias) if bias is not None else None
         return self._op_proxy(
             ttir.Conv2dOp,
-            [in0, weight, bias],
```
tapspatel, line 2946: shouldn't we have the bias be passed?

ddesaiTT: passing as an kwarg later

Issue #5551, opened 2025-10-29
CI not passing on main branch


sdjordjevicTT:
Example failed job:
https://github.com/tenstorrent/tt-mlir/actions/runs/18893967931/job/53929837916

Failure message:
```
Setting latest tag on the image ghcr.io/tenstorrent/tt-mlir/tt-mlir-manylinux-ubuntu-22-04:dt-3ecfa947b29a6548e775508b0bae96a4844481a8521ddcb035d4abe990c27a6e
time="2025-10-29T02:04:51Z" level=fatal msg="initializing source docker://ghcr.io/tenstorrent/tt-mlir/tt-mlir-manylinux-ubuntu-22-04:dt-3ecfa947b29a6548e775508b0bae96a4844481a8521ddcb035d4abe990c27a6e: reading manifest dt-3ecfa947b29a6548e775508b0bae96a4844481a8521ddcb035d4abe990c27a6e in ghcr.io/tenstorrent/tt-mlir/tt-mlir-manylinux-ubuntu-22-04: manifest unknown"
```

sdjordjevicTT:
Reopening the issue, as we observed a new CI failure now:
https://github.com/tenstorrent/tt-mlir/actions/runs/18905257596/job/53963148308

test_layouts.py tests are failing on all_close check:
```
Running program:  abs  in runtime:  DeviceRuntime.TTNN
--------------------------------
Result:
ttnn.Tensor([[ 9.0234e-01,  4.8047e-01,  ...,  7.7637e-02,  4.9414e-01],
             [ 6.4062e-01,  1.0781e+00,  ...,  2.0938e+00,  1.2031e+00],
             ...,
             [ 3.3691e-02,  8.1250e-01,  ...,  2.2266e-01,  1.8047e+00],
             [ 1.2891e+00,  2.0312e+00,  ...,  1.2422e+00,  9.0820e-02]], shape=Shape([384, 192]), dtype=DataType::BFLOAT16, layout=Layout::TILE)
--------------------------------
Golden result:
ttnn.Tensor([[ 9.0234e-01,  4.8047e-01,  ...,  7.7637e-02,  4.9414e-01],
             [ 6.4062e-01,  1.0781e+00,  ...,  2.0938e+00,  1.2031e+00],
             ...,
             [ 2.2754e-01,  1.2266e+00,  ...,  2.4219e+00,  4.2773e-01],
[...]

PR #5607, opened 2025-10-30
Fix ttnn.jit python crash in CI

vtangTT:
### Ticket
https://github.com/tenstorrent/tt-mlir/issues/5551

### Problem description
python module initialization crashing
broken run: https://github.com/tenstorrent/tt-mlir/actions/runs/18939107901/job/54073713301
fixed run: https://github.com/tenstorrent/tt-mlir/actions/runs/18947585576

### DIFF
--- a/.github/test_scripts/pykernel.sh
+++ b/.github/test_scripts/pykernel.sh
@@ -13,14 +13,9 @@ ln -sf $INSTALL_DIR/tt-metal third_party/tt-metal/src/tt-metal
 if [ ! -d "$BUILD_DIR/python_packages/ttnn_jit" ]; then
     ln -sf $WORK_DIR/tools/ttnn-jit $BUILD_DIR/python_packages/ttnn_jit
 fi
-if [ ! -d "$BUILD_DIR/python_packages/ttnn_jit/runtime" ]; then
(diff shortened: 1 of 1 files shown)


.github/test_scripts/pykernel.sh
```diff
@@ -13,14 +13,9 @@ ln -sf $INSTALL_DIR/tt-metal third_party/tt-metal/src/tt-metal
 if [ ! -d "$BUILD_DIR/python_packages/ttnn_jit" ]; then
     ln -sf $WORK_DIR/tools/ttnn-jit $BUILD_DIR/python_packages/ttnn_jit
 fi
-if [ ! -d "$BUILD_DIR/python_packages/ttnn_jit/runtime" ]; then
```
nsumrakTT, line 16: if ttnn-jit wheel is not built this adds runtime package. Will this actually work?

vtangTT: yes, in https://github.com/tenstorrent/tt-mlir/blob/main/tools/ttnn-jit/_src/__init__.py it'll default to the except branch and import from mlir runtime whicih should be foudn in the pythonpath

PR #5612, opened 2025-10-30
elementwisefusion and insertdstregisteraccess skip generics with no c…

anuragsingh-tt:
…ompute ops

### What's changed
- Elementwise fusion: skip generics with no compute ops to avoid largest dst type assertion.
- InsertDstRegisterAccess: skip regions without compute ops before querying largest dst type.


lib/Dialect/D2M/Transforms/ElementwiseFusion.cpp
```diff
@@ -364,6 +364,22 @@ struct FuseD2MElementwiseOpsPattern : public OpRewritePattern<GenericOp> {
...
+          if (op->hasTrait<D2MGenericRegionComputeOpTrait>()) {
+            hasCompute = true;
+            return WalkResult::interrupt();
+          }
+          return WalkResult::advance();
+        })
+        .wasInterrupted();
+    if (!hasCompute) {
+      return failure();
+    }
```
nsmithtt: Maybe this should be a utility method on the generic op?

lib/Dialect/D2M/Transforms/ElementwiseFusion.cpp
```diff
@@ -364,6 +364,12 @@ struct FuseD2MElementwiseOpsPattern : public OpRewritePattern<GenericOp> {
     }
 
     assert(consumer.getNumRegions() == 1u);
+
+    // Skip regions that don't have compute ops.
+    if (!consumer.hasComputeOpsInRegion(0)) {
+      return failure();
```
wenbinlyuTT: Is this check needed? Coverage report at https://app.codecov.io/gh/tenstorrent/tt-mlir/pull/5612 shows that this is never hit at all even with your newly added tests.

It seems that the `isValidElementwiseFusionTarget()` check above has already intercepted the empty genericOp.

PR #5615, opened 2025-10-30
Force build_all=true

### LINKED ISSUE #5614: call-prepare-run.yml's selective component test through changed file analysis is unreliable
Change #5513 breaks the EmitC FullOp [build step in the speedy build](https://github.com/tenstorrent/tt-mlir/actions/runs/18951516586/job/54117178150). This was not noticed because the files changed in #5513 do not apparently match the criteria to run the EmitC build step, so it ends up being later, unrelated commits that do cause EmitC to build, to show this issue.

The above failing job was observed by checking out mlir main at commit for #5513 and forcing it to build all components.

jameszianxuTT:
### Ticket
Temporarily fixes #5614 
 
### Problem description

### What's changed

### Checklist

### DIFF
--- a/.github/workflows/call-prepare-run.yml
+++ b/.github/workflows/call-prepare-run.yml
@@ -121,12 +121,14 @@ jobs:
 
       - id: is-all
         run: |
-          if [[ "${{ steps.components.outputs.workflow_any_changed }}" == "true" ]]; then
-            echo "Detected changes in workflow or settings files, building all components"
-            echo "build_all=true" >> "$GITHUB_OUTPUT"
-          else
-            echo "build_all=${{ inputs.build_all }}" >> "$GITHUB_OUTPUT"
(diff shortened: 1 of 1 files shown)


jameszianxuTT:
Test run w/ emitC testgroup removed - https://github.com/tenstorrent/tt-mlir/actions/runs/18954160718

Still expected to fail in PR to main being cooked

jameszianxuTT:
@nsumrakTT - I am not sure how exactly this PR should look, but it seems that fails are slipping through into main because the current test impact analysis strategy doesn't provide enough coverage. 

I am not sure what that coverage would be, but it's too narrow right now, so I suggest disabling this test impact analysis feature entirely to prevent additional regressions while more knowledgeable stakeholders can determine which filesets can really have an impact on which tests.

Issue #5621, opened 2025-10-31
Add decomposition for rope in decode mode


mtopalovicTT:
If we have rope which has these inputs:
(1, 8, 1, 64),  # input
(1, 1, 64),  # cos input
(1, 1, 64),  # sin input

this will fail on pcc check because pytorch outputs shape 1, 8, 1, 64 but ttnn outputs 1, 8, 32, 64 (see https://github.com/tenstorrent/tt-metal/issues/31567). Add decomposition which will slice output of rope in case we are running in decode mode.

Interestingly this doesn't impact llama correctness.

sdjordjevicTT:
Completed with this one:
https://github.com/tenstorrent/tt-mlir/pull/8465

PR #5623, opened 2025-10-31
ttnn.rotary_embedding op + fusing

### LINKED ISSUE #5535: Fusing for rotary_embeding

mtopalovicTT:
This PR add `ttnn.rotary_embedding` op and fusing pattern for fusing this op.

Initially we added `ttnn.rotary_embedding_llama` op with hope that we are going to fuse RoPE into it, but it turns out that kernel implementation of that op supports only specific case of RoPE where pair wise rotation is done. `ttnn.rotary_embedding` peforms half rotation which is exactly what are we getting from current llama models that we are running.

RoPe formula in simple form looks like this:
`x * cos_cache + rotate_half(x) * sin_cache`

where x is input embedding and cos/sin cache are precomputed cos and sin matrices which execute the rotation. `rotate_half` function takes input embedding and split is in half at last dimension negates second half and concatenates negated second half and first half.

Tested that I manage to test:

* EmitPy - generation is working I also tested the code itself and it's working
* EmitC - working.  Tested using ttrt
* Tested using xla llama and tt-forge benchmarks. I can see the op is fused and PCC is good.
* Tested quen and llama isolated `apply_rotary_emb`

The thing that doesn't work as expected (look builder test) is that output shape of rope when we are in decode mode is wrong (more here https://github.com/tenstorrent/tt-mlir/issues/5621). This doesn't affect the output somehow but we should still fix it using decomposition until we get some from ttnn team. This is only problem when rope result is output of function and we try to copy it to host buffer, is where there is discrepancy between expected shape and ttnn tensor shape.

Sorry for this large PR but we wanted to get things out as fast as possible.

closes #5535

### DIFF
--- a/lib/Dialect/TTNN/Transforms/TTNNFusing.cpp
+++ b/lib/Dialect/TTNN/Transforms/TTNNFusing.cpp
@@ -422,8 +422,8 @@ public:
         TTNNConv2dWithActivation<ReluOp>, TTNNConv2dWithActivation<Relu6Op>,
         TTNNConv2dWithActivation<SiluOp>, TTNNConv2dWithActivation<SigmoidOp>,
         TTNNMatmulAndLinearWithActivation<MatmulOp, SigmoidOp>,
-        TTNNMatmulAndLinearWithActivation<LinearOp, SigmoidOp>,
-        RoPEFusing>(&getContext());
+        TTNNMatmulAndLinearWithActivation<LinearOp, SigmoidOp>, RoPEFusing>(
+        &getContext());
 
     GreedyRewriteConfig config;
     config.setUseTopDownTraversal(true);


lib/OpModel/TTNN/TTNNOpModel.cpp
```diff
@@ -2708,6 +2708,97 @@ llvm::Expected<size_t> OpModel<RotaryEmbeddingLlamaOp>::getOpRuntime(
...
+  }
+
+  ::ttnn::TensorSpec inputSpec = inputSpecExp.get();
+  ::ttnn::TensorSpec cosSpec = cosSpecExp.get();
+  ::ttnn::TensorSpec sinSpec = sinSpecExp.get();
+
+  auto rotaryEmbeddingOpQuery = [=]() {
+    return ::ttnn::graph::query_op_constraints(
+        ::ttnn::experimental::rotary_embedding, device, inputSpec, cosSpec,
+        sinSpec, tokenIndex, detail::getNullableMemoryConfig(outputLayout));
```
sgholamiTT, line 2749: OpModel should be implemented similar to [runtime](https://github.com/tenstorrent/tt-mlir/pull/5623/files#diff-885de6d0ead492e550f5070c6f5f178ad40a3de152ec1bc0fe3b4dc7a6692682R23-R26).
The op accepts a `compute_kernel_config` argument which is also supported in [metal](https://github.com/tenstorrent/tt-metal/blob/420cfbcda58bbe84558864a5c816cfb060bc7902/ttnn/cpp/ttnn/operations/experimental/transformer/rotary_embedding/rotary_embedding.hpp#L21C9-L21C98). Can we utilize that argument and pass it to the backend ?
I believe you can use `conversion::getDeviceComputeKernelConfig` for conversion too.
```c++
return ::ttnn::graph::query_op_constraints(
        ::ttnn::experimental::rotary_embedding, device, inputSpec, cosSpec,
        sinSpec, tokenIndex, detail::getNullableMemoryConfig(outputLayout),
        conversion::getDeviceComputeKernelConfig(deviceComputeKernelConfig));
```

lib/OpModel/TTNN/TTNNOpModel.cpp
```diff
@@ -2708,6 +2708,97 @@ llvm::Expected<size_t> OpModel<RotaryEmbeddingLlamaOp>::getOpRuntime(
...
+
+  ::ttnn::TensorSpec inputSpec = inputSpecExp.get();
+  ::ttnn::TensorSpec cosSpec = cosSpecExp.get();
+  ::ttnn::TensorSpec sinSpec = sinSpecExp.get();
+
+  // Create query closure
+  auto rotaryEmbeddingOpQuery = [=]() {
+    return ::ttnn::graph::query_op_runtime(
+        ::ttnn::experimental::rotary_embedding, device, inputSpec, cosSpec,
+        sinSpec, tokenIndex, detail::getNullableMemoryConfig(outputLayout));
```
sgholamiTT, line 2793: Similar here.

sgholamiTT (approved):
Thanks for doing this. OpModel changes look good to me. One minor comment inlined.
Please also add test in `TestOpModelLib.cpp`.
I think you need to rebase to main for CI to pass (?)

lib/OpModel/TTNN/TTNNOpModel.cpp
mtopalovicTT: Sure

mtopalovicTT:
> Thanks for doing this. OpModel changes look good to me. One minor comment inlined. Please also add test in `TestOpModelLib.cpp`. I think you need to rebase to main for CI to pass (?)

I can but again I don't see any value in adding test in TestOpModelLIb since I'm not testing different virtual grids/shard sizes buffers etc. At this moment I don't need that fine grain control. Let me know if you disagree.

lib/Dialect/TTNN/IR/TTNNOps.cpp
```diff
@@ -3782,6 +3784,55 @@ mlir::LogicalResult RotaryEmbeddingLlamaOp::verify() {
...
+
+  auto devicePredicate = [](mlir::RankedTensorType type) {
+    return ttnn::utils::isTensorOnDevice(type);
+  };
+
+  if (!llvm::all_of(inputTypes, devicePredicate)) {
+    return emitOpError("all input and output tensors must be on device.");
+  }
+
+  auto tilePredicate = [](mlir::RankedTensorType type) {
```
sdjordjevicTT: Should we handle this as a workaround? We can always transform there, right?

lib/Dialect/TTNN/IR/TTNNOps.cpp
```diff
@@ -3782,6 +3784,55 @@ mlir::LogicalResult RotaryEmbeddingLlamaOp::verify() {
...
+
+  SmallVector<RankedTensorType> inputTypes = {inputType, cosType, sinType,
+                                              outputType};
+
+  auto devicePredicate = [](mlir::RankedTensorType type) {
+    return ttnn::utils::isTensorOnDevice(type);
+  };
+
+  if (!llvm::all_of(inputTypes, devicePredicate)) {
+    return emitOpError("all input and output tensors must be on device.");
```
sdjordjevicTT: Same for this...

test/python/golden/ttir_ops/test_ttir_rope.py
```diff
@@ -0,0 +1,157 @@
+# SPDX-FileCopyrightText: (c) 2024 Tenstorrent AI ULC
```
sdjordjevicTT, line 1: Can you move this file under:
test/python/golden/ttir_ops/test_ttir_rope.py

test/python/golden/ttir_ops/test_ttir_rope.py
```diff
@@ -0,0 +1,157 @@
...
+            ],
+            # Mark as fail becase of https://github.com/tenstorrent/tt-metal/issues/31567
+            # I will follow up with decomposition to slice tensor until there is
+            # some resolution from metal team.
+            # Issue for decompositon https://github.com/tenstorrent/tt-mlir/issues/5621
+            marks=pytest.mark.xfail,
+        ),
+    ],
+)
+@pytest.mark.parametrize("dtypes", [[torch.bfloat16] * 3])
```
sdjordjevicTT, line 116: You can parametrize this test with:
@pytest.mark.parametrize("target", ["ttnn", "emitpy"])

With this, you are getting the same test for the emitpy...

test/ttmlir/EmitPy/transformer/rope.mlir
```diff
@@ -0,0 +1,13 @@
+// RUN: ttmlir-opt --ttir-to-emitpy-pipeline -o %t.mlir %s
```
sdjordjevicTT: With the above emitpy target in the builder, you don't need this one?

include/ttmlir/Dialect/TTIR/IR/TTIROps.td
```diff
@@ -5438,7 +5438,31 @@ def TTIR_ConcatenateHeadsOp: TTIR_NamedOp<"concatenate_heads"> {
   let hasVerifier = 1;
 }
 
+def TTIR_RotaryEmbeddingOp : TTIR_Op<"rotary_embedding"> {
```
sdjordjevicTT: Do we see a need for TTIR op? The fuising is done on the TTNN level?

mtopalovicTT: I wanted to test emitc/emitpy easily i.e I didn't want to have bunch of ops in the test which will be fused into single one just to test that lowering is working. I think this was the reason although I have a feeling that I had stronger reason I just cant remember it... Now thinking about it your comment makes sense to me. I remove it, we can use builder test anyway.

lib/Dialect/TTNN/IR/TTNNOps.cpp
mtopalovicTT: Done

lib/Dialect/TTNN/IR/TTNNOps.cpp
mtopalovicTT: Done

test/python/golden/ttir_ops/test_ttir_rope.py
mtopalovicTT: Done

test/python/golden/ttir_ops/test_ttir_rope.py
mtopalovicTT: Done

test/ttmlir/EmitPy/transformer/rope.mlir
mtopalovicTT: Done

include/ttmlir/Dialect/TTIR/IR/TTIROps.td
mtopalovicTT: Removed

sdjordjevicTT (approved):
This is great!

PR #5628, opened 2025-10-31
Added example modifications

FConstantinos:
Changes to the examples and some clearing up of semantics

### DIFF
--- a/test/pykernel/gen/colman_fused_eltwise_bcast.py
+++ b/test/pykernel/gen/colman_fused_eltwise_bcast.py
@@ -25,11 +25,14 @@ def eltwise_bcast(a_in, b_in, c_in, out, grid=None):
     # Parallelizing by columns here to get reuse on C
     cols_per_core = math.ceil(col_tiles / (grid[0] * grid[1]))
 
-    a_accessor = Stream(a_in, index_type=IndexType.TILE)
-    b_accessor = Stream(b_in, index_type=IndexType.TILE)
-    c_accessor = Stream(c_in, index_type=IndexType.TILE)
-    out_accessor = Stream(out, index_type=IndexType.TILE)
+    a_accessor = TensorAccessor(a_in, index_type=IndexType.TILE)
+    b_accessor = TensorAccessor(b_in, index_type=IndexType.TILE)
+    c_accessor = TensorAccessor(c_in, index_type=IndexType.TILE)
+    out_accessor = TensorAccessor(out, index_type=IndexType.TILE)
     
+    # NOTE: (Kostas) I don’t understand why a CircularBuffer needs to be associated with a tensor accessor.
+    #                Perhaps we need to know its specific type? Or to prevent mixups of tensors on the same cb?
+    # Whenever we wait()/reserve() on a CB, we wait for *shape* amount of data/space and get back a RingView pointer to the CB memory
     a_in_cb = CircularBuffer(a_accessor, shape=(granularity,1), buffer_factor=2)
     b_in_cb = CircularBuffer(b_accessor, shape=(granularity,1), buffer_factor=2)
     # NOTE: should it be declared somewhere that c_in_cb contains a row padded to tile,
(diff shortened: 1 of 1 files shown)


test/pykernel/gen/colman_fused_eltwise_bcast.py
```diff
@@ -45,13 +48,28 @@ async def compute():
         
         for ct in range(start_col_tile, end_col_tile):
             # Reuse C across rows of A, B
-            c_block = c_in_cb.wait().pop()
+            # The following waits until *shape* number of tiles (1 in this case) is available and returns a RingView object as defined in:
+            # https://github.com/tenstorrent/tt-lang/blob/main/python/sim/cbsim/ringview.py.
+            # TODO: Perhaps consider making RingView pointers that come from wait()/reserve() read/write only respectively?
+            c_block = c_in_cb.wait()
```
zoecarver, line 54: I think the current API is that we just call pop and then the wait is implicit.

test/pykernel/gen/colman_fused_eltwise_bcast.py
```diff
@@ -45,13 +48,28 @@ async def compute():
...
+                # again, these wait until *shape* number of tiles (*granularity*, i.e 4 in this case) are available and return RingView pointers:
+                a_block = a_in_cb.wait() # blocking 
+                b_block = b_in_cb.wait() # blocking
                 # NOTE: Please consider making non-approx the default for eltwise unary, but leave the option for the user to specify approx=True
+                out_block = out_cb.reserve() # blocking
+                # RingView operations, these also need to be defined
                 out_block = a_block * b_block + tt.tanh(c_block, approx=False)
-                out_cb.push(out_block)
+                # finalize push, this advances the cb pointers, the writing happened at the line above
+                out_cb.push()
```
zoecarver, line 64: Push also doesn't exist; this would also be a pop to signal data is ready

test/pykernel/gen/colman_fused_eltwise_bcast.py
```diff
@@ -45,13 +48,28 @@ async def compute():
...
+                # finalize pop, this advances the cb pointers, essentially freeing the memory
+                # After poping, the corresponding RingView(a_block) points to stale data.
+                # Should probably make it an error to access it at that point, at least in sim.
+                a_in_cb.pop()
```
zoecarver, line 68: This would be implicit IIUC

test/pykernel/gen/colman_fused_eltwise_bcast.py
```diff
@@ -25,11 +25,14 @@ def eltwise_bcast(a_in, b_in, c_in, out, grid=None):
     # Parallelizing by columns here to get reuse on C
     cols_per_core = math.ceil(col_tiles / (grid[0] * grid[1]))
 
-    a_accessor = Stream(a_in, index_type=IndexType.TILE)
-    b_accessor = Stream(b_in, index_type=IndexType.TILE)
-    c_accessor = Stream(c_in, index_type=IndexType.TILE)
-    out_accessor = Stream(out, index_type=IndexType.TILE)
+    a_accessor = TensorAccessor(a_in, index_type=IndexType.TILE)
```
zoecarver, line 28: Why did you change this from a `Stream`? I think `Stream` is correct here. IIUC TensorAccessor is for low-level NOC address calculations (see `matmul_multicore_demo.py`), while Stream is the abstraction for DMA operations between DRAM and L1. When using CircularBuffers with `dma()` calls, you need `Stream` not `TensorAccessor`.

FConstantinos: mainly because of the following comment and because I understand these from the usage here to be wrappers for accessing tensors (a_in, b_in etc are all torch tensors). I didn't know the other context, thanks for pointing out, I'll have a look. Either way I don't have a strong opinion on this.

https://github.com/tenstorrent/tt-mlir/commit/26df1ac1228c7a620981c81e6f1ded6a8cca6cdf#r167286010

zoecarver: @nsmithtt can you clarify Stream vs TensorAccessor in this example? I think I'm confused about TensorAccessor's semantics or maybe I am misunderstanding the example more generally.

nsmithtt: It was just a naming thing, we prefer TensorAccessor over Stream since it maps better to an existing tt-metal data structure of the same name.

test/pykernel/gen/colman_fused_eltwise_bcast.py
brnorris03, line 54: Does the Python interface allow the number of tiles to be specified (with a default of 1)?

test/pykernel/gen/colman_fused_eltwise_bcast.py
brnorris03: Pop advances the cb counter, and effectively invalidates `out_block`, I don't think that's how data-ready signaling is accomplished. The metal interface has push specifically for that purpose, and I don't think it's appropriate to use pop in this case -- a different thread should be popping this CB entry (e.g., dm).

Another thing I don't quite understand is why this has to be explicit. The compiler can automatically insert push after all operations involved in the assignment are completed. But I guess if we want users to control when signaling happens, then you need push (or something equivalent, but not pop, IMO).

PR #5629, opened 2025-10-31
[CI] Dynamic CIv2 offload

nsumrakTT:
### Problem description
The CI is congested due to lack of machines.
Some should be offloaded to CIv2.

### What's changed
This PR makes it possible to create offload rules dynamically, without PR.
DevEx can test and tune config based on queue data.  

### Checklist

### DIFF
--- a/.github/scripts/python/generate_test_matrix.py
+++ b/.github/scripts/python/generate_test_matrix.py
@@ -9,6 +9,7 @@ import test_common
 from itertools import product
 from datetime import datetime
 import random
+import re
 
 default_duration = 150.0  # default duration in seconds if not found in _test_durations
 do_array_unroll_for = [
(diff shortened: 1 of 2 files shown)


.github/scripts/python/generate_test_matrix.py
```diff
@@ -40,6 +42,125 @@ def unroll_arrays(tests):
...
+    # Load CIv2 offload configuration
+    offload = []
+    try:
+        with open("_civ2_offload_config.json", "r") as f:
+            offload = json.load(f)
+    except (FileNotFoundError, json.JSONDecodeError) as e:
+        print(f"Warning: Could not load _civ2_offload_config.json: {e}")
+        return test_matrix
+
+    current_hour = datetime.now().hour
```
jameszianxuTT: (nit) Force UTC here?

This should take the runner's local time but it seems safer to explicitly normalize timezone, lest some random TZ config setting change introduce some funny™️  bugs

.github/workflows/call-prepare-run.yml
```diff
@@ -176,6 +176,9 @@ jobs:
           cat << 'EOF' > _test_durations
             ${{ vars.TEST_DURATIONS }}
           EOF
+          cat << 'EOF' > _civ2_offload_config.json
+            ${{ vars.CIV2_OFFLOAD_CONFIG }}
+          EOF
```
jameszianxuTT, line 181: nit: Might be nice to dump the contents of this variable to stdout, so devs that don't have access to repo vars can understand it a bit better?

We have seen issues before with differential failure between CIv2 and CIv1, and introducing this CI mechanism seems like it could make it harder to know what is running where, and why. Dumping this to stdout might be a cheap way to get some more visibility on that.

nsumrakTT: The ruleset can be seen by expanding the first line (script) of the "Generate test matrix" step.
Script prints out resulting matrix and rules that actually influenced decisions to stdout. 
Anyway, this is designed as internal DevEx tool for CI control rather than public stuff.

.github/scripts/python/generate_test_matrix.py
nsumrakTT: Our servers are set to UTC but it is good point because we should not rely on this.

nsumrakTT:
@jameszianxuTT Did I resolved your comments or you have more concerns?

PR #5630, opened 2025-10-31
[ttnn-jit] Using graph capture to produce IR

sgholamiTT:
### Problem description
Need to use graph capture to generate IR in `ttnn-jit`. The current method uses python's `AST` library which is neither scalable nor efficient. We also can't support `for` loops and `if` statements to be `jit`ed.

### What's changed
In order to do this, we wrap the call to the python function in `begin_graph_capture` and `end_graph_capture`. This will produce a graph similar to the following picture for this `cosh` function:
```python
def cosh(input_tensor):
    e_pos_x = ttnn.exp(input_tensor)
    e_neg_x = ttnn.exp(ttnn.neg(input_tensor))
    nr_term = ttnn.add(e_pos_x, e_neg_x)
    output = ttnn.multiply(nr_term, 0.5)
    return output
```

image

The graph produced by `graph_capture` is complicated and unnecessarily large. Therefore, we simplify this graph in multiple steps to achieve a graph similar to this:
Screenshot 2025-10-31 at 10 54 35 AM
Then we produce the IR from this simplified graph.

### Notes
- I ran all tests in `./test/ttnn-jit/test_eltwise.py` and confirmed that the IRs produced by this approach and the previous one are identical.
- I also ran all tests in `./test/ttnn-jit/test_layout.py` and confirmed the IRs are identical (the tests are skipped right now due to some issues).
- Using this new approach we can now decorate the functions with `for` loops and `if` statements with `jit` (which we couldn't before). For instance:
```python
def if_inside_for_multiple_branches(input_tensor, iterations=3):
    output = input_tensor
    for i in range(iterations):
        if i == 0:
            output = ttnn.exp(output)
        elif i == 1:
            output = ttnn.sin(output)
        else:
            output = ttnn.cos(output)
    return output

```
generates the following IR when decorated with `jit` (with `iterations=4`):
```mlir
#l1 = #ttnn.buffer_type<l1>
#ttnn_layout = #ttnn.ttnn_layout<(d0, d1) -> (d0, d1), <1x1>, memref<2x2x!ttcore.tile<32x32, bf16>, #l1>, <block_sharded>>
module {
  func.func @op_wrapper(%arg0: tensor<64x64xbf16, #ttnn_layout>) -> tensor<64x64xbf16, #ttnn_layout> {
    %0 = "ttnn.get_device"() <{mesh_offset = #ttnn<mesh_offset 0x0>, mesh_shape = #ttnn<mesh_shape 1x1>}> : () -> !ttnn.device
    %1 = "ttnn.exp"(%arg0) {ttnn.hoist_generic_via_d2m} : (tensor<64x64xbf16, #ttnn_layout>) -> tensor<64x64xbf16, #ttnn_layout>
    %2 = "ttnn.sin"(%1) {ttnn.hoist_generic_via_d2m} : (tensor<64x64xbf16, #ttnn_layout>) -> tensor<64x64xbf16, #ttnn_layout>
    %3 = "ttnn.cos"(%2) {ttnn.hoist_generic_via_d2m} : (tensor<64x64xbf16, #ttnn_layout>) -> tensor<64x64xbf16, #ttnn_layout>
    %4 = "ttnn.cos"(%3) {ttnn.hoist_generic_via_d2m} : (tensor<64x64xbf16, #ttnn_layout>) -> tensor<64x64xbf16, #ttnn_layout>
    return %4 : tensor<64x64xbf16, #ttnn_layout>
  }
}
```
- Added tests in `test_eltwise` to test `for` loops and `if` statements.
- We also decided to maintain both `graph` based and `ast` based approaches for IR generation. In order to do that, a flag was added to `jit` decorator: `use_ast_compiler` which is set to `False` by default. I also parametrized some tests to test both approaches for IR generation.

### New Files
- `graph_trace_compiler.py`: includes all code for graph clusterization and simplification as well as IR generation logic; the main body of this PR.
- `ir_generator.py`: wraps both approaches for IR generation and exposes a convenient function `generate_ir` which generates IR based on `use_ast_compiler` flag (defaulted to `False`).

### Disclaimer
_Some code that performs clusterization and simplification is written by @svuckovicTT found [here](https://github.com/tenstorrent/tt-metal/compare/main...svuckovic/graph_trace#diff-298626e7e892bf7e09a05129f81b7a36edcafad89e83ba8df25451aa7ac9c14cR1)._

### Checklist

### DIFF
--- a/.github/build-docker-images.sh
+++ b/.github/build-docker-images.sh
@@ -22,14 +22,10 @@ CIBW_IMAGE_NAME=ghcr.io/$REPO/tt-mlir-manylinux-2-34
 DOCKER_TAG=$(./.github/get-docker-tag.sh)
 echo "Docker tag: $DOCKER_TAG"
 
-# Are we on main branch
-ON_MAIN=$(git branch --show-current | grep -q main && echo "true" || echo "false")
-
 build_and_push() {
     local image_name=$1
     local dockerfile=$2
-    local on_main=$3
-    local from_image=$4
+    local from_image=$3
 
     IMAGE_EXISTS=false
     if docker manifest inspect $image_name:$DOCKER_TAG > /dev/null; then
@@ -62,18 +58,13 @@ build_and_push() {
         docker push $image_name:$DOCKER_TAG
       fi
     fi
-
-    if [ "$ON_MAIN" = "true" ]; then
-        echo "Pushing latest tag for $image_name"
-        docker buildx imagetools create $image_name:$DOCKER_TAG --tag $image_name:latest --tag $image_name:$DOCKER_TAG
-    fi
 }
 
-build_and_push $BASE_IMAGE_NAME .github/Dockerfile.base $ON_MAIN
-build_and_push $BASE_IRD_IMAGE_NAME .github/Dockerfile.ird $ON_MAIN base
-build_and_push $CI_IMAGE_NAME .github/Dockerfile.ci $ON_MAIN
-build_and_push $IRD_IMAGE_NAME .github/Dockerfile.ird $ON_MAIN ci
-build_and_push $CIBW_IMAGE_NAME .github/Dockerfile.cibuildwheel $ON_MAIN
+build_and_push $BASE_IMAGE_NAME .github/Dockerfile.base
+build_and_push $BASE_IRD_IMAGE_NAME .github/Dockerfile.ird base
+build_and_push $CI_IMAGE_NAME .github/Dockerfile.ci
+build_and_push $IRD_IMAGE_NAME .github/Dockerfile.ird ci
+build_and_push $CIBW_IMAGE_NAME .github/Dockerfile.cibuildwheel
 
 echo "All images built and pushed successfully"
 echo "CI_IMAGE_NAME:"
--- a/.github/workflows/call-build-docker.yml
+++ b/.github/workflows/call-build-docker.yml
@@ -48,7 +48,7 @@ jobs:
   build-image:
     needs: check-if-docker-exist
     if: needs.check-if-docker-exist.outputs.docker-image == ''
-    runs-on: tt-ubuntu-2204-large-stable
+    runs-on: docker-builder
     outputs:
       docker-image: ${{ steps.build.outputs.docker-image }}
       docker-tag: ${{ steps.build.outputs.docker-tag }}
--- a/lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
+++ b/lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
@@ -221,18 +221,19 @@ private:
 // This pattern detects and fuses a sequence of operations that implement
 // softmax:
 // 1. exp(x)
-// 2. sum(exp(x)) along a dimension with keep_dim=true
-// 3. broadcast the sum
-// 4. divide exp(x) by the broadcasted sum
+// 2. sum reduce(exp(x)) along a dimension with keep_dim=true
+// 3. (optional) broadcast the sum reduce
+// 4. divide exp(x) by the (broadcasted) sum reduce
 class SoftmaxFusionPattern : public mlir::OpRewritePattern<DivOp> {
   using mlir::OpRewritePattern<DivOp>::OpRewritePattern;
 
 public:
-  // Pattern: div(exp(x), broadcast(sum(exp(x), keep_dim=true))).
+  // Pattern: div(exp(x), sum reduce(exp(x), keep_dim=true)) or
+  //          div(exp(x), broadcast(sum reduce(exp(x), keep_dim=true))).
   mlir::LogicalResult
   matchAndRewrite(DivOp divOp, mlir::PatternRewriter &rewriter) const final {
 
-    // Get the numerator (exp) and denominator (broadcast).
+    // Get the numerator (exp) and denominator.
     mlir::Value numerator = divOp.getLhs();
     mlir::Value denominator = divOp.getRhs();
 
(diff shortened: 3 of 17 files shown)


arminaleTT (approved):
Please also update the layout tests to use the graph tracing engine and ensure they pass. Otherwise looks good!

tools/ttnn-jit/_src/graph_trace_compiler.py
```diff
@@ -0,0 +1,811 @@
...
+    def get_children_vertices(self, vertex_id: int) -> List[Vertex]:
+        """Get all vertices that this vertex connects to."""
+        vertex = self.get_vertex(vertex_id)
+        return [self.vertices[conn_id] for conn_id in vertex.connections]
+
+    def get_parent_vertices(self, vertex_id: int) -> List[Vertex]:
+        """Get all vertices that connect to this vertex."""
+        return [v for v in self.vertices.values() if vertex_id in v.connections]
+
+    def merge_vertex(self, vertex: Vertex) -> None:
```
vtangTT: this seems to be unused? can del or am i missing something?

tools/ttnn-jit/_src/graph_trace_compiler.py
```diff
@@ -0,0 +1,811 @@
...
+    def __str__(self) -> str:
+        return (
+            f"ClusterVertex(vertices={self.vertices}, connections={self.connections})"
+        )
+
+    def __repr__(self) -> str:
+        return self.__str__()
+
+
+class Graph:
```
vtangTT: can we call this `JitGraph` or something less common

tools/ttnn-jit/_src/ir_generator.py
```diff
@@ -0,0 +1,81 @@
...
+    print_and_verify_ir(ir, "TTIRCompiler (AST-based)", debug)
+
+    return ir
+
+
+def generate_ir_from_graph(f, debug, *args, **kwargs):
+    begin_graph_capture(RunMode.NO_DISPATCH)
+    f(*args)
+    captured_graph = end_graph_capture()
+    # visualize(captured_graph, file_name=f.__name__ + "_graph.svg")
```
vtangTT, line 39: maybe turn this on for debug? I'm not sure what it looks like, maybe it's too chunky to even print?

tools/ttnn-jit/_src/graph_trace_compiler.py
```diff
@@ -0,0 +1,811 @@
...
+        try:
+            return float(arg)
+        except (ValueError, TypeError):
+            return None
+
+    def _is_input_tensor(self, arg):
+        """Check if argument is an input tensor (not a reference to previous operation)."""
+        return arg.startswith("Tensor(")
+
+    def _get_tensor_reference(self, arg):
```
vtangTT: Is it possible to parse these when we load the graph, maybe somewhere in `add_vertex`? So we parse once during graph creation instead of on every access to a node.

test/ttnn-jit/test_eltwise.py
```diff
@@ -373,8 +438,260 @@ def sinh(input_tensor):
...
+        dtype,
+        op,
+        1,
+        buffer_type=ttnn.BufferType.L1,
+        use_ast_compiler=use_ast_compiler,
+    )
+
+
+# ------------------------------------------------------------
+# Control flow tests
```
vtangTT: can we move these to another test file

tools/ttnn-jit/api.py
```diff
@@ -12,6 +12,7 @@ def jit(
     max_grid: tuple[int, int] = (7, 7),
     compile_only: bool = False,
     debug: bool = False,
+    use_ast_compiler: bool = False,
```
vtangTT: nit: not my favourite flag name, wdyt about smthn like `graph_capture: bool =True` instead or something similar

sgholamiTT:
> Please also update the layout tests to use the graph tracing engine and ensure they pass. Otherwise looks good!

Thanks @arminaleTT. Those tests use [the default flags](https://github.com/tenstorrent/tt-mlir/blob/main/test/ttnn-jit/test_layouts.py#L57C27-L57C45) for `jit`(which is using the graph tracing engine [after this PR](https://github.com/tenstorrent/tt-mlir/pull/5630/files#diff-56ecaf7adc5fddf6a6c0bdeadfa345f6931b4de6be45ebd18b90fcae4fd3618d)). Thus, they're using the graph tracing engine by default.
In other words, "any" `jit`ed function will use this engine by default, unless specified otherwise.

tools/ttnn-jit/_src/ir_generator.py
sgholamiTT: Not sure, this will dump a pretty large `.svg` file (it takes >1 sec just to generate the file). Printing is also too chunky as you said. So I've commenting this so that I (or anyone else working on this) have an easy way for visualization.

vtangTT: ok makes sense to me

tools/ttnn-jit/_src/graph_trace_compiler.py
sgholamiTT: This is a great suggestion. Added some logic to extract and store this information in a `Vertex` of a `simplified_graph`. The lookup should now be `O(1)`.

test/ttnn-jit/test_eltwise.py
sgholamiTT: Moved to `test_control_flow.py` under `tests/ttnn-jit/`.
Also moved the `run_op_test` function to `tests/ttnn-jit/utils.py` as it was used by multiple test files.

sgholamiTT:
@xanderchin should I merge this ?

test/ttnn-jit/test_control_flow.py
```diff
@@ -0,0 +1,258 @@
...
+        else:
+            output = ttnn.cos(output)
+    return output
+
+
+# ------------------------------------------------------------
+# Control flow tests
+# ------------------------------------------------------------
+@pytest.mark.parametrize(
+    "h , w, max_grid", [(32, 32, (0, 0)), (64, 64, (0, 0)), (256, 256, (7, 7))]
```
xanderchin: Would suggest having a much larger tensor on the 8x8 grid because this test case will overlap with the single core test -- there's only one tile per core.

tools/ttnn-jit/_src/graph_trace_compiler.py
```diff
@@ -0,0 +1,854 @@
...
+
+        # For some reason, ttnn::ones calls somehow always point to next node in the graph + the correct node where its resulting tensor is consumed - fixing this by taking only the consumption node
+        if added_vertex.params["name"] == "ttnn::ones":
+            if len(added_vertex.connections) == 2:
+                added_vertex.connections = [added_vertex.connections[-1]]
+
+    return new_graph
+
+
+class GraphToIRCompiler:
```
xanderchin: Maybe `GraphToIRTranslator`?

tools/ttnn-jit/_src/graph_trace_compiler.py
```diff
@@ -0,0 +1,854 @@
...
+            if parsed_arg.arg_type == "constant":
+                # This is a numeric constant
+                constant_value = parsed_arg.parsed_value
+
+        # Generate the MLIR operation
+        return self._create_ttnn_op(
+            op_name, result_type, operands, device, constant_value
+        )
+
+    def _create_ttnn_op(
```
xanderchin, line 750: Lets split out the list of supported ops. Maybe into a separate module?

test/ttnn-jit/test_control_flow.py
sgholamiTT: Added the following sizes and grids for all the tests:
```python
TEST_SHAPE_GRIDS = [
    (32, 32, (0, 0)),
    (64, 64, (0, 0)),
    (256, 256, (7, 7)),
    (2048, 2048, (7, 7)),
    (64, 128, (0, 0)),
]
```

tools/ttnn-jit/_src/graph_trace_compiler.py
sgholamiTT: Added all the supported ops with their categories in `tools/ttnn-jit/_src/supported_ops.py`.

PR #5635, opened 2025-10-31
D2M: d2m-allocate cleanup+prep for buffer analysis integration

vroubtsovTT:
### Problem description
This PR effects some pending cleanup items for `d2m-allocate` and some incremental functionality changes:

### What's changed
1. The way streams were inserted had been (more) broken because I ignored the need to manage reblocking views and correct result typing. This has now been fixed, I hope, also in order to introduce some optionality in how stream buffer size and shape is selected.
2. As a first step towards future integration with "buffer analysis API" the pass can now select a stream buffer size different in shard shape/size from the tensor data. The binary mode for this logic (either keep the same size or choose the minimum size consistent with blocking factors and keeping the grid shape unchanged) is controlled by a new testing-only `test-buffer-size-policy` pass option.
3. Some original lit tests for the allocator were disabled in the interim and have become stale. They have now been replaced and a first builder test for the allocator has been created.
4. Because it is difficult to write tests that anticipate levels of memory pressure, a new testing-only pass option, `test-assume-l1-capacity` was added to allow mocking of L1 capacity.
5. `createTTIRToTTMetalMiddleendPipeline()` had a bug whereby it failed to copy over some pipeline options into `allocateOptions`.
6. The latest `GenericOp::isExplicitDatamovementForm()` changes in `main` have been integrated.
7. General logic flow streamlined somewhat so that the same entries of the analysis structure are populated by a single step. Pass steps renamed for better clarity. Misc improvements in tracing and error messages. New utils for printing iterables and "shape-like" objects.

### Checklist
- [x] Existing tests provide coverage for changes
- [x] `test/ttmlir/Dialect/D2M/allocate/allocate_{spill,reblock}.mlir` replace some disabled tests that became stale
- [x] `test/python/golden/test_metal_allocate.py` exercises reblocking logic triggered by the "min" policy
- [x] `test/ttmlir/Dialect/D2M/allocate/allocate_explicit_datamovement.mlir` is the updated and renamed version of the `allocate_empty_indexing_maps.mlir` original
- [x] `test/ttmlir/Dialect/D2M/allocate/allocate_multiple_uses.mlir`  added as verification for #5760 fix

### DIFF
--- a/.github/workflows/call-build-macos.yml
+++ b/.github/workflows/call-build-macos.yml
@@ -2,6 +2,11 @@ name: Build on macos-latest
 
 on:
   workflow_dispatch:  # Manual trigger
+    inputs:
+      run-linter:
+        description: 'Run linter'
+        type: boolean
+        default: false
   workflow_call:
 
 env:
@@ -72,7 +77,7 @@ jobs:
         cmake --build ${{ steps.strings.outputs.build-output-dir }} --config ${{ matrix.build.build_type }}
 
     - name: Lint
-      if: matrix.build.enable_runtime == 'OFF'
+      if: inputs.run-linter && matrix.build.enable_runtime == 'OFF'
       shell: bash
       run: |
         source env/activate
--- a/.github/workflows/call-prepare-run.yml
+++ b/.github/workflows/call-prepare-run.yml
@@ -181,7 +181,7 @@ jobs:
           EOF
           TARGET_DURATION=${{ vars.TARGET_DURATION }}
           if [ -z "$TARGET_DURATION" ]; then
-            TARGET_DURATION=30
+            TARGET_DURATION=70
           fi
           python .github/scripts/python/generate_test_matrix.py .github/settings/tests.json "$TARGET_DURATION" "$comp"
           echo "test-matrix=$(cat _test_matrix.json | jq -c)" >> $GITHUB_OUTPUT
--- a/docs/supplemental/On Commuting Elementwise Multiply with Matmul in TTIR - Mihailo Djordjevic - Gaus edukativni centar.tex
+++ b/docs/supplemental/On Commuting Elementwise Multiply with Matmul in TTIR - Mihailo Djordjevic - Gaus edukativni centar.tex
@@ -1,0 +1,611 @@
+\documentclass[11pt,a4paper]{article}
+\usepackage[utf8]{inputenc}
+\usepackage[T1]{fontenc}
+\usepackage[serbian,english]{babel}
+\usepackage{newtxtext}
+\usepackage{newtxmath}
+\usepackage{microtype}
+\usepackage{bm}
+\usepackage{mathtools}
+\let\openbox\relax
+\usepackage{amsthm}
+\usepackage[a4paper,margin=27mm,headsep=8mm,footskip=12mm]{geometry}
+\usepackage{titlesec}
+\usepackage{setspace}
+\setstretch{1.08}
+\titlespacing*{\section}{0pt}{1.3ex plus .5ex}{0.9ex}
+\titlespacing*{\subsection}{0pt}{1.1ex plus .5ex}{0.7ex}
+\titlespacing*{\subsubsection}{0pt}{1ex plus .3ex}{0.6ex}
+\setlength{\parskip}{0.35em}
+\setlength{\parindent}{12pt}
+\usepackage{enumitem}
+\setlist{itemsep=0.2em,topsep=0.3em,leftmargin=2.2em}
+\usepackage{array}
+\usepackage{booktabs}
+\usepackage{multirow}
+\usepackage{graphicx}
+\usepackage{xcolor}
+\usepackage[unicode]{hyperref}
+\hypersetup{
+  colorlinks=true,
+  linkcolor=black,
+  citecolor=black,
+  urlcolor=blue!55!black,
+  pdfauthor={Mihailo Đorđević},
+  pdftitle={On Commuting Elementwise Multiply with Matmul in TTIR}
+}
+\usepackage{cleveref}
+\newtheorem{theorem}{Theorem}[section]
+\newtheorem{lemma}[theorem]{Lemma}
+\newtheorem{corollary}[theorem]{Corollary}
+\newtheorem{proposition}[theorem]{Proposition}
+\theoremstyle{definition}
+\newtheorem{definition}[theorem]{Definition}
+\newtheorem{remark}[theorem]{Remark}
+\newtheorem{example}[theorem]{Example}
+\newcommand{\R}{\mathbb{R}}
+\newcommand{\Had}{\circ}
+\newcommand{\Mat}{\,}
+\newcommand{\diag}{\operatorname{diag}}
+\newcommand{\Diag}{\operatorname{Diag}}
+\newcommand{\vecop}{\operatorname{vec}}
+\newcommand{\rank}{\operatorname{rank}}
+\newcommand{\supp}{\operatorname{supp}}
+\title{\vspace{-6mm}\Large On Commuting Elementwise Multiply with Matmul in TTIR:\\
+Necessary and Sufficient Conditions, Rewrite Patterns, and Extensions}
+\author{\foreignlanguage{serbian}{Mihailo Đorđević}}
+\date{}
+\begin{document}
+\maketitle
+\vspace{-6mm}
+\begin{abstract}
+This note studies when an elementwise product can be moved across a matrix multiplication without changing the numerical result. Two base transformations are considered: commuting the elementwise mask to the first operand of \texttt{matmul} and commuting it to the second operand. I give simple criteria that decide when each rewrite is valid. When the base rewrites fail, I discuss extensions that pre-process the left input through a fixed linear operator so that the expression can still be reordered in a useful way. Proofs and counterexamples are deferred to the appendices.
+\end{abstract}
+\tableofcontents
+\clearpage
+\section{Introduction and scope}
+\label{sec:intro}
+The TTIR graph often contains the pattern
+\[
+(A \Had B) \, C,
+\]
+where $A,B,C \in \R^{n\times n}$, $\Had$ denotes the Hadamard (elementwise) product, and juxtaposition denotes standard matrix multiplication. This work provides precise conditions for two shape preserving rewrites:
+\begin{itemize}
+  \item \textbf{Transformation A} (move the mask to the left input): find $B',C'$ depending only on $B,C$ such that
+  \[
+    (A \Had B) \, C = (A\,B') \Had C' \quad \text{for all } A.
+  \]
+  \item \textbf{Transformation B} (move the mask to the output of \texttt{matmul}): find $B',C'$ depending only on $B,C$ such that
+  \[
+    (A \Had B) \, C = A \Had (B' C') \quad \text{for all } A.
+  \]
+\end{itemize}
+When neither transformation is globally valid, I also describe extended rewrites that introduce a fixed linear pre-processing of $A$ before the \texttt{matmul}.
+\subsection{Notation}
+All matrices are real $n\times n$ unless stated otherwise. For a vector $u\in\R^n$, $\diag(u)$ is the diagonal matrix with diagonal $u$. For a matrix $X$, $\vecop(X)$ stacks its columns into a vector. The support of a vector $x$ is $\supp(x)=\{i: x_i\ne 0\}$. The $k$th column of $X$ is $X_{\cdot k}$. I use $I$ for the identity, $\bm{1}$ for the all ones vector, and $J=\bm{1}\bm{1}^\top$ for the all ones matrix.
+\subsection{Elementary facts used throughout}
+I collect the basic identities that will be used in the analysis. Proofs are standard and omitted; they are recalled in Appendix~A for completeness.
+\paragraph{Hadamard product.}
+For $X,Y\in\R^{n\times n}$, the Hadamard product $X\Had Y$ is defined entrywise, $(X\Had Y)_{ij}=X_{ij}Y_{ij}$. It is bilinear, commutative and associative. Two interactions will be repeatedly used:
+\begin{align}
+X \Had (u v^\top) &= \diag(u)\,X\,\diag(v), \label{eq:had-rank1}\\
+\vecop(X\Had Y) &= \Diag(\vecop(Y))\,\vecop(X). \label{eq:vec-had}
+\end{align}
+Identity \eqref{eq:had-rank1} says that a rank one mask $uv^\top$ induces left and right diagonal scalings. Identity \eqref{eq:vec-had} converts Hadamard multiplication into a diagonal linear map in the vectorized space.
+\paragraph{Matrix multiplication and vectorization.}
+For conforming matrices, $\vecop(AXB)=(B^\top\!\otimes A)\vecop(X)$. This turns matching of matrix identities into matching of linear maps in the vectorized form. I will use this to equate coefficients with respect to the basis matrices $E^{(ik)}$ that have a single one in position $(i,k)$.
+\paragraph{Linearity in $A$.}
+In both transformations, $B$ and $C$ are fixed while $A$ is arbitrary. Each side is linear in $A$. Equality for all $A$ holds if and only if the two linear maps agree on a basis, equivalently if the coefficient of each $a_{ik}$ matches in every output position. This is the key idea behind the criteria in Section~\ref{sec:dev-criteria} and the proofs in Appendix~A.
+\subsection{Problem statement}
+Given $B,C\in\R^{n\times n}$ and arbitrary $A\in\R^{n\times n}$, decide:
+\begin{enumerate}[label=\textbf{Q\arabic*.},wide]
+  \item When do there exist $B',C'$ depending only on $B,C$ such that $(A\Had B)C=(AB')\Had C'$ for all $A$?
+  \item When do there exist $B',C'$ depending only on $B,C$ such that $(A\Had B)C=A\Had(B'C')$ for all $A$?
+  \item If neither is possible in general, what structured cases allow it?
+  \item If a base rewrite is not valid, can we insert a fixed linear pre-processing of $A$ to obtain a useful reordered form without changing the final result?
+\end{enumerate}
+\vspace{3cm}
+\section{Main results}
+\label{sec:dev-criteria}
+This section states the conditions in plain English with minimal notation and many examples. Mathematical proofs are in Appendix~A.
+\subsection{Transformation A: when can we move the mask to the left input}
+\label{subsec:dev-A}
+\paragraph{Target form.}
+We want fixed $B',C'$ that depend only on $B,C$ such that
+\[
+(A \Had B)\,C \;=\; (A\,B') \Had C' \quad \text{for every } A.
+\]
+No shape changes and no extra operators are allowed.
+\subsubsection*{Practical criterion}
+Think column by column through $C$.
+\begin{enumerate}[label=\arabic*)]
+  \item Pick a column $j$ of $C$.
+  \item Find the row indices where this column is nonzero:
+        \[
+        \mathrm{support}(C_{\cdot j})=\{\,k : C_{k j}\ne 0\,\}.
+        \]
+  \item Look at those columns of $B$: $B_{\cdot k}$ for all $k$ in that set.
+  \item \textbf{Requirement.} All these $B_{\cdot k}$ must point in the same direction, that is, they must be scalar multiples of one another.
+\end{enumerate}
+If the requirement holds for every column $j$ of $C$, then the rewrite is valid. If it fails for any column, the rewrite is not valid.
+\paragraph{Intuition.}
+Each column of $C$ decides which columns of $B$ are allowed to interact in the same output column. When two or more columns of $B$ are mixed together by the same column of $C$, those $B$-columns must be proportional. If a column of $C$ is 1-sparse, there is nothing to check for that column.
+\subsubsection*{Quick checklists}
+\begin{itemize}
+  \item Dense $C$ (no zeros): then every column of $C$ mixes all columns of $B$, so all columns of $B$ must be mutually proportional. Equivalently, $B$ must be rank 1.
+  \item Each column of $C$ has at most one nonzero: the rewrite always passes for any $B$.
+  \item Block-sparse $C$ where columns only use indices from disjoint blocks $\mathcal{K}_1,\dots,\mathcal{K}_r$: inside each block $\mathcal{K}_t$, all columns of $B$ must be proportional. Different blocks may use different directions.
+\end{itemize}
+\subsubsection*{How to build $B'$ and $C'$ when it passes}
+For each column $j$ of $C$:
+\begin{itemize}
+  \item pick a unit direction $u^{(j)}$ equal to the common direction of the involved $B_{\cdot k}$,
+  \item write $B_{\cdot k}=\alpha_k u^{(j)}$ for all $k$ with $C_{k j}\ne 0$,
+  \item set $C'_{\cdot j}=u^{(j)}$ and $B'_{k j}=\alpha_k C_{k j}$, with zeros elsewhere.
+\end{itemize}
+Then $(A B') \Had C'$ reproduces $(A \Had B) C$ for all $A$.
+\subsubsection*{Examples that pass}
+\paragraph{Example A1. Overlapping supports, consistent proportionality.}
+\[
+B=\begin{bmatrix}
+1 & 2 & 1\\
+2 & 4 & 2\\
+3 & 6 & 3
+\end{bmatrix},
+\qquad
+C=\begin{bmatrix}
+1 & 0 & 0\\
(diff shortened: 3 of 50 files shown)


test/python/golden/test_metal_allocate.py
```diff
@@ -0,0 +1,48 @@
+# SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+
+import pytest
+
+from builder.base.builder_utils import compile_and_execute_ttir
+
+# borrow currently constrained way to build matmul inputs:
+from test_metal_matmul import create_matmul_constrained_inputs as create_matmul_inputs
```
vroubtsovTT, line 10: This kind of import (into a generic local name) was intentional, as I needed to borrow the exact way of creating matmul golden inputs that's working right now but did not want to move that method to `test_utils.py`. If smth else is preferable, please comment.

include/ttmlir/Dialect/D2M/Analysis/Allocation/Utils.h
```diff
@@ -107,7 +129,117 @@ auto make_lexicographical_field_comparator(Fields... fields) {
...
 
+template <typename T>
+SmallVector<T> concat(ArrayRef<T> a, SmallVector<T> b) {
+  return concat(a, ArrayRef<T>(b));
+}
+
+template <typename T>
+SmallVector<T> concat(SmallVector<T> a, ArrayRef<T> b) {
+  return concat(ArrayRef<T>(a), b);
+}
```
nsmithtt, line 167: There is already `llvm::concat`, perhaps we should defer to their library if a utility already exists?

test/python/golden/test_metal_allocate.py
nsmithtt: Maybe we should have a d2m_utils.py or common one? Not a big deal though in my opinion the way it is.

lib/Dialect/D2M/Transforms/Allocate.cpp
```diff
@@ -362,195 +440,351 @@ class D2MAllocate final : public impl::D2MAllocateBase<D2MAllocate> {
...
+        // Decide which operands might/must have streams. Note that
+        // the actual stream creation decision is only final after
+        // we have a feasible planner solution.
+
+        const AffineMap &indexingMap = indexingMaps[operandIndex];
+
+        if (operandCtx.isOutput) {
+          // L1 outputs are currently allocated in L1 so won't use streams
+          // unless allowed to do so in `allowL1OutputSpilling` mode. DRAM
+          // outputs always need to be streamed.
```
nsmithtt, line 634: Does this condition get skipped if `allowL1OutputSpilling` is true? I didn't see how `operandCtx.requiresStream` gets set for this case.

nsmithtt: Maybe it just throws an error later, maybe that's OK for now as long as out existing DRAM tests are still passing (those are datamovement only generic's though I think so perhaps they're already skipped by this logic).

nsmithtt:
Will continue reviewing later!

lib/Dialect/D2M/Transforms/Allocate.cpp
vroubtsovTT: ~I actually neglected the code path for output streaming except to add comments where more logic would be needed. (Currently any output streams will get asserted promptly by one of the lower passes.)~

~Since I can't make an e2e test for it and since we don't know the official IR for streaming outputs (will it be done via streams? or special views?) I would like to resolve this with something like `TT_assertv(false, "TODO output streaming is not yet fully implemented")`. I have already made sure that no allocate lit tests use `allowL1OutputSpilling` as of this PR.~

There is currently a pass error emitted via 
```
     if (operandCtx.isOutput) {
            // We get here if output streaming was enabled and the output
            // ended up spilled to DRAM. But there is no full support for
            // output streams/views as of yet, so for now just bail.
            TT_ALLOC_ERROR("inserting output streams not implemented yet");
            return failure();
     } /* else */
```
in `insertOperandStreams()` and the pass fails. This happens only if output spilling option is enabled and only if some output gets actually spilled. But reality is that this code path isn't officially supported yet, pending deciding how we actually do it (stream_layout, special view, etc).  The `/* else */` shows how the code will be altered once the logic is agreed on: the if-branch will emit the appropriate output stream/view, the else-branch will continue with `insertStream()`.

include/ttmlir/Dialect/D2M/Analysis/Allocation/Utils.h
vroubtsovTT: I started out with some ambitions to improve on having to do `llvm::to_vector(llvm::concat<...>....)` but lost patience with type inference... I'll go back to llvm version.

vroubtsovTT: I've replaced prev impl with these typing-saving overloads:
```
template <typename T>
SmallVector<T> concatToVector(const SmallVector<SmallVector<T>> &vs) {
  SmallVector<T> r;
  for (const auto &v : vs) {
    r.append(v.begin(), v.end());
  }
  return r;
}

template <typename T, typename... RangeTs>
auto concatToVector(RangeTs &&...ranges)
    -> std::enable_if_t<(sizeof...(RangeTs) > 1), SmallVector<T>> {
  return llvm::to_vector(
      llvm::concat<const T>(std::forward<RangeTs>(ranges)...));
}
```

lib/Dialect/D2M/Transforms/Allocate.cpp
```diff
@@ -811,112 +1050,313 @@ class D2MAllocate final : public impl::D2MAllocateBase<D2MAllocate> {
...
+  /// Calculate full "shard-only" blocking factors (defined as full blocking
+  /// factors divided by blocking factors)
+  static SmallVector<int64_t> getShardBlockFactors(d2m::GenericOp genericOp) {
+    SmallVector<int64_t> r = genericOp.getFullBlockFactors();
+    SmallVector<int64_t> blockFactors = genericOp.getBlockFactorsValue();
+    TT_debug(r.size() == blockFactors.size());
+    for (std::size_t d = 0; d < r.size(); ++d) {
+      TT_debugv(blockFactors[d] > 0, "unexpected block factor {} for dim {}",
+                blockFactors[d], d);
+      r[d] /= blockFactors[d];
```
nsmithtt, line 1329: Why are we dividing by the current block factors?  It's legal to pick any of the "full" block factors regardless of how they're currently set.

lib/Dialect/D2M/Transforms/Allocate.cpp
```diff
@@ -811,112 +1050,313 @@ class D2MAllocate final : public impl::D2MAllocateBase<D2MAllocate> {
...
     }
 
     return success();
   }
 
-  static void assignAddressAndAlignment(RewriterBase &rewriter,
-                                        memref::AllocOp op,
-                                        Planner::AllocSizeT offset,
-                                        const MemorySpaceInfo &info) {
+  LogicalResult insertStream(RewriterBase &rewriter, OpOperand &operand,
```
nsmithtt, line 1133: Out of scope for this PR, but following our streams 2.0 discussion yesterday.  It might be nice to have this be a method or standalone utility function, something like:
```c++
struct ParallelizationResult {
  SmallVector<memref::AllocOp> cbAllocations;
  SmallVector<memref::DellocOp> deallocations;
  ...
};

FailureOr<ParallelizationResult> GenericOp::setParallelization(
  RewriterBase &rewriter,
  std::optional<SmallVector> newBlockFactors,
  std::optional<SmallVector> newGridShape);
```

Where this function could be used by the GridSelection pass and Allocator pass.  It can infer the required IR transformation based on just these inputs.

The other benefit of this abstraction is that the region ops fixup + block args in place setting could be part of this routine.  This would enable the IR update to be encapsulated entirely in one place.

lib/Dialect/D2M/Transforms/Allocate.cpp
nsmithtt: Oh oops, forgot about this one, I think we should address this before landing.  Or if we want to define the policy as being "allocator will only divide further the existing block factors, i.e. it will never upsize a buffer", this is OK with me too.  Although in the future I could see an argument for enabling that.

vroubtsovTT: This impl is restricted to keeping the grid shape constant, because there is no legal way to stream outputs as commented elsewhere. So, the factors that "come from the grid" are not really considered available. I really just wanted to have one reblocking option enabled, to learn how to reblock and clone generics so it's accepted by the verifiers and all downstream passes.

lib/Dialect/D2M/Transforms/Allocate.cpp
vroubtsovTT: I chatted with Vincent about his generic recreator/cloner in GridSelection. Everybody has some ugliness in their version. I have spent considerable time trying alternatives, e.g. signature change support available with TypeConverters or via Operation::clone() with IRMapping. 

One difficulty is that we don't have enough GenericOp::build()s to make it easy to clone the op with all its possible threads. I gave up and stayed with in-place modification just to have something working well enough to pass some e2e. TypeConverters in MLIR's `main` now have the ability to define "context-aware" type replacement rules, which I think will make it much easier to deal with block arg replacement.

lib/Dialect/D2M/Transforms/Allocate.cpp
vroubtsovTT: The intent was to let `operandCtx.requiresStream` remain set to its construction value of `false`. I use a style with my context structs that (almost) all fields have an initializer so that an instance never exists with some indeterminate content, including immediately after the constructor call. So, this logic seems correct to me and the output streamingness is dealt with (and gated with a WIP error abort) in a later step.

lib/Dialect/D2M/Transforms/Allocate.cpp
nsmithtt: Hmm, but this code has nothing to do with the grid, this is taking the full block factors and dividing them by the current block factors.

lib/Dialect/D2M/Transforms/Allocate.cpp
nsmithtt: This method based approach is proposed to be in-place, note that it's a method on the generic op itself.

> Everybody has some ugliness in their version

Exactly my argument for creating a common abstraction.  Let's encapsulate all of the ugliness in one place, otherwise this is a recipe for unmaintainable code.

lib/Dialect/D2M/Transforms/Allocate.cpp
vroubtsovTT: It is just a calc shortcut for me so I didn't have to repeat the same `concatInversePermutationMap()` tricks just to recover the indexing space axes ( m, n, k, etc) but instead rely on this already done for me in the various block factor getters.

If I call `{G0, G1, G2,...}` and `{S0, S1, S2,...}` the grid and shard limits of the indexing space axes (what I call "extents" in code) then the current definitions of factors are such that
```
block_factors = {G0, G1, G2,...} / grid_shape
```
and 
```
full_block_factors = block_factors * {S0, S1, S2,...}
```
So if I want the "shard-only factors", I can recover them as `full_block_factors / block_factors`.

nsmithtt: Maybe let me put it another way, the FullBlockFactors already do _not_ include the generic op's grid factors.

nsmithtt: Note the implementation of `getFullBlockFactors` uses `flattenedOperandShardShapes`.  They are already "shard only".

vroubtsovTT: Correct. Those operandShardShapes are the factors I want to get at. They are multiplied by block_factors to produce full_block_factors. I am merely undoing this multiplication.
```
  auto currentBlockFactors = getBlockFactorsValue();
  auto factorizations = flatInverseMap.compose(flattenedOperandShardShapes);
  TT_assert(currentBlockFactors.size() == factorizations.size());

  // Multiply back in the current block factors to normalize the result.
  for (std::size_t i = 0; i < currentBlockFactors.size(); ++i) {
    factorizations[i] *= currentBlockFactors[i];
  }
```
Again, I am not trying for "maximum possible reblocking" here. I need to keep the grid extents the same because there is no way (at the moment) to produce a re-parallelized generic if one allows grid to change (because it means the output operand need a view/stream/whatever that reflects reblocking and we don't support that now). So I just want the subset of full factors that are "shard-only".

This PR is a POC for me to learn to reblock correctly. I just want one valid reblocking that's different from a no-op that I could observe go through the e2e cycle successfully. Eventually this logic has to change anyway because a full resizing policy will be in `GenericOpBufferAnalysis` and will need to produce a list of reblocking options, not just one. As we implement that we will have to deal with the issue of output streaming as a prerequisite.

nsmithtt: > Again, I am not trying for "maximum possible reblocking" here. I need to keep the grid extents the same because there is no way (at the moment) to produce a re-parallelized generic if one allows grid to change (because it means the output operand need a view/stream/whatever that reflects reblocking and we don't support that now). So I just want the subset of full factors that are "shard-only".

My point is that this is not related to the grid factors, this is already handled by the full block factors routine.  Working through an example:
```mlir
%lhs : memref<2x3x4x5>
%rhs : memref<3x6x5x7>
%out : memref<2x6x4x7>

generic ins(%lhs, %rhs) outs(%out) {
  blocking_factors = [1, 1, 3],
  grid = [2, 6],
  index_maps = matmul
}
```
The block factors never include the generic's grid shape, in this example the full block factors will return:
-  `[4, 7, 15]`

We can derive the steps that the full block factors routine is doing:
1. The code in full block factors is taking the current operand shard shapes and applying them to that map to get: `[4, 7, 5]`
2. And then multiplying it out by the existing block factors: `[1, 1, 3]`
3. This gives you the full block factors: `[4, 7, 15]`

What you're then doing is dividing by the current block factors:
- `[4, 7, 15]` / `[1, 1, 3]` -> `[4, 7, 5]`

Essentially what this division is doing, it disallowing the allocator to choose a larger allocation for `k` dim, i.e you're clamping it to the existing `k` of `3`.

nsmithtt: There is however code elsewhere in this PR that _does_ consider the output factors effectively ignoring the `4, 7` factors in this example.  But that is a separate routine above `getParticipatingDimMask `.

vroubtsovTT: Your example follows my formulas for block_factors and full_block_factors earlier. There is no disagreement there.

> Essentially what this division is doing, it disallowing the allocator to choose a larger allocation for k dim, i.e you're clamping it to the existing k of 3.

_That was my explicit goal._  My intent was to only allow the range `[1, current k]`. Dividing by the grid guarantees to give me such a valid range.

The "max" mode name is a misnomer, it should have been called "default", as I had it at one point. The code only enables min mode when the option string is "min" explicitly, it will keep the existing buffer sizing for any other option value.  The "min"/"max" nomenclature in my mind refers to the min/max limits of an allowed range for a dim. How that range is determined is not fully fleshed out by this PR, it will be done later along with more `GenericOpBufferAnalysis` work. I really just needed one mode where I could lower mem pressure for prototyping purposes and e2e testing.

lib/Dialect/D2M/Transforms/Allocate.cpp
vroubtsovTT: https://github.com/tenstorrent/tt-mlir/issues/5759

lib/Dialect/D2M/Transforms/Allocate.cpp
vroubtsovTT: > There is however code elsewhere in this PR that does consider the output factors effectively ignoring the 4, 7 factors in this example. But that is a separate routine above getParticipatingDimMask()
 
For a while this code existed with only the shard-only calculation needed and that felt good because it was insulated from needing to use `utils::concatInversePermutationMap()` directly, as that implies some restrictions on the allowed class of affine maps (permutation, projected permutation, etc). It seems useful to derive shard extent based on other attributes/getters already available on the op and not reach for any "under the hood" details.

The late need for `getGridAndShardExtents()` (which _does_ reach under the hood) arose when I realized that the block factory machinery had some ambiguity when dealing with operands of _mixed_ element types. E.g. tiled input, f32 outputs or vice versa. Tiling effectively represents yet another way of blocking work that is not reflected in [full] block factor values. I needed this case to work, however, because our e2e builder tests generate such IR all the time:
```
  func.func @VLAD(%arg0: memref<384x384xf32>, %arg1: memref<384x384xf32>) -> memref<1x1x384x384xf32, #ttcore.shard<1536x4>, #ttcore.memory_space<l1>> {
    %alloc = memref.alloc() : memref<1x1x12x12x!ttcore.tile<32x32, f32>, #ttcore.shard<49152x4096>, #ttcore.memory_space<l1>>
    %alloc_0 = memref.alloc() : memref<1x1x384x384xf32, #ttcore.shard<1536x4>, #ttcore.memory_space<l1>>
    d2m.to_layout %arg0, %alloc_0 : memref<384x384xf32> into memref<1x1x384x384xf32, #ttcore.shard<1536x4>, #ttcore.memory_space<l1>> hostInfo = <logical_shape = 384x384, dim_alignments = 32x32, collapsed_intervals = dense<[[0, 1], [1, 2]]> : tensor<2x2xi64>, undef, l1>
    d2m.generic {block_factors = [1, 1], grid = #ttcore.grid<1x1>, indexing_maps = [affine_map<(d0, d1) -> (d0, d1)>, affine_map<(d0, d1) -> (d0, d1)>], iterator_types = [#ttcore.iterator_type<parallel>, #ttcore.iterator_type<parallel>], threads = [#d2m.thread<compute>]}
        ins(%alloc_0 : memref<1x1x384x384xf32, #ttcore.shard<1536x4>, #ttcore.memory_space<l1>>)
        outs(%alloc : memref<1x1x12x12x!ttcore.tile<32x32, f32>, #ttcore.shard<49152x4096>, #ttcore.memory_space<l1>>)  {
    ^compute0(%cb0: !d2m.cb<memref<384x384xf32, #ttcore.memory_space<l1>>>, %cb1: !d2m.cb<memref<12x12x!ttcore.tile<32x32, f32>, #ttcore.memory_space<l1>>>):
      %0 = d2m.wait %cb0 : <memref<384x384xf32, #ttcore.memory_space<l1>>> -> memref<384x384xf32, #ttcore.memory_space<l1>>
      %1 = d2m.reserve %cb1 : <memref<12x12x!ttcore.tile<32x32, f32>, #ttcore.memory_space<l1>>> -> memref<12x12x!ttcore.tile<32x32, f32>, #ttcore.memory_space<l1>>
      "d2m.tile_tilize_block"(%0, %1) : (memref<384x384xf32, #ttcore.memory_space<l1>>, memref<12x12x!ttcore.tile<32x32, f32>, #ttcore.memory_space<l1>>) -> ()
    }

    %alloc_5 = memref.alloc() : memref<1x1x384x384xf32, #ttcore.shard<1536x4>, #ttcore.memory_space<l1>>
    d2m.generic {block_factors = [1, 1], grid = #ttcore.grid<1x1>, indexing_maps = [affine_map<(d0, d1) -> (d0, d1)>, affine_map<(d0, d1) -> (d0, d1)>], iterator_types = [#ttcore.iterator_type<parallel>, #ttcore.iterator_type<parallel>], threads = [#d2m.thread<compute>]}
        ins(%alloc : memref<1x1x12x12x!ttcore.tile<32x32, f32>, #ttcore.shard<49152x4096>, #ttcore.memory_space<l1>>)
        outs(%alloc_5 : memref<1x1x384x384xf32, #ttcore.shard<1536x4>, #ttcore.memory_space<l1>>)  {
    ^compute0(%cb0: !d2m.cb<memref<12x12x!ttcore.tile<32x32, f32>, #ttcore.memory_space<l1>>>, %cb1: !d2m.cb<memref<384x384xf32, #ttcore.memory_space<l1>>>):
      %0 = d2m.wait %cb0 : <memref<12x12x!ttcore.tile<32x32, f32>, #ttcore.memory_space<l1>>> -> memref<12x12x!ttcore.tile<32x32, f32>, #ttcore.memory_space<l1>>
      %1 = d2m.reserve %cb1 : <memref<384x384xf32, #ttcore.memory_space<l1>>> -> memref<384x384xf32, #ttcore.memory_space<l1>>
      "d2m.tile_untilize_block"(%0, %1) : (memref<12x12x!ttcore.tile<32x32, f32>, #ttcore.memory_space<l1>>, memref<384x384xf32, #ttcore.memory_space<l1>>) -> ()
    }
    return %alloc_5 : memref<1x1x384x384xf32, #ttcore.shard<1536x4>, #ttcore.memory_space<l1>>
  }
```

 So as a workaround I added `getGridAndShardExtents()` so I can apply "tile adjustment factors":
```
        SmallVector<int64_t> shardShapeRescaled =
            canonicalMap.compose(shardExtents);
        // TODO(vroubtsov) not sure if there's a better option right now for
        // adjusting to input/output tile shape changes:
        if (operandCtx.isOutput) {
          for (std::size_t t = 0; t < 2; ++t) {
            const std::size_t d = shardShapeRescaled.size() - 2 + t;
            shardShapeRescaled[d] =
                (shardShapeRescaled[d] * inputTileFactors[t]) /
                outputTileFactors[t];
          }
        }
```
Any perceived inconsistency is entirely due to this tile issue, which probably deserves some discussion/clarification. (was meant to happen in yesterday's meeting but the agenda was overfull).

nsmithtt: > That was my explicit goal. My intent was to only allow the range [1, current k]. Dividing by the grid guarantees to given me such a valid range.

But why? It's totally fine to pick a block factor of `1` for example, it's not related to the generic op's grid shape.  I.e. block factors of `[1, 1, 1]` are totally fine an legal in the above example, as long as it fits in memory.

vroubtsovTT: I envisaged the logic to be a rescaling "new extents[dim] = old extent[dim] * rescaling[dim]" and wanted some correct way of computing "rescaling". This value starts with all-ones and then gets modified based on some "policy":
```
      {
        SmallVector<int64_t> rescaling(rank, 1);
        if (useMinPolicy) {
          const llvm::BitVector participationMask =
              getParticipatingDimMask(genericOp);
          const SmallVector<int64_t> shardFactors =
              getShardBlockFactors(genericOp);
          for (auto d = participationMask.find_first_unset(); d >= 0;
               d = participationMask.find_next_unset(d)) {
            rescaling[d] = shardFactors[d];
          }
        }

        for (std::size_t d = 0; d < rank; ++d) {
          gridExtents[d] *= rescaling[d];
          shardExtents[d] /= rescaling[d];

          blockFactors[d] *= rescaling[d];
        }
        TT_ALLOC_DEBUG("rescaling {}, new block factors {}", asSeq(rescaling),
                       asSeq(blockFactors));
      }
```
There is a point above where a choice within `[1, max]` is being made, for every `dim`. Doesn't matter that 1 is always a valid choice at the edge of the valid range, the important point is to design for such a choice. This bit of logic needs to know the other end of that range, the max, but it is not explicitly thinking in terms of grid divisions at all, it thinks in terms of rescaling along m,n,k axes and what valid ranges along all of them are.

Just a way to reason about valid choices here, with a mind in the future to be emitting a problem definition for some "planner" that is stated in terms of discrete decision variables. E.g. this could be modified to generate all possible reblockings as a cross product of some allowed factors. One day this could be something that generates a k-ary decision variable for a new MILP Planner instead of a single choice, etc.

nsmithtt: > I envisaged the logic to be a rescaling "new extents[dim] = old extent[dim] * rescaling[dim]" 

Ok this is what I was getting at.  As long as this is deliberately our policy I'm ok with it, I just want to call it out somewhere as a comment because it is ignoring parts of the legal solution space and potentially ignoring the most optimal choice.  In a follow on change we should enable this part of the solution space, by removing the division you're doing here.  Note that there's no reason we couldn't just enable this today, there is no constraint prohibiting this.

vroubtsovTT: I've updated the WIP comment to make the interim state of affairs there more intentional:

```
      // WIP: There is a set of valid stream buffer sizes and selecting them
      // optimally is one of the goals of this pass. Eventually, this decision
      // will involve integration with the `GenericOpBufferAnalysis` API. In the
      // interim, however, we implement a simple binary choice policy, "keep
      // incoming shapes or pick a small-ish buffer" chosen by the
      // `testBufferSizePolicy` test pass option. Choices made by the policy
      // will be captured in new block factors `genericCtx.reblockedFactors` and
      // grid/shard dim extents in `grid/shardExtents`.
```

nsmithtt: Sounds great! Thanks!

PR #5589, opened 2025-10-29
[ttnn.jit] enable_cache flag + re-enable layout tests w/o cache

vtangTT:
### Ticket
https://github.com/tenstorrent/tt-mlir/issues/5550#issuecomment-3462833326

### Problem description
ND error in tests due to JitCache

### What's changed
- Add flag that disables JitCache (for now) 
- re-enable `test_layout.py tests`

Qualified by running entire `pytest test/ttnn-jit/test_layouts.py -v -ra` 10 times in loop and seeing no failures.

Also ran 5x CI pipelines with no failures observed

Issue #5550, opened 2025-10-29
ttnn-jit golden verification


azecevicTT:
tt-mlir commit: `1e307a306a6c332fa032368c2c70c4489e09d7cc`

Update: it seems that the same test is flaky failing for a lot shapes, some examples:
[abs-384-192-(2, 2)](https://github.com/tenstorrent/tt-mlir/runs/53966689911)
[abs-32-192-(5, 0)](https://github.com/tenstorrent/tt-mlir/pull/5506/checks?check_run_id=53966547121)
[abs-384-768-(7, 2)](https://github.com/tenstorrent/tt-mlir/pull/5506/checks?check_run_id=53966547121)
```
_______________ test_l1_block_sharded_shapes[abs-256-672-(6, 1)] _______________
device = MeshDevice(1x1 grid, 1 devices), h = 256, w = 672, max_grid = (6, 1)
op = <function abs at 0x7fd9b11c56c0>
    @pytest.mark.parametrize(
        "h , w, max_grid",
        BLOCK_SHARDED_SHAPE_GRIDS,
        ids=[f"{h}-{w}-{grid}" for h, w, grid in BLOCK_SHARDED_SHAPE_GRIDS],
    )
    @pytest.mark.parametrize("op", [abs])
    def test_l1_block_sharded_shapes(device, h, w, max_grid, op):
        if max_grid in GRIDS_FAILING_ALL_SHAPES:
            pytest.xfail("Grid fails for all shapes. Issue: #5415")
    
>       run_op_test(
            device,
            h,
            w,
            max_grid,
            torch.float16,
            op,
            num_inputs=1,
            buffer_type=ttnn.BufferType.L1,
        )
test/ttnn-jit/test_layouts.py:81: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
device = MeshDevice(1x1 grid, 1 devices), h = 256, w = 672, max_grid = (6, 1)
dtype = torch.float16, op = <function abs at 0x7fd9b11c56c0>, num_inputs = 1
buffer_type = <BufferType.L1: 1>
    def run_op_test(
        device, h, w, max_grid, dtype, op, num_inputs, buffer_type=ttnn.BufferType.L1
    ):
        if buffer_type == ttnn.BufferType.L1:
            inputs = [
[...]

PR #5636, opened 2025-11-01
[ttnn.jit] fix `ProgramDescCache`

vtangTT:
### Ticket
https://github.com/tenstorrent/tt-mlir/issues/5550

follow up fix to https://github.com/tenstorrent/tt-mlir/pull/5589 

### Problem description
ND behaviour in `ProgramDescCache`

### What's changed
Problem:
- cached `ProgramDescriptors` did not take into account possible change in runtime args such as tensor buffer addresses

Fix (short-time)
- construct `ProgramDescCache` hash with tensor buffer addresses included
- this basically means it's very hard to get a cache hit unless tensors are allocated with the same address

Actual fix
- requires `override_runtime_arguments` to be implemented for ttnn.generic op in tt-metal
- when done, also need to have an `overrideArgs` to update the cached ProgramDescriptor in mlir runtime before passing it to ttnn.generic and the ttnn op infrastructure
    - note: there's an impl of this already, can remove and push in the next PR.

New test:
- trying its best to test for this ND behaviour, methodology in comments

The original ND behvaiour found in `test_layouts.py` was also qualified by:
- running the whole suite 100 times, with no failures

### Checklist

### DIFF
--- a/runtime/lib/ttnn/operations/generic/generic_op.cpp
+++ b/runtime/lib/ttnn/operations/generic/generic_op.cpp
@@ -269,7 +269,11 @@ void overrideArgs(
   for (size_t i = 0; i < programDescriptor->cbs.size(); ++i) {
     const auto *cbDesc = programDesc->cbs()->Get(i);
     auto &cb = programDescriptor->cbs[i];
-    cb.buffer = ioTensors[cbDesc->buffer()->tensor_operand_index()].buffer();
+
+    // Not all CBs have a backing L1 buffer.
+    if (cbDesc->buffer()) {
+      cb.buffer = ioTensors[cbDesc->buffer()->tensor_operand_index()].buffer();
+    }
   }
 }
 


runtime/lib/ttnn/operations/generic/generic_op.cpp
```diff
@@ -251,17 +285,28 @@ void run(const ::tt::target::ttnn::GenericOp *op, ProgramContext &context) {
...
   auto *programDesc = op->program();
-  std::size_t hash =
-      ttsl::hash::hash_objects_with_default_seed(programDesc, programDescCache);
+
+  // Note: need to hash tensor buffers: even if we override the arg here during
+  // ProgramDescriptor creation, once the Program is cached, in ttnn, it won't
+  // matter, unless we implement `override_runtime_arguments` in ttnn.generic
+  // op.
+  std::vector<void *> tensorBuffers(ioTensors.size());
+  std::vector<uint32_t> tensorBufferAddresses(ioTensors.size());
```
jnie-TT, line 298: I think if we're pushing back later then we shouldn't preallocate elements, maybe you meant to reserve the size instead of allocating it directly?

vtangTT: oops fixed, ty

vtangTT:
image

Issue #5633, opened 2025-10-31
OnPR hung in main during EmitC tests - batchnorm training op?


jameszianxuTT:
Attempting to repro locally.

jameszianxuTT:
Repro candidate from emitc.sh build + test script:
```
// build script
llvm-lit -sv test/ttmlir/EmitC/TTNN
python tools/ttnn-standalone/ci_compile_dylib.py

// test script
llvm-lit build/test/ttmlir/EmitC/TTNN 
ttrt emitc build/test/ttmlir/EmitC/TTNN --flatbuffer build/test/ttmlir/EmitC/TTNN
```

jameszianxuTT:
Could be batch norm op hanging due to tt-metal bug https://github.com/tenstorrent/tt-mlir/issues/5601

Local repro shows hang at 
```
2025-10-31 20:44:56,033 - INFO - test command for binary=build/test/ttmlir/EmitC/TTNN/batch_norm/batch_norm_training.mlir.ttnn is: /opt/ttmlir-toolchain/venv/bin/ttrt run build/test/ttmlir/EmitC/TTNN/batch_norm/batch_norm_training.mlir.ttnn --program-index all --loops 1 --save-artifacts  --artifact-dir /localdev/jameszianxu/tt-mlir/ttrt-artifacts 
2025-10-31 20:44:57,683 - INFO - set log file=
2025-10-31 20:44:57,683 - INFO - setting artifacts folder path=/localdev/jameszianxu/tt-mlir/ttrt-artifacts
```

PR #5637, opened 2025-11-01
Disable batch_norm_training emitC test due to hang

jameszianxuTT:
#5633


azecevicTT (approved):

jameszianxuTT:
Alternative theory is that hangs are due to discovered failure - see #5639 

I recall a slack thread about these particular tests hanging during bringup, I am not sure what the status of those tests were when merged into main. However, I am pretty sure CI was compromised at that point so the hung tests were not run.

jameszianxuTT:
This does not look like the right solution, testing locally on pure matmul tests
` ttrt emitc build/test/ttmlir/EmitC/TTNN/matmul/ --flatbuffer build/test/ttmlir/EmitC/TTNN/matmul/`

Also exhibits a hang.

jameszianxuTT:
This looks hung in the same place (emitc ttrt tests), so this is not the correct path - https://github.com/tenstorrent/tt-mlir/actions/runs/18998635982/job/54262525052?pr=5637

Pursuing #5639 as alternative hypothesis.

PR #5622, opened 2025-10-31
[CI] Build everything

nsumrakTT:
### Problem description
Optional components makes bugs slip into main

### What's changed
Remove optional components, always build everything

### Checklist
- [ ] New/Existing tests provide coverage for changes

PR #5299, opened 2025-10-09
Adding ttrt emitc

jgrimTT:
### Ticket
Closes [#5298 ](https://github.com/tenstorrent/tt-mlir/issues/5298)

### Problem description
`ttrt` has no mechanism for executing `EmitPy` modules
EmitC is named `ttnn-standalone` in builder. 

### What's changed

Wrote `tools/ttrt/common/emitc.py`.
Added to `api.py` and `CMakeLists.txt`.
Updated utils to accommodate.
Removed `--emitc` from `run.py`.
Relocated repeated infrastructure in `run.py` and `emitpy.py` to `util.py`.
Renamed ttnn-standalone -> EmitC in builder.
Added functions to `dylib.cpp` to allow for running a `.so` module without variables provided by a corresponding flatbuffer run.
Added Documentation.
Cleaned up `emitpy.py`
Added CI test to `builder.sh` and added test for `ttrt emitpy` while I was at it

### Checklist
- [ ] New/Existing tests provide coverage for changes

PR #5639, opened 2025-11-01
Disable emitc.sh ttrt tests due to hang

### LINKED ISSUE #5633: OnPR hung in main during EmitC tests - batchnorm training op?

jameszianxuTT:
### Ticket
Closes #5633 

### Problem description
Skipping hanging testgroup introduced in #5299 but possibly uncaught by bugged test infra not triggering emitc tests, patched by #5622 but now exposing hangs into main.

### What's changed
Disable new ttrt emitc testgroup.

### Checklist

### DIFF
--- a/.github/settings/tests.json
+++ b/.github/settings/tests.json
@@ -34,7 +34,6 @@
   { "runs-on": "n300",   "image": "tracy",  "script": "ttnn_runtime.sh", "args": "runtime/test/ttnn/python/n300" },
   { "runs-on": "n300-llmbox", "image": "tracy",  "script": "ttnn_runtime.sh", "args": "runtime/test/ttnn/python/distributed/llmbox" },
   { "runs-on": "n300",   "image": "tracy",  "script": "runtime_debug.sh" },
-  { "runs-on": "n150",   "image": "speedy",  "script": "emitc.sh" },
   { "runs-on": "n150",   "image": "speedy",  "script": "ttnn_standalone.sh" },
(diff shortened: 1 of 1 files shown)


jameszianxuTT:
This is an alternative to #5637 which might be the wrong solution.

jameszianxuTT:
I would usually be against disabling testgroups, but this particular one was added on 3 days ago (Oct 29) so we didn't have test coverage before that anyways. Also I suspect this test coverage would have hung as soon as it was committed, so it was not providing any coverage at all.

Attaching known datapoints:

image




To test the theory that this testgroup was always hanging, I branch off from the commit `849ecd` and cherry-pick the change `568d75` to build & run all tests and trigger onPR on a limited testset to check if it hangs. [That run is here](https://github.com/tenstorrent/tt-mlir/actions/runs/18999097220).  If it also hangs, that means the testgroup should definitely be reverted as it was never functional. This was not noticed before because these tests were not run due to a separate bug #5614 


edit: [emitc ttrt tests were always broken](https://github.com/tenstorrent/tt-mlir/actions/runs/18999097220/job/54263694352) as shown  from result of onPR run just attempting those tests from the commit in main where they were merged.

jameszianxuTT:
Note - if this passes we can get MLIR main CI to be green. However, it is still not trustworthy, there needs to be a review of existing tests in the pipeline to make sure we are running all the tests we expect.

kmabeeTT:
This is great investigation and initiative, much appreciated @jameszianxuTT  - this case makes sense to force merge right now.

PR #5649, opened 2025-11-02
Allowing empty tensors for reshape

sdjordjevicTT:
### Ticket
Closes https://github.com/tenstorrent/tt-mlir/issues/5643

### Problem description
TTIR and TTNN dialect reshapes didn't allow reshaping of an empty tensor.

### What's changed
Allow reshaping of empty tensor. Added silicon builder test to verify the behavior.

### Checklist

### DIFF
--- a/test/ttmlir/Dialect/TTIR/data_movement/reshape/reshape_tests_negative.mlir
+++ b/test/ttmlir/Dialect/TTIR/data_movement/reshape/reshape_tests_negative.mlir
@@ -39,7 +39,7 @@ module {
   func.func @reshape_infer_dim_negative(%arg0: tensor<2x32x32xbf16>) -> tensor<32x2x32xbf16> {
     %0 = ttir.empty() : tensor<32x2x32xbf16>
     %1 = "ttir.reshape"(%arg0, %0) <{shape = [32: i32, -1: i32, -32: i32]}> : (tensor<2x32x32xbf16>, tensor<32x2x32xbf16>) -> tensor<32x2x32xbf16>
-    // CHECK: error: 'ttir.reshape' op All dimensions must be positive except the one with -1
+    // CHECK: error: 'ttir.reshape' op All dimensions must be >= 0 except the one with -1
     return %1 : tensor<32x2x32xbf16>
   }
 }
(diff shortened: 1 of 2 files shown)


svuckovicTT:
0-sized tensor dim hurts my brain 😅

Interestingly, numpy supports these, e.g. you can do:
```py
>>> np.ones((1, 0, 3))
array([], shape=(1, 0, 3), dtype=float64)
>>> np.ones((1, 2, 3))
array([[[1., 1., 1.],
        [1., 1., 1.]]])
```

But there doesn't seem to be a way to retrieve any values from a tensor with 0-sized dim - it has no values by definition (i.e. its volume is `x0 * x1 * ... * 0 * ... * xn = 0`).

While these are useful for concatenation, dynamic tensor shaping and similar scenarios, they don't seem to be scenarios that TT-NN can handle, so I think we shouldn't support these in the compiler without at least having a wider discussion and/or talking to the TT-NN folks to see what their intentions are here...

```py
>>> w = ttnn.ones((1, 0, 3))
>>> w
ttnn.Tensor([[]], shape=Shape([1, 0, 3]), dtype=DataType::BFLOAT16, layout=Layout::ROW_MAJOR)
>>> e = ttnn.ones((0, 5, 0))
>>> ttnn.concat(e, w)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/localdev/svuckovic/_workspace/repos/tt-mlir/third_party/tt-metal/src/tt-metal/ttnn/ttnn/decorators.py", line 377, in __call__
    result = self.function(*function_args, **function_kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: object of type 'ttnn._ttnn.tensor.Tensor' has no len()
```

EDIT:

I take it back, looks like dim0 tensors work out in TT-NN.

```py
>>> d = ttnn.open_device(device_id=0)
2025-11-03 09:16:26.940 | warning  |           Metal | Unable to bind worker thread to CPU Core. May see performance degradation. Error Code: 22 (hardware_command_queue.cpp:74)
>>> q = ttnn.ones((1, 0, 3), device=d)
>>> w = ttnn.ones((1, 5, 3), device=d)
>>> e = ttnn.concat([q, w], dim=1)
>>> e
ttnn.Tensor([[[ 1.7173e-38,  0.0000e+00,  0.0000e+00],
              [ 1.0000e+00,  1.0000e+00,  1.0000e+00],
              ...,
              [ 1.0000e+00,  1.0000e+00,  1.0000e+00],
              [ 1.0000e+00,  1.0000e+00,  1.0000e+00]]], shape=Shape([1, 5, 3]), dtype=DataType::BFLOAT16, layout=Layout::ROW_MAJOR)
```

So looks like they support this today :)

sdjordjevicTT:
Yeah, this PR doesn't come before I tried it on pure TTNN :D I added one reshape test that proves that it works on TTNN as well. I also tried a couple of more ops to check, and they all passed.

This comes as a repair item from the speech t5 failing model, as it has some weird indexing, and we are getting this IR:
https://github.com/tenstorrent/tt-mlir/issues/5643

I can try running it with enable-aggressive-simplification=true, but this is by default false, hence we are getting this IR in TTIR.

PR #5652, opened 2025-11-03
Move stablehlo like scatterOp decomposition to stablehlo to ttir conversion

djstefanovicTT:
### Ticket
Part of https://github.com/tenstorrent/tt-mlir/issues/5466

### Problem description
StableHLO scatterOp and TTIR scatterOp have the same semantics, but TTNN scatterOp has different. Instead of decomposition of TTIR version, this is moved directly to stableHLO patterns.

### What's changed
Moved all patterns that decompose `ttir.scatter` to stableHLO to TTIR pipeline. Updated all tests to accommodate this change. Changed all ops that use `ttir.scatter` in their decomposition to target directly the result of previous decomposition. 

### Checklist

### DIFF
--- a/lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
+++ b/lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
@@ -3207,12 +3207,11 @@ public:
               sliceOutputShape, updatesType.getElementType(), nullptr);
 
           // Create slice op.
-          auto slicedUpdates =
-              rewriter.create<ttir::SliceStaticOp>(
-                  scatterOp.getLoc(), slicedUpdatesType, updates,
-                  rewriter.getI32ArrayAttr(sliceStarts),
-                  rewriter.getI32ArrayAttr(sliceEnds),
-                  rewriter.getI32ArrayAttr(sliceSteps));
+          auto slicedUpdates = rewriter.create<ttir::SliceStaticOp>(
+              scatterOp.getLoc(), slicedUpdatesType, updates,
+              rewriter.getI32ArrayAttr(sliceStarts),
+              rewriter.getI32ArrayAttr(sliceEnds),
+              rewriter.getI32ArrayAttr(sliceSteps));
           // create fill cache op for this batch.
           cache = rewriter.create<mlir::tt::ttir::FillCacheOp>(
               scatterOp.getLoc(),
@@ -3391,8 +3390,8 @@ public:
 
       // Create ScatterInDimOp.
       rewriter.replaceOpWithNewOp<ttir::ScatterOp>(
-          srcOp, outputType, inputTensor, finalIndexTensor,
-          updateTensor, dimAttr);
+          srcOp, outputType, inputTensor, finalIndexTensor, updateTensor,
+          dimAttr);
       return success();
     }
 
@@ -3421,9 +3420,8 @@ public:
           mlir::cast<RankedTensorType>(flattenedInput.getType());
 
       Value scatterResult = rewriter.create<ttir::ScatterOp>(
-          srcOp.getLoc(), flattenedInputType.getShape(),
-          flattenedInputType.getElementType(), flattenedInputType.getEncoding(),
-          flattenedInput, finalIndexTensor, flattenedUpdate, dimAttr);
+          srcOp.getLoc(), flattenedInputType, flattenedInput, finalIndexTensor,
+          flattenedUpdate, dimAttr);
 
       // Reshape result back to original input shape.
       Value reshapedResult =
@@ -3655,11 +3653,13 @@ private:
       auto endsAttr = rewriter.getI32ArrayAttr(ends);
       auto stepsAttr = rewriter.getI32ArrayAttr(steps);
 
+      RankedTensorType sliceResultType = RankedTensorType::get(
+          sliceShape, indexType.getElementType(), indexType.getEncoding());
+
       Value dimensionIndices = rewriter.create<ttir::SliceStaticOp>(
           ttmlir::utils::appendLocationSuffix(
               op.getLoc(), "_dim_" + std::to_string(dim) + "_slice"),
-          sliceShape, indexType.getElementType(), indexType.getEncoding(),
-          indexTensor, beginsAttr, endsAttr, stepsAttr);
+          sliceResultType, indexTensor, beginsAttr, endsAttr, stepsAttr);
 
       // Multiply by stride if stride > 1.
       if (strides[dim] > 1) {
@@ -3674,22 +3674,26 @@ private:
                 op.getLoc(), "_stride_" + std::to_string(dim)),
             dimIndexType, scalarAttr);
 
+        RankedTensorType multiplyResultType = RankedTensorType::get(
+            sliceShape, indexType.getElementType(), indexType.getEncoding());
+
         dimensionIndices = rewriter.create<ttir::MultiplyOp>(
             ttmlir::utils::appendLocationSuffix(
                 op.getLoc(), "_dim_" + std::to_string(dim) + "_stride_mul"),
-            sliceShape, indexType.getElementType(), indexType.getEncoding(),
-            dimensionIndices, strideTensor);
+            multiplyResultType, dimensionIndices, strideTensor);
       }
 
       // Add to flat indices.
       if (flatIndices == nullptr) {
         flatIndices = dimensionIndices;
       } else {
+        RankedTensorType addResultType = RankedTensorType::get(
+            sliceShape, indexType.getElementType(), indexType.getEncoding());
+
         flatIndices = rewriter.create<ttir::AddOp>(
             ttmlir::utils::appendLocationSuffix(
                 op.getLoc(), "_add_dim_" + std::to_string(dim)),
-            sliceShape, indexType.getElementType(), indexType.getEncoding(),
-            flatIndices, dimensionIndices);
+            addResultType, flatIndices, dimensionIndices);
       }
     }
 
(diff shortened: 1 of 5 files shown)


sdjordjevicTT:
Can we remove TTIR scatterOp from the TTIR dialect as well? :)

test/ttmlir/Dialect/TTNN/simple_scatter.mlir
```diff
@@ -1,23 +1,26 @@
-// RUN: ttmlir-opt --ttir-to-ttnn-backend-pipeline -o %t %s
+// RUN: ttmlir-opt --stablehlo-to-ttir-pipeline --ttir-to-ttnn-backend-pipeline -o %t %s
```
mmanzoorTT: Same comment. Use `ttir.scatter_in_dim` op instead of `stablehlo.scatter` op in the test.

test/ttmlir/Conversion/StableHLOToTTIR/kv_cache_fusing.mlir
```diff
@@ -1,4 +1,4 @@
-// RUN: ttmlir-opt -ttir-fusing -o %t %s
+// RUN: ttmlir-opt --stablehlo-to-ttir-pipeline -ttir-fusing -o %t %s
```
mmanzoorTT: My suggestion is to not use `--stablehlo-to-ttir-pipeline` here. Use `ttir.scatter_in_dim` op instead of `stablehlo.scatter` op in the test. This is testing TTIR fusion pass; so adding a different pipeline (stablehlo) doesn't seem right.

test/ttmlir/Silicon/TTNN/n150/data_movement/scatter/scatter.mlir
```diff
@@ -1,4 +1,4 @@
-// RUN: ttmlir-opt --ttir-to-ttnn-backend-pipeline="system-desc-path=%system_desc_path%" -o %t.mlir %s
+// RUN: ttmlir-opt --stablehlo-to-ttir-pipeline --ttir-to-ttnn-backend-pipeline="system-desc-path=%system_desc_path%" -o %t.mlir %s
```
mmanzoorTT: Same here. use `ttir.scatter_in_dim` op.

test/ttmlir/Conversion/StableHLOToTTIR/kv_cache_fusing.mlir
djstefanovicTT: Perhaps this test should be moved somewhere else then? It could not work with `ttir.scatter_in_dim` op because the fusing logic now happens in stableHLOtoTTIR. `ttir.scatter` op was equivalent to `stablehlo.scatter` op, so not much was changed. The other two test can be changed to use `ttir.scatter_in_dim` op.

test/ttmlir/Conversion/StableHLOToTTIR/kv_cache_fusing.mlir
```diff
@@ -0,0 +1,132 @@
...
```
mmanzoorTT: Do we still need `ttir-fusing` pass here? Fusion logic is moved to stablehlo->to->ttir conversion. right?

mmanzoorTT (commented):
is there any particular reason to keep `scatter_in_dim` op? why not to rename it as `scatter` op?

test/ttmlir/Conversion/StableHLOToTTIR/kv_cache_fusing.mlir
mmanzoorTT: you are right. you should remove `-ttir-fusing`; this pass is not doing any fusion for scatter op.

test/ttmlir/Conversion/StableHLOToTTIR/kv_cache_fusing.mlir
djstefanovicTT: Fusing pass removes additions and multiplications and so on, it is not strictly necessary.

lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
```diff
@@ -2693,6 +2733,320 @@ class StableHLOToTTIROpIotaOpConversionPattern
...
+  }
+
+  static bool isEffectively1D(ArrayRef<int64_t> shape) {
+    return llvm::count_if(shape, [](int64_t dim) { return dim != 1; }) <= 1;
+  }
+};
+} // namespace
+
+namespace {
+class CacheFillUpdatePattern
```
vmilosevic: :warning: **clang\-diagnostic\-error** :warning:
redefinition of `` CacheFillUpdatePattern ``

test/ttmlir/Silicon/TTNN/n150/data_movement/scatter/scatter.mlir
```diff
@@ -6,34 +6,47 @@
 // shape(input) == shape(output)
 // single index dimension (len(scatter_dims_to_operand_dims) == 1)
 
-func.func @scatter_simple_1(%arg0: tensor<1x3x320x320xf32>, %arg1: tensor<1x1xi32>, %arg2: tensor<1x3x32x32xf32>) -> tensor<1x3x320x320xf32> {
```
sdjordjevicTT, line 9: Just double-check if we have this test coverage in stablehlo tests.

test/ttmlir/Silicon/TTNN/n150/data_movement/scatter/scatter.mlir
```diff
@@ -6,34 +6,47 @@
 // shape(input) == shape(output)
```
sdjordjevicTT, line 6: Can we also move this silicon test to a ttir builder test?

lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
```diff
@@ -148,6 +149,44 @@ parseFloatFromStringAttr(const mlir::StringAttr &stringAttr, float &result) {
   }
   return success();
 }
+// Helper function to create a reshape operation.
+static ttir::ReshapeOp createReshapeOp(PatternRewriter &rewriter, Location loc,
```
mmanzoorTT: I suggest to move it to `ttir::utils` so it is accessible everywhere.

lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
```diff
@@ -148,6 +149,44 @@ parseFloatFromStringAttr(const mlir::StringAttr &stringAttr, float &result) {
...
+  auto shapeAttr =
+      rewriter.getI32ArrayAttr(llvm::SmallVector<int32_t>(targetShape));
+
+  return ttir::utils::createDPSOp<ttir::ReshapeOp>(
+      rewriter, loc, targetShape, inputType.getElementType(),
+      inputType.getEncoding(), input, shapeAttr);
+}
+
+// Helper function to flatten a tensor to 1D.
+static Value flattenTensor(PatternRewriter &rewriter, Location loc,
```
mmanzoorTT: I suggest to move it to `ttir::utils` so it is accessible everywhere.

lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
```diff
@@ -3116,6 +3155,238 @@ class StableHLOToTTIROpIotaOpConversionPattern
 };
 } // namespace
 
+namespace {
+class CacheFillUpdatePattern
```
mmanzoorTT, line 3120: It was proposed to move fusion patterns to TTNN dialect. I am not sure about moving this fusion pattern to stablehlo conversion. Thoughts? @nsmithtt @sdjordjevicTT @azecevicTT

sdjordjevicTT: I wouldn't block on this. It seems much easier to implement in the stablehlo dialect area of abstraction than at the TTNN level. Regardless of our choice, I prefer to proceed with this change for now, and we can later adjust it as needed when we determine the appropriate level.

mmanzoorTT: sounds good

lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
ajakovljevicTT: I would also add a more detailed comment about what the inputs into flattenTensor are (specifically for "suffix")

lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
```diff
@@ -3116,6 +3155,238 @@ class StableHLOToTTIROpIotaOpConversionPattern
...
+          // Slice steps is 1 for every dim as we do not wish to skip any
+          SmallVector<int32_t> sliceSteps = {1, 1, 1, 1};
+
+          // Slice output shape is the same as the fill value shape, except the
+          // batch dim is 1 since we sliced out a single batch.
+          SmallVector<int64_t> sliceOutputShape(updatesType.getShape());
+          sliceOutputShape[0] = 1;
+
+          // Encoding should not be set when this pass is run. Guard against it.
+          assert(!updatesType.getEncoding());
```
ajakovljevicTT: Add an assert message here

lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
```diff
@@ -3116,6 +3155,238 @@ class StableHLOToTTIROpIotaOpConversionPattern
...
+      // Unlike ttnn.fill_cache, we can perform ttnn.update_cache on the entire
+      // batch at once. However this requires that the fill value is in the form
+      // [1, num_heads, B, head_size]. So, we must permute the updates tensor to
+      // this shape.
+      if (batchSize > 1) {
+        SmallVector<int64_t> permutedShape = ttmlir::utils::applyPermutation(
+            updatesType.getShape(), {2, 1, 0, 3});
+
+        // Encoding should not be set when this pass is run. Guard against it.
+        assert(!updatesType.getEncoding());
```
ajakovljevicTT: Again, missing a comment

Issue #5655, opened 2025-11-03
TTIR gather -> embedding op decomposition generality


sdjukicTT:
I am opening this issue to note what is currently supported in TTIR to TTIR decomposition for `ttir.gather` to `ttir.embedding`.

There are multiple cases of gather op based on slice sizes of dimensions that are being indexed (dims in `start_index_map`). Dims that are being indexed fall into one of three categories:
1. 1/N - slice size of 1, selecting one row
2. N/N - slice size is the full dim, so there is no actual indexing, it is the same as non-indexed dims
3. M/N - slice size M > 1, we select a few rows

TTIR to TTIR gather decomposition currently supports a few cases:
1. If all indexed dims are in 1. category (1/N), it can be mapped to `ttnn.embedding` without extra steps related to `start_index_map`
2. If there is at least one dim from 2. category (N/N), dims that are taken whole are removed from `start_index_map` and treated as non-indexed dims. This requires slicing the start_indices tensor to remove the extra dims. Currently we support only the case where exactly one dim is left after removing N/N dims from `start_index_map`. This is because if there were more dims, it would introduce complexity, with possible additional slice and concat ops to get only the meaningful indices from `start_index_map`.
3. If there is a dim from 3. category (M/N) we support that case if it is the only dim actually indexed (that isn't taken whole, isn't N/N). This is because this case requires expansion of start_indices. For example if indices were `[2, 0, 5]` and slice size for that dim is `2`, we need to expand indices to `[[2, 3], [0, 1], [5, 6]]`. Handling other dims together with this expansion would introduce complexity.

As for dimensions that aren't being indexed, we support them being taken whole (slice size equal to dim size in input, N/N).

Also, TTIR to TTIR decomposition ignores that indices that are out of range are expected to be clamped (https://openxla.org/stablehlo/spec#gather). We assume that cases we get from models will not have out of range indices. This would always add a clamp op to the decomposition (because in compile time it is not known whether it is needed or not).

For full generality:
- support for any combination of slice sizes should be added
- clamp indices to range

As of now, there hasn't been a case that needs a clamp op. There were a few cases of unsupported combination of slice sizes. So this is not high priority, opening an issue just to note what we support.

Related issues:
https://github.com/tenstorrent/tt-mlir/issues/4138
https://github.com/tenstorrent/tt-xla/issues/318
https://github.com/tenstorrent/tt-mlir/issues/5660

mmilosevicTT:
[...]

PR #5656, opened 2025-11-03
Fix handling of singleton indexed dims in TTIR gather

sdjukicTT:
### Ticket
Related, notes what is supported https://github.com/tenstorrent/tt-mlir/issues/5655

### Problem description
Dimensions that are in `start_index_map` but have size 1 (and slice size 1) can be treated both as dims that are taken whole (N/N) and dims that are normally indexed (1/N). Currently they were always treated as N/N dims, but greater generality is accomplished if we sometimes treat them as 1/N.
 
### What's changed
Instead of always removing 1/1 dims from `start_index_map`, we remove them if:
1. there is a partially indexed dim (M/N) and there can be no other indexed dims
2. all indexed dims are 1/1 or N/N so we remove all but one 1/1 dim.

### Checklist


sdjukicTT:
I have cherry-picked changes from https://github.com/tenstorrent/tt-mlir/pull/5684 to solve both problems at once. Thank you @mmanzoorTT for solving the problem.

Issue #5646, opened 2025-11-01
[TTIR] Fix non legal Reverse decomposition


sdjordjevicTT:
This fails to compile when TTIR.Reverse op is before TTIR.ConvTranspose2d op, because in this case, we are leaving the reverse op in the IR, but due to some failure in the conv op validation, it stays in IR, and as it is non-legal, the compile fails. We should mark TTIR.Reverse op conditionally legal based on whether it is before conv op.

Example failure IR:
```MLIR
module {
  func.func @main_conv_reverse(%arg_input: tensor<1x64x1x6400xf32>, %arg_weight: tensor<1x8x32x64xf32>, %arg_bias: tensor<32xf32>) -> tensor<1x32x1x25600xf32> {
    %999 = ttir.empty() : tensor<1x8x32x64xf32>
    %1000 = "ttir.reverse"(%arg_weight, %999) <{dimensions = array<i64: 0, 1>}> : (tensor<1x8x32x64xf32>, tensor<1x8x32x64xf32>) -> tensor<1x8x32x64xf32>
    %1001 = ttir.empty() : tensor<1x32x1x25600xf32>
    %1002 = ttir.empty() : tensor<1x32x1x1xf32>
    %1003 = "ttir.reshape"(%arg_bias, %1002) <{shape = [1 : i32, 32 : i32, 1 : i32, 1 : i32]}> : (tensor<32xf32>, tensor<1x32x1x1xf32>) -> tensor<1x32x1x1xf32>
    %1004 = "ttir.convolution"(%arg_input, %1000, %1003, %1001) <{
        batch_group_count = 1 : i64,
        convolution_layout = #ttir<
            convolution_layout 
            input_batch = 0, // N[0] -> 1
            input_feature = 1, // C[1] -> 64
            input_spatial_dimensions = 2x3, // 2x3 HxW (1x6400)
            kernel_output_feature = 2,
            kernel_input_feature = 3,
            kernel_spatial_dimensions = 0x1, // 0x1 KHxKW (1x8)
            output_batch = 0,
            output_feature = 1,
            output_spatial_dimensions = 2x3>,
        feature_group_count = 1 : i64,
        input_dilation = array<i64: 1, 4>,
        padding = array<i64: 0, 0, 5, 5>,
        weight_dilation = array<i64: 1, 1>,
        window_reversal = array<i1: false, false>,
[...]

PR #5657, opened 2025-11-03
Fixing padding logic for transposed convolution

djstefanovicTT:
### Ticket
https://github.com/tenstorrent/tt-mlir/issues/5646
https://github.com/tenstorrent/tt-mlir/issues/5647

### Problem description
 TTIR.Convolution op decomposes into the TTIR.ConvTransposed2d op with wrong order of padding. If a reverseOP precedes such faulty convolution, it will stay in the IR even though it is marked illegal. 

### What's changed
Added partial legality to `ttir.reverseOp` in ttir decomposition. Canonicalized removes these leftover operations. Fixed padding order.

### Checklist

### DIFF
--- a/lib/Conversion/TTIRToTTIRDecomposition/TTIRToTTIRDecomposition.cpp
+++ b/lib/Conversion/TTIRToTTIRDecomposition/TTIRToTTIRDecomposition.cpp
@@ -618,11 +618,7 @@ private:
       ConversionPatternRewriter &rewriter, ttir::ConvolutionOp op,
       OpAdaptor adaptor, mlir::tt::ttir::ConvolutionLayoutAttr convLayoutAttr,
       Value input, Value weight, llvm::ArrayRef<int64_t> outputShape) const {
-    // Determine if the stablehlo.convolution op represents a regular or
-    // transposed convolution, based on Torch-MLIR lowering patterns:
-    // https://github.com/llvm/torch-mlir/blob/main/lib/Conversion/TorchToStablehlo/Linear.cpp
-    // and XLA patterns: convolution is transposed if the input dilation is
-    // greater than 1.
+
     bool isTransposed = ttir::utils::isTransposedConv(op);
 
     auto strideAttr = rewriter.getDenseI32ArrayAttr({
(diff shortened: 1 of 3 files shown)


lib/Conversion/TTIRToTTIRDecomposition/TTIRToTTIRDecomposition.cpp
```diff
@@ -729,10 +719,6 @@ struct ConvolutionToConv2dPattern : public ConvolutionDecompositionPattern {
       // conv_transposed2d op.
       // https://github.com/llvm/torch-mlir/blob/main/lib/Conversion/TorchToStablehlo/Linear.cpp
       paddingAttr = rewriter.getDenseI32ArrayAttr({
```
sdjordjevicTT, line 717: Can you please add a comment order in which the padding is expected in the TTIR op?

lib/Conversion/TTIRToTTIRDecomposition/TTIRToTTIRDecomposition.cpp
```diff
@@ -623,17 +623,7 @@ struct ConvolutionToConv2dPattern : public ConvolutionDecompositionPattern {
     // https://github.com/llvm/torch-mlir/blob/main/lib/Conversion/TorchToStablehlo/Linear.cpp
```
sdjordjevicTT: Can we move this comment above the helper function?

lib/Conversion/TTIRToTTIRDecomposition/TTIRToTTIRDecompositionPass.cpp
```diff
@@ -119,6 +119,21 @@ struct TTIRToTTIRDecompositionPass
       return (dimArg->size() == 1 || dimArg->size() == rank);
     });
 
+    target.addDynamicallyLegalOp<ttir::ReverseOp>([&](ttir::ReverseOp op) {
+      // Only decompose if not used by a non-transposed convolution
```
svuckovicTT: If I'm parsing this correctly, this comment doesn't seem accurate. The code says that it's okay to mark legal (i.e. not decompose) if reverse is used by a transposed conv (not non-transposed).

djstefanovicTT: It should say "Only decompose if not used by a transposed convolution."

PR #5659, opened 2025-11-03
Enable EmitC tests on CI

azecevicTT:
A lot of bugs slipped through on EmitC tests during the period of (partial) inactivity of them on CI:
- `ttrt emitc` expects `create_inputs_for_*`; tests that didn't have inputs skipped creating this function
- const eval functions were treated as forward functions during forward function discovery in `ttrt emitc`
- because of the way Python GC works, sometimes the code from the `.so` would be called after that `.so` is closed


jgrimTT (commented):
Lgtm!

tapspatel (approved):

Issue #5645, opened 2025-11-01
[TTNN] Fix LinearOp workaround


sdjordjevicTT:
This TTIR graph fails to compile:
```MLIR
module {
  func.func @main_right_transpose(%arg_a : tensor<12x24x64xf32>, %arg_b : tensor<12x24x64xf32>, %arg_bias : tensor<12x24x24xf32>) -> tensor<12x24x24xf32>{
    %0 = ttir.empty() : tensor<12x24x24xf32>
    %1 = "ttir.linear"(%arg_a, %arg_b, %arg_bias, %0) <{transpose_a = false, transpose_b = true}> : (tensor<12x24x64xf32>, tensor<12x24x64xf32>, tensor<12x24x24xf32>, tensor<12x24x24xf32>) -> tensor<12x24x24xf32>
    return %1 : tensor<12x24x24xf32>
  }

  func.func @main_left_transpose(%arg_a : tensor<12x24x64xf32>, %arg_b : tensor<12x24x64xf32>, %arg_bias : tensor<12x64x64xf32>) -> tensor<12x64x64xf32>{
    %0 = ttir.empty() : tensor<12x64x64xf32>
    %1 = "ttir.linear"(%arg_a, %arg_b, %arg_bias, %0) <{transpose_a = true, transpose_b = false}> : (tensor<12x24x64xf32>, tensor<12x24x64xf32>, tensor<12x64x64xf32>, tensor<12x64x64xf32>) -> tensor<12x64x64xf32>
    return %1 : tensor<12x64x64xf32>
  }
}
```

The compile fails in the workaround pass because it doesn't properly handle linear op with transpose attributes.

Issue #5648, opened 2025-11-02
[TTNN] Add an option to disable workaround pass


sdjordjevicTT:
This can be used in builder tests in the following way:
- Run the op test with this pipeline option disabled
- The test will run the code without a workaround
- Add xfail flag to this test so when the metal folks fix this issue, we will know on the uplift with the xpass

PR #5661, opened 2025-11-03
Fixing linearOp workaround to include transpose attributes

sdjordjevicTT:
### Ticket
Closes https://github.com/tenstorrent/tt-mlir/issues/5645
Closes https://github.com/tenstorrent/tt-mlir/issues/5662

### Problem description
Linear op workaround fails to calculate output shape when transpose attributes are provided to the op.

### What's changed
Handle the shape calculation when inputs are transposed. Added a test to verify the new behavior.

### Checklist

### DIFF
--- a/test/ttmlir/Dialect/TTNN/linear/linear_tests_positive.mlir
+++ b/test/ttmlir/Dialect/TTNN/linear/linear_tests_positive.mlir
@@ -66,6 +66,7 @@ module {
     %1 = "ttir.linear"(%arg0, %arg1, %bias, %0) <{transpose_a = true}> : (tensor<64x128xbf16>, tensor<64x128xbf16>, tensor<128x128xbf16>, tensor<128x128xbf16>) -> tensor<128x128xbf16>
     return %1 : tensor<128x128xbf16>
   }
+
   func.func @linear_2d_2d_transpose_bias(%arg0: tensor<64x128xbf16>, %arg1: tensor<64x128xbf16>, %bias: tensor<64x64xbf16>) -> tensor<64x64xbf16> {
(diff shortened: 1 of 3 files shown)


test/ttmlir/Dialect/TTNN/linear/linear_tests_positive.mlir
```diff
@@ -88,4 +90,24 @@ module {
     %1 = "ttir.linear"(%arg0, %arg1, %bias, %0) <{transpose_a = true, transpose_b = true}> : (tensor<64x128xbf16>, tensor<128x64xbf16>, tensor<128x128xbf16>, tensor<128x128xbf16>) -> tensor<128x128xbf16>
     return %1 : tensor<128x128xbf16>
   }
+
+  func.func @main_batch_linear_with_bias_right_transpose(%arg_a : tensor<12x24x64xf32>, %arg_b : tensor<12x24x64xf32>, %arg_bias : tensor<12x24x24xf32>) -> tensor<12x24x24xf32>{
```
mtopalovicTT, line 94: We should consider adding flag for workarounds so that we can disable them in builder tests. Then we could test that we are getting failure from ttnn and once they fix the op we will get clear signal from test that workaround needs to be removed.

sdjordjevicTT: I created this issue over the weekend :)
https://github.com/tenstorrent/tt-mlir/issues/5648

sdjordjevicTT: I actually implemented it here :)
https://github.com/tenstorrent/tt-mlir/pull/5667

PR #5667, opened 2025-11-03
Adding pipeline option to disable workaround pass

sdjordjevicTT:
### Ticket
Closes https://github.com/tenstorrent/tt-mlir/issues/5648

### Problem description
We didn't have the option to disable workaround passes in the ttir-to-ttnn pipeline. It is useful for a developer to write tests without any workarounds applied and to mark those tests with xfail.

### What's changed
Introduced a pipeline option to enable a workaround pass. This option defaults to true but can be set to false in builder tests. One builder test was added to demonstrate this usage.

### Checklist

### DIFF
--- a/include/ttmlir/Dialect/TTNN/Pipelines/TTNNPipelines.h
+++ b/include/ttmlir/Dialect/TTNN/Pipelines/TTNNPipelines.h
@@ -215,10 +215,11 @@ struct TTIRToTTNNBackendPipelineOptions
 
   // Option to enable/disable the workaround pass.
   //
-  Option<bool> workaroundsEnabled{
-      *this, "enable-workaround-pass",
-      llvm::cl::desc("An option to disable/enable the whole workaround pass."),
-      llvm::cl::init(true)};
+  Option<bool> disableWorkarounds{
+      *this, "disable-workarounds",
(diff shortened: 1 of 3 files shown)


include/ttmlir/Dialect/TTNN/Pipelines/TTNNPipelines.h
```diff
@@ -215,6 +215,11 @@ struct TTIRToTTNNBackendPipelineOptions
 
   // Option to enable/disable the workaround pass.
   //
+  Option<bool> workaroundsEnabled{
+      *this, "enable-workaround-pass",
```
svuckovicTT:
```suggestion
  Option<bool> workaroundsEnabled{
      *this, "workarounds-enabled",
```

Having the same name for the option in code and as argument makes it so much easier to find and reason about these.

svuckovicTT: Or even better, since workarounds are enabled by default, rename to `disable-workarounds` which is set to false by default.

sdjordjevicTT: Makes sense, let me push the new version :)

PR #5681, opened 2025-11-03
Implement `emitpy` Execution Via `pytest` & Builder

ctodTT:
Added emitpy execution (with golden checks) to builder_runtime

### DIFF
--- a/test/python/golden/ttir_ops/convolution/test_conv_transpose2d.py
+++ b/test/python/golden/ttir_ops/convolution/test_conv_transpose2d.py
@@ -124,6 +124,10 @@ def test_conv_transpose2d(
         pytest.xfail(
             "Metal issue: https://github.com/tenstorrent/tt-metal/issues/33449"
         )
+    if test_id == "emitpy-bf16-batch16_segmentation_training":
+        pytest.xfail(
+            "Out of Memory: Not enough space to allocate 99328 B L1_SMALL buffer across 64 banks, where each bank needs to store 1552 B, but bank size is only 32768 B"
+        )
 
     if bias_shape:
         input_shapes = [input_shape, weight_shape, bias_shape]


jgrimTT (approved):

nsumrakTT (approved):

tools/builder/base/builder_utils.py
```diff
@@ -37,6 +37,19 @@
...
+        self.name = file_path
+
+        # temporary state value to check if test failed
+        self.test_result = "pass"
+
+    @staticmethod
+    def get_py_file_extension():
+        return ".py"
+
+
```
jgrimTT, line 46: Why use a class here instead of passing just the file path into the execute fn like `execute_fb`?

vvukomanTT (approved):

tools/builder/base/builder_runtime.py
```diff
@@ -565,39 +621,130 @@ def execute_fb(
...
+                    target_layout = template_input.layout
+                    target_device = template_input.device()
+                    # target_device = template_input.device()
+                    corrected_inputs.append(
+                        ttnn.as_tensor(
+                            golden_input,
+                            dtype=template_input.dtype,
+                            layout=target_layout,
+                            device=target_device,
+                            memory_config=ttnn.MemoryConfig(
```
tapspatel: can we always guarantee that this is the memory config of the input?

tools/builder/base/builder_runtime.py
```diff
@@ -565,39 +621,130 @@ def execute_fb(
...
+                for golden_input_dict, template_input in zip(
+                    golden_torch_tensors.values(), inputs
+                ):
+                    # Use the layout and device from the template_input
+                    golden_input = golden_input_dict[0]
+                    target_layout = template_input.layout
+                    target_device = template_input.device()
+                    # target_device = template_input.device()
+                    corrected_inputs.append(
+                        ttnn.as_tensor(
```
tapspatel, line 672: lets use our mlir runtime pybinded APIs and limit the amount of TTNN calls we do

tools/builder/base/builder_runtime.py
```diff
@@ -565,39 +621,130 @@ def execute_fb(
...
+    if not disable_golden:
+        golden_outputs_torch = []
+        for i in range(len(outputs)):
+            golden_outputs_torch.append(golden_torch_tensors[f"output_{i}"][0])
+
+        # Convert program outputs to torch tensors
+        output_tensors_torch = []
+        for i, output in enumerate(outputs):
+            # Transfer from device to host and convert to torch
+            output_host = ttnn.from_device(output)
```
tapspatel, line 691: use mlir runtime binded api

tools/builder/base/builder_runtime.py
jgrimTT: In this case, I don't think we can. Emitpy modules take `<class 'ttnn._ttnn.tensor.Tensor'>` instead of the runtime type `<class '_ttmlir_runtime.runtime.Tensor'>`

tools/builder/base/builder_runtime.py
jgrimTT: Addressed, thx

Issue #5663, opened 2025-11-03
error: failed to legalize operation 'ttir.reverse' that was explicitly marked illegal


mmilosevicTT:
### Describe the bug

After resolution of stablehlo.pad in https://github.com/tenstorrent/tt-mlir/issues/5305 RegNet test started failing with 
```
error: failed to legalize operation 'ttir.reverse' that was explicitly marked illegal 
```

In the continuation is the output of the test.

[regnet_reverse_output.txt](https://github.com/user-attachments/files/23303604/regnet_reverse_output.txt)

### Steps to reproduce the issue

1. Go to TT-XLA repo.
2. Run following command `pytest --runxfail tests/jax/single_chip/models/regnet/regnet_y_040/test_regnet_y_040.py`

djstefanovicTT:
@sdjordjevicTT If any operation breaks mlir translation before reverseOps are transformed, the conversion stops and reverseOp remains. If convolution cannot be properly translated, reverse op remains and IR fails to legalize. The problem that occurs is likely because stablehlo convolution supports negative values for padding, but TTIR does not.
 ```
error: 'ttir.conv2d' op Padding attribute values must be >= 0.
    %305 = "ttir.convolution"(%arg699, %303, %304) <{batch_group_count = 1 : i64, convolution_layout = #ttir<convolution_layout input_batch = 3, input_feature = 0, input_spatial_dimensions = 1x2, kernel_output_feature = 3, kernel_input_feature = 0, kernel_spatial_dimensions = 1x2, output_batch = 2, output_feature = 3, output_spatial_dimensions = 0x1>, feature_group_count = 1 : i64, input_dilation = array<i64: 1, 1>, padding = array<i64: 0, -1, 0, -1>, weight_dilation = array<i64: 2, 2>, window_reversal = array<i1: false, false>, window_strides = array<i64: 1, 1>}> : (tensor<1x14x14x512xf32>, tensor<1x7x7x1088xf32>, tensor<1x1x512x1088xf32>) -> tensor<1x1x512x1088xf32>
```
This is the error that happens if `ttir.reverseOp` is marked as legal.

sdjordjevicTT:
Yep, the PR that you made should uncover these cases more easily:
https://github.com/tenstorrent/tt-mlir/pull/5657

sdjordjevicTT:
Renamed the issue title based on this :)

vladimirjovanovicTT:
Closed by #5691

PR #5691, opened 2025-11-04
Negative padding in stableHLO convolution.

djstefanovicTT:
### Ticket
https://github.com/tenstorrent/tt-mlir/issues/5663

### Problem description
StableHLO convolution supports negative padding and TTIR convolution does not.

### What's changed
Checks if convolution has negative values for pad. If that is the case, add slice op in the end and set convolutions pad to 0.

### Checklist


lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
```diff
@@ -1348,10 +1348,43 @@ class StableHLOToTTIRConvolutionOpConversionPattern
...
+    bool hasNegativePadding = false;
+    ArrayRef<int64_t> paddingArray = paddingAttr.asArrayRef();
+    SmallVector<int64_t> adjustedPadding;
+    for (int64_t pad : paddingArray) {
+      if (pad < 0) {
+        hasNegativePadding = true;
+        pad = 0;
+      }
+      adjustedPadding.push_back(pad);
+    }
```
mmanzoorTT:
```suggestion
    bool hasNegativePadding = llvm::any_of(paddingArray, [](int64_t p) { return p < 0; });
    ArrayRef<int64_t> paddingArray = paddingAttr.asArrayRef();
    SmallVector<int64_t> adjustedPadding = llvm::to_vector(
    llvm::map_range(paddingArray, [](int64_t p) { return std::max<int64_t>(p, 0); }));
```

lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
```diff
@@ -1348,10 +1348,37 @@ class StableHLOToTTIRConvolutionOpConversionPattern
             : rewriter.getDenseBoolArrayAttr(
                   SmallVector<bool>(numSpatialDims, false));
 
-    ttir::utils::replaceOpWithNewDPSOp<mlir::tt::ttir::ConvolutionOp>(
-        rewriter, srcOp, outputType, adaptor.getLhs(), adaptor.getRhs(),
-        Value(), windowStridesAttr, paddingAttr, inputDilationAttr,
-        kernelDilationAttr, windowReversalAttr,
+    bool hasNegativePadding = llvm::any_of(paddingArray, [](int64_t p) { return p < 0; });
```
vmilosevic: :warning: **clang\-diagnostic\-error** :warning:
use of undeclared identifier `` paddingArray ``; did you mean `` paddingAttr ``?
```suggestion
    bool hasNegativePadding = llvm::any_of(paddingAttr, [](int64_t p) { return p < 0; });
```

vmilosevic (commented):

lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
```diff
@@ -1348,10 +1348,38 @@ class StableHLOToTTIRConvolutionOpConversionPattern
             : rewriter.getDenseBoolArrayAttr(
                   SmallVector<bool>(numSpatialDims, false));
 
-    ttir::utils::replaceOpWithNewDPSOp<mlir::tt::ttir::ConvolutionOp>(
-        rewriter, srcOp, outputType, adaptor.getLhs(), adaptor.getRhs(),
-        Value(), windowStridesAttr, paddingAttr, inputDilationAttr,
-        kernelDilationAttr, windowReversalAttr,
+    ArrayRef<int64_t> paddingArray = paddingAttr.asArrayRef();
```
sdjordjevicTT, line 1693: Can you just add a comment, so that the logic is easier to follow...

Something like this:
```
// Negative padding handling strategy:
// 1. Convolution ops don't support negative padding directly, so we split the
//    operation into two steps when negative padding is detected
// 2. First, run convolution with padding clamped to zero (resulting in a larger
//    intermediate output than the final desired shape)
// 3. Then, slice the intermediate output to achieve the effect of negative padding
//    (cropping edges based on the magnitude of negative padding values)
//
// Example: padding=[-1, -1] means "crop 1 pixel from each edge of the output"
//   - Run conv with padding=[0, 0] to get larger intermediate result
//   - Slice [1:-1, 1:-1] to remove the edges and get final output
```

sdjordjevicTT (approved):

Issue #5600, opened 2025-10-30
[ttnn.jit] Assert Triggered in GenericLowerDMA When Compiling Composite Ops with DRAM Operands


arminaleTT:
Stack trace for `cosh`

```
Unsupported memory space
UNREACHABLE executed at /localdev/arminale/tt-mlir/lib/Dialect/TTCore/IR/TTCoreOpsTypes.cpp:1375!
PLEASE submit a bug report to https://github.com/llvm/llvm-project/issues/ and include the crash backtrace.
Stack dump:
0.	Program arguments: ttmlir-opt --convert-ttnn-to-ttir --ttcore-register-device --ttir-to-ttmetal-pipeline=ttnn-mode=true input_dram.mlir -o armin_dram.mlir --mlir-print-ir-after-all
 #0 0x000055dffff695f0 llvm::sys::PrintStackTrace(llvm::raw_ostream&, int) (/localdev/arminale/tt-mlir/build/bin/ttmlir-opt+0x62c55f0)
 #1 0x000055dffff69a62 SignalHandler(int, siginfo_t*, void*) Signals.cpp:0:0
 #2 0x00007fda2cff5520 (/lib/x86_64-linux-gnu/libc.so.6+0x42520)
 #3 0x00007fda2d0499fc pthread_kill (/lib/x86_64-linux-gnu/libc.so.6+0x969fc)
 #4 0x00007fda2cff5476 gsignal (/lib/x86_64-linux-gnu/libc.so.6+0x42476)
 #5 0x00007fda2cfdb7f3 abort (/lib/x86_64-linux-gnu/libc.so.6+0x287f3)
 #6 0x000055dffff08fc8 llvm::install_out_of_memory_new_handler() (/localdev/arminale/tt-mlir/build/bin/ttmlir-opt+0x6264fc8)
 #7 0x000055dfffc79809 mlir::tt::ttcore::DeviceAttr::getMemoryMap(mlir::MemRefType, unsigned long, std::optional<mlir::AffineMap>, unsigned long) const /localdev/arminale/tt-mlir/lib/Dialect/TTCore/IR/TTCoreOpsTypes.cpp:1377:1
[...]

PR #5693, opened 2025-11-04
[ttnn.jit] Fix Stream Insertion and CB Op Insertion for Ops with DRAM Operands

arminaleTT:
### Ticket
https://github.com/tenstorrent/tt-mlir/issues/5600

This PR fixes two separate problems related to JIT ops and DRAM operands

### Problem 1
When a `d2m.generic` has DRAM inputs and outputs, a merged reader-writer kernels is created in `GenericHWThreadSelection` with all DMA ops in one block. `D2MToTTKernel` inserts CB release ops (push_back, pop_front) at the end of the parent block. This guarantees a deadlock as we await on the output CB before pushing the input.

```
^datamovement0:
    reserve_back
    // noc read
    wait_front
    // noc write
    push_back
    pop_front
```
### Fix
When merging DMA threads, wrap each thread in a `scf.execute_region` op with a `no_inline` attribute. This maintains a separate block for each thread, so "end of block" insertion does not create overlapping CB reservations.

```
^datamovement0:
    scf.execute_region no_inline {
        reserve_back
        // noc read
        push_back
    }
    scf.execute_region no_inline {
        wait_front
        // noc write
        pop_front
    }
```

`no_inline` prevents canonicalization from flattening the above structure prematurely. It is removed in `D2MToTTKernel` once CB ops are lowered.

### Problem 2
When JIT-ing a composite op with DRAM operands, an assertion is triggered in `GenericLowerDMA`. The root cause is in `GridSelection` when adding streams for TTNN dram tensors.  

Assuming all DRAM TTNN tensors, when analyzing the operands of the second generic, %1 is already streamed. However, the previous pass logic would implicitly drop the stream view and then insert a new `stream_layout` op. This broke the connection of the stream tensor to the original tensor, triggering the assert in a later pass.
```
%1 = d2m.generic ...
%2 = d2m.generic (ins: %1)
```

### Fix
When inserting streams for TTNN DRAM tensors, if the input to a `d2m.generic` is the output of a previous `d2m.generic`, directly set it as the input and do not insert a stream 

### Checklist

### DIFF
--- a/lib/Dialect/D2M/Transforms/GenericHWThreadSelection.cpp
+++ b/lib/Dialect/D2M/Transforms/GenericHWThreadSelection.cpp
@@ -171,25 +171,15 @@ public:
 
   void wrapBlockOpsInExecuteRegion(PatternRewriter &rewriter, Block &block,
                                    Location loc) const {
-    SmallVector<Operation *> opsToWrap;
-    for (Operation &blockOp : llvm::make_early_inc_range(block)) {
-      opsToWrap.push_back(&blockOp);
-    }
-
-    if (opsToWrap.empty()) {
-      return;
-    }
-
+    Block *executeRegionBlock = rewriter.splitBlock(&block, block.begin());
     rewriter.setInsertionPointToStart(&block);
-    auto executeRegionOp =
-        rewriter.create<scf::ExecuteRegionOp>(loc, TypeRange{});
-    // Prevent canonicalization from inlining the execute_region op.
-    executeRegionOp->setAttr("no_inline", rewriter.getUnitAttr());
+    auto executeRegionOp = rewriter.create<scf::ExecuteRegionOp>(
+        loc, TypeRange{}, /*no_inline=*/rewriter.getUnitAttr());
 
-    Block *executeRegionBlock = &executeRegionOp.getRegion().emplaceBlock();
-    for (Operation *opToMove : opsToWrap) {
-      opToMove->moveBefore(executeRegionBlock, executeRegionBlock->end());
-    }
+    // splitBlock creates the executeRegionBlock in the parent region of
+    // "block".
+    executeRegionBlock->getParent()->getBlocks().remove(executeRegionBlock);
+    executeRegionOp.getRegion().push_back(executeRegionBlock);
 
     rewriter.setInsertionPointToEnd(executeRegionBlock);
     rewriter.create<scf::YieldOp>(loc);


test/ttnn-jit/test_eltwise.py
```diff
@@ -377,6 +412,14 @@ def test_composite_ops(device, h, w, max_grid, dtype, op):
     run_op_test(device, h, w, max_grid, dtype, op, 1, buffer_type=ttnn.BufferType.L1)
 
 
+@pytest.mark.parametrize("h , w", DRAM_INTERLEAVED_SHAPES)
+@pytest.mark.parametrize("dtype", [torch.bfloat16, torch.float32])
+@pytest.mark.parametrize("op", [cosh, sinh])
+def test_composite_ops_dram(device, h, w, dtype, op):
```
vtangTT, line 418: I think I mentioned this on brandon's pr, but that one hasn't gone in yet. 

Could we move these composite tests to a separate file, maybe `test_composite.py` or `test_eltwise_composite.py`

arminaleTT: I would prefer to check this in here and then @brapananTT can move them and add more coverage in his pr

vtangTT: ok works for me

vtangTT (approved):

lib/Dialect/D2M/Transforms/GenericHWThreadSelection.cpp
```diff
@@ -131,14 +139,61 @@ class D2MGenericMergeDatamovementThreadsRewritePattern
...
 
     return success();
   }
+
+  void wrapBlockOpsInExecuteRegion(PatternRewriter &rewriter, Block &block,
+                                   Location loc) const {
+    SmallVector<Operation *> opsToWrap;
+    for (Operation &blockOp : llvm::make_early_inc_range(block)) {
+      opsToWrap.push_back(&blockOp);
+    }
```
nsmithtt: You might be able to do this routine more succinctly with `rewriter.splitBlock`, something like:
```c++
Block* executeRegion = rewriter.splitBlock(&block, block.begin());
rewriter.setInsertionPointToStart(&block);
auto executeRegionOp =
  rewriter.create<scf::ExecuteRegionOp>(loc, TypeRange{}, /*no_inline=*/rewriter.getUnitAttr());
executeRegionOp.getRegion().push_back(executeRegion);
rewriter.create<scf::YieldOp>(loc);
```

arminaleTT: TIL, thanks!

Issue #5696, opened 2025-11-04
[build] The install target is broken and TTMLIRConfig.cmake does not include all settings


brnorris03:
### Describe the bug

The install target fails on main (missing libraries).

```bash
-- Up-to-date: /Users/bnorris/tt/tt-mlir/install/tools/ttnn-standalone/compile_so.cpp
CMake Error at tools/tt-alchemist/csrc/cmake_install.cmake:41 (file):
  file INSTALL cannot find
  "/Users/bnorris/tt/tt-mlir/build/tools/tt-alchemist/csrc/libtt-alchemist-lib.dylib":
  No such file or directory.
Call Stack (most recent call first):
  tools/tt-alchemist/cmake_install.cmake:42 (include)
  tools/cmake_install.cmake:82 (include)
  cmake_install.cmake:67 (include)


FAILED: [code=1] CMakeFiles/install.util 
cd /Users/bnorris/tt/tt-mlir/build && /opt/ttmlir-toolchain/venv/lib/python3.11/site-packages/cmake/data/bin/cmake -P cmake_install.cmake
```

Moreover, targets that may be needed by downstream builds are not installed, even when built, most notably, `libTTMLIRCompilerStatic.a`.

The generated `TTMLIRConfig.cmake` does not include settings required for integrating in downstream projects through cmake (`TTMLIR_LIBRARY_DIRS`, `TTMLIR_TOOLS_DIR`, etc):
```cmake
find_package(MLIR REQUIRED CONFIG
             HINTS "")

set(TTMLIR_EXPORTED_TARGETS "TTMLIRTTKernelToEmitC;MLIRTTTransforms;MLIRTTCoreDialect;MLIRTTNNDialect;TTMLIRTTNNUtils;MLIRTTKernelDialect;MLIRTTMetalDialect;TTNNOpModelLib;MLIRTTNNInterfaces;coverage_config")
set(TTMLIR_CMAKE_DIR "/Users/bnorris/tt/tt-mlir/build/lib/cmake/ttmlir")
set(TTMLIR_BINARY_DIR "")
set(TTMLIR_INCLUDE_DIRS "/Users/bnorris/tt/tt-mlir/build/include;/Users/bnorris/tt/tt-mlir/include")
set(TTMLIR_LIBRARY_DIRS "")
set(TTMLIR_TOOLS_DIR "")

# Provide all our library targets to users.
if(NOT TARGET TTMLIRSupport)
[...]

PR #5698, opened 2025-11-04
[cmake] Fix a number of build issues, including the clean and install targets for TTMLIR

brnorris03:
### Ticket
Closes https://github.com/tenstorrent/tt-mlir/issues/5696

### Problem description
The install target fails (`cmake --install <build_dir>`). Also the generated `TTMLIRConfig.cmake` is incomplete.

### What's changed
1. CMake Configuration Fixes
   - `TTMLIRConfig.cmake.in`: Store install prefix directly from `CMAKE_INSTALL_PREFIX` at configure time (no runtime path computation needed)
   - `lib/CMakeLists.txt`: 
     - Added `INTERFACE_INCLUDE_DIRECTORIES` to `TTMLIRCompilerStatic` target to export include paths for consumers
     - Made `SHARDY_LIBS` a `PRIVATE` dependency instead of `PUBLIC` to avoid exposing Shardy symbols in the public interface
     - Conditionally export `TTMLIRCompilerStatic` via CMake only when `TTMLIR_ENABLE_STABLEHLO=OFF` (always installed, but CMake target export requires no external dependencies)
   - `cmake/modules/TTMLIRInstall.cmake`: New module with explicit whitelist of TTMLIR targets for export, avoiding accidental inclusion of third-party StableHLO, Shardy, etc. targets
   - `cmake/modules/CMakeLists.txt`: Centralized installation logic for all TTMLIR targets via `TTMLIRTargets` export

2. Target Definition Fixes
   - `lib/Dialect/TTNN/Interfaces/CMakeLists.txt`: Changed from `add_mlir_interface_library` to `add_mlir_library` for `MLIRTTNNInterfaces`
   - `lib/Scheduler/CMakeLists.txt`: Changed from `add_mlir_library` to `add_mlir_dialect_library` for `MLIRScheduler`
   - `lib/OpModel/TTNN/CMakeLists.txt`: Changed from `add_library` to `add_mlir_library` for `TTNNOpModelLib` and fixed `INTERFACE_INCLUDE_DIRECTORIES`

3. StableHLO Integration Handling
   - `cmake/modules/TTMLIRInstall.cmake`: Conditionally exclude pipeline targets (`MLIRTTIRPipelines`, `MLIRTTNNPipelines`) and StableHLO integration targets from export when `TTMLIR_ENABLE_STABLEHLO=ON` to avoid transitive dependencies on third-party libraries

4. Install Test
   - `test/install/`: New standalone test project that uses `find_package(TTMLIR)` and links against a tt-mlir library and an upstream MLIR library
   - `.github/scripts/bash/test-install.sh`: Script to configure, build, and run the install test
   - `.github/workflows/call-build-release.yml`: Added install test step after default installation

5. Component Management
   - `lib/CMakeLists.txt`: Install `TTMLIRCompilerStatic` with new `StaticLib` component
   - `.github/workflows/call-build-release.yml`: Added `StaticLib` component to CI install step and verification check for `libTTMLIRCompilerStatic.a`

6. Build Options
   - `CMakeLists.txt`: Added `TTMLIR_ENABLE_ALCHEMIST` option to conditionally build `tt-alchemist` tool, similar to other optional targets
   - `tools/CMakeLists.txt`: Wrapped `tt-alchemist` subdirectory in conditional

### Checklist

### DIFF
--- a/.github/build_scripts/test-install.sh
+++ b/.github/build_scripts/test-install.sh
@@ -5,7 +5,7 @@
 #
 # Script to install and test TTMLIR by building a standalone project
 # against the installation using a standalone test project in `test/install`.
-# 
+#
 # This script expects the following environment variables to be set:
 #   WORK_DIR    - Repository root directory
 #   BUILD_DIR   - Build artifacts directory (where TTMLIR was built)


.github/workflows/call-build-release.yml
```diff
@@ -133,12 +133,28 @@ jobs:
...
         source env/activate
+        rm -rf ${{  steps.strings.outputs.install-output-dir }}
+        cmake --install ${{ steps.strings.outputs.build-output-dir }}
+        test -f "${{ steps.strings.outputs.install-output-dir }}/lib/libTTMLIRCompilerStatic.a" || { echo "ERROR: libTTMLIRCompilerStatic.a not found in ${{ steps.strings.outputs.install-output-dir }}/lib/" && exit 1; }
+
+    - name: Test install
+      shell: bash
+      run: |
+        ${{ steps.strings.outputs.work-dir }}/.github/scripts/bash/test-install.sh \
+          "${{ steps.strings.outputs.install-output-dir }}"
```
nsumrakTT: Do we need static build in release?
If we don't, and you just want to test feature in CI, can we do this in debug/code coverage build (call-build-debug.yml)?

.github/workflows/call-build-release.yml
```diff
@@ -133,12 +133,28 @@ jobs:
...
+
+    - name: Install specific components
+      shell: bash
+      run: |
+        source env/activate
+        rm -rf ${{  steps.strings.outputs.install-output-dir }}
         cmake --install ${{ steps.strings.outputs.build-output-dir }} --component Test
         cmake --install ${{ steps.strings.outputs.build-output-dir }} --component SharedLib
+        cmake --install ${{ steps.strings.outputs.build-output-dir }} --component StaticLib
```
nsumrakTT: If component is not used, it should not be installed.
Install and build directories of release build are packed and uploaded and then downloaded and unpacked in each test job.
Thus, if you want just to test if build feature work, please do it in debug build.

.github/workflows/call-build-release.yml
brnorris03: Probably not required in release specifically, happy to move to debug as you suggest.
Update: there were some issues with having this in debug, so it's back in release -- see related comment below for more details.

.github/workflows/call-build-release.yml
brnorris03: Will do!

brnorris03: Actually, after doing this and thinking about it a bit more, I think the release workflow is more appropriate for the install test, so I moved it back there (but removed any extra static artifacts from the upload -- they are transient). Here are my reasons for having this in the release workflow instead of debug.
1. The downstream users of the installed project should be primarily using the release build, so it makes more sense to test that. 
2. The debug build is configured with coverage, which incurs extra dependencies that would not normally be in an installed version of the project and so shouldn't be added to the `test/install` stand-alone project that is used for testing (but would have to be added if using the debug build). An alternative is to add an entire extra build without coverage just for testing install, but that increases CI time and I think is not a good idea.

Does that sound reasonable?

nsumrakTT: A note from my perspective.
So we are using CI to check for regression. If testing of static lib/install is in question we can use debug build. It is the same makefile and build system and for testing it should be fine. It is part of the build process but no tests depends on it and its length is not critical.
The release is used for testing and these builds are in the critical path meaning that each PR length depends on the length of it. There are 2 release builds we use (speedy and tracy) and you added make/test steps to both. Let's make it only for tracy which is currently shorter using "component" feature (see docs "CI" section)?
I'm don't know about test/install standalone project and I didn't understand how CI will increase in time if used on debug build.

nsmithtt:
This is awesome, I think we should probably trigger tt-xla and tt-forge-fe CI against this PR just to make sure that the frontend projects aren't broken because of the new install flow.

@ctodTT, can you show @brnorris03 how to trigger those projects CI against a tt-mlir branch?

.github/workflows/call-build-release.yml
brnorris03: I agree testing in just one of the release builds is sufficient and will make that change. 

The stand-alone cmake project in test/install is a test of the install -- in other words, it's a tiny "downstream user" project, which exercises the install in exactly the same way as any normal user of the installed project would (in a very limited way obviously). If you try configuring and building it with the debug configuration (as it currently is defined), you will also require to do coverage in that project as well -- i.e., add more complexity to its cmake (which doesn't really make sense since the goal is to mimic normal use -- and requiring users to accomodate a tt-mlir coverage-enabled build is not something we'd do).

So to test the installed package in the debug build, one would have to add a separate tt-mlir Release configuration, full build of that, install, then finally use that install with the small test project. It has nothing to do with static libraries, this is testing a full installation and in fact is only linking with dynamic libraries for that particular test. The reason the post-install file check looks for the static library is that it's one of the targets that failed to install previously.

brnorris03: However, adding a `Release` configuration to the debug workflow doesn't seem that clean to me, so another alternative is to create a (small) new `call-install-release.yml` just for configuring and building in Release mode, installing, and testing the installed project. Would that be better in your opinion.

> Let's make it only for tracy which is currently shorter using "component" feature (see docs "CI" section)?

I did make this update (and it works fine, including having no effect on downstream projects and adding a total of 2 seconds to the release build and no extra artifacts), but will revise if another alternative seems better.

brnorris03:
> This is awesome, I think we should probably trigger tt-xla and tt-forge-fe CI against this PR just to make sure that the frontend projects aren't broken because of the new install flow.
> 
> @ctodTT, can you show @brnorris03 how to trigger those projects CI against a tt-mlir branch?

Successful downstream tests with https://github.com/tenstorrent/tt-mlir/pull/5698/commits/ef1ce92e6926617512ed61165e96a8e2bfc9de92

* tt-xla
https://github.com/tenstorrent/tt-xla/actions/runs/19111458603

* tt-forge
https://github.com/tenstorrent/tt-forge-fe/actions/runs/19111410687

nsmithtt (approved):
LGTM!

.github/workflows/call-build-release.yml
```diff
@@ -127,6 +127,24 @@ jobs:
         source env/activate
         cmake --build ${{ steps.strings.outputs.build-output-dir }}
 
+    - name: Install (default full install)
```
nsumrakTT: Could you please take a look at docs:
https://docs.tenstorrent.com/tt-mlir/ci.html#release-build-components

Instead of modifying workflow you can create component - simple shell script inside .github/build_scripts that would perform what is done in these 2 steps (and above script) and add it to .github/settings/build.json to tracy build.

brnorris03: Done. Sorry, I thought you were referring to the cmake (installable) components before, didn't realize you meant the json file.

nsumrakTT (approved):
Great, thanks!

.github/build_scripts/test-install.sh
```diff
@@ -0,0 +1,70 @@
+#!/bin/bash
```
vroubtsovTT, line 1: Is using `#!/usr/bin/env bash` here considered better practice?

brnorris03: I don't think so, not for bash in this case. I want to use specifically the one that comes with Ubuntu (or MacOS).

PR #5699, opened 2025-11-04
[D2M] Insert dummy generic if func returns view

vwellsTT:
### Ticket
NA

### Problem description
D2M views are only materialized by generics, but if a view is returned from a func, no generic consumes it.  We need a way to properly return results of views.

### What's changed
Tiny new pass to find any case where a func is returning an unmaterialized view and insert a generic op to materialize it (and return the result of that instead).

### Checklist

### DIFF
--- a/lib/Dialect/D2M/Transforms/MaterializeViewReturns.cpp
+++ b/lib/Dialect/D2M/Transforms/MaterializeViewReturns.cpp
@@ -97,33 +97,20 @@ public:
     // Process each function in the module to find unmaterialized view returns.
     module.walk([&](func::FuncOp funcOp) {
       funcOp.walk([&](func::ReturnOp returnOp) {
-        SmallVector<Value> newOperands;
-        bool modified = false;
+        builder.setInsertionPoint(returnOp);
 
-        // Inspect each return value to determine if it needs materialization.
-        for (Value operand : returnOp.getOperands()) {
-          Operation *definingOp = operand.getDefiningOp();
+        // Inspect each return operand to determine if it needs materialization.
+        for (OpOperand &opOperand : returnOp->getOpOperands()) {
+          Operation *definingOp = opOperand.get().getDefiningOp();
 
           if (isViewOp(definingOp)) {
             // Insert a generic op to materialize the view before returning.
             // This ensures the tensor transformation represented by the view
             // actually occurs, rather than just being a symbolic operation.
-            builder.setInsertionPoint(returnOp);
             Value materialized =
-                materializeView(builder, returnOp.getLoc(), operand);
-            newOperands.push_back(materialized);
-            modified = true;
-          } else {
-            // Non-view values can be returned directly without modification.
-            newOperands.push_back(operand);
+                materializeView(builder, returnOp.getLoc(), opOperand.get());
+            opOperand.set(materialized);
           }
-        }
-
-        // Replace the return op if any views were materialized.
-        if (modified) {
-          builder.setInsertionPoint(returnOp);
-          builder.create<func::ReturnOp>(returnOp.getLoc(), newOperands);
-          returnOp.erase();
         }
       });
     });
--- a/test/ttmlir/Dialect/D2M/materialize_view_returns.mlir
+++ b/test/ttmlir/Dialect/D2M/materialize_view_returns.mlir
@@ -107,3 +107,25 @@ func.func @mixed_returns(%arg0: tensor<1x1x8x24x!ttcore.tile<32x32, f32>, #layou
   // CHECK: return %[[MATERIALIZED]], %arg1
   return %view, %arg1 : tensor<8x8x1x3x!ttcore.tile<32x32, f32>, #layout8x8>, tensor<8x8x1x3x!ttcore.tile<32x32, f32>, #layout8x8>
 }
+
+// Test higher-rank tensors (3D uncollapsed) for permute TM use case
+// Grid has leading 1 (1x4x4 -> 1x8x2), demonstrating higher rank with legal grid
+// CHECK-LABEL: @higher_rank_view_return
+#layout_6d_1x4x4 = #ttcore.metal_layout<logical_shape = 64x96x192, dim_alignments = 32x32x32, collapsed_intervals = dense<[[0, 1], [1, 2], [2, 3]]> : tensor<3x2xi64>, undef, l1>
+#layout_6d_1x8x2 = #ttcore.metal_layout<logical_shape = 64x96x192, dim_alignments = 32x32x32, collapsed_intervals = dense<[[0, 1], [1, 2], [2, 3]]> : tensor<3x2xi64>, undef, l1>
+func.func @higher_rank_view_return(%arg0: tensor<1x4x4x2x3x6x!ttcore.tile<32x32, f32>, #layout_6d_1x4x4>) -> tensor<1x8x2x1x6x6x!ttcore.tile<32x32, f32>, #layout_6d_1x8x2> {
+  // CHECK: %[[VIEW:.*]] = d2m.view_layout %arg0
+  %view = d2m.view_layout %arg0 : tensor<1x4x4x2x3x6x!ttcore.tile<32x32, f32>, #layout_6d_1x4x4> -> tensor<1x8x2x1x6x6x!ttcore.tile<32x32, f32>, #layout_6d_1x8x2>
+
+  // CHECK: d2m.empty() : tensor<1x8x2x1x6x6x!ttcore.tile<32x32, f32>
+  // CHECK: %[[MATERIALIZED:.*]] = d2m.generic
+  // CHECK-SAME: grid = #ttcore.grid<1x8x2>
+  // CHECK-SAME: threads = [#d2m.thread<datamovement>]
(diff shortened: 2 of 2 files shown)


lib/Dialect/D2M/Transforms/MaterializeViewReturns.cpp
```diff
@@ -0,0 +1,139 @@
...
+#define GEN_PASS_DEF_D2MMATERIALIZEVIEWRETURNS
+#include "ttmlir/Dialect/D2M/Transforms/Passes.h.inc"
+
+namespace {
+
+// Check if an operation implements the ViewOpInterface.
+// View operations represent tensor transformations using affine mappings
+// without actual data movement, requiring materialization before use.
+bool isViewOp(Operation *op) {
+  return op && mlir::isa<d2m::ViewOpInterface>(op);
```
vmilosevic: :warning: **llvm\-prefer\-isa\-or\-dyn\-cast\-in\-conditionals** :warning:
isa\_and\_nonnull\<\> is preferred over an explicit test for null followed by calling isa\<\>
```suggestion
  return isa_and_nonnullr::isa<d2m::ViewOpInterface>(op);
```

lib/Dialect/D2M/Transforms/MaterializeViewReturns.cpp
```diff
@@ -0,0 +1,134 @@
+// SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
```
vwellsTT, line 1: worth noting that in some ways, this type of approach doesn't adhere to LLVMic standards that IR should always be **correct**, because we're relying on this pass to fix broken IR generated by LowerToLayout etc. if it emits a view which is returned directly.  Debatably, we ought to reverse this and always emit a dummy generic w/ every view, and then selectively remove dead ones or something--but that has a lot of overhead.  Would like to get a quick workaround like this landed for other PRs, but curious if others think we should approach this differently.

nsmithtt: FYI @tapspatel / @anuragsingh-tt / @dlokeTT who have also been hitting this issue.

test/ttmlir/Dialect/D2M/materialize_view_returns.mlir
```diff
@@ -0,0 +1,103 @@
...
+  %view = d2m.view_layout %arg0 : tensor<1x1x8x24x!ttcore.tile<32x32, f32>, #layout1x1> -> tensor<8x8x1x3x!ttcore.tile<32x32, f32>, #layout8x8>
+
+  // CHECK: d2m.empty() : tensor<8x8x1x3x!ttcore.tile<32x32, f32>
+  // CHECK: %[[MATERIALIZED:.*]] = d2m.generic
+  // CHECK: ins(%[[VIEW]]
+  // CHECK: d2m.reserve
+  // CHECK: d2m.yield
+  // CHECK: return %[[MATERIALIZED]], %arg1
+  return %view, %arg1 : tensor<8x8x1x3x!ttcore.tile<32x32, f32>, #layout8x8>, tensor<8x8x1x3x!ttcore.tile<32x32, f32>, #layout8x8>
+}
```
nsmithtt, line 109: Can we add a test that does higher rank than 2d?  This is an immediate requirement for the permute TM use case.

lib/Dialect/D2M/Transforms/MaterializeViewReturns.cpp
```diff
@@ -0,0 +1,134 @@
...
+    OpBuilder builder(&getContext());
+
+    // Process each function in the module to find unmaterialized view returns.
+    module.walk([&](func::FuncOp funcOp) {
+      funcOp.walk([&](func::ReturnOp returnOp) {
+        SmallVector<Value> newOperands;
+        bool modified = false;
+
+        // Inspect each return value to determine if it needs materialization.
+        for (Value operand : returnOp.getOperands()) {
```
nsmithtt: Instead of tracking `modified` and having to replace the whole op, you could iterate the `getOpOperands` which would let you just `if (isView) { ... opOperand.set(materialized) ... }`

lib/Dialect/D2M/Transforms/MaterializeViewReturns.cpp
```diff
@@ -0,0 +1,134 @@
...
+  // Allocate output storage for the materialized view result.
+  auto layout =
+      mlir::dyn_cast_or_null<ttcore::MetalLayoutAttr>(tensorType.getEncoding());
+  auto emptyOp = builder.create<d2m::EmptyOp>(
+      loc, tensorType.getShape(), tensorType.getElementType(), layout);
+
+  // Extract the grid from the tensor's layout to determine core distribution.
+  ttcore::GridAttr grid = getGridFromType(tensorType);
+  if (!grid) {
+    // Fall back to single-core execution when grid information is unavailable.
```
nsmithtt: Is this case ever true?

lib/Dialect/D2M/Transforms/MaterializeViewReturns.cpp
```diff
@@ -0,0 +1,134 @@
...
+  // causing the datamovement hardware to fetch data according to the view's
+  // affine map transformation.
+  auto genericOp = builder.create<GenericOp>(
+      loc, viewResult, emptyOp.getResult(),
+      [](OpBuilder &builder, Location loc, ValueRange blockArgs) {
+        Value output =
+            builder.create<d2m::ReserveOp>(loc, blockArgs[1]).getResult();
+        builder.create<d2m::YieldOp>(loc, output);
+      },
+      ThreadType::Datamovement, grid, SmallVector<int64_t>{1, 1});
```
nsmithtt, line 82: Does this work on silicon? I'd expect the thread type to be Compute here.  We might need to insert an explicit copy...

Or if we want to keep the Datamovement thread type we need to manually insert the dma in (affine form) between the reserve and yield.

lib/Dialect/D2M/Transforms/MaterializeViewReturns.cpp
anuragsingh-tt: > always emit a dummy generic w/ every view, and then selectively remove dead ones 

+1 for this being the long term solution. The dummy generic will be preserved if it is returned by the func.

lib/Dialect/D2M/Transforms/MaterializeViewReturns.cpp
vwellsTT: Shouldn't be, I'll assert instead

lib/Dialect/D2M/Transforms/MaterializeViewReturns.cpp
vwellsTT: Oh, I guess I haven't actually run it, only compiled it; I guess maybe it'd be good to add some trivial builder tests to this PR and expand 'em in my follow-on.

lib/Dialect/D2M/Transforms/MaterializeViewReturns.cpp
vwellsTT: Yeah, I lean towards that--the downside, to me, is that we insert view ops in a couple different places, so we need to like add some utility instead that inserts ViewLayoutOp AND dummyGeneric, and then somehow make sure that's always how people insert ViewLayoutOps, I guess? It'd be nice if we could somehow enforce that, idk

lib/Dialect/D2M/Transforms/MaterializeViewReturns.cpp
vwellsTT: added explicit dma, it seems to work fine for me locally

lib/Dialect/D2M/Transforms/MaterializeViewReturns.cpp
vwellsTT: nifty

nsmithtt (approved):
Awesome!

nsmithtt:
> This seems to do what is asked for, but I have one question (shows how long I've been in my own rabbit hole, I am forgetting everything else): why can't to_layout be used for this instead of emitting a custom generic op?

This is handling the lowering of the to_layout op.

vroubtsovTT:
> > This seems to do what is asked for, but I have one question (shows how long I've been in my own rabbit hole, I am forgetting everything else): why can't to_layout be used for this instead of emitting a custom generic op?
> 
> This is handling the lowering of the to_layout op.

I had thought that to_layout would end up materializing the data movement that it itself implies, so still confused.

nsmithtt:
> > > This seems to do what is asked for, but I have one question (shows how long I've been in my own rabbit hole, I am forgetting everything else): why can't to_layout be used for this instead of emitting a custom generic op?
> > 
> > 
> > This is handling the lowering of the to_layout op.
> 
> I had thought that to_layout would end up materializing the data movement that it itself implies, so still confused.

to_layout can be lowered into explicit datamovement or it can be lowered as just a view in certain situations.  Views are largely preferable because they can be stacked with other views and evaluated all together.  There is however this edge case where a view is returned from the main func, hence the need for a generic op to materialize it.

Issue #5278, opened 2025-10-08
[ttnn.jit] Support Rank 3 and 4 Tensors in JIT Frontend


arminaleTT:
Currently we only support rank 2 tensors because we do not do dim collapsing for the TTNNLayoutAttr, or call a getter that would perform it automatically

PR #5700, opened 2025-11-04
[ttnn.jit] Add Support for Rank 3 and 4 Tensors in JIT Frontend

brapananTT:
### Ticket
#5278 

### Problem description
Currently, we only support rank 2 tensors in ttnn jit because we do not support dim collapsing for the TTNNLayoutAttr, or call a getter that would perform it automatically.

### What's changed
Added new function `_get_collapsed_linear_affine_map()` in `ttnn-jit/_src/utils.py` which implements dimension collapsing for higher rank shapes. This is wired into `ttir_ast.py` and `graph_trace_compiler.py` `_create_tensor_layout()`in order to support rank 3 and 4 tensors. Tests for rank 3 and 4 tensors are added in `test_eltwise.py` and `test_layouts.py.`

### Checklist

### DIFF
--- a/test/ttnn-jit/test_layouts.py
+++ b/test/ttnn-jit/test_layouts.py
@@ -17,11 +17,11 @@ from utils import (
     run_op_test,
 )
 
-# Generates all shapes with 1 to 4 tiles per core in each dimension, with every grid from single core to 8x8.
-# TTNN grids are (Width, Height), while tensor shapes are (Height, Width).
+
 BLOCK_SHARDED_SHAPE_GRIDS = []
 
-# 2D shapes
+# Generates all rank 2 shapes with 1 to 4 tiles per core in each dimension, with every grid from single core to 8x8.
+# TTNN grids are (Width, Height), while tensor shapes are (Height, Width).
 BLOCK_SHARDED_SHAPE_GRIDS.extend(
     [
         ((h * 32 * (grid_h + 1), w * 32 * (grid_w + 1)), (grid_w, grid_h))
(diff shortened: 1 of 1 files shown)


test/ttnn-jit/test_eltwise.py
```diff
@@ -17,32 +17,47 @@
...
+    ((1024, 2048), (7, 7)),
     # Ensure non-square grid dims are interpreted correctly.
-    (64, 128, (3, 0)),
-    (64, 128, (0, 1)),
-    (96, 128, (3, 2)),
+    ((64, 128), (3, 0)),
+    ((64, 128), (0, 1)),
+    ((96, 128), (3, 2)),
+    # Include rank 3 and 4 tensors
+    ((1, 32, 32), (0, 0)),
```
arminaleTT: These are too many tests. Let's reduce them to
1. 3d single core
2. 3d all cores
3. 3d non-square grid
4. 4d single core
5. 4d all cores
6. 4d non-square grid

test/ttnn-jit/test_eltwise.py
```diff
@@ -17,32 +17,47 @@
...
+    ((256, 256)),
+    ((1024, 2048)),
+    ((2048, 2048)),
+    ((1024, 32)),
+    ((32, 1024)),
+    ((64, 1024)),
+    ((1024, 64)),
+    # Include rank 3 and 4 tensors
+    ((1, 32, 32)),
+    ((1, 64, 64)),
```
arminaleTT: just 1 for each of 3d and 4d is enough, but make sure the leading dims aren't all 1

test/ttnn-jit/test_layouts.py
```diff
@@ -18,18 +19,48 @@
...
+BLOCK_SHARDED_SHAPE_GRIDS.extend(
+    [
+        ((h * 32 * (grid_h + 1), w * 32 * (grid_w + 1)), (grid_w, grid_h))
+        for h, w, grid_h, grid_w in itertools.product(
+            range(1, 5), range(1, 5), range(8), range(8)
+        )
+    ]
+)
+
+# 3D shapes
```
arminaleTT: Please comment on what list of shapes this covers

tools/ttnn-jit/_src/utils.py
```diff
@@ -78,3 +79,75 @@ def _get_num_pos_args(func: Callable):
         ]
     )
     return num_pos_args
+
+
+def _get_collapsed_linear_affine_map(
```
arminaleTT, line 84: Include that this function is for creating `TTNNLayoutAttr` and its default behaviour must match TTNN's dim collapsing behaviour. Refer to the C++ function that this is based on

PR #5709, opened 2025-11-05
EmitPy GlobalOps

amilovanovicTT:
### Problem description
Added new ops to the EmitPy dialect and implemented their translation to Python. These will be used to support the consteval logic.
### What's changed
- GlobalOp : defines a global variable
- GlobalStatementOp : represents a Python global statement
- GetGlobalOp : obtains access to a global variable
- AssignGlobalOp : assigns a value to a global variable

Added lit tests to verify positive and negative scenarios.


include/ttmlir/Dialect/EmitPy/IR/EmitPyOps.td
```diff
@@ -197,4 +199,79 @@ def EmitPy_ConstantOp : EmitPy_Op<"constant", [ConstantLike]> {
...
+  let arguments = (ins FlatSymbolRefAttr:$name);
+  let results = (outs AnyType:$result);
+  let assemblyFormat = "$name `:` type($result) attr-dict";
+}
+
+def EmitPy_AssignGlobalOp : EmitPy_Op<"assign_global",
+    [DeclareOpInterfaceMethods<SymbolUserOpInterface>]> {
+  let summary = "Assign a value to a global variable";
+  let description = [{
+    The `emitpy.assign_global` assignes the value to an existing named global variable.
```
sdjordjevicTT, line 273: Can you please add an example here as well?

include/ttmlir/Dialect/EmitPy/IR/EmitPyOps.td
```diff
@@ -197,4 +199,79 @@ def EmitPy_ConstantOp : EmitPy_Op<"constant", [ConstantLike]> {
...
+    %0 = emitpy.global_statement @x : !emitpy.opaque<"[ttnn.Tensor]">
+    ```
+  }];
+
+  let arguments = (ins FlatSymbolRefAttr:$name);
+  let results = (outs AnyType:$result);
+  let assemblyFormat = "$name `:` type($result) attr-dict";
+}
+
+def EmitPy_GetGlobalOp : EmitPy_Op<"get_global",
```
sdjordjevicTT, line 249: I am not sure I am getting it. What is the difference between EmitPy_GlobalStatementOp and EmitPy_GetGlobalOp?

lib/Dialect/EmitPy/IR/EmitPyOps.cpp
```diff
@@ -378,3 +391,66 @@ LogicalResult ConstantOp::verify() {
...
+  }
+
+  if (parser.parseAttribute(initialValue)) {
+    return failure();
+  }
+
+  return success();
+}
+
+LogicalResult GlobalOp::verify() {
```
sdjordjevicTT, line 470: Can we add lit tests to verify that both positive and negative scenarios are hit?

lib/Target/Python/TranslateToPython.cpp
```diff
@@ -399,14 +402,46 @@ static LogicalResult printOperation(PythonEmitter &emitter,
   return emitter.emitAttribute(constantOp->getLoc(), value);
 }
 
+static LogicalResult printOperation(PythonEmitter &emitter, GlobalOp globalOp) {
```
sdjordjevicTT, line 405: It would be great if we can also write few lit tests for translation as well.

include/ttmlir/Dialect/EmitPy/IR/EmitPyOps.td
amilovanovicTT: EmitPy_GetGlobalOp is used to get the value of a certain global variable so that we can use it within the function scope.
EmitPy_GlobalStatementOp is used to produce this kind of code in Python:
`global g_cached_result_forward_const_eval_0` 
at the beginning of the function body.

include/ttmlir/Dialect/EmitPy/IR/EmitPyOps.td
amilovanovicTT: Done

lib/Dialect/EmitPy/IR/EmitPyOps.cpp
amilovanovicTT: Sure

include/ttmlir/Dialect/EmitPy/IR/EmitPyOps.td
sdjordjevicTT: Got it, I wasn't very familiar with the globals in Python :D

include/ttmlir/Dialect/EmitPy/IR/EmitPyOps.td
svuckovicTT: @amilovanovicTT I don't see an example, did you forget to push the changes?

lib/Dialect/EmitPy/IR/EmitPyOps.cpp
svuckovicTT, line 470: Verify valid Python variable name here as well.

include/ttmlir/Dialect/EmitPy/IR/EmitPyOps.td
```diff
@@ -197,4 +199,79 @@ def EmitPy_ConstantOp : EmitPy_Op<"constant", [ConstantLike]> {
...
+  let arguments = (ins SymbolNameAttr:$sym_name,
+                       EmitPy_OpaqueOrTypedAttr:$initial_value);
+
+  let hasVerifier = 1;
+  let assemblyFormat = [{
+       $sym_name custom<EmitPyGlobalOpInitialValue>($initial_value) attr-dict
+  }];
+}
+
+def EmitPy_GlobalStatementOp : EmitPy_Op<"global_statement",
```
svuckovicTT, line 224: Add verifier to check if `FlatSymbolRefAttr:$name` is a valid Python variable name.

sdjordjevicTT, line 224: After some thinking, I think we don't need this op. During function printing, we can easily find all the EmitPy_AssignGlobalOp, right? And just print their symbol names at the begining of the function printer...

Something like this:
```cpp
//LogicalResult printFunctionBody(...
// ...
  // Collect all global variables that are modified in this function
  llvm::SmallVector<StringRef> modifiedGlobals;
  blocks.front().walk([&](SetGlobalOp setGlobalOp) {
    modifiedGlobals.push_back(setGlobalOp.getSymName());
  });

  // Emit global declarations if needed
  if (!modifiedGlobals.empty()) {
    os << "global ";
    llvm::interleave(
        modifiedGlobals, [&](StringRef name) { os << name; },
        [&]() { os << ", "; });
    os << "\n";
  }
```

svuckovicTT: Tbh that feels a bit hacky. `global x` has semantic meaning in code so should be represented in IR IMO. It also matters where it sits, e.g. these two programs are different:

```py
foo = 5
def bar():
    global foo
    foo += 1
    print(foo)
```

```py
foo = 5
def bar():
    foo += 1
    global foo
    print(foo)
```

Additionally, we might end up adding `nonlocal` keyword as well at some point, so `AssignGlobalOp` would end up being used for both `global` and `nonlocal` (after a proper renaming of the op, of course).

svuckovicTT: I've given these ops some thought today, and I think having 4 of these ops is reasonable when considering the fact that `FuncOp` is `IsolatedFromAbove`. Can expand more if needed.

sdjordjevicTT: Yeah, I agree, let's add some tests, it would be much easier to comprehend the ops usage alongside them.

include/ttmlir/Dialect/EmitPy/IR/EmitPyOps.td
amilovanovicTT: Yes, sorry, should be visible now

lib/Dialect/EmitPy/IR/EmitPyOps.cpp
amilovanovicTT: Done

include/ttmlir/Dialect/EmitPy/IR/EmitPyOps.td
amilovanovicTT: Done

lib/Dialect/EmitPy/IR/EmitPyOps.cpp
amilovanovicTT: I added the negative one here, and the positive one can be found on ```amilovanovic/emitpyConstEval``` because it validates generated global ops for consteval path.

lib/Dialect/EmitPy/IR/EmitPyOps.cpp
```diff
@@ -99,6 +102,65 @@ static LogicalResult verifyInitializationAttribute(Operation *op,
...
+      symbolTable.lookupNearestSymbolFrom<GlobalOp>(op, op.getNameAttr());
+  if (!global) {
+    return op->emitOpError("'")
+           << op.getName() << "' does not reference a valid emitpy.global";
+  }
+
+  return success();
+}
+
+static bool isValidPythonIdentifier(StringRef name) {
```
sdjordjevicTT: Can we return LogicalResult here, and instead of returning false, for each case return op.emitOpError with an appropriate error message?

Also, please add one negative test case for each emitted error to verify all paths.

lib/Dialect/EmitPy/IR/EmitPyOps.cpp
sdjordjevicTT: I agree that in this PR, we should focus on testing the dialect independently. Therefore, can we add a few examples for each added operation that are not related to the consteval path? For this purpose, I was considering a test that generates a valid IR without any passes or integration. For example, something like this:

```mlir
// RUN: ttmlir-opt -o %t %s
// RUN: FileCheck %s --input-file=%t

// Test to check valid global operation creation
module {
  // CHECK: "emitpy.global"
  "emitpy.global"() <{sym_name = "valid_name", initial_value = "5"}> : () -> ()
}

// Test to check valid global operation retrieval
...
}
```

This should help ensure that we thoroughly test the dialect functionality.

lib/Target/Python/TranslateToPython.cpp
sdjordjevicTT: Something similar to the positive examples I wrote above, I think that the test should go from emitpy dialect and verify with a file check the generated Python code:
```mlir
// RUN: ttmlir-opt -o %t %s
// RUN: ttmlir-translate --mlir-to-python -o %t2 %t
// RUN: FileCheck %s --input-file=%t2
...
```

lib/Dialect/EmitPy/IR/EmitPyOps.cpp
amilovanovicTT: Done

sdjordjevicTT (approved):
This is great! Thanks for adding tests!

test/ttmlir/Dialect/EmitPy/global_negative.mlir
```diff
@@ -0,0 +1,94 @@
...
+module {
+  // CHECK: error: 'emitpy.global' op requires attribute 'initial_value'
+  "emitpy.global"() <{sym_name = "missing_initial_value"}> : () -> ()
+}
+
+// -----
+
+module {
+  // CHECK: error: 'emitpy.global' op variable name must not be a keyword
+  emitpy.global @global = #emitpy.opaque<"None">
```
svuckovicTT: I see different assembly formats here, is it on purpose? If not, I think we should go for the explicitly defined one, without the quotes.

test/ttmlir/Dialect/EmitPy/global_negative.mlir
```diff
@@ -0,0 +1,94 @@
...
+}
+
+// -----
+
+module {
+  // CHECK: error: @ identifier expected to start with letter or '_'
+  emitpy.global @123invalid = #emitpy.opaque<"None">
```
svuckovicTT, line 27: This test has 2 errors, let's always test one at a time - fix for all the negative tests.
```suggestion
  emitpy.global @123invalid = #emitpy.opaque<"initial_value">
```

test/ttmlir/Dialect/EmitPy/global/negative_global.mlir
```diff
@@ -0,0 +1,94 @@
...
+}
+
+// -----
+
+module {
+  emitpy.global @typed_global = 0
+
+  func.func @assign_global_type_mismatch(%arg0: !emitpy.opaque<"[ttnn.Tensor]">) -> !emitpy.opaque<"[ttnn.Tensor]"> {
+    // CHECK: error: 'emitpy.assign_global' op result type ('!emitpy.opaque<"[ttnn.Tensor]">') does not match global's type ('i64')
+    %0 = "emitpy.assign_global"(%arg0) <{name = @typed_global}> : (!emitpy.opaque<"[ttnn.Tensor]">) -> !emitpy.opaque<"[ttnn.Tensor]">
```
svuckovicTT: Would be nice to have custom assembly format for `AssignGlobalOp` as well :)

test/ttmlir/Dialect/EmitPy/global_negative.mlir
```diff
@@ -0,0 +1,94 @@
...
+
+// -----
+
+module {
+  func.func @get_global_nonexistent() -> !emitpy.opaque<"[ttnn.Tensor]"> {
+    // CHECK: error: 'emitpy.get_global' op 'nonexistent_global' does not reference a valid emitpy.global
+    %0 = emitpy.get_global @nonexistent_global : !emitpy.opaque<"[ttnn.Tensor]">
+    return %0 : !emitpy.opaque<"[ttnn.Tensor]">
+  }
+}
```
svuckovicTT, line 81: Add another test like this, but create a global that is named differently.

test/ttmlir/Dialect/EmitPy/global_negative.mlir
```diff
@@ -0,0 +1,94 @@
+// RUN: not ttmlir-opt --split-input-file %s 2>&1 | FileCheck %s
```
svuckovicTT, line 1: Move these global tests from `Dialect/EmitPy/global` to `Dialect/EmitPy` and rename to `global.mlir` and `global_negative.mlir`.

test/ttmlir/Dialect/EmitPy/global/positive_global.mlir
```diff
@@ -0,0 +1,144 @@
...
+
+// -----
+
+module {
+  // CHECK: emitpy.global @value = #emitpy.opaque<"None">
+  emitpy.global @value = #emitpy.opaque<"None">
+
+  func.func @assign_value(%arg0: !emitpy.opaque<"[ttnn.Tensor]">) -> !emitpy.opaque<"[ttnn.Tensor]"> {
+    // CHECK: %{{.*}} = "emitpy.assign_global"(%{{.*}}) <{name = @value}> : (!emitpy.opaque<"[ttnn.Tensor]">) -> !emitpy.opaque<"[ttnn.Tensor]">
+    %0 = "emitpy.assign_global"(%arg0) <{name = @value}> : (!emitpy.opaque<"[ttnn.Tensor]">) -> !emitpy.opaque<"[ttnn.Tensor]">
```
svuckovicTT: I guess this is valid IR since we're manually writing the test, but I still think it'd be nice to first capture the `@value` with `global_statement` op.

test/ttmlir/Translate/EmitPy/global_to_python.mlir
```diff
@@ -0,0 +1,64 @@
...
+
+// -----
+
+module {
+  // CHECK: global_var_2 = 100
+  emitpy.global @global_var_2 = 100 : i64
+
+  func.func @set_global_2() -> () {
+    // CHECK: def set_global_2():
+    // CHECK global global_var_2
```
svuckovicTT:
```suggestion
    // CHECK: global global_var_2
```

test/ttmlir/Translate/EmitPy/global_to_python.mlir
svuckovicTT, line 1: I'd rename this as well, just `global.mlir` works.

test/ttmlir/Dialect/EmitPy/global_negative.mlir
amilovanovicTT: Changed this to ```emitpy.global @global_var_missing_initial_value =```

test/ttmlir/Dialect/EmitPy/global/negative_global.mlir
amilovanovicTT: Agreed, added custom assembly format

Now ```AssignGlobalOp``` looks like : 
```emitpy.assign_global @value = %arg0 : !emitpy.opaque<"[ttnn.Tensor]">```
and doesn't have a return value.

PR #5710, opened 2025-11-05
[CI] Allow only one docker build at a time

nsumrakTT:
### Ticket
slack

### Problem description
When docker is built on main, during build time all PRs and onPush starts building the same image

### What's changed
Allow only one docker build at a time

### Checklist


vvukomanTT:
I didn't refresh to see the newest commit when approving, will we move the tagging with "latest" somewhere else?

nsumrakTT:
@vvukomanTT Actually it is already moved to set-latest-tag job

Issue #5654, opened 2025-11-03
Inverse reshapes around slice ops


mvasiljevicTT:
In yolo models there is a pattern:
```
                input <1, 1, NHW, C>
                        |
        reshape [1, 1, NHW, C] -> [N, H, W, C]
           /                            \
slice [:, :, :, :C/2]               slice [:, :, :, C/2:]
          |                              |
reshape -> [1, 1, NHW, C/2]         reshape -> [1, 1, NHW, C/2]
          |                              |
  out <1, 1, NHW, C/2>              out <1, 1, NHW, C/2>
```
Since reshapes are not affecting slice ranges, it can be rewritten as:

```
             input <1, 1, NHW, C>
           /                      \
slice [:, :, :, :C/2]         slice [:, :, :, C/2:]
          |                         |
  out <1, 1, NHW, C/2>        out <1, 1, NHW, C/2>
```
Inverse reshapes around slice can be erased if their lowest changing dim is higher than highest slicing dim

mvasiljevicTT:
Same as [this issue](https://github.com/tenstorrent/tt-mlir/issues/4731)

Issue #4731, opened 2025-08-29
Commute patterns for slice and concat in yolo


odjuricicTT:

LPanosTT:
@odjuricicTT Could we add a TTIR graph snippet of some patterns that EIO should be removing the TMs from, but do not?

odjuricicTT:
We see this pattern in almost all yolo models. @rpavlovicTT Could you add an IR example?

rpavlovicTT:
[yolov8_ttir.mlir.txt](https://github.com/user-attachments/files/23160545/yolov8_ttir.mlir.txt)

rpavlovicTT:
Result is that `slice_static` op has many `reshape` ops as users which are not erased.

<img width="1405" height="788" alt="Image" src="https://github.com/user-attachments/assets/ec5ebea3-b5b5-44be-bee7-7f08c8add5ab" />

LPanosTT:
@mvasiljevicTT Does this PR address this issue? https://github.com/tenstorrent/tt-mlir/pull/5713

mvasiljevicTT:
> [@mvasiljevicTT](https://github.com/mvasiljevicTT) Does this PR address this issue? [#5713](https://github.com/tenstorrent/tt-mlir/pull/5713)

Yes, I didn't see we already had an issue for that so I made a new one 😅

PR #5713, opened 2025-11-05
[EIO] Add Commute Patterns for Slice with Reshape

mvasiljevicTT:
### Ticket
[Link to Github Issue](https://github.com/tenstorrent/tt-mlir/issues/5654)
[Link to Github Issue](https://github.com/tenstorrent/tt-mlir/issues/4731)

### Problem description
In yolov9 model there is a pattern:
```
                input <1, 1, NHW, C>
                        |
        reshape [1, 1, NHW, C] -> [N, H, W, C]
           /                            \
slice [:, :, :, :C/2]               slice [:, :, :, C/2:]
          |                              |
reshape -> [1, 1, NHW, C/2]         reshape -> [1, 1, NHW, C/2]
          |                              |
  out <1, 1, NHW, C/2>              out <1, 1, NHW, C/2>
```
Since reshapes are not affecting slice ranges, it should be rewritten as:

```
             input <1, 1, NHW, C>
           /                      \
slice [:, :, :, :C/2]         slice [:, :, :, C/2:]
          |                         |
  out <1, 1, NHW, C/2>        out <1, 1, NHW, C/2>
```

### What's changed
Reshape can commute with slice if it is not affecting set of elements slice contains with and without that reshape. 

The is the case when slicing is done only on dims lower than the lowest dim changed by reshape, which we define in code as **if the leftmost slicing dim is right to the rightmost reshape dim**. For example, these cannot commute:
```
reshape [a,  b, c,  d]                  
     -> [a', b, c', d]                slice [:, :x, :, y:] 
         _______/                                 \_______      
  rightmost reshape dim                    leftmost slicing dim
```
When commuting we need to update slice args - begins, ends and steps:
- _left to the leftmost slicing dim_ they all have default values : 
  - `begin = 0`
  - `end = dimSize`
  - `step = 1`
so we only need to update `end` to new `dimSize`
- other dims are also _right to the rightmost reshape dim_ so they are not affected by reshape and keep original values.

Because reshape can change number of dimensions, we use **right-to-left** indexes as invariants, which is why they are called `rigtmostReshapeDimRTL`  and `leftmostSlicingDimRTL` in code.

This can be generalized by defining possible slicing dims as ones where cumulative products of dimensions match before and after reshape, for example `reshape [2,4,5,2,4] -> [8,5,2,2,2]` can commute with slice on dim 2, making slice on dim 1, which is not covered by this PR as we don't have need for this yet.

### Checklist

### DIFF
--- a/lib/Dialect/TTIR/Transforms/EraseInverseOps/SliceCommutePatterns.cpp
+++ b/lib/Dialect/TTIR/Transforms/EraseInverseOps/SliceCommutePatterns.cpp
@@ -65,11 +65,18 @@ public:
         rewriter.getArrayAttr(newSliceEnds),
         rewriter.getArrayAttr(newSliceSteps));
 
+    // All users must be identical TMs.
+    // We must not reference `permuteUser` during/after replacements, as it will
+    // be erased on its turn.
     SmallVector<Operation *> users(op->getUsers());
+    assert(llvm::all_of(users,
+                        [&](Operation *user) {
+                          return checkIdenticalTms(permuteUser, user);
+                        }) &&
+           "isCommuteUpwardsViable/Favorable should have ensured all users "
+           "are identical TMs");
+
     for (auto *user : users) {
-      assert(checkIdenticalTms(permuteUser, user) &&
-             "isCommuteUpwardsViable/Favorable should have ensured all users "
-             "are identical TMs");
       rewriter.replaceOp(user, newSlice);
     }
   }
@@ -175,11 +182,18 @@ public:
     SliceStaticOp newSlice =
         createReshapedSliceStaticOp(op, newReshape.getResult(), rewriter);
 
+    // All users must be identical TMs.
+    // We must not reference `reshapeUser` during/after replacements, as it will
+    // be erased on its turn.
     SmallVector<Operation *> users(op->getUsers());
+    assert(llvm::all_of(users,
+                        [&](Operation *user) {
+                          return checkIdenticalTms(reshapeUser, user);
+                        }) &&
+           "isCommuteUpwardsViable/Favorable should have ensured all users "
+           "are identical TMs");
+
     for (auto *user : users) {
-      assert(checkIdenticalTms(reshapeUser, user) &&
-             "isCommuteUpwardsViable/Favorable should have ensured all users "
-             "are identical TMs");
       rewriter.replaceOp(user, newSlice);
     }
   }
--- a/lib/Dialect/TTIR/Transforms/EraseInverseOps/BroadcastCommutePatterns.cpp
+++ b/lib/Dialect/TTIR/Transforms/EraseInverseOps/BroadcastCommutePatterns.cpp
@@ -232,12 +232,16 @@ public:
         rewriter, op->getLoc(), tmResultType, newReshape,
         newBroadcastDimensions);
 
+    // All users must be identical TMs.
+    // We must not reference `reshapeUser` during/after replacements, as it will
+    // be erased on its turn.
     SmallVector<Operation *> users(op->getUsers());
-    for (auto *user : users) {
-      assert(checkIdenticalTms(reshapeUser, user) &&
-             "isCommuteUpwardsViable/Favorable should have ensured all users "
-             "are identical TMs");
-    }
+    assert(llvm::all_of(users,
+                        [&](Operation *user) {
+                          return checkIdenticalTms(reshapeUser, user);
+                        }) &&
+           "isCommuteUpwardsViable/Favorable should have ensured all users "
+           "are identical TMs");
 
     for (auto *user : users) {
       rewriter.replaceOp(user, newBroadcast);
@@ -338,12 +342,16 @@ public:
         rewriter, op->getLoc(), tmResultType, newPermute,
         newBroadcastDimensions);
 
+    // All users must be identical TMs.
+    // We must not reference `permuteUser` during/after replacements, as it will
+    // be erased on its turn.
     SmallVector<Operation *> users(op->getUsers());
-    for (auto *user : users) {
-      assert(checkIdenticalTms(permuteUser, user) &&
-             "isCommuteUpwardsViable/Favorable should have ensured all users "
-             "are identical TMs");
-    }
+    assert(llvm::all_of(users,
+                        [&](Operation *user) {
+                          return checkIdenticalTms(permuteUser, user);
+                        }) &&
+           "isCommuteUpwardsViable/Favorable should have ensured all users "
+           "are identical TMs");
 
     for (auto *user : users) {
       rewriter.replaceOp(user, newBroadcast);
--- a/lib/Dialect/TTIR/Transforms/EraseInverseOps/ConcatCommutePatterns.cpp
+++ b/lib/Dialect/TTIR/Transforms/EraseInverseOps/ConcatCommutePatterns.cpp
@@ -76,11 +76,18 @@ public:
     ConcatOp newConcat = utils::createDPSOp<ConcatOp>(
         rewriter, op->getLoc(), newConcatType, newConcatOperands, newConcatDim);
 
+    // All users must be identical TMs.
+    // We must not reference `permuteUser` during/after replacements, as it will
+    // be erased on its turn.
     SmallVector<Operation *> users(op->getUsers());
+    assert(llvm::all_of(users,
+                        [&](Operation *user) {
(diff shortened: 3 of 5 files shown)


lib/Dialect/TTIR/Transforms/EraseInverseOps/SliceCommutePatterns.cpp
```diff
@@ -144,13 +145,290 @@ class TTIRCommutePermuteThroughSlice
...
+        rewriter.getI32ArrayAttr(targetShapeInt32));
+  }
+
+  bool isCommuteUpwardsViable(SliceStaticOp op,
+                              ReshapeOp reshapeUser) const override {
+    // Reshape can commute if its rightmost reshape dim
+    // is to the left of leftmost slicing dim
+    int64_t rightmostReshapeDimRTL = getRightmostReshapeDimRTL(reshapeUser);
+    int64_t leftmostSlicedDimRTL = getLeftmostSlicedDimRTL(op);
+    if (rightmostReshapeDimRTL == -1 || leftmostSlicedDimRTL == -1) {
```
mtopalovicTT, line 389: If we are doing unsqeeze or squeeze this will fail right, but it shouldn't right?

lib/Dialect/TTIR/Transforms/EraseInverseOps/SliceCommutePatterns.cpp
```diff
@@ -144,13 +145,290 @@ class TTIRCommutePermuteThroughSlice
...
+  }
+
+  bool isCommuteDownwardsFavorable(SliceStaticOp op,
+                                   ReshapeOp reshapeOperand) const override {
+    // Commuting downwards is favorable if all other operands satisfy one
+    // of the following:
+    // - Are an identical TM
+    // - Are on a consteval-able path
+
+    auto dps = mlir::cast<mlir::DestinationStyleOpInterface>(op.getOperation());
```
mtopalovicTT: slice has just one operand which is input, you don't need for loop

lib/Dialect/TTIR/Transforms/EraseInverseOps/SliceCommutePatterns.cpp
```diff
@@ -144,13 +145,290 @@ class TTIRCommutePermuteThroughSlice
...
+    for (mlir::OpOperand *operand : dps.getDpsInputOperands()) {
+      auto operandValue = operand->get();
+      if (checkIdenticalTms(operandValue.getDefiningOp(), reshapeOperand) ||
+          ttcore::valueTracesToConstantArgs(operandValue)) {
```
mtopalovicTT: Why do we need this           ttcore::valueTracesToConstantArgs(operandValue)

lib/Dialect/TTIR/Transforms/EraseInverseOps/SliceCommutePatterns.cpp
```diff
@@ -144,13 +145,290 @@ class TTIRCommutePermuteThroughSlice
...
+                                   ReshapeOp reshapeOperand) const override {
+    // Commuting downwards is favorable if all other operands satisfy one
+    // of the following:
+    // - Are an identical TM
+    // - Are on a consteval-able path
+
+    auto dps = mlir::cast<mlir::DestinationStyleOpInterface>(op.getOperation());
+    for (mlir::OpOperand *operand : dps.getDpsInputOperands()) {
+      auto operandValue = operand->get();
+      if (checkIdenticalTms(operandValue.getDefiningOp(), reshapeOperand) ||
```
mtopalovicTT: checkIdenticalTms this should be true for slice right when we are commuting reshape through slice

lib/Dialect/TTIR/Transforms/EraseInverseOps/SliceCommutePatterns.cpp
```diff
@@ -144,13 +145,290 @@ class TTIRCommutePermuteThroughSlice
...
+    // Larger right-to-left position means further left
+    return rightmostReshapeDimRTL > leftmostSlicedDimRTL;
+  }
+
+  bool isCommuteDownwardsFavorable(SliceStaticOp op,
+                                   ReshapeOp reshapeOperand) const override {
+    // Commuting downwards is favorable if all other operands satisfy one
+    // of the following:
+    // - Are an identical TM
+    // - Are on a consteval-able path
```
mtopalovicTT: Do we need to check that only user of reshape is slice or this is checked somewhere else?

lib/Dialect/TTIR/Transforms/EraseInverseOps/SliceCommutePatterns.cpp
```diff
@@ -144,13 +145,290 @@ class TTIRCommutePermuteThroughSlice
...
+            originalOutputShape[originalNdims - 1 - i];
+      }
+    }
+
+    // Update begins and steps to match new ends:
+    // - begins = 0
+    // - steps = 1
+    // for dims to the left of leftmost sliced dim
+
+    auto originalBegins = op.getBegins().getValue();
```
mtopalovicTT: Join this into above loop?

lib/Dialect/TTIR/Transforms/EraseInverseOps/SliceCommutePatterns.cpp
mvasiljevicTT: Forgot that is a valid type of reshape, fixed now 😅

lib/Dialect/TTIR/Transforms/EraseInverseOps/SliceCommutePatterns.cpp
mvasiljevicTT: Good point. In my understanding of how `isCommuteDownwardsViable/Favorable` were intended to be used this means it is always favorable, so I changed it like that now.

lib/Dialect/TTIR/Transforms/EraseInverseOps/SliceCommutePatterns.cpp
mvasiljevicTT: We don't 😄 
It is all removed now as the function just returns true

lib/Dialect/TTIR/Transforms/EraseInverseOps/SliceCommutePatterns.cpp
mvasiljevicTT: Yup, since slice has only one operand none of this is needed

lib/Dialect/TTIR/Transforms/EraseInverseOps/SliceCommutePatterns.cpp
mvasiljevicTT: It is checked [here](https://github.com/tenstorrent/tt-mlir/blob/120c1f82f16a96121f123ec9c5d43d91c7082a6a/include/ttmlir/Dialect/TTIR/Transforms/EraseInverseOps/EraseInverseOps.h#L76C1-L81C10)

lib/Dialect/TTIR/Transforms/EraseInverseOps/SliceCommutePatterns.cpp
mvasiljevicTT: Thanks, looks better now

odjuricicTT (approved):

lib/Dialect/TTIR/Transforms/EraseInverseOps/SliceCommutePatterns.cpp
```diff
@@ -140,22 +140,269 @@ class TTIRCommutePermuteThroughSlice
...
+
+    SliceStaticOp newSlice =
+        createReshapedSliceStaticOp(op, newReshape.getResult(), rewriter);
+
+    SmallVector<Operation *> users(op->getUsers());
+    for (auto *user : users) {
+      assert(checkIdenticalTms(reshapeUser, user) &&
+             "isCommuteUpwardsViable/Favorable should have ensured all users "
+             "are identical TMs");
+      rewriter.replaceOp(user, newSlice);
```
azecevic000: This will probably cause the same issue as this one https://github.com/tenstorrent/tt-mlir/pull/5767. Please fix this before merging.

mvasiljevicTT: Thanks! Changed in all EIO upwards patterns

Issue #5527, opened 2025-10-28
Failed to legalize: stablehlo.dynamic_update_slice


ndrakulicTT:
Here are examples:
```
stablehlo.dynamic_update_slice %1, %2, %3, %4, %5, %c_184, %c_184 : (tensor<3x1x24x64x32xbf16>, tensor<1x1x24x64x32xbf16>, tensor<i64>, tensor<i64>, tensor<i64>, tensor<i64>, tensor<i64>) -> tensor<3x1x24x64x32xbf16>
stablehlo.dynamic_update_slice %1, %2, %3, %4, %5, %c_184, %c_184 : (tensor<3x4x12x64x32xbf16>, tensor<1x4x12x64x32xbf16>, tensor<i64>, tensor<i64>, tensor<i64>, tensor<i64>, tensor<i64>) -> tensor<3x4x12x64x32xbf16>
stablehlo.dynamic_update_slice %1, %2, %3, %4, %5, %c_184, %c_184 : (tensor<3x16x6x64x32xbf16>, tensor<1x16x6x64x32xbf16>, tensor<i64>, tensor<i64>, tensor<i64>, tensor<i64>, tensor<i64>) -> tensor<3x16x6x64x32xbf16>
stablehlo.dynamic_update_slice %1, %2, %3, %4, %5, %c_184, %c_184 : (tensor<3x64x3x64x32xbf16>, tensor<1x64x3x64x32xbf16>, tensor<i64>, tensor<i64>, tensor<i64>, tensor<i64>, tensor<i64>) -> tensor<3x64x3x64x32xbf16>
stablehlo.dynamic_update_slice %1, %2, %3, %4, %5, %c_146, %c_146 : (tensor<3x1x24x49x32xbf16>, tensor<1x1x24x49x32xbf16>, tensor<i64>, tensor<i64>, tensor<i64>, tensor<i64>, tensor<i64>) -> tensor<3x1x24x49x32xbf16>
stablehlo.dynamic_update_slice %1, %2, %3, %4, %5, %c_146, %c_146 : (tensor<3x4x12x49x32xbf16>, tensor<1x4x12x49x32xbf16>, tensor<i64>, tensor<i64>, tensor<i64>, tensor<i64>, tensor<i64>) -> tensor<3x4x12x49x32xbf16>
[...]

PR #5714, opened 2025-11-05
Enable CPU-hoisting for `stablehlo.dynamic_update_slice`

dmilinkovicTT:
### Ticket
#5527 

### Problem description
`stablehlo.dynamic_update_slice` conversion to TTIR isn't supported. We need to hoist execution of this op to the CPU.

Most of the efforts for supporting this op are focused on enabling integer support in CPU-hoisted funcs.

### What's changed
- Allowed integer types in CPU-hoisted functions
- Appended `linalg.copy` to the end of CPU-hoisted functions if DPS semantics can't be enabled by other means
- Allowed creating `TTIRHoistPass` which would target only certain ops
- Generalized implementation of `TTIRHoistAnalyze` and `TTIRHoistTransform`, so that they receive a predicate which determines whether an op should be hoisted
- Disabled execution of `ElementTypeNormalization` on CPU module
- Fixed lowering of `ttnn.full` to flatbuffer
- Enabled CPU-hoisting stablehlo ops by default

### Checklist

### DIFF
--- a/tools/op-by-op-infra/op_by_op_infra/mlir_module_splitter.py
+++ b/tools/op-by-op-infra/op_by_op_infra/mlir_module_splitter.py
@@ -124,7 +124,7 @@ class MLIRModuleSplitter:
 
         if any(isinstance(op, _ttcore_ops_gen.CPUModuleOp) for op in funcs_list):
             raise ValueError(
-                "MLIRModuleSplitter does not support modules containing hoisted ops."
+                "MLIRModuleSplitter does not support modules containing CPU-hoisted ops."
             )
 
         # Extract operations from DeviceModuleOp if present.


lib/Dialect/TTNN/Pipelines/TTNNPipelines.cpp
```diff
@@ -20,6 +20,7 @@
 
 #include "mlir/Pass/PassManager.h"
 #include "mlir/Transforms/Passes.h"
+#include <mlir/IR/BuiltinOps.h>
```
vwellsTT: nit: "" for mlir includes

lib/Dialect/TTIR/Transforms/HoistCPUOps.cpp
```diff
@@ -180,27 +201,21 @@ static void hoistOperationToFunction(mlir::Operation *opToHoist,
     // Clone the operation, but modify its type if needed.
     auto *clonedOp = builder.clone(*opToHoist, mapping);
 
-    // Update operand types to f32 for tensor types.
+    // Update operand types to supported tensor types.
```
vwellsTT, line 204: we cannot relax this constraint without supporting non-f32 types in the runtime, imo

vwellsTT: sync'ed offline, I missed above changes which add some support for i32, and David says this does work at runtime.  I think it might be good to "formalize" support for i32 + f32 vs just f32 in the runtime TensorWrapper and so on, but don't feel super strongly I guess

vwellsTT (approved):
I still think it'd be good to add some `builder` tests before merging, FWIW
But otherwise great work, having int support is nice!

mtopalovicTT (approved):

lib/Dialect/TTIR/Pipelines/TTIRPipelines.cpp
```diff
@@ -30,6 +30,7 @@
 #include "ttmlir/Dialect/TTCore/Transforms/Passes.h"
 #include "ttmlir/Dialect/TTIR/Transforms/Passes.h"
 #include "ttmlir/Transforms/Passes.h"
+#include <stablehlo/dialect/StablehloOps.h>
```
azecevic000: The build will fail if `TTMLIR_ENABLE_STABLEHLO=false`; put this in the section under that is guarded, also use `""` instead of `<>`

lib/Dialect/TTIR/Pipelines/TTIRPipelines.cpp
```diff
@@ -72,10 +83,17 @@ void createStableHLOToTTIRPipeline(
...
-               stablehlo::StablehloDialect>());
+
+    // Fallback allow-listed SHLO ops to the CPU module.
+    pm.addPass(createHoistAllowlistedStablehloOpsPass());
+
+    // Run StableHLOToTTIR pass again on the DeviceModule to produce a failure
+    // if any unsupported ops have not been hoisted.
+    auto &devicePm = pm.nest<ttcore::DeviceModuleOp>().nest<mlir::ModuleOp>();
+    passOptions.enablePartialConversion = false;
+    devicePm.addPass(createConvertStableHLOToTTIRPass(passOptions));
```
azecevic000, line 95: The natural order here should be `SHLO->TTIR` and then cpu fallback. Otherwise, we might end up with a cpu op that could be executed on device. I understand from implementation perpective why you did it this way, one alternative to this would be to change the type of `SHLO->TTIR` from conversion to greedy rewriter, but that would require some non-trivial changes. Other sound option would be to first run `SHLO->TTIR` with `enablePartialConversion=true`, then cpu fallback, and finally `SHLO->TTIR` with `enablePartialConversion=false` (this second run would be pretty much verification run).

Edit: Thanks for pointing it out for me offline, I missed `createConvertStableHLOToTTIRPass` before `if (options.enableCPUFallback)`, so the latter proposal is actually already implemented here.

lib/Dialect/TTIR/Transforms/HoistCPUOps.cpp
```diff
@@ -293,33 +309,40 @@ static void hoistOperationToFunction(mlir::Operation *opToHoist,
   opToHoist->erase();
 }
 
-// An analysis class which currently relies on manually tagging ops with a
-// `should_hoist` attribute, but in the future will also tag fall-back ops, etc.
 namespace {
+// Predicate type for determining whether an op should be hoisted.
+using ShouldHoistPredicate = std::function<bool(mlir::Operation *)>;
```
azecevic000: Nit: I find convention of putting trailing `T`, `Ty` or `Type` very useful in cases like this one. `ShouldHoistPredicate` sounds more like a function name than function type name.

lib/Dialect/TTIR/Transforms/HoistCPUOps.cpp
```diff
@@ -397,23 +433,42 @@ class TTIRHoistTransform
...
+    return ((op->getDialect()->getTypeID() == TypeID::get<Dialects>()) || ...);
+  };
+  auto pass = std::make_unique<TTIRHoistTransform>(customPredicate);
+  return pass;
+}
+
+template <typename... Ops>
+std::unique_ptr<mlir::Pass> createTTIRHoistTransformForOps() {
+  const auto customPredicate = [](mlir::Operation *op) {
+    return (llvm::isa<Ops>(op) || ...);
```
azecevic000: Nit: `llvm::isa<Ops...>(op)` should be the same and it's little more readable IMO

lib/Dialect/TTIR/Pipelines/TTIRPipelines.cpp
dmilinkovicTT: VSCode keeps inserting random non-required headers for some reason

lib/Dialect/TTIR/Pipelines/TTIRPipelines.cpp
dmilinkovicTT: Synced offline

dmilinkovicTT:
@vwellsTT thanks for the review!

I've added a silicon test for `dynamic_update_slice`, and I've updated some hoisted op builder tests so that they are parametrized with both f32 and i32 types.

I've also adjusted the ptr type in `WrappedTensor`, as we've discussed offline

PR #5716, opened 2025-11-05
Support for batch > 1 fill/update cache.

LPanosTT:
### Ticket
https://github.com/tenstorrent/tt-xla/issues/1852

### Problem description
- `ttnn.fill_cache` does not support updating all batches at once.
- `ttnn.update_cache` **does** support updating all batches at once, however to do so the input must be in the form `[1, num_heads, B, head_size]`

### What's changed
- Update the fusion pattern for fill/update cache
   - For the fill cache case, insert a slice on the batch to pick out every batch and apply fill_cache to each one
   - For the update cache case, permute the fill values to `[1, num_heads, batch, head_size]`

### Checklist

### DIFF
--- a/lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
+++ b/lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
@@ -1154,9 +1154,11 @@ public:
           SmallVector<int64_t> sliceOutputShape(updatesType.getShape());
           sliceOutputShape[0] = 1;
 
+          // Encoding should not be set when this pass is run. Guard against it.
+          assert(!updatesType.getEncoding());
+
           RankedTensorType slicedUpdatesType = RankedTensorType::get(
-              sliceOutputShape, updatesType.getElementType(),
-              updatesType.getEncoding());
+              sliceOutputShape, updatesType.getElementType(), nullptr);
 
           // Create slice op.
           auto slicedUpdates = ttir::utils::createDPSOp<SliceStaticOp>(
@@ -1189,9 +1191,11 @@ public:
       if (batchSize > 1) {
         SmallVector<int64_t> permutedShape = ttmlir::utils::applyPermutation(
             updatesType.getShape(), {2, 1, 0, 3});
-        RankedTensorType permutedUpdatesType =
-            RankedTensorType::get(permutedShape, updatesType.getElementType(),
-                                  updatesType.getEncoding());
+
+        // Encoding should not be set when this pass is run. Guard against it.
+        assert(!updatesType.getEncoding());
+        RankedTensorType permutedUpdatesType = RankedTensorType::get(
+            permutedShape, updatesType.getElementType(), nullptr);
         updates = ttir::utils::createDPSOp<PermuteOp>(
             rewriter, scatterOp.getLoc(), permutedUpdatesType, updates,
             rewriter.getDenseI64ArrayAttr({2, 1, 0, 3}));


lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
```diff
@@ -1109,29 +1109,98 @@ class CacheFillUpdatePattern : public mlir::OpRewritePattern<ScatterOp> {
...
-          scatterOp, scatterOp.getResult().getType(), // Result type
-          cache,                                      // Cache tensor
-          updates,                                    // Updates tensor
-          *CachePositions,                            // Cache Idx
-          batchOffsetAttr                             // Batch offset
+      // Unlike ttnn.fill_cache, we can perform ttnn.update_cache on the entire
+      // batch at once. However this requires that the fill value is in the form
+      // [1, num_heads, B, head_size]. So, we must permute the updates tensor to
+      // this shape.
+      if (batchSize > 1) {
```
odjuricicTT, line 1191: When reading the cache in the next decode iteration, how do we know that it is permuted?

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
```diff
@@ -1109,29 +1109,98 @@ class CacheFillUpdatePattern : public mlir::OpRewritePattern<ScatterOp> {
 
     Value cache = scatterOp.getInput();
     Value updates = scatterOp.getUpdate();
-    auto batchOffsetAttr = rewriter.getI32IntegerAttr(0);
+
+    int32_t batchSize =
+        mlir::cast<RankedTensorType>(cache.getType()).getShape()[0];
```
azecevic000: Nit: I recommend to update types of `cache` and `updates` to `auto`, so that you don't have to use cast to `RankedTensorType` for `batchSize` and `updatesType`.

test/ttmlir/Silicon/TTNN/n150/fusing/fill_cache_batch32.mlir
azecevic000, line 1: Can you move this test to `test/ttmlir/Dialect/TTIR/fusing` dir, with some smaller batch (2 is sufficient). Batch 32 silicon test should be a builder test.

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
```diff
@@ -1109,29 +1109,98 @@ class CacheFillUpdatePattern : public mlir::OpRewritePattern<ScatterOp> {
...
-          *CachePositions,                            // Cache Idx
-          batchOffsetAttr                             // Batch offset
+      // Unlike ttnn.fill_cache, we can perform ttnn.update_cache on the entire
+      // batch at once. However this requires that the fill value is in the form
+      // [1, num_heads, B, head_size]. So, we must permute the updates tensor to
+      // this shape.
+      if (batchSize > 1) {
+        SmallVector<int64_t> permutedShape(updatesType.getShape());
+        permutedShape = ttmlir::utils::applyPermutation(updatesType.getShape(),
+                                                        {2, 1, 0, 3});
```
azecevic000:
```suggestion
        SmallVector<int64_t> permutedShape = ttmlir::utils::applyPermutation(updatesType.getShape(), {2, 1, 0, 3});
```

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
```diff
@@ -1109,29 +1109,98 @@ class CacheFillUpdatePattern : public mlir::OpRewritePattern<ScatterOp> {
...
+          SmallVector<int32_t> sliceSteps = {1, 1, 1, 1};
+
+          // Slice output shape is the same as the fill value shape, except the
+          // batch dim is 1 since we sliced out a single batch.
+          SmallVector<int64_t> sliceOutputShape(updatesType.getShape());
+          sliceOutputShape[0] = 1;
+
+          RankedTensorType slicedUpdatesType = RankedTensorType::get(
+              sliceOutputShape, updatesType.getElementType(),
+              updatesType.getEncoding());
```
azecevic000: I would `assert(!updatesType.getEncoding())` before this and set encoding to `nullptr`. If `TTNNLayoutAttr` really exists `updatesType.getEncoding()` will be wrong type of encoding for this op, since encoding is dependent on shape. Other option is to have full `if (encoding) {...} else {...}`, but it would be an overkill and just make the code less readable, `assert` would be there only to guard against running the pass after encoding is set.

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
```diff
@@ -1109,29 +1109,98 @@ class CacheFillUpdatePattern : public mlir::OpRewritePattern<ScatterOp> {
...
+      // batch at once. However this requires that the fill value is in the form
+      // [1, num_heads, B, head_size]. So, we must permute the updates tensor to
+      // this shape.
+      if (batchSize > 1) {
+        SmallVector<int64_t> permutedShape(updatesType.getShape());
+        permutedShape = ttmlir::utils::applyPermutation(updatesType.getShape(),
+                                                        {2, 1, 0, 3});
+        RankedTensorType permutedUpdatesType =
+            RankedTensorType::get(permutedShape, updatesType.getElementType(),
+                                  updatesType.getEncoding());
```
azecevic000: Same here.

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
```diff
@@ -1109,29 +1109,98 @@ class CacheFillUpdatePattern : public mlir::OpRewritePattern<ScatterOp> {
...
+        updates = ttir::utils::createDPSOp<PermuteOp>(
+            rewriter, scatterOp.getLoc(), permutedUpdatesType, updates,
+            rewriter.getDenseI64ArrayAttr({2, 1, 0, 3}));
+      }
+      cache = rewriter.create<UpdateCacheOp>(
```
odjuricicTT, line 1203: As discussed offline, we should now be able to replace this with PagedUpdateCache. Once that is done i'll take a another look at the PR

odjuricicTT (approved):

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
LPanosTT: As further discussed offline. Llama's hugging face implementation forces identical cache position across batches. It does this by masking out padding tokens in between the prompt and generated tokens. Thus, using `paged_update_cache` for the purpose of being able to specify a cache position for each batch is irrelevant.

PR #5718, opened 2025-11-05
[Uplift Enhancement] Permit custom metal commit hash for manual uplift, with safety check to disallow downgrade + metal branch commits

jameszianxuTT:
### Ticket
None

### Problem description

Not much flexibility to manual uplift job. We might want to uplift a smaller section of commits from metal, but we can't since the job selects the most recent commit.

### What's changed
- Add optional custom metal commit (default nothing = use tip of metal main like before)
- Add safety checks for custom metal commit -> Disallow uplifting a metal downgrade, disallow uplifting to a non-main metal branch

This is a pure QOL change for uplifters. It could have been done manually before by manually editing a commit post-creation

This is an additional enhancement on top of #5616 

### Checklist
- [x] New/Existing tests provide coverage for changes (see comments for validation)

### DIFF
--- a/.github/workflows/schedule-nightly-uplift.yml
+++ b/.github/workflows/schedule-nightly-uplift.yml
@@ -77,7 +77,7 @@ jobs:
           echo "Current tt-metal version: $CURRENT_TT_METAL_VERSION"
           echo "Custom tt-metal version: ${{ env.UPLIFTED_TT_METAL_VERSION }}"
 
-          # Verify custom commit is on main branch
+          # Verify custom commit is on main metal branch
           MAIN_TIP=$(gh api repos/tenstorrent/tt-metal/commits/main --jq '.sha')
           COMPARE_TO_MAIN=$(gh api repos/tenstorrent/tt-metal/compare/${{ env.UPLIFTED_TT_METAL_VERSION }}...${MAIN_TIP} --jq '.status')
 
@@ -92,7 +92,7 @@ jobs:
           # Verify custom commit is not a downgrade from current version
           COMPARE_RESULT=$(gh api repos/tenstorrent/tt-metal/compare/${CURRENT_TT_METAL_VERSION}...${{ env.UPLIFTED_TT_METAL_VERSION }} --jq '.status')
 
-          if [ "$COMPARE_RESULT" = "behind" ] || [ "$COMPARE_RESULT" = "diverged" ]; then
+          if [ "$COMPARE_RESULT" = "behind" ]; then
             echo "::error::Custom commit hash ${{ env.UPLIFTED_TT_METAL_VERSION }} would be a downgrade from current version $CURRENT_TT_METAL_VERSION (status: $COMPARE_RESULT)"
             echo "::error::Please use a commit hash that is ahead of the current version."
             exit 1


jameszianxuTT:
Verification of 4 cases:

1. I fed a custom commit hash that would downgrade metal relative to current tt-mlir main. We should never do this. [This does not go through as expected](https://github.com/tenstorrent/tt-mlir/actions/runs/19109125847).
2. I fed a custom commit hash from a tt-metal branch. Since we should never uplift to a non-main metal commit, [this is also explicitly disallowed.](https://github.com/tenstorrent/tt-mlir/actions/runs/19109363381/job/54602016979#step:6:38)
3. I fed a custom commit hash that (1) is an upgrade from main and (2) is not at tip of metal. This is the primary usecase of the custom commit hash. [This does work as expected.](https://github.com/tenstorrent/tt-mlir/actions/runs/19109403525/job/54602161614)
4. I fed no custom commit hash to test default behaviour (create uplift branch with custom name, using tip of metal). This works as expected, and [was the previous behavior.](https://github.com/tenstorrent/tt-mlir/actions/runs/19109471192/job/54602401725#step:5:16)

jameszianxuTT:
Example screenshot:
image

jameszianxuTT:
cc @ctodTT, newly appointed uplifter

jameszianxuTT:
This is to improve QOL specifically for the stacked uplift workflow:
image

.github/workflows/schedule-nightly-uplift.yml
```diff
@@ -51,12 +55,50 @@ jobs:
...
+            echo "::error::Please use a commit hash from the main branch."
+            exit 1
+          fi
+
+          echo "Validation passed: Custom commit is on main branch"
+
+          # Verify custom commit is not a downgrade from current version
+          COMPARE_RESULT=$(gh api repos/tenstorrent/tt-metal/compare/${CURRENT_TT_METAL_VERSION}...${{ env.UPLIFTED_TT_METAL_VERSION }} --jq '.status')
+
+          if [ "$COMPARE_RESULT" = "behind" ] || [ "$COMPARE_RESULT" = "diverged" ]; then
```
brataTT: If already pointing to a non-main branch, this could fail.
I'd prefer getting rid of the "diverged" comparison here to encourage getting out of using a non-main branch faster.

.github/workflows/schedule-nightly-uplift.yml
```diff
@@ -51,12 +55,50 @@ jobs:
...
+          if [ -n "${{ inputs.custom_metal_commit_hash }}" ]; then
+            UPLIFTED_TT_METAL_VERSION="${{ inputs.custom_metal_commit_hash }}"
+            echo "Using custom tt-metal commit hash: $UPLIFTED_TT_METAL_VERSION"
+          else
+            UPLIFTED_TT_METAL_VERSION=$(gh api repos/tenstorrent/tt-metal/commits/main --jq '.sha')
+            echo "Using latest tt-metal commit from main: $UPLIFTED_TT_METAL_VERSION"
+          fi
+          echo "UPLIFTED_TT_METAL_VERSION=$UPLIFTED_TT_METAL_VERSION" >> $GITHUB_ENV
+
+      - name: Validate no downgrade + using main when using custom commit hash for safety
```
brataTT, line 71: Thanks for adding this

Having this as fatal to the PR creation job instead of an easy to ignore warning makes sense
If someone wants to bypass this (older to get rid of regression, non-main to catch future regressions faster), manually editing later will always be an option

.github/workflows/schedule-nightly-uplift.yml
jameszianxuTT: Good point, thanks.

PR #5519, opened 2025-10-27
Add macOS build support with runtime stubs

zoecarver:
### Problem description
Currently, you cannot run the compiler pipeline on macOS due to missing runtime (linker errors). This stubs out the runtime so that you can still run the compiler + passes on macOS.

### What's changed
* Added a stub file
* Integrated into build
* Docs

PR #5723, opened 2025-11-05
[no-fork] Add macOS build support with runtime stubs (#5519)

zoecarver:
Problem description

Currently, you cannot run the compiler pipeline on macOS due to missing runtime (linker errors). This stubs out the runtime so that you can still run the compiler + passes on macOS.

What's changed

Added a stub file
Integrated into build
Docs

Exactly the same as #5519 (no fork)


vtangTT (approved):

zoecarver:
Thank you both!

PR #5729, opened 2025-11-05
Add suppport for composite op sharding

sshonTT:
### Ticket
[Link](https://github.com/tenstorrent/tt-xla/issues/1838)

### Problem description
For composite op running

### What's changed
This change enables sharding propagation through StableHLO stablehlo.composite operations by:
1. Flattening each composite callsite: inline (decompose) the callee body into the caller and tag the cloned ops with a group marker for later recovery.
2. Propagating sharding across the now-flat graph so standard passes can do their job without needing composite-aware logic.
3. Re-outlining the exact grouped region back into a stablehlo.composite op, restoring its original name and user-visible attributes.

This keeps composite boundaries in the final IR while letting sharding flow across them during optimization.

### Checklist

### DIFF
--- a/lib/Dialect/StableHLO/Transforms/FlattenComposite.cpp
+++ b/lib/Dialect/StableHLO/Transforms/FlattenComposite.cpp
@@ -79,6 +79,14 @@ flattenOneComposite(mlir::stablehlo::CompositeOp comp,
   // 3) Prepare mapping from callee arguments to composite operands.
   mlir::IRMapping mapping;
   auto &calleeEntry = callee.getBody().front();
+  if (static_cast<int64_t>(calleeEntry.getNumArguments()) !=
+      comp.getNumOperands()) {
+    comp.emitOpError() << "number of operands (" << comp.getNumOperands()
+                       << ") does not match number of callee arguments ("
+                       << calleeEntry.getNumArguments() << ") in function '"
+                       << leaf << "' during inlining.";
+    return mlir::failure();
+  }
   for (int64_t i = 0; i < static_cast<int64_t>(comp->getNumOperands()); ++i) {
     mlir::BlockArgument arg = calleeEntry.getArgument(i);
     mapping.map(arg, comp->getOperand(i));
@@ -120,6 +128,13 @@ flattenOneComposite(mlir::stablehlo::CompositeOp comp,
     return mlir::failure();
   }
 
+  if (static_cast<int64_t>(ret.getNumOperands()) != comp.getNumResults()) {
+    comp.emitOpError() << "number of return operands (" << ret.getNumOperands()
+                       << ") does not match number of composite results ("
+                       << comp.getNumResults() << ") in function '" << leaf
+                       << "' during inlining.";
+    return mlir::failure();
+  }
   for (int64_t i = 0; i < static_cast<int64_t>(comp->getNumResults()); ++i) {
     mlir::Value mapped = mapping.lookupOrNull(ret.getOperand(i));
     if (!mapped) {


lib/Dialect/StableHLO/Transforms/FlattenComposite.cpp
```diff
@@ -0,0 +1,164 @@
...
+    llvm::raw_string_ostream os(groupName);
+    os << "composite_" << callee.getSymName();
+  }
+  mlir::StringAttr groupAttr = builder.getStringAttr(groupName);
+  mlir::UnitAttr seedAttr = builder.getUnitAttr();
+
+  // 3) Prepare mapping from callee arguments to composite operands.
+  mlir::IRMapping mapping;
+  {
+    auto calleeEntry = &callee.getBody().front();
```
vmilosevic: :warning: **llvm\-qualified\-auto** :warning:
`` auto calleeEntry `` can be declared as `` auto *calleeEntry ``
```suggestion
    auto *calleeEntry = &callee.getBody().front();
```

lib/Dialect/StableHLO/Transforms/ReoutlineComposite.cpp
```diff
@@ -0,0 +1,373 @@
...
+      return false;
+    }
+  }
+
+  // Captures: operands whose defining op is not in the group (or is a block
+  // arg).
+  llvm::SmallDenseSet<mlir::Value> captureSet;
+  for (auto *op : sorted) {
+    for (mlir::Value v : op->getOperands()) {
+      if (auto defOp = v.getDefiningOp()) {
```
vmilosevic: :warning: **llvm\-qualified\-auto** :warning:
```suggestion
      if (auto *defOp = v.getDefiningOp()) {
```

lib/Dialect/StableHLO/Utils/ShardyUtils.cpp
```diff
@@ -287,9 +287,32 @@ getDefaultTensorSdyShardingAttr(MLIRContext *context, llvm::StringRef meshName,
 // Get the argument sharding attributes.
 llvm::SmallVector<mlir::sdy::TensorShardingAttr>
 getInShardingAttrs(MLIRContext *context, func::FuncOp &funcOp,
-                   mlir::sdy::MeshOp &globalMeshOp) {
+                   mlir::sdy::MeshOp &globalMeshOp, bool createIfMissing) {
   llvm::SmallVector<mlir::sdy::TensorShardingAttr> inShardingAttrs;
 
+  auto createDefaultShardingAttr = [&](uint32_t argNum) {
```
hkwonTT, line 295: I think it's a great idea.

lib/Dialect/StableHLO/Utils/ShardyUtils.cpp
```diff
@@ -298,34 +321,14 @@ getInShardingAttrs(MLIRContext *context, func::FuncOp &funcOp,
...
-
-      // Set argument with it's default sdy.sharding attribute
-      mlir::DictionaryAttr newDictAttr =
-          shardy_utils::addDictionaryAttrSdyShardingAnnotation(context,
-                                                               shardingAttr);
-      funcOp.setArgAttrs(arg.getArgNumber(), newDictAttr);
+      shardingAttr = createDefaultShardingAttr(arg.getArgNumber());
+    }
+    if (!shardingAttr) {
+      continue;
```
hkwonTT: Are you sure we can skip appending `shardingAttrs` to the `inShardingAttrs` vector? I thought this function is supposed to return a `shardingAttr` for each argument. Otherwise, we can’t be sure the indices line up between `shardingAttrs` and the arguments.
@tapspatel How do you think?

sshonTT: I don't have any issue to remove it. I will remove this line.

include/ttmlir/Dialect/StableHLO/Transforms/Passes.td
```diff
@@ -236,4 +255,22 @@ def InsertExplicitReshardsPass : Pass<"insert-explicit-reshards", "::mlir::Modul
...
 }
 
+def ReoutlineCompositePass : Pass<"reoutline-composite", "::mlir::ModuleOp"> {
+  let summary = "Reconstruct composite ops from flattened groups marked by reoutline.* attributes.";
+  let description = [{
+    This pass scans for ops annotated with `reoutline.group` and attempts to
+    re-outline them into a single composite/call if the group forms a closed,
+    movable subgraph within the function. Captured inputs become function arguments
+    and escaping values become function results. If a group is not closed or cannot
+    be moved safely due to side effects, it is skipped and the flattened form remains.
```
hshahTT, line 299: Potentially stupid question, but this kind of implies that re-constructing the composite op isn't needed to allow the graph to later execute and work: if that's the case then what is the benefit of reconstructing the composite op back?

include/ttmlir/Dialect/StableHLO/Utils/ShardyUtils.h
```diff
@@ -109,7 +109,8 @@ getDefaultTensorSdyShardingAttr(MLIRContext *context, llvm::StringRef meshName,
 // Get the argument sharding attributes.
 llvm::SmallVector<mlir::sdy::TensorShardingAttr>
 getInShardingAttrs(MLIRContext *context, func::FuncOp &funcOp,
-                   mlir::sdy::MeshOp &globalMeshOp);
+                   mlir::sdy::MeshOp &globalMeshOp,
+                   bool createIfMissing = true);
```
hshahTT, line 113: Could you add a comment on what the `createIfMissing` flag is supposed to do here?

include/ttmlir/Dialect/StableHLO/Transforms/Passes.td
sshonTT: It would run without re-composing. But if we can make back it to composite op, it would convert to one (or multiple) ttir op from [here](https://github.com/tenstorrent/tt-mlir/blob/main/lib/Conversion/StableHLOToTTIR/StableHLOLegalizeCompositePass.cpp#L48), which gives us performance benefit.

include/ttmlir/Dialect/StableHLO/Utils/ShardyUtils.h
hshahTT: Actually disregard this, my bad. I initially thought that this was related to your reoutline constants in ShardingUtils.h, but it's not. The variable name is clear enough without needing comments.

sshonTT: lol I just added comment. No worries, I think there’s no harm in adding the comments. Thanks a lot for the review!

test/ttmlir/Conversion/StableHLOToTTIR/composite/test_sharding_composite.mlir
```diff
@@ -0,0 +1,59 @@
+// REQUIRES: stablehlo
+// RUN: ttmlir-opt -split-input-file  --stablehlo-pipeline --stablehlo-to-ttir-pipeline -o %t %s
+// RUN: FileCheck %s --input-file=%t
+
+module @TwoGelus attributes {mhlo.cross_program_prefetches = [], mhlo.frontend_attributes = {xla.sdy.meshes = "{mesh = #sdy.mesh<[\22_axis_0\22=2]>}"}, mhlo.input_output_alias = [], mhlo.is_dynamic = false, mhlo.use_auto_spmd_partitioning = false} {
```
hshahTT, line 5: Could you also add failure cases for both passes? Not sure when the FlattenComposite would fail but we should definitely have a test case for Reoutline failing and leaving the ops flattened.

hshahTT (changes requested):

test/ttmlir/Conversion/StableHLOToTTIR/composite/test_sharding_composite.mlir
sshonTT: I added failure case for ReoutlineComposite pass, but regarding FlattenComposite pass, it is not expected to be failed as it is simple unrolling behaviour.

sshonTT: I added a test case for the flatten-composite pass to verify that the flattened ops are correctly grouped.

include/ttmlir/Dialect/StableHLO/Transforms/Passes.td
```diff
@@ -145,6 +145,25 @@ def DecoupleConstFanoutPass : Pass<
...
+    This pass inlines each `stablehlo.composite` by cloning its callee body at the call site.
+    Every cloned op is annotated with `reoutline.group = "composite_<callee>"`.
+    The first cloned op additionally receives `reoutline.seed` (UnitAttr).
+    The original composite op is erased after wiring the cloned return values to the users.
+
+    The pass is intended to run **before** sharding propagation so that propagation happens
+    on a flattened graph. A later re-outlining pass may reconstruct composites by grouping
+    ops with the same `reoutline.group` attribute if the subgraph is closed and movable.
+  }];
+
```
tapspatel, line 194: can you put an example inline of what this pass is doing

see `ApplyArgumentShardStatusPass`

include/ttmlir/Dialect/StableHLO/Transforms/Passes.td
```diff
@@ -236,4 +255,22 @@ def InsertExplicitReshardsPass : Pass<"insert-explicit-reshards", "::mlir::Modul
...
+    This pass scans for ops annotated with `reoutline.group` and attempts to
+    re-outline them into a single composite/call if the group forms a closed,
+    movable subgraph within the function. Captured inputs become function arguments
+    and escaping values become function results. If a group is not closed or cannot
+    be moved safely due to side effects, it is skipped and the flattened form remains.
+
+    Use this after sharding propagation and reshards insertion if you want to
+    restore composite structure for subsequent conversions.
+  }];
+
```
tapspatel, line 348: insert example inline here

include/ttmlir/Dialect/StableHLO/Utils/ShardyUtils.h
tapspatel: this was my mistake tbh when creating this function. On second thought, we should make it a pure function where it is not modifying any state outside of it, and a separate call adds a sharding annotation to an arg which the caller controls. Not blocking but fyi

tapspatel: I see `getOutShardingAttrs` was created correctly

lib/Dialect/StableHLO/Transforms/FlattenComposite.cpp
```diff
@@ -0,0 +1,159 @@
...
+#include "ttmlir/Dialect/StableHLO/Transforms/Passes.h.inc"
+
+/// Inline a single stablehlo.composite op. Returns success if it was flattened.
+static mlir::LogicalResult
+flattenOneComposite(mlir::stablehlo::CompositeOp comp,
+                    mlir::SymbolTable &symTable) {
+  mlir::Operation *op = comp.getOperation();
+  mlir::OpBuilder builder(op);
+
+  // 0) Read metadata from the original composite.
```
tapspatel, line 25: really like your style of comments, I will be stealing this for myself :)

lib/Dialect/StableHLO/Transforms/FlattenComposite.cpp
```diff
@@ -0,0 +1,159 @@
...
+    : public impl::FlattenCompositePassBase<FlattenCompositePass> {
+  using impl::FlattenCompositePassBase<
+      FlattenCompositePass>::FlattenCompositePassBase;
+
+public:
+  void runOnOperation() override {
+    mlir::ModuleOp module = getOperation();
+
+    // If the module is not sharded, we skip this pass.
+    if (!shardy_utils::isShardedModule(module)) {
```
tapspatel, line 177: not part of this PR but I think we should have a normalizeArgumentResultShardings pass (or some equivalent name), similar to the anlayzeMesh pass that normalizes the mesh. It would make any modifications to the graph like inserting a replicated sharding. I believe we also have duplicate logic for inserting replicating shardings in analyze mesh as well.

lib/Dialect/StableHLO/Transforms/FlattenComposite.cpp
```diff
@@ -0,0 +1,159 @@
...
+      return;
+    }
+
+    mlir::SymbolTable symTable(module);
+
+    for (mlir::func::FuncOp funcOp : module.getOps<mlir::func::FuncOp>()) {
+      for (auto compositeOp : llvm::make_early_inc_range(
+               funcOp.getOps<mlir::stablehlo::CompositeOp>())) {
+        if (mlir::failed(flattenOneComposite(compositeOp, symTable))) {
```
tapspatel: insert some error message if failed

eg
```
rootModule.emitError("some error");
```

lib/Dialect/StableHLO/Transforms/FlattenComposite.cpp
```diff
@@ -0,0 +1,159 @@
...
+namespace mlir::tt::stablehlo {
+#define GEN_PASS_DEF_FLATTENCOMPOSITEPASS
+#include "ttmlir/Dialect/StableHLO/Transforms/Passes.h.inc"
+
+/// Inline a single stablehlo.composite op. Returns success if it was flattened.
+static mlir::LogicalResult
+flattenOneComposite(mlir::stablehlo::CompositeOp comp,
+                    mlir::SymbolTable &symTable) {
+  mlir::Operation *op = comp.getOperation();
+  mlir::OpBuilder builder(op);
```
tapspatel: nit: 

usually we want to initialize builder in terms of the context, and not the op
```
mlir::OpBuilder builder(context);
```

you can pass it in as an argument to this helper

lib/Dialect/StableHLO/Transforms/FlattenComposite.cpp
```diff
@@ -0,0 +1,159 @@
...
+
+  // 0) Read metadata from the original composite.
+  mlir::StringAttr origName = builder.getStringAttr(comp.getName());
+  // Preserve original composite attributes.
+  mlir::DictionaryAttr origCompAttrs = comp.getCompositeAttributes();
+  if (!origCompAttrs) {
+    // Create an empty attribute if none exists.
+    origCompAttrs = mlir::DictionaryAttr::get(builder.getContext());
```
tapspatel: nit:

pass in context as well
```
MLIRContext *context = rootModule.getContext();
```

lib/Dialect/StableHLO/Transforms/FlattenComposite.cpp
```diff
@@ -0,0 +1,159 @@
...
+  }
+
+  // 1) Resolve callee from 'decomposition' (SymbolRefAttr).
+  // stablehlo.composite "name" %arg {decomposition = @foo}
+  auto decompAttr = op->getAttrOfType<mlir::SymbolRefAttr>("decomposition");
```
tapspatel: make `decomposition` a string literal

lib/Dialect/StableHLO/Transforms/FlattenComposite.cpp
```diff
@@ -0,0 +1,159 @@
...
+  if (!decompAttr) {
+    // Some forks expose an accessor; use it if available:
+    // decompAttr = comp.getDecompositionAttr();
+    return mlir::failure();
```
tapspatel, line 43: insert error message 

eg from another file
```
funcOp.emitError("In function ")
          << funcOp.getName() << " argument #: " << arg->getArgNumber()
          << " does not have an argument dictionary. This is required in order "
             "to do analysis on ttir.name tensor annotations";
```

lib/Dialect/StableHLO/Transforms/FlattenComposite.cpp
```diff
@@ -0,0 +1,159 @@
...
+  }
+
+  // SymbolRefAttr may be nested; use the leaf for func name.
+  mlir::StringAttr leaf = decompAttr.getLeafReference();
+  mlir::func::FuncOp callee = symTable.lookup<mlir::func::FuncOp>(leaf);
+  if (!callee || callee.getBody().empty()) {
+    return mlir::failure();
```
tapspatel, line 64: error message

lib/Dialect/StableHLO/Transforms/FlattenComposite.cpp
```diff
@@ -0,0 +1,159 @@
...
+  }
+
+  // We will inline *before* the composite op.
+  builder.setInsertionPoint(comp);
+  // mlir::OpBuilder builder(comp);
```
tapspatel: remove commented line

lib/Dialect/StableHLO/Transforms/FlattenComposite.cpp
```diff
@@ -0,0 +1,159 @@
...
+    clonedOps.push_back(cloned);
+  }
+
+  // 5) Handle callee terminator: replace composite results with mapped return
+  // values. We expect a func.return with N operands, matching the composite's
+  // results.
+  auto ret = llvm::dyn_cast<mlir::func::ReturnOp>(
+      callee.getBody().front().getTerminator());
+  if (!ret) {
+    return mlir::failure();
```
tapspatel, line 128: error message, will help external callers of this pass know where it failed

tapspatel (approved):
<edit>
I reread the comments on the issue and it sounds like you guys have already discussed this. I'm not aware of the full scope of composite ops so you may disregard the following comments. 

nothing blocking. although I would propose another approach to this

I believe in this situation, you just need to add a propagation rule to `stablehlo.composite`. Essentially, stablehlo.composite is just a form of calling another function from your main control flow. We shouldn't restrict ourselves for propagation of having to inline tbh, even though we do it today (I put in this code so not related to your changes, just my thoughts). 

Propagation can and should be able to flow through

example
```
module @FlattenCompositeGrouping attributes {mhlo.cross_program_prefetches = [], mhlo.input_output_alias = [], mhlo.is_dynamic = false, mhlo.use_auto_spmd_partitioning = false} {
  sdy.mesh @mesh = <["_axis_0_updated"=1, "_axis_0"=2]>
  func.func @main(%arg0: tensor<32x32xf32> {sdy.sharding = #sdy.sharding<@mesh, [{}, {"_axis_0"}]>, ttcore.shard_status = #ttcore.shard_status<presharded>}) -> (tensor<32x32xf32> {ttcore.shard_status = #ttcore.shard_status<unsharded>}) {
    %0 = stablehlo.reshape %arg0 {sdy.sharding = #sdy.sharding_per_value<[<@mesh, [{?}, {?}, {"_axis_0", ?}]>]>} : (tensor<32x32xf32>) -> tensor<1x32x32xf32>
    %1 = stablehlo.reshape %0 {sdy.sharding = #sdy.sharding_per_value<[<@mesh, [{?}, {"_axis_0", ?}]>]>} : (tensor<1x32x32xf32>) -> tensor<32x32xf32>
    %2 = stablehlo.composite "tenstorrent.gelu_tanh" %1 {sdy.sharding = #sdy.sharding_per_value<[<@mesh, [{?}, {"_axis_0", ?}]>]>, composite_attributes = {approximate = "tanh"}, decomposition = @tenstorrent.gelu_tanh.impl} : (tensor<32x32xf32>) -> tensor<32x32xf32>
    return %2 : tensor<32x32xf32>
  }
  func.func private @tenstorrent.gelu_tanh.impl(%arg0: tensor<32x32xf32> {ttcore.shard_status = #ttcore.shard_status<unsharded>, sdy.sharding = #sdy.sharding<@mesh, [{}, {"_axis_0"}]>}) -> (tensor<32x32xf32> {ttcore.shard_status = #ttcore.shard_status<unsharded>}) {
    %cst = stablehlo.constant dense<5.000000e-01> : tensor<f32>
    %cst_0 = stablehlo.constant dense<1.000000e+00> : tensor<f32>
    %cst_1 = stablehlo.constant dense<0.797884583> : tensor<f32>
    %cst_2 = stablehlo.constant dense<4.471500e-02> : tensor<f32>
    %cst_3 = stablehlo.constant dense<3.000000e+00> : tensor<f32>
    %0 = stablehlo.broadcast_in_dim %cst_3, dims = [] : (tensor<f32>) -> tensor<32x32xf32>
    %1 = stablehlo.broadcast_in_dim %cst_2, dims = [] : (tensor<f32>) -> tensor<32x32xf32>
    %2 = stablehlo.broadcast_in_dim %cst_1, dims = [] : (tensor<f32>) -> tensor<32x32xf32>
    %3 = stablehlo.broadcast_in_dim %cst_0, dims = [] : (tensor<f32>) -> tensor<32x32xf32>
    %4 = stablehlo.broadcast_in_dim %cst, dims = [] : (tensor<f32>) -> tensor<32x32xf32>
    %5 = stablehlo.multiply %arg0, %4 : tensor<32x32xf32>
    %6 = stablehlo.power %arg0, %0 : tensor<32x32xf32>
    %7 = stablehlo.multiply %6, %1 : tensor<32x32xf32>
    %8 = stablehlo.add %arg0, %7 : tensor<32x32xf32>
    %9 = stablehlo.multiply %8, %2 : tensor<32x32xf32>
    %10 = stablehlo.tanh %9 : tensor<32x32xf32>
    %11 = stablehlo.add %10, %3 : tensor<32x32xf32>
    %12 = stablehlo.multiply %5, %11 : tensor<32x32xf32>
    return %12 : tensor<32x32xf32>
  }
}
```

after running propagation, you get this

```
module @FlattenCompositeGrouping attributes {mhlo.cross_program_prefetches = [], mhlo.input_output_alias = [], mhlo.is_dynamic = false, mhlo.use_auto_spmd_partitioning = false} {
  sdy.mesh @mesh = <["_axis_0_updated"=1, "_axis_0"=2]>
  func.func @main(%arg0: tensor<32x32xf32> {sdy.sharding = #sdy.sharding<@mesh, [{}, {"_axis_0"}]>, ttcore.shard_status = #ttcore.shard_status<presharded>}) -> (tensor<32x32xf32> {sdy.sharding = #sdy.sharding<@mesh, [{?}, {"_axis_0", ?}]>, ttcore.shard_status = #ttcore.shard_status<unsharded>}) {
    %0 = stablehlo.reshape %arg0 {sdy.sharding = #sdy.sharding_per_value<[<@mesh, [{?}, {?}, {"_axis_0", ?}]>]>} : (tensor<32x32xf32>) -> tensor<1x32x32xf32>
    %1 = stablehlo.reshape %0 {sdy.sharding = #sdy.sharding_per_value<[<@mesh, [{?}, {"_axis_0", ?}]>]>} : (tensor<1x32x32xf32>) -> tensor<32x32xf32>
    %2 = stablehlo.composite "tenstorrent.gelu_tanh" %1 {composite_attributes = {approximate = "tanh"}, decomposition = @tenstorrent.gelu_tanh.impl, sdy.sharding = #sdy.sharding_per_value<[<@mesh, [{?}, {"_axis_0", ?}]>]>} : (tensor<32x32xf32>) -> tensor<32x32xf32>
    return %2 : tensor<32x32xf32>
  }
  func.func private @tenstorrent.gelu_tanh.impl(%arg0: tensor<32x32xf32> {sdy.sharding = #sdy.sharding<@mesh, [{}, {"_axis_0"}]>, ttcore.shard_status = #ttcore.shard_status<unsharded>}) -> (tensor<32x32xf32> {sdy.sharding = #sdy.sharding<@mesh, [{?}, {"_axis_0", ?}]>, ttcore.shard_status = #ttcore.shard_status<unsharded>}) {
    %cst = stablehlo.constant dense<5.000000e-01> : tensor<f32>
    %cst_0 = stablehlo.constant dense<1.000000e+00> : tensor<f32>
    %cst_1 = stablehlo.constant dense<0.797884583> : tensor<f32>
    %cst_2 = stablehlo.constant dense<4.471500e-02> : tensor<f32>
    %cst_3 = stablehlo.constant dense<3.000000e+00> : tensor<f32>
    %0 = stablehlo.broadcast_in_dim %cst_3, dims = [] {sdy.sharding = #sdy.sharding_per_value<[<@mesh, [{?}, {"_axis_0", ?}]>]>} : (tensor<f32>) -> tensor<32x32xf32>
    %1 = stablehlo.broadcast_in_dim %cst_2, dims = [] {sdy.sharding = #sdy.sharding_per_value<[<@mesh, [{?}, {"_axis_0", ?}]>]>} : (tensor<f32>) -> tensor<32x32xf32>
    %2 = stablehlo.broadcast_in_dim %cst_1, dims = [] {sdy.sharding = #sdy.sharding_per_value<[<@mesh, [{?}, {"_axis_0", ?}]>]>} : (tensor<f32>) -> tensor<32x32xf32>
    %3 = stablehlo.broadcast_in_dim %cst_0, dims = [] {sdy.sharding = #sdy.sharding_per_value<[<@mesh, [{?}, {"_axis_0", ?}]>]>} : (tensor<f32>) -> tensor<32x32xf32>
    %4 = stablehlo.broadcast_in_dim %cst, dims = [] {sdy.sharding = #sdy.sharding_per_value<[<@mesh, [{?}, {"_axis_0", ?}]>]>} : (tensor<f32>) -> tensor<32x32xf32>
    %5 = stablehlo.multiply %arg0, %4 {sdy.sharding = #sdy.sharding_per_value<[<@mesh, [{?}, {"_axis_0", ?}]>]>} : tensor<32x32xf32>
    %6 = stablehlo.power %arg0, %0 {sdy.sharding = #sdy.sharding_per_value<[<@mesh, [{?}, {"_axis_0", ?}]>]>} : tensor<32x32xf32>
    %7 = stablehlo.multiply %6, %1 {sdy.sharding = #sdy.sharding_per_value<[<@mesh, [{?}, {"_axis_0", ?}]>]>} : tensor<32x32xf32>
    %8 = stablehlo.add %arg0, %7 {sdy.sharding = #sdy.sharding_per_value<[<@mesh, [{?}, {"_axis_0", ?}]>]>} : tensor<32x32xf32>
    %9 = stablehlo.multiply %8, %2 {sdy.sharding = #sdy.sharding_per_value<[<@mesh, [{?}, {"_axis_0", ?}]>]>} : tensor<32x32xf32>
    %10 = stablehlo.tanh %9 {sdy.sharding = #sdy.sharding_per_value<[<@mesh, [{?}, {"_axis_0", ?}]>]>} : tensor<32x32xf32>
    %11 = stablehlo.add %10, %3 {sdy.sharding = #sdy.sharding_per_value<[<@mesh, [{?}, {"_axis_0", ?}]>]>} : tensor<32x32xf32>
    %12 = stablehlo.multiply %5, %11 {sdy.sharding = #sdy.sharding_per_value<[<@mesh, [{?}, {"_axis_0", ?}]>]>} : tensor<32x32xf32>
    return %12 : tensor<32x32xf32>
  }
}
```

and depending on the composite type, you can define an identity mapping or some other form.

sshonTT:
@tapspatel, To elaborate a bit on what we discussed offline — I initially considered implementing the approach you suggested. However, this would have required defining a separate sharding strategy for each composite op, which raised a few concerns. Specifically, I wasn’t sure how to maintain consistency in the composite op’s sharding spec, or how to properly apply our existing global-to-local shape logic in that setup.

I did try modifying the Shardy side so that sharding propagation could handle composite ops and propagate correctly into their functions. But during the global-to-local shape update, I ran into a mismatch between the op’s local input shape and the shape signature defined in the function itself.

Of course, one possible solution would be to recalculate and redefine the function’s signature based on the output sharding values, but I decided to rely on the native Shardy propagation mechanism instead. This approach seemed cleaner, as it avoids having to maintain composite-specific handling and still allows further passes to proceed even if re-outlining fails, as long as there are no other issues.

That said, I completely agree that this approach is not perfect, given the tight dependencies between passes and the need to maintain shared context.

Since you’ve approved the PR and I’ve already shared the graph that might cause potential issues, I think it would make sense to go ahead and merge this PR first, and then we can continue exploring a more robust solution together.

nsmithtt:
Does shardy work across function calls?  My thinking is we could just replace composites with a `func.call`, run shardy, then move them back to composites.  Or if @tapspatel is possible that'd be even better.

sshonTT:
@nsmithtt
It seems that there’s currently no sharding rule defined for [func.call](https://github.com/openxla/shardy/blob/main/shardy/dialect/sdy/transforms/propagation/op_sharding_rule_registry.cc#L232).
We might need to introduce a propagation rule for `func.call`, similar to what we’ve been discussing for `composite_call`.

I actually have a [branch](https://github.com/tenstorrent/tt-mlir/commit/7904cddf7d0383bb8f8fdd4f9f22eb446607a4c9) implementing Tap’s suggested approach. It works, but it’s a bit non-trivial — I had to manually set up the mesh and sharding information for the composite function arguments and then run propagation again. After that, we still face shape mismatches in the global-to-local pass, since StableHLO doesn’t yet have verification logic that’s aware of Shardy (I asked about this [here](https://github.com/openxla/stablehlo/issues/2872)). As a workaround, I could have forced the composite function signature to match manually calculated local shapes, though that feels somewhat error-prone.

Also considering what we want to achieve from the composite op — we aim to apply sharding to a single ttnn.composite op such as GELU or SDPA.
This means the composite op could be defined in a backward direction (ttnn → ttir → shlo → torch, etc.), so we don’t want to reorder or insert new ops inside the composite function.
However, depending on the sharding configuration, operations like all_reduce / all_gather can still be inserted (e.g., before %3 = stablehlo.dot_general in the example below).
```
module @SyncTensorsGraph.86 attributes {mhlo.cross_program_prefetches = [], mhlo.frontend_attributes = {xla.sdy.meshes = "{mesh = #sdy.mesh<[\22_axis_0\22=2]>}"}, mhlo.input_output_alias = [], mhlo.is_dynamic = false, mhlo.use_auto_spmd_partitioning = false} {
  func.func @main(%arg0: tensor<1x12x32x128xbf16> {mhlo.frontend_attributes = {xla.sdy.sharding = "#sdy.sharding<@mesh, [{}, {}, {}, {\22_axis_0\22}]>"}, mhlo.sharding = "{devices=[1,2]<=[2]}"}, %arg1: tensor<1x12x32x128xbf16> {mhlo.frontend_attributes = {xla.sdy.sharding = "#sdy.sharding<@mesh, [{}, {}, {}, {\22_axis_0\22}]>"}, mhlo.sharding = "{devices=[1,2]<=[2]}"}, %arg2: tensor<1x12x32x128xbf16> {mhlo.frontend_attributes = {xla.sdy.sharding = "#sdy.sharding<@mesh, [{}, {}, {}, {\22_axis_0\22}]>"}, mhlo.sharding = "{devices=[1,2]<=[2]}"}) -> tensor<1x12x32x128xbf16> {
    %0 = stablehlo.composite "tt.scaled_dot_product_attention" %arg2, %arg1, %arg0 {composite_attributes = {is_causal = "True", scale = 1.000000e+00 : f32}, decomposition = @tt.scaled_dot_product_attention.impl} : (tensor<1x12x32x128xbf16>, tensor<1x12x32x128xbf16>, tensor<1x12x32x128xbf16>) -> tensor<1x12x32x128xbf16>
    return %0 : tensor<1x12x32x128xbf16>
  }
  func.func private @tt.scaled_dot_product_attention.impl(%arg0: tensor<1x12x32x128xbf16>, %arg1: tensor<1x12x32x128xbf16>, %arg2: tensor<1x12x32x128xbf16>) -> tensor<1x12x32x128xbf16> {
    %0 = stablehlo.reshape %arg0 : (tensor<1x12x32x128xbf16>) -> tensor<12x32x128xbf16>
    %1 = stablehlo.transpose %arg1, dims = [0, 1, 3, 2] {result_layout = dense<[2, 3, 1, 0]> : tensor<4xindex>, xla_shape = "bf16[1,12,128,32]{2,3,1,0}"} : (tensor<1x12x32x128xbf16>) -> tensor<1x12x128x32xbf16>
    %2 = stablehlo.reshape %1 : (tensor<1x12x128x32xbf16>) -> tensor<12x128x32xbf16>
    %3 = stablehlo.dot_general %0, %2, batching_dims = [0] x [0], contracting_dims = [2] x [1] : (tensor<12x32x128xbf16>, tensor<12x128x32xbf16>) -> tensor<12x32x32xbf16>
```
This prevents the graph from being converted to a backend ttnn.composite op, unless we handle additional pattern-matching and legalization logic on the TTIR side, which would require extra work.
With this PR, however, there is semantic check logic, if such a situation occurs, it simply stops trying to form a composite op and leaves it flattened. Since all semantic information is still preserved as the original composite op, we can later reconstruct (reoutline) it as a proper composite op.

That brings me back to [this discussion](https://github.com/tenstorrent/tt-mlir/pull/5729#issuecomment-3513853163).
While I understand the approach isn’t ideal, I think it still has its own merits compared to implementing sharding propagation directly through function calls.

lib/Dialect/StableHLO/Transforms/FlattenComposite.cpp
```diff
@@ -0,0 +1,186 @@
...
+
+  if (!callee) {
+    comp.emitOpError() << "failed to resolve callee function '" << leaf
+                       << "' referenced by attribute '"
+                       << sharding_utils::kDecompositionAttr << "' (full ref: '"
+                       << decompAttr
+                       << "'). Please ensure the symbol is defined and visible "
+                          "to the symbol table.";
+    return mlir::failure();
+  } else if (callee.getBody().empty()) {
```
vmilosevic: :warning: **llvm\-else\-after\-return** :warning:
```suggestion
  } if (callee.getBody().empty()) {
```

AleksKnezevic:
The backward definition of the composite op is an excellent point @sshonTT. Given this approach allows us to:

1. add composites from FE without needing to define individual shard rules
2. retain composite if there are no introduced CCLs and equally important
3. break the composite if a CCL is introduced, without needing individual validation of each composite downstream

I propose we proceed, even if it's a little unorthodox.

test/ttmlir/Conversion/StableHLOToTTIR/composite/test_sharding_composite.mlir
sshonTT: Hi @hshahTT, could you review once again?

nsmithtt:
@sshonTT that makes sense, thank you for the comprehensive comment!  Definitely seems like it's a tricky problem.  I wonder, is this something we could eventually upstream?

sshonTT:
@nsmithtt Thanks for the comment! Just to clarify, which upstream are you referring to?

nsmithtt:
> @nsmithtt Thanks for the comment! Just to clarify, which upstream are you referring to?

I suppose this would go to stablehlo repo (if that's where the shardy interfaces are defined).  This seems like a common pass that stablehlo could use.

sshonTT:
> > @nsmithtt Thanks for the comment! Just to clarify, which upstream are you referring to?
> 
> I suppose this would go to stablehlo repo (if that's where the shardy interfaces are defined). This seems like a common pass that stablehlo could use.

Oh, I hadn’t really considered that direction. As far as I know, StableHLO doesn’t currently use any Shardy interfaces.
In the current PR, there’s only one place that actually depends on Shardy information ([here](https://github.com/tenstorrent/tt-mlir/pull/5729/files#diff-f4f1d88f09a1146b3053fd696160c216ece6dbcd98ea38d7ec2c80c4164ad989R162)), and if we remove that, it might be possible to upstream the passes themselves.
That said, the pass is written under the assumption that propagation is handled via Shardy, so the key question would be whether StableHLO could recognize and integrate that Shardy logic internally.

lib/Dialect/StableHLO/Transforms/FlattenComposite.cpp
```diff
@@ -0,0 +1,187 @@
...
+    llvm::raw_string_ostream os(groupName);
+    os << "composite_" << callee.getSymName();
+  }
+  mlir::StringAttr groupAttr = builder.getStringAttr(groupName);
+  mlir::UnitAttr seedAttr = builder.getUnitAttr();
+
+  // 3) Prepare mapping from callee arguments to composite operands.
+  mlir::IRMapping mapping;
+  auto &calleeEntry = callee.getBody().front();
+  for (int64_t i = 0; i < static_cast<int64_t>(comp->getNumOperands()); ++i) {
```
hshahTT, line 90: Could you add an assert just before the for loop to make sure that `calleeEntry.getNumArguments() == comp->getNumOperands()`?

lib/Dialect/StableHLO/Transforms/FlattenComposite.cpp
```diff
@@ -0,0 +1,187 @@
...
+  // 5) Handle callee terminator: replace composite results with mapped return
+  // values. We expect a func.return with N operands, matching the composite's
+  // results.
+  auto ret = llvm::dyn_cast<mlir::func::ReturnOp>(
+      callee.getBody().front().getTerminator());
+  if (!ret) {
+    comp.emitOpError() << "expected callee function '" << leaf
+                       << "' to end with a 'return' operation.";
+    return mlir::failure();
+  }
```
hshahTT, line 129: Can you also assert that the number of operands for the return op is the same as the number of results for the composite op? You mentioned it in your comment but didn't add a check for it

lib/Dialect/StableHLO/Transforms/FlattenComposite.cpp
sshonTT: Added an assert.

lib/Dialect/StableHLO/Transforms/FlattenComposite.cpp
sshonTT: Added an assert.

PR #5730, opened 2025-11-05
[d2m] work around two asserts in `ExplicitDatamovementForm` + flatbuffer assertion for memrefs of tiles. 

zoecarver:
### Ticket
N/A

### Problem description

  1. Fix `d2m.generic` verifier to skip yield check for explicit datamovement form

  2. Fix d2m-grid-selection pass to skip explicit datamovement form

  3. Add `ttcore::getScalarShape` utility and fix flatbuffer generation for TileType memrefs

### What's changed
  1. Verifier: Guard yield terminator checks with `!isExplicitDatamovementForm()`. In explicit datamovement form, users provide custom terminators in each region, so the standard yield requirement doesn't apply.
  2. `GridSelection` pass: Early-exit when `isExplicitDatamovementForm()` is `true`. These operations have empty `indexing_maps` by design, and users specify grids explicitly, so the pass shouldn't attempt optimization.
  3. Flatbuffer generation:
    - Added `ttcore::getScalarShape(MemRefType)` utility that unwraps TileType and computes scalar element shape
    - Replaced assertion with proper handling using `getOperandInnerElementType()` (existing utility) and new `getScalarShape()`

### DIFF
--- a/lib/Dialect/D2M/Transforms/GridSelection.cpp
+++ b/lib/Dialect/D2M/Transforms/GridSelection.cpp
@@ -653,6 +653,10 @@ public:
         d2m::utils::getSquareTargetGrid(targetGridShape);
 
     module.walk([&](d2m::GenericOp genericOp) {
+      // Skip explicit datamovement form - users manage grids manually
+      if (genericOp.isExplicitDatamovementForm()) {
+        return;
+      }
       assignGrids(genericOp, targetGridShape, targetSquareGridShape);
     });
   }
--- a/test/ttmlir/Dialect/D2M/Transforms/grid_selection_explicit_datamovement.mlir
+++ b/test/ttmlir/Dialect/D2M/Transforms/grid_selection_explicit_datamovement.mlir
@@ -1,0 +1,37 @@
+// RUN: ttmlir-opt --split-input-file --ttcore-register-device --d2m-grid-selection %s | FileCheck %s
+
+// Verify that d2m-grid-selection pass skips d2m.generic operations in
+// explicit datamovement form. These operations have empty indexing_maps
+// and users manage grids manually, so the pass should not attempt to
+// assign or optimize grids.
+
+#layout = #ttcore.metal_layout<logical_shape = 64x64, dim_alignments = 32x32, collapsed_intervals = dense<[[0, 1], [1, 2]]> : tensor<2x2xi64>, undef, l1>
+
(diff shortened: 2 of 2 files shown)


lib/Target/TTMetal/TTMetalToFlatbuffer.cpp
```diff
@@ -523,13 +523,16 @@ tensorValueToFlatbuffer(FlatbufferObjectCache &cache, Value value) {
...
-
-  assert(!mlir::isa<ttcore::DeviceLayoutInterface>(memref.getLayout()));
+  // Get scalar element type and shape (unwrapping TileType if present)
+  Type scalarElementType = ttcore::getOperandInnerElementType(value);
+  llvm::SmallVector<int64_t> scalarShape = ttcore::getScalarShape(memref);
   std::vector<int32_t> shape =
-      ttmlir::utils::castContainer<std::vector<int32_t>>(memref.getShape());
+      ttmlir::utils::castContainer<std::vector<int32_t>>(scalarShape);
+
+  ttcore::DataType dtype = ttcore::elementTypeToDataType(scalarElementType);
```
nsmithtt: I'm a little unsure what to do for this code.  I take it we're trying to handle the case where the main entry func has tiled inputs.  It feels like we might be missing flatbuffer representation for this, i.e. a tile shape on `binary.fbs::TensorDesc` or something. So `tensor<2x3x!tile<32x32, bfp8_b>>` ->
```fbs
TensorDesc {
  shape = {2, 3},
  tile_shape = {32, 32},
  ...
  layout....dtype = Bfp8_b,
}
```

nsmithtt: I think this piece shouldn't be needed for now, we shouldn't expect the function entry to have tiled inputs (at least not yet).   We can make this change later when we need support.

zoecarver: Dropped this commit now that we use scalar tensors for args. Thank you, Nick!

nsmithtt (approved):
Awesome, thanks!

vtangTT (approved):
pykernel changes lgtm

zoecarver:
Thank you both!!

zoecarver:
Sorry! Didn't mean to pull in the pykernel changes

Issue #4037, opened 2025-07-08
Split heads fusion operation for reshape + permute (transpose)


ddilbazTT:
We need a fusion pass that will live under `lib/Dialect/TTIR/Transforms/TTIRFusing.cpp` which corresponds to splitting heads.

The following examples are generated from llama attention layer TTIR MLIR.
```
# Split Heads for Query (Q):
%9 = "ttir.reshape"(%7, %8) <{shape = [1 : i32, 32 : i32, 3072 : i32]}> : (tensor<32x3072xbf16>, tensor<1x32x3072xbf16>) -> tensor<1x32x3072xbf16> loc(#loc85)

%11 = "ttir.reshape"(%9, %10) <{shape = [1 : i32, 32 : i32, 24 : i32, 128 : i32]}> : (tensor<1x32x3072xbf16>, tensor<1x32x24x128xbf16>) -> tensor<1x32x24x128xbf16> loc(#loc86)

%13 = "ttir.permute"(%11, %12) <{permutation = array<i64: 0, 2, 1, 3>}> : (tensor<1x32x24x128xbf16>, tensor<1x24x32x128xbf16>) -> tensor<1x24x32x128xbf16> loc(#loc87)
```

```
# Split Heads for Key (K)
%18 = "ttir.reshape"(%16, %17) <{shape = [1 : i32, 32 : i32, 1024 : i32]}> : (tensor<32x1024xbf16>, tensor<1x32x1024xbf16>) -> tensor<1x32x1024xbf16> loc(#loc90)

%20 = "ttir.reshape"(%18, %19) <{shape = [1 : i32, 32 : i32, 8 : i32, 128 : i32]}> : (tensor<1x32x1024xbf16>, tensor<1x32x8x128xbf16>) -> tensor<1x32x8x128xbf16> loc(#loc91)

%22 = "ttir.permute"(%20, %21) <{permutation = array<i64: 0, 2, 1, 3>}> : (tensor<1x32x8x128xbf16>, tensor<1x8x32x128xbf16>) -> tensor<1x8x32x128xbf16> loc(#loc92)
```

```
# Split Heads for Value (V)
%27 = "ttir.reshape"(%25, %26) <{shape = [1 : i32, 32 : i32, 1024 : i32]}> : (tensor<32x1024xbf16>, tensor<1x32x1024xbf16>) -> tensor<1x32x1024xbf16> loc(#loc95)

[...]

PR #5525, opened 2025-10-27
broadcast fold before fusion and fix matmul with bias

ddilbazTT:
### Ticket
#5366 
#5464

### Problem description
I had added a fusion pattern for matmul -> reshape -> add which does not work well with different bias shapes.
It originally would take in a graph like the following:
matmul_AB = matmul(A, B)
reshape_AB = reshape(matmul_AB)
output = add (reshape_AB, bias)

and change it to:
reshape_bias = reshape(bias)
linear_output = linear(A, B, reshape_bias)
output = reshape(linear_output)

However, this is a rigid way of doing things. It would be ideal if we could:
- fold broadcast into add op so we get the broadcasted shape beforehand. This is done by moving broadcast fold before ttir fusions.
- Not reshape bias and check if it would be compatible with expected linear output shape. This would remove the reshape before linear op.

So, the new way of fusing would be:
- check if bias is broadcast compatible with matmul output shape
- check if expected output shape from linear op has the same volume as add op output shape
linear_output = linear(A, B, bias)
output = reshape(linear_output) 

### What's changed
- Moved broadcast folding before ttir fusion pass.
- Removed reshape before linear op.
- Assigned reshape output shape using expected broadcasted output shape, as bias might broadcast the output. 
- Added cases to test changes.
- Edited softmax fusion to handle sequences with/ without broadcast.
- Edited tests with broadcast in the sequence to use broadcast fold before ttir fusions.

### Checklist
- [X] New/Existing tests provide coverage for changes

PR #5740, opened 2025-11-06
Multi-Head Attention Fusion pass w/split_query_key_value_and_split_heads

ddilbazTT:
### Ticket
https://github.com/tenstorrent/tt-mlir/issues/4037 
Please note, this only covers Multi-Head Attention aspect of the fusion. There needs to be a follow-up PR for grouped query attention, which will reuse and be similar to multi-head attention pattern matching.

### Problem description
In multi-head attention, query/ key/ value go through a similar route:
Attention weight -> reshape 
Reshaped Attention weight, Q/K/V weight, (Q/K/V  Bias) -> matmul/linear -> reshape -> permute (x3) 

Key can be transposed (wrt Value and Query). 

So there are three matmul/linear, reshape and permute operations.

Instead, we can concatenate weight, (biases) and use split_query_key_value_and_split_heads op so there is only one matmul/linear op. This would also remove the reshapes and permutes following.

To be able to implement this with ttnn.split_query_key_value_and_split_heads op, we need to:
- concatenate weights
- (optional) concatenate biases
- run a single matmul/ linear op to replace the three matmul/ linear ops
- reshape matmul/ linear output to unsqueeze batch size and sequence length (this will be folded by reshape folder if identity reshape, which could happen based on bias shape broadcasting)
- run split_query_key_value_and_split_heads
The outputs of ttnn.split_query_key_value_and_split_heads would replace each permute op output, respectively.

attention weight -> reshape
Q/K/V weights -> concatenate
(Q/K/V  Biases -> concatenate)
Reshaped attention weight, concatenated QKV weight, (concatenated QKV bias) -> matmul/ linear -> split_query_key_value_and_split_heads 

### What's changed
Added split_query_key_value_and_split_heads fusion pass.

### DIFF
--- a/lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
+++ b/lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
@@ -2007,16 +2007,6 @@ public:
     int32_t batchSize = inputShape[I_BATCH_SIZE];
     int32_t sequenceLength = inputShape[I_SEQUENCE_LENGTH];
 
-    auto queryType = permuteOps[0].getType();
-    auto keyType = permuteOps[1].getType();
-    auto valueType = permuteOps[2].getType();
-
-    auto queryShape = queryType.getShape();
-    auto keyShape = keyType.getShape();
-    auto valueShape = valueType.getShape();
-    int32_t numHeads = queryType.getShape()[O_NUM_KV_HEADS];
-    int32_t hiddenSize = queryType.getShape()[O_HEAD_SIZE] * numHeads;
-
     // Concatenate weights along dimension determined by transposeB attribute.
 
     TypedValue<RankedTensorType> queryWeightMatrix = matmulOps[0].getB();
@@ -2025,6 +2015,11 @@ public:
 
     MatMulOpType queryMatmulOp = matmulOps[0];
     std::size_t dimToConcatWeights = queryMatmulOp.getTransposeB() ? 0 : 1;
+
+    // Assert that queryMatmulOp A and B have rank 2.
+    TT_assertv((queryMatmulOp.getA().getType().getRank() == 2 &&
+                queryMatmulOp.getB().getType().getRank() == 2),
+               "Expected rank 2 for MatMulOp operands A and B");
 
     ttir::ConcatOp concatenatedWeightMatrix = concatenateAlongDim(
         rewriter, valueWeightMatrix.getLoc(),
@@ -2077,6 +2072,16 @@ public:
     // Create reshape operation to convert from [batch*seq, hidden*3] to [batch,
     // seq, hidden*3]. If it is the same output shape, reshape folder will fold
     // this operation.
+    auto queryType = permuteOps[0].getType();
+    auto keyType = permuteOps[1].getType();
+    auto valueType = permuteOps[2].getType();
+
+    auto queryShape = queryType.getShape();
+    auto keyShape = keyType.getShape();
+    auto valueShape = valueType.getShape();
+    int32_t numHeads = queryType.getShape()[O_NUM_KV_HEADS];
+    int32_t hiddenSize = queryType.getShape()[O_HEAD_SIZE] * numHeads;
+
     SmallVector<int64_t> reshapeToSplitShape = {batchSize, sequenceLength,
                                                 hiddenSize * 3};
     auto reshapeElementType = queryMatmulOp.getType().getElementType();
@@ -2189,9 +2194,11 @@ private:
     llvm::SetVector<mlir::Operation *> sortedOps = topologicalSort(opsToMove);
 
     // Move only operations that appear after queryPermuteOp.
+
     for (mlir::Operation *op : sortedOps) {
-      if (op->getBlock() == queryPermuteOp->getBlock() &&
-          !op->isBeforeInBlock(queryPermuteOp)) {
+      TT_assertv(op->getBlock() == queryPermuteOp->getBlock(),
+                 "Expected all ops to be in the same block.");
+      if (!op->isBeforeInBlock(queryPermuteOp)) {
         op->moveBefore(queryPermuteOp);
       }
     }
@@ -2370,6 +2377,7 @@ private:
       return false;
     }
 
+    ArrayRef<int64_t> firstBShape = matrixOps[0].getB().getType().getShape();
     for (MatMulOpType matrixOp : matrixOps) {
       // Each op must have the same input as the reshape op output.
       if (matrixOp.getA() != reshapeOp.getResult()) {
@@ -2378,6 +2386,10 @@ private:
       auto bShape = matrixOp.getB().getType().getShape();
       // Each op must have a 2D weight matrix.
       if (bShape.size() != 2) {
+        return false;
+      }
+      // Check that B shape is the same for all ops.
+      if (bShape != firstBShape) {
         return false;
       }
     }
@@ -2392,6 +2404,23 @@ private:
     if ((matrixOps[0].getTransposeB() != matrixOps[1].getTransposeB()) ||
         (matrixOps[0].getTransposeB() != matrixOps[2].getTransposeB())) {
       return false;
+    }
+
+    // If matrix op is Linear Op, check that bias is present and has the same
+    // shape.
+    if constexpr (std::is_same_v<LinearOp, MatMulOpType>) {
+      TypedValue<RankedTensorType> queryBias = matrixOps[0].getBias();
+      TypedValue<RankedTensorType> keyBias = matrixOps[1].getBias();
+      TypedValue<RankedTensorType> valueBias = matrixOps[2].getBias();
+      if (!queryBias || !keyBias || !valueBias) {
+        return false;
+      }
+      auto queryBiasShape = queryBias.getType().getShape();
+      auto keyBiasShape = keyBias.getType().getShape();
+      auto valueBiasShape = valueBias.getType().getShape();
+      if (queryBiasShape != keyBiasShape || queryBiasShape != valueBiasShape) {
+        return false;
+      }
     }
 
     return true;


lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
```diff
@@ -1924,6 +1973,577 @@ class ConcatenateHeadsUpdatePattern : public mlir::OpRewritePattern<ReshapeOp> {
...
+
+    // Create concat op along given dimension.
+    return ttir::utils::createDPSOp<ttir::ConcatOp>(
+        rewriter, loc, concatenatedType, tensors,
+        rewriter.getSI32IntegerAttr(dim) /* axis */);
+  }
+
+  // Recursively collect all defining ops of a value and their dependencies.
+  // Avoid cycles or duplicates using a visited set.
+  void collectRecursivePreprocessingOps(Value val, Operation *QueryPermuteOp,
```
jameszianxuTT: This function seems specialized to a PermuteOp, is there a reason to include it as an Operation* rather than a PermuteOp*?

jameszianxuTT (approved):
Read over this changeset, I think it is structured reasonably and clearly. 

1. I wonder if your nice ascii diagrams in the PR desc will get messed up by the automatic  git commit message character-per-line limit. Maybe they could be embedded into the source somehow?
2. More comments in private helper functions describing what they do might make this more readable. Like that `populateMatmulOps` goes through consumers of the reshape's result in a sorted order (and the purpose of sorting?) 
3. It seems like concatenateAlongDim might be a useful general helper utility (?) rather than scoped to just this fuser class? Except for condition `// Verify all tensors have the same rank and shape except last dim` seems overly specialized, but if it works for your case, then also maybe there's no need to overcomplicate

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
ddilbazTT: Previously, I was hoisting the query linear op instead of query permute op. Operation* made it easier to switch, but in current context it is a permute op as you said. As a follow-up, I might add it as a utility since it would help others with ssa dominance issues. 
Edit: I will keep it as Operation since I need it for `getBlock` and `isBeforeInBlock`

ddilbazTT:
> Read over this changeset, I think it is structured reasonably and clearly.
> 
> 1. I wonder if your nice ascii diagrams in the PR desc will get messed up by the automatic  git commit message character-per-line limit. Maybe they could be embedded into the source somehow?
> 2. More comments in private helper functions describing what they do might make this more readable. Like that `populateMatmulOps` goes through consumers of the reshape's result in a sorted order (and the purpose of sorting?)
> 3. It seems like concatenateAlongDim might be a useful general helper utility (?) rather than scoped to just this fuser class? Except for condition `// Verify all tensors have the same rank and shape except last dim` seems overly specialized, but if it works for your case, then also maybe there's no need to overcomplicate

Great feedback, thanks! I will do a follow-up PR for porting some utilities introduced in this PR. I also appreciate the request for more code comments, and agree that I should move the diagrams to comments.

ddilbazTT:
This PR needed the following PRs merged:
- https://github.com/tenstorrent/tt-mlir/pull/5149
- https://github.com/tenstorrent/tt-mlir/pull/5525
- https://github.com/tenstorrent/tt-mlir/pull/5276

LPanosTT (approved):

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
```diff
@@ -1924,6 +1924,588 @@ class ConcatenateHeadsUpdatePattern : public mlir::OpRewritePattern<ReshapeOp> {
...
+    if (reshapeOutputShape[0] != reshapeInputShape[I_BATCH_SIZE] *
+                                     reshapeInputShape[I_SEQUENCE_LENGTH] ||
+        reshapeOutputShape[1] != reshapeInputShape[I_HIDDEN_DIMENSION]) {
+      return false;
+    }
+
+    // if size != 3, return false.
+    if (matrixOps.size() != 3) {
+      return false;
+    }
```
mmanzoorTT: I'll suggest to move this check at the beginning of this function to avoid the above computation in case of failure. Secondly, this comment doesn't provide any context/detail. Please update the comment and provide a reason for failure in case of `size!=3`. I am assuming that you are fusing query/key/value so the size must be 3 but this assumption can be incorrect.

odjuricicTT:
@ddilbazTT Thanks for adding this! Skimmed over the PR and it looks ok from a high level. @mtopalovicTT has much more context on this and should probably take a look.

Did you run this on a model e2e? We expect to find this pattern in llama right? Could you try https://github.com/tenstorrent/tt-xla/blob/main/examples/pytorch/llama.py

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
```diff
@@ -1924,6 +1924,588 @@ class ConcatenateHeadsUpdatePattern : public mlir::OpRewritePattern<ReshapeOp> {
...
+    collectFromOperands(permuteOps[2]);
+
+    // Hoist preprocessing ops before Query PermuteOp — all of which are after
+    // Query PermuteOp.
+    for (Operation *preOp : preOpsToMove) {
+      preOp->moveBefore(queryPermuteOp);
+    }
+  }
+
+  PermuteOp getPermuteOp(mlir::Operation *op) const {
```
mtopalovicTT: This doesn't need to be a generic function. You can constrain it to OpType and just do op.getResult()

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
```diff
@@ -1924,6 +1924,588 @@ class ConcatenateHeadsUpdatePattern : public mlir::OpRewritePattern<ReshapeOp> {
...
+  }
+
+  bool validatePermuteOps(ReshapeOp reshapeOp,
+                          llvm::SmallVector<PermuteOp> &permuteOps) const {
+    // Size must be 3 for Q, K, V.
+    if (permuteOps.size() != 3) {
+      return false;
+    }
+
+    Value queryTensor = permuteOps[0]->getResult(0);
```
mtopalovicTT:
```
TypedValue<RankedTensorType> queryTensor = permuteOps[0].getResult();
...
ArrayRef<int64_t> queryShape = queryTensor.getType().getShape();
```

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
```diff
@@ -1924,6 +1924,588 @@ class ConcatenateHeadsUpdatePattern : public mlir::OpRewritePattern<ReshapeOp> {
...
+      return false;
+    }
+
+    if (queryShape[O_NUM_KV_HEADS] != keyShape[O_NUM_KV_HEADS] ||
+        queryShape[O_NUM_KV_HEADS] != valueShape[O_NUM_KV_HEADS]) {
+      // Number of kv heads must match.
+      return false;
+    }
+
+    if (queryShape[O_SEQUENCE_LENGTH] != valueShape[O_SEQUENCE_LENGTH]) {
```
mtopalovicTT: seq len for key doesn't have to be the same? If they are not the same then you cannot concatenate right? Or you can concatenate but you need to pad the lesser dims seq len dims before calling split qkv heads op and then slice afterwards

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
```diff
@@ -1924,6 +1924,588 @@ class ConcatenateHeadsUpdatePattern : public mlir::OpRewritePattern<ReshapeOp> {
...
+              });
+  }
+
+  bool isKeyTransposed(ArrayRef<int64_t> keyShape,
+                       ArrayRef<int64_t> queryShape) const {
+    return (keyShape[K_HEAD_SIZE] == queryShape[O_HEAD_SIZE] &&
+            keyShape[K_SEQUENCE_LENGTH] == queryShape[O_SEQUENCE_LENGTH]);
+  }
+
+  bool validatePermuteOps(ReshapeOp reshapeOp,
```
mtopalovicTT, line 2262: This is very important function with bunch of logic. Would be good if you can leave comments on the thought process.

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
```diff
@@ -1924,6 +1924,588 @@ class ConcatenateHeadsUpdatePattern : public mlir::OpRewritePattern<ReshapeOp> {
...
+  //
+  // Q/K/V Weights -----> Concat Weights \
+  //                                      \
+  // (opt)Q/K/V Biases -> Concat Biases --+--> Matmul/Linear --> Split QKV
+  //                                      /
+  // Attention ---------> Reshape -------/
+
+  // Reshape input is [batch_size, sequence_length, hidden_dimensions].
+  // For Multi-Head Attention:
+  // Query is [batch_size, number of kv heads, sequence_length, head_size].
```
mtopalovicTT, line 1954: query is num_heads and k/v is num_heads or num_kv heads depending on if num_heads is different for k and v right?

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
```diff
@@ -1924,6 +1924,588 @@ class ConcatenateHeadsUpdatePattern : public mlir::OpRewritePattern<ReshapeOp> {
...
+  // Key is [batch_size, number of kv heads, sequence_length, head_size].
+  // Value is [batch_size, number of kv heads, sequence_length, head_size].
+  // If key is transposed:
+  // Key is [batch_size, number of kv heads, head_size, sequence_length].
```
mtopalovicTT, line 1958: can you give IR example of how would fusing look like if key is transposed and if key is not transposed?

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
```diff
@@ -1924,6 +1924,588 @@ class ConcatenateHeadsUpdatePattern : public mlir::OpRewritePattern<ReshapeOp> {
...
+
+    // Create concat op along given dimension.
+    return ttir::utils::createDPSOp<ttir::ConcatOp>(
+        rewriter, loc, concatenatedType, tensors,
+        rewriter.getSI32IntegerAttr(dim) /* axis */);
+  }
+
+  // Recursively collect all defining ops of a value and their dependencies.
+  // Avoid cycles or duplicates using a visited set.
+  void collectRecursivePreprocessingOps(Value val, Operation *QueryPermuteOp,
```
mtopalovicTT: nit queryPermuteOp

mtopalovicTT: Not sure if you can use this helper instead to grab all operations recursively.

getUseDefChain
then filterOperations
and then topologicalSort

which will get you all the operations recursively topologically sorted for k/v. Then you can check if operation is before q in block.

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
```diff
@@ -1924,6 +1924,588 @@ class ConcatenateHeadsUpdatePattern : public mlir::OpRewritePattern<ReshapeOp> {
...
+
+    // Concatenate weights along dimension determined by transposeB attribute.
+
+    Value queryWeightMatrix =
+        (mhaType == TYPE_BIASED) ? linearOps[0].getB() : matmulOps[0].getB();
+    Value keyWeightMatrix =
+        (mhaType == TYPE_BIASED) ? linearOps[1].getB() : matmulOps[1].getB();
+    Value valueWeightMatrix =
+        (mhaType == TYPE_BIASED) ? linearOps[2].getB() : matmulOps[2].getB();
+    std::size_t queryWeightRank =
```
mtopalovicTT: queryWeightRank = queryWeightMatrix.getType().getRank()?

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
```diff
@@ -1924,6 +1924,588 @@ class ConcatenateHeadsUpdatePattern : public mlir::OpRewritePattern<ReshapeOp> {
...
+    int32_t hiddenSize = queryType.getShape()[O_HEAD_SIZE] * numHeads;
+
+    // Concatenate weights along dimension determined by transposeB attribute.
+
+    Value queryWeightMatrix =
+        (mhaType == TYPE_BIASED) ? linearOps[0].getB() : matmulOps[0].getB();
+    Value keyWeightMatrix =
+        (mhaType == TYPE_BIASED) ? linearOps[1].getB() : matmulOps[1].getB();
+    Value valueWeightMatrix =
+        (mhaType == TYPE_BIASED) ? linearOps[2].getB() : matmulOps[2].getB();
```
mtopalovicTT: I would add assert here to check that queryWeightMatrix keyWeightMatrix and valueWeightMatrix all have ranks of size 2

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
```diff
@@ -1924,6 +1924,588 @@ class ConcatenateHeadsUpdatePattern : public mlir::OpRewritePattern<ReshapeOp> {
...
+    std::size_t queryWeightRank =
+        mlir::cast<RankedTensorType>(queryWeightMatrix.getType())
+            .getShape()
+            .size();
+
+    std::size_t dimToConcatWeights =
+        (mhaType == TYPE_BIASED)
+            ? linearOps[0].getTransposeB()
+                  ? 0
+                  : (queryWeightRank - 1) // 0 if transposed, else last dim.
```
mtopalovicTT: This should be 1 per your invariant: 
      if (bShape.size() != 2) {
        return false;
      }

    if (reshapeOutputShape.size() != 2) {
      return false;
    }

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
```diff
@@ -1924,6 +1924,588 @@ class ConcatenateHeadsUpdatePattern : public mlir::OpRewritePattern<ReshapeOp> {
...
+
+      ttir::ConcatOp concatenatedBias = concatenateAlongDim(
+          rewriter, valueBias.getLoc(),
+          ValueRange{queryBias, keyBias, valueBias}, biasRank - 1);
+
+      // Get the output shape from one of the original linear ops.
+      ArrayRef<int64_t> originalLinearOutputShape =
+          linearOps[0].getResult().getType().getShape();
+
+      bool transposeA = linearOps[0].getTransposeA();
```
mtopalovicTT: same here I would asssert that transposeA is false

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
```diff
@@ -1924,6 +1924,588 @@ class ConcatenateHeadsUpdatePattern : public mlir::OpRewritePattern<ReshapeOp> {
...
+      std::size_t biasRank =
+          mlir::cast<RankedTensorType>(queryBias.getType()).getShape().size();
+
+      ttir::ConcatOp concatenatedBias = concatenateAlongDim(
+          rewriter, valueBias.getLoc(),
+          ValueRange{queryBias, keyBias, valueBias}, biasRank - 1);
+
+      // Get the output shape from one of the original linear ops.
+      ArrayRef<int64_t> originalLinearOutputShape =
+          linearOps[0].getResult().getType().getShape();
```
mtopalovicTT: can you extract this: linearOps[0] to same variable and you is below. Something like
LinearOp queryProjection = linearOps[0];

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
```diff
@@ -1924,6 +1924,588 @@ class ConcatenateHeadsUpdatePattern : public mlir::OpRewritePattern<ReshapeOp> {
     return true;
   }
 };
+
+class SplitQueryKeyValueAndSplitHeadsUpdatePattern
```
mtopalovicTT, line 1929: To me it seems that it would save us some code if we would template on MatMul type since majority of checks are the same for both types of matmuls. The only difference is when you need to create actual concatenated matmul type is where the difference is but you can you if constexpr (std::is_same<MatMulType, LinearOp>) and then add logic for concatenating bias.

Both matmul and linear builder accept typerange and value range arguments so you can create both ops in one go, and you won't need to constantly check the enum. Let me know if this makes sense I can help you with implementing the code, should be much of work.

ddilbazTT:
> @ddilbazTT Thanks for adding this! Skimmed over the PR and it looks ok from a high level. @mtopalovicTT has much more context on this and should probably take a look.
> 
> Did you run this on a model e2e? We expect to find this pattern in llama right? Could you try https://github.com/tenstorrent/tt-xla/blob/main/examples/pytorch/llama.py

That would be grouped query attention so it wouldn't be handled in this PR. However, some older llama versions use multi-head attention and are targeted with this PR.

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
ddilbazTT: Yes, I didn't add the distinction currently since we're doing mha. But I will add the distinction for GQA.

mtopalovicTT:
We should also add silicon tests for this fusing.

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
ddilbazTT: Based on our offline discussion, we are not using queryWeightRank anymore anyways.

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
ddilbazTT: That's already checked in `validateMatmulOrLinearOps`

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
ddilbazTT: That's checked in `validateMatmulOrLinearOps`

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
ddilbazTT: That's checked in `validateMatmulOrLinearOps`

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
ddilbazTT: Thanks for this suggestion and pointers on how to do this. I agree with you that this fusion pattern should be templated.

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
ddilbazTT: Is this what you are suggesting?

```
template<typename OpType>
PermuteOp getPermuteOp(OpType op) const {
  // Permute op comes from: linear/ matmul op -> reshape op -> permute op.
  auto users = op.getResult().getUsers();  // Changed from getResult(0)
  if (llvm::range_size(users) != 1) {
    return nullptr;
  }
  ReshapeOp reshapeOp = mlir::dyn_cast<ReshapeOp>(*users.begin());
  if (!reshapeOp) {
    return nullptr;
  }
  auto reshapeUsers = reshapeOp.getResult().getUsers();
  if (llvm::range_size(reshapeUsers) != 1) {
    return nullptr;
  }
  PermuteOp permuteOp = mlir::dyn_cast<PermuteOp>(*reshapeUsers.begin());
  if (!permuteOp) {
    return nullptr;
  }
  return permuteOp;
}
```

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
```diff
@@ -1924,6 +1924,485 @@ class ConcatenateHeadsUpdatePattern : public mlir::OpRewritePattern<ReshapeOp> {
...
+    linearOutputShape.back() = originalLinearOutputShape.back() * 3;
+
+    // Create linear operation with concatenated weights and bias.
+    RankedTensorType linearOutputType = RankedTensorType::get(
+        linearOutputShape, queryMatmulOp.getType().getElementType());
+
+    Value matmulDpsOutput = rewriter.create<EmptyOp>(
+        queryMatmulOp.getLoc(), linearOutputType.getShape(),
+        linearOutputType.getElementType(), linearOutputType.getEncoding());
+
```
vmilosevic, line 2057: :warning: **google\-readability\-todo** :warning:
missing username/bug in TODO
```suggestion
    // TODO(nsmith): Extend to support Grouped Query Attention.
```

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
mtopalovicTT: Now you can use MatMulOpType instead of Operation* op here. You don't need additional template.

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
ddilbazTT: There are example tests for transposed/ not transposed. Is that what you meant? Or did you mean that I should add a comment here?

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
ddilbazTT: Done! Would appreciate if you could take a look again.

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
```diff
@@ -1924,6 +1924,479 @@ class ConcatenateHeadsUpdatePattern : public mlir::OpRewritePattern<ReshapeOp> {
...
+    // Filter to get operations.
+    llvm::SetVector<mlir::Operation *> opsToMove =
+        ttmlir::utils::filterOperations(allValues.getArrayRef());
+
+    // Topologically sort to maintain dependencies.
+    llvm::SetVector<mlir::Operation *> sortedOps = topologicalSort(opsToMove);
+
+    // Move only operations that appear after queryPermuteOp.
+    for (mlir::Operation *op : sortedOps) {
+      if (op->getBlock() == queryPermuteOp->getBlock() &&
```
mtopalovicTT: I would assert that op->getBlock() == queryPermuteOp->getBlock()

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
```diff
@@ -1924,6 +1924,479 @@ class ConcatenateHeadsUpdatePattern : public mlir::OpRewritePattern<ReshapeOp> {
...
+    int32_t numHeads = queryType.getShape()[O_NUM_KV_HEADS];
+    int32_t hiddenSize = queryType.getShape()[O_HEAD_SIZE] * numHeads;
+
+    // Concatenate weights along dimension determined by transposeB attribute.
+
+    TypedValue<RankedTensorType> queryWeightMatrix = matmulOps[0].getB();
+    TypedValue<RankedTensorType> keyWeightMatrix = matmulOps[1].getB();
+    TypedValue<RankedTensorType> valueWeightMatrix = matmulOps[2].getB();
+
+    MatMulOpType queryMatmulOp = matmulOps[0];
```
mtopalovicTT, line 2016: assert that queryMatmulOp A and B rank are of size 2

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
```diff
@@ -1924,6 +1924,479 @@ class ConcatenateHeadsUpdatePattern : public mlir::OpRewritePattern<ReshapeOp> {
...
+    if (reshapeOutputShape.size() != 2) {
+      return false;
+    }
+    if (reshapeOutputShape[0] != reshapeInputShape[I_BATCH_SIZE] *
+                                     reshapeInputShape[I_SEQUENCE_LENGTH] ||
+        reshapeOutputShape[1] != reshapeInputShape[I_HIDDEN_DIMENSION]) {
+      return false;
+    }
+
+    for (MatMulOpType matrixOp : matrixOps) {
```
mtopalovicTT, line 2381: Should we validate that all shapes match for all 3 matmul operations?

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
```diff
@@ -1924,6 +1924,479 @@ class ConcatenateHeadsUpdatePattern : public mlir::OpRewritePattern<ReshapeOp> {
...
+    std::size_t dimToConcatWeights = queryMatmulOp.getTransposeB() ? 0 : 1;
+
+    ttir::ConcatOp concatenatedWeightMatrix = concatenateAlongDim(
+        rewriter, valueWeightMatrix.getLoc(),
+        ValueRange{queryWeightMatrix, keyWeightMatrix, valueWeightMatrix},
+        dimToConcatWeights);
+
+    Value concatenatedBias = nullptr;
+    if constexpr (std::is_same_v<LinearOp, MatMulOpType>) {
+      Value queryBias = matmulOps[0].getBias();
```
mtopalovicTT, line 2031: Same for bias we should validate that all 3 biasses have exact shape.

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
```diff
@@ -1924,6 +1924,479 @@ class ConcatenateHeadsUpdatePattern : public mlir::OpRewritePattern<ReshapeOp> {
...
+    int32_t sequenceLength = inputShape[I_SEQUENCE_LENGTH];
+
+    auto queryType = permuteOps[0].getType();
+    auto keyType = permuteOps[1].getType();
+    auto valueType = permuteOps[2].getType();
+
+    auto queryShape = queryType.getShape();
+    auto keyShape = keyType.getShape();
+    auto valueShape = valueType.getShape();
+    int32_t numHeads = queryType.getShape()[O_NUM_KV_HEADS];
```
mtopalovicTT: These are not used before you create reshape. You can move them below.

mtopalovicTT (approved):

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
ddilbazTT: I check this in `validateMatmulOrLinearOps`, and I'll assert here since it shouldn't reach this point.

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
ddilbazTT: We check that the input shape matches by checking against reshape output. Checking for output shape would differ between MHA and GQA. I'll add the check here for now, but for GQA this might be moved around.

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
ddilbazTT: Added to `validateMatmulOrLinearOps`

PR #5746, opened 2025-11-07
Fix ToLayoutOp folding to preserve DRAM staging when ops are in between

rpavlovicTT:
## Problem description

This commit fixes a bug in the ToLayoutOp consecutive folding optimization that was incorrectly eliminating DRAM staging operations, potentially causing memory allocation issues.

The `foldConsecutiveToLayoutOp` function was merging consecutive ToLayoutOps without considering whether there were operations between them that depended on DRAM staging for memory pressure management.

Before fix (incorrect folding):
```
    relu -> %148 (L1 block-sharded)
      |-> to_layout -> %149 (DRAM interleaved)
      |    [other ops that were scheduled assuming DRAM]
      |    '-> to_layout -> %153 (L1 block-sharded)
      |         '-> conv2d_2
      '-> to_layout -> %150 (L1 width-sharded)
           '-> conv2d_1
```

After canonicalization, %149 was incorrectly folded away:
```
    relu -> %148 (L1 block-sharded)
      |-> to_layout -> %153 (L1 block-sharded) [X] WRONG
      |    '-> conv2d_2
      '-> to_layout -> %150 (L1 width-sharded)
           '-> conv2d_1
```

This breaks memory staging because the ops in between were analyzed/scheduled assuming the tensor would be in DRAM. Folding bypasses DRAM and keeps the tensor in L1, causing memory pressure the optimizer didn't account for.

## Change 

Added a check in `foldConsecutiveToLayoutOp` to reject folding when:
1. There are operations between the producer and consumer ToLayoutOps, AND
2. Producer moves to DRAM buffer type, AND
3. Consumer moves to L1 buffer type

Piggy-backed fix in Optimizer:
- Try skipping spill to DRAM on the *nearest* forked branch in the schedule.

### Testing

Tested on full resnet50, where we end up with 4 more spill to dram (to_memory_config) ops. Perf dropped by ~10 fps measured by `ttrt` benchmarking.

### DIFF
--- a/lib/Dialect/TTNN/IR/TTNNOps.cpp
+++ b/lib/Dialect/TTNN/IR/TTNNOps.cpp
@@ -1834,6 +1834,28 @@ mlir::OpFoldResult foldConsecutiveToLayoutOp(ttnn::ToLayoutOp op) {
     return nullptr;
   }
 
+  // Don't fold if there are operations between producer and consumer when:
+  // - Producer moves to DRAM buffer type, AND
+  // - Consumer moves to L1 buffer type
+  //
+  // This preserves intentional DRAM staging for memory allocation safety.
+  // The ops in between may have been analyzed/scheduled assuming the tensor
+  // goes through DRAM. Folding would bypass DRAM and keep the tensor in L1,
+  // potentially causing memory pressure that wasn't accounted for.
+  Operation *nextOp = producerOp->getNextNode();
+  bool hasOpsBetween = (nextOp != op.getOperation());
+
+  if (hasOpsBetween) {
+    MemoryConfigAttr producerMemConfig = producerOp.getMemoryConfigAttr();
+    MemoryConfigAttr consumerMemConfig = op.getMemoryConfigAttr();
+
+    if (producerMemConfig && consumerMemConfig &&
+        producerMemConfig.getBufferType().getValue() == BufferType::DRAM &&
+        consumerMemConfig.getBufferType().getValue() == BufferType::L1) {
+      return nullptr;
+    }
+  }
+
   if (!op.getDtype()) {
     op.setDtypeAttr(producerOp.getDtypeAttr());
   }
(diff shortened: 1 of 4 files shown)


rpavlovicTT:
https://github.com/tenstorrent/tt-xla/actions/runs/19170832241/job/54817445835

Perf benchmark run with this fix included, there is 1-2% perf drop across CNN models.

lib/Dialect/TTNN/IR/TTNNOps.cpp
```diff
@@ -1821,6 +1821,28 @@ mlir::OpFoldResult foldConsecutiveToLayoutOp(ttnn::ToLayoutOp op) {
     return nullptr;
   }
 
+  // Don't fold if there are operations between producer and consumer when:
```
mtopalovicTT, line 1837: Hmmm. Shouldn't we always not fold if producer has more then one user?

rpavlovicTT: In my problem case, the producer has one user. 

I was thinking about forked producer generally, but didn't want to forbid it. Reason is we fold consumer toLayout op and connect its input to producer's input, while producer op still remains alive in the IR. This looks valid to me.
Something like 

```
input → producer → consumerA
             ↘
              → consumerB* (folding op)

input → producer → consumerA
    ↘
      → consumerB' (folded)
```

sdjordjevicTT: I am not sure that I understand this as well. So this is your example that we are trying to solve?
```
|-> to_layout -> %149 (DRAM interleaved)
      |    [other ops that were scheduled assuming DRAM]
      |    '-> to_layout -> %153 (L1 block-sharded)
      |         '-> conv2d_2
```
      
In this example, [other ops that were scheduled assuming DRAM] aren't consumers of %149?|

rpavlovicTT: Correct, they aren't consumers of that tensor. But they expect that tensor %149 is placed in DRAM.

odjuricicTT (approved):

PR #5370, opened 2025-10-16
Changing name generation for EmitPy

djstefanovicTT:
### Ticket
https://github.com/tenstorrent/tt-mlir/issues/4934

### Problem description
EmitPy uses `v1, v2, ...` for all variable names that are generated.

### What's changed
Changed naming convention to use function names. Now generates `matmul_1, add_1, ...`. It follows the existing scope indentation as before, but instead of fixed v it uses name of the function it is used in.

Issue #4934, opened 2025-09-15
[EmitPy] Name variables in generated code


svuckovicTT:
Today:
Codegen'd variables are named `v0`, `v1`, ...

Tomorrow:
Codegen'd variables names make sense. For example, they can be named `matmul0`, `matmul1`, `add0`, etc. Anything better than that is a bonus.

PR #5747, opened 2025-11-07
Changing name generation for EmitPy

djstefanovicTT:
### Ticket
https://github.com/tenstorrent/tt-mlir/issues/4934

### Problem description
EmitPy uses `v1, v2, ...` for all variable names that are generated.

### What's changed
Changed naming convention to use function names. Now generates `matmul_1, add_1, ...`. It follows the existing scope indentation as before, but instead of fixed v it uses name of the function it is used in.

Due to mistakes in merging with main, new PR had to be made. Old one: https://github.com/tenstorrent/tt-mlir/pull/5370

### DIFF
--- a/test/ttmlir/Translate/EmitPy/global.mlir
+++ b/test/ttmlir/Translate/EmitPy/global.mlir
@@ -15,9 +15,9 @@ module {
   emitpy.global @global_var_1 = #emitpy.opaque<"None">
 
   func.func @set_global_1(%arg0 : !emitpy.opaque<"[ttnn.Tensor]">) -> () {
-    // CHECK: def set_global_1(inputs):
+    // CHECK: def set_global_1([[INS:.*]]):
     // CHECK: global global_var_1
-    // CHECK: global_var_1 = inputs
+    // CHECK: global_var_1 = [[INS]]
     %0 = emitpy.global_statement @global_var_1 : !emitpy.opaque<"[ttnn.Tensor]">
     emitpy.assign_global @global_var_1 = %arg0 : !emitpy.opaque<"[ttnn.Tensor]">
     return
@@ -33,7 +33,7 @@ module {
   func.func @set_global_2() -> () {
     // CHECK: def set_global_2():
     // CHECK: global global_var_2
-    // CHECK: global_var_2 = var_0
+    // CHECK: global_var_2 = [[VAR:.*]]
     %0 = emitpy.global_statement @global_var_2 : i64
     %1 = "emitpy.constant"() <{value = 110 : i64}> : () -> i64
     emitpy.assign_global @global_var_2 = %1 : i64


svuckovicTT:
I'm not sold on this approach tbh.

There's no mechanism to supply variable names from outside (except for callopaqueop). Some TT-NN specific things are leaked into the emitpy dialect:
- logic for naming func arg to "inputs", which is also brittle.
- names of variables are created by taking everything to the right of the last `.` character

I'm inclined to approve this as it's better than what we have today, but I do feel that it's the wrong direction. Having a pass that tacks on discardable attrs seems like a more flexible approach, and doesn't require leaking anything into emitpy.

Thoughts? @azecevicTT @sdjordjevicTT @djstefanovicTT

lib/Conversion/TTNNToEmitPy/EmitPyPostProcessing.cpp
```diff
@@ -0,0 +1,105 @@
...
+            std::to_string(inputCounter[calleeAttr.getValue()]++);
+        // Insert all results into valueNames for multi-result operations
+        for (unsigned i = 0; i < callOp.getNumResults(); ++i) {
+          std::string argName = callOp.getNumResults() > 1
+                                    ? baseName + "_" + std::to_string(i)
+                                    : baseName;
+          valueNames.insert({callOp.getResult(i), argName});
+        }
+        StringAttr suggestName = StringAttr::get(ctx, baseName);
+        callOp->setAttr("suggest_name", suggestName);
```
svuckovicTT: I'd suggest (heh) that we always prefix these with `emitpy` - makes it clear which dialect it belongs to. IR parser might also be reading these (I have to check) and we might be able to even capture these and run a verifier on them (lots of ifs though).

Additionally, I think shortening the string to `emitpy.name` would make IR more readable.

include/ttmlir/Conversion/Passes.td
```diff
@@ -198,6 +198,15 @@ def ConvertTTNNToEmitPy : Pass<"convert-ttnn-to-emitpy", "::mlir::ModuleOp"> {
   let dependentDialects = ["mlir::tt::emitpy::EmitPyDialect", "mlir::tt::ttnn::TTNNDialect"];
 }
 
+def EmitPyPostProcessing : Pass<"emitpy-post-processing", "::mlir::ModuleOp"> {
```
svuckovicTT: This name is vary vague as it doesn't indicate what kind of processing it's doing. It often pays to be as accurate as possible when naming passes, suggestion below:
```suggestion
def EmitPyNameVars : Pass<"emitpy-name-vars", "::mlir::ModuleOp"> {
```

test/ttmlir/EmitPy/TTNN/load_input/load_input.mlir
```diff
@@ -11,25 +11,17 @@
 // RUN: FileCheck %s --check-prefix=CUSTOM-FULL --input-file=%t.custom_full.mlir
```
svuckovicTT, line 12: Can you add a couple tests - something to cover the golden path + maybe a test to see what happens when there's collisions in variable names (in func args, callops, return names, etc.)

lib/Target/Python/TranslateToPython.cpp
```diff
@@ -159,7 +164,7 @@ std::string PythonEmitter::getSubscriptName(SubscriptOp op) {
   std::string name;
   llvm::raw_string_ostream ss(name);
   auto index = op.getIndex();
-  ss << "[" << getOrCreateName(index) << "]";
+  ss << "[" << getOrCreateName(index, "subscript") << "]";
```
svuckovicTT: I think it makes sense to revert the default behaviour to what it was before, using `v_{index}` for variables for which the names aren't supplied.

test/ttmlir/Translate/EmitPy/global.mlir
```diff
@@ -15,9 +15,9 @@ module {
   emitpy.global @global_var_1 = #emitpy.opaque<"None">
 
   func.func @set_global_1(%arg0 : !emitpy.opaque<"[ttnn.Tensor]">) -> () {
-    // CHECK: def set_global_1(v1):
+    // CHECK: def set_global_1(inputs):
     // CHECK: global global_var_1
-    // CHECK: global_var_1 = v1
+    // CHECK: global_var_1 = inputs
```
svuckovicTT: Other tests shouldn't rely on any specific naming logic.


```suggestion
 func.func @set_global_1(%arg0 : !emitpy.opaque<"[ttnn.Tensor]">) -> () {
    // CHECK: def set_global_1([[INS:.*]]):
// CHECK: global global_var_1
    // CHECK: global_var_1 = [[INS]]
```

PR #5754, opened 2025-11-07
Skip mesh_shard creation for scalar replicate outputs

sshonTT:
### Ticket
[Link to Github Issue](https://github.com/tenstorrent/tt-xla/issues/1993#issuecomment-3502650909)

### Problem description
We can skip generating mesh_shard op (shard-to-full) for scalar tensor with replicate sharding, and tt-metal doesn't handle this case yet so it raises error.

### What's changed
Avoid generating mesh_shard op (shard-to-full) when the result is a scalar tensor with replicate sharding, as this case is semantically no-op.
Now this skip logic is only for scalar value, I will find it could be completely safe for all replicate cases.

### Checklist

### DIFF
--- a/test/ttmlir/Conversion/StableHLOToTTIR/ccl/ccl_ops_shardy.mlir
+++ b/test/ttmlir/Conversion/StableHLOToTTIR/ccl/ccl_ops_shardy.mlir
@@ -290,11 +290,8 @@ module @sdy_manual_computation_constant {
 // CHECK-SAME: shard_direction = #ttcore.shard_direction<full_to_shard>
 // CHECK-SAME: shard_shape = array<i64: 1>
 // CHECK-SAME: shard_type = #ttcore.shard_type<replicate>
-// CHECK: "ttir.mesh_shard"
-// CHECK-SAME: shard_dims = array<i64: -1>
-// CHECK-SAME: shard_direction = #ttcore.shard_direction<shard_to_full>
-// CHECK-SAME: shard_shape = array<i64: 1>
-// CHECK-SAME: shard_type = #ttcore.shard_type<replicate>
+// CHECK: return
+// CHECK-SAME: tensor<f32>
 
 // -----
 
--- a/test/ttmlir/Conversion/StableHLOToTTIR/ccl/e2e_dp_shardy.mlir
+++ b/test/ttmlir/Conversion/StableHLOToTTIR/ccl/e2e_dp_shardy.mlir
@@ -157,8 +157,5 @@ module @jit_loss_dp attributes {mhlo.num_partitions = 8 : i32, mhlo.num_replicas
 // CHECK-SAME: shard_direction = #ttcore.shard_direction<full_to_shard>
 // CHECK-SAME: shard_shape = array<i64: 8, 1>
 // CHECK-SAME: shard_type = #ttcore.shard_type<identity>
-// CHECK: "ttir.mesh_shard"
-// CHECK-SAME: shard_dims = array<i64: -1>
-// CHECK-SAME: shard_direction = #ttcore.shard_direction<shard_to_full>
-// CHECK-SAME: shard_shape = array<i64: 1>
-// CHECK-SAME: shard_type = #ttcore.shard_type<replicate>
+// CHECK: return
+// CHECK-SAME: : tensor<f32>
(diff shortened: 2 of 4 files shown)


AleksKnezevic (approved):

wooseokTT:
@sshonTT @AleksKnezevic I believe this case only works when we are using tt-xla frontend where mesh_shard ops will be identity due to expecting pre-sharded buffers. If we are expecting single-device tensor, then we still need to pick one out of multiple buffers, right?

sshonTT:
@wooseokTT Here's my idea, and please let me know if I am understanding wrong.
 
Regardless of whether inputs are pre-sharded, if the output is a scalar (rank-0) we shouldn’t need to emit a mesh_shard (shard ->full). I think by semantics, a scalar result should be identical on all devices.
If per-device scalars can differ, then running shard->full is ill-defined, we must first make them equal via an explicit all-reduce/aggregation (or define a selection policy) before exposing a single-device value.
In other words, for replicated scalars we skip mesh_shard (no-op). If it’s not guaranteed replicated, we should materialize the correct collective rather than relying on shard->full.

wooseokTT:
@sshonTT  I didn't mention about pre-sharded inputs. If tt-xla expects pre-sharded outputs, the mesh_shard op with identity will return multiple buffers without doing anything. With "replicate", since all buffers are identical, we can get rid of the mesh_shard op. It is possible since the input / output tensor dims are identical. So, no need to generate them.

However, if we are computing the stablehlo graph with different frontend such as tt-rt, we expect output to be single device tensor. Then, the mesh_shard op makes sense as we pick one out of many during shard_to_full operation. 

So, if needed, you can make this as an option and enable them with tt-xla frontend.

sshonTT:
@wooseokTT I see what you mean — tt-rt (or other frontends) would depend on the mesh_shard op itself since it creates a single buffer. Then yeah I will defnitly make this as an option.
I have a quick (maybe unrelated) question though: do you know if ttrt handles the scalar tensor case? Since ttrt might still be using the tt-nn mesh_shard op, the same error should occur in that situation as well, right?

wooseokTT:
@sshonTT 

do you know if ttrt handles the scalar tensor case? 
=> There is no difference between scalar vs other tensors, I believe. Scalar value is single dim tensor, I think.

Since ttrt might still be using the tt-nn mesh_shard op, the same error should occur in that situation as well, right?
=> If stablehlo->ttir->ttnn->ttnn.fb includes mesh_shard op, ttrt will run with runtime/ccl/mesh_shard.cpp. I still don't understand what you mean by "the same error" here. Can you elaborate with more details?

sshonTT:
@wooseokTT A scalar is a 0-dim tensor. When running mesh_shard in TTNN (tt-metal), I’m seeing the error described [here](https://github.com/tenstorrent/tt-xla/issues/1993#issue-3590260676):


```
    torch_xla.sync(wait=True)
  File "/home/qb-training/pglusac/tmp/pytorch/xla/torch_xla/torch_xla.py", line 87, in sync
    torch_xla._XLAC._xla_step_marker(
RuntimeError: TT_FATAL @ /home/qb-training/pglusac/tt-xla/third_party/tt-mlir/src/tt-mlir/third_party/tt-metal/src/tt-metal/ttnn/core/tensor/xtensor/partition.cpp:37: std::all_of( normalized_dims.begin(), normalized_dims.end(), [tensor_dims](int dim) { return dim >= 0 && dim < tensor_dims; })
info:
Invalid dimension index; got dims: SmallVector([0]), tensor dimension: 0
```

This suggests tt-metal assumes the operand tensor must have rank > 0. And I guess ttrt might see the same issue as flows are same.

wooseokTT:
@sshonTT  It's been a while since I was debugging ttnn, but as far as I remember, the tensor<f32> becomes tensor<1xf32> while we are converting dialect, I believe, during type conversion. I guess this has been changed somehow, then. 

It seems that the error caused by the mesh_shard op with scalar value is different issue. We may change scalar to 1 dim tensor or we can put dim 1 if there is no dim info, then, I believe.

sshonTT:
Following an offline chat with @wooseokTT:

- We should not delete the mesh_shard op when the return type is annotated as `ttcore.shard_status<unsharded>` (e.g., `tensor<f32> { ttcore.shard_status = #ttcore.shard_status<unsharded> }`), since it returns a single buffer.
- tt-metal does handle `mesh_shard` on scalar (0-dim) tensors.
- If a frontend can consume a multi-buffer result, skipping the op can be an optimization, but it carries potential semantic risk 
- Since torch.sum returns a 0-dim tensor by default (keepdim=False), we will not convert it to a 1-dim tensor for now.

sshonTT:
Close as other possible solution(#5804) as a workaround.

PR #5757, opened 2025-11-07
[Codegen] Improve tt-alchemist build

### LINKED ISSUE #5250: [tt-Alchemist] Find a better way to integrate alchemist
Currently we install tt-alchemist as a wheel into tt-xla's the python enviroment and then find the venv via a enviroment variable from C++ world. We should think of a nicer way of packaging up alchemist from consumption from the frontend.

This is a somewhat lower priority code cleanup issue.

### LINKED ISSUE #5668: [Codegen] Remove unnecessary copies of lib objects due to wheel
Today, build copies all the .so files into single location:
```
${TT_MLIR_HOME}/build/tools/tt-alchemist/python/tt_alchemist/lib
```

From there, running the `build-wheel-tt-alchemist` target will run `pip wheel -w ...` which will (by pip wheel design) copy everything from the above dir to:
```
${TT_MLIR_HOME}/build/tools/tt-alchemist/python/build/lib/tt_alchemist/lib
```

### LINKED ISSUE #5672: [Codegen] Dynamically link mlir compiler to libtt-alchemist.so
Statically linking the whole compiler just bloats the binary (~350mb) and introduces conflicts when someone tries to link to both compiler and tt-alchemist lib. This will speed up tt-xla builds as well.

### LINKED ISSUE #5673: [Codegen] Provide a way to copy templates without a wheel
tt-xla builds and installs the wheel only because they need the templates dir. If we were to provide them with a way to copy these without needing the wheel, their build process could/would be faster.

svuckovicTT:
### Ticket
Closes #5668
Closes #5672
Closes #5673
Closes #5250

### Problem description
Building alchemist as a tool and wheel is pretty clunky and slow.

### What's changed
- Dynamically load `TTMLIRCompiler.so` instead of statically linking `TTMLIRCompilerStatic.a` => faster compiles
- Don't build python wheel, but rather the lib itself + install templates and metal dependencies => orders of magnitude faster compiles
- Add editable wheel install target => easier debugging, much faster compile than full wheel
- "Normal" wheel remains for distribution / CI testing
- Alchemist lib is built by default (no noticable overhead) but the wheels are default OFF

### Checklist

### DIFF
--- a/tools/CMakeLists.txt
+++ b/tools/CMakeLists.txt
@@ -26,6 +26,7 @@ endif()
 add_subdirectory(ttmlir-translate)
 add_subdirectory(ttnn-standalone)
 
-if (TTMLIR_ENABLE_ALCHEMIST)
+# TTMLIR_ENABLE_RUNTIME is needed because tt-metal doesn't build without it => can't install TT-NN dependencies
+if (TTMLIR_ENABLE_ALCHEMIST AND TTMLIR_ENABLE_RUNTIME)
   add_subdirectory(tt-alchemist)
 endif()


lib/CMakeLists.txt
```diff
@@ -74,25 +74,18 @@ add_mlir_library(TTMLIRCompilerStatic
...
   ${SHARDY_LIBS}
 )
 
-target_include_directories(TTMLIRCompilerStatic
-  INTERFACE
-    $<BUILD_INTERFACE:${TTMLIR_SOURCE_DIR}/include>
-    $<BUILD_INTERFACE:${TTMLIR_BINARY_DIR}/include>
-    $<INSTALL_INTERFACE:include>
-)
-
```
svuckovicTT, line 87: @nsmithtt @brnorris03 

Hey folks, wanted to ask your opinions on this piece of the change.

Context: I'm trying to improve the tt-alchemist build flow so that it's faster and doesn't rely on the static compiler lib (hundreds of megabytes), among various other improvements.

I tried using the shared variant `TTMLIRCompiler` instead, but was confused as to how it's supposed to work. Running `ldd` on it shows that it doesn't find metal dependencies:
```
$ ldd build/lib/libTTMLIRCompiler.so
	linux-vdso.so.1 (0x00007ffdae3eb000)
	libtt_metal.so => not found
	libdevice.so => not found
	_ttnncpp.so => not found
	libtt_stl.so => not found
	libtracy.so.0.10.0 => not found
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f1b7964c000)
	libz.so.1 => /lib/x86_64-linux-gnu/libz.so.1 (0x00007f1b79630000)
	libstdc++.so.6 => /lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007f1b79404000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f1b793e4000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f1b791bb000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f1b8d157000)
```

I'm no build expert by any means - my understanding is that, in an ideal scenario, we'd be running an install step after building our dependencies (e.g. metal libs) in order to get those into our `build/...` dirs. But we don't, so we seem to be stuck with manually copying these metal libs around whenever we're creating other targets that rely on these (ttrt, ttnnopmodellib, etc.).

As a way to unblock myself, I've added the metal libs relative path to the runpath field, but it's a hack I'm not particularly fond of.

My question is: how is the `libTTMLIRCompiler.so` meant to be used, and is it meant to be used for linking into other targets, such as tt-alchemist?

svuckovicTT: @tenstorrent/forge-developers-mlir-build

nsmithtt: Yeah libTTMLIRCompiler.so is meant to be used by whoever needs it.  A couple points of clarification:
- Dynamic linking vs static linking isn't going to make the tt-alchemist package smaller, it's almost 500MB in and of itself.  The only thing it will do is make tt-alchemist incremental builds faster.  If anything static linking will usually result in an overall smaller binary footprint because the linker can strip out unused symbols.
- libTTMLIRCompiler.so by default will set the RPATH == `.`, so as long as you package the metal libs in the same directory as it, it'll find them.  You can try copying them to the same dir and running ldd again.

svuckovicTT: The packaged wheel won't be smaller, but when using in-source (no wheel, or editable install of the wheel), the total footprint will be smaller, which is how it'll be used for tt-mlir and tt-xla after this change.

Gotcha, so should we always copy metal libs into our `build/lib` dir? Any downsides to that?

nsmithtt: > The packaged wheel won't be smaller, but when using in-source (no wheel, or editable install of the wheel), the total footprint will be smaller, which is how it'll be used for tt-mlir and tt-xla after this change.

How so?

> Gotcha, so should we always copy metal libs into our `build/lib` dir? Any downsides to that?

Typically this is done by the cmake install step which would copy all the libs to the right place in the package.

svuckovicTT: >How so?

Today:
tt-alchemist lib links in `libTTMLIRCompilerStatic.a` producing a ~500mb binary.

Tomorrow:
tt-alchemist lib dynamically links to `libTTMLIRCompiler.so` producing a ~20mb binary. Effectively it's referencing the already existing compiler dylib in `build/lib`, it doesn't need to copy it anywhere.

nsmithtt: Yeah, but `libTTMLIRCompiler.so` itself is 500mb.  Anyway, I agree that it'll result in faster builds for users of tt-alchemist.

lib/CMakeLists.txt
```diff
@@ -74,25 +74,18 @@ add_mlir_library(TTMLIRCompilerStatic
...
 #
 add_library(TTMLIRCompiler SHARED RegisterAll.cpp)
 
-
-set_target_properties(TTMLIRCompiler PROPERTIES INSTALL_RPATH "$ORIGIN")
-set_target_properties(TTMLIRCompiler PROPERTIES BUILD_WITH_INSTALL_RPATH TRUE)
+set_target_properties(TTMLIRCompiler PROPERTIES
+  INSTALL_RPATH "$ORIGIN:$ORIGIN/../../third_party/tt-metal/src/tt-metal/build/lib"
+  BUILD_WITH_INSTALL_RPATH TRUE
+)
```
brnorris03, line 97: Setting the install RPATH to point to a build tree would prevent the use of installed versions of the compiler that shouldn't require the presence of build trees. Did setting `BUILD_RPATH` not do what you expected?

tools/tt-alchemist/external_project_example/CMakeLists.txt
```diff
@@ -0,0 +1,27 @@
+cmake_minimum_required(VERSION 3.24)
+project(tt-alchemist-example)
+
+# Require TTMLIR_BUILD_DIR to be set
+if(NOT DEFINED TTMLIR_BUILD_DIR)
+  message(FATAL_ERROR
+    "TTMLIR_BUILD_DIR must be set. Example:\n"
+    "  cmake -DTTMLIR_BUILD_DIR=/path/to/tt-mlir/build .."
+  )
+endif()
```
brnorris03, line 10: If there is no specific reason to require the build tree, perhaps consider using an installed compiler, e.g.,  with `find_package(TTMLIR CONFIG HINTS ${TTMLIR_INSTALL_PATHS})`.

lib/CMakeLists.txt
svuckovicTT: Ah, wasn't aware that there were separate RPATHs between build and install, thank you! :)

I'll update to this:
```cmake
set_target_properties(TTMLIRCompiler PROPERTIES
  BUILD_RPATH "$ORIGIN:$ORIGIN/../../third_party/tt-metal/src/tt-metal/build/lib"
  INSTALL_RPATH "$ORIGIN"
  BUILD_WITH_INSTALL_RPATH FALSE
)
```

tools/tt-alchemist/external_project_example/CMakeLists.txt
svuckovicTT: That'd be ideal - our own frontends don't actually install the compiler, but rely on `ExternalProject_Add` mechanism, so they use env vars to find things within third party projects :(

For that reason, I'd prefer to keep this "format" of the example.

svuckovicTT:
> All your goal seems very good, but I would like to comment on word "distribution".
> 
> Our CI should use fastest variant of the wheel inside PR build-test process, essentially dev variant. It doesn't have to be packaged in a wheel, it can be run directly from the source or installed as editable wheel because our CI docker has all the necessary dev tools and build/install artifacts. Our PR should show regression and it should be as fast as possible.
> 
> If "distribution" means release version, something ment for the end user, then it should be built in Build Wheels/Nightly workflows because release process picks up nightly builds from Nightly run. This wheel should be build on manylinux image and tested on the base tt-forge docker which is far away from our CI docker - there are no dev libs and dev tools, it is basically bare linux with just a few libraries installed. Test done on release wheel resembles much closer what our users will experience when installing our releases and it reveals totally different set of problems than PR's.

In this scenario, the term "distributable" wheel is there to differentiate from the editable install wheel. The wheel isn't intended to be used by the end user, at least not in the nearest future, it's there to be share-able between repos/teams within the co.

However, you do raise a good point by asking whether we should be using the "distributable" wheel for testing purposes. It probably makes sense to make this leaner - let's talk offline, it's an orthogonal change anyway.

svuckovicTT:
@nsmithtt @vwellsTT @mtopalovicTT PTAL 😊

jmcgrathTT (approved):

PR #5767, opened 2025-11-10
Fix use-after-free in permute-reduce EIO pattern

azecevicTT:
`llvm-lit` test failed on macos build https://github.com/tenstorrent/tt-mlir/actions/runs/19220308155/job/54937104945

The issue is that we `replaceOp` in each iteration, at some point we also replace `permuteUser` which we use subsequently in `checkIdenticalTms`.

Workflow on which I tested the change: https://github.com/tenstorrent/tt-mlir/actions/runs/19227210017/job/54957089590


nsmithtt:
Awesome, thanks!

LPanosTT:
Thanks!

PR #5775, opened 2025-11-10
Remove PointToPoint fallback from runtime

### LINKED ISSUE #4602: Adopt ttnn.point_to_point
point_to_point was implemented on runtime as 'host fallback' because it wasn't landed on tt-metal at that time.
The PR implementing point_to_point on tt-metal repo had been landed.
Reimplement point_to_point as using real TTNN api.

Note:
ttnn.point_to_point adopt `fabric`. We cannot enable `fabric` by default because all other CCL ops using `fabric` wasn't generalized enough to use. So implement and testing is okay, but we should stay as 'host fallback' until all the CCL ops using `fabric` become generalized.

hkwonTT:
### Ticket
closes #4602

### Problem description
point_to_point was implemented in the runtime as a host fallback because it hadn’t landed in `tt-metal` at the time.
The PR that implements point_to_point in the `tt-metal` repo has now landed.

### What's changed
- Reimplement `point_to_point` using real TTNN CCL API.
- Enforce the same-device P2P constraint.
- Update CCL decomposition to use PointToPoint instead of allowing same-device P2P; replace with `Assign` op.
- Update the emitPy implementation

### Checklist

### DIFF
--- a/tools/tt-alchemist/templates/python/local/utils.py
+++ b/tools/tt-alchemist/templates/python/local/utils.py
@@ -3,6 +3,7 @@
 # SPDX-License-Identifier: Apache-2.0
 import ttnn
 import ttnn_supplemental
+import math
 
 
 # Monkey-patch ttnn with ttnn_supplemental objects
@@ -25,7 +26,7 @@ class DeviceGetter:
     def __del__(self):
         if self._instance is not None:
             ttnn.close_mesh_device(self._instance)
-        ttnn.set_fabric_config(ttnn.FabricConfig.DISABLED)
+            ttnn.set_fabric_config(ttnn.FabricConfig.DISABLED)
 
     @classmethod
     def get_device(cls, mesh_shape):
@@ -39,7 +40,9 @@ class DeviceGetter:
                     f"mesh_shape must be a non-empty list or tuple of positive integers, got {mesh_shape}"
                 )
             cls._mesh_shape = mesh_shape
-            ttnn.set_fabric_config(ttnn.FabricConfig.FABRIC_1D)
+
+            if math.prod(mesh_shape) >= 2:
+                ttnn.set_fabric_config(ttnn.FabricConfig.FABRIC_1D)
             cls._instance = ttnn.open_mesh_device(
                 mesh_shape=ttnn.MeshShape(mesh_shape),
                 l1_small_size=cls.l1_small_size,


include/ttmlir/Dialect/TTNN/IR/TTNNOps.td
```diff
@@ -2458,16 +2458,16 @@ def TTNN_PointToPointOp: TTNN_Op<"point_to_point"> {
...
     The operation returns a multi-device tensor whose buffer layout follows the mesh configuration.
   }];
 
   let arguments = (ins AnyRankedTensor:$input,
-                         DenseI64ArrayAttr:$send_coord,
-                         DenseI64ArrayAttr:$receive_coord,
-                         Optional<AnyRankedTensor>:$accum_tensor);
+                         DenseI64ArrayAttr:$sender_coord,
+                         DenseI64ArrayAttr:$receiver_coord,
+                         Optional<AnyRankedTensor>:$optional_output_tensor);
```
azecevic000: Do you have a use case for this? We removed optional output tensor from some other TTNN ops that we model because we don't use them currently.

hkwonTT: Yes. We are using this [here](https://github.com/tenstorrent/tt-mlir/blob/77f5051e4a82e1bd34fb502960e4b682f7100126/lib/Conversion/TTIRToTTNN/TTIRToTTNN.cpp#L2074) and [here](https://github.com/tenstorrent/tt-mlir/blob/77f5051e4a82e1bd34fb502960e4b682f7100126/lib/Conversion/TTIRToTTNN/TTIRToTTNN.cpp#L1845).

jnie-TT: It would be great if we could find a way to make this out of place and remove the optional output tensor for consistency, and also this makes consteval more tricky to model. If it's not possible for now please model this op as having memory effects on the output tensor

hkwonTT: I cannot figure out how to make this out of place yet. What should I do to model this op as having memory effects on the output tensor?

jnie-TT: `Arg<Optional<AnyRankedTensor>, "optional output tensor", [MemWrite]>` should do the trick. As an example please see `TTNN_BatchNormTrainingOp`

tools/tt-alchemist/templates/python/local/utils.py
```diff
@@ -35,6 +39,7 @@ def get_device(cls, mesh_shape):
                     f"mesh_shape must be a non-empty list or tuple of positive integers, got {mesh_shape}"
                 )
             cls._mesh_shape = mesh_shape
+            ttnn.set_fabric_config(ttnn.FabricConfig.FABRIC_1D)
```
svuckovicTT: Is this safe to do by default? I've asked once but never got a clear answer.

hkwonTT: Good point! As per my experience, we need to set `Fabric Config` only when it's multi-device. I'll update the code. Thanks!

Issue #5766, opened 2025-11-10
Set of failing tt-xla models after an uplift


ajakovljevicTT:
### Describe the bug

After an uplift, there is a set of tt-xla torch models failing with the following issue:
```
RuntimeError: TT_FATAL @ /localdev/ajakovljevic/tt-xla/third_party/tt-mlir/src/tt-mlir/third_party/tt-metal/src/tt-metal/ttnn/core/tensor/tensor_utils.cpp:55: new_volume == old_volume
info:
Invalid arguments to reshape
```
The issue has been bisected to this tt-mlir commit: https://github.com/tenstorrent/tt-mlir/commit/6fe566ea8d0fba651fb3d7f0793cf8cf85186212

Providing logs of tt-xla runs in the case it works and does not work below:

[log_working.log](https://github.com/user-attachments/files/23450982/log_working.log)
[log_not_working.log](https://github.com/user-attachments/files/23450983/log_not_working.log)

skhatriTT:
Hey @mtopalovicTT, what is our status on this one?

mtopalovicTT:
@skhatriTT This is fixed with #5786

PR #5786, opened 2025-11-11
Fix rope H dim tile alignment

mtopalovicTT:
We are getting errors from nightly xla because rotary_embedding is padding seq len dim when token_idx is not passed. Since it's tricky to figure out when to pass token_idx, this PR adds two changes:

* adds workaround for rotary_embedding when seq_len is not tile aligned
* adds hifi4 math fidelity to rotary_embedding - this is because with lofi (default value) we are getting low PCC with rotary embedding. I added this workaround explicitly in runtime not to waste time and to unblock uplift. I will model it it IR after this is merged


include/ttmlir/Dialect/TTNN/Transforms/Workarounds/Decomposition/RotaryEmbeddingOpRewritePattern.h
```diff
@@ -0,0 +1,24 @@
...
+    : public OpRewritePattern<RotaryEmbeddingOp> {
+public:
+  using OpRewritePattern<RotaryEmbeddingOp>::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(RotaryEmbeddingOp srcOp,
+                                PatternRewriter &rewriter) const override;
+};
+} // namespace mlir::tt::ttnn::workarounds::decomposition
+
+#endif // TTMLIR_DIALECT_TTNN_TRANSFORMS_PASSES_TTNNFUSINGOPREWRITEPATTERN_H
```
vmilosevic: :warning: **llvm\-header\-guard** :warning:
\#endif for a header guard should reference the guard macro in a comment
```suggestion
#endif // TTMLIR_DIALECT_TTNN_TRANSFORMS_WORKAROUNDS_DECOMPOSITION_ROTARYEMBEDDINGOPREWRITEPATTERN_H
```

lib/Dialect/TTNN/IR/TTNNOps.cpp
```diff
@@ -3895,10 +3895,6 @@ ::mlir::LogicalResult RotaryEmbeddingOp::verify() {
   mlir::RankedTensorType sinType = getSinCache().getType();
   mlir::RankedTensorType outputType = getResult().getType();
 
-  if (inputType != outputType) {
-    return emitOpError("input and output tensor types must match.");
-  }
-
```
azecevic000, line 3901: What are legitimate cases where this is too restrictive?

lib/Dialect/TTNN/Transforms/Workarounds/Decomposition/RotaryEmbeddingOpRewritePattern.cpp
```diff
@@ -0,0 +1,53 @@
+// // SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#include "ttmlir/Dialect/TTNN/Transforms/Workarounds/Decomposition/RotaryEmbeddingOpRewritePattern.h"
```
azecevic000, line 5: Nit: add new line below this one

lib/Dialect/TTNN/Transforms/Workarounds/Decomposition/RotaryEmbeddingOpRewritePattern.cpp
```diff
@@ -0,0 +1,53 @@
...
+  ArrayRef<int64_t> resultShape = resultType.getShape();
+  if (resultShape.size() < 2) {
+    return failure();
+  }
+
+  if (resultShape[resultShape.size() - 2] % TILE_HEIGHT == 0) {
+    return failure();
+  }
+
+  int64_t originalSeqLen = resultShape[resultShape.size() - 2];
```
azecevic000: Nit: rearrange the order, i.e.
```cpp
int64_t originalSeqLen = ...;
if (originalSeqLen % ...) {
  ...
}
```

runtime/lib/ttnn/operations/transformer/rotary_embedding.cpp
```diff
@@ -20,6 +20,9 @@ static void runRotaryEmbedding(const ::tt::target::ttnn::RotaryEmbeddingOp *op,
       tensorPool.getTTNNTensorAndValidate(op->sin_cache());
   std::optional<uint32_t> tokenIndex = op->token_index();
   std::optional<::ttnn::DeviceComputeKernelConfig> computeConfig;
+
+  std::get<::ttnn::WormholeComputeKernelConfig>(*computeConfig).math_fidelity =
+      MathFidelity::HiFi4;
```
azecevic000: Will this work on blackhole (or better question, does blackhole also have low PCC problem with LoFi)? It would be better to model this in compiler so that we can do it in the workaround pass, but it can be done in a subsequent PR.

lib/Dialect/TTNN/IR/TTNNOps.cpp
mtopalovicTT: Some nightly models on xla that had reshape after rope. When we decompose rope we rewrite output to respect ttnn constraints and thats when the validation fails if I leave this line...

runtime/lib/ttnn/operations/transformer/rotary_embedding.cpp
mtopalovicTT: Not sure didn't test it, could be.  Currently runtime doesn't even support blackhole compute config.

sdjordjevicTT: Just please create an issue so we can model this in the compiler.

test/python/golden/ttir_ops/fusing/test_rope_fusing.py
```diff
@@ -12,12 +12,8 @@
 pytestmark = pytest.mark.frontend("ttir")
```
sdjordjevicTT, line 12: Can you move this file under:
test/python/golden/ttir_ops/fusing/test_ttir_rope.py

mtopalovicTT: Not sure how I didn't move it the first time...

runtime/lib/ttnn/operations/transformer/rotary_embedding.cpp
mtopalovicTT: https://github.com/tenstorrent/tt-mlir/issues/5790

jameszianxuTT:
Linking to issue #5766

PR #5788, opened 2025-11-11
Support lowering partially indexed DMAs

nsmithtt:
This commit handles a missing lowering case for DMA's where the DMA is partially indexed and does not have the numElems attribute set.  This implicitly means, gather all of the indices that aren't applied (similar to python slice semantics).

This is implemented by:
- Adding a trivial loop around the unapplied indices over the remaining shape.
- Disambiguating this DMA form by introducing `hasExplicitSize`.  If false it means we need to lower via bullet 1, if true, it means some other lowering proved the unapplied indices are contiguous i.e. via affine lowering.

As an example, the following dma:
```mlir
%src : memref<1x2x2x2x!ttcore.tile<32x32, f32>, #ttcore.shard<16384x4096>, #l1_>
%dst :  memref<2x2x!ttcore.tile<32x32, f32>, #l1_>
%0 = d2m.dma %src[%c0, %c1], %dst : !d2m.mem_tx
```

Becomes:
```mlir
%2 = scf.for %arg1 = 0 to 2 step 1 iter_args(%arg2 = %1) -> (!d2m.mem_tx) {
  %3 = scf.for %arg3 = 0 to 2 step 1 iter_args(%arg4 = %arg2) -> (!d2m.mem_tx) {
    %tx = d2m.dma %src [%c0, %c1, %arg1, %arg3], %dst [%arg1, %arg3], <1> : !d2m.mem_tx
    scf.yield %tx : !d2m.mem_tx
  }
  scf.yield %3 : !d2m.mem_tx
}
```

### DIFF
--- a/lib/Dialect/D2M/Transforms/GenericGenerateDatamovement.cpp
+++ b/lib/Dialect/D2M/Transforms/GenericGenerateDatamovement.cpp
@@ -104,21 +104,22 @@ public:
     AffineMapAttr indexingMapAttr =
         (operandIndexingMap) ? AffineMapAttr::get(*operandIndexingMap)
                              : nullptr;
+    AffineMapAttr srcIndexingMapAttr = isOutput ? nullptr : indexingMapAttr;
+    AffineMapAttr dstIndexingMapAttr = isOutput ? indexingMapAttr : nullptr;
 
-    // affine map is associated with dst when isOutput is true, src otherwise
-    Value result;
-    if (isOutput) {
-      result = builder
-                   .create<d2m::DMAOp>(loc, src, dst, indexingMapAttr,
-                                       coreIndex, mcastShape)
-                   .getResult();
-    } else {
-      result = builder
-                   .create<d2m::DMAOp>(loc, src, indexingMapAttr, dst,
-                                       coreIndex, mcastShape)
-                   .getResult();
+    IntegerAttr numElemsAttr = nullptr;
+    // If mcast is programmed, we will implicitly set the numElems because we
+    // know we already gathered the data into a contiguous buffer.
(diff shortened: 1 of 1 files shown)


lib/Dialect/D2M/Transforms/GenericLowerDMAs.cpp
```diff
@@ -361,7 +360,66 @@ class D2MGenericLowerAffineDMAsRewritePattern : public OpRewritePattern<DMAOp> {
...
+          dma.getDstIndices(), dma.getMcastStartIndex(), dma.getMcastShape(),
+          /*numElems=*/1);
+      return success();
+    }
+
+    auto [lbs, ubs, steps] =
+        getLoopBounds(rewriter, dma->getLoc(), unappliedShape);
+
+    // Create a loop around the unapplied indices.
+    auto nullDmaTx = rewriter.create<d2m::NullTxOp>(dma.getLoc());
```
bgrady-tt, line 399: We _could_ test for contiguity here and collapse this into fewer DMAOps. Although it seems that the working assumption is that the partial indexing is a result of non-contiguity.

I think that is unnecessary right now, but I think an issue should be filed to look into it if ND TM ops use this mode heavily.

nsmithtt: This seems like it can get really tricky, what if your partially applied dma depends on a loop variable like:
```mlir
src : memref<2x2x4x4>
dst : memref<4x4>
scf.for i in ..:
  dma src[0, i] dst
```

You have to do a pretty complicated analysis (consteval'ing the loop) to determine contiguity for all loop iterations.

PR #5794, opened 2025-11-12
Adding emitpy reduction tests

sdjordjevicTT:
### Ticket
N/A

### Problem description
We don't have reduction tests for EmitPy.

### What's changed
Restructured reduction builder tests:
- Move reduction tests under a separate file
- File issues for currently non-supported cases
- Added emitpy execution target in tests

### Checklist

### DIFF
--- a/lib/Dialect/TTNN/IR/TTNNWorkaroundsPass.cpp
+++ b/lib/Dialect/TTNN/IR/TTNNWorkaroundsPass.cpp
@@ -744,13 +744,9 @@ TTNNOperandsWorkaroundsFactory::createReductionOpOperandsWorkarounds(
 // Factory method to create a set of workarounds for reduce product op operands.
 // tt-metal only supports full product reduction for bfloat16 data type.
 TTNNOperandsWorkarounds
-TTNNOperandsWorkaroundsFactory::createReduceProdOpOperandsWorkarounds(
-    mlir::Type elementType, bool allDimensions) {
-  bool isDataTypeWARequired = allDimensions && !elementType.isBF16();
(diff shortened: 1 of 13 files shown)


lib/Dialect/TTNN/IR/TTNNWorkaroundsPass.cpp
```diff
@@ -744,13 +744,9 @@ TTNNOperandsWorkaroundsFactory::createReductionOpOperandsWorkarounds(
...
 TTNNOperandsWorkarounds
-TTNNOperandsWorkaroundsFactory::createReduceProdOpOperandsWorkarounds(
-    mlir::Type elementType, bool allDimensions) {
-  bool isDataTypeWARequired = allDimensions && !elementType.isBF16();
+TTNNOperandsWorkaroundsFactory::createReduceProdOpOperandsWorkarounds() {
   TTNNOperandWorkarounds bf16Workaround;
-  if (isDataTypeWARequired) {
-    bf16Workaround.tensorDataTypeWorkaround = ttcore::DataType::BFloat16;
-  }
+  bf16Workaround.tensorDataTypeWorkaround = ttcore::DataType::BFloat16;
```
azecevic000, line 749: Does having `allDimensions=true` still require dtype workaround?

sdjordjevicTT: Yes, it is required not only for all dimension reduction but also for other dimensional types, as the PCC golden check fails without it.

sdjordjevicTT: The issue on the metal side:
https://github.com/tenstorrent/tt-metal/issues/24297

PR #5793, opened 2025-11-12
Use variables for C and C++ compiler in CMake

giordano:
### Ticket

N/A

### Problem description

I installed dependencies with https://github.com/tenstorrent/tt-metal/blob/757741f41567d22dae126cb7823c6bec3b3c1050/install_dependencies.sh, but that installs Clang with executables named `clang-17`/`clang++-17`.  I invoked the top-level CMake script with `-DCMAKE_C_COMPILER=clang-17 -DCMAKE_CXX_COMPILER=clang++-17`, but then this other CMake script goes on to ignore that setting and hardcodes `clang`/`clang++`.

### What's changed

Use the appropriate CMake variables, which may be supplied by the user from the command line, instead of hardcoding possibly non-existing executable names.

### Checklist
- [ ] New/Existing tests provide coverage for changes

PR #5795, opened 2025-11-12
Use variables for C and C++ compiler in CMake

nsmithtt:
There's no point in hardcoding them, users may set those variables when configuring `tt-mlir` but this CMake is hardcoding different values.

Also set `CMAKE_BUILD_PARALLEL_LEVEL` to avoid job timeouts.

### DIFF
--- a/.github/Dockerfile.ci
+++ b/.github/Dockerfile.ci
@@ -25,7 +25,7 @@ RUN git log -1
 RUN <<EOT
 export CMAKE_BUILD_PARALLEL_LEVEL=16
 # ANCHOR: environment_build
-cmake -B env/build env
+cmake -B env/build env -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++
 cmake --build env/build
 # ANCHOR_END: environment_build
 EOT


nsmithtt:
Build image fails, doesn't seem worth it to invest in this further at this time.  We can reopen if this is truly blocking a third party.

giordano:
As elaborated in https://github.com/tenstorrent/tt-mlir/pull/5793#issue-3617133800, the [documented way to install the dependencies on Ubuntu](https://docs.tenstorrent.com/tt-mlir/getting-started.html#ubuntu) installs Clang with versioned executables `clang-17`/`clang++-17`, not `clang`/`clang++`.

> if this is truly blocking a third party.

Honestly I have no clue how on Earth this works for anyone, at least when following the documentation.

giordano:
BTW, the problem is that now the build is correctly picking up the same compiler building llvm-project as for the overall project: https://github.com/tenstorrent/tt-mlir/actions/runs/19582314444/job/56083043505#step:6:147
```
#11 [ci-build 7/7] RUN <<EOT (# ANCHOR: environment_build...)
#11 0.619 -- The CXX compiler identification is GNU 11.4.0
#11 0.704 -- The C compiler identification is GNU 11.4.0
[...]
#11 476.1 [ 40%] Performing configure step for 'llvm-project'
#11 476.1 -- The C compiler identification is GNU 11.4.0
#11 476.2 -- The CXX compiler identification is GNU 11.4.0
```
while, for example, in the successful build https://github.com/tenstorrent/tt-mlir/actions/runs/21830611841/job/62987596320#step:6:146
```
#11 [ci-build 7/7] RUN <<EOT (# ANCHOR: environment_build...)
#11 0.476 -- The CXX compiler identification is GNU 11.4.0
#11 0.540 -- The C compiler identification is GNU 11.4.0
[...]
#11 957.3 [ 40%] Performing configure step for 'llvm-project'
#11 957.5 -- The C compiler identification is Clang 17.0.6
#11 957.6 -- The CXX compiler identification is Clang 17.0.6
```
this was using two different compilers.  The simple solution would be to build everything with a single compiler (`clang-17`/`clang++-17`)

> doesn't seem worth it to invest in this further at this time

This took me two minutes of investigation.

giordano:
Also, these three lines, which are ***only*** in your CI setup, https://github.com/tenstorrent/tt-mlir/blob/7781efc34ad12ed43869b57e9748abc3e0e986a3/.github/Dockerfile.base#L57-L59 is the reason why the build happens to work in CI, but not for anyone else when following your documentation, because manually creating the symlinks isn't suggested anywhere.  Sure, I could do it myself, but hardcoding paths of compiler is also not a great practice to start with, when there's no way to change them.

nsmithtt:
Ok retrying

nsumrakTT (approved):

nsmithtt:
Well, concerningly, this very innocuous change is causing massive fallout:
```
error: unknown: 'func.func' op symbol's parent must have the SymbolTable trait
```

I'll try and repro in IRD.

vwellsTT (approved):

giordano:
Thanks for completing this PR!

nsmithtt:
> Thanks for completing this PR!

Sorry for all the churn, it became a pretty drawn out process 😅. Glad to see it finally land

Issue #4214, opened 2025-07-25
Add LLM specific ops to TTNN IR


odjuricicTT:
In order to run LLMs more performantly the following ops need to be added:

 - [ ] ttnn.experimental.nlp_concat_heads
 - [ ] ttnn.experimental.nlp_concat_heads_decode
 - [ ] ttnn.experimental.nlp_create_qkv_heads
 - [ ] ttnn.experimental.nlp_create_qkv_heads_decode
 - [ ] ttnn.experimental.paged_fill_cache
 - [ ] ttnn.experimental.paged_update_cache
 - [ ] ttnn.experimental.rotary_embedding_llama
 - [ ] ttnn.rms_norm
 - [ ] ttnn.slice
 - [ ] ttnn.transformer.paged_scaled_dot_product_attention_decode
 - [ ] ttnn.transformer.scaled_dot_product_attention
 - [ ] ttnn.unsqueeze_to_4D

These are the ops found in the llama3.2 architecture. More will be added to the list as we take a look at other architectures.

For most of these ops fusing patterns will be needed, more investigation is required to see how to approach adding them.

AleksKnezevic:
@ddilbazTT can you please add the split/concat heads issues you have here.

ddilbazTT:
Hi all, 

I have added support for concatenating q-k-v heads through a fusion pattern: https://github.com/tenstorrent/tt-mlir/pull/4068 

I have self-assigned split heads op support and fusion pattern to myself: https://github.com/tenstorrent/tt-mlir/issues/4037 As discussed with Aleks, the way to go about it will be to concatenate the qkv inputs first then adding split heads op after.

umalesTT:
I added 2 more issues under this umbrella issue for LLM training related ops directly supported in metal.
Fusing for both of them will be done on frontend level through composite patching, so we don't need to worry about their fusing patterns.

rpavlovicTT:
https://github.com/tenstorrent/tt-mlir/issues/5530 linking related issue here for better visibility

sdjordjevicTT:
@odjuricicTT I am going over some old issues. I believe we can close this one?

Issue #5738, opened 2025-11-06
`getOpRuntime` for `ttnn.paged_update_cache` sporadically hangs.


LPanosTT:
The test is located in: `test/unittests/OpModel/TTNN/Op/TestOpModelInterface.cpp` `TEST_F(OpModelBase, PagedUpdateCacheOpInterface)`

At the bottom of the test there is a section under an `if` block which does not run. Enable it to repro.

NOTE: In rare instances this part does not hang. I do not know the conditions in which it passes.

My theory is that garbage data is passed as the `page_table` and `update_index`. Since these tensors are integer tensors meant to index the `cache` input, this garbage data causes the hang. Although I am unsure if this is true.

The reason I think this might be the case is since ttnn will use the generated `TensorSpec` to allocate a tensor to use as a mock input. In this flow, I did not see anything indicating that the memory which is allocated is wiped, or set to a set value.

CC: @sgholamiTT @odjuricicTT @ckaravasilisTT 

NOTE: This test is not in `main` as of the making of this issue (Nov 6, 2025, 1:22 PM EST). But it will be soon

sgholamiTT:
Thanks for adding this, Lewis.

I am willing to discuss this issue in the next perf weekly sync. It is possible that we populate garbage data in tensors and the ops that are dependant on the actual data can go to undefined behavior (maybe unpredictable is a better word?).
If that's the case, we might be seeing hangs when running some models in the future which would be difficult to debug.

fyi, @xanderchin @rpavlovicTT @gfengTT

gfengTT:
So if you run a runtime query immediately after a constraint query with the same parameters, the program hangs? This is like what we saw in https://github.com/tenstorrent/tt-metal/issues/25772. Can you try to cherry-pick https://github.com/tenstorrent/tt-metal/commit/c6e5a0ee53ba77cb8d6d6261079167ef6a7ff805 ?

PR #5805, opened 2025-11-12
Model `ttnn.paged_scaled_dot_product_attention_decode` op

LPanosTT:
### Ticket
Closes https://github.com/tenstorrent/tt-mlir/issues/4214?issue=tenstorrent%7Ctt-mlir%7C4264

### Problem description
We do not model `ttnn.paged_scaled_dot_product_attention_decode`

NOTE: There is an issue which causes `PagedUpdateCacheOp` silicon/runtime tests to hang. It seems to be because we cannot choose which integers are actually placed in the page table tensor, which is used to index the cache. Generating random values for the page table will not suffice as we do not know what happens if the indices in the tensor are out of bounds (likely causing hang). This op also indexes its inputs using a page table tensor. And so, we do not test runtime in `TestOpModelInterface.cpp` or have a silicon test: https://github.com/tenstorrent/tt-mlir/issues/5738

### What's changed
- Model `PagedScaledDotProductAttentionDecode` in runtime, ttnn, ttir, and shlo -> ttir as custom call.
    - Requires workaround forcing the `pageTable` and `curPosTensor` args to be in ROW MAJOR.
- Add OpModelLib and OpModelInterface tests
- Also removed a verification invariant for `PagedUpdateCache` which makes sure `numHeads == 32` as this does not appear to be necessary from testing I've done.
- Had to include `NOLINT` barriers for a specific linter error caused by the use/passing around of an optional `APFloat` this was investigated heavily when bringing up `ScaledDotProductAttentionDecode` and we believe it is a false-negative.
    - This same `NOLINT` is included in some areas of `ScaledDotProductAttentionDecode` currently.
### Checklist

### DIFF
--- a/include/ttmlir/Dialect/TTIR/IR/TTIROps.td
+++ b/include/ttmlir/Dialect/TTIR/IR/TTIROps.td
@@ -3165,6 +3165,7 @@ def TTIR_PagedFillCacheOp : TTIR_NamedOp<"paged_fill_cache"> {
 
   let results = (outs AnyRankedTensor:$result);
 
+  // The cache tensor is modified in place by this operation, thus it acts as the DPS output of this operation.
   let extraClassDeclaration = [{
       MutableOperandRange getDpsInitsMutable() { return getCacheMutable(); }
   }];


test/unittests/OpModel/TTNN/Lib/TestOpModelLib.cpp
```diff
@@ -5378,6 +5378,171 @@ INSTANTIATE_TEST_SUITE_P(ScaledDotProductAttentionDecodeTests,
...
+  detail::ExpectedResult expectedResult;
+};
+
+class OpModelPagedScaledDotProductAttentionDecodeParam
+    : public OpModelTest,
+      public testing::WithParamInterface<
+          PagedScaledDotProductAttentionDecodeOpParam> {
+protected:
+  void RunTest() {
+    // NOLINTBEGIN(clang-analyzer-cplusplus.NewDelete)
```
vmilosevic, line 5506: :warning: **clang\-tidy\-nolint** :warning:
unmatched `` NOLINTBEGIN `` comment without a subsequent `` NOLINTEND `` comment

vmilosevic (commented):

lib/Dialect/TTNN/Interfaces/TTNNOpModelInterface.cpp
```diff
@@ -2043,6 +2043,138 @@ llvm::Expected<size_t> ScaledDotProductAttentionDecodeOp::getOpRuntime(
...
+
+  TypedValue<RankedTensorType> attentionMask = op.getAttentionMask();
+  TypedValue<RankedTensorType> curPosTensor = op.getCurPosTensor();
+  TypedValue<RankedTensorType> attentionSink = op.getAttentionSink();
+
+  int32_t idx = 4;
+  if (attentionMask) {
+    ret.attentionMaskShape =
+        llvm::SmallVector<int64_t>(attentionMask.getType().getShape());
+    ret.attentionMaskLayout = inputs[idx];
```
sgholamiTT, line 2098: I'd recommend adding sth like
```c++
assert(inputs.size() > idx && "some error");
```
in each `if` statement to avoid getting out of bound errors.

lib/OpModel/TTNN/TTNNOpModel.cpp
```diff
@@ -2587,6 +2587,160 @@ llvm::Expected<size_t> OpModel<ScaledDotProductAttentionDecodeOp>::getOpRuntime(
...
+  std::optional<uint32_t> slidingWindowSize = std::nullopt;
+
+  auto pagedScaledDotProductAttentionDecodeOpQuery = [=]() {
+    return ::ttnn::graph::query_op_constraints(
+        ::ttnn::transformer::paged_scaled_dot_product_attention_decode, device,
+        querySpec, keySpec, valueSpec, pageTableSpec, isCausal,
+        attentionMaskSpec, curPosTensorSpec, attentionSinkSpec, scaleFloat,
+        slidingWindowSize, detail::getNullableMemoryConfig(outputLayout),
+        /*program_config=*/std::nullopt,
+        /*compute_kernel_config=*/std::nullopt);
```
sgholamiTT, line 2654: Nit: Let's either do:
```c++
/* sliding_window_size =*/std::nullopt,
```
or
```c++
slidingWindowSize, detail::getNullableMemoryConfig(outputLayout),
programConfig, computeKernelConfig);
```
For consistency.

lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
```diff
@@ -4140,6 +4140,160 @@ class StableHLOToTTIRScaledDotProductAttentionDecodeOpConversionPattern
...
+      if (isCausalStringAttr.getValue().lower() == "true") {
+        isCausal = true;
+      } else if (isCausalStringAttr.getValue().lower() == "false") {
+        isCausal = false;
+      } else {
+        return rewriter.notifyMatchFailure(
+            srcOp, "is_causal attribute must be true or false. Received \"" +
+                       isCausalStringAttr.getValue() + "\".");
+      }
+    }
```
azecevic000, line 4214: It would make sense to have this as a helper function `parseBoolAttr` (either as a separate lib or static function in this file). I would assume that other ops would require the same logic in the future (and it already repeats several times in this conversion alone).

lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
```diff
@@ -4140,6 +4140,160 @@ class StableHLOToTTIRScaledDotProductAttentionDecodeOpConversionPattern
...
+    if (scaleStringAttr) {
+      float _scale;
+      if (!llvm::to_float(scaleStringAttr.getValue(), _scale)) {
+        return rewriter.notifyMatchFailure(
+            srcOp,
+            "scale attribute string must be convertible to float. Received \"" +
+                scaleStringAttr.getValue() + "\".");
+      }
+      scale = _scale;
+    }
```
azecevic000, line 4227: The same goes from FloatAttr.

Issue #3883, opened 2025-06-20
[Bounty $500]CPU Hoisted TTIR ReduceOrOp Fails at Runtime


vwellsTT:
# Bounty $[500] [CPU Hoisted TTIR ReduceOrOp Fails at Runtime]

## 📝 Background

Theoretical support for hoisting TTIR ReduceOrOp to CPU fallback has been added; however, while the op compiles, it fails at runtime with an error like:
```
FAILED test/python/golden/test_ttir_ops.py::test_hoisted_reduce_or[ttnn-dim_args0-4x4] - builder.base.builder_utils.TTBuilderGoldenException: Failed: program-level output golden comparison failed, actual_pcc=0.2947140183650278 < expected_pcc=0.99
FAILED test/python/golden/test_ttir_ops.py::test_hoisted_reduce_or[ttmetal-dim_args0-4x4] - builder.base.builder_utils.TTBuilderGoldenException: Failed: program-level output golden comparison failed, actual_pcc=0.29421406235980374 < expected_pcc=0.99
```

## 🎯 What Success Looks Like
* `test_hoisted_reduce_or` in `test_ttir_ops.py` should be enabled and pass

## 💡 Problem to Solve
This op's lowering to LLVM IR for CPU execution must be incorrect at some stage; the current lowering executes, but does not produce correct results.

### Steps to reproduce the issue
1. Uncomment the "xfail" directive for `test_hoisted_reduce_or` in `test_ttir_ops.py`
2. Build ttrt (`cmake --build build && cmake --build build -- ttrt`), and run pytest: `python -m pytest test/python/golden/test_ttir_ops.py::test_hoisted_reduce_or -vv --path=output --sys-desc=ttrt-artifacts/system_desc.ttsys` -- note that after recent changes, this will both build the binary AND run it
3. observe the output--it should show a failure

### Point of Contact
Any general questions about our CPU fallback pipeline, etc. can be addressed to me here or on discord

Mountagha:
hey @skhatriTT  @marty1885 can I be assigned to this ?

marty1885:
[...]

PR #5814, opened 2025-11-12
Reduce or op cpufallback

Mountagha:
### Ticket
#3883 

### Problem description
Theoretical support for hoisting TTIR ReduceOrOp to CPU fallback has been added; however, while the op compiles, it fails at runtime with an error like:
```
FAILED test/python/golden/test_ttir_ops.py::test_hoisted_reduce_or[ttnn-dim_args0-4x4] - builder.base.builder_utils.TTBuilderGoldenException: Failed: program-level output golden comparison failed, actual_pcc=0.2947140183650278 < expected_pcc=0.99
FAILED test/python/golden/test_ttir_ops.py::test_hoisted_reduce_or[ttmetal-dim_args0-4x4] - builder.base.builder_utils.TTBuilderGoldenException: Failed: program-level output golden comparison failed, actual_pcc=0.29421406235980374 < expected_pcc=0.99
```
### What's changed
The TTIRToTTIRDecomposition pattern of ReduceOR op was simply mapping it to a simple reduce sum. More have been added to mimic the ReduceOr (sum, compare, notEqual, typecast) Ops to be more aligned with the semantic of Reduce_or.
casted the reduce_or_golden output to float for type conformity (though we might need to later adapt in case of a tensor with integer values)

### Checklist


Mountagha:
Ping @vwellsTT  Waiting for your inputs. :smile:

lib/Conversion/TTIRToTTIRDecomposition/TTIRToTTIRDecomposition.cpp
```diff
@@ -2236,9 +2236,27 @@ struct ReductionOrPattern : public OpConversionPattern<ttir::ReduceOrOp> {
...
+        DenseElementsAttr::get(reduceOutputType, zerAttr);
+    mlir::Value zeroConstant = rewriter.create<ttir::ConstantOp>(
+        ttmlir::utils::appendLocationSuffix(op.getLoc(), "_zeroConstant"),
+        reduceOutputType, zeroConstantAttr);
+
+    // Compare sum != 0.
+    mlir::Value cmpOp = ttir::utils::createDPSOp<ttir::NotEqualOp>(
+        rewriter, op.getLoc(), reduceOutputType, sumOp, zeroConstant);
+
+    // typecast boolean result to float type.
```
vwellsTT: nit: capital T

vwellsTT (approved):
I think this pattern makes more sense than existing one to me--an alternative would be to create a separate rewriter for the CPU fallback target vs TTNN.  Maybe @azecevicTT or someone else can also take a look to confirm.

Mountagha:
> I think this pattern makes more sense than existing one to me--an alternative would be to create a separate rewriter for the CPU fallback target vs TTNN. Maybe @azecevicTT or someone else can also take a look to confirm.

Aight I will wait on them inputs then. Going to fix mlir tests meanwhile

vwellsTT:
> @vwellsTT @Mountagha In principle I agree that this decomposition is more complete than the previous one, my only concern is that this might result in perf regression in TTNN if the previous implementation worked well for TTNN. It can get ugly, but one possible solution might be adding a TTNN fusing pattern:
typecast(neq(sum(x), zero)) -> sum(x)

That makes sense.  I don't personally have a strong preference between that and just adding a separate pattern for `ReductionOrPattern` and `ReductionOrTTNNPattern` with old logic (since we do have separate TTIRDecomp targets already anyway)--do you have thoughts? Or is there someone else who we should check with?

Mountagha:
hello @azecevicTT trigger the CI for me if you still around. Thanks

azecevicTT (approved):

Issue #5091, opened 2025-09-25
Add reduction to ttir.scatter op


vladimirjovanovicTT:
[This PR](https://github.com/tenstorrent/tt-metal/pull/26757) is adding support for reduction (opt_reduction argument) in ttnn scatter op.
We need to add corresponding MLIR support for the reduction parameter.

vladimirjovanovicTT:
fyi @abogdanovicTT @pglusacTT

jojo-771771:
I can work on this. Please assign!

jojo-771771:
I can work on this. Please assign!

jojo-771771:
I can work on this. Please assign!

jojo-771771:
I can work on this. Please assign!

PR #5821, opened 2025-11-13
Adding reduction type to ttir.scatter op

abogdanovicTT:
### Ticket
Closes [#5091](https://github.com/tenstorrent/tt-mlir/issues/5091)
Related to [#31909](https://github.com/tenstorrent/tt-metal/issues/31909)

### Problem description
MLIR didn't have support for the reduction in ScatterOp.

### What's changed

- The current ttir.scatter and ttir.scatter_in_dim are extended with a scatter_reduce_type attribute, which enables reduction operations supported in TT-Metal. There is currently an open issue in TT-Metal for [scatter reduction](https://github.com/tenstorrent/tt-metal/issues/31909), and once it’s resolved, MLIR support will be added. 
- A workaround in ScatterOp has also been added to support bf16 and int32 input tensors.

### Checklist

### DIFF
--- a/test/python/golden/mlir_snippets/ttir/ttir_scatter_in_dim.mlir
+++ b/test/python/golden/mlir_snippets/ttir/ttir_scatter_in_dim.mlir
@@ -1,6 +1,6 @@
 module {
   func.func @model(%arg0: tensor<1x16x1x1xf32>, %arg1: tensor<1x4x1x1xi32>, %arg2: tensor<1x4x1x1xf32>) -> tensor<1x16x1x1xf32> {
-    %1 = "ttir.scatter"(%arg0, %arg1, %arg2) <{dim = 1 : i32, scatter_reduce_type = #ttcore.reduce_type<invalid>}> : (tensor<1x16x1x1xf32>, tensor<1x4x1x1xi32>, tensor<1x4x1x1xf32>) -> tensor<1x16x1x1xf32>
+    %1 = "ttir.scatter"(%arg0, %arg1, %arg2) <{dim = 1 : i32, scatter_reduce_type = #ttcore.reduce_type<sum>}> : (tensor<1x16x1x1xf32>, tensor<1x4x1x1xi32>, tensor<1x4x1x1xf32>) -> tensor<1x16x1x1xf32>
     return %1 : tensor<1x16x1x1xf32>
   }
 }
--- a/tools/builder/ttir/ttir_builder.py
+++ b/tools/builder/ttir/ttir_builder.py
@@ -1977,7 +1977,7 @@ class TTIRBuilder(Builder):
         index: Operand,
         source: Operand,
         dim: int,
-        scatter_reduce_type: ttcore.ReduceType,
+        scatter_reduce_type: Attribute,
         output_type: Optional[torch.dtype] = None,
         loc: Optional[str] = None,
         unit_attrs: Optional[List[str]] = None,
--- a/tools/golden/mapping.py
+++ b/tools/golden/mapping.py
@@ -3571,13 +3571,44 @@ def ttir_scatter_golden(
     index: GoldenMapTensor,
     source: GoldenMapTensor,
     dim: IntegerAttr,
+    scatter_reduce_type: Attribute,
     output_type_mlir: Type,
 ) -> GoldenMapTensor:
     dim_value = unpack_mlir_attr(dim)
+    # TODO(tapspatel): Scatter reduce type is shown as basic MLIR Attribute, should be properly handled as ReduceType attribute
+    scatter_string = str(scatter_reduce_type)
+    possible_strings = {
+        f"#ttcore.reduce_type<{s}>": s for s in ["sum", "prod", "max", "min", "invalid"]
+    }
+    if scatter_string not in possible_strings:
+        raise ValueError(f"Invalid scatter reduce type: {scatter_string}")
+    scatter_reduce_type = possible_strings[scatter_string]
+
     output_dtype = mlir_type_to_torch_dtype(output_type_mlir)
     index_copy = index.clone()
     index_copy = index_copy.to(torch.int64)
-    return torch.scatter(input_tensor, dim_value, index_copy, source).to(output_dtype)
+    if scatter_reduce_type == "sum":
+        out_tensor = torch.scatter_reduce(
+            input_tensor, dim_value, index_copy, source, reduce="sum"
+        )
+    elif scatter_reduce_type == "prod":
+        out_tensor = torch.scatter_reduce(
+            input_tensor, dim_value, index_copy, source, reduce="prod"
+        )
+    elif scatter_reduce_type == "max":
+        out_tensor = torch.scatter_reduce(
+            input_tensor, dim_value, index_copy, source, reduce="amax"
+        )
+    elif scatter_reduce_type == "min":
+        out_tensor = torch.scatter_reduce(
+            input_tensor, dim_value, index_copy, source, reduce="amin"
+        )
+    elif scatter_reduce_type == "invalid":
+        out_tensor = torch.scatter(input_tensor, dim_value, index_copy, source)
+    else:
+        raise ValueError(f"Invalid scatter reduce type: {scatter_reduce_type}")
+
+    return out_tensor.to(output_dtype)
 
 
 def ttir_reverse_golden(


lib/Conversion/TTIRToTTNN/TTIRToTTNN.cpp
```diff
@@ -1760,7 +1713,8 @@ class ScatterInDimOpConversionPattern
     rewriter.replaceOpWithNewOp<ttnn::ScatterOp>(
         op, this->getTypeConverter()->convertType(op.getType()),
         adaptor.getInput(), adaptor.getIndex(), adaptor.getSource(),
-        rewriter.getI32IntegerAttr(op.getDim()), nullptr);
+        rewriter.getI32IntegerAttr(op.getDim()),
+        adaptor.getScatterReduceTypeAttr(), nullptr);
```
umalesTT: Can you please add comment on what this nullptr stands for? Similar as in other conversions, something like ```/*memory_config=*/```

lib/Dialect/TTNN/IR/TTNNWorkaroundsPass.cpp
```diff
@@ -219,7 +189,9 @@ TTNNOperandsWorkaroundsFactory::createScatterOpOperandsWorkarounds(
 
   bool isLayoutWorkaroundRequired =
       (inputLayoutAttr.isTiled() && inputType.getElementType().isF32()) ||
-      (sourceLayoutAttr.isTiled() && sourceType.getElementType().isF32());
+      (sourceLayoutAttr.isTiled() && sourceType.getElementType().isF32()) ||
+      (inputLayoutAttr.isTiled() && inputType.getElementType().isBF16()) ||
+      (sourceLayoutAttr.isTiled() && sourceType.getElementType().isBF16());
```
umalesTT: Are we sure tile layout is unsupported only for these 2 types? 
If tile layout is generally unsupported we can just check ```isTiled()```, skipping the type inspection?

runtime/lib/ttnn/operations/data_movement/scatter.cpp
```diff
@@ -23,9 +23,35 @@ void run(const ::tt::target::ttnn::ScatterOp *op, ProgramContext &context) {
...
+    scatter_reduce_type =
+        ::ttnn::operations::data_movement::scatter::ScatterReductionType::AMIN;
+    break;
+  case 2: // Max -> AMAX
+    scatter_reduce_type =
+        ::ttnn::operations::data_movement::scatter::ScatterReductionType::AMAX;
+    break;
+  default:
+    scatter_reduce_type =
+        ::ttnn::operations::data_movement::scatter::ScatterReductionType::ADD;
```
nvukobratTT: I this a valid interpretation? E.g., if we don't specify the reduction type, it'll be reduce sum? I.e., is this what both TTIR and metal will expect? :))

vladimirjovanovicTT: @nvukobratTT sum reduction is default behavior for embedding_bw ttnn op, which is a special case of scatter, so this seems in line with this.

include/ttmlir/Dialect/TTIR/IR/TTIROps.td
```diff
@@ -2221,6 +2221,65 @@ def TTIR_MeanOp : TTIR_ReductionOp<"mean"> {
   }];
 }
 
+def TTIR_MulOp : TTIR_ReductionOp<"mul"> {
```
sdjordjevicTT: Talked offline, we should remove this op. We already have TTIR_ProdOp, which does the same thing.

lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
```diff
@@ -2442,6 +2447,14 @@ class StableHLOToTTIRScatterOpConversionPattern
     auto outputType = mlir::cast<RankedTensorType>(
         this->getTypeConverter()->convertType(srcOp.getResults()[0].getType()));
 
+    // Convert reduceType stablehlo attribute into ttir attribute
+    llvm::ErrorOr<ttcore::ReduceType> scatter_reduce_type =
```
azecevic000: Nit: use camelCase

runtime/lib/ttnn/operations/data_movement/scatter.cpp
```diff
@@ -23,9 +23,35 @@ void run(const ::tt::target::ttnn::ScatterOp *op, ProgramContext &context) {
   std::optional<::ttnn::MemoryConfig> outputMemoryConfig =
       ::tt::runtime::ttnn::utils::createMemoryConfigIfNeeded(
           op->memory_config());
+  // TT Metal currently only supports the following scatter reduction types:
+  // SUM, MULTIPLY, AMIN, AMAX.
+  ::ttnn::operations::data_movement::scatter::ScatterReductionType
+      scatter_reduce_type;
```
azecevic000: Same here.

lib/Target/TTNN/TTNNToFlatbuffer.cpp
```diff
@@ -745,9 +745,9 @@ createOp(FlatbufferObjectCache &cache, ScatterOp op) {
       getOperandThroughDPSOps(op.getSource()));
   auto output = cache.getOrCreate(op.getResult(), tensorValueToFlatbuffer);
   auto memoryConfig = getMemoryConfigIfNeeded(cache, op);
-  return ::tt::target::ttnn::CreateScatterOp(*cache.fbb, input, output, index,
-                                             sourceTensor, op.getDim(),
-                                             memoryConfig);
+  return ::tt::target::ttnn::CreateScatterOp(
+      *cache.fbb, input, output, index, sourceTensor, op.getDim(), memoryConfig,
+      static_cast<uint32_t>(op.getScatterReduceType()));
```
azecevic000: Please create a flatbuffer enum of `ScatterReduceType` instead of casting to `uint32_t`.

runtime/lib/ttnn/operations/data_movement/scatter.cpp
azecevic000: Synced with @vladimirjovanovicTT offline, we have to see what's the default behaviour in TTNN (when reduce type is `std::nullopt`). If it's not add, we also have to change the type of `scatter_reduce_type` in TTNN (and TTIR) dialect to `OptionalAttr<TTCore_ReduceTypeAttr>`. As it stands (when the type is `TTCore_ReduceTypeAttr`), default here should be a simple `assert(false && "unknown reduce type);"`, otherwise we might end up with some new reduce type XYZ that is converted to ADD (I believe we have compiler option that should prohibit this behaviour, but the point still stands).

vladimirjovanovicTT: @azecevicTT 
https://github.com/tenstorrent/tt-metal/blob/34722ed36aacbcf4a457e06c793db3e79df4e5b3/ttnn/cpp/ttnn/operations/data_movement/scatter/device/scatter_device_operation.cpp#L89

Seems that we should map to this?

test/ttmlir/Dialect/TTNN/simple_scatter.mlir
```diff
@@ -1,23 +1,29 @@
...
+    %2 = "ttir.scatter"(%arg0, %1, %arg1, %0) <{index_vector_dim = 1 : i32, indices_are_sorted = false, input_batching_dims = array<i32>, inserted_window_dims = array<i32: 0>, scatter_dims_to_operand_dims = array<i32: 0>, scatter_indices_batching_dims = array<i32>, unique_indices = false, update_window_dims = array<i32: 1, 2, 3>, scatter_reduce_type = #ttcore.reduce_type<sum>}> : (tensor<1x3x320x320xf32>, tensor<1x1xi32>, tensor<1x3x32x32xf32>, tensor<1x3x320x320xf32>) -> tensor<1x3x320x320xf32>
+    // CHECK: %{{[0-9]+}} = "ttnn.scatter"({{.*}}) <{dim = 0 : i32, scatter_reduce_type = #ttcore.reduce_type<sum>}> : (tensor<1x3x320x320xf32, {{.*}}>, tensor<1x3x32x32xsi32, {{.*}}>, tensor<1x3x32x32xf32, {{.*}}>) -> tensor<1x3x320x320xf32, {{.*}}>
     return %2 : tensor<1x3x320x320xf32>
     // CHECK: return %{{[0-9]+}} : tensor<1x3x320x320xf32, {{.*}}>
   }
 
-  // gpt-oss - multi-dimensional scatter with scatter operation broken into multiple scatter operations each handling index_shape[dim] < 256
```
ddilbazTT, line 12: Instead of replacing the old test cases, can we add the new ones separately? For example, gpt-oss is testing multi-dimensional scatter.

runtime/lib/ttnn/operations/data_movement/scatter.cpp
azecevic000: https://github.com/tenstorrent/tt-metal/blob/34722ed36aacbcf4a457e06c793db3e79df4e5b3/ttnn/cpp/ttnn/operations/data_movement/scatter/device/kernels/dataflow/reader_scatter.cpp#L144

Not sure why it's called INVALID (I guess one interpretation would be that the old value is INVALIDated), but it's a bit ambiguous because INVALID is used a lot of times in enums to signify value that would result in a runtime error.

Thanks for looking into this, `std::optional` is definitely misused in TTNN in this case, you would use `std::optional<T>` to extend the set of possible values with one additional value that's not reprensentible in set of `T`, but it never make sense to have `std::optional` with user-defined enum because you can always represent that value explicitly (INVALID in this case). I would definitely add `INVALID` to our own `TTCore_ReduceTypeAttr` (@abogdanovicTT please add a comment about sementics of this value when adding it; same for FB schema). For the op, we can either do `DefaultValuedAttr<TTCore_ReduceTypeAttr, INVALID>`, or plain `TTCore_ReduceTypeAttr`. I'm okay with either of these approaches, might even prefer plain `TTCore_ReduceTypeAttr` where we always set it explicitly in the compiler, it leaves less room for silent errors.

runtime/lib/ttnn/operations/data_movement/scatter.cpp
```diff
@@ -23,9 +23,40 @@ void run(const ::tt::target::ttnn::ScatterOp *op, ProgramContext &context) {
...
+    break;
+  case ::tt::target::ttnn::ScatterReduceType::Max: // Max -> AMAX
+    scatterReduceType =
+        ::ttnn::operations::data_movement::scatter::ScatterReductionType::AMAX;
+    break;
+  case ::tt::target::ttnn::ScatterReduceType::Invalid: // Invalid -> INVALID
+    scatterReduceType = ::ttnn::operations::data_movement::scatter::
+        ScatterReductionType::INVALID;
+    break;
+  default:
```
vmilosevic: :warning: **clang\-diagnostic\-covered\-switch\-default** :warning:
default label in switch which covers all enumeration values

sgholamiTT (commented):
Please update `OpModel` implementation for scatter too:
1. Update the interface call ([here](https://github.com/tenstorrent/tt-mlir/blob/25abcd0744f45f99667c6f0a924cb704448a923b/lib/Dialect/TTNN/Interfaces/TTNNOpModelInterface.cpp#L1072C1-L1075C40) and [here](https://github.com/tenstorrent/tt-mlir/blob/25abcd0744f45f99667c6f0a924cb704448a923b/lib/Dialect/TTNN/Interfaces/TTNNOpModelInterface.cpp#L1092C1-L1095C30)) and pass the `scatter_reduce_type` to the APIs in `TTNNOpModel.h`
2. Then update the APIs in `TTNNOpModel.h` and `TTNNOpModel.cpp` to get this information and pass it to the [metal backend](https://github.com/tenstorrent/tt-metal/blob/b68d39c6ad12e63c69fed095ac7be6d15b5e2831/ttnn/cpp/ttnn/operations/data_movement/scatter/scatter.hpp#L25) ([here](https://github.com/tenstorrent/tt-mlir/blob/25abcd0744f45f99667c6f0a924cb704448a923b/lib/OpModel/TTNN/TTNNOpModel.cpp#L1713C1-L1713C49) and [here](https://github.com/tenstorrent/tt-mlir/blob/25abcd0744f45f99667c6f0a924cb704448a923b/lib/OpModel/TTNN/TTNNOpModel.cpp#L1759)).
3. Finally update the tests [here](https://github.com/tenstorrent/tt-mlir/blob/25abcd0744f45f99667c6f0a924cb704448a923b/test/unittests/OpModel/TTNN/Lib/TestOpModelLib.cpp#L689-L745) and [here](https://github.com/tenstorrent/tt-mlir/blob/25abcd0744f45f99667c6f0a924cb704448a923b/test/unittests/OpModel/TTNN/Op/TestOpModelInterface.cpp#L1273-L1313) to test this new feature.

------

- Note 1: You might need to add some conversion functions [in this file](https://github.com/tenstorrent/tt-mlir/blob/main/include/ttmlir/OpModel/TTNN/Conversion.h) to convert `TTCore_ReduceTypeAttr` to `std::string` (which I believe is what metal expects).
- Note 2: Most of what I said applies to the runtime code too. Please update that as well (for instance, [here](https://github.com/tenstorrent/tt-mlir/blob/25abcd0744f45f99667c6f0a924cb704448a923b/runtime/lib/ttnn/operations/data_movement/scatter.cpp#L13-L32)).

include/ttmlir/OpModel/TTNN/Conversion.h
```diff
@@ -132,6 +132,9 @@ std::optional<::ttnn::DeviceComputeKernelConfig>
 getDeviceComputeKernelConfig(const std::optional<DeviceComputeKernelConfigAttr>
                                  &deviceComputeKernelConfig);
 
+::ttnn::operations::data_movement::scatter::ScatterReductionType
+getScatterReductionType(const ttcore::ReduceTypeAttr &reduceTypeAttr);
+
```
sgholamiTT, line 145: Let's change this API to:
```c++
std::optional<::ttnn::operations::data_movement::scatter::ScatterReductionType>
getScatterReductionType(const std::optional<ttcore::ReduceTypeAttr> &reduceTypeAttr);
```
To make the call site clear.

include/ttmlir/OpModel/TTNN/TTNNOpModel.h
```diff
@@ -468,13 +468,15 @@ struct OpModel<ScatterOp> {
       ttcore::GridAttr deviceGrid, llvm::ArrayRef<int64_t> inputShape,
       TTNNLayoutAttr inputLayout, llvm::ArrayRef<int64_t> indexShape,
       TTNNLayoutAttr indexLayout, llvm::ArrayRef<int64_t> sourceShape,
-      TTNNLayoutAttr sourceLayout, int32_t dim, TTNNLayoutAttr outputLayout);
+      TTNNLayoutAttr sourceLayout, int32_t dim, TTNNLayoutAttr outputLayout,
+      std::optional<ttcore::ReduceTypeAttr> optReduction);
```
sgholamiTT: Nit: Our convention is to keep `outputLayout` as the last argument:
```c++
TTNNLayoutAttr sourceLayout, int32_t dim,
std::optional<ttcore::ReduceTypeAttr> optReduction, TTNNLayoutAttr outputLayout);
```

include/ttmlir/OpModel/TTNN/TTNNOpModel.h
```diff
@@ -468,13 +468,15 @@ struct OpModel<ScatterOp> {
...
 
   static llvm::Expected<size_t>
   getOpRuntime(llvm::ArrayRef<int64_t> inputShape, TTNNLayoutAttr inputLayout,
                llvm::ArrayRef<int64_t> indexShape, TTNNLayoutAttr indexLayout,
                llvm::ArrayRef<int64_t> sourceShape, TTNNLayoutAttr sourceLayout,
-               int32_t dim, TTNNLayoutAttr outputLayout);
+               int32_t dim, TTNNLayoutAttr outputLayout,
+               std::optional<ttcore::ReduceTypeAttr> optReduction);
```
sgholamiTT: Same here.

lib/OpModel/TTNN/Conversion.cpp
```diff
@@ -645,6 +645,30 @@ TTNNLayoutAttr getLayoutAttrFromTensorSpec(MLIRContext *context,
...
+    return ScatterReductionType::MULTIPLY;
+  case ttcore::ReduceType::Invalid:
+    return ScatterReductionType::INVALID;
+  case ttcore::ReduceType::Mean:
+  case ttcore::ReduceType::Std:
+  case ttcore::ReduceType::Var:
+    // These reduction types are not supported by scatter operation
+    throw std::runtime_error("Unsupported ReduceType for scatter operation");
+  }
+  throw std::runtime_error("Invalid ReduceType");
```
sgholamiTT: Following the API change: 
We can return `std::nullopt` here and let the caller decide what to do with that (which seems to be more aligned with other APIs in this file).

lib/OpModel/TTNN/TTNNOpModel.cpp
```diff
@@ -1680,12 +1681,22 @@ llvm::Expected<OpConstraints> OpModel<ScatterOp>::getOpConstraints(
...
+  // Convert optReduction to ScatterReductionType enum
+  std::optional<
+      ::ttnn::operations::data_movement::scatter::ScatterReductionType>
+      optReductionType = ::ttnn::operations::data_movement::scatter::
+          ScatterReductionType::INVALID;
+  if (optReduction.has_value()) {
+    optReductionType =
+        conversion::getScatterReductionType(optReduction.value());
+  }
+
```
sgholamiTT, line 1758: Following the API change:
This complexity will all go away and we can have:
```c++
auto optReductionType = conversion::getScatterReductionType(optReduction);
```

lib/OpModel/TTNN/TTNNOpModel.cpp
```diff
@@ -1680,12 +1681,22 @@ llvm::Expected<OpConstraints> OpModel<ScatterOp>::getOpConstraints(
...
   // Create query closure
   auto scatterOpQuery = [=]() {
     return ::ttnn::graph::query_op_constraints(
         ::ttnn::scatter, device, inputSpec, dim, indexSpec, sourceSpec,
         detail::getNullableMemoryConfig(outputLayout),
-        /* opt_reduction */ std::nullopt);
+        /* opt_reduction_type */ optReductionType);
```
sgholamiTT: Nit: No need for the comment. The variable name clearly describes the meaning:
```c++
detail::getNullableMemoryConfig(outputLayout), optReductionType);
```

lib/OpModel/TTNN/TTNNOpModel.cpp
```diff
@@ -1725,12 +1737,22 @@ llvm::Expected<size_t> OpModel<ScatterOp>::getOpRuntime(
...
+  // Convert optReduction to ScatterReductionType enum
+  std::optional<
+      ::ttnn::operations::data_movement::scatter::ScatterReductionType>
+      optReductionType = ::ttnn::operations::data_movement::scatter::
+          ScatterReductionType::INVALID;
+  if (optReduction.has_value()) {
+    optReductionType =
+        conversion::getScatterReductionType(optReduction.value());
+  }
+
```
sgholamiTT, line 1806: Can also change to:
```c++
auto optReductionType = conversion::getScatterReductionType(optReduction);
```

lib/OpModel/TTNN/TTNNOpModel.cpp
```diff
@@ -1725,12 +1737,22 @@ llvm::Expected<size_t> OpModel<ScatterOp>::getOpRuntime(
...
   // Create query closure
   auto scatterOpRuntimeQuery = [=]() {
     return ::ttnn::graph::query_op_runtime(
         ::ttnn::scatter, device, inputSpec, dim, indexSpec, sourceSpec,
         detail::getNullableMemoryConfig(outputLayout),
-        /* opt_reduction */ std::nullopt);
+        /* opt_reduction_type */ optReductionType);
```
sgholamiTT: Same here.

lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
```diff
@@ -2511,162 +2181,510 @@ class StableHLOToTTIRLogicalAndBitwiseOpConversionPattern
...
+  // DEVICE_TO_DEVICE = 1 : A channel for sending data between
+  // devices.
+  kChannelTypeDeviceToDevice = 1,
+  // DEVICE_TO_HOST = 2 : A channel for sending data from the
+  // device to the host. Can only be used with a Send operation.
+  kChannelTypeDeviceToHost = 2,
+  // HOST_TO_DEVICE = 3 : A channel for sending data from the host to
+  // the device. Can only be used with a Recv operation.
+  kChannelTypeHostToDevice = 3,
+};
```
vmilosevic, line 2446: :warning: **llvm\-else\-after\-return** :warning:
```suggestion
    } if (isa<mlir::stablehlo::MulOp>(op)) {
```

lib/OpModel/TTNN/Conversion.cpp
```diff
@@ -710,6 +710,27 @@ getSDPAProgramConfig(
   return sdpaConfig;
 }
 
+std::optional<std::string> getScatterReductionType(
+    const std::optional<ttcore::ReduceTypeAttr> &reduceTypeAttr) {
+  switch (reduceTypeAttr.value().getValue()) {
+  case ttcore::ReduceType::Sum:
+    return std::optional<std::string>("add");
```
azecevic000: Nit: `std::optional` is usually not constructed explicitly unless the semantics is ambiguous, i.e. `return "add";` is sufficient.

lib/OpModel/TTNN/Conversion.cpp
```diff
@@ -710,6 +710,27 @@ getSDPAProgramConfig(
   return sdpaConfig;
 }
 
+std::optional<std::string> getScatterReductionType(
+    const std::optional<ttcore::ReduceTypeAttr> &reduceTypeAttr) {
+  switch (reduceTypeAttr.value().getValue()) {
```
azecevic000, line 713: This is dangerous because it potentially contains two nullptr dereferences. IMO `std::optional` on `mlir::Attribute` (and its descendants) class doesn't make much sense, because an `Attribute` can be a nullptr, but if you go with this API, you have to decide what meaning do `std::nullopt` (for `std::optional`) and `nullptr` (for `ttcore::ReduceTypeAttr`) have (it's okay for either to be iligeal, but you should at least assert that).

lib/Target/TTNN/TTNNToFlatbuffer.cpp
```diff
@@ -833,6 +833,25 @@ createOp(FlatbufferObjectCache &cache, ReduceScatterOp op) {
...
+  case ttcore::ReduceType::Sum:
+    return ::tt::target::ttnn::ScatterReduceType::Sum;
+  case ttcore::ReduceType::Max:
+    return ::tt::target::ttnn::ScatterReduceType::Max;
+  case ttcore::ReduceType::Min:
+    return ::tt::target::ttnn::ScatterReduceType::Min;
+  case ttcore::ReduceType::Prod:
+    return ::tt::target::ttnn::ScatterReduceType::Prod;
+  default:
+    return ::tt::target::ttnn::ScatterReduceType::Invalid;
```
azecevic000: Since you've already added `ttcore::ReduceType::Invalind` it's better to have:
```cpp
case ttcore::ReduceType::Invalind:
  return ::tt::target::ttnn::ScatterReduceType::Invalid;
default:
  llvm_unreachable("...");
```
This will ensure that if in the future a new reduce type is introduced, it won't be silently mapped to `Invalid`, but will fail explicitly until it's properly mapped, which is much better option.

lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
```diff
@@ -2511,162 +2181,510 @@ class StableHLOToTTIRLogicalAndBitwiseOpConversionPattern
...
+    if (isa<mlir::stablehlo::AddOp>(op)) {
+      return ttcore::ReduceType::Sum;
+    } else if (isa<mlir::stablehlo::MulOp>(op)) {
+      return ttcore::ReduceType::Prod;
+    } else if (isa<mlir::stablehlo::MaxOp>(op)) {
+      return ttcore::ReduceType::Max;
+    } else if (isa<mlir::stablehlo::MinOp>(op)) {
+      return ttcore::ReduceType::Min;
+    } else {
+      return ttcore::ReduceType::Invalid;
```
azecevic000: This is a pretty strong assumption that will fail in a lot of cases. `Invalid` does have a well-defined semantics, so it shouldn't be a catch-all case. Wouldn't `Invalid` map to `stablehlo::return` (or something similar)? The `else` block should be `llvm::Error`, which means we couldn't recognize the type of reduction, which is much better than a silent failure.

lib/OpModel/TTNN/Conversion.cpp
azecevic000: Just to clarify on this comment, both `.value()` and `.getValue()` are potential nullptr dereferences. If either of those is never expected to be null, at least add:
```
assert(reduceTypeAttr && *reduceTypeAttr && "reduceTypeAttr doesn't contain value");
```
before this line (IMO it doesn't make much sense to accept `std::optional` just to `assert` that it always has a value, but it's your call).

azecevicTT (approved):
LGTM, added a clarification for one of the previous comments.

lib/Dialect/TTNN/Transforms/Workarounds/TTNNWorkaroundsPatterns.cpp
```diff
@@ -493,6 +493,10 @@ class TTNNAllReduceWorkarounds : public OpRewritePattern<ttnn::AllReduceOp> {
       return op.emitOpError() << "std is not supported";
     case ttcore::ReduceType::Var:
       return op.emitOpError() << "var is not supported";
+    case ttcore::ReduceType::Prod:
+      return op.emitOpError() << "prod is not supported";
+    case ttcore::ReduceType::Invalid:
+      return op.emitOpError() << "invalid is not supported";
```
jameszianxuTT, line 500: Is `invalid` case not supported because of the overall context of this being inside a AllReduce workaround, so that a no-op reduction would make no sense here? I don't see it being "unsupported" in the same way that the other ReduceTypes are unsupported ; i.e. lack of TTNN support; but rather a pointless case where your AllReduce doesn't come with a reduction type

jameszianxuTT: Also, as a nitpick, `Invalid` seems like a somewhat confusing name for this enum, as it looks like an error rather than a no-op reduction

runtime/lib/ttnn/operations/data_movement/scatter.cpp
```diff
@@ -23,10 +23,33 @@ void run(const ::tt::target::ttnn::ScatterOp *op, ProgramContext &context) {
   std::optional<::ttnn::MemoryConfig> outputMemoryConfig =
       ::tt::runtime::ttnn::utils::createMemoryConfigIfNeeded(
           op->memory_config());
+  // TT Metal currently only supports the following scatter reduction types:
+  // SUM, MULTIPLY, AMIN, AMAX - applied reduction to source tensor
+  // INVALID - copy source to output tensor and apply no reduction(pass
```
jameszianxuTT, line 28: Might be nice to see this comment elsewhere (eg. enum definitions) since it's not obvious from the naming of `Invalid`

lib/Dialect/TTNN/Transforms/Workarounds/TTNNWorkaroundsPatterns.cpp
abogdanovicTT: `ttcore::ReduceType::Invalid` is new in ` ttcore::ReduceType` enum, added for reduction in ScatterOp, so for AllReduceOp this case is not supported

runtime/lib/ttnn/operations/data_movement/scatter.cpp
vladimirjovanovicTT: I agree the name is a bit confusing; scatter flag Invalid is defined in metal so we just followed this naming:
https://github.com/tenstorrent/tt-metal/blob/f080a17769edcc969da9a24526a95f2014f1b6af/ttnn/cpp/ttnn/operations/data_movement/scatter/scatter_enums.hpp#L19

lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
```diff
@@ -2603,162 +2273,510 @@ class StableHLOToTTIRLogicalAndBitwiseOpConversionPattern
...
+         "Region should have exactly one block");
+
+  // Add, Prod, Max, Min are the only supported reduce types for now.
+  // Invalid is default reduction type in TT-Metal, used when we do not do
+  // reduction, only copy source to output.
+  auto &block = region.front();
+  for (Operation &op : block) {
+    if (isa<mlir::stablehlo::AddOp>(op)) {
+      return ttcore::ReduceType::Sum;
+    } else if (isa<mlir::stablehlo::MulOp>(op)) {
```
vmilosevic: :warning: **llvm\-else\-after\-return** :warning:
```suggestion
    } if (isa<mlir::stablehlo::MulOp>(op)) {
```

PR #5827, opened 2025-11-13
Enable matmul sharding

### LINKED ISSUE #5562: Enable sharding for ttnn.matmul

rpavlovicTT:
This change enables matmul/linear ops for sharding in optimizer pass. For matmul/linear ops we can't just pass NULL mem config with sharded input because TTNN actually defaults to DRAM interleaved matmul in such case. Instead, we provide partial memory config with only buffer type (L1) and memory layout (WS, HS, BS) and let TTNN do its part then.

Code cleanup:
- Removed old CheckShardCompatible in favor of new op constraint validation lib
- Fixed interleaved->sharded check and reshard check to use new API
- Now reshard saves all viable config paths instead of first one only
- Keeping old customCheckShardCompatible for testing purposes
- Removed tensorL1UsageCap requirement by constraint validation lib, instead we attach it to the IR and use it as module's attribute

### Ticket
Closes #5562

### DIFF
--- a/lib/Dialect/TTNN/Analysis/ShardSolver.cpp
+++ b/lib/Dialect/TTNN/Analysis/ShardSolver.cpp
@@ -11,12 +11,17 @@
 #include "ttmlir/Dialect/TTNN/Analysis/OpConfigAttrs.h"
 #include "ttmlir/Dialect/TTNN/IR/TTNNOps.h"
 #include "ttmlir/Dialect/TTNN/IR/TTNNOpsAttrs.h"
+#include "ttmlir/Dialect/TTNN/Utils/OptimizerUtils.h"
 #include "ttmlir/Dialect/TTNN/Utils/PassOverrides.h"
+#include "ttmlir/Dialect/TTNN/Utils/Utils.h"
+#include "ttmlir/Dialect/TTNN/Validation/OpConstraintValidation.h"
 #include "ttmlir/Support/Logger.h"
 #include "ttmlir/Utils.h"
 
+#include "mlir/IR/Attributes.h"
 #include "mlir/IR/BuiltinTypes.h"
 #include "mlir/IR/Location.h"
+#include "mlir/IR/Visitors.h"
 #include "mlir/Interfaces/DestinationStyleOpInterface.h"
 #include "mlir/Support/LLVM.h"
 #include "llvm/ADT/DenseSet.h"
@@ -27,8 +32,7 @@
 #include "llvm/Support/ErrorHandling.h"
 #include "llvm/Support/raw_ostream.h"
 
-#include "mlir/IR/Attributes.h"
-#include "mlir/IR/Visitors.h"
+#include <unordered_map>
 #include <utility>
 #include <vector>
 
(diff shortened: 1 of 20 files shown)


lib/Dialect/TTNN/Utils/Utils.cpp
```diff
@@ -19,6 +21,23 @@
 
 namespace mlir::tt::ttnn::utils {
 
+float getTensorL1UsageCap(Operation *op, float defaultValue) {
+  // Walk up to find the module operation that has the attribute
+  Operation *moduleOp = op;
+  while (moduleOp && !isa<ModuleOp>(moduleOp)) {
+    moduleOp = moduleOp->getParentOp();
+  }
```
azecevic000: Nit:
```suggestion
  ModuleOp moduleOp = op->getParentOfType<ModuleOp>();
```

rpavlovicTT:
perf benchmark with this commit https://github.com/tenstorrent/tt-xla/actions/runs/19332181661/job/55299832763

lib/Dialect/TTNN/Transforms/OptimizerPasses/OptimizerPassesWrapper.cpp
```diff
@@ -47,6 +50,13 @@ class OptimizerPassesWrapper
       op_model::SingletonDeviceContext::getInstance().openDevice();
     }
 
+    // Set tensorL1UsageCap as a module attribute so it's accessible to nested
+    // passes without parameter threading.
+    Operation *op = getOperation();
+    OpBuilder builder(op->getContext());
+    op->setAttr(utils::g_TensorL1UsageCapAttrName,
```
odjuricicTT, line 57: This seems like a bit of a hack just to avoid parameter threading. l1usagecap is just a compile time setting and not tied to IR or the module in any way.

Not going to block on this, but my preference would be to stick to passing the param everywhere. Or i'm open to if there is another cleaner solution.

lib/Dialect/TTNN/Analysis/ShardSolver.cpp
```diff
@@ -49,21 +52,96 @@ getUniqueOpSpecificAttrs(const std::vector<OpConfig> &configs) {
   return attrVec;
 }
 
+// Returns unique test configs for Matmul/Linear ops.
```
odjuricicTT: Can the logic for this and the function below be moved to LegalOpLayout / LagalOpConfig analysis?
e.g. not generate the output layouts at all if not needed for some ops there?

lib/Dialect/TTNN/Transforms/OptimizerPasses/OptimizerPassesWrapper.cpp
rpavlovicTT: This is much more MLIR native way of solving it. I like it much more than having to pass single float in dozens of functions and different analysis. It's especially neat as this single wrapper pass sets attribute and then unsets when passes are done.

lib/Dialect/TTNN/Analysis/ShardSolver.cpp
rpavlovicTT: Discussed offline -- we still need layout generation, but will move these functions to utilities.

PR #5842, opened 2025-11-13
GridSelection for row major tensors

anuragsingh-tt:
### What's changed
Refactor out the physical shape calculation when updating the generic and streamlayout ops. When calculating the physical shape to determine the optimal grid in GridSelection, treat row majors as tilized to avoid small shards. Additionally, loosen the assertion in updateStreamLayouts to enable generics to use streams within their own regions.

### DIFF
--- a/lib/Dialect/D2M/Transforms/GridSelection.cpp
+++ b/lib/Dialect/D2M/Transforms/GridSelection.cpp
@@ -138,10 +138,9 @@ static llvm::SmallVector<int64_t> computePhysicalShape(
   if (auto tileType =
           mlir::dyn_cast<ttcore::TileType>(tensorType.getElementType())) {
     tileShape = llvm::to_vector(tileType.getShape());
-  }
-  // Always tile-align when calculating the physical shape, even in the row
-  // major case.
-  if (tileShape.empty()) {
+  } else {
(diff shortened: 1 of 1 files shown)


lib/Dialect/D2M/Transforms/GridSelection.cpp
```diff
@@ -128,6 +128,36 @@ computeGridAwareDimAlignments(ArrayRef<int64_t> logicalShape,
...
+// dimension alignments and then deriving the physical shape (always
+// tile-aligned).
+static llvm::SmallVector<int64_t> computePhysicalShape(
+    ttcore::MetalLayoutAttr layout, mlir::RankedTensorType tensorType,
+    ArrayRef<int64_t> targetSquareGridShape, OpBuilder &builder) {
+  llvm::SmallVector<int64_t> tileShape;
+  if (auto tileType =
+          mlir::dyn_cast<ttcore::TileType>(tensorType.getElementType())) {
+    tileShape = llvm::to_vector(tileType.getShape());
+  }
```
nsmithtt: Nit: I think it'd be more natural to write `} else {` here.

nsmithtt (approved):

PR #5847, opened 2025-11-13
Change runner from 'docker-builder' to 'builder' for temp workaround

jmcgrathTT:
### Ticket
Slack DevEx ticket

### Problem description
There are issues with the runner `forge-builder-5`, but I can't find another runner with the label `docker-builder`.

image

https://github.com/tenstorrent/tt-mlir/actions/runs/19304270673/job/55339782267

I also don't have permissions to change the labels of our other forge builder runners.

image

As a temp solution, I have switched this to a builder box to unblock devs for today.

### DIFF
--- a/.github/workflows/call-build-docker.yml
+++ b/.github/workflows/call-build-docker.yml
@@ -48,7 +48,7 @@ jobs:
   build-image:
     needs: check-if-docker-exist
     if: needs.check-if-docker-exist.outputs.docker-image == ''
-    runs-on: docker-builder
+    runs-on: builder
     outputs:
       docker-image: ${{ steps.build.outputs.docker-image }}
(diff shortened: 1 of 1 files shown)


vvukomanTT:
@jmcgrathTT Hey John, we can close this PR since we have already merged a different quick fix for this #5862.

Note: the PR for a proper fix will be created today 👍

nsumrakTT (changes requested):
Please close this PR.

The docker-builder label is added to prevent multiple docker builds at the same time.
When there are multiple changes in docker file that require docker build in push (on main) all PRs also build the same docker image until push build is finished. Since docker build lasts for >3h, whole CI stops on docker step for hours, while forked/community PRs fail. We had this situation several times and we decided to have single docker builder so that docker is built only once and all other workflows will wait for it.

Issue #5743, opened 2025-11-06
[ttnn.jit] Test multichip support


sohaibnadeemTT:

sohaibnadeemTT:
The tests for eltwise ops with tensors sharded across devices (in different configs) are all passing and merged in #5849

sohaibnadeemTT:
# Test Coverage:

The tests run eltwise ops with tensors sharded across devices in different configs and check jit (with graph capture) output against regular ttnn for:
- all close of each shard
- matching ttnn memory configs of output tensors
- matching shapes shards on each device

This is on a T3K/Quietbox/Llmbox with row dispatch axis used (column dispatch axis fails since mlir runtime tries to use cores being used for dispatch).

### Operations:
- Unary Eltwise Operations
    - **`exp`**
    - **`cosh`** (Composite implemented with exp, neg, add, multiply)
- Binary Eltwise Operations: **`add`**

There's also an interop test that uses jit output as input into a ttnn.sum to check compatibility

### Tensor Shapes
- single tile: (32, 32)
- large tensor: (2048, 2048)
- high aspect ratio: (1024, 32) and (32, 1024)

### Sharding Configs
- ReplicateTensorToMesh: replicate tensor across all devices
- ShardTensorToMesh: shard tensor across dimension 0 or 1 and distribute across all devices
- ShardTensor2dMesh: shard tensor across 2 dimensions and distribute across 2d grid of devices

PR #5849, opened 2025-11-13
Add tests for eltwise ops with mesh tensors in ttnn jit

sohaibnadeemTT:
This also add mesh device fixtures and an option to set mesh_mapper in create_dram_tensor, both for use in the tests.

### Ticket
Link to Github Issue: https://github.com/tenstorrent/tt-mlir/issues/5743

### Checklist

### DIFF
--- a/test/ttnn-jit/conftest.py
+++ b/test/ttnn-jit/conftest.py
@@ -5,6 +5,7 @@
 import pytest
 import ttnn
 import torch
+from loguru import logger
 
 
 @pytest.fixture(scope="module")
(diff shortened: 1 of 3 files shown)


test/ttnn-jit/conftest.py
```diff
@@ -4,10 +4,151 @@
...
 import ttnn
+from loguru import logger
 
 
 @pytest.fixture(scope="module")
 def device():
     d = ttnn.open_device(device_id=0)
     yield d
     ttnn.close_device(d)
+
```
sohaibnadeemTT, line 16: These changes for mesh device fixture are from tt-metal/conftest.py

test/ttnn-jit/conftest.py
```diff
@@ -4,10 +4,151 @@
...
+
+# Reset fabric config to DISABLED if not None, and do nothing otherwise
+# Temporarily require previous state to be passed in as even setting it to DISABLED might be unstable
+# This is to ensure that we don't propagate the instability to the rest of CI
+def reset_fabric(fabric_config):
```
arminaleTT, line 21: Is this a bug, or just a general limitation of fabric?

test/ttnn-jit/test_mesh_tensor_eltwise.py
```diff
@@ -0,0 +1,70 @@
...
+    memory_configs_equal,
+    create_dram_tensor,
+)
+
+
+def add(a, b):
+    return ttnn.add(a, b)
+
+
+@pytest.mark.parametrize("h, w", [(2048, 2048)])
```
arminaleTT: I know this is supposed to be a first PR, but can we have some more shapes here? Please add:

- high aspect ratio
- single tile shard shape

arminaleTT (commented):
Could you also please add:
- A unary test
- A composite op (e.g. multiply accumulate)
- An interop test (JIT op producer, TTNN op consumer)

If they fail, just xfail them with an issue

test/ttnn-jit/conftest.py
sohaibnadeemTT: This comment is copied over from metal too so I don't know the full story but digging into the PR, it looks like this was added back in March when they did have to pass in the existing fabric_config even for resetting since fabric was unstable; so yeah it appears to be a limitation for now.

test/ttnn-jit/test_mesh_tensor_eltwise.py
sohaibnadeemTT: added

sohaibnadeemTT:
@arminaleTT Added the following:

- Changed the tests for T3K/quietbox, so I can test ShardTensor2dMesh too
- Add a check to compare shard shapes between jit and ttnn outputs
- Added unary eltwise (exp),  composite (cosh), and interop (sum after main op)
- added additional tensor shapes (high aspect ratios, single tile)

sohaibnadeemTT:
@arminaleTT  Also I've changed the tests to only check with graph_capture=True to keep the test permutations small (it's already 144 (taking ~4 minutes) = 4 shapes *  4 ops * 9 sharding types)

arminaleTT (approved):

PR #5850, opened 2025-11-13
Fix macOS build

nsmithtt:
Instead of searching for the gtest package (which is typically provided by LLVM) just check for its existence and early out if not available.


nsmithtt:
Not a real issue, closing.

PR #5851, opened 2025-11-13
Add pass for register custom sharding rule

sshonTT:
### Ticket
N/A

### Problem description
We don't have sharding rule for custom ops, so we cannot run it on multi device

### What's changed
This pass registers custom sharding rules in the current MLIRContext so that Shardy can correctly propagate shardings through operations that do not have built-in sharding rules in Shardy. Starting from stablehlo.custom_call op, we can define any newly introduced op by our own rule.

### Checklist

### DIFF
--- a/lib/Dialect/StableHLO/Transforms/RegisterCustomShardingRule.cpp
+++ b/lib/Dialect/StableHLO/Transforms/RegisterCustomShardingRule.cpp
@@ -83,9 +83,9 @@ private:
   getCustomCallShardingRule(mlir::stablehlo::CustomCallOp op) const {
     llvm::StringRef target = op.getCallTargetName();
 
-    auto it = customCallShardingRules.find(target);
-    if (it != customCallShardingRules.end()) {
-      return it->second(op);
+    auto shardOpFunc = customCallShardingRules.lookup(target);
+    if (shardOpFunc) {
+      return shardOpFunc(op);
     }
 
     op.getOperation()->emitWarning()


lib/Dialect/StableHLO/Transforms/RegisterCustomShardingRule.cpp
```diff
@@ -0,0 +1,124 @@
...
+  bool shouldKeepOutputShardingsDivisible(mlir::Operation *) const {
+    return true;
+  }
+
+private:
+  mlir::sdy::OpShardingRuleAttr
+  getCustomCallShardingRule(mlir::stablehlo::CustomCallOp op) const {
+    llvm::StringRef target = op.getCallTargetName();
+
+    auto it = customCallShardingRules.find(target);
```
anuragsingh-tt: You can use `llvm::DenseMap::lookup` here to get the rule directly.

test/ttmlir/Dialect/StableHLO/decouple_constant_fanout/broadcast_in_dim_clone.mlir
```diff
@@ -1,8 +1,4 @@
 // REQUIRES: stablehlo
-// This file incorporates work covered by the following copyright and permission notice:
```
sshonTT, line 2: I know this isn’t directly related to this PR, but I included it since it’s a simple change.

AleksKnezevic:
This is awesome @sshonTT. Do we want to use a custom or composite of for SDPA? You can probably get this in for paged attention? The sharding rules will be the same.

sshonTT:
@AleksKnezevic I think using both should be fine, since they seem to map to a single TTNN op anyway. Of course, we’ll need to check the performance to be sure.
And as you mentioned, if the same sharding rules apply, we should be able to map them in the same way for paged attention as well.

lib/Dialect/StableHLO/Transforms/RegisterCustomShardingRule.cpp
sshonTT: Changed it to `lookup`

sshonTT:
@AleksKnezevic I thought about it again, and I think the custom call approach might actually give us more benefits unless head sharding is confirmed. With a composite op, if sharding propagation determines that a CCL operation needs to be inserted inside the composite function, the whole thing ends up running as flattened ops.
On the other hand, with a custom op, the CCL operations would be inserted before the custom call for replication, but the core computation can still run as a single TTNN op.

AleksKnezevic:
@sshonTT, that makes sense. If we find the way we handle composite ops doesn't work for some composites, is it possible to handle them through this flow (i.e. if the composite op is in some table then apply a custom rule to it, otherwise flatten and propagate)? 

This is not a change I'm requesting now, just something to think about.

sshonTT:
@AleksKnezevic Yes, I think that would be possible. Adding new rules isn’t difficult with the current setup, and we would just need to register exceptions for the cases where flattening shouldn’t apply.

tapspatel (approved):

test/ttmlir/Dialect/StableHLO/decouple_constant_fanout/broadcast_in_dim_clone.mlir
mmanzoorTT: IIRC this copyright notice was added as this test (or part of it) was copied from some other project. I think @tapspatel had added this. I think, we should keep this copyright if it is still the case.

mmanzoorTT (approved):

env/patches/shardy.patch
```diff
@@ -831,3 +831,16 @@ index 47ce53e..63b2424 100644
        .Case<stablehlo::BitcastConvertOp>(
            [](stablehlo::BitcastConvertOp bitcastConvert) {
              ArrayRef<int64_t> inShape =
+@@ -685,6 +716,12 @@ OpShardingRuleAttr createOpShardingRule(Operation* op,
+                   /*isBlocked=*/usedByRngBitGenerator)
+               .build();
+         }
++        // Check if the custom call implements the ShardingRuleOpInterface.
```
sshonTT, line 838: This seems upstream-able so I updated a patch to Shardy (https://github.com/openxla/shardy/pull/885)

test/ttmlir/Dialect/StableHLO/decouple_constant_fanout/broadcast_in_dim_clone.mlir
sshonTT: This test file is something I added in an earlier commit, and it looks like this copyright line got carried over while copy-pasting. @tapspatel, to be sure, do you think we should remove it? I don’t see the same notice in other test files.

sshonTT:
@vwellsTT, @nsmithtt Could you please take a look as code owners?

nsmithtt (approved):
Approving cmake changes!

PR #5852, opened 2025-11-13
Add explicit push/pop release operations for circular buffers

zoecarver:
The other half of tt-lang's [Add explicit push/pop release operations for circular buffers](https://github.com/tenstorrent/tt-lang/pull/26)

### Ticket
N/A

### Problem description
Circular buffers previously used implicit release semantics where resources were automatically freed at block end. This made synchronization points less clear and didn't match the hardware semantics as closely as it could. Users had to understand that `reserve()` implicitly pushed at the end, which wasn't obvious from the API. There are still cases for the implicit behavior, but some users would prefer it be explicit. 

### What's changed
Added explicit `d2m.push` and `d2m.pop` operations for circular buffer release semantics. Producers now call `push()` after `reserve()` to signal data is ready, and consumers call `pop()` after `wait()` to signal consumption is complete. The conversion pass detects explicit release ops and skips automatic insertion, while existing IR without explicit releases still works with automatic insertion.

  Changes:
  - Added `d2m.push` and `d2m.pop` operations to D2M dialect
  - Created `D2MCBReleaseOpRewriter` template for lowering to ttkernel ops
  - Modified `D2MCBOpRewriter` to skip automatic release when explicit ops present

### DIFF
--- a/test/ttmlir/Conversion/D2MToTTKernel/explicit_push_pop.mlir
+++ b/test/ttmlir/Conversion/D2MToTTKernel/explicit_push_pop.mlir
@@ -10,11 +10,11 @@
 module {
   // Test explicit push/pop for producer/consumer pattern
   // CHECK-LABEL: func.func private @compute_kernel
-  func.func @test_explicit_push_pop(%in0: memref<1x1x1x1x!ttype_f32, #ttcore.shard<4096x4096>, #l1_>,
-                                    %out: memref<1x1x1x1x!ttype_f32, #ttcore.shard<4096x4096>, #l1_>) {
+  func.func @test_explicit_push_pop(%in0: memref<1x1x1x1x!ttype_f32, #ttcore.shard<4096x4096, 1>, #l1_>,
+                                    %out: memref<1x1x1x1x!ttype_f32, #ttcore.shard<4096x4096, 1>, #l1_>) {
     d2m.generic {block_factors = [1, 1], grid = #ttcore.grid<1x1>, indexing_maps = [#map_, #map_], iterator_types = [#parallel_, #parallel_], threads = [#d2m.thread<compute>]}
-        ins(%in0 : memref<1x1x1x1x!ttype_f32, #ttcore.shard<4096x4096>, #l1_>)
-        outs(%out : memref<1x1x1x1x!ttype_f32, #ttcore.shard<4096x4096>, #l1_>)  {
+        ins(%in0 : memref<1x1x1x1x!ttype_f32, #ttcore.shard<4096x4096, 1>, #l1_>)
+        outs(%out : memref<1x1x1x1x!ttype_f32, #ttcore.shard<4096x4096, 1>, #l1_>)  {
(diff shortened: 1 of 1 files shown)


include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
```diff
@@ -1202,6 +1204,66 @@ def D2M_ReserveOp : D2M_CBOp<"reserve"> {
   }];
 }
 
+def D2M_PushOp : D2M_GenericRegionOp<"push",
+  [ MemoryEffects<[MemRead, MemWrite]>
+  , DeclareOpInterfaceMethods<BufferizableOpInterface>
```
nsmithtt: This and one below should be:
```
    DeclareOpInterfaceMethods<BufferizableOpInterface, [
      "bufferizesToMemoryRead",
      "bufferizesToMemoryWrite",
      "bufferize",
      "getAliasingValues",
      "getBufferType"
    ]>
```

lib/Dialect/D2M/IR/D2MOps.cpp
```diff
@@ -1595,9 +1596,39 @@ mlir::LogicalResult d2m::GenericOp::bufferize(
...
+        }
+      }
+
+      // Memref-semantic ops use the memref arg directly
+      for (OpOperand *use : memrefSemanticUses) {
+        rewriter.modifyOpInPlace(use->getOwner(), [&]() {
+          use->set(newArg);
+        });
+      }
+
```
nsmithtt: I think we shouldn't be doing this inside of generic op bufferization and instead let this be handled by implementing the bufferization interface for push/pop.

zoecarver: We don't want `to_tensor` ops on push/pop, so I had to add this logic here. Not sure if this is better than what we had before (previous version of this PR)...

zoecarver: The alternative is to have wait and reserve's bufferization handle the `to_tensor` ops, but then they need to reach up into their def which is pretty terrible. WDYT? Am I missing something?

nsmithtt: Yeah I think this is because your implementation of `bufferize` is empty.  That should be implemented to emit the memref version of the op.

nsmithtt: `to_tensor` is an ephemeral op that's only used as a placeholder during the bufferization pass itself so that the IR can be in a legal state throughout the pass.

zoecarver: OK I get it now, we are adding to_tensor here and then bufferizing with to_buffer later. Will update push and pop to do this as well. Sorry, I was super confused about what was getting to_tensor and thought push/pop didn't need it because they don't return anything.

nsmithtt: Np! There is a subtle art to bufferization that took me (more than) a few op implementations for it to click 😅

PR #5875, opened 2025-11-14
[D2M/TTNN] allocate cb at address

vtangTT:
### Ticket
Link to Github Issue

### Problem description

### What's changed

### Checklist


include/ttmlir/Dialect/TTNN/IR/TTNNOpsAttrs.td
```diff
@@ -751,11 +760,26 @@ def TTNN_KernelCBAttr: TTNN_Attr<"KernelCB", "kernel_cb"> {
   let parameters = (ins "uint32_t":$total_size,
                         TTNN_CoreRangeSetAttr:$core_ranges,
                         ArrayRefParameter<"KernelCBFormatAttr">:$formats,
-                        OptionalParameter<"KernelCBGlobalBufferAddressOfTensorAttr">:$buffer);
+                        OptionalParameter<"mlir::Attribute">:$buffer);
```
vroubtsovTT, line 764: I don't understand this change... why lower the apparent type safety here?

include/ttmlir/Dialect/TTNN/IR/TTNNOpsAttrs.td
```diff
@@ -742,6 +742,15 @@ def TTNN_KernelCBGlobalBufferAddressOfTensorAttr : TTNN_Attr<"KernelCBGlobalBuff
   let assemblyFormat = "`<` $tensor_operand_index `>`";
 }
 
+def TTNN_KernelCBAddressAttr : TTNN_Attr<"KernelCBAddress", "kernel_cb_address"> {
+  let summary = "Statically allocatedCB address";
+  let description = [{
+    TTNN Generic CB Descriptor argument for statically allocated CB address.
+  }];
+  let parameters = (ins "uint32_t":$cb_address);
```
vroubtsovTT: Should this not be `int64_t`? The allocator certainly treats "address" as a 64-bit value and I see int64_ts used in a few other existing ops... but maybe it's the allocator that's wrong.

lib/Conversion/D2MToTTNN/D2MToTTNN.cpp
```diff
@@ -174,19 +174,38 @@ class D2MGenericRewriter : public OpConversionPattern<d2m::GenericOp> {
...
+        cbDescriptor = ttnn::KernelCBAttr::get(
+            ctx, numPages * pageSize, coreRangeSet, {cbFormat}, cbAddress);
+      } else {
+        // If not a memref or not streamed
+        ttnn::KernelCBGlobalBufferAddressOfTensorAttr globalCBIndexOfTensor;
+
         TT_assertv(ttcore::getMemorySpace(cb_memref) ==
                        ttcore::MemorySpace::DeviceL1,
                    "Can only alias L1 buffers.");
         globalCBIndexOfTensor =
```
vroubtsovTT: Is there a reason to separate a `globalCBIndexOfTensor` declaration and assignment by that assert? Could be simply
```
TT_assertv(ttcore::...
auto globalCBIndexOfTensor = ttnn::KernelCBGlobalBufferAddressOfTensorAttr::get(ctx, i);
```

lib/Target/TTNN/TTNNToFlatbuffer.cpp
```diff
@@ -2297,18 +2297,33 @@ createOp(FlatbufferObjectCache &cache, GenericOp op) {
...
     }
-    ::flatbuffers::Offset<::tt::target::ttnn::KernelGlobalCBIndexOfTensor>
-        buffer;
-    if (kernelCbAttr.getBuffer()) {
-      buffer = ::tt::target::ttnn::CreateKernelGlobalCBIndexOfTensor(
-          *cache.fbb, kernelCbAttr.getBuffer().getTensorOperandIndex());
+    ::tt::target::ttnn::KernelGlobalCBOrAddress bufferType =
+        ::tt::target::ttnn::KernelGlobalCBOrAddress::NONE;
+    ::flatbuffers::Offset<void> buffer;
+    if (auto bufferAttrBase = kernelCbAttr.getBuffer(); bufferAttrBase) {
```
vroubtsovTT: the `; bufferAttrBase` part isn't needed here, right?

runtime/lib/ttnn/operations/generic/generic_op.cpp
```diff
@@ -270,9 +300,16 @@ void overrideArgs(
...
 
-    // Not all CBs have a backing L1 buffer.
-    if (cbDesc->buffer()) {
-      cb.buffer = ioTensors[cbDesc->buffer()->tensor_operand_index()].buffer();
+    // Not all CBs have a backing L1 buffer. Statically allocated CBs are
+    // compile time only.
+    if (cbDesc->buffer_type() == ::tt::target::ttnn::KernelGlobalCBOrAddress::
+                                     KernelGlobalCBIndexOfTensor) {
+      if (const auto *tensorBuffer =
+              cbDesc->buffer_as_KernelGlobalCBIndexOfTensor();
```
vroubtsovTT, line 307: confused about nested if-checks here... if the buffer type is `KernelGlobalCBIndexOfTensor`, can `cbDesc->buffer_as_KernelGlobalCBIndexOfTensor()` ever return null?

test/ttmlir/Conversion/D2MToTTNN/sanity.mlir
```diff
@@ -39,15 +39,15 @@ module {
 
     // CHECK-NOT: memref.alloc()
     // CHECK-NOT: d2m.stream_layout
-    %storage_in = memref.alloc() : memref<1x1x1x1x!ttcore.tile<32x32, f32>, #ttcore.shard<4096x4096>, #ttcore.memory_space<l1>>
+    %storage_in = memref.alloc() {address = 1024 : i64, alignment = 32 : i64} : memref<1x1x1x1x!ttcore.tile<32x32, f32>, #ttcore.shard<4096x4096>, #ttcore.memory_space<l1>>
```
vroubtsovTT: see, these addresses are i64s...

vroubtsovTT (changes requested):

include/ttmlir/Dialect/TTNN/IR/TTNNOpsAttrs.td
vtangTT: `$buffer` should either be a `KernelCBGlobalBufferAddressOfTensorAttr` or `KernelCBAddressAttr`. Unsure how I would be able to express this in tablegen..

Type safety should still be insured with the `verify` check in `lib/Dialect/TTNN/IR/TTNNOpsAttrs.cpp`

include/ttmlir/Dialect/TTNN/IR/TTNNOpsAttrs.td
vtangTT: No, I think you're right and it shoudl be `int64_t`. I didn't give much thought and just defaulted it to `uint32_t`

lib/Conversion/D2MToTTNN/D2MToTTNN.cpp
vtangTT: Oops, yea this was sloppy, ty 🙏

lib/Target/TTNN/TTNNToFlatbuffer.cpp
vtangTT: yes not needed. ty 🙏

runtime/lib/ttnn/operations/generic/generic_op.cpp
vtangTT: If `buffer_type` is set, then it should be guaranteed that it should never return null. So under normal circumstances the if statement is redundant..

I'm ok with leaving it in just as extra defense measure, but also fine with taking it out.

include/ttmlir/Dialect/TTNN/IR/TTNNOpsAttrs.td
vroubtsovTT: We want tablegen to be as declarative as possible, and a generic `Attribute` seems like it works against that... I think we have this either/or situation in a few places... `AnyAttrOf<[My_Attr1, My_Attr2]>` I think is the tablegen trick for expressing a union of attr types.

Can something like `OptionalAttr<AnyAttrOf<[KernelCBGlobalBufferAddressOfTensorAttr, KernelCBAddressAttr]>>` work?  I found one example in the llvm codebase that seems to validate this syntax:
```
def AMDGPU_DPPOp : AMDGPU_Op<"dpp", [SameTypeOperands, AllTypesMatch<["result", "old", "src"]>]>,
  Arguments<(ins AnyType:$old,
                  AnyType:$src,
                  AMDGPU_DPPPermAttr:$kind,
                  OptionalAttr<AnyAttrOf<[I32Attr, ArrayAttr, UnitAttr]>>:$permArgument,
                  DefaultValuedAttr<I32Attr, "0xf">:$row_mask,
...
```

vroubtsovTT: After discussing elsewhere, let's stick with `OptionalParameter<"mlir::Attribute">` since we've already done it within similar TTNN attr types, but to help keep things more "declarative" add a comment about the allowed types a la
```
OptionalParameter<"mlir::Attribute", "CB buffer address, one of: KernelCBGlobalBufferAddressOfTensorAttr, KernelCBAttr">:$buffer)
```

PR #5879, opened 2025-11-16
[D2M] Implement graph coloring-based DST register allocation

brnorris03:
### Ticket
Closes https://github.com/tenstorrent/tt-mlir/issues/5923

## Problem

The existing destination register allocator `D2MInsertDstRegisterAccess` performs a number of separable tasks (in addition to assigning DST slices) and implements a linear scan DST allocation strategy, requiring extra logic to identify reuse, potentially leading to inefficient register usage that can exceed hardware limits. Experimenting with different allocation strategies is difficult in the current design.

## Changes

Refactors the implementation to make it possible to use different DST allocation strategies. Introduces graph coloring-based DST (destination register) allocation into the existing `D2MInsertDstRegisterAccess` pass, which can reduce DST usage by reusing slices for non-interfering values. It also eliminates the need for subsequent "DST fusion" transformation. Chaitin-Briggs graph coloring is now the default strategy.

### Architecture

```
                 DstAnalysis (abstract base)
                        ^
                        | (factory functions)
                        |
        +-----------------------+
        |                       |
createBasicDstAnalysis()        |
        |               createGreedyDstAnalysis()
        |               createChaitinBriggsDstAnalysis()
        v                       |
DstAnalysisBasic        DstAnalysisGraphColoring
(no graph coloring)             |
                                | uses
                                v
                      ColoringStrategy (abstract base)
                           ^       ^
               implements /         \ implements
                         /           \
         ChaitinBriggsColoring   GreedyColoring


InterferenceGraph (namespace with utility functions):
  - buildIndexGraphFromDstOperations()
  - computeChromatic_Lowerbound()
```

All strategies rely on the existing static `collectDstAccesses` function to create a `CopyInfoMap` structure for the transformation (it also collects guard dimensions for conditional reloads and identifies intermediate results between compute ops). The difference between strategies is how the slice indices in `copyInfos` get populated:
- Basic (existing pass functionality): `getSlice()` computes indices on-the-fly during collectDstAccesses
- Graph coloring: `getSlice()` looks up pre-computed indices from `analysisResult.operationSlices`.

### Implementation Details

The core of the graph-coloring implementation is the construction of an accurate interference graph (each node represents a value, and an edge between two nodes indicates that the corresponding values are simultaneously live). This requires tracking several details beyond standard liveness. 

The conditions for adding an interference edge are:
- Two values are simultaneously live (overlapping lifetimes).
- Two input operands of the same operation.
- In-place operation: between input and result values.
- Values produced by compute-like operations (primarily binary ops) that may be reused.

### Strategy Selection

Three strategies available via pass option `--d2m-insert-dst-register-access='allocation-strategy=<strategy>'` or pipeline option `--ttmetal-pipeline="dst-strategy=<strategy>"`.

1. `chaitin` (default): Chaitin-Briggs simplification-based coloring
2. `greedy`: Sequential first-available-color assignment (ordered by degree)
3. `basic`: Original sequential allocation (no reuse)

#### Fallback to Linear Allocation

When graph coloring strategies are selected, certain operations fall back to linear (sequential) allocation because they aren't tracked by `DstAnalysis`. This occurs because the analysis identifies DST accesses based on explicit `AffineLoadOp` and `AffineStoreOp` operations in the IR, so operations without these explicit markers don't get pre-assigned slots:

- `TileBcastOp` intermediates: The broadcast op itself isn't in the pre-computed slot map; the pass traces through to the underlying `AffineLoadOp` which has an assigned slot.
- Single-use compute intermediates: Values passed directly between compute ops without explicit stores may not have pre-assigned slots in the analysis.
- `tile_matmul` inside `linalg.generic`: The graph coloring analysis runs before matmul lowering (linalg->affine conversion), so matmul regions use BlockArguments instead of `AffineLoadOp`. Since the analysis cannot (currently) trace block arguments back to load operations, matmul DST accesses are not recorded and fall back to linear allocation. This will be addressed in a future PR.

Future work: Extend `DstAnalysis` to cover all DST accesses so every operation has a pre-assigned slot, eliminating the need for fallback allocation. The next steps would be 
to more fully decouple the analysis from the transformation, reducing some redundant 
traversals (to identify/collect DST accesses). 

## Tests

- All existing `insert-dst-register*.mlir` lit tests pass with all three strategies (parameterized tests).
- Several new tests added based on regressions observed in pytest runs.
- New unit (google) tests for `DstCapacityAnalysis` and `GraphColoringStrategy`.
- Updated the pytest configuration to generate new tests based on the pytest CLI `--dst-allocation-strategy` option (if not specified, only default settings are used as before). E.g., to test all strategies:
```pytest -v --dst-allocation-strategy=greedy,basic,chaitin test/python/golden```
- Created `.github/test_scripts/d2m_tests.sh` and added a test config that calls it. 
  It runs the golden pytests with all three DST strategies as described above.

### DIFF
--- a/lib/Dialect/D2M/Analysis/DstAnalysisGraphColoring.cpp
+++ b/lib/Dialect/D2M/Analysis/DstAnalysisGraphColoring.cpp
@@ -40,7 +40,7 @@ identifyDstAccesses(mlir::Operation *op, mlir::Region &region) {
   int64_t index = 0;
 
   region.walk([&](mlir::Operation *operation) {
-    // Check if this is a D2M compute operation with DST interface
+    // Check if this is a D2M compute operation with DST interface.
     auto computeOp =
         llvm::dyn_cast<OperandLoadStoreRegisterOpInterface>(operation);
     if (!computeOp) {
@@ -50,8 +50,7 @@ identifyDstAccesses(mlir::Operation *op, mlir::Region &region) {
     auto operandsLoadFromDst = computeOp.getOperandsLoadFromDstRegister();
 
     // For each operand that loads from DST, find the corresponding load
-    // operation. This traces through intermediate compute ops such as
-    // tile_bcast.
+    // operation. This traces through intermediate ops such as tile_bcast.
     for (int64_t operandIdx : operandsLoadFromDst) {
       mlir::Value operand = operation->getOperand(operandIdx);
 
@@ -75,9 +74,9 @@ identifyDstAccesses(mlir::Operation *op, mlir::Region &region) {
       }
     }
 
-    // Check if the result store needs a separate DST slice.
+    // Check if this operation's result store operationneeds a separate DST
+    // slice.
     if (!computeOp.getDstRegInPlace()) {
-      // Find the store operation that stores this operation's result.
       for (mlir::Operation *user : operation->getUsers()) {
         if (auto storeOp = llvm::dyn_cast<mlir::affine::AffineStoreOp>(user)) {
           if (seenOps.insert(storeOp.getOperation()).second) {
@@ -117,7 +116,7 @@ public:
           continue;
         }
 
-        // Store identified accesses for reuse by transformation passes
+        // Store identified accesses for reuse by transformation passes.
         result.dstAccesses.append(dstAccesses.begin(), dstAccesses.end());
 
         // Build interference graph and coalescing constraints.
@@ -210,20 +209,16 @@ public:
 
         // Propagate colors from contracted graph back to original nodes.
         std::vector<unsigned> coloring(dstAccesses.size());
+        unsigned colorUsed = 0; // at least one color is used.
         for (size_t i = 0; i < dstAccesses.size(); ++i) {
           size_t leader = nodeToLeader[i];
           size_t contractedIdx = leaderToContracted[leader];
           coloring[i] = contractedColoring[contractedIdx];
+          colorUsed = std::max(colorUsed, coloring[i] + 1);
         }
         normalizeColorOrder(coloring);
 
-        // Find actual number of colors used.
-        unsigned colorsUsed = 0;
-        for (unsigned color : coloring) {
-          colorsUsed = std::max(colorsUsed, color + 1);
-        }
-
-        maxSlicesNeeded = std::max(maxSlicesNeeded, colorsUsed);
+        maxSlicesNeeded = std::max(maxSlicesNeeded, colorUsed);
 
         // Record per-operation breakdown.
         for (size_t i = 0; i < dstAccesses.size(); ++i) {
--- a/lib/Dialect/D2M/Analysis/GraphColoringStrategy.cpp
+++ b/lib/Dialect/D2M/Analysis/GraphColoringStrategy.cpp
@@ -26,8 +26,10 @@ mlir::LogicalResult ChaitinBriggsColoring::colorGraph(
   std::vector<size_t> simplifyStack;
   std::vector<bool> removed(numNodes, false);
 
-  while (true) {
-    bool found = false;
+  // Repeatedly find and remove nodes with degree < numColors until none remain.
+  bool foundNode = true;
+  while (foundNode) {
+    foundNode = false;
     for (size_t node = 0; node < numNodes; ++node) {
       if (removed[node]) {
         continue;
@@ -44,15 +46,11 @@ mlir::LogicalResult ChaitinBriggsColoring::colorGraph(
       if (degree < numColors) {
         simplifyStack.push_back(node);
         removed[node] = true;
-        found = true;
+        foundNode = true;
         break;
       }
-    }
-
-    if (!found) {
-      break;
-    }
-  }
+    } // End of inner loop.
+  } // End of outer loop.
 
   // Selection phase: assign colors by popping from stack.
   while (!simplifyStack.empty()) {

### CODE REVIEW

File: lib/Dialect/D2M/Transforms/GraphColoringStrategy.cpp [hunk 1]
```diff
@@ -0,0 +1,202 @@
...
+    }
+
+    coloring[v] = color;
+  }
+
+  // Handle nodes with degree >= K (spill nodes).
+  unsigned nextColor = 0;
+  for (auto v : nodes) {
+    if (!coloring.count(v)) {
+      coloring[v] = nextColor % numColors;
```
vmilosevic: :warning: **clang\-analyzer\-core\.DivideZero** :warning:
Division by zero

lib/Dialect/D2M/IR/D2MOps.cpp
```diff
@@ -25,6 +25,39 @@
 
 namespace mlir::tt::d2m {
 
+LogicalResult AcquireDstOp::verify() {
+  // TODO: The verifier should be smarter about checking for release_dst across
```
missing username/bug in TODO
```suggestion
  // TODO(nsmith): The verifier should be smarter about checking for release_dst across
```

test/unittests/D2M/TestDstCapacityAnalysis.cpp
```diff
@@ -0,0 +1,465 @@
...
+
+  auto funcOp =
+      module->lookupSymbol<func::FuncOp>("test_f16_capacity_exceeded");
+  ASSERT_TRUE(funcOp);
+
+  // Verify the analysis computes the correct capacity (16 for f16 with fullSyncEn=true)
+  DstCapacityAnalysis analysis(funcOp, /*fullSyncEn=*/true);
+  EXPECT_EQ(analysis.getMinDstCapacity(), 16u);
+
+  // TODO: Add integration test using PassManager to verify the pass fails
```
vmilosevic: :warning: **google\-readability\-todo** :warning:
missing username/bug in TODO
```suggestion
  // TODO(nsmith): Add integration test using PassManager to verify the pass fails
```

vmilosevic (commented):

include/ttmlir/Dialect/D2M/Analysis/DstAnalysis.h
```diff
@@ -0,0 +1,72 @@
+// SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#ifndef TTMLIR_DIALECT_D2M_ANALYSIS_DSTREQUIREMENTANALYSIS_H
+#define TTMLIR_DIALECT_D2M_ANALYSIS_DSTREQUIREMENTANALYSIS_H
```
vmilosevic: :warning: **llvm\-header\-guard** :warning:
```suggestion
#ifndef TTMLIR_DIALECT_D2M_ANALYSIS_DSTANALYSIS_H
#define TTMLIR_DIALECT_D2M_ANALYSIS_DSTANALYSIS_H
```

include/ttmlir/Dialect/D2M/Analysis/DstAnalysis.h
```diff
@@ -0,0 +1,72 @@
...
+  /// Get human-readable name of this analysis strategy.
+  ///
+  /// Used for diagnostics, debugging, and command-line option descriptions.
+  /// \return Strategy name (e.g., "basic", "graph-coloring", "greedy").
+  virtual llvm::StringRef getStrategyName() const = 0;
+};
+
+} // namespace mlir::tt::d2m
+
+#endif // TTMLIR_DIALECT_D2M_ANALYSIS_DSTREQUIREMENTANALYSIS_H
```
vmilosevic: :warning: **llvm\-header\-guard** :warning:
```suggestion
#endif // TTMLIR_DIALECT_D2M_ANALYSIS_DSTANALYSIS_H
```

vmilosevic (commented):

include/ttmlir/Dialect/D2M/Analysis/DstAnalysis.h
```diff
@@ -0,0 +1,72 @@
...
```
vmilosevic: :warning: **llvm\-header\-guard** :warning:
\#endif for a header guard should reference the guard macro in a comment
```suggestion
#endif // TTMLIR_DIALECT_D2M_ANALYSIS_DSTANALYSIS_H
```

include/ttmlir/Dialect/D2M/Analysis/DstAnalysis.h
vmilosevic: :warning: **llvm\-header\-guard** :warning:
```suggestion
#ifndef TTMLIR_DIALECT_D2M_ANALYSIS_DSTANALYSIS_H
#define TTMLIR_DIALECT_D2M_ANALYSIS_DSTANALYSIS_H
```

lib/Dialect/D2M/Analysis/DstAnalysisGraphColoring.cpp
```diff
@@ -0,0 +1,172 @@
...
+    // Get operand indices that must be loaded from DST register
+    auto operandsLoadFromDst = computeOp.getOperandsLoadFromDstRegister();
+
+    // For each operand that loads from DST, find the corresponding load
+    // operation
+    for (int64_t operandIdx : operandsLoadFromDst) {
+      mlir::Value operand = operation->getOperand(operandIdx);
+
+      // Trace back through the SSA chain to find the affine.load
+      if (auto definingOp = operand.getDefiningOp()) {
```
vmilosevic: :warning: **llvm\-qualified\-auto** :warning:
`` auto definingOp `` can be declared as `` auto *definingOp ``
```suggestion
      if (auto *definingOp = operand.getDefiningOp()) {
```

vmilosevic (commented):

lib/Dialect/D2M/Transforms/InsertDstRegisterGC.cpp
```diff
@@ -0,0 +1,1257 @@
...
+
+namespace {
+
+// Loop context information for DST index computation.
+// Tracks the loop nest containing a DST access and its iteration space.
+struct LoopContext {
+  llvm::SmallVector<mlir::affine::AffineForOp, 4>
+      loopNest;                             // Outermost to innermost.
+  llvm::SmallVector<int64_t, 4> tripCounts; // Trip count for each loop.
+  int64_t totalIterations;                  // Product of all trip counts.
```
vmilosevic: :warning: **cppcoreguidelines\-use\-default\-member\-init** :warning:
use default member initializer for `` totalIterations ``
```suggestion
  int64_t totalIterations{1};                  // Product of all trip counts.
```

lib/Dialect/D2M/Transforms/InsertDstRegisterGC.cpp
```diff
@@ -0,0 +1,1257 @@
...
+
+  LoopContext() : totalIterations(1) {}
```
vmilosevic: :warning: **cppcoreguidelines\-use\-default\-member\-init** :warning:
use default member initializer for `` totalIterations ``
```suggestion
  LoopContext() : {}
```

.github/test_scripts/d2m_tests.sh
```diff
@@ -0,0 +1,16 @@
...
+
+if [ -n "$REQUIREMENTS" ]; then
+    eval "pip install $REQUIREMENTS"
+fi
+
+for strategy in "basic" "greedy" "chaitin-briggs"; do
+    echo -e "\n----------------------------------------------------------------------------------"
+    echo "Running D2M Python tests $@ with DST_ALLOCATION_STRATEGY=$strategy"
+    DST_ALLOCATION_STRATEGY="$strategy" pytest -v --junit-xml="$TEST_REPORT_PATH" "$@"
+done
```
nsumrakTT: It is very bad strategy to use env variables for pytest tests.
When needed for some kind of analyses, superset identification, or bisection, tests would have the same name (it is not 1-1 identification for the test name and test).
Much better way is to use pytest parametrization.

brnorris03: Yes, not a fan of env variables, but wanted to enable testing least intrusively. Now that it works, I will try to figure out how to parameterize just the tests that involve the affected pipeline.

lib/Dialect/D2M/Analysis/DstAnalysisGraphColoring.cpp
```diff
@@ -0,0 +1,274 @@
...
+        if (!region.getOps<AcquireDstOp>().empty()) {
+          continue;
+        }
+
+        auto dstAccesses = identifyDstAccesses(genericOp, region);
+        if (dstAccesses.empty()) {
+          continue;
+        }
+
+        // Store identified accesses for reuse by transformation passes
```
vwellsTT: nit: .

lib/Dialect/D2M/Analysis/DstAnalysisGraphColoring.cpp
```diff
@@ -0,0 +1,274 @@
...
+/// Uses getOperandsLoadFromDstRegister() to identify loads and
+/// getDstRegInPlace() to determine if stores need separate DST slices.
+static llvm::SmallVector<std::pair<mlir::Operation *, int64_t>>
+identifyDstAccesses(mlir::Operation *op, mlir::Region &region) {
+  llvm::SmallVector<std::pair<mlir::Operation *, int64_t>> dstAccesses;
+  llvm::DenseSet<mlir::Operation *> seenOps;
+  int64_t index = 0;
+
+  region.walk([&](mlir::Operation *operation) {
+    // Check if this is a D2M compute operation with DST interface
```
vwellsTT: nit: .

include/ttmlir/Dialect/D2M/Transforms/GraphColoringStrategy.h
```diff
@@ -0,0 +1,104 @@
...
+///
+/// Slower than the greedy algorithm but produces high-quality colorings with
+/// guaranteed optimality for K-colorable graphs; best for complex interference
+/// graphs.
+///
+/// Reference: G. J. Chaitin, "Register allocation & spilling via graph
+/// coloring," SIGPLAN Not., vol. 17, no. 6, pp. 98-105, Jun. 1982.
+/// P. Briggs, "Register Allocation via Graph Coloring," PhD thesis,
+/// Rice University, 1992.
+class ChaitinBriggsColoring : public ColoringStrategy {
```
vwellsTT, line 79: general remark; as I recall, in a severely constrained case such as an interval graph, a true optimal solution to graph coloring can be computed in polynomial time (@vroubtsovTT and I discussed this a couple months ago).  The constraints I recall are like: 1 register type, no loops, etc. though I'm certainly not up on these things.
I'm curious, do we not meet such constraints? If so, I'm not clear on why we want ChaitinBriggs and Greedy at all vs linear solution

lib/Dialect/D2M/Analysis/DstAnalysisGraphColoring.cpp
```diff
@@ -0,0 +1,274 @@
...
+        strategyName(name) {}
+
+  DstAnalysisResult analyze(mlir::Operation *op) override {
+    DstAnalysisResult result;
+    unsigned maxSlicesNeeded = 0;
+
+    mlir::WalkResult walkResult = op->walk([&](GenericOp genericOp)
+                                               -> mlir::WalkResult {
+      for (auto &region : genericOp.getRegions()) {
+        // Skip if already has DST allocation
```
vwellsTT, line 109: nit: .

.github/test_scripts/d2m_tests.sh
brnorris03: Discussed various options with @ctodTT and implemented this by automatically parameterizing the relevant pytests when the `--dst-allocation-strategy` is provided. This way explicit additional marking of tests is not needed and also ensures that new tests are included without relying on test writers to figure out whether they should be marked for this or not.
Test names include the parameter, e.g., before:
```test/python/golden/test_composite_functions.py::test_geglu_split[ttmetal-f32-shape0]```
now
```test/python/golden/test_composite_functions.py::test_geglu_split[ttmetal-f32-shape0-dst:greedy]```

File: lib/Dialect/D2M/Analysis/DstAnalysisGraphColoring.cpp [hunk 15]
```diff
@@ -0,0 +1,274 @@
...
+          size_t leader = nodeToLeader[i];
+          size_t contractedIdx = leaderToContracted[leader];
+          coloring[i] = contractedColoring[contractedIdx];
+        }
+        normalizeColorOrder(coloring);
+
+        // Find actual number of colors used.
+        unsigned colorsUsed = 0;
+        for (unsigned color : coloring) {
+          colorsUsed = std::max(colorsUsed, color + 1);
```
vwellsTT: would tracking max colors used during traveral not be simpler?

lib/Dialect/D2M/Analysis/GraphColoringStrategy.cpp
```diff
@@ -0,0 +1,493 @@
...
+    const std::vector<std::vector<size_t>> &adjacencyList, unsigned numColors,
+    std::vector<unsigned> &coloring) {
+  coloring.assign(adjacencyList.size(), UINT_MAX);
+  size_t numNodes = adjacencyList.size();
+
+  // Simplification phase: remove nodes with degree < K.
+  std::vector<size_t> simplifyStack;
+  std::vector<bool> removed(numNodes, false);
+
+  while (true) {
```
vwellsTT: personally, I think `while(true)` is a bad pattern for production code, and it'd be better to move break conditions in here

include/ttmlir/Dialect/D2M/Transforms/GraphColoringStrategy.h
brnorris03: The graph coloring approaches reuse registers without having to have postprocessing. The actual algorithm(s) used in this PR are not the main point -- the main new capability is the ability to do different algorithms. I could implement anything else instead of Chaitin-Briggs with very minimal localized additions (and not touch the actual transformation pass almost at all). They also do use fewer registers than the current linear + fixups strategy for some sequences of fused ops.

I think replacing Chaitin with a better algorithm as a next step would be great and a much easier exercise than it would have been in the pre-refactored implementation.

brnorris03: Do you have a reference for the preferred graph coloring strategy you discussed? I don't mind replacing the Chaitin-Briggs one with something else that can also handle loops (not necessary yet in this PR but will be in the near future to remove the special loop handling). 

Is it the 2006 "Optimal register allocation for SSA-form programs in polynomial time"?

lib/Dialect/D2M/Analysis/DstAnalysisGraphColoring.cpp
brnorris03: Definitely -- changed it to eliminate this loop.

lib/Dialect/D2M/Analysis/GraphColoringStrategy.cpp
brnorris03: agreed, removed.

zoecarver:
Can I suggest we land this if the refactor looks good and continue the specific algorithm discussion after? A big refactor like this will probably run into conflicts and other issues sooner rather than later so I think it makes sense to land after the architecture is reviewed. The specific algorithm can be improved after the fact. This will also unblock tt-lang work. 

CC @vwellsTT @brnorris03

test/python/golden/conftest.py
```diff
@@ -15,11 +15,12 @@
 
 ALL_BACKENDS = set(["ttnn", "ttmetal", "emitc", "emitpy"])
 ALL_SYSTEMS = set(["n150", "n300", "llmbox", "tg", "p150", "p300"])
-
+DST_ALLOCATION_STRATEGIES = ["basic", "greedy", "chaitin"]
```
vwellsTT, line 18: personally, I think "chaitin-briggs" is clearer, there is a prior "Chaitin" algorithm which "Chaitin-Briggs" refined, if I recall my grad school classes correctly

include/ttmlir/Dialect/D2M/Transforms/GraphColoringStrategy.h
vwellsTT: > Do you have a reference for the preferred graph coloring strategy you discussed? 

No, I think we just referenced Wikipedia on interval graphs: https://en.wikipedia.org/wiki/Interval_graph which mentions there's a linear greedy alg to color interval graphs (and cites a textbook).  I wonder if your greedy strategy would be optimal for our actual graphs...but I am a bit rusty on register allocation stuff, would have to spend more time looking it up.

Regardless, my point was more whether Chaitin-Briggs may be overkill for our use cases or not.  I think it adds a lot of complex code, and I'm not sure we get much mileage from it vs simpler strategies.  But I agree the engineering is good s.t. it's modular and silo'ed from rest of the code, so I don't think it really hurts anything to have a configurable option for Chaitin-Briggs (and perhaps in the future we will have more complex IR s.t. more complex handling is needed).

vroubtsovTT: My understanding is that if the problem

1. has straight-line code (does not involve control flow) and
2. does not involve instruction scheduling (op order is given) and 
3. all registers are the "same",

then the register allocation problem becomes one of ["interval graph coloring"](https://homepages.gac.edu/~sskulrat/Courses/2015F-375/lectures/g2.pdf) which is optimally solved by a greedy/linear scan type of algorithm.  The apparent contradiction with the problem known to be NP-hard is resolved by noting that these ["interval graphs"](https://en.wikipedia.org/wiki/Interval_graph) form an "easier" sub-class of general conflict graphs, one for which certain hard problems are in fact solvable in polynomial time. (Similar to how, say, CSP problems are NP-hard but become "easy" on graphs that are in fact trees, etc.)

My understanding is also that our thread-level lowerings have both control flow and are free to look for optimal op orders (reduction trees, etc), so neither (1) nor (2) above apply. So, it seems like we're justified in using a classic graph coloring algorithm here.

(As a side note, the mem allocator pass has both (1) and (2) but not (3), because memory buffer requests have different sizes.)

vwellsTT: > My understanding is also that our thread-level lowerings have both control flow and are free to look for optimal op orders (reduction trees, etc), so neither (1) nor (2) above apply

Hmm, do we have control flow w.r.t. to these allocations though? (i.e. do they happen within loop bodies? My impression is no, but I haven't worked with 'em closely).

vroubtsovTT: Ditto, I am not closely familiar with this part of the pipeline, but I think when @mbagherbeikTT  & Co translate large expressions by walking them in various topologically sorted orders (for Sethi-Ullman, whatever) this is when we have some control over instruction ordering.

vwellsTT: Ah, was not aware.  Good point then.

PR #5881, opened 2025-11-17
[D2M] Extract the linalg-to-affine transformations from the D2MInsertDstRegisterAccess pass 

### LINKED ISSUE #5912: Extract linalg-to-affine conversion from D2MInsertDstRegisterAccess pass

brnorris03:
This PR extracts the linalg-to-affine conversion logic from `D2MInsertDstRegisterAccessPass` into a new standalone `D2MLinalgToAffine` pass as part of a larger DST analysis refactoring that would enable different DST allocation strategies and more flexible use of the DST analysis by other passes.

### Ticket
Closes #5912

### Problem description

`D2MInsertDstRegisterAccess` previously performed two distinct operations:
1. Converting `linalg.generic` operations to affine loops.
2. Managing DST register allocation and access.

### What's changed

Extracted linalg-to-affine conversion logic into a new standalone pass
`D2MLinalgToAffine`. This is a refactoring without functional changes - users of
`TTMetalPipelines` are not affected.

New pass `D2MLinalgToAffine`:
- Converts `linalg.generic` ops to affine loops within `d2m.generic` regions
- When `useTileMatmul=false`: skips ops containing `tile_matmul` (leaves them
  for `InsertDstRegisterAccess`)
- When `markRootLoops=true`: marks converted affine loops with `d2m.linalg_root`
  attribute for downstream coordination
- Includes proper error handling: checks `FailureOr` result from
  `linalg::linalgOpToAffineLoops()` before dereferencing

Modified pass `D2MInsertDstRegisterAccess`:
- Removed internal linalg-to-affine conversion logic for linalg ops that don't
  contain `tile_matmul`.
- Added precondition check in `runOnOperation()`: fails with diagnostic if
  unconverted `linalg.generic` ops remain (except `tile_matmul` when
  `useTileMatmul=false`). Without this, downstream passes could crash
  cryptically due to unexpected IR.
- Now only processes affine loops marked with `d2m.linalg_root` attribute
- Removes marker attribute immediately after identification.
- Still handles `tile_matmul_block` creation when `useTileMatmul=false`. 
 This will be extracted in a subsequent PR.

Pipeline changes in `TTMetalPipelines.cpp`:
- Added `d2m::createD2MLinalgToAffine()` immediately before
  `d2m::createD2MInsertDstRegisterAccess()`.
- Configured with `useTileMatmul` matching downstream pass and
  `markRootLoops=true`.

Pass definition in `Passes.td`:
- Added `D2MLinalgToAffine` definition with options `useTileMatmul` and
  `markRootLoops`.

#### Tests
- New test file `test/ttmlir/Dialect/D2M/Transforms/linalg_to_affine.mlir` -
  standalone tests for the new pass.
- New test file
`test/ttmlir/Dialect/D2M/Transforms/insert_dst_register_access_missing_linalg_to_affine.mlir`
  - verifies new precondition check.
- Updated 10 existing `insert_dst_register_access_*.mlir` test files to run
  `--d2m-linalg-to-affine` before `--d2m-insert-dst-register-access`.

### Checklist

### DIFF
--- a/lib/Dialect/D2M/IR/D2MOps.cpp
+++ b/lib/Dialect/D2M/IR/D2MOps.cpp
@@ -1229,7 +1229,7 @@ static mlir::LogicalResult verifyAffineBlocking(
               })
               .Case<RankedTensorType>(
                   [&](RankedTensorType tensorType) -> LogicalResult {
-                    if (!mlir::dyn_cast_or_null<ttcore::MetalLayoutAttr>(
+                    if (!mlir::dyn_cast<ttcore::MetalLayoutAttr>(
                             tensorType.getEncoding())) {
                       return emitOpError("tensor operand must have a metal "
                                          "layout encoding, "


lib/Dialect/D2M/Transforms/Allocate.cpp
```diff
@@ -1192,20 +1189,6 @@ class D2MAllocate final : public impl::D2MAllocateBase<D2MAllocate> {
...
-    if (!llvm::isa<func::ReturnOp>(lastOp)) {
-      OpBuilder::InsertionGuard guard(rewriter);
-      {
-        rewriter.setInsertionPointAfter(lastOp);
-        rewriter.create<memref::DeallocOp>(lastOp->getLoc(),
-                                           allocOp.getResult());
-      }
-    }
-  }
-
```
vroubtsovTT, line 1259: Curious...why this change?

brnorris03: Trying to use `OwnershipBasedBufferDeallocationPass` at a later time to avoid issues with deallocs being reordered by other passes. Still WIP...

As to the reason -- I have a more sizeable writeup, but the ultra-short version is, the allocate pass deallocates buffers while their corresponding streams are still live. Moreover, its liveness analysis does not include newly created streams, hence when deallocs are inserted, they are not taken into account.

Update: that may have been a red herring. It is true that Allocate returns different orderings of deallocs and addresses between different runs, but that was not causing the issues I was hunting down at the time. Allocate is not modified in this PR.

lib/Dialect/D2M/IR/D2MOps.cpp
```diff
@@ -1206,6 +1213,41 @@ ::mlir::LogicalResult d2m::GenericOp::verify() {
...
+                    }
+                    return success();
+                  })
+              .Default([](Type) { return success(); });
+
+      if (failed(result)) {
+        return failure();
+      }
+    }
+
```
brnorris03, line 1247: Not related to the main goal of this PR, but did encounter verifier crashes while debugging things, which prompted this addition.

The new test that goes with it is `test/ttmlir/Dialect/D2M/generic/generic_missing_device_layout.mlir`

include/ttmlir/Dialect/TTMetal/Pipelines/TTMetalPipelines.h
```diff
@@ -149,6 +149,14 @@ struct TTIRToTTMetalPipelineOptions
...
       llvm::cl::init("max")};
 
+  // Option to control whether d2m-allocate inserts deallocs or delegates to
+  // d2m-insert-deallocs pass.
+  Option<bool> insertDeallocs{
+      *this, "insert-deallocs",
+      llvm::cl::desc("Insert deallocs in d2m-allocate pass (true) or use "
+                     "separate d2m-insert-deallocs pass (false)."),
+      llvm::cl::init(false)};
+
```
nsmithtt: I think this is no longer used.

lib/Dialect/D2M/IR/D2MOps.cpp
```diff
@@ -597,6 +599,7 @@ mlir::LogicalResult d2m::StreamLayoutOp::bufferize(
 mlir::bufferization::AliasingValueList d2m::StreamLayoutOp::getAliasingValues(
     mlir::OpOperand &, const mlir::bufferization::AnalysisState &) {
   bufferization::AliasingValueList result;
+  result.addAlias({getResult(), bufferization::BufferRelation::Equivalent});
```
nsmithtt, line 602: Seems somewhat irrelevant to the core change.  That said, I suppose it's correct so won't block the PR on it.

lib/Dialect/D2M/IR/D2MOps.cpp
nsmithtt: Awesome!  Definitely an improvement for the verifier.

lib/Dialect/D2M/IR/D2MOps.cpp
```diff
@@ -1515,11 +1554,13 @@ d2m::GenericOp::getOperandGridShapes() {
       mlir::tt::ttcore::DeviceLayoutInterface layout =
           mlir::dyn_cast<mlir::tt::ttcore::DeviceLayoutInterface>(
               memrefType.getLayout());
+      assert(layout && "memref operand must have a device layout attribute");
```
nsmithtt: Perhaps the line above should use `mlir::cast` which will cast on our behalf.

lib/Dialect/D2M/IR/D2MOps.cpp
```diff
@@ -1515,11 +1554,13 @@ d2m::GenericOp::getOperandGridShapes() {
...
       gridShapes.emplace_back(layout.getGridShape(memrefType));
     } else {
       auto tensorType = mlir::cast<RankedTensorType>(operand.getType());
       ttcore::MetalLayoutAttr layout =
           mlir::cast<ttcore::MetalLayoutAttr>(tensorType.getEncoding());
+      assert(layout && "tensor operand must have a metal layout encoding");
```
nsmithtt: I think this assert is redundant because `mlir::cast` will assert if the cast isn't legal.

lib/Dialect/D2M/Transforms/InsertDstRegisterAccess.cpp
```diff
@@ -99,26 +121,32 @@ struct D2MInsertDstRegisterAccessRewriter final
...
-        rewriter.setInsertionPoint(linalgGenericOp);
-        // Apply linalg to affine loops pass.
-        auto linalgLoops =
-            linalg::linalgOpToAffineLoops(rewriter, linalgGenericOp);
-        if (failed(linalgLoops)) {
-          linalgToAffineFailed = true;
+        // This should not happen - all other linalg ops should have been
+        // converted by LinalgToAffine pass.
+        linalgToAffineFailed = true;
+      });
```
nsmithtt: I know this code is keeping the `linalgToAffineFailed` bool in the way it used to work, but it'd be more natural to return `WalkResult::interrupt()` and then check `walkResult.wasInterrupted()`.

lib/Dialect/D2M/Transforms/InsertDstRegisterAccess.cpp
```diff
@@ -789,7 +829,29 @@ class D2MInsertDstRegisterAccess
       D2MInsertDstRegisterAccess>::D2MInsertDstRegisterAccessBase;
 
   void runOnOperation() final {
-    MLIRContext *ctx = &getContext();
+    ModuleOp module = getOperation();
```
nsmithtt: We should call this var `moduleOp` to avoid keywords from c++20.

third_party/CMakeLists.txt
nsmithtt, line 1: I appreciate cmake cleanup, but perhaps we can do this as part of a separate PR?  It seems unrelated to the core change.

lib/Dialect/D2M/Transforms/InsertDstRegisterAccess.cpp
brnorris03: Updated (although this particular code is going to change again in the DST PR where this conversion is extracted into a separate utility).

include/ttmlir/Dialect/TTMetal/Pipelines/TTMetalPipelines.h
brnorris03: Yep, removed; thanks for catching it!

third_party/CMakeLists.txt
brnorris03: Oops, I think I must have accidentally opened it and VSCode autoformatted it. Reverted.

lib/Dialect/D2M/IR/D2MOps.cpp
brnorris03: Yes, I know at the moment it doesn't matter, but I think it's harmless and better to be consistent.

nsmithtt (approved):
Looks great, thank you!

PR #5885, opened 2025-11-17
Docker build in the CI refactor

vvukomanTT:
### Ticket
/

### Problem description
Execute the docker build job in the CI on CIv1 builder machines labelled with `docker-builder`.

### What's changed
The docker build job in the CI is now executed on the CIv1 builder machines.
Two changes were required to enable this:
- Set the permissions of the workspace to the current user to avoid the permisisons issue
- Limit the cmake's parallel level during the docker image building to avoid OOM error which is caused when the toolchain is built without the parallel level limiting

### Checklist
- [ ] Docker build job passing with the changes from this PR -> [job](https://github.com/tenstorrent/tt-mlir/actions/runs/19425546544/job/55572082145)

### DIFF
--- a/.github/build-docker-images.sh
+++ b/.github/build-docker-images.sh
@@ -46,7 +46,7 @@ build_and_push() {
         sudo apt-get update && sudo apt-get install -y cmake build-essential
 
         # Toolchain build causes the OOM error if parallel level is not limited
-        export CMAKE_BUILD_PARALLEL_LEVEL=16
+        export CMAKE_BUILD_PARALLEL_LEVEL=32
 
         echo "Building image $image_name:$DOCKER_TAG"
         docker build \


.github/workflows/call-build-docker.yml
```diff
@@ -48,11 +48,14 @@ jobs:
...
     needs: check-if-docker-exist
     if: needs.check-if-docker-exist.outputs.docker-image == ''
-    runs-on: tt-ubuntu-2204-large-stable
+    runs-on: docker-builder
     outputs:
       docker-image: ${{ steps.build.outputs.docker-image }}
       docker-tag: ${{ steps.build.outputs.docker-tag }}
     steps:
+      - name: Fix workspace permissions issue
+        run: sudo chown -R $USER:$USER $GITHUB_WORKSPACE
```
nsumrakTT, line 57: I would consider using docker container with base image for all workflows where docker is not needed.
We could prevent all:
- chowns and chmods
- pip --force-reinstall problems (install in system pip repo, docker-venv)
- set git safe dir, git clean
and all other stuff that we have in our workflows that are consequence of using docker image-no docker image setups on same machine.

nsumrakTT (approved):
The comment I had is just to think about it, not specific to this PR

.github/workflows/call-build-docker.yml
vvukomanTT: I agree that using the base image would make the mentioned issues disappear but my main concern is which base image are we using? 
Because I am not a fan of using the "latest" image in our own workflows.
1. that is a tag used for developers to more easily pull the newest docker image and "latest" could be semantically incorrect when the workflow is ran. 
2. "latest" could be itself broken or the tagging logic for it could be broken

If either of those happen, all of a sudden all our workflows are using that semantically incorrect/broken image which has nothing to do with them because they might have working images on that commit. 

Those are some of the reasons we are using the calculated current tag in the image names in all our workflows right now.

nsumrakTT: Ok, make sense.
Although I disagree with "latest", lets merge this and we'll figure it out in subsequent PR(s).

PR #5892, opened 2025-11-17
Added support for dynamically loading resnet builder ops.

tapspatel:
This is a massive refactor effort. The changes can be summarized as follows:

For all ttir ops found in Resnet, the underlying builder op has changed

- remove op_proxy to allow for more explicit build
- add @tag decorator to dynamically create ttir.op <-> builder op mapping
- add @parse decorator to dynamically create the ttir.op from another parsed op from a module
- add @split decorator to dynamically split an op from a module to create independent single op modules

These changes also fully support the golden infrastructure.

Future changes will be made to reduce code bloat once working functionality of ops are in.

### DIFF
--- a/runtime/tools/chisel/test/test_main_cli.py
+++ b/runtime/tools/chisel/test/test_main_cli.py
@@ -8,6 +8,7 @@ from pathlib import Path
 import pytest
 
 
+@pytest.mark.skip
 @pytest.mark.parametrize(
     "ttir_path, function_name",
     [
--- a/test/python/golden/experimental/test_mpmd_ops.py
+++ b/test/python/golden/experimental/test_mpmd_ops.py
@@ -25,7 +25,7 @@ def multiple_meshes(
     builder: StableHLOBuilder,
     unit_attrs: Optional[List[str]] = None,
 ):
-    return builder.add(in0, in1, unit_attrs=unit_attrs)
+    return builder.add(in0, in1)
 
 
 @pytest.mark.parametrize(
--- a/test/python/golden/test_ttir_parse_split_ops.py
+++ b/test/python/golden/test_ttir_parse_split_ops.py
@@ -1,0 +1,50 @@
+# SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+
+import pytest
+import torch
+from typing import Callable, List, Optional, Tuple, Union
+from collections import OrderedDict
+from functools import reduce
+import operator
+
+from builder.base.builder_utils import build_module, load_mlir_file, split_mlir_file
+
+pytestmark = pytest.mark.frontend("ttir")
+
+resnet_ops_ir = """module {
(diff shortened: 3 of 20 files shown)


test/python/golden/test_ttir_parse_split_ops.py
```diff
@@ -0,0 +1,50 @@
...
+    %22 = ttir.empty() : tensor<12x31x56x114xf32>
+    %23 = \"ttir.pooling\"(%arg10, %22) <{base_dilations = array<i64: 1, 1, 1, 1>, operandSegmentSizes = array<i32: 1, 1>, padding = array<i64: 0, 0, 0, 0, 0, 0, 0, 0>, pooling_method = #ttir<pooling_method Max>, window_dilations = array<i64: 1, 1, 1, 1>, window_dimensions = array<i64: 1, 1, 3, 3>, window_strides = array<i64: 1, 1, 2, 2>}> : (tensor<12x64x114x114xf32>, tensor<12x31x56x114xf32>) -> tensor<12x31x56x114xf32>
+    return %17 : tensor<1x1024xf32>
+  }
+}"""
+
+
+@pytest.mark.parametrize("mlir_text", [resnet_ops_ir])
+def test_resnet_ops(mlir_text: str, request, device):
+    mlir_module, builder = load_mlir_file(mlir_text)
```
ddilbazTT, line 49: I think we should move the resnet mlir to a file instead of a text block.

ddilbazTT (approved):

runtime/tools/chisel/test/test_main_cli.py
```diff
@@ -8,6 +8,7 @@
 import pytest
 
 
+@pytest.mark.skip
```
ctodTT, line 11: Nit, but an issue here for tracking would be nice if applicable

test/python/golden/experimental/test_mpmd_ops.py
```diff
@@ -25,7 +25,7 @@ def multiple_meshes(
     builder: StableHLOBuilder,
     unit_attrs: Optional[List[str]] = None,
```
ctodTT, line 26: On all changes like this, unless we want to add back functionality later (in which case I think we need a comment about it), I think the `unit_attrs` parameter should be removed as well since it's now unused and would clean up the api greatly. This probably falls under the future changes you mention in the PR message, but wanted to bring attention to it anyway

test/python/golden/test_ttir_parse_split_ops.py
ctodTT: +1 on this, makes tracking easier

tools/builder/ttir/ttir_builder.py
```diff
@@ -33,7 +48,83 @@ def __init__(
...
+            return func
+
+        return decorator
+
+    def split(name):
+        def decorator(func):
+            func._split = name
+            return func
+
+        return decorator
```
ctodTT, line 72: I'd have some cursory docstrings on these since these are going to be decently widely used

test/python/golden/experimental/test_mpmd_ops.py
tapspatel: we need to keep it for cpu hoisting, its still a required parameter

test/python/golden/test_ttir_parse_split_ops.py
tapspatel: yup, we have around several 100 modules from various models we need to run regression for, we will set this up

tools/builder/ttir/ttir_builder.py
tapspatel: ill add some comments once this is finalized, there's still lots of refactoring to come ontop of this

Issue #5855, opened 2025-11-14
Optimize MeanOp usage in ScaledSumToMeanPattern


mvasiljevicTT:
Currently efficientnet model has pattern:
```
reshape [1, 1, NHW, C] -> [N, H, W, C]
|
mean <dim = [1,2]>
|
reshape [N, 1, 1, C] -> [1, 1, N, C]
```
which produces this block of tt-metal ops (for shape [12, 7, 7, 1152]):
```
Op 1: ReshapeDeviceOperation         -    137,715 ns (    0.14 ms)
Op 2: FillPad                        -    177,137 ns (    0.18 ms)
Op 3: Reduce                         -     37,868 ns (    0.04 ms)
Op 4: Transpose                      -     38,004 ns (    0.04 ms)
Op 5: Reduce                         -      9,406 ns (    0.01 ms)
Op 6: Transpose                      -     19,492 ns (    0.02 ms)
Op 7: ReshapeDeviceOperation         -     10,290 ns (    0.01 ms)

Total DEVICE FW DURATION [ns]: 429,912 ns (0.43 ms)
```
If it was to be changed with this pattern instead:
```
reshape [1, 1, NHW, C] -> [N, 1, HW, C]
|
mean <dim = 2>
|
reshape [N, 1, 1, C] -> [1, 1, N, C]
```
it would have these metal ops (for same shape):
```
Op 1: ReshapeDeviceOperation         -    103,999 ns (    0.10 ms)
Op 2: FillPad                        -     54,699 ns (    0.05 ms)
Op 3: Reduce                         -     11,391 ns (    0.01 ms)
Op 4: ReshapeDeviceOperation         -     10,442 ns (    0.01 ms)

Total DEVICE FW DURATION [ns]: 180,531 ns (0.18 ms)
```
having 2x better device perf for that block, which is expected to improve total device perf for efficientnet by 8%.

PR #3915, opened 2025-06-25
Create function to register TM fusion patterns to a RewritePatternSet

LPanosTT:
### Ticket
https://github.com/tenstorrent/tt-mlir/issues/3877

### Problem description
- Sequences of TMs may be equivalent to shorter sequences of TMs, we should fuse these cases.
- These sequences may appear at any time during a graph transformation, namely EIO
- We should be able to register these patterns in any pattern set so these patterns can be fused during a pass in which such patterns may appear

### What's changed
- Created a dialect-wide function which can be used to register TM fusion patterns to any RewritePatternSet
- Added one pattern rewrite which appears in the ttir graph of resnet50 (from forge-fe) **during** EIO
- Register this pattern on tandem with commute patterns in EIO so that they may be fused during commuting of the ops as to open new pathways for TMs to commute during EIO
- Added a test to ensure this fusion is accurate and is executed

### Checklist
- [X] New/Existing tests provide coverage for changes

PR #5899, opened 2025-11-18
[TTIRFusing] Improve ScaledSumToMeanPattern

mvasiljevicTT:
### Ticket
[Link to Github Issue](https://github.com/tenstorrent/tt-mlir/issues/5855)

### Problem description
In efficientnet model, global avg pool is lowered as mean, which surrounded by reshapes is:
```
reshape [1, 1, NHW, C] -> [N, H, W, C]
|
mean <dim = [1,2]>
|
reshape [N, 1, 1, C] -> [1, 1, N, C]
```
This can be improved so that reduce is only done on one dim, like:
```
reshape [1, 1, NHW, C] -> [N, 1, HW, C]
|
mean <dim = 2>
|
reshape [N, 1, 1, C] -> [1, 1, N, C]
```
### What's changed
Added reshape and changed meanOp in `ScaledSumToMeanPattern`. This reshape is later folded with previous reshape so it doesn't affect perf directly.

With this change, inverse permutes around reduce were not erased, as there was a reshape in between they couldn't commute with. 
It is hard and risky to define general commute patterns for reshape and permute, so existing `PermuteReshapePermuteFusionPattern` in `TTIRTMFusionPatterns`, introduced with [this PR](https://github.com/tenstorrent/tt-mlir/pull/3915), is modified to cover this case too.

### Checklist

### DIFF
--- a/lib/Dialect/TTIR/IR/TTIRTMFusionPatterns.cpp
+++ b/lib/Dialect/TTIR/IR/TTIRTMFusionPatterns.cpp
@@ -30,19 +30,22 @@ namespace mlir::tt::ttir {
 // reshape (N, C, H, W) -> (N, C, 1, H*W)
 // permute(0, 2, 3, 1): [N, 1, H*W, C]
 //
-// The result of this sequence is identical to the following reshape:
+// Both cases result in the same final shape. The result of either sequence is
+// identical to the following reshape:
 // reshape (N, H, W, C) -> (N, 1, H*W, C)
 
-namespace {
-
 // Reshape: n c h w -> n 1 c h*w
-SmallVector<int64_t> computeCase1Reshape(ArrayRef<int64_t> inputShape) {
+static SmallVector<int64_t> computeCase1Reshape(ArrayRef<int64_t> inputShape) {
+  assert(inputShape.size() == 4 &&
+         "Expected 4D input tensor as output of 4D permutation");
   return SmallVector<int64_t>{inputShape[0], 1, inputShape[1],
                               inputShape[2] * inputShape[3]};
 }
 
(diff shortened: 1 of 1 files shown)


lib/Dialect/TTIR/IR/TTIRTMFusionPatterns.cpp
```diff
@@ -8,18 +8,59 @@
...
-// reshape op when the following conditions are met:
-// Original shape: [A, B, C, D]
-// permute(0, 3, 1, 2): [A, D, B, C]
-// reshape (A, D, B, C) -> (A, 1, D, B*C)
-// permute(0, 1, 3, 2): [A, 1, B*C, D]
+// reshape op.
 //
-// The result of this sequence is identical to the following reshape:
-// reshape (A, B, C, D) -> (A, 1, B*C, D)
+// It is hard to express all pairs of permutations and reshapes that can
```
azecevic000, line 18: I tinkered with some ideas for the generality of `permute(reshape(permute(x))` pattern. I have an algorithm that I'm pretty confident covers all cases when the folding is possible, regardless of the `x`s rank. I'll have to test it further to make sure it indeed works as intended, but it works for these two patterns at least.

I'm okay with your solution going forward for now.

mvasiljevicTT: Thanks for taking the time to think about this, I would love to hear your ideas for improvement!

mvasiljevicTT: synced offline, will be done as separate pr

PR #5903, opened 2025-11-18
[MLA] Enable sharding for RMSNorm op 

bmalesevicTT:
### Ticket
This PR investigates the behavior and correctness of the RMSNorm op when executed in sharded mode (see issue [#5706](https://github.com/tenstorrent/tt-mlir/issues/5706)
) and adds RMSNorm to the set of ops supported for sharding. The goal was to validate functional correctness under sharding, identify limitations, and document the constraints observed on real model-sized examples.

The example used in this PR is taken from a real model that runs on Metal. Although the graph representation appears row-major, the actual execution for this op should be tiled, as it is here.

For computation purposes, the tiling of the weights must match the tiling of the input tensor. This causes the RMSNorm weights tensor to include padding—effectively reserving 32× the size of the weights in L1. With larger batch sizes this space becomes useful parallelization capacity, but in the LLama-derived example tested here, the overall storage requirement exceeds L1 once the batch size goes above 6. If bias is also included, the maximum feasible batch size for L1-sharded execution drops to 5.

### Testing Approach

Ideally, this would be tested end-to-end with the TTIR builder, since it validates the entire pipeline from TTIR → execution → golden output. However, the TTIR builder currently does not work when the optimizer feature is enabled.

To work around this:

- Lit tests were added to validate the correctness of sharded TTIR → TTNN-IR lowering.
- The TTNN builder infrastructure was then used to execute the same example and verify golden results at runtime.

This two-step strategy covers both sides:
 - correctness of the compiler pipeline under sharding, and
 - correctness of execution and numerical results.

This PR provides evidence that RMSNorm works under sharded mode within the documented batch-size and L1-memory constraints, and it adds the test infrastructure required to validate this path.

### Notes for improvements:
-Testing via the TTNN builder uses a hacky approach, adding a temporary, hardcoded width-sharded L1 tiled layout via unitattr, since the builder currently only supports a DRAM-interleaved layout. This workaround enables proper sharded testing, but layout configuration in TTNN builder tests should be made flexible and non-hardcoded in the future, issue [#5907](https://github.com/tenstorrent/tt-mlir/issues/5907).
-Testing via TTIR builder should be possible with the added test, once issue [#TTIR builder: Segfault when optimizer is enabled #5909](https://github.com/tenstorrent/tt-mlir/issues/5909) is resolved.
-Technically RMSNorm doesn't have `bias` in the Torch implementation, but TTNN implementation has it?

### DIFF
--- a/test/ttmlir/Silicon/TTNN/n150/optimizer/rmsnorm_w_bias_sharding.mlir
+++ b/test/ttmlir/Silicon/TTNN/n150/optimizer/rmsnorm_w_bias_sharding.mlir
@@ -1,16 +1,15 @@
 // REQUIRES: opmodel
-// RUN: ttmlir-opt --ttir-to-ttnn-backend-pipeline="system-desc-path=%system_desc_path% enable-optimizer=true memory-layout-analysis-enabled=true" %s -o %t.mlir
+// RUN: ttmlir-opt --ttir-to-ttnn-backend-pipeline="system-desc-path=%system_desc_path% enable-optimizer=true memory-layout-analysis-enabled=true" %s -o %t.mlir --mlir-print-local-scope
 // RUN: FileCheck %s --input-file=%t.mlir
 // RUN: ttmlir-translate --ttnn-to-flatbuffer -o %t.ttnn %t.mlir
 
 // Test RMSNorm sharding for LLM models, without weights, with bias, Llama 3.2 example variant
 
 func.func @rmsnorm_sharding(%arg0: tensor<1x1x32x2048xbf16>, %arg1: tensor<2048xbf16>) -> tensor<1x1x32x2048xbf16> {
-    // CHECK-DAG: #[[LAYOUT_0:.*]] = #ttnn.ttnn_layout{{.*}}width_sharded{{.*}}
 
     %0 = ttir.empty() : tensor<1x1x32x2048xbf16>
     // First RMSNorm operation with width sharding: input [1,1,32,2048xbf16], no weights, bias [2048xbf16]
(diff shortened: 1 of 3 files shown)


test/ttmlir/Silicon/TTNN/n150/optimizer/rmsnorm_w_bias_sharding.mlir
```diff
@@ -0,0 +1,21 @@
+// REQUIRES: opmodel
+// RUN: ttmlir-opt --ttir-to-ttnn-backend-pipeline="system-desc-path=%system_desc_path% enable-optimizer=true memory-layout-analysis-enabled=true" %s -o %t.mlir
+// RUN: FileCheck %s --input-file=%t.mlir
+// RUN: ttmlir-translate --ttnn-to-flatbuffer -o %t.ttnn %t.mlir
+
+// Test RMSNorm sharding for LLM models, without weights, with bias, Llama 3.2 example variant
+
+func.func @rmsnorm_sharding(%arg0: tensor<1x1x32x2048xbf16>, %arg1: tensor<2048xbf16>) -> tensor<1x1x32x2048xbf16> {
+    // CHECK-DAG: #[[LAYOUT_0:.*]] = #ttnn.ttnn_layout{{.*}}width_sharded{{.*}}
```
rpavlovicTT: Use `--mlir-print-local-scope` above and check layouts inline instead of capturing them

rpavlovicTT:
Hey @tapspatel, can someone from your team take a look here? Majority of changes are in builder infra.

PR #5915, opened 2025-11-18
[D2M] Support for add, mul, sub, div, pow unary with scalars

### LINKED ISSUE #5917: [D2M] Add support for binary ops with scalars
Title

dlokeTT:
### Ticket
Closes #5917 

### Problem description
Compiler doesn't have support for binary ops such as add, subtract, multiply, divide and pow with scalars. We only have support for two tile variants which means we turn scalars into tiles. This is inefficient and creates multiple bottlenecks in the pipeline which include allocating memory for the tensor, tilizing the tensors, using slices in dest for these tiles instead of an SFPU LReg and the llk calls themselves being faster. 

### What's changed
New class called `D2M_GenericRegionComputeBinaryWithScalarOptionDstOp` for binary ops that can have rhs be tileOrScalar when llk api supports binop with scalar
New TTKernel ops for the unary tile versions of these ops.
Updated `InsertDstRegisterAccess.cpp` to handle binary ops that could be scalars by only loading the lhs if rhs is scalar
Added lit tests

### Checklist

### DIFF
--- a/lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
+++ b/lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
@@ -726,52 +726,31 @@ public:
       if (isScalarRhs) {
         // Handle scalar operand - need to use unary scalar ops
         const auto dstIdx = adaptor.getLhs();
+        auto loc = op->getLoc();
 
-        // Get the scalar value from the constant
-        if (auto constOp =
-                adaptor.getRhs()
-                    .template getDefiningOp<mlir::arith::ConstantOp>()) {
-          float floatVal = 0.0f;
-          if (auto floatAttr =
-                  mlir::dyn_cast<mlir::FloatAttr>(constOp.getValue())) {
-            floatVal = floatAttr.getValueAsDouble();
-          } else if (auto intAttr = mlir::dyn_cast<mlir::IntegerAttr>(
-                         constOp.getValue())) {
-            floatVal = static_cast<float>(intAttr.getInt());
-          }
-
-          // Create the appropriate unary scalar op based on the D2M op type
-          if constexpr (std::is_same_v<ConcreteOp, d2m::TileAddOp>) {
-            uint32_t bits;
-            std::memcpy(&bits, &floatVal, sizeof(float));
-            auto scalarParam = i32(rewriter, op->getLoc(), bits);
-            rewriter.create<ttkernel::AddUnaryTileOp>(op->getLoc(), dstIdx,
-                                                      scalarParam);
-          } else if constexpr (std::is_same_v<ConcreteOp, d2m::TileSubOp>) {
-            uint32_t bits;
-            std::memcpy(&bits, &floatVal, sizeof(float));
-            auto scalarParam = i32(rewriter, op->getLoc(), bits);
-            rewriter.create<ttkernel::SubUnaryTileOp>(op->getLoc(), dstIdx,
-                                                      scalarParam);
-          } else if constexpr (std::is_same_v<ConcreteOp, d2m::TileMulOp>) {
-            uint32_t bits;
-            std::memcpy(&bits, &floatVal, sizeof(float));
-            auto scalarParam = i32(rewriter, op->getLoc(), bits);
-            rewriter.create<ttkernel::MulUnaryTileOp>(op->getLoc(), dstIdx,
-                                                      scalarParam);
-          } else if constexpr (std::is_same_v<ConcreteOp, d2m::TileDivOp>) {
-            uint32_t bits;
-            std::memcpy(&bits, &floatVal, sizeof(float));
-            auto scalarParam = i32(rewriter, op->getLoc(), bits);
-            rewriter.create<ttkernel::DivUnaryTileOp>(op->getLoc(), dstIdx,
-                                                      scalarParam);
-          } else if constexpr (std::is_same_v<ConcreteOp, d2m::TilePowOp>) {
-            // For power, cast to integer directly (not reinterpret bits)
-            uint32_t intVal = static_cast<uint32_t>(floatVal);
-            auto scalarParam = i32(rewriter, op->getLoc(), intVal);
-            rewriter.create<ttkernel::PowUnaryTileOp>(op->getLoc(), dstIdx,
-                                                      scalarParam);
-          }
+        // Create the appropriate unary scalar op based on the D2M op type
+        if constexpr (std::is_same_v<ConcreteOp, d2m::TileAddOp>) {
+          // Bitcast the scalar value to i32 to pass as parameter
+          auto scalarParam = rewriter.create<arith::BitcastOp>(
+              loc, rewriter.getI32Type(), adaptor.getRhs());
+          rewriter.create<ttkernel::AddUnaryTileOp>(loc, dstIdx, scalarParam);
+        } else if constexpr (std::is_same_v<ConcreteOp, d2m::TileSubOp>) {
+          auto scalarParam = rewriter.create<arith::BitcastOp>(
+              loc, rewriter.getI32Type(), adaptor.getRhs());
+          rewriter.create<ttkernel::SubUnaryTileOp>(loc, dstIdx, scalarParam);
+        } else if constexpr (std::is_same_v<ConcreteOp, d2m::TileMulOp>) {
+          auto scalarParam = rewriter.create<arith::BitcastOp>(
+              loc, rewriter.getI32Type(), adaptor.getRhs());
+          rewriter.create<ttkernel::MulUnaryTileOp>(loc, dstIdx, scalarParam);
+        } else if constexpr (std::is_same_v<ConcreteOp, d2m::TileDivOp>) {
+          auto scalarParam = rewriter.create<arith::BitcastOp>(
+              loc, rewriter.getI32Type(), adaptor.getRhs());
+          rewriter.create<ttkernel::DivUnaryTileOp>(loc, dstIdx, scalarParam);
+        } else if constexpr (std::is_same_v<ConcreteOp, d2m::TilePowOp>) {
+          // For power, convert float value to integer (not bitcast)
+          auto scalarParam = rewriter.create<arith::FPToSIOp>(
+              loc, rewriter.getI32Type(), adaptor.getRhs());
+          rewriter.create<ttkernel::PowUnaryTileOp>(loc, dstIdx, scalarParam);
         }
       } else {
         // Binary tile operation
--- a/lib/Dialect/D2M/Transforms/InsertDstRegisterAccess.cpp
+++ b/lib/Dialect/D2M/Transforms/InsertDstRegisterAccess.cpp
@@ -312,17 +312,9 @@ public:
 
       // Collect loads to this op.
       for (int64_t operandIdx : computeOp.getOperandsLoadFromDstRegister()) {
-        // For tile+scalar ops, skip loading scalar operands (they're not tiles)
-        if (computeOp.supportsTileOrScalarRhs() && operandIdx == 1) {
-          auto rhsOperand = computeOp->getOperand(1);
-          auto rhsType = rhsOperand.getType();
-          // Check if rhs is a scalar: NOT a tile AND is int/float
-          bool isScalar =
-              !llvm::isa<ttcore::TileType>(rhsType) && rhsType.isIntOrFloat();
-          if (isScalar) {
-            // Skip scalar operands - they don't need to be loaded from dst
-            continue;
-          }
+        // Skip scalar operands - they don't need to be loaded from dst
+        if (computeOp.isScalarOperand(operandIdx)) {
+          continue;
         }
 
         if (auto potentialLoad = computeOp->getOperand(operandIdx)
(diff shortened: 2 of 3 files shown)


lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
```diff
@@ -360,13 +382,90 @@ class D2MNamedElementwiseRewriter final
...
+      // Check if this is a scalar tensor (rank 0)
+      if (convertedTensorType && convertedTensorType.getRank() == 0) {
+        if constexpr (!supportsScalar) {
+          return rewriter.notifyMatchFailure(
+              op, "Scalar operands only supported for add, multiply, subtract, "
+                  "and div ops");
+        }
+
+        // Extract the scalar constant value from the converted operand
+        auto convertedDefiningOp = origInput.getDefiningOp();
```
vmilosevic: :warning: **llvm\-qualified\-auto** :warning:
`` auto convertedDefiningOp `` can be declared as `` auto *convertedDefiningOp ``
```suggestion
        auto *convertedDefiningOp = origInput.getDefiningOp();
```

include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
```diff
@@ -106,6 +106,16 @@ def D2M_TileAddOp : D2M_GenericRegionComputeBinaryDstOp<"tile_add", D2M_TileComp
...
 
+def D2M_TileAddScalarOp : D2M_GenericRegionComputeUnaryDstOp<"tile_add_scalar", D2M_TileComputeOpTraitBundle> {
+    let summary = "D2M Tile Add Scalar Op";
+    let description = [{
+        The `tile_add_scalar` operation adds a scalar value to each element in the input tile.
+    }];
+
+    let arguments = (ins TTCore_Tile:$input, F32Attr:$scalar);
+    let results = (outs TTCore_Tile:$result);
+}
```
nsmithtt: Sorry, I think instead of introducing new ops, could we augment the existing ops?

```tablegen
def TileOrScalar = AnyTypeOf<[TTCore_Tile, FloatType, IntegerType]>;

def D2M_TileAddOp ... {
    let arguments = (ins TTCore_Tile:$lhs, TileOrScalar:$rhs);
}
```

Note: this will result in IR like:
```mlir
%t : ttcore.tile<32x32, f32>
%five = arith.constant 5.0 : f32
d2m.tile_add %t, %five
```

dlokeTT: Yeah, I was going to add that on. Just did this as a first step and wanted to run CI on it to make sure this doesn't cause issues with other parts of the codebase before moving forward, so I put a pr up.

nsmithtt: Cool sounds good!

dlokeTT: Although is there a reason for augmenting at this level? I figured that it makes sense at the ttir level to have them be one op like a high level torch.add and at ttkernel they have to be different due to different llk calls but not sure what the d2m level is supposed to model exactly.

Also, having them as two different d2m ops currently works with the abstraction we have for unary ops being in place and binary allocating a new spot for dstRegister pass. Not that this is a big deal since we can probably handle this somewhere in the interface or something + that whole thing could be gone very soon with the changes to insertDst

nsmithtt: Yeah d2m is meant to be an intermediate representation, so generally it's better to air on the side of making things easier to read / write / more flexible.  We can hide the verbosity of having multiple ops until we lower to ttkernel.  I agree it complicates the conversion to ttkernel but it should be mechanical.

lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
```diff
@@ -360,13 +385,105 @@ class D2MNamedElementwiseRewriter final
...
+        // Handle d2m.full (already converted)
+        if (auto d2mFullOp =
+                mlir::dyn_cast_or_null<d2m::FullOp>(convertedDefiningOp)) {
+          auto fillValueAttr = d2mFullOp.getFillValue();
+          if (auto floatAttr = mlir::dyn_cast<FloatAttr>(fillValueAttr)) {
+            scalarAttr = floatAttr;
+            // Track the d2m.full op for erasure
+            scalarOpsToErase.push_back(d2mFullOp);
+            continue; // Skip adding this to tensorInputs
+          } else if (auto intAttr =
```
vmilosevic: :warning: **llvm\-else\-after\-return** :warning:
do not use `` else `` after `` continue ``
```suggestion
          } if (auto intAttr =
```

lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
```diff
@@ -360,13 +385,105 @@ class D2MNamedElementwiseRewriter final
...
+
+        // Handle ttir.full (not yet converted - shouldn't happen in adaptor)
+        if (auto fullOp =
+                mlir::dyn_cast_or_null<ttir::FullOp>(convertedDefiningOp)) {
+          auto fillValueAttr = fullOp.getFillValue();
+          if (auto floatAttr = mlir::dyn_cast<FloatAttr>(fillValueAttr)) {
+            scalarAttr = floatAttr;
+            scalarOpsToErase.push_back(fullOp);
+            continue; // Skip adding this to tensorInputs
+          } else if (auto intAttr =
```
vmilosevic: :warning: **llvm\-else\-after\-return** :warning:
do not use `` else `` after `` continue ``
```suggestion
          } if (auto intAttr =
```

include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
dlokeTT: This is ready for review with the changes to keep the binary ops the same at the d2m level

include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
```diff
@@ -74,6 +74,29 @@ class D2M_GenericRegionComputeBinaryDstOp<string mnemonic, list<Trait> traits =
...
   let extraClassDeclaration = [{
     mlir::SmallVector<int64_t> getOperandsLoadFromDstRegister() {
+      // Both operands are tiles, load both from DST
+      return {0, 1};
+    }
+  }];
+}
+
+// Binary ops that support tile-tile OR tile-scalar operations
+class D2M_GenericRegionComputeBinaryOrUnaryDstOp<string mnemonic, list<Trait> traits = []> :
```
nsmithtt: Can we call this `GenericRegionComputeBinaryWithScalarDstOp`?  To me it doesn't make sense to call them unary ops, despite llk using that convention.

include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
```diff
@@ -74,6 +74,29 @@ class D2M_GenericRegionComputeBinaryDstOp<string mnemonic, list<Trait> traits =
...
+    D2M_GenericRegionComputeOp<mnemonic, traits> {
+
+  let extraClassDeclaration = [{
+    bool supportsTileOrScalarRhs() { return true; }
+
+    mlir::SmallVector<int64_t> getOperandsLoadFromDstRegister() {
+      // If the RHS is a scalar type, it doesn't come from DST register
+      if (this->getNumOperands() >= 2) {
```
nsmithtt: Perhaps this should be an assert?

lib/Dialect/D2M/IR/D2MOps.cpp
```diff
@@ -146,6 +146,18 @@ ::mlir::LogicalResult d2m::FullOp::verify() {
...
+                                              mlir::MLIRContext *context) {
+  // Remove unused d2m.full ops (dead code elimination).
+  patterns.add(+[](FullOp op, mlir::PatternRewriter &rewriter) {
+    if (op.getResult().use_empty()) {
+      rewriter.eraseOp(op);
+      return success();
+    }
+    return failure();
+  });
+}
```
nsmithtt: Instead of this we should just mark the FullOp with the `Pure` trait, which will do this for us.

dlokeTT:
> I think it's better if we don't change TTIR and have a pass that promotes unit tensors separately. That way dialect conversion doesn't have to care about special casing scalars.

In my head I was thinking that it should be intrinsic to the dialect because as a high level dialect the ttir op should be able to support tensor or scalar instead of written as an optimization pass. Also, the ttnn pipeline can support this change as well in the future since these same ttnn ops support scalar variants instead of them writing a very similar optimization pass. Could be imagining this completely wrong though.

include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
dlokeTT: Hmm, this kind of implies it is a binary op that uses the scalar instead of representing that it could be tile-tile or tile-scalar. I just stuck with llk because I couldn't think of anything better. Maybe `GenericRegionComputeBinaryWithScalarVariantDstOp`?

nsmithtt: Sounds good!

nsmithtt:
> > I think it's better if we don't change TTIR and have a pass that promotes unit tensors separately. That way dialect conversion doesn't have to care about special casing scalars.
> 
> In my head I was thinking that it should be intrinsic to the dialect because as a high level dialect the ttir op should be able to support tensor or scalar instead of written as an optimization pass. Also, the ttnn pipeline can support this change as well in the future since these same ttnn ops support scalar variants instead of them writing a very similar optimization pass. Could be imagining this completely wrong though.

I see your argument, but we already have a well established was of representing scalars in stablehlo/ttir/ttnn (ttnn does have native scalar support, see `TTNN_ElementwiseBinaryCompositeScalarOp`).  TTNN is already doing the thing I proposed above where they pattern match unit tensors and promote them to be scalars.  TTIR and stablehlo for example are meant to be higher level, it's more generalized to have them only ever operate on tensors.

dlokeTT:
> I see your argument, but we already have a well established was of representing scalars in stablehlo/ttir/ttnn (ttnn does have native scalar support, see `TTNN_ElementwiseBinaryCompositeScalarOp`). 

Oh cool, I was looking thru some ttnn model graphs and didn't see any examples of `TTNN_ElementwiseBinaryCompositeScalarOp`. I guess they only do it only for power and not add, mul, div and sub. 

> TTNN is already doing the thing I proposed above where they pattern match unit tensors and promote them to be scalars. 

I see it now. They do it in `TTIRtoTTNN` but they also have a special pattern for just the `powOp` since that's the only one they support currently whereas we want to have it for a lot more ops so it makes sense to make it it's own pass. 

> TTIR and stablehlo for example are meant to be higher level, it's more generalized to have them only ever operate on tensors.

Oh ok, I kind of imagined it the other way but this makes sense too.

Cool, thanks for the discussion, I'll go ahead with your suggestions.

dlokeTT:
Updated following Nick's suggestions, I think this is ready for review again.

lib/Dialect/D2M/Transforms/InsertDstRegisterAccess.cpp
```diff
@@ -261,6 +262,19 @@ struct D2MInsertDstRegisterAccessRewriter final
 
       // Collect loads to this op.
       for (int64_t operandIdx : computeOp.getOperandsLoadFromDstRegister()) {
+        // For tile+scalar ops, skip loading scalar operands (they're not tiles)
+        if (computeOp.supportsTileOrScalarRhs() && operandIdx == 1) {
```
nsmithtt: Perhaps we should have an op interface method so this code (and the checks below) could read:
```
if (computeOp.isScalarOperand(operandIdx)) {
  continue;
}
```

lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
```diff
@@ -701,13 +719,70 @@ class D2MSFPUOpsRewriter : public OpConversionPattern<ConcreteOp> {
...
+                adaptor.getRhs()
+                    .template getDefiningOp<mlir::arith::ConstantOp>()) {
+          float floatVal = 0.0f;
+          if (auto floatAttr =
+                  mlir::dyn_cast<mlir::FloatAttr>(constOp.getValue())) {
+            floatVal = floatAttr.getValueAsDouble();
+          } else if (auto intAttr = mlir::dyn_cast<mlir::IntegerAttr>(
+                         constOp.getValue())) {
+            floatVal = static_cast<float>(intAttr.getInt());
+          }
```
nsmithtt: Instead of doing the cast statically in the compiler (which wouldn't support scalars coming from runtime values if we wanted to eventually support that), we should just do this in the IR via [arith.bitcast](https://mlir.llvm.org/docs/Dialects/ArithOps/#arithbitcast-arithbitcastop).

```c++
auto scalarParam = rewriter.create<arith::BitcastOp>(loc, rewriter.getI32Type(), adaptor.getRhs());
...
```

This avoids having to lookup the attribute and do the conversion statically.

nsmithtt (commented):
This is looking great! Two minor comments inline.

lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
```diff
@@ -701,13 +719,49 @@ class D2MSFPUOpsRewriter : public OpConversionPattern<ConcreteOp> {
...
+        } else if constexpr (std::is_same_v<ConcreteOp, d2m::TileDivOp>) {
+          auto scalarParam = rewriter.create<arith::BitcastOp>(
+              loc, rewriter.getI32Type(), adaptor.getRhs());
+          rewriter.create<ttkernel::DivUnaryTileOp>(loc, dstIdx, scalarParam);
+        } else if constexpr (std::is_same_v<ConcreteOp, d2m::TilePowOp>) {
+          // For power, convert float value to integer (not bitcast)
+          auto scalarParam = rewriter.create<arith::FPToSIOp>(
+              loc, rewriter.getI32Type(), adaptor.getRhs());
+          rewriter.create<ttkernel::PowUnaryTileOp>(loc, dstIdx, scalarParam);
+        }
```
nsmithtt, line 754: Instead of having a big if/else chain, we could add another tuple element to the map, in the same way we get `SFPUOp`, we could have a new tuple element `SFPUScalarOp` or something. @wenbinlyuTT could be a good person to sync with.

dlokeTT: Definitely agree with this especially as we start adding more support for binary ops with scalar variants. Wanted to just get this and part 2 in before the break so I added an issue here https://github.com/tenstorrent/tt-mlir/issues/6008 and will get to it once were back

PR #5933, opened 2025-11-19
Updated ttir builder for some ops + fix conv and pooling golden

dlokeTT:
Title, needed for running resnet e2e thru the load_mlir flow


dlokeTT:
Just doing this to run ci, not ready for review yet

Issue #3222, opened 2025-05-06
Share docker image between frontend projects


jmcgrathTT:
Use docker image created in tt-mlir (with prebuilt LLVM) as base image in frontends to optimize image build time.

nsumrakTT:
tt-xla now uses tt-mlir docker images.
Both are optimized (~25min for tt-mlir and ~15min for tt-xla) but stacking of images is not fully implemented.
The agreement is not to make things more complex while the current setup works well enough.

PR #5934, opened 2025-11-19
[CI] Optimize docker build

nsumrakTT:
### Ticket
https://github.com/tenstorrent/tt-mlir/issues/3222

### Problem description
Docker build can be optimized and faster

### What's changed
Use docker copy whenever is possible.
Remove caches and intermediate files to lower docker size.
Use staged build process so things are not built twice (or more).
Build only docker images that are necessary at the moment.

### Checklist

CI Wheel Build docker is tested in nightly:
https://github.com/tenstorrent/tt-mlir/actions/runs/19540358600

### DIFF
--- a/.github/Dockerfile.cibuildwheel
+++ b/.github/Dockerfile.cibuildwheel
@@ -1,32 +1,6 @@
 ARG FROM_TAG=latest
 
-FROM quay.io/pypa/manylinux_2_34_x86_64 AS toolchain-source
-SHELL ["/bin/bash", "-c"]
-
-# Create a directory for the build and toolchain
-ENV PROJECT_NAME=tt-mlir
-ENV BUILD_DIR=/home/build
-ENV TTMLIR_TOOLCHAIN_DIR=/opt/ttmlir-toolchain
-
-RUN echo "Building $PROJECT_NAME"
-
-RUN mkdir -p $BUILD_DIR && \
-    mkdir -p $TTMLIR_TOOLCHAIN_DIR
-
-# Copy the project to the container
-ADD . $BUILD_DIR/$PROJECT_NAME
-
-# Build the toolchain
-WORKDIR $BUILD_DIR/$PROJECT_NAME
-
-# Show last commit
-RUN git log -1
-
-RUN cmake -B env/build env && \
-    cmake --build env/build
-
-# Final stage
-FROM quay.io/pypa/manylinux_2_34_x86_64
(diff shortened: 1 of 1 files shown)


.github/build-docker-images.sh
```diff
@@ -9,6 +9,12 @@ set -e
 CHECK_ONLY=false
 if [[ "$1" == "--check-only" ]]; then
     CHECK_ONLY=true
+    dockbuild=$2
+else
+    dockbuild=$1
+fi
+if [ -z "$dockbuild" ]; then
+    dockbuild="all"
```
vvukomanTT: Could we reorder arguments for this script, `dockbuild` can be the 1st argument and `--check-only` can be the 2nd argument.
It makes sense to me because `dockbuild` is mandatory and we would avoid having arguments in different positions in different situations.

.github/Dockerfile.cibuildwheel
```diff
@@ -29,29 +33,13 @@ ENV NINJA_VERSION="1.11.1"
...
-# Build the toolchain
-WORKDIR $BUILD_DIR/$PROJECT_NAME
-
-# Show last 2 commits
-RUN git log -2
-
-RUN cmake -B env/build env && \
-    cmake --build env/build
+# Copy toolchain from CI image
+COPY --from=toolchain-source $TTMLIR_TOOLCHAIN_DIR $TTMLIR_TOOLCHAIN_DIR
```
nsmithtt, line 71: I'm not sure the changes to this file are legal.  The cibuildwheel is meant to use the base docker image with a very specific toolchain that everything must be compiled with. See https://github.com/pypa/manylinux

nsumrakTT: Thanks, haven't thought about that, I'll revert to env build just make it clean inside docker.

nsmithtt: Does this file really need to change at all?  It seems like this copy is still here

nsumrakTT: Yes, because intermediate build artifacts stay in docker and we don't want them.
This is a waste of ~1GB that is stored and transferred each time docker is needed.
With multistage build process we create final docker that has only files we need.

nsmithtt: I see, makes sense, thanks!

nsmithtt (approved):

PR #5938, opened 2025-11-19
Fix test output

nsumrakTT:
### Ticket
slack

### Problem description
Test output is lost in some situations.

### What's changed
Use os.system instead of subprocess.run to prevent loss of output when process abruptly exits.

### Checklist


nsumrakTT:
There is no difference on output.
The problem is, it seems, that tests output its errors to stdout which is discarded by pytest, instead of stderr. To see error stream builder.sh script need to add -s parameter but the output becomes huge, garbled with debug messages to the level of unusable - it not possible to easily see which tests are run, which passed and which failed.
So, I'm closing this PR.

PR #5941, opened 2025-11-19
Remove workaround for all_reduce and call TTNN API directly

### LINKED ISSUE #4999: Adopt all_reduce_async
Check the functionality and completeness of the all_reduce_async op. 
Determine whether it is ready to replace the current all_reduce workaround. 
If it is ready, introduce it into the runtime and dialect.

Must-have generality items:
 - Any rank
 - Any shape - Especially with not a multiple of tile size on last two dim
 - fp32 / bf16 support
 - Any cluster axis

Good to check:
 - Supported reduce types
 - Cases RS + AG decomposition cannot supports

hkwonTT:
### Ticket
closes #4999

### Problem description
The ttnn.all_reduce was decomposed into other operations (such as reduce_scatter + all_gather) because the backend did not yet fully support the general case.
Now that the TTNN backend does provide full generality for all_reduce, the decomposition workaround is no longer needed and should be removed.

### What's changed
 - Removed the decomposition pattern for all_reduce in the workaround logic.
 - Updated conversion/emission paths so that ttnn.all_reduce map directly to the TT-NN API.
 - Updated builder testcase

### Checklist


AleksKnezevic (approved):

runtime/lib/ttnn/operations/ccl/all_reduce.cpp
```diff
@@ -0,0 +1,40 @@
+// SPDX-FileCopyrightText: (c) 2024 Tenstorrent AI ULC
```
hkwonTT:
```suggestion
// SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
```

runtime/include/tt/runtime/detail/ttnn/ttnn.h
```diff
@@ -15,6 +15,7 @@
 #include "tt-metalium/program_cache.hpp"
 #include "ttnn/device.hpp"
 #include "ttnn/events.hpp"
+#include "ttnn/global_semaphore.hpp"
```
wooseokTT: @hkwonTT  Just curious, what is this for?

wooseokTT (approved):
LGTM

runtime/include/tt/runtime/detail/ttnn/ttnn.h
hkwonTT: My bad. This line left over from rebasing my old commit. It should have been removed, but it wasn't. I'll update the code.

wooseokTT: Got it.

PR #5945, opened 2025-11-19
[explorer] Improve tests for tt-explorer

madcampos:
### Ticket
N/A

### Problem description
We have gaps in tt-explorer's testing suite.

### What's changed
Added tests for the `preload` command and a simple test for the case of `convert` returning a collection instead of a list of graphs.

### Checklist

Resolves: https://github.com/tenstorrent/tt-mlir/issues/6034
Depends on: https://github.com/tenstorrent/model-explorer/pull/80

### DIFF
--- a/test/ttmlir/Explorer/graph_collections/collection1.mlir/col1-graph1.mlir
+++ b/test/ttmlir/Explorer/graph_collections/collection1.mlir/col1-graph1.mlir
@@ -1,2 +1,3 @@
+// RUN: ttmlir-opt --version
 module {
 }
--- a/test/ttmlir/Explorer/graph_collections/loose_file.mlir
+++ b/test/ttmlir/Explorer/graph_collections/loose_file.mlir
@@ -1,2 +1,3 @@
+// RUN: ttmlir-opt --version
 module {
 }
--- a/test/ttmlir/Explorer/graph_collections/collection1.mlir/col1-graph2.mlir
+++ b/test/ttmlir/Explorer/graph_collections/collection1.mlir/col1-graph2.mlir
@@ -1,2 +1,3 @@
+// RUN: ttmlir-opt --version
 module {
 }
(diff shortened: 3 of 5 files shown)


test/ttmlir/Explorer/graph_collections/collection1.mlir/col1-graph1.mlir
```diff
@@ -0,0 +1,11 @@
+// RUN: ttmlir-opt --ttir-to-ttnn-backend-pipeline="system-desc-path=%system_desc_path%" -o %t.mlir %s
```
brataTT: with [[existing mlir used in explorer tests]](https://github.com/tenstorrent/tt-mlir/blob/main/test/ttmlir/Explorer/load/forward_and_backward.mlir#L1) as reference, we only need to check if the mlir is valid.
so, the filecheck and ` // CHECK: = "ttnn.slice_static"` while ok to have, is not necessary

test/ttmlir/Explorer/graph_collections/loose_file.mlir
brataTT, line 1: I assume we only care about the following folder structure here and not the actual mlir contents, so them having the same graph in all of them won't matter?
- col1.mlir
  - graph1.mlir
  - graph2.mlir
- col2.mlir
  - graph1.mlir
  - graph2.mlir
- loose.mlir

madcampos: Exactly. It is only for testing the server logic, if it reads the tree structure correctly.

test/ttmlir/Explorer/graph_collections/collection1.mlir/col1-graph1.mlir
madcampos: Sorry, I don't quite follow... It means we can remove this line?

brataTT: I meant, keep only this line and remove the FileCheck and ttmlir-translate lines below.
Also remove the CHECK comment in line 7.

Issue #5931, opened 2025-11-18
[ttnn.jit] test failures when running on `--model x2` ird machines (2 harvested rows)


vtangTT:
### Describe the bug

Unsure of the entire scope of the issue. Can confirm for sure that `test_control_flow.py` is failing for many grid shapes with 
```
E       RuntimeError: TT_FATAL @ /localdev/vtang/tt-mlir/third_party/tt-metal/src/tt-metal/tt_metal/impl/allocator/bank_manager.cpp:405: num_shards.value() <= num_compute_banks
E       info:
E       Expected number of shards 64 to be less than or equal to total number of L1 banks 56 in compute cores
E       backtrace:
```

### Steps to reproduce the issue

1. Reserve ird with `ird reserve --docker-image mlir --cluster tt_aus wormhole_b0 --model x2`
2. build ttnn_jit
3. run pytests

vtangTT:
see test/python/golden/conftest.py which default opens the device with mesh_options.dispatch_core_type = ttrt.runtime.DispatchCoreType.ETH

vtangTT:
The problem is *most* likely because we are opening device without setting `dispatch_core_config` meaning it'll deafult to `WORKER`. 

On an N300, this will mean less compute cores, so we get a 7x8 grid.

Fix should be to set `dispatch_core_config` in `test/ttnn-jit/conftest.py` defaulting to eth dispatch.

This will hold true except once we move to tg and p150 which requires us to disable eth dispatch

vtangTT:
should be fixed with https://github.com/tenstorrent/tt-mlir/pull/5948

Issue #5057, opened 2025-09-22
[ttnn.jit] - Dynamically query runtime artifacts as part of jit runtime.


vtangTT:
We need system_desc to run D2M compiler. We do not want to package ttrt into the ttnn-jit package. As a solution to this problem, replicate how `ttrt query --save-artifacts` gets system descriptor during jit runtime, and then set the corresponding env var. `ttrt` itself uses mlir runtime bindings which should already be packaged into the ttnn-jit package and ready to use

Starting point:
- how ttrt queries system descriptor: https://github.com/tenstorrent/tt-mlir/blob/84cc86cf1111fd47bb6663ed010ace38d0a55095/tools/ttrt/common/query.py#L125

vtangTT:
https://github.com/tenstorrent/tt-mlir/pull/5948

jojo-771771:
I can work on this. Please assign!

jojo-771771:
I can work on this. Please assign!

jojo-771771:
I can work on this. Please assign!

PR #5948, opened 2025-11-19
[ttnn.jit] Dynamically query for system desc

vtangTT:
### Ticket
https://github.com/tenstorrent/tt-mlir/issues/5057
https://github.com/tenstorrent/tt-mlir/issues/5931

### Problem description
as descrbied in issue

### What's changed
- use `get_current_system_desc` , passing input tensor device if available
- on invocation, the `JitFunction` will check `SYSTEM_DESC_PATH` env var, if not set, it will run `_query_and_save_system_desc` to get it's own system desc instead of just error'ing out.
- helper `_get_dispatch_core_type` and `_get_cluster_type` to get `DispatchCoreType` based off what the cluster is
- in `test/ttnn-jit/conftest.py` set `DispatchCoreType` to `WORKER` for p150, if not set `ETH` for device init

### Checklist

### DIFF
--- a/tools/ttnn-jit/_src/jit.py
+++ b/tools/ttnn-jit/_src/jit.py
@@ -12,8 +12,8 @@ from ttmlir.passes import (
     ttnn_to_ttmetal_pipeline,
 )
 
-from ttnn_jit._src.utils import _cleanup_source_code, _get_dispatch_core_type
-from ttnn_jit._src.dispatch_op import _run_binary, _run_binary_from_capsule
+from ttnn_jit._src.utils import cleanup_source_code, get_dispatch_core_type
+from ttnn_jit._src.dispatch_op import run_binary, run_binary_from_capsule
 from ttnn_jit._src import JitCache
 from ttnn_jit._src.ir_generator import generate_ir
 from ttnn_jit._src import (
@@ -35,7 +35,7 @@ class JitFunction:
         graph_capture: bool,
     ):
         self.func = func
-        self.source_code = _cleanup_source_code(func)
+        self.source_code = cleanup_source_code(func)
         self.max_grid = max_grid
         self.compile_only = compile_only
         self.debug = debug
@@ -58,7 +58,7 @@ class JitFunction:
         Uses the MLIR runtime bindings directly, replicating the logic from
         ttrt query --save-artifacts.
         """
-        dispatch_core_type = _get_dispatch_core_type()
+        dispatch_core_type = get_dispatch_core_type()
         try:
             # Use input tensor device to query if available.
             if ttnn_device:
@@ -109,7 +109,7 @@ class JitFunction:
         # Cache hit, no need to compile.
         if self.cache and self.cache.contains(*args):
             fb_binary = self.cache.get(*args)
-            return _run_binary(fb_binary, args)
+            return run_binary(fb_binary, args)
 
         ir = generate_ir(
             self.graph_capture,
@@ -134,14 +134,14 @@ class JitFunction:
             fb_binary = self.cache.compile_and_insert(
                 str(ir), options, self.debug, *args
             )
-            return _run_binary(fb_binary, args)
+            return run_binary(fb_binary, args)
 
         ttnn_to_ttmetal_pipeline(ir, options)
         if self.debug:
             print("---- IR Dump after ttnn_to_ttmetal_pipeline ----")
             print(ir)
         fb_capsule = ttnn_to_flatbuffer_bin(ir)
-        return _run_binary_from_capsule(fb_capsule, args)
+        return run_binary_from_capsule(fb_capsule, args)
 
     @property
     def num_entries(self):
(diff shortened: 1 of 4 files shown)


tools/ttnn-jit/_src/jit.py
```diff
@@ -53,8 +57,55 @@ def __init__(
...
+        Uses the MLIR runtime bindings directly, replicating the logic from
+        ttrt query --save-artifacts.
+
+        Args:
+            input_tensors: Optional list of input tensors. If provided, will use the
+                          device from the first tensor to avoid creating a new device.
+        """
+        try:
+            # Query system descriptor with ETH dispatch (default, matching ttrt behavior)
+            dispatch_core_type = DispatchCoreType.ETH
```
xanderchin: We should figure out if this should be set to ETH for all boards/products including BH

vtangTT: now setting the core type based on what cluster type was queried, see `_get_dispatch_core_type`

tools/ttnn-jit/_src/utils.py
```diff
@@ -153,3 +155,23 @@ def _get_collapsed_linear_affine_map(
...
+    cluster_type = _get_cluster_type()
+    match cluster_type:
+        case ttnn.cluster.ClusterType.N150:
+            dispatch_core_type = DispatchCoreType.ETH
+        case ttnn.cluster.ClusterType.N300:
+            dispatch_core_type = DispatchCoreType.ETH
+        case ttnn.cluster.ClusterType.P150:
+            dispatch_core_type = DispatchCoreType.WORKER
+        case ttnn.cluster.ClusterType.T3K:
+            dispatch_core_type = DispatchCoreType.WORKER
```
vtangTT: @jnie-TT do you know if T3K should be a worker core type or eth?

CC: @nsmithtt

jnie-TT: AFAIK eth is good for t3k, but for galaxy we need worker

nsmithtt: @jnie-TT don't we need worker for t3k? It needs the eth cores for multi chip communication right?

jnie-TT: I think if you have remote chips that are not connected via PCIe, you need ethernet dispatch. For wormhole t3k it's a cluster of n300s so we need ETH, galaxy is all PCIe connected so we need worker

arminaleTT (approved):

tools/ttnn-jit/_src/jit.py
```diff
@@ -53,8 +53,48 @@ def __init__(
         # Hashing based off runtime tensor metadata.
         self.cache = JitCache(64) if enable_cache else None
 
+    def _query_and_save_system_desc(self, input_tensors=None):
```
vroubtsovTT: Looks odd that a method to query and save a "system descriptor" would declare that it can take some input tensors as input. Why not use an optional parameter that is whatever `ttnn_device`  below ends up being instead? That kind of API would make more sense to me.

tools/ttnn-jit/_src/jit.py
```diff
@@ -15,10 +12,14 @@
     ttnn_to_ttmetal_pipeline,
 )
 
-from ttnn_jit._src.utils import _cleanup_source_code
+from ttnn_jit._src.utils import _cleanup_source_code, _get_dispatch_core_type
```
vroubtsovTT: I believe Python convention is to use leading underscores for "private" symbols. I.e. it would be bad practice to then go ahead and import those "private" symbols the way is done here... perhaps `_get_dispatch_core_type` should not pretend to be "private"?

vtangTT: Using an underscore here since these are internal helper functions that should only really be used within python `_src` files.

Although, I do admit the convention is hazy and non-consistent through the project right now.

tools/ttnn-jit/_src/jit.py
vtangTT: Agreed. Passing device handle directly now if available ty 🙏

tools/ttnn-jit/_src/jit.py
vroubtsovTT: I find it odd that `_src` is named that way also.

But going back to `_get_dispatch_core_type`, using a leading underscore in Python has always been a "private" indicator to me, i.e. a method/attribute to be used internally within a class or a module. Certainly not to be imported from outside (I believe `import *` will not import such names for that reason).

I propose we rename this to `get_dispatch_core_type`.

vroubtsovTT (changes requested):

vtangTT:
@vroubtsovTT I see your point, thank you for bringing this up. I've renamed all the _ leading functions that are exposed now, cleaning up in other places as well.

I've named the python `_src` directory as it is following how jax has named their repo for python internal implementation details. [link](https://github.com/jax-ml/jax/tree/main/jax/_src)

vroubtsovTT:
> I've named the python `_src` directory as it is following how jax has named their repo for python internal implementation details. [link](https://github.com/jax-ml/jax/tree/main/jax/_src)

Thanks. `_src` might have been ok if there was a desire to create a "non-public" module.

PR #5950, opened 2025-11-20
loosening generic/dma verifiers for dma-only generics

anuragsingh-tt:
### What's changed
To enable DMA-only generics with different sizes for input/output shards, the GenericOp verifier has been refactored to exit early.

### DIFF
--- a/lib/Dialect/D2M/IR/D2MGenericRegionOps.cpp
+++ b/lib/Dialect/D2M/IR/D2MGenericRegionOps.cpp
@@ -124,6 +124,12 @@ mlir::LogicalResult DMAOp::verify() {
     return emitOpError("operands must have the same post-index rank");
   }
 
+  if (!std::equal(srcType.getShape().begin() + srcIndices,
+                  srcType.getShape().end(),
+                  dstType.getShape().begin() + dstIndices)) {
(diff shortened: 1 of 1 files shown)


lib/Dialect/D2M/IR/D2MGenericRegionOps.cpp
```diff
@@ -124,12 +124,6 @@ mlir::LogicalResult DMAOp::verify() {
     return emitOpError("operands must have the same post-index rank");
   }
 
-  if (!std::equal(srcType.getShape().begin() + srcIndices,
-                  srcType.getShape().end(),
-                  dstType.getShape().begin() + dstIndices)) {
-    return emitOpError("operands must have the same post-index shape");
-  }
-
```
nsmithtt, line 132: Hey @anuragsingh-tt, was thinking a bit more about this one, do you have a case where this verification must be relaxed?  At first glance it seems like the DMA src/dst shapes should be equal.

nsmithtt:
@anuragsingh-tt, let me know if we still need this change after our convo on Thursday.  i.e. the view should produce the input in the post-tm shape, so the input and output to the generic should be identical.

PR #5961, opened 2025-11-20
D2M: Implicit broadcast for 2D inputs

wenbinlyuTT:
### Ticket
Related-To: #3023

### What's changed
* When implicit bcast is detected in eltwise binary ops, an `unary_bcast` op is inserted to do sub-tile bcast on the row/col/scalar operand(s).
* Additionally, tensor collapsing is disabled for implicit bcast situations.
* The insert DST access pass is extended to support broadcast-style guard indices, in addition to the accumulation-style guard indices for reduce & matmul.
* All guards now reside in the same cloned affine load loop, each has its own condition for the guard.

### Known limitations
* 2D only.
* Fusion with other eltwise ops is disabled, otherwise the fusion pass generates ill-formed IR that triggers assertions in downstream passes. #5927 wasn't able to fix the issue. Track in #5968.
* When some (not all) inputs/output of a `ttir.add` op is the result of a implicit bcast op, the lower toLayout pass doesn't know how to handle the mismatching collapsed & uncollapsed tensor layouts. #5540 Could help.

### Checklist

### DIFF
--- a/include/ttmlir/Dialect/D2M/IR/D2MOpsEnums.td
+++ b/include/ttmlir/Dialect/D2M/IR/D2MOpsEnums.td
@@ -21,6 +21,38 @@ def D2M_ReduceDim : I32EnumAttr<"ReduceDim", "D2M ReduceDim", [
   let cppNamespace = "::mlir::tt::d2m";
 }
 
+def D2M_TileBcastTypeNone   : I32EnumAttrCase<"None",   0, "none">;
+def D2M_TileBcastTypeCol    : I32EnumAttrCase<"Col",    1, "col">;
+def D2M_TileBcastTypeRow    : I32EnumAttrCase<"Row",    2, "row">;
+def D2M_TileBcastTypeScalar : I32EnumAttrCase<"Scalar", 3, "scalar">;
+
+def D2M_TileBcastType : I32EnumAttr<"TileBcastType", "D2M TileBcastType", [
+  D2M_TileBcastTypeNone,
+  D2M_TileBcastTypeCol,
+  D2M_TileBcastTypeRow,
+  D2M_TileBcastTypeScalar
+]> {
+  let summary = "D2M tile-level broadcast types.";
+  let description = [{
+    The type of the input shape (not the direction) of a unary tile-level
+    broadcast operation.
+
+    For type Col, the input should be a single tile with a filled 0-column and
+    zeros elsewhere.
+
+    For type Row, the input should be a single tile with a filled 0-row and
+    zeros elsewhere.
+
+    For type Scalar, the input should be a single tile with a filled single
+    value at location [0,0], and zeros elsewhere.
+
+    The output will be a full tile filled with the broadcasted input values.
+  }];
+
+  let genSpecializedAttr = 0;
+  let cppNamespace = "::mlir::tt::d2m";
+}
+
 // D2M ThreadType mirrors D2M thread type and is used by D2M_ThreadAttr
 def D2M_ThreadTypeCompute : I32EnumAttrCase<"Compute", 0, "compute">;
 def D2M_ThreadTypeDatamovement : I32EnumAttrCase<"Datamovement", 1, "datamovement">;
--- a/lib/Conversion/D2MToTTKernel/D2MToTTKernelPass.cpp
+++ b/lib/Conversion/D2MToTTKernel/D2MToTTKernelPass.cpp
@@ -6,9 +6,7 @@
 
 #include "ttmlir/Dialect/D2M/Analysis/CBProducerConsumer.h"
 #include "ttmlir/Dialect/D2M/IR/D2M.h"
-#include "ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.h"
 #include "ttmlir/Dialect/TTCore/IR/TTCore.h"
-#include "ttmlir/Dialect/TTCore/IR/TTCoreOps.h"
 #include "ttmlir/Dialect/TTIR/IR/TTIROps.h"
 #include "ttmlir/Dialect/TTKernel/IR/TTKernel.h"
 #include "ttmlir/Dialect/TTKernel/IR/TTKernelOpsTypes.h"
--- a/lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
+++ b/lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
@@ -4,19 +4,11 @@
 
 #include "ttmlir/Conversion/TTIRToD2M/TTIRToD2M.h"
 
-#include "ttmlir/Dialect/TTCore/IR/TTCore.h"
-#include "ttmlir/Dialect/TTCore/IR/TTCoreOpsTypes.h"
-#include "ttmlir/Dialect/TTIR/IR/TTIROps.h"
-// D2M generic/region ops
 #include "ttmlir/Asserts.h"
 #include "ttmlir/Dialect/D2M/IR/D2M.h"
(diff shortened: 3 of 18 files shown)


include/ttmlir/Dialect/D2M/IR/D2MOpsEnums.td
```diff
@@ -21,6 +21,38 @@ def D2M_ReduceDim : I32EnumAttr<"ReduceDim", "D2M ReduceDim", [
   let cppNamespace = "::mlir::tt::d2m";
 }
 
+def D2M_TileBcastTypeNone   : I32EnumAttrCase<"None",   0, "none">;
```
nsmithtt, line 24: What does None mean? Should we even bother representing it in d2m?

lib/Conversion/D2MToTTKernel/D2MToTTKernelPass.cpp
```diff
@@ -159,6 +241,15 @@ struct ConvertD2MToTTKernel
...
     }
+
+    getOperation().walk([this](func::FuncOp func) {
+      auto threadType = func->getAttrOfType<ttkernel::ThreadTypeAttr>(
+          ttkernel::ThreadTypeAttr::name);
+      if (threadType &&
+          threadType.getValue() == ttkernel::ThreadType::Compute) {
+        reorderUnaryBcastOps(func);
+      }
+    });
```
nsmithtt: Maybe this should be a dedicated workaround pass?  I guess an argument could be made that the conversion isn't legal until this transformation is run.

lib/Conversion/D2MToTTKernel/D2MToTTKernelPass.cpp
```diff
@@ -56,6 +55,89 @@ struct ConvertD2MToTTKernel
...
+
+          // Find the one and only consumer.
+          for (auto *searchOp = bcastOp->getNextNode(); searchOp;
+               searchOp = searchOp->getNextNode()) {
+            if (searchOp->hasTrait<ttkernel::TTKernelFPUOpTrait>() ||
+                searchOp->hasTrait<ttkernel::TTKernelSFPUOpTrait>()) {
+              for (auto operand : searchOp->getOperands()) {
+                consumerOperands.push_back(operand);
+                if (operand == bcastDstIdx) {
+                  consumer = searchOp;
```
nsmithtt: Break?

lib/Conversion/D2MToTTKernel/D2MToTTKernelPass.cpp
```diff
@@ -56,6 +55,89 @@ struct ConvertD2MToTTKernel
     this->ttnnMode = rhs.ttnnMode;
   }
 
+  // TODO(wenbinlyuTT): remove this WA once unary_bcast DST issue is fixed.
```
nsmithtt: Do we have an issue for this?  Is this an llk fix?

lib/Dialect/D2M/Transforms/ElementwiseFusion.cpp
```diff
@@ -100,6 +100,17 @@ static bool isValidElementwiseFusionTarget(GenericOp gOp) {
     return false;
   }
 
+  // TODO(wenbinlyuTT): fusing ops w/ implicit bcast produces dangling
```
nsmithtt: Can you file an issue for this?

include/ttmlir/Dialect/D2M/IR/D2MOpsEnums.td
wenbinlyuTT: The none type isn't used, but it is a required template argument in some llk API, like
https://github.com/tenstorrent/tt-metal/blob/b8d340f6c92fec1a2a8d407bb356f0f0e77ff3f2/tt_metal/include/compute_kernel_api/tilize.h#L37
Is there any chance we will need to emit llk calls in the future? If not then I can remove it.

lib/Conversion/D2MToTTKernel/D2MToTTKernelPass.cpp
wenbinlyuTT: Suggestions from the LLK folks weren't helpful, and my attempted reproducer can't reproduce the bug even if it's using the exact same set of dumped kernels. Something in our runtime workflow might be contributing to the issue but I'm not sure. I'm still trying to figure out if this is a LLK issue or our issue...

lib/Dialect/D2M/Transforms/ElementwiseFusion.cpp
wenbinlyuTT: Did a little digging and narrowed down the bug, #5968

lib/Conversion/D2MToTTKernel/D2MToTTKernelPass.cpp
zoecarver: I agree that I'd love to not have this, it's crashing when I used it in my tests because it either sometimes can't find a consumer or finds too many. 

If dst reg pass is fixed, we shouldn't need it (Boyana's pass for example works fine with this op without `reorderUnaryBcastOps`)

lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
```diff
@@ -412,7 +522,7 @@ class D2MNamedElementwiseRewriter final
             [&](mlir::OpBuilder &bbBuilder, mlir::Location bbLoc,
                 mlir::ValueRange bbArgs) {
               createComputeRegion(bbBuilder, bbLoc, bbArgs, rewriter, loc,
-                                  numInputs, numOutputs);
+                                  numInputs, numOutputs, tileBcastTypes);
```
vmilosevic, line 541: :warning: **clang\-diagnostic\-c\+\+20\-extensions** :warning:
captured structured bindings are a C\+\+20 extension

vmilosevic (commented):

lib/Conversion/D2MToTTKernel/D2MToTTKernelPass.cpp
wenbinlyuTT: @nsmithtt I reproduced the error with my pure metal test program, let's see what the LLK folks say about my reproducer. Meanwhile I modified the insert DST pass so all the load-bcast pairs are in the front. It's a lot more natural and this WA is gone.

@zoecarver I took a new approach and you might want to try again.

lib/Dialect/D2M/Transforms/InsertDstRegisterAccess.cpp
```diff
@@ -401,34 +423,60 @@ struct D2MInsertDstRegisterAccessRewriter final
...
+  }
+
+  static void collectDstLoadAndBcast(GenericOp op, affine::AffineLoadOp loadOp,
+                                     d2m::TileBcastOp bcastOp,
+                                     CopyInfoMap &copyInfos, int dstSliceIdx,
+                                     Operation *outermostInnerComputeLoop) {
+    if (!outermostInnerComputeLoop) {
+      // If there is no outermostInnerComputeLoop, the common ancestor is the
+      // operation itself.
+      outermostInnerComputeLoop = loadOp; // Ugh, is this right?
```
nsmithtt: Perhaps we should validate your comment here and remove it?

wenbinlyuTT: Oops, this is a leftover comment to myself.

lib/Dialect/D2M/Transforms/InsertDstRegisterAccess.cpp
nsmithtt, line 1: FYI @brnorris03 , changes to this file

PR #5962, opened 2025-11-20
General `permute . reshape . permute -> reshape` pattern

azecevicTT:
https://github.com/tenstorrent/tt-mlir/pull/3915 and https://github.com/tenstorrent/tt-mlir/pull/5899 introduce specific `permute . reshape . permute -> reshape` patterns with hard-coded values of permutation.

This PR generalizes the approach such that:
```
This pattern fuses the sequence: PermuteOp -> ReshapeOp -> PermuteOp into a
single ReshapeOp when the following conditions are met:
Original shape: [A_1, A_2,.., A_k]
permute(p_1, p_2,..., p_k)    -> [A_1', A_2',.., A_k']
reshape([A_1', A_2',.., A_k']) -> [B_1, B_2,.., B_k]
permute(p_1', p_2',..., p_k') -> [B_1', B_2',.., B_k']

where:
- k is the rank of the input tensor;
- (p_1, p_2,..., p_k) and (p_1', p_2',..., p_k') are
  permutations of {0, 1, ..., k-1};
- B_i = (A_r', A_r+1',..., A_r+l') where 1 <= r <= k, l >= 0 and r + l <= k,
  for each 1 <= i <= k;
- flatten([B_1', B_2',.., B_k']) = [A_1, A_2,.., A_k].

The result of this sequence is identical to the following reshape:
reshape([A_1, A_2,.., A_k]) -> [B_1', B_2',.., B_k']
```

Special credits to @mvasiljevicTT for scrutinizing an algorithm during initial development, which revealed some edge cases that weren't previously covered.

### DIFF
--- a/lib/Dialect/TTIR/IR/TTIRTMFusionPatterns.cpp
+++ b/lib/Dialect/TTIR/IR/TTIRTMFusionPatterns.cpp
@@ -18,7 +18,7 @@ namespace mlir::tt::ttir {
 // Original shape: [A_1, A_2,.., A_k]
 // permute(p_1, p_2,..., p_k)    -> [A_1', A_2',.., A_k']
 // reshape([A_1', A_2',.., A_k']) -> [B_1, B_2,.., B_k]
-// permute(p_1', p_2',..., p_k') -> [B_1', 1, B_2',.., B_k']
+// permute(p_1', p_2',..., p_k') -> [B_1', B_2',.., B_k']
 //
 // where:
 // - k is the rank of the input tensor;
@@ -29,7 +29,7 @@ namespace mlir::tt::ttir {
 // - flatten([B_1', B_2',.., B_k']) = [A_1, A_2,.., A_k].
 //
 // The result of this sequence is identical to the following reshape:
-// reshape([A_1, A_2,.., A_k]) -> [B_1', 1, B_2',.., B_k']
+// reshape([A_1, A_2,.., A_k]) -> [B_1', B_2',.., B_k']
 //
 class PermuteReshapePermuteFusionPattern
     : public mlir::OpRewritePattern<PermuteOp> {


lib/Dialect/TTIR/IR/TTIRTMFusionPatterns.cpp
```diff
@@ -5,137 +5,145 @@
...
+               "input and output shapes must have the same rank; "
+               "inputShape={}, outputShape={}",
+               inputShape.size(), outputShape.size());
+
+    llvm::SmallVector<llvm::SmallVector<int64_t>> axesGroups;
+    const int64_t rank = inputShape.size();
+    int64_t inputIndex = 0;
+    for (int64_t outputIndex = 0; outputIndex < rank; ++outputIndex) {
+      if (inputShape[inputIndex] == outputShape[outputIndex]) {
+        assert(inputIndex < rank && "input index out of bounds");
```
mvasiljevicTT: Loop in last case can increase inputIndex up to the value of rank, this should be checked before indexing

lib/Dialect/TTIR/IR/TTIRTMFusionPatterns.cpp
```diff
@@ -5,137 +5,145 @@
...
 //
-// It is hard to express all pairs of permutations and reshapes that can
-// commute, so for now we only support the cases we need:
+// where:
+// - k is the rank of the input tensor;
+// - (p_1, p_2,..., p_k) and (p_1', p_2',..., p_k') are
+//   permutations of {0, 1, ..., k-1};
+// - B_i = (A_r', A_r+1',..., A_r+l') where 1 <= r <= k, l >= 0 and r + l <= k,
+//   for each 1 <= i <= k;
+// - flatten([B_1', B_2',.., B_k']) = [A_1, A_2,.., A_k].
```
mvasiljevicTT, line 29: From the comment seems like it is enough that these shapes match by value, but is the actual condition stronger than that?

lib/Dialect/TTIR/IR/TTIRTMFusionPatterns.cpp
```diff
@@ -5,137 +5,145 @@
...
                   mlir::PatternRewriter &rewriter) const override {
-    // Find matching case based on the second permutation.
-    ArrayRef<int64_t> secondPermutation = op.getPermutation();
-    const PermuteReshapePermutePatternSpec *matchedPattern =
-        findMatchingPattern(secondPermutation);
-    if (!matchedPattern) {
+
+    // Check if the final PermuteOp is preceded by a ReshapeOp.
+    auto reshapeOp = finalPermuteOp.getInput().getDefiningOp<ReshapeOp>();
+    if (!reshapeOp) {
```
mvasiljevicTT, line 45:
```suggestion
    if (!reshapeOp || !reshapeOp->hasOneUse()) {
```

lib/Dialect/TTIR/IR/TTIRTMFusionPatterns.cpp
```diff
@@ -5,137 +5,145 @@
...
       return failure();
     }
 
-    // Verify the pattern structure: Permute -> Reshape -> Permute.
-    ReshapeOp reshapeOp = op.getInput().getDefiningOp<ReshapeOp>();
-    if (!reshapeOp || !reshapeOp->hasOneUse()) {
+    // Check if the ReshapeOp is preceded by a PermuteOp.
+    auto originalPermuteOp = reshapeOp.getInput().getDefiningOp<PermuteOp>();
+    if (!originalPermuteOp) {
```
mvasiljevicTT, line 51:
```suggestion
    if (!originalPermuteOp || !originalPermuteOp->hasOneUse()) {
```

lib/Dialect/TTIR/IR/TTIRTMFusionPatterns.cpp
azecevic000: I was also thinking about this, some very rudimentary cost-benefit analysis (rudimentary in the sense that metrics is ops add vs ops removed):
1) if all three ops have multiple users then it clearly makes sense to *not* do fusion, since we will have all three ops executed for some path, so adding a fusion would only add an additional ReshapeOp;
2) if two ops have multiple users, it's pretty much even, because we are effectively replacing that one op with a single user with a new ReshapeOp; the assumption here is that all ops cost the same (in terms of execution time, memory, etc.);
3) if only one or none of them have multiple users, we are adding just one op and removing two or three; of course, this is still under the assumption that all ops have the same cost.

Only 1) seems like a clear case, I'll add that check, 2) might require some more sophisticated analysis if we come across such case in practice, but I would do the fusion for that case for now.

lib/Dialect/TTIR/IR/TTIRTMFusionPatterns.cpp
azecevic000: Not sure which shapes you are referring to, but shape has only cardinality of an axis attached to it, not an identifier. Something like `[A, B]: 2x2` has the same shape as `[B, A]: 2x2`.

lib/Dialect/TTIR/IR/TTIRTMFusionPatterns.cpp
azecevic000: Good catch, fixed and added a test.

lib/Dialect/TTIR/IR/TTIRTMFusionPatterns.cpp
```diff
@@ -5,137 +5,151 @@
...
 namespace mlir::tt::ttir {
 
-// This pattern fuses the sequence: Permute -> Reshape -> Permute into a single
-// reshape op.
+// This pattern fuses the sequence: PermuteOp -> ReshapeOp -> PermuteOp into a
+// single ReshapeOp when the following conditions are met:
+// Original shape: [A_1, A_2,.., A_k]
+// permute(p_1, p_2,..., p_k)    -> [A_1', A_2',.., A_k']
+// reshape([A_1', A_2',.., A_k']) -> [B_1, B_2,.., B_k]
+// permute(p_1', p_2',..., p_k') -> [B_1', 1, B_2',.., B_k']
```
sdjordjevicTT: Probably 1 by accident?

azecevic000: Yeah, good catch, thanks!

Issue #5924, opened 2025-11-18
[D2M] div unary tile doing multiplication


dlokeTT:
### Describe the bug

title

PR #5970, opened 2025-11-21
[D2M] Promote constant tensors to scalars

dlokeTT:
### Ticket
Part 2 of adding scalar support for binary ops. Depends on #5915

### Problem description
Need to convert tensors of a constant to a scalar when possible

### What's changed
Added a pass after `TTIRToD2M` that goes thru all the `d2m.full` ops and traces to see if they can be turned into constants in the compute ops. If so, removes the full op and all corresponding values such as tensor/tile from the generic op
Div scalar has an issue in tt-metal/tt-llk where it does multiplication instead of division. Issue filed https://github.com/tenstorrent/tt-mlir/issues/5924 for follow on

### Checklist

### DIFF
--- a/lib/Dialect/D2M/Transforms/ScalarizeConstTensors.cpp
+++ b/lib/Dialect/D2M/Transforms/ScalarizeConstTensors.cpp
@@ -22,74 +22,56 @@ namespace mlir::tt::d2m {
 
 namespace {
 
-// Struct to track where a constant value flows through the IR
 struct ConstantUseChain {
-  Operation *constOp;       // The d2m.full op
-  Attribute splatValue;     // The scalar value
-  GenericOp genericOp;      // The d2m.generic that uses it
-  unsigned genericInputIdx; // Which input of the generic
+  GenericOp genericOp;
+  unsigned genericInputIdx;
 };
 
-// Trace forward from a constant op through d2m.to_layout to d2m.generic ops
 static void
-traceConstantToGenericOps(Operation *constOp, Attribute splatValue,
+traceConstantToGenericOps(Operation *constOp,
                           SmallVectorImpl<ConstantUseChain> &chains) {
-
-  // Walk through all uses of the constant
   for (Operation *user : constOp->getUsers()) {
     Value currentValue = constOp->getResult(0);
     Operation *currentOp = user;
 
-    // Trace through d2m.to_layout ops
     while (auto toLayoutOp = dyn_cast<ToLayoutOp>(currentOp)) {
       currentValue = toLayoutOp.getResult(0);
       if (currentValue.getUsers().empty()) {
         break;
       }
-      // Move to next user
       // NOTE: This only follows the first user. If a to_layout result has
       // multiple users, only one path will be traced. This is acceptable for
       // the current use case but may miss optimization opportunities.
       currentOp = *currentValue.getUsers().begin();
     }
 
-    // Check if we reached a d2m.generic op
     if (auto genericOp = dyn_cast<GenericOp>(currentOp)) {
-      // Find which input operand this constant corresponds to
       for (auto [idx, input] : llvm::enumerate(genericOp.getInputs())) {
         if (input == currentValue) {
-          chains.push_back(
-              {constOp, splatValue, genericOp, static_cast<unsigned>(idx)});
+          chains.push_back({genericOp, static_cast<unsigned>(idx)});
         }
       }
     }
   }
 }
 
-// Helper to trace from a d2m.generic input to linalg.generic block arguments
 static void
 findLinalgBlockArgsForGenericInput(GenericOp genericOp,
                                    unsigned genericInputIdx,
                                    SmallVectorImpl<BlockArgument> &linalgArgs) {
-
-  // Walk through the generic region to find linalg.generic ops
   for (Region &region : genericOp.getRegions()) {
     region.walk([&](linalg::GenericOp linalgOp) {
       Block *linalgBlock = linalgOp.getBody();
 
-      // Check each linalg input to see if it traces back to our generic input
       for (auto [linalgArgIdx, linalgInput] :
            llvm::enumerate(linalgOp.getInputs())) {
-
         Value tracedValue = linalgInput;
         Operation *defOp = tracedValue.getDefiningOp();
 
-        // Trace back through d2m.wait operations
         if (auto waitOp = dyn_cast_or_null<WaitOp>(defOp)) {
           tracedValue = waitOp.getCb();
         }
 
-        // Check if this traces back to a generic block argument
         if (auto blockArg = dyn_cast<BlockArgument>(tracedValue)) {
           if (blockArg.getOwner()->getParentOp() == genericOp.getOperation() &&
               blockArg.getArgNumber() == genericInputIdx) {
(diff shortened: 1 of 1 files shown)


include/ttmlir/Dialect/D2M/IR/D2MOps.td
```diff
@@ -303,7 +303,7 @@ def D2M_EmptyOp : D2M_Op<"empty", [
 }
 
 def D2M_FullOp : D2M_Op<"full", [
-    MemoryEffects<[MemAlloc]>,
+    Pure,
```
wenbinlyuTT, line 306: What if a fullOp is consumed by something other than a binary op that supports scalar input and so does require an allocation? Will this break that use case or am I misunderstanding something here.

lib/Dialect/D2M/Transforms/ScalarizeConstTensors.cpp
```diff
@@ -0,0 +1,472 @@
...
+#include "mlir/IR/PatternMatch.h"
+#include "mlir/Transforms/GreedyPatternRewriteDriver.h"
+
+namespace mlir::tt::d2m {
+#define GEN_PASS_DEF_D2MSCALARIZECONSTTENSORS
+#include "ttmlir/Dialect/D2M/Transforms/Passes.h.inc"
+
+namespace {
+
+// Struct to track where a constant value flows through the IR
```
wenbinlyuTT: Comments are full sentences. https://docs.tenstorrent.com/tt-mlir/coding-guidelines.html#comments

lib/Dialect/D2M/Transforms/ScalarizeConstTensors.cpp
```diff
@@ -0,0 +1,472 @@
...
+struct ConstantUseChain {
+  Operation *constOp;       // The d2m.full op
+  Attribute splatValue;     // The scalar value
```
wenbinlyuTT: The first two fields of the struct doesn't seem to be used anywhere?
Also, just curious, why call it "splat"?

lib/Dialect/D2M/Transforms/ScalarizeConstTensors.cpp
```diff
@@ -0,0 +1,472 @@
...
+          }
+        }
+      }
+    });
+  }
+}
+
+// Helper to check if an operation uses a value as a scalarizable RHS operand
+static bool isScalarRhsUse(Operation *user, Value arg) {
+  if (auto computeOp = dyn_cast<OperandLoadStoreRegisterOpInterface>(user)) {
```
wenbinlyuTT, line 87: Just a brain dump: I've always thought checking `D2M_GenericRegionComputeOpTrait` is more natural and self-explanatory. But the `supportsTileOrScalarRhs()` method is defined on this interface, alas.

lib/Dialect/D2M/Transforms/ScalarizeConstTensors.cpp
```diff
@@ -0,0 +1,472 @@
...
+    if (arg.use_empty()) {
+      unusedIndices.push_back(i);
+    }
+  }
+  return unusedIndices;
+}
+
+// Helper to build new inputs list excluding unused ones
+template <typename RangeT>
+static SmallVector<Value> buildNewInputsList(RangeT &&inputs,
```
wenbinlyuTT: Nit: the function name can be more descriptive, so that you don't need the comment above.

lib/Dialect/D2M/Transforms/ScalarizeConstTensors.cpp
```diff
@@ -0,0 +1,472 @@
...
+  for (auto [idx, input] : llvm::enumerate(inputs)) {
+    if (!llvm::is_contained(unusedIndices, idx)) {
+      newInputs.push_back(input);
+    }
+  }
+  return newInputs;
+}
+
+// Helper to build mapping from old block args to new ones, skipping unused
+static IRMapping buildBlockArgMapping(Block *oldBlock, Block *newBlock,
```
wenbinlyuTT: Ditto, more descriptive function name that aligns with the helper function above.

lib/Dialect/D2M/Transforms/ScalarizeConstTensors.cpp
```diff
@@ -0,0 +1,472 @@
...
+    }
+    mapping.map(oldBlock->getArgument(oldArgIdx),
+                newBlock->getArgument(newArgIdx++));
+  }
+  return mapping;
+}
+
+// Helper to rebuild linalg.generic with fewer inputs
+static linalg::GenericOp
+rebuildLinalgGenericWithFewerInputs(linalg::GenericOp linalgOp,
```
wenbinlyuTT: Ditto, more descriptive function name that aligns with the helper functions above. "New" / "Fewer" don't really convey a lot of info.

lib/Dialect/D2M/Transforms/ScalarizeConstTensors.cpp
```diff
@@ -0,0 +1,472 @@
...
+  for (Operation &op : linalgBlock->getOperations()) {
+    rewriter.clone(op, mapping);
+  }
+
+  return newLinalgOp;
+}
+
+// Helper to rebuild d2m.generic with fewer inputs
+static GenericOp
+rebuildD2MGenericWithFewerInputs(GenericOp genericOp,
```
wenbinlyuTT: Ditto on function naming.

lib/Dialect/D2M/Transforms/ScalarizeConstTensors.cpp
```diff
@@ -0,0 +1,472 @@
...
+
+// Pattern to scalarize d2m.full constant tensor operands
+class ScalarizeFullOpPattern : public OpRewritePattern<FullOp> {
+public:
+  using OpRewritePattern<FullOp>::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(FullOp fullOp,
+                                PatternRewriter &rewriter) const override {
+    Attribute splatValue = fullOp.getFillValue();
+    if (!splatValue) {
```
wenbinlyuTT: Will this ever gets hit?

wenbinlyuTT (commented):
Did a quick pass, overall looks good. Mostly only nit picks and questions for myself.

Is this feature on tight deadline? If not it probably worth it to add support for `rsub/rdiv/rpow` so that we are not limited to RHS fullOps. It's your choice if you would rather do it in a follow-up PR.

include/ttmlir/Dialect/D2M/IR/D2MOps.td
dlokeTT: Changing this only tells the canonicalizer that the fullOp can be canonicalized out if it has no consumers. So if there is a consumer still that requires allocation, it won't get canonicalized out.

lib/Dialect/D2M/Transforms/ScalarizeConstTensors.cpp
dlokeTT: Not sure. Ran into it with some other work but it might be handled in an earlier pass. Let me try and remove it

lib/Dialect/D2M/Transforms/ScalarizeConstTensors.cpp
dlokeTT: Oh yea, this is an artifact from old approach. Will remove
Not sure actually, that's just what I've seen

lib/Dialect/D2M/Transforms/ScalarizeConstTensors.cpp
dlokeTT: It doesn't. Getting rid of it

lib/Dialect/D2M/Transforms/ScalarizeConstTensors.cpp
dlokeTT: Hmm yea I see your point. Everything so far was defined on the interface so I went with that

dlokeTT:
> LGTM! Please open an issue to track the LHS fullOp support we talked about.

https://github.com/tenstorrent/tt-mlir/issues/6126

Issue #5785, opened 2025-11-11
error: 'ttnn.avg_pool2d' op Kernel height 7 is greater than input height 1. This ttnn.avg_pool2d configuration is invalid.


mmilosevicTT:
### Describe the bug

Several versions of `hrnet` TT-XLA PyTorch model are failing in training with the `ttnn.avg_pool2d` after the fix for batch norm training in https://github.com/tenstorrent/tt-mlir/pull/5287. 

Sample outputs:
- [hrnet1_output.txt](https://github.com/user-attachments/files/23475403/hrnet1_output.txt)
- [hrnet2_output.txt](https://github.com/user-attachments/files/23475642/hrnet2_output.txt)

From both files we extract the same stablehlo op that is problematic:
```
%9 = "stablehlo.reduce_window"(%8, %cst_12) <{window_dimensions = array<i64: 1, 1, 7, 7>}> ({
    ^bb0(%arg40: tensor<bf16> loc("reduce-window.42"), %arg41: tensor<bf16> loc("reduce-window.42")):
      %51 = stablehlo.add %arg40, %arg41 : tensor<bf16> loc(#loc52)
      stablehlo.return %51 : tensor<bf16> loc(#loc)
    }) : (tensor<1x2048x13x13xbf16>, tensor<bf16>) -> tensor<1x2048x7x7xbf16> loc(#loc51)
```

Outputs of some versions of these models are provided above.

### Steps to reproduce the issue

1. Go to TT-XLA
2. Replace NOT_SUPPORTED_SKIP tag with EXPECTED_PASSING in the `tests/runner/test_config/torch/test_config_training_single_device.yaml`
3. Run `pytest tests/runner/test_models.py::test_all_models_torch[<TEST_NAME>]`, or any other of the mentioned models.
4. Full list of models failing with this error is present in `tests/runner/test_config/torch/test_config_training_single_device.yaml`

mmilosevicTT:
@sdjordjevicTT @djstefanovicTT @vladimirjovanovicTT Pinging for visibility. Will extract the graph later today.

djstefanovicTT:
```
%9 = "stablehlo.reduce_window"(%8, %cst_12) <{window_dimensions = array<i64: 1, 1, 7, 7>}> ({
    ^bb0(%arg40: tensor<bf16> loc("reduce-window.42"), %arg41: tensor<bf16> loc("reduce-window.42")):
[...]

PR #5331, opened 2025-10-13
Adding a reshape to `ttir.pooling` decomposition

ajakovljevicTT:
Currently, the `ttir.pooling` op only accepts 4D tensors as inputs. This PR adds a reshape so that the non-4D tensors can also go into the `ttir.pooling` op.

For example, ttir MLIR code looking like this:
```
  func.func @avg_pool_2d_example(%arg0: tensor<28x28xf32>) -> tensor<14x14xf32> {
    %0 = ttir.empty() : tensor<14x14xf32>

    %1 = "ttir.pooling"(%arg0, %0) <{
      base_dilations = array<i64: 1, 1>,
      operandSegmentSizes = array<i32: 1, 1>,
      padding = array<i64: 0, 0, 0, 0>,
      pooling_method = #ttir<pooling_method Average>,
      window_dilations = array<i64: 1, 1>,
      window_dimensions = array<i64: 2, 2>,
      window_strides = array<i64: 2, 2>
    }> : (tensor<28x28xf32>, tensor<14x14xf32>) -> tensor<14x14xf32>

    return %1 : tensor<14x14xf32>
  }
```
Will get translated to this:
```
      func.func @avg_pool_2d_example(%arg0: tensor<28x28xf32, #ttnn_layout>) -> tensor<14x14xf32, #ttnn_layout> {
        %0 = "ttnn.reshape"(%arg0) <{shape = [1 : i32, 1 : i32, 28 : i32, 28 : i32]}> : (tensor<28x28xf32, #ttnn_layout>) -> tensor<1x1x28x28xf32, #ttnn_layout1>
        "ttnn.deallocate"(%arg0) <{force = false}> : (tensor<28x28xf32, #ttnn_layout>) -> ()
        %1 = "ttnn.permute"(%0) <{permutation = array<i64: 0, 2, 3, 1>}> : (tensor<1x1x28x28xf32, #ttnn_layout1>) -> tensor<1x28x28x1xf32, #ttnn_layout2>
        "ttnn.deallocate"(%0) <{force = false}> : (tensor<1x1x28x28xf32, #ttnn_layout1>) -> ()
        %2 = "ttnn.reshape"(%1) <{shape = [1 : i32, 1 : i32, 784 : i32, 1 : i32]}> : (tensor<1x28x28x1xf32, #ttnn_layout2>) -> tensor<1x1x784x1xf32, #ttnn_layout3>
[...]

PR #5975, opened 2025-11-21
Pooling op reshape input

djstefanovicTT:
### Ticket
Closes https://github.com/tenstorrent/tt-mlir/issues/5785

### Problem description
Pooling ops require 4d tensors as input, and StableHLO can have 2d input.

### What's changed
Added reshapes to 4d if needed.

### Checklist


ajakovljevicTT:
Hey @djstefanovicTT , just checking if this will also solve the issue that this PR (https://github.com/tenstorrent/tt-mlir/pull/5331) is trying to solve?

djstefanovicTT:
> Hey @djstefanovicTT , just checking if this will also solve the issue that this PR (#5331) is trying to solve?

It is essentialy doing the same thing, just on a different level.

ajakovljevicTT:
> > Hey @djstefanovicTT , just checking if this will also solve the issue that this PR (#5331) is trying to solve?
> 
> It is essentialy doing the same thing, just on a different level.

Thanks, I will close then my PR as a duplicate.

sdjordjevicTT:
@odjuricicTT @mtopalovicTT you were discussing something similar here:
https://github.com/tenstorrent/tt-mlir/pull/5331

Is this not the right approach to fix this?

@djstefanovicTT, as stablehlo implementation doesn't require a 4D tensor, can we remove that limitation from TTIR and move it to the TTNN workaround pass?

lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
```diff
@@ -1904,12 +1904,72 @@ class StableHLOToTTIRReduceWindowOpConversionPattern
...
+    DenseI64ArrayAttr adjustedWindowDimensions = windowDimensions;
+    DenseI64ArrayAttr adjustedWindowStrides = windowStrides;
+    DenseI64ArrayAttr adjustedBaseDilations = baseDilations;
+    DenseI64ArrayAttr adjustedWindowDilations = window_dilations;
+    DenseI64ArrayAttr adjustedPadding = padding;
+
+    if (needsReshape) {
+      SmallVector<int64_t> winDims4D = {1, 1};
+      llvm::append_range(winDims4D, windowDimensions.asArrayRef());
+      adjustedWindowDimensions = rewriter.getDenseI64ArrayAttr(winDims4D);
```
azecevic000: Nit: this is a repeating pattern, you can extract it in some helper (private class) function, something like `DenseI64ArrayAttr prependOnesTo2D(DenseI64ArrayAttr)` (some better name is plus), and also add an assert that array indeed has two elements as a sanity check.

azecevicTT (approved):
Can you add a few builder tests? Otherwise, looks good.

tapspatel (approved):

Issue #5884, opened 2025-11-17
SplitQueryKeyValueAndSplitHeadsUpdatePattern breaking Segformer on tt-xla


mvasiljevicTT:
[This PR](https://github.com/tenstorrent/tt-mlir/pull/5740) introduced `SplitQueryKeyValueAndSplitHeadsUpdatePattern` that causes segformer model to break on [nightly run](https://github.com/tenstorrent/tt-forge/actions/runs/19418396278/job/55551133266) with error:
```
<unknown>:0: error: loc("reshape.4367"): operand #0 does not dominate this use
<unknown>:0: note: loc("reshape.4367"): see current operation: %159 = "ttnn.reshape"(%1109#1) <{shape = [8 : i32, 32 : i32, 256 : i32]}> : (tensor<1x8x32x256xbf16, #ttnn.ttnn_layout<(d0, d1, d2, d3) -> (d0 * 256 + d1 * 32 + d2, d3), <1x1>, memref<8x8x!ttcore.tile<32x32, bf16>, #ttnn.buffer_type<dram>>, <interleaved>>>) -> tensor<8x32x256xbf16, #ttnn.ttnn_layout<(d0, d1, d2) -> (d0 * 32 + d1, d2), <1x1>, memref<8x8x!ttcore.tile<32x32, bf16>, #ttnn.buffer_type<dram>>, <interleaved>>>
```
with line 1109 being:
```
[...]

PR #5977, opened 2025-11-21
Fix Scheduler for ops with multiple outputs

mvasiljevicTT:
### Ticket
[Link to Github Issue](https://github.com/tenstorrent/tt-mlir/issues/5884)

### Problem description
When running models that include ops with multiple outputs, such as split_query_key_value_and_split_heads, and the memory-layout-analysis-enabled flag is enabled, ops may be incorrectly sorted, with users of a multi-output op appearing before the op itself.

This occurs because the scheduler assumes ops have only one output, so it considers only the users of the first result.
### What's changed
Fixed dependency initialization in the scheduler so that all results are processed.

### Checklist

### DIFF
--- a/lib/Scheduler/Scheduler.cpp
+++ b/lib/Scheduler/Scheduler.cpp
@@ -114,7 +114,7 @@ llvm::SmallVector<mlir::Operation *> Scheduler::getSchedulableOps() {
   // std::stable_sort internally uses std::get_temporary_buffer which is
   // deprecated in C++17. The deprecation is in the standard library
   // implementation, not our usage.
-  // NOLINTNEXTLINE(clang-diagnostic-deprecated-declarations)
+  // NOLINTNEXTLINE
   std::stable_sort(schedulableOps.begin(), schedulableOps.end(),
                    [&](mlir::Operation *a, mlir::Operation *b) {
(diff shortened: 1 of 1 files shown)


test/unittests/TestScheduler/TestScheduler.cpp
```diff
@@ -225,3 +225,62 @@ TEST_F(SchedulerBase, VerifyFork) {
...
+
+  mlir::tt::scheduler::Scheduler scheduler(&func);
+  llvm::SmallVector<mlir::Operation *> scheduleableOps =
+      scheduler.getSchedulableOps();
+  ASSERT_EQ(scheduleableOps.size(), 1);
+  ASSERT_EQ(scheduleableOps[0], splitOp.getOperation());
+
+  scheduler.scheduleOp(scheduleableOps[0]);
+  scheduleableOps = scheduler.getSchedulableOps();
+  ASSERT_EQ(scheduleableOps.size(), 3);
```
rpavlovicTT, line 285: Should you check order of QKV consumer ops too? Just to make sure your partitioning works as expected.

mvasiljevicTT: Added checks now

Issue #5890, opened 2025-11-17
Invalid data size in runtime


sgligorijevicTT:
Bit of housekeeping work, this is a regression induced by https://github.com/tenstorrent/tt-mlir/pull/4768 back in september but seems like no corresponding MLIR issue was opened. XLA issue for ref: https://github.com/tenstorrent/tt-xla/issues/1306 

Running the model with logging on shows:
```
[32m            RuntimeTTNN[0m | [1m[38;5;240m   DEBUG[0m | Executing operation: %0 = "ttnn.get_device"() <{mesh_offset = #ttnn<mesh_offset 0x0>, mesh_shape = #ttnn<mesh_shape 1x1>}> : () -> !ttnn.device loc("gather.8568_constant"("gather.8568"))
[32m            RuntimeTTNN[0m | [1m[38;5;240m   DEBUG[0m | Executing operation: %1 = "ttnn.constant"(%0) <{dtype = #ttcore.supportedDataTypes<bf16>, layout = #ttnn.layout<tile>, memory_config = #ttnn.memory_config<<dram>, <interleaved>>, value = dense<[[0], [1]]> : tensor<2x1xsi32>}> : (!ttnn.device) -> tensor<2x1xbf16, #ttnn.ttnn_layout<(d0, d1) -> (d0, d1), <1x1>, memref<1x1x!ttcore.tile<32x32, bf16>, #ttnn.buffer_type<dram>>, <interleaved>>> loc(unknown)
[32m                 Always[0m | [1m[31m   FATAL[0m | Invalid data size
[1mLOG_ASSERT @ /localdev/sgligorijevic/tt-xla/third_party/tt-mlir/src/tt-mlir/runtime/lib/ttnn/operations/utils/utils.cpp:363: numElements * elementSize == data->size()
backtrace:
 --- /localdev/sgligorijevic/tt-xla/third_party/tt-mlir/install/lib/libTTMLIRRuntime.so(+0x58a5a0) [0x7fa05c7395a0]
 --- tt::runtime::ttnn::operations::utils::toTTNNTensor(flatbuffers::Vector<unsigned char, unsigned int> const*, tt::tt_metal::Shape const&, tt::tt_metal::DataType const&, tt::tt_metal::distributed::MeshDevice*, tt::tt_metal::Layout const&, tt::tt_metal::MemoryConfig const&)
[...]

Issue #5021, opened 2025-09-18
Remove Const 4d workaround for EmitPy


sdjordjevicTT:
Currently, the ttnn Python tensor creation function `ttnn.Tensor` only accepts a fixed 4-dim shape argument. Once this limitation is removed from ttnn code, we should remove the workaround.

TTNN issue:
https://github.com/tenstorrent/tt-metal/issues/28739

jojo-771771:
I can work on this. Please assign!

jojo-771771:
I can work on this. Please assign!

jojo-771771:
I can work on this. Please assign!

PR #5980, opened 2025-11-21
Removing emitpy workarounds and fixing flatubffer constants

sdjordjevicTT:
### Ticket
Closes https://github.com/tenstorrent/tt-mlir/issues/5890
Closes https://github.com/tenstorrent/tt-mlir/issues/5021

### Problem description
The EmitPy pipeline had 2 workarounds for the constant op that were fixed by the metal folks recently. Also, we were tightly coupling the constant op value element type with the data type of the constant op output, like in this issue reported:
https://github.com/tenstorrent/tt-mlir/issues/5890

### What's changed
Removed 2 not-needed workarounds and updated the flatbuffer lowering for constant op to decouple the value vector data type from the op output data type. Additionally, fixed EmitPy lowering for max pool op.

### Checklist

### DIFF
--- a/runtime/lib/ttnn/operations/utils/utils.cpp
+++ b/runtime/lib/ttnn/operations/utils/utils.cpp
@@ -353,57 +353,57 @@ createConv2dSliceConfig(const ::tt::target::ttnn::Conv2dSliceConfig *config) {
 }
 
 template <typename T>
-static ::ttnn::Tensor
-toTTNNTensorImpl(const ::flatbuffers::Vector<uint8_t> *value,
-                 const ::ttnn::Shape &shape, const ::ttnn::DataType &dataType,
-                 ::ttnn::MeshDevice *device, const ::ttnn::Layout &layout,
-                 const ::ttnn::MemoryConfig &memoryConfig) {
+static ::ttnn::Tensor toTTNNTensorImpl(
+    const ::flatbuffers::Vector<uint8_t> *input, const ::ttnn::Shape &shape,
+    const ::ttnn::DataType &outputDataType, ::ttnn::MeshDevice *device,
+    const ::ttnn::Layout &layout, const ::ttnn::MemoryConfig &memoryConfig) {
   std::uint64_t numElements = shape.volume();
   size_t elementSize = sizeof(T);
-  LOG_ASSERT(numElements * elementSize == value->size(), "Invalid data size");
+  LOG_ASSERT(numElements * elementSize == input->size(), "Invalid data size");
   std::vector<T> dataVec(numElements);
   for (size_t i = 0; i < numElements; i++) {
     if constexpr (std::is_same_v<T, bfloat16>) {
       dataVec[i] = bfloat16(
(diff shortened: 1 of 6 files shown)


lib/Conversion/TTNNToEmitPy/TTNNToEmitPy.cpp
```diff
@@ -656,12 +656,12 @@ class MaxPool2dOpConversionPattern
...
         emitter.template emit<std::vector<uint32_t>>(
             maxPool2dOp.getDilationAttr()),
+        emitter.emit(maxPool2dOp.getCeilMode()),
         emitter.emit(emitter.getMemoryConfig(maxPool2dOp.getResult()),
                      "memory_config"),
         emitter.emit(maxPool2dOp.getAppliedShardScheme(),
                      "applied_shard_scheme"),
-        emitter.emit(maxPool2dOp.getCeilMode(), "ceil_mode"),
-        emitter.emit(maxPool2dOp.getInPlaceHalo(), "in_place_halo"),
+        emitter.emit(!maxPool2dOp.getInPlaceHalo(), "reallocate_halo_output"),
```
svuckovicTT, line 664: Would you mind updating the td for maxpool? 😀

sdjordjevicTT: I was thinking of doing that in the separate review :) Needed this for local testing.

runtime/lib/ttnn/operations/utils/utils.cpp
```diff
@@ -354,54 +354,55 @@ ::ttnn::DeviceComputeKernelConfig createDeviceComputeKernelConfig(
...
 
 ::ttnn::Tensor toTTNNTensor(
-    const ::flatbuffers::Vector<uint8_t> *data, const ::ttnn::Shape &shape,
+    const ::flatbuffers::Vector<uint8_t> *value,
+    const ::ttnn::DataType &valueDataType, const ::ttnn::Shape &shape,
     const ::ttnn::DataType &dataType, ::ttnn::MeshDevice *device = nullptr,
     const ::ttnn::Layout &layout = ::ttnn::Layout::ROW_MAJOR,
     const ::ttnn::MemoryConfig &memoryConfig = ::ttnn::DRAM_MEMORY_CONFIG) {
-  switch (dataType) {
+  switch (valueDataType) {
```
jnie-TT: From what I understand you're trying to implement an implicit typecast for constant op, if that's the case could we rename `valueDataType` and `dataType` to `inputDataType` and `outputDataType`, this is consistent with the eltwise ops that support implicit typecast and also clearer in my opinion. To mirror this it might also be worth renaming `value` to `input` so that it's clear that `inputDataType` describes the data type of the input buffer, while `outputDataType` will be the data type of the returned ttnn tensor.

sdjordjevicTT: Totally agree! Will make these changes in the next push!

Issue #5918, opened 2025-11-18
ttnn.jit wheel fails in Nightly CI


sgholamiTT:
`ttnn.jit` wheel building fails in nightly CI.
See [this](https://github.com/tenstorrent/tt-mlir/actions/runs/19453735938/job/55664377395).

vtangTT:
https://github.com/tenstorrent/tt-mlir/pull/5983 fixes

Issue #5232, opened 2025-10-06
Bad requirements for ttnn-jit wheel


nsumrakTT:
### Describe the bug

When ttnn-jit wheel is installed without developer docker there are missing requirements.
Wrong version of torch is installed (non-cpu).
Missing numpy.

nsumrakTT:
Related PRs

Original where wheel is introduced:
https://github.com/tenstorrent/tt-mlir/pull/5114

PR that moves wheel test to compute machine:
https://github.com/tenstorrent/tt-mlir/pull/5168

sgholamiTT:
I can see that in your [PR](https://github.com/tenstorrent/tt-mlir/pull/5168), you've added two lines when `Install ttnn-jit whl` step is being done.
```
pip install --index-url https://download.pytorch.org/whl/cpu torch==2.7.0
pip install numpy
```
The goal is to be able to remove that and the CI should pass. Can you confirm ?

----
I will wait for your work to be merged and then start working on this.

nsumrakTT:
@sgholamiTT I've added these lines to try to make it work. It installs wrong torch version with nvidia.
But it also requires some apt-get, custom openmpi and more.
In this PR I have to revert to our docker as PR introduces big changes to CI. If that doesn't work, I'll have to disable test.
IMHO it should install and run with known requirements from user perspective.

vtangTT:
i believe https://github.com/tenstorrent/tt-mlir/pull/5983 has fixed this. Can we close @nsumrakTT

nsumrakTT:
Yes, but I have to say that it is solved by removing release wheel, and keeping only dev version.

PR #5983, opened 2025-11-21
[ttnn.jit] new wheel

vtangTT:
### Ticket
https://github.com/tenstorrent/tt-mlir/issues/5232
https://github.com/tenstorrent/tt-mlir/issues/5918

### Problem description

### What's changed
- jit wheel now only packages necessary mlir + jit libs, then patches the rpath to look for the dependent metal libs
    - mlir + jit libs include:
        - `libJITCPP.so` - JIT src lib (this is just the JitCache)
        - `_ttnn_jit.cpython-311-x86_64-linux-gnu.so` - JIT bindings
        - `_ttmlir.cpython-311-x86_64-linux-gnu.so` - MLIR bindings
        - `libTTMLIRRuntime.so` - MLIR runtime bindings
            - the linked metal libs to this lib are not found until we patch the rpath and it's installed in an expected location
    - rpath patches include:
        - `"$ORIGIN/../../ttnn/build/lib",` - to point to `site-packages/ttnn/build/lib` which is where ttnn libs will be installed through the wheel
        - `"$ORIGIN/../../../../../../build/lib"` - to point to `${TT_METAL_HOME}/build/lib`, which assumes you are running the metal dev venv (created using `create_venv.sh`) 
- `cmake --build build` will now default to building and installing the JIT wheel into venv instead of copying files into `build/python_packages`
    - also added new target: `cmake --build build -- ttnn-jit` will do the same
- CI will run all tests with the JIT wheel (just like how ttrt does it)
    - removed ttnn-jit from `call-build-wheels.yml`
    - moved jit tests to new test script: `ttnn_jit.sh`
    - removed `ttnn-jit.sh` build script since it's no longer needed

**Important Note**
- tt-metal runs in python3.10 env while tt-mlir runs in python3.11 env :(
- metal devs will have to rebuild everything (the venv with llvm as well) to build a compatible python3.10 
- we are testing the 3.11 wheel in our CI, leaving potential gaps for 3.10 specific problems
- solution is to probably push forward for a manylinux wheel and pypi

### Checklist
- [x] passing on-push: https://github.com/tenstorrent/tt-mlir/actions/runs/19579113779
- [x] passing nightly: https://github.com/tenstorrent/tt-mlir/actions/runs/19686929423

### DIFF
--- a/tools/ttnn-jit/__init__.py
+++ b/tools/ttnn-jit/__init__.py
@@ -2,9 +2,11 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-import ttnn
-from importlib import import_module
 import os
+import re
+import subprocess
+
+import ttnn
 
 # _build_metadata.py is generated only by wheel build.
 try:
@@ -12,15 +14,25 @@ try:
 except ImportError:
     METAL_GIT_SHA = None
 
-if METAL_GIT_SHA is not None:
-    # Runtime check against {TT_METAL_HOME} git SHA to ensure compatibility.
-    metal_home = os.environ.get("TT_METAL_HOME", None)
-    if metal_home is None:
+
+def _get_ttnn_pip_sha():
+    """Extract SHA from ttnn pip package version string."""
+    try:
+        output = subprocess.check_output(["pip", "show", "ttnn"], text=True)
+        if match := re.search(r"Version:.*\+g([a-f0-9]+)", output):
+            return match.group(1)[:8]
+    except (subprocess.CalledProcessError, AttributeError):
+        pass
+    return None
+
+
+def _get_metal_home_sha():
+    """Extract SHA from TT_METAL_HOME git repository."""
+    metal_home = os.environ.get("TT_METAL_HOME")
+    if not metal_home:
         raise RuntimeError("TT_METAL_HOME is not set")
     try:
-        import subprocess
-
-        cwd_metal_sha = subprocess.check_output(
+        return subprocess.check_output(
             ["git", "rev-parse", "--short=8", "HEAD"],
             cwd=metal_home,
             text=True,
@@ -28,10 +40,29 @@ if METAL_GIT_SHA is not None:
     except Exception:
         raise RuntimeError("Could not get TT_METAL_HOME git SHA")
 
-    if cwd_metal_sha != METAL_GIT_SHA[:8]:
+
+def _verify_metal_compatibility():
+    """Verify tt-metal compatibility via pip package or TT_METAL_HOME."""
+    expected_sha = METAL_GIT_SHA[:8]
+
+    # Try ttnn pip package first (wheel-based installation)
+    if actual_sha := _get_ttnn_pip_sha():
+        if actual_sha != expected_sha:
+            raise RuntimeError(
+                f"Incompatible ttnn version. Expected SHA {expected_sha}, got {actual_sha}"
+            )
+        return
+
+    # Fall back to TT_METAL_HOME check (local dev environment)
+    actual_sha = _get_metal_home_sha()
+    if actual_sha != expected_sha:
         raise RuntimeError(
-            f"Incompatible tt-metal version detected. Expected {METAL_GIT_SHA[:8]}, got {cwd_metal_sha}. Please use a compatible tt-metal version."
+            f"Incompatible tt-metal version. Expected SHA {expected_sha}, got {actual_sha}"
         )
+
+
+if METAL_GIT_SHA is not None:
+    _verify_metal_compatibility()
 
 
 from ttnn_jit.api import (


tools/ttnn-jit/setup.py
```diff
@@ -54,196 +25,276 @@ def extract_tt_metal_version() -> str | None:
...
+        ]
+        if self.dev_build:
+            rpath_patches.append(f"{self.tt_metal_home}/build/lib")
+        for src_dir, lib_name in libraries_to_copy:
+            src_file = src_dir / lib_name
+            if src_file.exists():
+                shutil.copy2(src_file, lib_dest_dir / lib_name)
+                print(f"  Copied: {lib_name}")
+                print(f"  Patching rpath: {rpath_patches}")
+                subprocess.check_call(
```
nsmithtt: Do we still need to use patchelf?  Seems like these libs were build under our purview w/ RPATH ORIGIN set via cmake.

tools/ttnn-jit/setup.py
```diff
@@ -54,196 +25,276 @@ def extract_tt_metal_version() -> str | None:
...
+                env=build_env,
+            )
+        print("=" * 80)
+
+        # Copy MLIR .so libraries into ttnn_jit/lib/
+        arch = CMakeBuild.get_arch()
+        self._copy_mlir_so(build_dir, arch)
+        self._copy_ttmlir_bindings(build_dir)
+        if not self.dev_build:
+            self._write_build_metadata()
```
nsmithtt: Instead of having to do all of these manual copies ideally we use a cmake install component to do this for us. See `python/setup.py` which just does:
```python
self.spawn(["cmake", "--build", str(build_dir), "--", "TTMLIRPythonModules"])
self.spawn(
    ["cmake", "--install", str(build_dir), "--component", "TTMLIRPythonWheel"]
)
```

I won't block the PR on this if it's too much work, but it's definitely something we should do because it's way less fragile.

vtangTT: Agreed on using cmake install, didn't realize we were doing this for the mlir python wheel.

.github/test_scripts/pykernel.sh
```diff
@@ -10,12 +10,9 @@ export PYTHONPATH="$BUILD_DIR/python_packages:$INSTALL_DIR/tt-metal/ttnn:$INSTAL
...
 mkdir -p $WORK_DIR/third_party/tt-metal/src
 ln -sf $INSTALL_DIR/tt-metal third_party/tt-metal/src/tt-metal
-if [ ! -d "$BUILD_DIR/python_packages/ttnn_jit" ]; then
-    ln -sf $WORK_DIR/tools/ttnn-jit $BUILD_DIR/python_packages/ttnn_jit
-fi
 
-echo "Running PyKernel/ttnn-jit tests..."
-pytest -v $WORK_DIR/test/pykernel/demo/test.py $WORK_DIR/test/ttnn-jit/ --junit-xml=$TEST_REPORT_PATH
+echo "Running PyKernel tests"
+pytest -v $WORK_DIR/test/pykernel/demo/test.py
```
nsumrakTT: Please don't forget `--junit-xml=$TEST_REPORT_PATH`
as this emits results to analytics.

.github/workflows/call-test.yml
```diff
@@ -91,7 +91,21 @@ jobs:
...
+      with:
+        name: ttnn-jit-whl-${{ matrix.build.image }}
+
+    - name: Install ttnn-jit run whls
+      if: matrix.build.image == 'tracy'
+      shell: bash
+      run: |
+        source env/activate
+        pip show ttnn-jit && pip uninstall -y ttnn-jit
+        pip install ttnn_jit*.whl --upgrade
```
nsumrakTT: Is this needed for every single test we have?
If it is not, then download and install when needed through test script:
https://docs.tenstorrent.com/tt-mlir/ci.html#downloading-artifacts

nsumrakTT (changes requested):
So we are building ttnn-jit wheel in every release build configuration. Then, on each test machine we download and install it.
- Do we need ttnn-jit wheel for each test?
- Do we need ttnn-jit for each configuration?
- Do we need ttnn-jit wheel for ttnn-jit test at all? The wheel is nice feature for release to the users, but for CI I don't see why it has to be built, uploaded and downloaded when it can run from source?
- Is wheel tested for release to the end user? Do we support manylinux as pykernel and tt-mlir (and ttnn wheel in metal)?

tools/ttnn-jit/setup.py
vtangTT: Yea, this crossed my mind. I think it actually makes more sense to put it in the cmake, since we are switching the default jit build flow to use the wheel anyways. In that case, idt we need the patchelf.

vtangTT:
>     * Do we need ttnn-jit wheel for each test?
No only for the ttnn-jit tests.

>     * Do we need ttnn-jit for each configuration?
No only for the tracy build config in which we run the jit tests

To this end, I will try to isolate the wheel download/install step to only the jit tests is possible.

>     * Do we need ttnn-jit wheel for ttnn-jit test at all? The wheel is nice feature for release to the users, but for CI I don't see why it has to be built, uploaded and downloaded when it can run from source?
Yes we want to run it on CI so devs are working as close to the production workflow as possible. I don't think this blows up build times either.

>     * Is wheel tested for release to the end user? Do we support manylinux as pykernel and tt-mlir (and ttnn wheel in metal)?
To test the wheel for release to user, we'd need a representative metal dev env in python3.10, which currently doesn't seem possible on CI. We could also maybe have a flow to test the wheel release in metal in a python3.11 env.
No manylinux support yet, but that is the next step forward.

nsumrakTT:
@vtangTT Thanks for explanation.
Please contain both build and test changes to shell scripts (not workflow files) so it is clear what ttnn jit component requires.

tools/ttnn-jit/setup.py
vtangTT: Actually, on second thought, I think it might be better to keep it in the `setup.py`. The RPATH patch is only necessary for a release wheel that's to be used in a metal install. In that case, I don't think we should be polluting the RPATH in our normal dev cmake flow.

tools/ttnn-jit/setup.py
vtangTT: switched over to use cmake install

.github/test_scripts/pykernel.sh
vtangTT: fixed ty 🙏

.github/workflows/call-test.yml
vtangTT: only downloading and installing for ttnn-jit tests

https://github.com/tenstorrent/tt-mlir/actions/runs/19644538369?pr=5983

tools/ttnn-jit/setup.py
nsmithtt: @vtangTT we already program RPATH this way see `lib/CMakeLists.txt`:
```cmake
set_target_properties(TTMLIRCompiler PROPERTIES INSTALL_RPATH "$ORIGIN")
set_target_properties(TTMLIRCompiler PROPERTIES BUILD_WITH_INSTALL_RPATH TRUE)
```

It's much more robust and portable to use cmake over patchelf.

vtangTT: Switched to using cmake for rpath 🙏

.github/workflows/call-build-release.yml
```diff
@@ -156,6 +156,9 @@ jobs:
         export UPLOAD_LIST="$WORK_DIR/_components_upload.json"
         echo "[" >$UPLOAD_LIST
         echo "{\"name\":\"ttrt-whl-$BUILD_NAME\",\"path\":\"$WORK_DIR/build/tools/ttrt/build/ttrt*.whl\"}," >> $UPLOAD_LIST
+        if ls $WORK_DIR/tools/ttnn-jit/build/ttnn_jit*.whl 1> /dev/null 2>&1; then
+          echo "{\"name\":\"ttnn-jit-whl-$BUILD_NAME\",\"path\":\"$WORK_DIR/tools/ttnn-jit/build/ttnn_jit*.whl\"}," >> $UPLOAD_LIST
+        fi
```
nsumrakTT: Can you please move this to ttnn-jit.sh along with cmake to build it?

.github/workflows/call-test.yml
```diff
@@ -91,7 +91,21 @@ jobs:
...
+      with:
+        name: ttnn-jit-whl-${{ matrix.build.image }}
+
+    - name: Install ttnn-jit run whls
+      if: matrix.build.image == 'tracy' && contains(toJson(matrix.build.tests), 'ttnn_jit')
+      shell: bash
+      run: |
+        source env/activate
+        pip show ttnn-jit && pip uninstall -y ttnn-jit
+        pip install ttnn_jit*.whl --upgrade
```
nsumrakTT: But this is still inside workflow instead of ttnn-jit.sh test script.
Please move it there.

.github/workflows/call-test.yml
nsumrakTT: Please move it to test script so in the future we know exactly which component needs which downloads and installs.

.github/workflows/call-build-release.yml
vtangTT: Ah sry, I misunderstood your previous comments. They shoudl now be moved into the scripts

.github/workflows/call-test.yml
vtangTT: Should now be moved into the test script

nsumrakTT (approved):
Great, thank you!

runtime/python/CMakeLists.txt
```diff
@@ -39,11 +39,19 @@ nanobind_add_module(
...
+    INSTALL_RPATH "${JIT_WHEEL_RPATHS}"
+    BUILD_WITH_INSTALL_RPATH TRUE
+  )
+else()
+  set_target_properties(_ttmlir_runtime PROPERTIES
+    INSTALL_RPATH "$ORIGIN"
+    BUILD_RPATH "$ORIGIN;$<TARGET_FILE_DIR:TTMLIRRuntime>"
+    BUILD_WITH_INSTALL_RPATH FALSE
+  )
+endif()
```
nsmithtt, line 54: I see what you mean about changing the runtime rpath properties based on `TTMLIR_ENABLE_TTNN_JIT` being set, but this enables such a drastic cleanup of setup.py I think it's still worth it.

.github/settings/tests.json
```diff
@@ -34,6 +34,7 @@
   { "runs-on": "n300",   "image": "tracy",  "script": "runtime_debug.sh" },
   { "runs-on": "n150",   "image": "speedy",  "script": "ttnn_standalone.sh" },
   { "runs-on": "n150",   "image": "tracy",  "script": "pykernel.sh" },
+  { "runs-on": "n150",   "image": "tracy",  "script": "ttnn_jit.sh" },
   { "runs-on": "n150",   "image": "speedy",  "script": "alchemist.sh" },
```
sohaibnadeemTT, line 38: @vtangTT  can we change this to

```
{ "runs-on": "n150",   "image": "tracy",  "script": "pykernel.sh" },
 { "runs-on": ["n150", "llmbox"],   "image": "tracy",  "script": "ttnn_jit.sh" },
```

so that jit tests stay enabled on t3k/llmbox

vtangTT: yep i see the merge conflict

vtangTT: fixed.

xanderchin (approved):

PR #5996, opened 2025-11-22
GridSelection pass TM support

nsmithtt:
Tensor manipulation ops like rearrange, permute and reshape are implemented via views that result in much more complicated affine transformations.  This trips up the grid selection pass which assumes that operand layouts will be compatible with the generic consumer.

Problematic example case:
```mlir
%in : tensor<1x1x1x3x32x32xf32>
// reshape view that collapses dims z & y together
%view = view_layout %in : tensor<1x1x96x32xf32>
d2m.generic ins(%view)
```

The grid selection pass only divides along the innermost dims, leaving `%in` to at most reside on a unit grid `1x1x1` since we cannot divide the inner dims (`32x32`) further.  However, the generic chooses to be divided onto a `3x1` grid.

The solution is to use intermediate views between rewriting operands and the generic itself to account for these complications.

This PR also adds support for:
- Rewriting operands that come directly from d2m.empty ops.
- More sophisticated view_layout fold patterns that explicitly check for reblocking patterns that undo each other.  Also erasing nop views.

### DIFF
--- a/lib/Dialect/D2M/Transforms/GridSelection.cpp
+++ b/lib/Dialect/D2M/Transforms/GridSelection.cpp
@@ -669,7 +669,13 @@ recreateGenericOp(d2m::GenericOp genericOp,
   assert(optimalOperandGrids.size() == genericOp.getNumOperands());
   for (const auto &[optimalGrid, operand] :
        llvm::zip(optimalOperandGrids, genericOp->getOpOperands())) {
+
     auto definingView = operand.get().getDefiningOp<d2m::ViewLayoutOp>();
+    if (!definingView) {
+      newOperands.push_back(operand.get());
+      continue;
+    }
+
     if (genericOp.isDpsInit(&operand) && definingView) {
       // This is a workaround to avoid type checking errors during/after
       // canonicalization.  There is an offline proposal being discussed to
@@ -682,6 +688,7 @@ recreateGenericOp(d2m::GenericOp genericOp,
       newOperands.push_back(definingView.getInput());
       continue;
     }
+
     auto tensorType =
         mlir::cast<mlir::RankedTensorType>(operand.get().getType());
     auto viewTensorType = reblockTensor(tensorType, optimalGrid);


bgrady-tt (approved):
LGTM as an incremental change. The changing layouts for EmptyOps producing init operands is definitely necessary. 

HOWEVER: I think overall the layout decisions for ops must in the near term be driven by analysis of the GenericOp's indexing_maps, which impose coupling between operand layouts. Currently we are just adjusting operand layouts _independently_ and (because of the symmetry of the shape optimizer) aligning all dimensions legally. 

This approach will not work for virtual grid shapes beyond the simplest aliased input-output eltwise ops. Consider a matmul with a very high aspect-ratio `in0` and `out` operand, and a square-ish `in1`. Choosing layouts independently, the in0 and out will use virtual grid layouts, but `in1` will be conventionally block sharded; the resulting GenericOp will fail verification! A similar problem also exists with mixed L1 and DRAM tensors; they are not guaranteed to be 'optimized' to have mutually compatible shard grids.

nsmithtt:
> HOWEVER: I think overall the layout decisions for ops must in the near term be driven by analysis of the GenericOp's indexing_maps, which impose coupling between operand layouts. Currently we are just adjusting operand layouts _independently_ and (because of the symmetry of the shape optimizer) aligning all dimensions legally.
> 
> This approach will not work for virtual grid shapes beyond the simplest aliased input-output eltwise ops. Consider a matmul with a very high aspect-ratio `in0` and `out` operand, and a square-ish `in1`. Choosing layouts independently, the in0 and out will use virtual grid layouts, but `in1` will be conventionally block sharded; the resulting GenericOp will fail verification! A similar problem also exists with mixed L1 and DRAM tensors; they are not guaranteed to be 'optimized' to have mutually compatible shard grids.

I completely agree (fyi @vwellsTT).  I think we should transition the GridSelection pass to be Analysis based and work in terms of generic op constraints, i.e. using the affine maps to solve for optimal shapes.  After the ops themselves have had their ideal grids chosen, we propagate outwards towards leaf nodes and try our best to layout inputs w/ the best consumer access pattern.

Issue #4578, opened 2025-08-20
☔ LLM Fusing patterns


odjuricicTT:
In order to run LLMs more performantly, we need to implement fusing patterns for the following ops found in sub issues.

These are the ops found in the llama3.2 architecture. More will be added to the list as we take a look at other architectures.

The first model to focus on is llama 3.2 1B with batch 32. Decode perf is currently more important then prefill. In order to run it e2e and get perf metrics from tt-xla, use the local ttxla venv and run:
```bash
pytest -svv ../tt-forge/benchmark/tt-xla/llms.py::test_llama_3_2_1b
```
The above command assumes you have the tt-forge repo checked out next to tt-xla.

If testing from tt-mlir, you can use the below version if ttnn / ttir (dated 14. nov). This is llama 3.2 1B decode with batch 32:

[llama_3_2_1B_bs32_ttir.mlir.txt](https://github.com/user-attachments/files/23548713/llama_3_2_1B_bs32_ttir.mlir.txt)

[llama_3_2_1B_bs32_ttnn.mlir.txt](https://github.com/user-attachments/files/23548715/llama_3_2_1B_bs32_ttnn.mlir.txt)

Here is the device perf csv generated using emitPy and only for 2/3 of the graph (full graph hangs):
[llama_perf.csv](https://github.com/user-attachments/files/23548119/llama_perf.csv)
A sheet with per op analysis from tt-perf-analysis:
https://docs.google.com/spreadsheets/d/1RuMT-Bd52PoLsuWCv4QhLIIyyT5-sqruZiNEYr3tIys/edit?gid=0#gid=0

Device time op visual:

<img width="1800" height="2400" alt="Image" src="https://github.com/user-attachments/assets/b1b02ed3-c4d2-4620-9bbf-41bc9e0eb92c" />

LPanosTT:
I have a branch for bringing up `ttnn.transformer.paged_scaled_dot_product_attention_decode`: `lpanos/paged_attn` its unpolished and I put something together quick so I could investigate integration with vllm

LPanosTT:
> I have a branch for bringing up `ttnn.transformer.paged_scaled_dot_product_attention_decode`: `lpanos/paged_attn` its unpolished and I put something together quick so I could investigate integration with vllm

This comment is outdated. This branch is flawed. This task has been taken up by others: https://github.com/tenstorrent/tt-mlir/issues/4214?issue=tenstorrent%7Ctt-mlir%7C4264

PR #5999, opened 2025-11-23
[TTIR] Add SDPA fusing pattern

jserbedzijaTT:
### Ticket
[Link to Github Issue](https://github.com/tenstorrent/tt-mlir/issues/5534)

### Problem description

Scaled Dot Product Attention (SDPA) is a fundamental building block in transformer architectures, computing 
```
attention(Q, K, V) = softmax(Q @ K^T * scale + mask) @ V
```
Currently, we lack a fusing pattern in TTIR to combine this chain of operations into a single `ttir.scaled_dot_product_attention `op.

### What's changed

**SDPA Fusing Pattern:**
- Added a pattern in the `TTIRFusing` pass that identifies the canonical attention computation pattern: 
```
softmax(Q @ K^T * scale + mask) @ V
```
- Fuses the matched sequence of operations into a single `ttir.scaled_dot_product_attention` operation
- Handles optional attention mask and scaling factors
- Broadcasts the mask to shape `[batch_size, 1, query_seq_len, key_seq_len]` if needed

### Lowering of `ttir.scaled_dot_product_attention`
- Updated the lowering pattern of `ttir.scaled_dot_product_attention` in the `TTIRToTTNN` pass. It now lowers to either `ttnn.scaled_dot_product_attention` or `ttnn.scaled_dot_product_attention_decode` depending on the `sequence_length` dimension in the `query` tensor (if `sequence_length` is 1, it uses the decode operation).
  - Note: This logic might need improvement. If anyone has a better idea, please feel free to suggest it.

**Workarounds:**
- **Padding workaround**: Pads sequence dimensions to multiples of tile size, then slices output back to original shape
- **SDPA decode config workaround**: Currently for some cases `ttnn.scaled_dot_product_attention_decode` op requires `program_config` to succeed so we create a default one in the workaround.

**Testing:**
- Added Python and MLIR tests for fusing and workaround

### Checklist

### DIFF
--- a/lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
+++ b/lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
@@ -2935,8 +2935,9 @@ private:
 };
 } // namespace
 
+namespace {
 // ============================================================================
-// SDPA Fusing - Semantic Pattern Matching
+// SDPA Fusing
 // ============================================================================
 //
 // This pattern identifies Scaled Dot Product Attention (SDPA) by its
@@ -2953,8 +2954,6 @@ private:
 // After the matched pattern is replaced with ttir.scaled_dot_product_attention
 // op, the mask is broadcasted to shape [batch_size, 1, query_seq_len,
 // key_seq_len]
-
-namespace {
 
 class SDPAFusing : public mlir::OpRewritePattern<MatmulOp> {
   using SDPAFusing::OpRewritePattern<MatmulOp>::OpRewritePattern;
@@ -3015,7 +3014,7 @@ private:
     components.attentionMatmul = matmul;
 
     // Check if this matmul has softmax feeding into it (attention scores @ V)
-    SoftmaxOp softmax = skipLayoutOps<SoftmaxOp>(matmul.getA());
+    SoftmaxOp softmax = findOpThroughLayoutOps<SoftmaxOp>(matmul.getA());
     if (!softmax) {
       return false;
     }
@@ -3024,7 +3023,7 @@ private:
 
     // Look for optional mask addition
     Value cursor = softmax.getInput();
-    if (auto maskAdd = skipLayoutOps<AddOp>(cursor)) {
+    if (auto maskAdd = findOpThroughLayoutOps<AddOp>(cursor)) {
       if (auto maskResult = tryExtractMask(maskAdd)) {
         components.mask = maskResult->first;
         cursor = maskResult->second;
@@ -3032,7 +3031,7 @@ private:
     }
 
     // Look for optional scale multiplication
-    if (auto mulOp = skipLayoutOps<MultiplyOp>(cursor)) {
+    if (auto mulOp = findOpThroughLayoutOps<MultiplyOp>(cursor)) {
       if (auto scaleResult = tryExtractScale(mulOp)) {
         components.scale = scaleResult->first;
         cursor = scaleResult->second;
@@ -3040,7 +3039,7 @@ private:
     }
 
     // Look for Q@K matmul (required)
-    auto qkMatmul = skipLayoutOps<MatmulOp>(cursor);
+    auto qkMatmul = findOpThroughLayoutOps<MatmulOp>(cursor);
     if (!qkMatmul) {
       return false;
     }
@@ -3126,8 +3125,8 @@ private:
   // Try to extract mask from an add operation.
   // Returns {mask, score_path} if found, nullopt otherwise.
   std::optional<std::pair<Value, Value>> tryExtractMask(AddOp addOp) const {
-    auto lhsMul = skipLayoutOps<MultiplyOp>(addOp.getLhs());
-    auto rhsMul = skipLayoutOps<MultiplyOp>(addOp.getRhs());
+    auto lhsMul = findOpThroughLayoutOps<MultiplyOp>(addOp.getLhs());
+    auto rhsMul = findOpThroughLayoutOps<MultiplyOp>(addOp.getRhs());
     if (lhsMul && looksLikeMask(addOp.getRhs())) {
       return std::make_pair(addOp.getRhs(), addOp.getLhs());
     }
@@ -3158,7 +3157,7 @@ private:
   // Returns nullptr if a non-layout op is encountered before finding the
   // target.
   template <typename OpType>
-  OpType skipLayoutOps(Value v) const {
+  OpType findOpThroughLayoutOps(Value v) const {
     while (Operation *defOp = v.getDefiningOp()) {
       if (auto targetOp = dyn_cast<OpType>(defOp)) {
         return targetOp;
@@ -3241,9 +3240,6 @@ private:
     return type.getShape()[1] == 1;
   }
 };
-
-} // namespace
-
 } // namespace
 
 class TTIRFusingPass : public impl::TTIRFusingBase<TTIRFusingPass> {
@@ -3310,5 +3306,5 @@ public:
     }
   }
 };
-
+} // namespace
 } // namespace mlir::tt::ttir


lib/Conversion/TTIRToTTNN/TTIRToTTNN.cpp
```diff
@@ -2066,15 +2066,61 @@ class ScaledDotProductAttentionOpConversionPattern
...
+          op, ttir_to_ttnn::utils::generatePermute(
+                  decodeOp.getResult(), {1, 2, 0, 3}, rewriter, op.getLoc()));
+    } else {
+      rewriter.replaceOpWithNewOp<ttnn::ScaledDotProductAttentionOp>(
+          op, this->getTypeConverter()->convertType(op.getType()),
+          adaptor.getQuery(), adaptor.getKey(), adaptor.getValue(),
+          adaptor.getAttentionMask(), op.getIsCausal(), adaptor.getScaleAttr(),
+          adaptor.getSlidingWindowSizeAttr(),
+          /*memory_config=*/nullptr);
+    }
```
jserbedzijaTT: As mentioned in the description, this is probably not the best place to put the logic of deciding whether to create `ttnn.scaled_dot_product_attention` or `ttnn.scaled_dot_product_attention_decode` operation so if anyone have a better idea feel free to comment.

odjuricicTT:
@jserbedzijaTT Great!

Have you tested this on llama e2e ([details](https://github.com/tenstorrent/tt-mlir/issues/4578))? If so, could you share the resulting IR?

include/ttmlir/Dialect/TTNN/IR/TTNNOps.td
```diff
@@ -2788,10 +2788,11 @@ def TTNN_ScaledDotProductAttentionDecodeOp : TTNN_Op<"scaled_dot_product_attenti
                          AnyRankedTensor:$value,
                          DefaultValuedAttr<BoolAttr, "true">:$is_causal,
                          Optional<AnyRankedTensor>:$attention_mask,
-                         AnyRankedTensor:$cur_pos_tensor,
+                         Optional<AnyRankedTensor>:$cur_pos_tensor,
                          Optional<AnyRankedTensor>:$attention_sink,
                          OptionalAttr<F32Attr>:$scale,
-                         OptionalAttr<TTNN_MemoryConfigAttr>:$memory_config);
+                         OptionalAttr<TTNN_MemoryConfigAttr>:$memory_config,
+                         OptionalAttr<TTNN_SDPAProgramConfigAttr>:$program_config);
```
sgholamiTT, line 2883: I believe we should pass this information to optimizer (it's missing [here](https://github.com/tenstorrent/tt-mlir/blob/5b898c5703f745a51ceef256440072a901973f7c/lib/Dialect/TTNN/Interfaces/TTNNOpModelInterface.cpp#L2005-L2012)).
In general, the optimizer code should be similar to the [runtime](https://github.com/tenstorrent/tt-mlir/pull/5999/files#diff-15f5d43d3eb83dedf87cc69d035e97ac10eb476d74ec4f6d38bd6b4a8fb1ed96R47-R52) code. 

We also need whether [this logic](https://github.com/tenstorrent/tt-mlir/blob/5b898c5703f745a51ceef256440072a901973f7c/lib/Dialect/TTNN/Interfaces/TTNNOpModelInterface.cpp#L1927) needs to change or not with this modification.

include/ttmlir/Dialect/TTNN/Transforms/Workarounds/Decomposition/ScaledDotProductAttentionPadSequenceDimRewriterPattern.h
```diff
@@ -0,0 +1,35 @@
...
+#define TTMLIR_DIALECT_TTNN_TRANSFORMS_WORKAROUNDS_DECOMPOSITION_SCALEDDOTPRODUCTATTENTIONPADSEQUENCEDIMREWRITERPATTERN_H
+
+#include "ttmlir/Dialect/TTNN/IR/TTNNOps.h"
+
+#include "mlir/IR/PatternMatch.h"
+#include "mlir/Support/LogicalResult.h"
+
+namespace mlir::tt::ttnn::workarounds::decomposition {
+
+// TTNN ScaledDotProductAttention requires sequence dimensions to be aligned to
```
odjuricicTT: Is there an issue for this on metal side? If so please link it here.

jserbedzijaTT: Linked

include/ttmlir/Dialect/TTNN/IR/TTNNOps.td
jserbedzijaTT: Done

lib/Conversion/TTIRToTTNN/TTIRToTTNN.cpp
```diff
@@ -2040,15 +2040,61 @@ class ScaledDotProductAttentionOpConversionPattern
...
   LogicalResult
   matchAndRewrite(ttir::ScaledDotProductAttentionOp op, OpAdaptor adaptor,
                   ConversionPatternRewriter &rewriter) const override {
-    rewriter.replaceOpWithNewOp<ttnn::ScaledDotProductAttentionOp>(
-        op, this->getTypeConverter()->convertType(op.getType()),
-        adaptor.getQuery(), adaptor.getKey(), adaptor.getValue(),
-        adaptor.getAttentionMask(), op.getIsCausal(), adaptor.getScaleAttr(),
-        adaptor.getSlidingWindowSizeAttr(),
-        /*memory_config=*/nullptr);
+    auto queryType = mlir::cast<RankedTensorType>(op.getQuery().getType());
```
azecevic000: Redundant cast. Anyway, you should use `adaptor.getQuery()` instead of `op.getQuery()` here.

lib/Conversion/TTIRToTTNN/TTIRToTTNN.cpp
```diff
@@ -2040,15 +2040,61 @@ class ScaledDotProductAttentionOpConversionPattern
...
+
+    // If sequence length of query is 1, we should use the decode op
```
azecevic000: Nit: add `.` at the end of the sentence

lib/Conversion/TTIRToTTNN/TTIRToTTNN.cpp
```diff
@@ -2040,15 +2040,61 @@ class ScaledDotProductAttentionOpConversionPattern
...
+    if (queryType.getDimSize(2) == 1) {
```
azecevic000: Please use `enum` for symbolic names, there are a lot of magic constants in this pattern.

lib/Conversion/TTIRToTTNN/TTIRToTTNN.cpp
```diff
@@ -2040,15 +2040,61 @@ class ScaledDotProductAttentionOpConversionPattern
...
+
+      // Broadcast mask head dimension if needed for decode op
+      Value attentionMask = adaptor.getAttentionMask();
+      if (attentionMask) {
+        auto maskType = mlir::cast<RankedTensorType>(attentionMask.getType());
+        int64_t numHeads = queryType.getDimSize(1);
+
+        if (numHeads > 1) {
+          SmallVector<int64_t> broadcastShape(maskType.getShape().begin(),
+                                              maskType.getShape().end());
```
azecevic000: Nit:
```suggestion
          SmallVector<int64_t> broadcastShape(maskType.getShape());
```

lib/Conversion/TTIRToTTNN/TTIRToTTNN.cpp
```diff
@@ -2040,15 +2040,61 @@ class ScaledDotProductAttentionOpConversionPattern
...
+          op.getLoc(), permutedQuery.getType(), permutedQuery, adaptor.getKey(),
+          adaptor.getValue(), op.getIsCausal(), attentionMask,
+          /*cur_pos_tensor=*/Value(), /*attention_sink=*/Value(),
+          adaptor.getScaleAttr(), /*memory_config=*/nullptr,
+          /*program_config=*/nullptr);
+
+      // Permute result back: [1, B, H, D] -> [B, H, 1, D]
+      rewriter.replaceOp(
+          op, ttir_to_ttnn::utils::generatePermute(
+                  decodeOp.getResult(), {1, 2, 0, 3}, rewriter, op.getLoc()));
```
azecevic000: Besides symbolic names for dimensions, consider using `inversePermutation` instead of hardcoding pattern; it's a cheap calculation, but it's much easier to follow through with it.

lib/Conversion/TTIRToTTNN/Utils.cpp
```diff
@@ -42,6 +42,23 @@ generateNHWFlatten(mlir::TypedValue<mlir::RankedTensorType> input,
...
 
+ttnn::PermuteOp generatePermute(mlir::TypedValue<mlir::RankedTensorType> input,
+                                ArrayRef<int64_t> permutation,
+                                PatternRewriter &rewriter,
+                                mlir::Location newLoc) {
+  RankedTensorType inputType = input.getType();
+  llvm::SmallVector<int64_t> outputShape;
+  for (int64_t dim : permutation) {
+    outputShape.push_back(inputType.getDimSize(dim));
+  }
```
azecevic000:
```suggestion
  llvm::SmallVector<int64_t> outputShape =
      ttmlir::utils::applyPermutation(inputType.getShape(), permutation);
```

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
```diff
@@ -2681,6 +2681,335 @@ class GeluFusionPattern : public mlir::OpRewritePattern<ttir::MultiplyOp> {
...
+      if (auto targetOp = dyn_cast<OpType>(defOp)) {
+        return targetOp;
+      }
+
+      if (isLayoutOp(defOp)) {
+        v = defOp->getOperand(0);
+        continue;
+      }
+
+      break;
```
azecevic000: This is very convoluted logic. I would rewrite it as
```cpp
  if (!isLayoutOp(defOp)) {
    return {};
  }
  v = defOp->getOperand(0);
```
i.e., without `continue/break`.

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
```diff
@@ -2681,6 +2681,335 @@ class GeluFusionPattern : public mlir::OpRewritePattern<ttir::MultiplyOp> {
...
+      }
+
+      if (!v.hasOneUse()) {
+        return false;
+      }
+
+      if (isLayoutOp(defOp)) {
+        v = defOp->getOperand(0);
+        continue;
+      }
```
azecevic000: Why are only layout ops considered, it's not mentioned anywhere in the docstring of the function.

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
```diff
@@ -2681,6 +2681,335 @@ class GeluFusionPattern : public mlir::OpRewritePattern<ttir::MultiplyOp> {
...
+    }
+
+    if (auto fullOp = dyn_cast<FullOp>(defOp)) {
+      if (auto fillValueAttr = dyn_cast<FloatAttr>(fullOp.getFillValue())) {
+        return fillValueAttr.getValue().convertToFloat();
+      }
+      if (auto fillValueAttr = dyn_cast<IntegerAttr>(fullOp.getFillValue())) {
+        return static_cast<float>(fillValueAttr.getValue().getSExtValue());
+      }
+    }
```
azecevic000:
```suggestion
    if (auto fullOp = scaleVal.getDefiningOp<FullOp>()) {
      return llvm::TypeSwitch<mlir::Attribute, float>(fullOp.getFillValue())
        .Case([](FloatAttr fillValueAttr) {
          return fillValueAttr.getValue().convertToFloat();
        })
        .Case([](IntegerAttr fillValueAttr) {
          return static_cast<float>(fillValueAttr.getValue().getSExtValue());
        });
    }
```

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
```diff
@@ -2681,6 +2681,335 @@ class GeluFusionPattern : public mlir::OpRewritePattern<ttir::MultiplyOp> {
...
+      outMask = addOp.getLhs();
+      outScorePath = addOp.getRhs();
+      return true;
+    }
+
+    return false;
+  }
+
+  // Try to extract scale from a multiply operation
+  bool tryExtractScale(MultiplyOp mulOp, float &outScale,
```
azecevic000: IMO `std::optional<T> f()` is much more better API choice than `bool f(T&)`.

lib/Dialect/TTNN/Transforms/Workarounds/Decomposition/ScaledDotProductAttentionDecodeConfigRewritePattern.cpp
```diff
@@ -0,0 +1,46 @@
...
+ScaledDotProductAttentionDecodeConfigRewritePattern::matchAndRewrite(
+    ScaledDotProductAttentionDecodeOp srcOp, PatternRewriter &rewriter) const {
+  if (srcOp.getProgramConfigAttr()) {
+    return failure();
+  }
+
+  MLIRContext *context = srcOp.getContext();
+
+  // Create a default SDPA program config
+  auto computeGridSize = CoreCoordAttr::get(context, /*x=*/8, /*y=*/8);
```
azecevic000: It's not the best approach to hard-code (8, 8) value. Is there a way to query device grid size here? For Blackhole, it's not going to be (8, 8) (it might even succeed, but performance is going to be suboptimal).

lib/Dialect/TTNN/Transforms/Workarounds/Decomposition/ScaledDotProductAttentionDecodeConfigRewritePattern.cpp
```diff
@@ -0,0 +1,46 @@
+// SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#include "ttmlir/Dialect/TTNN/Transforms/Workarounds/Decomposition/ScaledDotProductAttentionDecodeConfigRewritePattern.h"
```
azecevic000, line 5: Nit: add new line separator below this one

lib/Dialect/TTNN/Transforms/Workarounds/Decomposition/ScaledDotProductAttentionPadSequenceDimRewriterPattern.cpp
```diff
@@ -0,0 +1,189 @@
+// SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#include "ttmlir/Dialect/TTNN/Transforms/Workarounds/Decomposition/ScaledDotProductAttentionPadSequenceDimRewriterPattern.h"
```
azecevic000, line 5: Same here

lib/Dialect/TTNN/Transforms/Workarounds/Decomposition/ScaledDotProductAttentionDecodeConfigRewritePattern.cpp
```diff
@@ -0,0 +1,46 @@
...
+#include "ttmlir/Dialect/TTNN/IR/TTNNOpsAttrs.h"
+
+// Workaround which adds a default SDPAProgramConfig to
+// ScaledDotProductAttentionDecodeOp when no config is present because
+// it will fail otherwise.
+// Metal issue reference: https://github.com/tenstorrent/tt-metal/issues/32641
```
azecevic000: It's enough to have this comment in the header file, you can remove it here (same for the other decomp pattern).

lib/Dialect/TTNN/Transforms/Workarounds/Decomposition/ScaledDotProductAttentionPadSequenceDimRewriterPattern.cpp
```diff
@@ -0,0 +1,189 @@
...
+  auto tensorType = mlir::dyn_cast<RankedTensorType>(tensor.getType());
+  if (!tensorType) {
+    return tensor;
+  }
+
+  ArrayRef<int64_t> shape = tensorType.getShape();
+  int64_t rank = tensorType.getRank();
+  int64_t padAmount = targetLen - shape[dim];
+
+  SmallVector<int64_t> paddedShape(shape.begin(), shape.end());
```
azecevic000: Nit: `paddedShape(shape)`

lib/Dialect/TTNN/Transforms/Workarounds/Decomposition/ScaledDotProductAttentionPadSequenceDimRewriterPattern.cpp
```diff
@@ -0,0 +1,189 @@
...
+  SmallVector<int32_t> padding(rank * 2, 0);
+  padding[dim * 2 + 1] = padAmount;
+
+  auto paddedType =
+      utils::RankedTensorTypeFactory::create(tensorType, paddedShape);
+
+  return rewriter
+      .create<PadOp>(
+          loc, paddedType, tensor, rewriter.getDenseI32ArrayAttr(padding),
+          rewriter.getF32FloatAttr(padValue), rewriter.getBoolAttr(true),
```
azecevic000: Nit: add `/*param=*/` for literal true value.

lib/Dialect/TTNN/Transforms/Workarounds/Decomposition/ScaledDotProductAttentionDecodeConfigRewritePattern.cpp
jserbedzijaTT: Good catch!

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
jserbedzijaTT: Simplified

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
jserbedzijaTT: I ended up checking if every op in the chain has only single use regardless of the op type

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
jserbedzijaTT: I agree

tools/golden/mapping.py
```diff
@@ -2384,6 +2384,21 @@ def arange_golden(single_dim_tensor: GoldenMapTensor, **kwargs) -> GoldenMapTens
     return GoldenMapTensor(output_shards, single_dim_tensor.mesh_shape)
 
 
+def full_golden(**kwargs) -> GoldenMapTensor:
```
tapspatel: is this being used anywhere?

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
```diff
@@ -2843,6 +2844,317 @@ class GeluFusionPattern : public mlir::OpRewritePattern<ttir::MultiplyOp> {
...
+        return true;
+      }
+
+      current = defOp->getOperand(0);
+    }
+
+    return false;
+  }
+
+  Value traceToSourceTensor(Value v) const {
```
jameszianxuTT, line 3204: nit; It seems like there are a lot of variations on this backwards graph traversal and custom predicate/return type with slight differences between them here.  This one for example seems very similar to `skipLayoutOps`. I'm not sure if it's worth the effort to refactor, and this is just an observation. 

Also, the naming of `skipLayoutOps` is pretty confusing, since it's not obvious that it's doing a graph traversal.

runtime/lib/ttnn/operations/transformer/scaled_dot_product_attention_decode.cpp
```diff
@@ -21,10 +21,14 @@ static void runScaledDotProductAttentionDecodeOp(
   const ::ttnn::Tensor &key = tensorPool.getTTNNTensorAndValidate(op->key());
   const ::ttnn::Tensor &value =
       tensorPool.getTTNNTensorAndValidate(op->value());
-  const ::ttnn::Tensor &curPosTensor =
-      tensorPool.getTTNNTensorAndValidate(op->cur_pos_tensor());
   bool isCausal = op->is_causal();
 
+  std::optional<::ttnn::Tensor> curPosTensor = std::nullopt;
```
jameszianxuTT, line 26: The below comment on line 53 probably should be changed 
>   // The current position information is required for this op. It can either be
  // passed as a tensor or as a uint vector. The uint vector is not wrapped in a
  // std::optional so we must pass an empty vector.

tools/golden/mapping.py
```diff
@@ -3498,10 +3513,11 @@ def ttir_clamp_tensor_golden(
 def ttir_full_golden(
     shape_attr: DenseI32ArrayAttr,
     fill_value_attr: Union[IntegerAttr, FloatAttr],
+    dtype: torch.dtype = torch.float32,
```
tapspatel: I just merged some refactoring changes: https://github.com/tenstorrent/tt-mlir/commit/4ead5b2d66189abed9b74f89a6e858f782de5329

you can remove this since the output golden type will be based on what the user passes

```
def full(
        self,
        output_shape: List[int],
        output_type: torch.dtype,
        fill_value: Union[int, float],
        loc: Optional[str] = None,
    ) -> OpView:
        ttir_op = self.get_opview_from_method(TTIRBuilder.full)
        mlir_output_type = self._get_type_from_torch_dtype(output_type)
```

tools/golden/mapping.py
jserbedzijaTT: Nope, good catch

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
jserbedzijaTT: I renamed skipLayoutOps, but I won’t refactor the logic for now. I might do it in the future if a similar need arises for other operations.

runtime/lib/ttnn/operations/transformer/scaled_dot_product_attention_decode.cpp
jserbedzijaTT: I think it’s best to leave it as is, since it’s related to the curPosEmpty vector.

azecevicTT (approved):

Issue #6011, opened 2025-11-24
Add op-bypass feature from chisel into builder


jgrimTT:
### Is your feature request related to a problem? Please describe

Chisel offers a way to bypass a runtime operation by replacing the op output with the op's golden output. This feature needs to be migrated to builder

PR #6012, opened 2025-11-24
Add bypass-ops feature to builder

jgrimTT:
### Ticket
Closes [#6011](https://github.com/tenstorrent/tt-mlir/issues/6011)

### Problem description
Builder has no mechanism for bypassing an op that is malfunctioning in runtime. This feature has been requested

### What's changed
Added list of op locations `self._bypass_ops` to builder, and a function `builder.bypass(operand)` to add to that list. 
Added structure in post op callback to update device tensors with golden tensors when location matches. 

### Checklist

### DIFF
--- a/tools/builder/base/builder_runtime.py
+++ b/tools/builder/base/builder_runtime.py
@@ -282,7 +282,7 @@ class CallbackRuntimeConfig:
         check_atol: bool = True,
         check_rtol: bool = True,
         goldens={},
-        bypass_ops=[],
+        bypass_ops=None,
     ):
         self.device = device
         self.pcc = pcc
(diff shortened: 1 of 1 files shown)


tools/builder/base/builder_runtime.py
```diff
@@ -423,6 +446,7 @@ def execute_fb(
     check_atol: bool = False,
     check_rtol: bool = False,
     enable_intermediate_verification: bool = False,
+    bypass_ops: List[str] = [],
```
tapspatel: make this default to None

and in the function, 

if bypass_ops == None:
    bypass_ops = [ ]

tools/builder/base/builder_runtime.py
```diff
@@ -270,6 +282,7 @@ def __init__(
         check_atol: bool = True,
         check_rtol: bool = True,
         goldens={},
+        bypass_ops=[],
```
tapspatel: for goldens + bypass_ops, default to None and set empty dict/list in the init function.

tools/builder/base/builder_runtime.py
tapspatel: default to None and set empty list in function

Issue #5813, opened 2025-11-12
Reinstate Correct `ttnn.linear` Lowering


ctodTT:
### Describe the bug

As a result of an uplift (#5744), the behavior of `ttnn.linear` lowering had to be changed for a hotfix. The bug was that as of `tt-metal` commit `f9dc6d2`, more biases around `ttnn.linear` ops are fused. This triggers a runtime verification check in `tt-metal`, breaking most tests that contain a `ttnn.linear` op. To circumvent this, we decided to temporarily force decomposition of `ttnn.linear` ops to prevent the new behavior from triggering. This means a whole new set of tests had to be disabled (see #5789) for tracking on that.

To fix this, logic surrounding `ttnn.linear` lowering must be constructed to only generate fused ops that pass the bias shape check triggered by this change.

PR #5744, opened 2025-11-06
Uplift third_party/tt-metal to 094a4ed7d519a3d099831c0345ffa2a6d13f5b94 2025-11-06

vmilosevic:
This PR uplifts the third_party/tt-metal to the 094a4ed7d519a3d099831c0345ffa2a6d13f5b94

### Note

This uplift notably creates a workaround for a bug introduced by [this](https://github.com/tenstorrent/tt-metal/commit/f9dc6d2a9550375122c667fb04b417dc60d988cc) metal commit. The new behavior is now that `ttnn.linear` is _always_ lowered to a matmul + add to avoid triggering the new DRAM interleaving in metal with bad bias shapes. This should be treated as a regression.

This uplift necesarrily touches a lot of tests with the new temporary fusion behavior (always decompose until a better solution to the bias mismatch of the new dram interleaved linear ops can be implemented)

This means that a good number of the tests in this change must be either reimplemented or have their changes reverted. To this end, all the test changes have been collected in the last commit of this PR. An issue to track these general test changes can be found here: #5789 . Please create any sub-issues as you see fit

### Checklist
- **Frontend CI passing links**
  - [x] [tt-forge-fe](https://github.com/tenstorrent/tt-forge-fe/actions/workflows/on-pr.yml): https://github.com/tenstorrent/tt-forge-fe/actions/runs/19312424550
  - [x] [tt-xla (basic + models)](https://github.com/tenstorrent/tt-xla/actions/workflows/manual-test.yml): https://github.com/tenstorrent/tt-xla/actions/runs/19312426586
- **Follow-up Actions**
  - [x] **Issues filed** to follow up on incomplete changes (if any):
  - [x] **Frontend fix PRs** ready (if needed by this uplift):

Issue #5858, opened 2025-11-14
Fusing matmul + add into linear with bias


odjuricicTT:
After https://github.com/tenstorrent/tt-mlir/commit/b70f5f52bf39746383901626c2d6737ab6ee8852

We don't get liner ops in latest llama batch 32, we need to reenable this for perf reasons.

rpavlovicTT:
https://github.com/tenstorrent/tt-metal/pull/32510 this case was fixed, so potentially fusing should work in more cases

odjuricicTT:
As disscussed with @mtopalovicTT, the solution for this might be to not broadcast the bias which would keep the batch dim at 1 and fusing should be supported then.

ddilbazTT:
@odjuricicTT @rpavlovicTT I put up a PR for this: https://github.com/tenstorrent/tt-mlir/pull/6014 @mtopalovicTT and I had noticed that it's best if we model pytorch linear op, where bias is 1d.

ddilbazTT:
Closing this issue since [my PR went in Friday](https://github.com/tenstorrent/tt-mlir/commit/4937cb9e271ba2a07819c669a8cba40e1790d0ea) Now, we are not decomposing all linear ops into matmul and add.

PR #6014, opened 2025-11-25
Change linear op rewrite pattern 

ddilbazTT:
### Ticket
https://github.com/tenstorrent/tt-mlir/issues/5813
https://github.com/tenstorrent/tt-mlir/issues/6016

### Problem description
[During uplift 2 weeks ago](https://github.com/tenstorrent/tt-mlir/pull/5744), default tt-metal/ttnn linear op behaviour has changed. linear op is now mapping [PyTorch Linear](https://docs.pytorch.org/docs/stable/generated/torch.nn.Linear.html) This means that linear op bias needs to be 1d (or view, like 1x1x64)

@ctodTT relaxed the conditions to linear op workaround so that all linear ops would be broken down into matmul and add. However, this means no linear op is running as a linear op, even if they have 1d bias.

### What's changed
Linear op workaround is changed such that:
- only linear ops with non 1d bias or batched b input are broken down into matmul + add
- activation is passed from linear to matmul op correctly (previously, this was not passed)
- Tests marked unsupported or skipped are reinstated. 

### Checklist
- [Performance Benchmark](https://github.com/tenstorrent/tt-xla/actions/runs/19769666488 ): 
    - Errors seen are the same errors seen in other performance benchmark runs. I.e. (Circular buffer error, cv2 not found)
- [Model test push](https://github.com/tenstorrent/tt-xla/actions/runs/19769681564)
- [Basic Test](https://github.com/tenstorrent/tt-xla/actions/runs/19769684308)
- [Model Test Passing](https://github.com/tenstorrent/tt-xla/actions/runs/19772587791)

### DIFF
--- a/lib/Dialect/TTNN/Transforms/Workarounds/Decomposition/LinearOpRewritePattern.cpp
+++ b/lib/Dialect/TTNN/Transforms/Workarounds/Decomposition/LinearOpRewritePattern.cpp
@@ -14,20 +14,40 @@
 
 namespace mlir::tt::ttnn::workarounds::decomposition {
 
-// static bool isBatchedLinearOp(ttnn::LinearOp linearOp) {
-//   RankedTensorType inputBType = linearOp.getB().getType();
-//   auto inputBShape = inputBType.getShape();
-//   int64_t rank = inputBShape.size();
-//
-//   // if rank <= 2, it cannot be batched. Return false.
-//   // Check if batched: any dimension before the last 2 has size > 1.
-//   // i.e. <1x3x128x32xbf16> is batched because of 3.
-//   // i.e. <1x1x128x32xbf16> is not batched because all dims before last 2
-//   are 1.
-//   // i.e. <128xbf16> is not batched because rank < 2.
-//   return rank > 2 && llvm::any_of(inputBShape.drop_back(2),
-//                                   [](int64_t dim) { return dim > 1; });
-// }
+static bool isBatchedLinearOp(ttnn::LinearOp linearOp) {
+  RankedTensorType inputBType = linearOp.getB().getType();
+  auto inputBShape = inputBType.getShape();
+  int64_t rank = inputBShape.size();
+
+  // if rank <= 2, it cannot be batched. Return false.
+  // Check if batched: any dimension before the last 2 has size > 1.
+  // i.e. <1x3x128x32xbf16> is batched because of 3.
+  // i.e. <1x1x128x32xbf16> is not batched because all dims before last 2 are 1.
+  // i.e. <128xbf16> is not batched because rank < 2.
+  return rank > 2 && llvm::any_of(inputBShape.drop_back(2),
+                                  [](int64_t dim) { return dim > 1; });
+}
+
+// Helper function to check if bias is effectively 1D
+// Returns true if bias has only one non-unit dimension (e.g., <64>, <1x64>,
+// <1x1x64>).
+// Expects every dimension except last to be either 1 or absent.
+static bool isNotEffectively1DBias(TypedValue<RankedTensorType> bias) {
+  if (!bias) {
+    return false;
+  }
+
+  RankedTensorType biasType = bias.getType();
+  if (!biasType) {
+    return false;
+  }
+
+  auto biasShape = biasType.getShape();
+  auto rank = biasType.getRank();
+
+  return rank > 1 && llvm::any_of(biasShape.drop_back(1),
+                                  [](int64_t dim) { return dim > 1; });
+}
 
 // Calculate the output shape of a matmul operation following tt-metal's logic.
 // Reference: ttnn/cpp/ttnn/operations/matmul/matmul.cpp
@@ -37,13 +57,38 @@ computeMatmulOutputShape(llvm::ArrayRef<int64_t> shapeA, bool transposeA,
   int64_t rankA = shapeA.size();
   int64_t rankB = shapeB.size();
 
-  // if (rankA == 1 || rankB == 1) {
-  //   TT_assertv(false,
-  //              "Should not reach linear op workaround if rankA or rankB is
-  //              1");
-  // }
-
   SmallVector<int64_t> outputShape;
+
+  // Handle rank 1 cases
+  if (rankA == 1 && rankB == 1) {
+    // vector dot vector -> scalar (but represented as 1D tensor with size 1)
+    outputShape.push_back(1);
+    return outputShape;
+  }
+
+  if (rankA == 1) {
+    // vector-matrix: (K,) x (..., K, N) -> (..., N)
+    // Result shape is all batch dims from B plus the last dim
+    outputShape.append(shapeB.begin(), shapeB.end() - 2);
+    outputShape.push_back(transposeB ? shapeB[rankB - 2] : shapeB[rankB - 1]);
+    return outputShape;
+  }
+
+  if (rankB == 1) {
+    // matrix-vector: (..., M, K) x (K,) -> (..., M)
+    // Result shape is all dims from A except the last (contraction) dim
+    if (transposeA) {
+      // If A is transposed, the contraction dim is second-to-last, keep last
+      outputShape.append(shapeA.begin(), shapeA.end() - 2);
+      outputShape.push_back(shapeA[rankA - 1]);
+    } else {
+      // Normal case: contraction dim is last, keep all but last
+      outputShape.append(shapeA.begin(), shapeA.end() - 1);
+    }
+    return outputShape;
+  }
+
+  // Both inputs are at least rank 2
   SmallVector<int64_t> batchShapeA(shapeA.begin(), shapeA.end() - 2);
   SmallVector<int64_t> batchShapeB(shapeB.begin(), shapeB.end() - 2);
   mlir::OpTrait::util::getBroadcastedShape(batchShapeA, batchShapeB,
@@ -72,6 +117,18 @@ LogicalResult
 LinearOpRewritePattern::matchAndRewrite(ttnn::LinearOp srcOp,
                                         PatternRewriter &rewriter) const {
 
+  // Only decompose if bias exists AND (bias is non-1D OR input B is batched)
+  if (!srcOp.getBias()) {
+    return failure();
+  }
+
+  bool biasIsNon1D = isNotEffectively1DBias(srcOp.getBias());
+  bool inputBIsBatched = isBatchedLinearOp(srcOp);
+
+  if (!biasIsNon1D && !inputBIsBatched) {
+    return failure();
+  }
+
   RankedTensorType inputAType = srcOp.getA().getType();
   RankedTensorType inputBType = srcOp.getB().getType();
   RankedTensorType outputType = srcOp.getResult().getType();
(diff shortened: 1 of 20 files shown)


ddilbazTT:
@sdjordjevicTT @jserbedzijaTT I believe this PR would address missing activations in linears with activations. However, I believe you might also consider changing the linear with activation fusing pattern to absorb the repeat that comes before. That way, they would remain as linear with 1d bias.

ddilbazTT:
> Does this PR solve #5858?

Yes, I'll leave a comment there

lib/Dialect/TTNN/Transforms/Workarounds/Decomposition/LinearOpRewritePattern.cpp
```diff
@@ -91,11 +153,18 @@ LinearOpRewritePattern::matchAndRewrite(ttnn::LinearOp srcOp,
...
+    activationAttr = rewriter.getStringAttr(activation.value());
+  }
+
   MatmulOp matmulOp = rewriter.create<ttnn::MatmulOp>(
       ttmlir::utils::appendLocationSuffix(srcOp.getLoc(), "_decomp_matmul"),
       matmulOutputType, srcOp.getA(), srcOp.getB(), srcOp.getTransposeA(),
       srcOp.getTransposeB(),
-      /*matmul_program_config=*/mlir::Attribute(), /*activation=*/nullptr);
+      /*matmul_program_config=*/mlir::Attribute(),
+      /*activation=*/activationAttr);
```
mtopalovicTT: I'm not sure if this is correct. Activation is usually last op in the chain. For example we have matmul fallowed by add which is followed by some activation function.

lib/Dialect/TTNN/Transforms/Workarounds/Decomposition/LinearOpRewritePattern.cpp
```diff
@@ -14,20 +14,45 @@
...
+
+  for (int64_t dim : biasShape) {
+    if (dim > 1) {
+      nonUnitDims++;
+    }
+  }
+
+  // Bias is effectively 1D if it has at most one non-unit dimension
+  return nonUnitDims <= 1;
+}
```
mtopalovicTT, line 50:
```suggestion
  return rank > 1 && llvm::any_of(biasShape.drop_back(1),
                                  [](int64_t dim) { return dim > 1; });
```

ddilbazTT: I don't think this is equivalent to the original code. I changed it to 
```
// Helper function to check if bias is effectively 1D
// Returns true if bias has only one non-unit dimension (e.g., <64>, <1x64>,
// <1x1x64>)
static bool isEffectively1DBias(TypedValue<RankedTensorType> bias) {
  if (!bias) {
    return false;
  }

  RankedTensorType biasType = bias.getType();
  if (!biasType) {
    return false;
  }

  auto biasShape = biasType.getShape();
  
  // Count non-unit dimensions.
  return llvm::count_if(biasShape, [](int64_t dim) { return dim > 1; }) <= 1;
}
```
instead.

lib/Dialect/TTNN/Transforms/Workarounds/Decomposition/LinearOpRewritePattern.cpp
ddilbazTT: I'm passing linear activation attribute to matmul activation attribute when linear is decomposed. Is that wrong?

ddilbazTT: @sdjordjevicTT could you also review this?

ddilbazTT: Also @jserbedzijaTT - could you check if I'm passing in the activations correctly?

jserbedzijaTT: Looks like you are trying to rewrite activation(linear) into activation(matmul) + bias which is not mathematically correct, you should wrap both matmul and bias into the same activation function like activation(matmul + bias)

jserbedzijaTT:
A couple of weeks ago, I created an issue on the metal side to change the Activation parameter of matmul from a string to UnaryWithParam, and it looks like they implemented it:

```cpp
using Activation = std::variant<std::string, UnaryWithParam>;

struct MatmulOperation {
    static Tensor invoke(
        const Tensor& input_tensor_a,
        const Tensor& input_tensor_b,
        bool transpose_a = false,
        bool transpose_b = false,
        const std::optional<const MemoryConfig>& memory_config = std::nullopt,
        std::optional<const DataType> dtype = std::nullopt,
        const std::optional<const MatmulProgramConfig>& program_config = std::nullopt,
        const std::optional<const Activation>& activation = std::nullopt,
        std::optional<const DeviceComputeKernelConfig> compute_kernel_config = std::nullopt,
        std::optional<const CoreGrid> core_grid = std::nullopt,
        const std::optional<const tt::tt_metal::Tile>& output_tile = std::nullopt,
        std::optional<Tensor> optional_output_tensor = std::nullopt,
        const std::optional<const GlobalCircularBuffer>& global_cb = std::nullopt,
        const std::optional<tt::tt_metal::SubDeviceId>& sub_device_id = std::nullopt);
};
```

I’m not sure if it’s too much work to update matmul in TTNN to use this, but it would be a nice change.

lib/Dialect/TTNN/Transforms/Workarounds/Decomposition/LinearOpRewritePattern.cpp
mtopalovicTT: that won't work in case 1x64x1x1. You need to drop last dim.

lib/Dialect/TTNN/Transforms/Workarounds/Decomposition/LinearOpRewritePattern.cpp
mtopalovicTT: it should be either
matmul -> activation
or 
matmul -> add -> activation

ddilbazTT:
> A couple of weeks ago, I created an issue on the metal side to change the Activation parameter of matmul from a string to UnaryWithParam, and it looks like they implemented it:
> 
> ```c++
> using Activation = std::variant<std::string, UnaryWithParam>;
> 
> struct MatmulOperation {
>     static Tensor invoke(
>         const Tensor& input_tensor_a,
>         const Tensor& input_tensor_b,
>         bool transpose_a = false,
>         bool transpose_b = false,
>         const std::optional<const MemoryConfig>& memory_config = std::nullopt,
>         std::optional<const DataType> dtype = std::nullopt,
>         const std::optional<const MatmulProgramConfig>& program_config = std::nullopt,
>         const std::optional<const Activation>& activation = std::nullopt,
>         std::optional<const DeviceComputeKernelConfig> compute_kernel_config = std::nullopt,
>         std::optional<const CoreGrid> core_grid = std::nullopt,
>         const std::optional<const tt::tt_metal::Tile>& output_tile = std::nullopt,
>         std::optional<Tensor> optional_output_tensor = std::nullopt,
>         const std::optional<const GlobalCircularBuffer>& global_cb = std::nullopt,
>         const std::optional<tt::tt_metal::SubDeviceId>& sub_device_id = std::nullopt);
> };
> ```
> 
> I’m not sure if it’s too much work to update matmul in TTNN to use this, but it would be a nice change.

Is it ok if it's a different PR? I think this should be about cleaning up ttnn linear op workaround.

jserbedzijaTT:
> > A couple of weeks ago, I created an issue on the metal side to change the Activation parameter of matmul from a string to UnaryWithParam, and it looks like they implemented it:
> > ```c++
> > using Activation = std::variant<std::string, UnaryWithParam>;
> > 
> > struct MatmulOperation {
> >     static Tensor invoke(
> >         const Tensor& input_tensor_a,
> >         const Tensor& input_tensor_b,
> >         bool transpose_a = false,
> >         bool transpose_b = false,
> >         const std::optional<const MemoryConfig>& memory_config = std::nullopt,
> >         std::optional<const DataType> dtype = std::nullopt,
> >         const std::optional<const MatmulProgramConfig>& program_config = std::nullopt,
> >         const std::optional<const Activation>& activation = std::nullopt,
> >         std::optional<const DeviceComputeKernelConfig> compute_kernel_config = std::nullopt,
> >         std::optional<const CoreGrid> core_grid = std::nullopt,
> >         const std::optional<const tt::tt_metal::Tile>& output_tile = std::nullopt,
> >         std::optional<Tensor> optional_output_tensor = std::nullopt,
> >         const std::optional<const GlobalCircularBuffer>& global_cb = std::nullopt,
> >         const std::optional<tt::tt_metal::SubDeviceId>& sub_device_id = std::nullopt);
> > };
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I’m not sure if it’s too much work to update matmul in TTNN to use this, but it would be a nice change.
> 
> Is it ok if it's a different PR? I think this should be about cleaning up ttnn linear op workaround.

Yep no worries it just came to mind when I saw Activation attribute being passed to matmul :D

lib/Dialect/TTNN/Transforms/Workarounds/Decomposition/LinearOpRewritePattern.cpp
ddilbazTT: I just changed it, I would appreciate a look. In terms of activations, we are talking about: `relu`, `sigmoid`, `gelu`, `tanh` right? Is there any that I'm missing? I just updated `lib/Dialect/TTNN/Transforms/Workarounds/Decomposition/LinearOpRewritePattern.cpp` 

```
  // Step 3: If original linear op had activation, apply it to the add result.
  if (srcOp.getActivation()) {
    std::string activationStr = srcOp.getActivation()->str();
    Value activationResult;

    if (activationStr == "relu") {
      activationResult = rewriter.create<ttnn::ReluOp>(
          ttmlir::utils::appendLocationSuffix(srcOp.getLoc(), "_decomp_relu"),
          outputType, addOp.getResult());
    } else if (activationStr == "sigmoid") {
      activationResult = rewriter.create<ttnn::SigmoidOp>(
          ttmlir::utils::appendLocationSuffix(srcOp.getLoc(),
                                              "_decomp_sigmoid"),
          outputType, addOp.getResult());
    } else if (activationStr == "gelu") {
      activationResult = rewriter.create<ttnn::GeluOp>(
          ttmlir::utils::appendLocationSuffix(srcOp.getLoc(), "_decomp_gelu"),
          outputType, addOp.getResult());
    } else if (activationStr == "tanh") {
      activationResult = rewriter.create<ttnn::TanhOp>(
          ttmlir::utils::appendLocationSuffix(srcOp.getLoc(), "_decomp_tanh"),
          outputType, addOp.getResult());
    } else {
      matmulOp.emitError()
          << "Unsupported activation type in LinearOp decomposition: "
          << activationStr;
      return failure();
    }
    rewriter.replaceOp(srcOp, activationResult);
  } else {
    rewriter.replaceOp(srcOp, addOp.getResult());
  }
```

lib/Dialect/TTNN/Transforms/Workarounds/Decomposition/LinearOpRewritePattern.cpp
ddilbazTT: Just to be clear - isn't `1x64x1x1` a valid bias shape?

ddilbazTT: Doesn't this check if there is a dim greater than 1 except the last dim? So it would return true for dim shape 32x32, for example.

mtopalovicTT: Only last dim should be > 1, otherwise it will be broadcasted.

> Doesn't this check if there is a dim greater than 1 except the last dim? So it would return true for dim shape 32x32, for example.

My suggestion yes, but you can adjust it to your needs. I only wanted to show piece of code that checks if any dim except last one is > 1.

lib/Dialect/TTNN/Transforms/Workarounds/Decomposition/LinearOpRewritePattern.cpp
```diff
@@ -90,21 +145,54 @@ LinearOpRewritePattern::matchAndRewrite(ttnn::LinearOp srcOp,
...
-  rewriter.replaceOp(srcOp, addOp.getResult());
+  // Step 3: If original linear op had activation, apply it to the add result.
+  if (srcOp.getActivation()) {
+    std::string activationStr = srcOp.getActivation()->str();
+    Value activationResult;
+
+    if (activationStr == "relu") {
+      activationResult = rewriter.create<ttnn::ReluOp>(
+          ttmlir::utils::appendLocationSuffix(srcOp.getLoc(), "_decomp_relu"),
+          outputType, addOp.getResult());
```
mtopalovicTT: You can use this builder for all operations instead of doing them separately.

  /// Creates an operation with the given fields.
  Operation *create(Location loc, StringAttr opName, ValueRange operands,
                    TypeRange types = {},
                    ArrayRef<NamedAttribute> attributes = {},
                    BlockRange successors = {},
                    MutableArrayRef<std::unique_ptr<Region>> regions = {});

lib/Dialect/TTNN/Transforms/Workarounds/Decomposition/LinearOpRewritePattern.cpp
ddilbazTT: So I tried a few bias shapes:
- `<2x1x1>` - does not run because batch_size > 1
- `<1x34x1>` - runs on silicon
I think I need to add a check for batch_size > 1 when rank > 1 but otherwise, I think the function I have is correct. Let me know.

ddilbazTT: I think you're right, my bad for making this convoluted. I should only check last dim is not 1 and everything else is 1. I think with your suggestion I need to invert the return value or change the name of the function to `notEffectively1DBias` or something. However, <1x34x1> is not really an accurate linear bias if we're referencing torch, even if it can run on tt silicon.

lib/Dialect/TTNN/Transforms/Workarounds/Decomposition/LinearOpRewritePattern.cpp
ddilbazTT: Great suggestion, thanks!

Issue #5909, opened 2025-11-18
TTIR builder: Segfault when optimizer is enabled


bmalesevicTT:
This tool would be very beneficial for testing optimizer/sharding impact on validity of ops e2e, however it currently [breaks with a segfault #5283](https://github.com/tenstorrent/tt-mlir/issues/5283) in both tests that try enabling optimizer. Conv2d sharding test is referenced in the previous issue, while another example is added in [[MLA] Enable sharding for RMSNorm op #5903](https://github.com/tenstorrent/tt-mlir/issues/5903) . 
Both tests avoid the segfault if using `"enable-optimizer=false"` in pipeline options (which also disables sharding, so asserts for sharded output must be commented out for the test to pass).

The likely cause is a need for device access in the Optimizer (e.g., via `getOpConstraints`), which is not properly handled in the Python bindings for this section of the pipeline; currently, the device is only accessible during the execution phase. Quick attempts to add device to the TTIR→TTNN pipeline led to circular dependencies and other architectural complications.

### Impact:
- Blocks comprehensive end-to-end tests with optimizer and sharding enabled.
- Prevents validation of correctness and performance for sharded ops.
- Blocks enabling or adding similar end-to-end tests in the future.

### Request: 
Please investigate and address device/context handling in the TTIR builder pipeline to support optimizer and sharding with `"enable-optimizer=true, memory-layout-analysis-enabled=true"` set. This will unblock the affected tests and enable robust, end-to-end testing with optimizer and sharding enabled.

rpavlovicTT:
This should be revisited when mock device is integrated with tt-mlir (https://github.com/tenstorrent/tt-mlir/issues/6719)

bmalesevicTT:
@rpavlovicTT Did you check if TTIR builder works for optimizer passes now that mock device is used, to close this issue?

Issue #5283, opened 2025-10-08
Optimizer `test_conv2d_sharding` Fails In CI


ctodTT:
### Describe the bug

The optimizer `test_conv2d_sharding` test hits a segfault while running the ttir pipeline in CI. For the moment, the test has been disabled to get a more pressing PR merged. 

### Steps to reproduce the issue

See [this run](https://github.com/tenstorrent/tt-mlir/actions/runs/18354674740/job/52286718345?pr=5274) for more information

PR #6010, opened 2025-11-24
Skip EIO when no ops are flattened

odjuricicTT:
### Ticket
No ticket

### Problem description
The EraseInverseOps (EIO) pass should only run on functions that have at least one op that has been explicitly flattened (usually adding permute + reshape for pooling op to get (N, C, H, W) to (1, 1, N*H*W, C)) Currently, it runs on all functions regardless of whether they contain flattened operations, which may not be necessary and will just commute some of the TMs without erasing any, making it harder to have consistant fusing patterns and op locations in TTNN.

### What's changed
- Added `hasFlattenedCompatInfoAttr()` helper function that generically checks for the presence of `TTIR_FlattenedCompatInfoAttr` on any operation within a funcOp
- Modified the pass to skip funcOps that don't contain any operation with the flattened_compat_info attribute
- Added a `force` flag (default: false) to the pass options to override this behavior and always run the pass, maintaining backward compatibility
- Updated all 16 existing EraseInverseOps tests to use `force=true` to maintain their original behavior
- Added 2 new tests:
  - `skip_without_flattened_attr.mlir`: Verifies the pass is skipped when no attribute exists
  - `run_with_flattened_attr.mlir`: Verifies the pass runs and erases inverse TMs when the attribute exists
- Updated TMFusion test to use `force=true` flag
- Removed 2 bf_interleaved_policy tests that were sensitive to optimizer behavior changes

### Checklist
- [x] New/Existing tests provide coverage for changes

PR #6018, opened 2025-11-25
Revert rope back to ttnn and introduce op constraints check for fusing

### LINKED ISSUE #6004: Phi models are failing with TT_FATAL: Input X dimension (32) must be divisible by 64 for tiling in TT-XLA
Ref : https://github.com/tenstorrent/tt-xla/issues/2290

Most of the Phi models are failing in nightly with the below error

`TT_FATAL: Input X dimension (32) must be divisible by 64 for tiling `

When bisected, it is observed that https://github.com/tenstorrent/tt-mlir/commit/de4bf473d51124392422dda751875fb9b1279ae9 is responsible for this error.

Tagging @mtopalovicTT as he is the author of this commit to fix this issue

mtopalovicTT:
closes #6004

This PR reverts fusing of rope to TTNN. This is simply because there are a lot of constraints in metal regarding rope fusing which simply cannot be checked in fusing itself. Instead we propose to use op model get constraints api to check if rope can be fused. This has some side effects:

* Rope can only be fused with optimizer on
* Rope tests now require op model which cases golden tests to fail because of https://github.com/tenstorrent/tt-mlir/issues/5283 https://github.com/tenstorrent/tt-mlir/issues/5909. Problem is that golden test are opening the device by default and caching it for all the tests that are being executed, but get constraints api also needs to open the device and asserts because device is already opened.

### DIFF
--- a/lib/Dialect/TTNN/Pipelines/TTNNPipelines.cpp
+++ b/lib/Dialect/TTNN/Pipelines/TTNNPipelines.cpp
@@ -13,7 +13,7 @@
 #include "ttmlir/Dialect/TTCore/Utils/PopulateArgumentTypes.h"
 #include "ttmlir/Dialect/TTIR/Pipelines/TTIRPipelines.h"
 #include "ttmlir/Dialect/TTIR/Transforms/Passes.h"
-#include "ttmlir/Dialect/TTNN/Transforms/DevicePassWrapper.h"
+#include "ttmlir/Dialect/TTNN/Transforms/DevicePassesWrapper.h"
 #include "ttmlir/Dialect/TTNN/Transforms/Passes.h"
 #include "ttmlir/Support/Logger.h"
 #include "ttmlir/Transforms/Passes.h"
@@ -89,14 +89,14 @@ void createTTNNPipelineAnalysisPasses(
 #ifdef TTMLIR_ENABLE_OPMODEL
     ttnn::TTNNOptimizerOptions optimizerOptions(options);
     // Wrap all Optimizer passes with device lifecycle management.
-    DevicePassWrapperOptions wrapperOptions;
+    DevicePassesWrapperOptions wrapperOptions;
     wrapperOptions.devicePtr = options.devicePtr;
     wrapperOptions.tensorL1UsageCap = options.tensorL1UsageCap;
 
     ttnn::TTNNOperationValidationAndFallbackOptions validationOptions{
         options.tensorL1UsageCap};
 
-    pm.addPass(createDevicePassWrapper(
+    pm.addPass(createDevicePassesWrapper(
         [optimizerOptions, validationOptions](OpPassManager &innerPm) {
           // All Optimizer passes will be run inside the wrapper.
           innerPm.addPass(
@@ -139,11 +139,11 @@ void createTTNNFusingPass(OpPassManager &pm,
   if (options.enableFusing) {
     if (options.optimizerPassEnabled) {
 #ifdef TTMLIR_ENABLE_OPMODEL
-      DevicePassWrapperOptions wrapperOptions;
+      DevicePassesWrapperOptions wrapperOptions;
       wrapperOptions.devicePtr = options.devicePtr;
       wrapperOptions.tensorL1UsageCap = options.tensorL1UsageCap;
 
-      pm.addPass(createDevicePassWrapper(
+      pm.addPass(createDevicePassesWrapper(
           [](OpPassManager &innerPm) {
             TTNNFusingOptions fusingOptions;
             fusingOptions.enableOpConstraints = true;
--- a/include/ttmlir/Dialect/TTNN/Transforms/DevicePassesWrapper.h
+++ b/include/ttmlir/Dialect/TTNN/Transforms/DevicePassesWrapper.h
@@ -2,8 +2,8 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 
-#ifndef TTMLIR_DIALECT_TTNN_TRANSFORMS_DEVICEPASSWRAPPER_H
-#define TTMLIR_DIALECT_TTNN_TRANSFORMS_DEVICEPASSWRAPPER_H
+#ifndef TTMLIR_DIALECT_TTNN_TRANSFORMS_DEVICEPASSESWRAPPER_H
+#define TTMLIR_DIALECT_TTNN_TRANSFORMS_DEVICEPASSESWRAPPER_H
 
 #include "mlir/Pass/Pass.h"
 
@@ -16,8 +16,8 @@ class MeshDevice;
 namespace mlir::tt::ttnn {
 
 #ifdef TTMLIR_ENABLE_OPMODEL
-// Options for DevicePassWrapper pass
-struct DevicePassWrapperOptions {
+// Options for DevicePassesWrapper pass
+struct DevicePassesWrapperOptions {
   // External device pointer (if provided by frontend)
   std::shared_ptr<::tt::tt_metal::distributed::MeshDevice> devicePtr = nullptr;
   // Tensor L1 usage cap (fraction of available L1 memory)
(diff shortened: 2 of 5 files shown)


lib/Dialect/TTNN/Transforms/TTNNFusing.cpp
```diff
@@ -162,6 +167,275 @@ class TTNNMatmulAndLinearWithActivation : public mlir::OpRewritePattern<SrcOp> {
...
+// rotate_half([x1, x2, x3, x4]) = [-x3, -x4, x1, x2]
+//
+// This pattern is broken down into two sub-patterns:
+// 1. unrotatedProjection: matches x * cos
+// 2. rotatedProjection: matches rotate_half(x) * sin
+//
+// unrotatedProjection comes as following sequence of operations:
+// cos_unsqueezed = unsqueeze(cos) - where unsqueeze will be reshape which
+// prepends
+//  dimensions to cos to match x's rank.
```
rpavlovicTT: merge rows

lib/Dialect/TTNN/Transforms/TTNNFusing.cpp
```diff
@@ -162,6 +167,275 @@ class TTNNMatmulAndLinearWithActivation : public mlir::OpRewritePattern<SrcOp> {
...
+// unrotated_projection = multiply(x, cos_unsqueezed)
+//
+// rotatedProjections comes as following sequence of operations:
+// sin_unsqueezed = unsqueeze(sin) - where unsqueeze will be reshape which
+// prepends
+//  dimensions to sin to match x's rank.
```
rpavlovicTT: ditto

lib/Dialect/TTNN/Transforms/TTNNFusing.cpp
```diff
@@ -162,6 +167,275 @@ class TTNNMatmulAndLinearWithActivation : public mlir::OpRewritePattern<SrcOp> {
...
+    // Extract input layouts from the operation
+    std::vector<TTNNLayoutAttr> inputLayouts =
+        utils::extractInputLayouts(ropeOp.getOperation());
+
+    OpConfig config(mlir::cast<TTNNLayoutAttr>(resultType.getEncoding()));
+
+    auto backend = mlir::cast<OpModel>(ropeOp.getOperation());
+
+    llvm::Expected<ttnn::op_model::OpConstraints> constraintsExp =
+        backend.getOpConstraints(inputLayouts, config);
```
rpavlovicTT: You included OpConstraintValidation above, but did not use it here. Any specific reason?

jameszianxuTT:
cc @ddilbazTT  @jazpurTT 

Will retrigger uplift to get this into tt-xla hopefully before nightly if this can be merged.

test/python/golden/optimizer/test_conv_sharding.py
```diff
@@ -58,6 +58,8 @@ def test_conv2d_sharding(
     request,
     device,
 ):
+    print(device)
```
jserbedzijaTT: I guess this is not needed

mtopalovicTT:
> Looks good.
> 
> To confirm that with this PR and #6010 the microsoft phi regression will be solved? #6004

Even without eio fix it will work.

sdjordjevicTT (approved):

lib/Dialect/TTNN/Transforms/TTNNFusing.cpp
```diff
@@ -162,6 +167,273 @@ class TTNNMatmulAndLinearWithActivation : public mlir::OpRewritePattern<SrcOp> {
...
+
+    rewriter.replaceOp(srcOp, ropeOp.getResult());
+    return mlir::success();
+#else
+    return failure();
+#endif // TTMLIR_ENABLE_OPMODEL
+  }
+
+  bool matchUnrotatedProjection(MultiplyOp mulOp, Value &x,
+                                Value &cosUnsqueezed) const {
```
rpavlovicTT, line 307: Just suggestion, you don't have to apply it: in optimizer I used llvm::Expected for cases like this when you want to know if function failed/succeeded and returns some data along. In your example, it would be
```
llvm::Expected<pair<Value, Value>> matchUnrotatedProjection(MultiplyOp mulOp)
```
and you don't have to pass by ref x, cosUnsqueezed.

lib/Dialect/TTNN/Transforms/TTNNFusing.cpp
```diff
@@ -162,6 +167,273 @@ class TTNNMatmulAndLinearWithActivation : public mlir::OpRewritePattern<SrcOp> {
...
+    if (!inputType || inputType.getRank() < 1) {
+      return failure();
+    }
+
+    int64_t headDim = inputType.getShape()[inputType.getRank() - 1];
+    if (headDim <= 0 || headDim % 2 != 0) {
+      return failure();
+    }
+
+#ifdef TTMLIR_ENABLE_OPMODEL
```
rpavlovicTT: can whole `RoPEFusing` be enclosed by  `#ifdef TTMLIR_ENABLE_OPMODEL`?

lib/Dialect/TTNN/Pipelines/TTNNPipelines.cpp
```diff
@@ -128,6 +128,38 @@ void createTTNNPipelineLoweringPasses(
...
+void createTTNNFusingPass(OpPassManager &pm,
+                          const TTIRToTTNNBackendPipelineOptions &options) {
+  if (options.enableFusing) {
+    if (options.optimizerPassEnabled) {
+#ifdef TTMLIR_ENABLE_OPMODEL
+      OptimizerPassesWrapperOptions wrapperOptions;
+      wrapperOptions.devicePtr = options.devicePtr;
+      wrapperOptions.tensorL1UsageCap = options.tensorL1UsageCap;
+
+      pm.addPass(createOptimizerPassesWrapper(
```
rpavlovicTT: We can maybe consider renaming this wrapper since now it wraps not only "optimizer" passes, but fusing too. Although, fusing may be considered an optimisation.

lib/Dialect/TTNN/Transforms/TTNNFusing.cpp
mtopalovicTT: I fixed it in latest commit.

lib/Dialect/TTNN/Transforms/TTNNFusing.cpp
mtopalovicTT: I don't mind either but we don't have that practice in dialect codebase. Lets for now stick to this :D

lib/Dialect/TTNN/Transforms/TTNNFusing.cpp
mtopalovicTT: Yes

lib/Dialect/TTNN/Pipelines/TTNNPipelines.cpp
mtopalovicTT: createExternalDeviceWrapper?

rpavlovicTT: maybe `createDeviceTouchingPassesWrapper`? since it's not exclusively external device.

mtopalovicTT: Hmmm I'm not sure about that one, maybe drop Touching?

include/ttmlir/Dialect/TTNN/Transforms/DevicePassWrapper.h
```diff
@@ -35,19 +36,19 @@ struct OptimizerPassesWrapperOptions {
...
 //       [](OpPassManager &innerPm) {
 //         innerPm.addPass(createTTNNOptimizer(...));
 //         innerPm.addPass(createTTNNOperationValidationAndFallback(...));
 //       },
 //       options));
-std::unique_ptr<Pass> createOptimizerPassesWrapper(
-    std::function<void(OpPassManager &)> populatePipeline,
-    const OptimizerPassesWrapperOptions &options = {});
+std::unique_ptr<Pass>
+createDevicePassWrapper(std::function<void(OpPassManager &)> populatePipeline,
```
rpavlovicTT: This is also fine, I'd just keep plural DevicePassesWrapper

mtopalovicTT: Done

Issue #4288, opened 2025-07-31
[ OpModel ] Avoid equality comparison in opModelTests


sgholamiTT:
Change in `cbSize` (or `peakSize` etc) is common.
To avoid many changes in op model tests (such as `TestOpModelLib`) we should avoid exact comparisons and perform sth like `ASSERT_GT(..,0)` or `ASSERT_GT(...,1024)` etc.

Some CI run that fails due to this issue: https://github.com/tenstorrent/tt-mlir/actions/runs/16627177012/job/47052344107

Also, @jameszianxuTT mentioned sth odd that I quote here. This worth investigation.

> Caveat: Bisection is done by local execution of the relevant opmodel test group.
I see local [ FAILED ] in stdout for logicalAndTests with tt-metal commit pointing to 662a1a3  but not the commit before, which is why I attribute cause of failure to 662a1a3 .
However, running this test locally even when it shows the [ FAILED ]  string does not show failures in the final test report (shows all tests passed), and return code is 0. I am not familiar with this test infra, but I think this is a legitimate bisection. Someone who knows these tests better maybe can comment

-----
Error message in the CI run:

```
[  FAILED  ] LogicalAndTests/OpModelLogicalAndParam.LogicalAndOp/7, where GetParam() = 464-byte object <70-04 EA-9A CE-55 00-00 02-00 00-00 06-00 00-00 40-00 00-00 00-00 00-00 00-04 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 ... 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 01-00 00-00 00-00 00-00 00-30 00-00 00-00 00-00 00-08 00-00 00-00 00-00 00-08 00-00 00-00 00-00> (373 ms)
/localdev/jameszianxu/tt-mlir/test/unittests/OpModel/TTNN/Lib/TestOpModelLib.cpp:822: Failure
Expected equality of these values:
  cbSize
    Which is: 16384
  expectedCbSize
    Which is: 12288 
```

sgholamiTT:
FYI @rpavlovicTT @gfengTT @ckaravasilisTT @azecevicTT @arminaleTT @odjuricicTT.

ckaravasilisTT:
[...]

PR #6024, opened 2025-11-25
Relax OpModel equality tests

ckaravasilisTT:
### Ticket
#4288

### Problem description
OpModel unit tests have been using equality checks for op memory usage. This is redundant as we constantly change these numbers during metal uplifts.

### What changed
Replaced `EXPECT_EQ` with `EXPECT_GT` requiring memory usage to be > 0 in `TestOpModelLib.cpp` and `TestOpModelInterface.cpp`. Equality checks for `0` mem usage, boolean values, and layout checks remain unchanged. Prexisting `EXPECT_GE` were also left unchanged. One can refactor this further to shorten tests, but we may want to keep some of the infrastructure (e.g. `ExpectedResult` struct) to easily reintroduce other checks in the future.

### DIFF
--- a/test/unittests/OpModel/TTNN/Op/TestOpModelInterface.cpp
+++ b/test/unittests/OpModel/TTNN/Op/TestOpModelInterface.cpp
@@ -1865,7 +1865,7 @@ TEST_F(OpModelBase, SplitQueryKeyValueAndSplitHeadsOpInterface) {
     auto l1 = constraintsExp.get();
     const auto [cbSize, l1PeakSize, totalPeakSize, outputSize, outputLayout] =
         l1;
-    EXPECT_EQ(cbSize, 0);
+    EXPECT_GT(cbSize, 0);
     EXPECT_GE(l1PeakSize, 0);
     EXPECT_GT(outputSize, 0);
     EXPECT_GE(totalPeakSize, 0);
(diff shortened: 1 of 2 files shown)


arminaleTT (approved):
Thanks for doing this!

gfengTT:
Shall we use `EXPECT_GT` instead of `EXPECT_GE`? That would better differentiate tensors in DRAM (usage = 0) and tensors in L1 (usage > 0).

ckaravasilisTT:
Yes I agree @gfengTT . I will do this shortly.

sgholamiTT (approved):
Thanks for doing this! 👍🏻 
This was needed for sure.

test/unittests/OpModel/TTNN/Op/TestOpModelInterface.cpp
```diff
@@ -2980,9 +2960,9 @@ TEST_F(OpModelBase, PrepareConv2dWeightsTest) {
   ASSERT_TRUE(static_cast<bool>(constraintsExp));
   const auto &[cbSize, l1PeakSize, totalPeakSize, outputSize, outputLayout] =
       constraintsExp.get();
-  EXPECT_EQ(cbSize, 0);
```
rpavlovicTT, line 2983: Can you skip changing equalities with 0?

ckaravasilisTT: I've updated the tests with these tighter checks. We can later discuss if we should relax them further.

Issue #5859, opened 2025-11-14
Token index is sent to host from device to change layout


odjuricicTT:
In llama 3.2 1B (link)
We have:
```mlir
        %40 = "ttnn.typecast"(%arg3) <{dtype = #ttcore.supportedDataTypes<u32>}> : (tensor<32x1xsi32, #ttnn_layout14>) -> tensor<32x1xui32, #ttnn_layout21> loc(#loc219)
        "ttnn.deallocate"(%arg3) <{force = false}> : (tensor<32x1xsi32, #ttnn_layout14>) -> () loc(#loc219)
        %41 = "ttnn.reshape"(%40) <{shape = [1 : i32, 32 : i32]}> : (tensor<32x1xui32, #ttnn_layout21>) -> tensor<1x32xui32, #ttnn_layout21> loc(#loc219)
        "ttnn.deallocate"(%40) <{force = false}> : (tensor<32x1xui32, #ttnn_layout21>) -> () loc(#loc219)
        %42 = "ttnn.from_device"(%41) : (tensor<1x32xui32, #ttnn_layout21>) -> tensor<1x32xui32, #ttnn_layout22> loc(#loc1320)
        "ttnn.deallocate"(%41) <{force = false}> : (tensor<1x32xui32, #ttnn_layout21>) -> () loc(#loc1320)
        %43 = "ttnn.to_layout"(%42) <{layout = #ttnn.layout<row_major>}> : (tensor<1x32xui32, #ttnn_layout22>) -> tensor<1x32xui32, #ttnn_layout23> loc(#loc1320)
        "ttnn.deallocate"(%42) <{force = false}> : (tensor<1x32xui32, #ttnn_layout22>) -> () loc(#loc1320)
        %44 = "ttnn.to_device"(%43, %39) <{memory_config = #ttnn.memory_config<#dram, <interleaved>>}> : (tensor<1x32xui32, #ttnn_layout23>, !ttnn.device) -> tensor<1x32xui32, #ttnn_layout24> loc(#loc1320)
        "ttnn.deallocate"(%43) <{force = false}> : (tensor<1x32xui32, #ttnn_layout23>) -> () loc(#loc1320)
        %45 = "ttnn.embedding"(%44, %arg4) : (tensor<1x32xui32, #ttnn_layout24>, tensor<128256x2048xbf16, #ttnn_layout15>) -> tensor<1x32x2048xbf16, #ttnn_layout25> loc(#loc220)
```

[...]

Issue #5745, opened 2025-11-07
Fix hard-coded `uint32` expectation in runtime/update_cache.cpp


rpavlovicTT:
### Describe the bug

```cpp
    const ::tt::tt_metal::HostBuffer buffer =
        ::tt::tt_metal::host_buffer::get_host_buffer(indexOnHost);
    const auto &buf = buffer.view_as<uint32_t>();
    uint32_t upIdx = *buf.begin();
```

In runtime's update_cache.cpp we view `updateIndex`'s buffer as uint32 buffer. This may lead to an error if this tensor is created with signed int32 type for example. Error is:
```cpp
    TT_FATAL(*type_info_ == typeid(T), "Requested type T does not match the underlying buffer type.");
```

I think we should check what is underlying tensor's dtype and decide which view we want based on that. 

For the targeting `::ttnn::update_cache` op it doesn't really matter which kind of int we pass through as it will cast it to uint32_t anyway. 

```cpp
struct UpdateCacheOperation {
    static ttnn::Tensor invoke(
        const ttnn::Tensor& cache,
        const ttnn::Tensor& input,
        uint32_t update_idx,
        uint32_t batch_offset,
        std::optional<const DeviceComputeKernelConfig> compute_kernel_config = std::nullopt);
};
```

To avoid errors during runtime, in op model lib I made a workaround fix (not ideal)

https://github.com/tenstorrent/tt-mlir/blob/110a0c76665fb7b80aee0fc8e79b9e311054c869/lib/OpModel/TTNN/TTNNOpModel.cpp#L3972-L3974

PR #6029, opened 2025-11-25
Canonicalize `ttir.update_cache` to `ttir.paged_update_cache`

LPanosTT:
### Ticket
https://github.com/tenstorrent/tt-mlir/issues/5745
https://github.com/tenstorrent/tt-mlir/issues/5859

### Problem description
`ttnn.update_cache` requires that all users have the same cache position. This forces implementations in torch to use left padding. In addition to that the `update_idx` is a `uint32_t` arg on the ttnn op, not a tensor. Since we model the op using an integer tensor to carry this value, we must workaround this by **pulling the tensor off device and reading the integer from the buffer**. This is not ideal.
### What's changed
`ttnn.paged_update_cache` is also capable of updating a static cache. This op allows us to provide a different cache position (update_idx) for each user. This op also accepts an integer tensor as an argument for this, unlike `ttnn.update_cache`

Since `ttnn.paged_update_cache` has broader functionality, this PR adds a canonicalization to `ttir.update_cache` to transform it into a `ttnn.paged_update_cache`, utilizing its static-cache ability.

`ttnn.paged_update_cache` requires its fill value to be in the form `[1, num_users, num_heads, head_size]` unlike `ttnn.update_cache` which requires `[1, num_heads, num_users, head_size]` so we must insert a permute on this operand. It seems as though a permute is already in place on this operand for most models we compile from `transformers`, so this extra permute just gets folded away and we end up with one permute regardless.

If each user does indeed have the same cache position (i.e. the inputs are left-padded). A single cache position integer tensor will not suffice. `ttnn.paged_update_cache` _does_ require a cache position tensor of shape `[batch_size]` so, this tensor must be full of identical integers.

For some models in `transformers` left padding is simply required, and issues will arise if you attempt to provide a cache position which is not a singleton tensor. For this reason, as part of the canonicalization we will conditionally insert a repeat op on the cache position if it is a single integer.

### Checklist


odjuricicTT (approved):

lib/Dialect/TTIR/IR/TTIROps.cpp
```diff
@@ -3732,6 +3732,53 @@ ::mlir::LogicalResult mlir::tt::ttir::UpdateCacheOp::verify() {
...
+
+    // If the update index shape is [1] then repeat to num users
+    TypedValue<RankedTensorType> newUpdateIndex = op.getUpdateIndex();
+    if (updateIndexShape[0] == 1) {
+      auto newUpdateIndexShape = {numUsers};
+      auto newUpdateIndexType = RankedTensorType::get(
+          newUpdateIndexShape, newUpdateIndex.getType().getElementType(),
+          newUpdateIndex.getType().getEncoding());
+      auto repeatDims = rewriter.getDenseI64ArrayAttr({numUsers});
+      newUpdateIndex = utils::createDPSOp<RepeatOp>(rewriter, op.getLoc(),
```
rpavlovicTT: We recently got rid of DPS ops, so please rebase and adjust code.

Issue #5929, opened 2025-11-18
[D2M][TTNN][JIT][Graph Capture] Input arguments to ttnn ops not wired correctly when using graph capture instead of AST


mbagherbeikTT:
### Describe the bug

See title

### Test Case
- test/ttnn-jit/multiply_accumulate_trace.py::test_muladd_dram_compare[256-256]
- other sizes fail too

### TTNN IRs outputted 

- IR generated via AST runs on device and passes PCC
- IR generated via GC fails PCC when running on device, output is grossly different from golden

```
---- After TTIRCompiler (AST-based) ----
#dram = #ttnn.buffer_type<dram>
#ttnn_layout = #ttnn.ttnn_layout<(d0, d1) -> (d0, d1), <1x1>, memref<8x8x!ttcore.tile<32x32, bf16>, #dram>, <interleaved>>
module {
  func.func @muladd(%arg0: tensor<256x256xbf16, #ttnn_layout>, %arg1: tensor<256x256xbf16, #ttnn_layout>, %arg2: tensor<256x256xbf16, #ttnn_layout>) -> tensor<256x256xbf16, #ttnn_layout> {
    %0 = "ttnn.get_device"() <{mesh_offset = #ttnn<mesh_offset 0x0>, mesh_shape = #ttnn<mesh_shape 1x1>}> : () -> !ttnn.device
    %1 = "ttnn.multiply"(%arg1, %arg2) <{dtype = #ttcore.supportedDataTypes<bf16>}> {ttnn.hoist_generic_via_d2m} : (tensor<256x256xbf16, #ttnn_layout>, tensor<256x256xbf16, #ttnn_layout>) -> tensor<256x256xbf16, #ttnn_layout>
    %2 = "ttnn.add"(%1, %arg0) <{dtype = #ttcore.supportedDataTypes<bf16>}> {ttnn.hoist_generic_via_d2m} : (tensor<256x256xbf16, #ttnn_layout>, tensor<256x256xbf16, #ttnn_layout>) -> tensor<256x256xbf16, #ttnn_layout>
    return %2 : tensor<256x256xbf16, #ttnn_layout>
  }
}

---- After GraphToIRTranslator (Graph-based) ----
#dram = #ttnn.buffer_type<dram>
#ttnn_layout = #ttnn.ttnn_layout<(d0, d1) -> (d0, d1), <1x1>, memref<8x8x!ttcore.tile<32x32, bf16>, #dram>, <interleaved>>
module {
[...]

Issue #5873, opened 2025-11-14
[ttnn.jit] Multiply Accumulate fails All Close Check


brapananTT:
The JIT'ed multiply accumulate composite op:

```
def mul_add(input_tensor_a, input_tensor_b, input_tensor_c):
    mul_result = ttnn.multiply(input_tensor_b, input_tensor_c)
    return ttnn.add(mul_result, input_tensor_a)
```

 fails all close and pcc checks in `test/ttnn-jit/test_eltwise_composite.py`.

This was discovered after removing `torch.manual_seed(0)` from helper funcs `create_dram_tensor` and `create_sharded_tile_tensor` in `test/ttnn-jit/utils.py`.

brapananTT:
Reason for pcc / allclose failure discovered and tracked here: #5929

PR #6035, opened 2025-11-25
[D2M][TTNN][JIT][Graph Capture] Add Support for Levelized Graph, Reduction Ops, and Composite Ops

sgholamiTT:
### Ticket
#5929

### Problem description
This PR significantly enhances the `TTNN-JIT` compiler by adding support for:
- **Levelized graph traversal** for handling composite operations
- **Reduction operations** (`sum`, `mean`, `max`, `min`)
- **Composite operations** that decompose into multiple primitive operations (`digamma`)
- **Improved architecture** through operation registry pattern and modular design

### Motivation
The existing `TTNN-JIT` compiler had several limitations:
1. **No support for composite operations**: Operations like `digamma`, etc. that internally decompose into multiple primitive operations couldn't be compiled
2. **Missing reduction operations**: Common operations like `sum`, `mean`, `max`, `min` were not supported (due to lack of support for dynamic output shape inference)
3. **Rigid architecture**: Operation handling logic was scattered and difficult to extend
4. **Incomplete graph capture**: Final outputs lacked proper metadata for correct compilation

###  Architecture Changes
architecture_comparison

#### Levelized Graph Support (`levelized_graph.py`) (See the [metal PR](https://github.com/tenstorrent/tt-metal/pull/32676) for full details).
levelized_graph_expansion

Added comprehensive data structures for working with levelized graphs extracted from TTNN operation traces:
Key Components:
`LevelizedGraphVertex`: Wrapper class for graph vertices with helper methods
- Distinguishes between input tensors and operations
- Provides methods to extract tensor IDs and operation names
- Checks for supported operations and composite operations
- Identifies placeholder operations and final outputs

`LevelizedGraph`: High-level graph operations manager
- Extracts and orders tensor arguments
- Performs walks on the graph,
- Finds output vertices
- Validates graph visitability

#### Operation Registry Pattern (`op_registry.py`)

Introduced a flexible operation registry system that makes the compiler more generic:

Handler Interface (`OpHandler`):

```python
class OpHandler(ABC):
    def parse_arguments(op_name, arguments) -> Dict
    def validate_operands(op_name, operands) -> None
    def prepare_result_type(op_name, result_type, vertex, compiler_context) -> Type
    def prepare_operands(op_name, operands, parsed_args, result_type, device, compiler_context) -> List
    def create_operation(op_name, result_type, operands, device, parsed_args, ctx) -> Operation
    def add_common_attributes(result, ctx) -> None
```
Implemented Handlers:
- `UnaryOpHandler`: Single input operations (exp, log, sqrt, etc.)
- `BinaryOpHandler`: Two input operations with optional scalar constants
- `ReductionOpHandler`: Dimension reduction operations (sum, mean, max, min)
- `SquareOpHandler`: Special case that maps to multiply(x, x) (there's no `ttnn.square` for some reason!)
handler_flow

#### Return Modifier (`return_modifier.py`)
Solves a critical graph capture limitation where final outputs lack metadata (this is workaround, not a permanent solution):
Problem:
```python
def some_function(input_tensor):
    last_tensor = ttnn.some_op(...)
    return last_tensor  # ❌ Output metadata not captured
```
Solution:
```python
def some_function(input_tensor):
    last_tensor = ttnn.some_op(...)
    ttnn.identity(last_tensor)  # ✅ Placeholder marks output
    return last_tensor
```
- AST transformation to inject placeholder operations before returns
- Supports single and multiple return values
- Handles both simple variables and complex expressions
- Uses `ttnn.identity` as a passthrough placeholder
- Metal issue that should be solved to address this: https://github.com/tenstorrent/tt-metal/issues/33914

### Testing
Composite Operations (`test_composite.py`)
- test_digamma_dram: Special mathematical function
- test_complex_composite_dram: Multi-step composite operations
- test_nested_composite_dram: Nested composite operations

Reduction Operations (`test_reduction.py`)
- test_sum: Sum reduction with various dimensions and keepdim
- test_mean: Mean reduction
- test_max: Max reduction
- test_min: Min reduction

Composite Elementwise (`test_eltwise_composite.py`)
- test_composite_ops_l1: L1 sharded memory tests
- test_composite_ops_dram: DRAM interleaved tests
- test_large_shapes_muladd_l1/dram: Large shape tests for mul_add
Note: Reduction operations are currently marked as `skip` (not `xfail`) because `D2M` does not yet support them, which causes assertion crashes during compilation.

#### Notes:
- Composite operation expansion uses an iterative approach with a maximum depth of 10 to prevent infinite loops while allowing reasonable nesting.
- Operation registry is designed to be extensible. Future additions (e.g., ternary operations, matrix operations) can be easily added by implementing the OpHandler interface.
- Return modifier is a temporary workaround until graph capture properly handles final outputs or TTNN adds a dedicated no-op.

### Checklist


tools/ttnn-jit/_src/return_modifier.py
```diff
@@ -0,0 +1,207 @@
...
+"""
+There's a known limitation in graph capture that it does not populate the output tensor shape and info for ops without a consumer.
+For instance, imagine a function f like this:
+def someFunction(input_tensor):
+    # ... some functionality
+    last_tensor = ttnn.someOp(...)
+    return last_tensor
+In this case, the output tensor shape and info for ttnn.someOp(...) is not populated.
+To this aim, the following class inserts ttnn.identity calls before the return statements.
+So we canvert the aforemention function to this:
```
arminaleTT: "canvert" typo?

tools/ttnn-jit/_src/return_modifier.py
```diff
@@ -0,0 +1,207 @@
...
+def someFunction(input_tensor):
+    # ... some functionality
+    last_tensor = ttnn.someOp(...)
+    ttnn.identity(last_tensor)
+    return last_tensor
+
+Therefore, we can correctly capture the output tensor shape and info for ttnn.someOp(...).
+This also helps in identifying the final output tensor of the function.
+Note 1: that this code also supports multiple return values and expressions.
+Note 2: This is a temporary workaround for this problem. In order to fix it, we should either:
```
arminaleTT, line 28: If this is a temporary workaround, please create an issue, add these details there, and include the issue number in this comment. The issue doesn't necessarily need to be assigned to you, but without an issue this "temporary" workaround will become permanent

tools/ttnn-jit/_src/return_modifier.py
```diff
@@ -0,0 +1,207 @@
...
+- Add support for capturing the output tensor shape and info for ops without a consumer.
+- Add a ttnn.no-op as a pass through op to ttnn.
+Note 3: We assert that the original function does not contain any ttnn.identity calls.
+Note 4: This approach should work for any passthrough op. Rn, we use ttnn.identity as an example.
+"""
+
+# Placeholder op is used to mark the output of a function without performing any computation.
+# One example of a placeholder op is "identity" (ttnn.identity), but other passthrough
+# operations could also serve this purpose.
+placeholder_op_name = "ttnn.identity"
```
arminaleTT: nit: should be all caps

tools/ttnn-jit/_src/return_modifier.py
```diff
@@ -0,0 +1,207 @@
...
+
+def create_modified_function(f):
+    """
+    Create a modified version of function f that inserts passthrough calls before returns.
+    """
+    # Get the source code of the function
+    source = inspect.getsource(f)
+
+    # Assert that the original function does not contain any placeholder op calls
+    assert placeholder_op_name not in source, (
```
arminaleTT: Should we have a list of eligible placeholder ops so we don't have to fail if one of them is actually present?

Also, is there a test for this?

tools/ttnn-jit/_src/op_registry.py
```diff
@@ -0,0 +1,533 @@
+# SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+
+"""
+Operation Registry for TTNN JIT Compiler.
```
arminaleTT, line 6: I really like this. The AST parser should be migrated to use this as well

tools/ttnn-jit/_src/graph_trace_compiler.py
```diff
@@ -2,753 +2,572 @@
...
+        # Note: Use .*? for non-greedy match since ShardSpec contains nested parens
+        shard_shape_match = re.search(
+            r"shard_spec=ShardSpec\(.*?shape=\{([^}]+)\}", output_info_str
+        )
+        if shard_shape_match:
+            shape_str = shard_shape_match.group(1)
+            config["shard_shape"] = [int(x.strip()) for x in shape_str.split(",")]
+
+        # Extract grid if sharded
+        # Format: grid={[(x=0,y=0) - (x=0,y=0)]}
```
arminaleTT, line 146: It is theoretically possible that there be multiple `CoreRange`s in the grid attribute. We certainly cannot support it in JIT/D2M, but erroring out in a user-friendly way would be nice

tools/ttnn-jit/_src/graph_trace_compiler.py
```diff
@@ -2,753 +2,572 @@
...
-        self.stacking_level = 0
-
-        if len(connections) != len(set(connections)):
-            print(
-                f"Warning: Duplicate connections found for vertex {counter}: {connections}"
-            )
+        Returns:
+            Dict with keys: buffer_type, memory_layout, shard_shape, grid
+        """
+        import re
```
arminaleTT, line 114: I'm not a fan of how many regexes are used here. I defer to your judgement if it is more readable with regexes vs using`starts_with`, `ends_with`, `find` etc.

tools/ttnn-jit/_src/graph_trace_compiler.py
```diff
@@ -2,753 +2,572 @@
...
+        import math
+
+        if memory_config["shard_shape"]:
+            # Use shard shape for sharded tensors
+            shard_h, shard_w = memory_config["shard_shape"]
+            tiles_h = math.ceil(shard_h / 32)
+            tiles_w = math.ceil(shard_w / 32)
+            memref_shape = [tiles_h, tiles_w]
+            affine_map = _get_collapsed_linear_affine_map(
+                self.ctx, shape, self.max_grid
```
arminaleTT: fyi rebase required. There's no `max_grid` anymore

tools/ttnn-jit/_src/graph_trace_compiler.py
xanderchin: Also not a fan of using regexs but in this case the match groups are further processed, so think it's good solution.

tools/ttnn-jit/_src/supported_ops.py
```diff
@@ -36,20 +37,44 @@
     "pow_tensor",
 ]
 
-# All supported operations
-all_ops = set(unary_ops + binary_ops)
+# Reduction operations - operations that reduce tensor dimensions
+reduction_ops = [
+    "sum",
+    "mean",
```
vtangTT, line 58: per our last standup idt we should support mean right?

test/ttnn-jit/test_reduction.py
```diff
@@ -0,0 +1,171 @@
...
+        ((4, 64, 64), 1),
+        ((8, 128, 128), 1),
+        ((4, 64, 64), 2),
+        ((8, 128, 128), 2),
+    ],
+)
+@pytest.mark.parametrize("keepdim", [True, False])
+@pytest.mark.parametrize("dtype", [torch.float32, torch.bfloat16])
+@pytest.mark.skip(
+    reason="D2M Does not support reduction ops - causes assertion crashes"
```
vtangTT: nit: can we put what the assertion is here, or just only say `D2M does not support reduction ops`

test/ttnn-jit/test_composite.py
```diff
@@ -0,0 +1,103 @@
+# SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
```
vtangTT: i feel like these tests can go into `test_eltwise_composite.py` still unless you think they're special enough to warrant another file

tools/ttnn-jit/_src/graph_trace_compiler.py
```diff
@@ -2,753 +2,572 @@
...
-    # Check if it's a tensor reference (e.g., "tensor: 0" or "tensor:0")
-    match = re.match(r"tensor:\s*(\d+)", arg)
-    if match:
-        tensor_ref = int(match.group(1))
-        return ParsedArgument(
-            arg_type="tensor_ref", raw_value=arg, parsed_value=tensor_ref
-        )
+    def _find_optimal_depth(self, max_depth: int = 10) -> Tuple[int, LevelizedGraph]:
+        """
+        Find the minimum depth k where all level 1 operations are visitable.
```
vtangTT, line 74: nit: can we clarify what a level 1 op is in this comment 🙏

tools/ttnn-jit/_src/op_registry.py
vtangTT: agreed

tools/ttnn-jit/_src/return_modifier.py
sgholamiTT: Good point. [Metal issue](https://github.com/tenstorrent/tt-metal/issues/33914) has been created and linked as a TODO.

tools/ttnn-jit/_src/return_modifier.py
sgholamiTT: > Should we have a list of eligible placeholder ops so we don't have to fail if one of them is actually present?

I'm not sure. Since: 
1. I don't think `identity` is an important op to support in jit for now.
2. I'm not aware of any other op that qualifies as a placeholder (the op has to have no side effects, no huge memory footprint, not computationally expensive, return it's original input, etc.). `identity` is the only one I found.
3. As you mentioned too, this is a temporary workaround which should be fixed in metal.

> Also, is there a test for this?
 
I added a test that expects to fail with an `identity` op (if that's what you mean).

arminaleTT: Makes sense! And yes, that's what I meant by test

tools/ttnn-jit/_src/graph_trace_compiler.py
sgholamiTT: Added assertions to error out if multiple `coreRange`s exist.

tools/ttnn-jit/_src/graph_trace_compiler.py
sgholamiTT: I kinda get your point @arminaleTT. This is not the best solution in my view too. But it's a bit difficult to do this with `starts_with`, `ends_with` etc. 
And also, as @xanderchin mentioned, we're using the groups later so it's not the worst thing to use `re` here I guess. 
Let me know if you think otherwise.

tools/ttnn-jit/_src/graph_trace_compiler.py
sgholamiTT: Just rebased.
Confirmed with @brapananTT too what kind of sanity check I should run to make sure that max_grid removal works as expected. Feels like the rebase is fine.
I'll ask Brandon for a review later too just to double check.

tools/ttnn-jit/_src/supported_ops.py
sgholamiTT: I think there is some misunderstanding about the term `supported`.
`Supported` does not mean that the `op` is supported e2e. It means that the jit frontend can generate IR for the op and pass it to D2M.
By having all those reductions here, I'm stating the fact that we're able to generate correct IR for those. That's how I enabled @ckaravasilisTT to work on some tests for reductions.
I get that this might not be the best name and it might confuse people. Maybe this + the support matrix that you have can clearly state what is supported in jit and what is supported e2e.

Let me know what you think.

arminaleTT (approved):

vtangTT (approved):

test/ttnn-jit/test_eltwise_composite.py
```diff
@@ -103,12 +103,7 @@ def mul_add(input_tensor_a, input_tensor_b, input_tensor_c):
 def test_composite_ops_l1(device, shape, max_grid, dtype, op, memory_layout):
     num_inputs = 1
     if op is mul_add:
-        # num_inputs = 3
-        pytest.xfail(
-            "mul_add fails allclose, see https://github.com/tenstorrent/tt-mlir/issues/5873"
```
brapananTT, line 108: Assuming these muladd tests are passing, you could also link #5873 to be closed by this PR if you'd like

PR #5967, opened 2025-11-20
Add memory state printing API and debug prints in submit

jnie-TT:
### Describe the problem
We are lacking device memory state profiling in runtime

### What's changed
Added a public debug API to print device memory state, added device memory state printing in submit guarded by `TT_RUNTIME_DEBUG`

Example:
```
2025-11-20 20:45:47,960 - INFO - set log file=
2025-11-20 20:45:47,960 - INFO - setting artifacts folder path=/localdev/jnie/tt-mlir/ttrt-artifacts
                 Always |    DEBUG | Device grid size = { 8, 8 }
2025-11-20 20:45:48,508 - INFO - PASS: getting system_desc passed
2025-11-20 20:45:48,508 - INFO - results saved to=query_results.json
                 Always |    DEBUG | Device grid size = { 8, 8 }
2025-11-20 20:45:48,733 - INFO - evaluating binary=./build/test/ttmlir/Silicon/TTNN/n300/runtime/Output/add.mlir.tmp.ttnn
                 Always |    DEBUG | Device memory state before submit
                 Always |    DEBUG | Device DRAM memory state: MemoryView{numBanks: 12, totalBytesPerBank: 1024.000 MB, totalBytesAllocatedPerBank: 0.008 MB, totalBytesFreePerBank: 1023.992 MB, largestContiguousBytesFreePerBank: 1023.992 MB}
                 Always |    DEBUG | Device L1 memory state: MemoryView{numBanks: 64, totalBytesPerBank: 1.270 MB, totalBytesAllocatedPerBank: 0.000 MB, totalBytesFreePerBank: 1.270 MB, largestContiguousBytesFreePerBank: 1.270 MB}
                 Always |    DEBUG | Device L1_SMALL memory state: MemoryView{numBanks: 64, totalBytesPerBank: 0.062 MB, totalBytesAllocatedPerBank: 0.000 MB, totalBytesFreePerBank: 0.062 MB, largestContiguousBytesFreePerBank: 0.062 MB}
                 Always |    DEBUG | Device TRACE memory state: MemoryView{numBanks: 12, totalBytesPerBank: 0.000 MB, totalBytesAllocatedPerBank: 0.000 MB, totalBytesFreePerBank: 0.000 MB, largestContiguousBytesFreePerBank: 0.000 MB}
            RuntimeTTNN |    DEBUG | Starting execution of program: add
[...]

PR #6037, opened 2025-11-26
Add memory log level and memory logs correspondingly

jnie-TT:
### Problem description
Currently the only memory debug logs are on the program level #5967, we want to support more comprehensive logging.

### What's changed
Added `MemoryLogLevel` bit flag that describes the memory logging level, and runtime API `setMemoryLogLevel` to configure the logging level. Currently there are 2 levels:
* Program: This will log at the beginning and end of program execution. Useful for debugging memory footprint across program runs.
* Operation: This will log at the beginning of each operation within the program. Useful for debugging memory footprint within a program run, e.g. OOM errors.

### Checklist

### DIFF
--- a/runtime/include/tt/runtime/types.h
+++ b/runtime/include/tt/runtime/types.h
@@ -12,8 +12,6 @@
 #include <optional>
 #include <vector>
 
-#include "tt/runtime/utils.h"
-
 #pragma clang diagnostic push
 #pragma clang diagnostic ignored "-Wcovered-switch-default"
 #include "ttmlir/Target/Common/debug_info_generated.h"
@@ -29,6 +27,7 @@ using ::tt::runtime::flatbuffer::DeviceRuntime;
 using ::tt::runtime::flatbuffer::DispatchCoreType;
 using ::tt::runtime::flatbuffer::FabricConfig;
 using ::tt::runtime::flatbuffer::HostRuntime;
+using ::tt::runtime::flatbuffer::MemoryLogLevel;
 
 /*
 MemoryBlockTable is a list of memory blocks in the following format:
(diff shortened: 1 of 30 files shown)


runtime/include/tt/runtime/types.h
```diff
@@ -180,14 +179,9 @@ struct TensorDesc {
              const ::tt::target::DataType dataType,
              const std::optional<uint32_t> itemsize = {},
              const std::optional<std::vector<uint32_t>> &stride = {},
-             const std::optional<uint64_t> physicalVolume = {})
-      : shape(shape), dataType(dataType) {
-    this->itemsize = itemsize.value_or(utils::dataTypeElementSize(dataType));
-    this->stride = stride.value_or(utils::calculateStride(shape));
-    this->physicalVolume = physicalVolume.value_or(volume());
-  }
+             const std::optional<uint64_t> physicalVolume = {});
```
kmabeeTT, line 182: Just curious, what is the deal with the TensorDesc changes here and in other file?

jnie-TT: Circular dependency, utils was including this and this was including utils, I moved the implementation to the cpp file instead of jamming everything in the header.

jameszianxuTT (approved):
LGTM! 

Side comment:
- Distributed response/request for these pure, void-returning configuration commands seems like it could be templated out a bit more, I will look into ways later to reduce the boilerplate needed. It seems that at least the worker response builder and maybe controller response handler could be factored out or codegenned or something in this trivial case

runtime/include/tt/runtime/types.h
kmabeeTT: Ahh got it, thx

dlokeTT (approved):

PR #6038, opened 2025-11-26
Removed ttrt dependency in builder and cleaned up runtime APIs

tapspatel:
Compiling a module gives back the mlir file and a golden dictionary

```
mlir_path, goldens = compile_ttir_module_to_flatbuffer(
            split_module,
            split_builder,
            system_desc_path=system_desc_path,
            test_base=test_base,
            output_root=output_root,
            target="ttnn",
            module_dump=True,
        )
```

this can be fed into the runtime API for golden evaluation
```
golden_report = execute_fb(flatbuffer_path, goldens, device=device)
```

A golden report is returned which has the following format for each op executed on device
```
{
    'loc("-":9:10)': {
            device_0: {
                'result': 'pass', 
                'expected_pcc': 0.99, 
                'actual_pcc': 1.0, 
                'expected_atol': 1e-08, 
                'actual_atol': 0.0, 
                'expected_rtol': 1e-05, 
                'actual_rtol': 0.0, 
                'allclose': True, 
                'max': 0.0, 
                'mean_absolute_error': 0.0, 
                'root_mean_square_error': 0.0, 
                'cosine_similarity': 1.0000001192092896
            }
        }
    }
```


tools/builder/base/builder_utils.py
```diff
@@ -525,11 +486,11 @@ def decorated_func(*inputs):
 
                 result = fn(*inputs, builder)
 
-                outputs = result if hasattr(result, "__iter__") else (result,)
+                outputs = result if hasattr(result, "__iter__") else [result]
                 output_goldens: Dict[Operand, GoldenMapTensor] = {}
                 for op in outputs:
                     output_goldens[op] = builder._get_golden_tensor(op)
-                builder._set_goldens(output_goldens)
+                # builder._set_goldens(output_goldens)
```
dlokeTT: Meant to comment this?

dlokeTT (approved):

hkwonTT:
It might be a dumb question, but which ttrt dependency did you remove? I skimmed the change quickly, and it looks like some implementations copied from ttrt were simply rewritten in another file.

tapspatel:
@hkwonTT we were using ttrt to access the underlying nano binds. Now we are directly accessing them without ttrt intervention. As well, some code from ttrt was migrated to allow execution which will be cleaned up. We are trying to avoid having ttrt in the loop in builder runtime.

Going forward, all this code will be standardized into a common lib between ttrt + builder runtime, but folks need stuff running right now, hence some duplicate code.

tools/builder/base/builder_utils.py
tapspatel: yeah it actually doesn't do anything since the output goldens are already stored in the map :) I should remove it completely

PR #6041, opened 2025-11-26
Implement const-eval CPU-hoisting

dmilinkovicTT:
### Ticket
#6098

### Problem description
This PR introduces possibility of executing ops which form the const-eval subgraphs on the CPU by modifying and extending existing passes for CPU hoisting. The feature can be toggled using `enable-cpu-hoisted-const-eval` option in the `ttir-to-ttnn-backend-pipeline` (for now, disabled by default).

#### Before

```mlir
func.func private @forward_merge_return_multiple_values_const_eval_0(%arg0: tensor<32x32xbf16, #ttnn_layout>, %arg1: tensor<32x32xbf16, #ttnn_layout>, %arg2: tensor<32x32xbf16, #ttnn_layout>) -> (tensor<32x32xbf16, #ttnn_layout>, tensor<32x32xbf16, #ttnn_layout>) attributes {const_eval} {
  %0 = "ttnn.add"(%arg0, %arg1) <{dtype = #ttcore.supportedDataTypes<bf16>}> : (tensor<32x32xbf16, #ttnn_layout>, tensor<32x32xbf16, #ttnn_layout>) -> tensor<32x32xbf16, #ttnn_layout>
  %1 = "ttnn.add"(%arg1, %arg2) <{dtype = #ttcore.supportedDataTypes<bf16>}> : (tensor<32x32xbf16, #ttnn_layout>, tensor<32x32xbf16, #ttnn_layout>) -> tensor<32x32xbf16, #ttnn_layout>
  %2 = "ttnn.subtract"(%0, %1) <{dtype = #ttcore.supportedDataTypes<bf16>}> : (tensor<32x32xbf16, #ttnn_layout>, tensor<32x32xbf16, #ttnn_layout>) -> tensor<32x32xbf16, #ttnn_layout>
  "ttnn.deallocate"(%1) <{force = false}> : (tensor<32x32xbf16, #ttnn_layout>) -> ()
  return %0, %2 : tensor<32x32xbf16, #ttnn_layout>, tensor<32x32xbf16, #ttnn_layout>
}
func.func @forward_merge_return_multiple_values(%arg0: tensor<32x32xbf16, #ttnn_layout> {ttcore.argument_type = #ttcore.argument_type<input>}, %arg1: tensor<32x32xbf16, #ttnn_layout> {ttcore.argument_type = #ttcore.argument_type<parameter>}, %arg2: tensor<32x32xbf16, #ttnn_layout> {ttcore.argument_type = #ttcore.argument_type<parameter>}, %arg3: tensor<32x32xbf16, #ttnn_layout> {ttcore.argument_type = #ttcore.argument_type<constant>}) -> tensor<32x32xbf16, #ttnn_layout> {
  %0:2 = ttcore.load_cached(@forward_merge_return_multiple_values_const_eval_0, [%arg1, %arg2, %arg3]) : (tensor<32x32xbf16, #ttnn_layout>, tensor<32x32xbf16, #ttnn_layout>, tensor<32x32xbf16, #ttnn_layout>) -> (tensor<32x32xbf16, #ttnn_layout>, tensor<32x32xbf16, #ttnn_layout>)
  // ...
}
```

#### After

```mlir
ttcore.device_module {
  builtin.module attributes {ttcore.system_desc = #system_desc} {
    func.func private @forward_merge_return_multiple_values_const_eval_0(%arg0: tensor<32x32xbf16, #ttnn_layout>, %arg1: tensor<32x32xbf16, #ttnn_layout>, %arg2: tensor<32x32xbf16, #ttnn_layout>) -> (tensor<32x32xbf16, #ttnn_layout>, tensor<32x32xbf16, #ttnn_layout>) attributes {const_eval} {
      // from_device, to_layout, to_dtype ops...
      %12:2 = call @hoisted_forward_merge_return_multiple_values_const_eval_0_decl(%3, %6, %9, %10, %11) {ttir.cpu_hoist_call} : (tensor<32x32xf32, #ttnn_layout3>, tensor<32x32xf32, #ttnn_layout3>, tensor<32x32xf32, #ttnn_layout3>, tensor<32x32xf32, #ttnn_layout3>, tensor<32x32xf32, #ttnn_layout3>) -> (tensor<32x32xf32, #ttnn_layout3>, tensor<32x32xf32, #ttnn_layout3>)
      // to_device, to_layout, to_dtype ops...
      return %15, %18 : tensor<32x32xbf16, #ttnn_layout>, tensor<32x32xbf16, #ttnn_layout>
    }
    func.func @forward_merge_return_multiple_values(%arg0: tensor<32x32xbf16, #ttnn_layout> {ttcore.argument_type = #ttcore.argument_type<input>}, %arg1: tensor<32x32xbf16, #ttnn_layout> {ttcore.argument_type = #ttcore.argument_type<parameter>}, %arg2: tensor<32x32xbf16, #ttnn_layout> {ttcore.argument_type = #ttcore.argument_type<parameter>}, %arg3: tensor<32x32xbf16, #ttnn_layout> {ttcore.argument_type = #ttcore.argument_type<constant>}) -> tensor<32x32xbf16, #ttnn_layout> {
      %0:2 = ttcore.load_cached(@forward_merge_return_multiple_values_const_eval_0, [%arg1, %arg2, %arg3]) : (tensor<32x32xbf16, #ttnn_layout>, tensor<32x32xbf16, #ttnn_layout>, tensor<32x32xbf16, #ttnn_layout>) -> (tensor<32x32xbf16, #ttnn_layout>, tensor<32x32xbf16, #ttnn_layout>)
      //..
    }
    func.func private @hoisted_forward_merge_return_multiple_values_const_eval_0_decl(tensor<32x32xf32, #ttnn_layout3>, tensor<32x32xf32, #ttnn_layout3>, tensor<32x32xf32, #ttnn_layout3>, tensor<32x32xf32, #ttnn_layout3>, tensor<32x32xf32, #ttnn_layout3>) -> (tensor<32x32xf32, #ttnn_layout3>, tensor<32x32xf32, #ttnn_layout3>) attributes {ttir.hoisted_func}
  }
}
ttcore.cpu_module {
  builtin.module {
    llvm.func @hoisted_forward_merge_return_multiple_values_const_eval_0(%arg0: !llvm.ptr, %arg1: !llvm.ptr, %arg2: i64, %arg3: i64, %arg4: i64, %arg5: i64, %arg6: i64, %arg7: !llvm.ptr, %arg8: !llvm.ptr, %arg9: i64, %arg10: i64, %arg11: i64, %arg12: i64, %arg13: i64, %arg14: !llvm.ptr, %arg15: !llvm.ptr, %arg16: i64, %arg17: i64, %arg18: i64, %arg19: i64, %arg20: i64, %arg21: !llvm.ptr, %arg22: !llvm.ptr, %arg23: i64, %arg24: i64, %arg25: i64, %arg26: i64, %arg27: i64, %arg28: !llvm.ptr, %arg29: !llvm.ptr, %arg30: i64, %arg31: i64, %arg32: i64, %arg33: i64, %arg34: i64) attributes {arg_ranks = [2, 2, 2, 2, 2], ttir.hoisted_func} {
      // LLVM IR ops...
    }
  }
}
```

#### CPU-hoisting multiple operations at a time

The most of the changes in this PR are focused on enabling CPU-hoisting multiple ops into single CPU-hoisted function. Instead of providing a single op to the `hoistOperationToFunction`, it now accepts `CPUHoistedOpsDescriptor` which specifies:
- The ops to be hoisted
- The output values to be returned
- Hoisted function name

Additionally, `HoistAnalyzer` class has been refactored into a function type, to enable a more convenient logic reuse.

`CPUHoistTransform` is now instantiated with `CPUHoistAnalyzerType`, which allows for separation of concerns between **how** the ops get hoisted and **which** ops should be hoisted.

#### DPS-related considerations

When the CPU-hoisted functions are lowered to the LLVM dialect, their outputs can't be returned using `func.return` op because of the various RVO issues throughout the LLVM lowering process. Instead, CPU-hoisted functions receive output destinations as their arguments - effectively meaning that CPU-hoisted functions obey DPS semantics.

Before this PR, the passes which affected the DPS-ness of the CPU-hoisted functions are:
- `CPUHoistTransform` - ensures that the output argument is added to the hoisted function even if the operation which is being hoisted isn't itself in DPS
- `ReenableLostDPS` - ensures that DPS semantics in CPU-hoisted functions still holds, because it could have been optimized-out during the ttir -> linalg pipeline. If the hoisted operation itself doesn't carry the DPS semantics, a `linalg.copy` gets inserted to copy the results to the output argument.
- `RemoveReturnValues` - removes the return ops from the CPU-hoisted function, once the existence of DPS arguments has been confirmed

This PR refactors the DPS-related passes in the following way:
- `CPUHoistTransform` now produces CPU-hoisted functions in **non-DPS form**
- `ConvertCPUHoistedFunctionsToDPS` adds necessary output arguments and replaces returns by redirecting output arguments of the linalg ops to the newly added output arguments, or by inserting `linalg.copy` ops if the output producer ops don't implement DPS semantics
- `ReenableLostDPS` and `RemoveReturnValues` passes have been removed

The arguments behind the refactor:
- In a not-so-distant future, we'll need to support CPU-hoisting in EmitPy. Here, we won't be lowering the CPU-hoisted functions to LLVM dialect - instead, they will consist `ttnn.op.golden_function` invocations, which then invoke equivalent Torch operations and thus enable CPU execution. Because DPS functions don't make much sense in Python, we would need to disable DPS produced by `CPUHoistTransform`. Having two passes - one which disables, and one which enables DPS - is a bit too chaotic, and, IMHO, a better way forward is to have DPS initially disabled, and then enable it for the targets that require it.
- DPS has been removed from TTIR altogether, which allows the simplification of DPS ops handling in CPU-hoisting-related passes.

#### Pass ordering in TTIR to TTNN pipeline

The main change in terms of pass ordering is lifting the const-eval hoisting pass before TTIR -> TTNN conversions, because we need to use TTIR -> LinAlg conversion for CPU execution.

#### Const-eval related changes

The following changes are required to the const-eval pass because of the aforementioned const-eval lifting:
- Const-eval functions are made `private`, so that DCE can erase them in case their contained ops get optimized-away
- `conv2DWeights` attribute is now retained in const-eval functions in order to ensure proper TTNN Layouts
- Some adjustments were made in PowOp TTIR -> TTNN pattern in order to enable it to discover scalar pow through const-eval function boundary

### Checklist

### DIFF
--- a/test/ttmlir/Silicon/TTNN/n150/optimizer/conv2d/conv2d_prepare_weights_const_eval.mlir
+++ b/test/ttmlir/Silicon/TTNN/n150/optimizer/conv2d/conv2d_prepare_weights_const_eval.mlir
@@ -12,7 +12,7 @@ module {
   // extracted into a separate function (named with _const_eval_N suffix) and
   // replaced with a load_cached call in the main function.
   //
-  // CHECK: func.func @conv2d_const_eval_weights_const_eval_0
+  // CHECK: func.func private @conv2d_const_eval_weights_const_eval_0
   // CHECK: = "ttnn.get_device"()
   // CHECK: = "ttnn.prepare_conv2d_weights"(%{{.*}}, %{{.*}})
   //


lib/Conversion/TTIRToTTNN/TTIRToTTNN.cpp
```diff
@@ -531,25 +532,90 @@ class ClampOpConversionPattern : public OpConversionPattern<TTIROpTy> {
...
+
+    if (!loadCachedOp) {
+      return {op, value};
+    }
+
+    int valueIndex =
+        std::distance(loadCachedOp.getResults().begin(),
+                      llvm::find(loadCachedOp.getResults(), value));
+
+    if (valueIndex < 0 || valueIndex >= loadCachedOp.getNumResults()) {
```
vmilosevic: :warning: **clang\-diagnostic\-sign\-compare** :warning:
comparison of integers of different signs: `` int `` and `` unsigned int ``

vmilosevic (commented):

include/ttmlir/Dialect/TTCore/IR/TTCoreOps.td
```diff
@@ -139,7 +139,7 @@ def TTCore_DeviceOp : TTCore_Op<"device", [Symbol]> {
   let assemblyFormat = "$sym_name `=` $device_attr attr-dict";
 }
 
-def TTCore_LoadCachedOp : TTCore_Op<"load_cached", []> {
+def TTCore_LoadCachedOp : TTCore_Op<"load_cached", [Pure]> {
```
vwellsTT, line 144: to some extent, this seems "wrong" to me--this op DOES have side effects at runtime, etc.  I guess perhaps it is pure from the IR perspective though, hmm

include/ttmlir/Dialect/TTIR/IR/TTIROpsAttrs.td
```diff
@@ -80,6 +80,15 @@ def TTIR_HoistedCallAttr : AttrDef<TTIR_Dialect, "HoistedCall", [], "::mlir::Att
   }];
 }
 
+def TTIR_HoistedFuncAttr : AttrDef<TTIR_Dialect, "HoistedFunc", [], "::mlir::Attribute"> {
```
vwellsTT: nit: we can hoist for other reasons, perhaps CPUHoisted should be in the name here?

lib/Conversion/TTIRToLinalg/TTIRToLinalg.cpp
```diff
@@ -2022,6 +2022,82 @@ class ConstantOpConversionPattern
 };
 } // namespace
 
+namespace {
```
vwellsTT, line 2025: friendly advice: in future, I'd recommend splitting big PR into smaller PRs; e.g. these op lowerings could probably be separate. It makes reviewing easier/less daunting when PR is like 10 files instead of 50.  (I've done this many times too though, echoing advice I got from others 😆 ).

vwellsTT: Not a big deal here or anything though!

lib/Conversion/TTIRToTTNN/TTIRToTTNN.cpp
```diff
@@ -531,25 +532,88 @@ class ClampOpConversionPattern : public OpConversionPattern<TTIROpTy> {
...
+    if (!calleeFunc) {
+      return {op, value};
     }
 
-    auto fullOp = mlir::dyn_cast_if_present<mlir::tt::ttnn::FullOp>(op);
-    if (!fullOp) {
-      return {};
+    auto *terminatorOp = calleeFunc.getBody().back().getTerminator();
+    auto returnOp = mlir::dyn_cast<func::ReturnOp>(terminatorOp);
+    if (!returnOp) {
```
vwellsTT: I don't think we support any other terminators today--perhaps this should assert instead

lib/Dialect/TTNN/Transforms/Passes.cpp
```diff
@@ -485,9 +485,9 @@ class TTNNTuplifyTensors
     SmallVector<func::FuncOp, 1> targetFuncOpsInput;
     SmallVector<func::FuncOp, 1> targetFuncOpsResult;
     block->walk([&](func::FuncOp funcOp) {
-      // Skip private functions.
+      // Skip function declarations (CPU-hoisted functions).
```
vwellsTT, line 553: I don't feel super strongly, but are there cases where we do want to NOT skip private funcs? I think we have some other users of private (not 100% sure)

lib/Transforms/ConvertCPUHoistedFunctionsToDPS.cpp
```diff
@@ -0,0 +1,210 @@
...
+      const auto returnTypes = llvm::to_vector<4>(
+          llvm::map_range(returnedValues, [](mlir::Value val) {
+            auto tensorType =
+                llvm::dyn_cast_or_null<RankedTensorType>(val.getType());
+            assert(tensorType &&
+                   "Return type can't be casted to RankedTensorType!");
+            return tensorType;
+          }));
+
+      // Create new function type with extra output arguments and void return
```
vwellsTT, line 87: nit: comments in this func don't adhere to style guide (end punctation etc)

lib/Dialect/TTIR/Transforms/HoistCPUOps.cpp
```diff
@@ -8,37 +8,51 @@
...
+#include "mlir/IR/Value.h"
+#include "mlir/IR/ValueRange.h"
 #include "mlir/Transforms/DialectConversion.h"
 #include "mlir/Transforms/GreedyPatternRewriteDriver.h"
+#include "llvm/ADT/STLExtras.h"
 #include "llvm/ADT/SmallSet.h"
 #include "llvm/ADT/SmallString.h"
 #include "llvm/ADT/SmallVector.h"
+#include "llvm/Support/Casting.h"
+#include <mlir/Interfaces/DestinationStyleOpInterface.h>
```
vwellsTT: nit: bad headers

lib/Dialect/TTIR/Transforms/HoistCPUOps.cpp
```diff
@@ -47,318 +61,427 @@ getOperandTensorRanks(mlir::Operation *op) {
   return ranks;
 }
 
-// Generate unique name base on operation type + argument tensors dims & types.
-static llvm::SmallString<16> generateHoistedFuncName(mlir::Operation *op) {
+// Helper function which generates a unique name based on operation type +
+// argument tensors dims & types.
+std::string generateHoistedFuncName(const OpsVectorType &ops) {
```
vwellsTT: we prefer using llvm types as much as possible--can you not stick with `llvm::SmallString` for some reason here?

lib/Dialect/TTIR/Transforms/HoistCPUOps.cpp
```diff
@@ -47,318 +61,427 @@ getOperandTensorRanks(mlir::Operation *op) {
...
+// Helper function to collect the operations producing the output values of a
+// set of operations.
+static OpsVectorType
+collectOutputProducers(const OpsVectorType &operations,
+                       const ValuesVectorType &outputValues) {
+  OpsVectorType outputProducers;
+  for (auto outputValue : outputValues) {
+    auto *definingOp = outputValue.getDefiningOp();
+
+    assert(definingOp && "Output value does not have a defining operation!");
```
vwellsTT: nit: recommend using `TT_assertv` here instead, a bit cleaner

lib/Dialect/TTIR/Transforms/HoistCPUOps.cpp
```diff
@@ -47,318 +61,427 @@ getOperandTensorRanks(mlir::Operation *op) {
...
+    return hoistedOpsDescriptors;
+  };
+}
 
-private:
-  ShouldHoistPredicateType predicate;
-};
-} // namespace
+// HoistAnalyzer which hoists const-eval functions as a whole.
+CPUHoistAnalyzerType constEvalHoistAnalyzer() {
```
vwellsTT, line 432: this seems good for now, but in theory it'd be nice in the future to have some multi-op hoister that doesn't require the ops to be a shared func etc--I think if that existed, this special case might go away (you'd have to do some analysis similar to what const-eval does to compute func boundaries -> which ops need their results returned and which do not, etc).

lib/Dialect/TTIR/Transforms/HoistCPUOps.cpp
```diff
@@ -388,19 +511,10 @@ class TTIRHoistTransform
...
-      return op->hasAttr(ttir::ShouldHoistAttr::name);
-    };
-
-    const auto combinedPredicate = [&](mlir::Operation *op) {
-      return isOpManuallyTaggedForHoisting(op) || this->customPredicate(op);
-    };
-
-    TTIRHoistAnalyze analysisPass(combinedPredicate);
-    auto hoistOpSets = analysisPass.getHoistedOps(rootModule);
+    auto hoistedOpsDescriptors = analyzer(rootModule);
```
vwellsTT: nit: this is probably a case where explicit type is better than auto (same with my old code I guess :) ).

include/ttmlir/Dialect/TTCore/IR/TTCoreOps.td
dmilinkovicTT: Yeah, this bugged me for quite some time. 

On one hand, the side effect indeed does exist, at least from the runtime tensor caching perspective.
On the other hand, the cache should be an opaque abstraction, and the side effects it produces should be transparent.

I made it `Pure` in order to allow DCE to remove the const-evaled values which become optimized-away - this happens because we've lifted the const-eval pass in the pipeline.

Alternative approach would be to add a clean-up pass for unneeded const-eval functions. Wdyt?

vwellsTT: Yeah, hmm.  Probably not worth getting too hung up the pedantics themselves.  On the other hand, it does make me nervous that we might one day introduce a bug by using an upstream pass which say does a reordering which isn't truly legal, etc. right? Not sure tbh

lib/Conversion/TTIRToLinalg/TTIRToLinalg.cpp
dmilinkovicTT: Tbh, didn't expect the PR would get this big, I also got pretty frustrated by the size at one point.

Sorry for the inconvenience - lesson learned 😄

vwellsTT: Haha, no problem; like I said, I've definitely been there too! :)

lib/Dialect/TTIR/Transforms/HoistCPUOps.cpp
dmilinkovicTT: Good point, I'll adjust the implementation so that it's clear that `CPUHoistAnalyzerType` operates on functions, in order to avoid potential expectation that it could work across the function boundaries.

I'll also create an issue to explore this possibility in the future

dmilinkovicTT: #6097

include/ttmlir/Dialect/TTCore/IR/TTCoreOps.td
dmilinkovicTT: I understand the concern, and I'm interested in the specifics of such a pass which could break the LoadCachedOp's semantics; could you please come up with an example?

vwellsTT: No, I've thought for a bit and not come up with any.  I still think it's possible, but probably not super likely.  Maybe this is fine and we can add a TODO or something--curious to see if others have opinions

dmilinkovicTT: I've added a TODO comment

lib/Transforms/ConstEvalHoist.cpp
```diff
@@ -647,9 +650,33 @@ class ConstEvalHoistTransform
     builder.setInsertionPoint(originalFunc);
     auto newFuncOp = builder.create<func::FuncOp>(originalFunc.getLoc(),
                                                   newFuncName, funcType);
-    // Mark the new function as const-eval.
+    // Mark the new function as const-eval and private
```
vwellsTT: nit: .

lib/Transforms/ConstEvalHoist.cpp
```diff
@@ -647,9 +650,33 @@ class ConstEvalHoistTransform
...
+          std::find(originalFunc.getArguments().begin(),
+                    originalFunc.getArguments().end(), input);
+
+      if (maybeFunctionArgument == originalFunc.getArguments().end()) {
+        continue;
+      }
+
+      auto originalArgIdx = maybeFunctionArgument->getArgNumber();
+
+      // Check for existence of ttmlir::utils::g_conv2dWeightAttrName
```
vwellsTT: nit: .

lib/Dialect/TTNN/Pipelines/TTNNPipelines.cpp
```diff
@@ -203,66 +203,90 @@ void createTTIRToTTNNBackendPipeline(
...
+    // We need to re-run const-eval to pick up const prepare conv2d weight ops
+    // split during the analysis passes.
+    if (options.enableConstEval) {
+      devicePm.addPass(transforms::createConstEvalHoistTransform());
+    }
+    createTTNNPipelineLayoutDecompositionPass(devicePm, options);
+    if (options.enableTrace) {
+      devicePm.addPass(tt::ttnn::createTTNNTraceHoistTransform());
+    }
+    // Fold ttcore.optimization_barrier ops before deallocation
```
vwellsTT: nit: .

lib/Dialect/TTNN/Pipelines/TTNNPipelines.cpp
```diff
@@ -203,66 +203,90 @@ void createTTIRToTTNNBackendPipeline(
...
+    devicePm.addPass(
+        ttir::createElementTypeNormalization(elementTypeNormalizationOptions));
+
+    createTTNNPipelineTTIRPasses(devicePm, options);
+
+    ttir::TTIRQuantDataTypeConversionPassOptions quantOptions;
+    quantOptions.targetBitWidth = options.quantBitWidth;
+    devicePm.addPass(ttir::createTTIRQuantDataTypeConversionPass(quantOptions));
+
+    // Const-eval hoisting passes
```
vwellsTT: nit: .

lib/Dialect/TTIR/Transforms/HoistCPUOps.cpp
```diff
@@ -47,318 +60,432 @@ getOperandTensorRanks(mlir::Operation *op) {
...
-  // Create function types.
+// Helper function to hoist an arbitrary set of ops into a new function in
+// targetModule, generate a matching extern prototype in the sourceModule, and
+// replace the ops in the set with a callOp to the extern function.
+static void hoistOperationsToFunction(CPUHoistedOpsDescriptor &descriptor,
+                                      mlir::ModuleOp sourceModule,
+                                      mlir::ModuleOp targetModule) {
+  mlir::MLIRContext *context = sourceModule.getContext();
+
+  const auto outputProducers =
```
vwellsTT: nit: consider non-auto for these 3

lib/Dialect/TTIR/Transforms/HoistCPUOps.cpp
```diff
@@ -47,318 +60,432 @@ getOperandTensorRanks(mlir::Operation *op) {
...
+      // input argument to the set of operations.
+      if (llvm::is_contained(operations, operand.getDefiningOp())) {
+        continue;
+      }
 
-  llvm::SmallVector<int64_t, 3> ranks = getOperandTensorRanks(opToHoist);
-  mlir::MLIRContext *context = sourceModule.getContext();
-  mlir::OpBuilder typeBuilder(opToHoist);
+      // Insert the argument if not already present.
+      if (llvm::is_contained(inputArguments, operand) == false) {
```
vwellsTT: shouldn't this be an `else`?

lib/Conversion/TTIRToTTNN/TTIRToTTNN.cpp
```diff
@@ -531,25 +532,86 @@ class ClampOpConversionPattern : public OpConversionPattern<TTIROpTy> {
...
+      if (isa<FloatAttr, IntegerAttr>(fillValueAttr)) {
+        return fillValueAttr;
+      }
+    }
+    return {};
+  }
+
+  // Helper function to skip TTNN and TTIR reshape and typecast ops.
+  static std::pair<mlir::Operation *, mlir::Value>
+  skipReshapeTypecastOps(mlir::Operation *op, mlir::Value value) {
```
vwellsTT: nit: and toLayout

lib/Conversion/TTIRToTTNN/TTIRToTTNN.cpp
```diff
@@ -531,25 +532,86 @@ class ClampOpConversionPattern : public OpConversionPattern<TTIROpTy> {
...
+    while (mlir::isa_and_present<
+           mlir::tt::ttnn::ReshapeOp, mlir::tt::ttnn::TypecastOp,
+           mlir::tt::ttnn::ToLayoutOp, mlir::tt::ttnn::TypecastOp>(op)) {
```
vwellsTT: nit: typecast twice?

lib/Dialect/TTIR/Transforms/HoistCPUOps.cpp
dmilinkovicTT: Just an awkward `== false`, changed to use `!`

include/ttmlir/Dialect/TTCore/IR/TTCoreOps.td
azecevic000: Does `load_cache` compute the hoisted function on the first call (or generally on cache miss)? If that's the case, the issue might be that the hoisted function does contain non-pure ops, which makes it a non-pure function. It makes sense to go with this approach for now as the simplest one, and be aware that it might cause an issue in the future.

include/ttmlir/Dialect/TTIR/IR/TTIROpsAttrs.td
```diff
@@ -71,15 +71,24 @@ def TTIR_FlattenedCompatInfoAttr : AttrDef<TTIR_Dialect, "FlattenedCompatInfo",
...
+  let mnemonic = "cpu_hoisted_call";
+  let summary = "Indicates that this call operation has been CPU-hoisted";
   let description = [{
-    This attribute marks call operations that have been hoisted during optimization.
+    This attribute marks call operations that have been CPU-hoisted during optimization.
     It is used as a marker and doesn't carry any additional data.
   }];
 }
 
+def TTIR_CPUHoistedFuncAttr : AttrDef<TTIR_Dialect, "CPUHoistedFunc", [], "::mlir::Attribute"> {
```
azecevic000, line 83: Is this attr redundant? If `cpu_hoisted_func` <=> func inside `ttcore.cpu_module`, or is there a case where this might not be true?

lib/Conversion/TTIRToLinalg/TTIRToLinalg.cpp
```diff
@@ -2022,6 +2022,82 @@ class ConstantOpConversionPattern
...
+      return rewriter.notifyMatchFailure(op, "Unsupported fill_value type");
+    }
+
+    // Create DenseElementsAttr with appropriate element type.
+    DenseElementsAttr fillAttr;
+    if (auto floatType = dyn_cast<FloatType>(elementType)) {
+      APFloat apFloat(value);
+      bool lostInfo;
+      apFloat.convert(floatType.getFloatSemantics(),
+                      APFloat::rmNearestTiesToEven, &lostInfo);
```
azecevic000: Does `arith.constant` support non-rounded values? It seems to me that this part of the code does the same thing as `createDenseElementsAttr`, though I may be missing something.

lib/Conversion/TTIRToTTNN/TTIRToTTNN.cpp
```diff
@@ -531,25 +532,87 @@ class ClampOpConversionPattern : public OpConversionPattern<TTIROpTy> {
...
+      if (isa<FloatAttr, IntegerAttr>(fillValueAttr)) {
+        return fillValueAttr;
+      }
+    }
+    return {};
+  }
+
+  // Helper function to skip TTNN and TTIR reshape and typecast ops.
+  static std::pair<mlir::Operation *, mlir::Value>
+  skipReshapeTypecastLayoutOps(mlir::Operation *op, mlir::Value value) {
```
azecevic000: This seems redundant in the sense that `Operation *` (both as an argument and return value) will be `value.getDefiningOp()`. The only edge case I can see is when `op` argument is not `value.getDefiningOp()` and `op` is not one of the layout ops, will that ever be the case in usage?

lib/Conversion/TTIRToTTNN/TTIRToTTNN.cpp
```diff
@@ -531,25 +532,87 @@ class ClampOpConversionPattern : public OpConversionPattern<TTIROpTy> {
...
     }
 
-    auto fullOp = mlir::dyn_cast_if_present<mlir::tt::ttnn::FullOp>(op);
-    if (!fullOp) {
-      return {};
+    return {op, value};
+  }
+
+  // Helper function to trace back through const-eval function.
+  static std::pair<mlir::Operation *, mlir::Value>
```
azecevic000: The same question here.

lib/Conversion/TTIRToTTNN/TTIRToTTNN.cpp
```diff
@@ -531,25 +532,87 @@ class ClampOpConversionPattern : public OpConversionPattern<TTIROpTy> {
...
+
+    // If the value is received through const-eval, we need to trace it back.
+    std::tie(op, value) = getDefiningOpThroughConstEval(op, value);
+
+    if (auto attr = getConstantAttrFromFullOp<mlir::tt::ttir::FullOp>(op)) {
+      return attr;
+    }
+    if (auto attr = getConstantAttrFromFullOp<mlir::tt::ttnn::FullOp>(op)) {
+      return attr;
+    }
```
azecevic000, line 611: Accidentally doubled?

include/ttmlir/Dialect/TTIR/IR/TTIROpsAttrs.td
dmilinkovicTT: `cpu_hoisted_func` applies to both functions in the CPU module and their stubs (declarations) in the device module

lib/Conversion/TTIRToTTNN/TTIRToTTNN.cpp
dmilinkovicTT: ttir vs ttnn, to combat the situation where function in which PowOp resides has been lowered to TTNN, but the const-eval function hasn't been lowered yet

lib/Conversion/TTIRToTTNN/TTIRToTTNN.cpp
dmilinkovicTT: makes perfect sense, refactored. thanks!

lib/Conversion/TTIRToTTNN/TTIRToTTNN.cpp
dmilinkovicTT: makes perfect sense, refactored. thanks!

lib/Conversion/TTIRToLinalg/TTIRToLinalg.cpp
dmilinkovicTT: the only issue is that `createDenseElementsAttr` currently accepts an int64 value, but seems like it could be refactored to accept doubles and use `apFloat.convert`, just as in FullOp conversion pattern. 

is that what you ment?

lib/Dialect/TTIR/Transforms/HoistCPUOps.cpp
vwellsTT: ah, yeah

include/ttmlir/Dialect/TTCore/IR/TTCoreOps.td
vwellsTT: > Does load_cache compute the hoisted function on the first call (or generally on cache miss)?

yeah exactly

lib/Conversion/TTIRToLinalg/TTIRToLinalg.cpp
```diff
@@ -22,6 +22,7 @@
 #include "llvm/ADT/SmallVector.h"
 
 #include <cstdint>
```
vmilosevic, line 26: :warning: **llvm\-include\-order** :warning:
```suggestion
#include <llvm/ADT/APFloat.h>
#include <cstdint>
```

lib/Conversion/TTIRToLinalg/TTIRToLinalg.cpp
```diff
@@ -22,6 +22,7 @@
...
+#include <llvm/ADT/APFloat.h>
```
vmilosevic: :warning: **llvm\-include\-order** :warning:
```suggestion

```

vmilosevic (commented):

lib/Conversion/TTIRToTTNN/TTIRToTTNN.cpp
```diff
@@ -531,25 +532,85 @@ class ClampOpConversionPattern : public OpConversionPattern<TTIROpTy> {
...
+
+  // Helper function to skip TTNN and TTIR reshape and typecast ops.
+  static mlir::Value skipReshapeTypecastLayoutOps(mlir::Value value) {
     mlir::Operation *op = value.getDefiningOp();
-    while (mlir::isa_and_present<mlir::tt::ttnn::ReshapeOp,
-                                 mlir::tt::ttnn::TypecastOp>(op)) {
-      op = op->getOperand(0).getDefiningOp();
+    while (mlir::isa_and_present<
+           mlir::tt::ttnn::ReshapeOp, mlir::tt::ttnn::TypecastOp,
+           mlir::tt::ttnn::ToLayoutOp, mlir::tt::ttir::TypecastOp,
```
vwellsTT: super nit: order is ttnn: reshape typecast to_layout but ttir: typecast reshape to_layout, would look a little nicer if was the same for both

lib/Dialect/TTNN/Pipelines/TTNNPipelines.cpp
```diff
@@ -203,66 +203,90 @@ void createTTIRToTTNNBackendPipeline(
...
+
+    createTTNNPipelineDeallocPass(devicePm, options);
+  }
+
+  // Run lowering to LLVM pass on hoisted funcs.
+  {
+    auto &cpuPm = pm.nest<ttcore::CPUModuleOp>().nest<mlir::ModuleOp>();
+
+    ttir::LinalgToLLVMPipelineOptions linalgToLLVMOptions;
+    ttir::createTTIRToCPUPipeline(cpuPm, linalgToLLVMOptions);
```
vwellsTT, line 283: one thing I'm confused by: you have various code throughout that handles TTIR or TTNN.  But how is the TTNN actually lowered to LLVM--is there a TTNN -> TTIR pass in this pipeline now?

dmilinkovicTT: we actually still lower only TTIR to Linalg - const-eval pass has been lifted before the TTIR -> TTNN conversion.

> you have various code throughout that handles TTIR or TTNN

you mean PowOp handling? for PowOp mix of TTIR and TTNN is there because there was a situation where const-eval function (which contains FullOp) isn't yet converted to TTNN, but ops in the function in which the PowOp resides have been converted to TTNN

vwellsTT: > you mean PowOp handling? 

Yep!

>  for PowOp mix of TTIR and TTNN is there because there was a situation where const-eval function (which contains FullOp) isn't yet converted to TTNN, but ops in the function in which the PowOp resides have been converted to TTNN

I see, though I don't think I understand exactly how that can happen if you lift before TTIR -> TTNN; is this some synthetic testcase using different pipeline or something?

dmilinkovicTT: from my understanding, here's how it goes:
- const-eval pass on TTIR -> everything is in TTIR
- TTIR -> TTNN pass - at the exact moment of PowOp conversion:
  - const-eval function is still in TTIR (not picked up for conversion yet)
  - the residing function of the PowOp is in TTNN (at least, the ops before the PowOp)

vwellsTT: Ah, I think I see what you mean; the issue is not that the PowOp itself has been hoisted, but that something it consumes is possibly hoisted so you need to consider both TTIR and TTNN FullOps.  Yeah, that makes sense then.

lib/Conversion/TTIRToLinalg/TTIRToLinalg.cpp
```diff
@@ -287,12 +288,14 @@ static Value createReductionOpChain(Value input, RankedTensorType resultType,
 // Helper function to create DenseElementsAttr with a specific value based on
 // element type.
 static Attribute createDenseElementsAttr(RankedTensorType resultType,
```
azecevic000: While you are changing this function, can you please change the return type to `DenseElementsAttr`? With that change you can avoid weird casts to `DenseElementsAttr` on the callsite, and it should be backwards compatible since `DenseElementsAttr` is-a `Attribute`.

lib/Conversion/TTIRToLinalg/TTIRToLinalg.cpp
```diff
@@ -287,12 +288,14 @@ static Value createReductionOpChain(Value input, RankedTensorType resultType,
...
-  if (isa<FloatType>(elementType)) {
-    return DenseElementsAttr::get(
-        resultType,
-        APFloat(cast<FloatType>(elementType).getFloatSemantics(), value));
+  if (auto floatType = dyn_cast<FloatType>(elementType)) {
+    APFloat apFloat(value);
+    bool lostInfo;
+    apFloat.convert(floatType.getFloatSemantics(), APFloat::rmNearestTiesToEven,
+                    &lostInfo);
+    return DenseElementsAttr::get(resultType, apFloat);
```
azecevic000: Also, a general note (since this isn't a major scope of this PR), `DenseElementsAttr` that broadcasts a single value is known as `SplatElementsAttr`. So this could be rewritten as:
```suggestion
    return SplatElementsAttr::get(resultType, FloatAttr::get(floatType, value));
```

lib/Conversion/TTIRToTTNN/TTIRToTTNN.cpp
azecevic000: Oh, I see. It's kind of a weird situation, but we can easily get away with it because those two have the same semantics in both dialects.

lib/Dialect/TTIR/Transforms/HoistCPUOps.cpp
```diff
@@ -2,43 +2,56 @@
...
 // Hoist CPU ops to standalone funcs pass
 //===----------------------------------------------------------------------===//
 
-// Helper function to get ranks of an op's operands
-// we use this to populate attrs which we need to tensor unpacking operations
-// later.
+namespace {
+using OpsVectorType = llvm::SmallVector<mlir::Operation *, 4>;
+using ValuesVectorType = llvm::SmallVector<mlir::Value, 4>;
+using TypesVectorType = llvm::SmallVector<mlir::Type, 4>;
```
azecevic000: Is there a reason to put 4 here? General advice is to leave the number of stack-allocated elements to the library implementation, unless there is a strong assumption about the upper bound.

lib/Dialect/TTIR/Transforms/HoistCPUOps.cpp
```diff
@@ -47,318 +60,433 @@ getOperandTensorRanks(mlir::Operation *op) {
...
+
+  return inputArguments;
+}
 
-  auto f32Type = mlir::Float32Type::get(context);
-  auto i32Type =
+// Helper function to convert tensor types to CPU-compatible types.
+static mlir::RankedTensorType
+convertTensorType(mlir::RankedTensorType tensorType,
+                  mlir::MLIRContext *context) {
```
azecevic000: The `context` seems like a redundant argument here. I can't think of a use case where we want to create the output in some context other than the context in which the input is created, so it might even cause an unexpected issue. You can get a `tensorType` context with `tensorType.getContext()`.

lib/Dialect/TTIR/Transforms/HoistCPUOps.cpp
```diff
@@ -47,318 +60,433 @@ getOperandTensorRanks(mlir::Operation *op) {
...
+  } else if (elementType.isFloat()) {
+    convertedElementType = f32Type;
+  }
 
-  // Convert operands and gather types for function signature.
-  llvm::SmallVector<mlir::Type> operandTypes;
-  llvm::SmallVector<mlir::Value> convertedOperands;
+  if (elementType != convertedElementType) {
+    return mlir::RankedTensorType::get(
+        tensorType.getShape(), convertedElementType, tensorType.getEncoding());
```
azecevic000: What's the expected encoding of the `tensorType`? If it's `TTNNLayoutAttr`, this might cause an issue, since the element type is also encoded in the `LayoutAttr` itself.

lib/Dialect/TTIR/Transforms/HoistCPUOps.cpp
```diff
@@ -47,318 +60,433 @@ getOperandTensorRanks(mlir::Operation *op) {
...
+  return tensorType;
+}
+
+// Helper function to convert input arguments to CPU-compatible types,
+// inserting conversion ops as needed.
+static void
+performInputArgumentsConversion(mlir::OpBuilder &opBuilder,
+                                const ValuesVectorType &inputArguments,
+                                TypesVectorType &convertedArgumentTypes,
+                                ValuesVectorType &convertedArguments) {
```
azecevic000: Is there a good reason to have these as output arguments? Unless there is a one, I'm very against it, since it's much harder to follow the flow of data, and it's much more error-prone. Types are easily obtainable from the Values itself, and since these are going to be relatively small collections, we doesn't really save much with putting everything here.
```cpp
static ValuesVectorType
performInputArgumentsConversion(mlir::OpBuilder &opBuilder,
                                const ValuesVectorType &inputArguments);
```

lib/Dialect/TTNN/Transforms/TTNNLayout.cpp
```diff
@@ -337,6 +338,44 @@ class TTNNLayoutHoistedFuncCallRewriter
...
+        break;
+      }
+      auto funcResultType = funcOp.getResultTypes()[idx];
+      if (callResultType != funcResultType) {
+        loadCachedOp->getResult(idx).setType(funcResultType);
+        modified = true;
+      }
+    }
+
+    return modified ? success() : failure();
```
azecevic000: Nit:
```suggestion
    return success(modified);
```

lib/Transforms/ConstEvalHoist.cpp
```diff
@@ -647,9 +650,33 @@ class ConstEvalHoistTransform
     builder.setInsertionPoint(originalFunc);
     auto newFuncOp = builder.create<func::FuncOp>(originalFunc.getLoc(),
                                                   newFuncName, funcType);
-    // Mark the new function as const-eval.
+    // Mark the new function as const-eval and private.
```
azecevic000, line 653: Thanks for this!

lib/Transforms/ConvertCPUHoistedFunctionsToDPS.cpp
```diff
@@ -0,0 +1,212 @@
...
+#include "ttmlir/Dialect/TTCore/IR/TTCoreOps.h"
+#include "ttmlir/Dialect/TTIR/IR/TTIROps.h"
+
+#include "mlir/Dialect/Func/IR/FuncOps.h"
+#include "llvm/ADT/SmallVector.h"
+
+#include "mlir/Dialect/Linalg/IR/Linalg.h"
+#include "mlir/IR/BuiltinOps.h"
+#include "llvm/ADT/STLExtras.h"
+#include "llvm/Support/Casting.h"
```
azecevic000, line 18: Nit: `ttmlir` headers goes into one section, `mlir/llvm` into other (https://docs.tenstorrent.com/tt-mlir/coding-guidelines.html)

lib/Transforms/ConvertCPUHoistedFunctionsToDPS.cpp
```diff
@@ -0,0 +1,212 @@
...
+      auto returnOp = llvm::dyn_cast<func::ReturnOp>(
+          funcOp.getBody().back().getTerminator());
+
+      assert(returnOp &&
+             "Hoisted function does not have a valid return operation!");
+
+      auto returnedValues = returnOp.getOperands();
+
+      const auto returnTypes = llvm::to_vector<4>(
+          llvm::map_range(returnedValues, [](mlir::Value val) {
```
azecevic000: Nit: `llvm::map_to_vector`

lib/Transforms/ConvertCPUHoistedFunctionsToDPS.cpp
```diff
@@ -0,0 +1,212 @@
...
+            return tensorType;
+          }));
+
+      // Create new function type with extra output arguments and void return
+      // type.
+      llvm::SmallVector<Type, 4> inputTypes(funcOp.getArgumentTypes().begin(),
+                                            funcOp.getArgumentTypes().end());
+      for (auto returnType : returnTypes) {
+        inputTypes.push_back(returnType);
+      }
```
azecevic000: Nit:
```suggestion
      auto inputTypes = ttmlir::utils::flatten(funcOp.getArgumentTypes(), returnTypes);
```

lib/Dialect/TTIR/Transforms/HoistCPUOps.cpp
dmilinkovicTT: Good point, I'll remove this to avoid confusion, since we shouldn't have any tensor encodings assigned at this point (`TTNNLayout` pass should happen after this one)

runtime/lib/ttnn/operations/cpu/cpu.cpp
```diff
@@ -33,12 +33,19 @@ void run(const ::tt::target::ttnn::CpuOp *op, ProgramContext &context) {
...
-      fbInputs->Get(fbInputs->size() - 1));
+  // Last inputs are actually output destinations. We need to associate outputs
+  // with the output destination tensors already in the pool.
+  for (size_t i = 0; i < op->outs()->size(); ++i) {
+    const auto &fbOutput = op->outs()->Get(i);
+
+    ::ttnn::Tensor outTensor = context.getTensorPool().getTTNNTensorAndValidate(
+        fbInputs->Get(fbInputs->size() - op->outs()->size() + i));
+
+    context.getTensorPool().insertTTNNTensorAndValidate(fbOutput, outTensor);
```
pilkicTT, line 46: @dmilinkovicTT why are we adding the same tensor again to the pool? i would expect that the output tensor id and its coresponding input tensor id are the same, is that not the case?

dmilinkovicTT: The IDs are indeed different. For CPU hoisting, we're introducing DPS-styled output tensors explicitly, and we need to associate them with the corresponding result operands.

The following example should illustrate the cause behind the difference:
```
%out1 = ttir.empty() : tensor<...>
%out2 = ttir.empty() : tensor<...>
...
%res1, %res2 = call @cpu_hoisted_function(%in1, %in2, %out1, %out2) : ...
```

We need to associate `%res1` and `%res2` with `%out1` and `%out2`

dmilinkovicTT: The changes I've introduced here are just generalization of the existing single-output approach for multiple outputs

PR #6049, opened 2025-11-26
[ttnn.jit] Add Tests for Mixed Legacy Sharding Types

brapananTT:
### Ticket
#6023 [ttnn.jit] Test mixing legacy sharding types

### Problem description
Add tests for JIT'ed binary ops with operands with mixed legacy sharding types. Ex. `ttnn.add(A, B)` where A is height sharded and B is block sharded.

### What's changed
Added test file `test_mixed_legacy_sharding_types.py` in `test/ttnn-jit/`.
There are 3 cases for mixed inputs which are tested here: 

- Height Sharded and Width Sharded Operands
- Height Sharded and Block Sharded Operands
- Width Sharded and Block Sharded Operands

Test coverage includes tensors of different rank, dtype, and tests on graph capture / ast. 

### Checklist

### DIFF
--- a/test/ttnn-jit/test_mixed_legacy_sharding_types.py
+++ b/test/ttnn-jit/test_mixed_legacy_sharding_types.py
@@ -16,9 +16,6 @@ from utils import (
     run_op_test,
 )
 
-# WIthout support for JIT'ed binary ops with inputs with differing sharding grids,
-# much of the testing here should be expected to fail. The trivial case is when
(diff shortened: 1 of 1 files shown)


brapananTT:
In the trivial case of `max_grid=(0, 0)` tests are largely passing. Most of the more complex cases, for example when `max_grid` is `(0, 7), (7, 0), (7 7)`, tests fail due to an operand grid mismatch inside the d2m generic.

PR #6051, opened 2025-11-26
[ttnn.jit] add bfp8 support

ckaravasilisTT:
Added support to handle bfp8 data. 

Added bfp8 to `test_unary_op_l1` and `test_unary_op_dram` test parameters. Updated `run_op_test`, `create_sharded_tile_tensor`, `create_dram_tensor` to also take a `ttnn_dtype=None` argument. This is because bfp8 is not available through torch. This was the most minimal change. Another option is to refactor the tests to only use `ttnn.DataType` in the parameterization.

### DIFF
--- a/test/ttnn-jit/test_eltwise.py
+++ b/test/ttnn-jit/test_eltwise.py
@@ -169,10 +169,9 @@ def rsqrt(input_tensor):
     "dtype, ttnn_dtype",
     [
         (torch.float32, None),
-        (torch.bfloat16, None),
         (torch.bfloat16, ttnn.DataType.BFLOAT8_B),
     ],
-    ids=["f32", "bf16", "bfp8"],
+    ids=["f32", "bfp8"],
 )
 @pytest.mark.parametrize(
     "op",


test/ttnn-jit/test_eltwise.py
```diff
@@ -647,3 +647,47 @@ def test_interop_jit_and_ttnn_to_binary_dram(
...
+    "shape, max_grid",
+    [
+        ((64, 64), (0, 0)),
+        ((128, 128), (0, 0)),
+        ((256, 256), (0, 0)),
+        ((256, 256), (7, 7)),
+    ],
+)
+@pytest.mark.parametrize("graph_capture", [False, True])
+def test_bfp8_exp(device, shape, max_grid, graph_capture):
```
vtangTT: can we just add this to existing `test_unary_op_l1` and `test_unary_op_dram` by just adding bfp8 to the dtype parametrization?

ckaravasilisTT: Good point, I'll try that. May have to update `utils.py` also.

tools/ttnn-jit/_src/ttir_ast.py
```diff
@@ -57,6 +57,10 @@ def _mlir_dtype_from_ttnn_dtype(self, dtype):
                 return F32Type.get(self.ctx)
             case 2:
                 return IntegerType.get_unsigned(32, self.ctx)
+            case 3:
+                return ttcore.ir.TileType.get(
```
arminaleTT, line 61: Why are we creating the tile type here? Why can't we do it similar to BF16/F32?

ckaravasilisTT: Bfp8 doesn't exist as an MLIR type (no `BFP8Type.get()`), and it can only be represented as a `TileType`. I'm not sure if there's a cleaner way to do this.

PR #6053, opened 2025-11-26
ND Virtual Grid Support

bgrady-tt:
### Problem description
Currently D2M cannot support grid shapes that aren't 2D, requiring all higher-rank shapes be collapsed to 2D. Collapsed shapes are in general harder to work with (especially for TM ops), and don't provide fine grained control of the collapsing function itself.

### What's changed
- ND virtual grids are now supported for D2M generic
  - No major IR changes
- ND virtual grids are inferred in TTIRToD2M for rank>2 shapes if the pipeline option `collapse-tensors-2d=0` is set.

### Checklist

### DIFF
--- a/lib/Dialect/D2M/Transforms/GridSelection.cpp
+++ b/lib/Dialect/D2M/Transforms/GridSelection.cpp
@@ -336,9 +336,8 @@ shouldImplementAsVirtualGrid(RankedTensorType tensorType,
     return false;
   }
 
-  // Tensors with collapsed leading dims are not currently supported as virtual
-  // grids.
-  if (layout.hasCollapsedLeadingDims(tensorType.getShape())) {
+  // Tensors with collapsed dims are not currently supported as virtual grids.
+  if (layout.hasNonTrivialCollapsedDims(tensorType.getShape())) {
     return false;
   }
   if (physicalShape.size() != 2) {
--- a/lib/Dialect/TTCore/IR/TTCoreOpsTypes.cpp
+++ b/lib/Dialect/TTCore/IR/TTCoreOpsTypes.cpp
@@ -1252,7 +1252,8 @@ int64_t MetalLayoutAttr::getHostVolume() const {
   return getHostStrideAndVolume().second;
 }
 
-bool MetalLayoutAttr::hasCollapsedLeadingDims(ArrayRef<int64_t> deviceShape) {
+bool MetalLayoutAttr::hasNonTrivialCollapsedDims(
+    ArrayRef<int64_t> deviceShape) {
 
   auto numDims = deviceShape.size();
   auto normalizedIntervals = getNormalizedIntervals();
--- a/include/ttmlir/Dialect/TTCore/IR/TTCoreOpsTypes.td
+++ b/include/ttmlir/Dialect/TTCore/IR/TTCoreOpsTypes.td
@@ -509,8 +509,8 @@ def TTCore_MetalLayoutAttr : TTCore_Attr<"MetalLayout", "metal_layout", [TTCore_
       return getIndexAffineMap();
     }
 
-    // Returns true if this layout has leading dimensions that are collapsed together
-    bool hasCollapsedLeadingDims(ArrayRef<int64_t> deviceShape);
+    // Returns true if this layout has two or more dims collapsed into a single dim
+    bool hasNonTrivialCollapsedDims(ArrayRef<int64_t> deviceShape);
 
     bool isView() { return false; }
   }];


include/ttmlir/Dialect/D2M/Utils/VirtualGrid.h
```diff
@@ -13,8 +13,73 @@
 #include "ttmlir/Utils.h"
 #include "llvm/ADT/SmallVector.h"
 
+#include <cmath>
+
 namespace ttmlir::d2m::utils::grids {
 
+inline mlir::AffineMap prependResult(mlir::AffineMap map,
```
bgrady-tt, line 19: NOTE: these affine map utils will be moved to AffineMapUtils.h

lib/Dialect/D2M/IR/D2MOps.cpp
```diff
@@ -358,11 +356,12 @@ ToLayoutOp::CompoundComponents ToLayoutOp::compoundComponents() {
   components.isFormatChange =
       inputTensor.getElementType() != outputTensor.getElementType();
 
+  // XXX: address this properly
```
bgrady-tt: **NOTE**: The layout change detected by this code causes any ND <-> 2D to_layout to crash, even though it can be absorbed into a grid change to_layout. This transition should be detected and fused into grid change to_layout ops.

lib/Dialect/D2M/Transforms/GridSelection.cpp
```diff
@@ -277,83 +394,17 @@ bool shouldImplementAsVirtualGrid(ArrayRef<int64_t> physicalShape,
 
 static std::pair<llvm::SmallVector<int64_t>, bool>
 computeOptimalGrid(ArrayRef<int64_t> physicalShape,
-                   ArrayRef<int64_t> targetSquareGridShape,
-                   bool isInterleaved) {
-  if (!isInterleaved &&
-      shouldImplementAsVirtualGrid(physicalShape, targetSquareGridShape)) {
+                   ArrayRef<int64_t> targetSquareGridShape, bool isInterleaved,
+                   bool isTiled) {
+  if (shouldImplementAsVirtualGrid(physicalShape, targetSquareGridShape,
```
bgrady-tt: NOTE: Currently working on row-major support for this PR.

lib/Dialect/D2M/Transforms/GridSelection.cpp
```diff
@@ -157,10 +157,125 @@ int64_t getTargetGridVolume(ArrayRef<int64_t> targetSquareGridShape) {
...
+    }
+  }
+  TT_assertv(false,
+             "Unable to find 2D grid with volume {} within target grid {}",
+             volume, ttmlir::utils::formatIterable(targetGridShape, "x"));
+  return {};
+}
+
+/// Returns all positive factors of `value`
+inline llvm::SmallVector<int64_t> getFactors(size_t value) {
```
bgrady-tt: NOTE: I plan to replace implementation here with something cleaner/efficient, and move this into Utils.h

lib/Dialect/D2M/Transforms/LowerToLayout.cpp
```diff
@@ -292,21 +292,25 @@ class D2MSplitCompoundLayoutRewriter : public OpRewritePattern<ToLayoutOp> {
     computeVirtualGridBounceShape(ArrayRef<int64_t> virtualGridShape,
```
bgrady-tt, line 86: NOTE: This will be replaced with an implementation based on computeOptimalGrid().

lib/Dialect/D2M/Transforms/LowerToLayout.cpp
bgrady-tt, line 113: The code to support virtual grid to_layout splitting needs to be refactored here. The new approach will detect and prioritize splitting `system -> virtual` grid to_layout ops by inserting a bounce to a _conventionally block sharded grid_ that metal's `enqueue_(read|write)_buffer` can directly copy to.

i.e. `system -> conventional block sharded -> virtual grid sharded`. 

The remaining to_layout splitting logic can then proceed on decomposing the `system -> conventional block sharded` without having to deal with virtual grid complexities.

lib/Dialect/D2M/Utils/Utils.cpp
```diff
@@ -6,76 +6,103 @@
...
+  SmallVector<AffineExpr> results(map.getNumResults());
+  for (int64_t i = 0; i < map.getNumResults(); ++i) {
+    results[i] = dimSwappedMap.getResult(invResultPermutation[i]);
+  }
+  return mlir::AffineMap::get(map.getNumDims(), 0, results, map.getContext());
+}
+} // namespace detail
+
 // Calculate a reblocking affine map from inputShape to outputShape.
 mlir::AffineMap calculateReblockMap(mlir::ArrayRef<int64_t> inputShape,
```
bgrady-tt: The implementation here works, but is probably being overly clever. I will replace it with an implementation that is more straightforward.

lib/Dialect/TTCore/IR/TTCoreOpsTypes.cpp
bgrady-tt: I am unsure that GridAttr should even have this functionality. The physical grid can be derived directly using the output tensors grid shape and ShardLayoutAttr affine map. Trying to _guess_ a working shape by relying on conventions is very fragile.

lib/Dialect/D2M/Transforms/LowerToLayout.cpp
bgrady-tt: @vwellsTT PR #5540 actually cleans up the logic here, so I will probably just rebase on top of that and fix any remaining issues.

lib/Dialect/D2M/IR/D2MOps.cpp
nsmithtt: @vwellsTT PR removes these components, unclear if it resolves the core issue.

lib/Dialect/D2M/Transforms/GridSelection.cpp
nsmithtt, line 1: @vwellsTT can you review this file?

lib/Dialect/TTCore/IR/TTCoreOpsTypes.cpp
```diff
@@ -1450,6 +1466,19 @@ mlir::AffineMap DeviceAttr::getMemoryMap(MemRefType memrefType, size_t pageSize,
...
+      // virtualization map results
+      if (affineMap.getNumDims() > coreVirtMap.getNumResults()) {
+        auto dimsToRemove =
+            affineMap.getNumDims() - coreVirtMap.getNumResults();
+        llvm::SmallBitVector projectedDims(affineMap.getNumDims());
+        projectedDims.set(0, dimsToRemove);
+
+        affineMap = getProjectedMap(affineMap, projectedDims);
+        affineMap = affineMap.dropResults(projectedDims);
+      }
```
nsmithtt, line 1525: It's unclear if we should do this here, it kind of feels like we're just fixing things up at the last minute.  I'm wondering if we should instead enforce that `d2m.empty` and `memref.alloc` always has an embedded map that has the collapse builtin and assert that it can always be composed with a device map.

lib/Dialect/D2M/Transforms/GridSelection.cpp
```diff
@@ -151,20 +151,129 @@ findMaxDimAndAspectRatio(ArrayRef<int64_t> physicalShape) {
   return {maxDimIndex, aspectRatio};
 }
 
-int64_t getTargetGridVolume(ArrayRef<int64_t> targetSquareGridShape) {
-  return std::accumulate(targetSquareGridShape.begin(),
-                         targetSquareGridShape.end(), uint64_t{1},
-                         std::multiplies<uint64_t>());
+/// Finds a 2D grid (y, x) such that y * x = volume.
```
vwellsTT: Is this meant to mean that volume % x * y == 0, or as written? Seems weird grid would be entire volume (though perhaps volume is not full tensor volume I guess)

lib/Dialect/D2M/Transforms/GridSelection.cpp
```diff
@@ -151,20 +151,129 @@ findMaxDimAndAspectRatio(ArrayRef<int64_t> physicalShape) {
...
+  TT_assertv((targetGridShape[0] > 0 && targetGridShape[1] > 0),
+             "Target grid dimensions must be positive");
+
+  auto fitsTarget = [&](int64_t dimY, int64_t dimX) {
+    return dimY <= targetGridShape[0] && dimX <= targetGridShape[1];
+  };
+
+  int64_t y = 1;
+  // Find the largest factor of volume that is <= sqrt(volume)
+  for (int64_t i = static_cast<int64_t>(std::sqrt(volume)); i > 0; --i) {
```
vwellsTT: not so sure I get this either; if volume is 10 say, optimal grid is 5x2 or 10x1 or whatever, but sqrt 10 will cast down to 3, and you won't find a valid answer, right?

lib/Dialect/D2M/Transforms/GridSelection.cpp
```diff
@@ -151,20 +151,129 @@ findMaxDimAndAspectRatio(ArrayRef<int64_t> physicalShape) {
...
+      highFactors.push_back(value / i);
+      lowFactors.push_back(i);
+    }
+  }
+  return llvm::to_vector(
+      llvm::concat<int64_t>(llvm::reverse(lowFactors), highFactors));
+}
+
+static llvm::SmallVector<llvm::SmallVector<int64_t>>
+computeFactorCombinations(ArrayRef<llvm::SmallVector<int64_t>> factors) {
```
vwellsTT: pls add comment explaining combos of what

lib/Dialect/D2M/Transforms/LowerToLayout.cpp
```diff
@@ -362,23 +375,44 @@ class D2MSplitCompoundLayoutRewriter : public OpRewritePattern<ToLayoutOp> {
...
+        }
       }
 
-      auto layout = ttcore::MetalLayoutAttr::get(
-          ctx, baseLayout.getLogicalShape(), baseLayout.getDimAlignments(),
-          baseLayout.getCollapsedIntervals(), baseLayout.getOobVal(), memSpace,
-          baseLayout.getMemoryLayout(), AffineMap::get(ctx));
+      ttcore::MetalLayoutAttr layout;
+      if (hasVirtualGrid && !reblockVirtualGridShapes) {
+        // if virtual grid is preserved (not reblocked), use empty collapsed
```
vwellsTT: This file will probably need to be re-reviewed post re-base on my LowerToLayout rework (which hopefully will land today, btw), but I don't follow why virtual grid is preserved => no collapse, it seems that would => collapse intervals should be preserved as well to me

include/ttmlir/Dialect/D2M/Utils/VirtualGrid.h
```diff
@@ -37,40 +101,58 @@ createCoreVirtMaps(mlir::MLIRContext *context,
...
+        virtualGrid.size());
+
+    auto inverseMap =
+        prependResult(create1DtoNDMap(context, virtualGrid)
+                          .compose(createCollapseMap(context, targetGrid)),
+                      getAffineConstantExpr(0, context));
+
+    return {forwardMap, inverseMap};
+  } else {
+
```
vmilosevic: :warning: **llvm\-else\-after\-return** :warning:
```suggestion
  } 
```

include/ttmlir/Dialect/D2M/Utils/VirtualGrid.h
```diff
@@ -37,40 +101,58 @@ createCoreVirtMaps(mlir::MLIRContext *context,
...
+                         d3};
+      inverseMapExprs = {d1 * gridRowStride + d0, zero};
+    }
+    auto forward =
+        mlir::AffineMap::get(2 * virtualGridRank, 0, forwardMapExprs, context);
+    auto inverse =
+        mlir::AffineMap::get(virtualGridRank, 0, inverseMapExprs, context);
+    inverse = prependResult(inverse, getAffineConstantExpr(0, context));
+    return {forward, inverse};
   }
```
vmilosevic, line 115: :warning: **llvm\-else\-after\-return** :warning:
```suggestion
 
```

include/ttmlir/Utils.h
```diff
@@ -180,6 +180,64 @@ inline T identity(T x) {
...
+  }
+
+public:
+  AscendingStrideAdaptor(Container &&container)
+      : container(std::forward<Container>(container)),
+        indices(strideIterationOrder(this->container.size())) {}
+
+  class iterator {
+    const int64_t *indexPtr;
+    Container &container;
```
vmilosevic: :warning: **cppcoreguidelines\-avoid\-const\-or\-ref\-data\-members** :warning:
member `` container `` of type `` Container & `` is a reference

include/ttmlir/Utils.h
```diff
@@ -180,6 +180,64 @@ inline T identity(T x) {
   return x;
 }
 
+// Used by ascendingStrideIter to implement ascending stride range wrapper.
+template <typename Container>
+class AscendingStrideAdaptor {
+  Container container;
```
vmilosevic: :warning: **cppcoreguidelines\-avoid\-const\-or\-ref\-data\-members** :warning:
member `` container `` of type `` llvm::ArrayRef<long> & `` is a reference

include/ttmlir/Utils.h
vmilosevic: :warning: **cppcoreguidelines\-avoid\-const\-or\-ref\-data\-members** :warning:
member `` container `` of type `` llvm::ArrayRef<long> & `` is a reference

lib/Dialect/D2M/Transforms/LowerToLayout.cpp
```diff
@@ -85,21 +85,27 @@ class D2MLowerToLayoutRewriter : public OpRewritePattern<ToLayoutOp> {
...
-        ret = {divisor, virtualGridShape[1]};
-      } else {
-        int64_t divisor = std::gcd(virtualGridShape[1], deviceGridShape[1]);
-        ret = {virtualGridShape[0], divisor};
+
+      TT_assert(virtualGridShape.size() >= 2u);
+      // Collapse all leading dimensions into the first dimension of a 2D shape
+      llvm::SmallVector<int64_t> collapsedVirtualGridShape(2);
+      collapsedVirtualGridShape[0] = virtualGridShape[0];
+      for (int64_t i = 1; i < int64_t(virtualGridShape.size()) - 1; ++i) {
```
vmilosevic: :warning: **google\-readability\-casting** :warning:
C\-style casts are discouraged; use static\_cast
```suggestion
      for (int64_t i = 1; i < static_cast<int64_t>(virtualGridShape.size()) - 1; ++i) {
```

lib/Dialect/TTCore/IR/TTCoreOpsTypes.cpp
```diff
@@ -876,13 +873,23 @@ GridAttr::getPhysicalGridShape(ArrayRef<int64_t> deviceGridShape) {
...
+        if (y <= deviceGridShape[0] && x <= deviceGridShape[1]) {
+          if (checkInjective({y, x})) {
+            return SmallVector<int64_t>{y, x};
+          }
         }
+        return std::nullopt;
+      };
+      if (auto result = checkValid(y, x)) {
+        return result.value();
+      } else if (auto result = checkValid(x, y)) {
```
vmilosevic: :warning: **llvm\-else\-after\-return** :warning:
```suggestion
      } if (auto result = checkValid(x, y)) {
```

lib/Dialect/TTCore/IR/TTCoreOpsTypes.cpp
```diff
@@ -1450,6 +1482,19 @@ mlir::AffineMap DeviceAttr::getMemoryMap(MemRefType memrefType, size_t pageSize,
     // Core virtualization map must be composed before other maps.
     if (auto coreVirtMap = shardLayout.getCoreVirtualizationMap();
         !coreVirtMap.isEmpty()) {
+
+      // TODO: generalize to drop any results not present in the core
```
vmilosevic: :warning: **google\-readability\-todo** :warning:
missing username/bug in TODO
```suggestion
      // TODO(nsmith): generalize to drop any results not present in the core
```

bgrady-tt: fixed!

lib/Dialect/TTCore/IR/TTCoreOpsTypes.cpp
bgrady-tt: fixed 👍

include/ttmlir/Utils.h
bgrady-tt: fixed

include/ttmlir/Utils.h
bgrady-tt: fixed

include/ttmlir/Utils.h
bgrady-tt: fixed

include/ttmlir/Dialect/D2M/Utils/VirtualGrid.h
bgrady-tt: fixed

include/ttmlir/Dialect/D2M/Utils/VirtualGrid.h
bgrady-tt: fixed

lib/Dialect/D2M/Transforms/GridSelection.cpp
bgrady-tt: This was indeed wrong; latest branch checks _both_ x,y and y,x. Any factors greater than sqrt(X) are given by N/smaller_factor.

lib/Dialect/D2M/Transforms/GridSelection.cpp
bgrady-tt: The volume here is just the grid volume, not full tensor volume. I can change the variable name to make this clearer.

lib/Dialect/TTCore/IR/TTCoreOpsTypes.cpp
bgrady-tt: I agree that as-is the affine maps on ND tensors are 'lop-sided'; they must collapse all grid dims to 2D, but there are still ND shard dims. In the end there will be awkwardness _somewhere_, as the ND shapes ofc don't have fixed dimensions and must have dimensions dropped in one way or another.

Could you elaborate on this builtin collapse map?

lib/Dialect/D2M/Transforms/GridSelection.cpp
vwellsTT: makes sense, thanks

lib/Dialect/D2M/Transforms/GridSelection.cpp
bgrady-tt: renamed this to computeCartesianProduct and moved to utils, and added a descriptive comment.

lib/Dialect/D2M/Transforms/LowerToLayout.cpp
bgrady-tt: I did have to make substantial changes to LowerToLayout post rebase (as expected). 

I was able to cleanup the handling of virtual grids in `modifyDeviceType` and `createDeviceType` by splitting `system -> device virtual grid` toLayout into `system -> device block sharded -> device virtual grid` _as the first step in the pipeline_.  The rest of the pipeline then doesn't need to consider virtual grids corner cases when computing appropriate bounce shapes.

lib/Dialect/D2M/Transforms/GridSelection.cpp
bgrady-tt: row major virtual grids are supported now.

lib/Dialect/D2M/IR/D2MOps.cpp
bgrady-tt: This lack of proper handling for layout changes is now addressed properly by @vwellsTT PR **https://github.com/tenstorrent/tt-mlir/pull/5540**.

include/ttmlir/Dialect/D2M/Utils/VirtualGrid.h
bgrady-tt: Fixed 👍

lib/Dialect/D2M/Transforms/LowerToLayout.cpp
```diff
@@ -565,6 +573,61 @@ class D2MLowerToLayoutRewriter : public OpRewritePattern<ToLayoutOp> {
...
+    auto hasVirtualGrid =
+        [](std::optional<ttcore::MetalLayoutAttr> maybeLayout) -> bool {
+      if (!maybeLayout) {
+        return false;
+      }
+      auto indexMap = maybeLayout->getIndexAffineMap();
+      return !indexMap.isEmpty() && !indexMap.isIdentity();
+    };
+
+    // 0. Split virtual grid shapes earlier into a conventional block sharded
```
vwellsTT, line 585: I feel like this should be step 2, after System -> Device (like if you have 1 to_layout which is moving from system -> device, and then doing the virtual reblocking, I think we would need to emit the system -> device first, then the virtual reblock)?

lib/Dialect/TTCore/IR/TTCoreOpsTypes.cpp
```diff
@@ -1106,13 +1119,23 @@ MetalLayoutAttr MetalLayoutAttr::get(::mlir::MLIRContext *context,
                                      OOBVal oobVal, MemorySpace memorySpace,
                                      TensorMemoryLayout memoryLayout,
                                      DenseIntElementsAttr collapsedIntervals) {
+  return get(context, logicalShape, oobVal, memorySpace, memoryLayout,
+             collapsedIntervals, mlir::AffineMap::get(context));
+}
+// Getter with explicit collapsedIntervals, we calculate the alignments.
```
vwellsTT, line 1054: nit: \n

lib/Dialect/TTCore/IR/TTCoreOpsTypes.cpp
```diff
@@ -1220,6 +1243,28 @@ int64_t MetalLayoutAttr::getHostVolume() const {
   return getHostStrideAndVolume().second;
 }
 
+bool MetalLayoutAttr::hasCollapsedLeadingDims(ArrayRef<int64_t> deviceShape) {
+
```
vwellsTT, line 1257: nit: remove \n

vwellsTT: a bit confused what this func is actually doing, it seem like it's detecting whether there are any un-collapsed dims, and if so returns true, else false? (If so, should be renamed)

include/ttmlir/AffineMapUtils.h
```diff
@@ -114,22 +121,26 @@ fullyApplyAffineMap(mlir::OpBuilder &builder, mlir::Location loc,
...
+      llvm::SmallVector<int64_t>(map.getNumResults(), 0);
+  ttmlir::utils::sample(gridShape, [&](llvm::SmallVector<int64_t, 8> point) {
+    llvm::SmallVector<int64_t> virtualPoint = map.compose(point);
     for (size_t i = 0; i < virtualPoint.size(); ++i) {
       resultGridShape[i] = std::max(resultGridShape[i], virtualPoint[i] + 1);
+      lowerBound[i] = std::min(lowerBound[i], virtualPoint[i]);
     }
   });
+  if (assertResultStartsAtOrigin) {
+    TT_assertv(std::all_of(lowerBound.begin(), lowerBound.end(),
```
vwellsTT: nit: use llvm::all_of

include/ttmlir/Dialect/D2M/Utils/VirtualGrid.h
```diff
@@ -5,16 +5,80 @@
...
+    extendedResults.push_back(getAffineDimExpr(i, map.getContext()));
+  }
+
+  return mlir::AffineMap::get(map.getNumDims() + extraDims, map.getNumSymbols(),
+                              extendedResults, map.getContext());
+}
+
+inline mlir::AffineMap createCollapseMap(mlir::MLIRContext *context,
+                                         llvm::ArrayRef<int64_t> virtualGrid) {
+  using namespace mlir;
```
vwellsTT: nit: I think `using namespace` is forbidden in our style guide

include/ttmlir/Dialect/D2M/Utils/VirtualGrid.h
```diff
@@ -37,16 +101,28 @@ createCoreVirtMaps(mlir::MLIRContext *context,
...
+    return {forwardMap, inverseMap};
+  }
+
+  bool is2DWidthSharded = (virtualGridRank == 2) && virtualGrid[0] == 1;
+  bool is2DHeightSharded = (virtualGridRank == 2) && virtualGrid[1] == 1;
 
   TT_assertv((is2DWidthSharded || is2DHeightSharded),
              "Only supporting 2D width or height sharding (actual grid shape = "
-             "{})",
+             "{1})",
```
vwellsTT: this is wrong, llvm fmt uses 0-indexing.  And specifying index explicitly is more error prone than not, pls remove

include/ttmlir/Utils.h
```diff
@@ -127,6 +127,55 @@ std::string formatIterable(Range &&iterable, llvm::StringRef separator = ",") {
   return "(" + content + ")";
 }
 
+/// Returns all positive factors of `value`
```
vwellsTT, line 130: nit: end with .

include/ttmlir/Utils.h
```diff
@@ -127,6 +127,55 @@ std::string formatIterable(Range &&iterable, llvm::StringRef separator = ",") {
...
+// computeCartesianProduct(possibilities = {{1, 2}, {3, 4, 5}})
+//   -> {{1, 3}, {1, 4}, {1, 5}, {2, 3}, {2, 4}, {2, 5}}
+template <typename T>
+std::vector<llvm::SmallVector<T>>
+computeCartesianProduct(llvm::ArrayRef<llvm::SmallVector<T>> possibilities) {
+  if (possibilities.empty()) {
+    return {};
+  }
+
+  // recursively enumerate all combinations of the possibilities
```
vwellsTT: nit: proper sentence

lib/Dialect/D2M/Transforms/GridSelection.cpp
vwellsTT: I do think gridVolume would be clearer, FWIW

lib/Dialect/D2M/Transforms/GridSelection.cpp
```diff
@@ -153,36 +153,98 @@ findMaxDimAndAspectRatio(ArrayRef<int64_t> physicalShape) {
...
+                               ArrayRef<int64_t> targetSquareGridShape);
+
+static llvm::SmallVector<int64_t>
 computeOptimalVirtualGrid(ArrayRef<int64_t> physicalShape,
                           ArrayRef<int64_t> targetSquareGridShape) {
 
+  int64_t targetGridVolume = ttmlir::utils::volume(targetSquareGridShape);
+  if (physicalShape.size() != 2) {
+
+    // compute factors for all dims
```
vwellsTT: nit: proper sentence

lib/Dialect/D2M/Transforms/GridSelection.cpp
```diff
@@ -153,36 +153,98 @@ findMaxDimAndAspectRatio(ArrayRef<int64_t> physicalShape) {
...
+    SmallVector<SmallVector<int64_t>> factors =
+        llvm::to_vector(llvm::map_range(physicalShape, [](int64_t dim) {
+          return ttmlir::utils::getFactors(dim);
+        }));
+
+    auto factorCombinations =
+        ttmlir::utils::computeCartesianProduct<int64_t>(factors);
+
+    // try to find grid with largest volume that is <= target grid volume
```
vwellsTT: nit: proper sentence

lib/Dialect/D2M/Transforms/GridSelection.cpp
```diff
@@ -153,36 +153,98 @@ findMaxDimAndAspectRatio(ArrayRef<int64_t> physicalShape) {
...
+      int64_t gridVolume = ttmlir::utils::volume<int64_t>(grid);
+      if (gridVolume <= targetGridVolume && gridVolume > bestGridVolume) {
+        bestGrid = grid;
+        bestGridVolume = ttmlir::utils::volume<int64_t>(bestGrid);
+      }
+    }
+    return bestGrid;
+  }
+
+  // 2D HW Sharding
```
vwellsTT: nit: proper sentence

lib/Dialect/D2M/Transforms/LowerToLayout.cpp
```diff
@@ -85,21 +85,28 @@ class D2MLowerToLayoutRewriter : public OpRewritePattern<ToLayoutOp> {
...
-      llvm::SmallVector<int64_t> ret;
-      if (virtualGridShape[0] > deviceGridShape[0]) {
-        int64_t divisor = std::gcd(virtualGridShape[0], deviceGridShape[0]);
-        ret = {divisor, virtualGridShape[1]};
-      } else {
-        int64_t divisor = std::gcd(virtualGridShape[1], deviceGridShape[1]);
-        ret = {virtualGridShape[0], divisor};
+
+      TT_assert(virtualGridShape.size() >= 2u);
+      // Collapse all leading dimensions into the first dimension of a 2D shape
```
vwellsTT: nit: .

lib/Dialect/D2M/Transforms/LowerToLayout.cpp
```diff
@@ -565,6 +573,61 @@ class D2MLowerToLayoutRewriter : public OpRewritePattern<ToLayoutOp> {
...
+      auto l1Type = typeBuilder.modifyDeviceType(
+          targetInfo.type, *targetInfo.layout, targetGridShape,
+          ttcore::MemorySpace::DeviceL1, bounceShape);
+      auto l1Empty = createEmpty(l1Type);
+
+      // Create a ToLayoutOp from system to block sharded bounce tensor.
+      auto sysToBounceToLayout =
+          lowerSystemLayoutChange(rewriter, currentValue, l1Empty, op.getLoc());
+
+      // replace ToLayout producing virtual grid target with blockToVirtual
```
vwellsTT: nit: proper sentence

lib/Dialect/D2M/Transforms/LowerToLayout.cpp
```diff
@@ -565,6 +573,61 @@ class D2MLowerToLayoutRewriter : public OpRewritePattern<ToLayoutOp> {
...
+      auto blockToVirtual = lowerDatamovementGeneric(
+          rewriter, sysToBounceToLayout, op.getOutput(), op.getLoc());
+
+      // final output is from blockToVirtual
```
vwellsTT: nit: proper sentence

lib/Dialect/D2M/Transforms/LowerToLayout.cpp
```diff
@@ -565,6 +573,61 @@ class D2MLowerToLayoutRewriter : public OpRewritePattern<ToLayoutOp> {
...
+      rewriter.replaceOp(op, blockToVirtual);
+
+      // remaining pipeline splits the sys -> BS tolayout
```
vwellsTT: nit: proper sentence

lib/Dialect/TTCore/IR/TTCoreOpsTypes.cpp
```diff
@@ -837,35 +825,42 @@ GridAttr::getPhysicalGridShape(ArrayRef<int64_t> deviceGridShape) {
...
       for (int x = 0; x < physGridShape[1]; x++) {
-        auto vCoord = ArrayRef<int64_t>{map.compose({y, x})}.drop_front(1);
-        TT_assertv(vCoord.size() == 2ul, "Virtual grid coordinate must be 2D.");
+        auto fullVCoord = map.compose({y, x});
 
-        auto index = vCoord[0] * vGrid[1] + vCoord[1];
-        // if location is already occupied, the physical grid is invalid
-        if (bitmap[index]) {
-          return false;
+        // drop device index result from coord
```
vwellsTT: nit: proper sentence

lib/Dialect/TTCore/IR/TTCoreOpsTypes.cpp
```diff
@@ -837,35 +825,42 @@ GridAttr::getPhysicalGridShape(ArrayRef<int64_t> deviceGridShape) {
...
+        int64_t stride = 1;
+        for (auto [i, coord] : llvm::enumerate(vCoord)) {
+          index += coord * stride;
+          stride *= vGrid[i];
         }
-        // if the location is not within the virtual grid, the physical grid is
-        // invalid
-        if (vCoord[0] >= vGrid[0] || vCoord[1] >= vGrid[1]) {
+
+        // index must be in bounds and location not occupied
```
vwellsTT: nit: proper sentence

lib/Dialect/TTCore/IR/TTCoreOpsTypes.cpp
```diff
@@ -837,35 +825,42 @@ GridAttr::getPhysicalGridShape(ArrayRef<int64_t> deviceGridShape) {
...
-        bitmap[index] = true;
+        // If the location is not within the virtual grid, the physical grid is
+        // invalid.
+        for (size_t i = 0; i < vCoord.size(); ++i) {
+          if (vCoord[i] < 0 || vCoord[i] >= vGrid[i]) {
+            return false;
+          }
+        }
+
+        // mark location as occupied
```
vwellsTT: nit: proper sentence

lib/Dialect/D2M/IR/D2MOps.cpp
```diff
@@ -1217,14 +1217,8 @@ ::mlir::LogicalResult d2m::GenericOp::verify() {
       }
       AffineMap fwdMap = *maybeFwdMap;
 
-      // Drop the shard dim results from the virtual grid mapping.
-      if (fwdMap.getNumResults() % 2 != 0) {
```
vwellsTT, line 1221: is this not still true?

lib/Dialect/TTCore/IR/TTCoreOpsTypes.cpp
bgrady-tt: It scans the collapsed intervals and checks if they _actually_ collapse dims together i.e. `PxQxR` is collapsed to `(PQ)xR`. This is done by checking if _all_ the intervals have range==1. 

I could rename this to `hasNonTrivialCollapsedDims()` to convey this (where 'trivial' means every dim in the logical shape has its own collapse interval pair).

lib/Dialect/D2M/Transforms/LowerToLayout.cpp
bgrady-tt: I will move the virtual grid reblocking bounce to be after System -> Device. It helps to split off virtual grid bounce earlier, so the rest of the pipeline can compute bounce shapes in a more straightforward way.

lib/Dialect/D2M/Utils/Utils.cpp
bgrady-tt: greatly simplified implementation in latest version.

lib/Dialect/D2M/Transforms/LowerToLayout.cpp
bgrady-tt: Newest version of this is similar to logic in `computeOptimalBlockShardedGrid()`, but handles collapsing leading dims if needed and computes optimal block sharded bounce shape. Not worth additional coupling to save 10 lines of simple code.

lib/Dialect/D2M/IR/D2MOps.cpp
bgrady-tt: The code here doesn't align with the comment, but I think it is probably related to the logic for dropping all but two result from the mapping.

The core virtualization maps have a form like so (`gd` is a grid dim, `sd` is a shard dim):
```
(gd0, gd1, ... gdN, sd0, sd1, ... sdN) 
-> (
      expr(gd0, gd1, ... gdN), // phys grid Y; function of grid dims only
      expr(gd0, gd1, ... gdN), // phys grid X; function of grid dims only
      sd0, sd1, ... sdN        // identity mapping for all shard dims; grid dims cannot be used here
      )
```

So we are just dropping all but the first two results from the map (e.g. the physical grid results), as we only care in this context about virtGrid <-> physGrid mapping, not the shard dims themselves.

lib/Dialect/TTCore/IR/TTCoreOpsTypes.cpp
bgrady-tt: removed this code in latest version.

lib/Dialect/TTCore/IR/TTCoreOpsTypes.cpp
bgrady-tt: removed this code in latest version

lib/Dialect/TTCore/IR/TTCoreOpsTypes.cpp
bgrady-tt: removed this code in latest version

lib/Dialect/D2M/Transforms/LowerToLayout.cpp
bgrady-tt: fixed 👍

lib/Dialect/D2M/Transforms/LowerToLayout.cpp
bgrady-tt: fixed 👍

lib/Dialect/D2M/Transforms/LowerToLayout.cpp
bgrady-tt: fixed 👍

lib/Dialect/D2M/Transforms/LowerToLayout.cpp
bgrady-tt: fixed

lib/Dialect/D2M/Transforms/GridSelection.cpp
bgrady-tt: Fixed 👍

lib/Dialect/D2M/Transforms/GridSelection.cpp
bgrady-tt: Fixed 👍

lib/Dialect/D2M/Transforms/GridSelection.cpp
bgrady-tt: Fixed 👍

include/ttmlir/Utils.h
bgrady-tt: Fixed 👍

include/ttmlir/Utils.h
bgrady-tt: Fixed 👍

include/ttmlir/Dialect/D2M/Utils/VirtualGrid.h
bgrady-tt: Fixed 👍

include/ttmlir/Dialect/D2M/Utils/VirtualGrid.h
bgrady-tt: Fixed 👍

include/ttmlir/AffineMapUtils.h
bgrady-tt: Fixed 👍

lib/Dialect/D2M/Transforms/LowerToLayout.cpp
bgrady-tt: LowerToLayout has been substantially refactored after rebasing on Vincent's PR #5540

lib/Dialect/D2M/Transforms/GridSelection.cpp
bgrady-tt: This was renamed to be `gridVolume`. Comment was also updated to improve clarity.

include/ttmlir/AffineMapUtils.h
```diff
@@ -5,15 +5,15 @@
...
+#include "ttmlir/Asserts.h"
+#include "ttmlir/Utils.h"
+
 #include "mlir/Conversion/AffineToStandard/AffineToStandard.h"
 #include "mlir/Dialect/Affine/IR/AffineOps.h"
 #include "mlir/IR/AffineMap.h"
 #include "mlir/IR/BuiltinAttributes.h"
-#include "ttmlir/Asserts.h"
-#include "ttmlir/Utils.h"
 #include "llvm/ADT/SmallVector.h"
```
vmilosevic: :warning: **llvm\-include\-order** :warning:
```suggestion
#include "llvm/ADT/STLExtras.h"
#include "llvm/ADT/SmallVector.h"
```

include/ttmlir/AffineMapUtils.h
```diff
@@ -5,15 +5,15 @@
...
-
-#include <cstdint>
+#include "llvm/ADT/STLExtras.h"
```
vmilosevic, line 15: :warning: **llvm\-include\-order** :warning:
```suggestion

```

vmilosevic (commented):

lib/Dialect/TTCore/IR/TTCoreOpsTypes.cpp
```diff
@@ -828,67 +822,6 @@ static llvm::SmallVector<int64_t> applyCollapsedIntervalsAndAlignments(
   return resultShape;
 }
 
-SmallVector<int64_t>
```
bgrady-tt, line 831: NOTE: Deriving the physical grid shape from the GridAttr was a clumsy process that relied on guessing and verifying a plausible inverse shape based on solving `virtualGrid = inverseMap(physGrid)`, where inverseMap and virtualGrid are unknown. 

It is better to _unambiguously_ derive the physical grid shape by simply sampling the virtual grid shape through the core virtualization map.

lib/Dialect/TTCore/IR/TTCoreOpsTypes.cpp
vwellsTT: I would vote for that, imo current name is misleading because PxQxR -> Px(QR) would return true, right?

lib/Dialect/TTCore/IR/TTCoreOpsTypes.cpp
vwellsTT: does this entail changes w.r.t. what we discussed for the invMap? is this live somewhere in PR?

lib/Dialect/D2M/IR/D2MOps.cpp
vwellsTT: I see, was just curious if the number of *results* is not still constrained to 2 for this mapping or not

lib/Dialect/TTCore/IR/TTCoreOpsTypes.cpp
bgrady-tt: The only place `GridAttr::getPhysicalGridShape` was used to derive the physical grid shape for a `GenericOp` was in `TTMetalToFlatbuffer`. The relevant change is in commit **90727d7**, where the GenericOp has a member function `getPhysicalGridShape` added that derives the grid shape from the output tensors map and grid shape.

So this eliminates a major point of unreliability and complexity. But we still need the inverse map on the GridAttr (for now). I think the better scheme is to have both fwd and inv maps _in the tensor/memref layout_; both have to be created at the same time and consistent with each other. The GenericOp then inherits its physical grid attribute from the output tensor itself.

vwellsTT: Gotcha, makes sense to me

nsmithtt: I see why we need an inv map on the grid, because we truly need both views.  But I'd be reluctant to add the inv map to tensors/memrefs unless we have a concrete need.  It adds a lot of verbosity to the IR.

include/ttmlir/AffineMapUtils.h
bgrady-tt: Not present in latest commit

include/ttmlir/AffineMapUtils.h
bgrady-tt: Fixed.

bgrady-tt:
@nsmithtt This PR supports ND row-major tensors too 👍

lib/Dialect/TTCore/IR/TTCoreOpsTypes.cpp
bgrady-tt: This API was removed; the authoritative way to get a GenericOp's physical grid shape is via GenericOp::getPhysicalGridShape().

lib/Dialect/D2M/Transforms/LowerToLayout.cpp
bgrady-tt: @vwellsTT I have decided to commit the approach above as-is (splitting virtual grids into conventional bounce early).

A follow up PR is **needed** to improve handling of virtual grids in `LowerToLayout`; however my changes were becoming very unwieldy to get ND shapes through some of the new mapping change logic.

lib/Dialect/TTCore/IR/TTCoreOpsTypes.cpp
bgrady-tt: changed the name to `hasNonTrivialCollapsedDims()` 👍

lib/Dialect/D2M/Transforms/LowerToLayout.cpp
bgrady-tt: @vwellsTT After spending a good amount of time trying, I think a larger refactor is in order here. But this PR is not the place to do it.

vwellsTT: No objections on my end, I sympathize 😄

lib/Dialect/D2M/Transforms/LowerToLayout.cpp
vwellsTT: Okay, let know if we need to discuss or anything

PR #6060, opened 2025-11-27
[TTNN] Introduce PrepareConvTranspose2dWeight and PrepareConvTranspose2dBias

jserbedzijaTT:
### Ticket
[PrepareConvTranspose2dWeight](https://github.com/tenstorrent/tt-mlir/issues/4021)
[PrepareConvTranspose2dBias](https://github.com/tenstorrent/tt-mlir/issues/4022)

### Problem description
Implementation for `PrepareConvTranspose2dWeight` and `PrepareConvTranspose2dBias` were missing in the TTNN dialect.

### Checklist


include/ttmlir/OpModel/TTNN/TTNNOpModel.h
```diff
@@ -26,6 +26,10 @@ mlir::RankedTensorType
 getPreparedConv2dWeightsOutputTensor(Conv2dOp *op,
                                      Conv2dConfigAttr conv2dConfig);
 
+mlir::RankedTensorType
+getPreparedConvTranspose2dWeightsOutputTensor(ConvTranspose2dOp *op,
+                                              Conv2dConfigAttr conv2dConfig);
+
```
sgholamiTT: I don't think this is the right place to put this API.
Let's move it to another file (For example, you can put it under `include/ttmlir/OpModel/TTNN/OutputTensorInference.h` and implementation in `lib/OpModel/TTNN/OutputTensorInference.cpp`).

Please also add a TODO with a metal issue  (you can ask for feature request for exposing an API for this, as that would be the right way of handling this).

lib/OpModel/TTNN/TTNNOpModel.cpp
```diff
@@ -5089,6 +5131,145 @@ llvm::Expected<OpConstraints> OpModel<PrepareConv2dBiasOp>::getOpConstraints(
...
+    return llvm::createStringError(
+        llvm::inconvertibleErrorCode(),
+        "ConvTranspose2d weight tensor assumed to be on host.");
+  }
+  if (weightLayout.getDataType() != ttcore::DataType::Float32 &&
+      weightLayout.getDataType() != ttcore::DataType::BFloat16) {
+    return llvm::createStringError(
+        llvm::inconvertibleErrorCode(),
+        "ConvTranspose2d weight tensor assumed to be float32 or bfloat16.");
+  }
```
sgholamiTT: Is this the right place to check for these assumptions (and returning `inconvertibleErrorCode`)?
I thought one assumption of OpModel is that we can technically query the constraints with an incorrect set of input args for an op, and we'd still be able to handle that correctly (based on whatever metal assumes/does for the op).
That's why we have no other hardcoding condition on OpModel, because we basically rely on the metal assumptions and implementations.

@rpavlovicTT Please advice if I'm wrong.

lib/OpModel/TTNN/TTNNOpModel.cpp
```diff
@@ -5089,6 +5131,145 @@ llvm::Expected<OpConstraints> OpModel<PrepareConv2dBiasOp>::getOpConstraints(
...
+    return llvm::createStringError(
+        llvm::inconvertibleErrorCode(),
+        "ConvTranspose2d bias tensor assumed to be on host.");
+  }
+  if (biasLayout.getDataType() != ttcore::DataType::Float32 &&
+      biasLayout.getDataType() != ttcore::DataType::BFloat16) {
+    return llvm::createStringError(
+        llvm::inconvertibleErrorCode(),
+        "ConvTranspose2d bias tensor assumed to be float32 or bfloat16.");
+  }
```
sgholamiTT: Similar here.

sdjordjevicTT (approved):
Thanks!

include/ttmlir/OpModel/TTNN/TTNNOpModel.h
jserbedzijaTT: There was an issue that got closed because it was not urgent for us at the moment: https://github.com/tenstorrent/tt-metal/issues/21061. Do you think we should reopen it? @odjuricicTT

lib/Dialect/TTNN/Transforms/OptimizerPasses/TTNNPrepareConv2dWeightsAndBias.cpp
```diff
@@ -136,6 +139,122 @@ class TTNNPrepareConv2dWeightsAndBias
         conv2dOp.setConv2dConfigAttr(conv2dConfig);
       });
     });
+
+    moduleOp.walk([&](ttnn::ConvTranspose2dOp convTranspose2dOp) {
+      ttnn::TTNNLayoutAttr weightLayoutAttr = mlir::cast<ttnn::TTNNLayoutAttr>(
```
mtopalovicTT: This part
```
      ttnn::TTNNLayoutAttr weightLayoutAttr = mlir::cast<ttnn::TTNNLayoutAttr>(
          convTranspose2dOp.getWeight().getType().getEncoding());
      assert(weightLayoutAttr.getBufferType() ==
                 ttnn::BufferType::SystemMemory &&
             weightLayoutAttr.getLayout() == ttnn::Layout::RowMajor &&
             "Weight must be in system memory and row-major layout when "
             "calling TTNNPrepareConv2dWeightsAndBias pass.");

      ttnn::TTNNLayoutAttr biasLayoutAttr;
      if (convTranspose2dOp.getBias()) {
        biasLayoutAttr = mlir::cast<ttnn::TTNNLayoutAttr>(
            convTranspose2dOp.getBias().getType().getEncoding());
        assert(biasLayoutAttr.getBufferType() ==
                   ttnn::BufferType::SystemMemory &&
               biasLayoutAttr.getLayout() == ttnn::Layout::RowMajor &&
               "Bias must be in system memory and row-major layout when "
               "calling TTNNPrepareConv2dWeightsAndBias pass");
      }

      ttnn::TTNNLayoutAttr inputLayoutAttr = mlir::cast<ttnn::TTNNLayoutAttr>(
          convTranspose2dOp.getInput().getType().getEncoding());
      ttnn::TTNNLayoutAttr outputLayoutAttr = mlir::cast<ttnn::TTNNLayoutAttr>(
          convTranspose2dOp.getResult().getType().getEncoding());

      ttcore::GridAttr deviceGrid =
          ttcore::lookupDevice(moduleOp).getWorkerGrid();
      ttnn::MemoryConfigAttr inputMemConfigAttr =
          ttnn::MemoryConfigAttr::get(inputLayoutAttr, deviceGrid);

      // Input and output dtype attr on prepare api is used to specify
      // input/output dtype for the conv2d operation.
      auto inputDtypeAttr = mlir::tt::ttcore::DataTypeAttr::get(
          &getContext(), inputLayoutAttr.getDataType());
      auto outputDtypeAttr = mlir::tt::ttcore::DataTypeAttr::get(
          &getContext(), outputLayoutAttr.getDataType());

      // When weight dtype is set in conv2d config prepare API will typecast the
      // output into desired dtype. Until we have some different use case we
      // will put prepared bias/weight into the same dtype as input.
      auto conv2dConfig = convTranspose2dOp.getConv2dConfigAttr()
                              ? convTranspose2dOp.getConv2dConfigAttr()
                              : Conv2dConfigAttr::get(&getContext());

      Type inputElementType = inputLayoutAttr.getScalarElementType();
      conv2dConfig = conv2dConfig.withWeightsDtype(inputDtypeAttr.getValue());
```
is pretty much the same for conv2d and conv2dtranspose right

include/ttmlir/OpModel/TTNN/TTNNOpModel.h
odjuricicTT: I wouldn't revisit that for this purpose.

Talking with conv folks a long time ago about this we agreed to keep it as is. In a way query_op_constraints is the API to get the output of an op / metal function.

ctodTT (approved):

lib/Dialect/TTNN/Transforms/OptimizerPasses/TTNNPrepareConv2dWeightsAndBias.cpp
jserbedzijaTT: I refactored a logic so there is no more redudancy

test/python/golden/ttir_ops/convolution/test_conv2d.py
```diff
@@ -0,0 +1,173 @@
...
+    after each test to prevent OOM errors from accumulated allocations.
+    """
+    yield
+    conftest = sys.modules.get("conftest")
+    if conftest and conftest._current_device is not None:
+        tt_runtime.runtime.close_mesh_device(conftest._current_device)
+        tt_runtime.runtime.set_fabric_config(tt_runtime.runtime.FabricConfig.DISABLED)
+        conftest._current_device = None
+        conftest._current_device_target = None
+        conftest._current_device_mesh_shape = None
```
jameszianxuTT: Is it fully necessary to reset device between tests? This seems like it would set a bad precedent and could spread + hide other errors wrt. cross-test resource leaks. 

This might be related - [previous investigation in tt-xla](https://github.com/tenstorrent/tt-xla/issues/2227#issuecomment-3571706129) showed us that L1_SMALL leaks from conv2d was due to a metal limitation scoped to program cache specifically. 

You might be able to just use runtime to clear program cache via #6059 and have it work without the hammer of reset device between tests. Easy way to test this is also to just disable program cache (which is now on by default, #5651) and rerun your tests to see if you overrun L1.

However, if that is not an option or doesn't work, that is fine too. I would just suggest not using this nuke solution if a more precise one works.

jserbedzijaTT: I managed to do it with `device.clear_program_cache()`, thanks for the suggestion

jameszianxuTT: nice! good to know that's the only issue here.

brataTT (approved):

lib/Dialect/TTNN/IR/TTNNOps.cpp
```diff
@@ -832,34 +945,53 @@ ::mlir::LogicalResult mlir::tt::ttnn::ConvTranspose2dOp::verify() {
...
+    if (!biasIsPrepared) {
+      // Only check when the bias is not prepared because it changes the shape.
+      if (bias->getDimSize(bias->getRank() - 1) != outputChannels) {
+        return emitOpError()
+               << "Mismatch in bias tensor dimensions. "
+               << "Bias tensor has " << bias->getDimSize(bias->getRank() - 1)
+               << " channels, "
+               << "but the output tensor has " << outputChannels
+               << " channels.";
+      }
```
svuckovicTT: This isn't entirely correct. If we break up the forward fn into multiple fns, the producer for the conv op can be a func op's argument. I've encountered this when "prettifying" in codegen.

Here's a quickfix example:
https://github.com/tenstorrent/tt-mlir/compare/svuckovic/readability-mvp-4#diff-a6cb8b5b9ca05403b2ffed57b2fc9e353ac983c5f7ae34c7a1cb385e7af2777fR421-R430

Ideally, we check that the tensor shape is correct instead of looking at the producer

jserbedzijaTT: I updated the check for preparedWeight and removed the check for preparedBias since it does not affect the bias shape.

PR #6075, opened 2025-11-28
[ttnn.jit] Run Only a Subset of TTNN JIT Tests in onPush/onPR

arminaleTT:
### Ticket
N/A

### Problem description
TTNN JIT has a lot of tests and the testing matrix is growing quickly. Running all tests for every push in a PR is unsustainable. JIT tests are already around 15-20min.

### What's changed
- Separate some TTNN JIT tests to only run in the nightly and uplift jobs.
- Add ttnn-jit-nightly as an optional component
- Only run the nightly-exclusive jit tests when the ttnn-jit-nightly optional component is present. This saves about 10min for onPR

On PR
Nightly run:

### DIFF
--- a/.github/test_scripts/ttnn_jit.sh
+++ b/.github/test_scripts/ttnn_jit.sh
@@ -11,7 +11,7 @@ export PYTHONPATH="$INSTALL_DIR/tt-metal/ttnn:$INSTALL_DIR/tt-metal"
 echo "Downloading ttnn-jit wheel..."
 cd $WORK_DIR
 
-# This script may be called multiple times in the same test job. 
+# This script may be called multiple times in the same test job.
 # Delete the downloaded wheel since gh run will not overwrite an existing file.
 if [ -f ttnn_jit*.whl ]; then
     rm -f ttnn_jit*.whl


arminaleTT:
I picked the environment variable and the overall methodology based on what is done in the Mac OS build (See https://github.com/tenstorrent/tt-mlir/blob/main/.github/workflows/call-build-macos.yml#L90). I am open to other implementations. I just need the test script to know if we are in nightly or not

test/ttnn-jit/utils.py
```diff
@@ -121,7 +121,6 @@ def create_sharded_tile_tensor(
         shard_spec=shard_spec,
     )
 
-    print("Armin: creating tensor with memory config", memory_config)
```
vtangTT, line 124: cleanup

arminaleTT: Yes, this is being removed!

arminaleTT:
Thanks for the guidance @nsumrakTT 
I have made the changes you suggested to use optional components and avoid changing the workflow files

vtangTT (approved):
JIT changes lgtm ty! 🙏

nsumrakTT (approved):
Thank you!

PR #6095, opened 2025-12-01
`actBlockHOverride` search for conv2d ops

rpavlovicTT:
- Enable conv2d config search for `actBlockHOverride` field. 
- Force `deallocate_activation=true` on conv2d ops (should be safe as we don't shard forked ops). Added sanity check nevertheless and a unit test for it.
- Allow ConvTranspose2d to be sharded by optimizer, thus choosing working conv2d config along way.

### DIFF
--- a/tools/explorer/tt_adapter/src/tt_adapter/mlir.py
+++ b/tools/explorer/tt_adapter/src/tt_adapter/mlir.py
@@ -444,12 +444,14 @@ def parse_ttnn_ttnn_layout(attr):
 def parse_conv2d_config(attr):
     conv2d_config = ttnn.ir.Conv2dConfigAttr.maybe_downcast(attr)
     result = []
+    weights_dtype = str(
+        ttcore.DataType(conv2d_config.weights_dtype_as_int)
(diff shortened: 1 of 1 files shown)


tools/explorer/tt_adapter/src/tt_adapter/mlir.py
```diff
@@ -444,12 +444,14 @@ def parse_ttnn_ttnn_layout(attr):
 def parse_conv2d_config(attr):
     conv2d_config = ttnn.ir.Conv2dConfigAttr.maybe_downcast(attr)
     result = []
+    weights_dtype = str(
+        ttcore.DataType(conv2d_config.weights_dtype_as_int)
+        if conv2d_config.weights_dtype_as_int
+        else OVERRIDE_PARAMETER_DISABLED_STR
```
brataTT, line 450: Thanks for fixing this. Didn't notice there was a `None` case to handle here.

brataTT (approved):

jazpurTT (approved):

PR #6109, opened 2025-12-01
[ttnn.jit] Remove max_grid from jit decorator

brapananTT:
### Problem description
Passing in a max_grid parameter into the jit decorator restricts all operands of JIT'ed functions to have identical max_grids. This may not always be wanted, especially in the cases of JIT'ed ops with L1 and DRAM operands, or 2 L1 operands with different sharding types or sharding grids.

### What's changed
The max_grid from JIT frontend is removed. It now uses `tensor_arg` to retrieve the sharding grid via a helper function `_get_grid_from_bounding_box()`, within `tensor_translator`.py:

```
def _get_grid_from_bounding_box(tensor_arg):

    core_range_set = tensor_arg.memory_config().shard_spec.grid
    number_of_core_ranges = len(core_range_set.ranges())
    if number_of_core_ranges != 1:
        raise ValueError(
            "Tensor grids with more than one CoreRange are not supported. Tensor grids must be rectangular."
        )

    core_coord = core_range_set.bounding_box().grid_size()
    max_grid = (core_coord.x, core_coord.y)

    return max_grid
```

### Checklist

### DIFF
--- a/test/ttnn-jit/test_program_cache.py
+++ b/test/ttnn-jit/test_program_cache.py
@@ -32,15 +32,15 @@ def test_jit_cache(device):
     - Same operation, same parameters, only data differs
 
     Cache should miss when:
-    - Different JIT parameters (e.g., max_grid)
+    - Different JIT parameters (e.g., debug)
     - Different tensor properties (memory config, data type, shape)
     """
     shape1 = (256, 256)
     shape2 = (320, 320)
 
     # Create operations with different JIT parameters
-    op_single_core = jit(enable_cache=True)(abs)
-    op_full_grid = jit(enable_cache=True)(abs)
+    op_single_core = jit(debug=True, enable_cache=True)(abs)
+    op_full_grid = jit(debug=False, enable_cache=True)(abs)
 
     assert op_single_core.num_entries == 0, "No entries should be in the cache"
     assert op_full_grid.num_entries == 0, "No entries should be in the cache"


tools/ttnn-jit/_src/tensor_translator.py
```diff
@@ -111,36 +111,55 @@ def _ttcore_dtype_from_ttnn_dtype(dtype):
...
 
 
-def _get_grid(ctx, max_grid, memory_layout):
+def _get_grid_from_bounding_box(tensor_arg):
+
+    core_range_set = tensor_arg.memory_config().shard_spec.grid
+    number_of_core_ranges = len(core_range_set.ranges())
+    if number_of_core_ranges != 1:
+        raise ValueError(
+            f"CoreRangeSet should have only 1 CoreRange, but got {number_of_core_ranges}"
```
arminaleTT: Can you change this message to: "Tensor grids with more than one CoreRange are not supported. Tensor grids must be rectangular."? This is more actionable.

tools/ttnn-jit/_src/tensor_translator.py
```diff
@@ -111,36 +111,55 @@ def _ttcore_dtype_from_ttnn_dtype(dtype):
...
+
+
+def _get_grid(ctx, tensor_arg, memory_layout):
     # IMPORTANT: TTNN writes grids as (width, height) but compiler expects (height, width)
-    grid_size_x = max_grid[0] + 1
-    grid_size_y = max_grid[1] + 1
 
-    if memory_layout in (
+    if memory_layout == ttnn.TensorMemoryLayout.Interleaved:
+        return ttcore.ir.GridAttr.get(ctx, [1, 1])
```
arminaleTT: Add a constant tuple `DRAM_GRID_SIZE = (1,1)` to communicate that this is a special value

tools/ttnn-jit/_src/tensor_translator.py
```diff
@@ -190,7 +209,8 @@ def _create_dram_tensor_layout(ctx, tensor_arg, max_grid):
     )
 
 
-def _check_layout_supported(tensor_arg, max_grid):
+def _check_layout_supported(tensor_arg):
+    # todo: get max grid from tensor arg
```
arminaleTT: Stale todo?

test/ttnn-jit/test_program_cache.py
```diff
@@ -39,8 +39,8 @@ def test_jit_cache(device):
     shape2 = (320, 320)
 
     # Create operations with different JIT parameters
-    op_single_core = jit(max_grid=(0, 0), enable_cache=True)(abs)
-    op_full_grid = jit(max_grid=(7, 7), enable_cache=True)(abs)
```
arminaleTT, line 43: This test is confusing as written. There is no difference between `op_single_core` and `op_full_grid` now, though the cache behaviour remains the same since the tensor grids are still read and respected. Could you clean up the test?

cc: @vtangTT

arminaleTT (commented):
Thanks for doing this! A few comments to be addressed but otherwise looks good

test/ttnn-jit/test_program_cache.py
brapananTT: I could replace this with another jit parameter e.g. `debug=True` and `False`.  @vtangTT would this still be valid for the purposes of the test?

vtangTT: Yes that should work.

brapananTT: done!

Issue #6058, opened 2025-11-27
Remove program cache logs from op constraints api


odjuricicTT:
Current perf benchmark CI logs are full of the below:
```
2025-11-27 10:53:41.434 | info     |           Metal | Disabling and clearing program cache on MeshDevice 3 (mesh_device.cpp:710)
2025-11-27 10:53:41.434 | info     |           Metal | Enabling program cache on MeshDevice 3 (mesh_device.cpp:700)
2025-11-27 10:53:41.434 | info     |           Metal | Disabling and clearing program cache on MeshDevice 3 (mesh_device.cpp:710)
2025-11-27 10:53:41.528 | info     |           Metal | Enabling program cache on MeshDevice 3 (mesh_device.cpp:700)
2025-11-27 10:53:41.528 | info     |           Metal | Disabling and clearing program cache on MeshDevice 3 (mesh_device.cpp:710)
2025-11-27 10:53:41.529 | info     |           Metal | Enabling program cache on MeshDevice 3 (mesh_device.cpp:700)
2025-11-27 10:53:41.529 | info     |           Metal | Disabling and clearing program cache on MeshDevice 3 (mesh_device.cpp:710)
2025-11-27 10:53:41.529 | info     |           Metal | Enabling program cache on MeshDevice 3 (mesh_device.cpp:700)
2025-11-27 10:53:41.530 | info     |           Metal | Disabling and clearing program cache on MeshDevice 3 (mesh_device.cpp:710)
2025-11-27 10:53:41.530 | info     |           Metal | Enabling program cache on MeshDevice 3 (mesh_device.cpp:700)
2025-11-27 10:53:41.531 | info     |           Metal | Disabling and clearing program cache on MeshDevice 3 (mesh_device.cpp:710)
2025-11-27 10:53:41.531 | info     |           Metal | Enabling program cache on MeshDevice 3 (mesh_device.cpp:700)
2025-11-27 10:53:41.531 | info     |           Metal | Disabling and clearing program cache on MeshDevice 3 (mesh_device.cpp:710)
2025-11-27 10:53:41.532 | info     |           Metal | Enabling program cache on MeshDevice 3 (mesh_device.cpp:700)
2025-11-27 10:53:41.532 | info     |           Metal | Disabling and clearing program cache on MeshDevice 3 (mesh_device.cpp:710)
2025-11-27 10:53:41.532 | info     |           Metal | Enabling program cache on MeshDevice 3 (mesh_device.cpp:700)
[...]

PR #6111, opened 2025-12-01
Suppress getOpConstraint tt-metal logs in tt-mlir

vkovacevicTT:
### Ticket
Adresses #6058

### Problem description
Since #6094 we are using [LogLevelGuard from tt-metal](https://github.com/tenstorrent/tt-metal/blob/7c4c6782f3b73a01fe99b2dc93080fa7b240759b/ttnn/api/ttnn/graph/graph_query_op_constraints.hpp#L25-L39) (created [here](https://github.com/tenstorrent/tt-metal/pull/30746)) to suppress logs in tt-mlir that come from enabling/disabling program cache:

https://github.com/tenstorrent/tt-mlir/blob/6e0bace5a83ee45a654f560e96cc057a1c0a47ca/lib/OpModel/TTNN/TTNNOpModel.cpp#L67-L100

This way we suppress logging in tt-mlir, and then again in tt-metal.
Also, tt-metal LogLevelGuard returns all logger types to the saved level of `LogOp` logger, which is a bug.

We should consider migrating LogLevelGuard logic from tt-metal to tt-mlir, that way tt-mlir is responsible for suppressing redundant logs.

### DIFF
--- a/lib/OpModel/TTNN/TTNNOpModel.cpp
+++ b/lib/OpModel/TTNN/TTNNOpModel.cpp
@@ -37,6 +37,33 @@ namespace mlir::tt::ttnn::op_model {
 #ifdef TTMLIR_ENABLE_OPMODEL
 namespace operation {
 
+/// RAII helper to temporarily suppress tt-metal logs queries.
+struct MetalLogLevelGuard {
+  MetalLogLevelGuard(spdlog::level::level_enum logLevel,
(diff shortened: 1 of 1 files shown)


sdjordjevicTT:
@vkovacevicTT is this still relevant, can we close this one?

vkovacevicTT:
@sdjordjevicTT 
Not relevant anymore, closing :)

Issue #5997, opened 2025-11-23
Support scatter operation with shardy


AleksKnezevic:
Add support for scatter in UpdateGlobalToLocalShapes. 

@ddilbazTT this is needed for gptOss as well as [transformers upgrade](https://github.com/tenstorrent/tt-xla/issues/1020), so please prioritize. 

@sshonTT can help if needed.

jazpurTT:
There is also changes needed in this [branch](https://github.com/tenstorrent/tt-mlir/tree/jazpur/reduce-window) for the transformers uplift. We can add them to @ddilbazTT PR for this issue so we can have it all in a single commit on tt-mlir.

ddilbazTT:
@jazpurTT I think we should keep PRs separately, feel free to put up a PR for reduce window uint8 support.

ddilbazTT:
I would like to explain what is happening in gpt-oss. This is the unsupported scatter op:
```
    %209 = "stablehlo.scatter"(%0, %200, %208) <{scatter_dimension_numbers = #stablehlo.scatter<inserted_window_dims = [0, 1], scatter_dims_to_operand_dims = [0, 1], index_vector_dim = 2>}> ({
    ^bb0(%arg414: tensor<bf16> loc("scatter.925"), %arg415: tensor<bf16> loc("scatter.925")):
      stablehlo.return %arg415 : tensor<bf16> loc(#loc)
    }) : (tensor<71x32xbf16>, tensor<71x4x2xi64>, tensor<71x4xbf16>) -> tensor<71x32xbf16> loc(#loc606)
```
It is failing due to:
```
  // insertedWindowDims / scatterDimsToOperandDims must be a single dim and
  // equal for a cache update.
  if (insertedWindowDims.size() != 1 || scatterDimsToOperandDims.size() != 1 ||
      (insertedWindowDims.front() != scatterDimsToOperandDims.front())) {
    return false;
  }
```

The example we are dealing with has `inserted_window_dims = [0, 1], scatter_dims_to_operand_dims = [0, 1]` 

I think there is an easy way to address this error but a generalist conversion might be difficult/ more time-consuming. @jazpurTT can you attach the scatter failures you encountered?

ddilbazTT:
I talked with @sshonTT offline. He gave me a good direction to work with:
- Allow sharding on the input operand only if the sharded dimension is not included in scatter_dims_to_operand_dims; otherwise, force it to replicate.
[...]

PR #6124, opened 2025-12-02
Support scatter operation with shardy

ddilbazTT:
### Ticket
#5997 

### Problem description
Scatter is not supported with shardy fully in our stack. The current limitations introduced to allow certain scatter operations to pass through in `lib/Dialect/StableHLO/Transforms/UpdateGlobalToLocalShapes.cpp` are not a long-term solution, and fail with gpt-oss-20b scatter op. It is difficult to update global to local dims of scatter in compile time allowing [default shardy scatter rule builder to work.](https://github.com/openxla/shardy/blob/23359578eec0292a45958f560b19ea574d98e669/shardy/dialect/sdy/transforms/propagation/op_sharding_rule_registry.cc#L1097) This implies the need for adding our custom scatter sharding rule, which would be a fallback option when the default one wouldn't work without global to local shape modifications.

### What's changed
A solution to this problem is to write a custom sharding rule for scatter. This rule will be used when scatter input is sharded along scatter_dims_to_operand_dims. Otherwise, scatter will use the default rule. The rule follows as such:
- shard input, updates and output along sharding dim IF it is NOT in scatter_dims_to_operand_dims or inserted_window_dims. otherwise replicate
- Replicate indices
Furthermore, removed all scatter limitations in `lib/Dialect/StableHLO/Transforms/UpdateGlobalToLocalShapes.cpp` All shapes are passed as is. 
Added two examples:
    - `scatter_test_replicate_input`: This is from gpt-oss-20b op. Input is sharded along dim 1, which is in scatter_dims_to_operand_dims ([0, 1]). It is expected that input, updates and output are replicated.
    - `scatter_test_shard_input`: This is a test I wrote which has mutually exclusive sharding dim [1] and scatter_dims_to_operand_dims [0] / inserted_window_dims [0]. It is expected that input, update and output to be sharded.

### Checklist
- [X] Added test/ttmlir/Silicon/StableHLO/llmbox/sdy_scatter.mlir. 
- [X] Added more cases to test/ttmlir/Dialect/StableHLO/shardy/op_propagation_registry/scatter.mlir

Thanks @sshonTT for the recommendations.

### DIFF
--- a/test/ttmlir/Dialect/StableHLO/shardy/op_propagation_registry/scatter.mlir
+++ b/test/ttmlir/Dialect/StableHLO/shardy/op_propagation_registry/scatter.mlir
@@ -31,7 +31,7 @@ module {
 
 // -----
 
-// Test scatter with sharding along scatter_dims_to_operand_dims.
+// Test scatter with sharding along scatter_dims_to_operand_dims. The sharded input must be replicated.
 module {
   sdy.mesh @mesh = <["model"=1, "batch"=8]>
     func.func @scatter_test_replicate_input(%arg0: tensor<71x4xbf16> {sdy.sharding = #sdy.sharding<@mesh, [{}, {}]>},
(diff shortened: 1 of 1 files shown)


lib/Dialect/StableHLO/Transforms/UpdateGlobalToLocalShapes.cpp
```diff
@@ -290,17 +193,10 @@ static FailureOr<mlir::OperationState> createNewOperationState(
...
-              return mlir::success();
-            }
-
-            // If not a safe cache update, emit the error as before
-            scatterOp->emitError(
-                "Scatter operation is not supported in stablehlo-pipeline for "
-                "meshes not 1x1: "
-                "https://github.com/tenstorrent/tt-mlir/issues/3496.");
-            return mlir::failure();
+            // TODO: (@ddilbazTT) Implement updating attribute dictionary
```
vmilosevic: :warning: **google\-readability\-todo** :warning:
missing username/bug in TODO
```suggestion
            // TODO(nsmith): (@ddilbazTT) Implement updating attribute dictionary
```

lib/Dialect/StableHLO/Transforms/RegisterCustomShardingRule.cpp
```diff
@@ -63,6 +147,23 @@ getSDPAShardingRule(mlir::stablehlo::CustomCallOp op) {
       .build();
 }
 
+template <typename OpTy>
+struct StablehloShardingModel
+    : public mlir::sdy::ShardingRuleOpInterface::ExternalModel<
+          StablehloShardingModel<OpTy>, OpTy> {
+
+  mlir::sdy::OpShardingRuleAttr getShardingRule(mlir::Operation *op) const {
+    if (auto scatterOp = llvm::cast<mlir::stablehlo::ScatterOp>(op)) {
```
vmilosevic: :warning: **llvm\-prefer\-isa\-or\-dyn\-cast\-in\-conditionals** :warning:
cast\<\> in conditional will assert rather than return a null pointer
```suggestion
    if (auto scatterOp = dyn_cast::cast<mlir::stablehlo::ScatterOp>(op)) {
```

lib/Dialect/StableHLO/Transforms/RegisterCustomShardingRule.cpp
```diff
@@ -15,6 +15,90 @@ namespace mlir::tt::stablehlo {
...
+           sdy::kNullDim}, // [input_dim, indices_dim, updates_dim]
+          {inputDim},      // result_dim
+          inputType.getDimSize(inputDim),
+          sdy::FactorType::kPassThrough // Can be sharded.
+      );
+    }
+  }
+
+  // Replicate all scatter_indices dimensions.
+  for (int64_t indicesDim = 0; indicesDim < indicesRank; indicesDim++) {
```
anuragsingh-tt, line 126: We can use `llvm::enumerate` for the for loops instead.

lib/Dialect/StableHLO/Transforms/RegisterCustomShardingRule.cpp
```diff
@@ -63,6 +173,23 @@ getSDPAShardingRule(mlir::stablehlo::CustomCallOp op) {
       .build();
 }
 
+template <typename OpTy>
+struct StablehloShardingModel
+    : public mlir::sdy::ShardingRuleOpInterface::ExternalModel<
+          StablehloShardingModel<OpTy>, OpTy> {
+
+  mlir::sdy::OpShardingRuleAttr getShardingRule(mlir::Operation *op) const {
+    if (auto scatterOp = dyn_cast::cast<mlir::stablehlo::ScatterOp>(op)) {
```
vmilosevic: :warning: **clang\-diagnostic\-error** :warning:
`` dyn_cast `` is not a class, namespace, or enumeration

vmilosevic: :warning: **clang\-diagnostic\-error** :warning:
expected `` ( `` for function\-style cast or type construction

lib/Dialect/StableHLO/Transforms/RegisterCustomShardingRule.cpp
```diff
@@ -15,6 +16,115 @@ namespace mlir::tt::stablehlo {
...
+
+  auto dimNums = scatterOp.getScatterDimensionNumbers();
+  ArrayRef<int64_t> scatterDimsToOperandDims =
+      dimNums.getScatterDimsToOperandDims();
+  ArrayRef<int64_t> insertedWindowDims = dimNums.getInsertedWindowDims();
+
+  // Return null if input is not sharded along scatter_dims_to_operand_dims.
+  // This would fallback to default sharding rule in
+  // op_sharding_rule_registry.cc.
+  bool isShardedAlongScatterDims = false;
```
sshonTT, line 66: Is this variable necessary? It seems we can just return the default attribute on line 65.

ddilbazTT: How would that look?

sshonTT: Oh I got this the opposite way. But in this case, we can use `any_of`, something like
```
if (sdy::TensorShardingAttr inputSharding =
        mlir::sdy::getSharding(inputs.front())) {
  ArrayRef<mlir::sdy::DimensionShardingAttr> dimShardings =
      inputSharding.getDimShardings();

  if (!llvm::any_of( scatterDimsToOperandDims, [&](int64_t scatterDim) { 
       return scatterDim < static_cast<int64_t>(dimShardings.size()) &&            
                                       !dimShardings[scatterDim].getAxes().empty(); })) { 
        // Return null to fallback to default sharding rule. 
        return mlir::sdy::OpShardingRuleAttr();
  }
}
```

sshonTT: never mind, mines still looks bad

ddilbazTT: Yeah I think neither is better than the other, I'll leave as is.

lib/Dialect/StableHLO/Transforms/RegisterCustomShardingRule.cpp
```diff
@@ -2,6 +2,7 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 
+#include "shardy/dialect/sdy/ir/utils.h"
```
vwellsTT, line 5: nit: should this be separate include block from ttmlir? not familiar with what we do for SHLO/shardy normally I guess

vwellsTT (approved):
lgtm

lib/Dialect/StableHLO/Transforms/RegisterCustomShardingRule.cpp
ddilbazTT: Yeah we might need to restructure this file in the future anyways if we end up adding more custom rules. For now I think it's ok

hshahTT (approved):

ddilbazTT:
Ran tt-xla passing models after rebasing to [5b727af9e4e24ed6656048603a74efcbb9d2efe8 ](https://github.com/tenstorrent/tt-mlir/commit/5b727af9e4e24ed6656048603a74efcbb9d2efe8)https://github.com/tenstorrent/tt-xla/actions/runs/20013912058 

I haven't debugged yet, but a tt-mlir commit after [5b727af9e4e24ed6656048603a74efcbb9d2efe8](https://github.com/tenstorrent/tt-mlir/commit/5b727af9e4e24ed6656048603a74efcbb9d2efe8) and up until [5b727af9e4e24ed6656048603a74efcbb9d2efe8](https://github.com/tenstorrent/tt-mlir/commit/5b727af9e4e24ed6656048603a74efcbb9d2efe8) is causing tensor-parallel/ data-parallel failures. Which is why I had to rebase.

Issue #6134, opened 2025-12-03
[ttnn-jit] DRAM operands force all L1 operands on same op to be 1x1 block sharded


bgrady-tt:
### Describe the bug

DRAM interleaved tensors are represented with a 1x1 unit grid scheme in D2M, matching the TTNN convention. If a DRAM interleaved tensor is used as a d2m.generic operand (i.e. as the output of the `TTIRToD2M` pass), it forces the other operands to also be on a 1x1 grid to pass verification post `TTIRToD2M` pass.

This is problematic for TTNN jit operands in L1, as these tensors are likely already block sharded to a much larger grid and cannot have their grids modified.

Ideally it would be possible to shim all TTNN L1 operands with views to allow verification to pass, and optimize away these views in the downstream GridSelection pass.

PR #6135, opened 2025-12-03
[ttnn-jit] Wrap TTNN L1 operands in unit grid views to enable mixed L1/DRAM operands

bgrady-tt:
### Ticket
See #6134 

### Problem description
DRAM interleaved tensors are represented with a 1x1 unit grid scheme in D2M, matching the TTNN convention. If a DRAM interleaved tensor is used as a d2m.generic operand (i.e. as the output of the `TTIRToD2M` pass), it forces the other operands to also be on a 1x1 grid to pass verification post `TTIRToD2M` pass.

This is problematic for TTNN jit operands in L1, as these tensors are likely already block sharded to a much larger grid and cannot have their grids modified.

### What's changed
- **TTIRToD2M**: All L1 operands that are the result of a `MetalLayoutCastOp` (e.g. TTNN tensors) are wrapped in a reblocking view that makes them appear as 1x1 sharded. This mimicks the approach taken in the rest of TTIRToD2M; all operands start as 1x1 sharded to ensure shape verification passes.
- **Grid Selection**: During `insertTTNNDRAMStreams()` these unit grid reblocking views are stripped away, recovering the raw MetalLayoutCastOp results as the GenericOp operands. DRAM streams are then adjusted to match constraints set by the fixed L1 operand shapes.

### Checklist

### DIFF
--- a/test/ttmlir/Conversion/TTIRToD2M/ttnn_mixed_l1_dram_operand_translation.mlir
+++ b/test/ttmlir/Conversion/TTIRToD2M/ttnn_mixed_l1_dram_operand_translation.mlir
@@ -47,6 +47,13 @@ func.func @test_mixed_operands_eltwise_binary_l1_dram_dram(
     %arg1: tensor<512x2048xbf16, #ttnn_layout_dram>)
     -> tensor<512x2048xbf16, #ttnn_layout_dram>
 {
+  // CHECK: %[[L1_CAST:.*]] = ttir.ttnn_metal_layout_cast %arg0 {{.*}} -> tensor<2x2x8x32x!ttcore.tile<32x32, bf16>, #layout1>
+  // CHECK: %[[DRAM_CAST:.*]] = ttir.ttnn_metal_layout_cast %arg1 {{.*}} -> tensor<1x1x16x64x!ttcore.tile<32x32, bf16>, #layout>
+  // CHECK: %[[STREAM:.*]] = "d2m.stream_layout"(%[[DRAM_CAST]], %{{.*}}) : (tensor<1x1x16x64x!ttcore.tile<32x32, bf16>, #layout>, tensor<2x2x8x32x!ttcore.tile<32x32, bf16>, #layout1>) -> tensor<2x2x8x32x!ttcore.tile<32x32, bf16>, #layout2>
+  // CHECK: %[[OUT_STREAM:.*]] = "d2m.stream_layout"(%{{.*}}, %{{.*}}) : (tensor<1x1x16x64x!ttcore.tile<32x32, bf16>, #layout>, tensor<2x2x8x32x!ttcore.tile<32x32, bf16>, #layout1>) -> tensor<2x2x8x32x!ttcore.tile<32x32, bf16>, #layout2>
+  // CHECK: %{{.*}} = d2m.generic {block_factors = [1, 1], grid = #ttcore.grid<2x2>
+  // CHECK: ins(%[[L1_CAST]], %[[STREAM]] : tensor<2x2x8x32x!ttcore.tile<32x32, bf16>, #layout1>, tensor<2x2x8x32x!ttcore.tile<32x32, bf16>, #layout2>)
+  // CHECK: outs(%[[OUT_STREAM]] : tensor<2x2x8x32x!ttcore.tile<32x32, bf16>, #layout2>)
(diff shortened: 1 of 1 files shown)


lib/Dialect/D2M/Transforms/GridSelection.cpp
```diff
@@ -759,10 +760,85 @@ static bool hasTTNNOperands(d2m::GenericOp genericOp) {
     if (operand.getDefiningOp<ttir::TTNNMetalLayoutCastOp>()) {
       return true;
     }
+    // Bypass any views applied in TTIRToD2M to check if viewed tensor itself is
+    // the result of a TTNNMetalLayoutCastOp.
+    else if (auto view = operand.getDefiningOp<d2m::ViewLayoutOp>()) {
```
vmilosevic: :warning: **llvm\-else\-after\-return** :warning:
```suggestion
    if (auto view = operand.getDefiningOp<d2m::ViewLayoutOp>()) {
```

lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
```diff
@@ -130,8 +148,25 @@ class D2MNamedRewriterCommon {
...
+      if (metalTensorMemSpace == ttcore::MemorySpace::DeviceL1) {
+        // Reblock L1 operand to unit grid to align with other operands while
+        // preserving original TTNN tensor shape. These views will be removed in
+        // GridSelection by insertTTNNDRAMStreams().
+        auto unitReblockingView = rewriter.create<d2m::ViewLayoutOp>(
+            value.getLoc(), reblockTensor(metalTensorType, {1, 1}),
+            metalCastOp->getResult(0));
+        return unitReblockingView.getResult();
+      } else {
+        // For DRAM operands, we can return the metal cast result directly.
```
vmilosevic: :warning: **llvm\-else\-after\-return** :warning:
```suggestion
      }         // For DRAM operands, we can return the metal cast result directly.
```

lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
```diff
@@ -130,8 +148,25 @@ class D2MNamedRewriterCommon {
...
+        return metalCastOp->getResult(0);
+      }
```
vmilosevic, line 148: :warning: **llvm\-else\-after\-return** :warning:
```suggestion
     
```

vmilosevic (commented):

lib/Dialect/D2M/Transforms/GridSelection.cpp
bgrady-tt: fixed in latest version 👍

lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
bgrady-tt: Fixed in latest version 👍

lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
bgrady-tt: Fixed in latest version 👍

include/ttmlir/Dialect/D2M/Utils/Utils.h
```diff
@@ -24,6 +24,11 @@ calculateReblockMapForGrid(mlir::ArrayRef<int64_t> tensorShape,
                            mlir::ArrayRef<int64_t> newGridShape,
                            mlir::MLIRContext *context);
 
+// Return a new RankedTensorType by reblocking its device shape to match a new
+// grid shape.
+RankedTensorType reblockTensor(RankedTensorType oldTensor,
```
vwellsTT, line 14: I could've sworn I have some util like this somewhere but I can't find it 🤔

vwellsTT: Oh, I'm dumb--you're moving from GridSelection to here.  Sounds good!

bgrady-tt: I refactored it out of GridSelection, as I would have to duplicate it exactly in TTIRToD2M.

vwellsTT: Yep, makes more sense as shared util imo

arminaleTT:
This makes sense to me. How have you tested this change?

xanderchin (approved):

PR #6139, opened 2025-12-03
Fix PagedUpdateCacheOp to handle optional page_table parameter

odjuricicTT:
### Ticket
N/A

### Problem description
The `page_table` operand is optional in `PagedUpdateCacheOp` (defined as `Optional<AnyRankedTensor>` in the ODS), but the OpModel interface functions `getOpConstraints` and `getOpRuntime` were asserting exactly 4 inputs. This caused a crash when compiling models that use `PagedUpdateCacheOp` without a `page_table` tensor.

The crash manifested as:
```
Assertion `inputs.size() == 4' failed.
```

### What's changed
- Updated `getOpConstraints` and `getOpRuntime` in `TTNNOpModelInterface.cpp` to accept 3 or 4 inputs (matching the optional nature of `page_table`)
- Changed the `OpModel<PagedUpdateCacheOp>` function signatures in both the header and implementation to use `std::optional` for `pageTableShape` and `pageTableLayout` parameters
- Added conditional handling for the optional `page_table` tensor, following the pattern used by `PagedFillCacheOp`
- Added a new unit test `PagedUpdateCacheOpWithoutPageTable` to cover the case when `page_table` is not provided
- Added llama_3_2_1b_1_layer model test that exercises this code path

### Checklist


odjuricicTT:
Performance benchmark started: https://github.com/tenstorrent/tt-xla/actions/runs/19891280533

PR #6153, opened 2025-12-03
[ttnn.jit] Enable Block Sharded Matmuls Through the AST Frontend

arminaleTT:
### Problem description
TTNN JIT needs to support matmul

### What's changed
- Derive matmul output tensor and grid shapes in the AST frontend
  -  Both shape and grid follow the MxK * KxN -> MxN patten
- Derive the correct output tensor type for the top level `func.func` instead of assuming that it is the same as the input shapes
  - Create the `func.func` with a dummy return type originally to enable parsing and IR emitting for the ops
  - When processing the return statement, capture the derived output type for op that produces the return value
  - Create a new `func.func` with the captured return type, copy over the generated IR from the original `func.func` and delete it
- Some clean up in tensor processing code

### Checklist

### DIFF
--- a/test/ttnn-jit/nightly/test_matmul.py
+++ b/test/ttnn-jit/nightly/test_matmul.py
@@ -110,6 +110,7 @@ MATMUL_SHAPE_GRIDS = [
     ],
     ids=["bf16"],
 )
+@pytest.mark.parametrize("graph_capture", [False])
 def test_matmul_with_grids(device, shape_grids, dtype, ttnn_dtype, graph_capture):
     shapes, grids = shape_grids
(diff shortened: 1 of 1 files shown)


tools/ttnn-jit/_src/ttir_ast.py
```diff
@@ -143,6 +145,23 @@ def visit_FunctionDef(self, node):
                 self.visit(target)
 
         self.symbol_tables.pop()
+        # Create a new function type with the actual output type. Move the body from the old function to the new one.
```
vtangTT, line 148: lol..

PR #6154, opened 2025-12-03
[Bug Fix] Conv3d

vkovinicTT:
### Problem description
- conv3d op wasn't working with correct layouts for `weights`, `bias` and `output`

### What's changed
- set correct outputs for all those tensors

### DIFF
--- a/include/ttmlir/Dialect/TTNN/Transforms/Workarounds/Decomposition/Conv3dRewritePattern.h
+++ b/include/ttmlir/Dialect/TTNN/Transforms/Workarounds/Decomposition/Conv3dRewritePattern.h
@@ -1,0 +1,82 @@
+// SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
+//
+// SPDX-License-Identifier: Apache-2.0
+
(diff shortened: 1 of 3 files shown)


jserbedzijaTT:
I think it would be better to move this logic to `TTNNWorkaround` pass, see how it is done for conv2d op: `include/ttmlir/Dialect/TTNN/Transforms/Workarounds/Decomposition/Conv2dRewritePattern.h`

lib/Dialect/TTNN/Transforms/TTNNLayout.cpp
```diff
@@ -278,6 +278,11 @@ class TTNNLayoutRewriter : public OpInterfaceRewritePattern<ttir::TTIROp> {
       return mlir::isa<ttcore::TileType>(inputLayout.getElementType());
     }
 
+    // Conv3d produces ROW_MAJOR output at runtime (experimental op)
+    if (mlir::isa<ttir::Conv3dOp>(op)) {
+      return false;
+    }
+
```
jserbedzijaTT, line 285: Maybe move this also to workaround?

PR #6155, opened 2025-12-03
Fix builder test config for blackhole

gfengTT:
### Problem description
Builder tests need `--disable-eth-dispatch` flag when opening a BH device.

### DIFF
--- a/.github/test_scripts/builder.sh
+++ b/.github/test_scripts/builder.sh
@@ -15,7 +15,7 @@ PYTEST_ARGS=""
 FLATBUFFER=""
 
 [[ "$RUNS_ON" != "n150" ]] && PYTEST_ARGS="$PYTEST_ARGS --require-exact-mesh"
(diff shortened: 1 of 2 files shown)


wenbinlyuTT:
My two cents: `--disable-eth-dispatch` is used in a few different places and it would be a bit strange if some places have this "auto detect" while others don't. `ttrt query` is a very frequently hit example.

Maybe we can check `MeshDevice::ethernet_cores().size()` at the lowest level of our runtime and auto disable if the size is zero (and optionally print a warning). This approach is card ID agnostic and more natural: don't do eth dispatch when there is no eth cores...

gfengTT:
@wenbinlyuTT thanks for bringing it up. I agree it's a problem that `--disable-eth-dispatch` has to be manually added in a number of different places. In general we want that option whenever opening a Blackhole device. I'm not sure if `MeshDevice::ethernet_cores().size()` is the best option for that. If available, board type is the best indicator.

Issue #6025, opened 2025-11-25
[ttnn.jit] Error Out Gracefully on Unsupported Tensor Layouts


arminaleTT:
Ensure we do not crash. The error should be catch-able. Tests should verify the error

PR #6156, opened 2025-12-03
[ttnn.jit] Error Out Gracefully on Unsupported Tensor Layouts in TTNN JIT Frontend

brapananTT:
### Ticket
#6025 [ttnn.jit] Error Out Gracefully on Unsupported Tensor Layouts

### Problem description
We want to avoid crashes and catch any errors that result from invalid tensor layouts being passed into JIT'ed ops.

### What's changed
Wired up helper function` _check_layout_supported()` in `tensor_translator.py`. Added tests to verify unsupported tensor layouts are caught in `test/ttnn-jit/test_unsupported_tensor_layouts.py.`

### Checklist

### DIFF
--- a/test/ttnn-jit/test_unsupported_tensor_layouts.py
+++ b/test/ttnn-jit/test_unsupported_tensor_layouts.py
@@ -24,9 +24,10 @@ def exp(input_tensor_a):
 @pytest.mark.parametrize(
     "use_graph_capture", [True, False], ids=["graph_capture", "ast"]
 )
+@pytest.mark.skip(
+    reason="Error is raised from ttnn first, not JIT frontend.",
+)
 def test_l1_interleaved_not_supported(device, use_graph_capture):
-    if use_graph_capture:
-        pytest.skip("Using graph capture results in error from ttnn, not JIT frontend.")
 
     with pytest.raises(ValueError, match="Interleaved L1 tensors are not supported."):
         shape = (32, 32)
@@ -45,7 +46,7 @@ def test_l1_interleaved_not_supported(device, use_graph_capture):
             memory_config=memory_config,
         )
 
-        op_jit = ttnn_jit.jit(debug=True, graph_capture=use_graph_capture)(exp)
+        op_jit = ttnn_jit.jit(debug=False, graph_capture=use_graph_capture)(exp)
         output_tensor = op_jit(ttnn_tensor)
 
 
--- a/tools/ttnn-jit/_src/tensor_translator.py
+++ b/tools/ttnn-jit/_src/tensor_translator.py
@@ -196,7 +196,7 @@ def _check_layout_supported(tensor_arg):
             raise ValueError("Sharded DRAM tensors are not supported.")
 
     if mem_config.buffer_type.value == ttnn.BufferType.L1:
-        if mem_config.memory_layout != ttnn.TensorMemoryLayout.Interleaved:
+        if mem_config.memory_layout == ttnn.TensorMemoryLayout.Interleaved:
             raise ValueError("Interleaved L1 tensors are not supported.")
 
 


tools/ttnn-jit/_src/tensor_translator.py
```diff
@@ -212,39 +212,29 @@ def _create_dram_tensor_layout(ctx, tensor_arg):
 
 
 def _check_layout_supported(tensor_arg):
-    print(f"Checking layout supported for tensor: {tensor_arg}")
-    if tensor_arg.memory_config().is_sharded():
-        if tensor_arg.memory_config().shard_spec is None:
+    if str(tensor_arg.layout) != "Layout.TILE":
```
vtangTT: why do we have to stringify it, can we not check the enum?

tools/ttnn-jit/_src/tensor_translator.py
```diff
@@ -212,39 +212,29 @@ def _create_dram_tensor_layout(ctx, tensor_arg):
...
+        if mem_config.shard_spec is None:
             raise ValueError(
                 "Tensor is sharded but no legacy shard spec is present. ND Sharded tensors are not supported yet."
             )
 
-    if (
-        tensor_arg.memory_config().buffer_type == ttnn.BufferType.L1
-        and not tensor_arg.memory_config().is_sharded()
-    ):
+    if str(mem_config.buffer_type) == "BufferType.L1" and not mem_config.is_sharded():
```
vtangTT: same here, why stringify

test/ttnn-jit/test_unsupported_tensor_layouts.py
```diff
@@ -0,0 +1,100 @@
...
+    create_sharded_tile_tensor,
+    run_op_test,
+)
+
+
+def exp(input_tensor_a):
+    return ttnn.exp(input_tensor_a)
+
+
+@pytest.mark.parametrize("use_graph_capture", [True, False])
```
vtangTT: prob only need to check one of ast or graph_capture if they're running through the same code path ithink?

tools/ttnn-jit/_src/tensor_translator.py
brapananTT: I changed this initially because the original check (e.g. `mem_config.buffer_type == ttnn.BufferType.DRAM`) was working incorrectly. I can replace this with: `int(mem_config.buffer_type) == ttnn.BufferType.DRAM.value` for buffer type, and something like this for layout check: `int(tensor_arg.layout) == ttnn.Layout.TILE.value` although not 100% if `ttnn.Layout.TILE` is the right enum name

tools/ttnn-jit/_src/tensor_translator.py
brapananTT: see above

test/ttnn-jit/test_unsupported_tensor_layouts.py
brapananTT: My thought was that because graph capture will return back TT-FATAL / THROW on ttnn illegal configurations whereas ast will not, using ast / graph capture will cause differences of where an error will appear from (either from ttnn / metal or from jit frontend)

tools/ttnn-jit/_src/tensor_translator.py
vtangTT: `tensor_arg.layout == ttnn.TILE_LAYOUT` should work, `ttnn.Layout.Tile` should also work (according to what i see in metal)

`mem_config.buffer_type == ttnn.BufferType.DRAM` is also the way we do it in metal so unsure why it's not working..

Would rather cast to int then stringify if it doesn't work normally

brapananTT: > `mem_config.buffer_type == ttnn.BufferType.DRAM` is also the way we do it in metal so unsure why it's not working..

TY, I'll try this again

brapananTT: This works: `mem_config.buffer_type.value == ttnn.BufferType.DRAM`

vtangTT: probably some funkiness going on w/ nanobind/pybind type mixing... idk..

tools/ttnn-jit/_src/tensor_translator.py
```diff
@@ -180,23 +180,24 @@ def _create_dram_tensor_layout(ctx, tensor_arg):
...
-    ):
-        raise ValueError("Interleaved L1 tensors are not supported.")
-
-    if (
-        tensor_arg.memory_config().buffer_type == ttnn.BufferType.DRAM
-        and tensor_arg.memory_config().is_sharded()
-    ):
-        raise ValueError("DRAM tensors must be interleaved.")
+    if mem_config.buffer_type.value == ttnn.BufferType.L1:
+        if mem_config.memory_layout != ttnn.TensorMemoryLayout.Interleaved:
```
arminaleTT: I think this needs to be `==` instead?

vtangTT (approved):

Issue #6157, opened 2025-12-03
The UpdateGlobalToLocalShapes pass fails to update the sdy CCL ops' results to local shape in certain cases.


hshahTT:
### Describe the bug

When the UpdateGlobalToLocalShapes pass encounters a graph that has the output of a Shardy CCL op being fed as input to another Shardy CCL op, the pass fails with the following error: "number of output elements doesn't match expected number of elements". This pattern can be seen when running multi-batch inference on some models such as Llama 3 and Qwen 2.5.

For example, the following MLIR graph will fail due to the output of `sdy.all_reduce` being used as an input to `sdy.all_slice`:

```
sdy.mesh @mesh = <["x"=2, "y"=4]>

func.func @sharded_matmul(%arg0: tensor<544x8192xbf16> {sdy.sharding = #sdy.sharding<@mesh, [{"x"}, {}]>}, %arg1: tensor<8192x1024xbf16> {sdy.sharding = #sdy.sharding<@mesh, [{"x"}, {"y"}]>}) -> (tensor<544x1024xbf16> {sdy.sharding = #sdy.sharding<@mesh, [{"x"}, {"y"}]>}) {
    %0:4 = sdy.manual_computation(%arg0, %arg1) in_shardings=[<@mesh, [{"x"}, {}]>, <@mesh, [{"x"}, {"y"}]>] out_shardings=[<@mesh, [{}, {"x"}]>, <@mesh, [{}, {"y"}]>, <@mesh, [{}, {"y"}]>, <@mesh, [{"x"}, {"y"}]>] manual_axes = {} (%arg2: tensor<544x8192xbf16>, %arg3: tensor<8192x1024xbf16>) {
        %1 = sdy.all_to_all [{"x"}: 0->1] %arg2 out_sharding=<@mesh, [{}, {"x"}]> : tensor<544x8192xbf16>
        %2 = stablehlo.dot_general %1, %arg3, contracting_dims = [1] x [0] {sdy.sharding = #sdy.sharding_per_value<[<@mesh, [{}, {"y"}]>]>} : (tensor<544x8192xbf16>, tensor<8192x1024xbf16>) -> tensor<544x1024xbf16>
        %3 = sdy.all_reduce {"x"} %2 out_sharding=<@mesh, [{}, {"y"}]> : tensor<544x1024xbf16>
[...]

PR #6158, opened 2025-12-03
Account for Shardy CCL ops in `updateShapes()` in the UpdateGlobalToLocalShapes pass

hshahTT:
### Ticket
Fixes [6157](https://github.com/tenstorrent/tt-mlir/issues/6157)

### Problem description
The `updateShapes()` function in the UpdateGlobalToLocalShapes pass currently fails to update shape of the result tensors of Shardy CCL ops, which causes the pass to fail in certain scenarios (see attached issue for more context).

### What's changed
- Fixed the `updateShapes()` function to produce the correct local shape for Shardy CCL ops.

### Checklist

### DIFF
--- a/lib/Dialect/StableHLO/Transforms/UpdateGlobalToLocalShapes.cpp
+++ b/lib/Dialect/StableHLO/Transforms/UpdateGlobalToLocalShapes.cpp
@@ -267,7 +267,8 @@ static mlir::LogicalResult updateManualAxes(MLIRContext *context,
   return mlir::failure(result.wasInterrupted());
 }
 
-// Update all shapes in the module based on their sdy tensor sharding attribute.
+// Update all shapes in the module based on their sdy tensor sharding attribute,
+// and out_sharding attribute (for sdy CCL ops).
 static mlir::LogicalResult updateShapes(MLIRContext *context,
                                         mlir::OpBuilder &builder,
                                         mlir::sdy::MeshOp &globalMeshOp,
@@ -283,6 +284,8 @@ static mlir::LogicalResult updateShapes(MLIRContext *context,
       llvm::SmallVector<mlir::sdy::TensorShardingAttr> tensorShardings;
 
       if (namedAttr.getName() == outShardingAttrName) {
+        // For sdy CCL ops, we need to update the output shapes based on the
+        // out_sharding
         mlir::sdy::TensorShardingAttr tensorShardingAttr =
             mlir::cast<mlir::sdy::TensorShardingAttr>(namedAttr.getValue());
         tensorShardings.assign({tensorShardingAttr});
--- a/test/ttmlir/Dialect/StableHLO/shardy/update_sdy_shapes.mlir
+++ b/test/ttmlir/Dialect/StableHLO/shardy/update_sdy_shapes.mlir
@@ -1,0 +1,49 @@
+// REQUIRES: stablehlo
+// This file incorporates work covered by the following copyright and permission notice:
+// SPDX-FileCopyrightText: Copyright (c) 2024 The Shardy Authors
+// SPDX-License-Identifier: Apache-2.0
+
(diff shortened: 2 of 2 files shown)


tapspatel (commented):
aren't sdy ccls supposed to be global shapes? once we convert to stablehlo ccls, those are local shapes

sdy ccls shouldn't exist under manual computation op that has manual axes defined (it can if the manual axes isn't defined)

tapspatel:
looking at the issue, we need to fix the pattern rewriter, I don't remember exactly but we should use the adaptor to access the op elements and update from there

hshahTT:
> sdy ccls shouldn't exist under manual computation op that has manual axes defined (it can if the manual axes isn't defined)

At the point we call `updateShapes()` and the Sdy->SHLO ccl ops conversion method the manual axes aren't defined, so this should be valid.

> aren't sdy ccls supposed to be global shapes? once we convert to stablehlo ccls, those are local shapes

> looking at the issue, we need to fix the pattern rewriter, I don't remember exactly but we should use the adaptor to access the op elements and update from there

Yes, but the SHLO rewrite patterns seem to expect the Sdy CCL operands to be in a local shape, which can't happen in the sdy CCL followed by sdy CCL case if we use bottom-up traversal. 

So we need to either make it so that every operand to any given sdy CCL op is in local shape (which is what this PR currently does), or we need to change the rewrite patterns to accept either local or global shaped operands, and in the case of a global operand, have it convert to local shape during the rewrite. 

I think to get that to work we'd have to annotate the result of a Sdy CCL op with a `sdy.sharding` and account for any `sdy.sharding` within the rewrite pattern itself.

Any thoughts on this?

sshonTT:
To add a bit more context with example,
```
%91 = "stablehlo.dot_general"(%87, %90)
      <{dot_dimension_numbers = #stablehlo.dot<
        lhs_contracting_dimensions = [1],
        rhs_contracting_dimensions = [0]
      >}>
      {sdy.sharding = #sdy.sharding_per_value<[<@mesh, [{"_axis_0", ?}, {?}]>]>}
      : (tensor<512x1024xbf16>, tensor<1024x4096xbf16>) -> tensor<512x4096xbf16>

%92 = "sdy.all_reduce"(%91)
      <{out_sharding = #sdy.sharding<@mesh, [{"_axis_0", ?}, {?}]>,
        reduction_axes = #sdy<axis_ref_list{"_axis_1"}>
      }>
      : (tensor<512x4096xbf16>) -> tensor<1024x4096xbf16>

%93 = "sdy.all_slice"(%92)
      <{out_sharding = #sdy.sharding<@mesh, [{"_axis_0"}, {"_axis_1"}]>,
        slicing_axes = #sdy<list_of_axis_ref_lists[{}, {"_axis_1"}]>
      }>
      : (tensor<1024x4096xbf16>) -> tensor<1024x4096xbf16>
```

This sequence of sdy CCL ops is what gets generated due to the reshard. These are the ops right after updateShapes, but before convertShardyCCLToStableHLOCCL is applied.

As you can see, the input to sdy.all_reduce (%91) is correctly in local shape, but the following sdy CCL ops (sdy.all_reduce and then sdy.all_slice) do not have the correct local shapes reflected in their types. Because of that, the later lowering step receives incorrect shapes and fails.

In other words, the issue is not with the propagation into the first CCL, but with the fact that all subsequent sdy CCL ops still have global-shaped result types, and updateShapes never normalizes them into local-shaped types before we run the Stablehlo CCL conversion. That mismatch eventually causes the lowering to crash.

hshahTT:
I created this PR instead: https://github.com/tenstorrent/tt-mlir/pull/6421 which just uses top-down traversal for the pattern rewriter. I can later look into a better way to fix updating the shapes, but I wanted to get this in so it unblocks the Llama 70B in Galaxy demo.

PR #6160, opened 2025-12-03
[tt-lang] Python bindings for ttnn interop

zoecarver:
Add bindings for tt-lang's ttnn-interop, see PR: https://github.com/tenstorrent/tt-lang/pull/50

### DIFF
--- a/python/Passes.cpp
+++ b/python/Passes.cpp
@@ -422,10 +422,10 @@ void populatePassesModule(nb::module_ &m) {
           if (auto threadTypeAttr =
                   funcOp->getAttrOfType<mlir::tt::ttkernel::ThreadTypeAttr>(
                       mlir::tt::ttkernel::ThreadTypeAttr::name)) {
-            kernels.emplace_back(
-                funcOp.getName().str(),
-                mlir::tt::ttkernel::stringifyThreadType(threadTypeAttr.getValue())
-                    .str());
+            kernels.emplace_back(funcOp.getName().str(),
+                                 mlir::tt::ttkernel::stringifyThreadType(
+                                     threadTypeAttr.getValue())
+                                     .str());
           }
         });
         return kernels;


python/Passes.cpp
```diff
@@ -384,6 +385,82 @@ void populatePassesModule(nb::module_ &m) {
...
+      },
+      nb::arg("module"), nb::arg("symbol_name"));
+
+  m.def(
+      "get_ttkernel_names",
+      [](MlirModule module) {
+        mlir::Operation *moduleOp = unwrap(mlirModuleGetOperation(module));
+        auto mod = mlir::cast<ModuleOp>(moduleOp);
+
+        std::vector<std::tuple<std::string, std::string>> kernels;
```
tapspatel, line 420: comment on what this data structure contains

python/Passes.cpp
```diff
@@ -384,6 +385,82 @@ void populatePassesModule(nb::module_ &m) {
...
+        mod.walk([&](func::FuncOp funcOp) {
+          if (auto threadTypeAttr =
+                  funcOp->getAttrOfType<mlir::tt::ttkernel::ThreadTypeAttr>(
+                      mlir::tt::ttkernel::ThreadTypeAttr::name)) {
+            std::string threadType;
+            switch (threadTypeAttr.getValue()) {
+            case mlir::tt::ttkernel::ThreadType::Noc:
```
tapspatel: I believe we already have this autogenerated for us: `build/include/ttmlir/Dialect/TTKernel/IR/TTKernelOpsEnums.cpp.inc`

```
::llvm::StringRef stringifyThreadType(ThreadType val) {
  switch (val) {
    case ThreadType::Noc: return "noc";
    case ThreadType::Compute: return "compute";
    case ThreadType::Ethernet: return "ethernet";
  }
  return "";
}
```

zoecarver:
Both comments addressed. Thank you @tapspatel!

Issue #6048, opened 2025-11-26
Default layout for const-evaled args should be system_memory


odjuricicTT:
Currently we default to dram interleaved tile for most weights / inputs. This causes a lot of OOM problems for bigger models since once the input is const-evaled we are keeping both the original and const-evaled version in dram, greatly increasing memory requirements for the model. More details here: https://github.com/tenstorrent/tt-xla/issues/1927

In order to optimize this one approach is to find all args that inputs to const-eval functions and default them to system memory and add an explicit to_device and to_layout if needed.

PR #6166, opened 2025-12-04
Enforce host layout for const-eval functions' inputs

dmilinkovicTT:
### Ticket
#6048

### Problem description
Const-eval inputs are placed in DRAM by default. Since const-eval results are also stored in DRAM, this effectively means that before the original inputs is deallocated, memory footprint is twice the required size.

### What's changed
- Force const-eval function arguments to system memory
- Force function arguments which get consumed by `load_cached` ops to system memory
- Extend logic for handling `ttir.mesh_op` layout
- Update tests

### Checklist

### DIFF
--- a/runtime/test/ttnn/python/utils.py
+++ b/runtime/test/ttnn/python/utils.py
@@ -100,7 +100,9 @@ class ProgramTestRunner:
         if self.config.compute_golden:
             golden = self.config.compute_golden(inputs_torch)
 
-        return inputs_runtime_with_layout, golden
+        # Returnining inputs_torch to prevent Python GC to free memory of the Torch tensors,
+        # which can lead to invalid memory access if the inputs are expected in the host memory.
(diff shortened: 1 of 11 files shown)


runtime/test/ttnn/python/utils.py
```diff
@@ -100,7 +100,9 @@ def get_inputs_and_golden(self, device, borrow=True):
         if self.config.compute_golden:
             golden = self.config.compute_golden(inputs_torch)
 
-        return inputs_runtime_with_layout, golden
+        # Returnining inputs_torch to prevent Python GC to free memory of the Torch tensors,
+        # which can lead to invalid memory access if the inputs are expected in the host memory.
```
dmilinkovicTT, line 104: pretty hacky, but not sure how else could I handle the issue. 

also tried `Owned` instead of `Borrowed` storage, but got similar results

PR #6168, opened 2025-12-04
Row-major propagation for integer inputs

rpavlovicTT:
Implement TTNNRowMajorLayoutPropagation pass that tracks tensors from function
input arguments through the dataflow graph and maintains row-major layout until
operations explicitly require tile layout (e.g., matmul, conv). This optimization
eliminates redundant layout conversions by:

- Identifying row-major function inputs and removing redundant ToLayoutOps
- Propagating row-major layout through the computation graph using validation
- Stopping propagation at operations that require or produce tile layout, or are in-place ops.
- Updating OpConfigAnalysis to prefer matching existing layout attributes

For now, row-major propagation is done only for integer types, since that is helpful in LLMs. 
CNNs can have perf degradation if input images are kept in row-major, so we are not doing that yet.

Fixed untilize/tilize constraints in TTNNDecomposeLayouts pass wrt. tt-metal.

Disabled the hoisted_equal test and created issue https://github.com/tenstorrent/tt-mlir/issues/6183 to track
investigation and potential solutions.

Remove old bf_interleaved_policy and greedy_l1_interleaved_policy tests
that are no longer maintained. These tests were using deprecated optimizer
policies and are not actively used.


odjuricicTT (approved):

test/ttmlir/Silicon/TTNN/n150/Transforms/DecomposeLayouts/decomposing_layouts_from_host.mlir
```diff
@@ -115,14 +115,15 @@ module attributes {} {
         return %0 : tensor<64x128xbf16, #ttnn_layout_device_rm_bf16>
     }
 
-    // Test case when we move tensor from host to device for tile -> row-major case.
+    // Test case when we move tensor from host to device for tile -> row-major case for f32 data type.
```
sdjordjevicTT, line 118: For the sake of test coverage, now that we have changed a couple of tests that tilize/utilize on device, do we have at least one test case when the tilize/untilize is done on the host?

rpavlovicTT: Yes there are still such tests, ones named with prefix `from_host_to_host...`

PR #6169, opened 2025-12-04
[Optimizer] Temporary workaround for divide-by-zero in matmul kernel

mvasiljevicTT:
### Ticket
[Link to Github Issue](https://github.com/tenstorrent/tt-forge/issues/710)

### Problem description
When running the segformer model on the latest tt-xla with optimizer enabled, compilation crashes in the `TTNNOptimizer` pass while computing constraints for a `MatmulOp`.

The root cause is that the matmul kernel currently uses the logical shape instead of the padded shape to calculate the number of tiles for block-sharded input. When the last dimension is less than 32, this results in 0 tiles, leading to a divide-by-zero floating-point exception.

### What's changed
This PR introduces a temporary workaround to prevent crashes. 

It adds additional validation to `MatmulOp::getOpConstraints` in `TTNNOpModelInterface`, along with a new test that verifies the fix has not yet been uplifted in tt-metal. 

The test currently expects the matmul operation to crash. When it starts failing, that will indicate that [this fix on tt-metal](https://github.com/tenstorrent/tt-metal/commit/c579ac2ad7b1df0ab4ab35f308eb1e67e43751ca) is uplifted and this workaround should be removed.

### Checklist

### DIFF
--- a/test/unittests/OpModel/TTNN/Op/TestMatmulBlockShardedConstraint.cpp
+++ b/test/unittests/OpModel/TTNN/Op/TestMatmulBlockShardedConstraint.cpp
@@ -48,8 +48,8 @@ public:
 // IMPORTANT: If this test FAILS (EXPECT_DEATH fails because code doesn't
 // crash), it means the metal issue
 // (https://github.com/tenstorrent/tt-metal/pull/33777) is fixed and uplifted.
-// In that case, all changes from this PR should be reverted (the check in
-// MatmulOp::getOpConstraints and this test).
+// In that case, all changes from this PR should be reverted:
+// https://github.com/tenstorrent/tt-mlir/pull/6169
 TEST_F(OpModelTest, MatmulBlockShardedInputWithPadding) {
   llvm::SmallVector<int64_t> inputShapeA = {4096, 16};
   TTNNLayoutAttr inputLayoutA = CreateTiledLayout(


lib/Dialect/TTNN/Validation/OpConstraintValidation.cpp
```diff
@@ -104,6 +105,35 @@ validateConstraints(Operation *op, llvm::ArrayRef<TTNNLayoutAttr> inputLayouts,
             .concat(op->getName().getStringRef()));
   }
 
+  // Skip getOpConstraints for matmul with block sharded first operand if height
+  // or width is less than tile size
+  // TODO (mvasiljevic): remove once the fix in metal is merged and uplifted:
+  // https://github.com/tenstorrent/tt-metal/pull/33777
+  if (mlir::isa<ttnn::MatmulOp>(op) && !inputLayouts.empty() &&
```
odjuricicTT: Can you put this logic inside matmul constraints API? e.g. https://github.com/tenstorrent/tt-mlir/blob/main/lib/Dialect/TTNN/Interfaces/TTNNOpModelInterface.cpp#L2898

mvasiljevicTT: Sure, changed now

test/unittests/OpModel/TTNN/Op/TestMatmulBlockShardedConstraint.cpp
```diff
@@ -0,0 +1,86 @@
...
+};
+
+// Test that bypassing the check in MatmulOp::getOpConstraints causes a crash.
+// The check prevents crashes by returning early for block-sharded matmul inputs
+// with logical shape less than tile size.
+//
+// IMPORTANT: If this test FAILS (EXPECT_DEATH fails because code doesn't
+// crash), it means the metal issue
+// (https://github.com/tenstorrent/tt-metal/pull/33777) is fixed and uplifted.
+// In that case, all changes from this PR should be reverted (the check in
```
odjuricicTT: Link to this PR, that should be reverted.

odjuricicTT (approved):
Awesome!

test/unittests/OpModel/TTNN/Op/TestMatmulBlockShardedConstraint.cpp
mvasiljevicTT: added

Issue #6170, opened 2025-12-04
Distributed runtime API parity


jameszianxuTT:
Many distributed runtime APIs are not implemented - GetTensorDesc, HasLayout, ClearProgramCache etc... that the frontends may use.

Implement these. It should be mostly boilerplate, some small custom ser/des

PR #6171, opened 2025-12-04
Create distributed API for GetTensorDesc

jameszianxuTT:
### Ticket
Part of #6170 

### Problem description
::tt::runtime::distributed::getTensorDesc not implemented.

### What's changed
- Add flatbuffer ser/des for tt::runtime::TensorDesc.
- Add distributed runtime implementation for GetTensorDesc, which should return a single tensor desc
- Add testcase

### Checklist
- [x] Add runtime pytest runtime/test/ttnn/python/distributed/llmbox/test_multiprocess.py::test_GetTensorDesc

### DIFF
--- a/runtime/lib/distributed/runtime.cpp
+++ b/runtime/lib/distributed/runtime.cpp
@@ -194,7 +194,7 @@ void deallocateTensor(::tt::runtime::Tensor &tensorHandle, bool force) {
   ControllerSingleton::get().deallocateTensor(tensorHandle, force);
 }
 
-::tt::runtime::TensorDesc getTensorDesc(::tt::runtime::Tensor t){
+::tt::runtime::TensorDesc getTensorDesc(::tt::runtime::Tensor t) {
   assertControllerLaunched();
   return ControllerSingleton::get().getTensorDesc(t);
 }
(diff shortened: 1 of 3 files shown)


jameszianxuTT:
I have diagrammed what I've considered the "idiomatic" way to implement a distributed runtime API.

distributed-runtime-arch drawio

jameszianxuTT:
Please take a look at this PR @acolicTT and @pilkicTT as we do not have any other reviewers with familiarity to distributed runtime.

I've tried to make this easy to understand from what Jackson transferred over. The diagram in above comment hopefully helps with understanding all the boilerplate.

I will also put up #6206 on top of this PR which includes the two program cache related APIs, with very similar implementation to this one. That should unblock your PR @acolicTT  https://github.com/tenstorrent/tt-xla/pull/2404, if you do not wish to special-case the distributed runtime test.

runtime/lib/distributed/controller/controller.cpp
```diff
@@ -678,6 +665,32 @@ void Controller::deallocateTensor(::tt::runtime::Tensor &tensorHandle,
                                  std::move(commandBuilder));
 }
 
+::tt::runtime::TensorDesc
+Controller::getTensorDesc(const ::tt::runtime::Tensor &tensorHandle) {
+  auto commandBuilder = std::make_unique<::flatbuffers::FlatBufferBuilder>();
+
+  uint64_t commandId =
+      CommandFactory::buildGetTensorDescCommand(*commandBuilder, tensorHandle);
```
pilkicTT, line 673: we should consider modifying `CommandFactory::buildGetTensorDescCommand()` to just return a `std::pair<Command, CommandId>` (or something to that extent) - so that the users of the command factory don't need to create a fb builder... it seems that this pattern repeats in this code, for example: `pushToCommandAndResponseQueue()` should probably just return the future and not accept the future and promise...

Issue #5449, opened 2025-10-22
[D2M] add verifier to MetalLayoutAttr


vwellsTT:
### Is your feature request related to a problem? Please describe

currently, there are a lot of rules for MetalLayoutAttr (`dim_alignments` must cover logical shape, collapsed_intervals should have sensical values, etc), but we have no verifier.  This results in some mildly difficult to debug failures if a misformed layout is created, etc.

### Describe the solution you'd like

We should add a basic verifier that checks any conditions that can be checked in isolation.  It's worth noting that a fair number of issues cannot be check, since anything which requires a mix of the actual tensor shape and this attr is not verifiable in isolation.  However, we can ensure that any relationships intra-layout are well-formed.

PR #6174, opened 2025-12-04
[D2M] Add simple verifier to MetalLayoutAttr

vwellsTT:
### Ticket
https://github.com/tenstorrent/tt-mlir/issues/5449

### Problem description
When manually creating MetalLayoutAttrs, it's easy to violate assumptions which are central to its usage.

### What's changed
Add basic verification s.t. malformed MetalLayoutAttrs get rejected.  Worth noting that we cannot verify anything related to the actual tensor shape (so if a MetalLayoutAttr is attached a tensor with a mismatched shape, we can't catch that here), but we can check that the fields of the attr itself are all internally consistent.  Add a negative lit test to ensure coverage too.

### Checklist

### DIFF
--- a/lib/Dialect/TTCore/IR/TTCoreOpsTypes.cpp
+++ b/lib/Dialect/TTCore/IR/TTCoreOpsTypes.cpp
@@ -1165,40 +1165,36 @@ MetalLayoutAttr MetalLayoutAttr::get(::mlir::MLIRContext *context,
 
   // Logical shape must have at least 2 dimensions for tiling.
   if (logicalRank < 2) {
-    emitError() << "logical_shape must have at least 2 dimensions, got "
-                << logicalRank;
-    return ::mlir::failure();
+    return emitError() << "logical_shape must have at least 2 dimensions, got "
+                       << logicalRank;
   }
 
   // The dim_alignments rank must match logical_shape rank.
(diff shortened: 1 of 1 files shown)


lib/Dialect/TTCore/IR/TTCoreOpsTypes.cpp
```diff
@@ -1153,6 +1153,93 @@ MetalLayoutAttr MetalLayoutAttr::get(::mlir::MLIRContext *context,
...
+    MemorySpace memorySpace, TensorMemoryLayout memoryLayout,
+    ::mlir::AffineMap indexAffineMap) {
+
+  int64_t logicalRank = logicalShape.size();
+
+  // Logical shape must have at least 2 dimensions for tiling.
+  if (logicalRank < 2) {
+    emitError() << "logical_shape must have at least 2 dimensions, got "
+                << logicalRank;
+    return ::mlir::failure();
```
azecevic000: Nit: `InFlightDiagnostic` has an implicit cast to `LogicalResult`, so you can write `return emitError() << ...;`

vwellsTT: Oh, neat, thanks!

Issue #5955, opened 2025-11-20
Slice folding


mvasiljevicTT:
Swin model has consecutive slices that could be folded as one slice:
```
        %361 = "ttnn.slice_static"(%360) <{begins = [0 : i32, 0 : i32, 0 : i32, 0 : i32], ends = [1 : i32, 128 : i32, 133 : i32, 96 : i32], step = [1 : i32, 1 : i32, 1 : i32, 1 : i32]}> : (tensor<1x133x133x96xbf16, #ttnn_layout182>) -> tensor<1x128x133x96xbf16, #ttnn_layout197> loc(#loc1116)
        "ttnn.deallocate"(%360) <{force = false}> : (tensor<1x133x133x96xbf16, #ttnn_layout182>) -> () loc(#loc1116)
        %362 = "ttnn.slice_static"(%361) <{begins = [0 : i32, 0 : i32, 0 : i32, 0 : i32], ends = [1 : i32, 128 : i32, 128 : i32, 96 : i32], step = [1 : i32, 1 : i32, 1 : i32, 1 : i32]}> : (tensor<1x128x133x96xbf16, #ttnn_layout197>) -> tensor<1x128x128x96xbf16, #ttnn_layout101> loc(#loc1117)
        "ttnn.deallocate"(%361) <{force = false}> : (tensor<1x128x133x96xbf16, #ttnn_layout197>) -> () loc(#loc1117)
```

odjuricicTT:
@HalfBloodPrince010 Will pick this up!

HalfBloodPrince010:
Hi @mvasiljevicTT, good day! Could you please share a bit more context on the issue above so I can start looking into it? Thank you!

mvasiljevicTT:
Sure!

[SliceStaticOp](https://github.com/tenstorrent/tt-mlir/blob/8b2e9321d7e8c1d666b94dda1d27017213eae8cc/include/ttmlir/Dialect/TTIR/IR/TTIROps.td#L3827) is defined with three arrays, begins, ends, and steps, which specify slicing per dimension.
For a 4-dimensional tensor, the op:
```
slice_static(arr)
  <begins = [b1, b2, b3, b4],
   ends   = [e1, e2, e3, e4],
   steps  = [s1, s2, s3, s4]>
```
is equivalent to the following Python slice notation:
```
arr[b1:e1:s1, b2:e2:s2, b3:e3:s3, b4:e4:s4]
```

In this example from swin model we have:
```
[...]

PR #6182, opened 2025-12-05
Add folding for consecutive SliceStaticOp

HalfBloodPrince010:
**Motivation**
Swin model has consecutive slices that could be folded as one slice.

**Testing**

Added a test to validate redundant intermediate
slices are removed.

### Ticket
[Link to Github Issue](https://github.com/tenstorrent/tt-mlir/issues/5955)

### Problem description
Back to back Producer-Consumer SliceStaticOp
can be Folded by composing their begin/end/step
attributes into a single Op and rewiring the
consumer SliceStaticOp to slice directly from
the original produce input.

### What's changed

Consecutive Producer-Consumer SliceStaticOp
will be Folded into single SliceStaticOp.

Used the [folding pattern](https://mlir.llvm.org/docs/Canonicalization/)  do the Consecutive fold.
Updated Tablegen to indicate the SliceStaticOp has fold method by defining
```hasFolder = 1```.

### Example
```
input tensor shape: <1x8x8x4>
   |
slice 1 - <{begins = [0, 1, 0, 0], ends = [1, 8, 8, 4], steps = [1, 2, 1, 1]}> producing shape <1x4x8x4>
   |
slice 2 - <{begins = [0, 0, 0, 0], ends = [1, 4, 8, 4], steps = [1, 1, 2, 1]}> , producing shape <1x4x4x4>
```

After Slice 1, indices will be

```
Dim 0:  {0} 
Dim 1:  {1, 3, 5, 7}
Dim 2: {0, 1, 2, 3, 4, 5, 6, 7}
Dim 3: {0, 1, 2, 3}

```

they could be merged as one slice:

```
b_i = b1_i + s1_i * b2_i
e_i = b1_i + s1_i * (e2_i - 1) + 1
s_i = s1_i * s2_i
```

Using the above formula begins will be
```
Dim 0: b_0 = 0 + 1 * 0 = 0 
Dim 1: b_1 = 1 + 2 * 0 = 1 
Dim 2: b_2 = 0 + 1 * 0 = 0 
Dim 3: b_3 = 0 + 1 * 0 = 0
```

Ends will be
```
Dim 0: e_0 = 0 + 1 * (1-1) + 1 = 1 
Dim 1: e_1 = 1 + 2 * (4 - 1) + 1 = 8 
Dim 2: e_2 = 0 + 1 * (8 - 1) + 1 = 8 
Dim 3: e_3 = 0 + 1 * (4 - 1) + 1 = 4
```

### Note: The reason to do  (e2_i - 1) in the Ends formula is if ```producerOp``` has step > 1 and begin > 0, then we don't want it to overshoot. For instance if we didn't do ```-1```, then

```1 + 4 * 2 = 9```, which is incorrect

and Steps 
```
Dim 0: s_0 = 1 * 1 = 1
Dim 1: s_1 = 2 * 1 = 2
Dim 2: s_2 = 1 * 2 = 2
Dim 3: s_3 = 1 * 1 = 1
```

Merged slice on the original tensor by updating the begins, ends and step like above
```
output = slice_static(arr) 
<begins = [0, 1, 0, 0],
 ends = [1, 8, 8, 4],
 steps = [1, 2, 2, 1]>
```

### Checklist

### CODE REVIEW

**Comment by mvasiljevicTT** [comment]
Thank you for doing this!  
Could you please add tests where begins are not all zeros and steps are not all 1, and one where the first slice has other users.

**Comment by HalfBloodPrince010** [comment]
Sounds good, added a test case to cover multiple producer uses and also different begins, ends, and steps.

File: lib/Dialect/TTIR/IR/TTIROps.cpp [hunk 1]
```diff
@@ -1932,6 +1932,72 @@ ::mlir::LogicalResult mlir::tt::ttir::SliceStaticOp::verify() {
...
+      int32_t b1_i = mlir::cast<::mlir::IntegerAttr>(begins1[i]).getInt();
+      int32_t s1_i = mlir::cast<::mlir::IntegerAttr>(stepAttr1[i]).getInt();
+
+      // ith dimension in Consumer begins, end and step arrays
+      int32_t b2_i = mlir::cast<::mlir::IntegerAttr>(begins2[i]).getInt();
+      int32_t e2_i = mlir::cast<::mlir::IntegerAttr>(ends2[i]).getInt();
+      int32_t s2_i = mlir::cast<::mlir::IntegerAttr>(stepAttr2[i]).getInt();
+
+      int32_t b_i = b1_i + s1_i * b2_i;
+      int32_t e_i = b1_i + s1_i * e2_i;
```
mvasiljevicTT: After more careful calculation I see that it should be  `e_i = b1_i + s1_i * (e2_i - 1) + 1`, could you please update that?

test/ttmlir/Dialect/TTIR/canonicalize/slice_static_op_fold_tests.mlir
```diff
@@ -0,0 +1,28 @@
...
+    // CHECK-SAME: step = [2 : i32, 2 : i32, 2 : i32]
+    // CHECK: "ttir.slice_static"
+    // CHECK-SAME: begins = [0 : i32, 2 : i32, 2 : i32]
+    // CHECK-SAME: ends = [64 : i32, 64 : i32, 64 : i32]
+    // CHECK-SAME: step = [1 : i32, 2 : i32, 2 : i32]
+    %1 = "ttir.slice_static"(%arg0) <{begins = [0 : i32, 0 : i32, 0 : i32], ends = [128 : i32, 128 : i32, 128 : i32], step = [2 : i32, 2 : i32, 2 : i32]}> : (tensor<128x128x128xbf16>) -> tensor<64x64x64xbf16>
+    %2 = "ttir.slice_static"(%1) <{begins = [0 : i32, 2 : i32, 2 : i32], ends = [64 : i32, 64 : i32, 64 : i32], step = [1 : i32, 2 : i32, 2 : i32]}> : (tensor<64x64x64xbf16>) -> tensor<64x31x31xbf16>
+    %3 = ttir.empty() : tensor<64x64x64xbf16>
+    %4 = "ttir.permute"(%1, %3) <{permutation = array<i32: 2, 1, 0>}> : (tensor<64x64x64xbf16>, tensor<64x64x64xbf16>) -> tensor<64x64x64xbf16>
+    return %2: tensor<64x31x31xbf16>
```
mvasiljevicTT: Could you please divide second test into two tests - one with:
- Slice 1: 1x128x128x96 → 1x64x128x96, begins [0,1,0,0], ends [1,128,128,96], steps [1,2,1,1]
- Slice 2: 1x64x128x96 → 1x64x64x96, begins [0,0,0,0], ends [1,64,128,96], steps [1,1,2,1]

that does fold, and the other one where producer has multiple users so it doesn't fold.

In this test you don't use permute op anywhere later, so it will be removed from the graph, making producer slice still having only one user. You also don't need `ttir.empty()` for permute, as we don't use dps in ttir anymore. Permutation has to be i64 array.

lib/Dialect/TTIR/IR/TTIROps.cpp
HalfBloodPrince010: got it, understood now, if the produce step is >1 and begin > 0, it will overshoot in the end.

In the below example, if we use ```e_i = b1_i + s1_i * e2_i;```

Then it become, 1 + 2 * 64 = 129. Thank you. 

I will update e_i.

test/ttmlir/Dialect/TTIR/canonicalize/slice_static_op_fold_tests.mlir
HalfBloodPrince010: Sound good, I will break it into 2 tests, and also update the permute case and return it so its not eliminated.

HalfBloodPrince010:
Note: Build failure not related to the above change: ```Could not find a package configuration file provided by "tsl-robin-map"```.  https://github.com/tenstorrent/tt-mlir/pull/6221

test/ttmlir/Dialect/TTIR/canonicalize/slice_static_op_fold_tests.mlir
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+  func.func @consecutive_slice_static_no_fold_multiple_producer_uses(%arg0: tensor<128x128x128xbf16>) -> (tensor<64 x 31 x 31 x bf16>, tensor<64 x 64 x 64 x bf16>) {
+    // CHECK: "ttir.slice_static"
+    // CHECK-SAME: begins = [0 : i32, 0 : i32, 0 : i32]
+    // CHECK-SAME: ends = [128 : i32, 128 : i32, 128 : i32]
+    // CHECK-SAME: step = [2 : i32, 2 : i32, 2 : i32]
+    // CHECK: "ttir.slice_static"
+    // CHECK-SAME: begins = [0 : i32, 2 : i32, 2 : i32]
+    // CHECK-SAME: ends = [64 : i32, 64 : i32, 64 : i32]
```
mvasiljevicTT, line 31: When you run tests with `llvm-lit -sv test/ttmlir/Dialect/TTIR/canonicalize/slice_static_op_fold_tests.mlir` you can see that the last one is failing.

lib/Dialect/TTIR/IR/TTIROps.cpp
```diff
@@ -1932,6 +1932,72 @@ ::mlir::LogicalResult mlir::tt::ttir::SliceStaticOp::verify() {
...
 
+// back to back producer-consumer SliceStaticOp can be folded
+// into a single SliceStaticOp.
+static mlir::OpFoldResult
+foldConsecutiveSliceStatic(mlir::tt::ttir::SliceStaticOp consumerOp) {
+
+  if (auto producerSliceOp =
+          consumerOp.getInput()
+              .getDefiningOp<mlir::tt::ttir::SliceStaticOp>()) {
+    mlir::RankedTensorType inputType = consumerOp.getInput().getType();
```
mvasiljevicTT, line 1949: You need to check that producerSliceOp doesn't have other users.

HalfBloodPrince010: Updated to check if producerOp, and fold only if it has single use.

HalfBloodPrince010:
Tests are passing now,

```
llvm-lit -sv test/ttmlir/Dialect/TTIR/canonicalize/slice_static_op_fold_tests.mlir

Testing Time: 2.44s

Total Discovered Tests: 1
  Passed: 1 (100.00%)
```

HalfBloodPrince010:
Thank you @mvasiljevicTT for the detailed pointers, helped me understand a lot of things!

HalfBloodPrince010:
Hi @mvasiljevicTT , could you help merge, I updated the branch. Not sure if tests need to be re-run again.

azecevicTT (approved):
LGTM, thank you for the contribution!

Issue #6093, opened 2025-12-01
Gather operation validation and fallback error on Swin


vkovacevicTT:
On [latest](https://github.com/tenstorrent/tt-forge/actions/runs/19811523944/job/56754891619) nightly in tt-forge Swin fails in perf benchmarks with error:

`loc("gather.2770"): error: OperationValidationAndFallback: Operation failed validation and no fallback configuration worked`

PR #6191, opened 2025-12-05
[Swin Regression] Fix output shape for embedding op in OpModel

bmalesevicTT:
### Ticket
[Swin regression OperationValidationAndFallback](https://github.com/tenstorrent/tt-forge/issues/711)
#6093 

### Problem description
Starting from 2025-11-29 nightly, some embedding ops (originally gather ops) began failing to find a valid configuration.
The regression appeared after uplifting this tt-metal change: https://github.com/tenstorrent/tt-metal/commit/a23fcef172d6a0b3f00a0c67b9646412e505717f. 
The issue was not in tt-metal itself.
Instead, OpModel was providing an incorrect output shape to the embedding op.
This caused an unexpected reshape inside tt-metal, leading to the validation failure.
### What's changed
- Updated the expected outputShape in the OpModel for the embedding op.
- With the corrected shape, the op validates correctly and behaves as it did prior to the uplift for this op. 

The fix successfully reproduces and resolves the failure on the original problematic commit: https://github.com/tenstorrent/tt-xla/actions/runs/19949564074 - fix succeeded on top of previously reproduced error commit
https://github.com/tenstorrent/tt-xla/actions/runs/19962253979/job/57246087254 - fix succeeded on top of current xla main
https://github.com/tenstorrent/tt-xla/actions/runs/20037329985/job/57463132607 december 8th version (failing bge-m3 is from current main, unrelated to this change)

### DIFF
--- a/lib/OpModel/TTNN/TTNNOpModel.cpp
+++ b/lib/OpModel/TTNN/TTNNOpModel.cpp
@@ -6489,14 +6489,12 @@ llvm::Expected<size_t> OpModel<UpsampleOp>::getOpRuntime(
 struct EmbeddingOpArgs {
   ::ttnn::TensorSpec inputSpec;
   ::ttnn::TensorSpec weightSpec;
-  std::optional<::ttnn::TensorSpec> outputSpec;
 };
 
 llvm::Expected<EmbeddingOpArgs> getEmbeddingOpArgs(
     ::tt::tt_metal::distributed::MeshDevice *device,
     llvm::ArrayRef<int64_t> inputShape, TTNNLayoutAttr inputLayout,
-    llvm::ArrayRef<int64_t> weightShape, TTNNLayoutAttr weightLayout,
-    llvm::ArrayRef<int64_t> outputShape, TTNNLayoutAttr outputLayout) {
+    llvm::ArrayRef<int64_t> weightShape, TTNNLayoutAttr weightLayout) {
   auto inputSpecExp =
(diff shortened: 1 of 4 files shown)


arminaleTT (approved):

sgholamiTT (approved):

lib/OpModel/TTNN/TTNNOpModel.cpp
```diff
@@ -6451,8 +6451,15 @@ getEmbeddingOpArgs(::tt::tt_metal::distributed::MeshDevice *device,
 
   std::optional<::ttnn::TensorSpec> outputSpec = std::nullopt;
   if (outputLayout) {
+    // Compute the correct output shape for embedding operation:
+    // outputShape = inputShape + [embeddingDim]
+    // where embeddingDim = weightShape[-1]
+    llvm::SmallVector<int64_t> outputShape;
+    outputShape.append(inputShape.begin(), inputShape.end());
+    outputShape.push_back(weightShape.back());
```
rpavlovicTT: Question: is it possible to pass outputShape through op's OpModel interface? Is the shape assigned to output tensor valid at the time?

bmalesevicTT: I tried it now, it works that way as well. Should I use that way as more reliable?

rpavlovicTT: Yes, no need to recreate output shape if it is present in the IR already.

PR #6198, opened 2025-12-05
Force host memory layout for const-eval inputs

dmilinkovicTT:
### Ticket
#6048 

## Problem description
Const-eval inputs are placed in DRAM by default. Since const-eval results are also stored in DRAM, this effectively means that if the deallocation of original inputs doesn't happen (because inputs are retained on the frontend), memory footprint is twice the actually required size.

OOMs which can be pin-pointed to this exact issue have been observed on some models.

## What's changed
- Force const-eval function arguments to system memory
- Force function arguments which get consumed by `load_cached` op to system memory
- Extend the logic for handling `ttir.mesh_shard` input and output layouts

### Checklist
- [x] Revert temporary nanobind dependencies flag

### DIFF
--- a/test/ttmlir/Dialect/TTNN/trace/matmul_multiply_consteval.mlir
+++ b/test/ttmlir/Dialect/TTNN/trace/matmul_multiply_consteval.mlir
@@ -3,6 +3,9 @@
 
 module {
   // CHECK-LABEL: func.func private @matmul_with_multiply_const_eval_0
+  // Both arguments are expected to be in system memory.
+  // CHECK: "ttnn.to_device"(%arg1, %{{.*}})
+  // CHECK: "ttnn.to_device"(%arg0, %{{.*}})
   // CHECK: "ttnn.matmul"
 
   // CHECK-LABEL: func.func private @trace_0_matmul_with_multiply


runtime/test/ttnn/python/utils.py
```diff
@@ -100,7 +100,9 @@ def get_inputs_and_golden(self, device, borrow=True):
         if self.config.compute_golden:
             golden = self.config.compute_golden(inputs_torch)
 
-        return inputs_runtime_with_layout, golden
+        # Returnining inputs_torch to prevent Python GC to free memory of the Torch tensors,
+        # which can lead to invalid memory access if the inputs are expected in the host memory.
+        return inputs_runtime_with_layout, golden, inputs_torch
```
dmilinkovicTT: ugly workaround, but not sure how else can we handle this situation.

also tried with `owned` instead of `borrowed` storage, and got similar results

CMakeLists.txt
```diff
@@ -13,6 +13,8 @@ if (NOT DEFINED ENV{TTMLIR_ENV_ACTIVATED})
   message(FATAL_ERROR "tt-mlir environment not activated. Please run 'source env/activate'.")
 endif()
 
+set(NB_USE_SUBMODULE_DEPS ON CACHE BOOL "Use nanobind bundled dependencies")
```
dmilinkovicTT, line 16: Temp fix, should be reverted after #6221 has been merged

lib/Dialect/TTNN/Transforms/TTNNLayout.cpp
```diff
@@ -384,20 +384,34 @@ class TTNNLayoutMeshShardRewriter : public OpRewritePattern<ttir::MeshShardOp> {
...
+    // This rewrite handles the following cases:
+    // (1) *Non-identity* MeshShardOp receives device input or produces device
+    // output
+    //     In this case, we insert ToLayout ops to convert the input to host
+    //     layout and change the output layout to host layout.
+    // (2) *Identity* MeshShardOp receives host input and produces host output
+    //     In this case, we insert ToLayout ops to convert the input to device
+    //     layout and change the output layout to device layout.
+    //
+    //  TODO(#2291): This can be removed once the workaround pass can correctly
```
svuckovicTT: fyi https://github.com/tenstorrent/tt-mlir/pull/6202

runtime/test/ttnn/python/utils.py
```diff
@@ -100,7 +100,9 @@ def get_inputs_and_golden(self, device, borrow=True):
         if self.config.compute_golden:
             golden = self.config.compute_golden(inputs_torch)
 
-        return inputs_runtime_with_layout, golden
+        # Returnining inputs_torch to prevent Python GC to free memory of the Torch tensors,
```
svuckovicTT:
```suggestion
        # Returning inputs_torch to prevent Python GC to free memory of the Torch tensors,
```

runtime/test/ttnn/python/utils.py
svuckovicTT, line 106: open issue to properly fix this

jameszianxuTT: Sorry, to be clear, the torch tensors `inputs_runtime_with_layout` may get deallocated if the local `inputs_torch` is not returned and GC'd?

dmilinkovicTT: `inputs_torch` get deallocated, and since we're passing `torch_tensor.data_ptr()` to the runtime in `get_runtime_tensor_from_torch`, this leads to use-after-free scenario if the input is expected in the system memory (and to_device transfer never happens)

lib/Dialect/TTNN/Transforms/TTNNConstEvalInputsToSystemMemory.cpp
```diff
@@ -0,0 +1,291 @@
...
+  convertArgumentsOfForwardFunc(func::FuncOp funcOp) {
+    SmallVector<Type> argumentTypes;
+    SmallVector<BlockArgument> convertedArguments;
+
+    // Find relevant const-eval inputs and convert them to system
+    // memory.
+    //
+    for (auto blockArgument : funcOp.getRegion().getArguments()) {
+      auto tensorType =
+          mlir::dyn_cast<RankedTensorType>(blockArgument.getType());
```
mtopalovicTT: cast

include/ttmlir/Dialect/TTNN/Transforms/Passes.td
```diff
@@ -335,6 +335,23 @@ def TTNNWeightBFP8Conversion: Pass<"ttnn-weight-bfp8-conversion", "::mlir::Modul
   }];
 }
 
+def TTNNForceConstEvalInputsToSystemMemory : Pass<"ttnn-force-const-eval-to-system-memory", "::mlir::ModuleOp"> {
```
odjuricicTT: Nit: the "Force" in the name might be redundant.

lib/Dialect/TTNN/Transforms/TTNNLayout.cpp
```diff
@@ -384,20 +384,33 @@ class TTNNLayoutMeshShardRewriter : public OpRewritePattern<ttir::MeshShardOp> {
...
     // handle canonicalization of toLayout ops (#2102). Currently the
     // workaround pass cannot detect redundant toLayout ops as a result of
     // forcing the output layout and removing them.
-    if (!shouldMeshShardOpForceSystemMemory(op.getOperation())) {
-      return failure();
-    }
+    const bool shouldForceSystemMemory =
+        shouldMeshShardOpForceSystemMemory(op.getOperation());
+
+    const bool shouldTile = !shouldForceSystemMemory;
```
odjuricicTT: This is no longer needed?

test/ttmlir/Dialect/TTNN/const-eval/const-eval.mlir
```diff
@@ -3,6 +3,9 @@
 
 module {
   // CHECK-LABEL: func.func private @forward_const_eval_0
+  // Both const-eval arguments should be in system memory.
```
odjuricicTT, line 6: Please add a test with `enable-trace=true` and const-eval if it does not exist.

odjuricicTT (commented):
Looks great!

test/ttmlir/Dialect/TTNN/const-eval/const-eval.mlir
dmilinkovicTT: There is one - `matmul_multiply_consteval.mlir`. 

Should that be enough?

lib/Dialect/TTNN/Transforms/TTNNLayout.cpp
dmilinkovicTT: forgot about these changes, thanks!

lib/Dialect/TTNN/Transforms/TTNNForceConstEvalInputsToSystemMemory.cpp
```diff
@@ -0,0 +1,291 @@
...
+
+private:
+  // Converts the arguments of the forward function which are const-eval inputs
+  // to system memory. Returns the list of converted arguments.
+  //
+  // An argument is considered to be a const-eval input if it is only
+  // consumed by LoadCachedOps
+  //
+  SmallVector<BlockArgument>
+  convertArgumentsOfForwardFunc(func::FuncOp funcOp) {
```
mtopalovicTT: Are these two convertArgumentOfConstEvalFunc/convertArgumentsOfForwardFunc using anything from class? If not can you move them up? Its more easier for reading, since the way is currently I would need to jump up/down when looking at code.

lib/Dialect/TTNN/Transforms/TTNNConstEvalInputsToSystemMemory.cpp
```diff
@@ -0,0 +1,318 @@
...
+namespace {
+
+// Helper function to create a system memory layout attribute for the given
+// tensor type.
+//
+static TTNNLayoutAttr createSystemMemoryLayoutAttr(RankedTensorType type) {
+  static const std::array<std::pair<int64_t, int64_t>, 1> defaultCollapseDims =
+      {{{0, -1}}};
+
+  auto currentLayout = mlir::cast<TTNNLayoutAttr>(type.getEncoding());
```
mtopalovicTT, line 32: currentLayout.withBufferType(BufferType::SystemMemory);

lib/Dialect/TTNN/Transforms/TTNNConstEvalInputsToSystemMemory.cpp
```diff
@@ -0,0 +1,318 @@
...
+          convertArgumentsOfForwardFunc(funcOp);
+
+      // Next, we need to update all const-eval functions accordingly.
+      //
+      for (auto convertedArgument : convertedArguments) {
+        // Find the LoadCachedOps which use the converted argument.
+        //
+        llvm::SmallVector<ttcore::LoadCachedOp> loadCachedOps;
+        for (auto *userOp : convertedArgument.getUsers()) {
+          if (auto op = llvm::dyn_cast<ttcore::LoadCachedOp>(userOp)) {
```
mtopalovicTT: mlir::dyn_cast

lib/Dialect/TTNN/Pipelines/TTNNPipelines.cpp
```diff
@@ -264,6 +264,10 @@ void createTTIRToTTNNBackendPipeline(
     // split during the analysis passes.
     if (options.enableConstEval) {
       devicePm.addPass(transforms::createConstEvalHoistTransform());
+
+      // Now that all const-eval passes have run, we can force the const-eval
+      // function inputs to system memory.
+      devicePm.addPass(createTTNNConstEvalInputsToSystemMemory());
```
svuckovicTT: Consider putting this under a flag. If we missed some edge cases, it's good to have a switch we can easily use if some models in tt-xla fail.

lib/Dialect/TTNN/Transforms/TTNNConstEvalInputsToSystemMemory.cpp
mtopalovicTT: Can we assert here that all ops are load cached? If argument was converted that already means that all users are load cached so I would just assert here. Instead I would just do:

for (auto op : convertedArgument.getUsers()) {
     TT_assertv(mlir::dyn_cast<ttcore::LoadCachedOp>(op), "Expected load cached op");
     auto loadCached = mlir::cast<ttcore::LoadCachedOp>(op);

lib/Dialect/TTNN/Transforms/TTNNConstEvalInputsToSystemMemory.cpp
```diff
@@ -0,0 +1,318 @@
...
+
+    ttcore::DeviceAttr device = ttcore::lookupDevice(moduleOp);
+    TT_assertv(device, "Device not found");
+
+    moduleOp->walk([&](func::FuncOp funcOp) {
+      // We only want to process forward functions.
+      //
+      if (funcOp.isDeclaration() || ttmlir::utils::isConstEvalFunc(funcOp)) {
+        return;
+      }
```
svuckovicTT, line 218: If we ever add another "function type" to IR, that is neither forward nor consteval, this will fail - I'd suggest inverting this statement so that we check if it is a "forward" function. However, there's no good way capture this today.

This is more of a conversation starter than any kind of blocker for this PR - does anyone have ideas on how we could implement this in the near future? cc @sdjordjevicTT @azecevicTT

lib/Dialect/TTNN/Transforms/TTNNConstEvalInputsToSystemMemory.cpp
```diff
@@ -0,0 +1,318 @@
...
+    registry.insert<mlir::tt::ttcore::TTCoreDialect>();
+    registry.insert<mlir::func::FuncDialect>();
+  }
+
+private:
+  // Converts the arguments of the forward function which are const-eval inputs
+  // to system memory. Returns the list of converted arguments.
+  //
+  // An argument is considered to be a const-eval input if it is only
+  // consumed by LoadCachedOps
```
svuckovicTT:
```suggestion
  // An argument is considered to be a const-eval input if it is
  // consumed only by LoadCachedOps
```

lib/Dialect/TTNN/Transforms/TTNNConstEvalInputsToSystemMemory.cpp
```diff
@@ -0,0 +1,318 @@
...
+}
+
+// Helper function to determine whether to transfer the const-eval argument to
+// device memory.
+//
+// The argument should NOT be transferred to device memory only if its sole user
+// is a ttnn.to_layout op which already transfers it to system memory - this
+// happens if the const-eval function is CPU-hoisted.
+//
+static bool shouldTransferArgumentToDevice(BlockArgument blockArgument) {
```
mtopalovicTT, line 59: This assumption is wrong if you ask me. What if we hoist multiple ops where some ops have constraint that input must be on host and some ops require device. I would look at the uses and then I would decide if I need to add to_layout for sysmem/device.

lib/Dialect/TTNN/Transforms/TTNNConstEvalInputsToSystemMemory.cpp
```diff
@@ -0,0 +1,318 @@
...
+  return RankedTensorType::get(ty.getShape(), ty.getElementType(), newLayout);
+}
+
+// Helper function to determine whether the given tensor type is in system
+// memory.
+//
+static bool isSystemMemory(RankedTensorType ty) {
+  auto layout = mlir::cast<TTNNLayoutAttr>(ty.getEncoding());
+  return layout.getBufferType() == BufferType::SystemMemory;
+}
```
svuckovicTT, line 50: There's a `mlir::tt::ttnn::isSystemBufferType` utility already, would it make sense to use that one instead?

lib/Dialect/TTNN/Transforms/TTNNConstEvalInputsToSystemMemory.cpp
```diff
@@ -0,0 +1,318 @@
...
+    SmallVector<Type> constEvalArgumentTypes(
+        constEvalFuncOp.getFunctionType().getInputs());
+
+    // If the argument is already converted to system memory, skip it.
+    //
+    if (constEvalArgumentTypes[argumentIndex] == systemMemoryType) {
+      return;
+    }
+
+    constEvalArgumentTypes[argumentIndex] = systemMemoryType;
```
mtopalovicTT: can we move this part:
```
    constEvalArgumentTypes[argumentIndex] = systemMemoryType;

    // Update the function type.
    auto newConstEvalFunctionType = mlir::FunctionType::get(
        constEvalFuncOp->getContext(), constEvalArgumentTypes,
        constEvalFuncOp.getFunctionType().getResults());

    constEvalFuncOp.setFunctionType(newConstEvalFunctionType);
```

at the bottom of this fn. I got confused that after this sequence
    auto blockArgument = constEvalFuncOp.getArgument(argumentIndex);
is system memory...

lib/Dialect/TTNN/Transforms/TTNNConstEvalInputsToSystemMemory.cpp
```diff
@@ -0,0 +1,318 @@
...
+
+private:
+  // Converts the arguments of the forward function which are const-eval inputs
+  // to system memory. Returns the list of converted arguments.
+  //
+  // An argument is considered to be a const-eval input if it is only
+  // consumed by LoadCachedOps
+  //
+  SmallVector<BlockArgument>
+  convertArgumentsOfForwardFunc(func::FuncOp funcOp) {
```
svuckovicTT: Rename to something more descriptive, like `moveConstEvalArgsToHost`

lib/Dialect/TTNN/Transforms/TTNNConstEvalInputsToSystemMemory.cpp
```diff
@@ -0,0 +1,318 @@
...
+      // avoid TTNNDecomposeLayouts pass failing because of a redundant
+      // to_layout op, as the argument is already in system memory.
+      //
+      auto toLayoutOp =
+          mlir::cast<ttnn::ToLayoutOp>(*blockArgument.getUsers().begin());
+
+      // If the data type and layout of the to_layout op matches the argument
+      // type, we can remove it.
+      //
+      auto toLayoutOpResultType =
```
mtopalovicTT: What if instead of doing this we run canonicalizer pass before decomposition. Will that remove redundant ops?

lib/Dialect/TTNN/Transforms/TTNNConstEvalInputsToSystemMemory.cpp
dmilinkovicTT: I've noticed this pattern with `dyn_cast` and allowing the argument not to be a tensor type in many places. 

Should we generally avoid using this pattern? Do we ever have non-tensor function inputs?

lib/Dialect/TTNN/Transforms/TTNNConstEvalInputsToSystemMemory.cpp
dmilinkovicTT: we use `currentLayout` only to get `TensorMeshAttr`, is this required?

lib/Dialect/TTNN/Transforms/TTNNConstEvalInputsToSystemMemory.cpp
dmilinkovicTT: Not sure if I understand, as the function already checks whether the `to_layout` op is the **only** user of the argument.

lib/Dialect/TTNN/Transforms/TTNNConstEvalInputsToSystemMemory.cpp
dmilinkovicTT: Yep, it works. Thanks!

lib/Dialect/TTNN/Transforms/TTNNConstEvalInputsToSystemMemory.cpp
mtopalovicTT: ah nvm block argument is device block argument.

lib/Dialect/TTNN/Transforms/TTNNConstEvalInputsToSystemMemory.cpp
mtopalovicTT: I had the same thought. One thing that came up was to tag all forward functions from inside some pass which will run at the very begin of the pipeline. In this way we can be sure that those are all fw functions because today we operate on those functions as forward functions in our passes.

lib/Dialect/TTNN/Transforms/TTNNConstEvalInputsToSystemMemory.cpp
mtopalovicTT: Only tensor types so far.

lib/Dialect/TTNN/Transforms/TTNNConstEvalInputsToSystemMemory.cpp
svuckovicTT: Agreed. Might even make sense to "expect" a tag on a forward function from the FEs.

test/ttmlir/Dialect/TTNN/const-eval/const-eval.mlir
odjuricicTT: Yes, if you add checks that arg0 and arg1 from that test were moved to system memory.

dmilinkovicTT: Added

PR #6204, opened 2025-12-05
Record program metadata in ttmetal runtime

gfengTT:
This allows `ttrt perf` to not skip ttnn api duration and device kernel duration in ttmetal code path.


gfengTT:
Thanks @jameszianxuTT . Yes, it was suggested by @tapspatel during our conversation.

PR #6209, opened 2025-12-06
Added cpu_module, device_module and multiple function support to builder

tapspatel:
Added support for cpu_module, device_module and multiple functions in builder.

For example, user can now do the following

```
def my_module(builder: TTIRBuilder):
        @builder.func([(32, 32)], [torch.float32])
        def my_modela(in0: Operand, builder: TTIRBuilder):
            sigmoid0 = builder.sigmoid(in0)
            return sigmoid0

        @builder.func([(32, 32)], [torch.float32])
        def my_modelb(in0: Operand, builder: TTIRBuilder):
            sigmoid0 = builder.sigmoid(in0)
            return sigmoid0
```

which will generate

```
module {
  func.func @my_modela(%arg0: tensor<32x32xf32>) -> tensor<32x32xf32> {
    %0 = "ttir.sigmoid"(%arg0) : (tensor<32x32xf32>) -> tensor<32x32xf32>
    return %0 : tensor<32x32xf32>
  }
  func.func @my_modelb(%arg0: tensor<32x32xf32>) -> tensor<32x32xf32> {
    %0 = "ttir.sigmoid"(%arg0) : (tensor<32x32xf32>) -> tensor<32x32xf32>
    return %0 : tensor<32x32xf32>
  }
}
```


dlokeTT (approved):

dmilinkovicTT:
Hey @tapspatel, great work, thanks! I'll focus on some CPU-module related remarks.

What are the intended usecases of explicit CPU module generation? Currently, CPU module is being generated automatically in a few places throughout the StableHLO -> TTNN pipeline (non-convertible SHLO ops, manually tagged ops, const-eval (soon)...). I’m wondering whether letting users explicitly create CPU module offers enough advantages to justify the risks introduced by the module’s rigid function structure and required constraints.
  - For example, if we want to allow users to CPU-hoist certain ops, we could just add `ttir.should_hoist` attribute to the op and let the pipeline handle the heavy-lifting
  - We could also think of a similar approach for groups of operations
  
On the other hand, if we end up agreeing that explicit CPU module generation makes sense, we should keep in mind the following constraints:
  - Only allowlisted SHLO and TTIR ops should be in the CPU module
  - Functions shouldn't invoke other functions
  - All tensors should use either `f32` or `i32`
  - Matching function declaration prototypes should be added to the device module
  - ...possibly others - constraints are constantly evolving as the feature is under development

PR #6210, opened 2025-12-07
[TTNN] Introduce ttnn.div operation

jserbedzijaTT:
### Problem description
Currently, we have ttnn.divide supported in the TTNN dialect. The issue is that it does not support integer inputs, so we have to typecast them to float32. Metal also has a separate `ttnn.div` composite op that supports integer inputs and returns a float by default if the `dtype` parameter is not provided.

### What's changed
- Added support for the `ttnn.div` op in the TTNN dialect and updated the lowering of `ttir.div` to either `ttnn.div` or `ttnn.divide`, depending on the input dtypes.

**Note:** Metal is planning to consolidate these two operations into a single op at some point in the future: https://github.com/tenstorrent/tt-metal/issues/28936

### Checklist

### DIFF
--- a/runtime/lib/ttnn/operations/eltwise/binary/binary_composite.cpp
+++ b/runtime/lib/ttnn/operations/eltwise/binary/binary_composite.cpp
@@ -122,7 +122,6 @@ void run(const ::tt::target::ttnn::EltwiseBinaryCompositeOp *op,
   }
 }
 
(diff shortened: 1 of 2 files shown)


jserbedzijaTT:
There is an effort in tt-metal to consolidate these ops into one thus I'm closing this.

PR #6212, opened 2025-12-08
[ttnn.jit] Fix reductions (max/sum) (L1-only)

ckaravasilisTT:
Max/sum will now work properly through `ttnn-jit` for L1 sharded tensors and 1x1 grid.

### What changed
The `scaler` will now be created with a `TTNNLayoutAttr` if the input also has one, making it go through `TTNNMetalLayoutCastOp` instead of `ToLayout`.

Added tests.

### DIFF
--- a/tools/ttnn-jit/_src/graph_trace_compiler.py
+++ b/tools/ttnn-jit/_src/graph_trace_compiler.py
@@ -165,7 +165,8 @@ class GraphToIRTranslator:
                 f"Found {len(grid_matches)} CoreRange(s) in output_info. "
                 f"Only single CoreRange grids are currently supported."
             )
-        if len(unique_grids) == 1:
+
+        if len(grid_matches) == 1:
             x_min, y_min, x_max, y_max = map(int, grid_matches[0])
             # Grid size is (max - min + 1) for each dimension
             # But TTNN uses (width, height) while compiler uses (height, width)


lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
```diff
@@ -770,15 +769,38 @@ class D2MNamedReductionRewriter final
   // Create a reduction scaler value for a given type of tensor operand
   // (at the current 'builder' insertion point).
   static mlir::Value createScaler(mlir::OpBuilder &builder, mlir::Location loc,
```
vtangTT, line 771: Is this typo? Shouldn't it be `createScalar`..

ckaravasilisTT: Given its wide use in the codebase, I don't think it's a typo. It's meant to refer to _scaling_.

xanderchin: `createScalerOperand` maybe?

tools/ttnn-jit/_src/graph_trace_compiler.py
```diff
@@ -149,14 +149,14 @@ def _parse_memory_config_from_output_info(
         grid_pattern = r"\[\(x=(\d+),y=(\d+)\)\s*-\s*\(x=(\d+),y=(\d+)\)\]"
         grid_matches = re.findall(grid_pattern, output_info_str)
 
-        if len(grid_matches) > 1:
+        unique_grids = list(set(grid_matches))
+        if len(unique_grids) > 1:
```
sgholamiTT: Sorry for this. As you mentioned, [this](https://github.com/tenstorrent/tt-mlir/issues/6225) is fixed in my [PR](https://github.com/tenstorrent/tt-mlir/pull/6231). Please revert these changes and rebase after that was merged (currently, main is broken, so it might take a while for me to be able to merge that)

test/ttnn-jit/test_reduction.py
```diff
@@ -14,9 +14,49 @@
...
+    ((128, 128), (0, 0)),
+    ((64, 128), (0, 0)),
+    ((128, 64), (0, 0)),
+    ((256, 256), (0, 0)),
+]
+
+REDUCTION_OPS = [
+    ("max", ttnn.max),
+    ("sum", ttnn.sum),
+]
```
sgholamiTT, line 46: I like this. Can we extend this to all reductions ops and use it for both `L1` and `DRAM` tests ?
```python
REDUCTION_SHAPES = [
    # 2D shapes (shape, max_grid)
    ((32, 32), (0, 0)),
    ((128, 128), (0, 0)),
    ((64, 128), (0, 0)),
    ((128, 64), (0, 0)),
    ((256, 256), (0, 0)),
]
REDUCTION_OPS = [
    ("max", ttnn.max),
    ("sum", ttnn.sum),
    ("min", ttnn.min),
    ("mean", ttnn.mean)
]

def test_reductions_l1(....):
    """Test reduction operations with L1 block-sharded config."""
    if op_name == "sum" and dtype == torch.bfloat16:
        pytest.xfail("failing allclose for some shapes for bfloat16")
    if op_name == "min" || op_name == "mean":
       pytest.xfail("...") # or even pytest.skip() if that's the case.
    # ...
    run_op_test(
       ...
       buffer_type=ttnn.BufferType.L1,
       ...
    )

@pytest.mark.skip(reason="D2M does not support reduction ops in DRAM")
def test_reductions_dram(....):
    """Test reduction operations with DRAM config."""
    # ...
    run_op_test(
       ...
       buffer_type=ttnn.BufferType.DRAM,
       ...
    )
```

Then, we can get rid of everything else in this file (Line 60-203) which is much cleaner.
Note 1: I think it's okay for `REDUCTION_SHAPES` to not include any 3D-4D tensors for now (since you mentioned they fail rn).
Note 2: If we don't want to add tests for "keep_dim=False", that's okay too.
Note 3: I guess you should be able to combine `test_reductions_l1` and `test_reductions_dram` into a single test suite as well and make it even more clean (by parametrizing it over mem loc). Up to you.

ckaravasilisTT: I agree. I will refactor the tests shortly.

Issue #6144, opened 2025-12-03
Optimize permutes that could be reshapes


mvasiljevicTT:
In llama models there are a lot of permutes like:
```
%37 = "ttnn.permute"(%36) <{permutation = [2,0,1,3]}> : tensor<32x8x1x64xbf16> -> tensor<1x32x8x64xbf16>
```
that could be rewritten as reshape, which would be nops.

johnny19436:
Hi, @mvasiljevicTT I'm interested in this issue, could you assign it to me?

PR #6213, opened 2025-12-08
[TTNN] Canonicalize permute to reshape when only size-1 dims move

johnny19436:
### Ticket
#6144 

### What's changed
Canonicalizes ttnn.permute to ttnn.reshape when the permutation only moves size-1 dimensions and preserves the relative order of non-singleton dims.
The pass checks shape consistency, encoding equality, and verifies that only singleton axes are reordered.

### Checklist

### DIFF
--- a/lib/Dialect/TTNN/IR/TTNNOps.cpp
+++ b/lib/Dialect/TTNN/IR/TTNNOps.cpp
@@ -3267,11 +3267,6 @@ void mlir::tt::ttnn::PermuteOp::getCanonicalizationPatterns(
     auto inputType = op.getInput().getType();
     auto resultType = op.getResult().getType();
 
-    // Reshape cannot change layout/encoding, only the view.
-    if (inputType.getEncoding() != resultType.getEncoding()) {
-      return mlir::failure();
-    }
-
     auto inShape = inputType.getShape();
     auto outShape = resultType.getShape();
     auto permutation = op.getPermutation();
--- a/test/ttmlir/Dialect/TTNN/Transforms/Workarounds/multiply_decomposition_workaround.mlir
+++ b/test/ttmlir/Dialect/TTNN/Transforms/Workarounds/multiply_decomposition_workaround.mlir
@@ -1,4 +1,4 @@
-// RUN: ttmlir-opt --ttcore-register-device="system-desc-path=%system_desc_path%" --ttnn-workaround --canonicalize -o %t %s
+// RUN: ttmlir-opt --ttcore-register-device="system-desc-path=%system_desc_path%" --ttnn-workaround -o %t %s
 // RUN: FileCheck %s --input-file=%t
 
 #dram = #ttnn.buffer_type<dram>
--- a/test/ttmlir/Dialect/TTNN/convolution/simple_conv1d.mlir
+++ b/test/ttmlir/Dialect/TTNN/convolution/simple_conv1d.mlir
@@ -21,16 +21,12 @@ module {
   // Test a different ordering of dimensions
   func.func public @conv1d_test2(%arg0: tensor<1x7x768xbf16>, %arg1: tensor<1x192x768xbf16>) -> (tensor<1x7x768xbf16>) {
     // CHECK: "ttnn.reshape"
-    // CHECK-SAME: shape = [1 : i32, 7 : i32, 768 : i32, 1 : i32]
-    // CHECK: "ttnn.reshape"
     // CHECK-SAME: shape = [1 : i32, 192 : i32, 768 : i32, 1 : i32]
     // CHECK: "ttnn.permute"
-    // CHECK-SAME: permutation = array<i64: 0, 1, 3, 2>
-    // CHECK: "ttnn.permute"
     // CHECK-SAME: permutation = array<i64: 2, 1, 0, 3>
+    // CHECK: "ttnn.reshape"
+    // CHECK-SAME: shape = [1 : i32, 1 : i32, 7 : i32, 768 : i32]
     // CHECK: "ttnn.conv2d"
-    // CHECK: "ttnn.permute"
-    // CHECK-SAME: permutation = array<i64: 0, 1, 3, 2>
     // CHECK: "ttnn.reshape"
     // CHECK-SAME: shape = [1 : i32, 7 : i32, 768 : i32]
     %0 = "ttir.convolution"(%arg0, %arg1) <{batch_group_count = 1 : i64, convolution_layout = #ttir<convolution_layout input_batch = 0, input_feature = 2, input_spatial_dimensions = 1, kernel_output_feature = 2, kernel_input_feature = 1, kernel_spatial_dimensions = 0, output_batch = 0, output_feature = 2, output_spatial_dimensions = 1>, feature_group_count = 4 : i64, input_dilation = array<i64: 1>, padding = array<i64: 0, 0>, weight_dilation = array<i64: 1>, window_reversal = array<i1: false>, window_strides = array<i64: 1>}> : (tensor<1x7x768xbf16>, tensor<1x192x768xbf16>) -> tensor<1x7x768xbf16>
--- a/test/ttmlir/Dialect/TTNN/data_movement/permute/simple_permute.mlir
+++ b/test/ttmlir/Dialect/TTNN/data_movement/permute/simple_permute.mlir
@@ -1,12 +1,12 @@
 // RUN: ttmlir-opt --ttir-to-ttnn-backend-pipeline -o %t %s
 // RUN: FileCheck %s --input-file=%t
 module {
-  func.func @permute(%arg0: tensor<1x4x32x64xf32>) -> tensor<4x32x64x1xf32> {
+  func.func @permute(%arg0: tensor<1x4x32x64xf32>) -> tensor<4x64x32x1xf32> {
     // CHECK: "ttnn.permute"
-    // CHECK-SAME: permutation = array<i64: 1, 2, 3, 0>
+    // CHECK-SAME: permutation = array<i64: 1, 3, 2, 0>
     // CHECK-SAME: tensor<1x4x32x64xf32
-    // CHECK-SAME: tensor<4x32x64x1xf32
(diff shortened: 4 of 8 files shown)


include/ttmlir/Dialect/TTNN/IR/TTNNOps.td
```diff
@@ -2477,6 +2477,7 @@ def TTNN_PermuteOp : TTNN_Op<"permute"> {
     let results = (outs AnyRankedTensor:$result);
 
     let hasVerifier = 1;
+    let hasCanonicalizer = 1;
```
azecevic000, line 2482: I see that the example in the linked issue contains `ttnn.permute`, but this canonicalization shouldn't be layout dependent, so it makes more sense to have it in TTIR dialect, not sure about TTNN? @mvasiljevicTT wdyt?

lib/Dialect/TTNN/IR/TTNNOps.cpp
```diff
@@ -3264,6 +3264,76 @@ ::mlir::LogicalResult mlir::tt::ttnn::PermuteOp::verify() {
...
+void mlir::tt::ttnn::PermuteOp::getCanonicalizationPatterns(
+    mlir::RewritePatternSet &patterns, mlir::MLIRContext * /*context*/) {
+
+  patterns.add(+[](PermuteOp op, mlir::PatternRewriter &rewriter)
+                   -> mlir::LogicalResult {
+    // Require ranked tensors so we can reason about shapes.
+    auto inputType =
+        llvm::dyn_cast<mlir::RankedTensorType>(op.getInput().getType());
+    auto resultType =
+        llvm::dyn_cast<mlir::RankedTensorType>(op.getResult().getType());
```
azecevic000: Both of these types are statically known `RankedTensorType`, so `dyn_cast` is redundant in this case (as well as subsequent check).

lib/Dialect/TTNN/IR/TTNNOps.cpp
```diff
@@ -3264,6 +3264,76 @@ ::mlir::LogicalResult mlir::tt::ttnn::PermuteOp::verify() {
...
+    if (inputType.getEncoding() != resultType.getEncoding())
+      return mlir::failure();
+
+    auto inShape = inputType.getShape();
+    auto outShape = resultType.getShape();
+    auto permutation = op.getPermutation();
+
+    auto rank = inShape.size();
+    if (rank != outShape.size() || rank != permutation.size())
+      return mlir::failure();
```
azecevic000: This is done as a part of op verification, so we don't have to do it for every op transformation like this one.

lib/Dialect/TTNN/IR/TTNNOps.cpp
```diff
@@ -3264,6 +3264,76 @@ ::mlir::LogicalResult mlir::tt::ttnn::PermuteOp::verify() {
...
+
+    // We need static sizes to distinguish 1 vs non-1 dims.
+    if (llvm::any_of(inShape, mlir::ShapedType::isDynamic) ||
+        llvm::any_of(outShape, mlir::ShapedType::isDynamic))
+      return mlir::failure();
```
azecevic000: We currently don't support dynamic shapes; generally, a lot of components won't work with dynamic shapes, so `assert` would be more appropriate here.

lib/Dialect/TTNN/IR/TTNNOps.cpp
```diff
@@ -3264,6 +3264,76 @@ ::mlir::LogicalResult mlir::tt::ttnn::PermuteOp::verify() {
...
+
+    // Ensure the result shape matches the permuted input shape.
+    auto expectedOutShape =
+        ttmlir::utils::applyPermutation(inShape, permutation);
+    if (!llvm::equal(expectedOutShape, outShape))
+      return mlir::failure();
```
azecevic000: This is also done as a part of verification, you can remove it.

lib/Dialect/TTNN/IR/TTNNOps.cpp
```diff
@@ -3264,6 +3264,76 @@ ::mlir::LogicalResult mlir::tt::ttnn::PermuteOp::verify() {
...
+
+    // Collect indices of non-1 dims in the original order.
+    llvm::SmallVector<int64_t, 4> origNonOne;
```
azecevic000: Nit: you can leave `4` out, just `llvm::SmallVector<int64_t>`

lib/Dialect/TTNN/IR/TTNNOps.cpp
```diff
@@ -3264,6 +3264,76 @@ ::mlir::LogicalResult mlir::tt::ttnn::PermuteOp::verify() {
...
+    for (int64_t i = 0, e = inShape.size(); i < e; ++i)
+      if (inShape[i] != 1)
+        origNonOne.push_back(i);
```
azecevic000: We use https://clang.llvm.org/extra/clang-tidy/checks/readability/braces-around-statements.html, so lint check will fail in CI. You should add braces around each block, i.e.
```cpp
for (...) {
  if (...) {
    ...
  }
}
```

lib/Dialect/TTNN/IR/TTNNOps.cpp
```diff
@@ -3264,6 +3264,76 @@ ::mlir::LogicalResult mlir::tt::ttnn::PermuteOp::verify() {
...
+
+    // If non-1 dims change relative order, this is a real permute.
+    if (!llvm::equal(origNonOne, permNonOne))
+      return mlir::failure();
+
+    // Only singleton dims are moved: we can safely rewrite as reshape.
+    llvm::SmallVector<int32_t, 4> newShape;
+    newShape.reserve(outShape.size());
+    for (int64_t d : outShape)
+      newShape.push_back(static_cast<int32_t>(d));
```
azecevic000: I believe `llvm::SmallVector<int32_t> newShape(outShape.begin(), outShape.end());` should do the same.

include/ttmlir/Dialect/TTNN/IR/TTNNOps.td
odjuricicTT: It is layout independant, tho i'm fine with leaving this in ttnn. We might add a similar canonicalization in ttir as well.

lib/Dialect/TTNN/IR/TTNNOps.cpp
johnny19436: outShape is int64_t, and I think the range constructor can’t narrow int64_t → int32_t. Is it okay to keep the loop?

johnny19436:
Thanks for the review! I've applied suggested fixes and the PR is ready for another pass.

include/ttmlir/Dialect/TTNN/IR/TTNNOps.td
johnny19436: That is reasonable, should I move it to TTIR instead?

johnny19436:
Thanks so much for the detailed review! I’ve applied the fixes and ready for another look.

lib/Dialect/TTNN/IR/TTNNOps.cpp
```diff
@@ -3264,6 +3264,65 @@ ::mlir::LogicalResult mlir::tt::ttnn::PermuteOp::verify() {
...
+      return mlir::failure();
+    }
+
+    auto inShape = inputType.getShape();
+    auto outShape = resultType.getShape();
+    auto permutation = op.getPermutation();
+
+    // Collect indices of non-1 dims in the original order.
+    llvm::SmallVector<int64_t> origNonOne;
+    for (int64_t i = 0, e = inShape.size(); i < e; ++i) {
```
mtopalovicTT, line 3276: for (std::size_t i = 0; i < inShape.size(); ++i) {

lib/Dialect/TTNN/IR/TTNNOps.cpp
```diff
@@ -3264,6 +3264,65 @@ ::mlir::LogicalResult mlir::tt::ttnn::PermuteOp::verify() {
...
+
+  patterns.add(+[](PermuteOp op,
+                   mlir::PatternRewriter &rewriter) -> mlir::LogicalResult {
+    // Require ranked tensors so we can reason about shapes.
+    auto inputType =
+        mlir::cast<mlir::RankedTensorType>(op.getInput().getType());
+    auto resultType =
+        mlir::cast<mlir::RankedTensorType>(op.getResult().getType());
+
+    // Reshape cannot change layout/encoding, only the view.
```
mtopalovicTT: I"m not sure if this is correct. If you are doing this permute:
inputShape: 4, 5,1
outputShape: 1, 4, 5

inputShape encoding would have affine map which would look like this: (d0, d1, d2) -> (d0 * 5 + d1, d2)
outputShape encoding would have affine map which would look like this: (d0, d1, d2) -> (d0 * 4 + d1, d2)

but if we take encoding out of the picture then then we should be able to canonicalize this as reshape right?

lib/Dialect/TTNN/IR/TTNNOps.cpp
```diff
@@ -3264,6 +3264,65 @@ ::mlir::LogicalResult mlir::tt::ttnn::PermuteOp::verify() {
...
+      }
+    }
+
+    // If non-1 dims change relative order, this is a real permute.
+    if (!llvm::equal(origNonOne, permNonOne)) {
+      return mlir::failure();
+    }
+
+    // Only singleton dims are moved: we can safely rewrite as reshape.
+    llvm::SmallVector<int32_t> newShape;
```
mtopalovicTT: not sure if you can single line this:
```
llvm::SmallVector<int32_t> newShape(outShape.begin(), outShape.end());
```

lib/Dialect/TTNN/IR/TTNNOps.cpp
azecevic000: Yeah, this is definitely too restrictive. Not sure if we need to account for layout at all here, when optimizer is enabled it should find the right configuration, and in general case workarounds will cover individual op constraints, so even if we end up with invalid configuration here it shouldn't be a problem E2E.

azecevicTT (approved):

mmanzoorTT (approved):
SHLO changes look good

PR #6221, opened 2025-12-08
Unblock build failing due to missing nanobind submodule dep

brataTT:
### Ticket
None

### Problem description
- Without `NB_USE_SUBMODULE_DEPS` set, nanobind's CMake config tries to find `tsl-robin-map` as a separate installed CMake package 
- The package is not installed separately, causing the build to fail

### What's changed
- Added `set(NB_USE_SUBMODULE_DEPS ON)`. tells nanobind to create the `tsl::robin_map` target from bundled headers instead of searching for a separate CMake package

### Checklist


CMakeLists.txt
```diff
@@ -153,6 +153,7 @@ add_subdirectory(third_party)
 if (TTMLIR_ENABLE_STABLEHLO)
   set(STABLEHLO_BUILD_EMBEDDED ON)
   set(STABLEHLO_ENABLE_BINDINGS_PYTHON ON)
+  set(NB_USE_SUBMODULE_DEPS ON CACHE BOOL "Use nanobind bundled dependencies")
```
jameszianxuTT: Why is this gated behind stableHLO build flag? We could probably set it at top level. Also, maybe a comment about why this is necessary for new NB version

brataTT: That's where I was seeing the failure locally. For stablehlo's python bindings when building explorer.
I assume we should put this elsewhere to cover other places where we use nanobind.
But it seems to be enough to unblock CI, since we always have stablehlo enabled in both speedy and tracy builds.

That said, idk if a longer lived solution should be installing the dep through cmake instead, and remove this option entirely.

nsmithtt: Yeah maybe it's better to just put this at the top level.  Let's see how CI does

pilkicTT: also, i think we should pin versions of all build-required packages

PR #6228, opened 2025-12-08
[D2M] Refactor invMap calculation for LowerToLayout

vwellsTT:
### Ticket
NA

### Problem description
In previous LowerToLayout, I introduced a func which attempts to use inversePermutation on any non-identity map and then defaults back to identity on failure.  This is not logically correct in all cases.

### What's changed
Explicitly detect permute fwd maps, and only inversePermute those.  llvm_unreachable on maps we cannot handle.

### Checklist


include/ttmlir/AffineMapUtils.h
```diff
@@ -171,23 +174,30 @@ createGridInverseMapFromIndexMap(mlir::AffineMap indexMap, unsigned gridRank,
...
+    // Get the inverse permutation
+    auto invGridMap = mlir::inversePermutation(gridMap);
 
-  // If inverse is null (not a valid permutation), fall back to identity
-  if (!invGridMap) {
-    return createIdentityGridInverseMap(context);
-  }
+    // If inverse is null (not a valid permutation), fall back to identity
+    if (!invGridMap) {
+      return createIdentityGridInverseMap(context);
```
vwellsTT, line 184: this should be an assert instead now I guess, will change

bgrady-tt: Agree with this approach. For the intended use case (permuting dims with index map), we should be able to inverse the mapping. Otherwise, it is likely a misuse of the feature.

Issue #6225, opened 2025-12-08
[D2M][TTNN-JIT][Graph Capture] Multiple core ranges might be detected in the output info


sgholamiTT:
### Describe the bug

There are 2 problems:
1. After `NDSharding` support in `ttnn-jit` is merged, there might be multiple `CoreRanges` detected in the str received from graph trace. We should Only parse grid from `ShardSpec`, not from `NdShardSpec`.
2. We should fail the test in multiple `CoreRange`s are detected. Currently, the exception is handled and suppressed.  

Example of the output info:
```
  output_info: Tensor(storage=DeviceStorage(),tensor_spec=TensorSpec(logical_shape=Shape([2, 1, 32, 2048]),tensor_layout=TensorLayout(dtype=DataType::FLOAT32,page_config=PageConfig(config=TilePageConfig(tile=Tile(tile_shape={32, 32},face_shape={16, 16},num_faces=4))),memory_config=MemoryConfig(memory_layout=TensorMemoryLayout::WIDTH_SHARDED,buffer_type=BufferType::L1,shard_spec=ShardSpec(grid={[(x=0,y=0) - (x=7,y=7)]},shape={64, 32},orientation=ShardOrientation::ROW_MAJOR),nd_shard_spec=NdShardSpec(shard_shape=Shape([64, 32]),grid={[(x=0,y=0) - (x=7,y=7)]},orientation=ShardOrientation::ROW_MAJOR,shard_distribution_strategy=ShardDistributionStrategy::ROUND_ROBIN_1D),created_with_nd_shard_spec=0),alignment=Alignment([32, 32]))))
```
### Steps to reproduce the issue

Happens on:
```bash
pytest "test/ttnn-jit/test_eltwise.py::test_binary_ops[True-div-f32-shape_(2, 1, 32, 2048)_grid_(7, 7)_TensorMemoryLayout.WIDTH_SHARDED]" -svv
```

Issue #6219, opened 2025-12-08
[D2M][TTNN-JIT][Graph Capture] ttnn.div might receive three args


sgholamiTT:
### Describe the bug

Running into this
```
E               ValueError: Binary operation divide expects 2 operands, got 3

/opt/ttmlir-toolchain/venv/lib/python3.11/site-packages/ttnn_jit/_src/graph_trace_compiler.py:735: ValueError
=========================================================================================================================================== short test summary info ============================================================================================================================================
FAILED test_digamma.py::test_digamma_compare[256-256] - ValueError: Binary operation divide expects 2 operands, got 3
```
seems like `ttnn.divide` is picking up an extra operand somewhere.

```python
@ttnn_jit.jit(debug=True, graph_capture=True)
def digamma(input_a, one_tensor):
    t_log_out = ttnn.log(input_a)  # negative log is not useful here
    # 1/2(z)
    tmp_reciprocal = ttnn.divide(one_tensor, input_a)
    output = ttnn.multiply(tmp_reciprocal, 0.5)
    tmp = ttnn.multiply(tmp_reciprocal, tmp_reciprocal)
    val_square = tmp
    # (1/12) * x^2
    output = ttnn.subtract(output, ttnn.multiply(tmp, 0.083333333))
    # (1/120) * x^4
    tmp = ttnn.multiply(tmp, val_square)
    output = ttnn.add(output, ttnn.multiply(tmp, 0.008333333333333333))
    # (1/252) * x^6
    tmp = ttnn.multiply(tmp, val_square)
    output = ttnn.subtract(output, ttnn.multiply(tmp, 0.003968253968253968))
    # (1/240) * x^8
    tmp = ttnn.multiply(tmp, val_square)
    output = ttnn.add(output, ttnn.multiply(tmp, 0.004166666666666667))
    # (1/132) * x^10
    tmp = ttnn.multiply(tmp, val_square)
    output = ttnn.subtract(output, ttnn.multiply(tmp, 0.007575757575757576))
    # (691/32760) * x^12
    tmp = ttnn.multiply(tmp, val_square)
    output = ttnn.add(output, ttnn.multiply(tmp, 0.021092796092796094))
    # (1/12) * x^14
    tmp = ttnn.multiply(tmp, val_square)
    output = ttnn.subtract(output, ttnn.multiply(tmp, 0.08333333333333333))
    return ttnn.subtract(t_log_out, output)
@pytest.mark.parametrize(
    "h, w",
[...]

PR #6231, opened 2025-12-08
[D2M][TTNN-JIT][Graph Capture] fixes ttnn.divide receiving three arguments

sgholamiTT:
### Ticket
#6219
#6225

### Problem description
This PR fixes two issues in `ttnn-jit` for graph capture: 
#### 1. Fixes the issue where `ttnn.divide` received three arguments in graph capture (instead of two):
- **Root cause:** There is an extra argument in the captured graph for `ttnn.divide` defaulted to `0` (string) and that argument was incorrectly being detected as an argument to the operand. So graph trace compiler detected the two input arguments + 1 argument (for the default 0 string we were creating a constant tensor) -> 3 arguments.
- **Solution:** avoid creating constant tensors whenever two input arguments already exist for a `binaryOp`.
- **Testing:** Added the `compile_only` option to `run_op_test` and used that for `div` tests. Those tests previously were just being ignored. Now they run through `jit` but they just skip the `allclose` checks.

#### 2. Detecting multiple CoreRanges in the output info
- **Root cause:** After `NDSharding` support in `ttnn-jit` is merged, there might be multiple `CoreRange`s detected in the str received from graph trace. 
- **Solution:** Parse grid from `ShardSpec`, not from `NdShardSpec`.
- **Testing:** Previously some tests resulted in some warnings when multiple `CoreRanges` were detected, but they did not fail (while they should have) since the exception was being caught later in the flow. By changing the `ValueError` to `BaseException` we no longer suppress it with `try-catch` statements. So the failures will be correctly identified.  

An example of such an output info:
```
output_info: Tensor(...,memory_config=MemoryConfig(....,shard_spec=ShardSpec(grid={[(x=0,y=0) - (x=7,y=7)]},...),nd_shard_spec=NdShardSpec(...,grid={[(x=0,y=0) - (x=7,y=7)]},orientation=.....)
```

### Checklist

### DIFF
--- a/.github/Dockerfile.base
+++ b/.github/Dockerfile.base
@@ -4,7 +4,7 @@ SHELL ["/bin/bash", "-c"]
 # Set environment variables
 ENV DEBIAN_FRONTEND=noninteractive
 
-ARG TT_METAL_DEPENDENCIES_COMMIT=e1d6113542dba5683fed6e0e2363fbf26042631e
+ARG TT_METAL_DEPENDENCIES_COMMIT=93e6cd7c0d2dbde65698178c0ebcd50d0bc3f204
 
 # Install dependencies
 RUN <<EOT
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,6 +13,8 @@ if (NOT DEFINED ENV{TTMLIR_ENV_ACTIVATED})
   message(FATAL_ERROR "tt-mlir environment not activated. Please run 'source env/activate'.")
 endif()
 
+set(NB_USE_SUBMODULE_DEPS ON CACHE BOOL "Use nanobind bundled dependencies")
+
 option(TT_RUNTIME_ENABLE_PERF_TRACE "Enable performance mode" OFF)
 option(TTMLIR_ENABLE_RUNTIME "Enable runtime" OFF)
 option(TTMLIR_ENABLE_PYKERNEL "Enable python kernels" OFF)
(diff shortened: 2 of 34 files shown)


tools/ttnn-jit/_src/op_registry.py
```diff
@@ -243,10 +243,15 @@ def prepare_operands(
 
         Binary operations may have a constant argument that needs to be converted
         to a constant tensor and added to the operand list.
+
+        Only adds a constant tensor if we have fewer than 2 operands. If we already
+        have 2 tensor operands from in_edges, we ignore any constant in parsed_args
+        (which may be metadata/default values from the trace).
```
vtangTT, line 249: lol this is so jank/unfortunate :(

tools/ttnn-jit/_src/graph_trace_compiler.py
```diff
@@ -146,11 +146,21 @@ def _parse_memory_config_from_output_info(
         # Format: grid={[(x=0,y=0) - (x=0,y=0)]}
         # This represents a grid from (x_min, y_min) to (x_max, y_max)
         # Note: Multiple CoreRanges are not supported in JIT/D2M
-        grid_pattern = r"\[\(x=(\d+),y=(\d+)\)\s*-\s*\(x=(\d+),y=(\d+)\)\]"
-        grid_matches = re.findall(grid_pattern, output_info_str)
+        # Only parse grid from ShardSpec, not from NdShardSpec (which may also contain grid)
```
vtangTT, line 149: @arminaleTT is this the right way to approach this?

arminaleTT: Yeah this is fine for now. When we enable ND sharding, we can switch over to parsing NdShardSpec (NdShardSpec will always exist, even when using legacy APIs)

PR #6232, opened 2025-12-08
Fix image build: Pass NB_USE_SUBMODULE_DEPS to llvm 

brataTT:
### Ticket
None

### Problem description
Same as https://github.com/tenstorrent/tt-mlir/pull/6221

### What's changed
Use the same solution as https://github.com/tenstorrent/tt-mlir/pull/6221 for image build since llvm build during env build does not inherit the same compile option

### Checklist


brataTT:
Looks like image build actually passed later 🤔 
I'll revisit this if we notice build failing again

Issue #5558, opened 2025-10-29
[OpModelLib] Handle `many tensor` results in some ops


rpavlovicTT:
Some ops like for example `ttnn.split_query_key_value_and_split_heads` return more than 1 tensor. In our design we don't support having more than 1 `outputLayout`. 

2 things are important:
1. We must be able to get every tensor layout/tensor_spec from backend in our compiler side.
2. We must be able to give memory config to these ops in a different way rather than converting output tensor's layout to memory config. Potentially we can extend OpConfig to contain `MemoryConfigAttr` instead of `TTNNLayoutAttr` (or it can be any of those 2 types). Now `MemoryConfigAttr` contains `ShardSpec` anyway, so I don't think we lose any functionality.

skhatriTT:
@rpavlovicTT can you please provide our latest update on this issue?

gfengTT:
Waiting for metal team's approval on https://github.com/tenstorrent/tt-metal/pull/32193

rpavlovicTT:
Added myself as an assignee to handle things on tt-mlir side when uplift lands.

PR #6233, opened 2025-12-08
multiple output tensors (simple version)

gfengTT:
### Ticket
#5558 

### Problem description
Allow multiple output tensors.

### What's changed
1. The output of an op will be a vector instead of a single object.
2. Awaiting metal change to query_op_constraints, which will return a vector output_tensor_specs.
3. Before changing all references to output_tensor_spec, we extract the first element in the vector as the workaround.


sdjordjevicTT:
The above issue that this PR solves is closed, hence I will close this one.

PR #6235, opened 2025-12-08
[ttnn.jit] Dump kernel source to file

vtangTT:
### Ticket
Link to Github Issue

### Problem description
We want to be able to dump the cpp kernels to a dedicated folder when running jit in `compile_only` mode. This is completely separate from ttrt

### What's changed
- modified `TTNNToFlatbuffer` to take a `kernelDumpDir` field which generic op translation will dump the kernel source to
- this will work for both the api and CLI
- for jit, `compile_only` mode will dump kernels to `generated/ttnn-jit/`

### Checklist

### DIFF
--- a/python/Passes.cpp
+++ b/python/Passes.cpp
@@ -208,40 +208,45 @@ void populatePassesModule(nb::module_ &m) {
       std::string, std::unordered_map<unsigned int, mlir::tt::GoldenTensor>>>(
       m, "GoldenMap");
 
-  m.def("ttnn_to_flatbuffer_file",
-        [](MlirModule module, std::string &filepath,
-           const std::unordered_map<
-               std::string,
-               std::unordered_map<std::uint32_t, mlir::tt::GoldenTensor>>
-               &goldenMap = {},
-           const std::vector<std::pair<std::string, std::string>> &moduleCache =
-               {}) {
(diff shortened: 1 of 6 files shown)


python/Passes.cpp
```diff
@@ -208,40 +208,45 @@ void populatePassesModule(nb::module_ &m) {
...
-        });
+        if (mlir::failed(mlir::tt::ttnn::translateTTNNToFlatbuffer(
+                moduleOp, file, goldenMap, moduleCache, kernelDumpDir))) {
+          throw std::runtime_error("Failed to write flatbuffer to file: " +
+                                   filepath);
+        }
+      },
+      nb::arg("module"), nb::arg("filepath"),
+      nb::arg("golden_map") = nb::dict(), nb::arg("module_cache") = nb::list(),
+      nb::arg("kernel_dump_dir") = "");
```
nsmithtt, line 249: I'd somewhat prefer having a dedicated pybind func like `dump_kernels` that walks the module and manually calls `translateKernelFuncToCpp` instead of just piggybacking on flatbuffer serialization.  IMO these are separate debugging features, but I won't block the PR.

vtangTT:
i overcomplicated the problem

closing in favour of https://github.com/tenstorrent/tt-mlir/pull/6277

PR #6242, opened 2025-12-09
update ci ttrt logger level

dlokeTT:
Title

### DIFF
--- a/.github/test_scripts/builder.sh
+++ b/.github/test_scripts/builder.sh
@@ -40,7 +40,7 @@ if [[ "$runttrt" == "1" ]]; then
 
                 # Check if the corresponding TTNN file exists
                 if [ -f "$ttnn_file" ]; then
-                    echo "Found matching TTNN file: $ttnn_file"
+                    # echo "Found matching TTNN file: $ttnn_file"
                     FLATBUFFER=" --flatbuffer $ttnn_file "
                 else
                     echo "No matching TTNN file found for $basename_file, running without flatbuffer comparison"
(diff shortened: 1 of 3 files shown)


dlokeTT:
> I thought we'll increase ttrt output (especially output of the tests that are executed), but setting level to ERROR seems to lower the output even further.

Sorry I'm not following. Are you saying that setting it to ERROR, we should expect even more output?

runtime/lib/ttnn/runtime.cpp
```diff
@@ -1268,9 +1268,9 @@ getOpOutputRef(OpContext opContextHandle,
   case ::tt::target::ttnn::OpType::NLPCreateQKVHeadsDecodeOp:
   case ::tt::target::ttnn::OpType::SplitQueryKeyValueAndSplitHeadsOp:
   case ::tt::target::ttnn::OpType::DumpTensorOp: {
-    LOG_WARNING("getting output tensor is not supported for ",
-                ::tt::target::ttnn::EnumNamesOpType()[static_cast<size_t>(
-                    opContext.type_type())]);
+    // LOG_WARNING("getting output tensor is not supported for ",
+    //             ::tt::target::ttnn::EnumNamesOpType()[static_cast<size_t>(
+    //                 opContext.type_type())]);
```
jameszianxuTT, line 1273: Is there some more context about why this is being removed or about the PR in general? Were we seeing overly verbose output from ttrt in CI?

dlokeTT: hey sorry kind of doing this just to run CI but yes this output is overly verbose and seems to be taking a ton of time to print in CI so I was seeing if removing this helps times

dlokeTT:
this didn't do anything

PR #6243, opened 2025-12-09
[D2M] add to/from_host ops to replace to_layout overloads

vwellsTT:
### Ticket
NA

### Problem description
Currently we have two forms of d2m to_layout--one which is decomposed into other ops, and one which represents moving data to/from host.  This is not ideal for obvious reasons (confusing, make translation less direct, etc).

### What's changed
Introduce new explicit ops for moving data to and from host--these ops can be directly lowered into ttmetal enqueue read/write buffer ops.  Now LowerToLayout can completely decompose all to_layout ops and it can be illegal later in the pipeline as well.

### Checklist

### DIFF
--- a/.github/Dockerfile.base
+++ b/.github/Dockerfile.base
@@ -4,7 +4,7 @@ SHELL ["/bin/bash", "-c"]
 # Set environment variables
 ENV DEBIAN_FRONTEND=noninteractive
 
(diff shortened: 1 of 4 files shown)


wenbinlyuTT (approved):
All changes LGTM! Personally I would have named them `d2m.copy_{h2d,d2h}`, but `to_{host,device}` are also clear enough.

PR #6244, opened 2025-12-09
[ttnn.jit] digamma demo

vtangTT:
### Problem description
We want digamma demo on main

### What's changed
- const handling in python FE was broken, return `op.result` instead of `op` in both ast and graph trace
- fix import paths in jit python src
- use `cleanup_source_code` in graph trace or else it will fail if you don't inline decorate
- handle view layouts in `scalarizeconst` pass

### Checklist

### DIFF
--- a/test/ttnn-jit/demo/test_digamma.py
+++ b/test/ttnn-jit/demo/test_digamma.py
@@ -9,10 +9,6 @@ import torch
 import pytest
 
 
-def digamma_ttnn(input_tensor):
-    return ttnn.digamma(input_tensor)
-
-
 @ttnn_jit.jit(debug=True, graph_capture=False, enable_cache=True)
 def digamma(input_tensor):
     t_log_out = ttnn.log(input_tensor)  # negative log is not useful here
@@ -69,7 +65,6 @@ def digamma(input_tensor):
     [
         digamma,
         ttnn_jit.jit(debug=True, graph_capture=True, enable_cache=True)(digamma.func),
-        digamma_ttnn,
     ],
 )
 def test_digamma_metal_trace(h, w, op):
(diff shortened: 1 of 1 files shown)


lib/Dialect/D2M/Transforms/ScalarizeConstTensors.cpp
```diff
@@ -32,6 +32,9 @@ static Value getLayoutOrCastResult(Operation *op) {
   if (auto toLayoutOp = dyn_cast<ToLayoutOp>(op)) {
     return toLayoutOp.getResult(0);
   }
+  if (auto viewLayoutOp = dyn_cast<ViewLayoutOp>(op)) {
+    return viewLayoutOp.getResult();
+  }
```
wenbinlyuTT, line 37: It would be nice if we have a lit test for this :3

test/ttnn-jit/demo/test_digamma.py
```diff
@@ -0,0 +1,158 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+import ttnn
+import ttnn_jit
+import torch
+
+import pytest
+
+
+@ttnn_jit.jit(debug=True, graph_capture=False, enable_cache=True)
```
sgholamiTT, line 12: Is there a reason we can't test with `graph_capture=True` rn ?

arminaleTT: Yeah we should test both

nsmithtt (approved):
+1 on the lit test for the scalarize const tensor change, otherwise LGTM!

lib/Dialect/D2M/Transforms/ScalarizeConstTensors.cpp
vtangTT: added, ptal 🙏

test/ttnn-jit/demo/test_digamma.py
vtangTT: now testing graph capture in both trace and non-traced

wenbinlyuTT (approved):

PR #6245, opened 2025-12-09
Fixing errors with pad for ttnn.avgPool2d

djstefanovicTT:
### Ticket
closes https://github.com/tenstorrent/tt-mlir/issues/5785

### Problem description
Pooling op gets fused with pad op, but tt-metal has constrain that pad must be smaller than half of the kernel size. Verifier only checks that input dimensions are greater than kernel dimensions, disregarding padding.

### What's changed
If padding is greater than allowed, added decomposition to padOp + poolingOp. It is not done as a special case when it should not be fused because it is mathematically correct, but this is a hardware constraint. Fixed verifier logic.

### Checklist


lib/Conversion/TTIRToTTNN/TTIRToTTNN.cpp
```diff
@@ -1655,6 +1655,54 @@ class Pooling2dOpConversionPattern : public OpConversionPattern<TTIROpTy> {
...
+          mlir::cast<mlir::TypedValue<mlir::RankedTensorType>>(input),
+          nhwcShape, rewriter, op.getLoc());
+
+      // Apply padding to NHWC tensor.
+      llvm::SmallVector<int32_t> nhwcPadding = {
+          0, 0, paddingTop, paddingBottom, paddingLeft, paddingRight, 0, 0};
+      auto paddedInput = ttir_to_ttnn::utils::generatePad(
+          mlir::cast<mlir::TypedValue<mlir::RankedTensorType>>(
+              unflattenedInput.getResult()),
+          nhwcPadding, rewriter, op.getLoc());
```
sdjordjevicTT, line 1685: @odjuricicTT @vkovacevicTT 

I’m uncertain if we still need to maintain unique locations. Can you please provide more comments on whether we should add a prefix or postfix here?

svuckovicTT: I'd stray away from manually changing locations - it'll affect codegen (or anything else that expects a certain location format in the future)

mtopalovicTT:
@djstefanovicTT @sdjordjevicTT can we maybe move this fusing pattern into ttnn? We are going to move the decomposition of pooling op upstream at some point right? If we move this fusing pattern to ttnn we can use op model to check if we can fuse or not.

Other way is to add the check for that padding is < kernel_size / 2 during fusing. This constraint is not TTNN constraint only is constraint in other ML frameworks for pooling ops.

sdjordjevicTT:
> @djstefanovicTT @sdjordjevicTT can we maybe move this fusing pattern into ttnn? We are going to move the decomposition of pooling op upstream at some point right? If we move this fusing pattern to ttnn we can use op model to check if we can fuse or not.
> 
> Other way is to add the check for that padding is < kernel_size / 2 during fusing. This constraint is not TTNN constraint only is constraint in other ML frameworks for pooling ops.

I wouldn't block these PR changes now, as they will unblock a couple of things. But in general, I agree with your opinion. @djstefanovicTT let's create an issue to follow up once we merge this. @mtopalovicTT, are you ok with this?

PR #6260, opened 2025-12-10
Store convolution config tensors in DRAM

djstefanovicTT:
### Problem description
Some models fail upon execution because L1 cannot allocate enough space for tensors.

### What's changed
Added a flag in ttir conv2dconfig if tensor should be in DRAM. Same flag exists in ttnn config for convolution. In conversion from TTIR to TTNN convolution this flag is set to true.


sdjordjevicTT (approved):

svuckovicTT:
Can you add a sanity test as well please?

djstefanovicTT:
> Can you add a sanity test as well please?

It is set to true always, so it is tested already.

svuckovicTT:
> > Can you add a sanity test as well please?
> 
> It is set to true always, so it is tested already.

Sounds good!

Btw do you know if `L1_SMALL_SIZE` is used for anything besides conv halo data? If not, we should consider setting this to 0:
https://github.com/tenstorrent/tt-mlir/blob/903cd1d7848d0edde172fcd2a0f18150b0ed5043/include/Constants.h#L14

It is used to open device here:
https://github.com/tenstorrent/tt-mlir/blob/903cd1d7848d0edde172fcd2a0f18150b0ed5043/runtime/lib/ttnn/runtime.cpp#L466-L477

fyi @rpavlovicTT @odjuricicTT

odjuricicTT:
> Btw do you know if L1_SMALL_SIZE is used for anything besides conv halo data? If not, we should consider setting this to 0:

@svuckovicTT 

It's used for other pooling ops that don't have the option to store these in DRAM yet, so it cannot be 0 yet.

tapspatel (approved):

PR #6264, opened 2025-12-10
[ttnn.jit] default pcc check, optional all_close, add input constraints for certain ops

ckaravasilisTT:
### What's changed
Tests (`run_op_test`) now use `pcc` check by default, with an optional `all_close` check.

Added input transformations for `sqrt`, `tan`, `reciprocal`, similar to the ones in `test/python/golden/ttir_ops/eltwise/test_ttir_unary.py`.

### DIFF
--- a/test/ttnn-jit/utils.py
+++ b/test/ttnn-jit/utils.py
@@ -34,6 +34,12 @@ def _transform_reciprocal(t: torch.Tensor) -> torch.Tensor:
     return torch.where(t == 0, torch.tensor(1e-6, dtype=t.dtype), t)
 
 
+def _transform_digamma(t: torch.Tensor) -> torch.Tensor:
+    t = t * 1e5
+    t = torch.clamp(t, min=1)
+    return t
+
+
 def _transform_sqrt(t: torch.Tensor) -> torch.Tensor:
     """Ensure non-negative values."""
     return torch.abs(t)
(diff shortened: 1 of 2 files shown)


test/ttnn-jit/utils.py
```diff
@@ -37,6 +72,15 @@ def get_block_sharding_grid(shape):
...
+    """Check PCC between result and golden using ttnn.pearson_correlation_coefficient."""
+    result_torch = result.cpu().to_torch()
+    golden_torch = golden_result.cpu().to_torch()
+    pcc = ttnn.pearson_correlation_coefficient(golden_torch, result_torch)
+    print(f"PCC: {pcc} (threshold: {threshold})")
+    return pcc >= threshold, pcc
+
+
 def all_close_check(result, golden_result, atol=1e-1, rtol=1e-1, debug=True):
     if debug:
```
vtangTT: can we move this debug print to run_op_test so it'll print output+golden even if we only check pcc

ckaravasilisTT: Done

vtangTT:
Can you check if `test_digamma_dram` (unsure if i got this test name correct) in `test_eltwise_composite.py` passes pcc. It's xfailed rn, but I think it should pass pcc? If so can you enable it here too 🙏

ckaravasilisTT:
@vtangTT digamma does not pass pcc either. If the input is transformed similarly to the digamma test in `test/python/golden/test_composite_functions.py` it passes. I added that transformation as well. We should discuss if `utils.py` is the best place to have all these transformations long term.

Issue #6258, opened 2025-12-10
[D2M][TTNN-JIT][Graph Capture] Incorrect layout info for digamma


sgholamiTT:
### Describe the bug

Some weird behavior for graph trace for `digamma` for `L1_BLOCK_SHARDED` input.

### Steps to reproduce the issue

Reproducible with
```python
@ttnn_jit.jit(debug=True, graph_capture=True, enable_cache=True)
def digamma_ttnn(input_tensor):
    return ttnn.digamma(input_tensor)
@pytest.mark.parametrize(
    "op",
    [digamma_ttnn],
)
def test_digamma_compare(h, w, op):
    device = ttnn.open_device(device_id=0)
    # h, w = 1024, 1024
    dtype = torch.bfloat16
    max_grid = (7, 7)
    torch_tensor_a = torch.rand((h, w), dtype=dtype) * 100
    core_range = ttnn.CoreRange(
        ttnn.CoreCoord(0, 0), ttnn.CoreCoord(max_grid[0], max_grid[1])
    )
    core_range_set = ttnn.CoreRangeSet([core_range])
    tensor_spec = ttnn.TensorSpec(
        shape=(h, w),
        dtype=ttnn.DataType.BFLOAT16,
        layout=ttnn.TILE_LAYOUT,
        buffer_type=ttnn.BufferType.L1,
    ).block_sharded(core_range_set)
    input_a = ttnn.from_torch(torch_tensor_a, spec=tensor_spec, device=device)
    output_tensor = op(input_a)
    golden = ttnn.digamma(input_a)
    print(f"output_tensor\n: {output_tensor}")
    print(f"golden\n: {golden}")
    pcc = ttnn.pearson_correlation_coefficient(
        golden.cpu().to_torch(), output_tensor.cpu().to_torch()
    )
    ttnn.close_device(device)
    print(f"PCC: {pcc}")
    assert pcc > 0.99
```

Which gives this error:
```
---- IR Dump after GraphToIRTranslator (Graph-based) ----
#dram = #ttnn.buffer_type<dram>
#l1 = #ttnn.buffer_type<l1>
#ttnn_layout = #ttnn.ttnn_layout<(d0, d1) -> (d0, d1), <8x8>, memref<1x1x!ttcore.tile<32x32, bf16>, #l1>, <block_sharded>, exactGrid = true>
[...]

PR #6267, opened 2025-12-10
[D2M][TTNN-JIT][Graph Capture] fix digamma with L1 input

sgholamiTT:
### Ticket
#6258 
Main root cause (not solved in this PR): https://github.com/tenstorrent/tt-metal/issues/33914

### Problem description
When `ttnn.digamma` is used in `ttnn-jit` with graph capture, we failed to generate correct IR.

### Root cause
1. `return_modifier.py` adds `ttnn.identity` before return statements to populate `output_info` for the last top-level op,
2. For composite ops, the parent op (`digamma`) gets `output_info` populated, but internal ops don't. That's why the last `subtract` has bad layout information.
3. When `output_info` is empty, it defaults to `DRAM/INTERLEAVED` instead of using the parent's layout -> the issue we saw.

### Implementation
Inherit Parent Output Info in Composite Processing.
#### Changes made:
1. Modified `_get_output_type_from_vertex` signature:
    - Added optional `parent_output_info: Optional[List[str]] = None` parameter,
    - Updated logic to use parent's output_info when vertex's output_info is empty,
2. Updated `_process_composite_vertex`:
    - Parse parent composite vertex's `output_info` once at the start,
    - Pass `parent_output_info` to `_get_output_type_from_vertex` for all internal ops
3. Adds tests in `test_digamma.py` that tests digamma op with input in L1. 

### Checklist

### DIFF
--- a/test/ttnn-jit/demo/test_digamma.py
+++ b/test/ttnn-jit/demo/test_digamma.py
@@ -49,8 +49,8 @@ def digamma(input_tensor):
     return ttnn.subtract(t_log_out, output)
 
 
+@ttnn_jit.jit(debug=True, graph_capture=True, enable_cache=True)
 def digamma_ttnn(input_tensor):
-    """Wrapper function that uses ttnn.digamma composite op."""
     return ttnn.digamma(input_tensor)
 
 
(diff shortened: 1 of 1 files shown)


arminaleTT (approved):

test/ttnn-jit/demo/test_digamma.py
```diff
@@ -65,6 +70,7 @@ def digamma(input_tensor):
     [
         digamma,
         ttnn_jit.jit(debug=True, graph_capture=True, enable_cache=True)(digamma.func),
+        ttnn_jit.jit(debug=True, graph_capture=True, enable_cache=True)(digamma_ttnn),
```
vtangTT: nit: can you actually decorate it, instead of inlining it

i inlined the decoration of graph capture for the first one cuz i want to explicitly show `digamma` being decorated instead of just inlining

vtangTT (approved):
ty 🙏

PR #6270, opened 2025-12-10
added stablehlo ops

dlokeTT:
Added additional StableHLO ops to builder and their corresponding golden mappings and mlir snippets as tests. List of ops below: 

```
stablehlo.abs
stablehlo.and
stablehlo.batch_norm_grad
stablehlo.batch_norm_training
stablehlo.broadcast_in_dim
stablehlo.ceil
stablehlo.clamp
stablehlo.concatenate
stablehlo.constant
stablehlo.cosine
stablehlo.divide
stablehlo.exponential
```


tools/builder/stablehlo/stablehlo_builder.py
```diff
@@ -386,121 +386,2927 @@ def decorated_func(*inputs):
 
         return add_module, add_builder
 
+    ################ stablehlo.AndOp ###############
+
+    @tag(stablehlo.AndOp)
+    def and_(
```
tapspatel: can we rename this to `def and`
remove the _, including for other ops

tools/builder/stablehlo/stablehlo_builder.py
```diff
@@ -386,121 +386,2927 @@ def decorated_func(*inputs):
...
+                    new_op = stablehlo_op(lhs, rhs, loc=old_op.location)
+
+                    if not self._disable_golden_check:
+                        op_golden_function = get_golden_function(stablehlo_op)
+                        input0 = self._get_golden_tensor(old_op.lhs)
+                        input1 = self._get_golden_tensor(old_op.rhs)
+                        golden_output = op_golden_function(
+                            input0, input1, new_op.result.type.element_type
+                        )
+                        and_builder._set_golden_tensor(new_op, golden_output)
```
tapspatel: this should be new_op.result

tools/builder/stablehlo/stablehlo_builder.py
```diff
@@ -386,121 +386,2927 @@ def decorated_func(*inputs):
...
+                        and_builder._set_output_ordering([new_op])
```
tapspatel: new_op.result

tools/builder/stablehlo/stablehlo_builder.py
```diff
@@ -386,121 +386,2927 @@ def decorated_func(*inputs):
...
+        contract_dims_lhs: List[int],
+        batch_dims_rhs: List[int],
+        contract_dims_rhs: List[int],
+        loc: Optional[str] = None,
+        unit_attrs: Optional[List[str]] = None,
+        sharding_attr: Optional[sdy.TensorShardingPerValueAttr] = None,
+    ) -> OpResult:
+        stablehlo_op = self.get_opview_from_method(StableHLOBuilder.dot_general)
+
+        dot_dimension_numbers = stablehlo.DotDimensionNumbers.get(
```
tapspatel: we should pass this entire structure into the golden function, like we do for ttir.convolution

tools/builder/stablehlo/stablehlo_builder.py
```diff
@@ -386,121 +386,2927 @@ def decorated_func(*inputs):
...
+            context=self._ctx,
+            lhs_batching_dimensions=batch_dims_lhs,
+            lhs_contracting_dimensions=contract_dims_lhs,
+            rhs_batching_dimensions=batch_dims_rhs,
+            rhs_contracting_dimensions=contract_dims_rhs,
+        )
+
+        lhs_shape = list(in0.type.shape)
+        rhs_shape = list(in1.type.shape)
+        result_shape = self._compute_dot_general_result_shape(
```
tapspatel: we should get this from the golden output tensor

tools/builder/stablehlo/stablehlo_builder.py
dlokeTT: yea, just kept that since that's what it was. think this is the only refactored op that has it. as the other ops get refactored will change them

dlokeTT: oh also realized we can't since and is a keyword

dlokeTT: making it logical_and just like ttir does

tools/builder/stablehlo/stablehlo_builder.py
dlokeTT: why's that? I get its done for you but seems kind of weird/unnatural that we have to get that info from outside of the inputs to the ops. also what about the case where goldens are disabled?

dlokeTT: on second thought, I don't have a strong opinion about this. I see pros to using the goldens since they do all the work for you and that saves a lot of redundant code. but also still have a question about what we do when goldens are disabled

tools/builder/stablehlo/stablehlo_builder.py
dlokeTT: For ttir dot general we pass each of these params in explicitly. Are you saying to make this like the ttir convolution layout attr?

tools/builder/stablehlo/stablehlo_builder.py
```diff
@@ -398,35 +398,22 @@ def decorated_func(*inputs):
 
         return add_module, add_builder
 
-    ############### stablehlo.LogOp ###############
+    ################ stablehlo.AndOp ###############
```
tapspatel, line 401: extra # 

see andop

it makes it more easy to search how many ops we have completed the refactor for

tools/builder/stablehlo/stablehlo_builder.py
```diff
@@ -398,35 +398,22 @@ def decorated_func(*inputs):
...
 
-    @tag(stablehlo.LogOp)
-    def log(
+    @tag(stablehlo.AndOp)
+    def logical_and(
```
tapspatel, line 404: name this `and`, lets keep it consistent with shlo convention 
`https://openxla.org/stablehlo/spec#and`

tools/builder/stablehlo/stablehlo_builder.py
```diff
@@ -543,99 +532,108 @@ def neg(
...
-        if output_type is None:
-            mlir_output_type = self.get_type(in0)
-        else:
-            mlir_output_type = self._get_type_from_torch_dtype(output_type)
+        stablehlo_op = self.get_opview_from_method(StableHLOBuilder.compare)
 
-        input0 = self._get_golden_tensor(in0)
-        op_golden_function = get_golden_function(stablehlo_op)
-        golden_output = op_golden_function(input0, mlir_output_type)
+        # Get comparison direction attribute
```
tapspatel: nit: redundant comment, we already have lots of code bloat, and if the code can be explained by reading it, we shouldn't add comments

tools/builder/stablehlo/stablehlo_builder.py
```diff
@@ -543,99 +532,108 @@ def neg(
...
+        comparison_direction_attr = stablehlo.ComparisonDirectionAttr.get(
```
tapspatel: nice, this is exactly what I want to do in ttir as well :)

tools/builder/stablehlo/stablehlo_builder.py
```diff
@@ -543,99 +532,108 @@ def neg(
...
+            comparison_direction, context=self._ctx
+        )
 
-        if loc is None:
-            loc = self._get_location()
-        else:
-            loc = Location.name(loc)
+        # Get optional compare type attribute
```
tapspatel: redundant comment

tools/builder/stablehlo/stablehlo_builder.py
```diff
@@ -647,96 +645,131 @@ def rsqrt(
                 op.operation.attributes[attr_name] = UnitAttr.get(self._ctx)
 
         if not self._disable_golden_check:
+            lhs_golden = self._get_golden_tensor(lhs)
+            rhs_golden = self._get_golden_tensor(rhs)
+            op_golden_function = get_golden_function(stablehlo_op)
+            golden_output = op_golden_function(
+                lhs_golden,
+                rhs_golden,
+                comparison_direction,
```
tapspatel: the golden function should take the compare attribute, not its python type

this makes the parsing + split functionality more easy since we don't need to unpack the attribute

tools/builder/stablehlo/stablehlo_builder.py
```diff
@@ -647,96 +645,131 @@ def rsqrt(
...
+        old_op: stablehlo.CompareOp,
         global_dict: Dict[Operand, Operand],
     ) -> Tuple[Operation, Dict[OpResult, OpResult]]:
-        stablehlo_op = self.get_opview_from_parser(StableHLOBuilder.rsqrt_parser)
-        in0 = global_dict[old_op.operand]
+        stablehlo_op = self.get_opview_from_parser(StableHLOBuilder.compare_parser)
+        lhs = global_dict[old_op.lhs]
+        rhs = global_dict[old_op.rhs]
+
+        # Extract comparison direction from the attribute
```
tapspatel: golden function should take the attribute and its unpacked there, so we don't have to do this here

tools/builder/stablehlo/stablehlo_builder.py
```diff
@@ -647,96 +645,131 @@ def rsqrt(
...
+        # The attribute string looks like "#stablehlo<comparison_direction LT>"
+        # We need to extract just "LT"
+        comparison_direction_attr = old_op.comparison_direction
+        attr_str = str(comparison_direction_attr)
+        comparison_direction = attr_str.split()[-1].rstrip(">")
```
tapspatel: see ttir.convolution for an example, we should also have an equivalent `maybe_downcast` call that gets done

tools/builder/stablehlo/stablehlo_builder.py
```diff
@@ -647,96 +645,131 @@ def rsqrt(
...
+
+        # Extract compare type if present
+        compare_type_attr = old_op.compare_type if old_op.compare_type else None
```
tapspatel: inline this when creating the op

tools/builder/stablehlo/stablehlo_builder.py
```diff
@@ -647,96 +645,131 @@ def rsqrt(
...
-        old_loc = Location.unknown(old_ctx)
-        with old_ctx, old_loc:
-            rsqrt_module = Module.create()
-            rsqrt_builder = StableHLOBuilder(old_ctx, old_loc)
-            op_input_types = [old_op.operand.type]
+        old_context = old_op.context
+        old_loc = Location.unknown(old_context)
 
-            with InsertionPoint(rsqrt_module.body):
+        # Extract comparison direction from the attribute
```
tapspatel: same thing here with the golden function, pass in the attribute so we don't have to unpack here

tools/builder/stablehlo/stablehlo_builder.py
```diff
@@ -751,98 +784,92 @@ def sine(
...
-        return sine_module, sine_builder
+        return ceil_module, ceil_builder
 
-    ############### stablehlo.SqrtOp ###############
+    ################ stablehlo.DivOp ###############
 
-    @tag(stablehlo.SqrtOp)
-    def sqrt(
+    @tag(stablehlo.DivOp)
+    def divide(
```
tapspatel, line 703: rename to match stablehlo op name `div`

tools/builder/stablehlo/stablehlo_builder.py
```diff
@@ -2573,209 +2483,3091 @@ def select(
...
+
+        return clamp_module, clamp_builder
+
+    ################ stablehlo.ConcatenateOp ###############
+
+    @tag(stablehlo.ConcatenateOp)
+    def concatenate(
+        self,
+        inputs: List[Operand],
+        dim: int = 0,
```
tapspatel: dim is not defaulted to 0 in op definition, the user must provide it

https://github.com/openxla/stablehlo/blob/3934f366c978ee061081b393036382a622c0acd3/stablehlo/dialect/StablehloOps.td#L2263C3-L2266C5

```
let arguments = (ins
    Variadic<HLO_Tensor>:$inputs /*concatenate_i1*/,
    ConfinedAttr<I64Attr, [IntNonNegative]>:$dimension /*concatenate_c4, concatenate_i2*/
  );
```

tools/builder/stablehlo/stablehlo_builder.py
```diff
@@ -2573,209 +2483,3091 @@ def select(
...
+
+        return concatenate_module, concatenate_builder
+
+    ################ stablehlo.ConvertOp ###############
+
+    @tag(stablehlo.ConvertOp)
+    def convert(
+        self,
+        in0: Operand,
+        element_type: Type,
```
tapspatel: convert into python type (torch.dtype)

tools/builder/stablehlo/stablehlo_builder.py
dlokeTT: talked offline but making logical_and to match ttir builder. And is reserved keyword in python

PR #6283, opened 2025-12-11
where working e2e d2m with tests and with fusion for fp32 and bf16

### LINKED ISSUE #3029: d2m ttir.where

dlokeTT:
### Ticket
Closes #3029

### Problem description
Need support for where in d2m for llama and other models

### What's changed
Wired up the `ttir.where` op thru the d2m path. Support both bf16 and fp32
Added new ternary compute op class in d2m and proper handling in `D2MtoTTKernel.cpp`
In op scheduler, add middle child for ternary op support
Added lit, builder and fusion tests

### Checklist

### DIFF
--- a/lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
+++ b/lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
@@ -406,18 +406,24 @@ private:
     SmallVector<SmallVector<mlir::AffineExpr>> inputExprs(
         numInputs, SmallVector<mlir::AffineExpr>(outRank));
 
+    // Deduce output shape and build affine indexing maps for broadcasting.
+    // We iterate right-to-left (innermost to outermost) to align dimensions
+    // per NumPy semantics. Lower-rank inputs are implicitly unsqueezed with
+    // leading 1s. For each dim, we validate compatibility and mark inputs
+    // needing broadcast with constant-0 affine exprs (locked index).
     SmallVector<int64_t> deducedShape(outRank, -1);
     for (int i = -1; i >= -outRank; i--) {
       const int outDim = outRank + i;
 
-      // Gather dim sizes for all inputs (-1 for non-existent dims).
+      // Gather dim sizes for all inputs (-1 for missing dims in lower-rank
+      // tensors).
       SmallVector<int64_t> dimSizes;
       for (size_t j = 0; j < numInputs; ++j) {
         const int inputDim = inputRanks[j] + i;
         dimSizes.push_back((inputDim >= 0) ? inputShapes[j][inputDim] : -1);
       }
 
-      // Find the max dim size (excluding -1) for broadcast validation.
+      // NumPy broadcasting: dims of 1 or -1 can broadcast, others must match.
       int64_t maxDimSize = -1;
       for (int64_t dimSize : dimSizes) {
         if (dimSize != -1 && dimSize != 1) {
@@ -429,12 +435,11 @@ private:
           }
         }
       }
-      // If all dims are 1 or -1, use 1 as max.
       if (maxDimSize == -1) {
         maxDimSize = 1;
       }
 
-      // Determine which inputs need broadcast and set affine exprs.
+      // Set affine expr: constant 0 for broadcast dims, dim expr otherwise.
       for (size_t j = 0; j < numInputs; ++j) {
         const int inputDim = inputRanks[j] + i;
         const bool needsBcast =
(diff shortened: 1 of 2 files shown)


lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
```diff
@@ -351,6 +351,10 @@ using ComputeOpMap = OpMap<
   std::pair<d2m::TileMatmulOp,      std::pair<ttkernel::MatmulInitOp,              ttkernel::MatmulTilesOp>>,
   std::pair<d2m::TileMatmulBlockOp, std::pair<ttkernel::MatmulBlockInitOp,         ttkernel::ExperimentalMatmulBlockOp>>,
 
+  // Reductions FPU
+  std::pair<d2m::TileReduceSumOp,   std::pair<ttkernel::ComputeKernelHWStartupOp, ttkernel::ReduceTileOp>>,
+  std::pair<d2m::TileReduceMaxOp,   std::pair<ttkernel::ComputeKernelHWStartupOp, ttkernel::ReduceTileOp>>,
```
wenbinlyuTT: Column alignment of the ops.

wenbinlyuTT (approved):

lib/Dialect/D2M/Transforms/OpScheduler.cpp
```diff
@@ -182,6 +193,21 @@ struct D2MOpSchedulerRewriter final : public OpRewritePattern<GenericOp> {
...
+      unsigned leftWeight = node->left->weight;
+      unsigned middleWeight = node->middle->weight;
+      unsigned rightWeight = node->right->weight;
+
+      unsigned maxWeight = std::max({leftWeight, middleWeight, rightWeight});
+      unsigned equalCount = (leftWeight == maxWeight) +
+                            (middleWeight == maxWeight) +
+                            (rightWeight == maxWeight);
+      // If multiple children have max weight, we need extra register.
+      node->weight = maxWeight + (equalCount > 1 ? 1 : 0);
```
mbagherbeikTT: Small fix for this as you'd need 2 extra registers if all 3 children have the same weight
node->weight = maxWeight + (equal_count - 1)

dlokeTT: Actually realized this was wrong. For ternaries, it should be max(w[0], 1 + max(w[1], 1 + w[2])). It evaluates heaviest first, store (1 reg), evaluate next heaviest(while holding 1), store (2 regs), evaluate lightest (while holding 2). Redid this in latest commit

lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
```diff
@@ -369,65 +369,81 @@ class D2MNamedElementwiseRewriter final
...
+          } else {
+            TT_assertv(dimSize == maxDimSize,
+                       "Incompatible bcast dims {} & {}.", dimSize, maxDimSize);
+          }
+        }
+      }
+      // If all dims are 1 or -1, use 1 as max.
+      if (maxDimSize == -1) {
+        maxDimSize = 1;
+      }
```
wenbinlyuTT, line 440: This isn't equivalent to the logic before the change. Originally I wanted to just crash for cases like `lhsDimSize = 1 & rhsDimSize = -1`, basically when the tensors have different ranks.

The new logic is essentially doing implicit unsqueeze, and I guess it's OK? If this causes some issue later then the IR is probably wrong to begin with.

Please leave a comment stating this could be an implicit unsqueeze.

lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
```diff
@@ -369,65 +369,81 @@ class D2MNamedElementwiseRewriter final
...
-
-      deducedShape[outDim] = std::max(lhsDimSize, rhsDimSize);
+      // Gather dim sizes for all inputs (-1 for non-existent dims).
+      SmallVector<int64_t> dimSizes;
+      for (size_t j = 0; j < numInputs; ++j) {
+        const int inputDim = inputRanks[j] + i;
+        dimSizes.push_back((inputDim >= 0) ? inputShapes[j][inputDim] : -1);
+      }
+
+      // Find the max dim size (excluding -1) for broadcast validation.
```
wenbinlyuTT: The new logic here looks good, but please also adapt the old comment here so we also have a natural language description of the restriction.

wenbinlyuTT (approved):
GJ on the bcast support!
BTW, a few lit tests would be good.

lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
dlokeTT: Yea this is following numpy semantics. Will add comment

PR #6284, opened 2025-12-11
[ttnn.jit] deallocate intermediate tensors properly

vtangTT:
### Ticket
Link to Github Issue

### Problem description
Running into OOM issues on longer tests.

### What's changed
- we were not inserting deallcoates into the final IR; added `createTTNNDeallocate` to the pipeline
- always set `retain=True` for any tensors passed into the jit function; these input tensors are owned by the caller so we will never deallocate them in mlir runtime

### Checklist

### DIFF
--- a/runtime/python/runtime/utils.cpp
+++ b/runtime/python/runtime/utils.cpp
@@ -47,8 +47,8 @@ void registerRuntimeUtilsBindings(nb::module_ &m) {
       [](tt::runtime::Tensor tensor) -> nb::object {
         ::ttnn::Tensor &ttnn_tensor =
             ::tt::runtime::ttnn::utils::getTTNNTensorFromRuntimeTensor(tensor);
-        py::object py_tensor =
-            py::cast(ttnn_tensor, py::return_value_policy::copy);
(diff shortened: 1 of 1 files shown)


arminaleTT (approved):

runtime/python/runtime/utils.cpp
```diff
@@ -47,8 +47,8 @@ void registerRuntimeUtilsBindings(nb::module_ &m) {
       [](tt::runtime::Tensor tensor) -> nb::object {
         ::ttnn::Tensor &ttnn_tensor =
             ::tt::runtime::ttnn::utils::getTTNNTensorFromRuntimeTensor(tensor);
-        py::object py_tensor = py::cast(std::move(ttnn_tensor),
-                                        py::return_value_policy::reference);
+        py::object py_tensor =
+            py::cast(ttnn_tensor, py::return_value_policy::copy);
```
jameszianxuTT: Sorry, why are we making this change?

vtangTT: as per our offline discussion, this change has been reverted ty 🙏

tapspatel (approved):

PR #6286, opened 2025-12-11
Lower collective_permute to TTNN and fix open loop behavior

### LINKED ISSUE #4323: Collective Permute produces incorrect output
### Describe the bug

Collective Permute produces incorrect results when the source_target_pairs form an “open loop.”

**Example**
Input tensors:
```
tensor[0] = [1, 2, 3, 4]
tensor[1] = [5, 6, 7, 8]
```
Operation:
```
%out = "ttnn.collective_permute"(%3, %0) <{source_target_pairs = dense<[[0, 1]]> : ...
```

Expected output:
```
tensor[0] = [0, 0, 0, 0]  // no source writes to device 0
tensor[1] = [1, 2, 3, 4]  // device 1 receives data from device 0
```

Actual test output:
```
tensor[0] = [0, 0, 0, 0]
tensor[1] = [0, 0, 0, 0]
```

The data from tensor[0] is lost instead of being transferred to tensor[1].

### Steps to reproduce the issue

```

import torch

from typing import List, Tuple
from ttir_builder.utils import compile_to_flatbuffer
from ttir_builder import Operand, TTIRBuilder, Shape

def pseudo_golden_collective_permute(
    input_tensor: torch.Tensor,
    source_target_pairs: List[Tuple[int, int]],
):
    shards = list(torch.chunk(input_tensor, 2, dim=1))
    permuted_tensor = [torch.zeros_like(shard) for shard in shards]
    for source, target in source_target_pairs:
        permuted_tensor[target] = shards[source].clone()
    result_tensor = torch.cat(permuted_tensor, dim=1)
    return result_tensor

def collective_permute(in0: Operand, builder: TTIRBuilder):
    pair = [(0, 1)]
    input = builder._get_golden_tensor(in0)
    golden_output = pseudo_golden_collective_permute(input, pair)
    builder.set_graph_input_output([input], [golden_output])

    sharded = builder.mesh_shard(
        in0,
        shard_direction="#ttcore.shard_direction<full_to_shard>",
        shard_type="#ttcore.shard_type<devices>",
        shard_shape=(1, 2),
        shard_dims=(-1, 1),
    )
    reduced = builder.collective_permute(
        sharded,
        source_target_pairs=pair,
    )
    return builder.mesh_shard(
        reduced,
        shard_direction="#ttcore.shard_direction<shard_to_full>",
        shard_type="#ttcore.shard_type<devices>",
        shard_shape=(1, 2),
        shard_dims=(-1, 1),
    )

compile_to_flatbuffer(
    collective_permute,
    [(32, 32)],
    mesh_shape=(1, 2),
    test_base="CollectivePermute",
    )
```

### Expected behavior

Golden Match

_No response_

### LINKED ISSUE #6285: Relocate Collective Permute support
ttnn does not support the `collective permute` op, but we need to support `stablehlo.collective_permute`. So we added a fallback implementation in the runtime. We might have been able to implement it in the `TTIRtoTTNN` pass or the `StableHLOtoTTIR` pass, but at that time we did not have the PointToPoint op.
Now that we have PointToPoint op support, we can move the fallback implementation to the TTIRtoTTNN pass as a decomposition.

hkwonTT:
### Ticket
closes #6285 
closes #4323

### Problem description
 - `ttnn` does not support a native `collective_permute` op, but we still need to support `stablehlo.collective_permute`. Today this is handled by a fallback implementation in the runtime instead of in the compiler pipeline.  ￼
 - The current collective_permute implementation produces incorrect results when `source_target_pairs` form an open loop. 
 - `ttnn.point_to_point` can only send tensors between `MeshCoordinates` that share the same row or the same column.

### What's changed
 - Lower ttir.collective_permute to existing TTNN ops (point_to_point, assign, etc.) in TTIRToTTNN.
 - Implement correct StableHLO behavior for open loops and devices with no incoming edge (they get zeros).
 - Add a simple workaround to break a 2D hop into two 1D hops so point_to_point can route across 2D meshes.
 - Remove the old TTNN collective_permute op and related runtime / emitC / emitPy code.
 - Update and extend tests for collective_permute to cover these cases.

### Checklist

### DIFF
--- a/lib/Conversion/TTIRToTTNN/TTIRToTTNN.cpp
+++ b/lib/Conversion/TTIRToTTNN/TTIRToTTNN.cpp
@@ -2063,10 +2063,31 @@ class CollectivePermuteOpConversionPattern
 public:
   using OpConversionPattern<ttir::CollectivePermuteOp>::OpConversionPattern;
 
+private:
+  // Helper function to create a zero tensor using ClampScalarOp.
+  // We need a zero tensor with the same layout as the input. ttnn::ZerosOp
+  // cannot accept all required layout attributes (specifically, Alignment
+  // attribute), and ttnn::ZerosLikeOp does not exist in the TTNN dialect,
+  // so we use ClampScalarOp as a workaround.
+  mlir::Value createZerosTensor(ttir::CollectivePermuteOp op,
+                                mlir::Value inputTensor,
+                                ttnn::MemoryConfigAttr memoryConfigAttr,
+                                ConversionPatternRewriter &rewriter) const {
+    FloatAttr zeroAttr =
+        FloatAttr::get(Float32Type::get(rewriter.getContext()), 0.0f);
+    return rewriter
+        .create<ttnn::ClampScalarOp>(
+            op.getLoc(), this->getTypeConverter()->convertType(op.getType()),
+            inputTensor, zeroAttr, zeroAttr, memoryConfigAttr)
+        .getResult();
+  }
+
+public:
   LogicalResult
   matchAndRewrite(ttir::CollectivePermuteOp op, OpAdaptor adaptor,
                   ConversionPatternRewriter &rewriter) const override {
-    auto inputType = mlir::cast<RankedTensorType>(adaptor.getInput().getType());
+    RankedTensorType inputType =
+        mlir::cast<RankedTensorType>(adaptor.getInput().getType());
     ttnn::TTNNLayoutAttr layoutAttr =
         mlir::cast<ttnn::TTNNLayoutAttr>(inputType.getEncoding());
     // MemoryConfigAttr only exists if memLayout is not null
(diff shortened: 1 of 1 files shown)


tools/tt-alchemist/templates/python/local/utils.py
```diff
@@ -7,7 +7,6 @@
 
 
 # Monkey-patch ttnn with ttnn_supplemental objects
-ttnn.collective_permute = ttnn_supplemental.collective_permute
 
 
 class DeviceGetter:
```
svuckovicTT, line 8:
```suggestion
import math


class DeviceGetter:
```

tools/tt-alchemist/templates/python/local/ttnn_supplemental.py
svuckovicTT, line 1: Remove the whole `ttnn_supplemental.py` file and wherever it is `import`ed

lib/Conversion/TTIRToTTNN/TTIRToTTNN.cpp
```diff
@@ -2025,12 +2025,98 @@ class CollectivePermuteOpConversionPattern
...
+    ttnn::MemoryConfigAttr memoryConfigAttr = ttnn::MemoryConfigAttr::get(
+        op.getContext(), layoutAttr.getMemLayout(),
+        ttnn::BufferTypeAttr::get(op.getContext(), layoutAttr.getBufferType()),
+        std::nullopt);
+    ttcore::DataTypeAttr dTypeAttr = ttcore::DataTypeAttr::get(
+        rewriter.getContext(), layoutAttr.getDataType());
 
-    auto device = ::ttnn::utils::getOrInsertDevice(rewriter, op);
+    auto sourceTargetPairs = adaptor.getSourceTargetPairs();
+    auto sourceTargetPairsValues = sourceTargetPairs.getValues<int64_t>();
```
svuckovicTT: It's not obvious to me what the return types are, I would replace `auto` with actual types here.

lib/Conversion/TTIRToTTNN/TTIRToTTNN.cpp
```diff
@@ -2025,12 +2025,98 @@ class CollectivePermuteOpConversionPattern
...
+          adaptor.getInput(), zeroAttr, zeroAttr, memoryConfigAttr);
+      return success();
+    }
 
-    rewriter.replaceOpWithNewOp<ttnn::CollectivePermuteOp>(
-        op, this->getTypeConverter()->convertType(op.getType()),
-        adaptor.getInput(), device, adaptor.getSourceTargetPairs());
+    auto baseTensor = rewriter.create<ttnn::AssignOp>(
+        op.getLoc(), this->getTypeConverter()->convertType(op.getType()),
+        adaptor.getInput(), memoryConfigAttr, dTypeAttr);
```
svuckovicTT: Should this be:
```suggestion
    auto baseTensor = rewriter.create<ttnn::AssignOp>(
        op.getLoc(), this->getTypeConverter()->convertType(op.getType()),
        adaptor.getInput(), memoryConfigAttr, dTypeAttr).getResult();
```

The `auto` kind of obscures whether we're returning an op or a value, I've noticed this before but never looked into it - I guess that depends on the semantics of the op?

tools/tt-alchemist/templates/python/local/utils.py
svuckovicTT: The suggestion mechanism seems to be broken... What I meant to say here is: let's delete the monkey-patch comment and keep 2 empty lines as per PEP8.

lib/Dialect/TTNN/Transforms/Workarounds/Decomposition/PointToPointOpRewritePattern.cpp
```diff
@@ -0,0 +1,54 @@
...
+  if ((senderCoord[0] == receiverCoord[0]) ||
+      (senderCoord[1] == receiverCoord[1])) {
+    // If Row or Column are the same, no decomposition is needed.
+    return failure();
+  }
+
+  // Decompose the PointToPointOp into a sequence of PointToPointOps
+  ::llvm::SmallVector<int64_t, 2> intermediateCoordVec{senderCoord[0],
+                                                       receiverCoord[1]};
+  ::llvm::ArrayRef<int64_t> intermediateCoord = intermediateCoordVec;
```
jameszianxuTT, line 36: This is very speculative premature optimization - but it would be interesting to set the intermediateCoordVec to either of the diagonals `{senderCoord[0], receiverCoord[1]}` or `{receiverCoord[0], senderCoord[1]}` based on a global counter %2 or something

Because, if we have some train of collective permute from the same point A -> B with 2D hop, there are 2 paths around the "rectangle" from A->B , but we're just arbitrarily loading one of them.

I think this decomposition is definitely not the place to model noc congestion though, stateful compilation seems like a horrible idea as well, just an idle comment :)

PR #6287, opened 2025-12-11
D2M: Add 1D Matmul heuristic support

jdesousa-TT:
### Problem description
We currently run out of memory / utilize a fraction of the grid when we have high aspect ratio input tensors.

### What's changed
- Fixes lowering of virtual grid tensors to make sure operands that share dimensions match grid layouts
- Inserts views to ensure the generic sees the input tensors as matching (with each other and with the output)
- Fixes multicast lowering such that we can support both 2D multicasts (requires loopback) and 1D multicasts (optimally doesn't want loopback)
- Hard-codes CB initialization shapes to include the full device grid, instead of pulling from memref shapes which may have block factors that increase grid size beyond max worker grid

### Checklist

### DIFF
--- a/lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
+++ b/lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
@@ -1186,8 +1186,8 @@ public:
             op.getLoc(), rewriter.getI32Type(), numDestsIdx);
         auto mcastAddr =
             rewriter.create<ttkernel::ExperimentalGetNocMulticastAddrOp>(
-                op.getLoc(), virtX, virtY, mcastEndX, mcastEndY, dstL1Start,
-                nullptr);
+                op.getLoc(), virtX, virtY, virtMcastEndX, virtMcastEndY,
+                dstL1Start, nullptr);
         if (op.getIsLoopback()) {
           rewriter.create<ttkernel::NocAsyncWriteMulticastLoopbackSrcOp>(
               op.getLoc(), srcL1Start, mcastAddr, transferSize, numDests,


lib/Dialect/D2M/Transforms/GenericGenerateDatamovement.cpp
```diff
@@ -176,41 +195,52 @@ class D2MGenericGenerateDatamovementRewriter
...
     Value receiversReadySemaphore = createSemaphore(builder, loc, regions);
     Value senderFinishedSemaphore = createSemaphore(builder, loc, regions);
-    assert(mcastArgs.mcastCoreIndex.size() == mcastArgs.mcastShape.size());
-    assert(mcastArgs.conditions.size() == 1 &&
-           "Exactly one condition supported");
+    assert(mcastArgs.senderCoreIndex.size() == mcastArgs.mcastShape.size());
+    assert(mcastArgs.conditions.size() == 1 ||
+           mcastArgs.conditions.size() == 2 &&
+               "Exactly one or two conditions supported");
+    // build compound condition using all conditions generically !!
```
jdesousa-TT: fix me

lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
```diff
@@ -1182,15 +1185,15 @@ class D2MDMAWriteRewriter : public OpConversionPattern<d2m::DMAWriteOp> {
             rewriter.create<ttkernel::ExperimentalGetNocMulticastAddrOp>(
                 op.getLoc(), virtX, virtY, mcastEndX, mcastEndY, dstL1Start,
                 nullptr);
-        if (adaptor.getSrc() == adaptor.getDst()) {
+        if (op.getIsLoopback()) {
           // If src and dst refer to the same memref, we do not loopback mcast
```
jdesousa-TT: remove this comment

lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
```diff
@@ -1513,6 +1518,8 @@ class D2MSemaphoreUpdateRewriter : public OpConversionPattern<ConcreteOp> {
              "d2m.semaphore_set to single remote core is illegal.");
       auto [virtY, virtX] = getVirtualCoordsFromLogicalCoords(
           rewriter, op.getLoc(), chipDesc, op.getDstCoreIndex());
+      rewriter.create<ttkernel::DPrintOp>(
```
jdesousa-TT: rm

test/python/golden/test_metal_matmul.py
```diff
@@ -206,3 +206,63 @@ def test_matmul_ttnn_shapes_double_buffered(
...
+    ids=shape_str,
+)
+@pytest.mark.parametrize("target", ["ttmetal"])
+def test_matmul_1d_shapes(
+    shape: tuple[int, ...],
+    target: str,
+    request,
+    device,
+):
+    def matmul(
```
jdesousa-TT: clean this up

include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
```diff
@@ -806,12 +806,12 @@ def D2M_DMAOp : D2M_GenericRegionDatamovementOp<"dma",
       %tx = d2m.dma %src, %dst : (memref<1x1x6x8x!ttcore.tile<32x32, f32>, #ttcore.shard<...>, #dram>, memref<6x8x!ttcore.tile<32x32, f32>, #l1_>) -> !d2m.mem_tx
       ```
 
-      - Local to mcast, e.g. starting at offset core[1, 2] with mcast shape [4, 4] (src and dst have the same SSA value, implies NoC doesn't loopback)
+      - Local to mcast, e.g. starting at offset core[1, 2] with mcast shape [4, 4] (isLoopback is false)
```
nsmithtt, line 809: I'm ok with making loopback explicit, but I don't quite understand how 1d matmul is different than 2d in this regard.  IMO 1d means that only 1 side (either the LHS or RHS) is doing mcast, the other side is not.

lib/Dialect/D2M/Transforms/GenericGenerateDatamovement.cpp
```diff
@@ -142,28 +144,45 @@ class D2MGenericGenerateDatamovementRewriter
...
-        Value gridDimMinusOne = rewriter.create<arith::ConstantOp>(
-            loc, rewriter.getIndexType(), rewriter.getIndexAttr(numDests));
-        assert(iteratorType == ttcore::IteratorType::Reduction);
-        args.senderCoreIndex.push_back(zero);
-        args.mcastCoreIndex.push_back(one);
-        args.mcastShape.push_back(gridDimMinusOne);
-        args.mcastVolume *= numDests;
-
-        Value condition = rewriter.create<arith::CmpIOp>(
+        args.senderCoreIndex.push_back(Value(zero));
```
nsmithtt, line 173: Aren't there 2 cases for `!outputShardLayout.getCoreVirtualizationMap().isIdentity()` as well? There's the Parallel (non-mcast) case as well right?

lib/Dialect/D2M/Transforms/GenericGenerateDatamovement.cpp
```diff
@@ -142,28 +144,45 @@ class D2MGenericGenerateDatamovementRewriter
...
     for (auto [dim, iteratorType] : llvm::enumerate(mcastIterators)) {
       Value core = rewriter.create<CoreIndexOp>(
           loc, rewriter.getIndexType(), rewriter.getI64IntegerAttr(dim));
-      if (iteratorType == ttcore::IteratorType::Parallel) {
-        args.senderCoreIndex.push_back(Value(core));
-        args.mcastCoreIndex.push_back(Value(core));
-        args.mcastShape.push_back(Value(one));
+      auto outputShardLayout = mlir::cast<ttcore::ShardLayoutAttr>(
+          ttcore::getDeviceLayout(outputOperand));
+      if (outputShardLayout.getCoreVirtualizationMap().isIdentity()) {
```
nsmithtt, line 152: Can we avoid the extra conditional? Seems like we could just combine the mcast case regardless of it being identity or not.  If it's identity then the physical grid shape happens to match the virtual, if it's not identity then physical grid shape still returns the right thing.

include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
jdesousa-TT: 1D matmul does imply that only one of the operands is mcast, the issue is that we mcast that operand to all of the worker cores. Imagine core (0,0) is the sender, and we are mcasting the LHS to all cores, then the mcast grid is either 8x8 (which has to include the sender => loopback required), or we need to issue 2 mcasts ((0,1) -> (0,7) and (1,0) to (7,7)) and not use loopback. This is more complicated than just including the full worker grid and the sender core, which is why we use loopback for this case.

lib/Dialect/D2M/Transforms/GenericGenerateDatamovement.cpp
jdesousa-TT: Since we only support `1 by x` or `x by 1` grids for virtualization, it makes the parallel and reduction case the same. The sender is always (0,0) and the shape is always the respective dimension. 

I'm not sure if this generalizes to some weird cases we discussed, for example a `2 by x` virtual grid, where we mcast some part of the operand to half the grid, and some other part to the other half of the grid though. This is not really on the radar in terms of near-term support however.

include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
nsmithtt: Is there any win by issuing 2 mcasts?  It seems like it'd almost always be more performant and simpler to issue just one mcast over a bounding rectangle that covers the 1d grid.   Non-participating cores will get some data that's unused, but that shouldn't be an issue because of lockstep allocation.

lib/Dialect/D2M/Transforms/GenericGenerateDatamovement.cpp
nsmithtt: That makes sense for the mcast case, but what if you're an eltwise op using a 1d grid, shouldn't the mcast shape be 1 in that case for all operands?

include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
jdesousa-TT: Yes, I agree. I can't see a reason to issue two disjoint multicasts. I went with a single mcast, bounding box including the full **2D** grid, but this implies we have to include the sender, and use the loopback API for this.

Just for clarity in the terminology (referencing the terms TTNN uses):
- "2D Matmul" is made up of **two** 1D multicasts (one for each operand)
- "1D Matmul" is made up of **one** 2D multicast (same operand cast in both dimensions)

nsmithtt: > I went with a single mcast, bounding box including the full 2D grid, but this implies we have to include the sender, and use the loopback API for this

Perhaps I'm missing something, but I think loopback is only required if your src and dst buffers are not at the same address.  The way we code-gen mcast is there's one core (0,0) in this case that gathers to its own input at say address 0x8000, then it mcasts that buffer to all others in the grid at the same address 0x8000 and _without_ loopback because it's already gathered the data to its buffer at 0x8000.

lib/Dialect/D2M/Transforms/GenericGenerateDatamovement.cpp
jdesousa-TT: In the eltwise case, we wouldn't run this code at all, since there's no mcast dimensions. 

`calculateMcastIterators` returns an empty `mcastIterators` when all dims are parallel and then we just do default DMA generation.

line 95:
```cpp
return allParallel ? SmallVector<ttcore::IteratorType>() : mcastIterators;
```

line 266:
```cpp
bool isMcast = !mcastIterators.empty();
      if (isMcast) {
        createGatherMcastDMA(builder, loc, src, dst, outputOperand,
                             operandIndexingMap, grid, mcastIterators, regions);
      } else {
        Value memTx =
            createDMA(builder, loc, src, dst, operandIndexingMap, isOutput);
        builder.create<d2m::DMAWaitOp>(loc, memTx);
      }
 ```

File: include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td (hunk 5 above)
Comment (nsmithtt) [reply include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td:809]: All loopback means is that this NoC cmd field is programmed `NOC_CMD_BRCST_SRC_INCLUDE`, which tells the noc if it should include the src core in the copy or not.  So if you mcast to a rectangle without this bit set, there will be a hole in the rectangle at the source core.

File: lib/Dialect/D2M/Transforms/GenericGenerateDatamovement.cpp (hunk 6 above)
Comment (nsmithtt) [reply lib/Dialect/D2M/Transforms/GenericGenerateDatamovement.cpp:173]: Ok, I guess in a fusion case then where you're doing matmul and then you have a bunch of fused eltwise inputs that run afterwards, those additional operands will have parallel iterator type.

File: lib/Dialect/D2M/Transforms/GenericGenerateDatamovement.cpp (hunk 7 above)
Comment (jdesousa-TT) [reply lib/Dialect/D2M/Transforms/GenericGenerateDatamovement.cpp:152]: There's some special casing going on, before when we had an identity map (i.e. before virtual grids), we would force the parallel dim mcast shape to be 1:
```cpp
args.mcastShape.push_back(Value(one));
```

If we want to default to always use the physical shape of the dim, this works for the 1D case, but not the 2D case anymore, since we will get something bigger than 1 for the parallel dim too.

I suppose the "correct" way to handle this might be to set the mcast shape in virtual coordinates, (i.e. force 1 on the  parallel dim even for 1D mcast, then the mcast shape will be 1x64 for 1D) then apply the virtual to physical mapping on the mcast shape after the fact?

include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
jdesousa-TT: Hmm, I didn't look into the implementation details of the NoC APIs very deeply.

Played around with this some more and it seems I misunderstood the limitations of the regular `noc_async_write_multicast` API. I'll blame the documentation because it says:
```
* With this API, the multicast sender cannot be part of the multicast
 * destinations. If the multicast sender has to be in the multicast
 * destinations (i.e. must perform a local L1 write), the other API variant
 * *noc_async_write_multicast_loopback_src* can be used.
 ```
 
I assumed this mean you cannot include the src core in the range to mcast to. After experimenting, it seems to work even if the src core is included in the bounding box but NOT counted in `num_dests`. 
 
That being the case, I suppose we don't need loopback here and we can make the convention that our mcast core ranges always include the sender, and we decrement mcast volume by 1 when setting `num_dests` for non-loopback mcasts...?

The reason I say make this the convention across the board, is when we lower to TTKernel we have to calculate the number of destinations based on shape, and I'm not sure it would be easy/possible to determine at that level whether or not the range includes whatever core this executes on.

Comment (jdesousa-TT) [reply include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td:809]: See #6477, revised version with the aforementioned convention.

Comment (nsmithtt) [reply include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td:809]: Yeah that seems like a good convention since it'll make all of out mcast lowerings uniform.

File: lib/Dialect/D2M/Transforms/GenericGenerateDatamovement.cpp (hunk 7 above)
Comment (nsmithtt) [reply lib/Dialect/D2M/Transforms/GenericGenerateDatamovement.cpp:152]: Yeah I see what you mean, I was definitely oversimplifying. My original thinking is that we apply the virtual shape to the map to get back a physical shape.

My pseudo code attempt (note parallel iterator type remains unchanged regardless of there being a virtual grid):
```
      if (iteratorType == ttcore::IteratorType::Parallel) {
        args.senderCoreIndex.push_back(Value(core));
        args.mcastCoreIndex.push_back(Value(core));
        args.mcastShape.push_back(Value(one));
      } else {
        auto physicalGridShape = map.apply(virtualGridShape);
        int64_t numDests = physicalGridShape[dim];
        Value gridDim = rewriter.create<arith::ConstantOp>(
            loc, rewriter.getIndexType(), rewriter.getIndexAttr(numDests));
        assert(iteratorType == ttcore::IteratorType::Reduction);
        args.senderCoreIndex.push_back(zero);
        args.mcastCoreIndex.push_back(zero); // I guess this should be updated to zero now
        args.mcastShape.push_back(gridDim);
        args.mcastVolume *= numDests;
        // As you said, mcastVolume needs to be fixed up during ttkernel lowering instead of here.
        ...
```

PR #6288, opened 2025-12-11
Simplify const eval hoist

mtopalovicTT:
Changing the way we analyze graph for const eval ops. This way we can remove a lot of code.

### DIFF
--- a/lib/Transforms/ConstEvalHoist.cpp
+++ b/lib/Transforms/ConstEvalHoist.cpp
@@ -113,9 +113,9 @@ public:
   }
 
   ConstEvalAnalysisResults getAnalysisResults() {
-    llvm::DenseMap<mlir::Value, std::size_t> rootToId;
-    llvm::DenseMap<std::size_t, ConstEvalSubgraph> idToSubgraph;
-    std::size_t currentId = 0;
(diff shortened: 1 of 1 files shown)


lib/Transforms/ConstEvalHoist.cpp
```diff
@@ -51,109 +106,58 @@ namespace {
...
+        constParams(mlir::tt::ttcore::getConstsAndParams(funcOp)) {
     buildConstEvalSubgraphs();
   }
 
-  // Recursively collapse subsets into subgraphs by iterating over ops in order
-  // and grouping them based on root idx; also collect sets of params
-  // dependencies.
   ConstEvalAnalysisResults getAnalysisResults() {
-    llvm::SmallVector<ConstEvalSubgraph, 4> finalSubgraphs;
+    llvm::DenseMap<mlir::Value, std::size_t> rootToId;
```
vwellsTT: nit: I think std::size_t is ugly personally :) but if you want to use, should use everywhere I guess, there are regular `size_t` in this PR too

vwellsTT:
Great cleanup!

lib/Transforms/ConstEvalHoist.cpp
mtopalovicTT: Sure

Issue #6248, opened 2025-12-09
[D2M] On BH, logical coordinates are not converted to translated coordinates on device correctly


bgrady-tt:
### Describe the bug

Blackhole's translated coordinates place worker cores on a _discontinuous_ grid. 

D2M generates kernel code that dynamically converts logical coordinates on device to translated coordinates for DMA indexing logic. The current code assumes that a simple offset adjustment to XY coordinate is enough, but that appears to be incorrect for Blackhole.

sohaibnadeemTT:
also causes https://github.com/tenstorrent/tt-mlir/issues/5341

Issue #5341, opened 2025-10-14
D2M matmul tests of TTNN shapes hang on P150


wenbinlyuTT:
As of 9f0e66b, the D2M `test_matmul_ttnn_shapes_single/double_buffered` tests hang on P150.
All other test configurations are not affected.

sohaibnadeemTT:
This is fixed btw, closing.

PR #6305, opened 2025-12-12
Fix for BH coord translation issue

sohaibnadeemTT:
### Ticket
Link to Github Issue: https://github.com/tenstorrent/tt-mlir/issues/6248, https://github.com/tenstorrent/tt-mlir/issues/5341

### Problem description
Translated coordinates in BH are not contiguous so we cannot use a pair of translation offsets to convert coordinates from LOGICAL to TRANSLATED system.

### What's changed
We instead use the worker_logical_col_to_virtual_col and worker_logical_row_to_virtual_row tables (which are global arrays on device) to do this conversion.

### Checklist

### DIFF
--- a/include/ttmlir/Dialect/TTKernel/IR/TTKernelOps.td
+++ b/include/ttmlir/Dialect/TTKernel/IR/TTKernelOps.td
@@ -2575,6 +2575,34 @@ def TTKernel_InterleavedAddrGenFastGetNocAddrOp : TTKernel_Op<"interleaved_addr_
     }];
 }
 
+def TTKernel_MyXOp : TTKernel_Op<"my_x", [Pure]> {
+    let summary = "MyX";
+    let description = [{
+      Lowers to the tt-metal supported MY_X macro. This represents the virtual X coordinate of the current core.
+    }];
+
+    let arguments = (ins Optional<I8>:$noc);
+    let results = (outs Index:$x);
+
+    let assemblyFormat = [{
+      `(` ($noc^)? `)` attr-dict `:` functional-type(operands, results)
+    }];
+}
+
+def TTKernel_MyYOp : TTKernel_Op<"my_y", [Pure]> {
+    let summary = "MyY";
+    let description = [{
+      Lowers to the tt-metal supported MY_Y macro. This represents the virtual Y coordinate of the current core.
+    }];
+
+    let arguments = (ins Optional<I8>:$noc);
+    let results = (outs Index:$y);
+
+    let assemblyFormat = [{
+      `(` ($noc^)? `)` attr-dict `:` functional-type(operands, results)
+    }];
+}
+
 def TTKernel_MyLogicalXOp : TTKernel_Op<"my_logical_x_", [Pure]> {
     let summary = "MyLogicalX";
     let description = [{
--- a/lib/Target/TTKernel/TTKernelToCpp.cpp
+++ b/lib/Target/TTKernel/TTKernelToCpp.cpp
@@ -43,6 +43,8 @@ public:
 
     builder->create<emitc::IncludeOp>(loc, "tools/profiler/kernel_profiler.hpp",
                                       /*isStandard=*/false);
+    builder->create<emitc::IncludeOp>(loc, "firmware_common.h",
+                                      /*isStandard=*/false);
 
     if (threadType == ThreadType::Noc) {
 
--- a/lib/Conversion/TTKernelToEmitC/TTKernelToEmitC.cpp
+++ b/lib/Conversion/TTKernelToEmitC/TTKernelToEmitC.cpp
@@ -930,6 +930,10 @@ public:
     patterns.add<TTKernelInvokeSFPIOpRewriter>(typeConverter,
                                                funcOp.getContext());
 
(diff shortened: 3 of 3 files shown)


vmilosevic (commented):

lib/Target/TTKernel/TTKernelToCpp.cpp
```diff
@@ -11,6 +11,7 @@
 #include "ttmlir/Target/TTKernel/LLKs/experimental_matmul_llks_generated.h"
 #include "ttmlir/Target/TTKernel/LLKs/experimental_tilize_llks_generated.h"
 #include "ttmlir/Target/TTKernel/LLKs/experimental_untilize_llks_generated.h"
+#include "ttmlir/Target/TTKernel/LLKs/experimental_coord_translation_generated.h"
```
vmilosevic: :warning: **llvm\-include\-order** :warning:
```suggestion

```

include/ttmlir/Dialect/TTKernel/IR/TTKernelOps.td
```diff
@@ -2624,34 +2575,62 @@ def TTKernel_InterleavedAddrGenFastGetNocAddrOp : TTKernel_Op<"interleaved_addr_
     }];
 }
 
-def TTKernel_MyXOp : TTKernel_Op<"my_x", [Pure]> {
-    let summary = "MyX";
+def TTKernel_MyLogicalXOp : TTKernel_Op<"my_logical_x_", [Pure]> {
```
nsmithtt, line 2606: Polymage and other third parties still use my_x, I think we should keep my_x/y for parity and additionally introduce the new op w/ new behavior.

include/ttmlir/Target/TTKernel/LLKs/experimental_coord_translation.h
```diff
@@ -0,0 +1,28 @@
+// SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#ifndef TTMLIR_TARGET_TTKERNEL_LLKS_EXPERIMENTAL_COORD_TRANSLATION_H
+#define TTMLIR_TARGET_TTKERNEL_LLKS_EXPERIMENTAL_COORD_TRANSLATION_H
+
```
nsmithtt, line 7: We must be getting the header transitively somehow, perhaps we should `#include "firmware_common.h"` here though.

include/ttmlir/Dialect/TTKernel/IR/TTKernelOps.td
sohaibnadeemTT: I don't think that's a good idea. my_x/my_y is TRANSLATED and we currently convert my_x/my_y to LOGICAL when lowering d2m::CoreIndexOp (see D2MCoreIndexRewriter::matchAndRewrite) and then back later which is redundant. We should just be using the logical coordinates directly.

include/ttmlir/Target/TTKernel/LLKs/experimental_coord_translation.h
sohaibnadeemTT: `"firmware_common.h"` is included in all firmware (see `tt_metal/hw/firmware/src/tt-1xx/brisc.cc`) similar my_x/my_y, should we still be including it explicitly?

include/ttmlir/Dialect/TTKernel/IR/TTKernelOps.td
nsmithtt: This will break third party code that rely on these functions.  ttkernel dialect is meant to reflect the metal programming api, not be prescriptive on which parts of it you should or shouldn't use.  We can suggest to them to use logical x/y instead, but they may have reasons for sticking with my_x and my_y.

include/ttmlir/Target/TTKernel/LLKs/experimental_coord_translation.h
nsmithtt: Maybe we should use `"common_globals.h"` which seems to be the api header for these apis.

include/ttmlir/Dialect/TTKernel/IR/TTKernelOps.td
sohaibnadeemTT: Yeah that makes sense, I'll add back TTKernel_MyXOp/TTKernel_MyYOp too

sohaibnadeemTT: added

include/ttmlir/Target/TTKernel/LLKs/experimental_coord_translation.h
sohaibnadeemTT: added

sohaibnadeemTT:
@nsmithtt Going back from `common_globals.h` to `firmware_common.h` since the former is meant for compute/trisc kernels (includes "ckernel.h" and other stuff only meant for compute kernels)

sohaibnadeemTT:
image

test_metal_matmul passing on p150 in CI

PR #6318, opened 2025-12-13
[ttnn.jit] lit test infra

vtangTT:
### Problem description
We want to be able to check mlir emitted from jit python frontend, without having to run it e2e.

### What's changed
Solution: lit test the python files that have jit decorated w/ `compile_only=True`
- added basic lit tests for basic unary, binary, matmul, reduction ops
- moved `test_control_flow.py` to be completely lit test
- moved `test_math_fidelity.py` to be completely lit test

### Checklist

### DIFF
--- a/.github/test_scripts/ttnn_jit.sh
+++ b/.github/test_scripts/ttnn_jit.sh
@@ -35,7 +35,7 @@ else
         pytest -v $WORK_DIR/test/ttnn-jit/test_mesh_tensor_eltwise.py $WORK_DIR/test/ttnn-jit/test_matmul_smoketest.py --junit-xml=$TEST_REPORT_PATH
     else
         # Only run tests in the top level directory. These are always run.
-        llvm-lit -v --xunit-xml-output $TEST_REPORT_PATH $WORK_DIR/test/ttnn-jit/lit
+        llvm-lit -v --xunit-xml-output $TEST_REPORT_PATH $BUILD_DIR/test/ttnn-jit/lit
         pytest -v $WORK_DIR/test/ttnn-jit/*.py --junit-xml=$TEST_REPORT_PATH
     fi
 fi


ckaravasilisTT:
LIT test infrastructure LGTM! A few minor comments.

test/ttnn-jit/lit/test_ops.py
```diff
@@ -0,0 +1,78 @@
...
+    # CHECK: func.func @abs
+    # CHECK: "ttnn.abs"(%arg0) {ttnn.hoist_generic_via_d2m}
+    _ = abs(input_tensor_a_dram)
+
+    # CHECK: ---- IR Dump after TTIRCompiler (AST-based) ----
+    # CHECK: func.func @add
+    # CHECK: "ttnn.add"{{.*}} <{dtype = #ttcore.supportedDataTypes<bf16>}> {ttnn.hoist_generic_via_d2m}
+    _ = add(input_tensor_a_l1, input_tensor_b_l1)
+
+    # CHECK:---- IR Dump after GraphToIRTranslator (Graph-based) ----
```
ckaravasilisTT: space after `CHECK:`

test/ttnn-jit/lit/test_ops.py
```diff
@@ -0,0 +1,78 @@
...
+# REQUIRES: ttnn-jit
+
+import ttnn_jit
+import ttnn
+import torch
+
+from utils import create_sharded_tile_tensor, create_dram_tensor
+
+
+@ttnn_jit.jit(debug=True, compile_only=True, graph_capture=False)
```
ckaravasilisTT: is `debug=True` redundant? If so, maybe remove or add it to `test_math_fidelity.py` also for consistency

tools/ttnn-jit/_src/jit.py
```diff
@@ -41,7 +41,7 @@ def __init__(
         self.func = func
         self.source_code = cleanup_source_code(func)
         self.compile_only = compile_only
-        self.debug = debug
+        self.debug = debug | compile_only
```
sgholamiTT: I'm not a python expert but I think using `bitwise or` (`|`) for flags is a bit uncommon (and not safe, eg. If later someone changes `debug: Optional[bool]`, you'll get a type error for this line.)

If we want to enable debug when `compile_only` is on, I'd recommend: 
```python
self.debug = debug or compile_only
```
Feel free to ignore if you think otherwise.

test/ttnn-jit/lit/test_ops.py
vtangTT: fixed ty

test/ttnn-jit/lit/test_ops.py
vtangTT: fixed ty

tools/ttnn-jit/_src/jit.py
vtangTT: makes sense to me.. fixed ty

PR #6323, opened 2025-12-15
[fix] modify toolchain venv build before activating

woruyu:
### Summary
In setion `3. Please ensure that you do not`'s code block of file `getting-started.md`, readme don't support #include syntax, so just change it to text version.

### DIFF
--- a/docs/src/getting-started.md
+++ b/docs/src/getting-started.md
@@ -86,7 +86,8 @@ sudo chown -R "${USER}" "${TTMLIR_TOOLCHAIN_DIR}"
 3. Please ensure that you do not already have an environment (venv) activated before running the following commands:
 
 ```bash
-{{#include ../../../.github/Dockerfile.ci:environment_build}}
+cmake -B env/build env
+cmake --build env/build
 source env/activate
 ```
 


nsmithtt:
Hi @woruyu, thank you for the contribution, but I'm not sure I understand the fix here.  The purpose of the existing mdbook inline code is to paste in working / tested code.  It also reads identically to the change you propose here:

Screenshot 2025-12-15 at 8 58 21 AM

Please let me know if I'm misunderstanding something here, thanks!

See generated docs: https://docs.tenstorrent.com/tt-mlir/getting-started.html#setting-up-the-environment-manually

We are using mdbook and its md extensions, so not all docs will be rendered perfectly outside of mdbook.

woruyu:
@nsmithtt , Thanks for the clarification — I now understand that the {{#include}} syntax is intentional and works correctly when rendered via mdBook.

My original concern came from the user flow: following the TT-MLIR build instructions link from the rendered docs currently redirects to the GitHub README, where the mdBook include syntax is not expanded. From a reader’s perspective, this can be confusing and appear misleading, even though the mdBook output itself is correct.

Would it make sense to update the link target to point directly to the rendered mdBook documentation instead of the GitHub markdown, so readers stay within the intended rendering environment?
image
image

nsmithtt:
Hi @woruyu, thank you for bringing this to our attention!  This is a bad link on the tt-xla docs, it should link here: https://docs.tenstorrent.com/tt-mlir/getting-started.html#setting-up-the-environment-manually.

Do you want to open a PR against XLA to make this change?  No worries if not and we can do it.

woruyu:
> Hi @woruyu, thank you for bringing this to our attention! This is a bad link on the tt-xla docs, it should link here: https://docs.tenstorrent.com/tt-mlir/getting-started.html#setting-up-the-environment-manually.
> 
> Do you want to open a PR against XLA to make this change? No worries if not and we can do it.

Thanks for confirming. Here is PR, https://github.com/tenstorrent/tt-xla/pull/2606

nsmithtt:
> Thanks for confirming. Here is PR, [tenstorrent/tt-xla#2606](https://github.com/tenstorrent/tt-xla/pull/2606)

Thank you @woruyu!  Closing this in favor the linked tt-xla PR above.

PR #6326, opened 2025-12-15
[TTIRFusing] Add RMSNormFusionPattern

mvasiljevicTT:
### Ticket
[Link to Github Issue](https://github.com/tenstorrent/tt-mlir/issues/4588)

### Problem description
Pytorch llama models use rms_norm as:
```
x * torch.rsqrt(x.pow(2).mean(-1, keepdim=True) + self.eps)
```
This should be fused in TTIR as [RMSNormOp](https://docs.tenstorrent.com/tt-metal/latest/ttnn/ttnn/api/ttnn.rms_norm.html#ttnn.rms_norm).

### What's changed
- Added `RMSNormFusionPattern` that rewrites `(x * rsqrt(mean(x^2) + epsilon)) * gamma` if mean is done over last dimension as  `rms_norm <epsilon> (x, gamma)`.

- After that in some cases we end up with pattern:
```
reshape   [BT, D] -> [1, BT, D]
|
rms_norm  [1, BT, D]
|
reshape   [1, BT, D] -> [BT, D]
```
which we rewrite with `TTIRCommuteReshapeThroughRMSNorm` as:
```
rms_norm [BT, D]
```
This is done if the reshapes are inverse and don't change last dim, as `rms_norm` is always done over last dim in the current metal implementation.

- When compiling models with rms_norm with `optimizer_level=2`, running `getOpConstraints` for rms_norm with sharded input crashes in tt-metal with floating-point exception. 
This will be fixed with [this pr in tt-metal](https://github.com/tenstorrent/tt-metal/pull/34335). 
Until then, sharding for rms_norm is turned off.

- Added `EXPECT_DEATH` test that calls `getOpConstraints` for rms_norm with sharded input. 
Once [the tt-metal fix](https://github.com/tenstorrent/tt-metal/pull/34335) is merged and uplifted, the test will fail, signaling that we should bring back sharding for rms_norm.

- Using basic bf16 `rms_norm` caused [pcc erros in some models](https://github.com/tenstorrent/tt-xla/actions/runs/20299537098), which is why `RMSNormConfigRewritePattern` sets compute config for max percision. 

Regressions that remain:

| Model | Current PCC | Target PCC |
|-------|-------------|------------|
| `albert/masked_lm/pytorch-base_v2` | 0.9866 | 0.99 |
| `gpt_neo/causal_lm/pytorch-gpt_neo_1_3B` | 0.9889 | 0.99 |
| `llama/causal_lm/pytorch-llama_3_2_3b` | 0.9895 | 0.99 |
| `mistral/pixtral/pytorch` | 0.9256 | 0.96 |

- Added helper methods `lookThrough`, `findOpThrough`, `lookThroughLayoutOps` and refactored previous patterns where they could be used.

### Checklist
- [x] Merged fix in tt-metal
- [ ] [tt-xla nightly](https://github.com/tenstorrent/tt-xla/actions/runs/20345372899)

### DIFF
--- a/lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
+++ b/lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
@@ -3650,6 +3650,31 @@ public:
 
     llvm::SmallVector<int64_t> normalizedShape{inputType.getShape().back()};
 
+    // Reshape gamma to match normalized_shape if needed.
+    // Gamma may have extra dimensions (e.g., [1, 1, 2048] instead of [2048]).
+    auto gammaType = mlir::cast<RankedTensorType>(gamma.getType());
+    if (gammaType.getShape() != llvm::ArrayRef(normalizedShape)) {
+      // Check if gamma can be squeezed to normalized_shape.
+      // All dimensions except the last must be 1.
+      for (int64_t i = 0; i < gammaType.getRank() - 1; ++i) {
+        if (gammaType.getShape()[i] != 1) {
+          return mlir::failure();
+        }
+      }
+      // Check if the last dimension matches.
+      if (gammaType.getShape().back() != normalizedShape[0]) {
+        return mlir::failure();
+      }
+      // Reshape gamma to normalized_shape.
+      auto reshapedGammaType = RankedTensorType::get(
+          normalizedShape, gammaType.getElementType(), gammaType.getEncoding());
+      llvm::SmallVector<int32_t> targetShape(normalizedShape.begin(),
+                                             normalizedShape.end());
+      gamma = rewriter.create<ReshapeOp>(outerMul.getLoc(), reshapedGammaType,
+                                         gamma,
+                                         rewriter.getI32ArrayAttr(targetShape));
+    }
+
     // Create RMSNormOp with output shape matching input shape
     auto rmsNormOutputType =
         RankedTensorType::get(inputType.getShape(), outputType.getElementType(),


lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
```diff
@@ -3637,6 +3557,148 @@ class SDPAFusing : public mlir::OpRewritePattern<MatmulOp> {
...
+
+// Folds reshape pairs around RMSNormOp when the last dim is unchanged.
+class RMSNormReshapeFoldPattern : public mlir::OpRewritePattern<RMSNormOp> {
+  using mlir::OpRewritePattern<RMSNormOp>::OpRewritePattern;
+
+public:
+  mlir::LogicalResult
+  matchAndRewrite(RMSNormOp op, mlir::PatternRewriter &rewriter) const final {
+    auto inputReshape = op.getInput().getDefiningOp<ReshapeOp>();
+    if (!inputReshape || !op->hasOneUse()) {
```
odjuricicTT: Nit: you can check `!op->hasOneUse()` before initializing `inputReshape`

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
```diff
@@ -3637,6 +3557,148 @@ class SDPAFusing : public mlir::OpRewritePattern<MatmulOp> {
...
+    if (!meanOp) {
+      meanOp = addOp.getRhs().getDefiningOp<MeanOp>();
+      epsilon = addOp.getLhs();
+    }
+    if (!meanOp) {
+      return mlir::failure();
+    }
+
+    // TTNN RMS norm only supports normalization over the last dimension.
+    auto dimArg = meanOp.getDimArg();
```
odjuricicTT, line 3605: Not a blocker for this PR, but this is one more reason to move these fusing to TTNN IR, so we don't have to model TTNN ops on TTIR level.

cc: @mtopalovicTT

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
```diff
@@ -3637,6 +3557,148 @@ class SDPAFusing : public mlir::OpRewritePattern<MatmulOp> {
...
+    rewriter.replaceOpWithNewOp<RMSNormOp>(
+        outerMul, outerMul.getType(), x, gamma,
+        /*bias=*/nullptr, rewriter.getDenseI64ArrayAttr(normalizedShape),
+        rewriter.getF32FloatAttr(epsAttr.getValue().convertToFloat()));
+    return mlir::success();
+  }
+};
+
+// Folds reshape pairs around RMSNormOp when the last dim is unchanged.
+class RMSNormReshapeFoldPattern : public mlir::OpRewritePattern<RMSNormOp> {
```
odjuricicTT: Have you considered solving this problem by supporting reshape <-> rmsnorm commute in EIO?

test/ttmlir/Dialect/TTIR/fusing/rms_norm_fusion.mlir
```diff
@@ -0,0 +1,110 @@
+// RUN: ttmlir-opt --canonicalize --ttir-fusing %s -o %t.mlir
+// RUN: FileCheck %s --input-file=%t.mlir
+
+module {
+    // First RMS norm from llama - input comes from reshape (32x2048 -> 32x1x2048)
+    // Reshape pairs around rms_norm should be folded away.
+    // CHECK-LABEL: func.func @rms_norm_fusion_1
+    func.func @rms_norm_fusion_1(%arg0: tensor<32x2048xbf16>, %arg1: tensor<2048xbf16>) -> tensor<32x2048xbf16> {
+        // CHECK: %[[RESULT:.*]] = "ttir.rms_norm"(%arg0, %arg1)
```
odjuricicTT, line 9: Could you add this check to the single layer tests as well?
https://github.com/tenstorrent/tt-mlir/blob/main/test/ttmlir/Silicon/TTNN/n150/optimizer/models_perf_tests/llama_3_2_1b_decoder.mlir

test/unittests/OpModel/TTNN/Op/TestRMSNormWidthSharded.cpp
```diff
@@ -0,0 +1,71 @@
...
+class OpModelTest : public OpModelFixture {};
+
+// Test that RMS norm with width-sharded input causes a crash in metal.
+// This test verifies the workaround in DFShardingPolicy.cpp is still needed.
+// We are waiting for the following metal fix to be uplifted:
+// https://github.com/tenstorrent/tt-metal/pull/34335
+//
+// IMPORTANT: If this test FAILS (code doesn't crash), it means the metal fix
+// has been uplifted. In that case:
+// 1. Add RMSNormOp back to validForSharding in DFShardingPolicy.cpp
```
odjuricicTT, line 25: Awesome! Thanks for adding.

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
mvasiljevicTT: Added upwards commute pattern for reshape and rms norm, which is enough to erase inverse reshapes. 
We could also add downwards, but in llama layer only case of that is 
```
reshape [1, BT, D] -> [BT, D]
|
rms_norm [BT, D]
|
matmul [BT, D]...
```
so moving it downwards would just increase time of that rms_norm.

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
mvasiljevicTT: removed in new version

test/ttmlir/Dialect/TTIR/fusing/rms_norm_fusion.mlir
mvasiljevicTT: done

PR #6328, opened 2025-12-15
Support for rearrange op

nsmithtt:
This PR adds preliminary support for the rearrange op. Remaining work includes more robust pattern parsing and affine map conversion, supporting groups on the LHS of the ein expression.

rearrange op definition / spec: https://einops.rocks/api/rearrange/

### DIFF
--- a/test/python/golden/ttir_ops/eltwise/test_ttir_binary.py
+++ b/test/python/golden/ttir_ops/eltwise/test_ttir_binary.py
@@ -295,7 +295,7 @@ scalar_binary_ops = [
 @pytest.mark.parametrize("dtype", [torch.float32], ids=["f32"])
 @pytest.mark.parametrize("target", ["ttmetal"])
 @pytest.mark.parametrize(
-    "test_fn,scalar",
+    "test_fn,scalar_value",
     scalar_binary_ops,
     ids=[
         "add_scalar",
@@ -307,7 +307,7 @@ scalar_binary_ops = [
 )
 def test_scalar_binary_ops(
     test_fn: Callable,
-    scalar: float,
+    scalar_value: float,
     shape: Shape,
     dtype: torch.dtype,
     target: str,


lib/Dialect/TTIR/IR/TTIROps.cpp
```diff
@@ -1756,6 +1756,214 @@ ::mlir::OpFoldResult mlir::tt::ttir::ReshapeOp::fold(FoldAdaptor adaptor) {
...
+::mlir::LogicalResult mlir::tt::ttir::RearrangeOp::verify() {
+  llvm::StringRef patternStr = getPattern();
+  llvm::SmallVector<llvm::StringRef> parts;
+  patternStr.split(parts, "->");
+
+  if (parts.size() != 2) {
+    return emitOpError() << "pattern must contain exactly one '->' separator.";
+  }
+
+  mlir::FailureOr<AffineMap> failureOrMap = getInvPatternMap();
```
vwellsTT: nit: imo this var name should specificy it's the invMap, not just "Map"

lib/Dialect/TTIR/IR/TTIROps.cpp
```diff
@@ -1756,6 +1756,214 @@ ::mlir::OpFoldResult mlir::tt::ttir::ReshapeOp::fold(FoldAdaptor adaptor) {
...
+  if (failed(failureOrMap)) {
+    return emitOpError() << "failed to parse pattern " << patternStr << ".";
+  }
+
+  auto map = *failureOrMap;
```
vwellsTT: ditto here

lib/Dialect/TTIR/IR/TTIROps.cpp
```diff
@@ -1756,6 +1756,214 @@ ::mlir::OpFoldResult mlir::tt::ttir::ReshapeOp::fold(FoldAdaptor adaptor) {
...
+    // % h_size).
+    bool needsMod =
+        group.size() > 1; // mod only required for collapsed groups, we elide it
+                          // otherwise to reduce clutter.
+    int64_t stride = 1;
+    mlir::AffineExpr expr = mlir::getAffineConstantExpr(0, context);
+    for (int64_t i = static_cast<int64_t>(group.size()) - 1; i >= 0; --i) {
+      unsigned dimPos = dimToInputPos[group[i]];
+      expr = mlir::getAffineDimExpr(groupPos, context).floorDiv(stride);
+      if (needsMod) {
```
wenbinlyuTT: I think this still leads to a redundant modulo on the outermost dim if group size is > 2.
Simply check `if (i > 0) ` makes the ir look better.

lib/Dialect/TTIR/IR/TTIROps.cpp
```diff
@@ -1756,6 +1756,214 @@ ::mlir::OpFoldResult mlir::tt::ttir::ReshapeOp::fold(FoldAdaptor adaptor) {
...
+  SmallVector<int64_t> lastInputIndex = map.compose(lastOutputIndex);
+  SmallVector<int64_t> expectedInputShape;
+  for (int64_t i = 0; i < map.getNumResults(); i++) {
+    expectedInputShape.push_back(lastInputIndex[i] + 1);
+  }
+
+  if (getInput().getType().getShape() !=
+      ArrayRef<int64_t>(expectedInputShape)) {
+    return emitOpError() << "input tensor shape ("
+                         << ttmlir::utils::join(
```
vwellsTT, line 1792: nit: maybe better to just use llvm's `interleaveComma`?

lib/Dialect/TTIR/IR/TTIROps.cpp
```diff
@@ -1756,6 +1756,214 @@ ::mlir::OpFoldResult mlir::tt::ttir::ReshapeOp::fold(FoldAdaptor adaptor) {
...
+
+  return success();
+}
+
+mlir::FailureOr<::mlir::AffineMap>
+mlir::tt::ttir::RearrangeOp::getInvPatternMap(mlir::MLIRContext *context,
+                                              StringRef pattern,
+                                              ArrayRef<int64_t> shape) {
+  // We need to write a routine to convert the pattern string to an affine map.
+  // Example patterns:
```
vwellsTT, line 1807: I think this comment is great for specifying all these examples, but I'm curious if this is sort of a custom affine-like syntax you came up with, or based on some torch behavior, or what?
I wonder if we should have some specialized utility files for parsing these, depending on how much complexity is eventually expected

lib/Dialect/TTIR/IR/TTIROps.cpp
```diff
@@ -1756,6 +1756,214 @@ ::mlir::OpFoldResult mlir::tt::ttir::ReshapeOp::fold(FoldAdaptor adaptor) {
...
+      }
+
+      if (c == ' ') {
+        i++;
+        continue;
+      }
+
+      // Parse dimension name
+      size_t start = i;
+      while (i < pattern.size() && pattern[i] != ' ' && pattern[i] != ')' &&
```
vwellsTT, line 1886: something like this feels like maybe we should use more sophisticated token-reading tech say

lib/Dialect/TTIR/IR/TTIROps.cpp
```diff
@@ -1756,6 +1756,214 @@ ::mlir::OpFoldResult mlir::tt::ttir::ReshapeOp::fold(FoldAdaptor adaptor) {
...
+
+  auto inputDims = parseDims(inputPattern);
+  auto outputDims = parseDims(outputPattern);
+
+  // Build a map from dimension name to input position
+  llvm::DenseMap<llvm::StringRef, unsigned> dimToInputPos;
+  unsigned pos = 0;
+  for (const auto &group : inputDims) {
+    if (group.size() > 1) {
+      // Unsupported input groups.
```
vwellsTT: nit: this isn't clear to me (and I suppose isn't a "proper sentence" technically, just an adj + noun); I guess it means: "Input groups are currently unsupported"?

lib/Dialect/TTIR/IR/TTIROps.cpp
```diff
@@ -1756,6 +1756,214 @@ ::mlir::OpFoldResult mlir::tt::ttir::ReshapeOp::fold(FoldAdaptor adaptor) {
...
+        i++;
+        continue;
+      }
+
+      if (c == ' ') {
+        i++;
+        continue;
+      }
+
+      // Parse dimension name
```
vwellsTT: nit: .

lib/Dialect/TTIR/IR/TTIROps.cpp
```diff
@@ -1756,6 +1756,214 @@ ::mlir::OpFoldResult mlir::tt::ttir::ReshapeOp::fold(FoldAdaptor adaptor) {
...
+      result.push_back(currentGroup);
+    }
+
+    return result;
+  };
+
+  auto inputDims = parseDims(inputPattern);
+  auto outputDims = parseDims(outputPattern);
+
+  // Build a map from dimension name to input position
```
vwellsTT: nit: .

lib/Dialect/TTIR/IR/TTIROps.cpp
```diff
@@ -1756,6 +1756,214 @@ ::mlir::OpFoldResult mlir::tt::ttir::ReshapeOp::fold(FoldAdaptor adaptor) {
...
+    }
+
+    for (llvm::StringRef dim : group) {
+      dimToInputPos[dim] = pos;
+    }
+
+    pos++;
+  }
+
+  // Build the affine expressions for the output
```
vwellsTT: nit: .

lib/Dialect/TTIR/IR/TTIROps.cpp
```diff
@@ -1756,6 +1756,214 @@ ::mlir::OpFoldResult mlir::tt::ttir::ReshapeOp::fold(FoldAdaptor adaptor) {
...
+  llvm::SmallVector<mlir::AffineExpr> exprs;
+  exprs.resize(inputDims.size(), nullptr);
+  for (const auto [groupPos, group] : llvm::enumerate(outputDims)) {
+    assert(!group.empty());
+    // For flattening like b h -> (b h)@d, we create inverse map: (d / h_size, d
+    // % h_size).
+    bool needsMod =
+        group.size() > 1; // mod only required for collapsed groups, we elide it
```
vwellsTT: nit: M

lib/Dialect/TTIR/IR/TTIROps.cpp
```diff
@@ -1756,6 +1756,214 @@ ::mlir::OpFoldResult mlir::tt::ttir::ReshapeOp::fold(FoldAdaptor adaptor) {
...
+  llvm::SmallVector<llvm::StringRef> parts;
+  pattern.split(parts, "->");
+
+  assert(parts.size() == 2 &&
+         "RearrangeOp pattern must contain exactly one '->' separator.");
+
+  llvm::StringRef inputPattern = parts[0].trim();
+  llvm::StringRef outputPattern = parts[1].trim();
+
+  // Helper lambda to parse dimension names from a pattern
```
vwellsTT: nit: .

lib/Dialect/TTIR/IR/TTIROps.cpp
nsmithtt: Perhaps it should read `i > 0 && needsMod`?  I'm also not confident you can elide the mod in the i==0 case because this map will be composed with other maps which may need the mod to be correct.

include/ttmlir/Dialect/TTIR/IR/TTIROps.td
```diff
@@ -3895,6 +3895,23 @@ def TTIR_ReshapeOp: TTIR_NamedOp<"reshape", [TTIR_TensorManipulation]> {
     let hasFolder = 1;
 }
 
+def TTIR_RearrangeOp : TTIR_NamedOp<"rearrange", [TTIR_TensorManipulation]> {
+    let summary = "Tensor rearrange operation.";
+    let description = [{
```
bgrady-tt, line 3900: Just tagging that this needs a beefy description.

lib/Dialect/TTIR/IR/TTIROps.cpp
nsmithtt: This is `einops.rearrange` syntax, part of the unit test asserts that this affine map performs the exact transform as the einop rearrange. https://einops.rocks/api/rearrange/

ttnn models team explicitly requested this op to be supported, it plugs into our affinemap infra pretty nicely.

lib/Dialect/TTIR/IR/TTIROps.cpp
nsmithtt: @anuragsingh-tt is going to build this out into a proper parser that can support the full rearrange syntax.

lib/Dialect/TTIR/IR/TTIROps.cpp
wenbinlyuTT: If `i > 0` happened then group size must be larger than 1, `needsMod` is thus always true.

I have a vague feeling that if eliminating a redundant mod in one affine map leads to bugs after composition, the root cause is somewhere else... Like out-of-range indices from the output of the other map.

lib/Dialect/TTIR/IR/TTIROps.cpp
nsmithtt: One downside of `interleaveComma` is you need to create an output string on the stack to read into.

lib/Dialect/TTIR/IR/TTIROps.cpp
bgrady-tt: This is a nit for sure, but I agree that we should just lex alphanumeric identifiers that start with `_[a-zA-Z]`. Everything else should be an error. A simple loop here is fine IMO.

lib/Dialect/TTIR/IR/TTIROps.cpp
bgrady-tt: Fwiw, I have some code on branch that can generically eliminate redundant modulo in affine maps using range analysis. So I would vote to keep the affine map generation simple; we can optimize away redundant mods/floordiv elsewhere.

wenbinlyuTT: Perhaps we just don't do this ourselves. Maybe MLIR can simplify the affine maps based on known input rage constraints before the DMA passes, claude told me we have tools like:
```
mlir::simplifyAffineMap() - Basic simplification
mlir::composeAffineMapAndOperands() - Compose and simplify
FlatAffineValueConstraints - For range-based reasoning
mlir::affine::normalizeAffineMap() - Normalize representation
```
Right now the TM affine maps are very complicated and I constantly see stuff like `(d0 floordiv 32) mod 32` or `(d0 * 64 + d1) % 8` lowered all the way into the kernel cpp files.

Hopefully MLIR has built-in support for these and we don't need Z3 or some other dependencies.

lib/Dialect/TTIR/IR/TTIROps.cpp
```diff
@@ -1756,6 +1756,214 @@ ::mlir::OpFoldResult mlir::tt::ttir::ReshapeOp::fold(FoldAdaptor adaptor) {
...
+  }
+
+  SmallVector<int64_t> lastOutputIndex;
+  for (int64_t dim : getResult().getType().getShape()) {
+    lastOutputIndex.push_back(dim - 1);
+  }
+  SmallVector<int64_t> lastInputIndex = map.compose(lastOutputIndex);
+  SmallVector<int64_t> expectedInputShape;
+  for (int64_t i = 0; i < map.getNumResults(); i++) {
+    expectedInputShape.push_back(lastInputIndex[i] + 1);
```
bgrady-tt: nit: I don't think this is correct if we allow any affine expressions with negation or subtraction. 

_We may never need this,_ but it is strictly better to sample all four corners of the output shape (i.e. (0,0), (0,N), (M,0), and (M,N)) and find the max extent for each dim. For rect domain -> rect co-domain transforms this should always be enough.

lib/Dialect/TTIR/IR/TTIROps.cpp
nsmithtt: This is a map that is implementing reshape, once you have groups in rearrange it implies that you're merging dims so your sample space will overflow the modulus. I'll test with `i > 0 && needsMod`, but I'm not sure if it'll work.

lib/Dialect/TTIR/IR/TTIROps.cpp
bgrady-tt: Something akin to `torch.flip` would probably result in an affine map like this.

lib/Dialect/TTIR/IR/TTIROps.cpp
wenbinlyuTT: I have been using `i > 0` since rearrange3 came out and it's fine, it is equivalent to `i > 0 && needsMod`.

lib/Dialect/TTIR/IR/TTIROps.cpp
vwellsTT: well, it really takes a stream, not a string, right? But yeah idk how you'd fuse it with the emitOpError I guess off the top of my head, maybe not worth the hassle

lib/Dialect/TTIR/IR/TTIROps.cpp
vwellsTT: Gotcha, cool

lib/Dialect/TTIR/IR/TTIROps.cpp
vwellsTT: > @anuragsingh-tt is going to build this out into a proper parser that can support the full rearrange syntax.

nice

lib/Dialect/TTIR/IR/TTIROps.cpp
nsmithtt: Yeah you're right!

lib/Dialect/TTIR/IR/TTIROps.cpp
nsmithtt: Doesn't everything go out the window with subtraction in the picture?

lib/Dialect/D2M/Transforms/MaterializeViewReturns.cpp
```diff
@@ -130,19 +130,22 @@ class MaterializeViewReturnsPass
           // Pattern: %view = view_layout ... -> %host = to_host %view ->
           // return %host. We need to materialize the view BEFORE the
           // device-to-host transfer.
-          if (auto toHostOp =
-                  mlir::dyn_cast_if_present<d2m::ToHostOp>(definingOp)) {
-            Value toHostInput = toHostOp.getInput();
+          auto toLayoutOp =
+              mlir::dyn_cast_if_present<d2m::ToLayoutOp>(definingOp);
```
nsmithtt, line 134: @vwellsTT, I brought back the check for ToLayoutOp because this pass additionally runs before ToLayoutLowering.  Without this we end up with a stream going into a view which I could write a folder for, but since we were talking about moving away from a generic folder I figured this is the simpler change (and brings it back to the way it used to work).

vwellsTT: Gotcha, yeah that makes sense, I think we can clean it up as we go w/ stream changes

tapspatel (approved):

bgrady-tt (approved):
LGTM 👍

lib/Dialect/TTIR/IR/TTIROps.cpp
bgrady-tt: fwiw AffineMapUtils.h `applyMapToGrid() `does full bounds checking by sampling the entire map, and will work even if negative constants are used. But we can just update all these uses later if we need to.

PR #6334, opened 2025-12-15
Arguments for embedding in emitPy

djstefanovicTT:
### Ticket
closes https://github.com/tenstorrent/tt-mlir/issues/6333

### Problem description
EmitPy requires named arguments, but it was set as positional.

### What's changed
Added argument names for emitPy generation.

### DIFF
--- a/lib/Conversion/TTNNToEmitPy/TTNNToEmitPy.cpp
+++ b/lib/Conversion/TTNNToEmitPy/TTNNToEmitPy.cpp
@@ -2116,7 +2116,6 @@ public:
     llvm::SmallVector<mlir::Attribute> args{
         emitter.emit(embeddingOp.getInput()),
         emitter.emit(embeddingOp.getWeight()),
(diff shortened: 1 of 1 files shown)


lib/Conversion/TTNNToEmitPy/TTNNToEmitPy.cpp
```diff
@@ -2116,8 +2116,8 @@ class EmbeddingOpConversionPattern
     llvm::SmallVector<mlir::Attribute> args{
         emitter.emit(embeddingOp.getInput()),
         emitter.emit(embeddingOp.getWeight()),
-        emitter.emit(std::nullopt),
-        emitter.emit(layoutAttr),
+        emitter.emit(std::nullopt, "padding_idx"),
```
azecevic000: IMO we shouldn't even mention named arguments that we don't model at all. When they are positional, they are a necessary evil, but as a named argument, they don't provide any value and just clutter the generated code.

PR #6338, opened 2025-12-15
extending inner permutes (beyond 2D)

anuragsingh-tt:
### Problem description
Explained in title.

### What's changed
Enabled feature (updates to TTIR2D2M, GridSelection) and added pytest. 
And a minor CMake change to get rid of a warning during the build.

### DIFF
--- a/runtime/tools/chisel/CMakeLists.txt
+++ b/runtime/tools/chisel/CMakeLists.txt
@@ -2,8 +2,7 @@ add_custom_target(chisel
   COMMENT "python chisel package")
 
 add_custom_command(
-  TARGET chisel
-  POST_BUILD
   COMMAND python -m pip install -e .
+  TARGET chisel
   WORKING_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}"
 )
--- a/test/python/golden/test_metal_tms.py
+++ b/test/python/golden/test_metal_tms.py
@@ -64,5 +64,5 @@ def test_permute_abs(
         test_base=request.node.name,
         output_root=request.config.getoption("--path"),
         system_desc_path=request.config.getoption("--sys-desc"),
-        custom_pipeline=f"ttir-to-ttmetal-pipeline{{{' '.join(options)}}}",
+        custom_pipeline=f"ttir-to-ttmetal-pipeline{{{' '}}}",
     )


lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
```diff
@@ -1108,35 +1105,32 @@ class D2MPermuteRewriter final
...
-    // For inner permute, we need as streamLayout to do reblocking.
+    // For inner permute, we need a streamLayout to do reblocking.
     auto storage = rewriter.create<d2m::EmptyOp>(
         loc, permuted.physicalShape, inputTensorType.getElementType(),
         resultLayout);
     auto stream =
         rewriter.create<d2m::StreamLayoutOp>(loc, viewType, inputs[0], storage);
     inputs[0] = stream.getResult();
-
+    auto logicalRank = deviceRank / 2;
```
vwellsTT: nit: this shouldn't be an auto imo

runtime/tools/chisel/CMakeLists.txt
```diff
@@ -2,7 +2,8 @@ add_custom_target(chisel
   COMMENT "python chisel package")
 
 add_custom_command(
-  COMMAND python -m pip install -e .
   TARGET chisel
+  POST_BUILD
```
vwellsTT: why is this needed?

test/python/golden/test_metal_tms.py
```diff
@@ -0,0 +1,70 @@
...
+        [(32, 12, 100), [0, 2, 1]],
+        [(32, 11, 64), [0, 2, 1]],
+        # 4d inner permutes (llama-like)
+        [(1, 32, 12, 100), [0, 1, 3, 2]],
+        [(1, 32, 11, 64), [0, 1, 3, 2]],
+        [(1, 8, 11, 64), [0, 1, 3, 2]],
+    ],
+)
+@pytest.mark.parametrize("target", ["ttmetal"])
+def test_permute_abs(
```
vwellsTT, line 44: I wonder if: 1. we should have some outer dim permutes here too (or is that not supported yet? I thought it was simpler case but I might misremember) and 2. there's some redundancy w/ transpose tests I've added to the collapsing tests file--not sure one this, arguably, they are testing separate things so this is better

runtime/tools/chisel/CMakeLists.txt
anuragsingh-tt: There was a CMake warning during building (unrelated to permutes):

> CMake Warning (dev) ... (add_custom_command): Exactly one of PRE_BUILD, PRE_LINK, or POST_BUILD must be given. Assuming POST_BUILD to preserve backward compatibility. Policy CMP0175 is not set..

test/python/golden/test_metal_tms.py
anuragsingh-tt: 1. Outer dim permutes go through the TensorManipulatorRewriter as part of the rearrange PR; I will include those tests in `test_metal_tms.py` for that PR. 
2. Would it make sense to support collapsed-tensors=True on an inner permute for a 3D tensor?

lib/Dialect/D2M/Transforms/GridSelection.cpp
```diff
@@ -532,11 +532,26 @@ updateStreamLayoutOps(ArrayRef<StreamLayoutUpdateInfo> streamLayoutsToUpdate,
             storageLayout.getLogicalShape(), targetSquareGridShape,
             storageLayout.getNormalizedIntervals());
 
+    // If using a virtual grid, compute required forward index affine map.
+    AffineMap storageIndexMap = storageLayout.getIndexAffineMap();
+    if (info.isVirtualGrid) {
```
bgrady-tt, line 537: fwiw I don't think the storage tensor needs a correct index affine map. It shouldn't be used for anything.

runtime/tools/chisel/CMakeLists.txt
vwellsTT: I would marginally prefer if you could commit this in separate PR, it gets a bit messy bundling unrelated changes (though good on ya for fixing random things 😄 ) when trying to trace git history etc.

test/python/golden/test_metal_tms.py
vwellsTT: gotcha, sounds good for 1.
hmm, yeah I guess not--eventually, we ought to probably have some tests that we don't default collapse such cases I guess ha

lib/Dialect/D2M/Transforms/GridSelection.cpp
wenbinlyuTT: +1 on this, I think the affine map that defines the "actual work" of the `stream_layout` is recorded in the output tensor's layout. The storage tensor's map is usually `map(0)`.
This is probably also related to the `grid_selection` lit test change below, which I am having trouble to understand what's going on there.

anuragsingh-tt: I believe this change is needed. The tensor for the `%stream` created in TTIR2D2M and whose grid is selected in this pass is a physical allocation whose shard layout may sample core coordinates that are out of bounds if the virtual -> physical mapping is not accounted for. This index map is passed down as the `ShardLayoutAttr.coreVirtualizationMap`.

Example in final kernel (in a llama test case):
`%6 = "ttmetal.create_buffer"() <{address = 110848 : i64}> : () -> memref<16x4x1x2x1x1x!ttcore.tile<32x32, f32>, #ttcore.shard<4096x4096x4096, 1, (d0, d1, d2, d3, d4, d5) -> (((d1 + d2 + d0 * 4) floordiv 8) mod 8, (d1 + d2 + d0 * 4) mod 8, d3, d4, d5)>, #ttcore.memory_space<l1>>`

vs the incorrect:

`%6 = "ttmetal.create_buffer"() <{address = 110848 : i64}> : () -> memref<16x4x1x2x1x1x!ttcore.tile<32x32, f32>, #ttcore.shard<4096x4096x4096, 1, (d0, d1, d2, d3, d4, d5) -> (d0, d2, d1, d3, d5, d4)>, #ttcore.memory_space<l1>>`

bgrady-tt: If it works as-is, I am fine keeping this code 👍 It might be required to pass verification or somehow in the backend TTMetal passes.

Just fyi, the stream storage purpose is just allocating a shard-sized on the whole worker L1 grid; _the storage index map is never used (and shouldn't be)_. The grid shape and size doesn't matter to the underlying allocator; it always allocates the L1 the shape requirement is (imo) a mistake in the API.

nsmithtt: +1 the storage index map should never be referenced, seems like a deeper issue if it is.

Issue #6247, opened 2025-12-09
Runtime tilize on device fails with BLOCK_SHARDED target memory config


rpavlovicTT:
## Summary

After enabling FLOAT32 tilize on device in the runtime (commit 3fd5e1a65), tests that target BLOCK_SHARDED memory layout fail because tt-metal's `tilize_op.cpp` requires HEIGHT_SHARDED or INTERLEAVED memory layouts.

## Error

```
TT_FATAL: Input tensor memory layout must be HEIGHT_SHARDED but got TensorMemoryLayout::BLOCK_SHARDED
```

Location: `ttnn/operations/data_movement/tilize/device/tilize_op.cpp:42`

## Root Cause

The test function `@test_scalarize` has an input parameter with TILED + BLOCK_SHARDED layout:
```mlir
func.func @test_scalarize(%arg0: tensor<32x32xf32, #ttnn_layout1>) -> tensor<32x32xf32, #ttnn_layout1>
// where #ttnn_layout1 = TILED, L1, BLOCK_SHARDED
```

At runtime, when a host tensor (ROW_MAJOR) needs to be converted to this target layout, the `LayoutConverter::handleHostInputLayoutNoTypecast` does:

```cpp
if (shouldTilize && utils::canTilizeDataTypeOnDevice(outputDesc.dataType)) {
    ::ttnn::Tensor out = toDeviceIfNeeded(input, targetDevice);  // sends with BLOCK_SHARDED
    out = toLayoutIfNeeded(out);  // tilize fails here
    out = toMemoryConfigIfNeeded(out);
    return out;
}
```

The problem is `toDeviceIfNeeded` uses `outputDesc.memoryConfig` (BLOCK_SHARDED) immediately when sending to device, and then tilize fails because it doesn't support BLOCK_SHARDED.

## Proposed Fix

The `toDeviceIfNeeded` should NOT apply the target memory config immediately. Instead:

1. Send to device with INTERLEAVED memory config (or no specific config)
2. Perform tilize operation
3. Then explicitly change memory config to BLOCK_SHARDED via `toMemoryConfigIfNeeded`

The fix should be in `runtime/lib/ttnn/types/layout_converter.cpp`. The `handleHostInputLayoutNoTypecast` function (and similar functions) should be updated to:

```cpp
if (shouldTilize && utils::canTilizeDataTypeOnDevice(outputDesc.dataType)) {
    // Send to device without target memory config (use INTERLEAVED)
[...]

PR #6342, opened 2025-12-15
ttnn runtime fix: canTilizeOnDevice

vtangTT:
### Ticket
https://github.com/tenstorrent/tt-mlir/issues/6247

### Problem description
as described in issue

### What's changed
- add `canTilizeMemoryLayoutOnDevice`, only tilize for interleaved and height sharded
- re-enabled `ttnn_d2m_e2e.mlir`

### Checklist

### DIFF
--- a/runtime/lib/ttnn/types/layout_converter.cpp
+++ b/runtime/lib/ttnn/types/layout_converter.cpp
@@ -108,15 +108,16 @@ LayoutConverter::fromDeviceIfNeeded(const ::ttnn::Tensor &input) {
     return out;
   }
 
-  if (shouldTilize && utils::canTilizeOnDevice(outputDesc.dataType, outputDesc.memoryConfig)) {
+  if (shouldTilize &&
+      utils::canTilizeOnDevice(outputDesc.dataType, outputDesc.memoryConfig)) {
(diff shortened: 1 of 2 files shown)


runtime/lib/ttnn/types/layout_converter.cpp
```diff
@@ -108,14 +108,17 @@ ::ttnn::Tensor LayoutConverter::handleHostInputLayoutNoTypecast(
     return out;
   }
 
-  if (shouldTilize && utils::canTilizeDataTypeOnDevice(outputDesc.dataType)) {
+  if (shouldTilize && utils::canTilizeDataTypeOnDevice(outputDesc.dataType) &&
```
jameszianxuTT: I am not sure how possible it would be to wrap this something more graceful to represent the whole metal dtype + memory layout support matrix

eg. utils::canTilizeOnDevice(dataType, memoryConfig)

These branches are becoming pretty hard to read otherwise. It would also be a nicer way to express that there's some "support matrix" of datatype x memoryConfig that's supported in metal tilize/untilize on device in one place.

I don't really see a case where you'd query tilize support for dtype and mem layout separately also.

PR #6346, opened 2025-12-16
[ttnn.jit] Reorganize JIT Demo Tests

arminaleTT:
### Problem description
JIT demos are disorganized

### What's changed
- Move all demos under the demo folder
- Add a standalone demo for multiply-add

### DIFF
--- a/test/ttnn-jit/demo/test_multiply_accumulate.py
+++ b/test/ttnn-jit/demo/test_multiply_accumulate.py
@@ -10,18 +10,20 @@ import pytest
 from utils import all_close_check
 
 
+@ttnn_jit.jit(debug=True)
+def mul_add(input_tensor_a, input_tensor_b, input_tensor_c):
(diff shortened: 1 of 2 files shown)


test/ttnn-jit/demo/test_multiply_accumulate.py
```diff
@@ -0,0 +1,59 @@
...
+import ttnn
+import torch
+
+import pytest
+
+from utils import all_close_check
+
+
+def test_mul_add_demo(device):
+    @ttnn_jit.jit(debug=True)
```
vtangTT: can we decorate these function outside, to remain consistent with the digamma demo.

also ok with moving the digamma decoration inside, just want to be consistent

arminaleTT: Moved them outside

Issue #6099, opened 2025-12-01
Enable CPU-hoisted functions in EmitPy


dmilinkovicTT:
Currently, during the TTNN -> EmitPy, CPU module is being completely dismissed.

In order to support execution of the CPU-hoisted ops in EmitPy, we need to:
- Lower the CPU module to TTNN
- Use `ttnn.op.golden_function` in order to invoke Torch ops on the host

PR #6348, opened 2025-12-16
Enable CPU-hoisting on EmitPy

dmilinkovicTT:
### Ticket
#6099 

### Problem description
This PR implements CPU-hoisting functionality on EmitPy target by relying on TTNN library's golden function mechanism. Golden functions are functions attached to the TTNN ops which invoke equivalent Torch library operations, effectively executing the ops on the host CPU.

### What's changed
- Introduced `EmitPyLinkModules` pass
- Added golden mode support to the EmitPy conversion classes

### Checklist

### DIFF
--- a/test/ttmlir/EmitPy/cpu_hoisted_ops.mlir
+++ b/test/ttmlir/EmitPy/cpu_hoisted_ops.mlir
@@ -9,82 +9,82 @@
 // Verify CPU-hoisted functions are generated with golden_function calls
 
 // Binary arithmetic
-// CHECK-LABEL: def hoisted_ttir_add_32x32_32x32_func
+// CHECK-LABEL: def hoisted_ttir_add_32x32xf32_32x32xf32_func
 // CHECK: ttnn.add.golden_function
 
 // Binary division
-// CHECK-LABEL: def hoisted_ttir_div_32x32_32x32_func
+// CHECK-LABEL: def hoisted_ttir_div_32x32xf32_32x32xf32_func
 // CHECK: ttnn.divide.golden_function
 
 // Binary comparison
-// CHECK-LABEL: def hoisted_ttir_eq_32x32_32x32_func
+// CHECK-LABEL: def hoisted_ttir_eq_32x32xf32_32x32xf32_func
 // CHECK: ttnn.eq.golden_function
 
 // Binary logical
-// CHECK-LABEL: def hoisted_ttir_logical_and_32x32_32x32_func
+// CHECK-LABEL: def hoisted_ttir_logical_and_32x32xf32_32x32xf32_func
 // CHECK: ttnn.logical_and.golden_function
 
 // Binary min/max
-// CHECK-LABEL: def hoisted_ttir_maximum_32x32_32x32_func
+// CHECK-LABEL: def hoisted_ttir_maximum_32x32xf32_32x32xf32_func
 // CHECK: ttnn.maximum.golden_function
 
 // Matrix multiplication
-// CHECK-LABEL: def hoisted_ttir_matmul_32x64_64x32_func
+// CHECK-LABEL: def hoisted_ttir_matmul_32x64xf32_64x32xf32_func
 // CHECK: ttnn.matmul.golden_function
 
 // Unary activation
-// CHECK-LABEL: def hoisted_ttir_relu_32x32_func
+// CHECK-LABEL: def hoisted_ttir_relu_32x32xf32_func
 // CHECK: ttnn.relu.golden_function
 
 // Unary trigonometric
-// CHECK-LABEL: def hoisted_ttir_sin_32x32_func
+// CHECK-LABEL: def hoisted_ttir_sin_32x32xf32_func
 // CHECK: ttnn.sin.golden_function
 
 // Unary exponential/logarithmic
-// CHECK-LABEL: def hoisted_ttir_exp_32x32_func
+// CHECK-LABEL: def hoisted_ttir_exp_32x32xf32_func
 // CHECK: ttnn.exp.golden_function
 
-// CHECK-LABEL: def hoisted_ttir_log_32x32_func
+// CHECK-LABEL: def hoisted_ttir_log_32x32xf32_func
 // CHECK: ttnn.log.golden_function
 
 // Unary rounding
-// CHECK-LABEL: def hoisted_ttir_ceil_32x32_func
+// CHECK-LABEL: def hoisted_ttir_ceil_32x32xf32_func
 // CHECK: ttnn.ceil.golden_function
 
 // Unary sign/abs
-// CHECK-LABEL: def hoisted_ttir_abs_32x32_func
+// CHECK-LABEL: def hoisted_ttir_abs_32x32xf32_func
 // CHECK: ttnn.abs.golden_function
 
 // Unary reciprocal
-// CHECK-LABEL: def hoisted_ttir_reciprocal_32x32_func
+// CHECK-LABEL: def hoisted_ttir_reciprocal_32x32xf32_func
 // CHECK: ttnn.reciprocal.golden_function
 
 // Unary sqrt
-// CHECK-LABEL: def hoisted_ttir_sqrt_32x32_func
+// CHECK-LABEL: def hoisted_ttir_sqrt_32x32xf32_func
 // CHECK: ttnn.sqrt.golden_function
 
 // Pooling
-// CHECK-LABEL: def hoisted_ttir_max_pool2d_1x32x32x32_func
+// CHECK-LABEL: def hoisted_ttir_max_pool2d_1x32x32x32xf32_func
 // CHECK: ttnn.max_pool2d.golden_function
 
 // Reduction
-// CHECK-LABEL: def hoisted_ttir_sum_32x32_func
+// CHECK-LABEL: def hoisted_ttir_sum_32x32xf32_func
 // CHECK: ttnn.sum.golden_function
 
 // Data manipulation - reshape
-// CHECK-LABEL: def hoisted_ttir_reshape_32x32_func
+// CHECK-LABEL: def hoisted_ttir_reshape_32x32xf32_func
 // CHECK: ttnn.reshape.golden_function
 
 // Data manipulation - concat
-// CHECK-LABEL: def hoisted_ttir_concat_32x32_32x32_func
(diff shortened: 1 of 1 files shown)


lib/Dialect/TTCore/Transforms/TTCoreModuleWrap.cpp
```diff
@@ -5,20 +5,69 @@
...
 #define GEN_PASS_DEF_TTCOREUNWRAPDEVICEMODULEPASS
 #define GEN_PASS_DEF_TTCOREWRAPDEVICEMODULEPASS
+#define GEN_PASS_DEF_TTCORELINKCPUANDDEVICEMODULESPASS
 #include "ttmlir/Dialect/TTCore/Transforms/Passes.h.inc"
 
+namespace {
+
+// Helper function to unwrap a DeviceModuleOp into a top-level ModuleOp.
+//
+void unwrapDeviceModule(mlir::ModuleOp rootModule) {
```
dmilinkovicTT: note: this has been extracted from the `TTCoreUnwrapDeviceModulePass`, but the PR diff got interleaved

lib/Dialect/TTNN/Pipelines/TTNNPipelines.cpp
```diff
@@ -126,7 +126,7 @@ void createTTNNPipelineLoweringPasses(
   // Add pass to convert TTIR to TTNN.
   pm.addPass(createConvertTTIRToTTNNPass());
   // Add pass to remove unused values.
-  if (options.removeDeadValuesEnabled) {
+  if (removeDeadValuesEnabled) {
```
vmilosevic, line 130: :warning: **clang\-diagnostic\-error** :warning:
use of undeclared identifier `` removeDeadValuesEnabled ``

lib/Dialect/TTNN/Pipelines/TTNNPipelines.cpp
```diff
@@ -268,7 +268,7 @@ void createLowerTTIRToTTNNDevicePipeline(
     auto &devicePm = pm.nest<ttcore::DeviceModuleOp>().nest<mlir::ModuleOp>();
 
     // Run TTNN lowering passes on Device module.
-    createTTNNPipelineLoweringPasses(devicePm, options);
+    createTTNNPipelineLoweringPasses(devicePm, options.removeDeadValuesEnabled);
```
vmilosevic, line 266: :warning: **clang\-diagnostic\-error** :warning:
no matching function for call to `` createTTNNPipelineLoweringPasses ``

lib/Conversion/TTNNToEmitPy/EmitPyLinkModules.cpp
```diff
@@ -0,0 +1,149 @@
...
+    // call sites.
+    //
+    OpBuilder builder(&getContext());
+    llvm::SmallVector<func::FuncOp> declarations;
+    for (auto funcOp : deviceModule.getOps<func::FuncOp>()) {
+      if (!funcOp->hasAttr(ttir::CPUHoistedFuncAttr::name) ||
+          !funcOp.isDeclaration()) {
+        continue;
+      }
+      constexpr const char *kDeclSuffix = "_decl";
```
mtopalovicTT: can you make this global variable, I think its used in cpu hoisting also

lib/Conversion/TTNNToEmitPy/TTNNToEmitPyPass.cpp
```diff
@@ -39,8 +40,65 @@ class TTNNToEmitPyTypeConverter : public TypeConverter {
...
 struct ConvertTTNNToEmitPyPass
     : public tt::ttnn::impl::ConvertTTNNToEmitPyBase<ConvertTTNNToEmitPyPass> {
+
+  using BaseConstructor =
+      tt::ttnn::impl::ConvertTTNNToEmitPyBase<ConvertTTNNToEmitPyPass>;
+
+  ConvertTTNNToEmitPyPass() = default;
+  ConvertTTNNToEmitPyPass(const ConvertTTNNToEmitPyOptions &options)
+      : BaseConstructor(options) {}
+
```
mtopalovicTT, line 97:
```suggestion
  using tt::ttnn::impl::ConvertTTNNToEmitPyBase<ConvertTTNNToEmitPyPass>::ConvertTTNNToEmitPyBase;
```

lib/Dialect/TTNN/Pipelines/TTNNPipelines.cpp
```diff
@@ -374,26 +368,52 @@ void createTTNNToEmitPyDevicePipeline(
...
+
+// Pipeline which lowers the CPU module from TTIR to EmitPy using
+// TTNN golden functions.
+//
+void createTTIRToEmitPyCPUPipeline(OpPassManager &pm) {
+  auto &cpuPm = pm.nest<ttcore::CPUModuleOp>().nest<mlir::ModuleOp>();
+
+  // Prepare CPU module TTIR ops for TTNN lowering.
+  //
+  cpuPm.addPass(mlir::tt::ttir::createTTIRFusing());
```
mtopalovicTT, line 426: Hmmm i'm not sure how ideal is this, but then I don't have any better approach on my mind tbh....

lib/Conversion/TTNNToEmitPy/TTNNToEmitPyPass.cpp
```diff
@@ -102,6 +161,13 @@ struct ConvertTTNNToEmitPyPass
         signalPassFailure();
         return;
       }
+
+      // Enable torch tensor conversions if the golden mode is enabled.
+      //
+      if (enableGoldenMode) {
+        module.walk(
+            [&](func::FuncOp funcOp) { enableTorchConversion(funcOp); });
```
mtopalovicTT, line 165: This can be done using [unrealized materialization cast](https://www.jeremykun.com/2023/10/23/mlir-dialect-conversion/) but you would have to tag ranked tensors with torch tensor attribute or something similar to avoid conversion in case tensor is already converted.

dmilinkovicTT: TBH, not sure how can we tag the tensors types in the EmitPy type system except by adding an attribute to the `OpaqueType`, which feels wrong.

From my understanding, the best course of action would be to use `emitpy.OpaqueType<"torch.Tensor">` for golden ops.

Even still, IMHO, the current approach would be a bit simpler to grasp, since using `addSourceMaterialization` / `addTargetMaterialization` doesn't fit the usecase of golden execution that well - we want to keep using TTNN tensors for function inputs and results, and Torch tensors for everything in-between, instead of performing global type conversion. This partial nature of type conversion would make things a bit awkward.

Or have you had something else in mind?

lib/Dialect/TTNN/Pipelines/TTNNPipelines.cpp
dmilinkovicTT: The fusing pass here, or the overall approach in this pipeline?

tools/ttrt/common/emitpy.py
```diff
@@ -345,6 +345,8 @@ def signal_handler(sig, frame):
                         and node.name != "main"
                         and node.name[0:18] != "create_inputs_for_"
                         and not node.name.__contains__("_const_eval_")
+                        # TODO(dmilinkovic): this is getting out of hand, issue #6386.
+                        and not node.name.__contains__("hoisted_")
```
jgrimTT, line 349: Ha this really is getting out of hand. Can you add the same line to `execute_py` in `tools/builder/base/builder_runtime.py`?

dmilinkovicTT: Sure, added

lib/Dialect/TTNN/Pipelines/TTNNPipelines.cpp
mtopalovicTT: No, the overall approach.  Because of how CPU hoisting was originally made where we lower ttir into lingalg and not ttnn into lingalg, you have to explicitly add each of the passes where are needed for TTNN lowering here, so there is likelihood that someone is going to break emitpy CPU pipeline but not adding some important pipeline which is needed for some function call in ttnn (like flatten sliding window or some decomposition). But then again I'm not sure what would be better solution in this moment. Ideally you would have TTNN dialect in CPU module and not TTIR and then we would either lower ttnn into lingalg, or in your case to golden function. This is first thing that came up to my mind, there are other ways for sure.

lib/Conversion/TTNNToEmitPy/TTNNToEmitPyPass.cpp
mtopalovicTT: Materialization does exactly that. It adds materialization hook at the start and at the end of the function body, but you would have to use some different type (like opaquetype<"torch.Tensor">) for conversion to work properly. So you would run ttnntoemitpy pipeline on CPU module with type converter which would convert ranked tensor type to opaque torch type like you said. This would add materialization hooks which you can use to substitute for:
```
    auto toTorchOp = builder.create<emitpy::CallOpaqueOp>(
        funcOp.getLoc(), arg.getType(), "ttnn.to_torch", ValueRange{arg},
        nullptr, nullptr);
```

and ttnn.from_torch ops for function inputs and outputs. But regardless I wouldn't block this PR on that, it's just more MLIR way of doing type conversion.

lib/Dialect/TTNN/Pipelines/TTNNPipelines.cpp
dmilinkovicTT: Yeah, I agree that TTNN -> LinAlg would make much more sense, we could perhaps discuss this offline with the team and plan out the work, since I think it would be pretty risky to just convert all the existing conversion patterns to TTNN right away.

> so there is likelihood that someone is going to break emitpy CPU pipeline but not adding some important pipeline which is needed for some function call in ttnn (like flatten sliding window or some decomposition)

If an important pass gets skipped mistakenly, TTIR -> TTNN conversion would fail, and we wouldn't end up with a silent failure. However, this will happen only for the ops that are currently covered with EmitPy CPU-hoisting tests. 

If a new op gets added, and the new op requires some TTIR pass before lowering to TTNN, it could indeed fail. But I think this is a general testing blind spot across all targets when it comes to CPU-hoisting. I had some ideas on how to make the CPU-hoisting test coverage more robust, we could also discuss this offline with the team.

mtopalovicTT (approved):

PR #6365, opened 2025-12-17
Bug fixes and cleanups

rpavlovicTT:
TTIROps: Fold identity permutes where all swapped dims are size 1
- Permute [2,0,1,3] on 1x1x1x64 is now folded as no-op
- Eliminates trivial permutes from RoPE broadcast patterns

TTNNOps: Fix SplitQueryKeyValueAndSplitHeadsOp memory config
- Use q_out memory config as fallback

Runtime slice: Fix to pass memory config from TensorRef
- Both static and dynamic slice variants now respect output layout

### Checklist

### DIFF
--- a/lib/Dialect/TTIR/IR/TTIROps.cpp
+++ b/lib/Dialect/TTIR/IR/TTIROps.cpp
@@ -4366,11 +4366,28 @@ void mlir::tt::ttir::ReverseOp::getCanonicalizationPatterns(
 
 // PermuteOp with identity permutation is a no-op.
 // The input can be used directly as the output.
+// This includes:
+// 1. Sorted permutations like [0, 1, 2, 3]
+// 2. Permutations where all swapped dimensions have size 1
+//    (e.g., [2, 0, 1, 3] on shape 1x1x1x64 is a no-op)
 static mlir::OpFoldResult foldIdentityPermute(mlir::tt::ttir::PermuteOp op) {
+  // Case 1: True identity permutation (sorted)
   if (llvm::is_sorted(op.getPermutation())) {
     return op.getInput();
   }
-  return nullptr;
+
+  // Case 2: All non-identity dimension swaps are between dims of size 1
+  auto inputShape = op.getInput().getType().getShape();
+  auto perm = op.getPermutation();
+  for (size_t i = 0; i < perm.size(); ++i) {
+    if (perm[i] != static_cast<int64_t>(i)) {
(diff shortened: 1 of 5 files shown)


runtime/lib/ttnn/operations/transformer/split_query_key_value_and_split_heads.cpp
```diff
@@ -14,7 +14,11 @@ static void runSplitQueryKeyValueAndSplitHeadsOp(
     const ::tt::target::ttnn::SplitQueryKeyValueAndSplitHeadsOp *op,
     ProgramTensorPool &tensorPool) {
   std::optional<::ttnn::MemoryConfig> outputMemoryConfig =
-      ::tt::runtime::ttnn::utils::createMemoryConfigIfNeeded(op->memcfg());
+      op->memcfg()
+          ? ::tt::runtime::ttnn::utils::createMemoryConfigIfNeeded(op->memcfg())
+          : ::tt::runtime::ttnn::utils::createMemoryConfigIfNeeded(
+                ::tt::runtime::ttnn::utils::getTensorRefMemoryConfig(
+                    op->q_out()));
```
odjuricicTT, line 21: Why is this needed? Can this not be solved on IR level?

rpavlovicTT: It's either we do it here, or in TTNNToFlatbuffer.cpp. Currently, it's a mixed situation, some ops do it in runtime, some in translation.

lib/Dialect/TTIR/IR/TTIROps.cpp
```diff
@@ -4366,11 +4366,28 @@ ::mlir::LogicalResult mlir::tt::ttir::PermuteOp::verify() {
...
   }
-  return nullptr;
+
+  // Case 2: All non-identity dimension swaps are between dims of size 1
+  auto inputShape = op.getInput().getType().getShape();
+  auto perm = op.getPermutation();
+  for (size_t i = 0; i < perm.size(); ++i) {
+    if (perm[i] != static_cast<int64_t>(i)) {
+      // This dim position is changed - check if both dims involved are size 1
+      if (inputShape[i] != 1 || inputShape[perm[i]] != 1) {
```
azecevic000, line 4385: Nit: you can use
```cpp
for (auto [index, permuteIndex] : llvm::enumerate(perm)) {
  ...
}
```

Issue #6310, opened 2025-12-12
[D2M] decompose d2m tile_mask_boundary op for whole tiles


vwellsTT:
### Is your feature request related to a problem? Please describe

After introducing d2m.tile_mask_boundary as an op, we need a pass to decompose it into lower-level ops which implement the arithmetic.

### Describe the solution you'd like

As a first step, I want to introduce a small pass which decomposes tile_mask_boundary ops for full tiles, and asserts on partial tiles--this handling will be considerably more complex, and can be handled in a separate ticket

Issue #6309, opened 2025-12-12
[D2M] Add new high-level d2m op to describe filling tiled regions


vwellsTT:
### Is your feature request related to a problem? Please describe

It's convenient to have a high-level op which describes the filling of a tensor's padding regions with specific values. 

### Describe the solution you'd like

We'll insert the op during LowerToLayout.cpp--any layout change which involves change in fill value will get lowered to some type of d2m op.  Later on, a separate pass will decompose this op into relevant low-level d2m constructs (tile-filling ops, split tile handling, etc).

Issue #6308, opened 2025-12-12
[D2M] Add support for filling padded regions with specific oob_val


vwellsTT:
In D2M, we need support for actually propagating oob_vals into padded regions of tensors (currently, we only support `undef`, i.e. we are not obliged to fill with any particular values.

PR #6368, opened 2025-12-17
[D2M] Tiling Masking -- Complete Tiles

vwellsTT:
### Ticket
https://github.com/tenstorrent/tt-mlir/issues/6308
Closes [#6309](https://github.com/tenstorrent/tt-mlir/issues/6309) and https://github.com/tenstorrent/tt-mlir/issues/6310
### Problem description
We need some way to fill padded regions with specific values for various reduction ops etc.

### What's changed
This PR introduces `d2m.tile_mask_boundary` op to represent filling padded regions, a DecomposeMasking pass to decompose this into primitive ops *for discrete tile case only*.  Partial tile filling will be introduced in a follow-on.  Similarly, a follow-on will introduce logic to actually set oob_val per input op type (this PR does NOT hook up the pass e2e or anything).  `llvm-lit` and builder tests to confirm logic works.
Also introduce a patch to the SFPUTileLoopFission pass s.t. it will not try to incorrectly fission my kernels--may need some further thought on this.
Add support for custom dim_alignments to builder tests, s.t. I can pad a 32x32 into a 64x64 and test fill values.
Also, disallow view folding if reinterpretLayout is set, and fix logical bug in GridSelection.

### Checklist


include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
```diff
@@ -711,6 +711,47 @@ def D2M_TileTypecastOp : D2M_GenericRegionComputeUnaryDstOp<"tile_typecast", [D2
...
+        decomposition time, this op uses d2m.iter_index to determine the current tile's
+        position in the grid and computes which elements are within bounds.
+
+        This is a high-level op that gets decomposed in a later pass into primitive
+        tile arithmetic operations (comparisons, selects via mul/add).
+
+        Example:
+        ```mlir
+        %masked = d2m.tile_mask_boundary %input, [50, 50], <zero>
+            : !ttcore.tile<32x32, f32> -> !ttcore.tile<32x32, f32>
```
nsmithtt: It's a little confusing that this op takes the full logical shape as input, but it operates on tiles.  I feel like this op should take a 2d shape that marks the region within a single tile that you'd like to keep (implicit origin 0,0).  The logic for masking full tiles seems like it should just be captured by control flow:

```mlir
%y = iter_index(0)
%x = iter_index(1)
scf.if %y >= 50 && %x >= 50 {
  %masked = d2m.tile_mask_boundary %input, [0, 0], <zero>
}
```

OR we could have a shard version of this op, which is a bit closer to what you have here, but accepts a memref instead of a tile.

```mlir
d2m.block_mask_boundary %input, [50, 50], <zero> : memref<....>
```

include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
```diff
@@ -711,6 +711,47 @@ def D2M_TileTypecastOp : D2M_GenericRegionComputeUnaryDstOp<"tile_typecast", [D2
...
+        ```
+    }];
+
+    let arguments = (ins
+        TTCore_Tile:$input,
+        DenseI64ArrayAttr:$logical_shape,
```
nsmithtt: I think this might need to be a Value instead of an attribute, because the masking pattern will be different depending on which core you are in the multicore case.  So each core will need to calculate what range they need to mask.

nsmithtt:
I think the majority of the code is on the right track! Just a few comments inline surrounding the interface of the masking op.

include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
vwellsTT: hmm, yeah that makes sense, I'll rethink

lib/Dialect/D2M/Transforms/LowerToLayout.cpp
```diff
@@ -572,6 +604,43 @@ class D2MLowerToLayoutRewriter : public OpRewritePattern<ToLayoutOp> {
...
+    auto genericOp = rewriter.create<GenericOp>(
+        loc, input, output,
+        [&](OpBuilder &builder, Location innerLoc, ValueRange blockArgs) {
+          Value src =
+              builder.create<WaitOp>(innerLoc, blockArgs[0]).getResult();
+          Value dst =
+              builder.create<ReserveOp>(innerLoc, blockArgs[1]).getResult();
+
+          // Create index constants for logical bounds.
+          // TODO: For multicore, compute per-core bounds using core_index.
```
missing username/bug in TODO
```suggestion
          // TODO(nsmith): For multicore, compute per-core bounds using core_index.
```

vmilosevic (commented):

include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
```diff
@@ -711,6 +711,62 @@ def D2M_TileTypecastOp : D2M_GenericRegionComputeUnaryDstOp<"tile_typecast", [D2
...
+        d2m.block_mask %input, %output, %logical_rows, %logical_cols, <zero>
+            : (memref<2x3x!ttcore.tile<32x32, f32>>, memref<2x3x!ttcore.tile<32x32, f32>>)
+        ```
+    }];
+
+    let arguments = (ins
+        AnyRankedTensorOrMemRef:$input,
+        AnyRankedTensorOrMemRef:$output,
+        Index:$logical_rows,
+        Index:$logical_cols,
```
nsmithtt, line 755: Thinking about this some more, I think this might need to be a shape (variadic inputs) w/ maybe even additional attributes in the context of collapse.  If you had logical shape `[2, 3, 4]` and collapse `[[0, -1], [-1, -1]]` and dim alignments `1x32x32` then we'd need to mask out rows between collapsed dims.

I think it's fine to leave this as out of scope of this initial PR, but perhaps we want a more general interface that leaves the door open for that.

lib/Dialect/D2M/Transforms/DecomposeMasking.cpp
```diff
@@ -0,0 +1,242 @@
...
+    Type elementType = inputType.getElementType();
+    auto tileType = dyn_cast<ttcore::TileType>(elementType);
+    if (!tileType) {
+      return rewriter.notifyMatchFailure(op,
+                                         "element type must be a tile type");
+    }
+
+    Type scalarType = getScalarElementType(elementType);
+    ArrayRef<int64_t> blockShape = inputType.getShape();
+    size_t blockRank = blockShape.size();
```
nsmithtt, line 108: Should assert that blockRank >= 2

lib/Dialect/D2M/Transforms/DecomposeMasking.cpp
```diff
@@ -0,0 +1,242 @@
...
+
+    // Get shard shape (tiles per core). The block shape is what each core
+    // sees, so for a 4x4 grid with 1 tile per core, blockShape is [1, 1].
+    int64_t shardRows = blockShape[0];
+    int64_t shardCols = blockShape[1];
```
nsmithtt: blockShape[-2] & blockShape[-1]  (I suppose this does assume that dim alignments was tile aligned on the innermost dims)

include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
vwellsTT: Ah, hmm.  Yeah that's pretty annoying to handle, will think more and do it in a follow-up I guess, good point

PR #6373, opened 2025-12-17
Allow return values in CPU-hoisted functions

dmilinkovicTT:
### Ticket
N/A

### Problem description
Runtime isn't able to handle the values returned from the CPU-hoisted functions. Because of this, DPS semantics gets introduced explicitly in the CPU-hoisted functions through `ConvertCPUHoistedFunctionsToDPS` pass, which replaces return values with output arguments, and makes the functions void. 

This introduces some complexity to the TTNN pipelines. Having CPU-hoisted function actually return values would simplify the CPU-hoisting setup across the targets.

### What's changed
- Removed `ConvertCPUHoistedFunctionsToDPS` pass
- Fixed the bug in ttmetal target introduced in #6041 , where `CpuCommand` wasn't able to support multiple return values
- Introduced support for CPU-hoisted return values in `ttnn` and `ttmetal` runtimes
  - Unpacking the tensors returned as memref descriptors

### Checklist

### DIFF
--- a/runtime/include/tt/runtime/detail/common/dylib.h
+++ b/runtime/include/tt/runtime/detail/common/dylib.h
@@ -107,16 +107,17 @@ prepareSizesAndStrides(const std::vector<int64_t> &sizes,
                  [](uint32_t s) -> int64_t { return s; });
 }
 
-// Callback type for creating tensors from WrappedTensor.
+// Callback type for creating runtime tensors from the provided pre-allocated
+// data buffer.
 //
-// The callback receives the tensor reference, the WrappedTensor, and a
-// deletion callback function that should be called to free the buffer when
-// the created tensor is destroyed.
+// The callback receives the tensor reference, and a shared_ptr to the
+// pre-allocated data buffer. The callback is expected to return a
+// runtime-specific tensor.
 //
 template <typename TensorType, typename TensorRefType>
 using CreateTensorCallbackType =
-    std::function<TensorType(const TensorRefType *, const WrappedTensor &,
-                             std::function<void()> deletionCallback)>;
+    std::function<TensorType(const TensorRefType *, std::shared_ptr<void>)>;
+
 // Unpack tensors returned from a CPU-hoisted function.
 //
 // Takes the returned WrappedTensor array and a target-specific callback to
@@ -134,12 +135,16 @@ std::vector<TensorType> inline unpackTensors(
   for (size_t i = 0; i < numOutputs; ++i) {
     const WrappedTensor &wrappedTensor = outputArray[i];
 
-    // Deletion callback function to free the buffer allocated inside the
-    // CPU-hoisted function.
-    auto deletionCallback = [data = wrappedTensor.start]() { std::free(data); };
+    // Set up a shared_ptr with custom deleter to free the tensor buffer
+    // allocated by the CPU-hoisted function once the runtime tensor is
+    // destroyed.
+    // Note: we use the alignedStart pointer for the tensor data, but
+    // we need to free the original start pointer.
+    std::shared_ptr<void> dataPtr(
+        wrappedTensor.alignedStart,
+        [data = wrappedTensor.start](void *) { std::free(data); });
 
-    results.push_back(
-        createTensorCallback(outs->Get(i), wrappedTensor, deletionCallback));
+    results.push_back(createTensorCallback(outs->Get(i), dataPtr));
   }
 
   // Free the sizesAndStrides buffer allocated inside the CPU-hoisted function
--- a/runtime/lib/ttnn/operations/cpu/cpu.cpp
+++ b/runtime/lib/ttnn/operations/cpu/cpu.cpp
@@ -38,19 +38,11 @@ void run(const ::tt::target::ttnn::CpuOp *op, ProgramContext &context) {
   // Callback for tensor creation from WrappedTensor.
   common::CreateTensorCallbackType<::ttnn::Tensor,
                                    ::tt::target::ttnn::TensorRef>
-      createTensor =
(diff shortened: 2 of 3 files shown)


runtime/include/tt/runtime/detail/common/dylib.h
```diff
@@ -107,6 +107,44 @@ prepareSizesAndStrides(const std::vector<int64_t> &sizes,
...
+// Takes the returned WrappedTensor array and a callback to create tensors.
+// When freeData is true (default), frees all memory allocated by the CPU
+// function. When false, the callback takes ownership of wrapped.start.
+template <typename TensorType, typename TensorRefType>
+std::vector<TensorType> inline unpackTensors(
+    WrappedTensor *outputArray, size_t numOutputs,
+    const flatbuffers::Vector<flatbuffers::Offset<TensorRefType>> *outs,
+    std::function<TensorType(const TensorRefType *, const WrappedTensor &)>
+        createTensor,
+    bool freeData = true) {
```
vwellsTT: subjective, but personally I hate default param in C++, would rather you just update call sites to use false

lib/Dialect/LLVM/Transforms/EmitWrapperFuncs.cpp
```diff
@@ -129,11 +151,110 @@ void generateLLVMWrappersForArgRanks(ModuleOp moduleOp) {
...
+      auto loc = func.getLoc();
+      auto i64Ty = builder.getI64Type();
+
+      auto makeConst = [&](int64_t val) {
+        return builder.create<LLVM::ConstantOp>(loc, i64Ty,
+                                                builder.getI64IntegerAttr(val));
+      };
+
+      // Allocate output array and sizesAndStrides buffer.
+      constexpr int64_t wrappedTensorBytes = 32; // 4 fields * 8 bytes
```
vwellsTT: nit: I think we have loose convention of naming magic numbers like this `kWrappedTensorBytes`

runtime/include/tt/runtime/detail/common/dylib.h
```diff
@@ -107,6 +107,44 @@ prepareSizesAndStrides(const std::vector<int64_t> &sizes,
...
+
+  std::vector<TensorType> results;
+  results.reserve(numOutputs);
+
+  for (size_t i = 0; i < numOutputs; ++i) {
+    const WrappedTensor &wrapped = outputArray[i];
+    results.push_back(createTensor(outs->Get(i), wrapped));
+
+    if (freeData) {
```
vwellsTT: will this not invalidate the results created above? Not sure I get this in general

vwellsTT (approved):
Again this broadly looks great! Some questions inline

runtime/include/tt/runtime/detail/common/dylib.h
dmilinkovicTT: I've refactored the implementation a bit, hope it's cleaner now.

I've managed to make both TTNN and TTMetal tensors borrow the allocated storage, so this deallocation doesn't need to happen anymore

runtime/lib/ttnn/operations/cpu/cpu.cpp
```diff
@@ -35,20 +32,52 @@ void run(const ::tt::target::ttnn::CpuOp *op, ProgramContext &context) {
...
+         const common::WrappedTensor &wrapped) -> ::ttnn::Tensor {
+    ::ttnn::Shape shape = utils::toTTNNShape(*ref->desc()->shape());
+    ::ttnn::DataType dtype = ::tt::runtime::ttnn::utils::toTTNNDataType(
+        ref->desc()->layout()->memory_desc()->data_type());
 
-    context.getTensorPool().insertTTNNTensorAndValidate(fbOutput, outTensor);
-  }
+    // Tensor buffer is allocated by the CPU-hoisted function. It needs to be
+    // deallocated when the TTNN tensor is destroyed.
+    std::function<void()> deletionCallback = [allocatedData = wrapped.start]() {
```
vwellsTT: not a blocker or anything, but I wonder if there's a good common place this can be defined vs defining same free func multiple places

runtime/include/tt/runtime/detail/common/dylib.h
vwellsTT: nice, in some ways I like this better, though tbh in others I think it's a bit ugly to have the std::frees in multiple places :) but I think it's fine, can always clean things up later.
I guess maybe it'd make sense to define it in a shared header like here, not sure

runtime/lib/ttnn/operations/cpu/cpu.cpp
dmilinkovicTT: I've managed to extract the deallocation to the `unpackTensors` function

runtime/include/tt/runtime/detail/ttnn/utils.h
```diff
@@ -158,6 +158,42 @@ inline ::ttnn::Tensor createTTNNTensor(
...
+// remains valid for the lifetime of the tensor.
+//
+template <typename T>
+inline ::ttnn::Tensor createBorrowedTTNNTensor(void *rawData,
+                                               const ::ttnn::Shape &shape) {
+  auto dataPtr = std::shared_ptr<void>(rawData, [](void *) {
+    // No deletion needed, as the caller manages the buffer's lifetime.
+  });
+
+  return createBorrowedTTNNTensor<T>(dataPtr, shape);
```
vmilosevic: :warning: **clang\-diagnostic\-error** :warning:
call to function `` createBorrowedTTNNTensor `` that is neither visible in the template definition nor found by argument\-dependent lookup

vmilosevic (commented):

pilkicTT (approved):

Issue #6377, opened 2025-12-18
[D2M] Handling NOC read/write alignment boundaries in TMs


dlokeTT:
### Describe the bug

For reshape, I've encountered test cases where we currently can't do the reshape because the noc has to read and write on 16 byte boundaries which means we can't currently do reshapes on the singular element level. This should be a general problem for other tms such as permute and the shapes that cause these issues appear in prominent models such as llama.

Issue #6376, opened 2025-12-18
[D2M] Add support for 1D tensors


dlokeTT:
### Describe the bug

Llama 3.2 3B has 1D tensors used in reshapes and other operations which currently hit an assert in `ttirtod2m` complaining that the metal layout attribute has to be 2 dimensions or greater. Can either hack this by adding a leading dim of 1 to turn a 1D tensor into 2D or remove this assertion and handle 1D tensors thru the pipeline properly.

dlokeTT:
Closed as duplicate as #6659

Issue #6375, opened 2025-12-18
[D2M] calculateCoalescingFactors inefficient


dlokeTT:
### Describe the bug

In `D2MGenericLowerDMAs` the `calculateCoalescingFactor` function which looks to find the max coalescing factor we can do for dma reads samples every grid and shard to determine the coalescing factor. For TMs such as reshape this causes compile time to bloat exponentially. For a reshape found in llama 3.2 3B that goes from tensor<32x24x128x128xbf16> -> tensor<768x128x128xbf16> the `D2MGenericLowerDMAs` pass takes 7 minutes alone. We need to find a smarter way to find this factor that scales much better as we get larger TMs and ops that will be found in models

PR #6378, opened 2025-12-18
added d2m support for reshape and tests

dlokeTT:
### Problem description
Need to support reshape in d2m 

### What's changed
Added support for changing the rank of tensors in `projectLogicalMapToUnitDeviceSpace` which is needed for reshape and `reshapeLogicalMap` which is the affine map for reshape transformations. Assumes minimum rank is 2 for now.
Added builder tests for tile aligned tensors, random shaped tensors and tensors found in llama 3.2 3b. Lots of these are skipped for various reasons, such as compile time #6375, 1d support #6376, and noc unaligned reads/writes #6377.

Had an initial hacky solution for 1d tensors by adding a leading 1 to the metal layout attribute so it would pass the assertion but didn't get to handling all the edge cases so leaving that aside and marking the 1d llama tests as skip for now. I could pick this up when I'm back or we could try and support this natively in the pipeline.

### Checklist

### DIFF
--- a/lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
+++ b/lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
@@ -1329,16 +1329,24 @@ public:
     SmallVector<AffineExpr> deviceExprs;
 
     // Grid coordinate mapping (first inputLogicalRank results).
-    // Pass through last 2 grid dims for 2D grid compatibility, pad higher dims.
     for (unsigned i = 0; i < inputLogicalRank; ++i) {
-      unsigned revIdx = inputLogicalRank - 1 - i;
-      if (revIdx < 2 && revIdx < outputLogicalRank) {
-        deviceExprs.push_back(
-            builder.getAffineDimExpr(outputLogicalRank - 1 - revIdx));
-      } else if (revIdx < outputLogicalRank) {
-        deviceExprs.push_back(builder.getAffineDimExpr(revIdx));
+      if (inputLogicalRank == outputLogicalRank) {
+        // Same rank: identity mapping for grid (matches original behavior)
+        deviceExprs.push_back(builder.getAffineDimExpr(i));
+      } else if (inputLogicalRank < outputLogicalRank) {
+        // Expanding (e.g., 2D -> 3D): map input grid dims to output's last
+        // inputLogicalRank grid dims.
+        unsigned outputGridIdx = outputLogicalRank - inputLogicalRank + i;
+        deviceExprs.push_back(builder.getAffineDimExpr(outputGridIdx));
       } else {
-        deviceExprs.push_back(builder.getAffineConstantExpr(0));
+        // Contracting (e.g., 3D -> 2D): map last outputLogicalRank input grid
+        // dims to output grid, pad the rest with 0.
+        if (i < inputLogicalRank - outputLogicalRank) {
+          deviceExprs.push_back(builder.getAffineConstantExpr(0));
+        } else {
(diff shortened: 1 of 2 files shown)


lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
```diff
@@ -1279,77 +1435,19 @@ class D2MTensorManipulationOpRewriter
...
-    auto newOutTy = RankedTensorType::get(outTy.getShape(),
-                                          outTy.getElementType(), newLayout);
-
-    auto storage =
-        rewriter.create<d2m::EmptyOp>(op.getLoc(), outputs[0].getType());
-    auto view = rewriter.create<d2m::StreamLayoutOp>(
-        op.getLoc(), newOutTy, inputs[0], storage.getResult());
-
-    rewriter.replaceOp(op, unLayoutResult(rewriter, view->getResult(0),
-                                          op->getResult(0).getType()));
```
nsmithtt, line 1309: What was the purpose of this refactor? It's hard to tell if there were any changes made?

lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
```diff
@@ -1279,77 +1435,19 @@ class D2MTensorManipulationOpRewriter
...
-  }
 };
 } // namespace
 
-static AffineMap rearrangeLogicalMap(ttir::RearrangeOp op) {
-  mlir::FailureOr<AffineMap> maybeMap = op.getInvPatternMap();
-  assert(succeeded(maybeMap));
-  return *maybeMap;
-}
-
```
nsmithtt, line 1352: Why move the code?

lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
dlokeTT: There were no code changes. Was playing around with this and forgot to clean it up when I rebased. Reverted back to how its done in main

lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
dlokeTT: Also an artifact of some stuff I was playing with and forgot to undo. Put it back to where it was.

lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
```diff
@@ -1313,33 +1313,41 @@ class D2MTensorManipulationOpRewriter
...
+    for (unsigned i = 0; i < inputLogicalRank; ++i) {
+      unsigned revIdx = inputLogicalRank - 1 - i;
+      if (revIdx < 2 && revIdx < outputLogicalRank) {
+        deviceExprs.push_back(
+            builder.getAffineDimExpr(outputLogicalRank - 1 - revIdx));
+      } else if (revIdx < outputLogicalRank) {
+        deviceExprs.push_back(builder.getAffineDimExpr(revIdx));
+      } else {
+        deviceExprs.push_back(builder.getAffineConstantExpr(0));
+      }
```
nsmithtt, line 1350: Is this logically doing what it was doing before?

dlokeTT: Oh wait good catch. Ran into issues with this where the grid dim indices would collide for the cases where the ranks of the input and output are the same or the input grid rank is > than the output grid rank. Added handling for each case now where input rank is <, =, > the output rank. This should now logically match what was happening before.

PR #6392, opened 2025-12-18
[ttnn.jit] Rework of TTNN JIT E2E Tests

brapananTT:
### Problem description
Want to rework TTNN JIT e2e tests to reduce CI time.

### What's changed
Moved` test_eltwise.py`, `test_eltwise_composite.py`, `test_reduction.py` into nightly
Added `*_smoketest.py` tests for the above tests to replace in on-push, which include subsets of the nightly tests
No extra xfails are added from this PR

### Checklist

### DIFF
--- a/test/ttnn-jit/test_eltwise_smoketest.py
+++ b/test/ttnn-jit/test_eltwise_smoketest.py
@@ -272,14 +272,12 @@ def test_binary_ops_dram(device, shape, dtype, ttnn_dtype, op, graph_capture):
 def test_binary_ops_l1(
     device, shape, max_grid, shard_strategy, dtype, ttnn_dtype, op, graph_capture
 ):
-    if op in [pow, eq, ne, gt, ge, lt, le] and dtype == torch.float32:
+    if op in [div, pow, eq, ne, gt, ge, lt, le] and dtype == torch.float32:
         pytest.xfail("failing allclose for some shapes")
     if op in [maximum, minimum, pow] and ttnn_dtype == ttnn.DataType.BFLOAT8_B:
         pytest.xfail("failing allclose for some shapes for bfp8")
     if op == div and ttnn_dtype == ttnn.DataType.BFLOAT8_B:
-        pytest.xfail(
-            "RuntimeError: Input and output tile size should be same, from typecast_sharded_program_factory.cpp:44: input_tile_size == output_tile_size"
-        )
+        pytest.skip("ttnn.div does not support BFLOAT8_B")
 
     run_op_test(
         device,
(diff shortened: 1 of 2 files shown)


test/ttnn-jit/nightly/test_eltwise.py
```diff
@@ -291,8 +291,8 @@ def test_unary_op_dram(device, shape, dtype, ttnn_dtype, op, graph_capture):
     ],
 )
 @pytest.mark.parametrize("graph_capture", [True, False])
-def test_unary_op_l1(
-    device, shape, max_grid, shard_strategy, dtype, ttnn_dtype, op, graph_capture
+def test_unary_op_l1_block(
```
arminaleTT: Please merge all three sharding modes into one function, even if you have to add new parameters

test/ttnn-jit/nightly/test_eltwise.py
```diff
@@ -61,6 +61,7 @@
     ((2, 1, 32, 2048), (7, 7)),
 ]
```
arminaleTT, line 61: Remove:
- height sharded
  -  ((128, 32), (3, 0))
- block sharded
  - ((32, 64), (0, 0))

test/ttnn-jit/test_eltwise_smoketest.py
```diff
@@ -0,0 +1,700 @@
...
+    [add, sub, mul, div, pow, eq, ne, gt, ge, lt, le, maximum, minimum],
+)
+@pytest.mark.parametrize("graph_capture", [True, False])
+def test_binary_ops_l1(
+    device, shape, max_grid, shard_strategy, dtype, ttnn_dtype, op, graph_capture
+):
+    if op in [pow, eq, ne, gt, ge, lt, le] and dtype == torch.float32:
+        pytest.xfail("failing allclose for some shapes")
+    if op == div and ttnn_dtype == ttnn.DataType.BFLOAT8_B:
+        pytest.xfail("mysterious error msg")
```
arminaleTT: What is this mysterious error message? Please put it in a case and share it so it can be diagnosed/ruled out. 

Since this is div, are setting minimum bounds on the random values in the tensor so we don't get a division by (near) zero?

brapananTT: Error is `RuntimeError: Input and output tile size should be same` coming from the golden tensor calculation. Seems as if ttnn.div does not support bfp8 according to ttnn [docs](https://docs.tenstorrent.com/tt-metal/latest/ttnn/ttnn/api/ttnn.div.html)? added a skip for this case for now. 

Also added a minimum bound on div input tensors in `utils.py`

arminaleTT (approved):

vtangTT (approved):
lgtm ty!

sgholamiTT (approved):

Issue #6335, opened 2025-12-15
Enable SDPA fusing for LLM with new torch and transformers version


odjuricicTT:
## Problem

After upgrading torch and transformers versions in tt-xla (commit https://github.com/tenstorrent/tt-xla/commit/81e993b2bb855535af76a570701bb70b472a49db), the SDPA (Scaled Dot Product Attention) fusing pattern no longer matches for LLM models like Llama.

Here is what the new llama IR looks like:

[llama_3_2_1b_1layer_new.mlir.txt](https://github.com/user-attachments/files/24171921/llama_3_2_1b_1layer_new.mlir.txt)

All of the below description is generated by claude so some errors are expected:

## Root Cause

The new torch/transformers version generates a different attention pattern that the existing SDPA fusing code doesn't recognize:

### Old pattern (torch 2.5):
```
Q @ K^T -> scale -> mask_add -> softmax -> @ V
```

### New pattern (torch 2.9):
```
Q * scale -> LinearOp(Q_scaled, K^T * scale, mask) -> NaN_detection -> softmax -> where(nan_mask, zeros, softmax) -> @ V
```

## Key differences

| Aspect | Old Pattern | New Pattern |
|--------|-------------|-------------|
| Q@K computation | MatmulOp | LinearOp (fused matmul + bias add) |
| Scale application | Post-matmul multiply | Pre-matmul multiply on Q and K |
| Softmax output | Direct | Wrapped in WhereOp for NaN-safe attention |
| K/V heads | GQA (8 heads passed to SDPA) | Expanded (32 heads before SDPA) |
| Data type | bf16 | f32 for attention computation |

### Old Llama attention pattern (works):
```mlir
// Q @ K^T matmul
%103 = "ttir.dot_general"(%98, %102) <{batch_dims_lhs = array<i64: 0, 1>, batch_dims_rhs = array<i64: 0, 1>, contract_dims_lhs = array<i64: 3>, contract_dims_rhs = array<i64: 2>}> 
    : (tensor<32x32x1x64xbf16>, tensor<32x32x64x128xbf16>) -> tensor<32x32x1x128xbf16>

// Scale multiply (post-matmul)
%104 = "ttir.multiply"(%103, %cst) : (tensor<32x32x1x128xbf16>, tensor<1x1x1x1xbf16>) -> tensor<32x32x1x128xbf16>

// Mask add
%105 = "ttir.add"(%104, %mask) : ...

// Softmax (direct)
[...]

PR #6399, opened 2025-12-19
[TTNN] Improve SDPA Fusing

jserbedzijaTT:
### Ticket
[Link to Github Issue](https://github.com/tenstorrent/tt-mlir/issues/6335)

### Problem description
The SDPA (Scaled Dot-Product Attention) fusing pattern was not handling newer PyTorch decomposition patterns. Specifically:
- Split-scale pattern: `PyTorch 2.9+` applies` sqrt(1/head_dim)` to both `Q` and `K` separately before the matmul, rather than applying `1/sqrt(head_dim)` to the `QK^T` result.
- NaN handling: Newer torch versions emit a where operation after softmax to handle fully-masked rows (replacing NaN with zeros).

The existing TTIR fusing pattern couldn't match these variations, causing SDPA to remain unfused in models using newer torch versions.

### What's changed

- Moved SDPA fusing from TTIRFusing to TTNNFusing pass
- Added support for split-scale pattern where scale is applied to Q and K separately
- Added support for robust softmax with all-masked row handling (where op pattern)
- Updated tests to cover both single-scale and split-scale SDPA patterns

### Checklist

### DIFF
--- a/test/python/golden/ttir_ops/fusing/test_sdpa_fusing.py
+++ b/test/python/golden/ttir_ops/fusing/test_sdpa_fusing.py
@@ -462,6 +462,7 @@ def test_sdpa_split_scale_robust_softmax(
         output_root=request.config.getoption("--path"),
         system_desc_path=request.config.getoption("--sys-desc"),
         device=device,
+        compile_options=["enable-optimizer=true"],
     )
 
     # Check for appropriate fused op based on q_seq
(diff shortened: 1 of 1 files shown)


lib/Dialect/TTNN/Transforms/TTNNFusing.cpp
```diff
@@ -451,6 +968,7 @@ class TTNNFusingPass : public impl::TTNNFusingBase<TTNNFusingPass> {
 #ifdef TTMLIR_ENABLE_OPMODEL
     if (enableOpConstraints) {
       patterns.add<RoPEFusing>(&getContext());
+      patterns.add<SDPAFusing>(&getContext());
```
vmilosevic, line 970: :warning: **clang\-diagnostic\-error** :warning:
unknown type name `` SDPAFusing ``; did you mean `` RoPEFusing ``?
```suggestion
      patterns.add<RoPEFusing>(&getContext());
```

Issue #5904, opened 2025-11-18
Fallback pass: limit no. of combinations


rpavlovicTT:
### Describe the bug

Ops that have more than 5 or so operands, can possibly introduce infinite compile time if fallback mechanism is triggered for them. Reason is brute-force solution that seeks for valid fallback. We can limit number of searched configurations and also prune some that don't make sense (for example, concat op shouldn't combine inputs with different data types etc.).

bmalesevicTT:
Currently they are limited by default to 10^4 combinations, and the limit is open to set as a flag in TT-MLIR. Heuristic is also slightly modified to prefer changing multiple operands' **datatype**, rather than convert their layout **row-major <-> tile**. 

Other potential heuristic improvements were considered but not implemented, as none could be shown to improve fallback success without sacrificing generality. These include:
- Parsing validation error messages to detect whether failures are caused by layout or datatype issues and prioritizing fixes accordingly.
- Differentiating ops by operand requirements, for example:
  - Ops like concat that require uniform properties across all inputs.
  - Ops with special constraints on specific operands (e.g., first input, weights).
  - Ops with multiple distinct operand requirements (e.g., SDPA).
- Adjusting the cost model for different types of changes.
At the moment, there are no known real-world examples where the current heuristic takes an unreasonably long time to converge. As a result, there is also no concrete data to guide cost adjustments.

Issue #5259, opened 2025-10-07
Bfp8 with optimizer_on hangs in endless loop


dgolubovicTT:
When I run ResNet with Bfp8 and optimizer enabled, It stucks in endless loop:

<img width="1015" height="632" alt="Image" src="https://github.com/user-attachments/assets/94e8cfcb-0bd5-47ed-a637-678f1fa13e4b" />

To repro:
- checkout `dgolubovic/enable-torch-lower-df` (or tt-xla main if [this](https://github.com/tenstorrent/tt-xla/pull/1489) is merged)
- run: `pytest -svv tests/torch/single_chip/models/resnet/test_resnet.py::test_torch_resnet_inference[optimizer_on-bfp8]`

When optimizer is disabled, it fails with same error but without endless loop: 

`BatchNorm only supports float32/bfloat16, got bfp8 (unsupported dtype)`

I thought it is worth while to open issue and tag @odjuricicTT.

dgolubovicTT:
Here is the ttir for easier repro:

[resnet_infinite_loop_optimizer.txt](https://github.com/user-attachments/files/22885022/resnet_infinite_loop_optimizer.txt)

dgolubovicTT:
Update after talking with @rpavlovicTT: It is not endless loop, it is just terribly slow for this case... Lots of combinations to check.

mtopalovicTT:
I think even if we constraint bfp8 to conv/matmul... it will still be slow because optimizer would check all ops with different data types including bfp8 (@rpavlovicTT can correct me if I'm wrong)

rpavlovicTT:
I don't think it will be too slow, we can control how many configs to check per op. However, in the case of ops having 5+ operands our fallback pass (automatic workaround) goes nuts because we try all possible workarounds per operand. This explodes quickly in this case.

bmalesevicTT:
Adjusted to newer TTIR format, without DPS: 

[resnet_infinite_loop_optimizer_nodps.mlir.zip](https://github.com/user-attachments/files/24367732/resnet_infinite_loop_optimizer_nodps.mlir.zip)

PR #6400, opened 2025-12-19
[Optimizer] Limited number of fallback configs, heuristic change, duplicate fix, clearer error

bmalesevicTT:
This PR improves the Optimizer fallback pass by reducing configuration explosion, bounding worst-case runtime, and improving error reporting.

### 1) Duplicate configuration fix and heuristic change (Issue #5259)

- Fixes a bug where, due to fallback generation order, unsupported BFP8 row-major layouts silently fell back to tiled making all data types use silent fallback tiled layouts, producing duplicate operand configurations. Datatype is now changed first to avoid duplicates, with additional duplicate pruning kept as a safeguard.
For the reported case (`batch_norm` 5 operands), with the previous heuristic, finding a valid configuration per this op dropped from ~4,500 attempts to ~1,200 attempts.
This is especially impactful as the affected model had 50+ occurrences of this op requiring fallback.
This made debug compilation time of the model go from ~2h to 15-20min.
- Additionally, row-major is presumably (after analyzing the ops and their implementation constraints) rarely needed to change for many operands, compared to data type change, so layout change is now deprioritized in the sorting heuristic. For the reported case (`batch_norm` 5 operands), with the previous heuristic, finding a valid configuration per this op dropped from ~1,200 attempts to ~700 attempts. This made debug compilation time of the model go from 15-20min to ~7min.

*Note this particular issue is not relevant for `optimization_level=1` as that does additional optimizations (`enable-fusing-conv2d-with-multiply-pattern`) which remove the operation which triggers fallbacks.
To reproduce the issue and see detailed logs of number of tried combinations use the nodps version of file attached in issue, and run: `TTMLIR_LOGGER_LEVEL=Trace ttmlir-opt --ttir-to-ttnn-backend-pipeline="system-desc-path=ttrt-artifacts/system_desc.ttsys enable-optimizer=true enable-bfp8-conversion=true" ./resnet_infinite_loop_optimizer_nodps.mlir -o ./runs/resnet_infinite_loop_optimizer_nodps_ttnn.mlir --mlir-print-local-scope -debug-only='op-validation' |& tee ./resnet_infinite_loop_optimizer_nodps.txt`

### 2) Limit fallback search via adjustable flag (Issue #5904)

Adds a configurable limit on the number of fallback configurations attempted to prevent unbounded exponential search.
In Release builds, fallback is now usually fast, but models with ops that have many operands may still require a much larger search space. In Debug builds, however, ~10,000 attempts already take ~1–2 minutes per op, meaning that if many ops require fallback, compilation time can grow significantly.
The limit provides a safety net while remaining adjustable for cases that genuinely need deeper exploration.

### 3) Clearer error message on failure

When no valid fallback configuration is found, the emitted error now includes additional information about the problematic operation including its full IR, improving debuggability, especially in Release builds, and all info directly visible from tt-xla benchmarks.

### DIFF
--- a/include/ttmlir/Dialect/TTNN/Validation/OpConstraintValidation.h
+++ b/include/ttmlir/Dialect/TTNN/Validation/OpConstraintValidation.h
@@ -10,6 +10,7 @@
 
 #include "mlir/IR/Operation.h"
 #include "llvm/ADT/ArrayRef.h"
+#include "llvm/ADT/StringRef.h"
 
 namespace mlir::tt::ttnn {
 
@@ -26,7 +27,7 @@ enum class ValidationStatus {
 };
 
 // Convert ValidationStatus to string for error messages
-const char *validationStatusToString(ValidationStatus status);
+llvm::StringRef validationStatusToString(ValidationStatus status);
 
 // Result of a single constraint validation test.
 struct ValidationResult {
--- a/lib/Dialect/TTNN/Transforms/OptimizerPasses/OperationValidationAndFallback.cpp
+++ b/lib/Dialect/TTNN/Transforms/OptimizerPasses/OperationValidationAndFallback.cpp
@@ -448,8 +448,8 @@ createFallbackTransforms(TTNNLayoutAttr originalLayout,
   std::vector<BufferType> targetBufferTypes = {BufferType::DRAM,
                                                BufferType::SystemMemory};
 
-  for (ttcore::DataType targetDataType : targetDataTypes) {
-    for (Layout targetLayout : targetLayouts) {
+  for (Layout targetLayout : targetLayouts) {
+    for (ttcore::DataType targetDataType : targetDataTypes) {
       for (BufferType targetBufferType : targetBufferTypes) {
         // Skip if this is the same as original
         if (targetLayout == originalLayout.getLayout() &&
--- a/lib/Dialect/TTNN/Validation/OpConstraintValidation.cpp
+++ b/lib/Dialect/TTNN/Validation/OpConstraintValidation.cpp
@@ -13,6 +13,7 @@
 
 #include "mlir/IR/BuiltinOps.h"
 #include "llvm/ADT/ArrayRef.h"
+#include "llvm/ADT/StringRef.h"
 #include "llvm/ADT/Twine.h"
 #include "llvm/Support/Error.h"
 #include "llvm/Support/ErrorHandling.h"
@@ -21,7 +22,7 @@ namespace mlir::tt::ttnn {
 
 namespace op_constraint_validation {
 
-const char *validationStatusToString(ValidationStatus status) {
+llvm::StringRef validationStatusToString(ValidationStatus status) {
   switch (status) {
   case ValidationStatus::Success:
     return "Success";


include/ttmlir/Dialect/TTNN/Transforms/Optimizer.h
```diff
@@ -36,6 +36,7 @@ struct TTNNOptimizerOptions {
   bool rowMajorEnabled = false;
   float tensorL1UsageCap =
       0.95f; // Default to 100% of maximum free space in L1.
+  uint32_t maxFallbackAttempts = 10000;
```
rpavlovicTT: Not needed.

include/ttmlir/Dialect/TTNN/Transforms/Passes.td
```diff
@@ -319,7 +319,11 @@ def TTNNOperationValidationAndFallback: Pass<"ttnn-operation-validation-and-fall
     Option<"tensorL1UsageCap",
            "tensor-l1-usage-cap",
            "float", /*default=*/"1.0",
-           "Limit L1 memory usage to this fraction of available space (0.0-1.0)">
+           "Limit L1 memory usage to this fraction of available space (0.0-1.0)">,
```
rpavlovicTT, line 322: Can you remove this option, it's not used anymore. We pass cap through IR attribute.

lib/Dialect/TTNN/Pipelines/TTNNPipelines.cpp
```diff
@@ -93,9 +93,11 @@ void createTTNNPipelineAnalysisPasses(
     DevicePassesWrapperOptions wrapperOptions;
     wrapperOptions.devicePtr = options.devicePtr;
     wrapperOptions.tensorL1UsageCap = options.tensorL1UsageCap;
+    wrapperOptions.maxFallbackAttempts = options.maxFallbackAttempts;
 
-    ttnn::TTNNOperationValidationAndFallbackOptions validationOptions{
-        options.tensorL1UsageCap};
+    ttnn::TTNNOperationValidationAndFallbackOptions validationOptions;
+    validationOptions.tensorL1UsageCap = options.tensorL1UsageCap;
```
rpavlovicTT, line 98: Remove it from here too.

lib/Dialect/TTNN/Pipelines/TTNNPipelines.cpp
```diff
@@ -93,9 +93,11 @@ void createTTNNPipelineAnalysisPasses(
     DevicePassesWrapperOptions wrapperOptions;
     wrapperOptions.devicePtr = options.devicePtr;
     wrapperOptions.tensorL1UsageCap = options.tensorL1UsageCap;
+    wrapperOptions.maxFallbackAttempts = options.maxFallbackAttempts;
```
rpavlovicTT: Wrapper doesn't need this option?

lib/Dialect/TTNN/Transforms/OptimizerPasses/OperationValidationAndFallback.cpp
```diff
@@ -388,8 +462,10 @@ createFallbackTransforms(TTNNLayoutAttr originalLayout,
   std::vector<BufferType> targetBufferTypes = {BufferType::DRAM,
                                                BufferType::SystemMemory};
 
-  for (Layout targetLayout : targetLayouts) {
-    for (ttcore::DataType targetDataType : targetDataTypes) {
+  // Variable optimized out in debug builds, but useful for logging
+  [[maybe_unused]] size_t attemptedInsertions = 0;
```
rpavlovicTT: Remove this.

lib/Dialect/TTNN/Transforms/OptimizerPasses/OperationValidationAndFallback.cpp
```diff
@@ -401,28 +477,38 @@ createFallbackTransforms(TTNNLayoutAttr originalLayout,
...
+        attemptedInsertions++;
+        fallbackLayoutsSet.insert(result);
       }
     }
   }
 
-  return fallbackLayouts;
+  TTMLIR_TRACE(
+      ttmlir::LogComponent::OpValidation,
+      "Generated {} unique fallback layouts from {} target combinations",
```
rpavlovicTT, line 491: If you want attempted insertions, just multiply loop sizes.

lib/Dialect/TTNN/Transforms/OptimizerPasses/OperationValidationAndFallback.cpp
```diff
@@ -401,28 +477,38 @@ createFallbackTransforms(TTNNLayoutAttr originalLayout,
...
 
-        // Apply layout transformation if needed
-        if (targetLayout != originalLayout.getLayout()) {
-          result = result.withLayout(targetLayout, tensorShape);
-        }
-
-        // Apply data type transformation if needed
+        // Apply data type transformation first (before layout changes)
+        // This prevents issues where layout changes force incompatible layouts
+        // (e.g., BFP8 -> Tile) before we can try row-major with other dtypes
```
rpavlovicTT: Isn't BFP8 with RM incompatible layout? This comment is a bit confusing, I suggest removing it. It's pretty clear from the code what we do.

lib/Dialect/TTNN/Transforms/OptimizerPasses/OperationValidationAndFallback.cpp
```diff
@@ -781,14 +869,32 @@ bool tryConfigFallbacks(Operation *operation,
               if (result.isSuccess()) {
                 workingConfig = configAttr;
                 workingResult = result;
+                TTMLIR_DEBUG(ttmlir::LogComponent::OpValidation,
+                             "Found working config after {} failed attempts",
+                             failedAttempts);
```
rpavlovicTT: This should be TRACE log. Don't pollute too much DEBUG logs.

lib/Dialect/TTNN/Transforms/OptimizerPasses/Optimizer.cpp
```diff
@@ -60,7 +60,8 @@ TTNNOptimizerOptions::TTNNOptimizerOptions(
       memReconfigEnabled(pipelineOptions.memReconfigEnabled),
       maxLegalLayouts(pipelineOptions.maxLegalLayouts),
       rowMajorEnabled(pipelineOptions.rowMajorEnabled),
-      tensorL1UsageCap(pipelineOptions.tensorL1UsageCap) {}
+      tensorL1UsageCap(pipelineOptions.tensorL1UsageCap),
+      maxFallbackAttempts(pipelineOptions.maxFallbackAttempts) {}
```
rpavlovicTT: Remove these options from optimizer.

lib/Dialect/TTNN/Transforms/OptimizerPasses/Optimizer.cpp
```diff
@@ -188,6 +190,12 @@ class TTNNOptimizerBase : public ::mlir::OperationPass<::mlir::ModuleOp> {
           "Override tensor L1 usage cap in L1 Interleaved Fallback Analysis "
           "and Memory Layout Analysis. [0.0-1.0]"),
       ::llvm::cl::init(1.0f)};
+  ::mlir::Pass::Option<uint32_t> maxFallbackAttempts{
```
rpavlovicTT: Remove

lib/Dialect/TTNN/Transforms/OptimizerPasses/OperationValidationAndFallback.cpp
```diff
@@ -254,10 +292,17 @@ class TTNNOperationValidationAndFallback
...
+      TTMLIR_DEBUG(ttmlir::LogComponent::OpValidation,
+                   "Operation validation FAILED: {} operations checked before "
+                   "failure, {} fixed",
+                   totalOperationsChecked, operationsFixed);
+    } else {
+      TTMLIR_DEBUG(
+          ttmlir::LogComponent::OpValidation,
+          "Operation validation complete: {} operations checked, {} fixed",
+          totalOperationsChecked, operationsFixed);
+    }
```
rpavlovicTT: This can be single log

lib/Dialect/TTNN/Transforms/OptimizerPasses/OperationValidationAndFallback.cpp
```diff
@@ -242,9 +246,43 @@ class TTNNOperationValidationAndFallback
...
+                emitError(operation->getLoc())
+                << "OperationValidationAndFallback: Operation "
+                << operation->getName() << " with " << inputLayouts.size()
+                << " inputs failed validation (original error: " << statusName;
+            if (!originalResult.errorMessage.empty()) {
+              diag << " - " << originalResult.errorMessage;
+            }
+            diag << "). No fallback configuration worked after testing up to " +
+                        std::to_string(maxFallbackAttempts) + " combinations.";
+            diag << "\nOperation IR: " << os.str();
```
rpavlovicTT: This is code is ugly. 
1. Do we really need to emit this message? 
2. If we do, right place is to make a change in `ValidationStatus` and convert its error status to string.

lib/Dialect/TTNN/Transforms/OptimizerPasses/OperationValidationAndFallback.cpp
```diff
@@ -242,9 +246,43 @@ class TTNNOperationValidationAndFallback
...
+              statusName = "UnmatchedReferenceConfig";
+              break;
+            case op_constraint_validation::ValidationStatus::OutOfMemoryError:
+              statusName = "OutOfMemoryError";
+              break;
+            }
+
+            std::string opStr;
+            llvm::raw_string_ostream os(opStr);
+            operation->print(os);
```
rpavlovicTT: By doing `operation->emitError` you will get operation's IR printed.

rpavlovicTT (changes requested):

lib/Dialect/TTNN/Transforms/OptimizerPasses/OperationValidationAndFallback.cpp
bmalesevicTT: Yes, but only at TT-MLIR level as a `note`, it doesn't appear when running through TT-XLA, so the only info we got was the location of the op from `error`, which made finding the problematic op harder, especially on XLA nightly

lib/Dialect/TTNN/Transforms/OptimizerPasses/OperationValidationAndFallback.cpp
bmalesevicTT: Yes, it should be RM, my bad. 
As for removing, I wasn't sure it would be clear that this ordering is important without getting deeper context on the code, if in the future another combination turned out problematic and someone wanted to reorder these again, that they could potentially break these combinations.

include/ttmlir/Dialect/TTNN/Transforms/Passes.td
bmalesevicTT: Made a separate issue, as it requires slight reformatting of L1Interleaved etc. and this PR is already multiconcern https://github.com/tenstorrent/tt-mlir/issues/6429

lib/Dialect/TTNN/Transforms/OptimizerPasses/OperationValidationAndFallback.cpp
bmalesevicTT: I tried to make the intent of the comment clearer now

lib/Dialect/TTNN/Transforms/OptimizerPasses/OperationValidationAndFallback.cpp
bmalesevicTT: In the other reply I mentioned just location isn't quite enough info to analyze where the error originates. Original failure message and ttnn operation name are quite useful for this purpose. Tried to make it cleaner, please add if you see more space for improvement

vmilosevic (commented):

lib/Dialect/TTNN/Transforms/OptimizerPasses/OperationValidationAndFallback.cpp
```diff
@@ -401,28 +449,41 @@ createFallbackTransforms(TTNNLayoutAttr originalLayout,
...
         }
 
-        fallbackLayouts.push_back(result);
+        fallbackLayoutsSet.insert(result);
       }
     }
   }
 
-  return fallbackLayouts;
+  size_t totalCombinations =
```
vmilosevic: :warning: **clang\-analyzer\-deadcode\.DeadStores** :warning:
Value stored to `` totalCombinations `` during its initialization is never read

vmilosevic: :warning: **clang\-diagnostic\-unused\-variable** :warning:
unused variable `` totalCombinations ``

lib/Dialect/TTNN/Transforms/OptimizerPasses/OperationValidationAndFallback.cpp
bmalesevicTT: Added a short explanation in a comment now

include/ttmlir/Dialect/TTNN/Validation/OpConstraintValidation.h
```diff
@@ -25,6 +25,9 @@ enum class ValidationStatus {
   OutOfMemoryError
 };
 
+// Convert ValidationStatus to string for error messages
+const char *validationStatusToString(ValidationStatus status);
```
rpavlovicTT: Let's return StringRef

lib/Dialect/TTNN/Transforms/OptimizerPasses/OperationValidationAndFallback.cpp
```diff
@@ -242,9 +246,9 @@ class TTNNOperationValidationAndFallback
                          "Operation {} at {} fixed with fallback configuration",
                          operation->getName(), operation->getLoc());
           } else {
-            operation->emitError()
-                << "OperationValidationAndFallback: Operation failed "
-                   "validation and no fallback configuration worked";
+            emitValidationFailureError(operation, originalResult,
+                                       maxFallbackAttempts);
```
rpavlovicTT, line 250: This is much better now 👍🏻

lib/Dialect/TTNN/Transforms/OptimizerPasses/OperationValidationAndFallback.cpp
```diff
@@ -388,8 +448,8 @@ createFallbackTransforms(TTNNLayoutAttr originalLayout,
   std::vector<BufferType> targetBufferTypes = {BufferType::DRAM,
                                                BufferType::SystemMemory};
 
-  for (Layout targetLayout : targetLayouts) {
-    for (ttcore::DataType targetDataType : targetDataTypes) {
+  for (ttcore::DataType targetDataType : targetDataTypes) {
+    for (Layout targetLayout : targetLayouts) {
```
rpavlovicTT: Do you gain something by changing order of loops?

include/ttmlir/Dialect/TTNN/Transforms/Passes.td
rpavlovicTT: Fine, let's just have a follow up PR for #6429 soon, to avoid it getting buried in backlog.

lib/Dialect/TTNN/Transforms/OptimizerPasses/OperationValidationAndFallback.cpp
bmalesevicTT: no, adjusted it just to match order of applying for readability, but it doesn't make an actual difference, will revert

odjuricicTT (approved):

PR #6404, opened 2025-12-19
Draft: Analytical Coalescing Factor

bgrady-tt:
The GenericLowerDMA pass currently uses a brute force sampling-based approach to figure out the largest packet size that can be used for NoC transfers. For large

This branch adds an _analytical function_ for computing the coalescing factor using static analysis that is constant-time wrt tensor size. For simple cases (reblocking affine maps, dim permutations) this agrees with reference approach and is much faster. There are no known cases where this code returns an incorrect coalescing factor; however the results 
are not always _optimal_ (brute force check is a multiple of analytical result).

However this analytical check does fail on some complex affine maps. For all known  unanalyzable expressions, `computeCoalescingFactorAnalytically` automatically falls back  to a sampling based approach and prints a warning. If the derived coalesced chunk size is small  enough to violate noc alignment, sampling method is automatically used automatically to avoid creating new bugs due to inadequate coalescing.

### DIFF
--- a/test/unittests/AffineMapUtils/TestAffineMapUtils.cpp
+++ b/test/unittests/AffineMapUtils/TestAffineMapUtils.cpp
@@ -203,7 +203,7 @@ CoalescingTestResult compareCoalescingFactors(int64_t samplingFactor,
   if (samplingFactor == analyticalFactor) {
     return CoalescingTestResult::Success;
   }
-  if (analyticalFactor < samplingFactor &&
+  if (analyticalFactor != 0 && analyticalFactor < samplingFactor &&
       samplingFactor % analyticalFactor == 0) {
     return CoalescingTestResult::Subset;
   }


include/ttmlir/AffineMapUtils.h
```diff
@@ -364,6 +368,1373 @@ buildPhysicalToDeviceMap(mlir::ArrayRef<int64_t> physicalShape,
...
+
+    // Check if LHS is also floordiv/mod - error condition
+    if (auto lhsBinOp = llvm::dyn_cast<mlir::AffineBinaryOpExpr>(lhs)) {
+      auto lhsKind = lhsBinOp.getKind();
+      if (lhsKind == mlir::AffineExprKind::Mod ||
+          lhsKind == mlir::AffineExprKind::FloorDiv) {
+        if (!exprContainsDim(lhsBinOp.getLHS(), dimPos)) {
+          return -1;
+        } else {
+          // Precompute the value of the RHS constant for lhsBinOp
```
vmilosevic: :warning: **llvm\-else\-after\-return** :warning:
```suggestion
        }           // Precompute the value of the RHS constant for lhsBinOp
```

include/ttmlir/AffineMapUtils.h
```diff
@@ -364,6 +368,1373 @@ buildPhysicalToDeviceMap(mlir::ArrayRef<int64_t> physicalShape,
...
+                     kind == mlir::AffineExprKind::Mod) {
+            // discontinuity happens as soon as lhs floordiv expr changes; no
+            // need to analyze outer mod.
+            return analyzeShardResultExprForContiguity(lhsBinOp, dimBounds,
+                                                       dimPos, modulus);
+          } else {
+            // not handled
+            return -1;
+          }
+        }
```
vmilosevic: :warning: **llvm\-else\-after\-return** :warning:
```suggestion
       
```

include/ttmlir/AffineMapUtils.h
```diff
@@ -364,6 +368,1373 @@ buildPhysicalToDeviceMap(mlir::ArrayRef<int64_t> physicalShape,
...
+
+          if (lhsKind == mlir::AffineExprKind::FloorDiv &&
+              kind == mlir::AffineExprKind::FloorDiv) {
+            // Both inner and outer are floordiv, so collapse them into a
+            // single floordiv
+            int64_t combinedDivisor = innerRhsConstValue * modulus;
+            return analyzeShardResultExprForContiguity(
+                lhsBinOp.getLHS().floorDiv(combinedDivisor), dimBounds, dimPos,
+                combinedDivisor);
+          } else if (lhsKind == mlir::AffineExprKind::Mod &&
```
vmilosevic: :warning: **llvm\-else\-after\-return** :warning:

include/ttmlir/AffineMapUtils.h
```diff
@@ -364,6 +368,1373 @@ buildPhysicalToDeviceMap(mlir::ArrayRef<int64_t> physicalShape,
...
+/// @return Coalescing factor limited by stride misalignment:
+///         - std::nullopt: Strides are properly aligned (no constraint)
+///         - Positive value: Maximum coalescing factor due to stride issues
+inline int64_t analyzeShardDimStrides(mlir::AffineMap map,
+                                      mlir::ArrayRef<int64_t> shape,
+                                      unsigned numGridDims,
+                                      unsigned numGridResults) {
+  int numShardDims = shape.size() - numGridDims;
+  TT_assert(numShardDims > 0);
+  // TODO: fix this
```
vmilosevic: :warning: **google\-readability\-todo** :warning:
missing username/bug in TODO
```suggestion
  // TODO(nsmith): fix this
```

include/ttmlir/AffineMapUtils.h
```diff
@@ -364,6 +368,1373 @@ buildPhysicalToDeviceMap(mlir::ArrayRef<int64_t> physicalShape,
...
+      }
+      return stepsNeeded;
+    };
+
+    // Pattern 1: d_dimPos floorDiv N -> need N steps to change output
+    if (auto dimExpr = llvm::dyn_cast<mlir::AffineDimExpr>(lhs)) {
+      if (dimExpr.getPosition() == dimPos) {
+        return checkBoundsAndReturn(divisor);
+      } else {
+        // unrelated dim by itself poses no constraints
```
vmilosevic: :warning: **llvm\-else\-after\-return** :warning:
```suggestion
      }         // unrelated dim by itself poses no constraints
```

include/ttmlir/AffineMapUtils.h
```diff
@@ -364,6 +368,1373 @@ buildPhysicalToDeviceMap(mlir::ArrayRef<int64_t> physicalShape,
...
+        return -1;
+      }
```
vmilosevic: :warning: **llvm\-else\-after\-return** :warning:
```suggestion
     
```

include/ttmlir/AffineMapUtils.h
```diff
@@ -364,6 +368,1373 @@ buildPhysicalToDeviceMap(mlir::ArrayRef<int64_t> physicalShape,
...
+      int64_t mulValue = mulRhsConst.getValue();
+      // Now we know RHS is a constant, just check LHS for dim
+      if (exprIsSpecificDimExpr(maybeMul.value().getLHS(), dimPos)) {
+        // ceil(divisor / mulValue)
+        int64_t stepsNeeded = (divisor + mulValue - 1) / mulValue;
+        return checkBoundsAndReturn(stepsNeeded);
+      } else {
+        return -1;
+      }
```
vmilosevic: :warning: **llvm\-else\-after\-return** :warning:
```suggestion
      }         return -1;
     
```

include/ttmlir/AffineMapUtils.h
```diff
@@ -364,6 +368,1373 @@ buildPhysicalToDeviceMap(mlir::ArrayRef<int64_t> physicalShape,
...
+            if (!dimExpr || !constExpr) {
+              // lhs is binary op expr, can only analyze if this contains the
+              // only reference to the target dim.
+              if (exprContainsDim(sumOperand, dimPos)) {
+                // worst case assume that any change in the lhs expr will change
+                // the output
+                return analyzeGridResultExprForDiscontinuity(sumOperand,
+                                                             dimBounds, dimPos);
+              } else {
+                // don't try to analyze this for now
```
vmilosevic: :warning: **llvm\-else\-after\-return** :warning:
```suggestion
              }                 // don't try to analyze this for now
```

include/ttmlir/AffineMapUtils.h
```diff
@@ -364,6 +368,1373 @@ buildPhysicalToDeviceMap(mlir::ArrayRef<int64_t> physicalShape,
...
+                return 1;
+              }
```
vmilosevic: :warning: **llvm\-else\-after\-return** :warning:
```suggestion
             
```

include/ttmlir/AffineMapUtils.h
```diff
@@ -364,6 +368,1373 @@ buildPhysicalToDeviceMap(mlir::ArrayRef<int64_t> physicalShape,
...
+  case mlir::AffineExprKind::Mul: {
+    // If a mul expression contains the target dim, it implies any change in the
+    // dim will change the output unless this op is the LHS of a floorDiv or
+    // ceilDiv expression that contains the target dim.
+    // if not, by definition it's unconstrained.
+    if (!exprContainsDim(lhs, dimPos)) {
+      return -1;
+    } else {
+      return analyzeGridResultExprForDiscontinuity(lhs, dimBounds, dimPos);
+    }
```
vmilosevic: :warning: **llvm\-else\-after\-return** :warning:
```suggestion
    }       return analyzeGridResultExprForDiscontinuity(lhs, dimBounds, dimPos);
   
```

vmilosevic (commented):

test/unittests/AffineMapUtils/TestAffineMapUtils.cpp
```diff
@@ -0,0 +1,1134 @@
...
+        memoryMap, deviceShape, memoryMap.getNumDims() / 2,
+        memoryMap.getNumResults() - 1, 1);
+
+    // Also compute coalescing factor using the sampling-based approach
+    int64_t coalescingFactor = calculateCoalescingFactor(
+        memoryMap, deviceShape, 1, memoryMap.getNumDims() / 2);
+
+    if (coalescingFactor == coalescingFactorAnalytical) {
+      return TestResult::Success;
+    } else if (coalescingFactorAnalytical < coalescingFactor &&
```
vmilosevic: :warning: **llvm\-else\-after\-return** :warning:
```suggestion
    } if (coalescingFactorAnalytical < coalescingFactor &&
```

lib/Dialect/D2M/Transforms/GenericLowerDMAs.cpp
```diff
@@ -340,7 +340,28 @@ class D2MGenericLowerAffineDMAsRewritePattern : public OpRewritePattern<DMAOp> {
...
 
-    return {streamIndices, coalescingFactor};
+    //--------------- trial analytical coalescing approach --------------//
+    auto altCoalescingFactor = ttmlir::utils::analyzeShardDimContiguity(
+        memoryMap, memref.getShape(), memrefGridShape.size(),
+        memoryMap.getNumResults() - 1, elemSizeBytes);
+    llvm::dbgs() << "[CoalescingFactor] alt = " << altCoalescingFactor << " vs "
+                 << coalescingFactor << "\n";
+
+    if (coalescingFactor % size_t(altCoalescingFactor) != 0) {
```
vmilosevic: :warning: **google\-readability\-casting** :warning:
C\-style casts are discouraged; use static\_cast
```suggestion
    if (coalescingFactor % static_cast<size_t>(altCoalescingFactor) != 0) {
```

vmilosevic (commented):

include/ttmlir/AffineMapUtils.h
```diff
@@ -364,6 +368,1456 @@ buildPhysicalToDeviceMap(mlir::ArrayRef<int64_t> physicalShape,
...
+  case mlir::AffineExprKind::Mod:
+  case mlir::AffineExprKind::Mul: {
+    // If a mul expression contains the target dim, it implies any change in the
+    // dim will change the output unless this op is the LHS of a floorDiv or
+    // ceilDiv expression that contains the target dim.
+    // if not, by definition it's unconstrained.
+    if (!exprContainsDim(lhs, dimPos)) {
+      return -1;
+    } else {
+      return analyzeGridResultExprForDiscontinuity(lhs, dimBounds, dimPos,
```
vmilosevic: :warning: **llvm\-else\-after\-return** :warning:
```suggestion
    }       return analyzeGridResultExprForDiscontinuity(lhs, dimBounds, dimPos,
```

include/ttmlir/AffineMapUtils.h
```diff
@@ -364,6 +368,1456 @@ buildPhysicalToDeviceMap(mlir::ArrayRef<int64_t> physicalShape,
...
+                                                   numGridDims);
+    }
```
vmilosevic: :warning: **llvm\-else\-after\-return** :warning:
```suggestion
   
```

include/ttmlir/AffineMapAnalysis.h
```diff
@@ -0,0 +1,1328 @@
...
+
+    // Check if LHS is also floordiv/mod - error condition
+    if (auto lhsBinOp = llvm::dyn_cast<mlir::AffineBinaryOpExpr>(lhs)) {
+      auto lhsKind = lhsBinOp.getKind();
+      if (lhsKind == mlir::AffineExprKind::Mod ||
+          lhsKind == mlir::AffineExprKind::FloorDiv) {
+        if (!exprContainsDim(lhsBinOp.getLHS(), dimPos)) {
+          return -1;
+        } else {
+          // Precompute the value of the RHS constant for lhsBinOp
```
vmilosevic: :warning: **llvm\-else\-after\-return** :warning:
```suggestion
        }           // Precompute the value of the RHS constant for lhsBinOp
```

include/ttmlir/AffineMapAnalysis.h
```diff
@@ -0,0 +1,1328 @@
...
+                     kind == mlir::AffineExprKind::Mod) {
+            // discontinuity happens as soon as lhs floordiv expr changes; no
+            // need to analyze outer mod.
+            return analyzeShardResultExprForContiguity(
+                lhsBinOp, dimBounds, dimPos, numGridDims, modulus);
+          } else {
+            // not handled
+            return -1;
+          }
+        }
```
vmilosevic: :warning: **llvm\-else\-after\-return** :warning:
```suggestion
       
```

include/ttmlir/AffineMapAnalysis.h
```diff
@@ -0,0 +1,1328 @@
...
+
+          if (lhsKind == mlir::AffineExprKind::FloorDiv &&
+              kind == mlir::AffineExprKind::FloorDiv) {
+            // Both inner and outer are floordiv, so collapse them into a
+            // single floordiv
+            int64_t combinedDivisor = innerRhsConstValue * modulus;
+            return analyzeShardResultExprForContiguity(
+                lhsBinOp.getLHS().floorDiv(combinedDivisor), dimBounds, dimPos,
+                numGridDims, combinedDivisor);
+          } else if (lhsKind == mlir::AffineExprKind::Mod &&
```
vmilosevic: :warning: **llvm\-else\-after\-return** :warning:

include/ttmlir/AffineMapAnalysis.h
```diff
@@ -0,0 +1,1328 @@
...
+///         - std::nullopt: Strides are properly aligned (no constraint)
+///         - Positive value: Maximum coalescing factor due to stride issues
+inline int64_t analyzeShardDimStrides(mlir::AffineMap map,
+                                      mlir::ArrayRef<int64_t> shape,
+                                      unsigned numGridDims,
+                                      unsigned numGridResults,
+                                      int64_t elemSizeBytes) {
+  int numShardDims = shape.size() - numGridDims;
+  TT_assert(numShardDims > 0);
+  // TODO: fix this
```
vmilosevic: :warning: **google\-readability\-todo** :warning:
missing username/bug in TODO
```suggestion
  // TODO(nsmith): fix this
```

include/ttmlir/AffineMapAnalysis.h
```diff
@@ -0,0 +1,1328 @@
...
+      }
+      return stepsNeeded;
+    };
+
+    // Pattern 1: d_dimPos floorDiv N -> need N steps to change output
+    if (auto dimExpr = llvm::dyn_cast<mlir::AffineDimExpr>(lhs)) {
+      if (dimExpr.getPosition() == dimPos) {
+        return checkBoundsAndReturn(divisor);
+      } else {
+        // unrelated dim by itself poses no constraints
```
vmilosevic: :warning: **llvm\-else\-after\-return** :warning:
```suggestion
      }         // unrelated dim by itself poses no constraints
```

include/ttmlir/AffineMapAnalysis.h
```diff
@@ -0,0 +1,1328 @@
...
+        return -1;
+      }
```
vmilosevic: :warning: **llvm\-else\-after\-return** :warning:
```suggestion
     
```

include/ttmlir/AffineMapAnalysis.h
```diff
@@ -0,0 +1,1328 @@
...
+      int64_t mulValue = mulRhsConst.getValue();
+      // Now we know RHS is a constant, just check LHS for dim
+      if (exprIsSpecificDimExpr(maybeMul.value().getLHS(), dimPos)) {
+        // ceil(divisor / mulValue)
+        int64_t stepsNeeded = (divisor + mulValue - 1) / mulValue;
+        return checkBoundsAndReturn(stepsNeeded);
+      } else {
+        return -1;
+      }
```
vmilosevic: :warning: **llvm\-else\-after\-return** :warning:
```suggestion
      }         return -1;
     
```

include/ttmlir/AffineMapAnalysis.h
```diff
@@ -0,0 +1,1328 @@
...
+            if (!dimExpr || !constExpr) {
+              // lhs is binary op expr, can only analyze if this contains the
+              // only reference to the target dim.
+              if (exprContainsDim(sumOperand, dimPos)) {
+                // worst case assume that any change in the lhs expr will change
+                // the output
+                return analyzeGridResultExprForDiscontinuity(
+                    sumOperand, dimBounds, dimPos, numGridDims);
+              } else {
+                // don't try to analyze this for now
```
vmilosevic: :warning: **llvm\-else\-after\-return** :warning:
```suggestion
              }                 // don't try to analyze this for now
```

include/ttmlir/AffineMapAnalysis.h
```diff
@@ -0,0 +1,1328 @@
...
+                return 1;
+              }
```
vmilosevic: :warning: **llvm\-else\-after\-return** :warning:
```suggestion
             
```

include/ttmlir/AffineMapAnalysis.h
```diff
@@ -0,0 +1,1328 @@
...
+  case mlir::AffineExprKind::Mod:
+  case mlir::AffineExprKind::Mul: {
+    // If a mul expression contains the target dim, it implies any change in the
+    // dim will change the output unless this op is the LHS of a floorDiv or
+    // ceilDiv expression that contains the target dim.
+    // if not, by definition it's unconstrained.
+    if (!exprContainsDim(lhs, dimPos)) {
+      return -1;
+    } else {
+      return analyzeGridResultExprForDiscontinuity(lhs, dimBounds, dimPos,
```
vmilosevic: :warning: **llvm\-else\-after\-return** :warning:
```suggestion
    }       return analyzeGridResultExprForDiscontinuity(lhs, dimBounds, dimPos,
```

include/ttmlir/AffineMapAnalysis.h
```diff
@@ -0,0 +1,1328 @@
...
+                                                   numGridDims);
+    }
```
vmilosevic: :warning: **llvm\-else\-after\-return** :warning:
```suggestion
   
```

lib/Dialect/D2M/Transforms/GenericLowerDMAs.cpp
```diff
@@ -337,8 +303,53 @@ class D2MGenericLowerAffineDMAsRewritePattern : public OpRewritePattern<DMAOp> {
...
-    size_t coalescingFactor = calculateCoalescingFactor(
-        memoryMap, memrefGridShape, memrefShardShape, elemSizeBytes);
+
+    auto coalescingFactor = ttmlir::utils::analyzeShardDimContiguity(
+        memoryMap, memref.getShape(), memrefGridShape.size(), elemSizeBytes);
+
+    // When debug mode is enabled, also run the slow sampling-based coalescing
+    // factor computation and compares results. If their is a discrepancy, the
+    // sampled value is used as the fallback.
+    if (debugCoalescingInference) {
```
bgrady-tt: If the CLI option `debug-d2m-coalescing-inference` is set to `true`, extra details are logged about any differences between sampling based coalescing factor calculation and the new analytical one. The sample based coalescing is used instead of the analytical one, to allow E2E comparison (for diffing perf, functionality, etc).

test/unittests/AffineMapUtils/TestAffineMapUtils.cpp
```diff
@@ -0,0 +1,1567 @@
...
+
+    int64_t coalescingFactor = calculateCoalescingFactor(
+        memoryMap, deviceShape, 1, memoryMap.getNumDims() / 2);
+
+    CoalescingTestResult result =
+        compareCoalescingFactors(coalescingFactor, coalescingFactorAnalytical);
+
+    return {result,           memoryMap,
+            inputDeviceShape, outputDeviceShape,
+            coalescingFactor, coalescingFactorAnalytical};
```
vmilosevic, line 471: :warning: **clang\-diagnostic\-c\+\+11\-narrowing** :warning:
non\-constant\-expression cannot be narrowed from type `` size_t `` \(aka `` unsigned long ``\) to `` int64_t `` \(aka `` long ``\) in initializer list

vmilosevic (commented):

include/ttmlir/Dialect/D2M/Transforms/Passes.td
```diff
@@ -391,6 +391,11 @@ def D2MGenericLowerDMAs : Pass<"d2m-generic-lower-dmas", "::mlir::ModuleOp"> {
     %tx = d2m.dma %stream [%1, %iter2], %cb0
     ```
   }];
+  let options = [
+    Option<"debugCoalescingInference", "debug-d2m-coalescing-inference", "bool", /*default=*/"false",
```
bgrady-tt, line 395: Another more conservative option here is to have the analytical coalescing method be _optional_. Slow tests could enable it as needed.

include/ttmlir/AffineMapAnalysis.h
```diff
@@ -0,0 +1,1297 @@
...
+#include "mlir/IR/AffineMap.h"
+#include "llvm/ADT/DenseMap.h"
+#include "llvm/ADT/DenseSet.h"
+#include "llvm/ADT/STLExtras.h"
+#include "llvm/ADT/SmallVector.h"
+
+#include <numeric>
+#include <optional>
+
+namespace ttmlir::utils {
```
bgrady-tt, line 24: All of the analysis code for affine map expressions is in this header.

#### Breakdown
- `computeCoalescingFactorAnalytically` : top-level API that is called as 1:1 replacement for existing coalescing check.
- `analyzeShardResultExprForContiguity` : Analyzes a single _shard result_ AffineExpr and dim expression (i.e. d0) to find a lower-bound on maximum stride for that dim.
- `analyzeGridResultExprForDiscontinuity` : Analyzes a single _grid result_ AffineExpr and dim expression (i.e. d0) to detect when changes in the dim variable cause grid dimensions to change (i.e. switching banks)
- `analyzeShardDimStrides` : Analyzes affine map expressions to infer effective strides of each dim variable. Misalignment in strides (e.g. due to padding) leads to noncontiguous access.
- `simplifyAffineMapWithRangeAnalysis` : Uses shape bounds to simplify affine map result expressions. This is a generic pre-processing step, but results in less conservatism by eliminating mod/floordiv ops that either do nothing or are always `0`.

include/ttmlir/AffineMapAnalysis.h
```diff
@@ -0,0 +1,1297 @@
...
+  return expr;
+}
+
+/// Simplifies an affine map by eliminating redundant mod/floordiv operations
+/// using range analysis. If the upper bound of an operand is less than the
+/// modulus, the mod is removed; if less than the divisor, floordiv becomes 0.
+/// When performEquivalenceCheck is true, samples the domain to verify the
+/// simplified map is equivalent; returns the original map if not.
+inline mlir::AffineMap
+simplifyAffineMapWithRangeAnalysis(mlir::AffineMap map,
```
vwellsTT, line 49: are there any users where `performEquivalenceCheck` is overriden? Personally hate default params in C++, would vote to just remove or specify at call sites

include/ttmlir/AffineMapAnalysis.h
```diff
@@ -0,0 +1,1297 @@
...
+inline std::optional<mlir::AffineBinaryOpExpr>
+isBinaryMul(mlir::AffineExpr expr) {
+  auto lhsBinExpr = llvm::dyn_cast<mlir::AffineBinaryOpExpr>(expr);
+  return (lhsBinExpr && lhsBinExpr.getKind() == mlir::AffineExprKind::Mul)
+             ? std::optional<mlir::AffineBinaryOpExpr>(lhsBinExpr)
+             : std::nullopt;
+}
+
+inline void
+collectSumOperandsImpl(mlir::AffineExpr expr,
```
vwellsTT: don't get the motivation for having an impl func here

include/ttmlir/AffineMapAnalysis.h
```diff
@@ -0,0 +1,1297 @@
...
+    });
+    if (!equivalent) {
+      return map;
+    }
+  }
+
+  return simplifiedMap;
+}
+
+inline mlir::AffineExpr isBinaryAdd(mlir::AffineExpr expr) {
```
vwellsTT: I think this should have the same type as isBinaryMul below? And for both I think `isX` ought to return bool, something like `tryGetX` is appropriate for a func that might return a value or a nullopt/empty val.

include/ttmlir/AffineMapAnalysis.h
```diff
@@ -0,0 +1,1297 @@
...
+}
+
+inline mlir::SmallVector<mlir::AffineExpr>
+collectSumOperands(mlir::AffineExpr expr) {
+  mlir::SmallVector<mlir::AffineExpr> ops;
+  collectSumOperandsImpl(expr, ops);
+  return ops;
+}
+
+/// Finds the minimum gap to the next target boundary for any achievable sum.
```
vwellsTT: likely this is due to personal ignorance, but I honestly don't know what any of the terms mean.  I see "gap" is defined/given an example below, but I still find this unclear; what is "target" and "sum"? What is an "achievable sum"?

include/ttmlir/AffineMapAnalysis.h
```diff
@@ -0,0 +1,1297 @@
...
+  return foundDim;
+}
+
+inline bool exprIsSpecificDimExpr(mlir::AffineExpr expr, unsigned dimPos) {
+  return llvm::isa<mlir::AffineDimExpr>(expr) &&
+         llvm::dyn_cast<mlir::AffineDimExpr>(expr).getPosition() == dimPos;
+}
+
+/// Recursively analyzes contiguity of an affine expression to find the maximum
+/// coalescable run. Returns -1 if unconstrained, 1 on error, or the contiguity
```
vwellsTT: Not a big fan of using 1 & -1 as magic numbers here, maybe we should define a struct with an enum + std::variant/int value or something like that?

include/ttmlir/AffineMapAnalysis.h
```diff
@@ -0,0 +1,1297 @@
...
+    // RHS must be constant.
+    auto rhsConstExpr = llvm::dyn_cast<mlir::AffineConstantExpr>(rhs);
+    assert(rhsConstExpr && "Mod/FloorDiv RHS must be constant expression");
+
+    int64_t modulus = rhsConstExpr.getValue();
+    if (modulus <= 0) {
+      return 1;
+    }
+
+    // Check if LHS is also floordiv/mod - error condition.
```
vwellsTT: because precedence rules mean we should've already processed, or do we not support multiple floordiv/mods chained?

include/ttmlir/AffineMapAnalysis.h
```diff
@@ -0,0 +1,1297 @@
...
+      return expr;
+    }
+  }
+  return expr;
+}
+
+/// Simplifies an affine map by eliminating redundant mod/floordiv operations
+/// using range analysis. If the upper bound of an operand is less than the
+/// modulus, the mod is removed; if less than the divisor, floordiv becomes 0.
+/// When performEquivalenceCheck is true, samples the domain to verify the
```
bgrady-tt, line 46: NOTE: the equivalence check here is SLOW as it implies sampling both input and result map through the entire domain; this is mostly a debugging feature.

include/ttmlir/AffineMapAnalysis.h
```diff
@@ -0,0 +1,1297 @@
...
+  mlir::SmallVector<mlir::AffineExpr> ops;
+  collectSumOperandsImpl(expr, ops);
+  return ops;
+}
+
+/// Finds the minimum gap to the next target boundary for any achievable sum.
+/// For floor division semantics: gap = target - (sum % target), or target if
+/// sum is exactly on a boundary.
+/// Returns the minimum such gap across all achievable sums.
+inline int64_t minimizeGap(int64_t target, llvm::ArrayRef<int64_t> multipliers,
```
bgrady-tt: This helper is used to compute what the lower-bound on coalescing is when we find an expression like:
`(A*d0 + B*d1 + ... + CONSTANT) mod N`. In this case, we find the smallest difference between a multiple of N and the sum of everything _but_ the sub expression with the target dim. That gap implies the maximum coalescable stride before wrapping to zero across all possible values for other dim variables.

include/ttmlir/AffineMapAnalysis.h
bgrady-tt: I can make the equivalence check explicit at each call site 👍

include/ttmlir/AffineMapAnalysis.h
bgrady-tt: Without a helper function, the recursive call would need to pass an empty `results` vector, which seems clunky. A pure functional approach (just return `SmallVector` always) is fine by me, but results in a lot of pointless copying/merging of SmallVectors.

vwellsTT: ah, yeah fair enough

include/ttmlir/AffineMapAnalysis.h
vwellsTT: Gotcha, that makes more sense thanks

include/ttmlir/AffineMapAnalysis.h
bgrady-tt: This is a generic function that is new to me as well; it figures out the worst-case (minimum) gap to a threshold sum. This is sometimes called the change making problem or [subset sum](https://chaoxu.prof/files/papers/subset-sum-simple.pdf). This is basically Bellman's dynamic programming version of solving this problem.

The problem it solves is very similar to something along the lines of: '_given N quarters and M dimes, what is the closest to a **multiple of** $1.24 I can get without exceeding N*1.24 _?' . 

- The **target** here is $1.24
- The **multipliers** are the values of each coin type (25¢,10¢, ...)
- The **bounds** are N and M (max amount of each coin type)
- The **constant** is a fixed amount always added to the sum.

Achievable should just be 'possible'.

include/ttmlir/AffineMapAnalysis.h
bgrady-tt: I can refactor to use `std::variant` to explicitly signal the different cases here.

include/ttmlir/AffineMapAnalysis.h
bgrady-tt: That is an errant comment, I will remove it. This code does analyze nested floordiv/mod (with some limits).

include/ttmlir/AffineMapAnalysis.h
bgrady-tt: I can fix the inconsistency here; `getIfBinaryAdd` is more descriptive.

vwellsTT: yep, that name SGTM

vwellsTT:
I'll plan to finish reviewing later today or tomorrow btw, need a bit more time to wrap my head around some of the concepts here.  Broadly looking good though, I think these sort of robust utilities are pretty needed as we use affine maps more and more.

lib/Dialect/D2M/Transforms/GenericLowerDMAs.cpp
```diff
@@ -337,8 +304,66 @@ class D2MGenericLowerAffineDMAsRewritePattern : public OpRewritePattern<DMAOp> {
         viewInterface.applyViews();
     AffineMap memoryMap = device.getMemoryMap(underlyingMemrefAndView,
                                               0 /* use default page size*/);
-    size_t coalescingFactor = calculateCoalescingFactor(
-        memoryMap, memrefGridShape, memrefShardShape, elemSizeBytes);
+
+    auto coalescingFactor = ttmlir::utils::computeCoalescingFactorAnalytically(
+        memoryMap, memref.getShape(), memrefGridShape.size(), elemSizeBytes);
+
+    // If the coalescing factor is less than the sampling fallback threshold,
```
dlokeTT: Why choose 16 as min size for fallback? Also is this precautionary or have you found cases where analytical calculation is more conservative than sampling?

include/ttmlir/AffineMapUtils.h
```diff
@@ -9,11 +9,15 @@
 #include "ttmlir/Utils.h"
 
 #include "mlir/Conversion/AffineToStandard/AffineToStandard.h"
+
 #include "mlir/Dialect/Affine/IR/AffineOps.h"
 #include "mlir/IR/AffineMap.h"
 #include "mlir/IR/BuiltinAttributes.h"
 #include "llvm/ADT/STLExtras.h"
 #include "llvm/ADT/SmallVector.h"
+#include <atomic>
```
vwellsTT, line 18: nit: \n here

include/ttmlir/Dialect/D2M/Transforms/Passes.td
vwellsTT: yeah, off the top of my head I don't think we have any other "debug" passes, but I think it seems like a reasonable paradigm to me, idk

include/ttmlir/AffineMapAnalysis.h
```diff
@@ -0,0 +1,1383 @@
+// SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
```
vwellsTT, line 1: nit: 2026 I guess :)

include/ttmlir/AffineMapAnalysis.h
```diff
@@ -0,0 +1,1383 @@
...
+  int64_t value;
+  explicit ConstrainedBound(int64_t v) : value(v) {}
+};
+
+/// Represents an unanalyzable expression (treated as ConstrainedBound with
+/// value 1).
+struct UnanalyzableBound {};
+
+/// Variant type for contiguity analysis results.
+using ContiguityBound =
```
vwellsTT, line 40: thanks, I like this a lot better!

include/ttmlir/AffineMapAnalysis.h
```diff
@@ -0,0 +1,1383 @@
...
+    if (constExpr.getValue() < 0) {
+      return std::nullopt;
+    }
+    return constExpr.getValue();
+  }
+  if (auto binOp = llvm::dyn_cast<mlir::AffineBinaryOpExpr>(expr)) {
+
+    auto lhs = getExprUpperBound(binOp.getLHS(), dimBounds);
+
+    // Quick check for Mod with constant RHS.
```
vwellsTT: It took me a bit to follow why this matters/implies return values below, maybe comment could be more explicit

include/ttmlir/AffineMapAnalysis.h
```diff
@@ -0,0 +1,1383 @@
...
+
+  for (size_t i = 0; i < multipliers.size(); ++i) {
+    llvm::DenseSet<int64_t> newRemainders;
+    int64_t multiplierMod =
+        ((multipliers[i] % targetSum) + targetSum) % targetSum;
+    for (int64_t remainder : possibleRemainders) {
+      for (int64_t n = 0; n <= bounds[i]; ++n) {
+        int64_t newRemainder = (remainder + multiplierMod * n) % targetSum;
+        newRemainders.insert(newRemainder);
+        // Early exit: if we've covered all remainders, no need to continue
```
vwellsTT: nit: this and next comment need .

include/ttmlir/AffineMapAnalysis.h
```diff
@@ -0,0 +1,1383 @@
...
+  // Find the minimum non-zero gap to any integer multiple of targetSum for any
+  // possible remainder.
+  int64_t bestGap = targetSum;
+
+  for (int64_t remainder : possibleRemainders) {
+    int64_t gap = (remainder == 0) ? targetSum : (targetSum - remainder);
+    if (gap < bestGap) {
+      bestGap = gap;
+    }
+    // Early exit if gap is 1 (best possible non-exact alignment)
```
vwellsTT: nit: .

include/ttmlir/AffineMapAnalysis.h
```diff
@@ -0,0 +1,1383 @@
...
+    // If rhsConst % constVal != 0, return unanalyzable.
+    return UnanalyzableBound{};
+  }
+
+  case mlir::AffineExprKind::Mod:
+  case mlir::AffineExprKind::FloorDiv:
+  case mlir::AffineExprKind::CeilDiv: {
+    // RHS must be constant.
+    auto rhsConstExpr = llvm::dyn_cast<mlir::AffineConstantExpr>(rhs);
+    assert(rhsConstExpr && "Mod/FloorDiv RHS must be constant expression");
```
vwellsTT: nit: TT_assertv

include/ttmlir/AffineMapAnalysis.h
```diff
@@ -0,0 +1,1383 @@
...
+        }
+
+        // If we found the target dim, compute worst-case contiguity.
+        if (foundTargetDim && targetMultiplier > 0) {
+          if (constSum < 0) {
+            return UnanalyzableBound{}; // Negative constants not analyzed,
+                                        // fallback.
+          }
+
+          int64_t gap;
```
vwellsTT: nit: prefer if ya 0 init this

include/ttmlir/AffineMapAnalysis.h
```diff
@@ -0,0 +1,1383 @@
...
+          int64_t stepsNeeded =
+              (gap + multiplier - 1) / multiplier; // Ceiling division
+          return checkBoundsAndReturn(stepsNeeded);
+        }
+      }
+    }
+
+    // Cannot analyze, conservative fallback.
+    return UnanalyzableBound{};
+  }
```
vwellsTT: I think this whole case block is too big, it probably should be standalone func/funcs vs 300 lines in a switch

test/unittests/AffineMapUtils/TestAffineMapUtils.cpp
```diff
@@ -0,0 +1,1437 @@
...
+/// Success: factors match exactly.
+/// Subset: analytical < sampling and sampling is divisible by analytical.
+/// Failed: factors don't match and aren't in subset relationship.
+CoalescingTestResult compareCoalescingFactors(int64_t samplingFactor,
+                                              int64_t analyticalFactor) {
+  if (samplingFactor == analyticalFactor) {
+    return CoalescingTestResult::Success;
+  }
+  if (analyticalFactor < samplingFactor &&
+      samplingFactor % analyticalFactor == 0) {
```
vmilosevic, line 207: :warning: **clang\-analyzer\-core\.DivideZero** :warning:
Division by zero

nsmithtt (approved):

PR #6406, opened 2025-12-20
outer permute, part 1

anuragsingh-tt:
### Problem description
Pt 1: simple outer permute + tests.


lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
```diff
@@ -580,7 +580,8 @@ class D2MNamedElementwiseRewriter final
             loc,
             /* result tensor types */
             llvm::to_vector(
-                mlir::ValueRange(blockArgs.take_back(numOutputs)).getTypes()),
+                static_cast<mlir::ValueRange>(blockArgs.take_back(numOutputs))
```
nsmithtt: I don't think this is a valid static_cast, the old code reads more correct to me, we're constructing a ValueRange from an ArrayRef.

PR #6411, opened 2025-12-21
D2M: fix missed alignments for certain TMs

### LINKED ISSUE #6268: Rearrange PCC Error - Virtual Grid Host Upload
The following test (currently marked as skip) fails with a PCC error.  I'm almost certain this is due to host upload/download of the ND virtual grid'd tensor, similar to the bug that @wenbinlyuTT discovered for middle dims > 1.

```bash
pytest -svv "test/python/golden/test_rearrange.py::test_rearrange[ttmetal-shape9-z y x -> z (y x)]"
```

wenbinlyuTT:
Closes #6268

### Bug 1
When the TTIRToD2M pass creates StreamLayoutOps for TMs, it uses unit grids with default alignments and no collapsing.

In the GridSelection pass, a grid is applied to the storage of the StreamLayoutOps and it's possible to get an higher alignment requirement than the default.

Example: apply rearrange with collapse `z y x -> y (z x)` on a `33x2x8` tensor to get a `2x264` tensor on a `8x8` grid will endup with a `32x256` alignment requirement on the stream's storage.

When the GridSelection pass creates a new stream storage on an optimal grid, a reblock map from the new optimal storage to the old unit grid shape is prepended to the stream's affine map, so the semantics of the stream is preserved.

However, this reblock map makes no sense when these two shapes have different dimAlignments: their volumes are different and their strides are incompatiable and so the index transformation is incorrect.

Solution: when an alignment change is detected, we align-up the old unit grid shape with the new alignments, and calculate the correct reblock map using it.

Since alignment paddings are applied at the end of the dimensions, this align-up "aligns" the first step and doesn't affect the second step of the index transformation chain `optimal_output->unit_output->tm_map->input_index_map`.

### Bug 2
When applying rearrangement `z y x -> (z y) x` to `4x85x1055` to get `340x1055`, the stream's unit grid storage's shape would be `352x1056` and dimAlignments would be `256x256` from the start, so we won't encounter the bug above.

The missed alignment problem manifests later, in the LowerToLayout pass, and on the stream's input this time.

Since the GridSelection pass selected optimal (virtual) grid `2x3x8` for the input tensor, its device shape would be `2x3x8x2x32x160` with alignment `1x32x256` and no collapse.

When we lower the ToLayoutOp for the host to device transfer of this input we trigger the virtual bounce and collapse the optimal grid to be `6x8`.

This 3D to 2D collapse leads to a bigger alignment of `256x32x256` and changes the shape of the bounce tensor, and in this particular case, made it indivisible by grid `6x8`.

Solution: when an alignment change is detected, bounce the H2D transfer through an interleaved DRAM tensor on the unit grid, and let the DRAM->L1 lowering step take over.

### Misc
- Add an assert in reblock map calculation to catch obvious nonsensical inputs.
- Correct confusing variable names.

### DIFF
--- a/lib/Dialect/D2M/Transforms/LowerToLayout.cpp
+++ b/lib/Dialect/D2M/Transforms/LowerToLayout.cpp
@@ -138,8 +138,8 @@ class D2MLowerToLayoutRewriter : public OpRewritePattern<ToLayoutOp> {
       auto newIndexMap = referenceLayout.getIndexAffineMap();
       bool hasNonTrivialIndexMap =
           !newIndexMap.isEmpty() && !newIndexMap.isIdentity();
-      ttcore::MetalLayoutAttr layout;
 
+      ttcore::MetalLayoutAttr layout;
       if (hasNonTrivialIndexMap && virtualBounceNeeded) {
         // If the reference layout has a virtual grid, the bounce shape should
         // not copy its index map.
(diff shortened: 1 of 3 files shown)


lib/Dialect/D2M/Transforms/GridSelection.cpp
```diff
@@ -562,6 +556,23 @@ updateStreamLayoutOps(ArrayRef<StreamLayoutUpdateInfo> streamLayoutsToUpdate,
...
+    // calculation from the old unit grid storage shape to the new optimal grid
+    // storage shape (different strides & volumes). In this case, align-up the
+    // old storage shape by overwriting it with the new storage's shape on a
+    // unit grid.
+    llvm::SmallVector<int64_t> oldStorageShape(storageType.getShape());
+    if (!llvm::equal(storageLayout.getDimAlignments(),
+                     newStorageLayout.getDimAlignments())) {
+      TT_assert(
+          ttmlir::utils::volume(storageLayout.getGridShape(storageType)) == 1);
+      oldStorageShape = newStorageLayout.getDeviceShape(
```
bgrady-tt, line 570: This is a basic question, but does the alignment of the storage shape matter beyond (perhaps) requiring the allocation be aligned up? I am seeing a lot of workarounds for getting stream storage correct, but its layout is essentially _unused_ by any downstream pass; it is just a way to ensure allocation of a CB for streaming.

bgrady-tt: We are preparing to refactor stream layout out of existence, but it might be worth simply eliminating a lot of these (badly conceived) constraints? The stream storage could simply be:

```
// where MxN is the shard shape only
// affine map is enforced to be empty
// alignment _can_ be set to align up allocation if necessary
%storage = d2m.empty() : tensor<MxNxTYPE, ... index_map = map(0)>
```

lib/Dialect/D2M/Transforms/LowerToLayout.cpp
```diff
@@ -144,30 +138,44 @@ class D2MLowerToLayoutRewriter : public OpRewritePattern<ToLayoutOp> {
...
         auto [collapsedIntervals, dimAlignments] =
             computeGridAwareCollapsedIntervalsAndDimAlignments(referenceLayout,
                                                                targetGridShape);
-        layout = ttcore::MetalLayoutAttr::get(
-            ctx, referenceLayout.getLogicalShape(), dimAlignments,
-            collapsedIntervals, referenceLayout.getOobVal(), memSpace,
-            referenceLayout.getMemoryLayout(), newIndexMap);
-
+        // Bounce through interleaved DRAM if dimAlignments has changed.
+        if (llvm::equal(dimAlignments, referenceLayout.getDimAlignments())) {
```
bgrady-tt: I think we should _**always**_ bounce through DRAM; it facilitates testing close-to-L1-limits tensors and gives us 1 code path to debug. Thoughts?

lib/Dialect/D2M/Transforms/GridSelection.cpp
wenbinlyuTT: I am also under the impression that the storage shape's alignment doesn't matter much, I does feel like we're just handling our self-imposed alignment compatibility constraints for the reblock calculation here.

In fact, I am starting to question if we still need this whole 32 v.s. 256 (or 32x10 32x13 for BH?) alignment change now that we have virtual grids... Simple tile-aligned H&W dims should be enough?

Elimination of this alignment change can probably happen before stream 2.0 and will get rid a lot of WAs like this.

bgrady-tt: I agree that as long as ND virtual grids are mutually exclusive of dim collapsing (probably a wise decision), then dim alignments within virtual grid tensors are driven by tile alignment only.

nsmithtt: The reason we want to align to `grid_dim*32`, 256 in the case of WH, is to handled cases where you might have a logical width of 416.

Tilizing 416 gives you 13 tiles wide which doesn't divide onto any grid.  Instead if we align-up to 512, now we can fill the whole grid nicely.

lib/Dialect/D2M/Transforms/LowerToLayout.cpp
wenbinlyuTT: @bgrady-tt I have switched to always bounce through DRAM here and the tests are passing!

PR #6421, opened 2025-12-22
Use top-down traversal during Sdy CCLs to SHLO CCLs pattern rewriter

hshahTT:
### Ticket
Fixes https://github.com/tenstorrent/tt-mlir/issues/6157

### Problem description
The current pattern rewriter in `convertShardyCCLToStableHLOCCL()` within the `UpdateGlobalToLocalShapesPass` uses bottom-up traversal. This leads to cases where, when the result of a Sdy CCL op is fed as input to another Sdy CCL op, the rewriter fails to update the shape correctly and causes the pass to fail. See attached ticket for more context.

### What's changed
- Use top-down traversal instead so that the shapes are updated in the correct order.

### Checklist

### DIFF
--- a/lib/Dialect/StableHLO/Transforms/UpdateGlobalToLocalShapes.cpp
+++ b/lib/Dialect/StableHLO/Transforms/UpdateGlobalToLocalShapes.cpp
@@ -349,6 +349,9 @@ convertShardyCCLToStableHLOCCL(MLIRContext *context,
   }();
 
   config.enableConstantCSE(!hasProtectedConst);
+  // Fixes: https://github.com/tenstorrent/tt-mlir/issues/6157
+  // TODO(hshah): See if the above issue can be fixed by modifying the pattern
+  // rewriter directly (without needing to use top-down traversal)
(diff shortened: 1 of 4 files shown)


lib/Dialect/StableHLO/Transforms/UpdateGlobalToLocalShapes.cpp
```diff
@@ -349,6 +349,10 @@ convertShardyCCLToStableHLOCCL(MLIRContext *context,
   }();
 
   config.enableConstantCSE(!hasProtectedConst);
+  // Fixes: https://github.com/tenstorrent/tt-mlir/issues/6157
+  // TODO: See if the above issue can be fixed by modifying the pattern rewriter
```
vmilosevic: :warning: **google\-readability\-todo** :warning:
missing username/bug in TODO
```suggestion
  // TODO(nsmith): See if the above issue can be fixed by modifying the pattern rewriter
```

mmanzoorTT (approved):

PR #6422, opened 2025-12-22
[D2M] Add support for non-square grids in Grid Selection (for improved perf on BH)

sohaibnadeemTT:
### Ticket
Link to Github Issue

### Problem description

### What's changed

### Checklist

### DIFF
--- a/lib/Dialect/D2M/Transforms/GridSelection.cpp
+++ b/lib/Dialect/D2M/Transforms/GridSelection.cpp
@@ -448,7 +448,7 @@ normalizeOperandGridsForGeneric(
   }
 
   // For each loop dimension that is used by multiple inputs, promote the grid
-  // size associated with that loop dimension to the maximum across those
+  // size associated with that loop dimension to the minimum across those
   // inputs.
   for (auto &it : dimToInputOperandDims) {
     auto &entries = it.second;


lib/Dialect/D2M/Transforms/GridSelection.cpp
```diff
@@ -408,6 +403,121 @@ struct EmptyUpdateInfo {
...
+        continue;
+      }
+      int64_t loopDim = dimExpr.getPosition();
+      dimToInputOperandDims[loopDim].push_back(
+          std::make_pair(operandIndex, static_cast<unsigned>(operandDimIdx)));
+    }
+  }
+
+  // For each loop dimension that is used by multiple inputs, promote the grid
+  // size associated with that loop dimension to the maximum across those
```
jdesousa-TT: minimum

sohaibnadeemTT: fixed

lib/Dialect/D2M/Transforms/GridSelection.cpp
```diff
@@ -438,15 +545,23 @@ analyzeOperandsAndComputeGrids(d2m::GenericOp genericOp,
...
     auto [optimalGrid, isVirtualGrid] =
-        computeOptimalGrid(operandType, physShape, targetSquareGridShape);
+        computeOptimalGrid(operandType, physShape, targetGridShape);
 
     optimalOperandGrids.push_back(optimalGrid);
+    isOptimalOperandGridVirtual.push_back(isVirtualGrid);
+  }
+
+  optimalOperandGrids =
+      normalizeOperandGridsForGeneric(genericOp, optimalOperandGrids);
```
sohaibnadeemTT, line 560: **_Change no. 2_**: grids constraints must be moved earlier and when we update view_layout/stream_layout/empty ops, the operand layouts are aligned according to the constrained grids (see layoutWithOptimalGrid, we pass optimalGrid instead). This ensures that the operands for matmul get aligned equally to respect dim constraints. To highlight why this problem occurs when using non-square grids, consider the following example on BH which has a 10x13 compute grid:

Before computePhysicalShape:
Input 0 layout: 1x1x20x26x!ttcore.tile<32x32, bf16>
Input 1 layout: 1x1x26x26x!ttcore.tile<32x32, bf16>

computePhysicalShape aligns both inputs to a 10x13 grid (since they exceed grid threshold) so when view_layout/stream_layout/empty ops get updated, we get the following new layouts:
Input 0 layout: 10x13x2x2x!ttcore.tile<32x32, bf16>
Input 1 layout: 10x13x3x2x!ttcore.tile<32x32, bf16> (gets aligned)

which is problematic since now the inputs dims don't even align to be able to do matmul (input 0 is 20*26 while input 1 is 30*26). This is why we need to apply the constraints early on right after computing the optimal grids, and then use the optimal (constrained) grids to do alignment and get the new layouts.

lib/Dialect/D2M/Transforms/GridSelection.cpp
```diff
@@ -255,39 +251,37 @@ shouldImplementAsVirtualGrid(RankedTensorType tensorType,
...
 
 static ttcore::MetalLayoutAttr layoutWithOptimalGrid(
     ttcore::MetalLayoutAttr oldLayout, ArrayRef<int64_t> targetGridShape,
-    ArrayRef<int64_t> targetSquareGridShape, ArrayRef<int64_t> optimalGrid,
-    bool isVirtualGrid, OpBuilder &builder) {
+    ArrayRef<int64_t> optimalGrid, bool isVirtualGrid, OpBuilder &builder) {
   auto collapsedIntervals = oldLayout.getCollapsedIntervals();
 
   llvm::SmallVector<int64_t> newDimAlignments =
       ttcore::MetalLayoutAttr::computeGridAwareDimAlignments(
```
sohaibnadeemTT, line 269: Related to comment above on change no. 2, we need to use optimal (constrained) grids to do out alignment and get the new layout.

lib/Dialect/D2M/Transforms/GridSelection.cpp
```diff
@@ -411,6 +403,121 @@ struct EmptyUpdateInfo {
...
+    for (auto [operandIndex, operandDimIdx] : entries) {
+      TT_assertv(operandDimIdx < normalizedOperandGrids[operandIndex].size(),
+                 "operand dim index out of bounds on adjusted operand grids");
+      minFactor = std::min(minFactor,
+                           normalizedOperandGrids[operandIndex][operandDimIdx]);
+    }
+    for (auto [operandIndex, operandDimIdx] : entries) {
+      TT_assertv(operandDimIdx < normalizedOperandGrids[operandIndex].size(),
+                 "operand dim index out of bounds on adjusted operand grids");
+      normalizedOperandGrids[operandIndex][operandDimIdx] = minFactor;
```
sohaibnadeemTT, line 472: **_Change no. 1_**: Use min instead of max here since we need to ensure we respect grid constraints (e.g. on BH, if one operand has grid width 13 and the other has grid height 10, then we have to use 10 since BH has only 10 rows)

sohaibnadeemTT (commented):
The PR is based off of Jacob's PR (https://github.com/tenstorrent/tt-mlir/pull/6287) for 1D matmul support (so will rebase once that goes in) and adds only the following 2 significant changes (the rest is refactoring to use targetGridShape instead of targetSquareGridShape in Grid Selection):

PR #6430, opened 2025-12-23
Adding support for `xla.sdy.FuncResultSharding` function

ajakovljevicTT:
Adding support for `xla.sdy.FuncResultSharding`, which is produced as a result of maving sharded torch tesnors to host on `tt-xla`. This PR adds their conversion into `sdy.sharding_constraint`, which than get get properly ingested by our downstream passes.

### DIFF
--- a/include/ttmlir/Dialect/StableHLO/Utils/GSPMDUtils.h
+++ b/include/ttmlir/Dialect/StableHLO/Utils/GSPMDUtils.h
@@ -24,8 +24,6 @@ inline constexpr llvm::StringRef kSPMDFullToShardShapeCallTargetName =
     "SPMDFullToShardShape";
 inline constexpr llvm::StringRef kSPMDShardToFullShapeCallTargetName =
     "SPMDShardToFullShape";
-inline constexpr llvm::StringRef kXlaSdyFuncResultShardingCallTargetName =
-    "xla.sdy.FuncResultSharding";
 inline constexpr llvm::StringRef kXlaShardingAttr = "mhlo.sharding";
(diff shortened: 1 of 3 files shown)


lib/Dialect/StableHLO/Utils/ShardyUtils.cpp
sshonTT, line 881: Can't we reuse this function for `convertFuncResultShardingToConstraint`? It seems that it works similarly

include/ttmlir/Dialect/StableHLO/Utils/GSPMDUtils.h
```diff
@@ -24,6 +24,8 @@ inline constexpr llvm::StringRef kSPMDFullToShardShapeCallTargetName =
     "SPMDFullToShardShape";
 inline constexpr llvm::StringRef kSPMDShardToFullShapeCallTargetName =
     "SPMDShardToFullShape";
+inline constexpr llvm::StringRef kXlaSdyFuncResultShardingCallTargetName =
```
hshahTT: Could we move this to the ShardyUtils.h file? Since we don't really support GSPMD anymore we should also look into slowly deprecating this file.

lib/Dialect/StableHLO/Utils/ShardyUtils.cpp
ajakovljevicTT: Done. Had to change the function a bit and add some special cases for the `FuncResultSharding`. Let me know if you agree with the changes!

sshonTT (approved):

PR #6434, opened 2025-12-23
[consteval] make const-eval cache global

### LINKED ISSUE #3888: Share Consteval Cache Between Binaries

For generative models, we want to share some graph constants between different compiled binaries where they represent the same weights.

For example in Llama, we have different graphs for prefill and decode, as they require different input shapes (bulk token proccessing vs single token processing).

However, many of the input weights between these binaries are the same, and get constevalled to identical consteval graphs within each binary.

From FE, we reuse the same tensor wrapper (runtime tensor) representing the input graph constant (by preventing deallocation and reusing them between prefill & decode). However, we'd like to also reuse the consteval graphs generated from those input tensors between prefill and decode.

Based on discussion with @vwellsTT @jazpurTT @AleksKnezevic on June 20:
 
- Changes to the consteval tensor cache schema to partially hash on the consteval graph itself, so identical consteval graphs could be shared
- Binding the tensor cache to a device rather than to a binary, so we don't need to detatch the cache from the prefill binary and insert it into the decode binary

pilkicTT:
Currently, consteval cache (i.e. `TensorCache`) is created per flatbuffer binary. It turns out that there are cases (LLMs) where we have multiple identical const-eval graphs present in multiple binaries (modules), e.g. prefill & decode. For models that barely fit on a single chip (llama-8b, qwen-8b, etc.) this causes us to hit OOM; since the cached results for the identical const-eval graphs are stored two times, one in the prefill's cache and the other in the decode's cache.

This change modifies the const-eval scheme such that it is possible to have one unique (global) cache - reducing the memory footprint.

To enable this, we need a way to identify identical const-eval graphs in the runtime. After investigating some options, I've chosen to go with hashing the IR with `sha256` and storing that hash value in the flatbuffer binary for each const-eval function.

NOTE: trying to identify mathematically equivalent IRs is not part of the scope of this change - here we only care for identical IRs.

The hash is computed on the dumped IR string, which is created in the following way:
- the function name (e.g. "main_const_eval_123") is temporarily replaced with "anonymous_const_eval" so that it doesn't affect the computed hash
- the IR is printed using `useLocalScope()` and `printGenericOpForm()`.

To help detect false positives (semantically different functions having the same hash), we could implement a runtime debug check which recomputes output tensors each time and verifies that the same tensors are in the cache - however, I would prefer to do it as a follow up if needed. Having false positives seems unlikely, when operating under the assumption that all the info needed for execution is contained within ops and their attributes, which in turn all get dumped as IR text.

Once we have the hash, then we can use it as part of the key in the cache map in runtime. The cache key contains of:
- id of the device on which we executed the const-eval function
- function hash
- vector of tensor versions - which uniquely identify which input tensors were used to compute the cached result.

In runtime, the `TensorCache` class is now `GlobalTensorCache` which is a singleton. It is moved into `ttnn` namespace, since it is designed and used only in `ttnn` backend.

A callback mechanism ties cache entry lifetimes to the lifetimes of its input tensors, removing the need for explicit cache management in most cases. We were having issues with the previous design since the cache was tied to the binary; so if you wanted to cache binaries for reuse, you would need to manually clean up its cache.

Unit tests and end-to-end python test are added.

Closes #3888


runtime/include/tt/runtime/detail/ttnn/types/global_tensor_cache.h
```diff
@@ -0,0 +1,104 @@
...
+    return deviceId == other.deviceId && funcHash == other.funcHash &&
+           inputVersions == other.inputVersions;
+  }
+};
+
+std::string to_string(const CacheKey &key);
+
+struct CacheKeyHasher {
+  std::size_t operator()(const CacheKey &k) const {
+    // TODO: Optimize hash function if needed
```
vmilosevic: :warning: **google\-readability\-todo** :warning:
missing username/bug in TODO
```suggestion
    // TODO(nsmith): Optimize hash function if needed
```

vmilosevic (commented):

lib/Target/TTNN/TTNNToFlatbuffer.cpp
```diff
@@ -2197,9 +2200,63 @@ createDeallocateOp(FlatbufferObjectCache &cache, DeallocateOp op) {
   return ::tt::target::ttnn::CreateDeallocateOp(*cache.fbb, in, force);
 }
 
+// Computes SHA256 hashes for all const-eval functions in the module and stores
```
sdjordjevicTT, line 2203: It might be more effective to perform this at the end of the TTNN backend pipeline in a simple pass. This pass would calculate the hashes and embed the hash information into a function attribute. Then, in TTNNToFlatbuffer, we can simply use that attribute to embed it into a flatbuffer. This approach would allow us to reuse this information in the TTNN emitter path, enabling the application of this optimization as well.

pilkicTT: i would propose to leave this for later, when implementing this mechanism fully for emitter paths... since the bigger problem than sharing this piece of code is implementing runtime const-eval caching in emited code...

also, i am not sure that making a pass and having an attribute is a good approach, since you always need to make sure that the hash is computed on the final version of the graph... so, i think we should do it when serializing (either to flatbuffer or to code)... in that case we should just have this as a common method for computing hashes...

sdjordjevicTT: I agree, it makes sense.

acolicTT (approved):

lib/Target/TTNN/TTNNToFlatbuffer.cpp
```diff
@@ -2197,9 +2200,63 @@ createDeallocateOp(FlatbufferObjectCache &cache, DeallocateOp op) {
...
+// them in a map. These hashes will be stored in the flatbuffer and later used
+// during runtime to identify completely equal const-eval functions.
+//
+// We have cases where the multiple modules contain identical const-eval
+// functions (e.g. in LLMs, prefill and decode share many const-eval functions).
+//
+// NOTE: the goal for this hashing is not to identify functions which are
+// mathematically equivalent, but the completely identical ones.
+llvm::StringMap<std::string>
+computeConstEvalFuncHashesSHA256(mlir::ModuleOp module) {
```
azecevic000: Nit: rename `module` to `moduleOp` to avoid name collision with contextual keyword from newer standards

azecevicTT (approved):

lib/Target/TTNN/TTNNToFlatbuffer.cpp
pilkicTT: renamed it throughout this file

runtime/include/tt/runtime/detail/ttnn/types/global_tensor_cache.h
```diff
@@ -0,0 +1,112 @@
...
+#include <shared_mutex>
+#include <string>
+#include <unordered_map>
+#include <vector>
+
+// Forward declaration for friend class
+namespace ttsl {
+template <typename T>
+class Indestructible;
+}
```
vmilosevic: :warning: **llvm\-namespace\-comment** :warning:
namespace `` ttsl `` not terminated with a closing comment
```suggestion
} // namespace ttsl
```

vmilosevic (commented):

PR #6435, opened 2025-12-23
Add kernel dumping/loading support to builder tests

wenbinlyuTT:
### Problem description
Builder no longer stores the manually generated golden inputs & outputs and so ttrt doesn't have access to these special inputs & goldens anymore.
Example: the `test/python/golden/test_metal_matmul.py::test_matmul_single_core_8otpc` test manually restricts the input range to be (0.0001, 0.999).
But if we run the generated flatbuffer again with `ttrt run --print-input-output-tensors`, negative values are in the inputs.
This affects some debug workflow using ttrt (combine special input patterns and kernel modifications to debug PCC issues).

### What's changed
Add kernel dumping & loading cmdline args to pytest so the workflow remains supported.

### DIFF
--- a/test/python/golden/conftest.py
+++ b/test/python/golden/conftest.py
@@ -210,6 +210,59 @@ def pytest_addoption(parser):
         action="store_true",
         help="disable putting dispatch on ethernet cores - place it on worker cores instead; necessary on blackhole",
     )
+    parser.addoption(
+        "--dump-kernels",
+        action="store_true",
+        help="Dump kernels to disk as they are being executed",
+    )
+    parser.addoption(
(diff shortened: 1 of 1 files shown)


jgrimTT:
@tapspatel I think this should be guarded by the compile_and_execute functions as an argument `store_goldens_in_flatbuffer: bool = False`, but this functionality is definitely needed while ttrt is still a tool

wenbinlyuTT:
@tapspatel @jgrimTT I have repurposed this pr to port the kernel dumping & loading cmdline args to pytest.

jgrimTT (approved):
LGTM!

PR #6442, opened 2025-12-25
Adding support for overriding compute kernel config's math_fidelity and fp32_dest_acc_en

### LINKED ISSUE #6159: Need matmul/linear compute_config (HiFi4 + fp32_dest_acc_en=True) for red model accurracy
TLDR; Can we plz make ComputeKernelConfig controllable and change defaults for increasing accurracy of LLM red models?

**Problem:**

We have at least half a dozen red LLM models which have low PCC and benefit immediately from setting these flags for extra precision in [runtime/lib/ttnn/operations/matmul/matmul.cpp](https://github.com/tenstorrent/tt-mlir/blob/2cfaa5cf3828b7197915c30bbf32a3548259c89f/runtime/lib/ttnn/operations/matmul/matmul.cpp) like in testing commit https://github.com/tenstorrent/tt-mlir/commit/2cfaa5cf3828b7197915c30bbf32a3548259c89f which gives PCC increase from `0.94 -> 0.98 in qwen2.5`

```
    auto config = ::ttnn::WormholeComputeKernelConfig();
    config.fp32_dest_acc_en = true;
    config.math_fidelity = MathFidelity::HiFi4;
```

A few more examples (before vs after PCC) I had on hand easily generated from `f5c3e8e6f` (Nov29) commit in tt-xla (there are more beyond this):

```
python ~/scripts/compare_pcc.py group_of_qwen_phi_llms_no_flags.xml group_of_qwen_phi_llms_hifi4_fp32_dest_acc_en.xml
test_name                                                                                                      pcc_before  pcc_after  pcc_change
test_all_models_torch[phi2/causal_lm/pytorch-microsoft/phi-2-single_device-full-inference]                       0.928117   0.956141    0.028024
test_all_models_torch[phi3/causal_lm/pytorch-microsoft/Phi-3-mini-128k-instruct-single_device-full-inference]    0.951510   0.973153    0.021643
test_all_models_torch[phi3/causal_lm/pytorch-microsoft/Phi-3-mini-4k-instruct-single_device-full-inference]      0.916351   0.971760    0.055408
test_all_models_torch[qwen_2_5/causal_lm/pytorch-0_5b_instruct-single_device-full-inference]                     0.984860   0.993960    0.009100
test_all_models_torch[qwen_2_5/causal_lm/pytorch-1_5b_instruct-single_device-full-inference]                     0.942016   0.980478    0.038461
test_all_models_torch[qwen_2_5/causal_lm/pytorch-3b_instruct-single_device-full-inference]                       0.916324   0.928521    0.012197
```

I pre-emptively ran a bunch of perf tests (all llms) locally yesterday (Using tt-xla d1d64565 (12-01) and tt-forge 4b24a92 (12-02)) with the above hack to `matmul.cpp` for these flags, and see on average around 1% samples/sec performance impact with this change if it were to be made default:

Image

**Request:**

From talking to @odjuricicTT and @AleksKnezevic we figure it makes sense to make these the default values for now to reap the accuracy benefits with minimal perf impact, but have it optionally overriden using `CompilerConfig` flags which could be controlled from models if needed, somewhat like https://github.com/tenstorrent/tt-mlir/pull/5878/ changes for BFP8 weights (`CompilerConfig` flag and it's tt-xla counterpart change) and https://github.com/tenstorrent/tt-mlir/pull/3434 which introduced `ComputeKernelConfig` attribute for `Conv2D` usage.

The one challenge there is that matmul op in ttnn (around here, and elsewhere https://github.com/tenstorrent/tt-metal/blob/main/ttnn/cpp/ttnn/operations/matmul/device/matmul_op.cpp#L1435-L1441) has somewhat smart internal defaults to the `WormholeComputeKernelConfig` fields when a `WormholeComputeKernelConfig` is not provided (tt-mlir currently provides none, std::nullopt), so if we just use the default initialized version of that struct and set a single field like `fp32_dest_acc_en` we get downgraded to `LoFi`, so some kind of smarter setting of the fields (similar to how matmuld does it internally) would be needed to not "do the wrong thing" when using `CompilerConfig` to only set `fp32_dest_acc_en=False` for example.

**Maybe Short Term:**

If we're open to just making these the new defaults and following up with `CompilerConfig` + IR + controllability changes as next step it works for me becuase it instantly helps/unblocks half dozen RED models.

dgolubovicTT:
Adding support for overriding compute kernel config's math_fidelity and fp32_dest_acc_en for all TTNN operations that expose TTNN_DeviceComputeKernelConfig in TTNN dialect.

### Ticket
Closes #6159

### Problem description
We need to enable global overrides for compute kernel config's math_fidelity and fp32_dest_acc_en for all TTNN operations that expose TTNN_DeviceComputeKernelConfig in TTNN dialect. Both of those recover accuracy in bunch of models - described in #6159

### What's changed

1. Defined the Interface TTNN_ComputeKernelConfigOpInterface:
   - Added TTNN_ComputeKernelConfigOpInterface with getComputeConfigAttr() and setComputeConfigAttr() methods
   - Added placeholder verification. Currently not needed since ComputeKernelConfig doesn't introduce changes in IR consistency (purely a runtime hint)
   - Added TTNN_ComputeKernelConfigOpInterface trait to TTNN operations that expose TTNN_ComputeKernelConfig
     - TTNN_MatmulOp, TTNN_LinearOP
     - TTNN_Conv2dOp, TTNN_Conv3dOp, TTNN_ConvTranspose2dOp - TTNN_PrepareConv2dBiasOp, TTNN_PrepareConv2dWeightsOp - TTNN_PrepareConvTranspose2dBiasOp, TTNN_PrepareConvTranspose2dWeightsOp - TTNN_RotaryEmbeddingOp, TTNN_RotaryEmbeddingLlamaOp - TTNN_MeanOp, TTNN_MinOp, TTNN_MaxOp, TTNN_SumOp - TTNN_RMSNormOp, TTNN_BatchNormInferenceOp, TTNN_BatchNormTrainingOp, TTNN_SoftmaxOp

   Note: all new ops that expose TTNN_ComputeKernelConfig should have TTNN_ComputeKernelConfigOpInterface interface added as a trait.
         Adding this interface to an op enables automatic propagation of compute config overrides in MLIR pass TTNNSetComputeKernelConfig via external polymorphis
2. Add TTNNSetComputeKernelConfig pass
   - Defined TTNNSetComputeKernelConfig pass in [Passes.td](http://passes.td/): - Sets default ComputeKernelConfig parameters for operations - Only applies values that are NOT already set (preserves existing compiler decisions) - Supports all ComputeKernelConfig parameters: math_fidelity, math_approx_mode, fp32_dest_acc_en, packer_l1_acc, dst_full_sync_en - Uses TTNNComputeKernelConfigOpInterface to find applicable operations
   - Added .mlir tests for this pass in test/ttmlir/Dialect/TTNN/set_compute_kernel_config/:
       - conv2d_set_math_fidelity.mlir: Setting math_fidelity parameter
       - conv2d_set_fp32_dest_acc.mlir: Setting fp32_dest_acc_en parameter - conv2d_preserve_existing.mlir: Preserving existing values when they are already set - conv2d_set_multiple_params.mlir: Setting multiple parameters together - conv2d_partial_override.mlir: Adding parameters to partial config
3. Integrate TTNNSetComputeKernelConfig pass to TTIRToTTNNBackendPipeline and test for compute kernel config overrides e2e in builder.
   - Added pipeline options to TTNNPipelines: - compute-cfg-math-fidelity: Set math fidelity (lofi, hifi2, hifi3, hifi4) - compute-cfg-fp32-dest-acc-en: Set fp32 destination accumulation (true, false)
   - Integrated TTNNSetComputeKernelConfig pass into TTNN pipeline
       - Pass runs after BFP8 weights conversion, before optimizer analysis passes
   - Added end-to-end builder test - test_compute_kernel_config.py: tests for each op group.
4. Support for prepare conv ops:
   - Prepare conv weight/bias operations inherit compute config from respective convolution op.
   - Add one conv2d end to end silicon test with optimizer on just to test prepare conv weight/bias ops logic: test/ttmlir/Silicon/TTNN/n150/optimizer/conv2d/conv2d_comp_cfg.mlir

5. Added  TTNN to EmitC Conversion support for ComputeKernelConfig

### DIFF
--- a/include/ttmlir/Conversion/TTNNToEmitC/EmitCConversion.h
+++ b/include/ttmlir/Conversion/TTNNToEmitC/EmitCConversion.h
@@ -890,6 +890,7 @@ struct EmitCTypeConverter<::ttnn::WormholeComputeKernelConfig> {
     rso << TypeNameV<::ttnn::WormholeComputeKernelConfig> << "{";
 
     bool first = true;
+
     // math_fidelity
     if (auto mathFidelity = attr.getMathFidelity()) {
       if (!first) {


include/ttmlir/Dialect/TTNN/Pipelines/TTNNPipelines.h
```diff
@@ -347,6 +347,32 @@ struct TTIRToTTNNBackendPipelineOptions
           "matrix multiplication and convolution operations to bfp8_b."),
       llvm::cl::init(false)};
 
+  // ComputeKernelConfig options
+  // Note: computeCfgMathFidelity default value (HiFi4) is not used unless
+  // computeCfgMathFidelitySet is true. The default exists only to satisfy
+  // Option<T>'s requirement for an initialized value.
+  Option<MathFidelity> computeCfgMathFidelity{
```
azecevic000, line 355: Didn't try this, so I'm not sure if it would work, but `Option<std::optional<MathFidelity>>` does make sense here, so we don't have to maintain `computeCfgMathFidelitySet`.

include/ttmlir/Dialect/TTNN/Interfaces/TTNNTensorSpecInterface.td
```diff
@@ -98,6 +98,42 @@ def TTNN_MemoryConfigOpInterface : OpInterface<"TTNNMemoryConfigOpInterface"> {
...
+    >,
+    InterfaceMethod<
+      /*desc=*/[{
+        Set the compute kernel config attribute.
+      }],
+      /*retTy=*/"void",
+      /*methodName=*/"setComputeConfigAttr",
+      /*args=*/(ins "DeviceComputeKernelConfigAttr": $computeConfigAttr),
+      /*methodBody=*/[{
+        $_op->setAttr("compute_config", computeConfigAttr);
```
azecevic000: Nit: I think that `$_op.setComputeConfigAttr(computeConfigAttr);` should also work here.

lib/Dialect/TTNN/Transforms/TTNNSetComputeKernelConfig.cpp
```diff
@@ -0,0 +1,92 @@
...
+
+class TTNNSetComputeKernelConfig
+    : public impl::TTNNSetComputeKernelConfigBase<TTNNSetComputeKernelConfig> {
+
+public:
+  using impl::TTNNSetComputeKernelConfigBase<
+      TTNNSetComputeKernelConfig>::TTNNSetComputeKernelConfigBase;
+
+  void runOnOperation() final {
+    ModuleOp module = getOperation();
```
azecevic000: Nit: rename `module` to `moduleOp` (it's a keyword in newer standards)

lib/Dialect/TTNN/Transforms/TTNNSetComputeKernelConfig.cpp
```diff
@@ -0,0 +1,92 @@
...
+    MLIRContext *context = &getContext();
+
+    std::optional<MathFidelity> mathFidelityOverride = mathFidelity;
+
+    // Walk through all operations in the module
+    module->walk([&](Operation *op) {
+      // Check if operation implements ComputeKernelConfigOpInterface
+      if (auto ckConfigOp = dyn_cast<TTNNComputeKernelConfigOpInterface>(op)) {
```
azecevic000: Nit: you can denest by an early exit, i.e.
```cpp
auto computeConfigOp = dyn_cast<TTNNComputeKernelConfigOpInterface>(op);
if (!computeConfigOp) {
  return;
}
// main logic
```

lib/Dialect/TTNN/Transforms/TTNNSetComputeKernelConfig.cpp
```diff
@@ -0,0 +1,92 @@
...
+        } else {
+          TTMLIR_DEBUG(
+              ttmlir::LogComponent::General,
+              "  Existing config before override: nullptr (no config set)");
+        }
+
+        // Start with existing config or create empty one
+        if (!config) {
+          config = DeviceComputeKernelConfigAttr::get(context);
+        }
```
azecevic000: Shouldn't this go into an `else` branch above?

include/ttmlir/Dialect/TTNN/Pipelines/TTNNPipelines.h
dgolubovicTT: Hey, I tried this one before I added `computeCfgMathFidelitySet` and It turns out llvm doesn't support optional out of the box... I made it work like this. I would like to merge this PR as is, and I can create Issue to investigate cleaner way to handle this option. How does this sound?

lib/Dialect/TTNN/Transforms/TTNNSetComputeKernelConfig.cpp
dgolubovicTT: Yeah, thanks

include/ttmlir/Dialect/TTNN/Pipelines/TTNNPipelines.h
azecevic000: Yeah, sounds good to me!

dgolubovicTT: Created https://github.com/tenstorrent/tt-mlir/issues/6468 as a follow up in the next days

mtopalovicTT: doesn't getNumOccurrences tell you if this option is set or not?

dgolubovicTT: This getNumOccurrences works only when the option is set in the command line. When option is set programmatically, getNumOccurrences doesn't register that...

dgolubovicTT:
@odjuricicTT @azecevicTT fyi, after talking to @pilkicTT and @nvukobratTT I decided to set `fp32_dest_acc_en` to true by default. Why:
- Almost no perf decrease currently, according to #6159 
- Reduces need for user to think about this knob - helps with out of the box generality.

include/ttmlir/Dialect/TTNN/Pipelines/TTNNPipelines.h
```diff
@@ -348,6 +348,32 @@ struct TTIRToTTNNDevicePipelineOptions
...
+      *this, "compute-cfg-math-fidelity",
```
odjuricicTT, line 356: Can we shorten this to just `math-fidelity`?

include/ttmlir/Dialect/TTNN/Pipelines/TTNNPipelines.h
```diff
@@ -348,6 +348,32 @@ struct TTIRToTTNNDevicePipelineOptions
...
+          clEnumValN(MathFidelity::HiFi4, "hifi4", "High fidelity 4")),
+      llvm::cl::init(MathFidelity::HiFi4)};
+
+  // Internal flag to track whether computeCfgMathFidelity was explicitly set.
+  // This enables distinguishing "unset" from "explicitly set to any value".
+  // Frontend propagation must use setComputeCfgMathFidelity() to set this flag.
+  bool computeCfgMathFidelitySet = false;
+
+  Option<bool> computeCfgFp32DestAccEn{
+      *this, "compute-cfg-fp32-dest-acc-en",
```
odjuricicTT, line 372: Same as above?

lib/Dialect/TTNN/Pipelines/TTNNPipelines.cpp
```diff
@@ -272,6 +272,30 @@ void createTTIRToTTNNDevicePipeline(
     if (options.experimentalBfp8Weights) {
       devicePm.addPass(createTTNNWeightBFP8Conversion());
     }
+
+    // Apply ComputeKernelConfig settings before analysis passes.
+    // This ensures that analysis passes see the configured values.
+    // Check if math fidelity is explicitly set (CLI or programmatic).
+    bool mathFidelityExplicitlySet =
+        options.computeCfgMathFidelitySet ||
+        (options.computeCfgMathFidelity.getNumOccurrences() > 0);
```
odjuricicTT, line 281: As we have the same pattern for optimization level option, i know that this does not work when passing these options via cpp struck like we do from tt-xla. In that case getNumOccurrences is always 0 as far as i remember.

I'm for finding a better solution for this, maybe optional is the way to go (as mentioned above)

test/ttmlir/Dialect/TTNN/set_compute_kernel_config/conv2d_set_math_fidelity.mlir
```diff
@@ -0,0 +1,21 @@
+// RUN: ttmlir-opt --ttnn-set-compute-kernel-config="math-fidelity=hifi4" %s | FileCheck %s
+
+// CHECK-LABEL: func @test_conv2d_without_compute_config
+func.func @test_conv2d_without_compute_config(%arg0: tensor<1x1x1024x64xbf16>, %arg1: tensor<64x64x3x3xbf16>, %device: !ttnn.device) -> tensor<1x1x900x64xbf16> {
+  // CHECK: ttnn.conv2d
+  // CHECK-SAME: compute_config = #ttnn.device_compute_kernel_config<math_fidelity = hifi4>
```
odjuricicTT: Can you add a check that fp32 accumulation is not set?

odjuricicTT (approved):

lib/Dialect/TTNN/Pipelines/TTNNPipelines.cpp
dgolubovicTT: True, but we have flag options.computeCfgMathFidelitySet that is set whenever we set math fidelity programatically via set method... We use this to propagate from tt-xla and it works...

include/ttmlir/Dialect/TTNN/Pipelines/TTNNPipelines.h
dgolubovicTT: I wanted to have this prefix compute-cfg to indicate all knobs that belong to compute config...

test/ttmlir/Dialect/TTNN/set_compute_kernel_config/conv2d_set_math_fidelity.mlir
dgolubovicTT: Since it is true by default, I can add the check that it is true here where we don't set it

include/ttmlir/Conversion/TTNNToEmitC/EmitCConversion.h
```diff
@@ -865,6 +872,87 @@ struct EmitCTypeConverter<::ttnn::QueueId> {
...
+      }
+      first = false;
+      rso << ".packer_l1_acc = " << (packerL1Acc.getValue() ? "true" : "false");
+    }
+    // dst_full_sync_en
+    if (auto dstFullSyncEn = attr.getDstFullSyncEn()) {
+      if (!first) {
+        rso << ", ";
+      }
+      first = false;
```
vmilosevic: :warning: **clang\-analyzer\-deadcode\.DeadStores** :warning:
Value stored to `` first `` is never read

include/ttmlir/Conversion/TTNNToEmitC/EmitCConversion.h
```diff
@@ -865,6 +872,86 @@ struct EmitCTypeConverter<::ttnn::QueueId> {
...
+
+  static std::string convert(ttnn::DeviceComputeKernelConfigAttr attr) {
+    std::string buf;
+    llvm::raw_string_ostream rso(buf);
+    rso << TypeNameV<::ttnn::WormholeComputeKernelConfig> << "{";
+
+    bool first = true;
+    // math_fidelity
+    if (auto mathFidelity = attr.getMathFidelity()) {
+      if (!first) {
```
svuckovicTT, line 896:
```suggestion
    bool first = true;

    // math_fidelity
    if (auto mathFidelity = attr.getMathFidelity()) {
      if (!first) {
```

Add empty rows between each conditional block

include/ttmlir/Conversion/TTNNToEmitC/EmitCConversion.h
```diff
@@ -865,6 +872,86 @@ struct EmitCTypeConverter<::ttnn::QueueId> {
...
+        rso << ", ";
+      }
+      first = false;
+      rso << ".math_fidelity = ::ttnn::MathFidelity::";
```
svuckovicTT, line 900: Similarly to this comment here, it'd be nice to make a separate converter for `ttnn::MathFidelity` type

https://github.com/tenstorrent/tt-mlir/pull/6462/changes#r2652495016

dgolubovicTT: Makes sense, I will add it to this follow up [issue](https://github.com/tenstorrent/tt-mlir/issues/6501)

PR #6445, opened 2025-12-26
[TTNN] Allocate config tensors in DRAM by default for conv_transpose2d

jserbedzijaTT:
### Problem description
In the general path, we want to store config tensors in DRAM instead of L1 to prevent OOM errors when using the `conv_transpose2d` operation.

### What's changed
Created `Conv2dConfig` and set the `ConfigTensorsInDram` flag by default for `conv_transpose2d`.

### Checklist

### DIFF
--- a/include/Constants.h
+++ b/include/Constants.h
@@ -9,9 +9,9 @@
 
 namespace tt::constants {
 
-// Default L1 small size to use for the ttnn runtime (16kb).
+// Default L1 small size to use for the ttnn runtime (64kb).
 // This reserves a region of L1 memory for L1_SMALL buffers used by convs.
-constexpr static std::size_t L1_SMALL_SIZE = 1 << 14;
+constexpr static std::size_t L1_SMALL_SIZE = 1 << 16;
 
 } // namespace tt::constants
 


include/Constants.h
```diff
@@ -9,9 +9,9 @@
 
 namespace tt::constants {
 
-// Default L1 small size to use for the ttnn runtime (64kb).
+// Default L1 small size to use for the ttnn runtime (16kb).
 // This reserves a region of L1 memory for L1_SMALL buffers used by convs.
-constexpr static std::size_t L1_SMALL_SIZE = 1 << 16;
+constexpr static std::size_t L1_SMALL_SIZE = 1 << 14;
```
azecevic000: Can you please grep the project for other places where this appears (`l1SmallSize` etc.) and change it to the same value.

jserbedzijaTT: I decided to fiddle with L1_SMALL_SIZE in a separate PR to simplify things

odjuricicTT (approved):
Nice! Test on vgg_unet from tt-forge perf benchmarks for a possible perf regression.

PR #6449, opened 2025-12-26
[EmitPy] Add ExpressionOp for inline Python code generation

### LINKED ISSUE #6394: [EmitPy] Add ExpressionOp to support expression inlining

Currently, the EmitPy dialect generates variable definition for almost every operation in the IR. This makes the Python output hard to follow.

Introduce ExpressionOp to EmitPy dialect. This will allow to fold boilerplate variable definitions directly into their point of use, leading to cleaner Python output.

Current emission:
```Python
var_268 = _CONST_EVAL_CACHE[0]
utils_constEvalFuncWrapper_0 = utils.constEvalFuncWrapper(const_0, util_create_list_213, var_268)
```

Desired emission:
```Python
utils_constEvalFuncWrapper_0 = utils.constEvalFuncWrapper(const_0, util_create_list_213, _CONST_EVAL_CACHE[0])
```

### Describe alternatives you've considered

_No response_

### Additional context

_No response_

sdjordjevicTT:
### Ticket
Closes #6394

### Problem description
The EmitPy dialect previously generated Python code with many intermediate variable assignments, even for simple operations like tuple element access. For example, accessing a tuple element would generate:
```Python
var_1 = 0
var_2 = tuple[var_1]
some_function(var2)
```
Instead of the more concise:
```Python
some_function(tuple[0])
```
Additionally, the EmitPy interfaces were located in the IR directory, which doesn't follow standard MLIR project organization patterns where interfaces should be in their own directory.

### What's changed
Introduces EmitPy_ExpressionOp to enable inline Python expression generation, producing cleaner code without intermediate variable assignments. Also reorganizes EmitPy interfaces into a separate directory following MLIR conventions.

The ExpressionOp allows operations to be composed into single-line Python expressions when translating, improving code readability.

Summary of all changes:
1. Added EmitPy_ExpressionOp and EmitPy_YieldOp: New operations that allow composing multiple EmitPy operations into inline Python expressions. The ExpressionOp contains a single-block region where operations can be chained together and yielded as a single expression.
2. Reorganized EmitPy interfaces: Moved EmitPyInterfaces from IR/ to a dedicated Interfaces/ directory, following MLIR conventions and improving code organization.
3. Enhanced Python translator: Updated TranslateToPython.cpp to handle ExpressionOp by building inline expression strings when the do_not_inline attribute is not set. This generates cleaner Python code without intermediate variables.
4. Updated GetTupleElementOpConversionPattern: Modified to use the new ExpressionOp, enabling inline subscript operations like tuple[0] instead of creating intermediate literal variables.
5. Added comprehensive tests: Included tests for various expression patterns including nested expressions, literals, subscripts, and complex DAG structures in both dialect and translation tests.

### Checklist

### DIFF
--- a/include/ttmlir/Dialect/EmitPy/IR/EmitPyOps.td
+++ b/include/ttmlir/Dialect/EmitPy/IR/EmitPyOps.td
@@ -107,15 +107,9 @@ def EmitPy_LiteralOp : EmitPy_Op<"literal", [Pure, PyExpressionInterface]> {
 
   let hasVerifier = 1;
   let assemblyFormat = "$value attr-dict `:` type($result)";
-
-  let extraClassDeclaration = [{
-    bool hasSideEffects() {
-      return false;
-    }
-  }];
 }
 
-def EmitPy_SubscriptOp : EmitPy_Op<"subscript", [PyExpressionInterface]> {
+def EmitPy_SubscriptOp : EmitPy_Op<"subscript", [Pure, PyExpressionInterface]> {
   let summary = "Subscript operation";
   let description = [{
     With the `emitpy.subscript` operation the subscript operator `[]` can be applied
@@ -137,12 +131,6 @@ def EmitPy_SubscriptOp : EmitPy_Op<"subscript", [PyExpressionInterface]> {
 
   // TODO (amilovanovic) : add custom verify method
   let assemblyFormat = "$value `[` $index `]` attr-dict `:` functional-type(operands, results)";
-
-  let extraClassDeclaration = [{
-    bool hasSideEffects() {
-      return false;
-    }
-  }];
 }
 
 def EmitPy_AssignOp : EmitPy_Op<"assign"> {
@@ -320,7 +308,6 @@ def EmitPy_YieldOp : EmitPy_Op<"yield", [Pure, ReturnLike, Terminator,
 
   let arguments = (ins AnyType:$result);
   let assemblyFormat = "$result attr-dict `:` type($result)";
-  let hasVerifier = 1;
 }
 
 def EmitPy_ExpressionOp : EmitPy_Op<"expression", [
@@ -329,9 +316,10 @@ def EmitPy_ExpressionOp : EmitPy_Op<"expression", [
   let summary = "Expression operation to inline Python code";
   let description = [{
     The `emitpy.expression` operation returns a single SSA value yielded by its
-    single-basic-block region. The operations inside the region must form a single
-    Def-Use tree. The expression's operands are mapped to block arguments and can
-    be used within the expression body.
+    single-basic-block region. The operations inside the region must form a DAG
+    (directed acyclic graph) of computations that produces a single result value.
+    The expression's operands are mapped to block arguments and can be used
+    within the expression body.
 
     When generating Python code, operations within the expression body are emitted
     inline without intermediate variable assignments, producing cleaner code.
--- a/include/ttmlir/Dialect/EmitPy/Interfaces/EmitPyInterfaces.h
+++ b/include/ttmlir/Dialect/EmitPy/Interfaces/EmitPyInterfaces.h
@@ -6,6 +6,7 @@
 #define TTMLIR_DIALECT_EMITPY_INTERFACES_EMITPYINTERFACES_H
 
 #include "mlir/IR/OpDefinition.h"
+#include "mlir/Interfaces/SideEffectInterfaces.h"
 
 #include "ttmlir/Dialect/EmitPy/Interfaces/EmitPyInterfaces.h.inc"
 
--- a/include/ttmlir/Dialect/EmitPy/Interfaces/EmitPyInterfaces.td
+++ b/include/ttmlir/Dialect/EmitPy/Interfaces/EmitPyInterfaces.td
@@ -24,7 +24,7 @@ def PyExpressionInterface : OpInterface<"PyExpressionInterface"> {
 
       By default operation is marked as having side effects.
 
-       ```c++
+      ```c++
       class ConcreteOp ... {
       public:
         bool hasSideEffects() {
@@ -36,7 +36,9 @@ def PyExpressionInterface : OpInterface<"PyExpressionInterface"> {
     }],
       "bool", "hasSideEffects", (ins), /*methodBody=*/[{}],
        /*defaultImplementation=*/[{
-        return true;
+        // Check if the operation has memory effects
+        // Operations without memory effects are considered pure
+        return !::mlir::isMemoryEffectFree($_op);
     }]>,
   ];
 }
--- a/lib/Dialect/EmitPy/IR/EmitPyOps.cpp
+++ b/lib/Dialect/EmitPy/IR/EmitPyOps.cpp
@@ -574,28 +574,6 @@ GetGlobalOp::verifySymbolUses(SymbolTableCollection &symbolTable) {
 }
 
 //===----------------------------------------------------------------------===//
-// YieldOp
-//===----------------------------------------------------------------------===//
-
-LogicalResult YieldOp::verify() {
-  auto *parentOp = getOperation()->getParentOp();
-  auto expressionOp = dyn_cast<ExpressionOp>(parentOp);
-  if (!expressionOp) {
-    return emitOpError("must be nested in an emitpy.expression");
-  }
-
-  Type yieldType = getResult().getType();
-  Type expressionType = expressionOp.getResult().getType();
-  if (yieldType != expressionType) {
-    return emitOpError("yielded type ")
-           << yieldType << " does not match expression result type "
-           << expressionType;
-  }
-
-  return success();
-}
-
-//===----------------------------------------------------------------------===//
 // ExpressionOp
 //===----------------------------------------------------------------------===//
 


include/ttmlir/Dialect/EmitPy/IR/EmitPyOps.h
```diff
@@ -7,10 +7,11 @@
 
 #include "ttmlir/Dialect/EmitPy/IR/EmitPy.h"
 #include "ttmlir/Dialect/EmitPy/IR/EmitPyAttrs.h"
-#include "ttmlir/Dialect/EmitPy/IR/EmitPyInterfaces.h"
+#include "ttmlir/Dialect/EmitPy/Interfaces/EmitPyInterfaces.h"
```
vmilosevic: :warning: **llvm\-include\-order** :warning:
```suggestion
#include "ttmlir/Dialect/EmitPy/IR/EmitPyTypes.h"
#include "ttmlir/Dialect/EmitPy/Interfaces/EmitPyInterfaces.h"
```

include/ttmlir/Dialect/EmitPy/IR/EmitPyOps.h
```diff
@@ -7,10 +7,11 @@
...
 #include "ttmlir/Dialect/EmitPy/IR/EmitPyTypes.h"
```
vmilosevic, line 10: :warning: **llvm\-include\-order** :warning:
```suggestion

```

lib/Target/Python/TranslateToPython.cpp
```diff
@@ -73,7 +73,7 @@ inline LogicalResult interleaveCommaWithError(const Container container,
 namespace {
 /// Emitter that uses dialect specific emitters to emit Python code.
 struct PythonEmitter {
-  explicit PythonEmitter(raw_ostream &os) : os(os) {
+  explicit PythonEmitter(raw_ostream &os) : os(os), inExpressionContext(false) {
```
vmilosevic: :warning: **cppcoreguidelines\-use\-default\-member\-init** :warning:
use default member initializer for `` inExpressionContext ``
```suggestion
  explicit PythonEmitter(raw_ostream &os) : os(os), {
```

lib/Target/Python/TranslateToPython.cpp
```diff
@@ -160,11 +166,18 @@ struct PythonEmitter {
 
   /// The set of names used in the current scope.
   std::stack<std::set<std::string>> usedNames;
+
+  /// Track whether we're currently emitting inside an expression
+  bool inExpressionContext;
```
vmilosevic: :warning: **cppcoreguidelines\-use\-default\-member\-init** :warning:
use default member initializer for `` inExpressionContext ``
```suggestion
  bool inExpressionContext{false};
```

lib/Target/Python/TranslateToPython.cpp
```diff
@@ -498,6 +648,18 @@ LogicalResult PythonEmitter::emitOperation(Operation &op) {
 }
 
 LogicalResult PythonEmitter::emitOperand(Value value, std::string name) {
+  // Check if this value is from an inline expression that has been cached
+  if (auto defOp = value.getDefiningOp()) {
```
vmilosevic: :warning: **llvm\-qualified\-auto** :warning:
```suggestion
  if (auto *defOp = value.getDefiningOp()) {
```

lib/Dialect/EmitPy/Interfaces/EmitPyInterfaces.cpp
```diff
@@ -0,0 +1,10 @@
+// SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#include "ttmlir/Dialect/EmitPy/Interfaces/EmitPyInterfaces.h"
+
+namespace mlir::tt::emitpy {
+#define GET_OP_CLASSES
+#include "ttmlir/Dialect/EmitPy/IR/EmitPyInterfaces.cpp.inc"
```
vmilosevic: :warning: **clang\-diagnostic\-error** :warning:
`` ttmlir/Dialect/EmitPy/IR/EmitPyInterfaces.cpp.inc `` file not found

include/ttmlir/Dialect/EmitPy/Interfaces/EmitPyInterfaces.td
```diff
@@ -24,12 +24,14 @@ def PyExpressionInterface : OpInterface<"PyExpressionInterface"> {
...
-      # Example: Pure operations like literals, arithmetic
-      class ConcreteOp:
-          def hasPythonSideEffects(self):
-              # Override for pure operations like literals, arithmetic
-              return False
+       ```c++
+      class ConcreteOp ... {
+      public:
+        bool hasSideEffects() {
+          // That way we can override the default implementation.
```
svuckovicTT:
```suggestion
          // This way we can override the default implementation.
```

include/ttmlir/Dialect/EmitPy/IR/EmitPyOps.td
```diff
@@ -7,9 +7,10 @@
 
 include "EmitPyAttrs.td"
 include "EmitPyBase.td"
-include "EmitPyInterfaces.td"
+include "ttmlir/Dialect/EmitPy/Interfaces/EmitPyInterfaces.td"
```
svuckovicTT:
```suggestion
include "EmitPyInterfaces.td"
```

include/ttmlir/Dialect/EmitPy/IR/EmitPyOps.td
```diff
@@ -106,6 +107,12 @@ def EmitPy_LiteralOp : EmitPy_Op<"literal", [Pure, PyExpressionInterface]> {
 
   let hasVerifier = 1;
   let assemblyFormat = "$value attr-dict `:` type($result)";
+
+  let extraClassDeclaration = [{
+    bool hasSideEffects() {
+      return false;
+    }
+  }];
```
azecevic000: What's the purpose of this function, isn't `hasSideEffects() != hasTrait<Pure>()` always true?

azecevic000:
An idea that came to mind, I've opened an issue with the description https://github.com/tenstorrent/tt-mlir/issues/6538. LMK what you think about it, it's definitely something that should be done in a subsequent PR if we are going with it as this one is already a big change as-is.

include/ttmlir/Dialect/EmitPy/IR/EmitPyOps.td
sdjordjevicTT: Mostly copy/pasted implementation details from EmitC dialect. It should serve a different purpose than a pure trait. It is a part of a PyExpressionInterface::hasSideEffects(). The Pure trait is for MLIR optimization passes, while the hasSideEffects() method from PyExpressionInterface is used during Python code generation to determine if operations can be safely inlined in expressions. It overrides the default implementation from PyExpressionInterface (which returns true) to indicate that this literal operation has no side effects in the context of Python expression generation.

sdjordjevicTT:
> An idea that came to mind, I've opened an issue with the description #6538. LMK what you think about it, it's definitely something that should be done in a subsequent PR if we are going with it as this one is already a big change as-is.

Agree. I won't block this PR, but will follow up with this issue after merging this one.

include/ttmlir/Dialect/EmitPy/IR/EmitPyOps.td
sdjordjevicTT: Actually, I need to keep "ttmlir/Dialect/EmitPy/Interfaces/EmitPyInterfaces.td" because EmitPyInterfaces.td isn't in the same dir as EmitPyOps.td file.

include/ttmlir/Dialect/EmitPy/Interfaces/EmitPyInterfaces.h
```diff
@@ -0,0 +1,12 @@
+// SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
```
nsmithtt: 2026

include/ttmlir/Dialect/EmitPy/IR/EmitPyOps.td
azecevic000: Discussed offline, I agree that it's not totally redundant. One thing I would change though is that instead of `hasSideEffects()` returning `true` by default, it can return `!hasTrait<Pure>()` by default, and just conditionally override non `Pure` ops (like `call_opaque` etc.).

include/ttmlir/Dialect/EmitPy/IR/EmitPyOps.td
```diff
@@ -286,4 +299,86 @@ def EmitPy_AssignGlobalOp : EmitPy_Op<"assign_global",
...
+}
+
+def EmitPy_ExpressionOp : EmitPy_Op<"expression", [
+    HasOnlyGraphRegion, IsolatedFromAbove,
+    SingleBlockImplicitTerminator<"YieldOp">]> {
+  let summary = "Expression operation to inline Python code";
+  let description = [{
+    The `emitpy.expression` operation returns a single SSA value yielded by its
+    single-basic-block region. The operations inside the region must form a single
+    Def-Use tree. The expression's operands are mapped to block arguments and can
```
azecevic000: `...must form a single Def-Use tree...`

I don't get this part; the given example actually has two different uses of the same def (`arg1`). The intention is probably to emphasize that there aren't divergent paths in the region, but a single basic block condition already guarantees that.

include/ttmlir/Dialect/EmitPy/IR/EmitPyOps.td
```diff
@@ -286,4 +299,86 @@ def EmitPy_AssignGlobalOp : EmitPy_Op<"assign_global",
...
+    be used within the expression body.
+
+    When generating Python code, operations within the expression body are emitted
+    inline without intermediate variable assignments, producing cleaner code.
+
+    The optional `do_not_inline` attribute can be used to force the translator to
+    emit the expression as separate statements rather than inline.
```
azecevic000, line 328: Is inlining the default behaviour? Might be beneficial to have a simple logic that always inlines if the number of uses is 1 (unless `do_not_inline` is specified), and not inline if number of uses > 1, because it could force expensive recalculation. Though in this case, `do_not_inline` would have to be `BoolAttr` so that we can force inlining even when number of uses is > 1, so I'm not sure if it's worth it at the moment, just noted it here to have that in mind if we ever observe undesirable behaviour in the wild in the future.

include/ttmlir/Dialect/EmitPy/Interfaces/EmitPyInterfaces.td
```diff
@@ -24,12 +24,14 @@ def PyExpressionInterface : OpInterface<"PyExpressionInterface"> {
...
-              # Override for pure operations like literals, arithmetic
-              return False
+       ```c++
+      class ConcreteOp ... {
+      public:
+        bool hasSideEffects() {
+          // This way we can override the default implementation.
+          return false;
+        }
+      };
```
azecevic000: I don't get this example, we won't emit this code neither in Python nor C++, we just use it for the ExpressionOp.

lib/Conversion/TTNNToEmitPy/TTNNToEmitPy.cpp
```diff
@@ -2365,17 +2365,40 @@ class GetTupleElementOpConversionPattern
...
+        this->getTypeConverter()->convertType(getTupleElementOp.getType());
+
+    // Create an expression op to inline the subscript operation
+    auto loc = getTupleElementOp->getLoc();
+    auto exprOp = rewriter.create<emitpy::ExpressionOp>(
+        loc, resultType, ValueRange{adaptor.getOperand()});
+
+    // Setup the expression body
+    {
+      OpBuilder::InsertionGuard guard(rewriter);
```
azecevic000, line 2379: This is nice!

lib/Dialect/EmitPy/IR/EmitPyOps.cpp
```diff
@@ -573,5 +573,201 @@ GetGlobalOp::verifySymbolUses(SymbolTableCollection &symbolTable) {
...
+    return emitOpError("must be nested in an emitpy.expression");
+  }
+
+  Type yieldType = getResult().getType();
+  Type expressionType = expressionOp.getResult().getType();
+  if (yieldType != expressionType) {
+    return emitOpError("yielded type ")
+           << yieldType << " does not match expression result type "
+           << expressionType;
+  }
```
azecevic000: I believe that the combination of `Terminator` and `ParentOneOf<["ExpressionOp"]>` traits defined in td should guarantee both of these conditions, so explicit verification should be redundant.

include/ttmlir/Dialect/EmitPy/IR/EmitPyOps.td
sdjordjevicTT: I agree, this is a great suggestion, will update the code.

include/ttmlir/Dialect/EmitPy/Interfaces/EmitPyInterfaces.td
sdjordjevicTT: This is a C++ concrete op class example with the interface applied. Not the generated code.

include/ttmlir/Dialect/EmitPy/IR/EmitPyOps.td
sdjordjevicTT: Agree, confused docs a bit, will update to:
The operations inside the region must form a DAG (directed acyclic graph) of computations that produces a single result value.

include/ttmlir/Dialect/EmitPy/IR/EmitPyOps.td
sdjordjevicTT: Yes, inlining is the default behavior. The current design keeps things simple and predictable. Let's revisit if we encounter performance issues in real-world usage.

lib/Dialect/EmitPy/IR/EmitPyOps.cpp
sdjordjevicTT: You are right, I will remove this!

PR #6450, opened 2025-12-26
[emitc] Add EmitC Support for chained TensorAccessorArgs

brnorris03:
### Ticket
N/A

### Problem description

tt-metal kernels extensively use chained `TensorAccessorArgs` to automatically manage compile-time and runtime argument offsets, e.g.:

```cpp
auto args_src = TensorAccessorArgs<3, 7>();
auto args_dst = TensorAccessorArgs<
    args_src.next_compile_time_args_offset(),
    args_src.next_common_runtime_args_offset()>();
...
```

The current TTKernel to EmitC conversion does not support this, forcing manual offset calculations that can be error-prone and do not take advantage of the tt-metal interface that computes these offsets automatically.

### What's changed

Extended `TTKernel.TensorAccessorArgsOp` with:
- Optional `prev_args` operand - references previous accessor for automatic offset chaining.
- Optional `cta_expr`/`crta_expr` string attributes - constexpr expressions that override chaining.
- When lowering to emitc: explicit attrs > chaining > literal constants.

Added verifier for:

1. Mutual Exclusivity: `prev_args` and `cta_base/crta_base` are mutually exclusive
   - When `prev_args` is present, `cta_base` and `crta_base` must NOT be provided
   - When `prev_args` is NOT present, both `cta_base` and `crta_base` MUST be provided

2. Constant Requirement: When not using expressions, bases must be constants
   - `cta_base` must be a constant when `cta_expr` is not provided
   - `crta_base` must be a constant when `crta_expr` is not provided

**Impact on existing behavior:**
- **Breaking change**: `TensorAccessorArgsOp` now emits C++ using `emitc.verbatim` instead of `emitc.call_opaque`.
  - Generated C++ code is functionally identical (same `TensorAccessorArgs<CTA, CRTA>()` calls).
  - Generated MLIR EmitC IR structure changes from anonymous SSA values to named variables
- New optional `prev_args`, `cta_expr`, `crta_expr` enable chaining patterns not previously expressible (non-breaking change).

tt-metal use case coverage:

| tt-metal Pattern | ttmlir Support | Example |
|------------------|----------------|---------|
| `TensorAccessorArgs<cta_base, crta_base>()` | `ttkernel.TensorAccessorArgs(%cta, %crta)` | Basic case with constant bases |
| `TensorAccessorArgs<prev.next_cta(), prev.next_crta()>()` | `ttkernel.TensorAccessorArgs(prev = %prev)` | Pure chaining from previous args |
| `TensorAccessorArgs<cta_base, crta_base>()` with expressions | `ttkernel.TensorAccessorArgs(%cta, %crta) cta_expr = "..." crta_expr = "..."` | Bases with C++ expression overrides |
| `TensorAccessorArgs<prev.next_cta()>(crta_expr)` | `ttkernel.TensorAccessorArgs(prev = %prev) crta_expr = "..."` | Chaining with runtime crta override |

**Testing:**
- 6 positive tests: basic literal offsets, pure chaining, expression attributes (`cta_expr`, `crta_expr`, both), chaining with expression override.
- 9 negative tests: parser error handling for syntax and type errors, non-constant operands.

### Checklist

### DIFF
--- a/python/TTKernelModule.cpp
+++ b/python/TTKernelModule.cpp
@@ -100,10 +100,22 @@ void populateTTKernelModule(nb::module_ &m) {
 
   tt_attribute_class<tt::ttkernel::ArgSpecAttr>(m, "ArgSpecAttr")
       .def_static("get",
-                  [](MlirContext ctx, std::vector<tt::ttkernel::ArgAttr> rtArgs,
-                     std::vector<tt::ttkernel::ArgAttr> ctArgs) {
-                    return wrap(tt::ttkernel::ArgSpecAttr::get(unwrap(ctx),
-                                                               rtArgs, ctArgs));
+                  [](MlirContext ctx, std::vector<MlirAttribute> rtArgs,
+                     std::vector<MlirAttribute> ctArgs) {
+                    std::vector<tt::ttkernel::ArgAttr> _rt_args, _ct_args;
+
+                    for (const auto &x : rtArgs) {
+                      _rt_args.emplace_back(
+                          mlir::cast<tt::ttkernel::ArgAttr>(unwrap(x)));
+                    }
+
+                    for (const auto &x : ctArgs) {
+                      _ct_args.emplace_back(
+                          mlir::cast<tt::ttkernel::ArgAttr>(unwrap(x)));
+                    }
+
+                    return wrap(tt::ttkernel::ArgSpecAttr::get(
+                        unwrap(ctx), _rt_args, _ct_args));
                   })
       .def_prop_ro_static("name",
                           [](nb::handle) {
--- a/test/python/test_ttkernel_argspec.py
+++ b/test/python/test_ttkernel_argspec.py
@@ -11,16 +11,13 @@ from ttmlir.dialects import ttkernel
 
 with Context() as ctx, Location.unknown():
     # Create ArgAttr instances for rt_args and ct_args.
-    # ArgAttr.get() returns MlirAttribute, downcast to typed ArgAttr.
-    rt_arg1 = ttkernel.ir.ArgAttr.maybe_downcast(
-        ttkernel.ir.ArgAttr.get(ctx._CAPIPtr, 0, 0, True)
-    )
-    rt_arg2 = ttkernel.ir.ArgAttr.maybe_downcast(
-        ttkernel.ir.ArgAttr.get(ctx._CAPIPtr, 0, 1, False)
-    )
-    ct_arg1 = ttkernel.ir.ArgAttr.maybe_downcast(
-        ttkernel.ir.ArgAttr.get(ctx._CAPIPtr, 1, 2, True)
-    )
+    # ArgAttr.get() returns MlirAttribute for MLIR interop.
+    rt_arg1_attr = ttkernel.ir.ArgAttr.get(ctx._CAPIPtr, 0, 0, True)
+    rt_arg2_attr = ttkernel.ir.ArgAttr.get(ctx._CAPIPtr, 0, 1, False)
+    ct_arg1_attr = ttkernel.ir.ArgAttr.get(ctx._CAPIPtr, 1, 2, True)
+
+    # Downcast to access typed properties.
+    rt_arg1 = ttkernel.ir.ArgAttr.maybe_downcast(rt_arg1_attr)
 
     # Verify ArgAttr properties work on downcast object.
     # CHECK: rt_arg1 operand_index: 0
@@ -28,8 +25,10 @@ with Context() as ctx, Location.unknown():
     # CHECK: rt_arg1 is_uniform: True
     print(f"rt_arg1 is_uniform: {rt_arg1.is_uniform}")
 
-    # ArgSpecAttr.get() returns MlirAttribute for MLIR interop.
-    spec_attr = ttkernel.ir.ArgSpecAttr.get(ctx._CAPIPtr, [rt_arg1, rt_arg2], [ct_arg1])
+    # ArgSpecAttr.get() accepts MlirAttribute objects and returns MlirAttribute.
+    spec_attr = ttkernel.ir.ArgSpecAttr.get(
+        ctx._CAPIPtr, [rt_arg1_attr, rt_arg2_attr], [ct_arg1_attr]
+    )
 
     # Downcast to access typed properties.
     spec = ttkernel.ir.ArgSpecAttr.maybe_downcast(spec_attr)


lib/Conversion/TTKernelToEmitC/TTKernelToEmitC.cpp
```diff
@@ -533,29 +533,58 @@ class TTKernelTensorAccessorArgsOpRewriter
...
+    os.flush();
+    std::string varName = "tensor_accessor_args_" + ssaName.substr(1);
+
+    // Build CTA/CRTA expression with priority: expr attr > chaining > literal.
+    auto buildArgExpr = [&](StringAttr exprAttr, Value baseValue,
+                            StringRef chainMethodName) -> std::string {
+      if (exprAttr) {
+        // Explicit constexpr string expression (overrides chaining).
+        return exprAttr.getValue().str();
+      } else if (op.getPrevArgs()) {
```
vmilosevic: :warning: **llvm\-else\-after\-return** :warning:
```suggestion
      } if (op.getPrevArgs()) {
```

zoecarver (approved):

python/TTKernelModule.cpp
```diff
@@ -86,9 +86,9 @@ void populateTTKernelModule(nb::module_ &m) {
           "get",
           [](MlirContext ctx, uint32_t argTypeValue, size_t operandIndex,
              bool isUniform = true) {
-            return wrap(tt::ttkernel::ArgAttr::get(
```
tapspatel, line 89: any reason we are not wrapping the type?

brnorris03: Is there a good reason to wrap it? As far as I understand it, with nanobind we don't need explicit wrapping because we don't have to go through the MLIR CAPI (for MlirAttribute), we get a concrete instance directly with `nb::class_<T>` for `tt_attribute_class<tt::ttkernel::ArgAttr>`.

brnorris03: If we always wrap, then also need `get` implementations to deal with the custom type explicitly, approximately:
```c++
diff --git a/python/TTKernelModule.cpp b/python/TTKernelModule.cpp
index b13725616..c937d03f7 100644
--- a/python/TTKernelModule.cpp
+++ b/python/TTKernelModule.cpp
@@ -86,9 +86,9 @@ void populateTTKernelModule(nb::module_ &m) {
           "get",
           [](MlirContext ctx, uint32_t argTypeValue, size_t operandIndex,
              bool isUniform = true) {
-            return tt::ttkernel::ArgAttr::get(
+            return wrap(tt::ttkernel::ArgAttr::get(
                 unwrap(ctx), static_cast<tt::ttkernel::ArgType>(argTypeValue),
-                operandIndex, isUniform);
+                operandIndex, isUniform));
           },
           nb::arg("ctx"), nb::arg("argTypeValue"), nb::arg("operandIndex"),
           nb::arg("isUniform") = true)
@@ -100,10 +100,20 @@ void populateTTKernelModule(nb::module_ &m) {
 
   tt_attribute_class<tt::ttkernel::ArgSpecAttr>(m, "ArgSpecAttr")
       .def_static("get",
-                  [](MlirContext ctx, std::vector<tt::ttkernel::ArgAttr> rtArgs,
-                     std::vector<tt::ttkernel::ArgAttr> ctArgs) {
-                    return wrap(tt::ttkernel::ArgSpecAttr::get(unwrap(ctx),
-                                                               rtArgs, ctArgs));
+                  [](MlirContext ctx, std::vector<MlirAttribute> rtArgs,
+                     std::vector<MlirAttribute> ctArgs) {
+                    std::vector<tt::ttkernel::ArgAttr> rtArgsUnwrapped;
+                    std::vector<tt::ttkernel::ArgAttr> ctArgsUnwrapped;
+                    for (const auto &arg : rtArgs) {
+                      rtArgsUnwrapped.push_back(
+                          mlir::cast<tt::ttkernel::ArgAttr>(unwrap(arg)));
+                    }
+                    for (const auto &arg : ctArgs) {
+                      ctArgsUnwrapped.push_back(
+                          mlir::cast<tt::ttkernel::ArgAttr>(unwrap(arg)));
+                    }
+                    return wrap(tt::ttkernel::ArgSpecAttr::get(
+                        unwrap(ctx), rtArgsUnwrapped, ctArgsUnwrapped));
                   })
       .def_prop_ro_static("name",
                           [](nb::handle) {
```

tapspatel: yeah, this is one of those things where I've heard differing opinions from folks on how to handle. My understanding is as follows

The CAPIs were introduced to support ABI (application boundary interface for C++) and RTTI. For ABI, it acts as a simple wrapper around python such that when changes upstream happen to the underlying C++ objects, we don't have to maintain and support the python side handling of such differences. For RTTI, consider the situation where you have 2 separate shared libraries that each expose the same MLIR C++ type via nano bind. see: https://nanobind.readthedocs.io/en/latest/faq.html?utm_source=chatgpt.com#how-can-i-avoid-conflicts-with-other-projects-using-nanobind. They list a bunch of concerns and how it could break both packages. There's also solutions to this listed, as well as defining your nb module as local with `nb::module_local())` or making opaque types with your shared types `NB_OPAQUE_TYPE`. You can probably guess it's hard to enforce this between different projects. By making the types opaque to python using the CAPI, python no longer has to worry about the types or maintaining a type registry for RTTI, and it forces the user by design to have to cast in C++ instead (see `maybe_downcast` in include/ttmlir/Bindings/Python/TTMLIRModule.h).

You can absolutely get away with not having to use the CAPIs but I think it's more of a fail-safe in case mlir significantly changes their architecture towards bindings or users are trying some finicky things with different libs (compiled with different versions of their compiler) etc. 

Two popular project we use significantly by google are stablehlo and shardy. See `stablehlo/integrations/python/StablehloModule.cpp` and `shardy/integrations/python/ir/sdy_module.cc`, They also follow the wrap/unwrap pattern for their custom types/attributes for the most part, but they use `mlir_attribute_subclass` whereas we have a slightly different nanobind class definition, but the idea of type erasure in python is still there.

tapspatel: For the custom type comment, I agree it adds some boilerplate code, but imo the explicit cast is by design - it keeps the onus out of having to maintain the type in nanobind and instead explicitly fail during the cast instead,

tapspatel: discussed offline, the requirements around tt-lang are the user will construct some custom attribute, and we probably don't want that to be type erased, since if they want to access any member variables from it, they will have to downcast in python, which is not pythonic.


Instead, we should follow a mechanism like so
```
extern "C" {
bool isSomeAttr(MlirAttribute attr) {
  return mlir::isa<SomeAttr>(unwrap(attr));
}

MlirAttribute someAttrGet(MlirContext ctx, int someValue) {
  return wrap(SomeAttr::get(unwrap(ctx), someValue));
}
}

mlir::python::nanobind_adaptors::mlir_attribute_subclass(
      m, "SomeAttr", isSomeAttr)
      .def_classmethod(
          "get",
          [](nb::object cls, int someValue,
             MlirContext ctx) {
            return cls(someAttrGet(ctx, someValue));
          })
     .def_classmethod(
          "from_attribute",
          [](nb::object cls, MlirAttribute attr) {
              if (!isSomeAttr(attr)) {
                  throw std::runtime_error("Attribute is not a SomeAttr");
              }
              return cls(attr);
          })
```

Where we can support both cases of returning the actual type + if we have a type erasured object, we can reconstruct the type aware object.

tapspatel: This is a wider discussion, I'm okay with PR as is and we can reevaluate next steps during our tech discussion.

pilkicTT (approved):

Issue #5247, opened 2025-10-07
Mish activation function


mvasiljevicTT:
Yolo4 uses mish activation function that is lowered through torch-xla as 
`x * tanh(where(x > C, x, log1p(e^x)))`

Since mish op is supported in tt-metal, it should be added in TTIR and TTNN dialects, and this pattern should be fused into it. It should also be added as fused activation in convolution.

mvasiljevicTT:
Op is added in TTIR and TTNN dialects with  [this](https://github.com/tenstorrent/tt-mlir/pull/5914/files) and [this](https://github.com/tenstorrent/tt-mlir/pull/5475) pr. TTIRFusing pattern still needed.

HalfBloodPrince010:
Hi @mvasiljevicTT thanks again for helping. Could you assign me this issue? And, could you provide some reference pointers for the above fusion task?

HalfBloodPrince010:
Hi @mvasiljevicTT , just checking in on this when you get a chance. Apologies, if you are busy. Thank you!

mvasiljevicTT:
Hi @HalfBloodPrince010!

Here you can get a bit more context about [yolov4 and mish activation](https://cenk-bircanoglu.medium.com/review-of-yolov4-architecture-f488ec32c1c4). There you can see that in yolov4 model mish is implemented as `x * torch.tanh(F.softplus(x))`.

Lowered through [tt-xla](https://github.com/tenstorrent/tt-xla), we get [yolov4_ttir.txt](https://github.com/user-attachments/files/24291887/yolov4_ttir.txt). Every time you see `"ttir.log1p"` op in this ir, it probably comes from one mish activation.
For better undestanding of the graph, you can visualize it with [tt-explorer](https://docs.tenstorrent.com/tt-mlir/tt-explorer/tt-explorer.html) where you could notice this pattern:


<img width="928" height="1250" alt="Image" src="https://github.com/user-attachments/assets/79e67cfe-8a7c-49e7-88af-66e55854668d" />


Your goal is for this to be rewritten as just one `mish` op. You would do that by making TTIRFusing rewrite pattern, [like the one for hardsigmoid activation](https://github.com/tenstorrent/tt-mlir/blob/72778d4412f719cf95a43cdcd529f0d242983b8b/lib/Dialect/TTIR/Transforms/TTIRFusing.cpp#L3843) for example.

### Repro and testing
[...]

PR #6451, opened 2025-12-26
Add fusion pattern for MishOp

HalfBloodPrince010:
### Ticket
[Link to Github Issue](https://github.com/tenstorrent/tt-mlir/issues/5247)

### Problem description

Yolo4 uses mish activation function that is lowered through torch-xla as
```x * tanh(where(x > C, x, log1p(e^x)))```
here, input to the tanh op is a numerically
stable variant of Softplus function

### Motivation

IR has a list of Ops like log1p, tanh, multiply
which is basically a Mish Activation function
for which the Op has already been added.
We can fuse these different ops into a single
mish op.

### What's changed

- Added a fusing pattern for mish activation function

### Testing

Added a test to verify the changes, both with and
without typecasting for multiple operands.

#### Before
before_mish

#### After
after_mish_fusion_pattern

```
llvm-lit -sv ./test/ttmlir/Dialect/TTIR/fusing/mish_fusing.mlir 

Testing Time: 0.05s

Total Discovered Tests: 1
  Passed: 1 (100.00%)
```

### Checklist

### DIFF
--- a/lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
+++ b/lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
@@ -571,6 +571,97 @@ public:
   }
 };
 
+std::optional<mlir::Value> matchNumericallyStableSoftplus(mlir::Value op) {
+  // common utility to match the numerically stable softplus
+  // NumericallyStableSoftplus = where[cond, x, ln(1 + e^x)]
+  WhereOp whereOp = op.getDefiningOp<WhereOp>();
+  if (!whereOp || !whereOp->hasOneUse()) {
+    return std::nullopt;
+  }
+  auto softplusInput = whereOp.getSecond();
+
+  // [cond, x, ln(1 + e^x)]
+  Log1pOp log1pOp = whereOp.getThird().getDefiningOp<Log1pOp>();
+  if (!log1pOp || !log1pOp->hasOneUse()) {
+    return std::nullopt;
+  }
+
+  GreaterThanOp gtOp = whereOp.getFirst().getDefiningOp<GreaterThanOp>();
+  if (!gtOp) {
+    return std::nullopt;
+  }
+
+  if (gtOp.getLhs() != softplusInput) {
+    return std::nullopt;
+  }
+
+  ExpOp expOp = log1pOp.getInput().getDefiningOp<ExpOp>();
+  if (!expOp || !expOp->hasOneUse()) {
+    return std::nullopt;
+  }
+
+  if (softplusInput != expOp.getInput()) {
+    return std::nullopt;
+  }
+
+  return softplusInput;
+}
+
+// Mish Fusion Pattern
+class MishFusingPattern : public mlir::OpRewritePattern<MultiplyOp> {
+  using mlir::OpRewritePattern<MultiplyOp>::OpRewritePattern;
+
+public:
+  // Match Pattern -  mish(x) = x * tanh(Softplus(x))
+  // where, Softplus(x) = x > C? x : ln(1 + e^x) (numerically stable variant of
+  // softplus(x)) Because for very large x, (1 + e^x) ~ e^x ln(e^x) = x
+  mlir::LogicalResult
+  matchAndRewrite(MultiplyOp multiplyOp,
+                  mlir::PatternRewriter &rewriter) const final {
+    mlir::Value lhs = utils::lookThrough<TypecastOp>(multiplyOp.getLhs());
+    mlir::Value rhs = utils::lookThrough<TypecastOp>(multiplyOp.getRhs());
+
+    // Match multiply(x, tanh(softplus(x))) and
+    // multiply(tanh(softplus(x)), x)
+    mlir::Value originalInput;
+    TanhOp tanhOp = nullptr;
+    if (auto lhsTanhOp = lhs.getDefiningOp<TanhOp>()) {
+      tanhOp = lhsTanhOp;
+      originalInput = rhs;
+    } else if (auto rhsTanhOp = rhs.getDefiningOp<TanhOp>()) {
+      tanhOp = rhsTanhOp;
+      originalInput = lhs;
+    }
+
+    if (!tanhOp || !tanhOp->hasOneUse()) {
+      return mlir::failure();
+    }
+
+    // Check if tanh's input is softplus
+    auto softplusInput = matchNumericallyStableSoftplus(tanhOp.getInput());
+    if (!softplusInput || *softplusInput != originalInput) {
+      return mlir::failure();
+    }
+
+    auto inputType = originalInput.getType();
+    auto outputType = multiplyOp.getResult().getType();
+    auto mishOp =
+        rewriter.create<MishOp>(multiplyOp.getLoc(), inputType, originalInput);
+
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by HalfBloodPrince010** [comment]
Also, tested on the yolov4 ttir

ttnn_post_mish_fusion

**Comment by HalfBloodPrince010** [comment]
[Not related to current PR]

But just checking), should this be casted to MultiplyOp's `output_type` instead?

https://github.com/tenstorrent/tt-mlir/blob/72778d4412f719cf95a43cdcd529f0d242983b8b/lib/Dialect/TTIR/Transforms/TTIRFusing.cpp#L566

The tests are not catching it because in IR we explicitly typecast back to the result. 

Let me know if I have missed something.

File: lib/Dialect/TTIR/Transforms/TTIRFusing.cpp [hunk 1]
```diff
@@ -573,6 +573,97 @@ class SiluFusionPattern : public mlir::OpRewritePattern<MultiplyOp> {
...
+  auto softplusInput = whereOp.getSecond();
+
+  // [cond, x, ln(1 + e^x)]
+  Log1pOp log1pOp = whereOp.getThird().getDefiningOp<Log1pOp>();
+  if (!log1pOp || !log1pOp->hasOneUse()) {
+    return std::nullopt;
+  }
+
+  GreaterThanOp gtOp = whereOp.getFirst().getDefiningOp<GreaterThanOp>();
+  if (!gtOp) {
```
mtopalovicTT, line 590: Check for one use

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
```diff
@@ -573,6 +573,97 @@ class SiluFusionPattern : public mlir::OpRewritePattern<MultiplyOp> {
...
+    return std::nullopt;
+  }
+
+  if (gtOp.getLhs() != softplusInput) {
```
mtopalovicTT, line 594: check that gtOp.lhs and gtOp.rhs have one use

test/ttmlir/Dialect/TTIR/fusing/mish_fusing.mlir
```diff
@@ -0,0 +1,35 @@
...
+module {
+    // CHECK-LABEL: func.func @mish_fusing
+    func.func @mish_fusing(%arg0:tensor<1x32x480x640xbf16>) -> tensor<1x32x480x640xbf16> {
+        %1 = "ttir.constant"() <{value = dense<2.000000e+01> : tensor<1x32x480x640xbf16>}> : () -> tensor<1x32x480x640xbf16>
+        %2 = "ttir.gt"(%arg0, %1) : (tensor<1x32x480x640xbf16>, tensor<1x32x480x640xbf16>) -> tensor<1x32x480x640xi1>
+        %3 = "ttir.exp"(%arg0) : (tensor<1x32x480x640xbf16>) -> tensor<1x32x480x640xbf16>
+        %4 = "ttir.log1p"(%3) : (tensor<1x32x480x640xbf16>) -> tensor<1x32x480x640xbf16>
+        %5 = "ttir.where"(%2, %arg0, %4) : (tensor<1x32x480x640xi1>, tensor<1x32x480x640xbf16>, tensor<1x32x480x640xbf16>) -> tensor<1x32x480x640xbf16>
+        %6 = "ttir.tanh"(%5) : (tensor<1x32x480x640xbf16>) -> tensor<1x32x480x640xbf16>
+        // CHECK-NOT: "ttir.multiply"
```
mtopalovicTT, line 13: I'm not sure if we need dialect test anymore. It would be better if you can create builder test which can test fusing and also compare against torch golden. You have one example inside `test/python/golden/ttir_ops/fusing/test_rope_fusing.py`. Just have in mind that `test_rope_fusing` has two tests which are skipped for reasons where shouldn't matter in your case.

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
HalfBloodPrince010: Thank you @mtopalovicTT for the comments. To confirm my understanding: the gt op (e.g., `%cond = ttir.gt %x, %C`) may have multiple users without impacting the fusion, since it doesn't have side effects right?

test/ttmlir/Dialect/TTIR/fusing/mish_fusing.mlir
HalfBloodPrince010: Added a builder-based test for Mish fusion, following the same pattern as the RoPE fusion tests. The existing lit test is kept for additional coverage, but can be removed if we decide to consolidate on builder tests, similar to other recent changes. Let me know what do you think.

mvasiljevicTT:
> [Not related to current PR]
> 
> But just checking), should this be casted to MultiplyOp's `output_type` instead?
> 
> https://github.com/tenstorrent/tt-mlir/blob/72778d4412f719cf95a43cdcd529f0d242983b8b/lib/Dialect/TTIR/Transforms/TTIRFusing.cpp#L566
> 
> The tests are not catching it because in IR we explicitly typecast back to the result.
> 
> Let me know if I have missed something.

Nice catch, thanks for pointing this out!

HalfBloodPrince010:
Thank you @mvasiljevicTT. I ran pre-commit and rebased the changes, could you help run the workflow?

HalfBloodPrince010:
Hi @mvasiljevicTT Good day, could you help get these changes merged? Thank you.

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
mtopalovicTT: You are right. Do we have to check hasOneUse for any of the ops in this PR?

HalfBloodPrince010:
@mvasiljevicTT hmm strange, tests were passing previously, now its failing.

I guess I need to update the artifacts directory and func signature like in this commit: https://github.com/tenstorrent/tt-mlir/commit/e2510c468806d8148242be1197d9004dd5c21313#diff-7547a2e36953cd9be0d5082673f347bca95fa07d2278b9be39f9c12d7c86044e

This change was made last week, will update and raise a revision.

mvasiljevicTT:
> @mvasiljevicTT hmm strange, tests were passing previously, now its failing.
> 
> I guess I need to update the artifacts directory and func signature like in this commit: [e2510c4#diff-7547a2e36953cd9be0d5082673f347bca95fa07d2278b9be39f9c12d7c86044e](https://github.com/tenstorrent/tt-mlir/commit/e2510c468806d8148242be1197d9004dd5c21313#diff-7547a2e36953cd9be0d5082673f347bca95fa07d2278b9be39f9c12d7c86044e)
> 
> This change was made last week, will update and raise a revision.

That's right, that should be the only change you need

HalfBloodPrince010:
@mvasiljevicTT sounds good, I updated the test, could you help run workflow and merge the changes? thank you!

HalfBloodPrince010:
Tests are passing, but was cancelled

```
test/python/golden/ttir_ops/fusing/test_mish_fusing.py::test_mish_fusion[ttnn-f32-shape0] PASSED [ 11%]
```

PR #6453, opened 2025-12-27
Add chain merging support (A→B, 3-way merge) for DFShardingPolicy

rpavlovicTT:
## Summary
Add chain merging capability to DFShardingPolicy to keep tensors in L1 across chain boundaries:

- **A→B Merge**: When Chain A's output feeds Chain B's first operand, keep Chain A's output in L1 while Chain B executes
- **3-Way Merge**: Support merging two chains (A and C) into a single consumer chain B at a join operation
- **Validation**: Validate that merged chains can execute within L1 constraints using `additionalL1Usage`
- **Helper functions**: `findAllMergeCandidates`, `validateChainBWithMergedInput`, `validateThreeWayMergeJoinOp`, `applyChainMerges`

This enables better L1 utilization by avoiding unnecessary DRAM spills between consecutive chains.

**Stacked on**: #6452

### DIFF
--- a/lib/Dialect/TTNN/Analysis/DFShardingPolicy.cpp
+++ b/lib/Dialect/TTNN/Analysis/DFShardingPolicy.cpp
@@ -12,6 +12,7 @@
 #include "ttmlir/Dialect/TTNN/IR/TTNNOps.h"
 #include "ttmlir/Dialect/TTNN/IR/TTNNOpsAttrs.h"
 #include "ttmlir/Dialect/TTNN/Utils/Utils.h"
+#include "ttmlir/Dialect/TTNN/Validation/OpConstraintValidation.h"
 #include "ttmlir/Scheduler/Scheduler.h"
 #include "ttmlir/Support/Logger.h"
 #include "ttmlir/Utils.h"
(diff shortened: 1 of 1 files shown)


lib/Dialect/TTNN/Analysis/DFShardingPolicy.cpp
```diff
@@ -20,6 +21,1079 @@
...
+    const std::vector<L1Reservation> &l1Reservations,
+    const llvm::DenseMap<Operation *, TTNNLayoutAttr> &l1ResidentsLayoutMap =
+        kEmptyL1ResidentsLayoutMap) {
+
+  auto joinOpPosIt = schedulePositionMap.find(joinOp);
+  if (joinOpPosIt == schedulePositionMap.end()) {
+    return false;
+  }
+  int64_t joinOpPos = joinOpPosIt->second;
+
```
odjuricicTT, line 252: Shoud we fail the merge if join op is not scheduled right after end of chain A?

rpavlovicTT: Not really because this check is passed we additionally check `validateScheduleRangeWithReservation` (line 859). Potentially I can remove this `validateChainBWithMergedInput` in favour of the other more generic check. However, I did not want to make more changes once code converged in an acceptable state.

odjuricicTT (approved):

PR #6456, opened 2025-12-27
Fix ShardSolver matmul activation handling and elementwise binary core shrink

rpavlovicTT:
## Summary
Fix two ShardSolver issues that cause hangs/crashes during L1 sharding:

### 1. Matmul/Linear activation handling
- Add `shouldUseIgnorePhysicalLayout()` helper to detect when matmul/linear ops have fused activations
- When activation is present, we need full layout because the internal unary op cannot handle partial memory configs (crashes in validate_shard_spec)
- **TODO(tt-metal#34500)**: Remove this workaround once tt-metal handles partial memory configs in fused activations. 
### 2. Elementwise binary core shrink rejection
- Reject configs where elementwise binary ops (Add, Multiply, Minimum) would shrink the core count from sharded input to sharded output
- Implicit resharding in these ops causes hangs
- **Workaround for**: github.com/tenstorrent/tt-metal/issues/34765

### 3. Fix preprocessFirstOp layout comparison
- Use original layout for comparison instead of the ignorePhysicalLayout-modified version

### DIFF
--- a/include/ttmlir/OpModel/TTNN/TTNNOpModel.h
+++ b/include/ttmlir/OpModel/TTNN/TTNNOpModel.h
@@ -993,7 +993,8 @@ struct OpModel<MatmulOp> {
       ttcore::GridAttr deviceGrid, llvm::ArrayRef<int64_t> inputShapeA,
       TTNNLayoutAttr inputLayoutA, llvm::ArrayRef<int64_t> inputShapeB,
       TTNNLayoutAttr inputLayoutB, TTNNLayoutAttr outputLayout, bool transposeA,
-      bool transposeB);
+      bool transposeB,
+      std::optional<llvm::StringRef> activation = std::nullopt);
 
   static llvm::Expected<size_t>
   getOpRuntime(llvm::ArrayRef<int64_t> inputShapeA, TTNNLayoutAttr inputLayoutA,
--- a/lib/Dialect/TTNN/Interfaces/TTNNOpModelInterface.cpp
+++ b/lib/Dialect/TTNN/Interfaces/TTNNOpModelInterface.cpp
@@ -2948,7 +2948,7 @@ MatmulOp::getOpConstraints(const std::vector<TTNNLayoutAttr> &inputs,
   return opConstraintsCache().getOrCompute(
       op_model::OpModel<MatmulOp>::getOpConstraints, *this, deviceGrid,
       inputShapeA, inputs[0], inputShapeB, inputs[1], opConfig.outputLayout,
-      getTransposeA(), getTransposeB());
+      getTransposeA(), getTransposeB(), getActivation());
 }
 
 llvm::Expected<size_t>
--- a/lib/OpModel/TTNN/TTNNOpModel.cpp
+++ b/lib/OpModel/TTNN/TTNNOpModel.cpp
@@ -4171,7 +4171,7 @@ llvm::Expected<OpConstraints> OpModel<MatmulOp>::getOpConstraints(
     ttcore::GridAttr deviceGrid, llvm::ArrayRef<int64_t> inputShapeA,
     TTNNLayoutAttr inputLayoutA, llvm::ArrayRef<int64_t> inputShapeB,
     TTNNLayoutAttr inputLayoutB, TTNNLayoutAttr outputLayout, bool transposeA,
-    bool transposeB) {
+    bool transposeB, std::optional<llvm::StringRef> activation) {
 #ifdef TTMLIR_ENABLE_OPMODEL
   ::tt::tt_metal::distributed::MeshDevice *device =
       SingletonDeviceContext::getInstance().getDevice();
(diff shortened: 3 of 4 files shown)


lib/Dialect/TTNN/Analysis/ShardSolver.cpp
```diff
@@ -38,6 +38,22 @@
 
 namespace mlir::tt::ttnn {
 
+// Helper to check if matmul/linear op should use IgnorePhysicalLayout.
+// When activation is present, we need full layout because the internal unary op
+// for activation cannot handle partial memory configs (crashes in
+// validate_shard_spec).
+// TODO(tt-metal#34500): Remove activation check once tt-metal handles partial
```
odjuricicTT, line 45: Can we have an xfail / expects_death op model constraint test for this that will fail when the tt-metail issue fix is uplifted?

lib/Dialect/TTNN/Analysis/ShardSolver.cpp
```diff
@@ -38,6 +38,22 @@
...
+// memory configs in fused activations.
+static bool shouldUseIgnorePhysicalLayout(Operation *op) {
+  if (auto matmulOp = mlir::dyn_cast<ttnn::MatmulOp>(op)) {
+    return !matmulOp.getActivation().has_value();
+  }
+  if (auto linearOp = mlir::dyn_cast<ttnn::LinearOp>(op)) {
+    return !linearOp.getActivation().has_value();
```
odjuricicTT, line 52: Should this be in part of validation / constraint API?

lib/Dialect/TTNN/Analysis/ShardSolver.cpp
```diff
@@ -197,6 +212,25 @@ bool ShardSolver::resolveStep() {
           for (std::size_t i = 0; i < results.size(); ++i) {
             const auto &result = results[i];
             if (result.isSuccess()) {
+              // For elementwise binary ops with sharded input, reject configs
+              // that shrink the core count. Implicit resharding in these ops
+              // causes hangs. See: github.com/tenstorrent/tt-metal/issues/34765
+              if (inputLayout.hasShardedL1TensorMemoryLayout() &&
```
odjuricicTT, line 218: Same as above? Can we add a test to fail when this is resolved?

lib/Dialect/TTNN/Analysis/ShardSolver.cpp
```diff
@@ -197,6 +212,25 @@ bool ShardSolver::resolveStep() {
...
+                  result.actualOutputLayout.hasShardedL1TensorMemoryLayout() &&
+                  llvm::isa<ttnn::AddOp, ttnn::MultiplyOp, ttnn::MinimumOp>(
+                      consumerOp)) {
+                int64_t inputCores = inputLayout.getGrid().getGridVolume();
+                int64_t outputCores =
+                    result.actualOutputLayout.getGrid().getGridVolume();
+                if (outputCores < inputCores) {
```
odjuricicTT, line 225: Same as above, should this be in constraint / validation code?

lib/Dialect/TTNN/Analysis/ShardSolver.cpp
rpavlovicTT: Not really, as it affects how we generate test configs for shard solver. I think there may be bugs if we wait until validation to control it. Anyway, this is fixed in tt-metal and uplift should come soon, so it will be easy to remove this code and test it again.

lib/Dialect/TTNN/Analysis/ShardSolver.cpp
rpavlovicTT: Added a test. I had to make amendment in TTNN OpModel to handle activation parameter. This is also done in https://github.com/tenstorrent/tt-mlir/pull/6457/, anyway later I will handle the conflict.

lib/Dialect/TTNN/Analysis/ShardSolver.cpp
rpavlovicTT: Not easy to do since this is a hang scenario. Skipping for now.

lib/Dialect/TTNN/Analysis/ShardSolver.cpp
rpavlovicTT: It could be in validation lib (at the time when we know backend tensor layout), but I don't see it being much cleaner. I'd rather spoil shard solver code than validation lib.

odjuricicTT: Yes but anything else that uses validation lib can hit this same issue. We are moving towards it being used in multiple other passes.

odjuricicTT (approved):

lib/Dialect/TTNN/Analysis/ShardSolver.cpp
rpavlovicTT: I agree, but only shard solver actually shards tensors right now, so it is safe. Also, we don't observe hang during graph capture, but only in runtime...

PR #6460, opened 2025-12-29
Refactoring TTNNBuilder

jgrimTT:
### Ticket
Link to Github Issue

### Problem description
TTNNBuilder still does op creation the old way through `op_proxy`

### What's changed
Implemented updated op handling for 61/63 ops
Added updated mappings
Added snippets

### Checklist

### DIFF
--- a/tools/builder/ttnn/ttnn_builder.py
+++ b/tools/builder/ttnn/ttnn_builder.py
@@ -615,7 +615,7 @@ class TTNNBuilder(Builder):
 
         op_map_dictionary = {}
         op_map_dictionary[old_op.result] = new_op_result
-        return erf_module, erf_builder
+        return new_op, op_map_dictionary
 
     ############### ttnn.ErfcOp ###############
 
--- a/test/python/golden/test_parse_split_ops.py
+++ b/test/python/golden/test_parse_split_ops.py
@@ -80,4 +80,3 @@ def test_stablehlo_parsing_splitting_ops(mlir_snippet, request, device):
 def test_ttnn_parsing_splitting_ops(mlir_snippet, request, device):
     mlir_ir_string = ttnn_mlir_snippets[mlir_snippet]
     mlir_module, builder = load_mlir_file(mlir_ir_string, target="ttnn")
-    split_modules = split_mlir_file(mlir_module, builder)


ndrakulicTT:
This looks good overall.

One thought: many of these ops fall into a few repeatable paterns, so we could group them and auto-generate the `@parse`, `@split`, and `@tag` functions instead of defining all three per op. Groups like unary (single input), binary (two inputs), ternary, and variadic (e.g., concat) would cover most ops, with a small “special” bucket for truly custom behavior.

With grouping plus `inspect.signature` for matching kwargs/args, most ops could be handled automatically, and in many cases we’d only need to define the custom golden behavior instead of the full set of functions.

This would reduce boilerplate, improve coverage, and automatically surface missing builder support when new ops appear in `ttmlir.dialect.ttnn`.

tapspatel:
@ndrakulicTT totally agreed. We initially had that in builder where we had groupings of ops to reduce boilerplate. However, more requirements were being added and the structure was too rigid to keep the consolidated functions. So we ended up breaking everything apart. 

Our strategy is 2 folds, first we are going as deep as possible to get all current and future features "working", so we know what the structure of the code will look like. Alongside that, keep adding as many ops as possible, once that is done (only a few features are left to experiment and add in), we will refactor all this boilerplate to make it more generic.

tools/builder/base/builder_apis.py
```diff
@@ -847,6 +847,9 @@ def compile_ttnn_to_flatbuffer(
         module, builder = build_module(
             fn,
             "ttnn",
+            module_dump=module_dump,
+            output_root=output_root,
+            base=test_base + "5",
```
tapspatel: what is this 5

tools/builder/ttnn/ttnn_builder.py
```diff
@@ -343,2390 +343,7515 @@ def decorated_func(*inputs):
...
+                    rhs = inputs[1]
+                    result = old_op.result.type
+
+                    new_op = ttnn_op(result, lhs, rhs, loc=old_op.location)
+                    new_op_result = new_op.result
+
+                    if not self._disable_golden_check:
+                        op_golden_function = get_golden_function(ttnn_op)
+                        input0 = self._get_golden_tensor(old_op.lhs)
+                        input1 =
```
tapspatel: dim is not default, so we should force user to provide it

tapspatel: remove default arg

tools/golden/mapping.py
```diff
@@ -4231,13 +4200,516 @@ def clamp(min_val, x, max_val):
 ################ TTNN Op Golden Functions ###############
 
 
+def ttnn_abs_golden(
```
tapspatel: duplicate ttnn_abs_golden function

tools/golden/mapping.py
```diff
@@ -86,7 +86,15 @@ class GoldenMapTensor:
...
 
-    def __init__(self, shard_map: Dict[int, torch.Tensor], mesh_shape: Tuple[int, int]):
+    def __init__(
+        self,
+        shard_map: Dict[int, torch.Tensor] | "GoldenMapTensor",
+        mesh_shape: Tuple[int, int],
+    ):
+        # Allow constructing from another GoldenMapTensor for convenience
+        if isinstance(shard_map, GoldenMapTensor):
+            shard_map = shard_map.shard_map
```
tapspatel: we should be using __eq__ for this, although I see it's already been overridden to something else which is not good. Let's remove this and use __eq__ and fix wherever its being used to a different approach

tools/golden/mapping.py
```diff
@@ -357,6 +365,13 @@ def unpack_mlir_attr(attr):
     if isinstance(attr, DenseElementsAttr):
         array = np.array(attr)
         return array
+    if isinstance(attr, Attribute):
```
tapspatel: this won't work because Attribute is the parent class for any type of Attr, we should avoid string based parsing at all costs

See 
```
convolution_layout = ttir.ir.ConvolutionLayoutAttr.maybe_downcast(
        convolution_layout_attr
    )
```

for how to downcast the attribute if you have the parent instance.

tapspatel (changes requested):
feedback

tools/builder/ttnn/ttnn_builder.py
```diff
@@ -343,2390 +343,7515 @@ def decorated_func(*inputs):
...
+                    rhs = inputs[1]
+                    result = old_op.result.type
+
+                    new_op = ttnn_op(result, lhs, rhs, loc=old_op.location)
+                    new_op_result = new_op.result
+
+                    if not self._disable_golden_check:
+                        op_golden_function = get_golden_function(ttnn_op)
+                        input0 = self._get_golden_tensor(old_op.lhs)
+
```
tapspatel: unnecessary comment

tapspatel, line 3508: pow header 
############### ttnn.PowTensorOp ###############

tapspatel: some places you created ranked tensor type, others `create_ttnn_tensor`

Issue #6465, opened 2025-12-29
Refactor CPU-hoisted function deduplication semantics


dmilinkovicTT:
Currently, in `HoistCPUOps` pass, if a CPU-hoisted function declaration with a corresponding name already exists, its creation will be skipped, as it is assumed that the function implementation is essentially encoded in the function's name (e.g. `hoisted_shlo_add_32x32_32x32_func`). 

This approach only makes sense when hoisting single ops at a time (although the approach is still fragile). Since hoisting multiple ops at a time has been enabled in #6041, this approach should be revisited and refactored.

Issue #6448, opened 2025-12-26
error: 'func.call' op incorrect number of operands for callee


ndrakulicTT:
Repro in tt-xla:
`pytest -svv tests/runner/test_models.py::test_all_models_torch[vit/pytorch-vit_b_16-single_device-full-training]`

gives:
```
error: 'func.call' op incorrect number of operands for callee
```

This is last MLIR that was generated:
[shlo_compiler_1766760121971.txt](https://github.com/user-attachments/files/24348093/shlo_compiler_1766760121971.txt)

This affects 13 models:
```
pytorch_mamba_mamba-1.4b-hf_nlp_causal_lm_huggingface
pytorch_mamba_mamba-370m-hf_nlp_causal_lm_huggingface
pytorch_mamba_mamba-790m-hf_nlp_causal_lm_huggingface
pytorch_swin_swin_b_cv_image_cls_torchvision
pytorch_swin_swin_s_cv_image_cls_torchvision
pytorch_swin_swin_t_cv_image_cls_torchvision
pytorch_swin_swin_v2_b_cv_image_cls_torchvision
pytorch_swin_swin_v2_s_cv_image_cls_torchvision
pytorch_swin_swin_v2_t_cv_image_cls_torchvision
pytorch_vit_vit_b_16_cv_image_cls_torchvision
pytorch_vit_vit_b_32_cv_image_cls_torchvision
pytorch_vit_vit_l_16_cv_image_cls_torchvision
pytorch_vit_vit_l_32_cv_image_cls_torchvision
```

dmilinkovicTT:
Seems to be CPU-hoisting related, taking a look

dmilinkovicTT:
Confirmed it's a bug in CPU-hoisting pass, the fix will land on Monday

PR #6466, opened 2025-12-29
Fix arguments mismatch issue in CPU-hoisted operations

dmilinkovicTT:
### Ticket
#6448

**Note: this PR is a somewhat-dirty hotfix. A proper fix should be delivered as part of #6465.**

### Problem description
During CPU hoisting, when the same value is used multiple times as an argument of an op, current implementation will skip it, because deduplication of the input arguments is being performed. This issue has been introduced as part of #6041. 

Deduplication makes sense when hoisting multiple ops at the same time (e.g. for const-eval CPU-hoisting), but it shouldn't be performed during single-op hoisting.

Additionally (but unrelated to the issue), element types and 0D-tensor input aren't encoded in the CPU-hoisted function names, which makes the IR hard to follow and debug.

### Example

📝 **Given**: For the `dynamic_update_slice` op in the following snippet:
```
%base_2d = stablehlo.constant dense<[[0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0], [0, 0, 0, 0]]> : tensor<4x4xi32>
%update_2d = stablehlo.constant dense<[[9, 9], [9, 9]]> : tensor<2x2xi32>
%start_2d = stablehlo.constant dense<1> : tensor<i32>
    
    
%result_2d = stablehlo.dynamic_update_slice %base_2d, %update_2d, %start_2d, %start_2d
      : (tensor<4x4xi32>, tensor<2x2xi32>, tensor<i32>, tensor<i32>) -> tensor<4x4xi32>
```

✅ **Expected behavior**: The following signature for the CPU-hoisted function is used:
```
func.func @hoisted...(tensor<4xi32>, tensor<2xi32>, tensor<i32>, tensor<i32>) -> tensor<4xi32>
--------------------------------------------------- ^^^^^^^^^^^^^^^^^^^^^^^^ -----------------
```

❌ **Actual behavior**: `%start_2d` argument gets deduplicated, and the following signature is used:
```
func.func private @hoisted...(tensor<4xi32>, tensor<2xi32>, tensor<i32>) -> tensor<4xi32>
----------------------------------------------------------- ^^^^^^^^^^^ -----------------
```

### What's changed
- Made the input arguments deduplication optional.
- Improved hoisted functions naming.

### Checklist

### DIFF
--- a/lib/Dialect/TTIR/Transforms/HoistCPUOps.cpp
+++ b/lib/Dialect/TTIR/Transforms/HoistCPUOps.cpp
@@ -67,7 +67,12 @@ llvm::SmallString<64> generateHoistedFuncName(const OpsVectorType &ops) {
               mlir::dyn_cast<mlir::RankedTensorType>(operand.getType())) {
         uniqueName += "_";
         llvm::raw_svector_ostream os(uniqueName);
-        llvm::interleave(tensorType.getShape(), os, "x");
+        // Emit shape and element type. Example: 2x3xf32.
+        if (!tensorType.getShape().empty()) {
+          llvm::interleave(tensorType.getShape(), os, "x");
+          os << "x";
+        }
+        tensorType.getElementType().print(os);
       }
     }
   }
@@ -79,7 +84,9 @@ llvm::SmallString<64> generateHoistedFuncName(const OpsVectorType &ops) {
 }
 
 // Helper function to collect input arguments to a set of operations.
-static ValuesVectorType collectInputArguments(const OpsVectorType &operations) {
+// If shouldDeduplicate is true, only unique arguments are collected.
+static ValuesVectorType collectInputArguments(const OpsVectorType &operations,
+                                              bool shouldDeduplicate) {
   ValuesVectorType inputArguments;
 
   for (auto *op : operations) {
@@ -90,10 +97,11 @@ static ValuesVectorType collectInputArguments(const OpsVectorType &operations) {
         continue;
       }
 
-      // Insert the argument if not already present.
-      if (!llvm::is_contained(inputArguments, operand)) {
-        inputArguments.push_back(operand);
+      if (shouldDeduplicate && llvm::is_contained(inputArguments, operand)) {
+        continue;
       }
+
+      inputArguments.push_back(operand);
     }
   }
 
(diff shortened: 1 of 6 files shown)


mtopalovicTT:
> Deduplication makes sense when hoisting multiple ops at the same time (e.g. for const-eval CPU-hoisting), but it shouldn't be performed during single-op hoisting.

What is the difference between deduplicating when there are multiple hoisted ops and deduplicating when we only have single hoisted op?

dmilinkovicTT:
> What is the difference between deduplicating when there are multiple hoisted ops and deduplicating when we only have single hoisted op?

@mtopalovicTT When hoisting a single op:
```
%res = ttir.add %a, %a
```

If deduplication is performed, the generated CPU-hoisted function would receive only one argument.

However, if deduplication is skipped, the generated CPU-hoisted function would be more general, and could facilitate hoisted `ttir.add` ops with different arguments. This is a small optimization that prevents having multiple similar-ish CPU-hoisted functions.

On the other hand, when hoisting the following ops of a const-eval subgraph:

```
%res0 = ttir.add %a, %b
%res1 = ttir.add %a, %c
return %res0, %res1
```

It doesn't make much sense to have `%a` duplicated and to have the generated CPU-hoisted function receive 4 arguments (2 for first `add`, and 2 for the second), because this CPU-hoisted function will be used only in the context of this specific const-eval subgraph - no reuse is expected.

Of course, if we ever end up wanting to avoid deduplication on sets of multiple ops, we could do that with the proposed implementation, since this behavior is controllable though a flag in the descriptor.

mtopalovicTT:
> However, if deduplication is skipped, the generated CPU-hoisted function would be more general, and could facilitate hoisted `ttir.add` ops with different arguments. This is a small optimization that prevents having multiple similar-ish CPU-hoisted functions.

Can you give me an ir example where hoisting would fail today on main branch when we are hoisting single op which reuses input argument? Again I don't see why would we add new logic to handle cases when we don't want to deduplicate, it should work the same regardless of number of ops we are hoisting unless there is something I'm missing.

> This is a small optimization that prevents having multiple similar-ish CPU-hoisted functions.

I'm not sure good an optimization this really is. If you ask me we should generate hoisted functions by hashing the body of "to be generated" hoisted function and we should use that hash value to check if hoisted function already exist. Today we are just using main function name + shapes of arguments which is just wrong.

dmilinkovicTT:
@mtopalovicTT 

> Can you give me an ir example where hoisting would fail today on main branch when we are hoisting single op which reuses input argument?

```
%51 = stablehlo.dynamic_update_slice %cst_1, %50, %c_10, %c_11, %c_11, %c_11 : (tensor<3x197x1x768xbf16>, tensor<1x197x1x768xbf16>, tensor<i64>, tensor<i64>, tensor<i64>, tensor<i64>) -> tensor<3x197x1x768xbf16> loc(#loc352)
// ...
%80 = stablehlo.dynamic_update_slice %cst_1, %79, %c_11, %c_11, %c_11, %c_11 : (tensor<3x197x1x768xbf16>, tensor<1x197x1x768xbf16>, tensor<i64>, tensor<i64>, tensor<i64>, tensor<i64>) -> tensor<3x197x1x768xbf16> loc(#loc381)
```

First and second `dynamic_update_slice` would yield the same CPU-hoisted function name (since the input and output types match), but the first would attempt to use 4 arguments, and the second one would attempt to use 3 arguments (because of the deduplication). This would lead to compilation failure, as described in the issue.

> Today we are just using main function name + shapes of arguments which is just wrong.

Completely agreed - indeed, this is the main culprit behind the issue. I've thought of this solution to only be temporary, to unblock the training folks. I've opened an issue (#6465) to properly handle this.

mtopalovicTT:
> #6465

I see so it's naming collision. Yea we should definitely just hash the whole hoisted function to avoid duplication of functions and collisions.

Lets go with proper approach as soon as you fix this.

PR #6474, opened 2025-12-30
Improve CPU-hoisted function deduplication

dmilinkovicTT:
### Ticket
#6465 

### Problem description
Currently, deduplication of CPU-hoisted functions relies on checking for the equality of the function names, because when single ops are CPU-hoisted, the op and the input tensor shapes are encoded in the function name. This approach is a bit fragile, especially since different classes of CPU-hoisted operations are starting to emerge (e.g. const-eval subgraphs). 

Additionally, we've already experienced an issue with the current approach for deduplication, and the issue was hotfixed as part of #6466.

This PR proposes a different approach, similar to the approach taken for const-eval cache collision detection: each function's IR is hashed, and this hash value uniquely represents the CPU-hoisted function. The prefix of the hash value is encoded in the function name to ensure uniqueness.

### What's changed
- Refactored `hoistOperationsToFunction` so that declarations and definitions are created in separate functions
- Extracted IR hashing from `TTNNToFlatbuffer`
- Applied IR hashing in CPU-hoisting pass
- Simplified CPU-hoisted functions naming
- Disabled suffixing CPU-hoisted declarations with `_decl` suffix

### Checklist

### DIFF
--- a/include/ttmlir/Support/IRHasher.h
+++ b/include/ttmlir/Support/IRHasher.h
@@ -1,4 +1,4 @@
-// SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
+// SPDX-FileCopyrightText: (c) 2026 Tenstorrent AI ULC
 //
 // SPDX-License-Identifier: Apache-2.0
 
@@ -8,10 +8,10 @@
 #include "mlir/Dialect/Func/IR/FuncOps.h"
 #include "mlir/IR/BuiltinOps.h"
 #include "llvm/ADT/ScopeExit.h"
+#include "llvm/ADT/SmallString.h"
 #include "llvm/ADT/StringExtras.h"
 #include "llvm/Support/SHA256.h"
-
-#include <string>
+#include "llvm/Support/raw_ostream.h"
 
 namespace mlir::tt {
 
@@ -24,7 +24,8 @@ namespace mlir::tt {
 // Uses printGenericOpForm to skip using custom printers and get an
 // explicit form of all op attributes and types.
 //
-inline std::string hashFuncOp(func::FuncOp func) {
+// SHA256 hex digest is 64 characters.
+inline llvm::SmallString<64> hashFuncOp(func::FuncOp func) {
   auto originalSymName = func.getSymName();
 
   // RAII guard to undo the temporary changes made to the function
@@ -55,18 +56,18 @@ inline std::string hashFuncOp(func::FuncOp func) {
   });
 
   auto flags = mlir::OpPrintingFlags().useLocalScope().printGenericOpForm();
-  std::string s;
-  llvm::raw_string_ostream os(s);
+  llvm::SmallString<1024> irBuffer;
+  llvm::raw_svector_ostream os(irBuffer);
 
   func.print(os, flags);
 
   auto digest = llvm::SHA256::hash(llvm::ArrayRef<uint8_t>(
-      reinterpret_cast<const uint8_t *>(s.data()), s.size()));
+      reinterpret_cast<const uint8_t *>(irBuffer.data()), irBuffer.size()));
 
-  constexpr bool lowercase = true;
-
-  return llvm::toHex(llvm::ArrayRef<uint8_t>(digest.data(), digest.size()),
-                     lowercase);
+  llvm::SmallString<64> result;
+  llvm::raw_svector_ostream hexOs(result);
+  hexOs << llvm::toHex(digest, /*LowerCase=*/true);
+  return result;
 }
 
 } // namespace mlir::tt
--- a/lib/Dialect/TTIR/Transforms/HoistCPUOps.cpp
+++ b/lib/Dialect/TTIR/Transforms/HoistCPUOps.cpp
@@ -328,7 +328,7 @@ static func::FuncOp createCPUHoistedFunctionDefinition(
                                  ttmlir::utils::FunctionType::ForwardCPU);
 
   // Finally, hash the function implementation and set the func_hash attribute.
-  std::string funcHash = hashFuncOp(funcDefinition);
+  auto funcHash = hashFuncOp(funcDefinition);
   funcDefinition->setAttr("func_hash", builder.getStringAttr(funcHash));
 
   return funcDefinition;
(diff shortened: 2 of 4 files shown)


include/ttmlir/Conversion/Passes.td
```diff
@@ -224,15 +224,15 @@ def EmitPyLinkModules : Pass<"emitpy-link-modules", "::mlir::ModuleOp"> {
...
       ttcore.cpu_module {
         module {
-          func.func @hoisted_function() { ... }  // Definition of hoisted function
+          func.func @cpu_hoisted_function() { ... }  // Definition of hoisted function
         }
       }
       ttcore.device_module {
         module {
           emitpy.import "ttnn"
-          func.func private @hoisted_function_decl()  // Stub declaration
```
vwellsTT, line 233: subjectively, I found it a bit easier to debug with slightly different names in device_module vs cpu_module, so I added _decl prefix even though it's not needed.  Don't strongly object to this change or anything, but I do think there's a small downside when you get an MLIR error and can mistake which version of the func it's complaining about.

include/ttmlir/Support/IRHasher.h
```diff
@@ -0,0 +1,74 @@
+// SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
```
vwellsTT: nit: 2026

include/ttmlir/Support/IRHasher.h
```diff
@@ -0,0 +1,74 @@
...
+            func->getParentOfType<mlir::ModuleOp>().lookupSymbol<func::FuncOp>(
+                callOp.getCallee())) {
+      auto calleeHash = hashFuncOp(calleeFunc);
+      callOp->setAttr("callee_hash",
+                      mlir::StringAttr::get(func.getContext(), calleeHash));
+    }
+  });
+
+  auto flags = mlir::OpPrintingFlags().useLocalScope().printGenericOpForm();
+  std::string s;
```
vwellsTT: nit: technically we prefer llvm::SmallString or similar when not inconvenient; I think for this case it would be fine?

lib/Dialect/TTIR/Transforms/HoistCPUOps.cpp
```diff
@@ -243,161 +215,261 @@ struct CPUHoistedOpsDescriptor {
...
-      }
+  return funcDeclaration;
+}
 
-      // Check if this is the output producing op. If it is, keep track of it
-      // for later.
-      if (llvm::is_contained(outputProducers, opToHoist)) {
-        clonedOutputProducers.push_back(clonedOp);
-      }
+static std::optional<func::FuncOp>
```
vwellsTT: debatably would be simpler to just return a null FuncOp for empty case and ditch optional (but in some ways this is clearer, so don't feel strongly either way)

include/ttmlir/Dialect/TTIR/Transforms/Passes.td
```diff
@@ -51,15 +51,15 @@ def CPUHoistTransform: Pass<"cpu-hoist-transform", "::mlir::ModuleOp">
...
             return %1 : tensor<32x32xbf16>
           }
-          func.func private @hoisted_ttir_add_32x32xbf16_32x32xbf16_32x32xbf16_func_decl(tensor<32x32xbf16>, tensor<32x32xbf16>, tensor<32x32xbf16>) -> tensor<32x32xbf16>
+          func.func private @cpu_hoisted_ttir_add_<hash>(tensor<32x32xbf16>, tensor<32x32xbf16>, tensor<32x32xbf16>) -> tensor<32x32xbf16>
         }
       }
       ttcore.cpu_module {
         builtin.module {
-          func.func @hoisted_ttir_add_32x32xbf16_32x32xbf16_32x32xbf16_func(%arg0: tensor<32x32xbf16>, %arg1: tensor<32x32xbf16>, %arg2: tensor<32x32xbf16>) -> tensor<32x32xbf16> attributes {arg_ranks = [2, 2, 2, 2], result_ranks = [2]} {
+          func.func @cpu_hoisted_ttir_add_<hash>(%arg0: tensor<32x32xbf16>, %arg1: tensor<32x32xbf16>, %arg2: tensor<32x32xbf16>) -> tensor<32x32xbf16> attributes {arg_ranks = [2, 2, 2, 2], result_ranks = [2]} {
```
vwellsTT, line 62: will these still have "add" at all, or just be pure hash?

vwellsTT (approved):
Another good improvement! A few minor nits inline

include/ttmlir/Dialect/TTIR/Transforms/Passes.td
vwellsTT: looked into this, my impression is yes, will have op-name for single-op funcs, will have nothing for const-eval hoists, is that right?

dmilinkovicTT: for individual ops, the op is encoded in the name. and for const-eval funcs, the name of the const-eval func itself is encoded into the CPU-hoisted function name

lib/Target/TTNN/TTNNToFlatbuffer.cpp
```diff
@@ -2367,12 +2337,12 @@ createOp(FlatbufferObjectCache &cache, ttcore::LoadCachedOp op,
...
   assert(itHash != constEvalFuncHashes.end() &&
          "Const-eval function hash not found in const-eval function hash map!");
-  const std::string &funcHash = itHash->second;
+  const auto &funcHash = itHash->second;
 
   // Create the LoadCachedOp with indices instead of inputs
   return ::tt::target::ttnn::CreateLoadCachedOpDirect(
       *cache.fbb, &ins, op.getCallee().str().c_str(), programIdx, &outputs,
-      funcHash.c_str());
+      funcHash.str().c_str());
```
vmilosevic: :warning: **clang\-diagnostic\-error** :warning:
no member named `` c_str `` in `` llvm::StringRef ``

vmilosevic (commented):

include/ttmlir/Conversion/Passes.td
dmilinkovicTT: makes sense. tbh, I'm not strongly opinionated on either of the approaches.

the main reason behind the change was an attempt to slightly reduce the mental overhead, as some folks here were confused about the difference in names, thinking that there's a deeper technical reason behind it, and I was also a bit confused when I've initially started looking into the CPU-hoisting.

> which version of the func it's complaining about

the only place where I can currently see function name being useful in debugging is argument type mismatches; however, since only the declarations ever get `call`-ed, I would assume having the `decl` suffix would not add value there

wdyt?

vwellsTT: Hmm, yeah maybe you're right.  Ideally we won't be getting compilation errors anyway 😄

PR #6477, opened 2025-12-30
D2M: Add 1D Matmul Heuristic and supporting lowering changes

jdesousa-TT:
### Problem description
We currently run out of memory / utilize a fraction of the grid when we have high aspect ratio input tensors.

### What's changed
- Fixes lowering of virtual grid tensors to make sure operands that share dimensions match grid layouts
- Inserts views to ensure the generic sees the input tensors as matching (with each other and with the output)
- Fixes multicast lowering such that we can support both 2D multicasts and 1D multicasts
- Hard-codes CB initialization shapes to include the full device grid, instead of pulling from memref shapes which may have block factors that increase grid size beyond max worker grid
- Changes the implementation of CoreIndexOp to apply the grid virtualization mapping. This requires removing the dependence on this op at the ttkernel level, replaced with MyLogicalX/MyLogicalY ops instead.

### Checklist

### DIFF
--- a/lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
+++ b/lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
@@ -12,6 +12,7 @@
 #include "ttmlir/Dialect/TTKernel/IR/TTKernelOps.h"
 #include "ttmlir/Utils.h"
 
+#include "mlir/Conversion/AffineToStandard/AffineToStandard.h"
 #include "mlir/Dialect/Affine/IR/AffineOps.h"
 #include "mlir/Dialect/Affine/ViewLikeInterfaceUtils.h"
 #include "mlir/Dialect/Func/IR/FuncOps.h"
@@ -24,7 +25,6 @@
 #include "mlir/IR/Value.h"
 #include "mlir/Transforms/DialectConversion.h"
 
-#include "mlir/Conversion/AffineToStandard/AffineToStandard.h"
 #include <type_traits>
 #include <utility>
 
@@ -1267,7 +1267,6 @@ public:
   matchAndRewrite(d2m::CoreIndexOp op, d2m::CoreIndexOpAdaptor adaptor,
                   ConversionPatternRewriter &rewriter) const final {
 
-    // Base physical (logical) coordinates.
     Value logicalY = rewriter.create<ttkernel::MyLogicalYOp>(op.getLoc());
     Value logicalX = rewriter.create<ttkernel::MyLogicalXOp>(op.getLoc());
 
--- a/lib/Dialect/D2M/Transforms/GridSelection.cpp
+++ b/lib/Dialect/D2M/Transforms/GridSelection.cpp
@@ -20,6 +20,8 @@
 #include "llvm/ADT/DenseMap.h"
 #include "llvm/ADT/SmallVector.h"
 
+#include <cstdint>
+
 namespace mlir::tt::d2m {
 
 //--------------------------------------------------------
@@ -430,14 +432,14 @@ normalizeOperandGridsForGeneric(
   llvm::SmallVector<llvm::SmallVector<int64_t>> normalizedOperandGrids(
       optimalOperandGrids.begin(), optimalOperandGrids.end());
 
-  unsigned numInputs = genericOp.getNumDpsInputs();
+  uint64_t numInputs = genericOp.getNumDpsInputs();
   // Map: loopDim -> list of (operandIndex, operandDimIdx) pairs that reference
   // this loop dimension in their indexing maps.
-  llvm::DenseMap<int64_t, llvm::SmallVector<std::pair<unsigned, unsigned>>>
+  llvm::DenseMap<int64_t, llvm::SmallVector<std::pair<uint64_t, uint64_t>>>
       dimToInputOperandDims;
 
   auto indexingMaps = genericOp.getIndexingMapsValue();
-  for (unsigned operandIndex = 0; operandIndex < numInputs; ++operandIndex) {
+  for (uint64_t operandIndex = 0; operandIndex < numInputs; ++operandIndex) {
     AffineMap operandIndexingMap = indexingMaps[operandIndex];
     auto results = operandIndexingMap.getResults();
     for (auto [operandDimIdx, expr] : llvm::enumerate(results)) {
@@ -447,7 +449,7 @@ normalizeOperandGridsForGeneric(
       }
       int64_t loopDim = dimExpr.getPosition();
       dimToInputOperandDims[loopDim].push_back(
-          std::make_pair(operandIndex, static_cast<unsigned>(operandDimIdx)));
+          std::make_pair(operandIndex, static_cast<uint64_t>(operandDimIdx)));
     }
   }
 
@@ -501,9 +503,9 @@ normalizeOperandGridsForGeneric(
 
       AffineMap indexingMap = genericOp.getIndexingMap(operandIndex);
       auto results = indexingMap.getResults();
-      if (results.size() != normalizedOperandGrids[operandIndex].size()) {
-        continue;
-      }
+      TT_assertv(results.size() == normalizedOperandGrids[operandIndex].size(),
+                 "indexing map results size does not match normalized operand "
+                 "grids size");
 
       for (auto [resultIdx, expr] : llvm::enumerate(results)) {
         auto dimExpr = mlir::dyn_cast<AffineDimExpr>(expr);
--- a/test/python/golden/test_metal_matmul.py
+++ b/test/python/golden/test_metal_matmul.py
@@ -253,7 +253,7 @@ def test_matmul_1d_shapes(
         device=device,
         custom_pipeline=f"ttir-to-ttmetal-pipeline{{{' '.join(options)}}}",
         test_base=request.node.name,
-        module_dump=True,
+        save_artifacts=True,
         print_ir=True,
         output_root=request.config.getoption("--path"),
         system_desc_path=request.config.getoption("--sys-desc"),


include/ttmlir/Dialect/TTKernel/IR/TTKernelOps.td
```diff
@@ -2608,11 +2608,11 @@ def TTKernel_NocSemaphoreSetMulticastLoopbackOp : TTKernel_Op<"noc_semaphore_set
     let arguments = (ins TTKernel_Semaphore:$src_local_l1_addr,
                          TTKernel_NocAddr:$dst_noc_addr_multicast,
                          I32:$num_dests,
-                         BoolAttr:$linked,
-                         BoolAttr:$multicast_path_reserve);
+                         OptionalAttr<BoolAttr>:$linked,
```
bgrady-tt: Having multiple optional arguments can cause silent bugs; Armin and I have both dealt with issues related to this. Better to not use more than one optional attribute.

lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
```diff
@@ -1272,6 +1279,46 @@ class D2MCoreIndexRewriter : public OpConversionPattern<d2m::CoreIndexOp> {
...
+    }
+    if (mapAttr.isEmpty() && op.getDim() == 1) {
+      rewriter.replaceOp(op, logicalX);
+      return success();
+    }
+    // We add 1 here because the grid mapping includes a leading device index
+    // result
+    auto dimAttr = op.getDim() + 1;
+
+    llvm::SmallBitVector dropDims(mapAttr.getNumResults(), 1);
```
bgrady-tt: Add comment about why the dims and results are being dropped.

lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
```diff
@@ -1665,6 +1717,7 @@ void populateD2MToTTKernelPatterns(
   patterns.add<ttkernel::D2MGetGlobalOperandRewriter>(typeConverter, ctx, ttnnMode);
   patterns.add<ttkernel::D2MDMAReadRewriter>(typeConverter, ctx, &associatedDMAWaits, &cbProducerConsumer);
   patterns.add<ttkernel::D2MDMAWriteRewriter>(typeConverter, ctx, &associatedDMAWaits, &cbProducerConsumer);
+  populateAffineToStdConversionPatterns(patterns);
```
bgrady-tt, line 1707: add comment about this lowering AffineApply ops to arith

lib/Target/TTMetal/TTMetalToFlatbuffer.cpp
```diff
@@ -438,8 +438,15 @@ memrefTypeToCircularBufferConfigFlatbuffer(FlatbufferObjectCache &cache,
   auto extendedMapping = extendMappingForHigherDimGrid(
       device.getWorkerGrid().getMapping(), memrefGridShape.size());
 
+  SmallVector<int64_t> workerGridShape =
```
bgrady-tt, line 445: why is this necessary?

lib/Dialect/D2M/Transforms/GenericGenerateDatamovement.cpp
```diff
@@ -123,50 +124,61 @@ class D2MGenericGenerateDatamovementRewriter
...
         Value condition = rewriter.create<arith::CmpIOp>(
-            loc, rewriter.getI1Type(), mlir::arith::CmpIPredicate::eq, core,
+            loc, rewriter.getI1Type(), mlir::arith::CmpIPredicate::eq, virtCore,
             zero);
         args.conditions.push_back(condition);
       }
     }
 
+    if (!outputShardLayout.getCoreVirtualizationMap().isEmpty()) {
+      auto coreVirtMap = outputShardLayout.getCoreVirtualizationMap();
```
bgrady-tt, line 172: I would leave a comment here about why the affine map is being manipulated.

lib/Target/TTMetal/TTMetalToFlatbuffer.cpp
jdesousa-TT: `memrefGridShape` can be bigger than the `workerGridShape` depending on the sharding / blocking scheme. This fails at runtime because we attempt to create CBs on cores that don't exist. 

I asked a while ago if there was objection to just defaulting this to the worker grid shape since CBs are allocated in lockstep across cores anyway, seemed to be none. If you have a less janky idea I'm in

lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
bgrady-tt, line 1191: nit: Since multicast shape is known at compile-time, I think we could just create `numDestMinusOne` as a constant ahead of time.

lib/Dialect/D2M/Transforms/GridSelection.cpp
```diff
@@ -998,7 +1116,9 @@ static void assignGrids(d2m::GenericOp genericOp,
         insertTTNNDRAMStreams(genericOp, targetSquareGridShape);
   }
 
-  recreateGenericOp(genericOp, optimalOperandGrids);
+  llvm::SmallVector<llvm::SmallVector<int64_t>> normalizedOperandGrids =
+      normalizeOperandGridsForGeneric(genericOp, optimalOperandGrids);
+  recreateGenericOp(genericOp, normalizedOperandGrids);
```
bgrady-tt: I am not sure this new grid selection logic should apply to ops with TTNN operands. For TTNN interop we maintain grid shapes even if they are incompatible/wrong.

bgrady-tt (commented):
I have some minor concerns that need addressing, but close to LGTM.

lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
jdesousa-TT: I think we'd have to change the definition of DMA op to do this since the mcastShape is a `Variadic<Index>` which are one or more runtime values.

lib/Target/TTMetal/TTMetalToFlatbuffer.cpp
bgrady-tt: Ah ok this makes sense. I would note that in a comment; that the adjusted shape is unused (CBs aren't accessed by host) and spoofed as a workaround.

lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
bgrady-tt: I guess this is strictly more flexible, but it is a bit weird.

jdesousa-TT:
@nsmithtt & @bgrady-tt, made some changes after our conversations, please take another look!

include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
```diff
@@ -1604,11 +1604,41 @@ def D2M_IterIndexOp : D2M_IndexOp<"iter_index"> {
...
+  ]> {
     let summary = "Core Index op.";
     let description = [{
       Return the index of this core's coordinate inside the generic op's grid dimension.
     }];
+
+    // Optional physical-to-virtual grid mapping. When present, this op returns
+    // the virtualized core coordinate for `dim`.
+    let arguments = (ins
+      ConfinedAttr<I64Attr, [IntMinValue<0>, IntMaxValue<1>]>:$dim,
```
nsmithtt: The max value is not 1, it's variable based on the rank of the grid.

include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
```diff
@@ -1604,11 +1604,41 @@ def D2M_IterIndexOp : D2M_IndexOp<"iter_index"> {
...
+      OptionalAttr<AffineMapAttr>:$phys_to_virt_map
```
nsmithtt, line 1622: This _must_ be identical to the map that exists on the grid attribute on the generic op, I think we should avoid the redundancy if possible.  Perhaps I'm missing the reasoning here.

nsmithtt: Ok upon further review I realized that when we move the regions to top level module funcs the core_index op no longer has access to its generic parent :(

If there's a way to not have to do this that'd be great, but I'm not seeing it at the moment, so ignore my comment above.

lib/Dialect/D2M/Transforms/GridSelection.cpp
```diff
@@ -405,6 +407,121 @@ struct EmptyUpdateInfo {
   bool isVirtualGrid = false;
 };
 
+// This function normalizes the operand grids for a generic operation by
+// ensuring that the grids are consistent across all operands that share the
+// same loop dimension. We also need to make sure that the grids respect any
+// constraints implied by the outputs' grids. If multiple operands participate
+// in the same loop dimension, the corresponding grid extents must agree.
+static llvm::SmallVector<llvm::SmallVector<int64_t>>
+normalizeOperandGridsForGeneric(
```
nsmithtt, line 418: @vwellsTT do you mind auditing this file?

lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
```diff
@@ -23,6 +24,7 @@
 #include "mlir/IR/Value.h"
 #include "mlir/Transforms/DialectConversion.h"
 
+#include "mlir/Conversion/AffineToStandard/AffineToStandard.h"
```
vwellsTT: nit: wrong block

lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
```diff
@@ -1257,16 +1260,39 @@ class D2MCoreIndexRewriter : public OpConversionPattern<d2m::CoreIndexOp> {
...
-    assert(op.getDim() == 0 ||
-           op.getDim() == 1 &&
-               "Expected core index dim to be in range 0-1, failing.");
-    if (op.getDim()) {
-      auto coreIndex = rewriter.create<ttkernel::MyLogicalXOp>(op.getLoc());
-      rewriter.replaceOp(op, coreIndex);
-    } else {
-      auto coreIndex = rewriter.create<ttkernel::MyLogicalYOp>(op.getLoc());
-      rewriter.replaceOp(op, coreIndex);
+    // Base physical (logical) coordinates.
```
vwellsTT: nit: confusing comment, is it virtual or physical, or some juxtaposition?

lib/Dialect/D2M/Transforms/GridSelection.cpp
```diff
@@ -405,6 +407,121 @@ struct EmptyUpdateInfo {
...
+
+  // First, normalize input operand grids for operands that share loop
+  // dimensions. For example, in a matmul, the two inputs share the reduction
+  // dimension. If their independently chosen optimal grids differ along that
+  // dimension, promote the grid factor for that *dimension only* to the
+  // maximum across all inputs that share it.
+  llvm::SmallVector<llvm::SmallVector<int64_t>> normalizedOperandGrids(
+      optimalOperandGrids.begin(), optimalOperandGrids.end());
+
+  unsigned numInputs = genericOp.getNumDpsInputs();
```
vwellsTT: nit: subjectively I prefer uint32_t to unsigned, feels more consistent with other types we're using (int64_t, etc), and generally prefer being explicit about int sizes in C++.

lib/Dialect/D2M/Transforms/GridSelection.cpp
```diff
@@ -405,6 +407,121 @@ struct EmptyUpdateInfo {
...
+  if (outputConstraints) {
+    for (auto [operandIndex, operand] :
+         llvm::enumerate(genericOp->getOpOperands())) {
+      if (genericOp.isDpsInit(&operand)) {
+        continue;
+      }
+
+      AffineMap indexingMap = genericOp.getIndexingMap(operandIndex);
+      auto results = indexingMap.getResults();
+      if (results.size() != normalizedOperandGrids[operandIndex].size()) {
```
vwellsTT: when would this happen? should it be an assert?

lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
jdesousa-TT: yeah should be removed altogether. logical is in the name of the following ops, ty

lib/Dialect/D2M/Transforms/GridSelection.cpp
jdesousa-TT: Good catch, this shouldn't happen. Better to assert than silently not apply the normalization in some cases. Fixed.

PR #6480, opened 2025-12-31
[runtime] Exposed getDeviceTensors in runtime (#6059)

acolicTT:
Exposed getDeviceTensors call in runtime, so that frontends can use it to access single shard of device tensor.

### DIFF
--- a/runtime/test/ttnn/python/n300/test_runtime_api.py
+++ b/runtime/test/ttnn/python/n300/test_runtime_api.py
@@ -67,7 +67,23 @@ def test_open_mesh_device(
     ttrt.runtime.close_mesh_device(device)
 
 
+# Converts runtime device tensor to torch tensor.
+# It first moves device tensor to host, and creates torch tensor from host tensor.
+def device_tensor_to_torch_tensor(runtime_tensor):
+    host_tensor = ttrt.runtime.to_host(runtime_tensor, untilize=True)
+    assert len(host_tensor) == 1
+
+    shape = host_tensor[0].get_shape()
+    host_tensor[0].get_dtype()
+    torch_tensor = torch.zeros(
+        shape, dtype=ttrt_datatype_to_torch_dtype(host_tensor[0].get_dtype())
+    )
+    ttrt.runtime.memcpy(torch_tensor.data_ptr(), host_tensor[0])
+    return torch_tensor
+
+
 # Testing get_device_tensors API.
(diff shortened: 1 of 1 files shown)


pilkicTT (approved):

jameszianxuTT:
Sorry, so for context, is this to facilitate https://github.com/tenstorrent/tt-xla/pull/2707?

acolicTT:
> Please add a testcase for this API. Since you've already made the binding for it in runtime/python/runtime/runtime.cpp this shouldn't be too complex. I think runtime/test/ttnn/python/n300/test_runtime_api.py would be a suitable place to add it and just check the continued functionality of this API
> 
> Also, if it is not too urgent to get this in, it would be nice going forward to implement the distributed API and distributed tests for new runtime APIs instead of leaving them as unimplemented. We can discuss this offline too :)

Added.

Maybe it is better to check this in alone to unblock PJRT refactoring, (since distributed API can be implemented independently).

runtime/test/ttnn/python/n300/test_runtime_api.py
```diff
@@ -63,3 +65,93 @@ def test_open_mesh_device(
...
+            assert (
+                len(host_tensor) == 1
+            ), f"Expected 1 host tensor, got {len(host_tensor)}"
+
+            torch_output_tensor = torch.zeros(shape, dtype=dtype)
+            ttrt.runtime.memcpy(torch_output_tensor.data_ptr(), host_tensor[0])
+
+            assert_pcc(torch_tensor, torch_output_tensor, threshold=0.99)
+
+        ttrt.runtime.deallocate_tensor(device_tensor, force=True)
```
jameszianxuTT, line 115: The logic here is a bit hard to follow so I suggest adding some more comments. My understanding is that you check if a tensor can be replicated to a 1xN mesh, then pull its shards back to host and check that the replicas are equal to the original in these steps:

1. create a random torch_tensor
2. create owned runtime_tensor from (1)
3. create device_tensor by moving (2) to device
4. create device_tensors list by calling get_device tensors on (2) -> one replica per device in the 1x2 mesh
5. iterate over the 2 device tensors, and for each tensor:
6. create host_tensor by moving the device tensor to host
7. create a blank torch_output_tensor torch.zeroes
8. memcpy the contents of the host tensor into the torch_output_tensor from (7)
9. check equality of (8) and (1)

jameszianxuTT (approved):

Issue #6386, opened 2025-12-18
Implement proper forward function detection across targets


dmilinkovicTT:
As @svuckovicTT pointed out [here](https://github.com/tenstorrent/tt-mlir/pull/6198#discussion_r2609881621), we don't have a proper way of differentiating between *forward* functions and other types of functions (const-eval, CPU-hoisted, input generators, main function...). 

Instead, various conditions for testing if function is a *forward* are spread throughout the codebase, allowing for bugs and inconsistent behaviors.

We should try to enable more robust *forward* function detection mechanism across the targets.

PR #6502, opened 2026-01-05
Introduce an API for determining function types

dmilinkovicTT:
### Ticket
#6386 

### Problem description
Currently, we don't have a proper way of detecting if a function is a "forward" function. Across the codebase, in many places, we're checking if a certain function is **not** of a certain type (e.g. const-eval and CPU-hoisted) in order to detect forward functions, which is bug-prone and non-scalable as new types of functions get introduced.

This PR aims to introduce a more robust mechanism for handling various function types.

### What's changed
- Introduced `FunctionType` enum and related helper functions
- Applied correct function types to all generated functions
- Added a pass at the start of TTNN / TTMetal pipelines to assign `ForwardDevice` function type to all functions without a type assigned

### Checklist

### DIFF
--- a/lib/Dialect/TTNN/Transforms/TTNNRecoverStructure.cpp
+++ b/lib/Dialect/TTNN/Transforms/TTNNRecoverStructure.cpp
@@ -96,7 +96,8 @@ public:
 
 private:
   bool isCandidateFn(func::FuncOp funcOp) {
-    return ttmlir::utils::isForwardDeviceFunc(funcOp);
+    return ttmlir::utils::isForwardDeviceFunc(funcOp) &&
+           !funcOp.getFunctionType().getInputs().empty();
   }
 
   bool isCandidateOp(Operation *op) {


lib/Dialect/TTNN/Pipelines/TTNNPipelines.cpp
```diff
@@ -225,6 +225,12 @@ void createTTIRToTTNNDevicePipeline(
   // Create device module, if not already present.
   pm.addPass(ttcore::createTTCoreWrapDeviceModulePass());
 
+  // Mark all public functions as forward functions before any hoisting
+  // transforms. This provides a consistent mechanism for identifying
+  // forward functions downstream.
+  pm.nest<ttcore::DeviceModuleOp>().nest<mlir::ModuleOp>().addPass(
+      ttcore::createTTCoreMarkFunctionsAsForwardPass());
```
svuckovicTT: Would it make sense to have this pass run at the beginning of `createTTIRToTTNNBackendPipeline`?

include/ttmlir/FunctionTypes.h
```diff
@@ -0,0 +1,209 @@
...
+namespace ttmlir::utils {
+
+/// Enum representing the different types of functions in the compilation
+/// pipeline.
+enum class FunctionType {
+  /// Forward function - the main computational graph intended for device
+  /// execution.
+  Forward,
+  /// Const-eval function - a function that performs constant evaluation
+  /// at compile time.
```
svuckovicTT: Wouldn't this be runtime though?

include/ttmlir/FunctionTypes.h
```diff
@@ -0,0 +1,209 @@
...
+  /// Kernel function - a low-level function representing a kernel
+  /// operation.
+  Kernel,
+  /// Input generator function - generates input data for testing/execution.
+  /// Relevant for EmitPy and EmitC targets.
+  InputGenerator,
+  /// Main function - the entry point for the generated program.
+  /// Relevant for EmitPy and EmitC targets.
+  Main,
+};
```
svuckovicTT, line 61: Nit: for readability, consider adding blank lines after each enum value.

lib/Dialect/TTCore/Transforms/CMakeLists.txt
```diff
@@ -1,5 +1,6 @@
 add_mlir_dialect_library(MLIRTTTransforms
         OptimizationBarrierFold.cpp
+        MarkFunctionsAsForward.cpp
```
svuckovicTT, line 2:
```suggestion
add_mlir_dialect_library(MLIRTTTransforms
        MarkFunctionsAsForward.cpp
        OptimizationBarrierFold.cpp
```

svuckovicTT (approved):

lib/Dialect/TTNN/Pipelines/TTNNPipelines.cpp
dmilinkovicTT: If it were at the beginning of the pipeline, we would run it against the root module, which could itself contain both Device and CPU module, and thus, it could tag both forward and CPU-hoisted functions created during the StableHLO pipeline.

This way, we're ensuring the pass runs only on the Device module, and it picks up only actual forward functions.

svuckovicTT: I see, we differentiate between "device forward" fns (`Forward`) and "cpu-hoisted forward" funcs (`CPUHoisted`).

I would propose we change the taxonomy a bit - we already have MLIR modules separating the devices we run on (cpu vs tt hardware [device]), so I'm thinking it would make sense to remove the `CPUHoisted` enum and use `Forward` instead. That way it's more clear what `Forward` means, regardless of what hardware it runs on - cpu hoisted funcs are still part of the main computational graph.

Thoughts?

dmilinkovicTT: Makes sense, but the issue with the current implementation is that CPU-hoisted function declarations (stubs) in the Device module also get assigned `CPUHoisted` function type.

One of my goals with this PR is to remove all assumptions about the nature of the functions in pass implementations, and query the function type instead. For example, some passes are relying on a somewhat-fragile assumption that (function declaration <=> CPU-hoisted function declaration) - true for now, but risky in the long run.

I should probably introduce `CPUHoistedDeclaration` function type to resolve this issue.

But even with this, I think that having CPU-hoisted function definitions marked as `forward` carries some risks:
- For emitpy, the modules get merged, after which we don't have info about which functions are *really* forward. And this information seems to be relevant to the ttrt / builder runtime executors ([example](https://github.com/tenstorrent/tt-mlir/blob/a5755a04e656727bb137fbfd4e0920b067eb4fa8/tools/builder/base/builder_runtime.py#L708)).
- My aim with #6361 was to decouple the pass implementation from the information about whether it operates on CPU or the Device module (except for, of course, passes which need to operate on both modules, such as CPU-hoisting and module linking passes). My assumption is that marking CPU-hoisted functions as `Forward` could (in the future)  break some passes which really expect to operate on true-forward functions. 

After all, `Forward` might not be the best name - something like `ModelEntryPoint` or `ForwardDeviceGraph` (but a bit less verbose) could perhaps fit better. Wdyt?

svuckovicTT: Agreed on declarations and adding a tag for that.

> For emitpy, the modules get merged, after which we don't have info about which functions are really forward. And this information seems to be relevant to the ttrt / builder runtime executors ([example](https://github.com/tenstorrent/tt-mlir/blob/a5755a04e656727bb137fbfd4e0920b067eb4fa8/tools/builder/base/builder_runtime.py#L708)).

Since builder works with code in this case (instead of IR), it will always have to rely on naming convention, unless we change its input (e.g. we add decorators in generated code). Not sure I'm following how the current PR might help address this issue?

> My aim with https://github.com/tenstorrent/tt-mlir/pull/6361 was to decouple the pass implementation from the information about whether it operates on CPU or the Device module (except for, of course, passes which need to operate on both modules, such as CPU-hoisting and module linking passes). My assumption is that marking CPU-hoisted functions as Forward could (in the future) break some passes which really expect to operate on true-forward functions.

Agreed on this, it kind of exposes our premise that a 1-dim attribute is enough to solve all of these problems. Having a kind of hierarchy would be more expressive, but also comes with drawbacks of managing/using it, so probably not worth going down that road...

Maybe something like `ForwardCPU` and `ForwardDevice` might make sense? It'd be the only type that has flavors, and I'd expect (hope?) there wouldn't be a need for more of these in the future :D Still thinking out loud here, please share your thoughts!


To circle back to the original comment, would it work if we changed the pass from `createTTCoreMarkFunctionsAsForwardPass` to `createTTCoreMarkFunctionTypes`, moved it to the front of the `createTTIRToTTNNBackendPipeline` pipeline and have it mark all the fns that exist at that point? That way it would be the first thing to run and all the following passes could rely on function types having been set.

dmilinkovicTT: > (e.g. we add decorators in generated code)

That was indeed my idea, and it was pushed to this branch in the meantime (waited for #6348 to get merged). Does this fit well with the ongoing efforts of modularizing / splitting the generated Python files?

> Maybe something like ForwardCPU and ForwardDevice might make sense? It'd be the only type that has flavors, and I'd expect (hope?) there wouldn't be a need for more of these in the future :D Still thinking out loud here, please share your thoughts!

Yep, makes perfect sense.

> would it work if we changed the pass from createTTCoreMarkFunctionsAsForwardPass to createTTCoreMarkFunctionTypes, moved it to the front of the createTTIRToTTNNBackendPipeline pipeline and have it mark all the fns that exist at that point? That way it would be the first thing to run and all the following passes could rely on function types having been set.

Actually, the reasoning in my previous comments was flawed - we could already move it to the top without changing the semantics of the pass, as it will tag only the functions which haven't already been tagged, which will skip potential `ForwardCPU` functions which get tagged during the CPU-hoisting pass in the SHLO pipeline. In short:
- all compiler-generated functions should get tagged during creation
- the untagged ones at the beginning of the pipeline can be assumed to be `ForwardDevice`

Maybe we could introduce a constraint which would require every `FuncOp` to be assigned a type?

svuckovicTT: > That was indeed my idea, and it was pushed to this branch in the meantime (waited for https://github.com/tenstorrent/tt-mlir/pull/6348 to get merged). Does this fit well with the ongoing efforts of modularizing / splitting the generated Python files?

Yeah, this makes sense for testing, but not for end users. I think it's okay if we put the decorators under a flag, to be used for testing infra only, e.g. `--add-test-decorators`.

> Maybe we could introduce a constraint which would require every FuncOp to be assigned a type?

Agreed!

lib/Dialect/TTNN/Transforms/TTNNRecoverStructure.cpp
```diff
@@ -95,8 +96,7 @@ class TTNNRecoverStructure
 
 private:
   bool isCandidateFn(func::FuncOp funcOp) {
-    return !funcOp.isPrivate() && !ttmlir::utils::isConstEvalFunc(funcOp) &&
-           !funcOp.getFunctionType().getInputs().empty();
+    return ttmlir::utils::isForwardDeviceFunc(funcOp);
```
svuckovicTT:
```suggestion
    return ttmlir::utils::isForwardDeviceFunc(funcOp) && !funcOp.getFunctionType().getInputs().empty();;
```

lib/Dialect/TTNN/Transforms/Passes.cpp
```diff
@@ -247,7 +248,7 @@ class TTNNInputFunctionCreatorBase {
         rewriter.modifyOpInPlace(funcOp, [&]() { funcOp.setSymName("_main"); });
       }
 
-      if (funcOp.isPrivate() || ttmlir::utils::isConstEvalFunc(funcOp)) {
+      if (funcOp.isPrivate()) {
```
svuckovicTT, line 251: Does this not change the logic?

dmilinkovicTT: It shouldn't, as `ConstEval` funcs are already private

test/python/golden/mlir_snippets/ttir/stablehlo_dynamic_update_slice.mlir
jgrimTT, line 1: Move this file to `test/python/golden/mlir_snippets/stablehlo`

test/python/golden/mlir_snippets/ttir/ttir_cpu_module.mlir
jgrimTT, line 1: Move this file to `test/python/golden/mlir_snippets/stablehlo`

jgrimTT (approved):
Requested small changes, otherwise looks great!

test/python/golden/mlir_snippets/ttir/stablehlo_dynamic_update_slice.mlir
dmilinkovicTT: The test is failing when I move it; seems like the builder infra can handle device/CPU module separation only on TTIR graphs?

test/python/golden/mlir_snippets/ttir/ttir_cpu_module.mlir
dmilinkovicTT: Same as above

PR #6515, opened 2026-01-07
D2M: fix incorrect placement of copy loop guards

wenbinlyuTT:
### Problem description
After #5961, the load guards are moved to the inside of the cloned affine loops, in the hope that all the guarded and normal loads can share the same affine loop nest.

Unfortunately, the init hoisting pass hoists the `copy_tile_init` call to the outside of the guard, so it's invoked unconditionally, causing performance regression in matmul tests.

### What's changed
- The load guard now dominates the cloned affine loop nest.
- As a result, each guarded load now lives in its own loop nest.
- Bcast loop nests are inserted to the top of the compute tiling loops #6516.
- All normal loads still share the same loop nest.

### DIFF
--- a/lib/Dialect/D2M/Transforms/InsertDstRegisterAccess.cpp
+++ b/lib/Dialect/D2M/Transforms/InsertDstRegisterAccess.cpp
@@ -681,17 +681,12 @@ public:
               AffineMap dstAccessMap, ValueRange dstAccessIndices) {
             auto loc = record.loadStore.getLoc();
             Value cb = record.loadStore.getMemref();
-            const bool isBcastGuard = record.bcast.has_value();
-            auto guard = createLoadLoopGuard(rewriter, loc, record.guardDims,
-                                             isBcastGuard);
-            if (guard) {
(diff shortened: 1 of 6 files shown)


lib/Dialect/D2M/Transforms/InsertDstRegisterAccess.cpp
```diff
@@ -843,20 +834,43 @@ struct D2MInsertDstRegisterAccessRewriter final
...
+      return {skeleton, mapper};
+    };
+
+    auto [copyLoop, copyLoopMapper] = cloneLoopSkeleton(rewriter, loopNestOrOp);
 
     for (auto record : loadStoreRecords) {
+      mlir::IRMapping irMapper = copyLoopMapper;
+      if (!record.guardDims.empty()) {
+        const bool isBcastGuard = record.bcast.has_value();
+        // TODO(wenbinlyuTT): remove once bcast init doesn't reset the DST bank.
```
nsmithtt: Can you file an issue?

nsmithtt (approved):

PR #6520, opened 2026-01-07
Miscellaneous fixes and improvements

wenbinlyuTT:
Two small improvements I've accumulated in the past month:
- Make the dumped kernel names more clear
  - Before `..._datamovement_..._data_movement0_noc1__0_0-4_4`
  - After `..._datamovement_..._ncrisc_noc1__y0x0-y4x4`
- Extend #6076 and create unique build dir for each cmake build type because RelWithDebInfo & ASan are also useful
  - Create a `build` symlink to point to the most recent build type, similar to what tt-metal does

### DIFF
--- a/tools/builder/base/builder_runtime.py
+++ b/tools/builder/base/builder_runtime.py
@@ -33,127 +33,6 @@ class TTBuilderGoldenException(Exception):
     pass
 
 
-# Global setting for printing input/output tensors
-_print_input_output_tensors = False
-
-# Global setting for printing top k golden differences
-# None means disabled, any positive integer enables printing that many differences
-_golden_diff_topk = None
-
-
-def set_print_input_output_tensors(enabled: bool):
-    """
-    Set the global print_input_output_tensors setting.
-
-    When enabled, input tensors, output tensors, and golden tensors will be
-    printed during execution. This is useful for debugging golden comparison
-    failures.
-
-    Parameters
-    ----------
-    enabled : bool
-        Whether to enable printing of input/output tensors
-    """
-    global _print_input_output_tensors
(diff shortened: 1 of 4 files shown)


jameszianxuTT (approved):
runtime change lgtm

tools/builder/base/builder_runtime.py
```diff
@@ -33,6 +33,127 @@ class TTBuilderGoldenException(Exception):
     pass
 
 
+# Global setting for printing input/output tensors
+_print_input_output_tensors = False
```
tapspatel: im personally not a fan of setting a global this way. Would having a logging module support your use case? I have plans to add it in next week for this exact purpose of printing tensors (amongst other things)

wenbinlyuTT: I'm not sure what is this logging module approach but yeah as long as we can have a CLI argument to pytest that prints these info we're happy.
What should I do for this PR though? 80% of the python changes here are actually from Claude...

tapspatel: the python logging module is adding support for this: https://docs.python.org/3/library/logging.html
It's pretty native to python and allows users to set various levels as well: https://docs.python.org/3/library/logging.html#levels
It's also integrated into pytest: https://docs.pytest.org/en/latest/how-to/logging.html#live-logs

I'm okay if you wanna keep the helper print functions like since I'll change them to logger prints
```
def print_topk_diff(golden: torch.Tensor, output: torch.Tensor, output_index: int):
```

but I would remove the global variables and pytest fixtures.

wenbinlyuTT: Removed the feature in favor of proper logging.

PR #6521, opened 2026-01-07
D2M Single Thread Refactor

bgrady-tt:


include/ttmlir/Dialect/D2M/Utils/Utils.h
```diff
@@ -5,6 +5,8 @@
 #ifndef TTMLIR_DIALECT_D2M_UTILS_UTILS_H
 #define TTMLIR_DIALECT_D2M_UTILS_UTILS_H
 
+#include "mlir/IR/Builders.h"
```
vmilosevic, line 9: :warning: **llvm\-include\-order** :warning:
```suggestion
#include "mlir/IR/AffineMap.h"
#include "mlir/IR/Builders.h"
```

include/ttmlir/Dialect/D2M/Utils/Utils.h
```diff
@@ -5,6 +5,8 @@
...
+#include "mlir/IR/AffineMap.h"
```
vmilosevic, line 8: :warning: **llvm\-include\-order** :warning:
```suggestion

```

lib/Dialect/D2M/Transforms/DecoupleLoadStoreOpsFromCompute.cpp
```diff
@@ -0,0 +1,510 @@
...
+    }
+
+    // Determine which operand is remote (has a view/stream layout)
+    // In dma-only form, one operand typically has a view transformation
+    Value loadMemref = loadOp.getMemref();
+    Value storeMemref = storeOp.getMemref();
+
+    // Find operand indices (verification guarantees these exist)
+    auto opOperands = generic->getOpOperands();
+    auto loadOperandIt = llvm::find_if(opOperands, [&](OpOperand &opOperand) {
```
vmilosevic: :warning: **llvm\-qualified\-auto** :warning:
`` auto loadOperandIt `` can be declared as `` auto *loadOperandIt ``
```suggestion
    auto *loadOperandIt = llvm::find_if(opOperands, [&](OpOperand &opOperand) {
```

lib/Dialect/D2M/Transforms/DecoupleLoadStoreOpsFromCompute.cpp
```diff
@@ -0,0 +1,510 @@
...
+      return opOperand.get() == loadMemref;
+    });
+    auto storeOperandIt = llvm::find_if(opOperands, [&](OpOperand &opOperand) {
```
vmilosevic: :warning: **llvm\-qualified\-auto** :warning:
`` auto storeOperandIt `` can be declared as `` auto *storeOperandIt ``
```suggestion
    auto *storeOperandIt = llvm::find_if(opOperands, [&](OpOperand &opOperand) {
```

include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
```diff
@@ -668,6 +668,7 @@ def D2M_TileTilizeBlockOp : D2M_GenericRegionComputeOp<"tile_tilize_block",
 
     let arguments = (ins AnyRankedTensorOrMemRef:$input,
                          AnyRankedTensorOrMemRef:$output);
+    let results = (outs AnyRankedTensorOrMemRef:$result);
```
bgrady-tt, line 704: tilize/untilize block ops need DPS style interface with result to be bufferized properly after changes to how output buffers are created.

include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
```diff
@@ -755,9 +758,11 @@ def D2M_BlockMaskOp : D2M_GenericRegionComputeOp<"block_mask",
         Index:$logical_cols,
         TTCore_OOBValAttr:$fill_value);
 
+    let results = (outs AnyRankedTensorOrMemRef:$result);
```
bgrady-tt, line 794: Block masking ops need to have a result (DPS) to be bufferized properly.

lib/Dialect/D2M/Transforms/GridSelection.cpp
```diff
@@ -874,6 +946,83 @@ recreateGenericOp(d2m::GenericOp genericOp,
...
+                  auto elementType = tensorType.getElementType();
+                  auto shardType =
+                      RankedTensorType::get(shardShape, elementType);
+                  remoteLoadOp.getResult().setType(shardType);
+
+                  // Also update the localBuffer's defining operation's result
+                  // type to match the shard shape
+                  Value localBuffer = remoteLoadOp.getLocalBuffer();
+                  if (localBuffer) {
+                    if (auto defOp = localBuffer.getDefiningOp()) {
```
vmilosevic: :warning: **llvm\-qualified\-auto** :warning:
```suggestion
                    if (auto *defOp = localBuffer.getDefiningOp()) {
```

lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
```diff
@@ -379,11 +384,44 @@ class D2MNamedRewriterCommon {
...
-                                                 mlir::TypeRange inputs,
-                                                 mlir::TypeRange outputs) {
+  // Get grid dimension indices where multicast should happen.
+  // Multicast is needed on grid dimensions where the indexing map has a
+  // parallel iterator type. Returns empty vector if no multicast is needed.
+  static SmallVector<int64_t> getMulticastGridDims(AffineMap indexingMap,
+                                                   ArrayAttr iteratorTypes) {
+    SmallVector<int64_t> mcastGridDims;
+
+    // Iterate over the indexing map results (one per grid dimension)
```
vwellsTT: nit: . comments here

lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
```diff
@@ -398,13 +436,78 @@ class D2MNamedRewriterCommon {
...
-              ? builder.create<d2m::WaitOp>(loc, arg).getResult()
-              : builder.create<d2m::ReserveOp>(loc, arg).getResult();
-      operands.push_back(acquire);
+
+    // Process input operands - create remote_load operations using result
+    // form
+    for (size_t i = 0; i < inputs.size(); ++i) {
+      auto cbArg = block->getArgument(i);
+      auto cbType = mlir::cast<d2m::CBType>(cbArg.getType());
+      auto shardType = cbType.getUnderlying();
```
vwellsTT: nit: bad auto imo

lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
```diff
@@ -1738,19 +1966,24 @@ class TTIRToD2MPass final
...
     target.addLegalDialect<::mlir::linalg::LinalgDialect>();
+    target.addLegalDialect<::mlir::arith::ArithDialect>();
     target.addLegalDialect<mlir::tt::d2m::D2MDialect>();
     target.addLegalDialect<mlir::tt::ttcore::TTCoreDialect>();
 
     // Keep some TTIR ops legal if they don't have D2M equivalents.
     target.addLegalOp<ttir::TTNNMetalLayoutCastOp>();
 
+    // Tensor empty is used within GenericOp regions
+    target.addLegalOp<::mlir::tensor::EmptyOp>();
```
vwellsTT, line 1986: was curious, do these represent local/scratch buffers then, or what?

lib/Dialect/D2M/IR/D2MGenericRegionOps.cpp
```diff
@@ -283,59 +172,912 @@ void DMAWriteOp::getEffects(
...
+          "be present");
+    }
+    return emitOpError(
+        "must have either circular buffer or local buffer; exactly one must be "
+        "present");
+  }
+
+  // Verify result presence constraints
+  // Explicit CB form: CB present, localBuffer absent, result must NOT be
+  // present Implicit form: localBuffer present, CB absent, result MUST be
```
vwellsTT: new line? and .s

lib/Dialect/D2M/IR/D2MOps.cpp
```diff
@@ -1795,6 +1802,40 @@ mlir::AffineMap d2m::GenericOp::getIndexingMap(int64_t operandIndex) {
...
 }
 
+AffineMap d2m::GenericOp::getIndexingMapForOperand(Value operand) {
+  // Find the operand index for the given value
+  for (unsigned i = 0; i < getNumOperands(); ++i) {
+    if (getOperand(i) == operand) {
+      return getIndexingMap(i);
+    }
+  }
+  TT_assertv(false, "Operand not found in GenericOp");
```
vwellsTT: I think this sort of pattern (`assert(false); return ...;`) is exactly what llvm::unreachable is for, no?

lib/Dialect/D2M/IR/D2MOps.cpp
```diff
@@ -1795,6 +1802,40 @@ mlir::AffineMap d2m::GenericOp::getIndexingMap(int64_t operandIndex) {
...
+  return getParticipatingLoopDims(outputOperandIndex);
+}
+
+int64_t d2m::GenericOp::getOperandIndex(Value operand) {
+  for (unsigned i = 0; i < getNumOperands(); ++i) {
+    if (getOperand(i) == operand) {
+      return i;
+    }
+  }
+  TT_assertv(false, "Operand not found in GenericOp");
```
vwellsTT: ditto on unreachable here

lib/Dialect/D2M/IR/D2MOps.cpp
```diff
@@ -1795,6 +1802,40 @@ mlir::AffineMap d2m::GenericOp::getIndexingMap(int64_t operandIndex) {
   return getIndexingMapsValue()[operandIndex];
 }
 
+AffineMap d2m::GenericOp::getIndexingMapForOperand(Value operand) {
+  // Find the operand index for the given value
```
vwellsTT: nit: .s in this file too

lib/Dialect/D2M/Transforms/Allocate.cpp
```diff
@@ -1134,6 +1216,105 @@ class D2MAllocate final : public impl::D2MAllocateBase<D2MAllocate> {
...
+        TT_assert(!genericOp->getRegions().empty());
+        Region &region = genericOp->getRegions().front();
+        TT_assert(region.hasOneBlock());
+        Block &block = region.getBlocks().front();
+        Type cbArgType = block.getArgument(operandIndex).getType();
+        auto cbType = mlir::dyn_cast<d2m::CBType>(cbArgType);
+        TT_assert(cbType != nullptr);
+        Type cbUnderlyingType = cbType.getUnderlying();
+
+        // Check if a stream was inserted for this operand during this pass
```
vwellsTT, line 1240: it sort of seems like co-inserting remote_load/store w/ stream would be cleaner, but maybe that can be done in follow-ups too

lib/Dialect/D2M/Transforms/GridSelection.cpp
```diff
@@ -874,6 +946,83 @@ recreateGenericOp(d2m::GenericOp genericOp,
               clonedOp->getResult(0).setType(
                   mlir::cast<d2m::CBType>(clonedOp->getOperand(0).getType())
                       .getUnderlying());
+            } else if (auto remoteLoadOp =
+                           llvm::dyn_cast<d2m::RemoteLoadOp>(clonedOp)) {
+              // Only update result type if the result exists (implicit form)
```
vwellsTT: I assume this pass runs before lowering to explicit form, can this be an assert instead?

lib/Dialect/TTMetal/Pipelines/TTMetalPipelines.cpp
```diff
@@ -248,7 +270,7 @@ void createTTIRToTTMetalPipeline(OpPassManager &pm,
 
   // Run regular ttir to ttmetal pipelines on IR in DeviceModule.
   createTTIRToTTMetalFrontendPipeline(devicePm, options);
-  createTTIRToTTMetalMiddleendPipeline(devicePm, options);
+  createTTIRToTTMetalUnifiedMiddleendPipeline(devicePm, options);
```
vwellsTT: definitely a nit, but I prefer keeping ME vs UME name

vwellsTT (commented):
I think I need another pass on the new files, but mostly LGTM.  Definitely a ton of nits to be made about sentence comments and so on, and few other questions/suggestions so far.

lib/Dialect/D2M/Transforms/ConvertLocalLoadStoreOpsToAliasedCBs.cpp
```diff
@@ -0,0 +1,320 @@
...
+// Assumes that the operand index in the generic op equals the CB block arg
+// index.
+static Value findAssociatedCB(Operation *op, Value memrefOperand) {
+  GenericOp generic = op->getParentOfType<GenericOp>();
+  if (!generic) {
+    return Value();
+  }
+
+  // Find which operand index this memref corresponds to
+  unsigned operandIndex = UINT_MAX;
```
vwellsTT: nit: not a fan of this pattern personally, seems cleaner to use an optional and maybe llvm::enumerate for the loop

lib/Dialect/D2M/Transforms/ConvertLocalLoadStoreOpsToAliasedCBs.cpp
```diff
@@ -0,0 +1,320 @@
...
+      Value localBuffer = remoteStore.getLocalBuffer();
+      if (!localBuffer) {
+        remoteStore.emitWarning(
+            "remote_store does not have a local buffer operand, skipping "
+            "conversion");
+        continue;
+      }
+
+      // Find the memref.alloc that produces the local buffer being stored
+      memref::AllocOp allocOp = findAllocOp(localBuffer);
```
vwellsTT, line 367: so this alloc is generated via the tensor::EmptyOps from earlier? I wonder if it'd be better to have custom op + custom bufferization instead in the long run?

lib/Dialect/D2M/Transforms/GenerateOuterLoops.cpp
```diff
@@ -0,0 +1,255 @@
...
+    loopedBlock->addArguments(
+        region.getArgumentTypes(),
+        SmallVector<mlir::Location>(region.getArgumentTypes().size(),
+                                    generic.getLoc()));
+    rewriter.setInsertionPointToStart(loopedBlock);
+    scf::LoopNest loopNest = buildLoopNest(
+        rewriter, generic.getLoc(), loopBounds, regionBlock, loopedBlock);
+
+    // Mark all loops in the nest with an attribute to prevent re-processing
+    for (scf::ForOp loop : loopNest.loops) {
```
vwellsTT, line 223: this seems unintuitive to me, all loops in the nest are "outer"? As opposed to what then?

lib/Dialect/D2M/Transforms/LowerLoadStoreOpsToExplicitCBForm.cpp
```diff
@@ -0,0 +1,543 @@
...
+  for (Value operand : definingOp->getOperands()) {
+    if (auto reserveOp = findReserveOp(operand)) {
+      return reserveOp;
+    }
+  }
+
+  return nullptr;
+}
+
+// Recognize and simplify the load-store idiom where a remote_load and
```
vwellsTT, line 112: is it not possible to prevent emitting a remote_load/store on local buffers?

lib/Dialect/D2M/Transforms/GenerateOuterLoops.cpp
bgrady-tt: The outer loop nests implied by the genericOp block factors. I guess I could change this to be "d2m.generic_blocking_loops"

lib/Dialect/D2M/Transforms/ConvertLocalLoadStoreOpsToAliasedCBs.cpp
bgrady-tt: Originally I used a custom op (`d2m.acquire_buffer`), but it is just not compatible with how linalg and other builtin MLIR passes treat alloc ops. They will freely substitute memref.alloc or tensor.empty. This keeps all allocations as `memref.alloc` even those done within the genericOp body.

bgrady-tt: fwiw this was I believe the consensus in the last meeting; all allocations including those in the generic regions would be done as `memref.alloc`

lib/Dialect/D2M/Transforms/ConvertLocalLoadStoreOpsToAliasedCBs.cpp
bgrady-tt: I can refactor this 👍

lib/Dialect/TTMetal/Pipelines/TTMetalPipelines.cpp
bgrady-tt: Ok I can rename it to keep it consistent.

lib/Dialect/D2M/Transforms/LowerLoadStoreOpsToExplicitCBForm.cpp
bgrady-tt: I _like_ this idiom, because it allows expressing simple passthrough datamovement (input tensor -> output tensor) without having to worry about local/remote distinction. It is always valid, and the lowering code just does the right thing.

bgrady-tt: for example, code generating a datamovement style generic op can always just generate a pair of load and store ops. It doesn't need to special case for every combo of local/remote, remote/local, remote/remote

lib/Dialect/D2M/Transforms/GridSelection.cpp
bgrady-tt: I can make this an assert.

lib/Dialect/D2M/Transforms/Allocate.cpp
bgrady-tt: Yeah this logic needs to be cleaned up post Streams 2.0 . It will probably change a great deal.

lib/Dialect/D2M/IR/D2MOps.cpp
bgrady-tt: will fix this to use unreachable

lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
bgrady-tt: yes, they represent both (A) memory allocation/declaration that must be fulfilled in allocator post-bufferization (B) the buffer itself (which for now will always be lowered to a CB).

lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
bgrady-tt: ok I can make this explicit

lib/Dialect/D2M/Transforms/ConvertLocalLoadStoreOpsToAliasedCBs.cpp
bgrady-tt: ok I can refactor this.

lib/Dialect/D2M/Transforms/ConvertLocalLoadStoreOpsToAliasedCBs.cpp
vwellsTT: yeah, that's right--hard to keep straight :)

lib/Dialect/D2M/Transforms/LowerLoadStoreOpsToExplicitCBForm.cpp
vwellsTT: hmm, I guess that's consistent with how we do like to_layouts or whatever, okay

include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
```diff
@@ -861,165 +866,6 @@ def D2M_TileTransposeOp : D2M_GenericRegionComputeOp<"tile_transpose", D2M_TileC
...
-        return mlir::cast<::mlir::tt::ttcore::MemorySpaceAttr>(getSrcMemRefType().getMemorySpace()).getValue();
-      }
-      ::mlir::tt::ttcore::MemorySpace getDstMemorySpace() {
-        return mlir::cast<::mlir::tt::ttcore::MemorySpaceAttr>(getDstMemRefType().getMemorySpace()).getValue();
-      }
-
-    }];
-}
-
-
```
nsmithtt, line 1022: 🙌

include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
```diff
@@ -1167,6 +1011,361 @@ def D2M_DMAWaitOp : D2M_GenericRegionDatamovementOp<"dma_wait", [MemoryEffects<[
     let assemblyFormat = [{ $mem_tx attr-dict }];
 }
 
+//===----------------------------------------------------------------------===//
+// D2M Generic Region Remote Load/Store Ops (Based on Affine Load/Store)
+//===----------------------------------------------------------------------===//
+
+def D2M_RemoteLoadOp : D2M_GenericRegionDatamovementOp<"remote_load",
+  [ AttrSizedOperandSegments
```
nsmithtt, line 1020: Probably should have DPS trait?

include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
```diff
@@ -1636,6 +1835,20 @@ def D2M_IterIndexOp : D2M_IndexOp<"iter_index"> {
     }];
 }
 
+def D2M_IMIndexOp : D2M_IndexOp<"im_index"> {
+    let summary = "Indexing Map Index op.";
+    let description = [{
+      Return the index for the given indexing map dimension. This op represents
```
nsmithtt: I'm missing the nuance here, can you provide an example of how this is different than IterIndex?

lib/Dialect/D2M/IR/D2MOps.cpp
bgrady-tt: fixed

include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
bgrady-tt: `IMIndex` corresponds to fully block-factor aware index lowering. The example should clarify :) 

`IterIndex` is a more literal replacement with the outer loop variable. This is needed to properly support implicit broadcast, which uses `iter_index` to _directly reference_ the raw outer loop variable to check for the first loop iteration. If the blocked variant is used instead, implicit broadcast breaks.

```
Example:
indexing map (d0,d1) -> (d0,d1)
block factors are [2,2]
im_index(0) ---lowers to---> 2*core_index(0) + %outer_loop_iv0
iter_index(0) ---lowers to---> %outer_loop_iv0
```

include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
bgrady-tt: It isn't DPS always though; in explicit form it is side-effecting?

nsmithtt: DPS is an op interface that establishes these semantics and has a first class relationship with one-shot bufferization.  This is a good bit of documentation explaining what the interface is and why it's useful: https://mlir.llvm.org/docs/Bufferization/#destination-passing-style

bgrady-tt: I can add this to both remote load and store ops. Hopefully it doesn't perturb bufferization implementation much.

include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
nsmithtt: I see, so it's a shorthand for the above expression.  I think the name and description is a little confusing, I'm wondering if we should just copy cuda names here.  Could we call it `block_index` or something?

- `iter_index`: Local block index (current block processed by this core)
- `block_index`: Global block index (renamed from `im_index`), description defines it in terms of `block_dim * core_index + iter_index`.
- `core_index`: Core's index in grid

Perhaps as a follow on we should be even more explicit and go with:
- `local_block_index`: renamed from `iter_index`
- `global_block_index`: renamed from `im_index`
- `core_index`: Core's index in grid

On this note, it might be nice to also have `block_dim` operation so that you can programmatically get the block factor in the kernel and manually expand this.

nsmithtt (approved):
Awesome work, this is a huge step in the right direction for d2m.  I didn't have time to review most of the passes, but signing off on the high level design and new ops that were introduced.  Would like to still get your thoughts on the im_index op naming.  I'm thinking that since cuda and triton have adopted the block concept, perhaps we should follow suit.

include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
bgrady-tt: block index is a better name. I will fix this before merging.

bgrady-tt: Agree on local vs global distinction for indexing as well. Something to refine later for sure.

include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
bgrady-tt: added `DestinationStyleInterface` to both ops this morning 👍

include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
bgrady-tt: Renamed `IMIndex` to `BlockIndex` this morning 👍

PR #6540, opened 2026-01-09
[d2m] enable FPU lowering for binary add, sub, mul

ckaravasilisTT:
### Problem description
Currently, `add`/`sub`/`mul` are lowered as SFPU ops with both input operands stored in Dst. We want to use the FPU for these ops for increased performance and future fusion opportunities.

### What's changed
These binary ops now use the FPU as the default path. Exceptions to this are when Rhs is a scalar, and when both inputs come from Dst, where the SFPU is used. When only one input comes from Dst, `binary_dest_reuse_tiles` (FPU) executes the op and the Dst input is used directly. These ops will initially be excluded from fusion until an initial version of FPU fusion is implemented.

### DIFF
--- a/include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
+++ b/include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
@@ -101,7 +101,7 @@ class D2M_GenericRegionComputeBinaryWithScalarOptionDstOp<string mnemonic, list<
 }
 
 // FPU binary ops that support tile-tile OR tile-scalar operations
-class D2M_GenericRegionComputeFPUBinaryWithScalarOptionOp<string mnemonic, list<Trait> traits = []> :
+class D2M_GenericRegionComputeFPUOrSFPUBinary<string mnemonic, list<Trait> traits = []> :
     D2M_GenericRegionComputeOp<mnemonic, traits> {
 
   let extraClassDeclaration = [{
@@ -151,7 +151,7 @@ defvar D2M_TileComputeOpTraitBundle = [
 // Type constraint that accepts either a Tile or a scalar value (float/int)
 def TileOrScalar : AnyTypeOf<[TTCore_Tile, AnyFloat, AnyInteger]>;
 
-def D2M_TileAddOp : D2M_GenericRegionComputeFPUBinaryWithScalarOptionOp<"tile_add", [Pure, D2M_SkipOpEltwiseFusionTrait]> {
+def D2M_TileAddOp : D2M_GenericRegionComputeFPUOrSFPUBinary<"tile_add", [Pure, D2M_SkipOpEltwiseFusionTrait]> {
     let summary = "D2M Tile Add Op";
     let description = [{
         The `tile_add` operation adds two tiles element-wise on the FPU,
@@ -163,7 +163,7 @@ def D2M_TileAddOp : D2M_GenericRegionComputeFPUBinaryWithScalarOptionOp<"tile_ad
     let results = (outs TTCore_Tile:$result);
 }
 
-def D2M_TileSubOp : D2M_GenericRegionComputeFPUBinaryWithScalarOptionOp<"tile_sub", [Pure, D2M_SkipOpEltwiseFusionTrait]> {
+def D2M_TileSubOp : D2M_GenericRegionComputeFPUOrSFPUBinary<"tile_sub", [Pure, D2M_SkipOpEltwiseFusionTrait]> {
     let summary = "D2M Tile Sub Op";
     let description = [{
         The `tile_sub` operation subtracts two tiles element-wise on the FPU,
@@ -175,7 +175,7 @@ def D2M_TileSubOp : D2M_GenericRegionComputeFPUBinaryWithScalarOptionOp<"tile_su
     let results = (outs TTCore_Tile:$result);
 }
 
-def D2M_TileMulOp : D2M_GenericRegionComputeFPUBinaryWithScalarOptionOp<"tile_mul", [Pure, D2M_SkipOpEltwiseFusionTrait]> {
+def D2M_TileMulOp : D2M_GenericRegionComputeFPUOrSFPUBinary<"tile_mul", [Pure, D2M_SkipOpEltwiseFusionTrait]> {
     let summary = "D2M Tile Mul Op";
     let description = [{
         The `tile_mul` operation multiplies two tiles element-wise on the FPU,
--- a/lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
+++ b/lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
@@ -418,11 +418,12 @@ using ComputeOpMap = OpMap<
   std::pair<d2m::TileLezOp,         std::pair<ttkernel::LezTileInitOp,             ttkernel::LezTileOp>>,
   std::pair<d2m::TileTypecastOp,    std::pair<ttkernel::TypecastTileInitOp,        ttkernel::TypecastTileOp>>,
 
-  // Elementwise SFPU Binary (can also handle scalar operands).
-  // Add/Sub/Mul fall back to SFPU if FPU rewriter fails.
+  // Elementwise FPU Binary with SFPU fallback.
   std::pair<d2m::TileAddOp,         std::pair<ttkernel::AddBinaryTilesInitOp,      ttkernel::AddBinaryTilesOp>>,
   std::pair<d2m::TileSubOp,         std::pair<ttkernel::SubBinaryTilesInitOp,      ttkernel::SubBinaryTilesOp>>,
   std::pair<d2m::TileMulOp,         std::pair<ttkernel::MulBinaryTilesInitOp,      ttkernel::MulBinaryTilesOp>>,
+
+  // Elementwise SFPU Binary.
   std::pair<d2m::TileBitwiseAndOp,  std::pair<ttkernel::BinaryBitwiseTileInitOp,   ttkernel::BitwiseAndBinaryTilesOp>>,
   std::pair<d2m::TileBitwiseOrOp,   std::pair<ttkernel::BinaryBitwiseTileInitOp,   ttkernel::BitwiseOrBinaryTilesOp>>,
   std::pair<d2m::TileBitwiseXorOp,  std::pair<ttkernel::BinaryBitwiseTileInitOp,   ttkernel::BitwiseXorBinaryTilesOp>>,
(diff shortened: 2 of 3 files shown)


lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
```diff
@@ -802,6 +802,76 @@ class D2MSFPUOpsRewriter : public OpConversionPattern<ConcreteOp> {
...
+        rewriter.create<ttkernel::AddUnaryTileOp>(loc, dstIdx, scalarParam);
+      }
+      // TODO: Add other scalar ops here (sub, mul)
+    } else {
+      // Tile-tile operation - use FPU path (reads directly from CBs)
+      auto cbA = getCB(rewriter, op.getLhs());
+      auto cbB = getCB(rewriter, op.getRhs());
+      auto outCB = getOutCB(rewriter, op);
+
+      auto insertionPoint = rewriter.getInsertionPoint();
```
vmilosevic: :warning: **google\-readability\-todo** :warning:
missing username/bug in TODO
```suggestion
      // TODO(nsmith): Add other scalar ops here (sub, mul)
```

lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
```diff
@@ -802,6 +802,76 @@ class D2MSFPUOpsRewriter : public OpConversionPattern<ConcreteOp> {
...
+                                              adaptor.getRhs(), dstIdx);
+      }
+      // TODO: Add other FPU binary ops here (sub, mul)
+    }
+
+    rewriter.eraseOp(op);
+    return success();
+  }
+};
+} // namespace
```
vmilosevic, line 983: :warning: **google\-readability\-todo** :warning:
missing username/bug in TODO
```suggestion
      // TODO(nsmith): Add other FPU binary ops here (sub, mul)
```

include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
```diff
@@ -100,6 +100,27 @@ class D2M_GenericRegionComputeBinaryWithScalarOptionDstOp<string mnemonic, list<
   }];
 }
 
+// FPU binary ops that support tile-tile OR tile-scalar operations
+class D2M_GenericRegionComputeFPUBinaryWithScalarOptionOp<string mnemonic, list<Trait> traits = []> :
+    D2M_GenericRegionComputeOp<mnemonic, traits> {
+
+  let extraClassDeclaration = [{
+    bool supportsTileOrScalarRhs() { return true; }
```
nsmithtt, line 108: The semantics of this are a little weird because FPU doesn't support scalar operands, only the SFPU does.  Is this all or nothing or can we support conditionally lowering to either the FPU or SFPU depending on the constraints / context?

nsmithtt (commented):
Only reviewed the dialect changes so far, one comment inline regarding scalar operand support.

include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
ckaravasilisTT: I agree the semantics are a bit weird. It's basically ops that _prefer_ the FPU, but fall back to SFPU when needed. Conditional lowering takes place in `D2MToTTKernel` via `D2MFPUBinaryWithScalarRewriter`. I can add more comments in the `.td` file but I'm not sure how to make it cleaner semantically except breaking it into two ops (`TileAddOp`, `TileAddScalarOp`).

lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
```diff
@@ -1663,17 +1879,19 @@ void populateD2MToTTKernelPatterns(
                ttkernel::D2MSFPUOpsRewriter<d2m::TileLezOp>,
                ttkernel::D2MSFPUOpsRewriter<d2m::TileTypecastOp>,
 
+               // FPU Binary with scalar option.
+               ttkernel::D2MFPUBinaryWithScalarRewriter<d2m::TileAddOp>,
+               ttkernel::D2MFPUBinaryWithScalarRewriter<d2m::TileSubOp>,
+               ttkernel::D2MFPUBinaryWithScalarRewriter<d2m::TileMulOp>,
```
nsmithtt: So instead of having an uber-rewriter that does both FPU and SFPU, you can keep the `ttkernel::D2MSFPUOpsRewriter<d2m::TileAddOp>` cases below.  The patterns here should take precedence because they appear earlier in this list, but if they return a rewrite failure then the SFPU ones below will be matched as fallback.

ckaravasilisTT: Updated to follow this approach

test/ttmlir/Conversion/D2MToTTKernel/sfpu_op_placement_after_operands.mlir
```diff
@@ -1,3 +1,4 @@
+// UNSUPPORTED: true
```
nsmithtt: Do we plan on fixing these before this PR lands?

test/python/golden/test_ttir_eltwise_fusion.py
```diff
@@ -571,6 +572,7 @@ def where_with_unary_chains(
 @pytest.mark.parametrize("shape", [(128, 128)])
 @pytest.mark.parametrize("dtype", [torch.bfloat16], ids=["bf16"])
 @pytest.mark.parametrize("target", ["ttmetal"])
+@pytest.mark.skip(reason="TODO: Binary ops FPU fusion is not supported yet")
```
nsmithtt: Do we plan on fixing these before this PR lands?

nsmithtt (commented):
Overall looks good to me! Just want to know what our plan is for the tests that are now marked as skip.  Also we probably need a new test or two to make sure we're exercising the new codepaths you added.

test/ttmlir/Conversion/D2MToTTKernel/sfpu_op_placement_after_operands.mlir
ckaravasilisTT: Yep, I will be updating these tests (and adding new ones) shortly.

test/python/golden/test_ttir_eltwise_fusion.py
ckaravasilisTT: Fusion tests may have to be skipped until an updated fusion plan that includes the FPU lands. TBD

brnorris03:
Would it be possible to land the ttkernel changes first? (tt-lang waiting on that part)

ckaravasilisTT:
> Would it be possible to land the ttkernel changes first? (tt-lang waiting on that part)

I think it can land as is. I can add further lit tests at a later time. @brnorris03

include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
wenbinlyuTT: I think we can replace the `FPU` in the name to `FPUOrSFPU` so that it's clear this class of ops can select different code paths? It aligns with the `getOperandsLoadFromDstRegister()` function below.

In the future we may also choose the SFPU path if we really need to preserve f32 accuracy or for something similar, which I imagine will also be integrated here. Therefore we may even drop `WithScalarOption` from the name so that it doesn't read like it's dedicated to a single condition.

lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
```diff
@@ -386,16 +419,17 @@ using ComputeOpMap = OpMap<
   std::pair<d2m::TileTypecastOp,    std::pair<ttkernel::TypecastTileInitOp,        ttkernel::TypecastTileOp>>,
 
   // Elementwise SFPU Binary (can also handle scalar operands).
+  // Add/Sub/Mul fall back to SFPU if FPU rewriter fails.
   std::pair<d2m::TileAddOp,         std::pair<ttkernel::AddBinaryTilesInitOp,      ttkernel::AddBinaryTilesOp>>,
+  std::pair<d2m::TileSubOp,         std::pair<ttkernel::SubBinaryTilesInitOp,      ttkernel::SubBinaryTilesOp>>,
+  std::pair<d2m::TileMulOp,         std::pair<ttkernel::MulBinaryTilesInitOp,      ttkernel::MulBinaryTilesOp>>,
```
wenbinlyuTT, line 424: Nit: move these to their own section called `Elementwise FPU binary with SFPU fallback`, and the comment about scalar can be removed.

lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
```diff
@@ -811,6 +848,139 @@ class D2MSFPUOpsRewriter : public OpConversionPattern<ConcreteOp> {
...
+    bool rhsFromDst = operandFromDst(op.getRhs());
+
+    if (!lhsFromDst && !rhsFromDst) {
+      // Both from CB: standard FPU path
+      return emitFPUTiles(rewriter, op, adaptor, loc);
+    }
+
+    if (lhsFromDst && rhsFromDst) {
+      // Both from DST: fallback to SFPU
+      return failure();
```
wenbinlyuTT, line 880: Nit: personally I prefer to put all failing scenarios to the front.

test/ttmlir/Conversion/D2MToTTKernel/single_tile_ops.mlir
```diff
@@ -1,3 +1,4 @@
+
```
wenbinlyuTT: Nit: extra line

test/ttmlir/Conversion/D2MToTTKernel/single_tile_ops.mlir
```diff
@@ -181,12 +182,12 @@ module {
...
         // CHECK-NOT: d2m.tile_bcast
-        // CHECK: ttkernel.init_sfpu
+        // CHECK: ttkernel.binary_op_init_common
         // CHECK: ttkernel.unary_bcast_init(%{{.*}}, %{{.*}}, <scalar>)
         // CHECK: ttkernel.unary_bcast(%{{.*}}, %{{.*}}, %{{.*}}, <scalar>)
         %0 = "d2m.tile_bcast"(%arg0) <{bcast_type = #d2m<tile_bcast_type scalar>}> : (!ttype_f32) -> !ttype_f32
-        // CHECK: ttkernel.mul_binary_tile_init
-        // CHECK: ttkernel.mul_binary_tile
+        // CHECK: ttkernel.binary_dest_reuse_tiles_init(%{{.*}}, <mul>, <dest_to_srca>)
+        // CHECK: ttkernel.binary_dest_reuse_tiles(%{{.*}}, %{{.*}}, %{{.*}}, <mul>, <dest_to_srca>)
```
wenbinlyuTT, line 216: We will also need a test that emits `dest_to_srcb`

ckaravasilisTT: Added a similar test for `dest_to_srcb`

lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
ckaravasilisTT: Updated!

include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
ckaravasilisTT: I also like this naming a bit better. Updated.

nsmithtt:
+1 to Wenbin's comments, otherwise LGTM

PR #6545, opened 2026-01-09
outer permute, part 2: decompose complex permutes

anuragsingh-tt:
### Problem description
D2M has two kinds of permutes:
- inner permute (swaps last two dimensions)
- outer permute (swaps dimensions but keeps the last dimension fixed).

A permute that performs both needs to be split up. 

### What's changed
Created DecomposeComplexPermute pass to break down complex permutes. At most, the permutation will be decomposed into three permutes: Outer1 -> Inner -> Outer2. Added an attribute to these new permutes to prevent existing TTIR folding (which would undo these decompositions). 

### Checklist
- test/ttmlir/Dialect/D2M/decompose_complex_permute.mlir: testing DecomposeComplexPermute pass outcome
- additional test_permute cases in test_metal_tms.py. Note that more complex tests (e.g. inner, outer, then inner) do not work end to end due to downstream issues stacking views on top of streams; this will be revisited after the stream refactor.

### DIFF
--- a/lib/Dialect/D2M/Transforms/DecomposeComplexPermute.cpp
+++ b/lib/Dialect/D2M/Transforms/DecomposeComplexPermute.cpp
@@ -39,93 +39,87 @@ llvm::SmallVector<int64_t> createInnerPermute(size_t rank) {
   return perm;
 }
 
-// Create an outer permute that swaps positions i and j, keeping last position
-// fixed. Acts on positions [0..n-2], keeps last position fixed.
-llvm::SmallVector<int64_t> createOuterSwap(size_t rank, size_t i, size_t j) {
-  llvm::SmallVector<int64_t> perm(rank);
-  int64_t *data = perm.data();
-  for (size_t k = 0; k < rank; ++k) {
-    data[k] = (k == i) ? j : (k == j ? i : k);
+// Check if permutation is identity.
+bool isIdentity(llvm::ArrayRef<int64_t> perm) {
+  for (size_t i = 0; i < perm.size(); ++i) {
+    if (perm[i] != static_cast<int64_t>(i)) {
+      return false;
+    }
   }
-  return perm;
+  return true;
 }
 
-// Apply permutation p to array arr: result[i] = arr[p[i]].
-llvm::SmallVector<int64_t> applyPerm(llvm::ArrayRef<int64_t> arr,
-                                     llvm::ArrayRef<int64_t> p) {
-  llvm::SmallVector<int64_t> result;
-  for (size_t i = 0; i < p.size(); ++i) {
-    result.push_back(arr[p[i]]);
-  }
-  return result;
-}
-
-// Decompose a complex permutation into a sequence of inner and outer permutes.
-// Inner permute: only swaps the last two positions (n-2, n-1).
+// Decompose a complex permutation into at most, a sequence of Outer -> Inner ->
+// Outer permute. Inner permute: only swaps the last two positions (n-2, n-1).
 // Outer permute: acts on positions [0..n-2], keeps last position fixed.
 //
-// Algorithm: greedy left-to-right placement.
-// 1. For each position i from 0 to n-3:
-//    - If current[i] != target[i], find where target[i] is.
-//    - If it's at position n-1, do an inner swap first.
-//    - Then do an outer swap to place it correctly.
-// 2. After fixing positions 0..n-3, fix last two with an inner swap if needed.
 llvm::SmallVector<llvm::SmallVector<int64_t>>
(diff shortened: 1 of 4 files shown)


lib/Dialect/D2M/Transforms/DecomposeComplexPermute.cpp
```diff
@@ -0,0 +1,184 @@
+// SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
```
vwellsTT, line 1: nit: year

lib/Dialect/D2M/Transforms/DecomposeComplexPermute.cpp
```diff
@@ -0,0 +1,184 @@
...
+  for (size_t i = 0; i < rank - 2; ++i) {
+    perm[i] = i;
+  }
+  perm[rank - 2] = rank - 1;
+  perm[rank - 1] = rank - 2;
+  return perm;
+}
+
+// Check if permutation is identity.
+bool isIdentity(llvm::ArrayRef<int64_t> perm) {
```
vwellsTT, line 43: I think you could do:
```
bool isIdentity(llvm::ArrayRef<int64_t> perm) {
  return llvm::equal(perm, llvm::seq<int64_t>(0, perm.size()));
}
```
which is simpler, but less clear; debatable whether it's better or not

lib/Dialect/D2M/Transforms/DecomposeComplexPermute.cpp
```diff
@@ -0,0 +1,184 @@
...
+  for (size_t i = 0; i < perm.size(); ++i) {
+    if (perm[i] != static_cast<int64_t>(i)) {
+      return false;
+    }
+  }
+  return true;
+}
+
+// Decompose a complex permutation into at most, a sequence of Outer -> Inner ->
+// Outer permute. Inner permute: only swaps the last two positions (n-2, n-1).
```
vwellsTT, line 53: honestly I feel like this is becoming a bit over-stated, seem to have read definition of inner permute like 5x so far (including previously in this same file).  I guess it's better to overcommunicate than undercommunicate, idk

lib/Dialect/D2M/Transforms/DecomposeComplexPermute.cpp
```diff
@@ -0,0 +1,184 @@
...
+// Outer permute: acts on positions [0..n-2], keeps last position fixed.
```
vwellsTT, line 54: To make sure I understand complex case, if I want say [2,3,1,0], the sequence is:
outer permute [0,1,2,3] -> [2,1,0,3]
inner permute [2,1,0,3] -> [2,1,3,0]
outer permute [2,1,3,0] -> [2,3,1,0]
is that right?

lib/Dialect/D2M/Transforms/DecomposeComplexPermute.cpp
```diff
@@ -0,0 +1,184 @@
...
+decomposePermutation(llvm::ArrayRef<int64_t> permutation) {
+  llvm::SmallVector<llvm::SmallVector<int64_t>> result;
+  size_t n = permutation.size();
+
+  // Outer permute: last dimension is identity.
+  if (permutation[n - 1] == static_cast<int64_t>(n - 1)) {
+    return result;
+  }
+
+  // Case 1: P[n-2] == n-1, can do with just Outer + Inner.
```
vwellsTT, line 66: this seems to me like it's just an inner permute, am I missing something?
either way, nit: don't like Case 1: and then General Case, should stick with Case N, or maybe "Edge Case" + "General Case"

lib/Dialect/TTIR/IR/TTIROps.cpp
```diff
@@ -4392,6 +4392,10 @@ static mlir::OpFoldResult foldIdentityPermute(mlir::tt::ttir::PermuteOp op) {
 // If the producer is a PermuteOp we can compose the permutation attributes
 // into `op`, and set the input to the producers input.
 static mlir::OpFoldResult foldConsecutivePermute(mlir::tt::ttir::PermuteOp op) {
+  // Don't fold decomposed permutes - they were intentionally split.
+  if (op->hasAttr("decomposed")) {
```
vwellsTT, line 4396: MLIR makes this type of ad hoc attr easy to use, but I think we should be a bit more conservative about it; in the past, core team has suggested we should at least define these attrs in some dialect specific tablegen and access via variables instead of string literals (see `shouldHoistAttr` for an example)

vwellsTT (approved):
Core logic seems good to me, some comments inline

PR #6554, opened 2026-01-12
Op by op infra update

sdjukicTT:
### Problem description
Op by op infra needed updates to support new tt-xla use https://github.com/tenstorrent/tt-xla/pull/2796

### What's changed
- removed OpView from OpWrapper class, leaving only op as a string
- added extract_ops_from_module and execute_extracted_op methods
- added support for storing multiple origin models for ops
- moved generating sub_modules from ModuleSplitter split method because we don't always want submodules right away (like in extract_ops_from_module)
- added optional debug prints

### Checklist
- this infra is used in a test in tt-xla repo

### DIFF
--- a/tools/op-by-op-infra/op_by_op_infra/utils.py
+++ b/tools/op-by-op-infra/op_by_op_infra/utils.py
@@ -20,9 +20,18 @@ from ttmlir.ir import (
     Type,
 )
 
-_LOC_LINE_PATTERN = re.compile(r"^#loc.*$", re.MULTILINE)
-_LOC_INLINE_PATTERN = re.compile(r"\s*loc\((?:[^()]*(?:\([^()]*\))*)*\)")
-_SDY_MESH_PATTERN = re.compile(r"\s*sdy.mesh.*$", re.MULTILINE)
+# Patterns for preprocessing MLIR module strings
+_LOC_LINE_PATTERN = re.compile(r"^#loc.*$", re.MULTILINE)  # Match #loc lines
+_LOC_INLINE_PATTERN = re.compile(
+    r"\s*loc\((?:[^()]*(?:\([^()]*\))*)*\)"
+)  # Match inline loc(...) annotations
+_SDY_MESH_PATTERN = re.compile(r"\s*sdy.mesh.*$", re.MULTILINE)  # Match sdy.mesh lines
+
+
+def _replace_mlir_identifier(text: str, old_name: str, new_name: str) -> str:
+    """Replace SSA value name (e.g., %arg0), avoiding partial matches like %arg01."""
+    pattern = re.escape(old_name) + r"(?=[^a-zA-Z0-9_]|$)"
+    return re.sub(pattern, new_name, text)
 
 
 @dataclass(frozen=True)
@@ -114,6 +123,7 @@ class OpWrapper:
 
     Extracts and caches all necessary information from the live OpView at construction
     time, so the wrapper can outlive the MLIR context.
+    If the op has constant operands, they are stored, to later be added to the module.
     """
 
     # ----- Public methods and properties -----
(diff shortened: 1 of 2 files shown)


tools/op-by-op-infra/op_by_op_infra/utils.py
```diff
@@ -20,6 +20,10 @@
     Type,
 )
 
+_LOC_LINE_PATTERN = re.compile(r"^#loc.*$", re.MULTILINE)
```
ajakovljevicTT: Write comments for what this catches.

test/python/op_by_op/test_mlir_module_splitter.py
```diff
@@ -9,31 +9,31 @@
 
 def test_shlo_module_split(shlo_module_str: str):
     splitter = MLIRModuleSplitter()
-    sub_modules = splitter.split(shlo_module_str)
-    sub_ops = splitter.sub_ops
+    sub_ops = splitter.split(shlo_module_str)
+    sub_modules = splitter.sub_modules
 
     assert len(sub_ops) == len(sub_modules) == 3
```
acicovicTT: Suggestion: expected number of modules should be passed as a parameter to tests functions in this file, since it is not obvious where these constants came from. If you define them in `fixtures.py` alongside paths to `.mlir` source files, it will be a little cleaner.

tools/op-by-op-infra/op_by_op_infra/utils.py
```diff
@@ -114,37 +123,68 @@ def __init__(
...
+
+        result_mapping = {}
+        for i, result in enumerate(op.results):
+            original_name = result.get_name()
+            standardized_name = f"%res{i}"
+            result_mapping[original_name] = standardized_name
+
+        # Replace operand and result names in op_string
+        for original, standardized in {**result_mapping, **operand_mapping}.items():
+            pattern = re.escape(original) + r"(?=[^a-zA-Z0-9_]|$)"
```
acicovicTT: You can extract this one as a constant as well. +1 on Andrej's suggestion to make a comment for each regex.

tools/op-by-op-infra/op_by_op_infra/workflow.py
```diff
@@ -74,3 +76,67 @@ def run_op_by_op_workflow(
...
+        compile_only, debug_print=debug_print
+    )
+    execution_results = []
+
+    for op in workflow_internal.progress_bar(ops, desc="Executing submodules..."):
+        sub_module = op.as_module()
+        execution_result = executor.execute(sub_module)
+        execution_results.append(execution_result)
+
+    return workflow_internal.convert_results_to_pydantic_models(
```
acicovicTT, line 140: Should we also do this in `split_and_execute` and similar functions in `worklow_internal.py`?

acicovicTT (commented):
Approved with suggestions.

tools/op-by-op-infra/op_by_op_infra/utils.py
sdjukicTT: Added comments.

test/python/op_by_op/test_mlir_module_splitter.py
sdjukicTT: Extracted and defined them in `fixtures.py`.

tools/op-by-op-infra/op_by_op_infra/utils.py
sdjukicTT: This one depends on original name, so it can't be constant, but I extracted logic into a method to make this more readable.

tools/op-by-op-infra/op_by_op_infra/workflow.py
sdjukicTT: It is not known whether `split_and_execute` will always need to make results pydantic models, so I would leave it as is for now.

Issue #6537, opened 2026-01-09
Big conv2d fails optimizer validation


ppadjinTT:
### Describe the bug

When trying to pass a larger convolution through the optimizer with optimization level 1, this error appears:

`<unknown>:0: error: loc("convolution.5"): OperationValidationAndFallback: Operation ttnn.conv2d failed validation (original error: OutOfMemoryError - Not enough L1 memory). No fallback configuration worked (tested up to 10000 combinations).` This error **does not appear** when optimization level is 0.

This is the configuration of the given convolution:
- input: (2, 960, 128, 128) bf16
- weight: (320, 960, 3, 3) bf16
- padding: 1
- stride: 1
- dilation: 1
- groups: 1

## From TTIR
Run the attached repro.mlir file with following command: 
```
TTMLIR_RUNTIME_LOGGER_LEVEL=Trace TTRT_LOGGER_LEVEL=Debug  ./build/bin/ttmlir-opt --ttir-to-ttnn-backend-pipeline="system-desc-path=$SYSTEM_DESC_PATH optimization-level=1" --mlir-print-ir-after-all  -o out.mlir repro.mlir &> repro.log
```

## From tt-xla
```
import torch
import torch.nn as nn
import torch_xla
import torch_xla.core.xla_model as xm
import torch_xla.runtime as xr


def run_conv2d_repro():
    xr.set_device_type("TT")

    # Match optimization level from sdxl-pipeline.py
    torch_xla.set_custom_compile_options({"optimization_level": 1})

    # Create conv2d with exact parameters from failing SDXL op
    model = nn.Conv2d(
        in_channels=960,
        out_channels=320,
        kernel_size=3,
        stride=1,
        padding=1,
        dilation=1,
        groups=1,
        bias=False,  # No bias in the failing op
    )
    model = model.to(torch.bfloat16)
    model = model.eval()

    # Create input with exact shape from failing op
    # input: tensor<2x960x128x128xbf16>
    input_tensor = torch.randn(2, 960, 128, 128, dtype=torch.bfloat16)

    # Get torch reference output (on CPU)
    print(f"Input shape: {input_tensor.shape}")
    print(f"Weight shape: {model.weight.shape}")
    print(f"Computing torch reference output...")
    with torch.no_grad():
        torch_output = model(input_tensor)
    print(f"Output shape: {torch_output.shape}")

    # Compile with TT backend
    print(f"Compiling model with backend='tt'...")
[...]

Issue #6433, opened 2025-12-23
[Optimizer] Conv2d no fallback configuration L1 OOM


bmalesevicTT:
When running optimization_level=1 on models with large conv2d e.g. `detr_object_detection_pytorch_resnet_50`, fallback pass failed to find a fitting configuration. This is because L1 is always OOM with proper configurations, as `conv2d_slice_config` set to L1 is crossing the L1 limit and is never changed. 
The specific conv2d op is extracted to a unit test attached here for repro, with the layouts that model compilation with `optimization_level=1` failed to find a fallback for.
For debug testing, to see all the fallback configurations (running the lit test might take too long, because of checking many fallback combinations, but it eventually fails): 
```
LOGGER_LEVEL=trace ttmlir-opt --ttcore-register-device --ttnn-operation-validation-and-fallback conv2d_l1_oom_fallback.mlir --debug-only='op-validation'
```

[conv2d_l1_oom_fallback.mlir.zip](https://github.com/user-attachments/files/24314026/conv2d_l1_oom_fallback.mlir.zip)

bmalesevicTT:
Linked PR fixed OOM L1 problems for mentioned models, however both still aren't functional with `optimization_level=1` because of other issues:
- `tests/runner/test_models.py::test_all_models_torch[detr/object_detection/pytorch-resnet_50-single_device-inference]` - DRAM OOM being investigated
- `tests/runner/test_models.py::test_all_models_torch[resnet/pytorch-resnet_50_hf_high_res-single_device-inference]`- low pcc, explained in https://github.com/tenstorrent/tt-metal/issues/36487

PR #6569, opened 2026-01-13
[Optimizer] Set conv2d_slice_config to DRAM for L1 OOM fallbacks

bmalesevicTT:
### Ticket
https://github.com/tenstorrent/tt-mlir/issues/6433
https://github.com/tenstorrent/tt-mlir/issues/6537

### Summary
When trying fallbacks for OOM L1, try setting `conv2d_slice_config` to DRAM after trying other conv2d config fields such as `act_h_block_override`. 
This ensures that fallback configurations avoid further L1 pressure and allows compilation to proceed successfully in OOM scenarios observed in the referenced issues.

### Checklist
- [x] Additional repros in cited issues.

### DIFF
--- a/lib/Dialect/TTNN/Transforms/OptimizerPasses/OperationValidationAndFallback.cpp
+++ b/lib/Dialect/TTNN/Transforms/OptimizerPasses/OperationValidationAndFallback.cpp
@@ -802,22 +802,22 @@ bool tryConfigFallbacks(Operation *operation,
   Conv2dConfigAttr originalConfigAttr =
       conv2dAttrs.conv2dConfig.has_value()
           ? conv2dAttrs.conv2dConfig.value()
-          : Conv2dConfigAttr::getDefault(operation->getContext());
+          : Conv2dConfigAttr::get(operation->getContext());
 
-  // Create search space focused on act_block_h_override
-  // Try all multiples of 32 from 1024 down to 32
-  Conv2dConfigSearchSpace searchSpace;
-  searchSpace.actBlockHOverride.push_back(0);
+  // Create search spaces for act_block_h_override
+  // L1Full supports 0; DRAM slices don't
+  // TODO(bmalesevic, #6634): Remove when tt-metal backend supports
+  // act_block_h=0 with DRAM slicing
+  Conv2dConfigSearchSpace l1SearchSpace;
+  l1SearchSpace.actBlockHOverride.push_back(0);
   for (uint32_t val = 1024; val >= 32; val -= 32) {
-    searchSpace.actBlockHOverride.push_back(val);
+    l1SearchSpace.actBlockHOverride.push_back(val);
   }
 
-  TTMLIR_DEBUG(ttmlir::LogComponent::OpValidation,
-               "Trying config fallbacks for operation {} at {} with {} "
-               "act_block_h_override values",
-               operation->getName(), operation->getLoc(),
-               searchSpace.actBlockHOverride.size());
-
+  Conv2dConfigSearchSpace dramSearchSpace;
+  for (uint32_t val = 1024; val >= 32; val -= 32) {
+    dramSearchSpace.actBlockHOverride.push_back(val);
+  }
   // Use Conv2dConfigParams to unset act_block_h_override so the generator
   // will iterate through the search space values
   Conv2dConfigParams configParams(originalConfigAttr);
@@ -844,26 +844,20 @@ bool tryConfigFallbacks(Operation *operation,
                 Conv2dSliceType::DramHeight};
 
             for (Conv2dSliceType sliceType : sliceConfigsToTry) {
-              Conv2dConfigSearchSpace currentSearchSpace = searchSpace;
+              // Use l1SearchSpace for L1Full (includes 0), dramSearchSpace for
+              // DRAM slices (excludes 0)
+              const Conv2dConfigSearchSpace &currentSearchSpace =
+                  (sliceType == Conv2dSliceType::L1Full) ? l1SearchSpace
+                                                         : dramSearchSpace;
 
               auto conv2dSliceConfig = Conv2dSliceConfigAttr::get(
                   convOp->getContext(), sliceType, 0);
               convOp.setConv2dSliceConfigAttr(conv2dSliceConfig);
 
-              // Remove act_block_h_override=0 for DRAM slice configs
-              // TODO(bmalesevic, #6634): Remove when tt-metal backend supports
-              // act_block_h=0
-              if (sliceType != Conv2dSliceType::L1Full) {
-                currentSearchSpace.actBlockHOverride.erase(
-                    std::remove(currentSearchSpace.actBlockHOverride.begin(),
-                                currentSearchSpace.actBlockHOverride.end(), 0),
-                    currentSearchSpace.actBlockHOverride.end());
-              }
-
               Conv2dConfigGenerator configGenerator(
                   &convOp, baseConfig, currentSearchSpace, filterOutFn);
 
-              TTMLIR_DEBUG(ttmlir::LogComponent::OpValidation,
+              TTMLIR_TRACE(ttmlir::LogComponent::OpValidation,
                            "Trying slice config: {} with {} act_block_h values",
                            stringifyEnum(sliceType),
                            currentSearchSpace.actBlockHOverride.size());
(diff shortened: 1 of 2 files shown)


lib/Dialect/TTNN/Transforms/OptimizerPasses/OperationValidationAndFallback.cpp
```diff
@@ -837,6 +837,15 @@ bool tryConfigFallbacks(Operation *operation,
   bool foundConfig =
       llvm::TypeSwitch<Operation *, bool>(operation)
           .Case<ttnn::Conv2dOp, ttnn::ConvTranspose2dOp>([&](auto convOp) {
+            // Set DRAM slice config to avoid L1 memory usage during OOM
+            // fallback. This is set here because tryConfigFallbacks is only
+            // called for OOM errors. Only Conv2dOp supports slice config.
+            if constexpr (std::is_same_v<decltype(convOp), ttnn::Conv2dOp>) {
```
rpavlovicTT: Use llvm constructs for checking op type. Open an issue to implement slice config for `ConvTranspose2d` as there is support in ttnn now.

lib/Dialect/TTNN/Transforms/OptimizerPasses/OperationValidationAndFallback.cpp
rpavlovicTT: This function will exit if there is no conv2d config. This means slice config won't be tried.

lib/Dialect/TTNN/Transforms/OptimizerPasses/OperationValidationAndFallback.cpp
```diff
@@ -837,6 +837,15 @@ bool tryConfigFallbacks(Operation *operation,
...
+              auto conv2dSliceConfig = Conv2dSliceConfigAttr::get(
+                  convOp->getContext(), Conv2dSliceType::DramHeight, 0);
```
rpavlovicTT: Any reason you chose `DramHeight`? 

Comment suggests DRAM WIDTH here

```cpp
struct Op2DSliceConfig {
    // Determines the dimension along which the input & output tensors are sliced.
    // Slices based on [N, H, W, C] shape.
    // Using width slicing is more efficient as it reduces memory usage. This is because the overlap of data between
    // cores is minimized in width slicing, reducing the size of the Halo output. If the Height & Width dimensions are
    // similar, then use Width slicing. Use Height slicing if the Height dimension is significantly larger than the
    // Width dimension.
    enum class SliceType : uint8_t {
        DRAM_HEIGHT,
        DRAM_WIDTH,
        L1_FULL  // This option can be used to force conv2d with a DRAM Input to move it to L1, and output will be in
                 // L1.
    };
    SliceType slice_type = SliceType::DRAM_WIDTH;

    // Number of slices that the output tensor should be divided into.
    uint32_t num_slices = 0;
};

File: lib/Dialect/TTNN/Transforms/OptimizerPasses/OperationValidationAndFallback.cpp [hunk 3]
```diff
@@ -837,6 +837,15 @@ bool tryConfigFallbacks(Operation *operation,
...
+              convOp.setConv2dSliceConfigAttr(conv2dSliceConfig);
```
rpavlovicTT, line 855: We should not set slice config immediately. We should try all 3 options starting from L1 FULL.

ppadjinTT:
:partying_face:

lib/Dialect/TTNN/Transforms/OptimizerPasses/OperationValidationAndFallback.cpp
bmalesevicTT: I tried using `mlir::dyn_cast` inside a combined `.Case<Conv2dOp, ConvTranspose2dOp>`, but this doesn't work cleanly because:

1. TypeSwitch with multiple types in one `.Case` passes the common base type to the lambda. With `auto convOp` it passes a reference, but `mlir::dyn_cast` between reference types causes casting errors in LLVM's type system. 
2. Using `auto *convOp` to force pointer semantics doesn't work either - `Conv2dConfigGenerator` requires the actual typed pointer (`Conv2dOp*` or `ConvTranspose2dOp*`) in its constructor - it stores the op in a `std::variant<Conv2dOp*, ConvTranspose2dOp*>`.
3. Using `dyn_cast` to get the typed pointer inside the lambda still requires if/else branches for all the slice config logic and debug logs.

Separate `.Case` calls seemed like the best solution for now. The duplication is temporary - when ConvTranspose2dOp gains slice config support (#6639), we merge them.

I considered extracting the duplicate code, but it would require 10+ parameters and hurt readability for a temporary situation.

lib/Dialect/TTNN/Transforms/OptimizerPasses/OperationValidationAndFallback.cpp
```diff
@@ -797,11 +797,12 @@ bool tryConfigFallbacks(Operation *operation,
...
-  if (!conv2dAttrs.conv2dConfig.has_value()) {
-    return false;
-  }
 
-  Conv2dConfigAttr originalConfigAttr = conv2dAttrs.conv2dConfig.value();
+  // Use existing config or default if none exists
+  Conv2dConfigAttr originalConfigAttr =
+      conv2dAttrs.conv2dConfig.has_value()
+          ? conv2dAttrs.conv2dConfig.value()
+          : Conv2dConfigAttr::getDefault(operation->getContext());
```
rpavlovicTT: use `get` instead of `getDefault`. `get` returns empty config which you want here. `default` will populate it with some values.

lib/Dialect/TTNN/Transforms/OptimizerPasses/OperationValidationAndFallback.cpp
```diff
@@ -836,10 +837,95 @@ bool tryConfigFallbacks(Operation *operation,
...
+              convOp.setConv2dSliceConfigAttr(conv2dSliceConfig);
+
+              // Remove act_block_h_override=0 for DRAM slice configs
+              // TODO(bmalesevic, #6634): Remove when tt-metal backend supports
+              // act_block_h=0
+              if (sliceType != Conv2dSliceType::L1Full) {
+                currentSearchSpace.actBlockHOverride.erase(
+                    std::remove(currentSearchSpace.actBlockHOverride.begin(),
+                                currentSearchSpace.actBlockHOverride.end(), 0),
+                    currentSearchSpace.actBlockHOverride.end());
```
rpavlovicTT: cleaner is to create new search space without 0

lib/Dialect/TTNN/Transforms/OptimizerPasses/OperationValidationAndFallback.cpp
```diff
@@ -836,10 +837,95 @@ bool tryConfigFallbacks(Operation *operation,
...
+              }
+
+              Conv2dConfigGenerator configGenerator(
+                  &convOp, baseConfig, currentSearchSpace, filterOutFn);
+
+              TTMLIR_DEBUG(ttmlir::LogComponent::OpValidation,
```
rpavlovicTT: TRACE

lib/Dialect/TTNN/Transforms/OptimizerPasses/OperationValidationAndFallback.cpp
```diff
@@ -836,10 +837,95 @@ bool tryConfigFallbacks(Operation *operation,
...
+            return false;
+          })
+          // ConvTranspose2dOp doesn't support slice config attribute yet
+          // TODO(bmalesevic, #6639): Move ConvTranspose2dOp to case above when
+          // slice config attribute is supported
+          .Case<ttnn::ConvTranspose2dOp>([&](ttnn::ConvTranspose2dOp convOp) {
             Conv2dConfigGenerator configGenerator(&convOp, baseConfig,
                                                   searchSpace, filterOutFn);
 
+            TTMLIR_DEBUG(ttmlir::LogComponent::OpValidation,
```
rpavlovicTT: TRACE

lib/Dialect/TTNN/Transforms/OptimizerPasses/OperationValidationAndFallback.cpp
bmalesevicTT: Assumed it prefers those values since its after optimizer, but makes sense to align with having no set configuration - now changed to use get().

Issue #6388, opened 2025-12-18
[Optimizer] Return type row-major integer, expected tiled


bmalesevicTT:
Analysis of running nightly ~500 models shows 4 of them (YOLOv6 variants) fail only when optimizer is enabled because final result of the model is expected to be tiled, but the Optimizer pass keeps integers in row-major format. 

Proposition for solution: Keep/insert a 'toLayout' op at the very end of the graph to tilize the return tensor, so that the signature of the function matches it.

Reproducible on branch `bmalesevic/temp-xla-opt-lvl-1`, or simply turn optimizer on in tt-xla:
- TT-XLA nightly tests: `pytest tests/jax/single_chip/test_data_types.py::test_dtypes`, specifically for integer types
- TT-XLA nightly models:
1. YOLOv6n (Nano)
pytest tests/runner/test_models.py -k "pytorch-yolov6n-single_device-full-inference" -svv

2. YOLOv6s (Small)
pytest tests/runner/test_models.py -k "pytorch-yolov6s-single_device-full-inference" -svv

3. YOLOv6m (Medium)
pytest tests/runner/test_models.py -k "pytorch-yolov6m-single_device-full-inference" -svv

4. YOLOv6l (Large)
pytest tests/runner/test_models.py -k "pytorch-yolov6l-single_device-full-inference" -svv

umalesTT:
Yeah, I encountered this one too; running optimizer with optimization level set to 1 or 2 on Llama 1B fine-tuning results in this error, failing on ttnn.arange op
fyi @vladimirjovanovicTT

bmalesevicTT:
Communicated with @umalesTT, his example is a separate problem.
Mentioned YOLOv6 variants and unit data type tests work with `optimization_level=1` after merging the fix.

PR #6570, opened 2026-01-13
[Optimizer] Fix function return layout mismatch

bmalesevicTT:
### Ticket
https://github.com/tenstorrent/tt-mlir/issues/6388

### Summary
This PR fixes a function return layout mismatch by inserting toLayout operations for return operands to match the expected Tile layout.

The fix must be applied in the same pass where the breaking change is
introduced. Graph validation fails immediately for this class of errors,
preventing execution from reaching the OpValidation pass where layout
corrections would normally occur. To address this, `toLayout` ops are
added at the end of the row-major propagation pass if needed, to ensure return operand layouts are valid
before graph validation.

### Checklist
- [x] Validated with cited issue repro.

### DIFF
--- a/lib/Dialect/TTNN/Transforms/OptimizerPasses/RowMajorLayoutPropagation.cpp
+++ b/lib/Dialect/TTNN/Transforms/OptimizerPasses/RowMajorLayoutPropagation.cpp
@@ -10,6 +10,7 @@
 #include "ttmlir/Dialect/TTNN/IR/TTNNOpsAttrs.h"
 #include "ttmlir/Dialect/TTNN/IR/TTNNOpsTypes.h"
 #include "ttmlir/Dialect/TTNN/Transforms/Passes.h"
+#include "ttmlir/Dialect/TTNN/Utils/TransformUtils.h"
 #include "ttmlir/Dialect/TTNN/Validation/OpConstraintValidation.h"
 #include "ttmlir/FunctionTypes.h"
 #include "ttmlir/Support/Logger.h"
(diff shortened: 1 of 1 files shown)


lib/Dialect/TTNN/Transforms/OptimizerPasses/RowMajorLayoutPropagation.cpp
```diff
@@ -79,6 +79,9 @@ class TTNNRowMajorLayoutPropagation
 
       llvm::DenseMap<Operation *, Layout> opLayoutConstraints;
       propagateRowMajorLayout(func, rowMajorArgs, opLayoutConstraints);
+
+      // After propagation, insert ToLayoutOps before returns where needed
+      insertToLayoutOpsBeforeReturns(func, rewriter);
```
rpavlovicTT: Can we instead stop propagation if user is `ReturnOp` and its layout is different wrt. `isTiled()` property?

bmalesevicTT: Consolidated return handling into the main propagation loop for efficiency. 
Stopping at ReturnOp without inserting ToLayoutOp would mismatch types when ops must output row-major but the signature expects tiled. I considered checking to stop propagation earlier to avoid this, but that would add complexity. Current approach inserts ToLayoutOp only when needed.
Let me know if you had a different approach in mind.

lib/Dialect/TTNN/Transforms/OptimizerPasses/RowMajorLayoutPropagation.cpp
```diff
@@ -307,6 +317,62 @@ class TTNNRowMajorLayoutPropagation
...
+        ttnn::LayoutAttr::get(rewriter.getContext(),
+                              expectedLayout.getLayout()),
+        ttcore::DataTypeAttr::get(rewriter.getContext(),
+                                  expectedLayout.getDataType()),
+        ttnn::MemoryConfigAttr::get(
+            rewriter.getContext(), expectedLayout.getMemLayout(),
+            ttnn::BufferTypeAttr::get(rewriter.getContext(),
+                                      expectedLayout.getBufferType()),
+            /*shardSpec=*/std::nullopt));
+
```
rpavlovicTT: Can we use 

https://github.com/tenstorrent/tt-mlir/blob/efdde5a128c8a63b7df38ab12c06d998fe28dce9/include/ttmlir/Dialect/TTNN/Utils/TransformUtils.h#L21-L29

lib/Dialect/TTNN/Transforms/OptimizerPasses/RowMajorLayoutPropagation.cpp
rpavlovicTT: That's fine.

PR #6571, opened 2026-01-13
Filter unique ops for both discover + parted scripts

nsmithtt:


tools/scripts/parted.py
```diff
@@ -10,28 +10,27 @@
...
+        regions = entry.regions
+        if len(regions) > 0:
+            assert len(regions) == 1
+            blocks = regions[0].blocks
+            assert len(blocks) == 1
+            yield from filter_dialect_ops(
+                blocks[0], dialects=dialects, op_filter=op_filter
+            )
+        else:
+            for op in block.operations:
```
ddilbazTT: Isn't this looping over the same for loop condition? Just wanted to check if that's what you wanted.

nsmithtt: Oh good catch, thank you!

PR #6575, opened 2026-01-14
Fix Documentation for EmbeddingBackwardOp

mmilosevicTT:
### Ticket
NA

### Problem description
In the documentation we do not use proper phrasing for input gradients. They do not come from forward pass, but from previous backpropagation steps.

### What's changed
Change in documentation to use proper phrasing for the origin of input gradients.


sdjordjevicTT (approved):
Great! Thanks!

Issue #6585, opened 2026-01-14
[codegen] tt-alchemist errors out when generating Python from TTNN


svuckovicTT:
Recent PR https://github.com/tenstorrent/tt-mlir/pull/6361 reworked pipelines, which left TTNN->EmitPy pipeline in an invalid state for tt-alchemist.

PR #6589, opened 2026-01-14
Fix tt-alchemist's generate-python API for TTNN IR input

svuckovicTT:
### Ticket
#6585 

### Problem description
```
tt-alchemist generate-python [TTNN IR file] -o some_output_dir
```
results in an error - `DeviceModuleOp` can't be converted by EmitPy->Python translation.

Pipelines have recently been reworked and there are additional passes that need to run in tt-alchemist so that it can work with TTNN IR as an input.

### What's changed
Added a temporary workaround pipeline to run the additional pipelines/passes.
Added a CI test for this specific case.

### Checklist

### DIFF
--- a/include/ttmlir/Dialect/TTNN/Pipelines/TTNNPipelines.h
+++ b/include/ttmlir/Dialect/TTNN/Pipelines/TTNNPipelines.h
@@ -533,7 +533,7 @@ void createTTIRToEmitCPipeline(OpPassManager &pm,
 void createTTIRToEmitPyPipeline(OpPassManager &pm,
                                 const TTIRToEmitPyPipelineOptions &options);
 
-void createWorkaroundMixedTTIRTTNNToEmitPyPipeline(
+void createTTNNToEmitPyPipeline(
     OpPassManager &pm, const TTNNToEmitPyDevicePipelineOptions &options);
 
 void registerTTNNPipelines();
--- a/lib/Dialect/TTNN/Pipelines/TTNNPipelines.cpp
+++ b/lib/Dialect/TTNN/Pipelines/TTNNPipelines.cpp
@@ -500,17 +500,15 @@ void createTTIRToEmitPyPipeline(OpPassManager &pm,
   pm.addPass(createEmitPyLinkModulesPass());
 }
 
-// Workaround pipeline for lowering mixed TTIR/TTNN to EmitPy.
+// Complete pipeline for lowering TTNN to EmitPy.
 //
-// This is a temporary pipeline used when the input is already in TTNN dialect
-// (e.g., from tt-alchemist with TTNN input). It handles the case where the
-// device module contains TTNN ops but the CPU module may still need TTIR
-// processing.
+// This pipeline is used when the input is already in TTNN dialect, and assumes
+// the CPU module is still in TTIR.
 //
 // Device module: TTNN -> EmitPy.
 // CPU module: TTIR -> TTNN -> EmitPy (with golden functions).
 //
-void createWorkaroundMixedTTIRTTNNToEmitPyPipeline(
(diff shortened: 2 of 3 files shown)


include/ttmlir/Dialect/TTNN/Pipelines/TTNNPipelines.h
```diff
@@ -533,6 +533,9 @@ void createTTIRToEmitCPipeline(OpPassManager &pm,
 void createTTIRToEmitPyPipeline(OpPassManager &pm,
                                 const TTIRToEmitPyPipelineOptions &options);
 
+void createWorkaroundMixedTTIRTTNNToEmitPyPipeline(
```
sdjordjevicTT: I see this as a valid pipeline, not needing to include a workaround in its name?
Let's name it:
`void createTTNNToEmitPyPipeline()`

sdjordjevicTT (approved):

dmilinkovicTT (approved):
LGTM for unblocking the alchemist failures.

But since this has proven that the previously merged pipeline refactor isn't the solution that covers all use-cases, I assume we should attempt to simplify the pipelines.

One idea:
1. Each E2E pass pipeline starts with a pass which would categorize all ops by their dialect into submodules:
```mlir
module {
   device_module {
       ttir_ops_module {
         // ...ops
       }
       // ... or
       ttnn_ops_module {
         // ...ops
       }
   }
   cpu_module {
      llvm_ops_module {
         // ...ops
      }
   }
}
```

(of course, we should assert that only one such module exists inside Device/CPU module)

2. Each E2E pipeline can now **statically**, using nested pass-managers, populate passes which should lower the corresponding modules - for example, `EmitPy` pipeline could:
  - Lower `device_module.ttir_ops_module` to TTNN, and then to EmitPy
  - Lower `device_module.ttnn_ops_module` to EmitPy
  - Lower `cpu_module.ttir_ops_module` to TTNN, and then to EmitPy
  - Fail if `cpu_module.llvm_ops_module` exist
  - ...

This could, at least in theory, simplify the pipelines by enabling **one E2E pipeline per target** rule, and such E2E pipelines could accept inputs from multiple stages of compilation.

3. At the end of E2E pipelines, previously added temporary modules get unwrapped.

@svuckovicTT wdyt?

EDIT: ah, I forgot about the requirement for compiling TTIR to TTNN only once for both Flatbuffer and EmitPy - this doesn't seem to fit well with the proposed idea...

include/ttmlir/Dialect/TTNN/Pipelines/TTNNPipelines.h
svuckovicTT: Sure, just note that there's also TTIR on the CPU side, so it's not strictly TTNN->EmitPy.

sdjordjevicTT: Yeah, yeah, fine with that. If we want more specific, maybe something like this:
`void createTTNNAndCPUtoEmitPyPipeline()`

But I would rather keep it as a simpler `void createTTNNToEmitPyPipeline()`

svuckovicTT: I've already renamed it back to `createTTNNToEmitPyPipeline`, will merge as-is. We are already thinking on how to rework pipelines a bit so that it makes more sense for the whole flow. The way it is today is not terribly confusing so we can live with it.

@dmilinkovicTT and I are on top of it, will share ideas once we work the details out :)

sdjordjevicTT: Great! Thanks!

Issue #6591, opened 2026-01-14
TTNN Layout Pass Fails to Handle mesh_shard in Const Eval Case


hkwonTT:
## Problem

The TTNN layout pass fails when calling const_eval functions containing `mesh_shard` via `LoadCachedOp`, causing type mismatches.

### Symptoms

Compilation error:
```
error: 'ttcore.load_cached' op Operand type mismatch at index 0
%1 = ttcore.load_cached(@main_const_eval_1, [%arg155]) : (tensor<4096x4096xbf16>) -> tensor<4096x512xbf16>
```

The error shows:
- Operand type: `tensor<4096x4096xbf16, #ttnn.buffer_type<dram>>`
- Expected type: `tensor<4096x4096xbf16, #ttnn.buffer_type<system_memory>>`

### Root Cause

1. The function signature of `@main_const_eval_1` (containing `mesh_shard`) is updated to `SystemMemory` by the `TTNNLayoutFuncInputOutputTypeRewriter` pattern.
2. However, the `LoadCachedOp` operand (`%arg155`) still retains the `DRAM` layout
3. This causes a type mismatch:
   - Function signature expects: `SystemMemory`
   - Actual operand has: `DRAM` layout

### Steps to reproduce the issue

Compile below MLIR with this command:
`ttmlir-opt --ttnn-layout example.mlir`

example.mlir:
```mlir
func.func private @main_const_eval_1(%arg0: tensor<4096x4096xbf16>) -> tensor<4096x512xbf16> {
  %0 = "ttir.mesh_shard"(%arg0) : (tensor<4096x4096xbf16>) -> tensor<4096x512xbf16>
  return %0 : tensor<4096x512xbf16>
}

func.func @main(%arg155: tensor<4096x4096xbf16> {ttnn.layout = <DRAM>}) {
  // Error: %arg155 has DRAM layout, but @main_const_eval_1 expects SystemMemory
  %1 = ttcore.load_cached(@main_const_eval_1, [%arg155]) : (tensor<4096x4096xbf16>) -> tensor<4096x512xbf16>
}
```

Result:
```
error: 'ttcore.load_cached' op Operand type mismatch at index 0
    %1 = ttcore.load_cached(@main_const_eval_1, [%arg155]) : (tensor<4096x4096xbf16>) -> tensor<4096x512xbf16>
```

hkwonTT:
#6631 may fix this issue.

PR #6592, opened 2026-01-14
Fix TTNN layout pass to handle mesh_shard in const_eval case

hkwonTT:
### Ticket
#6591 

### Problem description
The TTNN layout pass fails when calling const_eval functions containing `mesh_shard` via `LoadCachedOp`, causing type mismatches. When the function signature is updated to `SystemMemory` for `mesh_shard` operations, the `LoadCachedOp` operands still retain their original layout (e.g., `DRAM`), resulting in compilation errors.

### What's changed
Added operand type conversion in `LoadCachedOp` rewriting to match function signatures by inserting `ToLayoutOp` when needed. Enhanced `shouldForceInputSystemMemory` to detect when `LoadCachedOp` operands should be forced to `SystemMemory` based on callee function expectations.

### Checklist

### DIFF
--- a/tools/scripts/d2m-discover.py
+++ b/tools/scripts/d2m-discover.py
@@ -22,6 +22,24 @@ from ttmlir.dialects import func, ttir
 # Operations matching these names will be excluded from analysis.
 OP_IGNORELIST = [
     "func.return",
+    # ttcore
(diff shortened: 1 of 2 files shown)


hkwonTT:
Duplicated with #6631

PR #6628, opened 2026-01-14
hidden visibility to prevent collision when multiple python packages …

anuragsingh-tt:
…are loaded simultaneously


CMakeLists.txt
```diff
@@ -76,6 +76,9 @@ endif()
 
 set(CMAKE_BUILD_WITH_INSTALL_NAME_DIR ON)
 
+set(CMAKE_C_VISIBILITY_PRESET hidden)
+set(CMAKE_CXX_VISIBILITY_PRESET hidden)
```
nsmithtt, line 80: I don't think we can do this without explicitly exporting our APIs, I'd expect this to break front-end builds because all public APIs are no longer visible.

anuragsingh-tt: I can move this to the Python `CMakeLists.txt` but we already use the `ttmlir` namespace so this actually shouldn't be an issue. Deleting this PR.

Issue #6009, opened 2025-11-24
☂️ Execution of const-eval subgraphs on the host CPU


dmilinkovicTT:
Currently, TT-MLIR supports extracting const-eval-able operation subgraphs into functions whose results are being cached and reused across forward invocations.

Additionally, TT-MLIR supports hoisting execution of certain operations to the CPU on the host machine, which is currently being used to execute some non-supported SHLO ops through SHLO -> linalg -> ... -> LLVM path.

This issue encapsulates the efforts required for enabling CPU-hoisting of the const-eval subgraphs. The main motivation comes from the expectation that PCC should be improved due to the fact that we can use higher-precision data types on the CPU.

Issue #2145, opened 2025-02-07
Enable canonicalizer on linalg without removing non-returned values


vwellsTT:
In the linalg to LLVM pipeline introduced by CPU hoisting, I was forced to disable the `createCanonicalizerPass()` pass because it removes linalg ops which aren't returned, and I want to avoid returning values to prevent messy RVOs.  However, once we start hoisting multiple ops into the same func etc, canonicalization might be useful, and we might need to find a workaround so we can re-enable it without optimizing away values we care about.
(For example, we could perhaps define a `dummy` func with variadic arguments, and pass all the values we wish the call to yield to a fake call to `dummy` at the end of the function.  Then run canonicalizer, then a pass to remove `dummy` call, then rest of the pipeline).

PR #6633, opened 2026-01-15
Enable const-eval CPU-hoisting by default

dmilinkovicTT:
### Ticket
#6009 #2145 

### Problem description
As TTIR -> Linalg conversion coverage is becoming sufficient, we should start moving towards CPU-hoisted const-eval enabled by default.

This PR includes various fixes required for enabling the feature.

### What's changed
- Extend CPU-hoisting pass implementation so that in includes `DenseElementsAttr` conversion to CPU-supported types
- Disabled CPU-hoisting const-eval subgraphs if CCLs or non-identity `mesh_shard` ops are present in the subgraph
  - Will be explored further as part of #6709 
- Disabled CPU-hoisting const-eval subgraphs if only creation ops are present
- Disabled CPU-hoisting const-eval subgraphs for EmitC target
   - Will be re-enabled after #6100 lands
- Updated optimizer-related passes so that they apply on forward-device functions 
- Re-enabled canonicalization pass on linalg ops, since RVOs aren't an issue anymore (#6373)
- Introduced new approach for integer signedness in CPU-hoisted functions
  - Should be reconsidered on a broader level as part of #6797 
- Disabled multithreaded compilation because of mutex ownership issues in `MetalContext`
  - Should be re-enabled after [tt-metal issue](https://github.com/tenstorrent/tt-metal/issues/31041) is resolved

### Checklist

### DIFF
--- a/test/ttmlir/Dialect/TTNN/optimizer/conv2d_l1_oom_fallback.mlir
+++ b/test/ttmlir/Dialect/TTNN/optimizer/conv2d_l1_oom_fallback.mlir
@@ -1,5 +1,5 @@
 // REQUIRES: opmodel
-// RUN: ttmlir-opt --ttcore-register-device --ttnn-operation-validation-and-fallback -o %t %s
+// RUN: ttmlir-opt --ttcore-register-device --ttcore-mark-functions-as-forward --ttnn-operation-validation-and-fallback -o %t %s
 // RUN: FileCheck %s --input-file=%t
 //
 // This test verifies that when a Conv2d operation with conv2d_slice_config=<l1_full>


AleksKnezevic:
This is awesome @dmilinkovicTT! Before merging you should run a tt-xla nightly pipeline as I suspect this will have an effect on overall model accuracy (I assume mostly positive).

dmilinkovicTT:
@AleksKnezevic thanks!

I've been running `basic-tests-nightly` and `model-tests-passing` during the last few weeks against this branch, and no PCC degradations were observed.

However, I won't be merging this PR until [this one](https://github.com/tenstorrent/tt-xla/pull/3032) is merged (where the flag is disabled by default, for now).

When this PR gets uplifted into TT-XLA, I'll create a branch for enabling the flag in TT-XLA. This way, the pipelines runs won't be polluted with the failures unrelated to the CPU-hoisted const-eval.

Wdyt?

AleksKnezevic:
Works for me as long as we don't keep the defaults different (between tt-xla and tt-mlir) for too long.

PR #6638, opened 2026-01-15
Support presharded input and output execution in ttrt

tapspatel:
This PR support execution of mlir graphs that have presharded inputs/outputs via ttrt.

In introduces a RuntimeTensorShardingAttr for tensor types, which encode the shard status of that tensor and it's local shape. Shard status can be one of presharded/unsharded and local shape is the shape of the tensor on device. This information gets captured during lowering in stablehlo-pipeline during `createApplyArgumentShardStatusPass`. Then, when translating to flatbuffer executable, each input/output tensor will get appended with this information, so runtime knows how to generate these inputs and collect the outputs.

### DIFF
--- a/include/ttmlir/Dialect/TTCore/IR/TTCoreOpsTypes.td
+++ b/include/ttmlir/Dialect/TTCore/IR/TTCoreOpsTypes.td
@@ -688,6 +688,24 @@ def TTCore_MeshesAttr : TTCore_Attr<"Meshes", "meshes"> {
   }];
 }
 
+def TTCore_RuntimeTensorShardingAttr : TTCore_Attr<"RuntimeTensorSharding", "runtime_tensor_sharding"> {
+  let summary = "Runtime tensor sharding information";
+  let description = [{
+    Holds the runtime sharding information for this tensor.
+    This includes whether the tensor is presharded/unsharded and its per device shape.
+    This attribute is required for graphs coming from torch-xla/jax where the inputs/outputs to the graph can be presharded.
+    Any tool built ontop of mlir runtime needs to know how to generate the inputs (unsharded or presharded) and this attribute encapsulate that information.
+  }];
+  let parameters = (ins
+    "ShardStatusAttr":$shardStatus,
+    "RankedTensorType":$localShape);
+
+  let assemblyFormat = [{
+        `shard_status` `=` $shardStatus `,`
+        `local_shape` `=` $localShape
+  }];
+}
+
 //===----------------------------------------------------------------------===//
 // TT type definitions
 //===----------------------------------------------------------------------===//
--- a/include/ttmlir/Target/Utils/FuncOpToProgram.h
+++ b/include/ttmlir/Target/Utils/FuncOpToProgram.h
@@ -73,15 +73,54 @@ Program<OpT> funcOpToProgram(
   program.name = entry.getSymName().data();
 
   for (auto &input : entry.getBody().getArguments()) {
-    program.inputs.push_back(cache.getOrCreate(input, tensorValueToFlatbuffer));
+    // Get argument encoding to determine sharding status.
+    mlir::DictionaryAttr argAttrDict =
+        entry.getArgAttrDict(input.getArgNumber());
+    ttcore::ShardStatus shardStatus = ttcore::ShardStatus::Unsharded;
+    mlir::RankedTensorType localShape =
+        mlir::cast<mlir::RankedTensorType>(input.getType());
+
+    if (argAttrDict) {
+      auto runtimeTensorShardingAttr =
+          argAttrDict.get(mlir::tt::ttcore::RuntimeTensorShardingAttr::name);
+      if (runtimeTensorShardingAttr) {
+        auto rtsAttr = mlir::cast<mlir::tt::ttcore::RuntimeTensorShardingAttr>(
+            runtimeTensorShardingAttr);
+        shardStatus = rtsAttr.getShardStatus().getValue();
+        localShape = rtsAttr.getLocalShape();
+      }
+    }
+
+    program.inputs.push_back(cache.getOrCreate(input, tensorValueToFlatbuffer,
+                                               shardStatus, localShape));
   }
 
   mlir::AsmState printState(entry, printFlags);
   entry.getBody().walk([&](mlir::Operation *op) {
     if (auto returnOp = dyn_cast_if_present<func::ReturnOp>(op); returnOp) {
-      for (auto output : returnOp.getOperands()) {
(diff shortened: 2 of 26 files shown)


hshahTT:
Thanks @tapspatel ! Could you also revert the changes to the following files made in https://github.com/tenstorrent/tt-mlir/pull/6504 ? This will also help test your changes in this PR.

- [test_shardy_ops_n300.py](https://github.com/tenstorrent/tt-mlir/pull/6504/files#diff-f25baf69164d015fcbfc245e3a9d4a77a4264063047d777bf0f6d1d3f5572992)
- [llmbox/sdy_all_slice.mlir](https://github.com/tenstorrent/tt-mlir/pull/6504/files#diff-51a3b5942f0126970a40d9ddc5b755df450dabb7c96403cff719f6e67be85884)
- [n300/sdy_all_slice.mlir](https://github.com/tenstorrent/tt-mlir/pull/6504/files#diff-4b0900898a28aab9e1721a23f6eeb0af1b8cf54ff4489af7d5c5944d2069e60b)
- [n300/sdy_fill_cache1x2.mlir](https://github.com/tenstorrent/tt-mlir/pull/6504/files#diff-f470388852aff866d045176ad59dc8c8d636f403315943737615a5424382e7b0)

jameszianxuTT:
Could you elaborate a bit more in the PR description about the approach taken here? A bit more context would be helpful when reviewing :)

tools/ttrt/common/util.py
```diff
@@ -1005,23 +1018,67 @@ def populate_inputs(self, init_fn, golden_inputs=[]):
...
+
+                    torch_tensor = init_fn(
+                        local_shape,
+                        dtype=Binary.Program.from_data_type(
+                            i["desc"]["layout"]["memory_desc"]["data_type"]
+                        ),
+                    )
+                    for shard_index in range(num_devices):
+                        tensor_shard = torch_tensor.clone()
+                        tensor_shards.append(tensor_shard)
```
jameszianxuTT, line 1094: Does this logic for "reassembling" a sharded tensor for input/output work for partially sharded presharded tensors? 

Eg. on a 4x2 mesh, if the tensor is replicated over the size-4 dim but sharded over the size-2 dim.

From my reading, the runtime tensor sharding attr only captures the local shape and whether the tensor is presharded or unsharded, but it seems like we'd need more information about the exact sharding spec to properly gather the results.

tapspatel: it will work for partially sharded tensors. All we need is the per device shape, so it doesn't matter how the tensor is being replicated/sharded and on what axis.

tapspatel: I purposely left out gathering the results out of ttrt for the scope of this PR. You can add in sharding information to this runtime config to do that, but this only focuses on ensuring ttrt can support these types of graphs. For gathering the resultant tensor into your specified sharding config, we should look at how tt-xla does it and mirror something common in ttmlir.

jameszianxuTT: That makes sense, understood.

lib/Dialect/StableHLO/Transforms/ApplyArgumentShardStatus.cpp
```diff
@@ -13,15 +13,10 @@ namespace mlir::tt::stablehlo {
...
-                                             bool gspmdAnnotationsExist,
-                                             bool shardyAnnotationsExist) {
-  llvm::StringRef annotationsKeyWord =
-      shardyAnnotationsExist ? mlir::sdy::TensorShardingAttr::name
-                             : mlir::tt::gspmd_utils::kXlaShardingAttr;
-
+static mlir::LogicalResult
+updateShardStatus(MLIRContext *context, mlir::ModuleOp &module,
+                  mlir::OpBuilder &builder, func::FuncOp &funcOp,
+                  bool gspmdAnnotationsExist, bool shardyAnnotationsExist) {
```
hshahTT, line 19: Nit: We don't use `gspmdAnnotationsExist` in this function anymore, so this can be removed from the argument list.

include/ttmlir/Dialect/TTCore/IR/TTCoreOpsTypes.td
```diff
@@ -688,6 +688,24 @@ def TTCore_MeshesAttr : TTCore_Attr<"Meshes", "meshes"> {
   }];
 }
 
+def TTCore_RuntimeTensorShardingAttr : TTCore_Attr<"RuntimeTensorSharding", "runtime_tensor_sharding"> {
+  let summary = "Runtime tensor sharding information";
+  let description = [{
+    Holds the runtime sharding information for this tensor.
+    This includes whether the tensor is presharded/unsharded and it's per device shape.
```
vwellsTT: grammar nit: its not it's

include/ttmlir/Target/Utils/FuncOpToProgram.h
```diff
@@ -73,15 +73,54 @@ Program<OpT> funcOpToProgram(
   program.name = entry.getSymName().data();
 
   for (auto &input : entry.getBody().getArguments()) {
-    program.inputs.push_back(cache.getOrCreate(input, tensorValueToFlatbuffer));
+    // Get argument encoding to determine sharding status
```
vwellsTT: nit: .

lib/Target/TTNN/TTNNToFlatbuffer.cpp
```diff
@@ -254,7 +282,9 @@ createOp(FlatbufferObjectCache &cache, ToMemoryConfigOp op) {
   // needed.
   auto memoryConfig = toFlatbuffer(cache, op.getMemoryConfig());
 
-  auto output = cache.getOrCreate(op.getResult(), tensorValueToFlatbuffer);
+  auto output = cache.getOrCreate(op.getResult(), tensorValueToFlatbuffer,
```
vwellsTT: I would suggest overloading `getOrCreate` (or creating a new func like `getOrCreateNoSharding` or something) to have a version with no ShardStatus which defaults to Unsharded.  This will make the delta to this file smaller, and generally avoid the need for passing ShardStatus everywhere, which imo is desirable since the actual use case seems narrow.

tapspatel: yeah good point

tapspatel:
@svuckovicTT you can support that use case with the base functionality in this pr

pretty much there is a runtime tensor config which can be attached to any tensor, input/output or intermediate. This PR only supports inputs/outputs but we would need to add a pass which can analyze the entire graph and add in this attribute for each intermediate tensor (depending on the op like ttir mesh shard etc). Then change the ttnntoflatbuffer to query this information from the tensor attribute. That will then get stored in the flatbuffer

PR #6642, opened 2026-01-15
Restructured mlir python bindings to fit CAPI pattern

tapspatel:
This PR restructures MLIR bindings to better fit CAPI pattern. This also supports working with native python types, instead of having to downcast in order to access object members and attributes, which is not pythonic. 

It introduces the following structure:

Attributes:
- all attributes are to inherit from `mlir::python::nanobind_adaptors::mlir_attribute_subclass`
- associated `isSomeAttr` must be implemented 
- associated `.def_classmethod("get",` must be implemented
- associated `.def_classmethod("from_attribute",` must be implemented

See `GridAttr` for an example.

Types:
- all types are to inherit from `mlir::python::nanobind_adaptors::mlir_type_subclass`
- associated `isSomeType` must be implemented 
- associated `.def_classmethod("get",` must be implemented
- associated `.def_classmethod("from_attribute",` must be implemented

See `TileType` for an example.

Enums:
- enums are to be binded as a separate class in their respective source file

See `MlirDataTypeEnum` for an example

Other:
- CAPI headers must be defined in respective header file eg. `include/ttmlir-c/TTAttrs.h`
- Function implementations must be defined in respective source file eg. `lib/CAPI/TTCoreAttrs.cpp`
- unwrap/wrap logic should only happen in source file, NOT python module file

### DIFF
--- a/include/ttmlir-c/TTAttrs.h
+++ b/include/ttmlir-c/TTAttrs.h
@@ -12,18 +12,117 @@
 extern "C" {
 #endif
 
-MLIR_CAPI_EXPORTED MlirAttribute ttmlirTTGridAttrGet(MlirContext ctx,
-                                                     int64_t *shape,
-                                                     size_t shapeSize);
+//===----------------------------------------------------------------------===//
+// TileType
+//===----------------------------------------------------------------------===//
+
+MLIR_CAPI_EXPORTED bool ttmlirIsTileType(MlirType type);
+
+MLIR_CAPI_EXPORTED MlirType ttmlirTileTypeGet(MlirContext ctx, int64_t height,
+                                              int64_t width,
+                                              MlirAttribute dataType);
+
+MLIR_CAPI_EXPORTED int64_t ttmlirTileTypeGetHeight(MlirType type);
+
+MLIR_CAPI_EXPORTED int64_t ttmlirTileTypeGetWidth(MlirType type);
+
+MLIR_CAPI_EXPORTED MlirAttribute ttmlirTileTypeGetDataType(MlirType type);
+
+//===----------------------------------------------------------------------===//
+// GridAttr
+//===----------------------------------------------------------------------===//
+
+MLIR_CAPI_EXPORTED bool ttmlirIsGridAttr(MlirAttribute attr);
+
+MLIR_CAPI_EXPORTED MlirAttribute ttmlirGridAttrGet(MlirContext ctx,
+                                                   intptr_t nShape,
+                                                   const int64_t *shape);
+
+MLIR_CAPI_EXPORTED intptr_t ttmlirGridAttrGetShapeSize(MlirAttribute attr);
+
+MLIR_CAPI_EXPORTED int64_t ttmlirGridAttrGetShapeElem(MlirAttribute attr,
+                                                      intptr_t pos);
+
+//===----------------------------------------------------------------------===//
+// ReduceTypeAttr
+//===----------------------------------------------------------------------===//
+
+enum MlirReduceTypeEnum {
+  MlirReduceTypeSum,
+  MlirReduceTypeMean,
+  MlirReduceTypeMax,
+  MlirReduceTypeMin,
+  MlirReduceTypeStd,
+  MlirReduceTypeVar,
+  MlirReduceTypeProd,
+  MlirReduceTypeInvalid,
+};
+typedef enum MlirReduceTypeEnum MlirReduceTypeEnum;
+
+MLIR_CAPI_EXPORTED bool ttmlirIsReduceTypeAttr(MlirAttribute attr);
+
+MLIR_CAPI_EXPORTED MlirAttribute
+ttmlirReduceTypeAttrGet(MlirContext ctx, MlirReduceTypeEnum reduceType);
+
+MLIR_CAPI_EXPORTED MlirAttribute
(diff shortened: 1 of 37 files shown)


include/ttmlir-c/TTAttrs.h
```diff
@@ -12,19 +12,46 @@
...
 #endif
 
-MLIR_CAPI_EXPORTED MlirAttribute ttmlirTTGridAttrGet(MlirContext ctx,
-                                                     int64_t *shape,
-                                                     size_t shapeSize);
+//===----------------------------------------------------------------------===//
+// GridAttr
+//===----------------------------------------------------------------------===//
+
+MLIR_CAPI_EXPORTED bool ttcoreIsGridAttr(MlirAttribute attr);
```
nsmithtt: I think it might be better to stick with the `ttmlir` prefix, unless we update all API's.

tapspatel: we'll have to update all these binded APIs, its much more easy to maintain this structure and many stakeholders (ttlang, builder, ttnn-jit) - as a matter of fact, we are not even using these CAPIs today :P our entire implementation is in the python module files. 

but I'll keep the prefix ttmlir

nsmithtt: Sounds good, ttmlir isn't ideal I agree, but let's keep it consistent and flip the switch all at once if we want to choose a new prefix

include/ttmlir/Bindings/Python/TTMLIRModule.h
```diff
@@ -5,6 +5,8 @@
 #ifndef TTMLIR_BINDINGS_PYTHON_TTMLIRMODULE_H
 #define TTMLIR_BINDINGS_PYTHON_TTMLIRMODULE_H
 
+#include "mlir-c/BuiltinAttributes.h"
```
vmilosevic: :warning: **llvm\-include\-order** :warning:
```suggestion
#include "mlir-c/Bindings/Python/Interop.h"
#include "mlir-c/BuiltinAttributes.h"
```

include/ttmlir/Bindings/Python/TTMLIRModule.h
```diff
@@ -5,6 +5,8 @@
...
+#include "mlir-c/IR.h"
 #include "mlir-c/Bindings/Python/Interop.h"
```
vmilosevic, line 8: :warning: **llvm\-include\-order** :warning:
```suggestion

```

lib/CAPI/TTCoreAttrs.cpp
```diff
@@ -9,28 +9,65 @@
 
 using namespace mlir::tt::ttcore;
 
-MlirAttribute ttmlirTTGridAttrGet(MlirContext ctx, int64_t *shape,
-                                  int shapeSize) {
-  return wrap(GridAttr::get(unwrap(ctx), {shape, shape + shapeSize},
-                            mlir::AffineMap::get(unwrap(ctx))));
+template <typename AttrTy>
+AttrTy unwrapAttr(MlirAttribute attr) {
+  return mlir::cast<AttrTy>(unwrap(attr));
```
brnorris03: I don't know if this helper is worth having, mainly adds another stack frame if anything fails making it harder to go to the actual problem. But given how simple most of the callers are, this is very much just a minor nit that can be disregarded.

tapspatel:
making draft due to some additional changes I am adding

include/ttmlir/CAPI/TTCoreEnums.h
```diff
@@ -0,0 +1,24 @@
+// SPDX-FileCopyrightText: (c) 2026 Tenstorrent AI ULC
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#ifndef MLIR_CAPI_TTCOREENUMS_H
+#define MLIR_CAPI_TTCOREENUMS_H
```
vmilosevic: :warning: **llvm\-header\-guard** :warning:
```suggestion
#ifndef TTMLIR_CAPI_TTCOREENUMS_H
#define TTMLIR_CAPI_TTCOREENUMS_H
```

vmilosevic (commented):

include/ttmlir/CAPI/TTCoreEnums.h
```diff
@@ -0,0 +1,24 @@
...
+inline mlir::tt::ttcore::DataType &unwrap(MlirDataType dataType) {
+  assert(dataType.ptr && "unexpected null dataType");
+  return *(static_cast<mlir::tt::ttcore::DataType *>(dataType.ptr));
+}
+
+inline MlirDataType wrap(mlir::tt::ttcore::DataType &dataType) {
+  return {&dataType};
+}
+
+#endif // MLIR_CAPI_TTCOREENUMS_H
```
vmilosevic: :warning: **llvm\-header\-guard** :warning:
```suggestion
#endif // TTMLIR_CAPI_TTCOREENUMS_H
```

lib/CAPI/TTCoreAttrs.cpp
tapspatel: Agreed, I'll remove it.

test/python/capi/test_ttcore.py
brnorris03, line 1: Nice! Thanks for adding!

Issue #6417, opened 2025-12-22
[EmitPy] Extend emitpy dialect to support Python classes


sdjordjevicTT:
The current emitpy dialect is effective for emitting Python function calls and expressions, but it lacks first-class support for Python classes and instances. This limits our ability to:
- Represent object-oriented Python APIs (e.g. libraries with class-based interfaces)
- Emit idiomatic Python code without artificial temporaries
- Mirror the design philosophy of the [emitc](chatgpt://generic-entity?number=0) dialect for Python

To address this, we should extend emitpy with nominal class types, class definitions, and instance attribute/method access, while keeping the dialect syntax-preserving and semantics-free.

### Proposed Design

1. New Types
    - !emitpy.class<"Name">
Nominal Python class type
2. New / Extended Ops
    ```mlir
    # Class definition
    emitpy.class_def @MyClass {
    emitpy.method_def @__init__(%self, %x) { ... }
    emitpy.method_def @foo(%self) { ... }
    }
    # ...
    # Class instantiation
    # Possibly reuse emitpy.calll_opaque_op
    # ...
    # Attribute access
    %v = emitpy.attr_access %obj "field": (!emitpy.class<"MyClass">) -> !emitpy.object
    # ...
    # Attribute assignment
    emitpy.attr_assign %obj "field", %value
    # ...
    # Method call
    # Possibly reuse emitpy.calll_opaque_op
    # ...
    ```

PR #6649, opened 2026-01-15
[EmitPy] ClassOp

azecevicTT:
### Ticket
https://github.com/tenstorrent/tt-mlir/issues/6417

### Problem description
EmitPy could only represent Python at the function/statement level. It lacked a first-class representation for classes, instance fields, and method kinds, so there was no faithful way to model patterns like:
- load weights once in `__init__`
- store them on `self`
- expose a `forward()` method that uses those fields

This blocked emitting idiomatic Python classes for TTNN/TTIR models and forced everything into standalone functions with extra parameters.

### What's changed
Added operations to model class constructs directly:

- `emitpy.class` for class definitions with optional base list.
- `emitpy.get_attr` for attribute access (`obj.attr`).
- `emitpy.set_attr` for attribute assignment (`obj.attr = value`).

`emitpy.class` verification now enforces Python-like constraints:

- At most one `__init__`.
- `__init__` must return nothing.
- `emitpy.method_kind` on `func.func` is validated (`instance`, `staticmethod`,
  `classmethod`).
- Instance/class methods must have a receiver arg; if named, it must be
  `self` (instance) or `cls` (classmethod).
- Class body accepts only EmitPy ops or `func.func`.

Class bases use a Python-like syntax and only print `attributes {}` when extra
attrs exist:

```
emitpy.class @Model(#emitpy.opaque<"torch.nn.Module">) {
  ...
}

emitpy.class @NoBase {
  ...
}
```

Translator now emits:

- `class ...:` blocks from `emitpy.class`
- attribute access/assignment from `emitpy.get_attr` / `emitpy.set_attr`
- inline attribute access in `emitpy.expression`
- `self` as default first arg when inside class scope

```mlir
emitpy.class @Model(#emitpy.opaque<"torch.nn.Module">) {
  func.func @__init__(%self: !emitpy.class<"Model">) {
    %w = emitpy.call_opaque "ttnn.load_weight"() : () -> !emitpy.opaque<"ttnn.Tensor">
    emitpy.set_attr %self, "weight", %w : (!emitpy.class<"Model">, !emitpy.opaque<"ttnn.Tensor">)
    return
  }

  func.func @forward(%self: !emitpy.opaque<"Model">,
                     %input: !emitpy.opaque<"ttnn.Tensor"> {emitpy.name = "input"})
      -> !emitpy.opaque<"ttnn.Tensor"> {
    %w = emitpy.get_attr %self, "weight" : (!emitpy.class<"Model">) -> !emitpy.opaque<"ttnn.Tensor">
    %r = emitpy.call_opaque "ttnn.matmul"(%input, %w)
        : (!emitpy.opaque<"ttnn.Tensor">, !emitpy.opaque<"ttnn.Tensor">)
        -> !emitpy.opaque<"ttnn.Tensor">
    return %r : !emitpy.opaque<"ttnn.Tensor">
  }
}
```

```py
class Model(torch.nn.Module):
    def __init__(self):
        self.weight = ttnn.load_weight()

    def forward(self, input):
        return ttnn.matmul(input, self.weight)
```

Notes on some design decisions:
- methods are modeled with `func.func` op with additional attributes like `emitpy.method_kind`; an alternative was to introduce `emitpy.func` op, though it would a big overhead at the moment with little benefits
- methods are expected to have a first (receiver) parameter named `self` (or `cls` in a case of `classmethod`) with `emitpy.name`; While this isn't a strict requirement of the Python language it's convention that virtually any Python code adheres to. This should help with verification and the generated code will emit this names which is what all users would expect.

### Checklist

### DIFF
--- a/include/ttmlir/Dialect/EmitPy/IR/EmitPyOps.td
+++ b/include/ttmlir/Dialect/EmitPy/IR/EmitPyOps.td
@@ -166,7 +166,7 @@ def EmitPy_GetAttrOp : EmitPy_Op<"get_attr", [Pure, PyExpressionInterface]> {
     Example:
 
     ```mlir
-    %w = emitpy.get_attr %self, "weight" : (!emitpy.opaque<"Model">) -> !emitpy.opaque<"ttnn.Tensor">
+    %w = emitpy.get_attr %self, "weight" : (!emitpy.class<"Model">) -> !emitpy.opaque<"ttnn.Tensor">
     ```
     ```python
     # Code emitted:


lib/Target/Python/TranslateToPython.cpp
```diff
@@ -167,6 +168,19 @@ struct PythonEmitter {
...
+
+  /// RAII helper function to manage entering/exiting Python class scopes.
+  struct ClassScope {
+    ClassScope(PythonEmitter &emitter) : emitter(emitter) {
+      emitter.classDepth++;
+    }
+    ~ClassScope() { emitter.classDepth--; }
+
+  private:
+    PythonEmitter &emitter;
```
vmilosevic: :warning: **cppcoreguidelines\-avoid\-const\-or\-ref\-data\-members** :warning:
member `` emitter `` of type `` PythonEmitter & `` is a reference

include/ttmlir/Dialect/EmitPy/IR/EmitPyOps.td
```diff
@@ -157,6 +157,99 @@ def EmitPy_AssignOp : EmitPy_Op<"assign"> {
...
+        emitpy.set_attr %self, "weight", %w : (!emitpy.opaque<"Model">, !emitpy.opaque<"ttnn.Tensor">)
+        return
+      }
+    }
+    ```
+  }];
+
+  let arguments = (ins
+    SymbolNameAttr:$sym_name,
+    OptionalAttr<ArrayAttr>:$bases
```
svuckovicTT: I was confused by `bases` first read-through, would suggest renaming to `baseClasses`, or adding a comment somewhere in the definition.

include/ttmlir/Dialect/EmitPy/IR/EmitPyOps.td
```diff
@@ -157,6 +157,99 @@ def EmitPy_AssignOp : EmitPy_Op<"assign"> {
...
+def EmitPy_GetAttrOp : EmitPy_Op<"get_attr", [Pure, PyExpressionInterface]> {
+  let summary = "Get attribute from an object";
+  let description = [{
+    The `emitpy.get_attr` operation represents Python attribute access
+    via the dot operator.
+
+    Example:
+
+    ```mlir
+    %w = emitpy.get_attr %self, "weight" : (!emitpy.opaque<"Model">) -> !emitpy.opaque<"ttnn.Tensor">
```
svuckovicTT: `emitpy.opaque<"Model">` - is `Model` supposed to be the name of the class? Could we introduce a `ClassType`?

lib/Dialect/EmitPy/IR/EmitPyOps.cpp
```diff
@@ -158,6 +159,169 @@ LogicalResult isValidPythonIdentifier(Operation *op, StringRef name) {
...
+  int initCount = 0;
+  Block &body = getBody().front();
+  for (Operation &op : body.getOperations()) {
+    if (auto funcOp = dyn_cast<func::FuncOp>(op)) {
+      StringRef methodKind = "instance";
+      if (auto methodKindAttr =
+              funcOp->getAttrOfType<StringAttr>("emitpy.method_kind")) {
+        methodKind = methodKindAttr.getValue();
+        if (methodKind != "instance" && methodKind != "staticmethod" &&
+            methodKind != "classmethod") {
```
svuckovicTT, line 197:
```suggestion
        if (methodKind not in ["instance", "staticmethod", "classmethod"]) {
```

I find this more readable, but it does construct a whole list object - leaving it as a soft suggestion

svuckovicTT (approved):
Looks great! Would like to understand the class type modelling before merging.

include/ttmlir/Dialect/EmitPy/IR/EmitPyOps.td
azecevic000: Yeah, I agree, I will change the name to `baseClasses`.

lib/Dialect/EmitPy/IR/EmitPyOps.cpp
azecevic000: I think you switched languages here :D I also prefer such construct in Python, but there isn't a simple oneliner in C++ that I'm aware of, so for a small collection of values this is probably the most readable solution.

include/ttmlir/Dialect/EmitPy/IR/EmitPyOps.td
azecevic000: Discussed offline, `ClassType` is definitely a preferable solution here.

lib/Dialect/EmitPy/IR/EmitPyOps.cpp
svuckovicTT: Oh lol, that's funny, I have indeed imagined python land 😄

Issue #6311, opened 2025-12-12
[D2M] Decomp tile_mask_boundary op for partial tiles


vwellsTT:
### Is your feature request related to a problem? Please describe

In previous tickets we handle decomp for full-tile cases; we also need to handle the more complex partial tile cases.

### Describe the solution you'd like

For partial tile cases, the decomp pass must do more complicated arithmetic; loose plan is to create hardcoded mask tiles in place in the IR, perform appropriate actions to cancel out padded data and fill it with given oob_val's.

PR #6655, opened 2026-01-16
[D2M] Support partial masking 

vwellsTT:
### Ticket
[#6311](https://github.com/tenstorrent/tt-mlir/issues/6311)

### Problem description
We need to support filling oob_values that are not aligned with tile boundaries; we maybe need to fill sections of a tile in 3 possible ways: row-edge, col-edge, or corner (row-edge and col-edge)

### What's changed
This ended up being a bit heftier than I hoped, I can try to decompose for reviewing if requested.  Key points:
* DecomposeMasking.cpp -- rework to support partial tiles using single generic op with multiple loops to handle cases.  We use dynamic loop bounds to shard across cores, separate loop regions for various cases (pass through full tiles, pass through w/ row mask, pass through w/ col mask, pass through with double mask for corner, pass through pure padding along rows, pass through pure padding along cols). Use 2 scratch CB inputs (1 tile each) for the masks themselves.
* InsertDstRegisterAccess.cpp -- a bunch of changes here to support flattened dst s.t. my scratch tiles being different size than inputs isn't a problem, make managing dst tiles a bit simpler hopefully.  Allow both scf.for and affine.for, and memref.load/store and affine.load/store in this pass as well--found dynamic loops easier to express as non-affine, though this could be cleaned up a bit
* D2MOps.td -- added a "scratch_cbs" field to generic ops--these indicate input cbs which allocator should NOT stream.  This required some minor changes to a few other files where generic ops are constructed.

### Checklist
- [x] Fix double acquire_dst bug


lib/Dialect/D2M/Transforms/DecomposeMasking.cpp
```diff
@@ -30,222 +43,365 @@ static double getOOBValAsFloat(ttcore::OOBVal oobVal) {
...
+    // Global region: [0, lastValidRow+1) x [lastValidCol+1, totalTileCols)
+    // =========================================================================
+    {
+      rewriter.setInsertionPointAfter(insertionPoint);
+      auto [rowStart, rowEnd] = computeLocalBounds(
+          rewriter, loc, 0, lastValidRow + 1, coreY, shardTileRows);
+      auto [colStart, colEnd] = computeLocalBounds(
+          rewriter, loc, lastValidCol + 1, totalTileCols, coreX, shardTileCols);
+      auto loop = createLocalLoop(rowStart, rowEnd, colStart, colEnd, emitFill);
+      insertionPoint = loop;
```
vmilosevic: :warning: **clang\-analyzer\-deadcode\.DeadStores** :warning:
Value stored to `` insertionPoint `` is never read

lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
```diff
@@ -768,6 +927,10 @@ class D2MSFPUOpsRewriter : public OpConversionPattern<ConcreteOp> {
               loc, rewriter.getI32Type(), adaptor.getRhs());
           rewriter.create<ttkernel::PowUnaryTileOp>(loc, dstIdx, scalarParam);
         }
+        // Scalar ops operate in-place on DST slot - replace with the same
+        // dstIdx
+        rewriter.replaceOp(op, dstIdx);
+        return success();
       } else {
         // Binary tile operation
```
vmilosevic: :warning: **llvm\-else\-after\-return** :warning:
```suggestion
      }         // Binary tile operation
```

lib/Dialect/D2M/Transforms/InsertDstRegisterAccess.cpp
```diff
@@ -52,44 +271,71 @@ struct D2MInsertDstRegisterAccessRewriter final
       : OpRewritePattern<GenericOp>(ctx), useTileMatmul(useTileMatmul),
         maxDstPhysicalSizeTiles(maxDstPhysicalSizeTiles) {};
 
-  // Records a CB<->DST affine.load/store op, which DST slice it accesses, and
+  // Records a CB<->DST load/store op, which DST slice it accesses, and
   // some special considerations for looping over the tensor shard while doing
-  // DST accumulation/broadcast. The CB->load->bcast->DST sequence is also
-  // modeled as a CB->DST load.
-  template <typename LoadOrStoreTy>
```
vwellsTT, line 59: @wenbinlyuTT ~~btw I think I'm sort of reverting some of your changes with this, partially due to merge conflicts--I can switch to using a template of variant approach if you like, but kinda prefer not templating for something like this (where we take in 1 of 2 options) myself, lmk what you think.~~
Ended up redoing rebase and do use templates, but still might be good to get your review on this file FYI

include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
```diff
@@ -738,30 +770,38 @@ def D2M_BlockMaskOp : D2M_GenericRegionComputeOp<"block_mask",
...
+               memref<1x1x!ttcore.tile<32x32, f32>>, memref<1x1x!ttcore.tile<32x32, f32>>)
         ```
     }];
 
     let arguments = (ins
-        AnyRankedTensorOrMemRef:$input,
-        AnyRankedTensorOrMemRef:$output,
+        AnyTypeOf<[AnyRankedTensorOrMemRef, D2M_CB]>:$input,
+        AnyTypeOf<[AnyRankedTensorOrMemRef, D2M_CB]>:$output,
+        Optional<AnyTypeOf<[AnyRankedTensorOrMemRef, D2M_CB]>>:$row_mask_cb,
```
vwellsTT: My original plan was to make these optional and pass iff we need (i.e. if tile is NOT tiled aligned on a given dim), but that makes things more complex so for now I always pass 'em and just fill with nothing as needed.  Not sure if I should change back or leave as optional for future optimization work.

include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
```diff
@@ -747,6 +747,37 @@ def D2M_TileTypecastOp : D2M_GenericRegionComputeUnaryDstOp<"tile_typecast", [D2
 // D2M Masking Ops
 //===----------------------------------------------------------------------===//
 
+def D2M_ExperimentalTileFillOp : D2M_GenericRegionComputeOp<"experimental::tile_fill"> {
```
nsmithtt: In the d2m dialect I don't think we need to mark anything experimental, that's purely a ttkernel thing.

include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
```diff
@@ -1719,4 +1759,42 @@ def D2M_GetGlobalOperandOp : D2M_GenericRegionOp<"get_global_operand", [Pure]> {
     let assemblyFormat = [{ `(` $operand_index `)` attr-dict `:` type($result) }];
 }
 
+//===----------------------------------------------------------------------===//
+// D2M Experimental Index Tile Ops (Used for partial tile masking)
+//===----------------------------------------------------------------------===//
+
+def D2M_ExperimentalWriteRowMaskTileOp : D2M_GenericRegionOp<"experimental::write_row_mask_tile",
```
nsmithtt: Same here, no need for experimental

lib/Dialect/D2M/Transforms/DecomposeMasking.cpp
```diff
@@ -30,222 +43,371 @@ static double getOOBValAsFloat(ttcore::OOBVal oobVal) {
...
+    Value globalCoreStart =
+        rewriter.create<arith::MulIOp>(loc, coreIdx, shardSizeVal);
+
+    Value globalRegionStartVal =
+        rewriter.create<arith::ConstantIndexOp>(loc, globalRegionStart);
+    Value globalRegionEndVal =
+        rewriter.create<arith::ConstantIndexOp>(loc, globalRegionEnd);
+
+    // localStart = max(globalRegionStart - globalCoreStart, 0).
+    //            = globalRegionStart - min(globalRegionStart, globalCoreStart).
```
nsmithtt: Do we still need these comments?

include/ttmlir/Dialect/D2M/IR/D2MOps.td
```diff
@@ -245,7 +245,8 @@ def D2M_GenericOp : D2M_Op<"generic",
                          I64ArrayAttr:$block_factors,
                          AffineMapArrayAttr:$indexing_maps,
                          TTCore_IteratorTypeArrayAttr:$iterator_types,
-                         D2M_ThreadArrayAttr:$threads);
+                         D2M_ThreadArrayAttr:$threads,
+                         OptionalAttr<DenseI64ArrayAttr>:$scratch_inputs);
```
nsmithtt: FYI @jdesousa-TT / @wenbinlyuTT / @xanderchin, this could be a stop-gap solution for supplying a generic op with scratch inputs.

lib/Dialect/D2M/Transforms/DecomposeMasking.cpp
vwellsTT: This is sort of my pseudo-code explanation of the bounds checks, not commented out code, but I'll make it clearer 👍

include/ttmlir/Dialect/D2M/IR/D2MOps.td
vwellsTT: Yeah, curious to get more thoughts on this--from Fusion WG meeting, perhaps having only a single scratch input is preferred? I could pretty easily refactor to have a 2-tile scratch instead of 2 1-tile scratches I guess

lib/Dialect/D2M/Transforms/InsertDstRegisterAccess.cpp
```diff
@@ -285,20 +301,32 @@ struct D2MInsertDstRegisterAccessRewriter final
...
+        if (auto affineFor = dyn_cast<affine::AffineForOp>(op)) {
+          if (affineFor->hasAttr("d2m.linalg_root")) {
+            loopOp = affineFor;
+            loopRegion = &affineFor.getRegion();
+          }
+        } else if (auto scfFor = dyn_cast<scf::ForOp>(op)) {
+          if (scfFor->hasAttr("d2m.linalg_root")) {
+            loopOp = scfFor;
+            loopRegion = &scfFor.getRegion();
+          }
```
nsmithtt: nit: this could probably be simplified:
```mlir
          if (mlir::isa<affine::AffineForOp, scf::ForOp> && op->hasAttr("d2m.linalg_root")) {
            loopOp = op;
            loopRegion = &op->getRegion(0);
          }
```

lib/Dialect/D2M/Transforms/InsertDstRegisterAccess.cpp
```diff
@@ -371,8 +405,13 @@ struct D2MInsertDstRegisterAccessRewriter final
...
         types.hasLinalgGeneric = true;
-      } else if (auto forOp = dyn_cast<affine::AffineForOp>(op)) {
-        if (forOp->hasAttr("d2m.linalg_root")) {
+      } else if (auto affineFor = dyn_cast<affine::AffineForOp>(op)) {
+        if (affineFor->hasAttr("d2m.linalg_root")) {
+          types.hasMarkedAffineLoops = true;
+        }
+      } else if (auto scfFor = dyn_cast<scf::ForOp>(op)) {
+        // Also check scf.for loops (scheduled path).
+        if (scfFor->hasAttr("d2m.linalg_root")) {
```
nsmithtt, line 414: Can probably be simplified in the same way

nsmithtt (approved):
I think the dst changes are probably OK, would like to get some additional eyes on it from @wenbinlyuTT and @dlokeTT who have worked with the pass more recently.

nsmithtt:
Be sure to sync with @bgrady-tt for coordinating landing this w/ the unified thread PR

vwellsTT:
> Be sure to sync with @bgrady-tt for coordinating landing this w/ the unified thread PR

Yeah I think it's probably my turn to let him merge first and then deal with rebase fallout on my end this time 😄

lib/Dialect/D2M/Transforms/InsertDstRegisterAccess.cpp
```diff
@@ -580,8 +642,12 @@ struct D2MInsertDstRegisterAccessRewriter final
       auto iteratorTypes = gOp.getIteratorTypesValue();
 
       for (int64_t dim : nonParticipatingLoopDims) {
-        TT_assert(iteratorTypes[dim] == ttcore::IteratorType::Reduction);
-        guardDims.insert(dim);
+        // Only add reduction dims to guardDims - these need loop guards to
+        // ensure CB<->DST copies happen at the correct iteration.
+        // Parallel non-participating dims are broadcasts and don't need guards.
```
wenbinlyuTT: I am a bit confused about this. Broadcasts need loop guards and they're being captured by the function below (although there's a bug in its impl atm and the guard isn't guarding on the tiling loops).
I'm assuming you changed the assert because you are seeing parallel non-participating dims here and they are not actually broadcasts?

vwellsTT: yeah, I was hitting this assert for sure w/o broadcasts--don't remember exactly why I thought this was the fix though tbh (definitely partially LLM-driven).  Thinking about this now though, I think probably the real issue that for my loops, we aren't really using loop dims in traditional way, which => loop dims are considered non-participating and not-reduction, and hitting this assert.  I guess I should try to find some other short-circuit condition for my case s.t. we avoid this check (or somehow classify dims as `participating`, but that's somewhat painful for my case where we have custom looping logic inside the generic instead of top-level loops)

I'll see if maybe I can just skip this check on scratch inputs, theorize that was the problem

include/ttmlir/Dialect/D2M/IR/D2MOps.td
bgrady-tt: Fwiw we will need 1 scratchpad per dtype, so being able to support multiple is better.

I think using the same scratchpad for masking and general fusion is probably fine for initial bringup. Not sure if these two use cases be unified; since required space for masking is really small having it be seperate seems better.

bgrady-tt: @ckaravasilisTT you should take a look at this 👆

Issue #6561, opened 2026-01-13
Failed to legalize operation 'stablehlo.scatter' in densenet xray variant


kamalrajkannan78:
### Describe the bug

- densenet121_xray failed due to `error: failed to legalize operation 'stablehlo.scatter' ` - [jan13_densenet_xray.log](https://github.com/user-attachments/files/24582425/jan13_densenet_xray.log)
- Issue originates from this [masked scatter](https://github.com/mlmed/torchxrayvision/blob/df3c84c8c31bebbeb8696d400c862e615110bdfc/torchxrayvision/models.py#L499C5-L499C75) operation. - [jan13_masked_scatter.log](https://github.com/user-attachments/files/24582427/jan13_masked_scatter.log)
- Problematic SHLO IR - [shlo.mlir.zip](https://github.com/user-attachments/files/24582581/shlo.mlir.zip)

### Steps to reproduce the issue

- Follow below steps on tt-xla.

```
git checkout kkannan/jan13_densenet_xray_issue
git submodule update --init --recursive

# model run 
pytest tests/runner/test_models.py::test_all_models_torch[densenet/pytorch-densenet121_xray-single_device-inference] -svv --runxfail 

# sanity run 
pytest tests/torch/ops/test_masked_scatter.py -svv
```

kamalrajkannan78:
I think this issue is similar to https://github.com/tenstorrent/tt-mlir/issues/6555

@sdjordjevicTT could you pls take a look?

sdjordjevicTT:
> I think this issue is similar to [#6555](https://github.com/tenstorrent/tt-mlir/issues/6555)
> 
> [@sdjordjevicTT](https://github.com/sdjordjevicTT) could you pls take a look?

This is now under @mrakitaTT ownership.

PR #6657, opened 2026-01-16
Fix stablehlo.scatter validity logic

acicovicTT:
## Issue

https://github.com/tenstorrent/tt-mlir/issues/6561

## Description

Fix `inserted_window_dims` validation in StableHLO `ScatterOp` conversion to correctly follow the StableHLO spec. The previous validation logic checked `inserted_window_dims` against `rank(updates)` instead of `rank(inputs)`.

#### Fix (ref [StableHLO Scatter Specification](https://github.com/openxla/stablehlo/blob/main/docs/spec.md#scatter))

- (C8): Validate `update_window_dims` against `rank(updates)` (unchanged)
- (C11): Validate `inserted_window_dims` against `rank(inputs)` (fixed)
- (C2): Replace the overlap/coverage logic with the proper spec constraint: `rank(inputs) = size(update_window_dims) + size(inserted_window_dims) + size(input_batching_dims)`

### DIFF
--- a/lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
+++ b/lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
@@ -3774,7 +3774,8 @@ private:
     RankedTensorType indexType = op.getScatterIndices().getType();
     ArrayRef<int64_t> indexShape = indexType.getShape();
 
-    // Check update_window_dims (C8): indices into updates tensor.
+    // Verify openxla.org/stablehlo/spec#scatter (C8):
+    // indices into updates tensor.
     for (auto dim : updateWindowDims) {
       if (dim < 0 || dim >= updateRank) {
         return rewriter.notifyMatchFailure(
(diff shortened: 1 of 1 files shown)


lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
```diff
@@ -3771,38 +3774,29 @@ class StableHLOToTTIRScatterOpConversionPattern
     RankedTensorType indexType = op.getScatterIndices().getType();
     ArrayRef<int64_t> indexShape = indexType.getShape();
 
-    // Create array to track which dimensions are covered.
-    llvm::SmallVector<bool> dimsCovered(updateRank, false);
-
-    // Check update_window_dims.
+    // Check update_window_dims (C8): indices into updates tensor.
```
ajakovljevicTT: nit: can we also link the stablehlo.scatter doc so that C8 makes sense in this context?

acicovicTT: Done

Issue #6663, opened 2026-01-16
True Multihost Support


jameszianxuTT:
A tracking issue related to bugs / missing features related to a true multihost setup. 

"true" multihost refers to a multihost setup on two different physical machines connected over SSH, unlike the current tests which run on a single LLMBox with shared NIC + FS on wormhole.

jameszianxuTT:
This is also related to https://github.com/tenstorrent/tt-xla/issues/2052?issue=tenstorrent%7Ctt-xla%7C2983

PR #6664, opened 2026-01-16
Multihost bugfixes from 2xBH QB bringup

jameszianxuTT:
### Ticket
Partial resolution to #6663 

### Problem description
Some missing features needed for true multihost (using MPI over ssh) on two different machines

### What's changed
- Forward controller hostname to worker executable to support true remote multihost
- Pybind MCA options using string/string map
- Enable disable_eth_dispatch_on_blackhole for subprocess_get_system_descriptor to avoid segfault on blackhole arch
- Change `--map-by rankfile:file=` to `--rankfile` to avert MPI parsing bug with global filepaths

### Checklist

### DIFF
--- a/runtime/test/ttnn/python/utils.py
+++ b/runtime/test/ttnn/python/utils.py
@@ -157,15 +157,20 @@ def subprocess_get_system_descriptor(request, disable_eth_dispatch_on_blackhole=
     folder_name = "-".join([request.fspath.basename, request.node.name, "artifacts"])
     artifacts_dir = f"{os.getcwd()}/{folder_name}"
 
+    ttrt_query_cmd = [
+        "ttrt",
+        "query",
+        "--save-artifacts",
+        "--artifact-dir",
+        artifacts_dir,
+    ]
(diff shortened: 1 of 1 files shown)


runtime/python/runtime/runtime.cpp
```diff
@@ -131,8 +131,17 @@ void registerRuntimeBindings(nb::module_ &m) {
...
+          "with_mca_options",
+          [](tt::runtime::MultiProcessArgs &self, nb::dict mcaOptions) {
+            std::map<std::string, std::string> mcaOptionsMap;
+            for (auto [key, value] : mcaOptions) {
+              mcaOptionsMap[nb::cast<std::string>(key)] =
+                  nb::cast<std::string>(value);
+            }
+            return self.withMcaOptions(mcaOptionsMap);
+          },
+          nb::rv_policy::reference_internal)
```
jameszianxuTT, line 144: Note - hecking this cast, not sure if nanobind can automatically handle it

PR #6667, opened 2026-01-18
fix print_ir in builder for all pipelines

dlokeTT:
### Problem description
Print_ir being enabled wasn't printing ir for generic pipelines in builder

### What's changed

### Checklist

### DIFF
--- a/tools/builder/base/builder_apis.py
+++ b/tools/builder/base/builder_apis.py
@@ -1073,7 +1073,7 @@ def compile_stablehlo_to_flatbuffer(
     # We need to generate golden dictionary before pipeline run because pipeline run modifies the graph in place.
     input_output_goldens, intermediate_goldens = builder.golden_map
 
-    stablehlo_pipeline(module, " ".join(shlo_pipeline_options), print_ir=bool(print_ir))
+    stablehlo_pipeline(module, " ".join(shlo_pipeline_options), print_ir=print_ir)
     print(f"`{fn.__name__}` successfully ran stablehlo-pipeline.")
     print(module)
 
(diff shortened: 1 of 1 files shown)


tools/builder/base/builder_apis.py
```diff
@@ -1051,7 +1051,7 @@ def compile_stablehlo_to_flatbuffer(
     # We need to generate golden dictionary before pipeline run because pipeline run modifies the graph in place.
     input_output_goldens, intermediate_goldens = builder.golden_map
 
-    stablehlo_pipeline(module, " ".join(shlo_pipeline_options))
+    stablehlo_pipeline(module, " ".join(shlo_pipeline_options), print_ir=bool(print_ir))
```
tapspatel: do we need to typecast this? I see its a union of str + bool, but im guessing it should only be bool

tools/builder/base/builder_apis.py
```diff
@@ -1060,7 +1060,9 @@ def compile_stablehlo_to_flatbuffer(
         with open(filename, "w") as f:
             f.write(str(module))
 
-    stablehlo_to_ttir_pipeline(module, " ".join(shlo_to_ttir_pipeline_options))
+    stablehlo_to_ttir_pipeline(
+        module, " ".join(shlo_to_ttir_pipeline_options), print_ir=bool(print_ir)
```
tapspatel: typecast

tapspatel (commented):
thanks for addressing feedback! one small comment, otherwise nice stuff

python/Passes.cpp
```diff
@@ -69,10 +69,18 @@ void populatePassesModule(nb::module_ &m) {
 
   m.def(
       "stablehlo_pipeline",
```
nsmithtt, line 71: At some point it might be better to remove all of these nanobinded pipeline calls and across the board use (which can accept any pass or pipeline from string):
https://github.com/tenstorrent/tt-mlir/blob/a0beec115d1cc51690f4035b46bfae41259a9bf7/tools/builder/base/builder_utils.py#L125

dlokeTT: Yeah true. Tried doing this with cursor and ran into some issues with jit packaging. I'll try and look at this later when I have some free time to come up with a cleaner approach.

PR #6669, opened 2026-01-19
Adding Builder API args as pytest flags

jgrimTT:
### Ticket
Link to Github Issue

### Problem description
`compile_and_execute_ttir` args like `save_artifacts` have to be set in the code. 

### What's changed
Added boolean builder API arguments as pytest flags and wrote `get_request_kwargs` in `conftest.py` to handle pytest flags in tests .

```
    compile_and_execute_ttir(
        module,
        **get_request_kwargs(request),
        device=device,
    )
```

### Checklist

### DIFF
--- a/.github/get-docker-tag.sh
+++ b/.github/get-docker-tag.sh
@@ -5,6 +5,6 @@
 
 # Calculate hash from the following files. This hash is used to tag the docker images.
 # Any change in these files will result in a new docker image build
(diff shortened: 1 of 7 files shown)


wenbinlyuTT:
Thanks! This is much needed!
There're quite a few builder tests that hard code `print_ir = True` or some other value, can we also remove them here? Just to make things more predictable.

Issue #6637, opened 2026-01-15
Enable CPU-hoisted ops on host-sharded tensors


dmilinkovicTT:
As observed by @agobeljicTT, ViT + LoRA fails in multichip environment (n300):

```
[32m            RuntimeTTNN[0m | [1m[38;5;240m   DEBUG[0m | Executing operation: %1184 = "ttnn.reshape"(%1183) <{shape = [5 : i32, 1 : i32, 768 : i32]}> : (tensor<5x768xf32, #ttnn.ttnn_layout<(d0, d1) -> (d0, d1), <1x1>, memref<1x24x!ttcore.tile<32x32, f32>, #ttnn.buffer_type<dram>>, <interleaved>>>) -> tensor<5x1x768xf32, #ttnn.ttnn_layout<(d0, d1, d2) -> (d0 * 32 + d1, d2), <1x1>, memref<5x24x!ttcore.tile<32x32, f32>, #ttnn.buffer_type<dram>>, <interleaved>>> loc("reshape.3856")
[32m            RuntimeTTNN[0m | [1m[38;5;240m   DEBUG[0m | Executing operation: "ttnn.deallocate"(%1183) <{force = false}> : (tensor<5x768xf32, #ttnn.ttnn_layout<(d0, d1) -> (d0, d1), <1x1>, memref<1x24x!ttcore.tile<32x32, f32>, #ttnn.buffer_type<dram>>, <interleaved>>>) -> () loc("reshape.3856")
[32m            RuntimeTTNN[0m | [1m[38;5;240m   DEBUG[0m | Executing operation: %1185 = "ttnn.to_layout"(%1184) <{layout = #ttnn.layout<row_major>}> : (tensor<5x1x768xf32, #ttnn.ttnn_layout<(d0, d1, d2) -> (d0 * 32 + d1, d2), <1x1>, memref<5x24x!ttcore.tile<32x32, f32>, #ttnn.buffer_type<dram>>, <interleaved>>>) -> tensor<5x1x768xf32, #ttnn.ttnn_layout<(d0, d1, d2) -> (d0 + d1, d2), <1x1>, memref<5x768xf32, #ttnn.buffer_type<dram>>, <interleaved>>> loc(unknown)
[...]

PR #6672, opened 2026-01-19
Support multi-device tensors in CPU-hoisted ops

dmilinkovicTT:
### Ticket
#6637 

### Problem description
Execution of CPU-hoisted functions doesn't support multi-device tensors, since `getRawHostDataPtr` doesn't make sense for sharded/replicated tensors - there is no single host buffer, but rather, multiple host buffers logically representing the mesh sharded/replicated tensors.

This PR introduces basic support for CPU-hoisted ops execution in TTNN runtime by:
- Collecting the input shards of multi-device input tensors
- Performing the CPU-hoisted function on each shard
- Aggregating the output shards into a single multi-device tensor for each output

**Note**: Multi-device tensors are currently created only by using `tt::runtime::createMultiDeviceHostTensor`, which always creates a sharded tensor, even if the strategy is set to `replicate`. That's why we need to perform CPU-hoisted computations on all "shards" (even if the "shards" are actually "replicas") - we can't differentiate between the two with the existing APIs.

### Checklist

### DIFF
--- a/runtime/test/ttnn/python/n300/test_multichip_cpu_hoisting.py
+++ b/runtime/test/ttnn/python/n300/test_multichip_cpu_hoisting.py
@@ -1,0 +1,132 @@
+# SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+
+import os
+import ttrt
+import ttrt.runtime
+import torch
+from ttrt.common.util import *
+from .constants import FLATBUFFER_BASE_PATH
+
+from ..utils import (
+    Helper,
+    DeviceContext,
+    assert_pcc,
+    get_to_layout_inputs,
+    get_runtime_tensor_from_torch,
+)
+
+MESH_SHAPE = [1, 2]
+
+
+def get_input_spec(program, input_index):
+    """Extract shape and dtype from program input specification."""
+    program_input = program.inputs[input_index]
+    shape = program_input["desc"]["shape"]
+    dtype = Binary.Program.from_data_type(
+        program_input["desc"]["layout"]["memory_desc"]["data_type"]
+    )
+    return shape, dtype
+
+
+def create_multi_device_input(shape, dtype):
+    """Create a multi-device host tensor with random data per shard."""
+    runtime_dtype = Binary.Program.to_data_type(dtype)
+    num_devices = MESH_SHAPE[0] * MESH_SHAPE[1]
+    torch_shards = [torch.randn(shape, dtype=dtype) for _ in range(num_devices)]
+
+    runtime_tensor = ttrt.runtime.create_multi_device_host_tensor(
+        [shard.data_ptr() for shard in torch_shards],
+        list(shape),
+        list(torch_shards[0].stride()),
+        torch_shards[0].element_size(),
+        runtime_dtype,
+        {},
+        MESH_SHAPE,
+    )
+    return runtime_tensor, torch_shards
+
+
+def create_single_device_input(shape, dtype):
+    """Create a single-device host tensor with random data."""
+    torch_tensor = torch.randn(shape, dtype=dtype)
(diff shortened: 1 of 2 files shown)


runtime/lib/ttnn/operations/cpu/cpu.cpp
```diff
@@ -43,32 +74,171 @@ void run(const ::tt::target::ttnn::CpuOp *op, ProgramContext &context) {
...
+              return ::tt::runtime::ttnn::utils::getRawHostDataPtr(
+                  input[shardIdx]);
+            } else {
+              // Single-device tensor: use the same data for all shards.
+              return ::tt::runtime::ttnn::utils::getRawHostDataPtr(input);
+            }
+          },
+          inputTensors[i]);
+      inputDataPtrs.push_back(dataPtr);
+    }
```
jameszianxuTT, line 187: nit: This loop, and the use of std::variant above, seem a little bit overcomplicated. I imagine you could rearrange this loop so the outer loop goes over the input index, and the inner loop goes over the shard index.

I think you could even do away with the variant vector and use a vector<vector<::ttnn::Tensor>> and for the replicated case just use a size 1 vector. 

I'm not sure exactly how partial sharding is represented, but if it's a case where `::ttnn::distributed::get_device_tensors` returns a number of shards less than the number of devices, then using a `vector<vector<tensor>>` seems like it would make it easier to handle those cases as you could just loop over the number of shards in the inner loop, rather than forcing this to the number of devices in the outer loop.

---

Note - this is just a nit for readability/simplicity. I'm not sure what would be the best practice here w/r/t perf, so if this comment makes no sense feel free to disregard :)

runtime/lib/ttnn/operations/cpu/cpu.cpp
```diff
@@ -43,32 +74,171 @@ void run(const ::tt::target::ttnn::CpuOp *op, ProgramContext &context) {
...
-  };
+// Executes CPU-hoisted operation for single-chip workloads.
+void runSingleChip(
+    common::WrappedFunc fn,
+    const flatbuffers::Vector<flatbuffers::Offset<tt::target::ttnn::TensorRef>>
+        *fbInputs,
+    const flatbuffers::Vector<flatbuffers::Offset<tt::target::ttnn::TensorRef>>
+        *fbOutputs,
+    ProgramContext &context) {
+  LOG_DEBUG("CPU op: executing single-chip path...");
```
jameszianxuTT: nit: I think this output log would spam the output without providing much useful information. Is this meant to be merged in?

jameszianxuTT:
Edit: Also, how is this path tested?

runtime/lib/ttnn/operations/cpu/cpu.cpp
dmilinkovicTT: Thanks for the feedback! I've simplified the implementation.

> I imagine you could rearrange this loop so the outer loop goes over the input index, and the inner loop goes over the shard index.

I assume you meant the other way around? We first need to collect a specific shard for all inputs, then execute the function, and then, move on to the next shard.

And regarding partial sharding - I'm not even sure it will ever happen in practice. From my limited experience with `tt-xla`, input tensors seem to always be sharded across the same number of devices.

But if it does, I still think this approach couldn't handle it well without adding quite a bit of complexity. For example, imagine:
- Input 0: unsharded
- Input 1: sharded across 2 chips
- Input 2: sharded across 4 chips

We would need to somehow create a mapping which would tell us which shards of input 1 correspond to which shards of input 2.

dmilinkovicTT:
> Edit: Also, how is this path tested?

Good point, I've added a runtime test.

runtime/lib/ttnn/operations/cpu/cpu.cpp
jameszianxuTT: Right, that makes sense. Thanks for the explanation and the refactor!

PR #6682, opened 2026-01-20
Add shlo compiler graphs for large models

btsoiTT:
Branch for shlo compiler graphs


btsoiTT:
The qwen and gpt oss graphs currently exceed the maximum file size allowed for the pre-commit. Am I allowed to change the limit (500KB)?

nsmithtt:
> The qwen and gpt oss graphs currently exceed the maximum file size allowed for the pre-commit. Am I allowed to change the limit (500KB)?

I didn't realize these would be so large, a few of them are 1.5MB, to avoid bloating the repo, maybe we should check these into a separate personal/public repo, or just leave them on this branch for folks.

Issue #2768, opened 2025-04-01
StableHLO to TTIR conversion fails for custom_call op


umalesTT:
When selecting indices of top k weights via jax.numpy.argpartition (relevant for EfficientNeRF training) SHLO->TTIR conversion fails with ``` error: failed to legalize operation 'stablehlo.custom_call' ```. Function being called is mhlo.topk and SHLO code is given below:

``` 
module @jit_argpartition attributes {mhlo.num_partitions = 1 : i32, mhlo.num_replicas = 1 : i32} {
  func.func public @main(%arg0: tensor<262144xf32> {mhlo.sharding = "{replicated}"}) -> tensor<262144xi32> {
    %cst = stablehlo.constant dense<0.000000e+00> : tensor<f32>
    %c = stablehlo.constant dense<262144> : tensor<i32>
    %c_0 = stablehlo.constant dense<0> : tensor<i32>
    %cst_1 = stablehlo.constant dense<1.000000e+00> : tensor<f32>
    %0 = stablehlo.negate %arg0 : tensor<262144xf32>
    %1:2 = stablehlo.custom_call @mhlo.topk(%0) {mhlo.attributes = {k = 242145 : i64, largest = true}, mhlo.version = 1 : i64} : (tensor<262144xf32>) -> (tensor<242145xf32>, tensor<242145xi32>)
    %2 = stablehlo.broadcast_in_dim %cst_1, dims = [] : (tensor<f32>) -> tensor<262144xf32>
    %3 = stablehlo.broadcast_in_dim %c_0, dims = [] : (tensor<i32>) -> tensor<242145xi32>
    %4 = stablehlo.compare  LT, %1#1, %3,  SIGNED : (tensor<242145xi32>, tensor<242145xi32>) -> tensor<242145xi1>
    %5 = stablehlo.broadcast_in_dim %c, dims = [] : (tensor<i32>) -> tensor<242145xi32>
    %6 = stablehlo.add %1#1, %5 : tensor<242145xi32>
    %7 = stablehlo.select %4, %6, %1#1 : tensor<242145xi1>, tensor<242145xi32>
    %8 = stablehlo.broadcast_in_dim %7, dims = [0] : (tensor<242145xi32>) -> tensor<242145x1xi32>
    %9 = stablehlo.broadcast_in_dim %cst, dims = [] : (tensor<f32>) -> tensor<242145xf32>
[...]

PR #6686, opened 2026-01-21
Support TopK Op

### LINKED ISSUE #3131: [Bounty $500] Add `TopK` op
- Add TTIR op: TTNN implementation seems reasonable to me, so I would model it as such (that's the next point)
- Add TTNN op: it should model https://github.com/tenstorrent/tt-metal/blob/main/ttnn/cpp/ttnn/operations/reduction/topk/topk.hpp without the last three parameters (`sub_core_grids`, `indices_tensor` and `optional_output_tensors`); for the results, it should have two `RankedTensorType`s (`result` and `indices`)
- Add lowering from StableHLO `custom_call` op with `mhlo.topk` as function to TTIR op (related issue https://github.com/tenstorrent/tt-mlir/issues/2768)
- Add flatbuffer schema and translation from TTNN to flatbuffer
- Add runtime support for the op
- Add OpModel for the op
- (optional): Add conversion from TTNN to EmitC
- (optional): Add conversion from TTNN to EmitPy
- (optional): Add an op to the `builder`: https://docs.tenstorrent.com/tt-mlir/builder/adding-a-ttir-op.html

### LINKED ISSUE #7100: Add topk op support to TTIR and TTNN dialects
Add e2e support for the `ttnn.topk` op in both TTIR and TTNN dialects.

This op selects the top-k values and their indices along a given dimension. It is used in MoE router logic (e.g. GPT OSS decode) to select the top-k experts per token.

**TTNN API reference:**
- Inputs: tensor `[B, E]`, `k` (int), `dim` (int)
- Outputs: values `[B, K]`, indices `[B, K]` (uint32)

**Scope:**
- Add `topk` op to TTIR dialect
- Add `topk` op to TTNN dialect
- TTIR → TTNN lowering
- Lit tests and Python tests

elhewaty:
Fixes: https://github.com/tenstorrent/tt-mlir/issues/3131

## Milestones

- [x] **Add TTIR op**: Model after TTNN implementation patterns
- [x] **Add TTNN op**: Model [tt-metal topk.hpp](https://github.com/tenstorrent/tt-metal/blob/main/ttnn/cpp/ttnn/operations/reduction/topk/topk.hpp) without last three parameters (sub_core_grids, indices_tensor, optional_output_tensors). Returns two RankedTensorTypes (result and indices)
- [x] **Add lowering from StableHLO**: Support custom_call op with mhlo.topk as function to TTIR op (related to #2768)
- [x] **Add flatbuffer schema**: Translation from TTNN to flatbuffer
- [x] **Add runtime support** for the op
- [x] **Add OpModel** for the op
- [x] **(optional)** Add conversion from TTNN to EmitC
- [x] **(optional)** Add conversion from TTNN to EmitPy
- [x] **(optional)** Add an op to the builder following [documentation](https://docs.tenstorrent.com/tt-mlir/builder/adding-a-ttir-op.html)


include/ttmlir/Dialect/TTIR/IR/TTIROps.td
```diff
@@ -5740,4 +5740,32 @@ def TTIR_SplitQueryKeyValueAndSplitHeadsOp: TTIR_NamedOp<"split_query_key_value_
...
+    The boolean option `sorted` if True, will make sure that the returned `k` elements are sorted.
+    the `memory_config` parameter is optional which represents the memory configuration for the operation.
+  }];
+
+  let arguments = (ins AnyRankedTensor:$input_tensor,
+                       UI32Attr:$k,
+                       DefaultValuedAttr<SI32Attr, "-1">:$dim,
+                       DefaultValuedAttr<BoolAttr, "true">:$largest,
+                       DefaultValuedAttr<BoolAttr, "true">:$sorted);
+                       // OptionalAttr<>:$memory_config); defer memory_config for now not sure of the type
```
azecevic000: `memory_config` shouldn't be modeled in TTIR

include/ttmlir/Dialect/TTIR/IR/TTIROps.td
```diff
@@ -5740,4 +5740,32 @@ def TTIR_SplitQueryKeyValueAndSplitHeadsOp: TTIR_NamedOp<"split_query_key_value_
...
+
+  let results = (outs AnyRankedTensor:$values,
+                 AnyRankedTensor:$indices);
+
+  let hasVerifier = 1;
+
+  let assemblyFormat = [{
+    $input_tensor `,` $k (`dim` $dim^)? (`largest` $largest^)? (`sorted` $sorted^)?
+    attr-dict `:` type($input_tensor) `->` type($values) `,` type($indices)
```
azecevic000: If we are going with a custom assembly format I would prefer `k=X, sorted=Y...`, but you can also leave default format.

lib/Dialect/TTIR/IR/TTIROps.cpp
```diff
@@ -4662,6 +4662,49 @@ ::mlir::LogicalResult mlir::tt::ttir::ArgMaxOp::verify() {
...
+// TopKOp verification
+::mlir::LogicalResult mlir::tt::ttir::TopKOp::verify() {
+  RankedTensorType inputType = getInputTensor().getType();
+  int64_t inputRank = inputType.getRank();
+
+  // ttnn.topk Limitations:
+  // Input tensor must be in 4D [N, C, H, W] format.
+  if (inputRank != 4) {
+    return emitOpError("input tensor must be in 4D [N, C, H, W] format");
+  }
```
azecevic000: TTNN limitations shouldn't leak into TTIR. For a lot of these constraints we can work around them when going from TTIR to TTNN. TTIR verification should only check theoretical (mathematical) constraints, e.g. `k is in range [-n, n)` etc.

elhewaty:
Hi @azecevicTT, I hope we can call this a little progress!

I have a question regarding testing the `Constraints` and `Runtime` APIs.
From the docs:

```
  auto constraintsExp = getOpConstraints(yourOp.getOperation());
  if (constraintsExp) {
      auto l1 = constraintsExp.get();
      const auto &[cbSize, l1PeakSize, totalPeakSize, outputSize, outputLayout] = l1;
      EXPECT_EQ(cbSize, /* some expected value */);
      EXPECT_EQ(l1PeakSize, /* some expected value */);
      EXPECT_EQ(totalPeakSize, /* some expected value */);
      EXPECT_EQ(outputSize, /* some expected value */);
  } else {
      FAIL() << "Missing L1 constraints; Error="
          << llvm::toString(constraintsExp.takeError()) << std::endl;
  }
```

what are these values `cbSize, l1PeakSize, totalPeakSize, outputSize, outputLayout`?
I mean what do they represent? 
and how can I expect their values[Maybe I will know after  get an idea of what are they]?

azecevic000:
@elhewaty @sgholamiTT can help you with this one

elhewaty:
@sgholamiTT , Any hint?

elhewaty:
@azecevicTT please sgare you thoughts on the current code?

sgholamiTT:
@elhewaty Thanks for your contribution. 🚀 

Regrading your question, `OpConstraints` struct is defined/documented as follows:
https://github.com/tenstorrent/tt-mlir/blob/8a0e9638113a0eef0b56f4425ade13248f982b7b/include/ttmlir/OpModel/TTNN/TTNNOpConstraints.h#L20-L25
Unfortunately, the document your'e referring to is a bit out of date. We have recently refactored our tests and just expect these values to be non-negative: 
- If the output tensor layout is determined to be in `L1`, they should be positive,
  - [This is an example](https://github.com/tenstorrent/tt-mlir/blob/8a0e9638113a0eef0b56f4425ade13248f982b7b/test/unittests/OpModel/TTNN/Lib/TestOpModelLib.cpp#L563-L570), note the `layoutL1Interleaved` arg passed to the constraint API.
- If the output tensor layout is determined to be in `DRAM`, they should be zero,
  - [This is an example](https://github.com/tenstorrent/tt-mlir/blob/8a0e9638113a0eef0b56f4425ade13248f982b7b/test/unittests/OpModel/TTNN/Lib/TestOpModelLib.cpp#L555-L561), note the `layoutDRAM` arg passed to the constraint API.

Please take a look at the current implementation of our tests [here](https://github.com/tenstorrent/tt-mlir/blob/main/test/unittests/OpModel/TTNN/Lib/TestOpModelLib.cpp) and [here](https://github.com/tenstorrent/tt-mlir/blob/main/test/unittests/OpModel/TTNN/Op/TestOpModelInterface.cpp) and follow the same pattern.

Let me know if you had any other question, and thanks again for your catribution.

elhewaty:
@sgholamiTT would you review the current implementation?

sgholamiTT:
The OpModel changes LGTM. But you should add UTs [here](https://github.com/tenstorrent/tt-mlir/blob/main/test/unittests/OpModel/TTNN/Lib/TestOpModelLib.cpp) and [here](https://github.com/tenstorrent/tt-mlir/blob/main/test/unittests/OpModel/TTNN/Op/TestOpModelInterface.cpp) for the op as I described in my previous comment.

elhewaty:
@sgholamiTT, Does it require a special hardware?

sgholamiTT:
@elhewaty Sorry, but I fail to understand your question.
In general, to be able to get the memory footprint of a `Program` on any hardware, yes, you probably need to have access to the actual hardware (so that you can literally run the `Program` and monitor its memory footprint). Tenstorrent is no exception too, and I believe we have ways of giving access to our hardware for bounty programs (@tapspatel is that correct?)

But in this case, I think it should be very trivial to "guess" the memory based on what I described in [this](https://github.com/tenstorrent/tt-mlir/pull/6686#issuecomment-3791430820) comment.
[This](https://github.com/tenstorrent/tt-mlir/blob/8a0e9638113a0eef0b56f4425ade13248f982b7b/test/unittests/OpModel/TTNN/Lib/TestOpModelLib.cpp#L541-L577) UT is a great example of how the memory footprint could change based on the fact that the output layout is set to be on `L1` or `DRAM`.

include/ttmlir/Dialect/TTIR/IR/TTIROps.td
```diff
@@ -5730,4 +5730,25 @@ def TTIR_SplitQueryKeyValueAndSplitHeadsOp: TTIR_NamedOp<"split_query_key_value_
...
+  }];
+
+  let arguments = (ins AnyRankedTensor:$input_tensor,
+                       I32Attr:$k,
+                       DefaultValuedAttr<I32Attr, "-1">:$dim,
+                       DefaultValuedAttr<BoolAttr, "true">:$largest,
+                       DefaultValuedAttr<BoolAttr, "true">:$sorted);
+
+  let results = (outs AnyRankedTensor:$values,
+                 AnyRankedTensor:$indices);
```
azecevic000: Nit (alignment):
```suggestion
  let results = (outs AnyRankedTensor:$values,
                      AnyRankedTensor:$indices);
```

lib/Dialect/TTIR/IR/TTIROps.cpp
```diff
@@ -4676,6 +4676,35 @@ ::mlir::LogicalResult mlir::tt::ttir::ArgMaxOp::verify() {
...
+
+  // Normalize dim to check if it's effectively the last dimension
+  int normalizedDim = dim < 0 ? dim + inputRank : dim;
+  if (normalizedDim < 0 || normalizedDim >= inputRank) {
+    return emitOpError() << "specified dimension should be between "
+                         << -inputRank << " and " << (inputRank - 1)
+                         << ", but got: " << dim;
+  }
+
+  int normalizedK = K < 0 ? K + inputRank : K;
```
azecevic000: IMO it doesn't make sense for k to be negative (for dim it's okay and standard practice).

lib/Dialect/TTNN/IR/TTNNOps.cpp
```diff
@@ -4672,4 +4672,33 @@ ::mlir::LogicalResult mlir::tt::ttnn::AssignOp::verify() {
...
+
+  // Normalize dim to check if it's effectively the last dimension
+  int normalizedDim = dim < 0 ? dim + inputRank : dim;
+  if (normalizedDim < 0 || normalizedDim >= inputRank) {
+    return emitOpError() << "specified dimension should be between "
+                         << -inputRank << " and " << (inputRank - 1)
+                         << ", but got: " << dim;
+  }
+
+  int normalizedK = K < 0 ? K + inputRank : K;
```
azecevic000: Same here.

test/ttmlir/Dialect/TTIR/basic_topk.mlir
azecevic000, line 1: TBH this test doesn't add any real value, you can remove it. You can add negative tests (i.e. tests that will fail on verification), there are few examples in this dir that you can check.

azecevic000 (commented):
@elhewaty This looks really good at the moment, I've left a few comments inline, keep up the good work!

elhewaty:
Thanks for the review, I will add more work today.

sgholamiTT (approved):
Approving for OpModel changes.
Thanks for your contribution @elhewaty 🚀

elhewaty:
while trying to test ttir.top_k: 
```
pytest test/python/golden/test_ttir_ops.py::test_topk
====================================================================== test session starts ======================================================================
platform linux -- Python 3.11.14, pytest-9.0.2, pluggy-1.6.0
rootdir: /root/tt-mlir/test/python/golden
configfile: pytest.ini
plugins: json-report-1.5.0, forked-1.6.0, metadata-3.1.1
collected 0 items / 1 error                                                                                                                                     

============================================================================ ERRORS =============================================================================
_______________________________________________________________ ERROR collecting test_ttir_ops.py _______________________________________________________________
test/python/golden/test_ttir_ops.py:14: in <module>
    from builder.ttir.ttir_builder import TTIRBuilder
build/python_packages/builder/ttir/ttir_builder.py:18: in <module>
    from builder.base.builder import *
build/python_packages/builder/base/builder.py:16: in <module>
    from golden import GoldenMapTensor, get_golden_function
build/python_packages/golden/__init__.py:4: in <module>
    from .mapping import *
build/python_packages/golden/mapping.py:5539: in <module>
    ttir.TopKOp: torch.topk,
    ^^^^^^^^^^^
E   AttributeError: module 'ttmlir.dialects.ttir' has no attribute 'TopKOp'
```

How can we expose ttir ops to python?

azecevic000:
@tapspatel @jgrimTT for the previous question

jgrimTT:
Adding it to `TTIROPs.td` should be the what gets the op generated in `build/python_packages/ttmlir/dialects/_ttir_ops_gen.py`. You typically have to `rm -rf build` and rebuild to get it to generate, have you done so?

elhewaty:
@ jgrimTT , I didn't know this.
Thanks, it works now.

elhewaty:
Hey @azecevicTT , 
Can you review the current progress?

elhewaty:
@azecevicTT would you run the CI and share your thoughts?

lib/Conversion/TTNNToEmitC/TTNNToEmitC.cpp
```diff
@@ -4416,6 +4416,36 @@ class PagedFillCacheOpConversionPattern
 };
 } // namespace
 
+namespace {
+class TTNNToEmitCTopKOpConversionPattern
```
sdjordjevicTT, line 4498: CI fails on TTNNToEmitC conversion. The code generates a call to a `ttnn::top_k` ttnn operation, which doesn't exist. You have two options here:
1. Rename TTIR\TTNN ops to be called topk instead of top_k:
```mlir
def TTNN_TopKOp : TTNN_Op<"top_k", [TTNN_MemoryConfigOpInterface]> {
->
def TTNN_TopKOp : TTNN_Op<"topk", [TTNN_MemoryConfigOpInterface]> {
```
2. In the conversion pattern override:
```C++
std::string getPrefixSwapPattern() const override { return "ttnn::topk"; }
```

To me, the 1. one option looks much better, as most frameworks define operation as `topk` without the _...

CI error:
```C++
//...
standalone/build/CMakeFiles/ttnn-dylib.dir/cmake_pch.hxx -MD -MT CMakeFiles/ttnn-dylib.dir/ttnn-dylib.cpp.o -MF CMakeFiles/ttnn-dylib.dir/ttnn-dylib.cpp.o.d -o CMakeFiles/ttnn-dylib.dir/ttnn-dylib.cpp.o -c /__w/tt-mlir/tt-mlir/tools/ttnn-standalone/ttnn-dylib.cpp
/__w/tt-mlir/tt-mlir/tools/ttnn-standalone/ttnn-dylib.cpp:4:44: error: no member named 'top_k' in namespace 'ttnn'
    4 |   ::std::vector<::ttnn::Tensor> v3 = ttnn::top_k(v2, 5, 4294967295, true, true, ::std::nullopt);
      |                                      ~~~~~~^
1 error generated.
ninja: build stopped: subcommand failed.
```

elhewaty: I think we should do it in the conversion pattern as well, as it still fails, right?

elhewaty: Hey @sdjordjevicTT,
I tried both approaches and it didn't work! so I reverted the second approach and it is now the first.
any hint why does it fails?

CC: @azecevicTT

elhewaty: Ping!

tools/builder/ttir/ttir_builder.py
```diff
@@ -13204,3 +13204,92 @@ def split_module(
...
+        in0: Operand,
+        k: int,
+        dim: int = -1,
+        largest: bool = True,
+        sorted: bool = True,
+        output_type: Optional[torch.dtype] = None,
+        loc: Optional[str] = None,
+        unit_attrs: Optional[List[str]] = None,
+    ) -> Tuple[OpResult, OpResult]:
+        """
```
jgrimTT: Nit: remove docstring

jgrimTT (changes requested):
This looks great on the builder side, it does still need a split and a parse function in `ttir_builder.py`

lib/Conversion/TTNNToEmitC/TTNNToEmitC.cpp
sdjordjevicTT: Could you please provide more details on how you verified that it isn't working?

elhewaty: I updated the PR, and re-run the CI:
It was converted to
```
ttnn::topktopk
```
Which is not found

tools/builder/ttir/ttir_builder.py
elhewaty: Great, thanks for review.
BTW, why do we need the split and parser functions?

jgrimTT: Ofc! Builder has APIs `load_mlir_file` and `split_mlir_file` which allow users to pass in a module as a string and have it converted into a builder object with golden tensors. The load/split functionality is useful for running golden checks, debugging, and pinpointing PCC errors without having to manually recreate a module in builder

lib/Conversion/TTNNToEmitC/TTNNToEmitC.cpp
sdjordjevicTT: Because it isn't correct. You should generate a solution that will produce:
```
ttnn::topk(...
```

To debug this, you do not need to rerun CI, target the ttir-to-emitc pipeline:
```
ttmlir-opt --ttnn-to-emitc-device-pipeline ...topk_test_file...
```

elhewaty: Yes to generate this we can enforce the generated code you mention, but the current implementation fails with 
```
ttnn::topk
```
As we can see in the CI, will enforcing the prefix change this?

elhewaty: The CI error

```
/__w/tt-mlir/tt-mlir/tools/ttnn-standalone/ttnn-dylib.cpp:4:44: error: no member named 'topk' in namespace 'ttnn'
    4 |   ::std::vector<::ttnn::Tensor> v3 = ttnn::topk(v2, 5, 4294967295, true, true, ::std::nullopt);
      |                                      ~~~~~~^
1 error generated.
ninja: build stopped: subcommand failed.
```

elhewaty: The current implementation is option 1.

sdjordjevicTT: Oh, I see, thanks for clarification, the generated code is correct, indeed. The thing that you need to do, is to include the required topk header file in all ttnn-precompiled.hpp files...

elhewaty: I will update the PR now, BTW, would you assign a next issue maybe a little harder?

elhewaty: BTW, it doesn't have to be a bounty.

elhewaty:
@sgholamiTT Can you check the failures here?
```
[ RUN      ] OpModelTest.TopK
/__w/tt-mlir/tt-mlir/test/unittests/OpModel/TTNN/Lib/TestOpModelLib.cpp:1602: Failure
Expected equality of these values:
  opCstr.outputL1BufferSize
    Which is: 4096
  0

[  FAILED  ] OpModelTest.TopK (40 ms)

```

tools/builder/ttir/ttir_builder.py
elhewaty: Hey @jgrimTT ,
Can we implement the builder/golden with values not values and indices?
the problem is: values works fine and the tests passes, but when returning the indices
tensor with the values I get pcc error [0.98 < 0.99] and for emitc [0.95 < 0.99],
I think the problem happens as the internal implementation of the kernel itself
is not the same in case of duplicated values, for example:

```
Input tensor:
tensor([[95, 95, 50, 30],
        [80, 80, 80, 20],
        [70, 65, 70, 60],
        [40, 40, 40, 40]])
```

which indices we choose for this case? do we choose first or last? and for torch which element
does it choose?

I updated the PR with the solution I mentioned.

test/unittests/OpModel/TTNN/Lib/TestOpModelLib.cpp
```diff
@@ -1569,6 +1569,45 @@ TEST_F(OpModelTest, Sort) {
...
+  EXPECT_EQ(opCstr.outputL1BufferSize, 0);
+
+  constraintsExp = op_model::OpModel<TopKOp>::getOpConstraints(
+      CreateWorkerGrid(), tensorShape, layoutDRAM, 4, -1, true, true,
+      layoutL1Interleaved);
+  EXPECT_TRUE(static_cast<bool>(constraintsExp));
+  opCstr = constraintsExp.get();
+  EXPECT_GT(opCstr.cbL1PeakSize, 0);
+  EXPECT_GT(opCstr.tensorL1PeakSize, 0);
+  EXPECT_EQ(opCstr.outputL1BufferSize, 0);
```
elhewaty: @sgholamiTT the CI fails at this line!

tools/builder/ttir/ttir_builder.py
elhewaty: @jgrimTT, reminder.

anuragsingh-tt: This is a great point @elhewaty. May I suggest modifying the golden function you have implemented to return a canonicalized form of indices (e.g. sort the indices per row/col sequentially 0, 1, ...) and do the same for the builder functions. That should normalize the comparison between golden values and device values.

test/unittests/OpModel/TTNN/Lib/TestOpModelLib.cpp
sgholamiTT: Yes, this means that the output buffer size is `4096` which is greater than 0. Please just change this line to:
```
EXPECT_GT(opCstr.outputL1BufferSize, 0);
```
And it should be good to go.

tools/builder/ttir/ttir_builder.py
jgrimTT: I'll let you decide for yourself how to handle that golden return, @anuragsingh-tt had a good suggestion. Slightly lower pcc happens sometimes for ops, and setting a custom pcc `pcc=0.95` in `compile_and_execute_ttir` is a valid solution for your test

elhewaty: @anuragsingh-tt  I couldn't find a way to sort op_indices in the builder, any hint?

elhewaty: Would you please run the CI? I need to see if there are any crashes.

elhewaty: Any hint why the CI fails because of a file that was not touched🤔?

lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
```diff
@@ -6506,6 +6506,68 @@ class StableHLOToTTIRScaledDotProductAttentionOpConversionPattern
...
+
+    // Convert i64 to i32
+    int32_t kValueI32 = static_cast<int32_t>(kValueI64.getSExtValue());
+    kAttr = IntegerAttr::get(rewriter.getIntegerType(32), kValueI32);
+
+    auto largestAttr = dictAttr.getAs<BoolAttr>("largest");
+    if (!largestAttr) {
+      return rewriter.notifyMatchFailure(
+          srcOp, "Missing largest attribute in mhlo.attributes");
+    }
```
azecevic000, line 6563: @umalesTT @sgligorijevicTT Based on [this](https://github.com/tenstorrent/tt-mlir/issues/2768) issue, `largest` is present in both examples, do you happen to know if that's always guaranteed? We default to `true` in TTIR; that's why I'm not sure if we should always expect it here, or default to `true` if it's not present.

include/ttmlir/Dialect/TTIR/IR/TTIROps.td
```diff
@@ -5741,4 +5741,25 @@ def TTIR_SplitQueryKeyValueAndSplitHeadsOp: TTIR_NamedOp<"split_query_key_value_
...
+    If `dim` is not provided, the last dimension of the input_tensor is used.
+    If `largest` is True, the `k` largest elements are returned. Otherwise, the `k` smallest elements are returned.
+    The boolean option `sorted` if True, will make sure that the returned `k` elements are sorted.
+  }];
+
+  let arguments = (ins AnyRankedTensor:$input_tensor,
+                       I32Attr:$k,
+                       DefaultValuedAttr<I32Attr, "-1">:$dim,
+                       DefaultValuedAttr<BoolAttr, "true">:$largest,
+                       DefaultValuedAttr<BoolAttr, "true">:$sorted);
```
azecevic000: `dim` and `largest` have reasonable defaults, but I'm more inclined to default `sorted` to false. I would assume that it incurs a performance penalty, so unless user explicitly expects values sorted we shouldn't assume them.

https://github.com/tenstorrent/tt-metal/blob/main/ttnn/cpp/ttnn/operations/reduction/topk/topk.hpp

TTNN doesn't default any of these values, IMO you should leave these defaults in TTIR (except changing `sorted` to `false`) but have TTNN attributes as not default.

lib/Conversion/TTNNToEmitC/TTNNToEmitC.cpp
```diff
@@ -4468,6 +4468,36 @@ class PagedFillCacheOpConversionPattern
...
+        srcOp, adaptor, rewriter);
+
+    llvm::SmallVector<mlir::Attribute> args{
+        emitter.emit(srcOp.getInputTensor()),
+        emitter.emit(srcOp.getK()),
+        emitter.emit(srcOp.getDim()),
+        emitter.emit(srcOp.getLargest()),
+        emitter.emit(srcOp.getSorted()),
+        emitter.emit(std::nullopt) | emitter.emit(srcOp.getMemoryConfig()),
+
```
azecevic000: Nit: remove extra line

lib/Conversion/TTNNToEmitC/TTNNToEmitC.cpp
```diff
@@ -4468,6 +4468,36 @@ class PagedFillCacheOpConversionPattern
...
+    ttnn_to_emitc::EmitCTTNNEmitter<mlir::tt::ttnn::TopKOp> emitter(
+        srcOp, adaptor, rewriter);
+
+    llvm::SmallVector<mlir::Attribute> args{
+        emitter.emit(srcOp.getInputTensor()),
+        emitter.emit(srcOp.getK()),
+        emitter.emit(srcOp.getDim()),
+        emitter.emit(srcOp.getLargest()),
+        emitter.emit(srcOp.getSorted()),
+        emitter.emit(std::nullopt) | emitter.emit(srcOp.getMemoryConfig()),
```
azecevic000:
```suggestion
        emitter.emit(srcOp.getValues()) | emitter.emit(srcOp.getMemoryConfig()),
```

tools/builder/ttir/ttir_builder.py
jameszianxuTT: https://github.com/tenstorrent/tt-mlir/actions/runs/22119847735/job/63952381822?pr=6686#step:16:3893

Something broken was merged into main accidentally, will be fixed soon https://github.com/tenstorrent/tt-mlir/pull/7091 but requires a rebase on your end to pick up once this PR is merged, sorry about that.

lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
elhewaty: I think it will always has a value, here is mhlo definition of topk: https://github.com/tensorflow/mlir-hlo/blob/master/mhlo%2FIR%2Fhlo_ops.td#L3271

elhewaty:
@sgholamiTT , can the last commit cause these build failures?

lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
azecevic000: Great find, thank you!

lib/Conversion/TTNNToEmitC/TTNNToEmitC.cpp
azecevic000: Sorry for the confusion, the suggestion should've been
```cpp
        emitter.emit(srcOp.getMemoryConfig()) | emitter.getMemoryConfig(srcOp.getValues()),
```

This is the cause of the fail in the CI.

lib/Conversion/TTNNToEmitPy/TTNNToEmitPy.cpp
```diff
@@ -3775,6 +3775,38 @@ class AssignOpConversionPattern
...
+    ttnn_to_emitpy::EmitPyTTNNEmitter<tt::ttnn::TopKOp> emitter(
+        srcOp, adaptor, rewriter, this->isGoldenModeEnabled());
+
+    llvm::SmallVector<mlir::Attribute> args{
+        emitter.emit(srcOp.getInputTensor()),
+        emitter.emit(srcOp.getK()),
+        emitter.template emit<int32_t>(srcOp.getDim()),
+        emitter.emit(srcOp.getLargest()),
+        emitter.emit(srcOp.getSorted()),
+        emitter.emit(std::nullopt | emitter.getMemoryConfig(srcOp.getValues()),
```
azecevic000, line 3879: This should be `srcOp.getMemoryConfig() | emitter.getMemoryConfig(srcOp.getValues())`

azecevic000 (approved):
@elhewaty Dialects and flatbuffer changes LGTM!

Once again, sorry for the confusion that I caused in one of the previous comments. I've updated suggestion there, from what I saw it's the reason why CI jobs fail (might be something in addition to that, I will run the CI once again when you push changes).

lib/Conversion/TTNNToEmitPy/TTNNToEmitPy.cpp
elhewaty: Thanks for help.
I will update it now, would you you please assign another issue for me? Hope it is more involved.

elhewaty:
@azecevicTT would you please take a look at the failures?

sgholamiTT:
Hey @elhewaty 
Aleksander is a bit swamped with work so I thought I could help here.
There are two CI issues:
1.
```
/__w/tt-mlir/tt-mlir/lib/Conversion/TTNNToEmitC/TTNNToEmitC.cpp:4491:33: error: invalid operands to binary expression ('::std::optional< ::mlir::tt::ttnn::MemoryConfigAttr>' and 'mlir::Attribute')
 4491 |         srcOp.getMemoryConfig() | emitter.getMemoryConfig(srcOp.getValues()),
      |         ~~~~~~~~~~~~~~~~~~~~~~~ ^ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

It's illegal to use `|` operand on `std::optional<...>` and `mlir::Attribute`.

2.
```
[ 8/18][51.8s] /opt/ttmlir-toolchain/venv/bin/clang-tidy -export-fixes /tmp/tmpdoohetx3/tmphmdvgbd4.yaml -extra-arg-before=-DDISABLE_STATIC_ASSERT_TESTS -extra-arg-before=-D__cpp_structured_bindings=202400 -p=/__w/tt-mlir/tt-mlir/build_tidy --warnings-as-errors=* /__w/tt-mlir/tt-mlir/lib/OpModel/TTNN/TTNNOpModel.cpp
/__w/tt-mlir/tt-mlir/third_party/tt-metal/src/tt-metal/ttnn/api/ttnn/graph/graph_query_op_constraints.hpp:44:54: error: too few arguments to function call, expected 9, have 6 [clang-diagnostic-error]
   44 |     return std::forward<Op>(op)(std::get<Is>(args)...);
      |            ~~~~~~~~~~~~~~~~~~~~                      ^
/__w/tt-mlir/tt-mlir/third_party/tt-metal/src/tt-metal/ttnn/api/ttnn/graph/graph_query_op_constraints.hpp:49:12: note: in instantiation of function template specialization 'ttnn::graph::detail::invoke_op_impl<std::vector<tt::tt_metal::Tensor> (*&)(const tt::tt_metal::Tensor &, unsigned int, signed char, bool, bool, const std::optional<tt::tt_metal::MemoryConfig> &, const std::optional<tt::tt_metal::CoreRangeSet> &, const std::optional<tt::tt_metal::Tensor> &, const std::optional<std::tuple<tt::tt_metal::Tensor, tt::tt_metal::Tensor>> &), std::tuple<tt::tt_metal::Tensor, unsigned int, signed char, bool, bool, std::optional<tt::tt_metal::MemoryConfig>>, 0UL, 1UL, 2UL, 3UL, 4UL, 5UL>' requested here
   49 |     return invoke_op_impl(
      |            ^
/__w/tt-mlir/tt-mlir/third_party/tt-metal/src/tt-metal/ttnn/api/ttnn/graph/graph_query_op_constraints.hpp:142:62: note: in instantiation of function template specialization 'ttnn::graph::detail::invoke_op<std::vector<tt::tt_metal::Tensor> (*&)(const tt::tt_metal::Tensor &, unsigned int, signed char, bool, bool, const std::optional<tt::tt_metal::MemoryConfig> &, const std::optional<tt::tt_metal::CoreRangeSet> &, const std::optional<tt::tt_metal::Tensor> &, const std::optional<std::tuple<tt::tt_metal::Tensor, tt::tt_metal::Tensor>> &), std::tuple<tt::tt_metal::Tensor, unsigned int, signed char, bool, bool, std::optional<tt::tt_metal::MemoryConfig>>>' requested here
  142 |             outputs = detail::extract_output_tensors(detail::invoke_op(op, transformed_args));
      |                                                              ^
/__w/tt-mlir/tt-mlir/lib/OpModel/TTNN/TTNNOpModel.cpp:7394:27: note: in instantiation of function template specialization 'ttnn::graph::query_op_constraints<std::vector<tt::tt_metal::Tensor> (*)(const tt::tt_metal::Tensor &, unsigned int, signed char, bool, bool, const std::optional<tt::tt_metal::MemoryConfig> &, const std::optional<tt::tt_metal::CoreRangeSet> &, const std::optional<tt::tt_metal::Tensor> &, const std::optional<std::tuple<tt::tt_metal::Tensor, tt::tt_metal::Tensor>> &), const tt::tt_metal::TensorSpec &, unsigned int, signed char, const bool &, const bool &, std::optional<tt::tt_metal::MemoryConfig>>' requested here
 7394 |     return ::ttnn::graph::query_op_constraints(
      |                           ^
/__w/tt-mlir/tt-mlir/third_party/tt-metal/src/tt-metal/ttnn/api/ttnn/graph/graph_query_op_runtime.hpp:64:9: error: no matching function for call to 'invoke_op' [clang-diagnostic-error]
   64 |         detail::invoke_op(op, transformed_args);
      |         ^~~~~~~~~~~~~~~~~
/__w/tt-mlir/tt-mlir/third_party/tt-metal/src/tt-metal/ttnn/api/ttnn/graph/graph_query_op_runtime.hpp:136:25: note: in instantiation of function template specialization 'ttnn::graph::capture_op_trace<std::vector<tt::tt_metal::Tensor> (*)(const tt::tt_metal::Tensor &, unsigned int, signed char, bool, bool, const std::optional<tt::tt_metal::MemoryConfig> &, const std::optional<tt::tt_metal::CoreRangeSet> &, const std::optional<tt::tt_metal::Tensor> &, const std::optional<std::tuple<tt::tt_metal::Tensor, tt::tt_metal::Tensor>> &), const tt::tt_metal::TensorSpec &, unsigned int, signed char, const bool &, const bool &, std::optional<tt::tt_metal::MemoryConfig>>' requested here
  136 |         auto trace_id = capture_op_trace(op, device, std::forward<Args>(args)...);
      |                         ^
/__w/tt-mlir/tt-mlir/lib/OpModel/TTNN/TTNNOpModel.cpp:7424:27: note: in instantiation of function template specialization 'ttnn::graph::query_op_runtime<std::vector<tt::tt_metal::Tensor> (*)(const tt::tt_metal::Tensor &, unsigned int, signed char, bool, bool, const std::optional<tt::tt_metal::MemoryConfig> &, const std::optional<tt::tt_metal::CoreRangeSet> &, const std::optional<tt::tt_metal::Tensor> &, const std::optional<std::tuple<tt::tt_metal::Tensor, tt::tt_metal::Tensor>> &), const tt::tt_metal::TensorSpec &, unsigned int, signed char, const bool &, const bool &, std::optional<tt::tt_metal::MemoryConfig>>' requested here
 7424 |     return ::ttnn::graph::query_op_runtime(
      |                           ^
/__w/tt-mlir/tt-mlir/third_party/tt-metal/src/tt-metal/ttnn/api/ttnn/graph/graph_query_op_constraints.hpp:48:6: note: candidate template ignored: substitution failure [with Op = std::vector<tt::tt_metal::Tensor> (*&)(const tt::tt_metal::Tensor &, unsigned int, signed char, bool, bool, const std::optional<tt::tt_metal::MemoryConfig> &, const std::optional<tt::tt_metal::CoreRangeSet> &, const std::optional<tt::tt_metal::Tensor> &, const std::optional<std::tuple<tt::tt_metal::Tensor, tt::tt_metal::Tensor>> &), Tuple = tuple<typename __decay_and_strip<Tensor>::__type, typename __decay_and_strip<unsigned int>::__type, typename __decay_and_strip<signed char>::__type, typename __decay_and_strip<bool>::__type, typename __decay_and_strip<bool>::__type, typename __decay_and_strip<optional<MemoryConfig>>::__type>]
   48 | auto invoke_op(Op&& op, Tuple& args) {
      |      ^
/__w/tt-mlir/tt-mlir/third_party/tt-metal/src/tt-metal/ttnn/api/ttnn/graph/graph_query_op_runtime.hpp:69:9: error: no matching function for call to 'invoke_op' [clang-diagnostic-error]
   69 |         detail::invoke_op(op, transformed_args);
      |         ^~~~~~~~~~~~~~~~~
/__w/tt-mlir/tt-mlir/third_party/tt-metal/src/tt-metal/ttnn/api/ttnn/graph/graph_query_op_constraints.hpp:48:6: note: candidate template ignored: substitution failure [with Op = std::vector<tt::tt_metal::Tensor> (*&)(const tt::tt_metal::Tensor &, unsigned int, signed char, bool, bool, const std::optional<tt::tt_metal::MemoryConfig> &, const std::optional<tt::tt_metal::CoreRangeSet> &, const std::optional<tt::tt_metal::Tensor> &, const std::optional<std::tuple<tt::tt_metal::Tensor, tt::tt_metal::Tensor>> &), Tuple = tuple<typename __decay_and_strip<Tensor>::__type, typename __decay_and_strip<unsigned int>::__type, typename __decay_and_strip<signed char>::__type, typename __decay_and_strip<bool>::__type, typename __decay_and_strip<bool>::__type, typename __decay_and_strip<optional<MemoryConfig>>::__type>]
   48 | auto invoke_op(Op&& op, Tuple& args) {
      |      ^
```
Which is complaining that [these three variables](https://github.com/tenstorrent/tt-metal/blob/70a9be53848dc6d6c5a8eaee6b31980315df11fc/ttnn/cpp/ttnn/operations/reduction/topk/topk.hpp#L24-L26) are unassigned in the `query_op_constraints` and `query_op_runtime` calls [here](https://github.com/tenstorrent/tt-mlir/pull/6686/changes#diff-63aa5b4c29464f31b7a8e2787813778daf6433e2888d0de717277320d9e2b3b7) (You can assign `std::nullopt` to them).

After you pushed the change, please ping me so I can re-run the CI job again.

elhewaty:
Hey @sgholamiTT 
would you please run the CI? my local environment corrupted and I am working on fixing it.

elhewaty:
@jgrimTT ,
Regarding the golden, I find t easier[and maybe enough] as it is right now, so if you can approve it please do, so we can move this forward.

elhewaty:
@sdjordjevicTT ,
I think the CI will pass -I hope- so please take a final look at it. If it is ok to approve, please also mention the related parties needed to merge the PR.

sgholamiTT (approved):

sdjordjevicTT (approved):
Nice! Thanks for the contribution!

jgrimTT (approved):

elhewaty:
@sdjordjevicTT , who else do we need to approve to merge this?

pilkicTT (approved):

PR #6690, opened 2026-01-21
Add optimizer refactoring proposal to spec document

rpavlovicTT:
Merges the pass-based architecture proposal into the TTNN optimizer spec as Section 5. 

The proposal outlines replacing DFShardingPolicy, ShardSolver, and most analysis pipeline with a simpler greedy approach based on empirical analysis of 50+ models.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -15,7 +15,7 @@ LICENSE @nsmithtt
 # Docs
 *.md @nsmithtt @sdjordjevicTT
 /docs/ @nsmithtt @sdjordjevicTT
-/docs/src/specs/ttnn-optimizer.md @rpavlovicTT
+/docs/src/specs/ttnn-optimizer.md @rpavlovicTT @odjuricicTT
 
 # tt-mlir Bindings
 /include/ttmlir/Bindings/ @tapspatel
--- a/docs/src/specs/ttnn-optimizer.md
+++ b/docs/src/specs/ttnn-optimizer.md
@@ -36,6 +36,8 @@ Key limitation: Only tracks first operand; single edge failure breaks entire cha
    - 5.4 [Problems with Current Approach](#54-problems-with-current-approach)
    - 5.5 [Empirical Findings](#55-empirical-findings)
    - 5.6 [Proposed Architecture](#56-proposed-architecture)
+      - 5.6.1 [Pass 1: Layout Propagation](#561-pass-1-layout-propagation)
+      - 5.6.2 [Pass 2: L1 Spill Management](#562-pass-2-l1-spill-management)
    - 5.7 [Simplicity Benefits](#57-simplicity-benefits)
    - 5.8 [Optimization Strategies](#58-optimization-strategies)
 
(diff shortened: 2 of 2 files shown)


docs/src/specs/ttnn-optimizer.md
```diff
@@ -755,3 +764,166 @@ Chain merging is a step towards global optimization but still operates on pre-bu
...
+
+When any edge in an L1 chain fails validation, the entire chain spills to DRAM. There's no mechanism for partial success—a single incompatible operation forces all connected operations out of L1 memory.
+
+#### 5.4.3 Designed for Linear Chains
+
+The ShardSolver's bitset-based constraint propagation assumes linear chain structure. Complex graph topologies like forks, joins, and diamonds require special-case handling outside the solver, adding to the overall complexity.
+
+#### 5.4.4 Backtracking Provides Little Practical Benefit
+
+In theory, the constraint satisfaction approach allows the solver to backtrack and find valid configurations. In practice, operations typically either keep their input layout or require a reshard—there's rarely a meaningful choice to optimize over.
```
odjuricicTT: It could be added that shard solver assumes and solves for the case when ops have multiple valid outputs for a single sharded input. While in practise this is rarely the case.

docs/src/specs/ttnn-optimizer.md
```diff
@@ -755,3 +764,166 @@ Chain merging is a step towards global optimization but still operates on pre-bu
...
+
```
odjuricicTT, line 808: Another big problem to add here is the use of precomputed output layouts in solver. e.g. If op constraints api returns a valid sharded output that we did not create in legal layout analysis we discard it.

docs/src/specs/ttnn-optimizer.md
```diff
@@ -755,3 +764,166 @@ Chain merging is a step towards global optimization but still operates on pre-bu
...
+### 5.5 Empirical Findings
+
+We analyzed the compiled IR for 50+ models including Segformer, ResNet50, and 45 LLM variants (Llama, Falcon, Gemma, Phi, Qwen). Key findings:
+
+**Memory pressure is rare.** All tested models show 40-94% L1 headroom. The L1 budget of ~1364 KB per core is rarely stressed.
```
odjuricicTT: Maybe point out that this is the case because of optimizer unnecesary spills, not because the models don't require the full use of L1.

e.g. ResNet with batch 16 and sharded fork-joins would probably not leave a lot of headroom (same for yolo).

docs/src/specs/ttnn-optimizer.md
```diff
@@ -755,3 +764,166 @@ Chain merging is a step towards global optimization but still operates on pre-bu
...
 
 The L1 reservation timeline mechanism introduced here provides the foundation for global memory tracking.
+
+---
+
+## 5. Proposed Refactoring: Pass-Based Architecture
+
+### 5.1 Summary
+
+We propose removing the DFShardingPolicy, ShardSolver, and most of the TTNNOptimizer analysis pipeline, replacing them with a simpler pass-based architecture. Our empirical analysis across 50+ models shows that a greedy approach can match or exceed current performance while being significantly easier to understand and maintain.
```
odjuricicTT, line 776: :) Awesome!

docs/src/specs/ttnn-optimizer.md
```diff
@@ -755,3 +764,166 @@ Chain merging is a step towards global optimization but still operates on pre-bu
...
+We propose two independent passes with clear responsibilities:
+
+#### 5.6.1 Pass 1: L1 Sharding Propagation
+
+A simple edge-based sharding picker. For each operation in schedule order:
+
+1. Look at input edges and their layouts
+2. Enumerate valid (config, layout) pairs for the op
+3. Pick the first pair that works with the inputs
+4. If no L1 layout works, fall back to DRAM
```
odjuricicTT: Can you expand a bit more on this pass?

Does is distinguis between fork / join / single ops? Does check memory requrements or not? Does it potentially leave the graph in a state where OOM is expected and the relies on the next pass to fix this?

docs/src/specs/ttnn-optimizer.md
```diff
@@ -755,3 +764,166 @@ Chain merging is a step towards global optimization but still operates on pre-bu
...
+- Track live tensors and their sizes
+- Keep tensors in L1 as long as possible
+- When budget is exceeded, spill the tensor with longest remaining lifetime
+
+**Fork handling (borrowed from DF sharding 2.0):**
+- Allow fork tensors to stay in L1 for their full lifetime when space permits
+- Modify op configs as needed (e.g., conv2d `deallocate_activation=false` for fork inputs)
+
+**Memory hierarchy:**
+- Use L1 Interleaved when sharding is not feasible
```
odjuricicTT: Not completely clear on what the first / second pass does. Would use L1 interleaved not be part of the first pass?

docs/src/specs/ttnn-optimizer.md
```diff
@@ -755,3 +764,166 @@ Chain merging is a step towards global optimization but still operates on pre-bu
...
+- Decisions are local and easy to trace
+- Fewer special cases for graph patterns
+
+This simplicity translates to faster development, easier debugging, and more predictable behavior.
+
+### 5.8 Optimization Strategies
+
+While greedy allocation provides a functional baseline, the pass-based architecture enables more sophisticated strategies to reach parity with the current optimizer.
+
+#### 5.8.1 Beam Search for Layout Propagation
```
odjuricicTT: Same question as above, would this just be part of layout propagation faze and not validate memory requirements?

docs/src/specs/ttnn-optimizer.md
```diff
@@ -755,3 +764,166 @@ Chain merging is a step towards global optimization but still operates on pre-bu
...
+
+We propose two independent passes with clear responsibilities:
+
+#### 5.6.1 Pass 1: L1 Sharding Propagation
+
+A simple edge-based sharding picker. For each operation in schedule order:
+
+1. Look at input edges and their layouts
+2. Enumerate valid (config, layout) pairs for the op
+3. Pick the first pair that works with the inputs
```
odjuricicTT: Let's discuss offline the two possiblisites here and their tradeoffs:
1. prioritize picking the input that is the same as the output of previous op
2. prioritize reshard and picking max grid input for this op locally

.github/CODEOWNERS
```diff
@@ -15,6 +15,7 @@ LICENSE @nsmithtt
 # Docs
 *.md @nsmithtt @sdjordjevicTT
 /docs/ @nsmithtt @sdjordjevicTT
+/docs/src/specs/ttnn-optimizer.md @rpavlovicTT
```
odjuricicTT: You will not be able to check in changes to this doc with only you as the owner.

rpavlovicTT: I added you too

PR #6693, opened 2026-01-21
Changing order of typecast for sharded tensors

djstefanovicTT:
### Ticket
closes https://github.com/tenstorrent/tt-xla/issues/2885

### Problem description
Problem occurs when op validation pass inserts dtype cast to bf16 for some op -- for example, max_pool2d requires bf16 even when our model is compiled with f32 dtype.

When this typecast is done on tensor that is just spilled to dram in the previous step, these 2 transformations are folded into single ToLayoutOp. Then, in decompose layouts pass this single op is decomposed into:

1. Typecast (f32->bf16)
2. ToMemoryConfig (L1 sharded -> DRAM int.)

This ordering is not valid, because Typecast can't be done on sharded tensor unless both input and output tiles are same size.

### What's changed
Changed the order of decompositions if the tensor is sharded.

### Checklist

### DIFF
--- a/lib/Dialect/TTNN/Transforms/TTNNDecomposeLayouts.cpp
+++ b/lib/Dialect/TTNN/Transforms/TTNNDecomposeLayouts.cpp
@@ -802,15 +802,17 @@ private:
     if (output.isTilized()) {
       // If the input is sharded, typecast should happen after converting to
       // memory.
-      currentInput = input.isL1Sharded() ? this->createToMemoryConfigOpIfNeeded(
-                                               op, rewriter, currentInput, info)
-                                         : currentInput;
-      currentInput = this->createDataTypeCastingOpIfNeeded(op, rewriter,
-                                                           currentInput, info);
-      currentInput = input.isL1Sharded()
-                         ? currentInput
-                         : this->createToMemoryConfigOpIfNeeded(
-                               op, rewriter, currentInput, info);
+      if (input.isL1Sharded()) {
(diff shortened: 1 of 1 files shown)


lib/Dialect/TTNN/Transforms/TTNNDecomposeLayouts.cpp
```diff
@@ -790,10 +800,17 @@ class TTNNDecomposeLayouts
...
+                                               op, rewriter, currentInput, info)
+                                         : currentInput;
       currentInput = this->createDataTypeCastingOpIfNeeded(op, rewriter,
                                                            currentInput, info);
-      currentInput = this->createToMemoryConfigOpIfNeeded(op, rewriter,
-                                                          currentInput, info);
+      currentInput = input.isL1Sharded()
+                         ? currentInput
+                         : this->createToMemoryConfigOpIfNeeded(
+                               op, rewriter, currentInput, info);
```
svuckovicTT: This is a cool way to write a little less code, but fairly confusing. I'd suggest rewriting it in a more readable pattern.

svuckovicTT (approved):

lib/Dialect/TTNN/Transforms/TTNNDecomposeLayouts.cpp
djstefanovicTT: Updated. Please confirm if it is better now.

Issue #6694, opened 2026-01-21
[EmitPy] const-eval opens already opened device


azecevicTT:
When `target-module=true` and `forward` receives device, const eval funcs still calls `utils.DeviceGetter.get_device` which results in segfault. It should be changed to receive already initialized device if exists (relevant for `target-module=true` usecase).

PR #6708, opened 2026-01-22
[EmitPy] Fix const-eval when `target-module=true`

azecevicTT:
### Ticket
https://github.com/tenstorrent/tt-mlir/issues/6694

### Problem description
When running TTIR → EmitPy with `target-module=true`, the exported entrypoint is expected to take `(input, device)` and avoid opening a device inside the generated function body (no `DeviceGetter.get_device(...)` calls). This behavior was correct for the `forward` entrypoint, but **const-eval functions** were still materializing a device via the singleton path, which breaks the “device is passed in” contract for module export.

Additionally, const-eval functions may be emitted in two equally valid styles:
- **Singleton-device path**: const-eval function opens device in the body (via `DeviceGetter`).
- **Explicit-device-arg path**: const-eval function receives `device` as a parameter.

EmitPy utilities and the const-eval call site needed to support both styles.

### What's changed
- Added a **new option** to `convert-ttnn-to-emitpy`:
  - `target-module` (plumbed from the existing TTNN EmitPy pipeline option): this has to be done during `ttnn-to-emitpy` pass, and not during the existing `TTNNPrepareModuleForExport` because otherwise we would have to change `ttcore.load_cached` 

- In `ConvertTTNNToEmitPyPass`:
  - When `target-module=true`, const-eval functions are updated to **accept an explicit `device` argument**.
  - Any `ttnn.get_device` operations inside const-eval functions are replaced with that argument.

- In the EmitPy lowering of `ttcore.load_cached`:
  - The generated wrapper call now **passes `device` when the enclosing function has a device argument** (using type-based detection via the pass type converter).

- Updated EmitPy Python utilities (`utils.py`):
  - Const-eval wrappers now invoke const-eval functions as `constEvalFunc(inputs, device)` when a device is provided, otherwise `constEvalFunc(*inputs)`.
  - This supports both singleton-device and explicit-device-arg const-eval styles.

- Added a regression test:
  - `test/ttmlir/EmitPy/target_module_const_eval.mlir`
  - Verifies const-eval + `target-module=true` produces:
    - `forward(input, device)`
    - const-eval function signature includes `device`
    - no `utils.DeviceGetter.get_device`
    - wrapper call passes `device`

### DIFF
--- a/lib/Conversion/TTNNToEmitPy/TTNNToEmitPyPass.cpp
+++ b/lib/Conversion/TTNNToEmitPy/TTNNToEmitPyPass.cpp
@@ -135,42 +135,7 @@ struct ConvertTTNNToEmitPyPass
     // they can avoid materializing `ttnn.get_device` in the function body.
     //
     if (this->targetModule) {
-      IRRewriter rewriter(&getContext());
-      mlir::tt::ttnn::DeviceType deviceType =
-          mlir::tt::ttnn::DeviceType::get(&getContext());
-
-      module.walk([&](func::FuncOp funcOp) {
-        if (!ttmlir::utils::isConstEvalFunc(funcOp) || funcOp.isExternal()) {
-          return;
-        }
-
-        // Add device argument to the const-eval function signature.
-        //
-        auto originalFuncType = funcOp.getFunctionType();
-        SmallVector<Type> newInputTypes(originalFuncType.getInputs().begin(),
-                                        originalFuncType.getInputs().end());
-        newInputTypes.push_back(deviceType);
-        auto newFuncType = FunctionType::get(&getContext(), newInputTypes,
-                                             originalFuncType.getResults());
-
-        funcOp.setFunctionType(newFuncType);
-
-        Block &entryBlock = funcOp.getBody().front();
-        BlockArgument deviceArg =
-            entryBlock.addArgument(deviceType, funcOp.getLoc());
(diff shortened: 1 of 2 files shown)


lib/Conversion/TTNNToEmitPy/TTNNToEmitPyPass.cpp
```diff
@@ -129,6 +130,49 @@ struct ConvertTTNNToEmitPyPass
...
+        //
+        SmallVector<mlir::tt::ttnn::GetDeviceOp> getDeviceOps;
+        funcOp.walk([&](mlir::tt::ttnn::GetDeviceOp op) {
+          getDeviceOps.push_back(op);
+        });
+        for (auto op : getDeviceOps) {
+          rewriter.replaceOp(op, deviceArg);
+        }
+      });
+    }
```
svuckovicTT, line 139: I would suggest moving this into a function + opening an issue to "fix" it properly (i.e. not have this logic live here). Not sure what the best way to improve this would be, but we can be vague about it in the issue, you can assign it to me.

@amilovanovicTT has also had trouble with the `load cached` op conversion. Probably makes sense to remove the logic from the conversion and add a separate pass to adjust everything around the op.

azecevic000: Agreed, I've opened the https://github.com/tenstorrent/tt-mlir/issues/6746.

Issue #5812, opened 2025-11-12
Support ttir.prod op lowering to CPU


sdjordjevicTT:
Not all combinations of ttir.prod to CPU lowering are supported at this moment.

Repro test:
```bash
pytest -svv test/python/golden/ttir_ops/reduction/test_reduction.py::test_reduction_cpu_hoisted_ops -k "prod"
```

Issue #5811, opened 2025-11-12
Support ttir.max and ttir.sum ops lowering to CPU


sdjordjevicTT:
Not all combinations of ttir.max and ttir.sum to CPU lowering are supported at this moment.

Repro test:
```bash
pytest -svv test/python/golden/ttir_ops/reduction/test_reduction.py::test_reduction_cpu_hoisted_ops -k "max"
```
```bash
pytest -svv test/python/golden/ttir_ops/reduction/test_reduction.py::test_reduction_cpu_hoisted_ops -k "sum"
```

Issue #5810, opened 2025-11-12
Support ttir.min op lowering to CPU


sdjordjevicTT:
ttir.argmax to CPU lowering isn't supported at this moment.

Repro test:
```bash
pytest -svv test/python/golden/ttir_ops/reduction/test_reduction.py::test_reduction_cpu_hoisted_ops -k "min"
```

PR #6711, opened 2026-01-22
Add TTIR -> Linalg patterns (part 2)

dmilinkovicTT:
### Ticket
#6009 #3232 #5810 #5811 #5812

### Problem description
As CPU-hoisted const-eval is being tested on models in `tt-xla`, new ops which require linalg lowering are emerging.

### What's changed

#### TTIRToTTIR decompositions
 - Added decomposition for Embedding -> Gather to reuse the TTIR -> Linalg pattern for Gather op

#### Added patterns
  - LogicalAndOp, LogicalOrOp, LogicalXorOp – Logical binary operations that convert float inputs to boolean (non-zero = true), apply the TOSA logical operation,  
  then convert back to float (true = 1.0, false = 0.0)                                                                                                             
  - MinOp – Reduction operation using tosa::ReduceMinOp chain, supports dim_arg and keep_dim attributes                                                            
  - ProdOp – Product reduction using tosa::ReduceProductOp chain, supports dim_arg and keep_dim attributes                                                         
  - MinimumOp, MaximumOp – Elementwise binary min/max operations via tosa::MinimumOp / tosa::MaximumOp                                                             
  - Relu6Op – Converts to tosa::ClampOp with min=0 and max=6                                                                                                       
  - ClampScalarOp – Converts to tosa::ClampOp with scalar min/max attributes                                                                                       
  - LinearOp – Full linear layer support via tosa::MatMulOp with:                                                                                                  
    - Optional transpose of A and B operands                                                                                                                       
    - 2D→3D and >3D→3D reshaping for TOSA matmul compatibility                                                                                                     
    - Batch dimension broadcasting via tosa::TileOp                                                                                                                
    - Optional bias addition                                                                                                                                       
  - LayerNormOp – Decomposed into elementary operations: mean computation, centering, variance, inverse standard deviation (rsqrt), normalization, and optional weight/bias application                                                                                                                                          
                                                                                                                                                                   
####  Modified Patterns                                                                                                                                                
                                                                                                                                                                   
  - SoftmaxOp – Rewritten to decompose into elementary TOSA ops (ReduceMax, Sub, Exp, ReduceSum, Reciprocal, Mul) instead of using linalg::SoftmaxOp which doesn't support linalg-to-loops lowering                                                                                                                                 
  - ConstantOp – Fixed to handle cases where types already match after HoistCPUOps conversion                                                                      
  - getDimsFromAttribute – Fixed to normalize negative dimension indices     

### Checklist

### DIFF
--- a/lib/Conversion/TTIRToTTIRDecomposition/TTIRToTTIRDecomposition.cpp
+++ b/lib/Conversion/TTIRToTTIRDecomposition/TTIRToTTIRDecomposition.cpp
@@ -679,13 +679,13 @@ private:
 } // namespace
 
 //===----------------------------------------------------------------------===//
-// Embedding to Gather Pattern (CPUFallback)
+// Embedding to Gather Pattern
 //===----------------------------------------------------------------------===//
 
 namespace {
 // Converts ttir.embedding to ttir.gather for CPU fallback path.
-// This avoids using tensor.gather which lacks bufferization support in upstream
-// MLIR.
+// This allows us not to use tensor.gather which lacks bufferization support in
+// upstream MLIR.
 struct EmbeddingToGatherConversionPattern
     : public OpConversionPattern<ttir::EmbeddingOp> {
   using OpConversionPattern<ttir::EmbeddingOp>::OpConversionPattern;
@@ -713,7 +713,12 @@ struct EmbeddingToGatherConversionPattern
     auto indicesShape = inputType.getShape();
     int64_t indicesRank = indicesShape.size();
     auto weightShape = weightType.getShape();
-    int64_t embeddingDim = weightShape[1];
+    int64_t weightRank = weightType.getRank();
+
+    // Weight is "effectively 2D" with shape (1, 1, ..., 1, vocab_size,
+    // embedding_dim). The embedding dimension is always the last dimension.
+    int64_t embeddingDim = weightShape[weightRank - 1];
+    int64_t vocabDimIndex = weightRank - 2;
 
     // Add trailing dimension of size 1 to indices for index_vector_dim.
     SmallVector<int64_t> newIndicesShape(indicesShape.begin(),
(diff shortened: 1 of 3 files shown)


lib/Conversion/TTIRToLinalg/TTIRToLinalg.cpp
```diff
@@ -1563,6 +1595,84 @@ class LogicalNotOpConversionPattern
...
+  matchAndRewrite(TTIROpTy op, typename TTIROpTy::Adaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    Value lhs = adaptor.getLhs();
+    Value rhs = adaptor.getRhs();
+
+    auto resultType = dyn_cast<RankedTensorType>(
+        this->getTypeConverter()->convertType(op.getResult().getType()));
+    assert(resultType && "Result type must be a ranked tensor type.");
+
+    // Convert both inputs to boolean tensors
```
vwellsTT: nit: . these comments

lib/Conversion/TTIRToLinalg/TTIRToLinalg.cpp
```diff
@@ -1925,28 +2065,77 @@ class ElementwiseOpConversionPattern : public OpConversionPattern<TTIROpTy> {
...
 
-    Value input = adaptor.getInput();
-    const size_t inputSize =
-        dyn_cast<RankedTensorType>(input.getType()).getShape().size();
-    const int32_t dimension = (op.getDimension() < 0)
-                                  ? op.getDimension() + inputSize
-                                  : op.getDimension();
+    // Normalize dimension to be positive.
+    int32_t dim = op.getDimension();
+    if (dim < 0) {
```
vwellsTT: maybe we should have a tiny helper to normalize dims--seems like this pattern gets repeated a few times now

lib/Conversion/TTIRToLinalg/TTIRToLinalg.cpp
```diff
@@ -2089,6 +2313,9 @@ class SliceStaticOpConversionPattern
 } // namespace
 
 namespace {
+// Converts ttir.embedding to ttir.gather, which then gets lowered to
+// linalg.generic by GatherOpConversionPattern. This avoids using tensor.gather
+// which lacks bufferization support in upstream MLIR.
```
vwellsTT: I would somewhat prefer if this sort of TTIR -> TTIR decomp happens in TTIRToTTIRDecomp code--we already have specialized target for cpu hoisting there I think, so we should be able to do ttir.embedding to ttir.gather easily

Issue #6604, opened 2026-01-14
[d2m] ttir.reshape op coverage


nsmithtt:
Generate all usages of this op and add them as unit tests by following this guide here: 
- #6602

nsmithtt:
Currently failing 19 / 150 due to noc alignment issues

nsmithtt:
The remaining failures are due to noc alignment or allocator which are more fundamental and outside the scope of the reshape op.  We can close for now and move tracking of remaining work to more specific issues.

PR #6714, opened 2026-01-22
[d2m] pre processing to remove ttir.reshapes and ttir.broadcasts when dealing with constant tensors

dlokeTT:
### Problem description
Unneccessary `ttir.reshape` and `ttir.broadcasts` that would change shape of tensors to match constant tensors. Helps in clarifying what reshapes we actually need to support in d2m as part of #6604 

### What's changed
Added two ttir passes, `--ttir-move-reshape-to-constant` `--ttir-fold-constant-reshape-broadcast` that first move the reshape from the real tensor to const tensor then remove any reshapes or broadcasts on the const tensor by creating a new const tensor of this shape. These passes are meant to be used in d2m pipeline since we have handling for const tensors in `scalarizeConstTensors`
Added lit tests

### Checklist

### DIFF
--- a/lib/Dialect/TTIR/Transforms/FoldConstantReshapeBroadcast.cpp
+++ b/lib/Dialect/TTIR/Transforms/FoldConstantReshapeBroadcast.cpp
@@ -1,4 +1,4 @@
-// SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
+// SPDX-FileCopyrightText: (c) 2026 Tenstorrent AI ULC
 //
 // SPDX-License-Identifier: Apache-2.0
 
@@ -38,7 +38,7 @@ static DenseElementsAttr findSplatConstantThroughChain(Value value) {
     return nullptr;
   }
 
-  // Direct constant case
+  // Direct constant case.
   if (auto constantOp = dyn_cast<ConstantOp>(defOp)) {
     auto denseAttr = dyn_cast<DenseElementsAttr>(constantOp.getValue());
     if (denseAttr && denseAttr.isSplat()) {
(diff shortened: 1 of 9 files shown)


lib/Dialect/TTIR/Transforms/FoldConstantReshapeBroadcast.cpp
```diff
@@ -0,0 +1,189 @@
+// SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
```
vwellsTT: nit: year

lib/Dialect/TTIR/Transforms/FoldConstantReshapeBroadcast.cpp
```diff
@@ -0,0 +1,189 @@
...
+// Returns the splat DenseElementsAttr if found, nullptr otherwise.
+// This handles patterns like:
+//   reshape(broadcast(reshape(constant))) where constant is a splat.
+static DenseElementsAttr findSplatConstantThroughChain(Value value) {
+  Operation *defOp = value.getDefiningOp();
+  if (!defOp) {
+    return nullptr;
+  }
+
+  // Direct constant case
```
vwellsTT: nit: sentence this + all below comments pls

lib/Dialect/TTIR/Transforms/MoveReshapeToConstant.cpp
```diff
@@ -0,0 +1,170 @@
+// SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
```
vwellsTT: nit: year

lib/Dialect/TTIR/Transforms/MoveReshapeToConstant.cpp
```diff
@@ -0,0 +1,170 @@
...
+//   %reshaped = ttir.reshape(%act) : tensor<32x2560xf32> ->
+//   tensor<32x1x2560xf32> %result = ttir.pow(%reshaped, %const) :
+//   tensor<32x1x2560xf32>
+//
+// Transforms to:
+//   %const = ttir.constant() : tensor<32x1x2560xf32>
+//   %const_reshaped = ttir.reshape(%const) : tensor<32x1x2560xf32> ->
+//   tensor<32x2560xf32> %act = ... : tensor<32x2560xf32> %result =
+//   ttir.pow(%act, %const_reshaped) : tensor<32x2560xf32>
+class MoveReshapeToConstantPattern
```
vwellsTT, line 54: I wonder if there are any edge-cases w/ implicit bcast on the input that could cause problems--does the other pass ensure that type of pattern gets prevented?

dlokeTT: Talked offline but yeah this is something we would have to worry about if implicit broadcasts were part of the pipeline but they aren't in d2m yet. If/when we add that pass, it should be run after these passes

PR #6715, opened 2026-01-22
cursor

dlokeTT:
Test to see if this simple thing works. Could be used as a replacement for fixing `print_ir` flag on default pipelines


dlokeTT:
Doesn't work because of some issue with jit finding builder module. Might look into this later

Issue #6671, opened 2026-01-19
Fold `ttir.repeat` when `repeat_dimensions` are 1


azecevicTT:
The op is neutral (no-op) in this case.

PR #6720, opened 2026-01-23
Repeat identity folding

djstefanovicTT:
### Ticket
closes https://github.com/tenstorrent/tt-mlir/issues/6671

### Problem description
Repeat was not folded when all repeating dimensions were 1.

### What's changed
Added check that applies said fold. Removed now redundant check if repeat is done only once.

### Checklist

### DIFF
--- a/test/ttmlir/Conversion/TTIRToLinalg/repeat.mlir
+++ b/test/ttmlir/Conversion/TTIRToLinalg/repeat.mlir
@@ -50,14 +50,4 @@ module {
     %0 = "ttir.repeat"(%arg0) <{repeat_dimensions = array<i64: 2, 1>}> : (tensor<2x3xbf16>) -> tensor<4x3xbf16>
     return %0 : tensor<4x3xbf16>
   }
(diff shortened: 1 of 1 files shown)


test/ttmlir/Conversion/TTIRToLinalg/repeat.mlir
```diff
@@ -54,8 +54,8 @@ module {
   // Test repeat with no repetition (1x factor)
   func.func @repeat_identity(%arg0: tensor<2x3xf32>) -> tensor<2x3xf32> {
     // CHECK-LABEL: func.func @repeat_identity
-    // CHECK: tosa.const_shape
-    // CHECK: tosa.tile
```
azecevic000, line 58: IMO it makes more sense to remove this function from this test because it doesn't test anything TTIR->Linalg related.

djstefanovicTT: Updated.

PR #6725, opened 2026-01-23
[TTIRFusing] Fix matching in RMSNormFusionPattern

mvasiljevicTT:
### Problem description
Current implementation of `RMSNormFusionPattern` doesn't match all rms norms in some qwen models.

### What's changed
To match pattern for `RMSNorm` we looked through all layout ops. In some qwen models it tried to match one too many reshapes and break matching because of input and output shape mismatch.

This is why we define `lookThroughLayoutOpsIf` and `preservesDim` predicate that is used with it in this pattern.

It is also used to refactor `RMSNormCommutePatterns` and should be used to simplify other `EraseInverseOp` patterns.

### Checklist
- [x] [perf benchmark for qwen](https://github.com/tenstorrent/tt-xla/actions/runs/21283054357)  showing big improvements on some models
- [ ] [perf benchmark](https://github.com/tenstorrent/tt-xla/actions/runs/21292407473)

### DIFF
--- a/lib/Dialect/TTIR/Utils/Utils.cpp
+++ b/lib/Dialect/TTIR/Utils/Utils.cpp
@@ -116,16 +116,6 @@ bool preservesDim(mlir::Operation *op, int64_t dim) {
   }
 
   return llvm::TypeSwitch<mlir::Operation *, bool>(op)
-      .Case<PermuteOp>([&](PermuteOp permute) {
-        auto perm = permute.getPermutation();
-        // Dimension must stay in the same position.
-        if (perm[dim] != dim) {
-          return false;
-        }
-        // All dimensions after must still map to positions after.
-        auto permAfter = perm.drop_front(dim + 1);
-        return llvm::all_of(permAfter, [&](int64_t p) { return p > dim; });
-      })
       .Case<RepeatInterleaveOp>([&](RepeatInterleaveOp repeat) {
         int64_t repeatDim = repeat.getDim();
         if (repeatDim < 0) {
--- a/include/ttmlir/Dialect/TTIR/Utils/Utils.h
+++ b/include/ttmlir/Dialect/TTIR/Utils/Utils.h
@@ -326,27 +326,6 @@ mlir::Value lookThrough(mlir::Value value) {
   return value;
 }
 
-// Traces backward through specified ops that satisfy the predicate.
-template <typename... Ops, typename PredicateFn>
-mlir::Value lookThroughIf(mlir::Value value, PredicateFn &&predicate) {
-  while (auto *op = value.getDefiningOp()) {
-    if (llvm::isa<Ops...>(op) && predicate(op)) {
-      value = op->getOperand(0);
-    } else {
-      break;
-    }
-  }
-  return value;
-}
-
-// Traces backward through specified ops that satisfy the predicate to find
-// an operation of type OpTy.
-template <typename OpTy, typename... Ops, typename PredicateFn>
-OpTy findOpThroughIf(mlir::Value value, PredicateFn &&predicate) {
-  return lookThroughIf<Ops...>(value, std::forward<PredicateFn>(predicate))
-      .template getDefiningOp<OpTy>();
-}
-
 // Traces backward from a value through specified ops to find an operation of
 // type OpTy.
 template <typename OpTy, typename... Ops>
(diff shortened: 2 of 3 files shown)


lib/Dialect/TTIR/Utils/Utils.cpp
```diff
@@ -62,4 +66,108 @@ mlir::LogicalResult broadcastValue(mlir::PatternRewriter &rewriter,
...
+        int64_t repeatDim = repeat.getDim();
+        if (repeatDim < 0) {
+          repeatDim += inputRank;
+        }
+        return repeatDim != dim;
+      })
+      .Case<ReshapeOp>([&](ReshapeOp reshapeOp) {
+        // Convert LTR dim to RTL position.
+        int64_t inputDimRTL = inputRank - 1 - dim;
+        int64_t foundRTL = findMatchingDimRTL(reshapeOp, inputDimRTL);
```
vmilosevic, line 129: :warning: **clang\-diagnostic\-error** :warning:
use of undeclared identifier `` findMatchingDimRTL ``

lib/Dialect/TTIR/Utils/Utils.cpp
```diff
@@ -4,6 +4,10 @@
 
 #include "ttmlir/Dialect/TTIR/Utils/Utils.h"
 
+#include "llvm/ADT/TypeSwitch.h"
+
+#include <numeric>
+
```
jserbedzijaTT: Are these needed

mvasiljevicTT: `TypeSwitch` is, but looks like`numeric` is already included, thanks

mvasiljevicTT:
> I think we should start using canonicalization/folding patterns soon for TMs instead of writing utility functions to skip them, because things will get complicated quickly and we’ll probably miss some potential matches.

That would be nice, but most TM combinations are not foldable (perm+rshp), so as long as they are moving freely in EIO we could expect them anywhere. :/

lib/Dialect/TTIR/Utils/Utils.cpp
```diff
@@ -62,4 +64,108 @@ mlir::LogicalResult broadcastValue(mlir::PatternRewriter &rewriter,
...
+  auto outputShape = outputType.getShape();
+  int64_t inputRank = inputType.getRank();
+  int64_t outputRank = outputType.getRank();
+
+  // Normalize negative dimension.
+  if (dim < 0) {
+    dim += inputRank;
+  }
+
+  return llvm::TypeSwitch<mlir::Operation *, bool>(op)
```
mtopalovicTT, line 118: Is it safe to look through permute?

test/ttmlir/Dialect/TTIR/fusing/rms_norm_fusion.mlir
```diff
@@ -207,4 +207,34 @@ module {
...
+        // CHECK: "ttir.reshape"(%arg0)
+        // CHECK-SAME: (tensor<32x4096xbf16>) -> tensor<32x1x32x128xbf16>
+        // CHECK: "ttir.rms_norm"
+        // CHECK-SAME: (tensor<32x1x32x128xbf16>, tensor<128xbf16>) -> tensor<32x1x32x128xbf16>
+        %0 = "ttir.constant"() <{value = dense<9.99999997E-7> : tensor<32x1x32x1xf32>}> : () -> tensor<32x1x32x1xf32>
+        %1 = "ttir.constant"() <{value = dense<2.000000e+00> : tensor<32x1x32x128xf32>}> : () -> tensor<32x1x32x128xf32>
+        %2 = "ttir.reshape"(%arg1) <{shape = [1 : i32, 1 : i32, 1 : i32, 128 : i32]}> : (tensor<128xbf16>) -> tensor<1x1x1x128xbf16>
+        %3 = "ttir.broadcast"(%2) <{broadcast_dimensions = array<i64: 32, 1, 32, 1>}> : (tensor<1x1x1x128xbf16>) -> tensor<32x1x32x128xbf16>
+        // This reshape changes the last dim from 4096 to 128 - fusion should NOT look through it
+        %4 = "ttir.reshape"(%arg0) <{shape = [32 : i32, 1 : i32, 32 : i32, 128 : i32]}> : (tensor<32x4096xbf16>) -> tensor<32x1x32x128xbf16>
```
mtopalovicTT, line 227: can we add some different test examples in builder with different TMs through which we look through to validate that output is correct?

lib/Dialect/TTIR/Utils/Utils.cpp
mvasiljevicTT: On second thought, no it isn't, and it shouldn't be part of 'layout ops'. 

Changed the logic to look only through ops that could be reverted by setting shape and type to the expected shape and type.

test/ttmlir/Dialect/TTIR/fusing/rms_norm_fusion.mlir
mvasiljevicTT: There were no builder tests for rms norm fusing, added them now :D

mvasiljevicTT:
More rms norm matching in llama 8b model causes this error
```
TT_FATAL: Out of Memory: Not enough space to allocate 117440512 B DRAM buffer across 12 banks, where each bank needs to store 9787392 B, but bank size is 1073741792 B (allocated: 976657856 B, free: 97083936 B, largest free block: 4587840 B) 
```
Pausing merging this PR until we solve that, as it would make regression on perf benchmark now.

mvasiljevicTT:
Perf benchmark now passing after changes on tt-forge : https://github.com/tenstorrent/tt-xla/actions/runs/21859900359

Issue #6042, opened 2025-11-26
EmitPy - ConstEval, Improvements


vcanicTT:
Currently, when we generate the Python code for some model or test graph, the whole code related to the model is placed in the same, single file. 

Here, I would like to present how we can structure the generated code, starting from const eval part of the code.

Today, const eval:
- is in the same file as the model,
- does caching using global variables,
- is implemented using wrapper functions for fetching the result.

That is causing huge number of lines of the code, and very hard searching and changing.

I propose:
- to make additional folder as part of the bigger model folder, and put there files for const eval functions, each function in the separate file, eg. model/const_eval/main_const_eval_22.py;
- to remove all global variables from the code, they are generally bad solution, and we should avoid using them. We should generate more Pythonic code. Instead, we should have hash map/structure/lookup table where we can store all intermediate values from const eval graphs, that can be part of some general cache.py file, where we can put all structures of this kind; 
- to remove wrappers from the code;
- to remove and reduced all unnecessary code related to const eval. Here, I mean when we store and load the results. LOAD and STORE can be part of the cache structure. 

The point of this issue is to launch general idea how to produce well-structured TTNN code. If we look at models in TT-Metal, we can see that they are written in many files using different features, like pipelines, trace, etc. 

Let's start thinking in that way, how to organize models in more than one file. Feature const eval is great starting point. If we do these few changes the main file can be reduced from ~20000 lines to less than 6000.

Related issues:
https://github.com/tenstorrent/tt-mlir/issues/6043
https://github.com/tenstorrent/tt-mlir/issues/6044

@svuckovicTT @sdjordjevicTT Can you, please, take a look? Also, if you think, we can make some meeting, and doc, aiming to sketch up main components of the code.

PR #6729, opened 2026-01-23
[EmitPy] Implement fileOp and decouple consteval logic into separate file

amilovanovicTT:
### Ticket
#6042

### Problem Description
Previously, the Python code generation process emitted all consteval logic directly into the model file. This resulted in bloated, monolithic files that were difficult to navigate and maintain. To improve code readability and modularity, this PR introduces a mechanism to decouple constant evaluation from the primary execution logic.

The generated output is now split into two distinct files:
* **Consteval file**: Encapsulates all constant evaluation logic and provides a clean API for the main file, hiding implementation details.
* **Main file**: Contains the core model logic and invokes the consteval API.

### Changes
* Introduced `fileOp` to the `EmitPy` dialect. Each `fileOp` region now represents the complete content of a single output file.
* Added the following passes in `TTNNToEmitPyDevicePipeline`:
    - `TTNNPrepareConstEvalCaching`:
      - Prepares the const-eval results for caching by packing the results of all LoadCachedOps into a single global caching dictionary per forward function
    - `TTNNFileSplit`:
      - Organizes TTNN IR into FileOp containers (`emitpy::FileOp` or `emitc::FileOp`, selected via the `target` pass option)
    - `EmitPyConstEvalCachingPass`: 
      - Post-conversion pass that wraps const-eval function calls and their dict caching ops in an `if not dict:` guard.
 * Updated/extended the following passes: 
    - `TTNNToEmitPy` : 
      - LoadCachedOp conversion pattern now converts to the plain function call, since caching infrastructure is now handled by other passes
      - Added conversion for the new ttcore dictionary ops
    - `EmitPyLinkModules` : 
      - If TTNNFileSplit is performed, CPU definitions are placed in the consteval file, and declarations from the main file are used to create a proper import statement

* **Tooling Updates**:
    * **tt-alchemist**: Now automatically emits both files into the specified solution folder. To disable TTNNFileSplit pass use `--pipeline-options "split-files=true"`.
    * **ttmlir-translate**: Added a `-emitpy-file-id` flag for targeted emission (`-emitpy-file-id="main"` or `-emitpy-file-id="consteval"`). If no ID is specified, the tool generates a single stream with clear comment headers labeling each file segment.

### Future improvements
As a follow-up, I will add another pass at the end of `TTNNToEmitPyDevicePipeline` to pack the ops into the `emitpy::ExpressionOp`, improving code readability.

### Checklist

### DIFF
--- a/lib/Dialect/TTNN/Pipelines/TTNNPipelines.cpp
+++ b/lib/Dialect/TTNN/Pipelines/TTNNPipelines.cpp
@@ -54,6 +54,10 @@ void createTTNNPipelineTTIRPasses(
     pm.addPass(mlir::tt::ttir::createTTIRQuantDequantConversion());
   }
   pm.addPass(mlir::tt::createTTIRToTTIRDecompositionPass());
+  pm.addPass(mlir::createCanonicalizerPass());
+  if (options.implicitBroadcastFoldingEnabled) {
+    pm.addPass(mlir::tt::ttir::createTTIRImplicitBroadcastFold());
+  }
   if (options.enableFusing) {
     pm.addPass(mlir::tt::ttir::createTTIRFusing(fusingOptions));
   }
--- a/.github/workflows/call-build-release.yml
+++ b/.github/workflows/call-build-release.yml
@@ -121,6 +121,7 @@ jobs:
           -DTTMLIR_ENABLE_STABLEHLO=ON \
           -DTTMLIR_ENABLE_OPMODEL=${{ matrix.build.enable_op_model }} \
           -DTTMLIR_ENABLE_ALCHEMIST_WHEEL=${{ matrix.build.enable_alchemist_whl }} \
+          -DTT_USE_SYSTEM_SFPI=ON \
           -DTT_RUNTIME_DEBUG=${{ matrix.build.enable_runtime_debug }} \
           -S ${{ steps.strings.outputs.work-dir }}
 
--- a/.github/workflows/call-lint.yml
+++ b/.github/workflows/call-lint.yml
@@ -54,6 +54,7 @@ jobs:
         -DTTMLIR_ENABLE_RUNTIME_TESTS=ON \
         -DTTMLIR_ENABLE_STABLEHLO=ON \
         -DTTMLIR_ENABLE_OPMODEL=ON \
+        -DTT_USE_SYSTEM_SFPI=ON \
         -S ${{ steps.strings.outputs.work-dir }}
 
     - name: Lint
--- a/.github/workflows/call-test-ttsim.yml
+++ b/.github/workflows/call-test-ttsim.yml
@@ -147,6 +147,8 @@ jobs:
           export TT_METAL_HOME="$WORK_DIR/third_party/tt-metal/src/tt-metal"
           export LD_LIBRARY_PATH="$INSTALL_DIR/lib:${TTMLIR_TOOLCHAIN_DIR}/lib:${LD_LIBRARY_PATH}"
           export SYSTEM_DESC_PATH="${GITHUB_WORKSPACE}/ttrt-artifacts/system_desc.ttsys"
+          export TT_METAL_INSPECTOR="0"
+          export TT_METAL_INSPECTOR_RPC="0"
           mkdir -p test_reports
 
           run_sim_test() {
--- a/.github/workflows/call-test.yml
+++ b/.github/workflows/call-test.yml
@@ -157,6 +157,8 @@ jobs:
         TTRT_LOGGER_LEVEL: "WARNING"
         LOGURU_LEVEL: "ERROR"
         TT_METAL_LOGGER_LEVEL: "FATAL"
+        TT_METAL_INSPECTOR: "0"
+        TT_METAL_INSPECTOR_RPC: "0"
       run: |
         # Run Tests
         source env/activate
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -45,7 +45,7 @@ option(TTMLIR_ENABLE_TOOLS "Enable building core tools (ttmlir-opt, ttmlir-lsp-s
 option(TTMLIR_ENABLE_ALCHEMIST "Enable tt-alchemist lib" ON)
 option(TTMLIR_ENABLE_ALCHEMIST_WHEEL "Enable tt-alchemist wheel" OFF)  # Expensive to build
 option(CODE_COVERAGE "Enable coverage reporting" OFF)
-option(TT_USE_SYSTEM_SFPI "Use system path for SFPI. SFPI is used to compile firmware." ON)
+option(TT_USE_SYSTEM_SFPI "Use system path for SFPI. SFPI is used to compile firmware." OFF)
 
 if (TTMLIR_ENABLE_RUNTIME)
   option(TT_RUNTIME_ENABLE_TTNN "Enable TTNN Runtime (requires TTMLIR_ENABLE_RUNTIME)" ON)
--- a/include/Constants.h
+++ b/include/Constants.h
@@ -13,6 +13,14 @@ namespace tt::constants {
 // This reserves a region of L1 memory for L1_SMALL buffers used by convs.
 constexpr static std::size_t L1_SMALL_SIZE = 1 << 16;
 
+// Used only in unittests:
+// todo(arminaleTT): look into dynamically adjusting this
+// getOpRuntime() uses trace capture to run and measure the runtime of an op.
+// This requires the device to be opened with sufficient trace region size.
+// This number is currently set based on manual testing of supported ops to
+// accommodate the highest required trace buffer size (2004992B)
+static constexpr size_t opModelDefaultTraceRegionSize = 6000000;
+
 } // namespace tt::constants
 
 #endif // TTMLIR_DIALECT_TTNN_UTILS_CONSTANTS_H
--- a/include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
+++ b/include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
@@ -1164,8 +1164,8 @@ def D2M_RemoteLoadOp : D2M_GenericRegionDatamovementOp<"remote_load",
         -> memref<SY x SX x f32, ...>
       ```
 
-      The explicit CB form takes a CB block arg as additional input; the load is
-      produced into the CB. The CB provided must be the CB block arg associated
+      The explicit CB form takes a CB (from `d2m.get_cb`) as additional input;
+      the load is produced into the CB. The CB provided must be the CB associated
       with the operand. In explicit CB form, no local buffer is required.
 
       ```mlir
@@ -1332,6 +1332,15 @@ def D2M_RemoteLoadOp : D2M_GenericRegionDatamovementOp<"remote_load",
         return mlir::cast<CBType>(cb.getType());
       }
 
+      /// Return the CB port number from this op's CB operand.
+      /// The op must be in explicit CB form with a GetCBOp-defined CB.
+      unsigned getCBPort() {
+        assert(isExplicitCBForm() && "getCBPort requires explicit CB form");
+        auto getCBOp = getCb().getDefiningOp<GetCBOp>();
+        assert(getCBOp && "CB operand must be defined by a GetCBOp");
+        return getCBOp.getPort();
+      }
+
       /// Returns true if the circular buffer operand is present (explicit CB form).
       bool isExplicitCBForm() { return static_cast<bool>(getCb()); }
 
@@ -1426,7 +1435,7 @@ def D2M_RemoteStoreOp : D2M_GenericRegionDatamovementOp<"remote_store",
         -> memref<GY x GX x SY x SX x f32, ...>
       ```
 
-      - Explicit CB form stores data from a concrete CB block arg to the remote operand.
+      - Explicit CB form stores data from a concrete CB (from `d2m.get_cb`) to the remote operand.
 
       ```mlir
       d2m.remote_store %generic_operand[%i, %j] from %cb
@@ -1505,6 +1514,16 @@ def D2M_RemoteStoreOp : D2M_GenericRegionDatamovementOp<"remote_store",
         Value cb = getCb();
         if (!cb) return CBType();
         return mlir::cast<CBType>(cb.getType());
+      }
+
+      /// Return the CB port number from this op's CB operand, or UINT_MAX if
+      /// Return the CB port number from this op's CB operand.
+      /// The op must be in explicit CB form with a GetCBOp-defined CB.
+      unsigned getCBPort() {
+        assert(isExplicitCBForm() && "getCBPort requires explicit CB form");
+        auto getCBOp = getCb().getDefiningOp<GetCBOp>();
+        assert(getCBOp && "CB operand must be defined by a GetCBOp");
+        return getCBOp.getPort();
       }
 
       /// Returns true if the circular buffer operand is present (explicit CB form).
@@ -1964,9 +1983,10 @@ def D2M_GetScratchFromCBOp : D2M_GenericRegionOp<"get_scratch_from_cb",
   let summary = "Unwrap scratch circular buffer into a memref.";
   let description = [{
     The `get_scratch_from_cb` operation extracts the underlying memref from
-    a scratch circular buffer (CB) block argument. This op is inserted by the
-    LowerScratchAllocate pass at the beginning of the `d2m.generic` region to
-    make the scratch buffer available as a plain memref for use by downstream
+    a scratch circular buffer (CB) obtained via `d2m.get_cb`. This op is
+    inserted by the LowerScratchAllocate pass at the beginning of the
+    `d2m.generic` region to make the scratch buffer available as a plain memref
+    for use by downstream
     scratch allocation ops (e.g. `d2m.scratch_allocate` lowered to
     `memref.subview`).
 
@@ -2184,6 +2204,29 @@ def D2M_CoreIndexOp : D2M_GenericRegionOp<"core_index",
 //===----------------------------------------------------------------------===//
 // D2M Remote Access ops (Used in TTMetal Lowering)
 //===----------------------------------------------------------------------===//
+
+def D2M_GetCBOp : D2M_GenericRegionOp<"get_cb", [Pure]> {
+    let summary = "Get a circular buffer for the given port.";
+    let description = [{
+      Obtain a CB handle for the specified port number.  Each unique port maps
+      to exactly one physical circular buffer.
+
+      The port number is a local identifier — it does not correspond to the
+      parent generic op's operand index.  The binding between a CB port and a
+      generic operand is established by `remote_load`/`remote_store` ops.
+
+      ```mlir
+      %cb0 = d2m.get_cb(0) : !d2m.cb<memref<2x2x!ttcore.tile<32x32, f32>, #l1_>>
+      %buf = d2m.wait %cb0 : !d2m.cb<memref<...>> -> memref<...>
+      ```
+    }];
+
+    let arguments = (ins ConfinedAttr<I64Attr, [IntMinValue<0>]>:$port,
+                         OptionalAttr<I64Attr>:$operand_index);
+    let results = (outs D2M_CB:$result);
+
+    let assemblyFormat = [{ `(` $port `)` (`operand_index` `=` $operand_index^)? attr-dict `:` type($result) }];
+}
 
 def D2M_GetGlobalOperandOp : D2M_GenericRegionOp<"get_global_operand", [Pure]> {
     let summary = "Get global operand op.";
--- a/include/ttmlir/Dialect/D2M/IR/D2MOps.td
+++ b/include/ttmlir/Dialect/D2M/IR/D2MOps.td
@@ -373,6 +373,12 @@ def D2M_GenericOp
       static Value findAssocOperand(::mlir::tensor::EmptyOp emptyOp);
       static Value findAssocCBByOperandIndex(Operation *op, unsigned operandIndex);
       static Value findAssocCBByOperand(Operation *op, Value operand);
+
+      /// Get the allocation (tensor.empty or memref.alloc) op result at
+      /// position `operandIndex` from the region's first block using
+      /// positional counting.  Returns null if not found.
+      static Value getOperandAlloc(Region &region, unsigned operandIndex);
+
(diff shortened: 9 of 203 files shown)


jgrimTT:
Could you add an emitpy consteval test in `test/python/golden`? I tried to run `pytest test/python/golden/test_ttir_ops.py::test_conv2d_consteval` with an `target="emitpy"`, but it crashed

lib/Conversion/TTNNToEmitPy/TTNNToEmitPy.cpp
```diff
@@ -2518,26 +2520,51 @@ class LoadCachedOpConversionPattern
...
-    //
+    // beginning of the function body. Find or create a constant op representing
+    // the outer dictionary key for the parent function.
     emitpy::GlobalStatementOp global = nullptr;
+    emitpy::ConstantOp funcNameKey = nullptr;
+    auto parentOp = loadCachedOp->getParentOfType<func::FuncOp>();
+    if (!parentOp) {
+      return failure();
+    }
+    Block &entryBlock = parentOp.getBody().front();
```
vmilosevic: :warning: **clang\-diagnostic\-error** :warning:
redefinition of `` entryBlock ``

amilovanovicTT:
> Could you add an emitpy consteval test in `test/python/golden`? I tried to run `pytest test/python/golden/test_ttir_ops.py::test_conv2d_consteval` with an `target="emitpy"`, but it crashed

Hi @jgrimTT, there was a bug in this PR such that only one main module was created from the generated Python code in the builder. This is now fixed with two modules being created (one for the main file and one for the consteval file), so this test now passes.

vvukomanTT (approved):

jgrimTT:
> > Could you add an emitpy consteval test in `test/python/golden`? I tried to run `pytest test/python/golden/test_ttir_ops.py::test_conv2d_consteval` with an `target="emitpy"`, but it crashed
> 
> Hi @jgrimTT, there was a bug in this PR such that only one main module was created from the generated Python code in the builder. This is now fixed with two modules being created (one for the main file and one for the consteval file), so this test now passes.

Awesome, thanks! Can you change the test to parameterize target? 
`@pytest.mark.parametrize("target", ["ttnn", "emitpy"])`
Or add any consteval emitpy test

include/ttmlir/Conversion/Passes.td
```diff
@@ -257,6 +257,16 @@ def EmitPyLinkModules : Pass<"emitpy-link-modules", "::mlir::ModuleOp"> {
   let dependentDialects = ["mlir::tt::emitpy::EmitPyDialect"];
 }
 
+def EmitPySplitFiles: Pass<"emitpy-split-files", "::mlir::ModuleOp"> {
```
svuckovicTT: This pass isn't tied to the dialect, but is use-case specific - I think it should be named along the lines of `CodegenSplitFiles`.

lib/Conversion/TTNNToEmitPy/EmitPySplitFiles.cpp
```diff
@@ -0,0 +1,341 @@
...
+    // Categorize top-level module operations.
+    for (auto &op : llvm::make_early_inc_range(moduleOp.getOps())) {
+      if (isa<emitpy::FileOp>(op)) {
+        continue;
+      }
+      if (isa<emitpy::ImportOp>(op)) {
+        importOps.push_back(&op);
+        continue;
+      }
+      if (op.hasAttr("emitpy.consteval")) {
```
svuckovicTT: I was going to suggest checking if this is a func op and then if it is a consteval func, but I see that block just below so I'm not sure what this block tries to catch?

lib/Conversion/TTNNToEmitPy/EmitPySplitFiles.cpp
```diff
@@ -0,0 +1,341 @@
...
+            subscriptOp->erase();
+          }
+        }
+
+        // Erase explicitly marked consteval operations from the original
+        // function body.
+        for (Operation *op : llvm::reverse(constevalOps)) {
+          op->erase();
+        }
+      }
```
svuckovicTT: Move this into a function (maybe multiple) and document the intent.

include/ttmlir/Conversion/Passes.td
```diff
@@ -257,6 +257,16 @@ def EmitPyLinkModules : Pass<"emitpy-link-modules", "::mlir::ModuleOp"> {
...
+  let summary = "Organize EmitPy operations into separate files.";
+  let description = [{
+    This pass organizes EmitPy operations into two files:
+    - main.py: Runtime execution logic.
+    - consteval.py: Complete consteval logic.
+  }];
+  let constructor = "createEmitPySplitFilesPass()";
```
mtopalovicTT: Why is constructor needed?

lib/Dialect/TTNN/Pipelines/TTNNPipelines.cpp
```diff
@@ -524,6 +524,10 @@ void createTTIRToEmitPyPipeline(OpPassManager &pm,
   // Link Device and CPU modules into the root module.
   //
   pm.addPass(createEmitPyLinkModulesPass());
+
+  // Split linked module into main and consteval files.
+  //
+  pm.addPass(createCodegenSplitFilesPass());
```
sdjordjevicTT: Can we add this under the pipeline switch? I still see a value that, for example, sometimes we want to keep everything in a single file. :)

lib/Conversion/TTNNToEmitPy/CodegenSplitFiles.cpp
```diff
@@ -0,0 +1,536 @@
...
+//
+//   def main_const_eval_0(input):               # Const-eval function
+//       ...
+//
+//   def _main(input):                           # Main with inline caching
+//       global _CONST_EVAL_CACHE
+//       ...
+//       utils_constEvalFuncWrapper_0 =
+//       utils.constEvalFuncWrapper("main_const_eval_0", ..., _CONST_EVAL_CACHE,
+//       ...) utils_constEvalFuncWrapper_0_0 = utils_constEvalFuncWrapper_0[0]
```
mtopalovicTT: format

lib/Conversion/TTNNToEmitPy/CodegenSplitFiles.cpp
```diff
@@ -0,0 +1,536 @@
...
+//
+// AFTER (split files):
+//   main.py:
+//     import ttnn
+//     import utils
+//     from consteval import consteval__main
+//
+//     def _main(input):
+//         var_0 = consteval__main(input)         # Call to consteval caching
+//         function
```
mtopalovicTT: what is function here?

lib/Conversion/TTNNToEmitPy/CodegenSplitFiles.cpp
```diff
@@ -0,0 +1,536 @@
...
+//     _CONST_EVAL_CACHE = {}
+//
+//     def cpu_hoisted_const_eval_0(input):      # Moved from main
+//         ...
+//
+//     def main_const_eval_0(input):             # Moved from main
+//         result = cpu_hoisted_const_eval_0(...)
+//         ...
+//
+//     def consteval__main(inputs):              # Extracted consteval caching
```
mtopalovicTT: format

lib/Conversion/TTNNToEmitPy/CodegenSplitFiles.cpp
```diff
@@ -0,0 +1,536 @@
...
+//===----------------------------------------------------------------------===//
+// Data Structures and Helper Functions
+//===----------------------------------------------------------------------===//
+
+namespace {
+// Constants for file names and operation names.
+constexpr const char *MAIN_FILE_NAME = "main";
+constexpr const char *CONSTEVAL_FILE_NAME = "consteval";
+constexpr const char *CONSTEVAL_WRAPPER_PREFIX = "constEvalFuncWrapper";
+constexpr const char *CACHE_CONSTEVAL_ATTR_NAME = "emitpy.cache_consteval";
```
mtopalovicTT: When I search "emitpy.cache_consteval" I find multiple places where we are using string literal to get or set this attribute. Can you move this declaration to some header file and substitute all those references with one global?

mtopalovicTT (commented):
I'm curious about the placement of the split file logic in the pipeline. Currently, CodegenSplitFilesPass runs after the conversion to EmitPy, which means the splitting logic needs to understand EmitPy-specific
  details like how load_cache_op gets decomposed.

  Have you considered moving the split file logic earlier in the pipeline, before the TTNN → EmitPy conversion? The potential benefits would be:

  1. Cleaner separation: The split logic would operate on higher-level IR (TTNN) and wouldn't need to be aware of EmitPy decomposition details
  2. More robust: Changes to how specific ops are lowered to EmitPy wouldn't affect the splitting logic
  3. Simpler implementation: We could categorize ops based on their consteval/cpu-hoisted attributes at the TTNN level, create the appropriate fileOp regions, and let the conversion passes handle each region
  independently

  The flow would be:
  - Before conversion: Analyze and split ops into fileOp regions (consteval file vs. main file) based on op attributes
  - During conversion: Each fileOp region gets converted to EmitPy independently

The way you are doing it seems a bit fragile and complex imo. Having said that I'm not sure what would be the drawbacks of doing the split before lowering.

amilovanovicTT:
> I'm curious about the placement of the split file logic in the pipeline. Currently, CodegenSplitFilesPass runs after the conversion to EmitPy, which means the splitting logic needs to understand EmitPy-specific details like how load_cache_op gets decomposed.
> 
> Have you considered moving the split file logic earlier in the pipeline, before the TTNN → EmitPy conversion? The potential benefits would be:
> 
> 1. Cleaner separation: The split logic would operate on higher-level IR (TTNN) and wouldn't need to be aware of EmitPy decomposition details
> 2. More robust: Changes to how specific ops are lowered to EmitPy wouldn't affect the splitting logic
> 3. Simpler implementation: We could categorize ops based on their consteval/cpu-hoisted attributes at the TTNN level, create the appropriate fileOp regions, and let the conversion passes handle each region
>    independently
> 
> The flow would be:
> 
> * Before conversion: Analyze and split ops into fileOp regions (consteval file vs. main file) based on op attributes
> * During conversion: Each fileOp region gets converted to EmitPy independently
> 
> The way you are doing it seems a bit fragile and complex imo. Having said that I'm not sure what would be the drawbacks of doing the split before lowering.

I agree with you, moving the split files logic earlier in the pipeline sounds like a better approach. Having said that, we will need to refactor how we handle loadCachedOp in the codegen path, because currently some ops from loadCachedOp decomposition need to be in the main file, while others need to be in the consteval file. I have synced offline with @svuckovicTT, and we agreed that I should investigate this further, since loadCachedOp has brought some hacky solutions in the codegen path previously. For now, I'll leave this PR open and get back to you soon.

include/ttmlir/Conversion/Passes.td
```diff
@@ -207,6 +207,22 @@ def ConvertTTNNToEmitPy : Pass<"convert-ttnn-to-emitpy", "::mlir::ModuleOp"> {
...
+  let summary = "Wrap complete consteval caching logic into the one if-guard.";
+  let description = [{
+    Post-conversion pass that walks functions containing a cache dictionary
+    and wraps const-eval calls (tagged with `emitpy.const_evaled`) and their
+    dict-set expressions in an `emitpy.if` guard:
+
+      if not dict:
+          dict["key"] = callee(inputs...)
+
+    This pass is to be used regardless of whether the TTNNFileSplit pass was performed.
```
svuckovicTT: Remove.

include/ttmlir/Dialect/EmitPy/IR/EmitPyOps.td
```diff
@@ -594,4 +594,39 @@ def EmitPy_ExpressionOp : EmitPy_Op<"expression", [
...
 }
 
+def EmitPy_FileOp
+    : EmitPy_Op<"file", [IsolatedFromAbove, NoRegionArguments, SymbolTable,
+                        NoTerminator]> {
+  let summary = "A file container operation";
+  let description = [{
+    A `file` represents a single Python file.
+
+    If `-file-id=id` flag is used with `ttmlir-translate`, then only the ops of the `emitpy.file` with the matching id are emitted.
```
svuckovicTT: Here it says `-file-id` but elsewhere it says `-emitpy-file-id`, can we consolidate. I slightly prefer `-file-id` as it's shorter.

include/ttmlir/Dialect/TTNN/Pipelines/TTNNPipelines.h
```diff
@@ -510,6 +510,11 @@ struct TTNNToEmitPyDevicePipelineOptions
           "original IR/code. Highly experimental; please file issues at "
           "https://github.com/tenstorrent/tt-mlir/issues"),
       llvm::cl::init(false)};
+
+  Option<bool> splitFiles{
+      *this, "split-files",
+      llvm::cl::desc("Split code into two files: consteval and main."),
```
svuckovicTT:
```suggestion
      llvm::cl::desc("Enables TTNNFileSplit pass"),
```

include/ttmlir/Dialect/EmitPy/IR/EmitPyOps.td
svuckovicTT:
```suggestion
    If `-file-id=id` flag is used with `ttmlir-translate`, then only the matching `emitpy.file` op is emitted.
```

include/ttmlir/Dialect/TTNN/Transforms/Passes.td
```diff
@@ -501,6 +501,45 @@ def TTNNSimplifyLocsForCodegen: Pass<"ttnn-simplify-locs-for-codegen", "::mlir::
...
+  ];
+  let options = [
+    Option<"target", "target", "FileSplitTarget",
+           /*default=*/"",
+           "Target dialect for file splitting",
+           [{::llvm::cl::values(
+              clEnumValN(FileSplitTarget::EmitPy, "emitpy", "EmitPy target"),
+              clEnumValN(FileSplitTarget::EmitC, "emitc", "EmitC target")
+           )}]>
+  ];
```
svuckovicTT, line 549:
```suggestion
  }];

  let dependentDialects = [
    "mlir::tt::emitpy::EmitPyDialect",
    "mlir::emitc::EmitCDialect",
    "mlir::tt::ttcore::TTCoreDialect"
  ];

  let options = [
    Option<"target", "target", "FileSplitTarget",
           /*default=*/"",
           "Target dialect for file splitting",
           [{::llvm::cl::values(
              clEnumValN(FileSplitTarget::EmitPy, "emitpy", "EmitPy target"),
              clEnumValN(FileSplitTarget::EmitC, "emitc", "EmitC target")
           )}]>
  ];
```

lib/Dialect/EmitPy/IR/EmitPyOps.cpp
```diff
@@ -11,6 +11,7 @@
 #include "llvm/Support/Casting.h"
 #include "llvm/Support/LogicalResult.h"
 
+#include "llvm/Support/Casting.h"
```
svuckovicTT:
```suggestion
```

svuckovicTT: ^ double include

lib/Dialect/TTNN/Pipelines/TTNNPipelines.cpp
```diff
@@ -467,10 +467,20 @@ void createTTNNToEmitPyDevicePipeline(
     }
   }
 
+  devicePm.addPass(createTTNNPrepareConstEvalCaching());
+  // Optionally split code into two files: consteval and main.
```
svuckovicTT:
```suggestion
  // Optionally run TTNNFileSplit pass
```

I'd avoid mentioning how many files there are. That will change soon and then these comments will be incorrect.

lib/Dialect/TTNN/Pipelines/TTNNPipelines.cpp
```diff
@@ -563,7 +573,16 @@ void createTTIRToEmitPyPipeline(OpPassManager &pm,
...
+
+  TTNNToEmitPyDevicePipelineOptions emitPyDeviceOptions;
+  emitPyDeviceOptions.targetModule = options.targetModule;
+  emitPyDeviceOptions.loadInputTensorsFromDisk =
+      options.loadInputTensorsFromDisk;
+  emitPyDeviceOptions.tensorLoadDirectory = options.tensorLoadDirectory;
+  emitPyDeviceOptions.tensorLoadFilePrefix = options.tensorLoadFilePrefix;
+  emitPyDeviceOptions.tryRecoverStructure = options.tryRecoverStructure;
+  emitPyDeviceOptions.splitFiles = options.splitFiles;
+  createTTNNToEmitPyDevicePipeline(pm, emitPyDeviceOptions);
```
svuckovicTT: Why is this needed?

lib/Dialect/TTNN/Transforms/TTNNToPython.cpp
```diff
@@ -26,7 +26,8 @@ LogicalResult emitTTNNAsPython(ModuleOp origOp, llvm::raw_ostream &os) {
     return failure();
   }
 
-  if (emitpy::translateToPython(op, os).failed()) {
+  std::string fileId = "";
+  if (emitpy::translateToPython(op, os, fileId).failed()) {
```
svuckovicTT: Can we make `fileId` a default-valued arg and set it to `std:nullopt`?

tools/builder/base/builder_runtime.py
svuckovicTT, line 1: I think we shouldn't be splitting files for builder. If there's fileops in the IR, let's error out here.

Btw it's a good idea to avoid basing logic on things that are prone to change and have no validation mechanism in compile-time, like comment strings `#=== main.py ===`.

tools/tt-alchemist/csrc/lib/generate_python.cpp
```diff
@@ -103,21 +93,56 @@ bool TTAlchemist::generatePython(const std::string &input_file,
     return false;
   }
 
-  // Create main.py with the generated Python code
+  // Check if the module has file ops (split-files mode).
+  bool hasSplitFiles = false;
+  module->walk([&](mlir::tt::emitpy::FileOp) { hasSplitFiles = true; });
```
svuckovicTT:
```suggestion
  module->walk([&hasSplitFiles](mlir::tt::emitpy::FileOp) { hasSplitFiles = true; });
```

tools/tt-alchemist/csrc/lib/generate_python.cpp
svuckovicTT, line 1: As mentioned earlier, we want to support the "legacy" scenario with no file ops and the old load caching mechanism to enable a smoother transition when uplifting into tt-xla.

lib/Dialect/TTNN/Pipelines/TTNNPipelines.cpp
amilovanovicTT: Agreed, changed.

lib/Dialect/TTNN/Transforms/TTNNToPython.cpp
amilovanovicTT: Sure, done.

lib/Dialect/TTNN/Pipelines/TTNNPipelines.cpp
amilovanovicTT: Not needed, removed.

include/ttmlir/Dialect/EmitPy/IR/EmitPyOps.td
amilovanovicTT: Changed everything to `-file-id`.

tools/tt-alchemist/csrc/lib/generate_python.cpp
sdjordjevicTT: @amilovanovicTT @svuckovicTT I wouldn't keep legacy just for transition scenarios. I would keep it as a feature, for example, when you want to create a TTNN repro, you would probably want to have everything in one file, hence the default option should be split into multiple files, but we should aim to have a working non-split version as well.

include/ttmlir/Dialect/EmitPy/IR/EmitPyOps.td
amilovanovicTT: Had to revert to `-emitpy-file-id` to avoid collision with MLIR's EmitC `-file-id`option, which was causing a duplicate registration error.

tools/builder/base/builder_runtime.py
amilovanovicTT: Done

lib/Conversion/TTNNToEmitPy/EmitPyLinkModules.cpp
```diff
@@ -108,48 +146,31 @@ class EmitPyLinkModulesPass
...
     }
 
     // Move all operations from CPU module (CPU-hoisted function
-    // definitions).
+    // definitions). If TTNNFileSplit was performed, move them to the consteval
+    // file. Otherwise, move them to the root module.
+    //
+    Block *blockToMoveCPUOpsTo = constevalFileOp
+                                     ? &constevalFileOp.getBodyRegion().front()
+                                     : blockToMoveCPUOpsTo = &rootBody;
```
vmilosevic: :warning: **clang\-analyzer\-deadcode\.DeadStores** :warning:
Although the value stored to `` blockToMoveCPUOpsTo `` is used in the enclosing expression, the value is never actually read from `` blockToMoveCPUOpsTo ``

tools/tt-alchemist/csrc/lib/generate_python.cpp
amilovanovicTT: Currently, there is a `splitFiles` option of  `TTNNToEmitPyDevicePipeline` to enable/disable `TTNNFileSplit` pass. 
When it comes to tt-xla, I could add `codegen_split_files` option to enable codegen with 
```
codegen_py(
        model, x, export_path="resnet50_codegen_true_new",
        compiler_options={"codegen_split_files": True}
    )
```
@svuckovicTT @sdjordjevicTT What is your take on this?

svuckovicTT: @sdjordjevicTT my bad on communication, the "single file" mode won't be removed, we need it for various test infras as well (ttir-builder, tt-xla model tests, ...)

@amilovanovicTT yup, we need this as an option thru tt-xla as well.

PR #6734, opened 2026-01-23
[TTIRFusion] Update LinearOp fusion pattern

### LINKED ISSUE #6733: [TTIRFusion] LinearOp incorrect output shape
TTIR fusion assumes that `LinearOp` supports broadcasting for all the arguments and generates fusion pattern accordingly. However, tt-metal only support broadcasting for `bias`. LinearOp output shape depends on `matmul` operation and `bias` is broadcasted to match `matmul` shape.

Consider the graph below.
```
func.func @matmul_bias_7(%arg0: tensor<256x1024xbf16>, %arg1: tensor<1024x1024xbf16>, %arg2: tensor<1024xbf16>) -> tensor<1x256x1024xbf16> {
  %0 = "ttir.matmul"(%arg0, %arg1) <{transpose_a = false, transpose_b = false}> : (tensor<256x1024xbf16>, tensor<1024x1024xbf16>) -> tensor<256x1024xbf16>
  %1 = "ttir.reshape"(%0) <{shape = [1 : i32, 256 : i32, 1024 : i32]}> : (tensor<256x1024xbf16>) -> tensor<1x256x1024xbf16>
  %2 = "ttir.reshape"(%arg2) <{shape = [1 : i32, 1 : i32, 1024 : i32]}> : (tensor<1024xbf16>) -> tensor<1x1x1024xbf16>
  %3 = "ttir.add"(%1, %2) : (tensor<1x256x1024xbf16>, tensor<1x1x1024xbf16>) -> tensor<1x256x1024xbf16>
  return %3 : tensor<1x256x1024xbf16>
}
```

TTIR fusion pass will convert it to
```
func.func @matmul_bias_7(%arg0: tensor<256x1024xbf16>, %arg1: tensor<1024x1024xbf16>, %arg2: tensor<1024xbf16>) -> tensor<1x256x1024xbf16> {
    %0 = "ttir.reshape"(%arg2) <{shape = [1 : i32, 1 : i32, 1024 : i32]}> : (tensor<1024xbf16>) -> tensor<1x1x1024xbf16>
    %1 = "ttir.linear"(%arg0, %arg1, %0) <{transpose_a = false, transpose_b = false}> : (tensor<256x1024xbf16>, tensor<1024x1024xbf16>, tensor<1x1x1024xbf16>) -> tensor<1x256x1024xbf16>
    return %1 : tensor<1x256x1024xbf16>
}
```

It is expected that `ttir.linear` will generate output `tensor<1x256x1024xbf16>` due to `bias`. But tt-metal generates output as `tensor<256x1024xbf16>` which can cause failures later.

The correct graph should be
```
func.func @matmul_bias_7(%arg0: tensor<256x1024xbf16>, %arg1: tensor<1024x1024xbf16>, %arg2: tensor<1024xbf16>) -> tensor<1x256x1024xbf16> {
    %0 = "ttir.linear"(%arg0, %arg1, %arg2) <{transpose_a = false, transpose_b = false}> : (tensor<256x1024xbf16>, tensor<1024x1024xbf16>, tensor<1024xbf16>) -> tensor<256x1024xbf16>
    %1 = "ttir.reshape"(%0) <{shape = [1 : i32, 256 : i32, 1024 : i32]}> : (tensor<256x1024xbf16>) -> tensor<1x256x1024xbf16>
    return %1 : tensor<1x256x1024xbf16>
}
```

mmanzoorTT:
### Ticket
closes #6733

### Problem description
Fusion pattern always uses broadcasted shape (for reshaped matmul fusion) as output shape although tt-metal only uses broadcasted shape if padded H dim of bias does not match with the TILE_HEIGHT. 

### What's changed
- Generate LinearOp with broadcasted shape if H dims of bias does not match with TILE_HEIGHT; otherwise, use matmul shape as output shape.
- Update TTIR and TTNN LinearOp verification to match this behaviour.

### Checklist
- [X] Existing/New tests provide coverage for changes

### DIFF
--- a/lib/Dialect/TTIR/IR/TTIROps.cpp
+++ b/lib/Dialect/TTIR/IR/TTIROps.cpp
@@ -15,6 +15,7 @@
 #include "ttmlir/Dialect/TTIR/Utils/VerificationUtils.h"
 #include "ttmlir/Dialect/TTNN/IR/TTNNOpsAttrs.h"
 #include "ttmlir/Dialect/TTNN/Types/Types.h"
+#include "ttmlir/Dialect/TTNN/Utils/Utils.h"
 #include "ttmlir/Utils.h"
 
 #include "mlir/Dialect/Bufferization/IR/Bufferization.h"
@@ -3297,13 +3298,32 @@ bool mlir::tt::ttir::TTNNMetalLayoutCastOp::bufferizesToMemoryWrite(
 
     // Verify that the dimensions of the matmul of A and B are broadcast
     // compatible with input bias.
-    llvm::SmallVector<int64_t> matmulShape = expectedOutputShape;
-    if (!mlir::OpTrait::util::getBroadcastedShape(matmulShape, biasShape,
-                                                  expectedOutputShape)) {
+    llvm::SmallVector<int64_t> broadcastShape;
+    if (!mlir::OpTrait::util::getBroadcastedShape(expectedOutputShape,
+                                                  biasShape, broadcastShape)) {
       return emitOpError("Bias shape(")
              << ttmlir::utils::join(biasShape, ",")
              << ") is not broadcast compatible with the matmul output shape("
-             << ttmlir::utils::join(matmulShape, ",") << ")";
+             << ttmlir::utils::join(expectedOutputShape, ",") << ")";
+    }
+
+    // tt-metal uses a composite LinearOp where the bias is added after the
+    // matmul, and ttnn.add supports broadcasting of both operands. Otherwise,
+    // tt-metal lowers to a fused LinearOp, which uses the matmul result shape
+    // as the output shape. The composite LinearOp requires that the bias
+    // second-to-last dim (of padded shape) does not match the tile height.
+    // Update the expected output shape to the fully broadcasted shape when:
+    // 1) The matmul result is a scalar (vector x vector), so the inferred
+    //    output shape is empty and must be derived from the bias via
+    //    broadcasting, or
+    // 2) The bias second-to-last dim (of padded shape) does not match the
+    //    tile height, indicating the composite LinearOp is used.
+    llvm::SmallVector<int64_t> paddedBiasShape =
+        ttnn::utils::getTilePaddedShape(biasShape);
+    if (expectedOutputShape.empty() ||
+        (paddedBiasShape.size() > 1 &&
+         paddedBiasShape[paddedBiasShape.size() - 2] != ttnn::TILE_HEIGHT)) {
+      expectedOutputShape = broadcastShape;
     }
   }
 
--- a/lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
+++ b/lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
@@ -6,6 +6,8 @@
 #include "ttmlir/Dialect/TTIR/IR/TTIROps.h"
 #include "ttmlir/Dialect/TTIR/Transforms/Passes.h"
 #include "ttmlir/Dialect/TTIR/Utils/Utils.h"
+#include "ttmlir/Dialect/TTNN/Types/Types.h"
+#include "ttmlir/Dialect/TTNN/Utils/Utils.h"
 #include "ttmlir/Utils.h"
 
 #include "mlir/Analysis/TopologicalSortUtils.h"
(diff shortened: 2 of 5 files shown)


lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
```diff
@@ -1348,24 +1349,21 @@ class MatmulWithBiasFusionPattern : public mlir::OpRewritePattern<AddOp> {
       TypedValue<RankedTensorType> bias =
           (addOp.getLhs() == reshapeOp.getResult()) ? addOp.getRhs()
                                                     : addOp.getLhs();
-      auto biasType = bias.getType();
+
+      if (llvm::isa_and_nonnull<ttir::ReshapeOp>(bias.getDefiningOp())) {
```
mtopalovicTT: You can use short hand version here maybe:
if (auto reshape = bias.getDefiningOp<ReshapeOp>()) {
}

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
```diff
@@ -1348,24 +1349,21 @@ class MatmulWithBiasFusionPattern : public mlir::OpRewritePattern<AddOp> {
       TypedValue<RankedTensorType> bias =
```
mtopalovicTT:
```
TTIR fusion assumes that LinearOp supports broadcasting for all the arguments and generates fusion pattern accordingly. However, tt-metal only support broadcasting for bias.
```

Do we need to validate in `getFusableReshapedMatmulOp` that multiply is not being broadcasted?

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
mmanzoorTT: Thanks for the suggestion.

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
mmanzoorTT: tt-metal treats the bias as defining output shape and broadcasts the matmul result if the trailing dims of bias and matmul matches. I have updated the fusion pattern to use bias shape as output shape if trailing dims match; otherwise, use matmul shape. 
I also have updated TTIR and TTN LinearOp verification to generate output shape accordingly.

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
```diff
@@ -1327,59 +1327,58 @@ class MatmulWithBiasFusionPattern : public mlir::OpRewritePattern<AddOp> {
...
+    auto biasType = bias.getType();
+    // If the bias’s trailing two dimensions match the trailing dimensions of
+    // the matmul result, the bias is applied per-matrix. In this case, LinearOp
+    // broadcasts the matmul result across the leading dimensions of the bias.
+    // So the output type should follow the bias shape.
+    if (biasType.getRank() >= 2 && outputType.getRank() >= 2 &&
+        llvm::equal(outputType.getShape().take_back(2),
+                    biasType.getShape().take_back(2))) {
+      outputType = biasType;
+    }
```
mtopalovicTT: I don't get this part. If trailing dimensions are the same we take bias shape as output shape of linear, this don't seem correct to me. What if we have bias which has shape 1, 1, 32, 32 and matmul has shape 2, 2, 32, 32?

mmanzoorTT: You are right. I was only considering `bias` shapes. LinearOp should use broadcasted shape if the trailing dims match otherwise use matmul shape. 
If bias is 1, 1, 32, 32 and matmul is 2, 2, 32, 32 then output will be 2, 2, 32, 32
if bias is 1, 1, 32, 32 and matmul is 32, 32 then the output will be 1, 1, 32, 32

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
```diff
@@ -15,6 +15,7 @@
 #include "mlir/Transforms/GreedyPatternRewriteDriver.h"
 #include "llvm/ADT/SmallSet.h"
 #include "llvm/ADT/TypeSwitch.h"
+#include "llvm/Support/raw_ostream.h"
```
mtopalovicTT: Not needed?

lib/Dialect/TTIR/IR/TTIROps.cpp
```diff
@@ -3297,13 +3297,28 @@ ::mlir::LogicalResult mlir::tt::ttir::LinearOp::verify() {
...
+    llvm::ArrayRef<int64_t> outputShapeRef = expectedOutputShape;
+    // Update the expected output shape to the fully broadcasted shape when:
+    // 1) The matmul result is a scalar (vector x vector), so the inferred
+    //    output shape is empty and must be derived from the bias via
+    //    broadcasting, or
+    // 2) The bias has rank >= 2 and its trailing two dimensions match the
+    //    trailing dimensions of the matmul output, meaning the bias is applied
+    //    per-matrix and broadcast across leading dimensions.
+    if (outputShapeRef.empty() ||
+        ((biasType->getRank() >= 2) && (outputShapeRef.size() >= 2) &&
```
mtopalovicTT: What is the expected output shape if outputShapeRef and biasType trailing dims are not equal? it shouldn't be whatever is the result of matmul alone, so we should fail the validation no? This seems like some ttnn limitation (which is probably justified) but we should at least fail the op validation or decompose into matmul + add.

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
mtopalovicTT: Is it correct to use matmul shape in else case?

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
mmanzoorTT: Thanks, removed.

lib/Dialect/TTIR/IR/TTIROps.cpp
mmanzoorTT: [tt-metal](https://github.com/tenstorrent/tt-metal/blob/2fc0378862fb2abaffcdb119b9982b3a70802848/ttnn/cpp/ttnn/operations/matmul/matmul.cpp#L116)  uses bias's second-to-last dimension (of padded shape) to decide whether to post process bias with ttnn.add (as composite op) or process it along with the matmul/linear op (as fused op). ttnn.add supports broadcasting of both LHS and RHS; however, fused op only broadcast bias to match matmul shape. 

```
        // If bias second-to-last dimension doesn't match tile height, must post-process
        if (bias_padded_shape[-2] != tile_height) {
            post_process_bias = true;
        }
```

`TILE_HEIGHT=32` as `defined in include/ttmlir/Dialect/TTNN/Types/Types.h`

For example, If bias is 1, 1, 768 and matmul is 64, 768 then ttnn output is 64, 768. As the second-to-last dim will be 32 after padding, so tt-metal will use matmul shape as output shape. 
If bias is 1, 64, 768 then ttnn.add will be used for adding bias and second-to-last dim is not 32 so output shape will be 1, 64, 768.

There are other scenarios which can trigger post processing of bias (depends on program_config and core_grid) but we are not modelling them in TTNN dialect. So I am ignoring those scenarios.

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
mmanzoorTT: Please have a look at comment below https://github.com/tenstorrent/tt-mlir/pull/6734#discussion_r2739890575

lib/Dialect/TTIR/IR/TTIROps.cpp
mtopalovicTT: Okay maybe my assumption that bias can have batch dimensions which are > 1 is wrong. If it's wrong than this is fine.

What would happen if H dimension is tile aligned but W dimension is not
For example:
matmul shape is 1, 1, 32, 256
bias shape is 32, 512

will metal use composite or fused?

mmanzoorTT: These shapes are not compatible. W dim (256/512) cannot be broadcasted. Compiler will never create such addition. If you create add with these shape then canonicalize pass will fail.
```
func.func @dot_general_with_bias_5(%arg0: tensor<1x1x32x2048xbf16>, %arg1: tensor<2048x256xbf16>, %bias: tensor<32x512xbf16>) -> tensor<1x1x32x512xbf16> {
  %0 = "ttir.matmul"(%arg0, %arg1) <{transpose_a = false, transpose_b = false}> : (tensor<1x1x32x2048xbf16>, tensor<2048x256xbf16>) -> tensor<1x1x32x256xbf16>
  %2 = "ttir.add"(%0, %bias) : (tensor<1x1x32x256xbf16>, tensor<32x512xbf16>) -> tensor<1x1x32x512xbf16>
  return %2 : tensor<1x1x32x512xbf16>
}
```
Here is the output
```
ttmlir-opt test/test.mlir 
test/test.mlir:22:10: error: 'ttir.add' op operand shape (32, 512) is not broadcast compatible with inferred operand shapes (1, 1, 32, 256)
    %2 = "ttir.add"(%0, %bias) : (tensor<1x1x32x256xbf16>, tensor<32x512xbf16>) -> tensor<1x1x32x512xbf16>
         ^
test/test.mlir:22:10: note: see current operation: %1 = "ttir.add"(%0, %arg2) : (tensor<1x1x32x256xbf16>, tensor<32x512xbf16>) -> tensor<1x1x32x512xbf16>
```

mtopalovicTT: Yea not sure why I gave that example that is obviously not broadcastable, but now I forgot what case I had in mind...

PR #6741, opened 2026-01-24
using TTMLIR_PYTHON_VERSION throughout ./env/init_venv.sh

iharabukhouski:
### Ticket
na

### Problem description
one part of the script allows for custom python version (ie line 14) and the other does not (ie lines 28, 30, 31, 32)

### What's changed
replaced hardcoded references to "python" with "$TTMLIR_PYTHON_VERSION"

### Checklist
~- [ ] New/Existing tests provide coverage for changes~ - na

### DIFF
--- a/tools/builder/base/builder_apis.py
+++ b/tools/builder/base/builder_apis.py
@@ -59,6 +59,7 @@ def _compile_and_execute(
     check_atol: bool = False,
     check_rtol: bool = False,
     enable_intermediate_verification: bool = False,
+    dump_memory: bool = False,
     **compile_kwargs,
 ) -> str:
     builder, compiled_bin, input_output_goldens, intermediate_goldens = compile_fn(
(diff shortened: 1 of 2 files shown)


env/init_venv.sh
```diff
@@ -25,8 +25,8 @@ if [ ! -d "$TTMLIR_TOOLCHAIN_DIR" ]; then
   source $ENV_DIR/activate
 fi
 
-python -m pip install --upgrade pip
+$TTMLIR_PYTHON_VERSION -m pip install --upgrade pip
```
nsmithtt, line 28: FYI after running `source env/activate` which also activates the python environment, the `python` command is symlinked to the version of python that was used to create that venv.

iharabukhouski: yep, i realized that after creating the pr and that's why i closed it
i am sorry for that miss on my par. i will try to be more careful in the future to avoid over spamming tt team

nsmithtt: No worries! Thanks for looking it!

PR #6743, opened 2026-01-25
TTNN/D2M Integration Pt1

vtangTT:
### Ticket
Link to Github Issue

### Problem description
Initial commit for TTNN + D2M integration effort. This PR contains everything but optimizer integration + fallback (and maybe some minor rewrite passes).

Initial design [doc](https://docs.google.com/document/d/1NXtWf1ng8bfArJMFMptuOAI9eJSlrshQ7G1ZTydCeHE/edit?usp=sharing), IR examples might not be up to date but the idea is the same

### What's changed
- new `ttnn.d2m_subgraph` op that reference the ttnn subgraph to be compiled through d2m
    - the subgraph will be held in a private func in the same module
- new `TTNND2MFusion` pass that detects simple eltwise fusion groups (method commented in code)
- new `createTTNNPipelineD2MPass` pipeline under nested PassManager
    - will only run on the newly defined op to compile the nested ttnn subgraph through D2M
- new `TTNND2MCollaspe` pass to inline the compile ttnn.generics and kernel code back into the main func/module 
- new `enable-d2m-fusing-pass` (defaults to false) that enables `TTNND2MFusion` detection pass and subsequent pass

### Checklist
    - lit tests for each pass/pipeline that was added
    - `ttnn_d2m_e2e.mlir` should be changed to just call the whole TTNN pipeline flow (not individual passes one after other) once optimizer integration is complete

### DIFF
--- a/include/ttmlir/Dialect/TTNN/IR/TTNNOps.td
+++ b/include/ttmlir/Dialect/TTNN/IR/TTNNOps.td
@@ -2939,7 +2939,7 @@ def TTNN_GenericOp : TTNN_Op<"generic", [TTNN_MemoryConfigOpInterface, MemoryEff
     let hasVerifier = 1;
 }
 
-def TTNN_DispatchD2MOp : TTNN_MemoryEffectOp<"dispatch_d2m",
+def TTNN_D2MSubgraphOp : TTNN_MemoryEffectOp<"d2m_subgraph",
   [AttrSizedOperandSegments]> {
   let summary = "Dispatch D2M compiled subgraph.";
   let description = [{
@@ -2958,7 +2958,7 @@ def TTNN_DispatchD2MOp : TTNN_MemoryEffectOp<"dispatch_d2m",
     Example:
     ```mlir
     // Before D2M compilation
-    %result = ttnn.dispatch_d2m @d2m_subgraph
+    %result = ttnn.d2m_subgraph @d2m_subgraph
         ins(%a : tensor<...>)
         outs(%out : tensor<...>) : tensor<...>
 
@@ -2967,7 +2967,7 @@ def TTNN_DispatchD2MOp : TTNN_MemoryEffectOp<"dispatch_d2m",
     }
 
     // After D2M compilation
-    %result = ttnn.dispatch_d2m @d2m_subgraph
+    %result = ttnn.d2m_subgraph @d2m_subgraph
         ins(%a : tensor<...>)
         outs(%out : tensor<...>) : tensor<...>
 
--- a/lib/Conversion/TTNNToTTIR/TTNNToTTIRPass.cpp
+++ b/lib/Conversion/TTNNToTTIR/TTNNToTTIRPass.cpp
@@ -36,7 +36,7 @@ struct ConvertTTNNToTTIRPass
 
   void collectD2MSubgraphs(ModuleOp moduleOp) {
     d2mSubgraphs.clear();
-    moduleOp.walk([&](ttnn::DispatchD2MOp dispatchOp) {
+    moduleOp.walk([&](ttnn::D2MSubgraphOp dispatchOp) {
       if (func::FuncOp func = dispatchOp.getD2MMainFunc()) {
         d2mSubgraphs.insert(func);
       }
@@ -66,8 +66,7 @@ struct ConvertTTNNToTTIRPass
 
     // Mark TTNN dialect as dynamically legal for all ops that:
     // 1. Do not have the `hoist_generic_via_d2m` attribute, and
-    // 2. Are not inside a D2M subgraph function (referenced by dispatch_d2m
-    // ops).
+    // 2. Are not inside a D2M subgraph function (referenced by d2m_subgraph)
     target.addDynamicallyLegalDialect<ttnn::TTNNDialect>([this](Operation *op) {
       if (isInsideD2MSubgraph(op)) {
         return false;
--- a/lib/Dialect/TTNN/IR/TTNNOps.cpp
+++ b/lib/Dialect/TTNN/IR/TTNNOps.cpp
@@ -4746,10 +4746,10 @@ mlir::tt::ttnn::PagedScaledDotProductAttentionDecodeOp::verify() {
 }
 
 //===----------------------------------------------------------------------===//
-// DispatchD2MOp
+// D2MSubgraphOp
 //===----------------------------------------------------------------------===//
 
-::mlir::LogicalResult mlir::tt::ttnn::DispatchD2MOp::verify() {
+::mlir::LogicalResult mlir::tt::ttnn::D2MSubgraphOp::verify() {
   func::FuncOp mainFunc = getD2MMainFunc();
   if (!mainFunc) {
     return emitOpError("Could not find D2M function '")
@@ -4793,7 +4793,7 @@ mlir::tt::ttnn::PagedScaledDotProductAttentionDecodeOp::verify() {
   return success();
 }
 
-func::FuncOp mlir::tt::ttnn::DispatchD2MOp::getD2MMainFunc() {
+func::FuncOp mlir::tt::ttnn::D2MSubgraphOp::getD2MMainFunc() {
   ModuleOp parentModule = getOperation()->getParentOfType<ModuleOp>();
   if (!parentModule) {
     return nullptr;
@@ -4805,7 +4805,7 @@ func::FuncOp mlir::tt::ttnn::DispatchD2MOp::getD2MMainFunc() {
       SymbolTable::lookupSymbolIn(parentModule, nameAttr));
 }
 
-void mlir::tt::ttnn::DispatchD2MOp::getEffects(
+void mlir::tt::ttnn::D2MSubgraphOp::getEffects(
     SmallVectorImpl<SideEffects::EffectInstance<MemoryEffects::Effect>>
         &effects) {
   effects.emplace_back(MemoryEffects::Read::get(),
--- a/lib/Dialect/TTNN/Pipelines/TTNNPipelines.cpp
+++ b/lib/Dialect/TTNN/Pipelines/TTNNPipelines.cpp
@@ -634,11 +634,11 @@ void registerTTNNPipelines() {
       "XLA/Torch. ",
       mlir::tt::ttnn::createRecoverStructureXLATorchPipeline);
 
-  // TTNN D2M pipeline - runs D2M compilation on TTNN dispatch_d2m ops.
+  // TTNN D2M pipeline - runs D2M compilation on TTNN d2m_subgraph ops.
   //
   mlir::PassPipelineRegistration<>(
       "ttnn-through-d2m-pipeline",
-      "Pipeline to compile D2M subgraphs inside ttnn.dispatch_d2m ops.",
+      "Pipeline to compile D2M subgraphs inside ttnn.d2m_subgraph ops.",
       [](OpPassManager &pm) {
         auto &devicePm =
             pm.nest<ttcore::DeviceModuleOp>().nest<mlir::ModuleOp>();
--- a/lib/Dialect/TTNN/Transforms/TTNNCollaspeD2M.cpp
+++ b/lib/Dialect/TTNN/Transforms/TTNNCollaspeD2M.cpp
@@ -27,7 +27,7 @@ public:
     ModuleOp moduleOp = getOperation();
     SmallVector<func::FuncOp> funcsToDelete;
 
-    moduleOp.walk([&](DispatchD2MOp dispatchOp) {
+    moduleOp.walk([&](D2MSubgraphOp dispatchOp) {
       func::FuncOp mainFunc = dispatchOp.getD2MMainFunc();
       if (failed(inlineDispatchOp(moduleOp, dispatchOp))) {
         signalPassFailure();
@@ -43,7 +43,7 @@ public:
 
 private:
   LogicalResult inlineDispatchOp(ModuleOp parentModule,
-                                 DispatchD2MOp dispatchOp) {
+                                 D2MSubgraphOp dispatchOp) {
     func::FuncOp mainFunc = dispatchOp.getD2MMainFunc();
     if (!mainFunc) {
       return dispatchOp.emitOpError("could not find D2M function '")
--- a/lib/Dialect/TTNN/Transforms/TTNND2MFusing.cpp
+++ b/lib/Dialect/TTNN/Transforms/TTNND2MFusing.cpp
@@ -284,7 +284,7 @@ private:
     rewriter.create<func::ReturnOp>(loc, returnValues);
 
     rewriter.setInsertionPoint(firstOp);
-    auto dispatchOp = rewriter.create<DispatchD2MOp>(
+    auto dispatchOp = rewriter.create<D2MSubgraphOp>(
         loc, outputTypes, inputs.getArrayRef(), outputBuffers,
         SymbolRefAttr::get(rewriter.getContext(), funcName));
 
--- a/test/ttmlir/Dialect/TTNN/dispatch_d2m/collaspe.mlir
+++ b/test/ttmlir/Dialect/TTNN/dispatch_d2m/collaspe.mlir
@@ -6,9 +6,9 @@
 module {
   // CHECK-LABEL: func.func @one_d2m_subgraph
   func.func @one_d2m_subgraph(%arg0: tensor<32x32xf32, #ttnn_layout>, %out: tensor<32x32xf32, #ttnn_layout>) -> tensor<32x32xf32, #ttnn_layout> {
-    // CHECK-NOT: ttnn.dispatch_d2m
+    // CHECK-NOT: ttnn.d2m_subgraph
     // CHECK: "ttnn.generic"
-    %0 = ttnn.dispatch_d2m @d2m_subgraph_0
+    %0 = ttnn.d2m_subgraph @d2m_subgraph_0
         ins(%arg0 : tensor<32x32xf32, #ttnn_layout>)
         outs(%out : tensor<32x32xf32, #ttnn_layout>) : tensor<32x32xf32, #ttnn_layout>
     return %0 : tensor<32x32xf32, #ttnn_layout>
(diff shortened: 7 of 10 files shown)


test/ttmlir/Dialect/TTNN/dispatch_d2m/collaspe.mlir
```diff
@@ -0,0 +1,118 @@
...
+module {
+  // CHECK-LABEL: func.func @one_d2m_subgraph
+  func.func @one_d2m_subgraph(%arg0: tensor<32x32xf32, #ttnn_layout>, %out: tensor<32x32xf32, #ttnn_layout>) -> tensor<32x32xf32, #ttnn_layout> {
+    // CHECK-NOT: ttnn.dispatch_d2m
+    // CHECK: "ttnn.generic"
+    %0 = ttnn.dispatch_d2m @d2m_subgraph_0
+        ins(%arg0 : tensor<32x32xf32, #ttnn_layout>)
+        outs(%out : tensor<32x32xf32, #ttnn_layout>) {
+      builtin.module {
+        func.func @d2m_subgraph_0(%input: tensor<32x32xf32, #ttnn_layout>, %output: tensor<32x32xf32, #ttnn_layout>) -> tensor<32x32xf32, #ttnn_layout> {
```
nsmithtt: Did we consider just putting `d2m_subgraph_0` as a private func somewhere else in the top level module?

vtangTT: Do you mean for this test case or for the design in general

nsmithtt: The design in general, having a nested module like this feels a little weird, there's globals like the device and system descriptors that belong to the top level module, it seems cleaner to have all funcs/ops belong to the same module and reference its symbols / attrs.

nsmithtt: e.g.

We could additionally tag these funcs with `d2m.subgraph` or something so that the ttnn pass manager knows not to traverse into them.

```mlir
  func.func @one_d2m_subgraph(%arg0: tensor<32x32xf32, #ttnn_layout>, %out: tensor<32x32xf32, #ttnn_layout>) -> tensor<32x32xf32, #ttnn_layout> {
    %0 = ttnn.dispatch_d2m @d2m_subgraph_0
        ins(%arg0 : tensor<32x32xf32, #ttnn_layout>)
        outs(%out : tensor<32x32xf32, #ttnn_layout>)
    return %0 : tensor<32x32xf32, #ttnn_layout>
  }

  func.func private @d2m_subgraph_0(%input: tensor<32x32xf32, #ttnn_layout>,
                                    %output: tensor<32x32xf32, #ttnn_layout>)
                                    -> tensor<32x32xf32, #ttnn_layout> attributes {d2m.subgraph} {
    "ttnn.generic"(%input, %output) <{program = #ttnn.program<kernels = [#ttnn.read_kernel<symbol_ref = @kernel0, core_ranges = <[#ttnn.core_range<(0,0), (0,0)>]>, ct_args = [#ttnn.kernel_arg_cb_buffer_index<0>], common_rt_args = [], rt_args = []>], cbs = [<total_size = 4096, core_ranges = <[#ttnn.core_range<(0,0), (0,0)>]>, formats = [<buffer_index = 0, dtype = f32, page_size = 4096>], buffer = #ttnn.kernel_cb_global_buffer_address_of_tensor<0>>], semaphores = []>}> : (tensor<32x32xf32, #ttnn_layout>, tensor<32x32xf32, #ttnn_layout>) -> ()
    return %output : tensor<32x32xf32, #ttnn_layout>
  }

  func.func private @kernel0() attributes {ttkernel.thread = #ttkernel.thread<noc>} {
    return
  }
```

vtangTT: as per offline discussion, will be changing design to above and pipecleaning the d2m pipeline passes to handle running on multiple funcs.

vtangTT: `ttnn.dispatch_d2m` is now referencing a private function holding the ttnn subgraph to be compiled through D2M.

The D2M pipeline will now only run once over the entire module as discussed.

include/ttmlir/Dialect/TTNN/IR/TTNNOps.td
```diff
@@ -2878,6 +2878,73 @@ def TTNN_GenericOp : TTNN_Op<"generic", [TTNN_MemoryConfigOpInterface, MemoryEff
     let hasVerifier = 1;
 }
 
+def TTNN_DispatchD2MOp : TTNN_MemoryEffectOp<"dispatch_d2m",
+  [AttrSizedOperandSegments,
+   DestinationStyleOpInterface]> {
```
rpavlovicTT: We don't use DPS in TTNN dialect anymore

lib/Dialect/TTNN/Pipelines/TTNNPipelines.cpp
```diff
@@ -290,7 +293,22 @@ void createTTIRToTTNNDevicePipeline(
...
+      if (!options.optimizerPassEnabled) {
+        llvm::llvm_unreachable_internal(
+            "D2M fusing pass only supported with optimizer enabled");
+      }
+      devicePm.addPass(tt::ttnn::createTTNND2MFusing());
+    }
+
     createTTNNPipelineAnalysisPasses(devicePm, options);
+
+    if (options.enableD2MFusing) {
```
rpavlovicTT, line 311: Should we check for `options.optimizerPassEnabled` here as well?

lib/Dialect/TTNN/Transforms/TTNND2MFusing.cpp
```diff
@@ -0,0 +1,309 @@
...
+  static bool isElementwiseOp(Operation *op) {
+    if (!op) {
+      return false;
+    }
+    // Eltwise unary ops.
+    if (mlir::isa<AbsOp, CbrtOp, CeilOp, SignOp, CosOp, ExpOp, ErfOp, ErfcOp,
+                  FloorOp, GeluOp, IsFiniteOp, LogicalNotOp, BitwiseNotOp,
+                  NegOp, TanOp, TanhOp, ReciprocalOp, ReluOp, SinOp, SqrtOp,
+                  RsqrtOp, SigmoidOp, HardsigmoidOp, SiluOp, MishOp, LogOp,
+                  Log1pOp, Expm1Op>(op)) {
```
rpavlovicTT, line 52: Kinda weird we have cleaner way of checking for eltwise ops in ttnn IR. Just noting we should have such trait in the dialect.

lib/Dialect/TTNN/Transforms/TTNND2MFusing.cpp
```diff
@@ -0,0 +1,309 @@
...
+
+      if (group.size() >= 2) {
+        for (Operation *op : group) {
+          visitedOps.insert(op);
+        }
+        fusionGroups.push_back(std::move(group));
+      }
+    }
+
+    LLVM_DEBUG({
```
rpavlovicTT: Please use TTMLIR_DEBUG

include/ttmlir/Dialect/TTNN/IR/TTNNOps.td
nsmithtt: Due to allocator requirements I think we will need ttnn to pre-allocate the output tensor(s) for us.  That said, this op doesn't need to have the DPS trait, it can just have outputs explicitly passed in + memory effects and that should be sufficient.

lib/Dialect/TTNN/Pipelines/TTNNPipelines.cpp
vtangTT: The first check on line 296 already guarantees optimizer is enabled by this point, so would be redundant to check here.

lib/Dialect/TTNN/Transforms/TTNND2MFusing.cpp
vtangTT: I don't see an existing trait for this in the TTNN dialect :(

The reason I did it this way so I could match what is supported in the `TTNNToTTIR` conversion pass.

vtangTT: I can add a helper to distinguish eltwise ops, not sure how we want to handle the case where they're not supported in the conversion pass?

lib/Dialect/TTNN/Transforms/TTNND2MFusing.cpp
vtangTT: Changed to use `TTMLIR_DEBUG` under log component optimizer

include/ttmlir/Dialect/TTNN/IR/TTNNOps.td
vtangTT: removed the DPS trait

lib/Dialect/TTNN/Transforms/TTNND2MFusing.cpp
```diff
@@ -0,0 +1,307 @@
...
+      }
+    }
+
+    TTMLIR_DEBUG(ttmlir::LogComponent::Optimizer,
+                 "ElementwiseFusionAnalysis found {} valid fusion groups",
+                 fusionGroups.size());
+    for (const auto &group : fusionGroups) {
+      TTMLIR_DEBUG(ttmlir::LogComponent::Optimizer,
+                   "  Fusion group with {} ops", group.size());
+      for (Operation *op : group) {
```
vmilosevic: :warning: **clang\-analyzer\-deadcode\.DeadStores** :warning:
Value stored to `` op `` during its initialization is never read

vmilosevic: :warning: **clang\-diagnostic\-unused\-variable** :warning:
unused variable `` op ``

include/ttmlir/Dialect/TTNN/IR/TTNNOps.td
sgholamiTT: Related [issue](https://github.com/tenstorrent/tt-mlir/issues/6821), cc @nsmithtt

lib/Dialect/TTNN/Transforms/TTNND2MFusing.cpp
rpavlovicTT: `optimizer` log is already overused. Feel free to add `fusing` component and use it.

lib/Dialect/TTNN/Transforms/TTNND2MFusing.cpp
rpavlovicTT: I meant we should have it, as in we should try to create such trait. Right now I don't see better way than yours.

vtangTT: Created issue here https://github.com/tenstorrent/tt-mlir/issues/6834

lib/Dialect/TTNN/Transforms/TTNND2MFusing.cpp
vtangTT: Added `D2MFusion` log component

include/ttmlir/Dialect/TTNN/IR/TTNNOps.td
```diff
@@ -2939,6 +2936,67 @@ def TTNN_GenericOp : TTNN_Op<"generic", [TTNN_MemoryConfigOpInterface, MemoryEff
...
+
+  let results = (outs Variadic<AnyRankedTensor>:$results);
+
+  let hasVerifier = 1;
+
+  let assemblyFormat = [{
+    $d2m_func attr-dict `\n`
+    `ins` `(` $inputs `:` type($inputs) `)` `\n`
+    `outs` `(` $outputs `:` type($outputs) `)` (`:` type($results)^)?
+  }];
```
sgholamiTT, line 2995: I don't know about the main reason you decided to add `assemblyFormat` for this op, but this format prints sth like this:
```mlir
%1 = "ttnn.matmul"(%arg0, %arg1) <{transpose_a = false, transpose_b = false}> : (tensor<64x128xbf16, #ttnn_layout>, tensor<128x256xbf16, #ttnn_layout>) -> tensor<64x256xbf16, #ttnn_layout>
%2 = "ttnn.empty"(%0) <{dtype = #ttcore.supportedDataTypes<bf16>, layout = #ttnn.layout<tile>, memory_config = #ttnn.memory_config<#l1, <interleaved>>, shape = #ttnn.shape<64x256>}> : (!ttnn.device) -> tensor<64x256xbf16, #ttnn_layout>
%3 = ttnn.dispatch_d2m @d2m_subgraph_0
ins(%1 : tensor<64x256xbf16, #ttnn_layout>)
outs(%2 : tensor<64x256xbf16, #ttnn_layout>) : tensor<64x256xbf16, #ttnn_layout>
%4 = ... 
```
Which is inconsistent with all other TTNN ops. If we just simply remove it we'll get:
```mlir
%1 = "ttnn.matmul"(%arg0, %arg1) <{transpose_a = false, transpose_b = false}> : (tensor<64x128xbf16, #ttnn_layout>, tensor<128x256xbf16, #ttnn_layout>) -> tensor<64x256xbf16, #ttnn_layout>
%2 = "ttnn.empty"(%0) <{dtype = #ttcore.supportedDataTypes<bf16>, layout = #ttnn.layout<tile>, memory_config = #ttnn.memory_config<#l1, <interleaved>>, shape = #ttnn.shape<64x256>}> : (!ttnn.device) -> tensor<64x256xbf16, #ttnn_layout>
%3 = "ttnn.dispatch_d2m"(%1, %2) <{d2m_func = @d2m_subgraph_0, operandSegmentSizes = array<i32: 1, 1>}> : (tensor<64x256xbf16, #ttnn_layout>, tensor<64x256xbf16, #ttnn_layout>) -> tensor<64x256xbf16, #ttnn_layout>
%4 = ...
```
Which seems perfectly aligned with other ops.
Also, I guess our policy is to avoid writing `assemblyFormat` for `TTNN` ops (as no other op has it).

But feel free to ignore this comment if there are good reasons for having `assemblyFormat` that I'm not aware of.

sgholamiTT (approved):

include/ttmlir/Dialect/TTNN/IR/TTNNOps.td
vtangTT: Originally added this to mirror the structure of the `d2m.generic` since it used to hold the subgraph in it's nested region. That is not the case anymore.

It's more readable with the format imo, but no strong opinion on this 🤷

lib/Dialect/TTNN/IR/TTNNOps.cpp
```diff
@@ -4739,4 +4745,56 @@ ::mlir::LogicalResult mlir::tt::ttnn::AssignOp::verify() {
...
+
+  for (auto [idx, pair] : llvm::enumerate(
+           llvm::zip(getResults().getTypes(), mainFunc.getResultTypes()))) {
+    auto [resultType, funcResultType] = pair;
+    if (resultType != funcResultType) {
+      return emitOpError("D2M function return type ")
+             << idx << " mismatch: expected " << resultType << ", got "
+             << funcResultType;
+    }
+  }
```
azecevic000, line 4791: Should you also do this for input types of the function?

lib/Conversion/TTNNToTTIR/TTNNToTTIRPass.cpp
```diff
@@ -34,7 +34,24 @@ struct ConvertTTNNToTTIRPass
...
+    });
+  }
+
+  bool isInsideD2MSubgraph(Operation *op) {
+    func::FuncOp parentFunc = op->getParentOfType<func::FuncOp>();
+    return parentFunc && d2mSubgraphs.contains(parentFunc);
+  }
+
   void runOnOperation() final {
+    ModuleOp module = getOperation();
```
azecevic000: Nit: rename to `moduleOp`

lib/Dialect/TTNN/Pipelines/TTNNPipelines.cpp
```diff
@@ -285,7 +288,22 @@ void createTTIRToTTNNDevicePipeline(
       devicePm.addPass(createTTNNSetComputeKernelConfig(setConfigOptions));
     }
 
+    if (options.enableD2MFusing) {
+      if (!options.optimizerPassEnabled) {
+        llvm::llvm_unreachable_internal(
+            "D2M fusing pass only supported with optimizer enabled");
+      }
```
azecevic000, line 304: I'm not sure if this is the right approach from the user experience perspective. If the user wants `d2m-fusing` he shouldn't be concerned with the internal dependencies of the pass. Maybe logging a warning that the optimizer is not explicitly enabled and that the `d2m-fusing` will enable it as its dependency could enhance debugability while maintaining seamless UX.

lib/Dialect/TTNN/Transforms/TTNNCollaspeD2M.cpp
```diff
@@ -0,0 +1,97 @@
...
+    SmallVector<func::FuncOp> funcsToDelete;
+
+    moduleOp.walk([&](DispatchD2MOp dispatchOp) -> WalkResult {
+      func::FuncOp mainFunc = dispatchOp.getD2MMainFunc();
+      if (failed(inlineDispatchOp(moduleOp, dispatchOp))) {
+        signalPassFailure();
+        return WalkResult::interrupt();
+      }
+      funcsToDelete.push_back(mainFunc);
+      return WalkResult::advance();
```
azecevic000: Nit: you aren't using the return type of the lambda as a result of `walk` function, you can use a version with lambda that returns `void`.

lib/Dialect/TTNN/Transforms/TTNND2MFusing.cpp
```diff
@@ -0,0 +1,307 @@
...
+      for ([[maybe_unused]] Operation *op : group) {
+        TTMLIR_DEBUG(ttmlir::LogComponent::D2MFusion, "    {} at {}",
+                     op->getName(), op->getLoc());
+      }
+    }
+  }
+
+  // BFS backward from exit op and include producer only if all its consumers
+  // are in the fusion set. BFS order guarantees correctness since consumers are
+  // processed before their producers.
```
azecevic000: Isn't this DFS on (consumer->producer) graph?

lib/Dialect/TTNN/IR/TTNNOps.cpp
vtangTT: added verifier for inputs

lib/Conversion/TTNNToTTIR/TTNNToTTIRPass.cpp
vtangTT: fixed

lib/Dialect/TTNN/Pipelines/TTNNPipelines.cpp
vtangTT: will now `llvm::errs` message and then turn on optimizer as a dep

lib/Dialect/TTNN/Transforms/TTNNCollaspeD2M.cpp
vtangTT: fixed

lib/Dialect/TTNN/Transforms/TTNND2MFusing.cpp
vtangTT: yes.. my mistake, comments have been fixed

Issue #6748, opened 2026-01-26
ArgMax produces mismatched tensor (si32) and memref (ui32) types


bmalesevicTT:
Nightly test model `yoloworld` ran as:
```python
pytest tests/runner/test_models.py::test_all_models_torch[yoloworld/pytorch-small_640-single_device-inference] -svv
```
fails for `optimization_level=1` in `tt-mlir/lib/Dialect/TTNN/Analysis/ScalarDataTypeAnalysis.cpp` with:
```
assert(layoutAttr.getScalarElementType() == tensorType.getElementType() &&
       "Expected scalar element type in TTNNLayoutAttr to match tensor element type");
Type mismatch detected!
```
Problem origin:
```
Operation: ttnn.argmax
Location: loc("reduce.37")
Tensor element type: si32
Layout scalar element type: ui32
```
The core reason is existing logic for `ttnn.argmax` lowering (`TTIRToTTNN.cpp`), where the discrepancy between memref and tensor type is ignored, so tt-xla can still observe it as `si32`, while tt-metal expects `u32`. 

### Potential solutions
- The reason this comes up in optimizer is the assert checking this consistency, so one solution could be special casing for this to be ignored, as it should still work for this model.
- A better approach seems to be actually syncing them and pass assert naturally.

run simple tt-mlir repro below as 
```
ttmlir-opt --ttir-to-ttnn-backend-pipeline="system-desc-path=ttrt-artifacts/system_desc.ttsys optimization-level=1" repro.mlir --mlir-print-local-scope
```
```
module attributes {} {
  func.func @argmax_example(%arg0: tensor<3x3xi32> {ttcore.argument_type = #ttcore.argument_type<input>, ttcore.shard_status = #ttcore.shard_status<unsharded>, ttir.name = "args_0"}) -> (tensor<3xi32> {ttcore.shard_status = #ttcore.shard_status<unsharded>}) {
    %0 = "ttir.argmax"(%arg0) <{dim_arg = [1 : i32], keep_dim = false}> : (tensor<3x3xi32>) -> tensor<3xi32>
    return %0 : tensor<3xi32>
  }
}
```

PR #6749, opened 2026-01-26
[Optimizer] Fix `ttnn.argmax` result dtype mismatch

bmalesevicTT:
### Ticket
#6748 

RM pass changes encoding type based on backend (OpModel API) response, which is what introduces the dtype discrepancy in tensor (si32) and memref (ui32) types for `ttnn.argmax` op.
### Final implemented solution (edited after discussion)
Insert a conversion toLayout op (later converted to typecast) where dtypes mismatch - yes, this inserts runtime no-op typecasts, but it makes the conversion explicit rather than hiding the mismatch. This is the most transparent approach and aligns with what already happens at `optimizer_level=0`.
I confirmed that with optimizer_level=0, typecast ops already exist for these conversions. This approach maintains consistency: when RM pass detects a dtype mismatch, it inserts a toLayout op (later converted to typecast). This ensures the mismatch doesn't go unchecked.

Benefit vs. optimizer_level=0: RM pass removes other toLayout ops by propagating row-major layout to ops like ArgMaxOp, rather than the workaround's explicit casting.
Tradeoff: Typecast happens eagerly (immediately after dtype mismatch) rather than lazily (only when a consumer needs signed integer)

### Background
- tt-metal hardware: Uses unsigned integers (`ui32`) for indices
- TTNN op: Use signed integers (`si32/i32`) for arithmetic operations
- The conversion is functionally safe since indices are always positive (bitwise identical representation)

### Checklist
- [x] Added test (`test/ttmlir/Dialect/TTNN/optimizer/argmax_datatype_conversion.mlir`) 
      verifying ui32 conversion and typecast insertion
- [x] Fixes `yoloworld` model test with `optimization_level=1`

### DIFF
--- a/lib/Dialect/TTNN/Transforms/OptimizerPasses/RowMajorLayoutPropagation.cpp
+++ b/lib/Dialect/TTNN/Transforms/OptimizerPasses/RowMajorLayoutPropagation.cpp
@@ -197,6 +197,11 @@ private:
   // type. Inserts a toLayout op with TILE layout (required for on-device
   // typecast) to perform the dtype conversion. Returns true if conversion was
   // inserted, indicating RM propagation should stop at this point.
+  //
+  // TODO(bmalesevic, #6783): Once issue is addressed (on-device typecasting
+  // support for RM tensors), RM propagation should be allowed to continue after
+  // dtype conversion without forcing conversion to tile layout (which currently
+  // stops RM propagation).
   bool handleDtypeConversionIfNeeded(Operation *user, IRRewriter &rewriter,
                                      TTNNLayoutAttr rmOutputLayout,
                                      Type backendDataType,


rpavlovicTT:
I agree with @azecevicTT point, this is too hacky of a solution. His proposals make sense to try out.

@mtopalovicTT I don't think we need to turn on WA here, op validation & fallback pass should fix it if needed, but in this case we didn't reach that pass because of the assert in RM pass.

mtopalovicTT:
@rpavlovicTT RM pass didn't assert in this case, it was data type analysis. RM pass changed the encoding dtype without changing tensor element type which triggered assert down the line. RM pass should only change page layout right?

rpavlovicTT:
> @rpavlovicTT RM pass didn't assert in this case, it was data type analysis. RM pass changed the encoding dtype without changing tensor element type which triggered assert down the line. RM pass should only change page layout right?

Apologies, I didn't understand your previous comment. Yes, this is possibly a bug. We will investigate.

bmalesevicTT:
Yes, RM pass changes encoding type based on backend (OpModel API) response, which is what introduces the dtype discrepancy here.
I synced with @rpavlovicTT about the three potential approaches:
  1. Change layout's dtype to match tensor element type - risky and conceptually wrong, since we'd be overriding what the backend (tt-metal) says is correct, even if it passes runtime we want to avoid
  2. Relax dtype checks specifically for certain ops like argmax - also seems risky and generally we want to avoid such discrepancies in IR, not hide them
  3. Insert a conversion op where dtypes mismatch - yes, this inserts no-op typecasts, but it makes the conversion explicit rather than hiding the mismatch. This is the most transparent approach and aligns with what already happens at `optimizer_level=0`.

I confirmed that with optimizer_level=0, typecast ops already exist for these conversions. This approach maintains consistency: when RM pass detects a dtype mismatch, it inserts a toLayout op (later converted to typecast). This ensures the mismatch doesn't go unchecked.

Benefit vs. optimizer_level=0: RM pass removes other toLayout ops by propagating row-major layout to ops like ArgMaxOp, rather than the workaround's explicit casting.
Tradeoff: Typecast happens eagerly (immediately after dtype mismatch) rather than lazily (only when a consumer needs signed integer)

lib/Dialect/TTNN/Transforms/OptimizerPasses/RowMajorLayoutPropagation.cpp
```diff
@@ -193,6 +193,60 @@ class TTNNRowMajorLayoutPropagation
...
+    RankedTensorType backendResultType = RankedTensorType::get(
+        userResultType.getShape(), backendDataType, rmOutputLayout);
+    user->getResult(0).setType(backendResultType);
+
+    rewriter.setInsertionPointAfter(user);
+
+    // Create layout with TILE (required for on-device typecast) and the
+    // original tensor element type. TTNNDecomposeLayouts requires TILE
+    // layout for on-device dtype conversion; ROW_MAJOR would force a
+    // host round-trip (from_device → to_dtype → to_device).
```
rpavlovicTT, line 230: Hm I see now this forced round-trip, however https://github.com/tenstorrent/tt-metal/commit/544e13fd2afe806568f2da71f7f4ce31034efb9b added support for it. Let's keep your code change as is, but open an issue to remove forceful host typecasting and instead do it on device for RM tensors. Leave a comment above this function to let RM prop pass with typecast insertion once new issue is addressed.

PR #6751, opened 2026-01-26
Improving emitpy unit test coverage

sdjordjevicTT:
### Ticket
N/A

### Problem description
Currently TTNN EmitPy pipeline lacks 100% of unit test coverage with the builder.

### What's changed
Added builder tests for TTNN EmitPy for a couple of ops:
- Pooling ops (MaxPool, AvgPool, GlobalAvgPool, MaxPoolWithIndicies)
- Normalization ops (BatchNorm, Softmax, RMSNorm)
- Matmul ops (Matmul, Linear)

Fixed a couple of not-supported ops in EmitPy and EmitC.

### Checklist


runtime/lib/ttnn/operations/pool/pool2d.cpp
```diff
@@ -102,7 +102,8 @@ void runMaxPool2dOp(
   std::array<uint32_t, 2> kernelSize, stride, dilation;
   std::copy_n(op->kernel_size()->begin(), 2, kernelSize.begin());
   std::copy_n(op->stride()->begin(), 2, stride.begin());
-  std::copy_n(op->dilation()->begin(), 2, dilation.begin());
+  std::copy_n(op->extra_params_as_MaxPool2dExtraParams()->dilation()->begin(),
+              2, dilation.begin());
```
jameszianxuTT: Nit: The changes to poolOp definition / removing dilation as an argument seem unnecessary. It is a fair point that dilation doesn't make sense as an argument to avg pooling, and it's not forwarded in runtime. 

I guess there are overall two options to deal with representation of dilation in IR as it doesn't pertain to both MaxPool2D and AvgPool2D:
- Default expose attr, don't pass in runtime to AvgPool and check in validator that dilation isn't set
- Default hide attr, need to manually set + parse out of extra_params for MaxPool2D, and any other ops that may accept a dilation

I'm not sure if other types of pooling would be exposed in the future, but to me it seems like `dilation` is an inherent property of any pooling op, and while it doesn't make sense to do so for an avg pool, I'm not sure if it's our role to disallow this in the compiler by dropping it from the td. If the goal is to prevent a user from performing this avg pool with dilation assuming it's a mistake, then maybe it would be safer to assert this or check in op validation. 

This is just a suggestion, I don't have strong feelings about this either way though, so feel free to disregard.

include/ttmlir/Dialect/TTIR/Utils/VerificationUtils.h
```diff
@@ -243,6 +244,37 @@ OutputTensorDims getPool2dOutputDims(PoolOp *op) {
...
+getPoolingDilation(Pool2dOp op) {
+  // Handle both pointer and non-pointer types
+  using OpType = std::remove_pointer_t<Pool2dOp>;
+
+  // For AvgPool2d, dilation is not applicable and defaults to (1, 1)
+  if constexpr (std::is_same_v<OpType, mlir::tt::ttir::AvgPool2dOp>) {
+    return std::make_pair(1, 1);
+  } else {
+    // Check if op is a pointer or not
+    if constexpr (std::is_pointer_v<Pool2dOp>) {
```
mtopalovicTT: Can you drop this part and just allow ref/value types?

include/ttmlir/Dialect/TTIR/Utils/VerificationUtils.h
```diff
@@ -243,6 +244,37 @@ OutputTensorDims getPool2dOutputDims(PoolOp *op) {
...
 }
 
+template <typename Pool2dOp>
+static llvm::Expected<std::pair<int32_t, int32_t>>
+getPoolingDilation(Pool2dOp op) {
+  // Handle both pointer and non-pointer types
+  using OpType = std::remove_pointer_t<Pool2dOp>;
+
+  // For AvgPool2d, dilation is not applicable and defaults to (1, 1)
+  if constexpr (std::is_same_v<OpType, mlir::tt::ttir::AvgPool2dOp>) {
```
mtopalovicTT:
```cpp
if constexpr (std::is_same_v< Pool2dOp, mlir::tt::ttir::MaxPool2dOp>) {
} else if constexpr (std::is_same_v<Pool2dOp, mlir::tt::ttir::AvgPool2dOp>) {
} else {
     static_assert(....)
}
```

lib/Conversion/TTIRToTTNN/TTIRToTTNN.cpp
```diff
@@ -1703,18 +1703,19 @@ class Pooling2dOpConversionPattern : public OpConversionPattern<TTIROpTy> {
...
-        ttmlir::utils::getPairOfInteger<int32_t>(adaptor.getDilation());
-    assert(dilationPairOrError && "Expected valid dilation attribute");
-    DenseI32ArrayAttr dilationAttr = rewriter.getDenseI32ArrayAttr(
-        {dilationPairOrError->first, dilationPairOrError->second});
+    DenseI32ArrayAttr dilationAttr;
 
     // TTNN only supports lowering of AvgPool2dOp with dilation of (1, 1).
-    if constexpr (std::is_same_v<TTIROpTy, ttir::AvgPool2dOp>) {
-      if (dilationPairOrError->first != 1 || dilationPairOrError->second != 1) {
+    if constexpr (!std::is_same_v<TTIROpTy, ttir::AvgPool2dOp>) {
```
mtopalovicTT: do we validate dilation in ttnnops.cpp for max and avg pooling?

runtime/lib/ttnn/operations/pool/pool2d.cpp
sdjordjevicTT: Agree, I am not sure either. For the sake of this PR, I will revert the change for the avg_pool API. It would be smarter to handle such things in a separate PR while we also agree on the public API.

sdjordjevicTT:
> FYI we are migrating the builder ops and mappings to have a different structure (cutting out `op_proxy`), it makes life easier for us if new ops are added to the builder ecosystem in the new format in the future but isn't a dealbreaker

Hey @jgrimTT, I rewrote the newly added global_avg_pool2d op in the TTIR builder. Can you please take a look at it?

Could you also please guide me on how to verify the parser and split functions that I added for this op?

Issue #6756, opened 2026-01-27
Issues with shardy and gather in a new transformers version


umalesTT:
### Describe the bug

Creation of attention mask is changed in new transformers version, and on batch-sharded inputs StableHLO compilation fails on ```stablehlo.gather``` on ```InsertExplicitReshardsPass```.
So it is a blocker on all data-parallel training tasks.
Below is a minimal repro.

```
// REQUIRES: stablehlo
// RUN: ttmlir-opt --stablehlo-pipeline="mesh-shape=2,4" %s 2>&1 | FileCheck %s
//
// This test reproduces the bug where gather with collapsed_slice_dims=[0,1]
// and operand sharded on the indexed dimension causes:
//   error: 'sdy.all_reduce' op reduction axis "_axis_0" overlaps with operand dimension sharding
module {
  sdy.mesh @mesh = <["_axis_0"=2, "_axis_1"=4]>
  // Minimal repro: gather with collapsed_slice_dims=[0,1] on sharded operand
  func.func @gather_allreduce_overlap_bug(
      %arg0: tensor<32x64xi1> {sdy.sharding = #sdy.sharding<@mesh, [{"_axis_0"}, {}]>},
      %arg1: tensor<32x64x2xi64> {sdy.sharding = #sdy.sharding<@mesh, [{"_axis_0"}, {}, {}]>}
  ) -> tensor<32x64xi1> {
    %0 = "stablehlo.gather"(%arg0, %arg1) <{
      dimension_numbers = #stablehlo.gather<
        collapsed_slice_dims = [0, 1],
        start_index_map = [0, 1],
        index_vector_dim = 2
      >,
      slice_sizes = array<i64: 1, 1>
    }> : (tensor<32x64xi1>, tensor<32x64x2xi64>) -> tensor<32x64xi1>
    return %0 : tensor<32x64xi1>
  }
}
// CHECK: error: 'sdy.all_reduce' op reduction axis "_axis_0" overlaps with operand dimension sharding


```

In [shardy implementation](https://github.com/openxla/shardy/blob/23359578eec0292a45958f560b19ea574d98e669/shardy/dialect/sdy/transforms/propagation/op_sharding_rule_registry.cc#L192) for some cases they add CCLs, in this issue all-reduce, and for us it fails on this concrete case as all-reduce is done across sharded dimension.
[...]

PR #6769, opened 2026-01-27
Fix gather sharding rule to use replication for collapsed_slice_dim

sshonTT:
### Ticket
[Link to Github Issue](https://github.com/tenstorrent/tt-mlir/issues/6756)

### Problem description
When operand is sharded along collapsed_slice_dims, the default Shardy gather rule using kReduction causes InsertExplicitReshardsPass to insert sdy.all_reduce after the gather operation. However, this leads to an error:

 'sdy.all_reduce' op reduction axis "_axis_0" overlaps with operand dimension sharding or replicated axes

### What's changed
This happens because the propagated output sharding conflicts with the reduction axis of the all_reduce operation.

Change kReduction to kNeedReplication for gather's collapsed_slice_dims. This forces an all-gather on the operand before the gather operation instead of an all-reduce after, avoiding the axis overlap issue.

- kReduction: keep operand sharded -> gather -> all-reduce (causes error)
- kNeedReplication: all-gather -> gather (works correctly)

### Checklist

### DIFF
--- a/env/patches/shardy.patch
+++ b/env/patches/shardy.patch
@@ -648,9 +648,25 @@ index a73917b..78cc827 100644
                       llvm::SmallDenseSet<Value>& seenValues) {
    EdgeNode target = edge.target;
 diff --git a/shardy/dialect/sdy/transforms/propagation/op_sharding_rule_registry.cc b/shardy/dialect/sdy/transforms/propagation/op_sharding_rule_registry.cc
-index ab93067..283d6ad 100644
+index ab93067..b181797 100644
 --- a/shardy/dialect/sdy/transforms/propagation/op_sharding_rule_registry.cc
 +++ b/shardy/dialect/sdy/transforms/propagation/op_sharding_rule_registry.cc
+@@ -187,12 +187,12 @@ void addGatherScatterFactors(
+
+   // We add factors for all collapsed slice dimensions.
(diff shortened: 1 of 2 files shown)


env/patches/shardy.patch
```diff
@@ -651,7 +651,24 @@ diff --git a/shardy/dialect/sdy/transforms/propagation/op_sharding_rule_registry
...
+@@ -188,10 +188,12 @@ void addGatherScatterFactors(
+
+   // We add factors for all collapsed slice dimensions.
+   for (int64_t collapsedSliceDim : collapsedSliceDims) {
+-    // To keep the operand dim sharded for gather, we need an all-reduce on the
+-    // result.
++    // For gather: use kNeedReplication to all-gather operand before gather,
++    // instead of kReduction which would insert all-reduce after gather.
++    // kReduction can cause "reduction axis overlaps with operand sharding" errors.
++    // For scatter: use kPassThrough as before.
```
mmanzoorTT:
```suggestion
+    // For scatter: use kPassThrough 
```
and add the reason as you did for `gather`.
`as before` doesn't provide any meaningful context.

sshonTT:
I've opened an issue [upstream](https://github.com/openxla/shardy/issues/968) regarding this, as I believe it should be addressed at the upstream level.

PR #6770, opened 2026-01-27
[d2m] support for 1d tensors

### LINKED ISSUE #6659: [d2m] support 1d tensors
```mlir
  func.func @component_3_ttir_typecast_72(%arg0: tensor<1xi64>) -> tensor<1xf32> {
    %0 = "ttir.typecast"(%arg0) <{conservative_folding = false}> : (tensor<1xi64>) -> tensor<1xf32>
    return %0 : tensor<1xf32>                                                                                                                                                                                         }
```

Fails with:
```
TTCoreOpsTypes.cpp:995: static llvm::SmallVector<int64_t> mlir::tt::ttcore::MetalLayoutAttr::computeTileAlignments(ArrayRef<int64_t>, ArrayRef<int64_
t>): Assertion `logicalRank >= 2' failed.
```

dlokeTT:
### Ticket
Closes #6659 

### Problem description
Currently there is no support for 1d tensors in the d2m pipeline and it would error out. We need to be able to support 1d tensors in general thru d2m and especially since they are needed for supporting models such as gpt_oss and qwen
### What's changed
- Added a pass, `d2m-rank-normalization` at the ttir level that rewrites ir that has < 2d tensors to use 2d tensors by prepending 1s.
- Added 1d tests for unary, binary and ternary ops
- Added tests for reshaping ops from 1d -> nd and vice versa 
- Added lit tests for `rankNormalization` pass
### Checklist


lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
```diff
@@ -30,6 +30,75 @@
 namespace mlir::tt {
 
 namespace {
+
+/// Virtualize a shape to at least 2D by prepending a dimension of 1.
+/// Device operations require at least 2D tensors, so 1D tensors are
+/// expanded to [1, N] internally. This is a zero-cost type reinterpretation.
+static SmallVector<int64_t> virtualizeTo2D(ArrayRef<int64_t> shape) {
```
vwellsTT: nit: I think this naming is confusing, would specify 1DTo2D or something

test/python/golden/test_metal_tms.py
```diff
@@ -202,71 +193,21 @@ def permute(
...
+    # LLAMA 3.2 3B TESTS
     ((18, 128), (1, 18, 128)),
     ((18, 128), (1, 1, 18, 128)),
-    pytest.param(((18,), (18, 1)), marks=ONE_D_SKIP),
-    pytest.param(((18,), (1, 1, 18)), marks=ONE_D_SKIP),
-    pytest.param(((1, 1024, 3072), (1024, 3072)), marks=SLOW_COMPILE_SKIP),
-    pytest.param(((1, 128256, 3072), (128256, 3072)), marks=SLOW_COMPILE_SKIP),
     ((1, 18, 128), (18, 128)),
     ((1, 18, 128), (1, 1, 18, 128)),
-    pytest.param(((1, 1, 18), (18,)), marks=ONE_D_SKIP),
```
vwellsTT, line 240: why are so many of these being removed?

include/ttmlir/Target/Utils/FlatbufferObjectCache.h
```diff
@@ -8,11 +8,15 @@
...
 #include "llvm/ADT/DenseMap.h"
 
+#include <optional>
+
 namespace mlir::tt {
 
 struct FlatbufferObjectCache {
   ::flatbuffers::FlatBufferBuilder *fbb;
   DenseMap<const void *, ::flatbuffers::uoffset_t> objectMap;
+  // Map to track global_id assigned to each Value (for buffer aliasing)
```
vwellsTT: nit: fix comments pls

include/ttmlir/Dialect/D2M/Transforms/Passes.td
```diff
@@ -861,5 +861,78 @@ def D2MDecomposeComplexPermute: Pass<"d2m-decompose-complex-permute", "::mlir::M
   }];
 }
 
+def D2MPromote1DTo2D: Pass<"d2m-promote-1d-to-2d", "::mlir::ModuleOp"> {
```
vwellsTT: super nit: I would personally call this Expand1DTo2D over Promote--to me, promote implies more of a type change (e.g. scalar to tensor).  But I think it's pretty subjective/debatable

lib/Target/TTMetal/TTMetalToFlatbuffer.cpp
```diff
@@ -1026,6 +1029,50 @@ std::shared_ptr<void> translateTTMetalToFlatbuffer(
         // Unqualified walk will visit the root op itself last, we should
         // ignore this.
         return;
+      } else if (auto expandOp = dyn_cast_if_present<memref::ExpandShapeOp>(op);
+                 expandOp) {
+        // Shape manipulation ops are compile-time only. Create a new BufferRef
+        // with the expanded shape but sharing the same global_id as the source
+        // (they point to the same memory). This is needed for 1D tensor support
+        // where we expand 1D to 2D before copying to the aligned bounce buffer.
```
vwellsTT: and it's the case that this op cannot change data layout in memory, only expand rank trivially?

lib/Dialect/D2M/Transforms/RankNormalization.cpp
```diff
@@ -0,0 +1,263 @@
...
+      // Track all values that have been promoted to 2D (for downstream
+      // propagation)
+      DenseSet<Value> promotedValues;
+
+      for (unsigned idx : input1DIndices) {
+        BlockArgument arg = entryBlock.getArgument(idx);
+
+        // Check if all uses of this 1D argument are reshape ops.
+        // If so, skip inserting a promotion reshape since the existing
+        // reshapes already handle the dimension change.
```
vwellsTT: this makes me wonder, what if there's a downcasting reshape in the input? i.e.
```
%n = ttir.reshape (some 2x32 -> 64)
%m = ttir.exp(%n)
```
what would happen?

include/ttmlir/Dialect/D2M/Transforms/Passes.td
nsmithtt: +1 or RankNormalization

nsmithtt (commented):
Took a quick look, do we need the memref aliasing stuff / complicated host view? I expected to just flat out rewrite the ranks destructively.

So at the func level if you have:
```mlir
func main(%a: tensor<f32>) ->  tensor<f32> {
  %0 = exp(%a) :  tensor<f32> ->  tensor<f32>
  return %0
}
```

We'd just replace it with:
```mlir
func main(%a: tensor<1x1xf32>) {
  %0 = exp(%a) :  tensor<1x1xf32> ->  tensor<1x1xf32>
  return %0
}
```

We can do this because adding 1's to the outer dims doesn't change the layout in any way, the scalar would be packed the same way no matter what.

If my proposal works then it can just be a pass that runs in isolation and no existing code has to change.

nsmithtt:
Also the pass can be written in kind of a dialect conversion style, you need a special func op rewriter that changes the func signature, but then you can just have a generic rewriter, pseudo code:

```mlir
matchAndRewrite(Operation* op, Adaptor adaptor, Rewriter rewriter) {
  for (auto opOperand : adaptor->getOpOperands()) {
    if (getRank(opOperand->getType()) >= 2) {
      continue;
    }

    auto *cast2d = rewriter.create<UnrealizedConversionCast>(as2D(opOperand->getType(), operand);
    rewriter.modifyOpInPlace(op, [&]() {
      opOperand.set(cast2d.getResult());
    });
  }

  for (auto result : adaptor->getResults()) {
    if (getRank(result->getType()) >= 2) {
      continue;
    }

    auto *cast2d = rewriter.create<UnrealizedConversionCast>(as2D(result->getType(), result);
    rewriter.replaceAllUsesWith(result, cast2d);
  }

  return success();
}
```

All of the unrealized conversion casts should cancel each other out after the conversion pass has been fully evaluated.

test/python/golden/test_metal_tms.py
dlokeTT: These were added as part of tests for llama 3.2 3b that were added last year but now we have some new models to focus on and we plan to test their shapes using builder load/parse functionality

lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
```diff
@@ -30,6 +30,75 @@
...
+  auto virtOutput = virtualizeTo2D(outputShape);
+  return virtInput == virtOutput;
+}
+
+/// Compute logical map for reshape: linearize output coords, delinearize to
+/// input coords. This handles rank changes (e.g., 2D -> 3D).
+/// Returns a map from output logical coords to input logical coords.
+/// This generalized version accepts explicit shapes, allowing callers to pass
+/// virtualized shapes for 1D tensor handling.
+static AffineMap reshapeLogicalMapFromShapes(MLIRContext *ctx,
```
bgrady-tt: This might do the same thing as `calculateReblockMap`, which also does ND -> flat -> ND thing to do reblocking?

dlokeTT:
> Took a quick look, do we need the memref aliasing stuff / complicated host view? I expected to just flat out rewrite the ranks destructively.
> 
> So at the func level if you have:
> 
> ```mlir
> func main(%a: tensor<f32>) ->  tensor<f32> {
>   %0 = exp(%a) :  tensor<f32> ->  tensor<f32>
>   return %0
> }
> ```
> 
> We'd just replace it with:
> 
> ```mlir
> func main(%a: tensor<1x1xf32>) {
>   %0 = exp(%a) :  tensor<1x1xf32> ->  tensor<1x1xf32>
>   return %0
> }
> ```
> 
> We can do this because adding 1's to the outer dims doesn't change the layout in any way, the scalar would be packed the same way no matter what.
> 
> If my proposal works then it can just be a pass that runs in isolation and no existing code has to change.

I initially thought if we changed the input arg types, then it would do a reshape on host or cause some issues but I can play around with this to see if it works

nsmithtt:
> I initially thought if we changed the input arg types, then it would do a reshape on host or cause some issues but I can play around with this to see if it works

Let me know if it works, my thinking is that adding 1's to the outer dimension is a no-op in terms of data layout, so it should be fine to destructively rewrite it.

dlokeTT:
> > I initially thought if we changed the input arg types, then it would do a reshape on host or cause some issues but I can play around with this to see if it works
> 
> Let me know if it works, my thinking is that adding 1's to the outer dimension is a no-op in terms of data layout, so it should be fine to destructively rewrite it.

Yeah this works. Updated pr to follow this strategy

lib/Dialect/D2M/Transforms/RankNormalization.cpp
dlokeTT: Using new structure, the result would be rewritten as `1x64` and then both input and output of `ttir.exp` would be `1x64`

lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
dlokeTT: no longer needed as approach to this changed

lib/Dialect/D2M/Transforms/RankNormalization.cpp
```diff
@@ -0,0 +1,269 @@
...
+namespace mlir::tt::d2m {
+#define GEN_PASS_DEF_D2MRANKNORMALIZATION
+#include "ttmlir/Dialect/D2M/Transforms/Passes.h.inc"
+
+namespace {
+
+static constexpr int64_t minRank = 2;
+
+/// Promotes a shape array to minRank by prepending 1s.
+static SmallVector<int32_t> promoteShape(ArrayRef<int32_t> shape) {
```
vwellsTT: nit: I would also rename this + below to "expandShape" or normalize or similar (personally prefer expand, normalize seems less precise imo)

lib/Dialect/D2M/Transforms/RankNormalization.cpp
```diff
@@ -0,0 +1,269 @@
...
+  using OpConversionPattern<func::FuncOp>::OpConversionPattern;
+
+  LogicalResult
+  matchAndRewrite(func::FuncOp funcOp, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    // Non-external functions are handled by the standard function conversion.
+    if (!funcOp.isExternal()) {
+      return failure();
+    }
+    if (!ttmlir::utils::isForwardCPUDeclarationFunc(funcOp)) {
```
vwellsTT: is this needed? Doesn't seem related to me (indeed, seems inverted to me?)

test/python/golden/test_metal_tms.py
```diff
@@ -74,8 +69,6 @@
         [(1, 18, 8, 128), [0, 2, 1, 3]],
         [(1, 128, 24, 128), [0, 2, 1, 3]],
         [(1, 128, 8, 128), [0, 2, 1, 3]],
-        # 4d outer permutes (tilized)
-        [(1, 18, 8, 128), [1, 0, 2, 3]],
```
vwellsTT, line 78: why are some many of these removed?

lib/Dialect/D2M/Transforms/RankNormalization.cpp
```diff
@@ -0,0 +1,269 @@
...
+    });
+    addSourceMaterialization(materializeCast);
+    addTargetMaterialization(materializeCast);
+  }
+
+private:
+  static Value materializeCast(OpBuilder &builder, Type type, ValueRange inputs,
+                               Location loc) {
+    assert(inputs.size() == 1 && "Expected single input.");
+    return builder.create<UnrealizedConversionCastOp>(loc, type, inputs.front())
```
vwellsTT, line 75: not sure I follow how this gets squashed?

test/ttmlir/Dialect/D2M/Transforms/rank_normalization.mlir
```diff
@@ -0,0 +1,280 @@
...
+// =============================================================================
+
+// CHECK-LABEL: func.func @all_2d_no_promotion
+// CHECK-SAME: (%arg0: tensor<8x16xf32>) -> tensor<8x16xf32>
+// CHECK: %0 = "ttir.abs"(%arg0) : (tensor<8x16xf32>) -> tensor<8x16xf32>
+// CHECK: return %0 : tensor<8x16xf32>
+func.func @all_2d_no_promotion(%arg0: tensor<8x16xf32>) -> tensor<8x16xf32> {
+  %0 = "ttir.abs"(%arg0) : (tensor<8x16xf32>) -> tensor<8x16xf32>
+  return %0 : tensor<8x16xf32>
+}
```
vwellsTT, line 280: might be good to have tests w/ implicit bcast (e.g. add a 1D tensor to a 3D tensor type thing)?

test/python/golden/test_metal_tms.py
dlokeTT: Permute test was accidentally removed. The rest are old remnants from using llama 3.2 3b tests but now we have new models to go off of and a new structure for testing those thru mlir snippets.

lib/Dialect/D2M/Transforms/RankNormalization.cpp
dlokeTT: These checks are not needed anymore as I set the benefit to lower so it runs after standard functions are handled and always deals with external funcs

lib/Dialect/D2M/Transforms/RankNormalization.cpp
dlokeTT: I believe this is done by MLIR in dialect conversion. When I `applyFullConversion` it automatically tries to remove all `UnrealizedConversionCastOp` pairs that are inverses of each other.

PR #6773, opened 2026-01-27
Add support for non-square grids

sohaibnadeemTT:
### Ticket
Link to Github Issue: https://github.com/tenstorrent/tt-mlir/issues/6776, https://github.com/tenstorrent/tt-mlir/issues/6619

### Problem description

### What's changed

### Checklist

### DIFF
--- a/lib/Dialect/D2M/Transforms/GridSelection.cpp
+++ b/lib/Dialect/D2M/Transforms/GridSelection.cpp
@@ -474,18 +474,18 @@ normalizeOperandGridsForGeneric(
   // dimensions. For example, in a matmul, the two inputs share the reduction
   // dimension. If their independently chosen optimal grids differ along that
   // dimension, promote the grid factor for that *dimension only* to the
-  // maximum across all inputs that share it.
+  // maximum across all operands that share it.
   llvm::SmallVector<llvm::SmallVector<int64_t>> normalizedOperandGrids(
       optimalOperandGrids.begin(), optimalOperandGrids.end());
 
-  uint64_t numInputs = genericOp.getNumDpsInputs();
+  uint64_t numOperands = genericOp->getNumOperands();
   // Map: loopDim -> list of (operandIndex, operandDimIdx) pairs that reference
   // this loop dimension in their indexing maps.
   llvm::DenseMap<int64_t, llvm::SmallVector<std::pair<uint64_t, uint64_t>>>
-      dimToInputOperandDims;
+      dimToOperandDims;
 
   auto indexingMaps = genericOp.getIndexingMapsValue();
-  for (uint64_t operandIndex = 0; operandIndex < numInputs; ++operandIndex) {
+  for (uint64_t operandIndex = 0; operandIndex < numOperands; ++operandIndex) {
     AffineMap operandIndexingMap = indexingMaps[operandIndex];
     auto results = operandIndexingMap.getResults();
     for (auto [operandDimIdx, expr] : llvm::enumerate(results)) {
@@ -494,15 +494,15 @@ normalizeOperandGridsForGeneric(
         continue;
       }
       int64_t loopDim = dimExpr.getPosition();
-      dimToInputOperandDims[loopDim].push_back(
+      dimToOperandDims[loopDim].push_back(
           std::make_pair(operandIndex, static_cast<uint64_t>(operandDimIdx)));
     }
   }
 
-  // For each loop dimension that is used by multiple inputs, promote the grid
+  // For each loop dimension that is used by multiple operands, promote the grid
   // size associated with that loop dimension to the minimum across those
-  // inputs.
-  for (auto &it : dimToInputOperandDims) {
+  // operands.
+  for (auto &it : dimToOperandDims) {
     auto &entries = it.second;
     if (entries.size() < 2) {
       continue;
@@ -522,48 +522,6 @@ normalizeOperandGridsForGeneric(
       TT_assertv(operandDimIdx < normalizedOperandGrids[operandIndex].size(),
                  "operand dim index out of bounds on adjusted operand grids");
       normalizedOperandGrids[operandIndex][operandDimIdx] = minFactor;
-    }
-  }
-
-  // Compute grid dim constraints implied by the generic's outputs. These
-  // constraints describe which loop dimensions must agree across operands.
-  auto outputIndexingMap =
-      genericOp.getIndexingMapsValue()[genericOp.getOutputs()
-                                           .getBeginOperandIndex()];
-  auto outputShape =
-      optimalOperandGrids[genericOp.getOutputs().getBeginOperandIndex()];
-  std::optional<SmallVector<int64_t>> outputConstraints =
-      utils::computeDimConstraints(
-          llvm::ArrayRef<AffineMap>(outputIndexingMap),
-          llvm::ArrayRef<SmallVector<int64_t>>(outputShape));
-
-  // Ensure that input operand grid shapes respect any constraints implied by
-  // the outputs' grids. If multiple operands participate in the same loop
-  // dimension, the corresponding grid extents must agree.
-  if (outputConstraints) {
-    for (auto [operandIndex, operand] :
-         llvm::enumerate(genericOp->getOpOperands())) {
-      if (genericOp.isDpsInit(&operand)) {
-        continue;
-      }
-
-      AffineMap indexingMap = genericOp.getIndexingMap(operandIndex);
-      auto results = indexingMap.getResults();
-      TT_assertv(results.size() == normalizedOperandGrids[operandIndex].size(),
-                 "indexing map results size does not match normalized operand "
-                 "grids size");
-
-      for (auto [resultIdx, expr] : llvm::enumerate(results)) {
-        auto dimExpr = mlir::dyn_cast<AffineDimExpr>(expr);
-        if (!dimExpr) {
-          continue;
-        }
-        int64_t dimPos = dimExpr.getPosition();
-        int64_t constraint = (*outputConstraints)[dimPos];
-        if (constraint != 0) {
-          normalizedOperandGrids[operandIndex][resultIdx] = constraint;
-        }
-      }
     }
   }
 
--- a/test/python/golden/test_metal_matmul.py
+++ b/test/python/golden/test_metal_matmul.py
@@ -202,14 +202,18 @@ def test_matmul_ttnn_shapes_double_buffered(
         # width sharded in0, block sharded in1
         pytest.param(
             (32, 4096, 2048),
-            marks=pytest.mark.skip(reason="Constraining output grids not supported"),
+            marks=pytest.mark.skip(
+                reason="Not enough L1 space after constraining grids"
+            ),
         ),
         (32768, 32, 32),  # height sharded in0, block sharded in1
         (32, 32, 32768),  # block sharded in0, width sharded in1
         # block sharded in0, height sharded in1
         pytest.param(
             (2048, 4096, 32),
-            marks=pytest.mark.skip(reason="Constraining output grids not supported"),
+            marks=pytest.mark.skip(
+                reason="Not enough L1 space after constraining grids"
+            ),
         ),
     ],
     ids=shape_str,


lib/Dialect/D2M/Transforms/GridSelection.cpp
```diff
@@ -551,15 +546,25 @@ analyzeOperandsAndComputeGrids(d2m::GenericOp genericOp,
...
+        computeOptimalGrid(operandType, physShape, targetGridShape);
 
     optimalOperandGrids.push_back(optimalGrid);
+    isOptimalOperandGridVirtual.push_back(isVirtualGrid);
+  }
+
+  // Normalize the operand grids for the generic operation - see the comment on
+  // this function for details.
+  optimalOperandGrids =
+      normalizeOperandGridsForGeneric(genericOp, optimalOperandGrids);
```
sohaibnadeemTT, line 571: Change no. 2: grids constraints must be moved earlier and when we update view_layout/stream_layout/empty ops, the operand layouts are aligned according to the constrained grids (see layoutWithOptimalGrid, we pass optimalGrid instead). This ensures that the operands for matmul get aligned equally to respect dim constraints. To highlight why this problem occurs when using non-square grids, consider the following example on BH which has a 10x13 compute grid:

Before computePhysicalShape:
Input 0 layout: 1x1x20x26x!ttcore.tile<32x32, bf16>
Input 1 layout: 1x1x26x26x!ttcore.tile<32x32, bf16>

computePhysicalShape aligns both inputs to a 10x13 grid (since they exceed grid threshold) so when view_layout/stream_layout/empty ops get updated, we get the following new layouts:
Input 0 layout: 10x13x2x2x!ttcore.tile<32x32, bf16>
Input 1 layout: 10x13x3x2x!ttcore.tile<32x32, bf16> (gets aligned)

which is problematic since now the inputs dims don't even align to be able to do matmul (input 0 is 20x26 while input 1 is 30x26). This is why we need to apply the constraints early on right after computing the optimal grids, and then use the optimal (constrained) grids to do alignment and get the new layouts.

lib/Dialect/D2M/Transforms/GridSelection.cpp
```diff
@@ -253,39 +252,37 @@ shouldImplementAsVirtualGrid(RankedTensorType tensorType,
...
     ttcore::MetalLayoutAttr oldLayout, ArrayRef<int64_t> targetGridShape,
-    ArrayRef<int64_t> targetSquareGridShape, ArrayRef<int64_t> optimalGrid,
-    bool isVirtualGrid, OpBuilder &builder) {
+    ArrayRef<int64_t> optimalGrid, bool isVirtualGrid, OpBuilder &builder) {
   auto collapsedIntervals = oldLayout.getCollapsedIntervals();
 
   llvm::SmallVector<int64_t> newDimAlignments =
       ttcore::MetalLayoutAttr::computeGridAwareDimAlignments(
-          oldLayout.getLogicalShape(), targetSquareGridShape,
+          oldLayout.getLogicalShape(), optimalGrid,
```
sohaibnadeemTT, line 251: Related to comment above on change no. 2, we need to use optimal (constrained) grids to do out alignment and get the new layout.

lib/Dialect/D2M/Transforms/GridSelection.cpp
```diff
@@ -454,28 +449,28 @@ normalizeOperandGridsForGeneric(
...
-    int64_t maxFactor = 0;
+    int64_t minFactor = std::numeric_limits<int64_t>::max();
     TT_assertv(
         entries.size() <= normalizedOperandGrids.size(),
         "adjusted operand grids size does not match dim-operand mapping size");
     for (auto [operandIndex, operandDimIdx] : entries) {
       TT_assertv(operandDimIdx < normalizedOperandGrids[operandIndex].size(),
                  "operand dim index out of bounds on adjusted operand grids");
-      maxFactor = std::max(maxFactor,
+      minFactor = std::min(minFactor,
```
sohaibnadeemTT, line 518: Change no. 1: Use min instead of max here since we need to ensure we respect grid constraints (e.g. on BH, if one operand has grid width 13 and the other has grid height 10, then we have to use 10 since BH has only 10 rows)

lib/Dialect/D2M/Transforms/GridSelection.cpp
```diff
@@ -219,52 +219,50 @@ shouldImplementAsVirtualGrid(RankedTensorType tensorType,
...
-    bool isVirtualGrid, OpBuilder &builder) {
+    ArrayRef<int64_t> optimalGrid, bool isVirtualGrid, OpBuilder &builder) {
   auto collapsedIntervals = oldLayout.getCollapsedIntervals();
 
   llvm::SmallVector<int64_t> newDimAlignments =
       ttcore::MetalLayoutAttr::computeGridAwareDimAlignments(
-          oldLayout.getLogicalShape(), targetSquareGridShape,
-          oldLayout.getNormalizedIntervals());
+          oldLayout.getLogicalShape(), optimalGrid,
+          oldLayout.getNormalizedIntervals(), /*forceGridAlignment=*/true);
```
vmilosevic, line 252: :warning: **clang\-diagnostic\-error** :warning:
too many arguments to function call, expected 3, have 4

lib/Dialect/D2M/Transforms/GridSelection.cpp
```diff
@@ -697,19 +697,20 @@ updateStreamLayoutOps(ArrayRef<StreamLayoutUpdateInfo> streamLayoutsToUpdate,
 
     llvm::SmallVector<int64_t> storageDimAlignments =
         ttcore::MetalLayoutAttr::computeGridAwareDimAlignments(
-            storageLayout.getLogicalShape(), targetSquareGridShape,
-            storageLayout.getNormalizedIntervals());
+            storageLayout.getLogicalShape(), optimalGrid,
+            storageLayout.getNormalizedIntervals(),
+            /*forceGridAlignment=*/true);
```
vmilosevic, line 660: :warning: **clang\-diagnostic\-error** :warning:
too many arguments to function call, expected 3, have 4

lib/Dialect/TTCore/IR/TTCoreOpsTypes.cpp
```diff
@@ -1014,7 +1014,7 @@ MetalLayoutAttr::computeTileAlignments(ArrayRef<int64_t> logicalShape,
 // more predictable NoC traffic patterns.
 llvm::SmallVector<int64_t> MetalLayoutAttr::computeGridAwareDimAlignments(
```
vmilosevic, line 1015: :warning: **clang\-diagnostic\-error** :warning:
out\-of\-line definition of `` computeGridAwareDimAlignments `` does not match any declaration in `` mlir::tt::ttcore::MetalLayoutAttr ``

sohaibnadeemTT:
@vwellsTT I added some tests under `test/ttmlir/Dialect/D2M/Transforms/grid_selection_bh.mlir` so we can actually check the operand grids selected, let me know if I should add more. I also added tests under `test/ttmlir/Silicon/TTMetal/p150` (same as the n150 ones) but I don't know if we're getting any extra coverage with those, I can remove them.

lib/Dialect/TTCore/IR/TTCoreOpsTypes.cpp
```diff
@@ -1023,7 +1023,7 @@ llvm::SmallVector<int64_t> MetalLayoutAttr::computeGridAwareDimAlignments(
   const int64_t tensorGridRank = normalizedIntervals.size() / 2;
 
   assert(logicalRank >= 2);
-  assert(deviceGridRank == 2);
```
sohaibnadeemTT, line 1026: allow larger deviceGridRanks since we are passing optimal grids here (which can be virtual)

sohaibnadeemTT, line 1026: allowed larger deviceGridRanks since we are passing optimal grids here (which can be virtual)

lib/Dialect/TTCore/IR/TTCoreOpsTypes.cpp
```diff
@@ -1014,7 +1014,7 @@ MetalLayoutAttr::computeTileAlignments(ArrayRef<int64_t> logicalShape,
 // more predictable NoC traffic patterns.
 llvm::SmallVector<int64_t> MetalLayoutAttr::computeGridAwareDimAlignments(
     ArrayRef<int64_t> logicalShape, ArrayRef<int64_t> deviceGridShape,
-    ArrayRef<int64_t> normalizedIntervals) {
+    ArrayRef<int64_t> normalizedIntervals, bool forceGridAlignment) {
   constexpr std::array<int64_t, 2> tileShape =
```
sohaibnadeemTT, line 1018: forceGridAlignment is needed since we use computeGridAwareDimAlignments in 2 different scenarios:
- in computePhysicalShape, we are just trying to get an aligned shape to later compute the optimal grid, so we don't want to align to the full grid
- in layoutWithOptimalGrid, we are trying to align the tensor to the optimal grid so we need forceGridAlignment=true to ensure this

sohaibnadeemTT:
@nsmithtt @bgrady-tt thoughts?

lib/Dialect/D2M/Transforms/GridSelection.cpp
bgrady-tt: This makes sense; constrained optimal shapes should be generated from the start, not adjusted at the end.

nsmithtt (approved):

nsmithtt:
> Thanks for adding tests! Hmm, would be curious to get others' thoughts on this, I had in mind more the p150 tests would only be for things like explicit grid checks etc, so maybe the grid_selection_bh.mlir is sufficient. I could see an argument that some more complex checks in the p150 dir for specific ops could have value though (e.g. checking that the p150 eltwise add has correct 10x13 grid, AND that it compiles completely), so I could see either way--I guess up to you, or we could ping others for opinions or something.
> 
> Either way LGTM, thanks!

Yeah it would be good to have some key op types tested to ensure they use the full device grid, e.g. eltwise + matmul + reductions on R or C

vwellsTT:
> > Thanks for adding tests! Hmm, would be curious to get others' thoughts on this, I had in mind more the p150 tests would only be for things like explicit grid checks etc, so maybe the grid_selection_bh.mlir is sufficient. I could see an argument that some more complex checks in the p150 dir for specific ops could have value though (e.g. checking that the p150 eltwise add has correct 10x13 grid, AND that it compiles completely), so I could see either way--I guess up to you, or we could ping others for opinions or something.
> > Either way LGTM, thanks!
> 
> Yeah it would be good to have some key op types tested to ensure they use the full device grid, e.g. eltwise + matmul + reductions on R or C

So there's a test w/ this and a hardcoded device descriptor w/ 10x13 grid here: https://github.com/tenstorrent/tt-mlir/pull/6773/changes#diff-e1d59ec882075c5d9d6a3ede8f1415d2a39e53f67cc2a9912ed0ef10e60f4dd1
I didn't realize that was possible, but I think debatably that's good enough instead of a p150 test dir--downside is no full compilation, but I think it should cover basic IR checks for BH grid type?
Though yeah it would be good to add some kind of transpose/permute + reductions I guess

sohaibnadeemTT:
@jdesousa-TT @nsmithtt @bgrady-tt

While fixing some of failing 1d matmul tests, I noticed that my changes will nerf 1d matmul in some cases

E.g. operand 0 is (32, 4096) and operand 1 is (4096, 2048)  

Previously, we we were only using the normalized grids in the inverse views inserted for the generic so we would choose a layout of  1x64, 8x8, and 1x64 (output) for the operands but 1x64, 64x64, and 1x64 for the views into the generic

but with my change we are using the constrained grids to update the to_layout/stream_layout/view_layout as well and we have to take the minimum to respect grid constraints, so we end up with 1x8, 8x8, and 1x8 grids for both the operands’ physical layouts and the generic inputs’ views.

I think we need more sophisticated analysis that looks at what grid layouts we can support for each operand and try to pick the optimal grids that align with each other based on the indexing maps. 

For now, I’m thinking of just disabling the affected 1d matmul tests until we have such an analysis pass (see latest commit).

Issue #5951, opened 2025-11-20
Commute reshape with softmax


mvasiljevicTT:
ViT model has this pattern:
```
reshape (BH, N, N) -> (B, H, N, N)
.
. <- ops that only affect last two dims
.
reshape (B, H, N, N) -> (BH, N, N)
```
as:
<img width="332" height="673" alt="Image" src="https://github.com/user-attachments/assets/71bb7d99-6d9b-4210-9dd5-741171f17149" />

This reshape could commute with all the ops in the shown subgraph, but it is not currently implemented for softmax, repeat and where.

mvasiljevicTT:
This will also improve performance of swin model as it has 24 softmax-es each surrounded by two reshapes like:
```
%345 = "ttnn.reshape"(%344)  tensor<1083x49x49xbf16> -> tensor<361x3x49x49xbf16>
%346 = "ttnn.softmax"(%345) <{dimension = 3 : si32, numericStable = true}> 
%347 = "ttnn.reshape"(%346)  tensor<361x3x49x49xbf16> -> tensor<1083x49x49xbf16>
```
which would be rewritten with this commute as just:
```
%346 = "ttnn.softmax"(%345) <{dimension = 2 : si32, numericStable = true}> 
```

mvasiljevicTT:
@HalfBloodPrince010 if you are interested, this is a good starting point to getting to know EIO pass

#### About EIO pass
When convolutions are lowered through torch-xla they usually have `[N, C, H, W]` input shape. Our metal `conv2d` op needs flattened `[1, 1, NHW, C]` input shape which is why it is lowered as:
```
permute [N, C, H, W] - > [N, H, W, C]
reshape [N, H, W, C] -> [1, 1, NHW, C]
conv2d
reshape [1, 1, N'H'W', C'] -> [N' H', W', C']
permute  [N' H', W', C'] ->  [N' C', H', W']
```
If our model only had convolutions, inverse permutes would be one after another and cancel each other, then reshapes, and we would be left only with starting TMs (tensor manipulation ops) and ending. 

But since some other ops come in between, we define how reshape and permute can commute with them so that they can move up and down the graph hoping to come across other TMs that will cancel them (you are already familiar with folding patterns).

In `lib/Dialect/TTIR/Transforms/EraseInverseOps/EraseInverseOps.cpp` you can see the process:
- move (commute) all TMs up as possible 
- move all TMs down as possible
- repeat until number of TMs stops decreasing

[...]

PR #6779, opened 2026-01-28
[EIO] Add commute patterns for Softmax with Reshape

HalfBloodPrince010:
### Ticket
[Link to Github Issue](https://github.com/tenstorrent/tt-mlir/issues/5951)

### Problem description
Commute Patterns for Softmax with Reshape.
This commit adds changes to commute upwards and downwards the reshape op through the softmax using stride and reduction length of the softmax dimension.

### What's changed
Added Commute Patterns for Softmax with Reshape. Now,  Inverse TensorManipulation Ops like Reshape can be folded by commuting them upwards or downwards through the graph, improving the performance.

For eg, swin model as it has 24 softmax-es each surrounded by two reshapes, which could be folded away.

```
%345 = "ttnn.reshape"(%344)  tensor<1083x49x49xbf16> -> tensor<361x3x49x49xbf16>
%346 = "ttnn.softmax"(%345) <{dimension = 3 : si32, numericStable = true}> 
%347 = "ttnn.reshape"(%346)  tensor<361x3x49x49xbf16> -> tensor<1083x49x49xbf16>
```

which would be rewritten with this commute as just:

```
%346 = "ttnn.softmax"(%345) <{dimension = 2 : si32, numericStable = true}>
```

#### Update Softmax Dimension with Stride and reduction Length

Maps a softmax axis across a reshape by matching both: 
1. The reduction length of the axis, and 
2. Its contiguous stride.

```
     Consider a Tensor with shape = [2, 3, 4]
       i=0,j=0,k=0
       i=0,j=0,k=1
       i=0,j=0,k=2
       i=0,j=0,k=3
       i=0,j=1,k=0
       ...
     strides = [3*4, 4, 1]

Example 1 (Valid commute):
       Input shape      = [2, 3, 4]
       Softmax dim      = 2  (stride = 1, reduction length = 4)
       Reshape          = [6, 4]
The new softmax dim is 1, since output dim 1 still has stride = 1 and reduction length = 4.
     
Example 2 (Invalid Commute)
       Input shape      = [2, 3, 4]
       Softmax dim      = 1  (stride = 4, reduction length = 3)
       Reshape          = [2, 12]
output strides = [12, 1],  Hence, we can't apply softmax because no dimension in output has stride 4 and reduction len = 3
```

#### Upwards

```
   %0 = softmax(%arg0 : tensor<1x160x160x128xbf16>) <{dimension = 3}>
   %1 = reshape(%0) <{shape = [1 : i32, 25600 : i32, 128 : i32]}>
to
   %0 = reshape(%arg0 : tensor<1x160x160x128xbf16>) <{shape = [1 : i32, 25600 : i32, 128 : i32]}>
   %1 = softmax(%0) <{dimension = 2}>
```

#### Downwards
```
   %0 = reshape(%arg0 : tensor<1x160x160x128xbf16>) <{shape = [1 : i32, 25600 : i32, 128 : i32]}>
   %1 = softmax(%0) <{dimension = 2}>
 
to
   
   %0 = softmax(%arg0 : tensor<1x160x160x128xbf16>) <{dimension = 3}>
   %1 = reshape(%0) <{shape = [1 : i32, 25600 : i32, 128 : i32]}>
```

### Testing

```
pre-commit run --show-diff-on-failure --color=always --all-files
black....................................................................Passed
clang-format.............................................................Passed
Check copyright notices..................................................Passed
Trim Trailing Whitespace.................................................Passed
Fix End of Files.........................................................Passed
Check for added large files..............................................Passed
Fix include style........................................................Passed
```

```
llvm-lit -sv ./test/ttmlir/Dialect/TTIR/Transforms/EraseInverseOps/CommuteUpwards/commute_reshape_softmax.mlir 

Testing Time: 0.28s

Total Discovered Tests: 1
  Passed: 1 (100.00%)
```

```
llvm-lit -sv ./test/ttmlir/Dialect/TTIR/Transforms/EraseInverseOps/CommuteDownwards/commute_reshape_softmax.mlir 

Testing Time: 0.07s

Total Discovered Tests: 1
  Passed: 1 (100.00%)
```

### Checklist

### DIFF
--- a/lib/Dialect/TTIR/Transforms/EraseInverseOps/SoftmaxCommutePatterns.cpp
+++ b/lib/Dialect/TTIR/Transforms/EraseInverseOps/SoftmaxCommutePatterns.cpp
@@ -1,0 +1,259 @@
+// SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#include "ttmlir/Dialect/TTIR/Transforms/EraseInverseOps/EraseInverseOps.h"
+
+#include "mlir/IR/BuiltinTypes.h"
+#include "ttmlir/Dialect/TTCore/IR/Utils.h"
+#include "ttmlir/Dialect/TTIR/IR/TTIROps.h"
+#include "ttmlir/Dialect/TTIR/Utils/Utils.h"
+
+namespace mlir::tt::ttir {
+
+namespace {
+
+int64_t
+findNewSoftmaxDimByStrideAndReductionLen(llvm::ArrayRef<int64_t> inputShape,
+                                         llvm::ArrayRef<int64_t> outputShape,
+                                         int64_t softmaxDim) {
+  // Stride of new softmax dim in new shape must match stride of old
+  // softmax dim in old shape, and same for length.
+  //
+  // Example 1 (Valid commute):
+  //   Input shape      = [2, 3, 4]
+  //   Softmax dim      = 2  (stride = 1, reduction length = 4)
+  //   Reshape          = [6, 4]
+  // The new softmax dim is 1, since output dim 1 still has stride = 1 and
+  // reduction length = 4.
+  //
+  // Example 2 (Invalid Commute)
+  //   Input shape      = [2, 3, 4]
+  //   Softmax dim      = 1  (stride = 4, reduction length = 3)
+  //   Reshape          = [2, 12]
+  //   output strides = [12, 1],  Hence, we can't apply softmax because no
+  //   dimension in output has stride 4 and reduction len = 3
+  int64_t inputDims = inputShape.size();
+  int64_t outputDims = outputShape.size();
+
+  if (softmaxDim < 0) {
+    softmaxDim += inputDims;
+  }
+
+  assert(softmaxDim >= 0 && softmaxDim < inputDims &&
+         "Invalid softmax Dimension");
+
+  int64_t softmaxReductionLength = inputShape[softmaxDim];
+  int64_t inputStride = 1;
+  for (int64_t i = softmaxDim + 1; i < inputDims; ++i) {
+    inputStride *= inputShape[i];
+  };
+
+  // Commute is valid if
+  // 1. Input Stride of softmax dimension is preserved in output, and
+  // 2. The number of elements the over which softmax normalizes/reduces should
+  // also match
+  int64_t outputStride = 1;
+  for (int64_t dim = outputDims - 1; dim >= 0; dim--) {
+    if (outputStride == inputStride &&
+        outputShape[dim] == softmaxReductionLength) {
+      return dim;
+    }
+    outputStride *= outputShape[dim];
+  };
+
+  return -1;
+}
+
+template <CommuteDirection commuteDirection>
+// <TMOp, Op, CommuteDirection>
+class TTIRCommuteReshapeThroughSoftmax
+    : public TTIRCommuteOpRewritePattern<ReshapeOp, SoftmaxOp,
+                                         commuteDirection> {
+public:
+  using TTIRCommuteOpRewritePattern<
+      ReshapeOp, SoftmaxOp, commuteDirection>::TTIRCommuteOpRewritePattern;
+
+  // Consider the following IR pseudocode:
+  // Before commuting reshape op upwards
+  //   %a = softmax(%arg0, dim)
+  //   %b = reshape(%a)
+  // After commuting reshape op upwards
+  //   %x = reshape(%arg0)
+  //   %y = softmax(%x, UpdatedDim)
+  void performCommuteUpwardsRewrite(SoftmaxOp op, ReshapeOp reshapeUser,
+                                    PatternRewriter &rewriter) const override {
+    auto softmaxInputType = cast<RankedTensorType>(op.getInput().getType());
+    auto outputReshapeType =
+        cast<RankedTensorType>(reshapeUser.getResult().getType());
+
+    auto softmaxInputShape = softmaxInputType.getShape();
+    auto outputReshapeShape = outputReshapeType.getShape();
+    int64_t softmaxDim = static_cast<int64_t>(op.getDimension());
+    // We are mapping the softmax axis from softmax input to reshape user output
+    int64_t newSoftmaxDim = findNewSoftmaxDimByStrideAndReductionLen(
+        softmaxInputShape, outputReshapeShape, softmaxDim);
+    assert(newSoftmaxDim != -1 &&
+           "Invalid Softmax Dimension, isCommuteUpwardsFavorable should have "
+           "handled this");
+
+    // Create a new Reshape before the Softmax op
+    auto newInputReshapeType = RankedTensorType::get(
+        outputReshapeType.getShape(), softmaxInputType.getElementType(),
+        outputReshapeType.getEncoding());
+    SmallVector<int32_t> newReshapeShape(outputReshapeShape.begin(),
+                                         outputReshapeShape.end());
+
+    auto newInputReshape = rewriter.create<ReshapeOp>(
+        reshapeUser->getLoc(), newInputReshapeType, op.getInput(),
+        rewriter.getI32ArrayAttr(newReshapeShape));
+
+    auto newSoftmaxDimAttr = rewriter.getSI32IntegerAttr(newSoftmaxDim);
+
+    // Create a new Softmax Op with updated dimension attribute
+    auto newSoftmaxOp = rewriter.create<SoftmaxOp>(
+        op->getLoc(), outputReshapeType, newInputReshape.getResult(),
+        newSoftmaxDimAttr, op.getNumericStableAttr());
+
+    // All users must be identical TMs.
+    SmallVector<Operation *> users(op->getUsers());
+    assert(llvm::all_of(users,
+                        [&](Operation *user) {
+                          return checkIdenticalTms(reshapeUser, user);
+                        }) &&
+           "isCommuteUpwardsViable/Favorable should have ensured all users "
+           "are identical TMs");
+
+    for (auto *user : users) {
+      rewriter.replaceOp(user, newSoftmaxOp.getResult());
+    };
+  }
+
+  // Consider the following IR pseudocode:
+  // Before commuting reshape op downwards
+  //   %x = reshape(%arg0)
+  //   %y = softmax(%x, dim)
+  // After commuting reshape op downwards
+  //   %a = softmax(%arg0, updatedDim)
+  //   %b = reshape(%a)
+  void
+  performCommuteDownwardsRewrite(SoftmaxOp op, ReshapeOp reshapeOperand,
+                                 PatternRewriter &rewriter) const override {
+    auto reshapeOperandOutputShape =
+        reshapeOperand.getResult().getType().getShape();
+    auto reshapeOperandInputShape =
+        reshapeOperand.getInput().getType().getShape();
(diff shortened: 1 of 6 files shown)


HalfBloodPrince010:
Hi @mvasiljevicTT, could you help run the tests? Thank you.

lib/Dialect/TTIR/Transforms/EraseInverseOps/SoftmaxCommutePatterns.cpp
```diff
@@ -0,0 +1,297 @@
...
+
+  int64_t softmaxReductionLength = inputShape[softmaxDim];
+  // Assuming the shape products fit in int64_t
+  int64_t inputStride = 1;
+  for (int64_t i = softmaxDim + 1; i < inputDims; ++i) {
+    inputStride *= inputShape[i];
+  };
+
+  int64_t outputStride = 1;
+  SmallVector<int64_t> outputStrides(outputDims);
```
mvasiljevicTT: This will be an increasing array so instead of calculating stride for each dim and then matching, you can have one loop that travels through output shape from the back and accumulates the stride. 
Once the strides match, compare input and output shape at that dim - it can either match there, be a 1 in output shape in which case we continue, or mismatches in which case return -1.

lib/Dialect/TTIR/Transforms/EraseInverseOps/SoftmaxCommutePatterns.cpp
```diff
@@ -0,0 +1,297 @@
...
+    int64_t softmaxDim = static_cast<int64_t>(op.getDimension());
+    // We are mapping the softmax axis from softmax input to reshape user output
+    int64_t newSoftmaxDim = findNewSoftmaxDimByStrideAndReductionLen(
+        softmaxInputShape, outputReshapeShape, softmaxDim);
+    assert(newSoftmaxDim != -1 &&
+           "Invalid Softmax Dimension, isCommuteUpwardsFavorable should have "
+           "handled this");
+
+    // Create a new Reshape before the Softmax op
+    auto newReshapeInputType = RankedTensorType::get(
```
mvasiljevicTT: This name is misleading, this is input type for the softmax, which is output shape for the reshape.
You can assume elementType stays the same and reuse the type, like you do in downwrads.

lib/Dialect/TTIR/Transforms/EraseInverseOps/SoftmaxCommutePatterns.cpp
```diff
@@ -0,0 +1,297 @@
...
+
+    auto newInputReshape = rewriter.create<ReshapeOp>(
+        reshapeUser->getLoc(), newReshapeInputType, op.getInput(),
+        rewriter.getI32ArrayAttr(newReshapeShape));
+
+    // Softmax expects a signed integer attribute (si32)
+    auto si32Ty =
+        IntegerType::get(rewriter.getContext(), 32,
+                         mlir::IntegerType::SignednessSemantics::Signed);
+    auto newSoftmaxDimAttr = rewriter.getIntegerAttr(si32Ty, newSoftmaxDim);
```
mvasiljevicTT: Can `getSI32IntegerAttr` be used here?

lib/Dialect/TTIR/Transforms/EraseInverseOps/SoftmaxCommutePatterns.cpp
```diff
@@ -0,0 +1,297 @@
...
+  // transform this IR into: %0 = softmax(%arg0 : tensor<1x160x160x128xbf16>)
+  // <{dimension = 3}> %1 = reshape(%0) <{shape = [1 : i32, 25600 : i32, 128 :
+  // i32]}>
+  void
+  performCommuteDownwardsRewrite(SoftmaxOp op, ReshapeOp reshapeOperand,
+                                 PatternRewriter &rewriter) const override {
+    // Get the updated softmax dim
+    auto reshapeOperandOutputShape = reshapeOperand.getResult()
+                                         .getType()
+                                         .getShape(); // Reshape Operand Output
```
mvasiljevicTT: No need for comments when it is clear from the var name. On that note, check that all names are reasonable.

lib/Dialect/TTIR/Transforms/EraseInverseOps/SoftmaxCommutePatterns.cpp
```diff
@@ -0,0 +1,297 @@
...
+
+    // Create new reshape Op
+    // Note: For the new reshape op, type is the softmax op
+    // Before
+    //   %x = reshape(%arg0)
+    //   %y = softmax(%x, dim)
+    // After commuting downwards
+    //   %a = softmax(%arg0, updatedDim)
+    //   %b = reshape(%a)
+    // Hence, output type of reshape should match %y (original softmax op)
```
mvasiljevicTT: I like this explanation, you could move it above the function instead of current docstring as it is more informative on what this rewrite is doing.

lib/Dialect/TTIR/Transforms/EraseInverseOps/SoftmaxCommutePatterns.cpp
```diff
@@ -0,0 +1,297 @@
...
+  //   (1) the reduction length of the axis, and
+  //   (2) its contiguous stride.
+  // Consider a Tensor with shape = [2, 3, 4]
+  //   i=0,j=0,k=0
+  //   i=0,j=0,k=1
+  //   i=0,j=0,k=2
+  //   i=0,j=0,k=3
+  //   i=0,j=1,k=0
+  //   ...
+  // strides = [3*4, 4, 1]
```
mvasiljevicTT: Tbh i don't get this `i, j, k` part. Try changing it with the simpler explanation of what the algorithm is doing in the context of shapes - stride of new softmax dim in new shape must match stride of old softmax dim in old shape, and same for length.

lib/Dialect/TTIR/Transforms/EraseInverseOps/SoftmaxCommutePatterns.cpp
```diff
@@ -0,0 +1,297 @@
...
+public:
+  using TTIRCommuteOpRewritePattern<
+      ReshapeOp, SoftmaxOp, commuteDirection>::TTIRCommuteOpRewritePattern;
+
+  // Consider the following IR pseudocode:
+  // %0 = softmax(%arg0 : tensor<1x160x160x128xbf16>) <{dimension = 3}>
+  // %1 = reshape(%0) <{shape = [1 : i32, 25600 : i32, 128 : i32]}>
+  // This method will transform this IR into:
+  // %0 = reshape(%arg0 : tensor<1x160x160x128xbf16>) <{shape = [1 : i32, 25600
+  // : i32, 128 : i32]}> %1 = softmax(%0) <{dimension = 2}>
```
mvasiljevicTT: format this to be more readable, it doesn't have to keep real syntax as it is a pseudocode.

mvasiljevicTT (commented):
Sorry for the delay, thank you for doing this!!

I love that you are changing reduction dim when possible, instead of just matching ones where it stays the same!

lib/Dialect/TTIR/Transforms/EraseInverseOps/SoftmaxCommutePatterns.cpp
HalfBloodPrince010: got it, I will make the change to be more readable IR.

lib/Dialect/TTIR/Transforms/EraseInverseOps/SoftmaxCommutePatterns.cpp
HalfBloodPrince010: makes sense, I will merge the two loops, and remove outputStrides Vector.

lib/Dialect/TTIR/Transforms/EraseInverseOps/SoftmaxCommutePatterns.cpp
HalfBloodPrince010: This is the type for the new "reshape op" which becomes the input after commuting upwards, hence named it as `newReshapeInputType`.

This mirrors the `newSoftmaxInputType `naming used in the `downwards` commute function

Similar to:  https://github.com/tenstorrent/tt-mlir/blob/main/lib/Dialect/TTIR/Transforms/EraseInverseOps/RMSNormCommutePatterns.cpp#L39

let me know if I should just name it "newInputType" to avoid confusion.

lib/Dialect/TTIR/Transforms/EraseInverseOps/SoftmaxCommutePatterns.cpp
HalfBloodPrince010: Yes, I will update this, thanks!

lib/Dialect/TTIR/Transforms/EraseInverseOps/SoftmaxCommutePatterns.cpp
HalfBloodPrince010: got it, will remove these!

lib/Dialect/TTIR/Transforms/EraseInverseOps/SoftmaxCommutePatterns.cpp
HalfBloodPrince010: Yes, will do

lib/Dialect/TTIR/Transforms/EraseInverseOps/SoftmaxCommutePatterns.cpp
HalfBloodPrince010: got it, will remove that

lib/Dialect/TTIR/Transforms/EraseInverseOps/SoftmaxCommutePatterns.cpp
HalfBloodPrince010: @mvasiljevicTT  let me know if I am missing anything. Else, will raise the revision with the other changes.

mvasiljevicTT: In rmsnorm `newInputReshape` is referring to reshaping input of the rmsnorm, and `newInputType` is then type of the input of the rmsnorm. That is used as target shape in the reshape op.

Similarly, this could be called `newInputReshapeType` as type of `newInputReshape`, but `newReshapeInputType` is misleading that it is `input type` of the `new reshape`, when it is it's output type.


When creating new nodes you don't define input type. You already use input `Value`, `op.getInput()` in this case, that is defining the input edge of that node. You need to set type of the output edge, and the attributes, which is the target shape for the reshape.

So when later in the downwards you have
```
auto newSoftmaxOp = rewriter.create<SoftmaxOp>(
        op->getLoc(), newSoftmaxInputType, reshapeOperand.getInput(),
        newSoftmaxDimAttr, op.getNumericStableAttr());
```
that says - input edge of the new node is `reshapeOperand.getInput()` and output edge has `newSoftmaxInputType`. 

Since softmax doesn't change shape this is ok.

HalfBloodPrince010: @mvasiljevicTT thank you for explaining, I will update `newReshapeInputType` to `newInputReshapeType` and raise the revision.

HalfBloodPrince010:
Hi @mvasiljevicTT  I have made the changes, could you help run the tests? Thank you.

HalfBloodPrince010:
Hi @mvasiljevicTT , the tests are passing, could you help merge it?

Issue #5943, opened 2025-11-19
Extract ttnn allocator state and pass to d2m pipeline / allocator pass options


jdesousa-TT:
We need to utilize the ttmetal allocator state to signal the available range of memory for the d2m allocator to use when context switching into d2m mode through ttnn-jit.

sgholamiTT:
I've drafted a POC for this that can be found [here](https://github.com/tenstorrent/tt-mlir/compare/sgholami/test-binding?diff=unified&w). It needs [this](https://github.com/tenstorrent/tt-metal/compare/sgholami/test-binding) metal commit for the python bindings.
@jdesousa-TT I'm going to self-assign this.

PR #6795, opened 2026-01-28
Integrate Metal Allocator with TTNN-JIT for D2M Memory Management

sgholamiTT:
### Ticket
https://github.com/tenstorrent/tt-mlir/issues/5943

## Summary

This PR integrates Metal's memory allocator with `ttnn-jit` to provide the D2M compiler with accurate information about available `L1/DRAM` memory ranges at JIT compilation time. This enables D2M to make informed decisions about buffer allocation and avoid memory conflicts with tensors already allocated on the device.

## Motivation

When `ttnn-jit` compiles a function, the device may already have tensors allocated in L1 and DRAM memory. Without knowledge of the current memory state, D2M cannot safely allocate scratch buffers or intermediate tensors without risking memory conflicts. This is particularly important for:
1. **Scratch pad allocation**: D2M needs to know the contiguous free memory range available for intermediate computations.
2. **Output tensor pre-allocation**: Understanding how much memory the output tensor will require helps D2M plan memory layout and potentially reserve space.

## Problem Statement

Before this change:
- D2M had no visibility into the device's current memory allocation state
- JIT compilation could not communicate available memory ranges to the compiler pipeline
- Memory allocation conflicts could occur when D2M-generated code tried to allocate buffers in regions already used by input tensors

### 1. Metal Buffer Allocation API (C++ Bindings)
Added Python bindings for Metal's memory APIs in `runtime/python/runtime/utils.cpp`:

**L1 Memory Query APIs** (direct allocator queries, no allocation needed):
```cpp
get_l1_base_allocator_addr(mesh_device)              // Start of allocatable L1 region
get_lowest_occupied_compute_l1_address(mesh_device)  // End of free region (lowest occupied addr)
get_l1_size_per_core(mesh_device)                    // Total L1 size per core
```

**Buffer Allocation APIs** (for DRAM probing):
```cpp
MeshBufferWrapper  // Wrapper class exposing address(), size(), deallocate()
allocate_dram_buffer(mesh_device, buffer_size, page_size)  // Allocate DRAM buffer
```

### 2. MemoryAnalyzer Class

Available in `tools/ttnn-jit/_src/memory_analyzer.py`, which performs the following actions:

1. **L1 Memory** (direct query - no allocation):
   - Uses `get_l1_base_allocator_addr()` to get the start of allocatable region
   - Uses `lowest_occupied_compute_l1_address()` to get the end of free region
   - Stores the `[start, end)` address range directly

2. **DRAM Memory** (temporary buffer allocation):
   - Allocates the largest possible DRAM buffer using Metal's allocator
   - Records the `[start, end)` address range, then deallocates

3. **Pre-allocates output tensor**: Creates a `ttnn.empty` tensor matching the output tensor's
   properties (shape, dtype, memory config including sharding) to determine what address range
   the output will occupy. This considers:
   - Tensor shape and data type
   - Memory layout (L1 vs DRAM)
   - Sharding strategy (BLOCK, HEIGHT, WIDTH, or Interleaved)
   - Core grid configuration

4. **Generates D2M options**: Produces compiler option strings like:
   ```
   available-l1-addr-range=103552,1490944
   ```

This is an example of this process. 
Untitled drawing (3)

Note 1: The memory location dedicated to (pre-allocated for) the output tensor is not necessarily within the `available-l1-addr-range` (eg. memory fragmentation, or output tensor requested to be located on `DRAM`).
Note 2: As shown in the picture, we first acquire the address of the largest chunk of memory in L1 and DRAM and then we pre-allocate the output tensor. There are pros and cons to this:
- Pros: Larger memory will be available to D2M as scratch pad,
- Cons: If D2M uses all that memory, it might become impossible to allocate the output tensor.

@xanderchin and @nsmithtt decided that if we do it the other way around, D2M might actually end up with very small to no space in L1 to work with (in case of large output tensors, eg. `matmul` with small `k` and large `m` and `n`).

### 3. Integration with JIT Pipeline

Modified `tools/ttnn-jit/_src/jit.py` to:

1. Extract output tensor type from the tracing compiler
2. Instantiate `MemoryAnalyzer` with the device and output type
3. Pass L1 address range to D2M via compiler options

```python
ir, output_type = generate_ir(self.func, self.debug, *args, **kwargs)
memory_analyzer = MemoryAnalyzer(device, output_type)
options += memory_analyzer.get_l1_range_str()
```

### L1 Memory Range Extraction (Direct Query)

L1 buffers in Metal grow **downward** from high addresses. The `MemoryAnalyzer` directly queries
the allocator to determine the free region without allocating temporary buffers:

```
L1 Memory Layout (per core):
+----------------------------+ <-- High Address (l1_size_per_core)
|    Allocated L1 Buffers    |     (grow downward)
+----------------------------+ <-- lowest_occupied_compute_l1_address()
|                            |
|         FREE SPACE         |  <-- Available for D2M / CBs
|                            |
+----------------------------+ <-- get_l1_base_allocator_addr()
|  Reserved (firmware, etc.) |
+----------------------------+ <-- Low Address (0)
```

```python
start = get_l1_base_allocator_addr(device)
lowest = get_lowest_occupied_compute_l1_address(device)
end = lowest if lowest is not None else get_l1_size_per_core(device)
# Range: [start, end)
```

### DRAM Memory Range Extraction (Buffer Allocation)

For DRAM, temporary buffer allocation is used to probe the available range:

```python
buffer = allocate_dram_buffer(device, largest_free_size, largest_free_size)
start = buffer.address()
end = start + size
buffer.deallocate()
```

### Output Tensor Size Calculation

For sharded tensors, the per-core buffer size is calculated as:
```python
tiles_in_shard = (shard_shape[0] // 32) * (shard_shape[1] // 32)
size_bytes = tiles_in_shard * tile.get_tile_size(dtype)
```

For interleaved tensors:
```python
num_tiles = ceil(tensor.volume() / (32 * 32))
size_bytes = num_tiles * tile.get_tile_size(dtype)
```

### Memory Config from MLIR Encoding

The MLIR `TTNNLayoutAttr` encoding is parsed to reconstruct the exact memory config:
- Buffer type (L1/DRAM) from `memref.memory_space`
- Tensor memory layout (Interleaved, BlockSharded, HeightSharded, WidthSharded)
- Core grid shape for sharded tensors

## Future Work

- `get_dram_range_str()` and `get_output_range_str()` are implemented but not yet consumed 
  by D2M (prefixed with `not-implemented-`) See related issues: 
  - https://github.com/tenstorrent/tt-mlir/issues/6821
  - https://github.com/tenstorrent/tt-mlir/issues/6820
- These will be enabled when D2M adds support for DRAM range and output tensor pre-allocation hints
- DRAM currently uses temporary buffer allocation; if Metal exposes a `highest_occupied_dram_address()`
  API in the future, we can switch to the same direct query approach used for L1
- The way ttnn-jit interacts with D2M should be unified with how Optimizer will be interacting with D2M. See related issue:
  - https://github.com/tenstorrent/tt-mlir/issues/6822 

### Checklist
- [X] Nightly run passes ([Link](https://github.com/tenstorrent/tt-mlir/actions/runs/21496471090))
- [X] Added comprehensive test suite covering:
  - [X] Basic initialization with sharded/interleaved tensors
  - [X] Range string format validation
  - [X] All sharding strategies (BLOCK, HEIGHT, WIDTH)
  - [X] L1 and DRAM interleaved tensors
  - [X] Various tensor shapes (32x32 to 1024x1024)
  - [X] Multiple data types (bfloat16, float32)
  - [X] End-to-end JIT integration
  - [X] Disabled mode behavior

### DIFF
--- a/test/ttnn-jit/test_memory_analyzer.py
+++ b/test/ttnn-jit/test_memory_analyzer.py
@@ -33,7 +33,7 @@ def _generate_ir_for_func(func, *args):
     sig = inspect.signature(func)
     param_names = list(sig.parameters.keys())
     tensor_args = {param_names[i]: args[i] for i in range(len(args))}
-    return generate_ir(func, False, *args, _tensor_args=tensor_args)
+    return generate_ir(func, False, None, *args, _tensor_args=tensor_args)
 
 
 class TestMemoryAnalyzerBasic:
--- a/tools/ttnn-jit/_src/memory_analyzer.py
+++ b/tools/ttnn-jit/_src/memory_analyzer.py
@@ -40,9 +40,10 @@ class MemoryAnalyzer:
     * Note that the numbers in 2.2 and 3.2 are not yet supported by D2M, so we output them as "not-implemented-...".
     """
 
-    # Alignment requirement for addresses (must be multiple of 32)
-    # It was suggested by @audrey_kertesz to always round down the addresses to the nearest multiple of 32 for alignment.
-    ALIGNMENT = 32
+    # Alignment requirement for addresses
+    # It was suggested by @audrey_kertesz to always round down the addresses to the nearest ALIGNMENT.
+    ALIGNMENT_L1 = 32
+    ALIGNMENT_DRAM = 64
 
     def __init__(self, device, output_type, perform_memory_analysis: bool = True):
         self.device = device
@@ -102,9 +103,12 @@ class MemoryAnalyzer:
     ############################# Private Methods #############################
     ###########################################################################
 
-    def _round_down_to_alignment(self, value: int) -> int:
+    def _round_down_to_alignment(self, value: int, is_l1: bool = True) -> int:
         """Round down to the nearest multiple of ALIGNMENT."""
-        return (value // self.ALIGNMENT) * self.ALIGNMENT
+        if is_l1:
+            return (value // self.ALIGNMENT_L1) * self.ALIGNMENT_L1
+        else:
+            return (value // self.ALIGNMENT_DRAM) * self.ALIGNMENT_DRAM
 
     def _get_l1_addr_range(self) -> tuple[int, int]:
         """
@@ -137,7 +141,7 @@ class MemoryAnalyzer:
         lowest = get_lowest_occupied_compute_l1_address(self.device)
         # If no allocations exist, lowest is None, use full L1 size
         end = lowest if lowest is not None else get_l1_size_per_core(self.device)
-        end = self._round_down_to_alignment(end)
+        end = self._round_down_to_alignment(end, is_l1=True)
         return (start, end)
 
     def _bytes_to_kb(self, bytes_val: int) -> float:
@@ -213,7 +217,7 @@ class MemoryAnalyzer:
 
         start_addr = buffer.address()
         end_addr = start_addr + buffer_size
-        end_addr = self._round_down_to_alignment(end_addr)
+        end_addr = self._round_down_to_alignment(end_addr, is_l1=False)
 
         # Deallocate the buffer before returning
         buffer.deallocate()


nsmithtt:
In order to get the "Area of Interest" we should double check if it would be better to use device api `device->lowest_occupied_compute_l1_address` to get the highest legal address for the CB region.  I'm not sure if `largest_contiguous_bytes_free_per_bank` will return this number or not, it'd be worth double checking.

sgholamiTT:
@nsmithtt thanks for your comment, this is an excellent observation.

I just checked and the API you suggested seems to be working fine (producing the same addresses that we're generating rn).

I have 3 concerns with the implementation which makes me doubtful to commit the code:
1. Apparently, metal folks don't want us to rely on/work with raw addresses. After long conversations, the conclusion was that instead of doing funky arithmetics on addresses, we should only rely on `allocator` being able to allocate a `buffer` with size `x` (where `x` was originally received via `statistics.largest_free_block_bytes` but [the PR](https://github.com/tenstorrent/tt-metal/pull/36512) was rejected by @ayerofieiev-tt, So right now we're using `memory_view.largest_contiguous_bytes_free_per_bank` which apparently gives the same number). So they want us to do sth like this:
```
buffer = allocate_buffer(buffer_size=x,page_size=x)
start_addr = buffer.address()
end_addr = start_addr + x
buffer.deallocate()
```
which is what I did in this PR.
2. I can see calculations like this being used vastly in metal ([eg](https://github.com/tenstorrent/tt-metal/blob/ab5455a003eb1cea44465f2eba4bee65bca49913/ttnn/cpp/ttnn/operations/matmul/device/utilities/matmul_utilities.cpp#L74-L80)):
```
lowest = lowest_occupied_compute_l1_address()
end =  lowest.has_value() ? lowest.value() : l1_size_per_core 
start = get_base_allocator_addr()
```
However, I'm not sure if this approach is accurate with existing `CB`s on `L1` (I'm mostly certain that when D2M starts, there are no CBs on L1 that are not deallocated, but I'm not %100 sure!). My understanding of `L1` and the `addresses` are as follows:
```
L1 Memory Layout (per core):
+----------------------------+ <-- High Address (l1_size_per_core)
|    Allocated L1 Buffers    |  
+----------------------------+        ( Grow downward )
|    Allocated L1 Buffers    |
+----------------------------+ <-- lowest_occupied_compute_l1_address()
|                            |
|         FREE SPACE         |
|                            |
+----------------------------+ 
|            CBs             |
+----------------------------+ <-- get_l1_base_allocator_addr()
|  Reserved (firmware, etc.) |
+----------------------------+ <-- Low Address (0)
```
So `[get_l1_base_allocator_addr, lowest_occupied_compute_l1_address)` does not correctly take into account existing `CB`s.

3. There are no equivalent APIs to perform a similar thing for `DRAM` (I couldn't find any). Therefore, this approach will introduce inconsistency between how `DRAM` and `L1` available memory is calculated (we need to keep the logic for `DRAM` and we'll have a new logic for `L1`).

Can you please let me know if any of these concerns are valid or not.
Just for the record, I'm completely fine with committing the approach you suggested, and in fact, I believe that's the better way for doing this 😄

sgholamiTT:
@nsmithtt :
After talking with @xanderchin I decided to go ahead with your suggestion for `L1` calculation. 
Please take another look (particularly at [this commit](https://github.com/tenstorrent/tt-mlir/pull/6795/changes/8a51b46f4a72d3d01405cd6af9eabac922d7b170)) and let me what you think. I also updated the PR description to reflect this change.

------------

For documentation, what we discussed is as follows for the concerns I had in my previous comment:
1. It's not very important as this is a temporary approach. We'll be coming up with a more robust CLI for AllocatorPass that receives the full allocator state. This can be populated by both `jit` and `Optimizer` as discussed this morning with other folks @rpavlovicTT @odjuricicTT [issue](https://github.com/tenstorrent/tt-mlir/issues/6822)
2. There will be no existing CBs when we enter D2M.
3. It's fine to have two approaches for calculating the range for L1 and DRAM as this is a temporary approach, [issue](https://github.com/tenstorrent/tt-mlir/issues/6822).

test/ttnn-jit/test_memory_analyzer.py
arminaleTT, line 1: I'd put this in the regular tests. Should be fast enough

tools/ttnn-jit/_src/memory_analyzer.py
```diff
@@ -0,0 +1,406 @@
...
+        3.2 It stores the [start,end) address range of the output tensor.
+        3.3 It deallocates the output tensor.
+
+    * Note that the range in 3.2 is not necessarily within the ranges in 1.2 and 2.2, since the memory might be fragmented.
+    * Note that the numbers in 2.2 and 3.2 are not yet supported by D2M, so we output them as "not-implemented-...".
+    """
+
+    # Alignment requirement for addresses (must be multiple of 32)
+    # It was suggested by @audrey_kertesz to always round down the addresses to the nearest multiple of 32 for alignment.
+    ALIGNMENT = 32
```
xanderchin: DRAM has a higher alignment requirement IIRC. 64 for WH and 128 for BH. Can you verify here?

Won't block this PR now as the current priority is L1.

sgholamiTT: Yes, you are right. `DRAM` alignment is 32 for WH and 64 for BH.
Modified the code to reflect the difference between DRAM and L1 alignment (tho most probably this is not necessary as allocator will take care of that, it's just to be safe).

PR #6807, opened 2026-01-29
Include compute kernel config in TTNN OpModel

rpavlovicTT:
Matmuls were invoked in TTNNOpModel without compute kernel config creating discrepancy between op model and runtime.


odjuricicTT:
@rpavlovicTT @sgholamiTT These types of bugs where runtime op calls are differ from constraint op calls are very frequent. We should discuss a strategy for testing for this.

rpavlovicTT:
> @rpavlovicTT @sgholamiTT These types of bugs where runtime op calls are differ from constraint op calls are very frequent. We should discuss a strategy for testing for this.

I agree totally.

PR #6808, opened 2026-01-29
Add L1 fragmentation check for chain merges

### LINKED ISSUE #6798: Issue: L1 Memory Fragmentation - Tensor/CB Address Overlap Causes OOM
## Summary

L1 memory fragmentation issue: OpModel validates total L1 memory sizes but not address layout. Runtime allocator places L1 tensors at addresses that overlap with the CB region, causing OOM despite sufficient total L1 space.

## Reproduction

- Model: Gemma 1.2B
- Operation: Matmul with `fp32_dest_acc_en=true` (down-projection)
- Error: `L1 region [598016, 876544] clashes with CB region ending at 763072`

## Root Cause

**OpModel query (NO_DISPATCH mode):**
- L1 tensors have `addr=0` (not actually allocated)
- Only validates: `peak_cb + peak_l1 < usable_L1` ✓

**Runtime:**
- Allocator places L1 tensor at addr=598,016
- CB region extends to 763,072
- Overlap: 598,016 < 763,072 ✗

**Why input tensor is at low address:**
- The input tensor is the output of a previous op that left it L1 sharded
- That previous op had its own inputs in L1, pushing its output to a lower address
- This cascading effect results in the tensor being placed in the CB region of a subsequent op

## The Gap

Optimizer cannot verify: `min_L1_tensor_address >= reserved_base + cbL1PeakSize`

Because actual tensor addresses are unknown at compile time.

## Potential Fix

**Compile-time**: Track and validate L1 tensor placement addresses during optimization to ensure no overlap with CB region.

rpavlovicTT:
After a chain merge (3-way or RHS), the join op's output may get allocated at a low L1 address when squeezed below other tensors. When those tensors are deallocated, the output remains stranded at the low address and may overlap with the CB region of subsequent ops.

This adds a fragmentation check that validates the op immediately after the join op with additional L1 reservation equal to the size of tensors that could push the output to a low address:
- For 3-way merge: lhsSize + rhsSize (both inputs to join op)
- For RHS merge: sourceOutputSize (merged chain's output)

This fixes OOM errors in models like Gemma 1.2B where the GeGLU pattern's 3-way merge caused the down-projection matmul to fail due to CB/tensor address overlap.

Fixes #6798

### DIFF
--- a/lib/Dialect/TTNN/Analysis/DFShardingPolicy.cpp
+++ b/lib/Dialect/TTNN/Analysis/DFShardingPolicy.cpp
@@ -255,6 +255,13 @@ static const std::vector<L1Reservation> kEmptyL1Reservations;
 static const llvm::DenseMap<Operation *, TTNNLayoutAttr> kEmptyL1ResidentsMap;
 static const llvm::DenseMap<Operation *, OpConfig> kEmptyModifiedOpConfigs;
 
+// Forward declaration for fragmentation check used by merge validation.
+static bool validateOpAfterJoinWithFragmentationReserve(
+    const L1ChainConfig &chainB, size_t joinOpIndex,
+    uint64_t fragmentationReserve,
+    const llvm::DenseMap<Operation *, int64_t> &schedulePositionMap,
+    const std::vector<L1Reservation> &l1Reservations);
+
 // Validate that Chain B can execute with Chain A's output in L1.
 // This validates each op in Chain B (up to and including joinOp) with
 // Chain A's output as additional L1 usage. At the join op, we also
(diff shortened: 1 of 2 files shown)


odjuricicTT (approved):
@rpavlovicTT 
Another solution for this is to insert a ttnn.move to defragment the memory after the join op?

rpavlovicTT:
> @rpavlovicTT Another solution for this is to insert a ttnn.move to defragment the memory after the join op?

That is true, but only if we knew for sure that fragmentation is present. I suppose we can explore this option with allocator state integrated.

Issue #6587, opened 2026-01-14
[EmitPy] Improve SubscriptOp to work with dictionary type


amilovanovicTT:
Currently, ```SetValueForDictKeyOp``` and ```GetValueForDictKeyOp``` exist in the EmitPy dialect to add and retrieve values for specific dictionary keys. However, these two ops seem to be redundant as ```SubscriptOp``` could be extended to work with EmitPy_DictType.

PR #6809, opened 2026-01-29
[EmitPy] Refactor dictionary ops

amilovanovicTT:
### Ticket
#6587

### Problem description
Previously, `SetValueForDictKeyOp` and `GetValueForDictKeyOp` were used to manipulate dictionary entries. However, these operations were overly specialized, contradicting the MLIR philosophy of maintaining generic, reusable operations within a dialect.
### What's changed
- Introduced `EmitPy_StringType`
- Refactored `SubscriptOp`: Updated to support  `EmitPy_DictType`. When used with dictionaries, it accepts `EmitPy_StringType` as an index. This serves as the direct replacement for `GetValueForDictKeyOp`.
- Refactored `AssignOp`: Extended to support subscript assignment (e.g., `dict[key] = value`). This is done by using expressionOp that has the following form: subscriptOp, assignOp, constantOp (its result will be used as an input to the yieldOp), yieldOp. This serves as the direct replacement for `SetValueForDictKeyOp`.
- Deprecation: Removed `SetValueForDictKeyOp` and `GetValueForDictKeyOp` in favor of the more general operations above.

### Checklist

### DIFF
--- a/include/ttmlir/Dialect/EmitPy/IR/EmitPyOps.td
+++ b/include/ttmlir/Dialect/EmitPy/IR/EmitPyOps.td
@@ -138,13 +138,13 @@ def EmitPy_SubscriptOp : EmitPy_Op<"subscript", [Pure, PyExpressionInterface]> {
   }];
 
   let arguments = (ins
-    Arg<AnyTypeOf<[EmitPy_OpaqueType, EmitPy_DictType]>, "the value to subscript">:$value,
+    Arg<AnyTypeOf<[EmitPy_OpaqueType, EmitPy_DictType]>, "the container to subscript">:$container,
     Arg<AnyTypeOf<[Index, EmitPy_StringType]>, "index or key value">:$index);
 
   let results = (outs EmitPy_Type:$result);
 
   let hasVerifier = 1;
-  let assemblyFormat = "$value `[` $index `]` attr-dict `:` functional-type(operands, results)";
+  let assemblyFormat = "$container `[` $index `]` attr-dict `:` functional-type(operands, results)";
 }
 
 def EmitPy_GetAttrOp : EmitPy_Op<"get_attr", [Pure, PyExpressionInterface]> {
@@ -316,19 +316,15 @@ def EmitPy_AssignOp : EmitPy_Op<"assign",
 
     ```python
     # Code emitted:
-<<<<<<< HEAD
     target = value
     ...
     dict[5] = value
-=======
-    _CONST_EVAL_CACHE[5] = tensors
->>>>>>> main
     ```
   }];
 
   let arguments = (ins
     Arg<EmitPy_Type, "target">:$target,
-    Arg<EmitPy_Type, "value to set">:$value
+    Arg<EmitPy_Type, "value to assign">:$value
   );
 
   let hasVerifier = 1;
--- a/lib/Dialect/EmitPy/IR/EmitPyOps.cpp
+++ b/lib/Dialect/EmitPy/IR/EmitPyOps.cpp
@@ -577,9 +577,9 @@ LogicalResult LiteralOp::verify() {
 //===----------------------------------------------------------------------===//
 
 LogicalResult SubscriptOp::verify() {
-  Type valueType = getValue().getType();
+  Type valueType = getContainer().getType();
   Type indexType = getIndex().getType();
-  if (!isa<DictType>(valueType) && isa<StringType>(indexType)) {
+  if (isa<StringType>(indexType) && !isa<DictType>(valueType)) {
     return emitOpError() << "cannot use string index on non-dict type "
                          << valueType;
   }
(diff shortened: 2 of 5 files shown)


include/ttmlir/Dialect/EmitPy/IR/EmitPyOps.td
```diff
@@ -112,24 +112,33 @@ def EmitPy_LiteralOp : EmitPy_Op<"literal", [Pure, PyExpressionInterface]> {
...
+
+    Supports reading from any subscriptable object (arrays, dicts, etc.)
+    with integer or string keys.
 
     Example:
 
     ```mlir
-    %0 = emitpy.literal "0" : index
-    %1 = emitpy.subscript %arg0[%0] : (!emitpy.opaque<"[ttnn.Tensor]">, index) -> !emitpy.opaque<"ttnn.Tensor">
+    %dict = emitpy.global_statement @_CONST_EVAL_CACHE : !emitpy.dict
```
sdjordjevicTT, line 127: We can keep the array indexing example as well.

include/ttmlir/Dialect/EmitPy/IR/EmitPyOps.td
```diff
@@ -297,63 +306,38 @@ def EmitPy_CreateDictOp : EmitPy_Op<"create_dict", [Pure]> {
   }];
 }
 
-def EmitPy_SetValueForDictKeyOp : EmitPy_Op<"set_value_for_dict_key"> {
-  let summary = "Set dictionary value for a given key";
+def EmitPy_SetItemOp : EmitPy_Op<"set_item",
```
sdjordjevicTT: Instead of this op, can we improve the EmitPyAssign op to be more like the emitC one, so we can reuse it in this case?

amilovanovicTT: That is what I wanted to do first, but then it made more sense to me that the op was not producing a result
if we want it to emit sth like `dict["key"] = value`. On the other hand, if we want the op to emit a simple assignment `x = value`, it should return a result. 
However, we can make it return the value in both cases.

amilovanovicTT: Actually, maybe it's best to refactor EmitPyAssign to not have a result, since in Python `x = 5` is just a statement, so it's not returning any result.

lib/Dialect/EmitPy/IR/EmitPyOps.cpp
```diff
@@ -592,6 +593,20 @@ LogicalResult LiteralOp::verify() {
...
 }
 
+//===----------------------------------------------------------------------===//
+// SubscriptOp
+//===----------------------------------------------------------------------===//
+
+LogicalResult SubscriptOp::verify() {
+  Type valueType = getValue().getType();
+  Type indexType = getIndex().getType();
+  if (!isa<DictType>(valueType) && isa<StringType>(indexType)) {
```
svuckovicTT: nit: this feel a little more natural to me
```suggestion
  if (isa<StringType>(indexType) && !isa<DictType>(valueType)) {
```

nsumrakTT (approved):

include/ttmlir/Dialect/EmitPy/IR/EmitPyOps.td
```diff
@@ -112,49 +112,39 @@ def EmitPy_LiteralOp : EmitPy_Op<"literal", [Pure, PyExpressionInterface]> {
...
+    # Code emitted:
+    var_0 = subscript_arg[0]
+    ...
+    tensors = _CONST_EVAL_CACHE[5]
     ```
   }];
 
   let arguments = (ins
-    EmitPy_OpaqueType:$value);
+    Arg<AnyTypeOf<[EmitPy_OpaqueType, EmitPy_DictType]>, "the value to subscript">:$value,
```
svuckovicTT: Maybe call this `collection` or `container` - `value` is too vague.

include/ttmlir/Dialect/EmitPy/IR/EmitPyOps.td
```diff
@@ -297,63 +287,48 @@ def EmitPy_CreateDictOp : EmitPy_Op<"create_dict", [Pure]> {
...
+    ...
+    dict[5] = value
     ```
   }];
 
   let arguments = (ins
-    EmitPy_DictType:$dict,
-    AnyTypeOf<[Index, EmitPy_OpaqueType]>:$key
+    Arg<EmitPy_Type, "target">:$target,
+    Arg<EmitPy_Type, "value to set">:$value
```
svuckovicTT: nit:
```suggestion
    Arg<EmitPy_Type, "value to assign">:$value
```

lib/Dialect/EmitPy/IR/EmitPyOps.cpp
amilovanovicTT: Agree, done

include/ttmlir/Dialect/EmitPy/IR/EmitPyOps.td
amilovanovicTT: Done

include/ttmlir/Dialect/EmitPy/IR/EmitPyOps.td
amilovanovicTT: Done

PR #6824, opened 2026-01-29
Fix convertXlaSdyToSdyDictionary to handle replicated 0-dim scalar 

jameszianxuTT:
### Ticket
tt-xla uplift regression related to #6638 

### Problem description
After changes in #6638, XlaSdyToSdy conversion pass fails on scalars like:
`%arg16: tensor<i1> {mhlo.frontend_attributes = {xla.sdy.sharding = "#sdy.sharding<@mesh, []>"}, mhlo.sharding = "{replicated}"`

because it drops both the  xla.sdy.sharding due to empty dim shardings "[]", and the mhlo.sharding as well. This results in the ApplyArgumentShardStatus pass to annotate the input arg as unsharded rather than presharded, as presharded tensors must have a sharding annotation attached and both were stripped by the XlaSdyToSdy pass. 

This results in a failure at runtime since this is a presharded input scalar that is moved to device prior to execution, but a replicate mesh_shard op is applied, which is decomposed into a ttnn.distribute_tensor op, which only works on host tensors.

### What's changed
- Permit empty dim shardings in convertXlaSdyToSdyDictionary to allow conversion "xla.sdy.sharding ... <@mesh, []>" to "sdy.sharding <@mesh, []>" instead of dropping the attr entirely. 
- Add test case

### Checklist

Note - Credit to @hshahTT for help with debug and resolution

### DIFF
--- a/test/ttmlir/Dialect/StableHLO/xla_sdy_to_sdy/round_trip_attributes.mlir
+++ b/test/ttmlir/Dialect/StableHLO/xla_sdy_to_sdy/round_trip_attributes.mlir
@@ -24,6 +24,19 @@ module @ShardingConstraint attributes {mhlo.cross_program_prefetches = [], mhlo.
   }
 }
 
+// -----
+
+// CHECK-LABEL: module @ScalarShardingConstraint
+// Test scalar tensor (rank 0) with empty dimension shardings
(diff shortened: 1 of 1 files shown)


jameszianxuTT:
try fix regressions in #6638

Issue #6782, opened 2026-01-28
TTNN permute folding


mvasiljevicTT:
Current implementation of SDPA in [TTNNFusing](https://github.com/tenstorrent/tt-mlir/blob/main/lib/Dialect/TTNN/Transforms/TTNNFusing.cpp) adds new permutes. 

We need folding patterns for permute ttnn op like there are for [reshape](https://github.com/tenstorrent/tt-mlir/blob/f3c90ca2114e51293a1850bd645c05c2bf143fa0/lib/Dialect/TTNN/IR/TTNNOps.cpp#L1560).

This will also be needed when EIO is lowered to ttnn.

mvasiljevicTT:
@elhewaty, if you are interested this would be a nice issue for you:

- When two consecutive permutes come one after another, we should fold them into one. 
- If permutation is identity, it should be removed.

You can find more about folding patterns in general [here](https://mlir.llvm.org/docs/Canonicalization/).

Ping me if you have any questions!

elhewaty:
Great, please assign me.

elhewaty:
@mvasiljevicTT , would you review this https://github.com/tenstorrent/tt-mlir/pull/6825 when you have time?

elhewaty:
@mvasiljevicTT Would you please assign the next task?

elhewaty:
> @mvasiljevicTT Would you please assign the next task?

@mvasiljevicTT Ping

mvasiljevicTT:
@elhewaty if you are interested, [this](https://github.com/tenstorrent/tt-mlir/issues/5952) might be an interesting issue for you.

PR #6825, opened 2026-01-30
Fold identity/consecutive permutes

elhewaty:
Fixes: https://github.com/tenstorrent/tt-mlir/issues/6782

### DIFF
--- a/lib/Dialect/TTNN/IR/TTNNOps.cpp
+++ b/lib/Dialect/TTNN/IR/TTNNOps.cpp
@@ -3340,8 +3340,7 @@ mlir::OpFoldResult foldConsecutivePermutes(mlir::tt::ttnn::PermuteOp op) {
 
   mlir::Value newInput = permuteOperand.getInput();
   op->setOperand(0, newInput);
-  op.setPermutationAttr(
-      mlir::DenseI64ArrayAttr::get(op.getContext(), combinedPerm));
+  op.setPermutation(combinedPerm);
 
   return op.getResult();
 }
--- a/test/ttmlir/Dialect/TTNN/data_movement/permute/permute_folding.mlir
+++ b/test/ttmlir/Dialect/TTNN/data_movement/permute/permute_folding.mlir
@@ -11,16 +11,28 @@ module {
     return %0 : tensor<2x3x4xf32>
   }
 
-  func.func @consecutive_permute(%arg0: tensor<2x3x4xf32>) -> tensor<4x2x3xf32> {
+  func.func @consecutive_permute(%arg0: tensor<2x3x4xf32>) -> tensor<4x3x2xf32> {
     // CHECK: ttnn.permute
-    // CHECK-SAME: permutation = array<i64: 2, 0, 1>
+    // CHECK-SAME: permutation = array<i64: 2, 1, 0>
     // CHECK-NOT: ttnn.permute
     // CHECK-NEXT: return
     %0 = "ttnn.permute"(%arg0) <{permutation = array<i64: 2, 0, 1>}> :
         (tensor<2x3x4xf32>) -> tensor<4x2x3xf32>
-    %1 = "ttnn.permute"(%0) <{permutation = array<i64: 0, 1, 2>}> :
-        (tensor<4x2x3xf32>) -> tensor<4x2x3xf32>
-    return %1 : tensor<4x2x3xf32>
+    %1 = "ttnn.permute"(%0) <{permutation = array<i64: 0, 2, 1>}> :
+        (tensor<4x2x3xf32>) -> tensor<4x3x2xf32>
+    return %1 : tensor<4x3x2xf32>
+  }
+
+  func.func @consecutive_permute_4d(%arg0: tensor<5x2x3x4xf32>) -> tensor<4x3x2x5xf32> {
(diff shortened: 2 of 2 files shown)


lib/Dialect/TTNN/IR/TTNNOps.cpp
```diff
@@ -3314,6 +3314,77 @@ ::mlir::LogicalResult mlir::tt::ttnn::PermuteOp::verify() {
...
+static mlir::OpFoldResult foldIdentityPermute(mlir::tt::ttnn::PermuteOp op) {
+  llvm::ArrayRef<int64_t> perm = op.getPermutation();
+
+  bool isIdentity = true;
+  for (size_t i = 0; i < perm.size(); ++i) {
+    if (perm[i] != static_cast<int64_t>(i)) {
+      isIdentity = false;
+      break;
+    }
+  }
```
mvasiljevicTT: You could use `llvm::equal` here to compare permutation with sequence 0...perm.size()

lib/Dialect/TTNN/IR/TTNNOps.cpp
```diff
@@ -3314,6 +3314,77 @@ ::mlir::LogicalResult mlir::tt::ttnn::PermuteOp::verify() {
...
+    return nullptr;
+  }
+
+  llvm::ArrayRef<int64_t> firstPerm = permuteOperand.getPermutation();
+  llvm::ArrayRef<int64_t> secondPerm = op.getPermutation();
+
+  llvm::SmallVector<int64_t> combinedPerm(firstPerm.size());
+  for (size_t i = 0; i < secondPerm.size(); ++i) {
+    combinedPerm[i] = firstPerm[secondPerm[i]];
+  }
```
mvasiljevicTT: You can use `ttmlir::utils::applyPermutation` here

lib/Dialect/TTNN/IR/TTNNOps.cpp
```diff
@@ -3314,6 +3314,77 @@ ::mlir::LogicalResult mlir::tt::ttnn::PermuteOp::verify() {
...
+
+  auto originalType =
+      mlir::cast<mlir::RankedTensorType>(op.getResult().getType());
+  auto inputType = mlir::cast<mlir::RankedTensorType>(newInput.getType());
+
+  llvm::SmallVector<int64_t> newShape;
+  newShape.reserve(combinedPerm.size());
+  for (int64_t idx : combinedPerm) {
+    newShape.push_back(inputType.getDimSize(idx));
+  }
```
mvasiljevicTT: Permuting input shape with combined permutation `(p \cdot q) (x)` you will get the same shape as `p(q(x))` that was output shape of second permutation. 

Since permutation ops are already validated to have valid input and output shape for their permutation, you can trust the output shape of the outer permutation to be the shape of combined permutation, and therefore not change it at all. 

If that wasn't the case, this folding pattern wouldn't be valid as the next op would expect different shape to come out of these permutes. 
When writing patterns, either folders or complicated fusion patterns, you should always keep in mind that the rest of the graph should not be aware of that, and any branch (input/output) connecting that subgraph (pattern) to the rest of the graph should not be changed (must keep shape and dtype etc).

lib/Dialect/TTNN/IR/TTNNOps.cpp
```diff
@@ -3316,6 +3316,47 @@ ::mlir::LogicalResult mlir::tt::ttnn::PermuteOp::verify() {
...
+  if (!permuteOperand->hasOneUse()) {
+    return nullptr;
+  }
+
+  llvm::SmallVector<int64_t> combinedPerm = ttmlir::utils::applyPermutation(
+      permuteOperand.getPermutation(), op.getPermutation());
+
+  mlir::Value newInput = permuteOperand.getInput();
+  op->setOperand(0, newInput);
+  op.setPermutationAttr(
```
mvasiljevicTT: You can use `op.setPermutation(combinedPerm)` here. It builds the attribute from `ArrayRef`, which is compatible with `SmallVector`.

test/ttmlir/Dialect/TTNN/data_movement/permute/permute_folding.mlir
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  func.func @consecutive_permute(%arg0: tensor<2x3x4xf32>) -> tensor<4x2x3xf32> {
+    // CHECK: ttnn.permute
+    // CHECK-SAME: permutation = array<i64: 2, 0, 1>
+    // CHECK-NOT: ttnn.permute
+    // CHECK-NEXT: return
+    %0 = "ttnn.permute"(%arg0) <{permutation = array<i64: 2, 0, 1>}> :
+        (tensor<2x3x4xf32>) -> tensor<4x2x3xf32>
+    %1 = "ttnn.permute"(%0) <{permutation = array<i64: 0, 1, 2>}> :
```
mvasiljevicTT: Could you please add examples where none of the permutes are identity? In this test the second permute will get folded and you won't actually test consecutive permute folding.

elhewaty: Sorry, I missed this. I will update it. BTW, can you assign the next task?

Issue #6710, opened 2026-01-22
[ttnn-jit] ttmlir import paths are being set as top level python modules


vtangTT:
### Describe the bug

https://github.com/tenstorrent/tt-mlir/blob/e92de08990acb764bf922eca80b0b9f023b91606/tools/ttnn-jit/_src/__init__.py#L11

Originally done as a workaround to get the wheel up and running as the wheel bundles its own ttmlir bindings under `ttnn_jit/runtime`. By setting as top level modules, imports could stay as `import ttmlir` when installed as a wheel into tt-metal.

However, this will cause a double import problem, as we are creating an alias in `sys.modules` but the underlying module object has been loaded under a separate identity (the `ttnn-jit/runtime` one)

Will inevitably blow up in some c++ extension or mlir internal if they consider the module name or identity.

### Expected behavior

As per @tapspatel, we should be depending on the ttmlir wheel instead of installing it into our own ttnn-jit wheel, so this top level `import ttmlir` is available by default

vtangTT:
@tapspatel let us know if you do push a temporary workaround solution as discussed 🙏

tapspatel:
Okay so the temporary solution didn't work :( I'll need this fixed before I can merge in that PR. Pretty much any subproject of ttmlir bindings should install it as a dependency and not use the src files.

skhatriTT:
@vtangTT does the linked PR https://github.com/tenstorrent/tt-mlir/pull/6839 close this issue?

vtangTT:
@skhatriTT sry closed it per-emptively. Yes the linked PR closes this issue

Issue #6358, opened 2025-12-17
[ttnn.jit] editable install/faster dev workflow


vtangTT:
### Is your feature request related to a problem? Please describe

The current ttnn.jit dev flow involves installing the release wheel into our pip env. While this reflects end-user env and should be the way to go on CI, locally, it adds probably around 2 mins to build time. It also does not allow you to edit python files without a rebuild (one of the main benefits of using python)

### Describe the solution you'd like

 `pip install -e .` [docs](https://setuptools.pypa.io/en/latest/userguide/development_mode.html)

Note that this solution do not reflect release env at all. Ideally, there is a dev flow that is basically just _release but live_. While this is what I hoped editable installs were, they are not. Editable installs involve a completely different pip flow that can break separately from normal wheel builds

The best solution is to set up the editable pip install flow for local dev use, and have a **VERY** simple smoke test for it running on a nightly pipeline. Something as simple as `python -c "import ttnn_jit"`

### Describe alternatives you've considered

`add_mlir_python_modules`

Solution #2 already exists in the repo today, guarded by `USE_TTNN_JIT_WHEEL` cmake arg.

vtangTT:
here's a branch with some wip, not necessarily the way to go: https://github.com/tenstorrent/tt-mlir/pull/6347/files

vtangTT:
deprio, repopen if needed

PR #6839, opened 2026-01-30
[ttnn-jit] Wheel depends on ttmlir wheel

vtangTT:
### Ticket
https://github.com/tenstorrent/tt-mlir/issues/6710

### Problem description
ttmlir source files were bundled into the ttnn-jit wheel, causing the above linked issue. 

### What's changed
- ttnn-jit wheel now points to ttmlir wheel as a install dep; cmake flow will build + install ttmlir wheel before jit wheel
- cmake flow + target to build ttmlir wheel
- ttmlir wheel no longer packages ttrt (this was never intended)

### Checklist

### DIFF
--- a/python/CMakeLists.txt
+++ b/python/CMakeLists.txt
@@ -251,12 +251,26 @@ add_custom_command(
   COMMAND TTMLIR_DEV_BUILD=ON python -m build --wheel
   COMMAND ${CMAKE_COMMAND} -E touch ${CMAKE_CURRENT_BINARY_DIR}/.ttmlir-wheel-built
   WORKING_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}"
-  DEPENDS TTMLIRPythonModules
+  DEPENDS
+    TTMLIRPythonModules
+    TTMLIRPythonCAPI
+    TTMLIRCAPI
+    TTMLIRPythonExtensions
+    ${CMAKE_CURRENT_SOURCE_DIR}/TTMLIRModule.cpp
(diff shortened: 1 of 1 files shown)


python/CMakeLists.txt
```diff
@@ -240,7 +240,37 @@ if(APPLE)
...
+  COMMAND python -m pip install build
+  COMMAND TTMLIR_DEV_BUILD=ON python -m build --wheel
+  COMMAND ${CMAKE_COMMAND} -E touch ${CMAKE_CURRENT_BINARY_DIR}/.ttmlir-wheel-built
+  WORKING_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}"
+  DEPENDS
+    TTMLIRPythonModules
+    TTMLIRPythonCAPI
+    TTMLIRCAPI
+    TTMLIRPythonExtensions
+    ${CMAKE_CURRENT_SOURCE_DIR}/TTMLIRModule.cpp
```
vtangTT, line 259: for some reason it seems `TTMLIRPythonExtensions` is not enough, as I can change `TTMLIRModule.cpp` and the ttmlir wheel will not rebuild 👎

nsumrakTT (approved):
Why we build wheel instead of using editable variant, as all the components exists inside install/build directories which are already transferred between build and test machines? (just curious)

vtangTT:
> Why we build wheel instead of using editable variant, as all the components exists inside install/build directories which are already transferred between build and test machines? (just curious)

We have an open [issue](https://github.com/tenstorrent/tt-mlir/issues/6358) for editable install for jit wheel. 

I wasn't aware the ttmlir wheel editable works?

Issue #6505, opened 2025-12-08
Fix redundant all_to_all ops when resharding a fully replicated tensor


hshahTT:
### Describe the bug

Parent issue: https://github.com/tenstorrent/tt-xla/issues/1000

Currently when Shardy inserts a reshard on a fully replicated tensor (for ex, when a 2D tensor with sharding `{}, {}` gets resharded to `{}, {"model"}`), it gets lowered to a `sdy.all_slice` op, which gets naively lowered to a `reshape` --> `sdy.all_to_all` --> `slice` pattern of ops.

The intermediate `sdy.all_to_all` is unneeded since the original tensor was already fully replicated, and introduces significant memory overhead in the form of intermediate tensors.

hshahTT:
Talked with @ntarafdar and @AleksKnezevic about this. There's a [mesh_partition](https://github.com/tenstorrent/tt-metal/blob/bc87b76d58b111d2920cd3e909558f20f1b0ff6f/ttnn/cpp/ttnn/operations/ccl/mesh_partition/mesh_partition_nanobind.cpp#L18) op within ttnn that can be used as a direct replacement for `sdy.all_slice`.

The proposed approach is:

1. Add the `mesh_partition` op to both [TTIROps](https://github.com/tenstorrent/tt-mlir/blob/main/include/ttmlir/Dialect/TTIR/IR/TTIROps.td) and [TTNNOps](https://github.com/tenstorrent/tt-mlir/blob/main/include/ttmlir/Dialect/TTNN/IR/TTNNOps.td) within tt-mlir.
2. Within the [ShardyToStableHLOAllSliceOpRewritePattern](https://github.com/tenstorrent/tt-mlir/blob/c5544f54308685c08a1f33df0b05be3a8a098356/lib/Dialect/StableHLO/Transforms/ShardyCCLToStableHLOCCLPatterns.cpp#L385) wrap the current rewrite pattern (`reshape` -> `all_to_all` -> `slice`) inside a named composite op.
3. Add a [composite op legalization pattern](https://github.com/tenstorrent/tt-mlir/blob/c5544f54308685c08a1f33df0b05be3a8a098356/lib/Conversion/StableHLOToTTIR/StableHLOLegalizeCompositePass.cpp#L243) for the named composite op when lowering from SHLO to TTIR.
[...]

PR #6850, opened 2026-02-02
Optimize `sdy.all_slice` lowering when input tensor is fully replicated 

hshahTT:
### Ticket
Closes https://github.com/tenstorrent/tt-mlir/issues/6505

### Problem description
The current lowering pattern for `sdy.all_slice` into SHLO includes a `stablehlo.all_to_all` op in all cases. This is unnecessary for fully replicated tensors and incurs a large memory and time penalty during execution.

#### Main changes
- Introduced [ttnn.mesh_partition](https://docs.tenstorrent.com/tt-metal/latest/ttnn/ttnn/api/ttnn.mesh_partition.html) op in the TTIR and TTNN dialects
- Introduced a new lowering pattern for `sdy.all_slices` ops when the input is a fully replicated tensor:
    1. Wrap the `sdy.all_slice` op inside a `stablehlo.composite "sdy.all_slice"` named op
    2. Within the `StableHLOLegalizeCompositePass` convert each composite op to a series of `ttir.mesh_partition` ops (one for every sliced dimension).
    3. Convert the `ttir.mesh_partition` ops to `ttnn.mesh_partition` ops, and further to Flatbuffer.

**Caveat:** There's a workaround to force the input and output tensors of the `ttnn.mesh_partition` ops to be in ROW_MAJOR format. This is due to an [issue](https://github.com/tenstorrent/tt-metal/issues/37676) with the op's implementation in tt-metal. It will be removed once that's fixed.

#### Ancillary changes
- Renamed and moved constants related to the `ReoutlineComposite` and `FlattenComposite` passes to a new `StableHLOUtils.h` file.
- Moved the function that created a new private function within the `ReoutlineComposite` pass to the StableHLO utils file so it can be reused.
- Removed some unnecessary includes.

### Checklist

### DIFF
--- a/runtime/lib/ttnn/operations/ccl/mesh_partition.cpp
+++ b/runtime/lib/ttnn/operations/ccl/mesh_partition.cpp
@@ -16,7 +16,7 @@ void run(const ::tt::target::ttnn::MeshPartitionOp *op,
   const ::ttnn::Tensor &input = tensorPool.getTTNNTensorAndValidate(op->in());
   LOG_ASSERT(
       input.storage_type() == ::ttnn::StorageType::DEVICE,
-      "Input of mesh_partition must be DEVICE. id:", op->in()->global_id());
+      "Input of mesh_partition must be on DEVICE. id:", op->in()->global_id());
   int32_t dim = op->dim();
   std::optional<uint32_t> clusterAxis = op->cluster_axis();
   std::optional<::ttnn::MemoryConfig> outputMemoryConfig =
--- a/include/ttmlir/Dialect/TTNN/IR/TTNNWorkaroundsPass.h
+++ b/include/ttmlir/Dialect/TTNN/IR/TTNNWorkaroundsPass.h
@@ -248,8 +248,10 @@ public:
   static TTNNOperandsWorkarounds
   createMeshShardOpOperandsWorkarounds(ttcore::MeshShardType shardType);
 
-  // Create workarounds for mesh partition op operands. The input tensor is
-  // always in row-major layout.
+  // Create workarounds for mesh partition op operands. The input and output
+  // tensors are always in row-major layout.
+  // TODO (hshah): Remove once
+  // https://github.com/tenstorrent/tt-metal/issues/37676 is fixed.
   static TTNNOperandsWorkarounds createMeshPartitionOpOperandsWorkarounds();
 
   // Create workarounds for scatter op operands.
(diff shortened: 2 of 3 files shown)


lib/Conversion/StableHLOToTTIR/StableHLOLegalizeCompositePass.cpp
```diff
@@ -290,6 +293,98 @@ class TenstorrentLayerNormConversionPattern
...
+        return rewriter.notifyMatchFailure(srcOp, "Invalid mesh axis index.");
+      }
+      // Compute new shape for the result tensor, with original dimension
+      // divided by mesh axis size
+      int64_t meshAxisSize = meshShape[clusterAxes[i]];
+      if (newShape[tensorDims[i]] < 0 ||
+          newShape[tensorDims[i]] % meshAxisSize != 0) {
+        return rewriter.notifyMatchFailure(
+            srcOp,
+            "Dimension size must be static and divisible by mesh axis size.");
```
jameszianxuTT, line 372: What does static mean in this context? Is it possible to receive a -1 in newShape[tensorDims[i]] to mean a dynamic shape?

runtime/lib/ttnn/operations/ccl/mesh_partition.cpp
```diff
@@ -0,0 +1,30 @@
...
+#include <optional>
+namespace tt::runtime::ttnn::operations::ccl {
+
+void run(const ::tt::target::ttnn::MeshPartitionOp *op,
+         ProgramContext &context) {
+  ProgramTensorPool &tensorPool = context.getTensorPool();
+  const ::ttnn::Tensor &input = tensorPool.getTTNNTensorAndValidate(op->in());
+  LOG_ASSERT(
+      input.storage_type() == ::ttnn::StorageType::DEVICE,
+      "Input of mesh_partition must be DEVICE. id:", op->in()->global_id());
```
jameszianxuTT: nit: must be "on" DEVICE

lib/Conversion/StableHLOToTTIR/StableHLOLegalizeCompositePass.cpp
hshahTT: It's possible to receive a value of [ShapedType::kDynamic](https://github.com/llvm/llvm-project/blob/0286641f5c607e7b45510f699a82535a327c857b/mlir/include/mlir/IR/BuiltinTypeInterfaces.td#L214C30-L214C38) (which is `INT_MIN`) which means the dim is dynamic.

I'll actually just replace that first check with `newShape[tensorDims[I]] == ShapedType::kDynamic` so it's more explicit.

PR #6853, opened 2026-02-02
Fix Clamp scalar OP to also allow integer attributes

### LINKED ISSUE #6852: Clamp scalar crashes in runtime with unexpected dtype error
Clamp scalar operation is registered in runtime to only use float attributes, which is wrong for integer tensors. This causes a type mismatch asserting to fire in runtime.

sgligorijevicTT:
### Ticket
closes #6852

### Problem description
Clamp scalar takes min and max as floats, this forces the op to return float tensors even with int inputs, this causes an IR-runtime mismatch which causes an assert to fire.

### What's changed
Attributes are now a union of int and float

### Checklist

### DIFF
--- a/test/python/golden/ttir_ops/eltwise/test_ttir_ternary.py
+++ b/test/python/golden/ttir_ops/eltwise/test_ttir_ternary.py
@@ -155,15 +155,13 @@ def test_ternary_eltwise_ops_implicit_broadcast(
 @pytest.mark.parametrize("shape", [(64, 128)], ids=shape_str)
 @pytest.mark.parametrize("max_arg,min_arg", [(0.8, -0.5)])
 @pytest.mark.parametrize("target", ["ttnn", "ttmetal"])
-def test_clamp_scalar(
-    shape: Shape, max_arg, min_arg, dtype: torch.dtype, target: str, request, device
-):
+def test_clamp_scalar(shape: Shape, max_arg, min_arg, target: str, request, device):
     def module_clamp_scalar(builder: TTIRBuilder):
-        @builder.func([shape], [dtype])
+        @builder.func([shape], [torch.float32])
         def clamp_scalar(
             in0: Operand, builder: TTIRBuilder, unit_attrs: Optional[List[str]] = None
         ):
-            input_tensor = torch.rand(shape, dtype=dtype) * 2 - 1
+            input_tensor = torch.rand(shape, dtype=torch.float32) * 2 - 1
             builder.set_goldens(inputs={in0: input_tensor})
             return builder.clamp_scalar(
                 in0, max_arg=max_arg, min_arg=min_arg, unit_attrs=unit_attrs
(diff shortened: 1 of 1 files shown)


include/ttmlir/Target/TTNN/operations/eltwise.fbs
```diff
@@ -189,8 +189,8 @@ enum EltwiseUnaryCompositeOpType: uint32 {
 }
 
 table ClampScalarOpParams {
-  min: float;
-  max: float;
+  min: RhsParams;
```
sdjordjevicTT: What is this? :D Can we rename this to something more meaningful? Something like IntOrFloatUnion... Or something more descriptive than RhsParams...

lib/Dialect/TTIR/IR/TTIROps.cpp
```diff
@@ -202,15 +201,61 @@ void mlir::tt::ttir::ClampScalarOp::getCanonicalizationPatterns(
       return mlir::failure();
     }
 
+    // Helper to convert attribute to double
```
sdjordjevicTT: Can we move this to some helper utils file? This seems useful not only for this op...

lib/Dialect/TTIR/IR/TTIROps.cpp
```diff
@@ -202,15 +201,61 @@ void mlir::tt::ttir::ClampScalarOp::getCanonicalizationPatterns(
...
+    }
+
     // Calculate the tightest bounds
-    auto newMin = std::max(producerMin, consumerMin);
-    auto newMax = std::min(producerMax, consumerMax);
+    double newMinVal = std::max(*producerMinVal, *consumerMinVal);
+    double newMaxVal = std::min(*producerMaxVal, *consumerMaxVal);
+
+    // Helper to create attribute from double
+    auto doubleToAttr = [&](double value,
```
sdjordjevicTT: Same for this...

lib/Conversion/TTNNToEmitC/TTNNToEmitC.cpp
```diff
@@ -378,12 +378,32 @@ class ClampOpConversionPattern
...
-        emitter.emit(srcOp.getMax()),
-        emitter.emit(std::nullopt) | emitter.getMemoryConfig(srcOp.getResult()),
-    };
+
+    llvm::SmallVector<mlir::Attribute> args;
+    args.push_back(emitter.emit(srcOp.getInput()));
+
+    if constexpr (std::is_same_v<SourceOp, mlir::tt::ttnn::ClampScalarOp>) {
+      // Helper to convert attribute to APFloat for ClampScalarOp
+      auto attrToAPFloat = [](mlir::Attribute attr) -> llvm::APFloat {
```
sdjordjevicTT: This as well seems like a generic utility function that all the repo can benefit...

lib/Conversion/TTNNToEmitC/TTNNToEmitC.cpp
```diff
@@ -378,12 +378,32 @@ class ClampOpConversionPattern
...
+        if (auto floatAttr = mlir::dyn_cast<mlir::FloatAttr>(attr)) {
+          return floatAttr.getValue();
+        }
+        if (auto intAttr = mlir::dyn_cast<mlir::IntegerAttr>(attr)) {
+          return llvm::APFloat(
+              static_cast<float>(intAttr.getValue().getSExtValue()));
+        }
+        llvm_unreachable("Unsupported attribute type for clamp");
+      };
+      args.push_back(emitter.emit(attrToAPFloat(srcOp.getMin())));
```
mtopalovicTT: Should we emit int or float attribute here based on what is in output element type? Same for EmitPy conversion.

include/ttmlir/Target/TTNN/operations/eltwise.fbs
sgligorijevicTT: Agreed. For context this branch was spun out of my hacks to make smth work, my bad for forgetting to clean up the code.

I took this opportunity to unify that type and another identical type as `NumberType`

lib/Conversion/TTNNToEmitC/TTNNToEmitC.cpp
sgligorijevicTT: Yup good point, fixed.

include/ttmlir/Target/TTNN/operations/eltwise.fbs
sdjordjevicTT: Great, thanks a lot!

jgrimTT (approved):

PR #6860, opened 2026-02-02
[d2m] set reductions keep_dim=True and add reshapes to remove dim 

### LINKED ISSUE #6645: [d2m] d2m.sum keep_dim=false support
keep_dim=false fails:
```mlir
func.func @component_1(%22: tensor<32x1x8192xf32>) -> tensor<32x1xf32> {
  %25 = "ttir.sum"(%22) <{dim_arg = [2 : i32], keep_dim = false}> : (tensor<32x1x8192xf32>) -> tensor<32x1xf32>
  return %25 : tensor<32x1xf32>
}
```
```
Assertion `op.getKeepDimAttr().getValue() && "expected default keep_dim=true"' failed.
```

Either:
- We add explicit support for this in the d2m conversion
- We have a rank normalization path that changes keep_dim=true and then uses reshape to drop a rank

The former might be slightly preferable.

dlokeTT:
### Ticket
Closes #6645 

### Problem description
`keep_dim=False` on reductions is unsupported in D2M. Current solution to get these working thru the pipeline initially is to add reshapes after removing the dim and setting `keep_dim=True` to compile thru D2M pipeline

### What's changed
Added a pass that goes thru and sets `keep_dim=True` on reductions where they are set to false and adds reshapes after that remove the dim
Added lit testing for this pass
Added `keep_dim=False` to builder tests for metal reductions

### Checklist

### DIFF
--- a/include/ttmlir/Dialect/TTIR/Transforms/Passes.td
+++ b/include/ttmlir/Dialect/TTIR/Transforms/Passes.td
@@ -323,6 +323,26 @@ def TTIRDecomposeComplexReshape: Pass<"ttir-decompose-complex-reshape", "::mlir:
   }];
 }
 
+def TTIRReductionForceKeepDim: Pass<"ttir-reduction-force-keep-dim", "::mlir::ModuleOp">
+{
+  let summary = "Force all reductions to use keep_dim=true.";
+  let description = [{
+    This pass converts all reduction operations with keep_dim=false to
+    keep_dim=true, inserting a reshape after the reduction to restore the
+    original output shape.
+
+    Example:
+      Before:
+        %0 = "ttir.sum"(%input) {dim_arg = [1], keep_dim = false} : (tensor<2x3x4xf32>) -> tensor<2x4xf32>
+
+      After:
+        %0 = "ttir.sum"(%input) {dim_arg = [1], keep_dim = true} : (tensor<2x3x4xf32>) -> tensor<2x1x4xf32>
+        %1 = "ttir.reshape"(%0) <{shape = [2, 4]}> : (tensor<2x1x4xf32>) -> tensor<2x4xf32>
+  }];
+
+  let dependentDialects = ["mlir::tt::ttir::TTIRDialect"];
+}
+
 def TTIRMultiDeviceTensorAnnotation: Pass<"ttir-multi-device-tensor-annotation", "::mlir::ModuleOp">
 {
   let summary = "TTIR multi-device tensor annotation pass.";
--- a/lib/Dialect/TTIR/Transforms/ReductionForceKeepDim.cpp
+++ b/lib/Dialect/TTIR/Transforms/ReductionForceKeepDim.cpp
@@ -1,0 +1,95 @@
+// SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#include "ttmlir/Dialect/TTIR/IR/TTIROps.h"
+#include "ttmlir/Dialect/TTIR/Transforms/Passes.h"
+
+#include "mlir/Transforms/GreedyPatternRewriteDriver.h"
+
+namespace mlir::tt::ttir {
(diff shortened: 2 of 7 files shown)


test/python/golden/test_metal_reductions.py
```diff
@@ -84,6 +84,11 @@ def test_sum_3d(
     request,
     device,
 ):
+    if len(dim_arg) >= 2 and not keep_dim:
```
dlokeTT, line 87: These issues are related to a general affine map composition bug that happens when TM's are involved with compute ops and involve shapes with rank > 2. Did some debugging with @vwellsTT  and this issue seems to go away after https://github.com/tenstorrent/tt-mlir/pull/6902

vwellsTT: I'm hoping to merge my PR today, perhaps we can wait and avoid the issue then?

include/ttmlir/Dialect/TTIR/Transforms/Passes.td
```diff
@@ -300,6 +300,26 @@ def TTIRDecomposeComplexReshape: Pass<"ttir-decompose-complex-reshape", "::mlir:
   }];
 }
 
+def TTIRReductionKeepDim: Pass<"ttir-reduction-keep-dim", "::mlir::ModuleOp">
```
vwellsTT: I think this pass should be renamed to something clearer, like DecomposeKeepDim or something

lib/Dialect/TTIR/Transforms/ReductionForceKeepDim.cpp
```diff
@@ -0,0 +1,91 @@
...
+      return mlir::failure();
+    }
+
+    ArrayRef<int64_t> inputShape = reductionOp.getInput().getType().getShape();
+
+    // Calculate the output shape with keep_dim=true.
+    llvm::SmallVector<int64_t> keepDimShape(inputShape);
+
+    if (!reductionOp.getDimArg()) {
+      keepDimShape.assign(inputShape.size(), 1);
```
vwellsTT, line 41: shouldn't this just be the original shape, vs 1s-filled of same rank? Don't quite follow here, maybe needs a comment

lib/Dialect/TTIR/Transforms/ReductionKeepDim.cpp
```diff
@@ -0,0 +1,91 @@
...
+      for (mlir::Attribute reduceDim : reduceDims) {
+        int64_t reduceDimInt =
+            mlir::cast<mlir::IntegerAttr>(reduceDim).getInt();
+        reduceDimInt = (reduceDimInt + inputShape.size()) % inputShape.size();
+        keepDimShape[reduceDimInt] = 1;
+      }
+    }
+
+    auto originalType = reductionOp.getResult().getType();
+    auto keepDimType =
```
vwellsTT: nit: would probably call this "decompedType" or something

include/ttmlir/Dialect/TTIR/Transforms/Passes.td
dlokeTT: renamed to ReductionForceKeepDim since that's what this is doing

lib/Dialect/TTIR/Transforms/ReductionKeepDim.cpp
dlokeTT: done

lib/Dialect/TTIR/Transforms/ReductionForceKeepDim.cpp
dlokeTT: Added a comment but it should be 1s of same rank. In `TTIROps.td` for reductions it says 
```
The dim_arg attribute specifies which dimensions to reduce along. If not provided, the reduction is performed
over all dimensions, resulting in a scalar value (or a tensor with all dimensions of size 1 if keep_dim is true).
```

vwellsTT: I see, cool

PR #6863, opened 2026-02-03
[Optimizer] initial integration with d2m subgraph op

sgholamiTT:
### Ticket
Link to Github Issue

Integrate `ttnn.d2m_subgraph` (D2M fused subgraph dispatch) into the TTNN memory-layout optimizer so that layout assignment and spilling behave correctly when D2M regions sit between producer and consumer chains.

## Summary of changes

- **DFShardingPolicy**: Treat `D2MSubgraphOp` as an L1 chain boundary. When the current op is a D2M subgraph, close the current chain and open an isolated single-op chain for it so it still gets a resolved layout from the policy. When the *next* op is a D2M subgraph, do not extend the current chain across it (break the chain before the dispatch).
- **Optimizer**: (1) Apply the chosen layout to `D2MSubgraphOp` (result type, output `empty` buffer, and the callee D2M subgraph function’s types and body) via `applyChosenLayoutToD2MSubgraphOp`. (2) After `processSpillOps`, sync the D2M subgraph function’s argument and result types to the dispatch op’s *current* inputs/results via `syncD2MFuncTypesToDispatchInputs` so that if a producer was spilled to DRAM, the subgraph’s signature matches (dispatch input DRAM → subgraph arg DRAM).
- **OpConstraintValidation**: Skip OpModel-based validation for `D2MSubgraphOp` and accept any proposed output layout.

## Main decisions and assumptions

- **Chain boundary**: D2M subgraph is not fused into surrounding L1 chains; it is always a boundary. Layout for the op is still chosen (from consumer / policy) and applied; the op is just not part of a longer L1 chain for spilling/scheduling. This decision is made since it's not possible to estimate how much L1 the D2M fused ops need.
- **Layout source**: The chosen layout for `D2MSubgraphOp` comes from the same flow as other ops (legal configs / policy), not from an OpModel. Validation is skipped so any policy-chosen layout is accepted.
- **Type sync after spill**: Subgraph function types are synced to the dispatch’s inputs/results *after* `processSpillOps`. That way, if a producer chain is spilled to DRAM and the dispatch’s first operand becomes DRAM, the subgraph’s first argument type is updated to DRAM, keeping the verifier and downstream passes consistent.
- **Single result**: Only single-result `D2MSubgraphOp` is supported in the apply path; multi-result is asserted as not yet implemented.

## Limitations

- Multi-result `D2MSubgraphOp` is not yet supported in `applyChosenLayoutToD2MSubgraphOp` (assert).
- D2M subgraph has no OpModel; validation always accepts the proposed layout and does not check L1/backend constraints for the inner ops.

### Checklist
- [X] [CI green](https://github.com/tenstorrent/tt-mlir/actions/runs/21635795859) ✅


lib/Dialect/TTNN/Analysis/DFShardingPolicy.cpp
```diff
@@ -2301,6 +2301,31 @@ void DFShardingPolicy::run() {
                     ttnn::SliceStaticOp, ttnn::RotaryEmbeddingOp>(currentOp) &&
           legalConfigs.lookup(currentOp).size() > 0;
 
+      // Special handling for D2MSubgraphOp: treat as chain boundary and
+      // create a single-op chain so it gets a resolved layout from
+      // ShardSolver.
+      if (llvm::isa<ttnn::D2MSubgraphOp>(currentOp) &&
```
odjuricicTT, line 2307: Not creating this single op chain should work as well. ShardSolver does not need to resolve layouts for every op. The unresolved ops will just default to dram interleaved tile.

Did you run into issues without this?

sgholamiTT:
Closing this PR as we decided to implement this [differently](https://github.com/tenstorrent/tt-mlir/pull/7007).

PR #6864, opened 2026-02-03
[d2m] lower reshapes into reshapes + permute if needed

dlokeTT:
### Ticket
Helps with #6604 

### Problem description
Some reshapes involve transposing on the last dim so we need to be able to detect when it is required to do a permute and lower the reshape into a reshape/permute or combination of the two

### What's changed

- Added `d2m-decompose-complex-reshape` that runs at the ttir level and detects when permutes are needed and subsequently decomposes the reshape.
- `d2m-decompose-complex-reshape` handles two cases. 
    - Singleton transposes where dims of size 1 are shifted around and a permute is needed if the last dim is flipped with any other dims. e.g. `[1, 128] -> [128, 1]`
    - Flatten and swaps where the last dim of size 1 is flipped with other dims that are involved in a real reshape. e.g `[128, 4, 1] -> [1, 512]`
- Added logic that forces reshapes to be materialized before a permute in `TTIRToD2M.cpp` because otherwise we hit some assertions for affine maps when rank changing reshapes are composed with rank preserving affine maps like permute. Follow on is to identify how we can compose these because reshape is hitting this issue when other compute ops follow it as well, issue here: https://github.com/tenstorrent/tt-mlir/issues/6981
- Added builder tests for some reshapes found in models that exercise this reshape/permute path.
- Added lit tests for various edge cases detecting when reshapes need to be lowered into permutes, into reshapes + permutes or kept as is.

### Checklist


lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
```diff
@@ -1416,8 +1416,24 @@ class D2MPermuteRewriter final
         toLayoutOperandsAndResults(rewriter, {origInputs, origOutputs},
                                    /*tiled*/ true);
 
+    // Check if input comes from a view operation (StreamLayoutOp or
+    // ViewLayoutOp). If so, materialize it first to prevent view chaining that
+    // can cause affine map composition dimension mismatches when the views have
+    // different input/output ranks (e.g., reshape followed by permute).
+    Value inputValue = inputs[0];
+    if (auto defOp = inputValue.getDefiningOp()) {
```
vmilosevic: :warning: **llvm\-qualified\-auto** :warning:
```suggestion
    if (auto *defOp = inputValue.getDefiningOp()) {
```

lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
```diff
@@ -1416,8 +1416,24 @@ class D2MPermuteRewriter final
...
                                    /*tiled*/ true);
 
+    // Check if input comes from a view operation (StreamLayoutOp or
+    // ViewLayoutOp). If so, materialize it first to prevent view chaining that
+    // can cause affine map composition dimension mismatches when the views have
+    // different input/output ranks (e.g., reshape followed by permute).
+    Value inputValue = inputs[0];
+    if (auto *defOp = inputValue.getDefiningOp()) {
+      if (mlir::isa<d2m::StreamLayoutOp, d2m::ViewLayoutOp>(defOp)) {
+        // Materialize the view by creating an identity ToLayout
```
vwellsTT: nit: .

lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
```diff
@@ -1784,10 +1800,83 @@ class D2MTensorManipulationOpRewriter
...
+                                                       inputs[0], storage);
+
+    // Get the logical rank for the generic op from output layout.
+    unsigned logicalRank = layout.getLogicalShape().size();
+
+    // Capture values explicitly for the lambda.
+    Value inputOperand = stream.getResult();
+    Value outputOperand = outputs[0];
+
+    // Create a GenericOp to materialize the data movement. This prevents
```
vwellsTT: I think this is pretty redundant vs comment above func

lib/Dialect/D2M/Transforms/DecomposeComplexReshape.cpp
```diff
@@ -0,0 +1,344 @@
...
+  for (auto it = shape.rbegin(); it != shape.rend(); ++it) {
+    if (*it != 1) {
+      break;
+    }
+    ++count;
+  }
+  return count;
+}
+
+llvm::SmallVector<int64_t> swapLastTwo(llvm::ArrayRef<int64_t> shape) {
```
vwellsTT: nit: I would name swapLastTwoDims or something a bit more explicit

lib/Dialect/D2M/Transforms/DecomposeComplexReshape.cpp
```diff
@@ -0,0 +1,344 @@
...
+//
+// Examples:
+//   [1, 128] -> [128, 1]: permute (trailing 1 status differs)
+//   [128, 1] -> [1, 128]: permute (trailing 1 status differs)
+//   [32, 360] -> [32, 1, 1, 360]: NOT a permute (360 is last in both)
+//   [32, 128] -> [1, 32, 128]: NOT a permute (128 is last in both)
+//
+bool isPurePermute(llvm::ArrayRef<int64_t> inputShape,
+                   llvm::ArrayRef<int64_t> outputShape) {
+  if (inputShape.empty() || outputShape.empty()) {
```
vwellsTT: this seems like it should be an assert, we should never run into empty shapes AFAIK

lib/Dialect/TTIR/Transforms/DecomposeComplexReshape.cpp
```diff
@@ -0,0 +1,344 @@
...
+//   [128, 1] -> [1, 128]: permute (trailing 1 status differs)
+//   [32, 360] -> [32, 1, 1, 360]: NOT a permute (360 is last in both)
+//   [32, 128] -> [1, 32, 128]: NOT a permute (128 is last in both)
+//
+bool isPurePermute(llvm::ArrayRef<int64_t> inputShape,
+                   llvm::ArrayRef<int64_t> outputShape) {
+  if (inputShape.empty() || outputShape.empty()) {
+    return false;
+  }
+  if (getNonOneDims(inputShape) != getNonOneDims(outputShape)) {
```
vwellsTT, line 135: is rank mismatch itself (regardless of 1s) not an automatic disqualifier for being a permute?

lib/Dialect/TTIR/Transforms/DecomposeComplexReshape.cpp
```diff
@@ -0,0 +1,344 @@
...
+    return false;
+  }
+  return (inputShape.back() == 1) != (outputShape.back() == 1);
```
vwellsTT, line 138: Why is the 1s thing important? (I think func name should reflect you're not actually checking if this is just a permute, you're somehow checking if this is a permute with 1s--but I don't follow why 1s matter tbh.  If I have: 32x1x13 => 13x1x32, you don't count that is as "pure permute"? why?)

lib/Dialect/D2M/Transforms/DecomposeComplexReshape.cpp
```diff
@@ -0,0 +1,344 @@
...
+}
+
+// Check if singleton dimensions need to move relative to non-singleton dims.
+// Detects cases like (128, 4, 1) -> (1, 512) where a trailing 1 must become
+// a leading 1, requiring a permute before the reshape.
+//
+// Examples:
+//   [128, 4, 1] -> [1, 512]: true  (trailing 1 -> leading 1)
+//   [1, 128, 4] -> [1, 512]: false (leading 1 stays leading)
+//   [128, 1, 4] -> [1, 512]: true  (middle 1 -> leading)
```
vwellsTT: this will return false? confused

lib/Dialect/TTIR/Transforms/DecomposeComplexReshape.cpp
```diff
@@ -0,0 +1,344 @@
...
+//===----------------------------------------------------------------------===//
+
+// `refType` supplies element type and encoding for the result type.
+Value createReshape(IRRewriter &rewriter, Location loc, Value input,
+                    llvm::ArrayRef<int64_t> newShape,
+                    RankedTensorType refType) {
+  auto resultType = RankedTensorType::get(newShape, refType.getElementType(),
+                                          refType.getEncoding());
+  return rewriter
+      .create<ttir::ReshapeOp>(
```
vwellsTT, line 83: we should not be creating TTIR ops in a D2M transform--if this is meant to be run on ttir not d2m, it should be moved to TTIR dir

vwellsTT:
> Pure permutes where dims of size 1 are shifted around and a permute is needed if the last dim is flipped with any other dims. e.g. [1, 128] -> [128, 1]

why is the fact there are 1 dims important? This seems true for 2x128 => 128x2 as well to me

>Singleton permutes where the last dim of size 1 is flipped with other dims that are involved in a real reshape. e.g [128, 4, 1] -> [1, 512]

same question, not following why the singleton part matters

lib/Dialect/TTIR/Transforms/DecomposeComplexReshape.cpp
dlokeTT: I followed how `DecomposeComplexPermute.cpp` was done. I assumed that this is a transform on the ttir level that is meant for the d2m pipeline so that's why it was put there but I can move this to ttir dir or maybe we need a new place for these kinds of passes? `RankNormalization.cpp` was another d2m transforms pass that was run on ttir graphs. Happy to consolidate these to wherever though

dlokeTT: Do you have any suggestions @nsmithtt . I kind of agree with @vwellsTT that the passes should be part of ttir dir since they are structured based on dialect instead of which pipelines use the passe

dlokeTT:
> > Pure permutes where dims of size 1 are shifted around and a permute is needed if the last dim is flipped with any other dims. e.g. [1, 128] -> [128, 1]
> 
> why is the fact there are 1 dims important? This seems true for 2x128 => 128x2 as well to me
> 
> > Singleton permutes where the last dim of size 1 is flipped with other dims that are involved in a real reshape. e.g [128, 4, 1] -> [1, 512]
> 
> same question, not following why the singleton part matters

We discussed offline and will add some comments to clarify this but a reshape from (2, 128) -> (128, 2) is different than a permute from (2, 128) -> (128, 2). In this example, the first would look something like 
```
[0, 1],
[2, 3], 
[4, 5],
```
and the latter would be 
```
[0, 128],
[1, 129],
[2, 130],
```
So, a reshape being lowered to permute is only valid if one of the dims involved is a 1.

lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
dlokeTT: fixed

lib/Dialect/D2M/Transforms/DecomposeComplexReshape.cpp
dlokeTT: removed it altogether. I think you're right that we never hit this

lib/Dialect/D2M/Transforms/DecomposeComplexReshape.cpp
dlokeTT: You're right, good catch. This is false and no permute needed.

lib/Dialect/TTIR/Transforms/DecomposeComplexReshape.cpp
dlokeTT: Yes in the literal sense of the permute op, a rank mismatch is a disqualifier. This helper is used in the context of the decomposition which already factors in a reshape to make sure the ranks match. I updated the comments to handle some of these questions and be more clear

vwellsTT: Okay, that makes sense, but I don't really understand in what sense this is a "pure permute" then, would consider renaming

lib/Dialect/D2M/Transforms/DecomposeComplexReshape.cpp
```diff
@@ -0,0 +1,354 @@
...
+
+    if (llvm::ArrayRef<int64_t>(swappedInserted) != outputShape) {
+      input = createReshape(rewriter, loc, input, outputShape, outputType);
+    }
+    return input;
+  }
+
+  // Decompose a reshape where trailing 1s become leading 1s with a genuine
+  // reshape.  Strategy: collapse to (X, 1), swap to (1, X), reshape to output.
+  void decomposeSingletonPermute(IRRewriter &rewriter, ttir::ReshapeOp op) {
```
vwellsTT: honestly, not sure on this name either--seems like the "pure" case also involves singletons.  Maybe this should be called "collapseAndSwap" or something, idk

lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
```diff
@@ -1502,8 +1502,24 @@ class D2MPermuteRewriter final
...
                                    /*tiled*/ true);
 
+    // Check if input comes from a view operation (StreamLayoutOp or
+    // ViewLayoutOp). If so, materialize it first to prevent view chaining that
+    // can cause affine map composition dimension mismatches when the views have
+    // different input/output ranks (e.g., reshape followed by permute).
+    Value inputValue = inputs[0];
+    if (auto *defOp = inputValue.getDefiningOp()) {
+      if (mlir::isa<d2m::StreamLayoutOp, d2m::ViewLayoutOp>(defOp)) {
+        // Materialize the view by creating an identity ToLayout.
```
vwellsTT: mildly surprised this actually works, I don't remember any explicit identity -> dma lowering in LowerToLayout?

lib/Dialect/TTIR/Transforms/DecomposeComplexReshape.cpp
dlokeTT: updated naming. Calling a `purePermute` a `singletonTranspose` now and calling `singletonPermutes` `flattenAndSwap` to be more indicative of what they are doing

lib/Dialect/D2M/Transforms/DecomposeComplexReshape.cpp
dlokeTT: renamed to `flattenAndSwap`

dlokeTT:
> Some comments, new comments LGTM and most of logic seems fine, still think decomp on TTIR ops should go in TTIR dir (and arguably should be in TTIRToTTIRDecomp instead of new pass, we have support for enabling specific patterns for D2M alone etc. already)

Moved this to ttir dir. I think we should have a consensus on how to approach all these passes. I feel like having them in the ttir dir is the best since we are adding more and more of these passes and that will bloat decomposition but not opposed to that either.

lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
dlokeTT: Went back and double checked this, it actually doesn't do anything. The logic in the `tensorManipulationRewriter` is actually what prevents view chaining by materializing thru a non identity` to_layout`. Removed this dead code.

vwellsTT:
>Moved this to ttir dir. I think we should have a consensus on how to approach all these passes. I feel like having them in the ttir dir is the best since we are adding more and more of these passes and that will bloat decomposition but not opposed to that either.

My personal opinion is that TTIRToTTIRDecomp should be decomposed (ha) into multiple files, but still run as a single configurable pass, and this + the permute case should be subsumed by that.  Would be interested to see if others have thoughts though yeah

vwellsTT (approved):
I'm okay with this now, thanks for addressing feedback.  Would still be interested in canvasing opinions on whether all TTIR decomps should go in current file, or it's too bloated and we should make it a directory instead, or what; for now my personal vote would be adding this type of decomp to TTIRToTTIRDecomp file, but won't block approval on it

dlokeTT:
> I'm okay with this now, thanks for addressing feedback. Would still be interested in canvasing opinions on whether all TTIR decomps should go in current file, or it's too bloated and we should make it a directory instead, or what; for now my personal vote would be adding this type of decomp to TTIRToTTIRDecomp file, but won't block approval on it

Yeah completely agree. Let me file an issue for this and gather some opinions and come up with a clean refactor of this in a couple weeks when I have some free cycles

lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
```diff
@@ -1870,10 +1870,80 @@ class D2MTensorManipulationOpRewriter
...
 
-    rewriter.replaceOp(op, unLayoutResult(rewriter, view->getResult(0),
+    // Get the logical rank for the generic op from output layout.
+    unsigned logicalRank = layout.getLogicalShape().size();
+
+    // Capture values explicitly for the lambda.
+    Value inputOperand = stream.getResult();
+    Value outputOperand = outputs[0];
+
+    auto generic = rewriter.create<d2m::GenericOp>(
```
nsmithtt: I don't think we want this right? This is saying that all TM's are now eagerly evaluated.

lib/Dialect/TTIR/Transforms/DecomposeComplexReshape.cpp
nsmithtt: Agreed, it looks like it's now moved to TTIR so this looks good.

lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
dlokeTT: This was put in just as a stop gap to enable some of these reshape + transpose test cases where the affine map composition was failing. I could disable the reshape + transpose tests and remove this and then enable those tests again when the composition works or remove this materialization in the follow on pr that fixes the composition. Just did it this way so we could unblock some reshape cases for now since the affine map composition bug looked like a bigger issue

nsmithtt: Is this to work around the back to back stream issue? If so I think we should write a canonicalizer for that first and merge that in and rebase this on top.

dlokeTT: What was the back to back stream issue? I was hitting an issue on some cases where the reshape would change the grid dims and use virtual grids so when that combined with a permute that didn't need virtual grids, I would hit an affine map composition error since the # of dims wouldn't match

nsmithtt: Oh ok, the back to back streams is a different one that Anurag was hitting then.  Ok yeah, I'd prefer we get to the bottom of that affine apply bug, ideally if we have a simple repro that we can get in as a prerequisite PR to this one.  You can send the repro to me and/or @bgrady-tt to help if needed.

dlokeTT: Hmm removing this materialization and testing reshape + transpose cases I now hit some cases with a stream layout issue where it survives thru the whole pipeline and get an unsupported op error in `TTMetalToFlatbuffer` Is this the back to back stream layout issue? 

I also hit an allocation error and some cases pass so seem to be getting a mixed bag but not the composition error I was seeing earlier, but that was also with some extra code to create identity toLayout ops.

dlokeTT: @nsmithtt I removed changes to `TTIRToD2M.cpp` to not force materialization. With the stream layout folding changes https://github.com/tenstorrent/tt-mlir/pull/7114 we now pass a good amount of reshape cases that involve transposes. There's still some that fail due to affine map composition errors but I'm working on a followup pr to fix those

Issue #6866, opened 2026-02-03
AssertionError: No parser found for opview <class 'ttmlir.dialects._ttir_ops_gen.RMSNormOp'>


mmilosevicTT:
### Describe the bug

While debugging prefil graph for Llama 3.2 1B with sequence length 2048 and batch size 2 we stumbled upon a new error in parser:
```
AssertionError: No parser found for opview <class 'ttmlir.dialects._ttir_ops_gen.RMSNormOp'>
```

I am attaching the IR graph that does not pass parsing phase.

[llama_graph.txt](https://github.com/user-attachments/files/25039522/llama_graph.txt)

### Steps to reproduce the issue

Use builder in tt-mlir with simple load from a file.

### Expected behavior

RMSNorm should parse in builder.

mmilosevicTT:
@jgrimTT I will take a look at this to unblock this graph.

PR #6870, opened 2026-02-03
Add Proper Support in Builder for TTIR Ops: RMSNormOp and LogicalAndOp

mmilosevicTT:
### Ticket
Closes https://github.com/tenstorrent/tt-mlir/issues/6866

### Problem description
While debugging Llama 3.2 graphs it was noticed that `RMSNormOp` and `LogicalAndOp` are not supported in builder.

### What's changed
Introduce required mappings and functionalities in builder to support these two ops.

### Checklist
- [x] Added Test for Each Op

### DIFF
--- a/test/python/golden/ttir_ops/normalization/test_normalization.py
+++ b/test/python/golden/ttir_ops/normalization/test_normalization.py
@@ -83,6 +83,7 @@ def test_batch_norm(
     [
         ((32, 128), [128]),
         ((2, 4, 64), [64]),
+        ((1, 136, 2048), [2048]),
     ],
 )
 @pytest.mark.parametrize("has_weight", [True, False])


test/python/golden/test_ttir_ops.py
```diff
@@ -513,6 +513,52 @@ def batch_norm_training(
...
+            (1, 136, 2048),  # input
+            (1, 136, 2048),  # weight
+            (2048,),  # bias
+        ],
+    ],
+)
+@pytest.mark.parametrize("dtypes", [[torch.float32] * 3])
+@pytest.mark.parametrize("normalized_shape", [2048])
+@pytest.mark.parametrize("epsilon", [1e-5])
+def test_rms_norm(
```
sdjordjevicTT: rms_norm test already exists here:
https://github.com/tenstorrent/tt-mlir/blob/7dbe1771d1abcd433bf3b42e8ec3da67d3ca8dc1/test/python/golden/ttir_ops/normalization/test_normalization.py#L91

You can add the needed parameters to an existing test.

We are moving away from a single test_ttir_ops testing file to an organized folder structure under:
https://github.com/tenstorrent/tt-mlir/tree/main/test/python/golden/ttir_ops

tools/golden/mapping.py
```diff
@@ -972,6 +972,32 @@ def rms_norm_golden(
     return rms_norm.to(input.dtype)
 
 
+def ttir_rms_norm_golden(
```
sdjordjevicTT, line 975: Can you rename the existing rms_norm_golden instead of a new one?

mmilosevicTT: No, builder is not built to support functions that were present there before. The reason why I did not remove the existing implementation is that there is a TTNN implementation of that op as well, that used the same method.

test/python/golden/test_ttir_ops.py
mmilosevicTT: Will do, thanks for noting this.

mmilosevicTT: Moved the parametrised version to the file you suggested.

Issue #6736, opened 2026-01-23
[D2M][ttnn.jit] Set the default Output Tensor Layout to a maximally Block Sharded output layout


brapananTT:
We want to be able to set a default output layout when one is not specified within the ttnn.jit decorator (output layouts in ttnn.jit: #5280)
Support for a `DRAM interleaved` default layout is blocked by #6735.

Issue #6735, opened 2026-01-23
[D2M] Correctly handle L1 Sharded -> DRAM Interleaved to_layout in LowerToLayout pass


brapananTT:
The [LowerToLayout pass ](https://github.com/tenstorrent/tt-mlir/blob/main/lib/Dialect/D2M/Transforms/LowerToLayout.cpp#L699)currently handles a to_layout op from an `l1 sharded layout -> dram interleaved layout` in the following manner:

1. create an L1 interleaved view_layout + dma generic as an intermediate transform 
2. create a DRAM interleaved view_layout + dma generic as the output transform

An l1 interleaved layout is not supported, and this causes an error out.

The transformation logic needs modification in some form to support this `l1 sharded -> dram interleaved` case.

PR #6875, opened 2026-02-03
[ttnn.jit] Set the default Output Tensor Layout to Maximally L1 Block Sharded Output Layout

brapananTT:
### Ticket
[D2M][ttnn.jit] Set the default Output Tensor Layout to a maximally Block Sharded output layout #6736 

### Problem description
For integration with intermediate layout support coming to ttnn.jit, it is necessary to set a default output layout for the JIT'ed function if one is not provided within the ttnn.jit decorator. Setting this default layout to be DRAM Interleaved is currently unsupported (see issue #6735), so instead the default will be a maximally l1 block sharded layout.

### What's changed
- created helper `get_maximal_block_sharding_grid` and added logic in `tracing_compiler.py ` to create a l1 block sharded memory config based on the output tensor shape + insert into IR.
- Updated tests to check for the new default memory config when using `assert memory_configs_equal()`

### Checklist

### DIFF
--- a/test/ttnn-jit/lit/test_tracing_ir.py
+++ b/test/ttnn-jit/lit/test_tracing_ir.py
@@ -502,11 +502,12 @@ if __name__ == "__main__":
     # CHECK: ---- IR Dump after TracingCompiler (Tracing-based) ----
     # CHECK: func.func @permute_2d_func
     # CHECK-SAME: (%arg0: tensor<64x128xbf16, #ttnn_layout>)
-    # CHECK-SAME: -> tensor<128x64xbf16, #ttnn_layout{{[0-9]*}}>
+    # CHECK-SAME: -> [[OUT_TYPE:tensor<128x64xbf16, #ttnn_layout[0-9]*>]]
     # CHECK: %[[VAL:[0-9]+]] = "ttir.permute"(%arg0)
     # CHECK-SAME: permutation = array<i64: 1, 0>
-    # CHECK-SAME: (tensor<64x128xbf16, #ttnn_layout>) -> tensor<128x64xbf16, #ttnn_layout{{[0-9]*}}>
-    # CHECK: return %[[VAL]] : tensor<128x64xbf16, #ttnn_layout{{[0-9]*}}>
+    # CHECK-SAME: (tensor<64x128xbf16, #ttnn_layout>) -> tensor<{{.*}}>
+    # CHECK: %[[CONVERTED:[0-9]+]] = ttir.to_layout %[[VAL]]{{.*}} -> [[OUT_TYPE]]
+    # CHECK: return %[[CONVERTED]] : [[OUT_TYPE]]
     test_ir_generation(permute_2d_func, input_c)
 
     # 4D permute (attention-style BHSD -> BSHD)
@@ -516,11 +517,12 @@ if __name__ == "__main__":
     # CHECK: ---- IR Dump after TracingCompiler (Tracing-based) ----
     # CHECK: func.func @permute_4d_func
     # CHECK-SAME: (%arg0: tensor<1x8x32x64xbf16, #ttnn_layout>)
-    # CHECK-SAME: -> tensor<1x32x8x64xbf16, #ttnn_layout{{[0-9]*}}>
+    # CHECK-SAME: -> [[OUT_TYPE:tensor<1x32x8x64xbf16, #ttnn_layout[0-9]*>]]
     # CHECK: %[[VAL:[0-9]+]] = "ttir.permute"(%arg0)
     # CHECK-SAME: permutation = array<i64: 0, 2, 1, 3>
-    # CHECK-SAME: (tensor<1x8x32x64xbf16, #ttnn_layout>) -> tensor<1x32x8x64xbf16, #ttnn_layout{{[0-9]*}}>
-    # CHECK: return %[[VAL]] : tensor<1x32x8x64xbf16, #ttnn_layout{{[0-9]*}}>
+    # CHECK-SAME: (tensor<1x8x32x64xbf16, #ttnn_layout>) -> tensor<{{.*}}>
+    # CHECK: %[[CONVERTED:[0-9]+]] = ttir.to_layout %[[VAL]]{{.*}} -> [[OUT_TYPE]]
+    # CHECK: return %[[CONVERTED]] : [[OUT_TYPE]]
     test_ir_generation(permute_4d_func, input_4d)
 
     # ============================================================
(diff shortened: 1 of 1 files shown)


lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
```diff
@@ -1650,6 +1639,16 @@ class D2MToLayoutOpRewriter : public D2MNamedRewriterCommon,
   }
 
 public:
+  D2MToLayoutOpRewriter(const TypeConverter &typeConverter,
+                        mlir::MLIRContext *ctx,
+                        ttcore::MemorySpace defaultInputMemSpace,
+                        ttcore::MemorySpace defaultOutputMemSpace,
+                        bool ttnnMode, bool collapseTensors)
+      : D2MNamedRewriterCommon(defaultInputMemSpace, defaultOutputMemSpace,
```
vmilosevic: :warning: **clang\-diagnostic\-error** :warning:
no matching constructor for initialization of `` D2MNamedRewriterCommon ``

vmilosevic (commented):

lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
```diff
@@ -1590,6 +1590,24 @@ class D2MToLayoutOpRewriter : public D2MNamedRewriterCommon,
 
   using D2MNamedRewriterCommon::getMetalTensorFromTTNNTensor;
 
+  LogicalResult
```
arminaleTT, line 1686: This seems like a rebasing artifact

tools/ttnn-jit/_src/utils.py
```diff
@@ -101,3 +101,24 @@ def get_dispatch_core_type():
...
+        collapsed_dim = 1
+        for i in range(len(shape) - 1):
+            collapsed_dim *= shape[i]
+    else:
+        collapsed_dim = shape[0]
+    tile_shape = [collapsed_dim // 32, shape[-1] // 32]
+
+    grid = []
+    for dim in tile_shape:
+        for grid_dim in reversed(range(8)):
```
arminaleTT: Make the grid dims a parameter for this function. I don't know if we have the device available, but ideally we should read it from the device. If that's not possible right now, a hardcoded `[8, 8]` is still better than this

arminaleTT (commented):
Looks good! 
Two changes regarding the layout tests in nightly:
1. Could you disable the ND sharding tests and put a case on me about them? They are failing right now and will be even more messed up with this and grid selection. I need to evaluate their future
2. Could you pin the output layout for every layout test to be the same as the input?

lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
brapananTT: I did a bit of code cleanup here during the rebase and moved `matchAndRewrite` to the top of the class.

tools/ttnn-jit/_src/utils.py
brapananTT: I've updated this to get the core_grid from the first `tensor_arg`  in a helper function `get_core_grid_from_tensor_args(tensor_args)`:
```
tensor_arg = next(iter(tensor_args.values()))
device = tensor_arg.device()
return (device.core_grid.x, device.core_grid.y)
```
This should avoid hardcoding the max coregrid to [8, 8].

test/ttnn-jit/utils.py
```diff
@@ -76,19 +76,43 @@ def _get_input_transform(
...
+def get_block_sharding_grid(shape, core_grid):
+    """Infer a TTNN grid/end coord for block sharding the given logical tensor shape and device core grid"""
+    # collapse dims [0, -1)
+    if len(shape) > 2:
+        collapsed_dim = 1
+        for i in range(len(shape) - 1):
+            collapsed_dim *= shape[i]
+    else:
+        collapsed_dim = shape[0]
+    tile_shape = [collapsed_dim // 32, shape[-1] // 32]
```
sgholamiTT: Nit: Can you please import and use `TILE_WIDTH` and `TILE_HEIGHT` defined in `tensor_translator.py` instead of hardcoding 32 ?

test/ttnn-jit/utils.py
```diff
@@ -76,19 +76,43 @@ def _get_input_transform(
...
+
```
sgholamiTT: Is it possible to use `_calculate_tile_shape` defined in `tensor_translator.py` to get the `tile_shape` from `shape` ? I feel like the logics are the same.

test/ttnn-jit/utils.py
```diff
@@ -76,19 +76,43 @@ def _get_input_transform(
...
-        for grid_dim in reversed(range(8)):
+    for dim, max_grid in zip(tile_shape, core_grid):
+        for grid_dim in reversed(range(max_grid)):
             if dim % (grid_dim + 1) == 0:
                 grid.append(grid_dim)
                 break
     return list(reversed(grid))
 
 
+def get_expected_memory_config(tensor_shape, core_grid):
```
sgholamiTT: Nit: Subjective, but I think the name of this function is misleading. Maybe `get_expected_block_sharded_memory_config` is better ?

tools/ttnn-jit/_src/utils.py
```diff
@@ -101,3 +101,34 @@ def get_dispatch_core_type():
...
+        collapsed_dim = shape[0]
+    tile_shape = [collapsed_dim // 32, shape[-1] // 32]
+
+    grid = []
+    for dim, max_grid in zip(tile_shape, core_grid):
+        for grid_dim in reversed(range(max_grid)):
+            if dim % (grid_dim + 1) == 0:
+                grid.append(grid_dim)
+                break
+    return list(reversed(grid))
```
sgholamiTT, line 126: Isn't this function the same with `get_block_sharding_grid` defined in the other `utils.py` file ? 
If so, let's remove the redundant function and just import and use the one defined here.

tools/ttnn-jit/_src/tracing_compiler.py
```diff
@@ -112,6 +116,26 @@ def compile(self):
 
         # Insert output layout conversion if memory_config provided
         try:
+
+            if self.memory_config is None:
+                # If no memory_config is provided, set output layout to block-sharded
```
sgholamiTT, line 124: My understanding is that we're making decisions based on an argument that is received from the user and we're changing the behavior of the compiler if that arg is unset. If that's true:

Should we update `docs/src/ttnn-jit.md` with this decision ? I think the user should be clearly identified that:
1. If they specify an output layout, we respect it.
2. If not, we'll use a maximally block sharded layout.

Also, does this line need an update with this decision: 
https://github.com/tenstorrent/tt-mlir/blob/addf7c827d054da6b55af4e69fc485292391a33d/tools/ttnn-jit/api.py#L26
The `defaults to None` part should essentially say sth like: If unspecified, a maximally block sharded layout will be used.

brapananTT: Updated ttnn-jit.md and api.py to reflect this!

arminaleTT (approved):

Issue #6879, opened 2026-02-03
[D2M] Explicit Datamovement form of d2m.generic doesn't support aliased mode operands


bgrady-tt:
The basic reason for requiring stream_layout for explicit datamovement form is that the checks for aliasing in DMA lowering passes rely on the presence of StreamLayout to tell if aliasing is required. 

For regular D2M generic ops, checking if a StreamLayout is present is enough to tell if an operand is remote or aliased local CB, as the earlier passes will insert streams automatically if an operand cannot be in alias mode. 

The above is **_not true_** for the explicit form, which can in theory opt out of streams on any input/output, and therefore remote vs local operands must be inferred in a more rigorous way.

PR #6880, opened 2026-02-03
[D2M] Explicit Datamovement Generic Form Support

bgrady-tt:
Introduces a series of changes to enable explicit datamovement form of the d2m.generic op. 

See https://github.com/tenstorrent/tt-mlir/issues/6879 for more details on limitations wrt explicit form operands.

### DIFF
--- a/test/ttmlir/Dialect/D2M/generic/explicit_datamovement_matmul_mcast.mlir
+++ b/test/ttmlir/Dialect/D2M/generic/explicit_datamovement_matmul_mcast.mlir
@@ -1,0 +1,125 @@
+// RUN: ttmlir-opt --split-input-file --canonicalize %s | FileCheck %s
+
+// Test for explicit datamovement form of d2m.generic with matmul operation
+// using low-level multicast remote_load (mcore/mshape).
+//
+// Matmul: C[M,N] = sum_K A[M,K] * B[K,N]
+// Output grid: 2x4, K reduction blocks: 3
+// A: grid 2x3, shard 2x2 tiles (logical 4x6 tiles)
+// B: grid 3x4, shard 2x2 tiles (logical 6x8 tiles)
+// C: grid 2x4, shard 2x2 tiles (logical 4x8 tiles)
+//
+// Multicast pattern:
+//   A at [i,k] -> multicast along columns: mcore[0,0] mshape[1,4]
+//   B at [k,j] -> multicast along rows: mcore[0,0] mshape[2,1]
+//
+// Each k iteration loads A/B, computes a partial matmul, and stores C.
+// The inner linalg.generic reduction handles the tile-level contraction.
+
+#l1_ = #ttcore.memory_space<l1>
+#mapL = affine_map<(d0, d1, d2) -> (d0, d2)>
+#mapR = affine_map<(d0, d1, d2) -> (d2, d1)>
+#mapO = affine_map<(d0, d1, d2) -> (d0, d1)>
+
+#layout_a = #ttcore.metal_layout<logical_shape = 64x96, dim_alignments = 32x32, collapsed_intervals = dense<[[0, 1], [1, 2]]> : tensor<2x2xi64>, undef, l1, sharded, index_map = map(0)>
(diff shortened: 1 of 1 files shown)


nsmithtt:
Hey @bgrady-tt I think this might be a bad rebase?  Seems like a bunch of unrelated changes were pulled in the diff

lib/Dialect/D2M/Transforms/LowerLoadStoreOpsToDMA.cpp
```diff
@@ -372,61 +372,6 @@ class D2MLowerRemoteLoadRewritePattern : public OpRewritePattern<RemoteLoadOp> {
     return loopNest.results.front();
   }
 
-  // Calculate which dimensions are multicast dimensions based on the parent
```
bgrady-tt, line 375: The prior implementation of multicast lowering here relied on indexing maps and iterator type attributes. This wasn't technically necessary and as redoing work done earlier when lowering high-level multicast to low-level form (with explicit shape and start point).

bgrady-tt: This is all part of cleanup that gets end-of-middleend passes to stop using iterator types, blocking factors, and indexing maps so they only work with (something like) the explicit datamovement form.

lib/Dialect/D2M/IR/D2MGenericRegionOps.cpp
```diff
@@ -308,29 +308,40 @@ ::mlir::LogicalResult RemoteLoadOp::verify() {
...
-
-      for (int64_t gridDim : mcastDimIndices) {
-        if (gridDim < 0 ||
-            gridDim >= static_cast<int64_t>(indexingMap.getNumResults())) {
-          return emitOpError("mcast dimension index ")
-                 << gridDim << " is out of bounds for grid rank "
-                 << indexingMap.getNumResults();
-        }
+    // Forbid high-level mcast form in explicit datamovement form
+    if (genericOp.isExplicitDatamovementForm() && !getMcastDims().empty()) {
```
bgrady-tt, line 312: high-level multicast remote loads can only be lowered with the aid of indexing maps and iterator type attributes. So they are illegal in the low-level multicast form.

lib/Dialect/D2M/Transforms/Allocate.cpp
```diff
@@ -485,7 +485,7 @@ class D2MAllocate final : public impl::D2MAllocateBase<D2MAllocate> {
               op)) {
         // Skip memref.alloc operations that have a genericOp as parent
         if (llvm::isa<memref::AllocOp>(op) &&
-            llvm::isa<d2m::GenericOp>(op->getParentOp())) {
+            op->getParentOfType<d2m::GenericOp>()) {
```
bgrady-tt, line 488: In explicit datamovement form, the _immediate_ parent of a memref.alloc in a GenericOp region might be a `scf.for` loop. So use getParentOfType() to check all the way up the hierarchy.

bgrady-tt:
@nsmithtt There was indeed a weird rebase thing happening. It is fixed now.

test/ttmlir/Dialect/D2M/generic/explicit_datamovement_form_sequential_loops.mlir
```diff
@@ -0,0 +1,141 @@
...
+    scf.for %i = %c0 to %c1 step %c1 {
+      scf.for %j = %c0 to %c3 step %c1 {
+        %buffer_in = tensor.empty() : tensor<2x2x!ttcore.tile<32x32, f32>>
+        %buffer_out = tensor.empty() : tensor<2x2x!ttcore.tile<32x32, f32>>
+
+        // Load input shard at grid position [i, j]
+        %loaded = d2m.remote_load %buffer_in %arg0_stream[%i, %j]
+          : tensor<2x2x!ttcore.tile<32x32, f32>>,
+            tensor<2x3x2x2x!ttcore.tile<32x32, f32>, #layout>
+          -> tensor<2x2x!ttcore.tile<32x32, f32>>
```
nsmithtt, line 81: Can we also have a test that uses an mcast load? Seems like the verifier has a special case for that.

bgrady-tt: sure I can add a test for multicast load.

bgrady-tt: added a test.

PR #6881, opened 2026-02-04
Fix `ttnn.pad` logical shape and padded shape mismatch

pradhanaditya:
`ttnn::pad` may not update logical shape to match the padded shape after explicit padding. This causes `write_tensor` to fail when writing a padded host tensor to device.

### Problem description
The `ttnn.pad` operation does not work correctly for 4D+ tensors, especially when the result tensor of this operation is used by `ttnn.write_tensor` operation. The TTNN runtime reports a logical shape mismatch error when we try to write such a padded tensor to the device.

### What's changed
The `experimental::view` is used to change the metadata of the tensor to have the same logical and padded shape. This is a workaround and the proper fix should be in tt-metal's pad implementation.

A unit test has been added to exercise this path. I did a `ttrt run --save-artifacts` on the unit test as well and found that there are no runtime issues and the device tensor has same values as the padded host tensor after this change.

### Checklist

### DIFF
--- a/runtime/lib/ttnn/operations/data_movement/write_tensor.cpp
+++ b/runtime/lib/ttnn/operations/data_movement/write_tensor.cpp
@@ -9,6 +9,8 @@
 #include "tt/runtime/detail/ttnn/utils.h"
 #include "ttnn/tensor/tensor_impl.hpp"
 
+#include "ttnn/operations/experimental/reshape/view.hpp"
+
 namespace tt::runtime::ttnn::operations::data_movement {
 void run(const ::tt::target::ttnn::WriteTensorOp *op, ProgramContext &context) {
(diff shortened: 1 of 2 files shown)


pradhanaditya:
@xanderchin @nsmithtt Could you please take a look into this and let me know if there are better alternatives or if this can be fixed in the tt-metal runtime library?

nsmithtt:
@pradhanaditya, please take a look at the failing jobs, some tests are failing.

nsmithtt:
@tenstorrent/forge-developers-mlir-runtime, can we get a review on this?

nsmithtt:
Instead of doing this workaround in tt-mlir runtime, we will push on a proper fix from tt-metal:
- https://github.com/tenstorrent/tt-metal/issues/38144

Issue #6873, opened 2026-02-03
Regression on multiple models due to PoolingOp removal


mvasiljevicTT:
Since [this commit](352b608d4a8412a00e1f89334ea77e5721c2e7b1) we notice regressions:

| Model            | Status    | Value Change | % Change | Result     |
|------------------|-----------|--------------|----------|------------|
| tt-xla-resnet    | + SUCCESS | -364.97      | -44.3%   | REGRESSION |
| tt-xla-segformer | + SUCCESS | -1.53        | -6.3%    | REGRESSION |
| tt-xla-ufld_v2   | + SUCCESS | -49.48       | -35.7%   | REGRESSION |
| tt-xla-vovnet    | + SUCCESS | -161.14      | -30.1%   | REGRESSION |

In resnet, instead of 
```
max_pool2d<padding = array<i32: 1, 1>>
```
there is
```
pad<padding = array<i32: 0, 0, 0, 0, 1, 1, 1, 1>>
max_pool<padding = array<i32: 0,0>>
```
which is problematic because padOp currently doesn't commute with reshape and permute, and we get
```
        %60 = "ttnn.conv2d"(%59, %0#0, %0#1, %55) <{batch_size = 8 : i32, compute_config = #ttnn.device_compute_kernel_config<math_fidelity = hifi4, fp32_dest_acc_en = true>, conv2d_config = #ttnn.conv2d_config<weights_dtype = bf16, activation = <op_type = relu>, deallocate_activation = true, act_block_h_override = 64, enable_kernel_stride_folding = false, config_tensors_in_dram = true>, conv2d_slice_config = #ttnn.conv2d_slice_config<l1_full, 0>, dilation = array<i32: 1, 1>, dtype = #ttcore.supportedDataTypes<bf16>, groups = 1 : i32, in_channels = 3 : i32, input_height = 224 : i32, input_width = 224 : i32, kernel_size = array<i32: 7, 7>, out_channels = 64 : i32, padding = array<i32: 3, 3, 3, 3>, stride = array<i32: 2, 2>}> : (tensor<1x1x401408x3xbf16, #ttnn_layout143>, tensor<1x1x147x64xbf16, #ttnn_layout2>, tensor<1x1x1x64xbf16, #ttnn_layout3>, !ttnn.device) -> tensor<1x1x100352x64xbf16, #ttnn_layout144> loc(#loc3)
[...]

PR #6883, opened 2026-02-04
Optimize PadOp and Max/AvgPool2dOp fusing

acicovicTT:
### Issue

https://github.com/tenstorrent/tt-mlir/issues/6873

### What's changed

Move the `PadOp` and `Max/AvgPool2dOp` fusing to be done during conversion to TTIR, since now that `PoolingOp` is removed that is the latest moment at which the layout matches between the equivalent StableHLO ops (this is the main reason why the change wasn't made in TTIR to enable commuting past reshape/permute ops for `PadOp`). This will fix the perf regression caused by removing the `PoolingOp` from TTIR.

### Checklist

- [x] New/existing tests provide coverage for changes.
- [x] Perf. benchmarks confirm that the regression is fixed (ran on n300):
  - resnet50: ~450 -> 800+ [sample/s]
  - vovnet: ~360 -> 530+ [sample/s]
  - ufld_v2: ~90 -> 110+ [sample/s]

### DIFF
--- a/test/ttmlir/Conversion/StableHLOToTTIR/avgpool2d_op.mlir
+++ b/test/ttmlir/Conversion/StableHLOToTTIR/avgpool2d_op.mlir
@@ -86,8 +86,6 @@ func.func @test_avgpool2d_with_division(%arg0: tensor<1x14x14x112xbf16>) -> tens
 
 // -----
 
-// Test that stablehlo.pad is fused into avg_pool2d when padding only spatial
-// dimensions (NHWC layout).
 func.func @test_avgpool2d_with_pad_fusion(%arg0: tensor<1x112x112x64xbf16>) -> tensor<1x58x58x64xbf16> {
   %cst = stablehlo.constant dense<0.000000e+00> : tensor<bf16>
   %pad_value = stablehlo.constant dense<0.000000e+00> : tensor<bf16>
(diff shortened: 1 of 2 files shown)


lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
```diff
@@ -1285,78 +1285,6 @@ class RepVGGConvSumFusionPattern : public mlir::OpRewritePattern<AddOp> {
   }
 };
 
-// Fuses a PadOp into a Pool2dOp by combining their padding attributes.
-// This pattern matches when a Pool2dOp's input comes from a PadOp that only
-// pads spatial dimensions (H, W), and combines both paddings into a single
-// Pool2dOp with updated padding.
-template <typename Pool2dOp>
-class PadPool2dFusionPattern : public mlir::OpRewritePattern<Pool2dOp> {
```
mtopalovicTT, line 1293: It wouldn't hurt to leave this pattern.

acicovicTT: Thought about this, but decided against it do avoid having duplicated logic in two places + in majority of cases pooling ops will be surrounded by reshape or permute and this TTIR fusing will fail

Issue #6840, opened 2026-01-30
[Optimizer] ttnn.arange tile->row layout conversion not working.


ddilbazTT:
When I enable optimizer for gpt-oss-20b though tt-forge, I encounter an error with `ttnn.arange`:
```
error: 'ttnn.arange' op output tensor layout row_major must match output layout attribute tile
```

This is due to the following file check:
`third_party/tt-mlir/src/tt-mlir/lib/OpModel/TTNN/TTNNOpModel.cpp` 
```
//===----------------------------------------------------------------------===//
// ArangeOp
//===----------------------------------------------------------------------===//
// sgholamiTT: There are two reasons why receiving the start, end, and step as
// attributes is better than as integers:
//   1. That is the only valid way to aquire a pointer to MLIRContext.
//   2. Using getInt() member function of ::mlir::IntegerAttr is safer and more
//      mlir idiomatic than static_cast<int64_t>(start).
llvm::Expected<OpConstraints>
OpModel<mlir::tt::ttnn::ArangeOp>::getOpConstraints(
    mlir::tt::ttcore::GridAttr deviceGrid, ::mlir::IntegerAttr start,
    ::mlir::IntegerAttr end, ::mlir::IntegerAttr step,
    std::optional<mlir::tt::ttcore::DataType> dtype,
    std::optional<mlir::tt::ttnn::MemoryConfigAttr> memConfig,
    mlir::tt::ttnn::TTNNLayoutAttr outputLayout) {
#ifdef TTMLIR_ENABLE_OPMODEL
  ::tt::tt_metal::distributed::MeshDevice *device =
      SingletonDeviceContext::getInstance().getDevice();
  // ~~~~~~~~~~~~~~~~~~~~~ Note ~~~~~~~~~~~~~~~~~~~~~
  // The following default values are taken from Arrange's invoke function in
  // tt-metal/ttnn/cpp/ttnn/operations/creation.hpp
  const ::tt::tt_metal::DataType defaultDtypeInMetal =
      ::tt::tt_metal::DataType::BFLOAT16;
  const ::ttnn::MemoryConfig defaultMemoryConfigInMetal =
      ::ttnn::DRAM_MEMORY_CONFIG;
  const ::ttnn::Layout defaultLayoutInMetal = ::ttnn::ROW_MAJOR_LAYOUT;
  // ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

[...]

PR #6884, opened 2026-02-04
[Optimizer] Fix creation op layout attribute mismatch

bmalesevicTT:
### Ticket
#6840 

### Problem description
1. In `OptimizerPasses/OperationValidationAndFallback.cpp` layout attribute was never changed, even when the layout itself is - causing a later mismatch error.
2. ArangeOp in OpModel passed default `::ttnn::ROW_MAJOR_LAYOUT`, instead of reading from expected output layout.
### What's changed
1. `OptimizerPasses/OperationValidationAndFallback.cpp` layout attribute is now updated using `TTNNLayoutOpInterface`, matching the pattern in other `OptimizerPasses`
2. `ArangeOp` in `OpModel` only passes `::ttnn::ROW_MAJOR_LAYOUT` by default when there is no expected output layout, otherwise reads the expected value from it and passes that to backend.

The fix for 1 cannot be reproduced after 2 is fixed as well, hence test is not added. Fix 1 is still included to match existing patterns and was verified before fix 2 was discovered.

### DIFF
--- a/lib/OpModel/TTNN/TTNNOpModel.cpp
+++ b/lib/OpModel/TTNN/TTNNOpModel.cpp
@@ -6950,9 +6950,12 @@ OpModel<mlir::tt::ttnn::ArangeOp>::getOpConstraints(
     dataType = conversion::getDataType(dtype.value());
   }
   ::ttnn::MemoryConfig memoryConfig = defaultMemoryConfigInMetal;
+  ::ttnn::Layout layout = defaultLayoutInMetal;
   // Prefer the output layout if possible:
   if (outputLayout) {
     memoryConfig = conversion::getMemoryConfig(outputLayout);
+    layout =
+        outputLayout.isTiled() ? ::ttnn::TILE_LAYOUT : ::ttnn::ROW_MAJOR_LAYOUT;
   } else if (memConfig.has_value()) {
     memoryConfig = conversion::getMemoryConfig(memConfig.value());
   }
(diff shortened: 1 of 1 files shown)


lib/Dialect/TTNN/Transforms/OptimizerPasses/OperationValidationAndFallback.cpp
```diff
@@ -721,6 +721,14 @@ void applyFallbackTransformations(
                               result.actualOutputLayout);
     operation->getResult(0).setType(newResultType);
 
+    // Update the layout attribute for ops that have one (e.g., creation ops).
+    // The layout attribute must match the result type's layout.
+    if (TTNNLayoutOpInterface opWithLayoutIF =
+            mlir::dyn_cast<TTNNLayoutOpInterface>(operation)) {
+      opWithLayoutIF.setLayoutAttr(LayoutAttr::get(
+          operation->getContext(), result.actualOutputLayout.getLayout()));
+    }
```
odjuricicTT, line 726: Do we already have this logic in row major propagation or is it missing there as well?

bmalesevicTT: I checked, there is no such logic, but it seems unnecessary for RMPropagationPass since it never considers Creation ops. 
RMPropagationPass starts by checking for function inputs which were originally row-major and unnecessarily converted to Tile, and then propagates through all their users. Creation ops have no tensor inputs, hence are never traversed in this pass.

sgholamiTT (approved):

PR #6888, opened 2026-02-04
Add kv_cache argtype

### LINKED ISSUE #6302: Add KVCache ArgumentType for distinguishing KV cache tensor arguments
## Problem

Currently, function arguments are marked with one of these types via `ttcore.argument_type`:
- `input` - Model inputs
- `parameter` - Learnable parameters  
- `constant` - Constants
- `default` - Default type

KV cache tensors are currently marked as `input`, but they have different layout requirements and handling than regular inputs. For example:
- Regular inputs may be forced to row-major layout in certain cases
- KV cache tensors should **always** remain tiled

## Proposed Solution

Add a new `ArgumentType` enum value: `kv_cache` (or `kv_cache`).

### Changes needed:

Figure out if we have this information from the framework and mark KV cache function arguments with `ttcore.argument_type = #ttcore.argument_type<kv_cache>`.

vkovacevicTT:
### Ticket
Closes #6302

### Problem description
KV cache tensors are marked with `input` ttcore.argument_type.
User inputs come from host and are in RowMajor layout, while KV cache tensors are on device in Tile layout.
If we treat them as `input` arguments, they will be forced to RowMajor as user inputs and then tilized as TTIR op operands with to_layout op in TTNNLayout pass.

Currently, we avoid forcing RowMajor layout on UpdateCache and PagedUpdateCache op operands: 

https://github.com/tenstorrent/tt-mlir/blob/ac0cfa06d05980373265b008b9ffdaca8c7ba0c5/lib/Dialect/TTNN/Transforms/TTNNLayout.cpp#L608-L623

but not for FillCache and PagedFillCache ops.

E.g. in Falcon3 1B prefill we have patterns like:

```
%144 = "ttnn.slice_static"(%143) <{begins = [0 : i32, 0 : i32, 0 : i32, 0 : i32], ends = [1 : i32, 4 : i32, 17 : i32, 256 : i32], step = [1 : i32, 1 : i32, 1 : i32, 1 : i32]}> : (tensor<32x4x17x256xbf16, #ttnn_layout43>) -> tensor<1x4x17x256xbf16, #ttnn_layout52> loc(#loc462)
%145 = "ttnn.to_layout"(%arg6) <{layout = #ttnn.layout<tile>}> : (tensor<32x4x128x256xbf16, #ttnn_layout32>) -> tensor<32x4x128x256xbf16, #ttnn_layout33> loc(#loc463) <- Unnecessary tilize op
"ttnn.deallocate"(%arg6) <{force = false}> : (tensor<32x4x128x256xbf16, #ttnn_layout32>) -> () loc(#loc463)
"ttnn.fill_cache"(%145, %144) <{batch_offset = 0 : i32}> : (tensor<32x4x128x256xbf16, #ttnn_layout33>, tensor<1x4x17x256xbf16, #ttnn_layout52>) -> () loc(#loc462)
"ttnn.deallocate"(%144) <{force = false}> : (tensor<1x4x17x256xbf16, #ttnn_layout52>) -> () loc(#loc462)
%146 = "ttnn.slice_static"(%143) <{begins = [1 : i32, 0 : i32, 0 : i32, 0 : i32], ends = [2 : i32, 4 : i32, 17 : i32, 256 : i32], step = [1 : i32, 1 : i32, 1 : i32, 1 : i32]}> : (tensor<32x4x17x256xbf16, #ttnn_layout43>) -> tensor<1x4x17x256xbf16, #ttnn_layout52> loc(#loc462)
"ttnn.fill_cache"(%145, %146) <{batch_offset = 1 : i32}> : (tensor<32x4x128x256xbf16, #ttnn_layout33>, tensor<1x4x17x256xbf16, #ttnn_layout52>) -> () loc(#loc462)
"ttnn.deallocate"(%146) <{force = false}> : (tensor<1x4x17x256xbf16, #ttnn_layout52>) -> () loc(#loc462)
%147 = "ttnn.slice_static"(%143) <{begins = [2 : i32, 0 : i32, 0 : i32, 0 : i32], ends = [3 : i32, 4 : i32, 17 : i32, 256 : i32], step = [1 : i32, 1 : i32, 1 : i32, 1 : i32]}> : (tensor<32x4x17x256xbf16, #ttnn_layout43>) -> tensor<1x4x17x256xbf16, #ttnn_layout52> loc(#loc462)
"ttnn.fill_cache"(%145, %147) <{batch_offset = 2 : i32}> : (tensor<32x4x128x256xbf16, #ttnn_layout33>, tensor<1x4x17x256xbf16, #ttnn_layout52>) -> () loc(#loc462)
```

### What's changed
*Updated after discussion below:*
- ~~Added a new TTCore_ArgumentType - kv_cache.~~
- Added a pass `TTIRInferKVCacheArgumentTypes` that marks KV Cache tensors with ~~new 'kv_cache' argtype~~ `ttcore.kv_cache` attribute.
- Added some tests.

### Checklist
- [x] End-to-end tt-xla test on Llama-3.2-1B (Device perf is expectedly failing)
    - https://github.com/tenstorrent/tt-xla/actions/runs/21873037327 
- [x] End-to-end tt-xla test on all single chip llms
    - https://github.com/tenstorrent/tt-xla/actions/runs/21942268612

### DIFF
--- a/lib/Dialect/TTIR/Transforms/InferKVCacheArgumentTypes.cpp
+++ b/lib/Dialect/TTIR/Transforms/InferKVCacheArgumentTypes.cpp
@@ -3,7 +3,6 @@
 // SPDX-License-Identifier: Apache-2.0
 
 #include "ttmlir/Dialect/TTCore/IR/Utils.h"
-#include "ttmlir/Dialect/TTIR/IR/TTIROps.h"
 #include "ttmlir/Dialect/TTIR/IR/TTIROpsInterfaces.h"
 #include "ttmlir/Dialect/TTIR/Transforms/Passes.h"
 
@@ -13,35 +12,6 @@
 namespace mlir::tt::ttir {
 #define GEN_PASS_DEF_TTIRINFERKVCACHEARGUMENTTYPES
 #include "ttmlir/Dialect/TTIR/Transforms/Passes.h.inc"
-
-namespace {
-
-// Traces the cache operand back through cache operations to find the original
-// function argument.
-BlockArgument traceCacheToBlockArg(Value cacheOperand) {
-  Value current = cacheOperand;
-
-  while (current) {
-    if (auto blockArg = llvm::dyn_cast<BlockArgument>(current)) {
-      return blockArg;
-    }
-
-    Operation *defOp = current.getDefiningOp();
-    if (!defOp) {
-      return nullptr;
-    }
-
-    if (auto cacheOp = mlir::dyn_cast<CacheOpInterface>(defOp)) {
-      current = cacheOp.getCache();
-    } else {
-      return nullptr;
-    }
-  }
-
-  return nullptr;
-}
-
-} // namespace
 
 class TTIRInferKVCacheArgumentTypes
     : public impl::TTIRInferKVCacheArgumentTypesBase<
@@ -58,9 +28,10 @@ public:
       llvm::DenseSet<BlockArgument> cacheArgs;
 
       funcOp.walk([&](CacheOpInterface cacheOp) {
-        BlockArgument blockArg = traceCacheToBlockArg(cacheOp.getCache());
-        if (blockArg && blockArg.getOwner()->getParentOp() == funcOp) {
-          cacheArgs.insert(blockArg);
+        if (auto blockArg = llvm::dyn_cast<BlockArgument>(cacheOp.getCache())) {
+          if (blockArg.getOwner()->getParentOp() == funcOp) {
+            cacheArgs.insert(blockArg);
+          }
         }
       });
 
--- a/test/ttmlir/Dialect/TTIR/Transforms/InferKVCacheArgumentTypes/infer_kv_cache_argument_types.mlir
+++ b/test/ttmlir/Dialect/TTIR/Transforms/InferKVCacheArgumentTypes/infer_kv_cache_argument_types.mlir
@@ -33,22 +33,6 @@ module {
   }
 }
 
-// Test: chained cache ops trace back to original argument
-module {
-  // CHECK-LABEL: func.func @chained_cache_ops
-  // CHECK-SAME: tensor<1x32x64x512xbf16> {ttcore.kv_cache}
-  func.func @chained_cache_ops(
-      %cache: tensor<1x32x64x512xbf16>,
-      %input1: tensor<1x32x64x512xbf16>,
-      %input2: tensor<1x32x1x512xbf16>
-  ) -> tensor<1x32x64x512xbf16> {
-    %filled = "ttir.fill_cache"(%cache, %input1) <{batch_offset = 0: i32}> : (tensor<1x32x64x512xbf16>, tensor<1x32x64x512xbf16>) -> tensor<1x32x64x512xbf16>
-    %idx = "ttir.constant"() <{value = dense<63> : tensor<1xi32>}> : () -> tensor<1xi32>
-    %updated = "ttir.update_cache"(%filled, %input2, %idx) <{batch_offset = 0: i32}> : (tensor<1x32x64x512xbf16>, tensor<1x32x1x512xbf16>, tensor<1xi32>) -> tensor<1x32x64x512xbf16>
-    return %updated : tensor<1x32x64x512xbf16>
-  }
-}
-
 // Test: multiple caches in same function - both get marked
 module {
   // CHECK-LABEL: func.func @multiple_caches


include/ttmlir/Dialect/TTCore/IR/TTCoreOpsEnums.td
```diff
@@ -188,13 +188,15 @@ def TTCore_ArgumentType_Input : I32EnumAttrCase<"Input", 0, "input">;
...
 def TTCore_ArgumentType_Default : I32EnumAttrCase<"Default", 3, "default">;
+def TTCore_ArgumentType_KVCache : I32EnumAttrCase<"KVCache", 4, "kv_cache">;
 
 def TTCore_ArgumentType : I32EnumAttr<"ArgumentType", "Argument Type",
                             [
                               TTCore_ArgumentType_Input,
                               TTCore_ArgumentType_Parameter,
                               TTCore_ArgumentType_Constant,
                               TTCore_ArgumentType_Default,
+                              TTCore_ArgumentType_KVCache,
```
jameszianxuTT: This seems like a workaround that introduces a lot of complexity.

There are some side effects on the ArgType right now, and adding this additional ArgumentType::KVCache that acts like an input but is just not forced to row major seems like it will make problems in the future, eg. if people introduce additional passes that have side effect on ArgumentType::Input . Since your arg type inference happens ttir-ttnn there also needs to be consideration of arg type changing between passes.

From offline discussion, I understand the core idea of your change is to infer input tensor layout to avoid unnecessary to_layout being introduced, which ultimately models the (ttnn op constraint) expected layout for the first user of the input tensor. 

I don't think this exists right now, but I think it would be better to model these input tensor layout constraints in the op definitions and enforce them through a simple pass like extractInputFirstUserExpectedLayout, rather than introducing a new argument type with special treatment in some passes. 

Thanks for the detailed explanation and offline discussion :)

odjuricicTT: @jameszianxuTT Can you elaborate on what are the scenarios in which having the distinction between intpus and kv_cache would cause problems?

There are currently two usecases where the distinction helps. Row major forcing and handing inputs when trace is enabled. In both scenarios kv_cache should be treated differently then other inputs and it seems cleaner to have this info in the tensor instead of having to fetch it from consumer ops.

>Since your arg type inference happens ttir-ttnn there also needs to be consideration of arg type changing between passes

Agree on this part, idealy we would have this arg type set in the frontend at the same time when input / constant / param is defined.
@vkovacevicTT Is this an option?

jameszianxuTT: > @jameszianxuTT Can you elaborate on what are the scenarios in which having the distinction between intpus and kv_cache would cause problems?

There are places in which we have side effects based on the argumentType, like here in the [AnalyzeMesh pass](https://github.com/tenstorrent/tt-mlir/blob/28944ba89739ace8cf3c968739ec05a9983fb8d4/lib/Dialect/StableHLO/Transforms/AnalyzeMesh.cpp#L174). For the purposes of sharding annotation, we _do_ want KV cache to be treated the same way as an input. Technically in this case, since this is a shlo-shlo pass, it is not affected by Vladan's change which only alters arg type in a later pass. 

Adding a special argtype for kv cache that is different to input in some cases but not all, adds a lot of complexity when maintaining these side effects or adding new ones. Someone might add a branch on argumentType in the future that should apply to KV caches as well, but not realize that KV caches are special cased and cause hard to catch bugs.

> There are currently two usecases where the distinction helps. Row major forcing and handing inputs when trace is enabled. In both scenarios kv_cache should be treated differently then other inputs and it seems cleaner to have this info in the tensor instead of having to fetch it from consumer ops.

Understandable. Embedding this info in the tensor _somewhere_ is fine, I just don't think it should be by adding a special case argument type that is not really clearly divided from other inputs.

>Agree on this part, idealy we would have this arg type set in the frontend at the same time when input / constant / param is defined.

Theoretically yes in tt-xla this could be done as we have a pass to annotate argTypes which could be specialized to detect KV caches, though by previous comments I am not sure this is is a good idea.

sdjordjevicTT: Yeah, I agree with @jameszianxuTT. I don’t see a strong reason to introduce a new argtype, and I agree with all the points @jameszianxuTT mentioned.

Instead, I suggest we add a function attribute to tag the argument, for example:
```
func.func @main(%arg0: tensor<10x128x512xbf16> {ttcore.argument_type = #ttcore.argument_type<input>, ttcore.kv_cache, ....
```
This approach keeps the argument type unchanged while allowing us to attach additional metadata.

vkovacevicTT: @odjuricicTT 
Tagging it in the frontend can be a bit tricky since KV cache tensors are treated as regular user inputs, but I'm exploring some options.

odjuricicTT: Ok, makse sense. I like @sdjordjevicTT s proposal.

vkovacevicTT: Implemented @sdjordjevicTT's solution.

lib/Dialect/TTIR/Transforms/InferKVCacheArgumentTypes.cpp
```diff
@@ -0,0 +1,94 @@
...
+      TTIRInferKVCacheArgumentTypes>::TTIRInferKVCacheArgumentTypesBase;
+
+  void runOnOperation() final {
+    ModuleOp moduleOp = getOperation();
+    MLIRContext *context = moduleOp.getContext();
+
+    moduleOp.walk([&](func::FuncOp funcOp) {
+      llvm::DenseSet<BlockArgument> cacheArgs;
+
+      funcOp.walk([&](Operation *op) {
```
sdjordjevicTT: Instead of a cache op trait I suggested above, maybe we can define a CacheOpInterface, attach it to cache ops, and here only walk over the ops with the interface:
```
funcOp.walk([&](CacheOpInterface dma)
```

lib/Dialect/TTIR/Transforms/InferKVCacheArgumentTypes.cpp
```diff
@@ -0,0 +1,94 @@
...
+
+namespace mlir::tt::ttir {
+#define GEN_PASS_DEF_TTIRINFERKVCACHEARGUMENTTYPES
+#include "ttmlir/Dialect/TTIR/Transforms/Passes.h.inc"
+
+namespace {
+
+// Returns the cache operand for cache operations, or nullptr if not a cache op.
+Value getCacheOperand(Operation *op) {
+  return llvm::TypeSwitch<Operation *, Value>(op)
```
sdjordjevicTT: With the implemented interface I suggested below, we can just remove this helper method and use it in the function caller place:
```
Value cacheOperand = mlir::dyn_cast<CacheOpInterface>(op)
```

lib/Dialect/TTIR/Transforms/InferKVCacheArgumentTypes.cpp
```diff
@@ -0,0 +1,75 @@
...
+    Operation *defOp = current.getDefiningOp();
+    if (!defOp) {
+      return nullptr;
+    }
+
+    if (auto cacheOp = mlir::dyn_cast<CacheOpInterface>(defOp)) {
+      current = cacheOp.getCache();
+    } else {
+      return nullptr;
+    }
```
rpavlovicTT: is this part really needed? As, you call `traceCacheToBlockArg` with Cache tensor operand, do we expect it to trace through another cache op. That seems like a weird case, isn't it?

vkovacevicTT: Hm, good point. 

In a chain of cache ops:

```
...
%56 = "ttir.fill_cache"(%arg6, %55)
...
%58 = "ttir.fill_cache"(%56, %57)
...
%60 = "ttir.fill_cache"(%58, %59)
...
``` 
we would mark the first operand (%arg6) of first op, and then re-trace operands of next ops back to the same operand of first op.
So in that case it is redundant.

This could be necessary if cache tensor comes from a non-cache op like `ttir.reshape`.
Then we would need to trace through all ops, not just cache ops.

But currently I think there are no such cases, so we can discard tracing through ops and just mark the first operand of a kv_cache op if it is a function argument.

PR #6898, opened 2026-02-04
Add TTIR -> Linalg patterns (part 3)

dmilinkovicTT:
### Ticket
N/A

### Problem description
More ops need to be supported for CPU-hoisting, as some models are failing.

### What's changed
- Added a TTIR -> Linalg patterns for `AvgPool2DOp`, `GlobalAvgPool2DOp` and `PadOp`, and CPU-hoisted golden tests for these ops
- Added support for multi-result output producers in `HoistCPUOps`
- Added TTIR -> TTIR decomposition in CPUFallback mode for `SplitQueryKeyValueAndSplitHeadsOp`
- Added builder and golden infra for `SplitQueryKeyValueAndSplitHeadsOp`

### Checklist

### DIFF
--- a/lib/Dialect/TTIR/Transforms/HoistCPUOps.cpp
+++ b/lib/Dialect/TTIR/Transforms/HoistCPUOps.cpp
@@ -47,11 +47,6 @@ using OpsVectorType = llvm::SmallVector<mlir::Operation *>;
 using ValuesVectorType = llvm::SmallVector<mlir::Value>;
 using TypesVectorType = llvm::SmallVector<mlir::Type>;
 
-// Pair of (Operation*, result_index) to track which result of an operation
-// produces each output of the CPU-hoisted op set.
-using OutputProducerType = std::pair<mlir::Operation *, unsigned>;
-using OutputProducersVectorType = llvm::SmallVector<OutputProducerType>;
-
 // Helper function to get ranks of tensor values.
 // Used to populate attrs needed for tensor packing/unpacking operations.
 static llvm::SmallVector<int64_t>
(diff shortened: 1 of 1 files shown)


lib/Conversion/TTIRToTTIRDecomposition/TTIRToTTIRDecomposition.cpp
```diff
@@ -2360,6 +2360,214 @@ struct ArgMaxPattern : public OpConversionPattern<ttir::ArgMaxOp> {
 };
 } // namespace
 
+//===----------------------------------------------------------------------===//
+// SplitQueryKeyValueAndSplitHeadsOp decomposition
+//===----------------------------------------------------------------------===//
```
mtopalovicTT, line 2365: Not sure if this is going to work. If ttir fusing pass is after decomposition we might fuse this op and then the conversion to lingalg will fail.

dmilinkovicTT: This decomposition is applied in `CPUFallback` decomposition mode, which is being run only on CPU module in `createTTIRToLLVMCPUPipeline`. No fusing passes happen there

mtopalovicTT: Hmm okay so actually fusing is done only on device module so you won't fuse back which is fine in your case. But are you sure you are only decomposing on CPU module. From TTNNPipelines I see

```cpp
  pm.addPass(mlir::tt::createTTIRToTTIRDecompositionPass(decompOptions));
```
This is top level pass manager right, it will execute on initial module first right?

dmilinkovicTT: This op will get decomposed only in `CPUFallback` decomp mode, which currently does only happen in CPU module.

However, this wasn't the case up until #6789, so I assume you're referring to the situation before this PR?

mtopalovicTT: Yea my bad I'm out of sync locally thats what lead to confusion.

mtopalovicTT (approved):

lib/Dialect/TTIR/Transforms/HoistCPUOps.cpp
```diff
@@ -47,6 +47,11 @@ using OpsVectorType = llvm::SmallVector<mlir::Operation *>;
 using ValuesVectorType = llvm::SmallVector<mlir::Value>;
 using TypesVectorType = llvm::SmallVector<mlir::Type>;
 
+// Pair of (Operation*, result_index) to track which result of an operation
+// produces each output of the CPU-hoisted op set.
+using OutputProducerType = std::pair<mlir::Operation *, unsigned>;
```
vwellsTT: can you just use a map of Operation -> unsigned (or vice versa) and remove these? Seems better vs these aliases and then `llvm::contains` on a vector, AFAICT

dmilinkovicTT: turns out that IRMapping is sufficient for original -> CPU-hoisted value mapping, significantly simplifying the output handling... thanks!

Issue #6612, opened 2026-01-14
[d2m] move affine maps from `metal_layout` & `shard` attributes to `view_layout` and `empty` ops


nsmithtt:
A loosely related issue is that today, any index_map introduced by a stream_layout op (or a view_layout op) becomes part of the tensor’s type as part of the MetalLayoutAttr/ViewAttr.  While this does make it simple to see remapping throughout the IR, it makes type checking quite onerous; mismatches in mapping are not permitted anywhere where types are checked (e.g. func type vs return value, generic output vs return value, etc).

Going forward, it would be more convenient to instead encode the re-mapping on the stream_layout op itself; consumer ops would be forced to do use-def analysis to determine the mapping (but we already do this for other things today), and the MetalLayoutAttr would no longer carry mapping information.

Concretely, today’s IR:
```mlir
%stream = "d2m.stream_layout"(%alloc_1, %alloc_5) :
  (memref<1x1x4x1x!ttcore.tile<32x32, f32>, #ttcore.shard<4096x4096, 1>, #ttcore.memory_space<l1>>,
   memref<4x1x1x1x!ttcore.tile<32x32, f32>, #ttcore.shard<4096x4096, 1>, #ttcore.memory_space<l1>>) -> 
   memref<4x1x1x1x!ttcore.tile<32x32, f32>, #ttcore.view<(d0, d1, d2, d3) -> ((d0 + d2) floordiv 4, d1 + d3, (d0 + d2) mod 4, 0)>, #ttcore.memory_space<l1>>

d2m.generic {block_factors = [1, 1, 4], grid = #ttcore.grid<1x1>, ...}
        ins(%stream: memref<1x4x1x1x!ttcore.tile<32x32, f32>, #ttcore.view<(d0, d1, d2, d3) -> (d0 + d2, (d1 + d3) floordiv 4, 0, (d1 + d3) mod 4)>, #ttcore.memory_space<l1>>)
...
```

Would instead become something like:
```mlir
%view = "d2m.view_layout"(%alloc_1) {remapping=(d0, d1, d2, d3) -> ((d0 + d2) floordiv 4, d1 + d3, (d0 + d2) mod 4, 0)} :
  (memref<1x1x4x1x!ttcore.tile<32x32, f32>, #ttcore.shard<4096x4096, 1>, #ttcore.memory_space<l1>>, 
[...]

PR #6902, opened 2026-02-04
[D2M] Index map refactor

vwellsTT:
### Ticket
https://github.com/tenstorrent/tt-mlir/issues/6612

### Problem description
Currently, any affine mappings applied to a given tensor or memref become encoded as part of the `type`--this is undesirable in a variety of ways.

### What's changed
Remove affine maps from MetalLayoutAttr, ViewLayoutAttr, and ShardLayoutAttr.  Attach maps to view_layout and stream_layout ops, and virtual grid maps to d2m empty_op, memref.alloc etc.  Refactor any code which previously attached an affine map to a layout -- instead attach it to a view op.  Replace accesses for these index_maps with explicit queries for defining view/stream/tensor creation ops.
Update a whole lot of tests to use new syntax.

### Checklist

### DIFF
--- a/lib/Dialect/D2M/IR/D2MOps.cpp
+++ b/lib/Dialect/D2M/IR/D2MOps.cpp
@@ -1244,9 +1244,13 @@ void d2m::GenericOp::build(mlir::OpBuilder &builder,
       if (auto vgm = utils::getVirtualGridMapping(output)) {
         SmallVector<int64_t> physGridShape =
             d2m::utils::getPhysicalGridShape(output);
-        auto [_, invMap] = ttmlir::d2m::utils::grids::createCoreVirtMaps(
-            builder.getContext(), gridShape, physGridShape);
-        grid = builder.getAttr<ttcore::GridAttr>(gridShape, invMap);
+        // Only add virtualization to the grid when the physical grid shape
+        // differs from the logical grid shape.
+        if (!llvm::equal(gridShape, physGridShape)) {
+          auto [_, invMap] = ttmlir::d2m::utils::grids::createCoreVirtMaps(
+              builder.getContext(), gridShape, physGridShape);
+          grid = builder.getAttr<ttcore::GridAttr>(gridShape, invMap);
+        }
       }
 
       // 2. Check for a 2D→2D permutation reblocking on a ViewLayoutOp.
--- a/lib/Dialect/D2M/Transforms/LowerToLayout.cpp
+++ b/lib/Dialect/D2M/Transforms/LowerToLayout.cpp
@@ -951,10 +951,9 @@ public:
       auto targetRemapping = utils::getAssociatedRemapping(op.getOutput());
       bool remappingsDiffer = currentRemapping != targetRemapping;
 
-      // Compare virtualGridMappings: different TTNN shard strategies (e.g.
-      // height_sharded vs block_sharded) can produce identical MetalLayoutAttr
-      // types but still require a mapping change to preserve the shard strategy
-      // through the pipeline.
+      // Compare virtualGridMappings: different shard strategies (e.g.
+      // height_sharded vs block_sharded) produce different VGMs even when
+      // all other layout properties match.
       auto currentVGM = utils::getVirtualGridMapping(currentValue);
       auto targetVGM = utils::getVirtualGridMapping(op.getOutput());
       bool vgmsDiffer = currentVGM != targetVGM;
--- a/lib/Dialect/D2M/Utils/Utils.cpp
+++ b/lib/Dialect/D2M/Utils/Utils.cpp
@@ -195,9 +195,6 @@ SmallVector<int64_t> getPhysicalGridShape(Value tensorOrMemref) {
   // (virtualGridMapping attr) rather than on views or the layout attribute.
   // Check for an explicit virtualGridMapping first.
   if (auto vgm = utils::getVirtualGridMapping(tensorOrMemref)) {
-    // The virtualGridMapping is the inverse map (physical → virtual).
-    // Derive the physical grid shape from the virtual grid shape using
-    // the shared getPhysicalGridExtent utility.
     ttcore::DeviceLayoutInterface layout =
         ttcore::getDeviceLayout(tensorOrMemref);
     TT_assert(layout);
@@ -206,6 +203,12 @@ SmallVector<int64_t> getPhysicalGridShape(Value tensorOrMemref) {
     if (auto *definingOp = tensorOrMemref.getDefiningOp()) {
       ttcore::DeviceAttr device = ttcore::lookupDevice(definingOp);
       auto workerGridShape = device.getWorkerGrid().getShape();
+      // If the grid fits directly on the device, the physical grid shape
+      // equals the virtual grid shape — no volume factorization needed.
+      if (!ttmlir::d2m::utils::grids::requiresVirtualGrid(gridShape,
+                                                          workerGridShape)) {
+        return SmallVector<int64_t>(gridShape.begin(), gridShape.end());
+      }
       return ttmlir::d2m::utils::grids::getPhysicalGridExtent(gridShape,
                                                               workerGridShape);
     }
@@ -260,6 +263,20 @@ std::optional<AffineMap> getVirtualGridMapping(Value val) {
     // Trace through d2m.to_layout to its output EmptyOp.
     if (auto toLayoutOp = mlir::dyn_cast<ToLayoutOp>(defOp)) {
       return getVirtualGridMapping(toLayoutOp.getOutput());
+    }
+
+    // Trace through d2m.generic results to the corresponding output operand.
+    // VGMs live on EmptyOps, so we need to explicitly trace from the result
+    // to the output operand that produced it.
+    if (auto genericOp = mlir::dyn_cast<GenericOp>(defOp)) {
+      // Find which result index this value corresponds to.
+      for (auto [idx, result] : llvm::enumerate(genericOp.getResults())) {
+        if (result == val) {
+          Value outputOperand = genericOp.getOutputs()[idx];
+          return getVirtualGridMapping(outputOperand);
+        }
+      }
+      return std::nullopt;
     }
 
     // Trace through d2m.view_layout to its input.
--- a/lib/Dialect/D2M/Transforms/LowerLoadStoreOpsToExplicitCBForm.cpp
+++ b/lib/Dialect/D2M/Transforms/LowerLoadStoreOpsToExplicitCBForm.cpp
@@ -205,13 +205,13 @@ static void simplifyLoadStorePairs(ModuleOp moduleOp, IRRewriter &rewriter) {
     bool isRemoteStore = isRemoteOperand(storeMemref, storeOp.getOperation());
 
     TT_assert(!(isRemoteLoad && isRemoteStore));
-    if (isRemoteLoad) {
+    if (!isRemoteStore) {
       // Create the explicit CB form of remote_load (no localBuffer, has CB
       // operand)
       rewriter.create<RemoteLoadOp>(loc, loadMemref, loadOp.getIndices(),
                                     outputCB, loadOp.getMcastStartIndex(),
                                     loadOp.getMcastShape());
-    } else if (isRemoteStore) {
+    } else {
       rewriter.create<RemoteStoreOp>(loc, storeMemref, loadOp.getIndices(),
                                      inputCB);
     }


test/python/golden/test_metal_dma.py
```diff
@@ -33,6 +33,7 @@ def compile_dma_test(test_func, shape, request, device):
         target="ttmetal",
         device=device,
         custom_pipeline=pipeline,
+        save_artifacts=True,
```
jgrimTT: Heads up, I try to avoid having these args in the test files themselves because if you use `--save-artifacts` in the pytest CLI, it'll throw an error that you're passing in the arg twice

vwellsTT: Yeah I meant to get rid of this once I was done debugging but I missed it, thanks!

vwellsTT: imo we should try to enforce not allowing this sort of thing in the tests themselves, probably adds a fair amount of IO to the test runners etc.

lib/Dialect/D2M/IR/D2MOps.cpp
```diff
@@ -1091,7 +1117,15 @@ void d2m::GenericOp::build(mlir::OpBuilder &builder,
 
   if (!grid) {
     auto output = outputs[0];
-    auto gridShape = ttcore::getGridShape(output);
+    SmallVector<int64_t> gridShape;
+    if (ttcore::hasDeviceLayout(output)) {
```
bgrady-tt: Does it make sense to support GenericOp that don't have outputs with layout conforming to DeviceLayoutInterface? Without DeviceLayoutInterface, these would be host tensors right?

vwellsTT: Yeah probably this could be an assert

lib/Dialect/D2M/Utils/Utils.cpp
```diff
@@ -165,77 +168,220 @@ SmallVector<int64_t> getPhysicalGridShape(Value tensorOrMemref) {
   if (auto viewOp = tensorOrMemref.getDefiningOp<d2m::ViewOpInterface>()) {
     ttcore::DeviceAttr device = ttcore::lookupDevice(viewOp);
     auto deviceGridShape = device.getWorkerGrid().getShape();
-    auto outputGridShape = ttcore::getGridShape(tensorOrMemref);
+    SmallVector<int64_t> outputGridShape;
+    if (ttcore::hasDeviceLayout(tensorOrMemref)) {
+      outputGridShape = llvm::to_vector(ttcore::getGridShape(tensorOrMemref));
+    } else {
```
bgrady-tt: Also not sure here if this should be supported. The physical grid shape only makes sense for a device tensor with a valid layout attribute.

lib/Dialect/TTCore/IR/Utils.cpp
```diff
@@ -171,6 +173,62 @@ llvm::SmallVector<int64_t, 2> collapseGridTo2D(ArrayRef<int64_t> gridShape) {
...
+  }
+
+  // Try to find an optimal factorization (matches main's behavior).
+  // This finds factors near sqrt to balance Y and X dimensions.
+  auto result = findLegalPhysicalGridForVolume(volume, deviceGridShape);
+  if (!result.empty()) {
+    return result;
+  }
+
+  // Fallback: if no factorization fits (time-multiplexing needed),
```
bgrady-tt: I think this indicates that something is wrong; this shouldn't ever happen with (current version of) virtual grids. It is currently limited to reassigning N shards to M L1 mem banks, but it cannot do so if N > M.

bgrady-tt: Basically, the reblocking affine maps (i.e. any affine map that changes shard dims) should be factored away from virtual grid maps (i.e. affine map that rearranges shard->bank mapping).

vwellsTT: Ok, honestly wasn't sure on virtual grid state so I wanted to be conservative here, but I'll make this assert too then

bgrady-tt: To be clear, I am not saying we _can't_ do this. But the old virtual grids design verification was tightly scoped to handle constant grid volume mappings; I don't know if allowing these combined mappings will work reliably.

vwellsTT: yeah, I mean I don't see any benefit to trying to handle inputs we don't actually support, so I'd rather take your suggestion

lib/Dialect/D2M/IR/D2MOps.cpp
```diff
@@ -1005,23 +1040,35 @@ bool d2m::ViewLayoutOp::isReblockOnly() {
...
-    return getInput();
+    auto inputRemapping = utils::getAssociatedRemapping(getInput());
+    if (inputRemapping.has_value() && *inputRemapping == getRemapping()) {
+      return getInput();
+    }
+    // If the input has no associated remapping and types match, this view is
+    // generally a no-op — unless the input is in DRAM.  Views on DRAM values
+    // serve as DMA markers in the lowering pipeline (isRemoteOperand checks
+    // for ViewOpInterface to decide whether to emit remote load/store ops).
+    // TODO: Fix isRemoteOperand to check memory space directly instead of
```
missing username/bug in TODO
```suggestion
    // TODO(nsmith): Fix isRemoteOperand to check memory space directly instead of
```

lib/Dialect/D2M/Utils/VirtualGrid.cpp
```diff
@@ -91,4 +92,35 @@ createCoreVirtMaps(mlir::MLIRContext *context,
...
+  auto result =
+      mlir::tt::d2m::utils::findLegalPhysicalGridForVolume(volume, targetGrid);
+  if (!result.empty()) {
+    return result;
+  }
+
+  // Fallback for time-multiplexing: use row-major flatten + reshape.
+  // X dimension fills first (up to targetGrid[1]),
+  // Y dimension extends as needed (up to targetGrid[0]).
+  int64_t physicalY =
```
bgrady-tt: Still think this logic doesn't make sense here. The physical grid extent shouldn't be conflated with time-multiplexing?

vwellsTT: Hmm. I definitely added this deliberately at some point to fix some failing tests, but I don't remember which/whether some later fixes would resolve.  (And the comment itself + fix was definitely LLM-generated).  I will see if I can just remove this "fallback" completely or not

include/ttmlir/Dialect/D2M/Utils/Utils.h
```diff
@@ -48,13 +52,53 @@ SmallVector<Value> buildGridIndices(OpBuilder &builder, Location loc,
 // grid shape used if the tensor/memref was the output of a GenericOp.
 SmallVector<int64_t> getPhysicalGridShape(Value tensorOrMemref);
 
+// Returns the remapping associated with a value, if any.
+// Traces back through the defining op to find a ViewLayoutOp or StreamLayoutOp
+// and returns its remapping attribute. Returns std::nullopt if the value has
+// no associated remapping.  After the virtual-grid refactor, these remappings
+// are always reblockings — virtual grid info lives on EmptyOp attrs instead.
```
nsmithtt, line 65: Is it recursive or just returns the first one? i.e. chained views

lib/Dialect/D2M/Utils/VirtualGrid.cpp
bgrady-tt: Based on use, this function should definitely assert if `findLegalPhysicalGridForVolume` fails. the physical grid shape is supposed to be a 2D shape that fits within the targetGrid. It doesn't make sense for this to be time multiplexed; blocking factors are supposed to do this.

nsmithtt: ~~Shouldn't we assume that `findLegalPhysicalGridForVolume` must return something?  Otherwise it implies that something illegal was chosen to begin with.~~

+1 to what Brett said

lib/Dialect/D2M/Transforms/LowerToLayout.cpp
```diff
@@ -237,25 +237,26 @@ class D2MLowerToLayoutRewriter : public OpRewritePattern<ToLayoutOp> {
...
+    RankedTensorType createDeviceType(RankedTensorType systemType,
+                                      ttcore::MetalLayoutAttr referenceLayout,
+                                      RankedTensorType referenceType,
+                                      ArrayRef<int64_t> targetGridShape) {
       SmallVector<int64_t> tensorGridShape =
           llvm::to_vector(referenceLayout.getGridShape(referenceType));
 
       bool virtualBounceNeeded = tensorGridShape.size() > 2 ||
                                  ((tensorGridShape[0] > targetGridShape[0]) ||
                                   (tensorGridShape[1] > targetGridShape[1]));
```
bgrady-tt: nit: I think this exact `requiresVirtualGridCollapse` should be made into a utility function and called in all the relevant places.

lib/Dialect/D2M/Transforms/LowerMulticastLoads.cpp
```diff
@@ -182,16 +185,22 @@ class LowerMulticastLoadsRewriter : public OpRewritePattern<RemoteLoadOp> {
...
-    if (!outputShardLayout.getCoreVirtualizationMap().isEmpty()) {
-      // We project out the shard layout dims and results from the indexing
-      // map before applying since we are only concerned with the grid
-      // dimensions.
-      auto coreVirtMap = outputShardLayout.getCoreVirtualizationMap();
+    if (!grid.getMapping().isEmpty()) {
+      // Derive the physical 2D grid and compute the forward map
+      // (virtual → physical) to convert virtual multicast shapes.
+      ttcore::DeviceAttr device = ttcore::lookupDevice(genericOp);
+      auto physGridShape = utils::collapseToPhysicalGrid2D(
```
bgrady-tt, line 198: This replaces the existing virtualization map on outputShardLayout with a canonical collapse produced by `createCoreVirtMaps`. This will likely break for TTNN integration, which may create layouts with customized core virtualization maps that don't align with what is produced by `createCoreVirtMaps`

lib/Dialect/D2M/Utils/VirtualGrid.cpp
vwellsTT: Yeah, makes sense.  Locally that seems to work fine now (on the set of pytests I remember specifically having problems with), updated.

lib/Dialect/D2M/IR/D2MOps.cpp
```diff
@@ -74,10 +98,18 @@ mlir::LogicalResult d2m::EmptyOp::bufferize(
...
-      mlir::cast<MemRefType>(
-          *getBufferType(getResult(), options, state, invocationStack)));
+  auto bufferType = mlir::cast<MemRefType>(
+      *getBufferType(getResult(), options, state, invocationStack));
+  auto allocOp = rewriter.create<memref::AllocOp>(getLoc(), bufferType);
+
+  // Propagate virtualGridMapping as a discardable attribute on memref::AllocOp
+  // (we don't own AllocOp so we can't add a declared attribute).
+  if (auto vgm = getVirtualGridMappingAttr()) {
+    allocOp->setAttr("virtualGridMapping", vgm);
```
bgrady-tt: use d2m namespace here for attribute i.e. "d2m.virtualGridMapping"

lib/Dialect/D2M/Transforms/LowerMulticastLoads.cpp
vwellsTT: Good point, hmm.  I'm not super familiar with the TTNN flow, but I suppose querying from annotations on memref.allocs (which should exist after we bufferize virtual grids) would be more robust--would that handle TTNN's cases too?

lib/Dialect/D2M/Transforms/LowerToLayout.cpp
vwellsTT: good idea

include/ttmlir/Dialect/D2M/Utils/Utils.h
vwellsTT: non-recursive today; I'd expect chained view to be folded (today), so it seemed unneeded (and when we update view folding, this could be updated as well, imo).
Let me know if you disagree, I don't think chaining would be hard or anything.  Or perhaps I should just clarify current state (nonchaining) in the comment?

lib/Dialect/D2M/IR/D2MOps.cpp
vwellsTT: I should probably take my own advice and put this in a header/tablegen too in fact :)

include/ttmlir/Dialect/D2M/Utils/Utils.h
vwellsTT: I took the latter approach for now, but happy to revisit if you prefer chaining

nsmithtt: Not all views will be collapsed, only views that do trivial reblocking are to avoid affine map hell / type checking issues.  Yeah I think just clarify in the comment could be good, whatever is a useful primitive to have makes sense, but just wanted to be explicit stated.

vwellsTT: Makes sense.  Someone merged some upstream tests which break my tests on successive stream_layout ops actually this afternoon (😢 ) so I might need to revisit this after I finish debugging

azecevicTT (approved):
TTCore change LGTM

lib/Dialect/D2M/IR/D2MOps.cpp
```diff
@@ -106,7 +106,7 @@ mlir::LogicalResult d2m::EmptyOp::bufferize(
   // Propagate virtualGridMapping as a discardable attribute on memref::AllocOp
   // (we don't own AllocOp so we can't add a declared attribute).
   if (auto vgm = getVirtualGridMappingAttr()) {
-    allocOp->setAttr("virtualGridMapping", vgm);
+    allocOp->setAttr(d2m::utils::kVirtualGridMappingAttr, vgm);
```
bgrady-tt, line 109: Attribute string namespace should be in form "d2m.utils.kVirtualGridMappingAttr" (periods, not colons). This format is actually relied on by MLIR to namespace built-in attributes.

bgrady-tt (approved):
Approved! Looks good 👍

tapspatel (approved):

PR #6908, opened 2026-02-04
Generalize CacheFillUpdatePattern to also handle constant and unsqueezed indices

### LINKED ISSUE #6907: Scatter to UpdateCache conversion pattern is not entirely general
Two minor limitations:
1) it expects the indicies to be a result of an argument to a graph(checks for a blockArg trough the use-def chain)
2) it doesn't handle unsqueezed tensors for indices

sgligorijevicTT:
### Ticket
closes #6907

### Problem description
Turning on experimental compile brought down a unit test which uncovered some limitations of the pattern

### What's changed
Generalized the pattern

Heads up to perf team, I insert a reshape in one case. I expect cases that already worked to not enter that branch, and I expect the reshape to get folded/EIOd away, but there's a small chance is pops up and impacts perf.

PR in draft until 1) I run some test suites to validate correctness 2) Potentially add more tests

### Checklist


jameszianxuTT:
cc @jazpurTT

PR #6921, opened 2026-02-05
Add support for d2m.experimental.force_virtual_grid_layout attribute

bgrady-tt:
Adds support for detecting a string attribute `"d2m.experimental.force_virtual_grid_layout"` that forces selection of virtual grid layouts for all operands on the corresponding op. This attribute is propagated from named ttir ops to d2m.generic ops.

### DIFF
--- a/lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
+++ b/lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
@@ -792,6 +792,13 @@ private:
         loc, inputs, outputs, rewriter.getAffineMapArrayAttr(indexingMaps),
         rewriter.getArrayAttr(iteratorTypes));
 
+    // Propagate d2m::experimental::force_virtual_grid_layout attribute if
+    // present
+    if (auto attr =
+            op->getAttr("d2m::experimental::force_virtual_grid_layout")) {
+      generic->setAttr("d2m::experimental::force_virtual_grid_layout", attr);
+    }
+
     // Create one bb in 'generic''s region and set its arguments.
     auto insertPoint = rewriter.saveInsertionPoint();
     rewriter.startOpModification(generic);
@@ -939,6 +946,13 @@ private:
     auto generic = rewriter.create<d2m::GenericOp>(
         loc, inputs, outputs, rewriter.getAffineMapArrayAttr(indexingMaps),
         rewriter.getArrayAttr(iteratorTypes));
+
+    // Propagate d2m::experimental::force_virtual_grid_layout attribute if
+    // present
+    if (auto attr =
+            op->getAttr("d2m::experimental::force_virtual_grid_layout")) {
+      generic->setAttr("d2m::experimental::force_virtual_grid_layout", attr);
+    }
 
     // Create one bb in 'generic''s region and set its arguments.
     auto insertPoint = rewriter.saveInsertionPoint();
@@ -1215,6 +1229,13 @@ private:
     auto generic = rewriter.create<d2m::GenericOp>(
         loc, inputs, outputs, rewriter.getAffineMapArrayAttr(indexingMaps),
         rewriter.getArrayAttr(iteratorTypes));
+
+    // Propagate d2m::experimental::force_virtual_grid_layout attribute if
+    // present
+    if (auto attr =
+            op->getAttr("d2m::experimental::force_virtual_grid_layout")) {
+      generic->setAttr("d2m::experimental::force_virtual_grid_layout", attr);
+    }
 
     // Create one bb in 'generic''s region and set its arguments.
     auto insertPoint = rewriter.saveInsertionPoint();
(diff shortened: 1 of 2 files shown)


lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
```diff
@@ -792,6 +792,13 @@ class D2MNamedElementwiseRewriter final
         loc, inputs, outputs, rewriter.getAffineMapArrayAttr(indexingMaps),
         rewriter.getArrayAttr(iteratorTypes));
 
+    // Propagate d2m::experimental::force_virtual_grid_layout attribute if
```
vwellsTT, line 795: nit: fix comments.
And this should probably be a tiny func vs copy-pasted 5x imo

lib/Dialect/D2M/Transforms/GridSelection.cpp
```diff
@@ -226,11 +226,11 @@ shouldImplementAsVirtualGrid(RankedTensorType tensorType,
...
 }
 
-static std::pair<llvm::SmallVector<int64_t>, bool>
-computeOptimalGrid(mlir::RankedTensorType tensorType,
-                   ArrayRef<int64_t> physicalShape,
-                   ArrayRef<int64_t> targetSquareGridShape) {
-  if (shouldImplementAsVirtualGrid(tensorType, physicalShape,
+static std::pair<llvm::SmallVector<int64_t>, bool> computeOptimalGrid(
+    mlir::RankedTensorType tensorType, ArrayRef<int64_t> physicalShape,
+    ArrayRef<int64_t> targetSquareGridShape, bool forceVirtualGrid = false) {
```
vwellsTT, line 231: AFAICT you're updating all the callsite anyway, can we get rid of default param value?

lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
vwellsTT: I wonder if there are any other attrs we care about--might be better to have some general-purpose attr forwarding mechanism vs doing 1 at a time also

vwellsTT:
I also think it'd be better practice to create a tablegen-ic unitAttr, vs using string literals like this, FWIW--seems bad to have raw string literals in separate files (I guess at a minimum it should be constexpr'ed in shared header or something)

nsmithtt (commented):
Could we just make this the default?  Why aren't all grids virtual? i.e. let's get rid of all branches that check for "virtual grids" a passthrough virtual grid is just an identity map.

vwellsTT:
> Could we just make this the default? Why aren't all grids virtual? i.e. let's get rid of all branches that check for "virtual grids" a passthrough virtual grid is just an identity map.

I guess on this note, I do feel like "virtual" check are ballooning a lot of the code in LowerToLayout, GridSelection, etc. unfavorably.  I remember we talked about why LowerToLayout might care w.r.t. host data transfers, but I'd be in favor of trying to simplify the logic as much as we can too

bgrady-tt:
@nsmithtt I guess the intent and naming is a little off. This is basically an attribute to force **_HW sharding_** vs block sharding for 2D tensors, since ND tensors are forced to use virtual grids anyway.

bgrady-tt:
wrt having only virtual grids (and permit identity mappings), I would agree refactoring to support that is a good idea. But probably out of the scope of this PR :)

lib/Dialect/D2M/Transforms/GridSelection.cpp
bgrady-tt: sure I can get rid of the default.

lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
bgrady-tt: I can clean this up to have a single helper function here that 'forwards' all 'd2m.experimental' attributes on ttir ops and tag them to the d2m.generic.

bgrady-tt:
@vwellsTT As far as having a tablegen defined bool attribute, I can just add that to the generic op here. I don't want to do that for _ttir ops_, because this attribute is meaningless outside of D2M scope; its basically metadata for our test infrastructure.

lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
vwellsTT: yeah, I think something like that would be my vote, seems like this sort of thing could easily have more use-cases in future (and good infra to have, I wish MLIR had better attr-forwarding during conversion)

vwellsTT:
> @vwellsTT As far as having a tablegen defined bool attribute, I can just add that to the generic op here. I don't want to do that for _ttir ops_, because this attribute is meaningless outside of D2M scope; its basically metadata for our test infrastructure.

My comment is more a nit regarding defining the attr as just a string literal; I think it's better to specifically encode attrs vs creating them on the fly and check via strings; that way you can access via something like `D2M::ForceVirtualAttr::name`
(way ago I did something like this for cpu hoisting, see `ttir::ShouldHoistAttr` as an example).

It's just a style nit but I think it's better practice (I think Stefan or someone suggested I use that when I was doing raw strings like this), and it's helpful for organizing this kind of attr if we have more etc. to have them all defined in a central place

nsmithtt:
> @nsmithtt I guess the intent and naming is a little off. This is basically an attribute to force **_HW sharding_** vs block sharding for 2D tensors, since ND tensors are forced to use virtual grids anyway.

Yeah, maybe let's update the attribute to include height_width in the name or something.  I think we should prioritize the removal of virtual grid if statements and use virtual grid maps unconditionally.  After that refactor the naming will be even more confusing.

bgrady-tt:
This was definitely a half-baked idea, so let's abandon it 👍

Issue #6885, opened 2026-02-04
Issues with handling constant non-splat tensors in shardy


umalesTT:
Currently, some multichip workloads fail with
``` Shardy automatic parallelization currently does not support non-splat constant tensors. ```
in ``` UpdateGlobalToLocalShapes ```, when we try to shard constant tensor that has different values across elements. 

Tensors for which this happens are small and possible fix is to just keep all constant tensor replicated.
This issue happened after [this fix](https://github.com/tenstorrent/tt-mlir/issues/6756)

vladimirjovanovicTT:
Assigning P0 because this is a customer issue - blocks finetuning runs by Evol labs

ndrakulicTT:
For the repro:
1. go to tt-xla
2. checkout `ndrakulic/non_splat` (small change in compiler options for training)
3. run the test `pytest -svv tests/runner/test_models.py::test_all_models_torch[qwen_3/causal_lm/pytorch-0_6b-tensor_parallel-training]`

Workaround for this is to increase num of batches to be grater then 2 per device ( change in `third_party/tt_forge_models/qwen_3/causal_lm/pytorch/loader.py` `def load_inputs(self, dtype_override=None, batch_size=X):`

When batch_size is 4 then it fails with some OOM in runtime (this should be unreleated)

sshonTT:
@umalesTT, @ndrakulicTT 
I've opened a PR [here](https://github.com/tenstorrent/tt-mlir/pull/6928).

With this change, qwen_3/causal_lm/pytorch-0_6b-tensor_parallel-training runs end-to-end, although there is a PCC failure.

Note on PCC Failure, I verified that the PCC error persists even when forcing constant ops to be replicated (bypassing the logic in this patch). This suggests the issue is unrelated to this PR.

However, could you please verify this on your end to confirm?

umalesTT:
@sshonTT 
thanks for the quick fix!
This does fix significant scope of problems with non-splat tensors, but unfortunately not everything - some of our training workloads still fail on this issue, but now with non-splat and non-periodic constant tensor, e.g. ```[0, 1, ..., 63]``` - where shardy tries to shard them and then fails.

[...]

PR #6928, opened 2026-02-05
Allow sharding for periodic non-splat constants

sshonTT:
### Ticket
[Link to Github Issue](https://github.com/tenstorrent/tt-mlir/issues/6885)

### Problem description
Previously, GlobalToLocalShape did not support sharding for non-splat constant tensors. However, if the constant values are periodic (e.g., broadcasted) along the sharding axis, it is safe to shard because every device receives the identical slice.

### What's changed
This patch implements a periodicity check based on the sharding specification. If the constant is confirmed to be periodic, we allow the sharding operation so that each device can utilize the same data slice.

### Checklist


test/ttmlir/Dialect/StableHLO/global_to_local_shape/non_splat_constant_sharding_error.mlir
```diff
@@ -0,0 +1,17 @@
+// REQUIRES: stablehlo
+// RUN: not ttmlir-opt --stablehlo-pipeline %s -verify-diagnostics
+
+// Test that non-splat constant with non-periodic values on sharding dimension produces an error
+module @NonSplatConstantSelectOpError attributes {mhlo.cross_program_prefetches = [], mhlo.frontend_attributes = {xla.sdy.meshes = "{mesh = #sdy.mesh<[\22_axis_0\22=8]>}"}, mhlo.input_output_alias = [], mhlo.is_dynamic = false, mhlo.use_auto_spmd_partitioning = false} {
+  func.func @main(
```
mmanzoorTT, line 6: Please add a `CHECK` statement to verify the error. This way test will fail if the error message is not a match and test failed due to some other reason.

sshonTT: Yeah, I didn't know that I could do it this way. Thanks!

mmanzoorTT: For reference: https://github.com/tenstorrent/tt-mlir/blob/main/test/ttmlir/Dialect/TTIR/sort/sort_tests_negative.mlir

There are many other examples.

sshonTT: yup, I just added it. please see below. Thanks!

mmanzoorTT: Thanks.

PR #6930, opened 2026-02-05
Decompose all_reduce into recude_scatter + all_gather if beneficial

AleksKnezevic:
### Ticket
[Link to Github Issue](https://github.com/tenstorrent/tt-xla/issues/3047)

### Problem description
we always preform all_reduce as reduce_scatter followed by all_gather. It can be beneficial to split them up since the ag can be commuted through eltwise/reshape ops and combined in some cases. 

### What's changed
Added a pass to split ar into rs+ag when the ag's can be commuted. 

### Checklist


AleksKnezevic:
I was initially thinking to do this in SHLO but moved it to TTIR after today's tech sync. @mrakitaTT FYI 😄

Issue #6926, opened 2026-02-05
ttnn.sort fails with f32 input: "Input tensor data type must be BFLOAT16 or UINT16"


kmabeeTT:
### Describe the bug

(blocks tt-xla: https://github.com/tenstorrent/tt-xla/issues/3089, https://github.com/tenstorrent/tt-xla/issues/3027)

The ttnn.sort operation fails at runtime when it receives f32 input tensors, but the hardware implementation only supports BFLOAT16 or UINT16 data types. There was new assert added in tt-metal via https://github.com/tenstorrent/tt-metal/pull/36467, previously was low PCC. The compiler currently does not insert the necessary type conversion in the `TTNNWorkaroundsPass`.

```
  Error:
  TT_FATAL: Input tensor data type must be BFLOAT16 or UINT16, got DataType::FLOAT32
```

  Affected operations:
  - torch.topk() - uses sort for finding top-k values/indices
  - vLLM sampling operations - uses sort for nucleus/top-p sampling (via https://github.com/tenstorrent/tt-xla/issues/3027) in all models
  - Any other code path that uses ttnn.sort with f32 tensors, like tt-xla topk unit tests (also failing due to this)

  Root cause:

  The `createSortOpOperandsWorkarounds()` function in `lib/Dialect/TTNN/IR/TTNNWorkaroundsPass.cpp` (lines 820-843) currently only adds a workaround for the output indices dtype (f32→ui16) but does not handle
  input dtype conversion. When f32 tensors flow into sort (either from PyTorch inputs or intermediate computations), no automatic conversion to bf16 is inserted.

### Steps to reproduce the issue

Including TTIR graph from the opt/125M model here as attachment since large, and dedicated topk unit test, I confirmed can reproduce the signature when generating and running flatbuffer for both cases.   topk test is simpler.

[topk_test_xla_debug.log](https://github.com/user-attachments/files/25102632/topk_test_xla_debug.log)
[test_opt_generation_rm_hack_feb4_xla_debug2.log.gz](https://github.com/user-attachments/files/25102633/test_opt_generation_rm_hack_feb4_xla_debug2.log.gz)

FIXME - IR and ttmlir-opt cmds coming next

### Expected behavior

No assertion failure, and the compiler should automatically insert a ttnn.typecast operation to convert f32 inputs to bf16 before passing them to ttnn.sort, similar to how it handles the output indices conversion.

```
  Expected transformation:
  // Before (current - fails):
[...]

PR #6931, opened 2026-02-05
[TTNN] Add f32 input workaround for sort operation

kmabeeTT:
### Ticket
Closes https://github.com/tenstorrent/tt-mlir/issues/6926

### Problem description
The `ttnn.sort` operation in tt-metal only supports BFloat16 or UInt16 input tensors, but TTIR graphs were passing Float32 inputs. This caused runtime failures with the error:

```
TT_FATAL: Input tensor data type must be BFLOAT16 or UINT16, got DataType::FLOAT32
```

This affected operations like top-k sorting and top-p sampling used in LLM text generation. The TTNNWorkaroundsPass was only handling output indices type conversion but not input tensor type conversion.

### What's changed
Modified `createSortOpOperandsWorkarounds()` in TTNNWorkaroundsPass to automatically detect f32 input tensors and insert typecast operations to bf16 before the sort operation.

**Key Changes**:
 - Check input element type in `createSortOpOperandsWorkarounds()`
 - Insert `tensorDataTypeWorkaround = BFloat16` when input is f32
 - Insert `tensorDataTypeWorkaround = UInt16` when input is integer (not ui16)
 - Apply same conversion to output values tensor (must match input type)
 - Added comprehensive test coverage (new f32 and si32 input to sort tests)

**Impact**: Enables f32-based operations like top-k sorting and top-p sampling to work correctly. The compiler now automatically inserts necessary type conversions, eliminating runtime errors in LLM text generation.

### Checklist
- [x] New/Existing tests provide coverage for changes (2 new tests)

### DIFF
--- a/lib/Dialect/TTNN/IR/TTNNWorkaroundsPass.cpp
+++ b/lib/Dialect/TTNN/IR/TTNNWorkaroundsPass.cpp
@@ -822,7 +822,7 @@ TTNNOperandsWorkaroundsFactory::createReduceProdOpOperandsWorkarounds() {
 // expected data type will cause runtime to assert.
 // Issue page: https://github.com/tenstorrent/tt-mlir/issues/4405
 // tt-metal also only supports BFloat16 or UInt16 for input tensors, not
-// Float32. Issue page: https://github.com/tenstorrent/tt-mlir/issues/6926
+// Float32. Issue page: https://github.com/tenstorrent/tt-metal/issues/37322
 TTNNOperandsWorkarounds
 TTNNOperandsWorkaroundsFactory::createSortOpOperandsWorkarounds(
     ttnn::SortOp op) {
--- a/test/python/golden/ttir_ops/data_movement/test_data_movement.py
+++ b/test/python/golden/ttir_ops/data_movement/test_data_movement.py
@@ -396,7 +396,11 @@ def test_slice(
 
 # Sort tests
 @pytest.mark.parametrize("shape", [(1, 64, 64)], ids=shape_str)
-@pytest.mark.parametrize("dtype", [torch.bfloat16], ids=["bf16"])
+@pytest.mark.parametrize(
+    "dtype",
+    [torch.bfloat16, torch.float32],
+    ids=["bf16", "f32"],
+)
 @pytest.mark.parametrize("dim", [0, 1, 2])
 @pytest.mark.parametrize("descending", [True, False])
 @pytest.mark.parametrize("stable", [True, False])
--- a/test/python/golden/ttir_ops/workarounds/test_workarounds.py
+++ b/test/python/golden/ttir_ops/workarounds/test_workarounds.py
@@ -283,3 +283,46 @@ def test_sdpa_decode_no_workaround(
     )
 
     assert check_op(output, "scaled_dot_product_attention")
+
+
+@pytest.mark.parametrize("shape", [(1, 64, 64)], ids=shape_str)
+@pytest.mark.parametrize("dtype", [torch.float32], ids=["f32"])
+@pytest.mark.parametrize("dim", [1, 2])
+@pytest.mark.parametrize("target", ["ttnn"])
+@pytest.mark.xfail(
+    reason="Sort with float32 input requires workaround to convert to bfloat16. Metal issue: https://github.com/tenstorrent/tt-metal/issues/37322"
+)
+def test_sort_without_workaround(
+    shape: Shape,
+    dtype: torch.dtype,
+    dim: int,
+    target: str,
+    request,
+    device,
+):
+    """
+    Test sort operation with float32 input and workarounds disabled.
+    Should fail because metal sort expects bfloat16 input.
+    """
+
+    def module(builder: TTIRBuilder):
+        @builder.func([shape], [dtype])
+        def sort_wrapper(
+            in0: Operand, builder: TTIRBuilder, unit_attrs: Optional[List[str]] = None
+        ):
+            sort_0_values, sort_0_indices = builder.sort(
+                in0,
+                dim=dim,
+                descending=False,
+                stable=False,
+                unit_attrs=unit_attrs,
+            )
+            return sort_0_values
+
+    compile_and_execute_ttir(
+        module,
+        **get_request_kwargs(request),
+        device=device,
+        target=target,
+        pipeline_options=["disable-workarounds=true"],
+    )


kmabeeTT:
Some more info

## Root Cause Details

**Where it failed**: `createSortOpOperandsWorkarounds()` in `lib/Dialect/TTNN/IR/TTNNWorkaroundsPass.cpp:820-842`

**Why it failed**: The workaround factory only checked and converted output indices type (si32→ui16), but didn't check or convert input tensor types (f32→bf16).

## MLIR Transformation Example

**Before (fails at runtime)**:
```mlir
%values, %indices = "ttir.sort"(%input_f32) : (tensor<1x10xf32>) -> (tensor<1x10xf32>, tensor<1x10xi32>)
// Compiles to:
%values, %indices = "ttnn.sort"(%input_f32)  // ERROR: f32 not supported!
```

**After (works correctly)**:
```mlir
%values, %indices = "ttir.sort"(%input_f32) : (tensor<1x10xf32>) -> (tensor<1x10xf32>, tensor<1x10xi32>)
// Compiles to:
%input_bf16 = "ttnn.typecast"(%input_f32) {dtype = bf16}
%values_bf16, %indices = "ttnn.sort"(%input_bf16)
%values_f32 = "ttnn.to_layout"(%values_bf16) {dtype = f32}
```

## Test Coverage

**Unit Tests**:
- `Workarounds/sort_f32_input_workaround.mlir` - Tests workaround pass in isolation (f32)
- `Workarounds/sort_integer_input_workaround.mlir` - Tests workaround pass in isolation (si32)

**Models Now Passing**:
- OPT-125M with top-p sampling
- All LLM generation tests using sampling

lib/Dialect/TTNN/IR/TTNNWorkaroundsPass.cpp
```diff
@@ -821,24 +821,35 @@ TTNNOperandsWorkaroundsFactory::createReduceProdOpOperandsWorkarounds() {
...
+// tt-metal also only supports BFloat16 or UInt16 for input tensors, not
+// Float32. Issue page: https://github.com/tenstorrent/tt-mlir/issues/6926
 TTNNOperandsWorkarounds
 TTNNOperandsWorkaroundsFactory::createSortOpOperandsWorkarounds(
     ttnn::SortOp op) {
-  auto indicesElementType = op.getIndices().getType().getElementType();
+  // Check input tensor type - tt-metal only supports BFloat16 or UInt16
+  auto inputElementType = op.getInput().getType().getElementType();
+  TTNNOperandWorkarounds inputWorkaround;
+  TTNNOperandWorkarounds valuesWorkaround;
```
mmanzoorTT: You can use same variable for both input/output to avoid any mismatch.

lib/Dialect/TTNN/IR/TTNNWorkaroundsPass.cpp
```diff
@@ -821,24 +821,35 @@ TTNNOperandsWorkaroundsFactory::createReduceProdOpOperandsWorkarounds() {
...
+  if (inputElementType.isF32()) {
+    // Convert input from f32 to bf16
+    inputWorkaround.tensorDataTypeWorkaround = ttcore::DataType::BFloat16;
+    // Convert output values to match (must be same type as input after
+    // conversion)
+    valuesWorkaround.tensorDataTypeWorkaround = ttcore::DataType::BFloat16;
+  }
```
mmanzoorTT: if the input is integer and not a `uint16` then workaround should convert it to `uint16`. If the input is floating point and not a `bfloat16` then workaround should add workaround for `bfloat16`.

lib/Dialect/TTNN/IR/TTNNWorkaroundsPass.cpp
kmabeeTT: Done - changed to use `inputOutputWorkaround` for both input and output to ensure they match.

lib/Dialect/TTNN/IR/TTNNWorkaroundsPass.cpp
kmabeeTT: Done - Now handles both of these ...

  - Integer (not ui16) -> converts to ui16
  - Float (not bf16) -> converts to bf16
  
... and added new test sort_integer_input_workaround.mlir

kmabeeTT:
Pushed an extra commit for Asif's PR feedback and offline feedback about redundant tests, updated PR description, going to mark ready for review.  

==

Edit/Side Note - There is a separate existing bug nearby with the exact same sort op when run in release build from tt-xla this assert fires (because workaround pass hardcodes Indices output of sort op to be uint16 but ttnn sometimes returns uint32 depending on shape size)... opened seperate issue here: https://github.com/tenstorrent/tt-mlir/issues/6935 - signature is as follows, just mentioning it here since it's in the same code area:

```
            RuntimeTTNN |    DEBUG | Executing operation: %values, %indices = "ttnn.sort"(%35) <{descending = false, dim = 1 : si8, stable = false}> : (tensor<1x50272xf32, #ttnn.ttnn_layout<(d0, d1) -> (d0, d1), <1x1>, memref<1x1571x!ttcore.tile<32x32, f32>, #ttnn.buffer_type<dram>>, <interleaved>>>) -> (tensor<1x50272xf32, #ttnn.ttnn_layout<(d0, d1) -> (d0, d1), <1x1>, memref<1x1571x!ttcore.tile<32x32, f32>, #ttnn.buffer_type<dram>>, <interleaved>>>, tensor<1x50272xui16, #ttnn.ttnn_layout<(d0, d1) -> (d0, d1), <1x1>, memref<1x1571x!ttcore.tile<32x32, u16>, #ttnn.buffer_type<dram>>, <interleaved>>>) loc("sort.107")
                 Always |    FATAL | DataType mismatch, expected UINT16, got UINT32
DEBUG_ASSERT @ /localdev/kmabee/tt-xla/third_party/tt-mlir/src/tt-mlir/runtime/lib/ttnn/debug/debug_apis.cpp:30: expectedDataType == actualDataType
backtrace:
 --- tt::runtime::ttnn::debug::checkTensorRefMatchesTTNNTensor(tt::target::ttnn::TensorRef const*, tt::tt_metal::Tensor const&)
 --- tt::runtime::ttnn::ProgramTensorPool::insertRuntimeTensorAndValidate(tt::target::ttnn::TensorRef const*, tt::runtime::Tensor const&)
 --- tt::runtime::ttnn::ProgramTensorPool::insertTTNNTensorAndValidate(tt::target::ttnn::TensorRef const*, tt::tt_metal::Tensor const&, bool)
 --- tt::runtime::ttnn::operations::data_movement::run(tt::target::ttnn::SortOp const*, tt::runtime::ttnn::ProgramContext&)
 --- tt::runtime::ttnn::ProgramExecutor::runOperation(tt::target::ttnn::Operation const*)
 --- tt::runtime::ttnn::ProgramExecutor::execute()
 ```

File: lib/Dialect/TTNN/IR/TTNNWorkaroundsPass.cpp [hunk 3]
```diff
@@ -821,24 +821,36 @@ TTNNOperandsWorkaroundsFactory::createReduceProdOpOperandsWorkarounds() {
 // tt-metal generates indices of type UInt16. Any mismatch between generated and
 // expected data type will cause runtime to assert.
 // Issue page: https://github.com/tenstorrent/tt-mlir/issues/4405
+// tt-metal also only supports BFloat16 or UInt16 for input tensors, not
+// Float32. Issue page: https://github.com/tenstorrent/tt-mlir/issues/6926
```
sdjordjevicTT: Since we have added a workaround, please also open a Metal issue and link it here for reference. Here we tend to reference metal issues that will remove the workaround in the future instead of what caused the issue.

test/ttmlir/Dialect/TTNN/Transforms/Workarounds/sort_f32_input_workaround.mlir
```diff
@@ -0,0 +1,24 @@
+// RUN: ttmlir-opt --ttcore-register-device --ttnn-layout --convert-ttir-to-ttnn --ttnn-workaround --canonicalize -o %t %s
```
sdjordjevicTT, line 1: We have a builder sort test here, we should probably add f32, uint32, and uint16 for data type parametrization:
https://github.com/tenstorrent/tt-mlir/blob/9ec32e52aa8386d7c6491360db4140fc705e9996/test/python/golden/ttir_ops/data_movement/test_data_movement.py#L399

Also, can we add a sort no-workaround builder test here:
https://github.com/tenstorrent/tt-mlir/blob/main/test/python/golden/ttir_ops/workarounds/test_workarounds.py

sdjordjevicTT (commented):
Looks good, just a few comments to create a metal issue and to strengthen our builder tests.

lib/Dialect/TTNN/IR/TTNNWorkaroundsPass.cpp
kmabeeTT: Done, opened https://github.com/tenstorrent/tt-metal/issues/37322 and updated link, marking this resolved.

test/ttmlir/Dialect/TTNN/Transforms/Workarounds/sort_f32_input_workaround.mlir
kmabeeTT: Done, added a no-workaround test entry that xfails as expected, and added f32 to existing test_sort in test_data_movement.py as suggested (passes) but when adding torch.int16 and torch.int32 dtypes they fail with low PCC (I wasn't sure how to test input types a uint32 or uint16 since torch doesn't have those types). Maybe something to follow up here on the side and debug later..

kmabeeTT:
Rebased on latest and pushed some feedback, PTAL.

test/ttmlir/Dialect/TTNN/Transforms/Workarounds/sort_f32_input_workaround.mlir
sdjordjevicTT: Yeah, makes sense, this seems enough for now.

mmanzoorTT (approved):

PR #6932, opened 2026-02-06
Support presharded/unsharded graphs in builder

tapspatel:
This PR supports presharded args in builder ecosystem.

Users can now annotate operands with a shard status with `builder.preshard_arg` API.
```
def module3(builder: TTIRBuilder):
    @builder.func([(1, 1, 256, 512)], [torch.float32])
    def model(in0: Operand, builder: TTIRBuilder):
        builder.preshard_arg(in0, shard_dims=(-1, 3))
        in_shard = builder.mesh_shard(
            in0,
            shard_direction=MeshShardDirection.FullToShard.value,
            shard_type=MeshShardType.Identity.value,
            shard_shape=(1, 1, 1, 2),
            shard_dims=(-1, 3),
        )
        exp = builder.exp(in_shard)
        out_shard = builder.mesh_shard(
            exp,
            shard_direction=MeshShardDirection.ShardToFull.value,
            shard_type=MeshShardType.Devices.value,
            shard_shape=(1, 1, 1, 2),
            shard_dims=(-1, 3),
        )
        return out_shard
```

Users can also load presharded mlir graphs and pass in presharded tensors as inputs.

Golden evaluation and runtime is updated to handle such cases.

### DIFF
--- a/include/ttmlir/Dialect/TTCore/Utils/Mesh.h
+++ b/include/ttmlir/Dialect/TTCore/Utils/Mesh.h
@@ -17,7 +17,7 @@ namespace mlir::tt::ttcore::utils {
 // Determine hardware mesh config for DeviceAttr.
 // If none exists, the empty meshShape leads to single device config.
 // If either option.meshShape or meshes exists, use one of them.
-// If both exist, compare mesh and throw error if they are different.
+// If both exist, use mesh shape found in module.
 inline llvm::Expected<llvm::SmallVector<int64_t>>
 determineMeshShape(mlir::ModuleOp module, llvm::ArrayRef<int64_t> meshShape) {
   if (auto meshesAttr = module->getAttrOfType<MeshesAttr>(MeshesAttr::name)) {
@@ -27,12 +27,6 @@ determineMeshShape(mlir::ModuleOp module, llvm::ArrayRef<int64_t> meshShape) {
     }
     // For now, use the first meshShape.
     llvm::ArrayRef<int64_t> meshFromMeshes = meshAttr[0].getShape();
-    // If both meshes exist, they should be identical. Otherwise, throw error.
-    if (!meshShape.empty() && !llvm::equal(meshShape, meshFromMeshes)) {
-      return llvm::createStringError(
-          std::errc::invalid_argument,
-          "Option.meshShape and mesh info from graph should be identical.");
-    }
     return llvm::SmallVector<int64_t>(meshFromMeshes);
   }
   return llvm::SmallVector<int64_t>(meshShape);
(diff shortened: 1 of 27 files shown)


lib/Dialect/TTCore/Transforms/TTCoreRegisterDevice.cpp
```diff
@@ -13,6 +13,8 @@ namespace mlir::tt::ttcore {
 #define GEN_PASS_DEF_TTIRDEPRECATEDLOADSYSTEMDESC
 #include "ttmlir/Dialect/TTCore/Transforms/Passes.h.inc"
 
+#include <iostream>
```
sshonTT: Could you remove this?

hshahTT: Unneeded include

tapspatel: ah the good ol prints. will remove :)

include/ttmlir/Dialect/TTCore/Utils/Mesh.h
```diff
@@ -17,7 +17,7 @@ namespace mlir::tt::ttcore::utils {
 // Determine hardware mesh config for DeviceAttr.
 // If none exists, the empty meshShape leads to single device config.
 // If either option.meshShape or meshes exists, use one of them.
-// If both exist, compare mesh and throw error if they are different.
+// If both exist, use mesh shape found in module.
```
vwellsTT, line 20: I don't have full context, but is this desirable? Why do mismatches occur?

vwellsTT: (I broadly support crashing on mismatch vs silently picking one, unless there's a reason this is expected to happen)

tapspatel: yeah, this function arguably shouldn’t exist. It reflects a suboptimal design choice, but I left it in place because removing the mesh-shape pipeline option would require a large cleanup across the codebase.

On main, if an IR already has a mesh encoded and a user tries to specify a different mesh via this pipeline option, the system will throw a mismatch error. The key issue is that mesh shape is not meant to be a configurable option at this stage — it represents the state of the IR for the platform it is being compiled on at that point in the pipeline. If a user wants to compile a graph with a different mesh, the correct place to do that is in the StableHLO pipeline. Those passes explicitly modify the IR to reflect the new mesh configuration.

Ideally, any user constructing a TTIR graph should explicitly encode the mesh they intend to use; otherwise, it defaults to 1×1. This change will always use the encoded mesh.

Arguably, this might allow silent failures, since a user trying to pass a mesh shape will silently fail if there's already an encoded mesh. Let me think of how to support this without silently failing.

tapspatel: Talking with multi device folks, I'm going to remove the mesh shape entirely: https://github.com/tenstorrent/tt-mlir/issues/6972

PR #6933, opened 2026-02-06
Support sparse_matmul op for MoE

sshonTT:
### Ticket
[Link to Github Issue](https://github.com/tenstorrent/tt-xla/issues/3095)

### Problem description

MoE (Mixture of Experts) models like activate only a subset of experts per token, but the existing tt-mlir pipeline had no way to skip computation for inactive experts. 
All expert blocks were computed via standard matmul, wasting compute resources. 
Beyond the missing op itself, there was no support for sharding sparse_matmul across multiple devices (Expert Parallelism / Tensor Parallelism), which is required for production MoE deployments. 
Additionally, the EraseInverseOps pass had a heuristic issue: commuting reshape through elementwise ops with broadcast-containing consteval operands would create broadcast→reshape chains that materialize larger constants in DRAM, increasing memory pressure.

### What's changed

1. End-to-end sparse_matmul op support for MoE computation:

- Define SparseMatmulOp in both TTIR and TTNN dialects with 3 inputs (a, b, sparsity), attributes (is_input_a_sparse, is_input_b_sparse, nnz), and shape verifiers for 3 sparse modes:
  - is_input_b_sparse=true (gate/up projection): [A,B,M,K] @ [1,E,K,N] → [A,B,1,E,M,N]
  - is_input_a_sparse=true (down projection): [A,E,M,K] @ [1,E,K,N] → [A,E,M,N]
  - Both sparse: [1,E,M,K] @ [1,E,K,N] → [1,E,M,N]

- Add full conversion pipeline: stablehlo.custom_call @tt.sparse_matmul → ttir.sparse_matmul → ttnn.sparse_matmul, parsing mhlo.frontend_attributes for sparse mode and nnz

- Add FlatBuffer schema (SparseMatmulOp table) and serialization

2. Multi-device sharding support (EP + TP):

- Register custom Shardy sharding rule for sparse_matmul with Expert Parallelism (E dimension as kPassThrough factor) and Tensor Parallelism (K as kReduction factor for row-parallel, N as kPassThrough factor for column-parallel)

- Add operand workaround for sparsity tensor to enforce ROW_MAJOR layout at compile time, removing the previous runtime TILE→ROW_MAJOR conversion

- Fix strided slice sharding in UpdateGlobalToLocalShapes

3. Cost model improvement for elementwise commute patterns:
- Replace the boolean check in isCommuteDownwardsFavorable with a netDelta-based cost model: identical TMs score -1, consteval paths score 0, non-consteval paths score +1, and reshape through broadcast-containing consteval paths scores +1 (penalizing DRAM pressure from materialized broadcast→reshape chains)

### Checklist

### DIFF
--- a/include/ttmlir/Dialect/TTNN/IR/TTNNOps.td
+++ b/include/ttmlir/Dialect/TTNN/IR/TTNNOps.td
@@ -1607,6 +1607,12 @@ def TTNN_SparseMatmulOp : TTNN_Op<"sparse_matmul"> {
 
     let results = (outs AnyRankedTensor:$result);
 
+    let extraClassDeclaration = [{
+      wa::TTNNOperandsWorkarounds getOperandsWorkarounds() {
+        return wa::TTNNOperandsWorkaroundsFactory::createSparseMatmulOpOperandsWorkarounds();
+      }
+    }];
+
     let hasVerifier = 1;
 }
 
--- a/include/ttmlir/Dialect/TTNN/IR/TTNNWorkaroundsPass.h
+++ b/include/ttmlir/Dialect/TTNN/IR/TTNNWorkaroundsPass.h
@@ -327,6 +327,10 @@ public:
   static TTNNOperandsWorkarounds
   createPagedScaledDotProductAttentionDecodeOpOperandsWorkarounds(
       Operation *op);
+
+  // Create workarounds for sparse_matmul op operands.
+  // Sparsity tensor must be in ROW_MAJOR layout.
+  static TTNNOperandsWorkarounds createSparseMatmulOpOperandsWorkarounds();
 };
 
 } // namespace mlir::tt::ttnn::wa
--- a/lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
+++ b/lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
@@ -6859,7 +6859,8 @@ public:
             srcOp->getDiscardableAttr("mhlo.frontend_attributes"));
     if (!frontendAttributes) {
       return rewriter.notifyMatchFailure(
-          srcOp, "SparseMatmul op must have mhlo.frontend_attributes attribute.");
+          srcOp,
+          "SparseMatmul op must have mhlo.frontend_attributes attribute.");
     }
 
     // Parse is_input_a_sparse attribute
@@ -6873,8 +6874,9 @@ public:
         isInputASparse = false;
       } else {
         return rewriter.notifyMatchFailure(
-            srcOp, "is_input_a_sparse attribute must be true or false. Received \"" +
-                       isInputASparseStringAttr.getValue() + "\".");
+            srcOp,
+            "is_input_a_sparse attribute must be true or false. Received \"" +
+                isInputASparseStringAttr.getValue() + "\".");
       }
     }
 
@@ -6889,8 +6891,9 @@ public:
         isInputBSparse = false;
       } else {
         return rewriter.notifyMatchFailure(
-            srcOp, "is_input_b_sparse attribute must be true or false. Received \"" +
-                       isInputBSparseStringAttr.getValue() + "\".");
+            srcOp,
+            "is_input_b_sparse attribute must be true or false. Received \"" +
+                isInputBSparseStringAttr.getValue() + "\".");
       }
     }
 
@@ -6904,8 +6907,9 @@ public:
       int64_t nnz;
       if (nnzStringAttr.getValue().getAsInteger(10, nnz)) {
         return rewriter.notifyMatchFailure(
-            srcOp, "nnz attribute string must be convertible to integer. Received \"" +
-                       nnzStringAttr.getValue() + "\".");
+            srcOp,
+            "nnz attribute string must be convertible to integer. Received \"" +
+                nnzStringAttr.getValue() + "\".");
       }
       nnzAttr = rewriter.getI64IntegerAttr(nnz);
     }
@@ -6913,7 +6917,8 @@ public:
     // Get operands
     if (adaptor.getOperands().size() != 3) {
       return rewriter.notifyMatchFailure(
-          srcOp, "SparseMatmul op expects exactly 3 operands (a, b, sparsity).");
+          srcOp,
+          "SparseMatmul op expects exactly 3 operands (a, b, sparsity).");
     }
 
     Value inputA = adaptor.getOperands()[0];
@@ -6924,8 +6929,8 @@ public:
         getTypeConverter()->convertType(srcOp.getResult(0).getType()));
 
     rewriter.replaceOpWithNewOp<ttir::SparseMatmulOp>(
-        srcOp, outputType, inputA, inputB, sparsity,
-        isInputASparseAttr, isInputBSparseAttr, nnzAttr);
+        srcOp, outputType, inputA, inputB, sparsity, isInputASparseAttr,
+        isInputBSparseAttr, nnzAttr);
 
     return success();
   }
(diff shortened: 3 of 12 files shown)


include/ttmlir/Dialect/TTIR/IR/TTIROps.td
```diff
@@ -5728,4 +5728,78 @@ def TTIR_SplitQueryKeyValueAndSplitHeadsOp: TTIR_NamedOp<"split_query_key_value_
 
     let hasVerifier = 1;
 }
+
+//===----------------------------------------------------------------------===//
+// TTIR Sparse Matmul Op
+//===----------------------------------------------------------------------===//
+
+def TTIR_SparseMatmulOp : TTIR_NamedOp<"sparse_matmul"> {
+    let summary = "Sparse matrix multiplication for MoE (Mixture of Experts).";
```
azecevic000, line 5737: Do these constraints come from TTNN? I'm not against adding this, but the name suggests sparse GEMM, which we probably want to support at some point. Something along the line of `sparse_matmul_moe` might make more sense.

sshonTT: I see your point regarding the naming convention. However, I chose sparse_matmul primarily to align with the backend, as the underlying primitive in ttnn is explicitly named ttnn.sparse_matmul.

As you noted, the current constraints originate directly from ttnn. If ttnn lifts these constraints in the future, this op could naturally evolve to serve as the general Sparse GEMM without needing a rename. Given this, I thought maintaining consistency with the backend API might be better.

lib/Dialect/TTIR/Transforms/EraseInverseOps/ElementwiseCommutePatterns.cpp
```diff
@@ -132,23 +132,105 @@ class TTIRCommuteTmsThroughElementwiseRewriter
...
+    //
+    //   After commute (%a1's reshape is pushed below add):
+    //     %a0 = ...                                    : (1,16,1,32,1,128)
+    //     // reshape on %a0 removed — commuted below
+    //     %b0 = ...                                    : (32,128)
+    //     %b1 = reshape %b0                            : (1,32,128)
+    //     %b2 = broadcast %b1                          : (16,32,128)
+    //     %b3 = reshape %b2 (inverse reshape)          : (1,16,1,32,1,128)
+    //     %c0 = add(%a0, %b3)                          : (1,16,1,32,1,128)
+    //     %c1 = reshape %c0 (commuted from %a1)        : (1,16,1,32,128)
```
odjuricicTT: @sshonTT Can you share the exact IR for this case with a bit more ops in the surrounding graph?

I'm interested to take a look at if there is a different approach to the problem that could solve both allowing the reshape to commute and not materializing the broadcasted tensor.

cc: @mvasiljevicTT has the most context around this area

sshonTT: @odjuricicTT , @mvasiljevicTT , Here's shlo graph : https://gist.github.com/sshonTT/3a8bc9c7e32da75d145eb84fd2204a78 

Please concentrate on below lines and see how the reshape ops are commuted.
```
      %68 = "stablehlo.all_reduce"(%67) <{channel_handle = #stablehlo.channel_handle<handle = 1, type = 1>, replica_groups = dense<[[0, 1, 2, 3], [4, 5, 6, 7]]> : tensor<2x4xi64>}> ({
      ^bb0(%arg14: tensor<bf16> loc("custom-call.119"), %arg15: tensor<bf16> loc("custom-call.119")):
        %98 = stablehlo.add %arg14, %arg15 : tensor<bf16> loc(#loc45)
        stablehlo.return %98 : tensor<bf16> loc(#loc45)
      }) : (tensor<1x128x1x16x1x5760xbf16>) -> tensor<1x128x1x16x1x5760xbf16> loc(#loc45)
      %69 = stablehlo.reshape %68 : (tensor<1x128x1x16x1x5760xbf16>) -> tensor<1x128x16x5760xbf16> loc(#loc46)
      %70 = stablehlo.reshape %arg12 : (tensor<16x5760xbf16>) -> tensor<1x16x5760xbf16> loc(#loc47)
      %71 = stablehlo.reshape %70 : (tensor<1x16x5760xbf16>) -> tensor<16x5760xbf16> loc(#loc48)
      %72 = stablehlo.broadcast_in_dim %71, dims = [2, 3] : (tensor<16x5760xbf16>) -> tensor<1x128x16x5760xbf16> loc(#loc49)
      %73 = stablehlo.add %69, %72 : tensor<1x128x16x5760xbf16> loc(#loc50)
```

mvasiljevicTT: Moving reshape up the const eval path was intentionally added, . Ideally it should work like this:
```
- [EIO DOWN] : move reshape to after addOp, add inverse reshape on consteval path
- [EIO UP] : commute reshape and broadcast
- [ImplicitBroadcastFold] remove the broadcast
```
The problem here is that `broadcast commute` is done only in `EIO UP` pattern set, and it gets into a race with moving reshape back up:
```
- [EIO DOWN] : move reshape to after addOp, add inverse reshape on consteval path
- [EIO UP] : move reshape before addOp; too late to commute reshape and broadcast
- [ImplicitBroadcastFold] : nothing to remove
```
So the fix could be to add `broadcast commute pattern` to both up and down pattern sets, like on [this commit.](https://github.com/tenstorrent/tt-mlir/commit/77e4c6f23e1e2632d78717e33c1dcc1f8c4ca1b8)

I can't reproduce the DRAM OOM issue, but based on the ttnn ir it looks like it would solve your issue?

[sparse_matmul_eio_ttnn_mlir.txt](https://github.com/user-attachments/files/25240873/sparse_matmul_eio_ttnn_mlir.txt)

sshonTT: Thanks for the suggestion. However it doesn't solve oom. FYI, here's both stablehlo and ttir graph :  https://gist.github.com/sshonTT/80e353fe4b0f54fdf92b309c4db84ec4 

For repro, you can use https://github.com/tenstorrent/tt-mlir/tree/sshon/temp-issue-case which is basically revert my change in EIO.

odjuricicTT: @sshonTT Thanks for the info.

The use of 6d tensors seems to be causing multiple perf issues here. One of which is that some of these reshapes are very slow, another is that we are not able to have an implicit broadcast on add. As these seem to originate from the sparse matmul, we might have a different solution if there is any flexibility in the op (if we could represent these tensors in less dims).

As i don't have enough context around this, i've scheduled a quick sync tomorrow discuss this further. With the general idea of talking about the moe -> sparse_matmul lowering pipeline e2e.

cc: @mvasiljevicTT

sshonTT: As you suggested, I dug deeper into the modeling side instead of the EraseInverse op logic. It turned out the shape (..., 1, intermediate_mlp) in the last two dimensions was causing significant inefficiency in the tile memory layout.

I found that I could reshape the input by borrowing 32 from the batch dimension to make it tile-aligned:
```
        hidden_4d = dispatched.view(BD // 32, seq_len, 32, hidden_size) # which originally (BD, seq_len, 1, hidden_size)
        gate_up_out = torch.ops.tt.sparse_matmul(
            hidden_4d,
            gate_up_proj, # [E, hidden, intermediate *2]
            sparsity,
            nnz=0,
            is_input_a_sparse=False,
            is_input_b_sparse=True,
        )
        # [BD // 32, S, 1, E, 32, inter*2] which originally [BD, S, 1, E, 1, inter*2]
```
With this change, the DRAM pressure seems to be completely gone. I'll revert my changes to erase_inverse_op after testing.

Regarding our sync up, we can probably cancel it since the issue is resolved. However, I'm happy to keep it if you're interested in the details of how the MLP is lowered

odjuricicTT: As discussed offline, removing all EIO changes should work after the above batch dim is transfered to the second to last dim. With this in mind, i'll unblock the PR.

odjuricicTT (approved):
Approving for EIO part of the PR.

sshonTT:
Close as #7300 covers this.

Issue #6935, opened 2026-02-06
ttnn.sort workaround pass doesn't account for UINT32 indices on tensors > 32,768 elements


kmabeeTT:
### Describe the bug

tt-mlir's TTNNWorkaroundsPass hardcodes UINT16 for sort operation indices, but tt-metal dynamically chooses between UINT16 and UINT32 based on the padded tensor size.  See [doc](https://docs.tenstorrent.com/tt-metal/latest/ttnn/ttnn/api/ttnn.sort.html#ttnn.sort) that mentions both formats. When the input tensor's last dimension exceeds 32,768 elements, bitonic sort pads it to 65,536 (next power of 2), causing tt-metal to return UINT32 indices. This triggers a debug assertion  `DataType mismatch, expected UINT16, got UINT32` here:

```
            RuntimeTTNN |     INFO | Executing operation: %values, %indices = "ttnn.sort"(%37) <{descending = false, dim = 1 : si8, stable = false}> : (tensor<1x50272xbf16, #ttnn.ttnn_layout<(d0, d1) -> (d0, d1), <1x1>, memref<1x1571x!ttcore.tile<32x32, bf16>, #ttnn.buffer_type<dram>>, <interleaved>>>) -> (tensor<1x50272xbf16, #ttnn.ttnn_layout
<(d0, d1) -> (d0, d1), <1x1>, memref<1x1571x!ttcore.tile<32x32, bf16>, #ttnn.buffer_type<dram>>, <interleaved>>>, tensor<1x50272xui16, #ttnn.ttnn_layout<(d0, d1) -> (d0, d1), <1x1>, memref<1x1571x!ttcore.tile<32x32, u16>, #ttnn.buffer_type<dram>>, <interleaved>>>) loc("opt125m_graph_17_ttnn.mlir":182:29)
                 Always |    FATAL | DataType mismatch, expected UINT16, got UINT32
DEBUG_ASSERT @ /localdev/kmabee/tt-mlir/runtime/lib/ttnn/debug/debug_apis.cpp:30: expectedDataType == actualDataType
```

  Padding threshold:
  - Last dimension ≤ 32,768 → pads to ≤ 32,768 → tt-metal uses UINT16 ✓
  - Last dimension > 32,768 → pads to 65,536 → tt-metal uses UINT32 ✗ (tt-mlir expects UINT16)

  Affected operations:
  - torch.topk with large tensors (e.g., (1, 40000), (1, 50000))
[...]

PR #6936, opened 2026-02-06
[TTNN] Fix sort indices type selection for large tensors (UINT16, UINT32)

kmabeeTT:
### Ticket
https://github.com/tenstorrent/tt-mlir/issues/6935

### Problem description
The `ttnn.sort` operation fails in debug builds with `DataType mismatch, expected UINT16, got UINT32` when sorting tensors with last dimension > 32,768 (e.g., topk with 40,000 elements, OPT-125M with vocabulary size 50,272).

The root cause: TTNNWorkaroundsPass hardcodes `UINT16` for sort indices, but tt-metal dynamically chooses between `UINT16` and `UINT32` based on the **padded tensor size**. Bitonic sort requires power-of-2 padding, so tensors with last dimension > 32,768 pad to 65,536, triggering tt-metal to use `UINT32` indices. The mismatch causes debug assertions to fail.

### What's changed
Made TTNNWorkaroundsPass padding-aware by replicating tt-metal's type selection logic from `sort_device_operation.cpp`.

**Key Changes**:
- Added `nextPowerOfTwo()` helper to calculate padded tensor size
- Updated `createSortOpOperandsWorkarounds()` to choose `UINT16` or `UINT32` based on padded size (≥65,535 threshold)
- Updated comments to reflect dynamic type selection
- Added `test/ttmlir/Dialect/TTNN/Transforms/Workarounds/sort_vocab_size_uint32.mlir` regression test verifying UINT32 indices selection for large tensors (50,272 elements from OPT-125M vocabulary) that pad beyond the UINT16 threshold

**Impact**: Enables sort operations on large tensors (vocabulary sizes, large topk) in debug builds. Fixes latent correctness issue in release builds.

### Checklist

### DIFF
--- a/lib/Dialect/TTNN/IR/TTNNWorkaroundsPass.cpp
+++ b/lib/Dialect/TTNN/IR/TTNNWorkaroundsPass.cpp
@@ -858,14 +858,16 @@ TTNNOperandsWorkaroundsFactory::createSortOpOperandsWorkarounds(
   int64_t paddedLastDim = llvm::PowerOf2Ceil(lastDim);
   constexpr int64_t uint16Max = std::numeric_limits<uint16_t>::max();
 
-  // Strengthen logic to handle edge cases where indices type might be
-  // UInt16 but padded shape exceeds valid range
-  if (!isUInt16 && paddedLastDim < uint16Max) {
-    // Convert to UInt16 if padded size fits in uint16 range
-    indicesWorkaround.tensorDataTypeWorkaround = ttcore::DataType::UInt16;
-  } else if (!isUInt32) {
-    // Convert to UInt32 for larger sizes or if already UInt16 but out of range
-    indicesWorkaround.tensorDataTypeWorkaround = ttcore::DataType::UInt32;
+  // Determine the correct indices type based on padded size, then only
+  // apply a workaround if the current type doesn't already match.
+  if (paddedLastDim < uint16Max) {
+    if (!isUInt16) {
+      indicesWorkaround.tensorDataTypeWorkaround = ttcore::DataType::UInt16;
+    }
+  } else {
+    if (!isUInt32) {
+      indicesWorkaround.tensorDataTypeWorkaround = ttcore::DataType::UInt32;
+    }
   }
 
   return TTNNOperandsWorkarounds::createEmptyTTNNOperandsWorkarounds()


lib/Dialect/TTNN/IR/TTNNWorkaroundsPass.cpp
```diff
@@ -817,19 +817,50 @@ TTNNOperandsWorkaroundsFactory::createReduceProdOpOperandsWorkarounds() {
       .addOutputOperandWorkaround(bf16Workaround);
 }
 
+// Helper function to calculate next power of two (for bitonic sort padding)
+static int64_t nextPowerOfTwo(int64_t n) {
+  if (n <= 1)
```
vmilosevic: :warning: **google\-readability\-braces\-around\-statements** :warning:
```suggestion
  if (n <= 1) {
```

lib/Dialect/TTNN/IR/TTNNWorkaroundsPass.cpp
```diff
@@ -817,19 +817,50 @@ TTNNOperandsWorkaroundsFactory::createReduceProdOpOperandsWorkarounds() {
...
+    return 1;
+  // If n is already a power of two, return it
```
vmilosevic: :warning: **google\-readability\-braces\-around\-statements** :warning:
```suggestion
}
  // If n is already a power of two, return it
```

lib/Dialect/TTNN/IR/TTNNWorkaroundsPass.cpp
```diff
@@ -817,19 +817,50 @@ TTNNOperandsWorkaroundsFactory::createReduceProdOpOperandsWorkarounds() {
...
+  if ((n & (n - 1)) == 0)
```
vmilosevic: :warning: **google\-readability\-braces\-around\-statements** :warning:
```suggestion
  if ((n & (n - 1)) == 0) {
```

lib/Dialect/TTNN/IR/TTNNWorkaroundsPass.cpp
```diff
@@ -817,19 +817,50 @@ TTNNOperandsWorkaroundsFactory::createReduceProdOpOperandsWorkarounds() {
...
+    return n;
+  // Otherwise, compute the next power of two
```
vmilosevic: :warning: **google\-readability\-braces\-around\-statements** :warning:
```suggestion
}
  // Otherwise, compute the next power of two
```

vmilosevic (commented):

kmabeeTT:
FYI @mmanzoorTT / @djstefanovicTT  and others, Claude helped me debug this and whip up a fix quickly for this blocker issue preventing us from enabling vllm SamplingParams other than "greedy", let me know what you think or if you have another idea.

==

New unit test passing with fix, and original tests from tt-xla also passing. Here is diff before vs after fix for a case that was triggering this debug assert.

```
diff topk_graph_1_ttnn.mlir.fail topk_graph_1_ttnn.mlir.pass
7c7
< #ttnn_layout4 = #ttnn.ttnn_layout<(d0, d1) -> (d0, d1), <1x1>, memref<1x1563x!ttcore.tile<32x32, u16>, #dram>, <interleaved>>
---
> #ttnn_layout4 = #ttnn.ttnn_layout<(d0, d1) -> (d0, d1), <1x1>, memref<1x1563x!ttcore.tile<32x32, u32>, #dram>, <interleaved>>
18c18
<         %values, %indices = "ttnn.sort"(%1) <{descending = true, dim = 1 : si8, stable = false}> : (tensor<1x50000xbf16, #ttnn_layout3>) -> (tensor<1x50000xbf16, #ttnn_layout3>, tensor<1x50000xui16, #ttnn_layout4>)
---
>         %values, %indices = "ttnn.sort"(%1) <{descending = true, dim = 1 : si8, stable = false}> : (tensor<1x50000xbf16, #ttnn_layout3>) -> (tensor<1x50000xbf16, #ttnn_layout3>, tensor<1x50000xui32, #ttnn_layout4>)
21,22c21,22
<         %2 = "ttnn.typecast"(%indices) <{dtype = #ttcore.supportedDataTypes<si32>}> : (tensor<1x50000xui16, #ttnn_layout4>) -> tensor<1x50000xsi32, #ttnn_layout5>
<         "ttnn.deallocate"(%indices) <{force = false}> : (tensor<1x50000xui16, #ttnn_layout4>) -> ()
---
>         %2 = "ttnn.typecast"(%indices) <{dtype = #ttcore.supportedDataTypes<si32>}> : (tensor<1x50000xui32, #ttnn_layout4>) -> tensor<1x50000xsi32, #ttnn_layout5>
>         "ttnn.deallocate"(%indices) <{force = false}> : (tensor<1x50000xui32, #ttnn_layout4>) -> ()
```

lib/Dialect/TTNN/IR/TTNNWorkaroundsPass.cpp
```diff
@@ -817,19 +817,54 @@ TTNNOperandsWorkaroundsFactory::createReduceProdOpOperandsWorkarounds() {
       .addOutputOperandWorkaround(bf16Workaround);
 }
 
+// Helper function to calculate next power of two (for bitonic sort padding)
+static int64_t nextPowerOfTwo(int64_t n) {
```
azecevic000: There is the `llvm::PowerOf2Ceil` which implements this functionality.

lib/Dialect/TTNN/IR/TTNNWorkaroundsPass.cpp
```diff
@@ -817,19 +817,54 @@ TTNNOperandsWorkaroundsFactory::createReduceProdOpOperandsWorkarounds() {
...
+
+  // Check output indices type - tt-metal generates UInt16 or UInt32 based on
+  // padded size Bitonic sort pads to next power of 2, and uses UInt32 if padded
+  // size >= 65535
   auto indicesElementType = op.getIndices().getType().getElementType();
 
   TTNNOperandWorkarounds datatypeWorkaround;
   if (!(indicesElementType.isInteger(16) &&
+        indicesElementType.isUnsignedInteger()) &&
+      !(indicesElementType.isInteger(32) &&
```
azecevic000: This seems like a good place to strengthen this condition a bit further. What happens if the type is `u16`, but the padded shape is larger than the range? Something like
```cpp
// last dim logic
if (!u16 && lastDim < range) {
  // to u16
} else if (!u32) {
  // to u32
}
```

lib/Dialect/TTNN/IR/TTNNWorkaroundsPass.cpp
```diff
@@ -817,19 +817,54 @@ TTNNOperandsWorkaroundsFactory::createReduceProdOpOperandsWorkarounds() {
...
-    datatypeWorkaround.tensorDataTypeWorkaround = ttcore::DataType::UInt16;
+    // Calculate padded size to determine UInt16 vs UInt32
+    auto inputType = mlir::cast<RankedTensorType>(op.getInput().getType());
+    auto inputShape = inputType.getShape();
+    int64_t lastDim = inputShape[inputShape.size() - 1];
+    int64_t paddedLastDim = nextPowerOfTwo(lastDim);
+
+    // Match tt-metal's logic: use UInt32 if padded size >= uint16_t::max
+    // (65535)
+    if (paddedLastDim >= 65535) {
```
azecevic000: Seems like a tt-metal logic is off by one, so we should probably go with that, unfortunately (still it's a good idea to mention that in the comment, because it's not obvious). But please use either `0xFFFF` or `std::numeric_limits<uint16_t>::max()`.

lib/Dialect/TTNN/IR/TTNNWorkaroundsPass.cpp
kmabeeTT: good call, done - updated.

lib/Dialect/TTNN/IR/TTNNWorkaroundsPass.cpp
kmabeeTT: Done, using std::numeric_limits<uint16_t>::max() and updated comment slightly.

lib/Dialect/TTNN/IR/TTNNWorkaroundsPass.cpp
kmabeeTT: Ah thx good call, done updated.

kmabeeTT:
Rebased on latest, added feedback items and force pushed, please take a look.

kmabeeTT:
Update: Rebased and squashed feedback today, and ran against tt-xla nightly set of passing test_models.py models ([passing](https://github.com/tenstorrent/tt-xla/actions/runs/21843083088)), but and hit [hangs](https://github.com/tenstorrent/tt-mlir/actions/runs/21841836008/job/63039869287#step:15:2060) here in onPR in CI build step in what turned out to be a sort test `test/ttmlir/EmitC/TTNN/sort/sort.mlir` , then traced to deadlock in the recently added "strengthened conditon" from PR feedback.  Just pushed fix, details below:

--

> The emitc.sh CI step was hanging during llvm-lit on test/ttmlir/EmitC/TTNN/sort/sort.mlir. The ttmlir-opt process was stuck at 100% CPU in the TTNNWorkarounds pass.
> 
> Root cause: the indices type workaround was oscillating infinitely in the greedy pattern rewriter. For small tensors (where UInt16 is the correct indices type), the if/else-if structure didn't handle the "already correct" case:
> 
> ```
>   // BUG: when isUInt16=true and paddedLastDim < uint16Max:
>   //   first branch: !isUInt16(false) && ... → skipped
>   //   else-if: !isUInt32(true) → wrongly sets UInt32
>   // Next iteration flips it back to UInt16, ad infinitum
>   if (!isUInt16 && paddedLastDim < uint16Max) {
>       indicesWorkaround = UInt16;
>   } else if (!isUInt32) {
>       indicesWorkaround = UInt32;
>   }
> ```
> 
>   Fix: determine the target type first, then only apply a workaround if the current type doesn't match:
> 
> ```
>   if (paddedLastDim < uint16Max) {
>       if (!isUInt16) indicesWorkaround = UInt16;
>   } else {
>       if (!isUInt32) indicesWorkaround = UInt32;
>   }
> ```

PR #6941, opened 2026-02-06
Fix ttrt run/perf to return non-zero exit code when no binaries found

odjuricicTT:
### Ticket
N/A

### Problem description
When `ttrt run` or `ttrt perf` is called with a path that contains no matching
binaries (non-existent path, wrong file extension, empty directory), the tool
exits with code 0 (success).

Commands like:
```
ttrt run gsaenfaelsnfn

ttrt run llana.ttnn
```
Execute nothing but return success which does not make a good user experinece.

It also makes it impossible for CI/CD pipelines or
scripts to detect that nothing was actually executed.

### What's changed
Added a check at the end of `postprocess()` in both `Run` and `Perf` commands.
If after all processing the results list is still empty (no passes, no errors,
no skips), an error result is added with the message
`"no binaries found or executed at path: <path>"`, causing the tool to return
exit code 1.

**Verified scenarios:**
| Scenario | Before | After |
|---|---|---|
| Non-existent path | exit 0 | exit 1 |
| Wrong file extension | exit 0 | exit 1 |
| Empty dir (no binaries) | exit 0 | exit 1 |
| All binaries skipped | exit 1 | exit 1 (unchanged) |
| Binary execution fails | exit 1 | exit 1 (unchanged) |
| Valid binary succeeds | exit 0 | exit 0 (unchanged) |

### Checklist

### DIFF
--- a/.github/test_scripts/ttrt.sh
+++ b/.github/test_scripts/ttrt.sh
@@ -10,6 +10,10 @@
 set -e -o pipefail
 
 echo "Running TTRT tests"
-eval ttrt "$1" "$BUILD_DIR/test/ttmlir/$2" "$3"
+# Resolve the test path, expanding $RUNS_ON then mapping runner names back
+# to test directory names (e.g. n300-llmbox runner uses the llmbox test directory)
+TTRT_TEST_PATH=$(eval echo "$2")
+TTRT_TEST_PATH="${TTRT_TEST_PATH//n300-llmbox/llmbox}"
+ttrt "$1" "$BUILD_DIR/test/ttmlir/$TTRT_TEST_PATH" $3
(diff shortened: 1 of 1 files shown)


jgrimTT (approved):

odjuricicTT:
Also fixed the `n300-llmbox` CI runner test path in `ttrt.sh`. The runner name `n300-llmbox` was being substituted into the test path via `$RUNS_ON`, but the actual test directory is `llmbox` (not `n300-llmbox`). This was silently passing before because ttrt returned exit 0 when no binaries were found.

nsumrakTT (approved):
This problem persists for quite a while and finally you tackled it.
Thanks a lot!

PR #6953, opened 2026-02-06
D2M Arange Op

anuragsingh-tt:
### Problem description
Enabling support for the torch::arange operation in the D2M flow.

### What's changed
Added a new ArangeBlock op in the D2M dialect that is decomposed into low-level tile operations via the DecomposeArange pass. 
Added a new experimental LLK ExperimentalWriteFullIndexTileOp to write a scratch index tensor that is added per iteration of the arange loop. 
Added Arange tests in test_metal_tms.py

### DIFF
--- a/include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
+++ b/include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
@@ -2035,7 +2035,7 @@ def D2M_FillArangeTileOp : D2M_GenericRegionOp<"fill_arange_tile",
   [DeclareOpInterfaceMethods<MemoryEffectsOpInterface>]> {
   let summary = "Write full linear index pattern to a tile CB";
   let description = [{
-      The `experimental::write_full_linear_index_tile` operation writes a linear index pattern
+      The `fill_arange_tile` operation writes a linear index pattern
       to a tile CB, where element[i,j] = i * 32 + j (the linear index as float).
 
       This is used for arange operations. Each element gets its linear index value


include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
```diff
@@ -1418,7 +1418,7 @@ def D2M_RemoteStoreOp : D2M_GenericRegionDatamovementOp<"remote_store",
         llvm_unreachable("getMemRefType() called on tensor operand, use getShapedType() instead");
       }
     }];
-  }
```
vwellsTT, line 1421: nit: this seems correct, but does the rest of the stuff above it need an indent too? looks weird to have it be 2 tabs in I think?

include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
```diff
@@ -2027,4 +2027,89 @@ def D2M_WriteColMaskTileOp : D2M_GenericRegionOp<"write_col_mask_tile",
     let assemblyFormat = [{ `(` $validCols `)` `to` $output attr-dict `:` type($validCols) `,` type($output) }];
 }
 
+//===----------------------------------------------------------------------===//
+// D2M Arange Ops
+//===----------------------------------------------------------------------===//
+
+def D2M_ExperimentalWriteFullIndexTileOp : D2M_GenericRegionOp<"experimental::write_full_index_tile",
```
vwellsTT: I had same pattern, but in my PR Nick said we shouldn't use Experimental in D2M, only TTKernel (which makes sense I think).
Further, I think this should have a more clear name, "full" does not imply stepped 0-1023 to me; I guess "index" is supposed to convey that, but I wonder if there's a better name, doesn't seem as obvious to me as I'd like what this op means

include/ttmlir/Target/TTKernel/LLKs/experimental_padding_llks.h
```diff
@@ -125,6 +125,49 @@ ALWI void write_col_mask_tile(uint32_t validCols, uint32_t cb_id) {
 #endif
 }
 
+// =============================================================================
+// Index tile functions for arange operations
+// =============================================================================
+
+// Helper to convert uint32_t to F32 bit pattern
+inline uint32_t _uint_to_f32_bits_(uint32_t val) {
```
vwellsTT: I have created a "floats_to_bits" in TTKernelToCPP.cpp, I think this is completely equivalent?

lib/Dialect/D2M/Transforms/DecomposeArange.cpp
```diff
@@ -0,0 +1,193 @@
+// SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
```
vwellsTT: nit: 2026

lib/Dialect/D2M/Transforms/DecomposeArange.cpp
```diff
@@ -0,0 +1,193 @@
...
+
+    auto genericOp = op->getParentOfType<GenericOp>();
+    if (!genericOp) {
+      return rewriter.notifyMatchFailure(
+          op, "ArangeBlockOp must be inside a GenericOp");
+    }
+
+    ttcore::GridAttr gridAttr = genericOp.getGrid();
+    ArrayRef<int64_t> gridShape = gridAttr.getShape();
+    if (gridShape.size() < 2) {
```
vwellsTT: I would think this could be an assert instead

lib/Dialect/D2M/Transforms/DecomposeArange.cpp
```diff
@@ -0,0 +1,193 @@
...
+        rewriter.create<arith::ConstantIndexOp>(loc, numTileCols);
+
+    // === STEP 1: Write the scratch tile ===
+    TT_assert(indexTileMemref);
+    rewriter.create<ExperimentalWriteFullIndexTileOp>(loc, indexTileMemref);
+
+    // === STEP 2: Scalar constants for arange start and step ===
+    Value startF = rewriter.create<arith::ConstantOp>(
+        loc, elemType,
+        rewriter.getFloatAttr(elemType, static_cast<double>(start)));
```
vwellsTT, line 71: does floatAttr only take doubles? weird

include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
nsmithtt: +1, there's no such thing as experimental in d2m, that's a ttkernel thing that refers to a compiler implementation of something that should live in llk.

lib/Dialect/D2M/Transforms/GridSelection.cpp
```diff
@@ -947,20 +947,23 @@ recreateGenericOp(d2m::GenericOp genericOp,
...
-                  auto outputOperandType =
-                      clonedOp->getOperand(numIns + i).getType();
-                  clonedOp->getResult(i).setType(outputOperandType);
-                }
+            } else if (auto dstOp = llvm::dyn_cast<DestinationStyleOpInterface>(
+                           clonedOp)) {
+              int numIns, numOuts = 0;
+              if (auto segAttr =
+                      clonedOp->getAttrOfType<mlir::DenseI32ArrayAttr>(
+                          "operandSegmentSizes")) {
```
nsmithtt: There's no need to check for this attribute.  We've already proved DPS we can unconditionally use:
```
                numIns = dstOp.getNumDpsInputs();
                numOuts = clonedOp->getNumResults();
```

include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
anuragsingh-tt: Looks like the correct convention has gone awry in this file. I fixed the indents for the ops I added. Not sure if its worth changing the indents for others as it will muck up the line blame history.

include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
anuragsingh-tt: How about `WriteFullLinearIndexTileOp`?

lib/Dialect/D2M/Transforms/GridSelection.cpp
anuragsingh-tt: Yes, I was wondering why we did it with operandSegmentSizes before.

lib/Dialect/D2M/Transforms/DecomposeArange.cpp
anuragsingh-tt: When should we use `TT_assertv` vs `rewriter.notifyMatchFailure()`? 
From `AGENTS.md`:
> Inside patterns: Use `rewriter.notifyMatchFailure()` for pattern match failures

I do not see a consistent rule in the codebase.

vwellsTT: I think assert is appropriate for invariants regardless of scope--this seems to be an invariant imo

include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
```diff
@@ -2027,4 +2027,89 @@ def D2M_WriteColMaskTileOp : D2M_GenericRegionOp<"write_col_mask_tile",
     let assemblyFormat = [{ `(` $validCols `)` `to` $output attr-dict `:` type($validCols) `,` type($output) }];
 }
 
+//===----------------------------------------------------------------------===//
+// D2M Arange Ops
+//===----------------------------------------------------------------------===//
+
+def D2M_WriteFullLinearIndexTileOp : D2M_GenericRegionOp<"write_full_linear_index_tile",
```
nsmithtt: Maybe we can call this `fill_arange_tile`

include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
```diff
@@ -2027,4 +2027,89 @@ def D2M_WriteColMaskTileOp : D2M_GenericRegionOp<"write_col_mask_tile",
...
 
+//===----------------------------------------------------------------------===//
+// D2M Arange Ops
+//===----------------------------------------------------------------------===//
+
+def D2M_FillArangeTileOp : D2M_GenericRegionOp<"fill_arange_tile",
+  [DeclareOpInterfaceMethods<MemoryEffectsOpInterface>]> {
+  let summary = "Write full linear index pattern to a tile CB";
+  let description = [{
+      The `experimental::write_full_linear_index_tile` operation writes a linear index pattern
```
nsmithtt: Stale name in doc string

PR #6958, opened 2026-02-09
Enable eth connections in system desc and add mesh axis topology

pglusacTT:
### Ticket
https://github.com/tenstorrent/tt-xla/issues/3067

### Problem description
Multi-device setups need per-axis topology information (ring vs linear) to be propagated through the compiler so that CCL operations can select the correct communication pattern, as well as the global Fabric config. Additionally, the system descriptor's ethernet connection discovery was disabled via workarounds that suppressed all connection enumeration, which works fine in tests.

### What's changed

  - Mesh topology in DeviceAttr: Added an optional meshTopology parameter to DeviceAttr that stores a per-axis Topology (Ring or Linear). This is threaded through the register-device pass and TTIR-to-TTNN pipeline options.
  - Ethernet connection fix: Replaced architecture-specific guards in system_desc.cpp with a try-catch around get_connected_ethernet_core. Cores that connect to remote devices (outside the local cluster) are now skipped with a warning instead of silently suppressing all connections. TT-Metal doesn't provide an API that can handle this without a try/catch, added an issue.
  - getMappedDeviceIds runtime API: Added tt::runtime::getMappedDeviceIds() which queries SystemMesh to get logical-to-physical device ID mapping for a given mesh shape without requiring an open MeshDevice.
  - New MeshTopologyAttr: Added a standalone MeshTopologyAttr in the TTCore dialect for describing per-axis topologies.
  - Multihost test fix: Skipped chip_channels comparison in multihost system descriptor tests since channel assignments can differ across hosts.

### Checklist

### DIFF
--- a/lib/Dialect/TTCore/IR/TTCoreOpsTypes.cpp
+++ b/lib/Dialect/TTCore/IR/TTCoreOpsTypes.cpp
@@ -1798,10 +1798,8 @@ size_t DeviceAttr::getMemrefCBNumPages(MemRefType memrefType) const {
     return ::mlir::failure();
   }
 
-  if (!meshTopology.empty() &&
-      (meshShape.empty() || meshTopology.size() != meshShape.size())) {
-    emitError() << "expected meshTopology size to match meshShape rank";
-    return ::mlir::failure();
+  if (!meshTopology.empty() && meshTopology.size() != meshShape.size()) {
+    return emitError() << "expected meshTopology size to match meshShape rank";
   }
 
   return ::mlir::success();


lib/Dialect/TTCore/IR/TTCoreOpsTypes.cpp
```diff
@@ -1794,6 +1798,12 @@ ::mlir::LogicalResult DeviceAttr::verify(
     return ::mlir::failure();
   }
 
+  if (!meshTopology.empty() &&
+      (meshShape.empty() || meshTopology.size() != meshShape.size())) {
```
azecevic000: Nit: Is `meshShape.empty()` redundant in the condition? If `meshShape.empty()` and `!meshTopology.empty()` are true, then `meshTopology.size() != meshShape.size()` will also be true.

lib/Dialect/TTCore/IR/TTCoreOpsTypes.cpp
```diff
@@ -1794,6 +1798,12 @@ ::mlir::LogicalResult DeviceAttr::verify(
...
+    emitError() << "expected meshTopology size to match meshShape rank";
+    return ::mlir::failure();
```
azecevic000: Nit:
```suggestion
    return emitError() << "expected meshTopology size to match meshShape rank";
```

sohaibnadeemTT:
@pglusacTT how does this new mesh topology get determined?

pglusacTT:
> @pglusacTT how does this new mesh topology get determined?

The idea is analogous to mesh shape, it is set as a compilation parameter (in `createTTIRToTTNNBackendPipeline`).

The entire flow is as follows:
- The system descriptor is populated with ethernet connections (with physical IDs)
- tt-xla reads the system descriptor and opens the mesh
- For *that* mesh it matches the logical <-> physical coordinates/chip ids via `getMappedDeviceIds` added in this PR
- For each axis of the mesh, it checks the topolgy (ring, linear)
- During compilation, it passes both the mesh shape and the axis topologies

@sohaibnadeemTT I hope that makes sense!

nsmithtt:
> > @pglusacTT how does this new mesh topology get determined?
> 
> The idea is analogous to mesh shape, it is set as a compilation parameter (in `createTTIRToTTNNBackendPipeline`).
> 
> The entire flow is as follows:
> 
> * The system descriptor is populated with ethernet connections (with physical IDs)
> * tt-xla reads the system descriptor and opens the mesh
> * For _that_ mesh it matches the logical <-> physical coordinates/chip ids via `getMappedDeviceIds` added in this PR
> * For each axis of the mesh, it checks the topolgy (ring, linear)
> * During compilation, it passes both the mesh shape and the axis topologies
> 
> @sohaibnadeemTT I hope that makes sense!

I guess to @sohaibnadeemTT's point, isn't the topology something that tt-mlir device constructor could derive automatically?  It might be better to have this logic in tt-mlir instead of tt-xla.

Alternatively, perhaps we should at least assert in tt-mlir that the chosen topology is legal.

pglusacTT:
> > > @pglusacTT how does this new mesh topology get determined?
> > 
> > 
> > The idea is analogous to mesh shape, it is set as a compilation parameter (in `createTTIRToTTNNBackendPipeline`).
> > The entire flow is as follows:
> > 
> > * The system descriptor is populated with ethernet connections (with physical IDs)
> > * tt-xla reads the system descriptor and opens the mesh
> > * For _that_ mesh it matches the logical <-> physical coordinates/chip ids via `getMappedDeviceIds` added in this PR
> > * For each axis of the mesh, it checks the topolgy (ring, linear)
> > * During compilation, it passes both the mesh shape and the axis topologies
> > 
> > @sohaibnadeemTT I hope that makes sense!
> 
> I guess to @sohaibnadeemTT's point, isn't the topology something that tt-mlir device constructor could derive automatically? It might be better to have this logic in tt-mlir instead of tt-xla.
> 
> Alternatively, perhaps we should at least assert in tt-mlir that the chosen topology is legal.

The Fabric config must be set in tt-xla before opening the device, so this logic must happen before entering tt-mlir compilation. Having it here too would duplicate that logic. What do you think @nsmithtt ?

nsmithtt:
> The Fabric config must be set in tt-xla before opening the device, so this logic must happen before entering tt-mlir compilation. Having it here too would duplicate that logic. What do you think @nsmithtt ?

I see, ok that makes sense.  It could make sense to expose an API from tt-mlir that can return a topology given a system desc + mesh shape.  My only thinking is that it could be nice to have an error check during tt-mlir device registration that ensures the topology is legal.  It could be easy for a unit test or third party to program things incorrectly which would lead to a runtime error or hang.

pglusacTT:
> > The Fabric config must be set in tt-xla before opening the device, so this logic must happen before entering tt-mlir compilation. Having it here too would duplicate that logic. What do you think @nsmithtt ?
> 
> I see, ok that makes sense. It could make sense to expose an API from tt-mlir that can return a topology given a system desc + mesh shape. My only thinking is that it could be nice to have an error check during tt-mlir device registration that ensures the topology is legal. It could be easy for a unit test or third party to program things incorrectly which would lead to a runtime error or hang.

I agree that it is better to have some checks. I'll add the runtime API for mesh + system desc topology query and I'll have it return both the recommended fabric config (global) and the per-axis settings. Would that be okay? @nsmithtt

nsmithtt:
Ok, yeah I think having a ttmlir runtime API can work, that sounds good.

jameszianxuTT:
Interestingly on a true multihost setup with 2 bh QB machines (not in CI yet), we see this result captured by the try/catch
image

So, each p150 card seems to have two chips at (x=0, y=11) and (x=0, y=9) that interface with the QFSP ports on the back, and this would make sense since I there is a single, QFSP-DD connection from each card on one machine, to another card on the remote machine. 

Not to block the PR, this is just an interesting detail. I will look a bit more to confirm this is how the PCIe board's QFSP link interfaces with the eth interfaces on the chip. If anything, this could be another path to query cluster connectivity :p

PR #6960, opened 2026-02-09
Enable mock device support in the compiler optimization path

### LINKED ISSUE #6719: Use virtual device for all constraints api call

rpavlovicTT:
Previously, SingletonDeviceContext always opened a real device connected to TT hardware. This meant the OpModel-based optimizer passes (sharding, layout propagation, fusing, conv2d weight preparation) could only run on machines with physical accelerators.

This change introduces mock device mode alongside the existing real device mode:
- openDevice() opens a real device connected to hardware (unchanged)
- openMockDevice() configures a mock device using architecture info from the MLIR system descriptor, enabling getOpConstraints queries without hardware
- getOpRuntime is guarded to return an error in mock mode since it requires JIT compilation and trace capture on a real device
- ScopedSingletonDeviceGuard and all optimizer passes now use openMockDevice() since they only need constraint queries

This enables running the full compiler optimization pipeline (including OpModel constraint validation) on CI machines without TT hardware.

Fixes #6719 

Note: unit tests TestOpModelLib and TestOpModelInterface still work with real device -- opened issue https://github.com/tenstorrent/tt-mlir/issues/6961 to refactor them to work with mock device (for `getOpConstraints` case).

### DIFF
--- a/test/unittests/Validation/TestOpConstraintValidation.cpp
+++ b/test/unittests/Validation/TestOpConstraintValidation.cpp
@@ -248,36 +248,24 @@ TEST_F(OpConstraintValidationTest, UpdateCacheOpWithInvalidUpdateIndexType) {
 }
 
 // Test ValidationStatus::NotImplemented
-// ScatterOp returns ArchitecturalMismatch which maps to NotImplemented
+// AllocOp returns MissingMetalDefinition which maps to NotImplemented
 TEST_F(OpConstraintValidationTest, ValidationStatusNotImplemented) {
-  llvm::SmallVector<int64_t> inputShape = {1, 1, 32, 32};
-  auto layout = createTiledLayout(inputShape, BufferType::L1,
+  llvm::SmallVector<int64_t> tensorShape = {1, 1, 32, 32};
+  auto layout = createTiledLayout(tensorShape, BufferType::L1,
                                   TensorMemoryLayout::Interleaved);
   auto tensorType =
-      mlir::RankedTensorType::get(inputShape, builder.getBF16Type(), layout);
+      mlir::RankedTensorType::get(tensorShape, builder.getBF16Type(), layout);
 
-  auto input = builder.create<OnesOp>(
-      builder.getUnknownLoc(), tensorType,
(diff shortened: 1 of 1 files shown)


include/ttmlir/OpModel/TTNN/SingletonDeviceContext.h
```diff
@@ -6,11 +6,17 @@
...
+#include "ttmlir/Dialect/TTCore/IR/TTCoreOpsTypes.h"
+#include "ttmlir/Dialect/TTCore/IR/Utils.h"
+
 #include <cassert>
 #include <cstddef>
 #include <memory>
 
+namespace mlir {
+class Operation;
+} // namespace mlir
```
azecevic000, line 18: Is there a particular reason for using forward declaration here?

rpavlovicTT:
tt-xla perf bench run https://github.com/tenstorrent/tt-xla/actions/runs/21829885259

include/ttmlir/OpModel/TTNN/SingletonDeviceContext.h
```diff
@@ -6,11 +6,17 @@
 #define TTMLIR_OPMODEL_TTNN_SINGLETONDEVICECONTEXT_H
```
odjuricicTT, line 6: Can we completely remove `createDevicePassesWrapper` with this change? That would be nice.

Doesn't have to be part of this PR, but at least open na issue and assign.

rpavlovicTT: Yes potentially, but then we need to make following changes too:
- Remove ExternalDevice setting once we verify everything works with mock device. 
- Move tensorL1UsageCap attribute setting into TTNNPipelines.cpp directly (a few lines before/after the pass
  group)
-  Move TT_METAL_DISABLE_BACKTRACE into openMockDevice()

Then individual passes' ScopedSingletonDeviceGuard handles everything else.

rpavlovicTT: opened https://github.com/tenstorrent/tt-mlir/issues/7151 for it

PR #6962, opened 2026-02-09
Use LLD as a library instead of shelling out to `ld.lld-17`

dmilinkovicTT:
### Ticket
[Link to Github Issue](https://github.com/tenstorrent/tt-forge/issues/869)

### Problem description
`LLVMToDynamicLib`, key component of CPU-hoisting mechanism, relies on shelling-out to `ld.lld-17`. This is problematic because it imposes environment requirements for running `tt-mlir` with CPU-hoisting turned on.

For example, `forge-ubuntu-22-04-py3-11` docker image, used for running perf benchmarks, doesn't have clang installed, and CPU-hoisting [fails](https://github.com/tenstorrent/tt-xla/actions/runs/21820816400/job/62955317017) because of missing `ld.lld-17`.

As we don't want to impose any dev tooling-related constraints on the machines running `tt-mlir`, we must:
- build `LLD` from the `LLVM` source
- link LLD lib with `TTLLVMToDynamicLib` target
- use `lld::lldMain` instead of shelling-out

### Changes in artifact sizes
- `ttmlir-translate` - 1600MB -> 2290MB
- `_ttmlir.cpython-311-x86_64-linux-gnu.so` - 1955MB -> 2021MB

### Checklist
- [x] Examine binary sizes before/after this change


wenbinlyuTT:
I'm a bit confused, the `forge-ubuntu-22-04-py3-11` docker image will still need to install `lld` correct? I'm assuming it doesn't inherit from the tt-mlir image. We don't actually need a `lld` with the exact same git commit as our MLIR dependency.

If so, then the only problem on tt-mlir's side is that it hard code `ld.lld-17`, and we can solve it by simply removing the hard coded version and creating symlink(s) like we already do for `clangd` https://github.com/tenstorrent/tt-mlir/blob/addf7c827d054da6b55af4e69fc485292391a33d/.github/Dockerfile.base#L59

I'm asking this because now more than half of the metal uplifts requires a rebuild of the docker image, and it's already painfully long. If we can get away with using an older `lld` then it would be great.

dmilinkovicTT:
Hey, @wenbinlyuTT!

> I'm a bit confused, the forge-ubuntu-22-04-py3-11 docker image will still need to install lld correct?

With this PR, we're not requiring to have `lld` installed anywhere - we're building it from source and linking it statically to our compiler.

> If so, then the only problem on tt-mlir's side is that it hard code ld.lld-17, and we can solve it by simply removing the hard coded version and creating symlink(s) like we already do for clangd

That would work in our dev environments, but the main issue this PR addresses is the fact that users shouldn't need to have `lld` installed in order for `tt-mlir` to work.

nsumrakTT (approved):
Great work! Thank you for the quick solution that affected release requirements for all FEs.

PR #6964, opened 2026-02-09
Add named argument support

alexbaden:
Adds support for a new class of `ttnn` / `ttkernel` argument type to express constexpr scalar arguments to `ttnn.generic` op kernel functions. We use this in the Triton backend to pass per core arguments to the kernels (namely a start/end block offset) but it seemed simpler to pass the argument as a common RT arg in the tests. I added one lit test but would gladly take suggestions for other places where this functionality can be tested. 

@nsmithtt helped me sketch out the initial set of types. I made some modifications based on testing with the Triton backend and setting up the lit test but most of the changes are related to plumbing the args into the runtime.

### DIFF
--- a/test/ttmlir/Silicon/TTNN/n150/generic_op/generic_op.mlir
+++ b/test/ttmlir/Silicon/TTNN/n150/generic_op/generic_op.mlir
@@ -1,4 +1,4 @@
-// RUN: ttmlir-opt --ttcore-register-device="system-desc-path=/home/abaden/Projects/tenstorrent/tt-mlir/ttrt-artifacts/system_desc.ttsys" -o %t.mlir %s
+// RUN: ttmlir-opt --ttcore-register-device="system-desc-path=%system_desc_path%" -o %t.mlir %s
 // RUN: FileCheck %s --input-file=%t.mlir
 // RUN: ttmlir-translate --ttnn-to-flatbuffer -o %t.ttnn %t.mlir
 


sohaibnadeemTT:
~@nsmithtt @alexbaden I think it would be better to call it "named crt arg" to be more explicit here~

nsmithtt (commented):
@sohaibnadeemTT and I discussed offline, we think the naming is OK to stay the way it's proposed in this PR.

Precommit is failing CI, you'll have to locally run and then commit these changes:
```bash
pre-commit run --all-files
```

Otherwise I think the PR LGTM!

nsmithtt:
Oh, never mind my comment regarding precommit, it seems like GitHub is just having a bad day:
```error
Error: fatal: unable to access 'https://github.com/tenstorrent/tt-mlir/': The requested URL returned error: 502
```

I'll try kicking CI again this afternoon.

alexbaden:
Yeah, GitHub is having a really bad day… I did run pre_commit but can check again once GitHub recovers.

test/ttmlir/Silicon/TTNN/n150/generic_op/generic_op.mlir
```diff
@@ -157,9 +157,10 @@ module {
     emitc.call_opaque "cb_pop_front"(%2, %0) : (!emitc.opaque<"::tt::CB">, i32) -> ()
     return
   }
-  func.func private @write_kernel() attributes {tt.function_type = "kernel", ttkernel.arg_spec = #ttkernel.arg_spec< rt_args = [<arg_type = buffer_address, operand_index = 0>] ct_args = [<arg_type = cb_port, operand_index = 0>]>, ttkernel.thread = #ttkernel.thread<noc>} {
+  // Test named arguments by passing page size as a constant
+  func.func private @write_kernel() attributes {tt.function_type = "kernel", ttkernel.arg_spec = #ttkernel.arg_spec< rt_args = [<arg_type = buffer_address, operand_index = 0>, <arg_type = named_argument, argument_name = "page_size", operand_index = 1 >] ct_args = [<arg_type = cb_port, operand_index = 0>]>, ttkernel.thread = #ttkernel.thread<noc>} {
     %0 = "emitc.constant"() <{value = 1 : i32}> : () -> i32
-    %1 = "emitc.constant"() <{value = 4096 : i32}> : () -> i32
+    %1 = emitc.call_opaque "get_arg_val"(%0) {template_args = [#emitc.opaque<"uint32_t">]} : (i32) -> i32
```
nsmithtt: There are CI failures due to a device hang.  I think there's a bug here because `%0` is set to `1`, we need a new constant `%zero = "emitc.constant"() <{value = 0 : i32}> : () -> i32` and to use that here.

It's possible this could cause a hang because get_arg_val will return garbage which could be a huge noc tx size that reads off the end of mem.

alexbaden: The intended behavior was to pull the 4096 constant from a runtime arg, but I think you're right there's a bug. I'm sticking 4096 in `common_rt_args` but using `get_arg_val`. 

So, I could try to update to use `rt_args` (per core) or try `get_common_arg_val` with index 0. We are using `rt_args` and not `common_rt_args`, but it might be worth testing with `common_rt_args` too... 
How can I reproduce this on my wormhole card locally?

nsmithtt: No I think it's fine to use `get_arg_val` it's just that right now you're doing `get_arg_val(1)`, where it should be `get_arg_val(0)`.

PR #6967, opened 2026-02-09
Update python path to use same .so file everywhere

vwellsTT:
### Ticket
NA

### Problem description
Some custom pytest commands could cause a crash because a nanobind was regisered multiple times--investigation showed that this was due to multiple .so libs existing and be loaded in the PYTHONPATH

### What's changed
Removed a dir from the PYTHONPATH, ensure all imports of ttrt.runtime's .so use the same file

### Checklist
- [X]Existing tests provide coverage for changes
- [X] manual confirmation prev. failing commands work fine now

### DIFF
--- a/tools/ttnn-jit/_src/__init__.py
+++ b/tools/ttnn-jit/_src/__init__.py
@@ -69,7 +69,7 @@ except ModuleNotFoundError:
         load_binary_from_path,
         load_binary_from_capsule,
     )
-    from ttrt.runtime._ttmlir_runtime.utils import (
+    from _ttmlir_runtime.utils import (
         create_runtime_device_from_ttnn,
         create_runtime_tensor_from_ttnn,
         get_ttnn_tensor_from_runtime_tensor,


tapspatel:
ttrt is no longer included in any builder codepath. so it should be completely decoupled from it.

test/python/golden/conftest.py
```diff
@@ -2,7 +2,7 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 import pytest
-import _ttmlir_runtime as tt_runtime
+import ttrt.runtime._ttmlir_runtime as tt_runtime
```
tapspatel, line 5: we should not import ttrt in this situation, the runtime nano binds should be enough. Something else must be importing ttrt

tapspatel:
The other places I see ttrt being used is 

test/lit.cfg.py
runtime/test

tapspatel:
@nsmithtt I think ttrt should include _ttmlir_runtime because it needs a copy of the nano binds since it gets installed out of tree. However, the in source tree code should not use ttrt, it should use the built nanobind.so directly.

nsmithtt:
> @nsmithtt I think ttrt should include _ttmlir_runtime because it needs a copy of the nano binds since it gets installed out of tree. However, the in source tree code should not use ttrt, it should use the built nanobind.so directly.

Ok, yeah makes sense. So maybe we can't change the python path like this, what's weird is that this test reproduces the nanobind double load issue:
```
pytest test/python -k "test_reshape[ttmetal-f32-96x64_to_3x32x64]" -x -s
```

As far as I can tell, everything underneath `test/python` does not reference ttrt so it's unclear who's loading the ttrt version.

vwellsTT:
>@nsmithtt I think ttrt should include _ttmlir_runtime because it needs a copy of the nano binds since it gets installed out of tree. However, the in source tree code should not use ttrt, it should use the built nanobind.so directly.

I think that would work too, hmm.  Honestly I'm a bit confused why we have 2 copies of the .so to start with in the build dir, is that itself not a bad thing? (It seems be copied by shutil I think)

vwellsTT:
>As far as I can tell, everything underneath test/python does not reference ttrt so it's unclear who's loading the ttrt version.

Yeah, I actually couldn't find that either...must be multiple levels of indirection or something.

But I'm empirically confident it exists somewhere haha

vwellsTT:
I think longer-term fix is a bit more involved, doesn't seem like this is a blocker on anything other than weird pytest usage atm, closing

PR #6968, opened 2026-02-09
[d2m] L1 scratchpad for spilling in fused kernels

ckaravasilisTT:
L1 scratchpad implementation for the purpose of storing intermediate fusion results. The scratch buffer is materialized as an input operand to every `d2m.generic` that may need it. This approach is likely temporary until we have cleaner mechanisms for creating and handling CBs.

Two new ops: `get_scratch_from_cb` takes a scratch cb and returns a memref. `scratch_allocate` requests an L1 buffer.
Two new passes:
- `AddScratchInputs`: Goes over every `d2m.generic` and adds a scratch input to the ones that may need it (currently every generic with more than one binary FPU op). Scratch size is currently fixed  to 128KB of the same data format as the other inputs.
- `LowerScratchAllocate`: Introduces a `get_scratch_from_cb` op at the start of the generic and replaces each `scratch_allocate` with a subview into the scratch.

`get_scratch_from_cb` is lowered into `d2m.reserve` in `ConvertLocalLoadStoreOpsToAliasedCBs`.

### DIFF
--- a/include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
+++ b/include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
@@ -1839,7 +1839,7 @@ def D2M_GetScratchFromCBOp : D2M_GenericRegionOp<"get_scratch_from_cb",
   }];
 
   let arguments = (ins D2M_CB:$cb);
-  let results = (outs AnyNon0RankedMemRef:$result);
+  let results = (outs DeviceL1MemRef:$result);
 
   let assemblyFormat = [{ $cb attr-dict `:` type($cb) `->` type($result) }];
 


include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
```diff
@@ -1788,6 +1788,88 @@ def D2M_PopOp : D2M_GenericRegionOp<"pop",
...
+
+  let extraClassDeclaration = [{
+    CBType getCbType() {
+      return mlir::cast<CBType>(getCb().getType());
+    }
+  }];
+
+  let extraClassDefinition = [{
+    mlir::LogicalResult $cppClass::verify() {
+      auto cbType = mlir::cast<CBType>(getCb().getType());
```
bgrady-tt, line 1861: would also assert L1 memspace here 👍

include/ttmlir/Dialect/D2M/Transforms/Passes.td
```diff
@@ -930,4 +977,56 @@ def D2MConvertLocalLoadStoreOpsToAliasedCBs : Pass<"d2m-convert-local-load-store
...
+  let description = [{
+    This pass lowers `d2m.scratch_allocate` operations inside `d2m.generic` ops
+    into rank-reducing `memref.subview` operations of the scratch memref.
+
+    For each `d2m.generic` that has a `scratch_inputs` attribute, the pass:
+    1. Collects all `d2m.scratch_allocate` ops in the compute region and sorts
+       them by `slot` ID for deterministic layout.
+    2. Computes sequential element offsets for each allocation.
+    3. Validates that total allocations fit within the scratch buffer capacity.
+    4. Introduces a `d2m.get_scratch_from_cb` op at the start of the generic
```
bgrady-tt, line 992: I think this makes sense as-is 👍

include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
ckaravasilisTT: I changed the result type to `DeviceL1MemRef`

include/ttmlir/Dialect/D2M/Transforms/Passes.td
```diff
@@ -7,6 +7,53 @@
 
 include "mlir/Pass/PassBase.td"
 
+// Add scratch inputs to d2m.generic ops (post-bufferization)
```
nsmithtt, line 10: Just throwing this out there.  Might it be better to run this pass after fusion, we could find all `memref.alloc` ops that DONT participate in a `d2m.load` or `d2m.store` op and manifest an operand for each of them?  Then we don't need to automatically burn 128k no matter what.

ckaravasilisTT: The scratch size right now is basically a placeholder. We definitely want a more precise estimate, but getting this early means we have to be aware of what types of fusion we support and how much they spill. The only way scratch space is requested right now is only through `scratch_allocate` ops, which don't appear until later in the pipeline. I agree with not wanting to burn 128k no matter what, but it's not obvious to me how to infer accurate scratch requirements right now (with a number of fusion changes underway).

PR #6970, opened 2026-02-10
[D2M] Fix higher-rank ttir.sum lowering in D2MNamedReductionRewriter

### LINKED ISSUE #6647: [d2m] higher rank ttir.sum support
The following mlir snippet:
```mlir
  func.func @component_1_ttir_sum_18(%arg0: tensor<32x1x8192xf32>) -> tensor<32x1x1xf32> {
    %0 = "ttir.sum"(%arg0) <{dim_arg = [2 : i32], keep_dim = true}> : (tensor<32x1x8192xf32>) -> tensor<32x1x1xf32>
    return %0 : tensor<32x1x1xf32>
  }
```

Fails with:
```
error: 'd2m.generic' op shard shape mismatch between operand[0] shard_shape=[1024, 8192] and operand[2] shard_shape=[1024, 32] at affine dim d1
    %0 = "ttir.sum"(%arg0) <{dim_arg = [2 : i32], keep_dim = true}> : (tensor<32x1x8192xf32>) -> tensor<32x1x1xf32>
```

nsmithtt:
The reduction rewriter collapsed all tensors to 2D, corrupting dimension mapping for rank > 2 inputs. Pass noCollapse=true for higher-rank tensors and fix dimArgAsReduceDim to determine ReduceDim relative to the last two dimensions instead of using absolute bit positions.

Closes #6647

### DIFF
--- a/.github/workflows/call-build-debug.yml
+++ b/.github/workflows/call-build-debug.yml
@@ -75,6 +75,8 @@ jobs:
 
     - name: Build
       id: build
+      env:
+        CMAKE_BUILD_PARALLEL_LEVEL: "16"
       shell: bash
       run: |
         source env/activate


test/python/golden/test_metal_reductions.py
```diff
@@ -69,6 +69,75 @@ def test_sum(
     )
 
 
+@pytest.mark.skip_config(["p150"], ["p300"])
```
vwellsTT: nit: imo when we skip a test, probably best always have some TODO + ticket (assuming test should eventually not be skipped)

vwellsTT (approved):

test/python/golden/test_metal_reductions.py
nsmithtt: Good point

nsmithtt:
Pipeline is failing for an exposed bug fixed here:
- https://github.com/tenstorrent/tt-mlir/pull/6985

Issue #6877, opened 2026-02-03
const_eval introduces significant DRAM overhead for larger sequences


umalesTT:
During prefill generality effort it was noted that const_eval produces tensors of shape
```batch_size x num_attention_heads x sequence_length x sequence_length```
in fp32 precision. This is a big problem, given that we wish to scale sequence_length to much larger values than current 128. For even moderately sized lengths (e.g. 4096) this results in several GBs of DRAM overhead and makes attention effectively operate on 5x larger batch memory-wise.

Even with const_eval executing on cpu now, this still represents a problem, since output tensor will end up in our DRAM. 
I synced with @dmilinkovicTT offline, and a straightforward unblocker would be to hard-limit const_eval graphs on memory usage / size of their outputs (per-graph), while a longer term fix would require more discussion and a well-defined heuristics.

IRs below for reference

Llama 1b [ttnn graph](https://gist.github.com/umalesTT/3bd65f0eff28e316195189c136903869) (graphs 11 and 20 for reference, with seq_len = 128)
Llama 1b [ttir graph](https://gist.github.com/umalesTT/41061e662fd9613202e716dcbd72061b)
 
Llama 8b [ttnn graph](https://gist.github.com/umalesTT/0cf8259232a1571d1cb7a618a23080bf) (graph 171, for example)
Llama 8b [ttir graph](https://gist.github.com/umalesTT/1943e3dbbd27a1b3fe382f83d78c3474)

dmilinkovicTT:
Verified both graphs attached in the issue description against the latest tt-mlir, and all large constant tensors have been shrunk to scalars.

umalesTT:
Had to reopen because of https://github.com/tenstorrent/tt-mlir/pull/7353

dmilinkovicTT:
Implicit broadcasting re-enabled in #7552

PR #6973, opened 2026-02-10
Fold FullOp to scalar if users support implicit broadcasting

dmilinkovicTT:
### Ticket
#6877 

### Problem description
The following attention masking patterns occur in the llama 1B graph (attached in the issue):

#### Pattern 1

```
%2 = "ttir.constant"() <{value = dense<0xFF80> : tensor<2x1x128x128xbf16>}> : () -> tensor<2x1x128x128xbf16> loc(#loc)
...
%111 = "ttir.where"(%110, %3, %2) : (tensor<2x1x128x128xi1>, tensor<2x1x128x128xbf16>, tensor<2x1x128x128xbf16>) -> tensor<2x1x128x128xbf16> loc(#loc241)
```

#### Pattern 2

```
%1 = "ttir.constant"() <{value = dense<0xFFF0000000000000> : tensor<2x32x128x128xf64>}> : () -> tensor<2x32x128x128xf64> loc(#loc)
...
// many times:
%A = "ttir.eq"(%B, %1) : (tensor<2x32x128x128xf64>, tensor<2x32x128x128xf64>) -> tensor<2x32x128x128xi1> loc(#loc547)
```

The mask tensor size is proportional to the square of the sequence length, which takes a large portion of DRAM space, since these tensors are const-evaled, and thus never deallocated.

### What's changed
- Implemented folding `FullOp`, `ZerosOp` and `OnesOp` to 1-volume tensors if all their users support implicit operand broadcasting
  - This will lead to a negligible performance penalty, since TTNN ops should handle such implicit broadcasts as scalar broadcasts, but will lead to significant DRAM savings

### Note

Mask from the aforementioned Pattern 1 will still get explicitly broadcasted, since `WhereOp` has `ExplicateOperandBroadcastsTrait` The efforts for validating the current state of implicit broadcasting for `WhereOp` are WIP - #6995.

### Checklist

### DIFF
--- a/lib/Dialect/TTIR/Transforms/Broadcast.cpp
+++ b/lib/Dialect/TTIR/Transforms/Broadcast.cpp
@@ -5,18 +5,63 @@
 #include "ttmlir/Dialect/TTCore/IR/TTCore.h"
 #include "ttmlir/Dialect/TTIR/Transforms/Passes.h"
 #include "ttmlir/Dialect/TTIR/Utils/Utils.h"
+#include "ttmlir/Utils.h"
 
 #include "mlir/Transforms/GreedyPatternRewriteDriver.h"
+#include "llvm/ADT/SetVector.h"
 
 namespace mlir::tt::ttir {
 #define GEN_PASS_DEF_TTIRIMPLICITBROADCASTFOLD
+#define GEN_PASS_DEF_TTIRFOLDFULLTOSCALAR
 #include "ttmlir/Dialect/TTIR/Transforms/Passes.h.inc"
 
-// This algorithm works by first removing all explicit broadcasts from the
-// operands of an operation. While doing so, it calculates the result shape that
-// would result from implicit broadcasting, taking all operands into
-// consideration. If this shape differs from the target shape, we add an
-// explicit broadcast to the operation’s output to match the target shape.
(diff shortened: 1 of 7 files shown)


lib/Dialect/TTIR/Transforms/Broadcast.cpp
```diff
@@ -5,18 +5,63 @@
 #include "ttmlir/Dialect/TTCore/IR/TTCore.h"
 #include "ttmlir/Dialect/TTIR/Transforms/Passes.h"
 #include "ttmlir/Dialect/TTIR/Utils/Utils.h"
+#include "ttmlir/Utils.h"
 
 #include "mlir/Transforms/GreedyPatternRewriteDriver.h"
+#include "llvm/ADT/SetVector.h"
```
jserbedzijaTT, line 11: Nit: this include is probably not needed

jserbedzijaTT (approved):
Looks good!

lib/Dialect/TTIR/Transforms/Broadcast.cpp
dmilinkovicTT: `SetVector` is actually being used in `CreationToScalarRewriter`

PR #6985, opened 2026-02-10
Remove D2M_PackerMaskResetOp

### LINKED ISSUE #6978: [d2m] BH support `test/python/golden/test_metal_reductions.py`
Currently `test/python/golden/test_metal_reductions.py` subtests all fail with bad PCC on blackhole.  We need to debug and root cause why.

nsmithtt:
This op leaks an llk quirk for reduce into d2m dialect, i.e. the reduce_uninit op see #22904 for more info. Originally we modeled this op in d2m because we could use the info during InsertDstAccess pass to insert the uninit in an optimal way, however, this insertion no longer works for higher rank reductions.

The fix consists of a cleanup and a simplification:
- Remove D2M_PackerMaskResetOp from d2m entirely.
- Unconditionally insert the uninit op, this is non-optimal, but we can add support to the HoistInit pass going forward to safely lift out of loops.

This also fixes the BH PCC errors, enabled reduction tests for BH.

Closes #6978

### DIFF
--- a/test/python/golden/test_metal_reductions.py
+++ b/test/python/golden/test_metal_reductions.py
@@ -39,7 +39,6 @@ def create_reductions_constrained_inputs(input_shape, reduce_type, dim_arg, keep
     return module
 
 
-@pytest.mark.skip_config(["p150"], ["p300"])
 @pytest.mark.parametrize("m", [4, 8, 16])
 @pytest.mark.parametrize("n", [2, 4, 8])
 @pytest.mark.parametrize("dim_arg", [[0], [1], [0, 1]])
@@ -69,7 +68,6 @@ def test_sum(
     )
 
 
-@pytest.mark.skip_config(["p150"], ["p300"])
 @pytest.mark.parametrize("m", [4, 8, 16])
 @pytest.mark.parametrize("n", [2, 4, 8])
 @pytest.mark.parametrize("dim_arg", [[0], [1]])


include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
```diff
@@ -639,14 +639,6 @@ def D2M_TileBcastOp : D2M_GenericRegionComputeOp<"tile_bcast", [Pure, D2M_SkipOp
...
 }
 
-def D2M_PackerMaskResetOp : D2M_GenericRegionComputeOp<"packer_mask_reset"> {
-  let summary = "D2M Packer Mask Reset Op";
-  let description = [{
-        The `packer_mask_reset` operation resets the packer mask. This should be removed once LLK refactors the reduce API.
-        See tt-metal#22904 and tt-metal -> reduce.h.
-    }];
-}
-
```
nsmithtt, line 649: The crux of the changes:
1. Remove D2M_PackerMaskResetOp and related rewriters

lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
```diff
@@ -692,6 +692,7 @@ class D2MFPUOpsRewriter : public OpConversionPattern<ConcreteOp> {
       rewriter.create<ttkernel::ReduceTileOp>(
           op->getLoc(), cbA, cbB, adaptor.getA(), adaptor.getB(),
           adaptor.getC(), reduce_type, kernel_reduce_dim);
+      rewriter.create<ttkernel::ReduceUninitOp>(op->getLoc());
```
nsmithtt, line 695: The crux of the changes:

2. Unconditionally add ReduceUninitOp in the same way we insert Init ops.  We can explore hoisting as a future optimization.

Issue #6974, opened 2026-02-10
Uplift blocker due to: `TT_FATAL: Invalid arguments to reshape (assert.hpp:104)`


acicovicTT:
### Describe the bug

tt-mlir cannot be uplifted to tt-xla due to the following error:

> TT_FATAL: Invalid arguments to reshape (assert.hpp:104)

Commit that introduced the regression is https://github.com/tenstorrent/tt-mlir/commit/bdedd3bb52d229bbf31164f25fe0ec1a3582795b

Example of failed workflow, where logs can be found: https://github.com/tenstorrent/tt-xla/actions/runs/21854266376/job/63084777525

### Steps to reproduce the issue

Checkout `bdedd3bb52d229bbf31164f25fe0ec1a3582795b` and run any of the tests that are failing in the workflow above (except the perf benchmark, which was failing with another root cause).

tapspatel:
Repro steps

```
git clone git@github.com:tenstorrent/tt-xla.git
cd tt-xla
git submodule update --init --recursive
source venv/activate
cmake -G Ninja -DCMAKE_BUILD_TYPE=Debug -DUSE_CUSTOM_TT_MLIR_VERSION=bdedd3bb52d229bbf31164f25fe0ec1a3582795b -B build
cmake --build build
pytest -ssv tests/torch/test_torch_xla_basic.py::test_spmd_sharding[partially_sharded-input_shape0-axis_names0]
```

Running the graph in tt-xla produces the assert on reshape, but taking the graph to tt-mlir and running there does not reproduce this issue. Still investigating

tapspatel:
The error is related to torch xla inserting a dummy graph to gather sharded output. However, I was under the assumption that if an argument/result has a sdy.sharding annotation, it is presharded. else it is unsharded. Was not told this is not the case. Asked frontend multi chip folks what the correct solution is.

In the meantime, you can revert the commit if its blocking, otherwise I will try to get in a solution within another business day depending on answers.

tapspatel:
Since the blocker was merged, I will close this issue. Please reopen if any other problem persists.

PR #6988, opened 2026-02-11
Revert "Support presharded/unsharded graphs in builder (#6932)"

acicovicTT:
This reverts commit bdedd3bb52d229bbf31164f25fe0ec1a3582795b.

### Ticket

https://github.com/tenstorrent/tt-mlir/issues/6974


acicovicTT:
As agreed with tapspatel, I merged this revert because there are 10+ commits waiting for uplift, and also Mon and Tue are state holidays for the XLA team here, so I would like to push the tt-mlir ref as close to tt-mlir main as possible before the long weekend

PR #6989, opened 2026-02-11
Update StableHLO Sort Op Conversion

sdjukicTT:
### Ticket
/

### Problem description
StableHLO Sort Op sometimes produces TTIR Gather Ops, but we want to remove the generic TTIR Gather Op that mimics StableHLO Gather (https://github.com/tenstorrent/tt-mlir/issues/6579).

### What's changed
kKeyValue SortType case of StableHLOToTTIRSortOpConversionPattern handles a sort op with multiple inputs: first input is sorted along a dimension, and other inputs are rearranged in the same order.

Changed the kKeyValue case of the conversion pattern - removed TTIR Gather Ops and necessary transformations for its input, now TTIR Embedding Op is emitted directly.

### Builder changes
Changed StableHLO Builder to support kKeyValue SortType case (with multiple inputs and outputs) and added a test.

### How this case is handled now
Steps:
1. Permutation of sort indices that moves sortDim to rank-1.
2. Reshape sort indices to [prePost, dSort].
3. Compute flat indices: ArangeOp(0, total, step=dSort, dim=0) + indices_2d.
4. Per value tensor: permute so sortDim is last, flatten to [total, 1], EmbeddingOp, reshape to permuted shape, permute back.

### Checklist
- changed test to match new decomposition
- checked results manually and using builder for different cases

### DIFF
--- a/lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
+++ b/lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
@@ -5249,7 +5249,7 @@ private:
 } // namespace
 
 namespace {
-// Conversion: stablehlo::SortOp → ttir::SortOp + optional ttir::EmbeddingOp(s)
+// Conversion: stablehlo::SortOp -> ttir::SortOp + optional ttir::EmbeddingOp(s)
 //
 // StableHLO's SortOp supports sorting tuples of tensors with an arbitrary
 // comparator function. This pattern lowers such SortOps into TTIR by
@@ -5381,8 +5381,8 @@ public:
     // where i is the row (linearized non-sort dims) and j is the column
     // (sortDim). The stride of the sort dimension is now always 1, so we only
     // need to add a row offset: i * dSort. That offset pattern is exactly what
-    // ArangeOp(start=0, end=total, step=dSort, dim=0) produces — one entry per
-    // row, broadcast across columns — making the full formula just an addition:
+    // ArangeOp(start=0, end=total, step=dSort, dim=0) produces - one entry per
+    // row, broadcast across columns - making the full formula just an addition:
     //   flat_indices = ArangeOp(step=dSort) + indices_2d
     //
     // Steps:


lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
```diff
@@ -5362,97 +5361,122 @@ class StableHLOToTTIRSortOpConversionPattern
...
+    SmallVector<int64_t> perm;
+    perm.reserve(rank);
+    for (int64_t i = 0; i < rank; ++i) {
+      if (i != sortDim) {
+        perm.push_back(i);
+      }
+    }
+    perm.push_back(sortDim);
+    SmallVector<int64_t> invPerm = ttmlir::utils::inversePermutation(perm);
+    bool needsPermute = (sortDim != rank - 1);
```
azecevic000: You don't need this flag, folder should deal with a case when permutation is a no-op, the main logic stays clean.

lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
```diff
@@ -5362,97 +5361,122 @@ class StableHLOToTTIRSortOpConversionPattern
...
+
+    // Permute indices so sortDim is last, then reshape to [prePost, dSort].
+    Value indices2D = indices;
+    if (needsPermute) {
+      SmallVector<int64_t> permShape =
+          ttmlir::utils::applyPermutation(origShape, perm);
+      auto permType = RankedTensorType::get(permShape, indexElemType, encoding);
```
azecevic000: What's the purpose of `encoding` here? Are you sure that the change of shape won't cause encoding to be invalid if it exists? If the expectation is that encoding doesn't exist you don't have to use it. The same question is for other tensor type constructors as well.

lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
```diff
@@ -5362,97 +5361,122 @@ class StableHLOToTTIRSortOpConversionPattern
...
-    for (int64_t idx = 0; idx < rank; ++idx) {
-      if (idx == sortDim) {
-        toConcat.push_back(reshape);
-        continue;
+    ArrayRef<int64_t> origShape = indicesType.getShape();
+    Type indexElemType = indicesType.getElementType();
+    auto encoding = indicesType.getEncoding();
+
+    int64_t dSort = origShape[sortDim];
+    int64_t prePost = 1; // product of all non-sort dimensions
```
azecevic000: `indiciesType.getNumElements() / dSort`

lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
```diff
@@ -5362,97 +5361,122 @@ class StableHLOToTTIRSortOpConversionPattern
       return success();
     }
 
-    // SortType::kKeyValue — sort additional inputs using indices and GatherOp.
+    // SortType::kKeyValue — reorder additional inputs using sort indices
+    // and EmbeddingOp with flat indexing.
+    //
+    // Strategy: permute sortDim to the last position so the flat index formula
+    // reduces to a simple row-offset + sort_index with stride 1:
+    //   flat[i, j] = i * dSort + indices_2d[i, j]
```
acicovicTT: Might be worth explaining that dSort is size of sortDim, and that indices_2d is the output of ttir::SortOp

lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
```diff
@@ -5362,97 +5361,122 @@ class StableHLOToTTIRSortOpConversionPattern
       return success();
     }
 
-    // SortType::kKeyValue — sort additional inputs using indices and GatherOp.
+    // SortType::kKeyValue — reorder additional inputs using sort indices
+    // and EmbeddingOp with flat indexing.
+    //
+    // Strategy: permute sortDim to the last position so the flat index formula
+    // reduces to a simple row-offset + sort_index with stride 1:
```
acicovicTT: I would try to expand on this explanation. I did not understand the message before I took the time to understand the algorithm in detail

lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
```diff
@@ -5362,97 +5361,122 @@ class StableHLOToTTIRSortOpConversionPattern
...
+
+    // Flat indices: row i gets offset i*dSort, so flat[i,j] = i*dSort +
+    // idx[i,j].
+    auto rowOffsets = rewriter.create<ttir::ArangeOp>(
+        loc, indices2DType, /*start=*/0, /*end=*/total,
+        /*step=*/dSort, /*arange_dimension=*/0);
+    Value flatIndices =
+        rewriter.create<ttir::AddOp>(loc, indices2DType, rowOffsets, indices2D);
+
+    // Per value tensor: permute sortDim to last → flatten → EmbeddingOp →
```
acicovicTT: Please re-review all AI-generated comments and remove non-ASCII characters

lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
sdjukicTT: Removed it.

lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
sdjukicTT: Removed encoding from all type constructors.

lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
sdjukicTT: Fixed.

lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
sdjukicTT: Removed non-ASCII characters.

lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
sdjukicTT: Added a more detailed explanation.

PR #6995, opened 2026-02-11
Enable implicit broadcast for WhereOp

dmilinkovicTT:
### Ticket
#6877 

### Problem description
`WhereOp` now supports implicit broadcasting in `tt-metal`, so we can remove the workaround which explicates the broadcasts. 

This is especially beneficial for attention masks, since we can avoid materializing the full mask, and we can materialize only the splat value instead.

### Checklist


dmilinkovicTT:
Blocked due to tt-metal issue [#37632](https://github.com/tenstorrent/tt-metal/issues/37632)

PR #6998, opened 2026-02-11
Adding device ops handling to profiler package

jgrimTT:
### Ticket
None

### Problem description
Profiler doesn't support device ops

### What's changed
Added environment variable handling for device profiling.
Fixed `profiler_logs_dir` doesn't exist bug

### Checklist

### DIFF
--- a/tools/profiler/profiler.py
+++ b/tools/profiler/profiler.py
@@ -13,7 +13,7 @@ from tracy.process_ops_logs import process_ops
 
 
 @contextmanager
-def trace(log_dir: str, port: int, host_only: bool = False):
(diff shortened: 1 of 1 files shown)


tools/profiler/profiler.py
```diff
@@ -13,7 +13,7 @@
 
 
 @contextmanager
-def trace(log_dir: str, port: int):
+def trace(log_dir: str, port: int, host_only: bool = False):
```
tapspatel: tbh I think we should just always keep device profiling on and remove host_only

tapspatel (approved):

PR #7007, opened 2026-02-11
Integration of optimizer and D2MSubgraphOp

sgholamiTT:
### Ticket
https://github.com/tenstorrent/tt-mlir/issues/7029

## Summary

This PR integrates the TTNN memory-layout optimizer with `ttnn.d2m_subgraph` by implementing the constraint API for the op: D2M subgraph ops are now included in DFSharding-based layout analysis, and the optimizer correctly applies chosen layouts to the dispatch op, its output buffer(s), and the referenced D2M function. For now, **D2M is allowed to spill to DRAM**: producer→D2M edges use the default path (no L1 chain through D2M), so the optimizer places D2M input/output and its buffer in DRAM and keeps the D2M callee's types in sync with the dispatch after layout/spill decisions.

### 1. D2MSubgraphOp OpModel interface (`TTNNOpModelInterface.cpp`)

- **`getOpConstraints`**  
  The implementation walks the D2M function body in program order and, for each internal op that implements the OpModel interface:
  - Builds input layouts from a value-to-layout map: block args use the D2M's `inputs` (and `opConfig.outputLayout` for any extra args); SSA values from earlier ops use the layout stored when that op was processed.
  - Calls `getOpConstraints(internalOpInputLayouts, opConfig)` for the internal op.
  - Takes the **element-wise max** of L1 constraint fields (cbL1PeakSize, tensorL1PeakSize, peakL1MemorySize, outputL1BufferSize) and keeps the D2M's output layout.

- **`getOpRuntime`**  
  Similarly walks the body, builds per-op input layouts via the same value-to-layout map, calls `getOpRuntime` for each internal op, and returns the **sum** of their runtimes. The value-to-layout map uses each internal op's **actual** output layout (from the result type when present) so that a trailing `ToLayoutOp` (if present) is costed with the correct input layout.

### 2. Optimizer support for D2MSubgraphOp (`Optimizer.cpp`)

- **`applyChosenLayoutToD2MSubgraphOp`**  
  When the optimizer picks a layout for a D2M dispatch op, this function:
  - Sets the dispatch op's result type(s) to the new tensor type.
  - Updates each output buffer (Empty op): type, dtype, layout (tile/row-major), and memory config (including shard spec when applicable).
  - Updates the D2M function: entry block argument types from the dispatch's input types and the function's type signature (return type = chosen layout). Internal body op result types are **not** rewritten; instead, a single **`ttnn.ToLayoutOp`** is inserted at the end of the D2M function to convert the last op's output to the chosen layout X, and the function returns that result.

- **`syncD2MFuncTypesToDispatchInputs`**  
  After layout/spill decisions (e.g. a producer spilled to DRAM), syncs the D2M function's argument and result types to the dispatch op's current input/output types so the callee signature matches the caller.

- **Integration**  
  - When applying the chosen layout to an op, `D2MSubgraphOp` is handled by calling `applyChosenLayoutToD2MSubgraphOp` instead of the default path.
  - A post-processing walk over the function calls `syncD2MFuncTypesToDispatchInputs` on every `D2MSubgraphOp` so D2M callee types stay in sync after any spill or layout change.

## Tests

- **`d2m_optimizer_linear_chain.mlir`**  
  Linear chain matmul → matmul → d2m_subgraph (add, multiply) → matmul. With D2M spilling to DRAM, the optimizer places the chain in DRAM (matmul results, empty buffer, and D2M in/out in DRAM; function args remain L1). Checks that layouts and the final `to_layout` to DRAM are applied correctly and that the D2M callee's types match the dispatch (DRAM in/out).

- **`d2m_optimizer_fork_join.mlir`**  
  Fork-join pattern with a D2M subgraph on one branch. Checks that the optimizer applies layouts and spill (`to_layout`) consistently, that the D2M branch uses a DRAM output buffer and result, and that the D2M callee's types (including DRAM when spilled) stay in sync with the dispatch op.

## Notes

- D2MSubgraphOp with multiple results is not yet supported (assert in `applyChosenLayoutToD2MSubgraphOp`).
- Runtime for D2M is the sum of internal op runtimes; constraints use the max of internal op L1 usage so the scheduler sees a conservative peak for the subgraph.
- The [previous PR](https://github.com/tenstorrent/tt-mlir/pull/6863) that spilled to DRAM when a `D2M_SubgraphOp` was detected is no longer necessary.
- [This issue](https://github.com/tenstorrent/tt-mlir/issues/7025) tracks allowing `D2M_SubgraphOp` in L1 chain.
- [This issue](https://github.com/tenstorrent/tt-mlir/issues/7199) tracks building a cost model for unary/binary ops to avoid relying on ttnn for d2m_subgraph internals.

### Checklist

### DIFF
--- a/lib/Dialect/TTNN/Interfaces/TTNNOpModelInterface.cpp
+++ b/lib/Dialect/TTNN/Interfaces/TTNNOpModelInterface.cpp
@@ -4835,12 +4835,18 @@ D2MSubgraphOp::getOpConstraints(const std::vector<TTNNLayoutAttr> &inputs,
       internalOpInputLayouts.push_back(it->second);
     }
 
-    // Default constraints for the D2M subgraph op.
-    op_model::OpConstraints c(0, 0, 0, 0, opConfig.outputLayout);
+    // Use this op's actual output layout (from the map), not the D2M's
+    // top-level layout, so the constraint API is asked for the layout this op
+    // produces.
+    assert(op.getNumResults() > 0 && "OpModel op expected to have results");
+    TTNNLayoutAttr internalOpOutputLayout =
+        valueToLayout.lookup(op.getResult(0));
+    OpConfig internalOpConfig(internalOpOutputLayout);
 
-    // Reuse the output layout for the D2M subgraph op for all internal ops:
+    op_model::OpConstraints c(0, 0, 0, 0, internalOpOutputLayout);
+
     llvm::Expected<op_model::OpConstraints> expectedConstraints =
-        backend.getOpConstraints(internalOpInputLayouts, opConfig);
+        backend.getOpConstraints(internalOpInputLayouts, internalOpConfig);
     if (expectedConstraints) {
       c = *expectedConstraints;
     } else {
@@ -4885,11 +4891,16 @@ D2MSubgraphOp::getOpRuntime(const std::vector<TTNNLayoutAttr> &inputs,
       internalOpInputLayouts.push_back(it->second);
     }
 
-    // Default runtime for the D2M subgraph op.
-    size_t runtime = 0;
+    // Use this op's actual output layout (from the map), not the D2M's
+    // top-level layout, so the runtime API is costed for the layout this op
+    // produces.
+    assert(internalOp.getNumResults() > 0 &&
+           "OpModel op expected to have results");
+    OpConfig internalOpConfig(valueToLayout.lookup(internalOp.getResult(0)));
 
+    size_t runtime = 0;
     llvm::Expected<size_t> internalOpRuntime =
-        backend.getOpRuntime(internalOpInputLayouts, opConfig);
+        backend.getOpRuntime(internalOpInputLayouts, internalOpConfig);
     if (internalOpRuntime) {
       runtime = *internalOpRuntime;
     } else {
--- a/lib/Dialect/TTNN/Transforms/OptimizerPasses/Optimizer.cpp
+++ b/lib/Dialect/TTNN/Transforms/OptimizerPasses/Optimizer.cpp
@@ -94,6 +94,9 @@ void applyChosenLayoutToD2MSubgraphOp(D2MSubgraphOp dispatchOp,
           dispatchOp.getContext(), tensorMemoryLayoutAttr,
           BufferTypeAttr::get(dispatchOp.getContext(), bufferType),
           utils::createShardSpecIfNeeded(layoutAttr, deviceGrid)));
+    } else {
+      llvm::report_fatal_error(
+          "Expected EmptyOp for D2MSubgraphOp output buffer");
     }
   }
 


lib/Dialect/TTNN/Utils/OptimizerUtils.cpp
```diff
@@ -137,8 +137,18 @@ getUniqueTestConfigs(const std::vector<OpConfig> &consumerConfigs,
     return getUniqueTestConfigsForMatmulLinear(consumerConfigs);
   }
 
-  // For non-Matmul/Linear: only op-specific attrs matter, no output layout
-  // needed.
+  // For non-Matmul/Linear ops that still have output layout (e.g.
+  // D2MSubgraphOp), use configs as-is so validation and reference matching
+  // succeed.
+  if (!consumerConfigs.empty() && consumerConfigs.front().outputLayout) {
```
rpavlovicTT: This will break current logic. We intentionally leave layouts NULL. `consumerConfigs` is provided to pick up OpSpecificAttrs.

sgholamiTT: Reverted this change and updated the tests. For now we allow `D2M_SubgraphOp` to fall back to DRAM.
Created an [issue](https://github.com/tenstorrent/tt-mlir/issues/7025) for allowing it to be part of an L1 chain.

I described how we can possibly modify the call site and special-case `D2M_SubgraphOp` to be part of L1 chain, but that's not included in this PR as you suggested:
```c++
// We can change the call site and special-case D2MSubgraphOp. For example, in ShardSolver.cpp
llvm::SmallVector<OpConfig> testConfigs;
if (llvm::isa<ttnn::D2MSubgraphOp>(consumerOp)) {
  for (const OpConfig &c : consumerConfigs) {
    testConfigs.push_back(c);
  }
} else {
  testConfigs = optimizer_utils::getUniqueTestConfigs(
      consumerConfigs, shouldUseIgnorePhysicalLayout(consumerOp));
}
```

lib/Dialect/TTNN/Analysis/DFShardingPolicy.cpp
```diff
@@ -2298,7 +2298,8 @@ void DFShardingPolicy::run() {
                     ttnn::TypecastOp, ttnn::SiluOp, ttnn::MatmulOp,
                     ttnn::LinearOp, ttnn::MinimumOp, ttnn::GeluOp, ttnn::NegOp,
                     ttnn::RsqrtOp, ttnn::ConcatOp, ttnn::PowScalarOp,
-                    ttnn::SliceStaticOp, ttnn::RotaryEmbeddingOp>(currentOp) &&
+                    ttnn::SliceStaticOp, ttnn::RotaryEmbeddingOp,
+                    ttnn::D2MSubgraphOp>(currentOp) &&
```
odjuricicTT: You should be able to just remove D2MSubgraphOp from here and have it not sharded and resolved to dram interleaved tile by default.

Does that not work?

lib/Dialect/TTNN/Interfaces/TTNNOpModelInterface.cpp
```diff
@@ -4698,19 +4702,150 @@ AggregateTensorOp::getOpRuntime(const std::vector<TTNNLayoutAttr> &inputs,
...
+    for (mlir::Value operand : op.getOperands()) {
+      auto it = valueToLayout.find(operand);
+      assert(it != valueToLayout.end() &&
+             "D2M internal op operand must have a known layout");
+      internalOpInputLayouts.push_back(it->second);
+    }
+
+    // Reuse the output layout for the D2M subgraph op for all internal ops:
+    llvm::Expected<op_model::OpConstraints> expectedConstraints =
+        backend.getOpConstraints(internalOpInputLayouts, opConfig);
```
odjuricicTT: One concern here is that having one random ttnn op fail a constraints check for some implementation specific reason would effectively make the d2m_subgraph invalid thus probably making the whole graph / model fail compilation. Let's discuss possible solutions tomorrow on the sync.

cc: @xanderchin

lib/Dialect/TTNN/Analysis/DFShardingPolicy.cpp
sgholamiTT: This is removed.

lib/Dialect/TTNN/Interfaces/TTNNOpModelInterface.cpp
sgholamiTT: Right now, we return a default 0 opConstraint if the constraint of an internal is invalid.
As we discussed yesterday, I opened up an [issue](https://github.com/tenstorrent/tt-mlir/issues/7199) to track building a cost model for this purpose to not rely on the validity of `ttnn` ops anymore.

odjuricicTT (approved):

PR #7017, opened 2026-02-12
Negative pad decomposition

djstefanovicTT:
### Ticket
closes https://github.com/tenstorrent/tt-mlir/issues/6959

### Problem description
PyTorch supports negative padding in pad op as cropping. TTNN padOp does not allow this and causes errors.

### What's changed
Added decomposition to `ttir.slice_static` op if there is negative padding. 

### Checklist

### DIFF
--- a/include/ttmlir/Dialect/Debug/IR/DebugOps.td
+++ b/include/ttmlir/Dialect/Debug/IR/DebugOps.td
@@ -31,7 +31,7 @@ include "mlir/IR/CommonTypeConstraints.td"
 //===----------------------------------------------------------------------===//
 
 class Debug_BaseOp<string mnemonic, list<Trait> traits = []> :
-    Debug_Op<mnemonic, [Pure, SameOperandsAndResultType]> {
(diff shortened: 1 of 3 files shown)


lib/Conversion/TTIRToTTIRDecomposition/TTIRToTTIRDecomposition.cpp
```diff
@@ -2568,6 +2568,86 @@ struct SplitQueryKeyValueAndSplitHeadsDecompositionPattern
...
+      SmallVector<int64_t> slicedShape;
+      for (size_t i = 0; i < inputShape.size(); i++) {
+        slicedShape.push_back(sliceEnds[i] - sliceBegins[i]);
+      }
+      sliceResultType =
+          RankedTensorType::get(slicedShape, inputType.getElementType());
+    } else {
+      // Final output type when only slicing is needed.
+      sliceResultType = cast<RankedTensorType>(op.getType());
+    }
```
azecevic000, line 2618: IMO you don't have to optimize the case for the case when padding is not needed, you can put the calculation of the shape in the previous loop, that way it's much more obvious what the intention is.

PR #7020, opened 2026-02-12
Force replication of non-splat & non-periodic constant tensors

umalesTT:
### Ticket
Closes https://github.com/tenstorrent/tt-mlir/issues/6885

### Problem description
More details in [this comment](https://github.com/tenstorrent/tt-mlir/issues/6885#issuecomment-3861366454).
A fundamental fix in Shardy was considered, but the viable approaches reduce to case-by-case handling per consumer op. Given that, this workaround is proposed to unblock development for now.

### What's changed
Added ```ReplicateNonSplittableConstants``` pass, which forces replication of non-splat & non-periodic constant tensors. This pass must happen before ```InsertExplicitReshards``` pass. 

Expanded support in ```ShardyUtils``` for checking if sharded tensor is periodic along multiple sharding axes, given that in some training workloads such shardings, with more than one axis, appeared.

### Checklist

### DIFF
--- a/lib/Dialect/StableHLO/Transforms/UpdateGlobalToLocalShapes.cpp
+++ b/lib/Dialect/StableHLO/Transforms/UpdateGlobalToLocalShapes.cpp
@@ -64,11 +64,14 @@ static FailureOr<mlir::OperationState> createNewOperationState(
                       denseElementsAttr, newTypes[0], tensorShardings[0],
                       globalMeshOp);
 
-              assert(periodicAttr.has_value() &&
-                     "Non-splat, non-periodic constant reached "
-                     "UpdateGlobalToLocalShapes with a sharded annotation. "
-                     "ReplicateNonSplittableConstantsPass should have marked "
-                     "it as replicated.");
+              if (!periodicAttr.has_value()) {
+                constantOp.emitError(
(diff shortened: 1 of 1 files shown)


lib/Dialect/StableHLO/Transforms/UpdateGlobalToLocalShapes.cpp
```diff
@@ -46,28 +46,30 @@ static FailureOr<mlir::OperationState> createNewOperationState(
...
-                newAttr = periodicAttr.value();
-              } else {
-                constantOp->emitError("Shardy automatic parallelization "
-                                      "currently does not support "
-                                      "non-splat constant tensors, and the "
-                                      "values are not periodic "
-                                      "across the sharding dimension.");
-                return mlir::failure();
-              }
+              assert(periodicAttr.has_value() &&
```
sshonTT: @jameszianxuTT found that this `assert` is effectively no-op when release build that result to silence passing. Please change this to `emitError` or `llvm_unreachable`

umalesTT: Thanks, changed it to ```emitError```.

PR #7026, opened 2026-02-12
[D2M] enable packer L1 accumulation for matmul

vtangTT:
### Problem description
Profiling D2M matmul against TTNN [matmul](https://github.com/tenstorrent/tt-metal/blob/2b47a12f130beeeb57d1f119b121d93eed1a6e84/ttnn/cpp/ttnn/operations/matmul/device/kernels/compute/bmm_large_block_zm_fused_bias_activation.cpp#L4), it was observed that TTNN reloads partials only needs to reload partials on the last subblock, as opposed to D2M where we are reloading every subblock.

### What's changed
Need to enable packer L1 accumulation mode: `PACK((llk_pack_reconfig_l1_acc(1)))`
- added new d2m and ttkernel op
- `enable-packer-l1-acc` flag, piped into `InsertDstRegisterAccess` pass, only inserting the packer reconfig op when a matmul is present
     - don't create copy loops, but still rewrite load accesses to use DST indices.
     - always, unconditional disable l1 accum mode at the end of the compute block.. packer config state is observed to be persistent (at least in blackhole) across programs
- necessary `D2MToTTKernel` and `TTKernelToEmitc` rewriters

New perf results: https://docs.google.com/spreadsheets/d/1X8KXUc_iZhhnIOHAK-PoT8OewGLlQfnphl_gYBf3-HA/edit?gid=1297905637#gid=1297905637

Note: with L1 accumulation mode, we don't need to reload partials at all. TTNN reloaded once at the end as a safeguard against this FP32 [bug](https://github.com/tenstorrent/tt-metal/blob/main/docs/L1_ACCUMULATION_FP32_ANALYSIS.md). 

### Checklist

### DIFF
--- a/test/python/golden/test_metal_matmul.py
+++ b/test/python/golden/test_metal_matmul.py
@@ -2,7 +2,6 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-import os
 import pytest
 import torch
 from typing import List
@@ -10,13 +9,9 @@ from conftest import get_request_kwargs
 
 from test_utils import shape_str
 
-from ttmlir.ir import Context, Location, Module
-from ttmlir.passes import ttmetal_to_flatbuffer_bin
-from builder.base.builder_runtime import execute_fb
 from builder.base.builder_utils import Operand
 from builder.ttir.ttir_builder import TTIRBuilder
 from builder.base.builder_apis import compile_and_execute_ttir
-from golden import GoldenMapTensor
 
 pytestmark = pytest.mark.frontend("ttir")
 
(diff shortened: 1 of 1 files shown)


include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
```diff
@@ -884,6 +884,24 @@ def D2M_AcquireDstOp : D2M_GenericRegionComputeOp<"acquire_dst",
...
+      ```mlir
+      d2m.pack_reconfig_l1_acc(%enable) : (i32) -> ()
+      ```
+    }];
+
+    let arguments = (ins I32:$enable);
+
+    let assemblyFormat = [{ `(` $enable `)` attr-dict }];
+}
+
```
nsmithtt, line 903: I feel like this is leaking too much into the d2m dialect.  Can't we just tag the `affine.store` op with an attribute that enables accumulation?  Then this op only gets inserted during D2MToTTKernel conversion

nsmithtt: Spoke offline and agreed the way the PR is written is probably the best approach.  Let's rename this op to `d2m.set_l1_accumulate` with a description:
> Configures HW state such that all subsequent stores to L1 memory space will be accumulated to in place.

vtangTT: renamed 👍

include/ttmlir/Dialect/D2M/Transforms/Passes.td
```diff
@@ -239,6 +239,7 @@ def D2MInsertDstRegisterAccess : Pass<"d2m-insert-dst-register-access", "::mlir:
   let options = [
     Option<"useTileMatmul", "use-tile-matmul", "bool", /*default=*/"false", "Use tile_matmul instead of tile_matmul_block">,
     Option<"maxDstPhysicalSizeTiles", "max-dst-physical-size-tiles", "unsigned", "0", "Override the max dst physical size tiles or 0 if unset.">,
+    Option<"enablePackerL1Acc", "enable-packer-l1-acc", "bool", /*default=*/"false", "Enable accumulating partials in L1 instead of reloading from L1 to DST.">,
```
nsmithtt: Nit: let's also rename this to `enableL1Acc` / `enable-l1-acc`

include/ttmlir/Dialect/TTMetal/Pipelines/TTMetalPipelines.h
```diff
@@ -186,6 +186,11 @@ struct TTIRToTTMetalPipelineOptions
                      "analytical and sampling-based coalescing checks and "
                      "prints debug output comparing them."),
       llvm::cl::init(false)};
+
+  Option<bool> enablePackerL1Acc{
+      *this, "enable-packer-l1-acc",
```
nsmithtt: Same as above

lib/Dialect/D2M/Transforms/InsertDstRegisterAccess.cpp
```diff
@@ -937,7 +992,8 @@ struct D2MInsertDstRegisterAccessRewriter final
           dstAccessGenerator,
       llvm::function_ref<void(PatternRewriter &, LoadStoreRecord<LoadOrStoreTy>,
                               AffineMap, ValueRange)>
-          dstAccessRewriter) {
+          dstAccessRewriter,
+      bool rewriteOnly = false) {
```
nsmithtt: Nit: Can we rename this to enableL1Acc? I think it's easier to read that we won't emit a copy loop nest if it's enabled, rewriteOnly sounds like something to do with pattern rewriting and I was confused.

nsmithtt (approved):
Minor nits inline, but otherwise LGTM

include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
xanderchin: I think this is also leaking upwards but I'm fine with current PR.

Longer term, I think this pattern should trigger the packer reconfig insertion (init lifting should then do it's job to pull it out of the inner loops):

affine.load DST[...] --\
affine.load[ L1[...]  ->  Add -> affine.store L1 [...]

Right now we don't generate this pattern in the upstream IR. We should try to generate this pattern as it more accurately reflects the L1 accumulate operation.

include/ttmlir/Dialect/D2M/Transforms/Passes.td
vtangTT: fixed

include/ttmlir/Dialect/TTMetal/Pipelines/TTMetalPipelines.h
vtangTT: fixed

lib/Dialect/D2M/Transforms/InsertDstRegisterAccess.cpp
vtangTT: fixed

vtangTT:
https://github.com/tenstorrent/tt-mlir/issues/7147 follow up issue for bfp8

PR #7037, opened 2026-02-13
TTIR -> Linalg patterns (part 4)

dmilinkovicTT:
### Ticket
#5809 

### What's changed
- Added `TTIR` -> `Linalg` pattern for `argmax` op (which previously gets decomposed to single-dim version in `TTIRToTTIRDecomposition`)
- Added `TTIR` -> `Linalg` pattern for `embedding` op and removed `embedding` -> `gather` TTIRToTTIR decomposition, since `gather` op will soon be removed (cc: @sdjukicTT)
- Removed creating splat values where a more performant approach is possible (e.g. using const scalar tensor and relying on implicit broadcasting), in order to avoid huge dylibs ([example model failure](https://github.com/tenstorrent/tt-xla/actions/runs/21949504118/job/63396112660))

### Checklist

### DIFF
--- a/lib/Conversion/TTIRToLinalg/TTIRToLinalg.cpp
+++ b/lib/Conversion/TTIRToLinalg/TTIRToLinalg.cpp
@@ -51,12 +51,12 @@ static Value convertToBooleanTensor(Value input, Location loc,
   auto elementType = inputType.getElementType();
   assert(elementType.isF32());
 
-  // Create broadcastable zero constant.
-  SmallVector<int64_t> broadcastShape(inputType.getRank(), 1);
-  auto broadcastType = RankedTensorType::get(broadcastShape, elementType);
+  // Create zero constant.
+  SmallVector<int64_t> zeroShape(inputType.getRank(), 1);
+  auto zeroType = RankedTensorType::get(zeroShape, elementType);
   DenseElementsAttr zeroAttr =
-      DenseElementsAttr::get(broadcastType, rewriter.getF32FloatAttr(0.0f));
-  auto zeroConst = rewriter.create<tosa::ConstOp>(loc, broadcastType, zeroAttr);
+      DenseElementsAttr::get(zeroType, rewriter.getF32FloatAttr(0.0f));
+  auto zeroConst = rewriter.create<tosa::ConstOp>(loc, zeroType, zeroAttr);
 
   // For logical operations, non-zero means true.
   // So we need: (input != 0) which we get by computing !(input == 0).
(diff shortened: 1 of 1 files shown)


lib/Conversion/TTIRToLinalg/TTIRToLinalg.cpp
```diff
@@ -755,22 +711,22 @@ class CosOpConversionPattern : public OpConversionPattern<ttir::CosOp> {
         this->getTypeConverter()->convertType(op.getResult().getType()));
     assert(resultType && "Result type must be a ranked tensor type.");
 
-    // Create a scalar constant for π/2 using arith.constant
+    // Create a scalar constant for π/2 using arith.constant.
```
vwellsTT: nit: maybe this should be moved below now?

lib/Conversion/TTIRToLinalg/TTIRToLinalg.cpp
```diff
@@ -2068,6 +2107,147 @@ class ProdOpConversionPattern : public OpConversionPattern<ttir::ProdOp> {
...
+
+    auto resultType = dyn_cast<RankedTensorType>(
+        this->getTypeConverter()->convertType(op.getResult().getType()));
+    assert(resultType && "Result type must be a ranked tensor type.");
+
+    // After TTIRToTTIRDecomposition, dim_arg is either absent (reduce all
+    // dims) or has exactly 1 entry.
+    auto dimArg = op.getDimArg();
+    assert((!dimArg || dimArg->size() <= 1) &&
+           "Multi-dim argmax should have been decomposed");
```
vwellsTT: nit: I guess assertion error should end with . (or at least above one does--not sure if we have a super consistent philosophy for this)

lib/Conversion/TTIRToTTIRDecomposition/TTIRToTTIRDecomposition.cpp
```diff
@@ -678,98 +678,6 @@ struct GatherToEmbeddingConversionPattern
...
 
-//===----------------------------------------------------------------------===//
-// Embedding to Gather Pattern
-//===----------------------------------------------------------------------===//
-
-namespace {
-// Converts ttir.embedding to ttir.gather for CPU fallback path.
-// This allows us not to use tensor.gather which lacks bufferization support in
-// upstream MLIR.
-struct EmbeddingToGatherConversionPattern
```
vwellsTT, line 689: hmm, is completely removing this desirable (vs remove for cpu fallback or something)?

vwellsTT: I guess from PR description gather op is going to removed, okay.  Fine with me then 😄

Issue #3695, opened 2025-06-06
Model trace host fallback in compiler


jnie-TT:
### Describe the bug

Blocked by #3606 

With the current implementation, there is a runtime workaround in trace that falls back to host implicitly when copying tensors to device. We would like to remove this workaround and model this fallback in the compiler.

odjuricicTT:
I think that the solution is to not put the inputs onto the device by default before executing trace.

odjuricicTT:
@jnie-TT I've tried running resnet with trace trough tt-xla and it works! :)
The problem is that the e2e perf is worse than without trace, the most probable reason for this is this WA and the fact that we send the input tensor to device two times.

I think that the proper solution is to keep the input tesor in system memory if the enable_trace flag is set. Does this sound ok to you? Could you take this on?

jnie-TT:
@odjuricicTT it's weird that it's slower, how many iterations did you run it for? Could you try using the same input tensor for each iteration to make sure that when the inputs/weights are the same we don't see a delay (copy back to host won't happen here because the input tensor versions would be the same, similar to consteval)?

And yes I think for trace inputs should just be on host, but the problem is that alot of input tensors to trace are outputs of consteval, not direct inputs from the user.

odjuricicTT:
The current problem is that with trace enabled all input tensors are pre allocated. Copying a ttnn tensor to a pre-allocated location in dram is only supported from host memory. The compiler defaults to having all inputs in dram so what happens when trace is enabled is that we put all tensors in dram but then pull them back to host in order to then memcopy to the preallocated addresses. This is slow and we should model this behaviour in the compiler better.

One idea is to keep input tensors on host if trace is enabled and then only to the one memcopy to preallocated addresses on model forward.

PR #7039, opened 2026-02-13
Move Trace Inputs to Host Memory

sdjordjevicTT:
### Ticket
Closes https://github.com/tenstorrent/tt-mlir/issues/3695

### Problem description
Previously, trace inputs were handled inefficiently:
- Unnecessary transfers: All inputs were assumed to be in device memory, requiring a redundant device-to-host transfer (from_device) before copying them to trace input slots
- Workaround dependency: The implementation relied on a workaround (traceImplicitFromDevice) because TTNN only supported writing to pre-allocated device tensors from the host

### What's changed
This PR optimizes trace execution by moving trace inputs to host memory instead of device memory, improving memory management and reducing unnecessary device-to-host transfers during trace capture:
- Modified trace input handling: Trace inputs are now kept in system memory (host) rather than being moved to device memory
- Updated CaptureOrExecuteTraceOp: Added constant_input_mask field to distinguish between constant/parameter inputs (which remain on device) and regular inputs (which stay on host)
- Enhanced verification: Added comprehensive type checking to ensure input/output types match between capture and trace functions
- Optimized memory transfers: Eliminated redundant device-to-host transfers by directly copying host inputs to device trace slots
- Removed workaround: Deprecated the traceImplicitFromDevice workaround as it's no longer needed
- Updated Python tests to reflect the new behavior

### Checklist


lib/Dialect/TTNN/Transforms/TTNNTraceHoistTransform.cpp
```diff
@@ -505,6 +532,121 @@ class TTNNTraceHoistTransform
...
+      }
+
+      // Update mesh_shard operations' result types
+      for (auto &[meshShardOp, newResultType] : meshShardsToUpdate) {
+        meshShardOp.getResult().setType(
+            mlir::cast<RankedTensorType>(newResultType));
+      }
+
+      // Erase the ToLayoutOps
+      for (auto op : opsToErase) {
```
vmilosevic: :warning: **llvm\-qualified\-auto** :warning:
`` auto op `` can be declared as `` auto *op ``
```suggestion
      for (auto *op : opsToErase) {
```

include/ttmlir/Dialect/TTCore/IR/Utils.h
```diff
@@ -164,6 +164,40 @@ bufferization::BufferLikeType
...
+inline bool isConstantOrParameterArgumentType(func::FuncOp op,
+                                              size_t argIndex) {
+  ArgumentType argType = getFunctionArgumentType(op, argIndex);
+  return argType == ArgumentType::Constant ||
+         argType == ArgumentType::Parameter;
+}
+
+// Checks if the function argument has the ttcore.kv_cache attribute attached.
+inline bool isKVCacheArgument(func::FuncOp op, size_t argIndex) {
+  return op.getArgAttr(argIndex, ttcore::g_kvCacheAttrName) != nullptr;
```
azecevicTT: Here as well.

lib/Dialect/TTNN/Transforms/TTNNTraceHoistTransform.cpp
```diff
@@ -664,7 +916,17 @@ class TTNNTraceHoistTransform
...
 
-    return performHoistTransform(funcOp, opsToHoist);
+    // Perform the hoist transform
+    mlir::LogicalResult result = performHoistTransform(funcOp, opsToHoist);
+    if (failed(result)) {
+      return result;
+    }
+
+    if (failed(mergeToLayoutOpsWithFuncArgs(funcOp))) {
+      return result;
```
mtopalovicTT: return failure here and above?

runtime/test/ttnn/python/utils.py
```diff
@@ -75,26 +81,80 @@ def __init__(self, config: ProgramTestConfig, binary: Binary, program_index: int
...
+                )
+                rt_input_with_layout = ttrt.runtime.to_layout(
+                    rt_input, device, layout, True
+                )
+                inputs_runtime_with_layout.append(rt_input_with_layout)
+
+                # Cache constant inputs (both torch and runtime tensors)
+                if i in self.constant_input_indices:
+                    self.constant_inputs_torch_cache[i] = torch_input
+                    self.constant_inputs_runtime_cache[i] = rt_input_with_layout
```
pilkicTT: i would like us to avoid this complication because of one test... can't we just do this one-off manually in the test - it shouldn't be too bad?

one more reason why i would like to avoid this, is because i will refactor these tests soon...

include/ttmlir/Dialect/TTCore/IR/Utils.h
```diff
@@ -174,6 +174,40 @@ bufferization::BufferLikeType
...
+  return ArgumentType::Input;
+}
+
+// Checks if the function argument is of type Input.
+inline bool isInputArgumentType(func::FuncOp op, size_t argIndex) {
+  return getFunctionArgumentType(op, argIndex) == ArgumentType::Input;
+}
+
+// Checks if the function argument is of type Constant or Parameter.
+inline bool isConstantOrParameterArgumentType(func::FuncOp op,
```
mtopalovicTT, line 199: can you update call sites of `isConstOrParamArg` and replace them with your helper since it's pretty much same thing

lib/Dialect/TTNN/Pipelines/TTNNPipelines.cpp
```diff
@@ -362,10 +362,11 @@ void createTTIRToTTNNDevicePipeline(
         devicePm.addPass(mlir::createCanonicalizerPass());
       }
     }
-    createTTNNPipelineLayoutDecompositionPass(devicePm, options);
     if (options.enableTrace) {
       devicePm.addPass(tt::ttnn::createTTNNTraceHoistTransform());
     }
+    createTTNNPipelineLayoutDecompositionPass(devicePm, options);
```
mtopalovicTT, line 372: Can you add comment why this needs to be after.

runtime/lib/ttnn/operations/trace/capture_or_execute_trace.cpp
```diff
@@ -142,16 +137,37 @@ static void executeTrace(const ::tt::target::ttnn::CaptureOrExecuteTraceOp *op,
...
+                 inputSlotWrapper.getVersion());
+      LOG_ASSERT(inputTensorWrapper.getTensor().storage_type() ==
+                     ::ttnn::StorageType::DEVICE &&
+                 "Non-regular inputs must already be on device.");
+      LOG_DEBUG("Skipping copy for constant input ", i,
+                " since it is already on device and trace input slot is "
+                "persisted across traces. Version: ",
+                inputTensorWrapper.getVersion());
+      continue;
+    }
```
pilkicTT: i am not quite sure if we should assert this... 

we could allow for weights to update and the mechanism for updating the slots should work. on the other hand, in case of multiple traced graphs, we don't have this tested and currently, we'll probably hit some issues.

anyhow, we should update the comment to reflect this - because there is nothing invalid in user updating the weight even if he's using trace, it's just that our implementation currently is making that risky.

let's sync offline...

runtime/test/ttnn/python/utils.py
sdjordjevicTT: As discussed offline, reverted these changes

runtime/lib/ttnn/operations/trace/capture_or_execute_trace.cpp
sdjordjevicTT: @pilkicTT, please take a look at the latest iteration, whether it is ok :)

lib/Dialect/TTNN/Transforms/TTNNTraceHoistTransform.cpp
```diff
@@ -664,7 +949,17 @@ class TTNNTraceHoistTransform
...
```
jserbedzijaTT: We should return mlir::failure() here

lib/Dialect/TTNN/Transforms/TTNNTraceHoistTransform.cpp
```diff
@@ -311,31 +299,80 @@ class TTNNTraceHoistTransform
...
+      // capture. We create two types for each regular input:
+      // 1. Host-side type (system memory) for the function signature
+      // 2. Device-side slot type (DRAM) for persistent trace storage
+      RankedTensorType inputArgType = utils::RankedTensorTypeFactory::create(
+          originalRankedTensorType, BufferType::SystemMemory);
+
+      inputTypes.push_back(inputArgType);
+
+      // Create the new output result type with the updated buffer type for the
+      // trace input slot for this argument on host.
```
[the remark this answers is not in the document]
jserbedzijaTT: Double check this

lib/Dialect/TTNN/Transforms/TTNNTraceHoistTransform.cpp
```diff
@@ -2,6 +2,7 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 
+#include "mlir/IR/BuiltinTypes.h"
 #include "ttmlir/Dialect/TTCore/IR/Utils.h"
```
vmilosevic, line 7: :warning: **llvm\-include\-order** :warning:
```suggestion
#include "mlir/Support/LLVM.h"
#include "ttmlir/Dialect/TTCore/IR/Utils.h"
```

lib/Dialect/TTNN/Transforms/TTNNTraceHoistTransform.cpp
```diff
@@ -13,6 +14,7 @@
 #include "ttmlir/FunctionTypes.h"
 #include "ttmlir/Utils.h"
 #include <atomic>
+#include "mlir/Support/LLVM.h"
```
vmilosevic: :warning: **llvm\-include\-order** :warning:
```suggestion

```

PR #7045, opened 2026-02-13
Pre-allocator generic affine fusion

bgrady-tt:
**Note**: _This PR does not enable affine fusion or scalrep passes by default_. Currently only a few lit tests exercise these passes. This is due to lack of support in downstream passes to handle intermediate scratchpad allocations properly.

Big New Passes:
- **GenericAffineLoopFusion** : Iteratively fuses producer/consumer generic op pairs if correct until no more successful fusions are possible. This pass doesn't eliminate intermediate CBs/operands; it just focuses on correct loop fusion.
- **GenericAffineScalarReplacement** : Replaces intermediate operands and CBs (only used to pass intermediate results in a single generic op) and eliminates (forwards) matching store->load pairs using affine scalrep utilities. Uses `d2m.scratch_allocate` for allocating all intermediates (cannot be lowered as-is).

This PRs impact to default pipeline is:
- Moves outer loop generation even earlier in the pipeline (immediately post-bufferization)
- Introduces a new form of the d2m.generic that has affine blocking loops with symbolic bounds (`get_block_factor()`). 
- A new pass `LowerToExplicitForm` converts this affine form with unresolved blocking factors to the fully explicit generic form that has hardened block factors and scf.for loops. This conversion happens immediately post-allocator

### DIFF
--- a/include/ttmlir/Dialect/D2M/IR/D2MOps.td
+++ b/include/ttmlir/Dialect/D2M/IR/D2MOps.td
@@ -318,9 +318,8 @@ def D2M_GenericOp
         }
         return true;
       }
-      bool isAffineMapForm() { return !getIndexingMaps().empty(); }
+      bool isImplicitBlockedForm();
       bool isAffineBlockedForm();
-      bool isLoweredLoopForm() { return !isAffineMapForm(); }
       bool isExplicitDatamovementForm() {
         return getBlockFactors().empty() &&
                getIndexingMaps().empty() &&
--- a/lib/Dialect/D2M/IR/D2MOps.cpp
+++ b/lib/Dialect/D2M/IR/D2MOps.cpp
@@ -1929,15 +1929,39 @@ mlir::SmallVector<int64_t> d2m::GenericOp::getBlockFactorsValue() {
   });
 }
 
-bool d2m::GenericOp::isAffineBlockedForm() {
-  // block_factors, indexing_maps, and iterator_types must all be non-empty.
-  if (getBlockFactors().empty() || getIndexingMaps().empty() ||
-      getIteratorTypes().empty()) {
+/// Returns true if the generic op has non-empty block_factors, indexing_maps,
+/// and iterator_types attributes, and a single unified region.
+static bool hasBlockingAttributes(d2m::GenericOp genericOp) {
+  if (genericOp.getBlockFactors().empty() ||
+      genericOp.getIndexingMaps().empty() ||
+      genericOp.getIteratorTypes().empty()) {
     return false;
   }
 
-  // Must have a single region with Unified thread type.
-  if (getNumRegions() != 1 || getRegionThreadType(0) != ThreadType::Unified) {
+  return genericOp.getNumRegions() == 1 &&
+         genericOp.getRegionThreadType(0) == ThreadType::Unified;
+}
+
+bool d2m::GenericOp::isImplicitBlockedForm() {
+  if (!hasBlockingAttributes(*this)) {
+    return false;
+  }
+
+  // No affine blocking loops must be present.
+  bool hasBlockingLoop = false;
+  getRegion(0).walk([&](affine::AffineForOp forOp) {
+    if (forOp->hasAttr("d2m.blocking_loop")) {
+      hasBlockingLoop = true;
+      return WalkResult::interrupt();
+    }
+    return WalkResult::advance();
+  });
+
+  return !hasBlockingLoop;
+}
+
+bool d2m::GenericOp::isAffineBlockedForm() {
+  if (!hasBlockingAttributes(*this)) {
     return false;
   }
 


lib/Dialect/TTNN/Pipelines/TTNNPipelines.cpp
```diff
@@ -510,6 +510,7 @@ void createTTNNPipelineD2MPass(OpPassManager &pm) {
   // only works on top-level modules (doesn't run module has a parent op).
   ttmetal::TTIRToTTMetalPipelineOptions ttmetalOptions;
   ttmetalOptions.ttnnMode = true;
+  ttmetalOptions.enableAffineLoopFusionAndScalarReplacement = false;
```
bgrady-tt, line 513: Force affine fusion disabled until feature is more mature.

lib/Dialect/D2M/Transforms/ScheduleDMA.cpp
```diff
@@ -43,12 +43,12 @@ collectDMAOps(Block *block,
     if (auto remoteLoad = mlir::dyn_cast<RemoteLoadOp>(&op)) {
       // Get the CB operand and find which block argument it corresponds to.
       Value cb = remoteLoad.getCb();
-      if (auto blockArg = mlir::dyn_cast<BlockArgument>(cb)) {
+      if (auto blockArg = mlir::dyn_cast_or_null<BlockArgument>(cb)) {
```
bgrady-tt, line 46: Just minor cleanup to unsafe dyn_cast here

include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
```diff
@@ -644,6 +645,14 @@ def D2M_TileBcastOp : D2M_GenericRegionComputeOp<"tile_bcast", [Pure, D2M_SkipOp
...
 }
 
+def D2M_PackerMaskResetOp : D2M_GenericRegionComputeOp<"packer_mask_reset",
+  [D2M_SkipOpAffineLoopFusionTrait]> {
+  let summary = "D2M Packer Mask Reset Op";
+  let description = [{
+        The `packer_mask_reset` operation resets the packer mask. This should be removed once LLK refactors the reduce API.
+        See tt-metal#22904 and tt-metal -> reduce.h.
+    }];
+}
```
nsmithtt: This probably came in as a bad rebase

bgrady-tt: Yep, currently have this fixed on my local branch 😓 Doing some cleanup today before asking for reviews.

include/ttmlir/Dialect/TTMetal/Pipelines/TTMetalPipelines.h
```diff
@@ -186,6 +186,14 @@ struct TTIRToTTMetalPipelineOptions
                      "analytical and sampling-based coalescing checks and "
                      "prints debug output comparing them."),
       llvm::cl::init(false)};
+
+  // Option to enable affine loop fusion and scalar replacement passes in the
+  // middleend.
+  Option<bool> enableAffineLoopFusionAndScalarReplacement{
```
bgrady-tt, line 192: **NOTE**: The fusion and scratchpad allocation 'backend' cannot currently handle the output of these passes. This is being commit in stages to solidify other related refactors for a future PR to finally integrate into end-to-end fusion with intermediates.

include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
```diff
@@ -1979,6 +2066,22 @@ def D2M_BlockIndexOp : D2M_IndexOp<"block_index"> {
     }];
 }
 
+def D2M_BlockOffsetOp : D2M_IndexOp<"block_offset"> {
+    let summary = "Block Offset op.";
+    let description = [{
+      Represents block_factor * core_index as a constant-like operation. Useful
+      for representing blocking index math in an affine dialect compatible way.
+    }];
+}
```
nsmithtt, line 2067: Oh did the affine symbol approach not work?

lib/Dialect/D2M/IR/D2MGenericRegionOps.cpp
```diff
@@ -1243,7 +1250,9 @@ void IterIndexOp::inferResultRanges(
 }
 
 mlir::OpFoldResult IterIndexOp::fold(FoldAdaptor adaptor) {
-  return adaptor.getDimAttr();
+  // Cannot fold: runtime value depends on loop iteration, not the dimension
+  // index itself.
```
nsmithtt, line 1254: This is only true once the generic is in explicit loop form and then the iter_index op should probably be made illegal.  Perhaps we should revert this and update the verifier for this op to check which form the generic op is in and fail if it's in explicit loop form.

lib/Dialect/D2M/IR/D2MGenericRegionOps.cpp
```diff
@@ -1267,6 +1276,50 @@ mlir::OpFoldResult BlockIndexOp::fold(FoldAdaptor adaptor) {
...
+    mlir::SetIntRangeFn setResultRange) {
+  setResultRange(getResult(),
+                 getIndexRange(0, std::numeric_limits<uint32_t>::max()));
+}
+
+mlir::OpFoldResult BlockOffsetOp::fold(FoldAdaptor adaptor) {
+  // Block offset must remain explicit so later D2M lowering can preserve
+  // block/core index math structure.
+  return {};
+}
```
nsmithtt, line 1300: Can we remove the BlockOffsetOp?  Seems like it might no longer be referenced

nsmithtt: Ah nevermind, I assume this means that that affine symbol approach which seemed to support multiply didn't work?

lib/Dialect/D2M/Transforms/GenericAffineScalarReplacement.cpp
```diff
@@ -0,0 +1,440 @@
...
+}
+
+class D2MGenericAffineScalarReplacement
+    : public impl::D2MGenericAffineScalarReplacementBase<
+          D2MGenericAffineScalarReplacement> {
+public:
+  using D2MGenericAffineScalarReplacementBase::
+      D2MGenericAffineScalarReplacementBase;
+
+  void runOnOperation() final {
```
nsmithtt, line 402: Did you dig into why `affine::createAffineScalarReplacementPass` isn't sufficient? It should automatically run the analysis that we're manually deriving here.

include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
bgrady-tt: This is part of that approach; the constant offset (block_factor * core_index) is represented as a single op to avoid doing `constant * other_constant` in the affine map.

lib/Dialect/D2M/IR/D2MGenericRegionOps.cpp
bgrady-tt: BlockOffsetOp is definitely used in d2m-generate-outer-loops and throughout the fusion passes?

lib/Dialect/D2M/Transforms/GenericAffineScalarReplacement.cpp
bgrady-tt: There is a need to do at least three major transforms to get scalar replacement to work as-is:
1. Analyze all operands to find intermediates, and change them to be allocated internally within the GenericOp (`internalizeFusedGenericIntermediates`). This is necessary for stores to be eliminated.
2. Temporarily convert RemoteLoad to _direct style_ (load result is used by compute instead of buffer) to match expectations of affine-scalrep and (`convertRemoteLoadToDirectStyle`). Scalar replacement simply doesn't work otherwise.
3. `convertBlockOffsetsToTaggedConstants/restoreTaggedConstantsToBlockOffsets` - This 'compatibility mode' is unfortunately still necessary even with BlockOffsetOp; making this op `ConstantLike` is _not_ a good idea and caused downstream bugs with canonicalization.

bgrady-tt: Some of the above bridging puts IR in an invalid form, so it is better to keep that state ephemeral. Fwiw calling affineScalarReplace on a func by func level is basically all affine::createAffineScalarReplacementPass does.

lib/Dialect/D2M/IR/D2MGenericRegionOps.cpp
bgrady-tt: This results sometimes in canonicalization folding `IterIndex` op into `arith.constant` values. This seems incorrect? The 'return-an-attribute' pattern for ::fold implementations is meant for constant-like ops, but IterIndex op represents loop IVs?

PR #7055, opened 2026-02-15
[TTNN] Improve RoPE fusing pattern matching

jserbedzijaTT:
### Ticket
Closes https://github.com/tenstorrent/tt-mlir/issues/7135

### Problem description
RoPE (Rotary Position Embedding) was not being fused in some models because the `EraseInverseOps` pass commutes permutes through the graph, breaking the patterns that the previous fusing logic expected.

The RoPE computation follows the pattern:
```
output = (x * cos) + (rotate_half(x) * sin)
```
where `rotate_half(x) = concat(neg(slice(x, half:)), slice(x, :half))`.

### What's changed
- Refactored the RoPE structural matcher to handle more IR variations (commuted operands, TM chains, broadcasts)
- Added a semantic axis analyzer that propagates symbolic {B,H,S,D} labels through the subgraph to validate candidate inputs. This decouples fusion from any specific axis ordering, regardless of how permutes rearrange the tensor, the analyzer resolves which physical dimensions correspond to B, H, S, and D, making the pattern robust to upstream pass changes.
- When permutes remain in the op chain, a permute is applied to the output of the fused `RotaryEmbeddingOp` to restore the original axis order
- Added op model validation so fusion gracefully falls back when the fused op violates hardware constraints
- Added unit tests and model-extracted tests for RoPE fusing

### Checklist

### DIFF
--- a/lib/Dialect/TTNN/Transforms/CMakeLists.txt
+++ b/lib/Dialect/TTNN/Transforms/CMakeLists.txt
@@ -1,5 +1,6 @@
 if (TTMLIR_ENABLE_OPMODEL)
         add_definitions(-DTTMLIR_ENABLE_OPMODEL)
+        set(OPMODEL_SRCS Fusing/RoPEFusingPattern.cpp)
 endif()
 
 add_mlir_dialect_library(MLIRTTNNTransforms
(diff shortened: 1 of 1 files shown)


lib/Dialect/TTNN/Transforms/TTNNFusing.cpp
```diff
@@ -174,267 +174,745 @@ class TTNNMatmulAndLinearWithActivation : public mlir::OpRewritePattern<SrcOp> {
...
-//
-// This pattern is broken down into two sub-patterns:
-// 1. unrotatedProjection: matches x * cos
-// 2. rotatedProjection: matches rotate_half(x) * sin
-//
-// unrotatedProjection comes as following sequence of operations:
-// cos_unsqueezed = unsqueeze(cos) - where unsqueeze will be reshape which
-// prepends dimensions to cos to match x's rank.
-// unrotated_projection = multiply(x, cos_unsqueezed)
+namespace {
```
sdjordjevicTT: There are a lot of new lines here related to Rope fuisng. Should we create a separate header/cpp for this logic?

PR #7078, opened 2026-02-18
stablehlo to ttir conversion for scatter with 32-bit multidimensional indices

anuragsingh-tt:
...and added mlir test case to check this conversion.

### DIFF
--- a/lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
+++ b/lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
@@ -35,6 +35,7 @@
 #include "mlir/Transforms/DialectConversion.h"
 #include "stablehlo/dialect/StablehloOps.h"
 #include "llvm/ADT/APFloat.h"
+#include "llvm/ADT/APInt.h"
 #include "llvm/ADT/STLExtras.h"
 
 #include "llvm/ADT/SmallVector.h"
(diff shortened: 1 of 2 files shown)


acicovicTT:
Can you please update the PR title / description to highlight the fact that this PR adds support for 32-bit scatter indices specifically

lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
acicovicTT, line 5151: For my understanding, why was this change needed?

lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
```diff
@@ -4874,6 +4875,17 @@ class StableHLOToTTIRScatterOpConversionPattern
     // Get index tensor shape.
     RankedTensorType indexType = op.getScatterIndices().getType();
     ArrayRef<int64_t> indexShape = indexType.getShape();
+    auto indexElementType = dyn_cast<IntegerType>(indexType.getElementType());
+    if (!indexElementType) {
+      return rewriter.notifyMatchFailure(
+          op, "TTIR scatter requires integer scatter_indices element type.");
+    }
+    if (indexElementType.getWidth() != 32 &&
```
mmanzoorTT, line 4883: Question: Why not to upcast the `scatter_indices` instead of failing?

Issue #6826, opened 2026-01-30
Change split axis in decomposed All Reduce


mvasiljevicTT:
Compiled llama 70b splits tensors for reduce_scatter over batch dimension:

<img width="600" alt="Image" src="https://github.com/user-attachments/assets/413ade64-0b76-4949-88ac-4bafb09008f5" />

while in metal implementation it is done on last dimension, which should be faster.

That `reduce_scatter` op comes from decomposing `all_reduce` op that doesn't specify splitting axis, and is chosen in `TTNNAllReduceWorkarounds` to be "_first tensor dimension which is divisible by number of devices_". 

Metal implementation of all reduce by default uses _last divisible dimension_.

mvasiljevicTT:
One layer llama 70b in local goes from 15.6sps to 19.05 with this change.

CI: https://github.com/tenstorrent/tt-xla/actions/runs/21907458939

mvasiljevicTT:
https://github.com/tenstorrent/tt-xla/actions/runs/21993732844

mvasiljevicTT:
Merged.

PR #7083, opened 2026-02-18
Change splitting dim in AllReduce workaround 

mvasiljevicTT:
### Ticket
[Link to Github Issue](https://github.com/tenstorrent/tt-mlir/issues/6826)

### Problem description
When `all_reduce` is decomposed in `TTNNAllReduceWorkarounds`,  splitting axis for `reduce_scatter` is set to "_first tensor dimension which is divisible by number of devices_".

This should be changed to _last divisible dimension_, as is in tt-metal.

### What's changed
Changed logic in `TTNNAllReduceWorkarounds` to find last divisible dimension instead of the first, which showed perf boost on all models.

### Checklist
- [x] [Perf benchmark llmbox](https://github.com/tenstorrent/tt-xla/actions/runs/21993732844)

### DIFF
--- a/test/ttmlir/Dialect/TTNN/ccl/all_reduce/all_reduce_positive.mlir
+++ b/test/ttmlir/Dialect/TTNN/ccl/all_reduce/all_reduce_positive.mlir
@@ -94,3 +94,34 @@ module attributes {} {
     return %1 : tensor<1x38x128x515xf32>
   }
 }
+
+// -----
+
+// Verify default all_reduce dimensions lower to scatter/all_gather on the last
+// divisible dimension (dim 3 for this shape).
+module attributes {} {
+  // CHECK-LABEL: all_reduce_default_dims_last_divisible_dim3
+  func.func @all_reduce_default_dims_last_divisible_dim3(%arg0: tensor<32x256xbf16>) -> tensor<32x256xbf16> {
+    %1 = "ttir.all_reduce"(%arg0) <{cluster_axis = 1 : ui32, reduce_type = #ttcore.reduce_type<sum>}> : (tensor<32x256xbf16>) -> tensor<32x256xbf16>
+    // CHECK: "ttnn.reduce_scatter"
+    // CHECK-SAME: scatter_dim = 3 : si32
+    // CHECK: "ttnn.all_gather"
+    // CHECK-SAME: all_gather_dim = 3 : si32
+    return %1 : tensor<32x256xbf16>
+  }
+}
+
+// -----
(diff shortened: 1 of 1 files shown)


lib/Dialect/TTNN/Transforms/Workarounds/TTNNWorkaroundsPatterns.cpp
```diff
@@ -383,13 +383,19 @@ class TTNNAllReduceWorkarounds : public OpRewritePattern<ttnn::AllReduceOp> {
...
     // cluster axis on which we are performing the all reduce.
     auto sizeOfDevices = meshShape[clusterAxis];
     auto inputShape = inputType.getShape();
-    const auto *tensorDimDevice = llvm::find_if(
-        inputShape, [&](int64_t dim) { return dim % sizeOfDevices == 0; });
+    auto reversedInputShape = llvm::reverse(inputShape);
+    auto tensorDimRevIt =
+        llvm::find_if(reversedInputShape, [sizeOfDevices](int64_t dim) {
+          return dim % sizeOfDevices == 0;
+        });
```
svuckovicTT, line 395: Would this work?

```suggestion
    auto tensorDimRevIt = llvm::find_if(inputShape.rbegin(), inputShape.rend(),
        [sizeOfDevices](int64_t dim) {
          return dim % sizeOfDevices == 0;
        });
```

azecevic000: `rbegin/rend` is more of a trick from the old days; even STL leans heavily on the ranges these days, and llvm functions follow that design

mvasiljevicTT: @svuckovicTT  thanks for the suggestion!

Unlike `std::find_if`, `llvm::find_if` [only supports ranges](https://llvm.org/doxygen/namespacellvm.html#ac78c09ae232b2ce188ff590d51e3c268), and can't be used with a pair of iterators directly. I could make a range from them, but that would go back to `llvm::reverse`. (And I like using `llvm` instead of `std` whenever I can)

svuckovicTT: @azecevicTT @mvasiljevicTT Ah, I assumed `llvm::reverse` would materialize a new array (similar to `std::reverse`), thanks for pointing out the difference :)

PR #7090, opened 2026-02-18
[D2M] Split semaphore_wait properly during DMA lowering

bgrady-tt:
Semaphore waits are replicated to all split regions in `SplitUnifiedThread` and `ScheduleDMA` passes.

Semaphore wait _with optional reset value_, semaphore set, and semaphore inc are all illegal in Unified generic op form (for now). It isn't clear there is a need for more semaphore operations, nor how they should be implemented.

### DIFF
--- a/lib/Dialect/D2M/Transforms/SplitUnifiedThread.cpp
+++ b/lib/Dialect/D2M/Transforms/SplitUnifiedThread.cpp
@@ -6,6 +6,7 @@
 
 #include "ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.h"
 #include "ttmlir/Dialect/D2M/IR/D2MOps.h"
+#include "ttmlir/Dialect/D2M/Utils/DMAUtils.h"
 
 #include "mlir/Dialect/SCF/IR/SCF.h"
 #include "mlir/IR/IRMapping.h"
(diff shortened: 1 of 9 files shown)


lib/Dialect/D2M/Transforms/SplitUnifiedThread.cpp
```diff
@@ -51,6 +53,13 @@ class D2MSplitUnifiedThreadRewriter : public OpRewritePattern<GenericOp> {
     }
     Block *originalBlock = &originalRegion.front();
 
+    // Check that there are no illegal semaphore ops in the unified thread.
+    // Replicating these across two threads would create a race condition on the
+    // shared semaphore.
+    if (failed(utils::checkForIllegalSemaphoreOps(originalBlock))) {
```
nsmithtt, line 59: Won't block the PR on it, but I think it might be slightly cleaner to do this with an op interface or traits like `D2M_GenericRegionComputeOpTrait` / `D2M_GenericRegionDatamovementOpTrait`

bgrady-tt: Something like `D2M_SplitFormOpTrait` that is disallowed in unified form?

bgrady-tt: I think this approach won't work for `semaphore_wait` with reset, as some forms of the op are legal and some are not. So just going to leave it as-is.

nsmithtt: It could be an interface that checks dynamic state on the op, maybe a trait oversimplified it.  It's just much more maintainable to use interfaces / traits where possible because it'll enforce the IR is always legal and not just in a particular pass.

PR #7080, opened 2026-02-18
d2m: broadcast non tiled dim

anuragsingh-tt:
### Ticket
Resolves #7064 

### Problem description
isImplicitBcast was true only if tileBcastTypes had row/col/scalar broadcast. That only captures broadcast in tile-local dims, not broadcast on earlier tensor dims.

Also added builder test.

PR #7093, opened 2026-02-18
fix broadcast on tile row/col dimensions

anuragsingh-tt:
### Problem description
#7080 added broadcasts on non-tiled dims but tiled dims (row/col) were failing. 

### What's changed
Copy broadcast operand into Dst slot for op output using CopyDestValuesOp so binary_dest_reuse_tiles works properly (i.e. does not read wrong Dst tile). 
Also added MLIR test + builder tests + appropriate header.


ckaravasilisTT:
LGTM

nsmithtt (approved):

PR #7094, opened 2026-02-18
[D2M] Refactor GenericOp Forms

bgrady-tt:
Heavily refactors d2m middle-end pipeline and solidifies four GenericOp forms:
1. **Implicit Blocked Form** - Has blocking factors, indexing maps, and iterator types. Blocking loops are _implicit_.
  a. Loop interchange is flexible
  a. Used throughout d2m frontend
  b. Converted to affine blocked form by `GenerateOuterLoops` pass post-bufferization
2. NEW **Affine Blocked Form** - Has blocking factors, indexing maps, and iterator types. Blocking loops are explicit **`affine.for`**. 
  a. Loop interchange is fixed.
  b. Exists from GenerateOuterLoops pass to GenericLowerToExplicitForm pass
  c. Will be used to perform affine fusion and scalar replacement as well as blocking analysis
3. **Explicit Datamovement Form** - Has only grid and thread type attributes. Blocking loops are explicit `**scf.for**` loops.
  a. The only valid form of d2m.generic past GenericLowerToExplicitForm pass.
4. **External Symbol Form** - Has no region body; kernels are extracted into separate func.func and referenced in threads attribute as symbols.
  a. Used in backend.

### DIFF
--- a/include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
+++ b/include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
@@ -2038,7 +2038,7 @@ class D2M_IndexOp<string mnemonic, list<Trait> traits = []> : D2M_GenericRegionO
     let arguments = (ins ConfinedAttr<I64Attr, [IntMinValue<0>]>:$dim);
     let results = (outs Index:$result);
     let assemblyFormat = [{ `(` $dim `)` attr-dict `:` type($result) }];
-    let hasFolder = true;
+    let hasFolder = false;
 
     let builders =
     [
@@ -2105,7 +2105,7 @@ def D2M_CoreIndexOp : D2M_GenericRegionOp<"core_index",
 
     let results = (outs Index:$result);
     let assemblyFormat = [{ `(` $dim `)` attr-dict `:` type($result) }];
-    let hasFolder = true;
+    let hasFolder = false;
 
     let builders =
     [
--- a/lib/Dialect/D2M/IR/D2MGenericRegionOps.cpp
+++ b/lib/Dialect/D2M/IR/D2MGenericRegionOps.cpp
@@ -1248,10 +1248,6 @@ void IterIndexOp::inferResultRanges(
                  getIndexRange(0, std::numeric_limits<uint32_t>::max()));
 }
 
-mlir::OpFoldResult IterIndexOp::fold(FoldAdaptor adaptor) {
-  return adaptor.getDimAttr();
-}
-
 //===----------------------------------------------------------------------===//
 // BlockIndexOp
 //===----------------------------------------------------------------------===//
@@ -1267,10 +1263,6 @@ void BlockIndexOp::inferResultRanges(
     mlir::SetIntRangeFn setResultRange) {
   setResultRange(getResult(),
                  getIndexRange(0, std::numeric_limits<uint32_t>::max()));
-}
-
-mlir::OpFoldResult BlockIndexOp::fold(FoldAdaptor adaptor) {
-  return adaptor.getDimAttr();
 }
 
 //===----------------------------------------------------------------------===//
@@ -1318,16 +1310,6 @@ void CoreIndexOp::inferResultRanges(
     mlir::SetIntRangeFn setResultRange) {
   setResultRange(getResult(),
                  getIndexRange(0, std::numeric_limits<uint32_t>::max()));
-}
-
-mlir::OpFoldResult CoreIndexOp::fold(FoldAdaptor adaptor) {
-  // Only fold to the constant `dim` when no virtualization map is present.
-  // If a map is present, the result depends on runtime core coordinates and
-  // must not be folded.
-  if (adaptor.getPhysToVirtMapAttr()) {
-    return {};
-  }
-  return adaptor.getDimAttr();
 }
 
 // TileMatmulBlockOp verification


.github/get-docker-tag.sh
```diff
@@ -5,6 +5,6 @@
 
 # Calculate hash from the following files. This hash is used to tag the docker images.
 # Any change in these files will result in a new docker image build
-DOCKERFILE_HASH_FILES=".github/Dockerfile.base .github/Dockerfile.ci .github/Dockerfile.ird .github/Dockerfile.cibuildwheel env/CMakeLists.txt env/init_venv.sh env/build-requirements.txt env/ttnn-requirements.txt env/patches/shardy.patch env/patches/shardy_mpmd_pybinds.patch test/python/requirements.txt env/install-tt-triage.sh"
+DOCKERFILE_HASH_FILES=".github/Dockerfile.base .github/Dockerfile.ci .github/Dockerfile.ird .github/Dockerfile.cibuildwheel env/CMakeLists.txt env/init_venv.sh env/build-requirements.txt env/ttnn-requirements.txt env/patches/shardy.patch env/patches/shardy_mpmd_pybinds.patch env/patches/affine-allow-symbol-vars.patch test/python/requirements.txt env/install-tt-triage.sh"
```
nsmithtt: Is the diff up to date?  Just double checking this is meant to be part of this change

include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
```diff
@@ -1979,6 +2055,22 @@ def D2M_BlockIndexOp : D2M_IndexOp<"block_index"> {
     }];
 }
 
+def D2M_BlockOffsetOp : D2M_IndexOp<"block_offset"> {
+    let summary = "Block Offset op.";
+    let description = [{
+      Represents block_factor * core_index as a constant-like operation. Useful
+      for representing blocking index math in an affine dialect compatible way.
+    }];
+}
```
nsmithtt, line 2077: I'd somewhat prefer not having this op if it's not required. I thought we discovered that affine symbols could express this.

include/ttmlir/Dialect/D2M/IR/D2MOps.td
nsmithtt, line 329: Maybe we should distinguish between Forms and Thread types or something, not sure what to call it.  It seems like Forms constitute 3 distinct stages:
1. Implicit
2. Affine Blocked
3. Explicit

Unified thread seems like it might be orthogonal to these forms since it exists across multiple forms.  Maybe I'm being too pedantic here.

nsmithtt, line 333: It's minor, but this could be the 4th Form when we capture these Forms in documentation.

lib/Dialect/D2M/IR/D2MGenericRegionOps.cpp
```diff
@@ -61,6 +61,7 @@ static mlir::ConstantIntRanges getIndexRange(uint64_t umin, uint64_t umax) {
                                                mlir::APInt(width, umax));
 }
 
+//===----------------------------------------------------------------------===//
```
nsmithtt: Typo?

lib/Dialect/D2M/IR/D2MGenericRegionOps.cpp
```diff
@@ -1243,7 +1250,9 @@ void IterIndexOp::inferResultRanges(
 }
 
 mlir::OpFoldResult IterIndexOp::fold(FoldAdaptor adaptor) {
-  return adaptor.getDimAttr();
+  // Cannot fold: runtime value depends on loop iteration, not the dimension
+  // index itself.
+  return {};
```
nsmithtt: We lost our convo on the last PR

Nick:
> This is only true once the generic is in explicit loop form and then the iter_index op should probably be made illegal. Perhaps we should revert this and update the verifier for this op to check which form the generic op is in and fail if it's in explicit loop form.

Brett:
> This results sometimes in canonicalization folding IterIndex op into arith.constant values. This seems incorrect? The 'return-an-attribute' pattern for ::fold implementations is meant for constant-like ops, but IterIndex op represents loop IVs?

nsmithtt: I think these fold patterns are required to be hoisted out of loops, but if we can confirm that just `Pure` is enough then maybe it's better to just remove hasFolder from the Index class.

lib/Dialect/D2M/IR/D2MGenericRegionOps.cpp
```diff
@@ -1267,6 +1276,50 @@ mlir::OpFoldResult BlockIndexOp::fold(FoldAdaptor adaptor) {
...
+
+void GetBlockFactorOp::inferResultRanges(
+    ::llvm::ArrayRef<::mlir::ConstantIntRanges> argRanges,
+    mlir::SetIntRangeFn setResultRange) {
+  setResultRange(getResult(),
+                 getIndexRange(0, std::numeric_limits<uint32_t>::max()));
+}
+
+mlir::OpFoldResult GetBlockFactorOp::fold(FoldAdaptor adaptor) {
+  return adaptor.getDimAttr();
```
nsmithtt: If above thread is true, wouldn't this also be replaced with arith.constant?  Perhaps we should follow suit with whatever we discovered above.

lib/Dialect/D2M/IR/D2MOps.cpp
```diff
@@ -1850,6 +1842,27 @@ unsigned d2m::GenericOp::getNumDims() {
   return getIndexingMap(0).getNumDims();
 }
 
+unsigned d2m::GenericOp::getNumBlockFactors() {
+  return static_cast<unsigned>(getBlockFactors().size());
+}
+
+std::optional<Operation *> d2m::GenericOp::getOutermostBlockingLoopOp() {
```
nsmithtt: It could be safari grep failing me, but seems like this function is unused.

lib/Dialect/D2M/IR/D2MOps.cpp
```diff
@@ -1907,6 +1929,75 @@ mlir::SmallVector<int64_t> d2m::GenericOp::getBlockFactorsValue() {
...
+    return false;
+  }
+
+  // Each block factor dimension must have a corresponding affine.for loop
+  // whose upper bound is defined by a get_block_factor() op.
+  unsigned numBlockFactors = static_cast<unsigned>(getBlockFactors().size());
+  SmallVector<bool> factorUsed(numBlockFactors, false);
+  unsigned loopCount = 0;
+
+  getRegion(0).walk([&](affine::AffineForOp forOp) {
```
nsmithtt, line 2057: I wonder if it might be better for the generic op to carry an attribute that states which form it's in, it'd make these boolean queries on the op state much simpler and this code would move to verification instead with additional verification depending on the proclaimed form.

include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
nsmithtt: If it's a huge hassle, then we can keep this op

bgrady-tt: I don't mind eliminating this op. It didn't really make much difference in the end; the indexing math had to be converted to a 'compatibility' form anyway to make the indexing analyzable.

include/ttmlir/Dialect/D2M/IR/D2MOps.td
bgrady-tt: Agreed; we could collapse the two modes case-wise like so:

1. Implicit
2. Affine Blocked
3. Explicit
4. Split Explicit

bgrady-tt: Split is only allowed for Explicit form, so it is really just 4 possibilities, not 6 = 3 x 2.

.github/get-docker-tag.sh
bgrady-tt: Removed this; holdover from fusion pass with llvm patch bundled.

lib/Dialect/D2M/IR/D2MOps.cpp
bgrady-tt: I feel like deriving the state based on region contents is better. Ensuring alignment of the 'form' attribute with actual region after every pass in verification doesn't seem like much of a speedup versus querying form on demand. Perhaps it is an IR readability win though?

lib/Dialect/D2M/IR/D2MOps.cpp
bgrady-tt: You are right, it is no longer used. Removing this ...

lib/Dialect/D2M/IR/D2MGenericRegionOps.cpp
bgrady-tt: I will leave things as-is I think for now. I will do some research on the 'correct' way to fold these ops; I think `Pure` should be enough to get this to work. But I don't think it is a functional difference; gcc (SFPI) should do CSE perfectly fine once lowered to EmitC.

lib/Dialect/D2M/IR/D2MOps.cpp
bgrady-tt: It is indeed unused after late stage refactors. I will remove it.

include/ttmlir/Dialect/D2M/IR/D2MOps.td
nsmithtt: Sounds good to me!

lib/Dialect/D2M/IR/D2MGenericRegionOps.cpp
nsmithtt: Yeah agree, but from an IR cleanliness perspective it's easier to read if there aren't multiple instances of this op scattered around.  But we can tackle as a follow on.

lib/Dialect/D2M/IR/D2MOps.cpp
nsmithtt: Yeah I can see arguments either way, I'll leave it up to you.  The LLVM conf talk on "[Normal forms for MLIR](https://www.youtube.com/watch?v=Esw84hH1Ed0)" does propose an attribute, but I can see how this is more of a requirement when dealing with an IR state and not just an isolated op.  The other drawback of an attribute is adding even more stuff to the generic op, but we can revisit this one in the future if we think it'll help in some way.

vvukomanTT (approved):

include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
bgrady-tt: I am going to retain `BlockOffsetOp`, as I think it will eventually make full affine dialect integration more feasible (without compatibility mode roundtrips).

include/ttmlir/Dialect/D2M/IR/D2MOps.td
bgrady-tt: Added this to the PR description. I think this is right.

lib/Dialect/D2M/IR/D2MGenericRegionOps.cpp
bgrady-tt: fixed.

lib/Dialect/D2M/IR/D2MGenericRegionOps.cpp
bgrady-tt: Going to consistently remove the fold implementations for D2M Index ops. Based on implicit bcast bug, I don't think this pattern is safe/correct in all cases, so better to not do it.

Issue #3026, opened 2025-04-21
d2m ttir.concat


nsmithtt:

nsmithtt:
Concat is largely done, we will file new issues for tracking bugs + follow on work

PR #7097, opened 2026-02-19
D2M: concat op

wenbinlyuTT:
### Ticket
Related-To: #3026

### Problem description
This PR implements the concat op for the simple case of tile-aligned sizes on the concat dim, and all inputs & output fit in L1.

### What's changed
- At high-level, the concat op is modeled as piecewise-affine data movement
  - Along the concat dimension, different regions of the output tensor map to different input tensors
  - At compile-time, generate index calculation logic and if-else DMA chains to realize the piecewiseness
  - At run-time, output tiles determine which input tile to fetch using their grid + shard coordinates
- The core idea: introduce a new `d2m.composite_view` op
  - `ttir.concat` is lowered to this lightweight metadata view that aggregates multiple tensors
  - This op implements `ViewOpInterface` and behave like a regular view through most of the pipeline
  - Concat-specific lowering is deferred until right before the `LowerDMAToFullyIndexedForm` pass
    - A new `ExpandDMAReadCompositeView` pass lowers straight to fully-indexed DMA reads
    - GenericOp expanded to the actual list of inputs
    - Piecewise indexing logic and if-else fully-indexed DMA chain generation
    - Shared util functions are moved to `Utils.cpp`
- Multiple compiler infrastructure updates for multi-input views was required
  - Currently there are strong assumption of 1-1 tensor/memref/view/etc. everywhere
  - The concat op / composite view is inherently multi-rooted & fine-grained & run-time determined
  - Several passes (notably D2MAllocate, GridSelection, and view utilities) were extended to handle composite views
- In multiple passes the existing lowering logic are avoided and the composite view gets its own handler
  - This is also to prepare for the follow up work of sub-tile concat, which is even more alien to existing logic and demands fine-grained control

### Checklist

### DIFF
--- a/lib/Dialect/D2M/Transforms/GridSelection.cpp
+++ b/lib/Dialect/D2M/Transforms/GridSelection.cpp
@@ -4,7 +4,6 @@
 
 #include "ttmlir/Dialect/D2M/Transforms/Passes.h"
 
-#include "ttmlir/AffineMapAnalysis.h"
 #include "ttmlir/AffineMapUtils.h"
 #include "ttmlir/Asserts.h"
 #include "ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.h"
@@ -59,10 +58,6 @@ findMaxDimAndAspectRatio(ArrayRef<int64_t> physicalShape) {
   }
   return {maxDimIndex, aspectRatio};
 }
-
-static llvm::SmallVector<int64_t>
-computeOptimalBlockShardedGrid(ArrayRef<int64_t> physicalShape,
-                               ArrayRef<int64_t> targetSquareGridShape);
 
 static llvm::SmallVector<int64_t>
 computeOptimalVirtualGrid(ArrayRef<int64_t> physicalShape,
@@ -584,12 +579,19 @@ struct ViewLayoutUpdateInfo {
   llvm::SmallVector<int64_t> grid;
 };
 
+struct CompositeViewUpdateInfo {
+  d2m::CompositeViewOp op;
+  unsigned operandIndex;
+  llvm::SmallVector<int64_t> grid;
+};
+
 struct GridAnalysisResult {
   llvm::SmallVector<llvm::SmallVector<int64_t>> optimalOperandGrids;
   llvm::SmallVector<ToLayoutUpdateInfo> toLayouts;
   llvm::SmallVector<TTNNTensorUpdateInfo> ttnnTensors;
   llvm::SmallVector<EmptyUpdateInfo> emptyOps;
   llvm::SmallVector<ViewLayoutUpdateInfo> viewLayouts;
+  llvm::SmallVector<CompositeViewUpdateInfo> compositeViews;
 };
 
 // This function normalizes the operand grids for a generic operation by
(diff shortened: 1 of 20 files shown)


lib/Dialect/D2M/Transforms/GridSelection.cpp
```diff
@@ -985,6 +990,70 @@ updateStreamLayoutOps(ArrayRef<StreamLayoutUpdateInfo> streamLayoutsToUpdate,
...
+      auto inputOptimalGrid =
+          computeOptimalGrid(inputType, inputPhysShape, config);
+
+      auto currentGrid = inputLayout.getGridShape(inputType);
+      if (llvm::ArrayRef(currentGrid) == llvm::ArrayRef(inputOptimalGrid)) {
+        reblockedInputs.push_back(input);
+        continue;
+      }
+
+      auto viewTensorType = utils::reblockTensor(inputType, inputOptimalGrid);
```
vmilosevic: :warning: **clang\-diagnostic\-error** :warning:
no member named `` reblockTensor `` in namespace `` mlir::tt::d2m::utils ``

lib/Dialect/D2M/Transforms/LowerDMAToFullyIndexedForm.cpp
```diff
@@ -460,6 +800,11 @@ class D2MLowerDMAToFullyIndexedForm
       D2MLowerDMAToFullyIndexedForm>::D2MLowerDMAToFullyIndexedFormBase;
 
   void runOnOperation() final {
+    if (failed(expandCompositeViews(getOperation()))) {
```
nsmithtt: Should this just be its own pass?

lib/Dialect/D2M/Transforms/LowerDMAToFullyIndexedForm.cpp
```diff
@@ -303,6 +297,352 @@ static Value generateFullyIndexedDMAOps(
...
+}
+
+static LogicalResult expandCompositeViewsInGeneric(IRRewriter &rewriter,
+                                                   GenericOp gOp) {
+  CompositeViewOp compositeView = nullptr;
+  gOp.walk([&](DMAReadOp dmaRead) {
+    auto maybeCompositeView = dmaRead.getSrc().getDefiningOp<CompositeViewOp>();
+    if (maybeCompositeView) {
+      assert(compositeView == nullptr &&
+             "Unsupported multiple composite views in one GenericOp.");
```
nsmithtt: We should file an issue to track this support.

lib/Dialect/D2M/Transforms/LowerDMAToFullyIndexedForm.cpp
```diff
@@ -303,6 +297,352 @@ static Value generateFullyIndexedDMAOps(
   return loopNest.results.front();
 }
 
+static LogicalResult expandCompositeDMARead(IRRewriter &rewriter,
```
nsmithtt: @bgrady-tt can you sign off on this file?

nsmithtt (approved):

lib/Dialect/D2M/Transforms/LowerDMAToFullyIndexedForm.cpp
wenbinlyuTT: I'm about to split this out into its own pass that is immediately before `LowerDMAToFullyIndexedForm`, but yeah the logic here will stay identical.

wenbinlyuTT: @bgrady-tt I have split the pass and the logic is in the `ExpandDMAReadCompositeView.cpp` file now.

Issue #7081, opened 2026-02-18
[Conv3d] `c_out_block` causing problems


vkovinicTT:
### Describe the bug

Running `conv3d` causes this error:
```bash
2026-02-12 11:35:18.342 | critical |          Always | TT_FATAL: C_out_block must be a multiple of 32 and divide evenly into output channels. Got C_out_block=32 and output_channels=48. (assert.hpp:104)
2026-02-12 11:35:18,344 - ERROR - ERROR: test=ttnn_fb.ttnn experienced an error with exception=TT_FATAL @ /localdev/vkovinic/tt-xla/third_party/tt-mlir/src/tt-mlir/third_party/tt-metal/src/tt-metal/ttnn/cpp/ttnn/operations/experimental/conv3d/device/conv3d_device_operation.cpp:100: args.output_channels % args.config.C_out_block == 0 && args.config.C_out_block % tt::constants::TILE_WIDTH == 0
info:
C_out_block must be a multiple of 32 and divide evenly into output channels. Got C_out_block=32 and output_channels=48.
backtrace:
--- /localdev/vkovinic/tt-xla/third_party/tt-mlir/src/tt-mlir/third_party/tt-metal/src/tt-metal/build_Release/lib/libtt_metal.so(+0x53346a) [0x7fee7547c46a]
--- ttnn::experimental::prim::Conv3dDeviceOperation::validate_on_program_cache_miss(ttnn::experimental::prim::Conv3dParams const&, ttnn::experimental::prim::Conv3dInputs const&)
--- /localdev/vkovinic/tt-xla/third_party/tt-mlir/src/tt-mlir/third_party/tt-metal/src/tt-metal/build_Release/lib/_ttnncpp.so(_ZN4ttnn16device_operation6detail29launch_operation_with_adapterINS0_26MeshDeviceOperationAdapterINS_12experimental4prim21Conv3dDeviceOperationEEEEEvRKNT_22operation_attributes_tERKNS8_13tensor_args_tERNS8_21tensor_return_value_tEPN2tt8tt_metal11distributed10MeshDeviceE+0x1a2) [0x7fee741fedf2]
[...]

PR #7105, opened 2026-02-19
Conv3d workaround when channels is not divisible by 32.

djstefanovicTT:
### Ticket
closes https://github.com/tenstorrent/tt-mlir/issues/7081

### Problem description
Conv3d requires out channel dimension to be divisible by TILE_WIDTH. This was not enforced in mlir.

### What's changed
Added a workaround that pads input tensors to be divisible by TILE_WIDTH and slice afterwards. This is correct because padded dimensions would only be involved in rows that get sliced, and the result is the same.

### Checklist

### DIFF
--- a/include/ttmlir/Dialect/TTNN/Transforms/Workarounds/Decomposition/Conv3dPadOutputChannelsRewritePattern.h
+++ b/include/ttmlir/Dialect/TTNN/Transforms/Workarounds/Decomposition/Conv3dPadOutputChannelsRewritePattern.h
@@ -16,6 +16,7 @@ namespace mlir::tt::ttnn::workarounds::decomposition {
 // When output channels aren't aligned, this pattern pads the weight (and bias)
 // tensors, runs conv3d with padded output channels, then slices the result back
 // to the original size.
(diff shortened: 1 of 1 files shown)


include/ttmlir/Dialect/TTNN/Transforms/Workarounds/Decomposition/Conv3dPadOutputChannelsRewritePattern.h
```diff
@@ -0,0 +1,30 @@
...
+
+#include "mlir/IR/PatternMatch.h"
+#include "mlir/Support/LogicalResult.h"
+
+namespace mlir::tt::ttnn::workarounds::decomposition {
+
+// Conv3d requires output channels to be a multiple of TILE_WIDTH (32).
+// When output channels aren't aligned, this pattern pads the weight (and bias)
+// tensors, runs conv3d with padded output channels, then slices the result back
+// to the original size.
```
sdjordjevicTT, line 18: Please create an issue on tt-metal.

PR #7106, opened 2026-02-19
Fix op-by-op bug

sdjukicTT:
### Ticket
/

### Problem description
In op-by-op runs, flatbuffer executions were leaving the device in a bad state, causing subsequent tests to fail unless the device was manually reset. This happened because the previous implementation used multiprocessing with os._exit() to terminate worker processes, which intentionally bypasses C++ destructors. Since tt-metal relies on those destructors for hardware cleanup, the device was left in an inconsistent state after each run.

### What's changed
Flatbuffer execution is now performed in a fresh subprocess instead of a multiprocessing worker. Unlike os._exit(), a subprocess that exits normally runs C++ destructors, allowing tt-metal to properly tear down the hardware. A dedicated _flatbuffer_worker.py script and run_subprocess_worker() utility were introduced for this purpose. Compilation steps are unchanged and continue to use the persistent multiprocessing pool.

### Checklist

### DIFF
--- a/python/common/compile_and_run_utils.py
+++ b/python/common/compile_and_run_utils.py
@@ -321,7 +321,7 @@ def run_subprocess_worker(
             cmd, timeout=timeout, check=False, capture_output=True, text=True
         )
 
-        if os.path.getsize(result_path) == 0:
+        if not os.path.exists(result_path) or os.path.getsize(result_path) == 0:
             raise RuntimeError(
                 f"Worker `{runner_script}` crashed without writing results. "
                 f"Exit code: {proc.returncode}. Stderr: {proc.stderr}"


python/common/compile_and_run_internal.py
```diff
@@ -192,14 +169,20 @@ def create_binary_from_fb_file(fb_file_path: str) -> Binary:
...
     """
-    process_manager = get_process_manager()
-    result: RunProcessResult = process_manager.run(worker_fn, worker_args_without_queue)
-    return result.return_code
+    result = run_subprocess_worker(
+        "_flatbuffer_worker.py", (flatbuffer_file_path,), timeout
+    )
+
+    if result["status"] == "error":
+        raise RuntimeError("Flatbuffer execution failed")
```
vkovinicTT: Can we raise a more descriptive error here, e.g.:
```python
raise RuntimeError(f"Flatbuffer execution failed: {result.get('error', 'unknown')}")
```

tools/op-by-op-infra/op_by_op_infra/mlir_module_executor.py
```diff
@@ -248,8 +248,9 @@ def _run(self) -> bool:
                 self._mark_execution_step(
                     ExecutionPhase.EXECUTED_FLATBUFFER, run_passed=True
                 )
-        finally:
-            return self._execution_result.device_run_passed
+        except RuntimeError:
+            pass
```
vkovinicTT: we are now only handling `RuntimeError`, but what happens if some other error occurs, should we handle it?

python/common/_flatbuffer_worker.py
```diff
@@ -0,0 +1,46 @@
...
+    try:
+        from ttrt.common.api import API
+
+        API.initialize_apis()
+        run_instance = API.Run(args={"binary": flatbuffer_path})
+        return_code, _ = run_instance()
+
+        if return_code != 0:
+            with open(result_path, "w") as f:
+                json.dump({"status": "error", "error": f"{return_code}"}, f)
```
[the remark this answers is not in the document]
sdjukicTT: There is no need since stderr will be captured separately by run_subprocess_worker and included in the error reporting

tools/op-by-op-infra/op_by_op_infra/mlir_module_executor.py
sdjukicTT: Updated errror handling

tapspatel (approved):

PR #7114, opened 2026-02-19
[d2m] view(stream) canonicalization pattern

dlokeTT:
### Problem description
When running TMs such as reshape that involve a stream layout followed by a compute op, we run into 
```
Encountered unsupported op.
UNREACHABLE executed at /localdev/dloke/tt-mlir/lib/Target/TTMetal/TTMetalToFlatbuffer.cpp:1023!
```

This is due to a `d2m.stream_layout` surviving thru the pipeline because its not being composed properly with the `d2m.view_layout` that is created when the `d2m.to_layout` that takes the TM's output and sends it back to host is folded with the `d2m.to_layout` that grabs the tensor from host and brings it to device.

### What's changed
Added a folding pattern for views of a stream that composes their affine maps and turns it into one stream op which is lowered properly downstream

### Checklist

### DIFF
--- a/lib/Dialect/D2M/IR/D2MOps.cpp
+++ b/lib/Dialect/D2M/IR/D2MOps.cpp
@@ -826,9 +826,9 @@ mlir::LogicalResult StreamLayoutOp::verify() {
       *this, "storage", "result", getStorage().getType(), getResult().getType(),
       /*checkSameElementType*/ true,
       /*checkSameMemorySpace*/ false,
-      /*checkSameRank*/ true,
+      /*checkSameRank*/ false,
       /*checkSameGridShape*/ false,
-      /*checkSameShardShape*/ true);
+      /*checkSameShardShape*/ false);
   if (failed(storageResultVerification)) {
     return storageResultVerification;
   }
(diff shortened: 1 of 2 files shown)


lib/Dialect/D2M/IR/D2MOps.cpp
```diff
@@ -1025,56 +1025,80 @@ mlir::OpFoldResult d2m::ViewLayoutOp::fold(FoldAdaptor adaptor) {
...
-  // Don't fold through reinterpretLayout views - they legitimately change
-  // logical shape and cannot be skipped.
-  if (consecutiveView.getReinterpretLayout()) {
-    return nullptr;
+  // Compose view(stream(x)) → view(x) with composed index maps.
+  // The stream's ViewLayoutAttr maps streamDims → physicalDims,
+  // the view's ViewLayoutAttr maps viewDims → streamDims.
+  // Composing gives viewDims → physicalDims, allowing the view to point
+  // directly at the stream's underlying input buffer.
+  if (auto streamOp = getInput().getDefiningOp<d2m::StreamLayoutOp>()) {
```
nsmithtt: I think we might need to write this fold on stream op instead because I don't think it's valid to get rid of the stream.

So:
```
%s = stream {map=m} %i
%r = view {map=n} %s
```

Should become:
```
%r = stream {map=m(n)} %i
```

dlokeTT: Updated to follow this. One thing with this approach was I had to relax the stream_layouts verifier to allow for rank and shape mismatch

PR #7118, opened 2026-02-19
[TTNN] Improve SDPA fusing by adding matmul canonicalization

jserbedzijaTT:
### Problem description
The frontend lowers batched matmul by squeezing the leading `batch=1` dimension before each matmul (e.g., [1, heads, seq, dim] -> [heads, seq, dim]) and unsqueezing it after. This prevents the SDPA fusing pattern from matching because of those inserted reshapes. For example, the frontend produces IR like:

```
%0 = "ttir.reshape"(%q) <{shape = [8, 18, 64]}> : (tensor<1x8x18x64xf32>) -> tensor<8x18x64xf32>
%1 = "ttir.reshape"(%k) <{shape = [8, 64, 18]}> : (tensor<1x8x64x18xf32>) -> tensor<8x64x18xf32>
%2 = "ttir.matmul"(%0, %1) : (tensor<8x18x64xf32>, tensor<8x64x18xf32>) -> tensor<8x18x18xf32>
%3 = "ttir.reshape"(%2) <{shape = [1, 8, 18, 18]}> : (tensor<8x18x18xf32>) -> tensor<1x8x18x18xf32>
```

Instead of the equivalent 4D matmul:

`%0 = "ttir.matmul"(%q, %k) : (tensor<1x8x18x64xf32>, tensor<1x8x64x18xf32>) -> tensor<1x8x18x18xf32>`

### What's changed
- Added a `MatmulOp` canonicalization pattern at the TTIR level that absorbs leading squeeze/unsqueeze reshapes around matmul. When a matmul is bracketed by reshape ops that only remove/add leading size-1 dimensions (e.g., squeeze [1,B,M,K]->[B,M,K] on inputs, unsqueeze [B,M,N]->[1,B,M,N] on output), the pattern eliminates the reshapes and performs the matmul directly on the original higher-rank tensors.
- Added `f32-to-bf16` typecast insertion in the SDPA fusing pass. When SDPA inputs (Q, K, V, mask) are f32, the fusing pattern now inserts `TypecastOp` to `bf16` before constructing the fused SDPA op, and casts the output back to the original dtype. This is a workaround until `tt-metal` adds native f32 input support for SDPA ([tt-metal#36717](https://github.com/tenstorrent/tt-metal/issues/36717)).

### Checklist

### DIFF
--- a/test/ttmlir/Dialect/TTIR/canonicalize/matmul_squeeze_unsqueeze_canonicalize.mlir
+++ b/test/ttmlir/Dialect/TTIR/canonicalize/matmul_squeeze_unsqueeze_canonicalize.mlir
@@ -28,6 +28,19 @@ module {
     return %2 : tensor<1x64x32xbf16>
   }
 
+  // Only input B squeezed -- pattern fires via batch broadcasting.
+  func.func @matmul_one_input_b_squeezed(%arg0: tensor<64x128xbf16>, %arg1: tensor<1x128x32xbf16>) -> tensor<1x64x32xbf16> {
+    // CHECK-LABEL: @matmul_one_input_b_squeezed
+    // CHECK-NOT: "ttir.reshape"
+    // CHECK: "ttir.matmul"(%arg0, %arg1)
+    // CHECK-SAME: -> tensor<1x64x32xbf16>
+    // CHECK-NOT: "ttir.reshape"
+    %0 = "ttir.reshape"(%arg1) <{shape = [128 : i32, 32 : i32]}> : (tensor<1x128x32xbf16>) -> tensor<128x32xbf16>
+    %1 = "ttir.matmul"(%arg0, %0) : (tensor<64x128xbf16>, tensor<128x32xbf16>) -> tensor<64x32xbf16>
+    %2 = "ttir.reshape"(%1) <{shape = [1 : i32, 64 : i32, 32 : i32]}> : (tensor<64x32xbf16>) -> tensor<1x64x32xbf16>
+    return %2 : tensor<1x64x32xbf16>
+  }
+
   // No squeeze/unsqueeze reshapes -- pattern does not fire.
   func.func @matmul_no_reshape(%arg0: tensor<5x64x32xbf16>, %arg1: tensor<5x32x64xbf16>) -> tensor<5x64x64xbf16> {
     // CHECK-LABEL: @matmul_no_reshape


lib/Dialect/TTIR/IR/TTIROps.cpp
```diff
@@ -3479,6 +3479,187 @@ ::mlir::LogicalResult mlir::tt::ttir::MatmulOp::verify() {
...
+    resultShape.push_back(bShape.back());
+  }
+
+  return resultShape;
+}
+
+// MatmulOp canonicalization: absorb leading squeeze/unsqueeze reshapes.
+void mlir::tt::ttir::MatmulOp::getCanonicalizationPatterns(
+    mlir::RewritePatternSet &patterns, mlir::MLIRContext *context) {
+  // NOLINTBEGIN(clang-analyzer-core.StackAddressEscape)
```
vmilosevic: :warning: **clang\-tidy\-nolint** :warning:
unmatched `` NOLINTBEGIN `` comment without a subsequent `` NOLINTEND `` comment

kmabeeTT:
I tried this out in tt-xla (offline discussion/thread) against target model and it worked (extra SDPA fusion per layer in ttnn graphs occurs.

sdjordjevicTT (approved):

lib/Dialect/TTIR/IR/TTIROps.cpp
```diff
@@ -3479,6 +3479,184 @@ ::mlir::LogicalResult mlir::tt::ttir::MatmulOp::verify() {
...
+    mlir::Value inputA = op.getA();
+    mlir::Value inputB = op.getB();
+    mlir::Value newA = inputA;
+    mlir::Value newB = inputB;
+
+    auto squeezeA = inputA.getDefiningOp<ttir::ReshapeOp>();
+    auto squeezeB = inputB.getDefiningOp<ttir::ReshapeOp>();
+
+    int64_t squeezeDimsA = squeezeA ? isLeadingSqueeze(squeezeA) : -1;
+    int64_t squeezeDimsB = squeezeB ? isLeadingSqueeze(squeezeB) : -1;
```
mvasiljevicTT: What happens if inputs don't have reshapes?

lib/Dialect/TTIR/IR/TTIROps.cpp
```diff
@@ -3479,6 +3479,184 @@ ::mlir::LogicalResult mlir::tt::ttir::MatmulOp::verify() {
...
+  return getLeadingOnesRankDiff(reshapeOp.getType().getShape(),
+                                reshapeOp.getInput().getType().getShape());
+}
+
+// Computes the expected matmul output shape given two input shapes and
+// transpose flags. Returns std::nullopt if the shapes are incompatible.
+static std::optional<llvm::SmallVector<int64_t>>
+computeMatmulResultShape(llvm::ArrayRef<int64_t> inputAShape,
+                         llvm::ArrayRef<int64_t> inputBShape, bool transposeA,
+                         bool transposeB) {
```
mvasiljevicTT: Can this be deduced from input reshapes and original output shape? We know that the reshapes only add leading ones, so can we calculate number of leading ones for the output based on the leading ones on the inputs?

mvasiljevicTT (approved):

lib/Dialect/TTIR/IR/TTIROps.cpp
jserbedzijaTT: In the end, I removed this utility function and simplified the logic overall. The problematic cases were 1D inputs to matmul because if we dropped the squeeze TMs before them, we would semantically change the matmul op. I decided to skip canonicalization when 1D inputs are involved.

lib/Dialect/TTIR/IR/TTIROps.cpp
jserbedzijaTT: We skip the cannonicalization, it is checked here:
```
if (!isSqueezedA && !isSqueezedB) {
  return mlir::failure();
}
```

Issue #6312, opened 2025-12-12
[D2M] Per-op logic to set oob_values


vwellsTT:
### Is your feature request related to a problem? Please describe

As part of effort to support non-`undef` out-of-bounds values, we must create layout with correct enum value set per op.  For simple ops such as eltwise, `undef` remain appopriate.  For reductions etc, we must set them according to type (0 for addition-based ops, 1 for mul-based ops, etc).

### Describe the solution you'd like

I will change the TTIRToD2M pass itself s.t. when it inserts to_layout ops for each generic it inserts, their layouts properly reflect a sensible oob_val for the op itself.

vwellsTT:
https://github.com/tenstorrent/tt-mlir/pull/7119 handles this for reduction ops; however, a more detailed follow up is required for matmul.  Basic ideas:
* we implicitly get a 0s-padded tensor any time we read from host
* if we run most (but not all) eltwise ops on a 0-padded tensor, it preserves 0-paddedness (execption: exp, pow, some comparisons, etc)
* we should NOT do the work of computing masks when we don't need to--this is important for matmul, which runs often and is of great perf interest to us
* we need some kind of analysis to detect whether an input to a matmul has an actual dirty padded region based on chain of ops (if input comes directly from host, or through unbroken chain of non-dirtying ops, we're good; else we need to insert the pad op)

PR #7119, opened 2026-02-19
[D2M] support oob values on reductions

vwellsTT:
### Ticket
https://github.com/tenstorrent/tt-mlir/issues/6312

### Problem description
Now that we have support for using oob_values, we need to set certain ops to actually do so.

### What's changed
Both D2M reduction ops, sum and max, are set to 0 and -inf oob values respectively.  Adding folding patterns s.t. undef oob_value to_layouts can be folded with other values (since undef means "I don't have constraints").  Add basic testing validation too

### Checklist

### DIFF
--- a/lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
+++ b/lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
@@ -330,9 +330,12 @@ protected:
       result[0].push_back(createOptimalLayoutOp(operand, memorySpaces[0], tiled,
                                                 noCollapse, rewriter, oobVal));
     }
+    // Outputs always use Undef: they are destination buffers being written
+    // into, so their padding fill value is irrelevant.  Only inputs need
+    // identity-element OOB to prevent padded tiles from corrupting reductions.
     for (Value operand : operandsAndResults[1]) {
       result[1].push_back(createOptimalLayoutOp(operand, memorySpaces[1], tiled,
-                                                noCollapse, rewriter, oobVal));
+                                                noCollapse, rewriter));
     }
 
     return result;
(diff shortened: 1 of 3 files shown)


lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
```diff
@@ -907,6 +924,19 @@ class D2MNamedReductionRewriter final
                                enableMulticastInference) {}
 
 private:
+  // Return the identity OOB fill value for this reduction's tile op.
+  // Padded elements must not affect the reduction result.
+  static constexpr ttcore::OOBVal getReductionOOBVal() {
+    if constexpr (std::is_same_v<TileOp, d2m::TileReduceMaxOp>) {
+      return ttcore::OOBVal::NegInf;
```
nsmithtt, line 934: The oob val it meant to track the current state of the tensor, not the desired state.  In other words, it's only legal to set neginf if some masking op ran and actually set that padded areas to neginf.

That said, it is probably also valuable to track requirements which are easier to tag during dialect conversion.  Thinking out loud, we could make it part of to_layout op which would have an undef oob coming in and a neginf coming out.  Any unresolved oob changes automatically get lowered to mask ops during to_layout lowering.

vwellsTT: Let's follow-up offline, a bit unclear on this, imo if oob_value is part of the layout, it should be used like this

nsmithtt: Followed up offline, Vincent's approach here is better than what I was thinking where oobval is interpreted as an input requirement resolved by to_layout lowering.

nsmithtt (approved):

Issue #7002, opened 2026-02-11
Support presharded/unsharded graphs in builder


tapspatel:
**Question**
What’s going on with presharded/unsharded tensors?

**Background**
In multi-device graphs jax/torch-xla will annotate arguments/results with a sharding attribute. In addition, the framework will either “preshard” the tensor and give it to ttmlir runtime, or defer to ttmlir runtime to shard the tensor. 

**Goal**
Identify which situations the framework will presharded vs not.

For the purposes of this document, I will not refer to tensors as presharded or unsharded, as that is causing confusion. Instead, we will look at whether the framework calling mlir runtime provides a list of tensors or a single tensor for each argument and expects a list of tensors or a single tensor for each result.

**Framework**
torch_xla
jax

We will define the problem in two parts

- Arguments

When does the framework that calls mlir runtime provide a list of tensors for a given argument or just provide a single tensor for a given argument

- Results

When does the framework that calls mlir runtime expect a list of tensors for a given result or just a single tensor for a given result

Notes
we need to figure out all the situations in how torch_xla + jax answer the above two questions

- jax

arguments
if there is a sdy.sharding annotation, framework will provide a list of tensors, otherwise framework will provide a single tensor
results
if there is a sdy.sharding annotation, framework expects a list of tensors, otherwise framework expects a single tensor

- torchxla

arguments
will always have a sdy.sharding annotation
framework will always provide a list of tensors for each result
results
framework will always expect a list of tensors for each result (regardless if it has a sdy.sharding annotation or not)

**Conclusion**
ttmlir compiler can infer the shard status of the arguments. If the argument has a sdy.sharding annotation, the framework will provide a list of tensors. If not, the framework provides only a single tensor

ttmlir compiler will not be able to infer the shard status of results based solely on the shlo graph itself. The set of rules that govern torch_xla are different from jax. Therefore, we need support from frontend teams to provide this as a pipeline option.

**Implementation**
Two attributes
shard_status
presharded
unsharded
local shape
the shape of this tensor local to a device

Pass: ApplyArgumentShardStatus
for arguments, compiler can determine which arguments are presharded/unsharded
for results, caller of this pass must pass it in as a pipeline option (list of bool for each result to indicate its shard status, true = presharded, false = unsharded)

Pass: AnnotateLocalShapes
annotate each argument + result with its local shape based on the solved shardings from shardy

[...]

PR #7120, opened 2026-02-19
Support presharded/unsharded graphs in builder

tapspatel:
Ticket: https://github.com/tenstorrent/tt-mlir/issues/7002

Support presharded/unsharded mix of tensors across ttmlir ecosystem.

This PR needs to land in conjunction with tt-xla support: https://github.com/tenstorrent/tt-xla/issues/3386


include/ttmlir/Dialect/StableHLO/Transforms/Passes.td
```diff
@@ -328,6 +332,31 @@ def WrapUnderManualComputationPass : Pass<"wrap-under-manual-computation", "::ml
...
+{
+  let summary = "Annotate all arguments and results with their local shape.";
+  let description = [{
+    This pass will iterate through all arguments and results and annotate the local shape of that argument/result.
+    For single device graphs, the local shape is the original shape.
+    For multi-device graphs, the local shape can be sharded or replicated depending on the sharding config set in the graph.
+    This information is required by any tool calling ttmlir runtime to tell it how the inputs should be preprocessed.
+
+    Example, this pass will convert the following code:
+    ```mlir
```
tapspatel, line 346: taps: need to update

jgrimTT:
FYI I'm getting an error when I run `pytest test/python/golden/test_ttir_ops.py::test_presharded_arg`:

```
    def set_arg_attribute(
        self, operand: Operand, new_attr_name: str, new_attr: Attribute
    ):
        func_op = operand.owner.owner
    
        arg_attr_list = func_op.arg_attrs
        new_arg_attr_list = []
        for arg_number, arg_attrs in enumerate(arg_attr_list):
            if arg_number == operand.arg_number:
                new_arg_attr = {}
                for attr in arg_attrs:
                    new_arg_attr[attr.name] = attr.attr
                new_arg_attr[new_attr_name] = new_attr
>               new_arg_attr_list.append(DictAttr.get(new_arg_attr))
                                         ^^^^^^^^^^^^^^^^^^^^^^^^^^
E               RuntimeError: std::bad_cast
```

tapspatel:
@jgrimTT sorry fixed it

lib/Dialect/StableHLO/Transforms/ApplyArgumentShardStatus.cpp
```diff
@@ -28,184 +33,116 @@ updateShardStatus(MLIRContext *context, mlir::ModuleOp &module,
...
+sharding attribute. For arguments, the framework will either provide a list of
+tensors or a single tensor for each argument and give it to ttmlir runtime. For
+results, the framework will either expect a list of tensors or a single tensor
+from ttmlir runtime.
+
+The situations are as follows:
+* jax
+arguments
+- if there is a sdy.sharding annotation, the framework will provide a list of
+tensors, otherwise framework will provide a single tensor results
```
jameszianxuTT: Is this meant to be under the heading of 'results'?

include/ttmlir/Dialect/StableHLO/Transforms/Passes.td
```diff
@@ -72,6 +73,10 @@ def ApplyArgumentShardStatusPass : Pass<"apply-argument-shard-status", "::mlir::
     "::mlir::sdy::SdyDialect",
     "::mlir::tt::ttir::TTIRDialect"
   ];
+
+  let options = [
+    ListOption<"resultPresharded", "result-presharded", "int64_t", "Set whether each result is presharded or not">,
```
mmanzoorTT, line 78: Can we use bool instead of int64_t? We only need 0/1 for this option.

include/ttmlir/Dialect/StableHLO/Transforms/Passes.td
```diff
@@ -328,6 +333,46 @@ def WrapUnderManualComputationPass : Pass<"wrap-under-manual-computation", "::ml
   ];
 }
 
+def AnnotateLocalShapesPass : Pass<"annotate-local-shapes", "::mlir::ModuleOp">
```
mmanzoorTT, line 336: We have `Update Global to Local` shape pass which also updates the shapes based on shardy configs. What's the difference between these two passes? Can't we use one pass to handle all shape conversions for multi-device graph?

lib/Dialect/StableHLO/Transforms/ApplyArgumentShardStatus.cpp
```diff
@@ -223,18 +160,28 @@ class ApplyArgumentShardStatusPass
...
         return;
       }
     });
+
+    // Loop through all results and annotate it with its shard status.
+    llvm::SmallVector<int64_t> resultPreshardedRef = llvm::to_vector(
+        llvm::map_range(resultPresharded, [](int64_t val) { return val; }));
+    rootModule.walk([&](func::FuncOp funcOp) {
+      if (failed(mlir::tt::stablehlo::updateResultShardStatus(
+              context, rootModule, builder, funcOp, resultPreshardedRef))) {
```
sdjukicTT, line 179: Will there be a problem if there are multiple funcOps with different number of results?

include/ttmlir/Dialect/StableHLO/Transforms/Passes.td
```diff
@@ -42,14 +42,15 @@ def ApplyArgumentShardStatusPass : Pass<"apply-argument-shard-status", "::mlir::
 {
   let summary = "Annotate arguments with their shard status.";
   let description = [{
-    This pass will analyze the module arguments and annotate arguments with whether they are pre-sharded or not. It will determine this based on analyzing shardy or mhlo annotations.
-    If shardy or mhlo annotations exist, the argument/results is presharded. If not, it is unsharded.
+    This pass will analyze the module arguments and annotate arguments with whether they are pre-sharded or not.
+    For argument, if a sdy.sharding or mhlo.sharding exists, it is presharded.
+    For results, the user must pass that as a compiler option.
```
mmanzoorTT, line 47: How it will work when we have multiple functions in a module or we are adding compiler option for `main's` results only?

lib/Dialect/StableHLO/Transforms/ApplyArgumentShardStatus.cpp
```diff
@@ -28,184 +33,116 @@ updateShardStatus(MLIRContext *context, mlir::ModuleOp &module,
...
+updateResultShardStatus(MLIRContext *context, mlir::ModuleOp &module,
+                        mlir::OpBuilder &builder, func::FuncOp &funcOp,
+                        llvm::SmallVector<int64_t> resultPreshardedRef) {
+  // If user did not provide result shard status, we will assume all results are
+  // unsharded.
+  if (resultPreshardedRef.empty() && funcOp.getNumResults() > 0) {
+    module.emitWarning("User did not provide result shard status, assuming all "
+                       "results are unsharded.");
+    resultPreshardedRef =
+        llvm::SmallVector<int64_t>(funcOp.getNumResults(), false);
```
mmanzoorTT:
```suggestion
        llvm::SmallVector<int64_t>(funcOp.getNumResults(), 0);
```

lib/Dialect/StableHLO/Transforms/ApplyArgumentShardStatus.cpp
tapspatel: At the start of this pipeline, we inline everything so there's only one funcop with a result list, and that should match the list the user provides

include/ttmlir/Dialect/StableHLO/Transforms/Passes.td
tapspatel: we could but it was for code modularity reasons and this pass is doing different things than update global to local is doing. This pass populates data structures that are used by the runtime, so there's a pretty strict contract between the two. Whereas update global to local doesn't have anything to do with runtime and doesn't use the data structures populated by annotate local shapes.

include/ttmlir/Dialect/StableHLO/Transforms/Passes.td
tapspatel: we inline everything at the start of stablehlo-pipelines. So there's only a single func in module. The "user" (ie xla) only needs to provide list of parameters for main func results.

include/ttmlir/Dialect/StableHLO/Transforms/Passes.td
tapspatel: fixed

lib/Dialect/StableHLO/Transforms/ApplyArgumentShardStatus.cpp
tapspatel: fixed

lib/Dialect/StableHLO/Transforms/ApplyArgumentShardStatus.cpp
tapspatel: fixed

include/ttmlir/Dialect/StableHLO/Transforms/Passes.td
tapspatel: fyi I had to change this back to int64_t. I forgot why I originally did it but ListOption<bool> does not work in MLIR because std::vector<bool> is a special bit-packed specialization that cannot produce a MutableArrayRef<bool>, which is what PassOptions expects.

tapspatel:
@tenstorrent/forge-developers-mlir-core @mrakitaTT @nsmithtt just waiting on your reviews :)

Issue #7082, opened 2026-02-18
[Conv3d] Padding must be `(0,x,x)`. got `(1, 1, 1)`


vkovinicTT:
### Describe the bug

The current `conv3d` op does not support temporal padding. When running `conv3d` with `(1, 1, 1)` padding, the following error is raised:
```
Padding must be (0,x,x). got (1, 1, 1)
```

**Possible solutions:**

1. Add temporal padding support to `tt-metal`.
2. Add a workaround in MLIR that extracts temporal padding and decomposes it into a `pad` + `conv3d` op.

### Steps to reproduce the issue

Here is the TTIR: [conv3d_repro.txt](https://github.com/user-attachments/files/25388237/conv3d_repro.txt)

PR #7121, opened 2026-02-19
Conv3d with depth pad

djstefanovicTT:
### Ticket
closes https://github.com/tenstorrent/tt-mlir/issues/7082

### Problem description
Conv3d could not work with depth padding. PadOp cannot pad along dimensions that are not among last 3 if any is >4.

### What's changed
Added workaround to decompose conv3d with depth padding to padOp + conv3d without depth padding. Added decomposition to padOp that permutes dimension that needs padding to satisfy metal constraint and creates padOp along appropriate dimension and then permutes back to original shape.

### Checklist

### DIFF
--- a/include/ttmlir/Dialect/TTNN/Transforms/Workarounds/Decomposition/Conv3dBlockingRewritePattern.h
+++ b/include/ttmlir/Dialect/TTNN/Transforms/Workarounds/Decomposition/Conv3dBlockingRewritePattern.h
@@ -52,10 +52,10 @@ private:
   // - Divisible by 16 and ≤128 → use as-is
   // - > 128 → cap at 128 (max efficient blocking)
   uint32_t calculateOptimalCInBlock(uint32_t in_channels) const {
-    if (in_channels % 16 != 0) {
+    if (in_channels % 16 != 0 || in_channels > 128) {
       return 0; // Bypass validation
     }
-    return std::min(in_channels, 128u);
+    return in_channels;
   }
 
   std::optional<Conv3dConfigAttr>


jameszianxuTT:
Q - Are we also doing fusion of `conv 3d op (pad op (input))` -> conv3d with padding attr?

djstefanovicTT:
> Q - Are we also doing fusion of `conv 3d op (pad op (input))` -> conv3d with padding attr?

There is no fusion of convolution ops with pad ops.

include/ttmlir/Dialect/TTNN/Transforms/Workarounds/Decomposition/PadHighDimRewritePattern.h
```diff
@@ -0,0 +1,32 @@
...
+
+#ifndef TTMLIR_DIALECT_TTNN_TRANSFORMS_WORKAROUNDS_DECOMPOSITION_PADHIGHDIMREWRITEPATTERN_H
+#define TTMLIR_DIALECT_TTNN_TRANSFORMS_WORKAROUNDS_DECOMPOSITION_PADHIGHDIMREWRITEPATTERN_H
+
+#include "ttmlir/Dialect/TTNN/IR/TTNNOps.h"
+
+#include "mlir/IR/PatternMatch.h"
+#include "mlir/Support/LogicalResult.h"
+
+namespace mlir::tt::ttnn::workarounds::decomposition {
```
sdjordjevicTT, line 13: Please add a builder test for this case:
https://github.com/tenstorrent/tt-mlir/blob/96090c423381c82f0cfa298f86e794430db09a8a/test/python/golden/ttir_ops/data_movement/test_data_movement.py#L108

include/ttmlir/Dialect/TTNN/Transforms/Workarounds/Decomposition/Conv3dBlockingRewritePattern.h
```diff
@@ -52,10 +52,10 @@ class Conv3dBlockingRewritePattern : public mlir::OpRewritePattern<Conv3dOp> {
   // - Divisible by 16 and ≤128 → use as-is
   // - > 128 → cap at 128 (max efficient blocking)
   uint32_t calculateOptimalCInBlock(uint32_t in_channels) const {
-    if (in_channels % 16 != 0) {
+    if (in_channels % 16 != 0 || in_channels > 128) {
       return 0; // Bypass validation
     }
-    return std::min(in_channels, 128u);
+    return in_channels;
```
jameszianxuTT, line 58: I think this is the same as returning 0 (which metal treats as setting [C_in_block=in_channels](https://github.com/tenstorrent/tt-metal/blob/895406f9cfea413c7f73ebb1f9de5dcc54e06679/ttnn/cpp/ttnn/operations/experimental/conv3d/device/conv3d_device_operation.cpp#L141)) - meaning we will end up not picking any smaller C_in block.  

I think this is fine, though we could just return 0 in the body of the function

PR #7123, opened 2026-02-19
test_ttir_unary golden tests bf16 dtype fix

ckaravasilisTT:
Small fix to set the dtype correctly. It was previously always set to `torch.float32`, causing tests to fail when parameterized with bf16.

Added bf16 tests to `test_unary_ops`.


test/python/golden/ttir_ops/eltwise/test_ttir_unary.py
wenbinlyuTT, line 252: Just curious, we're not adding bf16 here because there're still other tests that are failing?

ckaravasilisTT: I'm not sure of the original reasoning. In my local testing, all but `log` pass for bf16. I have not yet looked into this error.

wenbinlyuTT: Thanks! Then it would be great if we can add it here with this PR, and xfail/skip log+metal+bf16. But it's your choice if you want it to be part of a follow-up work.

ckaravasilisTT: The log +bf16 issue ended up being pretty simple. I also added bf16 tests

PR #7140, opened 2026-02-20
Minor fixes for conv3d

djstefanovicTT:
### Problem description
Prepare weights for conv3d had multiple bugs. Conv3d also requires bfloat16 as data type.

### What's changed
Added workaround to force inputs and outputs for conv3d to bfloat16.  Added logic equivalent to metals prepare weights https://github.com/tenstorrent/tt-metal/blob/main/models/tt_dit/utils/conv3d.py#L118.

### Checklist

### DIFF
--- a/lib/Conversion/TTIRToTTNN/TTIRToTTNN.cpp
+++ b/lib/Conversion/TTIRToTTNN/TTIRToTTNN.cpp
@@ -1522,9 +1522,19 @@ public:
     auto outputDtypeAttr =
         rewriter.getAttr<ttcore::DataTypeAttr>(outputLayoutAttr.getDataType());
 
-    // Permute + reshape weight: (O, C/G, K_D, K_H, K_W) → (K_D*K_H*K_W*C/G, O)
-    Value reshapedWeight = reshapeWeightForConv3d(adaptor.getWeight(), weightTy,
-                                                  rewriter, op.getLoc());
+    // Compute optimal C_in blocking for weight preparation.
+    // Uses the padded (tile-aligned) channel count since reshapeWeightForConv3d
+    // pads C/G to TILE_WIDTH internally.
+    constexpr int64_t TILE_WIDTH_CONST = ttcore::TileType::getDefaultShape()[1];
+    uint32_t paddedInChann =
+        llvm::divideCeil(static_cast<int64_t>(weightTy.getDimSize(1)),
+                         TILE_WIDTH_CONST) *
+        TILE_WIDTH_CONST;
+    uint32_t cInBlock = ttnn::utils::calculateOptimalCInBlock(paddedInChann);
+
+    // Permute + block + reshape weight
+    Value reshapedWeight = reshapeWeightForConv3d(
+        adaptor.getWeight(), weightTy, rewriter, op.getLoc(), cInBlock);
 
     // Reshape bias tensor: (1, 1, 1, 1, O) → (1, O)
     Value reshapedBias = adaptor.getBias();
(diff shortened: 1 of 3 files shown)


lib/Conversion/TTIRToTTNN/TTIRToTTNN.cpp
```diff
@@ -1542,6 +1540,20 @@ class Conv3dOpConversionPattern : public OpConversionPattern<ttir::Conv3dOp> {
...
+    if (inChannels % TILE_WIDTH != 0) {
+      int32_t cinPadAmount = static_cast<int32_t>(
+          llvm::divideCeil(inChannels, TILE_WIDTH) * TILE_WIDTH - inChannels);
+      // Input is now NDHWC; pad dim 4 (C)
+      llvm::SmallVector<int32_t> padding = {0, 0, 0, 0, 0,
+                                            0, 0, 0, 0, cinPadAmount};
+      input = ttir_to_ttnn::utils::generatePad(
+          mlir::cast<TypedValue<RankedTensorType>>(input), padding, rewriter,
+          ttmlir::utils::appendLocationSuffix(op.getLoc(), "_pad_cin"));
+      inChannelsAttr = rewriter.getI32IntegerAttr(inChannels + cinPadAmount);
```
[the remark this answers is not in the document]
jameszianxuTT: This is fine for now, groups == 1 is a constraint in metal conv3d. We should probably assert that in the lowering as well!

lib/Conversion/TTIRToTTNN/TTIRToTTNN.cpp
```diff
@@ -1542,6 +1540,20 @@ class Conv3dOpConversionPattern : public OpConversionPattern<ttir::Conv3dOp> {
...
+    }
```
jserbedzijaTT, line 1570: We should move this to the workaround pass at some point.

lib/Conversion/TTIRToTTNN/TTIRToTTNN.cpp
```diff
@@ -1601,30 +1613,49 @@ class Conv3dOpConversionPattern : public OpConversionPattern<ttir::Conv3dOp> {
...
+      llvm::SmallVector<int32_t> padding = {0, 0, 0, 0, 0, 0, 0, cinPadAmount,
+                                            0, 0};
+      permutedWeight = ttir_to_ttnn::utils::generatePad(
+          mlir::cast<TypedValue<RankedTensorType>>(permutedWeight), padding,
+          rewriter, ttmlir::utils::appendLocationSuffix(loc, "_pad_cin"));
+    }
+
     int64_t flattenedDim =
-        kernelDepth * kernelHeight * kernelWidth * inChannPerGroup;
+        kernelDepth * kernelHeight * kernelWidth * paddedInChannPerGroup;
```
jserbedzijaTT: Correct me if I'm wrong, but this belongs to flatten sliding window pass, right?

djstefanovicTT:
> Nit: I think we should cover ourselves for the Conv3D metal constraints
> 
> Is it possible to assert that C_in_groups == C_in or 0 somewhere?
> 
> Otherwise the C_in_pad + reshape implemented here won't work without extra logic to pre-swizzle the C_in_block and will again fail silently.

In workaround we set it to 0 or c_in, does that cover it?

jameszianxuTT:
> In workaround we set it to 0 or c_in, does that cover it?

Yes, but should someone change that workaround in the future, conv3d will start silently failing PCC checks. I think the timing of the workaround might be after this pass though, so maybe it is impossible to make an assert at this point

My point is that there's an implicit connection required for correctness on what c_in_block gets set to, and if we set it to something other than 0 or c_in, the ttir-ttnn lowering is no longer valid and will produce bad results. The ttir-ttnn lowering here is specialized to the case where c_in_block=0 or c_in, and isn't correct for the more general case, and I just want to make that constraint explicit somehow if possible :)

djstefanovicTT:
> Yes, but should someone change that workaround in the future, conv3d will start silently failing PCC checks. I think the timing of the workaround might be after this pass though, so maybe it is impossible to make an assert at this point

I added the full conversion even when c_in is not 0, it fixes OOM issues as well.

jameszianxuTT:
Oh, nice! That's a good solution

PR #7122, opened 2026-02-19
[ttnn.jit] fix double buffering

vtangTT:
### Ticket
https://github.com/tenstorrent/tt-mlir/issues/6984

### Problem description
As described in issue

### What's changed
- changed GridSelection helper `insertTTNNDRAMStreams` to `inserTTTNDRAMViews` to let Allocator pass insert stream and double buffer properly
    - allocator would ignore already inserted streams, so inserting a stream here would prevent double buffering
- Allocator pass should always insert a stream for DRAM operands
- `D2MToTTNN` helper `createCBDescriptors` will now account for single/double buffered memrefs when calculating total cb size

### Checklist
- [x] New/Existing tests provide coverage for changes

PR #7144, opened 2026-02-21
[ttnn.jit] Infer Intermediate Layouts in D2M GridSelection Pass

arminaleTT:
### Problem description
TTNN JIT currently requires all intermediate tensors to have a valid `TTNNLayoutAttr` encoding before the start of the TTIRToTTMetal pipeline. This places a huge burden on the JIT frontend, as it has no machinery to validate or optimize these intermediate layouts. The `GridSelection` pass in the TTIRToTTMetal pipeline already performs this exact task, but exclusively for the "core" D2M compiler. This PR extends the `GridSelction` pass to pick "optimal" layouts for tensors coming from TTNN.

There are two key differences between ttnn.jit tensors and regular D2M tensors in this context:
1. In ttnn.jit, the `func.func` args have a fully specified `TTNNLayoutAttr` that is pinned and cannot be modified. This means any reblocking of a `func.func` arg is a remote load and requires a stream.

2. In ttnn.jit, the allocations happen through the ttnn compiler. This means that any inferred layout must be convertible to a `TTNNLayoutAttr` and set on the output of a `ttnn.empty` op. This means that the grid padding performed in GridSelection cannot be easily applied to ttnn tensors.

#### GridSelection Pass
- Merged the TTNN tensor processing code path into the the regular code path that infers optimal grids and added a `ttnnMode` pass argument.
  - All tensors will have an optimal grid inferred, be reblocked to the optimal grid, and the generic will be rewritten as expected. This happens through the same high level functions as core D2M, with ttnn-mode special processing contained to a few low level functions
  - For `func.func` args, the reblocking view is nontrivial and requires a stream insertion in the allocator
  - For DRAM `func.func` args, the streaming logic is unchanged. This logic inserts views for DRAM args to maximize grid utilization 
  - For intermediate tensors, the optimal grid is simply folded into the `d2m.empty` that allocates the tensor
- Setting `ttnnMode=true` disables all grid padding and applies the `collapsed_intervals` expected by TTNN.
- Introduced `GridAnalysisResult` and `GridSelectionConfig` structs to contain code complexity and simplify function signatures

#### D2M Allocate Pass
- Any generic operand with a non-trivial view (a view with non-identity mapping) will have a stream inferred and inserted. These non-trivial views represent real data movement and happen when `func.func` args are reblocked.

#### D2MToTTNN Pass
Inferred intermediate tensors are now bufferized and represented by `memref.alloc` ops with `DeviceLayoutAttr`
- Match `memref.alloc` ops and rewrite them as `ttnn.empty`, translating the mermef with `DeviceLayoutAttr` to a tensor with `TTNNLayoutAttr`
- Delete `memref.dealloc` ops
- Modify the `d2m.generic` rewriter to set the IOs and CBs that originate from `memref.alloc` ops. We need both the  TTNN tensor and the metal memref for the IO and CB respectively, which creates an annoying dependence between the rewrite patterns. Fixing this (and other pass clean up) is tracked #7158  

#### TTNN JIT Frontend
- Set `CREATE_INTERMEDIATE_LAYOUT` to `false` to stop creating output layouts.
Before:
```
#l1 = #ttnn.buffer_type<l1>
#ttnn_layout = #ttnn.ttnn_layout<(d0, d1) -> (d0, d1), <1x1>, memref<3x3x!ttcore.tile<32x32, bf16>, #l1>, <block_sharded>, exactGrid = true>
module {
  func.func @abs(%arg0: tensor<96x96xbf16, #ttnn_layout>) -> tensor<96x96xbf16, #ttnn_layout> {
    %0 = "ttir.abs"(%arg0) : (tensor<96x96xbf16, #ttnn_layout>) -> tensor<96x96xbf16, #ttnn_layout>
    %1 = ttir.empty() : tensor<96x96xbf16, #ttnn_layout>
    %2 = ttir.to_layout %0, %1 : tensor<96x96xbf16, #ttnn_layout> into tensor<96x96xbf16, #ttnn_layout> -> tensor<96x96xbf16, #ttnn_layout>
    return %2 : tensor<96x96xbf16, #ttnn_layout>
  }
}
```
After:
```
#l1 = #ttnn.buffer_type<l1>
#ttnn_layout = #ttnn.ttnn_layout<(d0, d1) -> (d0, d1), <1x1>, memref<3x3x!ttcore.tile<32x32, bf16>, #l1>, <block_sharded>, exactGrid = true>
module {
  func.func @abs(%arg0: tensor<96x96xbf16, #ttnn_layout>) -> tensor<96x96xbf16, #ttnn_layout> {
    %0 = "ttir.abs"(%arg0) : (tensor<96x96xbf16, #ttnn_layout>) -> tensor<96x96xbf16>
    %1 = ttir.empty() : tensor<96x96xbf16, #ttnn_layout>
    %2 = ttir.to_layout %0, %1 : tensor<96x96xbf16> into tensor<96x96xbf16, #ttnn_layout> -> tensor<96x96xbf16, #ttnn_layout>
    return %2 : tensor<96x96xbf16, #ttnn_layout>
  }
}
```
#### d2m.empty Op
- Fixed op traits so unused empty ops are correctly removed by the canonicalize pass

### Limitations
- TTNN JIT can no longer output DRAM interleaved tensors
  - GridSelection/Allocator will not infer DRAM buffers, and setting the output memory config to DRAM will fail due to #6735 
- The GridSelection logic is not truly identical between TTNN and D2M core tensors due to the lack of support for grid padding. More details in #7162 

### Follow Ups:
Follow up work is tracked in #7156.

### Testing
- Adapted IR tests for the ttnn tensor translation tests
- A few IR tests updated due to new `d2m.empty` canonicalization behaviour
- IR tests for TTNNThroughD2M pipeline have been temporarily disabled until the TTNNToTTIR pass is updated to remove intermediate layouts #7160 
- JIT tests are all passing except:
  - matmul: Requires cleanup in JIT front end #7163 
  - mesh tensor: We supported it without actually modelling multi-device semantics. We may be able to resurrect it with changes to D2MToTTNN #7161 
  - A small set of shapes x grids failing PCC: #7157 

Nightly run: https://github.com/tenstorrent/tt-mlir/actions/runs/22682500345

### DIFF
--- a/include/ttmlir/Dialect/D2M/IR/D2MOps.td
+++ b/include/ttmlir/Dialect/D2M/IR/D2MOps.td
@@ -43,7 +43,6 @@ def D2M_ViewLayoutOp
 
   let extraClassDeclaration = [{
     bool isReblockOnly();
-    AffineMap getResultAffineMap();
   }];
 
   let hasVerifier = 1;


lib/Conversion/D2MToTTNN/D2MToTTNN.cpp
```diff
@@ -25,6 +28,110 @@
...
+          ctx, d2m::utils::getPhysicalGridShape(memrefValue));
+      TT_assertv(gridShape.size() >= 2u,
+                 "Expected at least 2 dimensions in grid shape");
+      int64_t gridY = gridShape[gridShape.size() - 2];
+      memLayoutEnum = (gridY > 1) ? ttnn::TensorMemoryLayout::HeightSharded
+                                  : ttnn::TensorMemoryLayout::WidthSharded;
+    }
+  }
+
+  size_t rank = 2;
```
vwellsTT: nit: constexpr size_t kRank if this is fixed

lib/Conversion/D2MToTTNN/D2MToTTNN.cpp
```diff
@@ -25,6 +28,110 @@
...
+
+  SmallVector<int64_t> tileDims = {1, 1};
+  if (auto tileType =
+          mlir::dyn_cast<ttcore::TileType>(memrefType.getElementType())) {
+    tileDims[0] = tileType.getHeight();
+    tileDims[1] = tileType.getWidth();
+  }
+
+  // Compute logical shape in elements: (grid * shard * tileDims). If dims have
+  // been collapsed, we cannot recover the original uncollapsed shape.
```
vwellsTT, line 46: given we cannot recover original shape, I'm not sure what "logical shape" means here; I think it might be confusing vs our usage elsewhere as well

lib/Conversion/D2MToTTNN/D2MToTTNN.cpp
```diff
@@ -218,8 +325,19 @@ class D2MGenericRewriter : public OpConversionPattern<d2m::GenericOp> {
           ttnn::KernelCBFormatAttr::get(ctx, i, dtype, pageSize);
 
       ttnn::KernelCBGlobalBufferAddressOfTensorAttr globalCBIndexOfTensor;
-      if (auto castOp = mlir::dyn_cast_if_present<ttir::TTNNMetalLayoutCastOp>(
-              cb.getDefiningOp())) {
+
+      // This is brittle. Ideally we should specifically identify outputs and
```
vwellsTT: might be good to attach a TODO + ticket here

lib/Conversion/D2MToTTNN/D2MToTTNN.cpp
```diff
@@ -267,15 +391,42 @@ class D2MGenericRewriter : public OpConversionPattern<d2m::GenericOp> {
                    "Expected TTNNMetalLayoutCastOp producing stream input.");
         return {innerCastOp.getOperand(), viewOp.getInput()};
       }
+
+      if (auto allocOp = mlir::dyn_cast_if_present<memref::AllocOp>(
+              viewOp.getInput().getDefiningOp())) {
+        // This is a view on top of the output of a previous generic. This
```
vwellsTT: I'm having trouble understanding "the generic that implements the to_layout for the user-selected output layout." part, can you maybe reword?

lib/Conversion/D2MToTTNN/D2MToTTNN.cpp
```diff
@@ -267,15 +391,42 @@ class D2MGenericRewriter : public OpConversionPattern<d2m::GenericOp> {
...
+        "Expected stream_layout, view_layout, memref.alloc, ttnn.empty, or "
+        "cast op as operand.");
   }
 
   LogicalResult
   matchAndRewrite(d2m::GenericOp op, d2m::GenericOpAdaptor adaptor,
                   ConversionPatternRewriter &rewriter) const final {
+    // The ttnn.generic op requires ttnn tensor operands. Defer rewriting until
+    // memref.alloc operands are converted so we have the memref->ttnn
+    // tensor translations
```
vwellsTT: nit: .

lib/Dialect/D2M/Transforms/GridSelection.cpp
```diff
@@ -137,12 +160,25 @@ computeOptimalVirtualGrid(ArrayRef<int64_t> physicalShape,
...
-    ttcore::MetalLayoutAttr layout, mlir::RankedTensorType tensorType,
-    ArrayRef<int64_t> targetSquareGridShape, OpBuilder &builder) {
+// Compute physical shape for a MetalLayoutAttr. In TTNN mode, returns the raw
+// physical shape without alignment adjustments. Otherwise, computes grid-aware
+// dimension alignments and derives the physical shape (always tile-aligned).
+static llvm::SmallVector<int64_t>
+computePhysicalShape(Value operand, const GridSelectionConfig &config,
+                     OpBuilder &builder) {
+  // Pass through the unit-reblocking view, if present.
+  if (auto view = operand.getDefiningOp<d2m::ViewLayoutOp>()) {
```
vwellsTT, line 154: ~~nit: I would move this below the ttnn short circuit since it's unrelated~~ nvm I'm dumb

lib/Dialect/D2M/Transforms/GridSelection.cpp
```diff
@@ -218,54 +253,78 @@ shouldImplementAsVirtualGrid(RankedTensorType tensorType,
...
+    auto i64Ty = IntegerType::get(builder.getContext(), 64);
+    auto intervalTy = RankedTensorType::get({1, 2}, i64Ty);
+    // Collapse all leading dimensions into the height dimension.
+    collapsedIntervals =
+        DenseIntElementsAttr::get(intervalTy, llvm::ArrayRef<int64_t>({0, -1}));
+
+    // Set dim alignments: all 1s except last two are tile default shape.
+    newDimAlignments.assign(oldLayout.getLogicalShape().size(), 1);
+    newDimAlignments[newDimAlignments.size() - 1] =
+        ttcore::TileType::getDefaultShape()[0];
```
vwellsTT: nit: imo ugly to call getDefaultShape twice, I'd do it once and access twice vs calling twice and accessing in place

lib/Dialect/D2M/Transforms/GridSelection.cpp
```diff
@@ -442,24 +499,176 @@ static void optimizeToLayoutGrid(d2m::ToLayoutOp toLayoutOp,
   }
 }
 
+static bool isTTNNOperand(Value operand) {
+  if (operand.getDefiningOp<ttir::TTNNMetalLayoutCastOp>()) {
+    return true;
+  }
+  if (auto view = operand.getDefiningOp<d2m::ViewLayoutOp>()) {
+    return view.getInput().getDefiningOp<ttir::TTNNMetalLayoutCastOp>();
```
vwellsTT: I'm not 100% sure but I think Nick told me recently we don't always fold views, so perhaps recursing would be better, in case there's a view-chain

vwellsTT (commented):
I'm going to give this another pass tomorrow morning when I'm fresher, but I think it mostly seems good to me.  Some nits/suggestions for now.

lib/Conversion/D2MToTTNN/D2MToTTNN.cpp
```diff
@@ -25,6 +28,110 @@
...
+  MemRefType memrefType = mlir::cast<MemRefType>(memrefValue.getType());
+
+  auto bufferType = ttnn::BufferType::DRAM;
+  if (auto memSpace = mlir::dyn_cast_if_present<ttcore::MemorySpaceAttr>(
+          memrefType.getMemorySpace());
+      memSpace && memSpace.getValue() == ttcore::MemorySpace::DeviceL1) {
+    bufferType = ttnn::BufferType::L1;
+  }
+
+  auto deviceLayout = ttcore::getDeviceLayout(memrefValue);
```
vwellsTT, line 74: is this distinct from `memrefType.getLayout()` used elsewhere? If so, naming is bit confusing imo

lib/Dialect/D2M/Transforms/GridSelection.cpp
```diff
@@ -53,6 +59,23 @@ findMaxDimAndAspectRatio(ArrayRef<int64_t> physicalShape) {
   return {maxDimIndex, aspectRatio};
 }
 
+static llvm::SmallVector<int64_t> computeOptimalTTNNCompatibleBlockShardedGrid(
```
vwellsTT: It's not clear to me exactly how this logic differs from the regular, non-TTNN path, does that also want to select largest factor within grid bounds?

lib/Dialect/D2M/Transforms/GridSelection.cpp
```diff
@@ -442,24 +499,176 @@ static void optimizeToLayoutGrid(d2m::ToLayoutOp toLayoutOp,
...
+  auto streamOp = builder.create<d2m::StreamLayoutOp>(
+      castOp.getLoc(), streamOutputTensor, castOp.getResult(), storageOp);
+  castOp.getResult().replaceAllUsesExcept(streamOp.getResult(), streamOp);
+}
+
+static void
+optimizeTTNNMetalLayoutCastOpGrid(ttir::TTNNMetalLayoutCastOp castOp,
+                                  ArrayRef<int64_t> optimalGrid,
+                                  OpBuilder &builder) {
+
```
vwellsTT: nit: spurious \n?

lib/Dialect/D2M/Transforms/GridSelection.cpp
```diff
@@ -442,24 +499,176 @@ static void optimizeToLayoutGrid(d2m::ToLayoutOp toLayoutOp,
...
+                                      bool isVirtualGrid, OpBuilder &builder) {
+  if (auto viewOp = operand.getDefiningOp<d2m::ViewLayoutOp>()) {
+    auto originalOperand = viewOp.getInput();
+    viewOp.getResult().replaceAllUsesWith(originalOperand);
+    viewOp.erase();
+    operand = originalOperand;
+  }
+  // Do not "restream" metal -> ttnn -> metal sequences. This happens when the
+  // output of a generic is the input to another generic. The output is
+  // already streamed, but the cast back to ttnn silently erases the index
```
vwellsTT, line 516: >  but the cast back to ttnn silently erases the index map

I don't think that's legal, doesn't that mean we don't actually have the tensor in proper form?  I think it might require something akin to the MaterializeViewReturns pass for this case

lib/Dialect/D2M/Transforms/GridSelection.cpp
```diff
@@ -578,96 +787,128 @@ normalizeOperandGridsForGeneric(
...
   OpBuilder builder(toLayoutsToUpdate.front().op->getContext());
   for (auto &info : toLayoutsToUpdate) {
-    optimizeToLayoutGrid(info.op, targetGridShape, targetSquareGridShape,
-                         info.grid, info.isVirtualGrid, builder);
+    optimizeToLayoutGrid(info.op, config, info.grid, info.isVirtualGrid,
+                         builder);
+  }
+}
+
+// Phase 2b: Update TTNN tensors with their optimal grids.
```
vwellsTT: nit: doesn't really seem related to "phase 2", should probably just re-number with this as phase 3

lib/Dialect/D2M/Transforms/GridSelection.cpp
```diff
@@ -578,96 +787,128 @@ normalizeOperandGridsForGeneric(
...
-          emptyOpsToUpdate};
+  bool hasTTNNOperand =
+      llvm::any_of(genericOp->getOpOperands(), [](OpOperand &operand) {
+        return isTTNNOperand(operand.get());
+      });
+  if (config.ttnnMode && hasTTNNOperand) {
+    // Propagate normalized grids back into update plans. This is required for
+    // generic ops with DRAM TTNN tensors and L1 outputs. DRAM interleaved
+    // tensors do not support virtual grid reblocking, so we need to make sure
+    // the output grid is set to the normalized grid.
```
vwellsTT: I'm a bit confused why this is need, isn't the normalized grid already aligned w/ output grid?

lib/Dialect/D2M/Transforms/Allocate.cpp
```diff
@@ -1482,6 +1509,11 @@ class D2MAllocate final : public impl::D2MAllocateBase<D2MAllocate> {
       }
     };
 
+    // Non-trivial views need a stream to represent the implied data movement.
+    if (isNonTrivialView(operandCtx)) {
```
bgrady-tt: nit: since having non-trivial view and _not_ being in DMA only form overrides all other considerations, I would move this above the move complex scratch logic.

lib/Dialect/D2M/Transforms/Allocate.cpp
```diff
@@ -1193,7 +1191,7 @@ class D2MAllocate final : public impl::D2MAllocateBase<D2MAllocate> {
 
         if (!operandCtx.hasStream &&
             (useAlwaysStreamPolicy() ||
-             inferStreamRequirement(genericOp, operandCtx.operandIndex()))) {
```
bgrady-tt: optional nit: Might as well completely encapsulate this composite check as its quite fiddly and duplicated.
```cpp
       if (!operandCtx.hasStream &&
            (useAlwaysStreamPolicy() ||
             inferStreamRequirement(genericOp, operandCtx)))

       // becomes
       if (operandNeedsStream(genericOp, operandCtx))
```

lib/Dialect/D2M/IR/D2MOps.cpp
```diff
@@ -68,11 +68,12 @@ mlir::LogicalResult d2m::EmptyOp::bufferize(
   }
 
   // Don't bufferize if tensor has a ttnn_layout; lowering to ttnn generic.
-  if (options.allowUnknownOps &&
-      (mlir::isa<ttnn::TTNNLayoutAttr>(getResult().getType().getEncoding()) ||
-       mlir::isa<ttnn::TTNNNDLayoutAttr>(
-           getResult().getType().getEncoding()))) {
-    return success();
+  if (options.allowUnknownOps) {
```
bgrady-tt, line 97: nit: Would add helper to consolidate redundancy here and at line 120

lib/Dialect/D2M/IR/D2MOps.cpp
```diff
@@ -1065,22 +1069,24 @@ d2m::ViewLayoutOp::getBufferType(
...
+    return resultView.getAffineMap();
   }
+  return mlir::cast<ttcore::MetalLayoutAttr>(
+             mlir::cast<RankedTensorType>(getType()).getEncoding())
+      .getIndexAffineMap();
+}
 
-  auto resultLayout = mlir::cast<ttcore::MetalLayoutAttr>(
-      mlir::cast<RankedTensorType>(getType()).getEncoding());
+bool d2m::ViewLayoutOp::isReblockOnly() {
```
bgrady-tt, line 1222: This probably interacts with @vwellsTT PR, which tries to keep virtual grid fwd and inv affine maps seperate from views.

lib/Dialect/D2M/Transforms/GridSelection.cpp
```diff
@@ -442,24 +499,176 @@ static void optimizeToLayoutGrid(d2m::ToLayoutOp toLayoutOp,
...
+  if (operand.getDefiningOp<ttir::TTNNMetalLayoutCastOp>()) {
+    return true;
+  }
+  if (auto view = operand.getDefiningOp<d2m::ViewLayoutOp>()) {
+    return view.getInput().getDefiningOp<ttir::TTNNMetalLayoutCastOp>();
+  }
+  return false;
+}
+
+static void insertStreamForTTNNTensor(Value operand,
```
bgrady-tt: This needs to be updated to align with @vtangTT change to use views instead of streams here https://github.com/tenstorrent/tt-mlir/pull/7122. This is likely landing sometime very soon. All other affine map logic is mostly the same.

sgholamiTT:
This probably needs a rebase after [this PR](https://github.com/tenstorrent/tt-mlir/pull/7007) was merged. I suspect the following tests might fail (?): 
- [d2m_optimizer_fork_join](https://github.com/tenstorrent/tt-mlir/blob/main/test/ttmlir/Dialect/TTNN/optimizer/d2m_optimizer_fork_join.mlir)
- [d2m_optimizer_linear_chain](https://github.com/tenstorrent/tt-mlir/blob/main/test/ttmlir/Dialect/TTNN/optimizer/d2m_optimizer_linear_chain.mlir)

Let me know if you needed help rebasing. The fix for the test might involve some changes to [this](https://github.com/tenstorrent/tt-mlir/blame/main/lib/Dialect/TTNN/Transforms/OptimizerPasses/Optimizer.cpp#L115-L140) logic.

lib/Dialect/D2M/Transforms/GridSelection.cpp
bgrady-tt: For my own understanding, implicit padding is disallowed here because?

lib/Conversion/D2MToTTNN/D2MToTTNN.cpp
```diff
@@ -25,6 +28,108 @@
...
+
+  ttcore::GridAttr grid;
+  ttnn::TensorMemoryLayout memLayoutEnum;
+  ArrayRef<int64_t> gridShape = deviceLayout.getGridShape(memrefType);
+
+  if (mlir::isa<ttcore::InterleavedLayoutAttr>(memrefType.getLayout())) {
+    grid = ttcore::GridAttr::get(ctx, SmallVector<int64_t>(2, 1));
+    memLayoutEnum = ttnn::TensorMemoryLayout::Interleaved;
+  } else {
+    auto virtMap = d2m::utils::getVirtualGridMapping(memrefValue);
```
vmilosevic: :warning: **clang\-diagnostic\-error** :warning:
no member named `` getVirtualGridMapping `` in namespace `` mlir::tt::d2m::utils ``

lib/Dialect/D2M/Transforms/GridSelection.cpp
arminaleTT: I misunderstood the non-ttnn path! Removed the special ttnn path method

lib/Conversion/D2MToTTNN/D2MToTTNN.cpp
arminaleTT: No it is not! It just takes a `Value` instead of needing a `Type`

lib/Dialect/D2M/Transforms/GridSelection.cpp
arminaleTT: This was necessary pre-rebase as the generic and operand grids could end up out of sync. Confirmed no longer needed

lib/Dialect/D2M/Transforms/GridSelection.cpp
arminaleTT: This happened/happens when dealing with DRAM intermediates. I suspect this code is no longer relevant as the problematic structure might not be created at all. I will verify that and remove this if possible, in another PR

vwellsTT (approved):
thanks for addressing prior feedback, LGTM!

lib/Dialect/D2M/Transforms/GridSelection.cpp
vwellsTT: I suppose technically it might be worth ticketing this w/ TODO as well, but don't feel strongly

lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
```diff
@@ -298,6 +298,14 @@ class D2MNamedRewriterCommon {
           rewriter.getContext(), logicalShape, ttcore::OOBVal::Undef, memSpace,
           ttcore::TensorMemoryLayout::Sharded, emptyCollapseIntervals);
 
+    } else if (ttnnMode) {
```
bgrady-tt, line 301: Would add a comment here describing why collapsedIntervals have different defaults in TTNN mode.

lib/Dialect/D2M/Transforms/Allocate.cpp
```diff
@@ -933,106 +934,100 @@ class D2MAllocate final : public impl::D2MAllocateBase<D2MAllocate> {
...
+        // This decision variable must be bound to its incoming memspace
+        // in any of these cases:
+        //  - if it is placed in DRAM *explicitly*;
+        //  - if the incoming IR indicates that this alloc should be pinned
+        //    to its current memspace in any other explicit way (aggregated
+        //    into `isMemspaceBound`);
+        //  - if it is the output of a generic op and the enabled pass
+        //    options do not allow output spilling;
+        //  - (edge case) if it has zero generic op users;
+        const bool bound =
```
bgrady-tt, line 960: The logic at this point on origin/main uses memref context's `usedAsScratchInput` field to bind scratch inputs to initial L1 memspace (see https://github.com/tenstorrent/tt-mlir/pull/7276). I think this needs to be rebased; as-is it seems like it is blowing away this bug fix.

arminaleTT: Looking into it

bgrady-tt: My mistake, I forgot how this PR works. The logic is preserved correctly 😓

lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
bgrady-tt: this is a nit, I won't block PR over this.

PR #7148, opened 2026-02-22
[D2M] Don't insert mcast barrier

vtangTT:
### Ticket
Link to Github Issue

### Problem description
Multicasts are done on same noc, same VC so should not need the barrier to guarantee order.

### What's changed
- don't insert write barrier for mcasts
- set linked=true for mcasts so that we keep the mcast link

### Checklist


nsmithtt:
Just to confirm, we can safely do this because we have 2 mcasts back to back?
1. Mcast the tensor data to cores
2. Mcast the semaphore update

And the semaphore update will be linked

vtangTT:
> And the semaphore update will be linked

Yes that's correct.

Issue #6803, opened 2026-01-29
Add fusing pattern for fused_rms_minimal


mvasiljevicTT:
Multichip models (tp nxm) can't use rms_norm op that we currently support because they reduce over multiple chips.

When `fused_rms_minimal` is added, we need to add fusing pattern to match it.

odjuricicTT:
@mvasiljevicTT Let's profile the current state of llama 70b layer and see how much impact this would have before starting work on this.

odjuricicTT:
Also https://github.com/tenstorrent/tt-metal/issues/37746

mvasiljevicTT:
Decomposed rms norm has these device ops:

<img width="1998" height="694" alt="Image" src="https://github.com/user-attachments/assets/9454b639-eb71-4e84-b7ab-e8b2370f4ca9" />

running for 1500us, out of ~4000us for the whole layer, having 4 FillPad ops out of 6 in the whole layer.

Fused op in different scenarios:
- no trace 1 iter - 200us
- trace 1 iter - 108us
- trace 1, 100 iter avg - 70us
- trace 100, avg - 30us

In all scenarios decomposed op had similar ~1500us.

Script for the analysis on [this branch](https://github.com/tenstorrent/tt-metal/compare/mvasiljevic/fused_rms_profile).

mvasiljevicTT:
Comparison of 5 layer model with last AG removed, decomposed vs [fused](https://github.com/tenstorrent/tt-mlir/tree/mvasiljevic/distributed_rsm_fusing) DistributedRMSNorm on llama 70b tp 2x4: 

<p align="center">
  <img src="https://github.com/user-attachments/assets/c92111bb-3f4f-432e-9356-650ba9ba73d0" width="48%" />
  <img src="https://github.com/user-attachments/assets/f9970a8a-a5db-451c-8ae3-78efd4f14955" width="48%" />
</p>

~~leading to 2.53 tps on local machine (current nightly on CI had 1.66)~~ this measuring was done with local change of always using local sum fallback for all reduce workaround. We should check how that alone affects perf. Bars are using original workaround.

[perf benchmark for tp 2x4 llama](https://github.com/tenstorrent/tt-xla/actions/runs/22854425936) : 1.66 ->
[perf benchmark for galaxy llama](https://github.com/tenstorrent/tt-xla/actions/runs/22851010842) : 0.85 -> 1

mvasiljevicTT:
[...]

PR #7152, opened 2026-02-23
[TTNNWorkarounds] Change splitting axis logic in AllReduceWorkarounds

mvasiljevicTT:
### Ticket
[Link to Github Issue](https://github.com/tenstorrent/tt-xla/issues/3407)

### Problem description
Currently, splitting axis is determined as `last tensor dimension which is divisible by number of devices along the cluster axis on which we are performing the all reduce`.

In the `pmean` example currently hanging on tt-xla, there is a tiled tensor of shape `8192, 392`, and mesh `1,2`. 
In the current implementation `reduce_scatter` would use splitting axis -1 as `392` is divisible by 2, but it's number of tiles, `13`, is not, which is probable cause for ccl hang.

### What's changed
Logic in `AllReduceWorkarounds` is changed to use tiled shape for tiled tensors when looking for last divisible dimension.

If none of the dims are divisible there is no safe splitting axis for `reduce_scatter`, and  `rewriteAsAllGatherLocalReduce` is used instead of `rs+ag` decomposition.

When chosen dim doesn't have full tiles, after `reduce_scatter` devices may end up with different shape, which is why we need to pad the input before `reduce_scatter`, and then slice it after `all_gather`:
```
|x x|  |y y|         |x x|  |y y|         |a a|               |a a|  |a a|          |a a|  |a a|
|x x|  |y y|   pad   |x x|  |y y|    rs   |a a|          ag   |a a|  |a a|   slice  |a a|  |a a|
|x x|  |y y|    ->   |x x|  |y y|    ->         |b b|    ->   |b b|  |b b|    ->    |b b|  |b b|
                     |0 0|  |0 0|               |0 0|         |0 0|  |0 0|        
```

### Testing

All previously passing perf benchmark tests are working with this change.

Some models now have more local sums and less RS ops. For example, RMSNorm in llama 70b decode had AR with <32,1> tensor which will now be changed to AG + local sum. This is not affecting perf as the tensors are small anyway. Larger tensors have hidden dim that is usually power or 2, or divisible by 2 many times, so this won't affect them either.

### Checklist
- [x] [pmean tt-xla test](https://github.com/tenstorrent/tt-xla/actions/runs/22312591630/job/64549995718)
- [x] [metal issue](https://github.com/tenstorrent/tt-metal/issues/38538)
- [x] [perf benchmark](https://github.com/tenstorrent/tt-xla/actions/runs/22479428361)

### DIFF
--- a/lib/Dialect/TTNN/Transforms/Workarounds/TTNNWorkaroundsPatterns.cpp
+++ b/lib/Dialect/TTNN/Transforms/Workarounds/TTNNWorkaroundsPatterns.cpp
@@ -439,10 +439,8 @@ public:
       llvm::SmallVector<int64_t> paddedShape(inputShape.begin(),
                                              inputShape.end());
       paddedShape[selectedDim] = tilePaddedShape[selectedDim];
-      auto paddedType = RankedTensorType::get(
-          paddedShape, inputType.getElementType(),
-          mlir::cast<ttnn::TTNNLayoutAttr>(inputType.getEncoding())
-              .withTensorShape(paddedShape));
+      auto paddedType =
+          ttnn::utils::RankedTensorTypeFactory::create(inputType, paddedShape);
 
       reduceScatterInput = rewriter.create<ttnn::PadOp>(
           ttmlir::utils::appendLocationSuffix(loc, "_pad_for_reduce_scatter"),


lib/Dialect/TTNN/Transforms/Workarounds/TTNNWorkaroundsPatterns.cpp
```diff
@@ -390,59 +390,113 @@ class TTNNAllReduceWorkarounds : public OpRewritePattern<ttnn::AllReduceOp> {
...
+      padding[selectedDim * 2 + 1] =
+          tilePaddedShape[selectedDim] - inputShape[selectedDim];
+
+      llvm::SmallVector<int64_t> paddedShape(inputShape.begin(),
+                                             inputShape.end());
+      paddedShape[selectedDim] = tilePaddedShape[selectedDim];
+      auto paddedType = RankedTensorType::get(
+          paddedShape, inputType.getElementType(),
+          mlir::cast<ttnn::TTNNLayoutAttr>(inputType.getEncoding())
+              .withTensorShape(paddedShape));
```
azecevic000: You can use `ttnn::utils::RankedTensorTypeFactory::create` as a safer option here.

mvasiljevicTT:
Pausing this PR for now, discussion in [the issue](https://github.com/tenstorrent/tt-xla/issues/3407).

lib/Dialect/TTNN/Transforms/Workarounds/TTNNWorkaroundsPatterns.cpp
mvasiljevicTT: Thanks, changed now

mvasiljevicTT:
CI rerun:
- [ ] [perf benchmark llmbox](https://github.com/tenstorrent/tt-xla/actions/runs/22908870468)
- [x] [perf benchmark galaxy](https://github.com/tenstorrent/tt-xla/actions/runs/22908881872)
- [x] [galaxy benchmark before change](https://github.com/tenstorrent/tt-xla/actions/runs/22910764440/job/66482451307)

There is perf boost 0.85 -> 1.0 on galaxy llama when small tensors ([32,1]) from rms norm use local sum instead of reduce scatter, which is same boost we get with [distributed rms norm fusing](https://github.com/tenstorrent/tt-mlir/issues/6803)

Issue #6983, opened 2026-02-10
[D2M] DRAM reads should be scheduled to preferred datamovement thread


bgrady-tt:
- DRAM reads are most efficient when performed on NoC0 (region ^datamovement0)
- DRAM writes are most efficient when performed on NoC1 (region ^datamovement1)

The `D2MScheduleDMA` pass should treat DRAM memspace operands as a special case and schedule all DRAM reads (remote load) to datamovement0 region and all DRAM writes (remote store) to the datamovement1 region. 

The remaining L1 memspace loads and stores (if any) should be round-robin allocated between datamovement0 and datamovement1 as they are now.

Issue #6790, opened 2026-01-28
[D2M/TTNN] Use DataMovementConfigDescriptor for kernels


vtangTT:
### Describe the bug

Right now, `D2MToTTNN` assigns first datamovement kernel in the IR to `ReaderConfigDescriptor` and second to `WriterConfigDescriptor`. Tracing down to metal, this will assign to noc1 and noc0 respectively.

This is not aligned with `D2MToTTMetal`.

Instead, get rid of the notion of "reader" and writer". The first datamovement kernel goes to noc0 and second goes to noc1. Use `DataMovementConfigDescriptor` instead to have the flexibility on which noc to assign.

This already exists in tt-metal's `ProgramDescriptor` but will need to be wired up into ttnn dialect and fbs.

PR #7167, opened 2026-02-24
[D2M] Assign correct noc for DRAM loads/stores

vtangTT:
### Ticket
https://github.com/tenstorrent/tt-mlir/issues/6983
https://github.com/tenstorrent/tt-mlir/issues/6790

### Problem description
As described in issues above

### What's changed
D2M changes: 
- new `nocIndex` param on `D2M_ThreadAttr` that indicates which NoC to assign this dm thread to
- new helper `assignNocIndices` in `ScheduleDMA`: scoring prioritizes giving NoC0 to the thread with the largest multicast remote load
- `D2MToTTNN` and `D2MToTTMetal` changes to query this index instead of using arbitrary counter

TTNN changes:
- new dialect and fbs attr that mirrors `DataMovementConfigDescriptor` in [metal](https://github.com/tenstorrent/tt-metal/blob/e909aef5ae3453629ffa1453e9757b1ee007c99f/tt_metal/api/tt-metalium/program_descriptors.hpp#L81C8-L81C36)
- necessary toflatbuffer and generic op runtime changes

### Checklist

### DIFF
--- a/lib/Dialect/D2M/IR/D2MOps.cpp
+++ b/lib/Dialect/D2M/IR/D2MOps.cpp
@@ -1699,28 +1699,6 @@ MutableArrayRef<OpOperand> d2m::GenericOp::getInputsAndOutputsMutable() {
     return emitOpError("number of regions must match the number of threads");
   }
 
-  // Can only have one dm thread per NoC.
-  bool nocUsed[2] = {false, false};
-  for (auto thread : getThreads()) {
-    auto threadAttr = mlir::cast<ThreadAttr>(thread);
-    if (threadAttr.getThreadType() != ThreadType::Datamovement) {
-      continue;
-    }
-    int32_t nocIndex = threadAttr.getNocIndex();
-    if (nocIndex < 0) {
-      continue;
-    }
-    if (nocIndex > 1) {
-      return emitOpError("invalid noc index ")
-             << nocIndex << ", must be 0 or 1";
-    }
-    if (nocUsed[nocIndex]) {
-      return emitOpError("multiple datamovement threads assigned to noc ")
-             << nocIndex;
-    }
-    nocUsed[nocIndex] = true;
-  }
-
   // Output grid shape must equal the GenericOp grid shape.
   auto opGridShape = getGrid().getShape();
   for (auto output : getOutputs()) {


include/ttmlir/Dialect/TTNN/IR/TTNNOpsEnums.td
```diff
@@ -345,6 +345,18 @@ def TTNN_KernelCoreType : I32EnumAttr<"KernelCoreType", "TT KernelCoreType",
   let cppNamespace = "::mlir::tt::ttnn";
 }
 
+def TTNN_DataMovementProcessorRiscV0 : I32EnumAttrCase<"RiscV0", 0, "riscv0">;
```
bgrady-tt, line 349: nit: Would leave comment here for which is BRISC vs NCRISC.

test/ttmlir/Dialect/D2M/Transforms/schedule_dma.mlir
```diff
@@ -367,7 +367,7 @@ module {
     %alloc = memref.alloc() {alignment = 64 : i64} : memref<2x4x2x4x!ttcore.tile<32x32, f32>, #ttcore.shard<16384x4096, 1>, #l1>
 
     // CHECK: d2m.generic
-    // CHECK-SAME: threads = [#d2m.thread<datamovement>, #d2m.thread<datamovement>, #d2m.thread<compute>]
+    // CHECK-SAME: threads = [#d2m.thread<datamovement, noc = 0>, #d2m.thread<datamovement, noc = 0>, #d2m.thread<compute>]
```
bgrady-tt: Shouldn't the two datamovement threads have different assignments here?

include/ttmlir/Dialect/D2M/IR/D2MOpsAttrs.td
```diff
@@ -19,17 +19,24 @@ def D2M_ThreadAttr : AttrDef<D2M_Dialect, "Thread", [], "::mlir::Attribute"> {
   let summary = "Thread information for a generic op.";
   let description = [{
     Holds the thread information corresponding to a single generic op region.
+    For datamovement threads, nocIndex indicates the assigned NoC (0 or 1).
+    A value of -1 means unassigned.
```
bgrady-tt, line 23: We should validate noc assignments in `GenericOp::verify()` in D2MOps.cpp. For now, it shouldn't be possible to assign >1 thread to a given noc index. Also dynamic mode should be forbidden.

lib/Dialect/D2M/Transforms/ScheduleDMA.cpp
```diff
@@ -194,28 +244,45 @@ class D2MScheduleDMARewriter : public OpRewritePattern<GenericOp> {
...
+    // Not enough CBs to warrant splitting but still need to assign nocIndex on
+    // the existing single DM thread before returning failure.
+    if (numThreadsToUse <= 1 || cbWorkloads.size() <= 1) {
+      auto origAttr =
+          mlir::cast<ThreadAttr>(generic.getThreadsAttr().getValue()[0]);
+      if (origAttr.getNocIndex() != -1) {
+        return failure();
+      }
+
+      bool hasDRAMRead = llvm::any_of(dmaOps, [](const auto &entry) {
```
bgrady-tt: It seems like the logic for determining noc index based on if it is DRAM read/write is duplicated both here and in assignNoCIndices? Would be good to unify this to keep it from getting out of sync.

include/ttmlir/Dialect/D2M/IR/D2MOpsAttrs.td
vtangTT: added into d2m.generic verifier.. also added the verifier for the ttnn datamovement attr to not allow dynamic mode

lib/Dialect/D2M/Transforms/ScheduleDMA.cpp
vtangTT: commonized 🙏

include/ttmlir/Dialect/TTNN/IR/TTNNOpsEnums.td
vtangTT: commented

lib/Dialect/D2M/Transforms/ScheduleDMA.cpp
```diff
@@ -258,9 +269,12 @@ class D2MScheduleDMARewriter : public OpRewritePattern<GenericOp> {
     // Not enough CBs to warrant splitting but still need to assign nocIndex on
     // the existing single DM thread before returning failure.
     if (numThreadsToUse <= 1 || cbWorkloads.size() <= 1) {
-      SmallVector<unsigned, 2> dramReadCBs, dramWriteCBs;
-      getDRAMAccessCBs(dmaOps, dramReadCBs, dramWriteCBs);
-      int nocIndex = !dramWriteCBs.empty() ? 1 : 0;
+      bool writesDRAM = llvm::any_of(dmaOps, [](const auto &entry) {
+        auto store = mlir::dyn_cast<RemoteStoreOp>(entry.first);
```
bgrady-tt: would use `dyn_cast_or_null` here

lib/Dialect/D2M/Transforms/ScheduleDMA.cpp
```diff
@@ -59,58 +59,69 @@ collectDMAOps(Block *block,
...
+
+      if (auto load = mlir::dyn_cast<RemoteLoadOp>(op)) {
+        if (ttcore::getMemorySpace(load.getMemref()) !=
+            ttcore::MemorySpace::DeviceDRAM) {
+          continue;
+        }
+        if (load.isMcast()) {
+          auto layout = ttcore::getDeviceLayout(load.getMemref());
+          if (layout) {
+            auto shardShape = SmallVector<int64_t>(
```
bgrady-tt: worth prompting to see if there is an existing util for computing overall shard size (in bytes); I think there might be?

lib/Dialect/D2M/Transforms/ScheduleDMA.cpp
bgrady-tt, line 126: It is not clear to me how we ensure multicast loads are distributed fairly to both threads? We should at least validate that each thread has at max 1 multicast load to avoid hangs.

bgrady-tt (commented):
I think some more paranoid checks are needed to ensure multiple multicast loads don't end up on the same NoC.

lib/Dialect/D2M/Transforms/ScheduleDMA.cpp
vtangTT: fixed 👍

lib/Dialect/D2M/Transforms/ScheduleDMA.cpp
vtangTT: There is, now using the helper `DeviceAttr::getShardSizeInBytes`

lib/Dialect/D2M/Transforms/ScheduleDMA.cpp
vtangTT: Agreed with being extra cautious. I've asserted inside `assignNocIndices` to check this case. 

Note: in the future, once kernels get more complicated I'm unsure whether such logic would belong in `assignNocIndices` or upstream in `assignCBsToThread` or perhaps another helper..

lib/Dialect/D2M/IR/D2MOps.cpp
```diff
@@ -1699,6 +1699,28 @@ ::mlir::LogicalResult d2m::GenericOp::verify() {
...
+    int32_t nocIndex = threadAttr.getNocIndex();
+    if (nocIndex < 0) {
+      continue;
+    }
+    if (nocIndex > 1) {
+      return emitOpError("invalid noc index ")
+             << nocIndex << ", must be 0 or 1";
+    }
+    if (nocUsed[nocIndex]) {
+      return emitOpError("multiple datamovement threads assigned to noc ")
```
nsmithtt: This is technically OK if you wanted to assign both DM threads to the same noc.  It will also be common for this to happen in quasar, I think it might be better to remove this check.

vtangTT: This check was added as it was observed two DM threads on same noc would hang if both are multicasting (with a barrier afterwards). See [issue](https://github.com/tenstorrent/tt-mlir/issues/7297) with an IR dump.

@bgrady-tt and I agreed that we should assert this case for now.

nsmithtt: We'll have to remove this check in a month or so when we start on quasar.  Perhaps we need a more specific check that explicitly looks for this mcast case.

vtangTT: I'm already asserting in `ScheduleDMA` that we are not assigning two multicasts to the same thread, so I can remove the check here.. wdyt @bgrady-tt

bgrady-tt: Its ok to let this go I think, as long as multicast is being verified.

bgrady-tt (approved):
LGTM 👍

lib/Dialect/D2M/IR/D2MOps.cpp
vtangTT: removed

PR #7169, opened 2026-02-24
Remove traceback log from eth connections

pglusacTT:
### Ticket
N/A

### Problem description
Fetching device connections displayed the entire metal traceback. We originally included this to help identify false negatives (where an eth connection exists, but the metal API can't use it). Since we haven't encountered this scenario yet, the traceback was just causing confusion.

### What's changed
Remove traceback logging in system descriptor creation.

### Checklist

### DIFF
--- a/runtime/lib/common/system_desc.cpp
+++ b/runtime/lib/common/system_desc.cpp
@@ -74,7 +74,7 @@ getAllDeviceConnections(const std::vector<::tt::tt_metal::IDevice *> &devices) {
         connectedDevice = device->get_connected_ethernet_core(ethernetCore);
       } catch (const std::runtime_error &e) {
         LOG_WARNING("Skipping ethernet core (", ethernetCore.str(),
-                    ") on chip ", device->id(), ": ", e.what());
+                    ") on chip ", device->id());
         continue;
       }
       addConnection(device->id(), ethernetCore, std::get<0>(connectedDevice),


pglusacTT:
FYI @mstojkovicTT

runtime/lib/common/system_desc.cpp
```diff
@@ -74,7 +74,7 @@ getAllDeviceConnections(const std::vector<::tt::tt_metal::IDevice *> &devices) {
         connectedDevice = device->get_connected_ethernet_core(ethernetCore);
       } catch (const std::runtime_error &e) {
         LOG_WARNING("Skipping ethernet core (", ethernetCore.str(),
-                    ") on chip ", device->id(), ": ", e.what());
+                    ") on chip ", device->id());
```
pilkicTT, line 77: but isn't `tt-metal` logging the message as `TT_FATAL` regardless?

pglusacTT: It is, without the traceback, will be solved once the TODO above is resolved. I hope that's fine for now.

pilkicTT: yeah, it's fine, since we are waiting for proper API... but have you maybe tried setting `TT_METAL_DISABLE_BACKTRACE`, before `try-catch`? i am just looking to minimize any confusion, because there was some... :smile:

pilkicTT: oh, now i see, they are setting the backtrace in the exception... all good

Issue #6993, opened 2026-02-11
sort op doesn't use memory config


rpavlovicTT:
https://github.com/tenstorrent/tt-mlir/blob/48522b8451d31157463c6c1b4faf1da9517a1d95/lib/Target/TTNN/TTNNToFlatbuffer.cpp#L2358..L2365

We should create memory config based on the output tensor like in other ops.

PR #7186, opened 2026-02-24
Fix sort op memory config in flatbuffer

bmalesevicTT:
### Ticket
#6993 

### Problem description
In step of converting TTNN-IR to flattbuffer, the sort op was only using the memory config if it was explicitly set on the op attribute. If no explicit memory config was set, it passed 0 (null) to the flatbuffer.

### What's changed
Create a memory config based on the output tensor type (like other ops do) when no explicit memory config is set on the op.
This almost matches the standard pattern used by other ops like Pool2d, Slice, Repeat, Typecast, etc. but Sort has multiple results so we specifically take the first result ("getValues()") output tensor type. Now there is an overload to pass the result Value explicitly.

### Checklist

### DIFF
--- a/lib/Target/TTNN/TTNNToFlatbuffer.cpp
+++ b/lib/Target/TTNN/TTNNToFlatbuffer.cpp
@@ -3245,7 +3245,9 @@ createOp(FlatbufferObjectCache &cache, NLPCreateQKVHeadsDecodeOp op) {
 
                                   /*local_shape*/ std::nullopt);
 
-  auto memoryConfig = getMemoryConfigIfNeeded(cache, op, op.getQuery());
+  auto memoryConfig = op.getMemoryConfig()
+                          ? toFlatbuffer(cache, op.getMemoryConfig().value())
+                          : 0;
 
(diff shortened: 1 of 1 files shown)


lib/Target/TTNN/TTNNToFlatbuffer.cpp
```diff
@@ -2418,12 +2397,16 @@ createSortOp(FlatbufferObjectCache &cache, SortOp op) {
...
   bool stable = op.getStable();
-  std::optional<mlir::tt::ttnn::MemoryConfigAttr> memoryConfig =
-      op.getMemoryConfig();
 
-  return ::tt::target::ttnn::CreateSortOpDirect(
-      *cache.fbb, in, dim, descending, stable,
-      (memoryConfig ? toFlatbuffer(cache, memoryConfig.value()) : 0), &outputs);
+  // Could not use template function since SortOp has multiple results with
+  // specific getters
+  auto memoryConfig =
```
sdjordjevicTT: Can we, instead of scattering this logic all over the place, improve our template function or create a new one that will help with ops that have more than one output?

Issue #750, opened 2024-09-18
[LLAMA] Implement conversion of stablehlo.image to TTIR 


mmanzoorTT:
`stablehlo.imag` OP specification can be found [here](https://openxla.org/stablehlo/spec#imag).

mmanzoorTT:
@mrakitaTT Do we need this for LLama for tt-xla?

mrakitaTT:
Sorry for late reply, I've checked the SHLO graph from the `openllama_3b_v2` test and `stablehlo.imag` doesn't appear in that graph, so I would say this task is of a low priority.

Also, while ttnn [has a support](https://docs.tenstorrent.com/tt-metal/latest/ttnn/ttnn/api/ttnn.imag.html#ttnn.imag) for `ttnn.imag`, we don't support complex data types in TTIR yet.

ddilbazTT:
https://github.com/tenstorrent/tt-mlir/pull/7407

Issue #749, opened 2024-09-18
[LLAMA] Implement conversion of stablehlo.real to TTIR 


mmanzoorTT:
`stablehlo.real` OP specification can be found [here](https://openxla.org/stablehlo/spec#real).

AleksKnezevic:
duplicate of https://github.com/tenstorrent/tt-xla/issues/3060

Issue #748, opened 2024-09-18
[LLAMA] Implement conversion of stablehlo.complex to TTIR


mmanzoorTT:
`stablehlo.complex` OP specification can be found [here](https://openxla.org/stablehlo/spec#complex).

ddilbazTT:
@nsmithtt @uazizTT @mmanzoorTT Am I supposed to be creating a new file under `runtime/lib/ttnn/operations/eltwise` for complex operations? Or should I add a complexEltwiseBinaryOp under binary.cpp? I am assuming this function will use ComplexTensor struct defined in `third_party/tt-metal/src/tt-metal/ttnn/cpp/ttnn/operations/eltwise/complex/complex.hpp`

ddilbazTT:
Hi all, I need help because "Tensor" and "ComplexTensor" structs are not interchangeable with the way the codebase is designed right now. It seems implementing ComplexOp needs an extensive code refactoring.

I created a folder under `runtime/lib/ttnn/operations` named `complex` - which includes a header and a cpp file. [(Commit Link)](https://github.com/tenstorrent/tt-mlir/commit/d5693ce70527ec7c016d5a5656022e162bf66811) The most basic purpose of this commit is to include a ComplexTensor alternative of `run`. For example, this is the `run` operation for "concat" (from `runtime/lib/ttnn/operations/data_movement/concat.cpp`).

```
void run(const ::tt::target::ttnn::ConcatOp *op, ProgramContext &context) {
  ProgramTensorPool &tensorPool = context.getTensorPool();
  std::vector<::ttnn::Tensor> inputs;
  for (const auto &input : *op->inputs()) {
    inputs.push_back(tensorPool.at(input->global_id()));
  }
  int32_t dim = op->dim();
  ::ttnn::Tensor out = ::ttnn::concat(inputs, dim);
  tensorPool.insert_or_assign(op->out()->global_id(), out);
}
```

My `void run(const ::tt::target::ttnn::ComplexOp *op, ProgramContext &context)` is not correct, and this is related to needed codebase changes. I will explain the dilemma below.

If you look at `/runtime/lib/ttnn/runtime.cpp`, you will see the function 
```
Event submit(Device deviceHandle, Binary executableHandle,
             std::uint32_t programIndex,
             std::vector<Tensor> const &inputHandles,
             std::vector<Tensor> const &outputHandles) 
```
which invokes 
```
void runProgram(::ttnn::MeshDevice &meshDevice,
[...]

PR #7197, opened 2026-02-25
[DRAFT][PoC] Add support for complex data type and stablehlo.complex, stablehlo.real, stablehlo.imag

ddilbazTT:
### Ticket
#748 
#749 
#750 
https://github.com/tenstorrent/tt-xla/issues/3060 

### Problem description
In order to bring up Deepseek V3.2, we need to figure out a way to represent complex data types (i.e. complex<f32>) and add support for complex operations. I decided to target [complex](https://openxla.org/stablehlo/spec#complex), [real](https://openxla.org/stablehlo/spec#real), and [imaginary](https://openxla.org/stablehlo/spec#imag) ops as others can use these ops when decomposed.  

I decided to lower complex operations into arithmetic operations during TTIR as I do not think we should use tt-metal complex support. These are the following reasons why:
- tt-metal does not have proper complex op support. For example, multiplication is not supported.
- In ttnn, ComplexTensor is a struct resembling a tuple:
```
struct ComplexTensor {
    std::array<Tensor, 2> m_real_imag;

    ComplexTensor(const std::tuple<const Tensor&, const Tensor&>& real_imag);

    const Tensor& operator[](uint32_t index) const;
    const Tensor& real() const;
    const Tensor& imag() const;
    void deallocate();
};
```

During runtime, we cannot represent `mlir::complex` as a `ComplexTensor` without a lot of changes and even if we did, we wouldn't have proper op coverage in tt-metal anyways. For deepseek, we will eventually need support for add, div, multiply ops.

Based on pjrt, we can understand what the buffers for complex data types look like:

```
  // Paired F32 (real, imag), as in std::complex<float>.
  PJRT_Buffer_Type_C64,
  // Paired F64 (real, imag), as in std::complex<double>.
  PJRT_Buffer_Type_C128,
```

I also printed the buffers to understand how complex values are represented in memory. Basically, real and imaginary values are interleaved: `[re0 im0 re1 im1 re2 im2 ... ]` This means that the following two can be represented almost equally and with some PJRT runtime adjustments, would work 1:1
tensor<1x4xcomplexf32> -> tensor<1x4x2xf32> 
When looking at the last dimension, 0 corresponds to real values and 1 corresponds to imaginary values. 

The pjrt changes for this type conversion are in this branch: https://github.com/tenstorrent/tt-xla/compare/main...ddilbaz/complex

Having decided that complex data types will be rewritten with existing data types and 2 appended as last dim, I decided on the following:

- In TTIR, we will have two versions of complex, real, and imaginary ops. One version corresponds to operations using `mlir::complex`, the other corresponds to equivalent representation with last dim 2 appended.
For example, there will be these two that are converted to one another during ttir to ttir decompositions:
```
%0 = "ttir.stablehlo_complex"(%arg0, %arg1) : (tensor<2x4xf64>, tensor<2x4xf64>) -> tensor<2x4xcomplex<f64>>
%0 = "ttir.complex"(%arg0, %arg1) : (tensor<2x4xf64>, tensor<2x4xf64>) -> tensor<2x4x2xf64>
```
- In TTIRComplexDataTypeConversionPass, we will replace complex data types with unpacked versions and replace stablehlo equivalent complex ops with new complex ops using the new data standard.
- In TTIRToTTNN conversion, we will convert complex operations to existing TTNN operations. I will explain conversions.
    - complex op conversion:
```
%0 = "ttir.complex"(%arg0, %arg1) : (tensor<2x4xf32>, tensor<2x4xf32>) -> tensor<2x4x2xf32>
```
Here, we unsqueeze arguments 0 and 1, resulting in shape `tensor<2x4x1xf32>' Then we concatenate along last dim constructing the final output with shape `tensor<2x4x2xf32>`. 
    - real/ imag op conversions:
These two conversions are very similar. We will use `slice` along the last dim, resulting in a `2x4x1xf32` shape. Later, we will `reshape` to achieve `2x4xf32`

I have tested `torch.view_as_complex`, `torch.real`, `torch.imag` ops through tt-xla and they work. 

**I would appreciate feedback on this proof of concept before moving any further.**

### DIFF
--- a/test/ttmlir/Dialect/TTIR/ComplexDataTypeConversion/deepseek.mlir
+++ b/test/ttmlir/Dialect/TTIR/ComplexDataTypeConversion/deepseek.mlir
@@ -1,0 +1,27 @@
+// RUN: ttmlir-opt --ttir-complex-data-type-conversion %s
+
+func.func @main(%arg0: tensor<16x8xcomplex<f32>> {ttcore.argument_type = #ttcore.argument_type<input>, ttcore.runtime_tensor_sharding = #ttcore<runtime_tensor_sharding shard_status = <unsharded>, local_shape = tensor<16x8xcomplex<f32>>>, ttir.name = "args_1"}, %arg1: tensor<2x16x4x16xbf16> {ttcore.argument_type = #ttcore.argument_type<input>, ttcore.runtime_tensor_sharding = #ttcore<runtime_tensor_sharding shard_status = <unsharded>, local_shape = tensor<2x16x4x16xbf16>>, ttir.name = "args_0"}) -> (tensor<2x16x4x16xbf16> {ttcore.runtime_tensor_sharding = #ttcore<runtime_tensor_sharding shard_status = <unsharded>, local_shape = tensor<2x16x4x16xbf16>>}) attributes {tt.function_type = "forward_device"} {
+    %0 = "ttir.typecast"(%arg1) <{conservative_folding = false}> : (tensor<2x16x4x16xbf16>) -> tensor<2x16x4x16xf32>
+    %1 = "ttir.reshape"(%0) <{shape = [2 : i32, 16 : i32, 4 : i32, 8 : i32, 2 : i32]}> : (tensor<2x16x4x16xf32>) -> tensor<2x16x4x8x2xf32>
+    %2 = "ttir.slice_static"(%1) <{begins = [0 : i32, 0 : i32, 0 : i32, 0 : i32, 0 : i32], ends = [2 : i32, 16 : i32, 4 : i32, 8 : i32, 1 : i32], step = [1 : i32, 1 : i32, 1 : i32, 1 : i32, 1 : i32]}> : (tensor<2x16x4x8x2xf32>) -> tensor<2x16x4x8x1xf32>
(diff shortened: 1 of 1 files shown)


lib/Conversion/TTIRToTTNN/TTIRToTTNN.cpp
```diff
@@ -3235,6 +3369,9 @@ void populateTTIRToTTNNPatterns(MLIRContext *ctx, RewritePatternSet &patterns,
            DropoutOpConversionPattern,
            DebugOpConversionPattern<debug::DumpOp, ttnn::DumpTensorOp>,
            TopKOpConversionPattern
+           ComplexOpConversionPattern,
```
vmilosevic, line 3337: :warning: **clang\-diagnostic\-error** :warning:
type\-id cannot have a name
```suggestion
           ,
```

ddilbazTT:
I have discussed my proposals during our meeting and I will be closing this draft PR for the time being. Here are notes on the path forward:
- Look into [StablehloComplexMathExpanderPatterns](https://github.com/openxla/stablehlo/blob/main/stablehlo/transforms/StablehloComplexMathExpander.cpp) 
    - If this turns out to be a more practical way forward, I will scrap the proposal.
    - Currently, this seems supplementary to the proposal since it implements a handful of complex operations that are not common:
        - Complex logarithm
        - Complex square root
        - Complex exponential
- Rewrite the proposal with the following changes:
    - Move ComplexDataTypeConversionPass to Stablehlo
    - Add permutes to unpacked data or permute in PJRT runtime so that we do not have tilize/ untilize cost due to last dim being non-divisible by 16 (proposal was to unpack to 2)
- Follow-up with the following changes that need to be tested around:
    - Can we represent complex/ polar with cos and sin? Based off of Euler’s formula
    - Can we fuse this rope implementation into ttnn rotary embedding op?
    - Implement add/ multiplication decomposition in tt-mlir instead of torch decomposition.

Issue #7215, opened 2026-02-26
[ttnn.jit] nightly tests broken for certain non-square grids


vtangTT:
### Describe the bug

see https://github.com/tenstorrent/tt-mlir/actions/runs/22428033547/job/64941934518

raw log [link](https://productionresultssa1.blob.core.windows.net/actions-results/b8028c63-ea3c-4af5-94e1-4d66696428c7/workflow-job-run-148c9e9d-e309-5529-891c-3dc8e588d119/logs/job/job-logs.txt?rsct=text%2Fplain&se=2026-02-26T15%3A43%3A52Z&sig=ocMG%2BHRD7fEBv%2Bl7H3svVhZrggHZpZSX6kUx9T1gVO0%3D&ske=2026-02-26T16%3A43%3A38Z&skoid=ca7593d4-ee42-46cd-af88-8b886a2f84eb&sks=b&skt=2026-02-26T12%3A43%3A38Z&sktid=398a6654-997b-47e9-b12b-9515b896b4de&skv=2025-11-05&sp=r&spr=https&sr=b&st=2026-02-26T15%3A33%3A47Z&sv=2025-11-05)

`test/ttnn-jit/nightly` pytests are failing PCC for non-square grids, specifically `grid_(3, 1` and `grid_(1, 5)` (might be more)

Triage of possible commits on Feb25th, 2026 has shown that github.com/tenstorrent/tt-mlir/pull/6902 caused the regression.

### Steps to reproduce the issue

```
git checkout 9caeec95986d53d77341cafa68159fdf8f4ba0f0

// build with jit enabled -DTTMLIR_ENABLE_RUNTIME=ON -DTTMLIR_ENABLE_TTNN_JIT=ON

// smaller failing test suite
pytest test/ttnn-jit/nightly/test_eltwise.py::test_binary_ops_l1 -k mul -v

// larger failing test suite (superset of the first one)
pytest test/ttnn-jit/nightly/test_eltwise.py::test_binary_ops_l1 -v

PR #7239, opened 2026-02-27
[D2M] Index refactor ttnn-jit nightly fix

vwellsTT:
### Ticket
https://github.com/tenstorrent/tt-mlir/issues/7215

### Problem description
Previous PR broke some nightly tests.  Also had some functionality gaps (some pre-existing, some newly introduced).

### What's changed
The idea behind this change is that: 1). for virtual grids, we ought to carry both fwd and inv maps on the creation ops (prev PR only carried invMaps, before that PR index_map only carried fwd map), and 2). we should never "re-derive" the maps--we should always look them up.  This involves changing some verification logic etc.
Additionally, some logic for LowerLoadStoreOpsToExplicitCBForm needed updates from last PR; we allowed neither to be "remote", but did not always properly handle this case; now we ensure that non-remote operands get properly lowered to explicit form in all cases.  (This may have minor adverse perf implications?)

### Checklist
- [x] passing nightly tests https://github.com/tenstorrent/tt-mlir/actions/runs/22506471932

### DIFF
--- a/include/ttmlir/Conversion/TTNNToEmitC/EmitCConversion.h
+++ b/include/ttmlir/Conversion/TTNNToEmitC/EmitCConversion.h
@@ -2142,14 +2142,15 @@ public:
       return loadOp.getResult();
     }
 
-    // SortOp returns a std::vector<ttnn::Tensor> containing two elements:
-    // [0] = sorted tensor, [1] = corresponding indices.
-    // Extract both elements to replace the original SortOp.
-    if constexpr (std::is_same_v<TTNNOp, tt::ttnn::SortOp>) {
+    // SortOp and TopKOp returns a std::vector<ttnn::Tensor> containing two
+    // elements: [0] = values tensor, [1] = corresponding indices. Extract both
+    // elements to replace the original Op.
+    if constexpr (std::is_same_v<TTNNOp, tt::ttnn::SortOp> ||
+                  std::is_same_v<TTNNOp, tt::ttnn::TopKOp>) {
       assert(op.getNumResults() == 2 &&
-             "Expected two outputs for SortOp (sorted tensor and indices).");
+             "Expected two outputs (values tensor and indices).");
       using ReturnTy = std::vector<::ttnn::Tensor>;
-      auto sortOp = rewriter.create<emitc::CallOpaqueOp>(
+      auto callOp = rewriter.create<emitc::CallOpaqueOp>(
           op.getLoc(), rewriter.getType<emitc::OpaqueType>(TypeNameV<ReturnTy>),
           opConversionPattern.convertOpName(op), rewriter.getArrayAttr(args),
           /*template_args=*/nullptr, operands);
@@ -2168,7 +2169,7 @@ public:
 
         // Get reference to the i-th element in the result vector.
         auto subscriptOp = rewriter.create<emitc::SubscriptOp>(
-            op.getLoc(), lvalueType, sortOp.getResult(0), indexVal);
+            op.getLoc(), lvalueType, callOp.getResult(0), indexVal);
 
         // Load the actual tensor value from the reference.
         auto loadOp = rewriter.create<emitc::LoadOp>(
@@ -2180,7 +2181,7 @@ public:
       }
 
       rewriter.replaceOp(op, results);
-      return sortOp.getResult(0);
+      return callOp.getResult(0);
     }
 
     auto resultTypes = llvm::to_vector(
--- a/include/ttmlir/Dialect/TTIR/IR/TTIROps.td
+++ b/include/ttmlir/Dialect/TTIR/IR/TTIROps.td
@@ -5789,4 +5789,25 @@ def TTIR_SplitQueryKeyValueAndSplitHeadsOp: TTIR_NamedOp<"split_query_key_value_
 
     let hasVerifier = 1;
 }
+
+def TTIR_TopKOp : TTIR_NamedOp<"topk"> {
+  let summary = "Top-K selection operation.";
+  let description = [{
+    Returns the `k` largest or `k` smallest elements of the `input_tensor` along a given dimension `dim`.
+    If `dim` is not provided, the last dimension of the input_tensor is used.
+    If `largest` is True, the `k` largest elements are returned. Otherwise, the `k` smallest elements are returned.
+    The boolean option `sorted` if True, will make sure that the returned `k` elements are sorted.
+  }];
+
+  let arguments = (ins AnyRankedTensor:$input_tensor,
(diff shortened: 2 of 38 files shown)


azecevicTT (approved):

include/ttmlir/Dialect/D2M/Utils/Utils.h
```diff
@@ -15,11 +15,14 @@ class DeviceAttr;
...
-// Discardable attribute name for propagating virtualGridMapping through ops
-// we don't own (e.g. memref.alloc).  Uses the dialect prefix so MLIR can
-// verify it belongs to D2M.
-constexpr llvm::StringLiteral kVirtualGridMappingAttr =
-    "d2m.virtualGridMapping";
+// Discardable attribute names for propagating virtualGridMapping (inverse) and
+// virtualGridForwardMapping (forward) through ops we don't own (e.g.
+// memref.alloc).  Uses the dialect prefix so MLIR can verify they belong to
+// D2M.
+constexpr llvm::StringLiteral kVirtualGridInverseMappingAttr =
```
bgrady-tt, line 22: Very happy we are here. I think having both mappings in the same place makes verification and reblocking much simpler to get right.

lib/Dialect/D2M/Transforms/Allocate.cpp
```diff
@@ -1502,6 +1501,41 @@ class D2MAllocate final : public impl::D2MAllocateBase<D2MAllocate> {
...
+    //
+    // Skip this check for DRAM operands: DRAM is already treated as "remote"
+    // by simplifyLoadStorePairs, so the copy's data movement is preserved
+    // without needing a stream.
+    unsigned numInputs = genericOp.getNumDpsInputs();
+    if (operandIndex < numInputs) {
+      Value inputOperand = genericOp->getOperand(operandIndex);
+
+      // Check if input is in DRAM — if so, no stream needed for VGM reasons.
+      bool inputIsDRAM = false;
```
bgrady-tt: I believe the above check at LINE 1484 makes checking DRAM memspace here unnecessary?

Also I think the comments here are wrong; _DRAM doesn't weaken the requirement for streams if VGMs mismatch between input and output_. 

For _DMA only operations_ (with load directly feeding store), neither L1 or DRAM memspace operands require a stream buffer (view is sufficient). In that case we can directly perform DMA to/from remote directly on the local aliased CB; the remote operand doesn't need a stream_layout to provide a buffer. 

This usually manifests itself as a DRAM operand having a trivial view on a DMA-only op (to signal it is remote), but that isn't a property of the DRAM memspace itself.

lib/Dialect/D2M/Utils/Utils.cpp
```diff
@@ -273,40 +264,96 @@ std::optional<AffineMap> getVirtualGridMapping(Value val) {
...
+      for (auto [idx, result] : llvm::enumerate(genericOp.getResults())) {
+        if (result == val) {
+          Value outputOperand = genericOp.getOutputs()[idx];
+          return getVirtualGridForwardMapping(outputOperand);
+        }
+      }
+      return std::nullopt;
+    }
+
+    if (auto viewOp = mlir::dyn_cast<ViewLayoutOp>(defOp)) {
```
bgrady-tt: nit: view and stream can be handled together using ViewOpInterface

arminaleTT (approved):

lib/Dialect/D2M/IR/D2MOps.cpp
```diff
@@ -1717,31 +1710,14 @@ ::mlir::LogicalResult d2m::GenericOp::verify() {
...
-        TT_assert((shape.size() % 2) == 0ul);
-        outputGridShape.assign(shape.begin(), shape.begin() + shape.size() / 2);
-      } else {
-        outputGridShape = llvm::to_vector(ttcore::getGridShape(output));
-      }
-
-      if (outputGridShape != impliedVirtShape) {
-        return emitOpError("output grid shape does not match implied virtual "
-                           "grid shape from physical grid and inverse mapping");
-      }
```
bgrady-tt, line 1744: If discarding the above check (not sure why we are doing this), we need to replace it with something that still validates fwd/inv consistency. Note that fwd/inv VGM consistency is often only true only for specific shapes! Simply composing fwd+inv and checking it is an identity map will **_not_** work.

- For a given GenericOp, the GridAttr inverse map results should _always_ equal the output tensors inverse map results (ignoring extra device_id result in GridAttr map).
- For a given GenericOp, the output tensors grid shape applied to the output tensor's fwd VGM composed with inv VGM should be _identical_:
  - inv(fwd(output_grid_shape)) == output_grid_shape

bgrady-tt (changes requested):
Overall I think this is basically the right approach, however the verification of fwd and inv maps for grid attribute has been replaced with an inadequate check.

include/ttmlir/Dialect/D2M/Utils/Utils.h
vwellsTT: Yeah, this isn't strictly related to the bugfix, but I got so frustrated trying to debug inconsistencies last night I just went for this, imo it's a lot easier to use

lib/Dialect/D2M/Transforms/Allocate.cpp
vwellsTT: Yeah, I thought I'd cleaned this up; let me see if it's actually useful

vwellsTT: Ah, you reviewed right before my latest push, yeah removed this block completely, wasn't needed AFAICT

lib/Dialect/D2M/IR/D2MOps.cpp
```diff
@@ -1719,6 +1719,46 @@ ::mlir::LogicalResult d2m::GenericOp::verify() {
...
+          d2m::utils::getPhysicalGridShape(output);
+
+      // Drop the deviceID result (first result) from the inverse map.
+      AffineMap invMapNoDevice = gridInvMap.dropResult(0);
+
+      SmallVector<int64_t> impliedVirtShape =
+          ttmlir::utils::evalShape(invMapNoDevice, physicalGridShape);
+
+      SmallVector<int64_t> outputGridShape;
+      if (auto memrefType = mlir::dyn_cast<MemRefType>(output.getType())) {
```
bgrady-tt: small nit: I think this branch is unnecessary? Using `ttcore::getGridShape` should be fine here for any value of output (tensor or memref).

vwellsTT: good point 👍

Issue #7224, opened 2026-02-26
[mlir-snippets][kimik2_1t] snippet_silu_gate_matmul_64x4608_to_64x3584_bf16 PCC regression


anuragsingh-tt:
## Error signature

```
TTBuilderGoldenException: program-level output golden comparison failed, actual_pcc=0.1310873972258988 < expected_pcc=0.99
```

- Failure kind: `pcc_mismatch`
- Test case: `test/python/golden/test_snippets.py::test_model_snippet_compile_execute[kimik2_1t/snippet_silu_gate_matmul_64x4608_to_64x3584_bf16/silu_gate_matmul_64x4608_to_64x3584_bf16]`
- Snippet path: `test/python/golden/mlir_snippets/models/kimik2_1t/snippet_silu_gate_matmul_64x4608_to_64x3584_bf16.mlir`

## MLIR snippet

```mlir
module {
  func.func @silu_gate_matmul_64x4608_to_64x3584_bf16(%arg0: tensor<64x4608xbf16>, %arg1: tensor<64x4608xbf16>, %arg2: tensor<4608x3584xbf16>) -> tensor<64x3584xbf16> {
    %0 = "ttir.typecast"(%arg0) <{conservative_folding = false}> : (tensor<64x4608xbf16>) -> tensor<64x4608xf32>
    %1 = "ttir.sigmoid"(%0) : (tensor<64x4608xf32>) -> tensor<64x4608xf32>
    %2 = "ttir.multiply"(%0, %1) : (tensor<64x4608xf32>, tensor<64x4608xf32>) -> tensor<64x4608xf32>
    %3 = "ttir.typecast"(%2) <{conservative_folding = false}> : (tensor<64x4608xf32>) -> tensor<64x4608xbf16>
    %4 = "ttir.multiply"(%3, %arg1) : (tensor<64x4608xbf16>, tensor<64x4608xbf16>) -> tensor<64x4608xbf16>
    %5 = "ttir.matmul"(%4, %arg2) <{transpose_a = false, transpose_b = false}> : (tensor<64x4608xbf16>, tensor<4608x3584xbf16>) -> tensor<64x3584xbf16>
    return %5 : tensor<64x3584xbf16>
  }
}
```

anuragsingh-tt:
Resolved by #7183

PR #7274, opened 2026-02-27
d2m: adding a trait to selectively enable dst register to fp32

anuragsingh-tt:
### Ticket
#7224

### Problem description
Setting fp32_dest_acc_en=true for all ops is too broad, it should only apply to ops that partially accumulate in Dst. 

### What's changed
- Introduces a new TTKernel op trait (TTKernelDstAccumulatingOpTrait) and tags DST-accumulating ops (matmul/reduce variants).
- Updates D2M→TTMetal compute kernel config generation to enable fp32_dest_acc_en only when the kernel both (1) contains a DST-accumulating op and (2) uses a 32-bit CB port operand.
- Adds/updates MLIR tests and golden snippets to cover the regression scenario and validate the new compute config behavior.
 

### Checklist


lib/Conversion/D2MToTTMetal/D2MToTTMetal.cpp
```diff
@@ -45,6 +46,53 @@ class D2MGenericRewriter : public OpConversionPattern<d2m::GenericOp> {
...
+    bool hasDstAccumulatingOp = false;
+    kernelFunc.walk([&](Operation *op) {
+      if (!op->hasTrait<ttkernel::TTKernelDstAccumulatingOpTrait>()) {
+        return WalkResult::advance();
+      }
+      hasDstAccumulatingOp = true;
+      return WalkResult::interrupt();
+    });
+    if (!hasDstAccumulatingOp) {
+      return false;
```
wenbinlyuTT, line 69: Does this mean we will disable dst f32 accum for every op that isn't matmul/reduce? I'm not sure about how exactly this accum mode works, does it no affect the accuracy if the op only write once to the dst?

PR #7279, opened 2026-02-27
DST Tiling Analysis For Fusion Scratchpad

bgrady-tt:
This optimizes packing/chunking of shards into DST for all linalg.generic results in a d2m.generic blocking loop nest. It determines max DST capacity per op by accounting for execution class (SFPU vs FPU) and dtype (f32 halves DST tile capacity).

The API now returns a single aggregate object with per-result and global fields:

```cpp
struct DSTPackingInfo {
  llvm::SmallDenseMap<Value, DSTPackingPerResultInfo> perResult;
  int64_t numTilesPerResult = 0;
  int64_t numOuterLoopIters = 0;
};
```

where `DSTPackingPerResultInfo` is:
```cpp
struct DSTPackingPerResultInfo {
  int64_t numDstFlips = 0;
  int64_t numTilesPerFlip = 0;
};
```

- perResult stores {numDstFlips, numTilesPerFlip} for each linalg.generic output value.
- numOuterLoopIters is global and identical across all analyzed results.
- numTilesPerResult is global and identical across all analyzed results, with:
  - numTilesPerResult = numDstFlips * numTilesPerFlip

Optimization goals and constraints:
- Maximize common numOuterLoopIters across all linalg.generic results.
- Enforce identical numOuterLoopIters for all results.
- Enforce identical numTilesPerResult for all results.
- Minimum numDstFlips is 2.

### DIFF
--- a/lib/Conversion/D2MToTTNN/D2MToTTNN.cpp
+++ b/lib/Conversion/D2MToTTNN/D2MToTTNN.cpp
@@ -439,12 +439,6 @@ public:
       // The genericOp has a virtual grid. We need to recover the original
       // physical grid.
       auto output = op.getOutputs()[0];
-      mlir::ShapedType outputType =
-          mlir::cast<mlir::ShapedType>(output.getType());
-      auto shardLayout = mlir::dyn_cast<ttcore::ShardLayoutAttr>(
-          ttcore::getDeviceLayout(outputType));
-      TT_assertv(shardLayout, "Expected shardLayoutAttr for the output of a "
-                              "generic op with a virtual grid.");
 
       auto physicalGridShape = d2m::utils::getPhysicalGridShape(output);
       // TTNN grids are (Width, Height), while D2M grids are (Height, Width).
--- a/lib/Dialect/D2M/IR/D2MOps.cpp
+++ b/lib/Dialect/D2M/IR/D2MOps.cpp
@@ -1754,20 +1754,37 @@ MutableArrayRef<OpOperand> d2m::GenericOp::getInputsAndOutputsMutable() {
              << deviceVolume << ")";
     }
 
+    auto isDRAM = [](Value output) {
+      if (auto memrefType = mlir::dyn_cast<MemRefType>(output.getType())) {
+        return ttcore::getMemorySpace(memrefType) ==
+               ttcore::MemorySpace::DeviceDRAM;
+      }
+      if (auto tensorType =
+              mlir::dyn_cast<RankedTensorType>(output.getType())) {
+        if (auto layout = mlir::dyn_cast_if_present<ttcore::MetalLayoutAttr>(
+                tensorType.getEncoding())) {
+          return layout.getMemorySpace() == ttcore::MemorySpace::DeviceDRAM;
+        }
+      }
+      return false;
+    };
     // Verify per-output VGM consistency:
-    // 1. The output's inverse VGM must match the GridAttr's inverse map.
+    // 1. For non-DRAM outputs, the output's inverse VGM must match the
+    // GridAttr's inverse map.
     // 2. The inverse map applied to the physical grid shape must produce
     //    a virtual grid shape matching the output's grid shape.
     AffineMap gridInvMap = getGrid().getMapping();
     for (Value output : getOutputs()) {
-      auto outputInvMap = utils::getVirtualGridInverseMapping(output);
-      if (outputInvMap && *outputInvMap != gridInvMap) {
-        return emitOpError("grid inverse map does not match output operand's "
-                           "inverse VGM");
-      }
-      if (!outputInvMap && !gridInvMap.isEmpty()) {
-        return emitOpError("grid has an inverse map but output operand "
-                           "does not have a VGM");
+      if (!isDRAM(output)) {
+        auto outputInvMap = utils::getVirtualGridInverseMapping(output);
+        if (outputInvMap && *outputInvMap != gridInvMap) {
+          return emitOpError("grid inverse map does not match output operand's "
+                             "inverse VGM");
+        }
+        if (!outputInvMap && !gridInvMap.isEmpty()) {
+          return emitOpError("grid has an inverse map but output operand "
+                             "does not have a VGM");
+        }
       }
 
       SmallVector<int64_t> physicalGridShape =
--- a/lib/Dialect/D2M/Transforms/LowerLoadStoreOpsToExplicitCBForm.cpp
+++ b/lib/Dialect/D2M/Transforms/LowerLoadStoreOpsToExplicitCBForm.cpp
@@ -399,12 +399,11 @@ static PushPopInfo convertToExplicitCBForm(ModuleOp moduleOp,
     rewriter.eraseOp(allocOp);
   }
 
-  // Transform RemoteStoreOp (implicit form -> explicit CB form)
+  // Transform RemoteStoreOp: convert implicit form to explicit CB form, and
+  // ensure explicit form stores (e.g., created by simplifyLoadStorePairs) get
+  // reserve/push inserted for CB synchronization.
   SmallVector<RemoteStoreOp> remoteStoresToConvert;
   moduleOp->walk([&](RemoteStoreOp remoteStore) {
-    if (!remoteStore.isImplicitForm()) {
-      return;
-    }
     remoteStoresToConvert.push_back(remoteStore);
   });
 
--- a/lib/Dialect/D2M/Transforms/LowerToLayout.cpp
+++ b/lib/Dialect/D2M/Transforms/LowerToLayout.cpp
@@ -585,8 +585,13 @@ public:
     }
 
     Value viewOutput = output;
+    ttcore::GridAttr grid;
     if (outputInfo.isDRAM()) {
       viewOutput = buildConcreteView(output, outputInfo.type, inputInfo.type);
+      if (auto invMap = utils::getVirtualGridInverseMapping(input)) {
+        auto gridShape = llvm::to_vector(inputInfo.getGridShape());
+        grid = rewriter.getAttr<ttcore::GridAttr>(gridShape, *invMap);
+      }
     }
 
     const size_t gridRank = outputInfo.getGridShape().size();
@@ -615,7 +620,7 @@ public:
                       builder, innerLoc, viewOutput, indices, loadedData);
                   builder.create<YieldOp>(innerLoc, storeResult);
                 },
-                ThreadType::Unified)
+                ThreadType::Unified, grid)
             .getResult(0);
     return result;
   }
@@ -853,7 +858,7 @@ public:
       return emptyOp.getResult();
     };
 
-    // 1. SYSTEM→DEVICE: Transfer to L1/DRAM with same element type as input.
+    // SYSTEM→DEVICE: Transfer to L1/DRAM with same element type as input.
     if (!currentInfo.hasLayout() && targetInfo.hasLayout()) {
       // System transfer can ONLY change memory space, not element type.
       // Create intermediate with scalar element type (same as system input).
@@ -874,7 +879,7 @@ public:
       currentInfo = TensorInfo::from(currentValue);
     }
 
-    // 2. DRAM→L1: Must happen before other device ops.
+    // DRAM→L1: Must happen before other device ops.
     // Use target's layout characteristics.
     if (currentInfo.hasLayout() && currentInfo.isDRAM() &&
         targetInfo.hasLayout() && !targetInfo.isDRAM()) {
@@ -896,7 +901,7 @@ public:
       currentInfo = TensorInfo::from(currentValue);
     }
 
-    // 3. TILIZE: Before mapping (so mapping operates on final format).
+    // TILIZE: Before mapping (so mapping operates on final format).
     bool needsTilize =
         !ttcore::isTiled(currentInfo.type) && ttcore::isTiled(targetInfo.type);
     if (needsTilize && currentInfo.hasLayout()) {
@@ -913,7 +918,34 @@ public:
       currentInfo = TensorInfo::from(currentValue);
     }
 
-    // 4. MASKING: Apply boundary masking after tilization if needed.
+    // UNTILIZE: Before L1→DRAM or Device→System.
+    bool needsUntilize =
+        ttcore::isTiled(currentInfo.type) && !ttcore::isTiled(targetInfo.type);
+    if (needsUntilize) {
+      Type scalarType = targetInfo.type.getElementType();
+      // Avoid reblocking virtual grid shapes here. Output type here retains
+      // input's virtual grid shape; only transformation is to scalar dtype.
+      auto existingRemapping =
+          utils::getAssociatedRemapping(currentValue).value_or(AffineMap());
+      auto scalarType_ranked = typeBuilder.modifyDeviceType(
+          currentInfo.type, *currentInfo.layout, targetGridShape,
+          existingRemapping, /*memSpace=*/{}, /*newTensorGrid=*/{}, scalarType,
+          /*newTileShape=*/std::nullopt, /*reblockVirtualGridShapes=*/false);
+      auto scalarEmpty = createEmpty(scalarType_ranked);
+      currentValue = lowerFormatConversionGeneric(rewriter, currentValue,
+                                                  scalarEmpty, op.getLoc());
+      currentInfo = TensorInfo::from(currentValue);
+    }
+
+    // L1→DRAM (lowerDatamovementGeneric handles grid mismatch via views).
+    if (currentInfo.hasLayout() && !currentInfo.isDRAM() &&
+        targetInfo.hasLayout() && targetInfo.isDRAM()) {
+      currentValue = lowerDatamovementGeneric(rewriter, currentValue,
+                                              op.getOutput(), op.getLoc());
+      currentInfo = TensorInfo::from(currentValue);
+    }
+
+    // MASKING: Apply boundary masking after tilization if needed.
     // Insert TileMaskBoundaryOp when the target layout has non-Undef OOBVal
     // and padding exists.
     if (currentInfo.hasLayout() && ttcore::isTiled(currentInfo.type) &&
@@ -935,7 +967,7 @@ public:
       currentInfo = TensorInfo::from(currentValue);
     }
 
-    // 5. MAPPING CHANGE: Grid/index_map/logical_shape/dim_alignments (after
+    // MAPPING CHANGE: Grid/index_map/logical_shape/dim_alignments (after
     // tilize). Includes all reblocking (both virtual and normal grids). Must
     // happen in L1 (can't reblock in DRAM). Only when element type formats
     // match (tilize/untilize should happen first).
@@ -983,7 +1015,7 @@ public:
           // shapes don't divide evenly into tiles. Decompose via scalar space:
           // untilize → map in scalar space → tilize back.
 
-          // 5a. Untilize to scalar space (preserve current layout properties).
+          // Untilize to scalar space (preserve current layout properties).
           // Reblock virtual grid shape here to align with earlier splitting
           // phases that use reblocked intermediates to bounce virtual grid
           // shapes from host to device.
@@ -999,7 +1031,7 @@ public:
               rewriter, currentValue, untilizedEmpty, op.getLoc());
           currentInfo = TensorInfo::from(currentValue);
 
-          // 5b. Apply complex mapping change in scalar space.
+          // Apply complex mapping change in scalar space.
           // Build scalar target with ALL target's layout properties.
           auto scalarTargetLayout = ttcore::MetalLayoutAttr::get(
               rewriter.getContext(), targetInfo.layout->getLogicalShape(),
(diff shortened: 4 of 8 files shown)


lib/Dialect/D2M/Utils/Utils.cpp
```diff
@@ -14,15 +14,308 @@
...
+}
+
+static std::optional<int64_t> getShardSizeInTiles(Value outputValue) {
+  auto shapedType = mlir::dyn_cast<ShapedType>(outputValue.getType());
+  if (!shapedType || !shapedType.hasStaticShape()) {
+    return std::nullopt;
+  }
+
+  int64_t shardSizeTiles = 1;
+  for (int64_t dim : shapedType.getShape()) {
```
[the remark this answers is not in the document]
bgrady-tt: stale

include/ttmlir/Dialect/D2M/Utils/DstRegisterAnalysis.h
```diff
@@ -0,0 +1,34 @@
...
+
+struct DSTPackingInfo {
+  llvm::SmallDenseMap<Value, DSTPackingPerResultInfo> perResult;
+  int64_t numTilesPerResult = 0;
+  int64_t numOuterLoopIters = 0;
+};
+
+// Analyze linalg.generic ops in a unified d2m.generic region and compute legal
+// DST packing values that maximize common outer loop iterations.
+DSTPackingInfo analyzeGenericForDSTPacking(d2m::GenericOp generic);
```
nsmithtt: We should nest this inside of a struct see `include/ttmlir/Dialect/D2M/Analysis/CBProducerConsumer.h` as an example which enables you to opt into MLIR's `getAnalysis<DstRegisterAnalysis>()` which does caching etc.

include/ttmlir/Dialect/D2M/Utils/DstRegisterAnalysis.h
```diff
@@ -0,0 +1,34 @@
+// SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
```
nsmithtt: nit 2026

lib/Dialect/D2M/Utils/DstRegisterAnalysis.cpp
```diff
@@ -0,0 +1,288 @@
+// SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
```
nsmithtt: nit 2026

lib/Dialect/D2M/Utils/DstRegisterAnalysis.cpp
```diff
@@ -0,0 +1,288 @@
...
+  if (!generic.isUnifiedForm()) {
+    generic.emitOpError("expected unified form for DST packing analysis");
+    return DSTPackingInfo();
+  }
+  Region &unifiedRegion = generic.getRegion(0);
+
+  SmallVector<linalg::GenericOp> linalgOps;
+  unifiedRegion.walk([&](linalg::GenericOp op) { linalgOps.push_back(op); });
+
+  scf::ForOp commonImmediateParentBlockingLoop = nullptr;
```
nsmithtt: Instead of this, why don't we use a `DenseMap<Region*, linalg::GenericOp>` above so that linalg's are grouped by parent region, then we run an analysis per region.

lib/Dialect/D2M/Utils/DstRegisterAnalysis.cpp
```diff
@@ -0,0 +1,288 @@
...
+      mlir::dyn_cast<ttcore::TileType>(outputShapedType.getElementType());
+  if (!tileType) {
+    return std::nullopt;
+  }
+
+  const ttcore::DataType dataType = tileType.getDataType();
+  const bool isFp32 = dataType == ttcore::DataType::Float32;
+
+  int64_t maxDstTiles =
+      classifyLinalgExecutionClass(op) == DstExecutionClass::FPU ? 8 : 4;
```
nsmithtt: We should use/augment `ChipDesc::getDstLogicalSizeTiles` instead of magic numbers here.

lib/Dialect/D2M/Utils/DstRegisterAnalysis.cpp
```diff
@@ -0,0 +1,288 @@
...
+
+static std::optional<int64_t> getSmallestLegalChunkSize(int64_t shardSizeTiles,
+                                                        int64_t maxDstTiles) {
+  int64_t largestCandidate = std::min(maxDstTiles, shardSizeTiles / 2);
+  for (int64_t numTilesPerFlip = 1; numTilesPerFlip <= largestCandidate;
+       ++numTilesPerFlip) {
+    if ((2 * numTilesPerFlip) > shardSizeTiles) {
+      continue;
+    }
+    if ((shardSizeTiles % numTilesPerFlip) != 0) {
```
nsmithtt, line 104: Anything % 1 is 0, doesn't that mean that we're almost always returning 1 from this function?  Not sure if this is intended, perhaps we should be counting down?

lib/Dialect/D2M/Utils/DstRegisterAnalysis.cpp
```diff
@@ -0,0 +1,288 @@
...
+    std::optional<int64_t> maxDstTiles = getMaxDstTilesForLinalgOp(linalgOp);
+    if (!maxDstTiles) {
+      linalgOp.emitOpError("failed to compute max DST tile capacity");
+      return DSTPackingInfo();
+    }
+
+    std::optional<int64_t> numTilesPerFlip =
+        getSmallestLegalChunkSize(shardSizeTiles, *maxDstTiles);
+    if (!numTilesPerFlip) {
+      linalgOp.emitOpError("failed to find legal tiles per DST flip");
```
nsmithtt: Is this a legitimate failure mode? It seems like we should always be able to find a legal size.

lib/Dialect/D2M/Utils/DstRegisterAnalysis.cpp
```diff
@@ -0,0 +1,288 @@
...
+      return DSTPackingInfo();
+    }
+
+    int64_t numDstFlips = shardSizeTiles / *numTilesPerFlip;
+    if (numDstFlips <= 0) {
```
nsmithtt: This feels more like it should be an assertion, how can a negative result appear here?

lib/Dialect/D2M/Utils/DstRegisterAnalysis.cpp
```diff
@@ -0,0 +1,288 @@
...
+      linalgOp.emitOpError("expected positive DST flip count");
+      return DSTPackingInfo();
+    }
+    if (numDstFlips < 2) {
+      linalgOp.emitOpError("failed to satisfy num_dst_flips >= 2");
```
nsmithtt: Is this a legitimate failure mode? In some cases, i.e. shard size is 1 there might be nothing to flip, but I don't think we should fail.

lib/Dialect/D2M/Utils/DstRegisterAnalysis.cpp
```diff
@@ -0,0 +1,288 @@
...
+    return std::nullopt;
+  }
+
+  int64_t maxCandidate = numDstFlipsPerOp.front() / 2;
+  int64_t gcdNumDstFlips = numDstFlipsPerOp.front();
+  for (int64_t numDstFlips : numDstFlipsPerOp.drop_front()) {
+    // Enforce num_dst_flips >= 2, i.e.
+    // num_outer_loop_iters <= num_dst_flips / 2.
+    maxCandidate = std::min(maxCandidate, numDstFlips / 2);
+    gcdNumDstFlips = std::gcd(gcdNumDstFlips, numDstFlips);
```
nsmithtt, line 124: Seems like there's a bug here, if `numDstFlipsPerOp = [12, 8]` the way I'm reading the code `maxCandidate` will result in `4`, but it should be `2`.  GCD probably needs to be computed first?

lib/Dialect/D2M/Utils/DstRegisterAnalysis.cpp
```diff
@@ -0,0 +1,288 @@
...
+  }
+  TT_assertv(maxCandidate >= 1, "maxCandidate must be >= 1");
+
+  // Pick the largest legal factor so the common outer loop iteration count is
+  // maximized.
+  for (int64_t factor :
+       llvm::reverse(ttmlir::utils::getFactors(gcdNumDstFlips))) {
```
nsmithtt, line 131: Perhaps I'm missing the point here, but didn't we just calculate this above via `gcdNumDstFlips`?  It seems like we could compute this in 1 pass.

lib/Dialect/D2M/Utils/DstRegisterAnalysis.cpp
```diff
@@ -0,0 +1,288 @@
...
+    if (factor <= maxCandidate) {
+      return factor;
+    }
+  }
```
nsmithtt, line 135: Perhaps we should assert false down here?  This doesn't seem like an expected failure mode.

lib/Dialect/D2M/Utils/DstRegisterAnalysis.cpp
```diff
@@ -0,0 +1,288 @@
...
+    return DSTPackingInfo();
+  }
+
+  std::optional<int64_t> commonNumTilesPerResult;
+  for (const PendingDSTPackingResult &pending : pendingResults) {
+    int64_t numDstFlips = pending.numDstFlips / *commonNumOuterLoopIters;
+    if ((pending.numDstFlips % *commonNumOuterLoopIters) != 0 ||
+        numDstFlips < 2) {
+      generic.emitOpError("failed to satisfy common num_outer_loop_iters and "
+                          "num_dst_flips >= 2");
```
nsmithtt: Maybe I'm missing how it's supported, but I think we definitely need to support the shard volume == 1 case.

test/unittests/D2MGenericAnalysis/TestD2MGenericAnalysis.cpp
```diff
@@ -95,4 +157,453 @@ TEST_F(GenericOpAnalysisTest, CanAnalyzeDimConstraintsForIncompatibleShapes) {
...
+            3); // tile_sub(tile,tile) => FPU fp32
+  EXPECT_EQ(packing.perResult.lookup(outputValues[0]).numDstFlips, 15);
+  EXPECT_EQ(packing.perResult.lookup(outputValues[1]).numDstFlips, 5);
+  EXPECT_EQ(packing.perResult.lookup(outputValues[2]).numDstFlips, 15);
+  EXPECT_EQ(packing.perResult.lookup(outputValues[3]).numDstFlips, 5);
+  EXPECT_EQ(packing.numTilesPerResult, 15);
+  EXPECT_EQ(packing.numOuterLoopIters, 1);
+}
+
+TEST_F(GenericOpAnalysisTest, CanAnalyzeGenericForDSTPackingFPUBf16_8x8) {
```
vmilosevic: :warning: **google\-readability\-avoid\-underscore\-in\-googletest\-name** :warning:
avoid using "\_" in test name "CanAnalyzeGenericForDSTPackingFPUBf16\_8x8" according to Googletest FAQ

lib/Dialect/D2M/Utils/DstRegisterAnalysis.cpp
bgrady-tt: Discussed this with @jdesousa-TT and @xanderchin. It isn't possible to use this sort of double buffering scheme for single-tile case; we will need a bespoke fallback there (perhaps executing dummy loop iterations).

lib/Dialect/D2M/Utils/DstRegisterAnalysis.cpp
```diff
@@ -0,0 +1,285 @@
...
+
+  SmallVector<linalg::GenericOp> linalgOps;
+  unifiedRegion.walk([&](linalg::GenericOp op) { linalgOps.push_back(op); });
+
+  scf::ForOp commonImmediateParentBlockingLoop = nullptr;
+  for (linalg::GenericOp linalgOp : linalgOps) {
+    scf::ForOp immediateParentBlockingLoop =
+        getImmediateParentBlockingLoop(linalgOp);
+    if (immediateParentBlockingLoop == nullptr) {
+      linalgOp.emitOpError(
```
jdesousa-TT: do we need this assertion and the one below? I don't see blocking loops getting generated for the cases i'm testing. What if block factors are 1?

lib/Dialect/D2M/Utils/DstRegisterAnalysis.cpp
bgrady-tt: agreed

lib/Dialect/D2M/Utils/DstRegisterAnalysis.cpp
bgrady-tt: Basically, I don't think we know the precise mechanism we are going to use to handle single-tile. So for now, the analysis here just handles shard_size > 1 tile.

lib/Dialect/D2M/Utils/DstRegisterAnalysis.cpp
bgrady-tt: The GCD would not be valid here, as it could easily violate the constraint numDstFlips >= 2 for all ops:

dstFlipsPerOp = [6,12,18]
-> gcd = 6, min = 3
-> must make outer loop iters 3, so that each op has at least two dst flips

bgrady-tt: the second loop is basically iterating over all the valid settings for valid outer loop count and finding the biggest one that still keeps the worst-case linalg.generic with >= 2 DST flips.

lib/Dialect/D2M/Utils/DstRegisterAnalysis.cpp
bgrady-tt: Code has changed a little since yesterday, but as of now `maxCandidate` is the max possible _outer loop nest iteration count_. This is basically `min(numDstFlipsPerOp)/2`, which would be `8/2 = 4`.

lib/Dialect/D2M/Utils/DstRegisterAnalysis.cpp
bgrady-tt: This should probably just return a null analysis, you are right. 

I still don't know if we should pad single tile case to 2 tiles (and expect caller to understand this). This PR will probably evolve a little bit with fusion branch until we understand exactly what approach will be.

lib/Dialect/D2M/Utils/DstRegisterAnalysis.cpp
bgrady-tt: Ah did not know about this.

lib/Dialect/D2M/Utils/DstRegisterAnalysis.cpp
bgrady-tt: Changed the logic here to find largest chunk size.

lib/Dialect/D2M/Utils/DstRegisterAnalysis.cpp
bgrady-tt: As noted elsewhere, I think we need to settle on meaningful result to return when shard size is 1 tile.

lib/Dialect/D2M/Utils/DstRegisterAnalysis.cpp
bgrady-tt: I think we could do an analysis per region 👍

lib/Dialect/D2M/Utils/DstRegisterAnalysis.cpp
bgrady-tt: yeah, I think I need to clean up error handling in general.

lib/Dialect/D2M/Utils/DstRegisterAnalysis.cpp
nsmithtt: Maybe a simple workaround for the single tile case is just to force full dst mode?

lib/Dialect/D2M/Utils/DstRegisterAnalysis.cpp
nsmithtt: I guess, why can't the code be?
```c++
auto gcd = std::gcd(*numDstFlipsPerOp);
return ttmlir::utils::getFactors(gcd)[-2];
```

nsmithtt: Maybe this is too much of a nit, I found the code kind of hard to reason about, but I see how it works now so we can leave it as is.

lib/Dialect/D2M/Utils/DstRegisterAnalysis.cpp
jdesousa-TT: don't think there needs to be dependence here on the data movement level blocking (unless i'm missing something). just commenting this out worked for me

bgrady-tt: I have refactored to make the analysis results per-region (i.e. analyzed groups must all exist in the same region scope). The linalg.generic can now also exist at top-level scope in d2m.generic unified region.

lib/Dialect/D2M/Utils/DstRegisterAnalysis.cpp
bgrady-tt: For the analysis here, I am going to assume that `numDstFlips` = 2 and `numTilesPerFlip` = 1, and introduce a new member field `tilePadding` to represent dummy tiles inserted to pad single tile shard cases to >= 2 DST flips.

lib/Dialect/D2M/Utils/DstRegisterAnalysis.cpp
bgrady-tt: This is fixed in the latest version of this branch. Single tile shard case is handled by inserting a 'padding' field. Fusion branch doesn't yet handle this, but likely we will solve the case by executing dummy iterations and selectively packing out results. 

Additionally if analysis fails without breaking invariants (d2m.generic op is valid, but structure makes it unanalyzable), the returned analysis info for that generic op is simply null. Caller is expected to handle this.

lib/Dialect/D2M/Utils/DstRegisterAnalysis.cpp
bgrady-tt: converted this to an `TT_assertv`

lib/Dialect/D2M/Utils/DstRegisterAnalysis.cpp
bgrady-tt: this analysis no longer returns a std::optional, but asserts if shardSize is <= 0. For single tile shard case, we handle that as a special case and early return 1.

lib/Dialect/D2M/Utils/DstRegisterAnalysis.cpp
bgrady-tt: Using ChipDesc::getDstLogicalSizeTiles instead of hardcoded numbers here.

lib/Dialect/D2M/Utils/DstRegisterAnalysis.cpp
bgrady-tt: Changed to have a `llvm::DenseMap<Region *, DSTPackingRegionInfo> perRegion` mapping each Region* to an analysis result.

include/ttmlir/Dialect/D2M/Utils/DstRegisterAnalysis.h
bgrady-tt: Changed API to align with MLIR's analysis interface.

test/unittests/D2MGenericAnalysis/TestD2MGenericAnalysis.cpp
```diff
@@ -38,6 +60,61 @@ class GenericOpAnalysisTest : public ::testing::Test {
...
+      if (linalgOp.getOutputs().size() == 1u) {
+        outputValues.push_back(
+            {linalgOp->getParentRegion(), linalgOp.getOutputs().front()});
+      }
+    });
+    return outputValues;
+  }
+
+  static constexpr llvm::StringLiteral kModuleHeader = R"mlir(
+#system_desc = #ttcore.system_desc<[{role = host, target_triple = "x86_64-pc-linux-gnu"}], [{arch = <wormhole_b0>, grid = 8x8, coord_translation_offsets = 18x18, l1_size = 1499136, num_dram_channels = 12, dram_channel_size = 1073741824, noc_l1_address_align_bytes = 16, pcie_address_align_bytes = 32, noc_dram_address_align_bytes = 32, l1_unreserved_base = 1024, erisc_l1_unreserved_base = 1024, dram_unreserved_base = 1024, dram_unreserved_end = 1073741824, supported_data_types = [<f32>, <f16>, <bf16>, <bfp_f8>, <bfp_bf8>, <bfp_f4>, <bfp_bf4>, <bfp_f2>, <bfp_bf2>, <u32>, <u16>, <u8>, <si32>], supported_tile_sizes = [ 4x16,  16x16,  32x16,  4x32,  16x32,  32x32], dst_physical_size_tiles = 16, num_cbs = 32, num_compute_threads = 1, num_datamovement_threads = 2}], [0], [1 : i32], [ 0x0x0x0]>
```
nsmithtt: Is it possible to run the register device pass instead of explicit system desc here? It makes updating the system desc really difficult and tedious.

bgrady-tt: I can see if this is possible.

bgrady-tt: Added this change; should make this easier to maintain 👍

PR #7298, opened 2026-03-02
[D2M] CB Rework

vwellsTT:
### Ticket
https://github.com/tenstorrent/tt-mlir/issues/6946

### Problem description
CB handling is a bit awkward in IR, especially as we support more complex fusion, remove streams, etc.

### What's changed
Get rid of CB block args completely.  Treat internal allocations as valid allocations including CB--Allocator must give them space (but not allow spilling, etc for now).  Update various locations which access CBs as block args to access through allocs.  Update tests to match new form.

### Checklist
- [ ] New/Existing tests provide coverage for changes

PR #7304, opened 2026-03-02
[D2M] Add d2m.spatial op and lowering to ttnn.generic 

hkwonTT:
### What's changed

- **Op definition (TableGen)**  
  - `d2m.spatial`: `grid_ranges` attribute (non-overlapping device grid ranges), variadic inputs/outputs, one region per grid; each region has exactly one `d2m.generic`.
  - `d2m.spatial_yield`: terminator for spatial regions; returns region results that are aggregated as the spatial op's results in region order.

- **Pass integration**  
  - **GridSelection**: Handles `d2m.spatial` by assigning each region to its `grid_ranges` entry and processing the inner `d2m.generic` per region.  
  - **D2MToTTNN**: Lowers `d2m.spatial` to `ttnn.generic` (one per region with the correct grid/core range; results aggregated in region order).  
  - **TTNNToFlatbuffer**: For virtual core ranges with a mapping, computes the implied physical core range so spatial regions are emitted with the correct device cores.

- **TTCore: CoreRangeSetAttr and related types**  
  `d2m.spatial` needs a `grid_ranges` attribute (a set of non-overlapping device grid ranges). The types for this (e.g. `CoreRangeSetAttr`) currently exist in the TTNN dialect, but they are generic enough to be used by D2M and Metal as well, so they belong in TTCore. Moving the existing TTNN definitions into TTCore will be done in a separate PR. For this PR, the needed types are defined in TTCore (duplicating the TTNN definitions for now). A follow-up PR will remove the TTNN definitions and migrate usages to the TTCore types.
### Checklist


include/ttmlir/Dialect/D2M/Utils/VirtualGrid.h
```diff
@@ -41,6 +41,13 @@ createCoreVirtMaps(mlir::MLIRContext *context,
                    llvm::ArrayRef<int64_t> virtualGrid,
                    llvm::ArrayRef<int64_t> targetGrid);
 
+/// Creates a 2D affine map for grid physical -> (deviceId, virtual): takes
+/// physical grid coordinates (d0, d1) and produces (0, d0 - startY, d1 -
+/// startX). The leading 0 is the device id. Use for GridAttr mapping when the
+/// grid runs on a core range starting at (startY, startX).
+mlir::AffineMap createPhysicalToVirtualMap(mlir::MLIRContext *context,
```
bgrady-tt: would rename this to `createXYOffsetVirtualGridMap`; we use physical to virtual in other places in codebase to mean something else.

include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
```diff
@@ -1636,6 +1636,60 @@ def D2M_YieldOp : D2M_GenericRegionOp<"yield", [
...
+      "bufferizesToMemoryRead",
+      "bufferizesToMemoryWrite",
+      "bufferize",
+      "getAliasingValues",
+      "getBufferType"
+    ]>
+  ]> {
+    let summary = "Yield op for spatial op.";
+    let description = [{
+      Yield operation for spatial op. Return control flow to another thread. Corresponds to a circular buffer push.
```
bgrady-tt: Description here with circular buffer and thread seems like a holdover from regular d2m.yield op:

Change "Yield operation for spatial op. Return control flow to another thread. Corresponds to a circular buffer push." -->
"Yield operation for spatial op"

include/ttmlir/AffineMapUtils.h
```diff
@@ -35,6 +35,27 @@ llvm::SmallVector<int64_t> evalShape(mlir::AffineMap map, Vector shape) {
   }
   return result;
 }
+/// Returns a new shape by applying `map` to the input shape.
```
bgrady-tt: Add more descriptive comment here about why this differs from evalShape i.e. it doesn't assume the output shape starts at 0,0.

lib/Dialect/D2M/Utils/VirtualGrid.cpp
```diff
@@ -92,6 +92,18 @@ createCoreVirtMaps(mlir::MLIRContext *context,
   return {forwardMap, inverseMap};
 }
 
+mlir::AffineMap createPhysicalToVirtualMap(mlir::MLIRContext *context,
```
bgrady-tt: See other comment about renaming this.

lib/Dialect/D2M/IR/D2MGenericRegionOps.cpp
```diff
@@ -1745,6 +1745,17 @@ mlir::LogicalResult YieldOp::verify() {
 
   return ::mlir::success();
 }
+mlir::LogicalResult SpatialYieldOp::verify() {
+  auto spatial = getOperation()->getParentOfType<SpatialOp>();
+  if (!spatial
+      //|| !generic.hasPureTensorSemantics()
```
bgrady-tt: cleanup :)

lib/Dialect/D2M/IR/D2MOps.cpp
```diff
@@ -1997,6 +1998,28 @@ ::mlir::LogicalResult d2m::GenericOp::verify() {
 
   return success();
 }
+// Spatialop verification
+::mlir::LogicalResult d2m::SpatialOp::verify() {
```
bgrady-tt, line 2175: I think this should also verify that generic op contained in each region have physical grid shapes that fit in the region's CoreRange.

hkwonTT:
While debugging LIT test failures, I realized that setting the grid mapping attribute in the grid-selection pass is unnecessary. So I reverted the implementation and the related utility functions.

lib/Dialect/D2M/IR/D2MOps.cpp
nsmithtt: +1

lib/Dialect/D2M/Transforms/GridSelection.cpp
```diff
@@ -869,6 +870,16 @@ static void updateEmptyOps(ArrayRef<EmptyUpdateInfo> emptyOpsToUpdate,
   }
 }
 
+// Returns the enclosing d2m.spatial op if genericOp is inside one, otherwise
+// null.
+static d2m::SpatialOp getEnclosingSpatialOpIfPresent(d2m::GenericOp genericOp) {
+  mlir::Region *region = genericOp->getParentRegion();
+  if (!region) {
+    return d2m::SpatialOp();
```
nsmithtt: I think it'd be better to `assert(region)` only ops that have trait isolated from above are allowed to not belong to a parent region.

lib/Conversion/D2MToTTNN/D2MToTTNN.cpp
```diff
@@ -342,6 +361,151 @@ class D2MGenericRewriter : public OpConversionPattern<d2m::GenericOp> {
...
+    ArrayRef<d2m::GenericOp> generics,
+    ArrayRef<ttnn::CoreRangeSetAttr> regionCoreRangeSets,
+    const ttcore::DeviceAttr &device, SymbolTable &symbolTable,
+    ttmetal::MathFidelity mathFidelity) {
+  MLIRContext *ctx = builder.getContext();
+  llvm::SmallDenseMap<Value, size_t> valueToGlobalIOIndex;
+  SmallVector<Value> globalIos;
+
+  for (Value v : spatialOp.getInputs()) {
+    Value io = D2MGenericRewriter::extractIOAndCBFromGenericOperand(v).io;
```
vtangTT: can we extract these out as shared helpers instead of static members functions inside `D2MGenericRewriter`

same for `createCBDescriptors`, `createSemaphoreDescriptors`, `createKernelDescriptors`

include/ttmlir/AffineMapUtils.h
hkwonTT: Thanks for the comment. Unfortunately, this function was removed while updating grid-selection pass implementation.

lib/Dialect/D2M/IR/D2MOps.cpp
hkwonTT: Verification code has been updated. Thanks!

lib/Dialect/D2M/Transforms/GridSelection.cpp
hkwonTT: Agreed. The updated code asserts here. Thanks.

lib/Conversion/D2MToTTNN/D2MToTTNN.cpp
hkwonTT: I tried to preserve the code for GenericOp as much as possible. Placing functions that are commonly used across multiple classes outside the classes might be reasonable. I updated it as you suggested. Thanks!

hkwonTT:
Here are the major updates from the last review:
- Rebased on the latest commit.
- Updated the verify function for SpatialOp to ensure that the core grids in GenericOp fit within the grid ranges of the SpatialOp.
- Added LIT tests for op verification, grid selection, and d2m-to-ttnn. Also added silicon tests as end-to-end execution tests, but not golden checks. I will generate golden check tests with the builder if I have a chance.

hkwonTT:
The spatial op lowering implementation was broken by this change: #7298 .
Debugging it now.

include/ttmlir/Dialect/TTCore/IR/TTCoreOpsTypes.td
```diff
@@ -760,4 +760,71 @@ def TTCore_ArgumentTypeAttr : EnumAttr<TTCore_Dialect, TTCore_ArgumentType, "arg
...
+    Example:
+
+    ```mlir
+    // Define a core range spanning from (1,1) to (3,4)
+    // This represents a 3x4 rectangular grid of cores
+    #core_range = #ttcore.core_range<(1, 1), (3, 4)>
+    ```
+  }];
+
+  let parameters = (ins "CoreCoordAttr":$start_coord, "CoreCoordAttr":$end_coord);
```
nsmithtt, line 787: Won't block the PR, but I tend to prefer `shape` instead of `end_coord` because it's usually the information you're after when reading the IR.

include/ttmlir/Dialect/TTCore/IR/TTCoreOpsTypes.td
```diff
@@ -760,4 +760,71 @@ def TTCore_ArgumentTypeAttr : EnumAttr<TTCore_Dialect, TTCore_ArgumentType, "arg
...
+  let assemblyFormat = "`<` custom<CoordBracketStyle>($start_coord) `,` custom<CoordBracketStyle>($end_coord) `>`";
+
+  let extraClassDeclaration = [{
+    bool intersects(CoreRangeAttr other) const;
+  }];
+
+  let genVerifyDecl = 1;
+}
+
+def TTCore_CoreRangeSetAttr: TTCore_Attr<"CoreRangeSet", "core_range_set"> {
```
nsmithtt: I'd somewhat prefer we just use `OptionalArrayRefParameter<"CoreRangeAttr">:$core_ranges` directly in the spatial op definition because `CoreRangeSet` exists in tt-metal and implies something slightly different.

lib/Conversion/D2MToTTNN/D2MToTTNN.cpp
```diff
@@ -795,6 +797,281 @@ static LogicalResult convertGenerics(ModuleOp module,
...
+
+  rewriter.setInsertionPoint(spatialOp);
+  rewriter.replaceOpWithNewOp<ttnn::GenericOp>(
+      spatialOp, remapTable.getUnifiedIO(),
+      remapTable.getUnifiedAdditionalArgs(), mergedProgram,
+      ttnn::MemoryConfigAttr());
+  return success();
+}
+
+static LogicalResult convertSpatials(ModuleOp module) {
```
nsmithtt: `module` is a keyword in newer c++, we should use `moduleOp` for the var name.

include/ttmlir/Dialect/TTCore/IR/TTCoreOpsTypes.td
hkwonTT: I thought `CoreRangeSet` was used in the same way. I did not know there was a slight difference. Anyway, I have defined a new attribute, `CoreRangeArrayAttr`, and have used it for the `grid_ranges` attribute.

lib/Conversion/D2MToTTNN/D2MToTTNN.cpp
hkwonTT: Thanks. I updated variable names in the entire file.

include/ttmlir/Dialect/TTCore/IR/TTCoreOpsTypes.td
nsmithtt: Sounds good, yeah technically it could be used in the same way, but in tt-metal it's an attribute used for non-rectangular dispatches of a single program.  For example you want to dispatch to a logical 1x10 grid on wormhole, in tt-metal you'd encode that as `CoreRangeSet({[0,0], [0,7]}, {[1,0],[1,1]})`

include/ttmlir/Dialect/TTCore/IR/TTCoreOpsTypes.td
hkwonTT: So you are suggesting we represent a grid as {origin, shape}, instead of {start, end}, right?

In my current code, I already compute the shape from start/end, so having shape directly would simplify things.

While looking into this, I also noticed that we already have `TTCore_GridAttr`, which represents a grid as {shape, mapping}.

Given that, I think we have three options:
1. keep `CoreRangeAttr` as {start, end} (current PR),
2. change `CoreRangeAttr` to {origin, shape},
3. reuse `TTCore_GridAttr` instead.

My impression is that option 2 is the most straightforward for readability in the IR, since {origin, shape} is more direct than both {start, end} and {shape, mapping}. However, it would introduce a duplicate representation of a similar concept.

Would you prefer option 2, or do you think we should reuse `TTCore_GridAttr` instead?

nsmithtt: I prefer option 2 since I find it more readable, but also fine with keeping the PR the way it is.

bgrady-tt: My two cents: I think `TTCore_GridAttr` has a different use case and role than `CoreRangeAttr`. Representing the data as {origin, shape} keeps the grid shape translation invariant.

```mlir
// these all represent the same 3x4 grid starting at different origins
#core_range = #ttcore.core_range<(1, 1), (3, 4)>
#core_range = #ttcore.core_range<(0, 0), (3, 4)>
#core_range = #ttcore.core_range<(4, 2), (3, 4)>
```

bgrady-tt: But option1 (what exists as-is) is also fine. option3 seems to introduce coupling where there shouldn't be any.

lib/Dialect/D2M/IR/D2MOps.cpp
bgrady-tt: thanks 🙏

include/ttmlir/Dialect/TTCore/IR/TTCoreOpsTypes.td
hkwonTT: Okay. I'll update the CoreRangeAttr as (start, shape) soon. Thanks!

PR #7310, opened 2026-03-03
[D2M] Remove hacky fallback path in generic op builder

vwellsTT:
### Ticket
NA

### Problem description
In PR to remove index_maps, I missed some code which enables a hacky fallback in a generic builder to update output virtual grid in place when building a generic.  This is not correct--instead we should just ensure all creation site already have the correct output format.

### What's changed
Remove fallback fix-up path from generic builder, and add VGM-coverage to LowerToLayout in relevant places where we could create output w/ mismatched VGM.

### Checklist

### DIFF
--- a/include/ttmlir/OpModel/TTNN/TTNNOpModel.h
+++ b/include/ttmlir/OpModel/TTNN/TTNNOpModel.h
@@ -1521,13 +1521,13 @@ struct OpModel<ClampScalarOp> {
   static llvm::Expected<OpConstraints>
   getOpConstraints(ttcore::GridAttr deviceGrid,
                    llvm::ArrayRef<int64_t> inputShape,
-                   TTNNLayoutAttr inputLayout, llvm::APFloat min,
-                   llvm::APFloat max, TTNNLayoutAttr outputLayout);
+                   TTNNLayoutAttr inputLayout, mlir::Attribute min,
+                   mlir::Attribute max, TTNNLayoutAttr outputLayout);
 
   static llvm::Expected<size_t> getOpRuntime(llvm::ArrayRef<int64_t> inputShape,
                                              TTNNLayoutAttr inputLayout,
-                                             llvm::APFloat min,
-                                             llvm::APFloat max,
+                                             mlir::Attribute min,
+                                             mlir::Attribute max,
                                              TTNNLayoutAttr outputLayout);
 };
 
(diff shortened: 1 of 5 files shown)


lib/Dialect/D2M/Transforms/LowerToLayout.cpp
```diff
@@ -925,11 +951,13 @@ class D2MLowerToLayoutRewriter : public OpRewritePattern<ToLayoutOp> {
...
-      auto maskedEmpty =
-          rewriter
-              .create<d2m::EmptyOp>(op.getLoc(), currentInfo.type.getShape(),
-                                    currentInfo.type.getElementType(), layout)
-              .getResult();
+      auto maskedEmptyOp = rewriter.create<d2m::EmptyOp>(
+          op.getLoc(), currentInfo.type.getShape(),
+          currentInfo.type.getElementType(), layout);
+      if (layout) {
+        setVirtualGridMappingIfNeeded(maskedEmptyOp, layout, currentInfo.type);
```
nsmithtt: Wouldn't this be better handled as part of the op constructor?

vwellsTT: Yeah, I guess so.  I kind of don't like the initialization pattern of:
```
var x; // empty
if (cond) { x = a;} else { x = b; }
```
vs this
``` 
       var x = commonInit();
       if(!cond) {x.subfield = something;}
```
but I don't feel strongly, in general agree that constructing op correctly is better than modifying it after the fact.

nsmithtt: I meant having an EmptyOp builder that does `setVirtualGridMappingIfNeeded` inside of the builder.

vwellsTT: Ah, I see; I think that would make more sense

include/ttmlir/Dialect/D2M/IR/D2MOps.td
```diff
@@ -416,7 +416,13 @@ def D2M_EmptyOp
...
                       (ins "ArrayRef<int64_t>":$shape, "Type":$elementType), [{
       build($_builder, $_state, shape, elementType, nullptr);
-    }]>];
+    }]>,
+                  // Builder that computes and sets virtual grid mapping attrs
+                  // when the tensor's grid requires virtualization for the
+                  // given target device grid.
+                  OpBuilder<(ins "ArrayRef<int64_t>":$shape,
+                                 "Type":$elementType, "Attribute":$encoding,
+                                 "ArrayRef<int64_t>":$targetGridShape)>];
```
nsmithtt: Bad indentation

Issue #4872, opened 2025-09-10
[Bounty $100] Compare ops


tapspatel:
### Is your feature request related to a problem? Please describe

All ops found here: https://github.com/tenstorrent/tt-mlir/blob/ee1de78ea06b161973498036b3aeb3d196a213b5/lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp#L3196

Add the op implementation here: https://github.com/tenstorrent/tt-mlir/blob/main/tools/builder/stablehlo/stablehlo_builder.py
Sample: https://github.com/tenstorrent/tt-mlir/blob/main/tools/builder/ttir/ttir_builder.py

Add the respective golden function here: https://github.com/tenstorrent/tt-mlir/blob/main/tools/builder/base/builder_golden.py
see: https://github.com/tenstorrent/tt-mlir/blob/847a49ed9c9b9319a02c00b4e60c3c1437dc8df3/tools/builder/base/builder_golden.py#L2657

Add test case here: https://github.com/tenstorrent/tt-mlir/blob/main/test/python/golden/test_stablehlo_ops.py
see: https://github.com/tenstorrent/tt-mlir/blob/main/test/python/golden/test_ttir_ops.py

Gmin2:
can i take this @marty1885 ?

HariK3927:
@skhatriTT @marty1885 can you assign  to me?

marty1885:
@Gmin2 @HariK3927 Do not spam and keep in mind that we are all people, we don't need 3 messages across . Consider this a warning. I will ban if this behavior continue. As per out policy (see [Bounty TOS](https://docs.tenstorrent.com/bounty_terms.html) one person can only take on one bounty at a time. 

Would @Gmin2 Be ok taking on this one? I will assign @HariK3927 another.

lanchongyizu:
Hi @marty1885, could you assign this issue to me?

marty1885:
@lanchongyizu Not until Gmin2 gives up or enough time passes that we can reasonably assume he is no longer interested

aliraza556:
@marty1885 Is this one available to take? If so, I’d be happy to grab it. Thanks!

marty1885:
@aliraza556 nope. It is assigned

mudaly10-coder:
Auto-claim: I'm taking this on.
- Part of an automated bounty triage/claim workflow.
- If you'd like a different approach or more detail, please say so.

[...]

PR #7326, opened 2026-03-04
Add comparison operators (EQ, NE, GT, LT, GE, LE) to stablehlo_builder

toxfox69:
Implements issue #4872

Adds 6 comparison operator methods to the StableHLOBuilder class:
- eq(): Equal comparison
- ne(): Not equal comparison
- gt(): Greater than comparison
- lt(): Less than comparison
- ge(): Greater than or equal comparison
- le(): Less than or equal comparison

Each operator is implemented as a decorator-tagged method that delegates to a shared _compare_op() helper, which properly constructs stablehlo.CompareOp with the appropriate comparison direction and type attributes.

This enables StableHLO programs to use comparison operations through the high-level builder API.

### DIFF
--- a/tools/builder/stablehlo/stablehlo_builder.py
+++ b/tools/builder/stablehlo/stablehlo_builder.py
@@ -7914,3 +7914,112 @@ class StableHLOBuilder(Builder):
                                 sub_modules_and_builders.append(sub_op_module_builder)
 
         return sub_modules_and_builders
+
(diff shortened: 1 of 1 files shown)


jgrimTT:
This bounty is assigned to another dev. We only allow PRs from people assigned to the issue, and we assign those on a first-come-first-served basis. So I will be closing the issue, thanks!

Issue #7328, opened 2026-03-04
[Runtime vs EmitPy] Incorrect memory config on ttnn.to_layout op


svuckovicTT:
This IR runs thru runtime but fails thru emitpy (during execution of emitted TTNN python code).

Issue: runtime reads memory config object from parameter list of the op. For emitpy path, if the memory config parameter doesn't exist, we read it from the output tensor layout, which creates the difference, resulting in an error.

### Repro steps

```
# prerequisite: build and set up ttrt and tt-alchemist

# RUN THRU RUNTIME (works)
ttmlir-opt \
  --ttcore-register-device="system-desc-path=ttrt-artifacts/system_desc.ttsys" \
  --ttcore-wrap-device-module \
  repro_to_layout_bug.mlir -o q.mlir
ttmlir-translate --ttnn-to-flatbuffer q.mlir -o q.ttnn
ttrt run q.ttnn

# RUN THRU EMITPY (fails at running generated code)
tt-alchemist generate-python q.mlir -o q/
cd q
./run
```

Error msg:
```
Device: MeshDevice(1x1 grid, 1 devices)
2026-03-04 10:44:04.363 | critical |          Always | TT_FATAL: Physical shard shape (1, 64) must be tile {32, 32} sized! (assert.hpp:104)
Traceback (most recent call last):
  File "/localdev/svuckovic/_workspace/repos/tt-mlir/q/main.py", line 95, in <module>
    main()
  File "/localdev/svuckovic/_workspace/repos/tt-mlir/q/main.py", line 89, in main
    _main_0 = _main(create_inputs_for__main_0)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/localdev/svuckovic/_workspace/repos/tt-mlir/q/main.py", line 47, in _main
    ttnn_to_layout_1 = ttnn.to_layout(
                       ^^^^^^^^^^^^^^^
  File "/localdev/svuckovic/_workspace/repos/tt-mlir/third_party/tt-metal/src/tt-metal/ttnn/ttnn/decorators.py", line 451, in __call__
    result = self.function(*function_args, **function_kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #7309, opened 2026-03-03
[EmitPy] Add missing memory_config in TTNN->EmitPy op conversions

azecevicTT:
### Ticket
https://github.com/tenstorrent/tt-mlir/issues/7281

### Problem description
RepeatOp and few others didn't have memory config in conversion at all, or had conversion that could partially fail.

### Checklist
- [ ] New/Existing tests provide coverage for changes

PR #7329, opened 2026-03-04
[TTNN->EmitPy] Do NOT read mem cfg off output layout for to_layout op

svuckovicTT:
### Ticket
https://github.com/tenstorrent/tt-mlir/issues/7328

### Problem description
TTNN->EmitPy conversion for `ttnn.to_layout` creates memory config object from the output tensor layout. This is a common practice and is used across most (all?) other ops. Runtime path doesn't do this for this specific op, it only checks for memory config object from the existing op argument list.

This causes a mismatch in the way ops are called and (coincidentally) results in an assert firing for the Python path. Repro in the ticket

### What's changed
Removed the read of memory config from output tensor layout object.

### Checklist

### DIFF
--- a/lib/Conversion/TTNNToEmitPy/TTNNToEmitPy.cpp
+++ b/lib/Conversion/TTNNToEmitPy/TTNNToEmitPy.cpp
@@ -1151,6 +1151,9 @@ public:
         emitter.emit(toLayoutOp.getInput()),
         emitter.emit(toLayoutOp.getLayout()),
         emitter.emit(toLayoutOp.getDtype()),
+        // Emit MemoryConfig only if it exists as an argument to the op.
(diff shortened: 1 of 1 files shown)


svuckovicTT:
@sdjordjevicTT @azecevicTT @mtopalovicTT  Let's use tomorrow's tech sync to align on how we want to avoid these problems in the future.

svuckovicTT:
> @sdjordjevicTT @azecevicTT @mtopalovicTT Let's use tomorrow's tech sync to align on how we want to avoid these problems in the future.

https://github.com/tenstorrent/tt-mlir/pull/7186
https://github.com/tenstorrent/tt-mlir/pull/7309

^ just noting that there were two other PRs in the last couple days that deal with the same type of problem :)

PR #7345, opened 2026-03-04
[D2M][ttnn-jit] Support `CCL`s through jit

sgholamiTT:
### Summary
Adds support for collective communication (CCL) ops in the `ttnn-jit` tracing compiler: **all_gather**, **all_reduce**, and **reduce_scatter**. The tracer now emits the corresponding `ttir` CCL ops with correct output shapes based on mesh shape.

### Changes
- **jit_functions.py**: Add `AllGatherOpHandler`, `AllReduceOpHandler`, and `ReduceScatterOpHandler`; infer result shapes from `mesh_shape` (e.g. all_gather multiplies gather dim by `mesh_shape[cluster_axis]`, reduce_scatter divides scatter dim by it).
- **tracing_compiler.py**: Pass `mesh_shape` into the JIT context (from tensor args via `get_mesh_shape_from_tensor_args`).
- **utils.py**: Add `get_mesh_shape_from_tensor_args()` to read mesh shape from the first tensor’s device (`device.shape`), defaulting to `(1, 1)` for single device.
- **supported_ops.py**: Register `all_gather`, `all_reduce`, and `reduce_scatter` as supported ops.
- **test_tracing_ir.py**: Add basic CCL lit tests (all_gather, all_reduce, reduce_scatter, reduce_scatter max).
- **test_tracing_ir_ccl.py**: New lit test file covering more comprehensive ccl cases.

### Checklist

### DIFF
--- a/test/ttnn-jit/test_jit_type_hints.py
+++ b/test/ttnn-jit/test_jit_type_hints.py
@@ -1,0 +1,49 @@
+# SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+"""Tests for JIT-compiled functions that use ttnn type hints (e.g. ttnn.Tensor)."""
+
+import ttnn_jit
+import ttnn
+import torch
+import pytest
+
+from utils import create_dram_tensor
+
+
+@ttnn_jit.jit(compile_only=True)
+def single_exp_typed(input: ttnn.Tensor) -> ttnn.Tensor:
+    return ttnn.exp(input)
+
+
+@ttnn_jit.jit(compile_only=True)
+def add_typed(a: ttnn.Tensor, b: ttnn.Tensor) -> ttnn.Tensor:
+    return ttnn.add(a, b)
+
+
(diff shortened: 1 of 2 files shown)


tools/ttnn-jit/_src/utils.py
```diff
@@ -112,3 +112,25 @@ def get_core_grid_from_tensor_args(tensor_args):
...
+
+    For a mesh device, returns (mesh_rows, mesh_cols) from device.shape.
+    For a single device (no shape attribute or length < 2), returns (1, 1).
+    """
+    if not tensor_args:
+        return (1, 1)
+    tensor_arg = next(iter(tensor_args.values()))
+    device = tensor_arg.device()
+    try:
+        shape = getattr(device, "shape", None)
```
vtangTT: Why can't we do something like the following
```
    shape = device.shape
    if shape.dims() >= 2:
        return (shape[0], shape[1])
    return (1, 1)
```
Try catch seems unnecessary

sgholamiTT: Fixed.

PR #7348, opened 2026-03-04
[D2M] Cleanup and generalize DMA-Only Generic Op Support

bgrady-tt:
This PR resolves a bunch of special cases that existed for DMA only generic op. DMA-only generic ops did not require creation of stream_layout for remote operands accessed via DMA. This led to complex downstream code to handle all possible combinations of remote/local input.

DMA-only generics are now much less 'special' than previously. The only remaining special handling for DMA-only is that they are skipped during various compute-centric passes. They also require stream_layout for accessing remote operands, in similar fashion to 'regular' generic ops. There is some buffer allocation inefficiency here, but this should be optimizable when Streams 2.0 refactor is complete.

The remaining special handling logic is concentrated on the _forwarded load->store idiom_, where DMA passthrough is represented as a load whose result is exclusively consumed by a store. This is handled by having both remote_load and remote_store use the same CB; normally each load/store would use CB bound to their associated operand.


lib/Dialect/D2M/Transforms/Allocate.cpp
```diff
@@ -727,12 +728,22 @@ class D2MAllocate final : public impl::D2MAllocateBase<D2MAllocate> {
       // `operandCtx.defChain` and updates this memref's slot in
       // `analysis.memrefs`.
 
-      MemRefType memrefType = nullptr;
-      std::tie(operandCtx.root, memrefType, operandCtx.hasStream) =
-          analyzeOperandDefChain(genericOp, operand.get(), operandCtx.defChain);
+      FailureOr<OperandDefChainAnalysis> operandDefChainAnalysis =
```
bgrady-tt, line 731: The changes here are related to adding proper error handling if use-def analysis fails due to finding chains of view layout ops in non-canonical form. These _should_ have been canonicalized/folded properly earlier; it seems incorrect to try to fold them together during stream inference.

lib/Dialect/D2M/Transforms/LowerLoadStoreOpsToExplicitCBForm.cpp
```diff
@@ -260,36 +130,102 @@ static PushPopInfo convertToExplicitCBForm(ModuleOp moduleOp,
...
       }
+      if (failed(getLegalForwardableStore(remoteLoad, localBuffer,
+                                          forwardableStore))) {
+        return failure();
+      }
+    }
+
+    Value loadTargetCb = assocCb;
+    bool eraseForwardableStoreWithoutReplacement = false;
+    if (forwardableStore) {
```
bgrady-tt, line 168: This handles the case where only the input is streaming for a load->store forwardable pair (the store CB is aliased to a local CB). In this case the load directly targets the store's CB and then eliminates the redundant store op.

lib/Dialect/D2M/Transforms/Allocate.cpp
```diff
@@ -161,6 +161,13 @@ struct MemrefValueContext {
 
 using OperandDefChain = llvm::SmallVector<Operation *, 4>;
 
+struct OperandDefChainAnalysis {
+  Value root;
+  MemRefType type;
+  bool containsStream = false;
```
vwellsTT: nit: naming inconsistency bugs me here :)

lib/Dialect/D2M/Transforms/Allocate.cpp
vwellsTT: hmm, to me this seems like a case where it's an invariant => we *should* use assert over error emission

lib/Dialect/D2M/Transforms/LowerToLayout.cpp
```diff
@@ -1081,6 +1081,14 @@ class D2MLowerToLayoutRewriter : public OpRewritePattern<ToLayoutOp> {
       currentInfo = TensorInfo::from(currentValue);
     }
 
+    // 6b. DRAM→DRAM: Direct reblocking between DRAM buffers.
```
vwellsTT: nit: I think this should be full re-numbering

vwellsTT: but thanks for fixing this case, was on my TO-DO list!

vwellsTT:
A couple other notes: I think after my PR lands, I'm hoping to remove streams in fairly short order, rendering over-alloc concerns fairly minor.  Might be worth checking if extra memory pressure is breaking any nightly runs or anything though (I found some of them are pretty sensitive)

lib/Dialect/D2M/Transforms/Allocate.cpp
bgrady-tt: fair enough, I can fix this.

lib/Dialect/D2M/Transforms/Allocate.cpp
bgrady-tt: Hmm perhaps you are right. I am a little nervous about this not being checked in a release build, as it can easily cause a miscompile.

lib/Dialect/D2M/Transforms/LowerToLayout.cpp
bgrady-tt: I think I will remove the numbers here just to keep updating this similar. I don't think they serve much of a purpose anymore.

vwellsTT: Yeah, that seems less painful in the long term, good call

lib/Dialect/D2M/Transforms/LowerLoadStoreOpsToExplicitCBForm.cpp
```diff
@@ -260,36 +130,102 @@ static PushPopInfo convertToExplicitCBForm(ModuleOp moduleOp,
...
+    //   remote_load %buf %in[...]
+    //   remote_store %out[...] %buf
+    // Lower both to the same load-associated CB.
+    if (forwardableStore) {
+      auto pushOp = rewriter.create<PushOp>(loc, loadTargetCb);
+
+      rewriter.setInsertionPointAfter(pushOp);
+      auto waitOp = rewriter.create<WaitOp>(loc, loadTargetCb);
+
+      rewriter.setInsertionPointAfter(waitOp);
```
sohaibnadeemTT, line 208: Is there a drawback to inserting the explicit RemoteStoreOp in the same place of the original one (i.e. `rewriter.setInsertionPoint(forwardableStore)` instead of `rewriter.setInsertionPointAfter(waitOp)`). Since that would make sure the IR stays valid if it takes any operands that are defined just before the remote store (whereas if we put it after waitOp, we might get a use-before-define type error).

PR #7349, opened 2026-03-04
[TTIRFusing] Refactor linear fusing pattern

jserbedzijaTT:
## Summary

This PR paves the way for upcoming work on the `SplitQueryKeyValueAndSplitHeadsOp` fusing pattern and LHS-sharing matmul concatenations in attention layers. These downstream patterns require their producers to already be `LinearOp`s, rather than separate `MatmulOp + AddOp` pairs.

**Pass Ordering Conflict:** Currently, if the planned concatenation pattern fires before `MatmulWithBiasFusionPattern`, it merges the bare matmuls. This strands the bias additions, making them impossible to absorb into a fused linear op later.

**Abstraction Leak in TTIR:** The existing `MatmulWithBiasFusionPattern` had tt-metal-specific logic embedded at the TTIR level (checking if the padded bias second-to-last dimension matched `TILE_HEIGHT` of 32 to determine the output shape).

### What's changed

**TTIR Fusing (`TTIRFusing.cpp`)**: Rewrote `MatmulWithBiasFusionPattern` to remove the tt-metal tile-height workaround from the TTIR level. Moved the pattern to phase 1 of the fusing pass so it fires before other patterns.

**TTNN LinearOp verifier (`TTNNOps.cpp`)**: Updated to also accept the workarounded output shape alongside the mathematically correct broadcast shape. 

**TTNN LinearOp output shape workaround (`LinearOpOutputShapeRewritePattern.cpp`)**: Introduced a new workaround pattern that reshapes the LinearOp output from the broadcast shape to the matmul shape when tt-metal uses the fused kernel path.

### DIFF
--- a/include/ttmlir/Dialect/TTNN/Transforms/Workarounds/Decomposition/LinearOpOutputShapeRewritePattern.h
+++ b/include/ttmlir/Dialect/TTNN/Transforms/Workarounds/Decomposition/LinearOpOutputShapeRewritePattern.h
@@ -17,6 +17,7 @@ namespace mlir::tt::ttnn::workarounds::decomposition {
 // broadcasted shape. This pattern adjusts the LinearOp output from the
 // broadcasted shape to the matmul shape and inserts a ReshapeOp to restore the
 // original shape.
+// See: https://github.com/tenstorrent/tt-metal/issues/39392
 class LinearOpOutputShapeRewritePattern
     : public OpRewritePattern<ttnn::LinearOp> {
 public:


include/ttmlir/Dialect/TTNN/Transforms/Workarounds/Decomposition/LinearOpOutputShapeRewritePattern.h
```diff
@@ -0,0 +1,31 @@
...
+#include "mlir/IR/PatternMatch.h"
+#include "mlir/Support/LogicalResult.h"
+
+namespace mlir::tt::ttnn::workarounds::decomposition {
+
+// When the fused LinearOp kernel is used (padded bias second-to-last dim ==
+// TILE_HEIGHT), the hardware output shape is the matmul shape, not the
+// broadcasted shape. This pattern adjusts the LinearOp output from the
+// broadcasted shape to the matmul shape and inserts a ReshapeOp to restore the
+// original shape.
```
sdjordjevicTT, line 19: Do we have a metal issue for this?

jserbedzijaTT: Updated

Issue #7133, opened 2026-02-20
Fuse TopK op in TTNN using constraints API


odjuricicTT:
## Description

Add a fusing pattern in `TTNNFusing.cpp` that recognizes and fuses a sequence of sort + slice operations into a single `ttnn.topk` op.

### What the op does

`TopK` selects the top (or bottom) `k` elements along a given dimension from an input tensor. It returns two tensors: the `k` values and their original indices.

- **Input**: tensor `[B, ..., N]`, integer `k`, integer `dim`, bool `largest`
- **Output**: values tensor `[B, ..., K]`, indices tensor `[B, ..., K]` (uint32)

For example, given a `[1, 1, 32, 8192]` input with `k=50`, it returns `[1, 1, 32, 50]` values and indices tensors containing the top-50 elements per row.

This is critical for MoE (Mixture of Experts) router logic where the top-k experts need to be selected per token.

### Metal op reference

The underlying tt-metal implementation exists as `ttnn.topk`:
[`ttnn/operations/reduction/topk/`](https://github.com/tenstorrent/tt-metal/tree/main/ttnn/cpp/ttnn/operations/reduction/topk)

API signature: `ttnn.topk(input, k, dim=dim, largest=largest, sorted=sorted, sub_core_grids=None)`

### Dependencies

Blocked by:
- #7100 - Add topk op support to TTIR and TTNN dialects
- #3131 - Add TopK op (bounty)

sdjordjevicTT:
@odjuricicTT, since we added the op in the compiler, do you have some example TTIR that we can use to develop this?

odjuricicTT:
@sdjordjevicTT 
It should exist in the e2e gpt oss test here: https://github.com/tenstorrent/tt-xla/blob/main/tests/benchmark/test_llms.py#L877

Ir example:

```mlir
%values, %indices = "ttnn.sort"(%435) <{descending = true, dim = 1 : si8, stable = false}> : (tensor<32x32xbf16, #ttnn_layout40>) -> (tensor<32x32xbf16, #ttnn_layout40>, tensor<32x32xui16, #ttnn_layout107>) loc(#loc151)
"ttnn.deallocate"(%435) <{force = false}> : (tensor<32x32xbf16, #ttnn_layout40>) -> () loc(#loc151)
[...]

PR #7363, opened 2026-03-05
Adding TTNN fusing for TopK

sdjordjevicTT:
### Ticket
Solves https://github.com/tenstorrent/tt-mlir/issues/7133

### Problem description
The TTNN dialect was missing a pattern for fusing topk operation.

### What's changed
This PR introduces a new fusion pattern in the TTNN dialect that optimizes the common pattern of `sort` followed by two `slice` operations into a single `topk` operation. This optimization reduces the number of operations in the graph and improves performance by leveraging the native TopK implementation.

The pattern of sorting a tensor and then slicing to get the top-k elements is very common in machine learning workloads (e.g., beam search, top-k sampling, k-nearest neighbors). Previously, this required three separate operations:
1. `sort` - to order elements
2. `slice` on values - to extract top k values
3. `slice` on indices - to extract corresponding indices

With this fusion, these three operations are replaced by a single `topk` operation, which is more efficient.

### Input Pattern
```mlir
%values, %indices = "ttnn.sort"(%input) <{descending = true, dim = 1}>
%top_values = "ttnn.slice_static"(%values) <{begins = [0, 0], ends = [2, 3], step = [1, 1]}>
%top_indices = "ttnn.slice_static"(%indices) <{begins = [0, 0], ends = [2, 3], step = [1, 1]}>
```

### Output Pattern
```mlir
%top_values, %top_indices = "ttnn.topk"(%input) <{k = 3, dim = 1, largest = true, sorted = true}>
```

### Checklist

### DIFF
--- a/include/ttmlir/Dialect/TTNN/Transforms/Fusing/FusionValidator.h
+++ b/include/ttmlir/Dialect/TTNN/Transforms/Fusing/FusionValidator.h
@@ -34,9 +34,6 @@ struct FusionValidationConfig {
   /// Whether to apply layout workarounds.
   bool applyLayoutWorkarounds = true;
 
-  /// Maximum L1 memory usage cap (fraction, 1.0 = use all available).
-  float tensorL1UsageCap = 1.0f;
-
   /// Maximum number of fallback configurations to try (0 = unlimited).
   uint32_t maxFallbackAttempts = 10000;
 };
--- a/include/ttmlir/Dialect/TTNN/Transforms/Passes.td
+++ b/include/ttmlir/Dialect/TTNN/Transforms/Passes.td
@@ -239,11 +239,6 @@ def TTNNFusing: Pass<"ttnn-fusing", "::mlir::ModuleOp">
             "Enable op model constraint validation for fusion patterns. "
             "Requires TTMLIR_ENABLE_OPMODEL. When enabled, fused ops are "
             "validated in an isolated module before being applied.">,
-    Option<"tensorL1UsageCap",
-            "tensor-l1-usage-cap",
-            "float", /*default=*/"1.0",
-            "Limit L1 memory usage to this fraction of available space "
-            "(0.0-1.0). Only used when enable-op-constraints is true.">,
     Option<"maxFallbackAttempts",
             "max-fallback-attempts",
             "uint32_t", /*default=*/"10000",
--- a/lib/Dialect/TTNN/Transforms/Fusing/FusionValidator.cpp
+++ b/lib/Dialect/TTNN/Transforms/Fusing/FusionValidator.cpp
@@ -9,8 +9,6 @@
 #include "mlir/Pass/PassManager.h"
 #include "mlir/Transforms/Passes.h"
 
-#define DEBUG_TYPE "ttnn-fusion-validator"
-
 namespace mlir::tt::ttnn {
 
 FusionValidationResult FusionValidator::runValidationPipeline(ModuleOp module) {
@@ -35,7 +33,6 @@ FusionValidationResult FusionValidator::runValidationPipeline(ModuleOp module) {
   {
     PassManager pm(context);
     TTNNOperationValidationAndFallbackOptions validationOptions;
-    validationOptions.tensorL1UsageCap = config.tensorL1UsageCap;
     validationOptions.maxFallbackAttempts = config.maxFallbackAttempts;
     pm.addPass(mlir::tt::ttnn::createTTNNOperationValidationAndFallback(
         validationOptions));
--- a/lib/Dialect/TTNN/Transforms/Fusing/TopKFusingPattern.cpp
+++ b/lib/Dialect/TTNN/Transforms/Fusing/TopKFusingPattern.cpp
@@ -212,9 +212,11 @@ TopKFusing::matchAndRewrite(SortOp srcOp,
                   : computeTopKResultType(inputType, sortDim, sliceResult->k,
                                           inputType.getElementType());
   RankedTensorType indicesResultType =
-      indicesSlice ? mlir::cast<RankedTensorType>(indicesSlice.getType())
-                   : computeTopKResultType(inputType, sortDim, sliceResult->k,
-                                           rewriter.getI32Type());
+      indicesSlice
+          ? mlir::cast<RankedTensorType>(indicesSlice.getType())
+          : computeTopKResultType(inputType, sortDim, sliceResult->k,
+                                  IntegerType::get(rewriter.getContext(), 32,
+                                                   IntegerType::Signed));
 
   // Validate the fusion before creating it
   FusionValidator validator(rewriter.getContext(), validationConfig);
--- a/lib/Dialect/TTNN/Pipelines/TTNNPipelines.cpp
+++ b/lib/Dialect/TTNN/Pipelines/TTNNPipelines.cpp
@@ -160,13 +160,11 @@ void createTTNNFusingPass(OpPassManager &pm,
       wrapperOptions.devicePtr = options.devicePtr;
       wrapperOptions.tensorL1UsageCap = options.tensorL1UsageCap;
 
-      float l1Cap = options.tensorL1UsageCap;
       uint32_t fallbackAttempts = options.maxFallbackAttempts;
       pm.addPass(createDevicePassesWrapper(
-          [l1Cap, fallbackAttempts](OpPassManager &innerPm) {
+          [fallbackAttempts](OpPassManager &innerPm) {
             TTNNFusingOptions fusingOptions;
             fusingOptions.enableOpConstraints = true;
-            fusingOptions.tensorL1UsageCap = l1Cap;
             fusingOptions.maxFallbackAttempts = fallbackAttempts;
             innerPm.addPass(mlir::tt::ttnn::createTTNNFusing(fusingOptions));
           },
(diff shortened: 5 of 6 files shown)


lib/Dialect/TTNN/Transforms/Fusing/TopKFusingPattern.cpp
```diff
@@ -0,0 +1,167 @@
...
+
+  auto begins = llvm::cast<ArrayAttr>(beginsAttr);
+  auto ends = llvm::cast<ArrayAttr>(endsAttr);
+  auto steps = llvm::cast<ArrayAttr>(stepsAttr);
+
+  // Check that we're slicing from the beginning with unit step
+  auto beginValue = llvm::cast<IntegerAttr>(begins[dim]).getInt();
+  auto stepValue = llvm::cast<IntegerAttr>(steps[dim]).getInt();
+
+  if (beginValue != 0 || stepValue != 1) {
```
azecevic000: Other case would be when `endValue == n - 1` (or `n` if it's an exclusive range) and then `largest(topk) == smallest(sort)` (and vice versa).

sdjordjevicTT:
> Could you add a test with a graph that only returns one of the two outputs (i.e., either values or indices)? This is a common use case in Torch implementations (for ex: the DSA implementation has this).
> 
> An example of the type of graph I'm talking about:
> 
> ```
> module @SyncTensorsGraph.25 attributes {mhlo.cross_program_prefetches = [], mhlo.input_output_alias = [], mhlo.is_dynamic = false, mhlo.use_auto_spmd_partitioning = false, ttcore.meshes = #ttcore.meshes<[<"mesh" = 1x1>]>} {
>   ttcore.device_module {
>     builtin.module @SyncTensorsGraph.25 attributes {mhlo.cross_program_prefetches = [], mhlo.input_output_alias = [], mhlo.is_dynamic = false, mhlo.use_auto_spmd_partitioning = false, ttcore.meshes = #ttcore.meshes<[<"mesh" = 1x1>]>} {
>       func.func @main(%arg0: tensor<1x40xf32> {ttcore.argument_type = #ttcore.argument_type<input>, ttcore.local_shape = #ttcore<local_shape local_shape = tensor<1x40xf32>>, ttcore.shard_status = #ttcore.shard_status<unsharded>, ttir.name = "args_0"}) -> (tensor<1x5xi64> {ttcore.local_shape = #ttcore<local_shape local_shape = tensor<1x5xi64>>, ttcore.shard_status = #ttcore.shard_status<unsharded>}) {
>         %0 = "ttir.reshape"(%arg0) <{shape = [1 : i32, 1 : i32, 40 : i32]}> : (tensor<1x40xf32>) -> tensor<1x1x40xf32>
>         %1 = "ttir.reshape"(%0) <{shape = [1 : i32, 40 : i32]}> : (tensor<1x1x40xf32>) -> tensor<1x40xf32>
>         %2 = "ttir.arange"() <{arange_dimension = 0 : i64, end = 40 : si64, start = 0 : si64, step = 1 : si64}> : () -> tensor<40xi32>
>         %values, %indices = "ttir.sort"(%1) <{descending = true, dim = 1 : si32, stable = false}> : (tensor<1x40xf32>) -> (tensor<1x40xf32>, tensor<1x40xi32>)
>         %3 = "ttir.slice_static"(%indices) <{begins = [0 : i32, 0 : i32], ends = [1 : i32, 5 : i32], step = [1 : i32, 1 : i32]}> : (tensor<1x40xi32>) -> tensor<1x5xi32>
>         %4 = "ttir.typecast"(%3) <{conservative_folding = false}> : (tensor<1x5xi32>) -> tensor<1x5xi64>
>         return %4 : tensor<1x5xi64>
>       }
>     }
>   }
> }
> ```
> 
> This is the result of tracing a [Torch module](https://github.com/tenstorrent/tt-xla/blob/1a1020c15e99a37e471c45d4740436af2c8b7123/tests/torch/ops/test_topk.py#L43-L49) that uses only the indices output of the `torch.topk` op

@hshahTT, addressed this in the latest iteration, please take a look. Also added the pattern you posted as a test.

lib/Dialect/TTNN/Transforms/Fusing/TopKFusingPattern.cpp
```diff
@@ -0,0 +1,266 @@
...
+      nullptr                                     // memory_config (optional)
+  );
+
+  // Replace the slice operations with TopK results (only for used results)
+  if (valuesSlice) {
+    rewriter.replaceOp(valuesSlice, topkOp.getValues());
+  }
+  if (indicesSlice) {
+    rewriter.replaceOp(indicesSlice, topkOp.getIndices());
+  }
```
hshahTT, line 260: Could you add some logic to ensure that the output datatypes of the topk OP in IR match what the `ttnn.topk` op actually outputs?

From [TTNN documentation](https://docs.tenstorrent.com/tt-metal/latest/ttnn/ttnn/api/ttnn.topk.html):
- `values` should have the same datatype as the input tensor
- `indices` should always be UInt16

All tests in this PR have the returned `indices` tensor to be of type SI32, which is incorrect. I just ran this code on the tt-xla topk test and it failed due to this datatype mismatch.

hshahTT: From L210 it seems like `valuesResultType` is already set correctly. I think we just need to change L214 so that `indicesResultType` always has uint16 regardless of what the input type is. And if any of these outputs are outputs of the main graph function, then that datatype update should propagate to the function return type as well.

hshahTT: Discussed offline. This PR fixes the issue I talked about: https://github.com/tenstorrent/tt-mlir/pull/7324

include/ttmlir/Dialect/TTNN/Transforms/Passes.td
```diff
@@ -236,7 +236,20 @@ def TTNNFusing: Pass<"ttnn-fusing", "::mlir::ModuleOp">
...
             "bool", /*default=*/"false",
-            "When enabled fusing pass includes patterns which check for op constraints.">
+            "Enable op model constraint validation for fusion patterns. "
+            "Requires TTMLIR_ENABLE_OPMODEL. When enabled, fused ops are "
+            "validated in an isolated module before being applied.">,
+    Option<"tensorL1UsageCap",
+            "tensor-l1-usage-cap",
+            "float", /*default=*/"1.0",
+            "Limit L1 memory usage to this fraction of available space "
+            "(0.0-1.0). Only used when enable-op-constraints is true.">,
```
rpavlovicTT: You don't need this anymore. Use one from the pipeline options to pass to DevicePassWrapper and that's sufficient. `tensorL1UsageCap` will be removed from validation pass soon. I think @bmalesevicTT has that on her radar. I think you can just remove this option from all of your changes

lib/Dialect/TTNN/Transforms/Fusing/FusionValidator.cpp
```diff
@@ -0,0 +1,53 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+#include "ttmlir/Dialect/TTNN/Transforms/Fusing/FusionValidator.h"
+
+#include "ttmlir/Dialect/TTNN/Transforms/Passes.h"
+
+#include "mlir/Pass/PassManager.h"
+#include "mlir/Transforms/Passes.h"
+
+#define DEBUG_TYPE "ttnn-fusion-validator"
```
rpavlovicTT: Don't override this macro, it can mess up logs. If you want to enable your logs just run ttmlir-opt with -debug-only=fusion-validator

include/ttmlir/Dialect/TTNN/Transforms/Fusing/FusionValidator.h
```diff
@@ -0,0 +1,209 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+/// Configuration for fusion validation.
+struct FusionValidationConfig {
+  /// Whether to apply decomposition workarounds.
+  bool applyDecompositionWorkarounds = true;
+
+  /// Whether to apply layout workarounds.
+  bool applyLayoutWorkarounds = true;
```
rpavlovicTT, line 35: Do we need these? I assumed op fallback pass will perform layout workarounds if needed?

sdjordjevicTT: We have a couple of optimizer workarounds, hence didn't want to skip them. I added line bellow:
`workaroundOptions.optimizerEnabled = true;`

This will enable only optimizer specific layout workarounds.

sdjordjevicTT: Btw, agree that current options aren't most intuitive, i think we can create an issue to replace these configs with one enum config:
- NoLayoutWorkaorunds
- OptimizerOffLayoutWorkarounds
- OptimizerOnLayoutWorkarounds

Something like this...

rpavlovicTT: Yeah makes sense, and I'd even file an issue to remove those `OptimizerOnLayoutWorkarounds` if possible. I opened one for embedding op. Let's see if we can handle other workarounds through fallback pass.

include/ttmlir/Dialect/TTNN/Transforms/Passes.td
bmalesevicTT: Yes, thanks for the heads-up, removed in #7464

sdjordjevicTT: Thanks for pointing that out! Removed the tensorL1UsageCap option from this change.

include/ttmlir/Dialect/TTNN/Transforms/Fusing/FusionValidator.h
sdjordjevicTT: Create issue here:
https://github.com/tenstorrent/tt-mlir/issues/7465

PR #7365, opened 2026-03-05
Fix embedding op crash for dynamic weights with sharded inputs

rpavlovicTT:
Also reorder DecomposeLayouts to unshard (toMemoryConfig) before untilize (toLayout) when input is L1 sharded, since untilize_with_unpadding does not support sharded input with sharded output.

### DIFF
--- a/lib/Dialect/TTNN/Transforms/TTNNDecomposeLayouts.cpp
+++ b/lib/Dialect/TTNN/Transforms/TTNNDecomposeLayouts.cpp
@@ -694,10 +694,19 @@ private:
     // If the output data type is untilizable on device, untilize on device then
     // move to host
     if (info.shouldUntilize() && canUntilizeDataTypeOnDevice(input.dataType)) {
-      currentInput =
-          this->createToLayoutOpIfNeeded(op, rewriter, currentInput, info);
-      currentInput = this->createToMemoryConfigOpIfNeeded(op, rewriter,
-                                                          currentInput, info);
+      // If input is L1 sharded, unshard first since untilize doesn't support
+      // sharded input with sharded output.
+      if (input.isL1Sharded()) {
+        currentInput = this->createToMemoryConfigOpIfNeeded(op, rewriter,
+                                                            currentInput, info);
+        currentInput =
+            this->createToLayoutOpIfNeeded(op, rewriter, currentInput, info);
+      } else {
+        currentInput =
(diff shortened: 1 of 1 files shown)


include/ttmlir/Dialect/TTNN/IR/TTNNWorkaroundsPass.h
```diff
@@ -237,7 +237,10 @@ class TTNNOperandsWorkaroundsFactory {
   static TTNNOperandsWorkarounds createPool2DWithIndicesOpOperandsWorkarounds();
 
   // Create workarounds for embedding op operands.
-  static TTNNOperandsWorkarounds createEmbeddingOpOperandsWorkarounds();
+  // Only applies weight workaround for static weights (parameters or
+  // load_cached results). Dynamic weights are handled by the optimizer.
```
sdjordjevicTT: Will this affect execution when the optimizer is disabled? Instead of adding additional constraints here, could we disable the optimizer workaround entirely for the embedding op? As far as I understand, the fallback pass should already handle fixing the op.

In general, I’m against introducing additional layout-related constraints in workarounds. We already have a set of patterns that are only applied when the optimizer is active, so adding more conditions here could make the behavior harder to reason about.

rpavlovicTT: > Will this affect execution when the optimizer is disabled? 

Yes most likely. It can get tricky with opt-level 0. However, we are working on making our validation-fallback pass general and not-dependent on opt level. 

But your point is good, and I agree. This whole workaround was added just to increase performance because embedding op works better with RM weights (when weights tensor is huge). This op works with both Tile/RM weights, so this workaround is not for functionality at all. @djstefanovicTT has more context. 
So I did not want to revert it, because it would probably affect perf, but rather just disable explicit workaround when embedding op works with "normal" tensors (not param). 

Probably better solution is to enhance `TTNNRowMajorLayoutPropagation` pass and let it propagate weight tensor to embedding in RM layout. Today it doesn't because it only works with `input` args.

rpavlovicTT: I removed these changes in workaround patterns, opened a new issue https://github.com/tenstorrent/tt-mlir/issues/7434. And boiled down this PR only to TTNNDecomposeLayouts change.

sdjordjevicTT (approved):
Thanks!

PR #7380, opened 2026-03-06
Fix op conversions in TTNN->Emit[Py/C]

svuckovicTT:
### Ticket
Ad-hoc

### Problem description
Emitted TTNN Python code was producing different results to runtime when optimizer level 2 was used.

### What's changed
Fixed parameters used in ops when emitting to Py/C, either added/removed/fixed:
- compute configs
- memory configs
- output dtypes

### Checklist

### DIFF
--- a/lib/Conversion/TTNNToEmitC/TTNNToEmitC.cpp
+++ b/lib/Conversion/TTNNToEmitC/TTNNToEmitC.cpp
@@ -1157,6 +1157,8 @@ public:
         emitter.emit(embeddingOp.getWeight()),
         emitter.emit(std::nullopt),
         emitter.emit(layoutAttr),
+        emitc::OpaqueAttr::get(rewriter.getContext(),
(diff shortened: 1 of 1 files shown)


lib/Conversion/TTNNToEmitC/TTNNToEmitC.cpp
```diff
@@ -919,7 +917,7 @@ class MaxPool2dWithIndicesOpConversionPattern
         /*reallocate_halo_output=*/
         emitter.emit(srcOp.getReallocateHaloOutput()),
         /*return_indices=*/emitter.emit(true),
-        emitter.emit(/*dtype=*/ttcore::DataType::BFloat16),
+        emitter.emit(emitter.getOutputDtype(srcOp.getResult(0))),
```
vmilosevic: :warning: **clang\-diagnostic\-error** :warning:
too many arguments to function call, expected 0, have 1

Issue #7357, opened 2026-03-05
Fold `ttir.neg` op on constants


dmilinkovicTT:
We should be able to detect when `ttir.neg` op is executed on a constant tensor, and be able to fold fold(const) into a negated const tensor.

PR #7385, opened 2026-03-06
Implement constant folding for ttir.neg

amalkovTT:
### Ticket
#7357

### Problem description
Constant folding is not performed on ttir.neg operation with constant argument

### What's changed
Added folders for ttir.neg, ttir.full, ttir.ones, ttir.zeros.
Changed materializeConstant to create ttir.full, ttir.ones, ttir.zeros when possible.
Added tests for new canonicalizations.

### Checklist


lib/Dialect/TTIR/IR/TTIROps.cpp
```diff
@@ -809,6 +820,69 @@ mlir::tt::ttir::GetDimensionSizeOp::fold(FoldAdaptor adaptor) {
   return mlir::DenseElementsAttr::get<uint32_t>(resultType, dimSize);
 }
 
+//===----------------------------------------------------------------------===//
+// NegOp
+//===----------------------------------------------------------------------===//
+
+static mlir::Attribute negatedScalarAttribute(const mlir::Attribute &attr) {
```
dmilinkovicTT: MLIR types, such as `mlir::Attribute`, are meant to be used as value-types, since they are just thin wrappers around pointers. Passing them around as const references isn't *wrong* in terms of program correctness, but we should stick to using these as value types, since it's a more MLIR-idiomatic approach which avoids an additional indirection step.

lib/Dialect/TTIR/IR/TTIROps.cpp
```diff
@@ -809,6 +820,69 @@ mlir::tt::ttir::GetDimensionSizeOp::fold(FoldAdaptor adaptor) {
...
+  if (auto intAttr = llvm::dyn_cast_if_present<IntegerAttr>(attr)) {
+    auto attrValue = intAttr.getValue();
+    auto attrType = intAttr.getType();
+    return mlir::IntegerAttr::get(attrType, -attrValue);
+  }
+  return {};
+}
+
+// NegOp canonicalization: neg(full(x)) -> full(-x)
+void mlir::tt::ttir::NegOp::getCanonicalizationPatterns(
```
dmilinkovicTT: Great, the core logic seems to be correct!

However, there are scenarios where constant operands are going through ops which perform some transformations which should be "transparent"  to the logic of `NegOp` - for example, the input tensor can be reshaped, broadcasted, etc.

IIRC, there are some utilities in this file which you could use in order not to reinvent the wheel when it comes to handling these "transparent" ops.

lib/Dialect/TTIR/IR/TTIROps.cpp
```diff
@@ -809,6 +820,110 @@ mlir::tt::ttir::GetDimensionSizeOp::fold(FoldAdaptor adaptor) {
...
+        return mlir::success();
+      }
+
+      rewriter.replaceOpWithNewOp<mlir::tt::ttir::FullOp>(
+          op, resultType, shapeAttr, negatedFillValueAttr);
+      return mlir::success();
+    }
+
+    if (auto constant = mlir::dyn_cast_if_present<mlir::tt::ttir::ConstantOp>(
+            operand.getDefiningOp())) {
```
dmilinkovicTT: `operandOp` instead of `operand.getDefiningOp()`

dmilinkovicTT: `operandOp` instead of `operand.getDefiningOp()`

test/ttmlir/Dialect/TTIR/canonicalize/neg_fold_tests.mlir
```diff
@@ -0,0 +1,86 @@
...
+      value = dense<[[1, 2, 3], [4, 5, 6]]> : tensor<2x3xi32>
+    } : () -> tensor<2x3xi32>
+    // CHECK-LABEL: func.func @neg_constant
+    // CHECK-NOT: "ttir.neg"
+    // CHECK: "ttir.constant"
+    // CHECK-SAME: value = dense<{{\[\[}}-1, -2, -3], [-4, -5, -6{{\]\]}}>
+    %1 = "ttir.neg"(%0) : (tensor<2x3xi32>) -> tensor<2x3xi32>
+    return %1 : tensor<2x3xi32>
+  }
+}
```
dmilinkovicTT: We should also cover constant -> <transparent_op> -> neg scenario

lib/Dialect/TTIR/IR/TTIROps.cpp
```diff
@@ -809,6 +820,110 @@ mlir::tt::ttir::GetDimensionSizeOp::fold(FoldAdaptor adaptor) {
   return mlir::DenseElementsAttr::get<uint32_t>(resultType, dimSize);
 }
 
+//===----------------------------------------------------------------------===//
+// NegOp
+//===----------------------------------------------------------------------===//
+
+static mlir::Attribute negatedScalarAttribute(mlir::Attribute attr) {
```
dmilinkovicTT: nit: function names are generally more readable when they contain verbs in imperative, consider `negateScalarAttribute`

lib/Dialect/TTIR/IR/TTIROps.cpp
```diff
@@ -809,6 +820,110 @@ mlir::tt::ttir::GetDimensionSizeOp::fold(FoldAdaptor adaptor) {
...
+      for (llvm::APFloat val : denseAttr.getValues<llvm::APFloat>()) {
+        newValues.push_back(-val);
+      }
+      return DenseElementsAttr::get(tensorType, newValues);
+    }
+
+    if (llvm::isa<IntegerType>(elementType)) {
+      SmallVector<llvm::APInt> newValues;
+      newValues.reserve(denseAttr.getNumElements());
+      for (llvm::APInt val : denseAttr.getValues<llvm::APInt>()) {
```
dmilinkovicTT: nit: great that you've picked up the coding guidelines on overusing `auto`! 

but for scenarios such as this one, I think that we can relax the constraints and actually use `auto`

test/ttmlir/Dialect/TTIR/canonicalize/neg_fold_tests.mlir
dmilinkovicTT: Also, typecast is considered as a transparent op, so it might be good to cover scenarios where the element type changes

lib/Dialect/TTIR/IR/TTIRDialect.cpp
```diff
@@ -118,10 +118,51 @@ void TTIRDialect::initialize() {
 // TTIR constant materializer.
 //===----------------------------------------------------------------------===//
 
+// Check if fill_value attirbute of FullOp can be of this type.
+bool isValidFullValueType(Type t) {
+  return t.isF32() || t.isSignlessInteger(32);
```
dmilinkovicTT, line 123: why are signed integers not allowed here?

dmilinkovicTT: ah, I see, it's from the op definition.

PR #7391, opened 2026-03-06
[ttnn.jit][D2M] Add Support for DRAM Interleaved as Output Memory Config

### LINKED ISSUE #6735: [D2M] Correctly handle L1 Sharded -> DRAM Interleaved to_layout in LowerToLayout pass

arminaleTT:
### Ticket
Closes #6735 

### Problem description
The LowerToLayout Pass decomposes L1 sharded -> DRAM interleaved `d2m.to_layout` ops into L1 sharded -> L1 interleaved -> DRAM interleaved. This is incorrect, as D2M does not support L1 interleaved, and it is an unnecessary complication. Downstream DMA generation already supports going directly from L1 sharded to DRAM interleaved.

### What's changed
- Reorder `to_layout` decomposition:
  - Moved utilize and L1 -> DRAM above mapping changes in decomposition order
  - L1 sharded -> DRAM interleaved now creates a single DMA-only generic instead of two 
- For L1->DRAM DMA-only generics, set the generic grid VGM based on the L1 input, as the output will not have or require a VGM
- Modify generic op verification to exclude VGM verification for DRAM operands
  - We must explicitly check the memory space as opposed to checking if the producer is a view, as single tile DRAM tensors will have their view canonicalized away
- Fix bug in LowerLoadStoreOpsToExplicitCBForm
  - For L1 -> DRAM transfers, the pass incorrectly skipped inserting `d2m.reserve` and `d2m.push` for the corresponding `d2m.remote_store` 
  - Lower in the pipeline, this became a generic with a data movement kernel with a `cb_wait_front` but an empty compute kernel. This created a hang as there was no `cb_push_back`
- Change JIT DRAM tests to make outputs DRAM interleaved
- Added IR tests for expected `to_layout` decomposition and CB op insertion 

### Checklist

### DIFF
--- a/lib/Dialect/D2M/Transforms/LowerToLayout.cpp
+++ b/lib/Dialect/D2M/Transforms/LowerToLayout.cpp
@@ -858,7 +858,7 @@ public:
       return emptyOp.getResult();
     };
 
-    // 1. SYSTEM→DEVICE: Transfer to L1/DRAM with same element type as input.
+    // SYSTEM→DEVICE: Transfer to L1/DRAM with same element type as input.
     if (!currentInfo.hasLayout() && targetInfo.hasLayout()) {
       // System transfer can ONLY change memory space, not element type.
       // Create intermediate with scalar element type (same as system input).
@@ -879,7 +879,7 @@ public:
       currentInfo = TensorInfo::from(currentValue);
     }
 
-    // 2. DRAM→L1: Must happen before other device ops.
+    // DRAM→L1: Must happen before other device ops.
     // Use target's layout characteristics.
     if (currentInfo.hasLayout() && currentInfo.isDRAM() &&
         targetInfo.hasLayout() && !targetInfo.isDRAM()) {
@@ -901,7 +901,7 @@ public:
       currentInfo = TensorInfo::from(currentValue);
     }
 
-    // 3. TILIZE: Before mapping (so mapping operates on final format).
+    // TILIZE: Before mapping (so mapping operates on final format).
     bool needsTilize =
         !ttcore::isTiled(currentInfo.type) && ttcore::isTiled(targetInfo.type);
     if (needsTilize && currentInfo.hasLayout()) {
@@ -918,7 +918,7 @@ public:
       currentInfo = TensorInfo::from(currentValue);
     }
 
-    // 4. UNTILIZE: Before L1→DRAM or Device→System.
+    // UNTILIZE: Before L1→DRAM or Device→System.
     bool needsUntilize =
         ttcore::isTiled(currentInfo.type) && !ttcore::isTiled(targetInfo.type);
     if (needsUntilize) {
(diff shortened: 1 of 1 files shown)


lib/Dialect/D2M/IR/D2MOps.cpp
```diff
@@ -1754,20 +1754,36 @@ ::mlir::LogicalResult d2m::GenericOp::verify() {
              << deviceVolume << ")";
     }
 
+    auto isDRAM = [](Value output) {
+      if (auto memrefType = mlir::dyn_cast<MemRefType>(output.getType())) {
+        return ttcore::getMemorySpace(memrefType) ==
+               ttcore::MemorySpace::DeviceDRAM;
+      } else if (auto tensorType =
```
vmilosevic: :warning: **llvm\-else\-after\-return** :warning:
```suggestion
      } if (auto tensorType =
```

lib/Dialect/D2M/Transforms/LowerLoadStoreOpsToExplicitCBForm.cpp
```diff
@@ -399,12 +399,11 @@ static PushPopInfo convertToExplicitCBForm(ModuleOp moduleOp,
     rewriter.eraseOp(allocOp);
   }
 
-  // Transform RemoteStoreOp (implicit form -> explicit CB form)
+  // Transform RemoteStoreOp: convert implicit form to explicit CB form, and
+  // ensure explicit form stores (e.g., created by simplifyLoadStorePairs) get
+  // reserve/push inserted for CB synchronization.
   SmallVector<RemoteStoreOp> remoteStoresToConvert;
   moduleOp->walk([&](RemoteStoreOp remoteStore) {
-    if (!remoteStore.isImplicitForm()) {
```
arminaleTT, line 405: There is already an if and else down below handling implicit and explicit forms. This check was making the else branch dead code

lib/Dialect/D2M/Transforms/LowerToLayout.cpp
```diff
@@ -935,7 +967,7 @@ class D2MLowerToLayoutRewriter : public OpRewritePattern<ToLayoutOp> {
       currentInfo = TensorInfo::from(currentValue);
     }
 
-    // 5. MAPPING CHANGE: Grid/index_map/logical_shape/dim_alignments (after
```
bgrady-tt, line 938: Remove the numbering of the comments here.

PR #7394, opened 2026-03-06
[d2m] add int32 support for ops found in models

dlokeTT:
### Problem description
Some of the models we are targeting this quarter have ops that work on `i64` data which after `element-type-normalization` will turn them into `i32` so we need to support the i32 variants at the `ttkernel` level.

### What's changed
Added a host of changes in ttkernel to better natively support i32 variants of ops which usually have different init and tile functions
Added/removed skipping int32 builder tests for ttmetal for these common ops

### Checklist

### DIFF
--- a/lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
+++ b/lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
@@ -550,7 +550,7 @@ using ComputeOpMap = OpMap<
 // Int32 variants of compute ops. Ops not listed here have no int32 variant.
 using IntComputeOpMap = OpMap<
   // Unary SFPU (init unchanged, only the tile op differs).
-  std::pair<d2m::TileAbsOp,       std::pair<ttkernel::AbsTileInitOp,             ttkernel::AbsTileI32Op>>,
+  std::pair<d2m::TileAbsOp,        std::pair<ttkernel::AbsTileInitOp,             ttkernel::AbsTileI32Op>>,
   std::pair<d2m::TileNegativeOp,   std::pair<ttkernel::NegativeTileInitOp,        ttkernel::NegativeTileInt32Op>>,
   std::pair<d2m::TileReluOp,       std::pair<ttkernel::ReluTileInitOp,            ttkernel::ReluTileI32Op>>,
 
@@ -590,7 +590,7 @@ template <typename SrcOp, typename OpMap>
 using TTKernelOpPair = typename OpMapLookup<SrcOp, OpMap>::type;
 
 template <typename SrcOp, typename Map>
-constexpr bool hasIntMapping =
+constexpr bool hasMapping =
     !std::is_same_v<TTKernelOpPair<SrcOp, Map>, std::pair<void, void>>;
 
 // Some int32 TTKernel ops require an explicit dtype argument.
@@ -838,7 +838,7 @@ public:
         } else {
           rewriter.create<ttkernel::BinopWithScalarTileInitOp>(op->getLoc());
         }
-      } else if constexpr (hasIntMapping<ConcreteOp, IntComputeOpMap>) {
+      } else if constexpr (hasMapping<ConcreteOp, IntComputeOpMap>) {
         using IntInit =
             typename TTKernelOpPair<ConcreteOp, IntComputeOpMap>::first_type;
         const auto tileType =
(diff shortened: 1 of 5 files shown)


include/ttmlir/Dialect/TTKernel/IR/TTKernelOps.td
```diff
@@ -775,6 +881,37 @@ def TTKernel_MulBinaryTilesOp : TTKernel_SFPUOp<"mul_binary_tile", [TTKernel_Bin
     }];
 }
 
+def TTKernel_MulIntTileInitOp : TTKernel_InitOp<"mul_int_tile_init"> {
+    let summary = "Short init function";
+    let description = [{
+      Must be run before mul_int_tile.
+    }];
+
+    let arguments = (ins TTCore_DataTypeAttr:$dtype);
```
wenbinlyuTT, line 890: It's just baffling that this API has a dtype template arg while `add/sub_int_tile_init` doesn't 🤷, maybe they need to set some fidelity-related things.

lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
```diff
@@ -547,6 +547,28 @@ using ComputeOpMap = OpMap<
   // Elementwise SFPU Ternary.
   std::pair<d2m::TileWhereOp,       std::pair<ttkernel::WhereTileInitOp,           ttkernel::WhereTileOp>>
 >;
+// Int32 variants of compute ops. Ops not listed here have no int32 variant.
+using IntComputeOpMap = OpMap<
+  // Unary SFPU (init unchanged, only the tile op differs).
+  std::pair<d2m::TileAbsOp,       std::pair<ttkernel::AbsTileInitOp,             ttkernel::AbsTileI32Op>>,
```
wenbinlyuTT: Nit: alignment

lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
```diff
@@ -567,6 +589,17 @@ struct OpMapLookup<SrcOp, OpMap<std::pair<Key, Value>, Rest...>> {
 template <typename SrcOp, typename OpMap>
 using TTKernelOpPair = typename OpMapLookup<SrcOp, OpMap>::type;
 
+template <typename SrcOp, typename Map>
+constexpr bool hasIntMapping =
```
wenbinlyuTT: Nit: there's nothing `Int` about this lookup, maybe just call it `hasMapping`

lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
```diff
@@ -824,84 +871,51 @@ class D2MSFPUOpsRewriter : public OpConversionPattern<ConcreteOp> {
...
+        auto floatAttr = mlir::cast<FloatAttr>(attr);
+        auto floatVal = rewriter.create<arith::ConstantOp>(
+            loc,
+            rewriter.getF32FloatAttr(floatAttr.getValue().convertToFloat()));
+        return rewriter.create<arith::BitcastOp>(loc, rewriter.getI32Type(),
+                                                 floatVal);
+      };
+      auto minParam = toI32Param(op.getMinAttr());
+      auto maxParam = toI32Param(op.getMaxAttr());
+      if (mlir::isa<IntegerAttr>(op.getMinAttr())) {
```
wenbinlyuTT: IMO we can split the `toI32Param` lambda above into two and put them in the corresponding branch here. Personally I like the logic to happen as close to where it's used as possible, and consolidate if-else branches.

lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
```diff
@@ -824,84 +871,51 @@ class D2MSFPUOpsRewriter : public OpConversionPattern<ConcreteOp> {
...
-    } else if constexpr (std::is_same_v<SFPUOp, ttkernel::EqzTileOp> ||
-                         std::is_same_v<SFPUOp, ttkernel::NezTileOp> ||
-                         std::is_same_v<SFPUOp, ttkernel::GtzTileOp> ||
-                         std::is_same_v<SFPUOp, ttkernel::GezTileOp> ||
-                         std::is_same_v<SFPUOp, ttkernel::LtzTileOp> ||
-                         std::is_same_v<SFPUOp, ttkernel::LezTileOp>) {
+    } else if constexpr (arity == 1 &&
+                         hasIntMapping<ConcreteOp, IntComputeOpMap>) {
+      using IntSFPU =
+          typename TTKernelOpPair<ConcreteOp, IntComputeOpMap>::second_type;
```
wenbinlyuTT, line 915: We can statically assert `!needsDtypeArg<IntSFPU>` here I think, just in case someone did a partial update for more compute API change in the future.

lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
```diff
@@ -824,84 +871,51 @@ class D2MSFPUOpsRewriter : public OpConversionPattern<ConcreteOp> {
...
       }
-    } else if constexpr (std::is_same_v<SFPUOp, ttkernel::EqzTileOp> ||
-                         std::is_same_v<SFPUOp, ttkernel::NezTileOp> ||
-                         std::is_same_v<SFPUOp, ttkernel::GtzTileOp> ||
-                         std::is_same_v<SFPUOp, ttkernel::GezTileOp> ||
-                         std::is_same_v<SFPUOp, ttkernel::LtzTileOp> ||
-                         std::is_same_v<SFPUOp, ttkernel::LezTileOp>) {
+    } else if constexpr (arity == 1 &&
+                         hasIntMapping<ConcreteOp, IntComputeOpMap>) {
+      using IntSFPU =
```
wenbinlyuTT: We should call this `IntSFPUOp`, at first glance I thought this is a typo of `InitSFPU`...

test/python/golden/ttir_ops/eltwise/test_ttir_ternary.py
```diff
@@ -154,14 +156,17 @@ def where(
 
 @pytest.mark.parametrize("shape", [(64, 128)], ids=shape_str)
 @pytest.mark.parametrize("max_arg,min_arg", [(0.8, -0.5)])
+@pytest.mark.parametrize("dtype", [torch.float32, torch.bfloat16], ids=["f32", "bf16"])
```
wenbinlyuTT: We should merge this test with the i32 test below and just select min & max depending on the data type, we're using one pair for each type anyway.

wenbinlyuTT (commented):
All looks good, only a few nits.
A few lit tests to improve test coverage would also be nice, it is a great tool to catch catastrophic lowering bugs in day-to-day dev workflow.

include/ttmlir/Dialect/TTKernel/IR/TTKernelOps.td
dlokeTT: Yeah I agree. Hopefully as llks get refactored, this gets cleaned up during uplifts

PR #7407, opened 2026-03-09
Complex type conversion and stablehlo complex/ real/ imag op support

ddilbazTT:
### Ticket
[Link to Github Issue](https://github.com/tenstorrent/tt-xla/issues/3060)

### Problem description
Our compiler does not have complex data type and complex operation support ([stablehlo.complex](https://openxla.org/stablehlo/spec#complex), [stablehlo.real](https://openxla.org/stablehlo/spec#real), [stablehlo.imag](https://openxla.org/stablehlo/spec#imag)) 

### What's changed
Complex values are stored in memory as real and imaginary interleaved: [re0, im0, re1, im1, ...] Therefore, they can be unpacked with a trailing dim 2. For example: tensor<1x3xcomplex<f32>> → tensor<1x3x2xf32>

Adding a trailing dim of 2 will result in performance degradation due to tilize/ untilize. Therefore, in conversion patterns, we need to include a transpose/ permute to move the trailing dim to 0th dim. For example: tensor<1x3x2xf32> --> tensor<2x1x3xf32>

If we add support for the following stablehlo complex ops, we would rewrite other standard ops with complex data type: [stablehlo.complex](https://openxla.org/stablehlo/spec#complex), [stablehlo.real](https://openxla.org/stablehlo/spec#real), [stablehlo.imag](https://openxla.org/stablehlo/spec#imag)

We can cover conversion in stablehlo to ttir pipeline with two passes:
- StablehloComplexMathExpander: Currently, I wrote a temp patch for mult op since that's what we're seeing in Deepseek v3.2 RoPE. However, I intend to make upstream changes in stablehlo: https://github.com/openxla/stablehlo/issues/2913 
- StableHLOComplexDataTypeConversionPass: This pass replaces complex values with their unpacked versions, rewrites reshape and broadcast_in_dim ops with the unpacked complex values, and rewrites stablehlo.complex, stablehlo.real, stablehlo.imag ops using existing ops.
    - stablehlo.complex:
        - reshape/ unsqueeze real and imaginary to add a prepending dim 1: tensor<3xf32> --> tensor<1x3xf32>
        - concatenate along dim 0: tensor<1x3xf32>, tensor<1x3xf32> --> tensor<2x3xf32>
        - transpose: tensor<2x3xf32> --> tensor<3x2xf32>
    - stablehlo.real/ imag:
        - transpose: tensor<3x2xf32> --> tensor<2x3xf32>
        - slice along dim 0: tensor<2x3xf32> --> tensor<1x3xf32>
        - reshape/ squeeze: tensor<1x3xf32> --> tensor<3xf32>

### DIFF
--- a/lib/Dialect/StableHLO/Transforms/ComplexDataTypeConversion.cpp
+++ b/lib/Dialect/StableHLO/Transforms/ComplexDataTypeConversion.cpp
@@ -1,0 +1,363 @@
+// SPDX-FileCopyrightText: (c) 2024 Tenstorrent AI ULC
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#include "ttmlir/Dialect/StableHLO/Transforms/Passes.h"
+
+#include "mlir/Dialect/Func/IR/FuncOps.h"
+#include "mlir/Dialect/Func/Transforms/FuncConversions.h"
+#include "mlir/IR/BuiltinDialect.h"
+#include "mlir/IR/BuiltinTypes.h"
+#include "mlir/IR/PatternMatch.h"
+#include "mlir/Support/LogicalResult.h"
+#include "mlir/Transforms/DialectConversion.h"
+#include "stablehlo/dialect/StablehloOps.h"
+
+using namespace mlir;
+
+namespace mlir::tt::stablehlo {
+#define GEN_PASS_DEF_STABLEHLOCOMPLEXDATATYPECONVERSIONPASS
+#include "ttmlir/Dialect/StableHLO/Transforms/Passes.h.inc"
+
+//===----------------------------------------------------------------------===//
+// ComplexDataTypeConversion overview
+//===----------------------------------------------------------------------===//
+// Runs after `stablehlo-complex-math-expander` (real-domain elementwise ops).
+//
+//  (1) Unpack complex dtypes — append trailing dim of 2: [re0, im0, re1, ...]
+//        tensor<4xcomplex<f32>>    -->  tensor<4x2xf32>
+//        tensor<3x4xcomplex<f32>>  -->  tensor<3x4x2xf32>
+//      Affected: func args/returns, constants, reshape, broadcast_in_dim.
+//
+//  (2) Decompose complex/real/imag --> slice/concat/reshape.
+//      Trailing dim of 2 is not tile-divisible; decompositions transiently
+//      move it to the leading position before operating on it.
+//
+//===----------------------------------------------------------------------===//
+
+// ---------------------------------------------------------------------------
+// Transpose helpers
+// ---------------------------------------------------------------------------
+
+// Moves the trailing dimension to the front:
+//   tensor<d0 x d1 x ... x dN>  -->  tensor<dN x d0 x d1 x ... x d(N-1)>
+static Value transposeTrailingToLeading(Location loc, Value input,
+                                        OpBuilder &builder) {
+  auto type = mlir::cast<RankedTensorType>(input.getType());
+  int64_t rank = type.getRank();
+
+  SmallVector<int64_t> perm;
+  perm.push_back(rank - 1);
+  for (int64_t i = 0; i < rank - 1; ++i) {
+    perm.push_back(i);
+  }
+
+  SmallVector<int64_t> newShape;
+  newShape.push_back(type.getShape()[rank - 1]);
+  for (int64_t i = 0; i < rank - 1; ++i) {
+    newShape.push_back(type.getShape()[i]);
+  }
+
+  auto newType = RankedTensorType::get(newShape, type.getElementType());
+  return builder
+      .create<mlir::stablehlo::TransposeOp>(loc, newType, input,
+                                            builder.getDenseI64ArrayAttr(perm))
+      .getResult();
+}
+
+// Moves the leading dimension to the back:
+//   tensor<d0 x d1 x ... x dN>  -->  tensor<d1 x d2 x ... x dN x d0>
+static Value transposeLeadingToTrailing(Location loc, Value input,
+                                        OpBuilder &builder) {
+  auto type = mlir::cast<RankedTensorType>(input.getType());
+  int64_t rank = type.getRank();
+
+  SmallVector<int64_t> perm;
+  for (int64_t i = 1; i < rank; ++i) {
+    perm.push_back(i);
+  }
+  perm.push_back(0);
+
+  SmallVector<int64_t> newShape;
+  for (int64_t i = 1; i < rank; ++i) {
+    newShape.push_back(type.getShape()[i]);
+  }
+  newShape.push_back(type.getShape()[0]);
+
+  auto newType = RankedTensorType::get(newShape, type.getElementType());
+  return builder
+      .create<mlir::stablehlo::TransposeOp>(loc, newType, input,
+                                            builder.getDenseI64ArrayAttr(perm))
+      .getResult();
+}
+
+// ---------------------------------------------------------------------------
+// Conversion patterns
+// ---------------------------------------------------------------------------
+
+// Decomposes complex(re, im) -> tensor<...x2xf32>:
+//
+//   re: tensor<d0x...xdN>    im: tensor<d0x...xdN>
+//       |                            |
+//    reshape                      reshape
+//       |                            |
+//   tensor<1xd0x...xdN>      tensor<1xd0x...xdN>
+//       \                           /
+//              concatenate(dim=0)
+//                     |
+//          tensor<2xd0x...xdN>   <- [re_slice, im_slice, ...]
+//                     |
+//           transposeLeadingToTrailing
+//                     |
+//          tensor<d0x...xdNx2>   <- unpacked complex layout
+//
+namespace {
+class StablehloComplexToDecomposedPattern
+    : public OpConversionPattern<mlir::stablehlo::ComplexOp> {
+  using OpConversionPattern<mlir::stablehlo::ComplexOp>::OpConversionPattern;
+
+public:
+  LogicalResult
+  matchAndRewrite(mlir::stablehlo::ComplexOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    Location loc = op.getLoc();
+    auto lhsType = mlir::cast<RankedTensorType>(adaptor.getLhs().getType());
+
+    SmallVector<int64_t> unsqueezedShape;
+    unsqueezedShape.push_back(1);
+    for (auto d : lhsType.getShape()) {
+      unsqueezedShape.push_back(d);
+    }
+    auto unsqueezedType =
+        RankedTensorType::get(unsqueezedShape, lhsType.getElementType());
+    auto reshapedLhs = rewriter.create<mlir::stablehlo::ReshapeOp>(
+        loc, unsqueezedType, adaptor.getLhs());
+    auto reshapedRhs = rewriter.create<mlir::stablehlo::ReshapeOp>(
+        loc, unsqueezedType, adaptor.getRhs());
+
+    SmallVector<int64_t> concatShape;
+    concatShape.push_back(2);
+    for (auto d : lhsType.getShape()) {
+      concatShape.push_back(d);
+    }
+    auto concatType =
+        RankedTensorType::get(concatShape, lhsType.getElementType());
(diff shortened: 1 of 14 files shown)


lib/Dialect/StableHLO/Transforms/ComplexDataTypeConversion.cpp
```diff
@@ -0,0 +1,450 @@
...
+        op, resultType, sliceOp.getResult());
+    return success();
+  }
+};
+
+// stablehlo.imag(operand: tensor<...x2xfN>):
+//   1. Transpose trailing dim to front → [2, ...]
+//   2. Slice dim 0 at [1, 2)           → [1, ...]
+//   3. Reshape to squeeze the size-1   → [...]
+struct StablehloImagToDecomposedPattern
```
svuckovicTT: `StablehloRealToDecomposedPattern` and `StablehloImagToDecomposedPattern` are the same patterns, with the only difference being the slice indices. I'd suggest unifying then to a single implementation and having each of the patterns call the implementation. You could even templatize, like so:

svuckovicTT:
```C++
template <typename SourceOp, int64_t SliceOffset>
struct StablehloComplexPartDecomposedPattern
    : public OpConversionPattern<SourceOp> {
  using OpConversionPattern<SourceOp>::OpConversionPattern;

  LogicalResult matchAndRewrite(
      SourceOp op, 
      typename OpConversionPattern<SourceOp>::OpAdaptor adaptor,
    Location loc = op.getLoc();

    // Step 1: transpose [..., 2] → [2, ...]
    auto transposed =
        transposeTrailingToLeading(loc, adaptor.getOperand(), rewriter);
    auto transposedType = mlir::cast<RankedTensorType>(transposed.getType());
    int64_t rank = transposedType.getRank();
    auto transposedShape = transposedType.getShape();

    // Step 2: slice dim 0 at [SliceOffset, SliceOffset + 1) → plane, shape [1, ...]
    SmallVector<int64_t> begins(rank, 0),
        ends(transposedShape.begin(), transposedShape.end()), steps(rank, 1);
    begins[0] = SliceOffset;
    ends[0] = SliceOffset + 1;
    SmallVector<int64_t> sliceShape(transposedShape.begin(),
                                    transposedShape.end());
    sliceShape[0] = 1;
    auto sliceOp = rewriter.create<mlir::stablehlo::SliceOp>(
        loc, RankedTensorType::get(sliceShape, transposedType.getElementType()),
        transposed, rewriter.getDenseI64ArrayAttr(begins),
        rewriter.getDenseI64ArrayAttr(ends),
        rewriter.getDenseI64ArrayAttr(steps));

    // Step 3: squeeze size-1 leading dim away → [...]
    // Note: Use `this->getTypeConverter()` instead of `getTypeConverter()` 
    // because it is a dependent base class method.
    auto resultType = mlir::cast<RankedTensorType>(
        this->getTypeConverter()->convertType(op.getResult().getType()));
    rewriter.replaceOpWithNewOp<mlir::stablehlo::ReshapeOp>(
        op, resultType, sliceOp.getResult());
    return success();
  }
};

using StablehloRealToDecomposedPattern = 
    StablehloComplexPartDecomposedPattern<mlir::stablehlo::RealOp, 0>;
using StablehloImagToDecomposedPattern = 
    StablehloComplexPartDecomposedPattern<mlir::stablehlo::ImagOp, 1>;
```

lib/Dialect/StableHLO/Transforms/ComplexDataTypeConversion.cpp
```diff
@@ -0,0 +1,450 @@
...
+  matchAndRewrite(mlir::stablehlo::ReshapeOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    auto resultType = mlir::cast<RankedTensorType>(op.getResult().getType());
+    if (!mlir::isa<mlir::ComplexType>(resultType.getElementType())) {
+      return rewriter.notifyMatchFailure(op, "not a complex element type");
+    }
+
+    Location loc = op.getLoc();
+
+    // Step 1: transpose [..., 2] → [2, ...]
```
svuckovicTT: I wonder why we do these transposes (step 1 and 3).

```py
import torch
q = torch.arange(24).reshape(4, 3, 2)
torch.equal(q.permute(2, 0, 1).reshape(2, 3, 4).permute(1, 2, 0), q.reshape(3, 4, 2))
```
returns `True`

svuckovicTT: Ah, because it makes sense for TTNN later on, this is the whole idea of the pattern... Might be of value to add the reasoning in the comments

lib/Dialect/StableHLO/Transforms/ComplexDataTypeConversion.cpp
```diff
@@ -0,0 +1,450 @@
...
+    target.addIllegalOp<mlir::stablehlo::ImagOp>();
+
+    target.addDynamicallyLegalOp<mlir::stablehlo::ReshapeOp>(
+        [](mlir::stablehlo::ReshapeOp op) {
+          auto resultType =
+              mlir::cast<RankedTensorType>(op.getResult().getType());
+          return !mlir::isa<mlir::ComplexType>(resultType.getElementType());
+        });
+
+    target.addDynamicallyLegalOp<mlir::stablehlo::BroadcastInDimOp>(
```
svuckovicTT: Would this work?

```C++
target.addDynamicallyLegalOp<mlir::stablehlo::ConstantOp,
                             mlir::stablehlo::ReshapeOp,
                             mlir::stablehlo::BroadcastInDimOp>(
    [](mlir::Operation *op) {
      auto resultType = mlir::cast<mlir::RankedTensorType>(op->getResult().getType());
      return !mlir::isa<mlir::ComplexType>(resultType.getElementType());
    });
```

Another option:
```C++
auto isNotComplex = [](auto op) {
  auto resultType = mlir::cast<mlir::RankedTensorType>(op.getResult().getType());
  return !mlir::isa<mlir::ComplexType>(resultType.getElementType());
};

target.addDynamicallyLegalOp<mlir::stablehlo::ConstantOp>(isNotComplex);
target.addDynamicallyLegalOp<mlir::stablehlo::ReshapeOp>(isNotComplex);
target.addDynamicallyLegalOp<mlir::stablehlo::BroadcastInDimOp>(isNotComplex);
```

lib/Dialect/StableHLO/Transforms/ComplexDataTypeConversion.cpp
```diff
@@ -0,0 +1,450 @@
...
+    patterns.add<StablehloComplexToDecomposedPattern>(typeConverter,
+                                                      &getContext());
+    patterns.add<StablehloRealToDecomposedPattern>(typeConverter,
+                                                   &getContext());
+    patterns.add<StablehloImagToDecomposedPattern>(typeConverter,
+                                                   &getContext());
+    patterns.add<ReshapeComplexConversionPattern>(typeConverter, &getContext());
+    patterns.add<BroadcastInDimComplexConversionPattern>(typeConverter,
+                                                         &getContext());
+
```
svuckovicTT, line 345:
```C++
  patterns.add<ConstantComplexConversionPattern,
               StablehloComplexToDecomposedPattern,
               StablehloRealToDecomposedPattern,
               StablehloImagToDecomposedPattern,
               ReshapeComplexConversionPattern,
               BroadcastInDimComplexConversionPattern>(typeConverter, &getContext());
```

lib/Dialect/StableHLO/Transforms/ComplexDataTypeConversion.cpp
```diff
@@ -0,0 +1,450 @@
...
+          StableHLOComplexDataTypeConversionPass> {
+  using impl::StableHLOComplexDataTypeConversionPassBase<
+      StableHLOComplexDataTypeConversionPass>::
+      StableHLOComplexDataTypeConversionPassBase;
+
+  void runOnOperation() override {
+    mlir::ConversionTarget target(getContext());
+    target.addLegalDialect<mlir::stablehlo::StablehloDialect>();
+    target.addLegalDialect<mlir::func::FuncDialect>();
+    target.addLegalDialect<BuiltinDialect>();
```
svuckovicTT: Do we need these with partial conversion? My understanding is that all the existing ops that don't go thru a conversion pass are legalized in partial conversion, but it wouldn't be the first time the docs lied 😅

ddilbazTT: I get this error without `target.addLegalDialect<mlir::stablehlo::StablehloDialect>();`
```
failed to legalize operation 'stablehlo.complex' that was explicitly marked illegal
```

However, FuncDialect and BuiltinDialect are not needed so removing them! Thanks!

lib/Dialect/StableHLO/Transforms/ComplexDataTypeConversion.cpp
ddilbazTT: Great idea, thanks!

lib/Dialect/StableHLO/Transforms/ComplexDataTypeConversion.cpp
ddilbazTT: For sure, I need to polish the comments.

lib/Dialect/StableHLO/Transforms/ComplexDataTypeConversion.cpp
ddilbazTT: Templated real and imaginary conversion patterns!

ddilbazTT:
> Regarding patterns, I have a question about `ReshapeOp` and `BroadcastInDimOp`. All other patterns seem necessary for what we want to achieve, but these two seem unnecessary, but maybe I'm missing something.
> 
> For example for reshape, if we had:
> 
> ```mlir
> stablehlo.reshape(...) <SHAPExcomplex<f32>> -> <OTHER_SHAPExcomplex<f32>>
> ```
> 
> wouldn't type converter and conversions on necessary ops ensure that we now have
> 
> ```mlir
> stablehlo.reshape(...) <SHAPEx2xf32> -> <OTHER_SHAPEx2xf32>
> ```
> 
> which is correct.

@azecevicTT Thanks for this question! I would like to double-check my understanding on this as well.

If I'm not mistaken, type conversion defines the mapping rule whereas the conversion pattern rewrites the IR with this rule. 

The rewrite rule for reshape op is 1:1. However, for broadcast_in_dim, we do need to change the broadcast dimensions attribute. I.e.

```
%9 = stablehlo.broadcast_in_dim %8, dims = [1, 3] : (tensor<16x8xcomplex<f32>>) -> tensor<2x16x4x8xcomplex<f32>>
```

When we rewrite 2x16x4x8 as 2x16x4x8x2, dims needs to change, like:

```
%12 = stablehlo.broadcast_in_dim %11, dims = [1, 3, 4] : (tensor<16x8x2xf32>) -> tensor<2x16x4x8x2xf32>
```

So I added a conversion for both of these operations. I am editing them to be cleaner. For reshape it's a 1:1 but for broadcast_in_dim, we need to edit the dims attribute. 

Edit: I think I might have misunderstood the original question. The earlier conversion patterns of reshape and broadcast_in_dim did some unnecessary transpose that hopefully, I have now fixed. Currently, I defined a default conversion pattern which reshape uses, since it's just type conversion. I defined a broadcast specific conversion pattern since it needs to update attributes. Thanks for the feedback! @azecevicTT

lib/Dialect/StableHLO/Transforms/ComplexDataTypeConversion.cpp
svuckovicTT: Ah, that's because complexop is marked as illegal couple lines below. And the error msg is saying that there are some complex ops in the IR after the conversion, which sounds like a bug?

ddilbazTT: Correct me if I'm wrong about this. Aren't all dialects illegal by default unless we define them as legal?

svuckovicTT: That'd be true in the case of full conversion, but you're using partial conversion here, which allows all the pre-existing ops to "pass" if not explicitly marked illegal.

https://mlir.llvm.org/docs/DialectConversion/#modes-of-conversion

PR #7416, opened 2026-03-09
Account for tile count divisibility in AllReduce workaround

btsoiTT:
### Ticket
https://github.com/tenstorrent/tt-xla/issues/3249

### Problem description
AllReduce on tensors where the chosen scatter dimension has an odd tile count crashes at runtime.

### What's changed
Fixed `TTNNAllReduceWorkarounds` to check tile-count divisibility when selecting a scatter dimension for ReduceScatter. Also increase the memory limit for AllReduceGather to 0.12.

### DIFF
--- a/test/ttmlir/Dialect/TTNN/ccl/all_reduce/all_reduce_positive.mlir
+++ b/test/ttmlir/Dialect/TTNN/ccl/all_reduce/all_reduce_positive.mlir
@@ -116,12 +116,12 @@ module attributes {} {
 // Verify scatter/all_gather use the last divisible dimension (dim 2 here).
 module attributes {} {
   // CHECK-LABEL: all_reduce_default_dims_last_divisible_dim2
-  func.func @all_reduce_default_dims_last_divisible_dim2(%arg0: tensor<1x1x32x255xbf16>) -> tensor<1x1x32x255xbf16> {
-    %1 = "ttir.all_reduce"(%arg0) <{cluster_axis = 1 : ui32, reduce_type = #ttcore.reduce_type<sum>}> : (tensor<1x1x32x255xbf16>) -> tensor<1x1x32x255xbf16>
+  func.func @all_reduce_default_dims_last_divisible_dim2(%arg0: tensor<1x1x64x255xbf16>) -> tensor<1x1x64x255xbf16> {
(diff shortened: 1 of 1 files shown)


lib/Dialect/TTNN/Transforms/Workarounds/TTNNWorkaroundsPatterns.cpp
```diff
@@ -411,7 +431,7 @@ class TTNNAllReduceWorkarounds : public OpRewritePattern<ttnn::AllReduceOp> {
       // blowup, enforce a size constraint based on DRAM capacity.
       if (exceedsAllGatherReduceMemLimit(ttcore::getCurrentScopeSystemDesc(op),
                                          inputType, meshShape[clusterAxis],
-                                         0.05)) {
+                                         0.12)) {
```
btsoiTT, line 431: I need to increase the limit for the tt-xla gpt-oss-20b generative example to work. 

I am actually not sure if this MemLimit check is even necessary? There is no additional fallback in the original code, so if `exceedsAllGatherReduceMemLimit` returns false (note that the way the function is written, returning true means we are not within the limit and cannot proceed with the AllGather LocalReduce workaround), the program keeps going and gets undefined behavior or segfault somewhere. So it might be better to just remove the speculative memory limit check, and if we don't have enough memory we just OOM and crash anyways.

mvasiljevicTT:
@btsoiTT looks like we got into similar problem with AR workaround, I was working on the fix on [this PR](https://github.com/tenstorrent/tt-mlir/pull/7152) but paused it as it didn't seem to cover real cases. Now that it does we can consider merging it. 

I agree with your comment about `rewriteAsAllGatherLocalReduce` fallback, in my PR I removed `exceedsAllGatherReduceMemLimit` check and always do local reduce if none of the dims are valid.

There is some difference in a way we check divisibility, for tiled input I check only number of tiles and not number of elements, and if divisible pad the input to full num_dev*k tiles, so that each output is k tiles and all pads and slices are added in mlir and don't rely on shape matching in metal.

Could you please take a look at my PR and see if we can combine the efforts?

btsoiTT:
Thanks, I will take a look.

mvasiljevicTT:
@btsoiTT Did you get the chance to check if my PR solves your problem too? I am ready to merge it now.

btsoiTT:
Sorry for the delayed response, I was running it with the gpt oss 20b example and with tt-xla CI. Yes your PR looks good and works, you can go ahead and merge it, I will close this PR.

PR #7417, opened 2026-03-09
[D2M] DMA Optimizations

vtangTT:
### Problem description
See [proposal](https://docs.google.com/document/d/1rjyBRWJQMRTXdPn_JHNKdkNIsT51dkdTQh9M7OvaHuM/edit?usp=sharing) for optimizations IR snippets

### What's changed
Implemented 2/3 optimizations in the doc:
- coalescing read/write barriers (`sinkAllBarriers`)
- defer write barriers (`deferWriteBarriers`)
- canonicalization for `ttkernel.noc_async_read/write_barrier`

Refactor to add `D2M_DMAType` to `MemTx`
- by doing so, can completely delete `AssociatedDMAWaits`

### Checklist
- [x] passing nightly [link](https://github.com/tenstorrent/tt-mlir/actions/runs/23950504246)

### DIFF
--- a/test/ttmlir/Dialect/D2M/Transforms/dma_optimizations.mlir
+++ b/test/ttmlir/Dialect/D2M/Transforms/dma_optimizations.mlir
@@ -15,8 +15,8 @@ module attributes {} {
     %alloc0 = memref.alloc() {address = 1024 : i64, alignment = 16 : i64} : memref<2x4x2x4x!ttcore.tile<32x32, f32>, #ttcore.shard<16384x4096, 1>, #l1>
     %alloc1 = memref.alloc() {address = 5120 : i64, alignment = 16 : i64} : memref<2x4x2x4x!ttcore.tile<32x32, f32>, #ttcore.shard<16384x4096, 1>, #l1>
     %alloc_out = memref.alloc() {address = 9216 : i64, alignment = 16 : i64} : memref<2x4x2x4x!ttcore.tile<32x32, f32>, #ttcore.shard<16384x4096, 1>, #l1>
-    %stream0 = "d2m.stream_layout"(%arg0, %alloc0) <{remapping = #map4}> : (memref<2x4x2x4x!ttcore.tile<32x32, f32>, #ttcore.shard<16384x4096, 1>, #dram>, memref<2x4x2x4x!ttcore.tile<32x32, f32>, #ttcore.shard<16384x4096, 1>, #l1>) -> memref<2x4x2x4x!ttcore.tile<32x32, f32>, #ttcore.view<4>, #dram>
-    %stream1 = "d2m.stream_layout"(%arg1, %alloc1) <{remapping = #map4}> : (memref<2x4x2x4x!ttcore.tile<32x32, f32>, #ttcore.shard<16384x4096, 1>, #dram>, memref<2x4x2x4x!ttcore.tile<32x32, f32>, #ttcore.shard<16384x4096, 1>, #l1>) -> memref<2x4x2x4x!ttcore.tile<32x32, f32>, #ttcore.view<4>, #dram>
+    %stream0 = d2m.view_layout %arg0 remapping = #map4 : memref<2x4x2x4x!ttcore.tile<32x32, f32>, #ttcore.shard<16384x4096, 1>, #dram> -> memref<2x4x2x4x!ttcore.tile<32x32, f32>, #ttcore.view<4>, #dram>
+    %stream1 = d2m.view_layout %arg1 remapping = #map4 : memref<2x4x2x4x!ttcore.tile<32x32, f32>, #ttcore.shard<16384x4096, 1>, #dram> -> memref<2x4x2x4x!ttcore.tile<32x32, f32>, #ttcore.view<4>, #dram>
     d2m.generic {block_factors = [], grid = #ttcore.grid<2x4>, indexing_maps = [], iterator_types = [], threads = [#d2m.thread<datamovement>, #d2m.thread<compute>]}
         ins(%stream0, %stream1 : memref<2x4x2x4x!ttcore.tile<32x32, f32>, #ttcore.view<4>, #dram>, memref<2x4x2x4x!ttcore.tile<32x32, f32>, #ttcore.view<4>, #dram>)
         outs(%alloc_out : memref<2x4x2x4x!ttcore.tile<32x32, f32>, #ttcore.shard<16384x4096, 1>, #l1>)  {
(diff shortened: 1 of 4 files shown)


lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
```diff
@@ -1632,34 +1630,20 @@ class D2MDMAReadRewriter : public OpConversionPattern<d2m::DMAReadOp> {
     rewriter.create<ttkernel::NocAsyncReadOp>(op.getLoc(), srcNocAddr,
                                               dstL1Addr, size);
 
-    // Add attribute marking whether the DMA wait is for a read or write
-    // operation This will be used when loweing the wait ops because the current
-    // DMA op will be replaced with a NullTx.
-    auto dmaWaitOps = associatedDMAWaits->get(op);
```
bgrady-tt, line 1879: fwiw I like using enums here to clean this up 👍

test/ttmlir/Dialect/D2M/Transforms/dma_optimizations.mlir
```diff
@@ -0,0 +1,434 @@
...
+    %stream_1 = "d2m.stream_layout"(%arg1, %alloc_0) <{remapping = #map4}> : (memref<2x4x2x4x!ttcore.tile<32x32, f32>, #ttcore.shard<16384x4096, 1>, #dram>, memref<2x4x2x4x!ttcore.tile<32x32, f32>, #ttcore.shard<16384x4096, 1>, #l1>) -> memref<2x4x2x4x!ttcore.tile<32x32, f32>, #ttcore.view<4>, #dram>
+    d2m.generic {block_factors = [], grid = #ttcore.grid<2x4>, indexing_maps = [], iterator_types = [], threads = [#d2m.thread<datamovement>, #d2m.thread<compute>]}
+        ins(%stream : memref<2x4x2x4x!ttcore.tile<32x32, f32>, #ttcore.view<4>, #dram>)
+        outs(%stream_1 : memref<2x4x2x4x!ttcore.tile<32x32, f32>, #ttcore.view<4>, #dram>)  {
+    ^datamovement0(%cb0: !d2m.cb<memref<2x4x!ttcore.tile<32x32, f32>, #l1>>, %cb1: !d2m.cb<memref<2x4x!ttcore.tile<32x32, f32>, #l1>>):
+      %gi = d2m.core_index(0) : index
+      %gj = d2m.core_index(1) : index
+      // CHECK: d2m.reserve %cb0
```
bgrady-tt: Still reviewing, but this seems incorrect to me? Would this allow the following rewrite?

```cpp
// before optimization
read
read_barrier
compute // uses read!
write
write_barrier

// after optimization
read
compute // read is not pushed!
write
read_barrier
write_barrier
```

lib/Dialect/D2M/Transforms/DMAOptimizations.cpp
```diff
@@ -0,0 +1,452 @@
...
+  if (auto dmaWrite = mlir::dyn_cast<DMAWriteOp>(op)) {
+    return getCBFromLocalBuffer(dmaWrite.getSrc());
+  }
+  if (auto dmaWaitOp = mlir::dyn_cast<DMAWaitOp>(op)) {
+    return getCBForOp(dmaWaitOp.getMemTx().getDefiningOp());
+  }
+  return nullptr;
+}
+
+static bool areDifferentCBs(Value a, Value b) { return a && b && a != b; }
```
bgrady-tt: nit: I would expect this to check that `Value` a and b are actually CBs. Maybe rename to `areDifferentValues`?

lib/Dialect/D2M/Transforms/DMAOptimizations.cpp
```diff
@@ -0,0 +1,452 @@
...
+// Returns true if a barrier was deferred (forOp is updated to the new ForOp).
+// Returns false if no deferrable write barrier was found.
+//
+// Assumes the loop executes at least once (lb < ub). For a
+// zero-iteration loop, the epilogue would incorrectly pop from an unfilled CB.
+//===----------------------------------------------------------------------===//
+static bool deferOneWriteBarrier(scf::ForOp &forOp) {
+  if (auto tripCount = forOp.getStaticTripCount()) {
+    TT_assertv(tripCount->isStrictlyPositive(),
+               "write barrier deferral requires loop to execute at least once");
```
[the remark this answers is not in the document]
bgrady-tt: I think we can assume trip count > 1 here.

lib/Dialect/D2M/Transforms/DMAOptimizations.cpp
```diff
@@ -0,0 +1,452 @@
...
+    return areDifferentCBs(getCBForOp(sinkOver), sinkCB);
+  }
+  if (isReadBarrier(dmaWait)) {
+    if (auto pushOp = mlir::dyn_cast<PushOp>(sinkOver)) {
+      return areDifferentCBs(pushOp.getCb(), sinkCB);
+    }
+    if (mlir::isa<SemaphoreSetOp, SemaphoreWaitOp, SemaphoreIncOp>(sinkOver)) {
+      return false;
+    }
+  } else if (isWriteBarrier(dmaWait)) {
```
bgrady-tt, line 110: I don't think it should be possible to sink a read barrier _past_ a write barrier. In general we are assuming here that the overall baby RISC pipeline is basically reader->compute->writer.
```
for it in blocking_iters:
  read_all_inputs_to_cbs
  read_barrier_and_push
  // implicitly there is a 'compute' barrier here
  // that consumes from all input CBs and produce into all output CBs
  write_all_outputs_to_cbs
  write_barrier_and_pop
```

All read barriers and their associated CB pushes should probably be assumed to be dependencies for any write_barrier to execute. There may be exceptions where this is ok, but it seems hazardous and probably not worth any perf.

test/ttmlir/Dialect/D2M/Transforms/dma_optimizations.mlir
vtangTT: Oh I see, ya this would cause problems. 100% agreed with all your points that we should not be sinking read barriers past any writes under any circumstances.

lib/Dialect/D2M/Transforms/DMAOptimizations.cpp
vtangTT: fixed

lib/Dialect/D2M/Transforms/DMAOptimizations.cpp
vtangTT: Agreed. Changed so that reads will never sink past writes

vtangTT:
as per offline discussion, will hold off on merging until state of DMA is more stable on main

Issue #5124, opened 2025-09-28
Implement `TTCore_Tile::getFloatSemantics`


nsmithtt:
The float type interface was introduced in order to support using the ttcore tile type in arith and math dialects, however, we haven't thus far needed any upstream passes that call this interface method hence we left it unimplemented.

PR #7431, opened 2026-03-10
[d2m] element normalization

### LINKED ISSUE #6650: [d2m] Run element type normalization in pass pipeline
In order to support ops with i64 and i1 we should add ElementTypeNormalization (ttir-element-type-normalization) to our pass pipeline before ttir conversion.

e.g.
```mlir
  func.func @component_1_ttir_gt_6(%arg0: tensor<1x1x1x128xi64>, %arg1: tensor<1x1x1x1xi64>) -> tensor<1x1x1x128xi1> {
    %0 = "ttir.gt"(%arg0, %arg1) : (tensor<1x1x1x128xi64>, tensor<1x1x1x1xi64>) -> tensor<1x1x1x128xi1>
    return %0 : tensor<1x1x1x128xi1>
  }
```

There is however a bug with the ElementTypeNormalization pass which trivially replaces types, in this case it replaces `i64` -> `si32` and `i1` -> `bf16` resulting in:
```mlir
  func.func @component_1_ttir_gt_6(%arg0: tensor<1x1x1x128xsi32>, %arg1: tensor<1x1x1x1xsi32>) -> tensor<1x1x1x128xbf16> {
    %0 = "ttir.gt"(%arg0, %arg1) : (tensor<1x1x1x128xsi32>, tensor<1x1x1x1xsi32>) -> tensor<1x1x1x128xbf16>
    return %0 : tensor<1x1x1x128xbf16>
  }
```

This is obviously incorrect and leads to other issues down the line complaining about mismatched input and output element types.

dlokeTT:
### Ticket
Closes #6650 

### Problem description
We don't support f64/i64/i1 tensors on device. Normal `element-type-normalization` pass trivially replaces `i1` tensors with `bf16` but this can cause mismatches with input/output dtypes in d2m path without forcing typecasts. 

### What's changed
Added a new `predicateTypeAlignment` pass that looks at ops such as comparisons or where, and when the condition or result are `i1` tensors, rewrite these to match the dtype of the inputs or other inputs.
Added lit tests for this new predicate alignment pass and builder tests for `f64/i64/i1` tensors
Added this pass and element type normalization to the ttmetal pipeline

### Checklist

### DIFF
--- a/lib/Dialect/TTMetal/Pipelines/TTMetalPipelines.cpp
+++ b/lib/Dialect/TTMetal/Pipelines/TTMetalPipelines.cpp
@@ -83,14 +83,14 @@ void createTTIRToTTMetalFrontendPipeline(
     registerDeviceOptions.meshShape = llvm::to_vector(options.meshShape);
   }
   pm.addPass(ttcore::createTTCoreRegisterDevicePass(registerDeviceOptions));
+  pm.addPass(ttir::createPredicateTypeAlignment());
+  pm.addPass(ttir::createElementTypeNormalization(
+      ttir::ElementTypeNormalizationOptions()));
   pm.addPass(tt::createTTIRToTTIRDecompositionPass());
   pm.addPass(ttir::createTTIRExplicateTMs());
   pm.addPass(ttir::createTTIREraseInverseOps());
   pm.addPass(ttir::createTTIRMoveReshapeToConstant());
   pm.addPass(ttir::createTTIRFoldConstantReshapeBroadcast());
-  pm.addPass(ttir::createPredicateTypeAlignment());
-  pm.addPass(ttir::createElementTypeNormalization(
-      ttir::ElementTypeNormalizationOptions()));
   pm.addPass(ttir::createTTIRReductionForceKeepDim());
   pm.addPass(ttir::createTTIRRankNormalization());
   pm.addPass(ttir::createTTIRDecomposeComplexReshape());


lib/Dialect/TTIR/Transforms/ElementTypeNormalizationForTTMetal.cpp
```diff
@@ -0,0 +1,544 @@
+// SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
```
vwellsTT: nit: year

lib/Dialect/TTIR/Transforms/SmartElementTypeNormalization.cpp
```diff
@@ -0,0 +1,544 @@
...
+            if (floatType.getWidth() == 64) {
+              Type newElementType = mlir::Float32Type::get(type.getContext());
+              return mlir::RankedTensorType::get(
+                  type.getShape(), newElementType, type.getEncoding());
+            }
+          }
+
+          return type;
+        });
+    // Pass-through for non-tensor types (e.g. memref in function signatures)
```
vwellsTT: why would you encounter a memref? it seems like that shouldn't happen based on pass ordering?

include/ttmlir/Dialect/TTIR/Transforms/Passes.td
```diff
@@ -114,6 +114,40 @@ def ElementTypeNormalization: Pass<"ttir-element-type-normalization", "::mlir::M
   ];
 }
 
+def ElementTypeNormalizationForTTMetal
```
vwellsTT: yeah, not a fan of this name.  I think something like "ttir-normalize-element-types" would be more in keeping with typical pattern (though the exact normalization pattern is a bit unclear still, but I think that's fine).

include/ttmlir/Dialect/TTIR/Transforms/Passes.td
```diff
@@ -114,6 +114,40 @@ def ElementTypeNormalization: Pass<"ttir-element-type-normalization", "::mlir::M
...
+    steps. TTMetal does not support 64-bit or native i1 tensors; comparisons
+    and logical ops are expected to produce the same element type as their
+    operands (e.g. i32) rather than i1.
+
+    Step 1 — Downsize 64-bit to 32-bit:
+      Converts all tensor element types from i64 to i32 and f64 to f32 (including
+      constants). Tensor-of-tile and quantized types are left unchanged.
+
+    Step 2 — Align predicate-like results with data types:
+      So that no op is left producing i1 while others expect a data type:
```
vwellsTT: nit: this sort of nested colon pattern is awkward imo, maybe this sentence can be removed?

lib/Dialect/TTIR/Transforms/SmartElementTypeNormalization.cpp
```diff
@@ -0,0 +1,544 @@
...
+
+namespace {
+class I64ToI32AndF64ToF32TypeConverter : public TypeConverter {
+public:
+  I64ToI32AndF64ToF32TypeConverter() {
+    addConversion(
+        [](mlir::RankedTensorType type) -> std::optional<RankedTensorType> {
+          Type elementType = type.getElementType();
+
+          // Tensor-of-tile types use TTCore TileType as element type; do not
```
vwellsTT: can we have tile of f64 or the like then? Or does that never happen? Kind of confused why this is an exception

dlokeTT: It never happens because we only hit this on ttnn jit side and ttnn doesn't support that dtype. This is an exception so we don't run the type convertor on it because we didn't implement `getFloatSemantics`. There's an issue for this https://github.com/tenstorrent/tt-mlir/issues/5124 which would remove this check but that implementation seemed like it should be in a different pr that removes a lot of these checks all over the codebase. I'll put a comment saying this can be removed when that issue is implemented

include/ttmlir/Dialect/TTIR/Transforms/Passes.td
dlokeTT: changed it to smart element type normalization since its not trivially rewriting types like the ttnn version. could use a different word instead of smart or something to explain that this does non trivial work but couldn't think of anything else

lib/Dialect/TTIR/Transforms/SmartElementTypeNormalization.cpp
dlokeTT: The fabric tests in `test/python/golden/test_fabric_apis.py` use snippets that operate on the ttkernel level and are already bufferized and run the full pipeline

Maybe they shouldn't be running the full pipeline, let me talk to Sohaib and see about this but originally just put this in to fix CI

dlokeTT: going to remove this check since the full pipeline won't be running on the fabric tests after https://github.com/tenstorrent/tt-mlir/pull/7542

sdjordjevicTT:
Hi @dlokeTT can you elaborate more on what the difference is between this new smart element type normalization and the element type normalization we currently have? I am more in favour of adding additional options and extending the current pass that can handle both backends, than having a new smart one added to the codebase...

dlokeTT:
> Hi @dlokeTT can you elaborate more on what the difference is between this new smart element type normalization and the element type normalization we currently have? I am more in favour of adding additional options and extending the current pass that can handle both backends, than having a new smart one added to the codebase...

Hi, yes. I think the main difference in the passes is that we don't trivially replace `i1` tensors with `bf16`. For a certain subset of ops such as comparisons, the output tensor is always `i1` because that is what `stablehlo.compare` returns.  In ttnn path you guys can just promote that to bf16 because the ttnn op supports different output dtypes than input as it just does a typecast after the comparison.

For ttmetal path, if we trivially promote it to bf16 that will cause issues in lowering since the input/output dtypes don't match and there is no explicit typecast. We could add this typecast in the lowering as well but it seems forced for no reason since the dtype will likely be changed to something else down the line.

Originally, I was thinking that this pass could be used in the ttnn path as well since it should result in less typecasts but decided for now to just have it in the ttmetal pipeline. I can look into integrating this with the original pass and then enabling options for the different pipelines as well if that's the simplest way forward.

vwellsTT:
I still have same opinion as @sdjordjevicTT FWIW, I think we should try to avoid introducing redundancy vs adding it with a "we'll clean this up later" mindset.  It sounds like the differences between this pass and existing pass are fairly small and should be easy to make configurable imo

dlokeTT:
> I still have same opinion as @sdjordjevicTT FWIW, I think we should try to avoid introducing redundancy vs adding it with a "we'll clean this up later" mindset. It sounds like the differences between this pass and existing pass are fairly small and should be easy to make configurable imo

Updated this pr. We now run a pass before element type normalization that does the predicate alignment to match input/output dtypes for specific ops and then run the regular ttir element normalization pass. This should address concerns on redundant code and make both backends aligned. My thinking is also that this pass could act as an optimization for ttnn backend since it removes typecasts but that can be explored later.

lib/Dialect/TTIR/Transforms/ElementTypeNormalization.cpp
```diff
@@ -27,6 +27,12 @@ class ElementTypeConverter : public TypeConverter {
             return type;
           }
 
+          // Preserve tensor<!ttcore.tile<…>>: lowering tile to scalar f32
+          // breaks L1 sharded layouts and runtime CB sizing
```
vwellsTT: nit: end with .

lib/Dialect/TTIR/Transforms/PredicateTypeAlignment.cpp
```diff
@@ -0,0 +1,325 @@
...
+    state.addTypes(newResultType);
+    mlir::Operation *newOp = rewriter.create(state);
+    rewriter.replaceOp(op, newOp->getResults());
+    return mlir::success();
+  }
+};
+
+// Use signless i32 for 32-bit integer compare/logical results so later passes
+// see a single i32 kind.
+static Type normalizeI32Signless(Type elementType, MLIRContext *ctx) {
```
vwellsTT: not exactly following why this is needed

lib/Dialect/TTMetal/Pipelines/TTMetalPipelines.cpp
```diff
@@ -88,6 +88,11 @@ void createTTIRToTTMetalFrontendPipeline(
   pm.addPass(ttir::createTTIREraseInverseOps());
   pm.addPass(ttir::createTTIRMoveReshapeToConstant());
   pm.addPass(ttir::createTTIRFoldConstantReshapeBroadcast());
+  pm.addPass(ttir::createPredicateTypeAlignment());
+  {
+    ttir::ElementTypeNormalizationOptions elementTypeOpts;
+    pm.addPass(ttir::createElementTypeNormalization(elementTypeOpts));
```
vwellsTT: nit: this is not the regular structure for these passes; elsewhere we only use the {} around setting fields in the options.  See `  pm.addPass(ttcore::createTTCoreRegisterDevicePass(registerDeviceOptions));` bit above.

lib/Dialect/TTIR/Transforms/PredicateTypeAlignment.cpp
dlokeTT: remnant of old code, I forget why I was using it now and things pass so removed

lib/Dialect/TTMetal/Pipelines/TTMetalPipelines.cpp
```diff
@@ -88,6 +88,9 @@ void createTTIRToTTMetalFrontendPipeline(
   pm.addPass(ttir::createTTIREraseInverseOps());
   pm.addPass(ttir::createTTIRMoveReshapeToConstant());
   pm.addPass(ttir::createTTIRFoldConstantReshapeBroadcast());
+  pm.addPass(ttir::createPredicateTypeAlignment());
+  pm.addPass(ttir::createElementTypeNormalization(
```
sdjordjevicTT: Any reasons why this pass isn't the first pass in the TT-Metal pipeline? Should we consider normalizing types before anything else in the pipeline?

dlokeTT: not really, yeah makes sense to do it first. I moved it to the front now

Issue #7433, opened 2026-03-10
CSE pass incorrectly merges distinct d2m.empty() ops when MemoryEffects is omitted from D2M_EmptyOp


hkwonTT:
### Describe the bug

**Summary**
For functions that return multiple outputs, the compiler correctly generates one d2m.empty() per output (each with the same type as its output). After the removal of `MemoryEffects<[MemAlloc]>` from `D2M_EmptyOp` (e.g. in PR #7144 ), the CSE pass treats these allocations as a common subexpression and replaces them with a single `d2m.empty()`. That leaves only one buffer where the program semantics require two, and can lead to runtime failures (e.g. "Buffer is not allocated" when handling the second output).

**Root cause**
 - CSE considers two ops equivalent if they have the same type, same operands, and no side effects.
Without `MemoryEffects<[MemAlloc]>`, `d2m.empty()` is treated as side-effect-free.
 - Two `d2m.empty()` ops with the same result type and no operands are then folded into one; all uses of the second are replaced by the first.

### Steps to reproduce the issue

Minimal repro: a function that returns two outputs (e.g. tanh and exp of the same input). After the D2M pipeline (including CSE), there should be two allocations, but only one remains.

Example:
```
#dram = #ttnn.buffer_type<dram>
#ttnn_layout = #ttnn.ttnn_layout<(d0, d1) -> (d0, d1), <1x1>, memref<2x4x!ttcore.tile<32x32, f32>, #dram>, <interleaved>, exactGrid = true>
module {
  func.func @dual_ops(%arg0: tensor<64x128xf32, #ttnn_layout>) -> (tensor<64x128xf32, #ttnn_layout>, tensor<64x128xf32, #ttnn_layout>) {
    %1 = "ttir.tanh"(%arg0) : (tensor<64x128xf32, #ttnn_layout>) -> tensor<64x128xf32, #ttnn_layout>
    %2 = "ttir.exp"(%arg0) : (tensor<64x128xf32, #ttnn_layout>) -> tensor<64x128xf32, #ttnn_layout>
    return %1, %2 : tensor<64x128xf32, #ttnn_layout>, tensor<64x128xf32, #ttnn_layout>
  }
}
```
Commands:
`ttmlir-opt --ttir-to-ttmetal-pipeline="ttnn-mode=true system-desc-path=$SYSTEM_DESC_PATH" single_tanh_and_exp.mlir -o _single_tanh_and_exp_ttnn.mlir`
[...]

PR #7438, opened 2026-03-10
[D2M] Change d2m.emptyOp Traits to Avoid Incorrect Elimination by CSE Pass

arminaleTT:
### Ticket
Closes https://github.com/tenstorrent/tt-mlir/issues/7433

### Problem description
In #7144 the d2m.emptyOp was changed from having a `MemoryEffects<[MemAlloc]>` trait to having a `Pure` trait. This fixed the issue of the emptyOp not being canonicalized when it was unused. However, this also led the CSE pass to combine d2m.emptyOps when their types were identical, turning two tensor allocations into one. This is incorrect.

### What's changed
- Changed the trait on d2m.emptyOp from `Pure` to `MemoryEffectsOpInterface`. 
  - The previous `MemoryEffects<[MemAlloc]>` declares a side effect that is not associated with any value; thus it cannot be eliminated. With this implementation of `getEffects`, we tie the side effect to the result of the empty op
- Add lit tests to verify both canonicalization and CSE behaviour

### Checklist

### DIFF
--- a/include/ttmlir/Dialect/D2M/IR/D2MOps.td
+++ b/include/ttmlir/Dialect/D2M/IR/D2MOps.td
@@ -389,7 +389,8 @@ def D2M_GenericOp
 
 def D2M_EmptyOp
     : D2M_Op<
-          "empty", [Pure, DeclareOpInterfaceMethods<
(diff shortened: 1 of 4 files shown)


bgrady-tt (approved):
LGTM 👍

PR #7443, opened 2026-03-11
Set system sfpi to off by default, except for ci and wheels

nsumrakTT:
### Ticket
slack

### Problem description
Set system sfpi to off by default for dev environment.

### What's changed
For CI/wheels make sure system sfpi is set to ON.

### Checklist
- wheel build run: https://github.com/tenstorrent/tt-mlir/actions/runs/22960197324
(manually checked sfpi is system)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -45,7 +45,7 @@ option(TTMLIR_ENABLE_TOOLS "Enable building core tools (ttmlir-opt, ttmlir-lsp-s
 option(TTMLIR_ENABLE_ALCHEMIST "Enable tt-alchemist lib" ON)
 option(TTMLIR_ENABLE_ALCHEMIST_WHEEL "Enable tt-alchemist wheel" OFF)  # Expensive to build
 option(CODE_COVERAGE "Enable coverage reporting" OFF)
-option(TT_USE_SYSTEM_SFPI "Use system path for SFPI. SFPI is used to compile firmware." ON)
+option(TT_USE_SYSTEM_SFPI "Use system path for SFPI. SFPI is used to compile firmware." OFF)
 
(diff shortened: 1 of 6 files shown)


CMakeLists.txt
```diff
@@ -45,7 +45,7 @@ option(TTMLIR_ENABLE_TOOLS "Enable building core tools (ttmlir-opt, ttmlir-lsp-s
 option(TTMLIR_ENABLE_ALCHEMIST "Enable tt-alchemist lib" ON)
 option(TTMLIR_ENABLE_ALCHEMIST_WHEEL "Enable tt-alchemist wheel" OFF)  # Expensive to build
 option(CODE_COVERAGE "Enable coverage reporting" OFF)
-option(TT_USE_SYSTEM_SFPI "Use system path for SFPI. SFPI is used to compile firmware." ON)
+option(TT_USE_SYSTEM_SFPI "Use system path for SFPI. SFPI is used to compile firmware." OFF)
```
nsmithtt, line 48: Wait, why do we want this?  If folks are using the docker for development this should be ON

nsumrakTT: A few of folks are reporting that they have to install dependancies. This could be if metal uplifts sfpi and you are using new docker on a older branch. So, this PR makes local dev always pull local sfpi while CI and wheels uses system sfpi which will always be inline with docker and installed sfpi version.

nsmithtt: But we explicitly install it in the base docker here https://github.com/tenstorrent/tt-mlir/blob/4e0480eeb72e0221a6c57e9eb86f687f519e4aab/.github/Dockerfile.base#L49.  I've never had to manually install sfpi, so I'd like to better understand why folks need to do this.

vtangTT (approved):

PR #7448, opened 2026-03-11
Fix MorehCumSumOp missing memory config in flatbuffer translation

rpavlovicTT:
MorehCumSumOp was falling back to no memory config (0) when the op lacked an explicit memory_config attribute, instead of deriving it from the result tensor type. Use getMemoryConfigIfNeeded() like other ops.


svuckovicTT (approved):
Clean change, very well written, would approve twice if I could! 💯

PR #7452, opened 2026-03-12
[EmitPy] Decouple consteval wrapper creation from file splitting

amilovanovicTT:
### Problem description

Previously, only the split-files path in EmitPy codegen created a separate `consteval_forward` wrapper function per forward function, containing the complete consteval logic of that forward function. In the non-split-files path, the forward function itself handled the consteval and caching logic inline. This coupling between wrapper creation and file splitting made the `TTNNFileSplit` pass more complex than necessary and harder to extend.

### What's changed

Moved the logic that creates the consteval function wrapper to the `TTNNPrepareConstEvalCaching` pass. This enables consteval function wrappers to be generated regardless of file splitting, making the behavior consistent across both paths.

Key changes:
- **`TTNNPrepareConstEvalCaching`**  : Creates `consteval_forward` wrapper functions, caching dictionaries per forward function, and replaces `ttcore.LoadCachedOp`s with dictionary lookups.
- **`TTNNFileSplit`** : Simplified pass that only organizes operations into `FileOp` containers (`main` / `consteval`).
- **`EmitPyConstEvalCaching`** : Post-conversion pass that wraps const-eval calls in `if not dict:` guards and sets `emitpy.name` attributes for proper variable naming.

As a result, the `TTNNFileSplit` pass becomes much more concise — it only moves operations between file containers. The future plan is to introduce additional file splits (e.g., separating input generators) to keep the main file focused on model logic, and this decoupling makes the split-files pass more reusable.

### Checklist


lib/Conversion/TTNNToEmitPy/TTNNToEmitPy.cpp
```diff
@@ -2646,7 +2646,14 @@ class TupleOpConversionPattern
 } // namespace
 
 // LoadCachedOp conversion pattern
+// LoadCachedOp conversion pattern
```
svuckovicTT: duplicate line

lib/Conversion/TTNNToEmitPy/TTNNToEmitPy.cpp
```diff
@@ -2646,7 +2646,14 @@ class TupleOpConversionPattern
...
+//
+// Converts ttcore.load_cached to a plain function call. The caching
+// infrastructure (globals, if-guards) is handled by the
+// TTNNPrepareConstEvalCaching pass (pre-split) and the
+// EmitPyConstEvalCaching pass (post-conversion).
 //
+
 // Converts ttcore.load_cached to a plain function call. The caching
```
svuckovicTT, line 2650: Forgot to remove old comment?

lib/Dialect/TTNN/Transforms/TTNNPrepareConstEvalCaching.cpp
```diff
@@ -17,17 +19,52 @@ namespace mlir::tt::ttnn {
...
-// This pass prepares the const-eval results for caching by packing the results
-// of all LoadCachedOps into a single global caching dictionary per forward
-// function. For each forward function containing LoadCachedOps, it creates a
-// global dictionary (e.g., `_cached_forward`), retrieves it at the top of the
-// function body, and stores each LoadCachedOp's results under its callee name.
+// This pass prepares const-eval results for caching by consolidating all
+// LoadCachedOps within each forward function into a single global caching
+// dictionary. For each forward function containing LoadCachedOps, it:
+//
+//   1. Creates a global caching dictionary (e.g., `_cached_forward`).
```
svuckovicTT: Prefer using "cache [dictionary]" instead of "caching [dictionary]". The difference is subtle but meaningful:

```
Grammatical Difference

Caching dictionary: "Caching" acts as a present participle adjective describing the function of the dictionary (a dictionary that performs the action of caching).

Cache dictionary: "Cache" acts as a noun modifier (attributive noun) describing the purpose or type of the dictionary (a dictionary used for the purpose of caching). 

Merriam-Webster Dictionary
```

svuckovicTT (approved):

lib/Conversion/TTNNToEmitPy/TTNNToEmitPy.cpp
amilovanovicTT: Removed

lib/Conversion/TTNNToEmitPy/TTNNToEmitPy.cpp
amilovanovicTT: Yes, removed

lib/Dialect/TTNN/Transforms/TTNNPrepareConstEvalCaching.cpp
amilovanovicTT: Agreed, changed.

lib/Dialect/TTNN/Transforms/TTNNPrepareConstEvalCaching.cpp
```diff
@@ -59,30 +97,120 @@ class TTNNPrepareConstEvalCaching
...
+      }
+    }
+
+    // For each LoadCachedOp, store its results under one key in the caching
+    // dictionary. The key is the callee name of the LoadCachedOp.
+    for (auto loadCachedOp : loadCachedOps) {
+      auto *clonedLoadCachedOp = mapping.lookup(loadCachedOp);
+      builder.setInsertionPointAfter(clonedLoadCachedOp);
+      builder.create<ttcore::SetKeyValueOp>(
+          clonedLoadCachedOp->getLoc(), mapping.lookup(cacheDict),
```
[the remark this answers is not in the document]
amilovanovicTT: This is incorrect, since `IRMapping` stores both `DenseMap<Value, Value>` and `DenseMap<Operation*, Operation*>`.

Issue #6705, opened 2026-01-22
Remove matmul program config generation


rpavlovicTT:
https://github.com/tenstorrent/tt-mlir/blob/e542457c96c56c1cca55b7e8acda2fc1627238f8/lib/Dialect/TTNN/Analysis/MatmulProgramConfig.cpp#L154

This issue is to track the idea of removing generation of matmul program configs in tt-mlir compiler. Matmul op in tt-metal has its own logic of creating configs and we should rely on it. Problem for tt-mlir compiler is that doesn't work through NO_DISPATCH mode which we use for evaluating ops and memory layout decisions. Once we have NO_DISPATCH mode working with memory allocator state this should be possible to fix in tt-mlir.

PR #7454, opened 2026-03-12
[Greedy Optimizer] Layout propagation

rpavlovicTT:
Introduce a new greedy/beam-search layout propagation pass that decides
optimal tensor layouts (L1 sharded vs DRAM interleaved) for each op in
the TTNN graph.

The pass works in two phases:

Forward pass: walks ops in topological order, building per-operand input
candidate sets from producer beam states and reshard exploration. For each
op, evaluates the cross-product of input candidates x output layout hints
through OpModel constraint validation. Valid candidates are scored using
a multi-criteria function (L1 > DRAM, sharded > interleaved, fewer DRAM
input bytes, no reshard > reshard, more cores, lower L1 usage) and the
top-K are kept per op.

Backward pass (K > 1): consolidates beam at fork points using a greedy
strategy that maximizes reshard-free consumer edges, then patches
reshardLayouts for mismatched consumers.

IR transformation: applies chosen configs to op result types, inserts
ToMemoryConfigOp for reshards, updates function return types, handles
op-specific attributes (conv2d config, matmul program config), and
inserts DRAM spills before func.return for L1 outputs.

New components:
- LayoutPropagation (Analysis): core beam search, cross-product eval,
  input candidate generation with reshard fan-out, IR transformation
- OpModelStrategy (Analysis): LayoutScore scoring, per-op OutputHints
  dispatch (TypeSwitch), shouldExploreReshards policy, BeamCandidate
- LayoutPropagationObserver: NullObject observer interface for
  pluggable tracing/profiling without conditionals in hot paths
- GreedyLayoutPropagation (Transforms): MLIR pass wrapper that runs
  ScalarDataType, LegalTensorLayout, LegalOpLayout, and LegalOpConfig
  analyses before invoking LayoutPropagation per function

Infrastructure changes to existing files:
- OpConstraintValidation: extract checkConstraintsResult() for reuse,
  add template validateOperation<OpType>() for op-less validation,
  add cbPeakUsage field to ValidationResult
- Utils: add getPerCoreL1Usage() for per-core L1 memory accounting
- Logger: add GreedyOptimizer and ValidationFallback log components
- OperationValidationAndFallback: rename log component to
  ValidationFallback for clearer log filtering

Input layout filters:
- ConcatOp, ConcatenateHeadsOp, SliceStaticOp, SliceDynamicOp: reject
  all sharded inputs
- ReshapeOp, PermuteOp: reject width-sharded inputs

Unit tests:
- TestOpModelStrategy: scoring comparison, output hints dispatch
- TestBeamSearch: end-to-end beam search on synthetic IR
- TestReshardCandidates: reshard candidate generation and dedup

### DIFF
--- a/lib/Dialect/TTNN/Analysis/MemoryLayoutPropagation.cpp
+++ b/lib/Dialect/TTNN/Analysis/MemoryLayoutPropagation.cpp
@@ -577,11 +577,10 @@ MemoryLayoutPropagation::processOp(Operation *op) {
   return candidates;
 }
 
-bool MemoryLayoutPropagation::validateReshard(Operation *consumerOp,
-                                        Operation *producerOp,
-                                        TTNNLayoutAttr producerOutputLayout,
-                                        TTNNLayoutAttr reshardLayout,
-                                        size_t producerResultIdx) {
+bool MemoryLayoutPropagation::validateReshard(
+    Operation *consumerOp, Operation *producerOp,
+    TTNNLayoutAttr producerOutputLayout, TTNNLayoutAttr reshardLayout,
+    size_t producerResultIdx) {
   auto inputShape = mlir::cast<RankedTensorType>(
                         producerOp->getResult(producerResultIdx).getType())
                         .getShape();
@@ -865,9 +864,8 @@ MemoryLayoutPropagation::getInputCandidateSets(Operation *op) {
   return result;
 }
 
-std::vector<TTNNLayoutAttr>
-MemoryLayoutPropagation::generateReshardCandidates(RankedTensorType tensorType,
-                                             TTNNLayoutAttr currentLayout) {
+std::vector<TTNNLayoutAttr> MemoryLayoutPropagation::generateReshardCandidates(
+    RankedTensorType tensorType, TTNNLayoutAttr currentLayout) {
   // Only generate sharded-to-sharded reshard candidates. Resharding from
   // sharded to interleaved (DRAM or L1) almost always hurts performance.
   if (!tensorTypePossibleLayouts) {
@@ -961,7 +959,7 @@ MemoryLayoutPropagation::generateReshardCandidates(RankedTensorType tensorType,
 
 Operation *
 MemoryLayoutPropagation::getProducerForOperandIdx(Operation *op,
-                                            size_t tensorOperandIdx) {
+                                                  size_t tensorOperandIdx) {
   size_t tensorIdx = 0;
   for (auto operand : op->getOperands()) {
     if (!mlir::isa<RankedTensorType>(operand.getType())) {
@@ -1076,9 +1074,8 @@ void MemoryLayoutPropagation::consolidateBeam() {
                "consolidateBeam: backward pass complete");
 }
 
-size_t
-MemoryLayoutPropagation::resolveForForkPoint(Operation *forkOp,
-                                       llvm::ArrayRef<Operation *> consumers) {
+size_t MemoryLayoutPropagation::resolveForForkPoint(
+    Operation *forkOp, llvm::ArrayRef<Operation *> consumers) {
   const auto &forkBeam = beamState[forkOp];
   size_t bestK = 0;
   int bestFreeCount = -1;
@@ -1113,7 +1110,8 @@ MemoryLayoutPropagation::resolveForForkPoint(Operation *forkOp,
   return bestK;
 }
 
-TTNNLayoutAttr MemoryLayoutPropagation::getDRAMInterleavedFallback(Operation *op) {
+TTNNLayoutAttr
+MemoryLayoutPropagation::getDRAMInterleavedFallback(Operation *op) {
   if (op->getNumResults() == 0) {
     return nullptr;
   }
@@ -1194,7 +1192,7 @@ void MemoryLayoutPropagation::applyToIR() {
 }
 
 void MemoryLayoutPropagation::applyOpConfig(Operation *op,
-                                      const BeamCandidate &candidate) {
+                                            const BeamCandidate &candidate) {
   TTNNLayoutAttr chosenLayout = getOutputLayoutForResult(candidate, 0);
   if (!chosenLayout) {
     return;
@@ -1273,8 +1271,8 @@ void MemoryLayoutPropagation::applyOpConfig(Operation *op,
 }
 
 void MemoryLayoutPropagation::insertReshardOp(Operation *consumerOp,
-                                        size_t operandIndex,
-                                        TTNNLayoutAttr reshardLayout) {
+                                              size_t operandIndex,
+                                              TTNNLayoutAttr reshardLayout) {
   Value operand = consumerOp->getOperand(operandIndex);
   auto producerTensorType = mlir::cast<RankedTensorType>(operand.getType());
 
--- a/lib/Dialect/TTNN/Transforms/OptimizerPasses/GreedyMemoryLayoutPropagation.cpp
+++ b/lib/Dialect/TTNN/Transforms/OptimizerPasses/GreedyMemoryLayoutPropagation.cpp
@@ -6,10 +6,10 @@
 
 #include "ttmlir/Dialect/TTCore/IR/TTCoreOpsTypes.h"
 #include "ttmlir/Dialect/TTCore/IR/Utils.h"
-#include "ttmlir/Dialect/TTNN/Analysis/MemoryLayoutPropagation.h"
 #include "ttmlir/Dialect/TTNN/Analysis/LegalOpConfigAnalysis.h"
 #include "ttmlir/Dialect/TTNN/Analysis/LegalOpLayoutAnalysis.h"
 #include "ttmlir/Dialect/TTNN/Analysis/LegalTensorLayoutAnalysis.h"
+#include "ttmlir/Dialect/TTNN/Analysis/MemoryLayoutPropagation.h"
 #include "ttmlir/Dialect/TTNN/Analysis/OpConfig.h"
 #include "ttmlir/Dialect/TTNN/Analysis/OpModelStrategy.h"
 #include "ttmlir/Dialect/TTNN/Analysis/OpRules/ConvRules.h"
@@ -35,7 +35,7 @@
 
 namespace mlir::tt::ttnn {
 
-#define GEN_PASS_DEF_TTNNGREEDYLAYOUTPROPAGATION
+#define GEN_PASS_DEF_TTNNGREEDYMEMORYLAYOUTPROPAGATION
 #include "ttmlir/Dialect/TTNN/Transforms/Passes.h.inc"
 
 class TTNNGreedyMemoryLayoutPropagation
@@ -47,11 +47,13 @@ public:
 
   // Custom copy constructor: Pass::Option members are non-copyable, so we
   // delegate to the base copy constructor and default-initialize them.
-  TTNNGreedyMemoryLayoutPropagation(const TTNNGreedyMemoryLayoutPropagation &other)
+  TTNNGreedyMemoryLayoutPropagation(
+      const TTNNGreedyMemoryLayoutPropagation &other)
       : TTNNGreedyMemoryLayoutPropagationBase(other) {}
 
   // Pipeline constructor: accepts complex options beyond what tablegen handles.
-  TTNNGreedyMemoryLayoutPropagation(TTNNGreedyMemoryLayoutPropagationPipelineOptions opts)
+  TTNNGreedyMemoryLayoutPropagation(
+      TTNNGreedyMemoryLayoutPropagationPipelineOptions opts)
       : TTNNGreedyMemoryLayoutPropagationBase() {
     maxLegalLayouts = opts.maxLegalLayouts;
     rowMajorEnabled = opts.rowMajorEnabled;
@@ -189,7 +191,8 @@ protected:
 // Pipeline create function.
 std::unique_ptr<::mlir::Pass> createTTNNGreedyMemoryLayoutPropagation(
     TTNNGreedyMemoryLayoutPropagationPipelineOptions options) {
-  return std::make_unique<TTNNGreedyMemoryLayoutPropagation>(std::move(options));
+  return std::make_unique<TTNNGreedyMemoryLayoutPropagation>(
+      std::move(options));
 }
 
 } // namespace mlir::tt::ttnn
--- a/test/unittests/Optimizer/TestBeamSearch.cpp
+++ b/test/unittests/Optimizer/TestBeamSearch.cpp
@@ -189,8 +189,8 @@ TEST_F(BeamSearchTest, BeamWidthK1ProducesSingleCandidate) {
   }
 
   MemoryLayoutPropagation propagation(func, getDeviceGrid(), legalConfigs,
-                                /*tensorTypePossibleLayouts=*/nullptr,
-                                /*beamWidth=*/1);
+                                      /*tensorTypePossibleLayouts=*/nullptr,
+                                      /*beamWidth=*/1);
   propagation.run();
 
   const auto &beamState = propagation.getBeamState();
@@ -219,8 +219,8 @@ TEST_F(BeamSearchTest, BeamWidthK8ProducesMultipleCandidates) {
   }
 
   MemoryLayoutPropagation propagation(func, getDeviceGrid(), legalConfigs,
-                                /*tensorTypePossibleLayouts=*/nullptr,
-                                /*beamWidth=*/8);
+                                      /*tensorTypePossibleLayouts=*/nullptr,
+                                      /*beamWidth=*/8);
   propagation.run();
 
   const auto &beamState = propagation.getBeamState();
@@ -245,8 +245,8 @@ TEST_F(BeamSearchTest, BeamWidthK8CandidatesSortedByScore) {
   }
 
   MemoryLayoutPropagation propagation(func, getDeviceGrid(), legalConfigs,
-                                /*tensorTypePossibleLayouts=*/nullptr,
-                                /*beamWidth=*/8);
+                                      /*tensorTypePossibleLayouts=*/nullptr,
+                                      /*beamWidth=*/8);
   propagation.run();
 
   const auto &beamState = propagation.getBeamState();
@@ -274,8 +274,8 @@ TEST_F(BeamSearchTest, BackwardPassLinearChainAllOpsHaveFinalChoice) {
   }
 
   MemoryLayoutPropagation propagation(func, getDeviceGrid(), legalConfigs,
-                                /*tensorTypePossibleLayouts=*/nullptr,
-                                /*beamWidth=*/4);
+                                      /*tensorTypePossibleLayouts=*/nullptr,
+                                      /*beamWidth=*/4);
   propagation.run();
 
   const auto &beamState = propagation.getBeamState();
@@ -304,8 +304,8 @@ TEST_F(BeamSearchTest, BackwardPassForkPointResolvesWithoutCrash) {
   }
 
   MemoryLayoutPropagation propagation(func, getDeviceGrid(), legalConfigs,
-                                /*tensorTypePossibleLayouts=*/nullptr,
-                                /*beamWidth=*/8);
+                                      /*tensorTypePossibleLayouts=*/nullptr,
+                                      /*beamWidth=*/8);
 
   // Should not crash during fork resolution.
   EXPECT_NO_FATAL_FAILURE(propagation.run());
@@ -337,8 +337,8 @@ TEST_F(BeamSearchTest, BackwardPassForkPointBothConsumersInBeam) {
   }
 
   MemoryLayoutPropagation propagation(func, getDeviceGrid(), legalConfigs,
-                                /*tensorTypePossibleLayouts=*/nullptr,
-                                /*beamWidth=*/4);
+                                      /*tensorTypePossibleLayouts=*/nullptr,
+                                      /*beamWidth=*/4);
   propagation.run();
 
   const auto &beamState = propagation.getBeamState();
@@ -364,8 +364,8 @@ TEST_F(BeamSearchTest, BeamWidthK1NoConsolidation) {
   }
 
   MemoryLayoutPropagation propagation(func, getDeviceGrid(), legalConfigs,
-                                /*tensorTypePossibleLayouts=*/nullptr,
-                                /*beamWidth=*/1);
+                                      /*tensorTypePossibleLayouts=*/nullptr,
+                                      /*beamWidth=*/1);
   propagation.run();
 
   const auto &finalChoice = propagation.getFinalChoice();
@@ -423,8 +423,8 @@ TEST_F(BeamSearchTest, BeamWidthK8ComplexGraphDoesNotCrash) {
   }
 
   MemoryLayoutPropagation propagation(func, getDeviceGrid(), legalConfigs,
-                                /*tensorTypePossibleLayouts=*/nullptr,
-                                /*beamWidth=*/8);
+                                      /*tensorTypePossibleLayouts=*/nullptr,
+                                      /*beamWidth=*/8);
 
   EXPECT_NO_FATAL_FAILURE(propagation.run());
 
@@ -462,8 +462,8 @@ TEST_F(BeamSearchTest, NoValidCandidateFallbackToDRAM) {
   legalConfigs[reluOp.getOperation()] = {};
 
   MemoryLayoutPropagation propagation(func, getDeviceGrid(), legalConfigs,
-                                /*tensorTypePossibleLayouts=*/nullptr,
-                                /*beamWidth=*/8);
+                                      /*tensorTypePossibleLayouts=*/nullptr,
+                                      /*beamWidth=*/8);
 
   // Should not crash -- falls back to DRAM interleaved.
   EXPECT_NO_FATAL_FAILURE(propagation.run());
--- a/test/unittests/Optimizer/TestReshardCandidates.cpp
+++ b/test/unittests/Optimizer/TestReshardCandidates.cpp
@@ -196,7 +196,7 @@ TEST_F(ReshardCandidatesTest, WithTensorLayoutsMapDoesNotCrash) {
       buildTensorTypeLayoutsMap(bareType, shape);
 
   MemoryLayoutPropagation propagation(func, getDeviceGrid(), legalConfigs,
-                                &tensorLayouts, /*beamWidth=*/8);
+                                      &tensorLayouts, /*beamWidth=*/8);
 
   EXPECT_NO_FATAL_FAILURE(propagation.run());
 }
@@ -224,8 +224,8 @@ TEST_F(ReshardCandidatesTest, NullTensorLayoutsNoReshardCandidates) {
   legalConfigs[reluOp.getOperation()] = createElementwiseLegalConfigs(shape);
 
   MemoryLayoutPropagation propagation(func, getDeviceGrid(), legalConfigs,
-                                /*tensorTypePossibleLayouts=*/nullptr,
-                                /*beamWidth=*/8);
+                                      /*tensorTypePossibleLayouts=*/nullptr,
+                                      /*beamWidth=*/8);
   propagation.run();
 
   // With null tensor layouts map, no reshard ops should have been inserted.
@@ -280,8 +280,8 @@ TEST_F(ReshardCandidatesTest, BeamCandidatesWithTensorLayoutsMoreThanWithout) {
     legalConfigs[reluOp.getOperation()] = createElementwiseLegalConfigs(shape);
     legalConfigs[mulOp.getOperation()] = createElementwiseLegalConfigs(shape);
 
-    MemoryLayoutPropagation propagation(localFunc, getDeviceGrid(), legalConfigs,
-                                  layouts, /*beamWidth=*/8);
+    MemoryLayoutPropagation propagation(localFunc, getDeviceGrid(),
+                                        legalConfigs, layouts, /*beamWidth=*/8);
     propagation.run();
 
     size_t totalCandidates = 0;
@@ -344,9 +344,9 @@ TEST_F(ReshardCandidatesTest, ReshardExplorationK1vsK8) {
     legalConfigs[reluOp.getOperation()] = createElementwiseLegalConfigs(shape);
     legalConfigs[mulOp.getOperation()] = createElementwiseLegalConfigs(shape);
 
-    MemoryLayoutPropagation propagation(localFunc, getDeviceGrid(), legalConfigs,
-                                  /*tensorTypePossibleLayouts=*/nullptr,
-                                  beamWidth);
+    MemoryLayoutPropagation propagation(
+        localFunc, getDeviceGrid(), legalConfigs,
+        /*tensorTypePossibleLayouts=*/nullptr, beamWidth);
     propagation.run();
 
     size_t totalCandidates = 0;


test/unittests/Optimizer/TestReshardCandidates.cpp
```diff
@@ -0,0 +1,367 @@
...
+    return func;
+  }
+};
+
+//===----------------------------------------------------------------------===//
+// Reshard candidate generation via LayoutPropagation (tested indirectly
+// through the full pipeline since generateReshardCandidates is private)
+//===----------------------------------------------------------------------===//
+
+TEST_F(ReshardCandidatesTest, WithTensorLayoutsMap_DoesNotCrash) {
```
vmilosevic: :warning: **google\-readability\-avoid\-underscore\-in\-googletest\-name** :warning:
avoid using "\_" in test name "WithTensorLayoutsMap\_DoesNotCrash" according to Googletest FAQ

test/unittests/Optimizer/TestReshardCandidates.cpp
```diff
@@ -0,0 +1,367 @@
...
+  TensorTypeLayoutsMap tensorLayouts =
+      buildTensorTypeLayoutsMap(bareType, shape);
+
+  LayoutPropagation propagation(func, getDeviceGrid(), legalConfigs,
+                                &tensorLayouts, /*beamWidth=*/8);
+
+  EXPECT_NO_FATAL_FAILURE(propagation.run());
+}
+
+TEST_F(ReshardCandidatesTest, NullTensorLayouts_NoReshardCandidates) {
```
vmilosevic: :warning: **google\-readability\-avoid\-underscore\-in\-googletest\-name** :warning:
avoid using "\_" in test name "NullTensorLayouts\_NoReshardCandidates" according to Googletest FAQ

test/unittests/Optimizer/TestReshardCandidates.cpp
```diff
@@ -0,0 +1,367 @@
...
+  const auto &beamState = propagation.getBeamState();
+  for (const auto &[op, candidates] : beamState) {
+    for (const auto &candidate : candidates) {
+      EXPECT_TRUE(candidate.reshardLayouts.empty())
+          << "No reshards should be generated without tensor layouts map";
+    }
+  }
+}
+
+TEST_F(ReshardCandidatesTest, BeamCandidatesWithTensorLayouts_MoreThanWithout) {
```
vmilosevic: :warning: **google\-readability\-avoid\-underscore\-in\-googletest\-name** :warning:
avoid using "\_" in test name "BeamCandidatesWithTensorLayouts\_MoreThanWithout" according to Googletest FAQ

test/unittests/Optimizer/TestReshardCandidates.cpp
```diff
@@ -0,0 +1,367 @@
...
+  size_t candidatesWithout = runWithLayouts(nullptr);
+  size_t candidatesWith = runWithLayouts(&tensorLayouts);
+
+  // With tensor layouts, we may have same or more candidates due to
+  // reshard exploration. At minimum it shouldn't have fewer.
+  EXPECT_GE(candidatesWith, candidatesWithout)
+      << "Tensor layouts should enable at least as many candidates";
+}
+
+TEST_F(ReshardCandidatesTest, ReshardExploration_K1vsK8) {
```
vmilosevic: :warning: **google\-readability\-avoid\-underscore\-in\-googletest\-name** :warning:
avoid using "\_" in test name "ReshardExploration\_K1vsK8" according to Googletest FAQ

test/unittests/Optimizer/TestBeamSearch.cpp
```diff
@@ -0,0 +1,482 @@
...
+
+    return {addOp, reluOp, mulOp, add2Op};
+  }
+};
+
+//===----------------------------------------------------------------------===//
+// Beam width K=1 tests
+//===----------------------------------------------------------------------===//
+
+TEST_F(BeamSearchTest, BeamWidthK1_ProducesSingleCandidate) {
```
vmilosevic: :warning: **google\-readability\-avoid\-underscore\-in\-googletest\-name** :warning:
avoid using "\_" in test name "BeamWidthK1\_ProducesSingleCandidate" according to Googletest FAQ

test/unittests/Optimizer/TestBeamSearch.cpp
```diff
@@ -0,0 +1,482 @@
...
+          << "K=1 should produce at most 1 candidate per op";
+    }
+  }
+}
+
+//===----------------------------------------------------------------------===//
+// Beam width K=8 tests
+//===----------------------------------------------------------------------===//
+
+TEST_F(BeamSearchTest, BeamWidthK8_ProducesMultipleCandidates) {
```
vmilosevic: :warning: **google\-readability\-avoid\-underscore\-in\-googletest\-name** :warning:
avoid using "\_" in test name "BeamWidthK8\_ProducesMultipleCandidates" according to Googletest FAQ

test/unittests/Optimizer/TestBeamSearch.cpp
```diff
@@ -0,0 +1,482 @@
...
+  for (size_t i = 0; i < 3; ++i) {
+    auto it = beamState.find(ops[i]);
+    if (it != beamState.end()) {
+      EXPECT_LE(it->second.size(), 8u)
+          << "K=8 should produce at most 8 candidates per op";
+    }
+  }
+}
+
+TEST_F(BeamSearchTest, BeamWidthK8_CandidatesSortedByScore) {
```
vmilosevic: :warning: **google\-readability\-avoid\-underscore\-in\-googletest\-name** :warning:
avoid using "\_" in test name "BeamWidthK8\_CandidatesSortedByScore" according to Googletest FAQ

test/unittests/Optimizer/TestBeamSearch.cpp
```diff
@@ -0,0 +1,482 @@
...
+          << "Candidates should be sorted by score descending";
+    }
+  }
+}
+
+//===----------------------------------------------------------------------===//
+// Backward pass tests
+//===----------------------------------------------------------------------===//
+
+TEST_F(BeamSearchTest, BackwardPass_LinearChain_AllOpsHaveFinalChoice) {
```
vmilosevic: :warning: **google\-readability\-avoid\-underscore\-in\-googletest\-name** :warning:
avoid using "\_" in test name "BackwardPass\_LinearChain\_AllOpsHaveFinalChoice" according to Googletest FAQ

test/unittests/Optimizer/TestBeamSearch.cpp
```diff
@@ -0,0 +1,482 @@
...
+    EXPECT_NE(choiceIt, finalChoice.end())
+        << "Every op in beam should have a finalChoice";
+    if (choiceIt != finalChoice.end()) {
+      EXPECT_LT(choiceIt->second, candidates.size())
+          << "finalChoice index should be within beam bounds";
+    }
+  }
+}
+
+TEST_F(BeamSearchTest, BackwardPass_ForkPoint_ResolvesWithoutCrash) {
```
vmilosevic: :warning: **google\-readability\-avoid\-underscore\-in\-googletest\-name** :warning:
avoid using "\_" in test name "BackwardPass\_ForkPoint\_ResolvesWithoutCrash" according to Googletest FAQ

test/unittests/Optimizer/TestBeamSearch.cpp
```diff
@@ -0,0 +1,482 @@
...
+    EXPECT_NE(choiceIt, finalChoice.end())
+        << "Fork op should have a finalChoice";
+    if (choiceIt != finalChoice.end()) {
+      EXPECT_LT(choiceIt->second, addIt->second.size())
+          << "Fork finalChoice should be within bounds";
+    }
+  }
+}
+
+TEST_F(BeamSearchTest, BackwardPass_ForkPoint_BothConsumersInBeam) {
```
vmilosevic: :warning: **google\-readability\-avoid\-underscore\-in\-googletest\-name** :warning:
avoid using "\_" in test name "BackwardPass\_ForkPoint\_BothConsumersInBeam" according to Googletest FAQ

test/unittests/Optimizer/TestBeamSearch.cpp
```diff
@@ -0,0 +1,482 @@
...
+  auto mulIt = beamState.find(ops[2]);
+  EXPECT_NE(reluIt, beamState.end()) << "relu should be in beam state";
+  EXPECT_NE(mulIt, beamState.end()) << "multiply should be in beam state";
+}
+
+//===----------------------------------------------------------------------===//
+// K=1 backward pass is trivial (no consolidation)
+//===----------------------------------------------------------------------===//
+
+TEST_F(BeamSearchTest, BeamWidthK1_NoConsolidation) {
```
vmilosevic: :warning: **google\-readability\-avoid\-underscore\-in\-googletest\-name** :warning:
avoid using "\_" in test name "BeamWidthK1\_NoConsolidation" according to Googletest FAQ

test/unittests/Optimizer/TestBeamSearch.cpp
```diff
@@ -0,0 +1,482 @@
...
+  for (const auto &[op, idx] : finalChoice) {
+    EXPECT_EQ(idx, 0u) << "K=1 finalChoice should always be 0";
+  }
+}
+
+//===----------------------------------------------------------------------===//
+// Smoke test: complex graph with K=8
+//===----------------------------------------------------------------------===//
+
+TEST_F(BeamSearchTest, BeamWidthK8_ComplexGraph_DoesNotCrash) {
```
vmilosevic: :warning: **google\-readability\-avoid\-underscore\-in\-googletest\-name** :warning:
avoid using "\_" in test name "BeamWidthK8\_ComplexGraph\_DoesNotCrash" according to Googletest FAQ

test/unittests/Optimizer/TestBeamSearch.cpp
```diff
@@ -0,0 +1,482 @@
...
+  // At least some ops should be in beam (mul2 feeds return -> skipped).
+  EXPECT_GE(beamState.size(), 3u)
+      << "Complex graph should have at least 3 ops in beam";
+}
+
+//===----------------------------------------------------------------------===//
+// Edge case: no valid candidates -> DRAM fallback
+//===----------------------------------------------------------------------===//
+
+TEST_F(BeamSearchTest, NoValidCandidate_FallbackToDRAM) {
```
vmilosevic: :warning: **google\-readability\-avoid\-underscore\-in\-googletest\-name** :warning:
avoid using "\_" in test name "NoValidCandidate\_FallbackToDRAM" according to Googletest FAQ

lib/Dialect/TTNN/Analysis/LayoutPropagation.cpp
```diff
@@ -0,0 +1,1450 @@
...
+
+void LayoutPropagation::run() {
+  TTMLIR_DEBUG(ttmlir::LogComponent::GreedyOptimizer,
+               "LayoutPropagation::run() starting for func {0}",
+               func.getName());
+
+  observer_->onStart(func.getName(), beamWidth);
+
+  size_t opIndex = 0;
+  // Forward pass: propagate layouts in topological (IR) order.
```
odjuricicTT: Nit: in scheduled order?

lib/Dialect/TTNN/Analysis/MemoryLayoutPropagation.cpp
```diff
@@ -0,0 +1,1450 @@
...
+      return;
+    }
+    if (!legalConfigs.count(op)) {
+      return;
+    }
+    // Skip ops whose operands all derive from constant/parameter arguments.
+    // These ops (e.g., BFP8 typecast on weights) will be re-hoisted into
+    // const_eval functions. Promoting their output to L1 would cause the
+    // const_eval to return L1 tensors that starve other ops of L1 budget.
+    bool allFromConstEval = op->getNumOperands() > 0 &&
```
odjuricicTT, line 338: Should we also skip if number of operands is 0? e.g. ttnn.full?

odjuricicTT: Also until recently this was not supposed to be possible as const-eval was run just before optimizer, but then the pass was moved.
We could fix this to run another const-eval before optimizer and assert on this condition, what do you think?

include/ttmlir/Dialect/TTNN/Analysis/LayoutPropagationObserver.h
```diff
@@ -0,0 +1,97 @@
...
+/// LayoutPropagation.cpp.
+class LayoutPropagationObserver {
+public:
+  virtual ~LayoutPropagationObserver() = default;
+
+  /// Called at the start of layout propagation.
+  virtual void onStart(llvm::StringRef funcName, size_t beamWidth) {}
+
+  /// An input candidate for one operand of an op.
+  struct InputCandidate {
```
odjuricicTT: Could this be defined somewhere else to more clearly separate class responsibility, as i'd assume that this class is for logging only?

lib/Dialect/TTNN/Analysis/LayoutPropagation.cpp
```diff
@@ -0,0 +1,1450 @@
...
+
+  for (auto operand : op->getOperands()) {
+    auto tensorType = mlir::dyn_cast<RankedTensorType>(operand.getType());
+    if (!tensorType) {
+      continue;
+    }
+    auto currentLayout =
+        mlir::dyn_cast_or_null<TTNNLayoutAttr>(tensorType.getEncoding());
+    if (!currentLayout) {
+      continue;
```
odjuricicTT: When does this happen?

lib/Dialect/TTNN/Analysis/LayoutPropagation.cpp
```diff
@@ -0,0 +1,1450 @@
...
+    }
+
+    std::vector<InputCandidate> candidatesForOperand;
+    constexpr size_t kMaxInputCandidatesPerOperand = 64;
```
odjuricicTT: Should this be extracted as an option like beam width?

lib/Dialect/TTNN/Analysis/MemoryLayoutPropagation.cpp
```diff
@@ -0,0 +1,1450 @@
...
+    // If no producer in beam (func arg or unresolved), use current layout.
+    if (candidatesForOperand.empty()) {
+      InputCandidate ic;
+      ic.layout = currentLayout;
+      ic.producerCandidateIndex = 0;
+      ic.isReshard = false;
+      candidatesForOperand.push_back(ic);
+    }
+
+    if (producerBeam) {
```
odjuricicTT, line 828: Why this if? We could add this fallback even if op is unresolved?

lib/Dialect/TTNN/Analysis/MemoryLayoutPropagation.cpp
```diff
@@ -0,0 +1,1450 @@
...
+      auto ml = candidate.layout.getMemLayout();
+      if (ml && isShardedMemoryLayout(ml.getValue())) {
+        hasL1Sharded = true;
+      } else {
+        hasL1Interleaved = true;
+      }
+    }
+  }
+
+  if (!hasL1Sharded || hasL1Interleaved) {
```
odjuricicTT, line 620: Why !hasL1Sharded? Preventing DRAM -> L1 interleaved?

lib/Dialect/TTNN/Analysis/MemoryLayoutPropagation.cpp
```diff
@@ -0,0 +1,1450 @@
...
+    size_t maxCandidates) {
+  bool hasL1Sharded = false;
+  bool hasL1Interleaved = false;
+  for (const auto &candidate : candidates) {
+    if (candidate.layout.hasL1BufferType()) {
+      auto ml = candidate.layout.getMemLayout();
+      if (ml && isShardedMemoryLayout(ml.getValue())) {
+        hasL1Sharded = true;
+      } else {
+        hasL1Interleaved = true;
```
odjuricicTT, line 615: If ml == nullptr, should this be set to true?

lib/Dialect/TTNN/Analysis/LayoutPropagation.cpp
```diff
@@ -0,0 +1,1450 @@
...
+  // Reshape/Permute: width-sharded inputs round-trip through interleaved
+  // internally in tt-metal; height/block sharding is handled natively.
+  static auto rejectWidthSharded = [](TTNNLayoutAttr layout) {
+    auto ml = layout.getMemLayout();
+    return !(ml && ml.getValue() == TensorMemoryLayout::WidthSharded);
+  };
+  return llvm::TypeSwitch<Operation *, LayoutFilterFn>(op)
+      // ConcatenateHeads: cannot consume any sharded inputs.
+      // https://github.com/tenstorrent/tt-mlir/issues/7145
+      .Case<ConcatenateHeadsOp, ConcatOp>([](auto) { return rejectAllSharded; })
```
odjuricicTT: Most of these are temporary limitations / bugs, not rules. As such we want to make sure that we remove them as soon as this is possible and not just have an ever growing list. It would be good that for each of these we have:
1. Reference to metal issue
2. XFAIL Test in CI that will fail once the issue is resolved in metal and uplifted to tt-mlir.

Can you add an xfail test for this case? It should be possible with the hang detection feature?

lib/Dialect/TTNN/Analysis/MemoryLayoutPropagation.cpp
```diff
@@ -0,0 +1,1450 @@
...
+    return c.outputLayouts[ri];
+  }
+  return c.configHint.outputLayout;
+}
+
+/// Returns an optional filter predicate that rejects invalid input layouts
+/// for a given op.  When the filter returns false, the candidate is removed.
+/// Returns nullptr when no filtering is needed (all layouts accepted).
+using LayoutFilterFn = std::function<bool(TTNNLayoutAttr)>;
+static LayoutFilterFn getInputLayoutFilter(Operation *op) {
```
odjuricicTT, line 81: If possible we should have this function and all similar metal per op specific modeling / workarounds extracted to a single file. It is very useful to get an overview of "all" metal issues influencing optimizer in one place.

lib/Dialect/TTNN/Analysis/LayoutPropagation.cpp
```diff
@@ -0,0 +1,1450 @@
...
+      // Slice: sharded inputs produce incorrect results in tt-metal.
+      // https://github.com/tenstorrent/tt-metal/issues/39074
+      .Case<SliceStaticOp, SliceDynamicOp>(
+          [](auto) { return rejectAllSharded; })
+      .Case<ReshapeOp, PermuteOp>([](auto) { return rejectWidthSharded; })
```
odjuricicTT: Do we have an issue for this?

lib/Dialect/TTNN/Analysis/MemoryLayoutPropagation.cpp
```diff
@@ -0,0 +1,1450 @@
...
+}
+
+void LayoutPropagation::addReshardCandidates(
+    std::vector<InputCandidate> &candidates, Operation *op, Value operand,
+    TTNNLayoutAttr currentLayout, RankedTensorType tensorType,
+    const llvm::SmallVector<BeamCandidate, 0> *producerBeam,
+    Operation *producerOp, size_t resultIdx, size_t maxCandidates) {
+  // Skip reshards for operands derived from constant/parameter arguments.
+  // These will be re-hoisted into const_eval.
+  if (ttcore::valueTracesToConstantArgs(operand)) {
```
odjuricicTT, line 687: Same as in the other comment, we could eliminate this case by recalling const-eval, should discuss with @dmilinkovicTT

lib/Dialect/TTNN/Analysis/OpModelStrategy.cpp
```diff
@@ -0,0 +1,215 @@
...
+          auto memLayout = cfg.outputLayout.getMemLayout();
+          if (memLayout && isShardedMemoryLayout(memLayout.getValue())) {
+            result.fallbackHints.push_back(cfg);
+          }
+        }
+        return result;
+      });
+}
+
+bool shouldExploreReshards(Operation *op) {
```
odjuricicTT, line 16: Is this just a compile time optimization? We should be able to discard the sharded inputs based on constraints calls?

This should be documented.

odjuricicTT: Also can logic from getInputLayoutFilter be reused here?

lib/Dialect/TTNN/Analysis/LayoutPropagation.cpp
```diff
@@ -0,0 +1,1450 @@
...
+                aVol *= d;
+              }
+              for (int64_t d : bShape) {
+                bVol *= d;
+              }
+              return aVol > bVol;
+            });
+
+  // Keep top 8 candidates.
+  constexpr size_t kMaxReshardCandidates = 8;
```
odjuricicTT: Should this also be passed from options?

lib/Dialect/TTNN/Analysis/MemoryLayoutPropagation.cpp
```diff
@@ -0,0 +1,1450 @@
...
+      }
+    }
+    if (!seen) {
+      seenKeys.push_back(key);
+      deduped.push_back(layout);
+    }
+  }
+
+  // Sort by grid volume (product of grid dims) descending -- more cores first.
+  std::sort(deduped.begin(), deduped.end(),
```
odjuricicTT, line 934: Is it possible to use same / part of sorting from layout scoring?

lib/Dialect/TTNN/Analysis/LayoutPropagation.cpp
```diff
@@ -0,0 +1,1450 @@
...
+        }
+      }
+      if (!alreadyPresent) {
+        uniqueReshardLayouts.push_back(reshardLayout);
+      }
+    }
+  }
+
+  // Fan out: for each unique reshard layout, create one candidate per
+  // producer beam index (KxK evaluation).
```
odjuricicTT: Nit: Not KxK, but K x kMaxReshardCandidates? (could be renamed)

lib/Dialect/TTNN/Analysis/LayoutPropagation.cpp
```diff
@@ -0,0 +1,1450 @@
...
+          continue;
+        }
+        if (!validateReshard(op, producerOp, producerOutput, reshardLayout,
+                             resultIdx)) {
+          continue;
+        }
+      }
+      // TODO(rpavlovicTT): Reshard candidates for func args / unresolved
+      // producers are emitted without validation since there is no beam
+      // state to provide a producerOutputLayout.
```
odjuricicTT: Can it not be read from the op itself?

lib/Dialect/TTNN/Analysis/MemoryLayoutPropagation.cpp
```diff
@@ -0,0 +1,1450 @@
...
+
+    if (producerBeam) {
+      addL1InterleavedFallbacks(candidatesForOperand, op, producerBeam,
+                                producerOp, currentLayout, resultIdx,
+                                kMaxInputCandidatesPerOperand);
+    }
+
+    applyInputLayoutFilter(candidatesForOperand, op, currentLayout);
+
+    addReshardCandidates(candidatesForOperand, op, operand, currentLayout,
```
odjuricicTT, line 836: Can applyInputLayoutFilter be run after addReshardCandidates?

Or both before and after, to avoid the duplicated shouldExploreReshards logic?

lib/Dialect/TTNN/Analysis/OpModelStrategy.cpp
```diff
@@ -0,0 +1,215 @@
...
+      // PadOp: ttnn::pad only supports sharded output for
+      // HEIGHT_SHARDED + ROW_MAJOR inputs. Its tile-layout path
+      // (invoke_tile) crashes when given a sharded output memory config
+      // with interleaved input because it unconditionally dereferences
+      // input_tensor.shard_spec()->grid, which is nullopt for
+      // interleaved tensors. The op model validation doesn't catch this
+      // since it only checks sharding constraints when the input is
+      // already sharded. Until tt-metal fixes this, forbid sharded
+      // output hints for PadOp.
+      // Disabling Slice ops due to
```
odjuricicTT: Can you link an issue for this?

lib/Dialect/TTNN/Analysis/OpModelStrategy.cpp
```diff
@@ -0,0 +1,215 @@
...
+}
+
+OutputHints getOutputHints(Operation *op,
+                           const std::vector<OpConfig> &legalConfigs) {
+  return llvm::TypeSwitch<Operation *, OutputHints>(op)
+      .Case<MatmulOp, LinearOp>([&](auto) {
+        // Use partial configs: deduplicate by (bufferType, memLayout),
+        // set ignorePhysicalLayout=true. Backend decides physical layout.
+        auto partialConfigs =
+            optimizer_utils::getUniqueTestConfigsForMatmulLinear(legalConfigs);
```
odjuricicTT: Do we have an issue for matmul config picker to be improved so this becomes redundant?

lib/Dialect/TTNN/Analysis/OpModelStrategy.cpp
```diff
@@ -0,0 +1,215 @@
...
+            SliceDynamicOp, ConcatOp>([&](auto) {
+        auto nonShardedConfigs = filterNonSharded(legalConfigs);
+        return OutputHints{nonShardedConfigs, {}, /*attemptL1Sharding=*/false};
+      })
+      // SDPA decode: tt-metal requires K/V in DRAM, Q either
+      // height-sharded or DRAM, output either height-sharded or DRAM
+      // (GQA further restricts output to DRAM-only).  In practice the
+      // only valid output is DRAM-interleaved via the NULL hint —
+      // every sharded output hint is rejected. Use NULL hint only,
+      // no fallbacks.
```
odjuricicTT: What is preventing us from utilizing the case where the output can be hight sharded? And can we have na issue for that?

lib/Dialect/TTNN/Analysis/LayoutPropagation.cpp
```diff
@@ -0,0 +1,1450 @@
...
+      candidates.push_back(std::move(*result));
+      return sharded;
+    }
+    return false;
+  };
+
+  size_t numOperandSets = inputCandidateSets.size();
+
+  // If no tensor operands (e.g., constant-like ops), just try output hints.
+  if (numOperandSets == 0) {
```
odjuricicTT: These should be const-evaled as well and we should skip them?

lib/Dialect/TTNN/Analysis/LayoutPropagation.cpp
```diff
@@ -0,0 +1,1450 @@
...
+          opIndex, beamState[op][0].configHint.outputLayout.getBufferType(),
+          beamState[op][0].configHint.outputLayout.getMemLayout(),
+          beamState[op][0].score.coreCount, beamState[op][0].score.isSharded,
+          beamState[op][0].score.isL1, beamState[op][0].score.requiresReshard,
+          beamState[op][0].configHint.outputLayout);
+    }
+    ++opIndex;
+  });
+
+  recordInplaceOps(opIndex);
```
odjuricicTT: This is logging only? Nit: leave a comment.

lib/Dialect/TTNN/Analysis/LayoutPropagation.cpp
```diff
@@ -0,0 +1,1450 @@
...
+          }
+        }
+        if (isFork) {
+          break;
+        }
+      }
+
+      if (isFork) {
+        // Fork point: resolve once (first consumer to reach it sets it).
+        if (!resolvedForks.contains(producer)) {
```
odjuricicTT: Can this pass be simplified by not looking at op producers and resolving producer forks, but instead resolving current op forks by looking at the chosen candidates for it's users?

This would remove the resolve forks tracking as processing order guaranties all users are resolved.

odjuricicTT: Also i'm a bit confused here, how can the first consumer set the for if the rest of them are not resolved?

lib/Dialect/TTNN/Analysis/LayoutPropagation.cpp
```diff
@@ -0,0 +1,1450 @@
...
+  TTMLIR_DEBUG(ttmlir::LogComponent::GreedyOptimizer,
+               "LayoutPropagation: processed {0} ops with beamWidth={1}",
+               opIndex, beamWidth);
+
+  // Backward pass: consolidate beam at fork points (only for K > 1).
+  if (beamWidth > 1) {
+    consolidateBeam();
+  }
+
+  recordFinalChoices();
```
odjuricicTT: Also logging only?

odjuricicTT (commented):
Looks great overall! Looking forward to the next one :)

lib/Dialect/TTNN/Analysis/MemoryLayoutPropagation.cpp
rpavlovicTT: Hm, maybe we can't assert because in some weird debug cases we will run without const-eval. For example we need to emitPy but with all ops in a single func.

rpavlovicTT: For now I will skip these, I can't think of other ops than ttnn.full or ttnn.constant.

lib/Dialect/TTNN/Analysis/LayoutPropagation.cpp
rpavlovicTT: Yup

lib/Dialect/TTNN/Analysis/OpModelStrategy.cpp
rpavlovicTT: added documentation. We can't reuse `getInputLayoutFilter` because it basically rejects invalid cases, while `shouldExploreReshards` is just a compile time optimisation. For example, SDPA op doesn't benefit from exploring reshards today, but it doesn't need to reject sharded inputs. Some combination may work. More likely not, but it may work.

lib/Dialect/TTNN/Analysis/OpModelStrategy.cpp
rpavlovicTT: I think we just don't find that combo since it requires K & V in DRAM. It's very restrictive case. So for us to utilise it we'd have to spill K/V to DRAM and reshard Q to HS (or choose it if present among beam candidates).

lib/Dialect/TTNN/Analysis/OpModelStrategy.cpp
rpavlovicTT: Do you mean like auto matmul program config path? https://github.com/tenstorrent/tt-mlir/issues/6705 we have this issue for that

include/ttmlir/Dialect/TTNN/Analysis/LayoutPropagationObserver.h
rpavlovicTT: Yes absolutely

lib/Dialect/TTNN/Analysis/LayoutPropagation.cpp
rpavlovicTT: I have to see if it is fixed in the meantime, if not I will reference the issue and try making a XFAIL test

lib/Dialect/TTNN/Analysis/LayoutPropagation.cpp
rpavlovicTT: There is no issue, I wrote down this in the commit 
```
tt-metal a3ffa27bc4 ("Native reshape_tiled") made reshape accept
sharded inputs, but for width-sharded tensors it internally
round-trips through interleaved (sharded→interleaved→resharded),
which is slower than receiving interleaved directly. The greedy
optimizer was picking width-sharded input (no reshard penalty in
scoring) even though the backend's internal conversion made it
slower than an explicit ShardedToInterleaved + interleaved reshape.
```

I will open an issue for us to follow up on this and see if they made some improvements.

lib/Dialect/TTNN/Analysis/LayoutPropagation.cpp
rpavlovicTT: Probably doesn't, should be an assert

lib/Dialect/TTNN/Analysis/LayoutPropagation.cpp
rpavlovicTT: Yes it can be

lib/Dialect/TTNN/Analysis/MemoryLayoutPropagation.cpp
rpavlovicTT: I want to insert this fallback only from L1 sharded layouts. If producer beam is NULL then it isn't L1 sharded. I didn't find cases when this would matter. We can revisit at some point if we notice them.

lib/Dialect/TTNN/Analysis/MemoryLayoutPropagation.cpp
rpavlovicTT: As noted in some other comment, inputLayoutFilter and shouldExploreReshards are not same. First one is for correctness and second one is for scoping compile time. Anyway we can run inputLayoutFilter after adding reshards.

lib/Dialect/TTNN/Analysis/MemoryLayoutPropagation.cpp
rpavlovicTT: It'd be messy, I think this is simple and straightforward and should not be mixed.

lib/Dialect/TTNN/Analysis/LayoutPropagation.cpp
rpavlovicTT: Simplified this as suggested, forks are now processed after all consumers.

lib/Dialect/TTNN/Analysis/MemoryLayoutPropagation.cpp
rpavlovicTT: Yeah, thinking was there is no sense moving it to L1 interleaved if DRAM interleaved input works because we don't gain performance by an extra copy to L1.

lib/Dialect/TTNN/Analysis/MemoryLayoutPropagation.cpp
rpavlovicTT: If buffer is L1 and memLayout is null, I guess default option is interleaved.

lib/Dialect/TTNN/Analysis/MemoryLayoutPropagation.cpp
bmalesevicTT: Update, this additional const-eval pass is on main now https://github.com/tenstorrent/tt-mlir/commit/2525a1861e7e87779150881a0b845c25985c366b

include/ttmlir/Dialect/TTNN/Analysis/OpRules/ConvRules.h
```diff
@@ -0,0 +1,59 @@
...
+//
+// Output hints:
+//   Use full legal configs (carry Conv2dConfig tied to output hint).
+//   Attempt L1 sharding.
+//
+// Op-specific attributes:
+//   Set Conv2dConfig + DeviceComputeKernelConfig from candidate.
+//
+// Candidate tiebreaker:
+//   Prefer lower act_block_h_override (0=auto is optimal).
```
bmalesevicTT, line 26: If we do manually set another value(!=0), isn't usually the larger the better for performance? It is more L1 memory pressure, but can't we handle it as we do in OperationAndValidationFallback where L1 OOM is handled by making this smaller only when L1 OOM happens?

lib/Dialect/TTNN/Transforms/OptimizerPasses/GreedyMemoryLayoutPropagation.cpp
```diff
@@ -0,0 +1,195 @@
+// SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#include "ttmlir/Dialect/TTNN/Transforms/Passes.h"
+
+#include "ttmlir/Dialect/TTCore/IR/TTCoreOpsTypes.h"
+#include "ttmlir/Dialect/TTCore/IR/Utils.h"
+#include "ttmlir/Dialect/TTNN/Analysis/MemoryLayoutPropagation.h"
```
vmilosevic: :warning: **llvm\-include\-order** :warning:
```suggestion

```

lib/Dialect/TTNN/Transforms/OptimizerPasses/GreedyMemoryLayoutPropagation.cpp
```diff
@@ -0,0 +1,195 @@
...
+#include "ttmlir/Dialect/TTNN/Analysis/LegalOpConfigAnalysis.h"
+#include "ttmlir/Dialect/TTNN/Analysis/LegalOpLayoutAnalysis.h"
+#include "ttmlir/Dialect/TTNN/Analysis/LegalTensorLayoutAnalysis.h"
+#include "ttmlir/Dialect/TTNN/Analysis/OpConfig.h"
```
vmilosevic, line 13: :warning: **llvm\-include\-order** :warning:
```suggestion
#include "ttmlir/Dialect/TTNN/Analysis/MemoryLayoutPropagation.h"
#include "ttmlir/Dialect/TTNN/Analysis/OpConfig.h"
```

lib/Dialect/TTNN/Transforms/OptimizerPasses/GreedyMemoryLayoutPropagation.cpp
```diff
@@ -0,0 +1,195 @@
...
+#include "llvm/ADT/DenseMap.h"
+#include "llvm/Support/ErrorHandling.h"
+
+namespace mlir::tt::ttnn {
+
+#define GEN_PASS_DEF_TTNNGREEDYLAYOUTPROPAGATION
+#include "ttmlir/Dialect/TTNN/Transforms/Passes.h.inc"
+
+class TTNNGreedyMemoryLayoutPropagation
+    : public impl::TTNNGreedyMemoryLayoutPropagationBase<
```
vmilosevic, line 42: :warning: **clang\-diagnostic\-error** :warning:
unknown template name `` TTNNGreedyMemoryLayoutPropagationBase ``

lib/Dialect/TTNN/Transforms/OptimizerPasses/GreedyMemoryLayoutPropagation.cpp
```diff
@@ -0,0 +1,195 @@
...
+          TTNNGreedyMemoryLayoutPropagation> {
+public:
+  using impl::TTNNGreedyMemoryLayoutPropagationBase<
```
vmilosevic, line 45: :warning: **clang\-diagnostic\-error** :warning:
using declaration in class refers into `` impl:: ``, which is not a class

vmilosevic, line 45: :warning: **clang\-diagnostic\-error** :warning:
expected `` ; `` after using declaration
```suggestion
  using impl::TTNNGreedyMemoryLayoutPropagationBase;<
```

lib/Dialect/TTNN/Transforms/OptimizerPasses/GreedyMemoryLayoutPropagation.cpp
```diff
@@ -0,0 +1,195 @@
...
+      TTNNGreedyMemoryLayoutPropagation>::TTNNGreedyMemoryLayoutPropagationBase;
+
+  // Custom copy constructor: Pass::Option members are non-copyable, so we
+  // delegate to the base copy constructor and default-initialize them.
+  TTNNGreedyMemoryLayoutPropagation(const TTNNGreedyMemoryLayoutPropagation &other)
+      : TTNNGreedyMemoryLayoutPropagationBase(other) {}
```
vmilosevic, line 52: :warning: **clang\-diagnostic\-error** :warning:
member initializer `` TTNNGreedyMemoryLayoutPropagationBase `` does not name a non\-static data member or base class

lib/Dialect/TTNN/Transforms/OptimizerPasses/GreedyMemoryLayoutPropagation.cpp
```diff
@@ -0,0 +1,195 @@
...
+
+  // Pipeline constructor: accepts complex options beyond what tablegen handles.
+  TTNNGreedyMemoryLayoutPropagation(TTNNGreedyMemoryLayoutPropagationPipelineOptions opts)
+      : TTNNGreedyMemoryLayoutPropagationBase() {
```
vmilosevic, line 57: :warning: **clang\-diagnostic\-error** :warning:
member initializer `` TTNNGreedyMemoryLayoutPropagationBase `` does not name a non\-static data member or base class

lib/Dialect/TTNN/Transforms/OptimizerPasses/GreedyMemoryLayoutPropagation.cpp
```diff
@@ -0,0 +1,195 @@
...
+    maxLegalLayouts = opts.maxLegalLayouts;
```
vmilosevic, line 58: :warning: **clang\-diagnostic\-error** :warning:
use of undeclared identifier `` maxLegalLayouts ``; did you mean `` OptionNames::maxLegalLayouts ``?
```suggestion
    OptionNames::maxLegalLayouts = opts.maxLegalLayouts;
```

vmilosevic, line 58: :warning: **clang\-diagnostic\-error** :warning:
no viable overloaded `` = ``

lib/Dialect/TTNN/Transforms/OptimizerPasses/GreedyMemoryLayoutPropagation.cpp
```diff
@@ -0,0 +1,195 @@
...
+    rowMajorEnabled = opts.rowMajorEnabled;
```
vmilosevic, line 59: :warning: **clang\-diagnostic\-error** :warning:
use of undeclared identifier `` rowMajorEnabled ``

lib/Dialect/TTNN/Transforms/OptimizerPasses/GreedyMemoryLayoutPropagation.cpp
```diff
@@ -0,0 +1,195 @@
...
+    beamWidth = opts.beamWidth;
```
vmilosevic, line 60: :warning: **clang\-diagnostic\-error** :warning:
use of undeclared identifier `` beamWidth ``

lib/Dialect/TTNN/Transforms/OptimizerPasses/GreedyMemoryLayoutPropagation.cpp
```diff
@@ -0,0 +1,195 @@
...
+    maxInputCandidatesPerOperand = opts.maxInputCandidatesPerOperand;
```
vmilosevic, line 61: :warning: **clang\-diagnostic\-error** :warning:
use of undeclared identifier `` maxInputCandidatesPerOperand ``

lib/Dialect/TTNN/Transforms/OptimizerPasses/GreedyMemoryLayoutPropagation.cpp
```diff
@@ -0,0 +1,195 @@
...
+    maxReshardCandidates = opts.maxReshardCandidates;
```
vmilosevic, line 62: :warning: **clang\-diagnostic\-error** :warning:
use of undeclared identifier `` maxReshardCandidates ``

lib/Dialect/TTNN/Transforms/OptimizerPasses/GreedyMemoryLayoutPropagation.cpp
```diff
@@ -0,0 +1,195 @@
...
+    enableL1ShardingLayouts = opts.enableL1ShardingLayouts;
```
vmilosevic, line 63: :warning: **clang\-diagnostic\-error** :warning:
use of undeclared identifier `` enableL1ShardingLayouts ``

lib/Dialect/TTNN/Transforms/OptimizerPasses/GreedyMemoryLayoutPropagation.cpp
```diff
@@ -0,0 +1,195 @@
...
+    enableDecisionTrace = opts.enableDecisionTrace;
```
vmilosevic, line 64: :warning: **clang\-diagnostic\-error** :warning:
use of undeclared identifier `` enableDecisionTrace ``

lib/Dialect/TTNN/Transforms/OptimizerPasses/GreedyMemoryLayoutPropagation.cpp
```diff
@@ -0,0 +1,195 @@
...
+    decisionTraceDir = std::move(opts.decisionTraceDir);
```
vmilosevic, line 65: :warning: **clang\-diagnostic\-error** :warning:
use of undeclared identifier `` decisionTraceDir ``

lib/Dialect/TTNN/Transforms/OptimizerPasses/GreedyMemoryLayoutPropagation.cpp
```diff
@@ -0,0 +1,195 @@
...
+    enableCompileTimeStats = opts.enableCompileTimeStats;
```
vmilosevic, line 66: :warning: **clang\-diagnostic\-error** :warning:
use of undeclared identifier `` enableCompileTimeStats ``

lib/Dialect/TTNN/Transforms/OptimizerPasses/GreedyMemoryLayoutPropagation.cpp
```diff
@@ -0,0 +1,195 @@
...
+    overrideOutputLayout = std::move(opts.overrideOutputLayout);
+    overrideConv2dConfig = std::move(opts.overrideConv2dConfig);
+  }
+
+  void runOnOperation() final {
```
vmilosevic, line 71: :warning: **clang\-diagnostic\-error** :warning:
only virtual member functions can be marked `` final ``
```suggestion
  void runOnOperation() {
```

lib/Dialect/TTNN/Transforms/OptimizerPasses/GreedyMemoryLayoutPropagation.cpp
```diff
@@ -0,0 +1,195 @@
...
+#ifndef TTMLIR_ENABLE_OPMODEL
+    llvm::llvm_unreachable_internal(
+        "TTNNGreedyMemoryLayoutPropagation pass requires OpModel support to be "
+        "enabled.");
+#else
+    ModuleOp moduleOp = getOperation();
```
vmilosevic, line 77: :warning: **clang\-diagnostic\-error** :warning:
use of undeclared identifier `` getOperation ``; did you mean `` Operation ``?

lib/Dialect/TTNN/Transforms/OptimizerPasses/GreedyMemoryLayoutPropagation.cpp
```diff
@@ -0,0 +1,195 @@
...
+          static_cast<size_t>(maxReshardCandidates));
+      propagation.run();
+    });
+#endif
+  }
+
+protected:
+  ::mlir::Pass::Option<llvm::StringMap<OutputLayoutOverrideParams>,
+                       OutputLayoutOverrideParser>
+      overrideOutputLayout{
```
vmilosevic, line 179: :warning: **clang\-diagnostic\-error** :warning:
no matching constructor for initialization of `` ::mlir::Pass::Option<llvm::StringMap<OutputLayoutOverrideParams>, OutputLayoutOverrideParser> ``

lib/Dialect/TTNN/Transforms/OptimizerPasses/GreedyMemoryLayoutPropagation.cpp
```diff
@@ -0,0 +1,195 @@
...
+          *this, OptionNames::overrideOutputLayout,
+          ::llvm::cl::desc("Override output tensor layout for specific ops."),
+          ::llvm::cl::init(llvm::StringMap<OutputLayoutOverrideParams>())};
+  ::mlir::Pass::Option<llvm::StringMap<Conv2dConfigOverrideParams>,
+                       Conv2dConfigOverrideParser>
+      overrideConv2dConfig{
```
vmilosevic, line 185: :warning: **clang\-diagnostic\-error** :warning:
no matching constructor for initialization of `` ::mlir::Pass::Option<llvm::StringMap<Conv2dConfigOverrideParams>, Conv2dConfigOverrideParser> `` \(aka `` Option<StringMap<mlir::tt::ttnn::Conv2dConfigParams>, mlir::tt::ttnn::Conv2dConfigOverrideParser> ``\)

include/ttmlir/Dialect/TTNN/Analysis/OpRules/DataMovementRules.h
```diff
@@ -0,0 +1,73 @@
...
+//   SliceStaticOp, SliceDynamicOp: reject all sharded inputs
+//     https://github.com/tenstorrent/tt-metal/issues/39074
+//   ReshapeOp, PermuteOp: reject width-sharded inputs
+//     (round-trip through interleaved internally in tt-metal)
+//
+// Output hints:
+//   All: non-sharded only
+//     PadOp: invoke_tile crashes with sharded output + interleaved input
+//     https://github.com/tenstorrent/tt-metal/issues/40898
+//     SliceOps: https://github.com/tenstorrent/tt-metal/issues/38016
```
bmalesevicTT, line 27: The mentioned issue is for slice height-sharded input, so maybe this should be referenced for input layout restrictions as well?

include/ttmlir/Dialect/TTNN/Analysis/OpRules/DataMovementRules.h
```diff
@@ -0,0 +1,73 @@
...
+#include "ttmlir/Dialect/TTNN/Analysis/OpRules/OpRuleBook.h"
+
+namespace mlir::tt::ttnn {
+
+//===----------------------------------------------------------------------===//
+// Data movement ops: Reshape, Permute, Concat, Slice, Pad
+//
+// Input layout restrictions:
+//   ConcatOp: reject all sharded inputs
+//     https://github.com/tenstorrent/tt-mlir/issues/7145
```
bmalesevicTT, line 17: The issue is referring to ConcatenateHeadsOp, the linked tt-metal issue for ConcatOp seems to be resolved, doesn't seem to be the same problem?

include/ttmlir/Dialect/TTNN/Analysis/OpRules/MatmulRules.h
```diff
@@ -0,0 +1,39 @@
...
+//
+// Output hints:
+//   Filter L1-interleaved (worst of both worlds for matmul:
+//   no program config, HiFi4 fallback, L1 pressure).
+//   Use partial configs (dedup by bufferType+memLayout).
+//
+// Op-specific attributes:
+//   Set MatmulProgramConfig from candidate.
+//   Remove activation attr if programConfig has fused activation
+//   (tt-metal #35060: duplicate activations).
```
bmalesevicTT, line 23: Is it currently temporarily simpler to keep it in programConfig than as matmul attr? 
Since the last comment in the issue implicates it will be ignored from programConfig and applied only from matmul attr, which would silently break this - makes more sense to do it as implied there, opposite to this?

include/ttmlir/Dialect/TTNN/Analysis/OpRules/LayoutFilterUtils.h
```diff
@@ -0,0 +1,58 @@
...
+
+/// Filter legalConfigs to only include non-sharded (DRAM or L1-interleaved).
+inline std::vector<OpConfig>
+filterNonSharded(const std::vector<OpConfig> &legalConfigs) {
+  std::vector<OpConfig> result;
+  for (const auto &config : legalConfigs) {
+    if (!config.outputLayout) {
+      result.push_back(config);
+      continue;
+    }
```
bmalesevicTT, line 34: Is this possible, shouldn't we assert against legalConfig having no layout?

lib/Dialect/TTNN/Analysis/OpRules/ConvRules.cpp
```diff
@@ -0,0 +1,114 @@
...
+      return abh.has_value() ? abh.value() : 0;
+    }
+  }
+  return UINT32_MAX;
+}
+
+bool Conv2dRuleBook::preferCandidate(Operation * /*op*/, const BeamCandidate &a,
+                                     const BeamCandidate &b) const {
+  // Prefer act_block_h_override=0 (auto, best), then higher over lower.
+  // Ordering: 0 > 64 > 32 > ...
```
bmalesevicTT, line 66: As mentioned earlier, contradictory to other comments about preferCandidate where "For Conv2d: prefers lower act_block_h_override (0 = auto = optimal)." and "/// Tiebreaker: prefer lower act_block_h_override."

include/ttmlir/Dialect/TTNN/Analysis/OpRules/ConvRules.h
bmalesevicTT: Seems it even is implemented that way 0>higher>lower, but the comments in multiple places such as this one are misleading.

include/ttmlir/Dialect/TTNN/Analysis/OpRules/ConvRules.h
```diff
@@ -0,0 +1,59 @@
...
+  /// Output hints: full legal configs with Conv2dConfig.
+  OutputHints
+  getOutputHints(Operation *op,
+                 const std::vector<OpConfig> &legalConfigs) const override;
+
+  /// Apply Conv2dConfig + DeviceComputeKernelConfig from candidate.
+  void applyOpSpecificAttrs(Operation *op,
+                            const BeamCandidate &candidate) const override;
+
+  /// Tiebreaker: prefer lower act_block_h_override.
```
bmalesevicTT, line 46: Prefer act_block_h_override=0 (auto, best), then higher over lower.

include/ttmlir/Dialect/TTNN/Analysis/OpModelStrategy.h
```diff
@@ -0,0 +1,134 @@
...
+
+/// Score a backend validation result. Per-op customization via TypeSwitch.
+LayoutScore
+scoreCandidate(Operation *op, const OpConfig &config,
+               const op_constraint_validation::ValidationResult &result,
+               bool requiresReshard);
+
+/// Op-specific tiebreaker when two candidates have equal LayoutScores.
+/// Returns true if candidate a is preferred over b.
+/// For Conv2d: prefers lower act_block_h_override (0 = auto = optimal).
```
bmalesevicTT, line 128: Prefers act_block_h_override=0 (auto, best), then higher over lower.

lib/Dialect/TTNN/Analysis/OpRules/DataMovementRules.cpp
```diff
@@ -0,0 +1,91 @@
...
+
+namespace mlir::tt::ttnn {
+
+//===----------------------------------------------------------------------===//
+// ConcatRuleBook
+//===----------------------------------------------------------------------===//
+
+LayoutFilterFn ConcatRuleBook::getInputLayoutFilter() const {
+  // Concat: cannot consume any sharded inputs.
+  // https://github.com/tenstorrent/tt-mlir/issues/7145
```
bmalesevicTT, line 16: The open issues on both tt-mlir and tt-metal refer to concatenate_heads op, is it relevant to Concat as well?

lib/Dialect/TTNN/Analysis/OpRules/DataMovementRules.cpp
```diff
@@ -0,0 +1,91 @@
...
+  return layout_filter_utils::rejectWidthSharded;
+}
+
+bool ReshapeRuleBook::shouldExploreReshards() const { return false; }
+
+OutputHints ReshapeRuleBook::getOutputHints(
+    Operation * /*op*/, const std::vector<OpConfig> &legalConfigs) const {
+  // ReshapeOp, PermuteOp: width-sharded inputs round-trip through interleaved
+  // internally; sharded output not beneficial.
+  return layout_filter_utils::nonShardedOutputHints(legalConfigs);
```
bmalesevicTT, line 68: The explanation is not completely clear to me here, as we already excluded width-sharded inputs, and some untested rational assumption:
For height/block sharded tiled tensors where the reshape preserves per-core data locality (common case: batch/seq dimension reinterpretation), keeping sharded output is beneficial. 
If you found this is never truly beneficial, maybe an additional comment/issue to explain that?

lib/Dialect/TTNN/Analysis/MemoryLayoutPropagation.cpp
```diff
@@ -0,0 +1,1357 @@
...
+  return memLayout && isShardedMemoryLayout(memLayout.getValue());
+}
+
+/// Compute total bytes transferred from DRAM across all inputs.
+static uint64_t
+computeInputDramBytes(const std::vector<TTNNLayoutAttr> &inputLayouts) {
+  uint64_t dramBytes = 0;
+  for (const auto &layout : inputLayouts) {
+    if (layout && !layout.hasL1BufferType()) {
+      uint64_t tensorBytes = layout.getShardSizeInBytes();
```
bmalesevicTT, line 100: As this is interleaved, should we finally rename this to getSizePerCoreInBytes() or smth to avoid confusion of mentioning "shard" in this context?

lib/Dialect/TTNN/Analysis/MemoryLayoutPropagation.cpp
```diff
@@ -0,0 +1,1357 @@
...
+    Operation *op, const OpConfig &hint, size_t hintIdx,
+    const std::vector<TTNNLayoutAttr> &inputLayouts, bool anyReshard,
+    const llvm::SmallVector<size_t> &producerCandidateIndices,
+    const llvm::DenseMap<size_t, TTNNLayoutAttr> &reshardLayouts) {
+  auto result =
+      op_constraint_validation::validateOperation(op, inputLayouts, hint);
+  if (result.isSuccess()) {
+    BeamCandidate candidate;
+    candidate.configHint =
+        OpConfig(result.getFirstActualOutputLayout(), hint.opSpecificAttrs);
```
bmalesevicTT, line 143: If getFirstActualOutputLayout() returns null (empty actualOutputLayouts), the OpConfig is constructed with a null outputLayout, which then crashes in the TRACE log below when .getBufferType() / .getMemLayout() are called without a null check. This shouldn't happen for a success result, but using checkAndGetFirstActualOutputLayout() would assert early instead of crashing cryptically.

lib/Dialect/TTNN/Analysis/OpRules/OpRuleBook.cpp
```diff
@@ -0,0 +1,92 @@
...
+  // Primary: NULL hint only -- let the backend decide output from inputs.
+  // Fallback: sharded configs -- tried only when NULL yields non-sharded
+  // output.
+  OutputHints result;
+  result.attemptL1Sharding = true;
+  result.hints.push_back(OpConfig(TTNNLayoutAttr()));
+  for (const auto &cfg : legalConfigs) {
+    if (!cfg.outputLayout) {
+      continue;
+    }
```
bmalesevicTT, line 36: How does it happen that legal config result has no output layout?

lib/Dialect/TTNN/Analysis/MemoryLayoutPropagation.cpp
```diff
@@ -0,0 +1,1357 @@
...
+  }
+  if (chosenIdx >= it->second.size()) {
+    chosenIdx = 0;
+  }
+  return &it->second[chosenIdx];
+}
+
+/// Get the output layout for a specific result index from a beam candidate.
+/// Falls back to configHint.outputLayout for single-output ops or when
+/// outputLayouts is not populated.
```
bmalesevicTT, line 68: Why single-output ops? Isn't it the same as getting the first output for multiple output op

lib/Dialect/TTNN/Analysis/MemoryLayoutPropagation.cpp
```diff
@@ -0,0 +1,1357 @@
...
+    }
+    for (size_t pIdx = 0; pIdx < producerBeamSize; ++pIdx) {
+      if (candidates.size() >= maxCandidates) {
+        break;
+      }
+      // Validate the reshard is feasible for this producer candidate.
+      TTNNLayoutAttr producerOutput;
+      if (producerBeam) {
+        producerOutput =
+            getOutputLayoutForResult((*producerBeam)[pIdx], resultIdx);
```
bmalesevicTT, line 753: If producer beam has no output layouts validated on backend, this returns hint output, do we want that? What value does checking reshard for that case bring, seems like a confusing path

lib/Dialect/TTNN/Analysis/MemoryLayoutPropagation.cpp
```diff
@@ -0,0 +1,1357 @@
...
+      } else {
+        // Func args / unresolved producers: use the current IR layout.
```
bmalesevicTT, line 755: I would replace "unresolved" with "untracked"/"non-participating"? Please correct me if I got it wrong but these are the ops we never process in the beam search, hence they will never be resolved, as they don't even have a beam state, an example being ToLayoutOp?

lib/Dialect/TTNN/Analysis/MemoryLayoutPropagation.cpp
```diff
@@ -0,0 +1,1357 @@
...
+      Operation *consumer = consumers[0];
+      size_t consumerIdx = finalChoice[consumer];
+      const BeamCandidate &consumerChosen = beamState[consumer][consumerIdx];
+
+      // Find which tensor operand of consumer connects to this op.
+      for (size_t opIdx = 0;
+           opIdx < consumerChosen.producerCandidateIndices.size(); ++opIdx) {
+        if (getProducerForOperandIdx(consumer, opIdx) == op) {
+          size_t prodIdx = consumerChosen.producerCandidateIndices[opIdx];
+          finalChoice[op] = prodIdx < beamState[op].size() ? prodIdx : 0;
```
bmalesevicTT, line 1021: When is this ever false and why do we just let it pass with choosing its best one, wouldn't it suggest an important bug in finding prodIdx?

lib/Dialect/TTNN/Analysis/MemoryLayoutPropagation.cpp
```diff
@@ -0,0 +1,1357 @@
...
+    if (!alreadyInList) {
+      layoutsToExplore.push_back(ic.layout);
+    }
+  }
+
+  // Collect unique reshard layouts across all base layouts.
+  llvm::SmallVector<TTNNLayoutAttr> uniqueReshardLayouts;
+  for (const auto &baseLayout : layoutsToExplore) {
+    std::vector<TTNNLayoutAttr> reshardCandidates =
+        generateReshardCandidates(tensorType, baseLayout);
```
bmalesevicTT, line 714: I am confused as to why we generate these for each candidate as baseLayout, what differs between the reshards they generate? (apart from each baseLayout excluding itself and the non-sharded ones returning an empty vector) Wouldn't it be simpler to check for any sharded layouts and filter just once, instead of doing this whole loop and dedup logic.

include/ttmlir/Dialect/TTNN/Analysis/MemoryLayoutPropagation.h
```diff
@@ -0,0 +1,182 @@
...
+  /// Observer (NullObject pattern: always non-null, no-op when tracing
+  /// disabled).
+  std::unique_ptr<LayoutPropagationObserver> observer;
+
+  /// Process a single op. Returns top-K candidates sorted by score descending.
+  llvm::SmallVector<BeamCandidate, 0> processOp(Operation *op);
+
+  /// Build per-operand input candidate sets.
+  /// For greedy (K=1): each operand gets {resolved_producer_layout}
+  ///   + reshard candidates (if shouldExploreReshards).
```
bmalesevicTT, line 95: After adding producer's K/BeamWidth candidate outputs, if sharded candidates exist, addL1InterleavedFallbacks() is added and capped at maxInputCandidatesPerOperand, and only then potential reshards, also later capped at maxInputCandidatesPerOperand. So shouldn't L1 layout be mentioned here too, even if K=1, if that resolved one is sharded, we will have 1+L1Interleaved+reshards?

lib/Dialect/TTNN/Analysis/MemoryLayoutPropagation.cpp
```diff
@@ -0,0 +1,1357 @@
...
+
+      // Patch reshardLayouts for consumers that assumed a different producer
+      // candidate. Without this, applyToIR won't insert a ToMemoryConfigOp
+      // and the consumer gets a mismatched input layout.
+      size_t chosenK = finalChoice[op];
+      for (Operation *user : consumers) {
+        size_t userChosenIdx = finalChoice[user];
+        if (userChosenIdx >= beamState[user].size()) {
+          continue;
+        }
```
bmalesevicTT, line 1045: Which case causes this as expected behavior without implying bugs?

bmalesevicTT (commented):
Went through this to understand the new beam search infrastructure. Left some inline comments — mostly questions and minor observations, nothing blocking.

Issue #7117, opened 2026-02-19
broadcast not being folded


AleksKnezevic:
We're running out of dram in a certain model, is it possible to fold one or both of the broadcasts in the following graph snippet:

```
%271 = "ttir.reshape"(%270) <{shape = [1 : i32, 32 : i32, 30720 : i32]}> : (tensor<32x30720xf32>) -> tensor<1x32x30720xf32> loc(#loc170)
        %272 = "ttir.typecast"(%10) <{conservative_folding = false}> : (tensor<30720xbf16>) -> tensor<30720xf32> loc(#loc171)
        %273 = "ttir.reshape"(%272) <{shape = [1 : i32, 1 : i32, 30720 : i32]}> : (tensor<30720xf32>) -> tensor<1x1x30720xf32> loc(#loc172)
        %274 = "ttir.broadcast"(%273) <{broadcast_dimensions = array<i64: 1, 32, 1>}> : (tensor<1x1x30720xf32>) -> tensor<1x32x30720xf32> loc(#loc172)
        %275 = "ttir.add"(%271, %274) : (tensor<1x32x30720xf32>, tensor<1x32x30720xf32>) -> tensor<1x32x30720xf32> loc(#loc173)
        %276 = "ttir.reshape"(%275) <{shape = [1 : i32, 32 : i32, 6 : i32, 5120 : i32]}> : (tensor<1x32x30720xf32>) -> tensor<1x32x6x5120xf32> loc(#loc174)
        %277 = "ttir.reshape"(%276) <{shape = [1 : i32, 32 : i32, 1 : i32, 6 : i32, 5120 : i32]}> : (tensor<1x32x6x5120xf32>) -> tensor<1x32x1x6x5120xf32> loc(#loc175)
        %278 = "ttir.broadcast"(%277) <{broadcast_dimensions = array<i64: 1, 1, 1560, 1, 1>}> : (tensor<1x32x1x6x5120xf32>) -> tensor<1x32x1560x6x5120xf32> loc(#loc175)
        %279 = "ttir.reshape"(%278) <{shape = [1 : i32, 49920 : i32, 6 : i32, 5120 : i32]}> : (tensor<1x32x1560x6x5120xf32>) -> tensor<1x49920x6x5120xf32> loc(#loc176)
[...]

PR #7472, opened 2026-03-12
Memory optimization pass

djstefanovicTT:
### Ticket
Workaround for https://github.com/tenstorrent/tt-mlir/issues/7117

### Problem description
Adding pass that reorganizes operations for lower DRAM pressure.

### What's changed
Added several patterns that reorganize ops to reduce the volume of tensors allocated in DRAM. These patterns consist of:

- slice propagation patterns that move slice upwards through graph so actual operations are done on smaller tensors
- repeat relocation to inner dimensions to create tile aligned tensors that consume less memory 
- elementwise + reshape reorganization that also creates tile aligned tensors that consume less memory 

Perfomance benchmark tests do not point to any significant changes in execution times:
[Without pass](https://github.com/tenstorrent/tt-xla/actions/runs/22911147560)
[With pass](https://github.com/tenstorrent/tt-xla/actions/runs/22911177565)

### Checklist

### DIFF
--- a/include/ttmlir/Dialect/TTNN/Pipelines/TTNNPipelines.h
+++ b/include/ttmlir/Dialect/TTNN/Pipelines/TTNNPipelines.h
@@ -264,9 +264,10 @@ struct TTIRToTTNNDevicePipelineOptions
       llvm::cl::desc("Enable implicit broadcast folding pass."),
       llvm::cl::init(true)};
 
-  Option<bool> memoryManagementEnabled{
-      *this, "enable-memory-management-pass",
-      llvm::cl::desc("Enable memory management pass."), llvm::cl::init(true)};
+  Option<bool> dramSpaceSavingOptimizationEnabled{
+      *this, "enable-dram-space-saving-optimization-pass",
+      llvm::cl::desc("Enable DRAM space saving optimization pass."),
+      llvm::cl::init(false)};
 
   Option<bool> eraseInverseOpsEnabled{
       *this, "enable-erase-inverse-ops-pass",
--- a/lib/Dialect/TTNN/Pipelines/TTNNPipelines.cpp
+++ b/lib/Dialect/TTNN/Pipelines/TTNNPipelines.cpp
@@ -289,7 +289,7 @@ void createTTIRToTTNNDevicePipeline(
     createTTNNPipelineLoweringPasses(devicePm, options.removeDeadValuesEnabled);
     createTTNNFusingPass(devicePm, options);
 
-    if (options.memoryManagementEnabled) {
+    if (options.dramSpaceSavingOptimizationEnabled) {
       devicePm.addPass(createTTNNMemoryManagement());
     }
     createTTNNPipelineWorkaroundPass(devicePm, options);
--- a/test/ttmlir/Dialect/TTNN/data_movement/repeat/simple_repeat.mlir
+++ b/test/ttmlir/Dialect/TTNN/data_movement/repeat/simple_repeat.mlir
@@ -1,4 +1,4 @@
-// RUN: ttmlir-opt --split-input-file --ttir-to-ttnn-backend-pipeline="enable-implicit-broadcast-folding-pass=false enable-memory-management-pass=false" -o %t %s
+// RUN: ttmlir-opt --split-input-file --ttir-to-ttnn-backend-pipeline="enable-implicit-broadcast-folding-pass=false" -o %t %s
 // RUN: FileCheck %s --input-file=%t
 module {
   func.func @main(%arg0: tensor<1x16x32xf32>, %arg1: tensor<1x1x32xf32>) -> tensor<1x16x32xf32> {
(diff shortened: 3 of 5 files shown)


lib/Dialect/TTNN/Transforms/TTNNMemoryManagement.cpp
```diff
@@ -0,0 +1,638 @@
...
+
+static bool hasNonTileAlignedInnerDims(ArrayRef<int64_t> shape) {
+  int64_t rank = shape.size();
+  if (rank < 2) {
+    return false;
+  }
+  return (shape[rank - 1] % 32 != 0) || (shape[rank - 2] % 32 != 0);
+}
+
+class PropagateSliceThroughPermute
```
sshonTT, line 99: Could you add comments to each pattern explaining how the graph is transformed after the pattern is applied?

lib/Dialect/TTNN/Transforms/TTNNMemoryManagement.cpp
```diff
@@ -0,0 +1,638 @@
...
+  }
+  return inIdx == inRank && outIdx == outRank;
+}
+
+static bool hasNonTileAlignedInnerDims(ArrayRef<int64_t> shape) {
+  int64_t rank = shape.size();
+  if (rank < 2) {
+    return false;
+  }
+  return (shape[rank - 1] % 32 != 0) || (shape[rank - 2] % 32 != 0);
```
sshonTT: Not necessarily in this PR, but it might be worth considering tile size utilization. For instance, we could check if `(shape[rank - 1] % tile_size) < (utilization * tile_size)` to find target.

Also, could you get the tile size from ttcore? I believe there's an API to get the tile size for current device.

lib/Dialect/TTNN/Transforms/TTNNMemoryManagement.cpp
djstefanovicTT: Good suggestion, updated.

lib/Dialect/TTNN/Transforms/TTNNMemoryManagement.cpp
```diff
@@ -0,0 +1,638 @@
...
+        ttnn::ShapeAttr::get(rewriter.getContext(), adjustedDims);
+
+    auto newRepeatOp = rewriter.create<ttnn::RepeatOp>(
+        op.getLoc(), newRepeatType, op.getInput(), newRepeatDimsAttr,
+        /*memory_config=*/nullptr);
+
+    // The reshape stays the same — it takes the new repeat output
+    // (different shape, same element count) and produces the same final shape.
+    rewriter.replaceOp(op, newRepeatOp.getResult());
+    return success();
```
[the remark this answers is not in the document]
djstefanovicTT: Added stricter checks.

lib/Dialect/TTNN/Transforms/TTNNMemoryManagement.cpp
djstefanovicTT: Now using API to get tile size.

include/ttmlir/Dialect/TTNN/Pipelines/TTNNPipelines.h
```diff
@@ -264,6 +264,10 @@ struct TTIRToTTNNDevicePipelineOptions
       llvm::cl::desc("Enable implicit broadcast folding pass."),
       llvm::cl::init(true)};
 
+  Option<bool> memoryManagementEnabled{
```
sdjordjevicTT: It is a bit unclear name, in my opinion, much better would be something like this:
dramOptimizationEnabled

rpavlovicTT: Or even more specific as `dramSpaceSavingOptimizationEnabled`

djstefanovicTT:
> Use this tool https://github.com/rpavlovicTT/gh-perf-report to compare perf with and without you commit.
> 
> I did not look into code, but from what I understand from the PR description, you are moving ops so that less DRAM is consumed, right?
> 
> Another question, did you consider implications on L1 sharding and performance of the operations?
> 
> ```c++
> ReshapeElementwiseAdjusting
> ...
> 
>   // This pattern pulls a reshape before an elementwise op by reshaping each
>   // operand to the final target space, then rebuilding the eltwise op.
>   // This is only done when the elementwise tiled volume is greater than the
>   // reshape tiled volume.
>   // For example: add([1024, 1, 1, 1], [1024, 1, 1, 1]) -> [1024, 1, 1, 1]
>   // reshape [1024, 1, 1, 1] -> [1, 1, 32, 32]
>   // The result is equivalent to:
>   // reshape [1024, 1, 1, 1] -> [1, 1, 32, 32]
>   // reshape [1024, 1, 1, 1] -> [1, 1, 32, 32]
>   // add([1, 1, 32, 32], [1, 1, 32, 32]) -> [1, 1, 32, 32]
> ```
> 
> Here you created 2 reshapes instead of one op. Reshape ops usually can't be sharded, so here you are forcing us to use more ops with interleaved layout. Maybe it's worth it since you are saving on work volume of the add op, but maybe it is not. On top of, you are now dispatching 3 ops instead of 2. Anyway, I think more thorough investigation should be done.

I ran the tool to compare, and there are no major changes between the runs. Every model is between -6% and +10% samples per second. This reorganization is necessary for operations that are having significant tile misalignment, that operate on tenors that are 32 times larger than they are supposed to be. Perhaps a check could be added to only rearrange if op is already in dram? What other investigation should be done?

odjuricicTT (changes requested):
Some of the patterns introduced here can have negative perf implications and potential to "randomize" op order in the graph.

Also it's not always clear that we want to optimize for less DRAM in this way. My hunch is that for some OOM cases we want to approach the problem differently, e.g. for conv weight the long term solution could be that the tensor does not get to the device at all before it is prepared. And after prepare, it will be nicely tiled on the last dims.

Can we keep this off by default and use it only where needed for now?

PR #7375, opened 2026-03-05
D2M: Fusion revamp

jdesousa-TT:
### Ticket
#6586 

### What's changed
Describe the approach used to solve the problem.
Summarize the changes made and its impact.

### Checklist
- [ ] New/Existing tests provide coverage for changes

Issue #6586, opened 2026-01-14
☂️ [d2m] Fusion


nsmithtt:

jdesousa-TT:
current progress on `jdesousa/fusion5` 

added some sub-issues for outstanding side-work

PR #7477, opened 2026-03-12
D2M: Fusion Rework

jdesousa-TT:
**(changed source branch from #7375)**

### Ticket
#6586 

 ## Summary

This PR implements elementwise op fusion for the D2M pipeline, enabling adjacent `d2m.generic` ops to be fused into a single compute region. It also introduces a **Spill & Scratch** pass that manages intermediate tile storage when fused ops exceed DST register capacity, and includes several correctness fixes uncovered during testing.

  ### Core Features

  **Elementwise Fusion (`D2MElementwiseFusion`)**
  - Fuses adjacent `d2m.generic` ops into a single compute region, enabling multi-op SFPU compute sequences to be orchestrated within a single kernel. Intermediate results between ops are spilled to L1 scratch (managed by `D2MSpillAndScratch`).
  - Controlled via `--enable-elementwise-fusion` pipeline option (off by default).
  - Added `D2M_SkipOpEltwiseFusionTrait` to selectively exclude ops from fusion.

  **Spill & Scratch (`D2MSpillAndScratch`)** _(new pass, 704 lines)_
  - After fusion, intermediate tile buffers between fused generics are converted to scratch-space allocations in L1.
  - Inserts `d2m.scratch_space_loop` attributes on outer loops to track which loop dimensions require per-iteration scratch reuse.
  - Handles multi-dimensional tiling, cross-step scratch access, and correct scratch buffer shape/indexing.

  **Two-Phase Tiling (`GenericTileComputeLoops`)**
  - Refactored to drive tile compute loop generation from `DstRegisterAnalysis` rather than subblock factors, aligning tiling decisions with DST register constraints and the fusion structure.

  **DST Register Allocation (`InsertDstRegisterAccess`)**
  - Extended to handle the non-scheduled (fusion) path: both operands of f32 binary ops are now loaded into DST slots before compute (SFPU path), with output overwriting the first input's slot to maximize DST utilization.
  - DST slot tracking through chains of compute ops via `dstIntermediates` map.

  **f32 / SFPU Routing**
  - Binary ops (`tile_add`, `tile_sub`, `tile_mul`) now route to SFPU when either operand is `f32`, preserving accuracy. `getOperandsLoadFromDstRegister()` returns `{0, 1}` for f32 (both to DST/SFPU), `{}` for bf16 tile×tile (FPU reads from CBs directly).

  **New Op: `d2m.unpack_stall_on_pack`**
  - Inserted between back-to-back dependent `linalg.generic` ops in a fused region. Waits on a hardware semaphore to stall the UNPACK thread until the previous PACK write is committed to L1.

  ### Bug Fixes

  - **`tile_bcast` hoisting**: Removed `Pure` trait from `D2M_TileBcastOp` to prevent `AffineLoopInvariantCodeMotion` from incorrectly hoisting broadcast ops out of per-tile loops, fixing `test_implicit_bcast_inner_2D` f32 failures.
  - **DST slot for in-place ops after chains**: Added `getInPlaceDstSlice` lambda in `collectDstAccesses` to correctly look up the DST slot of a chained compute op's output (e.g. `tile_nez` following `tile_sub`), fixing `test_comparison_ops` f32 pack index errors.
  - **Accum guard on SFPU loads**: Disabled the Accum guard (`scf.if IV != 0`) for operands loaded from `getOperandsLoadFromDstRegister()`. For SFPU ops, DST is acquired/released per tile so there is no cross-iteration accumulation — the guard was preventing the RHS from ever being loaded in 1-iteration scratch loops, fixing all 27 `test_binary_ops_broadcast_shard_dims` f32 failures.

  ### Tests
  - New lit tests: `spill_and_scratch.mlir`, `insert_dst_register_access_f32_binary.mlir`, `generic_tile_compute_loops.mlir`, `control_dst_section.mlir`
  - New Python golden test: `test_ttir_binary_tree.py` covering fused binary subgraph execution

### Checklist

### DIFF
--- a/include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
+++ b/include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
@@ -113,23 +113,7 @@ class D2M_GenericRegionComputeFPUOrSFPUBinary<string mnemonic, list<Trait> trait
     // Returns the operand indices that need to be loaded from the dst register.
     // For FPU, we don't load from DST.
     // For SFPU (either via f32 operands or scalar operands) we load from DST.
-    mlir::SmallVector<int64_t> getOperandsLoadFromDstRegister() {
-      assert(this->getNumOperands() == 2 && "Binary operation must have 2 operands");
-      auto lhsType = this->getOperand(0).getType();
-      auto rhsType = this->getOperand(1).getType();
-
-      // If either operand is a float32, use SFPU to maintain accuracy. Load both operands from DST.
-      if (ttcore::getDataType(lhsType) == ttcore::DataType::Float32 || ttcore::getDataType(rhsType) == ttcore::DataType::Float32) {
-        return {0, 1};
-      }
-
-      // If RHS is a scalar (not a tile), only load LHS from DST (SFPU path)
-      if (!mlir::isa<::mlir::tt::ttcore::TileType>(rhsType)) {
-        return {0};
-      }
-      // Both operands are tiles (not float32) - FPU reads directly from CBs, no DST load needed
-      return {};
-    }
+    mlir::SmallVector<int64_t> getOperandsLoadFromDstRegister();
   }];
 }
 
--- a/lib/Dialect/D2M/IR/D2MGenericRegionOps.cpp
+++ b/lib/Dialect/D2M/IR/D2MGenericRegionOps.cpp
@@ -1602,6 +1602,46 @@ static mlir::OpFoldResult foldScalarIdentity(mlir::Operation *op,
   return nullptr;
 }
 
+// Shared implementation for D2M_GenericRegionComputeFPUOrSFPUBinary ops.
+// Returns the operand indices that need to be loaded from the dst register.
+// For FPU, we don't load from DST.
+// For SFPU (either via f32 operands or scalar operands) we load from DST.
+static SmallVector<int64_t>
+getOperandsLoadFromDstRegisterFPUOrSFPUBinary(Operation *op) {
+  assert(op->getNumOperands() == 2 && "Binary operation must have 2 operands");
+  auto lhsType = op->getOperand(0).getType();
+  auto rhsType = op->getOperand(1).getType();
+
+  // If either operand is a float32, use SFPU to maintain accuracy. Load both
+  // operands from DST.
+  if (mlir::tt::ttcore::getDataType(lhsType) ==
+          mlir::tt::ttcore::DataType::Float32 ||
+      mlir::tt::ttcore::getDataType(rhsType) ==
+          mlir::tt::ttcore::DataType::Float32) {
+    return {0, 1};
+  }
+
+  // If RHS is a scalar (not a tile), only load LHS from DST (SFPU path).
+  if (!mlir::isa<mlir::tt::ttcore::TileType>(rhsType)) {
+    return {0};
+  }
+  // Both operands are tiles (not float32) - FPU reads directly from CBs, no
+  // DST load needed.
+  return {};
+}
+
+SmallVector<int64_t> TileAddOp::getOperandsLoadFromDstRegister() {
+  return getOperandsLoadFromDstRegisterFPUOrSFPUBinary(getOperation());
+}
+
+SmallVector<int64_t> TileSubOp::getOperandsLoadFromDstRegister() {
+  return getOperandsLoadFromDstRegisterFPUOrSFPUBinary(getOperation());
+}
+
+SmallVector<int64_t> TileMulOp::getOperandsLoadFromDstRegister() {
+  return getOperandsLoadFromDstRegisterFPUOrSFPUBinary(getOperation());
+}
+
 mlir::OpFoldResult TileAddOp::fold(FoldAdaptor adaptor) {
   return foldScalarIdentity(getOperation(), adaptor.getRhs(),
                             [](auto v) { return v.isZero(); });


lib/Dialect/D2M/Transforms/SFPUTileLoopFission.cpp
```diff
@@ -245,32 +243,46 @@ struct D2MSFPUTileLoopFission
...
+          // Fission at the first store
+          if (fissionAtStore(currentLoop, rewriter)) {
+            changed = true;
 
-              if (fissionAtStore(forOp, rewriter)) {
-                changed = true;
+            // The fission creates a new loop after currentLoop.
+            // Both the modified currentLoop and the new loop may need
+            // further fissioning.
+            if (auto nextOp = currentLoop->getNextNode()) {
```
vmilosevic: :warning: **llvm\-qualified\-auto** :warning:
`` auto nextOp `` can be declared as `` auto *nextOp ``
```suggestion
            if (auto *nextOp = currentLoop->getNextNode()) {
```

lib/Dialect/D2M/Transforms/SpillAndScratch.cpp
```diff
@@ -0,0 +1,697 @@
...
+  SmallVector<Value> operands;
+  unsigned dimIdx = 0;
+
+  // Helper: add loop with normalization, push to exprs.
+  auto addLoop = [&](affine::AffineForOp loop) -> AffineExpr {
+    operands.push_back(loop.getInductionVar());
+    AffineExpr dim = getAffineDimExpr(dimIdx++, ctx);
+    int64_t lb = loop.getConstantLowerBound();
+    int64_t step = loop.getStepAsInt();
+    if (lb != 0)
```
vmilosevic: :warning: **google\-readability\-braces\-around\-statements** :warning:
```suggestion
    if (lb != 0) {
```

lib/Dialect/D2M/Transforms/SpillAndScratch.cpp
```diff
@@ -0,0 +1,697 @@
...
+      dim = dim - lb;
+    if (step != 1)
```
vmilosevic: :warning: **google\-readability\-braces\-around\-statements** :warning:
```suggestion
}
    if (step != 1)
```

vmilosevic: :warning: **google\-readability\-braces\-around\-statements** :warning:
```suggestion
    if (step != 1) {
```

lib/Dialect/D2M/Transforms/InsertSpillAndScratch.cpp
```diff
@@ -0,0 +1,697 @@
...
+  auto addLoop = [&](affine::AffineForOp loop) -> AffineExpr {
+    operands.push_back(loop.getInductionVar());
+    AffineExpr dim = getAffineDimExpr(dimIdx++, ctx);
+    int64_t lb = loop.getConstantLowerBound();
+    int64_t step = loop.getStepAsInt();
+    if (lb != 0)
+      dim = dim - lb;
+    if (step != 1)
+      dim = dim.floorDiv(step);
+    return dim;
```
vmilosevic, line 249: :warning: **google\-readability\-braces\-around\-statements** :warning:
```suggestion
}
    return dim;
```

lib/Dialect/D2M/Transforms/SpillAndScratch.cpp
```diff
@@ -0,0 +1,697 @@
...
+  // Dimension 0: scratchLoop (normalized pass-through, same step for both).
+  exprs.push_back(addLoop(consumer.scratchLoop));
+
+  // Outer col loop (allLoops[0]): raw IV without step normalization.
+  // The raw value (e.g., 0 or 6 for step-6) is used in the absolute tile index.
+  operands.push_back(consumer.allLoops.front().getInductionVar());
+  AffineExpr d_outerCol = getAffineDimExpr(dimIdx++, ctx);
+  {
+    int64_t lb = consumer.allLoops.front().getConstantLowerBound();
+    if (lb != 0)
```
vmilosevic: :warning: **google\-readability\-braces\-around\-statements** :warning:
```suggestion
    if (lb != 0) {
```

lib/Dialect/D2M/Transforms/InsertSpillAndScratch.cpp
```diff
@@ -0,0 +1,697 @@
...
+
+  // Outer col loop (allLoops[0]): raw IV without step normalization.
+  // The raw value (e.g., 0 or 6 for step-6) is used in the absolute tile index.
+  operands.push_back(consumer.allLoops.front().getInductionVar());
+  AffineExpr d_outerCol = getAffineDimExpr(dimIdx++, ctx);
+  {
+    int64_t lb = consumer.allLoops.front().getConstantLowerBound();
+    if (lb != 0)
+      d_outerCol = d_outerCol - lb;
+    // Do NOT normalize by step — we need the raw batch-start value.
```
vmilosevic, line 264: :warning: **google\-readability\-braces\-around\-statements** :warning:
```suggestion
}
    // Do NOT normalize by step â we need the raw batch-start value.
```

lib/Dialect/D2M/Transforms/SpillAndScratch.cpp
```diff
@@ -0,0 +1,697 @@
...
+      d_outerCol = d_outerCol - lb;
+    // Do NOT normalize by step — we need the raw batch-start value.
+  }
+
+  // Middle dimensions (allLoops[1..-2]): normalized pass-through (row loops).
+  SmallVector<AffineExpr> middleExprs;
+  for (size_t i = 1; i + 1 < consumer.allLoops.size(); ++i)
```
vmilosevic: :warning: **google\-readability\-braces\-around\-statements** :warning:
```suggestion
  for (size_t i = 1; i + 1 < consumer.allLoops.size(); ++i) {
```

lib/Dialect/D2M/Transforms/InsertSpillAndScratch.cpp
```diff
@@ -0,0 +1,697 @@
...
+  }
+
+  // Middle dimensions (allLoops[1..-2]): normalized pass-through (row loops).
+  SmallVector<AffineExpr> middleExprs;
+  for (size_t i = 1; i + 1 < consumer.allLoops.size(); ++i)
+    middleExprs.push_back(addLoop(consumer.allLoops[i]));
+
```
vmilosevic, line 272: :warning: **google\-readability\-braces\-around\-statements** :warning:
```suggestion
}

```

lib/Dialect/D2M/Transforms/SpillAndScratch.cpp
```diff
@@ -0,0 +1,697 @@
...
+    middleExprs.push_back(addLoop(consumer.allLoops[i]));
+
+  // Inner col loop (allLoops[-1]): raw IV (offset within the col batch).
+  operands.push_back(consumer.allLoops.back().getInductionVar());
+  AffineExpr d_innerCol = getAffineDimExpr(dimIdx++, ctx);
+  {
+    int64_t lb = consumer.allLoops.back().getConstantLowerBound();
+    if (lb != 0)
```
vmilosevic: :warning: **google\-readability\-braces\-around\-statements** :warning:
```suggestion
    if (lb != 0) {
```

lib/Dialect/D2M/Transforms/InsertSpillAndScratch.cpp
```diff
@@ -0,0 +1,697 @@
...
+  // Inner col loop (allLoops[-1]): raw IV (offset within the col batch).
+  operands.push_back(consumer.allLoops.back().getInductionVar());
+  AffineExpr d_innerCol = getAffineDimExpr(dimIdx++, ctx);
+  {
+    int64_t lb = consumer.allLoops.back().getConstantLowerBound();
+    if (lb != 0)
+      d_innerCol = d_innerCol - lb;
+    // Do NOT normalize — inner col loop should have step=1.
```
vmilosevic, line 281: :warning: **google\-readability\-braces\-around\-statements** :warning:
```suggestion
}
    // Do NOT normalize â inner col loop should have step=1.
```

lib/Dialect/D2M/Transforms/SpillAndScratch.cpp
```diff
@@ -0,0 +1,697 @@
...
+      d_innerCol = d_innerCol - lb;
+    // Do NOT normalize — inner col loop should have step=1.
+  }
+
+  // Absolute column tile index (invariant across producer/consumer steps).
+  AffineExpr absCol = d_outerCol + d_innerCol;
+
+  // Map to producer scratch layout using producer's column batch size.
+  exprs.push_back(absCol.floorDiv(prodColStep));
+  for (auto me : middleExprs)
```
vmilosevic: :warning: **google\-readability\-braces\-around\-statements** :warning:
```suggestion
  for (auto me : middleExprs) {
```

lib/Dialect/D2M/Transforms/InsertSpillAndScratch.cpp
```diff
@@ -0,0 +1,697 @@
...
+  }
+
+  // Absolute column tile index (invariant across producer/consumer steps).
+  AffineExpr absCol = d_outerCol + d_innerCol;
+
+  // Map to producer scratch layout using producer's column batch size.
+  exprs.push_back(absCol.floorDiv(prodColStep));
+  for (auto me : middleExprs)
+    exprs.push_back(me);
+  exprs.push_back(absCol % prodColStep);
```
vmilosevic, line 292: :warning: **google\-readability\-braces\-around\-statements** :warning:
```suggestion
}
  exprs.push_back(absCol % prodColStep);
```

lib/Dialect/D2M/Transforms/SpillAndScratch.cpp
```diff
@@ -0,0 +1,697 @@
...
+      // *subsequent* scratch_space_loop (i.e., consumer index > producer index).
+      // A consumer that appears before or at the same loop as the producer does
+      // not require cross-loop scratch spilling.
+      bool hasSubsequentExternalConsumer = false;
+      if (allocInfo.producer) {
+        size_t producerIdx =
+            static_cast<size_t>(allocInfo.producer - scratchLoops.data());
+        hasSubsequentExternalConsumer =
+            llvm::any_of(allocInfo.consumers, [&](const ConsumerLoopLoads &c) {
+              if (c.loop == allocInfo.producer)
```
vmilosevic: :warning: **google\-readability\-braces\-around\-statements** :warning:
```suggestion
              if (c.loop == allocInfo.producer) {
```

lib/Dialect/D2M/Transforms/InsertSpillAndScratch.cpp
```diff
@@ -0,0 +1,697 @@
...
+      // not require cross-loop scratch spilling.
+      bool hasSubsequentExternalConsumer = false;
+      if (allocInfo.producer) {
+        size_t producerIdx =
+            static_cast<size_t>(allocInfo.producer - scratchLoops.data());
+        hasSubsequentExternalConsumer =
+            llvm::any_of(allocInfo.consumers, [&](const ConsumerLoopLoads &c) {
+              if (c.loop == allocInfo.producer)
+                return false;
+              size_t consIdx =
```
vmilosevic, line 588: :warning: **google\-readability\-braces\-around\-statements** :warning:
```suggestion
}
              size_t consIdx =
```

vmilosevic (commented):

lib/Dialect/D2M/Transforms/Allocate.cpp
```diff
@@ -1377,12 +1377,22 @@ class D2MAllocate final : public impl::D2MAllocateBase<D2MAllocate> {
...
 
-        // Get the CB argument type for this operand
+        // Get the CB argument type for this operand.
+        // In TTNN mode (positional alloc counting, no d2m.get_cb), scratch
+        // inputs have no inner alloc so the positional index must be adjusted.
         TT_assert(!genericOp->getRegions().empty());
         Region &region = genericOp->getRegions().front();
         TT_assert(region.hasOneBlock());
+        unsigned allocIndex = operandIndex;
+        if (auto scratchInputs = mutableGenericOp.getScratchInputsAttr()) {
```
bgrady-tt, line 1522: I won't block on this, but I think filing an issue on this for cleanup is a good idea. These sorts of details (skip over scratch alloc indices) should be hidden behind d2m.generic API somehow.

include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
```diff
@@ -112,12 +112,19 @@ class D2M_GenericRegionComputeFPUOrSFPUBinary<string mnemonic, list<Trait> trait
 
     mlir::SmallVector<int64_t> getOperandsLoadFromDstRegister() {
       assert(this->getNumOperands() == 2 && "Binary operation must have 2 operands");
+      auto lhsType = this->getOperand(0).getType();
       auto rhsType = this->getOperand(1).getType();
+
+      // If either operand is a float32, use SFPU to maintain accuracy.
+      if (ttcore::getDataType(lhsType) == ttcore::DataType::Float32 || ttcore::getDataType(rhsType) == ttcore::DataType::Float32) {
+        return {0, 1};
```
vwellsTT: I think this func should document what this return value means, it's not obvious at all from context

test/ttmlir/Dialect/D2M/Transforms/spill_and_scratch.mlir
```diff
@@ -0,0 +1,314 @@
+// SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
```
vwellsTT: year

test/ttmlir/Dialect/D2M/Transforms/insert_dst_register_access_f32_binary.mlir
```diff
@@ -0,0 +1,114 @@
+// SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
```
vwellsTT: year

test/ttmlir/Dialect/D2M/Transforms/generic_tile_compute_loops.mlir
```diff
@@ -0,0 +1,294 @@
+// SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
```
vwellsTT: year

lib/Dialect/D2M/Transforms/SpillAndScratch.cpp
```diff
@@ -0,0 +1,704 @@
+// SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
```
vwellsTT: year

include/ttmlir/Dialect/D2M/Utils/DstRegisterAnalysis.h
```diff
@@ -40,7 +40,11 @@ class DSTPackingInfo {
 
 class DstRegisterAnalysis {
 public:
-  DstRegisterAnalysis(Operation *op);
+  /// \param op                      Root operation to analyze.
```
vwellsTT, line 43: not a fan of this syntax, and this comment is pretty redundant imo

lib/Dialect/D2M/Transforms/InsertSpillAndScratch.cpp
```diff
@@ -0,0 +1,704 @@
...
+#include "mlir/IR/Builders.h"
+#include "mlir/IR/PatternMatch.h"
+
+namespace mlir::tt::d2m {
+#define GEN_PASS_DEF_D2MSPILLANDSCRATCH
+#include "ttmlir/Dialect/D2M/Transforms/Passes.h.inc"
+
+namespace {
+
+/// Information about a loop nest with d2m.scratch_space_loop
```
vwellsTT, line 22: nit: sentence-ify comments in this file

include/ttmlir/Dialect/D2M/Transforms/Passes.td
```diff
@@ -115,16 +115,100 @@ def D2MSFPUTileLoopFission : Pass<"d2m-sfpu-tile-loop-fission", "::mlir::ModuleO
...
+  ];
+
+  let dependentDialects = [
+    "mlir::tt::d2m::D2MDialect",
+    "::mlir::affine::AffineDialect",
+    "::mlir::scf::SCFDialect"
   ];
+}
+
+def D2MSpillAndScratch : Pass<"d2m-spill-and-scratch", "::mlir::ModuleOp"> {
```
vwellsTT: I find this name pretty vague

include/ttmlir/Dialect/D2M/Utils/DstRegisterAnalysis.h
vwellsTT: though below could be useful

include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
vwellsTT: imo this function is too big to be encoded in a .td without linting etc now too

lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
```diff
@@ -120,6 +120,40 @@ static Value getDstIdxFromResult(Value d2mOpResult) {
...
+// value must already be defined.
+static void ensureDominatesInsertionPoint(OpBuilder &rewriter, Value value) {
+  Operation *defOp = value.getDefiningOp();
+  if (!defOp) {
+    return; // Block argument - always dominates.
+  }
+
+  Block *insertBlock = rewriter.getInsertionBlock();
+  if (defOp->getBlock() != insertBlock) {
+    return; // Different block - assume dominance is correct.
```
vwellsTT: "assume dominance is correct." why? 😄

lib/Dialect/D2M/Transforms/AddScratchInputs.cpp
```diff
@@ -45,24 +46,17 @@ static MemRefType findTiledInputType(GenericOp genericOp) {
...
 }
 
-// Check if a d2m.generic needs a scratch buffer. Scratch is added when the
-// compute region contains more than one binary FPU op (add/sub/mul), which
-// indicates a fused kernel that may need to spill intermediate results.
-// TODO(ckaravasilisTT): Use a more precise method to determine if a generic
-// needs a scratch buffer and of what size.
+// Check if a d2m.generic needs a scratch buffer. Scratch is needed when the
+// region contains more than one linalg.generic, which means elementwise fusion
+// produced a multi-op kernel whose intermediate results must be spilled to L1.
```
vwellsTT, line 51: definitely nitpicky, but is this true? would we ever fuse to ops which do not get used back-to-back, and thus do not require scratch intermediates?

include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
```diff
@@ -926,6 +933,17 @@ def D2M_SetL1AccumulateOp : D2M_GenericRegionOp<"set_l1_accumulate"> {
     let assemblyFormat = [{ `(` $enable `)` attr-dict }];
 }
 
+def D2M_UnpackStallOnPackOp : D2M_GenericRegionOp<"unpack_stall_on_pack"> {
```
nsmithtt, line 923: Not suggesting we change this PR, but maybe as a follow on, is this not an analysis we could do at ttkernel level?

lib/Dialect/D2M/Transforms/AddScratchInputs.cpp
```diff
@@ -79,8 +73,16 @@ static LogicalResult addScratchToGeneric(GenericOp genericOp) {
     return failure();
   }
 
-  // Only add scratch to fused generics with multiple binary FPU ops
-  // (add/sub/mul) that may need to spill intermediate results.
+  // Skip if already in explicit datamovement form: indexing_maps is empty,
+  // so the indexing-map manipulation below would be incorrect. Scratch inputs
+  // are only added to generics that are still in implicit blocked form.
+  if (genericOp.isExplicitDatamovementForm()) {
```
vwellsTT: I feel like this check should either always be true or false based on pass-ordering, and thus not needed, right?

lib/Dialect/D2M/Transforms/Allocate.cpp
nsmithtt: +1

jdesousa-TT: #7658

include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
jdesousa-TT: We _could_ do this at the ttkernel level. It's just much easier to pick the insert point when in linalg.generic form, since we just inspect the inputs/outputs of each linalg.generic and insert between them. 

Will make an issue since I don't love this detail being so high up either. #7659

include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
jdesousa-TT: This function exists in about 10 declarations now. We can refactor these out in a separate PR if need be, for now just updated the comment on this specific declaration, since it's the most complicated one thus far.

include/ttmlir/Dialect/D2M/Transforms/Passes.td
jdesousa-TT: `InsertSpillAndScratch` sticks with our action verbiage a bit more, changed to that

lib/Dialect/D2M/Transforms/AddScratchInputs.cpp
jdesousa-TT: Right now this is true, as the ElementwiseFusion pass wouldn't fuse ops that are independent. However, this could be different in the future and we'd want to update this comment I suppose.

lib/Dialect/D2M/Transforms/AddScratchInputs.cpp
jdesousa-TT: I think this case broadly just appears in testing, not in full pipeline runs. We had some failures in lit tests in this pass due to tests being pre-written in explicit form.

lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
jdesousa-TT: If the blocks mismatch, then there's two cases:
1. defOp is in a parent block -> so it dominates the current block
2. defOp is in an actual non-dominating block, but this can't happen since the store we are pulling the dstIdx from already uses the value, so the IR is already invalid if this is the case.

This function I'm not a fan of in general, but this is the problem I was facing

```cpp
d2m.compute_op // needs dstIdx after rewrite to ttkernel
%dstIdx = some arith function
memref.store(%dstIdx, %cb, %cbIdx)
```

when we rewrite this using the dstIdx from the store, the compute op references the dstIdx before defintion. This function ensures the dstIdx defining op is moved before the compute op where we require it.

vwellsTT: Gotcha, makes sense. I think comment should make similar explanation though.

lib/Dialect/D2M/Transforms/AddScratchInputs.cpp
vwellsTT: Can we just remove/rewrite the tests instead then? I'm generally opposed to adding special casing like this just for testing, it seems counter-productive to bloat the code with random conditionals just to keep tests happy for non-representative pass orderings.

jdesousa-TT: It seems you actually removed the failing test in your recent PR, it was this one `test/ttmlir/Dialect/D2M/generic/explicit_datamovement_form_sequential_loops.mlir`. 

Will remove this check then, and if we add any explicit form tests like this in the future, it will fail in this pass indicating we shouldn't test non-representative pass orderings then.

lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
jdesousa-TT: Updated the comment

include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
vwellsTT: To be clear, I'm not saying this func shouldn't exist, I think it should just be moved to an actual .h/.cpp file instead of a .td file if it's nontrivial (which it is now).
Thanks for adding comments, but tbh I still don't see what exactly {0,1} *means*; can that be clarified?

jdesousa-TT: Moved the definition for this one into `D2MGenericRegionOps.cpp`

jdesousa-TT: The comment mentions that the return value is the indices of the operands to load into dst. `{0,1}` implies the first two operands will be copied into dst.

vwellsTT: Great, thanks.  Ah, yeah I missed that, cool.

Issue #7372, opened 2026-03-05
[d2m] GenericOp update parallelization


nsmithtt:
## Overview

As a prerequisite change to building cost model heuristics for block factor selection in Allocate.cpp, we need a robust way of updating a generic op's block factors (and grid shape).  We propose a utility method either on the generic op itself, or somewhere common:

```c++
// We need to return a struct because the caller will very likely want to do:
//   rewriter.replaceAllUsesWith(oldGeneric, ret.returnView);
// Where the `returnView`, views the result back into the same result shape
// as the oldGeneric.
// Other struct members are here for convenience.
struct ParallelizedGeneric {
  SmallVector<ViewOp> operands;
  GenericOp genericOp;
  ViewOp returnView;
};

// Uses builder to return a new generic op and associated views.
// The caller is responsible for erasing `this` op if they want.
// Generic must be in Implicit Blocked or Affine Blocked form.
// Supports both tensors and members.
// At least one of newGrid or newBlockFactors must be set.
FailureOr<ParallelizedGeneric> GenericOp::withParallelization(
  OpBuilder& builder,
  std::optional<GridAttr> newGrid,
  std::optional<ArrayRef<int64_t>> newBlockFactors);
```

## Implementation

The implementation involves 3 main steps with step 3 being the most involved.

### 1. Reblock operands

Calculate the new operand grid shapes and use `view_layout` op to reblock them.  Use `view_layout` to reblock the result(tensors)/output(memrefs) back to the original shape.

Since the function args are optional, we should trivially initialize them to a canonical value:
```c++
assert(newGrid || newBlockFactors);

if (!newGrid) {
  newGrid = this->getGrid();
}

if (!newBlockFactors) {
  newBlockFactors = this->getBlockFactorsValue();
}
```

Calculating the operand grid shapes is surprisingly involved, luckily there is existing function `verifyAffineBlocking` which can be used for reference and actually does exactly what we need.  Just like this function, for each operand, we need to:
1. Invert its affine map, plug in `newGrid.getShape()` -> `resultFactors`
2. For each `resultFactors` promote 0's to 1's
3. For each `resultFactors` multiply by the respective block factor
4. Plug this back into the forwards affine map and you'll get the `newOperandGridShape`.
5. `newOperand[i] = utils::reblockTensor(oldOperand[i], newOperandGridShapes[i]);`

[...]

PR #7479, opened 2026-03-12
WithParallelization + GridSelection

anuragsingh-tt:
### Ticket
#7372 

### What's changed
Extract the reblocking logic from GridSelection::recreateGenericOp into a
reusable GenericOp::withParallelization utility, which delegates to a shared withParallelizationImpl helper that
handles operand wrapping via view_layout ops, generic op reconstruction,
region cloning with type repair, and optional return view creation.

GridSelection's recreateGenericOp is now a thin wrapper that calls
withParallelization and replaces the old generic with the new one.

### DIFF
--- a/lib/Dialect/D2M/IR/D2MOps.cpp
+++ b/lib/Dialect/D2M/IR/D2MOps.cpp
@@ -90,8 +90,10 @@ computeReblockedOperandGridShape(d2m::GenericOp genericOp, int64_t operandIndex,
     }
   }
 
-  // Reduction dims (not in output map): total = block_factor.
-  // Parallel dims: total = grid_dim * block_factor.
+  // Parallel dims start from the output grid extent and then multiply in the
+  // block factor.
+  // Reduction dims (i.e. not in the output map) contribute only their block
+  // factor.
   for (unsigned i = 0; i < numLoopDims; ++i) {
     if (loopTotals[i] == 0) {
       loopTotals[i] = newBlockFactors[i];
@@ -2387,20 +2389,13 @@ static void repairParallelizedRegionTypes(Block *newBlock) {
   }
 }
 
-// Rebuild the optional view that matches the original generic result/output.
+// Build a return view for the rebuilt generic when the caller requests one.
 static d2m::ViewLayoutOp createReturnView(d2m::GenericOp thisOp,
                                           d2m::GenericOp newGenericOp,
-                                          OpBuilder &builder,
-                                          bool generateReturnView) {
-  if (!generateReturnView) {
-    return d2m::ViewLayoutOp();
-  }
-
+                                          OpBuilder &builder) {
   OpBuilder::InsertionGuard guard(builder);
   builder.setInsertionPointAfter(newGenericOp);
   if (thisOp.getNumResults() > 0) {
-    // TODO(anuragsingh): insert multiple return views instead of just one.
-    // Required for Allocator.
     return builder.create<d2m::ViewLayoutOp>(thisOp.getLoc(),
                                              thisOp.getResult(0).getType(),
                                              newGenericOp.getResult(0));
@@ -2414,24 +2409,24 @@ static d2m::ViewLayoutOp createReturnView(d2m::GenericOp thisOp,
 }
 
 // Rewrite the IR:
-// - inserting operand views
-// - creating the new generic
-// - cloning/retyping the region ops
-// - optionally creating the return view.
+// - insert operand views
+// - create the new generic
+// - clone/retype the region ops
+// - optionally create the return view.
 static FailureOr<d2m::ParallelizedGeneric>
 withParallelizationImpl(d2m::GenericOp thisOp, OpBuilder &builder,
                         ArrayRef<Type> reblockedTypes, ttcore::GridAttr newGrid,
                         ArrayRef<int64_t> newBlockFactors,
                         bool generateReturnView) {
-  // Step 1: Reblock operands via explicit view_layout ops.
+  // Reblock operands via explicit view_layout ops.
   auto [reblockedOperands, operandViews] =
       createReblockedOperands(thisOp, builder, reblockedTypes);
 
-  // Step 2: Reconstruct the GenericOp shell with new attrs + operands.
+  // Reconstruct the GenericOp shell with new attrs + operands.
   auto newGenericOp = createParallelizedGenericShell(
       thisOp, builder, reblockedOperands, newGrid, newBlockFactors);
 
-  // Step 3: Reconstruct regions, remapping values and repairing type-dependent
+  // Reconstruct regions, remapping values and repairing type-dependent
   // ops.
   for (unsigned regionIndex = 0; regionIndex < thisOp.getNumRegions();
        ++regionIndex) {
@@ -2447,8 +2442,10 @@ withParallelizationImpl(d2m::GenericOp thisOp, OpBuilder &builder,
   }
 
   // Only return a view into the old generic op result if requested.
-  auto returnView =
-      createReturnView(thisOp, newGenericOp, builder, generateReturnView);
+  d2m::ViewLayoutOp returnView = nullptr;
+  if (generateReturnView) {
+    returnView = createReturnView(thisOp, newGenericOp, builder);
+  }
 
   return d2m::ParallelizedGeneric{std::move(operandViews), newGenericOp,
                                   returnView};
--- a/lib/Dialect/D2M/Transforms/GridSelection.cpp
+++ b/lib/Dialect/D2M/Transforms/GridSelection.cpp
@@ -1081,11 +1081,11 @@ deriveGenericGridAttr(d2m::GenericOp genericOp,
 }
 
 // Phase 5: Recreate the d2m.generic with updated operands.
-// After updating all ToLayout and StreamLayout ops, the generic's operands
-// now have new types with optimized grids. We must recreate the generic to
-// reflect these type changes. We derive the generic grid from the output
-// operand's chosen grid and re-derive block factors from the chosen operand
-// grids so the rebuilt generic reflects the full execution plan.
+// After updating ToLayout and StreamLayout ops, the generic's operands have
+// new types with the selected grids. The generic grid is still anchored by the
+// output operand's chosen grid, but we must re-materialize the generic attrs
+// from the selected operand grids after those rewrites so the rebuilt op stays
+// consistent with the new operand types and the derived block factors.
 static void
 recreateGenericOp(d2m::GenericOp genericOp,
                   ArrayRef<llvm::SmallVector<int64_t>> optimalOperandGrids) {


include/ttmlir/Dialect/D2M/IR/D2MOps.h
```diff
@@ -23,6 +23,8 @@
 
 namespace mlir::tt::d2m {
 
+struct ParallelizedGeneric;
```
vwellsTT, line 26: is this needed? I don't see any reference

lib/Dialect/D2M/Transforms/GridSelection.cpp
```diff
@@ -1032,205 +1032,87 @@ static void updateEmptyOps(ArrayRef<EmptyUpdateInfo> emptyOpsToUpdate,
...
-        tensorType.getShape(), viewTensorType.getShape(), builder.getContext());
-    auto view = builder.create<d2m::ViewLayoutOp>(
-        genericOp.getLoc(), viewTensorType, operand.get(), reblockMap,
-        /*reinterpretLayout=*/false);
-    newOperands.push_back(view.getResult());
+  auto [grid, blockFactors] =
+      deriveGridAndBlockFactors(genericOp, optimalOperandGrids, builder);
+  auto ret = genericOp.withParallelization(builder, grid, blockFactors,
+                                           /*generateReturnView=*/false);
+  if (failed(ret)) {
```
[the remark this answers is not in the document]
vwellsTT: this should probably assert, imo, I'd think withParallization should be expected to always return

lib/Dialect/D2M/IR/D2MOps.cpp
```diff
@@ -2183,6 +2313,308 @@ mlir::SmallVector<int64_t> d2m::GenericOp::getBlockFactorsValue() {
...
+    Block &oldBlock = oldRegion.front();
+    // Rebuild block argument types so CB args track reblocked operand shards.
+    SmallVector<Type> newArgTypes;
+    SmallVector<Location> newArgLocs;
+    newArgTypes.reserve(oldBlock.getNumArguments());
+    newArgLocs.reserve(oldBlock.getNumArguments());
+    for (auto [argIndex, oldArg] : llvm::enumerate(oldBlock.getArguments())) {
+      Type newArgType = oldArg.getType();
+      if (argIndex < reblockedOperands.size()) {
+        if (auto oldCBType = mlir::dyn_cast<d2m::CBType>(oldArg.getType())) {
```
vwellsTT: this will need some updating with my recent PR, now we shouldn't be passing CBs as block args anymore

lib/Dialect/D2M/Transforms/GridSelection.cpp
```diff
@@ -1032,187 +1032,87 @@ static void updateEmptyOps(ArrayRef<EmptyUpdateInfo> emptyOpsToUpdate,
...
-        mlir::cast<mlir::RankedTensorType>(operand.get().getType());
-    auto viewTensorType = utils::reblockTensor(tensorType, optimalGrid);
-    auto reblockMap = ttmlir::utils::calculateReblockMap(
-        tensorType.getShape(), viewTensorType.getShape(), builder.getContext());
-    auto view = builder.create<d2m::ViewLayoutOp>(
-        genericOp.getLoc(), viewTensorType, operand.get(), reblockMap,
-        /*reinterpretLayout=*/false);
-    newOperands.push_back(view.getResult());
+  auto [grid, blockFactors] =
+      deriveGridAndBlockFactors(genericOp, optimalOperandGrids, builder);
```
nsmithtt: I don't quite understand this, in the context of grid selection `withParallelization` should take the chosen grid shape for the generic op and blockFactors should be nullopt.

lib/Dialect/D2M/IR/D2MOps.cpp
```diff
@@ -2183,6 +2313,308 @@ mlir::SmallVector<int64_t> d2m::GenericOp::getBlockFactorsValue() {
...
+    std::optional<ArrayRef<int64_t>> newBlockFactors, bool generateReturnView) {
+  TT_assert((newGrid.has_value() || newBlockFactors.has_value()));
+  OpBuilder::InsertionGuard insertionGuard(builder);
+  builder.setInsertionPoint(*this);
+
+  ttcore::GridAttr normalizedGrid = newGrid.value_or(getGrid());
+  SmallVector<int64_t> normalizedBlockFactors =
+      newBlockFactors ? llvm::to_vector(*newBlockFactors)
+                      : getBlockFactorsValue();
+  if (normalizedBlockFactors.size() != getNumDims()) {
```
nsmithtt: Perhaps this should be an assert.

lib/Dialect/D2M/IR/D2MOps.cpp
```diff
@@ -61,6 +61,129 @@ getGridAndShardFromValue(Value v) {
...
 
+// Convenience wrapper for call sites that start from a Value instead of a Type.
+static std::pair<SmallVector<int64_t>, SmallVector<int64_t>>
+getGridAndShardFromValue(Value v) {
+  return getGridAndShardFromShapedType(mlir::cast<ShapedType>(v.getType()));
+}
+
+// Reblock a shaped type to a new grid shape, handling tensor/memref
+// conversions.
+static FailureOr<Type> reblockShapedType(Type oldType,
```
nsmithtt: Can you replace `lib/Dialect/D2M/Utils/Utils.h`:`reblockTensor` with your impl `reblockShapedType` here?

lib/Dialect/D2M/IR/D2MOps.cpp
```diff
@@ -61,6 +61,129 @@ getGridAndShardFromValue(Value v) {
...
+                                         ArrayRef<int64_t> newGridShape) {
+  auto oldShapedType = mlir::cast<ShapedType>(oldType);
+  ArrayRef<int64_t> oldShape = oldShapedType.getShape();
+  for (auto [idx, gridDim] : llvm::enumerate(newGridShape)) {
+    std::size_t shardIdx = idx + newGridShape.size();
```
nsmithtt: Instead of this please use the DeviceLayoutInterface to pull out the grid or shard parts, grep `getGridShape` and `getShardShape` for usage.  See `getDeviceLayout` helper function for easily getting it.

lib/Dialect/D2M/IR/D2MOps.cpp
```diff
@@ -61,6 +61,129 @@ getGridAndShardFromValue(Value v) {
...
+    int64_t gridExtent = oldShape[idx];
+    int64_t shardExtent = oldShape[shardIdx];
+    // If the new grid shape is not a valid reblock of the old shape, return
+    // failure.
+    if ((gridExtent * shardExtent) % gridDim != 0) {
```
nsmithtt: This should be an assert

lib/Dialect/D2M/IR/D2MOps.cpp
```diff
@@ -61,6 +61,129 @@ getGridAndShardFromValue(Value v) {
...
+    return MemRefType::get(newShape, oldMemRefType.getElementType(),
+                           oldMemRefType.getLayout(),
+                           oldMemRefType.getMemorySpace());
+  }
+  return failure();
+}
+
+static FailureOr<d2m::ViewLayoutOp> createReblockView(OpBuilder &builder,
+                                                      Location loc, Value input,
+                                                      Type outputType) {
```
nsmithtt: This is such a common use case for views I feel like it should be a builder on view_layout op.

lib/Dialect/D2M/IR/D2MOps.cpp
nsmithtt: Also it should accept a ShapedType and return a ShapedType, it's unclear if `FailureOr` is needed, as a rule of thumb we should ask ourselves, is this function used in the context of IR validation?
- If yes, then FailureOr is required.
- If no, we should probably be using asserts.

lib/Dialect/D2M/IR/D2MOps.cpp
```diff
@@ -2183,6 +2313,308 @@ mlir::SmallVector<int64_t> d2m::GenericOp::getBlockFactorsValue() {
...
+            if (metalLayout.getMemorySpace() !=
+                ttcore::MemorySpace::DeviceDRAM) {
+              operandViews.push_back(nullptr);
+              reblockedOperands.push_back(definingView.getInput());
+              continue;
+            }
+          }
+        }
+      }
+    }
```
nsmithtt: I don't quite understand why this is needed for outputs? If there is a reblocking view on the output then we definitely need to use the view.

lib/Dialect/D2M/IR/D2MOps.cpp
```diff
@@ -2183,6 +2313,308 @@ mlir::SmallVector<int64_t> d2m::GenericOp::getBlockFactorsValue() {
...
+// - optionally creating the return view.
+static FailureOr<d2m::ParallelizedGeneric>
+withParallelizationImpl(d2m::GenericOp thisOp, OpBuilder &builder,
+                        ArrayRef<Type> reblockedTypes, ttcore::GridAttr newGrid,
+                        ArrayRef<int64_t> newBlockFactors,
+                        bool generateReturnView) {
+  const std::size_t numInputs = thisOp.getInputs().size();
+  const std::size_t numOutputs = thisOp.getOutputs().size();
+
+  // Step 1: Reblock operands via explicit view_layout ops.
```
nsmithtt: Each of these steps seem deserving of their own helper function.

nsmithtt (commented):
Comments inline, but overall this looks great! Nice work

include/ttmlir/Dialect/D2M/IR/D2MOps.h
anuragsingh-tt: Its needed because the generated `GenericOp` declarations reference `ParallelizedGeneric` before its full definition appears later in the header.

nsmithtt: It's probably fine to just put the definition here

lib/Dialect/D2M/IR/D2MOps.cpp
anuragsingh-tt: I have a few legality checks in this function: input/output must be shaped and the volumes must match. Would a static helper with `FailureOr` be preferable instead?

nsmithtt: I think those should be asserts.  FailureOr should be used for cases where we need to deal with invalid IR

lib/Dialect/D2M/IR/D2MOps.cpp
anuragsingh-tt: This workaround existed in the `recreateGenericOps` implementation to write to the underlying base buffer for outputs to avoid later canonicalization errors.

anuragsingh-tt: Checking with @vtangTT

include/ttmlir/Dialect/D2M/IR/D2MOps.h
anuragsingh-tt: The full definition can't be before `#include "ttmlir/Dialect/D2M/IR/D2MOps.h.inc"` because it stores `ViewLayoutOp` and `GenericOp`. They are needed before the struct can be defined.

lib/Dialect/D2M/IR/D2MOps.cpp
```diff
@@ -60,6 +61,129 @@ getGridAndShardFromValue(Value v) {
...
+// conversions.
+static FailureOr<Type> reblockShapedType(Type oldType,
+                                         ArrayRef<int64_t> newGridShape) {
+  auto oldShapedType = mlir::cast<ShapedType>(oldType);
+  ArrayRef<int64_t> oldShape = oldShapedType.getShape();
+  for (auto [idx, gridDim] : llvm::enumerate(newGridShape)) {
+    std::size_t shardIdx = idx + newGridShape.size();
+    int64_t gridExtent = oldShape[idx];
+    int64_t shardExtent = oldShape[shardIdx];
+    // If the new grid shape is not a valid reblock of the old shape, return
```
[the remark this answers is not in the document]
anuragsingh-tt: Not relevant for the refactor.

lib/Dialect/D2M/Transforms/GridSelection.cpp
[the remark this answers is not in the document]
anuragsingh-tt: The assert is valid for GridSelection but I'm not sure yet for Allocate.

lib/Dialect/D2M/IR/D2MOps.cpp
```diff
@@ -2183,6 +2263,321 @@ mlir::SmallVector<int64_t> d2m::GenericOp::getBlockFactorsValue() {
...
+                                          OpBuilder &builder,
+                                          bool generateReturnView) {
+  if (!generateReturnView) {
+    return d2m::ViewLayoutOp();
+  }
+
+  OpBuilder::InsertionGuard guard(builder);
+  builder.setInsertionPointAfter(newGenericOp);
+  if (thisOp.getNumResults() > 0) {
+    // TODO: insert multiple return views instead of just one. Required for
```
missing username/bug in TODO
```suggestion
    // TODO(nsmith): insert multiple return views instead of just one. Required for
```

include/ttmlir/Dialect/D2M/IR/D2MOps.td
```diff
@@ -41,6 +41,8 @@ def D2M_ViewLayoutOp
   let assemblyFormat = [{
     $input `remapping` `=` $remapping attr-dict `:` type($input) `->` type($result)  }];
 
+  let builders = [OpBuilder<(ins "Type":$outputType, "Value":$input)>];
```
nsmithtt, line 44: Won't block the PR on this, minor nit, but it's probably clearer to have `ArrayRef<int64_t> newGridShape` instead of `"Type":$outputType`, this way it's easier to infer what this constructor does and probably easier to use at the call site.  But I could see an argument for having this one too.

lib/Dialect/D2M/IR/D2MOps.cpp
```diff
@@ -61,6 +62,79 @@ getGridAndShardFromValue(Value v) {
...
+  SmallVector<int64_t> operandGrid = operandMap.compose(loopTotals);
+  for (int64_t &dim : operandGrid) {
+    if (dim == 0) {
+      dim = 1;
+    }
+  }
+  return operandGrid;
+}
+
+static Type getShardTypeFromOperand(Value operand, Type oldType) {
```
nsmithtt: This might be good to make part of the DeviceLayoutInterface.

lib/Dialect/D2M/IR/D2MOps.cpp
```diff
@@ -61,6 +62,79 @@ getGridAndShardFromValue(Value v) {
...
+  auto operandShapedType = mlir::dyn_cast<ShapedType>(operand.getType());
+  auto layout = ttcore::getDeviceLayout(operandShapedType);
+  if (!layout) {
+    return oldType;
+  }
+
+  SmallVector<int64_t> shardShape(layout.getShardShape(operandShapedType));
+  if (auto oldTensorType = mlir::dyn_cast<RankedTensorType>(oldType)) {
+    return RankedTensorType::get(shardShape, oldTensorType.getElementType(),
+                                 oldTensorType.getEncoding());
```
nsmithtt: The encoding shouldn't be passed here, DeviceLayoutInterface is only legal for top-level, global shapes that include the grid.

lib/Dialect/D2M/IR/D2MOps.cpp
```diff
@@ -61,6 +62,79 @@ getGridAndShardFromValue(Value v) {
...
+  }
+  if (auto oldMemRefType = mlir::dyn_cast<MemRefType>(oldType)) {
+    return MemRefType::get(shardShape, oldMemRefType.getElementType(),
+                           oldMemRefType.getLayout(),
```
nsmithtt: The layout shouldn't be passed here, DeviceLayoutInterface is only legal for top-level, global shapes that include the grid.

lib/Dialect/D2M/IR/D2MOps.cpp
nsmithtt: Double check if this is needed, I'd prefer to land without this if possible.

lib/Dialect/D2M/IR/D2MOps.cpp
```diff
@@ -2183,6 +2263,321 @@ mlir::SmallVector<int64_t> d2m::GenericOp::getBlockFactorsValue() {
...
+  for (unsigned regionIndex = 0; regionIndex < thisOp.getNumRegions();
+       ++regionIndex) {
+    Region &oldRegion = thisOp.getRegion(regionIndex);
+    Region &newRegion = newGenericOp.getRegion(regionIndex);
+    if (oldRegion.empty()) {
+      continue;
+    }
+
+    Block *newBlock = cloneParallelizedRegion(
+        thisOp, newGenericOp, builder, oldRegion, newRegion, reblockedOperands);
```
nsmithtt, line 2440: Minor nit, won't block PR, but this code below might be better suited as part of `cloneParallelizedRegion` or as part of their own func.

lib/Dialect/D2M/IR/D2MOps.cpp
```diff
@@ -61,6 +62,79 @@ getGridAndShardFromValue(Value v) {
...
+// Convenience wrapper for call sites that start from a Value instead of a Type.
+static std::pair<SmallVector<int64_t>, SmallVector<int64_t>>
+getGridAndShardFromValue(Value v) {
+  return getGridAndShardFromShapedType(mlir::cast<ShapedType>(v.getType()));
+}
+
+// Derive this operand's target grid shape from the new generic execution
+// grid and block factors using the generic's indexing maps.
+static FailureOr<SmallVector<int64_t>>
+computeReblockedOperandGridShape(d2m::GenericOp genericOp, int64_t operandIndex,
```
vwellsTT, line 74: nit: sort of seems like this should be a member func for a given genericOp vs static standalone imo, but I guess subjective

lib/Dialect/D2M/IR/D2MOps.cpp
```diff
@@ -61,6 +62,79 @@ getGridAndShardFromValue(Value v) {
...
+  SmallVector<int64_t> loopTotals(numLoopDims, 0);
+  for (auto [outputDim, expr] : llvm::enumerate(outputMap.getResults())) {
+    if (auto dimExpr = mlir::dyn_cast<AffineDimExpr>(expr)) {
+      if (outputDim < newGridShape.size()) {
+        loopTotals[dimExpr.getPosition()] = newGridShape[outputDim];
+      }
+    }
+  }
+
+  // Reduction dims (not in output map): total = block_factor.
```
vwellsTT: nit: I think both this and below definition comment could be worded a bit more clearly, though I think differentiating between reductions dims and parallel dims is helpful here

lib/Dialect/D2M/IR/D2MOps.cpp
```diff
@@ -2183,6 +2263,321 @@ mlir::SmallVector<int64_t> d2m::GenericOp::getBlockFactorsValue() {
...
+  OpBuilder::InsertionGuard guard(builder);
+  builder.setInsertionPointToStart(newBlock);
+  for (Operation &op : oldBlock) {
+    builder.clone(op, mapping);
+  }
+
+  return newBlock;
+}
+
+// Rebuild the optional view that matches the original generic result/output.
```
vwellsTT: nit: I also find this a bit confusing, "optional" means it may or may not have been present before reblocking?

lib/Dialect/D2M/IR/D2MOps.cpp
```diff
@@ -2183,6 +2263,321 @@ mlir::SmallVector<int64_t> d2m::GenericOp::getBlockFactorsValue() {
...
+  if (!thisOp.getOutputs().empty()) {
+    return builder.create<d2m::ViewLayoutOp>(
+        thisOp.getLoc(), thisOp.getOutputs().front().getType(),
+        newGenericOp.getOutputs().front());
+  }
+  return d2m::ViewLayoutOp();
+}
+
+// Rewrite the IR:
+// - inserting operand views
```
vwellsTT: nit: I find the -ing tense on all of these a bit awkward, I would personally prefer if each bullet was a sentence (e.g. "-Insert operand views."

lib/Dialect/D2M/IR/D2MOps.cpp
```diff
@@ -2183,6 +2263,321 @@ mlir::SmallVector<int64_t> d2m::GenericOp::getBlockFactorsValue() {
...
+// - creating the new generic
+// - cloning/retyping the region ops
+// - optionally creating the return view.
+static FailureOr<d2m::ParallelizedGeneric>
+withParallelizationImpl(d2m::GenericOp thisOp, OpBuilder &builder,
+                        ArrayRef<Type> reblockedTypes, ttcore::GridAttr newGrid,
+                        ArrayRef<int64_t> newBlockFactors,
+                        bool generateReturnView) {
+  // Step 1: Reblock operands via explicit view_layout ops.
```
vwellsTT: nit: I find claude/llms really like inserting these numbered step comments but they're a bit annoying in the long run (if you change the order, you have to re-number a lot of place).  Imo cleaner to try to remove them here and just say "Reblock operands via explicit view_layout ops."

lib/Dialect/D2M/IR/D2MOps.cpp
```diff
@@ -2183,6 +2263,321 @@ mlir::SmallVector<int64_t> d2m::GenericOp::getBlockFactorsValue() {
...
+  return newBlock;
+}
+
+// Rebuild the optional view that matches the original generic result/output.
+static d2m::ViewLayoutOp createReturnView(d2m::GenericOp thisOp,
+                                          d2m::GenericOp newGenericOp,
+                                          OpBuilder &builder,
+                                          bool generateReturnView) {
+  if (!generateReturnView) {
+    return d2m::ViewLayoutOp();
```
vwellsTT: I think you should remove `generateReturnView` from this func and only call the func if the flag is enabled, vs passing it through and returning an empty view here--that seems quite a bit more confusing/error-prone.

lib/Dialect/D2M/Transforms/GridSelection.cpp
```diff
@@ -1032,187 +1033,83 @@ static void updateEmptyOps(ArrayRef<EmptyUpdateInfo> emptyOpsToUpdate,
   }
 }
 
+// Derive grid (including virtual grid mapping) from the
+// optimized operand grids selected by GridSelection, mirroring
+// GenericOp::build.
```
vwellsTT, line 1038: A bit confused on why this is re-derived vs taken from the output now?

vwellsTT (commented):
core logic mostly LGTM, a few nits (mostly on comments)

include/ttmlir/Dialect/D2M/IR/D2MOps.td
anuragsingh-tt: I was thinking the builder would "reblock to this output type." We will have to derive the full output type anyway for newGridShape but I suppose that can be done inside the builder.

lib/Dialect/D2M/IR/D2MOps.cpp
anuragsingh-tt: I can move it to Utils for future reuse but it is not a high level layout attribute like DeviceLayoutInterface suggests.

lib/Dialect/D2M/IR/D2MOps.cpp
anuragsingh-tt: Here "optional" just means the caller may request it or not via generateReturnView. I’ll reword the comment to make it more obvious.

lib/Dialect/D2M/IR/D2MOps.cpp
anuragsingh-tt: I will keep it as a local helper for now.

lib/Dialect/D2M/IR/D2MOps.cpp
anuragsingh-tt: I actually added the steps here myself to track with the implementation in the issue body. I can just remove it.

lib/Dialect/D2M/Transforms/GridSelection.cpp
anuragsingh-tt: Apologies, realize the comment below for Phase 5 was out of date. I updated it. 

We need to re-derive it after the operand rewrites. grid_selection_adjusted_grid.mlir demonstrates this:
- generic grid anchored by output grid
- operand specific reblock
- re-derivation to build generic after rewrites

Issue #7481, opened 2026-03-13
[TTNN] Replace MorehCumSum with just TTNN CumSum op


sdjordjevicTT:
TTNN now has official support for the cumsum op in:
https://github.com/tenstorrent/tt-metal/tree/main/ttnn/cpp/ttnn/operations/reduction/accumulation/cumsum

Even the moreh cum sum op imlementation targets the ttnn cumsum:
https://github.com/tenstorrent/tt-metal/blob/main/ttnn/cpp/ttnn/operations/moreh/moreh_cumsum/moreh_cumsum.cpp

PR #7483, opened 2026-03-13
[cumsum] Remove moreh cumsum op workarounds and retarget to native ttnn::cumsum

sdjordjevicTT:
### Ticket
Closes https://github.com/tenstorrent/tt-mlir/issues/7481

### Problem description
When cumsum was first introduced into the compiler, the only available TTNN kernel was ttnn::moreh_cumsum, which had two significant limitations: it only supported rank-4 tensors and only supported dim < 2. To work around these restrictions, the TTNN workarounds pass decomposed unsupported cumsum ops — broadcasting lower-rank inputs up to rank-4, and inserting transpose/permute pairs to rotate unsupported dimensions into a supported position before and after the op.

The TTNN team has since shipped a native ttnn::cumsum implementation that handles arbitrary ranks and any dimension directly. The Moreh workarounds are therefore no longer needed, and ttnn::moreh_cumsum is no longer the right call target — ttnn::cumsum should be used instead.

### What's changed
- Retarget ttnn.moreh_cumsum → ttnn.cumsum — The TTNN dialect op, runtime, and all conversion passes have been updated to use the new first-class ttnn::cumsum implementation instead of ttnn::moreh_cumsum. As part of this change, the dim attribute type is narrowed from i64 to i32 to match the new API signature. The FlatBuffer schema is updated accordingly (moreh_cumsum.fbs → cumsum.fbs). Cumsum tests are also reorganized — moved out of test_ttir_ops.py into the dedicated test_reduction.py with expanded coverage.
- Remove ttnn.moreh_cumsum workarounds — Two decomposition patterns (CumSumOpDimRewritePattern for dim >= 2 and CumSumOpRankRewritePattern for rank < 4) have been deleted from the TTNN workarounds pass. These patterns existed because the Moreh cumsum kernel had restrictions on supported dimensions and tensor ranks; those restrictions no longer apply.

### Checklist

### DIFF
--- a/test/python/golden/ttir_ops/reduction/test_reduction.py
+++ b/test/python/golden/ttir_ops/reduction/test_reduction.py
@@ -192,7 +192,8 @@ def test_reduction_cpu_hoisted_ops(
         pytest.param([(4, 4, 128)], 2, id="rank3_dim2"),
     ],
 )
-def test_cumsum(shapes: List[Shape], dim: int, request, device):
+@pytest.mark.parametrize("target", ["ttnn", "emitpy"])
+def test_cumsum(shapes: List[Shape], dim: int, request, target, device):
     def module(builder: TTIRBuilder):
         @builder.func(shapes, [torch.float32] * len(shapes))
         def cumsum(
@@ -205,6 +206,7 @@ def test_cumsum(shapes: List[Shape], dim: int, request, device):
     compile_and_execute_ttir(
         module,
         **get_request_kwargs(request),
+        target=target,
         device=device,
     )
 
@@ -219,11 +221,9 @@ def test_cumsum(shapes: List[Shape], dim: int, request, device):
     ],
     ids=["dim1", "dim0", "dim_negative"],
 )
-@pytest.mark.parametrize("target", ["ttnn", "ttmetal"])
 def test_hoisted_cumsum(
     shapes: List[Shape],
     dim: int,
-    target: str,
     request,
     device,
 ):
@@ -239,7 +239,6 @@ def test_hoisted_cumsum(
     compile_and_execute_ttir(
         module,
         test_base=request.node.name,
-        target=target,
         device=device,
         output_root=request.config.getoption("--path"),
         system_desc_path=request.config.getoption("--sys-desc"),


lib/Target/TTNN/TTNNToFlatbuffer.cpp
rpavlovicTT: Just be careful when you rebase, I committed ee7af103bc2f21a01d0be14c13d51e4a61a57d18

runtime/lib/ttnn/operations/reduction/cumsum.cpp
```diff
@@ -2,29 +2,29 @@
...
+void run(const ::tt::target::ttnn::CumSumOp *op, ProgramContext &context) {
   ProgramTensorPool &tensorPool = context.getTensorPool();
   const ::ttnn::Tensor &in = tensorPool.getTTNNTensorAndValidate(op->in());
 
   std::optional<::ttnn::MemoryConfig> outputMemoryConfig =
       ::tt::runtime::ttnn::utils::createMemoryConfigIfNeeded(op->memcfg());
 
-  ::ttnn::Tensor out =
-      ::ttnn::moreh_cumsum(in, op->dim(), std::nullopt, outputMemoryConfig);
+  ::ttnn::Tensor out = ::ttnn::cumsum(in, op->dim(), std::nullopt, false,
```
pilkicTT: nit: pass these hardcoded values as `constexpr` values

lib/Conversion/TTIRToTTNN/TTIRToTTNN.cpp
```diff
@@ -504,9 +504,11 @@ class CumSumOpConversionPattern : public OpConversionPattern<ttir::CumSumOp> {
   LogicalResult
   matchAndRewrite(ttir::CumSumOp op, OpAdaptor adaptor,
                   ConversionPatternRewriter &rewriter) const override {
-    rewriter.replaceOpWithNewOp<ttnn::MorehCumSumOp>(
+    rewriter.replaceOpWithNewOp<ttnn::CumSumOp>(
         op, this->getTypeConverter()->convertType(op.getType()),
-        adaptor.getInput(), adaptor.getDim(), nullptr);
+        adaptor.getInput(),
+        rewriter.getI32IntegerAttr(static_cast<int32_t>(adaptor.getDim())),
```
sgholamiTT, line 516: Nit: is there a particular reason we cannot change `TTIR_CumSumOp`'s `I64Attr:$dim` to `I32Attr:$dim` ? So that we don't need the down cast here ? 
I can see that the `window_dimension` in `stablehlo` is a `si64` ([here](https://openxla.org/stablehlo/spec#reduce_window)) which we lower to `TTIR_CumSumOp`. But I think it's much better to perform the casting at the top level conversion pass (`StablehloToTTIR`) rather than `TTIRToTTNN`.
No strong opinion on this.

sdjordjevicTT: My view is that we should introduce restrictions at the level where they actually become required. It’s possible that in the future, or in another backend such as D2M, TTIR may support dim as i64. Because of that, this feels more like a TTNN-specific constraint rather than a TTIR one.

For that reason, I think handling the cast in the TTIRToTTNN conversion layer is the more appropriate level of abstraction, instead of restricting the TTIR op itself earlier in StablehloToTTIR.

PR #7490, opened 2026-03-13
Update .gitignore to allow for Claude skills

svuckovicTT:
### Ticket
N/A

### Problem description
Whole of `.claude/` was git ignored, not allowing for claude skills to be added.

### What's changed
Add rule to exclude `.claude/skills/` from `.gitignore`.

Personal (local) claude skills should be put in `.claude/skills-local/`.

### Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,8 +5,7 @@ third_party/ttnn-python
 .DS_STORE
 .vscode/*
 .cache
-.claude/*
-!.claude/skills/
+.claude/settings.local.json
 *pycache*
 *.egg-info
 cluster_descriptor.yaml
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -30,7 +30,3 @@ targets tt-metal.
   set the environment variable `SYSTEM_DESC_PATH=$(pwd)/ttrt-artifacts/system_desc.ttsys`.
 - `ttrt run ...`: Runs a compiler generated flatbuffer on silicon.  Can either
   point to a flatbuffer file or a directory containing flatbuffers.
-
-# Claude skills
-- project level skills live in `.claude/skills` dir
-- personal skills should live in `.claude/skills-local` dir


.gitignore
```diff
@@ -6,6 +6,7 @@ third_party/ttnn-python
 .vscode/*
 .cache
 .claude/*
+!.claude/skills/
```
nsmithtt: Not necessarily against this change, but you can still add files that match against the ignorelist. This change runs the risk of accidentally adding skills that you don't intend to check in.

Also, there's no skills checked in, is that intended?

CLAUDE.md
```diff
@@ -30,3 +30,7 @@ targets tt-metal.
   set the environment variable `SYSTEM_DESC_PATH=$(pwd)/ttrt-artifacts/system_desc.ttsys`.
 - `ttrt run ...`: Runs a compiler generated flatbuffer on silicon.  Can either
   point to a flatbuffer file or a directory containing flatbuffers.
+
+# Claude skills
+- project level skills live in `.claude/skills` dir
+- personal skills should live in `.claude/skills-local` dir
```
sdjordjevicTT: I would remove this for personal skills.  Personal skills should live under:
~/.claude/skills/<skill-name>/SKILL.md

.gitignore
nsmithtt: nvm my comment here, I think we should follow the advice from their docs https://code.claude.com/docs/en/settings#available-scopes

i.e. change our gitignore to only ignore this file `.claude/settings.local.json`

CLAUDE.md
nsmithtt: +1

.gitignore
svuckovicTT: Adjusted.

Yeah, no skills yet, couple folks are cooking things atm tho, so wanted to set things up beforehand.

CLAUDE.md
svuckovicTT: Updated!

PR #7493, opened 2026-03-13
[builder] d2m.generic builder support + test

nsmithtt:
- Introduce support for writing d2m.generic op in builder infra.
- Disable goldens, d2m builder can only accept manual golden.
- TileMatmulBlock -> linalg.generic rewriter since it's much easier to use d2m.tile_matmul_block in builder tests.
- Introduce d2m.generic op attr d2m.skip_grid_selection for skipping participation in grid selection, useful if testing manually specified grids.
- conftest harness for setting up perf capture when invoked with tracy.
- system_desc pytest fixture for accessing that info from test.
- Allocator fix for choosing stream buffer size. Note this path is deprecated soon anyway.

### DIFF
--- a/test/python/golden/conftest.py
+++ b/test/python/golden/conftest.py
@@ -799,7 +799,7 @@ def pytest_collection_modifyitems(config, items):
                 )
             )
 
-        def skip_exec_handler(item, platform_config, reason):
+        def skip_exec_handler(item):
             # Set skip_exec attribute on the item instead of marking as skipped
             item.skip_exec = True
             xfail_reason = f"Execution marked as skip"


lib/Dialect/D2M/IR/D2MOps.cpp
```diff
@@ -1696,7 +1696,7 @@ ::mlir::LogicalResult d2m::GenericOp::verify() {
 
     // Only check yield terminator for non-explicit-datamovement form.
     // Explicit datamovement form allows users to manage terminators themselves.
-    if (!isExplicitDatamovementForm()) {
+    if (!isExplicitDatamovementForm() && !this->getRegion(0).empty()) {
```
vwellsTT, line 1766: debatably, could be better to assert/error on empty region?

lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
```diff
@@ -2011,6 +2011,87 @@ class D2MMeshShardOpRewriter : public OpConversionPattern<ttir::MeshShardOp> {
...
+    rewriter.eraseOp(op);
+
+    return llvm::success();
+  }
+
+  static SmallVector<mlir::AffineMap>
+  getAffineMapsArray(mlir::OpBuilder &builder, std::size_t arity,
+                     std::size_t rank) {
+    assert(arity == 3 && "expected 3 operands");
+    // TODO(#2592) handle higher ranks, if needed in this pass
```
vwellsTT: nit: sentence-ify comments

lib/Dialect/D2M/IR/D2MOps.cpp
vwellsTT: I guess really it's probably better like this, I suppose empty region is legal (though we should eventually add some canonicalizing-like pass to remove them probably)

nsmithtt: The problem is there's no way to construct the generic op from python with verification enabled, you must construct it with empty regions and then build the regions.  The problem is that the verifier seems to automatically kick in before we get to region building.

tools/builder/d2m/d2m_builder.py
```diff
@@ -261,7 +240,173 @@ def untilize(
...
+            len(input.type.shape) % 2 == 0
+        ), f"Input shape must be multiple of 2 {input.type.shape}"
+        grid_rank = len(input.type.shape) // 2
+        old_grid = list(input.type.shape)[:grid_rank]
+        old_shard = list(input.type.shape)[grid_rank:]
+        assert (
+            len(new_grid) == grid_rank
+        ), f"Mismatched input/output grid rank for in {new_grid} and out {old_grid}"
+        canonical_shape = [gd * sd for gd, sd in zip(old_grid, old_shard)]
+        output_shape = new_grid
```
xanderchin: From Claude: 
** `output_shape = new_grid` aliases the caller's list (line 261)**
```python
output_shape = new_grid  # alias, not copy
for i, d in enumerate(canonical_shape):
    output_shape.append(d // new_grid[i])  # mutates new_grid!
```
`append` on `output_shape` mutates the original `new_grid` list. If `reblock` is called twice with the same grid, the second call sees corrupted input. **Fix:** `output_shape = list(new_grid)`.

nsmithtt: Man I would have never caught that without a repro

tools/builder/base/builder_utils.py
```diff
@@ -315,3 +315,31 @@ def get_metal_tensor_layout(
...
+
+
+def affine_map_from_lambda(fn):
+    class Dim:
+        def __init__(self, position, name):
+            self.position = position
+            self.name = name
+
+    dims = tuple(
+        Dim(name, i) for name, i in enumerate(inspect.signature(fn).parameters)
```
wenbinlyuTT: The order seems to be swapped? `enumerage` returns `i, v`, but the `__init__` function swapped them back...

PR #7495, opened 2026-03-13
[D2M][ttnn-jit] Nightly perf collection for Superset

sgholamiTT:
## Add JIT performance collection to nightly CI
Runs JIT vs TTNN op-level performance benchmarks in the nightly pipeline and exports structured results to Superset for dashboard visualization.

### What
- **`ttnn_jit.sh`**: The `nightly` path calls `run_perf_collect.sh` to run perf benchmarks alongside the existing nightly tests.
- **`perf_tests.py`**: Parametrized pytest suite comparing JIT and non-JIT execution of `abs`, `exp`, `add`, `mul`, `matmul`, `fusion_test` across dtypes (`bf16`, `bfp8`) and memory configs (`dram_interleaved`). More configs are yet to be decided.
- **`conftest.py`**: Pytest fixture (`perf_device`) that manages device lifecycle and collects per-test kernel durations via `ttnn.ReadDeviceProfiler` / `ttnn.get_latest_programs_perf_data`. Writes a combined `perf_results.json` incrementally after each test.
- **`run_perf_collect.sh`**: Sets profiler environment variables, runs all perf tests in a single pytest session, invokes the summarizer, and appends the report directory to `$UPLOAD_LIST` for artifact upload.
- **`summarize_perf_results.py`**: Reads `perf_results.json`, groups JIT vs non-JIT results by case (op, shape, dtype, memory config), and writes one JSON report per case to a `reports/` subdirectory. Each report maps to a `benchmark_run` row in Superset with columns (`model`=op, `precision`=dtype, `config`=shape/memory/fidelity) and three measurements: `jit_kernel_duration_ns`, `ttnn_kernel_duration_ns`, `perf_ratio`.
- **`call-test.yml`**: Adds `UPLOAD_LIST` support and a generic upload step (`nsumrakTT/upload-multiple-artifacts`) so test scripts can register artifacts for upload.
- **`run_tests_from_json.py`**: Passes `JOB_ID` as an environment variable to test scripts.

### Superset integration

Reports are picked up by the existing `collect_data` action, SFTP'd to the perf ingestion server, and loaded into `sw_test.benchmark_run` / `sw_test.benchmark_measurement` tables. No changes needed to the data pipeline.

Dashboard: https://superset.tenstorrent.com/superset/dashboard/p/WpVZ5d8pj90/

### Note
This workflow should not increase the nightly duration as it runs in parallel with all other text matrix jobs.
Currently, the e2e process takes around 10 mins ([evidence](https://github.com/tenstorrent/tt-mlir/actions/runs/23060470115/job/66988179129))

### Test plan
- [x] Nightly CI run completes successfully and uploads artifacts ([Link](https://github.com/tenstorrent/tt-mlir/actions/runs/23217843772))
- [x] `collect_data` finds and processes all `perf_*_<job_id>.json` reports
- [x] Data appears in Superset `benchmark_run` / `benchmark_measurement` tables


.github/workflows/call-jit-perf-test.yml
```diff
@@ -0,0 +1,160 @@
+name: JIT Perf Test
```
nsumrakTT: Why did you create new test workflow instead of  just adding new test to `.github/settings/tests.json` file as described here: https://docs.tenstorrent.com/tt-mlir/ci.html?

test/ttnn-jit/perf_ci/run_perf_collect.sh
```diff
@@ -0,0 +1,77 @@
...
+for tid in "${TESTS[@]}"; do
+  echo "=============================================="
+  echo "Running test_op_compare[$tid] ..."
+  echo "=============================================="
+  export TT_METAL_PROFILER_DIR="$OUT_DIR/$tid"
+  mkdir -p "$TT_METAL_PROFILER_DIR"
+  if ! python -m tracy -m -r -p "pytest test/ttnn-jit/perf_ci/perf_tests.py::test_op_compare[$tid]"; then
+    echo "Warning: test_op_compare[$tid] exited with non-zero status (results may still be present)."
+  fi
+done
```
nsumrakTT: I don't understand why you collected tests and then execute it one by one? 
* Collect and run is time consuming as pytest will load all the modules required on each run.
* Why using pytest in the first place if you'll run tests one by one?
* If you decide to use pytests than you should utilize conftest.py and create fixtures that will generate perf reports.

.github/workflows/call-jit-perf-test.yml
```diff
@@ -0,0 +1,160 @@
...
+    - name: Set reusable strings
+      id: strings
+      shell: bash
+      env:
+        JOB_ID: ${{ steps.fetch-job-id.outputs.job_id }}
+      run: |
+        echo "work-dir=$(pwd)" >> "$GITHUB_OUTPUT"
+        echo "build-output-dir=$(pwd)/build" >> "$GITHUB_OUTPUT"
+        echo "install-output-dir=$(pwd)/install" >> "$GITHUB_OUTPUT"
+        echo "perf-output-dir=$(pwd)/jit_perf_results" >> "$GITHUB_OUTPUT"
```
nsumrakTT: You haven't utilized perf-output-dir set here.

test/ttnn-jit/perf_ci/run_perf_collect.sh
```diff
@@ -0,0 +1,77 @@
...
+
+echo ""
+echo "Results written under: $OUT_DIR"
+echo "Summarizing..."
+JOB_ID_ARG=""
+if [ -n "$JOB_ID" ]; then
+  JOB_ID_ARG="--job-id $JOB_ID"
+fi
+if python test/ttnn-jit/perf_ci/summarize_perf_results.py "$OUT_DIR" --output-dir "$OUT_DIR" $JOB_ID_ARG; then
```
nsumrakTT: Ouput dir is the same as input dir. This means it will upload all csv along with resulting json. Do you need these csv files?

.github/workflows/call-jit-perf-test.yml
vvukomanTT: The default method for adding new tests should be adding them to `.github/settings/tests.json`.
Was there a specific reason you've created a totally new and separate workflow?

test/ttnn-jit/perf_ci/run_perf_collect.sh
sgholamiTT: Good point — I've reworked this. Tests now run in a single `pytest` session (no more collect-then-run-one-by-one), and per-test perf data collection is handled through a `conftest.py` fixture.

The `perf_device` fixture manages the device lifecycle and uses `ttnn.ReadDeviceProfiler()` / `get_latest_programs_perf_data()` after each test to capture `DEVICE KERNEL DURATION [ns]` per test. Results are written incrementally to `perf_results.json`, which the summarizer then turns into the final JSON reports.

test/ttnn-jit/perf_ci/perf_tests.py
```diff
@@ -0,0 +1,101 @@
...
+    input_b = ttnn.from_torch(
+        torch_tensor_b,
+        dtype=ttnn_dtype,
+        layout=ttnn.TILE_LAYOUT,
+        device=device,
+        memory_config=memory_config,
+    )
+
+    function_to_test = (
+        ttnn_jit.jit(debug=True, enable_cache=True)(op) if jit_enabled else op
```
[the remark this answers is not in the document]
sgholamiTT: Fixed.

test/ttnn-jit/perf_ci/perf_tests.py
```diff
@@ -0,0 +1,101 @@
...
+    )
+    output_tensor = (
+        function_to_test(input_a)
+        if is_unary(op)
+        else function_to_test(input_a, input_b)
+    )
+
+    print(f"output_tensor\n: {output_tensor}")
+    ttnn.close_device(device)
```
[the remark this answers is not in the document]
sgholamiTT: Although this is not entirely true as the output will be truncated, I removed the `print` statement and added `assert output_tensor is not None` .

.github/workflows/schedule-nightly.yml
```diff
@@ -49,6 +49,14 @@ jobs:
       test_matrix: ${{ needs.prepare-run.outputs.test_matrix }}
       timeout: ${{ fromJson(needs.prepare-run.outputs.test_timeout) }}
 
+  jit-perf-test:
+    needs: [ build-image, release-build ]
+    uses: ./.github/workflows/call-jit-perf-test.yml
+    secrets: inherit
+    with:
+      docker_image: ${{ needs.build-image.outputs.docker-image }}
+
```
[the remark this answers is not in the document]
sgholamiTT: Modified the whole structure.

.github/workflows/call-jit-perf-test.yml
```diff
@@ -0,0 +1,160 @@
...
+      run: |
+        source env/activate
+        rm -f ttmlir*.whl ttnn_jit*.whl
+
+        gh run download ${{ github.run_id }} --repo ${{ github.repository }} --name ttmlir-whl-tracy
+        gh run download ${{ github.run_id }} --repo ${{ github.repository }} --name ttnn-jit-whl-tracy
+
+        pip show ttmlir &> /dev/null && pip uninstall -y ttmlir
+        pip show ttnn-jit &> /dev/null && pip uninstall -y ttnn-jit
+        pip install ttnn_jit*.whl --find-links . --upgrade
```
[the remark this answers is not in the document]
sgholamiTT: Modified the whole structure. Not applicable anymore.

test/ttnn-jit/perf_ci/perf_tests.py
```diff
@@ -0,0 +1,101 @@
...
+    [
+        True,
+        False,
+    ],
+)
+def test_op_compare(
+    h, w, op, dtype, ttnn_dtype, memory_config, memory_config_id, jit_enabled
+):
+    device = ttnn.open_device(device_id=0)
+    torch_tensor_a = torch.rand((h, w), dtype=dtype) * 100
```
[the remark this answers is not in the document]
sgholamiTT: Added a `perf_device` parameter to the test.

.github/workflows/call-jit-perf-test.yml
sgholamiTT: You're right — there's no strong reason for a separate workflow. The original motivation was that perf collection depended on tracy profiler tools (`capture-release`, `csvexport-release`), which required custom setup steps. We've since reworked the approach to use the in-memory APIs instead, which eliminates the tracy dependency entirely.
With that change, the perf test fits the standard pattern: I just added the test invocation to `ttnn_jit.sh` if `Nightly`.

.github/workflows/call-jit-perf-test.yml
sgholamiTT: Not applicable anymore as this file is removed.

test/ttnn-jit/perf_ci/run_perf_collect.sh
```diff
@@ -0,0 +1,58 @@
...
+  echo "Summary reports written to $OUT_DIR"
+else
```
nsumrakTT, line 61: You'll probably need to upload results. You can use:
```
echo "{\"name\":\"perf-reports-$JOB_ID\",\"path\":\"$OUT_DIR"}," >> $UPLOAD_LIST
```

Also, you can add JOB_ID and/or PERF_REPORT_PATH here so they are available in test script:
https://github.com/tenstorrent/tt-mlir/blob/222c28128062fcc90e5d66d2b11d549ac5c3e1c6/.github/scripts/python/run_tests_from_json.py#L47

test/ttnn-jit/perf_ci/summarize_perf_results.py
```diff
@@ -0,0 +1,302 @@
...
+            "model": op,
+            "model_type": "jit_vs_ttnn",
+            "run_type": "op_benchmark",
+            "precision": dtype,
+            "config": config,
+            "measurements": measurements,
+        }
+
+        filename = f"perf_{op}_{dtype}_{mem_cfg}{job_suffix}.json"
+        filepath = out_dir / filename
```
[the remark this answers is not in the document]
sgholamiTT: The report filename now includes shape: `perf_{op}_{shape}_{dtype}_{mem_cfg}.json` (e.g., `perf_abs_2048x2048_BFLOAT16_dram_interleaved.json`). This prevents silent overwrites if additional shapes are added later.

test/ttnn-jit/perf_ci/run_perf_collect.sh
sgholamiTT: `run_perf_collect.sh` now writes JSON reports to a `reports/` subdirectory under the `OUT_DIR `. The profiler's raw data (`perf_results.json`, any device logs) stays in the top-level output dir, while only the clean Superset-ready JSON files go into `reports/`. This means CI artifact uploads can target just the `reports/` subdirectory without picking up intermediate files.

.github/workflows/call-jit-perf-test.yml
sgholamiTT: You're right — there's no strong reason for a separate workflow. The original motivation was that perf collection depended on tracy profiler tools (`capture-release`, `csvexport-release`), which required custom setup steps. We've since reworked the approach to use the in-memory APIs instead, which eliminates the tracy dependency entirely.
With that change, the perf test fits the standard pattern: I just added the test invocation to `ttnn_jit.sh` if `Nightly`.

sgholamiTT:
@nsumrakTT @vvukomanTT 
I made the last changes according to your comments. Can you please take a look ? 

Note: the changes in `.github/settings/tests.json` are just formatting that was needed by pre-commit hooks. No functional changes there.

sgholamiTT:
@vtangTT Not sure about the metal perf pipeline, but as far as I know we should generate json reports similar to [this](https://github.com/tenstorrent/tt-xla/blob/e84fbb8d771af7f6565b2e1216d5176fa19f7848/tests/benchmark/utils.py#L512-L535) to be able to upload to Superset with minimal setup.
Most probably possible to make some changes in Superset to accept other types of input data but that requires jumping through hoops.

PR #7504, opened 2026-03-16
Add SHLO Composite op support for TopK

hshahTT:
### Ticket
https://github.com/tenstorrent/tt-xla/issues/3728

### Problem description
Currently we just have support for fusing TTIR sort and slice ops into a TTNN TopK op. https://github.com/tenstorrent/tt-xla/pull/3729 introduces SHLO composite op support for TopK from Torch. This allows the composite op to be lowered into a TTNN TopK op without fusing.

### What's changed
- Adding support for lowering the 3 different types of SHLO composite ops representing the TopK op:
    - `tenstorrent.topk_values`: when only the values tensor is used in the graph
    - `tenstorrent.topk_indices`: when only the indices tensor is used in the graph
    - `tenstorrent.topk`: when both the values and indices tensors are used in the graph
- Added TTNN Workarounds to enforce the input and output tensor requirements as described [here](https://docs.tenstorrent.com/tt-metal/latest/ttnn/ttnn/api/ttnn.topk.html):
    - Input tensor can only be bfloat16 or bfloat8.
    - Output values tensor must have the same datatype as the input tensor.
    - Output indices tensor must be ui16.

### Checklist

### DIFF
--- a/lib/Conversion/StableHLOToTTIR/StableHLOLegalizeCompositePass.cpp
+++ b/lib/Conversion/StableHLOToTTIR/StableHLOLegalizeCompositePass.cpp
@@ -16,11 +16,11 @@
 #include "mlir/IR/PatternMatch.h"
 #include "mlir/Pass/Pass.h"
 #include "mlir/Transforms/DialectConversion.h"
-#include "stablehlo/dialect/StablehloOps.h"
-
 #include "llvm/ADT/SmallSet.h"
 #include "llvm/ADT/StringRef.h"
-#include <shardy/dialect/sdy/ir/dialect.h>
+
+#include "shardy/dialect/sdy/ir/dialect.h"
+#include "stablehlo/dialect/StablehloOps.h"
 
 using namespace mlir;
 using namespace mlir::tt;
@@ -102,8 +102,8 @@ public:
     }
     if (adaptor.getOperands().size() != 1) {
       return rewriter.notifyMatchFailure(
-          srcOp,
-          "tenstorrent.topk composite op must have exactly one input operand.");
+          srcOp, srcOp.getName() +
+                     " composite op must have exactly one input operand.");
     }
 
     bool isTopKWithValues = srcOp.getName() == "tenstorrent.topk_values";
@@ -126,7 +126,7 @@ public:
                                          "op must have exactly one result.");
     }
 
-    SmallVector<RankedTensorType> resultTypes;
+    SmallVector<RankedTensorType, 2> resultTypes;
     if (isTopKWithBoth) {
       resultTypes = {
           mlir::cast<RankedTensorType>(srcOp.getResult(0).getType()),
@@ -144,25 +144,20 @@ public:
 
     if (compositeAttrs) {
       if (auto attr = compositeAttrs.getAs<IntegerAttr>("k")) {
-        APInt kValue = attr.getValue();
-        if (!kValue.isIntN(32)) {
-          return rewriter.notifyMatchFailure(srcOp,
-                                             "k value is too large for ui32: " +
-                                                 Twine(kValue.getSExtValue()));
+        int64_t val = attr.getInt();
+        if (!llvm::isInt<32>(val)) {
+          return rewriter.notifyMatchFailure(
+              srcOp, "k value is too large for i32: " + Twine(val));
         }
-        kAttr = IntegerAttr::get(rewriter.getIntegerType(32),
-                                 static_cast<int32_t>(kValue.getSExtValue()));
+        kAttr = rewriter.getI32IntegerAttr(static_cast<int32_t>(val));
       }
       if (auto attr = compositeAttrs.getAs<IntegerAttr>("dim")) {
-        APInt dimValue = attr.getValue();
-        if (!dimValue.isSignedIntN(32)) {
+        int64_t val = attr.getInt();
+        if (!llvm::isInt<32>(val)) {
           return rewriter.notifyMatchFailure(
-              srcOp, "dim value is too large for si32: " +
-                         Twine(dimValue.getSExtValue()));
+              srcOp, "dim value is too large for i32: " + Twine(val));
         }
-        dimAttr =
-            IntegerAttr::get(rewriter.getIntegerType(32),
-                             static_cast<int32_t>(dimValue.getSExtValue()));
+        dimAttr = rewriter.getI32IntegerAttr(static_cast<int32_t>(val));
       }
       if (auto attr = compositeAttrs.getAs<BoolAttr>("largest")) {
         largestAttr = attr;


include/ttmlir/Dialect/TTNN/IR/TTNNWorkaroundsPass.h
```diff
@@ -354,6 +355,15 @@ class TTNNOperandsWorkaroundsFactory {
   // Issue page: https://github.com/tenstorrent/tt-metal/issues/39128
   static TTNNOperandsWorkarounds
   createMoeExpertTokenRemapOpOperandsWorkarounds();
+
+  // Create workarounds for topk ops.
+  // Input must be BFloat16 or BFP_BFloat8.
+  // Output values must be same data type as input.
+  // Output indices must be uint16.
+  // Source for restrictions:
+  // https://docs.tenstorrent.com/tt-metal/latest/ttnn/ttnn/api/ttnn.topk.html
```
sdjordjevicTT: Instead, could you please create a metal issue to resolve the restriction and link that issue here?

hshahTT: Done

lib/Conversion/StableHLOToTTIR/StableHLOLegalizeCompositePass.cpp
```diff
@@ -77,6 +79,134 @@ class StableHLOToTTIRCompositeOpConversionPattern
...
+    if (compositeAttrs) {
+      if (auto attr = compositeAttrs.getAs<IntegerAttr>("k")) {
+        kAttr = IntegerAttr::get(
+            rewriter.getIntegerType(32),
+            static_cast<int32_t>(attr.getValue().getSExtValue()));
+      }
+      if (auto attr = compositeAttrs.getAs<IntegerAttr>("dim")) {
+        dimAttr = IntegerAttr::get(
+            rewriter.getIntegerType(32),
+            static_cast<int32_t>(attr.getValue().getSExtValue()));
```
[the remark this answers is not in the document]
hshahTT: Done

sshonTT (approved):

lib/Conversion/StableHLOToTTIR/StableHLOLegalizeCompositePass.cpp
```diff
@@ -18,6 +18,8 @@
 #include "mlir/Transforms/DialectConversion.h"
 #include "stablehlo/dialect/StablehloOps.h"
 
+#include "llvm/ADT/SmallSet.h"
```
vwellsTT, line 19: nit: these should be in same block as mlir headers above--not sure if the shlo one should be moved to separate or not, and I think below shardy one should use "" instead of <> and be in same block as SHLO too.  But at least move these ones to above block

lib/Conversion/StableHLOToTTIR/StableHLOLegalizeCompositePass.cpp
```diff
@@ -77,6 +79,145 @@ class StableHLOToTTIRCompositeOpConversionPattern
...
+          srcOp,
+          "tenstorrent.topk_values composite op must have exactly one result.");
+    }
+    if (isTopKWithIndices && srcOp->getNumResults() != 1) {
+      return rewriter.notifyMatchFailure(srcOp,
+                                         "tenstorrent.topk_indices composite "
+                                         "op must have exactly one result.");
+    }
+
+    SmallVector<RankedTensorType> resultTypes;
```
vwellsTT: nit: if max size is 2, seems like we should use size hint here (or std::array 2 or similar)

lib/Conversion/StableHLOToTTIR/StableHLOLegalizeCompositePass.cpp
```diff
@@ -77,6 +79,145 @@ class StableHLOToTTIRCompositeOpConversionPattern
...
+  matchAndRewrite(mlir::stablehlo::CompositeOp srcOp,
+                  mlir::stablehlo::CompositeOp::Adaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    if (!supportedOpNames.contains(srcOp.getName())) {
+      return failure();
+    }
+    if (adaptor.getOperands().size() != 1) {
+      return rewriter.notifyMatchFailure(
+          srcOp,
+          "tenstorrent.topk composite op must have exactly one input operand.");
```
vwellsTT: maybe better to use srcOp.getName() here to be more accurate about which op is erroring if it's actually values/indexes?

lib/Conversion/StableHLOToTTIR/StableHLOLegalizeCompositePass.cpp
```diff
@@ -77,6 +79,145 @@ class StableHLOToTTIRCompositeOpConversionPattern
...
+        if (!kValue.isIntN(32)) {
+          return rewriter.notifyMatchFailure(srcOp,
+                                             "k value is too large for ui32: " +
+                                                 Twine(kValue.getSExtValue()));
+        }
+        kAttr = IntegerAttr::get(rewriter.getIntegerType(32),
+                                 static_cast<int32_t>(kValue.getSExtValue()));
+      }
+      if (auto attr = compositeAttrs.getAs<IntegerAttr>("dim")) {
+        APInt dimValue = attr.getValue();
```
vwellsTT: might be able to rewrite this a bit cleaner like:
```
 if (auto attr = compositeAttrs.getAs<IntegerAttr>("dim")) {
    int64_t v = attr.getInt();
    if (!llvm::isInt<32>(v))
          srcOp, Twine("dim value is too large for si32: ") + Twine(v));
    dimAttr = rewriter.getI32IntegerAttr(static_cast<int32_t>(v));
  }
}
```
and same for above

lib/Conversion/StableHLOToTTIR/StableHLOLegalizeCompositePass.cpp
hshahTT: Done

lib/Conversion/StableHLOToTTIR/StableHLOLegalizeCompositePass.cpp
hshahTT: Done

lib/Conversion/StableHLOToTTIR/StableHLOLegalizeCompositePass.cpp
hshahTT: Good catch, updated

lib/Conversion/StableHLOToTTIR/StableHLOLegalizeCompositePass.cpp
hshahTT: Done

PR #7512, opened 2026-03-16
[D2M] Use CTArgs for CB Indexes

vwellsTT:
### Ticket
NA

### Problem description
As part of CB rework, we changed the order-dependence of CBs vs program arguments.  As a workaround, we used a static hardcoded order, but this is probably less ideal in general, and problematic for d2m.spatial ops (which might run multiple generics at once, requiring unique CB port assignments across multiple generics).

### What's changed
Go back to using explicit CT args for the CB port indexes for both ttnn-jit and d2m paths.

### Checklist


hkwonTT (approved):
Thank you!

include/ttmlir/Dialect/TTKernel/IR/TTKernelOps.td
```diff
@@ -3054,20 +3054,20 @@ def TTKernel_GetCompileArgValOp : TTKernel_Op<"get_compile_time_arg_val",
...
-      code-generation time rather than going through the compile-time
-      argument indirection.
+      Produces a CB value by reading the port from compile-time arguments
+      via `get_compile_time_arg_val(ct_arg_index)`.  This allows the
+      spatial op to remap CB ports per grid range by overriding
+      compile-time arguments.
     }];
 
-    let arguments = (ins I32Attr:$port);
+    let arguments = (ins I32Attr:$ct_arg_index);
```
nsmithtt: This op is meant to give you access to a CB type directly from the compiler, I think some downstream users might use this op so we shouldn't change its behavior.  We should use `get_compile_time_arg_val` like we were previously.

nsmithtt: Maybe I'm confusing issues here, @tapspatel showed me some IR the other day where we were lowering the CB to just be a cast to the enum type from int, is this tackling a different issue with CB lowering?

Aside from fixing the situations that result in a hard-coded cast to a `get_compile_time_arg_val` I'm not sure I see the issue.  Taps argument passing refactor will also obviate any new convention established here.

include/ttmlir/Dialect/TTKernel/IR/TTKernelOpsTypes.td
```diff
@@ -134,20 +134,33 @@ def TTKernel_ArgAttr : TTKernel_Attr<"Arg", "arg"> {
   }];
   let parameters = (ins "ArgType":$arg_type, "size_t":$operand_index,
                         DefaultValuedParameter<"bool", "true">:$is_uniform,
-                        DefaultValuedParameter<"StringAttr", "StringAttr::get($_ctxt, \"\")">:$argument_name);
+                        DefaultValuedParameter<"StringAttr", "StringAttr::get($_ctxt, \"\")">:$argument_name,
+                        DefaultValuedParameter<"int64_t", "-1">:$port);
```
nsmithtt: Do we still need the port as part of this attr? Ideally we can remove this too.  The way this was working is that the compiler doesn't do any CB port assignment, it's up to the runtime, instead the compiler will write an arg spec that requests from the runtime to give it the CB port for a given operand index.

lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
```diff
@@ -1915,21 +1915,26 @@ class D2MGetCBRewriter : public OpConversionPattern<d2m::GetCBOp> {
 
     int64_t port = op.getPort();
     // The operand_index records which generic op operand this CB backs;
-    // the port is the actual hardware CB port number.
+    // the port is the actual hardware CB port number (may differ for spatial).
     int64_t operandIndex = op.getOperandIndex().value_or(port);
```
nsmithtt: Should we assert that there is an operand index?  It's unclear if falling back to `port` is actually safe.

include/ttmlir/Dialect/TTKernel/IR/TTKernelOpsTypes.td
vwellsTT: hmm, let me double check.  I remember there was some issue with old order-based approach for TTNN-jit specifically, but perhaps there's a better soln than specifying here

lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
vwellsTT: yeah, makes sense

vwellsTT: of course, in the future if we have scratch CBs or temporaries I think we might have no operand idx for certain CBs.  But I guess that should be cleaned up later.

nsmithtt (approved):

PR #7513, opened 2026-03-16
[D2M] [WIP]use proper stream dst for stream_layout's cb

vwellsTT:
### Ticket
NA

### Problem description
In flatbuffer serialization for ttmetal, the new handling of streamLayoutOps is not correct--we insert the input to the stream as a placeholder "cb" and then overwrite it later.  This is not logically correct and is causing some problems in Brett's branch.

### What's changed
We ought to first map the hoisted allocs to their operands, and then when processing operands look up any hoisted allocs and use those as CB's for the non-aliased case (and use the operands themselves as CBs for the aliased case).  There seems to be a couple of bugs in Allocate.cpp with inserting streams properly as well.

### Checklist

### DIFF
--- a/lib/Dialect/D2M/Transforms/LowerLoadStoreOpsToExplicitCBForm.cpp
+++ b/lib/Dialect/D2M/Transforms/LowerLoadStoreOpsToExplicitCBForm.cpp
@@ -4,15 +4,12 @@
 
 #include "ttmlir/Dialect/D2M/Transforms/Passes.h"
 
-#include "ttmlir/Asserts.h"
+#include "mlir/Dialect/MemRef/IR/MemRef.h"
+#include "mlir/IR/PatternMatch.h"
 #include "ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.h"
 #include "ttmlir/Dialect/D2M/IR/D2MOps.h"
 #include "ttmlir/Dialect/D2M/Utils/CBUtils.h"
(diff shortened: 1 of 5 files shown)


lib/Dialect/D2M/Transforms/LowerLoadStoreOpsToExplicitCBForm.cpp
```diff
@@ -60,141 +60,95 @@ static ReserveOp findReserveOp(Value value) {
...
+// the unique store user that can be folded into the forward-able pair path.
+static LogicalResult getLegalForwardableStore(RemoteLoadOp remoteLoad,
+                                              Value localBuffer,
+                                              RemoteStoreOp &forwardableStore) {
+  forwardableStore = nullptr;
+  if (!localBuffer) {
+    return success();
+  }
+
+  forwardableStore = utils::findForwardableStore(remoteLoad);
```
vmilosevic, line 121: :warning: **clang\-diagnostic\-error** :warning:
no member named `` findForwardableStore `` in namespace `` mlir::tt::d2m::utils ``

vmilosevic (commented):

vwellsTT:
Difficult to distill working set of changes without @bgrady-tt 's changes which I was debugging for this, possibly best to just close my PR and fold these changes into that one?  Alternatively I could revert last commit and merge my changes to how we handle the cb assignments--this would be correct on its own, but not sufficient to fix problems due to cases where we improperly don't insert streams

Issue #7158, opened 2026-02-23
Refactor D2MToTTNN Pass


arminaleTT:
This pass has significantly grown in complexity and length. Further, there is a complex ordering relationship between the rewrite patterns. A clean rewrite should make the pass simpler and easier to understand.

The main purpose of this pass is to:
1. Create and configure a `ttnn.generic` for each `d2m.generic`
2. Create `ttnn.empty` ops for tensors that need to be allocated
3. Cleanup the IO modelling ops that are no longer needed after 1 & 2

The complexity comes from mapping different forms of `d2m.generic` operands to the correct `ttnn.generic` configuration. These operands can come from `stream_layout` ops, `ttnn_metal_layout_cast` ops, `view_layout` ops, or `memref.alloc` ops (new). Generating the correct configuration requires having both the TTNN and D2M representation for operand tensors

After the refactor, IR tests for this pass should be tightened

PR #7516, opened 2026-03-16
[D2M] D2MToTTNN Pass Refactor

arminaleTT:
### Ticket
Closes https://github.com/tenstorrent/tt-mlir/issues/7158

### Problem description
The D2MToTTNN conversion pass has grown organically as TTNN JIT functionality has expanded. This pass is responsible for:
1. Creating intermediate tensor allocations by rewriting `mermef.alloc` ops with memref types as `ttnn.empty`
2. Create output tensor allocations by rewriting `d2m.empty` and `d2m.full` ops as `ttnn.empty` and `ttnn.full`
3. Converting semaphore creation and reset ops from D2M to TTNN
4. Converting `d2m.generic` ops to `ttnn.generic` ops with the appropriate IO, CB, KernelDescriptor, and semaphores
  i. IO and CB discovery requires tracing inputs and handling streamed and aliased operands, from `func.func` args and inferred intermediates  
5. Cleaning up IR structures that are no longer needed after steps 1-4

The steps above must happen in order, specifically 4 needs to consume the results of 1 and 2. This is difficult to enforce in a dialect conversion pass and has led to weird code and frequent bugs as the D2M dialect has evolved

Further, the code in 4 has bloated from recent feature additions in TTNN JIT and D2M dialect without holistic refactoring.

### What's changed
**Suggestion to reviewers**: The first two commits add tests and move functions around without any logical changes. Review from the 3rd commit onwards to focus on the architectural and logical changes.

- Significantly improve test coverage of D2MToTTNN pass by adding new complicated lit tests. These tests specifically exercise various properties of TTNN Generic
- Change the D2MToTTNN pass from dialect conversion to a manual module walk, making the ordering dependency explicit. 
 - **The behaviour of the pass on existing IR is identical before and after the refactor**

The pass is now split into 4 stages:
### 1. Materialize TTNN tensors:
Accomplishes the same functionality as 1 & 2, with the only logic change being properly skiping allocs for hoisted CBs

### 2. Convert semaphores
Accomplishes the same functionality as 3. No logic changes

### 3. Convert generics
The `d2m.generic` to `ttnn.generic` conversion happens here.

1. Rewrite and separate IO and CB discovery
      a. IO discovery is now recursive and will look through any combination of views, casts, or streams, instead of assuming very specific combinations
      b. Output operands are specifically identified
      c. Tensors are categorized as `DRAM`, `L1Remote` (any L1 tensor with a view or stream op in its def chain), or `L1Local` (all remaining L1 tensors)
2. Significantly simplify CB aliasing logic, enabled by marking outputs and locality in the previous step
      a. CBs only alias `L1Local` tensors
3. Simplify handling of hoisted CBs
4. Maintain existing logic for SemaphoreDescriptors and KernelDescriptors

### 4. Cleanup and verify
Remove all allocs/deallocs, streams, views, casts, d2m emptys/fulls, and d2m semaphore ops. Previously, many of these ops were left to dangle quite deep into the pipeline

### Checklist
- Nightly run: https://github.com/tenstorrent/tt-mlir/actions/runs/23257550929

### DIFF
--- a/lib/Conversion/D2MToTTNN/D2MToTTNN.cpp
+++ b/lib/Conversion/D2MToTTNN/D2MToTTNN.cpp
@@ -286,11 +286,17 @@ createKernelDescriptors(Builder &builder, const ArrayAttr &threads,
   return kernelConfigs;
 }
 
+enum class OperandLocality {
+  L1Local,
+  L1Remote,
+  DRAM,
+};
+
 struct GenericOperandInfo {
   Value ioTensor;
   Value cbMemref;
   bool isOutput;
-  bool isDRAM;
+  OperandLocality locality;
   unsigned operandIdx;
 };
 
@@ -319,29 +325,9 @@ createCBDescriptors(Builder &builder,
         ttnn::KernelCBFormatAttr::get(ctx, i, dtype, pageSize);
 
     ttnn::KernelCBGlobalBufferAddressOfTensorAttr globalCBIndexOfTensor;
-
-    if (!info.isDRAM) {
-      if (info.isOutput) {
-        // L1 outputs are always aliased.
-        globalCBIndexOfTensor =
-            ttnn::KernelCBGlobalBufferAddressOfTensorAttr::get(ctx, i);
-      } else {
-        bool isLocalFuncArg =
-            mlir::dyn_cast_if_present<ttir::TTNNMetalLayoutCastOp>(
-                info.cbMemref.getDefiningOp());
-        bool isLocalIntermediate =
-            mlir::dyn_cast_if_present<memref::AllocOp>(
-                info.cbMemref.getDefiningOp()) &&
-            !mlir::isa_and_present<ttcore::CBLayoutAttr>(
-                cbMemref.getLayout()) &&
-            llvm::none_of(info.cbMemref.getUsers(), [](Operation *user) {
-              return mlir::isa<d2m::StreamLayoutOp>(user);
-            });
-        if (isLocalFuncArg || isLocalIntermediate) {
-          globalCBIndexOfTensor =
-              ttnn::KernelCBGlobalBufferAddressOfTensorAttr::get(ctx, i);
-        }
-      }
+    if (info.locality == OperandLocality::L1Local) {
+      globalCBIndexOfTensor =
+          ttnn::KernelCBGlobalBufferAddressOfTensorAttr::get(ctx, i);
     }
 
     cbDescriptors[i] = ttnn::KernelCBAttr::get(
@@ -630,22 +616,26 @@ static LogicalResult convertSemaphores(ModuleOp module,
   return success();
 }
 
-static Value findIOTensor(Value operand, DenseMap<Value, Value> &valueMapping) {
+static std::pair<Value, OperandLocality>
+findIOTensor(Value operand, DenseMap<Value, Value> &valueMapping,
+             OperandLocality currentLocality = OperandLocality::L1Local) {
   auto iter = valueMapping.find(operand);
   if (iter != valueMapping.end()) {
     auto tensorType = dyn_cast<RankedTensorType>(iter->second.getType());
     TT_assertv(tensorType, "expected mapped value to be a ranked tensor");
     TT_assertv(isa<ttnn::TTNNLayoutAttr>(tensorType.getEncoding()),
                "expected mapped value to be a TTNN tensor");
-    return iter->second;
+    return {iter->second, currentLocality};
   }
 
   auto *def = operand.getDefiningOp();
   if (auto stream = dyn_cast<d2m::StreamLayoutOp>(def)) {
-    return findIOTensor(stream.getInput(), valueMapping);
+    return findIOTensor(stream.getInput(), valueMapping,
+                        OperandLocality::L1Remote);
   }
   if (auto view = dyn_cast<d2m::ViewLayoutOp>(def)) {
-    return findIOTensor(view.getInput(), valueMapping);
+    return findIOTensor(view.getInput(), valueMapping,
+                        OperandLocality::L1Remote);
   }
   if (auto cast = dyn_cast<ttir::TTNNMetalLayoutCastOp>(def)) {
     // Input is already a TTNN tensor.
@@ -653,7 +643,7 @@ static Value findIOTensor(Value operand, DenseMap<Value, Value> &valueMapping) {
     TT_assertv(tensorType, "expected input to be a ranked tensor");
     TT_assertv(isa<ttnn::TTNNLayoutAttr>(tensorType.getEncoding()),
                "expected input to be a ranked tensor");
-    return cast.getInput();
+    return {cast.getInput(), currentLocality};
   }
 
   llvm_unreachable("unexpected operand def chain");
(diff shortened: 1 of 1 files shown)


lib/Conversion/D2MToTTNN/D2MToTTNN.cpp
```diff
@@ -154,795 +142,742 @@ convertMathFidelity(ttmetal::MathFidelity fidelity) {
...
-
-  LogicalResult
-  matchAndRewrite(d2m::StreamLayoutOp op, d2m::StreamLayoutOpAdaptor adaptor,
-                  ConversionPatternRewriter &rewriter) const final {
-    rewriter.replaceOp(op, adaptor.getInput());
-    return success();
-  };
-};
-} // namespace
+  auto def = operand.getDefiningOp();
```
vmilosevic: :warning: **llvm\-qualified\-auto** :warning:
`` auto def `` can be declared as `` auto *def ``
```suggestion
  auto *def = operand.getDefiningOp();
```

lib/Conversion/D2MToTTNN/D2MToTTNN.cpp
```diff
@@ -154,795 +142,742 @@ convertMathFidelity(ttmetal::MathFidelity fidelity) {
...
+
+  if (auto stream = dyn_cast<d2m::StreamLayoutOp>(def)) {
+    return stream.getStorage();
+  }
+  if (auto view = dyn_cast<d2m::ViewLayoutOp>(def)) {
+    if (isa_and_present<d2m::StreamLayoutOp>(view.getInput().getDefiningOp())) {
+      return view.getInput();
     } else {
-      return rewriter.notifyMatchFailure(op, "unsupported encoding type");
+      TT_assertv(isa<MemRefType>(operand.getType()),
```
vmilosevic: :warning: **llvm\-else\-after\-return** :warning:
```suggestion
    }       TT_assertv(isa<MemRefType>(operand.getType()),
```

lib/Conversion/D2MToTTNN/D2MToTTNN.cpp
```diff
@@ -154,795 +142,742 @@ convertMathFidelity(ttmetal::MathFidelity fidelity) {
...
+                 "expected operand to be a memref");
+      return operand;
     }
```
vmilosevic, line 667: :warning: **llvm\-else\-after\-return** :warning:
```suggestion
   
```

lib/Conversion/D2MToTTNN/D2MToTTNN.cpp
```diff
@@ -154,795 +142,742 @@ convertMathFidelity(ttmetal::MathFidelity fidelity) {
...
-      for (auto castOp : castsToReplace) {
-        rewriter.replaceOp(castOp, emptyOp.getResult());
-      }
+static LogicalResult cleanupAndVerify(ModuleOp module,
+                                      DenseMap<Value, Value> &valueMapping) {
+  // Replace all mapped values' uses with their TTNN equivalents.
+  // This handles d2m.empty/full results used directly by func.return.
+  for (auto &[oldVal, newVal] : valueMapping) {
+    oldVal.replaceAllUsesWith(newVal);
+  }
```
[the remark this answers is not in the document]
arminaleTT: This is intended behaviour. We are converting the types

lib/Conversion/D2MToTTNN/D2MToTTNN.cpp
```diff
@@ -154,795 +142,742 @@ convertMathFidelity(ttmetal::MathFidelity fidelity) {
...
-      if ((mlir::dyn_cast_if_present<ttir::TTNNMetalLayoutCastOp>(
-               cb.getDefiningOp()) ||
-           isAliasedOutput) &&
-          ttcore::getMemorySpace(cb_memref) !=
-              ttcore::MemorySpace::DeviceDRAM) {
+    ttnn::KernelCBGlobalBufferAddressOfTensorAttr globalCBIndexOfTensor;
 
+    if (ttcore::getMemorySpace(cbMemref) != ttcore::MemorySpace::DeviceDRAM) {
+      if (info.isOutput) {
+        // L1 outputs are always aliased.
```
bgrady-tt: fwiw I think we should either 

1. Fix this bug and just configure output as a stream (seperate operand and CB) if it is a stream
2. Use common code path in else branch below for outputs, but assert if output is from StreamLayout

If we have tests relying on using DRAM buffer addresses as local CBs, option #2 will fail. We really shouldn't have tests that rely on hacks like that; I would be happier if those tests were disabled.

lib/Conversion/D2MToTTNN/D2MToTTNN.cpp
```diff
@@ -154,795 +142,742 @@ convertMathFidelity(ttmetal::MathFidelity fidelity) {
...
-      if (auto inner2 =
-              inner.getInput().getDefiningOp<ttir::TTNNMetalLayoutCastOp>()) {
-        rewriter.replaceOp(op, inner2.getOperand());
-      }
+  if (auto stream = dyn_cast<d2m::StreamLayoutOp>(def)) {
+    return stream.getStorage();
+  }
+  if (auto view = dyn_cast<d2m::ViewLayoutOp>(def)) {
+    if (isa_and_present<d2m::StreamLayoutOp>(view.getInput().getDefiningOp())) {
+      return view.getInput();
```
bgrady-tt: I don't think this logic is correct; if there is a stream _anywhere_ in the use-def chain, its storage buffer must be the CB memref. We need to do better in verifier that we don't produce invalid IR like `stream -> view -> stream -> ...`.

bgrady-tt: But assuming valid IR, the only valid use-def chains for streams and views should never have multiple streams.

lib/Conversion/D2MToTTNN/D2MToTTNN.cpp
arminaleTT: ~I don't understand the comment. Was some text cut off? This is not a bug, output aliasing for L1 tensors is the expected behaviour in JIT. The if on line 322 ensures that we **never** alias DRAM tensors~

Nevermind, I see how we can end up aliasing incorrectly on memrefs from stream storage buffers. I am working on a fix

lib/Conversion/D2MToTTNN/D2MToTTNN.cpp
arminaleTT: Fixed!

lib/Conversion/D2MToTTNN/D2MToTTNN.cpp
arminaleTT: I have modified the upstream to mark DRAM operands based on the tensor layout so we avoid having to parse the CB def chain and reason about hoisted vs unhoisted CBs. As a result, we can categorically exclude DRAM operands from aliasing.
@hkwonTT also shared some IR that failed as a result of this issue. I have converted that to a lit test after confirming that it works

lib/Conversion/D2MToTTNN/D2MToTTNN.cpp
```diff
@@ -154,795 +142,746 @@ convertMathFidelity(ttmetal::MathFidelity fidelity) {
...
-                                               dtype, layout, memcfg);
-    return success();
-  };
-};
-} // namespace
+  for (auto &info : infos) {
+    auto ttnnTensor = mlir::cast<RankedTensorType>(info.ioTensor.getType());
+    auto ttnnLayout =
+        mlir::cast<ttnn::TTNNLayoutAttr>(ttnnTensor.getEncoding());
+    info.isDRAM = ttnnLayout.getBufferType() == ttnn::BufferType::DRAM;
```
arminaleTT: @bgrady-tt here is where we mark the `isDRAM` field checked in CB descriptor generation. Since we're reading the ground truth tensor, we don't have to do a def chain traversal

lib/Conversion/D2MToTTNN/D2MToTTNN.cpp
```diff
@@ -154,795 +142,742 @@ convertMathFidelity(ttmetal::MathFidelity fidelity) {
...
+  // Hoisted CBs do not need to be materialized as TTNN tensors.
+  if (auto cbLayout = mlir::dyn_cast_if_present<ttcore::CBLayoutAttr>(
+          memrefType.getLayout())) {
+    return success();
+  }
+
+  auto deviceAttr = ttcore::lookupDevice(op);
+  if (!deviceAttr) {
+    return op.emitOpError("could not find device attribute");
+  }
```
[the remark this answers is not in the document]
bgrady-tt: I don't think this is a practical problem, but if the suggested patch runs might as well do it.

lib/Conversion/D2MToTTNN/D2MToTTNN.cpp
bgrady-tt: I think this is more correct than before. However we should be determining aliasing mode more robustly for L1 outputs. I won't block PR on this (I think we completely avoid L1 streaming in grid selection iirc) but I would like to see output streaming support finished for all backends in future.

bgrady-tt: There should be an assert here (or elsewhere idk) that L1 memspace outputs _cannot_ be stream_layout for TTNN backend. Misconfiguring the CB by assuming it must be in L1 alias mode will lead to PCC wild goose chases for someone down the line.

arminaleTT: This has been addressed in the latest commit. Any L1 tensor with a `view` or `stream` in its def chain will be considered `L1Remote`. All such tensors, in addition to DRAM tensors, will be excluded from CB aliasing

PR #7517, opened 2026-03-16
Optimal sharding selection 

apandeyTT:
### Ticket
https://docs.google.com/document/d/1WKySS7MIuL-Aej4H6lpXJmHQba6P0P_w43lueGLXzZA/edit?tab=t.0

### Problem description
Multi-device StableHLO graphs require sharding annotations on function arguments to distribute computation across devices. Currently, these shardings must be manually specified by the framework (e.g., via `shard_spec_fn` in JAX/torch-xla). There is no automated way within the tt-mlir compiler to explore and select an optimal sharding configuration that minimizes collective communication (CCL) overhead while maximizing memory efficiency.

### What's changed
Adds an `AutoShardingPass` that automatically searches for the optimal sharding configuration by enumerating candidates and evaluating their cost.

**AutoShardingPass (`AutoSharding.cpp`):**
- Enumerates all valid per-argument sharding configurations (at most one dimension per tensor sharded along the mesh axis).
- Optionally inserts `sdy.sharding_constraint` on intermediate op results, cross-producting.
- **Open/Closed toggle**: Duplicates each config into open (Shardy can adjust) and closed (exact) variants.
- For each candidate, clones the module, applies sharding hints, runs the remaining StableHLO sub-pipeline (propagation, reshard-to-collectives, canonicalization), and evaluates a cost function.
- **Cost function**: Volume-aware CCL cost (each op weighted by type and scaled by communicated-elements / max-elements) minus a memory benefit (sharding large tensors saves device memory, with a 3x multiplier for weight/parameter tensors over activations).
- Selects the lowest net-cost configuration and applies it to the original module.
- Optionally dumps per-variant IR, winner graphs, and a summary table to disk.

**Pipeline integration:**
- Registered as `--auto-sharding` pass in `Passes.td` with options: `dump-variants`, `dump-dir`, `max-constraint-candidates`.
- Gated behind `enable-auto-sharding` pipeline option; runs after `AnalyzeMesh` and before `DecoupleConstFanout`.

**Bug fix in `ShardyUtils.cpp`:**
- `addDictionaryAttrSdyShardingAnnotation` now replaces existing `sdy.sharding` attributes instead of appending duplicates, preventing a `DictionaryAttr element names must be unique` crash. (not sure this is the best way?)

**Perf comparison script (`tools/scripts/collect_perf_results.py`):**
- End-to-end script that runs auto-sharding search, compiles both manual and auto-sharded variants through the full pipeline (StableHLO → TTIR → TTNN → flatbuffer), executes `ttrt perf` on device, and generates a side-by-side performance comparison.
- Preprocessing step strips `ttcore.runtime_tensor_sharding`, `ttcore.shard_status`, and `tt.mark_argument` custom calls to handle re-running the pipeline on winner MLIR.
- All artifacts are written to `generated/` (gitignored).

**Reference test data (`test/ttmlir/Silicon/StableHLO/n300/auto_sharding/`):**
- Minimal attention MLIR inputs (manual-sharded and clean variants) for perf comparison.
- Directory marked `config.unsupported = True` so files are not executed in CI.

**Results on minimal attention (N300, mesh 1x2):** 
To get all the variants (update the path to your tt-mlir directory)
```shell
ttmlir-opt --stablehlo-pipeline="mesh-shape=1,2 enable-auto-sharding=true dump-dir=/tt-mlir/generated/auto_sharding/sdy_minimal_attn" /tt-mlir/test/ttmlir/Silicon/StableHLO/n300/auto_sharding/sdy_minimal_attn_auto.mlir
```
Single command to run end to end (permute all the variants and decide a winner and compare with manually sharded graphs)
```shell
python tools/scripts/collect_perf_results.py \
  --manual-input test/ttmlir/Silicon/StableHLO/n300/auto_sharding/sdy_minimal_attn_manual.mlir \
  --auto-input test/ttmlir/Silicon/StableHLO/n300/auto_sharding/sdy_minimal_attn_auto.mlir \
  --dump-variants
``` 

- Manual: 570,099 ns (32 ops)
- Auto:   504,085 ns (30 ops)
- **Auto-sharding is 1.131x faster (+11.6%)**

### Checklist


vmilosevic (commented):

lib/Dialect/StableHLO/Transforms/AutoSharding.cpp
```diff
@@ -0,0 +1,877 @@
...
+}
+
+// Remove stablehlo.custom_call @tt.mark_argument ops from a module.
+// These tt-xla-specific identity ops block Shardy from propagating shardings
+// through the graph, preventing CCL insertion during cost evaluation.
+// Each call is replaced by forwarding its input directly to all users.
+static void stripMarkArgumentCalls(ModuleOp module) {
+  SmallVector<mlir::stablehlo::CustomCallOp> toErase;
+  module.walk([&](mlir::stablehlo::CustomCallOp callOp) {
+    if (callOp.getCallTargetName() != "tt.mark_argument")
```
vmilosevic: :warning: **google\-readability\-braces\-around\-statements** :warning:
```suggestion
    if (callOp.getCallTargetName() != "tt.mark_argument") {
```

lib/Dialect/StableHLO/Transforms/AutoSharding.cpp
```diff
@@ -0,0 +1,877 @@
...
+      return;
+    if (callOp.getNumOperands() != 1 || callOp.getNumResults() != 1)
```
vmilosevic: :warning: **google\-readability\-braces\-around\-statements** :warning:
```suggestion
}
    if (callOp.getNumOperands() != 1 || callOp.getNumResults() != 1)
```

vmilosevic: :warning: **google\-readability\-braces\-around\-statements** :warning:
```suggestion
    if (callOp.getNumOperands() != 1 || callOp.getNumResults() != 1) {
```

lib/Dialect/StableHLO/Transforms/AutoSharding.cpp
```diff
@@ -0,0 +1,877 @@
...
+      return;
+    if (callOp.getOperand(0).getType() != callOp.getResult(0).getType())
```
vmilosevic: :warning: **google\-readability\-braces\-around\-statements** :warning:
```suggestion
}
    if (callOp.getOperand(0).getType() != callOp.getResult(0).getType())
```

vmilosevic: :warning: **google\-readability\-braces\-around\-statements** :warning:
```suggestion
    if (callOp.getOperand(0).getType() != callOp.getResult(0).getType()) {
```

lib/Dialect/StableHLO/Transforms/AutoSharding.cpp
```diff
@@ -0,0 +1,877 @@
...
+      return;
+    callOp.getResult(0).replaceAllUsesWith(callOp.getOperand(0));
```
vmilosevic, line 271: :warning: **google\-readability\-braces\-around\-statements** :warning:
```suggestion
}
    callOp.getResult(0).replaceAllUsesWith(callOp.getOperand(0));
```

lib/Dialect/StableHLO/Transforms/AutoSharding.cpp
```diff
@@ -0,0 +1,877 @@
...
+    toErase.push_back(callOp);
+  });
+  for (auto callOp : toErase)
```
vmilosevic: :warning: **google\-readability\-braces\-around\-statements** :warning:
```suggestion
  for (auto callOp : toErase) {
```

lib/Dialect/StableHLO/Transforms/AutoSharding.cpp
```diff
@@ -0,0 +1,877 @@
...
+    callOp->erase();
+}
```
vmilosevic, line 277: :warning: **google\-readability\-braces\-around\-statements** :warning:
```suggestion
}
}
```

lib/Dialect/StableHLO/Transforms/AutoSharding.cpp
```diff
@@ -0,0 +1,849 @@
...
+  // --- Tier 1: set sdy.sharding on function arguments ---
+  constexpr bool isClosed = false;
+  for (size_t argIdx = 0; argIdx < config.argDimSharded.size(); ++argIdx) {
+    llvm::SmallVector<mlir::sdy::DimensionShardingAttr> dimShardings;
+
+    for (bool sharded : config.argDimSharded[argIdx]) {
+      if (sharded) {
+        auto axisRef = mlir::sdy::AxisRefAttr::get(context, shardAxisName);
+        dimShardings.push_back(mlir::sdy::DimensionShardingAttr::get(
+            context, {axisRef}, isClosed));
```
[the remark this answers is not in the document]
apandeyTT: Closed sharings were removed, will update the description

include/ttmlir/Dialect/StableHLO/Transforms/Passes.td
```diff
@@ -248,6 +248,42 @@ def AnalyzeMeshPass : Pass<"analyze-mesh", "::mlir::ModuleOp">
   ];
 }
 
+def AutoShardingPass : Pass<"auto-sharding", "::mlir::ModuleOp"> {
+  let summary = "Automatically search for optimal sharding configuration by evaluating CCL costs";
+  let description = [{
```
anuragsingh-tt, line 253: Can we provide some minimal sample IR before and after this pass? See other pass descriptions for examples.

lib/Dialect/StableHLO/Transforms/AutoSharding.cpp
```diff
@@ -0,0 +1,849 @@
...
+#include "llvm/Support/raw_ostream.h"
+
+#include <chrono>
+#include <limits>
+
+namespace mlir::tt::stablehlo {
+#define GEN_PASS_DEF_AUTOSHARDINGPASS
+#include "ttmlir/Dialect/StableHLO/Transforms/Passes.h.inc"
+
+namespace {
```
anuragsingh-tt, line 63: Can we label this namespace with banners specifying helper definitions, helper classes and the pass implementation? 

E.g:
```
// Helper definitions.
// (ConstraintCandidate, ShardingConfig, MeshInfo, etc.)
```

It will make it easier for a reader to follow.

apandeyTT: Yes, will do!

include/ttmlir/Dialect/StableHLO/Transforms/Passes.td
apandeyTT: I have provided the examples for this 
And how to use them in the description of this PR
You can find them around here: [[test/ttmlir/Silicon/StableHLO/n300/auto_sharding/sdy_minimal_attn_auto.mlir](https://github.com/tenstorrent/tt-mlir/blob/b6b196c1e02a69543a98a9d02478e18d7a9a6831/test/ttmlir/Silicon/StableHLO/n300/auto_sharding/sdy_minimal_attn_auto.mlir)]

lib/Dialect/StableHLO/Transforms/AutoSharding.cpp
```diff
@@ -0,0 +1,849 @@
...
+    }
+    const MeshInfo &meshInfo = *meshInfoOpt;
+
+    auto shardableAxes = meshInfo.getShardableAxes();
+    if (shardableAxes.empty()) {
+      llvm::errs() << "AutoSharding: no shardable axes (all size 1), "
+                   << "applying all-replicated config\n";
+      return;
+    }
+
```
anuragsingh-tt, line 498: This is a great flow below. Can we make it more explicit? Seems to me we are analyzing -> formulating a problem/solving it -> executing (applyShardingHints). Maybe we can have these steps in their own functions?

lib/Dialect/StableHLO/Transforms/AutoSharding.cpp
```diff
@@ -0,0 +1,849 @@
...
+#include "mlir/Pass/PassManager.h"
+#include "mlir/Transforms/Passes.h"
+
+#include "llvm/ADT/ScopeExit.h"
+#include "llvm/Support/FileSystem.h"
+#include "llvm/Support/Path.h"
+#include "llvm/Support/raw_ostream.h"
+
+#include <chrono>
+#include <limits>
```
anuragsingh-tt, line 28: NIT: Are these used?

include/ttmlir/Dialect/StableHLO/Pipelines/StableHLOPipelines.h
```diff
@@ -39,6 +39,21 @@ struct StableHLOPipelineOptions
       *this, OptionNames::automaticArgAnalysis,
       llvm::cl::desc("Automatically determine argument shardings.")};
 
+  Option<bool> enableAutoSharding{
+      *this, OptionNames::enableAutoSharding,
+      llvm::cl::desc("Enable auto-sharding for optimal CCL cost."),
+      llvm::cl::init(false)};
+
+  Option<bool> dumpVariants{
+      *this, OptionNames::dumpVariants,
```
nsmithtt, line 52: These should be prefixed with autoSharding

include/ttmlir/Dialect/StableHLO/Transforms/Passes.td
anuragsingh-tt: Yes, I meant in this pass description specifically as other pass descriptions include it.

lib/Dialect/StableHLO/Transforms/AutoSharding.cpp
```diff
@@ -0,0 +1,809 @@
...
+
+    return analysis;
+  }
+
+  std::optional<SearchResult> evaluateConfigs(ModuleOp rootModule,
+                                              const AnalysisResult &analysis) {
+    MLIRContext *context = rootModule.getContext();
+    ShardingCostModel costModel;
+
+    double bestCost = INFINITY;
```
anuragsingh-tt: NIT: can we use C++ instead of C macro here

anuragsingh-tt (approved):
Thanks for incorporating the changes! Look forward to seeing how this cost model develops.

include/ttmlir/Dialect/StableHLO/Transforms/Passes.td
```diff
@@ -248,6 +248,42 @@ def AnalyzeMeshPass : Pass<"analyze-mesh", "::mlir::ModuleOp">
...
+    The winning sharding configuration is applied to the original module, which
+    then continues through the normal pipeline.
+  }];
+
+  let options = [
+    Option<"dumpVariants", "dump-variants", "bool",
+           /*default=*/"false",
+           "Dump each sharding variant IR to disk for inspection">,
+    Option<"dumpDir", "dump-dir", "std::string",
+           /*default=*/"\"\"",
```
tapspatel: maybe default to tmp

include/ttmlir/Dialect/StableHLO/Utils/ShardingCostModel.h
```diff
@@ -0,0 +1,75 @@
...
+#include "mlir/IR/BuiltinOps.h"
+
+#include "llvm/ADT/SmallVector.h"
+
+#include <cstdint>
+#include <optional>
+
+namespace mlir::tt::stablehlo {
+
+// Tier 2: intermediate op result eligible for sdy.sharding_constraint.
```
tapspatel: can you add a comment of how we know which op results are eligible for a constraint to be applied.

test/ttmlir/Silicon/StableHLO/n300/auto_sharding/lit.local.cfg
```diff
@@ -0,0 +1 @@
+config.unsupported = True
```
tapspatel: why do we not support these tests?

lib/Dialect/StableHLO/Transforms/AutoSharding.cpp
```diff
@@ -0,0 +1,809 @@
...
+    return false;
+  }
+  module->print(fos);
+  fos << "\n";
+  return true;
+}
+
+//===----------------------------------------------------------------------===//
+// AutoShardingPass implementation.
+//===----------------------------------------------------------------------===//
```
tapspatel: We should use `//` per the coding guidelines, not scoped comments
https://github.com/tenstorrent/tt-mlir/blob/cb06ccf45bed3605b91a880c4d71733ee3d3c47a/docs/src/coding-guidelines.md?plain=1#L66

lib/Dialect/StableHLO/Transforms/AutoSharding.cpp
```diff
@@ -0,0 +1,809 @@
...
+};
+
+static std::optional<MeshInfo> extractMeshInfo(ModuleOp &module) {
+  llvm::SmallVector<mlir::sdy::MeshOp> meshOps =
+      shardy_utils::getMeshOps(module);
+  if (meshOps.empty()) {
+    return std::nullopt;
+  }
+
+  MeshInfo info;
```
tapspatel: while this isn't incorrect, a good design is to create a constructed object if you have all the information available. Let's make a construct and pass through meshOps which will populate info.axes.

```
struct MeshInfo {
...
    MeshInfo(meshOps[0]) {
        this->axes = ...
    }
};
```

include/ttmlir/Dialect/StableHLO/Pipelines/StableHLOPipelines.h
```diff
@@ -43,6 +43,22 @@ struct StableHLOPipelineOptions
...
+  Option<bool> autoShardingDumpVariants{
+      *this, OptionNames::dumpVariants,
+      llvm::cl::desc("Dump each sharding variant IR to disk for inspection."),
+      llvm::cl::init(false)};
+
+  Option<std::string> autoShardingDumpDir{
+      *this, OptionNames::dumpDir,
+      llvm::cl::desc("Directory to dump variants into (default: current "
+                     "directory)."),
+      llvm::cl::init("")};
```
tapspatel: init to /tmp/

include/ttmlir/Dialect/StableHLO/Transforms/Passes.td
```diff
@@ -248,6 +248,38 @@ def AnalyzeMeshPass : Pass<"analyze-mesh", "::mlir::ModuleOp">
...
```
tapspatel: default to /tmp/

test/ttmlir/Silicon/StableHLO/n300/auto_sharding/lit.local.cfg
tapspatel: should this be here? all tests should be supported

include/ttmlir/Dialect/StableHLO/Utils/PassOverrides.h
```diff
@@ -17,6 +17,10 @@ struct OptionNames {
       "automatic-arg-analysis";
   static constexpr llvm::StringRef enableAggressiveSimplification =
       "enable-aggressive-simplification";
+  static constexpr llvm::StringRef enableAutoSharding = "enable-auto-sharding";
+  static constexpr llvm::StringRef systemDescPath = "system-desc-path";
```
tapspatel: where is `systemDescPath` being used?

include/ttmlir/Dialect/StableHLO/Utils/ShardingCostModel.h
```diff
@@ -0,0 +1,69 @@
+// SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
```
tapspatel: 2026

include/ttmlir/Dialect/StableHLO/Utils/ShardingCostModel.h
```diff
@@ -0,0 +1,69 @@
...
+#include "mlir/IR/BuiltinOps.h"
+
+#include "llvm/ADT/SmallVector.h"
+
+#include <cstdint>
+
+namespace mlir::tt::stablehlo {
+
+struct ShardingConfig {
+  // Per-arg per-dim sharding (true = sharded on the axis).
```
tapspatel: put an example here 
eg
```
%arg0 tensor<32x32xf32> sdy.sharding
relates to 
argDimSharded = [...]
```

include/ttmlir/Dialect/StableHLO/Utils/ShardingCostModel.h
```diff
@@ -0,0 +1,69 @@
...
+struct ShardingResult {
+  double communicationCost;
+  double memoryBenefit;
+  double netCost;
+};
+
+// Tunable parameters for ShardingCostModel.
+struct ShardingCostModelOptions {
+  // Fixed latency penalty per CCL op (synchronization + setup overhead).
+  // Unitless; relative to bandwidth cost which is scaled by volume fraction.
```
tapspatel: is this comment right? `// Unitless; relative to bandwidth cost which is scaled by volume fraction.` this base latency really means "we don't know the exact setups associated with executing ccls, but here is a guess since a graph with no ccls shouldn't incur this base latency"

include/ttmlir/Dialect/StableHLO/Utils/ShardingCostModel.h
```diff
@@ -0,0 +1,69 @@
...
+  double baseCCLLatency = 1.0;
```
tapspatel: probably want a comment to say this is a magic number or experimentally drive so future readers of this code know where this number comes from

include/ttmlir/Dialect/StableHLO/Utils/ShardingCostModel.h
```diff
@@ -0,0 +1,69 @@
...
+
+  // Multiplier for memory benefit of sharding weight/parameter tensors vs
+  // activations. Weights are persistent on-device and dominate peak memory,
+  // so sharding them is ~3x more valuable than sharding transient activations.
+  double parameterMultiplier = 3.0;
```
tapspatel: probably want a comment to say this is a magic number or experimentally drive so future readers of this code know where this number comes from

include/ttmlir/Dialect/StableHLO/Utils/ShardingCostModel.h
```diff
@@ -0,0 +1,69 @@
...
+};
+
+// Cost model for evaluating sharding configurations by estimating communication
+// overhead from CCL ops and memory benefit from tensor distribution across a
+// mesh. Designed to be swappable (e.g., with a lookup-table approach) without
+// changing the AutoSharding pass.
+class ShardingCostModel {
+public:
+  using Options = ShardingCostModelOptions;
```
tapspatel: can we just put the `baseCCLLatency` and `parameterMultiplier` within this class as member variables? So we don't have to maintain another struct for it

include/ttmlir/Dialect/StableHLO/Utils/ShardingCostModel.h
```diff
@@ -0,0 +1,69 @@
...
+  // activations. Weights are persistent on-device and dominate peak memory,
+  // so sharding them is ~3x more valuable than sharding transient activations.
+  double parameterMultiplier = 3.0;
+};
+
+// Cost model for evaluating sharding configurations by estimating communication
+// overhead from CCL ops and memory benefit from tensor distribution across a
+// mesh. Designed to be swappable (e.g., with a lookup-table approach) without
+// changing the AutoSharding pass.
+class ShardingCostModel {
```
tapspatel: im not actually sure we need a class in this situation. A class implies we want to store some state. However, I see we are only using the functions inside this class to encapsulate common functionality. We should modify this file to just have the following

```
struct ShardingResult {
  double communicationCost;
  double memoryBenefit;
  double netCost;
};

ShardingResult evaluate(ModuleOp module, llvm::SmallVector<llvm::SmallVector<bool>>& argDimSharded, func::FuncOp originalFuncOp, int64_t meshAxisSize);
``` 

A user provides a module, a shardingConfig, the original func and a mesh axis size and they get back the sharding results. That's our API.

include/ttmlir/Dialect/StableHLO/Utils/ShardingCostModel.h
```diff
@@ -0,0 +1,69 @@
...
+
+private:
+  Options options;
+
+  double evaluateCommunicationCost(ModuleOp module, int64_t maxElements) const;
+
+  double evaluateMemoryBenefit(const ShardingConfig &config,
+                               func::FuncOp funcOp, int64_t meshAxisSize,
+                               int64_t maxElements) const;
+};
```
tapspatel: let's remove the class overhead and stick to the single API I presented above.

lib/Dialect/StableHLO/Utils/ShardingCostModel.cpp
```diff
@@ -0,0 +1,139 @@
+// SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
```
tapspatel: 2026

lib/Dialect/StableHLO/Utils/ShardingCostModel.cpp
```diff
@@ -0,0 +1,139 @@
...
+
+#include "llvm/ADT/StringMap.h"
+
+namespace mlir::tt::stablehlo {
+
+// Relative cost weights for collective communication ops.
+// all_gather / collective_permute: 1.0 (single data movement, no reduction)
+// reduce_scatter / all_reduce / all_to_all: 1.5 (reduction + data movement)
+// all_slice: 0.5 (local slicing, minimal communication)
+static const llvm::StringMap<double> cclWeights = {
```
tapspatel: migrate this to shardingcostmode.h. Also do we need string identifiers as the key? Can we just use the actual type eg. `mlir::stablehlo::AllGatherOp`

lib/Dialect/StableHLO/Utils/ShardingCostModel.cpp
```diff
@@ -0,0 +1,139 @@
...
+    {"stablehlo.all_gather", 1.0},         {"stablehlo.reduce_scatter", 1.5},
+    {"stablehlo.all_reduce", 1.5},         {"stablehlo.all_to_all", 1.5},
+    {"stablehlo.collective_permute", 1.0},
+};
+
+static const llvm::StringMap<double> compositeWeights = {
```
tapspatel: migrate this to shardingcostmode.h. Same with converting from string identifier to using op type. Reason for this is mainly type checking, so when something upstream changes, we can identify the error during compilation instead of a runtime key lookup error. 

Also might want to add are these magic numbers or experimentally driven.

lib/Dialect/StableHLO/Utils/ShardingCostModel.cpp
tapspatel: Also might want to add are these magic numbers or experimentally driven.

lib/Dialect/StableHLO/Utils/ShardingCostModel.cpp
```diff
@@ -0,0 +1,139 @@
...
+    {"sdy.all_slice", 0.5},
+    {"sdy.all_gather", 1.0},
+    {"sdy.reduce_scatter", 1.5},
+    {"sdy.all_reduce", 1.5},
+};
+
+ShardingCostModel::ShardingCostModel() = default;
```
tapspatel: remove

lib/Dialect/StableHLO/Utils/ShardingCostModel.cpp
```diff
@@ -0,0 +1,139 @@
...
+
+ShardingCostModel::ShardingCostModel(Options options)
```
tapspatel: remove

lib/Dialect/StableHLO/Utils/ShardingCostModel.cpp
```diff
@@ -0,0 +1,139 @@
...
+    : options(std::move(options)) {}
+
+ShardingResult ShardingCostModel::evaluate(ModuleOp module,
```
tapspatel: Single API
```
ShardingResult evaluate(ModuleOp module, llvm::SmallVector<llvm::SmallVector<bool>>& argDimSharded, func::FuncOp originalFuncOp, int64_t meshAxisSize);
```

lib/Dialect/StableHLO/Utils/ShardingCostModel.cpp
```diff
@@ -0,0 +1,139 @@
...
+      maxElements = std::max(maxElements, tt.getNumElements());
+    }
+  }
+  return maxElements;
+}
+
+// Evaluate cost of a lowered StableHLO module by counting and weighting CCL
+// ops. Each CCL op incurs a fixed latency overhead (setup, synchronization)
+// plus a bandwidth cost proportional to the volume of data communicated.
+double ShardingCostModel::evaluateCommunicationCost(ModuleOp module,
```
tapspatel: do you see this as being called by a pass user in the future? if so, we can also make it an api and put its function declaration in `include/ttmlir/Dialect/StableHLO/Utils/ShardingCostModel.h`. Otherwise, lets make it static to this file only.

lib/Dialect/StableHLO/Utils/ShardingCostModel.cpp
```diff
@@ -0,0 +1,139 @@
...
+
+// Compute the memory benefit of a sharding config. Sharding large argument
+// tensors across the mesh saves memory proportional to the element count.
+// Benefit is normalized relative to the largest argument.
+//
+// Weight/parameter tensors (heuristic: rank <= 2 with > 1024 elements) get a
+// higher multiplier because they are persistent on device and dominate peak
+// memory in large models. Activations (rank > 2, typically with a batch dim)
+// are transient and less valuable to shard for memory.
+double ShardingCostModel::evaluateMemoryBenefit(const ShardingConfig &config,
```
tapspatel: do you see this as being called by a pass user in the future? if so, we can also make it an api and put its function declaration in include/ttmlir/Dialect/StableHLO/Utils/ShardingCostModel.h. Otherwise, lets make it static to this file only.

lib/Dialect/StableHLO/Utils/ShardingCostModel.cpp
```diff
@@ -0,0 +1,139 @@
...
+                                           func::FuncOp originalFuncOp,
+                                           int64_t meshAxisSize) const {
+  int64_t maxElements = computeMaxElements(originalFuncOp);
+  double commCost = evaluateCommunicationCost(module, maxElements);
+  double memBenefit =
+      evaluateMemoryBenefit(config, originalFuncOp, meshAxisSize, maxElements);
+  return {commCost, memBenefit, commCost - memBenefit};
+}
+
+int64_t ShardingCostModel::computeMaxElements(func::FuncOp funcOp) {
```
tapspatel: make this static function

lib/Dialect/StableHLO/Utils/ShardingCostModel.cpp
```diff
@@ -0,0 +1,139 @@
...
+
+// Evaluate cost of a lowered StableHLO module by counting and weighting CCL
+// ops. Each CCL op incurs a fixed latency overhead (setup, synchronization)
+// plus a bandwidth cost proportional to the volume of data communicated.
+double ShardingCostModel::evaluateCommunicationCost(ModuleOp module,
+                                                    int64_t maxElements) const {
+  double totalCost = 0.0;
+
+  module.walk([&](Operation *op) {
+    StringRef opName = op->getName().getStringRef();
```
tapspatel: this will need to change to lookup the op type

lib/Dialect/StableHLO/Utils/ShardingCostModel.cpp
```diff
@@ -0,0 +1,139 @@
...
+    double opWeight = 0.0;
+
+    auto it = cclWeights.find(opName);
```
tapspatel: on second thought, we should actually lookup from a switch statement

static double getCCLWeight(Operation* op) {
switch(op) {
case mlir::stablehlo::AllGatherOp:
return 1.0;
...
default
return 0.0;
}
}

lib/Dialect/StableHLO/Utils/ShardingCostModel.cpp
```diff
@@ -0,0 +1,139 @@
...
+    if (it != cclWeights.end()) {
+      opWeight = it->second;
+    } else if (opName == "stablehlo.composite") {
```
tapspatel: is this right? When we run `lib/Dialect/StableHLO/Transforms/UpdateGlobalToLocalShapes.cpp` we convert all shardy ccls `convertShardyCCLToStableHLOCCL` into stablehlo ccls. So we shouldn't have any sdy ccls left. Also stablehlo.composite is a different op itself (see `test/ttmlir/Conversion/StableHLOToTTIR/composite_op.mlir`). Actually, we need to iterate through the stablehlo composite op and check if any ccls got inserted there.

lib/Dialect/StableHLO/Utils/ShardingCostModel.cpp
```diff
@@ -0,0 +1,139 @@
...
+      if (auto nameAttr = op->getAttrOfType<StringAttr>("name")) {
+        auto cit = compositeWeights.find(nameAttr.getValue());
+        if (cit != compositeWeights.end()) {
+          opWeight = cit->second;
+        }
+      }
+    }
+
+    if (opWeight > 0.0 && maxElements > 0) {
```
tapspatel: personal nit: we should actually do this during the loop

```
 module.walk([&](Operation *op) {
if (!isOpCCL(op)) {
continue;
}

double opWeight = getCCLWeight(op);
...
}
```

The reason for this is opWeight = 0.0 is implicitly implying that a non-ccl op is being skipped form the totalCost calculation, but that is not evident from just reading the code. The above snippet makes it very clear to skip non ccls from the total cost.

tapspatel: also can max elements ever be less than 1? 

see: 
```
int64_t ShardingCostModel::computeMaxElements(func::FuncOp funcOp) {
  int64_t maxElements = 1;
  for (auto arg : funcOp.getArguments()) {
    if (auto tt = dyn_cast<RankedTensorType>(arg.getType())) {
      maxElements = std::max(maxElements, tt.getNumElements());
    }
  }
  return maxElements;
}
```

lib/Dialect/StableHLO/Utils/ShardingCostModel.cpp
```diff
@@ -0,0 +1,139 @@
...
+      int64_t commElements = 1;
+      if (op->getNumOperands() > 0) {
+        if (auto tt = dyn_cast<RankedTensorType>(op->getOperand(0).getType())) {
+          commElements = tt.getNumElements();
+        }
+      }
+      double volumeFactor =
+          static_cast<double>(commElements) / static_cast<double>(maxElements);
+      totalCost += options.baseCCLLatency + opWeight * volumeFactor;
```
tapspatel: baseCCLLatency can just be a variable in this function, no need to modularize into a separate structure.

tapspatel: add a comment to explain why this particular cost function works.

lib/Dialect/StableHLO/Utils/ShardingCostModel.cpp
```diff
@@ -0,0 +1,139 @@
...
+//
+// Weight/parameter tensors (heuristic: rank <= 2 with > 1024 elements) get a
+// higher multiplier because they are persistent on device and dominate peak
+// memory in large models. Activations (rank > 2, typically with a batch dim)
+// are transient and less valuable to shard for memory.
+double ShardingCostModel::evaluateMemoryBenefit(const ShardingConfig &config,
+                                                func::FuncOp funcOp,
+                                                int64_t meshAxisSize,
+                                                int64_t maxElements) const {
+  if (meshAxisSize <= 1) {
```
tapspatel, line 104: im not sure I fully understand the purpose of this check here. Can you explain why we can't remove this and let the algorithm compute a mesh axis of 1?

lib/Dialect/StableHLO/Utils/ShardingCostModel.cpp
```diff
@@ -0,0 +1,139 @@
...
+    return 0.0;
+  }
+
+  double benefit = 0.0;
+  double fractionSaved = 1.0 - 1.0 / static_cast<double>(meshAxisSize);
```
tapspatel, line 118: add a comment to explain why this function works

lib/Dialect/StableHLO/Utils/ShardingCostModel.cpp
```diff
@@ -0,0 +1,139 @@
...
+}
+
+// Compute the memory benefit of a sharding config. Sharding large argument
+// tensors across the mesh saves memory proportional to the element count.
+// Benefit is normalized relative to the largest argument.
+//
+// Weight/parameter tensors (heuristic: rank <= 2 with > 1024 elements) get a
+// higher multiplier because they are persistent on device and dominate peak
+// memory in large models. Activations (rank > 2, typically with a batch dim)
+// are transient and less valuable to shard for memory.
```
tapspatel: do these comments match the definition of the function?

lib/Dialect/StableHLO/Utils/ShardyUtils.cpp
```diff
@@ -287,14 +287,18 @@ mlir::DictionaryAttr addDictionaryAttrSdyShardingAnnotation(
     std::optional<mlir::DictionaryAttr> dictAttr) {
   llvm::SmallVector<mlir::NamedAttribute> newArgAttrs;
 
+  auto shardingKey =
+      mlir::StringAttr::get(context, mlir::sdy::TensorShardingAttr::name);
+
   if (dictAttr) {
-    newArgAttrs =
-        SmallVector<mlir::NamedAttribute>(dictAttr.value().getValue());
+    for (auto namedAttr : dictAttr.value().getValue()) {
```
tapspatel: we shouldn't make changes to this api. This api is called when we want to add a sdy.sharding to the dictionaryattr. It is up to the user who calls this API to ensure that there already isn't a sdy.sharding attribute in the dictionary. See `removeDictionaryAttrSdyShardingAnnotations` if you want to remove the sdy.sharding attribute.

lib/Dialect/StableHLO/Transforms/AutoSharding.cpp
```diff
@@ -0,0 +1,570 @@
+// SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
```
tapspatel: 2026

lib/Dialect/StableHLO/Transforms/AutoSharding.cpp
```diff
@@ -0,0 +1,570 @@
...
+  module->print(fos);
+  fos << "\n";
+  return true;
+}
+
+//===----------------------------------------------------------------------===//
+// AutoShardingPass implementation.
+//===----------------------------------------------------------------------===//
+
+class AutoShardingPass : public impl::AutoShardingPassBase<AutoShardingPass> {
```
tapspatel: take struct AnalysisResult,   struct VariantResult   struct SearchResult   std::optional<AnalysisResult> analyzeModule(ModuleOp rootModule) evaluateConfigs applyBestConfig and writeSummary outside of the class.

lib/Dialect/StableHLO/Utils/ShardyUtils.cpp
tapspatel: btw the reason for this is when a user calls an API, what we are doing here is an unwanted side effect, of removing an existing sdy sharding. The user needs to be aware of this and should handle the case if there is already an existing sdy. sharding annotation.

lib/Dialect/StableHLO/Transforms/AutoSharding.cpp
```diff
@@ -0,0 +1,570 @@
...
+
+//===----------------------------------------------------------------------===//
+// Helper data structures.
+//===----------------------------------------------------------------------===//
+
+struct MeshInfo {
+  std::string meshName;
+  llvm::SmallVector<std::pair<std::string, int64_t>> axes;
+
+  MeshInfo() = default;
```
tapspatel: if the user shouldn't be calling the default constructor, it shouldn't be added and removed from the class definition.

lib/Dialect/StableHLO/Transforms/AutoSharding.cpp
```diff
@@ -0,0 +1,570 @@
...
+
+  void runOnOperation() final {
+    ModuleOp rootModule = getOperation();
+    MLIRContext *context = rootModule.getContext();
+
+    context->exitMultiThreadedExecution();
+    auto restoreGuard = llvm::make_scope_exit(
+        [context] { context->enterMultiThreadedExecution(); });
+
+    auto analysisOpt = analyzeModule(rootModule);
```
tapspatel: some code reorg stuff. Usually what's a good flow is you want to check for the conditions of something to see if it's valid, before you run an algorithm on it. You end up calling 

```
auto meshInfoOpt = extractMeshInfo(rootModule);
  if (!meshInfoOpt) {
    rootModule.emitWarning("AutoSharding: no mesh found in module, skipping");
    return std::nullopt;
  }
```

which return std::null opt, which makes for not so great function design, especially if someone wants to trace this code. A better way is to 

```
if(isModuleValid(rootModule)) {
//exit
}

auto analysisOpt = analyzeModule(rootModule);
```

so you don't have to keep conditions or null opt. The less paths in your code that diverge, the better. The less information that needs to be sent around the better.

lib/Dialect/StableHLO/Transforms/AutoSharding.cpp
```diff
@@ -0,0 +1,570 @@
...
+  };
+
+  struct SearchResult {
+    size_t bestIdx;
+    double bestCost;
+    llvm::SmallVector<VariantResult> results;
+  };
+
+  std::optional<AnalysisResult> analyzeModule(ModuleOp rootModule) {
+    auto meshInfoOpt = extractMeshInfo(rootModule);
```
tapspatel: do we need a new structure for this? we already have `using MeshMap =` in ` include/ttmlir/Dialect/StableHLO/Utils/ShardyUtils.` which seems to model the information you already want.

lib/Dialect/StableHLO/Transforms/AutoSharding.cpp
```diff
@@ -0,0 +1,570 @@
...
+    analysis.meshInfo = *meshInfoOpt;
+
+    if (analysis.meshInfo.axes.empty()) {
+      llvm::errs() << "AutoSharding: no shardable axes (all size 1), "
+                   << "applying all-replicated config\n";
+      return std::nullopt;
+    }
+
+    analysis.shardAxisName = analysis.meshInfo.axes[0].first;
+    analysis.meshAxisSize = analysis.meshInfo.axes[0].second;
```
tapspatel: DRY: don't repeat yourself. If you already have `  MeshInfo meshInfo;` in AnalysisResult, you don't need to store variables again in the structure. Either your data structure is not well defined or you need to find a different access pattern.

lib/Dialect/StableHLO/Transforms/AutoSharding.cpp
```diff
@@ -0,0 +1,570 @@
...
+    llvm::errs() << "AutoSharding: mesh='" << analysis.meshInfo.meshName
```
tapspatel: wrong comment. this prints to stderr but there's no error about this.

lib/Dialect/StableHLO/Transforms/AutoSharding.cpp
```diff
@@ -0,0 +1,570 @@
...
+    if (!meshInfoOpt) {
+      rootModule.emitWarning("AutoSharding: no mesh found in module, skipping");
+      return std::nullopt;
+    }
+
+    AnalysisResult analysis;
+    analysis.meshInfo = *meshInfoOpt;
+
+    if (analysis.meshInfo.axes.empty()) {
+      llvm::errs() << "AutoSharding: no shardable axes (all size 1), "
```
tapspatel: this is not a llvm::errs

lib/Dialect/StableHLO/Transforms/AutoSharding.cpp
```diff
@@ -0,0 +1,570 @@
...
+    double cost;
+    double commCost;
+    double memBenefit;
+  };
+
+  struct SearchResult {
+    size_t bestIdx;
+    double bestCost;
+    llvm::SmallVector<VariantResult> results;
+  };
```
tapspatel: im not opposed to this structs but can you find a way to better represent the information? Someone coming into this sees all these variables and would have no idea what each of them mean or how the algorithm is using them.

tapspatel (changes requested):
did 75%. please make the required edits and rerequest again.

lib/Dialect/StableHLO/Utils/ShardingCostModel.cpp
apandeyTT: Yes that is right, updated accordingly

lib/Dialect/StableHLO/Transforms/AutoSharding.cpp
apandeyTT: Added comments to explain each component

lib/Dialect/StableHLO/Transforms/AutoSharding.cpp
apandeyTT: The duplicate member variables were removed and replaced with accessor methods (shardAxisName() and meshAxisSize()) that delegate directly to meshInfo.axes[0]. This way there's a single source of truth (meshInfo), and the accessors just provide a convenient shorthand so callers don't have to write analysis.meshInfo.axes[0].first everywhere.

lib/Dialect/StableHLO/Utils/ShardingCostModel.cpp
apandeyTT: Replaced the string-based map lookup with llvm::TypeSwitch on concrete op types in getCCLWeight(). Used TypeSwitch over a raw switch since it's the idiomatic MLIR pattern for type-based dispatch. Please let me know if I can do it differently.

lib/Dialect/StableHLO/Utils/ShardingCostModel.cpp
apandeyTT: Updated to 4

lib/Dialect/StableHLO/Utils/ShardingCostModel.cpp
apandeyTT: You're right, removing it wouldn't change the result. With meshAxisSize = 1, fractionSaved = 1 - 1/1 = 0, so every benefit term would be zero anyway. The check is just an early exit to skip the loop when we know the answer is zero. Added a comment to make the intent clear.

lib/Dialect/StableHLO/Transforms/AutoSharding.cpp
```diff
@@ -0,0 +1,631 @@
...
+
+  for (int64_t bits = 0; bits < numConfigs; ++bits) {
+    ArgDimSharded config;
+    int64_t bitIdx = 0;
+    bool valid = true;
+    for (size_t argIdx = 0; argIdx < argRanks.size(); ++argIdx) {
+      llvm::SmallVector<bool> dimChoices;
+      int shardedCount = 0;
+      for (int64_t d = 0; d < argRanks[argIdx]; ++d) {
+        bool sharded = (bits >> bitIdx) & 1;
```
sshonTT: Correct me if I'm wrong, but bitIdx keeps incrementing up to totalDims in the inner loop, the 20-bit cap only applies to numConfigs (the outer bits range), not to bitIdx.

Two issues:
1. If totalDims > 64, the last iterations evaluate bits >> bitIdx with bitIdx >= 64, which is UB for int64_t (shift amount ≥ operand width).
2. Even when totalDims <= 64 but > 20, bits is always < 2^20 by construction, so (bits >> bitIdx) & 1 is always 0 for bitIdx >= 20. Dims past the first 20 are silently pinned to replicated rather than being enumerated. The "2^20 bit patterns" log message suggests we explore 2^20 configurations, but we only explore them over the first 20 dims.'


Probably want to either clamp the shift explicitly (treat bitIdx >= cap as false) or enumerate per-arg more directly (each arg has rank + 1 valid choices, so the full valid search space is ∏(rank_i + 1), typically way smaller than 2^totalDims).

lib/Dialect/StableHLO/Utils/ShardingCostModel.cpp
```diff
@@ -0,0 +1,150 @@
...
+
+    double opWeight = getCCLWeight(op);
+
+    int64_t commElements = 1;
+    if (op->getNumOperands() > 0) {
+      if (auto tt = dyn_cast<RankedTensorType>(op->getOperand(0).getType())) {
+        commElements = tt.getNumElements();
+      }
+    }
+    double volumeFactor =
```
sshonTT, line 90: This volumeFactor could be greater than 1.0. For concrete example using attention,
Q: [Batch, NumHeads, Sequence, HeadDims]
K and V have the same shape as Q.
And the attention score, which is an intermediate tensor (not an input argument), has the shape [Batch, NumHeads, Sequence, Sequence].
If we shard on NumHeads dims for attention score, then 
volumFactor = (B * NH * S * S) / (B * NH * S * HD)
So if the sequence length is larger than head_dim then the volumeFactor would be greater than 1.0

lib/Dialect/StableHLO/Transforms/AutoSharding.cpp
apandeyTT: Yes, you're right, thank you for catching this! 
I'll switch to per-arg enumeration: since we enforce at-most-one-sharded-dim per tensor, the valid search space is ∏(rank_i + 1), which is typically orders of magnitude smaller than 2^totalDims. This eliminates both the UB and the silent truncation, and removes the need for the bit-pattern cap entirely.

lib/Dialect/StableHLO/Utils/ShardingCostModel.cpp
apandeyTT: Yes, you're right that volumeFactor can exceed 1.0 when CCL ops operate on intermediate tensors larger than any input argument (the attention score example is a concrete case).

However, I believe this doesn't affect correctness for the cost model's purpose, since all configs are compared using the same maxElements denominator, the relative ordering is preserved. A volumeFactor > 1.0 correctly reflects that communicating a larger intermediate tensor should incur a proportionally higher bandwidth cost.

That said, the comment calling it a "fraction" is misleading, I'll update the comment to clarify it's a ratio (not bounded to [0, 1]) and add a note about intermediate tensors potentially exceeding the normalization base.

lib/Dialect/StableHLO/Transforms/AutoSharding.cpp
```diff
@@ -0,0 +1,627 @@
...
+    }
+    argRanks.push_back(tensorType.getRank());
+  }
+
+  // Compute total number of valid configs: ∏(rank_i + 1).
+  // Each factor is (rank + 1) choices: shard on one dim, or replicate.
+  int64_t numConfigs = 1;
+  for (auto rank : argRanks) {
+    numConfigs *= (rank + 1);
+    if (numConfigs > (1LL << 20)) {
```
sshonTT, line 175: I am a bit worried about here. The search space feels small, in real model we have as many as 500+ arg models, then they would be capped by here. can we make the search space bigger? Maybe we can use something like Beam Search?

lib/Dialect/StableHLO/Utils/ShardingCostModel.cpp
```diff
@@ -0,0 +1,155 @@
...
+        llvm::any_of(argDimSharded[argIdx], [](bool s) { return s; });
+    if (!isSharded) {
+      continue;
+    }
+
+    auto tensorType =
+        cast<RankedTensorType>(funcOp.getArgument(argIdx).getType());
+    int64_t numElements = tensorType.getNumElements();
+
+    // Heuristic: tensors with rank <= 4 and > 1024 elements are likely model
```
sshonTT: If the Frontend is tt-xla, we might not need this heristic as the weights could be distinguished by `ttcore.argument_type`. Could you add this?

lib/Dialect/StableHLO/Transforms/AutoSharding.cpp
```diff
@@ -0,0 +1,627 @@
...
+                               StringRef meshName, StringRef shardAxisName) {
+  MLIRContext *context = module.getContext();
+
+  auto funcOps = module.getOps<func::FuncOp>();
+  if (funcOps.empty()) {
+    return;
+  }
+  func::FuncOp funcOp = *funcOps.begin();
+
+  constexpr bool isClosed = false;
```
sshonTT, line 221: Could you explain why this is hard-coded as false? Don't we need a case for open variant?

lib/Dialect/StableHLO/Transforms/AutoSharding.cpp
```diff
@@ -0,0 +1,627 @@
...
+// ∏(rank_i + 1), which is much smaller than 2^totalDims because invalid
+// multi-shard-per-tensor configs are never generated.
+static llvm::SmallVector<ArgDimSharded>
+enumerateArgShardings(ModuleOp &module) {
+  llvm::SmallVector<ArgDimSharded> configs;
+  auto funcOps = module.getOps<func::FuncOp>();
+  if (funcOps.empty()) {
+    return configs;
+  }
+  func::FuncOp funcOp = *funcOps.begin();
```
sshonTT, line 159: I am wondering if this safe in case of composite function is in graph. Could you add at least a warning if there are multiple functions in graph?

lib/Dialect/StableHLO/Transforms/AutoSharding.cpp
```diff
@@ -0,0 +1,627 @@
...
+}
+
+// Generate a filesystem-safe directory name for a variant, e.g.
+// "variant_03_arg0-SR_arg1-R".
+static std::string configDirName(size_t idx, const ArgDimSharded &config) {
+  std::string result;
+  llvm::raw_string_ostream os(result);
+  os << llvm::format("variant_%02zu", idx);
+  for (size_t a = 0; a < config.size(); ++a) {
+    os << "_arg" << a << "-";
```
sshonTT, line 340: We probably need hash this dir name in case of we have arg over 500+. The PATH_MAX is limited to 4096 for many cases.

lib/Dialect/StableHLO/Transforms/AutoSharding.cpp
```diff
@@ -0,0 +1,627 @@
...
+      return;
+    }
+    LLVM_DEBUG(llvm::dbgs() << "AutoSharding: evaluating " << configs.size()
+                            << " configurations\n");
+
+    // Set up dump directory.
+    std::string dumpRoot;
+    if (!dumpDir.empty()) {
+      dumpRoot = createDumpRoot(dumpDir);
+      if (!dumpRoot.empty()) {
```
sshonTT: nit : I think we should emit warning for case where dump is failed.

sshonTT (commented):
I took another look and found a few more things to fix. Sorry about that!

lib/Dialect/StableHLO/Transforms/AutoSharding.cpp
apandeyTT: Agreed, this will be an issue for large models. Since the current exhaustive search doesn't scale to 500+ args anyway capped, the variant naming will be redesigned together with the greedy/layer-wise search in a follow-up, where the naming scheme will naturally change to reflect the new search strategy.

lib/Dialect/StableHLO/Transforms/AutoSharding.cpp
apandeyTT: Yes, Added warning

lib/Dialect/StableHLO/Transforms/AutoSharding.cpp
apandeyTT: isClosed = false, means the dimension sharding is "open" in Shardy's terminology. Shardy's propagation pass is free to further refine or add axes to that dimension. Since AutoSharding is providing hints that feed into propagation (which runs after us), we want open annotations so propagation has full freedom to optimize.

lib/Dialect/StableHLO/Utils/ShardingCostModel.cpp
apandeyTT: Thank you for this input, I wasn't aware of ttcore.argument_type before! Added support for it now. When the attribute is present, we use it directly to distinguish weights from activations, with the rank/size heuristic as a fallback for non-tt-xla frontends. This makes the cost model significantly more precise.

lib/Dialect/StableHLO/Transforms/AutoSharding.cpp
apandeyTT: Yes, this version doesn't handle large models with 500+ args, the exhaustive search is intentionally simple for the initial implementation and will hit the cap quickly in those cases. In later versions we've implemented greedy algorithms and layer-wise graph solving to find the most optimal sharding without the combinatorial explosion.

lib/Dialect/StableHLO/Transforms/AutoSharding.cpp
```diff
@@ -0,0 +1,636 @@
...
+
+//===----------------------------------------------------------------------===//
+// Pipeline and I/O helpers.
+//===----------------------------------------------------------------------===//
+
+// Build a sub-pipeline with the remaining StableHLO passes (everything that
+// normally runs after AutoSharding in the stablehlo-pipeline).
+static void addRemainingStableHLOPasses(OpPassManager &pm) {
+  pm.addPass(createDecoupleConstFanoutPass());
+  pm.addPass(createFlattenCompositePass());
```
vmilosevic: :warning: **clang\-diagnostic\-error** :warning:
use of undeclared identifier `` createFlattenCompositePass ``; did you mean `` createReoutlineCompositePass ``?
```suggestion
  pm.addPass(createReoutlineCompositePass());
```

vmilosevic (commented):

sshonTT (approved):

tapspatel (approved):
great improvement! approved

nsmithtt (approved):
Signing off! Looks great, nice work!

PR #7518, opened 2026-03-16
Add TTNN concat/reshape hang workarounds

sshonTT:
### Ticket
N/A

### Problem description
Multi-chip GPT-OSS TP could hang when tiled concat/reshape hit partial-tile tails that produced sub-32B NOC writes. It's indeterministic.

### What's changed
Add TTNN workaround patterns that only rewrite the unsafe cases: narrow tiled reshapes are routed through row-major, and concat pad/slice is restricted to the semantics-safe case where only the final operand is unaligned.

### Checklist

### DIFF
--- a/include/ttmlir/Dialect/TTNN/Transforms/Workarounds/Decomposition/ConcatOpPadDimRewritePattern.h
+++ b/include/ttmlir/Dialect/TTNN/Transforms/Workarounds/Decomposition/ConcatOpPadDimRewritePattern.h
@@ -1,0 +1,36 @@
+// SPDX-FileCopyrightText: (c) 2026 Tenstorrent AI ULC
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#ifndef TTMLIR_DIALECT_TTNN_TRANSFORMS_WORKAROUNDS_DECOMPOSITION_CONCATOPPADDIMREWRITEPATTERN_H
+#define TTMLIR_DIALECT_TTNN_TRANSFORMS_WORKAROUNDS_DECOMPOSITION_CONCATOPPADDIMREWRITEPATTERN_H
+
+#include "ttmlir/Dialect/TTNN/IR/TTNNOps.h"
+
+#include "mlir/IR/PatternMatch.h"
+#include "mlir/Support/LogicalResult.h"
+
+namespace mlir::tt::ttnn::workarounds::decomposition {
+
+// When a ConcatOp is in TILE_LAYOUT and any input's concat-dimension size
+// is not tile-aligned (i.e. not a multiple of 32), the runtime takes the
+// untilize_rm_retilize_concat path. That path calls untilize_with_unpadding
+// with extremely small NOC writes (e.g. 2-4 bytes per element), which hangs
+// on multi-device meshes.
+//
+// This narrowed pattern only applies when the final input is the only operand
+// whose concat dimension is not tile-aligned. In that case, padding the tail
+// input to a multiple of 32, concatenating, and trimming the trailing suffix
+// preserves semantics while bypassing the problematic untilize path.
+class ConcatOpPadDimRewritePattern
+    : public mlir::OpRewritePattern<ttnn::ConcatOp> {
+public:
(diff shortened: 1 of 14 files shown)


AleksKnezevic (approved):

lib/Dialect/TTNN/Transforms/Workarounds/Decomposition/ReshapeNarrowTiledRewritePattern.cpp
```diff
@@ -0,0 +1,119 @@
...
+      int64_t outputSecondLast = outputType.getShape()[outputRank - 2];
+      if (inputSecondLast == outputSecondLast ||
+          (inputSecondLast % kTileWidth == 0 &&
+           outputSecondLast % kTileWidth == 0)) {
+        return failure();
+      }
+    }
+  }
+
+  // Step 1: TILE -> ROW_MAJOR
```
vwellsTT: nit: can these be written as sentences?

include/ttmlir/Dialect/TTNN/Pipelines/TTNNPipelines.h
```diff
@@ -259,6 +259,12 @@ struct TTIRToTTNNDevicePipelineOptions
       llvm::cl::desc("Enable decomposition workaround pass."),
       llvm::cl::init(true)};
 
+  Option<bool> nocDmaHangWorkaroundsEnabled{
+      *this, "enable-noc-dma-hang-workarounds",
```
AleksKnezevic: Can we make the WA global?

sshonTT: I did it first, but made it as option.
Since the hang is non-deterministic, many existing tests were written without this workaround. I didn't want to refactor all of them right now, so I decided to keep it as an optional flag instead. (https://github.com/tenstorrent/tt-mlir/actions/runs/23165856217/job/67305552196)

AleksKnezevic: My worry in introducing specific passes for specific models is that it will grow out of control and we won't be able to run any model without specific overrides. Unless there is a real issue with making this global, I would really prefer it that way. You can probably use AI to help with most of the tests that would need refactoring.

sshonTT: Sounds good. I can work that way.

include/ttmlir/Dialect/TTNN/Transforms/Workarounds/Decomposition/ConcatOpPadDimRewritePattern.h
```diff
@@ -0,0 +1,36 @@
...
+namespace mlir::tt::ttnn::workarounds::decomposition {
+
+// When a ConcatOp is in TILE_LAYOUT and any input's concat-dimension size
+// is not tile-aligned (i.e. not a multiple of 32), the runtime takes the
+// untilize_rm_retilize_concat path. That path calls untilize_with_unpadding
+// with extremely small NOC writes (e.g. 2-4 bytes per element), which hangs
+// on multi-device meshes.
+//
+// This narrowed pattern only applies when the final input is the only operand
+// whose concat dimension is not tile-aligned. In that case, padding the tail
```
sdjordjevicTT, line 22: Do we have issues open on metal, and do we have a repro? What will qualify the removal of this workaround?

sshonTT: I’m working on a minimal repro, but since it's a non-deterministic hang, it's quite tricky to isolate. I’m considering using the entire graph, I’ll add a tracker to narrow down the cause.

sdjordjevicTT: Could you please post that issue here for tracking purposes?

Issue #7382, opened 2026-03-06
MLIR attribute to FB native


rpavlovicTT:
#### 1a. Implement `toNative()` — MLIR attr to T-type

Write `toNative()` functions that produce T-type objects from MLIR attributes. These parallel the existing `toFlatbuffer()` functions but produce lightweight in-memory objects instead of serialized offsets.

Start with `DeviceComputeKernelConfig` (5 scalar fields, no nesting), then `Conv2dConfig` (nested objects).

**Where:** `include/ttmlir/Target/Utils/MLIRToFlatbuffer.h` 

**Reference — existing `toFlatbuffer()` for DeviceComputeKernelConfig:**
```cpp
// MLIRToFlatbuffer.h:848-857
inline ::flatbuffers::Offset<::tt::target::ttnn::DeviceComputeKernelConfig>
toFlatbuffer(FlatbufferObjectCache &cache,
             ttnn::DeviceComputeKernelConfigAttr computeConfigAttr) {
  return ::tt::target::ttnn::CreateDeviceComputeKernelConfig(
      *cache.fbb, toFlatbuffer(cache, computeConfigAttr.getMathFidelity()),
      toFlatbuffer(cache, computeConfigAttr.getMathApproxMode()),
      toFlatbuffer(cache, computeConfigAttr.getFp32DestAccEn()),
      toFlatbuffer(cache, computeConfigAttr.getPackerL1Acc()),
      toFlatbuffer(cache, computeConfigAttr.getDstFullSyncEn()));
}
```

**New `toNative()` produces a T-type instead — no FlatBufferBuilder, no cache.**

#### 1b. Add T-type overloads in runtime utils

Add overloaded `utils::` functions that accept T-type objects in addition to the existing serialized pointer versions.

**Reference — existing serialized version:**
```cpp
// utils.cpp:427-454
::ttnn::DeviceComputeKernelConfig createDeviceComputeKernelConfig(
    const ::tt::target::ttnn::DeviceComputeKernelConfig *config);
```

**New overload accepts T-type reference:**
```cpp
::ttnn::DeviceComputeKernelConfig createDeviceComputeKernelConfig(
    const ::tt::target::ttnn::DeviceComputeKernelConfigT &config);
```

#### 1c. Three-way consistency test

Write a test that proves all three conversion paths produce identical TTNN objects:

- **Path A** (existing OpModel): `MLIR attr -> conversion::getDeviceComputeKernelConfig() -> TTNN`
[...]

PR #7526, opened 2026-03-17
MLIR attribute to FB native

pljubisavljevicTT:
### Ticket
[Link to Github Issue](https://github.com/tenstorrent/tt-mlir/issues/7382)
### Problem description
Three different paths for conversion from MLIR attribute to TTNN objects

### What's changed
Added toNative functions for ComputeKernelConfig and Conv2dConfig

### Checklist
- [ ] ThreeWayConsistency test for comparing results of three different conversion paths


include/ttmlir/Target/Utils/MLIRToFlatbuffer.h
```diff
@@ -566,6 +569,20 @@ toFlatbuffer(FlatbufferObjectCache &cache,
       *cache.fbb, toFlatbuffer(cache, coreRangeSetAttr.getCoreRanges()));
 }
 
+inline ::tt::target::ttnn::CoreRangeSetT
+toNative(ttnn::CoreRangeSetAttr coreRangeSetAttr) {
+  ::tt::target::ttnn::CoreRangeSetT coreRangeSetT;
+  ::flatbuffers::FlatBufferBuilder dummyBuilder;
+  FlatbufferObjectCache dummyCache(&dummyBuilder);
```
rpavlovicTT: We don't need these dummies. Easiest solution for you would be to mirror the existing toFlatbuffer() overloads but drop the cache param. For example

```cpp
  inline ::tt::target::ttnn::MathFidelity
  toNative(ttnn::MathFidelity mathFidelity) {
    // exact same switch body as toFlatbuffer(cache, mathFidelity)
  }

  inline ::tt::target::ttnn::TensorMemoryLayout
  toNative(ttnn::TensorMemoryLayout layout) { ... }

  inline ::tt::target::ttnn::Layout
  toNative(ttnn::Layout layout) { ... }

  inline ::tt::target::DataType
  toNative(ttcore::DataType dtype) { ... }

  // Optional wrapper — produces flatbuffers::Optional<T> from std::optional<MLIR_T>
  template <typename T>
  ::flatbuffers::Optional<decltype(toNative(std::declval<T>()))>
  toNative(const std::optional<T> &opt) {
    if (!opt) return ::flatbuffers::nullopt;
    return toNative(*opt);
  }
```

  Then toNative(DeviceComputeKernelConfigAttr) becomes clean:

```cpp
  inline ::tt::target::ttnn::DeviceComputeKernelConfigT
  toNative(ttnn::DeviceComputeKernelConfigAttr attr) {
    ::tt::target::ttnn::DeviceComputeKernelConfigT t;
    t.math_fidelity   = toNative(attr.getMathFidelity());
    t.math_approx_mode = toNative(attr.getMathApproxMode());
    t.fp32_dest_acc_en = toNative(attr.getFp32DestAccEn());
    t.packer_l1_acc    = toNative(attr.getPackerL1Acc());
    t.dst_full_sync_en = toNative(attr.getDstFullSyncEn());
    return t;
  }
```

test/unittests/MLIRAttrToFBNative/ThreeWayConsistency.cpp
```diff
@@ -0,0 +1,313 @@
...
+  }
+};
+
+class Conv2dConfigTest
+    : public ThreeWayConsistencyTest<mlir::tt::ttnn::Conv2dConfigAttr,
+                                     ::ttnn::Conv2dConfig> {
+protected:
+  std::optional<::ttnn::Conv2dConfig>
+  pathA(mlir::tt::ttnn::Conv2dConfigAttr conv2dConfigAttr) override {
+    // TODO(#2130)
```
rpavlovicTT: Comment unclear

test/unittests/MLIRAttrToFBNative/ThreeWayConsistency.cpp
```diff
@@ -0,0 +1,313 @@
...
+                        pathB->enable_weights_double_buffer,
+                        pathC->enable_weights_double_buffer);
+    EXPECT_PRED_FORMAT3(Equal<std::optional<bool>>,
+                        pathA->enable_kernel_stride_folding,
+                        pathB->enable_kernel_stride_folding,
+                        pathC->enable_kernel_stride_folding);
+  }
+};
+
+TEST_F(DeviceComputeKernelConfigTest, ComputeKernelConfig) {
```
rpavlovicTT: Use TEST_P with INSTANTIATE_TEST_SUITE_P rather than individual TEST_F cases.

You should parameterize over: 
- All fields set to non-default values, 
- All fields unset (nullopt)
- Partial field combinations

test/unittests/MLIRAttrToFBNative/ThreeWayConsistency.cpp
```diff
@@ -0,0 +1,313 @@
...
+};
+
+class Conv2dConfigTest
+    : public ThreeWayConsistencyTest<mlir::tt::ttnn::Conv2dConfigAttr,
+                                     ::ttnn::Conv2dConfig> {
+protected:
+  std::optional<::ttnn::Conv2dConfig>
+  pathA(mlir::tt::ttnn::Conv2dConfigAttr conv2dConfigAttr) override {
+    // TODO(#2130)
+    if (conv2dConfigAttr.hasCoreGrid()) {
```
rpavlovicTT: Why skipping this?

test/unittests/MLIRAttrToFBNative/ThreeWayConsistency.cpp
```diff
@@ -0,0 +1,313 @@
...
+        pathC == std::nullopt) {
+      if (pathA == std::nullopt && pathB == std::nullopt &&
+          pathC == std::nullopt) {
+        SUCCEED();
+        return;
+      }
+      FAIL();
+    }
+
+    EXPECT_PRED_FORMAT3(Equal<MathFidelity>, pathA->math_fidelity,
```
rpavlovicTT: I think it's simpler just to use instead of EXPECT_PRED_FORMAT3.
```cpp
  EXPECT_EQ(resultA->math_fidelity, resultB->math_fidelity);                                                                                                                                           
  EXPECT_EQ(resultA->math_fidelity, resultC->math_fidelity);  
```

test/unittests/MLIRAttrToFBNative/ThreeWayConsistency.cpp
```diff
@@ -0,0 +1,313 @@
...
+  }
+
+  ::testing::AssertionResult failure = ::testing::AssertionFailure();
+  failure << e1 << " == " << e2 << " && " << e1 << " == " << e3 << " failed;";
+  return failure;
+}
+
+template <>
+::testing::AssertionResult
+Equal<std::optional<ttnn::operations::unary::UnaryWithParam>>(
```
rpavlovicTT: I feel it is cleaner to drop these complex overrides for `EXPECT_PRED_FORMAT3` and use simple EQ comparisons as below.

test/unittests/MLIRAttrToFBNative/ThreeWayConsistency.cpp
```diff
@@ -0,0 +1,313 @@
...
+        SUCCEED();
+      }
+      FAIL();
+    }
+
+    EXPECT_PRED_FORMAT3(Equal<std::optional<tt::tt_metal::DataType>>,
+                        pathA->weights_dtype, pathB->weights_dtype,
+                        pathC->weights_dtype);
+    EXPECT_PRED_FORMAT3(
+        Equal<std::optional<ttnn::operations::unary::UnaryWithParam>>,
```
rpavlovicTT: You can use lambda like
```cpp
  auto compareUnaryWithParam = [](                                                                                                                                                                     
      const std::optional<ttnn::operations::unary::UnaryWithParam> &a,                                                                                                                                 
      const std::optional<ttnn::operations::unary::UnaryWithParam> &b) {                                                                                                                               
    if (a.has_value() != b.has_value()) return false;                                                                                                                                                  
    if (!a) return true;                                                                                                                                                                               
    return a->op_type == b->op_type && a->params == b->params;                                                                                                                                         
  };   
```
to compare UnaryWithParam fields

rpavlovicTT (commented):
Looking good generally, few comments for improvements

include/ttmlir/Target/Utils/MLIRToFlatbuffer.h
nsmithtt: @rpavlovicTT, I guess an argument for this PR's approach though is it avoids duplicate logic in `toFlatbuffer`.  To your point though, perhaps there's a way we can have the impl be in `toNative` and the `toFlatbuffer` impls use `toNative` under the hood to get the object and then just put that object into the cache.  Not sure this is possible/as performant

rpavlovicTT: That should be possible like 
```cpp                                                                                 
  inline ::flatbuffers::Offset<::tt::target::ttnn::DeviceComputeKernelConfig>
  toFlatbuffer(FlatbufferObjectCache &cache,                                                                                                                                                           
               ttnn::DeviceComputeKernelConfigAttr attr) {                                                                                                                                             
    auto t = toNative(attr);                                                                                                                                                                           
    return ::tt::target::ttnn::DeviceComputeKernelConfig::Pack(*cache.fbb, &t);                                                                                                                        
  }       
```

I didn't want to induce major changes to current `MLIRToFlatbuffer` implementation. This was an onboarding task  for Petar, as part of bigger effort to unify op invocation paths (OpModel & tt runtime) to always go through FB schema. If you agree, once we have `toNative` support across all types, we can just change `toFlatbuffer` to use the above implementation .

nsmithtt: I'd somewhat prefer doing the formulation you provided above, I just think it's a dangerous state to be in to have all of this serialization code being duplicated, they could get out of sync and introduce subtle bugs.

I wonder if @pljubisavljevicTT's dummy cache approach could work via a generic template and handle all these API's in one fell swoop:
```c++
  // Trait to detect flatbuffers::Offset<T>
  template <typename T>
  struct is_flatbuffer_offset : std::false_type {};

  template <typename T>
  struct is_flatbuffer_offset<flatbuffers::Offset<T>> : std::true_type {};

  template <typename T>
  inline auto toNative(T value)
      -> std::enable_if_t<!is_flatbuffer_offset<ToFlatbufferReturnType<T>>::value,
                          ToFlatbufferReturnType<T>> {
    FlatbufferObjectCache dummyCache(nullptr);
    return toFlatbuffer(dummyCache, value);
  }

  template <typename T>
  auto toNative(T value)
      -> std::enable_if_t<is_flatbuffer_offset<ToFlatbufferReturnType<T>>::value,
                          typename ToFlatbufferReturnType<T>::type::NativeTableType> {
    flatbuffers::FlatBufferBuilder fbb;
    FlatbufferObjectCache cache(&fbb);
    auto offset = toFlatbuffer(cache, value);
    fbb.Finish(offset);
    auto *fb = flatbuffers::GetRoot<
        typename ToFlatbufferReturnType<T>::type>(fbb.GetBufferPointer());
    return *fb->UnPack();
  }
```

This minimizes code change and avoids duplicate implementations.

nsmithtt: The only drawback is perf of `toNative`, but maybe it's worth seeing if it causes an issue.

nsmithtt: Also the above doesn't handle vargs, but could be adapted

include/ttmlir/Target/Utils/MLIRToFlatbuffer.h
```diff
@@ -225,6 +274,20 @@ toFlatbuffer(FlatbufferObjectCache &, ttnn::TensorMemoryLayout memLayout) {
   }
 }
 
+inline ::tt::target::ttnn::TensorMemoryLayout
+toNative(ttnn::TensorMemoryLayout memLayout) {
+  switch (memLayout) {
```
vmilosevic: :warning: **clang\-diagnostic\-switch** :warning:
enumeration value `` NDSharded `` not handled in switch

include/ttmlir/Target/Utils/MLIRToFlatbuffer.h
```diff
@@ -225,6 +274,20 @@ toFlatbuffer(FlatbufferObjectCache &, ttnn::TensorMemoryLayout memLayout) {
...
+  case ttnn::TensorMemoryLayout::Interleaved:
+    return ::tt::target::ttnn::TensorMemoryLayout::Interleaved;
+  case ttnn::TensorMemoryLayout::HeightSharded:
+    return ::tt::target::ttnn::TensorMemoryLayout::HeightSharded;
+  case ttnn::TensorMemoryLayout::WidthSharded:
+    return ::tt::target::ttnn::TensorMemoryLayout::WidthSharded;
+  case ttnn::TensorMemoryLayout::BlockSharded:
+    return ::tt::target::ttnn::TensorMemoryLayout::BlockSharded;
+  }
+}
```
vmilosevic, line 243: :warning: **clang\-diagnostic\-return\-type** :warning:
non\-void function does not return a value in all control paths

test/unittests/MLIRAttrToFBNative/ThreeWayConsistency.cpp
```diff
@@ -0,0 +1,328 @@
...
+              pathC->enable_kernel_stride_folding);
+  }
+};
+
+TEST_P(DeviceComputeKernelConfigTest, DeviceComputeKernelConfig) {
+  auto params = GetParam();
+  RunTest(params);
+}
+
+mlir::MLIRContext *contextDeviceComputeKernelConfig =
```
vmilosevic: :warning: **cppcoreguidelines\-avoid\-non\-const\-global\-variables** :warning:
variable `` contextDeviceComputeKernelConfig `` is non\-const and globally accessible, consider making it const

vmilosevic: :warning: **cppcoreguidelines\-avoid\-non\-const\-global\-variables** :warning:
variable `` contextDeviceComputeKernelConfig `` provides global access to a non\-const object; consider making the pointed\-to data `` const ``

test/unittests/MLIRAttrToFBNative/ThreeWayConsistency.cpp
```diff
@@ -0,0 +1,328 @@
...
+INSTANTIATE_TEST_SUITE_P(
+    DeviceComputeKernelConfigTest, DeviceComputeKernelConfigTest,
+    ::testing::ValuesIn(deviceComputeKernelConfigAttrlist));
+
+TEST_P(Conv2dConfigTest, Conv2dConfig) {
+  auto params = GetParam();
+  RunTest(params);
+}
+
+mlir::MLIRContext *contextConv2dConfig = Conv2dConfigTest::getContext();
```
vmilosevic: :warning: **cppcoreguidelines\-avoid\-non\-const\-global\-variables** :warning:
variable `` contextConv2dConfig `` is non\-const and globally accessible, consider making it const

vmilosevic: :warning: **cppcoreguidelines\-avoid\-non\-const\-global\-variables** :warning:
variable `` contextConv2dConfig `` provides global access to a non\-const object; consider making the pointed\-to data `` const ``

vmilosevic (commented):

include/ttmlir/Target/Utils/MLIRToFlatbuffer.h
rpavlovicTT: I agree with your point about duplication.   
OTOH, I think your last suggestion leads us to a kind of a wrong path. Our goal is to make native class types canonical form.  Your approach involves serialize-then-deserialize round trip just to get a T-type object.  

Can we then agree on making `toNative` hold the translation logic and `toFlatbuffer` being the thin wrapper?

nsmithtt: > Can we then agree on making toNative hold the translation logic and toFlatbuffer being the thin wrapper?

Sure although I'm not as confident what the right tradeoff is:
- `toFlatbuffer` as they are written today will serialize directly inline to a buffer, i.e. more performant.
- If we prioritize `toNative` to have the translation logic you need to construct the object and then copy it into flatbuffer.

TBH I'm impartial and probably we should favor whatever results in the most maintainable code.  Just pointing out that there are tradeoffs, but maybe flatbuffer serialization perf hit is negligible, not sure.

nsmithtt: I'm ok with either approach, my main concern was code duplication so however we resolve that I'm on board.

Issue #7419, opened 2026-03-09
[D2M][ttnn.jit] JIT'ed Composite Ops with Matmul Failing in D2MToTTNN pass


brapananTT:
A matmul op in the middle / end of a JIT'ed op chain is currently failing in `D2MToTTNN` pass.

Example JIT'ed composite op:

```
def matmul_composite(input0, input1):
    a = ttnn.abs(input0)
    b = ttnn.matmul(a, input1)
    c = ttnn.abs(b)
    return c
```

Errors out with the following message:

```
/tt-mlir/lib/Conversion/D2MToTTNN/D2MToTTNN.cpp:369: 
static IOAndCB mlir::tt::(anonymous namespace)::D2MGenericRewriter::extractIOAndCBFromGenericOperand(Value, Value): 
Assertion `castOp` failed, Expected TTNNMetalLayoutCastOp producing stream input.

```

The error message from D2MToTTNN.cpp comes from `extractIOAndCBFromGenericOperand`:

```
static IOAndCB extractIOAndCBFromGenericOperand(Value origOperand,
                                                  Value convertedOperand) {
    if (auto streamLayoutOp = mlir::dyn_cast_if_present<d2m::StreamLayoutOp>(
            origOperand.getDefiningOp())) {
      auto castOp = mlir::dyn_cast_if_present<ttir::TTNNMetalLayoutCastOp>(
          streamLayoutOp.getInput().getDefiningOp());
      TT_assertv(castOp,    // fails on this assert
                 "Expected TTNNMetalLayoutCastOp producing stream input.");
      return {castOp.getOperand(), streamLayoutOp.getStorage()};
    }
...
```

brapananTT:
@arminaleTT this issue seems to be fixed on your PR branch #7516. Setting the op in `test_matmul_smoketest.py` back to `matmul_composite` passes all tests.

brapananTT:
This issue was fixed by #7516. Composite matmul tests to be enabled in #7605

Issue #7163, opened 2026-02-23
[ttnn.jit] Adapt JIT Frontend Matmul Tracing To Intermediate Layout Inference


arminaleTT:
Previously, JIT frontend used hacks to infer the correct layouts for matmul outputs. This no longer works with intermediate layouts being inferred and should be removed. This was not done in #7144 to contain the PR scope

PR #7533, opened 2026-03-17
[D2M][ttnn.jit] Support JIT'ed Composite Matmuls

brapananTT:
### Ticket
#7419 JIT'ed Composite Ops with Matmul Failing in D2MToTTNN pass 

### Problem description
After fixing matmul tracing in ttnn-jit #7163, jit'ed graphs with matmul included as an intermediate / at end of graph would fail out in D2MToTTNN pass. 

### What's changed
Updated `extractIOAndCBFromGenericOperand` in `D2MToTTNN` to handle intermediate buffers between chained generics (memref.alloc wrapped by stream_layout) by adding helper `isOutputOfGeneric`. Updated `isAliasedOutput` to correctly return true for an intermediate data alloc used as a stream's input.

### Checklist
- [x] New/Existing tests provide coverage for changes: Updated tests in `test_matmul_smoketest.py` to use a composite jit'ed matmul instead of single matmuls. Tests passing.

### DIFF
--- a/lib/Conversion/D2MToTTNN/D2MToTTNN.cpp
+++ b/lib/Conversion/D2MToTTNN/D2MToTTNN.cpp
@@ -332,8 +332,6 @@ public:
       // TODO (#7158): This is brittle. Ideally we should specifically identify
       // outputs and handle them separately from inputs, but that will require a
       // larger refactor of this pass.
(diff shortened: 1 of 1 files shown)


arminaleTT (changes requested):
Please wait for #7516. It will either fix it, or make this change much easier

PR #7541, opened 2026-03-17
Revert "Revert "Implement constant folding for ttir.neg (#7385)" (#7509)"

amalkovTT:
This reverts commit 33c867d354463ffb2d8dd5a950f1691a90b6baf5.

### Ticket
#7357 

### Problem description
#7385 was reverted because it failed tests in tt-xla. The underlying issue should be fixed by [PR](https://github.com/tenstorrent/tt-xla/pull/3768) in tt-xla, so this can now be reverted.

### What's changed
reverts commit 33c867d354463ffb2d8dd5a950f1691a90b6baf5

### Checklist

### DIFF
--- a/lib/Dialect/TTIR/IR/TTIROps.cpp
+++ b/lib/Dialect/TTIR/IR/TTIROps.cpp
@@ -4891,7 +4891,7 @@ static mlir::Attribute convertScalarAttribute(mlir::TypedAttr typedAttr,
   if (auto floatAttr = mlir::dyn_cast<mlir::FloatAttr>(typedAttr)) {
     llvm::APFloat floatVal = floatAttr.getValue();
 
-    // Case A: Float -> Float (e.g., f64 -> f32)
+    // Case A: Float -> Float (e.g., f32 -> f64)
     if (auto targetFloatType = mlir::dyn_cast<mlir::FloatType>(targetType)) {
       bool losesInfo;
       floatVal.convert(targetFloatType.getFloatSemantics(),
@@ -4911,9 +4911,13 @@ static mlir::Attribute convertScalarAttribute(mlir::TypedAttr typedAttr,
   if (auto intAttr = mlir::dyn_cast<mlir::IntegerAttr>(typedAttr)) {
     llvm::APInt intVal = intAttr.getValue();
 
-    // Case C: Integer -> Integer (e.g., i64 -> i32)
+    // Case C: Integer -> Integer (e.g., i32 -> i64)
     if (auto targetIntType = mlir::dyn_cast<mlir::IntegerType>(targetType)) {
-      intVal = intVal.sextOrTrunc(targetIntType.getWidth());
+      if (intAttr.getType().isUnsignedInteger()) {
+        intVal = intVal.zextOrTrunc(targetIntType.getWidth());
+      } else {
+        intVal = intVal.sextOrTrunc(targetIntType.getWidth());
+      }
       return mlir::IntegerAttr::get(targetType, intVal);
     }
 


lib/Dialect/TTIR/IR/TTIROps.cpp
```diff
@@ -820,6 +821,33 @@ mlir::tt::ttir::GetDimensionSizeOp::fold(FoldAdaptor adaptor) {
...
+    return nullptr;
+  }
+
+  if (auto denseAttr = dyn_cast<DenseElementsAttr>(attr)) {
+    Type elementType = denseAttr.getElementType();
+    if (elementType.isInteger()) {
+      return denseAttr.mapValues(elementType,
+                                 [](const APInt &val) { return -val; });
+    }
+    if (elementType.isFloat()) {
```
mtopalovicTT, line 841: Nit comment on why bitcastToAPInt is needed. Also it would be good to mention that mapped values will still be of APFloat.

lib/Dialect/TTIR/IR/TTIROps.cpp
```diff
@@ -4851,6 +4879,90 @@ verifyReplicaGroups(mlir::DenseIntElementsAttr replicaGroups) {
...
+
+// FullOp folder
+::mlir::OpFoldResult mlir::tt::ttir::FullOp::fold(FoldAdaptor adaptor) {
+  auto fillValue = llvm::dyn_cast<TypedAttr>(getFillValueAttr());
+  RankedTensorType resultType = getResult().getType();
+
+  // Fill value is 32-bit float or 32-bit signless integer, but result type
+  // might differ.
+  auto convertedFillValue =
+      convertScalarAttribute(fillValue, resultType.getElementType());
```
mtopalovicTT, line 4947: Why is this needed? Is this because ElementTypeNormalization doesn't normalize attribute but it normalizes result type?

lib/Dialect/TTIR/IR/TTIRDialect.cpp
```diff
@@ -118,10 +118,51 @@ void TTIRDialect::initialize() {
 // TTIR constant materializer.
 //===----------------------------------------------------------------------===//
 
+// Check if fill_value attirbute of FullOp can be of this type.
+bool isValidFullValueType(Type t) {
```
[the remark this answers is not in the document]
mtopalovicTT: @amalkovTT I would just inline this into if statement, it's only used in materializatiion.

lib/Dialect/TTIR/IR/TTIROps.cpp
```diff
@@ -4851,6 +4879,90 @@ verifyReplicaGroups(mlir::DenseIntElementsAttr replicaGroups) {
   return std::nullopt;
 }
 
+// Helper to convert type of scalar attribute.
+mlir::Attribute convertScalarAttribute(mlir::TypedAttr typedAttr,
+                                       mlir::Type targetType) {
```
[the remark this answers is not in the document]
mtopalovicTT: @amalkovTT this is valid.

lib/Dialect/TTIR/IR/TTIROps.cpp
amalkovTT: For example, this is valid IR (and even used in some tests): `%0 = "ttir.full"() <{shape = array<i32: 1>, fill_value = -5 : i32}> : () -> tensor<1xf64>` (attribute is integer, but return type is float). Also, `ttir.full` is defined to have `fill_value` attribute of type i32 or f32, but the actual return type can be any other type (ui64, bf16...).

PR #7542, opened 2026-03-17
[d2m] fabric pipeline

dlokeTT:
### Problem description
Fabric tests were using ttkernel/post bufferization ir snippets but then running the full `ttir-to-ttmetal` pipeline caused unnecessary checks in ttir passes. This is a pre req for https://github.com/tenstorrent/tt-mlir/pull/7431

### What's changed
New fabric pipeline with `ttcore` passes and `ttmetal-backend` pipeline
Updated fabric tests

### Checklist

### DIFF
--- a/test/python/golden/test_fabric_apis.py
+++ b/test/python/golden/test_fabric_apis.py
@@ -20,6 +20,8 @@ from ttmlir.passes import (
 from ttmlir.passmanager import PassManager
 import math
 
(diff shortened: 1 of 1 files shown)


sohaibnadeemTT:
@nsmithtt This should just be using the backend pipeline but it does hit errors. I won't block @dlokeTT on this though, I can fix it in a follow-up PR.

dlokeTT:
> Why do we need a custom pipeline for this? Why don't the tests just manually call those passes as part of their pass pipeline string?

Updated to doing this

Issue #6475, opened 2025-12-30
[D2M] Slice OP and DMA restrictions


wenbinlyuTT:
Currently there are at least three ways a `ttir.slice_static` op can generate unsupported data movement patterns.
- The start index of the innermost dim leads to unaligned NoC addresses (16B for L1, 32B for DRAM).
- The innermost dim is strided (step > 1).
- The DMA coalescing factor crosses core boundaries (e.g. for a 64x64 tensor, slice the width with `[10:1:42]`, but the DMA lowering pass chooses a coalescing factor of 4).

All failing sliceOp tests in `test_data_movement.py` will pass if we align their innermost start indices and use step size 1.

PR #7547, opened 2026-03-17
D2M: support NoC-unfriendly slice ops

wenbinlyuTT:
### Ticket
Related-To: #6475

### Problem description
D2M slice op isn't working correctly if:
- The width slice start at a non-NoC-aligned address
- The width step/stride isn't 1 (violates minimal NoC transfer size requirement)

### What's changed
- Add 1D -> 2D support for the slice op in the rank normalization pass
- Add a high-priority rewriter in `TTIRToD2M` that:
  - If the width slice is NoC-unfriendly but not the height dim, do Transpose-Slice-Transpose
  - If both width & height are NoC-unfriendly, do Slice(crop width)-Transpose-Slice(height only)-Transpose
 - Teach `MaterializeViewReturns` to also recognize `view -> to_host` that doesn't immediately return
 - Add detailed comment in `LowerToLayout` about why a virtual grid reblocking shouldn't happen 

### Caveats
- The intermediate results are written to the host, and immediately fetched back
- This is subject to the permute op's limitation when dealing with tall/wide tensors #7559

### Checklist

### DIFF
--- a/test/python/golden/ttir_ops/data_movement/test_data_movement.py
+++ b/test/python/golden/ttir_ops/data_movement/test_data_movement.py
@@ -371,6 +371,11 @@ def test_cpu_hoistable_reshape_op(
         ((4, 5, 32, 64), [1, 2, 0, 0], [3, 4, 32, 64], [1, 1, 1, 1]),
         ((8, 6, 64, 32), [2, 1, 0, 0], [6, 5, 64, 32], [1, 1, 1, 1]),
         ((2, 4, 3, 32, 64), [0, 1, 1, 0, 0], [2, 3, 2, 32, 64], [1, 1, 1, 1, 1]),
+        # Simple 1D
+        ((64,), [0], [32], None),
+        # Strided 1D
+        ((70,), [3], [62], [7]),
+        ((2048,), [0], [2048], [3]),
         # Simple 2D
         ((64, 64), [0, 0], [32, 32], None),
         # Crop 2D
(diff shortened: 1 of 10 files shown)


test/python/golden/ttir_ops/data_movement/test_data_movement.py
```diff
@@ -379,7 +384,17 @@ def hoisted_reshape_wrapper(
         ((192, 64), [2, 0], [192, 64], [3, 1]),
         ((64, 192), [0, 2], [64, 192], [1, 3]),
         # Sample large 2D tensors
-        ((32, 131072), [0, 3], [32, 128 * 991], [2, 991]),
+        pytest.param(
+            (32, 131072),
+            [0, 3],
+            [32, 128 * 991],
+            [2, 991],
+            marks=pytest.mark.skip_config(
```
sdjordjevicTT, line 392: Can we skip this test case only for the ttmetal target? I guess ttnn target passes?

nsmithtt:
```python
((32, 131072), [0, 3], [32, 128 * 991], [2, 991]) | SkipIf(["ttmetal", "p150"], ["ttmetal", "p300"])
```

wenbinlyuTT: Unfortunately `SkipIf` won't work on this tuple that needs unpacking, I just specified the ttmetal target in `skip_config`.

Issue #6396, opened 2025-12-19
[Regression]`ttir.slice_static` Verification Failure during tt-mlir Uplift


chandrasekaranpradeep:
In the tt-forge-fe repository, while uplifting tt-mlir, the following two index op test cases failed during MLIR verification. The `ttir.index` op is decomposed to `ttir.slice_static` in `IndexToSliceConversionPattern`, and during the verification of `ttir.slice_static` op in `SliceStaticOp::verify()`, it fails with the error: `'ttir.slice_static' op For positive step, begin index must be less than or equal to end index for dimension 0.`

## Test Case 1

**Test:** `forge/test/mlir/operators/indexing/test_basic_slicing.py::test_multidimensional_slicing[negative_indexing_subrange]`

**Exact Error Message:**
```
error: 'ttir.slice_static' op For positive step, begin index must be less than or equal to end index for dimension 0. Got begin: 1 (-2), end: 0, step: 1, input shape: (3, 3, 3)
```

**TTIR:**
```mlir
module @SlicingModule attributes {ttcore.system_desc = #ttcore.system_desc<[{role = host, target_triple = "x86_64-pc-linux"}], [{arch = <wormhole_b0>, grid = 8x8, coord_translation_offsets = 18x18, l1_size = 1499136, num_dram_channels = 12, dram_channel_size = 1073741824, noc_l1_address_align_bytes = 16, pcie_address_align_bytes = 32, noc_dram_address_align_bytes = 32, l1_unreserved_base = 102656, erisc_l1_unreserved_base = 98304, dram_unreserved_base = 32, dram_unreserved_end = 1073125888, supported_data_types = [<f32>, <f16>, <bf16>, <bfp_f8>, <bfp_bf8>, <bfp_f4>, <bfp_bf4>, <bfp_f2>, <bfp_bf2>, <u32>, <u16>, <u8>, <si32>], supported_tile_sizes = [ 4x16,  16x16,  32x16,  4x32,  16x32,  32x32], dst_physical_size_tiles = 16, num_cbs = 32, num_compute_threads = 1, num_datamovement_threads = 2}], [0], [1 : i32], [ 0x0x0x0]>} {

  func.func @forward(%arg0: tensor<3x3x3xf32> {ttcore.argument_type = #ttcore.argument_type<input>, ttir.name = "x"}) -> (tensor<2x2x3xf32> {ttir.name = "SlicingModule.output_Index_1"}) {

[...]

PR #7564, opened 2026-03-19
Fix negative indexes for slice op

djstefanovicTT:
### Ticket
closes https://github.com/tenstorrent/tt-mlir/issues/6396

### Problem description
Index op with negative bounds has incorrect calculation when expanded to array.

### What's changed
Changed to always use positive bounds in calculations.

### Checklist


sdjordjevicTT:
Can we add one example with negative dims in the builder test as well:
test/python/golden/ttir_ops/data_movement/test_data_movement.py::test_slice

PR #7565, opened 2026-03-19
[EmitPy] Conversion of `ttnn.conv3d`

azecevicTT:
### Ticket
https://github.com/tenstorrent/tt-xla/issues/3656#issuecomment-4089354582

### Problem description
Self-explanatory.

### What's changed
Self-explanatory.

### Checklist


sdjordjevicTT:
> Lgtm. Maybe it would be nice to add a simple lit test to keep it consistent with EmitC, which has some.

Builder tests are enough, we should depricate those emitC lit tests, as we have a builder support for emitC as well. :)

PR #7566, opened 2026-03-19
[TTIR] Generalize matmul canonicalization to absorb leading dim merge/split reshapes

jserbedzijaTT:
## Summary
- Generalizes the TTIR matmul canonicalization to absorb leading dimension merge/split reshapes around matmul ops (e.g., `[32,32,S,D] → [1024,S,D]`), not just leading size-1 squeeze/unsqueeze
- Unifies the old leading-1 squeeze/unsqueeze pattern and the new general merge pattern into a single canonicalization, since the former is a strict subset of the latter
- Enables downstream SDPA fusing to match 4D `[B,H,S,D]` attention patterns from XLA/JAX that were previously hidden behind 3D batch matmuls

## Context
XLA/JAX emits `dot_general` with batch and head dimensions already merged before the attention matmuls. This produced an IR like:
```
%a = reshape [32,32,S,D] → [1024,S,D]
%b = reshape [32,32,D,S] → [1024,D,S]
%r = matmul(%a, %b)      → [1024,S,S]
%o = reshape %r           → [32,32,S,S]
```
The SDPA fusing pattern requires 4D tensors and couldn't look through these reshapes. This canonicalization absorbs them into the matmul, restoring the 4D form so SDPA fusing matches.

## Test plan
- [x] All 23 existing TTIR canonicalization tests pass (including updated squeeze/unsqueeze tests)
- [x] New test file with 8 cases covering general merge (positive) and edge cases (negative)
- [x] Verified SDPA fuses on a real XLA-generated LLaMA decode IR (`ttnn.scaled_dot_product_attention_decode` appears in output)

### DIFF
--- a/lib/Dialect/TTIR/IR/TTIROps.cpp
+++ b/lib/Dialect/TTIR/IR/TTIROps.cpp
@@ -3692,46 +3692,102 @@ void mlir::tt::ttir::LinearOp::getCanonicalizationPatterns(
 }
 // ANCHOR_END: adding_an_op_matmul_ttir_verify
 
-// Returns true if `longer` is `shorter` with leading size-1 dimensions
-// prepended.
-static bool isLeadingOnesToShorter(llvm::ArrayRef<int64_t> longer,
-                                   llvm::ArrayRef<int64_t> shorter) {
-  if (longer.size() <= shorter.size()) {
-    return false;
+// Returns the number of leading input dimensions that are merged into the
+// first output dimension. Returns 0 if the reshape is not a leading dimension
+// merge.
+//
+// A leading merge reshapes [d0, d1, ..., dk, t0, t1, ...] ->
+//                          [d0*d1*...*dk, t0, t1, ...]
+// where the trailing dimensions are preserved exactly.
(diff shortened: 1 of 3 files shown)


umalesTT:
Hey @jserbedzijaTT ,
we hit this problem with redundant reshapes on inputs for attention matmul before, but on torch side. Merging batch and head dimension resulted in redundant CCL's for us and it was [solved on frontend side](https://github.com/tenstorrent/tt-xla/pull/2364). 
Do you think we could unify approaches for torch and jax in, say, `StablehloFusing` pass? It seems to me that otherwise we would have same CCL problem on jax side

fyi @AleksKnezevic

jserbedzijaTT:
> Hey @jserbedzijaTT , we hit this problem with redundant reshapes on inputs for attention matmul before, but on torch side. Merging batch and head dimension resulted in redundant CCL's for us and it was [solved on frontend side](https://github.com/tenstorrent/tt-xla/pull/2364). Do you think we could unify approaches for torch and jax in, say, `StablehloFusing` pass? It seems to me that otherwise we would have same CCL problem on jax side
> 
> fyi @AleksKnezevic

I  plan on merging this anyway, could you open an issue for the scenario you described?

PR #7576, opened 2026-03-19
[D2M] Insert memref deallocs after hoisting

vwellsTT:
### Ticket
NA

### Problem description
Some areas of CB rework created new problems/ugliness to cleanup.

### What's changed
1 major fix (other 2 got resolved in other PRs):
1. insert deallocs after hoisting memref.allocs out of generic region

### Checklist


lib/Dialect/D2M/Transforms/Allocate.cpp
```diff
@@ -1667,12 +1667,17 @@ class D2MAllocate final : public impl::D2MAllocateBase<D2MAllocate> {
...
+    // Use the generic's compute grid for the CB layout.  The grid rank must
+    // match the shard rank so that downstream [grid..shard..] + ShardLayoutAttr
+    // expansion (serializer, D2MToTTNN) produces a valid memref.  Collapse
+    // the compute grid to 2D when the shard rank is smaller.
+    SmallVector<int64_t> computeGrid;
+    auto rawGrid = genericOp.getGrid().getShape();
+    if (rawGrid.size() <= shardShape.size()) {
+      computeGrid.assign(rawGrid.begin(), rawGrid.end());
+    } else {
+      computeGrid = ttcore::collapseGridTo2D(rawGrid);
```
nsmithtt: I also ran into this problem on my generic test PR [here](https://github.com/tenstorrent/tt-mlir/pull/7493/changes#diff-663d18a0d4e6ddb5296775835d5c3dd657c48103606b57629f2f320824c3586e), I opted for just filling 1's i.e. unit grid to match the rank of the shard.  There are some reduction examples where different operands have shard shapes that mismatch with the generic grid shape and TBH I don't think the juice is worth the squeeze to calc what grid the CB actually belongs to.

nsmithtt: That said, this might be moot once stream_layout is removed from allocator

lib/Dialect/D2M/Transforms/Allocate.cpp
```diff
@@ -1787,18 +1792,21 @@ class D2MAllocate final : public impl::D2MAllocateBase<D2MAllocate> {
...
+              // CBs are allocated across all cores, so the compute grid
+              // is the most logically accurate grid to use here.  Collapse
+              // to match shard rank for valid [grid..shard..] expansion.
+              auto rawGrid = op.getGrid().getShape();
+              SmallVector<int64_t> computeGrid;
+              if (rawGrid.size() <= shardShape.size()) {
+                computeGrid.assign(rawGrid.begin(), rawGrid.end());
+              } else {
+                computeGrid = ttcore::collapseGridTo2D(rawGrid);
+              }
```
nsmithtt: Maybe it should be a helper func since it appears in multiple places?

lib/Dialect/D2M/Transforms/Allocate.cpp
vwellsTT: I think this would still be relevant, but I agree, hadn't thought through all these edge cases.  As long as a unit-grid is functionally fine, I can get rid of the grid field on CBLayoutAttr completely and just always lower to a 1x1 grid in serialization--that seems a lot simpler and debatably clearer w.r.t. to pre-serialization IR too

nsmithtt: yeah let's go with that

vwellsTT: I ended up having to do similar in my final stream removal PR: https://github.com/tenstorrent/tt-mlir/pull/7606 to get things to work.  I think I'll just lump this part in there and remove it here, and just leave the other 2 fixes here

PR #7579, opened 2026-03-19
Builder: reset device after fb execution failure

wenbinlyuTT:
### Problem description
We're suspecting that failed flatbuffer execution can leave the device in a bad state and wreak havoc.

### What's changed
Detect builder flatbuffer execution failures and close-reopen the device (since we don't have a direct reset API).

### DIFF
--- a/tools/builder/base/builder_runtime.py
+++ b/tools/builder/base/builder_runtime.py
@@ -687,6 +687,12 @@ def execute_fb(
     -------
     Tuple[Dict[str, Dict], Dict[str, Dict]]
         golden_report, output_tensors
+
+    Notes
+    -----
+    If execution fails, the device will be automatically closed and reopened
+    using the device_reopener function from conftest.
+    This ensures device state is clean for the next test.
     """
     fbb = tt_runtime.binary.load_binary_from_capsule(compiled_bin)
     program_indices = range(fbb.get_num_programs())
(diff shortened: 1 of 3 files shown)


tools/builder/base/builder_runtime.py
```diff
@@ -798,6 +804,17 @@ def execute_fb(
...
+                import sys
+
+                if "conftest" in sys.modules:
+                    device_reopener = sys.modules["conftest"].device_reopener
+                    print(f"Flatbuffer execution failed, close & reopen the device")
+                    tt_runtime.runtime.close_mesh_device(device)
+                    device = device_reopener()
+                    print(f"Device successfully reopened after execution failure")
+            except Exception as reopen_error:
+                print(f"Warning: Could not reopen device: {reopen_error}")
```
jgrimTT, line 817: There is a function in `conftest.py` (`pytest_runtest_call`) that assesses whether the test passed. Can we move the close/reopen handling to this so it doesn't involve `builder_runtime.py`?  Also because this is only in `execute_fb` and so it won't catch emitc/emitpy failures

wenbinlyuTT: I'm closing this since the original issue we tried to solve is gone now, better not make things more complicated than necessary.

wenbinlyuTT:
Closing as the P150 hang doesn't happen anymore, will revisit if something similar happens again.

vwellsTT:
hmm, might still exist? https://github.com/tenstorrent/tt-mlir/actions/runs/23505080826/job/68412863739

PR #7588, opened 2026-03-21
Support per-tensor weight dtype overrides

### LINKED ISSUE #7581: Add support for per-tensor weight dtype overrides
Many times, we want to lower dtypes of some weights in model while leaving others in higher precision. We need support for this in mlir.

dgolubovicTT:
### Ticket
Closes #7581

### Problem description
Up until now, we only had option to override weight dtype for whole model. In many cases, we want to leave some parameters in higher precision while lowering other parameters. This PR introduces support for per-tensor weight overrides for matmuls.

Note: overriding dtype of weights of convolutions is still not possible because of this [issue](https://github.com/tenstorrent/tt-mlir/issues/5995)
Overriding dtypes of non-matmul ops is not currently supported. 

### What's changed
Support per-tensor weight dtype overrides
    
Add per-tensor weight dtype overrides via "ttcore.weight_dtype" function
argument attributes, supporting BFloat16, BFP_BFloat8, and BFP_BFloat4.
Per-arg annotations take priority over the global target-dtype option.
    
A new TTIRPropagateWeightDtype pass resolves overrides by tracing each
matmul/linear weight (B) operand backward through ops with the
TensorManipulation trait to find the originating func arg annotation.
The resolved dtype is set as a discardable attribute on the op. The
TTIR-to-TTNN conversion patterns for Matmul and Linear forward this
attribute to the TTNN op.
    
The TTNN WeightDtypeConversion pass is simplified to read the dtype
directly from the op attribute.
    
Remove the deprecated enable-bfp8-conversion and experimental-bfp8-weights
pipeline options and related ElementTypeNormalization logic.

Add lit and Python builder tests for the new per-arg weight dtype behavior.

### DIFF
--- a/lib/Dialect/TTIR/Transforms/PropagateWeightDtype.cpp
+++ b/lib/Dialect/TTIR/Transforms/PropagateWeightDtype.cpp
@@ -18,11 +18,11 @@ namespace {
 // func arg. If it carries "ttcore.weight_dtype", set it on the consumer op.
 static void resolveAndSetWeightDtype(mlir::Value weight,
                                      mlir::Operation *consumerOp) {
-  // Trace backward through TM ops and BroadcastOp.
+  // Trace backward through TM ops, BroadcastOp, and MeshShardOp.
   mlir::Value source = weight;
   while (auto *op = source.getDefiningOp()) {
     if (op->hasTrait<TensorManipulation::Trait>() ||
-        mlir::isa<BroadcastOp>(op)) {
+        mlir::isa<BroadcastOp, MeshShardOp>(op)) {
       source = op->getOperand(0);
     } else {
       break;
--- a/test/ttmlir/Dialect/TTIR/propagate_weight_dtype/propagate_through_mesh_shard.mlir
+++ b/test/ttmlir/Dialect/TTIR/propagate_weight_dtype/propagate_through_mesh_shard.mlir
@@ -1,0 +1,20 @@
+// RUN: ttmlir-opt --ttir-propagate-weight-dtype %s | FileCheck %s
+
+// Test that weight_dtype propagates through a mesh_shard op to the matmul.
+// MeshShardOp does not have the TensorManipulation trait, but should still
+// be traced through since it commonly appears on weight paths in tensor
+// parallel models.
+
+module {
+  // CHECK-LABEL: func.func @propagate_through_mesh_shard
+  func.func @propagate_through_mesh_shard(
+    %arg0: tensor<32x64xbf16>,
+    %arg1: tensor<64x128xbf16> {ttcore.argument_type = #ttcore.argument_type<parameter>, ttcore.weight_dtype = "bfp_bf8"}
+  ) -> tensor<32x64xbf16> {
+    %0 = "ttir.mesh_shard"(%arg1) <{shard_dims = array<i64: -1, 1>, shard_direction = #ttcore.shard_direction<full_to_shard>, shard_shape = array<i64: 1, 2>, shard_type = #ttcore.shard_type<identity>}> : (tensor<64x128xbf16>) -> tensor<64x64xbf16>
+    // CHECK: "ttir.matmul"
+    // CHECK-SAME: ttcore.weight_dtype = "bfp_bf8"
+    %1 = "ttir.matmul"(%arg0, %0) : (tensor<32x64xbf16>, tensor<64x64xbf16>) -> tensor<32x64xbf16>
+    return %1 : tensor<32x64xbf16>
+  }
+}


lib/Dialect/TTNN/Transforms/TTNNWeightDtypeConversion.cpp
```diff
@@ -2,29 +2,98 @@
...
+// one weight, one override). Falls back to globalDtype if the chain contains
+// an unrecognized op.
+static std::optional<ttcore::DataType>
+resolveWeightDtype(mlir::Value weight,
+                   std::optional<ttcore::DataType> globalDtype) {
+  // Walk backwards through known unary transform ops to find the originating
+  // block argument.
+  mlir::Value current = weight;
+  while (auto *defOp = current.getDefiningOp()) {
+    if (mlir::isa<ReshapeOp, TypecastOp, ToLayoutOp, PermuteOp, ToDeviceOp,
```
dgolubovicTT: I know this is ugly and hard for maintenance. We should talk about introducing some trait for TM ops so that we don't have to do this. This pattern repeats on many places in repo, so it is worth discussing

lib/Dialect/TTNN/Transforms/TTNNWeightDtypeConversion.cpp
```diff
@@ -2,29 +2,98 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 
+#include "mlir/Transforms/GreedyPatternRewriteDriver.h"
```
azecevic000: Nit: move this include to the old place

lib/Dialect/TTNN/Transforms/TTNNWeightDtypeConversion.cpp
azecevic000: Yeah, I agree, this should be a relatively minor change with no downsides. You might add it even in this PR, but I'm also okay if you decide to do it subsequently.

dgolubovicTT: FYI @azecevicTT I talked with @dmilinkovicTT  and he mentioned that passing through ops like this would fail on ttcore.load_cached(). So If our weight is constevaluated this won't work. 
I added lit test with consteval and changed how we traverse the graph to accomodate for arbitrary op that calls some function. Maybe I could handle only load_cached but I think this way is more general.

With this we avoid having allow list of ops to pass through, and also handle constevals.

lib/Dialect/TTNN/Transforms/TTNNWeightDtypeConversion.cpp
```diff
@@ -17,14 +18,149 @@ namespace mlir::tt::ttnn {
...
+// Falls back to globalDtype otherwise.
+static std::optional<ttcore::DataType>
+resolveWeightDtype(mlir::Value weight,
+                   std::optional<ttcore::DataType> globalDtype) {
+  auto blockArgs = traceToFuncArgs(weight);
+
+  // Only handle 1-to-1 mapping: exactly one contributing block argument.
+  if (blockArgs.size() != 1) {
+    return globalDtype;
+  }
```
dmilinkovicTT: I think that warning log could be helpful here

test/ttmlir/Dialect/TTNN/weight_conversion/matmul_per_arg_through_load_cached.mlir
```diff
@@ -0,0 +1,34 @@
...
+
+// Test per-arg weight dtype override when the weight flows through a
+// ttcore.load_cached op into a const_eval function. The pass should trace
+// through the load_cached callee to find the originating block argument and
+// its "ttcore.weight_dtype" annotation.
+
+#dram = #ttnn.buffer_type<dram>
+
+module attributes {} {
+  // The const_eval function that processes the weight (permute + to_layout).
```
dmilinkovicTT: No `to_layout`?

lib/Dialect/TTNN/Transforms/TTNNWeightDtypeConversion.cpp
```diff
@@ -17,14 +18,149 @@ namespace mlir::tt::ttnn {
...
+  return mlir::SymbolTable::lookupNearestSymbolFrom<mlir::func::FuncOp>(
+      moduleOp, calleeAttr.getAttr());
+}
+
+// Trace a value backwards through the use-def chain to find all contributing
+// block arguments in the outermost function. Crosses function boundaries by
+// detecting call-like ops (any op with a "callee" FlatSymbolRefAttr that
+// resolves to a FuncOp): traces the corresponding return operand to its
+// contributing block arguments and maps them back to the call op's operands.
+static llvm::SetVector<mlir::BlockArgument> traceToFuncArgs(mlir::Value start) {
```
dmilinkovicTT: With this, we're back to tracing trough all ops, instead of just TMs. 

Although probably correct for most of the cases, this traversal can still hypothetically lead to multiple block arguments.

I still think that we should be explicit here about the ops we're skipping, just like with your previous approach. 

So, my proposal would be to:
- skip only TMs
- "jump" over the LoadCached boundary and enter the const-eval function
- skip TMs again

It would simplify the code, and make debugging things easier.

lib/Dialect/TTNN/Transforms/TTNNWeightDtypeConversion.cpp
```diff
@@ -17,14 +18,149 @@ namespace mlir::tt::ttnn {
...
+  if (!calleeAttr) {
+    return nullptr;
+  }
+  auto moduleOp = op->getParentOfType<mlir::ModuleOp>();
+  if (!moduleOp) {
+    return nullptr;
+  }
+  return mlir::SymbolTable::lookupNearestSymbolFrom<mlir::func::FuncOp>(
+      moduleOp, calleeAttr.getAttr());
+}
```
dmilinkovicTT: nit: although this is more general, I would prefer this to be scoped only for const-eval functions. We don't know how could semantics of some potential future function types could mess up with our logic here, and at least for now, we don't have any other function types at this stage of the pipeline.

lib/Dialect/TTNN/Transforms/TTNNWeightDtypeConversion.cpp
dmilinkovicTT: Synced offline. Current approach fails on the following example:

```
const_eval_0(weight0, weight1)
    0, 1 = cpu_hoisted_func(weight0, weight1) // Some TMs hoisted to CPU
    2 = to_device(0)
    3 = to_device(1)
    return 2, 3

main(weight0, weight1)
    0, 1 = load_cached(const_eval_0, [weight0, weight1])
    matmul(X, 0)
    matmul(X, 1)
```

While traversing, when we get to `cpu_hoisted_func`, we don't know which weight should we pick, so we pick both. Ultimately, we end up with both weights in the set of collected block arguments, and decide to skip both, since the choice is ambiguous.

We've came up with a better approach offline: 
1. Propagate weight attribute from block argument to the corresponding matmul on TTIR level
    - before const-eval - graph is simpler, we just need to start from matmuls, skip the TMs, and we will reach a single block argument
2. In TTNN pass, just apply the typecasts directly before the matmul, based on the weight attribute previosly attached to matmul

test/ttmlir/Dialect/TTNN/weight_conversion/matmul_per_arg_through_load_cached.mlir
dgolubovicTT: don't have the need for this test anymore

lib/Dialect/TTIR/Transforms/PropagateWeightDtype.cpp
```diff
@@ -0,0 +1,70 @@
...
+namespace {
+
+// Trace the weight operand backward through TM ops to find the originating
+// func arg. If it carries "ttcore.weight_dtype", set it on the consumer op.
+static void resolveAndSetWeightDtype(mlir::Value weight,
+                                     mlir::Operation *consumerOp) {
+  // Trace backward through any op with the TensorManipulation trait.
+  mlir::Value source = weight;
+  while (auto *op = source.getDefiningOp()) {
+    if (op->hasTrait<TensorManipulation::Trait>()) {
```
dmilinkovicTT: Just a heads-up: `BroadcastOp` doesn't seem to have this trait. But we could handle this separately.

lib/Dialect/TTIR/Transforms/PropagateWeightDtype.cpp
```diff
@@ -0,0 +1,70 @@
...
+  auto dtypeAttr = funcOp.getArgAttrOfType<mlir::StringAttr>(
+      blockArg.getArgNumber(), "ttcore.weight_dtype");
+  if (dtypeAttr) {
+    consumerOp->setAttr("ttcore.weight_dtype", dtypeAttr);
+  }
+}
+
+} // namespace
+
+class TTIRPropagateWeightDtype
```
dmilinkovicTT, line 52: Nit: could this pass also take into account global dtype? So that the `TTNNWeightDtypeConversion` has a single, simple responsibility: apply the typecast based on the attached attribute?

Ofc, I won't block the PR on this.

lib/Dialect/TTNN/Pipelines/TTNNPipelines.cpp
```diff
@@ -70,6 +70,9 @@ void createTTNNPipelineTTIRPasses(
   // Infer kv_cache argument types from cache operations.
   pm.addPass(mlir::tt::ttir::createTTIRInferKVCacheArgumentTypes());
 
+  // Propagate per-arg weight_dtype annotations through TM ops to consumers.
+  pm.addPass(mlir::tt::ttir::createTTIRPropagateWeightDtype());
+
```
dmilinkovicTT, line 75: This runs before `TTIRToTTIRDecomposition`, and `DotGeneral` decomposition produces `ttir::matmul`. This means that `DotGenral` weights can't be targeted. Is that okay, or should we move the pass? Or handle `DotGeneral` in the pass implementation?

dgolubovicTT: I double checked that before adding this. Seems like `createTTIRToTTIRDecompositionPass` is added before in line 56 in this file.

lib/Dialect/TTIR/Transforms/PropagateWeightDtype.cpp
dgolubovicTT: Thanks, adding handling for broadcast along with a lit test.

lib/Dialect/TTIR/Transforms/PropagateWeightDtype.cpp
dgolubovicTT: Makes sense, but I would leave it as it is for this PR. I will have some more work on this so I will consider it then.

Issue #7339, opened 2026-03-04
[EIO] Add support for keep_dim=false in ReduceCommutePatterns


mvasiljevicTT:
Currently EIO can move reshapes only through reduce ops with keep_dim=true.

In gpt oss we have reduce ops with keep_dim=false that are blocking reshape from moving:

<img width="500"  alt="Image" src="https://github.com/user-attachments/assets/64e72c87-ac19-4e50-b784-aab940a0341d" />

This is single chip issue, adding multichip label just for visibility.

mvasiljevicTT:
~~Do this only if #7358 shows that we should not force keep_dim=True~~ We should have this anyway

mvasiljevicTT:
Will be done as part of [this PR](https://github.com/tenstorrent/tt-mlir/pull/7599).

PR #7599, opened 2026-03-23
EIO - generalize Permute-Reduce commute pattern to handle keepdim=False

sgligorijevicTT:
### Ticket
No dedicated issue, while testing out my PR in XLA https://github.com/tenstorrent/tt-xla/pull/3795 I noticed it would introduce a pref regression, specifically ~10% in Mobilenetv2. With this change I not only recover the old perf but I seem to get slightly better numbers.
EDIT:
I was made aware there is an issue for this after all, closes https://github.com/tenstorrent/tt-mlir/issues/7339

### Problem description
CommuteOpRewritePattern for Permute-Reduce pair has a check that `keepDim==true` inside `Viable()` functions. This is leaving potential reorderings on the table.

### What's changed
More code was added to handle dims being removed.

### Checklist

### DIFF
--- a/lib/Dialect/TTIR/Transforms/EraseInverseOps/ReduceCommutePatterns.cpp
+++ b/lib/Dialect/TTIR/Transforms/EraseInverseOps/ReduceCommutePatterns.cpp
@@ -114,23 +114,27 @@ private:
     return rewriter.create<PermuteOp>(loc, outputType, input, permutation);
   }
 
-  ArrayAttr permuteDims(ArrayAttr dimArg, ArrayRef<int64_t> permutation,
+  ArrayAttr permuteDims(std::optional<ArrayAttr> dimArg,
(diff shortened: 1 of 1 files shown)


mvasiljevicTT:
Thank you for doing this! We had an [issue opened](https://github.com/tenstorrent/tt-mlir/issues/7339) for this that wasn't picked yet :D

sgligorijevicTT:
Perf benchmark is not clean - https://github.com/tenstorrent/tt-xla/actions/runs/23487047238

sgligorijevicTT:
https://github.com/tenstorrent/tt-xla/actions/runs/23597879570 Turns out the performance failures I was seeing are probably not related, running a perf test workflow with this code rebased on top of what was just uplifted onto xla main passes. This is going in.

Issue #7535, opened 2026-03-17
Constant folding for common tensor manipulation operators


amalkovTT:
We should implement constant folding for common tensor manipulation operators like `ttir.reshape`, `ttir.permute`, `ttir.broadcast`, `ttir.repeat`, `ttir.repeat_interleave`.

PR #7601, opened 2026-03-23
Implement constant folding for some TM operators

amalkovTT:
### Ticket
#7535

### Problem description
Common tensor manipulation operators (ttir.reshape, ttir.broadcast, ttir.repeat, ttir.repeat_interleave, ttir.permute, ttir.slice_static) should be folded when their input is constant.

### What's changed
Fold methods of these operators were implemented/upgraded to do constant folding.

To fix regressions caused by these folders, gelu fusing was changed to work when it finds `ttir.ones` instead of `ttir.full` with `fill_value = 1`. 

`ElementwiseCommutePatterns.cpp` was changed to commute downwards even when one of the operands comes from a `ConstantLike` operator. This fixes the failed test where reshape operand was duplicated, because folding prevented commuting.

The logic that converts a `ConstantOp`'s `value` to a `FullOp`'s `fill_value` has been moved to also be used in constant materialization. This fixes the issue where a `ConstantOp` with a splat value was materialized after the last canonicalization pass, and then wasn't supported by the EmitC backend.

### Checklist

### DIFF
--- a/lib/Dialect/TTIR/IR/TTIROps.cpp
+++ b/lib/Dialect/TTIR/IR/TTIROps.cpp
@@ -57,7 +57,7 @@ namespace mlir::tt::ttir {
 //===----------------------------------------------------------------------===//
 
 // Reshape attribute if it is splat otherwise return nullptr.
-static DenseElementsAttr reshapeIfSplat(RankedTensorType type, Attribute attr) {
+static DenseElementsAttr reshapeIfSplat(ShapedType type, Attribute attr) {
   if (auto splat = llvm::dyn_cast<SplatElementsAttr>(attr)) {
     return splat.resizeSplat(type);
   }
@@ -77,12 +77,13 @@ static bool shouldFold(mlir::Operation *op) {
   return ttmlir::utils::volume(shapedType.getShape()) <= foldLimit;
 }
 
+// Helper to fold a tensor manipulation operation with a non-splat constant
+// argument using an index mapping function. The index mapping function takes
+// output coordinates and returns input coordinates.
 template <typename ElementType, typename Fun>
-static ::mlir::OpFoldResult foldNonSplatTMImpl(mlir::Operation *op,
-                                               DenseElementsAttr inputAttr,
-                                               Fun indexMap) {
-  ShapedType resultType = mlir::cast<ShapedType>(op->getResult(0).getType());
-
+static ::mlir::OpFoldResult foldNonSplatTM(ShapedType resultType,
+                                           DenseElementsAttr inputAttr,
+                                           Fun indexMap) {
   auto inputValues = inputAttr.getValues<ElementType>();
   llvm::SmallVector<ElementType> outputValues;
   outputValues.reserve(resultType.getNumElements());
@@ -96,7 +97,9 @@ static ::mlir::OpFoldResult foldNonSplatTMImpl(mlir::Operation *op,
     int64_t inputIndex = mlir::linearize(inputCoord, inputStrides);
     outputValues.push_back(inputValues[inputIndex]);
 
-    // Increment output coordinates
+    // Increment output coordinates in row-major order. Increment the innermost
+    // dimension; if it reaches the dimension size, reset it to 0 and carry into
+    // the next outer dimension.
     for (int64_t dim = outputShape.size() - 1; dim >= 0; --dim) {
       if (++outputCoord[dim] < outputShape[dim]) {
         break;
@@ -108,26 +111,34 @@ static ::mlir::OpFoldResult foldNonSplatTMImpl(mlir::Operation *op,
   return mlir::DenseElementsAttr::get(resultType, outputValues);
 }
 
-// Helper to fold a constant tensor manipulation operation with a non-splat
-// argument using an index mapping function. The index mapping function takes
-// output coordinates and returns input coordinates.
+// Helper to fold a constant tensor manipulation operation using an index
+// mapping function. The index mapping function takes output coordinates and
+// returns input coordinates.
 template <typename Fun>
 static ::mlir::OpFoldResult
-foldNonSplatTM(mlir::Operation *op, mlir::Attribute inputAttr, Fun indexMap) {
-  auto denseAttr =
-      llvm::dyn_cast_if_present<mlir::DenseElementsAttr>(inputAttr);
-  if (!denseAttr || !shouldFold(op)) {
+constantFoldTM(mlir::Operation *op, mlir::Attribute inputAttr, Fun indexMap) {
+  if (!inputAttr) {
     return nullptr;
   }
 
   ShapedType resultType = mlir::cast<ShapedType>(op->getResult(0).getType());
-  if (resultType.getElementType().isFloat()) {
-    return foldNonSplatTMImpl<llvm::APFloat>(op, denseAttr, indexMap);
-  }
-  if (resultType.getElementType().isInteger()) {
-    return foldNonSplatTMImpl<llvm::APInt>(op, denseAttr, indexMap);
+  if (auto foldResult = reshapeIfSplat(resultType, inputAttr)) {
+    return foldResult;
   }
 
+  if (!shouldFold(op)) {
+    return nullptr;
+  }
+
+  if (auto denseAttr =
+          llvm::dyn_cast_if_present<mlir::DenseElementsAttr>(inputAttr)) {
+    if (resultType.getElementType().isFloat()) {
+      return foldNonSplatTM<llvm::APFloat>(resultType, denseAttr, indexMap);
+    }
+    if (resultType.getElementType().isInteger()) {
+      return foldNonSplatTM<llvm::APInt>(resultType, denseAttr, indexMap);
+    }
+  }
   return nullptr;
 }
 
@@ -2745,18 +2756,6 @@ constantFoldSliceStatic(mlir::tt::ttir::SliceStaticOp op,
     return nullptr;
   }
 
-  if (auto foldResult = reshapeIfSplat(op.getResult().getType(), constInput)) {
-    return foldResult;
-  }
-
-  if (op.getResult().hasOneUse() &&
-      isa<mlir::tt::ttir::SliceStaticOp>(
-          op.getResult().use_begin()->getOwner())) {
-    // Don't fold if the result is consumed by another SliceStaticOp, as that
-    // would prevent efficient folding of consecutive SliceStaticOps.
-    return nullptr;
-  }
-
   auto inputShape = op.getInput().getType().getShape();
   llvm::SmallVector<int64_t> begins(inputShape.size());
   llvm::SmallVector<int64_t> step(inputShape.size());
@@ -2766,19 +2765,16 @@ constantFoldSliceStatic(mlir::tt::ttir::SliceStaticOp op,
     // Adjust negative begin.
     begins[i] = (begin < 0) ? (begin + inputShape[i]) : begin;
   }
-  if (auto foldResult = foldNonSplatTM(
-          op, constInput,
-          [&begins, &step](const llvm::SmallVector<int64_t> &outputCoord) {
-            llvm::SmallVector<int64_t> inputCoord(outputCoord.size());
-            for (size_t i = 0; i < inputCoord.size(); ++i) {
-              inputCoord[i] = begins[i] + step[i] * outputCoord[i];
-            }
-            return inputCoord;
-          })) {
-    return foldResult;
-  }
 
-  return nullptr;
+  return constantFoldTM(
+      op, constInput,
+      [&begins, &step](const llvm::SmallVector<int64_t> &outputCoord) {
+        llvm::SmallVector<int64_t> inputCoord(outputCoord.size());
+        for (size_t i = 0; i != inputCoord.size(); ++i) {
+          inputCoord[i] = begins[i] + step[i] * outputCoord[i];
+        }
+        return inputCoord;
+      });
 }
 
 // SliceStaticOp Folder
@@ -4461,28 +4457,15 @@ static mlir::OpFoldResult foldIdentityRepeat(mlir::tt::ttir::RepeatOp op) {
 
 static mlir::OpFoldResult constantFoldRepeat(mlir::tt::ttir::RepeatOp op,
                                              mlir::Attribute input) {
-  if (!input) {
-    return nullptr;
-  }
-
-  if (auto foldResult = reshapeIfSplat(op.getResult().getType(), input)) {
-    return foldResult;
-  }
-
-  auto inputShape = op.getInput().getType().getShape();
-  if (auto foldResult = foldNonSplatTM(
-          op, input,
-          [inputShape](const llvm::SmallVector<int64_t> &outputCoords) {
-            llvm::SmallVector<int64_t> inputCoords(outputCoords.size());
-            std::transform(outputCoords.begin(), outputCoords.end(),
-                           inputShape.begin(), inputCoords.begin(),
-                           std::modulus<int64_t>());
-            return inputCoords;
-          })) {
-    return foldResult;
-  }
-
-  return nullptr;
+  llvm::ArrayRef<int64_t> inputShape = op.getInput().getType().getShape();
+  return constantFoldTM(
+      op, input, [inputShape](const llvm::SmallVector<int64_t> &outputCoords) {
+        llvm::SmallVector<int64_t> inputCoords(outputCoords.size());
+        std::transform(outputCoords.begin(), outputCoords.end(),
+                       inputShape.begin(), inputCoords.begin(),
+                       std::modulus<int64_t>());
+        return inputCoords;
+      });
 }
 
 // RepeatOp Folder
@@ -4560,31 +4543,20 @@ foldIdentityRepeatInterleave(mlir::tt::ttir::RepeatInterleaveOp op) {
 static mlir::OpFoldResult
 constantFoldRepeatInterleave(mlir::tt::ttir::RepeatInterleaveOp op,
                              mlir::Attribute input) {
-  if (!input) {
-    return nullptr;
-  }
-
-  if (auto foldResult = reshapeIfSplat(op.getResult().getType(), input)) {
-    return foldResult;
-  }
-
-  if (auto foldResult = foldNonSplatTM(
-          op, input,
-          [dim = op.getDim(), repeats = op.getRepeats()](
-              const llvm::SmallVector<int64_t> &outputCoords) {
-            llvm::SmallVector<int64_t> inputCoords(outputCoords.size());
-            for (size_t i = 0; i < outputCoords.size(); ++i) {
-              inputCoords[i] = outputCoords[i];
-              if (i == static_cast<size_t>(dim)) {
-                inputCoords[i] /= repeats;
-              }
-            }
-            return inputCoords;
-          })) {
-    return foldResult;
-  }
-
-  return nullptr;
+  int32_t dim = op.getDim();
+  uint32_t repeats = op.getRepeats();
+  return constantFoldTM(
+      op, input,
+      [dim, repeats](const llvm::SmallVector<int64_t> &outputCoords) {
+        llvm::SmallVector<int64_t> inputCoords(outputCoords.size());
+        for (size_t i = 0; i != outputCoords.size(); ++i) {
+          inputCoords[i] = outputCoords[i];
+          if (i == static_cast<size_t>(dim)) {
+            inputCoords[i] /= repeats;
+          }
+        }
+        return inputCoords;
+      });
 }
 
 // RepeatInterleaveOp Folder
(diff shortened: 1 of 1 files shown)


lib/Dialect/TTIR/IR/TTIROps.cpp
```diff
@@ -2421,13 +2459,109 @@ foldConsecutiveSliceStatic(mlir::tt::ttir::SliceStaticOp consumerOp) {
...
+    if (std::reduce(outputShape.begin(), outputShape.end(), 1,
+                    std::multiplies<int64_t>()) > maxElements) {
+      // Avoid folding if the number of elements is too large, to prevent
+      // excessive compile time and memory usage.
+      return nullptr;
+    }
+
+    if (denseAttr.getElementType().isFloat()) {
+      return constantFoldNonSplatSliceStatic<mlir::APFloat>(op, denseAttr);
+    } else if (denseAttr.getElementType().isInteger()) {
```
vmilosevic: :warning: **llvm\-else\-after\-return** :warning:
```suggestion
    } if (denseAttr.getElementType().isInteger()) {
```

amalkovTT:
SDPA fusing tests are failing, but they will pass after I rebase to main after #7692 is merged.

lib/Dialect/TTIR/IR/TTIROps.cpp
```diff
@@ -2555,13 +2583,110 @@ foldConsecutiveSliceStatic(mlir::tt::ttir::SliceStaticOp consumerOp) {
...
+  llvm::SmallVector<ElemType> outputValues;
+
+  auto it = inputValues.begin();
+  it += startPos;
+  while (true) {
+    outputValues.push_back(*it);
+    int64_t dim = currIndex.size() - 1;
+    while (dim >= 0 &&
+           ((step[dim] > 0 && currIndex[dim] + step[dim] >= ends[dim]) ||
+            (step[dim] < 0 && currIndex[dim] + step[dim] <= ends[dim]))) {
```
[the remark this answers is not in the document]
amalkovTT: Since the case when the tensor is empty does not depend on the element type, it is handled before calling this function template. I added a comment and an assertion to document that this function is never called with an empty attribute.

lib/Dialect/TTIR/IR/TTIROps.cpp
```diff
@@ -2555,13 +2583,110 @@ foldConsecutiveSliceStatic(mlir::tt::ttir::SliceStaticOp consumerOp) {
...
+  llvm::SmallVector<int64_t> step(inputShape.size());
+  llvm::SmallVector<int64_t> currIndex(inputShape.size());
+  int64_t startPos = 0;
+  for (size_t i = 0; i < currIndex.size(); ++i) {
+    begins[i] = mlir::cast<mlir::IntegerAttr>(op.getBegins()[i]).getInt();
+    ends[i] = mlir::cast<mlir::IntegerAttr>(op.getEnds()[i]).getInt();
+    step[i] = mlir::cast<mlir::IntegerAttr>(op.getStep()[i]).getInt();
+    currIndex[i] = begins[i];
+    startPos += currIndex[i] * iterStepSize[i];
+    assert(iterStepSize[i] != 0 && step[i] != 0 && "Step size cannot be zero");
```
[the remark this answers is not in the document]
amalkovTT: Fixed.

lib/Dialect/TTIR/IR/TTIROps.cpp
```diff
@@ -2555,13 +2583,119 @@ foldConsecutiveSliceStatic(mlir::tt::ttir::SliceStaticOp consumerOp) {
...
+                                          llvm::ArrayRef<mlir::Attribute>{});
+    }
+    llvm::ArrayRef<int64_t> outputShape = op.getType().getShape();
+    constexpr int64_t maxElements = 1'000'000;
+    if (std::reduce(outputShape.begin(), outputShape.end(), 1,
+                    std::multiplies<int64_t>()) > maxElements) {
+      // Avoid folding if the number of elements is too large, to prevent
+      // excessive compile time and memory usage.
+      return nullptr;
+    }
```
dmilinkovicTT: We should extract this threshold logic, so that it can be reused in multiple patterns.

lib/Dialect/TTIR/IR/TTIROps.cpp
```diff
@@ -2555,13 +2583,119 @@ foldConsecutiveSliceStatic(mlir::tt::ttir::SliceStaticOp consumerOp) {
...
+    return foldSplat;
+  }
+
+  if (op.getResult().hasOneUse() &&
+      isa<mlir::tt::ttir::SliceStaticOp>(
+          op.getResult().use_begin()->getOwner())) {
+    // Don't fold if the result is consumed by another SliceStaticOp, as that
+    // would prevent folding of consecutive SliceStaticOps.
+    return nullptr;
+  }
```
dmilinkovicTT: But won't the other slice also be folded?

lib/Dialect/TTIR/IR/TTIROps.cpp
```diff
@@ -4234,6 +4368,10 @@ mlir::OpFoldResult mlir::tt::ttir::RepeatOp::fold(FoldAdaptor fold) {
   if (auto foldResult = foldConsecutiveRepeat(*this)) {
     return foldResult;
   }
+  if (auto foldResult =
+          reshapeIfSplatAndPresent(getResult().getType(), fold.getInput())) {
+    return foldResult;
+  }
```
dmilinkovicTT: What about non-splats? Shouldn't we also apply the logic of folding if the resulting ElementsAttr's size doesn't exceed a certain threshold?

lib/Dialect/TTIR/IR/TTIROps.cpp
```diff
@@ -4285,6 +4423,28 @@ ::mlir::LogicalResult mlir::tt::ttir::RepeatInterleaveOp::verify() {
...
+
+// RepeatInterleaveOp Folder
+mlir::OpFoldResult mlir::tt::ttir::RepeatInterleaveOp::fold(FoldAdaptor fold) {
+  if (auto foldResult = foldIdentityRepeatInterleave(*this)) {
+    return foldResult;
+  }
+  if (auto foldResult =
+          reshapeIfSplatAndPresent(getResult().getType(), fold.getInput())) {
+    return foldResult;
+  }
```
dmilinkovicTT: Same here

lib/Dialect/TTIR/IR/TTIRDialect.cpp
```diff
@@ -144,19 +145,19 @@ ::mlir::Operation *TTIRDialect::materializeConstant(OpBuilder &builder,
...
       }
       if (isOneAttr(splatValue)) {
         return builder.create<ttir::OnesOp>(loc, type, shape);
       }
-      if (elementsAttr.getElementType().isF32() ||
-          elementsAttr.getElementType().isSignlessInteger(32)) {
-        return builder.create<ttir::FullOp>(loc, type, shape, splatValue);
+
+      if (auto fillValueAttr = utils::splatToFillValue(builder, elementsAttr)) {
+        return builder.create<ttir::FullOp>(loc, type, shape, fillValueAttr);
```
dmilinkovicTT, line 163: We should add a comment which explains why this is safe to do.

include/ttmlir/Dialect/TTIR/Utils/Utils.h
```diff
@@ -419,6 +419,29 @@ inline mlir::Value reshapeAndCastToType(mlir::PatternRewriter &rewriter,
   return result;
 }
 
+// Converts a splat ElementsAttr to an i32/f32 attribute that can be used as a
+// fill value in ttir::FullOp. Returns nullptr on failure.
+inline mlir::Attribute splatToFillValue(mlir::OpBuilder &builder,
+                                        mlir::ElementsAttr valueAttr) {
+  if (!valueAttr.isSplat()) {
+    return nullptr;
+  }
```
dmilinkovicTT, line 430: This function can just receive `SplatElementsAttr`

include/ttmlir/Dialect/TTIR/Utils/Utils.h
```diff
@@ -419,6 +419,29 @@ inline mlir::Value reshapeAndCastToType(mlir::PatternRewriter &rewriter,
...
+
+  if (auto integerType =
+          mlir::dyn_cast<mlir::IntegerType>(valueAttr.getElementType())) {
+    auto fillValue = valueAttr.getSplatValue<llvm::APInt>();
+    if (integerType.isSigned()) {
+      return builder.getI32IntegerAttr(fillValue.getSExtValue());
+    }
+    return builder.getI32IntegerAttr(fillValue.getZExtValue());
+  }
+  if (valueAttr.getElementType().isIntOrFloat()) {
```
dmilinkovicTT: Can we use `isFloat()` here?

lib/Dialect/TTIR/IR/TTIROps.cpp
amalkovTT: Yes, but folding of consecutive slice operations was already implemented, and if the input is large, it can be more efficient to fold consecutive slice operations using that function, and to perform constant folding on the single slice in a later iteration.

include/ttmlir/Dialect/TTIR/Utils/Utils.h
amalkovTT: Yes, but this is the code from `ConstantOp` canonicalization which I just moved to another file so that I can reuse it in `TTIRDialect.cpp`, so I didn't want to change it if not necessary.

lib/Dialect/TTIR/IR/TTIROps.cpp
```diff
@@ -277,6 +283,33 @@ getShapeAsI32(mlir::RankedTensorType tensorType) {
...
+// to avoid dramatically increasing compile time and memory usage.
+static bool shouldFold(mlir::Operation *op) {
+  constexpr int64_t foldLimit = 1'000'000;
+  mlir::Type resultType = op->getResult(0).getType();
+  auto shapedType = mlir::dyn_cast<mlir::ShapedType>(resultType);
+  if (!shapedType) {
+    return false;
+  }
+  llvm::ArrayRef<int64_t> shape = shapedType.getShape();
+  if (std::reduce(shape.begin(), shape.end(), 1, std::multiplies<int64_t>()) >
```
dmilinkovicTT: IIRC, utils has `volume(shape)` helper

lib/Dialect/TTIR/IR/TTIRDialect.cpp
dmilinkovicTT: This doesn't seem to be resolved?

lib/Dialect/TTIR/IR/TTIROps.cpp
```diff
@@ -410,6 +443,14 @@ ::mlir::OpFoldResult mlir::tt::ttir::BroadcastOp::fold(FoldAdaptor adaptor) {
...
     return getInput();
   }
+
+  // If the input is constant and splat, perform constant folding.
+  Attribute constInput = adaptor.getInput();
+  RankedTensorType resultType = getResult().getType();
+  if (auto foldResult = reshapeIfSplatAndPresent(resultType, constInput)) {
+    return foldResult;
+  }
+
```
dmilinkovicTT: Can we perform folding for non-splats with size heuristics here?

lib/Dialect/TTIR/IR/TTIROps.cpp
```diff
@@ -788,37 +829,27 @@ void mlir::tt::ttir::ConstantOp::getCanonicalizationPatterns(
...
-    auto valueAttr = op.getValueAttr();
-    if (!valueAttr.isSplat()) {
-      return failure();
-    }
+  patterns.add(
+      +[](mlir::tt::ttir::ConstantOp op, mlir::PatternRewriter &rewriter) {
+        auto valueAttr = op.getValueAttr();
+        if (!valueAttr.isSplat()) {
+          return failure();
+        }
```
dmilinkovicTT, line 875: Has this just been reformatted? We should avoid such changes as much as we can, since they make the code review process more complicated

lib/Dialect/TTIR/IR/TTIROps.cpp
```diff
@@ -2630,6 +2681,103 @@ foldSliceOfConcat(mlir::tt::ttir::SliceStaticOp sliceOp) {
...
+  assert(denseAttr && !denseAttr.isSplat() && !denseAttr.empty() &&
+         "Expected a non-splat, non-empty dense attribute");
+
+  llvm::ArrayRef<int64_t> inputShape = op.getInput().getType().getShape();
+  // Calculate step size for iterating over any dimension.
+  llvm::SmallVector<int64_t> iterStepSize(inputShape.size());
+  std::partial_sum(inputShape.rbegin(), std::prev(inputShape.rend()),
+                   std::next(iterStepSize.rbegin()),
+                   std::multiplies<int64_t>());
+  iterStepSize.back() = 1;
```
dmilinkovicTT: We can maybe use `calculateStrides` from `Utils.h`?

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
```diff
@@ -2612,6 +2604,30 @@ class GeluFusionPattern : public mlir::OpRewritePattern<ttir::MultiplyOp> {
     }
     return mlir::dyn_cast_if_present<ttir::FullOp>(currentOp);
   }
+
+  // Same as `getFullOpThroughTMChain` but get OnesOp or FullOp with fill_value
+  // close to one.
+  Operation *getOneThroughTMChain(Value value) const {
```
dmilinkovicTT: IMHO, this introduces unnecessary duplication. We should refactor `getFullOpThroughTMChain` and/or `isScalarValue` so that they target `Zeros`, `Ones` and `Full`. Wdyt?

Generally, a great skill to develop is to always attempt to refactor existing code instead of introducing more code for new use-cases. Of course, this is often significantly trickier to perform compared to just adding new code, and regressions are sometimes hard to catch, but is nevertheless a good practice which keeps the code cleaner in most of the cases. :)

lib/Dialect/TTIR/IR/TTIROps.cpp
```diff
@@ -5090,6 +5374,80 @@ static mlir::OpFoldResult foldConsecutivePermute(mlir::tt::ttir::PermuteOp op) {
...
+constantFoldNonSplatPermute(mlir::tt::ttir::PermuteOp op,
+                            mlir::DenseElementsAttr inputElements) {
+  RankedTensorType resultType = op.getResult().getType();
+  llvm::ArrayRef<int64_t> outputShape = resultType.getShape();
+
+  // Calculate step size for iterating over any dimension.
+  llvm::SmallVector<int64_t> stepSize(outputShape.size());
+  std::partial_sum(outputShape.rbegin(), std::prev(outputShape.rend()),
+                   std::next(stepSize.rbegin()), std::multiplies<int64_t>());
+  stepSize.back() = 1;
```
dmilinkovicTT: Same here - `ttmlir::utils::calculateStrides` should be sufficient

lib/Dialect/TTIR/IR/TTIROps.cpp
amalkovTT: Yes, it was done by pre-commit git hook.

lib/Dialect/TTIR/IR/TTIRDialect.cpp
amalkovTT: It's done now.

dmilinkovicTT (approved):

amalkovTT:
Passed [uplift test](https://github.com/tenstorrent/tt-xla/actions/runs/23941744682/job/69830465028).

lib/Dialect/TTIR/IR/TTIROps.cpp
```diff
@@ -40,9 +40,15 @@
 #include "mlir/IR/Value.h"
 #include "llvm/ADT/STLExtras.h"
 #include "llvm/Support/ErrorHandling.h"
+#include <algorithm>
```
azecevic000, line 40: Nit: add an empty line that separates LLVM includes from STL includes. Also check if all of these includes are needed.

lib/Dialect/TTIR/IR/TTIROps.cpp
```diff
@@ -410,6 +438,14 @@ ::mlir::OpFoldResult mlir::tt::ttir::BroadcastOp::fold(FoldAdaptor adaptor) {
                    [](const int32_t dim) { return dim == 1; })) {
     return getInput();
   }
+
```
azecevic000: I know it's not the change introduced in this PR, but I only see now that we have two separate sections for `BroadcastOp` (one with fold and one with verify). Can you please merge them as part of this PR.

Also, it makes sense to separate these two folding patterns into separate functions, as is done for other folding methods.

lib/Dialect/TTIR/IR/TTIROps.cpp
```diff
@@ -2017,6 +2046,19 @@ static mlir::OpFoldResult foldConsecutiveReshape(mlir::tt::ttir::ReshapeOp op) {
...
 }
 
+// Fold reshape if input is constant
+static mlir::OpFoldResult constFoldReshape(mlir::tt::ttir::ReshapeOp op,
+                                           Attribute constInput) {
+  if (auto denseAttr = dyn_cast_if_present<DenseElementsAttr>(constInput)) {
+    RankedTensorType type = op.getResult().getType();
+    if (denseAttr.isSplat()) {
+      return denseAttr.resizeSplat(type);
+    }
```
azecevic000: Is this necessary, i.e. would `reshape` work with splat as well?

lib/Dialect/TTIR/IR/TTIROps.cpp
```diff
@@ -2630,6 +2676,99 @@ foldSliceOfConcat(mlir::tt::ttir::SliceStaticOp sliceOp) {
   return nullptr;
 }
 
+// Constant fold SliceStaticOp when the input is a constant, non-splat and
+// non-empty DenseElementsAttr.
+template <typename ElemType>
+static mlir::OpFoldResult
+constantFoldNonSplatSliceStatic(mlir::tt::ttir::SliceStaticOp op,
```
azecevic000: One idea for constant folding of TMs (specifically for non-splat tensors). Since the attribute is a linear buffer, they can all follow the same algorithmic approach:
```
outputBuffer[i] = inputBuffer[indexMap(i)]
```
`indexMap` will be different for various ops (for example, for `reshape` it's just identity, that's probably why it's already implemented for us), but it's always going to be some pattern of:
```
outputMultiDimIndex = linearToMultiDim(i, outputShape);
inputMultiDimIndex = /*op specific that manipulates how output maps to input*/;
outputBuffer[i] = inputBuffer[multiDimToLinear(inputMultiDimIndex, inputShape)];
```

This approach has tremendous benefit in that it's easy to follow and prove correctness, on top of having a unifying platform for all TM foldings (this can be even codified in the sense that previous block can be a function that takes a function as a parameter for specific op (for second line), but even copy-pasting can be benefical instead of having an ad-hoc approach for each op).

lib/Dialect/TTIR/IR/TTIROps.cpp
amalkovTT: You're right, I checked and their implementations are almost identical and `resizeSplat` is only needed if the output type doesn't need to have the same number of elements.

lib/Dialect/TTIR/IR/TTIROps.cpp
```diff
@@ -15,40 +14,123 @@
...
+  mlir::Type resultType = op->getResult(0).getType();
+  auto shapedType = mlir::dyn_cast<mlir::ShapedType>(resultType);
+  if (!shapedType) {
+    return false;
+  }
+  return ttmlir::utils::volume(shapedType.getShape()) <= foldLimit;
+}
+
+template <typename ElementType, typename Fun>
+static ::mlir::OpFoldResult foldNonSplatTMImpl(mlir::Operation *op,
```
azecevic000: The only thing used from `Operation *` is the output shape, so you might as well have `(llvm::ArrayRef<int64_t> outputShape, DenseElementsAttr inputAttr, Fun indexMap)` as input arguments, especially considering the call chain that reaches this function.

lib/Dialect/TTIR/IR/TTIROps.cpp
```diff
@@ -15,40 +14,123 @@
...
+    outputValues.push_back(inputValues[inputIndex]);
+
+    // Increment output coordinates
+    for (int64_t dim = outputShape.size() - 1; dim >= 0; --dim) {
+      if (++outputCoord[dim] < outputShape[dim]) {
+        break;
+      }
+      outputCoord[dim] = 0;
+    }
+  }
```
azecevic000, line 109: This seems correct to me, but it wouldn't hurt to add a few lines of explanation; it's a non-trivial algorithm.

lib/Dialect/TTIR/IR/TTIROps.cpp
```diff
@@ -5168,6 +5325,39 @@ static mlir::OpFoldResult foldConsecutivePermute(mlir::tt::ttir::PermuteOp op) {
...
+  if (auto foldResult = reshapeIfSplat(result.getType(), input)) {
+    return foldResult;
+  }
+
+  if (result.hasOneUse() &&
+      llvm::isa<ttir::PermuteOp>(result.use_begin()->getOwner()) &&
+      !op->hasAttr("decomposed")) {
+    // Don't constant fold yet if folding of consecutive permutes is possible
+    return nullptr;
+  }
```
azecevic000: Is this just for optimization reasons? If we have `constant -> permute -> permute` we do `constant -> permute` and then constant folding? For very large dense tensors this will make a difference, but if we haven't seen such case seems like a premature optimization to me tbh.

lib/Dialect/TTIR/IR/TTIROps.cpp
amalkovTT: I change it, although not to shape, but to shaped type since it's also used (it could also be deduced from the shape and the attribute type, but it's simpler like this).

lib/Dialect/TTIR/IR/TTIROps.cpp
amalkovTT: Yes, that was the idea. I removed it now, and I also removed the same optimization in the other function.

PR #7615, opened 2026-03-23
[D2M] Add native D2M support for `ttir.min` reduction

sgholamiTT:
## Summary

- Enable `min` reduction through the D2M path by decomposing `ttir.min` into `neg(max(neg(x)))` via a dedicated `TTIRDecomposeMinReduction` pass, reusing the existing `tile_reduce_max` op.
- The decomposition lives in its own pass (`--ttir-decompose-min-reduction`) following the pattern of other standalone decompositions (`TTIRDecomposeComplexPermute`, `TTIRDecomposeComplexReshape`). It is only needed by the TTMetal pipeline — the TTNN backend has native `ttnn.min` support.
- Add MLIR lit tests for min decomposition (2D and 3D, all reduction dims, OOB correctness) and golden Python tests (`test_min`, `test_min_unaligned`).
- Enable `ttnn.min` in `ttnn-jit` smoketest and nightly reduction tests (previously skipped).

## What is not working / skipped

Most L1 block-sharded `ttnn.min` JIT tests are skipped. Only four (shape, grid) combinations pass: `(32x32, 1x1)`, `(512x1024, 8x1)`, `(64x32, 1x2)`, and `(1024x1024, 1x8)`. All other L1 block-sharded min configurations fail with low PCC because the `neg → max → neg` decomposition produces intermediate TTNN-typed tensors that go through the TTNN-to-metal roundtrip (`TTNNMetalLayoutCastOp → ViewLayoutOp`). In this path, the max reduction's input retains `Undef` OOB (TTNN's default) instead of `NegInf`, so uninitialized accumulator/padding values corrupt the reduction result. DRAM min works correctly for all shapes and grids.
This issue summarizes what I've tried and a comprehensive list of other options that could be tried: https://github.com/tenstorrent/tt-mlir/issues/7617

## Test plan

- [x] `llvm-lit test/ttmlir/Conversion/TTIRToD2M/` — 11/11 pass
- [x] `llvm-lit` on TTNN pipeline min tests — pass (simple_reduce_min)
- [x] `pytest test/ttnn-jit/test_reduction_smoketest.py` — all pass/skip
- [x] `pytest test/ttnn-jit/nightly/test_reduction.py` — all pass/skip
- [x] `pytest test/python/golden/test_metal_reductions.py -k test_min` — pass

### DIFF
--- a/lib/Dialect/D2M/IR/D2MOps.cpp
+++ b/lib/Dialect/D2M/IR/D2MOps.cpp
@@ -513,18 +513,6 @@ struct ToLayoutFoldOOBUndefPattern : public OpRewritePattern<ToLayoutOp> {
       return failure();
     }
 
-    // Don't fold when the input is undef but the output requires a specific
-    // reduction identity (neginf/inf).  Undef padding is typically zero,
-    // which is NOT the correct identity for max (neginf) or min (inf)
-    // reductions.  Folding here would let the reduction read zero-valued
-    // padding instead of the required identity, corrupting the result for
-    // tensors whose logical shape is not tile-aligned.
-    if (inputLayout.getOobVal() == ttcore::OOBVal::Undef &&
-        (outputLayout.getOobVal() == ttcore::OOBVal::NegInf ||
-         outputLayout.getOobVal() == ttcore::OOBVal::Inf)) {
-      return failure();
-    }
-
     // Layouts match except OOB — the to_layout is a no-op.
     rewriter.replaceOp(op, op.getInput());
     return success();
--- a/lib/Dialect/TTIR/Transforms/DecomposeMinReduction.cpp
+++ b/lib/Dialect/TTIR/Transforms/DecomposeMinReduction.cpp
@@ -1,4 +1,4 @@
-// SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
+// SPDX-FileCopyrightText: (c) 2026 Tenstorrent AI ULC
 //
 // SPDX-License-Identifier: Apache-2.0
 
@@ -14,11 +14,10 @@ namespace {
 // Decompose min(x) into neg(max(neg(x))) so that backends without native
 // reduce_min (e.g. TTMetal/D2M) can reuse the existing reduce_max tile op.
 //
-// This is a standalone pass rather than part of TTIRToTTIRDecomposition because
-// it is only needed by the TTMetal backend — the TTNN backend has native
-// ttnn.min support.  Keeping it separate avoids adding a backend-specific
-// config mode to the shared decomposition pass and lets each pipeline opt in
-// by simply including this pass.
+// This is a standalone pass following the pattern of other isolated
+// decompositions (TTIRDecomposeComplexPermute, TTIRDecomposeComplexReshape).
+// It is only needed by the TTMetal pipeline — the TTNN backend has native
+// ttnn.min support.
 class TTIRDecomposeMinReduction
     : public impl::TTIRDecomposeMinReductionBase<TTIRDecomposeMinReduction> {
 public:
--- a/test/ttnn-jit/test_reduction_smoketest.py
+++ b/test/ttnn-jit/test_reduction_smoketest.py
@@ -30,6 +30,16 @@ REDUCTION_OPS = [
     ttnn.min,
 ]
 
+# L1 block-sharded min has PCC issues due to OOB padding values in the neg→max→neg decomposition.
+# Also the DRAM test fails with operand #0 does not dominate this use.
+# TODO(sgholami): Investigate L1 block-sharded min PCC failures. Issue: #7617
+SKIP_MIN_TESTS = [
+    (ttnn.BufferType.L1, (128, 128), (0, 0)),
+    (ttnn.BufferType.L1, (256, 256), (0, 0)),
+    (ttnn.BufferType.L1, (128, 64), (1, 0)),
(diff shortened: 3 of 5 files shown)


test/ttnn-jit/test_reduction_smoketest.py
```diff
@@ -41,8 +41,21 @@ def test_reductions(device, shape, max_grid, dim, op, dtype, buffer_type):
...
-    if op in [ttnn.mean, ttnn.min]:
-        pytest.skip(reason="Mean and Min are not currently supported in D2M")
+    if op == ttnn.mean:
+        pytest.skip(reason="Mean is not currently supported in D2M")
+
+    # L1 block-sharded min has PCC issues for shapes with many tiles on a
+    # single core.  The neg→max→neg decomposition produces incorrect results
+    # when the intermediate negated tensor is L1 block-sharded with a small
+    # grid.  DRAM min and L1 min on larger grids work correctly.
+    # TODO(sgholami): Investigate L1 block-sharded min PCC failures.
```
vtangTT: can we link issue 🙏

lib/Dialect/TTIR/Transforms/DecomposeMinReduction.cpp
```diff
@@ -0,0 +1,52 @@
+// SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
```
vwellsTT: nit: year

include/ttmlir/Dialect/TTIR/Transforms/Passes.td
```diff
@@ -391,6 +391,15 @@ def TTIRInferKVCacheArgumentTypes: Pass<"ttir-infer-kv-cache-argument-types", ":
   let dependentDialects = ["mlir::tt::ttcore::TTCoreDialect", "mlir::tt::ttir::TTIRDialect"];
 }
 
+def TTIRDecomposeMinReduction: Pass<"ttir-decompose-min-reduction", "::mlir::ModuleOp"> {
```
vwellsTT, line 404: not really specific to this PR, but I think we should have a discussion at some point about how all the various TTIR decomps are organized, currently a lot pre-exist in TTIRToTTIRDecomp, and then we have a ton of these D2M 1-off ones too; it'd be good to settle on standard organization, like a TTIRDecomp dir with a few different files or something maybe

lib/Dialect/D2M/IR/D2MOps.cpp
```diff
@@ -515,6 +515,18 @@ struct ToLayoutFoldOOBUndefPattern : public OpRewritePattern<ToLayoutOp> {
       return failure();
     }
 
+    // Don't fold when the input is undef but the output requires a specific
```
vwellsTT: Huh, I don't think we should fold when these fields differ period, and I thought we didn't when I added support for 'em.  Is that not the case? I don't think this particular special casing is sensible to me

lib/Dialect/TTIR/Transforms/DecomposeMinReduction.cpp
```diff
@@ -0,0 +1,52 @@
...
+
+namespace {
+
+// Decompose min(x) into neg(max(neg(x))) so that backends without native
+// reduce_min (e.g. TTMetal/D2M) can reuse the existing reduce_max tile op.
+//
+// This is a standalone pass rather than part of TTIRToTTIRDecomposition because
+// it is only needed by the TTMetal backend — the TTNN backend has native
+// ttnn.min support.  Keeping it separate avoids adding a backend-specific
+// config mode to the shared decomposition pass and lets each pipeline opt in
```
vwellsTT: FWIW I think decomp already has back-end specific config

lib/Dialect/D2M/IR/D2MOps.cpp
vwellsTT: For example, if input is +inf, output is -inf, you 100% cannot fold away the to_layout either unless there's no oob region at all, etc.

include/ttmlir/Dialect/TTIR/Transforms/Passes.td
sgholamiTT: Agreed -- the current situation is fragmented: some decompositions live in `TTIRToTTIRDecomposition` (under `Conversion/`), others are standalone passes in `TTIR/Transforms/`(e.g. `TTIRDecomposeComplexPermute`, `TTIRDecomposeComplexReshape`, and now this one). A dedicated `TTIRDecomp` directory or at least a settled convention on when to use the shared pass vs. standalone passes would help. Happy to participate in that discussion.

lib/Dialect/D2M/IR/D2MOps.cpp
sgholamiTT: You're right -- the special casing isn't the right approach. I removed the guard entirely and all e2e tests still pass (`golden test_min/test_min_unaligned`, `JIT` nightly, `JIT` smoketest -- 424 tests total). The `to_layout` at the `eltwise-to-reduction` boundary isn't actually matched by ToLayoutFoldOOBUndefPattern in practice because the layouts differ in more than just OOB (e.g. logical shape changes across a reduction). I originally added it out of an abundance of caution, but it turned out to be unnecessary. I also agree with your broader concern about folding when OOB values differ.

lib/Dialect/TTIR/Transforms/DecomposeMinReduction.cpp
sgholamiTT: Updated the comment to drop that incorrect claim.

Issue #7607, opened 2026-03-23
[Runtime] Safe Trace


pilkicTT:
Current implementation of `trace` does not properly handle multiple scenarios and is unsafe. There are several user footguns, which we should ideally handle.

The feature as implemented in `tt-metal` is intended for performance optimizations and special care needs to be taken when combining multiple captures and device allocations between them. In general, any device allocation outside of begin/end capture is unsafe - i.e. those buffers/tensors can be overwritten by execution previously captured trace (by the traced graph's intermediate tensors/buffers). This, however, is not accounted for in our implementation.

PR #7618, opened 2026-03-23
[runtime] fix basic multi-graph trace scenario

### LINKED ISSUE #7608: [runtime] basic multi-graph trace scenario
We need to handle executing multiple traced graphs in the runtime. Currently, it can easily happen that previously captured trace can overwrite any newly allocated inputs or program config tensors (of the newly captured trace).

E.g. this is causing hangs on llama-70b tenstorrent/tt-xla#3078. The prefill's trace is overwritting a reshape kernel's config tensor in the captured decode graph - causing a hang.

pilkicTT:
Current implementation of `trace` is very unsafe when executing multiple graphs - there exist no mechanism to protect the user from one trace overwriting input buffers of another graph (traced or not).

When we capture a trace, all operations captured are replayed exactly on the next execution of the trace - including all memory allocations/reads & writes. Consider the following (slightly simplified) sequence of events:
- all inputs for the graph A are allocated on the device, let's say that we fill the memory space on device up to address 0x100 (simplified)
- we capture the graph A - during capture various intermediate tensors are allocated for individual operations inside of the graph - up to the address 0x200
- during the capture we will deallocate all of the allocated intermediates, freeing up the memory, so again (after the capture is completed) everything after address 0x100 is free
- next graph comes, and we allocate its inputs on the device - these inputs cause the memory up to 0x200 to be filled - and we execute the graph
- now we execute the graph A again, the trace is replayed.

After the above sequence, the graph B's inputs now contain garbage values, since they were overwritten by the trace replay of graph A - because graph A used address range 0x100-0x200 for its intermediates, while graph B used the same range for its inputs.

To prevent this, cached trace invalidation mechanism is implemented. With this change, we will invalidate all existing cached traces after we capture a new graph. Invalidated traces will be recaptured on-demand - when the graph which has an invalidated entry is submitted for execution. This works because everything that needs to be persisted (weight tensors, any program config tensors, etc.) is already allocated, so when recapturing the intermediates of the graph won't be able to use that space - hence no overwrites on future trace replay.

In this change we ONLY cover the basic scenario, which assumes all graphs that will be executed are traced graphs and there are no persistent device tensor allocations outside of the captured graphs (which could be overwritten).

Other scenarios include:
- user allocates and holds on to a tensor on device (while there are existing captured traces)
- user executes a new non-traced graph
- user clears the program cache - while there are existing captured traces.

Covering these scenarios will be done in the future.

Regression e2e test is added - the test verifies that executing two traced graphs does not overwrite any of the weight tensors.

Closes #7608
Parent issue #7607

### DIFF
--- a/runtime/include/tt/runtime/detail/ttnn/types/trace_cache.h
+++ b/runtime/include/tt/runtime/detail/ttnn/types/trace_cache.h
@@ -13,12 +13,32 @@ namespace tt::runtime::ttnn {
 
 struct TraceData {
   ::ttnn::MeshTraceId traceId;
-  // Input tensor buffers to write the input tensors to
+  /// Input tensor buffers to write the input tensors to.
   std::vector<::tt::runtime::Tensor> inputTensors;
-  // Output tensor buffers to write the output tensors to
+  /// Output tensor buffers to write the output tensors to.
   std::vector<::tt::runtime::Tensor> outputTensors;
+  /// Trace cache generation id at capture time - used to detect stale traces.
+  uint64_t generationId = 0;
 };
 
+/// Cache for storing captured traces.
+///
+/// The captured traces are stored by a two-level key:
+/// - the first level key identifies the main program (binary id + main program
+/// index)
+/// - the second level key identifies the capture and execute program IDs for
+/// the trace.
+///
+/// The cache also maintains a generation id to allow detection of stale traces.
+/// Traces are inherently unsafe to reuse after certain operations (e.g. after a
+/// new graph is captured). This is because the old trace may overwrite memory
+/// allocated for the new graph - those allocations were not present at
+/// the time of the capture, the captured graph could have used that memory for
+/// its intermediate buffers.
+///
+/// To handle those cases, the cache provides the generation id - which should
+/// be incremented by runtime whenever it performs an operation that may
+/// invalidate existing traces.
 class TraceCache {
 public:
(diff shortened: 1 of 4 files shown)


runtime/include/tt/runtime/detail/ttnn/types/trace_cache.h
```diff
@@ -13,12 +13,32 @@ namespace tt::runtime::ttnn {
...
 struct TraceData {
   ::ttnn::MeshTraceId traceId;
-  // Input tensor buffers to write the input tensors to
+  /// Input tensor buffers to write the input tensors to.
   std::vector<::tt::runtime::Tensor> inputTensors;
-  // Output tensor buffers to write the output tensors to
+  /// Output tensor buffers to write the output tensors to.
   std::vector<::tt::runtime::Tensor> outputTensors;
+  /// Trace cache generation id at capture time - used to detect stale traces.
+  uint64_t generationId = 0;
```
jameszianxuTT, line 21: I think you probably have a good reason to not make this generationId const, right?

pilkicTT: yeah, generally the cpp core guidelines warn against using const or ref data members - https://clang.llvm.org/extra/clang-tidy/checks/cppcoreguidelines/avoid-const-or-ref-data-members.html

PR #7619, opened 2026-03-23
Allocator-driven blocking for reduction/matmul generics

anuragsingh-tt:
Introduce auto policy for reblocking:
- tunes blocking along a single reduction dimension (K) if it exists
- evaluates larger K-blocking factors (derived from existing shard factors) and selects the one that minimizes estimated streamed input buffer size
- minimum per-core shard size is ≥4 tiles

If no better candidate exists, it is a no-op. It does not yet tune M/N, handle multiple reduction dims, change streaming decisions, or use runtime feedback.

Includes lit test + pytest.

### DIFF
--- a/lib/Dialect/D2M/Transforms/Allocate.cpp
+++ b/lib/Dialect/D2M/Transforms/Allocate.cpp
@@ -789,8 +789,10 @@ class D2MAllocate final : public impl::D2MAllocateBase<D2MAllocate> {
       }
 
       for (std::size_t d = 0; d < rank; ++d) {
+        TT_assert(blockFactors[d] % originalBlockFactors[d] == 0);
         const int64_t rescaling = blockFactors[d] / originalBlockFactors[d];
         gridExtents[d] *= rescaling;
+        TT_assert(shardExtents[d] % rescaling == 0);
         shardExtents[d] /= rescaling;
       }
       genericCtx.reblockedFactors = blockFactors;


anuragsingh-tt:
> Can we transform this into an [MLIR analysis](https://mlir.llvm.org/docs/PassManagement/#analysis-management)? This way the allocator can just do:
> 
> ```c++
> auto graphAnalysis = getAnalysis<BlockFactorAnalysis>();
> auto genericOpAnalysis = graphAnalysis.at(genericOp);
> ```
> 
> Also see this file which may have started the boilerplate:
> 
> * `include/ttmlir/Dialect/D2M/Analysis/GenericOpBufferAnalysis.h`

My latest commit refactors the block analysis out into its own file but getAnalysis<T> compatibility is probably better suited for a broader refactor of the Allocate pass. Right now the analysis needs some options which are passed in as constructor args.

include/ttmlir/Dialect/D2M/Analysis/BlockFactorAnalysis.h
```diff
@@ -0,0 +1,42 @@
+// SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
```
vwellsTT: nit: year

lib/Dialect/D2M/Analysis/BlockFactorAnalysis.cpp
```diff
@@ -0,0 +1,234 @@
+// SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
```
vwellsTT: nit: year

lib/Dialect/D2M/Analysis/BlockFactorAnalysis.cpp
```diff
@@ -0,0 +1,234 @@
...
+
+namespace {
+
+//===----------------------------------------------------------------------===//
+// Block factor analysis logic
+//===----------------------------------------------------------------------===//
+
+/// Returns the index of the single reduction dimension, if any.
+static std::optional<std::size_t>
+getSingleTuningDim(ArrayRef<ttcore::IteratorType> iteratorTypes) {
```
vwellsTT: what does "tuning" mean here?

lib/Dialect/D2M/Analysis/BlockFactorAnalysis.cpp
```diff
@@ -0,0 +1,234 @@
...
+      return std::nullopt;
+    }
+    tuningDim = dim;
+  }
+  return tuningDim;
+}
+
+/// Estimates the total stream buffer bytes for a candidate grid and shard
+/// extents.
+static std::optional<uint64_t> estimateCBBytesForCandidate(
```
vwellsTT: nit: is this an estimate? I think this is an exact calculation based on an estimated/tentative set of params, but the func itself is a deterministic, precise calculation => name is misleading

lib/Dialect/D2M/Transforms/Allocate.cpp
```diff
@@ -264,9 +264,12 @@ class D2MAllocate final : public impl::D2MAllocateBase<D2MAllocate> {
   using Base = impl::D2MAllocateBase<D2MAllocate>;
   using Base::Base;
 
+  enum class BufferSizePolicy { Auto, Min, Max };
```
vwellsTT: probably better to have this defined somewhere common vs have same enum in multiple places (BlockFactorAnalysis.h and here) imo

lib/Dialect/D2M/Analysis/BlockFactorAnalysis.cpp
anuragsingh-tt: Can refine the name here.

test/ttnn-jit/test_matmul_smoketest.py
```diff
@@ -46,6 +46,20 @@ def matmul_composite(input0, input1):
...
+    shapes,
+    ttnn_dtype: ttnn.DataType | None,
+    op,
+) -> str:
+    if (
+        ttnn_dtype == ttnn.DataType.BFLOAT8_B
+        and op in (matmul, matmul_composite)
+        and shapes in ((128, 256, 256), (256, 256, 256))
+    ):
+        return "test-buffer-size-policy=max"
```
vtangTT, line 60: q: by setting the policy to max we're just ensuring that we are not blocking k for these shapes right? is it bc the shapes are too small?

test/ttnn-jit/test_matmul_smoketest.py
```diff
@@ -151,8 +169,6 @@ def test_matmul_f32_fallback(device):
     compiled_op = ttnn_jit.jit(debug=True, compile_only=False, fallback=True)(matmul)
     output = compiled_op(input0, input1)
 
-    assert compiled_op.fallback_used, "Expected fallback for this configuration"
-
```
sgholamiTT, line 155: I see no point in keeping this test as it's not exercising the fallback anymore. Can we remove it ?

tools/ttnn-jit/api.py
```diff
@@ -15,6 +15,7 @@ def jit(
     math_fidelity: ttnn.MathFidelity = ttnn.MathFidelity.HiFi4,
     memory_config: ttnn.MemoryConfig = None,
     fallback: bool = False,
+    extra_pipeline_options: str = "",
```
sgholamiTT, line 18: Can you please update `docs/src/ttnn-jit.md` with this new option ?

sgholamiTT, line 18: I'm on the fence about adding a str option to `jit` (if you think about the end-user, they will most probably have a hard time figuring out when/how to use this option).

My suggestion is to define:
```python
# can be defined on top of tools/ttnn-jit/api.py
from enum import Enum, auto

# We need to define this since: 
#   enum class BufferSizePolicy { Auto, Min, Max }; 
# defined in include/ttmlir/Dialect/D2M/Analysis/BlockFactorAnalysis.h 
# is not pybinded. If it's possible to pybind the C++ struct, that would be 
# better and cleaner, but don't want to push for too many changes here.
class BufferSizePolicy(Enum):
    MIN = auto()
    AUTO = auto()
    MAX = auto()
```
Which can be used like this:

```python
def jit(
    compile_only: bool = False,
    debug: bool = False,
    enable_cache: bool = False,
    math_fidelity: ttnn.MathFidelity = ttnn.MathFidelity.HiFi4,
    memory_config: ttnn.MemoryConfig = None,
    fallback: bool = False,
    buffer_size_policy: BufferSizePolicy = BufferSizePolicy.AUTO,
):
# ...
if buffer_size_policy == BufferSizePolicy.MAX: 
  options += "test-buffer-size-policy=max"
elif buffer_size_policy == BufferSizePolicy.MIN: 
  options += "test-buffer-size-policy=min"
```
This will be a bit more self-explanatory and would be easier/more likely to be configured by the user.

tools/ttnn-jit/api.py
sgholamiTT, line 32: Can we update the `Args` list with the new arg please ?

test/ttnn-jit/test_matmul_smoketest.py
```diff
@@ -90,6 +104,7 @@ def test_matmul_smoketest(device, shapes, input_layouts, dtype, ttnn_dtype, op):
     k = TILE_SIZE if shapes[1] is None else shapes[1] * grid_dim
     n = TILE_SIZE if shapes[2] is None else shapes[2] * grid_dim
 
+    original_shapes = shapes
```
sgholamiTT, line 108: Nit: I guess this is not needed. `get_allocator_policy_override` does not change the values of `shapes`.

test/ttnn-jit/test_matmul_smoketest.py
anuragsingh-tt: Yes, the PCC will drift bc the partial accumulations are too small for this precision.

test/ttnn-jit/test_matmul_smoketest.py
anuragsingh-tt: Sounds good, I am not sure what this test was doing anyways.

test/ttnn-jit/test_matmul_smoketest.py
anuragsingh-tt: It is needed because shapes is reassigned on line 109. The original shape was 256x256x256 which `get_allocator_policy_override` needs, not 2048x2048

tools/ttnn-jit/api.py
anuragsingh-tt: I have found in my use cases for debugging JIT that a general purpose escape hatch for passing arbitrary pipeline options (not just buffer size policy) to the D2M pipeline is useful. I will update both the doc string and the docs to clearly document this as an "advanced use only" option.

But I am open to suggestions on this. Perhaps we can address this more holistically in a follow up PR.

sgholamiTT: Sure that sounds fine if a general purpose arg is useful for you.
We can circle back later if we found it being misused.

lib/Dialect/D2M/Transforms/Allocate.cpp
```diff
@@ -766,35 +782,16 @@ class D2MAllocate final : public impl::D2MAllocateBase<D2MAllocate> {
...
+      if (const auto *bfResult = blockFactorAnalysis.lookup(genericOp)) {
+        blockFactors = bfResult->reblockedFactors;
+      }
 
-          blockFactors[d] *= rescaling[d];
-        }
-        TT_ALLOC_DEBUG("rescaling {}, new block factors {}", asSeq(rescaling),
-                       asSeq(blockFactors));
+      for (std::size_t d = 0; d < rank; ++d) {
+        const int64_t rescaling = blockFactors[d] / originalBlockFactors[d];
```
nsmithtt, line 793: Should we assert divisibility here?

anuragsingh-tt: Right now they are always multiples of the `originalBlockFactors` but I can add it for future proofing.

PR #7624, opened 2026-03-24
[ttnn.jit] Remove TTNN JIT Wheel Dependency on ttmlir Wheel

arminaleTT:
### Problem description
The ttmlir wheel is [no longer being published ](https://github.com/tenstorrent/tt-forge/commit/2feb734645885c9341a561e7d59c755fba32bcd0) to our index. The TTNN JIT wheel currently depends on that wheel. This needs to be fixed so the TTNN JIT wheel is actually usable without users having to download dev artifacts from GitHub Actions.

### What's changed
- Revert https://github.com/tenstorrent/tt-mlir/commit/8371504ec8ea15c29bb93aa5cd5f0f972c54d335
- Bundle TTMLIR libraries in the TTNN JIT wheel
- Use a MetaPathFinder to resolve the hardcoded ttmlir.* import prefix in compiled MLIR .so files without sys.path hacking which was done prior to https://github.com/tenstorrent/tt-mlir/commit/8371504ec8ea15c29bb93aa5cd5f0f972c54d335

#### Import resolution
The MLIR Python binding `.so` files are compiled with `MLIR_PYTHON_PACKAGE_PREFIX=ttmlir.`, baking C-level import paths like `import_("ttmlir._mlir_libs._mlir.ir")` into the binaries. When relocated to `ttnn_jit/ttmlir/`, these hardcoded paths would load a second copy of the `.so` from `build/python_packages/` on `PYTHONPATH`, causing a nanobind duplicate type registration crash.
- `tools/ttnn-jit/__init__.py`: Added a `_TtmlirRedirector` meta path finder that intercepts all bare `ttmlir.*` imports and redirects them to `ttnn_jit.ttmlir.*`. This does not mutate `sys.path`.
- `tools/ttnn-jit/_jit_site_initialize_0.py`** (new): JIT-wheel-specific override for `_site_initialize_0.py`, installed via CMake `RENAME` to replace the copy from the build directory.
- `tools/ttnn-jit/_src/*.py`: Changed all `from ttmlir.*` imports to `from ttnn_jit.ttmlir.*`. Changed `importlib.import_module(f"ttmlir.dialects.{dialect}")` to use the `ttnn_jit.ttmlir` prefix.

### Checklist
- [x] Nightly tests: https://github.com/tenstorrent/tt-mlir/actions/runs/24090840211

### DIFF
--- a/python/CMakeLists.txt
+++ b/python/CMakeLists.txt
@@ -241,7 +241,7 @@ if(APPLE)
 endif()
 
 # PyWheel Component to create install directives for the packaged TTMLIR bindings wheel:
-install(DIRECTORY ${CMAKE_BINARY_DIR}/python_packages/ DESTINATION . COMPONENT TTMLIRPythonWheel EXCLUDE_FROM_ALL)
+install(DIRECTORY ${CMAKE_BINARY_DIR}/python_packages/ttmlir DESTINATION . COMPONENT TTMLIRPythonWheel EXCLUDE_FROM_ALL)
 
 install(DIRECTORY ${CMAKE_BINARY_DIR}/python/dialects/ DESTINATION python/dialects COMPONENT Test EXCLUDE_FROM_ALL)
 install(DIRECTORY ${CMAKE_BINARY_DIR}/python_packages/ DESTINATION python_packages COMPONENT Test EXCLUDE_FROM_ALL)


tools/ttnn-jit/_src/__init__.py
```diff
@@ -4,25 +4,82 @@
...
-    get_l1_base_allocator_addr,
-    get_lowest_occupied_compute_l1_address,
-    get_l1_size_per_core,
-)
-from ttnn_jit._ttnn_jit import JitCache
+import sys
+
+# Try wheel installation path first, then fallback to local development path
+try:
+    import ttnn_jit.runtime.ttmlir as _ttmlir_module
```
vtangTT: 100% do not want to do this, it should work with how the import paths were setup before

vtangTT:
@tapspatel do we want to still build locally ttmlir wheel and only bundle ttmlir libs in a release env to avoid the problems you were seeing before? Or are you ok for us to just revert and bundle ttmlir libs into the jit wheel like we were doing before.

tools/ttnn-jit/_src/__init__.py
arminaleTT: We had a conversation offline. The old import statements definitely do not work since the directory structure has changed. Naively changing `ttmlir.*` to `ttnn_jit.ttmlir.*` also does not work since it's possible to load both the bundled and system level ttmlir libraries. This manifests as a crash caused by enum collisions in nanobind.

In the interest of not replicating the sys.path hacking, I've created a MetaPathFinder that intercepts and redirects the import calls

tools/ttnn-jit/CMakeLists.txt
```diff
@@ -32,17 +32,26 @@ if (USE_TTNN_JIT_WHEEL)
     COMPONENT TTNNJITWheel
     EXCLUDE_FROM_ALL
   )
+  install(DIRECTORY ${CMAKE_BINARY_DIR}/python_packages/ttmlir/
+    DESTINATION ttnn_jit/ttmlir
+    COMPONENT TTNNJITWheel
+    EXCLUDE_FROM_ALL
+  )
+  install(FILES ${CMAKE_CURRENT_SOURCE_DIR}/_jit_site_initialize_0.py
```
vtangTT: confused why the above install of `python_packages/ttmlir/` doesn't copy over the `_site_initialize_0.py` used in `_mlir_libs`? is it bc it's symlinked?

asking bc the new one just looks like a duplicate

tools/ttnn-jit/CMakeLists.txt
```diff
@@ -98,4 +107,10 @@ else ()
   add_custom_target(ttnn-jit ALL
     COMMENT "ttnn-jit package"
     DEPENDS TTNNJITModules _ttnn_jit _ttmlir_runtime TTMLIRPythonModules)
+  add_custom_command(
```
vtangTT, line 107: curious why we're symlinking here? 

wouldn't the ttnn-jit cmake install step install the `python_packages/ttmlir` into the wheel so we shouldn't need this? or am i missing smthn here?

tools/ttnn-jit/__init__.py
```diff
@@ -2,9 +2,46 @@
...
 import subprocess
+import sys
+
+
+class _TtmlirRedirector(importlib.abc.MetaPathFinder):
+    """Redirect bare ``ttmlir.*`` imports to ``ttnn_jit.ttmlir.*``.
+
+    The MLIR Python bindings .so files are compiled with
+    ``MLIR_PYTHON_PACKAGE_PREFIX=ttmlir.`` and hardcode C-level imports such as
+    ``import_("ttmlir._mlir_libs._mlir.ir")``.  When the ttmlir tree is bundled
```
vtangTT, line 19: good to know 👍

tools/ttnn-jit/CMakeLists.txt
arminaleTT: Yes, good catch. Removed

tools/ttnn-jit/CMakeLists.txt
arminaleTT: This is for `USE_TTNN_JIT_WHEEL==OFF` case. It keeps the imports functional. I've added a comment

python/CMakeLists.txt
```diff
@@ -241,37 +241,7 @@ if(APPLE)
 endif()
 
 # PyWheel Component to create install directives for the packaged TTMLIR bindings wheel:
-install(DIRECTORY ${CMAKE_BINARY_DIR}/python_packages/ttmlir DESTINATION . COMPONENT TTMLIRPythonWheel EXCLUDE_FROM_ALL)
+install(DIRECTORY ${CMAKE_BINARY_DIR}/python_packages/ DESTINATION . COMPONENT TTMLIRPythonWheel EXCLUDE_FROM_ALL)
```
vtangTT: this should still only be installing from `python_packages/ttmlir` or else we are copying everything (ttrt, builder, etc..) into the ttmlir wheel

but i guess if the ttmlir wheel is no longer being published this wouldn't matter... 🤷

arminaleTT:
> Thanks for doing this. LGTM. 💯
> 
> 1. Please run nightly on this branch if we're doubtful sth could break.
> 2. Does the `ttnn-jit` wheel work on metal dev env now ?
> 3. Please rebase after [this](https://github.com/tenstorrent/tt-mlir/pull/7698) was merged. Most probably some conflicts will appear.

1 & 3. Yes, will do!
2. Yes, @vtangTT and I tested it and it can be installed and imported on a metal dev env

vvukomanTT (approved):

PR #7636, opened 2026-03-25
Refreshing the device mapping tensor in reshape op

sshonTT:
### Ticket
[Link to Github Issue](https://github.com/tenstorrent/tt-metal/issues/40612)

### Problem description
The tiled reshape kernel in tt-metal uses a pre-computed mapping_tensor (stored in DRAM) that describes how to copy data segments from input tiles to output tiles. When the program cache is hit, this mapping tensor is reused from the previous invocation. However, when the output shape has non-tile-aligned dimensions (height or width not a multiple of the tile size), the mapping tensor's DRAM region gets corrupted between invocations, causing a NOC DMA hang.

Root cause (observed behavior):
Partial tiles (created by non-tile-aligned height/width) produce a different mapping structure with non-standard segment offsets. When the cached program is reused, the corrupted mapping data causes the reader kernel to issue NOC reads to invalid DRAM addresses, leading to a device hang. The exact corruption mechanism is not fully understood, but it is reproducible and deterministic for a given shape.

### What's changed
At the tt-mlir runtime layer, we check if the output shape's last two dimensions (height, width) are tile-aligned. If either is not, we pass TileReshapeMapMode::RECREATE to ttnn::reshape, which forces tt-metal to recompute and reallocate the mapping tensor on every program cache hit instead of reusing the stale one.
Performance impact: Negligible. The mapping tensor is small (tens of KB) and the host-side recomputation + PCIe transfer cost is sub-millisecond.

### Checklist

### DIFF
--- a/runtime/lib/ttnn/operations/data_movement/reshape.cpp
+++ b/runtime/lib/ttnn/operations/data_movement/reshape.cpp
@@ -8,6 +8,38 @@
 
 #include "tt/runtime/detail/ttnn/utils.h"
 
+namespace {
+
+// Tiled reshape can corrupt the cached mapping_tensor's DRAM region when
+// the output has non-tile-aligned height or width.  Force the mapping tensor to
+// be recreated on every invocation for such shapes.
+bool isNonTileAligned(const std::vector<int32_t> &outputShape,
+                      const std::array<uint32_t, 2> &tileShape) {
+  if (outputShape.empty()) {
+    return true;
+  }
+  int32_t widthDim = outputShape.back();
+  int32_t heightDim = outputShape.size() > 1
+                          ? outputShape[outputShape.size() - 2]
+                          : static_cast<int32_t>(tileShape[0]);
+  if (widthDim <= 0 || heightDim <= 0) {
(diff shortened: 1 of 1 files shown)


sshonTT:
@pilkicTT

The primary blocker I’m facing is the reshape op, not concat. I discovered the concat issue while debugging other models, but for our current model, reshape is the main issue. Since the concat hang was reportedly fixed with a NOC workaround, I initially thought the two might be related. However, the reshape issue is much more complex to reproduce, and I believe it's still being investigated.

As far as I can tell, this is the most minimal workaround possible. The alternative would be enabling [this PR](https://github.com/tenstorrent/tt-mlir/pull/7518), which carries a higher cost.

Regarding performance, my local runs looked fine, but I agree we need formal verification before merging. I’ve just triggered a benchmark test in tt-xla : https://github.com/tenstorrent/tt-xla/actions/runs/23652961224
If you have any suggestion to run other benchmark, please let me know.

We really need this workaround to support various batch sizes for GPT OSS models, as this hang is currently a major blocker for us.

PR #7641, opened 2026-03-25
Add TTNN builder @split support and deallocate op

sdjordjevicTT:
### Ticket
Solves https://github.com/tenstorrent/tt-mlir/issues/7264

## Summary
- Add `@split` methods for all TTNN ops in `ttnn_builder.py` (~43 ops covering unary, binary, ternary, variadic, and special patterns)
- Add full builder support (`@tag`, `@parse`, `@split`) for `ttnn.deallocate` op
- Extend `test_ttnn_parsing_splitting_ops` to call `split_mlir_file` (previously only parsed)
- Add `ttnn_deallocate.mlir` test snippet
- Skip `ttnn_all_gather.mlir` in split tests (`GetDeviceOp` has no tensor inputs)

### Checklist

### DIFF
--- a/test/python/golden/mlir_snippets/ttnn/ttnn_deallocate.mlir
+++ b/test/python/golden/mlir_snippets/ttnn/ttnn_deallocate.mlir
@@ -1,0 +1,8 @@
+#dram = #ttnn.buffer_type<dram>
+#ttnn_layout = #ttnn.ttnn_layout<(d0, d1) -> (d0, d1), <1x1>, memref<1x1x!ttcore.tile<32x32, f32>, #dram>, <interleaved>>
+module {
+  func.func @model(%arg0: tensor<32x32xf32, #ttnn_layout>) -> () {
+    "ttnn.deallocate"(%arg0) <{force = false}> : (tensor<32x32xf32, #ttnn_layout>) -> ()
+    return
+  }
+}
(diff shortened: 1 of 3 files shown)


jgrimTT:
A while back, the tools team made the decision to remove support for TTNNBuilder split, because we didn’t have a mechanism to handle splitting ttnn graphs with ops that take the device as an input (like `ttnn.rand`, `ttnn.to_device`) because those require an instance of `ttnn.get_device` to function and can't be split into single-op graphs. We didn't think it was worth maintaining a ttnn split API that had such limited functionality. It was on my to-do list to create the infra to support device-involved ttnn splits, but now that we're going after runtime golden checks again, I'm unsure if any of this is necessary. Let me know if you have thoughts, thanks!

sdjordjevicTT:
> A while back, the tools team made the decision to remove support for TTNNBuilder split, because we didn’t have a mechanism to handle splitting ttnn graphs with ops that take the device as an input (like `ttnn.rand`, `ttnn.to_device`) because those require an instance of `ttnn.get_device` to function and can't be split into single-op graphs. We didn't think it was worth maintaining a ttnn split API that had such limited functionality. It was on my to-do list to create the infra to support device-involved ttnn splits, but now that we're going after runtime golden checks again, I'm unsure if any of this is necessary. Let me know if you have thoughts, thanks!

Let's merge this in, and later add support for splitting ttnn modules as well. I see this need as:
- at TTIR and above levels, we don't have fusions that are applied, hence those fusions can be problematic(they can introduce PCC issues), hence I see a value in having this ability to split TTNN graphs as well...

PR #7649, opened 2026-03-25
Add optional per-op device synchronization for debugging

sshonTT:
### Ticket
N/A

### Problem description
Since ops are executed asynchronously, it’s hard to identify which one is actually hanging based on the current logs alone.

### What's changed
Add syncAfterOpIfNeeded() to ProgramExecutor that synchronizes all devices after each operation when TT_RUNTIME_SYNC_AFTER_OP env var is set. Guarded behind TT_RUNTIME_DEBUG to ensure zero overhead in release builds. Useful for isolating which op causes hangs or incorrect results on device.

### Checklist

### DIFF
--- a/runtime/lib/ttnn/program_executor.cpp
+++ b/runtime/lib/ttnn/program_executor.cpp
@@ -4,6 +4,12 @@
 
 #include "tt/runtime/detail/ttnn/program_executor.h"
 
+#if defined(TT_RUNTIME_DEBUG) && TT_RUNTIME_DEBUG == 1
+#include <cstdlib>
+
+#include <tt-metalium/distributed.hpp>
+#endif
+
 #include "operations/cache/load_cached.h"
 #include "operations/ccl/aggregate_tensor.h"
 #include "operations/ccl/all_gather.h"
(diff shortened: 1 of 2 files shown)


runtime/lib/ttnn/program_executor.cpp
```diff
@@ -579,4 +588,16 @@ void ProgramExecutor::dumpPerfCountersIfNeeded() {
 #endif
 }
 
+#if defined(TT_RUNTIME_DEBUG) && TT_RUNTIME_DEBUG == 1
+void ProgramExecutor::syncAfterOpIfNeeded() {
+  static const bool enabled =
+      std::getenv("TT_RUNTIME_SYNC_AFTER_OP") != nullptr;
+  if (enabled) {
+    LOG_DEBUG(LogType::LogRuntimeTTNN, "Synchronizing after operation");
```
jameszianxuTT: nit: idk if you want to print this after every op. It is going to be very noisy, and since this affects all ops, you could maybe just print it once at the start of program execution or not at all.

sshonTT: Yeah make sense. Just removed this line. Thanks!

Issue #7240, opened 2026-02-27
[StableHLO] Failed to legalize stablehlo.reduce_window for 3D max pooling (max_pool3d)


ashokkumarkannan1:
### Describe the bug

The `stablehlo.reduce_window` operation generated by `torch.ops.aten.max_pool3d_with_indices `fails to legalize during the StableHLO to TTIR conversion pass. This blocks Dense UNet 3D model  that use 3D max pooling. (complete log -[denseunet_3d.log](https://github.com/user-attachments/files/25608579/denseunet_3d.log))

Error message:
`loc("reduce-window.14"): error: failed to legalize operation 'stablehlo.reduce_window'`

Root op in PyTorch:
```
torch.ops.aten.max_pool3d_with_indices.default(
    x,               # shape: [1, 96, 16, 128, 128]
    [3, 3, 3],       # kernel_size
    [2, 2, 2],       # stride
    [1, 1, 1],       # padding
    [1, 1, 1],       # dilation
    False,           # ceil_mode
)
```

### Steps to reproduce the issue
- Run the below isolated op sanity test (in tt-xla) (sanity log for reference - [sanity_maxpool.log](https://github.com/user-attachments/files/25608589/sanity_maxpool.log))

```
git checkout akannan/denseunet_3d_maxpool
pytest tests/torch/ops/test_maxpool3d.py::test_maxpool3d[dense_unet_3d_config] -svv
```

ashokkumarkannan1:
@mrakitaTT could you please take a look?

PR #7671, opened 2026-03-26
Support 5D max reduce window case

sdjukicTT:
### Ticket
https://github.com/tenstorrent/tt-mlir/issues/7240

### Problem description
An instance of stablehlo.reduce_window op with MAX reducer appeared, that does reduction on 3 dimensions. This wasn't supported.

### What's changed
New case: 5D input (3D max pooling)
ReduceWindowOp lowering now handles rank-5 inputs (3D pooling, e.g. nn.MaxPool3d). Since TTIR only has MaxPool2d, the 3D pooling window [kD, kH, kW] is decomposed into two sequential 2D max pool passes:
  1. Pass 1 (H,W): Reshape [N,C,D,H,W] → [N*C*D, H, W, 1], MaxPool2d with [kH, kW]
  2. Pass 2 (D): Reshape [N*C*D, Hout, Wout, 1] → [N*C, D, Hout*Wout, 1], MaxPool2d with [kD, 1]

This is valid because max is associative: max(d,h,w) = max_d(max(h,w)). Only max pooling is supported — average pooling does not decompose this way.                              
   
 Removed checks (redundant with [StableHLO constraints](https://openxla.org/stablehlo/spec#reduce_window)):                                                                                                                             
  - "All inputs must have the same shape" — guaranteed by C2: same(shape(inputs...)).                                                                                               
  - "Mismatch between inputs and init values" — guaranteed by C1: size(inputs) = size(init_values) = size(results) = N.
  - "The number of inputs must equal the number of outputs" — also guaranteed by C1.                                                                                                
  - windowDimensions.size() != inputRank — guaranteed by C4: size(window_dimensions) = rank(inputs[0]).                                                                             
  - windowStrides.size() != inputRank — guaranteed by C6: size(window_strides) = rank(inputs[0]).                                                                                   
  - windowDilations.size() != inputRank — guaranteed by C10: size(window_dilations) = rank(inputs[0]).                                                                              
  - baseDilations.size() != inputRank — guaranteed by C8: size(base_dilations) = rank(inputs[0]).                                                                                   
                                                                                                                                                                                    
  New check: base dilations must all be 1:                                                                                                                                           

StableHLO allows base_dilations > 1 (C9 only requires > 0), which inserts zeros between input elements before windowing. TTIR's MaxPool2d has no equivalent — it operates on dense inputs only. This check was previously implicit (the attribute size mismatch checks would sometimes mask it), but now it's explicit and applies to all ranks, correctly rejecting inputs that TTIR cannot represent. 

### Checklist

### DIFF
--- a/lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
+++ b/lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
@@ -2454,6 +2454,10 @@ public:
       return rewriter.notifyMatchFailure(
           srcOp, "Failed to extract constant init values.");
     }
+    if (initValues->size() != srcOp.getInputs().size()) {
+      return rewriter.notifyMatchFailure(
+          srcOp, "Mismatch between init values and inputs count.");
+    }
     // Validate block body.
     Block &block = srcOp.getBody().getBlocks().front();
     auto &operations = block.getOperations();
@@ -2503,6 +2507,13 @@ public:
 
     BoolAttr countIncludesPad = rewriter.getBoolAttr(true);
 
+    // TTIR pooling ops do not support base dilations.
+    if (!llvm::all_of(baseDilations.asArrayRef(),
+                      [](int64_t d) { return d == 1; })) {
+      return rewriter.notifyMatchFailure(
+          srcOp, "Base dilations other than 1 are not supported.");
+    }
+
     // Handle the special case of lowering to CumSumOp.
     if (srcOp.getInputs().size() == 1) {
       std::optional<int64_t> dimension =
@@ -2517,13 +2528,6 @@ public:
       }
     }
 
-    // TTIR pooling ops do not support base dilations.
-    if (!llvm::all_of(baseDilations.asArrayRef(),
-                      [](int64_t d) { return d == 1; })) {
-      return rewriter.notifyMatchFailure(
-          srcOp, "Base dilations other than 1 are not supported.");
-    }
-
     // Handle 5D input (3D pooling) by decomposing into two 2D pooling passes.
     // This works because max is associative: max(i,j,k) = max_i(max(j,k)).
     if (inputRank == 5) {
@@ -2532,8 +2536,8 @@ public:
                                  windowDilations, padding, ceilMode);
     }
 
-    // Not a special case of CumSumOp - lowering to TTIR pooling ops is
-    // supported only for 2D and 4D input tensors.
+    // Lowering to a single TTIR pooling op is supported only for 2D and 4D
+    // input tensors.
     if (!(inputRank == 2 || inputRank == 4)) {
       return rewriter.notifyMatchFailure(srcOp, "Invalid input tensor rank.");
     }
@@ -2796,7 +2800,7 @@ private:
 
     // Non-spatial dimensions are folded into batch, so they must have trivial
     // window attributes. Stride > 1 would subsample, padding would change the
-    // output size — neither is handled by the batch-folding reshapes.
+    // output size - neither is handled by the batch-folding reshapes.
     for (size_t dim : nonSpatialDims) {
       if (windowStrides[dim] != 1 || padding[dim * 2] != 0 ||
           padding[dim * 2 + 1] != 0) {
@@ -2860,8 +2864,9 @@ private:
     auto paddingD = rewriter.getDenseI32ArrayAttr(
         {static_cast<int32_t>(pDLo), 0, static_cast<int32_t>(pDHi), 0});
 
-    // TODO: Fuse preceding stablehlo.PadOp into pooling padding, similar to
-    // the 4D path (getFusablePadOp / combinePaddingFromPadOp).
+    // A preceding stablehlo.PadOp could be fused into the pooling padding,
+    // similar to the 4D path (getFusablePadOp / combinePaddingFromPadOp), but
+    // the 5D case is uncommon enough that this optimization is deferred.
     SmallVector<Value> resultVals;
     for (size_t i = 0; i < srcOp.getInputs().size(); ++i) {
       Value input = adaptor.getInputs()[i];


lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
```diff
@@ -2768,6 +2752,221 @@ class StableHLOToTTIRReduceWindowOpConversionPattern
...
+
+    // Pool2d attributes for Pass 2 (D pooling).
+    auto kernelD = rewriter.getDenseI32ArrayAttr({static_cast<int32_t>(kD), 1});
+    auto strideD = rewriter.getDenseI32ArrayAttr({static_cast<int32_t>(sD), 1});
+    auto dilationD =
+        rewriter.getDenseI32ArrayAttr({static_cast<int32_t>(dD), 1});
+    auto paddingD = rewriter.getDenseI32ArrayAttr(
+        {static_cast<int32_t>(pDLo), 0, static_cast<int32_t>(pDHi), 0});
+
+    // TODO: Fuse preceding stablehlo.PadOp into pooling padding, similar to
```
vmilosevic: :warning: **google\-readability\-todo** :warning:
missing username/bug in TODO
```suggestion
    // TODO(nsmith): Fuse preceding stablehlo.PadOp into pooling padding, similar to
```

lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
```diff
@@ -2548,6 +2517,21 @@ class StableHLOToTTIRReduceWindowOpConversionPattern
       }
     }
 
+    // TTIR pooling ops do not support base dilations.
```
anuragsingh-tt: Is there a TODO to support this some other way?

lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
```diff
@@ -2768,6 +2752,221 @@ class StableHLOToTTIRReduceWindowOpConversionPattern
...
+        indicesOfValuesGreaterThanOne(windowDimensions.asArrayRef());
+    if (spatialDims.size() != 3) {
+      return rewriter.notifyMatchFailure(
+          srcOp, "5D reduce_window requires exactly 3 spatial dimensions.");
+    }
+
+    // Only max pooling is supported for 5D decomposition.
+    for (size_t i = 0; i < reductionOps.size(); ++i) {
+      if (!isa<mlir::stablehlo::MaxOp>(reductionOps[i]) ||
+          initValues[i] != NEG_INF) {
```
anuragsingh-tt, line 2787: What does initValues != NEG_INF imply?

acicovicTT:
Thx for removing validations that are guaranteed by SHLO! :) However, I think the one Copilot comment about `extractInitValues` might have a point, even if there is a verifier in the framework, it might confuse someone who is reading the code as to why we're indexing into something for which we have not validated the size, so I would leave that one check.

lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
sdjukicTT: This was just a missing check, base dilations aren't propagated and are assumed to be 1. It is out of scope of this PR whether this can be supported and if there is a need.

lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
```diff
@@ -2548,6 +2517,21 @@ class StableHLOToTTIRReduceWindowOpConversionPattern
...
+
+    // Handle 5D input (3D pooling) by decomposing into two 2D pooling passes.
+    // This works because max is associative: max(i,j,k) = max_i(max(j,k)).
+    if (inputRank == 5) {
+      return lowerReduceWindow5D(srcOp, adaptor, rewriter, *initValues,
+                                 reductionOps, windowDimensions, windowStrides,
+                                 windowDilations, padding, ceilMode);
+    }
+
     // Not a special case of CumSumOp - lowering to TTIR pooling ops is
```
acicovicTT: I would also update this comment, since it references the previous version of the code where cumsum lowering was immediately before the default case of maxpool2d/avgpool2d.

lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
```diff
@@ -2548,6 +2517,21 @@ class StableHLOToTTIRReduceWindowOpConversionPattern
       }
     }
 
+    // TTIR pooling ops do not support base dilations.
+    if (!llvm::all_of(baseDilations.asArrayRef(),
```
acicovicTT: I suggest moving this check to before cumsum lowering, lumping it together with other validations

lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
sdjukicTT: Together with the MaxOp check, this confirms we're looking at a true max pooling operation. Padded values are assumed to be NEG_INF.

lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
acicovicTT: We can do this in the follow up PR where we will break up this pattern matching class into 3 classes

sdjukicTT: Agreed, this isn't high priority since 5D case is rare in models.

lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
sdjukicTT: Updated.

lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
sdjukicTT: Moved.

PR #7678, opened 2026-03-27
Set nanobind==2.10.2 throghout the whole tt-mlir

ajakovljevicTT:
Since ninja runs steps in parallel, different version of nanobind through tt-mlir can mess up the building upstream(in tt-xla), causing an ABI mismatch between the `nanobind-static-python-runner.a` (compiled from 2.12.0, which removed nb_module_traverse/nb_module_clear) and python_runner.cpp (compiled against 2.10.2 headers, which declare them). Pinning nanobind to 2.10.2 everywhere to fix this.

### DIFF
--- a/tools/ttrt/requirements.txt
+++ b/tools/ttrt/requirements.txt
@@ -1,5 +1,4 @@
 --extra-index-url https://download.pytorch.org/whl/cpu
 torch@https://download.pytorch.org/whl/cpu/torch-2.9.1%2Bcpu-cp312-cp312-manylinux_2_28_x86_64.whl; sys_platform == "linux"
 torch==2.9.1; sys_platform == "darwin"
-# NOTE: Must match tt-metal's pinned nanobind version to avoid ODR violations.
 nanobind==2.10.2


wenbinlyuTT:
I'm not sure how safe this is when we load xla-mlir-metal into the same address space, if all of them use their own nanobind and the versions don't match.

Last time I did metal uplift (which rebuild the docker images), nanobind released version 2.11.0 & the MLIR build brought that in, causing weird run-time errors. I thought we must align the nanobind version across the whole stack (basically use the same version tt-metal uses since they're the ultimate upstream for forge).

Also, my intention was for the downgrade to happen in https://github.com/tenstorrent/tt-mlir/blob/e1ed34338a4e6fb9404317cc49d38bf1332ba851/env/CMakeLists.txt#L55 , when we build the docker image. I assume the header be swapped during `ttrt` installation is happening in the xla fe, not in tt-mlir?

ajakovljevicTT:
> I assume the header be swapped during ttrt installation is happening in the xla fe, not in tt-mlir?

@wenbinlyuTT Correct, it happens when tt-xla is installing ttrt. What is the nanobind version that we can agree on (also needs to take into account the nanobind version that llvm uses) in tt-mlir, as this discrepancy is blocking tt-mlir -> tt-xla uplift?

wenbinlyuTT:
Metal is pinning 2.10.2 in https://github.com/tenstorrent/tt-metal/blob/f27eb6d24bbdad394972e5a5977abc66a619be45/third_party/CMakeLists.txt#L205
In LLVM trunk they still do `>=2.9, <3.0` https://github.com/llvm/llvm-project/blob/15940b130a21c4a8c13f97f28676f5791e538102/mlir/python/requirements.txt#L2
So I guess pinning 2.10.2 across the stack should be able to get us a long way. If MLIR updates its requirement we have more than enough time to update it in tt-metal first.

ajakovljevicTT:
> Metal is pinning 2.10.2 in https://github.com/tenstorrent/tt-metal/blob/f27eb6d24bbdad394972e5a5977abc66a619be45/third_party/CMakeLists.txt#L205 In LLVM trunk they still do `>=2.9, <3.0` https://github.com/llvm/llvm-project/blob/15940b130a21c4a8c13f97f28676f5791e538102/mlir/python/requirements.txt#L2 So I guess pinning 2.10.2 across the stack should be able to get us a long way. If MLIR updates its requirement we have more than enough time to update it in tt-metal first.

@wenbinlyuTT I've pinned the nanobind to 2.10.2 in this PR in llvm and ttrt, which fixed the issue upstream on tt-xla. Can you take a look and see if this works for you?

svuckovicTT:
@azecevicTT @sgligorijevicTT fyi if this ends up breaking the emitpy-ttxla path

ajakovljevicTT:
> Approving, but wondering if there is some way that we can detect when metal updates the pin to a new version?

Issue for that is here: https://github.com/tenstorrent/tt-mlir/issues/7802

PR #7686, opened 2026-03-27
Greedy optimizer: L1 spill management pass

### LINKED ISSUE #7457: Greedy optimizer: L1SpillManagement pass
## Summary

Introduce the `TTNNGreedyL1SpillManagement` pass — a validation-based L1 budget enforcement pass that resolves over-committed L1 memory by spilling tensors to DRAM.

## Motivation

LayoutPropagation (Pass 1) validates each op-to-op edge in isolation. It confirms that a single producer-consumer pair can both fit in L1, but does not account for other tensors that are simultaneously live. This pass takes the L1 layout assignments from Pass 1 and adjusts them based on global memory pressure.

## How it works

The pass walks operations in schedule order, tracking all live tensors and their L1 sizes at each point. When total L1 usage exceeds the per-core memory budget, it must choose which tensor to spill to DRAM. The eviction strategy is **furthest next use** — the tensor whose next consumer is latest in the schedule gets spilled first, since it occupies L1 the longest before being needed again. This is the theoretically optimal strategy for minimizing total spills given a fixed budget.

After spilling a tensor, all downstream ops that consume it are revalidated with the updated (DRAM) input layout, since the op config may need to change when an input moves from L1 sharded to DRAM interleaved.

## Components

- `L1SpillManagement<MemoryTracker>` templated core with schedule walking and liveness tracking
- `SumL1MemoryTracker`: address-aware L1 allocation tracker with speculative fragmentation detection
- Cascading revalidation of consumers after spill
- `L1SpillObserver`: NullObject observer interface for tracing spill decisions

## Design Reference

[TTNN Optimizer spec, Section 5.6.2](docs/src/specs/ttnn-optimizer.md)

Part of the greedy optimizer umbrella.

rpavlovicTT:
## Summary

- Introduce `TTNNGreedyL1SpillManagement` pass — validation-based L1 budget enforcement using Belady's optimal page replacement (furthest next use) eviction strategy
- Add `enableGreedyOptimizer` pipeline option to select greedy optimizer (LayoutPropagation + L1SpillManagement) vs default TTNNOptimizer
- Wire L1SpillManagement after LayoutPropagation in the greedy pipeline

## Components

- `L1SpillManagement<MemoryTracker>` — templated core with schedule walking and liveness tracking
- `SumL1MemoryTracker` — address-aware L1 allocation tracker with speculative fragmentation detection
- Cascading revalidation of consumers after spill
- `L1SpillObserver` — NullObject observer interface for tracing spill decisions

## How it works

LayoutPropagation (Pass 1) validates each op-to-op edge in isolation. It confirms that a single producer-consumer pair can both fit in L1, but does not account for other tensors that are simultaneously live. This pass takes the L1 layout assignments from Pass 1 and adjusts them based on global memory pressure.

The pass walks operations in schedule order, tracking all live tensors and their L1 sizes at each point. When total L1 usage exceeds the per-core memory budget, it chooses which tensor to spill to DRAM using **furthest next use** — the tensor whose next consumer is latest in the schedule gets spilled first, since it occupies L1 the longest before being needed again.

## Test plan

- [x] `belady_spill.mlir` — verifies spilling under extreme L1 pressure (1% cap)
- [x] `no_spill_headroom.mlir` — verifies no spills with comfortable headroom
- [x] `beam_spill_with_reshards.mlir` — verifies spill interaction with reshards

Stacks on #7454
Closes #7457

### DIFF
--- a/include/ttmlir/Dialect/TTNN/Analysis/L1SpillManagement.h
+++ b/include/ttmlir/Dialect/TTNN/Analysis/L1SpillManagement.h
@@ -253,6 +253,13 @@ private:
                                const ScheduleData &data, uint64_t opL1Usage,
                                uint64_t cbPeakUsage, uint64_t outputL1Size);
 
+  /// Run contiguous-fit and CB-fragmentation checks on a validated op's
+  /// output. Returns the (possibly updated) L1 size to add to the live set,
+  /// or 0 if the output was demoted to DRAM.
+  uint64_t ensureFitsL1(Operation *op, int64_t pos, const ScheduleData &data,
+                        uint64_t opL1Usage, uint64_t cbPeakUsage,
+                        uint64_t l1Size);
+
   /// OOM recovery: demote to L1-interleaved, evict farthest-use, or spill self.
   void handleOOM(Operation *op, int64_t pos,
                  llvm::ArrayRef<OpResult> tensorResults,
--- a/lib/Dialect/TTNN/Analysis/L1SpillManagement.cpp
+++ b/lib/Dialect/TTNN/Analysis/L1SpillManagement.cpp
@@ -526,6 +526,26 @@ void L1SpillManagement<MemoryTracker>::processDeadTensors(
 }
 
 //===----------------------------------------------------------------------===//
+// ensureFitsL1
+//===----------------------------------------------------------------------===//
+
+template <typename MemoryTracker>
+uint64_t L1SpillManagement<MemoryTracker>::ensureFitsL1(
+    Operation *op, int64_t pos, const ScheduleData &data, uint64_t opL1Usage,
+    uint64_t cbPeakUsage, uint64_t l1Size) {
+  auto speculativeAddr = memoryTracker.wouldAllocateAt(l1Size);
+  if (!speculativeAddr) {
+    l1Size = handleNoFit(op, pos, data, opL1Usage, l1Size);
+    speculativeAddr = memoryTracker.wouldAllocateAt(l1Size);
+  }
+  if (l1Size > 0 && speculativeAddr &&
+      wouldCBsOverlapTensors(op, pos, cbPeakUsage, *speculativeAddr)) {
+    l1Size = handleFragmentation(op, pos, data, opL1Usage, cbPeakUsage, l1Size);
+  }
+  return l1Size;
+}
+
+//===----------------------------------------------------------------------===//
 // handleOOM
 //===----------------------------------------------------------------------===//
 
@@ -557,28 +577,8 @@ void L1SpillManagement<MemoryTracker>::handleOOM(
     uint64_t l1Size =
         result.outputL1Usage > 0 ? result.outputL1Usage : opL1Usage;
     if (l1Size > 0) {
-      // Same contiguous-fit + CB frag checks as the main success path.
-      // Without these, ops that pass validation after eviction can still
-      // fail at runtime due to address-space fragmentation.
-      auto speculativeAddr = memoryTracker.wouldAllocateAt(l1Size);
-      TTMLIR_DEBUG(ttmlir::LogComponent::GreedyOptimizer,
-                   "    POST_OOM_FIT: l1Size={0}, speculativeAddr={1}, "
-                   "lowestOccupied={2}, occupied={3}/{4}, "
-                   "cbPeakUsage={5}",
-                   l1Size, speculativeAddr ? (int64_t)*speculativeAddr : -1,
-                   memoryTracker.getLowestOccupiedAddress(),
-                   memoryTracker.getOccupiedL1(), l1BudgetPerCore,
-                   result.cbPeakUsage);
-      if (!speculativeAddr) {
-        l1Size = handleNoFit(op, pos, data, opL1Usage, l1Size);
-        speculativeAddr = memoryTracker.wouldAllocateAt(l1Size);
-      }
-      if (l1Size > 0 && speculativeAddr &&
-          wouldCBsOverlapTensors(op, pos, result.cbPeakUsage,
-                                 *speculativeAddr)) {
-        l1Size = handleFragmentation(op, pos, data, opL1Usage,
-                                     result.cbPeakUsage, l1Size);
-      }
+      l1Size =
+          ensureFitsL1(op, pos, data, opL1Usage, result.cbPeakUsage, l1Size);
     }
     if (l1Size > 0) {
       addResultsToLiveSet(l1Size);
@@ -980,22 +980,8 @@ void L1SpillManagement<MemoryTracker>::run() {
                    "cbPeakUsage={1}, outputL1={2} bytes",
                    ttmlir::opToString(op), result.cbPeakUsage, l1Size);
 
-      // Can the output fit contiguously in the free list?
-      auto speculativeOutputAddr = memoryTracker.wouldAllocateAt(l1Size);
-      if (!speculativeOutputAddr) {
-        l1Size = handleNoFit(op, pos, data, opL1Usage, l1Size);
-        speculativeOutputAddr = memoryTracker.wouldAllocateAt(l1Size);
-      }
-
-      // CB fragmentation check: would CBs overlap any tensor?
-      if (l1Size > 0 && speculativeOutputAddr &&
-          wouldCBsOverlapTensors(op, pos, result.cbPeakUsage,
-                                 *speculativeOutputAddr)) {
-        l1Size = handleFragmentation(op, pos, data, opL1Usage,
-                                     result.cbPeakUsage, l1Size);
-      }
-
-      // Add tensor results to live set (0 means demoted to DRAM, skip).
+      l1Size =
+          ensureFitsL1(op, pos, data, opL1Usage, result.cbPeakUsage, l1Size);
       if (l1Size == 0) {
         continue;
       }
--- a/lib/Dialect/TTNN/Pipelines/TTNNPipelines.cpp
+++ b/lib/Dialect/TTNN/Pipelines/TTNNPipelines.cpp
@@ -113,7 +113,24 @@ void createTTNNPipelineAnalysisPasses(
     ttnn::TTNNOperationValidationAndFallbackOptions validationOptions;
     validationOptions.maxFallbackAttempts = options.maxFallbackAttempts;
 
-    if (options.enableGreedyOptimizer) {
+    if (!options.enableGreedyOptimizer) {
+      // Default: chain-based TTNNOptimizer.
+      ttnn::TTNNOptimizerOptions optimizerOptions(options);
+      pm.addPass(createDevicePassesWrapper(
+          [optimizerOptions, validationOptions](OpPassManager &innerPm) {
+            innerPm.addPass(
+                mlir::tt::ttnn::createTTNNRowMajorLayoutPropagation());
+            innerPm.addPass(
+                mlir::tt::ttnn::createTTNNOptimizer(optimizerOptions));
+            innerPm.addPass(mlir::createCanonicalizerPass());
+            innerPm.addPass(
+                mlir::tt::ttnn::createTTNNOperationValidationAndFallback(
+                    validationOptions));
+            innerPm.addPass(
+                mlir::tt::ttnn::createTTNNPrepareConv2dWeightsAndBias());
+          },
+          wrapperOptions));
+    } else {
       // Greedy optimizer: two new passes replace TTNNOptimizer.
       TTNNGreedyMemoryLayoutPropagationPipelineOptions propagationOptions;
       propagationOptions.maxLegalLayouts = options.maxLegalLayouts;
@@ -137,24 +154,6 @@ void createTTNNPipelineAnalysisPasses(
               innerPm.addPass(
                   mlir::tt::ttnn::createTTNNGreedyL1SpillManagement());
             }
-            innerPm.addPass(mlir::createCanonicalizerPass());
-            innerPm.addPass(
-                mlir::tt::ttnn::createTTNNOperationValidationAndFallback(
-                    validationOptions));
-            innerPm.addPass(
-                mlir::tt::ttnn::createTTNNPrepareConv2dWeightsAndBias());
-          },
-          wrapperOptions));
-    } else {
-      // Default: chain-based TTNNOptimizer.
-      ttnn::TTNNOptimizerOptions optimizerOptions(options);
-      pm.addPass(createDevicePassesWrapper(
-          [optimizerOptions, validationOptions](OpPassManager &innerPm) {
-            // All Optimizer passes will be run inside the wrapper.
-            innerPm.addPass(
-                mlir::tt::ttnn::createTTNNRowMajorLayoutPropagation());
-            innerPm.addPass(
-                mlir::tt::ttnn::createTTNNOptimizer(optimizerOptions));
             innerPm.addPass(mlir::createCanonicalizerPass());
             innerPm.addPass(
                 mlir::tt::ttnn::createTTNNOperationValidationAndFallback(
--- a/test/ttmlir/Dialect/TTNN/optimizer/greedy/beam_spill_with_reshards.mlir
+++ b/test/ttmlir/Dialect/TTNN/optimizer/greedy/beam_spill_with_reshards.mlir
@@ -2,20 +2,22 @@
 // RUN: ttmlir-opt --ttir-to-ttnn-backend-pipeline="enable-greedy-optimizer=true tensor-l1-usage-cap=0.01" -o %t %s
 // RUN: FileCheck %s --input-file=%t
 //
-// Test: beam search + reshards under tight L1 budget.
-// Verify L1 spill management works correctly with reshard ops.
+// Test: L1 spill management under tight L1 budget (1% cap).
+// Verify ops are demoted to DRAM and the pass produces valid TTNN IR.
+
+// Verify outputs use DRAM buffer type under 1% cap.
+// CHECK: #dram = #ttnn.buffer_type<dram>
 
 module attributes {} {
   func.func @forward(%arg0: tensor<512x512xbf16>, %arg1: tensor<512x512xbf16>,
                      %arg2: tensor<512x512xbf16>) -> tensor<512x512xbf16> {
-    // CHECK: "ttnn.add"{{.*}} -> tensor<512x512xbf16, #{{.*}}>
+    // CHECK: "ttnn.add"{{.*}} -> tensor<512x512xbf16, #ttnn_layout>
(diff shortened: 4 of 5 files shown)


lib/Dialect/TTNN/Analysis/L1SpillManagement.cpp
```diff
@@ -0,0 +1,1321 @@
...
+
+  // Restore snapshot taken before the victim's allocation.
+  auto snapIt = addressSnapshots.find(allocIdx);
+  assert(snapIt != addressSnapshots.end() &&
+         "snapshot not found for evicted tensor");
+  memoryTracker.restoreSnapshot(snapIt->second);
+
+  // Replay events from the alloc point forward, skipping evicted tensors.
+  // Update snapshots during replay so future evictions see accurate state.
+  for (size_t i = allocIdx; i < l1EventLog.size(); ++i) {
```
odjuricicTT, line 636: For my understanding, currently you only replay the alloc / dealloc simulation. If the real metal allocator was used used here, we would have to revalidate every op, not just allocations?

lib/Dialect/TTNN/Analysis/L1SpillManagement.cpp
```diff
@@ -0,0 +1,1321 @@
...
+      }
+
+      // Add tensor results to live set (0 means demoted to DRAM, skip).
+      if (l1Size == 0) {
+        continue;
+      }
+      addResultsToLiveSet(l1Size);
+      observer_->onLiveAdded(op, pos, l1Size, pos,
+                             memoryTracker.getOccupiedL1());
+
```
odjuricicTT, line 991: Is this the same logic as in handleOOM?

lib/Dialect/TTNN/Analysis/L1SpillManagement.cpp
rpavlovicTT: Correct — currently the replay is address-simulation only: it restores a free-list snapshot and replays alloc/dealloc events to keep wouldAllocateAt() accurate after an eviction. 
`GetOpConstraints` is not re-called because today it doesn't see allocator state. Unfortunately this will be a limitation due to coupling of allocation states with op validation path.

rpavlovicTT: So yeah we will have to revalidate every op in between `allocIdx` and latest processed op. We will have to see if this slows us down significantly.

lib/Dialect/TTNN/Analysis/L1SpillManagement.cpp
rpavlovicTT: Yes it is same sequence of steps. I can make it nicer.

lib/Dialect/TTNN/Pipelines/TTNNPipelines.cpp
```diff
@@ -114,21 +113,57 @@ void createTTNNPipelineAnalysisPasses(
...
+                mlir::tt::ttnn::createTTNNOptimizer(optimizerOptions));
+            innerPm.addPass(mlir::createCanonicalizerPass());
+            innerPm.addPass(
+                mlir::tt::ttnn::createTTNNOperationValidationAndFallback(
+                    validationOptions));
+            innerPm.addPass(
+                mlir::tt::ttnn::createTTNNPrepareConv2dWeightsAndBias());
+          },
+          wrapperOptions));
+    }
```
svuckovicTT: Consider inverting the condition - shorter-block-first is more readable.

include/ttmlir/Dialect/TTNN/Analysis/L1SpillManagement.h
```diff
@@ -0,0 +1,313 @@
...
+  /// Precomputed CB fragmentation cushion (bytes). See kCBFragCushionFraction.
+  uint64_t cbFragCushion;
+
+  /// Observer (NullObject pattern: always non-null).
+  std::unique_ptr<L1SpillObserver> observer_;
+
+  /// Pluggable memory state tracker.
+  MemoryTracker memoryTracker;
+
+  /// Belady eviction ordering: max-heap by lastUsePosition, keyed by Value.
```
bmalesevicTT, line 146: The implemented algorithm is not Belady's for multiple consumer cases(fork ops), in the PR desc Belady's algorithm is correctly described as evicting the one with largest lastNextUse, while the algorithm implemented and described here evicts the one with largest lastLastUse. If we had 2 slots, and A B C A B B A, when inserting C, we would evict B if we use Belady's but we would evict A if we used the current implementation. 
This implementation might be better for our case, but I think calling it Belady's everywhere and wrong PR desc are misleading...let's align this somehow please, for easier debugging and maintenance

test/ttmlir/Dialect/TTNN/optimizer/greedy/beam_spill_with_reshards.mlir
bmalesevicTT, line 1: Why is this one "beam spill with reshards" and the next one "bellady spill" when they are almost the same?
They test nearly the same IR (one just has an extra relu).
Neither test actually differentiates between "beam" vs. "Belady" behavior or verifies reshards.

lib/Dialect/TTNN/Pipelines/TTNNPipelines.cpp
```diff
@@ -114,21 +113,56 @@ void createTTNNPipelineAnalysisPasses(
...
+            innerPm.addPass(
+                mlir::tt::ttnn::createTTNNPrepareConv2dWeightsAndBias());
+          },
+          wrapperOptions));
+    } else {
+      // Greedy optimizer: two new passes replace TTNNOptimizer.
+      TTNNGreedyMemoryLayoutPropagationPipelineOptions propagationOptions;
+      propagationOptions.maxLegalLayouts = options.maxLegalLayouts;
+      propagationOptions.rowMajorEnabled = options.rowMajorEnabled;
+      propagationOptions.beamWidth = 8;
```
bmalesevicTT, line 138: Shouldn't this be a pipeline option as well? It is in an option in `include/ttmlir/Dialect/TTNN/Transforms/Passes.td` with a default value of 8 anyway so the hardcoded value here isn't needed?

lib/Dialect/TTNN/Analysis/L1SpillManagement.cpp
```diff
@@ -0,0 +1,1307 @@
...
+  bool resolved = evictUntil(pos, data, [&]() {
+    return memoryTracker.wouldAllocateAt(outputL1Size).has_value();
+  });
+  if (!resolved) {
+    demoteToDram(op);
+    evictForDramCBGrowth(op, pos, data);
+    TTMLIR_DEBUG(ttmlir::LogComponent::GreedyOptimizer,
+                 "    NO_FIT: eviction exhausted, demoting to DRAM");
+    return 0;
+  }
```
bmalesevicTT, line 731: Curious, is there ever a case where this is expected? Seems to me it would imply a logic bug rather than a legitimate runtime condition. If we evict everything, than the op should be guaranteed to fit because it was validated in isolation in LayoutPropagation?

lib/Dialect/TTNN/Transforms/OptimizerPasses/GreedyL1SpillManagement.cpp
```diff
@@ -0,0 +1,61 @@
...
+      TTNNGreedyL1SpillManagement>::TTNNGreedyL1SpillManagementBase;
+
+  void runOnOperation() final {
+    ModuleOp moduleOp = getOperation();
+
+    // Get L1 budget from system description and usage cap.
+    ttcore::GridAttr deviceGrid =
+        ttcore::lookupDevice(moduleOp).getWorkerGrid();
+    ttcore::ChipDescAttr chipDesc = ttcore::getOpChipDescAttr(moduleOp);
+    float tensorL1UsageCap = utils::getTensorL1UsageCap(moduleOp);
```
bmalesevicTT, line 40: Does the cap still make sense to exist, as internal fragmentation and CB buffers now have their own, explicit, handling?

lib/Dialect/TTNN/Analysis/L1SpillManagement.cpp
```diff
@@ -0,0 +1,1307 @@
...
+  TTMLIR_DEBUG(ttmlir::LogComponent::GreedyOptimizer,
+               "    OOM: validation failed, trying demotion/eviction");
+
+  auto inputLayouts = utils::extractInputLayouts(op);
+  auto config = extractOpConfigFromIR(op);
+
+  // Evict from live set (Belady: farthest last-use first).
+  // Uses evictUntil which inserts reshards for already-processed consumers
+  // instead of cascade revalidation, preserving downstream sharded layouts.
+  auto result = memoryTracker.validate(op, inputLayouts, config);
```
bmalesevicTT, line 567: We already validated the exact same inputLayouts and config right before calling this function, this first check seems redundant and increases compile time?

lib/Dialect/TTNN/Analysis/L1SpillManagement.cpp
```diff
@@ -0,0 +1,1307 @@
...
+    Operation *op = data.schedule[pos];
+
+    processDeadTensors(pos, data);
+
+    // Ops with L1 output annotation get full processing.
+    // DRAM-output ops (no annotation) still need CB overlap checking against
+    // live L1 tensors -- skip only if the op can't be validated or there are
+    // no live L1 tensors that could clash.
+    auto l1Attr = op->getAttrOfType<IntegerAttr>("ttnn.output_l1_usage");
+    uint64_t opL1Usage = l1Attr ? l1Attr.getValue().getZExtValue() : 0;
```
bmalesevicTT, line 906: By this name I would assume this is the internal usage needed for compute or the sum of all usage, but is actually opOutputL1Usage

lib/Dialect/TTNN/Analysis/L1SpillManagement.cpp
```diff
@@ -0,0 +1,1307 @@
...
+                   "    NOT_IMPLEMENTED: evicting all live L1 tensors for {0}",
+                   ttmlir::opToString(op));
+      evictAllFromL1(pos, data);
+      ++spillCount;
+      continue;
+    }
+
+    if (result.isSuccess()) {
+      uint64_t l1Size =
+          result.outputL1Usage > 0 ? result.outputL1Usage : opL1Usage;
```
bmalesevicTT, line 976: Why would we do this? If `result.outputL1Usage=0` it either means OpModel usage calculation is wrong or the tensor isn't truly placed in L1 by backend. Overriding it with opL1Usage means either our attribute in IR isn't valid but we count as if it is, or we mask bad OpModel usage calculation

lib/Dialect/TTNN/Analysis/L1SpillManagement.cpp
```diff
@@ -0,0 +1,1307 @@
...
+      TTMLIR_DEBUG(ttmlir::LogComponent::GreedyOptimizer,
+                   "    BACKEND_ERROR at pos {0} for {1}: {2}. "
+                   "Spilling to DRAM.",
+                   pos, ttmlir::opToString(op), result.errorMessage);
+      for (auto r : tensorResults) {
+        spillToDram(r);
+      }
+      ++spillCount;
+      continue;
+    }
```
bmalesevicTT, line 1013: Maybe comment could be a bit clearer as it initially confused me. I presume we spill to DRAM not because of "crashing" but because of the mismatch it would create otherwise where outputs are in L1 in IR but are not tracked during this pass.

lib/Dialect/TTNN/Analysis/L1SpillManagement.cpp
```diff
@@ -0,0 +1,1307 @@
...
+
+    size_t ri = opResult.getResultNumber();
+    TTNNLayoutAttr resultLayout = (ri < result.actualOutputLayouts.size())
+                                      ? result.actualOutputLayouts[ri]
+                                      : chosenLayout;
+    if (!resultLayout) {
+      continue;
+    }
```
bmalesevicTT, line 317: Shouldn't this be an assert? Why would we allow a result without info about its layout, it could cause inconsistencies?

lib/Dialect/TTNN/Analysis/L1SpillManagement.cpp
```diff
@@ -0,0 +1,1307 @@
...
+  for (auto opResult : op->getResults()) {
+    auto tensorType = mlir::dyn_cast<RankedTensorType>(opResult.getType());
+    if (!tensorType) {
+      continue;
+    }
+
+    size_t ri = opResult.getResultNumber();
+    TTNNLayoutAttr resultLayout = (ri < result.actualOutputLayouts.size())
+                                      ? result.actualOutputLayouts[ri]
+                                      : chosenLayout;
```
bmalesevicTT, line 314: Why would we use the first output if not enough of them are gotten from backend compared to IR, shouldn't this be an assert?

lib/Dialect/TTNN/Analysis/L1SpillManagement.cpp
```diff
@@ -0,0 +1,1307 @@
...
+    ttcore::DataTypeAttr newDataTypeAttr =
+        ttcore::DataTypeAttr::get(op->getContext(), chosenLayout.getDataType());
+    dtypeOp.setDtypeAttr(newDataTypeAttr);
+  }
+
+  // Update L1 usage attribute.
+  if (chosenLayout.hasL1BufferType() && result.outputL1Usage > 0) {
+    OpBuilder builder(op->getContext());
+    op->setAttr("ttnn.output_l1_usage",
+                builder.getI64IntegerAttr(result.outputL1Usage));
```
bmalesevicTT, line 350: Curious, will we ever change this attr here, rather than overwrite it with the same value it already set in LayoutPropagation? Would the only case be: spilling some inputs to DRAM affected that outputL1Usage is now different than it was calculated in LayoutPropagation?

lib/Dialect/TTNN/Analysis/L1SpillManagement.cpp
```diff
@@ -0,0 +1,1307 @@
...
+    // Insert reshards for already-processed consumers instead of cascade
+    // revalidation. This preserves downstream ops' sharded layouts.
+    if (originalL1Layout.hasL1BufferType()) {
+      for (Operation *consumer : collectDownstreamConsumers(victimOp)) {
+        auto posIt = data.positionMap.find(consumer);
+        if (posIt == data.positionMap.end() || posIt->second >= pos) {
+          continue;
+        }
+        if (!mlir::dyn_cast<OpModel>(consumer)) {
+          continue;
```
bmalesevicTT, line 689: Do we want to silently pass here or rather make an assert or at least a log of some sort? If original had L1 input and now it has DRAM input, we wouldn't notice the change while it might break something

lib/Dialect/TTNN/Analysis/L1SpillManagement.cpp
```diff
@@ -0,0 +1,1307 @@
...
+template <typename MemoryTracker>
+uint64_t L1SpillManagement<MemoryTracker>::ensureFitsL1(
+    Operation *op, int64_t pos, const ScheduleData &data, uint64_t opL1Usage,
+    uint64_t cbPeakUsage, uint64_t l1Size) {
+  auto speculativeAddr = memoryTracker.wouldAllocateAt(l1Size);
+  if (!speculativeAddr) {
+    l1Size = handleNoFit(op, pos, data, opL1Usage, l1Size);
+    speculativeAddr = memoryTracker.wouldAllocateAt(l1Size);
+  }
+  if (l1Size > 0 && speculativeAddr &&
```
bmalesevicTT, line 541: `l1Size > 0` makes sure that we don't use the speculativeAddr when it doesn't actually use L1 for output, BUT it also prevents checking for CBOverlap on this op. Whenever we demoteToDRAM, we also check evictForDRAMCBGrowth as it could still overlap with tensors which are not the output of the current op, so it should be checked here too, we might still need some evictions because of it

lib/Dialect/TTNN/Analysis/L1SpillManagement.cpp
```diff
@@ -0,0 +1,1307 @@
...
+      inputLayouts = utils::extractInputLayouts(op);
+      result = memoryTracker.validate(op, inputLayouts, config);
+      return result.isSuccess();
+    });
+  }
+
+  if (result.isSuccess()) {
+    uint64_t l1Size =
+        result.outputL1Usage > 0 ? result.outputL1Usage : opL1Usage;
+    if (l1Size > 0) {
```
bmalesevicTT, line 579: `l1Size > 0` prevents checking for CBOverlap on this op after OOM is fixed and validation succeeds with output in DRAM. Whenever we demoteToDRAM in ensureFitsL1, we also check evictForDRAMCBGrowth as it could still overlap with tensors which are not the output of the current op, so it should be checked here too.

lib/Dialect/TTNN/Analysis/L1SpillManagement.cpp
```diff
@@ -0,0 +1,1307 @@
...
+                   liveValues.size());
+    }
+  } else {
+    // Stage 3: Op exceeds budget alone -- spill all results to DRAM.
+    observer_->onSelfSpill(op, pos);
+    TTMLIR_DEBUG(ttmlir::LogComponent::GreedyOptimizer,
+                 "    SPILL SELF: op exceeds budget alone");
+    for (auto r : tensorResults) {
+      spillToDram(r);
+    }
```
bmalesevicTT, line 600: Why would we spill here instead of demoteToDram? By spilling we still have an OOM situation on the op itself

lib/Dialect/TTNN/Analysis/L1SpillManagement.cpp
bmalesevicTT: In many other places we do this as well, noting here not to repeat the same concern at each occurrence

lib/Dialect/TTNN/Analysis/L1SpillManagement.cpp
```diff
@@ -0,0 +1,1307 @@
...
+      uint64_t l1Size =
+          result.outputL1Usage > 0 ? result.outputL1Usage : opL1Usage;
+
+      TTMLIR_DEBUG(ttmlir::LogComponent::GreedyOptimizer,
+                   "    VALIDATION SUCCESS: op {0}, "
+                   "cbPeakUsage={1}, outputL1={2} bytes",
+                   ttmlir::opToString(op), result.cbPeakUsage, l1Size);
+
+      l1Size =
+          ensureFitsL1(op, pos, data, opL1Usage, result.cbPeakUsage, l1Size);
```
bmalesevicTT, line 984: Why do we need both opL1Usage and l1Size, and why can they differ at all when they both represent opOutputL1Size/Usage, quite confusing naming as well, and as earlier mentioned it seems to be either dead code or masking a bug?

PR #7687, opened 2026-03-27
Greedy optimizer: decision trace and compile-time stats

rpavlovicTT:
## Summary

- Add `DecisionTrace` infrastructure for recording greedy optimizer decisions to JSON
- Add `CompileTimeStatsObserver` for per-op compile-time profiling
- Wire `enableDecisionTrace`, `decisionTraceDir`, `enableCompileTimeStats` pipeline options through both greedy passes

## Components

- **DecisionTrace** — records per-op beam candidates, fork resolutions, edge reshards, spill events, and final layout choices. Serializes to JSON for offline analysis and visualization.
- **DecisionTraceObserver** — implements both `LayoutPropagationObserver` and `L1SpillObserver` interfaces, populating a `DecisionTrace` during pass execution.
- **CompileTimeStatsObserver** — `LayoutPropagationObserver` that prints per-op compile-time statistics at DEBUG level for profiling optimizer bottlenecks.

## Pipeline options

- `--enable-decision-trace` — enable JSON trace output
- `--decision-trace-dir` — output directory (default: `ttrt-artifacts/decision_trace`)
- `--enable-compile-time-stats` — print per-op timing at DEBUG level

## Test plan

- [ ] Build with `cmake --build build`
- [ ] Run with `--enable-decision-trace` and verify JSON output
- [ ] Run with `--enable-compile-time-stats` and verify DEBUG output

Stacks on #7686

### DIFF
--- a/include/ttmlir/Dialect/TTNN/Diagnostics/CompileTimeStatsObserver.h
+++ b/include/ttmlir/Dialect/TTNN/Diagnostics/CompileTimeStatsObserver.h
@@ -6,8 +6,8 @@
 #define TTMLIR_DIALECT_TTNN_DIAGNOSTICS_COMPILETIMESTATSOBSERVER_H
 
 #include "ttmlir/Dialect/TTNN/Analysis/MemoryLayoutPropagationObserver.h"
+#include "ttmlir/Support/Logger.h"
 
-#include "llvm/Support/Debug.h"
 #include "llvm/Support/Format.h"
 #include "llvm/Support/raw_ostream.h"
 
@@ -77,7 +77,8 @@ public:
       totalValid += s.validResults;
     }
 
-    llvm::raw_ostream &os = llvm::errs();
+    std::string statsStr;
+    llvm::raw_string_ostream os(statsStr);
     os << "\n=== Greedy Optimizer Compile-Time Stats (func: " << funcName_
        << ") ===\n";
     os << "Total: " << llvm::format("%.0f", totalMs) << "ms across "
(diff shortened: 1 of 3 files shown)


include/ttmlir/Dialect/TTNN/Analysis/CompileTimeStatsObserver.h
```diff
@@ -0,0 +1,116 @@
+// SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
```
odjuricicTT: Maybe have a seperate dir for these since it's not analysis and it's logging only.

include/ttmlir/Support/Logger.h
```diff
@@ -63,6 +63,8 @@ inline constexpr const char *getLogComponentStr(LogComponent type) {
     return "df-sharding-policy";
   case LogComponent::OpValidation:
     return "op-validation";
+  case LogComponent::ValidationFallback:
```
svuckovicTT, line 66:
```suggestion
  case LogComponent::ValidationAndFallback:
```

rpavlovicTT: Will rename in a separate commit

lib/Dialect/TTNN/Diagnostics/DecisionTrace.cpp
```diff
@@ -0,0 +1,718 @@
...
+  std::error_code ec;
+  llvm::raw_fd_ostream os(path, ec);
+  if (ec) {
+    llvm::errs() << "DecisionTrace: failed to write merged trace to " << path
+                 << ": " << ec.message() << "\n";
+    return false;
+  }
+  os << llvm::formatv("{0:2}", llvm::json::Value(std::move(*root)));
+  return true;
+}
```
bmalesevicTT, line 367: Isn't this overcomplicating it a bit? These two passes already run in the same pipeline scope, so sharing an in-memory object is simpler and more reliable than serializing to disk between them. Sharing a DecisionTrace object there would just make the implicit coupling explicit?

PR #7688, opened 2026-03-27
[d2m] lower constants to fill tile

### LINKED ISSUE #4756: [D2M] support lowering `ttir.constant` to kernel op `fill_tile`
As per @wenbinlyuTT [here](https://github.com/tenstorrent/tt-mlir/pull/4733#issuecomment-3240305905)

The benefits are: 

- We don't need to store the entire tensor in the flatbuffer, reducing its size.
- The constant tensor can be created on the device with the required shape, no need to worry about host side alignment and we save the data transfer cost.
- The creation & fill of the constant tensor can potentially happen after our auto-inserted tilization as an optimization, since it doesn't actually need any tilization anyway. If we optimize hard enough, we can even generate a single CB tile right there in the kernel and reuse that tile for the entire op!

dlokeTT:
### Ticket
Closes #4756 

### Problem description
We should be able to support creating constants on device. The `fill_tile` llks can be used for zero initializing accumulators for matmuls and reductions

### What's changed
- Renamed `d2m.tile_fill` to `d2m.fill_tile` to match ttkernel op name
- Lower `ttir.full` to a d2m.generic with no inputs and one output tensor with a linalg generic that uses `d2m.fill_tile`
- Updated `scalarizeConstTensors` to pick up on the new lowering of `ttir.full` so it erases the fill_tile generic when ops have scalar variants
- Removed experimental `ttkernel::TileFill` op which was basically a copy of llk maintained compute kernel api `fill_tile`
-  Added int version of `fill_tile` to ttkernel
- Added proper handing for d2m to ttkernel for `fill_tile` and the different ttkernel ops it can lower to based on dtype
- Added builder tests for ttmetal path testing creating a constant and then performing unary, binary and ternary ops with this const tensor. Test f32, bf16 and i32 dtypes
- `ttnn.generic` op has a runtime assertion guaranteeing at least one input and one output tensor but this assumption fails for generator style ops so added a hack to just push the output tensor twice to the IO list to pass this assertion. Probably should fix this in ttnn generic runtime
- Updated lit tests to use `d2m.fill_tile`

### Checklist

### DIFF
--- a/lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
+++ b/lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
@@ -77,27 +77,21 @@ static std::pair<Value, Value> getMcastEndCoords(PatternRewriter &rewriter,
 }
 
 static Value getCB(ConversionPatternRewriter &rewriter, Value cb) {
-  if (memref::LoadOp loadOp =
-          mlir::dyn_cast<memref::LoadOp>(cb.getDefiningOp());
-      loadOp) {
+  if (auto loadOp = cb.getDefiningOp<memref::LoadOp>()) {
     assert(loadOp.getIndices().size() == 1 &&
            "Expected single index in load op, failing.");
-    return rewriter.getRemappedValue(loadOp.getMemref());
+    return rewriter.getRemappedValue(loadOp.getMemRef());
   }
 
-  if (auto affLoad =
-          mlir::dyn_cast_if_present<affine::AffineLoadOp>(cb.getDefiningOp())) {
+  if (auto affLoad = cb.getDefiningOp<affine::AffineLoadOp>()) {
     return rewriter.getRemappedValue(affLoad.getMemRef());
   }
 
-  if (mlir::isa<memref::SubViewOp>(cb.getDefiningOp())) {
-    memref::SubViewOp subViewOp =
-        mlir::cast<memref::SubViewOp>(cb.getDefiningOp());
+  if (auto subViewOp = cb.getDefiningOp<memref::SubViewOp>()) {
     return rewriter.getRemappedValue(subViewOp.getSource());
   }
 
-  if (mlir::isa<memref::CastOp>(cb.getDefiningOp())) {
-    memref::CastOp castOp = mlir::cast<memref::CastOp>(cb.getDefiningOp());
+  if (auto castOp = cb.getDefiningOp<memref::CastOp>()) {
     return rewriter.getRemappedValue(castOp.getSource());
   }
   llvm_unreachable("Expected load or subview op");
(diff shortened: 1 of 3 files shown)


lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
```diff
@@ -161,42 +166,92 @@ static void ensureDominatesInsertionPoint(OpBuilder &rewriter, Value value) {
...
+// Remapped L1 CB for a tile loaded from L1; null if the value is not an L1 load
+// (including DST-only producers such as fill_tile).
+static Value tryGetL1CbFromTensorShardLoad(ConversionPatternRewriter &rewriter,
+                                           Value v) {
+  Operation *def = v.getDefiningOp();
+  if (!def) {
+    return nullptr;
+  }
+  Value memref;
+  if (auto load = mlir::dyn_cast<memref::LoadOp>(def)) {
```
wenbinlyuTT: I think `v.getDefiningOp<memref::LoadOp>()` like in `operandFromDst()` can also work here. Same for the `affLoad` change above.

lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
```diff
@@ -161,42 +166,92 @@ static void ensureDominatesInsertionPoint(OpBuilder &rewriter, Value value) {
...
+  }
+  if (ttcore::getMemorySpace(memref) != ttcore::MemorySpace::DeviceL1) {
+    return nullptr;
+  }
+  return rewriter.getRemappedValue(memref);
+}
+
+// Walk the enclosing host `func.func` for the first L1 load / store (memref or
+// affine). Used when operand-local CB resolution is insufficient.
+static Value findFirstL1InputCbMemref(func::FuncOp func) {
```
wenbinlyuTT: This function can still be templated like the old `getInOrOutCB()`, just on two types: template <typename MemrefLoadOrStoreOp, typename AffineLoadOrStoreOp>.

test/python/golden/test_metal_constants.py
```diff
@@ -0,0 +1,196 @@
...
+
+from conftest import get_request_kwargs
+from test_utils import SkipIf, shape_str
+
+from builder.base.builder_utils import Operand
+from builder.ttir.ttir_builder import TTIRBuilder
+from builder.base.builder_apis import compile_and_execute_ttir
+
+pytestmark = pytest.mark.frontend("ttir")
+torch.manual_seed(0)
```
wenbinlyuTT: Debug leftover?

test/python/golden/test_metal_constants.py
```diff
@@ -0,0 +1,196 @@
...
+
+    def module(builder: TTIRBuilder):
+        @builder.func([shape], [dtype])
+        def constant_binary(
+            in0: Operand,
+            builder: TTIRBuilder,
+            unit_attrs: List[str] = None,
+        ):
+            c = builder.constant(floor, unit_attrs=unit_attrs)
+            return builder.maximum(in0, c, unit_attrs=unit_attrs)
```
wenbinlyuTT, line 139: Let's randomize the input and put the constant near the middle of the range, like what you already do in the ternary tests.

lib/Dialect/D2M/Transforms/ScalarizeConstTensors.cpp
```diff
@@ -19,7 +19,9 @@
 #include "mlir/IR/PatternMatch.h"
 #include "mlir/Transforms/GreedyPatternRewriteDriver.h"
```
wenbinlyuTT: Nit: no empty line.

Issue #7689, opened 2026-03-27
Implement a skill that adds TTNN ops


svuckovicTT:
Investigate implementing a Claude skill that will add support for TTNN ops to the compiler, which includes dialect definitions, conversion patterns, flatbuffer schemas, opmodel logic, workarounds, etc.

PR #7690, opened 2026-03-27
[Claude skill] Add add-op skill

svuckovicTT:
### Ticket
Closes https://github.com/tenstorrent/tt-mlir/issues/7689

### Problem description
Investigating whether TTNN ops can be added to the compiler via a skill.

### What's changed
Added an `add-op` skill.

The skill identifies the specific C++ TTNN API to be used for the op (when more than 1 signature present), and then does its best to implement the skill throughout the stack. It can implement multiple ops at once! It does not perform any fusing currently, that's TBD (maybe a separate skill), but everything else from SHLO, to TTNN, optimizer, emitted TTNN, ttir-builder, etc. is taken care of.

Upon completing all the steps, the skill launches a webserver to help the user understand what's going on: it shows some basic info on ops added, runs all the tests, reports on passes/failures, etc. Some screenshots below.

### Basic info:
image

### Tests:
image

### Emitted TTNN:
image

### Alchemist path:
image

### Checklist


.claude/skills/add-op/SKILL.md
```diff
@@ -0,0 +1,912 @@
+---
```
svuckovicTT, line 1: A lot of this can be simplified to use less text.

Overall, feels like adding details really does help the skill, but the current amount of it is overwhelming in a single file, and I've already cut down on some useful instructions. Maybe steps could be repackaged into separate .md files, then we could spawn them off in their own context and return info on the "level of success" of each step.

This would also allow for easier maintenance, e.g. optimizer folks could update the opmodel step(s), ttir builder could be maintained by tooling folks, etc.

No established best practices so I'm brainstorming out loud, opinions/ideas very welcome.

.claude/skills/add-op/SKILL.md
```diff
@@ -0,0 +1,912 @@
...
+### Build
+
+```bash
+source env/activate
+cmake --build build
+```
+
+If the build fails, fix the errors and rebuild before proceeding to tests.
+
+### Launch the review webserver
```
nsmithtt, line 833: Would it be better for Claude to just open a draft PR instead of all the web tools checked in here?

svuckovicTT: It's a question of iteration speed - the tool gives you the results in 15 seconds or less, and tests rerun with a refresh. Using CI would be far slower.

But I think I understand where you're coming from, this does feel a bit bloatware-ish. No obvious (non-trivial) solution comes to mind though when trying to think of running a local set of tests tied to a specific op or ops. Will think about this in the background, feel free to add suggestions if anything comes to mind!

PR #7706, opened 2026-03-30
[EmitPy] Refactor EmitPy CPU-hoisting

dmilinkovicTT:
### Ticket
#7742 

### Motivation

The CPU-hoisted path previously lowered ops through TTNN before converting to Python:

`TTIR → (preparation TTIR passes) → TTNN → EmitPy (golden mode) → ttnn.<op>.golden_function()`

This was problematic for the following reasons:

1. **TTNN is a hardware-oriented dialect, and represents a wrong abstraction layer for CPU-hoisting** — forcing CPU-hoisted ops through TTNN added unnecessary complexity for ops that just need to run on the host CPU.

2. **TTNN golden functions are inconsistent** — they're maintained by the tt-metal team and their signatures often don't match the ops themselves. This required a growing set of monkey-patch workarounds in `utils.py` (`perform_golden_workarounds`) to bridge the mismatches.

### What's changed

The new path converts TTIR ops directly to Python, bypassing TTNN entirely:

`TTIR → TTIRToEmitPyCPU → ttir_cpu.<op>()`

emitpy-cpu

**New `TTIRToEmitPyCPU` pass** (`lib/Conversion/TTIRToEmitPy/`) — converts TTIR ops to `emitpy.CallOpaqueOp` targeting `ttir_cpu.<mnemonic>` functions. Uses an `EmitPyCallBuilder` abstraction and templated patterns (unary, binary, reduction) for scalability.

**New `ttir_cpu.py`** (`tools/tt-alchemist/templates/python/local/`) — standalone pure-torch implementations of TTIR ops. Each public function matches a TTIR op mnemonic. No dependency on ttnn golden functions.

**Removed from TTNN→EmitPy path** — the `enableGoldenMode` flag, the `.golden_function` callee suffix, and all golden-mode plumbing. The `perform_golden_workarounds()` monkey-patching in `utils.py` is deleted.

**Simplified CPU pipeline** — the 4 TTIR prep passes and TTNN lowering are removed from `createTTIRToEmitPyCPUPipeline`.

### Commit-by-commit

1. Remove `enableGoldenMode` flag and `.golden_function` callee suffix from TTNNToEmitPy.
2. Add `TTIRToEmitPyCPU` pass with `EmitPyCallBuilder` and templated conversion patterns; rewire CPU pipeline.
3. Add `ttir_cpu.py` with pure-torch implementations of all CPU-hoisted ops.
4. Delete `perform_golden_workarounds()` from `utils.py`.
5. One-line fix in `builder_runtime.py`, so that `cpu_hoisted` functions are actually skipped during program execution
6. Add `"emitpy"` target to all CPU-hoisted Python golden tests. Add some missing tests.
7. Update lit tests to match new `ttir_cpu.*` call format.
8. Fix `memory_config` attr propagation in `EmitPyConversion.h` and `ttcore.device` propagation in `TTNNFileSplit.cpp`.

### Future work
- Unify the TTIR goldens - #7798 
- Fix CPU-hoisted function placement - #7797

### Checklist
- [x] MLIR lit tests updated
- [x] Python golden tests extended with `"emitpy"` target across all hoisted test cases

---
### tt-xla Validation

**Branch:** [`xla-validate/7706`](https://github.com/tenstorrent/tt-mlir/tree/xla-validate/7706)
**Base (uplifted):** `2a24a1fb4`

| Test Suite | Run | Status |
|------------|-----|--------|
| full-mlir-uplift-qualification | [Run #23929095048](https://github.com/tenstorrent/tt-xla/actions/runs/23929095048) | :hourglass: in_progress |

*Last validated: 2026-04-03 UTC*


lib/Conversion/TTIRToEmitPy/TTIRToEmitPyCPUPass.cpp
```diff
@@ -0,0 +1,1059 @@
...
+      resultTypes.push_back(typeConverter->convertType(t));
+    }
+    rewriter.replaceOpWithNewOp<emitpy::CallOpaqueOp>(
+        srcOp, resultTypes, callee, operands, rewriter.getArrayAttr(args),
+        rewriter.getArrayAttr(kwargNames));
+  }
+
+private:
+  Operation *srcOp;
+  ConversionPatternRewriter &rewriter;
```
vmilosevic: :warning: **cppcoreguidelines\-avoid\-const\-or\-ref\-data\-members** :warning:
member `` rewriter `` of type `` ConversionPatternRewriter & `` is a reference

dmilinkovicTT:
@sdjordjevicTT #7798

include/ttmlir/Conversion/Passes.td
```diff
@@ -193,13 +193,23 @@ def ConvertTTNNToEmitC : Pass<"convert-ttnn-to-emitc", "::mlir::ModuleOp"> {
   let dependentDialects = ["mlir::emitc::EmitCDialect", "mlir::tt::ttnn::TTNNDialect"];
 }
 
+def ConvertTTIRToEmitPyCPU : Pass<"convert-ttir-to-emitpy-cpu", "::mlir::ModuleOp"> {
```
svuckovicTT:
```suggestion
def ConvertTTIRCPUToEmitPy : Pass<"convert-ttir-cpu-to-emitpy", "::mlir::ModuleOp"> {
```

svuckovicTT: ^ just a suggestion - it's somewhat more clear to me this way, but will let you make the call

lib/Conversion/TTIRToEmitPy/TTIRCPUToEmitPyPass.cpp
```diff
@@ -0,0 +1,1256 @@
...
+    b.addKwarg("ceil_mode", adaptor.getCeilMode() ? "True" : "False");
+    b.addKwarg("count_include_pad",
+               adaptor.getCountIncludePad() ? "True" : "False");
+    b.replaceOp(rewriter);
+    return success();
+  }
+};
+
+// --- Conv2d ---
+class TTIRConv2dToEmitPy : public OpConversionPattern<ttir::Conv2dOp> {
```
svuckovicTT, line 1060: For all of these pattern conversions, consider using the approach used in ttnn->emitpy/c where we overload `emit<T>` for all the possible `T`s. It would be nice to have the same approach in all of these dialect conversions where we're doing just signature mapping, but it is a non-trivial chunk of work, so maybe sometime in the future, definitely don't block this PR on that.

lib/Conversion/TTIRToEmitPy/TTIRCPUToEmitPyPass.cpp
```diff
@@ -0,0 +1,1256 @@
...
+    if (adaptor.getWeight()) {
+      b.addOperand(adaptor.getWeight());
+    } else {
+      b.addKwarg("weight", "None");
+    }
+    if (adaptor.getBias()) {
+      b.addOperand(adaptor.getBias());
+    } else {
+      b.addKwarg("bias", "None");
+    }
```
amilovanovicTT, line 895: If both weight and bias are provided, then they are emitted without kwargs, which conflicts with the definition from ttir_cpu.py `def layer_norm(t, epsilon=1e-5, weight=None, bias=None, **_)`. However, this case when both weight and bias are provided is not covered in tests. Is there some particular reason for that?

tools/tt-alchemist/templates/python/local/ttir_cpu.py
```diff
@@ -0,0 +1,986 @@
+# SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
```
amilovanovicTT: Nit: 
```suggestion
# SPDX-FileCopyrightText: (c) 2026 Tenstorrent AI ULC
```

lib/Conversion/TTIRToEmitPy/TTIRCPUToEmitPyPass.cpp
```diff
@@ -0,0 +1,1256 @@
...
+                  ConversionPatternRewriter &rewriter) const override {
+    EmitPyCallBuilder b(op, getTypeConverter(), getCallee(op));
+    b.addOperand(adaptor.getInput());
+    b.addLiteral(std::to_string(adaptor.getDim()));
+    b.replaceOp(rewriter);
+    return success();
+  }
+};
+
+class TTIRUnsqueezeToEmitPy : public OpConversionPattern<ttir::UnsqueezeOp> {
```
amilovanovicTT, line 514: Nit:
```suggestion
// --- Unsqueeze ---
class TTIRUnsqueezeToEmitPy : public OpConversionPattern<ttir::UnsqueezeOp> {
```

amilovanovicTT (approved):
Looks great! Maybe you can just update https://github.com/tenstorrent/tt-mlir/blob/main/.claude/skills/add-op/SKILL.md to remove `enableGoldenMode` and `isGoldenModeEnabled()`.

include/ttmlir/Conversion/Passes.td
dmilinkovicTT: makes sense, adjusted

lib/Conversion/TTIRToEmitPy/TTIRCPUToEmitPyPass.cpp
dmilinkovicTT: I've deliberately chosen this less-robust approach because for TTIR, we operate on smaller subset of attributes that need complex logic for conversion, and the mappings are generally simpler compared to TTNN -> EmitPy conversion. But I'll think about this for a bit more, and possibly create a follow-up PR.

lib/Conversion/TTIRToEmitPy/TTIRCPUToEmitPyPass.cpp
dmilinkovicTT: Nice catch, thanks! Adjusted the pattern so that `kwargs` are always used, and adjusted lit and builder tests.

dmilinkovicTT:
@amilovanovicTT thanks! I'm working on generally expanding the mentioned skill so that CPU-hoisting is generally better covered. I'll have the PR soon.

PR #7723, opened 2026-03-31
[d2m] remove full op

dlokeTT:
### Problem description
We no longer need the `d2m.full` op since we can create tensors on device after https://github.com/tenstorrent/tt-mlir/pull/7688

### What's changed
Removed `d2m.full` op and everywhere it's found. Use `d2m.fill_tile` instead.
Also renamed `d2m.fill_tile` to `d2m.tile_fill` to follow naming convention of other d2m compute ops

### Checklist


lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
```diff
@@ -911,8 +1023,8 @@ class D2MNamedElementwiseRewriter final
 // ----------------------------------------------------------------------------
 //
 // Rewriting reduction ops is similar to the elementwise group except for
-// ops whose tiled counterparts require a scaler operand ('weights', etc).
-// This rewriter will emit a single tile scaler operand that will be
+// ops whose tiled counterparts require a scalar operand ('weights', etc).
```
vwellsTT: I wanna say scaler is a real term intentionally used here? not 100% sure tbh

lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
```diff
@@ -535,6 +535,118 @@ class D2MNamedRewriterCommon {
     return operands;
   }
 
+  /// Emit d2m.generic + linalg.generic + tile_fill + remote_store to produce
+  /// a ranked tensor of resultType filled with fillAttr. Sets outValue to the
```
vwellsTT: tbh this function signature seems unnecessarily convoluted to me; this seems like just a reasonably simple op replacement, and you just replace the ttir.full_op with the generic's result, no? maybe I'm missing something

lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
```diff
@@ -535,6 +535,118 @@ class D2MNamedRewriterCommon {
...
+    const std::size_t physicalRank =
+        ttcore::getDeviceLayout(output).getRank() / 2;
+
+    SmallVector<AffineMap> indexingMaps =
+        getIdentityAffineMapsArray(rewriter, 1, physicalRank);
+    auto parallel = ttcore::IteratorTypeAttr::get(
+        rewriter.getContext(), ttcore::IteratorType::Parallel);
+    SmallVector<Attribute> iteratorTypes(physicalRank, parallel);
+
+    auto generic = rewriter.create<d2m::GenericOp>(
```
vwellsTT, line 617: this code is pretty nasty; I wonder if we have/need a way to create the boiler-plates for a custom generic like this more compactly via helper func anywhere...

lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
dlokeTT: Yeah, I think you're right now that I'm looking at it again. Going to revert this

lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
dlokeTT: Yeah, this is op replacement in the case of a ttir.full but we also use this for reductions for the scaler that we multiply the reduction by which doesn't come from a ttir.full so I made it a helper func so this logic can be used  in both places

lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
dlokeTT: hmm yea doesn't look like there's any helpers. Could make sense to have a helper func for creating custom generics. I could do that here but then there'd only be one callsite. Maybe better to do in a follow on pr where the helper func is used more commonly and replaces more of the generic builders?

lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
dlokeTT: We'll probably also use a similar routine for reductions on outer dims as well

lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
vwellsTT: seems fine to me

lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
vwellsTT: helper function is fine, but this pattern is suboptimal.  Can you use: https://llvm.org/doxygen/classllvm_1_1FailureOr.html for this?

lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
dlokeTT: updated with helper

lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
dlokeTT: updated following this convention

PR #7728, opened 2026-03-31
Add support for composite SDPA

mstojkovicTT:
### Ticket
https://github.com/tenstorrent/tt-xla/pull/4020

### Problem description
Described in PR.

### What's changed
Added support and test for composite SDPA.


mstojkovicTT:
> We are processing `sdpa` op as `CustomCallOp`. Why do we need additional compositeOp?

If my understanding is correct, we are using these CustomCalls for vllm models, but we need to support other SDPA patterns from transformers lib. Composite ops are torch_xla native way of wrapping operations like SDPA, and lowering them trough SHLO (and shardy works on them) to TTIR.

ajakovljevicTT (approved):

PR #7653, opened 2026-03-25
Deprecate ttrt perf

nsmithtt:
We are adopting a new path for collecting perf reports in ttrt via tt-metal's tracy package. This gets us out of the business of having to maintain a complicated set of scripts that need to be perfectly aligned with the tt-metal project.

```bash
python -m tracy -r -v --output-folder prof -m ttrt run ...
```

PR #7730, opened 2026-03-31
[runtime] add per-op tracy zones (ttnn)

pilkicTT:
Add `tracy` profiling zones in `ttnn` executor.

- adds zone for `submit` - used for tracking all host-side activity when submitting a program for execution
- adds `program_execute` zone for each individual program we execute (program name is added to the zone)
- adds op zone, which is named after the op we're executing, i.e. `MatmulOp`, `EltwiseBinaryOp`, etc.

Ran perf benchmark on `tt-xla` (resnet) - no perf regression observed.

### DIFF
--- a/runtime/lib/ttnn/program_executor.cpp
+++ b/runtime/lib/ttnn/program_executor.cpp
@@ -104,6 +104,10 @@
 #include "tt/runtime/perf.h"
 #include "tt/runtime/utils.h"
 
+#include "tracy/Tracy.hpp"
+
+#include <cstring>
+
 namespace tt::runtime::ttnn {
 
 using LogType = ::tt::runtime::logger::LogType;
@@ -157,11 +161,15 @@ void ProgramExecutor::runCallback(
 }
 
 void ProgramExecutor::execute() {
+  ZoneScopedN("program_execute");
+  ZoneText(program->name()->c_str(), std::strlen(program->name()->c_str()));
   LOG_DEBUG(LogType::LogRuntimeTTNN,
             "Starting execution of program: ", program->name()->c_str());
   for (const ::tt::target::ttnn::Operation *op : *program->operations()) {
     LOG_DEBUG(LogType::LogRuntimeTTNN,
               "Executing operation: ", op->debug_info()->c_str());
+    // TODO(#7743): Remove these tracy messages.
+    // Currently they are being used by `ttrt perf` for parsing the csv output.
     perf::Env::get().tracyLogOpLocation(std::string(op->loc_info()->c_str()));
     perf::Env::get().tracyLogConstEvalProgram(constEvalProgram);
     perf::Env::get().tracyLogProgramMetadata(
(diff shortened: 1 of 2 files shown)


runtime/lib/ttnn/program_executor.cpp
```diff
@@ -157,15 +159,14 @@ void ProgramExecutor::runCallback(
 }
 
 void ProgramExecutor::execute() {
+  ZoneScopedN("program_execute");
+  const char *programName = program->name()->c_str();
```
vmilosevic: :warning: **clang\-analyzer\-deadcode\.DeadStores** :warning:
Value stored to `` programName `` during its initialization is never read

vmilosevic: :warning: **clang\-diagnostic\-unused\-variable** :warning:
unused variable `` programName ``

runtime/lib/ttnn/program_executor.cpp
```diff
@@ -182,6 +183,10 @@ std::vector<::tt::runtime::Tensor> ProgramExecutor::gatherOutputTensors() {
 }
 
 void ProgramExecutor::runOperation(const ::tt::target::ttnn::Operation *op) {
+  ZoneScoped;
+  const char *opName =
```
vmilosevic: :warning: **clang\-analyzer\-deadcode\.DeadStores** :warning:
Value stored to `` opName `` during its initialization is never read

vmilosevic: :warning: **clang\-diagnostic\-unused\-variable** :warning:
unused variable `` opName ``

runtime/lib/ttnn/program_executor.cpp
```diff
@@ -157,15 +159,14 @@ void ProgramExecutor::runCallback(
...
   LOG_DEBUG(LogType::LogRuntimeTTNN,
-            "Starting execution of program: ", program->name()->c_str());
+            "Starting execution of program: ", programName);
   for (const ::tt::target::ttnn::Operation *op : *program->operations()) {
     LOG_DEBUG(LogType::LogRuntimeTTNN,
               "Executing operation: ", op->debug_info()->c_str());
-    perf::Env::get().tracyLogOpLocation(std::string(op->loc_info()->c_str()));
-    perf::Env::get().tracyLogConstEvalProgram(constEvalProgram);
-    perf::Env::get().tracyLogProgramMetadata(
-        perf::Env::get().tracyProgramMetadata);
```
jameszianxuTT, line 168: nit: I guess if we're moving over to metal tracy module, which doesn't support extracting out these location info/consteval logs to the device perf timing. Still, since ttrt perf deprecation is somewhat [blocked](https://github.com/tenstorrent/tt-mlir/pull/7653#pullrequestreview-4021878747), maybe some notice that [this feature is now broken](https://github.com/tenstorrent/tt-mlir/blob/10558fabdb90c88d8264a522ff2fdc8897f28f8b/tools/ttrt/common/perf.py#L553-L696) would be helpful?

https://tenstorrent.slack.com/archives/C07A30HRX7C/p1774472100624459?thread_ts=1774465200.323719&cid=C07A30HRX7C

pilkicTT: yeah, you're totally right, i've missed that we're using this for parsing out the data... i'll see what i can do about it...

pilkicTT: i've reverted that part of the change - all of the messages are left intact, so we don't break `ttrt perf`. we'll remove it once we're ready for deprecation.

PR #7752, opened 2026-04-01
Add decomposition for DistributedRMSNormOp with non-(1,1,3…

sdjordjevicTT:
### Ticket
N/A

### Problem description
TTNN_DistributedRMSNormOp maps to the ttnn::fused_rms_minimal kernel, which only supports input shapes (1,1,32,M). For any other shape, the op fails at runtime.

### What's changed
Added a decomposition workaround that rewrites DistributedRMSNormOp into primitive TTNN ops when the input shape is not (1,1,32,M). The decomposition mimics tt-metal's rms_norm_pre_all_gather + all_gather + rms_norm_post_all_gather pattern:
- Pre all-gather: `multiply(x, x) → mean(x², dim=-1) to compute local E(x²)`
- All-gather: gathers local stats across devices along the cluster axis
- Post all-gather: `mean → add(eps) → rsqrt → multiply(x, inv_rms) → optional multiply(weight)`

Handles optional residual (added to input before stats computation) and optional weight (gamma). Shapes that are supported by the fused kernel (1,1,32,M) are left unchanged.

Once rms_norm_pre_all_gather and rms_norm_post_all_gather are introduced in the compiler, we should rewrite this decomposition with these added ops.

### Checklist

### DIFF
--- a/lib/Dialect/TTNN/Transforms/CMakeLists.txt
+++ b/lib/Dialect/TTNN/Transforms/CMakeLists.txt
@@ -59,7 +59,6 @@ add_mlir_dialect_library(MLIRTTNNTransforms
         Workarounds/Decomposition/ScaledDotProductAttentionPadTileDimsRewritePattern.cpp
         Workarounds/Decomposition/UpsampleOpRewritePattern.cpp
         Workarounds/Decomposition/RMSNormConfigRewritePattern.cpp
-
         Workarounds/Decomposition/DistributedRMSNormWidthShardInputRewritePattern.cpp
         Workarounds/Decomposition/ReduceScatterConfigRewritePattern.cpp
         Workarounds/Decomposition/RotaryEmbeddingOpRewritePattern.cpp
--- a/lib/Dialect/TTNN/Transforms/Decomposition/DistributedRMSNormDecompositionRewritePattern.cpp
+++ b/lib/Dialect/TTNN/Transforms/Decomposition/DistributedRMSNormDecompositionRewritePattern.cpp
@@ -10,13 +10,18 @@
 
 namespace mlir::tt::ttnn::decomposition {
 
-// Returns true if the input shape is (1,1,32,M) where M is a multiple of 32,
-// which is the only shape supported by the fused_rms_minimal kernel.
-// Mirrors the TT_FATAL assertion in rms_allgather_device_operation.cpp:
-//   shape[0]==1, shape[1]==1, shape[2]==32, shape[3]%32==0
-static bool isSupportedByFusedKernel(ArrayRef<int64_t> shape) {
+// Returns true if the op can be lowered to the fused_rms_minimal kernel.
+// The kernel requires:
+//   - input shape (1,1,32,M) where M is a multiple of 32
+//     (rms_allgather_device_operation.cpp: shape[0]==1, shape[1]==1,
+//      shape[2]==32, shape[3]%32==0)
+//   - a weight (gamma) tensor must be present; the kernel asserts
+//     gamma.has_value() (https://github.com/tenstorrent/tt-metal/issues/38211)
+static bool isSupportedByFusedKernel(ttnn::DistributedRMSNormOp op) {
+  ArrayRef<int64_t> shape =
+      mlir::cast<RankedTensorType>(op.getInput().getType()).getShape();
   return shape.size() == 4 && shape[0] == 1 && shape[1] == 1 &&
-         shape[2] == 32 && shape[3] % 32 == 0;
+         shape[2] == 32 && shape[3] % 32 == 0 && op.getWeight();
 }
 
 LogicalResult DistributedRMSNormDecompositionRewritePattern::matchAndRewrite(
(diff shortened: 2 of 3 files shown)


lib/Dialect/TTNN/Transforms/Workarounds/Decomposition/DistributedRMSNormDecompositionRewritePattern.cpp
```diff
@@ -0,0 +1,138 @@
...
+#include "ttmlir/Dialect/TTNN/Utils/TransformUtils.h"
+#include "ttmlir/Dialect/TTNN/Utils/Utils.h"
+#include "ttmlir/Utils.h"
+
+namespace mlir::tt::ttnn::workarounds::decomposition {
+
+// Returns true if the input shape is (1,1,32,M) which is the only shape
+// supported by the fused_rms_minimal kernel.
+static bool isSupportedByFusedKernel(ArrayRef<int64_t> shape) {
+  return shape.size() == 4 && shape[0] == 1 && shape[1] == 1 && shape[2] == 32;
```
odjuricicTT: Are you sure that this is the only limitation? This shape will be hit for all llms for rms norm.

Be sure to run the llmbox and galaxy perf benchmarks with this once the composite op lands as well.

test/python/golden/ttir_ops/normalization/test_normalization.py
```diff
@@ -313,6 +313,7 @@ def layer_norm(
         (1, 1, 32, 512),
         (1, 1, 32, 4096),
         (1, 1, 32, 8192),
+        (1, 1, 128, 128),
```
svuckovicTT, line 316: Does `M` in `(1,1,32,M)` represent any positive integer value? If so, add test for that as well.

svuckovicTT (approved):

lib/Dialect/TTNN/Transforms/Workarounds/Decomposition/DistributedRMSNormDecompositionRewritePattern.cpp
sdjordjevicTT: Checked this, we have a bunch of workarounds to make this op work. I am pretty sure it will work, but yeah, I will run benchmark tests to verify.

sdjordjevicTT: Btw, now thinking, we will run those tests once we add support for composite distributed rms norm...

lib/Dialect/TTNN/Transforms/Decomposition/DistributedRMSNormDecompositionRewritePattern.cpp
```diff
@@ -0,0 +1,138 @@
...
+  // local_stats = mean(x_sq, dim=-1, keep_dim=true)
+  // Output shape: input shape with last dim replaced by 1. The mean op
+  // handles the shape reduction internally.
+  SmallVector<int64_t> statsShape(inputShape.begin(), inputShape.end());
+  statsShape.back() = 1;
+  auto inputEncoding =
+      mlir::cast<ttnn::TTNNLayoutAttr>(inputType.getEncoding());
+  RankedTensorType statsType =
+      RankedTensorType::get(statsShape, inputType.getElementType(),
+                            inputEncoding.withTensorShape(statsShape));
```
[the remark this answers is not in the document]
sdjordjevicTT: This is not expected.

lib/Dialect/TTNN/Transforms/CMakeLists.txt
```diff
@@ -57,6 +59,7 @@ add_mlir_dialect_library(MLIRTTNNTransforms
         Workarounds/Decomposition/ScaledDotProductAttentionPadTileDimsRewritePattern.cpp
         Workarounds/Decomposition/UpsampleOpRewritePattern.cpp
         Workarounds/Decomposition/RMSNormConfigRewritePattern.cpp
+
```
jserbedzijaTT: Is this newline needed

lib/Dialect/TTNN/Transforms/Decomposition/DistributedRMSNormDecompositionRewritePattern.cpp
```diff
@@ -0,0 +1,149 @@
...
+        ttmlir::utils::appendLocationSuffix(loc, "_residual_add"), inputType, x,
+        op.getResidual());
+    x = addOp.getResult();
+  }
+
+  // --- Pre all-gather: compute local E(x^2) ---
+
+  // x_sq = multiply(x, x)
+  auto xSqOp = rewriter.create<ttnn::MultiplyOp>(
+      ttmlir::utils::appendLocationSuffix(loc, "_square"), inputType, x, x);
```
jserbedzijaTT, line 74: Maybe we can use `PowScalar` operation here?

sdjordjevicTT: I have more trust in Multiply. :D

lib/Dialect/TTNN/Transforms/CMakeLists.txt
sdjordjevicTT: Good catch, thanks!

PR #7753, opened 2026-04-01
[d2m] ttir.sum outer reduction support

dlokeTT:
### Problem description
Reductions on outer dim aren't supported in d2m. Found an example of a sum reduction like this in gpt oss

### What's changed

- Added lowering for outer-dimension reductions with a new rewriter that zero fills the output which is accumulated into.
- Refactored some of the reduction checks and helpers so they can be used in both rewriters
- Extended `d2m.remote_load` with `isNotConflicting` so reads inside a generic from output backed memrefs don’t conflict with the generic’s writes.
- `InsertDstRegisterAccess` expects the reduction dimension to be fully collapsed after reblocking and the auto policy can sometimes do partial blocking which corrupts the accumulation. Xfailed reduction tests that hit this problem and cause bad pcc.
- Added a check in `ConvertLocalLoadStoreOpsToAliasedCBs` where if a local buffer is a `WaitOp` on the same CB as the store target, drop the redundant `remote_store` to avoid a double push which caused race conditions between DM and compute leading to bad pcc
- Added golden metal reduction tests for outer dim sum reductions and updated allocator lit test

### Checklist

### DIFF
--- a/test/ttmlir/Dialect/D2M/allocate/allocate_reblock_auto.mlir
+++ b/test/ttmlir/Dialect/D2M/allocate/allocate_reblock_auto.mlir
@@ -45,8 +45,7 @@ module {
   // CHECK-MAX-LABEL: func.func @single_reduction_non_matmul()
   // CHECK-MAX: d2m.generic {block_factors = [1, 1], grid = #ttcore.grid<1x1>
   // CHECK-AUTO-LABEL: func.func @single_reduction_non_matmul()
-  // CHECK-AUTO: d2m.view_layout {{.*}} -> memref<1x8x4x1x!ttcore.tile<32x32, f32>, #ttcore.shard<4096x4096, 1>, #l1>
-  // CHECK-AUTO: d2m.generic {block_factors = [1, 8], grid = #ttcore.grid<1x1>
+  // CHECK-AUTO: d2m.generic {block_factors = [1, 1], grid = #ttcore.grid<1x1>
   func.func @single_reduction_non_matmul() -> memref<1x1x4x1x!ttcore.tile<32x32, f32>, #ttcore.shard<16384x4096, 1>, #l1> {
     %in = memref.alloc() : memref<1x1x4x8x!ttcore.tile<32x32, f32>, #ttcore.shard<16384x32768, 1>, #l1>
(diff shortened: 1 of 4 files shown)


test/ttmlir/Dialect/D2M/allocate/allocate_reblock_auto.mlir
```diff
@@ -58,13 +59,13 @@ module {
...
-  // CHECK-AUTO: d2m.view_layout {{.*}} -> memref<1x2x1x4x!ttcore.tile<32x32, f32>, #ttcore.shard<16384x4096, 1>, #l1>
-  // CHECK-AUTO: d2m.view_layout {{.*}} -> memref<2x1x4x1x!ttcore.tile<32x32, f32>, #ttcore.shard<4096x4096, 1>, #l1>
-  // CHECK-AUTO: d2m.generic {block_factors = [1, 1, 2], grid = #ttcore.grid<1x1>
-  // CHECK-AUTO: memref.alloc() {{.*}} : memref<1x4x!ttcore.tile<32x32, f32>, #ttcore.cb_layout<16384x4096, 2>, #l1>
-  // CHECK-AUTO: memref.alloc() {{.*}} : memref<4x1x!ttcore.tile<32x32, f32>, #ttcore.cb_layout<4096x4096, 2>, #l1>
-  func.func @auto_clamps_reduction_factor_at_four_tiles() -> memref<1x1x1x1x!ttcore.tile<32x32, f32>, #ttcore.shard<4096x4096, 1>, #l1> {
+  // CHECK-AUTO-LABEL: func.func @auto_matmul_k8_full_reduction_tile()
+  // CHECK-AUTO: d2m.view_layout {{.*}} -> memref<1x8x1x1x!ttcore.tile<32x32, f32>, #ttcore.shard<4096x4096, 1>, #l1>
+  // CHECK-AUTO: d2m.view_layout {{.*}} -> memref<8x1x1x1x!ttcore.tile<32x32, f32>, #ttcore.shard<4096x4096, 1>, #l1>
+  // CHECK-AUTO: d2m.generic {block_factors = [1, 1, 8], grid = #ttcore.grid<1x1>
```
anuragsingh-tt: This will likely reduce perf (PCC + exec time).

dlokeTT:
Talked offline with @anuragsingh-tt and will land this without allocator changes. Going to xfail the tests that are failing and create an issue for it

Edit: Issue here https://github.com/tenstorrent/tt-mlir/issues/7895

PR #7755, opened 2026-04-01
[wip] Worker owned host tensor caching

jameszianxuTT:
### Ticket
https://github.com/tenstorrent/tt-xla/issues/4011

### Problem description
CreateOwnedHostTensor is broadcast for every shard and replica to every worker, leading to host memory usage that scales with the device count in the entire cluster leading to host OOM on trivial sized models over a large enough cluster (eg. quad galaxy)

### What's changed
- Deduplicate worker tensor shards using a tensor cache, where cache entry lifetimes are tied to the lifetimes of the runtime tensors they back. 
- Add CreateBorrowedHostTensor API that accepts a shared_ptr rather than raw data ptr so this change may be scoped only to multihost worker

### Checklist

### DIFF
--- a/runtime/lib/distributed/worker/tensor_data_cache.cpp
+++ b/runtime/lib/distributed/worker/tensor_data_cache.cpp
@@ -3,6 +3,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 #include "tt/runtime/detail/distributed/worker/tensor_data_cache.h"
+#include "tt/runtime/detail/common/logger.h"
 #include <cstdint>
 #include <numeric>
 #include <tt_stl/reflection.hpp>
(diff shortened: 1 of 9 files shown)


runtime/lib/distributed/worker/tensor_data_cache.cpp
```diff
@@ -0,0 +1,55 @@
...
+
+uint64_t TensorDataCache::getTensorHash(const uint8_t *tensorDataPtr,
+                                        const std::vector<uint32_t> &shape,
+                                        const std::vector<uint32_t> &stride,
+                                        uint32_t itemSize,
+                                        ::tt::target::DataType dataType) {
+  return ttsl::hash::hash_objects_with_default_seed(
+      reinterpret_cast<uintptr_t>(tensorDataPtr), shape, stride, itemSize,
+      static_cast<uint16_t>(dataType));
+}
```
jameszianxuTT, line 53: To be safe, we can do a more through hash over part or all of the data or tombstone torchxla shard locations (leading to a fragmented and larger vaddr space, but still allows much lower actual RSS)

PR #7776, opened 2026-04-02
[TTIRBuilder] Add deferred device support and refactor SDPA fusing tests

jserbedzijaTT:
## Summary
- **Deferred device support in builder**: Add `DeferredDevice` class and `isinstance`-based check in `_compile_and_execute` to defer device opening until after compilation. The optimizer pipeline (needed for SDPA fusing) uses OpModel's internal mock device during compilation via `SingletonDeviceContext`. If a real device is already open (from the pytest `device` fixture), the mock device creation fails with "Cannot switch to real hardware while 1 device(s) are active." `DeferredDevice(request)` wraps the pytest request and opens the real device only after compilation completes, with automatic cleanup (close device + clear fixture cache) after execution. Tests simply pass `device=DeferredDevice(request)` instead of the normal `device` fixture.
- **Return `artifact_dir`** from `compile_and_execute_ttir` so tests can inspect compiled MLIR for fusing pattern verification.
- **Refactor `test_sdpa_fusing.py`** with composable building blocks (`gqa_broadcast_kv`, `pre_scale`, `post_scale_scores`, `compute_scores`, `simple_softmax`, `nan_safe_softmax`) and `SDPAShapes` dataclass. Shape configs are derived from real model IRs (LLaMA, Qwen, Phi, Falcon, Gemma, Mistral, BERT, ViT, SegFormer, Swin).
- **5 focused test functions** covering real model SDPA patterns: `split_scale_nan_safe` (Qwen/Phi/BERT/ViT), `post_scale_simple` (SegFormer), `pre_scale_q_nan_safe` (LLaMA decode), `pre_scale_k_nan_safe` (Falcon/Gemma/Mistral), `simple_softmax` (Swin). 7 representative shape configs covering MHA, MQA, GQA (4:1, 8:1), decode, and prefill modes.

## Test plan
- [x] Run `pytest test/python/golden/ttir_ops/fusing/test_sdpa_fusing.py` — all 42 SDPA fusing tests pass
- [ ] Run `pytest test/python/golden/ttir_ops/fusing/` — no regressions in other fusing tests
- [ ] Verify SDPA ops appear in compiled MLIR output (fusing pattern matched)

### DIFF
--- a/test/python/golden/ttir_ops/fusing/test_sdpa_fusing.py
+++ b/test/python/golden/ttir_ops/fusing/test_sdpa_fusing.py
@@ -15,7 +15,14 @@ from builder.ttir.ttir_builder import TTIRBuilder
 from builder.base.builder_apis import compile_and_execute_ttir
 from conftest import get_request_kwargs, DeferredDevice
 
-pytestmark = pytest.mark.frontend("ttir")
+pytestmark = [
+    pytest.mark.frontend("ttir"),
+    pytest.mark.skip_config(
+        ("p150",),
+        ("p300",),
+        reason="Optimizer mock device grid mismatch on Blackhole (https://github.com/tenstorrent/tt-mlir/issues/7809)",
+    ),
+]
 
 
 # ---------------------------------------------------------------------------


tools/builder/base/builder_apis.py
```diff
@@ -46,6 +46,44 @@
 # ----- Private APIs -----
 
 
+class DeferredDevice:
```
jgrimTT: Why'd you decide to put DeferredDevice in builder_apis instead of conftest?

jgrimTT (approved):

tools/builder/base/builder_apis.py
jserbedzijaTT: My bad, moved it to conftest

Issue #7360, opened 2026-03-05
[TTIR→TTNN] Floating-point exception when lowering ops with zero-size tensor dimensions in pocket_tts model


ashokkumarkannan1:
### Describe the bug

- The TTIR-to-TTNN lowering pass crashes with a fatal floating-point exception when processing graphs that contain tensors with a zero-size dimension (e.g., tensor<1x0x32xf32>).
- This was discovered while running the `pocket_tts` model. The `FlowLMModel.forward()` method passes a `[1, 0, 32] float32 tensor` through `torch.where `and `nn.Linear.`
- PyTorch handles zero-size tensors correctly, but when this graph is lowered from TTIR to TTNN, the compiler hits a floating-point exception
- The TTIR module is generated successfully and printed in full. The crash occurs immediately after, before any TTNN output is produced.

- TTIR graph that triggers the crash:
```
module @SyncTensorsGraph.12 attributes {mhlo.cross_program_prefetches = [], mhlo.input_output_alias = [], mhlo.is_dynamic = false, mhlo.use_auto_spmd_partitioning = false, ttcore.meshes = #ttcore.meshes<[<"mesh" = 1x1>]>} {
  ttcore.device_module {
    builtin.module @SyncTensorsGraph.12 attributes {mhlo.cross_program_prefetches = [], mhlo.input_output_alias = [], mhlo.is_dynamic = false, mhlo.use_auto_spmd_partitioning = false, ttcore.meshes = #ttcore.meshes<[<"mesh" = 1x1>]>} {
      func.func @main(%arg0: tensor<1x0x32xf32> {ttcore.argument_type = #ttcore.argument_type<input>, ttcore.runtime_tensor_sharding = #ttcore<runtime_tensor_sharding shard_status = <unsharded>, local_shape = tensor<1x0x32xf32>>} loc("p0.1"), %arg1: tensor<32xf32> {ttcore.argument_type = #ttcore.argument_type<input>, ttcore.runtime_tensor_sharding = #ttcore<runtime_tensor_sharding shard_status = <unsharded>, local_shape = tensor<32xf32>>} loc("p1.3")) -> (tensor<1x0x32xf32> {ttcore.runtime_tensor_sharding = #ttcore<runtime_tensor_sharding shard_status = <unsharded>, local_shape = tensor<1x0x32xf32>>}) {
[...]

Issue #5671, opened 2025-11-03
Enable enableAggressiveSimplification in StableHLOToTTIR pipeline


sdjordjevicTT:
@anuragsingh-tt already tried to enable this pipeline option with this PR:
https://github.com/tenstorrent/tt-mlir/commit/6e6cb79ffcf157c71e4fde2f144983487727221f

Let's revisit the problems right now and see what we can do to enable this pipeline option.

PR #7777, opened 2026-04-02
 Enable aggressive simplification by default in StableHLO-to-TTIR pipeline  

ajakovljevicTT:
We had an enable-aggressive-simplification flag in the StableHLO-to-TTIR pipeline that was off by default. Turning it on lets StableHLO do useful cleanup before we start lowering. However, aggressive simplification keeps constants as rank-0 scalars, and when our TTIR conversion later reshapes those to match expected ranks, several fusing patterns stopped recognizing them. This PR fixes those patterns to look through the reshape chains so fusing keeps working, and updates the affected tests accordingly.

Fixes https://github.com/tenstorrent/tt-mlir/issues/5671
Fixes https://github.com/tenstorrent/tt-mlir/issues/7360

### DIFF
--- a/lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
+++ b/lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
@@ -2497,7 +2497,8 @@ private:
   // it exists, given the result of the sequence.
   Value getXCubedInput(Value xCubedResult) const {
     if (PowOp xCubed = xCubedResult.getDefiningOp<ttir::PowOp>()) {
-      ttir::FullOp power = getFullOpThroughTMChain(xCubed.getRhs());
+      ttir::FullOp power = dyn_cast_or_null<ttir::FullOp>(
+          getScalarThroughTMChain(xCubed.getRhs()));
       if (!power) {
         return nullptr;
       }


sdjordjevicTT:
Just to clarify, we have some stablehlo -> ttir conversions that convert scalar(0D) tensors to 1D and reshape back? We have full support in the compiler for scalar types, hence I just want to make sure that we are doing the right thing here.

ajakovljevicTT:
> Just to clarify, we have some stablehlo -> ttir conversions that convert scalar(0D) tensors to 1D and reshape back? We have full support in the compiler for scalar types, hence I just want to make sure that we are doing the right thing here.

@sdjordjevicTT, thanks for bringing this up, Some of these changes are leftover from my testing, so they might not be fully needed

acicovicTT (approved):

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
```diff
@@ -370,6 +370,20 @@ static bool isFullOpWithValue(Value value, float expectedValue,
...
+  while (isa<ReshapeOp>(op)) {
+    auto input = op->getOperand(0);
+    if (mlir::cast<RankedTensorType>(input.getType()).getRank() != 0) {
+      break;
+    }
+    op = input.getDefiningOp();
+    if (!op) {
+      return false;
+    }
+  }
```
azecevic000, line 385: FYI @dmilinkovicTT @amalkovTT I think cases like this one (splat values) are major use cases in practice, so a sound strategy for constant folding might be covering all TMs for splat values. This is trivial for splat values, and brings the most value for other passes.

azecevicTT (approved):

ajakovljevicTT:
> > Just to clarify, we have some stablehlo -> ttir conversions that convert scalar(0D) tensors to 1D and reshape back? We have full support in the compiler for scalar types, hence I just want to make sure that we are doing the right thing here.
> 
> @sdjordjevicTT, thanks for bringing this up, Some of these changes are leftover from my testing, so they might not be fully needed

@sdjordjevicTT I've added specific tests for this. The issue is that for stablehlo -> TTIR `broadcast_in_dim` on, if the input output are different in rank, we have `reshape -> broadcast` pattern. Later on, the fusing passes look for Full or Zero op as scalars, and fail if they hit a reshape, even though it comes from a scalar. I've hit this issue when testing aggressive simplification, but it is not related to the simplification directly.

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
```diff
@@ -2474,7 +2497,7 @@ class GeluFusionPattern : public mlir::OpRewritePattern<ttir::MultiplyOp> {
   // it exists, given the result of the sequence.
   Value getXCubedInput(Value xCubedResult) const {
     if (PowOp xCubed = xCubedResult.getDefiningOp<ttir::PowOp>()) {
-      ttir::FullOp power = xCubed.getRhs().getDefiningOp<ttir::FullOp>();
+      ttir::FullOp power = getFullOpThroughTMChain(xCubed.getRhs());
```
vmilosevic: :warning: **clang\-diagnostic\-error** :warning:
no viable conversion from `` Operation * `` to `` ttir::FullOp ``

vmilosevic: :warning: **clang\-diagnostic\-error** :warning:
use of undeclared identifier `` getFullOpThroughTMChain ``; did you mean `` getScalarThroughTMChain ``?
```suggestion
      ttir::FullOp power = getScalarThroughTMChain(xCubed.getRhs());
```

PR #7783, opened 2026-04-02
Multihost QOL Fixes

### LINKED ISSUE #7046: Defensive Multihost Improvements
1. Add T/O for tt::runtime::ControllerSocket::connectToWorkers, since this can hang for many reasons.
2. Add check for extant worker proc's on children
3. Logs like `Connecting to runtime controller UF-EV-B8-GWH01:45271` should also include the child's host IP to reduce confusion
4. Better configurability for remote_docker.sh
5. Compatibility with --tcp-interface option for ttrun; we may need this in the future as alternative to letting tcp_btl_if_exclude and letting MPI guess
6. Better distributed reset
7. It seems unnecessary in tt-xla to introduce the extra indirection inside compiled C++ between named cluster topology "dual_bh_quietbox" etc. and the path to the rank bindings in tt-metal. This also makes it harder to test since repointing to a different requires a recompile. We should embed the path directly relative to tt metal home, and if really necessary introduce the indirection in python test infra.

jameszianxuTT:
### Ticket
Partially resolves #7046 

### Problem description
- Uncaught segfault in controller due to nullptr deref when workers disconnect - handle this explictly and sigabrt rather than let a sigsegv propagate up to the top of the process. 

### What's changed
- Handle null response with assert

### Checklist

### DIFF
--- a/runtime/lib/common/types.cpp
+++ b/runtime/lib/common/types.cpp
@@ -142,16 +142,22 @@ std::string MultiProcessArgs::toArgString() const {
     }
   }
 
-  // Hosts file
+  // Hosts file (default to /etc/mpirun/hostfile when no hosts or hostfile set)
   if (!hostsFilePath_.empty()) {
     oss << " ";
     oss << "--hostfile " << hostsFilePath_;
+  } else if (hosts_.empty()) {
+    oss << " ";
+    oss << "--hostfile /etc/mpirun/hostfile";
   }
 
-  // Rank file
+  // Rank file (default to /etc/mpirun/rankfile)
   if (!rankFilePath_.empty()) {
     oss << " ";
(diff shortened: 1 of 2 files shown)


runtime/lib/common/types.cpp
```diff
@@ -142,16 +142,22 @@ std::string MultiProcessArgs::toArgString() const {
...
   }
 
-  // Rank file
+  // Rank file (default to /etc/mpirun/rankfile)
   if (!rankFilePath_.empty()) {
     oss << " ";
     oss << "--rankfile " << rankFilePath_;
+  } else {
+    oss << " ";
+    oss << "--rankfile /etc/mpirun/rankfile";
```
jameszianxuTT, line 160: bad idea - this may not be necessary, and this backup file may not exist

PR #7786, opened 2026-04-02
Replace custom CCL canonicalization with SDY built-in patterns

sshonTT:
### Ticket
N/A

### Problem description
we currently use our own canonicalization. However, to take full advantage of community features, we want to adopt the standard version rather than defining our own.

### What's changed
Remove the custom AllReduceAllSliceToReduceScatterPattern and delegate to SDY's built-in canonicalization patterns for all collective ops (AllGather,AllSlice, AllReduce, AllToAll, ReduceScatter).

### Checklist

### DIFF
--- a/test/ttmlir/Dialect/StableHLO/shardy/shardy_ccl_canonicalization.mlir
+++ b/test/ttmlir/Dialect/StableHLO/shardy/shardy_ccl_canonicalization.mlir
@@ -1,0 +1,44 @@
+// REQUIRES: stablehlo
+// RUN: ttmlir-opt --shardy-ccl-canonicalization %s | FileCheck %s
+
+sdy.mesh @mesh = <["_axis_0"=2, "_axis_1"=4]>
+
+// all_reduce + multi-dim all_slice should be fused into
+// reduce_scatter (on the matching axis) + residual all_slice (on the rest).
+// CHECK-LABEL: func.func @reduce_scatter_fusion_multi_dim
+// CHECK-NOT: sdy.all_reduce
+// CHECK: sdy.reduce_scatter [{}, {}, {}, {"_axis_1"}]
+// CHECK: sdy.all_slice [{}, {}, {"_axis_0"}, {}]
+func.func @reduce_scatter_fusion_multi_dim(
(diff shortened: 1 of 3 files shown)


test/ttmlir/Dialect/StableHLO/shardy/shardy_ccl_canonicalization.mlir
```diff
@@ -0,0 +1,43 @@
...
+  %0 = sdy.all_reduce {"_axis_1"} %arg0 out_sharding=<@mesh, [{}, {}, {}, {}]> : tensor<4x1x32x2880xbf16>
+  %1 = sdy.all_slice [{}, {}, {}, {"_axis_1"}] %0 out_sharding=<@mesh, [{}, {}, {}, {"_axis_1"}]> : tensor<4x1x32x2880xbf16>
+  return %1 : tensor<4x1x32x2880xbf16>
+}
+
+// all_slice(all_gather(x)) with the same axes should cancel out to identity.
+// The dead all_gather remains (no Pure trait) but the all_slice is eliminated.
+// CHECK-LABEL: func.func @all_slice_cancels_all_gather
+// CHECK-NOT: sdy.all_slice
+// CHECK: return %arg0
```
hshahTT, line 38: Since you expect the dead all_gather to remain you should add a CHECK statement for that. So something like:
```
// CHECK-LABEL: func.func @all_slice_cancels_all_gather
// CHECK: sdy.all_gather
// CHECK-NOT: sdy.all_slice
// CHECK: return %arg0
```

As an aside, why does the all_gather remain? And what does "no Pure trait" mean here?

sshonTT: Thanks for the suggestion, I will add the CHECK for all_gather.

As for the "no Pure trait", operations without the Pure trait aren't automatically removed by DCE since they might have side effects. Once all_slice is eliminated, all_gather's output is no longer used, but the operation itself stays in the IR because it doesn't have the Pure trait.

If we later want the all_gather removed too, we’d need either a dedicated pattern that erases all_gather when its result is unused

hshahTT: Gotcha, thanks for the explanation. That seems like a separate issue so will approve the PR now

PR #7797, opened 2026-04-03
[EmitPy] Imrpove CPU-hoisted function placement and EmitPy imports

dmilinkovicTT:
### Ticket
N/A

### Problem description

Several issues with how CPU-hoisted functions were handled in the EmitPy pipeline:

- **CPU-hoisted declarations were placed in the wrong file.** `TTNNFileSplit` was moving CPU-hoisted declarations to the "consteval" file unconditionally, even when they were called from a different file (e.g., "main"). Declarations should live in the file that actually uses them.
- **`EmitPyLinkModules` was doing too much.** It was handling file-level concerns (finding consteval/main files, creating inter-file imports, choosing where to place CPU definitions) when its responsibility should be limited to CPU → Device module linking.
- **Import creation was scattered.** Module-level imports (`import ttnn`, `import utils`) lived in `ConvertTTNNToEmitPy`, inter-file imports were created in `EmitPyLinkModules`, and CPU imports were handled during linking — there was no single place responsible for imports.
- **Function types were implicit.** Consteval wrappers were identified by an ad-hoc `consteval_wrapper` discardable attribute, and inter-file declarations had no explicit type at all — both should use the established `FunctionType` enum for consistency and clarity.

#### 1. Refactor EmitPyLinkModules

Strips `EmitPyLinkModules` down to its core responsibility: linking CPU and Device modules. The pass now replaces CPU-hoisted declarations in the Device module with their corresponding definitions from the CPU module, moves CPU imports to the appropriate scope, and erases the CPU module. It no longer searches for main/consteval files, creates inter-file import statements, or decides where CPU definitions should be placed — all of that is now handled upstream or downstream.

#### 2. Move CPU-hoisted declarations to an appropriate file

Updates `TTNNFileSplit` to place each CPU-hoisted declaration in the file that actually contains its call site, rather than unconditionally cloning it into the consteval file. The pass walks the main and consteval files to find which one references each declaration and moves it there.

#### 3. Introduce ImportedDeclaration function type

Adds a new `FunctionType::ImportedDeclaration` to explicitly represent private declarations that refer to functions defined in a different file (created by `TTNNFileSplit`). These carry a `tt.imported_from` attribute indicating the source file.

#### 4. Introduce explicit function type for ConstEval wrappers

Replaces the ad-hoc `consteval_wrapper` discardable attribute with a proper `FunctionType::ConstEvalWrapper` enum value. All code that previously checked `funcOp->hasAttr(kWrapperAttr)` now uses `ttmlir::utils::isConstEvalWrapperFunc(funcOp)`. The old `kWrapperAttr` constant is removed.

#### 5. Adjust lit tests

Updates lit tests to reflect the new behavior: CPU-hoisted function definitions now appear right before their first caller (rather than grouped at the top of the consteval file), and inter-file imports no longer include CPU-hoisted functions that are defined locally.

#### 6. Introduce EmitPyAddImports pass

Extracts all import-creation logic into a dedicated `EmitPyAddImports` pass that runs after `EmitPyLinkModules`. This pass handles:
- Module-level imports (`import ttnn`, `import utils`, `import ttir_cpu`) — added per-file when file splitting is active, or at module scope otherwise. `import ttir_cpu` is only added to scopes that contain CPU-hoisted code.
- Imported-declaration imports (`from <file> import <func>`) — generated from `ImportedDeclaration` func ops, grouped by source file.

This replaces the import logic that was previously spread across `ConvertTTNNToEmitPy` (module-level imports) and `EmitPyLinkModules` (inter-file + CPU imports).

### Validations

On PR: https://github.com/tenstorrent/tt-mlir/actions/runs/23940436187

### Checklist

### DIFF
--- a/lib/Conversion/TTNNToEmitPy/EmitPyLinkModules.cpp
+++ b/lib/Conversion/TTNNToEmitPy/EmitPyLinkModules.cpp
@@ -3,18 +3,15 @@
 // SPDX-License-Identifier: Apache-2.0
 
 #include "ttmlir/Conversion/TTNNToEmitPy/TTNNToEmitPy.h"
-#include "ttmlir/Dialect/EmitPy/IR/EmitPyOps.h"
+#include "ttmlir/Dialect/EmitPy/IR/EmitPy.h"
 #include "ttmlir/Dialect/TTCore/IR/TTCoreOps.h"
 #include "ttmlir/FunctionTypes.h"
 
 #include "mlir/Dialect/Func/IR/FuncOps.h"
-#include "mlir/IR/Builders.h"
 #include "mlir/IR/BuiltinOps.h"
 #include "mlir/Pass/Pass.h"
-#include "llvm/ADT/SmallPtrSet.h"
 #include "llvm/ADT/SmallVector.h"
 #include "llvm/ADT/StringMap.h"
-#include "llvm/ADT/StringSet.h"
 
 namespace mlir::tt {
 
@@ -27,10 +24,6 @@ namespace {
 //
 // - Replacing the CPU-hoisted declarations in the Device module with the
 //   corresponding CPU-hoisted definitions from the CPU module.
-//
-// - Moving the imports from the CPU module to the corresponding scope in the
-//   Device module, which could be either the module itself, or a file inside
-//   the Device module (if TTNNFileSplit is performed);
 //
 // - Erasing the CPU module.
 //
@@ -109,8 +102,7 @@ public:
 
 private:
   // Link CPU module into the device module by replacing CPU-hoisted
-  // declarations with definitions and moving CPU imports to the enclosing
-  // scope of each declaration.
+  // declarations with definitions.
   //
   LogicalResult linkCPUModule(mlir::ModuleOp deviceModule,
                               mlir::ModuleOp cpuModule) {
@@ -123,17 +115,11 @@ private:
       }
     }
 
-    // Collect CPU module imports.
-    //
-    llvm::SmallVector<emitpy::ImportOp> cpuImports(
-        cpuModule.getOps<emitpy::ImportOp>());
-
     // Walk all CPU-hoisted declarations in the device module and replace
     // each with the corresponding definition from the CPU module.
     //
     // The definition is inserted before the first operation in the scope
     // that calls the symbol, so it appears right before its first usage.
-    // CPU imports are cloned into the scope once per scope.
     //
     llvm::SmallVector<func::FuncOp> decls;
     deviceModule.walk([&](func::FuncOp funcOp) {
@@ -142,7 +128,6 @@ private:
       }
     });
 
-    llvm::SmallPtrSet<Block *, 4> blocksWithCPUImportsCloned;
     for (auto decl : decls) {
       Block *scope = decl->getBlock();
 
@@ -163,12 +148,6 @@ private:
       }
       it->second->moveBefore(insertionPoint);
       decl->erase();
-
-      // Clone CPU imports into this scope (once per scope).
-      //
-      if (blocksWithCPUImportsCloned.insert(scope).second) {
-        cloneCPUImports(cpuImports, *scope);
-      }
     }
 
     return success();
@@ -192,28 +171,6 @@ private:
       }
     }
     return nullptr;
-  }
-
-  // Clone CPU module imports into the given scope, skipping duplicates.
-  //
-  void cloneCPUImports(ArrayRef<emitpy::ImportOp> cpuImports, Block &scope) {
-    llvm::StringSet<> existingImports;
-    for (auto &op : scope) {
-      if (auto importOp = dyn_cast<emitpy::ImportOp>(op)) {
-        existingImports.insert(importOp.getModuleName());
-      }
-    }
-
-    OpBuilder builder(&getContext());
-    builder.setInsertionPointToStart(&scope);
-
-    for (auto importOp : cpuImports) {
-      if (existingImports.contains(importOp.getModuleName())) {
-        continue;
-      }
-      builder.clone(*importOp);
-      existingImports.insert(importOp.getModuleName());
-    }
   }
 };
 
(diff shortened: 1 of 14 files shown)


include/ttmlir/Dialect/TTNN/Transforms/Passes.h
```diff
@@ -21,7 +21,6 @@
 namespace mlir::tt::ttnn {
 
 constexpr const char *kCacheDictAttr = "cache_dict";
-constexpr const char *kWrapperAttr = "consteval_wrapper";
```
amilovanovicTT, line 24: Remove this from `include/ttmlir/Conversion/TTNNToEmitPy/EmitPyConversion.h` as well

lib/Conversion/TTNNToEmitPy/EmitPyLinkModules.cpp
```diff
@@ -91,144 +66,154 @@ class EmitPyLinkModulesPass
...
-        op->moveBefore(blockToMoveCPUOpsTo, insertPoint);
+
+      // Find the first operation in the scope that calls this symbol
+      // and insert the definition before it.
+      //
+      Operation *insertionPoint = findFirstCaller(decl.getSymName(), *scope);
+      if (insertionPoint) {
+        it->second->moveBefore(insertionPoint);
+      } else {
+        it->second->moveBefore(decl);
```
amilovanovicTT: When can this be triggered? Maybe we should emit an error in that case

lib/Dialect/TTNN/Transforms/TTNNFileSplit.cpp
```diff
@@ -93,43 +93,78 @@ class TTNNFileSplit : public impl::TTNNFileSplitBase<TTNNFileSplit> {
...
+      bool found = false;
+      fileOp.walk([&](func::CallOp callOp) {
+        if (callOp.getCallee() == symbol) {
+          found = true;
+          return WalkResult::interrupt();
+        }
+        return WalkResult::advance();
+      });
+      return found;
+    };
```
amilovanovicTT, line 142: Nit:
```suggestion
    auto fileContainsCallTo = [](auto fileOp, StringRef symbol) {
      return fileOp
          .walk([&symbol](func::CallOp callOp) {
            return callOp.getCallee() == symbol ? WalkResult::interrupt()
                                                : WalkResult::advance();
          })
          .wasInterrupted();
```

lib/Conversion/TTNNToEmitPy/EmitPyLinkModules.cpp
```diff
@@ -91,144 +66,154 @@ class EmitPyLinkModulesPass
...
       }
+      decl->erase();
 
-      cpuModuleOp->erase();
+      // Clone CPU imports into this scope (once per scope).
+      //
+      if (blocksWithCPUImportsCloned.insert(scope).second) {
+        cloneCPUImports(cpuImports, *scope);
```
amilovanovicTT: There are some duplicate imports in the emitted code when the file split is performed (specifically `import ttnn`). 
Namely, `ConvertTTNNToEmitPy` creates `import ttnn` at the device module level (outside file ops). When the IR is translated to Python, module-level imports are emitted into every file. However, those device module level imports are not accounted for during this cloning. 
We should either fix this here or change the emission logic, whichever you prefer.

lib/Conversion/TTNNToEmitPy/TTNNToEmitPyPass.cpp
```diff
@@ -12,6 +12,8 @@
 #include "mlir/Dialect/Func/Transforms/FuncConversions.h"
 #include "mlir/Pass/PassManager.h"
```
amilovanovicTT, line 14: Nit:
```suggestion
```

lib/Conversion/TTNNToEmitPy/TTNNToEmitPyPass.cpp
```diff
@@ -121,9 +123,59 @@ struct ConvertTTNNToEmitPyPass
         return;
       }
     }
+
+    // Lower ImportedDeclaration func declarations to emitpy.import ops.
+    // These are private declarations created by TTNNFileSplit to represent
+    // functions defined in other files.
```
amilovanovicTT: Nit: maybe add sth like if performed, just to indicate that TTNNFilesSplit isn't necessarily performed

lib/Conversion/TTNNToEmitPy/EmitPyLinkModules.cpp
dmilinkovicTT: Makes sense, thanks! I've added deduplication logic in `TranslateToPython`, so that the rest of the passes don't have to deal with it.

svuckovicTT: IMO that's not a good solution. Translation should be a mechanical process, it shouldn't reason about IR. At the very worst, we can have a `dedup` pass at the end of the pipeline, but let's sync offline to discuss approaches.

dmilinkovicTT: Synced offline, continued with a different approach

lib/Target/Python/TranslateToPython.cpp
```diff
@@ -521,26 +521,19 @@ static LogicalResult printOperation(PythonEmitter &emitter,
 static LogicalResult printOperation(PythonEmitter &emitter, ModuleOp moduleOp) {
   PythonEmitter::Scope scope(emitter);
 
-  // When file ops exist, defer module-level imports so they appear inside
-  // each file section rather than before the first file label.
+  // When file ops exist, emit each file's contents with a label separator.
+  // Imports are already placed inside each FileOp by the conversion passes.
```
amilovanovicTT: Nit:
```suggestion
  // When file ops exist, emit the content of a specified file if the fileId
  // is present, otherwise emit each file's contents with a label separator.
```

PR #7817, opened 2026-04-04
Skip L1 ops in ConstEvalHoistTransform to prevent unaccounted L1 pressure

### LINKED ISSUE #7813: ConstEvalHoistTransform hoists L1 tensors, causing unaccounted L1 pressure in subsequent functions
## Summary

`ConstEvalHoistTransform` runs **after** `L1SpillManagement` in the pipeline and hoists `to_memory_config` ops whose outputs are in L1 into const_eval functions. These L1 tensors persist across function boundaries — when `@main` executes, the const_eval's L1 allocations are still live, but `L1SpillManagement` made its decisions assuming the full L1 budget was available.

## Observed Impact

On `qwen_3_4b` (branch `rpavlovic/test-greedy-pr5`), `@main_const_eval_187` produces a `tensor<1x151936xf32>` width-sharded across 63 cores at **311,296 bytes per shard** in L1. This is an argmax index tensor used late in the graph. It occupies 311K of L1 per core for the **entire duration** of `@main`, reducing the effective L1 budget by ~23%.

This contributes to a runtime L1 clash:
```
Statically allocated circular buffers in program clash with L1 buffers
on core range [(x=0,y=0) - (x=7,y=1)]. L1 buffer allocated at 854016
and static circular buffer region ends at 857376
```

## Root Cause

Pipeline order in `TTNNPipelines.cpp`:
1. `createTTNNPipelineAnalysisPasses` — includes `GreedyL1SpillManagement`
2. `createConstEvalHoistTransform()` — re-runs const-eval **after** analysis passes

After step 1, `@main` contains:
```mlir
%187 = ttcore.load_cached(@main_const_eval_187, []) 
    : () -> tensor<1x151936xf32, ... #dram, <interleaved>>  // const_eval returns DRAM

%3929 = "ttnn.to_memory_config"(%187) <{memory_config = #ttnn.memory_config<<l1>, <width_sharded>, ...>}>
    : (tensor<...#dram>) -> tensor<...#l1, <width_sharded>>  // reshards to L1 in @main
```

After step 2, `ConstEvalHoistTransform` hoists the `to_memory_config` into the const_eval function:
```mlir
// const_eval_187 now returns L1 width-sharded (311K/shard) instead of DRAM
%187 = ttcore.load_cached(@main_const_eval_187, [])
    : () -> tensor<1x151936xf32, ... #l1, <width_sharded>>  // 311K per core, persists in L1!
```

At runtime, const_eval runs first, allocates 311K in L1, then `@main` runs with that allocation still live — but its spill decisions assumed full L1 was available.

## Proposed Fix

`ConstEvalHoistTransform` should **not hoist ops whose outputs are in L1** into const_eval functions. The constant + reshape ops should be hoisted (they produce DRAM outputs), but the `to_memory_config` that reshards to L1 should remain in `@main` where `L1SpillManagement` can account for it.

Const-eval functions should only produce DRAM/system_memory outputs, since their results persist across function boundaries and consume L1 budget that the optimizer doesn't account for.

## Minimal Repro

```mlir
// test_const_eval_l1_bug.mlir
module @const_eval_l1_repro attributes {} {
  func.func @main(%arg0: tensor<1x1024xf32>) -> tensor<1x1024xf32> {
    %cst = "ttir.arange"() <{arange_dimension = 1 : i64, end = 1024 : si64, start = 0 : si64, step = 1 : si64}> : () -> tensor<1x1024xf32>
    %0 = "ttir.multiply"(%arg0, %cst) : (tensor<1x1024xf32>, tensor<1x1024xf32>) -> tensor<1x1024xf32>
    %1 = "ttir.add"(%0, %cst) : (tensor<1x1024xf32>, tensor<1x1024xf32>) -> tensor<1x1024xf32>
    return %1 : tensor<1x1024xf32>
  }
}
```

```bash
# After L1SpillManagement: const_eval returns DRAM
ttmlir-opt --ttcore-register-device="system-desc-path=system_desc.ttsys" \
  "--ttir-to-ttnn-backend-pipeline=enable-optimizer=true \
    enable-greedy-optimizer=true optimization-level=2 \
    system-desc-path=system_desc.ttsys" \
  -mlir-print-ir-after=ttnn-greedy-l1-spill-management \
  test_const_eval_l1_bug.mlir 2>&1 | grep "const_eval_0.*->"
# Output: -> tensor<1x1024xf32, ... #dram, <interleaved>>

# After full pipeline: const_eval returns L1!
ttmlir-opt --ttcore-register-device="system-desc-path=system_desc.ttsys" \
  "--ttir-to-ttnn-backend-pipeline=enable-optimizer=true \
    enable-greedy-optimizer=true optimization-level=2 \
    system-desc-path=system_desc.ttsys" \
  test_const_eval_l1_bug.mlir 2>/dev/null | grep "const_eval_0.*->"
# Output: -> tensor<1x1024xf32, ... #l1, <width_sharded>>
```

rpavlovicTT:
## Summary
- Prevents `ConstEvalHoistTransform` from hoisting ops whose results are in L1 into const_eval functions
- Const-eval outputs now stay in DRAM; the DRAM→L1 reshard remains in `@main` where `L1SpillManagement` can account for it

## Problem

`ConstEvalHoistTransform` runs after `L1SpillManagement` in the pipeline and can hoist `to_memory_config` ops with L1 outputs into const_eval functions. These L1 tensors persist across function boundaries — when `@main` executes, the const_eval's L1 allocations are still live, but `L1SpillManagement` made its decisions assuming the full L1 budget was available.

On `qwen_3_4b`, `const_eval_187` produced a 311K/core L1 width-sharded tensor (argmax indices) that consumed ~23% of L1 for the entire duration of `@main`, contributing to a runtime CB/tensor buffer clash.

## Fix

Added an L1 result check in `ConstEvalAnalyze::processOp()` — if any result has `TTNNLayoutAttr` with `hasL1BufferType()`, the op is skipped from const-eval hoisting.

## Test plan
- [x] Verified minimal repro: `arange` + `multiply` + `add` — const_eval returns DRAM after fix (was L1 before)
- [x] Verified full `qwen_3_4b` TTIR compilation — `const_eval_187` returns `#dram, <interleaved>` after fix
- [ ] Existing const_eval tests should pass (hoisting of DRAM ops is unaffected)

Closes #7813

### DIFF
--- a/lib/Transforms/ConstEvalHoist.cpp
+++ b/lib/Transforms/ConstEvalHoist.cpp
@@ -193,6 +193,22 @@ private:
       return;
     }
 
+    // Skip ops whose results are in L1. Const-eval functions persist their
+    // outputs across function boundaries, so L1 allocations from const-eval
+    // would consume L1 budget during @main execution without the
+    // L1SpillManagement pass being able to account for them.
+    for (auto result : op->getResults()) {
+      if (auto tensorType =
+              mlir::dyn_cast<mlir::RankedTensorType>(result.getType())) {
+        if (auto layoutAttr = mlir::dyn_cast_or_null<ttnn::TTNNLayoutAttr>(
+                tensorType.getEncoding())) {
+          if (layoutAttr.hasL1BufferType()) {
+            return;
+          }
+        }
+      }
+    }
+
     auto operandConstEval = [&](mlir::Value operand) {
       if (auto blockArg = mlir::dyn_cast<mlir::BlockArgument>(operand)) {
         return constParams.contains(blockArg);


rpavlovicTT:
> This is fine for now, but this introduces some pretty awkward coupling between the optimizer and the const-eval pass.
> 
> Could we create an issue to track this?
> 
> One idea: not sure would it be possible to emit `prepare_conv2d_weights` op before the second const-eval pass (and thus before the optimizer)? We would just need to utilize `DevicePassesWrapper` in another part of the pipeline.
> 
> This would eliminate the need for the third const-eval pass, and the IR would be more or less semantically identical after the optimizer - only layout decomposition, deallocation and tracing passes would be run.
> 
> Wdyt?

I agree with your thoughts, but I think it is not possible to create `prepare_conv_weights` ops before `conv2d` sharding which is done in optimizer passes. But we are working on removing the need for `prepare_conv_weights` in IR so we will be able to just remove the 3rd const eval pass.

PR #7818, opened 2026-04-04
Forward compute_kernel_config to OpModel for RMSNormOp

rpavlovicTT:
## Summary
- Forward `compute_kernel_config` from the IR to the OpModel query for `RMSNormOp`
- Fixes cbPeakUsage underreporting when `fp32_dest_acc_en=true` is set

## Problem

`RMSNormConfigRewritePattern` sets `fp32_dest_acc_en=true` on rms_norm ops for maximum numerical precision. With `fp32_dest_acc_en=true`, intermediate circular buffers use Float32 tiles (4096B) instead of Float16_b (2048B), significantly increasing CB peak usage.

However, the OpModel query passed `compute_kernel_config=std::nullopt`, causing the backend to use default `fp32_dest_acc_en=false` when estimating CB sizes. This led to cbPeakUsage being underreported by ~28% (585,728 reported vs 753,664 actual for `544x2560 bf16` rms_norm).

## Fix

Plumb `compute_kernel_config` through all three layers:
- `TTNNOpModel.h`: add `computeKernelConfig` parameter to `getOpConstraints`/`getOpRuntime`
- `TTNNOpModel.cpp`: convert via `getDeviceComputeKernelConfig()` and forward to `QUERY_OP_CONSTRAINTS`/`RUNTIME`
- `TTNNOpModelInterface.cpp`: extract `getComputeConfigAttr()` from the op and pass through

## Test plan
- [ ] Verified on `qwen_3_4b` TTIR: cbPeakUsage now reports 753,664 (matches runtime) instead of 585,728
- [ ] Default behavior unchanged: `computeKernelConfig` defaults to `std::nullopt` so callers without compute config are unaffected

### DIFF
--- a/include/ttmlir/OpModel/TTNN/TTNNOpModel.h
+++ b/include/ttmlir/OpModel/TTNN/TTNNOpModel.h
@@ -1513,23 +1513,26 @@ struct OpModel<BatchNormTrainingOp> {
 
 template <>
 struct OpModel<RMSNormOp> {
-  static llvm::Expected<OpConstraints>
-  getOpConstraints(ttcore::GridAttr deviceGrid,
-                   llvm::ArrayRef<int64_t> inputShape,
-                   TTNNLayoutAttr inputLayout,
-                   std::optional<llvm::ArrayRef<int64_t>> weightShape,
-                   std::optional<TTNNLayoutAttr> weightLayout,
-                   std::optional<llvm::ArrayRef<int64_t>> biasShape,
-                   std::optional<TTNNLayoutAttr> biasLayout,
-                   llvm::APFloat epsilon, TTNNLayoutAttr outputLayout);
+  static llvm::Expected<OpConstraints> getOpConstraints(
+      ttcore::GridAttr deviceGrid, llvm::ArrayRef<int64_t> inputShape,
+      TTNNLayoutAttr inputLayout,
+      std::optional<llvm::ArrayRef<int64_t>> weightShape,
+      std::optional<TTNNLayoutAttr> weightLayout,
+      std::optional<llvm::ArrayRef<int64_t>> biasShape,
+      std::optional<TTNNLayoutAttr> biasLayout, llvm::APFloat epsilon,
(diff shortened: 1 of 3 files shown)


odjuricicTT (approved):

include/ttmlir/OpModel/TTNN/TTNNOpModel.h
```diff
@@ -1513,23 +1513,26 @@ struct OpModel<BatchNormTrainingOp> {
...
+  static llvm::Expected<OpConstraints> getOpConstraints(
+      ttcore::GridAttr deviceGrid, llvm::ArrayRef<int64_t> inputShape,
+      TTNNLayoutAttr inputLayout,
+      std::optional<llvm::ArrayRef<int64_t>> weightShape,
+      std::optional<TTNNLayoutAttr> weightLayout,
+      std::optional<llvm::ArrayRef<int64_t>> biasShape,
+      std::optional<TTNNLayoutAttr> biasLayout, llvm::APFloat epsilon,
+      TTNNLayoutAttr outputLayout,
+      std::optional<DeviceComputeKernelConfigAttr> computeKernelConfig =
+          std::nullopt);
```
sgholamiTT, line 1525: Nit: Not blocking the PR, but for the sake of consistency, I'd keep `TTNNLayoutAttr outputLayout` as the last argument and add `std::optional<DeviceComputeKernelConfigAttr> computeKernelConfig` before that (we don't need the default `std::nullopt` here as you have modified the only caller of this function in `lib/Dialect/TTNN/Interfaces/TTNNOpModelInterface.cpp`):

```c++
      // ...
      std::optional<TTNNLayoutAttr> biasLayout, llvm::APFloat epsilon,
      std::optional<DeviceComputeKernelConfigAttr> computeKernelConfig,
      TTNNLayoutAttr outputLayout);
```

sgholamiTT (approved):

sdjordjevicTT (approved):

PR #7823, opened 2026-04-06
[d2m] rms norm

dlokeTT:
### Problem description
D2M doesn't directly support `ttir.rms_norm` so we need to lower it to supported ttir ops

### What's changed
Decomposed `ttir.rms_norm` into basic ttir ops which are supported in ttmetal
Updated verifier/lit tests to support leading 1 dim for weight and bias since rank normalization runs before this and updates those tensors to be 2d with a leading 1
Added ttmetal to rms_norm builder tests

### Checklist

### DIFF
--- a/lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
+++ b/lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
@@ -3065,75 +3065,6 @@ public:
   }
 };
 
-class D2MRMSNormDecompositionRewriter
-    : public OpConversionPattern<ttir::RMSNormOp> {
-  using OpConversionPattern<ttir::RMSNormOp>::OpConversionPattern;
-
-  LogicalResult
-  matchAndRewrite(ttir::RMSNormOp op, ttir::RMSNormOp::Adaptor adaptor,
-                  ConversionPatternRewriter &rewriter) const override {
-    Location loc = op.getLoc();
-    Value x = adaptor.getInput();
-    auto inputType = cast<RankedTensorType>(x.getType());
-    int64_t rank = inputType.getRank();
-
-    auto xSquared = rewriter.create<ttir::MultiplyOp>(loc, inputType, x, x);
-
-    SmallVector<int64_t> reducedShape(inputType.getShape());
-    reducedShape[rank - 1] = 1;
-    auto reducedType = RankedTensorType::get(
-        reducedShape, inputType.getElementType(), inputType.getEncoding());
-
-    auto meanOp = rewriter.create<ttir::MeanOp>(
-        loc, reducedType, xSquared.getResult(), rewriter.getBoolAttr(true),
-        rewriter.getI32ArrayAttr({static_cast<int32_t>(rank - 1)}));
-
-    float epsilon = op.getEpsilon().convertToFloat();
-    auto epsOp = rewriter.create<ttir::FullOp>(
-        loc, reducedType, rewriter.getF32FloatAttr(epsilon));
-
-    auto addEps = rewriter.create<ttir::AddOp>(
-        loc, reducedType, meanOp.getResult(), epsOp.getResult());
-
-    auto rsqrt =
-        rewriter.create<ttir::RsqrtOp>(loc, reducedType, addEps.getResult());
-
-    auto normalized =
-        rewriter.create<ttir::MultiplyOp>(loc, inputType, x, rsqrt.getResult());
-
-    Value result = normalized.getResult();
-
-    auto reshapeToInputRank = [&](Value v) -> Value {
-      auto vType = cast<RankedTensorType>(v.getType());
-      if (vType.getRank() == rank) {
-        return v;
-      }
-      SmallVector<int64_t> newShape(rank - vType.getRank(), 1);
-      newShape.append(vType.getShape().begin(), vType.getShape().end());
-      auto reshapedType = RankedTensorType::get(
-          newShape, vType.getElementType(), vType.getEncoding());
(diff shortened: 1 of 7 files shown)


lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
```diff
@@ -3065,6 +3065,75 @@ class D2MSliceStaticOpNoCConstraintsRewriter
...
+    auto xSquared = rewriter.create<ttir::MultiplyOp>(loc, inputType, x, x);
+
+    SmallVector<int64_t> reducedShape(inputType.getShape());
+    reducedShape[rank - 1] = 1;
+    auto reducedType = RankedTensorType::get(
+        reducedShape, inputType.getElementType(), inputType.getEncoding());
+
+    auto meanOp = rewriter.create<ttir::MeanOp>(
+        loc, reducedType, xSquared.getResult(), rewriter.getBoolAttr(true),
+        rewriter.getI32ArrayAttr({static_cast<int32_t>(rank - 1)}));
```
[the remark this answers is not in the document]
vwellsTT: yeah, I'm confused by this too--does it not need to be conditionally n or n-1?

lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
```diff
@@ -3065,6 +3065,75 @@ class D2MSliceStaticOpNoCConstraintsRewriter
   }
 };
 
+class D2MRMSNormDecompositionRewriter
```
vwellsTT: I think most of this logic LGTM, but I'm not sure it's good to put decomposition into TTIRToD2M.cpp vs a separate pre-conversion TTIRToTTIR decomp pass like we've done for similar patterns? Not really sure

lib/Dialect/TTIR/IR/TTIROps.cpp
```diff
@@ -5798,20 +5798,44 @@ ::mlir::LogicalResult mlir::tt::ttir::RMSNormOp::verify() {
...
-    if (biasType.getShape() != normalizedShape) {
+    ArrayRef<int64_t> biasShape = getBias().getType().getShape();
+    size_t normRank = normalizedShape.size();
+    size_t biasRank = biasShape.size();
+    if (biasRank != normRank && biasRank != normRank + 1) {
+      return emitOpError(
+          "bias tensor rank must match normalized_shape rank or exceed it by "
+          "at most one (single leading broadcast dimension)");
+    }
+    if (biasShape.take_back(normRank) != normalizedShape) {
```
vwellsTT: are these changes safe from TTNN perspective? I'm not familiar enough with this op to have opinions

vwellsTT:
>Updated verifier/lit tests to support leading 1 dim for weight and bias since rank normalization runs before this and updates those tensors to be 2d with a leading 1

Does this mean if we ran this decomp before rank normalization it would simplify?

lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
```diff
@@ -3146,6 +3215,10 @@ void populateTTIRToD2MPatterns(MLIRContext *ctx, RewritePatternSet &patterns,
     D2MAllGatherRewriter
   >(typeConverter, ctx, defaultInputMemSpace, defaultOutputMemSpace, ttnnMode, collapseTensors, enableMulticastInference);
 
+  // High-priority decompositions that expand fused TTIR ops into primitives.
+  patterns.add<D2MRMSNormDecompositionRewriter>(typeConverter, ctx,
+                                                /*benefit=*/10);
```
nsmithtt: Do we really need to set the benefit?  I would expect the RMS pattern match to happen regardless of benefit since it's the only pattern for it

lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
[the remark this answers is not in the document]
dlokeTT: updated to match what the op actually supports. all the test cases/variations found in models were just reducing on the last dim so had that initially.

lib/Dialect/TTIR/IR/TTIROps.cpp
dlokeTT: Not really sure, so I moved this into it's own pass before rank normalization so we don't have to modify this

lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
dlokeTT: Nick had suggested doing it in `TTIRToD2M.cpp`. Moved it into its own pass to follow how other decomps have been done in the past and it removes updating the verifier

lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
dlokeTT: Yeah you're right, we didn't need to set the benefit. I followed @vwellsTT suggestion and moved this into it's own pass to follow how we did other decompositions such as min, reshape and permute. This also lets us not mess with the verifier since it runs before rank normalization, making overall changes less impactful.

I think we should consolidate on how to do all these decompositions since it keeps coming up as we're doing more and more of them to get ops working e2e. We have the `TTIRToTTIRDecomposition.cpp` pass which does some decomps, then we have some passes that do decomps for certain ops such as this and reshape/permute, and we do some decomps in `TTIRToD2M.cpp` such as the comparisons, clamp_tensor and others. I can come up with a proposal for all these so we have a set way to do decomps in the future.

Issue #7803, opened 2026-04-03
SliceOp with complex tensors is not legalized


ppadjinTT:
### Describe the bug

This PR https://github.com/tenstorrent/tt-mlir/pull/7407  introduced complex datatypes in tt-mlir and the followup conversion patterns for supporting different ops on complex tensors.

However, slice op decomposition is still missing, as seen in the example bellow.
When stablehlo.slice op has complex type as input/output, no decomposition is applies so complex types remain, however, the rest of the ops work with real tensors after the `StableHLOComplexDataTypeConversionPass`.

Compiler tries to insert a typecast from complex to/from real tensor and fails with:
```
error: failed to legalize unresolved materialization from ('tensor<64x32x2xf32>') to ('tensor<64x32xcomplex<f32>>') that remained live after conversion
```

I attached the whole execution log:

[error.log](https://github.com/user-attachments/files/26459819/error.log)

### tt-xla repro

```
import torch
import torch.nn as nn
import torch_xla
import torch_xla.runtime as xr


class ComplexSlice(nn.Module):
    def forward(self, freqs: torch.Tensor) -> torch.Tensor:
        return freqs[:, :28]


xr.set_device_type("TT")
device = torch_xla.device()

# Complex RoPE frequencies, same as QwenEmbedLayer3DRope.rope_params
index = torch.arange(64, dtype=torch.float32)
inv_freq = 1.0 / torch.pow(10000.0, torch.arange(0, 64, 2, dtype=torch.float32) / 64)
freqs = torch.polar(torch.ones(64, 32), torch.outer(index, inv_freq))

model = ComplexSlice().to(device)
compiled = torch.compile(model, backend="tt")

with torch.no_grad():
    result = compiled(freqs.to(device))

print(f"Output: {result.shape}, {result.dtype}")
```

### stablehlo repro
```
#loc1 = loc("/root/tt-xla/python_package/tt_torch/torch_overrides.py":22:0)
#loc2 = loc("/root/tt-xla/venv/lib/python3.12/site-packages/torch_xla/_dynamo/dynamo_bridge.py":756:0)
#loc3 = loc("/root/tt-xla/venv/lib/python3.12/site-packages/torch_xla/_dynamo/dynamo_bridge.py":859:0)
#loc4 = loc("/root/tt-xla/venv/lib/python3.12/site-packages/torch_xla/_dynamo/dynamo_bridge.py":737:0)
[...]

PR #7824, opened 2026-04-06
Adding conversions for slice and concat ops with complex tensors

ppadjinTT:
### Ticket
https://github.com/tenstorrent/tt-mlir/issues/7803

### Problem description
Slice and Concat operations working on complex tensors are not legalized. We need an explicit conversion pass. We have to explain how does slice/concat change when we convert complex tensor `d1 x d2 x..x dn` -> real tensor `d1 x d2 x..x dn x 2`.

### What's changed
Added `ComplexSliceOpConversionPattern` and `ComplexConcatenateOpConversionPattern` which legalize these ops.

`ComplexSliceOpConversionPattern`: When we do `complex_tensor[slice]`, that is equivalent to `real_tensor[new_slice]` where `new_slice=[*slice, 0:2:1]`, because we want to select everything in the new trailing dimension.

`ComplexConcatenateOpConversionPattern:` Concatenate op looks the same, we just change the inputs/result type from complex to real. 

### Checklist
- [DONE] Add FileCheck tests 
- [TODO] Add tt-xla level tests for a module that produced these (Qwen Image Positional Embeddings torch.nn.Module)

### DIFF
--- a/lib/Dialect/StableHLO/Transforms/ComplexDataTypeConversion.cpp
+++ b/lib/Dialect/StableHLO/Transforms/ComplexDataTypeConversion.cpp
@@ -228,8 +228,8 @@ public:
     auto newResultType = mlir::cast<RankedTensorType>(
         this->getTypeConverter()->convertType(op.getResult().getType()));
     rewriter.replaceOpWithNewOp<OpTy>(op, TypeRange{newResultType},
-                                     adaptor.getOperands(),
-                                     op.getProperties());
+                                      adaptor.getOperands(),
+                                      op.getProperties());
     return success();
   }
 };


lib/Dialect/StableHLO/Transforms/ComplexDataTypeConversion.cpp
```diff
@@ -286,6 +286,61 @@ class ComplexBroadcastInDimOpConversionPattern
...
+      ConversionPatternRewriter &rewriter) const override {
+    auto newResultType = mlir::cast<RankedTensorType>(
+        this->getTypeConverter()->convertType(op.getResult().getType()));
+
+    // same concat, just changing inputs/result from complex to float
+    rewriter.replaceOpWithNewOp<mlir::stablehlo::ConcatenateOp>(
+        op, newResultType, adaptor.getInputs(), op.getDimension());
+    return success();
+  }
+};
```
azecevic000: There is a `ComplexTypeDefaultConversionPattern`, it has a subtle bug though, in `replaceOpWithNewOp` where `adaptor.getAttributes()` should be added as a last argument. That pattern should cover this case.

azecevicTT (approved):

PR #7834, opened 2026-04-06
[D2M] Refactor - move all compute CB op insertion into thread splitter

vtangTT:
### Ticket
Link to Github Issue

### Problem description
CB sync ops (reserve/push/wait/pop) were inserted across 3 separate passes in the DMA lowering pipeline:
- ConvertLocalLoadStoreOpsToAliasedCBs
- LowerLoadStoreOpsToExplicitCBForm
- LowerLoadStoreOpsToDMA 

Each pass had implicit ordering contracts with the others: pass 2 assumed pass 1 had already handled aliased ops, pass 3 assumed pass 2 had converted everything to explicit CB form, and the splitter assumed all compute-side CB ops were already present. A bug in any pass's assumptions about prior passes would silently produce incorrect IR.
                                                                                                                                                                                                                                                                                        
Pros:
  - single source of truth for compute-side CB op insertion
  - net reduction of line of code w/ same coverage                                                                                                                                                                                                                
  - the splitter is the natural place for CB ops since they are the inter-thread communication mechanism
  - easier bring-up for future DMA ops: change 1 pass vs changing 3

Cons:                                                                                                                                                                                                                                                                                 
  - splitter is more complex: went from ~180 lines to ~700 
  - harder to test individual concerns in isolation: all CB insertion is now tested through `--d2m-split-unified-thread`

### What's changed
Consolidated `ConvertLocalLoadStoreOpsToAliasedCBs` and `LowerLoadStoreOpsToExplicitCBForm` into `SplitUnifiedThread`. New enhanced splitter now: 
- clones all ops from the unified thread into both DMA and compute regions
- compute thread: inserts CB sync ops for each remote_load/store based on aliased vs streaming
    - aliased ops get full reserve/push/wait/pop
    - streaming loads get wait/pop 
    - streaming stores get reserve/push
    - shared buffer pairs are detected and handled with proper CB sharing
- DMA thread: converts streaming remote_load/store to explicit CB form, aliased ops are erased
- cleans up dead ops from both threads (non-DMA ops from DMA, non-compute ops from compute) like before

### Checklist
- [x] passing nightly: https://github.com/tenstorrent/tt-mlir/actions/runs/24268512289

### DIFF
--- a/lib/Dialect/D2M/Transforms/SplitUnifiedThread.cpp
+++ b/lib/Dialect/D2M/Transforms/SplitUnifiedThread.cpp
@@ -288,6 +288,9 @@ static LogicalResult processComputeLoads(Block *computeBlock,
       rewriter.create<PushOp>(loc, cb);
       auto waitOp = rewriter.create<WaitOp>(loc, cb);
 
+      // Replace all uses of the alloc, not just the load's operand as
+      // downstream compute ops reference the alloc result directly and
+      // must read from the CB. Assumes 1:1 alloc-to-load relationship.
       rewriter.replaceAllUsesWith(allocOp.getResult(), waitOp.getResult());
       if (loadOp.getResult()) {
         rewriter.replaceAllUsesWith(loadOp.getResult(), waitOp.getResult());
@@ -358,6 +361,8 @@ static LogicalResult processComputeStores(Block *computeBlock,
       if (allocOp && isLocalAlloc(allocOp, computeBlock)) {
         rewriter.setInsertionPoint(allocOp);
         auto reserveOp = rewriter.create<ReserveOp>(loc, cb);
+        // Replace all uses as compute ops reference the alloc directly and
+        // must write into the CB. Assumes 1:1 alloc-to-store relationship.
         rewriter.replaceAllUsesWith(allocOp.getResult(), reserveOp.getResult());
       } else if (auto waitOp = localBuffer.getDefiningOp<WaitOp>()) {
         // The alloc was already replaced by a WaitOp from an earlier load


vtangTT:
@copilot+claude-sonnet-4.5 review this PR

sohaibnadeemTT:
Looks good, but why are CB ops for DMA threads also not inserted in SplitUnifiedThread? It seems like the natural place for all CB ops to be inserted.

include/ttmlir/Dialect/D2M/Utils/CBUtils.h
```diff
@@ -8,6 +8,12 @@
 #include "mlir/IR/PatternMatch.h"
 #include "llvm/ADT/DenseMap.h"
 
+namespace mlir {
```
vwellsTT: better to group with GenericOp below I guess?

lib/Dialect/D2M/Transforms/SplitUnifiedThread.cpp
```diff
@@ -18,170 +21,707 @@ namespace mlir::tt::d2m {
...
+// ---------------------------------------------------------------------------
+// Helpers
+// ---------------------------------------------------------------------------
+
+// A streaming op requires real DMA. Returns true if the remote memref
+// implies data movement (view layout, DRAM) or the local buffer is a
+// streaming CB (CBLayoutAttr). Aliased ops (plain L1 buffers accessing
+// L1 shard operands) need no DMA.
+static bool isStreamingOp(Value memref, Value localBuffer) {
+  // View ops are streaming, except for reinterpret view_layout ops
```
vwellsTT: I feel like this nomenclature is a bit inaccurate/confusing; imo the buffer check is more correct for "streaming".  I think that if for some reason a view exists, but doesn't get streaming buffers (not sure that can even happen), it would still need DMA of course, but if that's a real case this should be named something like "needsDMA" or whatnot instead of "isStreaming"?

lib/Dialect/D2M/Transforms/SplitUnifiedThread.cpp
```diff
@@ -18,170 +21,707 @@ namespace mlir::tt::d2m {
...
+    }
+  });
+
+  for (RemoteLoadOp loadOp : loads) {
+    Location loc = loadOp.getLoc();
+    Value memref = loadOp.getMemref();
+    Value localBuffer = loadOp.getLocalBuffer();
+    Value cb = findAssociatedCB(loadOp, memref, rewriter, cache, portCounters);
+    if (!cb) {
+      loadOp.emitWarning("could not find associated CB, skipping conversion");
```
vwellsTT: can this validly happen? I feel like some of these should be errors vs warnings + skip, but not 100% sure

lib/Dialect/D2M/Transforms/SplitUnifiedThread.cpp
```diff
@@ -18,170 +21,707 @@ namespace mlir::tt::d2m {
...
+    // DMA thread: convert streaming ops to explicit CB form.
+    convertDMAToExplicitCBForm(dmBlock, rewriter, dmaCache, portCounters,
+                               toErase);
+
+    // Clean up: first erase non-DMA ops from DMA and DMA ops from compute
+    // (this removes users of old ops, making them use_empty), then erase
+    // the collected old ops themselves.
+    eraseCollectedOps(rewriter, toErase);
+    eraseDeadOps(rewriter, dmBlock, /*isDatamovementThread=*/true);
+    eraseDeadOps(rewriter, computeBlock, /*isDatamovementThread=*/false);
```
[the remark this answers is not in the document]
vtangTT: incorrect, all uses are dropped or replaced

lib/Dialect/D2M/Transforms/SplitUnifiedThread.cpp
```diff
@@ -18,170 +21,707 @@ namespace mlir::tt::d2m {
...
+    // store's (output) memref for CB lookup so both ops share the same port.
+    // L1-to-L1 pairs use the load's own (input) memref since the store side
+    // also redirects to the input memref for non-streaming loads.
+    Value cbMemref = memref;
+    auto pairIt = sharedPairs.find(loadOp.getLocalBuffer());
+    if (pairIt != sharedPairs.end() && !l1Pair) {
+      cbMemref = pairIt->second.second.getMemref();
+    }
+
+    // Streaming -> convert to explicit CB form.
```
vwellsTT: nit: what does this comment mean, isn't everything being converted  to explicit CB form?

lib/Dialect/D2M/Transforms/SplitUnifiedThread.cpp
```diff
@@ -18,170 +21,707 @@ namespace mlir::tt::d2m {
...
+    // load's (input) CB when the load is aliased, to match compute.
+    Value cbMemref = memref;
+    auto storePairIt = sharedPairs.find(storeOp.getLocalBuffer());
+    if (storePairIt != sharedPairs.end()) {
+      RemoteLoadOp pairedLoad = storePairIt->second.first;
+      if (!isStreamingOp(pairedLoad.getMemref(), pairedLoad.getLocalBuffer())) {
+        // Load is aliased — use its (input) CB.
+        cbMemref = pairedLoad.getMemref();
+      }
+      // else: load is streaming — use store's own (output) CB (default).
```
vwellsTT: nit: sentence-ify

lib/Dialect/D2M/Transforms/SplitUnifiedThread.cpp
```diff
@@ -18,170 +21,707 @@ namespace mlir::tt::d2m {
...
+    }
+
+    // Streaming -> convert to explicit CB form.
```
vwellsTT: same comment as above, and I have same nit

lib/Dialect/D2M/Transforms/SplitUnifiedThread.cpp
```diff
@@ -18,170 +21,707 @@ namespace mlir::tt::d2m {
...
+    rewriter.create<RemoteStoreOp>(
+        storeOp.getLoc(), memref, storeOp.getIndices(), cb,
+        storeOp.getStartDevice(), storeOp.getDeviceMcastShape(),
+        storeOp.getSemaphore(), storeOp.getSemaphoreIndices());
+    toErase.insert(storeOp);
+  }
+}
+
+// ---------------------------------------------------------------------------
+// Dead-op cleanup
```
vwellsTT, line 558: what does this cleanup that doesn't get automatically DCE'ed by `canonicalize`?

lib/Dialect/D2M/Transforms/SplitUnifiedThread.cpp
vtangTT: hmm yea agreed, will rename to `needsDMA`

lib/Dialect/D2M/Transforms/SplitUnifiedThread.cpp
vtangTT: every implicit-form remote_load/store should have a findable CB.. so will change these to return error

lib/Dialect/D2M/Transforms/SplitUnifiedThread.cpp
vtangTT: we are doing thread-aware cleanup, where we remove all non-DMA ops from DM thread (i.e. compute ops) and removing DMA ops from compute thread. 

these ops will still have users so we iteratively erase them (erase leaf, then their now dead producer)

lib/Dialect/D2M/Transforms/SplitUnifiedThread.cpp
vtangTT: fixed

include/ttmlir/Dialect/D2M/Utils/CBUtils.h
vtangTT: fixed

lib/Dialect/D2M/Transforms/SplitUnifiedThread.cpp
vtangTT: yep, removed

lib/Dialect/D2M/Transforms/SplitUnifiedThread.cpp
vtangTT: fixed

lib/Dialect/D2M/Utils/CBUtils.cpp
```diff
@@ -136,4 +137,30 @@ Value findAssociatedCB(Operation *op, Value memrefOperand, IRRewriter &rewriter,
...
+    return nullptr;
+  }
+
+  // Direct case: value is directly produced by memref.alloc.
+  if (auto allocOp = mlir::dyn_cast<memref::AllocOp>(definingOp)) {
+    return allocOp;
+  }
+
+  // Trace through operations that might pass the buffer through
+  // (e.g., view-like ops, cast ops, etc.)
```
anuragsingh-tt: The header comment says “through a chain of view-like operations” but the code recursively walks all operands of any defining op. And what if multiple views consume the same root alloc? This would replace uses of all of them. Is that expected?

lib/Dialect/D2M/Transforms/SplitUnifiedThread.cpp
```diff
@@ -18,170 +21,696 @@ namespace mlir::tt::d2m {
 
 namespace {
 
+// ---------------------------------------------------------------------------
+// Helpers
+// ---------------------------------------------------------------------------
+
+// Returns true if the remote_load/store requires real DMA. This is the case
+// when the remote memref has a view layout, is in DRAM, or the local buffer is
```
anuragsingh-tt, line 29: NIT: these could be separate items.

lib/Dialect/D2M/Transforms/SplitUnifiedThread.cpp
```diff
@@ -18,170 +21,696 @@ namespace mlir::tt::d2m {
...
+    DenseSet<Operation *> toErase;
+
+    // Compute thread: insert CB sync ops for implicit-form remote ops.
+    if (failed(processSharedBufferPairs(computeBlock, rewriter, computeCache,
+                                        portCounters, toErase)) ||
+        failed(processComputeLoads(computeBlock, rewriter, computeCache,
+                                   portCounters, toErase)) ||
+        failed(processComputeStores(computeBlock, rewriter, computeCache,
+                                    portCounters, toErase))) {
+      return failure();
```
anuragsingh-tt, line 704: We modified the IR before this point, so if we return a failure the IR has been modified. I am wondering if we should do the analysis prior to the execution.

vtangTT: Hmm, to me these errors mean that an upstream pass produced non-valid IR to the splitters so I think its ok to just return an error even tho the IR has half modified. In this case, adding a pre-validation walk seems to be extra burden, especially if the problem didn't originate in this pass.

lib/Dialect/D2M/Utils/CBUtils.cpp
vtangTT: This is in fact a bug, fixed now to only trace through view-like operations as described by the comment.

>what if multiple views consume the same root alloc

I don't think this is an existing pattern in the pipeline right now? And the old pipeline (prior to this refactor) handled this case the same way since it also assumes a 1:1 alloc-to-load relationship. If multi-view sharing is introduced, i think we'd need to scope the replacement to only the relevant chain of ops.

lib/Dialect/D2M/Transforms/SplitUnifiedThread.cpp
anuragsingh-tt: Generally in an `OpRewritePattern`, returning `failure()` after mutating IR can leave the greedy driver with partially rewritten state which is not the right contract. But I guess the fallout is low, its really more of a debugging thing and we are not doing this consistently in passes anyways. Should be fine.

lib/Dialect/D2M/Utils/CBUtils.cpp
anuragsingh-tt: > I don't think this is an existing pattern in the pipeline right now?

Yes.

>since it also assumes a 1:1 alloc-to-load relationship. If multi-view sharing is introduced, i think we'd need to scope the replacement to only the relevant chain of ops.

Yes. But maybe we can future proof? Is it necessary to change all the uses of `localBuffer` at this time not just the `allocOp.getResult`? If it is required, let's document the baked in assumption here and we can move on.

anuragsingh-tt: I am thinking of a case where we do one large allocations and have multiple subviews of that alloc that refer to different parts of the buffer. But I guess that is not really our model right now. Not sure if it will change...

vtangTT: > Is it necessary to change all the uses of localBuffer at this time not just the allocOp.getResult? If it is required, let's document the baked in assumption here and we can move on.

yes since compute ops reference the alloc directly and must write into the CB.

vtangTT: added comments describing said assumption

PR #7859, opened 2026-04-07
Add experimental KV cache dtype conversion pass

kdimicTT:
### Problem description
KV cache tensors consume significant device memory. Storing them in lower-precision block floating-point formats (BFP8) reduces memory usage.

### What's changed
                                                                                                                                             
- New TTNNKVCacheDtypeConversion pass is added. Converts KV cache function argument types to a target BFP dtype, inserts ttnn.typecast ops on inputs to fill_cache and update_cache so written data matches the new cache format.
- Added block format dtype support in TensorDesc so the runtime correctly handles BFP-typed KV cache tensors.
- Generalized WeightDtype to BFPDtype since it is now shared across conversion passes. 

### Note

BFP4 is not a supported for KV cache. `UpdateKVCacheOperation::validate_on_program_cache_miss` in tt-metal explicitly allows only FLOAT32, BFLOAT16, and BFLOAT8_B for both the input and cache tensors, causing TT_FATAL at runtime for BFLOAT4_B.

### Checklist


include/ttmlir/Dialect/TTNN/Pipelines/TTNNPipelines.h
```diff
@@ -11,7 +11,7 @@
 #include "ttmlir/Dialect/TTNN/Utils/MathFidelityParser.h"
```
vmilosevic, line 12: :warning: **llvm\-include\-order** :warning:
```suggestion
#include "ttmlir/Dialect/TTNN/Utils/BFPDtypeParser.h"
#include "ttmlir/Dialect/TTNN/Utils/MathFidelityParser.h"
```

include/ttmlir/Dialect/TTNN/Pipelines/TTNNPipelines.h
```diff
@@ -11,7 +11,7 @@
...
 #include "ttmlir/Dialect/TTNN/Utils/MemoryLayoutAnalysisParams.h"
 #include "ttmlir/Dialect/TTNN/Utils/PassOverrides.h"
-#include "ttmlir/Dialect/TTNN/Utils/WeightDtypeParser.h"
+#include "ttmlir/Dialect/TTNN/Utils/BFPDtypeParser.h"
```
vmilosevic: :warning: **llvm\-include\-order** :warning:
```suggestion

```

lib/Dialect/TTNN/Transforms/TTNNKVCacheDtypeConversion.cpp
```diff
@@ -0,0 +1,170 @@
+#include "ttmlir/Dialect/TTCore/IR/TTCore.h"
+#include "ttmlir/Dialect/TTCore/IR/Utils.h"
+#include "ttmlir/Dialect/TTNN/IR/TTNNOps.h"
+#include "ttmlir/Dialect/TTNN/Transforms/Passes.h"
+#include "ttmlir/Dialect/TTNN/Utils/Utils.h"
```
vmilosevic: :warning: **llvm\-include\-order** :warning:
```suggestion
#include "ttmlir/Dialect/TTNN/Utils/BFPDtypeParser.h"
#include "ttmlir/Dialect/TTNN/Utils/Utils.h"
```

lib/Dialect/TTNN/Transforms/TTNNKVCacheDtypeConversion.cpp
```diff
@@ -0,0 +1,170 @@
...
+#include "ttmlir/Dialect/TTNN/Utils/BFPDtypeParser.h"
```
vmilosevic, line 9: :warning: **llvm\-include\-order** :warning:
```suggestion

```

lib/Dialect/TTNN/Transforms/TTNNWeightDtypeConversion.cpp
```diff
@@ -7,7 +7,7 @@
 #include "ttmlir/Dialect/TTNN/IR/TTNNOps.h"
 #include "ttmlir/Dialect/TTNN/Transforms/Passes.h"
 #include "ttmlir/Dialect/TTNN/Utils/Utils.h"
```
vmilosevic: :warning: **llvm\-include\-order** :warning:
```suggestion
#include "ttmlir/Dialect/TTNN/Utils/BFPDtypeParser.h"
#include "ttmlir/Dialect/TTNN/Utils/Utils.h"
```

lib/Dialect/TTNN/Transforms/TTNNWeightDtypeConversion.cpp
```diff
@@ -7,7 +7,7 @@
...
-#include "ttmlir/Dialect/TTNN/Utils/WeightDtypeParser.h"
+#include "ttmlir/Dialect/TTNN/Utils/BFPDtypeParser.h"
```
vmilosevic, line 9: :warning: **llvm\-include\-order** :warning:
```suggestion

```

vmilosevic (commented):

runtime/lib/common/types.cpp
```diff
@@ -28,6 +31,13 @@ size_t TensorDesc::volume() const {
   return utils::product(shape.cbegin(), shape.cend());
 }
 
+size_t TensorDesc::sizeBytes() const {
+  if (utils::isBlockFormatDataType(dataType)) {
+    return (physicalVolume / 1024) * utils::blockFormatTileSizeBytes(dataType);
+  }
+  return physicalVolume * itemsize;
```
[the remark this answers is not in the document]
dgolubovicTT: Agreed! I would add some assert before this:
```
     LOG_ASSERT(physicalVolume % 1024 == 0,                                                                                                                                                               
                 "Block format tensor physicalVolume must be a multiple of 1024 "
                 "(elements per 32x32 tile), got ",                                                                                                                                                        
                 physicalVolume); 
```

runtime/lib/common/types.cpp
```diff
@@ -19,7 +19,10 @@ TensorDesc::TensorDesc(const std::vector<uint32_t> &shape,
                        const std::optional<std::vector<uint32_t>> &stride,
                        const std::optional<uint64_t> physicalVolume)
     : shape(shape), dataType(dataType) {
-  this->itemsize = itemsize.value_or(utils::dataTypeElementSize(dataType));
+  this->itemsize =
+      itemsize.value_or(utils::isBlockFormatDataType(dataType)
+                            ? 0
+                            : utils::dataTypeElementSize(dataType));
```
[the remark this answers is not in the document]
dgolubovicTT: It is dangerous to set `itemsize` to 0. It will silently fail if someone uses it on block formats. `sizeBytes()` is the main utility, but I see `itemsize` in several places in runtime. Maybe it is better to use value 1 as approximation and also add some warning there - LOG_WARNING

lib/Dialect/TTNN/Transforms/TTNNKVCacheDtypeConversion.cpp
```diff
@@ -0,0 +1,173 @@
+// SPDX-FileCopyrightText: (c) 2024 Tenstorrent AI ULC
```
dgolubovicTT, line 1: 2026

dgolubovicTT (commented):
Can you look into differences between Paged Fill and Update cache and regular Fill and Update cache. We probably want to test and support both versions. 

In pass `TTNNKVCacheDtypeConversionPass` you register 2 patterns:
```
 patterns.add<KVCacheDtypePattern<FillCacheOp>>(&getContext(), dtype);                                                                                                                                    
  patterns.add<KVCacheDtypePattern<UpdateCacheOp>>(&getContext(), dtype);
```
While we have also PagedFillCacheOp and PagedUpdateCacheOp. Why we only handle FillCacheOp and UpdateCacheOp above?

On the other side, the pass's first phase rewrites all function arguments tagged with {ttcore.kv_cache} to the bfp dtype - regardles of which ops consume them. So after that phase, every cache tensor in the function is BFP-typed. But the second phase only inserts typecasts for FillCacheOp and UpdateCacheOp. If a PagedFillCacheOp receives a bf16 input and writes into a now-BFP typed cache, there's a dtype mismatch with no typecast to bridge it.                                         

Some of these ops require input has same type as cache, and some don't. But I would make input and cache same dtype always just for clarity...

kdimicTT:
> In pass `TTNNKVCacheDtypeConversionPass` you register 2 patterns:
> 
> ```
>  patterns.add<KVCacheDtypePattern<FillCacheOp>>(&getContext(), dtype);                                                                                                                                    
>   patterns.add<KVCacheDtypePattern<UpdateCacheOp>>(&getContext(), dtype);
> ```
> 
> While we have also PagedFillCacheOp and PagedUpdateCacheOp. Why we only handle FillCacheOp and UpdateCacheOp above?


PagedFillCacheOp is now handled in the pass. PagedUpdateCacheOp is intentionally excluded - tt-metal enforces that the input tensor must be FLOAT32 or BFLOAT16 (paged_update_cache_device_operation.cpp:185-187), so inserting a block-format typecast on the input causes TT_FATAL. It also handles the dtype mismatch internally, so adding it is not necessary.

lib/Dialect/TTNN/Transforms/TTNNKVCacheDtypeConversion.cpp
```diff
@@ -0,0 +1,174 @@
...
+        RankedTensorType::get(localShapeType.getShape(), newElementType,
+                              localShapeType.getEncoding());
+    return ttcore::LocalShapeAttr::get(ctx, newLocalShapeType);
+  }
+
+  static void changeCacheArgTypes(func::FuncOp funcOp,
+                                  ttcore::DataType targetDtype) {
+    if (funcOp.isDeclaration()) {
+      return;
+    }
```
dmilinkovicTT, line 85: We should have a more tight check here - `utils::isForwardDeviceFunction(funcOp)`.

It's just a defensive measure - in the future, we might have some non-declarations which shouldn't have this transformation applied automatically.

test/ttmlir/Dialect/TTNN/kv_cache_conversion/update_cache_bfp8.mlir
dmilinkovicTT, line 1: The test name suggests it tests `ttnn.update_cache` op, but the test is actually a negative test for `ttnn.paged_update_cache`.

lib/Dialect/TTNN/Transforms/TTNNKVCacheDtypeConversion.cpp
```diff
@@ -0,0 +1,174 @@
...
+    // written data matches the new cache dtype.
+    mlir::RewritePatternSet patterns(&getContext());
+    patterns.add<KVCacheDtypePattern<FillCacheOp>>(&getContext(), dtype);
+    patterns.add<KVCacheDtypePattern<UpdateCacheOp>>(&getContext(), dtype);
+    patterns.add<KVCacheDtypePattern<PagedFillCacheOp>>(&getContext(), dtype);
+
+    if (failed(
+            mlir::applyPatternsGreedily(getOperation(), std::move(patterns)))) {
+      signalPassFailure();
+    }
```
dmilinkovicTT, line 168: IMHO, greedy pattern rewriter is overkill here. 

A simple `walk` should do the trick, as aiming to transform each KV-cache mutation op at most once.

lib/Dialect/TTNN/Transforms/TTNNKVCacheDtypeConversion.cpp
```diff
@@ -0,0 +1,174 @@
...
+    // function signature.
+    getOperation().walk(
+        [&](func::FuncOp funcOp) { changeCacheArgTypes(funcOp, dtype); });
+
+    // Insert typecast operations on the input operands of all cache ops so
+    // written data matches the new cache dtype.
+    mlir::RewritePatternSet patterns(&getContext());
+    patterns.add<KVCacheDtypePattern<FillCacheOp>>(&getContext(), dtype);
+    patterns.add<KVCacheDtypePattern<UpdateCacheOp>>(&getContext(), dtype);
+    patterns.add<KVCacheDtypePattern<PagedFillCacheOp>>(&getContext(), dtype);
```
dmilinkovicTT, line 163: What about `PagedUpdateCacheOp`? 

One of the tests asserts that the typecast indeed isn't applied for it, but the PR lacks explanations on why is it excluded.

dmilinkovicTT (commented):
Nice work!

The test coverage in this PR should include:
- E2E pipeline lit tests (similar to the ones in `test/ttmlir/Dialect/TTNN`
- Builder tests, to actually run the produced TTNN graphs and ensure that the outputs are correct

lib/Dialect/TTNN/Transforms/TTNNKVCacheDtypeConversion.cpp
dmilinkovicTT: Oh okay, I saw your previous response explaining this. Nevertheless, noting this tt-metal limitation explicitly in a comment would be a good idea.

Also, what if we quantize the KV cache in a graph which contains a `PagedUpdateCacheOp`? The op's input isn't quantized, so, we're expecting a mismatch here? Is that supported through tt-metal?

kdimicTT: For `PagedUpdateCacheOp` writing bfloat16 input into the bfp8 cache is handled internally. The mismatch is intentional, since casting input tensor to bfp8 would cause TT_FATAL as explained above.  I will add a comment so it is clear.

test/ttmlir/Dialect/TTNN/kv_cache_conversion/update_cache_bfp8.mlir
kdimicTT: You're right, the file was misnamed. Renamed it to paged_update_cache_bfp8.mlir and added a separate test for ttnn.update_cache.

test/ttmlir/Dialect/TTNN/kv_cache_conversion/update_cache_pipeline.mlir
```diff
@@ -0,0 +1,15 @@
...
+
+module {
+  // CHECK-LABEL: func.func @forward
+  // CHECK-SAME: %arg0: tensor<{{.*}}bfp_bf8{{.*}} {ttcore.kv_cache}
+  // CHECK-SAME: %arg1: tensor<{{.*}}bf16
+  func.func @forward(%arg0: tensor<1x32x64x512xbf16>, %arg1: tensor<1x32x1x512xbf16>) -> tensor<1x32x64x512xbf16> {
+    // CHECK-NOT: "ttnn.typecast"
+    // CHECK: "ttnn.paged_update_cache"
+    %update_index = "ttir.constant"() <{value = dense<0> : tensor<1xi32>}> : () -> tensor<1xi32>
+    %1 = "ttir.update_cache"(%arg0, %arg1, %update_index) <{batch_offset = 0 : i32}> : (tensor<1x32x64x512xbf16>, tensor<1x32x1x512xbf16>, tensor<1xi32>) -> tensor<1x32x64x512xbf16>
```
dmilinkovicTT, line 12: Should we have `PagedUpdateCache` here? For `UpdateCache`, I'm not sure why `// CHECK-NOT: "ttnn.typecast"` passes.

dmilinkovicTT (approved):

test/ttmlir/Dialect/TTNN/kv_cache_conversion/update_cache_pipeline.mlir
kdimicTT: It passes because `ttir.update_cache` is always canonicalized to `ttir.paged_update_cache` (in `TTIROps.cpp:5151`) before lowering TTIR to TTNN.

dgolubovicTT (approved):

runtime/lib/common/types.cpp
```diff
@@ -28,6 +36,17 @@ size_t TensorDesc::volume() const {
   return utils::product(shape.cbegin(), shape.cend());
 }
 
+size_t TensorDesc::sizeBytes() const {
+  if (utils::isBlockFormatDataType(dataType)) {
+    LOG_ASSERT(physicalVolume % 1024 == 0,
+               "Block format tensor physicalVolume must be a multiple of 1024 "
+               "(elements per 32x32 tile), got ",
+               physicalVolume);
+    return (physicalVolume / 1024) * utils::blockFormatTileSizeBytes(dataType);
```
pilkicTT, line 35: have we tested running KV cache with shape not divisible by tile size?

runtime/lib/common/types.cpp
```diff
@@ -19,7 +19,15 @@ TensorDesc::TensorDesc(const std::vector<uint32_t> &shape,
...
-  this->itemsize = itemsize.value_or(utils::dataTypeElementSize(dataType));
+  if (utils::isBlockFormatDataType(dataType)) {
+    if (itemsize.has_value()) {
+      LOG_WARNING("itemsize is not meaningful for block format dtypes; use "
+                  "sizeBytes() instead. Defaulting to 1.");
+    }
+    this->itemsize = 1;
+  } else {
+    this->itemsize = itemsize.value_or(utils::dataTypeElementSize(dataType));
+  }
```
pilkicTT: i don't like having this field exposed - the users now have to know the internal details to be able to use this. there are also other problems with `TensorDesc` in general, which we can fix in further changes, but for now, i would like us to look into removing this field altogether, and have one `elementSize()` getter which will assert that the format is not block format (which requires tile layout, hence the element size doesn't have much sense).

lib/Dialect/TTNN/Transforms/TTNNKVCacheDtypeConversion.cpp
```diff
@@ -0,0 +1,169 @@
...
+    // Insert typecast operations on the input operands of all cache ops so
+    // written data matches the new cache dtype. PagedUpdateCacheOp is excluded:
+    // tt-metal enforces FLOAT32/BFLOAT16 for that op and handles any dtype
+    // mismatch internally.
+    mlir::OpBuilder builder(&getContext());
+    getOperation().walk([&](func::FuncOp funcOp) {
+      if (!ttmlir::utils::isForwardDeviceFunc(funcOp)) {
+        return;
+      }
+      funcOp.walk(
```
rpavlovicTT: Simplify with
```cpp
  funcOp.walk([&](Operation *op) {
    llvm::TypeSwitch<Operation *>(op)
        .Case<FillCacheOp, UpdateCacheOp, PagedFillCacheOp>(
            [&](auto cacheOp) {
              insertTypecastIfNeeded(builder, cacheOp, dtype);
            });
  });
```

lib/Dialect/TTNN/Transforms/TTNNKVCacheDtypeConversion.cpp
```diff
@@ -0,0 +1,169 @@
...
+    // function signature.
+    getOperation().walk(
+        [&](func::FuncOp funcOp) { changeCacheArgTypes(funcOp, dtype); });
+
+    // Insert typecast operations on the input operands of all cache ops so
+    // written data matches the new cache dtype. PagedUpdateCacheOp is excluded:
+    // tt-metal enforces FLOAT32/BFLOAT16 for that op and handles any dtype
+    // mismatch internally.
+    mlir::OpBuilder builder(&getContext());
+    getOperation().walk([&](func::FuncOp funcOp) {
```
rpavlovicTT: Merge these 2 top level walks

runtime/lib/common/utils.cpp
```diff
@@ -169,6 +169,39 @@ std::uint32_t dataTypeElementSize(::tt::target::DataType dataType) {
...
+  switch (dataType) {
+  case ::tt::target::DataType::BFP_Float8:
+  case ::tt::target::DataType::BFP_BFloat8:
+    return 1088; // 1024 mantissa + 64 shared exponents
+  case ::tt::target::DataType::BFP_Float4:
+  case ::tt::target::DataType::BFP_BFloat4:
+    return 576; // 512 mantissa + 64 shared exponents
+  case ::tt::target::DataType::BFP_Float2:
+  case ::tt::target::DataType::BFP_BFloat2:
+    return 320; // 256 mantissa + 64 shared exponents
```
rpavlovicTT: It'd be nice if we derived these constants from named constants. A tile is 32×32 = 1024 elements. BFP packs elements in groups of 16 that share one 8-bit exponent, therefore 64 shared exponent bytes per tile. Mantissas take mantissaBits/8 bytes per element. So tileBytes = (1024 × mantissaBits)/8 + 64. Plug in 8/4/2 and you get 1088/576/320. Check if tt-metal exposes these constants somewhere that we can use, or define them.

rpavlovicTT (approved):

runtime/lib/common/types.cpp
kdimicTT: Yes, it was tested with shapes not divisible by tile size which passes because the compiler pads tensors  internally. The `physicalVolume % 1024 == 0` check is left as a precaution.

pilkicTT: can we add that test case to `test_kv_cache_dtype.py`?

pilkicTT (approved):

PR #7878, opened 2026-04-08
Sharding of composites

acicovicTT:
## Summary

This changeset adds capabilities for proper sharding of composites through Shardy by converting them to `stablehlo.custom_call` ops with registered sharding rules (during the StableHLO pipeline), and introduces a fusion pass that combines custom calls with surrounding CCL ops into distributed variants of those custom calls. One of the problematic scenarios that this PR is trying to solve is insertion of CCLs into the flattened graph which breaks composite reoutlining logic, and carries a perf penalty (because conversion between composites and specialized ops cannot be performed without reoutlining). The other problem this PR is trying to solve is fusion of CCLs with custom calls into distributed variants of custom calls.

Logically, the PR is split into two parts:

### Part 1: Composite-to-CustomCall conversion with sharding rules (RMS norm as example)

- **Refactored `FlattenCompositePass` to `FlattenOrConvertCompositesPass`**: The pass now has the following behavior: composites listed in `kCompositesWithCustomSharding` (shared source of truth for which composites should be converted to custom calls) are converted to `stablehlo.custom_call` ops instead of being flattened. All other composites are flattened as before (old behavior). Shardy processes custom calls as opaque / atomic ops.
- As an example of usage, a **custom sharding rule for RMS norm** composite was added in `RegisterCustomShardingRule.cpp`. Also, `CustomCallRMSNormConversionPattern` was added for custom call conversion to TTIR target `ttir::RMSNormOp`.

### Part 2: Fuse custom_calls with CCLs into distributed variants (Distributed RMS norm as example)

- **New** `FuseDistributedCustomCallsPass` with `FuseRMSNormWithCCLPattern` that matches:
  ```
  all_gather(local_input) -> gathered_input
  all_gather(local_weight) -> gathered_weight
  custom_call @tenstorrent.rms_norm(gathered) -> gathered_result
  sdy.all_slice(gathered_result) -> local_result
  ```
  And fuses into:
  ```
  custom_call @tenstorrent.distributed_rms_norm(local_input, local_weight) -> local_result
  ```
- `CustomCallDistributedRMSNormConversionPattern` that lowers the fused custom call to `ttir::DistributedRMSNormOp`.

## Design decisions and compromises

- **Conversion of composites to custom calls**. This was a deliberate decision made to make it easier to generate distributed ops without a burden on the frontend/framework layer of ensuring proper sharding and emitting a special distributed version of a composite. With this changeset, the compiler can now automatically detect fusion patterns involving CCLs and replace the composite with a distributed variant, if exists. The downside is that custom calls don't have a decomposition function to fall back on.

- **Fusion pass is part of the StableHLO pipeline** (after `ReoutlineComposite`), even though it makes target-specific decisions (e.g., knowing that `tenstorrent.distributed_rms_norm` exists and what attributes it needs). This is a deliberate compromise: the CCL patterns are recognized easier before other passes (canonicalizer, etc.) can break them apart. Attempting to recognize the same pattern later in the StableHLO-to-TTIR conversion would be fragile. It also keeps all sharding-related passes in the same place.

## Tests

- Fusion test for distributed RMS norm covers all the scenarios added by this PR, but it is a tt-xla PR that depends on this one to check-in: https://github.com/tenstorrent/tt-xla/pull/4280
- Manually ran benchmarks on galaxy and llmbox, which validated the fusion of distributed RMS norm, and also the TTNN-level decomposition because input shapes for llama are not yet aligned with expected shapes for distributed RMS norm.


lib/Conversion/StableHLOToTTIR/StableHLOLegalizeCompositePass.cpp
```diff
@@ -290,62 +349,125 @@ class TenstorrentRMSNormConversionPattern
...
+      : OpConversionPattern<mlir::stablehlo::CustomCallOp>(context) {}
 
-    auto normalizedShapeAttr = compositeAttrs.get("normalized_shape");
-    SmallVector<int64_t> normalizedShapeVec;
+  LogicalResult
+  matchAndRewrite(mlir::stablehlo::CustomCallOp srcOp,
+                  mlir::stablehlo::CustomCallOp::Adaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    if (adaptor.getCallTargetNameAttr() != kTTRMSNormCustomCallTargetName ||
+        !srcOp->hasAttr(kHasCustomShardingAttr)) {
```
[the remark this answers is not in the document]
acicovicTT: This is an invalid assesment

lib/Dialect/StableHLO/Pipelines/StableHLOPipelines.cpp
```diff
@@ -111,6 +113,10 @@ void createStableHLOPipeline(OpPassManager &pm,
   // Re-outline composite ops from flattened groups.
   pm.addPass(createReoutlineCompositePass());
 
+  // Fuse custom_calls with surrounding CCL ops into
+  // distributed variants that handle cross-device communication internally.
+  pm.addPass(createFuseDistributedCustomCallsPass());
+
```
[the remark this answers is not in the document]
acicovicTT: This will be addresses as a fusion pattern test in tt-xla

lib/Dialect/StableHLO/Transforms/FuseDistributedCustomCalls.cpp
```diff
@@ -0,0 +1,201 @@
...
+    : public OpRewritePattern<mlir::stablehlo::CustomCallOp> {
+  using OpRewritePattern::OpRewritePattern;
+
+public:
+  LogicalResult matchAndRewrite(mlir::stablehlo::CustomCallOp customCallOp,
+                                PatternRewriter &rewriter) const override {
+
+    // Only operate on custom_calls that were converted from composites with
+    // custom sharding rules.
+    if (!customCallOp->hasAttr(kHasCustomShardingAttr)) {
```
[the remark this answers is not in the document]
acicovicTT: This is an invalid assesment

lib/Dialect/StableHLO/Transforms/FuseDistributedCustomCalls.cpp
```diff
@@ -0,0 +1,201 @@
...
+      return rewriter.notifyMatchFailure(
+          customCallOp, "rms_norm operand does not come from an all_gather op");
+    }
+
+    // Check that the custom_call has exactly one user: an sdy.all_slice
+    // composite that slices the gathered result back to local shape.
+    if (!customCallOp.getResult(0).hasOneUse()) {
+      return rewriter.notifyMatchFailure(
+          customCallOp, "rms_norm result has multiple uses, cannot fuse");
+    }
```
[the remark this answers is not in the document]
acicovicTT: Addressed

lib/Dialect/StableHLO/Transforms/FuseDistributedCustomCalls.cpp
```diff
@@ -0,0 +1,201 @@
...
+    localOperands.push_back(inputAllGather.getOperand(0));
+    // Operand 1+: weight/bias, may come from all_gather or directly.
+    for (unsigned i = 1; i < customCallOp.getNumOperands(); ++i) {
+      mlir::Value operand = customCallOp.getOperand(i);
+      if (auto opAllGather =
+              operand.getDefiningOp<mlir::stablehlo::AllGatherOp>()) {
+        localOperands.push_back(opAllGather.getOperand(0));
+      } else {
+        localOperands.push_back(operand);
+      }
```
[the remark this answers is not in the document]
acicovicTT: Addressed

lib/Dialect/StableHLO/Transforms/FuseDistributedCustomCalls.cpp
```diff
@@ -0,0 +1,201 @@
...
+    }
```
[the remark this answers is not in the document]
acicovicTT: Addressed

include/ttmlir/Dialect/StableHLO/Utils/StableHLOUtils.h
```diff
@@ -29,6 +29,35 @@ inline constexpr llvm::StringLiteral kCompDecompositionKey("decomposition");
...
+// internal attribute key.
+inline constexpr llvm::StringLiteral
+    kCustomCallCompositeAttrsKey("tt.composite_attributes");
+
+// UnitAttr tag on custom_call ops that were converted from composites
+// with custom sharding rules.
+inline constexpr llvm::StringLiteral
+    kHasCustomShardingAttr("tt.has_custom_sharding");
+
+inline constexpr llvm::StringLiteral
```
mrakitaTT, line 45: Add comment header here too since you did below for `kDistributedRmsNormTargetName`

lib/Dialect/StableHLO/Transforms/FlattenOrConvertComposites.cpp
```diff
@@ -217,11 +217,53 @@ static void eraseDeadPrivateCallees(mlir::ModuleOp module) {
...
 
-// The pass that flattens all stablehlo.composite ops in the module.
-struct FlattenCompositePass
-    : public impl::FlattenCompositePassBase<FlattenCompositePass> {
-  using impl::FlattenCompositePassBase<
-      FlattenCompositePass>::FlattenCompositePassBase;
+// Returns true if the given composite name has a custom sharding rule
+// registered in kCompositesWithCustomSharding.
+static bool hasCustomShardingRule(llvm::StringRef name) {
+  for (const auto &entry : utils::kCompositesWithCustomSharding) {
```
mrakitaTT: Better not to use auto where it's not super obvious without the context what the type is ([guidelines](https://docs.tenstorrent.com/tt-mlir/coding-guidelines.html#using-auto-to-deduce-type)).

lib/Dialect/StableHLO/Transforms/FlattenOrConvertComposites.cpp
```diff
@@ -217,11 +217,53 @@ static void eraseDeadPrivateCallees(mlir::ModuleOp module) {
...
+    if (entry == name) {
+      return true;
+    }
+  }
+  return false;
```
mrakitaTT: I think this can be replaced with `return std::find(std::begin(utils::kCompositesWithCustomSharding), std::end(utils::kCompositesWithCustomSharding), name) != std::end(utils::kCompositesWithCustomSharding);`

lib/Dialect/StableHLO/Transforms/FuseDistributedCustomCalls.cpp
```diff
@@ -0,0 +1,219 @@
...
+class FuseRMSNormWithCCLPattern
+    : public OpRewritePattern<mlir::stablehlo::CustomCallOp> {
+  using OpRewritePattern::OpRewritePattern;
+
+public:
+  LogicalResult matchAndRewrite(mlir::stablehlo::CustomCallOp customCallOp,
+                                PatternRewriter &rewriter) const override {
+
+    // Only operate on custom_calls that were converted from composites with
+    // custom sharding rules.
```
mrakitaTT, line 55: Maybe we should emit some error in case we somehow end up emitting `@tenstorrent.distributed_rms_norm` directly from the frontend?

test/ttmlir/Conversion/StableHLOToTTIR/composite/test_flatten_result_pos_multi_result.mlir
mrakitaTT, line 1: I know you are adding a fusion test in a separate PR, but maybe you can also add some smaller lit tests for parts of this logic? For example for composite->custom_call conversion.

mrakitaTT (approved):

include/ttmlir/Dialect/StableHLO/Utils/StableHLOUtils.h
acicovicTT: Done

lib/Dialect/StableHLO/Transforms/FlattenOrConvertComposites.cpp
acicovicTT: Done, found llvm::is_contained

test/ttmlir/Conversion/StableHLOToTTIR/composite/test_flatten_result_pos_multi_result.mlir
acicovicTT: Done

lib/Dialect/StableHLO/Transforms/FlattenOrConvertComposites.cpp
acicovicTT: Not relevant anymore, but I agree

lib/Dialect/StableHLO/Transforms/FuseDistributedCustomCalls.cpp
acicovicTT: I thought about it and decided to just emit failure, as is suggested for obvious failures to match a pattern. notifyMatchFailure is suggested when further clarification is needed, which I don't think is a case here

Issue #7832, opened 2026-04-06
Complex Conversion pattern can't enter sdy.manual_computation regions


ppadjinTT:
### Describe the bug

When we perform sharding on a torch module and also perform complex tensor arithmetic, the complex arithmetics become part of the `sdy.manual_computation` block.  However, `StablehloComplexMathExpanderPass` fires before `ConvertStableHLOToTTIRPass` (where `sdy.manual_computation` becomes inlined) and therefore we cannot rewrite the complex arithmetic ops inside the `sdy.manual_computation` block. This leaves complex arithmetic ops which can't (and shouldn't) be legalized.

This is the symptom error message of this problem:
```
error: failed to legalize unresolved materialization from ('tensor<16x16x2xf32>') to ('tensor<16x16xcomplex<f32>>') that remained live after conversion
```
Attaching whole log here:

[repro.log](https://github.com/user-attachments/files/26538303/repro.log)

### tt-xla repro:
```
import os

import numpy as np
import torch
import torch.nn as nn
import torch_xla
import torch_xla.distributed.spmd as xs
import torch_xla.runtime as xr
from torch_xla.distributed.spmd import Mesh

DIM = 128
NUM_HEADS = 4
HEAD_DIM = DIM // NUM_HEADS  # 32
SEQ_LEN = 16
SEED = 42
sharding = False


xr.set_device_type("TT")
torch_xla.set_custom_compile_options({"optimization_level": 1})

if sharding:
    os.environ["CONVERT_SHLO_TO_SHARDY"] = "1"
    xr.use_spmd()
    num_devices = xr.global_runtime_device_count()
    device_ids = np.array(range(num_devices))
    mesh = Mesh(device_ids, (1, num_devices), ("batch", "model"))
    print(f"SPMD: {num_devices} devices, mesh (1, {num_devices})")


class ComplexRoPELinear(nn.Module):
    def __init__(self, dim, num_heads, seq_len):
        super().__init__()
        self.num_heads = num_heads
        self.head_dim = dim // num_heads
        self.proj = nn.Linear(dim, dim, bias=False)

        half_dim = self.head_dim // 2
        theta = 10000.0
        freq_indices = torch.arange(0, half_dim, dtype=torch.float32)
        inv_freq = 1.0 / torch.pow(theta, freq_indices / half_dim)
        positions = torch.arange(seq_len, dtype=torch.float32)
        angles = torch.outer(positions, inv_freq)
        
[...]

PR #7884, opened 2026-04-08
Allowing `StableHLOComplexDataTypeConversionPass` to enter `sdy.manual_computation` regions. 

ppadjinTT:
### Ticket
https://github.com/tenstorrent/tt-mlir/issues/7832

### Problem description
This PR https://github.com/tenstorrent/tt-mlir/pull/7407 introduced support for operations with complex arguments. Currently, the `StableHLOComplexDataTypeConversionPass` which handles conversion of individual ops with complex tensor to equivalent operations on real tensors cannot enter regions marked with `sdy.manual_computation`. Issue above highlights one such scenario. 

### Why this is the case? 

We do not have a pattern that catches `sdy.manual_computation` op, so the block argument of `sdy.manual_computation` remains complex type. This means that later when we try to match complex patterns for ops inside this region, these ops will have complex arguments (becase the block argument is complex). 

The problem is that for Complex rewrite patterns to kick in (for example patterns for stablehlo.real, stablehlo.imag, ...), the arguments are expected to be real tensors already. Since this is not the case, we can't match any pattern and we are left with illegal complex tensors. 

### What's changed
- added ShardyManualComputationComplexConversionPattern which
       a) rewrites block arguments from complex-> real with trailing dim 2, which allows the complex patterns to kick in for ops inside the `sdy.manual_computation` region.
       b) properly adds another `{}` empty sharding dim for the additional real dimension that is added (we always wanna replicate that dim) 

- added llvm test for this pattern

### Checklist
- [x] tt-xla model test passing: https://github.com/tenstorrent/tt-xla/actions/runs/25075296404

### DIFF
--- a/lib/Dialect/StableHLO/Transforms/ComplexDataTypeConversion.cpp
+++ b/lib/Dialect/StableHLO/Transforms/ComplexDataTypeConversion.cpp
@@ -447,8 +447,20 @@ struct StableHLOComplexDataTypeConversionPass
         mlir::stablehlo::BroadcastInDimOp>(isNotComplexType);
 
     target.addIllegalOp<mlir::stablehlo::ComplexOp, mlir::stablehlo::RealOp,
-                        mlir::stablehlo::ImagOp,
-                        mlir::sdy::ManualComputationOp>();
+                        mlir::stablehlo::ImagOp>();
+
+    auto hasComplexType = [](TypeRange types) {
+      return llvm::any_of(types, [](Type t) {
+        auto rtt = mlir::dyn_cast<RankedTensorType>(t);
+        return rtt && mlir::isa<mlir::ComplexType>(rtt.getElementType());
+      });
+    };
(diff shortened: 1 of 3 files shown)


ddilbazTT (approved):

lib/Dialect/StableHLO/Transforms/ComplexDataTypeConversion.cpp
```diff
@@ -349,6 +454,13 @@ struct StableHLOComplexDataTypeConversionPass
     target.addIllegalOp<mlir::stablehlo::ComplexOp, mlir::stablehlo::RealOp,
                         mlir::stablehlo::ImagOp>();
 
+    target.addDynamicallyLegalOp<mlir::sdy::ManualComputationOp>(
```
ddilbazTT, line 458: I don't think we should check `hasComplexElementType`. There is a chance the complex element type is created inside the manual computation and then used inside it, never being one of the arguments or the return types.

sdjukicTT (approved):
LGTM

PR #7919, opened 2026-04-09
split insert dst register pass

### LINKED ISSUE #8345: [D2M][Dst Register] Refactor `insertDstRegisterAccess` into two passes
Refactor for scheduled vs non-scheduled.

sgholamiTT:
### Motivation
https://github.com/tenstorrent/tt-mlir/issues/8345
The monolithic `D2MInsertDstRegisterAccess` pass (~2050 lines) handled two fundamentally different DST register access patterns in a single file:

- **Unscheduled** (matmul, reduction): uses a bump allocator, generates three separate loop nests (load, compute, store), handles L1 accumulation guards.
- **Scheduled** (element-wise / SFPU): uses a stack allocator, performs in-place DST rewrites with a cloned store loop, handles the `d2m.scheduled` attribute produced by `OpScheduler`, and owns the fallback path for regions where the root loop has been canonicalized away.

Mixing both strategies in one pass made reasoning about correctness harder, obscured the division of responsibility, and made it difficult to test each path in isolation.

### What changed

The single `InsertDstRegisterAccess.cpp` is replaced by a shared header + three implementation files:

| File | Role |
|------|------|
| `InsertDstRegisterAccessShared.{h,cpp}` | Shared types (`CopyInfoMap`, `DstIntermediatesMap`, `LoadStoreRecord`, …) and utility functions (`recordDstAccess`, `buildIndices`, `insertAcquireDst`, `cloneAffineLoopSkeleton`, `emitDstCopyNest`, `insertDstRegisterAccessFinalize`, …) used by both passes. |
| `InsertDstRegisterAccessUnscheduled.cpp` | Handles roots **without** `d2m.scheduled`. Bump allocator (`DstSliceAllocationState`, declared locally), three-nest data-copy generation. Explicitly skips scheduled roots. |
| `InsertDstRegisterAccessScheduled.cpp` | Handles roots **with** `d2m.scheduled` (consumes the attribute). Stack allocator (`DstStackAllocator`, declared locally), in-place + cloned data-copy generation, `memref.load`→`memref.store` copy patterns, and the no-`linalg_root` fallback. Explicitly skips unscheduled roots. |

The pipeline in `D2MPipelines.cpp` now invokes the unscheduled pass first, then the scheduled pass—matching the order in which these patterns appear in the lowering flow.

`Passes.td` defines the two new passes with separate names (`d2m-insert-dst-register-access-unscheduled` / `d2m-insert-dst-register-access-scheduled`) and documents their contracts (including example IR for each).

### Design principles

- **Zero copy-paste**: every shared function lives in `*Shared.*`. Each per-pass file contains only the collection and emission logic that genuinely diverges; the cloned-nest copy emitter is a single shared template (`emitDstCopyNest`) used by both.
- **Explicit contracts**: each pass skips loops outside its territory (the unscheduled pass skips loops carrying `d2m.scheduled`; the scheduled pass skips `d2m.linalg_root` loops that don't), and both call `verifyInsertDstRegisterAccessPreconditions` (walks the module for any remaining `linalg.generic` and signals pass failure if so).
- **Attribute handshake**: `d2m.scheduled` is the discriminator. The unscheduled pass skips loops that carry it; the scheduled pass requires it (or the fallback condition) and removes it after processing.
- **Deterministic emission**: shared maps use `llvm::MapVector` so iteration order is the deterministic insertion order, keeping IR output stable run-to-run.

### Test changes

The 14 existing `insert_dst_register_access_*.mlir` tests were updated to invoke only the pass(es) actually needed. Since these tests do not run `OpScheduler`, none of their IR carries `d2m.scheduled`, so they now invoke only `--d2m-insert-dst-register-access-unscheduled`.

Seven new test files were added:

| Test | Exercises |
|------|-----------|
| `*_scheduled_eltwise.mlir` | Scheduled path: unary, binary chain, intermediate results through DST, stack allocator, `d2m.scheduled` consumption. |
| `*_scheduled_fallback.mlir` | Fallback path: flat compute block with no `d2m.linalg_root` loop. |
| `*_scheduled_end_to_end.mlir` | Full pipeline: `linalg-to-affine` → `op-scheduler` → both DST passes. Verifies `d2m.scheduled` is produced and consumed end-to-end. |
| `*_scheduled_memref_copy.mlir` | `memref.load` → `memref.store` copy pattern (DecomposeMasking-like) alongside compute ops. |
| `*_scheduled_skips_unscheduled_roots.mlir` | Scheduled pass correctly ignores roots without `d2m.scheduled`. |
| `*_unscheduled_skips_scheduled_roots.mlir` | Unscheduled pass correctly ignores roots with `d2m.scheduled`. |
| `*_l1_acc.mlir` | L1 accumulation guard (`enable-l1-acc=true`): matmul with multi-iteration `scf.for` tiling, verifies `d2m.set_l1_accumulate` insertion. |

The shared `--d2m-linalg-to-affine` precondition is exercised in the existing `*_missing_linalg_to_affine.mlir`, which now has two `RUN:` lines (one per split pass) over the same IR.

- [x] All D2M lit tests pass (`test/ttmlir/Dialect/D2M/`: 126/126).
- [x] New / existing tests cover the split routing, both allocators, fallback, end-to-end through the op scheduler, L1 accumulation, and the precondition check.


nsmithtt:
@dlokeTT / @wenbinlyuTT can you review?

dlokeTT (approved):
Looks good to me as a first step. I think it would be good to unify these again later on when we don't have a concept of scheduled/unscheduled. I'm not sure of those concepts/op scheduler are as valid anymore after Jacobs refactor of fusion

sgholamiTT:
Agree @dlokeTT. These passes will be simplified much more in later PRs. 
As the first step, I need to distinguish them clearly to make later refactoring easier.

include/ttmlir/Dialect/D2M/Transforms/InsertDstRegisterAccessShared.h
```diff
@@ -0,0 +1,257 @@
...
+  bool didStoreToDst() { return storedToDst; }
+  int getCurrSliceIndex() { return nextSliceIndex - 1; }
+
+private:
+  int64_t nextSliceIndex = 0;
+  bool storedToDst = false;
+};
+
+// Stack-based allocator -- used by the scheduled path.
+class DstStackAllocator {
```
wenbinlyuTT: Nit: maybe this should be moved to the scheduled pass?

wenbinlyuTT (approved):
Overall LGTM, makes it easier to reason about DST-related changes!

nsmithtt (approved):
Signing off on cmake changes + high level refactor

include/ttmlir/Dialect/D2M/Transforms/Passes.td
```diff
@@ -254,70 +254,58 @@ def D2MLinalgToAffine : Pass<"d2m-linalg-to-affine", "::mlir::ModuleOp"> {
...
-        }
-      }
-      affine.for %arg3 = 0 to 3 {
-        affine.for %arg4 = 0 to 2 {
-          %1 = affine.load %dst[%arg3, %arg4] : memref<3x2x!tt.tile<32x32, f32>, #dst>
-          affine.store %1, %cb2[%arg3, %arg4] : memref<3x2x!tt.tile<32x32, f32>, #l1>
-        }
-      }
-    ```
+    Requires prior `--d2m-linalg-to-affine` (no `linalg.generic` in compute
```
vwellsTT: broadly, I think getting rid of example IR is a downgrade here :)

include/ttmlir/Dialect/D2M/Transforms/Passes.td
```diff
@@ -254,70 +254,58 @@ def D2MLinalgToAffine : Pass<"d2m-linalg-to-affine", "::mlir::ModuleOp"> {
...
+    Also handles the **fallback** when a region has compute but no
+    `d2m.linalg_root` loop at region entry (e.g. single-iteration loop folded
+    away): treats the region as scheduled-shaped and inserts DST access for
+    the flat block.
+
+    Run **after** `d2m-insert-dst-register-access-unscheduled` when both passes
+    are needed.
+
+    Requires prior `--d2m-linalg-to-affine` (no `linalg.generic` in compute
+    regions).
```
vwellsTT: I wonder if there's a graceful way to encode this sort of pre-condition (and debatably, above ordering between these 2 passes as well).  Perhaps we could have a simple `.walk` w/ asserts that there aren't any linalg.generic ops?

lib/Dialect/D2M/Transforms/InsertDstRegisterAccessScheduled.cpp
```diff
@@ -0,0 +1,556 @@
+// SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
```
vwellsTT: Would it make sense to put these files in a shared subdir?  I personally think it's better for grouping multiple files, but could see arguments the other way too

lib/Dialect/D2M/Transforms/InsertDstRegisterAccessScheduled.cpp
```diff
@@ -0,0 +1,556 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#include "ttmlir/Dialect/D2M/Transforms/InsertDstRegisterAccessShared.h"
```
vwellsTT: imo this should go in below block, though it's a bit of a unusual case I guess

lib/Dialect/D2M/Transforms/InsertDstRegisterAccessShared.cpp
```diff
@@ -0,0 +1,1023 @@
...
+  LDBG() << outputStackStr;
+
+  return currSliceIndex;
+}
+
+unsigned DstStackAllocator::deallocate() {
+  assert(!(inputStack.empty() && outputQueue.empty()) &&
+         "Deallocating non-existent dst slice");
+
+  unsigned id;
```
vwellsTT: nit: uninit kind of "smells" even though all branches are covered, imo better to flatten out a layer of loops to default init to case 1, then branch other 2 cases and potentially overwrite.  Other options: use ternary (w/ some restructuring this is good) or lambda

lib/Dialect/D2M/Transforms/InsertDstRegisterAccessShared.cpp
```diff
@@ -0,0 +1,1023 @@
...
+
+  if (!inputStack.empty()) {
+    id = inputStack.pop_back_val();
+  } else {
+    if (outputQueue.size() > 1) {
+      id = outputQueue.at(outputQueue.size() - 2);
```
vwellsTT: accessing second to last elt of a "queue" is fundamentally goofy thing to do; I think this code needs algorithmic/DS rethinking (not sure I get the intent tbh)

lib/Dialect/D2M/Transforms/InsertDstRegisterAccessShared.cpp
```diff
@@ -0,0 +1,1023 @@
...
+      outputQueue.erase(outputQueue.end() - 2);
+    } else {
+      id = outputQueue.back();
+      outputQueue.pop_back();
+    }
+  }
+
+  sliceStack.push_back(id);
+
+  LDBG() << "======== DEALLOCATE =========";
```
vwellsTT: we seem to have mixed philosophies on usage of LDBG but personally I find having this much in blocks here pretty distracting from the rest of the code, not sure what others think

lib/Dialect/D2M/Transforms/InsertDstRegisterAccessShared.cpp
```diff
@@ -0,0 +1,1023 @@
...
+
+  std::string sliceStackStr = "SliceStack = ";
+  for (auto it : sliceStack) {
+    sliceStackStr += std::to_string(it) + ",";
+  }
+  LDBG() << sliceStackStr;
+
+  std::string inputStackStr = "InputStack = ";
```
vwellsTT: throughout PR, pls remove this type of pattern and use llvm's utils instead (`llvm::interleaveComma` or something like that).  This is much uglier, potentially bad perf (might be aggregated w/o usaged in non-debug mode), and uses non-llvm types w/o need

lib/Dialect/D2M/Transforms/InsertDstRegisterAccessShared.cpp
```diff
@@ -0,0 +1,1023 @@
...
+  SmallVector<Value> guardIVs;
+  for (Value loopIV : collectAncestorLoopIVs(contextOp)) {
+    if (!accessDependsOnIV(loadOrStore, loopIV)) {
+      guardIVs.push_back(loopIV);
+    }
+  }
+  return guardIVs;
+}
+
+bool hasTileMatmul(Operation *op) {
```
vwellsTT, line 174: nit: can write this marginally more compactly as:
```
bool hasTileMatmul(Operation *op) {
  return op->walk([](d2m::TileMatmulOp) {
    return WalkResult::interrupt();
  }).wasInterrupted();
}
```

lib/Dialect/D2M/Transforms/InsertDstRegisterAccessShared.cpp
```diff
@@ -0,0 +1,1023 @@
...
+
+  SmallVector<Value> guardIVs;
+  if (assocCB && emitGuard) {
+    guardIVs = getGuardLoopIVs(loadOrStore, outermostInnerComputeLoop);
+  }
+
+  iter->second.record(loadOrStore, dstSlice, guardIVs);
+}
+
+void recordDstAccess(affine::AffineLoadOp loadOrStore, CopyInfoMap &copyInfos,
```
vwellsTT: nit: loadOrStore seems like a silly name for all of these when it's concretely one or the other? I guess the idea is that's what the impl calls them? maybe fine idk

lib/Dialect/D2M/Transforms/InsertDstRegisterAccessShared.cpp
```diff
@@ -0,0 +1,1023 @@
...
+          .getResult();
+
+  const auto cmpPredicate =
+      isBcastGuard ? arith::CmpIPredicate::eq : arith::CmpIPredicate::ne;
+
+  auto zero = rewriter.create<arith::ConstantOp>(
+      loc, rewriter.getIndexType(),
+      rewriter.getIntegerAttr(rewriter.getIndexType(), 0));
+
+  for (Value guardIV : guardIVs) {
```
vwellsTT, line 764: just overwriting based on last guardIV seen doesn't seem intended/correct?

lib/Dialect/D2M/Transforms/InsertDstRegisterAccessUnscheduled.cpp
```diff
@@ -0,0 +1,522 @@
...
+      if (computeOp.isScalarOperand(operandIdx)) {
+        continue;
+      }
+      auto potentialLoad = computeOp->getOperand(operandIdx)
+                               .getDefiningOp<affine::AffineLoadOp>();
+      if (potentialLoad && notDstMemspace(potentialLoad)) {
+        ++totalCBLoads;
+      }
+    }
+    const bool noAccumGuardForLoads = totalCBLoads >= 2;
```
vwellsTT, line 98: significance of 2 is not clear to me, probably good to add a comment

lib/Dialect/D2M/Transforms/InsertDstRegisterAccessUnscheduled.cpp
```diff
@@ -0,0 +1,522 @@
...
+        assert(!dstSliceAllocationState.didStoreToDst() &&
+               "Multiple stores from last op to dst not supported");
+
+        const bool rhsIsScalar = computeOp.isScalarOperand(1);
+
+        int dstSlice = -1;
+        if (dstRegInPlace || rhsIsScalar) {
+          bool isUnaryOp = computeOp->getNumOperands() == 1;
+          bool isTileMatmul = mlir::isa<d2m::TileMatmulOp>(computeOp);
+          bool isReduction = mlir::isa<d2m::TileReduceMaxOp>(computeOp) ||
```
vwellsTT: nit: use varidiac `isa<X, Y, Z>` instead

lib/Dialect/D2M/Transforms/InsertDstRegisterAccessUnscheduled.cpp
```diff
@@ -0,0 +1,522 @@
...
```
vwellsTT: same here

lib/Dialect/D2M/Transforms/InsertDstRegisterAccessUnscheduled.cpp
```diff
@@ -0,0 +1,522 @@
...
+          dstSliceAllocationState.setStoreToDst();
+        }
+        collectDstStoreAccess(scratchStore, copyInfos, dstSlice,
+                              outermostInnerComputeLoop);
+      } else {
+        assert(user->hasTrait<D2MGenericRegionComputeOpTrait>());
+        assert(computeOp->hasOneUse() &&
+               "Currently we do not support multiple "
+               "users in the same compute dst region.");
+        assert(computeOp->getNumResults() == 1);
```
vwellsTT: might be good to have equivalent message here

lib/Dialect/D2M/Transforms/InsertDstRegisterAccessUnscheduled.cpp
```diff
@@ -0,0 +1,522 @@
...
+          dstSlice = firstInputDstSlice;
+          dstSliceAllocationState.setStoreToDst();
+        } else {
+          dstSlice = dstSliceAllocationState.allocate();
+          dstSliceAllocationState.setStoreToDst();
+        }
+        collectDstStoreAccess(scratchStore, copyInfos, dstSlice,
+                              outermostInnerComputeLoop);
+      } else {
+        assert(user->hasTrait<D2MGenericRegionComputeOpTrait>());
```
vwellsTT: I think better to use TT_assert everywhere vs mixing assert and TT_assert, esp. in same file

lib/Dialect/D2M/Transforms/InsertDstRegisterAccessShared.cpp
vwellsTT: (This would also make LDBG blocks 1 line instead of 5+, which would help with my above concerns, btw)

include/ttmlir/Dialect/D2M/Transforms/InsertDstRegisterAccessShared.h
```diff
@@ -0,0 +1,257 @@
+// SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
```
anuragsingh-tt: Should 2025 be updated?

include/ttmlir/Dialect/D2M/Transforms/InsertDstRegisterAccessShared.h
```diff
@@ -0,0 +1,257 @@
...
+  SmallVector<Value> guardIVs = {};
+
+  LoadStoreRecord(LoadOrStoreTy loadStore,
+                  std::optional<d2m::TileBcastOp> bcast, int dstSlice,
+                  ArrayRef<Value> guardIVs)
+      : loadStore(loadStore), bcast(bcast), dstSlice(dstSlice),
+        guardIVs(guardIVs.begin(), guardIVs.end()) {}
+};
+
+struct CopyInfo {
```
anuragsingh-tt, line 60: I am not sure five parallel containers makes sense here. Maybe we should have a DstAccess struct that contains an enum for the kind of Dst Access, e.g. Load/Store and have helpers (replaceLoadWithDst/replaceStoreWithDst) ingest this DstAccess struct? That way the logic is centralized and not repeated.

include/ttmlir/Dialect/D2M/Transforms/InsertDstRegisterAccessShared.h
```diff
@@ -0,0 +1,257 @@
...
+#define TTMLIR_DIALECT_D2M_TRANSFORMS_INSERTDSTREGISTERACCESSSHARED_H
+
+#include "ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.h"
+
+#include "mlir/Dialect/Affine/IR/AffineOps.h"
+#include "mlir/Dialect/MemRef/IR/MemRef.h"
+#include "mlir/Dialect/SCF/IR/SCF.h"
+#include "mlir/IR/BuiltinOps.h"
+#include "mlir/IR/PatternMatch.h"
+#include "mlir/Interfaces/DestinationStyleOpInterface.h"
```
anuragsingh-tt: Are these imports actually used? E.g. DPS?

include/ttmlir/Dialect/D2M/Transforms/InsertDstRegisterAccessShared.h
```diff
@@ -0,0 +1,257 @@
...
+#include "llvm/ADT/DenseMap.h"
+#include "llvm/ADT/STLExtras.h"
+#include "llvm/ADT/SmallVector.h"
+
+#include <deque>
+#include <optional>
+
+namespace mlir::tt::d2m {
+
+class GenericOp;
```
anuragsingh-tt: Why forward declare?

include/ttmlir/Dialect/D2M/Transforms/InsertDstRegisterAccessShared.h
anuragsingh-tt: This might be better in a follow up PR as it is a nontrivial change. Won't block on this one.

include/ttmlir/Dialect/D2M/Transforms/InsertDstRegisterAccessShared.h
```diff
@@ -0,0 +1,257 @@
...
+
+LogicalResult verifyInsertDstRegisterAccessPreconditions(ModuleOp moduleOp);
+
+// ---------------------------------------------------------------------------
+// Shared data structures
+// ---------------------------------------------------------------------------
+
+namespace detail {
+
+struct OperationTypes {
```
anuragsingh-tt: Can we change this name to be more specific to Dst?

include/ttmlir/Dialect/D2M/Transforms/InsertDstRegisterAccessShared.h
```diff
@@ -0,0 +1,257 @@
...
+    memrefStores.emplace_back(store, std::nullopt, dstSlice, ArrayRef<Value>{});
+  }
+
+  SmallVector<LoadStoreRecord<affine::AffineLoadOp>> loads;
+  SmallVector<LoadStoreRecord<affine::AffineStoreOp>> stores;
+  SmallVector<LoadStoreRecord<memref::LoadOp>> memrefLoads;
+  SmallVector<LoadStoreRecord<memref::StoreOp>> memrefStores;
+};
+
+using CopyInfoMap = DenseMap<Operation *, CopyInfo>;
```
anuragsingh-tt: Is this deterministic?

lib/Dialect/D2M/Transforms/InsertDstRegisterAccessScheduled.cpp
```diff
@@ -0,0 +1,556 @@
...
+        loadStoreDstAccessGenerator,
+    llvm::function_ref<void(PatternRewriter &, LoadStoreOpTy, AffineMap,
+                            ValueRange)>
+        dstAccessReplacement,
+    bool enableL1Acc = false) {
+  if (loadStoreOps.empty()) {
+    return;
+  }
+
+  for (auto [loadStore, bcast, dstSliceIndex, guardIVs] : loadStoreOps) {
```
anuragsingh-tt, line 188: Can we use `const auto&` here?

lib/Dialect/D2M/Transforms/InsertDstRegisterAccessScheduled.cpp
```diff
@@ -0,0 +1,556 @@
...
+    rewriter.clone(*loopNestOrOp, irMapper)->walk([&](Operation *op) {
+      if (!mlir::isa<affine::AffineForOp, affine::AffineYieldOp,
+                     affine::AffineApplyOp>(op)) {
+        op->dropAllUses();
+        rewriter.eraseOp(op);
+      }
+    });
+  }
+
+  for (auto [loadStore, bcast, dstSliceIndex, guardIVs] : loadStoreOps) {
```
anuragsingh-tt: Same as above for all loops in this file.

lib/Dialect/D2M/Transforms/InsertDstRegisterAccessShared.cpp
```diff
@@ -0,0 +1,1023 @@
+// SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
```
anuragsingh-tt: Update for all files?

lib/Dialect/D2M/Transforms/InsertDstRegisterAccessUnscheduled.cpp
anuragsingh-tt, line 1: There seems to be logic that is duplicated across these files which ideally should be refactored such that they use the same codepath. e.g. cloneLoopSkeleton should be called across both files instead of having separate clone and skeleton logic across the two files.

include/ttmlir/Dialect/D2M/Transforms/InsertDstRegisterAccessShared.h
```diff
@@ -0,0 +1,257 @@
...
+
+void collectDstStoreAccess(affine::AffineStoreOp storeOp,
+                           CopyInfoMap &copyInfos, int dstSlice,
+                           Operation *outermostInnerComputeLoop);
+void collectDstStoreAccess(memref::StoreOp storeOp, CopyInfoMap &copyInfos,
+                           int dstSlice, Operation *outermostInnerComputeLoop);
+
+void collectDstLoadWithAccumAnalysis(affine::AffineLoadOp loadOp,
+                                     int64_t operandIdx,
+                                     ArrayRef<Value> carriedOutputRegions,
```
anuragsingh-tt: ValueRange instead?

test/ttmlir/Dialect/D2M/Transforms/insert_dst_register_access_l1_acc.mlir
```diff
@@ -0,0 +1,77 @@
+// SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
+//
+// SPDX-License-Identifier: Apache-2.0
+
+// Tests for L1 accumulation guard insertion in the unscheduled pass.
+// When enable-l1-acc=true and the root loop contains a tile_matmul,
+// the pass inserts a d2m.set_l1_accumulate guarded by an scf.if that
+// checks whether the outer tiling loop has reached its second iteration.
+//
+// The test uses hand-crafted IR with a matmul (4x4 * 4x2 -> 4x2) tiled
```
anuragsingh-tt: Is this comment really necessary?

lib/Dialect/D2M/Transforms/InsertDstRegisterAccessUnscheduled.cpp
```diff
@@ -0,0 +1,522 @@
...
+        rewriter.replaceOp(record.loadStore, dstLoad.getResult());
+      }
+    };
+
+    createCopyLoop<affine::AffineLoadOp>(rewriter, loopNestOrOp, copyInfo.loads,
+                                         loadAccessGenerator,
+                                         loadAccessRewriter,
+                                         /*enableL1Acc=*/enableL1Acc);
+
+    // Step 2: store copy loop.
```
vwellsTT, line 276: nit: function comment says "Generates 3 loop nests" but we only have steps for 2 of them here--perhaps compute loop is pre-existing? if so top-level comment should be re-worded

lib/Dialect/D2M/Transforms/InsertDstRegisterAccessUnscheduled.cpp
```diff
@@ -0,0 +1,522 @@
...
+  });
+  return {copyInfos, dstIntermediates};
+}
+
+// ---------------------------------------------------------------------------
+// Unscheduled-only: copy loop generation (clones loop skeletons)
+// ---------------------------------------------------------------------------
+
+template <typename LoadOrStoreTy>
+static void createCopyLoop(
```
vwellsTT: this function could use some comments, I find the logic difficult to follow with 3 lambdas etc

vwellsTT: I also wonder if there's an easier way to commonize the code in general vs this lambda-based system

lib/Dialect/D2M/Transforms/InsertDstRegisterAccessUnscheduled.cpp
```diff
@@ -0,0 +1,522 @@
...
+            rewriter, gOp, *loopRegion, dstCapacity, loopOp, packerL1Acc,
+            copyInfos, dstIntermediates,
+            [](PatternRewriter &rw, Location loc, Value dst,
+               const CopyInfoMap &ci,
+               bool l1Acc) { dataCopyGenerate(rw, loc, dst, ci, l1Acc); });
+
+        return WalkResult::advance();
+      });
+
+      // The unscheduled pass does NOT handle the "no linalg_root" fallback.
```
vwellsTT: imo this comment isn't terribly meaningful on its own like this, probably would remove

test/ttmlir/Dialect/D2M/Transforms/insert_dst_register_access_unscheduled_skips_scheduled_roots.mlir
```diff
@@ -0,0 +1,42 @@
+// SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
```
vwellsTT: as a general note, we end with a lot of insert_dst mlir tests, are they actually testing unique cases? Would commonizing some of them be more convenient?

vwellsTT (commented):
I think the core refactor idea looks sound to me, but I think the code could use some polishing before merging

xanderchin (approved):

lib/Dialect/D2M/Transforms/InsertDstRegisterAccessScheduled.cpp
```diff
@@ -0,0 +1,561 @@
...
+          if (valueToStore.getType() != cbType.getElementType()) {
+            valueToStore = rewriter
+                               .create<d2m::DstReinterpretCastOp>(
+                                   loc, cbType.getElementType(), valueToStore)
+                               .getResult();
+          }
+
+          rewriter.create<affine::AffineStoreOp>(
+              loc, dstLoad.getResult(), cb, l1AccessMap, l1AccessIndices);
+        },
```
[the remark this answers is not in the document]
sgholamiTT: Fixed — using the casted valueToStore now.

lib/Dialect/D2M/Transforms/InsertDstRegisterAccessScheduled.cpp
```diff
@@ -0,0 +1,561 @@
...
+          if (valueToStore.getType() != dstType.getElementType()) {
+            valueToStore =
+                rewriter
+                    .create<d2m::DstReinterpretCastOp>(
+                        op.getLoc(), dstType.getElementType(), valueToStore)
+                    .getResult();
+          }
+
+          rewriter.replaceOpWithNewOp<affine::AffineStoreOp>(
+              op, op.getValue(), dst, dstAccessMap, dstAccessIndices);
```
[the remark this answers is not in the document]
sgholamiTT: Fixed — replaceOpWithNewOp now uses the casted valueToStore.

lib/Dialect/D2M/Transforms/InsertDstRegisterAccessUnscheduled.cpp
```diff
@@ -0,0 +1,522 @@
...
+        assert(!dstSliceAllocationState.didStoreToDst() &&
+               "Multiple stores from last op to dst not supported");
+
+        const bool rhsIsScalar = computeOp.isScalarOperand(1);
+
+        int dstSlice = -1;
+        if (dstRegInPlace || rhsIsScalar) {
+          bool isUnaryOp = computeOp->getNumOperands() == 1;
+          bool isTileMatmul = mlir::isa<d2m::TileMatmulOp>(computeOp);
+          bool isReduction = mlir::isa<d2m::TileReduceMaxOp>(computeOp) ||
```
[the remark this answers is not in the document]
sgholamiTT: Fixed — `TileReduceMeanOp` added; the check now matches the affine-store branch above.

include/ttmlir/Dialect/D2M/Transforms/InsertDstRegisterAccessShared.h
sgholamiTT: Done — moved `DstStackAllocator` to `InsertDstRegisterAccessScheduled.cpp`'s anonymous namespace, and the bump allocator to `Unscheduled.cpp` for symmetry.

include/ttmlir/Dialect/D2M/Transforms/Passes.td
sgholamiTT: Agreed — added an example IR for each pass.

include/ttmlir/Dialect/D2M/Transforms/Passes.td
sgholamiTT: This is already enforced — both passes call `verifyInsertDstRegisterAccessPreconditions(moduleOp)` from `runOnOperation`, which walks for any remaining `linalg.generic` and `signalPassFailure()`s. Tightened the wording in `Passes.td` to call this out.

lib/Dialect/D2M/Transforms/InsertDstRegisterAccessScheduled.cpp
sgholamiTT: Hmm, left as-is for now — the convention in `lib/Dialect/D2M/Transforms/` is to group related files by name prefix (`InsertDstRegisterAccess*`, `Generic*`, `LowerDMA*`); making this one a subdir would be a one-off I guess. 
Open to it as a project-wide cleanup though (?).

lib/Dialect/D2M/Transforms/InsertDstRegisterAccessScheduled.cpp
sgholamiTT: Fixed.

lib/Dialect/D2M/Transforms/InsertDstRegisterAccessShared.cpp
sgholamiTT: Fixed — `id` default-initialized to 0 then overwritten in each ternary branch (if that's what you mean).

lib/Dialect/D2M/Transforms/InsertDstRegisterAccessShared.cpp
sgholamiTT: Agreed it's awkward. I'd say this is out of scope for this PR (since this PR is just a refactoring and this was an existing logic). 
Added a comment explaining the intent (tail of `outputQueue` is the running output for the next consumer, so we drop the second-from-tail) and a TODO to revisit when the allocator is reworked.

lib/Dialect/D2M/Transforms/InsertDstRegisterAccessShared.cpp
sgholamiTT: Done — collapsed the four `LDBG` blocks into a single `debugDumpDstStackAllocator` helper using `LDBG_OS + llvm::interleaveComma`. Each call site is one line now.

lib/Dialect/D2M/Transforms/InsertDstRegisterAccessShared.cpp
sgholamiTT: Done — manual loops replaced with `llvm::interleaveComma` inside an `LDBG_OS` callback so they're never aggregated when the debug type is off.

lib/Dialect/D2M/Transforms/InsertDstRegisterAccessShared.cpp
sgholamiTT: Done.

lib/Dialect/D2M/Transforms/InsertDstRegisterAccessShared.cpp
sgholamiTT: Renamed the parameter to `op` in the four public overloads. The templated `recordDstAccessImpl` stays as `LoadOrStoreTy` since it really is generic.

lib/Dialect/D2M/Transforms/InsertDstRegisterAccessShared.cpp
sgholamiTT: Hmm that's not what's being done here. I think this one might read as overwrite at first glance but it's actually a left-fold: 
`guard = guard {AND|OR} cmp_i` for each IV:
- `AND` for bcast init guards starting from `true`, 
- `OR` for accumulation guards starting from `false`. 

Added a comment in createLoadLoopGuard spelling that out.

lib/Dialect/D2M/Transforms/InsertDstRegisterAccessUnscheduled.cpp
sgholamiTT: Added a comment.
TL;DR the accumulation guard only makes sense for ops with a single CB-fed accumulator; with ≥2 distinct CB inputs, all of them must be reloaded each iteration so we suppress the guard.

lib/Dialect/D2M/Transforms/InsertDstRegisterAccessUnscheduled.cpp
sgholamiTT: Done.

lib/Dialect/D2M/Transforms/InsertDstRegisterAccessUnscheduled.cpp
sgholamiTT: Done.

lib/Dialect/D2M/Transforms/InsertDstRegisterAccessUnscheduled.cpp
sgholamiTT: Done — both branches now share the longer, more informative message.

lib/Dialect/D2M/Transforms/InsertDstRegisterAccessUnscheduled.cpp
sgholamiTT: Done. Also revisited all asserts in the new code.

include/ttmlir/Dialect/D2M/Transforms/InsertDstRegisterAccessShared.h
sgholamiTT: Done.

include/ttmlir/Dialect/D2M/Transforms/InsertDstRegisterAccessShared.h
sgholamiTT: Agree. Will try to do it in a follow up PR.

include/ttmlir/Dialect/D2M/Transforms/InsertDstRegisterAccessShared.h
sgholamiTT: Cleaned up.

include/ttmlir/Dialect/D2M/Transforms/InsertDstRegisterAccessShared.h
sgholamiTT: Removed. `D2MGenericRegionOps.h` already brings in the full type.

include/ttmlir/Dialect/D2M/Transforms/InsertDstRegisterAccessShared.h
sgholamiTT: Renamed: `OperationTypes -> DstRegionOpClassification`, `getOperationTypes -> classifyDstRegionOps`.

lib/Dialect/D2M/Transforms/InsertDstRegisterAccessScheduled.cpp
sgholamiTT: No. MLIR op accessors (`.getLoc`, `.getIndices`, etc) aren't const-marked.

lib/Dialect/D2M/Transforms/InsertDstRegisterAccessScheduled.cpp
sgholamiTT: Same as above; not possible.

lib/Dialect/D2M/Transforms/InsertDstRegisterAccessShared.cpp
sgholamiTT: Done.

include/ttmlir/Dialect/D2M/Transforms/InsertDstRegisterAccessShared.h
sgholamiTT: Done.

test/ttmlir/Dialect/D2M/Transforms/insert_dst_register_access_l1_acc.mlir
sgholamiTT: Not super necessary but I find it helpful. Trimmed it to 3 lines.

lib/Dialect/D2M/Transforms/InsertDstRegisterAccessUnscheduled.cpp
sgholamiTT: Reworded — the function wraps the pre-existing compute loop with two copy nests (`CB→DST` before, `DST→CB` after); the compute loop is left in place and rewritten to go through DST.

lib/Dialect/D2M/Transforms/InsertDstRegisterAccessUnscheduled.cpp
sgholamiTT: Removed — already documented in `Passes.td`.

include/ttmlir/Dialect/D2M/Transforms/InsertDstRegisterAccessShared.h
sgholamiTT: Switched to `llvm::MapVector` for both maps which is deterministic.

lib/Dialect/D2M/Transforms/InsertDstRegisterAccessUnscheduled.cpp
sgholamiTT: Pulled the clone logic and the per-record emit loop up into `Shared.{h,cpp}` as `cloneAffineLoopSkeleton` + a templated `emitDstCopyNest`. `createCopyLoop` is gone; `dataCopyGenerateWithClone` is now a small adapter on top of `emitDstCopyNest`. The unscheduled-only knobs (`enableL1Acc`, `per-IV guard`) are parameters on the shared helper.

lib/Dialect/D2M/Transforms/InsertDstRegisterAccessUnscheduled.cpp
sgholamiTT: The control flow is now in a single shared template `emitDstCopyNest` in `Shared.{h,cpp}`. Both passes go through it; the scheduled side is a 15-line adapter. The lambda-pair shape stays (`copyGenerator` + `accessReplacer`) since each pass legitimately wants different code there, but they're no longer wrapping duplicated control flow.
I believe there's always better ways of doing things and this is no exception. Let me know if you have other suggestions.

test/ttmlir/Dialect/D2M/Transforms/insert_dst_register_access_unscheduled_skips_scheduled_roots.mlir
sgholamiTT: Hmm I'm not sure.
The two precondition tests had identical IR and just different RUN lines, so merged them into the existing `insert_dst_register_access_missing_linalg_to_affine.mlir` with two RUN lines, deleted the `_scheduled` variant. 
The other tests each pin a distinct contract (different allocators, different routing, different IR shapes) and merging them would either require gymnastics with `--check-prefix` or drop coverage.
Let me know what you think.

vwellsTT: Yeah, I don't feel super strongly here or anything, probably okay to revisit later

lib/Dialect/D2M/Transforms/InsertDstRegisterAccessUnscheduled.cpp
```diff
@@ -0,0 +1,445 @@
...
+  DstIntermediatesMap dstIntermediates;
+
+  auto getInPlaceDstSlice = [&](OperandLoadStoreRegisterOpInterface op) -> int {
+    for (int64_t operandIdx : op.getOperandsLoadFromDstRegister()) {
+      if (op.isScalarOperand(operandIdx)) {
+        continue;
+      }
+      auto *defOp = op->getOperand(operandIdx).getDefiningOp();
+      if (defOp) {
+        auto it = dstIntermediates.find(defOp);
```
vmilosevic: :warning: **llvm\-qualified\-auto** :warning:
`` auto it `` can be declared as `` auto *it ``
```suggestion
        auto *it = dstIntermediates.find(defOp);
```

vmilosevic (commented):

lib/Dialect/D2M/Transforms/InsertDstRegisterAccessUnscheduled.cpp
vwellsTT: yeah, imo that's a clearer spot for it, thanks

lib/Dialect/D2M/Transforms/InsertDstRegisterAccessUnscheduled.cpp
vwellsTT: Nothing super elegant jumped out at me, just took me a while to parse this--so probably good for now, and that sounds better.  Thanks

include/ttmlir/Dialect/D2M/Transforms/Passes.td
vwellsTT: Ah great

lib/Dialect/D2M/Transforms/InsertDstRegisterAccessScheduled.cpp
vwellsTT: I think I made the first Transform subdir yesterday FWIW haha.  But yeah don't feel strongly here

lib/Dialect/D2M/Transforms/InsertDstRegisterAccessShared.cpp
vwellsTT: Fair enough, thanks!

lib/Dialect/D2M/Transforms/InsertDstRegisterAccessScheduled.cpp
sgholamiTT: Ugh sorry I answered the comments and then rebased (so I missed your recent change).
I will try to move the files in a follow up PR.

lib/Dialect/D2M/Transforms/InsertDstRegisterAccessShared.cpp
vwellsTT: Ah yeah my bad, I guess I just saw the re-assignment and not the RHS of the arg (emitting an op).

lib/Dialect/D2M/Transforms/InsertDstRegisterAccessScheduled.cpp
vwellsTT: I think most MLIR types (ops, attrs, etc) are intended to be copy-iterated like this anyway, FWIW

lib/Dialect/D2M/Transforms/InsertDstRegisterAccessScheduled.cpp
```diff
@@ -0,0 +1,676 @@
...
+
+            dstIntermediates[computeOp] = {allocatedIndex,
+                                           outermostInnerComputeLoop};
+          }
+        }
+
+        // Reserve any extra DST scratch slices the op declares via the
+        // interface so they don't collide with operand/output slots. No
+        // deallocate: the slot is logically owned by the op for its
+        // lifetime.  TODO(https://github.com/tenstorrent/tt-mlir/issues/8081):
```
anuragsingh-tt, line 366: Is this the right issue?

vwellsTT (approved):
Thanks for addressing all my feedback :)

anuragsingh-tt (approved):
Thanks for your patience.

lib/Dialect/D2M/Transforms/InsertDstRegisterAccessScheduled.cpp
sgholamiTT: Hmm actually this was ported from [main](https://github.com/tenstorrent/tt-mlir/blob/0d56907670cff01431df4027b4c2c316b187aa20/lib/Dialect/D2M/Transforms/InsertDstRegisterAccess.cpp#L2036-L2044) (was added in [this PR](https://github.com/tenstorrent/tt-mlir/commit/75d025c16b7782445dc87e3e0d18ebd5ea1d3d8c) from @dlokeTT)
Not super sure about this but it doesn't seem irrelevant to me.

sgholamiTT:
Parent issue tracking follow up works as well as any other remaining work for dst register pass:
https://github.com/tenstorrent/tt-mlir/issues/8344

lib/Dialect/D2M/Transforms/InsertDstRegisterAccessScheduled.cpp
anuragsingh-tt: Ah, now I remember. Yeah this should be addressed, basically scratch shouldn't be pushed onto the stack allocator. It will show up in that situation.

anuragsingh-tt: Not a blocker for this PR

PR #7920, opened 2026-04-10
VoVNet optimization Issue #4349

### LINKED ISSUE #4349: [Bounty $2000] Tune a model - VoVNet
_This is a technical bounty suitable for developers with experience in ML model optimization, C++, and performance engineering. Knowledge of neural network architectures and low-level optimization techniques is highly beneficial._

## Goal

Your mission, should you choose to accept it, is to optimize VoVNet model to achieve **~1400 FPS** throughput using our TT-NN C++ library. 

**Starting Point**: Use our existing toolchain to bootstrap your optimization:
- Leverage our `tt-mlir` compiler through `tt-alchemist` tool to create a baseline standalone solution that's already partially optimized
- From this foundation, apply manual optimizations and hand-tuning techniques to reach the target performance

**Feedback Welcome**: Providing detailed feedack on the experience using `tt-alchemist`, `tt-mlir` and all the other tools/docs/guides/resources is **strongly** encouraged. Documenting any pain points, unclear workflows, missing features, or suggestions for improvement is appreciated. This feedback is valuable and will be considered as part of the overall contribution.

The challenge is to push beyond automated optimization limits through creative manual tuning, graph optimization, memory layout improvements, kernel optimization, and other performance engineering techniques.

## Resources
This section contains a list of resources across various repos. Some resources are tools, others are guides.

### [tool] tt-alchemist

- [Docs](https://docs.tenstorrent.com/tt-mlir/tt-alchemist.html)
- Codegen tool that converts TTIR dialect .mlir files into standalone solutions that use TT-NN library (both C++ and Python available).

### [tool] tt-explorer

- [Docs](https://docs.tenstorrent.com/tt-mlir/tt-explorer/tt-explorer.html#tt-explorer)
- Multipurpose tool that can visualize models from `.mlir` files + can be used to tune a model via GUI.

### [guide + tool] Device perf profiling

- [Usage docs](https://docs.tenstorrent.com/tt-metal/latest/tt-metalium/tools/device_program_profiler.html#device-program-profiler)
- [Our tracy fork](https://github.com/tenstorrent-metal/tracy)
- For profiling, we use the industry standard tracy library.

### [tool] TT-NN Visualizer

>_Works only on TT-NN code written in Python (not C++)!_

- [Repo and docs](https://github.com/tenstorrent/ttnn-visualizer)
- Useful multi-purpose tool for interactive visualization, detailed memory profiling and more - full list of features [here](https://github.com/tenstorrent/ttnn-visualizer?tab=readme-ov-file#features).

### [guide] Advanced perf optimizations - metal trace

- [Docs](https://github.com/tenstorrent/tt-metal/blob/main/tech_reports/AdvancedPerformanceOptimizationsForModels/AdvancedPerformanceOptimizationsForModels.md#advanced-performance-optimizations-for-models)
- This optimization technique is very useful to reduce op dispatch overhead, and to hide input/output transfer latency between host and device.

## Technical Specifications

**Model**: `VoVNet-v2`
**Variant**: `timm/ese_vovnet19b_dw.ra_in1k` ([huggingface](https://huggingface.co/timm/ese_vovnet19b_dw.ra_in1k))

**Input Specifications**:
- Image size: `3x224x224` (Channels x Height × Width)
- Recommended batch size: 1-32 (experimentation encouraged)
- Data format: `ttnn.bfloat16`

**Target Hardware**: 
- Wormhole N150

**Performance Target**: 
- **~1400 FPS** sustained throughput

**Deliverables**:
- Optimized TT-NN C++ implementation (not Python)
- Performance benchmarking results
- Documentation of optimization techniques applied
- Feedback report on all the resources used (docs, tools, guides, ...)

>_Currently, we are not worried about accuracy, as we'll be running this model with "fake" inputs/weights. We do want to keep the model in `ttnn.bfloat16` data format._

## How-to

>_We are currently going through a refactor, so running some things might seem a little convoluted. This section will cover the steps and will be updated as we streamline things in the coming weeks._

#### Step 1.
Clone the following two repos:
- `tt-forge`: https://github.com/tenstorrent/tt-forge
- `tt-forge-fe`: https://github.com/tenstorrent/tt-forge-fe
```
git clone git@github.com:tenstorrent/tt-forge.git
git clone git@github.com:tenstorrent/tt-forge-fe.git
```

#### Step 2.
Setup the `tt-forge-fe` repo, following the instructions found [here](https://docs.tenstorrent.com/tt-forge-fe/getting_started_build_from_source.html). Building from source is necessary.

#### Step 3.
Activate the `tt-forge-fe` python environment, and then cd into your local `tt-forge` repo.
Open [this benchmark script](https://github.com/tenstorrent/tt-forge/blob/db7e29d61e341e51332e0b8043f7af58ca9109e7/benchmark/run_benchmark_wh.sh#L48) to analyze how models are run. You'll find the command for vovnet here as well, which you will use to generate the MLIR file.

#### Step 4.
Run the vovnet benchmark command found in previous step:
```
python benchmark/benchmark.py -p tt-forge-fe -m vovnet -ts classification -bs 8 -df bfloat16 -lp 32 -o forge-benchmark-e2e-tt-forge-fe-vovnet_osmr.json
```

This is the place to change the batch size, data formats, etc. However, as previously mentioned, we want to stick to bfloat16 for now.

This command will run [this model](https://github.com/tenstorrent/tt-forge/blob/db7e29d61e341e51332e0b8043f7af58ca9109e7/benchmark/tt-forge-fe/vovnet.py#L71), you may experiment with various config settings, found [here](https://github.com/tenstorrent/tt-forge/blob/db7e29d61e341e51332e0b8043f7af58ca9109e7/benchmark/tt-forge-fe/vovnet.py#L127).

Make sure that the perf of the model makes sense. At the time of writing, the perf for the linked command is ~250 samples/sec.

The benchmark command will generate a couple MLIR files, usually found in `~/testify/ll-sw/VovnetTimm/mlir_reports`. We are interested in the `ttir.mlir` MLIR file. This is a high-level description of the model, without any optimizations enabled on it - we will need to rerun these optimizations in the next step.

#### Step 5.
Clone another repo - `tt-mlir`:
```
git clone git@github.com:tenstorrent/tt-mlir.git
```
Build according to [the docs](https://docs.tenstorrent.com/tt-mlir/getting-started.html), activate its python env.

Build `tt-alchemist` according to these [docs](https://docs.tenstorrent.com/tt-mlir/tt-alchemist.html).

Convert the `~/testify/ll-sw/VovnetTimm/mlir_reports/ttir.mlir` file from TTIR dialect to EmitC dialect by running `tt-alchemist`. To turn on the optimizer, add this flag:
```
--pipeline-options "enable-optimizer=true memory-layout-analysis-enabled=false" 
```

You'll end up with a directory that has all the necessary files to build and run the model from C++, using the `run` script provided.

#### Step 6.
Confirm the perf is similar to before.

The first run should be slower, as the kernels are compiled for each op before running. Second loop of the model should give the expected performance.

For example, one could loop 10 times and print time take for each __batch__:
```C++
#include <iostream>
#include <chrono>
...
  for (int i = 0; i < 10; i++) {
    auto start = std::chrono::high_resolution_clock::now();
    std::vector<ttnn::Tensor> v202 = forward(v201);
    ttnn::Tensor t = ttnn::from_device(v202[0]);   // block until tensor is read from device to host
    auto end = std::chrono::high_resolution_clock::now();
    std::chrono::duration<double> duration = end - start;
    std::cout << "Time taken for batch: " << duration.count() << " seconds" << std::endl;
  }
```

#### Step 7.
Optimize and document experience with all the resources! :)

**Koyeb Machines:** Training should be run on Koyeb-hosted N150 instances.

Request access here:[ https://www.koyeb.com/solutions/tenstorrent](https://www.koyeb.com/solutions/tenstorrent)

Wait for the onboarding email (from Koyeb) with detailed instructions to get started.

No separate guide will be provided, setup is straightforward.

dimitrijepesic:
### Ticket
https://github.com/tenstorrent/tt-mlir/issues/4349

### Problem description
VoVNet baseline from tt-alchemist runs at ~68 FPS. Target is ~1400 FPS. The main bottleneck is that the generated code re-prepares weights from host on every forward call, and conv2d can't be used with Metal Trace.

### What's changed
Result: 68 → 254 FPS (3.7x speedup)

What I did:
1. Cached conv2d weights — weights are prepared once on the first call and reused. 
2. Moved layout ops to device — replaced 45 host-side to_layout calls with device-side tilize/untilize so they don't block trace.
3. Switched to prim::conv2d — the high-level conv2d does host writes every call which blocks Metal Trace. Replaced 35 of 41 convs with the lower-level prim::conv2d that runs purely on device. 6 large convs still use the high-level API because they don't fit in L1.
4. Metal Trace — works on the forward body. Showed that the bottleneck is now pure compute, not dispatch overhead.

Next approach to reach 1400 FPS:
1. Lower math fidelity (LoFi) — accuracy doesn't matter for this bounty
2. Double buffering in conv config
3. Remove batchnorm/SE-block ops (not needed without accuracy requirements)
4. L1 sharding for activations

### Checklist

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1212806203871143


sdjordjevicTT:
Wow, thanks for the contribution @dimitrijepesic! We will take some time this week to look at it!

svuckovicTT:
Hey @dimitrijepesic, will find some time to take a look tomorrow, thanks!

dimitrijepesic:
Hello @svuckovicTT I'd love to hear if there's any feedback on my code so far and if you have any suggestions on my next steps for this optimization

docs/vovnet-bounty/ttnn-standalone.cpp
```diff
@@ -0,0 +1,5479 @@
...
+  ttnn::constEvalFuncWrapper(v389, v390, v391);
+  ::std::vector<::ttnn::Tensor> v392 = g_cached_result_forward_const_eval_57;
+  ::ttnn::Tensor v393 = v392[0];
+  ttnn::distributed::MeshDevice *v394 = ttnn::DeviceGetter::getInstance();
+  auto _to171 = std::chrono::high_resolution_clock::now();
+  ::ttnn::Tensor v395 = ttnn::tilize_with_zero_padding(
+      v2,
+      ::ttnn::MemoryConfig{::ttnn::TensorMemoryLayout::INTERLEAVED,
+                           ::ttnn::BufferType::DRAM, ::std::nullopt});
+  if (g_timing_enabled) g_op_times["tilize"] += std::chrono::duration<double>(std::chrono::high_resolution_clock::now() - _to171).count();
```
svuckovicTT, line 2020: This measuring approach has its flaws - ops are dispatched asynchronously, so you might want to call `ttnn.synchronize_device(device)` to wait for the accelerator to finish running the op. However, I'd suggest using the tooling mentioned in the original bounty issue (much better support for Python TTNN) to measure op performance, it's a nicer experience overall.

docs/vovnet-bounty/ttnn-standalone.cpp
```diff
@@ -0,0 +1,5479 @@
...
+  ::std::vector<::ttnn::Tensor> *v111 = &g_cached_result_forward_const_eval_1;
+  ttnn::constEvalFuncWrapper(v109, v110, v111);
+  ::std::vector<::ttnn::Tensor> v112 = g_cached_result_forward_const_eval_1;
+  ::ttnn::Tensor v113 = v112[0];
+  ::std::function<::std::vector<::ttnn::Tensor>(::std::vector<::ttnn::Tensor>)>
+      v114 = &forward_const_eval_2;
+  ::std::vector<::ttnn::Tensor> v115 = util_create_vec(v29);
+  ::std::vector<::ttnn::Tensor> *v116 = &g_cached_result_forward_const_eval_2;
+  ttnn::constEvalFuncWrapper(v114, v115, v116);
+  ::std::vector<::ttnn::Tensor> v117 = g_cached_result_forward_const_eval_2;
```
svuckovicTT, line 1627: Instead of running this on each forward call, and using the caching mechanism with `g_weights_cached`, a slightly better approach would be to yank the weight loading and consteval code into their own functions/methods.

docs/vovnet-bounty/ttnn-standalone.cpp
```diff
@@ -0,0 +1,5479 @@
...
+      ::ttnn::Layout::ROW_MAJOR, ::std::nullopt,
+      ::ttnn::MemoryConfig{::ttnn::TensorMemoryLayout::INTERLEAVED,
+                           ::ttnn::BufferType::SYSTEM_MEMORY, ::std::nullopt});
+  ::ttnn::Tensor v82 = ttnn::ones(
+      ::ttnn::Shape({1, 1, 1, 512}), ::ttnn::DataType::BFLOAT16,
+      ::ttnn::Layout::ROW_MAJOR, ::std::nullopt,
+      ::ttnn::MemoryConfig{::ttnn::TensorMemoryLayout::INTERLEAVED,
+                           ::ttnn::BufferType::SYSTEM_MEMORY, ::std::nullopt});
+  ::ttnn::Tensor v83 = ttnn::ones(
+      ::ttnn::Shape({192, 512, 1, 1}), ::
```
svuckovicTT, line 5464: Next to `ttnn.synchronize_device()`, `ttnn::from_device()` is also a good way to wait for the device to finish working as this is a blocking call that will wait for the output tensor.

svuckovicTT (commented):
Hey @dimitrijepesic, thanks for posting this PR. Great work on perf, just two more 4x improvements and we're way past the mark :) I'll leave some thoughts on what I think would be good additions :)

As I'm sure you've noticed, the generated code isn't the most readable. There have been some improvements to the tools recently, but mostly in the EmitPy(thon) path - no opposition if you decided to switch to Python, but we can work with this too. Improving readability of the code would allow for easier reviews for us, and faster iterations on improvements for yourself.

We've found LLMs to be really good at refactoring generated models into something similar to a pytorch implementation, when properly guided. Some things to consider:
- deduplicating copies of consteval functions, naming them properly, and moving to their own file
- using a dictionary (map) structure to hold the weights and consteval results
- moving the tensor loading code outside of the "main" (forward) function
- packing the model code into its own file and class
- pack the ops into reusable layers
- have the main.py be a short driver of the code
- etc.

Keep in mind that these are all suggestions, but would help the whole process meaningfully. A good example to look at (even though it's in Python) would be this implemetation of a CLIP model: https://github.com/tenstorrent/tt-xla/tree/svuckovic/sdxl-clip-models/clip_resampler_codegen_1

In the description of this PR, you mention metal trace, but I didn't see metal trace being used in the code.

> Remove batchnorm/SE-block ops (not needed without accuracy requirements)

Another important point I'd like to cover - removing ops from the model wouldn't be a valid path :)

Great work so far, looking forward to seeing the next set of improvements!

dimitrijepesic:
@svuckovicTT  @sdjordjevicTT  

Hello again! I worked on this issue more, and got the FPS from 236 to 293 in this iteration. As of right now, I'm stuck on what to do next, so I'll briefly discuss what I've done and am really open to advice on what can I try next.

I got FPS improvements by implementing metal trace, increasing batch size to 16, fusing convolution, add and ReLU. To be clear about the fusion: I didn't drop any ops from the model. The conv - add - ReLU chain just gets folded into a single conv call via Conv2dConfig.activation - the math is identical, it's just one op at the call site instead of three.

I tried to use BFLOAT_B weights to unlock batch size 32, but that didn't give me any improvement. Also, I tried jumping straight to batch=32 a few times and kept hitting an OSA concat shape mismatch. I also tried pushing conv outputs to L1-interleaved instead of DRAM, but that runs into "Writes are not supported during trace capture" because to_memory_config(L1) allocates dynamically per iter.

I refactored the code for better visibility, so now instead of two enormous files I have:
- main.cpp which works as a driver loop
- model_ttnn.cpp which holds the model as a class, and also covers forward, prepare_constevals and bn_folds functions
- inputs, utils etc.

I am not sure which direction I should take next. I was thinking about using tt-explorer for shard tuning, double-buffering or static L1 buffer pre-allocation, or another approach I just can't seem to find. I would love to hear your thoughts on what I should try first.

Thank you so much for the feedback!

jberkowitzTT:
This is great progress — 236 to 293 FPS is a strong jump, and it sounds like you’ve already picked up the biggest graph-level wins.

If I were choosing what to try next, I’d start with static L1 buffer pre-allocation so you can make the memory path trace-friendly. Since L1-interleaved looked promising, that feels like the highest-signal next step.

After that, I’d prioritize tt-explorer shard tuning on the hottest convs. My guess is the next gains will come more from placement/sharding efficiency than from another high-level model change.

I’d treat batch=32 as a separate debugging track for now. Since BFLOAT_B didn’t help, I wouldn’t make that the main optimization focus until the concat mismatch is resolved.

I’d leave double-buffering until after memory layout and sharding are in a better place.

So in order: static L1 pre-allocation first, tt-explorer second, batch-32 debugging third, double-buffering after that.

Also, the refactor sounds like a good move — that should make targeted profiling a lot easier.

Really nice work so far.

jberkowitzTT:
Hi @dimitrijepesic - wanted to check in for any updates or questions you had.

Thanks!

dimitrijepesic:
Hi @jberkowitzTT I have updates but I didn't have the hardware to test it. I came to Belgrade today and will be testing tonight 😁 Will push updates tomorrow if I get an improvement

dimitrijepesic:
@jberkowitzTT 

Hey! Update after the last round of suggestions. I think I've hit the ceiling at 293 FPS for a single chip, and I want to check the 1400 target with you before picking the next direction.

I wanted to add a quick hardware note I should've flagged earlier is that I'm on the Wormhole Quietbox (4× N300 Wormhole B0 chips). All my work so far has been targeting a single chip via the tt-alchemist standalone path, which is what the bounty doc prescribes. So everything below is per-chip.

What I tried this round:

Static L1 pre-allocation got the mechanism working but every variant lost FPS. Best case I got was 263 which is down about 10%. Any L1 I reserve eats into the slicer's free budget, and the convs that go through op_slicer pick more aggressive slice configs that cost more per iter than the DRAM round-trips I was saving. Codegen already has L1 basically maxed.

Next I tried tt-explorer/shard overrides. I implemented per-conv shard_layout overrides on both prep and prim paths. Mechanism is clean but auto-shard is already picking the right layout on every prim-path conv I tested (conv 7 HEIGHT_SHARDED matched at 292 vs 293). The convs where there'd be real upside (12/21/30/39 the 4 OSA pointwise convs that are ~70% of compute, currently on the slicer fallback) don't fit prim's L1 budget no matter which layout I pick. Without Tracy any further shard-layout tuning is guesswork.

Batch=32 debugging led me to the finding that the concat shape mismatch you flagged wasn't actually the blocker, it was L1 budget. I got 281 FPS which is worse than baseline.

I also re-emitted through tt-alchemist with the optimizer on to see what shard configs it picks. At b=8 it wants to L1-shard 3 of 4 hot OSA convs, which sounded good, but the optimizer picks 61-core grids assuming all 64 cores are usable, and the runtime reserves 8 for dispatch (56 compute banks per chip). That got a runtime crash. To rule out the possibility that the crash was hiding a real win, I applied the optimizer's chosen layouts as manual overrides at b=8 and got 262 FPS, so the layouts wouldn't have fit anyway, the upstream bug was just masking that.

I also ran the bounty's exact Step 5 command on the extracted TTIR: the emitted cpp builds clean but crashes on iter 0 with a CB clash that my refactor handles via try/catch in the prim path. So my refactor is roughly already where "manual optimization on top of the emitted cpp" is supposed to land.

Now, all three independent experiments (L1 alloc, batch sweep, b=8 sharding) all converge: per chip is saturated around 293 FPS for this model at 224² b=16. The public benchmark lists VoVNet at 240 target / 248 current, so I'm ~22% over the per-chip target. The 4 hot OSA convs sit on the slicer because they don't fit prim L1 at b=16, and they're 70% of total compute which is the wall I'm hitting.

If 1400 is total across the box, it matches data-parallel across ~5 of the 8 chips we have, but the prescribed tt-alchemist standalone path is single-device because multi-chip replication isn't part of the documented flow.

So the questions I'd love your read on:

1. Is the 1400 number per-chip or total throughput across the quietbox? If total, is mesh device/data-parallel replication the intended path, and is there a recommended pattern for replicating a single-device traced binary across the mesh?
2. If it's per-chip, do you see a lever I've missed? The only one I haven't fully tried is hand-rolled ShardSpecs (core_grid + act_block_h) on convs 12/21/30/39, but without a Tracy-enabled runtime that's pure guesswork.
3. Would "61-core grid vs 56 compute banks" error be considered a bug in tt-mlir? It is a real defect if anyone targets single-chip with the optimizer on.

Happy to go in any direction once we nail down which throughput number we're actually after. Thanks for the patience!

jberkowitzTT:
Hi @dimitrijepesic - pulling in @svuckovicTT to review your questions

svuckovicTT:
Hey @dimitrijepesic - just letting you know this is on my radar!

svuckovicTT:
Hey @dimitrijepesic, thanks for pushing on this.

To cover your questions:
1. 1400 fps is the target for a single n150 (Wormhole) chip.
2. Can you expand more on what kind of problems you're having with tracy? I imagine it would be next to impossible to tune perf without it - you wouldn't really know where you're leaking time.
3. It might! Depends on how you've set things up though. If you look at the spec for Wormhole cards (https://tenstorrent.com/hardware/cards#compare, choose Wormhole), you'll see that n300 cards have 64 cores per chip vs n150 cards that have 72. It might just be that you're targeting an n150 but running on an n300, which would explain the error. If you only have access to an n300, we can scale the perf target to account for missing cores `(1400*56/64) = 1225`. Please consider opening an issue with info on how you're running MLIR when hitting this core-count-mismatch error!


I've just tried codegenerating this model in bfloat16, batch-size of 1, using optimization level 2 (which actually L1 shards the activations) and manually added metal tracing - this is what I'm seeing on an n150:
```
Device: MeshDevice(1x1 grid, 1 devices)
Run 0 (eager): 4.9139s (0.20 FPS)
Trace captured
Run 2 (trace): 0.0029s (339.62 FPS)
Run 3 (trace): 0.0029s (342.25 FPS)
```

I'd imagine the bar being quite higher with bfp8_b weights + bigger batch sizes + metal trace with 2 command queues + some other model specific tunings.

Looking at your code, all the ops are DRAM inteleaved, none of them are L1 sharded.

My suggestion would be to switch over to Python codegen - it makes a lot of tooling easier to use and is a nicer experience in general. You'll also get actual weights serialized to disk, so you can track output accuracy vs a torch CPU implementation.  Here's the example script I used to generate the code: 
```
import timm
import torch
import torch_xla.runtime as xr
from tt_torch import codegen_py


EXPORT_PATH = "vovnet_codegen"


def main():
    # Set up XLA runtime for TT backend
    xr.set_device_type("TT")

    # Load ese_vovnet19b_dw.ra_in1k from HuggingFace via timm
    model = timm.create_model("ese_vovnet19b_dw.ra_in1k", pretrained=True)
    model.eval()
    model = model.to(torch.bfloat16)
    x = torch.randn(1, 3, 224, 224, dtype=torch.bfloat16)

    compiler_options = {
        "codegen_split_files": True,
        "optimization_level": 2,
        # "experimental_weight_dtype": "bfp_bf8",  # you can even try going further into bfp8_b with weights!
    }

    codegen_py(
        model,
        x,
        export_path=EXPORT_PATH,
        compiler_options=compiler_options,
    )


if __name__ == "__main__":
    main()

```

dimitrijepesic:
Thank you so much @svuckovicTT for feedback! I'll check all of this today, I'm hoping to get something done by tomorrow.

dimitrijepesic:
Hi @svuckovicTT ,

I followed your suggestion to switch to Python codegen and ran into a chain of issues that I want to walk through, because they all point in the same direction.

First I tried codegen_py on Quietbox with VoVNet. It fails with a `ValueError: Error code: 13` on `torch_xla._XLAC._xla_sync_multi`, and I get the same crash on resnet18. Plain MNIST inference works fine, and all the IRs (vhlo, shlo, ttir, ttnn) emit successfully before it dies, so the compile pipeline is reaching the end of the pass list and the failure is on the trace-execute step. I tried codegen_cpp next and hit the same error 13 in the same spot, but with a try/except I could still keep the emitted TTNN MLIR and feed it forward by hand.

Next step was `tt-alchemist generate-cpp` on that MLIR. The parser rejects it with a `chipDescs` syntax error. The wheel `pjrt-plugin-tt 1.1.0` is emitting MLIR with newer `ChipDescAttr` fields (`coord_translation_offsets`, `dram_grid`, `dram_bank_to_logical_worker_noc0/1`) than what the tt-mlir checkout under tt-xla main currently understands. I traced these fields back to commits `d6102f30d` (dram_grid) and `13041813f` (bank-to-worker), checked out tt-mlir at `d6102f30d^`, and rebuilt `libtt-alchemist-lib.so` against that older revision. The parser then accepts the wheel-emitted MLIR cleanly, and the generation succeeds: 2202-line `ttnn-standalone.cpp` plus CMakeLists, run script, and a bundled ttnn-install. It builds without warnings.

Runtime is where it falls over:
```
TT_FATAL: Expected number of shards 64 to be less than or equal to total number of L1 banks 56 in compute cores
  at bank_manager.cpp:430
```

This happens at `ttnn::multiply` in the eSE block (line 2030 of the emitted cpp), where the optimizer picked an `8x8 = 64-shard` HEIGHT_SHARDED layout. That fits N150 but breaks N300, This is the same root cause I flagged a week ago: tt-mlir grid selection doesn't subtract the dispatch reservation. Back then I was seeing 61-shard layouts, now with the newer optimizer it's gone up to the full 64.

I hand-patched the two offending lines, replacing `CoreCoord{7,7}` with `CoreCoord{7,3}` and adjusting the shard shape to `{1024,32}` so the math divides cleanly on 32 cores. The build is fine and it runs longer this time, but it hits the next layer of the same problem:

```
TT_FATAL: Illegal kernel placement for reshard_reader, Kernels cannot be placed on dispatch cores!
  at program.cpp:143
```

The `reshard` device op also assumes full grid availability and tries to place a kernel on a dispatch-reserved core. So the assumption isn't isolated to one op, it's spread across the codegen and several device ops. Hand-patching just keeps surfacing the next layer with the same baked-in N150-grid assumption.

While I was waiting for builds during all this, I ran a sanity check with raw `torch_xla` on the same model, just `model.to(xla_device); model(x)`. Clean wheel install, `TT_VISIBLE_DEVICES=0`. The numbers were:

```
batch=1:    3.43 ms/iter,    298 FPS
batch=8:    3.34 ms/iter,  2,395 FPS
batch=16:   3.34 ms/iter,  4,795 FPS
batch=32:   3.28 ms/iter,  9,752 FPS
batch=64:   3.29 ms/iter, 19,446 FPS
batch=128:  3.25 ms/iter, 39,331 FPS
```

Per-iter latency stays flat at around 3.3 ms from b=1 all the way to b=128, so this is fixed-overhead dominated and there's no compute saturation in sight. I verified the output against CPU: max diff 0.04, mean diff 0.008, PCC > 0.99, which is just bfloat16 quantization noise. So the hardware itself can clearly deliver far above the target. It's specifically the codegen path that can't reach those numbers on N300.

I have two asks:

1. Can I get an N150 instance on Koyeb? The bounty doc points to Koyeb N150s in the first place, and every codegen issue I've hit looks N300-specific (the 64-compute-core assumption baked across the optimizer and multiple device ops). 

2. Could the deadline be pushed? I think there's real progress here (the 39k FPS proof, a clean reproduction chain all the way through tt-alchemist, and several consistent upstream bug instances that confirm the §15 pattern), and the discovery that the prescribed codegen path is effectively gated on N150 access changes the timeline a bit.

Thanks for the patience on all this. If you want to repro any of it locally, I can push the generated cpp, the hand-patched version, and the torch_xla bench script to the branch.

svuckovicTT:
Hey @dimitrijepesic - glad to see you're making progress!

Let me start by addressing some of the issues you've listed here:
> ValueError: Error code: 13 on torch_xla._XLAC._xla_sync_multi

This is a "catch all" error message that isn't very useful unfortunately. To get a better understanding of why the process is erroring out, you may build the DEBUG flavor of tt-xla, and run with `TTXLA_LOGGER_LEVEL=DEBUG`.

On the core mismatch front, I think it's best we run the model in data parallel mode, i.e. use both chips and parallelize over the batch dim. This should be fairly trivial to do (couple lines in the codegen script), you may find an example here, which you can tailor for codegen:
https://github.com/tenstorrent/tt-xla/blob/main/examples/pytorch/resnet_dp.py#L44-L55

As for the preliminary perf numbers, I'm not convinced they are accurate :) If I had to venture a guess, I'd assume you're not waiting on the output tensor(s) to materialize, so you're only measuring dispatch.

@jberkowitzTT can we extend the bounty deadline?

jberkowitzTT:
I can extend the bounty another week & we can reevaluate then 

cc @svuckovicTT @dimitrijepesic

jberkowitzTT:
New due date: June 5th

samiyadev786:
@dimitrijepesic Are you still working on this issue? The extension date has already passed.

@jberkowitzTT, if @dimitrijepesic is no longer able to continue or decides to give it up, could you please assign the issue to me? I understand they may be busy, and I'd be happy to work on it and submit a complete solution PR.

jberkowitzTT:
This is due on June 18th, @dimitrijepesic please confirm if you intend to complete this by then. 

I've also been instructed by the engineering team to close out this bounty since unfortunately this is no longer a top priority. As a result, we will not reassign this but will keep it open until the 18th.

Thank you!

dimitrijepesic:
Hi @jberkowitzTT, I'll sadly have to forfeit this bounty as I won't have enough time to finish it. Thank you so much for the support along the way! :)

PR #7925, opened 2026-04-10
[OpModel] Introduce `opModelExempt` and `ASSIGN_OR_RETURN` macro

sgholamiTT:
### Ticket
https://github.com/tenstorrent/tt-mlir/issues/7872

### Summary

This PR introduces three improvements to the OpModel infrastructure that collectively remove ~2,000 lines of repetitive boilerplate:

1. **`OpModelExempt` trait** — A new marker trait that exempts ops from implementing the `TTNN_OpModelInterface`. Previously, every TTNN op was forced to declare `getOpConstraints`/`getOpRuntime` methods via the base op classes, and ops that didn't support OpModel had to provide stub implementations that returned error strings. Now, ops can simply add `OpModelExempt` to their trait list to opt out entirely. This removes ~488 lines of boilerplate stub implementations from `TTNNOpModelInterface.cpp`.

2. **`ASSIGN_OR_RETURN` macro** — A shared utility macro (in `TTNNOpModel.h`) that unwraps `llvm::Expected` values, returning the error from the enclosing function on failure. Eliminates the repetitive check-and-extract boilerplate pattern. Applied to 243 `convertToTensorSpec` call sites in `TTNNOpModel.cpp` (~960 lines removed) and 83 `checkDeviceWorkerGrid` call sites in `TTNNOpModelInterface.cpp` (~330 lines removed).

3. **`getValidatedDeviceGrid` helper** — Combines `checkDeviceWorkerGrid` (validation) and `lookupDevice(...).getWorkerGrid()` (retrieval) into a single function returning `llvm::Expected<ttcore::GridAttr>`. These two calls were always paired across 78 call sites.

### Changes

**`ASSIGN_OR_RETURN` macro (`TTNNOpModel.h`)**
- Defined `ASSIGN_OR_RETURN(lhs, expr)` macro using the Clang statement-expression extension (`lhs = ({ ... })`). The single-statement form is required to be safe inside unbraced `if`/`else`/`for`/`while` bodies — the original three-statement expansion would silently produce wrong behavior in `if (cond) ASSIGN_OR_RETURN(...);` because only the first statement was guarded by the `if`. The macro internally suppresses Clang's `-Wgnu-statement-expression-from-macro-expansion` warning so callers don't have to.
- Shared across `TTNNOpModel.cpp` and `TTNNOpModelInterface.cpp`.

**New trait (`TTNNTraits.h` / `TTNNTraits.td`)**
- Added `OpModelExempt` native op trait.

**Base op classes (`TTNNBase.td`)**
- Added `IfOpModelRequired` TableGen class that conditionally includes `DeclareOpInterfaceMethods<TTNN_OpModelInterface>` only when `OpModelExempt` is absent.
- Updated `TTNN_Op`, `TTNN_InplaceOp`, `TTNN_MemFreeOp`, and `TTNN_MemoryEffectOp` to use `IfOpModelRequired`.

**Ops marked exempt (`TTNNOps.td`)**
- 30 ops tagged with `OpModelExempt`: `GetDeviceOp`, `ToDeviceOp`, `FromDeviceOp`, `SparseMatmulOp`, `AllToAllDispatchOp`, `AllToAllDispatchMetadataOp`, `AllToAllCombineOp`, `SelectiveReduceCombineOp`, `MoeExpertTokenRemapOp`, `DistributedRMSNormOp`, `DistributedLayerNormOp`, `AllocOp`, `AllGatherOp`, `ReduceScatterOp`, `AllReduceOp`, `AllReduceAsyncOp`, `MeshShardOp`, `PointToPointOp`, `WriteTensorOp`, `BeginTraceCaptureOp`, `EndTraceCaptureOp`, `ExecuteTraceOp`, `CaptureOrExecuteTraceOp`, `GenericOp`, `DumpTensorOp`, `LoadTensorOp`, `DistributeTensorOp`, `AggregateTensorOp`, `CreateGlobalSemaphoreOp`, `ResetGlobalSemaphoreOp`.

**Stub removal + grid check cleanup (`TTNNOpModelInterface.cpp`)**
- Removed all `getOpConstraints`/`getOpRuntime` stub implementations for the exempt ops (~488 lines deleted).
- Added `getValidatedDeviceGrid` helper function combining grid validation and retrieval.
- Replaced 83 `checkDeviceWorkerGrid` boilerplate blocks with `ASSIGN_OR_RETURN` calls (~330 lines removed).

**TensorSpec boilerplate cleanup (`TTNNOpModel.cpp`)**
- Replaced 243 `convertToTensorSpec` error-check-unwrap blocks with `ASSIGN_OR_RETURN` calls (~960 lines removed).

**Optimizer fallback for exempt ops (`OpConstraintValidation.cpp`)**
- Before this PR, ops without OpModel returned a stub `OpNotSupportedError` that `validateConstraints` translated to `ValidationStatus::NotImplemented` via `handleAllErrors`. The optimizer (`GreedyMemoryLayoutPropagation`, `L1SpillManagement`, `OperationValidationAndFallback`) branches on `result.isNotImplemented()` to fall back gracefully. After removing those stubs, `dyn_cast<OpModel>(op)` returns `nullptr` for exempt ops and the old code path called `llvm::reportFatalInternalError` — crashing the compiler.
- Detect `OpModelExempt` at the op-cast boundary and synthesize a `NotImplemented` `ValidationResult` directly so the optimizer's fallback path is preserved. Non-exempt ops without OpModel still trip the fatal error (real programming bug).

**Documentation (`docs/src/ttnn-op-constraints.md`)**
- Updated "Operations Not Supported" section to document `OpModelExempt` trait.
- Updated Step 2 code example to show `ASSIGN_OR_RETURN` usage.

### Test plan
- [x] `TestOpModelLib` passes
- [x] `TestOpModelInterface` passes
- [x] `ValidationTests` passes — including new `OpConstraintValidationTest.OpModelExemptOpsReturnNotImplementedNotFatal` regression test that exercises two independent exempt ops (`GetDeviceOp`, `AllocOp`) and `static_assert`s the trait, so a future refactor that drops the trait or removes the fallback fails loudly instead of silently regressing the optimizer.
- [x] Existing `OpConstraintValidationTest.ValidationStatusNotImplemented` confirmed to surface `NotImplemented` (not crash) for an exempt op.


lib/OpModel/TTNN/TTNNOpModel.cpp
```diff
@@ -52,6 +52,14 @@ namespace mlir::tt::ttnn::op_model {
...
 #define QUERY_OP_RUNTIME(op, device, ...)                                      \
   ::ttnn::graph::query_op_runtime(WRAP_OP(op), device, __VA_ARGS__)
+
+#define CONCAT_IMPL_(a, b) a##b
+#define CONCAT_(a, b) CONCAT_IMPL_(a, b)
+#define ASSIGN_OR_RETURN(lhs, expr)                                            \
+  auto CONCAT_(_tt_expected_, __LINE__) = (expr);                              \
+  if (!CONCAT_(_tt_expected_, __LINE__))                                       \
+    return CONCAT_(_tt_expected_, __LINE__).takeError();                        \
+  lhs = std::move(CONCAT_(_tt_expected_, __LINE__).get())
```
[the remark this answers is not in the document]
sgholamiTT: The `do { ... } while(false)` approach doesn't work here because `lhs` is a declaration (e.g., `::ttnn::TensorSpec inputSpec`) in every usage — the variable would go out of scope at the end of the `do` block. The `__LINE__` concatenation is also still necessary regardless of scoping, since a single function can contain multiple `ASSIGN_OR_RETURN` calls that each need a distinct temporary. This is the same design used by [Abseil's](https://notes.haifengjin.com/tech/software_engineering/c%2B%2B/absl/) `ASSIGN_OR_RETURN` and similar macros in other projects for the same reason.

include/ttmlir/Dialect/TTNN/IR/TTNNBase.td
```diff
@@ -42,24 +42,36 @@ def TTNN_Dialect : Dialect {
...
+//===----------------------------------------------------------------------===//
+// TTNN traits definition for OpModel.
+//===----------------------------------------------------------------------===//
+
+// Returns [DeclareOpInterfaceMethods<TTNN_OpModelInterface, ...>] when
+// OpModelExempt is absent from traits, or [] when present.
+class IfOpModelRequired<list<Trait> traits> {
+  list<Trait> value = !if(!empty(!filter(t, traits, !eq(t, OpModelExempt))),
+      [DeclareOpInterfaceMethods<TTNN_OpModelInterface,
+        ["getOpConstraints", "getOpRuntime"]>], []);
```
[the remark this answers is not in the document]
sgholamiTT: The logic is correct as written. `!filter(t, traits, !eq(t, OpModelExempt))` produces a list containing only `OpModelExempt` entries. When `OpModelExempt` is absent, that list is empty, so `!empty(...)` returns `true`, and the `!if` takes the `then` branch which adds `DeclareOpInterfaceMethods`. When `OpModelExempt` is present, the filter finds it, `!empty` returns `false`, and the `!if` takes the `else` branch (empty list). This matches the comment on lines 49-50 and the intended behavior — verified by the fact that all 27 exempt ops compile and pass tests without stub implementations.

sgholamiTT:
fyi @sdjordjevicTT

arminaleTT (approved):
LGTM

include/ttmlir/OpModel/TTNN/TTNNOpModel.h
```diff
@@ -12,6 +12,18 @@
 #include "llvm/ADT/ArrayRef.h"
 #include "llvm/Support/Error.h"
 
+// Unwrap an llvm::Expected<T> expression: on success assign the value to `lhs`,
+// on failure propagate the error by returning it from the enclosing function.
+// clang-format off
+#define TTMLIR_CONCAT_IMPL_(a, b) a##b
+#define TTMLIR_CONCAT_(a, b) TTMLIR_CONCAT_IMPL_(a, b)
+#define ASSIGN_OR_RETURN(lhs, expr)                                            \
```
xanderchin, line 20: ### Claude Review: Macro is not `do { ... } while(0)` wrapped — unsafe in unbraced `if`/`else` (TTNNOpModel.h line 20)

The `ASSIGN_OR_RETURN` macro expands to **three separate statements**:
```cpp
auto _tt_expected_42 = (expr);
if (!_tt_expected_42) return _tt_expected_42.takeError();
lhs = std::move(_tt_expected_42.get())
```

If used inside an unbraced `if`:
```cpp
if (condition)
  ASSIGN_OR_RETURN(Foo foo, bar());  // Only first statement is guarded!
```

This would expand to:
```cpp
if (condition)
  auto _tt_expected_42 = bar();       // guarded by if
if (!_tt_expected_42)                  // ALWAYS executed
  return _tt_expected_42.takeError();  // ALWAYS executed
foo = std::move(_tt_expected_42.get()) // ALWAYS executed
```

**Impact:** Currently safe — all 9 usages inside `if` blocks use braces. But the macro is a public API in a header file, and any future usage without braces will silently produce wrong behavior without compiler warnings.

**Why `do { ... } while(0)` won't work here:** The macro declares `lhs` as a new variable in some usages (`ASSIGN_OR_RETURN(Foo foo, ...)`), which means the variable must escape the macro scope. This is a fundamental tension — the variable needs to be visible after the macro, but the macro needs multi-statement safety.

**Fix:** Use the GCC statement-expression extension (already used elsewhere in LLVM):
```cpp
#define ASSIGN_OR_RETURN(lhs, expr)                                            \
  lhs = ({                                                                     \
    auto TTMLIR_CONCAT_(_tt_expected_, __LINE__) = (expr);                     \
    if (!TTMLIR_CONCAT_(_tt_expected_, __LINE__))                              \
      return TTMLIR_CONCAT_(_tt_expected_, __LINE__).takeError();              \
    std::move(TTMLIR_CONCAT_(_tt_expected_, __LINE__).get());                  \
  })
```

Or split into two macros: `DECLARE_OR_RETURN(type, name, expr)` for declarations and `ASSIGN_OR_RETURN(lhs, expr)` (wrappable in `do{...}while(0)`) for assignments.

xanderchin (approved):

lib/Dialect/TTNN/Interfaces/TTNNOpModelInterface.cpp
```diff
@@ -4979,62 +4449,6 @@ MeshPartitionOp::getOpRuntime(const std::vector<TTNNLayoutAttr> &inputs,
...
-  return issueErrorForGetOpRuntime(
-      getOperation(), detail::ReasonForLackOfSupport::MissingMetalDefinition);
-}
-
-//===----------------------------------------------------------------------===//
-// AllReduceAsyncOp - TTNN Op Model Interface
-//===----------------------------------------------------------------------===//
-
-llvm::Expected<op_model::OpConstraints>
-AllReduceAsyncOp::getOpConstraints(const std::vector<TTNNLayoutAttr> &inputs,
```
rpavlovicTT, line 5159: Optimizer relies on capturing `NotImplemented` error from OpValidation. This will likely break it. Search for `isNotImplemented`.

rpavlovicTT (commented):
One major concern. 
Also +1 on Xander's comment. 

Once addressed please run all perf benchmark tests in tt-xla https://github.com/tenstorrent/tt-xla/actions/workflows/manual-benchmark.yml (N150 configuration) and make sure nothing breaks.

azecevicTT (approved):
Nice cleanup (with a caveat that @rpavlovicTT has a valid concern)!

include/ttmlir/OpModel/TTNN/TTNNOpModel.h
sgholamiTT: Fixed.

lib/Dialect/TTNN/Interfaces/TTNNOpModelInterface.cpp
sgholamiTT: Good catch — removing the interface stubs broke the implicit contract the optimizer relies on (`isNotImplemented()` from `OpConstraintValidation`). Without a stub, the dispatch fell through to `llvm::reportFatalInternalError`, which would crash the optimizer instead of letting it skip the op.

Fixed in `OpConstraintValidation::validateConstraints`: when an op doesn't implement `TTNN_OpModelInterface`, we now check for the `OpModelExempt` trait and synthesize a `ValidationResult::notImplemented(...)` instead of fatal-erroring. This restores the previous behavior for exempt ops without forcing them to carry a stub interface implementation.

Added a regression test in `test/unittests/Validation/TestOpConstraintValidation.cpp` (`OpModelExemptOpsReturnNotImplementedNotFatal`) that builds `GetDeviceOp` and `AllocOp` (both exempt), `static_assert`s the trait is present, `ASSERT_FALSE`s that they implement the `OpModel` interface (so we know we're hitting the new code path and not the interface dispatch), and asserts `isNotImplemented()` without crashing. Any future change that drops the trait, the trait check, or reverts to the fatal path will fail this test.

The [perf benchmark](https://github.com/tenstorrent/tt-xla/actions/runs/24682018575) is currently running. Will wait for ✅  on that before merging.

sgholamiTT:
`xla`'s perf benchmark on commit 0bff6ecf3d425eea28f87cb5f904659b3bc442b4 is green ([link](https://github.com/tenstorrent/tt-xla/actions/runs/24682018575)) ✅ 
Except for two test suites that are not relevant to these changes ([failure1](https://github.com/tenstorrent/tt-xla/actions/runs/24682018575/job/72184492967), [failure2](https://github.com/tenstorrent/tt-xla/actions/runs/24682018575/job/72184493029)).

nsmithtt (approved):
Signing off on docs changes

lib/Dialect/TTNN/Interfaces/TTNNOpModelInterface.cpp
rpavlovicTT: Please run LLMBox tests too, you will have to manually set opt-level=2 like here https://github.com/tenstorrent/tt-xla/pull/4341. Soon to be default. Don't run those that are currently failing as noted in the PR. Feel free to run on that branch with your tt-mlir commit.

rpavlovicTT (approved):
Approved code changes, merge once CI pipelines are green.

sgholamiTT:
To test this change on LLMBox perf tests the following was done:
1. I created a branch `sgholami/test-op-model-exempt` that is based off of [this PR](https://github.com/tenstorrent/tt-xla/pull/4341)'s branch `rpavlovic/tp_opt2_selective`,
2. I uplifted the `tt-mlir`'s commit SHA to 0bff6ecf3d425eea28f87cb5f904659b3bc442b4 which is the tip of the current PR. This was committed and pushed to `sgholami/test-op-model-exempt` (see [here](https://github.com/tenstorrent/tt-xla/compare/main...refs/heads/sgholami/test-op-model-exempt)).
3. I then ran the perf benchmark for LLMBox for `sgholami/test-op-model-exempt`.

The perf benchmark is available [here](https://github.com/tenstorrent/tt-xla/actions/runs/24726741493). 
There are some tests that are failing (which seems unrelated to current changes and most probably caused by the uplift) but most tests pass. I believe the intention for running this was to make sure that ccl ops (that mostly exist in LLMs) will still pass with this change. Given that there are some LLMs passing on the perf benchmark, I think we have ✅ to merge this PR.

PR #3626, opened 2025-06-01
Dynamic includes generation based on function ops.

wusatosi:
### Ticket
Fix #772 .

### Problem description
Currently `TTKernelToCpp` generate `emitc::IncludeOp` statically based on the thread type, which is hacky.

### What's changed
This PR adds the includes dynamically based on the kinds of operations in the target region.
The include map is a manually derived map that is statically stored in a `llvm::StringMap`.

This is done by declaring a static op -> include map (currently one-to-one).
Walk through the target region and collect mapped includes in a set.
Sorting them and emitting them at the top of the file.

### TODOs:
- Please double check my includes, I have listed out operations I am unsure of in https://github.com/tenstorrent/tt-mlir/issues/772#issuecomment-2921122647 , please follow up on that issue or comment on the line directly in this PR, I will keep the table in the comment and the table in the source code in sync.
- How do we want to approach testing for this?

### Checklist
- [ ] New/Existing tests provide coverage for changes

PR #7926, opened 2026-04-10
Dynamically pick required kernel includes

### LINKED ISSUE #772: Implement specific op includes generation
Implement proper creation of `emitc::IncludeOp`s, taking into account differences between unary and binary op kernel includes, which of them exist in general, which are specific to unary vs binary (if any) and which are specific to the concrete op. 

For example, `exp` call needs `compute_kernel_api/eltwise_unary/exp.h` included.

wenbinlyuTT:
### Ticket
Closes #772
Supersedes #3626

### Problem description
Right now the `TTKernelToCPP` pass inserts every API header that might potentially be used.
We need to include only the necessary ones to reduce bloat and prepare for device 2.0 API migration.

### Task Scoping
- All API modeled in the `TTKernel` dialect should map to one or more headers.
  - Prefer exact kernel API name matching to eliminate prefix collisions, use `starts_with` only when it's safe.
  - Prefer precise mapping: if header A includes function F, and header B includes header A, use header A.
  - Avoid internal headers: for example, don't use `dataflow_api_addrgen.h` for `get_noc_addr_from_bank_id`.
  - No missing/duplicated/unused headers and/or definitions.
  - Maintain stable insertion order to avoid randomly invalidate the compilation cache.
  - System headers should use `#included <>`, while the device API headers should use `#include ""`.
  - Ignore `sfpu_split_includes.h`, include headers that contain the actual definitions.

- Special cases:
  - Always needed: `<cstdint>`.
  - Thread-dependent:`cb_*/get_arg_val/get_common_arg_val` need `api/compute/{cb_api,common}.h` in the compute thread, but `api/dataflow/dataflow_api.h` in the datamovement thread.
  - Context-dependent: `reduce.h` requires `REDUCE_OP/DIM` template parameter definitions before its inclusion.
  - Leaked internal headers: `llk_defs.h` for some op enum definitions, `firmware_common.h` for `experimental::convert_logical_x/y_to_translated`.
  - The naughty `TTKernelToEmitC` pass:
    - Inserts ops using `CallOpaqueOp/VerbatimOp` w/o proper headers (e.g. `std::min/max`, `float_to_bits`, `TensorAccessor`, `add_unary_tile` and more).
    - Inserts `get_compile_time_arg_val(N)` using the `LiteralOp`, also w/o its header.
  - Other special headers: profiler, debug print.
  - Experimental kernel snippets under `include/ttmlir/Target/TTKernel/LLKs` that are inserted verbatim, must be inserted after the API headers.

### What's changed
- Create a static table in the new `TTKernelIncludesMap.h` with exact header mappings.
- Walk the final EmitC string calls to collect headers, so `TTKernelToEmitC` injections (`<algorithm>` for `std::min/max`, etc.) are also handled.
- Walk both the `CallOpaqueOp` and the `VerbatimOp`.
- Collect all required headers into a `std::set`, which is naturally deduplicated and alphanumerically sorted.
- Internal headers `firmware_common.h` and `llk_defs.h` are both transitively included, so we no longer insert them.
- Deleted non-existent `TransposeUninitOp`.
- Corrected a bug/typo: the pass checks `experimental::get_logical_mesh_position`, but the op/callee is `experimental::get_my_logical_mesh_position`.
- `get_compile_time_arg_val` and `TensorAccessor` still rely on transitive inclusion.

### Checklist


include/ttmlir/Target/TTKernel/TTKernelIncludesMap.h
```diff
@@ -0,0 +1,274 @@
...
+struct HeaderRequirement {
+  llvm::StringRef computeHeader;
+  llvm::StringRef dataflowHeader;
+};
+
+// This maps the EXACT kernel API name to its required API header(s).
+inline const llvm::StringMap<HeaderRequirement> &getCalleeToHeadersMap() {
+  static llvm::StringMap<HeaderRequirement> headerMap;
+  if (headerMap.empty()) {
+    // clang-format off
```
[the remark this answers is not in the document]
wenbinlyuTT: Switched to static const map, it is guaranteed to execute once and is thread safe.

lib/Target/TTKernel/TTKernelToCpp.cpp
```diff
@@ -33,153 +35,185 @@
 #include "llvm/ADT/ScopeExit.h"
 #include "llvm/Support/raw_ostream.h"
 
+#include <set>
```
vwellsTT: I guess we should use llvm set/vectors instead?

include/ttmlir/Dialect/TTKernel/IR/TTKernelOps.td
```diff
@@ -3042,19 +3042,6 @@ def TTKernel_TransposeTileOp : TTKernel_Op<"transpose_wh_tile"> {
     }];
 }
 
-def TTKernel_TransposeUninitOp : TTKernel_Op<"transpose_wh_uninit"> {
```
vwellsTT, line 3045: why is this removed?

vwellsTT (approved):
Like the idea, a couple minor nits

include/ttmlir/Dialect/TTKernel/IR/TTKernelOps.td
wenbinlyuTT: This doesn't even exist in tt-metal anymore, and it's not used in the compiler.

anuragsingh-tt (approved):

lib/Target/TTKernel/TTKernelToCpp.cpp
wenbinlyuTT: Unfortunately `SmallSet` requires a known static size, and `DenseSet` isn't ordered so leds to very low jit cache hit rate. I was only able to use `SmallVector`.

vwellsTT: Fair, sounds good!

PR #7940, opened 2026-04-12
[d2m] Grid selection for tilized concat ops can pick multi core

### LINKED ISSUE #7936: [d2m] Qwen 3 Embedding 4B Concat Op
```
func.func @concat_0(%arg0: tensor<1x32x128x64xbf16>, %arg1: tensor<1x32x128x64xbf16>) -> tensor<1x32x128x128xbf16> {
  %0 = "ttir.concat"(%arg0, %arg1) <{dim = 3 : si32}> : (tensor<1x32x128x64xbf16>, tensor<1x32x128x64xbf16>) -> tensor<1x32x128x128xbf16>
  return %0 : tensor<1x32x128x128xbf16>
}
```
This snippet runs into this failure
```
[1] concat_0 — COMPILE FAIL
    exception: Failed to run pass manager
    diagnostics:
      can't find feasible allocation because all 9 var(s) are bound
      error: 'func.func' op required L1 memory usage 1572864 exceeds memory capacity 1395424 (usable space is [103712, 1499136))
```

dlokeTT:
### Ticket
Closes #7936 

### Problem description
Concat ran into an OOM allocator issue because we were trying to tilize both inputs to concat on single core

### What's changed
- Updated grid selection to look at `to_layout` input to concat and if it doesn't come from a view, allow grid selection to pick an optimal grid
- Added lit and builder tests

### Checklist

### DIFF
--- a/lib/Dialect/D2M/Transforms/GridSelection.cpp
+++ b/lib/Dialect/D2M/Transforms/GridSelection.cpp
@@ -273,9 +273,41 @@ static RankedTensorType tensorWithOptimalGrid(RankedTensorType oldTensor,
   return RankedTensorType::get(deviceShape, newElementType, newLayout);
 }
 
+// Rebuild a ToLayoutOp and its associated EmptyOp at a new target type.
+// The EmptyOp builder handles VGM (virtual grid mapping) computation
+// automatically when the grid requires virtualization.
+// Returns the new ToLayoutOp, or nullptr if the old op has no EmptyOp output.
+static d2m::ToLayoutOp
+rebuildToLayoutWithType(d2m::ToLayoutOp toLayoutOp, RankedTensorType newType,
+                        const GridSelectionConfig &config, OpBuilder &builder) {
+  auto emptyOp = toLayoutOp.getOutput().getDefiningOp<d2m::EmptyOp>();
+  if (!emptyOp) {
+    return nullptr;
+  }
+
+  builder.setInsertionPoint(emptyOp);
+  auto newEmptyOp = builder.create<d2m::EmptyOp>(
+      emptyOp.getLoc(), newType.getShape(), newType.getElementType(),
+      newType.getEncoding(), config.targetSquareGridShape);
+
+  builder.setInsertionPoint(toLayoutOp);
+  return builder.create<d2m::ToLayoutOp>(toLayoutOp.getLoc(),
+                                         toLayoutOp.getInput(), newEmptyOp);
+}
+
+// Erase a ToLayoutOp and its EmptyOp output (if no longer used).
+static void eraseToLayoutAndEmpty(d2m::ToLayoutOp toLayoutOp) {
+  auto emptyOp = toLayoutOp.getOutput().getDefiningOp<d2m::EmptyOp>();
+  toLayoutOp.erase();
+  if (emptyOp && emptyOp.getResult().use_empty()) {
+    emptyOp.erase();
+  }
+}
+
 // Update a ToLayoutOp and its associated EmptyOp to use a specified grid by
 // recreating the MetalLayoutAttr with the given grid and proper dimension
-// alignments.
+// alignments, then inserting a reblock ViewLayoutOp to preserve IR correctness
+// for downstream consumers.
 static void optimizeToLayoutGrid(d2m::ToLayoutOp toLayoutOp,
                                  const GridSelectionConfig &config,
                                  ArrayRef<int64_t> optimalGrid,
@@ -312,42 +344,12 @@ static void optimizeToLayoutGrid(d2m::ToLayoutOp toLayoutOp,
 
   RankedTensorType newTensorType =
       tensorWithOptimalGrid(outputType, config, optimalGrid, builder);
-  builder.setInsertionPoint(emptyOp);
 
-  // Determine if the chosen grid is virtual (exceeds 2D device bounds or is
-  // ND). Note: VGM is NOT propagated from the to_layout's input here — the
-  // output EmptyOp has its own grid/shard strategy. VGM for DMA addresses
-  // is traced through the stream's input at DMA lowering time.
-  mlir::AffineMapAttr virtualGridInverseMapping;
-  mlir::AffineMapAttr virtualGridForwardMapping;
-  auto device = ttcore::lookupDevice(toLayoutOp);
-  auto workerGridShape = device.getWorkerGrid().getShape();
-  bool isVirtual = ttmlir::d2m::utils::grids::requiresVirtualGrid(
-      optimalGrid, workerGridShape);
-  if (isVirtual) {
-    auto physicalGridShape = utils::findLegalPhysicalGridForVolume(
-        ttmlir::utils::volume<int64_t>(optimalGrid),
-        config.targetSquareGridShape);
-    TT_assertv(
-        !physicalGridShape.empty(),
-        "Unable to find 2D rect that can fit virtual grid {} within "
-        "device grid {}",
-        ttmlir::utils::formatIterable(optimalGrid, "x"),
-        ttmlir::utils::formatIterable(config.targetSquareGridShape, "x"));
-    auto [forwardMap, inverseMap] =
-        ttmlir::d2m::utils::grids::createCoreVirtMaps(
-            builder.getContext(), optimalGrid, physicalGridShape);
-    virtualGridInverseMapping = AffineMapAttr::get(inverseMap);
-    virtualGridForwardMapping = AffineMapAttr::get(forwardMap);
+  auto newToLayoutOp =
+      rebuildToLayoutWithType(toLayoutOp, newTensorType, config, builder);
+  if (!newToLayoutOp) {
+    return;
   }
-
-  auto newEmptyOp = builder.create<d2m::EmptyOp>(
-      emptyOp.getLoc(), newTensorType, virtualGridInverseMapping,
-      virtualGridForwardMapping);
-
-  builder.setInsertionPoint(toLayoutOp);
-  auto newToLayoutOp = builder.create<d2m::ToLayoutOp>(
-      toLayoutOp.getLoc(), toLayoutOp.getInput(), newEmptyOp);
 
   // Reblock it back to original shape to preserve IR correctness.
   // The view chain that applyViews composes through depends on this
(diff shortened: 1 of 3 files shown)


lib/Dialect/D2M/Transforms/GridSelection.cpp
```diff
@@ -838,29 +849,76 @@ updateCompositeViewOps(ArrayRef<CompositeViewUpdateInfo> compositeViewsToUpdate,
...
+      auto tileShape = tileType.getShape();
+
+      // Derive input dim alignments from the output, overriding the concat
+      // dim with tile-only alignment so each input's contribution isn't
+      // independently inflated.
+      llvm::SmallVector<int64_t> inputAlignments(outDimAlignments.begin(),
+                                                 outDimAlignments.end());
+      int64_t logicalRank = inputLayout.getLogicalShape().size();
+      int64_t concatLogicalDim =
+          concatDim < 0 ? logicalRank + concatDim : concatDim;
```
wenbinlyuTT: The TTIRToD2M pass should have already made sure the concatDim is in [0, rank).

lib/Dialect/D2M/Transforms/GridSelection.cpp
```diff
@@ -838,29 +849,76 @@ updateCompositeViewOps(ArrayRef<CompositeViewUpdateInfo> compositeViewsToUpdate,
...
+      if (concatLogicalDim >= 0 &&
```
wenbinlyuTT: ... and so this `if` is always true?

lib/Dialect/D2M/Transforms/GridSelection.cpp
```diff
@@ -838,29 +849,76 @@ updateCompositeViewOps(ArrayRef<CompositeViewUpdateInfo> compositeViewsToUpdate,
...
+          concatLogicalDim < static_cast<int64_t>(inputAlignments.size())) {
+        int64_t tileIdx = (concatLogicalDim >= logicalRank - 2)
```
wenbinlyuTT: You can just do the subtraction, the check below only compare against 0 anyway.

lib/Dialect/D2M/Transforms/GridSelection.cpp
```diff
@@ -838,29 +849,76 @@ updateCompositeViewOps(ArrayRef<CompositeViewUpdateInfo> compositeViewsToUpdate,
...
+                              ? concatLogicalDim - (logicalRank - 2)
+                              : -1;
+        inputAlignments[concatLogicalDim] =
+            (tileIdx >= 0) ? tileShape[tileIdx] : 1;
```
wenbinlyuTT: If the concat dim isn't height/width, should we set the alignment requirement to 1, 32, or leave as-is? I remember the alignment is cumulative and thus at least 32, not sure if this has changed lately...

This probably has no effect since the inputs are "passive", but I just want to make sure the numbers make sense.

wenbinlyuTT (commented):
Mostly LGTM, I only have one question about the inputs' alignments.

lib/Dialect/D2M/Transforms/GridSelection.cpp
dlokeTT: yea oops, ai slop. removed

lib/Dialect/D2M/Transforms/GridSelection.cpp
dlokeTT: removed

lib/Dialect/D2M/Transforms/GridSelection.cpp
dlokeTT: yea we should probably just keep it as 1 since that's the alignment requirement in other places.

wenbinlyuTT (approved):
Thanks for fixing this!

lib/Dialect/D2M/Transforms/GridSelection.cpp
```diff
@@ -820,13 +820,24 @@ updateCompositeViewOps(ArrayRef<CompositeViewUpdateInfo> compositeViewsToUpdate,
   OpBuilder builder(compositeViewsToUpdate.front().op->getContext());
   for (const auto &info : compositeViewsToUpdate) {
     auto compositeView = info.op;
+    int32_t concatDim = compositeView.getDim();
```
anuragsingh-tt, line 822: Let's use the composite view terminology instead of concat.

test/python/golden/ttir_ops/data_movement/test_data_movement.py
```diff
@@ -66,6 +66,17 @@
         ([(3, 2, 96, 64), (3, 2, 32, 64), (3, 2, 64, 64)], 2),
         ([(4, 2, 64, 64), (4, 1, 64, 64), (4, 3, 64, 64)], 1),
         ([(3, 3, 64, 64), (2, 3, 64, 64), (1, 3, 64, 64)], 0),
+        ##################################
+        #    Large tensors (multi-core)  #
+        ##################################
+        # These exercise the grid selection path where concat inputs are
+        # large enough that single-core allocation would overflow L1,
+        # requiring the to_layout ops to distribute data across cores.
```
anuragsingh-tt, line 74: This is 15 new tests, do they come from a model? This is more a compiler change so maybe just one test is needed here + the lit tests.

lib/Dialect/D2M/Transforms/GridSelection.cpp
```diff
@@ -838,29 +849,72 @@ updateCompositeViewOps(ArrayRef<CompositeViewUpdateInfo> compositeViewsToUpdate,
...
+      // When the input comes directly from a to_layout op with a single use,
+      // update the to_layout's grid so that data is physically distributed
+      // across multiple cores, preventing L1 overflow when multiple concat
+      // inputs must coexist in L1 on a single core.
+      if (auto toLayoutOp = input.getDefiningOp<d2m::ToLayoutOp>();
+          toLayoutOp && input.hasOneUse()) {
+        auto emptyOp = toLayoutOp.getOutput().getDefiningOp<d2m::EmptyOp>();
+        if (emptyOp) {
+          builder.setInsertionPoint(emptyOp);
+          auto newEmptyOp = builder.create<d2m::EmptyOp>(
```
anuragsingh-tt: This is kind of a one off rewrite, we already have `optimizeToLayoutGrid`. Can we unify the flows by refactoring the common logic?

lib/Dialect/D2M/Transforms/GridSelection.cpp
wenbinlyuTT: @anuragsingh-tt I will probably rename CompositeView to PiecewiseView or just ConcatView soon, this can leave as-is.

test/ttmlir/Dialect/D2M/Transforms/grid_selection_concat.mlir
```diff
@@ -0,0 +1,65 @@
...
+module attributes {ttcore.device = #any_device_3} {
+  func.func @concat_three_inputs(%arg0: tensor<1x32x128x64xbf16>, %arg1: tensor<1x32x128x64xbf16>, %arg2: tensor<1x32x128x64xbf16>) -> tensor<1x32x128x192xbf16> {
+    // CHECK-LABEL: func.func @concat_three_inputs
+    // All three to_layout ops should use multi-core grids (not 1x1x1x1).
+    // CHECK: d2m.to_layout %arg0, %{{.*}} : tensor<1x32x128x64xbf16> into tensor<1x8x4x2x
+    // CHECK: d2m.to_layout %arg1, %{{.*}} : tensor<1x32x128x64xbf16> into tensor<1x8x4x2x
+    // CHECK: d2m.to_layout %arg2, %{{.*}} : tensor<1x32x128x64xbf16> into tensor<1x8x4x2x
+    // CHECK: d2m.composite_view
+    // CHECK: d2m.generic
+    %0 = "ttir.concat"(%arg0, %arg1, %arg2) <{dim = 3 : si32}> : (tensor<1x32x128x64xbf16>, tensor<1x32x128x64xbf16>, tensor<1x32x128x64xbf16>) -> tensor<1x32x128x192xbf16>
```
anuragsingh-tt, line 62: How would outer permute (or any op that is just a view) + concat look?

test/python/golden/ttir_ops/data_movement/test_data_movement.py
dlokeTT: Some do. Overall, I don't think adding more tests is bad, just gives us more coverage. If CI time is getting too long maybe we should refactor tests and have some run in nightly and not run all the different pipelines but that is probably a bigger discussion and should be done separately.

test/ttmlir/Dialect/D2M/Transforms/grid_selection_concat.mlir
dlokeTT: This would just follow the old pattern where it would default the grid to `1x1`. None of the concats I've found in models are following tms so I just made this change as a quick optimization to unblock some ops. @wenbinlyuTT is still looking at concats overall and trying to create a more thorough and optimized solution

lib/Dialect/D2M/Transforms/GridSelection.cpp
dlokeTT: We could extract the empty + to_layout creation into a helper but the alignment logic is different from independent grid selection. Trying to keep this pr as a small fix to unblock the ops found in the models and not holistically change concat. @wenbinlyuTT will still be taking a look at concat

lib/Dialect/D2M/Transforms/GridSelection.cpp
anuragsingh-tt: Okay. I spoke to @wenbinlyuTT offline and he had mentioned `ConcatView` would be too narrow. We should align on the terms here and use that consistently throughout the codebase instead of interchanging Concat/CompositeView etc.

lib/Dialect/D2M/Transforms/GridSelection.cpp
anuragsingh-tt: > We could extract the empty + to_layout creation into a helper

+1. We should unify the rewrite machinery in this file. I would prefer handling it in this PR rather than cleaning it up later.

dlokeTT: updated

anuragsingh-tt (approved):

PR #7670, opened 2026-03-26
[D2M] Clean up scratch handling w/ new cb structure

vwellsTT:
### Ticket
Link to Github Issue

### Problem description
We have reworked the way streamed buffered are handled, but still pass in scratch allocs as explicit args.  This is inconsistent.

### What's changed
Treat scratch allocs as the same as local streaming allocation--create local memref.allocs inside generic, attach CBLayoutAttr during allocation and allocate real space, then eventually hoist out of the generic and pass via additionalArgs instead of operands.

### Checklist
- [ ] New/Existing tests provide coverage for changes

PR #7948, opened 2026-04-13
[d2m] scratch buffer refactor

ckaravasilisTT:
This PR continues and completes the work started in #7670 .

### What's changed
Scratch buffers are no longer modelled as optional input operands to `d2m.generics`. The first scratch pass `InsertScratchBuffers` (renamed) now adds a local `memref.alloc` with `CBLayoutAttr` inside the generic, along with a new `d2m.scratch_init` op consuming the memref. This new op survives until `LowerScratchAllocate`. The `memref.alloc` later gets hoisted out of the generic and passed via `additionalArgs`. Modified the `GlobalOperandRewriter` in `D2MToTTKernel` to use the `TypeConverer` for CB-backed memrefs.


include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
```diff
@@ -2138,58 +2138,29 @@ def D2M_ScratchAllocateOp : D2M_Op<"scratch_allocate", [
...
-    for use by downstream
-    scratch allocation ops (e.g. `d2m.scratch_allocate` lowered to
-    `memref.subview`).
+    Marks a `memref.alloc` as the scratch buffer for intermediate spilling
+    inside a `d2m.generic` region. This op consumes the memref to prevent
+    canonicalization from removing the otherwise-unused alloc.
 
-    The result type matches the CB's underlying memref type.
+    Inserted by `AddScratchInputs` and erased by `LowerScratchAllocate`
+    after `scratch_allocate` ops are replaced with subviews of this memref.
```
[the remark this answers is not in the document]
vwellsTT: this nit is good

lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
```diff
@@ -2156,16 +2156,32 @@ class D2MGetGlobalOperandRewriter
     func::FuncOp entry = op->getParentOfType<func::FuncOp>();
     ArgAttr arg;
     size_t argIndex;
-    Type arg_result_type;
+    Type argResultType;
+
+    if (auto memrefType =
```
vwellsTT, line 2161: I'm a bit confused why this special handling is needed vs previous behavior, do you recall? It seems like at this point the allocation is the same as any other (just an external alloc op which feed an additionalArg)

lib/Dialect/D2M/Transforms/InsertScratchBuffers.cpp
```diff
@@ -0,0 +1,144 @@
+// SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
```
vwellsTT: nit: year

vwellsTT (approved):
a couple small nits/questions, but this logic LGTM, thanks for taking this over!

lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
ckaravasilisTT: Iirc the issue was that the `get_global_operand` memref was a plain i32 buffer address, while the consumer ops expected `CBType`. This caused a materializations cast from i32 to ttyernel::CBType which would fail.

lib/Dialect/D2M/Transforms/InsertScratchBuffers.cpp
```diff
@@ -0,0 +1,144 @@
...
+  // Skip if not in compute-only form.
+  if (!genericOp.isComputeOnlyForm()) {
+    return failure();
+  }
+
+  // Only add scratch to fused generics with multiple linalg.generic ops,
+  // which indicates elementwise fusion produced a multi-op kernel whose
+  // intermediate results must be spilled to L1.
+  if (!needsScratch(genericOp)) {
+    return failure();
```
anuragsingh-tt: Shouldn't this and the one above it return success?

lib/Dialect/D2M/Transforms/InsertScratchBuffers.cpp
```diff
@@ -0,0 +1,144 @@
...
+  }
+
+  // Find a tiled input to get the tile type.
+  MemRefType refMemRefType = findTiledInputType(genericOp);
+  if (!refMemRefType) {
+    return failure();
```
anuragsingh-tt: What if needsScratch is true but !refMemRefType is also true?

lib/Dialect/D2M/Transforms/InsertScratchBuffers.cpp
```diff
@@ -0,0 +1,144 @@
...
+  }
+
+  ttcore::TileType tileType = getTileType(refMemRefType);
+  if (!tileType) {
+    return failure();
```
anuragsingh-tt: Is this reachable?

lib/Dialect/D2M/Transforms/InsertScratchBuffers.cpp
```diff
@@ -0,0 +1,144 @@
...
+      genericOp.getContext(), ttcore::MemorySpace::DeviceL1);
+
+  // Shard memref type for the scratch buffer inside the region.
+  MemRefType scratchShardMemRefType =
+      MemRefType::get(scratchShardShape, tileType, cbLayout, l1MemorySpace);
+
+  // Insert memref.alloc + scratch_init at the start of the region body.
+  Region &region = genericOp.getRegion(0);
+  if (region.empty()) {
+    return failure();
```
anuragsingh-tt: Is this reachable?

ckaravasilisTT: I don't think so. Given `needsScratch()` earlier, I think this can be safely removed.

lib/Dialect/D2M/Transforms/InsertScratchBuffers.cpp
ckaravasilisTT: Not really, this was a conservative check from the first version of the pass. I will remove.

lib/Dialect/D2M/Transforms/InsertScratchBuffers.cpp
ckaravasilisTT: That shouldn't really happen. I suppose an assertion might be more appropriate than silently skipping the scratch. Let me know if you have a preference.

lib/Dialect/D2M/Transforms/InsertScratchBuffers.cpp
ckaravasilisTT: The way `addScratchToGeneric` is called we basically ignore its return value, so it doesn't make a functional difference. Given the function name, I'm not sure if 'success' would be much better. Maybe just change the return type to void?

lib/Dialect/D2M/Transforms/InsertScratchBuffers.cpp
anuragsingh-tt: Sure, we can assert.

lib/Dialect/D2M/Transforms/InsertScratchBuffers.cpp
anuragsingh-tt: Sort of a NIT so I'm fine either way

PR #7949, opened 2026-04-13
[D2M] Initialize CB read pointers for scratch and intermediate output buffers in fused kernels

### LINKED ISSUE #7539: Single tile fusion cases fail because of unpack/pack sync

### LINKED ISSUE #7574: Reenable fusion tests for small shapes
- test/python/golden/test_ttir_eltwise_fusion.py
- test/python/golden/test_composite_functions.py

jdesousa-TT:
In fused compute kernels, scratch CBs and intermediate output CBs are written to (via PACK) and read from (via UNPACK) within the same compute block. Without push+wait, the CB read pointer is never initialized, so copy_tile addresses the wrong location.

For scratch CBs: insert reserve→push→wait at initialization (mirroring the input CB pattern) to set up the read pointer.

For aliased output CBs reused as intermediates: detect store-then-load patterns via alias analysis and insert push+wait before the first intermediate read to initialize the read pointer.

Re-enable elementwise fusion in test_ttir_eltwise_fusion.py and test_composite_functions.py, and un-skip the binary reduction tree test. Re-enable the 256x256 binary tree test case (issue #7539).

### Ticket
#7574 , #7539

### Checklist

### DIFF
--- a/lib/Dialect/D2M/Transforms/SplitUnifiedThread.cpp
+++ b/lib/Dialect/D2M/Transforms/SplitUnifiedThread.cpp
@@ -144,6 +144,76 @@ static void insertPopBeforeTerminator(PatternRewriter &rewriter, Location loc,
   rewriter.create<PopOp>(loc, cb);
 }
 
+// Build the set of values that alias `value` through view-like operations
+// in the given block. Includes `value` itself and any view-like results
+// derived from it.
+static llvm::SmallPtrSet<Value, 8> buildValueAliasSet(Value value,
+                                                      Block *block) {
+  llvm::SmallPtrSet<Value, 8> aliases;
+  aliases.insert(value);
+  bool changed = true;
+  while (changed) {
+    changed = false;
+    for (Operation &op : *block) {
+      if (!mlir::isa<mlir::ViewLikeOpInterface>(op)) {
+        continue;
+      }
+      for (OpOperand &operand : op.getOpOperands()) {
+        if (aliases.contains(operand.get())) {
+          for (Value result : op.getResults()) {
+            if (aliases.insert(result).second) {
+              changed = true;
+            }
+          }
+          break;
+        }
+      }
+    }
+  }
+  return aliases;
+}
+
+// Find all top-level operations in `block` that contain a memref.load from a
+// value in `aliases`, where a preceding top-level operation stored to those
+// aliases. Each such point is a PACK→L1→UNPACK transition that requires
+// push+wait to initialize the CB read pointer. In fused kernels the output CB
+// may be used as a ping-pong intermediate across many compute phases, producing
+// multiple transitions.
+// If `beforeOp` is non-null, only considers operations before it.
+static SmallVector<Operation *>
+findAllIntermediateReadPoints(const llvm::SmallPtrSet<Value, 8> &aliases,
+                              Block *block, Operation *beforeOp = nullptr) {
+  SmallVector<Operation *> readPoints;
+  bool hasStoredBefore = false;
+  for (Operation &op : *block) {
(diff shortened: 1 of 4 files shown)


lib/Dialect/D2M/Transforms/SplitUnifiedThread.cpp
```diff
@@ -144,6 +144,76 @@ static void insertPopBeforeTerminator(PatternRewriter &rewriter, Location loc,
   rewriter.create<PopOp>(loc, cb);
 }
 
+// Build the set of values that alias `value` through view-like operations
+// in the given block. Includes `value` itself and any view-like results
+// derived from it.
+static llvm::SmallPtrSet<Value, 8> buildValueAliasSet(Value value,
```
vwellsTT, line 150: I think this is a bit crude, since MLIR has utils to directly track users etc.  I think something like:
```
static llvm::SmallPtrSet<Value, 8> buildValueAliasSet(Value value, Block *block) {
  llvm::SmallPtrSet<Value, 8> aliases;
  llvm::SmallVector<Value, 8> worklist = {value};

  while (!worklist.empty()) {
    Value v = worklist.pop_back_val();
    if (!aliases.insert(v).second)
      continue;
    for (Operation *user : v.getUsers()) {
      if (user->getBlock() != block)
        continue;
      if (!mlir::isa<mlir::ViewLikeOpInterface>(user))
        continue;
      for (Value result : user->getResults())
        worklist.push_back(result);
    }
  }
  return aliases;
}
```
might be more efficient?

lib/Dialect/D2M/Transforms/SplitUnifiedThread.cpp
```diff
@@ -144,6 +144,76 @@ static void insertPopBeforeTerminator(PatternRewriter &rewriter, Location loc,
...
+  return aliases;
+}
+
+// Find all top-level operations in `block` that contain a memref.load from a
+// value in `aliases`, where a preceding top-level operation stored to those
+// aliases. Each such point is a PACK→L1→UNPACK transition that requires
+// push+wait to initialize the CB read pointer. In fused kernels the output CB
+// may be used as a ping-pong intermediate across many compute phases, producing
+// multiple transitions.
+// If `beforeOp` is non-null, only considers operations before it.
```
vwellsTT, line 182: this is never used anyway, right? probably should remove if not used

lib/Dialect/D2M/Transforms/SplitUnifiedThread.cpp
```diff
@@ -364,6 +434,26 @@ static LogicalResult processComputeStores(Block *computeBlock,
...
+        for (Operation *readPoint :
+             findAllIntermediateReadPoints(aliases, computeBlock, storeOp)) {
+          rewriter.setInsertionPoint(readPoint);
+          if (!firstTransition) {
+            rewriter.create<PopOp>(loc, cb);
+          }
+          rewriter.create<PushOp>(loc, cb);
+          rewriter.create<WaitOp>(loc, cb);
+          firstTransition = false;
+        }
```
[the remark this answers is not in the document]
vwellsTT: big if true

vwellsTT (approved):

xanderchin (approved):

test/python/golden/test_ttir_eltwise_fusion.py
vtangTT, line 382: are we missing `enable-elementwise-fusion=true` flag here?

vtangTT (approved):
would prefer if you used the suggested worklist approach for `buildValueAliasSet` but not a blocker

changes makes sense and lgtm ty 👍

PR #7959, opened 2026-04-14
[EmitPy] Separate forward inputs into activations and named weights

amilovanovicTT:
### Problem description
Generated Python code from the EmitPy pipeline treats all forward function arguments as a single, flat tuple, making it hard to distinguish activations from weights. 
### What's changed
This PR improves code readability by separating forward function inputs into activations and weights. Moreover, it packs the weights into a global dictionary where each weight is stored under a key corresponding to its fully qualified name. 

- `TTNNSplitForwardFuncArgsByType`: A pass introduced to rewrite a forward function signature to accept one tuple for activations and one dictionary for weights:
  - Argument type info is extracted from the `ttcore.argument_type` attribute, so all function arguments must carry this attribute for the pass to process a function. 
  - Weight dictionary keys are derived from `ttir.name` attributes when present, or auto-generated otherwise (e.g., `weight_0`). 

- `TTNNTuplifyTensors`: This pass remains unchanged, but the set of functions it processes is reduced. For functions with split input, it is now only responsible for tuplifying the result.
- `TTNNInputFunctionCreatorBase`: Now generates separate input generator functions for activations and weights when the forward function input is split.
- `TTNNPrepareConstEvalCaching`: Activations are not passed to the consteval wrapper functions anymore.
- `TTNNToEmitPy`: Conversion patterns for `SetKeyValueOp` and `GetKeyValueOp` are updated to avoid unnecessary list wrapping for single values.
- `EmitPyNameVars`: Activations and weights names are now inferred from the preserved `ttcore.original_activation/weight_names` attributes.

_Note:_ This is relevant only for the canonical Python codegen path. Additionally, this will be enabled for the canonical C++ codegen path. Other codegen paths (EmitC dylib and EmitPy target module) remain unchanged. 

### Checklist

### DIFF
--- a/include/ttmlir/Dialect/TTNN/Transforms/Passes.td
+++ b/include/ttmlir/Dialect/TTNN/Transforms/Passes.td
@@ -178,6 +178,37 @@ def TTNNCreateMainForTest: Pass<"ttnn-create-main-for-test", "::mlir::ModuleOp">
   }];
 }
 
+def TTNNSplitActivationsAndWeights : Pass<"ttnn-split-activations-and-weights", "::mlir::ModuleOp"> {
+  let summary = "Split forward function arguments into activations tuple and weights dictionary.";
+  let description = [{
+    This pass rewrites each forward function of device module by separating its tensor arguments into activations and weights.
+    Activation arguments are packed into a tuple, while weight arguments are packed into a dictionary.
+
+    Given a forward function like this:
+
+    ```mlir
+    func.func @forward(%arg0: tensor<32x32xbf16> {ttcore.argument_type = #ttcore.argument_type<input>},
+                       %arg1: tensor<32x32xbf16> {ttcore.argument_type = #ttcore.argument_type<parameter>})
+        -> tensor<32x32xbf16> {
+      ...
+    }
+    ```
+
+    The pass will produce:
+
+    ```mlir
+    func.func @forward(%arg0: tuple<tensor<32x32xbf16>>,
+                       %arg1: !ttcore.dict)
+        -> tensor<32x32xbf16>
+        attributes {ttcore.original_argument_types = [...]} {
+      %input0 = ttcore.get_tuple_element %arg0[0] : ...
+      %weight0 = ttcore.get_key_value %arg1["weight_0"] : ...
+      ...
+    }
+    ```
+  }];
+}
+
 def TTNNTuplifyTensors: Pass<"ttnn-tuplify-tensors", "::mlir::ModuleOp"> {
   let summary = "Tuplify tensors in all funcs within a module.";
   let description = [{
@@ -551,7 +582,7 @@ def TTNNPrepareConstEvalCaching : Pass<"ttnn-prepare-const-eval-caching", "::mli
   let description = [{
     Packs the results of all LoadCachedOps into a single global caching
     dictionary per forward function.  For each forward function containing
-    LoadCachedOps it creates a global dictionary (e.g. `_cached_forward`),
+    LoadCachedOps it creates a global dictionary (e.g. `ce_cache_forward`),
     retrieves it at the top of the function body, and stores each
     LoadCachedOp result under its callee name.
   }];
--- a/include/ttmlir/Dialect/TTNN/Transforms/Passes.td
+++ b/include/ttmlir/Dialect/TTNN/Transforms/Passes.td
@@ -178,7 +178,7 @@ def TTNNCreateMainForTest: Pass<"ttnn-create-main-for-test", "::mlir::ModuleOp">
   }];
 }
 
-def TTNNSplitActivationsAndWeights : Pass<"ttnn-split-activations-and-weights", "::mlir::ModuleOp"> {
+def TTNNSplitForwardFuncArgsByType : Pass<"ttnn-split-forward-func-args-by-type", "::mlir::ModuleOp"> {
   let summary = "Split forward function arguments into activations tuple and weights dictionary.";
   let description = [{
     This pass rewrites each forward function of device module by separating its tensor arguments into activations and weights.
--- a/include/ttmlir/Dialect/TTNN/Transforms/Passes.td
+++ b/include/ttmlir/Dialect/TTNN/Transforms/Passes.td
@@ -178,6 +178,37 @@ def TTNNCreateMainForTest: Pass<"ttnn-create-main-for-test", "::mlir::ModuleOp">
   }];
 }
 
+def TTNNSplitActivationsAndWeights : Pass<"ttnn-split-activations-and-weights", "::mlir::ModuleOp"> {
+  let summary = "Split forward function arguments into activations tuple and weights dictionary.";
+  let description = [{
+    This pass rewrites each forward function of device module by separating its tensor arguments into activations and weights.
+    Activation arguments are packed into a tuple, while weight arguments are packed into a dictionary.
+
+    Given a forward function like this:
+
+    ```mlir
+    func.func @forward(%arg0: tensor<32x32xbf16> {ttcore.argument_type = #ttcore.argument_type<input>},
+                       %arg1: tensor<32x32xbf16> {ttcore.argument_type = #ttcore.argument_type<parameter>})
+        -> tensor<32x32xbf16> {
+      ...
+    }
+    ```
+
+    The pass will produce:
+
+    ```mlir
+    func.func @forward(%arg0: tuple<tensor<32x32xbf16>>,
+                       %arg1: !ttcore.dict)
+        -> tensor<32x32xbf16>
+        attributes {ttcore.original_argument_types = [...]} {
+      %input0 = ttcore.get_tuple_element %arg0[0] : ...
+      %weight0 = ttcore.get_key_value %arg1["weight_0"] : ...
+      ...
+    }
+    ```
+  }];
+}
+
 def TTNNTuplifyTensors: Pass<"ttnn-tuplify-tensors", "::mlir::ModuleOp"> {
   let summary = "Tuplify tensors in all funcs within a module.";
   let description = [{
@@ -568,7 +599,7 @@ def TTNNPrepareConstEvalCaching : Pass<"ttnn-prepare-const-eval-caching", "::mli
   let description = [{
     Packs the results of all LoadCachedOps into a single global caching
     dictionary per forward function.  For each forward function containing
-    LoadCachedOps it creates a global dictionary (e.g. `_cached_forward`),
+    LoadCachedOps it creates a global dictionary (e.g. `ce_cache_forward`),
     retrieves it at the top of the function body, and stores each
     LoadCachedOp result under its callee name.
   }];
--- a/include/ttmlir/FunctionTypes.h
+++ b/include/ttmlir/FunctionTypes.h
@@ -95,6 +95,10 @@ constexpr inline llvm::StringLiteral kImportedDeclarationValue =
 
 /// Attribute name for the source file of an imported declaration.
 constexpr inline llvm::StringLiteral kImportedFromAttrName = "tt.imported_from";
+
+/// Attribute name marking whether a forward device function inputs were
+/// split into activations and weights by TTNNSplitActivationsAndWeights.
+constexpr inline llvm::StringLiteral kSplitInputAttrName = "tt.is_split_input";
 } // namespace detail
 
 /// Returns the string value for the given function type.
@@ -198,13 +202,24 @@ inline void clearFunctionType(mlir::func::FuncOp funcOp) {
   funcOp->removeAttr(detail::kFunctionTypeAttrName);
 }
 
+/// Removes the split input attribute from the given function.
+inline void clearSplitInput(mlir::func::FuncOp funcOp) {
+  funcOp->removeAttr(detail::kSplitInputAttrName);
+}
+
 //===----------------------------------------------------------------------===//
 // Convenience query functions
 //===----------------------------------------------------------------------===//
 
-/// Returns true if the function is marked as a forward device function.
+/// Returns true if the function is a forward device function.
 inline bool isForwardDeviceFunc(mlir::func::FuncOp funcOp) {
   return hasFunctionType(funcOp, FunctionType::ForwardDevice);
+}
+
+/// Returns true if the function's inputs were split into activations and
+/// weights.
+inline bool isSplitInput(mlir::func::FuncOp funcOp) {
+  return funcOp->hasAttr(detail::kSplitInputAttrName);
 }
 
 /// Returns true if the function is marked as a forward CPU function.
@@ -274,6 +289,13 @@ inline void setImportedFrom(mlir::func::FuncOp funcOp,
                             llvm::StringRef fileName) {
   funcOp->setAttr(detail::kImportedFromAttrName,
                   mlir::StringAttr::get(funcOp->getContext(), fileName));
+}
+
+/// Marks a forward device function as having split inputs by
+/// TTNNSplitActivationsAndWeights.
+inline void setSplitInput(mlir::func::FuncOp funcOp) {
+  funcOp->setAttr(detail::kSplitInputAttrName,
+                  mlir::UnitAttr::get(funcOp->getContext()));
 }
 
 /// Gets the source file attribute from an imported declaration function.
--- a/include/ttmlir/FunctionTypes.h
+++ b/include/ttmlir/FunctionTypes.h
@@ -96,9 +96,10 @@ constexpr inline llvm::StringLiteral kImportedDeclarationValue =
 /// Attribute name for the source file of an imported declaration.
 constexpr inline llvm::StringLiteral kImportedFromAttrName = "tt.imported_from";
 
-/// Attribute name marking whether a forward device function inputs were
-/// split into activations and weights by TTNNSplitActivationsAndWeights.
-constexpr inline llvm::StringLiteral kSplitInputAttrName = "tt.is_split_input";
+/// Attribute name marking whether a forward device function arguments were
+/// split into activations and weights by TTNNSplitForwardFuncArgsByType.
+constexpr inline llvm::StringLiteral kSplitForwardFuncArgsByTypeAttrName =
+    "tt.has_split_func_args_by_type";
 } // namespace detail
 
 /// Returns the string value for the given function type.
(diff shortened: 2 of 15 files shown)


lib/Dialect/TTNN/Transforms/TTNNPrepareConstEvalCaching.cpp
```diff
@@ -147,18 +154,38 @@ class TTNNPrepareConstEvalCaching
...
   }
 
+  /// Return the index of the activations argument if it exists.
+  std::optional<unsigned> getActivationArgIdx(func::FuncOp funcOp) {
+    if (!ttmlir::utils::isSplitInput(funcOp)) {
+      return std::nullopt;
+    }
+
+    for (unsigned i = 0; i < funcOp.getNumArguments(); ++i) {
+      if (isa<TupleType>(funcOp.getArgument(i).getType())) {
```
dmilinkovicTT, line 164: Is there a way we could be a bit more explicit with detecting the activations? 

This could potentially break if we end up introducing another tuple argument.

lib/Dialect/TTNN/Transforms/TTNNPrepareConstEvalCaching.cpp
```diff
@@ -147,18 +154,38 @@ class TTNNPrepareConstEvalCaching
...
-                           forwardFunc.getArgumentTypes().end());
+
+    llvm::SmallVector<Type> wrapperArgTypes = {dictType};
+    auto activationIdx = getActivationArgIdx(forwardFunc);
+    for (unsigned i = 0; i < forwardFunc.getNumArguments(); ++i) {
+      if (activationIdx && *activationIdx == i) {
+        continue;
+      }
+      wrapperArgTypes.push_back(forwardFunc.getArgument(i).getType());
+    }
```
dmilinkovicTT, line 187: Could we pass the weights dictionary to the const-eval wrapper (if the split was performed), instead of passing each argument separately?

I assume this would simplify some changes in this file (we could get ride of the for loop which skips the activations argument and which is repeated quite a few times), while making the generated code more readable.

lib/Dialect/TTNN/Transforms/Passes.cpp
```diff
@@ -263,8 +264,34 @@ class TTNNDeallocate : public impl::TTNNDeallocateBase<TTNNDeallocate> {
   }
 };
 
+/// Recovers the original flat argument indices for activations and weights
+/// from the ttcore.original_argument_types attribute. This is needed to
```
dmilinkovicTT, line 268: The comment seems to be cut-out.

lib/Dialect/TTNN/Transforms/Passes.cpp
```diff
@@ -289,115 +316,199 @@ class TTNNInputFunctionCreatorBase {
         rewriter.modifyOpInPlace(funcOp, [&]() { funcOp.setSymName("_main"); });
       }
 
-      if (funcOp.isPrivate()) {
+      if (!ttmlir::utils::isForwardDeviceFunc(funcOp) || funcOp.isPrivate()) {
         return mlir::WalkResult::skip();
       }
```
dmilinkovicTT, line 321: Do we ever have private forward device functions?

include/ttmlir/FunctionTypes.h
```diff
@@ -276,6 +291,13 @@ inline void setImportedFrom(mlir::func::FuncOp funcOp,
                   mlir::StringAttr::get(funcOp->getContext(), fileName));
 }
 
+/// Marks a forward device function as having split inputs by
+/// TTNNSplitActivationsAndWeights.
+inline void setSplitInput(mlir::func::FuncOp funcOp) {
+  funcOp->setAttr(detail::kSplitInputAttrName,
```
dmilinkovicTT, line 297: We should either enforce that the `funcOp` is indeed a forward device function, or modify the comment above.

I prefer the first approach.

lib/Dialect/TTNN/Pipelines/TTNNPipelines.cpp
```diff
@@ -530,8 +530,11 @@ void createTTNNToEmitPyDevicePipeline(
     devicePm.addPass(createTTNNTuplifyTensors(tuplifyOptions));
     devicePm.addPass(createTTNNPrepareModuleForExport());
   } else {
-    // In canonical path, run tuplification + input generation/loading.
+    // In canonical path, first split forward functions inputs that have
+    // ttcore.argument_type on all args into activations and weights. Then
+    // tuplify everything else.
     //
+    devicePm.addPass(createTTNNSplitActivationsAndWeights());
```
dmilinkovicTT, line 537: A bit of a philosophical topic, but is this really the right place in the pipeline for this transformation?

On one hand, performing the split is significantly easier before other transformations take place, and I assume this was the main reason behind choosing this point in the pipeline.

But, on the other hand - tuplification, input function generation, and many other potential future transformations would need to branch their implementations to cover both split and non-split inputs.

There probably isn't a single correct answer for this, but IMHO, this position would probably make more sense if we were to perform the split on both EmitPy targets, because it would eliminate branching in downstream passes. 

In the current setup, having this pass execute a bit later in the pipeline would clearly communicate that this transformation as a nice-to-have prettification transformation rather than a core functionality.

Wdyt?

dmilinkovicTT (commented):
Nice work! This is the first batch of comments, I haven't been focusing that much on low-level details, I'll do that in a follow-up pass.

lib/Conversion/TTNNToEmitPy/EmitPyNameVars.cpp
```diff
@@ -21,6 +23,39 @@ namespace mlir::tt {
...
+    auto typeAttr = mlir::dyn_cast<ttcore::ArgumentTypeAttr>(attr);
+    assert(typeAttr &&
+           "Expected attribute to be of type ttcore::ArgumentTypeAttr");
+    if (typeAttr.getValue() == ttcore::ArgumentType::Input) {
+      hasActivations = true;
+    } else if (typeAttr.getValue() == ttcore::ArgumentType::Parameter ||
+               typeAttr.getValue() == ttcore::ArgumentType::Constant) {
+      hasWeights = true;
+    }
+  }
```
svuckovicTT, line 46: Would it make sense to catch other arg types here as well, and error out? What happens if there are other args but they aren't of these types, does the whole flow work?

lib/Conversion/TTNNToEmitPy/EmitPyNameVars.cpp
```diff
@@ -21,6 +23,39 @@ namespace mlir::tt {
...
+
+  if (hasActivations && argIdx == 0) {
```
svuckovicTT: Do we have a guarantee that they'll always be at idx=0? Also, there can be more than 1 activation tensor for some models.

lib/Conversion/TTNNToEmitPy/EmitPyNameVars.cpp
```diff
@@ -90,13 +125,12 @@ class EmitPyNameVarsPass : public impl::EmitPyNameVarsBase<EmitPyNameVarsPass> {
...
                 funcOp.getArgAttrOfType<StringAttr>(i, "emitpy.name")) {
           argName = existingNameAttr.getValue().str();
-        } else if (isa<emitpy::DictType>(funcOp.getArgument(i).getType())) {
-          argName = "ce_cache";
+        } else if (auto inferredName =
+                       inferArgNameFromOriginalTypes(funcOp, i)) {
+          argName = *inferredName;
         } else {
           argName = funcOp.getNumArguments() > 1 ? "input_" + std::to_string(i)
                                                  : "input";
```
svuckovicTT: Maybe we should change these `input_`s to `arg_`s

include/ttmlir/Dialect/TTNN/Transforms/Passes.td
```diff
@@ -178,6 +178,37 @@ def TTNNCreateMainForTest: Pass<"ttnn-create-main-for-test", "::mlir::ModuleOp">
   }];
 }
 
+def TTNNSplitActivationsAndWeights : Pass<"ttnn-split-activations-and-weights", "::mlir::ModuleOp"> {
```
svuckovicTT: Would prefer a more descriptive name here, something like `SplitForwardFuncArgsByType`

include/ttmlir/FunctionTypes.h
```diff
@@ -95,6 +95,10 @@ constexpr inline llvm::StringLiteral kImportedDeclarationValue =
 
 /// Attribute name for the source file of an imported declaration.
 constexpr inline llvm::StringLiteral kImportedFromAttrName = "tt.imported_from";
+
+/// Attribute name marking whether a forward device function inputs were
+/// split into activations and weights by TTNNSplitActivationsAndWeights.
+constexpr inline llvm::StringLiteral kSplitInputAttrName = "tt.is_split_input";
```
svuckovicTT: Similar to pass name, rename to something more accurate/descriptive.

include/ttmlir/FunctionTypes.h
svuckovicTT: +1 for erroring out if not a fwd func.

lib/Dialect/TTNN/Transforms/Passes.cpp
```diff
@@ -811,6 +918,203 @@ class TTNNCreateMainForTest
...
+      //
+      entryBlock.eraseArguments(newArgsCount, oldFunctionType.getNumInputs());
+
+      // Mark the function input as split.
+      //
+      ttmlir::utils::setSplitInput(funcOp);
+    }
+  }
+
+  bool isMissingArgTypes(func::FuncOp funcOp) {
```
svuckovicTT: Consider inverting to `containsArgTypes`

lib/Dialect/TTNN/Transforms/Passes.cpp
```diff
@@ -811,6 +918,203 @@ class TTNNCreateMainForTest
...
+    return llvm::any_of(
+        llvm::seq<unsigned>(0, funcOp.getFunctionType().getNumInputs()),
+        [&](unsigned i) {
+          return funcOp.getArgAttrOfType<ttcore::ArgumentTypeAttr>(
+                     i, ttcore::ArgumentTypeAttr::name) == nullptr;
+        });
+  }
+
+  void
+  classifyFuncArguments(func::FuncOp funcOp,
```
svuckovicTT: This fn is quite confusing, some items are index-dependant, others aren't. I think that originalarg types and names should be handled in another fn.

lib/Dialect/TTNN/Transforms/Passes.cpp
```diff
@@ -811,6 +918,203 @@ class TTNNCreateMainForTest
...
+        } else {
+          assert(weightArgIndices.count(idx) && "Expected weight argument!");
+          getElement = rewriter
+                           .create<ttcore::GetKeyValueOp>(
+                               funcOp.getLoc(), originalType, weightsDict,
+                               originalArgNames[weightsNameIdx++])
+                           ->getResult(0);
+        }
+
+        rewriter.replaceAllUsesWith(entryBlock.getArgument(newArgsCount + idx),
```
svuckovicTT, line 1081: Is there a method that would search only within the function currently being addressed?

lib/Dialect/TTNN/Transforms/Passes.cpp
```diff
@@ -811,6 +918,203 @@ class TTNNCreateMainForTest
...
+                           .create<ttcore::GetTupleElementOp>(
+                               funcOp.getLoc(), originalType, activationsTuple,
+                               activationsTupleIdx++)
+                           ->getResult(0);
+        } else {
+          assert(weightArgIndices.count(idx) && "Expected weight argument!");
+          getElement = rewriter
+                           .create<ttcore::GetKeyValueOp>(
+                               funcOp.getLoc(), originalType, weightsDict,
+                               originalArgNames[weightsNameIdx++])
```
svuckovicTT: As discussed offline, let's keep `originalArgNames` and `originalArgTypes` for both activations and weights (basically all function args) and attach these to each func arg separately (tuple of acts, dict of weights, set of whatevers), if possible.

lib/Conversion/TTNNToEmitPy/EmitPyNameVars.cpp
```diff
@@ -90,13 +125,12 @@ class EmitPyNameVarsPass : public impl::EmitPyNameVarsBase<EmitPyNameVarsPass> {
...
 
-        // Check if an emitpy.name attribute already exists for this argument.
-        // If so, use it instead of generating a new name.
         if (auto existingNameAttr =
                 funcOp.getArgAttrOfType<StringAttr>(i, "emitpy.name")) {
           argName = existingNameAttr.getValue().str();
-        } else if (isa<emitpy::DictType>(funcOp.getArgument(i).getType())) {
-          argName = "ce_cache";
+        } else if (auto inferredName =
+                       inferArgNameFromOriginalTypes(funcOp, i)) {
```
svuckovicTT: After the suggested changes that original arg names/types should match the length of args for both the activations tuple and weights dict, can we random access orig types by index and remove this method?

svuckovicTT (commented):
Left some comments. I might've misunderstood some things, will take another look after we've simplified.

lib/Dialect/TTNN/Transforms/TTNNPrepareConstEvalCaching.cpp
amilovanovicTT: I added the corresponding `ttcore.original_activation_names`/`ttcore.original_weight_names` array attribute to the activation/weight argument to preserve the original `ttir.names`, so now I used that to identify the weights.

include/ttmlir/FunctionTypes.h
amilovanovicTT: Added an assert that the funcOp is of forward device type.

lib/Dialect/TTNN/Transforms/Passes.cpp
amilovanovicTT: Yes, in `TTNNRecoverStructure`, e.g., when `_main` is split into private forward-device helper funcs.

lib/Dialect/TTNN/Pipelines/TTNNPipelines.cpp
amilovanovicTT: Synced offline: Will preserve the current place in the pipeline.

include/ttmlir/FunctionTypes.h
amilovanovicTT: Changed to `kSplitForwardFuncArgsByTypeAttrName = "tt.has_split_func_args_by_type"`.

lib/Dialect/TTNN/Transforms/Passes.cpp
amilovanovicTT: Fixed

lib/Dialect/TTNN/Transforms/Passes.cpp
amilovanovicTT: I kept `originalArgNames` separately for activations/weights as an argument-level array attribute and `originalArgTypes ` as a function-level array attribute.

lib/Dialect/TTNN/Transforms/Passes.cpp
amilovanovicTT: `replaceAllUsesWith` on a `BlockArgument` is already function-scoped since it just rewires the use-list of that value, and block arg's uses can only live inside that function.

lib/Conversion/TTNNToEmitPy/TTNNToEmitPy.cpp
```diff
@@ -3003,26 +3021,53 @@ class FuncOpConversionPattern
...
         }
+        if (originalActivationNames[i]) {
+          funcOp.setArgAttr(i, ttcore::g_originalActivationNamesAttrName,
+                            originalActivationNames[i]);
+        }
+        if (originalWeightNames[i]) {
+          funcOp.setArgAttr(i, ttcore::g_originalWeightNamesAttrName,
+                            originalWeightNames[i]);
+        }
+      }
```
svuckovicTT, line 3062: confirm logic + add comments

lib/Dialect/TTNN/Transforms/Passes.cpp
```diff
@@ -263,8 +265,35 @@ class TTNNDeallocate : public impl::TTNNDeallocateBase<TTNNDeallocate> {
...
+                                      SmallVector<size_t> &weightIndices) {
+  auto origTypesAttr =
+      funcOp->getAttrOfType<ArrayAttr>("ttcore.original_argument_types");
+  assert(origTypesAttr && "Expected ttcore.original_argument_types on function "
+                          "that has split arguments by type!");
+  for (unsigned i = 0; i < origTypesAttr.size(); ++i) {
+    auto typeAttr = mlir::cast<ttcore::ArgumentTypeAttr>(origTypesAttr[i]);
+    if (typeAttr.getValue() == ttcore::ArgumentType::Input) {
+      activationIndices.push_back(i);
+    } else {
```
svuckovicTT: Confirm arg type is `ttcore::ArgumentType::Parameter/Constant` and error out if other type

lib/Dialect/TTNN/Transforms/Passes.cpp
svuckovicTT: Add comment above as to why we're skipping fwd&&private fns

svuckovicTT (approved):

lib/Conversion/TTNNToEmitPy/TTNNToEmitPy.cpp
```diff
@@ -3003,26 +3021,63 @@ class FuncOpConversionPattern
...
 
-      // Restore emitpy.name attributes.
+      // Restore preserved attributes.
       for (unsigned i = 0; i < funcOp.getNumArguments(); ++i) {
         if (emitPyNames[i]) {
           funcOp.setArgAttr(i, ttnn_to_emitpy::kNameAttr, emitPyNames[i]);
         }
       }
+      if (activationNamesAttr) {
+        funcOp.setArgAttr(activationNamesAttrIdx,
```
vmilosevic, line 3054: :warning: **clang\-analyzer\-core\.CallAndMessage** :warning:
1st function call argument is an uninitialized value

lib/Conversion/TTNNToEmitPy/TTNNToEmitPy.cpp
```diff
@@ -3003,26 +3021,63 @@ class FuncOpConversionPattern
...
+                          ttcore::g_originalActivationNamesAttrName,
+                          activationNamesAttr);
+      }
+      if (weightNamesAttr) {
+        funcOp.setArgAttr(weightNamesAttrIdx,
```
vmilosevic, line 3059: :warning: **clang\-analyzer\-core\.CallAndMessage** :warning:
1st function call argument is an uninitialized value

vmilosevic (commented):

Issue #7474, opened 2026-03-12
Greedy optimizer: Conv2d/ConvTranspose2d picks arbitrary sharding when LayoutScores tie


rpavlovicTT:
## Problem

When the greedy optimizer scores Conv2d/ConvTranspose2d candidates, block_sharded and height_sharded layouts at the same core count and L1 usage produce identical `LayoutScore` values. The final choice depends on arbitrary sort order, which can pick block_sharded when height_sharded would be significantly faster at runtime.

This was observed on UNet where two Conv2d ops account for **+980 us (+89%)** of the total Conv2d device time regression vs baseline.

## Case 1: Block_sharded input → Conv2d output tie

**Op 26** (`conv_transpose2d`, ICh=256 OCh=128 K=2x2 S=2x2):
- Input: block_sharded 8x8 (from previous conv2d)
- Backend validates both BS 8x8 (64 cores, 32768 l1use) and HS 64x1 (64 cores, 32768 l1use) — identical LayoutScores
- Greedy picks BS arbitrarily; baseline uses HS
- **Runtime: BS = 428 us, HS = 103 us (4.2x slower)**

## Case 2: L1-interleaved input → Conv2d output tie

**Op 28** (`conv2d`, ICh=256 OCh=128 K=3x3 S=1x1):
- Input: L1-interleaved 8x8 (from concat — sharded concat disabled due to tt-metal #39419)
- Backend validates both BS 8x8 and HS 64x1 at same score
- Greedy picks BS; baseline explicitly reshards concat output to HS 64x1 before conv2d
- **Runtime: BS = 875 us, HS = 220 us (4x slower)**

In both cases the compiler sets `act_block_h_override = 0` (auto), but tt-metal runtime picks ABH=64 for block_sharded vs ABH=8 for height_sharded, leading to dramatically different kernel performance.

## Proposed Fix

Extend `preferCandidate()` in `OpModelStrategy.cpp` to prefer height_sharded over block_sharded for Conv2d/ConvTranspose2d when LayoutScores are equal. This is a tiebreaker only — it does not change scoring priority, just breaks ties in favor of HS which consistently produces better runtime configs.

## Files

- `lib/Dialect/TTNN/Analysis/OpModelStrategy.cpp` — `preferCandidate()`
- `lib/Dialect/TTNN/Analysis/LayoutPropagation.cpp` — sort using `preferCandidate`

## Perf Data

| Op | Baseline (HS) | Greedy (BS) | Delta |
[...]

PR #7960, opened 2026-04-15
Implement height sharded preference for #7474

JonathanQ-Maker:
### Ticket
PR for #7474 

### Problem description
According to #7474, when the greedy optimizer scores Conv2d/ConvTranspose2d ``HeightSharded`` and ``BlockSharded`` candidates with the same score, the tie breaker depends on arbitrary sorting order. Even though it has been empirically observed that ``HeightSharded`` consistently produces significantly faster runtime results.

### What's changed
It was proposed that a tie breaker be implemented for ``conv2d`` and ``conv_transpose2d`` that prefers ``HeightSharded`` layouts. This has been implemented here.

- Added tie breaker logic to ``Conv2dRuleBook``'s ``preferCandidate()`` that prefers ``HeightSharded`` layout over all others. As suggested by original issue.
- Added unit-test to confirm ``HeightSharded`` is preferred over all other types.

### Checklist

### DIFF
--- a/lib/Dialect/TTNN/Analysis/OpRules/ConvRules.cpp
+++ b/lib/Dialect/TTNN/Analysis/OpRules/ConvRules.cpp
@@ -60,6 +60,30 @@ static uint32_t getActBlockHOverride(const BeamCandidate &c) {
   return UINT32_MAX;
 }
 
+static std::optional<TensorMemoryLayout>
+getTensorMemoryLayout(const BeamCandidate &c) {
+  // First try conv2dConfig's shard_layout if explicitly set
+  if (auto *conv2d =
+          std::get_if<Conv2dAttrs>(&c.configHint.opSpecificAttrs)) {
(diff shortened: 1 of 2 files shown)


rpavlovicTT (commented):
Unfortunately, it's not as simple as just picking HS over BS. We will have to add a certain heuristics for choosing best sharding strategy based on tensor shape (wide/narrow/rectangular). This requires some empirical data from TT HW. I will investigate this issue little bit more, and update the issue.

JonathanQ-Maker:
I see, thank you for the update. If you identified a fix would be awesome if you referenced it such that I can get an update.

rpavlovicTT:
@JonathanQ-Maker FYI I did make some improvements in conv rules [026940fee6f2343b2f0516fce5a2efc7562b26a9](https://github.com/tenstorrent/tt-mlir/commit/026940fee6f2343b2f0516fce5a2efc7562b26a9). So I will close the #7474 and this PR as well.

PR #7974, opened 2026-04-15
[EraseInverseOps] Add TTIRCommuteReshapeThroughReduce pattern

rpavlovicTT:
## Summary
- Add `TTIRCommuteReshapeThroughReduce` pattern to the EraseInverseOps pass, enabling reshape ops to commute through reduce ops (both upward and downward)
- Uses stride-matching to map reduce dimensions between original and reshaped tensor shapes, rejecting cases where reshape splits or merges the reduce dimension

## Performance
- n150 benchmarks: https://github.com/tenstorrent/tt-xla/actions/runs/24445828995 (all pass except gemma — pre-existing issue fixed by 8b78aede6d)
- n300-llmbox benchmarks: https://github.com/tenstorrent/tt-xla/actions/runs/24459002367

### DIFF
--- a/lib/Dialect/TTIR/Transforms/EraseInverseOps/ReduceCommutePatterns.cpp
+++ b/lib/Dialect/TTIR/Transforms/EraseInverseOps/ReduceCommutePatterns.cpp
@@ -269,41 +269,6 @@ private:
     return true;
   }
 };
-// Find the corresponding reduce dimension in a reshaped tensor by matching
-// stride and reduction length. Returns -1 if the reduce dimension was split
-// or merged by the reshape (commute not possible).
-static int64_t findReduceDimInReshapedTensor(llvm::ArrayRef<int64_t> fromShape,
-                                             llvm::ArrayRef<int64_t> toShape,
-                                             int64_t reduceDim) {
-  int64_t fromRank = fromShape.size();
-  int64_t toRank = toShape.size();
-
-  if (reduceDim < 0) {
-    reduceDim += fromRank;
-  }
-
-  assert(reduceDim >= 0 && reduceDim < fromRank && "Invalid reduce dimension");
-
-  int64_t reductionLength = fromShape[reduceDim];
-
-  // Compute stride of reduceDim in fromShape.
-  int64_t fromStride = 1;
-  for (int64_t i = reduceDim + 1; i < fromRank; ++i) {
-    fromStride *= fromShape[i];
-  }
-
-  // Find a dimension in toShape with the same stride and length.
-  int64_t toStride = 1;
-  for (int64_t dim = toRank - 1; dim >= 0; dim--) {
-    if (toStride == fromStride && toShape[dim] == reductionLength) {
-      return dim;
-    }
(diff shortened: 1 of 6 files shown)


lib/Dialect/TTIR/Transforms/EraseInverseOps/ReduceCommutePatterns.cpp
```diff
@@ -269,6 +269,222 @@ class TTIRCommutePermuteThroughReduce
...
+  int64_t reductionLength = fromShape[reduceDim];
+
+  // Compute stride of reduceDim in fromShape.
+  int64_t fromStride = 1;
+  for (int64_t i = reduceDim + 1; i < fromRank; ++i) {
+    fromStride *= fromShape[i];
+  }
+
+  // Find a dimension in toShape with the same stride and length.
+  int64_t toStride = 1;
```
mvasiljevicTT, line 296: There is similar logic for this in softmax commute pattern and will probably be needed for more patterns as we add them so it would be nice if we had some utils function for that. 
I wrote [findMatchingDimRTL](https://github.com/tenstorrent/tt-mlir/blob/9ff1f65b1595c264c25c7a6f1cb46b050650ea4b/lib/Dialect/TTIR/Utils/Utils.cpp#L67) with this idea in mind some time ago, but it didn't have usage at the moment, can it be used in this case directly or with some change?

lib/Dialect/TTIR/Transforms/EraseInverseOps/ReduceCommutePatterns.cpp
```diff
@@ -269,6 +269,222 @@ class TTIRCommutePermuteThroughReduce
...
+        reshapeOutputType.getEncoding());
+    SmallVector<int32_t> reshapeShape(reshapeOutputShape.begin(),
+                                      reshapeOutputShape.end());
+    auto newReshape = rewriter.create<ReshapeOp>(
+        reshapeUser->getLoc(), newReshapeType, op.getInput(),
+        rewriter.getI32ArrayAttr(reshapeShape));
+
+    // Create reduce with mapped dims on the reshaped tensor.
+    auto newReduce = createReduceOp(op, newReshape.getResult(), newDimArgs,
+                                    reshapeOutputShape, rewriter);
```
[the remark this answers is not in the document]
mvasiljevicTT: +1

mvasiljevicTT (commented):
Thank you for adding this! 
In the benchmark tests reduce usually comes with `keep_dim=true`, but that is not necessary so you should either cover `keep_dim=false` or guard it to only true.

lib/Dialect/TTIR/Transforms/EraseInverseOps/ReduceCommutePatterns.cpp
rpavlovicTT: Done

mvasiljevicTT (approved):

sdjordjevicTT (approved):

PR #7983, opened 2026-04-15
D2M: Add TTMetal lowering for d2m.spatial

### LINKED ISSUE #7674: [d2m] Spatial op ttmetal backend support
D2M::Spatial Op supports the `ttnn` backend so far.

We also need `ttmetal` backend support for D2M::Spatial Op.

IR that includes D2M::Spatial Op should be able to be compiled without `ttnn-mode=true` option.

This would make builder tests simpler.

hkwonTT:
### Ticket
closes #7674 

### Problem description
`d2m.spatial` was only really supported on the TTNN path, so IR with spatial ops did not compile cleanly on TTMetal without ttnn-mode=true. 

### What's changed
1. convert-d2m-to-ttmetal
The pass now lowers `d2m.spatial` on the TTMetal path. It runs in two steps: first it lowers everything inside the spatial op (while the spatial op itself is still allowed), then it runs a second step that removes the spatial op and turns it into `ttmetal.enqueue_program`. When several regions are merged, kernel argument indices (for example global_semaphore) are updated so they still point at the right operands.

### Checklist

### DIFF
--- a/include/ttmlir/Conversion/D2MToTTMetal/D2MToTTMetal.h
+++ b/include/ttmlir/Conversion/D2MToTTMetal/D2MToTTMetal.h
@@ -22,7 +22,8 @@ void populateD2MToTTMetalPatterns(
     MLIRContext *ctx, RewritePatternSet &patterns, TypeConverter &typeConverter,
     ttmetal::MathFidelity mathFidelity = ttmetal::MathFidelity::HiFi4);
 
-void populateSpatialOpPatterns(MLIRContext *ctx, RewritePatternSet &patterns);
+void populateD2MToTTMetalSpatialOpPattern(MLIRContext *ctx,
+                                          RewritePatternSet &patterns);
 
 std::unique_ptr<OperationPass<ModuleOp>> createConvertD2MToTTMetalPass();
 std::unique_ptr<OperationPass<ModuleOp>>
--- a/include/ttmlir/Conversion/D2MToTTMetal/D2MToTTMetal.h
+++ b/include/ttmlir/Conversion/D2MToTTMetal/D2MToTTMetal.h
@@ -22,6 +22,8 @@ void populateD2MToTTMetalPatterns(
     MLIRContext *ctx, RewritePatternSet &patterns, TypeConverter &typeConverter,
     ttmetal::MathFidelity mathFidelity = ttmetal::MathFidelity::HiFi4);
 
+void populateSpatialOpPatterns(MLIRContext *ctx, RewritePatternSet &patterns);
+
 std::unique_ptr<OperationPass<ModuleOp>> createConvertD2MToTTMetalPass();
 std::unique_ptr<OperationPass<ModuleOp>>
 createConvertD2MToTTMetalPass(const d2m::ConvertD2MToTTMetalOptions &options);
--- a/lib/Conversion/D2MToTTMetal/D2MToTTMetal.cpp
+++ b/lib/Conversion/D2MToTTMetal/D2MToTTMetal.cpp
@@ -9,15 +9,21 @@
 #include "ttmlir/Dialect/D2M/Utils/Utils.h"
 #include "ttmlir/Dialect/TTCore/IR/TTCoreOpsTypes.h"
 #include "ttmlir/Dialect/TTCore/IR/Utils.h"
-#include "ttmlir/Dialect/TTIR/IR/TTIROps.h"
 #include "ttmlir/Dialect/TTKernel/IR/TTKernelOps.h"
 #include "ttmlir/Dialect/TTMetal/IR/TTMetalOps.h"
 
 #include "mlir/Dialect/Func/IR/FuncOps.h"
 #include "mlir/Dialect/MemRef/IR/MemRef.h"
+#include "mlir/IR/AffineMap.h"
+#include "mlir/IR/BuiltinAttributes.h"
+#include "mlir/IR/PatternMatch.h"
+
+#include "llvm/ADT/DenseMap.h"
+#include "llvm/ADT/TypeSwitch.h"
 
 #include <cstdint>
 #include <mlir-c/IR.h>
+#include <optional>
 
 namespace mlir::tt::ttmetal {
 
@@ -61,13 +67,10 @@ public:
 
   static ArrayAttr convertThreadsToKernelConfigs(
       Builder &builder, mlir::ValueRange inputOutputOperands, ArrayAttr threads,
-      ArrayRef<int64_t> physicalGridShape, const SymbolTable &symbolTable,
+      CoreRangeAttr coreRange, const SymbolTable &symbolTable,
       ttmetal::MathFidelity mathFidelity) {
     SmallVector<Attribute> kernelConfigs;
     int unassignedNocCounter = 0;
-
-    auto coreRange = ttmetal::CoreRangeAttr::getPhysicalCoreRange(
-        builder.getContext(), physicalGridShape);
 
     for (Attribute threadAttr : threads) {
       d2m::ThreadAttr thread = mlir::cast<d2m::ThreadAttr>(threadAttr);
@@ -182,10 +185,10 @@ public:
     }
 
     ArrayAttr threads = op.getThreads();
-    auto physicalGridShape = op.getPhysicalGridShape();
+    CoreRangeAttr coreRange = coreRangeAttrFromOp(rewriter, op);
     auto kernelConfigs = convertThreadsToKernelConfigs(
-        rewriter, op.getInputsAndOutputs(), threads, physicalGridShape,
-        symbolTable, mathFidelity_);
+        rewriter, op.getInputsAndOutputs(), threads, coreRange, symbolTable,
+        mathFidelity_);
     rewriter.replaceOpWithNewOp<ttmetal::EnqueueProgramOp>(
         op, args, cbs, cbPorts, kernelConfigs,
         op.getFabricConnectionConfigAttr());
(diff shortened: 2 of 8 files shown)


lib/Conversion/D2MToTTMetal/D2MToTTMetal.cpp
```diff
@@ -182,19 +185,53 @@ class D2MGenericRewriter : public OpConversionPattern<d2m::GenericOp> {
...
 };
+
+// CoreRange from d2m.generic grid: project the full virtual bbox through
+// virt_to_physical when the grid map is usable; otherwise
+// getPhysicalCoreRange(getPhysicalGridShape()).
+CoreRangeAttr D2MGenericRewriter::coreRangeAttrFromOp(Builder &builder,
+                                                      d2m::GenericOp op) {
+  MLIRContext *ctx = builder.getContext();
+  ttcore::GridAttr grid = op.getGrid();
+  AffineMap v2p = grid.getVirtToPhysicalMap();
```
vwellsTT: nit: personally would prefer more explicit variable name like virtToPhysMap or something

lib/Conversion/D2MToTTMetal/D2MToTTMetal.cpp
```diff
@@ -434,6 +471,297 @@ class D2MViewLayoutRewriter : public OpConversionPattern<d2m::ViewLayoutOp> {
...
+    return mergedFabricConfig && enqueueFabricConfig &&
+           mergedFabricConfig != enqueueFabricConfig;
+  }
+
+  static LogicalResult
+  collectRegionOps(Region &region, SmallVector<Operation *> &preEnqueueOps,
+                   SmallVector<Operation *> &postEnqueueOps,
+                   ttmetal::EnqueueProgramOp &regionEnqueueProgram) {
+    Block &block = region.front();
+    bool seenEnqueue = false;
```
vwellsTT: it seems like this is redundant w/ enqueueCount field, no? if count == 0, seenEnqueue = false, else true?

lib/Conversion/D2MToTTMetal/D2MToTTMetal.cpp
```diff
@@ -452,4 +780,8 @@ void populateD2MToTTMetalPatterns(MLIRContext *ctx, RewritePatternSet &patterns,
   patterns.add<ttmetal::D2MGenericRewriter>(ctx, mathFidelity);
 }
 
+void populateSpatialOpPatterns(MLIRContext *ctx, RewritePatternSet &patterns) {
```
vwellsTT, line 783: does this have to be converted in separate pass vs the other d2m ops?

vwellsTT (commented):
I think adding some pytests to validate spatial ops actually run would be good too

lib/Conversion/D2MToTTMetal/D2MToTTMetal.cpp
hkwonTT: Yes, it needs to be handled in a separate pass.

Inside d2m.spatial, there is a nested region containing d2m.generic. What I wanted was for d2m.generic to be lowered first, followed by d2m.spatial. However, `applyFullConversion` runs in preorder, so d2m.spatial gets processed before d2m.generic, regardless of the pattern adding order.

Because of this, I ended up using a two-phase approach: first lowering all ops except d2m.spatial, and then running a separate pass to lower only d2m.spatial.

lib/Conversion/D2MToTTMetal/D2MToTTMetal.cpp
hkwonTT: You're right. That is redundant. Thanks! I'll update the code.

include/ttmlir/Conversion/D2MToTTMetal/D2MToTTMetal.h
```diff
@@ -22,6 +22,8 @@ void populateD2MToTTMetalPatterns(
     MLIRContext *ctx, RewritePatternSet &patterns, TypeConverter &typeConverter,
     ttmetal::MathFidelity mathFidelity = ttmetal::MathFidelity::HiFi4);
 
+void populateSpatialOpPatterns(MLIRContext *ctx, RewritePatternSet &patterns);
```
anuragsingh-tt, line 25: Can we specify this is D2MToTTMetal in the name

lib/Conversion/D2MToTTMetal/D2MToTTMetal.cpp
```diff
@@ -182,19 +185,53 @@ class D2MGenericRewriter : public OpConversionPattern<d2m::GenericOp> {
...
+CoreRangeAttr D2MGenericRewriter::coreRangeAttrFromOp(Builder &builder,
+                                                      d2m::GenericOp op) {
+  MLIRContext *ctx = builder.getContext();
+  ttcore::GridAttr grid = op.getGrid();
+  AffineMap v2p = grid.getVirtToPhysicalMap();
+  ArrayRef<int64_t> vshape = grid.getShape();
+  const unsigned n = v2p.getNumResults();
+  if (v2p.isEmpty() || vshape.size() != 2 || v2p.getNumDims() != 2 ||
+      (n != 2 && n != 3)) {
+    return CoreRangeAttr::getPhysicalCoreRange(ctx, op.getPhysicalGridShape());
```
anuragsingh-tt, line 217: Is this a precondition or is this fallback legal? If the former, can we assert instead?

lib/Conversion/D2MToTTMetal/D2MToTTMetal.cpp
```diff
@@ -434,6 +473,295 @@ class D2MViewLayoutRewriter : public OpConversionPattern<d2m::ViewLayoutOp> {
...
+/// GlobalSemaphore, and CBPort. Runtime resolves CBPort by indexing the merged
+/// `cbs` list, then reads that entry's hardware port.
+class SpatialOpRewriter : public OpConversionPattern<d2m::SpatialOp> {
+public:
+  using OpConversionPattern<d2m::SpatialOp>::OpConversionPattern;
+
+  LogicalResult
+  matchAndRewrite(d2m::SpatialOp op, d2m::SpatialOp::Adaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const final {
+    (void)adaptor;
```
anuragsingh-tt, line 491: NIT: `maybe_unused` instead?

lib/Conversion/D2MToTTMetal/D2MToTTMetal.cpp
hkwonTT: Good catch, thanks. I also prefer to use more explicit variable name. The name changed during an AI-assisted cleanup, and I did not catch that it became unclear. I will update it to something more explicit.

lib/Conversion/D2MToTTMetal/D2MToTTMetal.cpp
```diff
@@ -434,6 +473,295 @@ class D2MViewLayoutRewriter : public OpConversionPattern<d2m::ViewLayoutOp> {
...
+    SpatialRemapTable remapTable;
+    SmallVector<ttmetal::EnqueueProgramOp> regionEnqueues;
+    SmallVector<Value> mergedCbs;
+    SmallVector<int64_t> mergedCbPorts;
+    SmallVector<Attribute> mergedKernelConfigs;
+    ttcore::FabricConnectionConfigAttr mergedFabricConfig = nullptr;
+    SmallVector<Operation *> preEnqueueOps;
+    SmallVector<Operation *> postEnqueueOps;
+
+    for (auto [regionIndex, region] : llvm::enumerate(op.getRegions())) {
```
anuragsingh-tt, line 508: Is region used?

lib/Conversion/D2MToTTMetal/D2MToTTMetal.cpp
```diff
@@ -434,6 +473,295 @@ class D2MViewLayoutRewriter : public OpConversionPattern<d2m::ViewLayoutOp> {
...
+      auto spatialCoreRange =
+          mlir::cast<ttcore::CoreRangeAttr>(gridRanges.getValue()[regionIndex]);
+      CoreRangeAttr spatialMetalRange =
+          ttCoreSpatialRangeToTtmetalCoreRange(rewriter, spatialCoreRange);
+      ttmetal::EnqueueProgramOp enqueueProgram = nullptr;
```
anuragsingh-tt, line 513: Why not FailureOr instead?

lib/Conversion/D2MToTTMetal/D2MToTTMetal.cpp
```diff
@@ -434,6 +473,295 @@ class D2MViewLayoutRewriter : public OpConversionPattern<d2m::ViewLayoutOp> {
...
+    rewriter.create<ttmetal::EnqueueProgramOp>(
+        op.getLoc(), remapTable.getUnifiedArgs(), mergedCbs, mergedCbPorts,
+        rewriter.getArrayAttr(mergedKernelConfigs), mergedFabricConfig);
+
+    for (Operation *operation : postEnqueueOps) {
+      rewriter.moveOpBefore(operation, op);
+    }
+
+    for (ttmetal::EnqueueProgramOp enqueueProgram : regionEnqueues) {
+      rewriter.eraseOp(enqueueProgram);
```
anuragsingh-tt, line 561: If we erase the op below won't it also erase these?

include/ttmlir/Conversion/D2MToTTMetal/D2MToTTMetal.h
hkwonTT: Thanks. Suggestion accepted.

lib/Conversion/D2MToTTMetal/D2MToTTMetal.cpp
hkwonTT: Great catch. Precondition and fallback were mixed. I separated them and added asserts for the preconditions.

lib/Conversion/D2MToTTMetal/D2MToTTMetal.cpp
hkwonTT: Thanks. Updated to `maybe_unused`.

lib/Conversion/D2MToTTMetal/D2MToTTMetal.cpp
hkwonTT: Yes, it is used as an argument to collectRegionOps, which collects EnqueueProgramOp and pre/post ops.

lib/Conversion/D2MToTTMetal/D2MToTTMetal.cpp
hkwonTT: Thanks. Updated to use FailureOr.

lib/Conversion/D2MToTTMetal/D2MToTTMetal.cpp
hkwonTT: Good catch. Removed this line.

anuragsingh-tt (approved):

test/python/golden/test_spatial.py
```diff
@@ -0,0 +1,572 @@
+# SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
```
hkwonTT:
```suggestion
# SPDX-FileCopyrightText: (c) 2026 Tenstorrent AI ULC
```

PR #7986, opened 2026-04-16
Fix conv3d opModel

djstefanovicTT:
### Problem description
Conv3d config became optional in tt-metal. If no config is passed, no default value is needed. Op model was still creating empty config for conv3d that caused mismatch between actual execution and op model.

### What's changed
Made conv3d config optional in op model.


sdjordjevicTT (approved):
Nice catch!

azecevicTT (approved):

PR #7987, opened 2026-04-16
Reshard improvements, per operand filtering, reshape view optimisation

rpavlovicTT:
## Summary
- **Stratified reshard candidate selection**: take top-N per sharding type (block, height, width) instead of global top-N by grid volume, preventing block_sharded from crowding out height_sharded candidates.
- **Per-operand input layout filter**: `getInputLayoutFilter()` now takes an `operandIdx` parameter. For matmul/linear, LHS (operand 0) allows all sharding types (width/height/block all supported by tt-metal), while RHS (operand 1) rejects all sharded layouts. DRAM width-sharded RHS is a tt-metal special case not yet supported by our program config generation.
- **Filter-aware reshard exploration**: pre-filter reshard targets before beam fan-out to prevent doomed layouts from consuming budget slots, and check filter when deciding whether interleaved-to-sharded exploration is needed.
- **Reshape view optimization**: view-eligible reshapes now use NULL output hints so they inherit the input's memory config, preserving the zero-cost view path in tt-metal. Previously, the greedy optimizer's L1-first scoring would upgrade DRAM tensors to L1 before a reshape, breaking the view optimization and turning a free metadata-only op into a real data move. On Llama 70B 2-layer prefill, this reduces device time by 4.2% and eliminates 31 device ops.
- **Universal tiebreaker**: prefer candidates with more sharded inputs when scores are equal.

## Test plan
- [x] tt-xla perf benchmarks passed on n150 — almost all improvements, few CNN regressions (known, will be addressed separately) https://github.com/tenstorrent/tt-xla/actions/runs/24460504416
- [x] LLMBOX perf tests have perf improved (tested locally)


lib/Dialect/TTNN/Analysis/OpRules/OpRuleBook.cpp
```diff
@@ -75,6 +91,9 @@ const OpRuleBook &getRuleBook(Operation *op) {
     reg(SliceStaticOp::getOperationName(), &slice);
     reg(SliceDynamicOp::getOperationName(), &slice);
     reg(ReshapeOp::getOperationName(), &reshape);
+
+    // TODO: split permute's from reshape's rule book
```
vmilosevic: :warning: **google\-readability\-todo** :warning:
missing username/bug in TODO
```suggestion
    // TODO(nsmith): split permute's from reshape's rule book
```

vmilosevic (commented):

mtopalovicTT (approved):

include/ttmlir/Dialect/TTNN/Analysis/OpRules/MatmulRules.h
```diff
@@ -29,6 +29,9 @@ struct MatmulRuleBook : OpRuleBook {
   getOutputHints(Operation *op,
                  const std::vector<OpConfig> &legalConfigs) const override;
 
+  /// Reject all sharded RHS inputs (operand 1). LHS and bias are unrestricted.
```
bmalesevicTT, line 32: Is it confirmed matmul can't support L1 sharded weights in our current compiler path? In tt-metal very specific cases work depending on matmul config, but it seems we can reach some, not always doomed as comments suggest.
The 1D multicast + L1-sharded-weights variants are hard to drive from tt-mlir because runtime recomputes the program config, **but** for 2D multicast with L1 BLOCK_SHARDED output and L1 WIDTH_SHARDED weights (in0 is interleaved or L1 HEIGHT_SHARDED) we are already in a supported, compiler-reachable combination.
Until DRAM sharding is enabled, I would guess they could be useful for perf and not fully ignored?

rpavlovicTT: We don't support weight sharding yet through compiler, since weights are parameters and usually big tensors which don't make much sense to move to L1. I am rejecting RHS sharding since I encountered bad PCC with it. Although I couldn't reproduce in isolation (single op test), I observed bad values through tensor dumping when running whole model e2e. For now this is the safest bet.

bmalesevicTT (approved):

odjuricicTT (approved):
Awesome!

PR #7992, opened 2026-04-16
Refactor Mesh 2D Normalization

### LINKED ISSUE #7507: Refactor mesh normalization in shardy passes
There are a couple of places where device mesh is normalized - this should be refactored so normalization is performed once on a common code path.

sdjukicTT:
### Ticket
https://github.com/tenstorrent/tt-mlir/issues/7507

### Problem description
Mesh normalization (expanding 0D/1D meshes to 2D) was duplicated in three places: inline in `AnalyzeMeshPass`, in `normalize1DMeshTo2D` called from` ConvertXlaSdyToSdyPass`, and in `parseMeshFromFrontendAttributes` in `GSPMDUtils.cpp.` Each used slightly different logic and inconsistent axis naming (`_updated` vs `_aux`).

### What's changed
- Mesh normalization now happens in exactly one place: `normalizeMeshTo2D` at the top of `AnalyzeMeshPass`. It handles 0D, 1D, and 2D meshes uniformly.                            
- Removed normalization and custom-call-to-sharding-constraint conversion from `ConvertXlaSdyToSdyPass`.                                                                         
- Removed inline normalization logic from `AnalyzeMeshPass`'s SDY branch.                               
- `parseMeshFromFrontendAttributes` now creates the mesh as-is (1D stays 1D) instead of expanding to 2D itself.                                                                  
- Added `addMeshToModule(module, name, MeshMap)` overload to support creating meshes with arbitrary axis count.                                                                  
- Standardized auxiliary axis naming to `_aux` everywhere.                                                                                                                       
- Updated all affected test expectations.    

### Checklist


lib/Dialect/StableHLO/Transforms/ConvertXlaSdyToSdy.cpp
```diff
@@ -43,20 +42,6 @@ class ConvertXlaSdyToSdyPass
         return;
       }
     }
-
```
acicovicTT, line 46: Is it okay to just remove this since ConvertXlaSdyToSdy happens before AnalyzeMeshPass?

acicovicTT: I guess the main idea with refactoring this file was move of `convertCustomCallToShardingConstraint` to AnalyzeMeshPass, so pls check whether `convertCustomCallToShardingConstraint` needs to happen for some of the passes in between

sdjukicTT: I checked by reviewing code and by running tests.

acicovicTT (approved):

PR #8002, opened 2026-04-16
[d2m] tag intermediate scratch allocs with d2m.blocking_maps

ckaravasilisTT:
Tag intermediate `memref.allocs` (not backed by `d2m.generic` operands) within fused `d2m.generics` with an appropriate affine map to assist the Allocator when reblocking.


anuragsingh-tt:
Can we add a quick lit test to see that it survives through GenerateOuterLoops down to allocator?

anuragsingh-tt (approved):

PR #8010, opened 2026-04-17
[Optimizer] Fix conv2d shard layout and concat mixed-input handling

rpavlovicTT:
## Summary

Two related optimizer fixes that address CNN model perf regressions and segformer compilation failure:

**1. Conv2d shard layout propagation to tt-metal runtime**

When the greedy optimizer chooses a sharded output layout for conv2d, embed the shard layout into the \`Conv2dConfig\`'s \`shard_layout\` field in \`getOutputHints\`. Without this, the tt-metal runtime's auto-shard heuristic picks a different (often slower) shard type when the input is later spilled to DRAM. Fixes a 4x regression on 128x128 spatial conv2d ops (876μs → 224μs, matching legacy).

Additional fixes:
- \`withBufferType(DRAM)\` memref dimensions: append \`.withTensorShape()\` at all call sites converting sharded L1 layouts to DRAM interleaved (6 sites across \`MemoryLayoutPropagation\`, \`L1SpillManagement\`, \`OperationValidationAndFallback\`). Previously the per-core shard dimensions leaked into the DRAM memref (e.g., \`memref<32x4>\` instead of \`memref<2048x4>\`).
- Add same-shard-type constraint for conv2d via \`Conv2dRuleBook::isValidOutputHintForInputs\` to prevent costly HS→BS reshards.

**2. Sibling-operand spill for backend constraint failures**

When L1 spill management evicts one concat operand to DRAM under CB-fragmentation pressure, the concat fails validation with a structural constraint error (\"All tensors must be sharded or all must be interleaved\") because the remaining operands are still L1 sharded.

In \`handleFragmentation\`, after partial eviction, distinguish OOM errors from backend constraint errors via \`ValidationResult\`'s status enum. On non-OOM failure, spill all remaining L1 operands of the op to DRAM, then fall through to \`demoteToDram\` so the op runs all-DRAM. Intentionally do not keep a sharded output after sibling spill to avoid a known runtime bug (https://github.com/tenstorrent/tt-metal/issues/41469).

Refactor: extract \`evictValue\` helper from \`evictUntil\` so both the farthest-use eviction loop and the sibling-operand path share the same tracker-rebuild + consumer-reshard bookkeeping.

## Test plan

- [x] UNet compiles and runs successfully on n150 silicon
- [x] Segformer compiles and runs successfully on n150 silicon (previously aborted with compiler error)
- [x] CNN perf benchmark on n150 non-shared runner vs April 15 nightly baseline:
  - resnet: **+26%**
  - unet: **+18%**
  - vovnet: **+17%**
  - efficientnet: **+7%**
  - segformer: **+26%**
  - mobilenetv2: neutral
- [x] Full CI pipeline on PR

### DIFF
--- a/lib/Dialect/TTNN/Analysis/L1SpillManagement.cpp
+++ b/lib/Dialect/TTNN/Analysis/L1SpillManagement.cpp
@@ -652,6 +652,53 @@ void L1SpillManagement<MemoryTracker>::markEvictedAndRebuild(Value victim) {
 //===----------------------------------------------------------------------===//
 
 template <typename MemoryTracker>
+void L1SpillManagement<MemoryTracker>::evictValue(Value victim, int64_t pos,
+                                                  const ScheduleData &data) {
+  liveValues.erase(victim);
+  Operation *victimOp = victim.getDefiningOp();
+  uint64_t freedBytes = memoryTracker.getTensorSize(victim);
+
+  // Save original L1 layout before spilling.
+  auto tensorType = mlir::cast<RankedTensorType>(victim.getType());
+  TTNNLayoutAttr originalL1Layout =
+      mlir::cast<TTNNLayoutAttr>(tensorType.getEncoding());
+
+  TTMLIR_DEBUG(ttmlir::LogComponent::GreedyOptimizer,
+               "    EVICT: {0} (L1: {1} bytes)", ttmlir::opToString(victimOp),
+               freedBytes);
+  observer_->onEviction(victimOp, pos, freedBytes);
+
+  spillToDram(victim);
+  memoryTracker.removeTensorFromSizes(victim);
+  markEvictedAndRebuild(victim);
+
+  // Insert reshards for already-processed consumers instead of cascade
+  // revalidation. This preserves downstream ops' sharded layouts.
+  if (originalL1Layout.hasL1BufferType()) {
+    for (Operation *consumer : collectDownstreamConsumers(victimOp)) {
+      auto posIt = data.positionMap.find(consumer);
+      if (posIt == data.positionMap.end() || posIt->second >= pos) {
+        continue;
+      }
+      if (!mlir::dyn_cast<OpModel>(consumer)) {
+        continue;
+      }
+      if (isa<ToLayoutOp>(consumer)) {
+        continue;
+      }
+      for (unsigned i = 0; i < consumer->getNumOperands(); ++i) {
+        Value operand = consumer->getOperand(i);
+        Operation *defOp = operand.getDefiningOp();
+        if (isa_and_nonnull<ToMemoryConfigOp>(defOp) &&
+            defOp->getOperand(0) == victim) {
+          insertReshardForConsumer(consumer, i, originalL1Layout);
+        }
+      }
+    }
+  }
+}
+
+template <typename MemoryTracker>
 bool L1SpillManagement<MemoryTracker>::evictUntil(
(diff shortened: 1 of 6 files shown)


lib/Dialect/TTNN/Analysis/L1SpillManagement.cpp
```diff
@@ -1174,6 +1175,17 @@ void L1SpillManagement<MemoryTracker>::evictForDramCBGrowth(
       op_constraint_validation::validateOperation(op, inputLayouts, config,
                                                   /*additionalL1Usage=*/0);
   if (!result.isSuccess()) {
+    // Concat requires all inputs to be homogeneous (all sharded or all
+    // interleaved). When we spill one of its operands to DRAM, the concat
+    // gets mixed inputs and fails validation here. The fallback pass runs
+    // later and can resolve this by spilling the remaining sharded operands
+    // to DRAM. Downgrade to a warning to let compilation proceed.
+    if (isa<ConcatOp>(op)) {
```
odjuricicTT: Hmm, this seems pretty hacky. Maybe there are other way to try to handle this case more generally:
1. If op is invalid after operand spill proceed to spill other operands and check again
2. Have per op spill rule and handle all concat input then here (might be overkill)
3. Call op validation right here and then deal with the result

What do you think?

odjuricicTT (commented):
Nice perf gain :)

lib/Dialect/TTNN/Analysis/L1SpillManagement.cpp
rpavlovicTT: Thanks for the ideas. I decided to spill all operands in case op failed validation (with non-OOM error) after some of its operands have been evicted. After this we demote op's output to DRAM as well, and revalidate. If final validation fails, we will emit an error.

rpavlovicTT:
https://github.com/tenstorrent/tt-xla/actions/runs/24587634576  vs Apr 15th nightly

```
  ┌─────────────────────────────────┬─────────────┬──────────┬──────────┐                                      
  │              Model              │ Samples/s Δ │ Change % │  Result  │                                      
  ├─────────────────────────────────┼─────────────┼──────────┼──────────┤                                      
  │ resnet                          │ +246.83     │ +27%     │ IMPROVED │                                      
  ├─────────────────────────────────┼─────────────┼──────────┼──────────┤                                      
  │ unet_for_conditional_generation │ +0.98       │ +38%     │ IMPROVED │                                      
  ├─────────────────────────────────┼─────────────┼──────────┼──────────┤                                      
  │ resnet_jax                      │ +16.03      │ +29%     │ IMPROVED │                                      
  ├─────────────────────────────────┼─────────────┼──────────┼──────────┤                                      
  │ segformer                       │ +8.00       │ +27%     │ IMPROVED │
  ├─────────────────────────────────┼─────────────┼──────────┼──────────┤                                      
  │ unet                            │ +19.40      │ +15%     │ IMPROVED │
  ├─────────────────────────────────┼─────────────┼──────────┼──────────┤                                      
  │ vovnet                          │ +63.38      │ +11%     │ IMPROVED │
  ├─────────────────────────────────┼─────────────┼──────────┼──────────┤                                      
  │ efficientnet                    │ +13.69      │ +4%      │ NEUTRAL  │
  ├─────────────────────────────────┼─────────────┼──────────┼──────────┤                                      
  │ mobilenetv2                     │ +29.71      │ +3%      │ NEUTRAL  │
  └─────────────────────────────────┴─────────────┴──────────┴──────────┘   

```

odjuricicTT (approved):

PR #8016, opened 2026-04-17
Add ttnn.experimental.moe_gpt op to TTIR/TTNN dialects

jserbedzijaTT:
## Summary

- Adds `ttnn.experimental.moe_gpt` op end-to-end (TTIR/TTNN dialects, conversion, flatbuffer, runtime, EmitC/EmitPy, op model stub, builder API, golden).
- Moves bf16 DRAM-INTERLEAVED → bfp4_b TILE DRAM HEIGHT_SHARDED weight prep and the WH ring-bank placement into a compiler workaround pattern (`MoeGptLayoutRewritePattern`), replacing the runtime-side prep.
- Fixes `toTTNNCoreRangeSet` in the runtime: was `std::set<CoreRange>` which sorts by `start_coord` during flatbuffer deserialization and silently destroyed the moe_gpt ring-bank permutation `[5,0,7,8,11,3,10,2,9,6,4,1]`. Switched to `std::vector` to preserve emitted order.
- Adds a builder test (`test_moe_dispatch_metadata_moe_gpt[fabric_1d_ring-4x8-ttnn]`) that exercises `all_to_all_dispatch_metadata → moe_gpt` on GPT-OSS 120B shapes. Verifies routing-metadata outputs (token_counts, activation_records, token_indices) with PCC=1.0 on all 32 devices. `tilize_out` / `tilize_out_rm` are not directly verified — they are internal staging buffers for `selective_reduce_combine` whose per-expert byte-offset layout (`shard_offset_per_expert_bytes * expert_id`) does not match a simple `local_e_idx * TILE + srow` walk for `E_per_device > 2`; tt-metal's own `test_moe_gpt_e2e.py` skips direct verification for the same reason.

## Draft, not ready to land

- `MoeGptLayoutRewritePattern` hardcodes the WH N300 bank ring order `kWormholeMoeGptBankRingOrder`; should be derived from the system descriptor at compile time.
- `ToDeviceOp` result encoding uses a grid-affine-derived CoreRangeSet that cannot mirror the explicit permuted set in `memory_config`; in debug builds `insertTTNNTensorAndValidate` would fail on that op. Proper fix is extending `TTNNLayoutAttr` to carry an explicit `CoreRangeSet`.
- Fused-decode end-to-end correctness for outputs 3/4 is not covered; follow-up PR should chain `selective_reduce_combine` and verify the `K × B × H` combine output as tt-metal does.
- Only tested on `fabric_1d_ring` at 4x8 mesh.

## Test plan

- [x] `check-ttmlir` compiler tests pass locally
- [x] `test_moe_dispatch_metadata_moe_gpt[fabric_1d_ring-4x8-ttnn]` passes on a 4x8 galaxy
- [x] `test_moe_dispatch_metadata[fabric_1d_ring-4x8-ttnn]` still passes
- [ ] emitpy target for the same test
- [ ] e2e verification via `selective_reduce_combine` chain


sdjordjevicTT (approved):

tools/golden/mapping.py
```diff
@@ -1568,6 +1568,274 @@ def ttir_all_to_all_dispatch_metadata_golden(
...
 
 
+def _swiglu_reference(gate, up, alpha=1.702, clamp_limit=7.0):
+    """SwiGLU activation matching tt-metal's SFPU implementation."""
+    gate_c = torch.clamp(gate, max=clamp_limit)
+    up_c = torch.clamp(up, min=-clamp_limit, max=clamp_limit)
+    return (up_c + 1.0) * gate_c * torch.sigmoid(alpha * gate_c)
+
+
+def moe_gpt_golden(
```
mmilosevicTT, line 1578: This should be registered with builder and also should add chisel wrapper.

tools/golden/mapping.py
```diff
@@ -1568,6 +1568,274 @@ def ttir_all_to_all_dispatch_metadata_golden(
...
+    raw_w1: torch.Tensor,
+    raw_w2: torch.Tensor,
+    hidden_size: int,
+    cluster_axis: int = 0,
+    num_worker_cores: int = 72,
+    token_counts_shape: Tuple = None,
+    activation_records_shape: Tuple = None,
+    token_indices_shape: Tuple = None,
+    tilize_out_shape: Tuple = None,
+    tilize_out_rm_shape: Tuple = None,
```
jgrimTT: All golden args should be mlir types and then unpacked with `unpack_mlir_attr`. And should also use `output_type_mlir`

tools/builder/base/builder_runtime.py
```diff
@@ -769,6 +769,17 @@ def execute_fb(
...
+                    save_torch_tensor(
+                        output_shard_torch,
+                        program_artifact_dir,
+                        f"device_output_{i}_device_{device_id}.pt",
+                    )
+                    save_torch_tensor(
+                        golden_shard_torch,
+                        program_artifact_dir,
+                        f"golden_output_{i}_device_{device_id}.pt",
+                    )
```
jgrimTT, line 782: That's helpful, thx :)

tools/builder/ttir/ttir_builder.py
```diff
@@ -181,6 +181,209 @@ def _to_dispatch_layout_for_golden(
...
+                                src_s = min(s, int(src_input.shape[2]) - 1)
+                                dest_output[k, b_local, s, :] = src_input[
+                                    local_idx, src_b, src_s, :
+                                ]
+
+            if valid_mapping:
+                golden = GoldenMapTensor(output_shards, input_tensor.mesh_shape)
+
+        return golden
+
```
jgrimTT: There shouldn't be any extra golden handling in ttir_builder.py, see how other ops use `get_golden_function`

tools/builder/ttir/ttir_builder.py
```diff
@@ -16887,6 +17090,283 @@ def decorated_func(*inputs):
...
+        token_counts_type: Optional[torch.dtype] = None,
+        activation_records_shape: Optional[Shape] = None,
+        activation_records_type: Optional[torch.dtype] = None,
+        token_indices_shape: Optional[Shape] = None,
+        token_indices_type: Optional[torch.dtype] = None,
+        tilize_out_shape: Optional[Shape] = None,
+        tilize_out_type: Optional[torch.dtype] = None,
+        tilize_out_rm_shape: Optional[Shape] = None,
+        tilize_out_rm_type: Optional[torch.dtype] = None,
+        unit_attrs: Optional[List[str]] = None,
```
jgrimTT, line 16903: Add `loc` as optional arg

tools/builder/ttir/ttir_builder.py
```diff
@@ -16887,6 +17090,283 @@ def decorated_func(*inputs):
...
+        for name in result_names:
+            old_result = getattr(old_op, name)
+            new_result = getattr(new_op, name)
+            result_type = old_result.type
+            shape = tuple(int(dim) for dim in result_type.shape)
+            dtype = mlir_type_to_torch_dtype(result_type.element_type)
+            golden = GoldenMapTensor(
+                {0: torch.zeros(shape, dtype=dtype)},
+                mesh_shape=self._mesh_shape,
+            )
```
jgrimTT, line 17082: Use `{i: torch.zeros().clone() for i in range(mesh_shape[0] * mesh_shape[1])}` to preserve sharding config

jgrimTT (changes requested):
Added some comments, feel free to reach out with questions about the golden mappings :)

tools/golden/mapping.py
mtopalovicTT: I've looked at adding chisel wrapper for this op, but it gets complicated a lot, requires a lot of refactoring for preparing the weights for golden. I would add this as is without chisel as this op will get replaced with gpt_compute op soon which will have weight preparation as separate op and this will become non issue.

mtopalovicTT:
@jgrimTT I've went through your comments. Please check this today since we need this op asap. :)

tools/builder/ttir/ttir_builder.py
jgrimTT: Remove unused functions

jgrimTT (approved):

tools/builder/ttir/ttir_builder.py
mtopalovicTT: Done

pilkicTT (approved):

tools/golden/mapping.py
mmilosevicTT: Fine by me. If we do not add ttnn golden, chisel can stay without it for now.

PR #8017, opened 2026-04-17
Multihost API - Get Worker Debug Stats

jameszianxuTT:
### Ticket
None

### Problem description
We currently have no way to introspect on the worker internals. As multihost has some worker behaviour diverging from the controller, this information gap becomes a hindrance as we cannot easily test for worker-internal mechanics. 

Current debug stats API is not suitable as debug stats is a process-local singleton, so we can query debug stats on the controller, but since the workers live on separate processes, their debug stats are inaccessible. 

### What's changed
- Introduce new multihost API getWorkerDebugStats, to mirror existing debug APIs to get debug stats. This will return a vector of WorkerDebugStatsEntry, which are hostname, DebugStatsMap (string -> i64 dictionary)

### Checklist
- [ ] Added multihost test

### DIFF
--- a/runtime/python/runtime/runtime.cpp
+++ b/runtime/python/runtime/runtime.cpp
@@ -2,7 +2,16 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 
+#include <algorithm>
+#include <cstddef>
+#include <cstdint>
+#include <iterator>
+#include <map>
+#include <optional>
 #include <sstream>
+#include <string>
+#include <unordered_map>
+#include <vector>
 
 #include "tt/runtime/debug.h"
 #include "tt/runtime/perf.h"
--- a/runtime/python/runtime/runtime.cpp
+++ b/runtime/python/runtime/runtime.cpp
@@ -362,8 +362,11 @@ void registerRuntimeBindings(nb::module_ &m) {
   m.def("shutdown_distributed_runtime",
         &tt::runtime::shutdownDistributedRuntime,
         "Shutdown the distributed runtime");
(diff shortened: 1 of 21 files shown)


jameszianxuTT:
Main use case for now -> we want to introspect on worker-side tensor caches as that feature is brought up.

jameszianxuTT:
To introspect on worker-specific DRAM leaks suspect -> https://github.com/tenstorrent/tt-xla/issues/3791

runtime/python/runtime/runtime.cpp
```diff
@@ -362,6 +365,27 @@ void registerRuntimeBindings(nb::module_ &m) {
...
+      .def("__repr__", [](const tt::runtime::WorkerDebugStatsEntry &entry) {
+        std::ostringstream oss;
+        oss << "Worker: " << entry.hostname << "\n";
+        // Sort keys for consistent output
+        std::vector<std::string> keys;
+        keys.reserve(entry.stats.size());
+        for (const auto &[key, value] : entry.stats) {
+          keys.push_back(key);
+        }
+        std::sort(keys.begin(), keys.end());
```
pilkicTT, line 389: please include, vector, algorithm, string, etc.

pilkicTT (approved):

jgrimTT (approved):
TTRT change approved

jameszianxuTT:
i will also add a json repr for the pytorch binding if I can do that with minimal changes (maybe separate PR), the caller can use these stats more effectively if they're in a structured format

PR #8022, opened 2026-04-20
Support added to fix Convolution Feature Group Count

sonalibaskaran2499:
### Ticket
https://github.com/tenstorrent/tt-metal/issues/45396

### Problem description
When Shardy applies tensor-parallel sharding to a grouped or depthwise `stablehlo.convolution`:

- `lhs_input_feature_dim` becomes local (global / shard factor)
- `feature_group_count` is **not** updated and stays at the global value

StableHLO verification then fails because it requires
`lhs_input_feature_dim % feature_group_count == 0` and
`in_channels_per_group = lhs_input_feature_dim / feature_group_count`.

This blocks compilation of any model that combines channel-dim TP sharding with a grouped/depthwise conv — notably the depthwise `Conv1d` inside the `GatedDeltaNet` linear-attention layers in **Qwen 3.5**.

### What's changed
After `updateShapes` runs in `UpdateGlobalToLocalShapesPass`, walk all `stablehlo.convolution` ops with `feature_group_count > 1` and recompute the attribute from the now-local operand types:

```
in_channels_per_group   = rhs_kernel_input_feature_dim         // invariant under sharding
new_feature_group_count = local_lhs_input_feature_dim / in_channels_per_group
```

For depthwise conv (`in_channels_per_group == 1`) this simplifies to
`new_feature_group_count = local_lhs_input_feature_dim`.

The pass is a no-op on:
- Non-grouped convs (`feature_group_count <= 1`)
- Convs where the values were already consistent
- Cases where `local_lhs_input_feature_dim` is not divisible by `in_channels_per_group` (guard against bad rewrites)

### Checklist

### DIFF
--- a/lib/Dialect/StableHLO/Transforms/UpdateGlobalToLocalShapes.cpp
+++ b/lib/Dialect/StableHLO/Transforms/UpdateGlobalToLocalShapes.cpp
@@ -402,6 +402,65 @@ static mlir::LogicalResult updateShapes(MLIRContext *context,
   return mlir::success();
 }
 
+// Shardy shards the convolution's feature dim across operands but leaves
+// feature_group_count at its global value, so the local (already-reduced) input
+// feature dim is no longer divisible by it and StableHLO verification rejects
+// the op. The fix is to rescale the count from the localized types.
+// Issue: https://github.com/tenstorrent/tt-metal/issues/45396
+//
+// in_channels_per_group is invariant under sharding, so the new count follows
+// from the local LHS feature dim:
+//   in_channels_per_group = global_in_feature_dim / feature_group_count
+//   new_feature_group_count = local_in_feature_dim / in_channels_per_group
+// (depthwise conv has in_channels_per_group == 1, so the count equals the local
+// feature dim.)
+static mlir::LogicalResult
+fixConvolutionFeatureGroupCount(MLIRContext *context,
+                                mlir::ModuleOp &rootModule) {
+  mlir::WalkResult result =
+      rootModule.walk([&](mlir::stablehlo::ConvolutionOp convOp) {
+        uint64_t currentFGC = convOp.getFeatureGroupCount();
+        // A standard convolution (feature_group_count == 1) keeps a single
+        // group no matter how the feature dim is sharded, so its count never
+        // goes stale and there is nothing to recompute.
+        if (currentFGC <= 1) {
+          return WalkResult::advance();
+        }
+
+        auto lhsType =
+            mlir::cast<mlir::RankedTensorType>(convOp.getLhs().getType());
+        auto rhsType =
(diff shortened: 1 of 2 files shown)


lib/Dialect/StableHLO/Transforms/UpdateGlobalToLocalShapes.cpp
```diff
@@ -402,6 +402,54 @@ static mlir::LogicalResult updateShapes(MLIRContext *context,
...
+        convOp.getDimensionNumbers().getInputFeatureDimension();
+    int64_t kernelInputFeatureDim =
+        convOp.getDimensionNumbers().getKernelInputFeatureDimension();
+
+    int64_t localInputChannels = lhsType.getDimSize(inputFeatureDim);
+    int64_t inChannelsPerGroup = rhsType.getDimSize(kernelInputFeatureDim);
+
+    if (inChannelsPerGroup <= 0 ||
+        localInputChannels % inChannelsPerGroup != 0) {
+      return;
```
sshonTT: Nit : Could you add `emitError` on here? So that we could understand what's the issue

lib/Dialect/StableHLO/Transforms/UpdateGlobalToLocalShapes.cpp
```diff
@@ -402,6 +402,54 @@ static mlir::LogicalResult updateShapes(MLIRContext *context,
...
+// the op. The fix is to rescale the count from the localized types.
+// Issue: https://github.com/tenstorrent/tt-metal/issues/45396
+//
+// in_channels_per_group is invariant under sharding, so the new count follows
+// from the local LHS feature dim:
+//   in_channels_per_group = global_in_feature_dim / feature_group_count
+//   new_feature_group_count = local_in_feature_dim / in_channels_per_group
+// (depthwise conv has in_channels_per_group == 1, so the count equals the local
+// feature dim.)
+static void fixConvolutionFeatureGroupCount(MLIRContext *context,
```
sshonTT: Could you add a test to make sure conversion is well done? You can add tests under `test/ttmlir/Dialect/StableHLO/global_to_local_shape`

lib/Dialect/StableHLO/Transforms/UpdateGlobalToLocalShapes.cpp
sonalibaskaran2499: @sshonTT I've added the emitError

lib/Dialect/StableHLO/Transforms/UpdateGlobalToLocalShapes.cpp
sonalibaskaran2499: @sshonTT I've added the mlir test in test/ttmlir/Dialect/StableHLO/global_to_local_shape

sshonTT (approved):

PR #8028, opened 2026-04-20
[TTNNFusing] Fix attention mask broadcast for SDPA

mvasiljevicTT:
### Ticket
[xla pcc issue](https://github.com/tenstorrent/tt-xla/issues/4206)
[known metal sdpa issue](https://github.com/tenstorrent/tt-metal/issues/39910)

### Problem description
Even though pcc on prefill is good, outputs for multichip LLMs are mostly bad, like in [this nightly run](https://github.com/tenstorrent/tt-xla/actions/runs/24642681050/job/72084379748). This is especially present in galaxy llama and gpt oss.

### What's changed
The root cause is attention_mask broadcasting in the decode sdpa op. 

Without checking and asserting shape, device ops assumes that it has same batch and num_heads as Q. When provided with tensor of batch_size 1 instead, it still reads rest of the data in expected size. When these are all 0, and KV cache has 0 on future places,  it might not look that bad with outputs repeating sensible frequent words, but if something else is already written there output can be completely random. 

Output for the first user was always ok as it was the only one with correct attention_mask. The fix was needed only for decode, so pcc that we calculate on prefill will not be affected.

Some perf drop is expected, but it will be back once we support `is_causal` and `sliding_window` built-in masks.

It should not change pcc of any model as we measure it today because we check only first user prefill, and this is fixing decode for non-first tokens. 

### Checklist
- [x] [perf benchmark on llmbox](https://github.com/tenstorrent/tt-xla/actions/runs/24664607317/job/72120802836) - mlir change only (gpt oss still expected to be weird)
- [x] [perf benchmark on galaxy](https://github.com/tenstorrent/tt-xla/actions/runs/24666905318/job/72129144995) - with [xla fix for gpt oss](https://github.com/tenstorrent/tt-xla/pull/4316)
- [x] [perf benchmark n150](https://github.com/tenstorrent/tt-xla/actions/runs/24710067795)

### DIFF
--- a/lib/Dialect/TTIR/IR/TTIROps.cpp
+++ b/lib/Dialect/TTIR/IR/TTIROps.cpp
@@ -6599,13 +6599,13 @@ mlir::tt::ttir::ScaledDotProductAttentionDecodeOp::verify() {
     if (attentionMaskType.getShape().size() != 4) {
       return emitOpError("Attention mask must be a 4D tensor");
     }
-    if (attentionMaskType.getShape()[0] != 1 &&
-        attentionMaskType.getShape()[0] != batchSize) {
+    if (attentionMaskType.getShape()[0] != 1) {
+      return emitOpError("Attention mask dim 0 must be 1");
+    }
+    if (attentionMaskType.getShape()[1] != 1 &&
+        attentionMaskType.getShape()[1] != batchSize) {
       return emitOpError("Attention mask batch size must be 1 (broadcast) or "
                          "match query batch size");
-    }
-    if (attentionMaskType.getShape()[1] != 1) {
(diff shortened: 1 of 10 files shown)


mvasiljevicTT:
cc @odjuricicTT

odjuricicTT (approved):
Nice!

This will affect single chip models as well? Can you run benchmarks for those?

mtopalovicTT (approved):

mvasiljevicTT:
> Nice!
> 
> This will affect single chip models as well? Can you run benchmarks for those?

Sure, added now

ndrakulicTT (approved):
`tools/golden/mapping.py` looks good

PR #8037, opened 2026-04-20
[Optimizer] Add RmsNormRuleBook with sharded-input scoring override

rpavlovicTT:
## Summary
- Adds `OpRuleBook::adjustScore` hook so per-op rules can rewrite a candidate's `LayoutScore` after base scoring, with access to the output `OpConfig`, all input layouts, and the `requiresReshard` flag.
- Adds `RmsNormRuleBook` with an input layout filter (rejects height-sharded and partial-bbox sharded operand-0 inputs) and an `adjustScore` override that makes rms_norm score candidates based on kernel-dispatch reality instead of output sharding.
- Adds `isFullBboxSharded` helper in `LayoutFilterUtils.h` — checks whether a sharded layout's core grid forms a full rectangular bounding box (required by the sharded layernorm kernel's TT_FATAL at `layernorm_device_operation.cpp:173`).

## Why

The generic `LayoutScore` derives `coreCount` from the output layout, but sharding-gated kernels (layernorm/rms_norm) key their parallelism off the **input** layout:
- sharded input → `LayerNormShardedProgramFactory`, uses `input.grid.volume` cores (full-bbox required)
- interleaved input → `LayerNormMultiCoreProgramFactory`, tile-row-parallel (parallelism = `NC*Ht` of input shape)

Previously the lexicographic `LayoutScore` order ranked `requiresReshard=false` above `coreCount`, so an interleaved-input candidate with a high-core sharded **output** beat a sharded-input candidate with a reshard — even though the latter is the strictly faster kernel path. The 2D decode-shape case was the worst: kernel actually ran on 1 core despite the output being `width_sharded/1x45`.

`RmsNormRuleBook::adjustScore` fixes this by:
1. On sharded input: set `coreCount = input.grid.volume` and clear `requiresReshard` (kernel speedup dominates ~5–10 µs reshard cost on decode-shape tensors).
2. On interleaved input: set `coreCount = NC * Ht` of the input shape (the real kernel parallelism, capped by `split_work_to_cores`).

The filter prunes sharded candidates the kernel would reject at op-model validation, saving compile time.

## Test plan

- [x] gpt_oss_20b_decode_2lyr recompile: all 5 rms_norms on sharded input
- [x] tt-xla perf benchmark (n150, LLMs): 0 regressions, 7 improvements

### DIFF
--- a/include/ttmlir/Dialect/TTNN/Analysis/OpRules/NormalizationRules.h
+++ b/include/ttmlir/Dialect/TTNN/Analysis/OpRules/NormalizationRules.h
@@ -1,0 +1,63 @@
+// SPDX-FileCopyrightText: (c) 2026 Tenstorrent AI ULC
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#ifndef TTMLIR_DIALECT_TTNN_ANALYSIS_OPRULES_NORMALIZATIONRULES_H
+#define TTMLIR_DIALECT_TTNN_ANALYSIS_OPRULES_NORMALIZATIONRULES_H
+
+#include "ttmlir/Dialect/TTNN/Analysis/OpRules/OpRuleBook.h"
+
+namespace mlir::tt::ttnn {
+
+//===----------------------------------------------------------------------===//
+// Normalization ops: RMSNorm (LayerNorm has the same kernel dispatch logic
+// and could share rules in a follow-up).
+//
+// Kernel dispatch (layernorm_device_operation.cpp:16-22):
+//   input.is_sharded() ? LayerNormShardedProgramFactory
+//                      : LayerNormMultiCoreProgramFactory
+//
+// The sharded kernel's validator enforces
+// (layernorm_device_operation.cpp:148-173):
+//   - input cannot be HEIGHT_SHARDED
+//   - shard grid must be a full rectangular bounding box
+//     (shard_spec.grid.num_cores() == bbox_num_cores)
+//   - output must also be sharded (not height sharded)
+//
+// Without an input filter, partial-grid sharded candidates reach op-model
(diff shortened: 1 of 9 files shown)


include/ttmlir/Dialect/TTNN/Analysis/OpRules/NormalizationRules.h
```diff
@@ -0,0 +1,63 @@
...
+  /// sharded-input candidate is the strictly faster runtime path.
+  ///
+  /// The adjustment also intentionally lets a reshard-bearing candidate win
+  /// over a no-reshard alternative when the reshard produces a full-bbox
+  /// sharded input: the generic LayoutScore lexicographic order ranks
+  /// `requiresReshard=false` above `coreCount`, which blocks sharded-input
+  /// candidates whose only cost vs an interleaved-input peer is the inserted
+  /// reshard. For rms_norm the kernel speedup (~3x+ on decode shapes)
+  /// dominates the reshard cost, so `requiresReshard` is cleared on
+  /// sharded-input candidates to prefer the better-parallelized path.
```
odjuricicTT, line 54: Does ti make sense to have some of this logic in general scoring path as well? Does sharded -> sharded beat perf of interleaved -> sharded in other ops?

odjuricicTT (approved):

include/ttmlir/Dialect/TTNN/Analysis/OpRules/NormalizationRules.h
rpavlovicTT: Remains to be investigated. Generally depends on how kernels are written and whether an op just reshards interleaved input to proper sharded layout, or it actually has some specific kernel which is slower like in this case of normalisation op. I will gather better info around this in near future.

PR #8040, opened 2026-04-20
[d2m] single-core eltwise -> reduction fusion on a single dim

ckaravasilisTT:
Adds initial support for fusing an elementwise `d2m.generic` producer into a reduction `d2m.generic` consumer along the existing elementwise-fusion path. Scope is deliberately narrow: single-core, a single reduction iterator on the consumer, reduction over a column dimension. Off by default, enabled via `enable-eltwise-reduction-fusion=true`.

### What's changed

- **`D2MElementwiseFusion`**
  - Renamed to `GenericFusion`.
  - `FuseD2MEltwiseReductionOpsPattern` that reuses the consumer-centric fused op builder.
  - Validates the consumer as a single-reduction-dim op and reuses the producer-only checks via shared helpers (`producerResultUsedOnlyByConsumer`, `passesSharedFusionShapeChecks`, `fuseOverOperand`), also applied to the existing elementwise pattern.

- **`DstRegisterAnalysis`**
  - Per-result `numTilesPerResult` added to `DSTPackingPerResultInfo` so heterogeneous-output fused regions can drive Phase 1 capacity per op.
  - New `trySharedParallelChunking` fallback: when the GCD-based factorization collapses to a single outer iteration and outputs are heterogeneous (eltwise + reduction), chunks each op's shard along a shared factor of the row dim so Phase 1 produces matching outer loops across ops.
  - Relaxed the "identical num_tiles_per_result across ops" precondition; the region-level value is kept for back-compatibility.

- **`D2MInsertSpillAndScratch`**: `getConsumerScratchAccessMap` gains a row-cross-step path (consumer `scratchLoop` step != producer's), complementing the existing column cross-step case.

- **`D2MGenericTileComputeLoops`**: Phase-1 tiling now uses the per-result `numTilesPerResult` instead of the region-level value.

- **`D2M_TileFillOp`** marked with `D2M_SkipOpEltwiseFusionTrait`.

- Builder test in `test/python/golden/test_metal_reduction_fusion.py`
- Lit test `eltwise_reduction_fusion.mlir‎`


lib/Dialect/D2M/Transforms/GenericFusion.cpp
```diff
@@ -557,6 +579,132 @@ static GenericOp createFusedGeneric(OpOperand *fusedOperand, GenericOp producer,
...
+  if (!isValidReductionFusionConsumer(consumer)) {
+    return false;
+  }
+
+  if (!producerResultUsedOnlyByConsumer(producer, consumer)) {
+    return false;
+  }
+
+  return passesSharedFusionShapeChecks(fusionTargetOperand, producer, consumer);
+}
```
sgholamiTT, line 661: `isEltwiseReductionFusable(fusionTargetOperand)` looks like a call to `isElementwiseFusable(fusionTargetOperand)`, no ?
(I think we can get rid of `unsigned dstCapacity` arg in  `isElementwiseFusable` as it's not used).

lib/Dialect/D2M/Utils/DstRegisterAnalysis.cpp
```diff
@@ -153,6 +153,115 @@ struct PendingDSTPackingResult {
   int64_t numDstFlips = 0;
 };
 
+// Fallback chunking for regions with heterogeneous output shapes (e.g. eltwise
+// fused with a reduction), where the GCD-of-numDstFlips factorization
+// collapses to a single outer iteration. Splits each op's shard along a shared
+// row factor of outputShape[0] so Phase 1 produces matching outer loops that
+// SpillAndScratch can fuse.
+//
+// TODO(ckaravasilisTT): derive the shared parallel dim from the affine
```
sgholamiTT, line 163: Nit: Link an issue for this ?

sgholamiTT (commented):
Overal LGTM 👍🏻 
Please just add lit tests for all the changes in `lib/Dialect/D2MTransforms/`.

sgholamiTT (approved):

lib/Dialect/D2M/Transforms/GenericFusion.cpp
ckaravasilisTT: I agree they are very similar but I'm inclined to leave them separate for now. Validation of the consumer differs, and abstracting it into a single function (with additional parameters) I think might be confusing.

I agree about `dstCapacity`.

lib/Dialect/D2M/Utils/DstRegisterAnalysis.cpp
ckaravasilisTT: I will create a number of follow-up issues.

xanderchin (approved):

include/ttmlir/Dialect/D2M/Pipelines/D2MPipelines.h
```diff
@@ -62,6 +62,13 @@ struct D2MPipelineOptions : public PassPipelineOptions<D2MPipelineOptions> {
       llvm::cl::desc("Enable elementwise fusion of d2m.generic ops."),
       llvm::cl::init(false)};
 
+  Option<bool> enableEltwiseReductionFusion{
+      *this, "enable-eltwise-reduction-fusion",
```
vwellsTT, line 66: nit: fwiw, I think this name feels a bit inconsistent vs "enable-elementwise-fusion" above (since this seems to be a single eltwise w/ a single reduction, whereas above it's mass fusion).  Perhaps above should be renamed to "enable-elementwise-chain-fusion" or something?

lib/Dialect/D2M/Utils/DstRegisterAnalysis.cpp
```diff
@@ -285,14 +412,8 @@ computeDSTPackingForRegion(d2m::GenericOp generic,
...
     int64_t numTilesPerResult = numDstFlips * pending.numTilesPerFlip;
-    if (!commonNumTilesPerResult) {
-      commonNumTilesPerResult = numTilesPerResult;
-    }
-    if (*commonNumTilesPerResult != numTilesPerResult) {
-      generic.emitOpError(
-          "expected identical num tiles per result for all linalg.generic "
-          "outputs");
-      return std::nullopt;
+    if (!regionNumTilesPerResult) {
```
vwellsTT, line 415: the intent is just to set this for the first iteration only I guess? maybe a comment explaining why would be good

vwellsTT (approved):

include/ttmlir/Dialect/D2M/Pipelines/D2MPipelines.h
ckaravasilisTT: Yes, the intended scope quite a bit and this isn't obvious from the name. I will look into updating this.

lib/Dialect/D2M/Utils/DstRegisterAnalysis.cpp
ckaravasilisTT: Yes, it records the first result’s value only, but this is only kept for backwards compatibility. `DSTPackingRegionInfo::numTilesPerResult` is no longer used to set phase-1 loops, `GenericTileComputeLoops` uses `DSTPackingPerResultInfo::numTilesPerResult` instead. `DSTPackingRegionInfo::numTilesPerResult` can probably be removed in a future cleanup.

PR #8041, opened 2026-04-21
[d2m] rand lowering

dlokeTT:
### Problem description
Need to support `ttir.rand` in d2m for sampling graphs in models

### What's changed
- Added lowering thru ttmetal pipeline for `ttir.rand` including typecasting rand from `f32` to output dtype if a different dtype is specified. Also had each core use their grid position + seed to produce different values across different tiles following ttnn protocol.
- Updated element type normalization to normalize both the dtype and the attribute because it was causing errors with `i32` and `si32` mismatch
- Added lit tests for different dtypes for rand
- Added builder test file for metal rand including special harnesses for checking rand produces random values

### Checklist

### DIFF
--- a/lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
+++ b/lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
@@ -27,6 +27,7 @@
 #include "mlir/IR/Value.h"
 #include "mlir/Transforms/DialectConversion.h"
 
+#include <cstdint>
 #include <type_traits>
 #include <utility>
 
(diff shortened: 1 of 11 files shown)


lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
```diff
@@ -2537,6 +2537,114 @@ class D2MConstantFillOpRewriter : public OpConversionPattern<ConcreteOp>,
...
+
+          return {linalgGeneric.getResult(0)};
+        });
+
+    mlir::Value result =
+        unLayoutResult(rewriter, generic->getResult(0), randResultType)
+            ->getResult(0);
+
+    if (needsCast) {
+      result = rewriter.create<ttir::TypecastOp>(loc, finalResultType, result);
```
vwellsTT: I'm not certain this is guaranteed to work with this `applyPartialConversion` style pass, from what I read there doesn't seem to be a guarantee it will re-execute the ttir::TypecastOp rewriter if you re-emit one in another rewriter.

lib/Dialect/TTIR/IR/TTIROps.cpp
```diff
@@ -793,7 +793,17 @@ ::mlir::LogicalResult mlir::tt::ttir::RandOp::verify() {
   auto dtype = getDtype();
   auto outputType = getResult().getType().getElementType();
 
-  if (dtype != outputType) {
+  // Normalize both sides so that pre/post `ttir-element-type-normalization`
```
vwellsTT: why does this happen, given you haven't changed element type normalization here?

lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
```diff
@@ -1618,6 +1619,71 @@ class D2MWriteColMaskTileRewriter
...
+    auto i32Ty = rewriter.getI32Type();
+    auto i32FromBits = [&](uint32_t bits) -> Value {
+      return rewriter
+          .create<arith::ConstantOp>(
+              loc, i32Ty,
+              rewriter.getIntegerAttr(i32Ty, llvm::APInt(/*numBits=*/32, bits)))
+          .getResult();
+    };
+
+    // Per-core seed: user_seed XOR (x*C1) XOR (y*C2) with hash-prime
```
vwellsTT: confused what x vs C1 and y vs C2 are, can you expand?

dlokeTT: x and y are the cores logical coordinates. C1 and C2 were constants we were using to turn the coordinates into an even distribution across [0, 2^32) so each core is getting a different seed. This is following what ttnn does. I updated comment to be more descriptive

lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
dlokeTT: Claude is saying that since we marked the ttir dialect as illegal, the pass will make sure the typecast is converted but changing it to d2m just to simplify

lib/Dialect/TTIR/IR/TTIROps.cpp
dlokeTT: yeah, just updated element type norm instead of this. was a Claude hotfix

lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
vwellsTT: that makes sense mathematically but I think the point I was confused by is that the kPrimeX/Y values are just hardcoded? Is there some justification for the values?
Regardless, I suppose it's fine if TTNN does the same 👍

dlokeTT: Ahh, yeah. I asked deepwiki and it said those are just some special numbers that are good to be used in whatever formula ttnn picked to make sure each core gets a different seed. Don't really know too much besides that

lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
```diff
@@ -1618,6 +1619,78 @@ class D2MWriteColMaskTileRewriter
...
+        rewriter.create<arith::MulIOp>(loc, xI32, i32FromBits(kPrimeX));
+    Value mulY =
+        rewriter.create<arith::MulIOp>(loc, yI32, i32FromBits(kPrimeY));
+    Value seedWithX = rewriter.create<arith::XOrIOp>(loc, userSeed, mulX);
+    Value effSeed = rewriter.create<arith::XOrIOp>(loc, seedWithX, mulY);
+    rewriter.create<ttkernel::RandTileInitOp>(loc, effSeed);
+    rewriter.setInsertionPoint(insertionPoint->getBlock(), insertionPoint);
+
+    Value fromVal =
+        i32FromBits(llvm::bit_cast<uint32_t>(op.getFrom().convertToFloat()));
```
vwellsTT: why cast these to i32 at all?

dlokeTT: Cleaned this up a bit. Now we just take the APFloat from the f32 attribute and bit cast it to an int. We need to do this because the llk takes a uint32 that represents the float value

vwellsTT: I wonder why ha, seems inconvenient for now reason.  But oh well

vwellsTT (approved):

lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
dlokeTT: This is littered all over the llks, I have no idea why. Hopefully it gets cleaned up

PR #8054, opened 2026-04-21
RoPE fusing false positive

### LINKED ISSUE #8042: [TTNN] Expanded RoPE fusing pattern produces incorrect concat at batch=1, corrupting decode output
### Problem description

The RoPE fusing pattern introduced in #7751 (`16ec8f80f`) generates incorrect `ttnn.concat` ops in the decode graph when `batch=1` and `optimization_level=1`. The result is garbage output from Llama models at batch=1 while batch=2 produces correct output.

Reported in tt-xla: https://github.com/tenstorrent/tt-xla/issues/4325

### Bisect result

- **First bad commit**: `16ec8f80fbc34da0fab3fd847d94aca240b14082` — `[TTNN] Add expanded (trig-identity) RoPE fusing pattern (#7751)`
- **Last good commit**: `c5575fc46d70ee245a36d69488e0a3781eb269a0` — `[EmitPy] Refactor EmitPy CPU-hoisting (#7706)`

### Root cause theory

In the bad commit's decode graph, the RoPE concat ops are passed the same tensor twice instead of the two complementary halves:

```
// Bad (batch=1, opt_level=1)
%26 = "ttnn.concat"(%23, %23) <{dim = 3 : si32}>   // sin concat'd with itself
%27 = "ttnn.concat"(%25, %25) <{dim = 3 : si32}>   // cos concat'd with itself
```

```
// Good (last good commit)
%34 = "ttnn.concat"(%30, %33) <{dim = 3 : si32}>   // two distinct tensors
```

The fusing pattern appears to incorrectly CSE or alias the two RoPE half-arguments at batch=1, causing both inputs to the concat to resolve to the same value. This corrupts the rotary embeddings for every transformer layer in the decode step.

Prefill is unaffected (prefill logits are identical between good and bad commits). Only the decode forward pass at batch=1 is broken.

### Conditions

- **Model**: Llama-3.2-3B (and likely other Llama/RoPE-based models)
- **Batch size**: 1 only — batch=2 produces correct output
- **optimization_level**: 1 only — opt_level=0 is correct

### Decode graph IR (before/after bad commit)

https://gist.github.com/kmabeeTT/a6f8b027d1d87e24749c0d7642cb1441

Contains `graph_3_ttir.mlir` and `graph_3_ttnn.mlir` for both the good and bad commits. The relevant concat ops are around line 301 in the bad ttnn graph.

djstefanovicTT:
### Ticket
closes https://github.com/tenstorrent/tt-mlir/issues/8042

### Problem description
Fusing pattern for rope was triggered in situations that are not valid.

### What's changed
Added stricter check for rope fusing pattern.

### DIFF
--- a/lib/Dialect/TTNN/Transforms/Fusing/RoPEFusingPattern.cpp
+++ b/lib/Dialect/TTNN/Transforms/Fusing/RoPEFusingPattern.cpp
@@ -318,9 +318,10 @@ private:
 // ---------------------------------------------------------------------------
 
 // Skip transparent TM ops (typecast, permute) to reach the semantic source.
+template <typename... AllowedOps>
 Value skipTMs(Value v) {
   while (Operation *defOp = v.getDefiningOp()) {
-    if (isa<TypecastOp, PermuteOp>(defOp)) {
+    if (isa<AllowedOps...>(defOp)) {
       v = defOp->getOperand(0);
     } else {
       break;
@@ -329,6 +330,8 @@ Value skipTMs(Value v) {
 
   return v;
 }
+
+Value skipTMs(Value v) { return skipTMs<TypecastOp, PermuteOp>(v); }
 
 // Collect a value and all predecessors reachable through TM ops.
 // Includes RepeatOp so that pre-broadcast values (with size-1 dims) are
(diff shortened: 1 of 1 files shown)


lib/Dialect/TTNN/Transforms/Fusing/RoPEFusingPattern.cpp
```diff
@@ -330,6 +330,19 @@ Value skipTMs(Value v) {
   return v;
 }
 
+// skip all TM ops and return the result
+Value skipAllTMs(Value v) {
+  while (Operation *defOp = v.getDefiningOp()) {
+    if (isa<TypecastOp, PermuteOp, RepeatOp, ReshapeOp>(defOp)) {
```
sdjordjevicTT: On the TTIR level, we have the TTIR_TensorManipulation trait. Maybe we should do something similar here?

sdjordjevicTT (approved):

kmabeeTT:
Thanks for PR - hit retry on failing job.  Testing this out on my side (Edit: thumbs up [comment](https://github.com/tenstorrent/tt-mlir/issues/8042#issuecomment-4291245347)).

PR #8061, opened 2026-04-22
[d2m] int32 dtype support for min, max and sum reductions

dlokeTT:
### Problem description
Need this for llms as there's an i32 sum.

### What's changed
- Added ttkernel ops for `sfpu_reduce` and `sfpu_reduce_init` and d2m ops for reducing integers since they go down the sfpu path, not fpu.
- Add handling for int32 for writing padding in experimental llks
- Integer handling in `decomposeMasking.cpp`
- Added builder tests for i32 including unaligned. Also adding deterministic cycling between attrs like keep_dim, dtype, reduce_dim to not bloat test suite
- Added lit tests for all extensions to reductions

### Checklist

### DIFF
--- a/include/ttmlir/Dialect/D2M/IR/D2MOpsInterfaces.td
+++ b/include/ttmlir/Dialect/D2M/IR/D2MOpsInterfaces.td
@@ -196,9 +196,10 @@ def D2M_OperandLoadStoreRegisterOpInterface : OpInterface<"OperandLoadStoreRegis
     InterfaceMethod<
       /*desc=*/[{
         Returns the number of DST scratch slices this op needs beyond its
-        operand/output slices. The insert-dst-register-access pass reserves
-        this many slices from the DST allocator and records their indices on
-        the op (e.g. via `dst_scratch_index`) for the lowering to consume.
+        operand/output slices. Only 0 or 1 are supported today: when 1, the
+        insert-dst-register-access pass reserves a slice from the DST
+        allocator and records its index on the op via `dst_scratch_index`
+        for the lowering to consume.
       }],
       /*retTy=*/"int64_t",
       /*methodName=*/"getNumDstScratchSlices",
--- a/include/ttmlir/Dialect/D2M/IR/D2MOpsInterfaces.td
+++ b/include/ttmlir/Dialect/D2M/IR/D2MOpsInterfaces.td
@@ -193,6 +193,19 @@ def D2M_OperandLoadStoreRegisterOpInterface : OpInterface<"OperandLoadStoreRegis
         return false;
       }]
     >,
+    InterfaceMethod<
+      /*desc=*/[{
+        Returns the number of DST scratch slices this op needs beyond its
+        operand/output slices. The insert-dst-register-access pass reserves
+        this many slices from the DST allocator and records their indices on
+        the op (e.g. via `dst_scratch_index`) for the lowering to consume.
+      }],
+      /*retTy=*/"int64_t",
+      /*methodName=*/"getNumDstScratchSlices",
+      /*args=*/(ins),
+      /*methodBody=*/"",
+      /*defaultImplementation=*/"return 0;"
+    >,
   ];
 }
 
--- a/include/ttmlir/Dialect/TTKernel/IR/TTKernelOps.td
+++ b/include/ttmlir/Dialect/TTKernel/IR/TTKernelOps.td
@@ -661,6 +661,9 @@ def TTKernel_SFPUReduceInitOp : TTKernel_InitOp<"sfpu_reduce_init"> {
       Unlike reduce_init this call carries no CB/DST operands: it only
       configures the SFPU for the given PoolType and DataFormat.
 
+      Only Sum/Max reductions over Int32 tiles are supported today; the
+      verifier enforces this.
+
       Lowers to: sfpu_reduce_init<PoolType, DataFormat>().
     }];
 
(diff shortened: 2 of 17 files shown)


lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
```diff
@@ -956,6 +956,117 @@ class D2MFPUOpsRewriter : public OpConversionPattern<ConcreteOp> {
...
+    ttcore::DataType dataFormat = inTileType.getDataType();
+
+    auto insertionPoint = rewriter.getInsertionPoint();
+    Value cbA = getCB(rewriter, op.getA());
+    Value outCB = getOutCB(rewriter, op);
+    Value cIdx = adaptor.getC();
+
+    // cIdx is defined inside the scf loops, so re-materialize a constant
+    // copy for use at the hoisted init point.
+    auto cIdxConst = cIdx.template getDefiningOp<arith::ConstantOp>();
```
anuragsingh-tt: Should be able to use `getConstantIntValue` here instead.

lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
```diff
@@ -956,6 +956,117 @@ class D2MFPUOpsRewriter : public OpConversionPattern<ConcreteOp> {
...
+      kernelReduceDim = ttkernel::ReduceDim::Scalar;
+      break;
+    }
+
+    // i32 element type is enforced by the op verifier; fill_tile_int and
+    // binary_max_int32_tile below rely on that invariant.
+    const auto inTileType = mlir::cast<ttcore::TileType>(op.getA().getType());
+    ttcore::DataType dataFormat = inTileType.getDataType();
+
+    auto insertionPoint = rewriter.getInsertionPoint();
```
anuragsingh-tt: Should be able to use `InsertionGuard` here instead.

include/ttmlir/Dialect/D2M/IR/D2MOpsInterfaces.td
```diff
@@ -193,6 +193,19 @@ def D2M_OperandLoadStoreRegisterOpInterface : OpInterface<"OperandLoadStoreRegis
...
     >,
+    InterfaceMethod<
+      /*desc=*/[{
+        Returns the number of DST scratch slices this op needs beyond its
+        operand/output slices. The insert-dst-register-access pass reserves
+        this many slices from the DST allocator and records their indices on
+        the op (e.g. via `dst_scratch_index`) for the lowering to consume.
+      }],
+      /*retTy=*/"int64_t",
+      /*methodName=*/"getNumDstScratchSlices",
```
anuragsingh-tt, line 205: The PR says later that "setDstScratchIndex supports exactly one scratch slice" so can we narrow this description?

lib/Dialect/D2M/IR/D2MGenericRegionOps.cpp
```diff
@@ -1747,6 +1747,59 @@ void TileUntilizeBlockOp::getEffects(
...
+}
+
+template <typename OpT>
+static mlir::LogicalResult verifySFPUTileReduce(OpT op) {
+  // The SFPU reduce lowering uses i32-only TTKernel ops (fill_tile_int,
+  // binary_max_int32_tile, add_int_tile), so restrict this op to i32 tiles.
+  auto isI32Tile = [](mlir::Value v) {
+    auto intType = mlir::dyn_cast<mlir::IntegerType>(
+        mlir::cast<mlir::tt::ttcore::TileType>(v.getType()).getElementType());
+    return intType && intType.getWidth() == 32;
```
anuragsingh-tt: This should be signed.

include/ttmlir/Dialect/TTKernel/IR/TTKernelOps.td
```diff
@@ -654,6 +654,54 @@ def TTKernel_ReduceTileOp : TTKernel_FPUOp<"reduce_tile", [TTKernel_TernaryOpTra
...
+
+    let arguments = (ins TTKernel_ReduceTypeAttr:$reduce_type,
+                         TTCore_DataTypeAttr:$data_format);
+
+    let assemblyFormat = [{
+      `(` $reduce_type `,` $data_format `)` attr-dict `:` functional-type(operands, results)
+    }];
+}
+
+def TTKernel_SFPUReduceTileOp : TTKernel_SFPUOp<"sfpu_reduce"> {
```
anuragsingh-tt, line 680: Lets add verifiers to these.

lib/Dialect/D2M/Transforms/InsertDstRegisterAccess.cpp
```diff
@@ -871,16 +891,10 @@ struct D2MInsertDstRegisterAccessRewriter final
...
-            bool isReduction = mlir::isa<d2m::TileReduceMaxOp>(computeOp) ||
-                               mlir::isa<d2m::TileReduceSumOp>(computeOp) ||
-                               mlir::isa<d2m::TileReduceMeanOp>(computeOp);
-            assert(
-                (isUnaryOp || isTileMatmul || isReduction || rhsIsScalar) &&
-                "Only unary ops, tile matmul, reductions, and tile+scalar ops "
-                "supported for destination register in place, multi-operand "
-                "ops "
-                "would reference wrong tile, but those ops should be setting "
-                "output tile.");
```
anuragsingh-tt, line 883: Would e.g. a subsequent unary op reference the right tile after this change to put scratch at the back?

dlokeTT: Good catch. If we fused a unary op onto the back of this reduction, it wouldn't. I'll add a comment about it and also filed an issue: https://github.com/tenstorrent/tt-mlir/issues/8081. This would only happen for int reductions though and there's only one I'm seeing in the models and a typecast follows it which we're currently not even fusing. Definitely need to fix this but those changes are pretty extensive I think and hopefully we can approach this pass better instead of dumping more special casing.

anuragsingh-tt (approved):

PR #8064, opened 2026-04-22
[StableHLOToTTIR] Decompose gather with multiple partial-slice indexed dims

### LINKED ISSUE #8062: Vilt QA : error: failed to legalize operation 'stablehlo.gather'
### Describe the bug

- Vilt model on QA task fails with `error: failed to legalize operation 'stablehlo.gather'`
- Issue originates from this [interpolate](https://github.com/huggingface/transformers/blob/7d9754a05193eb79b1d86aa744b622b8068008cd/src/transformers/models/vilt/modeling_vilt.py#L106) op

### Repro

In tt-xla repo, 

```
git checkout kkannan/apr22_vilt_qa_gather_issue
git submodule update --init --recursive

pytest tests/runner/test_models.py::test_all_models_torch[vilt/question_answering/pytorch-Vqa-single_device-inference] -svv
pytest tests/torch/ops/test_interpolate.py -svv
```

### Logs

- [apr22_vilt_qa.log](https://github.com/user-attachments/files/26956310/apr22_vilt_qa.log)
- [apr22_interpolate.log](https://github.com/user-attachments/files/26956314/apr22_interpolate.log)

kamalrajkannan78:
### Ticket

- fixes https://github.com/tenstorrent/tt-mlir/issues/8062 
- tenstorrent/tt-mlir#5655

### Problem description
`StableHLOGatherToEmbeddingPattern` explicitly bails on gathers that have two or more indexed dimensions with partial slice sizes (`1 < sliceSize < dimSize`):

```cpp
if (partialIndexedDims && (remainingIndexedDims != 1)) {
  return rewriter.notifyMatchFailure(srcOp, "...");
}
```
Because the other three existing gather patterns don't cover this shape either, any such gather fails StableHLO→TTIR legalization with failed to legalize operation 'stablehlo.gather'.

### What's changed

Added a new sibling pattern `StableHLOGatherMultiPartialFlattenPattern` that generalizes the existing `StableHLOGatherToEmbeddingPattern` to the multi-partial case, reusing its machinery rather than duplicating it:

1. Flatten the contiguous block of partial indexed dims in the operand into a single dim (e.g. 1×12×12×768 → 1×144×768) via the shared `permuteInput` + `reshapeInput` helpers.
2. Combine multi-component indices into a 1-D flat index (e.g. h*W + w) via the shared `flattenStartIndices` + `reshapeStartIndices` helpers.
3. Build one N-D window-offset constant and broadcast-add it to the flat indices to produce expanded indices of shape [prod(windowSizes), M] — a new shared helper `expandStartIndicesMultiDim` generalizes the existing 1-D `expandStartIndices`.
4. Emit a **single** `ttir.embedding` call, then reshape + permute the output to place the window axes.

The existing `StableHLOGatherToEmbeddingPattern` is unchanged behaviourally; its private helpers were moved to `public` so the new pattern can reuse them,and one new static helper (`expandStartIndicesMultiDim`) was added alongside them. The new pattern is guarded so it only fires when ≥ 2 partial indexed dims are present and defers to existing patterns otherwise.

Match conditions (all must hold):
- At least two partial indexed dims
- No batching dims, no collapsed_slice_dims
- Non-indexed dims fully sliced
- All indexed dims are partial (no mixing with full-N/N)
- Partial indexed dims are contiguous in operand layout
- offset_dims are the leading output positions

### Checklist

### Logs

- [apr22_interpolate_before_fix.log](https://github.com/user-attachments/files/26961797/apr22_interpolate.log)
- [apr24_interpolate_after_fix.log](https://github.com/user-attachments/files/27057849/apr24_interpolate_after_final_fix.log)

### DIFF
--- a/lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
+++ b/lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
@@ -21,6 +21,7 @@
 #include "mlir/Dialect/Func/IR/FuncOps.h"
 #include "mlir/Dialect/Func/Transforms/FuncConversions.h"
 #include "mlir/Dialect/Traits.h"
+#include "mlir/Dialect/Utils/IndexingUtils.h"
 #include "mlir/IR/Builders.h"
 #include "mlir/IR/BuiltinAttributeInterfaces.h"
 #include "mlir/IR/BuiltinAttributes.h"
@@ -4821,7 +4822,7 @@ public:
     return success();
   }
 
-private:
+public:
   // In StableHLO, startIndexMap attribute refers to which dims of input
   // we are indexing (with startIndices). We need these dims to be
   // flattened together to be the first dim of transformed input (that is
@@ -5106,6 +5107,81 @@ private:
                               expectedOutputType.getElementType(),
                               expectedOutputType.getEncoding()),
         reshapedOutput, inverseOutputPermutation);
+  }
+
+  // Multi-dimensional generalization of expandStartIndices. Expands
+  // startIndices [1, M] to [prod(windowSizes), M] where row r carries a flat
+  // offset built from decoding r (row-major) over windowSizes and scaling each
+  // component by the corresponding stride of the flattened indexed operand.
+  // Example: windowSizes=[3,3], partialDimStrides=[12,1]
+  //   row=0 (dh=0,dw=0) -> offset=0
+  //   row=1 (dh=0,dw=1) -> offset=1
+  //   row=3 (dh=1,dw=0) -> offset=12
+  //   row=8 (dh=2,dw=2) -> offset=26
+  static ttir::AddOp
+  expandStartIndicesMultiDim(ConversionPatternRewriter &rewriter, Location loc,
+                             Value startIndices, ArrayRef<int64_t> windowSizes,
+                             ArrayRef<int64_t> partialDimStrides) {
+    auto startIndicesType =
+        mlir::cast<RankedTensorType>(startIndices.getType());
+    auto startIndicesShape = startIndicesType.getShape();
+    assert(startIndicesShape.size() == 2 && startIndicesShape[0] == 1 &&
+           "expandStartIndicesMultiDim expects 2D startIndices [1, M]");
+    assert(windowSizes.size() == partialDimStrides.size());
+
+    int64_t totalWindow = 1;
+    for (int64_t ws : windowSizes) {
+      totalWindow *= ws;
+    }
+    int64_t M = startIndicesShape[1];
+
+    // Compute the flat offset for each window position (row-major over
+    // windowSizes).
+    SmallVector<int64_t> windowOffsets(totalWindow);
+    for (int64_t row = 0; row < totalWindow; ++row) {
+      int64_t flatOffset = 0;
+      int64_t remaining = row;
+      for (int64_t d = static_cast<int64_t>(windowSizes.size()) - 1; d >= 0;
+           --d) {
+        int64_t off_d = remaining % windowSizes[d];
+        remaining /= windowSizes[d];
+        flatOffset += off_d * partialDimStrides[d];
+      }
+      windowOffsets[row] = flatOffset;
+    }
+
+    SmallVector<int64_t> expandedShape = {totalWindow, M};
+    auto elemType = startIndicesType.getElementType();
+    auto expandedType = RankedTensorType::get(expandedShape, elemType,
+                                              startIndicesType.getEncoding());
+
+    // Build offset matrix [totalWindow, M]: row r has all elements =
+    // windowOffsets[r]. The caller routes startIndices through
+    // flattenStartIndices, which always produces float-typed indices, so we
+    // only need the float construction path here.
+    SmallVector<float> matrixData(totalWindow * M);
+    for (int64_t row = 0; row < totalWindow; ++row) {
+      float v = static_cast<float>(windowOffsets[row]);
+      for (int64_t col = 0; col < M; ++col) {
+        matrixData[row * M + col] = v;
+      }
+    }
+    auto offsetAttr =
+        DenseElementsAttr::get(expandedType, llvm::ArrayRef(matrixData));
+
+    auto offsetConstant = rewriter.create<ttir::ConstantOp>(
+        ttmlir::utils::appendLocationSuffix(loc, "_windowOffsetConstant"),
+        expandedType, offsetAttr);
+
+    SmallVector<int64_t> broadcastDimensions = {totalWindow, 1};
+    auto broadcastedStartIndices = rewriter.create<ttir::BroadcastOp>(
+        ttmlir::utils::appendLocationSuffix(loc, "_broadcastStartIndices"),
+        expandedType, startIndices,
+        rewriter.getDenseI64ArrayAttr(broadcastDimensions));
+
+    return rewriter.create<ttir::AddOp>(
+        ttmlir::utils::appendLocationSuffix(loc, "_expandedStartIndices"),
+        expandedType, broadcastedStartIndices, offsetConstant);
   }
 };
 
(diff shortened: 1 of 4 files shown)


meenakshiramanathan1 (approved):

lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
```diff
@@ -5620,6 +5620,415 @@ class StableHLOGatherToSliceRepeatConcatPattern
...
+
+    auto operandType =
+        mlir::cast<RankedTensorType>(srcOp.getOperand().getType());
+    auto indicesType =
+        mlir::cast<RankedTensorType>(srcOp.getStartIndices().getType());
+    auto operandShape = operandType.getShape();
+    auto indicesShape = indicesType.getShape();
+
+    // Identify partial indexed dims (1 < sliceSize < inputDim), in operand
+    // order. They must appear in the same order in start_index_map for the
```
anuragsingh-tt: Is startIndexMap == partialDims? Can we verify this?

lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
```diff
@@ -5620,6 +5620,415 @@ class StableHLOGatherToSliceRepeatConcatPattern
...
+    if (!indicesElemType.isIntOrIndex()) {
+      return rewriter.notifyMatchFailure(
+          srcOp, "start_indices element type must be integer");
+    }
+
+    // Per-dim row-major strides over partialDims.
+    SmallVector<int64_t> strides(partialDims.size(), 1);
+    for (int64_t k = static_cast<int64_t>(partialDims.size()) - 2; k >= 0;
+         --k) {
+      strides[k] = strides[k + 1] * operandShape[partialDims[k + 1]];
```
anuragsingh-tt: Let's use computeStrides here instead.

lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
```diff
@@ -5620,6 +5620,415 @@ class StableHLOGatherToSliceRepeatConcatPattern
...
+          loc, sliceOutType, scalarConst,
+          rewriter.getDenseI64ArrayAttr(bcastDims));
+    };
+
+    // flat = sum_k (componentIndices[k] * stride[k])
+    Value flatIndices;
+    auto sliceOutType =
+        mlir::cast<RankedTensorType>(componentIndices[0].getType());
+    for (int64_t k = 0; k < indexVecSize; ++k) {
+      Value strideConst = broadcastToSliceShape(buildScalarConst(strides[k]));
```
anuragsingh-tt: We can precompute this instead of doing it every time in the loop.

lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
```diff
@@ -5620,6 +5620,415 @@ class StableHLOGatherToSliceRepeatConcatPattern
...
+    // sub-gather.
+    SmallVector<int64_t> windowSizes;
+    for (int64_t pd : partialDims) {
+      windowSizes.push_back(sliceSizes[pd]);
+    }
+
+    SmallVector<Value> subGatherResults;
+    SmallVector<int64_t> currentOffset;
+
+    std::function<void(int64_t)> enumerate = [&](int64_t dim) {
```
anuragsingh-tt: We can do this iteratively instead of recursively.

lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
```diff
@@ -5620,6 +5620,415 @@ class StableHLOGatherToSliceRepeatConcatPattern
...
+    // partial dims (except firstPD's axis, which becomes the merged axis)
+    // stripped. The simplest consistent choice: produce a rank-
+    // (flatOperandShape.size() + batchRank) tensor whose shape matches
+    // srcOp's output except the window axes have size 1. We'll emit at the
+    // rank = flatOperandShape.size() + (indices rank - 1) and then reshape.
+    //
+    // Concretely: start from srcOp output shape, for every partial dim's
+    // output axis set size 1, then drop the axes for all but the first
+    // partial dim. Equivalently, build subGatherShape = resultShape with
+    // window axes dropped except at firstPD's axis set to 1.
```
anuragsingh-tt: The comments for this function are a bit excessive. Please consider condensing.

lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
```diff
@@ -5620,6 +5621,420 @@ class StableHLOGatherToSliceRepeatConcatPattern
...
+        mlir::cast<RankedTensorType>(srcOp.getStartIndices().getType());
+    auto operandShape = operandType.getShape();
+    auto indicesShape = indicesType.getShape();
+
+    // Identify partial indexed dims (1 < sliceSize < inputDim), in operand
+    // order. They must appear in the same order in start_index_map for the
+    // flattening arithmetic to be well-defined without extra transposes.
+    SmallVector<int64_t> partialDims;
+    for (int64_t d : startIndexMap) {
+      if (operandShape[d] > 1 && sliceSizes[d] > 1 &&
```
sdjukicTT: Nit: `operandShape[d] > 1` is redundant here

lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
```diff
@@ -5620,6 +5621,420 @@ class StableHLOGatherToSliceRepeatConcatPattern
...
+// (each with a single indexed dim and slice size 1). The sub-gathers are
+// concatenated along the original window axes.
+//
+// Match conditions:
+// - At least two partial indexed dims
+// - No batching dims, no collapsed_slice_dims
+// - Non-indexed dims are fully sliced
+// - All indexed dims are partial (not mixed with full-N/N)
+// - Partial indexed dims are contiguous in operand layout
+// - offset_dims are the leading output positions
```
sdjukicTT, line 5715: Can we extract all of these checks to a checkBasicLegality helper? To make pattern more readable

lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
```diff
@@ -5620,6 +5621,420 @@ class StableHLOGatherToSliceRepeatConcatPattern
...
+    int64_t indexVecSize = static_cast<int64_t>(startIndexMap.size());
+    if (indexVectorDim >= static_cast<int64_t>(indicesShape.size())) {
+      return rewriter.notifyMatchFailure(
+          srcOp,
+          "Implicit trailing index_vector_dim not supported by this pattern");
+    }
+    if (indicesShape[indexVectorDim] != indexVecSize) {
+      return rewriter.notifyMatchFailure(
+          srcOp, "index_vector_dim size does not match start_index_map size");
+    }
```
sdjukicTT: Can we remove these indexVectorDim checks? By C3 of https://openxla.org/stablehlo/spec#gather, we know multi-component index_vector_dim must be explicit and index_vector_dim size must match start_index_map arity.

lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
```diff
@@ -5620,6 +5621,420 @@ class StableHLOGatherToSliceRepeatConcatPattern
...
+    Value flatOperand = rewriter.create<mlir::stablehlo::ReshapeOp>(
+        loc, flatOperandType, srcOp.getOperand());
+
+    // Compute flat indices: flat[..] = sum_k(indices[.., k] * stride[k])
+    // where stride[k] is the product of partial dim sizes to the right of k.
+    auto indicesElemType = indicesType.getElementType();
+    if (!indicesElemType.isIntOrIndex()) {
+      return rewriter.notifyMatchFailure(
+          srcOp, "start_indices element type must be integer");
+    }
```
sdjukicTT: The spec https://openxla.org/stablehlo/spec#gather (I2) already requires integer type, this is another verifier invariant being re-checked.

lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
```diff
@@ -5620,6 +5621,420 @@ class StableHLOGatherToSliceRepeatConcatPattern
...
+      return rewriter.create<mlir::stablehlo::ConstantOp>(loc, scalarType,
+                                                          attr);
+    };
+
+    // Broadcast a scalar-rank-matched const to the full slice shape of a
+    // single index component (same shape as componentIndices[k]).
+    auto broadcastToSliceShape = [&](Value scalarConst) -> Value {
+      auto sliceOutShape =
+          mlir::cast<RankedTensorType>(componentIndices[0].getType())
+              .getShape();
```
sdjukicTT: Nit: each invocation of the lambda re-derives `sliceOutShape` from `componentIndices[0].getType()` and rebuilds `sliceOutType`. The outer scope already caches `sliceOutType` right after the lambda. The lambda could just use that captured value instead of re-deriving.

lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
```diff
@@ -5620,6 +5621,420 @@ class StableHLOGatherToSliceRepeatConcatPattern
...
+    for (int64_t k = 0; k < indexVecSize; ++k) {
+      strideConsts.push_back(
+          broadcastToSliceShape(buildScalarConst(strides[k])));
+    }
+
+    // flat = sum_k (componentIndices[k] * stride[k])
+    Value flatIndices;
+    for (int64_t k = 0; k < indexVecSize; ++k) {
+      Value scaled = rewriter.create<mlir::stablehlo::MulOp>(
+          loc, sliceOutType, componentIndices[k], strideConsts[k]);
```
sdjukicTT: `strides[indexVecSize - 1]` is always 1 by construction of row-major strides, so the last component's multiply is always identity. Can we skip that MulOp?

lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
```diff
@@ -5620,6 +5621,420 @@ class StableHLOGatherToSliceRepeatConcatPattern
...
+      auto scalarType = RankedTensorType::get(scalarShape, indicesElemType);
+      APInt apVal(indicesBitwidth, static_cast<uint64_t>(v), /*isSigned=*/true);
+      auto attr =
+          DenseIntElementsAttr::get(scalarType, llvm::ArrayRef<APInt>{apVal});
+      return rewriter.create<mlir::stablehlo::ConstantOp>(loc, scalarType,
+                                                          attr);
+    };
+
+    // Broadcast a scalar-rank-matched const to the full slice shape of a
+    // single index component (same shape as componentIndices[k]).
```
sdjukicTT: Could we collapse the two helpers into one, building a ConstantOp of right shape from the beginning, without BroadcastInDimOp, something like: 
   ```
  auto buildBroadcastConst = [&](int64_t v) -> Value {                                                                                                                     
    auto sliceOutType =                                                                                                                                                    
        mlir::cast<RankedTensorType>(componentIndices[0].getType());
    APInt apVal(indicesBitwidth, static_cast<uint64_t>(v), /*isSigned=*/true);                                                                                             
    auto attr = DenseIntElementsAttr::get(sliceOutType, apVal);                                                                                                            
    return rewriter.create<mlir::stablehlo::ConstantOp>(loc, sliceOutType, attr);                                                                                          
  };
```

File: lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp [hunk 13]
```diff
@@ -5620,6 +5621,420 @@ class StableHLOGatherToSliceRepeatConcatPattern
...
+    // Offset_dims for sub-gather: since we require original offsetDims to be
+    // the leading positions [0, 1, ..., operandRank-1], the sub-gather's
+    // offset dims are simply the leading positions of the flat operand:
+    // [0, 1, ..., flatOperandRank-1]. Batch dims fill the trailing positions.
+    SmallVector<int64_t> subOffsetDims;
+    subOffsetDims.reserve(flatOperandShape.size());
+    for (int64_t i = 0; i < static_cast<int64_t>(flatOperandShape.size());
+         ++i) {
+      subOffsetDims.push_back(i);
+    }
```
sdjukicTT: The two loops (`subSliceSizes` and `subOffsetDims`) are small enough to inline, and the second one can be expressed with std::iota, something like:                                              
   
```
  SmallVector<int64_t> subOffsetDims(flatOperandShape.size());                                                                                                             
  std::iota(subOffsetDims.begin(), subOffsetDims.end(), 0);
```

lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
```diff
@@ -5620,6 +5621,420 @@ class StableHLOGatherToSliceRepeatConcatPattern
...
+        rewriter.getContext(),
+        /*offsetDims=*/subOffsetDims,
+        /*collapsedSliceDims=*/{},
+        /*operandBatchingDims=*/{},
+        /*startIndicesBatchingDims=*/{},
+        /*startIndexMap=*/{flatIndexedDim},
+        /*indexVectorDim=*/indexVectorDim);
+
+    // Sub-gather output type: srcOp result shape but with partial output axes
+    // collapsed to size 1 (same as Option A; we'll reshape them back later).
```
sdjukicTT: Seems like comment has reference to something that is no longer relevant ('same as Option A')

lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
```diff
@@ -5620,6 +5621,420 @@ class StableHLOGatherToSliceRepeatConcatPattern
...
+    auto resultType = mlir::cast<RankedTensorType>(srcOp.getType());
+    SmallVector<int64_t> partialDimOutputAxes;
+    for (int64_t pd : partialDims) {
+      partialDimOutputAxes.push_back(offsetDims[pd]);
+    }
```
sdjukicTT: We can assume `offsetDims[d] == d` because of the "offset dims are leading" guard. Under that guard, the lookup offsetDims[pd] is pd. So `partialDimOutputAxes` equals `partialDims`

lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
```diff
@@ -5620,6 +5621,420 @@ class StableHLOGatherToSliceRepeatConcatPattern
...
+
+    // Build sub-gather output shape: start from result shape, set all partial
+    // window axes to 1, then drop the extra window axes (all but the first),
+    // since the flat operand has one fewer dim per merged partial dim.
```
sdjukicTT: This can be done in less confusing way.  The sub-gather output is just `window shape + batch shape`, which the pattern already has:
```                                                                                                                                         
  SmallVector<int64_t> subGatherShape;                                                                                                                                     
  subGatherShape.reserve(subSliceSizes.size() + indicesShape.size() - 1);
                                                                                                                                                                           
  // Window axes: subSliceSizes already has 1 at flatIndexedDim, full elsewhere.
  subGatherShape.append(subSliceSizes.begin(), subSliceSizes.end());                                                                                                       
                                                                                                                                                                           
  // Batch axes: indicesShape minus indexVectorDim.
  for (int64_t i = 0; i < static_cast<int64_t>(indicesShape.size()); ++i) {                                                                                                
    if (i != indexVectorDim) {                                                                                                                                             
      subGatherShape.push_back(indicesShape[i]);
    }                                                                                                                                                                      
  }
```
One pass, no sort/erase, no reliance on partialDimOutputAxes, and mirrors the flatOperandShape construction (leading/middle/trailing segments).

sdjukicTT (commented):
Thank you for adding this!

Two notes:
1. Could we emit a single embedding instead of prod(slice_sizes) of them? StableHLOGatherToEmbeddingPattern already does this for the single-partial-dim case via expandStartIndices - it expands the index tensor by slice_size times and runs one embedding, rather than emitting one embedding per window offset.                                                      
The multi-partial case is the same idea, just 2-D: build a constant window-offset tensor of shape [Hw, Ww] holding dhxW + dw, broadcast-add it to the flat indices to get expanded indices of shape [..., Hw, Ww], run one embedding on the flat [HxW, C] operand, then reshape/transpose to place the window axes.                                        
  This would avoid the prod(slice_sizes) sub-gather dispatches and the whole concat tree, reducing the number of ops. With example of 3x3 slice sizes, it's fine, but if it was a larger window size, this wouldn't scale well.

2. More broadly, the multi-partial case is a generalization of what StableHLOGatherToEmbeddingPattern already does, so most of the machinery could be shared rather than duplicated.
The existing pattern has helpers for exactly the pieces this pattern reimplements -permuteInput / reshapeInput for flattening the indexed dims of the operand, flattenStartIndices for combining multi-component indices into a 1-D index, expandStartIndices for broadening indices to cover a window, and reshapeAndPermuteOutput for putting output axes back in place.                                                                                                                                                       
Refactoring those helpers out (or letting this new pattern call into them) and making them reusable would keep the two paths consistent - same flattening convention, same index arithmetic, same output reshape - and shrink this pattern to mostly the multi-dim window-expansion logic.

lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
```diff
@@ -5620,6 +5710,276 @@ class StableHLOGatherToSliceRepeatConcatPattern
...
+    auto resultType = mlir::cast<RankedTensorType>(srcOp.getType());
+    int64_t outputRank = resultType.getRank();
+    int64_t operandRank = operandType.getRank();
+    int64_t numWindow = static_cast<int64_t>(windowSizes.size());
+    int64_t numBatch = static_cast<int64_t>(indicesShape.size()) - 1;
+
+    SmallVector<int64_t> permutation(outputRank);
+    for (int64_t t = 0; t < outputRank; ++t) {
+      if (t < operandRank) {
+        auto it = llvm::find(partialDims, t);
```
vmilosevic: :warning: **llvm\-qualified\-auto** :warning:
`` auto it `` can be declared as `` auto *it ``
```suggestion
        auto *it = llvm::find(partialDims, t);
```

vmilosevic (commented):

lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
```diff
@@ -5107,9 +5108,98 @@ class StableHLOGatherToEmbeddingPattern
...
+
+    SmallVector<int64_t> expandedShape = {totalWindow, M};
+    auto elemType = startIndicesType.getElementType();
+    auto expandedType = RankedTensorType::get(expandedShape, elemType,
+                                              startIndicesType.getEncoding());
+
+    // Build offset matrix [totalWindow, M]: row r has all elements =
+    // windowOffsets[r]. Element type matches startIndices.
+    DenseElementsAttr offsetAttr;
+    if (mlir::isa<mlir::FloatType>(elemType)) {
```
sdjukicTT: This helper is called with indices that are converted to float, so we can delete `else` branch.

lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
```diff
@@ -5620,6 +5710,276 @@ class StableHLOGatherToSliceRepeatConcatPattern
...
+    auto operandShape = operandType.getShape();
+    auto indicesShape = indicesType.getShape();
+
+    Location loc = srcOp.getLoc();
+
+    // Reuse shared helpers from StableHLOGatherToEmbeddingPattern to flatten
+    // the operand's indexed dims, combine multi-component indices into a flat
+    // 1-D index, and reshape indices to 2-D.
+    auto input = mlir::cast<mlir::TypedValue<mlir::RankedTensorType>>(
+        srcOp.getOperand());
```
sdjukicTT: Please source operand SSA values from adaptor (e.g. adaptor.getOperands()[0]) when building new ops.

lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
```diff
@@ -5620,6 +5710,276 @@ class StableHLOGatherToSliceRepeatConcatPattern
...
+        mlir::cast<RankedTensorType>(expandedIndices.getResult().getType());
+    SmallVector<int64_t> embeddingOutShape(
+        expandedIndicesType.getShape().begin(),
+        expandedIndicesType.getShape().end());
+    embeddingOutShape.push_back(reshapedInput.getType().getShape()[1]);
+    auto embeddingOutType = RankedTensorType::get(
+        embeddingOutShape, reshapedInput.getType().getElementType(),
+        reshapedInput.getType().getEncoding());
+    auto embeddingOp = rewriter.create<ttir::EmbeddingOp>(
+        loc, embeddingOutType, expandedIndices.getResult(), reshapedInput);
```
sdjukicTT: Minor inconsistency: appendLocationSuffix(loc, "_embedding")

lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
```diff
@@ -5620,6 +5710,276 @@ class StableHLOGatherToSliceRepeatConcatPattern
...
+    SmallVector<int64_t> windowSizes;
+    windowSizes.reserve(partialDims.size());
+    for (int64_t pd : partialDims) {
+      windowSizes.push_back(sliceSizes[pd]);
+    }
+    SmallVector<int64_t> partialDimSizes;
+    partialDimSizes.reserve(partialDims.size());
+    for (int64_t pd : partialDims) {
+      partialDimSizes.push_back(operandShape[pd]);
+    }
```
sdjukicTT, line 5871: Same loop as above, can be done once

lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
```diff
@@ -5620,6 +5710,276 @@ class StableHLOGatherToSliceRepeatConcatPattern
...
+            if (!llvm::is_contained(partialDims, j)) {
+              ++nonIndexedIdx;
+            }
+          }
+          permutation[t] = numWindow + numBatch + nonIndexedIdx;
+        }
+      } else {
+        permutation[t] = numWindow + (t - operandRank);
+      }
+    }
```
sdjukicTT, line 5956: This permutation building can be less complicated, something like:                                                                                                                                                                                                                                                                                            
```                                                                                                                                                                                                  
  SmallVector<int64_t> permutation(outputRank);                                                                                                                                                                                                                                                                    
  int64_t partialBegin = partialDims.front();                                                                                                                                                                                                                                                                      
                             
  // Window block (intermediate axes [0, numWindow)) lands on result positions                                                                                                                                                                                                                                     
  // [partialBegin, partialBegin + numWindow) — the partial-indexed operand axes.
  for (int64_t j = 0; j < numWindow; ++j) {                                                                                                                                                                                                                                                                        
    permutation[partialBegin + j] = j;                                                                                                                                                                                                                                                                             
  }                                                                                                                                                                                                                                                                                                                
  // Batch block (intermediate axes [numWindow, numWindow + numBatch)) lands on                                                                                                                                                                                                                                    
  // result positions [operandRank, outputRank).
  for (int64_t b = 0; b < numBatch; ++b) {                                                                                                                                                                                                                                                                         
    permutation[operandRank + b] = numWindow + b;
  }                                                                                                                                                                                                                                                                                                                
  // Non-indexed-operand block lands on the remaining result positions in operand
  // order: axes [0, partialBegin) and [partialBegin + numWindow, operandRank).                                                                                                                                                                                                                                    
  int64_t intIdx = numWindow + numBatch;                                                                                                                                                                                                                                                                           
  for (int64_t i = 0; i < partialBegin; ++i) {                                                                                                                                                                                                                                                                     
    permutation[i] = intIdx++;                                                                                                                                                                                                                                                                                     
  }                                                                                                                                                                                                                                                                                                                
  for (int64_t i = partialBegin + numWindow; i < operandRank; ++i) {
    permutation[i] = intIdx++;                                                                                                                                                                                                                                                                                     
  } 
```

sdjukicTT (approved):
Looks great!
One more request: since there is already `stablehlo.gather` support in StableHLO Builder, can you add a test (parametrization) for this case?

kamalrajkannan78:
> Looks great! One more request: since there is already `stablehlo.gather` support in StableHLO Builder, can you add a test (parametrization) for this case?

Added now.

kamalrajkannan78:
> > Looks great! One more request: since there is already `stablehlo.gather` support in StableHLO Builder, can you add a test (parametrization) for this case?

@sdjukicTT ,

Hit two issues in the golden infrastructure when adding the parametrized test for multi-partial case: [CI run](https://github.com/tenstorrent/tt-mlir/actions/runs/25103426139/job/73584540147)

- gather_golden had an assertion collapsed_slice_dims == start_index_map, which the window-gather case (collapsed_slice_dims=[]) doesn't satisfy. Relaxed the assertion to also accept the empty-collapsed case.
- After the assertion, the permutation step was passing a source→target mapping directly to torch.permute, which expects target→source. The two existing test cases happened to be involutions (their own inverse) so the bug was latent. Our new case isn't an involution so it surfaced. Fixed by explicitly inverting before passing to permute.

Both fixes are in `tools/golden/mapping.py`.

jgrimTT (approved):

PR #8069, opened 2026-04-22
InvokeOpLib foundation

pljubisavljevicTT:
### Ticket
Link to Github Issue

### Problem description

### What's changed
This PR introduces TTNNOpInvokeLib, a shared library that consolidates op parameter resolution and op-calling logic into one place, giving OpModel and Runtime a single call site. Op parameters inside the library are represented as NativeTable types. A NativeTable type is a plain mutable C++ struct with real fields (std::string, std::vector, std::unique_ptr<>, …) — the in-memory form of a schema table, in contrast to the Table type which is a read-only view over serialized bytes accessed through vtable offsets.               
   
  Foundation changes:                                                                                                                                                                                                                         
  - Added toNative() conversion overloads in `MLIRToFlatbuffer.h` for core types; `toFlatbuffer()` becomes a wrapper of `toNative()`                                                                                                                                                                                                                               
  - Moved shared utility functions from Runtime into `OpInvoke/utils` so both OpModel and Runtime use the same implementation                                                                                                             
  - Added helper functions in `Conversion.cpp` and `TTNNOpModel.cpp` that subsequent op PRs will build on
  
  Subsequent PRs in this stack wire individual ops (Conv2d, Matmul, Eltwise) through unified_op_lib end-to-end.


lib/OpModel/TTNN/Conversion.cpp
```diff
@@ -496,7 +560,9 @@ getConv2dConfig(const std::optional<Conv2dConfigAttr> &conv2dConfig) {
     config.shard_layout = std::nullopt;
   }
 
-  config.core_grid = std::nullopt;
+  if (conv2dConfig->getCoreGrid()) {
```
rpavlovicTT, line 569: Should this be put to subsequent PR for conv2d?

lib/OpModel/TTNN/TTNNOpModel.cpp
```diff
@@ -264,6 +250,37 @@ getNullableMemoryConfig(TTNNLayoutAttr layout) {
...
+std::optional<::tt::target::ttnn::MemoryConfigT>
+getNullableMemoryConfigT(TTNNLayoutAttr layout) {
+  if (!layout) {
+    return std::nullopt;
+  }
+  return conversion::getMemoryConfigT(layout);
+}
+
+std::unique_ptr<::tt::target::ttnn::TensorRefT>
+getOutputTensorRefT(TTNNLayoutAttr layout) {
```
rpavlovicTT, line 238: What is this function for?

runtime/include/tt/runtime/detail/ttnn/operations/utils.h
```diff
@@ -22,9 +22,6 @@ bool isTilized(const ::tt::target::ttnn::TensorRef *tensorRef);
 
 ::ttnn::DataType getDataType(const ::tt::target::ttnn::TensorRef *tensorRef);
 
-::ttnn::operations::unary::UnaryOpType
-toTTNNUnaryOpType(::tt::target::ttnn::EltwiseUnaryOpType unaryOpType);
-
```
rpavlovicTT, line 27: Should this be part of eltwise unary PR?

unified_op_lib/include/utils/utils.h
```diff
@@ -0,0 +1,106 @@
...
+#pragma clang diagnostic pop
+#include "ttnn/operations/conv/conv2d/conv2d.hpp"
+#include "ttnn/operations/matmul/device/config/matmul_program_config_types.hpp"
+
+// Macros to wrap overloaded functions for use with
+// query_op_constraints/runtime. These create a generic lambda that forwards
+// arguments, letting the compiler resolve the correct overload based on the
+// actual argument types.
+// clang-format off
+#define WRAP_OP(op)                                                         \
```
rpavlovicTT: misaligned `\`

unified_op_lib/include/utils/utils.h
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+#define QUERY_OP_CONSTRAINTS(op, device, ...)                                  \
+  ::ttnn::graph::query_op_constraints(WRAP_OP(op), device, __VA_ARGS__)
+
+#define QUERY_OP_RUNTIME(op, device, ...)                                      \
+  ::ttnn::graph::query_op_runtime(WRAP_OP(op), device, __VA_ARGS__)
+// clang-format on
+
+namespace unifiedOpLib {
```
rpavlovicTT: namespace should be `unified_op_lib` or whatever name we chose

unified_op_lib/lib/utils/utils.cpp
```diff
@@ -0,0 +1,577 @@
...
+
+  if (config.dst_full_sync_en) {
+    computeKernelConfig.dst_full_sync_en = *config.dst_full_sync_en;
+  }
+
+  return computeKernelConfig;
+}
+
+///////
+///================================================================================================
```
rpavlovicTT: weird pattern, please follow code style guidelines

unified_op_lib/lib/utils/utils.cpp
```diff
@@ -0,0 +1,577 @@
...
+  case ::tt::target::BufferType::L1:
+    return ::ttnn::BufferType::L1;
+  case ::tt::target::BufferType::SystemMemory:
+    return ::ttnn::BufferType::SYSTEM_MEMORY;
+  case ::tt::target::BufferType::L1Small:
+    return ::ttnn::BufferType::L1_SMALL;
+  case ::tt::target::BufferType::Trace:
+    return ::ttnn::BufferType::TRACE;
+  }
+};
```
rpavlovicTT: trailing semicolon

unified_op_lib/lib/utils/utils.cpp
```diff
@@ -0,0 +1,577 @@
...
+toTTNNCoreRange(const tt::target::ttnn::CoreRange &coreRange) {
+  tt::tt_metal::CoreCoord start = toTTNNCoreCoord(coreRange.start_coord());
+  tt::tt_metal::CoreCoord end = toTTNNCoreCoord(coreRange.end_coord());
+  return tt::tt_metal::CoreRange(start, end);
+}
+
+tt::tt_metal::CoreRangeSet
+toTTNNCoreRangeSet(const tt::target::ttnn::CoreRangeSetT &coreRangeSet) {
+  std::set<tt::tt_metal::CoreRange> coreRanges;
+  for (const tt::target::ttnn::CoreRange coreRange : coreRangeSet.core_ranges) {
```
rpavlovicTT: should use `for (const auto&` or `for (const tt::target::ttnn::CoreRange&` to avoid copies. Check other for loops too.

include/ttmlir/Target/Utils/MLIRToFlatbuffer.h
```diff
@@ -541,21 +594,47 @@ toFlatbuffer(FlatbufferObjectCache &cache, ElementsAttr elementsAttr) {
...
+
+inline ::tt::target::ttnn::CoreRangeSetT
+toNative(ttnn::CoreRangeSetAttr coreRangeSetAttr) {
+  if (!coreRangeSetAttr) {
+    return {};
+  }
+  ::tt::target::ttnn::CoreRangeSetT coreRangeSetT;
+
+  auto core_ranges = coreRangeSetAttr.getCoreRanges().vec();
+  for (auto core_range : core_ranges) {
```
rpavlovicTT: `const auto&`

include/ttmlir/Target/Utils/MLIRToFlatbuffer.h
```diff
@@ -541,21 +594,47 @@ toFlatbuffer(FlatbufferObjectCache &cache, ElementsAttr elementsAttr) {
...
+}
+
+inline ::tt::target::ttnn::CoreRangeSetT
+toNative(ttnn::CoreRangeSetAttr coreRangeSetAttr) {
+  if (!coreRangeSetAttr) {
+    return {};
+  }
+  ::tt::target::ttnn::CoreRangeSetT coreRangeSetT;
+
+  auto core_ranges = coreRangeSetAttr.getCoreRanges().vec();
```
rpavlovicTT: no need for copy here

include/ttmlir/Target/Utils/MLIRToFlatbuffer.h
```diff
@@ -791,33 +946,58 @@ toFlatbuffer(FlatbufferObjectCache &cache,
...
+  ::tt::target::ttnn::Conv2dConfigT conv2dConfigT;
+
+  conv2dConfigT.weights_dtype = toNative(config.getWeightsDtype());
   if (config.getActivation()) {
-    activation = toFlatbuffer(cache, config.getActivation());
+    conv2dConfigT.activation =
+        std::make_unique<::tt::target::ttnn::UnaryWithParamT>(
+            toNative(config.getActivation()));
+  } else {
+    conv2dConfigT.activation = nullptr;
```
rpavlovicTT: This is redundant most likely.

unified_op_lib/include/utils/utils.h
```diff
@@ -0,0 +1,106 @@
...
+tt::tt_metal::CoreRange
+toTTNNCoreRange(const tt::target::ttnn::CoreRange &coreRange);
+
+tt::tt_metal::CoreRangeSet
+toTTNNCoreRangeSet(const tt::target::ttnn::CoreRangeSetT &coreRangeSet);
+
+::ttnn::operations::unary::UnaryOpType
+toTTNNUnaryOpType(::tt::target::ttnn::UnaryOpType unaryOpType);
+
+::ttnn::DataType toTTNNDataType(::tt::target::DataType dataType);
```
rpavlovicTT: Duplicate of same function on line 44

rpavlovicTT (commented):
I'd explain what native type is in the PR description. 
Few general notes:
1. UnifiedOpLib is non-informative. Suggested replacements: TTNNOpDispatch. 
2. We need some tests in this PR too. Something like we did at the start with 3-way consistency tests. Assert `toNative(X)` is structurally equal to `UnPack(serialize(toFlatbuffer(X)))` (round-trip)
3. Op specific utils should be deferred to op specific PRs.
4. Can we complete `MlirToFlatbuffer` changes for all the types? Like use `toNative` in each `toFlatbuffer`. Does this break something? I am just asking because I'd like to avoid having partially refactored that translation unit, but if it is necessary it is okay.

lib/OpModel/TTNN/Conversion.cpp
pljubisavljevicTT: No, because assert above is removed and issue is closed.

lib/OpModel/TTNN/TTNNOpModel.cpp
pljubisavljevicTT: Output tensor in NativeTable type which is passed to unified_op_lib will be created with this function.

runtime/lib/ttnn/operations/conv/conv3d.cpp
sgholamiTT, line 55: I fail to realize this code.
The local is still `::ttnn::experimental::prim::Conv3dConfig config;`, but only the first assignment was rewritten and now references a non‑existent variable.
Should we have both of the variables:
```c++
std::optional<::ttnn::experimental::prim::Conv3dConfig> conv3dConfig;
::ttnn::experimental::prim::Conv3dConfig config;
```
Or maybe you intended to replace all usages and forgot ?

runtime/lib/ttnn/utils/utils.cpp
```diff
@@ -387,10 +286,15 @@ fromTTNNCoreRange(const tt::tt_metal::CoreRange &coreRange) {
...
 toTTNNCoreRangeSet(const tt::target::ttnn::CoreRangeSet &coreRangeSet) {
   std::set<tt::tt_metal::CoreRange> coreRanges;
+
   for (const tt::target::ttnn::CoreRange *coreRange :
        *coreRangeSet.core_ranges()) {
-    coreRanges.emplace(toTTNNCoreRange(*coreRange));
+    coreRanges.emplace(
+        unified_op_lib::operations::utils::toTTNNCoreRange(*coreRange));
+    coreRanges.emplace(
+        unified_op_lib::operations::utils::toTTNNCoreRange(*coreRange));
```
sgholamiTT: Duplicate ?

include/ttmlir/Target/Utils/MLIRToFlatbuffer.h
```diff
@@ -829,34 +1010,73 @@ toFlatbuffer(FlatbufferObjectCache &cache, ttnn::Conv2dSliceType sliceType) {
...
+  if (config.getHOutBlock()) {
+    conv3dConfigT.h_out_block = *config.getHOutBlock();
+  }
+  if (config.getCOutBlock()) {
+    conv3dConfigT.c_out_block = *config.getCOutBlock();
+  }
+  if (config.getCInBlock()) {
+    conv3dConfigT.c_in_block = *config.getCInBlock();
+  }
+  // compute_with_storage_grid_size not set here
```
sgholamiTT, line 1075: Hmm this can become a real bug which could become difficult to trace.
2 asks:
1. Is there a reason we cannot set `compute_with_storage_grid_size` here ? (I realize that the original `toFlatbuffer(FlatbufferObjectCache &cache, ttnn::Conv3dConfigAttr config)` was also silently ignoring this attr).
2. If it's not possible to fix it, please create an issue with a `TODO` so we can track this.

runtime/lib/ttnn/operations/utils/utils.cpp
```diff
@@ -22,8 +22,9 @@ bool isTilized(const ::tt::target::ttnn::TensorRef *tensorRef) {
 }
 
 ::ttnn::DataType getDataType(const ::tt::target::ttnn::TensorRef *tensorRef) {
-  return ::tt::runtime::ttnn::utils::toTTNNDataType(
-      tensorRef->desc()->layout()->memory_desc()->data_type());
+  target::ttnn::TensorRefT tensorRefT;
+  tensorRef->UnPackTo(&tensorRefT);
+  return unified_op_lib::operations::utils::getDataType(tensorRefT);
```
sgholamiTT: Hmm, these functions are called on hot paths in op runs.
`view->UnPackTo(&native)` walks the entire view recursively and deep‑copies every field into freshly heap‑allocated native objects. This is very expensive compared to the existing `O(1)` implementation.
Can we find an alternative for this, and if not, I think we might wanna measure compile time for some big graphs where this might have some actual effects.

Correct me if I'm wrong, but what if we don't wrap at all, and just call the enum converter directly. For these two functions, the only thing that actually had to move into `unified_op_lib` was the `target::DataType → ttnn::DataType` mapping (`toTTNNDataType`). The runtime side can keep its existing helper:
```c++
::ttnn::DataType getDataType(const ::tt::target::ttnn::TensorRef *tensorRef) {
  return unified_op_lib::operations::utils::toTTNNDataType(
      tensorRef->desc()->layout()->memory_desc()->data_type());
}
```
That preserves the original codegen while still routing through the unified enum table — which is the only thing you actually wanted to dedupe here.

runtime/lib/ttnn/utils/utils.cpp
```diff
@@ -47,9 +50,9 @@ bool isOnDevice(const ::ttnn::StorageType &storageType) {
 }
 
 bool inSystemMemory(const ::tt::target::ttnn::TensorRef *tensorRef) {
-  const ::tt::target::ttnn::StorageType storageType =
-      tensorRef->desc()->layout()->memory_desc()->storage_type();
-  return storageType == ::tt::target::ttnn::StorageType::Host;
+  ::tt::target::ttnn::TensorRefT tensorRefT;
+  tensorRef->UnPackTo(&tensorRefT);
+  return unified_op_lib::operations::utils::inSystemMemory(tensorRefT);
```
sgholamiTT: Same applies here.

include/ttmlir/Target/Utils/MLIRToFlatbuffer.h
```diff
@@ -379,6 +429,15 @@ toFlatbuffer(FlatbufferObjectCache &cache, const std::optional<T> &optValue) {
...
 }
 
+template <typename T>
+::flatbuffers::Optional<decltype(toNative(std::declval<T>()))>
+toNative(const std::optional<T> &opt) {
+  if (!opt) {
+    return ::flatbuffers::nullopt;
+  }
+  return toNative(*opt);
+}
```
sgholamiTT, line 454: Not sure why `toNative(std::optional<T>)` returns `flatbuffers::Optional<…>`, not `std::optional<…>`. Given the PR description ("plain mutable C++ struct with real fields … std::optional"), this is surprising for me. 
For types whose `toNative` already returns a `flatbuffers::Optional` (e.g. `BoolAttr`), wrapping in `std::optional` then calling this overload nests two Optionals, which is at best confusing.

runtime/lib/ttnn/operations/CMakeLists.txt
```diff
@@ -134,7 +135,7 @@ target_include_directories(TTRuntimeTTNNOps PUBLIC
 )
 
 target_include_directories(TTRuntimeTTNNOps SYSTEM PUBLIC "$<BUILD_INTERFACE:${TTMETAL_INCLUDE_DIRS}>")
-target_link_libraries(TTRuntimeTTNNOps PUBLIC TTNN_LIBRARY TTRuntimeTTNNDebug TTRuntimeTTNNTypes TTRuntimeTTNNUtils TRACY_LIBRARY)
+target_link_libraries(TTRuntimeTTNNOps PUBLIC TTNN_LIBRARY TTRuntimeTTNNDebug TTRuntimeTTNNTypes TTRuntimeTTNNUtils TRACY_LIBRARY TTNNOpDispatch)
```
sgholamiTT: Directory is `unified_op_lib/`, namespace is `unified_op_lib::operations::utils`, but the CMake target was renamed to `TTNNOpDispatch`. Maybe pick one ?

lib/OpModel/TTNN/TTNNOpModel.cpp
```diff
@@ -3,8 +3,10 @@
 // SPDX-License-Identifier: Apache-2.0
 
 #include "ttmlir/OpModel/TTNN/TTNNOpModel.h"
+#include "ttmlir/Target/Utils/MLIRToFlatbuffer.h"
 #include "ttmlir/Utils.h"
 #include "llvm/ADT/SmallVector.h"
+#include <tuple>
```
sgholamiTT: wrong include order ?

lib/OpModel/TTNN/TTNNOpModel.cpp
```diff
@@ -14,6 +16,7 @@
 #include "ttmlir/Dialect/TTNN/IR/TTNNOpsAttrs.h"
 #include "ttmlir/OpModel/TTNN/Conversion.h"
 #include "ttmlir/OpModel/TTNN/SingletonDeviceContext.h"
+#include <utils/utils.h>
```
sgholamiTT: Include absolute path.

lib/OpModel/TTNN/Conversion.cpp
```diff
@@ -248,6 +249,21 @@ ::tt::tt_metal::CoreRangeSet getCoreRangeSet(const TTNNLayoutAttr &layout) {
   return ::tt::tt_metal::CoreRangeSet(coreRangeSet);
 }
 
+::tt::target::ttnn::CoreRangeSetT
+getCoreRangeSetT(const TTNNLayoutAttr &layout) {
+  ::tt::target::ttnn::CoreRangeSetT coreRangeSetT;
+  assert(layout.getGrid().getVirtToPhysicalMap().isEmpty() == false);
```
sgholamiTT: Nit: Prefer `assert(!layout.getGrid().getVirtToPhysicalMap().isEmpty())`.

sgholamiTT (commented):
Thanks for doing this. Overall LGTM 👍🏻 
Few comments inlined.

sgholamiTT (approved):

nsumrakTT:
Huh it seems you have a merge problem, please rebase your commits on main

CMakeLists.txt
```diff
@@ -208,6 +208,9 @@ add_subdirectory(test)
 endif()
 add_subdirectory(tools)
 add_subdirectory(runtime)
+if (TTMLIR_ENABLE_RUNTIME OR TTMLIR_ENABLE_OPMODEL)
+  add_subdirectory(unified_op_lib)
```
rpavlovicTT: Rename

runtime/test/ttnn/utils.cpp
```diff
@@ -40,7 +40,7 @@ Layout getDramInterleavedRowMajorLayout(::tt::target::DataType dataType) {
 ::tt::runtime::Layout getHostRowMajorLayout(::tt::target::DataType dataType) {
   LOG_ASSERT(getCurrentDeviceRuntime() == DeviceRuntime::TTNN);
   ::ttnn::DataType ttnnDataType =
-      ::tt::runtime::ttnn::utils::toTTNNDataType(dataType);
+      unified_op_lib::operations::utils::toTTNNDataType(dataType);
```
rpavlovicTT: Still old namespace

lib/OpModel/TTNN/CMakeLists.txt
```diff
@@ -19,9 +19,9 @@ if (TTMLIR_ENABLE_OPMODEL)
     endif()
 
     # Link to tt-metal libs and include directories
-    target_include_directories(${LIB_NAME} SYSTEM PUBLIC "$<BUILD_INTERFACE:${TTMETAL_INCLUDE_DIRS}>")
+    target_include_directories(${LIB_NAME} SYSTEM PUBLIC "$<BUILD_INTERFACE:${TTMETAL_INCLUDE_DIRS}>" PRIVATE ${PROJECT_SOURCE_DIR}/unified_op_lib/include)
```
rpavlovicTT: Replace after you rename path

lib/OpModel/TTNN/Conversion.cpp
```diff
@@ -276,6 +285,33 @@ getShardSpec(const TTNNLayoutAttr &layout) {
                                    ::tt::tt_metal::ShardOrientation::ROW_MAJOR);
 }
 
+std::optional<::tt::target::ttnn::ShardSpecT>
+getShardSpecT(const TTNNLayoutAttr &layout) {
```
rpavlovicTT, line 291: Can we remove old implementation that returns `ShardSpec` ? Or will you do it in subsequent PR?

runtime/lib/ttnn/operations/utils/utils.cpp
pljubisavljevicTT: These wrapper functions only exist while we incrementally migrate ops into TTNNOpInvoke. Once all ops are routed through the unified lib, the runtime-side helpers (getDataType, inSystemMemory, …) go away entirely - call sites will hit the unified path directly without any UnPackTo operations.
We already run end-to-end performance benchmarks with these changes and the perf is unchanged.

include/ttmlir/Target/Utils/MLIRToFlatbuffer.h
pljubisavljevicTT: NativeTable types use flatbuffers::Optional<T> for their optional fields, so toNative returns the same type to allow direct assignment.

lib/OpModel/TTNN/Conversion.cpp
pljubisavljevicTT: Some ops are still using this function.

rpavlovicTT (commented):
Looking good overall. One of the last asks is to move sources to lib/OpInvoke/TTNN/ (parallel to lib/OpModel/TTNN/) and public headers to include/ttmlir/OpInvoke/TTNN/ (parallel to include/ttmlir/OpModel/TTNN/). CMake target TTNNInvokeOpLib. Runtime can include via #include "ttmlir/OpInvoke/TTNN/utils.h" — same pattern as it already uses for ttmlir/Target/....

rpavlovicTT (approved):

include/ttmlir/Target/Utils/MLIRToFlatbuffer.h
```diff
@@ -388,6 +437,22 @@ toFlatbuffer(FlatbufferObjectCache &cache, const std::optional<T> &optValue) {
...
 
+template <typename>
+struct is_fb_optional : std::false_type {};
+template <typename T>
+struct is_fb_optional<::flatbuffers::Optional<T>> : std::true_type {};
+
+template <typename T>
+auto toNative(const std::optional<T> &opt) {
+  using R = decltype(toNative(std::declval<T>()));
+  using Result = std::conditional_t<is_fb_optional<R>::value, R,
```
azecevic000, line 448: Does it ever happen that `toNative` of non-optional type `T` returns `flatbuffers::Optional<R>`? Tbh, I would consider that weird behaviour.

include/ttmlir/Target/Utils/MLIRToFlatbuffer.h
```diff
@@ -679,86 +768,159 @@ toFlatbuffer(FlatbufferObjectCache &, ttnn::UnaryOpType unaryOpType) {
...
 }
 
+inline ::tt::target::ttnn::UnaryOpType
+toFlatbuffer(FlatbufferObjectCache &, ttnn::UnaryOpType unaryOpType) {
+  return toNative(unaryOpType);
+}
+
+inline ::tt::target::ttnn::MatmulMultiCoreReuseProgramConfigT
+toNative(ttnn::MatmulMultiCoreReuseProgramConfigAttr matmulConfigAttr) {
+  ::tt::target::ttnn::MatmulMultiCoreReuseProgramConfigT matmulConfigT;
```
azecevic000: `matmulConfigT` can be a very misleading variable name. Trailing `T` or `_t` is by convention used exclusively for type names (to signal that it's a type), we should keep it that way.

lib/OpInvoke/TTNN/utils/utils.cpp
```diff
@@ -0,0 +1,250 @@
...
+  const auto memLayout = toTTNNTensorMemoryLayout(targetMemLayout);
+
+  // Verify that shard spec is present only for sharded memory layouts
+  if (callType == CallType::EXECUTE) {
+    const bool hasShardSpec =
+        (memcfg.shard_spec != nullptr) || (memcfg.nd_shard_spec != nullptr);
+    LOG_ASSERT(
+        hasShardSpec == isSharded(targetMemLayout),
+        "A shard spec must be present if and only if the tensor is sharded");
+  }
```
rpavlovicTT: Honestly I think we can drop this if and assertion inside. It's the only reason we are passing CallType to InvokeLib resolveOpParams as argument, right?

rpavlovicTT:
Update PR title to "InvokeOpLib foundation"

include/ttmlir/Target/Utils/MLIRToFlatbuffer.h
pljubisavljevicTT: `toNative(BoolAttr)` is an example.

lib/OpModel/TTNN/TTNNOpModel.cpp
```diff
@@ -241,6 +226,37 @@ getNullableMemoryConfig(TTNNLayoutAttr layout) {
...
+getNullableMemoryConfigT(TTNNLayoutAttr layout) {
+  if (!layout) {
+    return std::nullopt;
+  }
+  return conversion::getMemoryConfigT(layout);
+}
+
+std::unique_ptr<::tt::target::ttnn::TensorRefT>
+getOutputTensorRefT(TTNNLayoutAttr layout) {
+  auto memoryConfigT = getNullableMemoryConfigT(layout);
```
rpavlovicTT: Rename this var as @azecevicTT suggested before. Avoid trailing `T` on var names. You can call it memoryConfigNative or something similar.

rpavlovicTT (approved):

PR #8073, opened 2026-04-22
Fix GIL segfault in debug hook Python callbacks

ndrakulicTT:
The runtime (in tt-xla) executes on a C++ thread that has released the Python GIL. The debug hook callbacks (`pre_op`, `post_op`, `pre_program`, `post_program`) were invoking Python callables without re-acquiring the GIL, causing a segfault whenever a Python function was registered as a hook.

**Fixes:**
- Add `nb::gil_scoped_acquire` in each lambda before invoking the Python callable

### Related

Part of the problem was addressed in #7891 (commit `52a6149`), which changed `runCallback` arguments from pass-by-copy to `const` reference, eliminating unnecessary copies of the callable on each invocation. This PR builds on that by fixing the remaining root cause - the missing GIL reacquisition.

### Testing

No standalone test could be added within tt-mlir - the crash requires a live device execution path that invokes the runtime from a GIL-released thread. The fix is validated in tt-xla via:
```
import torch
import torch_xla
import ttrt


def preop(a,b,c):
    print("simple test")


a = torch.ones((128, 128)).to("xla")
b = torch.rand((128,128)).to("xla")

hooks = ttrt.runtime.DebugHooks.get(pre_op=preop)
print(hooks)

a = torch.ones((128, 128)).to("xla")
b = torch.rand((128,128)).to("xla")
print(a * 2 + b)
torch_xla.sync(wait=True)

ttrt.runtime.unregister_hooks()
```
without this fix, it segfaults.


acolicTT (approved):

jameszianxuTT (approved):
Nice find, is it possible to add a testcase for this in tt-xla ~or tt-mlir?~

ndrakulicTT:
@jameszianxuTT, not yet I have PR to enable this working on local machines, will be taking look in the near future how to test this in `tt-xla`

PR #8078, opened 2026-04-22
chisel hotfix to pass tests on main

dlokeTT:
### Problem description
Chisel tests are failing on main

### What's changed
Added params for debug printing so the strings match

### Checklist


ndrakulicTT (approved):
Yeah this was my oversight, thanks for the fix.

PR #8082, opened 2026-04-22
Add slice_static workaround when circular buffer > L1 size

ddilbazTT:
### Ticket
https://github.com/tenstorrent/tt-metal/issues/42882

### Problem description
One of the customer models fails during the following slice static operation:
```
%23 = "ttnn.slice_static"(%22) <{begins = [0 : i32, 2 : i32, 0 : i32], ends = [6 : i32, 3 : i32, 258112 : i32], step = [1 : i32, 1 : i32, 1 : i32]}> : (tensor<6x3x258112xf32, #ttnn_layout1>) -> tensor<6x1x258112xf32, #ttnn_layout1> loc(#loc21)
```

The reason for failure is that circular buffer size needs to be smaller than L1 size. The following equality is not met:
`output_shape[-1] * input.element_size (= # of bytes) * 2 (double-buffering) < L1 size` This comes from tt-metal:

```
    const ttnn::Shape& output_shape = output.padded_shape();
    const uint32_t unpadded_row_size_bytes = output_shape[-1] * input.element_size();
    const uint32_t cb_page_size = tt::round_up(unpadded_row_size_bytes, alignment);
```

In broader terms, the last dimension of the tensor being sliced determines the circular buffer page sizes. So if last dimension is little, there will be smaller chunks. If the last dimension is bigger, there are bigger chunks. In this failing case, the chunks are too big to write into L1. 

### What's changed
Added slice_static workaround which permutes the last dimension with one that wouldn't exceed L1 size, operates the slice on the permuted tensor, then permutes back to original orientation. 

By permuting, the size of the data being stored in total does not change. However, the chunk size decreases whilst the number of chunks increases. 

### Checklist

### DIFF
--- a/include/ttmlir/Dialect/TTNN/Transforms/Workarounds/Decomposition/SliceStaticOpRewritePattern.h
+++ b/include/ttmlir/Dialect/TTNN/Transforms/Workarounds/Decomposition/SliceStaticOpRewritePattern.h
@@ -1,4 +1,4 @@
-// SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
+// SPDX-FileCopyrightText: (c) 2026 Tenstorrent AI ULC
 //
 // SPDX-License-Identifier: Apache-2.0
 
--- a/include/ttmlir/Dialect/TTNN/Transforms/Workarounds/Decomposition/SliceStaticOpRewritePattern.h
+++ b/include/ttmlir/Dialect/TTNN/Transforms/Workarounds/Decomposition/SliceStaticOpRewritePattern.h
@@ -14,7 +14,9 @@ namespace mlir::tt::ttnn::workarounds::decomposition {
 
 // tt-metal's slice op fails when the circular buffer for the last output
 // dimension exceeds L1 memory (check: output_shape[-1] * elem_size * 2 >
-// l1_size). This pattern decomposes such a slice into permute -> slice ->
+// l1_size). A factor of 2 is used because of double-buffering: the reader and
+// writer work simultaneously on the same CB.
+// This pattern decomposes such a slice into permute -> slice ->
 // permute, swapping the last dimension with a smaller-output dimension so
 // the CB fits in L1.
 class SliceStaticOpRewritePattern
--- a/include/ttmlir/Dialect/TTNN/Transforms/Workarounds/Decomposition/SliceStaticOpRewritePattern.h
+++ b/include/ttmlir/Dialect/TTNN/Transforms/Workarounds/Decomposition/SliceStaticOpRewritePattern.h
@@ -1,0 +1,31 @@
+// SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#ifndef TTMLIR_DIALECT_TTNN_TRANSFORMS_WORKAROUNDS_DECOMPOSITION_SLICESTATICOPREWRITEPATTERN_H
+#define TTMLIR_DIALECT_TTNN_TRANSFORMS_WORKAROUNDS_DECOMPOSITION_SLICESTATICOPREWRITEPATTERN_H
+
+#include "ttmlir/Dialect/TTNN/IR/TTNNOps.h"
+
+#include "mlir/IR/PatternMatch.h"
+#include "mlir/Support/LogicalResult.h"
+
+namespace mlir::tt::ttnn::workarounds::decomposition {
+
+// tt-metal's slice op fails when the circular buffer for the last output
+// dimension exceeds L1 memory (check: output_shape[-1] * elem_size * 2 >
+// l1_size). This pattern decomposes such a slice into permute -> slice ->
+// permute, swapping the last dimension with a smaller-output dimension so
+// the CB fits in L1.
+class SliceStaticOpRewritePattern
+    : public OpRewritePattern<ttnn::SliceStaticOp> {
+public:
+  using OpRewritePattern<ttnn::SliceStaticOp>::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(ttnn::SliceStaticOp srcOp,
+                                PatternRewriter &rewriter) const override;
+};
+
+} // namespace mlir::tt::ttnn::workarounds::decomposition
+
+#endif // TTMLIR_DIALECT_TTNN_TRANSFORMS_WORKAROUNDS_DECOMPOSITION_SLICESTATICOPREWRITEPATTERN_H
--- a/lib/Dialect/TTNN/Transforms/Workarounds/Decomposition/SliceStaticOpRewritePattern.cpp
+++ b/lib/Dialect/TTNN/Transforms/Workarounds/Decomposition/SliceStaticOpRewritePattern.cpp
@@ -1,4 +1,4 @@
-// SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
+// SPDX-FileCopyrightText: (c) 2026 Tenstorrent AI ULC
 //
 // SPDX-License-Identifier: Apache-2.0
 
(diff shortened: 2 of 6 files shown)


include/ttmlir/Dialect/TTNN/Transforms/Workarounds/Decomposition/SliceStaticOpRewritePattern.h
```diff
@@ -0,0 +1,31 @@
+// SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
```
vwellsTT: nit: year

lib/Dialect/TTNN/Transforms/Workarounds/Decomposition/SliceStaticOpRewritePattern.cpp
```diff
@@ -0,0 +1,138 @@
+// SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
```
vwellsTT: nit: year

include/ttmlir/Dialect/TTNN/Transforms/Workarounds/Decomposition/SliceStaticOpRewritePattern.h
```diff
@@ -0,0 +1,31 @@
...
+#define TTMLIR_DIALECT_TTNN_TRANSFORMS_WORKAROUNDS_DECOMPOSITION_SLICESTATICOPREWRITEPATTERN_H
+
+#include "ttmlir/Dialect/TTNN/IR/TTNNOps.h"
+
+#include "mlir/IR/PatternMatch.h"
+#include "mlir/Support/LogicalResult.h"
+
+namespace mlir::tt::ttnn::workarounds::decomposition {
+
+// tt-metal's slice op fails when the circular buffer for the last output
```
vwellsTT, line 15: I feel like this wording is maybe a bit confusing, or maybe I'm not getting it, but how does swapping dims prevent OOM? It seems that having too large of a CB in general would cause OOM, regardless of which dim it's on?

vwellsTT (commented):
It would probably be ideal to add a runtime builder test for this as well if possible

include/ttmlir/Dialect/TTNN/Transforms/Workarounds/Decomposition/SliceStaticOpRewritePattern.h
ddilbazTT: If I understand correctly, the last dimension controls how the data is being stored. Even thought the total data size is the same, it is stored in different size chunks. So when last dim is large, it means there are less chunks that are bigger. When the last dim is smaller, there are more chunks of smaller size. When the chunk size is too big, it exceeds L1 memory.

ddilbazTT: By swapping the last dim with one that is smaller, we are reducing the chunk size but increasing the number of chunks to stay within the L1 limit.

ddilbazTT: The circular buffer page size is coming from tt-metal file: https://github.com/tenstorrent/tt-metal/blob/main/ttnn/cpp/ttnn/operations/data_movement/slice/device/slice_program_factory_rm.cpp
```
    const ttnn::Shape& output_shape = output.padded_shape();
    const uint32_t unpadded_row_size_bytes = output_shape[-1] * input.element_size();
    const uint32_t cb_page_size = tt::round_up(unpadded_row_size_bytes, alignment);
```
In the failing case, I printed out some of these values:
```
unpadded_row_size_bytes: 1032448
output_shape[-1]: 258112
input.element_size(): 4
```

vwellsTT (approved):
Okay, I still think RT test would be ideal to build confidence but won't block

include/ttmlir/Dialect/TTNN/Transforms/Workarounds/Decomposition/SliceStaticOpRewritePattern.h
```diff
@@ -0,0 +1,31 @@
...
+// dimension exceeds L1 memory (check: output_shape[-1] * elem_size * 2 >
```
azecevic000, line 16: Where does the factor 2 come from? It should be documented.

lib/Dialect/TTNN/Transforms/Workarounds/Decomposition/SliceStaticOpRewritePattern.cpp
```diff
@@ -0,0 +1,138 @@
...
+
+namespace mlir::tt::ttnn::workarounds::decomposition {
+
+LogicalResult
+SliceStaticOpRewritePattern::matchAndRewrite(ttnn::SliceStaticOp srcOp,
+                                             PatternRewriter &rewriter) const {
+
+  RankedTensorType outputType = srcOp.getResult().getType();
+  RankedTensorType inputType =
+      mlir::cast<RankedTensorType>(srcOp.getInput().getType());
```
azecevic000: Nit: redundant cast

azecevicTT (approved):
The general condition for decomposition looks fine, good find. I left one comment inline about the factor of 2, it should be properly documented.

include/ttmlir/Dialect/TTNN/Transforms/Workarounds/Decomposition/SliceStaticOpRewritePattern.h
```diff
@@ -0,0 +1,31 @@
...
+// l1_size). This pattern decomposes such a slice into permute -> slice ->
+// permute, swapping the last dimension with a smaller-output dimension so
+// the CB fits in L1.
+class SliceStaticOpRewritePattern
```
sdjordjevicTT, line 22: Can we please create an issue on tt-metal side for this, so we can remove the workaround when the issue get resolved?

test/ttmlir/Dialect/TTNN/Transforms/Workarounds/slice_static_workaround.mlir
```diff
@@ -0,0 +1,49 @@
+// RUN: ttmlir-opt --ttcore-register-device --ttnn-workaround -o %t %s
```
sdjordjevicTT, line 1: Please add one regression test that it expects to fail once the workaround isn't present here:
https://github.com/tenstorrent/tt-mlir/blob/main/test/python/golden/ttir_ops/workarounds/decomposition/test_decomposition_workarounds.py

include/ttmlir/Dialect/TTNN/Transforms/Workarounds/Decomposition/SliceStaticOpRewritePattern.h
ddilbazTT: It is for double buffering - but I'll make sure to explain that better in the code comments.

include/ttmlir/Dialect/TTNN/Transforms/Workarounds/Decomposition/SliceStaticOpRewritePattern.h
ddilbazTT: https://github.com/tenstorrent/tt-metal/issues/42882 Currently I have this one open!

test/ttmlir/Dialect/TTNN/Transforms/Workarounds/slice_static_workaround.mlir
ddilbazTT: Added

ddilbazTT:
I will be merging this today, thanks for the prompt reviews! @vwellsTT @azecevicTT @sdjordjevicTT

PR #8085, opened 2026-04-22
Add support for all torch.gather semantics via the composite op

hshahTT:
### Ticket
https://github.com/tenstorrent/tt-xla/issues/3726
https://github.com/tenstorrent/tt-xla/issues/4329

### Problem description
Currently we do not support all torch.gather semantics, especially when operand and index tensors are sharded. The main reason is that the implementation of [torch.gather lowering in OpenXLA](https://github.com/openxla/xla/blob/7b523e1e4f3d0790730cf0fe97ef37841da8f111/xla/hlo/builder/lib/slicing.cc#L211-L220) doesn't use batching dims, and instead relies on a [workaround](https://github.com/openxla/stablehlo/blob/a93085ced0919a8b0071f54b431a8ed8c9c12c75/rfcs/20240311-gather-scatter-batching-dims.md?plain=1#L34-L45) to represent batching dims that Shardy is not aware of.

### What's changed
- Added support for torch.gather on sharded input and index tensors. Note that the tensors must be sharded the same way, and that they cannot be sharded on the gather dim.
- Added a new pass called `RewriteCompositeDecompFunctionsPass` that allows us to access the decomposition function of a given named composite op and rewrite it. This is used to rewrite the `stablehlo.gather` op provided by Torch-XLA into an equivalent representation that uses batching dims.
- Added workarounds to prevent tt-metal errors when ttnn.gather is called on a 1D or >4D tensor.

### Note on torch.gather -> stablehlo.gather mapping
Since torch.gather has very simple semantics compared to stablehlo.gather; given the inputs to the torch.gather call (i.e., given input tensor, index tensor, and dim) we can construct a stablehlo.gather op that uses batching dims (and is thus sharding aware):

Suppose `rank(input)` = `rank(index)` = `N` and `dim` = `D`, then the torch.gather can be lowered to:

1. Slice input if there's any dim where input.size(dim) > index.size(dim) where dim != D (torch.gather allows such a case even though it ignores such dims during the actual gather anyways)
2. reshape `index` by adding a 1-sized dim to the end (so if `index` was initially shaped `[4, 16, 512]` it is now `[4, 16, 512, 1]`
3. Create a `stablehlo.gather` op with input and reshaped index as inputs and the following as attributes:
    a. `slice_sizes`: Array of length N filled with only 1s
    b. `operand_batching_dims`: Array of all nums from 0 to N-1 except for D
    c. `start_indices_batching_dims`: Same as operand_batching_dims
    d. `collapsed_slice_dims`: Always [D]
    e. `offset_dims`: Always empty
    f. `start_index_map`: Always [D]
   g. `index_vector_dim`: Always N

This is the logic implemented by `rewriteTenstorrentGatherDecomp`

### Checklist


hshahTT:
As per offline discussion with @mrakitaTT and @acicovicTT, the approach in https://github.com/tenstorrent/tt-mlir/pull/8599 will be used instead.

PR #8094, opened 2026-04-23
Add tt-metal bisect script for uplift regression hunting

sdjordjevicTT:
### Ticket
N/A

### Summary
- Adds `tools/scripts/bisect_metal.sh` — a binary search tool that finds the exact tt-metal commit that introduced a regression between two uplift versions
- For each bisect step: checks out tt-metal, rebuilds tt-mlir, and runs the user-provided test command
- Optionally invokes Claude Code (`CLAUDE_AUTO_FIX=1`) to auto-fix intermediate build breaks caused by API changes

### Usage
```bash
# Basic
./tools/scripts/bisect_metal.sh <good_metal_commit> <bad_metal_commit> "<test_command>"

# Example: find which metal commit broke collective_permute
./tools/scripts/bisect_metal.sh 90c914ef258b cc45c083679 \
  "source env/activate && pytest -svv 'test/python/golden/test_ttir_ops.py::test_collective_permute[f32-mesh_shape1-source_target_pairs1-30x60]'"

# With auto-fix for build breaks
CLAUDE_AUTO_FIX=1 ./tools/scripts/bisect_metal.sh 90c914ef258b cc45c083679 "<test_command>"
```
### Motivation
During uplifts, tt-metal often has 200+ commits between versions. When a test regresses, manually bisecting requires rebuilding tt-mlir for each candidate, and many intermediate metal commits introduce API changes that break the tt-mlir build. This script automates the entire process and can optionally use Claude Code to patch through build breaks, reducing a multi-hour manual task to a single command.

### Checklist

### DIFF
--- a/tools/scripts/bisect_metal.sh
+++ b/tools/scripts/bisect_metal.sh
@@ -28,6 +28,8 @@
 #   BISECT_LOG        - Log file path (default: $TT_MLIR_ROOT/bisect_metal.log)
 #   BUILD_TIMEOUT     - Timeout in seconds for the build step (default: 1800 = 30min)
 #   TEST_TIMEOUT      - Timeout in seconds for the test step (default: 300 = 5min)
+#   VERBOSE           - Set to "1" to print build output to terminal (default: 0, log-only)
+#   RESET_DEVICE      - Set to "1" to run "tt-smi -r" after test failures (default: 0)
 
 set -euo pipefail
 
@@ -47,6 +49,8 @@ CLAUDE_AUTO_FIX="${CLAUDE_AUTO_FIX:-0}"
 BISECT_LOG="${BISECT_LOG:-$TT_MLIR_ROOT/bisect_metal.log}"
 BUILD_TIMEOUT="${BUILD_TIMEOUT:-1800}"
 TEST_TIMEOUT="${TEST_TIMEOUT:-300}"
+VERBOSE="${VERBOSE:-0}"
+RESET_DEVICE="${RESET_DEVICE:-0}"
 
 TT_METAL_DIR="$TT_MLIR_ROOT/third_party/tt-metal/src/tt-metal"
 THIRD_PARTY_CMAKE="$TT_MLIR_ROOT/third_party/CMakeLists.txt"
@@ -83,6 +87,8 @@ usage() {
     echo "  CLAUDE_AUTO_FIX=1  - Enable Claude Code to auto-fix build breaks"
     echo "  BUILD_TIMEOUT=1800 - Build timeout in seconds (default: 30min)"
     echo "  TEST_TIMEOUT=300   - Test timeout in seconds (default: 5min)"
+    echo "  VERBOSE=1          - Print build output to terminal (default: log-only)"
+    echo "  RESET_DEVICE=1     - Run 'tt-smi -r' after test failures (default: off)"
     exit 1
 }
 
@@ -117,7 +123,14 @@ build_tt_mlir() {
 
     # Reconfigure to pick up the new tt-metal (skip the download/update steps
     # since we already checked out manually)
-    if ! timeout "$BUILD_TIMEOUT" cmake --build "$BUILD_DIR" -- -j$(nproc) 2>&1 | tee "$build_log"; then
+    local build_ok=0
+    if [ "$VERBOSE" = "1" ]; then
+        timeout "$BUILD_TIMEOUT" cmake --build "$BUILD_DIR" -- -j$(nproc) 2>&1 | tee "$build_log" || build_ok=$?
+    else
+        timeout "$BUILD_TIMEOUT" cmake --build "$BUILD_DIR" -- -j$(nproc) > "$build_log" 2>&1 || build_ok=$?
+    fi
+
+    if [ "$build_ok" -ne 0 ]; then
         log "${RED}Build failed. See: $build_log${NC}"
         return 1
     fi
@@ -191,6 +204,12 @@ run_test() {
         return 0
     else
         log "${RED}Test FAILED${NC}"
+        if [ "$RESET_DEVICE" = "1" ]; then
+            log "${YELLOW}Resetting device with tt-smi -r...${NC}"
+            if ! tt-smi -r 2>&1 | tee -a "$BISECT_LOG"; then
+                log "${RED}tt-smi -r failed${NC}"
+            fi
+        fi
         return 1
     fi
 }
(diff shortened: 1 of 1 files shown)


tools/scripts/bisect_metal.sh
```diff
@@ -0,0 +1,315 @@
...
+fi
+git -C "$TT_METAL_DIR" fetch origin 2>/dev/null || true
+
+# Collect the list of commits between good and bad (oldest first)
+mapfile -t COMMITS < <(git -C "$TT_METAL_DIR" rev-list --reverse "$GOOD_COMMIT..$BAD_COMMIT")
+TOTAL=${#COMMITS[@]}
+
+if [ "$TOTAL" -eq 0 ]; then
+    die "No commits found between $GOOD_COMMIT and $BAD_COMMIT. Check commit order and ensure both exist in tt-metal."
+fi
```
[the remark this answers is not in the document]
sdjordjevicTT: This is a developer productivity script, not a production pipeline. Adding upfront verification means building and testing twice more (good + bad) before the bisect even starts — that could easily add 30+ minutes to an already long process.

tools/scripts/bisect_metal.sh
```diff
@@ -0,0 +1,315 @@
...
+            fi
+        fi
+    fi
+
+    if [ "$build_ok" = false ]; then
+        log "${YELLOW}SKIP: Build failed for ${commit:0:12}, treating as bad (build break).${NC}"
+        # Build break = bad, narrow right half
+        first_bad_idx=$mid
+        hi=$((mid - 1))
+
```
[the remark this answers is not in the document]
sdjordjevicTT: I'd leave this as-is. The comment is worth considering but not worth acting on for a v1 script.

tools/scripts/bisect_metal.sh
```diff
@@ -0,0 +1,315 @@
...
+if [ "$#" -lt 3 ]; then
+    usage
+fi
+
+GOOD_COMMIT="$1"
+BAD_COMMIT="$2"
+TEST_CMD="$3"
+
+# Save original state to restore later
+ORIGINAL_METAL_VERSION=$(grep -oP 'set\(TT_METAL_VERSION "\K[a-f0-9]+' "$THIRD_PARTY_CMAKE")
```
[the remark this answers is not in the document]
sdjordjevicTT: I'd fix the multi-match issue with head -1 but keep grep -oP since it's more readable than the nested sed suggestion

tools/scripts/bisect_metal.sh
```diff
@@ -0,0 +1,315 @@
...
+set -u
+BUILD_DIR="${BUILD_DIR:-$TT_MLIR_ROOT/build}"
+CLAUDE_AUTO_FIX="${CLAUDE_AUTO_FIX:-0}"
+BISECT_LOG="${BISECT_LOG:-$TT_MLIR_ROOT/bisect_metal.log}"
+BUILD_TIMEOUT="${BUILD_TIMEOUT:-1800}"
+TEST_TIMEOUT="${TEST_TIMEOUT:-300}"
+
+TT_METAL_DIR="$TT_MLIR_ROOT/third_party/tt-metal/src/tt-metal"
+THIRD_PARTY_CMAKE="$TT_MLIR_ROOT/third_party/CMakeLists.txt"
+
```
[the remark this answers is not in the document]
sdjordjevicTT: If someone is bisecting a tt-metal regression, they've already built tt-mlir at least once — meaning $TT_METAL_DIR exists.

tools/scripts/bisect_metal.sh
```diff
@@ -0,0 +1,342 @@
...
+    echo "  TEST_TIMEOUT=300   - Test timeout in seconds (default: 5min)"
+    exit 1
+}
+
+# Update TT_METAL_VERSION in third_party/CMakeLists.txt
+set_metal_version() {
+    local commit="$1"
+    log "${CYAN}Setting TT_METAL_VERSION to $commit${NC}"
+    sed -i "s|^set(TT_METAL_VERSION \"[a-f0-9]*\")|set(TT_METAL_VERSION \"$commit\")|" "$THIRD_PARTY_CMAKE"
+}
```
ddilbazTT, line 100: What happens if `set_metal_version` or `checkout_metal` fail? They don't seem to be guarded.

tools/scripts/bisect_metal.sh
```diff
@@ -0,0 +1,342 @@
...
+
+log "Found $TOTAL commits between good and bad."
+
+# --- Binary search -----------------------------------------------------------
+
+lo=0
+hi=$((TOTAL - 1))
+# Track the first known bad index (starts as the last commit)
+first_bad_idx=$hi
+
```
ddilbazTT, line 323: This script assumes hi is going to fail the test. However, that is not a necessary assumption, and is volatile to user error. I think it would be better that first_bad_idx is initialized to -1. And after the while loop, if first_bad_idx is still -1, then there is no bad commit. 
```
if [ "$first_bad_idx" -eq -1 ]; then
    log "No bad commit found — all commits passed!"
else
    log "First bad commit: ${COMMITS[$first_bad_idx]}"
fi
```

ddilbazTT: Or before doing binary search, the script could start by ensuring that hi is going to fail. If it doesn't, it could stop executing anything and report that there is no bad commit. This way, if the user entered the commit incorrectly, there wouldn't be unnecessary search.

tools/scripts/bisect_metal.sh
```diff
@@ -0,0 +1,342 @@
...
+Do NOT modify files under third_party/.
+After making changes, run: cmake --build $BUILD_DIR -- -j\$(nproc)
+PROMPT
+
+    # Run claude code with the prompt
+    # Output streams to bisect_claude.log so you can: tail -f bisect_claude.log
+    if command -v claude &>/dev/null; then
+        local fix_result=0
+        timeout 600 claude --dangerously-skip-permissions --output-format stream-json \
```
ddilbazTT: Is `--dangerously-skip-permissions` necessary?

tools/scripts/bisect_metal.sh
sdjordjevicTT: Good question. With set -e active, if they fail, the script exits immediately, and now that we've added the cleanup trap, that exit triggers restoration. So failures during the bisect loop are handled.

tools/scripts/bisect_metal.sh
sdjordjevicTT: Changed to a safer variant:
- --permission-mode auto

tools/scripts/bisect_metal.sh
sdjordjevicTT: Great tip! Applied both tips! :)

sdjordjevicTT:
Hi @ddilbazTT, thanks for a great review! I addressed your comments. Please take one more look! :)

nsmithtt (approved):

ddilbazTT (approved):

PR #8095, opened 2026-04-23
[EmitC] Fix trace ops

amilovanovicTT:
### Problem description
There were multiple issues with `TTNNToEmitC` conversion of trace ops:
- `WriteTensorOp`: conversion pattern was calling non-existing `tt::tt_metal::write_tensor`
- `CaptureOrExecuteTraceOp`: 
    - converted op supported only one output
    - name collision between global input slots and input parameters
    - name collision in case of multiple `capture_or_execute` ops per forward function

### What's changed
- `WriteTensorOp`: conversion pattern is now calling `tt::tt_metal::copy_to_device`
- `CaptureOrExecuteTraceOp`: 
    - enabled multiple outputs by wrapping them in a result tuple
    - unique trace function suffix is used to prevent name collisions   
    - removed device parameter from the `capture_or_execute` signature as it was previously unused
 - Added simple lit test 

### Checklist

### DIFF
--- a/lib/Conversion/TTNNToEmitC/TTNNToEmitC.cpp
+++ b/lib/Conversion/TTNNToEmitC/TTNNToEmitC.cpp
@@ -7,6 +7,7 @@
 #include "ttmlir/Conversion/TTNNToEmitC/EmitCConversion.h"
 #include "ttmlir/Dialect/TTCore/IR/TTCoreOps.h"
 #include "ttmlir/Dialect/TTNN/IR/TTNNOps.h"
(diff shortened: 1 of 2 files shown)


sdjordjevicTT (approved):
Thanks for catching and fixing this! We should consider adding some tests that will compile the emitted C++ code and execute it, to prevent these kinds of regressions in the future. Also, please consider adding one codegen example with trace enabled in the tt-xla repo as well.

svuckovicTT (approved):

Issue #8030, opened 2026-04-20
TTNNOperationValidationAndFallback converts KV cache dtype 


kdimicTT:
**Branch:** `kdimic/ttnn-kv-cache-dtype-conversion`

### Description
Running ttir-to-ttnn pipeline on **Llama 3.2 3B** with `experimental-kv-cache-dtype=bfp_bf8`, the resulting TTNN IR differs between opt levels in a way that breaks the KV cache.

`%arg6` is the KV cache tensor passed as a function argument. Its dtype is set by `TTNNKVCacheDtypeConversion` when `experimental-kv-cache-dtype` is specified. 

With optimization_level = 2 `TTNNOperationValidationAndFallback` inserts `to_layout` conversions (bfp_bf8 → bf16) before `fill_cache`:

```mlir
%178 = "ttnn.to_layout"(%arg6) <{dtype = #ttcore.supportedDataTypes<bf16>, layout = #ttnn.layout<tile>, memory_config = #ttnn.memory_config<<dram>, <interleaved>>}>
  : (tensor<32x8x128x128x!ttcore.tile<32x32, bfp_bf8>, #ttnn.ttnn_layout<(d0, d1, d2, d3) -> (d0 * 1024 + d1 * 128 + d2, d3), <1x1>, memref<1024x4x!ttcore.tile<32x32, bfp_bf8>, #ttnn.buffer_type<dram>>, <interleaved>>>)
 -> tensor<32x8x128x128xbf16, #ttnn.ttnn_layout<(d0, d1, d2, d3) -> (d0 * 1024 + d1 * 128 + d2, d3), <1x1>, memref<1024x4x!ttcore.tile<32x32, bf16>, #ttnn.buffer_type<dram>>, <interleaved>>>

%179 = "ttnn.to_layout"(%177) <{dtype = #ttcore.supportedDataTypes<bf16>, layout = #ttnn.layout<tile>, memory_config = #ttnn.memory_config<<dram>, <interleaved>>}>
[...]

PR #8098, opened 2026-04-23
PagedUpdate|FillCacheOp validation in BeamSearch

rpavlovicTT:
PagedUpdateCacheOp is an in-place op with no tensor result, so isValidAnalysisTarget returned false and the optimizer never validated its input layouts. The fill-value input (operand 1) must be L1 height-sharded with numUsers cores, but OperationValidationAndFallback had no way to steer upstream ops toward that layout — it could only try DRAM-output fallbacks, all of which fail the kernel's is_sharded() assertion.

- GreedyMemoryLayoutPropagation: catch `FillCacheOp` and `PagedUpdateCacheOp` and register a null-output OpConfig in legalConfigs so `processOp` validates their input layout combinations.
- MemoryLayoutPropagation::run(): widen gate via isBeamSearchTarget so both ops enter processOp and drive upstream reshards backward.

PagedUpdateCacheRuleBook: getInputLayoutFilter(operand=1) forces requireL1HeightSharded so the beam search only considers L1 HeightSharded reshard candidates for the fill value. OpModel then picks the specific {numUsers, 1} grid. A proper ToMemoryConfigOp (tracked by L1SpillManagement) is inserted instead of the old workaround ToLayoutOp.

PagedUpdateCacheOpRewritePattern: disabled (returns failure immediately). The optimizer now handles the layout constraint end-to-end.

Related https://github.com/tenstorrent/tt-mlir/issues/8030

### DIFF
--- a/include/ttmlir/Dialect/TTNN/Utils/OptimizerUtils.h
+++ b/include/ttmlir/Dialect/TTNN/Utils/OptimizerUtils.h
@@ -6,13 +6,43 @@
 #define TTMLIR_DIALECT_TTNN_UTILS_OPTIMIZERUTILS_H
 
 #include "ttmlir/Dialect/TTNN/Analysis/OpConfig.h"
+#include "ttmlir/Dialect/TTNN/IR/TTNNOps.h"
 #include "ttmlir/Dialect/TTNN/IR/TTNNOpsAttrs.h"
 
+#include "mlir/IR/Operation.h"
 #include "llvm/ADT/SmallVector.h"
 
 #include <vector>
 
 namespace mlir::tt::ttnn::optimizer_utils {
+
+/// Returns true if the op produces a ranked tensor result that the layout
+/// optimizer can assign a memory layout to.
+inline bool opHasTensorResult(mlir::Operation *op) {
+  if (op->getNumResults() == 0) {
+    return false;
+  }
+  if (!llvm::isa<mlir::RankedTensorType>(op->getResult(0).getType())) {
+    return false;
+  }
+  if (llvm::isa<EmptyOp>(op)) {
+    return false;
+  }
+  return true;
+}
+
+/// Returns true for in-place ops with no tensor result that still need
+/// input layout validation and upstream reshard propagation in the beam search.
+inline bool isSinkOp(mlir::Operation *op) {
+  return llvm::isa<FillCacheOp, PagedUpdateCacheOp>(op);
+}
+
+/// Returns true if this op should participate in the greedy beam search —
+/// either because it produces a tensor output (normal path) or because it
+/// is a sink that drives upstream input layout decisions.
+inline bool isBeamSearchTarget(mlir::Operation *op) {
+  return opHasTensorResult(op) || isSinkOp(op);
+}
 
 // Create affine maps that translate a virtual grid layout to a physical
 // grid layout and vice versa for a single device based on the specified tensor
--- a/lib/Dialect/TTNN/Analysis/LegalOpLayoutAnalysis.cpp
+++ b/lib/Dialect/TTNN/Analysis/LegalOpLayoutAnalysis.cpp
@@ -137,18 +137,6 @@ bool incompatibleWithOverride(
 }
 
 // Skip operations that don't have nontrivial output tensors.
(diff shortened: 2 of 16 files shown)


vmilosevic (commented):

include/ttmlir/Dialect/TTNN/Utils/OptimizerUtils.h
```diff
@@ -6,14 +6,44 @@
 #define TTMLIR_DIALECT_TTNN_UTILS_OPTIMIZERUTILS_H
 
 #include "ttmlir/Dialect/TTNN/Analysis/OpConfig.h"
+#include "ttmlir/Dialect/TTNN/IR/TTNNOps.h"
 #include "ttmlir/Dialect/TTNN/IR/TTNNOpsAttrs.h"
 
 #include "llvm/ADT/SmallVector.h"
```
vmilosevic: :warning: **llvm\-include\-order** :warning:
```suggestion
#include "mlir/IR/Operation.h"
#include "llvm/ADT/SmallVector.h"
```

include/ttmlir/Dialect/TTNN/Utils/OptimizerUtils.h
```diff
@@ -6,14 +6,44 @@
...
+#include "mlir/IR/Operation.h"
```
vmilosevic, line 12: :warning: **llvm\-include\-order** :warning:
```suggestion

```

test/unittests/Optimizer/TestLegalLayoutAnalysis.cpp
```diff
@@ -4,6 +4,7 @@
 
 #include "ttmlir/Dialect/TTCore/IR/TTCore.h"
 #include "ttmlir/Dialect/TTNN/Analysis/LegalOpLayoutAnalysis.h"
+#include "ttmlir/Dialect/TTNN/Utils/OptimizerUtils.h"
```
vmilosevic: :warning: **llvm\-include\-order** :warning:
```suggestion

```

lib/Dialect/TTNN/Analysis/LegalOpLayoutAnalysis.cpp
```diff
@@ -137,18 +137,6 @@ bool incompatibleWithOverride(
 }
 
 // Skip operations that don't have nontrivial output tensors.
```
bmalesevicTT: Leftover comment

lib/Dialect/TTNN/Analysis/OpRules/TransformerRules.cpp
```diff
@@ -79,4 +79,24 @@ OutputHints SplitQKVRuleBook::getOutputHints(
...
+//===----------------------------------------------------------------------===//
+// PagedUpdateCacheRuleBook
+//===----------------------------------------------------------------------===//
+
+LayoutFilterFn
+PagedUpdateCacheRuleBook::getInputLayoutFilter(unsigned operandIdx) const {
+  // Operand 1 (fill value) must be L1 height-sharded.
+  // Reject interleaved and all other sharding types so the beam search
+  // is forced to explore HeightSharded reshard candidates. OpModel then
+  // picks the specific grid (numUsers cores) that satisfies the kernel.
```
bmalesevicTT: Are we sure the {numUsers, 1} grid will always be included after the cutoff to maxReshardCandidatesPerType, especially as numUsers can be smaller? 
Maybe also a warning about how choosing smaller maxReshardCandidatesPerType or maxInputCandidatesPerOperand, or batch=numUsers being small could fail this op by not keeping valid grid?

bmalesevicTT (approved):

odjuricicTT (approved):

sdjordjevicTT (approved):

lib/Dialect/TTNN/Analysis/OpRules/TransformerRules.cpp
rpavlovicTT: Actually tt-metal doesn't have hard constraint on this dim to be eq. to `numUsers`. So by forcing HS we should be fine.

Issue #8071, opened 2026-04-22
Enable distributed rms norm


mvasiljevicTT:
Distributed rms norm is now supported through composite ops, but is currently decomposing in llama 70b tp2x4.

This should be fixed so that we use ttnn.distributed_rms_norm op. Based on [this analysis](https://github.com/tenstorrent/tt-mlir/issues/6803#issuecomment-4023124664), it is expected to massively improve device time, and with that also e2e time.

mvasiljevicTT:
Fix on branch 

[galaxy CI](https://github.com/tenstorrent/tt-xla/actions/runs/24781748566)
[llmbox CI](https://github.com/tenstorrent/tt-xla/actions/runs/24789272982/job/72553349035) - with opt level 1

locally it breaks due to this [prefill issue](https://github.com/tenstorrent/tt-mlir/issues/8068), --decode-only works fine. 

This change is affecting only decode.

mvasiljevicTT:
Works fine on CI with 10sps on galaxy, so it is not blocked by the prefill issue

sdjordjevicTT:
@mvasiljevicTT, can we close this one?

mvasiljevicTT:
@sdjordjevicTT there is still [one loose end](https://github.com/tenstorrent/tt-mlir/issues/8205) regarding this effort, it will be finished soon

sdjordjevicTT:
Oh, I see, thanks for the update!

PR #8099, opened 2026-04-23
Enable DistributedRMSNorm in Llama

mvasiljevicTT:
### Ticket
[Link to Github Issue](https://github.com/tenstorrent/tt-mlir/issues/8071)

### Problem description
`DistributedRMSNorm` is currently lowered as a composite op, but is decomposed in llama 70b because it has shape `(32, 1, M)`, and only `(1, 1, 32, M)` is supported.
#### General support:
- added commute pattern for `DistributedRMSNorm` so that reshapes around it can be erased, leaving op in shape `(32, M)`
- added workaround to reshape `DistributedRMSNorm` of any `(1,* 32)` shape to canonical `(1, 1, 1, 32)`, and loosened decomposition check to accept `(1,* 32)` relying on the workaround to fix shape later.

#### Changes needed to make it work with trace enabled:
- `empty op` that was added in the workaround to be used as a buffer was breaking chain of hoistable ops.
 It is now moved to the top of the graph, included in prelude and recognized as a constant, so all `distributed_rms_norm` ops with same shape (which is all in llama) now use same main arg as input for the buffer. (`DistributedRMSNormWidthShardInputRewritePattern`,  `TTNNTraceHoistTransform`)
- after warmup, capture run was creating new `GlobalSemaphore`s, which was breaking trace capture. 
Now root is responsible for creating and storing semaphores as needed (`getOrCreateImplicitGlobalSemaphore`), and child runs have `parentContext` needed to trigger it, so semaphores are created only once during warmup run, and capture run reuses them.

#### Llama trace llmbox test
- added ttir of one-layer llama 70b on llmbox and test that covers:
   - compiling with mesh 2,4 and all flags used in perf benchmarks in tt-xla inluding trace, constevals and opt level 1
   - file check on fused ops in ttnn - distributed_rms_norm, sdpa, rope, concat_heads
   - ttrt run with program cache
 
### Checklist
- [x] [galaxy perf benchmark](https://github.com/tenstorrent/tt-xla/actions/runs/24781748566/job/72519554720) 5.3 -> 10.4 sps
- [x] [llmbox perf benchmark](https://github.com/tenstorrent/tt-xla/actions/runs/24789272982/job/72553349035) 2.61 -> 4.2 sps


include/ttmlir/Dialect/TTNN/Transforms/Workarounds/Decomposition/DistributedRMSNormReshapeToCanonicalShapeRewritePattern.h
```diff
@@ -0,0 +1,31 @@
...
+
+#include "mlir/IR/PatternMatch.h"
+#include "mlir/Support/LogicalResult.h"
+
+namespace mlir::tt::ttnn::workarounds::decomposition {
+
+// Reshapes the input/residual of a DistributedRMSNormOp into the canonical
+// (1, 1, 32, M) shape required by the tt-metal fused_rms_minimal kernel and
+// reshapes the result back afterwards.  Must run before
+// DistributedRMSNormWidthShardInputRewritePattern (enforced via a higher
```
sdjordjevicTT: In MLIR, the `PatternBenefit` serves primarily as a heuristic for prioritizing patterns rather than as a strict sequencing mechanism. If you require a specific execution order, we will need to find an alternative way to enforce that. I suggest relocating it to `lib/Dialect/TTNN/Transforms/Decomposition/DistributedRMSNormDecompositionRewritePattern.cpp` once we confirm that we are targeting the fused kernel.

.github/settings/tests.json
```diff
@@ -32,6 +30,7 @@
     { "runs-on": "n150",   "image": "tracy",  "script": "ttnn_runtime.sh", "args": "runtime/test/ttnn/python/distributed/n150" },
     { "runs-on": "n300",   "image": "tracy",  "script": "ttnn_runtime.sh", "args": "runtime/test/ttnn/python/n300" },
     { "runs-on": "llmbox", "image": "tracy",  "script": "ttnn_runtime.sh", "args": "runtime/test/ttnn/python/distributed/llmbox" },
+    { "runs-on": "llmbox", "image": "speedy", "script": "ttrt.sh", "args": ["run", "Dialect/TTNN/trace/scenarios/llmbox", "--enable-program-cache", "--loops", "10"] },
```
sdjordjevicTT: I am just doble checking, Is this intended change, we want to add this CI job?

sdjordjevicTT (commented):
Looks good, but I left one critical comment that we should address.

include/ttmlir/Dialect/TTNN/Transforms/Workarounds/Decomposition/DistributedRMSNormReshapeToCanonicalShapeRewritePattern.h
mvasiljevicTT: Thanks, moved now to decomposition

.github/settings/tests.json
mvasiljevicTT: It needs to be added as a separate entry as it is the only test we have that runs program compiled with trace, which need `enable-program-cache` and `loops`.

This might be because it should not be done that way. Not sure who to ping for that? 

I am still not sure what is the best place for that test

sdjordjevicTT (approved):
Looks great! Thanks!

.github/settings/tests.json
sdjordjevicTT: Me neither, but at least if we tend to run this on silicon, can you please move the test under the Silicon folder?

nsumrakTT (approved):

.github/settings/tests.json
mvasiljevicTT: Currently whole `build/test/ttmlir/Silicon` will be used with `ttrt run`, and this [test fails](https://github.com/tenstorrent/tt-mlir/actions/runs/24832976203/job/72688588946) without `enable-program-cache` flag. Some of the tests there have ttrt run line in the file, but I think that is a bug, and they are actually run twice.

mvasiljevicTT:
Note: Using `parentContext` is a temporary solution for fixing trace issues with global semaphores. Passing semaphores as explicit program inputs using global semaphore pool is currently [on branch](https://github.com/tenstorrent/tt-mlir/compare/mvasiljevic/distributed_rms_norm_wo_test_change), and will be done as separate PR.

pilkicTT (approved):
>Note: Using parentContext is a temporary solution for fixing trace issues with global semaphores. Passing semaphores as explicit program inputs using global semaphore pool is currently [on branch](https://github.com/tenstorrent/tt-mlir/compare/mvasiljevic/distributed_rms_norm_wo_test_change), and will be done as separate PR.

thanks

odjuricicTT (approved):

PR #8109, opened 2026-04-24
[D2M] Add lowering of scalar runtime kernel args

tapspatel:


lib/Conversion/TTKernelToEmitC/TTKernelToEmitC.cpp
```diff
@@ -444,6 +444,87 @@ class TTKernelToEmitCOpaqueRewriter : public OpConversionPattern<SourceOp> {
...
+      mlir::OpPrintingFlags flags;
+      op.getResult().printAsOperand(os, flags);
+      os.flush();
+    }
+    std::string varName = "_rc" + ssaName.substr(1);
+
+    if (mlir::isa<Float32Type>(resultType)) {
+      // "{{" escapes to "{" in EmitC verbatim; "}" needs no escaping.
+      std::string code = "union {{ uint32_t _u; float _f; } " + varName + "; " +
+                         varName + "._u = {};";
```
nsmithtt: This union trick is technically UB.  We should use this pattern instead (or vice versa):
```
      builder->create<emitc::VerbatimOp>(
          loc, "inline uint32_t float_to_bits(float f) { "
               "uint32_t r; __builtin_memcpy(&r, &f, sizeof(r)); return r; }");
```

include/ttmlir/Dialect/TTKernel/IR/TTKernelOps.td
```diff
@@ -3597,6 +3597,23 @@ def TTKernel_GetCompileArgValOp : TTKernel_Op<"get_compile_time_arg_val",
 // TTKernel Helper functions
 //===----------------------------------------------------------------------===//
 
+def TTKernel_ReinterpretCastOp : TTKernel_Op<"reinterpret_cast", [Pure]> {
```
nsmithtt: Naming nit, this should be called `BitcastOp`, C++ reinterpret_cast will actually do int->float or float->int conversion, but that's not what we're doing with this op.

nsmithtt (approved):
Looks great! Please address the 2 comments inline, once those are resolved LGTM

include/ttmlir/Dialect/TTKernel/IR/TTKernelOpsTypes.td
```diff
@@ -121,6 +121,7 @@ def TTKernel_MeshPosition : TTKernel_Type<"MeshPosition", "mesh_position"> {
     }];
 }
 
+
```
vwellsTT: nit: rm

lib/Conversion/TTKernelToEmitC/TTKernelToEmitC.cpp
vwellsTT: I think we have some version of this elsewhere already as a util func--maybe we should commonize to shared header if needed s.t. we can avoid re-emitting, FWIW

lib/Target/TTMetal/TTMetalToFlatbuffer.cpp
```diff
@@ -863,9 +892,27 @@ std::shared_ptr<void> translateTTMetalToFlatbuffer(
 
     cqBuilder.inputs.reserve(entry.getBody().getArguments().size());
     for (auto &input : entry.getBody().getArguments()) {
-      cqBuilder.inputs.push_back(
-          cache.getOrCreate(input, bufferValueToFlatbuffer, systemDesc, 0));
-      tensorInputs.push_back(tensorValueToFlatbuffer(cache, input));
+      if (mlir::isa<IntegerType, IndexType, FloatType>(input.getType())) {
```
vwellsTT: nit: don't feel strongly, but maybe it's worth slapping a "isSupportedScalarType" func into a util header for this specific `isa` set, since it's used several times

vwellsTT (approved):

lib/Target/TTMetal/TTMetalToFlatbuffer.cpp
tapspatel: fixed

include/ttmlir/Dialect/TTKernel/IR/TTKernelOpsTypes.td
tapspatel: fixed

PR #8126, opened 2026-04-27
Remove unnecessary stablhelo.scatter decomposition check

sdjukicTT:
### Ticket
https://github.com/tenstorrent/tt-xla/issues/4407
Wormhole lb test log after change: [log.txt](https://github.com/user-attachments/files/27126667/log.txt)

### Problem description
There's a StableHLO Scatter Op decomposition condition that checks if input tensor indexed dim is always 0 in single-dimensional scatter case, but TTIR and TTNN support any indexed dim.

### What's changed
Removed unnecessary condition and added a test for this case.

### Checklist

### DIFF
--- a/test/ttmlir/Conversion/StableHLOToTTIR/scatter_op.mlir
+++ b/test/ttmlir/Conversion/StableHLOToTTIR/scatter_op.mlir
@@ -108,6 +108,24 @@ module @jit_scatter attributes {} {
         return %result : tensor<1x4x3xi64>
     }
 
+    func.func public @test_scatter_nonzero_operand_dim(%operand: tensor<2x4x3xbf16>, %updates: tensor<2x1x3xbf16>) -> tensor<2x4x3xbf16> {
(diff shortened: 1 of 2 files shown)


acicovicTT (approved):

test/ttmlir/Conversion/StableHLOToTTIR/scatter_op.mlir
mrakitaTT, line 1: Maybe you can add test for some more dimensions, since you are now removing check for all dimensions, just in case we got lucky and it works only on dim 0 and 1 :)

mrakitaTT (approved):

test/ttmlir/Conversion/StableHLOToTTIR/scatter_op.mlir
sdjukicTT: I added a second test where dim=2

sdjordjevicTT:
Nothing blocking, but we should consider adding builder tests for different stablehlo scatter variants in order to have a silicon test coverage. Please consider adding those tests as a follow-up.

Issue #4563, opened 2025-08-20
D2M Gather


nsmithtt:
# Gather

This issue tracks the implementation / lowering of the gather op into d2m
generic.

## Prereqs

- `riscv_l1` memory space
  - #4564 
- TTIRToTTIRDecomp matches this op up to an embedding op, we don't want to do
  this for D2M and in fact we'd want to do the opposite and lower embedding to
  gather as part of TTIR to d2m lowering.
  - https://github.com/tenstorrent/tt-mlir/issues/3024
- Some investigation of non-padded L1 tensors
- i64 -> i32 conversion, do we need to use the host to do this?
- Research mlir [tensor.gather](https://mlir.llvm.org/docs/Dialects/TensorOps/#tensorgather-tensorgatherop), eventually we want to support lowering this op too.

## TTIR Gather Op

Snippet copy paste of the gather op:

```mlir
%input = ... : tensor<5x3xf32>         // Input tensor with shape [5,3]
%indices = ... : tensor<2xi64>         // Indices tensor with values [2, 1]
%output = ttir.empty() : tensor<3xf32> // Output tensor
%result = ttir.gather(%input, %indices, %output) {
    offset_dims = [0],                 // Output dimensions that are gathered from input
    collapsed_slice_dims = [0],        // Input dimensions that are collapsed
    operand_batching_dims = [],        // Batch dimensions of the input
    start_indices_batching_dims = [],  // Batch dimensions of the indices
    start_index_map = [0],             // Maps indices to input dimensions
    index_vector_dim = 0,              // Which dimension of indices contains the index vector
    slice_sizes = [1, 3],              // Size of the slice to extract from each position
    indices_are_sorted = false         // Whether indices are sorted
} : tensor<5x3xf32>, tensor<2xi64>, tensor<3xf32> -> tensor<3xf32>
```

## Generic Lowering

There are a lot of features and attributes on the gather op, for the purposes of
this issue we are going to simplify and just focus on a simple gather case
(typical access pattern of an embedding op):
- `input`: Your input tensor, from which we are gathering from. Has shape
  100x240xf32.
- `indices`: In this example we'll just be gather rows from the input. Has shape
  40xi32, contents [20, 2, 1, 9, ...] meaning we'll gather these input rows in that order.
- `output`: The output tensor we are gathering into. Has shape 40x240xf32, contains
  input rows listed by `indices`.

```mlir
%input = ... : tensor<100x240xf32>
[...]

PR #8133, opened 2026-04-27
[d2m] embedding op

### LINKED ISSUE #3024: d2m ttir.gather

dlokeTT:
### Ticket
Closes #3024

### Problem description
Need to support ttir.embedding in d2m

### What's changed
- Lower `ttir.embedding` into a  datamovement `d2m.generic` with a new `d2m.embedding` op that takes in the index, weight and output tensor
- Added a new `d2m.indexed_row_copy` memref op for post bufferization that deals with the scratch buffers needed for staging the dma
- Allocator uses the embedding table dtype for the output instead of the weights
- The `D2MToTTKernel` lowering implements the row_copy_index as a data movement row gather over collapsed 2D sharded L1 memrefs
- Added a new ttkernel op `LoadFromL1Op` used for loading an i32 value from risc cores L1 and its conversion to emitC
- Added lit test coverage for embedding for dialect conversion passes and bufferization
- Added builder tests for different shapes and dtypes

### Checklist


lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
```diff
@@ -3044,6 +3044,105 @@ class D2MArangeOpRewriter : public OpConversionPattern<ttir::ArangeOp>,
...
+        mlir::cast<ttcore::MetalLayoutAttr>(outputTensorType.getEncoding());
+    auto grid = ttcore::GridAttr::get(
+        rewriter.getContext(), outputLayout.getGridShape(outputTensorType));
+    ArrayAttr threads = rewriter.getArrayAttr(
+        {rewriter.getAttr<d2m::ThreadAttr>(d2m::ThreadType::Datamovement)});
+
+    auto generic = rewriter.create<d2m::GenericOp>(
+        loc, TypeRange(outputs), inputs, outputs,
+        /*additionalArgs=*/ValueRange(), grid, rewriter.getI64ArrayAttr({}),
+        rewriter.getAffineMapArrayAttr({}), rewriter.getArrayAttr({}), threads,
```
nsmithtt: If we program these affine maps and iterator types then I don't think we'll need any special casing in the GridSelection pass and potentially the other passes as well.  I outlined how this could look for the gather op which is semantically almost identical [for reference](https://github.com/tenstorrent/tt-mlir/issues/4563):

There are a lot of features and attributes on the gather op, for the purposes of
this issue we are going to simplify and just focus on a simple gather case
(typical access pattern of an embedding op):
- `input`: Your input tensor, from which we are gathering from. Has shape
  100x240xf32.
- `indices`: In this example we'll just be gather rows from the input. Has shape
  40xi32, contents [20, 2, 1, 9, ...] meaning we'll gather these input rows in that order.
- `output`: The output tensor we are gathering into. Has shape 40x240xf32, contains
  input rows listed by `indices`.

```mlir
%input = ... : tensor<100x240xf32>
%indices = ... : tensor<4xi32>
%output = ttir.empty() : tensor<40x240xf32>
%devInput = to_layout %input : tensor<2x4x64x64xf32, metal_layout<...>>  // Padded to tile then divided on 2x4 grid
%devIndices = to_layout %indices : tensor<2x1x1x20xi32, metal_layout<...>> // Normalized rank, then divided on 2x1 grid
%devOutput = to_layout %output : tensor<2x4x32x64xf32, metal_layout<>> // Padded to tile then divided on 2x4 grid
%7 = ttir.generic {
  block_factors = [1, 1],
  grid = #ttcore.grid<2x4>,
  indexing_maps = [
    affine_map<(d0, d1) -> (d0, d1)>, // Eltwise / Parallel Datamovement
    affine_map<(d0, d1) -> (d0, 0)>,  // Broadcast to all core columns
    affine_map<(d0, d1) -> (d0, d1)>  // Eltwise / Parallel Datamovement
  ],
  iterator_types = [#ttcore.iterator_type<parallel>, #ttcore.iterator_type<parallel>],
  threads = [#ttir.thread<datamovement>]}
    ins(devInput, devIndices)
    outs(devOutput)  {
^datamovement(
  %input: memref<64x64xf32, #ttcore.memory_space<l1>>,
  %indices: memref<1x20xi32, #ttcore.memory_space<l1>>,
  %out: memref<32x64xf32, #ttcore.memory_space<l1>>,
  ): {
  %riscv = memref.memory_space_cast %indices : #ttcore.memory_space<riscv_l1>
  scf.for %i in 20 {
    %index = memref.load %riscv[%i]
    dma %input[%index], %out[%i]
  }
}
```

Let me know if that makes sense.

include/ttmlir/Dialect/D2M/Analysis/Allocation/Utils.h
```diff
@@ -431,17 +431,20 @@ inline std::array<int64_t, 2> getEffectiveTileShape(Type elementType) {
...
-          .getElementType();
-  for (std::size_t operandIndex = 1;
-       operandIndex < genericOp.getOutputs().getBeginOperandIndex();
-       ++operandIndex) {
-    TT_assertv(inputElementType ==
-                   mlir::cast<MemRefType>(
-                       genericOp->getOperand(operandIndex).getType())
-                       .getElementType(),
-               "expected no change in tile shapes across generic op inputs");
+  Type inputElementType = genericOp.getIndexedRowGatherDataElementType();
```
nsmithtt: Why do we need this change? @anuragsingh-tt can you review?

include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
```diff
@@ -1548,6 +1548,84 @@ def D2M_DMAWaitOp : D2M_GenericRegionDatamovementOp<"dma_wait", [MemoryEffects<[
     let assemblyFormat = [{ $mem_tx attr-dict `:` qualified(type($mem_tx)) }];
 }
 
+def D2M_IndexedRowCopyOp : D2M_GenericRegionDatamovementOp<"indexed_row_copy",
```
nsmithtt, line 1616: Not opposed to introducing this primitive, but this might be accomplished by extending `local_copy` to support dynamic indices:
```mlir
scf.for %i in ... {
  d2m.local_copy %src[%I], %dst[%i]
}
```

@vtangTT, not sure how much work that would be.

nsmithtt (approved):
One comment on the allocator change, and another high level direction that would be worth investigating as follow on work.  Tentatively approving as long as @anuragsingh-tt signs off on allocator

include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
dlokeTT: Yeah I think this can be done. I can do this in a follow on, lmk what you think @vtangTT

include/ttmlir/Dialect/D2M/Analysis/Allocation/Utils.h
dlokeTT: We need this because allocator currently trips up if the inputs to a generic are different types which they are for embedding since the index tensor is some integer type and the embedding tensor is usually a float. This is a kind of hacky fix to tell allocator to look at the embedding tensors dtype instead of just the first input.

dlokeTT: Cleaned this up a bit. We still need to special case a little because the normal affine verifier fails because embedding has indexing based on the embedding tensor but removed the tablegen helper and the embedding specific path in this file now more generally just checks the effective tile shape instead of dtype since in embedding all tensors are row major

include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
vtangTT: Yes, we should expand the `local_copy` op to support dynamic runtime symbols. The mechanism would be similar to `affine.load`'s existing dim+symbol operand convention and would remove a lot of the code added for the new op.

I think changes should hopefully just boil down to some modifications in the .td for `local_copy` and some fixes in `LowerDMAToFullyIndexedForm`

Ideally, this PR uses the `local_copy` but I don't want to block, so can do as a follow on.

dlokeTT: Filed an issue here to track switching over to `local_copy` https://github.com/tenstorrent/tt-mlir/issues/8186

include/ttmlir/Dialect/D2M/Analysis/Allocation/Utils.h
anuragsingh-tt, line 1: This is a better invariant to check here than what existed before.

lib/Dialect/D2M/IR/D2MOps.cpp
```diff
@@ -1787,7 +1796,7 @@ ::mlir::LogicalResult d2m::GenericOp::verify() {
   bool hasGrid = mlir::isa<MemRefType>(getOutputs().front().getType()) ||
                  (rankedTensorType && rankedTensorType.getEncoding());
   SmallVector<AffineMap> indexingMaps = getIndexingMapsValue();
-  if (hasGrid && !indexingMaps.empty()) {
+  if (hasGrid && !indexingMaps.empty() && !hasIndexedRowGather(*this)) {
```
anuragsingh-tt: Why are we skipping these validations?

anuragsingh-tt (approved):
LGTM. Thanks for your patience.

+1 to DMA local_copy change. I think we should also try to use existing DMA optimizations instead of explicitly managing the read/write barriers in TTKernel. But that's an optimization that can be a follow up.

lib/Dialect/D2M/IR/D2MOps.cpp
dlokeTT: Reworked this a bit. We now verify but skip the check on the shard shape comparison when the indexing map implies broadcasting across loop dimensions which is what embedding trips up on

PR #8106, opened 2026-04-23
[d2m] Introduce SynchronizableOpInterface

sohaibnadeemTT:
### Problem description/ What's changed
This PR introduces the `SynchronizableOpInterface`, an interface for inferring which ops need synchronization in a general manner for use in passes like `Allocator` (to determine which memref allocs need to be back-buffered) and `SplitUnifiedThread` (to correctly insert cb ops based on producer/consumer relationships of the ops that need synchronization). This will reduce the surface-area for bugs and allow us to more easily add future ops that require synchronization. 

We also add a new op `d2m.synchronized_region`, which allows us encapsulate a bunch of non-synchronized ops into an op that generically implements `SynchronizableOpInterface`.

Currently, the interface is implemented for `remote_load`, `remote_store`, `linalg.generic` (via an external model), `local_copy`, `tile_tilize_block`, `tile_untilize_block`, and `synchronized_region`.

### Checklist
- [x] New/Existing tests provide coverage for changes

PR #8026, opened 2026-04-20
Introduce SynchronizableOpInterface + related CB rework

sohaibnadeemTT:
### Ticket
Link to Github Issue

### Problem description
Provide context for the problem.

### What's changed
Describe the approach used to solve the problem.
Summarize the changes made and its impact.

### Checklist
- [ ] New/Existing tests provide coverage for changes

PR #8135, opened 2026-04-27
[d2m] Refactor SplitUnifiedThread (and other backend passes) using SynchronizableOpInterface

sohaibnadeemTT:
### Ticket
Link to Github Issue: https://github.com/tenstorrent/tt-mlir/issues/8308

### Problem description/What's changed
This PR refactors aims to refactor the SplitUnifiedThread pass to insert CB ops using producer/consumer relationship info derived using the getCBUsageInfo utility function in SynchronizableOpInterfaceUtils.cpp, generalizing the usage of CBs and synchronization so we don't need special case handling for each op.

The getCBUsageInfo can provide us information on what the producer and consumer is of each CB. Using this, we can insert reserve+push ops on producer side and wait+pop ops on consumer side. If either producer/consumer is an aliased load or store, we move the CB ops to the other side since the aliased op will be erased. 

This PR also changes the get_cb op to be referenced by its operand index in the generic op. Enabling this requires converting in-generic alloc's that back aliased loads/stores to a new `d2m.operand_alias` op so that they can be hoisted similar to real allocs and therefore be referred to by their operand index in the generic op. This significantly simplifies referencing CBs and pushes down the need to use CB port numbers down to runtime lowering where they are actually needed. As a result, some backend passes (including D2MToTTMetal and D2MToTTNN) have been cleaned up.

### Checklist


vmilosevic (commented):

lib/Dialect/D2M/Transforms/HoistCBAllocs.cpp
```diff
@@ -93,59 +81,35 @@ class D2MHoistCBAllocs : public impl::D2MHoistCBAllocsBase<D2MHoistCBAllocs> {
...
+    // hoist aliased CBs
+    genericOp->walk([&](memref::AllocOp allocOp) {
+      if (auto aliasForOperandAttr =
+              allocOp->getAttrOfType<IntegerAttr>("d2m.alias_for_operand")) {
+        // llvm::errs() << "hoisting aliased CB: " <<
+        // genericOp.getOperand(operandIndex) << "\n";
+        rewriter.setInsertionPoint(genericOp);
+        auto aliasedOperand =
+            genericOp.getOperand(aliasForOperandAttr.getInt());
+        // TODO: add back later
```
vmilosevic: :warning: **google\-readability\-todo** :warning:
missing username/bug in TODO
```suggestion
        // TODO(nsmith): add back later
```

test/unittests/D2MGenericAnalysis/TestD2MGenericAnalysis.cpp
```diff
@@ -1066,9 +1066,5 @@ func.func @test(
   }
   ASSERT_EQ(constantCount, 2u);
   EXPECT_EQ(collapseCount, 3u);
-  EXPECT_EQ(loadCount, 2u);
   EXPECT_EQ(storeCount, 1u);
-  EXPECT_EQ(addCount, 1u);
-}
-
 } // namespace mlir::tt::d2m
```
vmilosevic, line 1075: :warning: **clang\-diagnostic\-error** :warning:
expected `` } ``

lib/Dialect/D2M/Transforms/SplitUnifiedThread.cpp
```diff
@@ -25,10 +28,13 @@ namespace {
 // Helpers
 // ---------------------------------------------------------------------------
 
-// Returns true if the remote_load/store requires real DMA. This is the case
-// when the remote memref has a view layout, is in DRAM, or the local buffer is
-// a streaming CB (CBLayoutAttr). Aliased ops do not need DMA and return false.
-static bool needsDMA(Value memref, Value localBuffer) {
+// TODO: move into some DMA utils place alongside load/store aliasing logic
```
vmilosevic: :warning: **google\-readability\-todo** :warning:
missing username/bug in TODO
```suggestion
// TODO(nsmith): move into some DMA utils place alongside load/store aliasing logic
```

lib/Dialect/D2M/Transforms/SplitUnifiedThread.cpp
```diff
@@ -25,10 +28,13 @@ namespace {
...
+// which should be separated out to separate function after
+// inferOperandAliasing inferOperandAliasing needs to be run prior to
+// assigning memory to actually benefit from reduced mem usage from aliasing
+// TODO: add specific checks for load/store (mcast load, fabric store)
```
vmilosevic: :warning: **google\-readability\-todo** :warning:
missing username/bug in TODO
```suggestion
// TODO(nsmith): add specific checks for load/store (mcast load, fabric store)
```

lib/Dialect/D2M/Transforms/SplitUnifiedThread.cpp
```diff
@@ -45,491 +51,372 @@ static bool needsDMA(Value memref, Value localBuffer) {
...
-        for (Value result : op.getResults()) {
-          if (aliasedValues.insert(result).second) {
-            changed = true;
-          }
+Value traceComputeMemrefToCB(Value value, GenericOp genericOp) {
+  llvm::errs() << "tracing value: " << value << "\n";
+  while (value) {
+    // check if its a cb (hoisted generic arg with cb layout attr),
+    if (auto memrefType = mlir::dyn_cast<MemRefType>(value.getType())) {
+      // TODO: add back later
```
vmilosevic: :warning: **google\-readability\-todo** :warning:
missing username/bug in TODO
```suggestion
      // TODO(nsmith): add back later
```

lib/Dialect/D2M/Transforms/SplitUnifiedThread.cpp
```diff
@@ -45,491 +51,372 @@ static bool needsDMA(Value memref, Value localBuffer) {
...
+      // if (mlir::isa<ttcore::CBLayoutAttr>(memrefType.getLayout())) {
+      if (llvm::find(genericOp.getAdditionalArgs(), value) !=
+          genericOp.getAdditionalArgs().end()) {
+        // Skip scratch buffers
+        if (auto scratchBufferAttr =
+                mlir::dyn_cast<Operation *>(value.getDefiningOp())
```
vmilosevic: :warning: **clang\-analyzer\-core\.CallAndMessage** :warning:
Called C\+\+ object pointer is null

lib/Dialect/D2M/Transforms/SplitUnifiedThread.cpp
```diff
@@ -45,491 +51,372 @@ static bool needsDMA(Value memref, Value localBuffer) {
...
-      }
-    }
-    if (opUsesValue) {
-      lastUse = &op;
+    // Otherwise keep tracing up the chain, if we reach an op we don't support,
+    // stop tracing and return nullptr
+    if (auto subviewOp = mlir::dyn_cast<memref::CollapseShapeOp>(definingOp)) {
+      value = subviewOp.getSrc();
+      continue;
+    } else if (auto subviewOp = mlir::dyn_cast<memref::SubViewOp>(definingOp)) {
```
vmilosevic: :warning: **llvm\-else\-after\-return** :warning:
do not use `` else `` after `` continue ``
```suggestion
    } if (auto subviewOp = mlir::dyn_cast<memref::SubViewOp>(definingOp)) {
```

lib/Dialect/D2M/Transforms/SplitUnifiedThread.cpp
```diff
@@ -45,491 +51,372 @@ static bool needsDMA(Value memref, Value localBuffer) {
...
-      stores.push_back(op);
-    }
-  });
+// from cb usage info, check for load-store pairs and insert aliased cb ops for
+// alias side
+static LogicalResult processSharedBufferPairs(
+    Block *computeBlock, PatternRewriter &rewriter,
+    llvm::DenseMap<Value, utils::CBUsageInfo> &cbUsageInfo) {
+  for (auto [localBuffer, usageInfo] : cbUsageInfo) {
+    auto producer = usageInfo.producers.front();
```
vmilosevic: :warning: **llvm\-qualified\-auto** :warning:
`` auto producer `` can be declared as `` auto *producer ``
```suggestion
    auto *producer = usageInfo.producers.front();
```

lib/Dialect/D2M/Transforms/SplitUnifiedThread.cpp
```diff
@@ -45,491 +51,372 @@ static bool needsDMA(Value memref, Value localBuffer) {
...
+    auto consumer = usageInfo.consumers.front();
```
vmilosevic: :warning: **llvm\-qualified\-auto** :warning:
`` auto consumer `` can be declared as `` auto *consumer ``
```suggestion
    auto *consumer = usageInfo.consumers.front();
```

lib/Dialect/D2M/Transforms/SplitUnifiedThread.cpp
```diff
@@ -45,491 +51,372 @@ static bool needsDMA(Value memref, Value localBuffer) {
...
-      if (auto reserveOp = traceToDefiningOp<ReserveOp>(src)) {
-        srcCb = reserveOp.getCb();
-      } else {
-        srcCb = findAssociatedCB(copyOp, src, rewriter, cache, portCounters);
-      }
-      if (srcCb) {
-        rewriter.setInsertionPoint(copyOp);
-        rewriter.create<PushOp>(loc, srcCb);
+  computeBlock->walk([&](Operation *op) {
+    // TODO: see if removed check for explicit CB form
```
vmilosevic: :warning: **google\-readability\-todo** :warning:
missing username/bug in TODO
```suggestion
    // TODO(nsmith): see if removed check for explicit CB form
```

lib/Dialect/D2M/Transforms/SplitUnifiedThread.cpp
```diff
@@ -45,491 +51,372 @@ static bool needsDMA(Value memref, Value localBuffer) {
...
+        if (synchronizedOp.isConsumer(operand)) {
+          Location loc = synchronizedOp.getLoc();
+          Value localBuffer = operand.get();
+          unsigned cbOperandIdx =
+              synchronizedOp->getParentOfType<GenericOp>().getOperandIndex(
+                  localBuffer);
+
+          // get the associated producer for this operand
+          // Assumes only one producer for this local buffer
+          auto associatedProducer = cbUsageInfo[localBuffer].producers.front();
```
vmilosevic: :warning: **llvm\-qualified\-auto** :warning:
`` auto associatedProducer `` can be declared as `` auto *associatedProducer ``
```suggestion
          auto *associatedProducer = cbUsageInfo[localBuffer].producers.front();
```

lib/Dialect/D2M/Transforms/SplitUnifiedThread.cpp
```diff
@@ -45,491 +51,372 @@ static bool needsDMA(Value memref, Value localBuffer) {
...
+        if (synchronizedOp.isProducer(operand)) {
+          Location loc = synchronizedOp.getLoc();
+          Value localBuffer = operand.get();
+          unsigned cbOperandIdx =
+              synchronizedOp->getParentOfType<GenericOp>().getOperandIndex(
+                  localBuffer);
+
+          // get the associated consumer for this operand
+          // Assumes only one consumer for this local buffer
+          auto associatedConsumer = cbUsageInfo[localBuffer].consumers.front();
```
vmilosevic: :warning: **llvm\-qualified\-auto** :warning:
`` auto associatedConsumer `` can be declared as `` auto *associatedConsumer ``
```suggestion
          auto *associatedConsumer = cbUsageInfo[localBuffer].consumers.front();
```

vmilosevic (commented):

lib/Dialect/D2M/Transforms/SplitUnifiedThread.cpp
```diff
@@ -45,491 +51,372 @@ static bool needsDMA(Value memref, Value localBuffer) {
...
 // ---------------------------------------------------------------------------
 // DMA thread: convert implicit-form ops to explicit CB form
 // ---------------------------------------------------------------------------
 
+// Erase aliased load and store ops (no DMA needed).
+static LogicalResult eraseAliasedLoadStoreOps(
+    PatternRewriter &rewriter,
+    llvm::DenseMap<Value, utils::CBUsageInfo> &cbUsageInfo) {
+  for (auto [localBuffer, usageInfo] : cbUsageInfo) {
+    auto producer = usageInfo.producers.front();
```
vmilosevic: :warning: **llvm\-qualified\-auto** :warning:
`` auto producer `` can be declared as `` auto *producer ``
```suggestion
    auto *producer = usageInfo.producers.front();
```

lib/Dialect/D2M/Transforms/SplitUnifiedThread.cpp
```diff
@@ -45,491 +51,372 @@ static bool needsDMA(Value memref, Value localBuffer) {
...
+    auto consumer = usageInfo.consumers.front();
```
vmilosevic: :warning: **llvm\-qualified\-auto** :warning:
`` auto consumer `` can be declared as `` auto *consumer ``
```suggestion
    auto *consumer = usageInfo.consumers.front();
```

lib/Dialect/D2M/Transforms/SplitUnifiedThread.cpp
```diff
@@ -545,63 +432,27 @@ convertDMAToExplicitCBForm(Block *dmBlock, PatternRewriter &rewriter,
...
-    // L1-to-L1 pairs use the load's own (input) memref since the store side
-    // also redirects to the input memref for aliased loads.
-    Value cbMemref = memref;
-    auto pairIt = sharedPairs.find(loadOp.getLocalBuffer());
-    if (pairIt != sharedPairs.end() && !l1Pair) {
-      cbMemref = pairIt->second.second.getMemref();
-    }
+    // removed  needsDMA and isL1ToL1Pair checks tocheck for aliased load/store
+    // with just checking aliased load/store
+    // TODO: l1-to-l1 pair is equivalent to checking aliased load + aliased
```
vmilosevic: :warning: **google\-readability\-todo** :warning:
missing username/bug in TODO
```suggestion
    // TODO(nsmith): l1-to-l1 pair is equivalent to checking aliased load + aliased
```

lib/Conversion/D2MToTTMetal/D2MToTTMetal.cpp
```diff
@@ -137,51 +159,74 @@ class D2MGenericRewriter : public OpConversionPattern<d2m::GenericOp> {
...
+        argMapping[operandIndex] = args.size();
         args.push_back(operand);
       } else if (mlir::isa<ttmetal::LocalSemaphoreType>(operand.getType())) {
+        argMapping[operandIndex] = args.size();
         args.push_back(operand);
-      } else if (mlir::isa<MemRefType>(operand.getType())) {
+      } else if (auto memrefType =
+                     mlir::dyn_cast_if_present<MemRefType>(operand.getType());
+                 memrefType) {
+        // TODO: add back later
```
vmilosevic: :warning: **google\-readability\-todo** :warning:
missing username/bug in TODO
```suggestion
        // TODO(nsmith): add back later
```

lib/Conversion/D2MToTTMetal/D2MToTTMetal.cpp
```diff
@@ -131,57 +153,83 @@ class D2MGenericRewriter : public OpConversionPattern<d2m::GenericOp> {
...
   }
 
+  Value getUnderlyingMemref(Value operand) const {
+    if (auto view = mlir::dyn_cast_if_present<d2m::ViewLayoutOp>(
+            operand.getDefiningOp());
+        view) {
+      return view.getInput();
+    } else {
+      return operand;
+    }
```
vmilosevic: :warning: **llvm\-else\-after\-return** :warning:
```suggestion
    }       return operand;
   
```

vmilosevic (commented):

arminaleTT (approved):
Please also run nightly on this to stress TTNN JIT

lib/Dialect/D2M/Transforms/SplitUnifiedThread.cpp
```diff
@@ -57,643 +57,417 @@ static bool needsDMA(Value memref, Value localBuffer) {
...
+    auto isBlockingLoop = [](Operation *op) {
+      return mlir::isa<scf::ForOp>(op) && op->hasAttr("d2m.blocking_loop");
+    };
+    while (outermostOp->getParentOp() != genericOp.getOperation() &&
+           !isBlockingLoop(outermostOp->getParentOp())) {
+      outermostOp = outermostOp->getParentOp();
+      if (!mlir::isa<scf::ForOp>(outermostOp) &&
+          !mlir::isa<linalg::GenericOp>(outermostOp)) {
+        llvm::errs() << "op " << *outermostOp << "\n";
+        assert(false && "outermost loop op is not a scf.for op");
```
vtangTT: this should be a return failure instead no?

test/ttmlir/Dialect/D2M/Transforms/split_unified_thread.mlir
```diff
@@ -401,88 +427,36 @@ module attributes {ttcore.system_desc = #system_desc} {
...
       %core0 = d2m.core_index(0) : index
       %core1 = d2m.core_index(1) : index
-      %buf = memref.alloc() : memref<1x1x!ttcore.tile<32x32, f32>, #l1>
-      %0 = d2m.remote_load %buf %input[%core0, %core1] : memref<1x1x!ttcore.tile<32x32, f32>, #l1>, memref<4x3x1x1x!ttcore.tile<32x32, f32>, #ttcore.shard<4096x4096, 1>, #l1> -> memref<1x1x!ttcore.tile<32x32, f32>, #l1>
-      %1 = d2m.remote_store %output[%core0, %core1] %buf : memref<4x3x1x1x!ttcore.tile<32x32, f32>, #ttcore.shard<4096x4096, 1>, #l1>, memref<1x1x!ttcore.tile<32x32, f32>, #l1> -> memref<1x1x!ttcore.tile<32x32, f32>, #l1>
-    }
-    return
-  }
-
-  // Test 11: Self-read/write (read-modify-write on same buffer)
```
vtangTT, line 415: this case is deleted/no longer applies thanks to the operand_alias op right

lib/Conversion/D2MToTTMetal/D2MToTTMetal.cpp
```diff
@@ -278,25 +312,13 @@ class MemrefAllocRewriter : public OpConversionPattern<memref::AllocOp> {
     auto fwd = op->getAttrOfType<AffineMapAttr>(
         d2m::utils::kVirtualGridForwardMappingAttr);
 
+    assert((mlir::isa<ttcore::ShardLayoutAttr, ttcore::InterleavedLayoutAttr,
```
vtangTT: use TT_assert?

include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
```diff
@@ -3159,4 +3176,37 @@ def D2M_PrintOp : D2M_GenericRegionOp<"print", [MemoryEffects<[MemRead, MemWrite
     }];
 }
 
+def D2M_OperandAliasOp : D2M_GenericRegionOp<"operand_alias", [
```
vwellsTT, line 3179: one open question we had in our IR refactor group a couple weeks back is whether this op ought to just replace the alloc op, or whether it should replace alloc and the remote_load/store ops from the sharded tensor into the aliased tensor.  Does this version only do the former, or the latter?

lib/Conversion/D2MToTTMetal/D2MToTTMetal.cpp
```diff
@@ -26,6 +26,7 @@
 #include "llvm/ADT/STLExtras.h"
 #include "llvm/ADT/TypeSwitch.h"
 
+#include "mlir/IR/Builders.h"
```
vwellsTT: nit: wrong block

lib/Conversion/D2MToTTMetal/D2MToTTMetal.cpp
```diff
@@ -131,57 +153,69 @@ class D2MGenericRewriter : public OpConversionPattern<d2m::GenericOp> {
...
-                    : IntegerAttr()) {
-          unsigned idx = static_cast<unsigned>(cbForOp.getInt());
-          assert(idx < cbs.size() && "d2m.cb_for_operand out of range");
-          cbs[idx] = operand;
-        } else {
+        // MemrefAllocRewriter).
+        if (auto aliasOp = mlir::dyn_cast<d2m::OperandAliasOp>(
+                op.getOperands()[operandIndex].getDefiningOp())) {
+          // OperandAliasOp's input is the generic's operand that this CB
+          // aliases. It could be a function argument of the parent func or an
```
vwellsTT: In practice shouldn't sharded allocOp's always be passed as function arguments (s.t. it could NOT really be an allocOp)?

lib/Dialect/D2M/Transforms/Allocate.cpp
```diff
@@ -1364,6 +1364,21 @@ class D2MAllocate final : public impl::D2MAllocateBase<D2MAllocate> {
                   L1memInfo, analysis.sequencing))) {
             return failure();
           }
+        } else {
+          // Mark allocs as aliased for HoistCBAllocs pass
```
vwellsTT: nit: .

lib/Dialect/D2M/Transforms/HoistCBAllocs.cpp
```diff
@@ -93,59 +81,20 @@ class D2MHoistCBAllocs : public impl::D2MHoistCBAllocsBase<D2MHoistCBAllocs> {
...
 
-  /// Copy VGM attrs from the operand's defining alloc onto |externalAlloc|.
-  static void copyVGMFromOperand(d2m::GenericOp genericOp, int64_t operandIdx,
-                                 memref::AllocOp externalAlloc) {
-    Value externalOperand = genericOp.getInputsAndOutputs()[operandIdx];
-    auto copyAttrs = [&](Operation *src) {
-      if (auto vgm = src->getAttrOfType<AffineMapAttr>(
-              d2m::utils::kVirtualGridInverseMappingAttr)) {
-        externalAlloc->setAttr(d2m::utils::kVirtualGridInverseMappingAttr, vgm);
+    // Hoist aliased CBs
```
vwellsTT: nit: sentence

lib/Dialect/D2M/Transforms/HoistCBAllocs.cpp
```diff
@@ -93,59 +81,20 @@ class D2MHoistCBAllocs : public impl::D2MHoistCBAllocsBase<D2MHoistCBAllocs> {
...
       }
     }
-  }
-
-  /// Find the regular operand index for a CB alloc by tracing through
-  /// its remote_load/remote_store memref operand back to the generic's
-  /// operands.
-  /// In the case that a remote load and store both use the same AllocOp,
-  /// consider the allocation as corresponding to the _input_ operand.
-  static int64_t findRegularOperandIndex(d2m::GenericOp genericOp,
```
vwellsTT, line 103: great cleanup :)

lib/Dialect/D2M/Transforms/SplitUnifiedThread.cpp
```diff
@@ -2,13 +2,16 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 
+#include "ttmlir/Dialect/D2M/IR/D2MTraits.h"
```
vwellsTT: nit: I think this belongs in block below

lib/Dialect/D2M/Transforms/SplitUnifiedThread.cpp
```diff
@@ -57,643 +57,417 @@ static bool needsDMA(Value memref, Value localBuffer) {
...
+    Operation *outermostOp = op;
+    auto isBlockingLoop = [](Operation *op) {
+      return mlir::isa<scf::ForOp>(op) && op->hasAttr("d2m.blocking_loop");
+    };
+    while (outermostOp->getParentOp() != genericOp.getOperation() &&
+           !isBlockingLoop(outermostOp->getParentOp())) {
+      outermostOp = outermostOp->getParentOp();
+      if (!mlir::isa<scf::ForOp>(outermostOp) &&
+          !mlir::isa<linalg::GenericOp>(outermostOp)) {
+        llvm::errs() << "op " << *outermostOp << "\n";
```
vwellsTT: probably best to do what copilot says (avoid direct errs() stream + assert(false) in favor of emitOpError or similar) I think

lib/Dialect/D2M/Transforms/SplitUnifiedThread.cpp
```diff
@@ -57,643 +57,417 @@ static bool needsDMA(Value memref, Value localBuffer) {
...
-            "could not find associated CB for shared pair");
-      }
-      rewriter.setInsertionPoint(storeOp);
+    if (mlir::isa<RemoteLoadOp>(producer) &&
+        mlir::isa<RemoteStoreOp>(consumer) &&
+        isAliasedStore(mlir::cast<RemoteStoreOp>(consumer))) {
+      Location loc = producer->getLoc();
+      unsigned cbOperandIdx =
+          producer->getParentOfType<GenericOp>().getOperandIndex(localBuffer);
+      // Set insertion point before consumer so GetCBOp dominates WaitOp/PopOp
```
vwellsTT: nit: .

lib/Dialect/D2M/Transforms/SplitUnifiedThread.cpp
```diff
@@ -57,643 +57,417 @@ static bool needsDMA(Value memref, Value localBuffer) {
...
-      }
-      rewriter.setInsertionPoint(loadOp);
+    } else if (mlir::isa<RemoteLoadOp>(producer) &&
+               mlir::isa<RemoteStoreOp>(consumer) &&
+               isAliasedLoad(mlir::cast<RemoteLoadOp>(producer))) {
+      Location loc = consumer->getLoc();
+      unsigned cbOperandIdx =
+          consumer->getParentOfType<GenericOp>().getOperandIndex(localBuffer);
+      // Set insertion point before producer so GetCBOp dominates
+      // ReserveOp/PushOp
```
vwellsTT: nit: .

lib/Dialect/D2M/Transforms/SplitUnifiedThread.cpp
```diff
@@ -57,643 +57,417 @@ static bool needsDMA(Value memref, Value localBuffer) {
...
-
-      Region *computeRegion = loadOp->getParentRegion();
-      rewriter.replaceUsesWithIf(
-          localBuffer, waitOp.getResult(), [&](OpOperand &use) {
-            return computeRegion->isAncestor(use.getOwner()->getParentRegion());
-          });
-      if (loadOp.getResult()) {
-        rewriter.replaceAllUsesWith(loadOp.getResult(), waitOp.getResult());
-      }
+      // get producers and insert reserve+push
```
vwellsTT: nit: sentence

lib/Dialect/D2M/Transforms/SplitUnifiedThread.cpp
```diff
@@ -57,643 +57,417 @@ static bool needsDMA(Value memref, Value localBuffer) {
...
+      for (auto &operand : synchronizedOp->getOpOperands()) {
+        if (synchronizedOp.isProducer(operand)) {
+          Location loc = synchronizedOp.getLoc();
+          Value localBuffer = operand.get();
+          unsigned cbOperandIdx =
+              synchronizedOp->getParentOfType<GenericOp>().getOperandIndex(
+                  localBuffer);
+
+          // get the associated consumer for this operand
```
vwellsTT: nit: sentence

lib/Target/TTMetal/TTMetalToFlatbuffer.cpp
```diff
@@ -1171,6 +1171,11 @@ std::shared_ptr<void> translateTTMetalToFlatbuffer(
         // Unqualified walk will visit the root op itself last, we should
         // ignore this.
         return;
+      } else if (auto operandAliasOp =
```
vwellsTT: I'm not 100% sure, but I loosely feel like we shouldn't preserve ops which don't have meaning by this stage.  Maybe TTMetal shouldn't have the aliasOp of its own, and D2MToTTMetal should be the end of the alias op?

lib/Dialect/D2M/Transforms/SplitUnifiedThread.cpp
```diff
@@ -57,643 +57,417 @@ static bool needsDMA(Value memref, Value localBuffer) {
...
-      toErase.insert(loadOp);
-      continue;
-    }
+// From cb usage info, check for load-store pairs and insert aliased cb ops for
+// alias side.
+static LogicalResult processSharedBufferPairs(
+    Block *computeBlock, PatternRewriter &rewriter,
+    llvm::DenseMap<Value, utils::CBUsageInfo> &cbUsageInfo) {
+  for (auto [localBuffer, usageInfo] : cbUsageInfo) {
+    auto *producer = usageInfo.producers.front();
```
vtangTT, line 257: can we guard these `.front()` calls in the case that these vectors are emtpy

lib/Dialect/D2M/Transforms/SplitUnifiedThread.cpp
```diff
@@ -57,643 +57,417 @@ static bool needsDMA(Value memref, Value localBuffer) {
...
-      }
+// Erase aliased load and store ops (no DMA needed).
+static LogicalResult eraseAliasedLoadStoreOps(
+    PatternRewriter &rewriter,
+    llvm::DenseMap<Value, utils::CBUsageInfo> &cbUsageInfo) {
+  for (auto [localBuffer, usageInfo] : cbUsageInfo) {
+    auto *producer = usageInfo.producers.front();
+    auto *consumer = usageInfo.consumers.front();
+
+    if (mlir::isa<RemoteStoreOp>(consumer) &&
```
vtangTT, line 397: can you clarify why the case of both load and store being aliased isn't possible here? in this case, we would only erase the store which is incorrect?

if the both aliased case is not possible can we add an assertion or some guard just for safety

test/ttmlir/Dialect/D2M/Transforms/split_unified_thread.mlir
```diff
@@ -401,88 +427,36 @@ module attributes {ttcore.system_desc = #system_desc} {
...
-  // Verifies: no remote ops in DMA, compute has reserve+push+wait for the
-  // accumulator CB, no extra push/wait/pop from store
-  // CHECK-LABEL: func.func @test_self_read_write_accumulator
-  func.func @test_self_read_write_accumulator() {
```
vtangTT, line 419: can you clarify why this doesn't need to be tested anymore?

include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
```diff
@@ -1468,8 +1468,8 @@ def D2M_LocalCopyOp : D2M_GenericRegionDatamovementOp<"local_copy",
 
       Explicit CB form (after LowerLoadStoreOpsToExplicitCBForm):
```
vtangTT: `LowerLoadStoreOpsToExplicitCBForm` doesn't exist anymore oops, would be great if you fixed this

vtangTT (commented):
local_copy changes make sense to me 👍 

alias operand lowering makes sense to me

xanderchin:
## Claude Review feedback

Reviewing the SynchronizableOpInterface application to production. The architectural refactoring is sound and most issues from the rejected PR #8026 are resolved. Four issues to consider, scoped to medium severity and higher (BUG numbering is internal — not the GitHub issue tracker):

### BUG-1 (High): `llvm_unreachable` for runtime-checked precondition

**File:** `lib/Dialect/D2M/Utils/SynchronizableOpInterfaceUtils.cpp` lines 84–90

```cpp
llvm::errs()
    << "Found use of non-pure op result outside the range of ops "
       "being wrapped: \n"
    << *user;
llvm_unreachable("Results of ops being wrapped in "
                 "SynchronizedRegionOp cannot be "
                 "used outside the range.");
```

This is a precondition check on caller-supplied input (the `[start, end)` range), not a static invariant. `llvm_unreachable` becomes `__builtin_unreachable` — undefined behavior — in release builds: the `llvm::errs()` will land in the log, but execution then enters UB rather than aborting cleanly.

For runtime-checked preconditions, use `llvm::report_fatal_error`, which always aborts cleanly:

```cpp
std::string msg;
llvm::raw_string_ostream os(msg);
os << "Found use of non-pure op result outside the range of ops "
      "being wrapped: " << *user;
llvm::report_fatal_error(llvm::StringRef(os.str()));
```

Flagged in the PR #8106 review (NEW-ISSUE-1) and not fixed there either.

### BUG-2 (Medium): `llvm_unreachable` for interface invariant

**File:** `lib/Dialect/D2M/Utils/SynchronizableOpInterfaceUtils.cpp` line 28

```cpp
if (synchronizedOp.isProducer(operand) &&
    synchronizedOp.isConsumer(operand)) {
  llvm_unreachable(
      "A single op operand cannot be both a producer and consumer");
}
```

This validates an interface invariant that may be violated by external implementations (notably `linalg::GenericOp` via external interface registration). Same UB-in-release problem as BUG-2; same fix:

```cpp
llvm::report_fatal_error(
    "A single op operand cannot be both a producer and consumer");
```

Flagged in the PR #8106 review (ISSUE-1) and not fixed there.

### BUG-3 (Medium): Copyright typo `2026wr`

**File:** `lib/Dialect/D2M/Transforms/HoistCBAllocs.cpp` line 1

```cpp
// SPDX-FileCopyrightText: (c) 2026wr Tenstorrent AI ULC
```

Stray `wr` after the year. Inherited from the rejected PR #8026; never fixed.

### BUG-4 (Medium): `D2M_SynchronizedRegionOp` has no verifier

**File:** `include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td` (around line 2198)

The op has variadic `consumers` + `producers`, a `SizedRegion<1>`, and `NoTerminator`/`SingleBlock` traits, but none validate:
- block argument count matches `consumers.size() + producers.size()`
- operand types match block argument types
- no operand appears in both `consumers` and `producers`

Currently safe because `wrapInSynchronizedRegion` constructs these carefully, but invalid IR can pass `mlir-opt -verify-each` cleanly. Carried over from PR #8106 (ISSUE-2).

lib/Dialect/D2M/Transforms/InsertDstRegisterAccess.cpp
```diff
@@ -669,9 +669,39 @@ struct D2MInsertDstRegisterAccessRewriter final
...
+        if (user->getBlock() != opBlock) {
+          continue;
+        }
+        if (!firstUser || user->isBeforeInBlock(firstUser)) {
+          firstUser = user;
+        }
+      }
+    }
+    return firstUser;
+  }
```
sgholamiTT: Can we promote this to `include/ttmlir/Utils.h` as this seems like a good utility function that can benefit other places (for instance, [this code](https://github.com/tenstorrent/tt-mlir/blob/46dedd805f63dfacb6dc26b920808672f7334349/lib/Transforms/ConstEvalHoist.cpp#L366-L380) is pretty much doing the same thing and could benefit from this function, though I'm not asking you to modify it, just flagging a potential user of this function).

lib/Dialect/D2M/Transforms/InsertDstRegisterAccess.cpp
```diff
@@ -669,9 +669,39 @@ struct D2MInsertDstRegisterAccessRewriter final
...
+    Operation *firstUser = nullptr;
+    Block *opBlock = op->getBlock();
+    for (Value result : op->getResults()) {
+      for (Operation *user : result.getUsers()) {
+        if (user->getBlock() != opBlock) {
+          continue;
+        }
+        if (!firstUser || user->isBeforeInBlock(firstUser)) {
+          firstUser = user;
+        }
```
sgholamiTT: What do you think of:
```suggestion
      mlir::Operation *userInBlock = block->findAncestorOpInBlock(*user);
      if (!userInBlock) {
        continue;
      }
      if (!firstUser || userInBlock->isBeforeInBlock(firstUser)) {
        firstUser = userInBlock;
      }
```

---------
Motivation: Consider this IR:
```mlir
func.func @example(%arg0: tensor<4xf32>, %cond: i1) -> tensor<4xf32> {
  %0 = ttir.add %arg0, %arg0 : tensor<4xf32>     // <-- op of interest
  scf.if %cond {                                 // (A) lives in op's block
    "some.consumer"(%0) : (tensor<4xf32>) -> ()  // direct user of %0, but its getBlock() is the scf.if region's block
     scf.yield
  }
  %1 = ttir.exp %0 : tensor<4xf32>               // (B) direct user in op's block
  return %1 : tensor<4xf32>
}
```

These two versions return different answers for `op = %0 = ttir.add`. Yours return `(B)` and the proposed version returns `(A)`. Though I'm not sure if this is particularly possible when looking for `aquire_dst` first user. 

Just thought this might be worth bringing up, but feel free to ignore this comment if it doesn't apply to your case.

lib/Dialect/D2M/Transforms/InsertDstRegisterAccess.cpp
```diff
@@ -669,9 +669,39 @@ struct D2MInsertDstRegisterAccessRewriter final
...
+    // remote_load ops. Move it to just before its first use to ensure compute
+    // ops are contiguous (important for SplitUnifiedThread which separates out
+    // contiguous regions of compute ops into synchronized regions).
+    // When there IS an outermost compute loop, the insertion point was already
+    // set correctly (inside or before the loop body), so no move is needed.
+    if (!outermostInnerComputeLoop) {
+      if (Operation *firstUser = findFirstUserInBlock(acquireDst)) {
+        acquireDst->moveBefore(firstUser);
+      }
+    }
```
sgholamiTT, line 684: Can we add a lit test that exercises this code path and fails if we remove this ?

I'm asking because I have a [refactoring PR](https://github.com/tenstorrent/tt-mlir/pull/7919) that targets this file, and I want to make sure that I preserve this behavior (since you'd probably merge your PR sooner than mine).

sgholamiTT (approved):
`InsertDstRegisterAccess` changes LGTM. 👍🏻 
Few nits inlined.

include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
sohaibnadeemTT: This version does the former, and only erases the aliased load/store in SplitUnifiedThread once cb ops (wait/reserve/push/pop) have been inserted. This is simpler for SplitUnifiedThread since we can still easily establish producer/consumer relationships in SplitUnifiedThread using getCBUsageInfo (erasing earlier would wouldn't allow us to use getCBUsageInfo), detect which consumers have an aliased load, and move the cb reserve+push onto the consumer before erasing the aliased load (similarly for aliased store).

lib/Conversion/D2MToTTMetal/D2MToTTMetal.cpp
sohaibnadeemTT: Oh I meant the parent func to the the generic here

test/ttmlir/Dialect/D2M/Transforms/split_unified_thread.mlir
sohaibnadeemTT: This isn't really a valid test case. In theory, yes you can do the compute in-place but we need to account for parallelization of the compute and DM threads here. Both the remote_load and compute are producers for %acc_buf therefore both will be pushing onto the CB which doesn't make sense and will cause undefined-behaviour.

test/ttmlir/Dialect/D2M/Transforms/split_unified_thread.mlir
sohaibnadeemTT: I guess this is the same question as below?

lib/Dialect/D2M/Transforms/SplitUnifiedThread.cpp
sohaibnadeemTT: If we have a load -> local buf -> store, and both the load and the store can be aliased, we cannot actually erase both ops because we still have to move the the data. In other word, we can only alias one side in general.

In the code above, we first check if consumer is an alias-able remote_store and if it's not then we check if the producer is an alias-able remote_load. If both are alias-able, we would pass the first check so only erase the store.

xanderchin (approved):

lib/Conversion/D2MToTTMetal/D2MToTTMetal.cpp
sohaibnadeemTT: fixed

lib/Dialect/D2M/Transforms/Allocate.cpp
sohaibnadeemTT: fixed

lib/Dialect/D2M/Transforms/HoistCBAllocs.cpp
sohaibnadeemTT: fixed

lib/Dialect/D2M/Transforms/SplitUnifiedThread.cpp
sohaibnadeemTT: fixed

lib/Dialect/D2M/Transforms/SplitUnifiedThread.cpp
sohaibnadeemTT: fixed

lib/Dialect/D2M/Transforms/SplitUnifiedThread.cpp
sohaibnadeemTT: fixed

lib/Dialect/D2M/Transforms/SplitUnifiedThread.cpp
sohaibnadeemTT: fixed

lib/Dialect/D2M/Transforms/SplitUnifiedThread.cpp
sohaibnadeemTT: fixed

lib/Target/TTMetal/TTMetalToFlatbuffer.cpp
sohaibnadeemTT: yeah fair point, fixed

lib/Dialect/D2M/Transforms/SplitUnifiedThread.cpp
sohaibnadeemTT: fixed to use emitOpError and  return failure

lib/Conversion/D2MToTTMetal/D2MToTTMetal.cpp
sohaibnadeemTT: fixed

include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
sohaibnadeemTT: fixed to SplitUnifiedThread

lib/Dialect/D2M/Transforms/SplitUnifiedThread.cpp
sohaibnadeemTT: Just to clarify, it's technically possibly for both load and store to be aliasable right now based on how we are checking here but we can only actually alias one side. In the following refactor for allocator, the aliasing decision logic will be changed to only be in allocator so we can add a guard then.

lib/Dialect/D2M/Transforms/SplitUnifiedThread.cpp
sohaibnadeemTT: We should probably assert here since it must only be one consumer and producer (can change in the future with DFBs but we don’t handle that now so shouldn’t allow it). Will add it after current CI run is green.

vwellsTT (approved):

lib/Dialect/D2M/Transforms/SplitUnifiedThread.cpp
vtangTT: Ok, yea this makes sense. 
```
In the following refactor for allocator, the aliasing decision logic will be changed to only be in allocator so we can add a guard then
```

yea this is def the way to go 👍

vtangTT (approved):
lgtm ty 👍 please add the assert for 1 consumer/producer

lib/Dialect/D2M/Transforms/InsertDstRegisterAccess.cpp
sohaibnadeemTT: moved

lib/Dialect/D2M/Transforms/InsertDstRegisterAccess.cpp
sohaibnadeemTT: Not relevant to my use case (since it's meant to be used  to find first use of aquire_dst when loops are canonicalized) but yeah your suggestion is more correct. Fixed.

lib/Dialect/D2M/Transforms/SplitUnifiedThread.cpp
sohaibnadeemTT: added

lib/Dialect/D2M/Transforms/InsertDstRegisterAccess.cpp
sohaibnadeemTT: Added `test/ttmlir/Dialect/D2M/Transforms/insert_dst_register_access_no_loop.mlir`, but note this will also be exercised in e2e tests (test/python/golden/d2m/test_tms.py::test_arange)

include/ttmlir/Dialect/D2M/Analysis/Allocation/Utils.h
```diff
@@ -408,8 +408,7 @@ inline bool hasNonTrivialView(Value value) {
            hasNonTrivialView(viewOp.getInput());
   }
   if (auto compositeViewOp = mlir::dyn_cast<d2m::CompositeViewOp>(definingOp)) {
-    return llvm::any_of(compositeViewOp.getCompositeInputs(),
-                        hasNonTrivialView);
+    return true;
```
anuragsingh-tt, line 411: This is probably correct behavior. Can we add a lit test? E.g. allocate_composite_view_forces_stream.mlir which checks that we force a stream.

lib/Dialect/D2M/Transforms/Allocate.cpp
```diff
@@ -1364,6 +1364,27 @@ class D2MAllocate final : public impl::D2MAllocateBase<D2MAllocate> {
                   L1memInfo, analysis.sequencing))) {
             return failure();
           }
+        } else {
```
anuragsingh-tt, line 1367: I am wondering what happens if the remote_load/store are themselves using aliases. findLocalBufferForOperandLoadStore early exits when it reaches these ops. 

Side note: I am not sure this HoistCBAlloc prep should be in insertOperandStreams, or in Allocate at all. But I get that it is convenient for now, we can tackle it later. It should probably move to HoistCBAllocs.

lib/Dialect/D2M/Transforms/SplitUnifiedThread.cpp
```diff
@@ -25,9 +28,6 @@ namespace {
 // Helpers
 // ---------------------------------------------------------------------------
 
-// Returns true if the remote_load/store requires real DMA. This is the case
```
anuragsingh-tt, line 28: I like this comment, it should be updated rather than deleted.

sohaibnadeemTT:
@anuragsingh-tt The surface area of this PR is already pretty big and conflicts with several PRs in-flight, so going to address your comments in a follow-up PR for the sake of efficiency.

PR #8137, opened 2026-04-27
[d2m] exclude scratch from GenericOp::getOperandAlloc

ckaravasilisTT:
**`GenericOp::getOperandAlloc:`** skip `memref.alloc` ops consumed by `d2m.scratch_init` so scratch buffers are not mis-matched as operand allocs (and later retyped by D2MAllocate).

Enabled previously-failing fusion test.


anuragsingh-tt (approved):
Let's not reblock scratch.

Issue #8141, opened 2026-04-27
TTNNWorkarounds skips TopKOp at opt_level >= 1, producing f32 indices typecasts and silently wrong outputs


kmabeeTT:
## Summary

When `optimizer-enabled=true`, `TTNNWorkarounds` restricts itself to a small whitelist of ops, and `TopKOp` is not in it. As a result, `createTopKOpOperandsWorkarounds` (which forces input bf16 + output indices ui16/ui32 + a `to_layout` round-trip back to the originally-requested integer dtype) is skipped at `opt_level >= 1`. The optimizer's own dtype propagation then fills in for the topk indices output, picking `f32` in some graph patterns. Downstream ops that consume those indices end up with mismatched operand types — silently wrong integer values at runtime.

## Repro

In tt-xla's vLLM sampler with ~3x perf-improvement changes via https://github.com/tenstorrent/tt-xla/pull/4334 (`integrations/vllm_plugin/vllm_tt/sampler.py::apply_top_k_top_p_fast`), four sequential `torch.topk` calls feed a per-chunk offset add → concat → gather. At `opt_level=1`, the resulting TTNN IR contains, for 3 of 4 chunks:

```mlir
%38 = "ttnn.typecast"(%indices_1) : (tensor<2x32xui16>) -> tensor<2x32xf32>     # should be si32
...
%55 = "ttnn.add"(%38, %1) <{dtype = si32}>
    : (tensor<2x32xf32>, tensor<1x1xsi32>) -> tensor<2x32xsi32>                  # mixed operand types
```

The chunk-offset add takes an `f32` operand under an `si32` declared dtype. End-to-end symptom: garbage tokens during non-greedy device sampling on Llama (e.g. `'! and!! the!!! river!!!!'` repetition patterns). At `opt_level=0` the workaround applies normally and all four chunks' indices uniformly typecast to `si32`; output is coherent.

## Expected

Either the optimizer-restricted whitelist should include `TopKOp`, or the `ttnn.add` verifier should reject mismatched operand types so this category of bug surfaces at compile time rather than runtime.

## Fix

I have a PR, it adds `ttnn::TopKOp::getOperationName()` to `enabledOpsForWorkaroundWithOptimizer` and a regression lit test that fails without the change and passes with it, will post.

Issue #8134, opened 2026-04-27
TopK workaround skipped under enable-optimizer; F32 input trips fallback cascade with no recovery via reshards


rpavlovicTT:
### Summary

`TopKOp`'s dtype workaround (which inserts an `f32 → bf16` typecast on its input) is **not in the optimizer's workaround whitelist**, so under `enable-optimizer=true` topk runs with whatever dtype the upstream produces. tt-metal's topk kernel only supports `bfloat16` / `bfloat8_b` input (see `topk_device_operation.cpp:145-148`), so an `f32` input always fails validation. The optimizer cannot recover, and the resulting fallback cascade silently produces a bad lowering downstream.

### Where

`lib/Dialect/TTNN/Transforms/Workarounds/TTNNWorkaroundsPatterns.cpp`, around line 668:

```cpp
const std::set<mlir::StringRef>
    TTNNWorkarounds::TTNNWorkarounds::enabledOpsForWorkaroundWithOptimizer = {
        ttnn::WhereOp::getOperationName(), ttnn::FullOp::getOperationName(),
        ttnn::EmbeddingOp::getOperationName(),
        ttnn::ScatterOp::getOperationName()};
```

The workaround pass branches on `optimizerEnabled`: if true, only this whitelist runs; if false, **all** TTNN ops run their workarounds. The TopK dtype workaround (`createTopKOpOperandsWorkarounds` in `lib/Dialect/TTNN/IR/TTNNWorkaroundsPass.cpp:1185`) sets `tensorDataTypeWorkaround = BFloat16` for non-bf16/bf8 inputs — but it is never invoked under the optimizer because `TopKOp` isn't whitelisted.

### Why the optimizer can't recover on its own

1. **Reshard candidates are dtype-preserving.** `MemoryLayoutPropagation::generateReshardCandidates` (`MemoryLayoutPropagation.cpp:951`) keys on the producer's `scalarElementType` and only enumerates layout/grid/sharding variants for that exact dtype. There is no candidate path that converts f32 → bf16, so the beam search has no valid configuration for topk when the producer outputs f32.

[...]

PR #8142, opened 2026-04-27
[Workarounds] Add TopKOp to enabledOpsForWorkaroundWithOptimizer

kmabeeTT:
### Ticket

[tt-mlir #8141](https://github.com/tenstorrent/tt-mlir/issues/8141)

### Problem description

- `TTNNWorkarounds` restricts itself to a small whitelist of ops when `optimizer-enabled=true`. `TopKOp` was missing, so `createTopKOpOperandsWorkarounds` (which forces input bf16 + output indices ui16/ui32 + the `to_layout` round-trip back to the originally-requested integer dtype) was skipped at `opt_level >= 1`. Without it, the optimizer's own dtype propagation could pick f32 for the topk indices output in some graph patterns, producing downstream `ttnn.add` ops with mismatched `(f32, si32)` operands and silently wrong integer values.

- Found via tt-xla's chunked-topk vLLM sampler (4 sequential topk calls → per-chunk offset add → concat → gather). At `opt_level=1`, 3 of 4 chunks got `f32` indices typecasts instead of `si32`, producing garbage tokens during non-greedy device sampling on Llama.

### What's changed

- Add `ttnn::TopKOp::getOperationName()` to `enabledOpsForWorkaroundWithOptimizer` in `lib/Dialect/TTNN/Transforms/Workarounds/TTNNWorkaroundsPatterns.cpp`. The existing `createTopKOpOperandsWorkarounds` then applies at `opt >= 1` the same way it does at `opt = 0`.

- Also adds `test/ttmlir/Dialect/TTNN/Transforms/Workarounds/topk_workaround_with_optimizer.mlir` — mirrors the existing `topk_workaround.mlir` and `topk_f32_input_workaround.mlir` but invokes the workaround pass with `ttnn-is-optimizer-enabled=true`. Three sub-cases: vocab ≤ 65535 (`ui16` indices), vocab > 65535 (`ui32` indices), and f32 input requiring bf16 conversion. All three fail without this commit (raw types from `ttir.topk` pass through `ttnn.topk` unchanged) and pass with it.

- Updates `test/ttmlir/Dialect/TTNN/optimizer/ttnn_fusing/topk/topk_ttnn_fusing_pass.mlir` 3D layouts from the artificial `(d0 * 3 + d1, d2)` 1x1-tile encoding to the realistic `(d0 * 32 + d1, d2)` 2x1-tile encoding that `--ttnn-layout` actually produces for `tensor<2x3x{8,4}x{f32,si32}>`. The fusion validator runs the topk workaround on the candidate fused op; with the artificial encoding, `TTNNLayoutAttr::withElementType` recreates the output layout from defaults, mismatches the function-declared return type, and the verifier rejects the fusion. CHECK lines for the 3D case are unchanged.

### Checklist

- [x] End-to-end impact on tt-xla's vLLM sampler at `opt_level=1`: garbage tokens become coherent text, and Llama-3.1-8B b=2 nongreedy-device throughput rises from 4.83 tok/s (broken pre-PR baseline) to 12.49 tok/s (2.59×). Llama-3.2-1B b=2 nongreedy-device: 5.81 → 21.41 tok/s (3.69×). Greedy paths and CPU sampling paths are unaffected (no regression).

### DIFF
--- a/lib/Dialect/TTNN/Transforms/Workarounds/TTNNWorkaroundsPatterns.cpp
+++ b/lib/Dialect/TTNN/Transforms/Workarounds/TTNNWorkaroundsPatterns.cpp
@@ -704,5 +704,8 @@ const std::set<mlir::StringRef>
     TTNNWorkarounds::TTNNWorkarounds::enabledOpsForWorkaroundWithOptimizer = {
         ttnn::WhereOp::getOperationName(), ttnn::FullOp::getOperationName(),
         ttnn::EmbeddingOp::getOperationName(),
-        ttnn::ScatterOp::getOperationName()};
+        ttnn::ScatterOp::getOperationName(),
+        // TopK's operands workaround forces input bf16 + indices ui16/ui32;
+        // without it, opt_level>=1 dtype propagation picks f32. See #8141.
+        ttnn::TopKOp::getOperationName()};
 } // namespace mlir::tt::ttnn
--- a/test/ttmlir/Dialect/TTNN/Transforms/Workarounds/topk_workaround_with_optimizer.mlir
+++ b/test/ttmlir/Dialect/TTNN/Transforms/Workarounds/topk_workaround_with_optimizer.mlir
@@ -1,0 +1,89 @@
+// RUN: ttmlir-opt --split-input-file --ttcore-register-device --ttnn-layout --convert-ttir-to-ttnn --ttnn-workaround="ttnn-is-optimizer-enabled=true" --canonicalize -o %t %s
+// RUN: FileCheck %s --input-file=%t
+
+// Regression test: TopKOp must remain in the workaround whitelist when the
+// optimizer is enabled.
+//
+// The TTNN workarounds pass restricts itself to a small set of ops when
+// `optimizer-enabled=true` (chain-based or greedy optimizer paths at
+// opt_level >= 1). TopKOp was previously absent from that set, so its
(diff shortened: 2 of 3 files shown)


rpavlovicTT (approved):
I opened issue for the same problem I had found yesterday https://github.com/tenstorrent/tt-mlir/issues/8134. I actually hit a hang due to consequence of top-K F32 dtype. Short-term we need to allow this workaround, I will work on more general/long-term solution.

sdjordjevicTT (approved):

kmabeeTT:
Circling back here - Rebased on latest and pushed a change to fix failing [CI](https://github.com/tenstorrent/tt-mlir/pull/8142/checks?check_run_id=73295370126) test: the existing 3D test case in `topk_ttnn_fusing_pass.mlir` used a hand-crafted compact 1x1-tile encoding `(d0 * 3 + d1, d2)` that doesn't match what `--ttnn-layout` produces for that shape. Adding TopKOp to the optimizer-enabled workaround whitelist surfaced this — the fusion validator now runs the topk workaround during validation, and `TTNNLayoutAttr::withElementType` rebuilds the output layout from defaults, producing `(d0 * 32 + d1, d2)` 2x1-tile and tripping the function-return-type verifier on the artificial encoding. Updated the four 3D layouts to the realistic encoding; CHECK lines didn't need to change. Verified locally — the 3D case fuses again and `topk_workaround.mlir` / `topk_workaround_with_optimizer.mlir` / `topk_full_pipeline.mlir` all still pass.

Issue #8120, opened 2026-04-27
d2m test failing on main


dlokeTT:
### Describe the bug

`test/python/golden/d2m/test_binary_tree.py::test_binary_tree[ttmetal-bf16-2048x2048]` is failing on main on p150 after cf42a9b982edb4ae9774b535a5de18dddfa5013b and 27710a0e02edf7adc523557c3b0a540e52b2edc3 due to stale merge. I tried to repro on p100 and couldn't reserve a p150. Skipping this until someone can investigate it

sgholamiTT:
`test_binary_tree[ttmetal-f32-2048x2048]` also fails on `n150` (see the [CI log](https://github.com/tenstorrent/tt-mlir/actions/runs/25010677784/job/73248864109?pr=8116#step:16:7395)).
To unblcok the [uplift](https://github.com/tenstorrent/tt-mlir/pull/8116), I [skipped](https://github.com/tenstorrent/tt-mlir/pull/8116/changes/4447b5f300e51ce661abeee9b5a5e63646ffac46) that case too which should be re-enabled when the fix is in place.

PR #8143, opened 2026-04-27
d2m: fuse sibling scratch loops in InsertScratch

anuragsingh-tt:
### Ticket
Solves #8120 

### Problem description
Encountered a 10x13 Blackhole regression when running fused D2M binary-tree pytests.

The original scratch spill path could produce scratch allocations that were too large for the available scratch buffer. The fix fused sibling scratch loops so producer and consumer work shared the same outer iteration space which exposed two additional issues:
- stale unpack_stall_on_pack ops could be carried over from the old unfused loop layout
- consumer scratch loads could miss prefix loop dimensions after fusion, producing rank-mismatched affine loads.

On simulator, the fused shape also exposed that experimental::unpack_stall_on_pack() was using PACK_DONE without initializing it first, causing a semaphore runtime error. 

### What's changed
- Fuse sibling scratch loops that are part of the same producer-to-consumer chain. 

```
Before:
scf.for { scratch_space_loop_1 { producer } }
scf.for { scratch_space_loop_2 { consumer } }

After:
scf.for {
    scratch_space_loop_1 { producer }
    scratch_space_loop_2 { consumer }
}
```
- Initialize PACK_DONE before posting in unpack_stall_on_pack. This is probably not the right ownership model and should be revisited. Issue here: #8144 
- With these new changes the pass is getting a bit large and should be split up into two separate passes. Issue here: #8145 
- New lit test added. bf16 p150 pytest restored.

### DIFF
--- a/lib/Dialect/D2M/Transforms/InsertSpillAndScratch.cpp
+++ b/lib/Dialect/D2M/Transforms/InsertSpillAndScratch.cpp
@@ -403,7 +403,8 @@ getScratchAccessMapAndOperands(ScratchLoopInfo &info, MLIRContext *ctx) {
 static SmallVector<affine::AffineForOp>
 getMissingConsumerPrefixLoops(ScratchLoopInfo &producer,
                               ScratchLoopInfo &consumer) {
-  // Find the shared outer loops the consumer needs to line up with producer scratch.
+  // Find the shared outer loops the consumer needs to line up with producer
+  // scratch.
   size_t consumerRank = 1 + consumer.allLoops.size();
   if (producer.scratchShape.size() <= consumerRank) {
     return {};
@@ -454,7 +455,8 @@ getConsumerScratchAccessMap(ScratchLoopInfo &producer,
     SmallVector<Value> operands;
     unsigned dimIdx = 0;
 
-    // Build the straightforward consumer index including any shared prefix loops.
+    // Build the straightforward consumer index including any shared prefix
+    // loops.
     auto addLoop = [&](affine::AffineForOp loop) {
       operands.push_back(loop.getInductionVar());
       AffineExpr dim = getAffineDimExpr(dimIdx++, ctx);
@@ -953,7 +955,8 @@ fuseScratchLoopSiblings(GenericOp genericOp,
 
     size_t bestBegin = scratchSpaceLoops.size();
     size_t bestEnd = 0;
-    // Pick the widest producer->consumer span so one fusion covers the whole chain.
+    // Pick the widest producer->consumer span so one fusion covers the whole
+    // chain.
     for (auto &intermediate : intermediates) {
       size_t producerIdx =
           static_cast<size_t>(intermediate.producer - loopShells.data());
--- a/lib/Dialect/D2M/Transforms/InsertSpillAndScratch.cpp
+++ b/lib/Dialect/D2M/Transforms/InsertSpillAndScratch.cpp
@@ -92,8 +92,7 @@ static int64_t estimateScratchDemand(ArrayRef<IntermediateAllocInfo> infos) {
     if (!info.producer) {
       continue;
     }
-    if (llvm::is_contained(info.producer->scratchShape,
-                           ShapedType::kDynamic)) {
+    if (llvm::is_contained(info.producer->scratchShape, ShapedType::kDynamic)) {
       return ShapedType::kDynamic;
     }
     demand += ttmlir::utils::volume<int64_t>(info.producer->scratchShape);
@@ -878,10 +877,12 @@ static bool fuseOuterScfLoops(SmallVector<affine::AffineForOp> &scratchLoops,
 
 static bool containsUnsafeFusedConsumer(ArrayRef<affine::AffineForOp> loops) {
   return llvm::any_of(loops, [](affine::AffineForOp loop) {
-    return loop.walk([](Operation *op) {
-      return isa<TileDivOp, TileWhereOp>(op) ? WalkResult::interrupt()
-                                             : WalkResult::advance();
-    }).wasInterrupted();
+    return loop
+        .walk([](Operation *op) {
+          return isa<TileDivOp, TileWhereOp>(op) ? WalkResult::interrupt()
+                                                 : WalkResult::advance();
+        })
+        .wasInterrupted();
   });
 }
 
--- a/lib/Dialect/D2M/Transforms/InsertSpillAndScratch.cpp
+++ b/lib/Dialect/D2M/Transforms/InsertSpillAndScratch.cpp
@@ -13,6 +13,8 @@
 #include "mlir/IR/Builders.h"
 #include "mlir/IR/PatternMatch.h"
 
+#include "llvm/ADT/SmallPtrSet.h"
+
 namespace mlir::tt::d2m {
 #define GEN_PASS_DEF_D2MINSERTSPILLANDSCRATCH
 #include "ttmlir/Dialect/D2M/Transforms/Passes.h.inc"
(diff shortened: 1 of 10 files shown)


lib/Dialect/D2M/Transforms/InsertSpillAndScratch.cpp
```diff
@@ -13,6 +14,10 @@
 #include "mlir/IR/Builders.h"
 #include "mlir/IR/PatternMatch.h"
 
+#include "llvm/ADT/SmallPtrSet.h"
```
vwellsTT, line 17: nit: this should be attached to mlir header block I think

lib/Dialect/D2M/Transforms/InsertSpillAndScratch.cpp
```diff
@@ -63,6 +68,38 @@ struct IntermediateAllocInfo {
   SmallVector<affine::AffineStoreOp> stores; // Stores to this alloc
 };
 
+// Compute total static scratch buffer capacity required by a GenericOp.
+static std::optional<int64_t> getScratchCapacity(GenericOp genericOp) {
+  std::optional<int64_t> capacity;
+  genericOp.walk([&](ScratchInitOp scratchInit) {
+    auto memrefType =
+        mlir::dyn_cast<MemRefType>(scratchInit.getScratch().getType());
+    if (!memrefType) {
```
vwellsTT: this should represent an error I think?

lib/Dialect/D2M/Transforms/InsertSpillAndScratch.cpp
```diff
@@ -63,6 +68,38 @@ struct IntermediateAllocInfo {
...
+      return;
+    }
+    ArrayRef<int64_t> shape = memrefType.getShape();
+    if (!llvm::is_contained(shape, ShapedType::kDynamic)) {
```
vwellsTT: also probably assert-able?

lib/Dialect/D2M/Transforms/InsertSpillAndScratch.cpp
```diff
@@ -63,6 +68,38 @@ struct IntermediateAllocInfo {
...
+      capacity = ttmlir::utils::volume<int64_t>(shape);
```
vwellsTT: confused by this func, it's just overwriting "capacity" each op it sees, not taking a max or anything?

lib/Dialect/D2M/Transforms/InsertSpillAndScratch.cpp
```diff
@@ -63,6 +68,38 @@ struct IntermediateAllocInfo {
...
+    }
+  });
+  return capacity;
+}
+
+// Estimate total scratch memory demand from intermediate allocations.
+static int64_t estimateScratchDemand(ArrayRef<IntermediateAllocInfo> infos) {
```
vwellsTT, line 101: curious why this is useful; we don't actually ever allocate the total capcity anyway, right?

lib/Dialect/D2M/Transforms/LowerScratchAllocate.cpp
```diff
@@ -105,7 +99,11 @@ class D2MLowerScratchAllocatePass
     }
 
     // Verify allocations fit in the scratch buffer.
-    int64_t scratchCapacity = getNumElements(scratchMemRefType);
+    assert(!llvm::is_contained(scratchMemRefType.getShape(),
+                               ShapedType::kDynamic) &&
+           "scratch memrefs must be static");
+    int64_t scratchCapacity =
```
vwellsTT, line 105: shouldn't this use above helper instead of reimplementing?

lib/Dialect/D2M/Transforms/InsertSpillAndScratch.cpp
```diff
@@ -596,6 +875,186 @@ static bool fuseOuterScfLoops(SmallVector<affine::AffineForOp> &scratchLoops,
   return true;
 }
 
+static bool containsUnsafeFusedConsumer(ArrayRef<affine::AffineForOp> loops) {
```
vwellsTT, line 911: might be good to have a comment defining "unsafe" in this context

ckaravasilisTT:
Thanks for doing this, LGTM. We will soon have to rethink scratch buffer limit and how to fail gracefully, but this works well for now.

ckaravasilisTT (approved):

lib/Dialect/D2M/Transforms/InsertSpillAndScratch.cpp
ckaravasilisTT: +1

lib/Dialect/D2M/Transforms/InsertSpillAndScratch.cpp
```diff
@@ -219,6 +400,29 @@ getScratchAccessMapAndOperands(ScratchLoopInfo &info, MLIRContext *ctx) {
...
+  while (parent && prefixes.size() < missingRank) {
+    if (auto parentLoop = dyn_cast<affine::AffineForOp>(parent)) {
+      prefixes.push_back(parentLoop);
+    }
+    parent = parent->getParentOp();
+  }
+
+  std::reverse(prefixes.begin(), prefixes.end());
+  return prefixes;
+}
```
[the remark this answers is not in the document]
anuragsingh-tt: This is a fair suggestion. I will switch to FailureOrs.

lib/Dialect/D2M/Transforms/InsertSpillAndScratch.cpp
```diff
@@ -184,6 +199,172 @@ static void collectLoadsInLoop(Value memref, ScratchLoopInfo &loopInfo,
...
+  for (auto loop : loops) {
+    loop.getInductionVar().replaceAllUsesWith(sharedLoop.getInductionVar());
+  }
+
+  Operation *sharedYield = sharedLoop.getBody()->getTerminator();
+  for (auto &ops : setupOpsBeforeLoop) {
+    for (Operation *op : ops) {
+      op->moveBefore(sharedLoop);
+    }
+  }
```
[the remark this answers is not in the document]
anuragsingh-tt: Sure we can add a side-effect check.

lib/Dialect/D2M/Transforms/InsertSpillAndScratch.cpp
anuragsingh-tt: I thought I added a header check a while back to the precommit to enforce the guidelines. Need to check if that's still working as intended.

lib/Dialect/D2M/Transforms/InsertSpillAndScratch.cpp
anuragsingh-tt: Fixing.

lib/Dialect/D2M/Transforms/InsertSpillAndScratch.cpp
anuragsingh-tt: The capacity value is useful because this pass decides whether sibling loop fusion is required before creating those scratch allocations. If it fits already, we don't fuse the loops.

lib/Dialect/D2M/Transforms/LowerScratchAllocate.cpp
anuragsingh-tt: They're doing slightly different things, they both use utils::volume though.

lib/Dialect/D2M/Transforms/InsertSpillAndScratch.cpp
anuragsingh-tt: Need to add a TODO here. There were some Dst related issues I didn't fully investigate as that's another rabbit hole.

vwellsTT (approved):

anuragsingh-tt:
[More detailed description here.](https://docs.google.com/document/d/1_kpf4jfHtQ-qJa1pLu4T5QDcrSFpdgoZdBnq1adhqXY/edit?tab=t.lcvhcxnsdp7v)

PR #8152, opened 2026-04-28
[runtime] fix profiling of traced graphs

### LINKED ISSUE #4216: GPT-OSS 120B TP8-DP4 decode-only fails: Event Synchronization not supported during trace capture
## Description

Running the GPT-OSS 120B TP8-DP4 benchmark in decode-only mode with tracy profiling crashes with `TT_FATAL: Event Synchronization is not supported during trace capture` from tt-metal's `fd_mesh_command_queue.cpp`.

Claude's theory:

The failure occurs during `tokenizer.batch_decode(next_token_ids_replicated.to("cpu"))` — the `.to("cpu")` triggers an event synchronization while a trace is being captured.

## Error

```
RuntimeError: TT_FATAL @ .../tt-metal/tt_metal/distributed/fd_mesh_command_queue.cpp:725: !trace_id_.has_value()
info:
Event Synchronization is not supported during trace capture.
```

Full traceback:
```
tests/benchmark/test_llms.py:1689: in test_gpt_oss_120b_tp_dp_galaxy_batch_size_128
    test_llm_tp(...)
tests/benchmark/test_llms.py:231: in test_llm_tp
    test_llm(...)
tests/benchmark/test_llms.py:128: in test_llm
    results = benchmark_llm_torch_xla(...)
tests/benchmark/benchmarks/llm_benchmark.py:480: in benchmark_llm_torch_xla
    _, iteration_times = generate_and_benchmark(...)
tests/benchmark/llm_utils/decode_utils.py:264: in generate_and_benchmark
    decoded = tokenizer.batch_decode(next_token_ids_replicated.to("cpu"))
python_package/tt_torch/torch_overrides.py:22: in __torch_function__
    return func(*args, **(kwargs or {}))
RuntimeError: TT_FATAL ... Event Synchronization is not supported during trace capture.
```

Additionally, tracy post-processing also fails:
```
AssertionError: Device 24 present in host logs but missing from cpp_device_perf_report.csv
```

## Repro Steps
Using a WH galaxy, on branch `odjuricic/gptoss-tp-dp` (PR #4205):

```bash
tracy -p -r -m pytest -svv tests/benchmark/test_llms.py::test_gpt_oss_120b_tp_dp_galaxy_batch_size_128 --num-layers 1 --decode-only --max-output-tokens 3
```

## Environment

- Branch: `odjuricic/gptoss-tp-dp`
- PR: #4205
- tt-metal source: `fd_mesh_command_queue.cpp:725`

pilkicTT:
This change fixes a couple of problems encountered when profiling models which use `trace` mode:
- `ReadMeshDeviceProfilerData()` is called during an active trace capture - this is not allowed in `tt-metal`, so we should skip doing this
- profiler data buffer on the device is overflown, so `tracy` reports missing profiler data for an op.

We now read the profiler data each time we execute a program. This is our best effort to avoid profiler data overflow. Overflow can still happen, e.g. we have captured more ops in the trace than the default limit for the buffer (1000). In that case, user can increase the buffer limit via `--op-support-count` flag in `tracy` CLI.

The function for reading the profiler data is renamed for clarity.

Closes tenstorrent/tt-xla#4216
Closes tenstorrent/tt-xla#4036

### DIFF
--- a/runtime/include/tt/runtime/detail/ttnn/program_executor.h
+++ b/runtime/include/tt/runtime/detail/ttnn/program_executor.h
@@ -73,9 +73,13 @@ private:
   void runOperation(const ::tt::target::ttnn::Operation *op);
 
   /**
-   * Dumps device profile counters if needed
+   * Conditionally reads profiler data from device buffer.
+   *
+   * The read will be performed in case:
+   * - we don't have an active trace capture.
+   * - force flag is set or we have executed at least `dumpDeviceRate` ops
    */
-  void dumpPerfCountersIfNeeded();
+  void readProfilerDataIfNeeded(bool force = false);
 
 #if defined(TT_RUNTIME_DEBUG) && TT_RUNTIME_DEBUG == 1
   /**
--- a/runtime/lib/ttnn/program_executor.cpp
+++ b/runtime/lib/ttnn/program_executor.cpp
@@ -210,12 +210,26 @@ void ProgramExecutor::execute() {
 #if defined(TT_RUNTIME_DEBUG) && TT_RUNTIME_DEBUG == 1
     syncAfterOpIfNeeded();
 #endif
+
     runOpCallback(debug::Hooks::get().getPostOperatorCallback(),
                   executableHandle, op, context.get());
-    dumpPerfCountersIfNeeded();
+    readProfilerDataIfNeeded();
   }
+
   runProgramCallback(debug::Hooks::get().getpostProgramCallback(),
                      executableHandle, context.get());
+
+  // Always read the profiler data after program execution. This handles trace
+  // replay case, since the op counting logic (in `readProfilerDataIfNeeded`)
+  // doesn't work there - we execute only one op (trace replay) which will
+  // indirectly execute lots of ops on the device. This will reduce the chance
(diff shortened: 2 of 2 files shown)


acolicTT (approved):

jameszianxuTT (approved):
> This is our best effort to avoid profiler data overflow. Overflow can still happen, e.g. we have captured more ops in the trace than the default limit for the buffer (1000). In that case, user can increase the buffer limit via --op-support-count flag in tracy CLI.

Other than seeing that ops are missing in the profile, is there any way to detect this case in runtime? It is not really related to this PR but just wondering.

pilkicTT:
> Other than seeing that ops are missing in the profile, is there any way to detect this case in runtime? It is not really related to this PR but just wondering.

@jameszianxuTT - currently, missing op info is a hard failure in the `tracy` CLI tool

jameszianxuTT:
> > Other than seeing that ops are missing in the profile, is there any way to detect this case in runtime? It is not really related to this PR but just wondering.
> 
> @jameszianxuTT - currently, missing op info is a hard failure in the `tracy` CLI tool

Ah, good to know, thanks!

PR #8153, opened 2026-04-28
[d2m] ttir.repeat decomposition

### LINKED ISSUE #6594: [d2m] ttir.repeat support
Seems like it could be trivially implemented with broadcast

jgrimTT:
### Ticket
Closes [#6954](https://github.com/tenstorrent/tt-mlir/issues/6594)

### Problem description
Support ttir.repeat by decomposing into ttir ops that are supported in d2m

### What's changed
Added support for ttir.repeat by decomposing into ttir.concat
Added lit and builder test

### Checklist


wenbinlyuTT (changes requested):
1. IMHO it's better to do the bit decomposition right away, it's cleaner:
```cpp
#include <bitset>

int64_t foo = 7;
std::bitset<64> bar(foo);
for (int i = 0; i < 64; i++) {
    // Do the doubling
    if (bar.test(i)) {
        printf("%d\n", i);
        // Concat the current value to the final output.
    }
    // Break if we've reached the highest set bit.
}
```
This will give you all the powers-of-two you need, then you can just double the concat until you reach the highest set bit and assemble the final product whenever you encounter a set bit, all in a single loop.
2. Non tile-aligned concat is supported so you might want to add builder tests for that as well, like `((2, 3), [5, 7])`.
3. Maybe I've missed something but why is this in TTIRToD2M instead of being a decomposition pattern?

jgrimTT:
> 1. IMHO it's better to do the bit decomposition right away, it's cleaner:
> 
> ```c++
> #include <bitset>
> 
> int64_t foo = 7;
> std::bitset<64> bar(foo);
> for (int i = 0; i < 64; i++) {
>     // Do the doubling
>     if (bar.test(i)) {
>         printf("%d\n", i);
>         // Concat the current value to the final output.
>     }
>     // Break if we've reached the highest set bit.
> }
> ```
> 
> This will give you all the powers-of-two you need, then you can just double the concat until you reach the highest set bit and assemble the final product whenever you encounter a set bit, all in a single loop. 2. Non tile-aligned concat is supported so you might want to add builder tests for that as well, like `((2, 3), [5, 7])`. 3. Maybe I've missed something but why is this in TTIRToD2M instead of being a decomposition pattern?

Thanks for the review @wenbinlyuTT! I think I addressed everything, lmk how it looks

lib/Dialect/TTIR/Transforms/DecomposeComposites.cpp
```diff
@@ -456,6 +456,130 @@ struct DecomposeSoftmaxPattern : public OpRewritePattern<SoftmaxOp> {
...
+      return input;
+    }
+
+    auto inputType = cast<RankedTensorType>(input.getType());
+    int64_t originalDimSize = inputType.getShape()[dim];
+
+    // Find the highest set bit to know when to stop building chunks
+    int64_t highestBit = 63 - __builtin_clzll(repeatCount);
+
+    // Optimization: if repeatCount is a power of 2, just do log2(N) doublings
```
wenbinlyuTT: I think the normal path below will produce identical results to this? We will just end up with partsToConcat[0] and that's the advantage of the bit decomposition approach.

lib/Dialect/TTIR/Transforms/DecomposeComposites.cpp
```diff
@@ -456,6 +456,130 @@ struct DecomposeSoftmaxPattern : public OpRewritePattern<SoftmaxOp> {
...
+      if (bit < highestBit) {
+        SmallVector<Value> inputs = {currentChunk, currentChunk};
+
+        SmallVector<int64_t> doubleShape(inputType.getShape().begin(),
+                                         inputType.getShape().end());
+        doubleShape[dim] = originalDimSize * (1LL << (bit + 1));
+        auto doubleType = RankedTensorType::get(
+            doubleShape, inputType.getElementType(), inputType.getEncoding());
+
+        auto dimAttr = rewriter.getSI32IntegerAttr(static_cast<int32_t>(dim));
```
wenbinlyuTT: This is loop-invariant and so can be lifted out from the for loop.

test/ttmlir/Conversion/TTIRToD2M/named_to_generic.mlir
```diff
@@ -450,6 +450,16 @@ module {
     return %0 : tensor<32x64xf32>
   }
 
+  // CHECK-LABEL: func @named_repeat
+  func.func @named_repeat(%arg0: tensor<32x32xf32>) -> tensor<64x96xf32> {
+    // CHECK-NOT: ttir.repeat
+    // CHECK-NOT: ttir.concat
+    // CHECK: "d2m.composite_view"
+    // CHECK: "d2m.composite_view"
```
wenbinlyuTT: If I understand it correctly there should be three composite views for this repeat, it's more accurate to do:
```
CHECK: "d2m.composite_view"
CHECK: "d2m.composite_view"
CHECK: "d2m.composite_view"
CHECK-NOT: "d2m.composite_view"
```

jgrimTT: Moved test to another file :)

wenbinlyuTT (approved):
All looks good now, thanks!

lib/Dialect/TTIR/Transforms/DecomposeComposites.cpp
```diff
@@ -456,6 +456,113 @@ struct DecomposeSoftmaxPattern : public OpRewritePattern<SoftmaxOp> {
...
+    assert(repeatCount > 0 && "repeatCount must be positive");
+
+    if (repeatCount == 1) {
+      return input;
+    }
+
+    auto inputType = cast<RankedTensorType>(input.getType());
+    int64_t originalDimSize = inputType.getShape()[dim];
+
+    // Find the highest set bit to know when to stop building chunks
```
vwellsTT: nit: sentence comments in this file pls

vwellsTT (approved):

anuragsingh-tt (approved):

PR #8160, opened 2026-04-28
Fold elementwise binary ops on constants

### LINKED ISSUE #8159: Fold elementwise binary ops on constants
We should be able to fold binary elementwise TTIR operators when their argument is a constant creation op.

List of operators that would be good to support:
- `ttir.add`
- `ttir.atan2`
- `ttir.bitwise_and`
- `ttir.bitwise_or`
- `ttir.bitwise_xor`
- `ttir.div`
- `ttir.eq`
- `ttir.ge`
- `ttir.gt`
- `ttir.logical_and`
- `ttir.logical_left_shift`
- `ttir.logical_or`
- `ttir.logical_right_shift`
- `ttir.logical_xor`
- `ttir.maximum`
- `ttir.minimum`
- `ttir.multiply`
- `ttir.ne`
- `ttir.pow`
- `ttir.remainder`
- `ttir.subtract`

We don't need to support activation functions (like `ttir.gel_bw`) since we don't expect them to appear with constant arguments. We also don't need to support `ttir.le` and `ttir.lt` since they are canonicalized to other operators.

When implementing fold methods for operators that didn't have them, it would also be good to check if they have canonicalization patterns which could be moved to fold methods.

amalkovTT:
### Ticket
Closes #8159 

### Problem description
We should be able to fold binary elementwise TTIR operators when their argument is a constant creation op.

### What's changed
Implemented folders for eltwise binary ops. While creating new folders, some existing canonicalizations were moved to folders. Implemented tests.

### Checklist


lib/Dialect/TTIR/IR/TTIROps.cpp
```diff
@@ -233,6 +233,161 @@ constantFoldEltwiseUnary(mlir::Operation *op, mlir::Attribute inputAttr,
...
+    llvm::SmallVector<int64_t> rhsCoord = resultCoord;
+    for (size_t i = 0; i != resultCoord.size(); ++i) {
+      if (lhsShape[i] == 1) {
+        lhsCoord[i] = 0;
+      }
+      if (rhsShape[i] == 1) {
+        rhsCoord[i] = 0;
+      }
+    }
+
```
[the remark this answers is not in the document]
amalkovTT: Fixed.

test/ttmlir/Dialect/TTIR/canonicalize/bitwise_xor_canonicalize_tests.mlir
dmilinkovicTT, line 1: We could perhaps move these to `eltwise_binary_op_fold_tests.mlir`, since we're not canonicalizing anymore?

lib/Dialect/TTIR/IR/TTIROps.cpp
```diff
@@ -233,6 +233,185 @@ constantFoldEltwiseUnary(mlir::Operation *op, mlir::Attribute inputAttr,
   return nullptr;
 }
 
+static bool fulfillFoldEltwiseBinaryConditions(mlir::Operation *op,
```
dmilinkovicTT: Nit: we should avoid present-tense `fulfill`, as the function doesn't actually fulfill anything by itself. :)

lib/Dialect/TTIR/IR/TTIROps.cpp
```diff
@@ -233,6 +233,185 @@ constantFoldEltwiseUnary(mlir::Operation *op, mlir::Attribute inputAttr,
...
+  }
+
+  // If exactly one input is a splat we resize it to the shape of the other,
+  // to be able to iterate trgough the same number of elements.
+  if (lhs.isSplat() && !rhs.isSplat()) {
+    lhs = lhs.resizeSplat(rhs.getType());
+  }
+  if (!lhs.isSplat() && rhs.isSplat()) {
+    rhs = rhs.resizeSplat(lhs.getType());
+  }
```
dmilinkovicTT: Can we move the both-splat case above this?

Generally, this function would be easier to follow if we divided it into three segments (not wit additional functions, but with early-return sections):
- both splat
- one splat
- none splat

lib/Dialect/TTIR/IR/TTIROps.cpp
```diff
@@ -233,6 +233,185 @@ constantFoldEltwiseUnary(mlir::Operation *op, mlir::Attribute inputAttr,
...
+
+  auto resultType = mlir::cast<ShapedType>(op->getResult(0).getType());
+  // If we have tensors of different shapes, we need to add leading dimensions
+  // of size 1 to the smaller one.
+  lhs = addLeadingDimsToMatchShape(lhs, resultType.getShape());
+  rhs = addLeadingDimsToMatchShape(rhs, resultType.getShape());
```
dmilinkovicTT, line 327: If the operand's size doesn't match the result size, `fulfillFoldEltwiseBinaryConditions` would return false, and we would never reach this?

lib/Dialect/TTIR/IR/TTIROps.cpp
```diff
@@ -233,6 +233,185 @@ constantFoldEltwiseUnary(mlir::Operation *op, mlir::Attribute inputAttr,
...
+    auto rhsSplatValue = rhs.getSplatValue<ElementType>();
+    ElementType result = mapFn(lhsSplatValue, rhsSplatValue);
+    return mlir::SplatElementsAttr::get(resultType, result);
+  }
+
+  llvm::SmallVector<ElementType> resultValues;
+  resultValues.reserve(resultType.getNumElements());
+
+  // If no broadcast is needed, just map over the input values.
+  if (lhs.getType() == resultType && rhs.getType() == resultType) {
```
dmilinkovicTT: Same here, this should always be true?

lib/Dialect/TTIR/IR/TTIROps.cpp
dmilinkovicTT: Why do we need to resize the splats? IMHO, it doesn't make the further steps simpler (if that was the intention).

In one-splat scenario, we could just iterate the non-splat `DenseElementsAttr` and use the splat value as the other operand.

lib/Dialect/TTIR/IR/TTIROps.cpp
```diff
@@ -268,11 +447,102 @@ class ApplyToAPFloat {
   Fun fn;
 };
 
-// Deduction guide to avoid having to write the template parameter at the
+// Callable that maps a C++ float binary function over two APFloats by
+// converting them to f32 and converting result to the original type. This
+// allows folding with non standard float types like bf16 using functions from
+// C++ standard library.
+template <typename Fun>
+class ApplyToAPFloats {
```
dmilinkovicTT: Did you maybe try to use variadic templates in order to combine `ApplyToAPFloats` and `ApplyToAPFloat` into a single helper class?

lib/Dialect/TTIR/IR/TTIROps.cpp
amalkovTT: It checks that the element type of the operand matches the element type of the result (while shape doesn't have to be the same). It also checks that if we have one splat and one non splat operand, than the non-splat one should have the same shape as the result, to avoid the dangerous case that we talked about. It allows broadcasting when both operands are non-splat or if only the splat one has different shape.

lib/Dialect/TTIR/IR/TTIROps.cpp
amalkovTT: SplatElementsAttr has getValues which is just a range for iterating over the same element, `size` times. In order to use that API we need to resize the splat.

It looked simpler to me, because otherwise we would need to have 3 cases that are all doing the same thing.

lib/Dialect/TTIR/IR/TTIROps.cpp
amalkovTT: Renamed to `check...`.

amalkovTT:
Running [uplift test](https://github.com/tenstorrent/tt-xla/actions/runs/25385149343) and [performance benchmark](https://github.com/tenstorrent/tt-xla/actions/runs/25385333149).

lib/Dialect/TTIR/IR/TTIROps.cpp
amalkovTT: I tried to organize this function better.

amalkovTT: Fixed.

amalkovTT:
[Uplift test passed](https://github.com/tenstorrent/tt-xla/actions/runs/25385149343) and [performance benchmark](https://github.com/tenstorrent/tt-xla/actions/runs/25385333149) shows no regressions against [baseline](https://github.com/tenstorrent/tt-xla/actions/runs/25426968948).

dmilinkovicTT (approved):

PR #8177, opened 2026-04-28
Add `ttir.invoke_external`

abrown:
In order to inject external functionality into the tt-mlir compilation pipeline, I am prototyping a new instruction:

```mlir
%result = ttir.invoke_external(%arg0, %arg...) {path = "external.mlir", entry = "foo"} : ...
```

This instruction would allow us to generate Tenstorrent MLIR and pass it through to tt-mlir as a file to be "linked" against. The `ttir.invoke_external` is then lowered to a `func.call` to a linked-in function. For example, this change will compile the MLIR above by merging in all of `external.mlir` into the currently-compiled module and replacing `ttir.invoke_external` with a `func.call @foo`. (The "MLIR in a file" approach seemed solid for prototyping; it is not load-bearing to this design and we could investigate passing the MLIR in an attribute instead in the future).

Something like `ttir.invoke_external` is key to allow injecting functionality prior to tt-xla's pipeline. As I understand it, tt-xla only allows a fixed, upstream subset of custom operations: [`custom_ops.py`]. This change would introduce a general mechanism for custom operations by using the same mechanism: a `stablehlo_custom_call` would wrap up a `tt.invoke_external` with the `path` and `entry` passed as `frontend_attributes` (e.g, [`scaled_dot_product_attention`]). Later, this change will lower `tt.invoke_external` to a `ttir.invoke_external` in the StableHLO to TTIR conversions.

[`custom_ops.py`]: https://github.com/tenstorrent/tt-xla/blob/9e338f8/python_package/tt_torch/custom_ops.py
[`scaled_dot_product_attention`]: https://github.com/tenstorrent/tt-xla/blob/9e338f8/python_package/tt_torch/custom_ops.py#L280-L286

This is a prototype: I'm not too attached to any of the details here and I'm especially interested in feedback on the approach.


sdjordjevicTT:
Hi @abrown, thanks for putting this together and for the detailed write-up. Before digging into the implementation details, could you elaborate a bit more on the end-to-end use case(s) you have in mind for this op? In particular, it would be helpful to better understand the intended scenarios where injecting external MLIR into the pipeline is needed, and how you see this fitting into the broader extensibility story.

abrown:
Sure: I'm working on [Triton](https://github.com/triton-lang/triton)-related things and one of our goals is being able to run Triton kernels on different kinds of hardware. We have ways to generate Tenstorrent MLIR program from Triton but were missing some way to integrate that with the rest of the model graph; that's what this PR is prototyping. @nsmithtt can probably give you more context; yeah, without that this definitely feels a bit out-of-the-blue!

sjw36:
@nsmithtt is this ready to land?

nsmithtt:
@sjw36 I think it needs to be rebased and we'll rerun CI

nsmithtt (approved):

abrown:
Ok, rebased.

sjw36:
@nsmithtt needs approval to run ci

PR #8178, opened 2026-04-29
Adding a new SaturatingRightShiftFoldPattern

aorlovicTT:
### Problem description
When running:
```
jax.random.uniform
jax.random.normal
jax.random.split
``` 
respectively, we get the following outputs on CPU:
```
[[0.48870957 0.6797972  0.6162715  0.5610161  0.4506446 ]]
[[-0.02830462  0.46713185  0.29570296  0.15354592 -0.12403282]]
[[1832780943  270669613] [  64467757 2916123636]]
```
and on TT:
```
[[0.4596039  0.2298019  0.83841693 0.91920793 0.35367072]]
[[ 1.8510936   2.1434302   1.1344137   1.5207957  -0.03319528]]
[[1832780943  270669613] [1154097635  473298049]]
```
### What's changed
In short, the operation that didn't work was: 
```
[0, 1] >> [32, 32]
```
on CPU the following gives the output of ```[0, 0]``` while on TT we still get ```[0, 1]``` (the reason why the first key in split is correct and the second isn't)
the reason being that the shift amount is 32 which is undefined behaviour.

After the change we on TT we should get:
```
[[0.48828125 0.6796875  0.61621094 0.5605469  0.4501953 ]]
[[-0.0293787   0.46682528  0.29554442  0.15235595 -0.12516783]]
[[1832780943  270669613] [  64467757 2916123636]]
```
Also for this to work we should have https://github.com/tenstorrent/tt-xla/pull/4427 merged (otherwise only the split one will work)
### Checklist


acicovicTT:
You can move this logic to `/lib/Dialect/TTIR/Transforms/TTIRFusing.cpp` and fuse on TTIR level

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
```diff
@@ -2990,6 +3034,68 @@ class ReshapeBroadcastReshapeToRepeatPattern
...
+//   %b = "ttir.broadcast"(%c) ... : (tensor<i32>) -> tensor<4xi32>
+//   %r = "ttir.logical_right_shift"(%x, %b)
+//          : (tensor<4xi32>, tensor<4xi32>) -> tensor<4xi32>
+//
+// Example (after):
+//   %r = "ttir.constant"() <{value = dense<0> : tensor<4xi32>}> : tensor<4xi32>
+//
+// The shift, broadcast, and constant feeding the RHS become dead and are
+// removed by later DCE.
+class SaturatingRightShiftFoldPattern
```
acicovicTT, line 3053: With minimal changes, you can make this work for 32 and 64 bit integers, e.g. similarly to how it was done here https://github.com/tenstorrent/tt-mlir/pull/7078/changes

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
```diff
@@ -2990,6 +3034,68 @@ class ReshapeBroadcastReshapeToRepeatPattern
   }
 };
 
+// Folds `ttir.logical_right_shift %x, splat-constant<C>` to a zero constant
```
sdjordjevicTT, line 3037: The fusing isn't the right place for this. This pattern folds logical_right_shift(x, large_const) → zero_constant. It's a constant-folding/canonicalization rule, eliminating an op when its result is statically determinable.

@amalkovTT has this folder in a draft PR. @amalkovTT can you send your PR here so @aorlovicTT can take a look at it?

amalkovTT: Its #8160. The folder in this PR is not really doing the same thing, but this can be implemented in a similar way by checking if adaptor.getRhs() is not null (which means that its constant), checking if it is a splat attribute with value larger than bit width, and returning a new splat attribute with 0.

PR #8182, opened 2026-04-29
[TTIR] Replace MoE bridge ops with ttir.composite

mtopalovicTT:
The 5 MoE ops (all_to_all_dispatch, all_to_all_dispatch_metadata,
all_to_all_combine, selective_reduce_combine, moe_expert_token_remap)
had no TTIR-level semantics — they were 1:1 passthroughs between
stablehlo.custom_call and the matching TTNN op. Letting
stablehlo.composite
flow through TTIR is not ideal because it would couple the TTIR dialect
to
stablehlo permanently and force every downstream pass/target to
whitelist it.
A self-contained ttir.composite op keeps TTIR independent and gives us a
single bridge construct.

Changes:
- Add ttir.composite (StrAttr name, variadic operands/results,
  DictionaryAttr composite_attributes) with a name-dispatched verifier
that
  preserves the checks of the deleted ops.
- Delete the 5 named TTIR ops and their verifiers.
- StableHLOToTTIR: rewrite custom_call patterns to emit ttir.composite.
- TTIRToTTNN: replace 5 per-op patterns with a single name-dispatching
  CompositeOpConversionPattern.
- Python builder: keep user-facing helpers (now emitting ttir.composite)
  and add generic @parse / @split dispatchers on ttir.CompositeOp.
- Drop the 4 deleted op-class entries from the golden mapping.
- Update lit tests (4 negative + 2 sparse_moe) to ttir.composite syntax.

### DIFF
--- a/include/ttmlir/Dialect/TTIR/IR/TTIROps.td
+++ b/include/ttmlir/Dialect/TTIR/IR/TTIROps.td
@@ -6254,7 +6254,7 @@ def TTIR_SparseMatmulOp : TTIR_NamedOp<"sparse_matmul"> {
 def TTIR_CompositeOp : TTIR_Op<"composite"> {
     let summary = "Opaque composite op that lowers 1:1 to a matching TTNN op.";
     let description = [{
-      Carries a symbolic `name` and a dictionary of `composite_attributes`. Used
+      Carries a symbolic `name` and a dictionary of `op_attributes`. Used
       as a bridge for `stablehlo.composite` / `stablehlo.custom_call` operations
       that map 1:1 to a TTNN op (no transformation is performed at the TTIR
       level). The TTIR->TTNN conversion dispatches on `name` and constructs the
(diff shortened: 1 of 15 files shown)


dlokeTT:
Hi @mtopalovicTT, I had put some thoughts here https://github.com/tenstorrent/tt-mlir/issues/7880 but this impacts d2m. We're starting to try and bring models up thru d2m and we can't match these ops. Currently we're trying to use the same ttir as y'all but maybe we should have a meeting discussing how to handle these ops? At least having some kind of hints to the compiler on how to decompose these into more basic ttir primitives like all_gather and such would be helpful. Lmk what you think

mtopalovicTT:
> Hi @mtopalovicTT, I had put some thoughts here #7880 but this impacts d2m. We're starting to try and bring models up thru d2m and we can't match these ops. Currently we're trying to use the same ttir as y'all but maybe we should have a meeting discussing how to handle these ops? At least having some kind of hints to the compiler on how to decompose these into more basic ttir primitives like all_gather and such would be helpful. Lmk what you think

I missed your comment on the issue. Yea I agree with you, thanks for letting me know. It seems that I went in wrong direction with this. Initially I wanted to remove bridge ops from ttir which don't have composites since they carry sharding metadata which is specific to ttnn, but on second thought those don't belong in ttir at all. I will close this PR, since it doesn't make sense at all, but I will need to take a look at what composites we have today in shlo to ttir composite legalization pass and if there are is straight path forward. We can sync offline next week maybe once I gather enough information. How does that sound?

Issue #7664, opened 2026-03-26
Alphafold3 Bringup


aorlovicTT:
Umbrella issue to keep track of progress of AF3 bringup.

PR #8185, opened 2026-04-29
Added gather index int32 support

aorlovicTT:
### Ticket
Related to #7664 

### Problem description
The current gather implementation didn't account for index tensor having INT32 dtype, so we can just convert them to UINT32, afaik we are not supporting negative indexes on gather op so this typecasting should be valid.

### What's changed

### Checklist

### DIFF
--- a/lib/Dialect/TTNN/IR/TTNNOps.cpp
+++ b/lib/Dialect/TTNN/IR/TTNNOps.cpp
@@ -3598,10 +3598,10 @@ mlir::tt::ttnn::ReduceScatterOp::fold(FoldAdaptor adaptor) {
   const ::mlir::RankedTensorType resultType = getResult().getType();
 
   if (!indexType.getElementType().isUnsignedInteger(16) &&
-      !indexType.getElementType().isUnsignedInteger(32)) {
-    return emitOpError() << "Index tensor must have an unsigned integer "
-                         << "type of ui16 or ui32, got "
-                         << indexType.getElementType();
+      !indexType.getElementType().isUnsignedInteger(32) &&
+      !indexType.getElementType().isSignedInteger(32)) {
+    return emitOpError() << "Index tensor must have type ui16, ui32, or si32, "
+                         << "got " << indexType.getElementType();
   }
 
   const int64_t inputRank = inputType.getRank();
--- a/runtime/lib/ttnn/operations/data_movement/gather.cpp
+++ b/runtime/lib/ttnn/operations/data_movement/gather.cpp
@@ -21,8 +21,14 @@ void run(const ::tt::target::ttnn::GatherOp *op, ProgramContext &context) {
       ::tt::runtime::ttnn::utils::createMemoryConfigIfNeeded(
           op->memory_config());
 
+  // Indices must be non negative, otherwise this will break
+  ::ttnn::Tensor newIndex = index;
+  if (index.dtype() == ::tt::tt_metal::DataType::INT32) {
+    newIndex = ::ttnn::typecast(index, ::tt::tt_metal::DataType::UINT32);
+  }
+
   ::ttnn::Tensor out =
-      ::ttnn::gather(input, dim, index, /*sparse_grad=*/false,
+      ::ttnn::gather(input, dim, newIndex, /*sparse_grad=*/false,
                      outputMemoryConfig, std::nullopt, std::nullopt);
 
   tensorPool.insertTTNNTensorAndValidate(op->out(), out);
--- a/runtime/lib/ttnn/operations/data_movement/gather.cpp
+++ b/runtime/lib/ttnn/operations/data_movement/gather.cpp
@@ -9,6 +9,16 @@
 
 namespace tt::runtime::ttnn::operations::data_movement {
 
+namespace {
+#if defined(TT_RUNTIME_DEBUG) && TT_RUNTIME_DEBUG == 1
+bool allNonNegative(const ::ttnn::Tensor &index) {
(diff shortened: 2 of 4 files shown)


runtime/lib/ttnn/operations/data_movement/gather.cpp
```diff
@@ -21,8 +21,14 @@ void run(const ::tt::target::ttnn::GatherOp *op, ProgramContext &context) {
       ::tt::runtime::ttnn::utils::createMemoryConfigIfNeeded(
           op->memory_config());
 
+  // Indices must be non negative, otherwise this will break
```
jameszianxuTT: It seems like you could assert here, unless this condition is enforced elsewhere

jameszianxuTT: (And also add a negative test for this case)

aorlovicTT: > (And also add a negative test for this case)

I'm not entirely sure how to make a negative test, since the negative indices would only be visible in runtime 😞

test/ttmlir/Silicon/TTNN/n150/data_movement/gather/gather.mlir
dmilinkovicTT, line 1: Generally, we're migrating away: from `Silicon` tests:
- If you need to perform some `CHECK`-s to confirm the IR is valid, use `Dialect` (lit) tests
- If you need to confirm that the runtime aspects of an op are correct, use `Builder` (python) tests

In your case, I'd say you need both.

lib/Dialect/TTNN/IR/TTNNOps.cpp
```diff
@@ -3598,10 +3598,10 @@ ::mlir::LogicalResult GatherOp::verify() {
...
 
   if (!indexType.getElementType().isUnsignedInteger(16) &&
-      !indexType.getElementType().isUnsignedInteger(32)) {
-    return emitOpError() << "Index tensor must have an unsigned integer "
-                         << "type of ui16 or ui32, got "
-                         << indexType.getElementType();
+      !indexType.getElementType().isUnsignedInteger(32) &&
+      !indexType.getElementType().isSignedInteger(32)) {
+    return emitOpError() << "Index tensor must have type ui16, ui32, or si32, "
+                         << "got " << indexType.getElementType();
```
dmilinkovicTT, line 3604: PR title and description are mentioning `i32`, but here, we're allowing `si32`, which is different from the MLIR perspective.

lib/Dialect/TTNN/IR/TTNNOps.cpp
```diff
@@ -3598,10 +3598,10 @@ ::mlir::LogicalResult GatherOp::verify() {
   const ::mlir::RankedTensorType resultType = getResult().getType();
 
   if (!indexType.getElementType().isUnsignedInteger(16) &&
-      !indexType.getElementType().isUnsignedInteger(32)) {
-    return emitOpError() << "Index tensor must have an unsigned integer "
-                         << "type of ui16 or ui32, got "
-                         << indexType.getElementType();
+      !indexType.getElementType().isUnsignedInteger(32) &&
+      !indexType.getElementType().isSignedInteger(32)) {
```
dmilinkovicTT, line 3602: We're modeling the TTNN dialect to be as close as possible to the TTNN library.

Since the op implementation in the library indeed only supports `ui16` and `ui32`, we shouldn't deviate here.

As we can't be sure that `typecast` can be performed (since there are possibly negative values), a more robust solution is required, possibly involving some index manipulation on TTIR level to compensate for the fact that `ttnn::gather` doesn't support signed indices (since TTIR op doesn't have any restrictions on integer signedness, and it should, IMHO, stay like that).

runtime/lib/ttnn/operations/data_movement/gather.cpp
```diff
@@ -9,6 +9,16 @@
 
 namespace tt::runtime::ttnn::operations::data_movement {
 
+namespace {
+#if defined(TT_RUNTIME_DEBUG) && TT_RUNTIME_DEBUG == 1
+bool allNonNegative(const ::ttnn::Tensor &index) {
+  std::vector<int32_t> values = index.to_vector<int32_t>();
```
dmilinkovicTT, line 15: This will cause the following performance degradations:
1. device -> host transfer, if the tensor is on device
2. copy into the vector's buffer, if the tensor is already on host.

Even with the `TT_RUNTIME_DEBUG` flag, this doesn't seem right.

Performance topics aside, the runtime should be the last place we attempt to put workarounds like this one.

runtime/lib/ttnn/operations/data_movement/gather.cpp
```diff
@@ -21,8 +31,15 @@ void run(const ::tt::target::ttnn::GatherOp *op, ProgramContext &context) {
       ::tt::runtime::ttnn::utils::createMemoryConfigIfNeeded(
           op->memory_config());
 
+  ::ttnn::Tensor newIndex = index;
+  if (index.dtype() == ::ttnn::DataType::INT32) {
+    DEBUG_ASSERT(allNonNegative(index),
+                 "ttnn::gather INT32 index must be non-negative");
+    newIndex = ::ttnn::typecast(index, ::ttnn::DataType::UINT32);
```
dmilinkovicTT, line 38: Once again, runtime should be the last resort for such workarounds. It's much harder to diagnose potential performance / memory layout issues when "ghost" operations such as this one are sneakily executed by the runtime.

PR #7738, opened 2026-04-01
D2M: Reblock eltwise ops in Allocator

anuragsingh-tt:
- add auto block factor analysis for eltwise D2M generics.
- wire blocked stream allocation behavior through allocate.
- fix parallelization to avoid re-deriving grids from reblocked output types.
- pytest + lit tests for new behavior

PR #8192, opened 2026-04-29
Re-enable d2m all gather test

sohaibnadeemTT:
### Problem description
In https://github.com/tenstorrent/tt-mlir/pull/7738, there was a change to skip stream insertion for DMA-only generics but the DMA-only generic for the all gather op does in fact require a CB since we are not aliasing either the remote_load or the remote_store (since the remote_load is required to fetch the data from the input, and the remote_store is a write over fabric).

### What's changed
Revert the change.

### Checklist


anuragsingh-tt (approved):

hkwonTT (approved):
Confirmed that the all_gather builder test passes.

PR #8196, opened 2026-04-29
D2M: object-oriented CB API migration

wenbinlyuTT:
### Problem description
We need to migrate to the OO-style device API to prepare for Quasar bringup.

This PR migrates the D2M backend from the free-function CB API to the new OO-style CB API.

```cpp
#include "api/compute/cb_api.h" // Old compute core header.
#include "api/dataflow/dataflow_api.h" // Old data movement core header.

#include "experimental/circular_buffer.h" // New unified header.

// Explicit CB object creation.
constexpr int32_t cb_id = get_compile_time_arg_val(0);
experimental::CircularBuffer cb(cb_id);

// CB method call mappings.
cb_reserve_back(cb, n) -> cb.reserve_back(n)
cb_wait_front(cb, n) -> cb.wait_front(n)
cb_push_back(cb, n) -> cb.push_back(n)
cb_pop_front(cb, n) -> cb.pop_front(n)
get_write_ptr(cb) -> cb.get_write_ptr()
get_read_ptr(cb) -> cb.get_read_ptr()

// Compute API still uses numerical IDs.
add_tiles_init(cb_id, cb_id);
add_tiles(cb_id, cb_id, 0, 0, 0);
```

### What's changed
- The migration is implemented in the `TTKernelToEmitC` & `TTKernelToCpp` passes.
  - The `TTKernel` dialect still uses the old CB API names, since it still captures the right semantics.
- Naming the CB identifiers:
  - A consistent & precise name is needed since the same CB port can appear as multiple SSA values in the IR.
  - Ideally the name of the CB should be `cb_n` where n is the CB port/ID, so the kernel is easy to understand.
  - In most cases, the CB ID is passed as a kernel compile-time argument.
    - The CB port from the D2M dialect is discarded when `D2MToTTKernel` lowers `d2m.get_cb(port)` to `ttkernel.get_compile_time_arg_val(ctArgIndex)`.
    - The CB port actually used by the runtime is assigned in the `D2mToTTMetal` pass, sequentially.
    - Additionally, the spatial op can remap CBs and so we must be careful to not produce misleading names.
    - Name these CBs as `cb_ctarg_{arg_index}` so at least it matches its index in the kernel's compile-time arguments.
  - It's also possible to create CBs from (common) runtime arguments, for them we use `cb_arg_{i}` where `i` is assigned sequentially as we walk the `funcOp`.
- Lowering kernel arguments:
  - Replaced `TTKernelToEmitCGetCompileArgValRewriter` with a more general `TTKernelToEmitCArgValRewriter` that handles all three types of kernel args.
  - For CB args, it attaches an internal attribute to indicate the index of the argument, preparing for naming the CB objects later.
- Lowering to OO CB API calls:
  - Lazily create the `experimental::CircularBuffer cb_xxx_n(...)` declaration once per block.
    - The compute API only uses CB IDs so there's no need to create a CB object if there's no CB method invocation.
    - Perform dominance analysis so same-port repeated SSA values and loop/cross-block uses reuse a single dominating `experimental::CircularBuffer` declaration.
  - `TTKernelToEmitCCBVoidMethodRewriter` handles void-return API calls like `ttkernel.cb_push_back -> cb_obj.push_back(...)`.
    - Use the `emitc::VerbatimOp` because these are statement-like method calls and is the most direct way to preserve the exact OO spelling we want.
  - `TTKernelToEmitCCBResultMethodRewriter` handles API calls that return values like `ttkernel.get_write_ptr -> cb_obj.get_write_ptr()`.
    - These are emitted as `emitc::LiteralOp` because the result participates in later expressions and needs to behave like a value.
- The `ttkernle::CBType -> !emitc.opaque<"::tt::CB">` lowering is untouched for now to simplify things (DFB migration will require integer type).
- CB header inclusion now scans `emitc::VerbatimOp` values and includes the experimental header when it sees `experimental::CircularBuffer`.
  - The old `cb_api.h` is removed except from the experimental tilize/untilize kernels.
- Delete the `GetArgValOp/GetCommonArgValOp` branches from `TTKernelToEmitCOpaqueRewriter::getTemplateArgs`.
- Various lit test updates & additions.

### Checklist

### DIFF
--- a/lib/Conversion/TTKernelToEmitC/TTKernelToEmitC.cpp
+++ b/lib/Conversion/TTKernelToEmitC/TTKernelToEmitC.cpp
@@ -28,7 +28,6 @@
 #include "llvm/ADT/StringRef.h"
 #include "llvm/Support/raw_ostream.h"
 
-#include <cctype>
 #include <string>
 
 using namespace mlir;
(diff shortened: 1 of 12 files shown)


lib/Conversion/TTKernelToEmitC/TTKernelToEmitC.cpp
```diff
@@ -106,6 +106,111 @@ datatypeToDataformatEnumValueOpaqueAttr(Builder &builder,
   return builder.getType<emitc::OpaqueAttr>(expression.c_str());
 }
 
+static std::string getCBName(Value cb) {
```
vwellsTT, line 109: nit: debatably should use LLVM string type instead

lib/Conversion/TTKernelToEmitC/TTKernelToEmitC.cpp
```diff
@@ -106,6 +106,111 @@ datatypeToDataformatEnumValueOpaqueAttr(Builder &builder,
...
+      cbIdxAttr = attr;
+    }
+  }
+
+  TT_assertv(cbIdxAttr, "CB value must have a stable lowering name");
+  return prefix + std::to_string(cbIdxAttr.getInt());
+}
+
+// Assign deterministic names to non-compile-time CB arguments before dialect
+// conversion. Runtime and common runtime arguments use their own common index
```
vwellsTT, line 130: what's a "common runtime" argument vs runtime?

lib/Conversion/TTKernelToEmitC/TTKernelToEmitC.cpp
```diff
@@ -572,20 +669,144 @@ class TTKernelBitcastOpRewriter
...
+  LogicalResult
+  matchAndRewrite(SourceOp op, typename SourceOp::Adaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const final {
+    auto operands = adaptor.getOperands();
+    assert(!operands.empty());
+
+    std::string callStr =
+        ensureCBDeclaration(operands.front(), op.getOperation(), rewriter) +
+        "." + methodName + "(";
+    for (size_t i = 1; i < operands.size(); ++i) {
```
vwellsTT: not sure I follow why this starts at 1 instead of 0

lib/Conversion/TTKernelToEmitC/TTKernelToEmitC.cpp
```diff
@@ -1322,33 +1545,30 @@ class ConvertTTKernelToEmitCPass
     populateMemRefToEmitCTypeConversion(typeConverter);
     populateMemRefToEmitCConversionPatterns(patterns, typeConverter);
 
+    // clang-format off
```
vwellsTT: nit: I think it's better not to do this so broadly

vwellsTT (approved):

lib/Conversion/TTKernelToEmitC/TTKernelToEmitC.cpp
wenbinlyuTT: These methods in the TTKernel dialect is still the old style `(cb_id, num_tiles)`, so the first operand is not needed in the OO-style API. But yeah this loop is redundant since there's only one arg left.

lib/Conversion/TTKernelToEmitC/TTKernelToEmitC.cpp
wenbinlyuTT: Runtime kernel arg can be set per-core, common runtime arg is the same for all cores (per-dispatch).

lib/Conversion/TTKernelToEmitC/TTKernelToEmitC.cpp
wenbinlyuTT: I'm constructing a new string inside the function, so StringRef wouldn't work, StringLiteral is really only meant to represent constexpr strings, and SmallString requires a length template argument which I don't want to set to some arbitrary value. So I guess we can just trust libstdc++'s small string optimization.

nsmithtt (approved):

PR #8199, opened 2026-04-30
[d2m] Move reblocking before memory planner so that aliasing decisions don't change post reblocking

sohaibnadeemTT:
### Checklist

### DIFF
--- a/lib/Dialect/D2M/Transforms/SplitUnifiedThread.cpp
+++ b/lib/Dialect/D2M/Transforms/SplitUnifiedThread.cpp
@@ -29,6 +29,15 @@ namespace {
 // when the remote memref has a view layout, is in DRAM, or the local buffer is
 // a streaming CB (CBLayoutAttr). Aliased ops do not need DMA and return false.
 static bool needsDMA(Value memref, Value localBuffer) {
+  // Check if the local buffer is a streaming CB.
+  if (localBuffer) {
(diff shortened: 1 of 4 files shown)


lib/Dialect/D2M/Transforms/SplitUnifiedThread.cpp
```diff
@@ -29,6 +29,15 @@ namespace {
 // when the remote memref has a view layout, is in DRAM, or the local buffer is
 // a streaming CB (CBLayoutAttr). Aliased ops do not need DMA and return false.
 static bool needsDMA(Value memref, Value localBuffer) {
+  // Check if the local buffer is a streaming CB.
```
vtangTT, line 32: just curious why we need to move this check earlier?

sohaibnadeemTT: This check is basically there to ensure the allocator and dma lowering don't disagree on aliasing. If allocator decides not to alias, it will tag the alloc with a CBLayoutAttr and that alloc gets hoisted (in HoistCBAllocs) and becomes a real CB (in runtime lowering). Therefore, it is actually too late to choose to alias here if CBLayoutAttr is already present and we must return true. This check was at the bottom of the function but should really checked first so should be at the top.

vtangTT (approved):

anuragsingh-tt (approved):

PR #8203, opened 2026-04-30
[d2m] Move reblocking before memory planner so that aliasing decisions don't change post reblocking

sohaibnadeemTT:
### Ticket
Link to Github Issue

### Problem description

### What's changed

### Checklist

### DIFF
--- a/lib/Dialect/D2M/Transforms/Allocate.cpp
+++ b/lib/Dialect/D2M/Transforms/Allocate.cpp
@@ -384,6 +384,10 @@ class D2MAllocate final : public impl::D2MAllocateBase<D2MAllocate> {
       return failure();
     }
 
+    if (failed(reblockGenerics(funcOp, analysis))) {
+      return failure();
+    }
+
     if (failed(analyzeGenericRegionAllocs(funcOp, analysis))) {
       return failure();
     }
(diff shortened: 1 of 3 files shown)


lib/Dialect/D2M/Transforms/Allocate.cpp
```diff
@@ -1215,16 +1215,18 @@ class D2MAllocate final : public impl::D2MAllocateBase<D2MAllocate> {
...
       Value newOutput = reblocked->returnView.getResult();
 
       // Move sequencing metadata to the new anchor op produced by the rewrite.
       SequenceT sequencePosition = analysis.sequencing[oldGenericOp];
-      analysis.sequencing.positionMap[sequencePosition] = sequenceAnchor;
+      analysis.sequencing.positionMap[sequencePosition] =
+          reblocked->genericOp.getOperation();
       analysis.sequencing.operationMap.erase(oldGenericOp.getOperation());
-      analysis.sequencing.operationMap[sequenceAnchor] = sequencePosition;
+      analysis.sequencing.operationMap[reblocked->genericOp.getOperation()] =
```
anuragsingh-tt, line 1225: So the sequenceAnchor is no longer the returnView on the reblocked generic? Is this needed? The sequencing is also used for deallocation so I am wondering if that could yield a situation where we memref.dealloc in between the generic and the return view that relies on that buffer. Maybe a targeted test for this?

Issue #8205, opened 2026-04-30
Proper fix for reusing global semaphores in capture run


mvasiljevicTT:
Before [this PR](https://github.com/tenstorrent/tt-mlir/pull/8099), models with ops that use global semaphores could not be used with trace, because semaphores were created again in the capture run.

That was quick fixed on that PR by giving both `parentContext` to each child, enabling them to have one global semaphore storage.

Cleaner way to do this is to propagate global semaphores the same way as input tensors, through existing `globalSemaphorePool`.

mvasiljevicTT:
Perf benchmark on branch rebased on current mlir uplift in tt-xla https://github.com/tenstorrent/tt-xla/actions/runs/25159365133

mvasiljevicTT:
During this effort, `DistributedRMSNormWidthShardInputRewritePattern.cpp` got unreasonably complicated for one workaround. It's tricky to divide adding semaphores as a new workaround as it is dependent on op having final core range.

We should consider moving semaphores into a separate pass that will come after optimizer, especially if this is something we expect more ops to have in the future (not sure yet if that is the case)

Original rewrite pattern should only have sharding and tilizing, so that it can be safely disabled when optimizer is used, and optimizer should be aware of all of those constraints.

Weight reshape should be moved to decomposition, next to activation reshape. 

Buffer initialization (empty op) should also be moved to other pass, as it is not intuitive to be there. I have a hunch that same ops will need explicit global semaphores and persistent buffers so that might be one pass. 

We should consider using this for CCLs too, we should investigate if there would be perf benefit of reusing semaphores and buffers if they were created on ttnn ir level instead of metal, as we would have more control of them.

mvasiljevicTT:
Switching to [mvasiljevic/test-global-sem-pool-pass](https://github.com/tenstorrent/tt-mlir/tree/mvasiljevic/test-global-sem-pool-pass) - [perf benchmark](https://github.com/tenstorrent/tt-xla/actions/runs/25446831084)

added separate passes for buffers and semaphores that could be later used for async ccls

now all distributed rms norm ops in one decode step use same buffer and semaphore

mvasiljevicTT:
To my understanding of trace, both buffers and semaphores stay in l1 through all decode steps. Since they all use same one buffer and one semaphore this is not a problem.

For ccls, each shape will need one buffer, this might be a problem for the optimizer as some part of l1 will be occupied all the time

mvasiljevicTT:
[...]

PR #8210, opened 2026-04-30
[Runtime][TTNN] Reuse global semaphores in capture run through the global semaphore pool

mvasiljevicTT:
### Ticket
[Link to Github Issue](https://github.com/tenstorrent/tt-mlir/issues/8205)

### Problem description
Ops that use global semaphores (like `distributed_rms_norm`) didn’t work with trace because the capture run created new semaphores instead of reusing the ones from warmup. 

This [commit](https://github.com/tenstorrent/tt-mlir/commit/5b85073695682d062a0ac7fe5888bfb5b410853d) fixed this as a quick unblock by giving each child `ProgramContext` a `parentContext` pointer and making the runtime op create the semaphore on warmup and cache it on the parent where capture run could later find it.

Cleaner way is to create them with `ttnn.create_global_semaphore` op, propagate to subprogram calls and trace the same way as input tensors, and resolve at runtime through existing global semaphore pool.

### What's changed

- `distributed_rms_norm` now takes an explicit `ttnn.global_semaphore` operand at the MLIR level, so the runtime dependency is visible in IR instead of being implicit. The workaround that lowers the op leaves the operand as `nullptr`; materialization is delegated to a dedicated allocation pass that runs after optimizer.

- The `DistributedRMSNormWidthShardInputRewritePattern` workaround is now strictly a layout/sharding workaround — it leaves `stats` (buffer) and `semaphore` operands as `nullptr` on the rebuilt op and delegates their materialization to the allocation passes.

- New `TTNN_DistributedOpInterface` for composite multi-device ops that need explicit scratch buffers and/or global semaphores in IR. The interface exposes `hasUnboundBuffers` / `allocateBuffers` and `hasUnboundSemaphores` / `allocateSemaphores`; implementations are idempotent and create their resources in the function prelude (right after `ttnn.get_device`) so trace hoisting stays clean. Currently only used for `DistributedRMSNormOp`, but will be used for future ccls, like `all_gather_async` and `reduce_scatter_async`

- Two dedicated passes drive allocation through the interface:
  - `TTNNAllocateDistributedOpBuffers` runs **before** the optimizer, so scratch buffers contribute to L1 budgeting (currently they are ignored, but that will be fixed separately). A canonicalize + CSE pair runs immediately after it to dedupe identical `EmptyOp`s across distributed ops. This way buffer of every dimension is allocated once and reused for all the ops that need it. This is ok for distributed rms norm as it doesn't care for the content of the buffer before use, when we come into ops that do this will be handled differently.
  - `TTNNAllocateDistributedOpSemaphores` runs **after** the optimizer (semaphore core range is derived from the finalized input shard spec) and **before** trace hoisting (semaphore SSA values must be visible at the trace boundary). Semaphores are not CSE'd because it was breaking resnet test, and it would not be contributing much as semaphores are small anyway and it would rely on op correctly resetting before and after usage.

- `ttnn.create_global_semaphore` is marked as a creation op (`TTCoreCreationOpTrait` + `NonCacheableTrait`), so the trace-hoist pass keeps it outside the trace region.

- Global semaphores flow through `func.call` and `ttnn.capture_or_execute_trace` the same way input tensors do, distinguished by their `GlobalSemaphoreType`:
  - For `func.call`, semaphores ride as ordinary SSA operands; the flatbuffer serializer partitions them into a separate `semaphore_inputs: [GlobalSemaphoreRef]` field on `FuncCallOp` (mirroring how tensor operands become `inputs: [TensorRef]`).
  - `ttnn.capture_or_execute_trace` gets an explicit `semaphore_inputs` variadic operand.
  - The trace-hoist pass scans inputs by SSA type and forwards semaphores into the trace function (as block args of `GlobalSemaphoreType`) and through the wrapping `func.call` / `CaptureOrExecuteTraceOp`.
  - At the program (function) level, function arguments of `GlobalSemaphoreType` are surfaced as `program.semaphore_inputs` in the flatbuffer.

- At runtime, semaphores are serialized as `GlobalSemaphoreRefs` and resolved through the existing per-`ProgramContext` global semaphore pool, the same way input tensors already use the tensor pool. When entering a subprogram or trace, the parent seeds the child's pool from the inputs it received, so capture and replay reuse the same semaphore objects.

### Checklist
- [ ] [perf benchmark for llama](https://github.com/tenstorrent/tt-xla/actions/runs/25312821286)

### DIFF
--- a/lib/Dialect/TTNN/IR/TTNNOps.cpp
+++ b/lib/Dialect/TTNN/IR/TTNNOps.cpp
@@ -14,7 +14,6 @@
 #include "ttmlir/Dialect/TTNN/IR/TTNNOpsInterfaces.cpp.inc"
 #include "ttmlir/Dialect/TTNN/IR/TTNNOpsTypes.h"
 #include "ttmlir/Dialect/TTNN/Types/Types.h"
-#include "ttmlir/Dialect/TTNN/Utils/OptimizerUtils.h"
 #include "ttmlir/Dialect/TTNN/Utils/TransformUtils.h"
 #include "ttmlir/Dialect/TTNN/Utils/Utils.h"
 #include "ttmlir/Dialect/TTNN/Utils/VerificationUtils.h"
@@ -3152,33 +3151,26 @@ void mlir::tt::ttnn::DistributedRMSNormOp::allocateBuffers(
   ttcore::DataType statsDataType =
       fp32DestAccEn ? ttcore::DataType::Float32 : ttcore::DataType::BFloat16;
 
-  auto physicalGrid =
-      ttcore::getCurrentScopeSystemDesc(*this).getChipDescs()[0].getGrid();
-  auto [virtToPhysicalMap, physicalToVirtMap] =
-      optimizer_utils::createSingleDeviceVirtualToPhysicalAffineMaps(
-          rewriter.getContext(), TensorMemoryLayout::WidthSharded,
-          physicalGrid);
+  ttcore::DeviceAttr deviceAttr = ttcore::lookupDevice(*this);
 
   // One tile (32x32) per device, width-sharded on core (0,0) in L1. The fused
   // kernel writes partial RMS statistics here and exchanges them across
   // devices via the all-gather.
-  SmallVector<int64_t> statsGridShape = {1, 1};
-  auto statsGrid = ttcore::GridAttr::get(rewriter.getContext(), statsGridShape,
-                                         virtToPhysicalMap, physicalToVirtMap);
-  auto statsMemLayoutAttr = TensorMemoryLayoutAttr::get(
-      rewriter.getContext(), TensorMemoryLayout::WidthSharded);
-
   SmallVector<int64_t> statsShape = {1, 1, 32, 32};
+  SmallVector<int64_t> statsGridShape = {1, 1};
   TTNNLayoutAttr statsLayout =
-      TTNNLayoutAttr::get(rewriter.getContext(), statsShape,
-                          ttcore::TileType::get(statsElementType),
-                          BufferType::L1, statsGrid, statsMemLayoutAttr);
+      TTNNLayoutAttr::Builder(rewriter.getContext(), statsShape,
+                              ttcore::TileType::get(statsElementType))
+          .setBufferType(BufferType::L1)
+          .setMemoryLayout(TensorMemoryLayout::WidthSharded)
+          .setGridShape(statsGridShape)
+          .buildWithCanonicalCorePlacement(deviceAttr);
 
   auto statsShapeAttr = ShapeAttr::get(rewriter.getContext(), statsShape);
   auto statsDtypeAttr =
       ttcore::DataTypeAttr::get(rewriter.getContext(), statsDataType);
   auto statsLayoutAttr = LayoutAttr::get(rewriter.getContext(), Layout::Tile);
-  auto statsMemConfig = MemoryConfigAttr::get(statsLayout, statsGrid);
+  auto statsMemConfig = MemoryConfigAttr::get(statsLayout);
 
   RankedTensorType statsResultType =
       RankedTensorType::get(statsShape, statsElementType, statsLayout);
--- a/lib/Dialect/TTNN/IR/TTNNOps.cpp
+++ b/lib/Dialect/TTNN/IR/TTNNOps.cpp
@@ -3126,7 +3126,7 @@ static ::mlir::LogicalResult verifyTTNNBatchNormOp(OpType op) {
 }
 
 bool mlir::tt::ttnn::DistributedRMSNormOp::hasUnboundBuffers() {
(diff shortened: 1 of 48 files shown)


mvasiljevicTT:
Major changes, updated the description. 

Llama 70b on perf benchmarks currently produces random outputs on main, not merging until that can be properly tested.

lib/Dialect/TTNN/IR/TTNNOps.cpp
```diff
@@ -3130,6 +3126,124 @@ ::mlir::LogicalResult mlir::tt::ttnn::DistributedRMSNormOp::verify() {
...
+  // devices via the all-gather.
+  SmallVector<int64_t> statsShape = {1, 1, 32, 32};
+  SmallVector<int64_t> statsGridShape = {1, 1};
+  TTNNLayoutAttr statsLayout =
+      TTNNLayoutAttr::Builder(rewriter.getContext(), statsShape,
+                              ttcore::TileType::get(statsElementType))
+          .setBufferType(BufferType::L1)
+          .setMemoryLayout(TensorMemoryLayout::WidthSharded)
+          .setGridShape(statsGridShape)
+          .buildWithCanonicalCorePlacement(deviceAttr);
```
dmilinkovicTT: Nit: `buildWithCanonicalCorePlacement` will indeed place the only shard on core `(0, 0)`, but it might be better to explicitly populate the placement by using `.setCoreRangeSet(crs).build()`.

The end result will be the same, but I think it's better to be explicit in this case, since the shard must be allocated on core `(0, 0)`.

This part of the code was indeed written by me in the workaround pattern - seems I didn't catch the issue back then. 😄

dmilinkovicTT (approved):

runtime/lib/ttnn/operations/normalization/distributed_rms_norm.cpp
```diff
@@ -65,16 +65,11 @@ void run(const ::tt::target::ttnn::DistributedRMSNormOp *op,
...
-  auto semaphore = context.getOrCreateImplicitGlobalSemaphore(
-      reinterpret_cast<uintptr_t>(op), [&]() {
-        return ::ttnn::global_semaphore::create_global_semaphore(
-            &meshDevice, shardSpec->grid, 0);
-      });
+  LOG_ASSERT(op->semaphore(),
+             "DistributedRMSNormOp expects an explicit global semaphore");
+  ::ttnn::GlobalSemaphore semaphore =
+      context.getGlobalSemaphorePool().getTTNNGlobalSemaphoreAndValidate(
+          op->semaphore());
```
svuckovicTT, line 72: Please update ttnntoemitpy.cpp and ttnntoemitc.cpp

mvasiljevicTT: Updated now, tested locally for ttnn-to-emitpy, but I can't currently test full ttir-to-emitpy pipeline because `ttnn.create_global_semaphore` is not supported. Opened issue for that: https://github.com/tenstorrent/tt-mlir/issues/8464

lib/Dialect/TTNN/IR/TTNNOps.cpp
mvasiljevicTT: thanks, done

runtime/lib/ttnn/operations/normalization/distributed_rms_norm.cpp
mvasiljevicTT: emitpy/emitc builder tests for distributed rms norm skipped now for this reason too

runtime/lib/ttnn/program_executor.cpp
```diff
@@ -155,10 +155,31 @@ ProgramExecutor::ProgramExecutor(
...
+  size_t expectedSemaphoreInputs =
+      program->semaphore_inputs() ? program->semaphore_inputs()->size() : 0;
+  LOG_ASSERT(programSemaphoreInputs.size() == expectedSemaphoreInputs,
+             "Program semaphore input size mismatch: ", expectedSemaphoreInputs,
+             " != ", programSemaphoreInputs.size());
+  if (program->semaphore_inputs()) {
+    size_t i = 0;
+    for (const ::tt::target::ttnn::GlobalSemaphoreRef *semaphoreRef :
+         *program->semaphore_inputs()) {
+      auto semaphorePtr = std::make_shared<::ttnn::GlobalSemaphore>(
```
pilkicTT: i think we could simplify this loop by having `collectSemaphoreInputs` return our runtime semaphore object, i.e. `::tt::runtime::GlobalSemaphore` and accept `programSemaphoreInputs` as a vector of those as well.

pilkicTT (approved):

runtime/lib/ttnn/program_executor.cpp
mvasiljevicTT: thanks, changed now

PR #8211, opened 2026-04-30
[build] Support user-managed tt-metal checkout via TTMLIR_TTMETAL_SOURCE_DIR

nsmithtt:
Add a TTMLIR_TTMETAL_SOURCE_DIR CMake cache variable. When set, the build uses the provided tt-metal checkout instead of having ExternalProject clone tt-metal at the pinned TT_METAL_VERSION. This lets developers iterate against a local tt-metal tree without fighting ExternalProject's download/update steps. TT_METAL_VERSION is not enforced when a user-managed checkout is supplied.

Replaces hardcoded third_party/tt-metal/src/tt-metal paths in third_party/CMakeLists.txt with a single TTMETAL_SOURCE_DIR that resolves to either the user-supplied path or the ExternalProject location.

### DIFF
--- a/third_party/CMakeLists.txt
+++ b/third_party/CMakeLists.txt
@@ -2,6 +2,41 @@ include(ExternalProject)
 
 set(TT_METAL_VERSION "2a9cc7c75378056aed986050654f52e8d4af97ad")
 
+set(TTMLIR_TTMETAL_SOURCE_DIR "" CACHE PATH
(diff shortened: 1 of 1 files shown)


jameszianxuTT:
That's a cool feature. I think this should be mentioned in the docs somewhere

anuragsingh-tt:
Yes! I support this feature for frontends too! It makes co-development easier.

vwellsTT (approved):

PR #8213, opened 2026-04-30
Added a case for rightshift

aorlovicTT:
### Ticket
Related to https://github.com/tenstorrent/tt-mlir/pull/8160

### Problem description

### What's changed
Added a case where every shift amount of RHS is greater then or equal to the one on LHS, so it just splats to 0.

### Checklist


amalkovTT:
Thanks!

PR #8215, opened 2026-04-30
[d2m] argmax

dlokeTT:
### Problem description
Need to support this op

### What's changed
- Lowering supports last-dimension argmax with bf16, f32, and i32 inputs.
- Added new `d2m.argmax` and `d2m.argmax_interleaved` to represent this dm op in tensor and memref land respectively.
- Implemented TTKernel lowering as a data-movement row scan:
      - reads packed bf16/f32/i32 values from L1 through scratch,
      - compares candidates (cmpf ogt for floats, signed cmpi sgt for i32),
      - writes the selected column index as i32.
  - Added D2M bufferization, memory effects, verifier checks, thread-arg normalization, and scratch-buffer handling for argmax.
- Added builder and lit tests

### Checklist

### DIFF
--- a/include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
+++ b/include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
@@ -1649,6 +1649,74 @@ def D2M_IndexedRowCopyOp : D2M_GenericRegionDatamovementOp<"indexed_row_copy",
     let hasVerifier = 1;
 }
 
+def D2M_ArgMaxInterleavedOp : D2M_GenericRegionDatamovementOp<"argmax_interleaved",
+  [ DeclareOpInterfaceMethods<MemoryEffectsOpInterface> ]> {
+    let summary = "D2M interleaved argmax op";
+    let description = [{
+      Data-movement primitive for computing argmax over the last logical
+      dimension of a row-major tensor. The op reads source elements from L1,
+      scans each row on the data-movement core, and writes 32-bit indices to
+      the output buffer.
+
+      The op is memref-only. Tensor-level semantic ops should bufferize to this
+      op after choosing a scratch circular buffer.
+    }];
+
+    let arguments = (ins
+      DeviceL1MemRef:$input,
+      DeviceL1MemRef:$output,
+      DeviceL1MemRef:$scratch,
+      I64Attr:$numRows,
+      I64Attr:$rowWidth,
+      DenseI64ArrayAttr:$rowShape
+    );
+
+    let assemblyFormat = [{
+      $input `,` $output `scratch` $scratch
+      `<` $numRows `,` $rowWidth `>`
+      attr-dict `:` type($input) `,` type($output) `,` type($scratch)
+    }];
+
+    let hasVerifier = 1;
+}
+
+def D2M_ArgMaxOp : D2M_GenericRegionDatamovementOp<"argmax",
+  [ DeclareOpInterfaceMethods<MemoryEffectsOpInterface>
+  , DeclareOpInterfaceMethods<BufferizableOpInterface, [
+      "bufferizesToMemoryRead",
+      "bufferizesToMemoryWrite",
+      "bufferize",
+      "getAliasingValues",
+      "getBufferType",
+      "hasTensorSemantics"
+    ]>
+  ]> {
+    let summary = "D2M argmax op";
+    let description = [{
+      Tensor-level data-movement argmax over the last logical dimension. During
+      bufferization, this op allocates a scratch circular buffer and lowers to
+      `d2m.argmax_interleaved`, which owns the backend dataflow implementation.
+    }];
+
+    let arguments = (ins
+      AnyRankedTensor:$input,
+      AnyRankedTensor:$output,
+      I64Attr:$numRows,
+      I64Attr:$rowWidth,
+      DenseI64ArrayAttr:$rowShape
+    );
+    let results = (outs AnyRankedTensor:$result);
+
+    let assemblyFormat = [{
+      $input `,` $output
+      `<` $numRows `,` $rowWidth `>`
+      attr-dict `:` type($input) `,` type($output)
+      `->` type($result)
+    }];
+
+    let hasVerifier = 1;
+}
+
 def D2M_EmbeddingOp : D2M_GenericRegionDatamovementOp<"embedding",
   [ DeclareOpInterfaceMethods<MemoryEffectsOpInterface>
   , DeclareOpInterfaceMethods<BufferizableOpInterface, [
--- a/lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
+++ b/lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
@@ -2585,52 +2585,44 @@ public:
 } // namespace
 
 namespace {
-static bool isI32ElementType(Type type) {
-  auto intType = mlir::dyn_cast<IntegerType>(type);
-  return intType && intType.getWidth() == 32;
+static Value compareArgMaxCandidates(OpBuilder &rewriter, Location loc,
+                                     Value candidate, Value currentMax) {
+  Type candidateType = candidate.getType();
+  if (mlir::isa<FloatType>(candidateType)) {
+    // Ordered greater-than keeps the existing maximum if either operand is NaN.
+    return rewriter.create<arith::CmpFOp>(loc, arith::CmpFPredicate::OGT,
+                                          candidate, currentMax);
+  }
+  if (d2m::utils::isI32ElementType(candidateType)) {
+    return rewriter.create<arith::CmpIOp>(loc, arith::CmpIPredicate::sgt,
+                                          candidate, currentMax);
+  }
+  llvm_unreachable("Unsupported argmax candidate type");
 }
 
-static Value loadArgMaxElement(OpBuilder &rewriter, Location loc,
-                               Value scratchCb, Value inputBase,
-                               AffineMap inputMemoryMap, MemRefType inputType,
-                               Value row, Value column,
-                               ttcore::ChipDescAttr chipDesc) {
-  Type elementType = inputType.getElementType();
+static Value loadArgMaxElementFromPacket(OpBuilder &rewriter, Location loc,
+                                         Value l1Ptr, Type elementType,
+                                         Value packetLane) {
   bool isBF16 = mlir::isa<BFloat16Type>(elementType);
   bool isF32 = mlir::isa<Float32Type>(elementType);
-  bool isI32 = isI32ElementType(elementType);
+  bool isI32 = d2m::utils::isI32ElementType(elementType);
   assert((isBF16 || isF32 || isI32) &&
          "argmax supports f32, bf16, and i32 inputs only");
 
-  Value packetColumn;
   Value wordLane;
   Value halfLane = index(rewriter, loc, 0);
   if (isBF16) {
-    Value eight = index(rewriter, loc, 8);
     Value two = index(rewriter, loc, 2);
-    Value packetIndex = rewriter.create<arith::DivSIOp>(loc, column, eight);
-    packetColumn = rewriter.create<arith::MulIOp>(loc, packetIndex, eight);
-    Value packetOffset =
-        rewriter.create<arith::SubIOp>(loc, column, packetColumn);
-    wordLane = rewriter.create<arith::DivSIOp>(loc, packetOffset, two);
+    wordLane = rewriter.create<arith::DivSIOp>(loc, packetLane, two);
     Value wordBase = rewriter.create<arith::MulIOp>(loc, wordLane, two);
-    halfLane = rewriter.create<arith::SubIOp>(loc, packetOffset, wordBase);
+    halfLane = rewriter.create<arith::SubIOp>(loc, packetLane, wordBase);
   } else {
-    Value four = index(rewriter, loc, 4);
-    Value packetIndex = rewriter.create<arith::DivSIOp>(loc, column, four);
-    packetColumn = rewriter.create<arith::MulIOp>(loc, packetIndex, four);
-    wordLane = rewriter.create<arith::SubIOp>(loc, column, packetColumn);
+    wordLane = packetLane;
   }
 
-  SmallVector<Value> inputLogicalIndices =
-      getCollapsed2DElementIndices(rewriter, loc, inputType, row, packetColumn);
-  Value inputNocAddr = buildMappedL1NocAddr(
-      rewriter, loc, inputBase, inputMemoryMap, inputLogicalIndices, chipDesc);
   Value wordLaneI32 =
       rewriter.create<arith::IndexCastOp>(loc, rewriter.getI32Type(), wordLane);
-  Value word = loadI32FromL1PacketThroughScratch(
-      rewriter, loc, scratchCb, inputNocAddr, wordLaneI32,
-      i32(rewriter, loc, 16), i32(rewriter, loc, 1));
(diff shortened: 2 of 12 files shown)


lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
```diff
@@ -2584,6 +2584,233 @@ class D2MIndexedRowCopyRewriter
...
+  Value bf16Value = rewriter.create<ttkernel::BitcastOp>(
+      loc, rewriter.getBF16Type(), bf16Bits);
+  return rewriter.create<arith::ExtFOp>(loc, rewriter.getF32Type(), bf16Value);
+}
+
+static Value compareArgMaxCandidates(OpBuilder &rewriter, Location loc,
+                                     Value candidate, Value currentMax) {
+  Type candidateType = candidate.getType();
+  if (mlir::isa<FloatType>(candidateType)) {
+    return rewriter.create<arith::CmpFOp>(loc, arith::CmpFPredicate::OGT,
```
vtangTT: curious how OGT handles NaN values or do we not care for this case?

lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
```diff
@@ -2584,6 +2584,233 @@ class D2MIndexedRowCopyRewriter
...
+    Value rowEndsBeforeTensorEnd = rewriter.create<arith::CmpIOp>(
+        loc, arith::CmpIPredicate::slt, tentativeEndRow, numRows);
+    Value endRow = rewriter.create<arith::SelectOp>(loc, rowEndsBeforeTensorEnd,
+                                                    tentativeEndRow, numRows);
+
+    Value startColumn = rewriter.create<arith::MulIOp>(
+        loc, myX, index(rewriter, loc, outputShardShape[1]));
+    Value tentativeEndColumn = rewriter.create<arith::AddIOp>(
+        loc, startColumn, index(rewriter, loc, outputShardShape[1]));
+    Value outputWidth = index(rewriter, loc, 1);
```
vtangTT: is it possible we pull the outputWidth from `outputShardShape[1]` instead of hard coding to 1 or am i misunderstanding here?

lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
```diff
@@ -2584,6 +2584,233 @@ class D2MIndexedRowCopyRewriter
...
+
+  if (isF32) {
+    return rewriter.create<ttkernel::BitcastOp>(loc, rewriter.getF32Type(),
+                                                word);
+  }
+  if (isI32) {
+    return word;
+  }
+
+  Value lowBits =
```
vtangTT, line 2638: can we get a comment explaining what's going on here? has smthn to do w/ endianess?

lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
```diff
@@ -2584,6 +2584,233 @@ class D2MIndexedRowCopyRewriter
 };
 } // namespace
 
+namespace {
+static bool isI32ElementType(Type type) {
```
vtangTT: this helper is duplicated in `D2MToTTKernel`, `D2MGenericRegionOps` and also inlined into the `isSupportedArgMaxInputElementType` helper in `TTIRToD2M`.. can we pull it out into utils soemwhere

lib/Dialect/D2M/IR/D2MGenericRegionOps.cpp
```diff
@@ -290,6 +291,179 @@ void IndexedRowCopyOp::getEffects(
...
+//===----------------------------------------------------------------------===//
+// ArgMax operations
+//===----------------------------------------------------------------------===//
+
+static bool isI32ElementType(Type type) {
+  auto intType = mlir::dyn_cast<IntegerType>(type);
+  return intType && intType.getWidth() == 32;
+}
+
+static bool isSupportedArgMaxInputElementType(Type type) {
```
vtangTT: duplicated in `TTIRToD2M` it seems, can we pull it out into utils somewhere

lib/Dialect/D2M/IR/D2MGenericRegionOps.cpp
```diff
@@ -290,6 +291,179 @@ void IndexedRowCopyOp::getEffects(
...
+    return ttcore::getBufferType(value.getType(), /*isView=*/false);
+  }
+  return mlir::failure();
+}
+
+bool ArgMaxOp::hasTensorSemantics() { return true; }
+
+static MemRefType createArgMaxScratchType(MLIRContext *ctx) {
+  auto scratchElementType = IntegerType::get(ctx, 32);
+  return createIndexedRowCopyScratchType(ctx, {1, kArgMaxScratchPageElements},
```
vtangTT: nit: is it possible to rename helper, kinda confusing that argmax is using a indexrowcopy op helper

lib/Dialect/D2M/IR/D2MGenericRegionOps.cpp
```diff
@@ -199,6 +199,7 @@ void DMAWriteOp::getEffects(
 //===----------------------------------------------------------------------===//
 
 static constexpr int64_t kIndexedRowCopyScratchPageElements = 1024;
+static constexpr int64_t kArgMaxScratchPageElements = 1024;
```
vtangTT, line 205: nit: comment to document why its 1024

lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
```diff
@@ -2584,6 +2584,233 @@ class D2MIndexedRowCopyRewriter
...
+
+        auto scanLoop = rewriter.create<scf::ForOp>(
+            loc, index(rewriter, loc, 1),
+            index(rewriter, loc, op.getRowWidth()), oneIndex,
+            ValueRange{initialMax, initialIndex},
+            [&](OpBuilder &builder, Location bodyLoc, Value column,
+                ValueRange iterArgs) {
+              Value currentMax = iterArgs[0];
+              Value currentIndex = iterArgs[1];
+              Value candidate = loadArgMaxElement(
```
vtangTT: To confirm, here for each iteration we are doing 1 NoC read of 16B through `loadI32FromL1PacketThroughScratch` helper? Or am I missing smthn here. 

This seems very inefficient, could we hoist a larger read above the scan loop?

lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
dlokeTT: It would return false so wouldn't replace but if the first value happened to be NaN it would be that. Added a comment but think this is a pretty rare situation

lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
dlokeTT: yea true we can, updated

lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
dlokeTT: added and yes its about how bf16 values are packed into the packet since `load_from_l1` gives us an i32 word.

lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
dlokeTT: pulled it into utils

lib/Dialect/D2M/IR/D2MGenericRegionOps.cpp
dlokeTT: done

lib/Dialect/D2M/IR/D2MGenericRegionOps.cpp
dlokeTT: yea

lib/Dialect/D2M/IR/D2MGenericRegionOps.cpp
dlokeTT: added

lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
dlokeTT: True, updated to hoist

lib/Dialect/D2M/Transforms/NormalizeThreadArgs.cpp
```diff
@@ -94,6 +94,20 @@ static void rewriteCapturedIndexedRowCopyOperands(IRRewriter &rewriter,
   }
 }
 
+static void rewriteCapturedArgMaxOperands(IRRewriter &rewriter,
```
sohaibnadeemTT, line 97: why is this needed? we should not be be rewriting args for each op separately, there is a general pass starting on line 152 that goes through generic operands and replaces the GetArgOp's result with uses of the generic operand.

include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
```diff
@@ -1649,6 +1649,74 @@ def D2M_IndexedRowCopyOp : D2M_GenericRegionDatamovementOp<"indexed_row_copy",
     let hasVerifier = 1;
 }
 
+def D2M_ArgMaxInterleavedOp : D2M_GenericRegionDatamovementOp<"argmax_interleaved",
```
sohaibnadeemTT, line 1652: why is this a datamovement op?

sohaibnadeemTT (changes requested):
The approach here is pretty hacky. Unless I'm mistaken, we are just using the generic op to wrap the D2M_ArgMaxOp/D2M_ArgMaxInterleavedOp op that only actually gets lowered in D2MToTTKernel, there is no use of data movement or compute primitives prior to that point and so we are just bypassing the entire pipeline up to this point, which is not fundamentally different from just handwriting a TTNN op.  

If we do need this urgently, perhaps we can make argmax a d2m level op not a generic region op and lower that directly in D2MToTTKernel. Since this will not compose nicely with any other generic op and therefore we shouldn't really be wrapping it in a generic just for the sake of it. Same goes for the embedding op recently added.

CC: @nsmithtt @xanderchin

dlokeTT:
Hey Sohaib, definitely agree with some of your points. We kind of are bypassing the entire pipeline because these ops are fundamentally custom kernels and can't really be represented thru the remote load/store and linalg generic flow. They aren't meant to be composed with other ops really. 

The reason for wrapping it in a d2m generic was just to get a lot of the machinery out creating a kernel, operands participating in grid selection and lower to layout and a lot of the plumbing for bufferization/cbs. For embedding we needed it more urgently for a model we are trying to bring online but argmax isn't as urgent. Maybe we should reconvene after tt-deploy about ops like this and how to handle them since there will be more (sort, topk...).

xanderchin:
@dlokeTT @sohaibnadeemTT agree with you both here. I would like to find a balance here, where we can maximally reuse d2m ops. We should also think about creating new building blocks, which lower naturally, that could assist with these sorting style operations.

sohaibnadeemTT:
Yeah from offline discussion with @dlokeTT it seems the fundamental issue here is the lack of argmax on SFPU/FPU which is why we need to use the RISC here. If that is case, we should still be able to use a remote_load + linalg + remote_store flow, except that inside the linalg perhaps we have a `d2m.argmax` op instead of the typical tile ops we use and this would lower into risc code instead of using SFPU/FPU.

include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
```diff
@@ -1649,6 +1649,74 @@ def D2M_IndexedRowCopyOp : D2M_GenericRegionDatamovementOp<"indexed_row_copy",
...
+      op after choosing a scratch circular buffer.
+    }];
+
+    let arguments = (ins
+      DeviceL1MemRef:$input,
+      DeviceL1MemRef:$output,
+      DeviceL1MemRef:$scratch,
+      I64Attr:$numRows,
+      I64Attr:$rowWidth,
+      DenseI64ArrayAttr:$rowShape
```
wenbinlyuTT, line 1671: `leadingShape` or `prefixShape` would be a lot more clear.

lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
```diff
@@ -2584,6 +2584,303 @@ class D2MIndexedRowCopyRewriter
...
+}
+
+static Value loadArgMaxElement(OpBuilder &rewriter, Location loc,
+                               Value scratchCb, Value inputBase,
+                               AffineMap inputMemoryMap, MemRefType inputType,
+                               Value row, Value column,
+                               ttcore::ChipDescAttr chipDesc) {
+  Type elementType = inputType.getElementType();
+  Value elementsPerPacket =
+      index(rewriter, loc, mlir::isa<BFloat16Type>(elementType) ? 8 : 4);
```
wenbinlyuTT, line 2658: This can be replaced by `getNocElementAlignmentL1()` now.

lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
```diff
@@ -2584,6 +2584,303 @@ class D2MIndexedRowCopyRewriter
...
+  assert((isBF16 || isF32 || isI32) &&
+         "argmax supports f32, bf16, and i32 inputs only");
+
+  Value wordLane;
+  Value halfLane = index(rewriter, loc, 0);
+  if (isBF16) {
+    Value two = index(rewriter, loc, 2);
+    wordLane = rewriter.create<arith::DivSIOp>(loc, packetLane, two);
+    Value wordBase = rewriter.create<arith::MulIOp>(loc, wordLane, two);
+    halfLane = rewriter.create<arith::SubIOp>(loc, packetLane, wordBase);
```
wenbinlyuTT, line 2618: Maybe we can extend `LoadFromL1Op` to support types other than i32? After all it's all just
```cpp
volatile Type* data = reinterpret_cast<volatile Type*>(address);
Type value = data[0];
```

lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
```diff
@@ -2584,6 +2584,303 @@ class D2MIndexedRowCopyRewriter
...
+    Value tentativeEndColumn = rewriter.create<arith::AddIOp>(
+        loc, startColumn, index(rewriter, loc, outputShardShape[1]));
+    Value outputWidth = index(rewriter, loc, outputShardShape[1]);
+    Value columnEndsBeforeTensorEnd = rewriter.create<arith::CmpIOp>(
+        loc, arith::CmpIPredicate::slt, tentativeEndColumn, outputWidth);
+    Value endColumn = rewriter.create<arith::SelectOp>(
+        loc, columnEndsBeforeTensorEnd, tentativeEndColumn, outputWidth);
+
+    auto columnLoop =
+        rewriter.create<scf::ForOp>(loc, startColumn, endColumn, oneIndex);
```
wenbinlyuTT, line 2814: This is looping over the output shard's width dim correct? Wouldn't this always has a single iteration for `argmax`, regardless of what `keep_dim` is?
Even with your planned hierarchical tree reduction optimization we will never see more than one column I guess (not sure what happens for `argmax` on other dims).

wenbinlyuTT (commented):
Did a first pass with some suggestions.
My main question is that, since the input's indexing map is set to be `(d0, 0)` and so the rows are all squeezed into the first column of the grid, all the NoC data transfers are therefore unnecessary?
If so, I think hierarchical tree reduction will eliminate the need for this packet-by-packet reduction and make all comparisons local, might as well get rid of all the NoC reads & writes now. In this PR everything is on the `Nx1` grid anyway.

PR #8217, opened 2026-04-30
[Builder] Fix crash on unallocated tensors in execute_fb

VarunKoyyalagunta:
Fix `execute_fb` crashing when `to_host` is called on an unallocated runtime tensor (e.g., produced by const_eval that yields no device tensor). Check `is_allocated()` before calling `to_host`; skip the
  golden comparison for unallocated outputs.

### DIFF
--- a/tools/builder/base/builder_runtime.py
+++ b/tools/builder/base/builder_runtime.py
@@ -855,17 +855,20 @@ def execute_fb(
                     output_device_tensors[device_id].get_data_buffer()
                 )
 
-                ref_tensor = output_device_tensors[device_id]
                 if len(data_buffer) == 0:
                     output_shard_torch = torch.empty(
-                        ref_tensor.get_shape(),
-                        dtype=runtime_dtype_to_torch_dtype(ref_tensor.get_dtype()),
+                        output_device_tensors[device_id].get_shape(),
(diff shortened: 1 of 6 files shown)


vmilosevic (approved):

nsumrakTT (approved):

vvukomanTT (approved):

tools/builder/base/builder_runtime.py
```diff
@@ -721,9 +819,19 @@ def execute_fb(
...
+                # as placeholder and skip to_host entirely.
+                output_host = None
+            else:
+                output_host = tt_runtime.runtime.to_host(
+                    runtime_output_tensor, untilize=True
+                )
+
+            if output_host is None:
+                # Unallocated tensor — nothing to read. Skip golden check.
+                continue
```
jgrimTT: It makes more sense to me to have this in the `if not runtime_output_tensor.is_allocated():` block, but that's just a matter of preference :)

jgrimTT (approved):
One comment, otherwise LGTM, thanks!

tools/builder/base/builder_runtime.py
VarunKoyyalagunta: i concur, modified!

PR #8218, opened 2026-04-30
[Builder] Add slice_dynamic op support

VarunKoyyalagunta:
Add `@parse` handler for `ttir.SliceDynamicOp` in `TTIRBuilder` and register `ttir_slice_dynamic_golden` in `mapping.py`.

### DIFF
--- a/test/python/golden/d2m/test_tms.py
+++ b/test/python/golden/d2m/test_tms.py
@@ -927,7 +927,7 @@ def test_slice(
         def slice_wrapper(
             in0: Operand, builder: TTIRBuilder, unit_attrs: Optional[List[str]] = None
         ):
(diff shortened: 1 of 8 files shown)


jgrimTT (changes requested):
The SliceDynamic golden looks good, we try to avoid adding partial support for ops though as it complicates builder. Could you add full support for the op? This entails adding a tag and split function as well (moved outside of the three ttir.SliceStaticOp functions), adding a test in `test/python/golden/test_ttir_ops.py` and one in `test/python/golden/mlir_snippets`. Also maybe renaming the ttir.SliceStaticOp builder functions from `slice` to `slice_static` for clarity

Issue #7973, opened 2026-04-15
[d2m] ttir-rank-normalization does not update ttir.broadcast dimensions after rank expansion


Aarsh2001:
### Describe the bug

`ttir-rank-normalization` can produce invalid TTIR when it expands low-rank tensors used by `ttir.broadcast`.

The pass correctly promotes low-rank tensor types (for example `tensor<1xf32>` -> `tensor<1x1xf32>` and `tensor<1200xf32>` -> `tensor<1x1200xf32>`), but it does not update the rank-sensitive `ttir.broadcast` attribute `broadcast_dimensions`.

As a result, valid input TTIR becomes invalid after the pass.

### Steps to reproduce the issue

Example

```mlir
module @m {
  ttcore.device_module {
    builtin.module @m {
      func.func public @main(%arg0: tensor<f32>) -> tensor<1200xf32> {
        %0 = "ttir.reshape"(%arg0) <{shape = [1 : i32]}> : (tensor<f32>) -> tensor<1xf32>
        %1 = "ttir.broadcast"(%0) <{broadcast_dimensions = array<i64: 1200>}> : (tensor<1xf32>) -> tensor<1200xf32>
        return %1 : tensor<1200xf32>
      }
    }
  }
}
```

Run:

```
ttmlir-opt \
  --mlir-disable-threading \
  --mlir-print-ir-before-all \
  --mlir-print-ir-module-scope \
  "--ttir-to-ttmetal-pipeline=mock-system-desc-arch=wormhole_b0" \
  <input.ttir.mlir> \
  -o /tmp/out.mlir
```
Output:
```
error: 'ttir.broadcast' op Input tensor rank should match output tensor rank.
```

### Expected behavior

The IR is valid before `ttir-rank-normalization`, but the pass expands the types and leaves the original broadcast_dimensions unchanged.

Conceptually, it rewrites:

```mlir
(tensor<1xf32>) -> tensor<1200xf32>
broadcast_dimensions = [1200]
```

into something like:

```mlir
(tensor<1x1xf32>) -> tensor<1x1200xf32>
broadcast_dimensions = [1200]
```

which is invalid because the broadcast attribute no longer matches the new rank.

Expected behavior:

When `ttir-rank-normalization` prepends leading `1s` to tensor ranks, it should also prepend leading `1s` to `ttir.broadcast's` broadcast_dimensions.

For the example above, after rank normalization the broadcast should become:

```mlir
[...]

PR #8225, opened 2026-05-01
broadcast normalization

anuragsingh-tt:
### Ticket
Solves #7973

### DIFF
--- a/lib/Dialect/TTIR/Transforms/RankNormalization.cpp
+++ b/lib/Dialect/TTIR/Transforms/RankNormalization.cpp
@@ -123,6 +123,8 @@ public:
       updateArangeDimension(arangeOp);
     } else if (auto sliceOp = dyn_cast<ttir::SliceStaticOp>(newOp)) {
       updateSliceStaticAttrs(sliceOp);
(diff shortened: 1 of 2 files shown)


lib/Dialect/TTIR/Transforms/RankNormalization.cpp
```diff
@@ -204,6 +206,35 @@ class GenericRankNormalizationPattern : public ConversionPattern {
...
+    ArrayRef<int64_t> currentBroadcastDimensions =
+        broadcastOp.getBroadcastDimensions();
+    if (static_cast<int64_t>(currentBroadcastDimensions.size()) ==
+        inputType.getRank()) {
+      return;
+    }
+
+    int64_t numDimsToAdd =
+        inputType.getRank() - currentBroadcastDimensions.size();
+    if (numDimsToAdd <= 0) {
```
vwellsTT: sort of seems like <0 should be an assert/error not silent pass through?

vwellsTT (approved):

lib/Dialect/TTIR/Transforms/RankNormalization.cpp
anuragsingh-tt: This is already handled in the verifier. Deleted it.

PR #8231, opened 2026-05-03
[runtime] Add scalar tensor support for TTMetal kernel args

nsmithtt:
Introduce a Scalar variant (uint32/int32/uint16/int16/uint8/int8/bool/float) and a createScalarTensor API that packs a scalar value directly into a MetalTensor handle, instead of going through a host buffer. processKernelArgs now reads the packed uint32_t out of the MetalTensor variant rather than memcpy'ing from a TensorDesc-backed host allocation.

Also:
- Allow IndexType in TTKernel_ArgResultType and map IndexType to DataType::UInt32 so index-typed scalar args can flow through the kernel argument plumbing.
- Add std::uint32_t alternative to the MetalTensor variant and stub out the unsupported callsites (toLayout, toHost, memcpy, mesh shard utils, tensor query helpers, etc.) with LOG_FATAL so the new variant is exhaustively handled.
- Expose create_scalar_tensor through the Python runtime bindings and pull in nanobind/stl/variant.h.

### DIFF
--- a/include/ttmlir/Dialect/TTCore/IR/TTCoreOpsTypes.h
+++ b/include/ttmlir/Dialect/TTCore/IR/TTCoreOpsTypes.h
@@ -218,6 +218,10 @@ inline std::optional<DataType> elementTypeToDataTypeImpl(Type elementType) {
     }
   }
 
+  if (mlir::isa<mlir::IndexType>(elementType)) {
+    return DataType::UInt32;
+  }
+
   return {};
 }
 
--- a/include/ttmlir/Dialect/TTKernel/IR/TTKernelOps.td
+++ b/include/ttmlir/Dialect/TTKernel/IR/TTKernelOps.td
@@ -3621,7 +3621,7 @@ def TTKernel_NocSemaphoreSetMulticastLoopbackOp : TTKernel_Op<"noc_semaphore_set
 // TTKernel Compile and runtime arguments operations
 //===----------------------------------------------------------------------===//
 
-def TTKernel_ArgResultType : AnyTypeOf<[I32, UI32, TTKernel_CB, TTKernel_L1Addr]>;
+def TTKernel_ArgResultType : AnyTypeOf<[Index, I32, UI32, TTKernel_CB, TTKernel_L1Addr]>;
 
 def TTKernel_GetArgValOp : TTKernel_Op<"get_arg_val", [Pure]> {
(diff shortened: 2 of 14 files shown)


svuckovicTT (approved):

sohaibnadeemTT:
We already have support for passing scalar values through the named argument, isn't this redundant?

nsmithtt:
> We already have support for passing scalar values through the named argument, isn't this redundant?

No because this will let you pass a scalar in through a top level function argument, the named arg approach is attribute based so it must be statically known by the compiler.  That said, named argument can be deprecated after this PR since this is a strict superset.

wenbinlyuTT (approved):
The name `ScalarTensor` felt a bit off since it's a tensor that doesn't participate in the compute like other tensors do (through MeshBuffer). But I can't come up with a more appropriate name.

hkwonTT (approved):

sohaibnadeemTT:
> No because this will let you pass a scalar in through a top level function argument, the named arg approach is attribute based so it must be statically known by the compiler. That said, named argument can be deprecated after this PR since this is a strict superset. 

@nsmithtt Gotcha, can we not pass an actual scalar through to runtime rather than making it a tensor variant?

nsmithtt:
> > No because this will let you pass a scalar in through a top level function argument, the named arg approach is attribute based so it must be statically known by the compiler. That said, named argument can be deprecated after this PR since this is a strict superset.
> 
> @nsmithtt Gotcha, can we not pass an actual scalar through to runtime rather than making it a tensor variant?

We can it'll just impact a lot of code, basically every data structure that carries tensors (maps/vectors) will need to change to be a variant instead and the runtime Tensor type is already a variant.  I agree that it's a cleaner interface, we probably could change the user API to be a variant if you think that's cleaner.  Let me know what you think

sohaibnadeemTT:
> We can it'll just impact a lot of code, basically every data structure that carries tensors (maps/vectors) will need to change to be a variant instead and the runtime Tensor type is already a variant. I agree that it's a cleaner interface, we probably could change the user API to be a variant if you think that's cleaner. Let me know what you think

Hmm ok fair point, looking at the code, it is not very flexible in the type of objects we can pass around purely in the runtime layer. Changes LGTM but can we add a test if there isn't one already for this path?

sohaibnadeemTT (approved):

PR #8232, opened 2026-05-03
[D2M] Add TensorAccessor-based DMA lowering path

nsmithtt:
Introduces an alternative shard-level DMA lowering that uses tt-metal's TensorAccessor at runtime instead of the fully-indexed form, gated by a new --use-tensor-accessor-dma option on TTIRToTTMetalPipeline and ConvertD2MToTTKernel. When enabled, D2MLowerDMAToFullyIndexedForm is skipped and shard-level dma_read/dma_write ops are lowered to NocAsyncRead/WriteTile against a TTKernel TensorAccessor.

Compiler:
- New TTKernel ops/types: tensor_accessor.dspec, dspec.shard_shape, dspec.tensor_strides, dspec.shard_strides, and the DSpec type.
- New ArgType cases TensorAccessor, TensorStride, and Reserved.
- ArgSpecAttr::append{CompileTime,Runtime}Arg gain a numSlots parameter so multi-slot kernel args reserve placeholder entries and keep subsequent indices stable.
- D2MDMAViaTensorAccessorRewriter for shard-level DMA reads/writes; computes CTA slot count to match TensorAccessorArgs::get_compile_time_args().
- TTKernelToEmitC support for the new DSpec ops via verbatim emission (the C++ return type is auto-deduced from the TensorAccessor template).
- D2MToTTNN rejects the new arg types (TTNN mode unsupported).
- NormalizeThreadArgs no longer collapses the view-applied memref so the get_arg memref preserves the #ttcore.view layout.

Flatbuffer / runtime:
- program.fbs gains KernelArgTensorAccessor, KernelArgTensorStride, and KernelArgReserved variants.
- runtime/lib/ttmetal/arguments.h constructs tt_metal::TensorAccessorArgs from the referenced mesh buffer and appends its compile-time or common runtime args; TensorStride emits the 2D row stride; Reserved slots zero-fill any unused capacity.
- executor_utils.h builds a BufferDistributionSpec and passes it through BufferShardingArgs for sharded mesh buffers.
- TTMetalToFlatbuffer page shape fix: tiled tensors now use tile size as the page shape rather than a 1xN row.

Tests:
- TTKernelToEmitC ttkernel.mlir gains a DSpec op test case.
- test_layout switches to tiled with arange_tile goldens and exercises the new --use-tensor-accessor-dma pipeline option.
- normalize_thread_args.mlir CHECK updated to reflect the preserved view layout on get_arg.

### DIFF
--- a/lib/Dialect/EmitPy/IR/EmitPyOps.cpp
+++ b/lib/Dialect/EmitPy/IR/EmitPyOps.cpp
@@ -936,6 +936,12 @@ LogicalResult ExpressionOp::verify() {
 
   // Ensure the terminator is a yield op
   auto yield = cast<YieldOp>(body.getTerminator());
+
+  // Guard against a yield with no operands — getResult() would be OOB.
+  if (yield.getNumOperands() == 0) {
+    return emitOpError("yielded value not defined within expression");
+  }
+
   Value yieldResult = yield.getResult();
 
   // Ensure the yield result is valid
--- a/lib/Dialect/EmitPy/IR/EmitPyOps.cpp
+++ b/lib/Dialect/EmitPy/IR/EmitPyOps.cpp
@@ -938,7 +938,7 @@ LogicalResult ExpressionOp::verify() {
   auto yield = cast<YieldOp>(body.getTerminator());
 
   // Guard against a yield with no operands — getResult() would be OOB.
-  if (yield.getNumOperands() == 0) {
+  if (yield.getOperation()->getNumOperands() == 0) {
     return emitOpError("yielded value not defined within expression");
   }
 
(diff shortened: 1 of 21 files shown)


arminaleTT:
Is this intended to become the default DMA lowering path? Asking primarily to see how we should prioritize migrating D2MToTTNN and TTNN JIT runtime. This will probably require work in the TTNN flatbuffer and maybe even TTNN generic itself

vmilosevic (commented):

lib/Dialect/EmitPy/IR/EmitPyOps.cpp
```diff
@@ -936,6 +936,12 @@ LogicalResult ExpressionOp::verify() {
 
   // Ensure the terminator is a yield op
   auto yield = cast<YieldOp>(body.getTerminator());
+
+  // Guard against a yield with no operands — getResult() would be OOB.
+  if (yield.getNumOperands() == 0) {
```
vmilosevic: :warning: **clang\-diagnostic\-error** :warning:
no member named `` getNumOperands `` in `` mlir::tt::emitpy::YieldOp ``

sohaibnadeemTT:
@nsmithtt If I'm understanding correctly, this lowering path only begins once we've converted DMA to fully indexed form (i.e. D2MToTTKernel and lower)? i.e. we're not trying to get rid of the existing DMA lowering logic which allows us to read from views/composite views and abstracts the coalescing needed for us.

PR #8234, opened 2026-05-04
[OpModel] Fix Y/X order in isLayoutLegalForTensorShape

rpavlovicTT:
### Bug

GridAttr.getShape() returns [Y, X] (rows, cols) per the createWorkerGrid convention, while ::tt::tt_metal::CoreCoord{x, y} takes X first.  The old code passed shape[0]=Y as .x and shape[1]=X as .y, producing a transposed worker rectangle on non-square chips.  

### Changes 

Swap to {shape[1], shape[0]} so .x = cols and .y = rows.

### DIFF
--- a/lib/OpModel/TTNN/TTNNOpModel.cpp
+++ b/lib/OpModel/TTNN/TTNNOpModel.cpp
@@ -550,9 +550,13 @@ bool isLayoutLegalForTensorShape(llvm::ArrayRef<int64_t> tensorShape,
   // we return false.
   try {
     auto tensorSpec = conversion::getTensorSpec(tensorShape, layout);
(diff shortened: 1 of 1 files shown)


bmalesevicTT (approved):
Nice
Note: Afaik this change only fixes the mixed-up [Y,X] → (x,y) mapping in LegalTensorLayoutAnalysis’s isLayoutLegalForTensorShape helper. The OpModel API (getOpConstraints / getOpRuntime) and runtime have always compared GridAttr vs CoreCoord correctly via Device::getDeviceConstraints / checkGrid, which is why this bug didn’t manifest on Wormhole and only affected layout pruning on Blackhole.

PR #8240, opened 2026-05-04
[TTNN] Fixing encoing incorrect shard shape

### LINKED ISSUE #8169: `TTNNDecomposeLayouts` produces a function-signature mismatch that a post-pass fixup paints over
Parent: #8162

## Problem

When `TTNNDecomposeLayouts` lowers a `ttnn.to_layout` to concrete ops
(`ttnn.to_device`, etc.), the new op's result type can have a different
physical memref shape than the `ToLayoutOp` it replaced — most commonly
when moving to L1 interleaved produces a flattened page layout. If that
result is a `func.return` operand, the `func.func`'s declared result
type now disagrees with what its terminator actually returns.

Concretely on the GPT-OSS 120B `expert_mapping` const-eval chain
(`@main_const_eval_*` of the decode graph): the function is hoisted at
const-eval time with its return type pinned to the original
2-D L1 layout. After `TTNNDecomposeLayouts`, the rebuilt `to_device`
advertises the flattened form instead:

    // Before TTNNDecomposeLayouts (function signature pinned at hoist time)
    func.func private @main_const_eval_5(...) -> tensor<32x128xui16, #l1_2d> {
      ...
      %0 = "ttnn.to_layout"(%cast)
        : (...) -> tensor<32x128xui16, #l1_2d>
      func.return %0 : tensor<32x128xui16, #l1_2d>
    }

    // After TTNNDecomposeLayouts (return-op operand type now diverges)
    func.func private @main_const_eval_5(...) -> tensor<32x128xui16, #l1_2d> {
      ...
      %0 = "ttnn.to_device"(...)
        : (...) -> tensor<1x4096xui16, #l1_flat>   // flat memref form
      func.return %0 : tensor<1x4096xui16, #l1_flat>  // ≠ declared
    }

The signature can't simply be updated to match: the function is called
from `ttcore.load_cached` in `@main`, whose result type would then need
to change too — and that result is in turn pinned at trace-hoist time to
match `ttnn.capture_or_execute_trace`'s capture-callee argument type.
Cascading the new type through would fail the `capture_or_execute_trace`
verifier.

## Current workaround

`TTNNDecomposeLayouts::syncFunctionSignatures` (introduced in commit
`7f06ecaad`, `lib/Dialect/TTNN/Transforms/TTNNDecomposeLayouts.cpp:88-110`).
After decomposition runs, walks every `func.func`, compares each
return-op operand type against the declared return type, and on
mismatch silently `setType()`s the operand back to the declared type:

    Value operand = returnOp.getOperand(i);
    Type declared = declaredReturnTypes[i];
    if (operand.getType() != declared) {
      operand.setType(declared);
    }

Because `Value::setType` mutates the producing `OpResult` in place, the
op that produced the mismatched memref (the rebuilt `to_device`) now
advertises the originally declared 2-D form. The function signature is
left unchanged, so `ttcore.load_cached` and the trace capture callee
keep agreeing.

The reconciliation pinning is what
`test/ttmlir/Dialect/TTNN/ccl/moe_gpt/gpt_oss_120b_decode_1_layer.mlir`
asserts via `CHECK: memref<32x128xui16, #l1>` and
`CHECK-NOT: memref<1x4096xui16`.

mtopalovicTT:
Issue #8169.

This PR fixes 4 callsites where encoding was or possibly could be generated with incorrect shard shape. This can happen when source encoding (DRAM, Interleaved) is changed into target (L1, Interleaved) using `with` builder from `TTNNLayout` where shard shape is simply copied from source encoding instead of recomputed.

closes #8169

### DIFF
--- a/lib/Dialect/TTNN/Transforms/OptimizerPasses/Optimizer.cpp
+++ b/lib/Dialect/TTNN/Transforms/OptimizerPasses/Optimizer.cpp
@@ -854,10 +854,16 @@ private:
       TTNNLayoutAttr layoutAttr =
           mlir::cast<TTNNLayoutAttr>(tensorType.getEncoding());
 
-      // Create a new tensor type with DRAM layout.
+      // Create a new tensor type with DRAM layout. The trailing
+      // .withTensorShape(...) reruns TTNNLayoutAttr::get(tensorShape, ...)
+      // so the memref shard shape is recomputed under the new (bufferType,
+      // memLayout) rather than carrying over the source layout's shard
+      // shape — see calculateLogicalShardShapeForL1Interleaved /
+      // calculateLogicalShardShapeForSharding (TTNNOpsAttrs.cpp).
       TTNNLayoutAttr dramLayout =
           layoutAttr.withBufferType(BufferType::DRAM)
-              .withMemoryLayout(TensorMemoryLayout::Interleaved);
+              .withMemoryLayout(TensorMemoryLayout::Interleaved)
+              .withTensorShape(tensorShape);
       RankedTensorType newTensorType = RankedTensorType::get(
           tensorShape, tensorType.getElementType(), dramLayout);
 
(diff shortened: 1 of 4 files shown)


mtopalovicTT:
Implicitly fixed with #8127

PR #8255, opened 2026-05-05
Gather Workaround for index tensor of type si32

aorlovicTT:
### Problem description
The current gather implementation didn't account for si32 indexes, since the trivial solution would be to just typecast it to ui32 and break the op if there are negative indexes and since ```torch.gather``` doesn't support negative indexes, in the offline discussion we agreed on a similar implementation of the 
```
jax.lax.gather(..., mode = "fill")
```
which populates the corresponding negative indexes in output tensor with fill values, for example:
```
array: [10, 20, 30, 40, 50]
index: [0, 2, -1, 3, -2]
output: [10, 30, fill_value, 40, fill_value]
```
The fill_value depends on the output dtype in the corresponding way:
```
float        => fill_value = NaN  
signed int   => fill_value = a large negative number
unsigned int => fill_value = a large positive number
```

### What's changed
Relaxed the verifier to accept **any** 32bit intiger
Added a WorkaroundPattern for the gatherop

### Checklist


dmilinkovicTT (approved):
Can we open a tt-metal ticket and add a link to it in the workaround implementation?

PR #8273, opened 2026-05-05
Conv2dOp to InvokeOpLib

pljubisavljevicTT:
### Ticket
Link to Github Issue

### Problem description                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                           
### What's changed
This is the second PR in the TTNNOpInvokeLib series. It introduces a shared library that unifies conv2d op dispatch across OpModel and runtime.

  Conv2dOpT (the NativeTable type for conv2d) is the interface passed into the library. In OpModel it is constructed from MLIR attributes via buildConv2dOpTFromMLIR; in Runtime it is already available directly from the deserialized flatbuffer. Both call sites then delegate to ttnn_op_invoke::callConv2d, which handles parameter resolution and dispatch for all three call types: constraint query, runtime query, and execution.

Depends on: https://github.com/tenstorrent/tt-mlir/pull/8069

### DIFF
--- a/include/ttmlir/OpInvoke/TTNN/Conv/Conv2dOp.h
+++ b/include/ttmlir/OpInvoke/TTNN/Conv/Conv2dOp.h
@@ -1,0 +1,48 @@
+// SPDX-FileCopyrightText: (c) 2026 Tenstorrent AI ULC
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#ifndef TTMLIR_OPINVOKE_TTNN_CONV_CONV2DOP_H
+#define TTMLIR_OPINVOKE_TTNN_CONV_CONV2DOP_H
+
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wcovered-switch-default"
+#include "ttmlir/Target/TTNN/operations/conv_generated.h"
+#pragma clang diagnostic pop
+#include "ttmlir/OpInvoke/TTNN/utils/utils.h"
+#include "ttnn/graph/graph_query_op_constraints.hpp"
+#include "ttnn/graph/graph_query_op_runtime.hpp"
+#include "ttnn/operations/conv/conv2d/conv2d.hpp"
+
+#include <optional>
+
+namespace ttnn_op_invoke {
+
+using Conv2dOpResult = std::variant<::ttnn::graph::ConstraintQueryResponse,
+                                    ::ttnn::graph::RuntimeQueryResponse,
+                                    ::ttnn::Conv2dResultWithOptions>;
+
+struct Conv2dResolvedParams {
+  std::array<uint32_t, 2> kernelSize;
+  std::array<uint32_t, 2> stride;
+  std::array<uint32_t, 2> dilation;
+  std::variant<std::array<uint32_t, 2>, std::array<uint32_t, 4>> padding;
+  std::optional<::ttnn::DataType> outputDtype;
+  std::optional<::ttnn::Conv2dConfig> conv2dConfig;
+  std::optional<::ttnn::DeviceComputeKernelConfig> computeConfig;
+  std::optional<::ttnn::MemoryConfig> outputMemoryConfig;
+  std::optional<::ttnn::Conv2dSliceConfig> sliceConfig;
+};
+
+Conv2dResolvedParams
+resolveConv2dParams(const ::tt::target::ttnn::Conv2dOpT &conv2dOp);
+
+Conv2dOpResult callConv2d(CallType callType,
+                          const ::tt::target::ttnn::Conv2dOpT &conv2dOp,
+                          TensorArg input, TensorArg weight,
+                          std::optional<TensorArg> bias,
+                          ::ttnn::MeshDevice *targetDevice);
+
+} // namespace ttnn_op_invoke
+
+#endif // TTMLIR_OPINVOKE_TTNN_CONV_CONV2DOP_H
--- a/include/ttmlir/OpInvoke/TTNN/utils/utils.h
+++ b/include/ttmlir/OpInvoke/TTNN/utils/utils.h
@@ -47,6 +47,57 @@ enum class CallType {
   EXECUTE,
 };
 
+using TensorArg = std::variant<const ::ttnn::Tensor *, ::ttnn::TensorSpec>;
+
+struct QueryTag {};
+struct ExecuteTag {};
+
+inline auto resolveTensorArg(TensorArg arg, QueryTag callType) {
+  return std::get<::ttnn::TensorSpec>(arg);
+}
+
+inline auto resolveTensorArg(TensorArg arg, ExecuteTag callType) {
+  return *std::get<const ::ttnn::Tensor *>(arg);
+}
+
+template <typename Result, bool ConstraintsImplemented = true,
+          bool RuntimeImplemented = true>
+Result callOp(auto op, CallType callType, auto makeTuple,
+              ::ttnn::MeshDevice *device) {
+  switch (callType) {
+  case CallType::QUERY_OP_CONSTRAINTS:
+    if constexpr (!ConstraintsImplemented) {
+      llvm_unreachable(
+          "Constraint query invoked for op without OpModel support");
+    } else {
+      return std::apply(
+          [&](auto &&...args) {
+            return ::ttnn::graph::query_op_constraints(op, device, args...);
+          },
+          makeTuple(QueryTag{}));
+    }
+  case CallType::QUERY_OP_RUNTIME:
+    if constexpr (!RuntimeImplemented) {
+      llvm_unreachable("Runtime query invoked for op without OpModel support");
+    } else {
+      return std::apply(
+          [&](auto &&...args) {
+            return ::ttnn::graph::query_op_runtime(op, device, args...);
+          },
+          makeTuple(QueryTag{}));
+    }
+  case CallType::EXECUTE: {
+    auto executeTuple = makeTuple(ExecuteTag{});
+    return std::apply(
+        [&](auto &&...args) {
+          return op(std::forward<decltype(args)>(args)...);
+        },
+        executeTuple);
+  }
+  }
+  llvm_unreachable("unhandled CallType");
+}
+
 } // namespace ttnn_op_invoke
 
 namespace ttnn_op_invoke::operations::utils {
@@ -70,7 +121,16 @@ toTTNNCoreRange(const tt::target::ttnn::CoreRange &coreRange);
 tt::tt_metal::CoreRangeSet
 toTTNNCoreRangeSet(const tt::target::ttnn::CoreRangeSetT &coreRangeSet);
 
+::ttnn::operations::unary::UnaryOpType
+toTTNNUnaryOpType(::tt::target::ttnn::UnaryOpType unaryOpType);
+
+::ttnn::operations::unary::UnaryOpType
+toTTNNUnaryOpType(::tt::target::ttnn::EltwiseUnaryOpType unaryOpType);
+
 ::ttnn::DataType getDataType(const ::tt::target::ttnn::TensorRefT &tensorRef);
+
+::ttnn::operations::unary::UnaryWithParam
+toTTNNUnaryWithParam(const ::tt::target::ttnn::UnaryWithParamT &unaryWithParam);
 
 bool inSystemMemory(const ::tt::target::ttnn::TensorRefT &tensorRef);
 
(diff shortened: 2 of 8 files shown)


unified_op_lib/include/conv/unifiedConv2dOp.h
```diff
@@ -0,0 +1,51 @@
+// SPDX-FileCopyrightText: (c) 2024 Tenstorrent AI ULC
```
rpavlovicTT: 2024->2026

include/ttmlir/OpInvoke/TTNN/Conv/Conv2dOp.h
```diff
@@ -0,0 +1,51 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#ifndef TT_RUNTIME_DETAIL_TTNN_OPERATIONS_UNIFIED_CONV2D_OP_H
+#define TT_RUNTIME_DETAIL_TTNN_OPERATIONS_UNIFIED_CONV2D_OP_H
+
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wcovered-switch-default"
+#include "ttmlir/Target/TTNN/operations/conv_generated.h"
+#pragma clang diagnostic pop
```
rpavlovicTT, line 11: Why do we need this?

lib/OpInvoke/TTNN/conv/conv2dOp.cpp
```diff
@@ -0,0 +1,135 @@
...
+
+Conv2dOpResult callConv2d(CallType callType,
+                          const ::tt::target::ttnn::Conv2dOpT &conv2dOpT,
+                          TensorArg input, TensorArg weight,
+                          std::optional<TensorArg> bias,
+                          ::ttnn::MeshDevice &targetDevice) {
+
+  Conv2dResolvedParams params = resolveConv2dParams(conv2dOpT, callType);
+
+  switch (callType) {
```
rpavlovicTT: Here we are reaching crucial piece of the project. 
The three switch arms are 80% duplicated argument lists — exactly the divergence bug surface the lib exists to eliminate, baked into the dispatch site. This pattern will repeat in every op. Before you continue with matmul/eltwise PRs, please propose a refactor that gets the conv2d argument list to one canonical place. There are several approaches (factor common args via std::tuple + std::apply, split into separate query/execute functions, type-tag dispatch, etc.) — pick what you think is cleanest, write a short proposal explaining your choice and the tradeoffs, then apply it to conv2d as the proof-of-concept. Future op PRs will follow the same pattern.

rpavlovicTT (changes requested):

lib/OpModel/TTNN/TTNNOpModel.cpp
```diff
@@ -5170,6 +5170,60 @@ llvm::Expected<size_t> OpModel<PagedFillCacheOp>::getOpRuntime(
...
+  conv2dOpT.kernel_size =
+      std::vector<int32_t>(kernel_size.begin(), kernel_size.end());
+  conv2dOpT.stride = std::vector<int32_t>(stride.begin(), stride.end());
+  auto reorderedPadding = detail::reorderPool2dPadding(padding);
+  std::visit(
+      [&conv2dOpT](const auto &arr) {
+        conv2dOpT.padding.assign(arr.begin(), arr.end());
+      },
+      reorderedPadding);
+  conv2dOpT.padding = std::vector<int32_t>(padding.begin(), padding.end());
```
sgholamiTT: The reordered padding is computed and copied into `conv2dOpT.padding`, then the next line overwrites it with the un‑reordered padding.

lib/OpInvoke/TTNN/Conv/Conv2dOp.cpp
```diff
@@ -0,0 +1,135 @@
...
+    std::copy_n(conv2dOpT.padding.begin(), 4, asymPadding.begin());
+    params.padding = asymPadding;
+  }
+
+  if (conv2dOpT.output_dtype.has_value()) {
+    params.outputDtype =
+        operations::utils::toTTNNDataType(*conv2dOpT.output_dtype);
+  } else if (conv2dOpT.out) {
+    params.outputDtype = operations::utils::getDataType(*conv2dOpT.out);
+  }
```
sgholamiTT, line 46: In `buildConv2dOpTFromMLIR`:
```c++
if (conv2dOpT.out) {
  conv2dOpT.output_dtype =
      conv2dOpT.out->desc->layout->memory_desc->data_type;
}
```
OpModel always sets `output_dtype` (from the output layout). 
Runtime only sets it when the MLIR op explicitly carries one (TTNNToFlatbuffer writes it conditionally).
I guess the point of the unification is to eliminate this kind of divergence.

lib/OpInvoke/TTNN/conv/conv2dOp.cpp
```diff
@@ -0,0 +1,135 @@
...
+      b = *std::get<const ::ttnn::Tensor *>(*bias);
+    }
+
+    return ::ttnn::conv2d(
+        in, wt, &targetDevice, conv2dOpT.in_channels, conv2dOpT.out_channels,
+        conv2dOpT.batch_size, conv2dOpT.input_height, conv2dOpT.input_width,
+        params.kernelSize, params.stride, params.padding, params.dilation,
+        conv2dOpT.groups, params.outputDtype, b, params.conv2dConfig,
+        params.computeConfig, params.outputMemoryConfig, params.sliceConfig);
+  }
```
sgholamiTT: Add `llvm_unreachable("...")`.

lib/OpModel/TTNN/TTNNOpModel.cpp
```diff
@@ -5310,29 +5357,26 @@ llvm::Expected<size_t> OpModel<Conv2dOp>::getOpRuntime(
...
+        weightSpec,
+        biasSpec.has_value()
+            ? std::optional<unified_op_lib::TensorArg>(*biasSpec)
+            : std::nullopt,
+        *device);
+
+    assert(
+        std::holds_alternative<::ttnn::graph::RuntimeQueryResponse>(result) &&
+        "Expected Conv2dOp constraints query to return "
+        "ConstraintQueryResponse");
```
sgholamiTT: Should be "runtime query" / "RuntimeQueryResponse".

runtime/lib/ttnn/operations/conv/conv2d.cpp
```diff
@@ -3,16 +3,16 @@
 // SPDX-License-Identifier: Apache-2.0
 
 #include "operations/conv/conv2d.h"
+#include "conv/unifiedConv2dOp.h"
 #include "tt/runtime/detail/common/logger.h"
+// #include "tt/runtime/detail/ttnn/operations/unified_op_lib/unifiedConv2dOp.h"
```
sgholamiTT: ?

unified_op_lib/include/conv/unifiedConv2dOp.h
```diff
@@ -0,0 +1,51 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#ifndef TT_RUNTIME_DETAIL_TTNN_OPERATIONS_UNIFIED_CONV2D_OP_H
+#define TT_RUNTIME_DETAIL_TTNN_OPERATIONS_UNIFIED_CONV2D_OP_H
```
sgholamiTT: `UNIFIED_OP_LIB_*_OP_H` ?

include/ttmlir/OpInvoke/TTNN/Conv/Conv2dOp.h
pljubisavljevicTT: In types_generated.h, included transitively via conv_generated.h, there is a -Wcovered-switch-default warning.

lib/OpInvoke/TTNN/Conv/Conv2dOp.cpp
pljubisavljevicTT: We're aware of this divergence. Not expected to be a problem and out of my scope here.

lib/OpModel/TTNN/TTNNOpModel.cpp
```diff
@@ -5198,6 +5198,59 @@ llvm::Expected<size_t> OpModel<PagedFillCacheOp>::getOpRuntime(
 //===----------------------------------------------------------------------===//
 // Conv2dOp
 //===----------------------------------------------------------------------===//
+
+#ifdef TTMLIR_ENABLE_OPMODEL
+static ::tt::target::ttnn::Conv2dOpT buildConv2dOpTFromMLIR(
```
rpavlovicTT, line 4562: This adds a new possible bug area. To manage it we have few options, the easiest for now is to have regression tests for each op where we check that buildOpTFromMLIR yields same native struct as op->FB->unpackTo path. Having such tests should guard us. So please add those in this PR for conv2d, and then later add them for other ops in upcoming PRs.

rpavlovicTT: Test strcuture should roughly be 

```
TEST_P(Conv2dOpTPathParity, BuildEqualsFlatbufferRoundTrip) {
    // Construct an MLIR Conv2dOp with the parameterized attributes
    ttnn::Conv2dOp mlirOp = buildTestConv2dOp(GetParam());
    TTNNLayoutAttr outputLayout = resolveOutputLayout(mlirOp);

    // Path A: OpModel-style construction
    // Mirror what OpModelInterface passes down to OpModel<Conv2dOp>::getOpConstraints
    Conv2dOpT opT_opmodel = buildConv2dOpTFromMLIR(
        mlirOp.getInChannels(),
        mlirOp.getOutChannels(),
        mlirOp.getBatchSize(),
        mlirOp.getInputHeight(),
        mlirOp.getInputWidth(),
        mlirOp.getKernelSize(),
        mlirOp.getStride(),
        mlirOp.getPadding(),
        mlirOp.getDilation(),
        mlirOp.getGroups(),
        mlirOp.getConv2dConfig(),
        mlirOp.getDeviceComputeKernelConfig(),
        mlirOp.getConv2dSliceConfig(),
        outputLayout);

    // Path B: FB serialization round-trip (what runtime sees)
    flatbuffers::FlatBufferBuilder fbb;
    FlatbufferObjectCache cache(&fbb);
    auto offset = mlir::tt::createOp(cache, mlirOp);
    fbb.Finish(offset);
    Conv2dOpT opT_fb;
    flatbuffers::GetRoot<::tt::target::ttnn::Conv2dOp>(fbb.GetBufferPointer())
        ->UnPackTo(&opT_fb);

    // Structural comparison (reuse the equality machinery from ToNativeConsistency.cpp)
    assertStructurallyEqual(opT_opmodel, opT_fb);
  }
  ```

File: lib/OpInvoke/TTNN/conv/conv2dOp.cpp [hunk 11]
```diff
@@ -0,0 +1,137 @@
...
+        },
+        makeTuple(QueryTag{}));
+  case CallType::EXECUTE:
+    return std::apply(
+        [&](auto &&...args) {
+          return ::ttnn::conv2d(std::forward<decltype(args)>(args)...);
+        },
+        makeTuple(ExecuteTag{}));
+  }
+  llvm_unreachable("unhandled CallType");
```
rpavlovicTT: Can we make a template function out of this switch? It should be same for every op, right?

include/ttmlir/OpInvoke/TTNN/utils/utils.h
```diff
@@ -38,6 +40,63 @@ enum class CallType {
...
+          std::string("Constraint query not implemented for ") + opName +
+          " yet";
+      return response;
+    }
+    return std::apply(
+        [&](auto &&...args) {
+          return ::ttnn::graph::query_op_constraints(
+              op, device, std::forward<decltype(args)>(args)...);
+        },
+        makeTuple(QueryTag{}));
```
rpavlovicTT: Lets put unreachable here, and put std::apply in else branch. 

```cpp
case CallType::QUERY_OP_CONSTRAINTS:
    if constexpr (!ConstraintsImplemented) {
      llvm_unreachable("Constraint query invoked for op without OpModel support");
    } else {
      return std::apply([&](auto &&...args) {
        return ::ttnn::graph::query_op_constraints(op, device, args...);
      }, makeTuple(QueryTag{}));
    }
```

rpavlovicTT: Also with the above fix, you won't need opName anymore

rpavlovicTT (changes requested):

rpavlovicTT (approved):

PR #8275, opened 2026-05-05
[D2M] Disambiguate ThreadAttr assembly format

nsmithtt:
The declarative format for `#d2m.thread<...>` had two optional groups both led by a comma:

    `<` $threadType (`,` $kernelSymbol^)? (`,` `noc` `=` $nocIndex^)? `>`

The tablegen parser commits to the first optional on seeing `,`, so `#d2m.thread<datamovement, noc = 0>` failed with "expected attribute value" while trying to read `noc` as a SymbolRefAttr. Switch to a hand-rolled parser/printer that peeks for the `noc` keyword after the comma to pick the right branch. The print form is unchanged, so existing IR round-trips identically.

### DIFF
--- a/include/ttmlir/Dialect/D2M/IR/D2MOpsAttrs.td
+++ b/include/ttmlir/Dialect/D2M/IR/D2MOpsAttrs.td
@@ -30,7 +30,11 @@ def D2M_ThreadAttr : AttrDef<D2M_Dialect, "Thread", [], "::mlir::Attribute"> {
     DefaultValuedParameter<"int32_t", "-1">:$nocIndex
   );
 
-  let assemblyFormat = [{ `<` $threadType (`,` $kernelSymbol^)? (`,` `noc` `=` $nocIndex^)? `>` }];
+  // Format: `<` threadType (`,` kernelSymbol)? (`,` `noc` `=` nocIndex)? `>`.
+  // Both optional groups start with `,`, which the declarative format cannot
+  // disambiguate, so we hand-roll the parser to peek for the `noc` keyword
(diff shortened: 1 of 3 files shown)


include/ttmlir/Dialect/D2M/IR/D2MOpsAttrs.td
```diff
@@ -30,7 +30,11 @@ def D2M_ThreadAttr : AttrDef<D2M_Dialect, "Thread", [], "::mlir::Attribute"> {
     DefaultValuedParameter<"int32_t", "-1">:$nocIndex
   );
 
-  let assemblyFormat = [{ `<` $threadType (`,` $kernelSymbol^)? (`,` `noc` `=` $nocIndex^)? `>` }];
+  // Format: `<` threadType (`,` kernelSymbol)? (`,` `noc` `=` nocIndex)? `>`.
```
vwellsTT, line 33: nit: I feel like this comment is repeated 3x, and I'm usually not a huge fan of comments which are sort of saying "we used to do X but now we do Y" instead of "we do Y"--I think LLMs generate these a lot, but the intended audience for the comment is future coders, not PR reviewers.
For this case, I'd suggest something like:
this file could just have the
```
  // Format: `<` threadType (`,` kernelSymbol)? (`,` `noc` `=` nocIndex)? `>`.
  // Custom format needed to handle ambiguous comma
```
full explanation comment in the .cpp file as is now.
In the .mlir file, only 
```
// Verify all valid forms of #d2m.thread<...> round-trip.
```
seems needed

vwellsTT (approved):

PR #8277, opened 2026-05-05
[TTIR] Add TTIRConsolidateStaticCacheUpdates pass to reduce transformers 5.5 StaticCache overhead

ssaliceTT:
### Ticket

tt-xla transformers 5.5 uplift performance regression

### Problem description

Uplifting tt-xla to **Transformers 5.5**  [4272](https://github.com/tenstorrent/tt-xla/pull/4272) caused a **~30% decode throughput regression** on TP LLM benchmarks (e.g. LLaMA 3.1 8B instruct TP, batch=32, 8-chip Wormhole).

**Root cause - Cache position tracking for Static Cache changed from Python int to `torch.tensor` in transformers 5.5 onwards:**

In transformers 5.2, `cache_position` was a plain Python `int`, never entering the compiled XLA graph. In 5.5, it's been deprecated and cache positions are internally tracked within the cache with `StaticCache.cumulative_length`, a `torch.tensor` on device that is updated within`StaticCache.update()`:

```python
# transformers 5.5 StaticCache.update() — runs inside compiled graph, once per layer
cache_position = torch.arange(kv_length, device=self.device) + self.cumulative_length
self.cumulative_length.add_(kv_length)
```

This means **every transformer layer** in the compiled decode graph gains an `add(cumulative_length, kv_length)` + `repeat(1→32)` op pair. For a 32-layer model, that's 32 redundant add/repeat sequences per decode step where all 32 values are identical — every layer's `cumulative_length` tracks the same global decode position.

**Impact observed on LLaMA 3.1 8B instruct TP as an example:**

| Metric | Transformers 5.2 (baseline) | Transformers 5.5 (before fix) |
|---|---|---|
| Decode SPS | 24.23 | 17.65 |
| Regression | — | **−27%** |
| Extra `ttnn.add (si32)` in decode graph | 1 | 32+ |
| Extra `ttnn.repeat` ops | 0 | 31+ |

### What's changed

Added `TTIRConsolidateStaticCacheUpdates`, a new TTIR-level pass that exploits the **lockstep invariant**: all per-layer `cumulative_length` block args receive the same `add(blockArg, constant_delta)` with identical delta and type at every decode step. The pass:

1. **Identifies** groups of CL block args where all members share the same `add(arg, constant_delta)` pattern with equal delta values.
2. **Designates** the last group's result as the canonical value.
3. **Redirects** all earlier return slots to the canonical result, eliminating N−1 redundant add chains.
4. **Erases** the now-dead intermediate add ops.
5. Relies on **CSE** (run immediately after via `mlir::createCSEPass()` in the pipeline) to collapse the remaining duplicate `ttnn.repeat` broadcasts that previously fanned out from each layer's distinct CL arg — these collapse to a single repeat since they all now read the same canonical arg.

**Pipeline wiring** (`TTNNPipelines.cpp`):
```cpp
// In TTIRToTTNN pipeline, before optimizer:
pm.addPass(mlir::tt::ttir::createTTIRConsolidateStaticCacheUpdates());
pm.addPass(mlir::createCSEPass());
```

**Result on LLaMA 3.1 8B instruct TP:**

| Metric | Baseline (5.2) | Before fix (5.5) | After fix (5.5 + pass) |
|---|---|---|---|
| Decode SPS | 24.23 | 17.65 | ~23.1 |
| Regression vs baseline | — | −27% | **~5%** |
| `ttnn.add (si32)` in decode graph | 1 | 32 | **1** |
| `ttnn.repeat` in decode graph | baseline | +31 | **+0** |

The residual ~4–5% regression could be from a separate issue of SDPA attention mask spillage to DRAM.

**Lit tests** added under `test/ttmlir/Dialect/TTIR/Transforms/ConsolidateStaticCacheUpdates/`:
- `consolidate_static_cache_updates.mlir` — basic consolidation, multi-group, type-mismatch, and negative (non-uniform delta, unmarked args) cases, with FileCheck assertions on argument count and op elimination.
- `infer_static_cache_cumulative_length_args.mlir` — verifies the inference pass marks args feeding `update_cache` / `paged_update_cache` (direct and through-add cases) and does not mark unrelated args.

### Results from most recent perf benchmark runs on tt-xla 

[n300-llmbox](https://github.com/tenstorrent/tt-xla/actions/runs/25390012445):
**SPS Regressions >5%:**
  falcon3_10b_tp:                                 -6.9%  (25.8 → 24.0263 SPS)
  qwen_3_32b_tp:                                   -6.2%  (4.57 → 4.2861 SPS)
  falcon3_7b_tp:                                   -5.9%  (34.43 → 32.4066 SPS)
  llama_3_1_8b_instruct_tp:                        -5.5%  (24.72 → 23.3631 SPS)

**SPS Improvements:**
  ministral_8b_tp:                                 +8.4%  (7.3 → 7.9163 SPS)

[n150](https://github.com/tenstorrent/tt-xla/actions/runs/25389990437):
**SPS Regressions >5%:**
  qwen_2_5_0_5b:                                   -7.5%  (76.86 → 71.0962 SPS)
  qwen_3_0_6b:                                     -5.2%  (35.77 → 33.9208 SPS)

**SPS Improvements:**
  vovnet:                                          +8.6%  (648.98 → 704.6497 SPS)
  mnist:                                           +6.2%  (13476.31 → 14316.4768 SPS)
  efficientnet:                                    +5.7%  (306.54 → 324.0591 SPS)

### Checklist
- [x] Perf benchmark result (pending CI run on this branch)


lib/Dialect/TTIR/Transforms/ConsolidateStaticCacheUpdates.cpp
dmilinkovicTT, line 1: If I understood correctly, this pass matches a pretty generic-looking pattern:

```
//   [mesh_shard(] add(blockArg-or-mesh_shard(blockArg),
//                     constant_delta) [)]  →  return operand
```

This could match some op patterns completely unrelated to the `StaticCache` mechanism.

Do we have some concrete static cache-related ops consuming the cumulative length arguments, so we can harden the pattern and be sure that the argument is indeed cumulative_length?

lib/Dialect/TTIR/Transforms/ConsolidateStaticCacheUpdates.cpp
```diff
@@ -0,0 +1,165 @@
...
+    return std::nullopt;
+  }
+  if (auto constOp = v.getDefiningOp<ConstantOp>()) {
+    auto dense = dyn_cast<DenseIntOrFPElementsAttr>(constOp.getValue());
+    if (dense && dense.isSplat() && dense.getElementType().isInteger()) {
+      return dense.getSplatValue<APInt>().getSExtValue();
+    }
+    return std::nullopt;
+  }
+  return std::nullopt;
```
dmilinkovicTT: nit: splat `ConstantOp` always canonicalizes to `FullOp`

test/ttmlir/Dialect/TTIR/Transforms/ConsolidateStaticCacheUpdates/consolidate_static_cache_updates.mlir
```diff
@@ -0,0 +1,113 @@
...
+    %a0 = "ttir.add"(%cl0, %delta) : (tensor<1xi64>, tensor<1xi64>) -> tensor<1xi64>
+    %a1 = "ttir.add"(%cl1, %delta) : (tensor<1xi64>, tensor<1xi64>) -> tensor<1xi64>
+    %a2 = "ttir.add"(%cl2, %delta) : (tensor<1xi64>, tensor<1xi64>) -> tensor<1xi64>
+    // Internal use of %cl0 — the pass replaces %cl0 (%arg0) with the
+    // canonical %cl2 (%arg2), so this add becomes add(%arg2, %arg3).
+    %internal = "ttir.add"(%cl0, %other) : (tensor<1xi64>, tensor<1xi64>) -> tensor<1xi64>
+    // After arg unification + CSE: one write-back add survives, %arg0/%arg1 gone.
+    // CHECK:     [[KEPT:%.+]] = "ttir.add"(%arg2,
+    // CHECK-NOT: "ttir.add"(%arg0,
+    // CHECK-NOT: "ttir.add"(%arg1,
```
dmilinkovicTT, line 123: These two would actually never fail, since there aren't operands named `%arg0` and `%arg1` 

Edit: ah, okay, turns out the argument names get renamed to `argN` notation anyway... In that case, we should probably rename the block arguments in the test to avoid confusion.

test/ttmlir/Dialect/TTIR/Transforms/ConsolidateStaticCacheUpdates/consolidate_static_cache_updates.mlir
```diff
@@ -0,0 +1,113 @@
...
+      %cl1: tensor<1xi64>,
+      %cl2: tensor<1xi64>,
+      %other: tensor<1xi64>
+  ) -> (tensor<1xi64>, tensor<1xi64>, tensor<1xi64>, tensor<1xi64>) {
+    %delta = "ttir.constant"() <{value = dense<1> : tensor<1xi64>}> : () -> tensor<1xi64>
+    %a0 = "ttir.add"(%cl0, %delta) : (tensor<1xi64>, tensor<1xi64>) -> tensor<1xi64>
+    %a1 = "ttir.add"(%cl1, %delta) : (tensor<1xi64>, tensor<1xi64>) -> tensor<1xi64>
+    %a2 = "ttir.add"(%cl2, %delta) : (tensor<1xi64>, tensor<1xi64>) -> tensor<1xi64>
+    // Internal use of %cl0 — the pass replaces %cl0 (%arg0) with the
+    // canonical %cl2 (%arg2), so this add becomes add(%arg2, %arg3).
```
dmilinkovicTT: So, are we just ending up with a bunch of unused block arguments (since CSE can't pick up these)? IMHO, this feels a bit wrong.

dmilinkovicTT (commented):
The pass provided here seems too generic, and could potentially match on some patterns completely unrelated to the static cache mechanism.

Depending on the urgency of this fix, we could either:
1. Harden the constraints in the pattern-matching logic, in order to ensure that pass never misfires on, if the fix is a high-prio issue.
2. Discuss about the actual architectural fit of this pass, since it feels like it should be higher up in the compiler stack (possibly even in the compiler frontend, since we can't (at least theoretically) be sure that all `cumulative_length` tensors indeed have the same value when we're already in the TTIR world), if the fix is a lower-prio issue.

lib/Dialect/TTNN/Pipelines/TTNNPipelines.cpp
```diff
@@ -92,6 +92,15 @@ void createTTNNPipelineTTIRPasses(
     pm.addPass(mlir::tt::ttir::createTTIRFusing(fusingOptions));
   }
   pm.addPass(mlir::tt::ttir::createTTIRFoldFullToScalar());
+  pm.addPass(mlir::tt::ttir::createTTIRInferStaticCacheCumulativeLengthArgs());
+  pm.addPass(mlir::tt::ttir::createTTIRConsolidateStaticCacheUpdates());
```
dmilinkovicTT: Why do we need to infer and tag the cumulative length args, and then have the consolidation pass transform them immediately after? 

Can't both analysis + transformation happen within a single pass? This way, we wouldn't need 2 separate passes, and we wouldn't need the cumulative length attr either.

lib/Dialect/TTNN/Pipelines/TTNNPipelines.cpp
```diff
@@ -92,6 +92,14 @@ void createTTNNPipelineTTIRPasses(
     pm.addPass(mlir::tt::ttir::createTTIRFusing(fusingOptions));
   }
   pm.addPass(mlir::tt::ttir::createTTIRFoldFullToScalar());
+  pm.addPass(mlir::tt::ttir::createTTIRConsolidateStaticCacheUpdates());
+  pm.addPass(mlir::createCSEPass());
+  // After consolidation, the eliminated per-layer cumulative_length block
+  // args have no remaining uses. RemoveDeadValues erases them from private
+  // helper functions (and propagates the signature change through call
+  // sites); public/externally-visible functions are left untouched, so the
+  // FFI/runtime ABI is preserved.
```
dmilinkovicTT: This is misleading, as we shouldn't have any private functions at this stage of the pipeline. The only function we're dealing with here is a single public forward function, which means that the arguments won't be cleaned up. 

That is okay for now, since this is a high-prio fix, but it would be nice to create a follow-up ticket to address the duplicated arguments. And I think we should remove the `RemoveDeadValuesPass` from this PR.

dmilinkovicTT (approved):
Thank you for implementing suggestions! Approving since this is a high-prio fix.

Please create a follow-up ticket to address the following issues:
- Unused block arguments aren't deleted
- We should reconsider the decision of fixing this at the TTIR level, since it's pretty hacky, and we're assuming that the cumulative length arguments all have the same value; instead, my hunch is that we should try to lift the fix upwards in the pipeline, possibly even to the frontend

PR #8278, opened 2026-05-06
[L1SpillManagement] Include sink ops in schedule for correct lastUse

rpavlovicTT:
## Summary

Cache-write sink ops (`PagedFillCacheOp`, `PagedUpdateCacheOp`, `FillCacheOp`) were excluded from the L1 spill simulator's schedule, so their operand uses landed outside `positionMap` and `lastUse` got truncated to earlier consumers — freeing L1 slots while runtime still pinned them.

- `buildScheduleData` now uses `isBeamSearchTarget` (matches `MemoryLayoutPropagation`) so sink ops contribute to `positionMap`.
- `run()` skips sink ops after `processDeadTensors` since they have no tensor result (avoids `extractOpConfigFromIR`'s `op->getResult(0)` assert).
- `isSinkOp` extended to cover `PagedFillCacheOp` (the prefill-time variant Gemma-4 emits).

### DIFF
--- a/lib/Dialect/TTNN/Analysis/L1SpillManagement.cpp
+++ b/lib/Dialect/TTNN/Analysis/L1SpillManagement.cpp
@@ -7,6 +7,7 @@
 #include "ttmlir/Dialect/TTNN/Analysis/OpConfigAttrs.h"
 #include "ttmlir/Dialect/TTNN/IR/TTNNOps.h"
 #include "ttmlir/Dialect/TTNN/IR/TTNNOpsAttrs.h"
+#include "ttmlir/Dialect/TTNN/Utils/OptimizerUtils.h"
 #include "ttmlir/Dialect/TTNN/Utils/Utils.h"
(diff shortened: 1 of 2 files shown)


lib/Dialect/TTNN/Analysis/L1SpillManagement.cpp
```diff
@@ -467,18 +468,14 @@ typename L1SpillManagement<MemoryTracker>::ScheduleData
...
-      return;
-    }
-    if (isa<EmptyOp>(op)) {
-      return;
-    }
-    bool hasTensorResult = llvm::any_of(op->getResults(), [](OpResult r) {
-      return mlir::isa<RankedTensorType>(r.getType());
-    });
-    if (!hasTensorResult) {
+    if (!optimizer_utils::isBeamSearchTarget(op)) {
```
bmalesevicTT, line 478: optimizer_utils::opHasTensorResult checks only the first result, while previous code here checked for any tensor result, I don't know an example where this would matter but worth mentioning

bmalesevicTT (approved):

PR #8281, opened 2026-05-06
Eltwise-unary ops to InvokeOpLib

pljubisavljevicTT:
### Ticket
Link to Github Issue

### Problem description                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                           
### What's changed
This is the fourth PR in the TTNNOpInvokeLib series. It introduces shared dispatch for eltwise unary and eltwise unary composite ops across OpModel and Runtime.                                                                                          
                  
  EltwiseUnaryOpT and EltwiseUnaryCompositeOpT (the NativeTable types for these ops) are the interfaces passed into the library. In OpModel they are constructed from MLIR attributes; in Runtime they are already available directly from the deserialized flatbuffer. Both call sites then delegate to ttnn_op_invoke::callEltwiseUnary / ttnn_op_invoke::callEltwiseUnaryComposite, which handle parameter resolution and dispatch for all three call types: constraint query, runtime query, and execution.

### DIFF
--- a/include/ttmlir/OpInvoke/TTNN/Eltwise/Unary/EltwiseUnaryOp.h
+++ b/include/ttmlir/OpInvoke/TTNN/Eltwise/Unary/EltwiseUnaryOp.h
@@ -1,0 +1,173 @@
+// SPDX-FileCopyrightText: (c) 2026 Tenstorrent AI ULC
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#ifndef TTMLIR_OPINVOKE_TTNN_ELTWISE_UNARY_ELTWISEUNARYOP_H
+#define TTMLIR_OPINVOKE_TTNN_ELTWISE_UNARY_ELTWISEUNARYOP_H
+
+#include "operations/eltwise/unary/unary.hpp"
+#include "ttmlir/OpInvoke/TTNN/utils/utils.h"
+#include "ttnn/graph/graph_query_op_constraints.hpp"
+#include "ttnn/graph/graph_query_op_runtime.hpp"
+#include "ttnn/tensor/tensor.hpp"
+
+#include <cstdint>
+#include <optional>
+#include <tuple>
+
+namespace ttnn_op_invoke {
+
+using EltwiseUnaryOpResult =
+    std::variant<::ttnn::graph::ConstraintQueryResponse,
+                 ::ttnn::graph::RuntimeQueryResponse, ::ttnn::Tensor>;
+
+struct EltwiseUnaryResolvedParams {
+  std::optional<::ttnn::MemoryConfig> outputMemoryConfig;
+  bool fastApproxMode = false;
+  std::optional<bool> approx;
+  std::optional<::ttnn::operations::unary::SigmoidMode> sigmoidMode;
+  std::optional<int32_t> vecMode;
+  std::optional<float> parameter;
+};
+
+EltwiseUnaryResolvedParams resolveEltwiseUnaryParams(
+    const ::tt::target::ttnn::EltwiseUnaryOpT &eltwiseUnaryOp);
+
+template <typename Tag>
+auto createEltwiseUnaryTuple(
+    Tag tag, const ::tt::target::ttnn::EltwiseUnaryOpT &eltwiseUnaryOp,
+    TensorArg input, const EltwiseUnaryResolvedParams &params) {
+  return std::make_tuple(resolveTensorArg(input, tag),
+                         params.outputMemoryConfig,
+                         /*optional_output_tensor=*/std::nullopt,
+                         /*sub_core_grids=*/std::nullopt);
+}
+
+template <typename Fn>
+EltwiseUnaryOpResult
+callEltwiseUnary(CallType callType,
+                 const ::tt::target::ttnn::EltwiseUnaryOpT &eltwiseUnaryOp,
+                 Fn ttnnOp, TensorArg input,
+                 ::ttnn::MeshDevice *device = nullptr) {
+
+  EltwiseUnaryResolvedParams params = resolveEltwiseUnaryParams(eltwiseUnaryOp);
+
+  auto makeTuple = [&](auto tag) {
+    return createEltwiseUnaryTuple(tag, eltwiseUnaryOp, input, params);
+  };
+
+  return callOp<EltwiseUnaryOpResult>(ttnnOp, callType, makeTuple, device);
+}
+
+template <typename Tag>
+auto createEltwiseUnaryTanhTuple(
+    Tag tag, const ::tt::target::ttnn::EltwiseUnaryOpT &eltwiseUnaryOp,
+    TensorArg input, const EltwiseUnaryResolvedParams &params) {
+  TT_INVOKE_ASSERT(params.approx.has_value(), "approx parameter not resolved");
+  return std::make_tuple(
+      resolveTensorArg(input, tag), params.outputMemoryConfig,
+      /*optional_output_tensor=*/std::nullopt, *params.approx,
+      /*sub_core_grids=*/std::nullopt);
+}
+
+template <typename Fn>
+EltwiseUnaryOpResult
+callEltwiseUnaryTanh(CallType callType,
+                     const ::tt::target::ttnn::EltwiseUnaryOpT &eltwiseUnaryOp,
+                     Fn ttnnOp, TensorArg input,
+                     ::ttnn::MeshDevice *device = nullptr) {
+
+  EltwiseUnaryResolvedParams params = resolveEltwiseUnaryParams(eltwiseUnaryOp);
+
+  auto makeTuple = [&](auto tag) {
+    return createEltwiseUnaryTanhTuple(tag, eltwiseUnaryOp, input, params);
+  };
+
+  return callOp<EltwiseUnaryOpResult>(ttnnOp, callType, makeTuple, device);
+}
+
+template <typename Tag>
+auto createEltwiseUnaryWithFastAndApproximateModeTuple(
+    Tag tag, const ::tt::target::ttnn::EltwiseUnaryOpT &eltwiseUnaryOp,
+    TensorArg input, const EltwiseUnaryResolvedParams &params) {
+  return std::make_tuple(resolveTensorArg(input, tag),
+                         /*approx=*/params.fastApproxMode,
+                         params.outputMemoryConfig,
+                         /*optional_output_tensor=*/std::nullopt,
+                         /*sub_core_grids=*/std::nullopt);
+}
+
+template <typename Fn>
+EltwiseUnaryOpResult callEltwiseUnaryWithFastAndApproximateMode(
+    CallType callType,
+    const ::tt::target::ttnn::EltwiseUnaryOpT &eltwiseUnaryOp, Fn ttnnOp,
+    TensorArg input, ::ttnn::MeshDevice *device = nullptr) {
+
+  EltwiseUnaryResolvedParams params = resolveEltwiseUnaryParams(eltwiseUnaryOp);
+
+  auto makeTuple = [&](auto tag) {
+    return createEltwiseUnaryWithFastAndApproximateModeTuple(
+        tag, eltwiseUnaryOp, input, params);
+  };
+
+  return callOp<EltwiseUnaryOpResult>(ttnnOp, callType, makeTuple, device);
+}
+
+template <typename Tag>
+auto createEltwiseUnarySigmoidTuple(
+    Tag tag, const ::tt::target::ttnn::EltwiseUnaryOpT &eltwiseUnaryOp,
+    TensorArg input, const EltwiseUnaryResolvedParams &params) {
+  TT_INVOKE_ASSERT(params.vecMode.has_value() && params.sigmoidMode.has_value(),
+                   "sigmoid params (vecMode, sigmoidMode) not resolved");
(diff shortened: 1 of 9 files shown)


unified_op_lib/lib/eltwise/unary/unifiedEltwiseUnaryOp.cpp
```diff
@@ -0,0 +1,54 @@
...
+
+namespace unified_op_lib {
+
+EltwiseUnaryResolvedParams resolveEltwiseUnaryParams(
+    const ::tt::target::ttnn::EltwiseUnaryOpT &eltwiseUnaryOpT,
+    CallType callType) {
+
+  EltwiseUnaryResolvedParams params;
+
+  params.fastApproxMode = true;
```
sgholamiTT: Pre‑PR runtime was:
`::ttnn::Tensor out = ttnnOp(in, /*parameter=*/false, outputMemoryConfig, ...);`

With this change, every runtime invocation of `Sqrt, Rsqrt, Erf, etc` now uses fast/approximate mode where it didn't before (which is not bad by itself). 
But that's a real numerical change in the runtime path and needs to:
1. Be mentioned in the PR description and,
2. Be reflected with some tests.

lib/OpInvoke/TTNN/Eltwise/Unary/EltwiseUnaryCompositeOp.cpp
```diff
@@ -0,0 +1,174 @@
...
+    params.min = clampParams->min.AsFP()->value;
+    params.max = clampParams->max.AsFP()->value;
+    break;
+  case ::tt::target::ttnn::NumberType::I32:
+    params.min = clampParams->min.AsI32()->value;
+    params.max = clampParams->max.AsI32()->value;
+    break;
+  default:
+    LOG_FATAL("unknown clamp scalar min/max type");
+  }
```
sgholamiTT, line 59: assert `min/max` types match in the build helper.

lib/OpModel/TTNN/TTNNOpModel.cpp
```diff
@@ -7306,17 +7662,47 @@ llvm::Expected<size_t> OpModel<GroupNormOp>::getOpRuntime(
...
-    return static_cast<int32_t>(intAttr.getValue().getSExtValue());
+static ::tt::target::ttnn::EltwiseUnaryCompositeOpT
+buildEltwiseUnaryCompositeClampScalarOpTFromMLIR(mlir::Attribute min,
+                                                 mlir::Attribute max,
+                                                 TTNNLayoutAttr outputLayout) {
+  ::tt::target::ttnn::EltwiseUnaryCompositeOpT eltwiseUnaryCompositeOpT;
+  eltwiseUnaryCompositeOpT.type =
+      ::tt::target::ttnn::EltwiseUnaryCompositeOpType::ClampScalar;
+
+  // int32_t adf = static_cast<int32_t>(intAttr.getValue().getSExtValue());
```
sgholamiTT: Remove ?

include/ttmlir/OpInvoke/TTNN/Eltwise/Unary/EltwiseUnaryOp.h
```diff
@@ -0,0 +1,201 @@
...
+                 ::ttnn::graph::RuntimeQueryResponse, ::ttnn::Tensor>;
+
+struct EltwiseUnaryResolvedParams {
+  std::optional<::ttnn::MemoryConfig> outputMemoryConfig;
+  bool fastApproxMode = false;
+  bool approx = false;
+  ::ttnn::operations::unary::SigmoidMode sigmoidMode;
+  int32_t vecMode;
+  float parameter;
+};
```
sgholamiTT, line 31: `sigmoidMode`, `vecMode`, `parameter` are only assigned for specific types. If a dispatcher (`callEltwiseUnarySigmoid`, `callEltwiseUnaryWithFloatParameter`) is called for an op whose type doesn't match what the resolver expects, it reads indeterminate values. 
Wrap each in `std::optional<…>` and have the dispatcher assert it's set, or default‑init all three. Same fix applies to the clamp scalar struct.

sgholamiTT (approved):
Overal LGTM 👍🏻 

A general comment:
This PR adds: 
`callEltwiseUnary`
`callEltwiseUnaryTanh`
`callEltwiseUnarySigmoid`
`callEltwiseUnaryWithFastAndApproximateMode`
`callEltwiseUnaryWithFloatParameter`
`callEltwiseUnaryComposite`
`callEltwiseUnaryCompositeClampScalar`
`callEltwiseUnaryCompositeClampTensor`
`callEltwiseUnaryCompositeWithFastAndApproximateMode`
…
each one a 3‑arm switch on `CallType` over basically the same skeleton. Each adds:
- a header alias for `TensorArg / *OpResult` (already duplicated)
- a `resolve*Params` function;
- a `call*` template body that hand‑writes the same query/runtime/execute branches.

Not that I have much better ideas at this point, but worth pausing to consider a single `dispatch<OpKind>(callType, opT, fn, ...)` traits‑driven entry point so the per‑op surface is a small specialization rather than a new free function. Have to think about the impl if this is what we wanna do.

That would also make `OpModel<TanhOp>` not need its own bespoke specialization carved out of `UnaryEltwiseOpModel`.

LMK what you think.

lib/OpModel/TTNN/TTNNOpModel.cpp
```diff
@@ -16,6 +16,8 @@
 #include "ttmlir/Dialect/TTNN/IR/TTNNOpsAttrs.h"
 #include "ttmlir/OpModel/TTNN/Conversion.h"
 #include "ttmlir/OpModel/TTNN/SingletonDeviceContext.h"
+#include <eltwise/unary/unifiedEltwiseUnaryCompositeOp.h>
+#include <eltwise/unary/unifiedEltwiseUnaryOp.h>
```
sgholamiTT: angle bracket.

unified_op_lib/include/eltwise/unary/unifiedEltwiseUnaryOp.h
```diff
@@ -0,0 +1,201 @@
...
+
+#include "ttnn/graph/graph_query_op_constraints.hpp"
+#include "ttnn/graph/graph_query_op_runtime.hpp"
+#include "ttnn/types.hpp"
+#include "utils/utils.h"
+
+#include <cstdint>
+#include <operations/eltwise/unary_ng/unary_ng.hpp>
+#include <optional>
+#include <ttnn/tensor/tensor.hpp>
```
sgholamiTT: include order. Applies to other PRs too. Not going to repeat this comment. Please scan other PRs and fix all instances.

lib/OpInvoke/TTNN/Eltwise/Unary/EltwiseUnaryCompositeOp.cpp
```diff
@@ -0,0 +1,174 @@
...
+#include "eltwise/unary/unifiedEltwiseUnaryCompositeOp.h"
+#include "tt/runtime/detail/common/logger.h"
+#include "tt/runtime/detail/ttnn/operations/utils.h"
+#include "tt/runtime/detail/ttnn/utils.h"
+#include "ttnn/graph/graph_query_op_constraints.hpp"
+#include "ttnn/operations/eltwise/unary/unary_composite.hpp"
+#include "utils/utils.h"
+#include <operations/functions.hpp>
+#include <optional>
+#include <variant>
```
sgholamiTT: Include order + spacing.

lib/OpInvoke/TTNN/Eltwise/Unary/EltwiseUnaryOp.cpp
```diff
@@ -0,0 +1,48 @@
...
+#include <variant>
+
+namespace ttnn_op_invoke {
+
+EltwiseUnaryResolvedParams resolveEltwiseUnaryParams(
+    const ::tt::target::ttnn::EltwiseUnaryOpT &eltwiseUnaryOp) {
+
+  EltwiseUnaryResolvedParams params;
+
+  params.fastApproxMode = false;
```
rpavlovicTT, line 18: I'd add a comment explaining that we are preserving hard-coded constant from runtime.

include/ttmlir/OpInvoke/TTNN/utils/utils.h
```diff
@@ -9,6 +9,7 @@
 #pragma clang diagnostic ignored "-Wcovered-switch-default"
 #include "ttmlir/Target/TTNN/program_generated.h"
 #pragma clang diagnostic pop
+#include "tt/runtime/detail/common/logger.h"
```
rpavlovicTT: Don't include runtime's logger in this lib. This lib should not depend on runtime. Instead let's use llvm's error handling, create 

```cpp
 #define TT_INVOKE_ASSERT(cond, msg)                                             \
    do {                                                                         \
      if (!(cond)) {                                                             \
        ::llvm::report_fatal_error(msg);                                         \
      }                                                                          \
    } while (0)
```

and use it instead of LOG_ASSERT.

lib/OpInvoke/TTNN/Eltwise/Unary/EltwiseUnaryCompositeOp.cpp
```diff
@@ -0,0 +1,155 @@
...
+  case ::tt::target::ttnn::NumberType::FP:
+    params.min = clampParams->min.AsFP()->value;
+    params.max = clampParams->max.AsFP()->value;
+    break;
+  case ::tt::target::ttnn::NumberType::I32:
+    params.min = clampParams->min.AsI32()->value;
+    params.max = clampParams->max.AsI32()->value;
+    break;
+  default:
+    LOG_FATAL("unknown clamp scalar min/max type");
```
rpavlovicTT: Use llvm's `report_fatal_error` in this case.

lib/OpInvoke/TTNN/Eltwise/Unary/EltwiseUnaryCompositeOp.cpp
```diff
@@ -0,0 +1,155 @@
...
+}
+
+EltwiseUnaryCompositeClampScalarResolvedParams
+resolveEltwiseUnaryCompositeClampScalarParams(
+    const ::tt::target::ttnn::EltwiseUnaryCompositeOpT
+        &eltwiseUnaryCompositeOp) {
+
+  EltwiseUnaryCompositeClampScalarResolvedParams params;
+
+  const auto *clampParams =
```
rpavlovicTT: Avoid auto type in cases like this, I want to read full type name when params are cast

lib/OpInvoke/TTNN/Eltwise/Unary/EltwiseUnaryOp.cpp
```diff
@@ -0,0 +1,48 @@
...
+    params.sigmoidMode = ::ttnn::operations::unary::SigmoidMode::ACCURATE;
+    params.vecMode =
+        static_cast<int32_t>(::ttnn::operations::unary::VecMode::RC);
+  }
+
+  if (eltwiseUnaryOp.type ==
+      ::tt::target::ttnn::EltwiseUnaryOpType::LeakyRelu) {
+    params.parameter =
+        eltwiseUnaryOp.params.AsEltwiseOpWithFloatParams()->parameter;
+  }
```
rpavlovicTT: For a readability sake I would change this to switch over unary op type. In default case just break;

lib/OpModel/TTNN/TTNNOpModel.cpp
```diff
@@ -6562,17 +6919,45 @@ llvm::Expected<size_t> OpModel<GroupNormOp>::getOpRuntime(
...
+  if (auto floatAttr = mlir::dyn_cast<mlir::FloatAttr>(min)) {
+    ::tt::target::ttnn::FloatingPointTypeT fp;
+    fp.value = floatAttr.getValue().convertToFloat();
+    params.min.Set(fp);
+  } else if (auto intAttr = mlir::dyn_cast<mlir::IntegerAttr>(min)) {
+    ::tt::target::ttnn::IntegralTypeT i32;
+    i32.value = static_cast<int32_t>(intAttr.getInt());
+    params.min.Set(i32);
+  } else {
+    assert(false && "Invalid clamp min attribute");
```
rpavlovicTT: use llvm unreachable in cases like this

lib/OpInvoke/TTNN/CMakeLists.txt
```diff
@@ -10,11 +12,8 @@ target_include_directories(TTNNOpInvokeLib PUBLIC
   $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include/ttmlir/OpInvoke>
   $<BUILD_INTERFACE:${PROJECT_BINARY_DIR}/include>
   $<BUILD_INTERFACE:${PROJECT_BINARY_DIR}/include/ttmlir/Target/Common>
-)
-
-target_include_directories(TTNNOpInvokeLib PRIVATE
-  ${PROJECT_SOURCE_DIR}/runtime/include
-  ${PROJECT_BINARY_DIR}/runtime/include
+  $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/runtime/include>
+  $<BUILD_INTERFACE:${PROJECT_BINARY_DIR}/runtime/include>
```
rpavlovicTT: We should not depend on runtime here

rpavlovicTT (approved):

PR #8282, opened 2026-05-06
[ttnn] Route integer reductions through fp32 instead of bf16 to avoid precision loss (e.g. 8400→8384)

### LINKED ISSUE #8279: D_FINE (Small) - ValueError: Make sure to align the spatial shapes with the sequence length of the encoder hidden states
### Describe the bug

- D_FINE small variant facing `ValueError: Make sure to align the spatial shapes with the sequence length of the encoder hidden states` - [may6_d_fine_small.log](https://github.com/user-attachments/files/27432152/may6_d_fine_small.log)
- Error arises because (spatial_shapes[:, 0] * spatial_shapes[:, 1]).sum() is not equal to sequence_length during tt run - [may6_d_fine_smalldetails_1.log](https://github.com/user-attachments/files/27432065/may6_d_fine_small_details.log)

```
sequence_length=8400
(spatial_shapes[:, 0] * spatial_shapes[:, 1]).sum()=8384
```
- Issue originates from sum op. [may6_sum_sanity.log](https://github.com/user-attachments/files/27432105/may6_sum_sanity.log), [test_sanity.py](https://github.com/user-attachments/files/27432117/test_sanity.py)

kamalrajkannan78:
### Ticket

- fixes https://github.com/tenstorrent/tt-mlir/issues/8279

### Problem description

- `createReductionOpOperandsWorkarounds` casts integer-input reductions to bf16 before `ttnn.sum` to work around [#21071](https://github.com/tenstorrent/tt-metal/issues/21071). 
- bf16's 7-bit mantissa silently rounds — e.g. `sum([6400, 1600, 400])` returns `8384` instead of `8400` 

### What's changed

- Switched the workaround target dtype from `BFloat16` to `Float32`. fp32 satisfies the same padding/transpose constraints and represents all integers up to 2^24 exactly, so int reductions now round-trip without precision loss.

### Checklist
- [x] verify the changes through local testing in N150

### Logs

- [may6_sum_sanity_before_fix.log](https://github.com/user-attachments/files/27433233/may6_sum_sanity.log)
- [may6_sum_sanity_after_fix.log](https://github.com/user-attachments/files/27433235/may6_sum_sanity_after_fix.log)

### DIFF
--- a/test/ttmlir/Dialect/TTNN/Transforms/Workarounds/reduction_ops_workaround.mlir
+++ b/test/ttmlir/Dialect/TTNN/Transforms/Workarounds/reduction_ops_workaround.mlir
@@ -7,17 +7,17 @@
 func.func public @test_reduce_max(%arg0: tensor<128x32xsi32, #ttnn_layout>) -> tensor<128xsi32, #ttnn_layout1> {
   // CHECK-LABEL: @test_reduce_max
   // CHECK: %[[ARG0:[0-9]+]] = "ttnn.to_layout"(%arg0)
-  // CHECK-SAME: dtype = #ttcore.supportedDataTypes<bf16>,
+  // CHECK-SAME: dtype = #ttcore.supportedDataTypes<f32>,
   // CHECK-SAME: tensor<128x32xsi32,
-  // CHECK-SAME: -> tensor<128x32xbf16,
+  // CHECK-SAME: -> tensor<128x32xf32,
   // CHECK: %[[MAX:[0-9]+]] = "ttnn.max"(%[[ARG0]])
   // CHECK-SAME: <{dim_arg = [1 : i32], keep_dim = false}>
-  // CHECK-SAME: tensor<128x32xbf16,
-  // CHECK-SAME: -> tensor<128xbf16,
+  // CHECK-SAME: tensor<128x32xf32,
+  // CHECK-SAME: -> tensor<128xf32,
   %0 = "ttnn.max"(%arg0) <{dim_arg = [1 : i32], keep_dim = false}> : (tensor<128x32xsi32, #ttnn_layout>) -> tensor<128xsi32, #ttnn_layout1>
   // CHECK: %{{[0-9]+}} = "ttnn.to_layout"(%[[MAX]])
   // CHECK-SAME: dtype = #ttcore.supportedDataTypes<si32>,
-  // CHECK-SAME: tensor<128xbf16,
+  // CHECK-SAME: tensor<128xf32,
   // CHECK-SAME: -> tensor<128xsi32,
   return %0 : tensor<128xsi32, #ttnn_layout1>
 }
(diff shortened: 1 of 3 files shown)


saiarthiraguram (approved):

azecevicTT (approved):

kamalrajkannan78:
@azecevicTT 

- 24 cases was failing in Push CI ([regress.zip](https://github.com/user-attachments/files/27448176/regress.zip))- all from the lit test `test/ttmlir/Dialect/TTNN/Transforms/Workarounds/reduction_ops_workaround.mlir`, which still pinned bf16 in its CHECK lines. 
- Updated those CHECKs to expect f32, matching the new workaround target dtype.

test/ttmlir/Dialect/TTNN/Transforms/Workarounds/reduction_ops_workaround.mlir
```diff
@@ -7,35 +7,35 @@
 func.func public @test_reduce_max(%arg0: tensor<128x32xsi32, #ttnn_layout>) -> tensor<128xsi32, #ttnn_layout1> {
```
sdjordjevicTT, line 7: Can you please add int in this reduction test, so we can verify with PCC that the workaround works?
https://github.com/tenstorrent/tt-mlir/blob/c44dec0c0380f60762d144277e93d8b389f23ecd/test/python/golden/ttir_ops/reduction/test_reduction.py#L47

kamalrajkannan78: @sdjordjevicTT 

After adding the suggested int32 parametrize, the current PR shows PCC failures. - [my_branch_failures.zip](https://github.com/user-attachments/files/27470755/my_branch_failures.zip)

I pushed [PR #8310](https://github.com/tenstorrent/tt-mlir/pull/8310) with only the int32 parametrize on main (no fp32 fix) and it produces the same failure pattern — confirming these are pre-existing limitations of the bf16 reduction workaround, not regressions introduced by this PR - [main_branch_failures.zip](https://github.com/user-attachments/files/27470753/main_branch_failures.zip)

Since this is a pre-existing issue and is currently blocking this PR, can we handle the int32 test addition in a separate follow-up PR?

sdjordjevicTT: I am not sure that I follow. If we apply the workaround from this PR, what will it fix? The int inputs still fail the PCC check, right?

kamalrajkannan78: It's a partial fix, not a complete one.

**What this PR fixes:** the d_fine model regression. sum([6400, 1600, 400]) was returning 8384 instead of 8400 because bf16 rounds integers above 2⁷ (~128). fp32 is exact up to 2²⁴ (~16.7M), so realistic small-magnitude integer reductions in models now produce correct results.

**What this PR doesn't fix:** arbitrary-magnitude integer reductions. The int32 PCC test feeds random values up to ±2³¹, which exceeds fp32's exact-int range too — no float-routing workaround can handle that. Full correctness needs native int reduction (https://github.com/tenstorrent/tt-metal/issues/21071).

**Net:** this PR is a strict improvement over bf16 within the limits of what a float workaround can deliver, but it can't make the int32 PCC test pass on full-range random inputs.

sdjordjevicTT: I see, thanks for the explanation! Btw, just please mark int32 test case in test/python/golden/ttir_ops/reduction/test_reduction.py with xfail, and that's it.

sdjordjevicTT (approved):

Issue #8036, opened 2026-04-20
Add DRAM sharding tests


sdjordjevicTT:
### Problem
No compiler tests validate the creation, conversion, or use of DRAM-sharded tensors in operations.

### What needs to change

- Add MLIR lit tests for DRAM-sharded tensors: MemoryConfigAttr with DRAM + HeightSharded/WidthSharded, ToMemoryConfigOp converting to/from DRAM sharded
- Add Python tests for end-to-end DRAM-sharded matmul (matching tt-metal's test_matmul_dram_sharded.py pattern)
- Test round-trip: MLIR assembly → flatbuffer → runtime deserialization with DRAM-sharded configs

Issue #8033, opened 2026-04-20
Add DRAM sharding helpers and update withBufferType()


sdjordjevicTT:
### Problem
Several helper functions assume sharding = L1. withBufferType(DRAM) forcibly resets the grid to 1x1, destroying any shard configuration. There's no hasShardedDRAMTensorMemoryLayout() counterpart to the existing L1 version.

### What needs to change

- TTNNLayoutAttr::withBufferType() (lib/Dialect/TTNN/IR/TTNNOpsAttrs.cpp:477-482) — don't reset grid to 1x1 for DRAM when the layout is sharded; use the DRAM grid instead
- Add hasShardedDRAMTensorMemoryLayout() helper in TTNNOpsAttrs.h alongside the existing L1 version
- Audit all callers of hasShardedL1TensorMemoryLayout() (TTNNOpsAttrs.cpp:62-67) — determine if each should also handle DRAM sharding or needs a new hasShardedTensorMemoryLayout() that covers both
- Update producesShardedL1Layout() (lib/Dialect/TTNN/Utils/Utils.cpp:368-371)
- Update isShardedMemoryLayout() (include/ttmlir/Dialect/TTNN/IR/TTNNOpsAttrs.h:35-39)

Issue #8031, opened 2026-04-20
Relax TTNN type system to allow DRAM sharded layouts


sdjordjevicTT:
### Problem

The TTNN dialect verification logic hardcodes that DRAM must be interleaved and sharding is L1-only. These blocks represent DRAM-sharded tensors in the IR, even though tt-metal fully supports them.

#### What needs to change

- verifyBufferAndMemoryLayout() (lib/Dialect/TTNN/IR/TTNNOpsAttrs.cpp:116-119) — allow DRAM + HeightSharded/WidthSharded (not BlockSharded, per tt-metal constraints)
- verifySharding() (lib/Dialect/TTNN/IR/TTNNOpsAttrs.cpp:139-141) — allow ShardSpec with DRAM buffer type
- verifyGridShape() (lib/Dialect/TTNN/IR/TTNNOpsAttrs.cpp:85-99) — allow DRAM buffers to have non-unit grids (the DRAM bank grid from DeviceAttr::getDramGrid())
- ToMemoryConfigOp::verify() (lib/Dialect/TTNN/IR/TTNNOps.cpp:1992-2001) — relax both DRAM-Interleaved-only and L1-only sharding constraints

PR #8285, opened 2026-05-06
[TTNN] Enable DRAM tensor sharding

dmilinkovicTT:
### Ticket
#8031, #8033 and #8036.

### Problem description
The TTNN dialect hardcoded `DRAM => Interleaved` and `Sharded => L1` assumptions, blocking representation of DRAM-sharded tensors that tt-metal supports.

This PR enables compiler-level support for DRAM-sharded layouts. This PR **doesn't** cause DRAM-sharded layouts to be emitted in any of the existing paths - that's a separate topic driven by the optimizer team.

### What's changed
- **Relax the TTNN type system**: verifiers now accept DRAM + sharded layouts; `to_memory_config` accepts DRAM↔DRAM reshards; canonical CRS builder places DRAM shards on a single row of the device's DRAM bank grid.
- **Runtime / OpModel / EmitC:** `CoreRangeSet` uses `std::vector` instead of `std::set` so DRAM-bank-to-worker order survives flatbuffer round-trip.
- **TTNNDecomposeLayouts** - route untilization to host for DRAM-sharded inputs, since `ttnn::untilize` doesn't support DRAM-sharded inputs - [tt-metal issue](https://github.com/tenstorrent/tt-metal/issues/43975).
- **ttnn_builder.py:** exposed sharded-layout params and extended `to_layout` op builder, so that we're able to test the DRAM-sharded layouts though the builder.

### Checklist

### DIFF
--- a/lib/Dialect/TTNN/IR/TTNNOps.cpp
+++ b/lib/Dialect/TTNN/IR/TTNNOps.cpp
@@ -2000,16 +2000,7 @@ static bool isValidDeviceLayout(TensorMemoryLayoutAttr memLayoutAttr) {
                        "sharded memory layouts");
   }
 
-  if (outputBufferType == BufferType::DRAM &&
-      outputMemoryLayout.getValue() != TensorMemoryLayout::Interleaved) {
-    return emitOpError(
-        "Device DRAM memory space only supports interleaved memory layout");
-  }
-
   if (outputLayout.hasShardedTensorMemoryLayout()) {
-    if (not outputLayout.hasShardedL1TensorMemoryLayout()) {
-      return emitOpError("Sharded tensors layout must reside in L1");
-    }
     ::llvm::SmallVector<int64_t> shardShape = outputLayout.getShardShape();
     // Currently TTNN backend only supports 2D shard shape
     if (shardShape.size() != 2) {
--- a/lib/Dialect/TTNN/IR/TTNNOps.cpp
+++ b/lib/Dialect/TTNN/IR/TTNNOps.cpp
@@ -1999,7 +1999,11 @@ static bool isValidDeviceLayout(TensorMemoryLayoutAttr memLayoutAttr) {
     return emitOpError("Device memory space only supports interleaved or "
                        "sharded memory layouts");
   }
-
+  if (outputBufferType == BufferType::DRAM &&
+      outputMemoryLayout.getValue() == TensorMemoryLayout::BlockSharded) {
+    return emitOpError(
+        "DRAM memory space doesn't support block sharded memory layout");
+  }
   if (outputLayout.hasShardedTensorMemoryLayout()) {
     ::llvm::SmallVector<int64_t> shardShape = outputLayout.getShardShape();
     // Currently TTNN backend only supports 2D shard shape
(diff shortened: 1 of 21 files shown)


mtopalovicTT (approved):

rpavlovicTT (approved):

lib/Dialect/TTNN/IR/TTNNOps.cpp
```diff
@@ -1996,16 +1996,7 @@ ::mlir::LogicalResult mlir::tt::ttnn::ToMemoryConfigOp::verify() {
                        "sharded memory layouts");
   }
 
-  if (outputBufferType == BufferType::DRAM &&
-      outputMemoryLayout.getValue() != TensorMemoryLayout::Interleaved) {
-    return emitOpError(
-        "Device DRAM memory space only supports interleaved memory layout");
-  }
```
bmalesevicTT, line 2007: In TTNNOpsAttrs.cpp it is noted that DRAM block sharding is not allowed, it probably has no path to reach here, but would make sense to have the check here too.

test/ttmlir/Dialect/TTNN/dram_sharded_layout.mlir
```diff
@@ -0,0 +1,64 @@
+// RUN: ttmlir-opt --split-input-file %s | FileCheck %s
+
+// Verify that DRAM-sharded TTNN layouts parse, verify, and round-trip cleanly.
+
+// Width-sharded across 12 DRAM banks (Blackhole-style dram grid).
```
bmalesevicTT: 12 DRAM banks is more common for Wormhole?

tools/builder/ttnn/ttnn_builder.py
```diff
@@ -285,15 +303,86 @@ def _create_memory_config_attr(
...
+    def _derive_canonical_core_range_set(
+        self,
+        buffer_type: ttnn.ir.BufferType,
+        grid_shape: List[int],
+        tensor_memory_layout: Optional[ttnn.ir.TensorMemoryLayout] = None,
+    ) -> ttnn.ir.CoreRangeSetAttr:
+        """
+        Canonical single-rectangle placement for a sharded layout's grid_shape.
+          - L1: mirrors the virtual grid orientation:
+            (0,0) -> (grid_shape[1]-1, grid_shape[0]-1).
```
bmalesevicTT: Doesn't each case of L1 sharding except block-sharded do row-major placement according to tt-metal? Which would mean height-sharding doesn't align with this approach at all while width-sharding is the one that needs adjustment if > worker_grid_width

jgrimTT (approved):
Builder changes look good, thanks!

tools/builder/ttnn/ttnn_builder.py
dmilinkovicTT: Yep, this helper only covers the subset of scenarios which don't require knowing the device worker grid dimensions. 

I've reconsidered the approach here, and, since we can't determine the worker grid dimensions in the TTNN builder infra (since the ops are getting instantiated before the device is created), I've removed this helper and added requirement for manually passing CRS in the tests.

Support for querying the system descriptor through TTNN builder infra should allow us to access the worker / DRAM grid dimensions and populate CRS automatically, but I have to check this with the tooling team, since it looks like a large change.

Thanks for the heads up!

sdjordjevicTT (approved):

pilkicTT (approved):

tapspatel (approved):

PR #8287, opened 2026-05-06
[Matmul,Linear] Drop top-level activation when program config carries fused activation

rpavlovicTT:
## Summary
- When the matmul program config (`MatmulMultiCoreReuseMultiCast{,1D}` or the two `DRAMSharded` variants) already carries a `fused_activation`, also passing a top-level activation trips tt-metal's `!parameters.user_fused_activation.has_value()` `TT_FATAL` for sharded inputs (`ttnn/operations/matmul/matmul.cpp:163`). On the OpModel side the beam search spuriously rejects every sharded matmul candidate; at runtime the op crashes once beam search picks the sharded path.
- Add a `programCarriesFusedActivation` helper in both OpModel (`namespace detail`, shared by `OpModel<MatmulOp>` and `OpModel<LinearOp>`) and runtime (duplicated because it links a separate translation unit of the same `MatmulProgramConfig` types) that suppresses the top-level activation string when the program config already carries one.
- Linear goes through the same `bound_matmul()` path as Matmul (`matmul.cpp:382`), so it needs the same fix in both OpModel and runtime. Sparse matmul is unaffected: its public API takes no `activation`, it bypasses `bound_matmul`, and pins `user_fused_activation` to `nullopt` internally.

Effect on Gemma-4 31B vLLM (Blackhole BHQB 4-chip): matmul/linear candidates with sharded operand 0 + fused activation are no longer rejected during op-model validation, and the runtime no longer crashes when beam search picks them.

## Test plan
- [x] `TestMatmulBlockShardedConstraint` — 2/2 passed (`MatmulBlockShardedInputWithPadding`, `MatmulActivationWithIgnorePhysicalLayout`)
- [x] `TestOpModelInterface --gtest_filter="*Linear*:*Matmul*"` — 6/6 passed
- [x] `TestOpModelLib --gtest_filter="*Linear*:*Matmul*"` — 30/30 passed
- [ ] CI: full unit + lit suites
- [ ] tt-xla validation against this PR

### DIFF
--- a/lib/OpModel/TTNN/TTNNOpModel.cpp
+++ b/lib/OpModel/TTNN/TTNNOpModel.cpp
@@ -539,6 +539,36 @@ auto getOpSymbol() {
   }
 }
 
+// Returns true if the matmul program config already carries a fused
+// activation.
+inline bool programCarriesFusedActivation(
+    const std::optional<::ttnn::operations::matmul::MatmulProgramConfig> &pc) {
+  if (!pc) {
+    return false;
+  }
+  return std::visit(
+      [](const auto &cfg) -> bool {
+        using T = std::decay_t<decltype(cfg)>;
+        if constexpr (
+            std::is_same_v<T, ::ttnn::operations::matmul::
+                                  MatmulMultiCoreReuseMultiCastProgramConfig> ||
+            std::is_same_v<T,
+                           ::ttnn::operations::matmul::
+                               MatmulMultiCoreReuseMultiCast1DProgramConfig> ||
+            std::is_same_v<
+                T, ::ttnn::operations::matmul::
+                       MatmulMultiCoreReuseMultiCastDRAMShardedProgramConfig> ||
+            std::is_same_v<
+                T,
+                ::ttnn::operations::matmul::
+                    MatmulMultiCoreReuseMultiCastBatchedDRAMShardedProgramConfig>) {
+          return cfg.fused_activation.has_value();
+        }
+        return false;
+      },
+      *pc);
+}
+
 } // namespace detail
 #endif // TTMLIR_ENABLE_OPMODEL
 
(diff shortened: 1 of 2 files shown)


runtime/lib/ttnn/operations/matmul/matmul.cpp
```diff
@@ -15,6 +15,41 @@
 
 namespace tt::runtime::ttnn::operations::matmul {
 
+// If the program config already carries the fused activation (kernel-fused
+// path, the only legal one for sharded inputs per
+// ttnn/operations/matmul/matmul.cpp:163), drop the top-level activation
+// string — passing both triggers the
+// `!parameters.user_fused_activation.has_value()` TT_FATAL when input is
+// sharded.  See TTNNOpModel.cpp::OpModel<MatmulOp>::getOpConstraints for the
+// matching beam-search-side fix.
```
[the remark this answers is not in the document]
rpavlovicTT: Not needed — the optimizer's beam search calls `applyOpSpecificAttrs` after selecting a candidate, which already clears the op-level `activation` attribute when the program config carries a fused activation (see `MatmulRules::applyOpSpecificAttrs`). By the time EmitPy/EmitC or the runtime serializer see the op, `activation` is already absent.

sgholamiTT (approved):

runtime/lib/ttnn/operations/matmul/matmul.cpp
```diff
@@ -15,6 +15,34 @@
 
 namespace tt::runtime::ttnn::operations::matmul {
 
+static bool programCarriesFusedActivation(
```
nvukobratTT, line 18: This duplication is unfortunate... If there is no other option, can we at least link them via function comments in both places? E.g., 
```
// Keep in sync with lib/OpModel/TTNN/TTNNOpModel.cpp::programCarriesFusedActivation
```

lib/OpModel/TTNN/TTNNOpModel.cpp
```diff
@@ -4549,15 +4579,16 @@ llvm::Expected<OpConstraints> OpModel<LinearOp>::getOpConstraints(
...
-  // Convert activation string to optional
-  std::optional<std::string> activationStr =
-      activation ? std::make_optional(activation->str()) : std::nullopt;
-
   // Convert program config attribute
   auto programConfig =
       programConfigAttr ? conversion::getMatmulProgramConfig(*programConfigAttr)
                         : std::nullopt;
 
+  std::optional<std::string> activationStr;
```
nvukobratTT, line 4587: A more general comment (I don't have the full width of the context here), when the top-level activation is dropped because the program config already carries one, can this produce some side effects in the rest of the runtime or optimizer? Should we log when this behaviour occurs?

runtime/lib/ttnn/operations/matmul/matmul.cpp
nvukobratTT: @pilkicTT @acolicTT any better ideas how to handle this case (ideally de-duplicate)?

nvukobratTT (approved):

runtime/lib/ttnn/operations/matmul/matmul.cpp
rpavlovicTT: Actually @pljubisavljevicTT's internship project strives to solve this duplication.

lib/OpModel/TTNN/TTNNOpModel.cpp
rpavlovicTT: No, it can't produce side effects. This is correctness fix, as carrying both would result in incorrect op's behaviour in runtime (or exception).

PR #8288, opened 2026-05-06
[TTIR] Cache valueTracesToConstantArgs in EraseInverseOps

sshonTT:
### Ticket
Link to Github Issue

### Problem description
TTIREraseInverseOps calls ttcore::valueTracesToConstantArgs from nearly every pattern match attempt, and each call walks the value's full use-def chain back to a function block argument. With O(num_ops) match attempts and an O(num_ops) walk per call, the pass is O(num_ops^2) in the size of the function — which dominates compile time on deep models.

### What's changed
Add ConstevalForwardAnalysis: a forward-propagation analysis that walks each func::FuncOp once in SSA order and propagates (hasConstOrParamArg, hasNonConstOrParamArg) flags from block args through op results. valueTracesToConstantArgs becomes an O(1) hash lookup, dropping the pass from O(num_ops^2) to O(num_ops) per rewrite epoch.

Correctness under pattern rewriters that mutate the IR is preserved by extending RewriterBase::Listener: any insert / modify / replace / erase flips a dirty flag and the next lookup rebuilds the cache from scratch, sidestepping MLIR's Value pointer recycling. Cache misses fall through to the original recursive walk so values created between rebuilds still get a correct answer.

The analysis is installed on the greedy driver via GreedyRewriteConfig::setListener and made visible to the inline valueTracesToConstantArgs helper through a thread_local pointer managed by ConstevalAnalysisScope.

### Checklist


lib/Dialect/TTCore/IR/Utils.cpp
```diff
@@ -236,4 +236,105 @@ getBufferType(Type type, bool isView, std::optional<MetalLayoutAttr> hostInfo) {
...
 }
 
+//===----------------------------------------------------------------------===//
+// ConstevalForwardAnalysis
+//===----------------------------------------------------------------------===//
+
+namespace {
+// Thread-local pointer to the currently-active analysis. Set by
+// ConstevalAnalysisScope. Read by valueTracesToConstantArgs.
+thread_local ConstevalForwardAnalysis *g_activeAnalysis = nullptr;
```
vmilosevic: :warning: **cppcoreguidelines\-avoid\-non\-const\-global\-variables** :warning:
variable `` g_activeAnalysis `` is non\-const and globally accessible, consider making it const

vmilosevic: :warning: **cppcoreguidelines\-avoid\-non\-const\-global\-variables** :warning:
variable `` g_activeAnalysis `` provides global access to a non\-const object; consider making the pointed\-to data `` const ``

vmilosevic (commented):

mvasiljevicTT (commented):
EIO changes are widely affecting models, please run and link all perf benchmark tests to confirm correctness

lib/Dialect/TTCore/IR/Utils.cpp
```diff
@@ -236,4 +236,105 @@ getBufferType(Type type, bool isView, std::optional<MetalLayoutAttr> hostInfo) {
...
+          // Treat pessimistically as non-const so the analysis returns
+          // `false` here; valueTracesToConstantArgs will fall back to the
+          // slow path on lookup of std::nullopt-style results.
+          hasNonConst = true;
+          continue;
+        }
+        hasConst |= it->second.first;
+        hasNonConst |= it->second.second;
+      }
+
```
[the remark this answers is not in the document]
sshonTT: Comment in original code misleading the logic. Change the comment.

include/ttmlir/Dialect/TTCore/IR/Utils.h
```diff
@@ -88,10 +90,96 @@ getConstsAndParams(mlir::func::FuncOp funcOp) {
...
+public:
+  // Captures `root` (typically a func::FuncOp). `lookup` lazily builds the
+  // cache on first use and rebuilds whenever a Listener notification has
+  // marked it dirty.
+  explicit ConstevalForwardAnalysis(mlir::Operation *root);
+
+  // Returns std::nullopt if the value is unknown to this analysis (e.g.
+  // created after the most recent rebuild or in an unvisited region).
+  // Callers should fall back to the slow path on nullopt.
+  std::optional<bool> lookup(mlir::Value v);
```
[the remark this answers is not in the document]
sshonTT: I've updated gtest to validate current logic vs og logic. In terms of compilation time, we don't have good way to measure it. Maybe following work can figure out this one.

sshonTT:
Measured compilation time with the DeepSeek v4 Flash model
 - Number of TTIR ops: 56k
 - Baseline: ~1000s
 - With this patch: 61s (17x faster)

sshonTT:
Hi @mvasiljevicTT , here are the benchmark test links
 - n300,llmbox : https://github.com/tenstorrent/tt-xla/actions/runs/25454428859
 - n150 : https://github.com/tenstorrent/tt-xla/actions/runs/25496640592

Fails in n150 tests, that seems already regression in nightly or server issues. So I think it's good to go.

include/ttmlir/Dialect/TTCore/IR/Utils.h
```diff
@@ -88,10 +90,102 @@ getConstsAndParams(mlir::func::FuncOp funcOp) {
...
+  ConstevalForwardAnalysis *previous_;
+};
+
+// Internal: returns the active scope's analysis, or nullptr if none.
+ConstevalForwardAnalysis *getActiveConstevalAnalysis();
+
 // This function will return true if a given Value is the result of operations
 // performed only between  block arguments in which have been marked as
 // consteval-able (Parameter or Constant ArgumentType).
 inline bool valueTracesToConstantArgs(const mlir::Value &value) {
```
dmilinkovicTT, line 94: This simple inline helper now has a side effect of using `getActiveConstevalAnalysis`, but the signature of the function in no way hints the user about this.

I think that leaving this helper as-is, and scoping the caching logic to the EIO pass privately would be a more robust solution.

include/ttmlir/Dialect/TTCore/IR/Utils.h
```diff
@@ -88,10 +90,102 @@ getConstsAndParams(mlir::func::FuncOp funcOp) {
...
+class ConstevalAnalysisScope {
+public:
+  explicit ConstevalAnalysisScope(ConstevalForwardAnalysis &analysis);
+  ~ConstevalAnalysisScope();
+  ConstevalAnalysisScope(const ConstevalAnalysisScope &) = delete;
+  ConstevalAnalysisScope &operator=(const ConstevalAnalysisScope &) = delete;
+
+private:
+  ConstevalForwardAnalysis *previous_;
+};
```
dmilinkovicTT: What concurrent modifications are we anticipating here, against which we need to defend ourselves?

MLIR infra exposes multi-threading in two similar ways:
- Passes targeting `ModuleOp` targeting **different** modules can run concurrently
- Passes targeting `FuncOp` targeting **different** functions can run concurrently

Concurrent modification is technically possible, but since passes naturally act on different modules/functions, the pass implementation would itself reveal the anti-pattern by reaching outside of the targeted module/function to modify the sibling ops.

EIO pass targets `ModuleOp` and  iteratively goes through `FuncOp`s  - pure single-threaded pass. Concurrent modifications are impossible here.

IMHO, dropping the mutlti-threaded  considerations from this PR would simplify it significantly.

dmilinkovicTT (commented):
Awesome discovery, thanks!

TLDR, two concerns:
- concurrent modifications shouldn't be possible, so we shouldn't need to scope the cache per thread
- the modified helper and the caching logic itself should, IMHO, live inside EIO pass as private implementation details - polluting the logic of a simple inline helper with complex caching logic is probably an overkill, since most of the callers won't come close the performance bottlenecks exposed by EIO

dmilinkovicTT:
@sshonTT btw, I assume @mvasiljevicTT suggested that you compare the performance results from the benchmark runs against the baseline and attach the comparison results here, so we can be sure that no perf regressions are introduced.

sshonTT:
@dmilinkovicTT Thanks for the great review, and yeah if we can scope out multithread, the pr will be much clearer, I will work that way. 
BTW, the benchmark test results I posted, I ran it with `Perf regression check: true` which means if there is a regression on the perf it should be failed.

include/ttmlir/Dialect/TTCore/IR/Utils.h
sshonTT: @dmilinkovicTT Addressed:

 - ttcore::valueTracesToConstantArgs reverted to its original body (no getActiveConstevalAnalysis side effect).
 - Cache (ConstevalForwardAnalysis, scope guard, cache-aware helper) moved to lib/Dialect/TTIR/Transforms/EraseInverseOps/ under mlir::tt::ttir.
 - Dropped thread_local -- EIO is single-threaded, plain file-static suffices.
 - EIO patterns call ttir::valueTracesToConstantArgs (falls back to ttcore:: on miss); non-EIO callers unchanged.

include/ttmlir/Dialect/TTCore/IR/Utils.h
sshonTT: Yeah, that is great point. I pushed the change to make it simple.

include/ttmlir/Dialect/TTIR/Transforms/EraseInverseOps/ConstevalForwardAnalysis.h
```diff
@@ -0,0 +1,125 @@
...
+#include "mlir/IR/Builders.h"
+#include "mlir/IR/PatternMatch.h"
+#include "mlir/IR/Value.h"
+#include "llvm/ADT/DenseMap.h"
+
+#include <optional>
+
+namespace mlir::tt::ttir {
+
+// Forward-propagation analysis for `ttcore::valueTracesToConstantArgs`.
```
dmilinkovicTT: This line isn't true anymore

include/ttmlir/Dialect/TTIR/Transforms/EraseInverseOps/ConstevalForwardAnalysis.h
```diff
@@ -0,0 +1,125 @@
...
+//
+// Semantics match `ttcore::valueTracesToConstantArgs` bug-for-bug: a value
+// traces to constant args iff its use-def chain reaches at least one
+// CONST/PARAM block argument and no non-CONST/PARAM block argument.
+//
+// Scope: private to the TTIREraseInverseOps pass. Made visible to commute
+// pattern files via the inline `valueTracesToConstantArgs` below. EIO is a
+// single-threaded pass (it walks func::FuncOps sequentially), so a plain
+// file-static "active analysis" pointer is sufficient -- no thread-local
+// storage required.
```
dmilinkovicTT: These Claude-generated comments are often too verbose and are adding some noise.

Many parts of this comment are focusing on unimportant implementation details, some stale details from past implementations (e.g. thread-local storage) etc.

Could we please clean up this comment?

include/ttmlir/Dialect/TTIR/Transforms/EraseInverseOps/ConstevalForwardAnalysis.h
```diff
@@ -0,0 +1,125 @@
...
+  // created after the most recent rebuild or in an unvisited region).
+  // Callers should fall back to the slow path on nullopt.
+  std::optional<bool> lookup(mlir::Value v);
+
+  size_t size() const { return cache_.size(); }
+
+  // Number of times the cache has been (re)built. Useful for tests that
+  // verify the listener-based invalidation actually keeps the cache live
+  // across pattern rewrites.
+  size_t rebuildCount() const { return rebuildCount_; }
```
dmilinkovicTT: Adding a public API function just for the sake of testing is an anti-pattern

include/ttmlir/Dialect/TTIR/Transforms/EraseInverseOps/ConstevalForwardAnalysis.h
```diff
@@ -0,0 +1,125 @@
...
+class ConstevalForwardAnalysis : public mlir::RewriterBase::Listener {
+public:
+  // Captures `root` (typically a func::FuncOp). `lookup` lazily builds the
+  // cache on first use and rebuilds whenever a Listener notification has
+  // marked it dirty.
+  explicit ConstevalForwardAnalysis(mlir::Operation *root);
+
+  // Returns std::nullopt if the value is unknown to this analysis (e.g.
+  // created after the most recent rebuild or in an unvisited region).
+  // Callers should fall back to the slow path on nullopt.
```
dmilinkovicTT: nit: but we lack an explanation of what it returns if value is known to this analysis, and it isn't clear from the function name either

include/ttmlir/Dialect/TTIR/Transforms/EraseInverseOps/ConstevalForwardAnalysis.h
```diff
@@ -0,0 +1,125 @@
...
+
+private:
+  void rebuild();
+
+  mlir::Operation *root_;
+  bool dirty_ = true;
+  size_t rebuildCount_ = 0;
+  // Per-value: (hasConstOrParamArg, hasNonConstOrParamArg). We need both bits
+  // because the result is `hasConst && !hasNonConst`.
+  llvm::DenseMap<mlir::Value, std::pair<bool, bool>> cache_;
```
dmilinkovicTT: nit 2: a map of structs with two boolean fields would be more readable afterwards.

lib/Dialect/TTIR/Transforms/EraseInverseOps/ConstevalForwardAnalysis.cpp
```diff
@@ -0,0 +1,121 @@
...
+}
+
+std::optional<bool> ConstevalForwardAnalysis::lookup(mlir::Value v) {
+  if (dirty_) {
+    rebuild();
+  }
+  auto it = cache_.find(v);
+  if (it == cache_.end()) {
+    return std::nullopt;
+  }
```
dmilinkovicTT, line 80: Just curious, does this actually ever happen?

Since we're rebuilding the entire cache after any IR modification, I assume not?

include/ttmlir/Dialect/TTIR/Transforms/EraseInverseOps/ConstevalForwardAnalysis.h
```diff
@@ -0,0 +1,125 @@
...
+  ~ConstevalAnalysisScope();
+  ConstevalAnalysisScope(const ConstevalAnalysisScope &) = delete;
+  ConstevalAnalysisScope &operator=(const ConstevalAnalysisScope &) = delete;
+
+private:
+  ConstevalForwardAnalysis *previous_;
+};
+
+// Internal: returns the active scope's analysis, or nullptr if none.
+ConstevalForwardAnalysis *getActiveConstevalAnalysis();
```
dmilinkovicTT: If I understand this correctly, the only reason we're introducing these scoping abstractions is because we want the analysis instances to be RAII-style `FuncOp`-scoped.

But at the moment of running EIO pass, we should have **exactly one function**:
- inliner pass is run at the start of the pipeline
- const-eval pass hasn't run yet

Which means that dropping the `FuncOp`-scoping requirement would simplify things drastically:
- `ConstevalForwardAnalysis` is scoped to the `EIO` pass instance (as a private member)
- the instance is passed to the rewrite pattern instances through constructor
- we drop the top-level inline `valueTracesToConstantArgs`, and instead make it a method on the `ConstevalForwardAnalysis`
- we drop the `ConstevalAnalysisScope` and related helpers.

dmilinkovicTT:
> I ran it with Perf regression check: true which means if there is a regression on the perf it should be failed.

ah okay, sorry, didn't know that this option existed

include/ttmlir/Dialect/TTIR/Transforms/EraseInverseOps/ConstevalForwardAnalysis.h
sshonTT: Right, I've fixed up the comments in PR.

include/ttmlir/Dialect/TTIR/Transforms/EraseInverseOps/ConstevalForwardAnalysis.h
sshonTT: Done. rebuildCount() is removed entirely along with the underlying counter.
The two tests that depended on it (CacheReusedAcrossLookups and the rebuild-count assertions in ListenerInvalidatesOnMutation) were dropped or rewritten to check the externally-observable behavior (post-mutation answers match the slow-path baseline) rather than the rebuild count.

include/ttmlir/Dialect/TTIR/Transforms/EraseInverseOps/ConstevalForwardAnalysis.h
sshonTT: Resolved by the API rewrite. lookup() returning std::optional<bool> is gone, replaced by bool valueTracesToConstantArgs(Value) whose name and signature describe the return value directly.

include/ttmlir/Dialect/TTIR/Transforms/EraseInverseOps/ConstevalForwardAnalysis.h
sshonTT: Done. The cache is now DenseMap<Value, Reachability> where Reachability is a struct with named hasConstOrParamArg / hasNonConstOrParamArg bools, and all callsites read them by name.

lib/Dialect/TTIR/Transforms/EraseInverseOps/ConstevalForwardAnalysis.cpp
sshonTT: You're right, it shouldn't happen in EIO. I will convert this to assert instead of silent fallback.

include/ttmlir/Dialect/TTIR/Transforms/EraseInverseOps/ConstevalForwardAnalysis.h
sshonTT: Yeah, I re-designed it following your suggestion.

include/ttmlir/Dialect/TTCore/IR/Utils.h
```diff
@@ -107,6 +107,14 @@ inline bool valueTracesToConstantArgs(const mlir::Value &value) {
   }
 
   for (auto blockArg : subgraphBlockArgs) {
+    // Require ownership by the same func::FuncOp before consulting its
+    // argument attributes; otherwise an inner-region block argument could
+    // accidentally be classified as CONST/PARAM via funcOp.getArgAttr at
+    // a matching argNumber. Non-function block arguments never carry
+    // ttcore::ArgumentType so the value cannot be const-eval-traceable.
+    if (blockArg.getOwner()->getParentOp() != funcOp.getOperation()) {
```
mtopalovicTT, line 115: Nice find. How did you detect this? I'm not aware of any op in ttir which is lowered into ttnn which has block inside itself but I might be wrong

include/ttmlir/Dialect/TTIR/Transforms/EraseInverseOps/ConstevalForwardAnalysis.h
```diff
@@ -0,0 +1,63 @@
...
+
+private:
+  void rebuild();
+
+  // Tracks both bits: a value traces to constant args iff its use-def chain
+  // reaches at least one CONST/PARAM arg and no non-CONST/PARAM arg.
+  struct Reachability {
+    bool hasConstOrParamArg = false;
+    bool hasNonConstOrParamArg = false;
+  };
```
mtopalovicTT, line 54: Why do you need both of these? tracesToConstOrParam checks if all paths trace back to const or param arg. We could just store DenseMap of value to bool where bool would mark if all paths from value back to block arg trace to constant or param right?

lib/Dialect/TTIR/Transforms/EraseInverseOps/ConstevalForwardAnalysis.cpp
```diff
@@ -0,0 +1,84 @@
...
+  dirty_ = false;
+}
+
+bool ConstevalForwardAnalysis::valueTracesToConstantArgs(mlir::Value v) {
+  if (dirty_) {
+    rebuild();
+  }
+  auto it = cache_.find(v);
+  if (it == cache_.end()) {
+    llvm_unreachable(
```
mtopalovicTT: Why not call here: return ttcore::valueTracesToConstantArgs(v) or just return false pessimistically?

mtopalovicTT (approved):
Approving. I would wait for other folks signoff.

include/ttmlir/Dialect/TTCore/IR/Utils.h
sshonTT: Honestly I implemented this with claude, I don't remember the specific case that hits this,  it was during DeepSeek-V4 bring-up, a pattern was misclassifying a value inside a nested-region as constant-traceable because the inner block arg's argNumber happened to collide with a CONST/PARAM index on the surrounding func::FuncOp, and isConstOrParamArg(blockArg, funcOp) blindly queried funcOp.getArgAttr(blockArg.getArgNumber(), ...).

include/ttmlir/Dialect/TTIR/Transforms/EraseInverseOps/ConstevalForwardAnalysis.h
sshonTT: The two bits preserve the exact semantics of ttcore::valueTracesToConstantArgs, which returns false when a value's use-def chain never reaches any function block argument. TTIR creation ops with zero inputs (ttir.empty, ...) hit this case

With a single tracesAllToConst bool propagated as AND of operand bools, a zero-operand op aggregates to vacuous true, flipping the answer for every creation op result. Tracking hasConstOrParamArg and hasNonConstOrParamArg separately keeps the cache bit-for-bit equivalent to the slow path:

| Case | 1-bit answer | Actual answer |
| :--- | :---: | :---: |
| Chain reaches const arg, never reaches non-const | true ✓ | true |
| Chain reaches both const and input | false ✓ | false |
| Chain reaches no block arg at all (`ttir.empty`, `ttir.arange`, ...) | true ✗ | **false** |

lib/Dialect/TTIR/Transforms/EraseInverseOps/ConstevalForwardAnalysis.cpp
sshonTT: right it came from [here](https://github.com/tenstorrent/tt-mlir/pull/8288/changes#r3236302391). but I think it was too aggressive. Switched this to slow-path fallback(return ttcore::valueTracesToConstantArgs(v);)

dmilinkovicTT (approved):
Thanks for implementing the suggestions!

sshonTT:
@mvasiljevicTT , @odjuricicTT could you take a look? Seems I need a approval from you to get this merged in.

test/unittests/Support/CMakeLists.txt
```diff
@@ -17,5 +17,7 @@ target_link_libraries(TTMLIRSupportTests
     MLIRIR
     MLIRFuncDialect
     MLIRTransforms
+    MLIRParser
```
mvasiljevicTT: Do we still need this?

mvasiljevicTT (approved):
great improvement!

test/unittests/Support/CMakeLists.txt
sshonTT: Good catch. I dropped MLIRParser and MLIRTTCoreDialect from this list.

PR #8297, opened 2026-05-06
[d2m] stale virtual grid fixes

dlokeTT:
### Problem description
Ran into problems with stale virtual grid metadata in grid selection and lower to layout.

### What's changed
- Added a helper that derives the new virtual grid mapping from the selected grid at rewrite.
- In `lowerToLayout.cpp`, don't use stale vgm attributes just when just grid shape matches. Only copy the vgm attrs if the maps structurally match the new type or just rederive the vgm.
- Added lit tests

### Checklist


anuragsingh-tt (approved):

vwellsTT (approved):
nit: can you just do like `TODO (#8301)` instead? I think that's more standard for us

PR #8299, opened 2026-05-06
Move TopK fusing to TTIR and rename FusionValidation to OpValidation

sdjordjevicTT:
### Ticket
Solves https://github.com/tenstorrent/tt-mlir/issues/8298

### Summary
- Move TopK fusing pattern from TTNN to TTIR dialect, with a corresponding TTNN decomposition pattern that validates the fused op via OpModel and decomposes if unsupported
- Rename FusionValidator/FusionValidationConfig/FusionValidationResult to OpValidator/OpValidationConfig/OpValidationResult and move from Transforms/Fusing/ to Transforms/, since the validator is used by both fusing and decomposition patterns
- Simplify pipeline branching in TTNNPipelines.cpp by deduplicating the fallback createTTNNDecomposition() path

### Checklist


mtopalovicTT (approved):

lib/Dialect/TTNN/Transforms/Decomposition/TopKDecompositionRewritePattern.cpp
```diff
@@ -0,0 +1,102 @@
...
+  // OpValidator. If validation succeeds, keep the op as-is.
+  if (validationConfig.has_value()) {
+    OpValidator validator(rewriter.getContext(), *validationConfig);
+
+    auto validationResult = validator.validateOp<ttnn::TopKOp>(
+        topkOp.getOperation(), topkOp.getLoc(),
+        {valuesResultType, indicesResultType}, topkOp.getInputTensor(),
+        topkOp.getKAttr(), topkOp.getDimAttr(), topkOp.getLargestAttr(),
+        topkOp.getSortedAttr(),
+        /*memory_config=*/nullptr);
```
rpavlovicTT, line 36: Without getting into details, I am wondering do we have to list out all of these op parameters here? This creates another bug surface like we have between OpModel vs Runtime. So we'd have to keep even this invocation point in mind when it comes to `TopK` op.

sdjordjevicTT: This is strictly a compiler interface API endpoint that we need to target in order to produce the op module in isolation. Not sure if there is a better way to tackle this?

include/ttmlir/Dialect/TTNN/Transforms/OpValidator.h
```diff
@@ -38,10 +38,10 @@ struct FusionValidationConfig {
   uint32_t maxFallbackAttempts = 10000;
 };
 
-/// Result of fusion validation.
-struct FusionValidationResult {
+/// Result of op validation.
+struct OpValidationResult {
```
rpavlovicTT, line 42: We already have this name used in `OpValidationLib`. Can you rename it or wrap in a new namespace to clearly distinguish them?

include/ttmlir/Support/Logger.h
```diff
@@ -41,7 +41,7 @@ enum class LogComponent {
   Test,
   General,
   D2MFusion,
-  FusionValidator,
+  OpValidator,
```
rpavlovicTT: Too similar to `OpValidation` above, please rename.

lib/Dialect/TTNN/Transforms/Decomposition/TopKDecompositionRewritePattern.cpp
rpavlovicTT: I see, I thought initially this goes to OpModel directly, but is actually passed through OpValidator which creates a sub-function with this op and runs ValidationAndFallback pass on it, is this correct?

include/ttmlir/Dialect/TTNN/Transforms/OpValidator.h
```diff
@@ -52,51 +52,50 @@ struct FusionValidationResult {
...
+  static OpValidationResult failure(Status status, const std::string &msg) {
     return {status, msg};
   }
 };
 
-/// Validates a single fused op by creating it in an isolated module,
+/// Validates a single op by creating it in an isolated module,
 /// applying workaround passes, and checking op model constraints.
-class FusionValidator {
+class OpValidator {
```
rpavlovicTT: This is also too similar to OpValidationLib. Can you think of better naming here? Generally this OpValidator just creates validation func wrapper and runs ValidationFallback pass, right?

lib/Dialect/TTNN/Transforms/Decomposition/TopKDecompositionRewritePattern.cpp
sdjordjevicTT: Yes, that's correct. This is the previously named fusing validator, but I chose to rename it since it serves as a more general validator rather than a strictly fusing one.

include/ttmlir/Dialect/TTNN/Transforms/OpValidator.h
sdjordjevicTT: Yeah, let me think about a better name. I will write you offline.

sdjordjevicTT: I come up with something like this, maybe:
IsolatedIRValidator

sdjordjevicTT: Synced offline:

IsolatedIRValidationWrapper

ndrakulicTT (approved):
Seams good on tooling part

include/ttmlir/Dialect/TTNN/Transforms/OpValidator.h
```diff
@@ -2,8 +2,8 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 
-#ifndef TTMLIR_DIALECT_TTNN_TRANSFORMS_FUSING_FUSIONVALIDATOR_H
-#define TTMLIR_DIALECT_TTNN_TRANSFORMS_FUSING_FUSIONVALIDATOR_H
+#ifndef TTMLIR_DIALECT_TTNN_TRANSFORMS_OPVALIDATOR_H
```
rpavlovicTT, line 5: Rename this file according to the new name `IsolatedIRValidationWrapper`. Also fix ifndef macros.

rpavlovicTT (approved):

include/ttmlir/Dialect/TTNN/Transforms/OpValidator.h
sdjordjevicTT: Missed this, will rename in the following PR, sorry :(

PR #8311, opened 2026-05-07
LinearOp: decompose fused-bias when bias rank exceeds matmul rank

meenakshiramanathan1:


meenakshiramanathan1:
Closing this since model fails with different issue in latest main.

PR #8316, opened 2026-05-07
Fuse `ttir.pad` into `ttir.conv3d`

### LINKED ISSUE #8296: Fuse pad into conv3d padding attribute
### Summary
When frameworks like PyTorch emit F.pad() followed by F.conv3d() (common for non-standard padding modes or asymmetric padding), the resulting IR after StableHLO→TTIR lowering contains a separate ttir.pad op feeding into ttir.conv3d. These two ops should be fused into a single ttir.conv3d with the padding folded into its padding attribute, eliminating the extra pad op and its associated memory traffic.

### Current State
- ttir.conv3d already has a padding attribute ([pD, pH, pW]), but only supports symmetric padding.
- No pad→conv fusion pattern exists in the codebase for conv3d.
- A reference pattern exists for fusing stablehlo.pad into pooling ops during StableHLO→TTIR lowering (getFusablePadOp / combinePaddingFromPadOp in StableHLOToTTIRPatterns.cpp).

### Proposed Approach
Add a TTIR fusing pattern that matches ttir.pad → ttir.conv3d and folds the pad values into the conv3d padding attribute.

azecevicTT:
### Ticket
https://github.com/tenstorrent/tt-mlir/issues/8296

### Problem description
The `ttir.conv3d` op supports symmetrical zero-padding of spatiotemporal dimensions. In standard lowering through StableHLO padding and conv3d are separate ops, which are lowered separately to `ttir.pad` and `ttir.conv3d`, respectively.

### What's changed
Added a fusing pattern when `ttir.pad` symmetrically zero-pads spatiotemporal dimensions.

### Checklist

### DIFF
--- a/lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
+++ b/lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
@@ -2992,6 +2992,77 @@ public:
   }
 };
 
+// Fuse a `ttir.pad` followed by `ttir.conv3d` into a single conv3d that
+// absorbs the pad into its `padding` attribute.
+//
+// Conv3d only supports symmetric, zero-valued padding (mode "zeros"), so the
+// fusion only fires when the producing pad is symmetric, has value 0.0, and
+// pads only the spatial (depth/height/width) dimensions.
+class PadConv3dFusionPattern : public mlir::OpRewritePattern<Conv3dOp> {
+  using mlir::OpRewritePattern<Conv3dOp>::OpRewritePattern;
+
+public:
+  mlir::LogicalResult
+  matchAndRewrite(Conv3dOp convOp,
+                  mlir::PatternRewriter &rewriter) const final {
+    auto padOp = convOp.getInput().getDefiningOp<PadOp>();
+    if (!padOp || !padOp->hasOneUse()) {
+      return mlir::failure();
+    }
+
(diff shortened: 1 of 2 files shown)


djstefanovicTT (approved):

sdjordjevicTT (approved):

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
```diff
@@ -2990,6 +2990,77 @@ class ReshapeBroadcastReshapeToRepeatPattern
...
+      return mlir::failure();
+    }
+
+    if (!padOp.getValue().isZero()) {
+      return mlir::failure();
+    }
+
+    auto inputRank = padOp.getInput().getType().getRank();
+    llvm::ArrayRef<int32_t> padArr = padOp.getPadding();
+    assert(padArr.size() == 2ul * inputRank && "Invalid padding array size");
```
mtopalovicTT, line 3023: Verifier is checking this right?

mtopalovicTT (approved):

lib/Dialect/TTIR/Transforms/TTIRFusing.cpp
azecevic000: Yeah, this is just a sanity check in case the semantics of `padding` attr ever change.

PR #8323, opened 2026-05-07
Fix create-main-for-test bug

amilovanovicTT:
### Problem description
tt-mlir uplift into tt-xla exposed a bug when `create-main-for-test` pipeline flag is set to true. Namely, `TTNNCreateMainForTest` pass assumes that a forward function has a single input tensor argument (plus the device argument that this pass injects). However, when `TTNNSplitForwardFuncArgsByType`  is performed (this pass is before `TTNNCreateMainForTest` in the pipeline), a forward function signature is changed so that it has a tuple argument for activations and/or a weight dictionary argument. 

### What's changed
As a hot-fix, `TTNNCreateMainForTest` is skipped on all forward functions on which `TTNNSplitForwardFuncArgsByType` pass was performed. A long-term solution will come as a follow-up.

### Checklist

### DIFF
--- a/lib/Dialect/TTNN/Transforms/Passes.cpp
+++ b/lib/Dialect/TTNN/Transforms/Passes.cpp
@@ -723,6 +723,16 @@ public:
       return;
     }
 
+    // main_for_test currently assumes that forward function has a single
+    // tuple input (plus the device that this pass injects). When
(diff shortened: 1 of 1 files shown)


sdjordjevicTT (approved):
Approved! Just open an issue for a proper follow-up.

Issue #8291, opened 2026-05-06
Error: failed to legalize unresolved materialization from ('tensor<0x2xf64>') to ('tensor<0xcomplex<f64>>') that remained live after conversion


kamalrajkannan78:
### Describe the bug

- `stablehlo` → `ttir` conversion decomposes `complex<T>` into real-pair (`...x2xT`) on op results, but doesn't rewrite the enclosing `func.func` result types or `return` operands. The leftover `unrealized_conversion_cast` between body and signature can't be folded: `error: failed to legalize unresolved materialization from 'tensor<0x2xf64>' to 'tensor<0xcomplex<f64>>' that remained live after conversion`
- Hit by Krea Realtime 14B (sharded RoPE `freqs` from `torch.polar`) and a minimal `torch.polar` test in tt-xla.

### Repro

- In tt-xla repo,

```
git checkout kkannan/may6_handle_complex
git submodule update --init --recursive

pytest tests/torch/ops/test_polar.py --runxfail -svv
```

### Logs

- [may6_polar_sharded_after_fix.log](https://github.com/user-attachments/files/27464067/may6_polar_sharded_after_fix.log)
- **Module shlo_set_mesh_attr** IR -

```
#loc1 = loc("p0.1")
#loc2 = loc("p1.5")
module @SyncTensorsGraph.15 attributes {mhlo.cross_program_prefetches = [], mhlo.input_output_alias = [], mhlo.is_dynamic = false, mhlo.use_auto_spmd_partitioning = false} {
  sdy.mesh @mesh = <["_axis_0_aux"=1, "_axis_0"=2]> loc(#loc)
[...]

PR #8324, opened 2026-05-07
Handle sdy.return during ComplexDataTypeConversion pass

ddilbazTT:
### Ticket
#8291 

### Problem description
ComplexDataTypeConversionPass was not handling sdy.return op type.

### What's changed
Added sdy.return op type conversion pattern and wrote a test case.

### Checklist

### DIFF
--- a/lib/Dialect/StableHLO/Transforms/ComplexDataTypeConversion.cpp
+++ b/lib/Dialect/StableHLO/Transforms/ComplexDataTypeConversion.cpp
@@ -424,7 +424,7 @@ public:
 
 // Rewrites `sdy.return` to use dialect-converted operand values (same role as
 // `populateReturnOpTypeConversionPattern` for `func.return`).
-class SdyReturnOpTypeConversionPattern
+class ShardyReturnOpTypeConversionPattern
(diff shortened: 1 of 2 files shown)


lib/Dialect/StableHLO/Transforms/ComplexDataTypeConversion.cpp
```diff
@@ -421,6 +421,21 @@ class ShardyManualComputationComplexConversionPattern
     return success();
   }
 };
+
+// Rewrites `sdy.return` to use dialect-converted operand values (same role as
+// `populateReturnOpTypeConversionPattern` for `func.return`).
+class SdyReturnOpTypeConversionPattern
```
sshonTT: nit: we use full Shardy* for class name in this file (https://github.com/tenstorrent/tt-mlir/pull/8324/changes#diff-de135d8393b0c465ab3061c7b3c2ac3e0f3acc208bf52b8169cc018ee4282adcL369). It will be better to keep consistency.

sshonTT (approved):
just nit but approved. Please be mind the copilot's reviews. They are seems reasonable.

PR #8334, opened 2026-05-08
d2m: enable aliasing for reblocked ops

anuragsingh-tt:
Instead of conservatively streaming for ops during reblocking, we can safely alias them provided the existing requirements (e.g. no nontrivial views are met). This will reduce the memory pressure.


lib/Dialect/D2M/Transforms/Allocate.cpp
```diff
@@ -1867,6 +1867,56 @@ class D2MAllocate final : public impl::D2MAllocateBase<D2MAllocate> {
     return Value();
   }
 
+  static bool isPurePermutationIndexingMap(AffineMap map) {
+    return map.getNumSymbols() == 0 &&
+           map.getNumResults() == map.getNumDims() && map.isPermutation();
+  }
+
+  bool canAliasBlockedAllParallelEltwiseOutput(
```
vwellsTT: nit: I would probably add a top-level comment here to explain what exactly this condition means, tedious to understand atm

vwellsTT (approved):

PR #8338, opened 2026-05-08
[L1SpillManagement] Reshard future consumers that reject DRAM after spill

rpavlovicTT:
## Summary

\`evictValue\` only reshards back to L1 for already-processed consumers. Future consumers with hard sharded-input constraints (e.g. \`paged_update_cache\`) silently end up with DRAM input and fail later validation in \`OperationValidationAndFallback\`.

## Change

Probe future consumers with \`additionalL1Usage=0\` (layout-only check). If they reject DRAM with \`MetalBackendError\`, insert reshard. Consumers that accept DRAM keep it.

Defense-in-depth alongside #8337.

### DIFF
--- a/lib/Dialect/TTNN/Analysis/L1SpillManagement.cpp
+++ b/lib/Dialect/TTNN/Analysis/L1SpillManagement.cpp
@@ -283,6 +283,9 @@ L1SpillManagement<MemoryTracker>::L1SpillManagement(
 template <typename MemoryTracker>
 OpConfig
 L1SpillManagement<MemoryTracker>::extractOpConfigFromIR(Operation *op) {
+  if (op->getNumResults() == 0) {
+    return OpConfig{};
+  }
   auto tensorType = mlir::cast<RankedTensorType>(op->getResult(0).getType());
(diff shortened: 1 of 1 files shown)


lib/Dialect/TTNN/Analysis/L1SpillManagement.cpp
```diff
@@ -737,11 +745,31 @@ void L1SpillManagement<MemoryTracker>::evictValue(Value victim, int64_t pos,
...
+        }
+      }
+      if (spilledOperandIdx.empty()) {
+        continue;
+      }
+
+      bool needsReshard = posIt->second < pos;
+      if (!needsReshard) {
+        auto consumerInputs = utils::extractInputLayouts(consumer);
+        auto consumerConfig = extractOpConfigFromIR(consumer);
```
bmalesevicTT, line 767: Won't this assert for ops with 0 results in IR, such as paged_update_cache? Can still be checked but with null output set here

rpavlovicTT: Good catch! Fixed it.

bmalesevicTT (approved):

PR #8343, opened 2026-05-08
Extended support for ttnn.group_norm and maxpool3d

jazpurTT:
### Ticket
unet3D model bing up

### Problem description
Compile failures + OOM during runtime due to several limitations in tt-mlir. ttnn.group_norm has no support for 5D inputs (3D conv models), and it is limited to only working on inputs that are tile aligned on it's C dim (plus other requirments). Also, our current decomp for maxpool3D support was causing OOM failures during runtime, both decomp passes, leave the last dim hardcoded to 1 and if the reshape that performs this is tilized it 32x's the memory needed.

### What's changed
- Added support to 5D inputs to ttnn.group_norm by relaxing the requirements on ttir and folding all three spatial dimensions into a single dim. We go from only supporting `[N, 1, H*W, C]` to `[N, 1, D*H*W, C]`
- Added a group norm workaround to pad C dim to be tile aligned, had to increase the number of groups proportionally to keep mathematical equivalency. Opened an issue on tt-metal to see if we can remove this workaround and handle it internally in https://github.com/tenstorrent/tt-metal/issues/43826
- Remove core grid attribute from group norm as it will be chosen during runtime since it's been enabled on tt-metal https://github.com/tenstorrent/tt-metal/pull/41515
- Changed maxpool3D decomp to move the C dim to the last dim instead of hard-coding it to 1:

Before
```
Pass 1 (H,W): Reshape [N,C,D,H,W] → [NCD, H, W, 1], MaxPool2d with [kH, kW]
Pass 2 (D): Reshape [NCD, Hout, Wout, 1] → [NC, D, HoutWout, 1], MaxPool2d with [kD, 1]
```
After
```
Pass 1 (H,W): Reshape and Permute [N,C,D,H,W] → [ND, H, W, C], MaxPool2d with [kH, kW]
Pass 2 (D): Reshape [ND, Hout, Wout, C] → [N, D, HoutWout, C], MaxPool2d with [kD, 1]
```

### Checklist

### DIFF
--- a/lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
+++ b/lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
@@ -2881,15 +2881,22 @@ private:
       RankedTensorType originalResultType = cast<RankedTensorType>(
           getTypeConverter()->convertType(srcOp.getResult(i).getType()));
 
+      // NDHWC input is already in the layout used for pooling; canonicalizing
+      // to NCDHW and permuting back would be a no-op transpose pair.
+      SmallVector<int64_t> permNDHWCToNCDHW = {0, 4, 1, 2, 3};
+      bool inputIsNDHWC = needsPermute && permToCanonical == permNDHWCToNCDHW;
+
       // Permute to canonical NCDHW layout if needed.
       SmallVector<int64_t> canonShape;
       if (needsPermute) {
         canonShape = ttmlir::utils::applyPermutation(inputType.getShape(),
                                                      permToCanonical);
-        input = rewriter.create<ttir::PermuteOp>(
-            srcOp.getLoc(),
-            RankedTensorType::get(canonShape, elemType, encoding), input,
-            permToCanonical);
+        if (!inputIsNDHWC) {
+          input = rewriter.create<ttir::PermuteOp>(
+              srcOp.getLoc(),
+              RankedTensorType::get(canonShape, elemType, encoding), input,
+              permToCanonical);
+        }
       } else {
         canonShape = SmallVector<int64_t>(inputType.getShape());
       }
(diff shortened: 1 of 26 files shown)


djstefanovicTT (approved):

sdjukicTT (approved):

lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
```diff
@@ -2912,17 +2913,23 @@ class StableHLOToTTIRReduceWindowOpConversionPattern
...
-      int64_t batchHW = N * C * D;
-      SmallVector<int64_t> shapeForHW = {batchHW, H, W, 1};
+      // Pass 1: Pool over H, W with channels preserved (NHWC).
+      // [N, C, D, H, W] -> permute -> [N, D, H, W, C] -> reshape ->
+      // [N*D, H, W, C].
+      SmallVector<int64_t> permNCDHWToNDHWC = {0, 2, 3, 4, 1};
+      SmallVector<int64_t> shapeNDHWC = {N, D, H, W, C};
+      Value ndhwc = rewriter.create<ttir::PermuteOp>(
+          srcOp.getLoc(), RankedTensorType::get(shapeNDHWC, elemType, encoding),
+          input, permNCDHWToNDHWC);
```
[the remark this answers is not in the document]
anuragsingh-tt: Seems like a good suggestion?

runtime/lib/ttnn/operations/normalization/group_norm.cpp
```diff
@@ -36,21 +36,24 @@ void run(const ::tt::target::ttnn::GroupNormOp *op, ProgramContext &context) {
       ::tt::runtime::ttnn::utils::createMemoryConfigIfNeeded(
           op->memory_config());
 
-  // Get core_grid: use compile-time value from flatbuffer if present,
-  // otherwise fallback to device grid size (required by ttnn::group_norm)
+  // Use core_grid from flatbuffer when the compiler specified one; otherwise
+  // pass nullopt so tt-metal can choose a valid grid (see tt-metal#40916).
+  // Defaulting to the full device grid was invalid for large Ht*W (e.g. 8x8
+  // when only 8x6 fits).
```
jameszianxuTT: Under what circumstances would compiler specify a core grid instead of using the new auto selection? It seems from the metal issue discussion that we should not try to solve for this ourselves because we have to model unstable group norm internal implementation constraints from ttnn, inside our compiler. 

If it is actually found to be unsafe to do so, I think it might be good to defensively warn or even throw here, with the eventual goal of always passing nullopt for the core grid. From metal side, it seems like this shouldn't even be a part of the API surface since it's a pure footgun, but that's a separate issue

jameszianxuTT (approved):
Approving runtime changes with nit

djstefanovicTT:
If the metal can appropriately handle nullptr for core grid, it would make sense to remove core grid attribute from `ttnn.group_norm` along with complex core grid calculations.

sgholamiTT (approved):
OpModel changes LGTM.

runtime/lib/ttnn/operations/normalization/group_norm.cpp
jazpurTT: I ended up removing the core grid attribute all together like @djstefanovicTT suggested

PR #8354, opened 2026-05-11
[golden] add chisel mappings for TTNN op goldens

ndrakulicTT:
## Summary
- Register ~140 TTNN ops in a new `CHISEL_GOLDEN_MAPPINGS` table with a chisel-shaped interface: `fn(op, inputs) -> GoldenMapTensor | tuple`.
- Add `chisel/executor.py` to re-key SSA inputs by operand role, run the registered golden, and normalize results to `(SSA outputs..., mutated in-place operands...)`.
- Classify TTNN ops in `chisel/ops.py` as non-executable (device / I/O / control flow) or in-place (operands written via `MemWrite`); add `GoldenNotImplementedError` and switch `IRModule` to a module-wide `AsmState`.
- Add `test_golden_execution.py`: run every TTNN op in a flatbuffer on meta tensors and assert output shapes/dtypes match the MLIR types.
- Drive-by golden fixes: `argmax` `None` dim, `batch_norm_training` running-stat update, `sdpa` causal-mask dtype/device.

## Follow-ups
- Unify `CHISEL_GOLDEN_MAPPINGS` with the existing `GOLDEN_MAPPINGS` into a single registry once the interfaces converge.
- Add goldens for the remaining ops currently listed in `_KNOWN_NOT_IMPLEMENTED_OPS`, grouped into:
  - **Constant ops**: `ttnn.constant` — needs custom handling since it materializes a value rather than transforming inputs.
  - **Quantization ops**: `ttnn.quantize`, `ttnn.dequantize`, `ttnn.requantize` — blocked on chisel-side support for quantized tensor element types.
  - **LLM-specific ops**: `ttnn.nlp_create_qkv_heads_decode`, `ttnn.nlp_concat_heads`, `ttnn.nlp_concat_heads_decode`, `ttnn.rotary_embedding_llama`.
  - **Multichip ops**: `ttnn.mesh_shard`, `ttnn.mesh_partition` (and other CCL/distribution ops as we extend coverage).

### DIFF
--- a/test/python/chisel/test_golden_execution.py
+++ b/test/python/chisel/test_golden_execution.py
@@ -40,7 +40,7 @@ _KNOWN_NOT_IMPLEMENTED_OPS: frozenset[str] = (
             "ttnn.nlp_concat_heads_decode",
             "ttnn.bitcast_convert",
             "ttnn.group_norm",
-            "ttnn.batch_norm_training"
+            "ttnn.batch_norm_training",
         }
     )
     | QUANTIZE_OP_NAMES
--- a/test/python/chisel/test_golden_execution.py
+++ b/test/python/chisel/test_golden_execution.py
@@ -40,6 +40,7 @@ _KNOWN_NOT_IMPLEMENTED_OPS: frozenset[str] = (
             "ttnn.nlp_concat_heads_decode",
             "ttnn.bitcast_convert",
             "ttnn.group_norm",
+            "ttnn.batch_norm_training"
         }
     )
     | QUANTIZE_OP_NAMES
--- a/test/python/chisel/test_golden_execution.py
+++ b/test/python/chisel/test_golden_execution.py
@@ -9,7 +9,6 @@ Ops without a registered golden are skipped, not failed.
 import pytest
 import torch
 
-from chisel.exceptions import GoldenNotImplementedError
 from chisel.executor import (
     build_role_keyed_inputs,
     execute_golden,
@@ -41,6 +40,7 @@ _KNOWN_NOT_IMPLEMENTED_OPS: frozenset[str] = (
             "ttnn.nlp_concat_heads",
             "ttnn.nlp_concat_heads_decode",
             "ttnn.bitcast_convert",
+            "ttnn.group_norm",
         }
     )
     | QUANTIZE_OP_NAMES
@@ -54,6 +54,15 @@ def _has_quantized_type(op) -> bool:
         if quant.QuantizedType.isinstance(value.type.element_type):
             return True
     return False
+
+
+@pytest.fixture(autouse=True)
+def _default_meta_device():
+    # Goldens may internally allocate tensors (e.g. SDPA causal mask,
+    # ttnn.arange/full/ones/zeros). Running them on the meta device keeps
+    # the test allocation-free and avoids per-golden device plumbing.
+    with torch.device("meta"):
+        yield
 
 
 def test_golden_execution(subtests, ir_module, binary, binary_path):
@@ -85,12 +94,7 @@ def test_golden_execution(subtests, ir_module, binary, binary_path):
                     inputs[name] = GoldenMapTensor({0: tensor}, (1, 1))
 
                 golden_inputs = build_role_keyed_inputs(op.opview, inputs, asm_state)
-                try:
-                    result = execute_golden(op.opview, golden_inputs)
-                except GoldenNotImplementedError as e:
-                    if e.op_name in _KNOWN_NOT_IMPLEMENTED_OPS:
-                        pytest.skip(str(e))
-                    raise
+                result = execute_golden(op.opview, golden_inputs)
 
                 op_outputs = get_op_outputs(op)
                 for i, op_out in enumerate(op_outputs):
--- a/test/python/chisel/test_golden_execution.py
+++ b/test/python/chisel/test_golden_execution.py
@@ -1,0 +1,121 @@
+# SPDX-FileCopyrightText: (c) 2026 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+"""
+Execute every TTNN op in a flatbuffer through the golden system on meta
+tensors (shape/dtype only) and verify each output matches the MLIR op type.
+Ops without a registered golden are skipped, not failed.
+"""
+import pytest
+import torch
+
+from chisel.exceptions import GoldenNotImplementedError
+from chisel.executor import (
+    build_role_keyed_inputs,
+    execute_golden,
+    get_provided_inplace_vals,
+)
+from chisel.ops import (
+    get_inplace_operands,
+    get_op_inputs,
+    get_op_outputs,
+    is_non_executable_op,
+    is_tensor_value,
+)
+from golden import GoldenMapTensor, get_chisel_golden_function
+from golden.mapping import mlir_type_to_torch_dtype
+from ttmlir.dialects import quant
+from utils import iterate_programs, QUANTIZE_OP_NAMES
+
+# Ops without a registered golden - skipped instead of failed. Remove an entry
+# once its golden lands. Quantization ops are included via QUANTIZE_OP_NAMES.
+_KNOWN_NOT_IMPLEMENTED_OPS: frozenset[str] = (
+    frozenset(
+        {
+            "ttnn.generic",
+            "ttnn.nlp_create_qkv_heads_decode",
+            "ttnn.rotary_embedding_llama",
+            "ttnn.constant",
+            "ttnn.mesh_shard",
+            "ttnn.mesh_partition",
+            "ttnn.nlp_concat_heads",
+            "ttnn.nlp_concat_heads_decode",
+            "ttnn.bitcast_convert",
+        }
+    )
+    | QUANTIZE_OP_NAMES
+)
+
+
+def _has_quantized_type(op) -> bool:
+    for value in list(op.operands) + list(op.results):
+        if not is_tensor_value(value):
+            continue
+        if quant.QuantizedType.isinstance(value.type.element_type):
+            return True
+    return False
+
+
+def test_golden_execution(subtests, ir_module, binary, binary_path):
+    asm_state = ir_module.get_asm_state()
+
+    for _prog_idx, prog_name in iterate_programs(binary):
+        for op in ir_module.get_function_ops(prog_name):
+            with subtests.test(prog=prog_name, op=op.name):
+                if is_non_executable_op(type(op.opview)):
+                    pytest.skip(f"golden not applicable for {type(op.opview).__name__}")
+
+                if _has_quantized_type(op):
+                    pytest.skip(
+                        f"quantized tensor types not yet supported (op: {op.name})"
+                    )
+
+                if (
+                    get_chisel_golden_function(type(op.opview)) is None
+                    and op.name in _KNOWN_NOT_IMPLEMENTED_OPS
+                ):
+                    pytest.skip(f"{op.name}: no golden registered")
+
+                inputs = {}
+                for operand in get_op_inputs(op):
+                    name = operand.get_name(asm_state)
+                    shape = list(operand.type.shape)
+                    dtype = mlir_type_to_torch_dtype(operand.type.element_type)
+                    tensor = torch.empty(shape, dtype=dtype, device="meta")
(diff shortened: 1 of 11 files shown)


tools/chisel/chisel/exceptions.py
```diff
@@ -0,0 +1,15 @@
...
+class GoldenNotImplementedError(NotImplementedError):
+    """Raised by execute_golden when no golden is registered for an op."""
+
+    def __init__(self, op):
+        op_name = op.name
+        op_type_name = type(op).__name__
+        super().__init__(f"{op_name}: no golden registered for {op_type_name}")
+        self.op = op
+        self.op_name = op_name
+        self.op_type_name = op_type_name
```
mmilosevicTT: Do we really need to store `op_name` and `op_type_name` when we have them in the `op`?

I would keep error class as simple as it is. Adding a variable `op_name` to replace `op.name` does not seem necessary to me (as in it is same number of characters).

tools/chisel/chisel/executor.py
```diff
@@ -0,0 +1,117 @@
+# SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
```
mmilosevicTT: nit:
```suggestion
# SPDX-FileCopyrightText: (c) 2026 Tenstorrent AI ULC
```

test/python/chisel/test_walk_program.py
```diff
@@ -5,23 +5,19 @@
...
 
-    for prog_idx in range(binary.get_num_programs()):
-        prog_name = binary.get_program_name(prog_idx)
+    for prog_idx, prog_name in iterate_programs(binary):
         mlir_ops = iter(ir_module.get_function_ops(prog_name))
 
         def _check(op_ctx):
             mlir_op = next(mlir_ops)
 
-            if mlir_op.name in _SKIP_OPS:
```
mmilosevicTT, line 24: I do not like this approach here. Can we keep this more open for future? We might need to skip additional ops in the tests for some quick fixes. Can we just do SKIP_OPS as a separate map to keep the code more consistent?

tools/chisel/chisel/executor.py
```diff
@@ -0,0 +1,117 @@
...
+    assert len(tensors) == expected, (
+        f"Golden for {type(op).__name__} returned {len(tensors)} tensor(s) "
+        f"but op has {ssa_count} SSA output(s) + {len(inplace_vals)} mutated "
+        f"operand(s); fix CHISEL_GOLDEN_MAPPINGS to return SSA outputs first, "
+        f"then one tensor per provided in-place operand"
+    )
+    return tensors
+
+
+def get_provided_inplace_vals(
```
mmilosevicTT: Should this be before `execute_golden`? From readability I expect this to be helper function before the main part, i.e. `execute_golden`

mmilosevicTT: Should this be prefixed with `_` as it is a private method?

tools/chisel/chisel/ops.py
```diff
@@ -18,22 +19,52 @@
...
+    ttnn.UpdateCacheOp: ("cache",),
+    ttnn.PagedUpdateCacheOp: ("cache",),
+    ttnn.FillCacheOp: ("cache",),
+    ttnn.PagedFillCacheOp: ("cache",),
+    ttnn.WriteTensorOp: ("device_tensor",),
+    ttnn.DumpTensorOp: (),
+    ttnn.DeallocateOp: (),
+    ttnn.BatchNormTrainingOp: ("running_mean", "running_var"),
+    ttnn.PointToPointOp: ("optional_output_tensor",),
+}
```
mmilosevicTT, line 46: Should these be prefixed with `_` as you have getter functions that you use to access elements? I would expect this not to go outside of this file then.

tools/chisel/chisel/ops.py
```diff
@@ -18,22 +19,52 @@
...
+    ttnn.GetDeviceOp,
+    ttnn.LoadTensorOp,
+    ttcore.LoadCachedOp,
+    func.CallOp,
+}
+
+# Op class -> operand role names mutated via `Arg<..., [MemWrite]>` in ODS.
+# Goldens return SSA results first, then one tensor per *provided* memwrite
+# operand (absent Optional operands are skipped - see get_provided_inplace_vals).
+# Hand-maintained; TODO: derive from ODS (see python_op_schema_codegen.py).
```
mmilosevicTT: Do we have a PR/ticket for this?

test/python/chisel/test_golden_execution.py
```diff
@@ -0,0 +1,121 @@
...
+                    shape = list(operand.type.shape)
+                    dtype = mlir_type_to_torch_dtype(operand.type.element_type)
+                    tensor = torch.empty(shape, dtype=dtype, device="meta")
+                    inputs[name] = GoldenMapTensor({0: tensor}, (1, 1))
+
+                golden_inputs = build_role_keyed_inputs(op.opview, inputs, asm_state)
+                try:
+                    result = execute_golden(op.opview, golden_inputs)
+                except GoldenNotImplementedError as e:
+                    if e.op_name in _KNOWN_NOT_IMPLEMENTED_OPS:
```
mmilosevicTT: This check is redundant, you have in lines 73-76 the skip for this, we do not need to repeat.

test/python/chisel/test_golden_execution.py
```diff
@@ -0,0 +1,121 @@
...
+                    assert list(result[i].shape) == expected_shape, (
+                        f"output[{i}] shape mismatch: got {list(result[i].shape)}, expected {expected_shape}"
+                        f"\nbinary: {binary_path}"
+                    )
+                    assert result[i].dtype == expected_dtype, (
+                        f"output[{i}] dtype mismatch: got {result[i].dtype}, expected {expected_dtype}"
+                        f"\nbinary: {binary_path}"
+                    )
+
+                inplace_vals = get_provided_inplace_vals(
```
mmilosevicTT: Uff, I do not like this. We are testing results of `execute_golden` with function that we used to implement it. This seems a bit circular.

tools/golden/mapping.py
```diff
@@ -1482,11 +1494,11 @@ def ttir_all_to_all_dispatch_metadata_golden(
     grouped_indices = expert_indices.group_by_axis(cluster_axis)
     grouped_scores = expert_scores.group_by_axis(cluster_axis)
 
-    # expert_mapping is replicated — get from any device
+    # expert_mapping is replicated - get from any device
     mapping_tensor = list(expert_mapping._shard_map.values())[0]
-    # Shape is [1, 1, D, E] — squeeze to [D, E], use row 0 since all rows identical
+    # Shape is [1, 1, D, E] - squeeze to [D, E], use row 0 since all rows identical
     mapping_2d = mapping_tensor.reshape(-1, mapping_tensor.shape[-1])  # [D, E]
-    mapping_row = mapping_2d[0]  # [E] — mapping_row[e] = device_id owning expert e
```
mmilosevicTT, line 1489: nit: Is this really a change that we need from claude garbage?

mmilosevicTT: ditto for all other places in the file

tools/golden/mapping.py
```diff
@@ -7915,3 +7937,1241 @@ def get_golden_function(ttir_op_class: type, **kwargs) -> Optional[Callable]:
...
+    return input_tensor.clone().to(mlir_type_to_torch_dtype(output_type_mlir))
+
+
+def ttnn_to_memory_config_golden(
+    input_tensor: GoldenMapTensor, output_type_mlir: Type
+) -> GoldenMapTensor:
+    return input_tensor.clone().to(mlir_type_to_torch_dtype(output_type_mlir))
+
+
+# Chisel golden interface: fn(op, inputs: Dict[str, GoldenMapTensor]) -> GoldenMapTensor | tuple
```
mmilosevicTT, line 8012: Could we leave a comment for why we need these for now?

tools/golden/mapping.py
```diff
@@ -7915,3 +7937,1241 @@ def get_golden_function(ttir_op_class: type, **kwargs) -> Optional[Callable]:
...
     assert False, f"No golden function found for TTIR operation: {ttir_op_class}"
+
+
+def ttnn_assign_golden(
+    input_tensor: GoldenMapTensor, output_type_mlir: Type
+) -> GoldenMapTensor:
+    return input_tensor.clone().to(mlir_type_to_torch_dtype(output_type_mlir))
+
+
+def ttnn_to_memory_config_golden(
```
mmilosevicTT: Should these be prefixed with `_` as well?

tools/golden/mapping.py
```diff
@@ -7915,3 +7937,1241 @@ def get_golden_function(ttir_op_class: type, **kwargs) -> Optional[Callable]:
...
+        weight_tensor=inputs["weight"],
+        bias_tensor=inputs["bias"],
+        k_attr=op.attributes["k"],
+        _num_experts_attr=op.attributes["num_experts"],
+        output_type_mlir=op.results[0].type.element_type,
+    )
+
+
+def chisel_ttnn_moe_expert_token_remap(op, inputs):
+    return ttir_moe_expert_token_remap_golden(
```
mmilosevicTT, line 8895: I am not a fan of `ttnn` function calling something that has `ttir` in prefix. Could we just remove `ttir` prefix then?

tools/golden/mapping.py
mmilosevicTT: This is not accurate description for this mapping file now.

mmilosevicTT (commented):
Great Work! 🎊 

I left some comments for this. In general seems logical.

test/python/chisel/test_walk_program.py
ndrakulicTT: This will be fixed in next PR, I will unify all op skips into one op_config place.

tools/chisel/chisel/ops.py
ndrakulicTT: I have created one: #8385

test/python/chisel/test_golden_execution.py
ndrakulicTT: removed

test/python/chisel/test_golden_execution.py
ndrakulicTT: The test is to test if the returned values matches the shapes/dtypes that MLIR expected for inplace operands. So we need to use `get_provided_inplace_vals` to get what operands are inplace

tools/golden/mapping.py
ndrakulicTT: Removed from this PR

tools/golden/mapping.py
ndrakulicTT: removed this helpers as used in only one place and are one liners

tools/golden/mapping.py
ndrakulicTT: Leaving it for now will be consolidated #8399, as there are quite a few functions that I used from `ttir`

tools/golden/mapping.py
ndrakulicTT: Fixed

test/python/chisel/test_golden_execution.py
```diff
@@ -0,0 +1,122 @@
...
+                    assert list(result[i].shape) == expected_shape, (
+                        f"output[{i}] shape mismatch: got {list(result[i].shape)}, expected {expected_shape}"
+                        f"\nbinary: {binary_path}"
+                    )
+                    assert result[i].dtype == expected_dtype, (
+                        f"output[{i}] dtype mismatch: got {result[i].dtype}, expected {expected_dtype}"
+                        f"\nbinary: {binary_path}"
+                    )
+
+                inplace_vals = get_flat_inplace_vals(op.opview)
```
mmilosevicTT, line 112: nit: Do we need `.opview` here?

mmilosevicTT (approved):
LGTM

test/python/chisel/test_golden_execution.py
ndrakulicTT: Yeah, op is just basic Op object, where OpView has all the op specific operand access by name and similar stuff

tools/golden/mapping.py
```diff
@@ -19,6 +28,7 @@
 import einops
 import torch
 import torch.nn.functional
+import torch.nn.functional as F
```
jgrimTT: Duplicate

jgrimTT (approved):
Approving golden mapping changes!

PR #8361, opened 2026-05-11
[D2M] Make remote_load/remote_store truly DPS after bufferization

nsmithtt:
The remote_load and remote_store ops advertised DestinationStyleOpInterface but still produced an SSA result in the post-bufferization memref form, contradicting DPS conventions and leaving a dead value in the IR.

Both ops now have no result once bufferized; the localBuffer (load) or destination memref (store) is the result. The tensor form retains a result since downstream tensor consumers and d2m.yield need it.

- bufferize() for both ops emits the no-result memref form and wraps the bufferized destination in a bufferization.to_tensor so pending tensor users see a valid SSA value until they bufferize themselves.
- Verifier now requires: implicit tensor form has a result; implicit memref form has no result.
- getValue() (AffineReadOpInterface) returns localBuffer when no result is present.
- Allocate.cpp drops setType() on the result and short-circuits the fallback re-typing when the localBuffer is already typed, preserving #ttcore.cb_layout<...> annotations set by stream insertion.
- LowerMulticastLoads uses the no-result builder for replacement ops.
- repairParallelizedRegionTypes guards on hasResultForm() before propagating the localBuffer type to the (possibly absent) result.
- Lit tests and one C++ unit test updated to drop the now-illegal result binding and arrow on memref-form remote_load / remote_store.

### DIFF
--- a/include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
+++ b/include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
@@ -1753,12 +1753,22 @@ def D2M_RemoteLoadOp : D2M_GenericRegionDatamovementOp<"remote_load",
       operation has multiple modes; the explicit CB form and unicast/multicast
       modes are both orthogonal and can be mixed and matched.
 
-      The basic form takes a local buffer as input and loads data into it:
+      The basic form takes a local buffer as input and loads data into it.
+      In tensor form (pre-bufferization), the op produces a result that aliases
+      the local buffer:
+
+      ```mlir
+      %result = d2m.remote_load %buffer %generic_operand[%i, %j]
+        : tensor<SY x SX x f32, ...>, tensor<GY x GX x SY x SX x f32, ...>
+        -> tensor<SY x SX x f32, ...>
+      ```
+
+      In memref form (post-bufferization), the op has no result; data is loaded
+      in-place into the local buffer:
 
       ```mlir
       d2m.remote_load %buffer %generic_operand[%i, %j]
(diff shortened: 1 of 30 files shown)


include/ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.td
anuragsingh-tt: Does this need to be updated?

anuragsingh-tt, line 1726: Does this need to be updated?

nsmithtt: I don't think so, unless I'm missing something.

nsmithtt: I don't think so, unless I'm missing something.

anuragsingh-tt: “In implicit form, the result matches the shard memref/tensor type.” 
There is no result now post-bufferization. This description is already getting pretty big though (perhaps it should be a doc?) so maybe specifying the tensor/memref distinction is not necessary.

anuragsingh-tt: Sorry, saw this comment in the implementation and got confused:
  // Load operations don't have results anymore (they load into CB)

anuragsingh-tt (approved):

sohaibnadeemTT (approved):
DMA changes lgtm

vtangTT (approved):
lgtm ty

dlokeTT (approved):

PR #8370, opened 2026-05-11
[D2M] Use offset-aware grid range handling in GridSelection

### LINKED ISSUE #8368: [D2M][GridSelection] Spatial grid selection drops VGM in d2m.generic grid attribute
## Problem

`d2m.spatial` can specify execution regions with `grid_ranges`, where each range includes both shape and offset.
In current `GridSelection`, offset semantics can be dropped while materializing selected grid information.

## Symptom

For non-origin spatial regions, VGM attached to `d2m.empty` (the defining op of `d2m.generic` output operand) disappears after `grid-selection`.
As a result, VGM also disappears from the `d2m.generic` grid attribute.

## Impact

This can make selected-grid/VGM mapping inconsistent with the intended spatial region and may lead to incorrect downstream grid-local behavior.

## Expected

`GridSelection` should preserve full grid-range semantics (shape + offset) and keep VGM consistent before/after transformation.

hkwonTT:
### Ticket
closes #8368 

### Problem description
`d2m.spatial` `grid_ranges` carry both shape and offset semantics.
In the previous `GridSelection` flow, selected grids were materialized in a way that did not fully preserve offset semantics, which could cause virtual grid mapping to be dropped or become inconsistent in non-origin spatial regions.

In particular, when a d2m.generic output operand is defined by d2m.empty, the virtual-to-physical mapping could lose offset consistency after grid selection, leading to mismatches in the resulting d2m.generic grid attribute.

### What's changed
 - Updated grid selection analysis/application to treat selected regions as full grid ranges (shape + offset) rather than shape-only.
 - Added offset-aware handling when updating virtual mappings so non-origin spatial semantics are preserved.
 - Introduced shared affine-map utilities for applying offsets consistently.
 - Added a regression test for the non-origin spatial path where d2m.generic uses a direct d2m.empty output operand, and verified VGM is preserved in the resulting d2m.generic grid attribute.

### Checklist

### DIFF
--- a/lib/Dialect/D2M/Transforms/GridSelection.cpp
+++ b/lib/Dialect/D2M/Transforms/GridSelection.cpp
@@ -37,8 +37,9 @@ namespace mlir::tt::d2m {
 
 static std::pair<mlir::AffineMapAttr, mlir::AffineMapAttr>
 deriveVirtualGridAttrs(Operation *anchorOp, ArrayRef<int64_t> selectedGrid,
-                       ArrayRef<int64_t> effectiveTargetGrid,
+                       const EffectiveTargetGridRange &effectiveTargetGridRange,
                        OpBuilder &builder) {
+  ArrayRef<int64_t> effectiveTargetGridShape = effectiveTargetGridRange.shape;
   auto device = ttcore::lookupDevice(anchorOp);
   auto workerGridShape = device.getWorkerGrid().getShape();
   bool isVirtual = ttmlir::d2m::utils::grids::requiresVirtualGrid(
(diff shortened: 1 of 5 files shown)


lib/Dialect/D2M/Transforms/GridSelection.cpp
```diff
@@ -36,37 +35,65 @@ namespace mlir::tt::d2m {
 // ----------------------------------------------------------------------------
 
 static std::pair<mlir::AffineMapAttr, mlir::AffineMapAttr>
-deriveVirtualGridAttrs(Operation *anchorOp, ArrayRef<int64_t> selectedGrid,
-                       ArrayRef<int64_t> effectiveTargetGrid,
+deriveVirtualGridAttrs(ArrayRef<int64_t> selectedGrid,
```
vwellsTT, line 38: nit: personally don't like reference-as-return style like this, I wonder if this could be kept more similar and the offset could be passed? idk
edit: wrong place, I meant this func `void GridAnalysis::analyzeGenericOp(GenericOp genericOp,`

vwellsTT (approved):
from my claude review rules:
🔴 Must fix
(none blocking)
🟡 Should fix
Add a lit test for requiresVirtualization=true && hasOffset=true (combined path in deriveVirtualGridAttrs).
Confirm or test the "1x1 selected grid, direct ToLayout input, non-origin spatial range" case — either add a test or a brief comment in optimizeToLayoutGrid explaining why the early return is safe there.
applyCompositeViewUpdate: take effectiveTargetGridRange as const EffectiveTargetGridRange & and rename to match siblings.
🟢 Nice to have
Drop the trailing return; in analyzeGenericOp.
Update D2MGridSelection's description to mention offset-aware VGM derivation for d2m.spatial ranges.
Consider returning the analysis result directly from analyzeGenericOp rather than mutating a partially-populated struct.

Some of this is redundant with my feedback, others might be worth taking a look at (won't block or anything though)

lib/Dialect/D2M/Transforms/GridSelection.cpp
hkwonTT: Gotcha. Updated as you suggested. And also updated most of the things claude review pointed out. Thanks!

PR #8371, opened 2026-05-11
[D2MBuilder] Add multi-device test scaffolding (mesh shard, global semaphore, fabric config)

hkwonTT:
### Ticket
Link to Github Issue

### Problem description
Authoring multi-device tests with D2MBuilder required several missing building blocks across the Python builder and bindings.
In practice, this meant test code had to rely on workarounds for mesh handling, generic op argument plumbing, semaphore creation, and system descriptor loading.
Because of these gaps, it was difficult to express realistic multi-device execution flows in a clean and consistent way at the builder level.

### What's changed
 - Added mesh_shard(...) to D2MBuilder:
   - Computes output shape based on shard direction and factors
   - Sets mesh encoding for shard outputs when needed
   - Emits d2m.MeshShardOp with shard type/direction/shape/dims attributes
 - Extended D2MBuilder.generic(...) plumbing:
   - Passes additional_args through _create_generic(...)
   - Adds and propagates fabric_connection_config into generic op creation
 - Added create_global_semaphore(...) to D2MBuilder:
   - Creates d2m.CreateGlobalSemaphoreOp directly from the builder API
 - Extended Python bindings:
   - Exposed d2m.ir.GlobalSemaphoreType.get(...)
 - Updated builder module assembly logic:
   - Ensured D2MBuilder modules also receive mesh module attributes, consistent with TTIR/TTNN builders

### Checklist

### DIFF
--- a/tools/builder/d2m/d2m_builder.py
+++ b/tools/builder/d2m/d2m_builder.py
@@ -499,3 +499,30 @@ class D2MBuilder(Builder):
             mcast_dims=mcast_dims,
             local_buffer=dst,
         )
+
+    def create_global_semaphore(
+        self,
+        output: Optional[Operand] = None,
+        output_type: Optional[RankedTensorType] = None,
+        value: int = 0,
+        unit_attrs: Optional[List[str]] = None,
+        loc: Optional[Union[str, Location]] = None,
+    ) -> OpView:
+        """Create a D2M create_global_semaphore operation."""
+        resolved_output_type, output_create_fn = self._resolve_output_spec(
+            output=output, output_type=output_type
+        )
+        return self._op_proxy(
+            d2m.CreateGlobalSemaphoreOp,
+            [],
+            unit_attrs=unit_attrs,
+            organize_d2m_args=lambda _inputs, output, _output_shape: (output,),
+            output_type=resolved_output_type,
+            output_shape=resolved_output_type.shape,
+            output_create_fn=output_create_fn,
+            d2m_kwargs={
+                "value": value,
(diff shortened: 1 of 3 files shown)


tools/builder/d2m/d2m_builder.py
```diff
@@ -165,6 +166,77 @@ def empty(
...
+                    raise ValueError(f"shard dim out of range: dim={dim}, rank={rank}")
+                if shard_direction == MeshShardDirection.FullToShard:
+                    if output_shape[dim] % factor != 0:
+                        raise ValueError(
+                            f"non-divisible shard: dim_size={output_shape[dim]}, "
+                            f"factor={factor}, dim={dim}"
+                        )
+                    output_shape[dim] //= factor
+                else:
+                    output_shape[dim] *= factor
```
jgrimTT, line 206: We usually skip adding ValueErrors like these in builder because they are caught and clearly communicated in `lib/Dialect/D2M/IR/D2MOps.cpp`

jgrimTT (approved):
Added a comment for a recommended change, otherwise LGTM!

tools/builder/d2m/d2m_builder.py
hkwonTT: Thanks for the comment. We need to know the output shape to create a `d2m.MeshShardOp`, and we should calculate it from the attributes. While calculating the output shape, we need to make sure all the attributes are correct. So I put `ValueErrors` here. FYI, `ttir_builder` uses the golden function to get the output shape, while the `d2m_builder` does not have one.

tapspatel (approved):

PR #8375, opened 2026-05-11
[d2m] Refactor Allocator (and other frontend passes) using SynchronizableOpInterface

sohaibnadeemTT:
### Ticket
Link to Github Issue: https://github.com/tenstorrent/tt-mlir/issues/8455

### Problem description/What's changed
This PR refactors the Allocator to use the SynchronizableOpInterface to detect which buffers need to be double-buffered (for synchronization between threads). We have a separate pass (MarkSynchronizedBuffers) to do this which tags the memref.alloc’s and the Allocator now only has to look for the d2m.synchronized_buffer attribute to identify these allocs. This contrasts with the previous behaviour of allocating a CB per operand which becomes problematic when we need intermediate CBs (e.g. if we have a remote_load of an operand followed by a local_copy, we need a CB for the local_copy output too but this is not related to an operand) or have multiple remote_loads of an operand.

Differentiating between different types of in-generic allocs is now done via an attribute on the memref.alloc itself: d2m.synchronized_buffer for buffers that need synchronization and d2m.scratch_buffer for scratch buffers. We also have a d2m.compute_intermediate: this is meant for memrefs that are the output of one linalg and the input of another. Prior to this refactor, we didn't actually allocate these and that happens to work out since we always fuse linalgs together into a single synchronized compute region and don’t generate the outputs of intermediary linalg.generic’s, so this attribute is meant to maintain parity with that behaviour. If we remove this, it may give OOM for cases where we may actually have enough memory after fusing and removing these intermediates but the Allocator pass is positioned earlier than the fusion pass (that fuses in-generic compute loops) so it isn’t aware of this. 

We also no longer use the CBLayoutAttr attribute to identify which allocs to register, we should be allocating allocs according to the attributes on them mentioned above (d2m.synchronized_buffer, d2m.scratch_buffer, d2m.compute_intermediate). Therefore, the CBLayoutAttr has been been moved down to the HoistCBAllocs pass and now mainly serves as the way to tell the backend after hoisting that this is an in-generic alloc (as opposed to an normal tensor) which should be backed with a CB (regardless of whether or not it actually needs synchronization, since this is the only mechanism we have for creating core local l1 memory that only lives for the duration of an op). 

Aliasing logic has also be cleaned up.

### Checklist


lib/Dialect/D2M/Transforms/Allocate.cpp
```diff
@@ -615,82 +631,50 @@ class D2MAllocate final : public impl::D2MAllocateBase<D2MAllocate> {
...
-                    L1memInfo.alignment);
-          }
+      std::optional<int32_t> numBuffers = std::nullopt;
+      genericOp->walk([&](memref::AllocOp allocOp) {
+        if (allocOp->getAttr("d2m.compute_intermediate")) {
+          // skip allocating, this is a contrace from fusion and compute
+          // lowering saying we will not actually use this buffer, it's just a
+          // placeholder, so we can safely skip it
+          return WalkResult::advance();
+        } else if (allocOp->getAttr("d2m.scratch_buffer")) {
```
vmilosevic: :warning: **llvm\-else\-after\-return** :warning:
```suggestion
        } if (allocOp->getAttr("d2m.scratch_buffer")) {
```

lib/Dialect/D2M/Transforms/Allocate.cpp
```diff
@@ -615,82 +631,50 @@ class D2MAllocate final : public impl::D2MAllocateBase<D2MAllocate> {
...
+        } else if (allocOp->getAttr("d2m.synchronized_buffer")) {
+          numBuffers =
+              allocOp->getAttrOfType<IntegerAttr>("d2m.synchronized_buffer")
+                  .getInt();
+        } else {
+          assert(false && "unexpected alloc op attribute");
+          // we should allow this but asserting for now to check it's not used
         }
-      }
+        // TODO: check not of more than one type (or maybe add a
```
vmilosevic: :warning: **google\-readability\-todo** :warning:
missing username/bug in TODO
```suggestion
        // TODO(nsmith): check not of more than one type (or maybe add a
```

lib/Dialect/D2M/Transforms/Allocate.cpp
```diff
@@ -1331,230 +1297,83 @@ class D2MAllocate final : public impl::D2MAllocateBase<D2MAllocate> {
...
-                                 rewriter.getI64IntegerAttr(operandIndex));
-              }
-            }
+  LogicalResult convertAliasedLoadStore(func::FuncOp funcOp,
+                                        FuncAnalysisData &analysis) {
+    IRRewriter rewriter(funcOp->getContext());
+    for (const auto &[genericOp, genericCtx] : analysis.generics) {
+      const auto &genericOpRef = genericOp;
+      for (const OperandContext &operandCtx : genericCtx.operands) {
+        // TODO: clean up
```
vmilosevic: :warning: **google\-readability\-todo** :warning:
missing username/bug in TODO
```suggestion
        // TODO(nsmith): clean up
```

lib/Dialect/D2M/Transforms/Allocate.cpp
```diff
@@ -1331,230 +1297,83 @@ class D2MAllocate final : public impl::D2MAllocateBase<D2MAllocate> {
...
+
+        genericOpRef->walk([&](RemoteLoadOp remoteLoadOp) {
+          // llvm::errs() << "remote load op: " << remoteLoadOp.getMemref()
+          //              << "\n";
+          if (remoteLoadOp.getMemref() == operandCtx.operand->get() &&
+              isAliasedLoad(remoteLoadOp)) {
+            // llvm::errs() << "replacing remote load op with aliased load
+            // op\n";
+            //  replace alloc with operand alias op
+            auto allocOp = remoteLoadOp.getLocalBuffer().getDefiningOp();
```
vmilosevic: :warning: **llvm\-qualified\-auto** :warning:
`` auto allocOp `` can be declared as `` auto *allocOp ``
```suggestion
            auto *allocOp = remoteLoadOp.getLocalBuffer().getDefiningOp();
```

lib/Dialect/D2M/Transforms/Allocate.cpp
```diff
@@ -1331,230 +1297,83 @@ class D2MAllocate final : public impl::D2MAllocateBase<D2MAllocate> {
...
+          //  DMA on both sides
+          auto producerRemoteLoad = mlir::dyn_cast<d2m::RemoteLoadOp>(
+              cbUsageInfo[remoteStoreOp.getLocalBuffer()].producers[0]);
+          if (remoteStoreOp.getMemref() == operandCtx.operand->get() &&
+              isAliasedStore(remoteStoreOp) &&
+              (!producerRemoteLoad ||
+               !producerRemoteLoad->hasAttr("d2m.aliased_load"))) {
+            // llvm::errs() << "replacing remote store op with aliased store
+            // op\n";
+            auto allocOp = remoteStoreOp.getLocalBuffer().getDefiningOp();
```
vmilosevic: :warning: **llvm\-qualified\-auto** :warning:
`` auto allocOp `` can be declared as `` auto *allocOp ``
```suggestion
            auto *allocOp = remoteStoreOp.getLocalBuffer().getDefiningOp();
```

lib/Dialect/D2M/Transforms/Allocate.cpp
```diff
@@ -1690,223 +1447,54 @@ class D2MAllocate final : public impl::D2MAllocateBase<D2MAllocate> {
...
-                            MemorySpace memspace) const {
-    if (isOperandExemptFromStreaming(operandCtx, memspace)) {
-      return false;
-    }
-    return inferStreamRequirement(genericOp, genericCtx, operandCtx, memspace);
+  static bool isNonTrivialView(const Value &operand) {
+    return allocation::hasNonTrivialView(operand);
   }
 
+  // TODO: remove since no longer used
```
vmilosevic: :warning: **google\-readability\-todo** :warning:
missing username/bug in TODO
```suggestion
  // TODO(nsmith): remove since no longer used
```

vmilosevic (commented):

lib/Dialect/D2M/Transforms/Allocate.cpp
```diff
@@ -1690,223 +1447,54 @@ class D2MAllocate final : public impl::D2MAllocateBase<D2MAllocate> {
...
                                     MemorySpace memspace) const {
-    if (isNonTrivialView(operandCtx)) {
+    if (isNonTrivialView(operandCtx.operand->get())) {
       return false;
     }
     return operandCtx.isOutput && !allowL1OutputSpilling &&
            memspace != MemorySpace::DeviceDRAM;
   }
 
+  // TODO: move into some DMA utils place alongside load/store aliasing logic
```
vmilosevic: :warning: **google\-readability\-todo** :warning:
missing username/bug in TODO
```suggestion
  // TODO(nsmith): move into some DMA utils place alongside load/store aliasing logic
```

lib/Dialect/D2M/Transforms/Allocate.cpp
```diff
@@ -1690,223 +1447,54 @@ class D2MAllocate final : public impl::D2MAllocateBase<D2MAllocate> {
...
-                                  MemorySpace memspace) const {
-    if (useAlwaysStreamPolicy()) {
+  bool canAliasOperand(d2m::GenericOp genericOp, Value memrefOperand) const {
+    auto memspace =
+        ttcore::getMemorySpace(mlir::cast<MemRefType>(memrefOperand.getType()));
+    auto operandIndex = genericOp.getOperandIndex(memrefOperand);
+    bool isOutput = llvm::find_if(genericOp.getOutputs(), [&](Value operand) {
+                      return operand == memrefOperand;
+                    }) != genericOp.getOutputs().end();
+    // TODO: remove???
```
vmilosevic: :warning: **google\-readability\-todo** :warning:
missing username/bug in TODO
```suggestion
    // TODO(nsmith): remove???
```

lib/Dialect/D2M/Transforms/Allocate.cpp
```diff
@@ -1915,6 +1503,9 @@ class D2MAllocate final : public impl::D2MAllocateBase<D2MAllocate> {
       return false;
     }
 
+    // TODO: remove??? redundant with mcast load check; call this function
```
vmilosevic: :warning: **google\-readability\-todo** :warning:
missing username/bug in TODO
```suggestion
    // TODO(nsmith): remove??? redundant with mcast load check; call this function
```

lib/Dialect/D2M/Transforms/GenericFusion.cpp
```diff
@@ -246,6 +247,152 @@ getFusedOperands(OpOperand *fusedOperand, GenericOp producer,
...
+struct LocalBufferAssociation {
+  LocalBufferAssociation() = default;
+  LocalBufferAssociation(Value operand, bool hasRemoteUse)
+      : operand(operand), hasRemoteUse(hasRemoteUse) {}
+
+  Value operand;
+  bool hasRemoteUse = false;
+};
+
+// TODO: this code is pretty bad, the check on buffer conflict is also not
```
vmilosevic: :warning: **google\-readability\-todo** :warning:
missing username/bug in TODO
```suggestion
// TODO(nsmith): this code is pretty bad, the check on buffer conflict is also not
```

lib/Dialect/D2M/Transforms/GenericFusion.cpp
```diff
@@ -307,6 +454,10 @@ static GenericOp createFusedGeneric(OpOperand *fusedOperand, GenericOp producer,
     appendSource(consumer.getOperation(), co.getOperandNumber());
   }
 
+  // TODO: use helper create buffer for tensor load/store (same as that used in
```
vmilosevic: :warning: **google\-readability\-todo** :warning:
missing username/bug in TODO
```suggestion
  // TODO(nsmith): use helper create buffer for tensor load/store (same as that used in
```

lib/Dialect/D2M/Transforms/SplitUnifiedThread.cpp
```diff
@@ -360,6 +333,7 @@ insertCBOpsForCompute(Block *computeBlock, PatternRewriter &rewriter,
           auto reserveOp = rewriter.create<ReserveOp>(loc, cb);
           rewriter.setInsertionPointAfter(synchronizedOp);
           rewriter.create<PushOp>(loc, cb);
+          // TODO: change isAliased to inherent attribute???
```
vmilosevic: :warning: **google\-readability\-todo** :warning:
missing username/bug in TODO
```suggestion
          // TODO(nsmith): change isAliased to inherent attribute???
```

lib/Dialect/D2M/IR/D2MOps.cpp
```diff
@@ -1629,6 +1629,7 @@ MutableArrayRef<OpOperand> d2m::GenericOp::getInputsAndOutputsMutable() {
 }
 
 // GenericOp verification
+// TODO: check all top level ops touching tensors/memrefs are synchronized
```
missing username/bug in TODO
```suggestion
// TODO(nsmith): check all top level ops touching tensors/memrefs are synchronized
```

lib/Dialect/D2M/IR/D2MOps.cpp
```diff
@@ -2057,7 +2058,11 @@ void GenericOp::getCanonicalizationPatterns(mlir::RewritePatternSet &patterns,
           }
 
           // Find the output alloc by positional counting.
-          Value outputAlloc = GenericOp::getOperandAlloc(region, dpsIOBoundary);
+          Value outputAlloc = nullptr;
+          // TODO: fix if needed, there are checks to get the right cb result,
```
missing username/bug in TODO
```suggestion
          // TODO(nsmith): fix if needed, there are checks to get the right cb result,
```

lib/Dialect/D2M/IR/D2MOps.cpp
```diff
@@ -2313,90 +2318,92 @@ static Block *cloneParallelizedRegion(d2m::GenericOp thisOp,
...
+static void repairParallelizedRegionTypes(Block *newBlock) {
+  // update the local buffers first
+  newBlock->walk([&](Operation *clonedOp) {
+    llvm::errs() << "repairing cloned op " << clonedOp->getName() << "\n";
+    if (auto remoteLoadOp = mlir::dyn_cast<d2m::RemoteLoadOp>(clonedOp)) {
+      llvm::errs() << "repairing remote load op " << remoteLoadOp->getName()
+                   << "\n";
+      // update the associated alloc/empty first
+      Value newMemref = remoteLoadOp.getMemref();
+      auto localBufferDefiningOp =
```
vmilosevic: :warning: **llvm\-qualified\-auto** :warning:
`` auto localBufferDefiningOp `` can be declared as `` auto *localBufferDefiningOp ``
```suggestion
      auto *localBufferDefiningOp =
```

lib/Dialect/D2M/IR/D2MOps.cpp
```diff
@@ -2313,90 +2318,92 @@ static Block *cloneParallelizedRegion(d2m::GenericOp thisOp,
...
       }
-      if (newAllocType != allocOp.getType()) {
-        auto newMemRefType = mlir::dyn_cast<MemRefType>(newAllocType);
-        TT_assert(newMemRefType);
-        allocOp.getResult().setType(newMemRefType);
+    } else if (auto remoteStoreOp =
+                   mlir::dyn_cast<d2m::RemoteStoreOp>(clonedOp)) {
+      // update the associated alloc/empty first
+      Value newMemref = remoteStoreOp.getMemref();
+      auto localBufferDefiningOp =
```
vmilosevic: :warning: **llvm\-qualified\-auto** :warning:
`` auto localBufferDefiningOp `` can be declared as `` auto *localBufferDefiningOp ``
```suggestion
      auto *localBufferDefiningOp =
```

lib/Dialect/D2M/IR/D2MOps.cpp
```diff
@@ -2313,90 +2318,92 @@ static Block *cloneParallelizedRegion(d2m::GenericOp thisOp,
...
-      unsigned numOuts = clonedOp.getNumResults();
+      unsigned numOuts = clonedOp->getNumResults();
       for (unsigned i = 0; i < numOuts; ++i) {
-        clonedOp.getResult(i).setType(
-            clonedOp.getOperand(numIns + i).getType());
+        clonedOp->getResult(i).setType(
+            clonedOp->getOperand(numIns + i).getType());
       }
     }
+    // TODO: still needed?
```
missing username/bug in TODO
```suggestion
    // TODO(nsmith): still needed?
```

vmilosevic (commented):

lib/Dialect/D2M/Transforms/HoistCBAllocs.cpp
```diff
@@ -30,16 +30,55 @@ class D2MHoistCBAllocs : public impl::D2MHoistCBAllocsBase<D2MHoistCBAllocs> {
...
+          // Skip hoisting, these are fake buffers added by fusion that are
+          // guaranteed to not be used
+          assert(!allocOp->getAttrOfType<IntegerAttr>("address") &&
+                 "compute intermediate buffer must not have address attribute");
+        } else if (allocOp->getAttr("d2m.scratch_buffer")) {
+          assert(allocOp->getAttrOfType<IntegerAttr>("address") &&
+                 "scratch buffer must have address attribute");
+          attachCBLayoutAttribute(rewriter, allocOp, 1);
           allocsToHoist.push_back(allocOp);
+        } else if (allocOp->getAttr("d2m.synchronized_buffer")) {
```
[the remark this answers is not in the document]
sohaibnadeemTT: fixed

lib/Dialect/D2M/Transforms/Allocate.cpp
```diff
@@ -1331,230 +1297,83 @@ class D2MAllocate final : public impl::D2MAllocateBase<D2MAllocate> {
...
+          if (remoteLoadOp.getMemref() == operandCtx.operand->get() &&
+              isAliasedLoad(remoteLoadOp)) {
+            // llvm::errs() << "replacing remote load op with aliased load
+            // op\n";
+            //  replace alloc with operand alias op
+            auto allocOp = remoteLoadOp.getLocalBuffer().getDefiningOp();
+            rewriter.setInsertionPoint(allocOp);
+            rewriter.replaceOpWithNewOp<d2m::OperandAliasOp>(
+                allocOp, allocOp->getResultTypes(), remoteLoadOp.getMemref());
+            remoteLoadOp->setAttr("d2m.aliased_load", rewriter.getUnitAttr());
```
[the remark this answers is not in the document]
sohaibnadeemTT: explicit cb form is introduced in SplitUnifiedThread pass

sgholamiTT (approved):
insertDstRegister changes LGTM 👍🏻

lib/Dialect/D2M/Pipelines/D2MPipelines.cpp
```diff
@@ -60,6 +60,7 @@ void createTTIRBufferizationPipeline(OpPassManager &pm,
   // bufferDeallocationOptions;
   // mlir::bufferization::buildBufferDeallocationPipeline(
   //    pm, bufferDeallocationOptions);
+  pm.addPass(d2m::createD2MMarkSynchronizedBuffers());
```
vtangTT: curious why this is included in the bufferization pipeline. Why not move it to after?

lib/Dialect/D2M/Transforms/Allocate.cpp
```diff
@@ -616,82 +625,49 @@ class D2MAllocate final : public impl::D2MAllocateBase<D2MAllocate> {
...
-                    L1memInfo.alignment);
+      genericOp->walk([&](memref::AllocOp allocOp) {
+        std::optional<int32_t> numBuffers = std::nullopt;
+        if (allocOp->getAttr("d2m.scratch_buffer")) {
+          numBuffers = 1;
+        } else if (allocOp->getAttr("d2m.synchronized_buffer")) {
+          // Skip allocating, this is a contract from fusion and compute
+          // lowering saying we will not actually use this buffer and it's just
+          // a placeholder, so we can safely skip it.
+          if (!allocOp->getAttr("d2m.compute_intermediate")) {
```
vtangTT, line 636: confused who/where this attr gets tagged onto the alloc? If it's a compute intermediate would it not just be a scratch? is it for memref allocs that are placeholders for the DST registers?

lib/Dialect/D2M/Transforms/SplitUnifiedThread.cpp
```diff
@@ -28,43 +28,16 @@ namespace {
...
-  if (auto memrefType = mlir::dyn_cast<MemRefType>(memref.getType())) {
-    if (ttcore::getMemorySpace(memrefType) == ttcore::MemorySpace::DeviceDRAM) {
-      return true;
-    }
-  }
-
-  return false;
+bool isAliasedStore(RemoteStoreOp storeOp) {
+  auto operandAliasOp =
+      mlir::dyn_cast<OperandAliasOp>(storeOp.getLocalBuffer().getDefiningOp());
```
vtangTT, line 33: can't we just read `d2m.aliased_store` attr tag?

lib/Dialect/D2M/Transforms/Allocate.cpp
```diff
@@ -616,82 +625,53 @@ class D2MAllocate final : public impl::D2MAllocateBase<D2MAllocate> {
...
+            numBuffers =
+                allocOp->getAttrOfType<IntegerAttr>("d2m.synchronized_buffer")
+                    .getInt();
           }
+        } else {
+          // We can allow this in the future but asserting for now to check it's
+          // not used.
+          std::string opStr;
+          llvm::raw_string_ostream(opStr) << *allocOp;
+          llvm::report_fatal_error(
```
vtangTT: nit: is genericOp->emitERror better here so it propagates failure up?

lib/Dialect/D2M/Transforms/HoistCBAllocs.cpp
```diff
@@ -30,16 +30,77 @@ class D2MHoistCBAllocs : public impl::D2MHoistCBAllocsBase<D2MHoistCBAllocs> {
...
+        if (allocOp->getAttr("d2m.scratch_buffer")) {
+          assert(allocOp->getAttrOfType<IntegerAttr>("address") &&
+                 "scratch buffer must have address attribute");
+          auto newAllocOp = attachCBLayoutAttribute(rewriter, allocOp, 1);
+          allocsToHoist.push_back(newAllocOp);
+        } else if (allocOp->getAttr("d2m.synchronized_buffer")) {
+          if (allocOp->getAttr("d2m.compute_intermediate")) {
+            // Skip hoisting, these are fake buffers added by fusion that are
+            // guaranteed to not be used
+            assert(
```
vtangTT: nit: TT_assert ?

lib/Dialect/D2M/Transforms/Allocate.cpp
```diff
@@ -1694,223 +1367,35 @@ class D2MAllocate final : public impl::D2MAllocateBase<D2MAllocate> {
...
-                                  MemorySpace memspace) const {
+  /// @return `true` if the remote_load/store for this operand does not require
+  /// real DMA. (i.e. can be aliased with it's corresponding operand). This is
+  /// the case when the operand:
+  ///  - does not have a non-trivial view layout
+  ///  - is in L1 and not DRAM
+  ///  - is not in an explicit datamovement form and the dimensions
+  ///  corresponding to the load/store indices
+  ///    are not broadcast or reduction.
+  /// This is a common function shared byt isAliasedLoad and isAliasedStore.
```
vtangTT: nit: typo

lib/Dialect/D2M/Transforms/SplitUnifiedThread.cpp
sohaibnadeemTT: I mainly added this for readability. Once aliasing has been chosen, we use the operand_alias op to check for aliasing.

lib/Dialect/D2M/Transforms/Allocate.cpp
sohaibnadeemTT: I haven't added this yet, but it will be in the MarkSynchronizedBuffers pass. This is meant for memrefs that are the output of one linalg and the input of another. Prior to this refactor, we didn't actually allocate these and that happens to work out since we always fuse linalgs together into a single sycnhornized compute region, so this attribute is meant to maintain parity with that behaviour. We shouldn't have to do this but if I remove it, it may give OOM for cases where we may actually have enough memory after fusing and removing these intermediates, but Allocator doesn't know this since it's positioned earlier

lib/Dialect/D2M/Transforms/Allocate.cpp
sohaibnadeemTT: fixed

lib/Dialect/D2M/Transforms/HoistCBAllocs.cpp
sohaibnadeemTT: fixed

lib/Dialect/D2M/Transforms/Allocate.cpp
sohaibnadeemTT: fixed

lib/Dialect/D2M/Pipelines/D2MPipelines.cpp
sohaibnadeemTT: I've moved it after, but the intention was to not have passes inserted in between bufferization and this pass (since they would most likely be hacks)

lib/Dialect/D2M/Transforms/Allocate.cpp
vtangTT: ok yea, so it does represent the DST registers?

sohaibnadeemTT: Not really, it's an actual alloc to store the output of a linalg. However, when we fuse, we are no longer generate the **full** output of these intermediary linalgs, we only generate the **full** output of the final linalg and use scratch whenever we need to store some intermediate values.  So this alloc never gets used.

lib/Dialect/D2M/Transforms/HoistCBAllocs.cpp
```diff
@@ -30,16 +30,77 @@ class D2MHoistCBAllocs : public impl::D2MHoistCBAllocsBase<D2MHoistCBAllocs> {
...
     // IR.  Erasing allocs shifts positional lookups.
     SmallVector<memref::AllocOp> allocsToHoist;
     for (Region &region : genericOp->getRegions()) {
       region.walk([&](memref::AllocOp allocOp) {
-        auto memrefType = allocOp.getType();
-        if (mlir::isa<ttcore::CBLayoutAttr>(memrefType.getLayout()) &&
-            allocOp->getAttrOfType<IntegerAttr>("address")) {
-          allocsToHoist.push_back(allocOp);
+        if (allocOp->getAttr("d2m.scratch_buffer")) {
+          assert(allocOp->getAttrOfType<IntegerAttr>("address") &&
```
anuragsingh-tt: Is assert the right legality check here? Does it go away in release builds?

sohaibnadeemTT: If there is a scratch buffer being used by compute, it must be an actual buffer with an address, hence the assert. 

The assert will go away in release. Is the suggestion that it should be checked in release or that it should not?

test/ttmlir/Dialect/D2M/allocate/allocate_reblock_cb_layout_consistency.mlir
```diff
@@ -1,92 +0,0 @@
-// RUN: ttmlir-opt --ttcore-register-device "--d2m-allocate=test-assume-l1-capacity=8388608" -o %t %s
```
sohaibnadeemTT, line 1: This test was meant to ensure that whenever we have a cb layout attribute, we assign an address too. However, we have moved the cb_layout attribute down to HoistCBAllocs pass now so this test is no longer necessary. The cb_layout attribute now mainly serves as the way to tell the backend that this is an in-generic alloc which should be backed with a CB (regardless of whether or not it actually needs synchronization, since we don't support any other mechanism for creating l1 memory that only lives for the duration of an op). Differentiating in-generic allocs is now done via an attribute on the memref.alloc itself (d2m.synchronized_buffer for buffers that need sychronization vs d2m.scratch_buffer for scratch buffers, etc.)

test/ttmlir/Dialect/D2M/allocate/allocate_tile_untilize_result_type.mlir
```diff
@@ -1,41 +0,0 @@
-// RUN: ttmlir-opt --ttcore-register-device "--d2m-allocate=stream-insert-policy=infer test-buffer-size-policy=max" -o %t %s
```
sohaibnadeemTT, line 1: tilize and untilize have made to been proper DPS (no result post-bufferization) so this test is no longer valid

test/ttmlir/Dialect/D2M/allocate/allocate_update_remote_ops_after_stream_insert.mlir
```diff
@@ -1,79 +0,0 @@
-// RUN: ttmlir-opt --d2m-allocate %s 2>&1 | FileCheck %s
```
sohaibnadeemTT, line 1: we don’t insert stream_layout in Allocator anymore, this test is stale (remote_load already has a view into it's input)

sohaibnadeemTT (commented):
Added explanations above for lit tests that are no longer needed and have been deleted

lib/Dialect/D2M/Transforms/HoistCBAllocs.cpp
anuragsingh-tt: I think TT_assertv is better here, thanks for changing.

lib/Dialect/D2M/Pipelines/D2MPipelines.cpp
anuragsingh-tt: Following up on this. Wondering why it is not immediately before Allocate? 

I see `InsertScratch` / `GenericApplyInterchange` / `DecomposeMasking` in between the `SynchronizedBuffers` pass and `Allocate`. First two don't really interact with this pass but the third seems to redundantly stamp synchronized_buffers.

lib/Dialect/D2M/Transforms/Allocate.cpp
anuragsingh-tt: Should MarkSynchronizedBuffers be picking up this responsibility? It sounds like a job for a fusion pass prior to Allocate. Maybe GenericFusion or a new pass right after it? 

> We shouldn't have to do this but if I remove it, it may give OOM for cases where we may actually have enough memory after fusing and removing these intermediates, but Allocator doesn't know this since it's positioned earlier

I see the issue here and it fits in a general bucket of problems in how backend fusion and Allocate interact right now. I think we need to do more fusion analysis in the frontend pre-Allocate instead of having Allocate overestimate the actual footprint realized by hardware bc the fusion backend passes are identifying savings. In my mind Allocate should only see either scratch or real CBs, no placeholders. We may even want to create a new op (not memref.alloc) that represents fused_intermediates instead of string attributes. Tbh this merits a further discussion. I have some ideas but am interested in your thoughts.

anuragsingh-tt: The above comment is an aside, not a blocker for this PR. Mostly a response to this:

> I haven't added this yet, but it will be in the MarkSynchronizedBuffers pass.

include/ttmlir/Dialect/D2M/Transforms/Passes.td
```diff
@@ -592,6 +592,31 @@ defvar D2MStreamingOptions = [
...
+    the CBLayoutAttr for these buffers.
+  }];
+
+  let dependentDialects = [
+    "::mlir::tt::d2m::D2MDialect",
+    "::mlir::memref::MemRefDialect"
+  ];
+
+  let options = [
+    Option<"numStreamBuffers",
```
anuragsingh-tt: Doesn't look like this is being used in the PR but I am wondering if this is going to be where the placeholder pipeline option that Allocate currently uses or something more involved? As we discussed offline I agree this pass should be _classifying_ which allocs need CBs but _depth_ of that decision (ie single, double back buffers) should be determined by the reblocking analysis as part of the candidate search space.

sohaibnadeemTT: Yeah this is just meant to be a default value when adding the d2m.synchronized_buffer attribute (basically the same thing we were doing when setting the CBLayoutAttr). Later, when reblocking chooses the optimal number of back buffers, it can update the d2m.synchronized_buffer to whatever it chose. Let me know what you think.

anuragsingh-tt: There is already a pipeline option in Allocate (num-stream-buffers). My preference is to not duplicate it, can we inherit from that instead?

anuragsingh-tt: [Created an issue to track.](https://github.com/tenstorrent/tt-mlir/issues/8443) It is on my plate.

lib/Dialect/D2M/Transforms/Allocate.cpp
```diff
@@ -614,86 +623,61 @@ class D2MAllocate final : public impl::D2MAllocateBase<D2MAllocate> {
...
+          // Skip allocating, this is a contract from fusion and compute
+          // lowering saying we will not actually use this buffer and it's just
+          // a placeholder, so we can safely skip it.
+          if (!allocOp->getAttr("d2m.compute_intermediate")) {
+            numBuffers =
+                allocOp->getAttrOfType<IntegerAttr>("d2m.synchronized_buffer")
+                    .getInt();
           }
+        } else {
+          // We can allow this in the future but asserting for now to check it's
```
anuragsingh-tt, line 642: What class of alloc would this be?

lib/Dialect/D2M/Transforms/Allocate.cpp
anuragsingh-tt: NIT: update this

lib/Dialect/D2M/Transforms/Allocate.cpp
```diff
@@ -1283,23 +1218,14 @@ class D2MAllocate final : public impl::D2MAllocateBase<D2MAllocate> {
   /// modifications:
   ///  - modify root alloc ops and any view layout ops to be in the final
   ///    memspace decided by the planner;
-  ///  - create local CB allocs inside the generic for streamed operands.
   ///
-  LogicalResult insertOperandStreams(func::FuncOp funcOp,
-                                     const FuncAnalysisData &analysis) {
+  LogicalResult remapMemorySpacesOnOperandViews(func::FuncOp funcOp,
```
anuragsingh-tt: super nit: not just views so maybe a better name for this?

lib/Dialect/D2M/Transforms/Allocate.cpp
```diff
@@ -2,13 +2,16 @@
...
+#include "ttmlir/Dialect/D2M/IR/D2MOps.h"
 #include "ttmlir/Dialect/D2M/Transforms/Passes.h"
 
 #include "ttmlir/AffineMapUtils.h"
 #include "ttmlir/Asserts.h"
 #include "ttmlir/Dialect/D2M/Analysis/Allocation/Planner.h"
 #include "ttmlir/Dialect/D2M/Analysis/Allocation/Utils.h"
 #include "ttmlir/Dialect/D2M/Analysis/BlockFactorAnalysis.h"
+#include "ttmlir/Dialect/D2M/IR/D2MGenericRegionOps.h"
+#include "ttmlir/Dialect/D2M/Utils/SynchronizableOpInterfaceUtils.h"
```
anuragsingh-tt: Is this used by the file? Looks like Allocate is just consuming the synchronized_buffer tag stamped in the pass prior to this.

lib/Dialect/D2M/Transforms/Allocate.cpp
```diff
@@ -1332,235 +1258,65 @@ class D2MAllocate final : public impl::D2MAllocateBase<D2MAllocate> {
...
         }
-        TT_assert(remappedMemSpace.has_value());
-        const MemorySpace finalRemappedMemSpace = *remappedMemSpace;
-        if (requiresCBAllocation(genericOp, genericCtx, operandCtx,
-                                 *remappedMemSpace)) {
-
-          // Record all aliases along the operand def-chain.
-          // This is needed because a generic operand may have several aliases
-          // along the def-chain (e.g. views, root alloc, etc.) and nested
-          // remote ops may reference any of them.
```
anuragsingh-tt, line 1343: How are we handling this now?

lib/Dialect/D2M/Transforms/Allocate.cpp
```diff
@@ -354,7 +359,7 @@ class D2MAllocate final : public impl::D2MAllocateBase<D2MAllocate> {
 
     // This pass works with two sets of IR objects: memrefs (defined by
```
anuragsingh-tt, line 353: Feel free to update this comment as it is a bit stale. :)

lib/Dialect/D2M/Transforms/Allocate.cpp
```diff
@@ -387,6 +392,10 @@ class D2MAllocate final : public impl::D2MAllocateBase<D2MAllocate> {
       return failure();
     }
 
+    if (failed(convertAliasedLoadStore(funcOp, analysis))) {
```
anuragsingh-tt: super nit: can we call this materializeOperandAliases? My mental model was analysis then execution for Allocator and this is kind of a pre-Allocate rewrite. Open to your thoughts.

lib/Dialect/D2M/Transforms/Allocate.cpp
sohaibnadeemTT: This was mainly used to update the result type of remote_load/remote_store (and tilize/untilize) after a CB layout attribute was attached to the memref:
a) we no longer attach a CB layout attribute on the local buffer
b) these ops are proper DPS now so we don't even have a result type post-bufferization (remote_load and remote_store were changed in @nsmithtt's recent PR (https://github.com/tenstorrent/tt-mlir/pull/8361) and tilize/untilize have been changed in this PR)

lib/Dialect/D2M/Transforms/Allocate.cpp
sohaibnadeemTT: fixed it up a bit, let me know if there's something im missing

lib/Dialect/D2M/Transforms/Allocate.cpp
sohaibnadeemTT: changed to `updateMemorySpaces`

lib/Dialect/D2M/Transforms/Allocate.cpp
sohaibnadeemTT: you're right these are stale changes, removed.

lib/Dialect/D2M/Transforms/Allocate.cpp
sohaibnadeemTT: fixed it up a bit, let me know if there's something else to add

lib/Dialect/D2M/Transforms/Allocate.cpp
sohaibnadeemTT: Makes sense, I've changed it to materializeAliasedLoadStore (still want to refer to load/store in the name since that is what is being aliased here)

lib/Dialect/D2M/Transforms/Allocate.cpp
sohaibnadeemTT: I haven't thought of a specific use-case yet to be honest but I was thinking it would just be general alloc that is not a synchronization buffer. We don't really use such a thing right now so I have guarded it with an assert so that folks don't accidentally create allocs that are meant to be one of the other classes but didn't attach the attribute correctly. If we do need this down the line, we can just remove the assert

lib/Dialect/D2M/Transforms/Allocate.cpp
```diff
@@ -387,6 +386,10 @@ class D2MAllocate final : public impl::D2MAllocateBase<D2MAllocate> {
...
```
vmilosevic: :warning: **clang\-diagnostic\-error** :warning:
use of undeclared identifier `` convertAliasedLoadStore ``

vmilosevic (commented):

vtangTT (approved):

include/ttmlir/Dialect/D2M/Transforms/Passes.td
sohaibnadeemTT: I've separated out numStreamBuffers definition as a separate defvar and re-used it in both markSynchronizedBuffers pass and DecomposeMasking pass. markSynchronizedBuffers also inherits from the pipeline option

nsmithtt (approved):

lib/Dialect/D2M/Transforms/Allocate.cpp
sohaibnadeemTT: I agree it should be set by GenericFusion but it's a pre-bufferization pass and the attributes would not be carried over from tensor.empty to memref.alloc, adding the logic in MarkSynchronizedBuffers seems like a simple solution for now to maintain similar behaviour.

lib/Dialect/D2M/Pipelines/D2MPipelines.cpp
sohaibnadeemTT: As mentioned above, I intended to have MarkSynchronizedBuffers be just after bufferization to mark memref.allocs created by the bufferization pass, and if there are other passes after that are manually introducing buffers, they should explicitly be tagging however they want (scratch, syncrhonized buffer, etc.).

lib/Dialect/D2M/Transforms/Allocate.cpp
anuragsingh-tt: I see your point. I think it should still be a separate pass from MarkSynchronized, maybe `MarkFusionIntermediates` or something. We should think about creating a new op long term instead of attributes.

anuragsingh-tt (approved):

PR #8380, opened 2026-05-12
[TTNN] Remove identity variant of `mesh_shard` op

dmilinkovicTT:
### Ticket
#8266

### Problem description
`Identity` variant of the `ttir::mesh_shard` op is a pure semantic decorator:
- it just performs the `global` -> `local` shape translation in the IR;
- it is a no-op from the runtime perspective.

This op gets lowered to `ttnn::mesh_shard` - a no-op which doesn't actually exist in the TTNN library.

Many TTIR/TTNN passes need to somehow work around this op, since it's an exception to most of the rules other ops are implicitly following.

Ideally, we should always have the **per-device (local) shapes** for pre-sharded tensor arguments and results in the IR. This requires patching the function argument / result types from global to local shapes in SHLO -> TTIR conversion. After that, the `Identity` variant of the `mesh_shard` op can be safely removed.

### What's changed (commit-by-commit)
1. **StableHLO: drop identity mesh_shard, retype presharded args:** 
  - Rewrite the Shardy and GSPMD conversion patterns so a presharded function arg no longer round-trips through an identity `ttir.mesh_shard` - instead, retype the block arg (and matching func result) to its per-device local shape and forward the value through
  - Add a `MeshShardingCollection` helper that consolidates the GSPMD and Shardy `MeshSharding` instantiation using the existing constructors (tt-xla picks this up in the uplift, and drops the duplicated instantiation logic)

2. **Remove identity mesh_shard op and all references:** 
  - Drop the `Identity` variant of `MeshShardType` and rip out every codepath that handled it

3. **builder: retype presharded arg to its per-device local shape.** 

3. **Test updates for identity mesh_shard removal.** Update or delete every `.mlir` and python test that referenced the identity shard type or `ttir.mesh_shard` for presharded args.

### Checklist

---
### tt-xla Validation

**Base (uplifted):** \`eb9005fa36\`
**tt-xla branch:** \`dmilinkovic/remove-identity-mesh-shard\`

| Test Suite | Run | Status |
|------------|-----|--------|
| full-mlir-uplift-qualification.json | [Run #26035653132](https://github.com/tenstorrent/tt-xla/actions/runs/26035653132) | :white_check_mark: passed |


jgrimTT (approved):
Builder changes lgtm!

include/ttmlir/Dialect/TTCore/IR/TTCoreOpsEnums.td
```diff
@@ -156,14 +156,12 @@ def TTCore_MeshShardDirection: I32EnumAttr<"MeshShardDirection", "TT MeshShardDi
   let cppNamespace = "::mlir::tt::ttcore";
 }
 
-def TTCore_MeshShardType_Identity  : I32EnumAttrCase<"Identity",  0, "identity">;
```
svuckovicTT, line 159: Should the indices below change to 0-indexed?

include/ttmlir/Dialect/TTNN/IR/TTNNOps.td
```diff
@@ -3013,30 +3013,6 @@ def TTNN_AllReduceAsyncOp: TTNN_Op<"all_reduce_async", [OpModelExempt]> {
...
-
-    let extraClassDeclaration = [{
-      wa::TTNNOperandsWorkarounds getOperandsWorkarounds() {
-        return wa::TTNNOperandsWorkaroundsFactory::createMeshShardOpOperandsWorkarounds(getShardType());
-      }
-    }];
-
-    let hasVerifier = 1;
-}
-
```
svuckovicTT, line 3069: ❤️ ❤️ ❤️

test/ttmlir/Dialect/TTIR/propagate_weight_dtype/propagate_through_mesh_shard.mlir
```diff
@@ -11,7 +11,7 @@ module {
     %arg0: tensor<32x64xbf16>,
     %arg1: tensor<64x128xbf16> {ttcore.argument_type = #ttcore.argument_type<parameter>, ttcore.weight_dtype = "bfp_bf8"}
   ) -> tensor<32x64xbf16> {
-    %0 = "ttir.mesh_shard"(%arg1) <{shard_dims = array<i64: -1, 1>, shard_direction = #ttcore.shard_direction<full_to_shard>, shard_shape = array<i64: 1, 2>, shard_type = #ttcore.shard_type<identity>}> : (tensor<64x128xbf16>) -> tensor<64x64xbf16>
```
svuckovicTT, line 14: Is this test needed anymore? We removed mesh_shard as an op thru which to propagate (now an empty list of ops), in one of the files above.

svuckovicTT: `lib/Dialect/TTNN/Transforms/TTNNKVCacheDtypeConversion.cpp`

svuckovicTT (commented):
Yeeeeesssss

Got a few more things to go thru before approving, will continue tmr!

include/ttmlir/Dialect/TTCore/IR/TTCoreOpsEnums.td
dmilinkovicTT: The only reason to keep it 1-indexed is to ensure that a FB compiled with an older compiler version fails explicitly in a new version of runtime (something like "ERROR: Invalid mesh shard type: 0"), instead of letting the runtime interpret it as the next enum entry, which can lead to data corruption and a cryptic error message.

But I'm not sure if we have any policies on FB backwards-compatability?

test/ttmlir/Dialect/TTIR/propagate_weight_dtype/propagate_through_mesh_shard.mlir
dmilinkovicTT: In `TTNNKVCacheDtypeConversion`, it was`ttnn::mesh_shard` which was being allowed as a pass-through op. However, here, it's `ttir::mesh_shard`, which still exists for non-identity scenarios.

include/ttmlir/Dialect/TTCore/IR/TTCoreOpsEnums.td
svuckovicTT: TBH I'm not sure if we have such a policy - we did have some form of versioning in fb/ttrt path which was producing more problems than it was solving, in the end we gave up on it.

All that to say - I don't know and have no strong opinions :)

svuckovicTT (approved):
LGTM!

sdjordjevicTT (approved):
Niceeeee!

acicovicTT (approved):
We discussed all of the concerns/questions offline, LGTM!

pilkicTT (approved):

mmilosevicTT (approved):
Chisel changes LGTM

nsmithtt (approved):

PR #8384, opened 2026-05-12
PagedSdpaDecode: workaround pass attribute for large head_dim

sshonTT:
### Ticket
[Link to Github Issue](https://github.com/tenstorrent/tt-xla/issues/4596#issuecomment-4426475057)

### Problem description
L1 OOM issue when running Gemma4 31b model which has head_dim >=256

### What's changed
Adds an optional program_config attribute on
ttnn.paged_scaled_dot_product_attention_decode and a TTNN workaround pattern that populates it when the default schedule overflows per-core L1. The runtime reads the attribute when set; the OpModel/validation path sees the same config that will execute.

This replaces the previous design that inlined the schedule into the runtime: by lifting the decision to a compile-time pattern, the runtime stays a thin executor and a future pass / external producer can supply its own program_config without forking the runtime heuristic.

Why head_dim is the discriminator
---------------------------------
tt-metal's SDPA decode program factory allocates per-core CBs whose footprint scales with DHt = head_dim / TILE_WIDTH (K/V tiles double- buffered) and with num_cores_per_head for the partial-accumulator CB:

  k_tiles               = Sk_chunk_t_cb_size * DHt  * 2
  v_tiles               = Sk_chunk_t_cb_size * vDHt * 2
  intermed_output_tiles = (PNHt * vDHt + 2 * PNHt) * (num_cores_per_head - 1)

For Gemma-4 31B (head_dim=512 global, head_dim=256 sliding-window) this overflows Blackhole's 1.5 MB usable L1 under the default schedule. Llama-class shapes (head_dim=128) with similar total K/V residency stay within budget, so the trigger is the per-CB scaling factor, not the aggregate K/V volume.

Override schedule
-----------------
When head_dim >= 256, set k_chunk_size=32 (one page, bounds Sk_chunk_t_cb_size) and cap max_cores_per_head_batch at min(32, blocks). The 32-core cap (vs the 64-core MAX_TREE_REDUCTION_ROUNDS limit) halves the intermed-output footprint, which is what keeps Gemma-4 31B at max_model_len >= 2048 (pageTableBlocks >= 64) inside per-core L1.

Components
----------
* IR: TTNN_PagedScaledDotProductAttentionDecodeOp gets an optional TTNN_SDPAProgramConfigAttr program_config.
* Flatbuffer: PagedScaledDotProductAttentionDecodeOp table carries a SDPAConfig program_config; TTNNToFlatbuffer serializes it from the attribute.
* Workaround pass: new PagedScaledDotProductAttentionDecodeProgramConfigRewritePattern fires only when head_dim >= 256 and program_config is unset; it populates the attribute from the device worker grid and the page-table block count.
* Runtime: prefers op->program_config() via utils::createSDPAProgramConfig when present; otherwise keeps the existing non-causal / Blackhole-causal fallbacks unchanged.
* OpModel: reads the attribute through conversion::getSDPAProgramConfig so the validation/fallback pass sees the same program_config the runtime executes; falls back to the existing core_grid-driven default when the attribute is absent.

### Checklist

### DIFF
--- a/include/ttmlir/Dialect/TTNN/IR/TTNNOps.td
+++ b/include/ttmlir/Dialect/TTNN/IR/TTNNOps.td
@@ -3640,7 +3640,8 @@ def TTNN_PagedScaledDotProductAttentionDecodeOp : TTNN_Op<"paged_scaled_dot_prod
                        Optional<AnyRankedTensor>:$cur_pos_tensor,
                        Optional<AnyRankedTensor>:$attention_sink,
                        OptionalAttr<F32Attr>:$scale,
-                       OptionalAttr<UI32Attr>:$sliding_window_size);
+                       OptionalAttr<UI32Attr>:$sliding_window_size,
+                       OptionalAttr<TTNN_SDPAProgramConfigAttr>:$program_config);
 
   let results = (outs AnyRankedTensor:$result);
 
--- a/include/ttmlir/Dialect/TTNN/Transforms/Workarounds/Decomposition/PagedScaledDotProductAttentionDecodeProgramConfigRewritePattern.h
+++ b/include/ttmlir/Dialect/TTNN/Transforms/Workarounds/Decomposition/PagedScaledDotProductAttentionDecodeProgramConfigRewritePattern.h
@@ -1,0 +1,39 @@
+// SPDX-FileCopyrightText: (c) 2026 Tenstorrent AI ULC
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#ifndef TTMLIR_DIALECT_TTNN_TRANSFORMS_WORKAROUNDS_DECOMPOSITION_PAGEDSCALEDDOTPRODUCTATTENTIONDECODEPROGRAMCONFIGREWRITEPATTERN_H
+#define TTMLIR_DIALECT_TTNN_TRANSFORMS_WORKAROUNDS_DECOMPOSITION_PAGEDSCALEDDOTPRODUCTATTENTIONDECODEPROGRAMCONFIGREWRITEPATTERN_H
+
+#include "ttmlir/Dialect/TTNN/IR/TTNNOps.h"
+
+#include "mlir/IR/PatternMatch.h"
+#include "mlir/Support/LogicalResult.h"
+
+namespace mlir::tt::ttnn::workarounds::decomposition {
+
+// Sets an explicit SDPAProgramConfig on PagedScaledDotProductAttentionDecodeOp
+// for head_dim >= 256, where the default TTNN schedule overflows per-core L1.
+// Scoped to opt-level=0; at opt-level>=1 OpValidationAndFallback owns this
+// decision.
+//
+// Metal issue: https://github.com/tenstorrent/tt-metal/issues/44311
+class PagedScaledDotProductAttentionDecodeProgramConfigRewritePattern
+    : public OpRewritePattern<ttnn::PagedScaledDotProductAttentionDecodeOp> {
+public:
+  PagedScaledDotProductAttentionDecodeProgramConfigRewritePattern(
+      MLIRContext *context, int64_t optimizationLevel)
+      : OpRewritePattern<ttnn::PagedScaledDotProductAttentionDecodeOp>(context),
+        optimizationLevel(optimizationLevel) {}
+
+  LogicalResult
+  matchAndRewrite(ttnn::PagedScaledDotProductAttentionDecodeOp srcOp,
+                  PatternRewriter &rewriter) const override;
+
+private:
+  int64_t optimizationLevel;
+};
+
+} // namespace mlir::tt::ttnn::workarounds::decomposition
+
+#endif // TTMLIR_DIALECT_TTNN_TRANSFORMS_WORKAROUNDS_DECOMPOSITION_PAGEDSCALEDDOTPRODUCTATTENTIONDECODEPROGRAMCONFIGREWRITEPATTERN_H
--- a/runtime/lib/ttnn/operations/transformer/paged_scaled_dot_product_attention_decode.cpp
+++ b/runtime/lib/ttnn/operations/transformer/paged_scaled_dot_product_attention_decode.cpp
@@ -4,6 +4,7 @@
 
 #include "operations/transformer/paged_scaled_dot_product_attention_decode.h"
 
+#include "tt/runtime/detail/ttnn/operations/utils.h"
 #include "tt/runtime/detail/ttnn/utils.h"
 
 namespace tt::runtime::ttnn::operations::transformer {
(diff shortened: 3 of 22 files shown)


rpavlovicTT (changes requested):
Hey @sshonTT , I'd explore other ways of handling PagedSDPA's program config. Currently , we have support in the compiler for Conv2dConfig, MatmulProgramConfig and maybe some others (but these are important). And those are generated in the optimizer passes and carefully evaluated (validation, L1 usage). We should strive to keep same mechanism for other ops like `PagedSDPA`. 
Workarounds are definitely not the best place for this. But any logic that manages L1 consumption should be contained within optimizer passes.  
For example, in `Conv2dConfig` case, we even have a "search space" for searching for the best param combination for a given conv2d op. However, this is not needed. Feel free to explore and propose a solution for `PagedSDPA` case.

sshonTT:
Thanks for the review, @rpavlovicTT. I completely agree with your points regarding architectural consistency. Centralizing the L1 evaluation logic and routing it through the OpModel validation path is definitely the ideal direction, similar to how Conv2dConfig and MatmulProgramConfig are handled.

However, there are a few reasons why I believe the current approach is necessary for this specific case right now:

1. opt_level=0 Support
For Gemma-4 (head_dim >= 256), an L1 overflow occurs with the default schedule. This is a hard hardware constraint rather than an optimization issue. Since the optimizer is disabled by default (opt_level=0), this fix must be applied regardless of the optimization level.

2. consistency
There's same pattern in [here](https://github.com/tenstorrent/tt-mlir/blob/main/lib/Dialect/TTNN/Transforms/Workarounds/Decomposition/SliceStaticOpRewritePattern.cpp#L44-L50)

3. Deterministic Fix vs. Performance Search
Conv2dConfig is designed to explore a search space to find performance trade-offs. In contrast, this PR is a deterministic workaround for a hardware constraint (e.g., deterministically mapping (head_dim, page_table_blocks) to (k_chunk=32, max_cores=32)). The optimizer's performance search mechanism isn't really applicable here. We can defnitly discuss this in the future. If you think it's reasonable, I can post an issue.

While I agree that full optimizer integration is the right long-term direction, I think we should merge this PR as-is because this is a model bringup stage.

Maybe we can use the new l1 aware framework from this one https://github.com/tenstorrent/tt-mlir/pull/8299 once it's merged in. What do you think?

rpavlovicTT:
@sshonTT 
To your points:
1. Yes `opt-level=0` is a problem case. In future we will have OpValidationPass included in this case, but until then unfortunately we have to rely on workaround. I propose to have your workaround enabled only for this opt-level. 
2. It's not completely the same pattern. i.e. your workaround is definitely not a **decomposition** pattern. 
3. For conv2d op we do search, but for matmul program config we actually fix it in compiler. So for `PagedSDPADecode` we could also create a fixed program config that depends on op's attributes and tensor shapes. 

I have few asks for you:
1. Limit your workaround for opt-level=0 case (there should already be `optimizationLevel` passed to `lib/Dialect/TTNN/Transforms/Workarounds/TTNNWorkaroundsPatterns.cpp`).
2. For opt-level > 0, open an issue describing that we need support for generating program config and validation of PagedSDPADecode op, and ability to tweak program config so that op fits L1. Something similar we do in ValidationAndFallbackPass for conv2d op. If you want to, you can implement fix for this by yourself or leave it for us to pick it up.

On a general note, we are striving to limit workarounds and replace most of them by automatic ValidationAndFallback pass. Let me know if you have questions.

sshonTT:
Hey @rpavlovicTT, I will follow your suggestion. 
I've gated this workaround behind when opt_level==0, and posted an issue for optimizer upon [here](https://github.com/tenstorrent/tt-mlir/issues/8417).

include/ttmlir/Dialect/TTNN/Transforms/Workarounds/Decomposition/PagedScaledDotProductAttentionDecodeProgramConfigRewritePattern.h
```diff
@@ -0,0 +1,37 @@
...
+#include "mlir/IR/PatternMatch.h"
+#include "mlir/Support/LogicalResult.h"
+
+namespace mlir::tt::ttnn::workarounds::decomposition {
+
+// Sets an explicit SDPAProgramConfig on PagedScaledDotProductAttentionDecodeOp
+// for head_dim >= 256, where the default TTNN schedule overflows per-core L1.
+// Scoped to opt-level=0; at opt-level>=1 OpValidationAndFallback owns this
+// decision.
+class PagedScaledDotProductAttentionDecodeProgramConfigRewritePattern
```
sdjordjevicTT, line 21: Can you please open an issue on the tt-metal side with a repro and attach it here as a comment?

test/ttmlir/Dialect/TTNN/Transforms/Workarounds/paged_sdpa_decode_program_config_workaround.mlir
```diff
@@ -0,0 +1,44 @@
+// RUN: ttmlir-opt --split-input-file --ttcore-register-device --ttnn-layout --ttnn-workaround -o %t %s
```
sdjordjevicTT, line 1: Please add a builder test that will catch a metal fix:
https://github.com/tenstorrent/tt-mlir/blob/main/test/python/golden/ttir_ops/workarounds/decomposition/test_decomposition_workarounds.py

test/ttmlir/Dialect/TTNN/Transforms/Workarounds/paged_sdpa_decode_program_config_workaround.mlir
```diff
@@ -0,0 +1,44 @@
...
+// RUN: FileCheck %s --input-file=%t
```
sdjordjevicTT, line 2: Please also add a builder test that tests this SDPA shape.

include/ttmlir/Dialect/TTNN/Transforms/Workarounds/Decomposition/PagedScaledDotProductAttentionDecodeProgramConfigRewritePattern.h
sshonTT: Yes, added it.

test/ttmlir/Dialect/TTNN/Transforms/Workarounds/paged_sdpa_decode_program_config_workaround.mlir
sshonTT: Yeah, added both xfailing case where doesn't take workaround pass, and positive one.

rpavlovicTT (approved):

include/ttmlir/Dialect/TTNN/IR/TTNNOps.td
```diff
@@ -3640,7 +3640,8 @@ def TTNN_PagedScaledDotProductAttentionDecodeOp : TTNN_Op<"paged_scaled_dot_prod
                        Optional<AnyRankedTensor>:$cur_pos_tensor,
                        Optional<AnyRankedTensor>:$attention_sink,
                        OptionalAttr<F32Attr>:$scale,
-                       OptionalAttr<UI32Attr>:$sliding_window_size);
+                       OptionalAttr<UI32Attr>:$sliding_window_size,
+                       OptionalAttr<TTNN_SDPAProgramConfigAttr>:$program_config);
```
svuckovicTT, line 3644: Can you please update the sdpa op conversions in `TTNNToEmitPy.cpp` and `TTNNToEmitC.cpp`? Thanks!

mtopalovicTT (approved):

include/ttmlir/Dialect/TTNN/IR/TTNNOps.td
sshonTT: Updated them. Thanks for review

runtime/lib/ttnn/operations/transformer/paged_scaled_dot_product_attention_decode.cpp
```diff
@@ -46,18 +47,22 @@ static void runPagedScaledDotProductAttentionDecodeOp(
...
-    // Preserve the existing causal decode scheduling while disabling the
-    // approximate exponential path on Blackhole.
     programConfig.emplace();
     programConfig->q_chunk_size = 0;
     programConfig->k_chunk_size = 0;
     programConfig->compute_with_storage_grid_size = computeGrid;
     programConfig->max_cores_per_head_batch = computeGrid.x * computeGrid.y;
+  }
+
+  if (programConfig.has_value() &&
```
jameszianxuTT, line 66: Did you mean to remove the comment? I'm not sure that it says anything useful to be fair
```
    // Preserve the existing causal decode scheduling while disabling the
    // approximate exponential path on Blackhole.
```

jameszianxuTT (approved):

runtime/lib/ttnn/operations/transformer/paged_scaled_dot_product_attention_decode.cpp
sshonTT: Yes, intentional. But I think we need a proper comment why we need this, and how we track it. So I've added issue number in tt-metal to the comment. Thanks.

PR #8386, opened 2026-05-12
[EmitPy] Fix targetModule path 

amilovanovicTT:
### Ticket
Part of #8258

### Problem description
Previously, two different paths in the TTNN-to-EmitPy pipeline served the same purpose of exporting a Python module for end-to-end testing: `targetModule` path and `TTNNCreateMainForTest` pass. On the frontend side, only the latter is used. However, there are a few points in favour of the `targetModule` path:
- `targetModule` was made as a completely separate path from the canonical one, whereas `TTNNCreateMainForTest` is part of the canonical codegen path
- `TTNNCreateMainForTest` leaves the input generator functions present in the generated code, which are redundant for testing since inputs are provided by the frontend
- `TTNNCreateMainForTest` pass conflicts with the recently added `TTNNSplitForwardFuncArgsByType` pass since they are both part of the canonical codegen path (the hot-fix is done in #8323, but this has to be properly addressed otherwise). However, `TTNNSplitForwardFuncArgsByType` pass that splits the forward function inputs into activations and weights is not really relevant for the testing

To reconcile these two, the agreement is to only use the `targetModule` path, and make a frontend/codegen contract on the tt-xla side that the forward function input tensors will always come as host, row-major. 
### What's changed
This PR refactors the `targetModule` path to work accordingly by: 
- fixing `TTNNTuplifyTensors` pass to perform forced tuplification (creating an empty tuple when input is empty) only on the forward function (this was previously applied to all funcOps, e.g., consteval funcs, and has caused errors in the emitted code)
- refactor `TTNNPrepareModuleForExport` to ensure that forward function input tensors are in the right form (host/device, layout, dtype) and that IR contains the true layout info

With these changes, `TTNNCreateMainForTest` pass can now be safely removed, provided that `targetModule` is set to be used on the frontend side. 
Workflows to prove the test parity: 
[Before (TTNNCreateMainForTest pass)](https://github.com/tenstorrent/tt-xla/actions/runs/25662480589)
[Now (targetModule path)](https://github.com/tenstorrent/tt-xla/actions/runs/26105781326)
### Checklist

### DIFF
--- a/include/ttmlir/Dialect/TTNN/Transforms/Passes.td
+++ b/include/ttmlir/Dialect/TTNN/Transforms/Passes.td
@@ -263,18 +263,20 @@ def TTNNTuplifyTensors: Pass<"ttnn-tuplify-tensors", "::mlir::ModuleOp"> {
     Option<"tuplifyInputIfEmpty",
             "tuplify-input-if-empty",
             "bool", /*default=*/"false",
-            "If input arg is originally empty, force create an empty tuple arg. This is useful for scenarios where "
-            "signatures are required have tuples on the input, like the EmitC dylib path.">
+            "If forward function input arg is originally empty, force create an empty tuple arg. This is useful for scenarios where "
+            "signatures are required to have tuples on the input, like the EmitC dylib or EmitPy targetModule path.">
   ];
 }
 
 def TTNNPrepareModuleForExport : Pass<"ttnn-prepare-module-for-export", "::mlir::ModuleOp"> {
   let summary = "Prepare module for Python module export.";
   let description = [{
-    This pass prepares the module for export as a Python module by:
-    1. Renaming the first public (non-private, non-const-eval) function to "forward"
+    Reshapes the first forward device function into the canonical entry-point
+    signature consumed by the frontends by:
+    1. Renaming the first forward device function to "forward"
     2. Adding a device argument to the function signature
     3. Replacing all ttnn.get_device ops with the device function argument
+    4. Preparing the forward function input tensors to be in the right form (host/device, layout, dtype)
 
     Given a forward function like this:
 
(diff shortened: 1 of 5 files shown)


lib/Dialect/TTNN/Transforms/Passes.cpp
```diff
@@ -1214,10 +1214,13 @@ class TTNNTuplifyTensors
...
+      // If `tuplifyInputIfEmpty` option is set and the function is a forward
+      // device function, tuplify the input even if the
       // function has no inputs.
       //
-      if ((tuplifyInputIfEmpty || !functionType.getInputs().empty()) &&
+      if (((tuplifyInputIfEmpty &&
+            ttmlir::utils::isForwardDeviceFunc(funcOp)) ||
+           !functionType.getInputs().empty()) &&
           llvm::all_of(functionType.getInputs(),
                        [](Type t) { return mlir::isa<RankedTensorType>(t); })) {
```
svuckovicTT: This is a little hard to parse, consider making it more readable.

If I'm not mistaken, there's an extra set of redundant parentheses.

lib/Dialect/TTNN/Transforms/Passes.cpp
```diff
@@ -1372,46 +1375,198 @@ class TTNNPrepareModuleForExport
...
+    for (auto getDeviceOp : getDeviceOps) {
       rewriter.replaceOp(getDeviceOp, deviceArg);
     }
+    return success();
+  }
+
+  // Re-type every forward function input tensor to the appropriate form
+  // (host/device, layout, dtype).
+  //
+  void prepareForwardFuncInputTensors(IRRewriter &rewriter,
```
svuckovicTT, line 1452: Add a .mlir test that checks this piece of code executed properly.

include/ttmlir/Dialect/TTNN/Transforms/Passes.td
```diff
@@ -263,18 +263,20 @@ def TTNNTuplifyTensors: Pass<"ttnn-tuplify-tensors", "::mlir::ModuleOp"> {
     Option<"tuplifyInputIfEmpty",
             "tuplify-input-if-empty",
             "bool", /*default=*/"false",
-            "If input arg is originally empty, force create an empty tuple arg. This is useful for scenarios where "
-            "signatures are required have tuples on the input, like the EmitC dylib path.">
+            "If forward function input arg is originally empty, force create an empty tuple arg. This is useful for scenarios where "
```
svuckovicTT, line 266: I think this `tuplifyInputIfEmpty` option was added to avoid issues with some mlir tests, but I'm not sure if it's worth it anymore - have you tried removing it to see if anything would fail?

svuckovicTT (approved):

include/ttmlir/Dialect/TTNN/Transforms/Passes.td
amilovanovicTT: This is still needed because of this
```
PythonModelRunner::forward(const std::vector<tt::runtime::Tensor> &inputs, tt::runtime::Device device)
``` 
which is used by `SOLoadedExecutableInstance::execute`.

svuckovicTT: Right, I meant to say - can we remove the option, but make the pass that used it behave as it was set to true?

svuckovicTT: Anyhow, that could be done in a follow-up change.

Issue #7656, opened 2026-03-25
[D2M] Resolve precision issues when aggressively blocking K dimension for matmuls


anuragsingh-tt:
### Is your feature request related to a problem? Please describe

For certain large matmul configurations, allocator `auto` increases the reduction dimension block factor beyond the pre-existing choice. This shrinks the local per-core reduction panel and increases the number of partial accumulations. 

In low-precision configurations and when L1 accumulation is enabled, this changes the rounding/accumulation behavior enough to cause measurable PCC degradation (e.g. 0.93 < 0.99). 

As a temporary mitigation, only the affected pytests are forced to use `test-buffer-size-policy=max` which preserves the larger local K panel. 

### Describe the solution you'd like

Use `fp32_dest_acc_en=true` for these matmuls instead. But this trades off DST capacity and may not fully compensate. So some experimentation will be required.

RustamKazmi:
Hi, I’m interested in working on this.

I’ll start by reproducing the precision degradation and evaluating the impact of K-dimension blocking and FP32 accumulation.

I have access to the relevant hardware, so I should be able to validate changes as well.

Please let me know if this is already being worked on or if I can start on it.

PR #8387, opened 2026-05-12
[D2M] enable l1 acc by default

### LINKED ISSUE #8366: [D2M] Make `enable-l1-acc` the default behavior
### Problem

The `enable-l1-acc` pipeline option in `D2MPipelineOptions` (and the `enableL1Acc` `Option<>` on both `D2MInsertDstRegisterAccess{Unscheduled,Scheduled}` passes) is **`false` by default**. Two things follow from this:

1. **It is opt-in, not on by default.** Every production caller that wants the working matmul path has to remember to pass `enable-l1-acc=true`, and silicon perf / accuracy reports against the pipeline's defaults reflect the *non-production* path.
2. **There is no profitability analysis driving the decision.** The compiler doesn't choose L1 acc per-loop based on the reduction dimension size, data format, memory pressure, or DST contention. The flag is a static toggle a user flips; the only compile-time logic on top is the `hasTileMatmul(loopOp)` gate that turns the feature off for non-matmul loops.

When `enable-l1-acc=false`, every matmul iteration reloads the running accumulator from L1 into a DST slice, computes the next partial, and stores it back. When `true`, the pass inserts `d2m.set_l1_accumulate` so the packer accumulates partials directly in L1 and the per-iteration CB→DST reload nest is skipped.

L1 accumulation was added in PR #7026 specifically because **TTNN's matmul kernel already does this** and the D2M path was losing perf relative to TTNN's matmul. The PR also documents that *not* doing L1 accumulation introduces visible precision loss for large bf16 matmuls.

### Evidence that the current default is the wrong one

In `test/python/golden/d2m/test_matmul.py` the bf16 matmul tests `xfail` when `enable_l1_acc=False`:

```python
if (
    dtype == torch.bfloat16
    and not enable_l1_acc
    and shape in ((2048, 2048, 2048), (1024, 2048, 2048))
):
    pytest.xfail(reason="bf16 PCC below threshold for these shapes")
```

In other words: **L1 acc isn't just a perf knob — it's a correctness requirement** for large bf16 shapes. Today the project ships with the *failing* default and relies on each caller to opt-in to the working mode.

The flag is also gated internally by `hasTileMatmul(loopOp)`:

```cpp
bool packerL1Acc = enableL1Acc && hasTileMatmul(loopOp);
```

so flipping it has **no effect on eltwise / reduction / non-matmul paths**. Only matmul-containing regions are affected.

### Requirements

The proposal has two steps; (1) is a low-risk default flip, (2) is a larger cleanup once the remaining caveats are resolved.

#### Step 1: flip the default to `true`

- [ ] Change the default of `enableL1Acc` in:
  - `D2MPipelineOptions::enableL1Acc` (`include/ttmlir/Dialect/D2M/Pipelines/D2MPipelines.h`),
  - `D2MInsertDstRegisterAccessUnscheduled` `enableL1Acc` option (`include/ttmlir/Dialect/D2M/Transforms/Passes.td`),
  - `D2MInsertDstRegisterAccessScheduled` `enableL1Acc` option (`include/ttmlir/Dialect/D2M/Transforms/Passes.td`),
  - all downstream consumers (`ttnn-jit`, `ttrt`, tt-xla) — search for `enable-l1-acc=` / `enableL1Acc =` to confirm explicit users already set the value they want.
- [ ] Remove the `xfail` markers in `test/python/golden/d2m/test_matmul.py` for the bf16 shapes that previously needed opt-in; tighten their PCC thresholds.
- [ ] Update / collapse the `enable-l1-acc=...` parametrisations in the python golden tests — those that pass `False` to exercise the no-L1-acc path become an explicit "regression for the escape-hatch path" test (or get deleted if Step 2 lands first).
- [ ] Verify on silicon for the matmul shapes currently in `check-perf` and the model perf CI; expected outcome is "same or better" perf and **better** PCC.
- [ ] Track the known precision interaction with the `auto` allocator policy ([#7656](https://github.com/tenstorrent/tt-mlir/issues/7656)): this is independent of the default flip but is exposed once the flip happens, so the issue needs the workaround (`test-buffer-size-policy=max`) to land first or to be applied in the pipeline default.

#### Step 2 (after Step 1 has soaked): drop the flag entirely

- [ ] Remove `enableL1Acc` from all three option lists.
- [ ] Drop the `if (enableL1Acc)` branches in `Shared.cpp::insertDstRegisterAccessFinalize`, `Unscheduled.cpp::dataCopyGenerate`, `Scheduled.cpp::dataCopyGenerateScheduledAll`, and `Shared.cpp::emitDstCopyNest`.
- [ ] Replace `packerL1Acc = enableL1Acc && hasTileMatmul(loopOp)` with just `hasTileMatmul(loopOp)`.
- [ ] Update `test/ttmlir/Dialect/D2M/Transforms/insert_dst_register_access_l1_acc.mlir`'s RUN line to drop the `="enable-l1-acc=true"` option.
- [ ] Delete the `*-l1-acc` parametrisations in the golden tests that only existed to flip the flag.

### Outcomes

- The default project behavior matches what every production caller already opts into.
- The matmul perf-vs-TTNN parity work from PR #7026 lands as the baseline, not the opt-in.
- Once Step 2 lands, the only L1-acc-related branch left in the code is the `hasTileMatmul` gate, which is part of the contract of the matmul lowering — not a tuning knob.

### Risks / open questions

- The TODO in `test_matmul.py::get_allocator_policy_override` plus [#7656](https://github.com/tenstorrent/tt-mlir/issues/7656) document a separate precision regression involving `auto` allocator policy. Step 1 should not regress anything that's working today (it's a default flip; callers can still pass `enable-l1-acc=false`), but Step 2 needs that issue resolved first.
- Per PR #7026's notes there is a documented tt-metal FP32 bug behind L1 accumulation ([L1_ACCUMULATION_FP32_ANALYSIS.md](https://github.com/tenstorrent/tt-metal/blob/main/docs/L1_ACCUMULATION_FP32_ANALYSIS.md)). TTNN works around it with one final reload; D2M may need an equivalent before Step 2 can drop the escape hatch. (This is a patch-on-top-of-L1-acc decision, not a "use-vs-don't-use" decision.)

### Why we do *not* propose a per-loop profitability heuristic

A natural-sounding alternative is "make the compiler decide whether to enable L1 acc per-loop based on reduction dim size / dtype / memory pressure". For this codebase, **today, that's premature complexity**:

- **The test suite gives no signal in that direction.** No test in `test/python/golden/d2m/test_matmul.py` (or anywhere else) `xfail`s with `enable_l1_acc=True` and passes with `False`. The evidence is strictly one-way: L1 acc either matches or strictly fixes the no-L1-acc case.
- **The theoretical "bad cases" don't pencil out either.** Tiny matmuls / `K==1` reductions still see the `scf.if iter != 0` guard short-circuit and the `set_l1_accumulate` op hoist out of the loop, so the worst-case overhead is one extra arith op per program — not a regression worth gating.
- **The compiler already makes the only decision that matters.** L1 acc is gated by `hasTileMatmul(loopOp)`, so eltwise / reduction / non-matmul paths are never affected. That's a simple but correct per-loop predicate; profitability analysis would refine the *same* predicate without changing the answer for any known workload.
- **Dtype-specific safety (e.g., FP32 final reload) is its own feature, not a profitability decision.** When that lands it should live next to `hasTileMatmul` inside the L1-acc lowering, not as a user-visible flag.

If real workload data later surfaces a *regression* (a case where L1 acc is genuinely a perf or accuracy loss), this issue would be amended to add a per-loop predicate alongside `hasTileMatmul`, e.g., `shouldUseL1Acc(loopOp) = hasTileMatmul(loopOp) && !<regression condition>`. Until then, the simpler change (Step 1 → Step 2 above) is both correct and what the codebase's evidence supports.

### References

- PR #7919 conversation: https://github.com/tenstorrent/tt-mlir/pull/7919
- L1-acc was introduced in PR #7026 ("[D2M] enable packer L1 accumulation for matmul").
- Existing xfails: `test/python/golden/d2m/test_matmul.py` lines 155-164 (single-buffered) and 227-236 (double-buffered).
- Related precision issue: https://github.com/tenstorrent/tt-mlir/issues/7656.

sgholamiTT:
### Ticket
https://github.com/tenstorrent/tt-mlir/issues/8366

### Problem description
l1-accumulation was turned off by default in d2m.

### What's changed
1. **Enable L1-accumulation by default in D2M** (the main goal of this PR). The user-facing knob is now `disable-l1-acc` (default `false`); to opt out, pass `disable-l1-acc=true`. The previous `enable-l1-acc` option (default `false`) has been removed; `disableL1Acc` is now the single source of truth across pass/pipeline options, internal C++ (`InsertDstRegisterAccess{Shared,Unscheduled,Scheduled}.{h,cpp}`), MLIR lit tests, and Python golden tests.
2. **Correctness fixes required to safely enable L1-acc by default** (gated in `InsertDstRegisterAccess{Unscheduled,Scheduled}`):
   - **Skip L1-acc when the matmul output element type is not in the packer's L1-acc supported set.** Per tt-llk's `PACK_L1_ACC_FORMATS`, only `Float32 / Float16 / BFloat16 / Int32 / UInt8` are supported. Block-float outputs (e.g. `bfp_bf8`) would silently produce garbage results (PCC ~0).
   - **Trigger the L1-acc guard on the outermost reduction loop, not the outermost ancestor loop.** A "reduction" loop here is an ancestor loop whose IV no recorded output store depends on. The previous unconditional outermost-ancestor behavior was wrong for batched matmuls — the outermost loop is the parallel batch dimension, so turning on L1-acc on the second batch iteration would accumulate into uninitialized L1 at a different output address (e.g. PCC ~0.31 for `test_matmul_3d_single_core[batch2_small]`).
   - **Skip L1-acc when the candidate reduction loop has constant trip count ≤ 1** (accumulation is a no-op anyway).
3. **Test updates:**
   - Restored `pytest.xfail` for `bf16 + no_l1_acc + (1024|2048)x2048x2048` matmul tests, with a comment linking to [#7656](https://github.com/tenstorrent/tt-mlir/issues/7656). These cases test the no-L1-acc path which has fundamental bf16 precision limitations from cumulative round-trips through the partial-sum CB; the proper fix (e.g. `fp32_dest_acc_en` or an fp32 partial-sum CB) is tracked in #7656.
   - Updated `insert_dst_register_access_l1_acc.mlir` to validate (a) the corrected L1-acc trigger on a real outer K-block reduction loop, and (b) that L1-acc is correctly **not** inserted when only parallel ancestor loops exist.

### Why no profitability analysis ?
The selective gating above is **correctness-based**, not profitability-based — it disables L1-acc only in cases where the packer cannot honor it or the trigger loop is not actually a reduction. We do **not** gate L1-acc on a profitability heuristic: based on the following evidence, matmul kernels almost always benefit from it.

**Evidence 1:**
image

**Evidence 2:**
This PR should also have positive effect on the [ttnn-jit dashboard](https://superset.tenstorrent.com/superset/dashboard/479/).  

| shape (M×K×N)   | dtype   | mem              | math fid. | L1 ON µs | L1 OFF µs | speedup |
|-----------------|---------|------------------|-----------|---------:|----------:|--------:|
| 2048×2048×2048  | bf16    | dram_interleaved | HiFi4     |    325.1 |     523.9 | 1.61x |
| 2048×2048×2048  | bf16    | l1_dram          | HiFi4     |    323.4 |     522.2 | 1.61x |
| 2048×2048×2048  | bfp8_b  | dram_interleaved | HiFi2     |    180.8 |     339.7 | 1.88x |
| 2048×2048×2048  | bfp8_b  | l1_dram          | HiFi2     |    180.2 |     336.8 | 1.87x |
| 2048×2048×1024  | bf16    | dram_interleaved | HiFi4     |    213.4 |     289.0 | 1.35x   |
| 2048×2048×1024  | bf16    | l1_dram          | HiFi4     |    180.0 |     285.7 | 1.59x   |
| 2048×2048×1024  | bfp8_b  | dram_interleaved | HiFi2     |    122.9 |     197.0 | 1.60x   |
| 2048×2048×1024  | bfp8_b  | l1_dram          | HiFi2     |    103.9 |     196.0 | 1.89x |
| 1024×2048×2048  | bf16    | dram_interleaved | HiFi4     |    179.6 |     281.0 | 1.56x   |
| 1024×2048×2048  | bf16    | l1_dram          | HiFi4     |    179.7 |     279.9 | 1.56x   |
| 1024×2048×2048  | bfp8_b  | dram_interleaved | HiFi2     |    104.2 |     186.8 | 1.79x   |
| 1024×2048×2048  | bfp8_b  | l1_dram          | HiFi2     |    103.9 |     186.2 | 1.79x   |
| 2048×1024×1024  | bf16    | dram_interleaved | HiFi4     |    111.2 |     147.7 | 1.33x   |
| 2048×1024×1024  | bf16    | l1_dram          | HiFi4     |     91.5 |     143.1 | 1.56x   |
| 2048×1024×1024  | bfp8_b  | dram_interleaved | HiFi2     |     62.5 |     100.3 | 1.60x   |
| 2048×1024×1024  | bfp8_b  | l1_dram          | HiFi2     |     52.8 |      98.4 | 1.86x   |
| 1024×1024×2048  | bf16    | dram_interleaved | HiFi4     |     91.8 |     141.0 | 1.54x   |
| 1024×1024×2048  | bf16    | l1_dram          | HiFi4     |     91.4 |     140.2 | 1.53x   |
| 1024×1024×2048  | bfp8_b  | dram_interleaved | HiFi2     |     53.5 |      94.2 | 1.76x   |
| 1024×1024×2048  | bfp8_b  | l1_dram          | HiFi2     |     53.1 |      93.6 | 1.76x   |
| 1024×1024×1024  | bf16    | dram_interleaved | HiFi4     |     58.3 |      83.2 | 1.43x   |
| 1024×1024×1024  | bf16    | l1_dram          | HiFi4     |     55.8 |      82.5 | 1.48x   |
| 1024×1024×1024  | bfp8_b  | dram_interleaved | HiFi2     |     35.4 |      57.8 | 1.63x   |
| 1024×1024×1024  | bfp8_b  | l1_dram          | HiFi2     |     34.8 |      57.9 | 1.67x   |
| 2048×1024×512   | bf16    | dram_interleaved | HiFi4     |    109.0 |     108.9 | 1.00x   |
| 2048×1024×512   | bf16    | l1_dram          | HiFi4     |     55.6 |      62.3 | 1.12x   |
| 2048×1024×512   | bfp8_b  | dram_interleaved | HiFi2     |     62.7 |      62.8 | 1.00x   |
| 2048×1024×512   | bfp8_b  | l1_dram          | HiFi2     |     33.5 |      40.5 | 1.21x   |
| 1024×1024×512   | bf16    | dram_interleaved | HiFi4     |     54.5 |      55.0 | 1.01x   |
| 1024×1024×512   | bf16    | l1_dram          | HiFi4     |     29.4 |      36.2 | 1.23x   |
| 1024×1024×512   | bfp8_b  | dram_interleaved | HiFi2     |     30.7 |      32.7 | 1.07x   |
| 1024×1024×512   | bfp8_b  | l1_dram          | HiFi2     |     20.2 |      24.7 | 1.22x   |
| 512×1024×2048   | bf16    | dram_interleaved | HiFi4     |     56.3 |      60.9 | 1.08x   |
| 512×1024×2048   | bf16    | l1_dram          | HiFi4     |     56.2 |      60.5 | 1.08x   |
| 512×1024×2048   | bfp8_b  | dram_interleaved | HiFi2     |     35.5 |      38.5 | 1.08x   |
| 512×1024×2048   | bfp8_b  | l1_dram          | HiFi2     |     34.8 |      38.3 | 1.10x   |
| 512×1024×1024   | bf16    | dram_interleaved | HiFi4     |     30.7 |      35.2 | 1.15x   |
| 512×1024×1024   | bf16    | l1_dram          | HiFi4     |     29.4 |      35.0 | 1.19x   |
| 512×1024×1024   | bfp8_b  | dram_interleaved | HiFi2     |     22.4 |      23.4 | 1.05x   |
| 512×1024×1024   | bfp8_b  | l1_dram          | HiFi2     |     21.5 |      23.1 | 1.07x   |
| 512×512×512     | bf16    | dram_interleaved | HiFi4     |     11.8 |      13.4 | 1.14x   |
| 512×512×512     | bf16    | l1_dram          | HiFi4     |     11.0 |      12.2 | 1.11x   |
| 512×512×512     | bfp8_b  | dram_interleaved | HiFi2     |      9.0 |       9.5 | 1.05x   |
| 512×512×512     | bfp8_b  | l1_dram          | HiFi2     |      8.3 |       8.8 | 1.05x   |

### Checklist
- [x] [Nightly](https://github.com/tenstorrent/tt-mlir/actions/runs/25811793178)

### DIFF
--- a/include/ttmlir/Dialect/D2M/Pipelines/D2MPipelines.h
+++ b/include/ttmlir/Dialect/D2M/Pipelines/D2MPipelines.h
@@ -210,7 +210,7 @@ struct D2MPipelineOptions : public PassPipelineOptions<D2MPipelineOptions> {
 
   Option<bool> enableL1Acc{*this, "enable-l1-acc",
                            llvm::cl::desc("Enable L1 accumulation."),
-                           llvm::cl::init(false)};
+                           llvm::cl::init(true)};
 };
 
 void createTTIRBufferizationPipeline(OpPassManager &pm,
--- a/include/ttmlir/Dialect/D2M/Pipelines/D2MPipelines.h
+++ b/include/ttmlir/Dialect/D2M/Pipelines/D2MPipelines.h
@@ -208,9 +208,11 @@ struct D2MPipelineOptions : public PassPipelineOptions<D2MPipelineOptions> {
                      "prints debug output comparing them."),
       llvm::cl::init(false)};
 
-  Option<bool> enableL1Acc{*this, "enable-l1-acc",
-                           llvm::cl::desc("Enable L1 accumulation."),
-                           llvm::cl::init(true)};
+  Option<bool> disableL1Acc{
+      *this, "disable-l1-acc",
+      llvm::cl::desc("Disable L1 accumulation (force reloading from L1 to DST "
+                     "instead of accumulating partials in L1)."),
+      llvm::cl::init(false)};
 };
 
 void createTTIRBufferizationPipeline(OpPassManager &pm,
--- a/include/ttmlir/Dialect/D2M/Transforms/InsertDstRegisterAccessShared.h
+++ b/include/ttmlir/Dialect/D2M/Transforms/InsertDstRegisterAccessShared.h
@@ -118,6 +118,33 @@ bool hasTileMatmul(Operation *op);
 // True iff `op` is any tile-level reduction op (FPU or SFPU variant).
 bool isTileReductionOp(Operation *op);
 
+// Returns true iff the packer hardware supports L1 accumulation when packing
+// into an output buffer of element type `dt`. The packer L1-acc path goes
+// through a fixed set of native formats; block-float (Bfp*) outputs are NOT
+// supported and would silently produce garbage results.
+//
+// Supported set follows tt-llk `PACK_L1_ACC_FORMATS`:
+//   Float32, Float16, BFloat16 (Float16_b), Int32, UInt8.
+bool isPackerL1AccumulationSupportedDataType(ttcore::DataType dt);
+
+// Returns true iff every `d2m.tile_matmul` in `loopOp` has a result tile
+// element type that is supported by the packer L1-accumulation path. Returns
+// true if there are no `d2m.tile_matmul` ops (caller is responsible for
+// gating on `hasTileMatmul`).
+bool allTileMatmulOutputsSupportPackerL1Acc(Operation *loopOp);
+
+// Find the outermost ancestor reduction loop IV of `acquireDst`, where
+// "reduction" means: no output store recorded in `copyInfos` depends on the
+// loop's induction variable. This is the loop whose iterations accumulate
+// into the same output L1 slot, and is therefore the correct trigger for
+// switching the packer to L1-acc mode.
+//
+// Returns nullptr if there is no such reduction loop, or if the candidate
+// reduction loop has a constant trip count <= 1 (in which case L1-acc is
+// not needed and the trigger comparison would never fire correctly).
+Value findOutermostReductionLoopIVForL1Acc(Operation *acquireDstOp,
+                                           const CopyInfoMap &copyInfos);
+
 // Stamp a pass-allocated scratch slice onto the op's `dst_scratch_index`
 // attribute for the TTKernel lowering to consume.  Today only supports ops
 // that need exactly one scratch slice.
--- a/include/ttmlir/Dialect/D2M/Transforms/InsertDstRegisterAccessShared.h
+++ b/include/ttmlir/Dialect/D2M/Transforms/InsertDstRegisterAccessShared.h
@@ -230,8 +230,8 @@ cloneAffineLoopSkeleton(PatternRewriter &rewriter, Operation *loopNestOrOp);
 //   2. Rewrites the original load/store via `accessReplacer` so it now
 //      goes through the DST register.
 //
-// `enableL1Acc=true` skips the upfront copy nest entirely (the L1 acc
-// guard preserves the running tile).
+// `disableL1Acc=false` (i.e. L1 acc on) skips the upfront copy nest
+// entirely (the L1 acc guard preserves the running tile).
 template <typename LoadOrStoreTy>
 void emitDstCopyNest(
     PatternRewriter &rewriter, Operation *loopNestOrOp,
@@ -242,7 +242,7 @@ void emitDstCopyNest(
     llvm::function_ref<void(PatternRewriter &, LoadStoreRecord<LoadOrStoreTy>,
                             AffineMap, ValueRange)>
         accessReplacer,
-    bool enableL1Acc = false);
+    bool disableL1Acc = true);
 
 std::pair<AffineMap, SmallVector<Value>>
 buildLinearizedDstAccess(PatternRewriter &rewriter, Operation *op, int dstSlice,
@@ -277,7 +277,7 @@ void insertPackerL1AccGuard(PatternRewriter &rewriter, Location loc,
 bool insertDstRegisterAccessFinalize(
     PatternRewriter &rewriter, GenericOp gOp, Region &region,
     unsigned dstCapacity, Operation *outermostInnerComputeLoop,
-    bool enableL1Acc, CopyInfoMap &copyInfos,
+    bool disableL1Acc, CopyInfoMap &copyInfos,
     DstIntermediatesMap &dstIntermediates,
     llvm::function_ref<void(PatternRewriter &, Location, Value,
                             const CopyInfoMap &, bool)>
--- a/include/ttmlir/Dialect/D2M/Transforms/Passes.td
+++ b/include/ttmlir/Dialect/D2M/Transforms/Passes.td
@@ -326,7 +326,7 @@ def D2MInsertDstRegisterAccessUnscheduled : Pass<"d2m-insert-dst-register-access
   ];
   let options = [
     Option<"maxDstPhysicalSizeTiles", "max-dst-physical-size-tiles", "unsigned", "0", "Override the max dst physical size tiles or 0 if unset.">,
-    Option<"enableL1Acc", "enable-l1-acc", "bool", /*default=*/"false", "Enable accumulating partials in L1 instead of reloading from L1 to DST.">,
+    Option<"enableL1Acc", "enable-l1-acc", "bool", /*default=*/"true", "Enable accumulating partials in L1 instead of reloading from L1 to DST.">,
   ];
 }
 
@@ -401,7 +401,7 @@ def D2MInsertDstRegisterAccessScheduled : Pass<"d2m-insert-dst-register-access-s
   ];
   let options = [
     Option<"maxDstPhysicalSizeTiles", "max-dst-physical-size-tiles", "unsigned", "0", "Override the max dst physical size tiles or 0 if unset.">,
-    Option<"enableL1Acc", "enable-l1-acc", "bool", /*default=*/"false", "Enable accumulating partials in L1 instead of reloading from L1 to DST.">,
+    Option<"enableL1Acc", "enable-l1-acc", "bool", /*default=*/"true", "Enable accumulating partials in L1 instead of reloading from L1 to DST.">,
   ];
 }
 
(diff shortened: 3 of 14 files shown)


anuragsingh-tt (approved):
Great data substantiating the change! Let's do it.

dlokeTT (approved):

xanderchin (approved):

PR #8394, opened 2026-05-12
[d2m] Infer unified-thread synchronized region from remote ops

nsmithtt:
In SplitUnifiedThread, the synchronized region wrapping compute was anchored on scf.for ops tagged with `d2m.blocking_loop`. This forced upstream passes to plant that attribute on the right loop, which is fragile and couples synchronization scope to a marker that doesn't mean anything on its own.

Infer the scope instead: collect the blocks that contain remote_load/remote_store ops at their immediate level, and walk up from each compute op until its block is one of those. The loads and stores that bound a compute region are exactly what defines that region, so this drops the tag dependency without changing the resulting IR shape.


lib/Dialect/D2M/Transforms/SplitUnifiedThread.cpp
```diff
@@ -112,21 +112,30 @@ LogicalResult wrapComputeInSynchronizedRegion(GenericOp genericOp,
...
+  // Collect the blocks that contain remote_load/store ops at their immediate
+  // level. These delimit the scope where compute is synchronized: the
+  // outermost compute ancestor sits in such a block alongside the loads/stores
+  // that bound it.
+  DenseSet<Block *> blocksWithRemoteOps;
+  genericOp.getRegion(0).walk([&](Operation *op) {
+    if (isa<RemoteLoadOp, RemoteStoreOp>(op)) {
+      blocksWithRemoteOps.insert(op->getBlock());
+    }
+  });
```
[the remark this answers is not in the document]
arminaleTT: Please address this by adding/updating the test

arminaleTT (approved):

lib/Dialect/D2M/Transforms/SplitUnifiedThread.cpp
```diff
@@ -112,21 +112,30 @@ LogicalResult wrapComputeInSynchronizedRegion(GenericOp genericOp,
...
   // they haven't been lowered yet into non-synchronized ops
   OpBuilder::InsertionGuard guard(rewriter);
+
+  // Collect the blocks that contain remote_load/store ops at their immediate
+  // level. These delimit the scope where compute is synchronized: the
+  // outermost compute ancestor sits in such a block alongside the loads/stores
+  // that bound it.
+  DenseSet<Block *> blocksWithRemoteOps;
+  genericOp.getRegion(0).walk([&](Operation *op) {
+    if (isa<RemoteLoadOp, RemoteStoreOp>(op)) {
```
nsmithtt: Instead of only matching remote load/store, we should collect any op with trait `SynchronizableOpInterface`.  The comment and variable name should be updated accordingly.

lib/Dialect/D2M/Transforms/SplitUnifiedThread.cpp
```diff
@@ -112,21 +112,30 @@ LogicalResult wrapComputeInSynchronizedRegion(GenericOp genericOp,
...
   OpBuilder::InsertionGuard guard(rewriter);
+
+  // Collect the blocks that contain SynchronizableOpInterface ops at their
+  // immediate level. These delimit the scope where compute is synchronized:
+  // the outermost compute ancestor sits in such a block alongside the
+  // synchronizable ops that bound it.
+  DenseSet<Block *> blocksWithSynchronizableOps;
+  genericOp.getRegion(0).walk([&](Operation *op) {
+    if (dyn_cast<SynchronizableOpInterface>(op)) {
+      blocksWithSynchronizableOps.insert(op->getBlock());
```
sohaibnadeemTT: nit: can we collect the parent op instead of the block for consistency with the rest of the fucntion?

lib/Dialect/D2M/Transforms/SplitUnifiedThread.cpp
```diff
@@ -112,21 +112,30 @@ LogicalResult wrapComputeInSynchronizedRegion(GenericOp genericOp,
...
+    }
+  });
+
```
sohaibnadeemTT, line 128: can we add an assert here to ensure that blocksWithSynchronizableOps is of size 1 (i.e blocking scope is not ambiguou)

sohaibnadeemTT (approved):
minor comments otherwise lgtm

sohaibnadeemTT:
@nsmithtt PR title/comment should be updated too after change to use any SyncrhonizableOpInterface op from our discussion

PR #8409, opened 2026-05-13
d2m: beam search on eltwise op reblocking to restrict search space

anuragsingh-tt:
New policy for eltwise op reblocking (bounded) now prioritizes moderate reblocking (best tradeoff between latency/memory) and performs beam search on a restricted search space set (top 64 candidates for now). Matmul policy remains as is for now (only one dimension is reblocked at this time--reduction--so the search space is quite limited). 

Follow up work is to enable reblocking for different ops at which point the names (AllParallel, Reduction) will change to better reflect true distinction. 
 
image

Edit:
The timing data on some full model compilations for the Allocate pass doesn't suggest a need to restrict the current search space right now, so I will keep the exhaustive search as the default policy. What is justified is a bias towards "moderate" reblocking which the "bounded" policy does pick. Keeping the flag for now but will re-orient cost model so moderate factors are biased even for full search.

### DIFF
--- a/lib/Dialect/D2M/Analysis/BlockFactorAnalysis.cpp
+++ b/lib/Dialect/D2M/Analysis/BlockFactorAnalysis.cpp
@@ -517,11 +517,11 @@ static SmallVector<int64_t> chooseReblockedFactors(
   case BlockFactorAnalysis::BufferSizePolicy::Auto:
     return applyAutoPolicy(genericOp, indexingMaps, iteratorTypes, gridExtents,
                            shardExtents, shardFactors, device, l1Attr,
-                           numBuffers, /*useBoundedEltwiseSearch=*/true);
-  case BlockFactorAnalysis::BufferSizePolicy::Exhaustive:
+                           numBuffers, /*useBoundedEltwiseSearch=*/false);
+  case BlockFactorAnalysis::BufferSizePolicy::Bounded:
     return applyAutoPolicy(genericOp, indexingMaps, iteratorTypes, gridExtents,
                            shardExtents, shardFactors, device, l1Attr,
-                           numBuffers, /*useBoundedEltwiseSearch=*/false);
+                           numBuffers, /*useBoundedEltwiseSearch=*/true);
   }
 
   llvm_unreachable("unknown buffer size policy");
--- a/lib/Dialect/D2M/Analysis/BlockFactorAnalysis.cpp
+++ b/lib/Dialect/D2M/Analysis/BlockFactorAnalysis.cpp
@@ -40,6 +40,7 @@ struct CandidateScore {
   uint64_t cbBytes = 0;
 };
 
+// Beam-search state before a full candidate legality and CB cost evaluation.
 struct PartialCandidate {
   SmallVector<int64_t> dimScales;
   uint64_t blockingVolume = 1;
@@ -154,7 +155,7 @@ static SmallVector<int64_t> buildEltwiseScaleSearchOrder(int64_t shardFactor) {
   });
 
   SmallVector<int64_t> limited;
-  // Always include 1 in the search order.
+  // Always include scale factor 1 so bounded search can leave a dim unchanged.
   limited.reserve(std::min(factors.size(), kEltwiseScalesPerDim));
   for (int64_t factor : factors) {
     if (limited.size() == kEltwiseScalesPerDim) {
@@ -372,7 +373,7 @@ applyAutoPolicy(GenericOp genericOp, ArrayRef<AffineMap> indexingMaps,
                 ArrayRef<int64_t> gridExtents, ArrayRef<int64_t> shardExtents,
                 ArrayRef<int64_t> shardFactors, ttcore::DeviceAttr device,
                 ttcore::MemorySpaceAttr l1Attr, uint32_t numBuffers,
-                bool useBoundedEltwiseSearch = true) {
+                bool useBoundedEltwiseSearch) {
   const SmallVector<int64_t> originalBlockFactors =
       genericOp.getBlockFactorsValue();
 
@@ -420,9 +421,7 @@ applyAutoPolicy(GenericOp genericOp, ArrayRef<AffineMap> indexingMaps,
     beam.push_back(
         PartialCandidate{SmallVector<int64_t>(currentDimScales), 1, 0});
 
-    for (std::size_t dimIndex = 0; dimIndex < config->candidateDims.size();
-         ++dimIndex) {
-      const std::size_t dim = config->candidateDims[dimIndex];
+    for (auto [dimIndex, dim] : llvm::enumerate(config->candidateDims)) {
       SmallVector<PartialCandidate> expanded;
       expanded.reserve(beam.size() * legalScales[dimIndex].size());
       for (const PartialCandidate &partial : beam) {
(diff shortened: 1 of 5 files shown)


lib/Dialect/D2M/Analysis/BlockFactorAnalysis.cpp
```diff
@@ -10,6 +10,7 @@
 
 #include "mlir/Dialect/Linalg/IR/Linalg.h"
 #include "mlir/Dialect/MemRef/IR/MemRef.h"
+#include "llvm/ADT/DenseMap.h"
```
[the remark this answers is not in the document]
vwellsTT: This seems like it's way outsized in rating, but good nit haha

lib/Dialect/D2M/Analysis/BlockFactorAnalysis.cpp
```diff
@@ -119,6 +129,67 @@ static bool isLexicographicallyLarger(ArrayRef<int64_t> lhs,
   return false;
 }
 
+static SmallVector<int64_t> buildEltwiseScaleSearchOrder(int64_t shardFactor) {
+  SmallVector<int64_t> factors = ttmlir::utils::getFactors(shardFactor);
+  llvm::SmallDenseMap<int64_t, int64_t, 8> factorRanks;
```
vwellsTT, line 135: super nit: a little weird to use naked SmallVector but llvm:: prefixed SmallDenseMap here vs line above imo

lib/Dialect/D2M/Analysis/BlockFactorAnalysis.cpp
```diff
@@ -39,6 +40,15 @@ struct CandidateScore {
   uint64_t cbBytes = 0;
 };
 
+struct PartialCandidate {
```
vwellsTT, line 44: nit: I think a 1-line comment explaining what this struct is vs a "real" candidate might be helpful

lib/Dialect/D2M/Analysis/BlockFactorAnalysis.cpp
```diff
@@ -119,6 +129,67 @@ static bool isLexicographicallyLarger(ArrayRef<int64_t> lhs,
...
+      return lhsPenalty < rhsPenalty;
+    }
+    if ((lhs == 1) != (rhs == 1)) {
+      return rhs == 1;
+    }
+    return lhs > rhs;
+  });
+
+  SmallVector<int64_t> limited;
+  // Always include 1 in the search order.
```
vwellsTT: I don't quite follow what this comment means, the number 1? 1 of something?

lib/Dialect/D2M/Analysis/BlockFactorAnalysis.cpp
```diff
@@ -300,7 +371,8 @@ applyAutoPolicy(GenericOp genericOp, ArrayRef<AffineMap> indexingMaps,
                 ArrayRef<ttcore::IteratorType> iteratorTypes,
                 ArrayRef<int64_t> gridExtents, ArrayRef<int64_t> shardExtents,
                 ArrayRef<int64_t> shardFactors, ttcore::DeviceAttr device,
-                ttcore::MemorySpaceAttr l1Attr, uint32_t numBuffers) {
+                ttcore::MemorySpaceAttr l1Attr, uint32_t numBuffers,
+                bool useBoundedEltwiseSearch = true) {
```
vwellsTT: does this need a default value? it seems like callers pass in true regardless => better to leave it as required imo

lib/Dialect/D2M/Analysis/BlockFactorAnalysis.cpp
```diff
@@ -333,6 +411,56 @@ applyAutoPolicy(GenericOp genericOp, ArrayRef<AffineMap> indexingMaps,
...
+  // Restrict the large eltwise search space to the top kEltwiseBeamWidth
+  // candidates unless the caller explicitly requests exhaustive search.
+  if (useBoundedEltwiseSearch &&
+      config->shapeClass == AutoShapeClass::AllParallelEltwise &&
+      boundedSearchSpace > kEltwiseBeamWidth) {
+    SmallVector<PartialCandidate> beam;
+    beam.push_back(
+        PartialCandidate{SmallVector<int64_t>(currentDimScales), 1, 0});
+
+    for (std::size_t dimIndex = 0; dimIndex < config->candidateDims.size();
```
vwellsTT: nit: maybe better to use llvm::enumerate?

lib/Dialect/D2M/Analysis/BlockFactorAnalysis.cpp
```diff
@@ -333,6 +411,56 @@ applyAutoPolicy(GenericOp genericOp, ArrayRef<AffineMap> indexingMaps,
...
+        for (auto [scaleRank, scale] : llvm::enumerate(legalScales[dimIndex])) {
+          PartialCandidate candidate = partial;
+          candidate.dimScales[dim] = scale;
+          candidate.blockingVolume *= scale;
+          candidate.moderationPenalty += scaleRank;
+          expanded.push_back(std::move(candidate));
+        }
+      }
+      trimPartialBeam(expanded);
+      beam = std::move(expanded);
```
vwellsTT, line 437: I feel like this logic could be cleaned up s.t. expanded + beam are the same vector, no?

vwellsTT: I guess maybe not because of the cleanup step? hmm

lib/Dialect/D2M/Analysis/BlockFactorAnalysis.cpp
```diff
@@ -333,6 +411,56 @@ applyAutoPolicy(GenericOp genericOp, ArrayRef<AffineMap> indexingMaps,
...
+    }
+
+    for (const PartialCandidate &candidate : beam) {
+      auto score = evaluateCandidate(
```
vwellsTT: nit: sort of violates our "only use auto when type is clear" rule, though I don't mind much for this type of case (in same file)

lib/Dialect/D2M/Analysis/BlockFactorAnalysis.cpp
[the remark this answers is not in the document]
anuragsingh-tt: I guess there is a non-transitive include somewhere.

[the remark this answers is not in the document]
anuragsingh-tt: There is no include for SmallDenseMap. It is in DenseMap.

lib/Dialect/D2M/Analysis/BlockFactorAnalysis.cpp
anuragsingh-tt: I guess I could have "using" in the file but I find it somewhat visually noisy

lib/Dialect/D2M/Analysis/BlockFactorAnalysis.cpp
anuragsingh-tt: It could go either way. I think keeping them separate is a little clearer here, beam is the current frontier and expanded is the full next frontier before trimming. I suppose I could reuse the same vector here but that would be additional bookkeeping.

lib/Dialect/D2M/Analysis/BlockFactorAnalysis.cpp
vwellsTT: are you saying the prefix is required for SmallDenseMap but not SmallVector? hmm, weird

vwellsTT (approved):

PR #8237, opened 2026-05-04
Tile layout concatenate with non-aligned last dim and large 0th dim circular buffer exceeding L1 size workaround

ddilbazTT:
### Ticket
https://github.com/tenstorrent/tt-metal/issues/43371

### Problem description
I am working on 0-day customer model support. The model is encountering an error during the following concatenate operation:
```
            RuntimeTTNN |    DEBUG | Executing operation: %13 = "ttnn.concat"(%12, %4) <{dim = 1 : si32}> : (tensor<1993728x3xsi32, #ttnn.ttnn_layout<(d0, d1) -> (d0, d1), <1x1>, memref<62304x1x!ttcore.tile<32x32, si32>, #ttnn.buffer_type<dram>>, <interleaved>>>, tensor<1993728x1xsi32, #ttnn.ttnn_layout<(d0, d1) -> (d0, d1), <1x1>, memref<62304x1x!ttcore.tile<32x32, si32>, #ttnn.buffer_type<dram>>, <interleaved>>>) -> tensor<1993728x4xsi32, #ttnn.ttnn_layout<(d0, d1) -> (d0, d1), <1x1>, memref<62304x1x!ttcore.tile<32x32, si32>, #ttnn.buffer_type<dram>>, <interleaved>>> loc("concatenate.36")
2026-04-28 20:38:16.627 | critical |          Always | TT_THROW: Statically allocated circular buffers on core range [(x=0,y=0) - (x=0,y=3)] grow to 16056096 B which is beyond max L1 size of 1499136 B (assert.hpp:104)
```

The issue is due to these reasons:
- the last dim of the concatenate ops are not tile aligned: 3 and 1
- in tt-metal, the tensors are converted to row major then the last two dims are permuted. 
    - 1993728x3 --> 3x1993728
    - 1993728x1 --> 1x1993728
- the last dim is used to calculate circular buffer space, which exceeds L1 

### What's changed
Now, the tensors are sliced into chunks which would not exceed L1 size and then concatenated in a concatenate operation with more number of tensors. Assuming that L1_SIZE is the max size a tensor could be supported in concatenate op, then the number of chunked tensors is n = TOTAL_SIZE/L1_SIZE. Now, instead of 1993728x3 tensor, there will be n times (1993728/n)x3  tensors being concatenated. 

### Checklist
- [x] mlir uplift qualifications: https://github.com/tenstorrent/tt-xla/actions/runs/25573466096

PR #8412, opened 2026-05-13
ConcatOpRewritePattern: reserve 20% L1 headroom in chunk-size formula

sonalibaskaran2499:
### Ticket
Follow-up to [#8237](https://github.com/tenstorrent/tt-mlir/pull/8237) 

### Problem description
PR [#8237](https://github.com/tenstorrent/tt-mlir/pull/8237) added the ConcatOpRewritePattern workaround to fix the original CB-overflow on ttnn.concat. The workaround chunks the concat along dim[-2] so each per-chunk CB region fits in usable L1.

The chunk-size formula at lib/Dialect/TTNN/Transforms/Workarounds/Decomposition/ConcatOpRewritePattern.cpp:128-130:
```
int64_t maxC = static_cast<int64_t>(l1UsableSize / (elemSizeBytes * 2));
int64_t chunkSize = (maxC / TILE_HEIGHT) * TILE_HEIGHT;
```
sizes each chunk's CB region at 100% of usable L1. For si32 with l1UsableSize ≈ 1,499,136 B this gives chunkSize = 187,392 and per-chunk CB after the internal transpose = 187,392 × 4 × 2 = 1,499,136 B — exactly the full L1 budget, with zero headroom for any other L1-resident tensor on the same cores.

In real model graphs, surrounding ops produce L1-resident intermediates on the workaround's target cores, and the CB region overlaps them at runtime:

```
TT_THROW: Statically allocated circular buffers in program 590 clash with L1
buffers on core range [0-0 - 0-2]. L1 buffer allocated at 1,433,280 and
static circular buffer region ends at 1,498,912
```
### What's changed
Cap the chunk-size at 80% of usable L1, leaving headroom on each core for co-resident allocations:

// Reserve ~20% L1 headroom so the per-chunk CB doesn't crowd out
// co-resident L1 tensors on the same cores 
int64_t maxC = static_cast<int64_t>(l1UsableSize * 0.8 / (elemSizeBytes * 2));
int64_t chunkSize = (maxC / TILE_HEIGHT) * TILE_HEIGHT;

### Logs
[qwen_3_5_concat_error.log](https://github.com/user-attachments/files/27714641/qwen_3_5_concat_error.log)

### Checklist

### DIFF
--- a/lib/Dialect/TTNN/Transforms/Workarounds/Decomposition/ConcatOpRewritePattern.cpp
+++ b/lib/Dialect/TTNN/Transforms/Workarounds/Decomposition/ConcatOpRewritePattern.cpp
@@ -107,9 +107,7 @@ ConcatOpRewritePattern::matchAndRewrite(ttnn::ConcatOp srcOp,
   uint64_t cbSizeAfterTranspose =
       static_cast<uint64_t>(dimMinusTwo) * elemSizeBytes * 2;
 
-  uint64_t l1UsableSize = ttcore::getCurrentScopeSystemDesc(srcOp)
-                              .getChipDescs()[0]
-                              .getUsableL1Size();
+  uint64_t l1UsableSize = utils::getUsableL1PerCore(srcOp);
 
   if (cbSizeAfterTranspose <= l1UsableSize) {
     // Existing path fits in L1, no workaround needed.


ddilbazTT (approved):

rpavlovicTT (changes requested):
I'd like to push back on landing this as-is. The fix solves the immediate symptom but locks in an L1-budgeting decision that I don't think belongs in the workaround pass at all, and adding `* 0.8` on top of the existing 0.95 cap makes that worse rather than better. I think we should do the proper fix in this PR.

 Why the current shape is wrong:

  1. Two definitions of "usable L1" now disagree. The optimizer side uses `utils::getUsableL1PerCore = chipL1 * tensorL1UsageCap - reservedL1Usage` (default cap 0.95, reserved = const-eval footprint). This pattern reads raw `chipDesc.getUsableL1Size()` and applies its own `0.8`. There is no principled relationship between the 0.8 here and the 0.95 there — they're independent magic numbers managing the same resource. 
  2. This pass runs before const-eval sets the reservation. So even after switching to `getUsableL1PerCore`,  the workaround can't see L1 reservations added by const-eval in this same compile. The 0.8 is essentially a guess at how much const-eval will eat — exactly the kind of guess we should not be encoding in a pattern.
  3. What this pattern does is OOM avoidance, not canonical decomposition. The bail-out  `(if (cbSizeAfterTranspose <= l1UsableSize) return failure();)` is the textbook signature of a fallback. The right home is `TTNNOperationValidationAndFallback`, which already runs post-layout,
  post-const-eval, calls the real `OpModel<ConcatOp>::getOpConstraints`, and has a dedicated `OutOfMemoryError` branch. Putting it there means using actual chosen layouts, seeing the real const-eval reservation, asking tt-metal directly, and auto-retiring when tt-metal/#43371 lands (the OpModel just stops reporting OOM and the fallback stops firing).

Concrete proposal for this PR:

TTNNOperationValidationAndFallback is gated on `opt-level ≥ 1` and on `TTMLIR_ENABLE_OPMODEL`, so we do still need a workaround-pass version for opt-0 / non-OpModel builds. Structure:

  - **opt-0 path** (this pattern): switch the budget query to `utils::getUsableL1PerCore(srcOp)` and drop the * 0.8. Best-effort and consistent with the rest of the compiler's L1 accounting. If 0.95 is too aggressive in practice the fix is to lower the global cap, not to layer another adjustment here.
  - **opt-level ≥ 1 path**: gate this pattern OFF via the `optimizationLevel` option that's already threaded into workaroundOptions, and register the concat fallback inside OperationValidationAndFallback's OOM branch. Same treatment for the twin `SliceStaticOpRewritePattern`.

Happy to pair on the `OperationValidationAndFallback` side if it helps.

sonalibaskaran2499:
Thanks for the detailed feedback @rpavlovicTT agreed on all three points.

I've updated this PR with the opt-0 only: switched the L1 budget query to utils::getUsableL1PerCore(srcOp) and dropped the * 0.8 multiplier. That removes the "two definitions of usable L1" problem you flagged in point 1.

For the opt-≥1, gating this pattern off and moving the OOM handling into TTNNOperationValidationAndFallback's OutOfMemoryError branch. I'd like to take you up on the pairing offer and tackle it in a follow-up PR. Since that change lives in a different pass and likely needs a small refactor to share the decomposition logic between the workaround and the fallback, I think it would be cleaner to address that in a dedicated follow-up PR.
Does that split work for you? Happy to pivot if you'd rather keep everything in one PR.

rpavlovicTT:
> Thanks for the detailed feedback @rpavlovicTT agreed on all three points.
> 
> I've updated this PR with the opt-0 only: switched the L1 budget query to utils::getUsableL1PerCore(srcOp) and dropped the * 0.8 multiplier. That removes the "two definitions of usable L1" problem you flagged in point 1.
> 
> For the opt-≥1, gating this pattern off and moving the OOM handling into TTNNOperationValidationAndFallback's OutOfMemoryError branch. I'd like to take you up on the pairing offer and tackle it in a follow-up PR. Since that change lives in a different pass and likely needs a small refactor to share the decomposition logic between the workaround and the fallback, I think it would be cleaner to address that in a dedicated follow-up PR. Does that split work for you? Happy to pivot if you'd rather keep everything in one PR.

Honestly, I would do it in this PR. This PR is already very small, so adding opt-level >=1 handling won't be a burden. Splitting this hot fix from proper fix (which can be deferred due to other priorities) just creates more mess overall. And I don't see this as a big blocker so we can spend few more days working on it.

PR #8413, opened 2026-05-13
[OpModel] Add CCL and DistributedRMSNorm op constraints support

### LINKED ISSUE #4599: Constraint API for ccl ops
When adding constraint API for `ccl` ops, the following issues were discovered:
```
TTNN_AllGatherOp
TTNN_ReduceScatterOp
TTNN_AllReduceOp
TTNN_CollectivePermuteOp
TTNN_PointToPointOp
```
1. It's not possible to test `opConstraint` and `opRuntime` APIs using the current test infrastructure since all tests right now use single device but these ops require multiple device. So currently the tests are expected to fail with clear messages/comments. However, this should be fixed if we decide to use the optimizer in a multi-device setting.
  - See `TestOpModelInterface.cpp` and `TestOpModelLib.cpp` for these ops: `TTNN_AllGatherOp, TTNN_ReduceScatterOp, TTNN_AllReduceOp`.
2. It's not possible to add `opConstraint` and `opRuntime` APIs for `TTNN_CollectivePermuteOp` since the `tt-metal` definition is missing for the op (there's no `ttnn::collective_permute` op defined in `tt-metal`.
3. It's not possible to add `opConstraint` and `opRuntime` APIs for `TTNN_PointToPointOp` since the `tt-metal` [API](https://github.com/tenstorrent/tt-metal/blob/32c08559e4340f36d63f3c3d8f16184b9a1cac9d/ttnn/cpp/ttnn/operations/point_to_point/point_to_point.hpp#L23C15-L23C30) receives a `GlobalSemaphore` object which is neither possible to be acquired from the op's [definition](https://github.com/tenstorrent/tt-mlir/blob/3cd9982731df1c25736272ac241786f3a636feb4/include/ttmlir/Dialect/TTNN/IR/TTNNOps.td#L2278) in `TTNNOps.td` nor it's trivial to initialize the object. So we might need some help from the metal team to provide easier to use APIs.

rpavlovicTT:
## Change

Implements OpModel constraints for multi-device ops:
- AllGatherOp, ReduceScatterOp, AllReduceOp — via DistributedTensorSpec with sharded topology; input shards distributed along clusterAxis of the mesh
- DistributedRMSNormOp — via DistributedTensorSpec with replicated topology (each device runs RMS locally; only stats are all-gathered); uses fused_rms_minimal with GlobalSemaphore created from input shard spec

Known limitation — single-axis DistributedTensorSpec: TTNNLayoutAttr describes only the single-device tensor layout; it carries no information about how the tensor is distributed across devices in the mesh. The MLIR CCL ops likewise carry only cluster_axis (a single mesh axis). As a result, getOpConstraints builds the DistributedTensorSpec with create_sharded_tensor_topology along the single clusterAxis (or create_fully_replicated_tensor_topology for fused_rms_minimal), even when the runtime tensor may actually be sharded along multiple mesh axes (e.g. [Shard{0}, Shard{1}] on a 2D mesh). For tensors that fit comfortably in L1 this does not change the constraint outcome; for tight L1 fits the per-device size estimate may differ from reality.

Infrastructure changes:
- SingletonDeviceContext: automatic fabric lifecycle (init on multi-chip open/reshape, disable on close); getComputeGridShape() queries actual device grid. `openDevice` sizes the mock env (and thus its fabric enablement) to the mesh actually opened — not the system descriptor's chip count — so opening a sub-mesh of a fabric-enabled env can't launch fabric on a device subset.
- Multi-chip op-model device selection: the "prefer the live cluster descriptor" discovery is restricted to single-chip devices; a multi-chip op-model device always uses the canned mock. The serialized *real* descriptor carries harvested/inactive eth cores that the CCL fabric-routing graph-capture cannot handle on real multi-chip CI hardware (n300/n300-llmbox), whereas the canned mock has a clean topology. Tracked in tt-mlir#9035.
- CCL op-model queries guard against a single-device op-model context (`requireMultiDeviceMesh`): a multi-device CCL query on a single-device device returns a graceful error instead of aborting.
- TTNNRowMajorLayoutPropagation: inserts the corrective `to_layout` when a stopping op *rejects* the RowMajor operand (op-model error) rather than when it merely produces a Tiled result — resolves the TODO left for this PR in #8953 while preserving the integer-input RowMajor fast path.
- SystemDescAttr::getDefault: new overload accepting chipGrid so MLIR device attr matches what the metal device actually reports
- DevicePassesWrapper: reads mesh shape from registered DeviceAttr and passes it to openMockDevice so multi-chip topologies trigger fabric init for CCL constraint queries
- MockDeviceFixture: Galaxy 6U (4x8) as base mock topology; all sub-meshes (N300, T3K, Galaxy) tested in one binary
- OpModelFixture::GetPhysicalGridSize: queries live device instead of hardcoded N300 constant
- TTMETAL_INCLUDE_DIRS: adds tt_metal/api/tt-metalium for metal_context.hpp

Workaround pattern: DistributedRMSNormWidthShardInputRewritePattern no longer writes a hard `memory_config` attribute on the new op. A pinned `memory_config` clashes with any later pass that re-decides the op's result tensor encoding (e.g. greedy MemoryLayoutPropagation choosing a `block_sharded` label that is equivalent to `width_sharded` on a 1xN grid). Passing `nullptr` lets the flatbuffer serializer and EmitC/Py paths derive `memory_config` from the result tensor type, keeping op-attr and result-encoding in sync regardless of what downstream passes choose.

## Testing

`TestOpModelLibMockDevice` exercises MeshPartition / AllGather / ReduceScatter / AllReduce / DistributedRMSNorm across WH + BH sub-meshes (N300, T3K) and the WH 4×8 Galaxy 6U, in a single mock-device binary. Added `CCLShardedOutputTest` covering the collectives' **L1 outputs** — interleaved and width/height/block-sharded — asserting acceptance + L1 placement, and asserting that an infeasible shard grid is rejected *gracefully* (a returned error, not an abort — the contract TTNNRowMajorLayoutPropagation depends on).

## Status

Rebased onto `main` (metal `af5ded324`). The MetalEnv migration this PR originally waited on (tt-metal#43754, tt-mlir#7631) landed on `main` and is used here. **There is no longer a pending tt-metal uplift.**

The metal pin now consumes **tt-umd#2896** (adds `chip_to_bus_id` to the UMD 6u cluster descriptor), so the **WH 4×8 (32-chip Galaxy 6U) mock CCL tests now pass in CI** — the fabric galaxy corner-pinning resolves the rack tray positions in the packaged environment. CI is green across n150 / n300 / n300-llmbox.

### Ticket
Fixes https://github.com/tenstorrent/tt-mlir/issues/4599

### DIFF
--- a/include/ttmlir/OpModel/TTNN/TTNNOpModel.h
+++ b/include/ttmlir/OpModel/TTNN/TTNNOpModel.h
@@ -2087,5 +2087,109 @@ struct OpModel<MeshPartitionOp> {
                TTNNLayoutAttr outputLayout);
 };
 
+//===----------------------------------------------------------------------===//
+// AllGatherOp
+//===----------------------------------------------------------------------===//
+
+template <>
+struct OpModel<AllGatherOp> {
+  static llvm::Expected<OpConstraints> getOpConstraints(
+      llvm::ArrayRef<int64_t> inputShape, TTNNLayoutAttr inputLayout,
+      int32_t allGatherDim, uint32_t clusterAxis,
+      std::optional<uint32_t> subDeviceId, std::optional<uint32_t> numLinks,
+      std::optional<ttcore::Topology> topology, TTNNLayoutAttr outputLayout);
+
+  static llvm::Expected<size_t> getOpRuntime(
+      llvm::ArrayRef<int64_t> inputShape, TTNNLayoutAttr inputLayout,
+      int32_t allGatherDim, uint32_t clusterAxis,
+      std::optional<uint32_t> subDeviceId, std::optional<uint32_t> numLinks,
+      std::optional<ttcore::Topology> topology, TTNNLayoutAttr outputLayout);
+};
+
+//===----------------------------------------------------------------------===//
+// ReduceScatterOp
+//===----------------------------------------------------------------------===//
+
+// Note: reduce_type is not forwarded — ttnn::reduce_scatter always performs
+// sum reduction (no reduce_type C++ API parameter).
+template <>
+struct OpModel<ReduceScatterOp> {
+  static llvm::Expected<OpConstraints>
+  getOpConstraints(llvm::ArrayRef<int64_t> inputShape,
+                   TTNNLayoutAttr inputLayout, int32_t scatterDim,
+                   uint32_t clusterAxis, std::optional<uint32_t> subDeviceId,
+                   std::optional<uint32_t> numLinks,
+                   std::optional<ttcore::Topology> topology,
+                   std::optional<DeviceComputeKernelConfigAttr> computeConfig,
+                   TTNNLayoutAttr outputLayout);
+
+  static llvm::Expected<size_t>
+  getOpRuntime(llvm::ArrayRef<int64_t> inputShape, TTNNLayoutAttr inputLayout,
+               int32_t scatterDim, uint32_t clusterAxis,
+               std::optional<uint32_t> subDeviceId,
+               std::optional<uint32_t> numLinks,
+               std::optional<ttcore::Topology> topology,
+               std::optional<DeviceComputeKernelConfigAttr> computeConfig,
+               TTNNLayoutAttr outputLayout);
+};
+
+//===----------------------------------------------------------------------===//
+// AllReduceOp
+//===----------------------------------------------------------------------===//
+// Note: reduce_type is not forwarded — ttnn::all_reduce always performs sum
+// and the C++ API has no reduce_type parameter.
+template <>
+struct OpModel<AllReduceOp> {
+  static llvm::Expected<OpConstraints> getOpConstraints(
+      llvm::ArrayRef<int64_t> inputShape, TTNNLayoutAttr inputLayout,
+      uint32_t clusterAxis, std::optional<uint32_t> subDeviceId,
+      std::optional<uint32_t> numLinks,
+      std::optional<ttcore::Topology> topology, TTNNLayoutAttr outputLayout);
+
+  static llvm::Expected<size_t>
+  getOpRuntime(llvm::ArrayRef<int64_t> inputShape, TTNNLayoutAttr inputLayout,
+               uint32_t clusterAxis, std::optional<uint32_t> subDeviceId,
+               std::optional<uint32_t> numLinks,
+               std::optional<ttcore::Topology> topology,
+               TTNNLayoutAttr outputLayout);
+};
+
+//===----------------------------------------------------------------------===//
+// DistributedRMSNormOp
+//===----------------------------------------------------------------------===//
+
+template <>
+struct OpModel<DistributedRMSNormOp> {
+  static llvm::Expected<OpConstraints> getOpConstraints(
+      llvm::ArrayRef<int64_t> inputShape, TTNNLayoutAttr inputLayout,
+      std::optional<llvm::ArrayRef<int64_t>> weightShape,
+      std::optional<TTNNLayoutAttr> weightLayout,
+      std::optional<llvm::ArrayRef<int64_t>> residualShape,
+      std::optional<TTNNLayoutAttr> residualLayout,
+      std::optional<llvm::ArrayRef<int64_t>> statsShape,
+      std::optional<TTNNLayoutAttr> statsLayout, uint32_t clusterAxis,
+      llvm::APFloat epsilon, std::optional<uint32_t> subDeviceId,
+      std::optional<uint32_t> numLinks,
+      std::optional<ttcore::Topology> topology,
+      std::optional<DeviceComputeKernelConfigAttr> computeConfig,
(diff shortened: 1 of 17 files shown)


include/ttmlir/OpModel/TTNN/TTNNOpModel.h
```diff
@@ -2018,5 +2018,94 @@ struct OpModel<MeshPartitionOp> {
...
+      std::optional<TTNNLayoutAttr> residualLayout,
+      std::optional<llvm::ArrayRef<int64_t>> statsShape,
+      std::optional<TTNNLayoutAttr> statsLayout, uint32_t clusterAxis,
+      llvm::APFloat epsilon, std::optional<uint32_t> subDeviceId,
+      std::optional<uint32_t> numLinks,
+      std::optional<ttcore::Topology> topology,
+      std::optional<DeviceComputeKernelConfigAttr> computeConfig,
+      std::optional<LayerNormShardedMultiCoreProgramConfigAttr> programConfig,
+      TTNNLayoutAttr outputLayout);
+};
```
sgholamiTT, line 2192: No `getOpRuntime()` impl ?

lib/Dialect/TTNN/Interfaces/TTNNOpModelInterface.cpp
sgholamiTT, line 4148: This needs to be updated ?
Or maybe there's a limitation for `getOpRuntime` that i'm not aware of.

lib/Dialect/TTNN/Interfaces/TTNNOpModelInterface.cpp
```diff
@@ -5075,39 +5110,77 @@ FullOp::getOpRuntime(const std::vector<TTNNLayoutAttr> &inputs,
...
+      inputs[0], getAllGatherDim(), getClusterAxis(), getSubDeviceId(),
+      getNumLinks(), getTopology(), opConfig.outputLayout);
 }
 
 //===----------------------------------------------------------------------===//
 // ReduceScatterOp - TTNN Op Model Interface
 //===----------------------------------------------------------------------===//
-
+// Note: reduce_type is always SUM in ttnn::reduce_scatter (no API parameter).
+// compute_config is intentionally not forwarded.
```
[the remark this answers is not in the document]
sgholamiTT: +1

sgholamiTT (approved):

include/ttmlir/OpModel/TTNN/TTNNOpModel.h
rpavlovicTT: added imps

vtangTT (approved):

sdjordjevicTT (approved):

rpavlovicTT:
## e2e QB2 (Blackhole, 4×P300) validation

Built tt-xla against this branch (`dc8916c77c`, tt-metal `af5ded324`) and ran all **15 tensor-parallel LLM benchmarks** at opt-level 2 (`--pcc-only`), with an A/B against the exact parent `f0f2b84`.

**9/15 pass** (incl. the two largest, llama-3.1-70B and gpt-oss-120B):
falcon3-7b/10b, qwen3-8b, llama-3.1-8b, ministral-8b, mistral-nemo, mistral-small-24b, llama-3.1-70b, gpt-oss-120b — all PCC ≥ 0.99.

**4 regressions, all qwen-family — L1 OOM** (`qwen_2_5_14b`, `qwen_3_14b`, `qwen_3_32b`, `qwen_2_5_coder_32b`). Confirmed PR-caused by A/B: `qwen_3_32b` passes 0.9995/0.9994 on the parent, OOMs here.

**2 unrelated:** `gpt_oss_20b` (+bs=1) fail on the parent too (hang/OOM) — separate pre-existing issue, not this PR.

**Root cause (full analysis in #4599):** enabling the CCL constraints lets the optimizer keep the lm_head `all_gather`'s input in L1 instead of spilling it to DRAM (as the parent did). That input is a `ttnn.reshape [32,37984]→[32,1,37984]` whose output layout is **32× over-tiled** — the collapse map `(d0*32+d1, d2)` with the unit middle dim puts each of the 32 batch rows in its own tile-row → **708 KB/core** vs the ~22 KB the data actually needs — which can't be placed in L1 at runtime. Small CCL tensors work fine in L1 (per-layer `all_reduce` runs L1 at 4 KB/core), so the feature itself is sound; this is a large-tensor layout-sizing edge.

**Suggested path:** add a conservative L1-size cap in the CCL ops' `getOpConstraints` (keep L1 for small tensors, fall back to DRAM above a safe per-core size) → recovers the 4 qwen tests while keeping the 9 green. The precise fragmentation-aware decision is deferred to the incoming truthful-memory-tracking work.

Separate follow-ups (not blocking this PR): the 32× `[B,1,N]` reshape output layout; the gpt-oss MoE instability.

Issue #7735, opened 2026-03-31
Erron when running gpt with Optimizer & D2M


sgholamiTT:
As suggested by @staylorTT, I tried to run `gpt-oss` model with both D2M and Optimizer to measure the perf gain. However, the following error prevents the perf collection:
`'ttnn.concat' op Output tensor dimension 0 must be equal to the sum of input tensor dimensions at the same axis (2), but got 1.`

Steps to reproduce:
1. Get the required machine on ird. I used:
```
ird reserve --timeout 1-0 --docker-image xla wormhole_b0 --num-pcie-chips 4 --model lb
```
2. Clone and build `tt-xla`:
```
git clone https://github.com/tenstorrent/tt-xla.git
cd tt-xla
git submodule update --init --recursive
source venv/activate
cmake -G Ninja -B build
cmake --build build
```
3. Run the gpt-oss model(s) with `d2m-fusion`. Unfortunately, most multi-chip graphs don't work with `ttrt perf` well. As a workaround, we can measure their e2e time using the benchmark test in `xla` and hack up the `tt-mlir` in `third_party` to enable `d2m-fusion`. In order to do that, make the following changes (this code is just a workaround to reproduce the error and should not be checked-in):
  - Add this line to `pjrt_implementation/inc/api/compile_options.h` ([here](https://github.com/tenstorrent/tt-xla/blob/71d741794c68aa319cbe31c013575eb3260693e2/pjrt_implementation/inc/api/compile_options.h#L91)):
```c++
  bool enable_d2m_fusing_pass = false;
```
  - Add this line to `pjrt_implementation/src/api/compile_options.cc` ([here](https://github.com/tenstorrent/tt-xla/blob/71d741794c68aa319cbe31c013575eb3260693e2/pjrt_implementation/src/api/compile_options.cc#L77)):
```c++
  options.enable_d2m_fusing_pass =
      internal::parseBoolOption(compile_options, "enable_d2m_fusing_pass")
          .value_or(options.enable_d2m_fusing_pass);
``` 
- Add this line to `pjrt_implementation/src/api/module_builder/module_builder.cc` ([here](https://github.com/tenstorrent/tt-xla/blob/71d741794c68aa319cbe31c013575eb3260693e2/pjrt_implementation/src/api/module_builder/module_builder.cc#L1028)):
```c++
  options.enableD2MFusing = compile_options.enable_d2m_fusing_pass;
```
[...]

PR #8414, opened 2026-05-13
[TTIR] Narrow RankNormalization scoping to skip only const_eval helpers

mvasiljevicTT:
### Problem description

Found during the **tt-mlir uplift in tt-xla**: `gpt_oss_120b_tp_dp_galaxy` (batch 128) OOMs after #8248 lands.

#8248 scoped `TTIRRankNormalization` to functions with at least one TTIR op in the body, as a proxy for "skip TTNN-only function bodies" (its tests are all `const_eval` helpers, mentioned as *e.g.* in the test comment). In the D2M + Optimizer path:

```
ConvertTTNNToTTIRPass -> D2MFrontend -> TTIRRankNormalization
```

a forward function can have **no TTIR ops in its body** yet carry **rank-1 tensors in its signature** (layer norm scales, `self_attn.sinks`). The proxy wrongly skipped it → rank-1 weights stayed rank-1 → TTNN picked a `<1x1>` single-core layout → DRAM fragmentation → OOM:

```
Out of Memory: Not enough space to allocate 3294625792 B DRAM buffer
... largest free block 215 MB
```

### What's changed

Replace the body-walk proxy with the precise predicate the test cases actually rely on:

- **Skip iff** `tt.function_type = "const_eval"`.
- Otherwise participate (pre-#8248 behavior).

`addLegalDialect<TTNNDialect>()` is kept, so any surviving TTNN ops are still non-candidates for rewriting and #7735 cannot regress.

**Test:** added `@forward_rank1_no_ttir_op` in `rank_normalization_scoping.mlir` — non-`const_eval` func, no TTIR op in body, rank-1 signature. Fails pre-PR, passes post-PR. The existing `@main_const_eval_*` cases continue to pin the `const_eval` skip behavior, so both halves of the predicate are locked in.

**Validation (local):**

| tt-mlir pin                | `gpt_oss_120b_tp_dp_galaxy` (batch 128) |
| -------------------------- | --------------------------------------- |
| `d61eba8f0^` (`9dcd8f5cc`) | passes                                  |
| `d61eba8f0`                | OOM                                     |

### Checklist
- [ ] [perf benchmark for gpt oss](https://github.com/tenstorrent/tt-xla/actions/runs/25800198765/job/75788059529)

### DIFF
--- a/lib/Dialect/TTIR/Transforms/RankNormalization.cpp
+++ b/lib/Dialect/TTIR/Transforms/RankNormalization.cpp
@@ -58,9 +58,10 @@ static bool needsRankExpansion(Type type) {
   return false;
 }
 
-/// Collects functions that participate in rank normalization. A function
-/// participates if it is external and its signature needs rank expansion, or
-/// if its body contains at least one TTIR-dialect op.
+/// Collects functions that participate in rank normalization:
+///   - external functions whose signature needs rank expansion;
+///   - all other non-external functions, except const_eval helpers (which are
+///     intentionally TTNN-only with rank<2 signatures).
 static DenseSet<func::FuncOp> collectParticipatingFuncs(ModuleOp module) {
   DenseSet<func::FuncOp> result;
   module.walk([&](func::FuncOp funcOp) {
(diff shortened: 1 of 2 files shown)


sdjordjevicTT (approved):

brapananTT (approved):
Thanks for adding this!

PR #8424, opened 2026-05-13
[D2M] Rework DstStackAllocator: explicit slot roles, unified allocator

### LINKED ISSUE #8081: [d2m] unary op fusion after int reductions
### Describe the bug

Title. Fusing a unary op onto the back of an int reduction would lead to performing the unary op on the wrong tile because of the way the allocation is handled.

### Expected behavior

_No response_

_No response_

### LINKED ISSUE #8347: [D2M] Redesign `DstStackAllocator` to remove the `outputQueue` second-from-tail hack
### Problem

`DstStackAllocator` (in `lib/Dialect/D2M/Transforms/InsertDstRegisterAccessScheduled.cpp`) allocates DST slices from a free-list `sliceStack`, parks operand reads on `inputStack` (LIFO), and parks store outputs on `outputQueue`. Its `deallocate()` uses a special case that @vwellsTT flagged on PR #7919:

```cpp
if (!inputStack.empty()) {
  id = inputStack.pop_back_val();
} else if (outputQueue.size() > 1) {
  id = outputQueue.at(outputQueue.size() - 2);
  outputQueue.erase(outputQueue.end() - 2);   // <-- this
} else {
  id = outputQueue.back();
  outputQueue.pop_back();
}
```

> "accessing second to last elt of a 'queue' is fundamentally goofy thing to do; I think this code needs algorithmic/DS rethinking (not sure I get the intent tbh)"

The intent is correct (the **tail** of `outputQueue` is the *running* output that the next consumer may still read from, so we deallocate the previous output instead and keep the tail live), but encoding "running output" as the last entry of a queue + a pop-from-the-middle is fragile and confusing. We parked this with an explanatory comment + TODO and bumped redesign to a follow-up.

The same data structure is also responsible for the two `TODO(#8081)` comments in the scratch-slice reservation loop in both `Scheduled.cpp::collectDstAccessesScheduled` and `Unscheduled.cpp::collectDstAccesses`: scratch slots land on `inputStack` (scheduled) / `currSliceIndex` (unscheduled) and are technically safe today but would interact badly with `deallocateAllButFirstInput()` once in-region fusion lands an op after an SFPU integer reduction (issue #8081).

### Requirements

- [ ] Replace the implicit "tail of `outputQueue` is the running output" convention with an explicit `currentOutputSlice` (or equivalent) so `deallocate()` becomes:
  - pop from `inputStack` if non-empty,
  - else release any prior outputs (cleanly) while preserving `currentOutputSlice` for the next consumer.

  In particular, no more `outputQueue.erase(outputQueue.end() - 2)`.
- [ ] Reflect the same shape in `getCurrSliceIndex()`, `getFirstInputSliceIndex()`, and `deallocateAllButFirstInput()` so callers (`collectDstAccessesScheduled`) don't need to know about the internal split.
- [ ] Move scratch slots out of `inputStack` (and out of "current slice" semantics on the bump allocator) so SFPU-int-reduction scratch reads don't pollute the operand stack. This should make the `TODO(#8081)` reservations in the two `getNumDstScratchSlices()` loops a no-op (they currently allocate from the same pool as operands).
- [ ] After this change, the comments referencing the second-from-tail special case and the `TODO(#8081)` markers should be removed.
- [ ] No regressions in the existing scheduled / SFPU-reduction lowerings; green `check-ttmlir` and (when available) green silicon perf runs for ops that exercise multi-input scheduled chains.

### Acceptance criteria

- `DstStackAllocator::deallocate()` has no `erase`-from-the-middle path.
- The two `TODO(https://github.com/tenstorrent/tt-mlir/issues/8081)` markers in `InsertDstRegisterAccess{Scheduled,Unscheduled}.cpp` are deleted along with the workaround they describe; #8081 (or its successor) is unblocked.
- All `insert_dst_register_access_*.mlir` lit tests pass with no IR-output diffs except the trivially-equivalent ones produced by cleaner deallocate ordering.

### References

- PR #7919 review thread: https://github.com/tenstorrent/tt-mlir/pull/7919#discussion_r3202993852
- Related bug: https://github.com/tenstorrent/tt-mlir/issues/8081
- Code: `lib/Dialect/D2M/Transforms/InsertDstRegisterAccessScheduled.cpp` (allocator class + `collectDstAccessesScheduled` scratch loop) and the matching scratch loop in `InsertDstRegisterAccessUnscheduled.cpp`.

### LINKED ISSUE #8364: [D2M] Use `DstStackAllocator` everywhere; remove `DstSliceAllocationState`
### Problem

`InsertDstRegisterAccess` has two allocator types today:

- `DstSliceAllocationState` (bump) — file-local in `InsertDstRegisterAccessUnscheduled.cpp`. ~15 lines. Each `allocate()` returns `nextSliceIndex++`; nothing is ever reclaimed. Used by the matmul / reduction lowering path.
- `DstStackAllocator` — file-local in `InsertDstRegisterAccessScheduled.cpp`. ~150 lines including debug-log helpers and the `outputQueue` middle-erase workaround. Supports LIFO reuse of input slots via `deallocate()`, the "running output" output queue, and a `deallocateAllButFirstInput()` helper.

The bump allocator is structurally a **subset** of the stack allocator: if you never call `deallocate()`, every `allocate()` call on `DstStackAllocator` pops a fresh slot off `sliceStack`, which is the exact bump-allocator semantics. There is no behavioral reason to have two separate types — only the API contract for "this caller doesn't need reuse" was different.

### Requirements

- [ ] Make `DstStackAllocator` (or the renamed replacement coming out of the allocator-redesign follow-up) usable from the unscheduled pass: it has to be constructible from the same `dstCapacity` value the unscheduled pass already computes per generic region.
- [ ] Replace every `DstSliceAllocationState` site in `InsertDstRegisterAccessUnscheduled.cpp` with a `DstStackAllocator` and forward `allocate()` calls without adding `deallocate()` calls. Semantics must be identical (each load gets a fresh slot, no reuse).
- [ ] Delete the `DstSliceAllocationState` class.
- [ ] If both files end up referring to the same allocator type, promote it to a shared anonymous-namespace utility (or to `Shared.{h,cpp}` if we're already exposing other symbols, but preferably keep it file-local to whichever pass remains after the "remove unscheduled" investigation lands).

### Ordering with other follow-ups

This is a **mechanical** consolidation. To avoid inheriting the `outputQueue` middle-erase hack in the unscheduled path during the transition, do this *after* the allocator-redesign follow-up, so the allocator we adopt everywhere is already cleaned up.

If the "remove unscheduled pass" investigation lands first, this issue becomes a no-op (the bump allocator disappears with the file).

### Acceptance criteria

- Only one DST-slot allocator type exists in `lib/Dialect/D2M/Transforms/InsertDstRegisterAccess*.cpp`.
- All `insert_dst_register_access_*.mlir` lit tests pass with no IR-output diffs.

### References

- PR #7919 conversation: https://github.com/tenstorrent/tt-mlir/pull/7919
- Related follow-ups: `DstStackAllocator` redesign; investigate removing the unscheduled pass.

sgholamiTT:
### Ticket
Closes #8347
Closes #8081
Closes #8364

### What's changed

**Simplified the DST slice allocator to a free-slot pool.** The pre-existing `DstStackAllocator` carried an `outputQueue` / `inputStack` / `sliceStack` split with a `deallocate()` method that used a `erase(end()-2)` middle-pop hack to keep the running output live. Every piece of that machinery — `outputQueue`, the middle-erase, and `deallocate()` itself — was dead code in tree (no caller, ever). The class has been simplified to a flat free list (`sliceStack`) plus a per-op operand list (`inputStack`) that exists only so `deallocateAllButFirstInput()` has slots to free, and renamed to `DstSliceAllocator` since it is no longer a stack. The only deallocation the algorithm performs is the multi-input fold case; everything else bump-allocates and lets slots persist until `release_dst`. The API now uses explicit `allocateInput()` / `allocateOutput()` / `allocateScratch()` instead of a boolean flag. The observable slot sequence is unchanged for every existing `insert_dst_register_access_*.mlir` test.

**Added `allocateScratch()` for per-op private scratch (#8081).** SFPU integer reductions reserve extra DST slices via `getNumDstScratchSlices()`. Before this PR those slots were allocated via the same `allocate()` path as operands, which left them on the operand stack and made them visible to `getCurrSliceIndex()`. A fused unary scheduled after `d2m.tile_sfpu_reduce_sum` would then write its output to the scratch slot instead of reusing the reduce's output slot in place. `allocateScratch()` parks the slot in its own pool, invisible to `inputStack` / `currSliceIndex`, so the next compute op cannot pick it up. Both the scheduled and unscheduled scratch-reservation loops route through it, and the `TODO(#8081)` warning comments are gone.

**Unified both passes onto the shared allocator (#8364).** Promoted `DstSliceAllocator` from `InsertDstRegisterAccess/Scheduled.cpp` to `InsertDstRegisterAccess/Shared.{h,cpp}` and removed the unscheduled pass's private `DstSliceAllocationState` bump allocator. The unscheduled `collectDstAccesses` now takes the per-region `dstCapacity` (already computed by the pass for `insertDstRegisterAccessFinalize`) and threads it through. Because the unscheduled path never calls `deallocateAllButFirstInput()`, every `allocateInput()` / `allocateOutput()` returns a fresh slot in 0, 1, 2, … order — bit-for-bit the previous bump-allocator behavior.

**Added regression test for #8081.** `insert_dst_register_access_sfpu_reduce_fused_unary.mlir` hand-crafts a `d2m.tile_negative` directly after a `d2m.tile_sfpu_reduce_sum` in the same generic region — the topology in-region fusion will produce once the downstream #8081 work lands. FileChecks that `dst_scratch_index = 1` on the reduce and `tile_negative` reads/writes `%dst[0]`, with an explicit `CHECK-NOT: %dst[1]` between the load and store of the unary's tile. Verified to fail against a pre-fix build.

### Pre-fix vs post-fix IR (regression test)

```
;; pre-fix: tile_negative latches onto the scratch slot
%9  = "d2m.tile_sfpu_reduce_sum"(%6, %8) <{dst_scratch_index = 1, ...}>
affine.store %9,  %dst[0]
%10 = affine.load %dst[0]
%11 = "d2m.tile_negative"(%10)
affine.store %11, %dst[1]   ;; <-- WRONG: writes scratch slot
%12 = affine.load %dst[1]   ;; <-- WRONG: reads scratch slot

;; post-fix: tile_negative reuses the reduce's output slot in place
%9  = "d2m.tile_sfpu_reduce_sum"(%6, %8) <{dst_scratch_index = 1, ...}>
affine.store %9,  %dst[0]
%10 = affine.load %dst[0]
%11 = "d2m.tile_negative"(%10)
affine.store %11, %dst[0]
%12 = affine.load %dst[0]
```

### Test plan

Full `check-ttmlir`: **1776 passed**, 3 expectedly failed, 213 unsupported — **0 unexpected failures**, no IR-output diffs in any existing `insert_dst_register_access_*.mlir` test. The new regression test passes here and fails against a pre-fix build (verified by reverting the scheduled scratch loop to `allocate()`).

### Checklist

- [x] New/Existing tests provide coverage for changes.
- [x] [Nightly](https://github.com/tenstorrent/tt-mlir/actions/runs/25832643983)


include/ttmlir/Dialect/D2M/Transforms/InsertDstRegisterAccess/Shared.h
```diff
@@ -102,10 +102,65 @@ struct DstAccessCollection {
...
+//
+// `currSliceIndex` is the most recently allocated *operand or output* slot
+// (never a scratch slot) so callers wanting to overwrite-in-place can grab
+// it directly.
+//
+// The scheduled pass uses the full API (LIFO reuse via `deallocate()` and
+// `deallocateAllButFirstInput()`).  The unscheduled pass uses only
+// `allocate()` / `allocateScratch()` / `getCurrSliceIndex()`, which gives
+// it the same bump-allocator behavior its old `DstSliceAllocationState`
+// had: every call returns a fresh slot in 0,1,2,... order.
```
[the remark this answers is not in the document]
vwellsTT: This comment is obviously wrong, but it seems a bit weird to declare this in a "Shared.h" if it's not in fact shared

lib/Dialect/D2M/Transforms/InsertDstRegisterAccess/Shared.cpp
```diff
@@ -50,6 +50,136 @@ LogicalResult verifyInsertDstRegisterAccessPreconditions(ModuleOp moduleOp) {
...
+} // namespace
+
+unsigned DstStackAllocator::allocate(bool isStore) {
+  TT_assertv(!sliceStack.empty(), "Out of dst slices");
+
+  unsigned id = sliceStack.pop_back_val();
+  currSliceIndex = id;
+
+  if (isStore) {
+    // The previous `currentOutput`, if any, has no live consumer at this
```
vwellsTT: I almost feel like we want a circular buffer or something instead of queue so this becomes trivial, no?

lib/Dialect/D2M/Transforms/InsertDstRegisterAccess/Shared.cpp
```diff
@@ -50,6 +50,136 @@ LogicalResult verifyInsertDstRegisterAccessPreconditions(ModuleOp moduleOp) {
...
+}
+
+unsigned DstStackAllocator::deallocate() {
+  TT_assertv(!(inputStack.empty() && retiredOutputs.empty() &&
+               !currentOutput.has_value()),
+             "Deallocating non-existent dst slice");
+
+  // Recycle in liveness order: most recent input first (LIFO), then any
+  // retired prior outputs, then the current output as a last resort
+  // (only legal once the next consumer is gone).  No middle-erase.
```
vwellsTT: nit: probably better not to mention "middle-erase" here, or queue order violations in other comment above; imo comments which cover what changed vs previous iteration are more confusing than helpful to a new reader

lib/Dialect/D2M/Transforms/InsertDstRegisterAccess/Shared.cpp
```diff
@@ -50,6 +50,136 @@ LogicalResult verifyInsertDstRegisterAccessPreconditions(ModuleOp moduleOp) {
...
+  unsigned id = 0;
+  if (!inputStack.empty()) {
+    id = inputStack.pop_back_val();
+  } else if (!retiredOutputs.empty()) {
+    id = retiredOutputs.pop_back_val();
```
vwellsTT: similar to what I said above, it feels like we should be able to just in-place "dealloc" vs storing retired outputs to me; I feel like we could rethink this process as just "use next available slot in buffer" and lazy-delete to simplify the record keeping.  Am I missing some complexity?

vwellsTT (approved):
I think this is definitely an improvement for clarity all-in-all; I'm not quite sure I buy that this is the cleanest form this algorithm can take though (it seems pretty simple at heart to me, I feel like a queue might just be the wrong choice).

However, I certainly might be missing something 😄

include/ttmlir/Dialect/D2M/Transforms/InsertDstRegisterAccess/Shared.h
```diff
@@ -102,10 +102,45 @@ struct DstAccessCollection {
...
-// `DstStackAllocator` in InsertDstRegisterAccess/Scheduled.cpp), since each is
-// only used by one pass.
+// ---------------------------------------------------------------------------
+// DstStackAllocator
+//
+// Free-slot pool for the DST register.  Slots are bump-allocated from
+// `sliceStack` and reclaimed only via `deallocateAllButFirstInput()` (the
+// multi-input fold case); `inputStack` exists so that path has the
+// current op's operand slots to free.  `scratchSlots` is a separate pool
+// so a fused-after-int-reduction op can't pick up the scratch slot as
```
anuragsingh-tt: This is a very specific situation that probably shouldn't be specified in this comment. Can we speak more generally about what this is buying us?

lib/Dialect/D2M/Transforms/InsertDstRegisterAccess/Scheduled.cpp
```diff
@@ -360,15 +225,10 @@ collectDstAccessesScheduled(GenericOp op, Region &region,
           }
         }
 
-        // Reserve any extra DST scratch slices the op declares via the
-        // interface so they don't collide with operand/output slots. No
-        // deallocate: the slot is logically owned by the op for its
-        // lifetime.  TODO(https://github.com/tenstorrent/tt-mlir/issues/8081):
-        // scratch lands on `inputStack`; safe today but a future in-region
-        // fusion would trip `deallocateAllButFirstInput()`.
+        // Reserve any per-op DST scratch slices (#8081).
```
anuragsingh-tt: Why is an issue being referenced here?

lib/Dialect/D2M/Transforms/InsertDstRegisterAccess/Unscheduled.cpp
```diff
@@ -205,13 +189,9 @@ collectDstAccesses(GenericOp gOp, Region &region,
       }
     }
 
-    // Reserve any extra DST scratch slices the op declares via the
-    // interface so they don't collide with operand/output slots.
-    // TODO(https://github.com/tenstorrent/tt-mlir/issues/8081): scratch
-    // becomes the new `getCurrSliceIndex()`; safe today but a future
-    // in-region fusion would trip it.
+    // Reserve any per-op DST scratch slices (#8081).
```
anuragsingh-tt: Why is an issue referenced here?

test/ttmlir/Dialect/D2M/Transforms/insert_dst_register_access_sfpu_reduce_fused_unary.mlir
```diff
@@ -0,0 +1,81 @@
...
+// Regression coverage for #8081 and the allocator redesign in #8347.
+//
+// Topology: an SFPU integer reduction (`d2m.tile_sfpu_reduce_sum`) whose
+// result is consumed by a fused unary op (`d2m.tile_negative`) inside the
+// same generic region.  The SFPU reduce declares
+// `getNumDstScratchSlices() == 1`, so the dst-access insertion pass
+// reserves a private scratch slot for it; the fused unary then asks for
+// an in-place DST slot via `getCurrSliceIndex()`.
+//
+// Pre-fix, the scheduled allocator pushed scratch onto `inputStack` and
```
anuragsingh-tt: Why are we explaining the fix history here?

lib/Dialect/D2M/Transforms/InsertDstRegisterAccess/Shared.cpp
```diff
@@ -50,6 +50,91 @@ LogicalResult verifyInsertDstRegisterAccessPreconditions(ModuleOp moduleOp) {
...
+unsigned DstStackAllocator::allocateScratch() {
+  TT_assertv(!sliceStack.empty(), "Out of dst slices");
+
+  unsigned id = sliceStack.pop_back_val();
+  scratchSlots.push_back(id);
+
+  // Intentionally do NOT update `currSliceIndex` or `inputStack`.
+  // Scratch is owned by the op for the lifetime of the region; it must
+  // not show up as a candidate for in-place reuse by later compute ops
+  // (the failure mode tracked in #8081).
```
anuragsingh-tt: Why is this issue referenced here?

include/ttmlir/Dialect/D2M/Transforms/InsertDstRegisterAccess/Shared.h
```diff
@@ -102,10 +102,45 @@ struct DstAccessCollection {
...
+// its in-place output (#8081).
+// ---------------------------------------------------------------------------
+
+class DstStackAllocator {
```
anuragsingh-tt: I don't think this is a stack allocator any more. `SlotAllocator` might be more appropriate here.

include/ttmlir/Dialect/D2M/Transforms/InsertDstRegisterAccess/Shared.h
```diff
@@ -102,10 +102,45 @@ struct DstAccessCollection {
...
+public:
+  DstStackAllocator() = delete;
+  explicit DstStackAllocator(unsigned dstSliceCapacityIn)
+      : dstSliceCapacity(dstSliceCapacityIn) {
+    initSliceStack();
+  }
+
+  // `isStore=true` skips the operand-list bookkeeping (the slot is the
+  // op's output, not an operand of the current op).
+  unsigned allocate(bool isStore = false);
```
anuragsingh-tt: nit: not a fan of using isStore like this. We should probably use an enum to represent the different states. Or have separate functions for operand, scratch, etc.

include/ttmlir/Dialect/D2M/Transforms/InsertDstRegisterAccess/Shared.h
```diff
@@ -102,10 +102,45 @@ struct DstAccessCollection {
...
+  unsigned allocateScratch();
+  void setStoreToDst() { storedToDst = true; }
+  bool didStoreToDst() const { return storedToDst; }
+  unsigned getCurrSliceIndex() const { return currSliceIndex; }
```
anuragsingh-tt: Is it legal for this function to return 0 (what currSliceIndex is initalized to)?

test/ttmlir/Dialect/D2M/Transforms/insert_dst_register_access_sfpu_reduce_fused_unary.mlir
```diff
@@ -0,0 +1,81 @@
...
+// updated `currSliceIndex`, so `getCurrSliceIndex()` returned the scratch
+// slot and the fused negate would read/write the wrong tile (slot 1, the
+// scratch slot, instead of slot 0, the reduce's output).
+//
+// Post-fix, `allocateScratch()` parks scratch in its own pool: the negate
+// reads/writes the reduce's output slot (slot 0), and `dst_scratch_index`
+// on the reduce points at the separately-allocated slot 1.
+// RUN: ttmlir-opt --ttcore-register-device --d2m-insert-dst-register-access-scheduled --canonicalize -o %t %s
```
anuragsingh-tt, line 14: Do we need an unscheduled test?

lib/Dialect/D2M/Transforms/InsertDstRegisterAccess/Shared.cpp
```diff
@@ -50,6 +50,136 @@ LogicalResult verifyInsertDstRegisterAccessPreconditions(ModuleOp moduleOp) {
...
+    // so a later `deallocate()` can recycle it cleanly without having to
+    // peek at queue tails.
+    if (currentOutput.has_value()) {
+      retiredOutputs.push_back(*currentOutput);
+    }
+    currentOutput = id;
+  } else {
+    inputStack.push_back(id);
+  }
+
```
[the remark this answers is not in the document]
sgholamiTT: Good catch. The right fix turned out to be deleting the state that could go stale, not papering it over with documentation: `currentOutput` and `retiredOutputs` only existed to feed `deallocate()`, and `deallocate()` has had zero in-tree callers since before this PR (the pre-existing `outputQueue + erase(end()-2)` machinery was never called either). Removed all three. The remaining state (`sliceStack`, `inputStack`, `scratchSlots`, `currSliceIndex`) is updated atomically by every operation, so there is no longer a path that can leave anything stale.

include/ttmlir/Dialect/D2M/Transforms/InsertDstRegisterAccess/Shared.h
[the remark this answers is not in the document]
sgholamiTT: Fixed by deleting `deallocate()` rather than the comment around it. Both passes now exercise the same API: `allocate()`, `allocateScratch()`, `getCurrSliceIndex()`, and (scheduled only) `getFirstInputSliceIndex()` + `deallocateAllButFirstInput()`.

lib/Dialect/D2M/Transforms/InsertDstRegisterAccess/Shared.cpp
sgholamiTT: You were right that the queue/LIFO framing wasn't earning its keep. After looking at the call sites more carefully, the only deallocation the algorithm performs is `deallocateAllButFirstInput()` (the multi-input fold case); everything else just bump-allocates and lets slots persist until `release_dst`. So I deleted the unused LIFO recycle path (`deallocate()`, `currentOutput`, `retiredOutputs`) entirely.

What's left is the free-slot pool you described: `sliceStack` is a free list of indices, `inputStack` tracks the current op's operand slots (needed by `getFirstInputSliceIndex()` + `deallocateAllButFirstInput()`), and `scratchSlots` is the per-op scratch pool. No queue ordering, no liveness bookkeeping, no caller-side invariants.
LMK what you think.

lib/Dialect/D2M/Transforms/InsertDstRegisterAccess/Shared.cpp
sgholamiTT: Fixed — and the deallocator the comment was justifying is gone too.

lib/Dialect/D2M/Transforms/InsertDstRegisterAccess/Shared.cpp
sgholamiTT: Not missing anything!
The LIFO `deallocate()` path was speculative API that no caller ever exercised, so the explicit `currentOutput` / `retiredOutputs` partition was bookkeeping for nothing. Deleted both, along with `deallocate()` itself. The allocator is now a free-slot pool plus a per-op operand list (for `getFirstInputSliceIndex()` / `deallocateAllButFirstInput()`); "next free" comes off the back of `sliceStack` and slots are reclaimed lazily.
LMK if this makes sense to you now.

include/ttmlir/Dialect/D2M/Transforms/InsertDstRegisterAccess/Shared.h
sgholamiTT: Fixed — reworded the scratch entry to generic terms: "so that a later compute op can never pick a scratch slot as a candidate for in-place reuse."

lib/Dialect/D2M/Transforms/InsertDstRegisterAccess/Scheduled.cpp
sgholamiTT: Dropped.

lib/Dialect/D2M/Transforms/InsertDstRegisterAccess/Unscheduled.cpp
sgholamiTT: Dropped.

test/ttmlir/Dialect/D2M/Transforms/insert_dst_register_access_sfpu_reduce_fused_unary.mlir
sgholamiTT: Dropped the pre/post-fix narrative.

lib/Dialect/D2M/Transforms/InsertDstRegisterAccess/Shared.cpp
sgholamiTT: Dropped.

include/ttmlir/Dialect/D2M/Transforms/InsertDstRegisterAccess/Shared.h
sgholamiTT: Renamed to `DstSliceAllocator`.

include/ttmlir/Dialect/D2M/Transforms/InsertDstRegisterAccess/Shared.h
sgholamiTT: Split `allocate(bool isStore)` into `allocateInput()` and `allocateOutput()`. Combined with the existing `allocateScratch()`, callers now express intent directly at the call site — no boolean flags.
I can also see arguments both ways, for sth like this:
```c++
enum class AllocateStrategy : unsigned { // Or AllocateFor, etc.
  Input,
  Output,
  Scratch
};
unsigned allocate(AllocateStrategy strategy) { /* ... */ };
```
Let me know if you prefer that. No strong opinion for me.

include/ttmlir/Dialect/D2M/Transforms/InsertDstRegisterAccess/Shared.h
sgholamiTT: Good catch — 0 is a valid slot index, so the uninitialized read was indistinguishable from a legitimate one. Changed `currSliceIndex` to `std::optional<unsigned>` (default `nullopt`); `getCurrSliceIndex()` now asserts that at least one allocation has happened before returning.

test/ttmlir/Dialect/D2M/Transforms/insert_dst_register_access_sfpu_reduce_fused_unary.mlir
sgholamiTT: Don't think it adds coverage. The failure mode is specific to the scheduled fallback: it relies on a later compute op picking up the polluted `getCurrSliceIndex()` as its in-place output.

dlokeTT (approved):
Nice changes

include/ttmlir/Dialect/D2M/Transforms/InsertDstRegisterAccess/Shared.h
anuragsingh-tt: I like the function approach

anuragsingh-tt (approved):

Issue #8385, opened 2026-05-12
[python] Derive in-place / MemWrite operand list from ODS


ndrakulicTT:
**Context**
`tools/chisel/chisel/ops.py` keeps a hand-maintained map of TTNN op class -> role names of operands that are mutated in place (ODS marker: `Arg<..., [MemWrite]>`). The chisel executor uses this list to return one extra tensor per provided in-place operand after the SSA results, so callers know which values were updated by the op.

```python
# tools/chisel/chisel/ops.py
CHISEL_INPLACE_OPS: dict[type, tuple[str, ...]] = {
    ttnn.UpdateCacheOp: ("cache",),
    ttnn.PagedUpdateCacheOp: ("cache",),
    ttnn.FillCacheOp: ("cache",),
    ttnn.PagedFillCacheOp: ("cache",),
    ttnn.WriteTensorOp: ("device_tensor",),
    ttnn.BatchNormTrainingOp: ("running_mean", "running_var"),
    ttnn.PointToPointOp: ("optional_output_tensor",),
    ...
}
```

This is fragile: new ops with `MemWrite` operands silently behave wrong in chisel (executor returns the wrong number of tensors, assertion fires only at test time), and the list drifts from the source of truth in ODS as ops evolve.

**Source of truth**
ODS already encodes this via per-operand decorators. Example from a TTNN op definition:
```
let arguments = (ins
    AnyRankedTensor:$input,
    Arg<AnyRankedTensor, "", [MemWrite]>:$cache,
    ...
);
```

`tools/scripts/python_op_schema_codegen.py` already parses `llvm-tblgen --dump-json` output and emits `OP_SCHEMA[<op-name>] = {"operands": (...), "attributes": (...), "results": (...)}`. We can extend it to capture per-arg decorators (specifically `MemWrite`) and expose them as part of the same sidecar.

**Proposed change**
1. Extend `python_op_schema_codegen.py` to record per-operand decorator info. Suggested shape:
   ```python
   OP_SCHEMA["ttnn.update_cache"] = {
       "operands":   ("cache", "input", ...),
       "memwrite_operands": ("cache",),
       "attributes": (...),
       "results":    (...),
   }
   ```
2. Replace `CHISEL_INPLACE_OPS` lookups in `chisel/ops.py` with a wrapper that reads `OP_SCHEMA[op.name].memwrite_operands` (and returns `()` for ops not in the schema).
[...]

PR #8426, opened 2026-05-14
[chisel] core: callbacks, executor, validators, builder tests

ndrakulicTT:
### Summary
chisel hooks `_ttmlir_runtime` debug callbacks and, for every op the runtime executes, runs the matching CPU golden in isolation and compares shape, dtype, and PCC against the live device tensor. Failures are recorded as structured JSONL records.

```python
import chisel

with chisel.session(results_path="chisel_result.jsonl") as report:
    run_program()
    print(f"{len(report.failures())} failing record(s)")
```

### What's in this PR

**Core (`tools/chisel/chisel/`)**
- `bind.py` — `bind` / `unbind` / `session()` context manager for one-call setup of all four DebugHooks callbacks.
- `callbacks.py` — `_default_pre_op` stashes a host copy of each device input; `_default_post_op` runs the golden in isolation and checks shape/dtype + PCC against the runtime output for every op output.
- `context.py` — `ChiselContext` singleton with `BinaryState` / `ProgramState` hierarchy; nested callback-scope and op-scope lifecycles.
- `op_configs.py` — per-op overrides (`no_golden`, `skip_isolated_pcc`); default table covers `ttnn.empty`, `ttnn.generic`, non-executable / trace / control-flow / I/O ops, in-place ops, quantization ops, and ops without a registered golden.
- `validators.py` — `check_shape_dtype`, per-shard `check_numerics` with configurable `PccConfig`.
- `recorder.py` — `ChiselRecorder` owns the in-memory ring buffer, optional JSONL sink, and per-binary debug-dump policy (`debug_chisel_dir`).
- `report.py` — pydantic record schema with per-status payload discriminator; `ChiselReport` supports filtering by op / check / program / status and offline loading from JSONL.
- `safety.py` — `chisel_safe` decorator: a chisel bug never kills the ttmlir runtime; `ChiselFailure` becomes a structured record, anything else becomes a `chisel_bug` record with traceback.
- `exceptions.py` — `ChiselFailure` hierarchy (`ShapeMismatch`, `DtypeMismatch`, `GoldenNotImplementedError`, `IrRuntimeMismatch`) and `UnexpectedStateError` for lifecycle violations.
- `utils.py` — tensor retrieval, op asm matching the flatbuffer debug_info, `CHISEL_DEBUG` pdb hook.

**Tests (`test/python/chisel/`)**
- `test_builder_chisel_integration.py`, `test_report.py`,

### TODOs deferred to follow-up PRs

- `bind.py:95` - validate `unbind()` against tt-xla for race conditions caused by asynchronous runtime execution.
- `callbacks.py:113` - input device tensors may be pulled to host multiple times across consumer ops; cache to avoid redundant copies.
- `op_configs.py:49` - chisel does not yet validate in-place tensor mutations; in-place ops are currently configured `no_golden=True`.
- `ops.py:37` - `CHISEL_INPLACE_OPS` is hand-maintained; derive from ODS (`python_op_schema_codegen.py`). #8385 

Follow-on feature branches already named in the seed commit message: `chisel-accumulated` (accumulation across ops), `chisel-skip-op` (skip mode), `chisel-load-cache` (load-cache feature).

### DIFF
--- a/tools/chisel/chisel/callbacks.py
+++ b/tools/chisel/chisel/callbacks.py
@@ -1,0 +1,172 @@
+# SPDX-FileCopyrightText: (c) 2026 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+"""Per-op chisel callbacks driven by the ttmlir runtime debug hooks.
+
+Validators raise ChiselFailure on the first failed shape/dtype/PCC check.
+chisel_safe wraps the per-op handler and turns the failure into a structured
+record — any subsequent checks on the same op are skipped. Other exceptions
+are recorded as chisel_bug.
+"""
+import logging
+from contextlib import contextmanager
+from enum import Enum
+from typing import Dict, Iterator, List
+
+from _ttmlir_runtime import runtime as tt_runtime
+from _ttmlir_runtime.binary import Binary
+from _ttmlir_runtime.runtime import CallbackContext, OpContext, TensorRef
+from ttmlir.ir import OpView, Value
+
+from golden import GoldenMapTensor
+
+from .context import ChiselContext, get_instance
+from .exceptions import IrRuntimeMismatch
+from .executor import build_role_keyed_inputs, execute_golden
+from .op_configs import ChiselOpConfig
+from .ops import SSAName, get_op_inputs, get_op_outputs
+from .report import ChiselRecord, NoGoldenPayload, SkippedNumericsPayload
+from .safety import chisel_safe
+from .utils import get_op_asm, retrieve_tensor
+from .validators import check_numerics, check_shape_dtype
+
+logger = logging.getLogger("chisel")
+
+
+class Phase(Enum):
+    PRE = "pre"
+    POST = "post"
+
+
+@contextmanager
+def _op_callback(
+    ctx: ChiselContext,
+    rt_binary: Binary,
+    rt_program_context: CallbackContext,
+    rt_op_context: OpContext,
+    *,
+    phase: Phase,
+) -> Iterator[None]:
+    """Open callback scope (and, on PRE, op scope); close them symmetrically."""
+    program = ctx.begin_callback(rt_binary, rt_program_context, rt_op_context)
+    try:
+        if phase is Phase.PRE:
+            program.begin_op()
+        yield
+    finally:
+        if phase is Phase.POST:
+            program.end_op()
+        ctx.end_callback()
+
+
+def _assert_op_matches_runtime(ctx: ChiselContext) -> None:
+    """Raise IrRuntimeMismatch if chisel and the runtime point at different ops."""
+    rt_debug = tt_runtime.get_op_debug_str(ctx.rt_op_context)
+    op = ctx.op
+    if rt_debug.strip() == get_op_asm(op).strip():
+        return
+    raise IrRuntimeMismatch(op, "ir_vs_runtime_op", rt_debug)
+
+
+def _validate_and_retrieve_tensor(
+    ctx: ChiselContext, mlir_value: Value, rt_tensor_ref: TensorRef
+) -> GoldenMapTensor:
+    op = ctx.op
+    check_shape_dtype(op, "mlir_vs_tensor_ref", mlir_value, rt_tensor_ref)
+    tensor = retrieve_tensor(ctx.rt_program_context, rt_tensor_ref)
+    check_shape_dtype(op, "mlir_vs_runtime_tensor", mlir_value, tensor)
+    return tensor
+
+
+def _run_isolation_golden(
+    op: OpView,
+    mlir_outputs: List[Value],
+    asm_state,
+    ssa_inputs: Dict[SSAName, GoldenMapTensor],
+) -> List[GoldenMapTensor]:
+    role_inputs = build_role_keyed_inputs(op, ssa_inputs, asm_state)
+    iso_outs = execute_golden(op, role_inputs)
+    for mlir_output, iso_out in zip(mlir_outputs, iso_outs, strict=True):
+        check_shape_dtype(op, "mlir_vs_golden", mlir_output, iso_out)
+    return iso_outs
+
+
+@chisel_safe
+def _default_pre_op(ctx: ChiselContext, config: ChiselOpConfig) -> None:
+    """Stash a host copy of every device input so POST can drive the golden."""
+    op = ctx.op
+    if config.no_golden:
+        ctx.write_record(
+            ChiselRecord(
+                op=op.name,
+                check="golden_not_implemented",
+                payload=NoGoldenPayload(),
+            )
+        )
+        return
+
+    _assert_op_matches_runtime(ctx)
+    asm_state = ctx.asm_state
+
+    mlir_op_inputs = get_op_inputs(op)
+    for mlir_input, rt_tensor_ref in zip(mlir_op_inputs, ctx.input_refs, strict=True):
+        # TODO(ndrakulic): Right now we are pulling input device tensors to host potentialy multiple times
+        tensor = _validate_and_retrieve_tensor(ctx, mlir_input, rt_tensor_ref)
+        ctx.stashed_inputs[mlir_input.get_name(asm_state)] = tensor
+
+
+@chisel_safe
+def _default_post_op(ctx: ChiselContext, config: ChiselOpConfig) -> None:
+    """Run the golden, validate shapes/dtypes, and PCC-check each output."""
+    if config.no_golden:
+        return
+
+    skip_isolated_pcc = config.skip_isolated_pcc
+    op = ctx.op
+    asm_state = ctx.asm_state
+
+    mlir_op_outputs = get_op_outputs(op)
+    if not mlir_op_outputs:
+        return
+
+    iso_outs = _run_isolation_golden(op, mlir_op_outputs, asm_state, ctx.stashed_inputs)
+
+    for mlir_output, output_ref, iso_out in zip(
+        mlir_op_outputs, ctx.output_refs, iso_outs, strict=True
+    ):
+        device_tensor = _validate_and_retrieve_tensor(ctx, mlir_output, output_ref)
+        ssa = mlir_output.get_name(asm_state)
+        if skip_isolated_pcc:
+            ctx.write_record(
+                ChiselRecord(
+                    op=op.name,
+                    check="numerics",
+                    ssa=ssa,
+                    payload=SkippedNumericsPayload(),
+                )
+            )
+            continue
+
+        check_numerics(ctx, op, ssa, iso_out, device_tensor)
+
+
+def run_op_callback(
+    rt_binary: Binary,
+    rt_program_context: CallbackContext,
+    rt_op_context: OpContext,
+    *,
+    phase: Phase,
+) -> None:
+    """Dispatch the PRE or POST handler for the current op."""
+    ctx = get_instance()
+    with _op_callback(ctx, rt_binary, rt_program_context, rt_op_context, phase=phase):
+        config = ctx.get_op_config(ctx.op)
+        if phase is Phase.PRE:
+            ok = _default_pre_op(ctx, config)
+            ctx.pre_failed = not ok
+        else:
+            # PRE recorded a failure — skip POST to avoid cascading into a
+            # chisel_bug from incomplete state.
+            if ctx.pre_failed:
+                return
+            _default_post_op(ctx, config)
--- a/tools/chisel/chisel/callbacks.py
+++ b/tools/chisel/chisel/callbacks.py
@@ -5,7 +5,7 @@
 
 Validators raise ChiselFailure on the first failed shape/dtype/PCC check.
 chisel_safe wraps the per-op handler and turns the failure into a structured
-record — any subsequent checks on the same op are skipped. Other exceptions
+record - any subsequent checks on the same op are skipped. Other exceptions
 are recorded as chisel_bug.
 """
 import logging
@@ -33,7 +33,7 @@ from .validators import check_numerics, check_shape_dtype
 logger = logging.getLogger("chisel")
 
 
-class Phase(Enum):
+class CallbackPhase(Enum):
     PRE = "pre"
     POST = "post"
 
@@ -45,16 +45,16 @@ def _op_callback(
     rt_program_context: CallbackContext,
     rt_op_context: OpContext,
     *,
-    phase: Phase,
+    phase: CallbackPhase,
 ) -> Iterator[None]:
     """Open callback scope (and, on PRE, op scope); close them symmetrically."""
     program = ctx.begin_callback(rt_binary, rt_program_context, rt_op_context)
     try:
-        if phase is Phase.PRE:
+        if phase is CallbackPhase.PRE:
             program.begin_op()
         yield
     finally:
-        if phase is Phase.POST:
+        if phase is CallbackPhase.POST:
             program.end_op()
         ctx.end_callback()
 
@@ -110,7 +110,7 @@ def _default_pre_op(ctx: ChiselContext, config: ChiselOpConfig) -> None:
 
     mlir_op_inputs = get_op_inputs(op)
     for mlir_input, rt_tensor_ref in zip(mlir_op_inputs, ctx.input_refs, strict=True):
-        # TODO(ndrakulic): Right now we are pulling input device tensors to host potentialy multiple times
+        # TODO(ndrakulic): Right now we are pulling input device tensors to host potentially multiple times
         tensor = _validate_and_retrieve_tensor(ctx, mlir_input, rt_tensor_ref)
         ctx.stashed_inputs[mlir_input.get_name(asm_state)] = tensor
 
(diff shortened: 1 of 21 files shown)


tools/chisel/chisel/op_configs.py
```diff
@@ -0,0 +1,90 @@
...
+    configs[ttnn.WriteTensorOp] = ChiselOpConfig(no_golden=True)
+    configs[ttnn.BatchNormTrainingOp] = ChiselOpConfig(no_golden=True)
+    configs[ttnn.PointToPointOp] = ChiselOpConfig(no_golden=True)
+
+    # Quantization ops not currently supported.
+    configs[ttnn.QuantizeOp] = ChiselOpConfig(no_golden=True)
+    configs[ttnn.DequantizeOp] = ChiselOpConfig(no_golden=True)
+    configs[ttnn.RequantizeOp] = ChiselOpConfig(no_golden=True)
+
+    # No golden registered yet
```
mmilosevicTT: nit: Missing dot.

tools/chisel/chisel/op_configs.py
```diff
@@ -0,0 +1,90 @@
...
+    configs[ttnn.RMSNormPreAllGatherOp] = ChiselOpConfig(no_golden=True)
+    configs[ttnn.SamplingOp] = ChiselOpConfig(no_golden=True)
+    configs[ttnn.SelectiveReduceCombineOp] = ChiselOpConfig(no_golden=True)
+    configs[ttnn.AllToAllDispatchMetadataOp] = ChiselOpConfig(no_golden=True)
+    configs[ttnn.D2MSubgraphOp] = ChiselOpConfig(no_golden=True)
+
+    return configs
+
+
+def get_no_golden_op_names() -> frozenset[str]:
```
mmilosevicTT:
```suggestion
def get_op_names_no_golden() -> frozenset[str]:
```

nit: It is a bit confusing to me if no_golden comes first.

tools/chisel/chisel/op_configs.py
```diff
@@ -0,0 +1,90 @@
...
+    configs[ttnn.NLPConcatHeadsOp] = ChiselOpConfig(no_golden=True)
+    configs[ttnn.NLPConcatHeadsDecodeOp] = ChiselOpConfig(no_golden=True)
+    configs[ttnn.BitcastConvertOp] = ChiselOpConfig(no_golden=True)
+    configs[ttnn.AsinhOp] = ChiselOpConfig(no_golden=True)
+    configs[ttnn.DistributedLayerNormOp] = ChiselOpConfig(no_golden=True)
+    configs[ttnn.RMSNormPreAllGatherOp] = ChiselOpConfig(no_golden=True)
+    configs[ttnn.SamplingOp] = ChiselOpConfig(no_golden=True)
+    configs[ttnn.SelectiveReduceCombineOp] = ChiselOpConfig(no_golden=True)
+    configs[ttnn.AllToAllDispatchMetadataOp] = ChiselOpConfig(no_golden=True)
+    configs[ttnn.D2MSubgraphOp] = ChiselOpConfig(no_golden=True)
```
mmilosevicTT: could we just do something like 
```
return {
ttnn.EmptyOp : ChiselOpConfig(skip_isolated_pcc = True),
...
ttnn.D2MSubgraphOp: ChiselOpConfig(no_golden=True),
}
```

tools/chisel/chisel/op_configs.py
```diff
@@ -0,0 +1,90 @@
...
+class ChiselOpConfig:
+    # no_golden: do not run the isolation golden; emit a NoGoldenPayload
+    # record per op so the report shows the op was visited but skipped.
+    # skip_isolated_pcc: run the isolation golden (shape/dtype still checked)
+    # but skip the PCC comparison.
+    no_golden: bool = False
+    skip_isolated_pcc: bool = False
+
+
+def default_configs() -> Dict[Type, ChiselOpConfig]:
```
mmilosevicTT: Could we narrow `Type` here to something like `Type[OpView]`?

tools/chisel/chisel/validators.py
```diff
@@ -0,0 +1,111 @@
...
+    # In both cases torch.isclose with these tolerances yields the PCC verdict
+    # directly, because correlation is undefined / 1.0 by convention. They do
+    # Use max_atol/max_rtol below to gate failure on the worst-case diffs.
+    atol: float = 1e-8
+    rtol: float = 1e-5
+
+    # When set, the numerics check also fails if the computed worst-case
+    # absolute/relative diff exceeds the threshold. None = disabled.
+    max_atol: Optional[float] = None
+    max_rtol: Optional[float] = None
```
mmilosevicTT: I am not a fan of namings in this class. We use `max_atol` and `max_rtol` but then we use threshold. PCC is a metric as well, I would rename it to `min_pcc` at least to keep consistent.

tools/chisel/chisel/validators.py
```diff
@@ -0,0 +1,111 @@
...
+        device_shard = device_shards[device_id]
+        atol, rtol, pcc = get_atol_rtol_pcc(
+            golden_shard, device_shard, cfg.atol, cfg.rtol
+        )
+        failed = pcc < cfg.threshold
+        if cfg.max_atol is not None and atol > cfg.max_atol:
+            failed = True
+        if cfg.max_rtol is not None and rtol > cfg.max_rtol:
+            failed = True
+        status = Status.NUMERICS_FAIL if failed else Status.OK
```
mmilosevicTT: We pass all metrics in the report with this, right? If yes then I am ok with this, if not I would like separate status fields to keep report informative.

tools/chisel/chisel/validators.py
```diff
@@ -0,0 +1,111 @@
...
+
+    # atol/rtol below are NOT failure thresholds — they are forwarded to
+    # get_pcc → torch.isclose to short-circuit two edge cases:
+    #   * single-element tensors (one scalar value to compare)
+    #   * constant tensors (all elements identical in golden and device)
+    # In both cases torch.isclose with these tolerances yields the PCC verdict
+    # directly, because correlation is undefined / 1.0 by convention. They do
+    # Use max_atol/max_rtol below to gate failure on the worst-case diffs.
+    atol: float = 1e-8
+    rtol: float = 1e-5
```
mmilosevicTT: I do not think these should be in this config. I think it is better for them to stay in a separate one, or at least have something like `min_pcc` dict which will combine threshold, atol and rtol in one thing. The cleanest way for me would be just a separate dataclass `PCCConfig`

tools/chisel/chisel/utils.py
```diff
@@ -0,0 +1,68 @@
...
+logger = logging.getLogger("chisel")
+
+
+def get_torch_tensor(tensor: Tensor) -> torch.Tensor:
+    rt_data_ptr = tensor.get_data_buffer()
+    rt_dtype = tensor.get_dtype()
+    dtype = mlir_datatype_to_torch_dtype(rt_dtype)
+    shape = tensor.get_shape()
+    torch_tensor = torch.frombuffer(rt_data_ptr, dtype=dtype)
+    return torch_tensor.reshape(shape).clone()
```
mmilosevicTT: Do we need a clone here?

tools/chisel/chisel/utils.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+def retrieve_tensor(
+    rt_program_context: CallbackContext, rt_tensor_ref: TensorRef
+) -> GoldenMapTensor:
+    device_tensor = tt_runtime.retrieve_tensor_from_pool(
+        rt_program_context, rt_tensor_ref
+    )
+    return GoldenMapTensor({0: get_torch_tensor(device_tensor)}, (1, 1))
```
mmilosevicTT, line 41: What if you get no tensor, null?

tools/chisel/chisel/report.py
```diff
@@ -0,0 +1,199 @@
...
+    DTYPE_MISMATCH = "dtype_mismatch"
+    NUMERICS_FAIL = "numerics_fail"
+    ERROR = "error"
+    SKIPPED_NUMERICS = "skipped_numerics"
+    NO_GOLDEN = "no_golden"
+    IR_RUNTIME_MISMATCH = "ir_runtime_mismatch"
+    CHISEL_BUG = "chisel_bug"
+
+
+NON_FAILURE_STATUSES = frozenset({Status.OK, Status.SKIPPED_NUMERICS, Status.NO_GOLDEN})
```
mmilosevicTT:
```suggestion
PASS_STATUS = frozenset({Status.OK, Status.SKIPPED_NUMERICS, Status.NO_GOLDEN})
```

tools/chisel/chisel/report.py
```diff
@@ -0,0 +1,199 @@
...
+    List,
+    Literal,
+    Optional,
+    Union,
+)
+
+from pydantic import BaseModel, ConfigDict, Field
+
+
+class Status(str, Enum):
```
mmilosevicTT:
```suggestion
class RecordStatus(str, Enum):
```

tools/chisel/chisel/context.py
```diff
@@ -0,0 +1,283 @@
...
+
+class ChiselContext:
+    def __init__(self) -> None:
+        self.binaries: Dict[int, "BinaryState"] = {}
+        # In-flight op-callback pointers; None outside an op callback.
+        self._current_callback_binary: Optional["BinaryState"] = None
+        self._current_callback_program: Optional["ProgramState"] = None
+        self.recorder: ChiselRecorder = ChiselRecorder()
+        self.op_configs: Dict[Type, ChiselOpConfig] = default_configs()
+        self.checks_config: ChiselChecksConfig = ChiselChecksConfig()
```
mmilosevicTT, line 52: Let's keep naming consistent. If one config uses plural other should too.

tools/chisel/chisel/recorder.py
```diff
@@ -0,0 +1,155 @@
...
+        self._file.flush()
+
+    def close(self) -> None:
+        if self._file is not None:
+            self._file.close()
+            self._file = None
+
+
+class ChiselRecorder:
+    # Owns the in-memory ring buffer, the optional JSONL sink, and the
```
mmilosevicTT: I prefer class description to be a docstring not a comment.

tools/chisel/chisel/recorder.py
```diff
@@ -0,0 +1,155 @@
...
+        os.makedirs(debug_dir, exist_ok=True)
+        stem = f"binary_{binary_id}"
+        mlir_path = os.path.join(debug_dir, stem + ".mlir")
+        fb_path = os.path.join(debug_dir, stem + ".ttnn")
+        with open(mlir_path, "w") as f:
+            f.write(mlir_source)
+        rt_binary.store(fb_path)
+        return mlir_path, fb_path
+    except Exception:
+        logger.exception("chisel debug dump failed")
```
mmilosevicTT: nit: Full sentence please.

tools/chisel/chisel/recorder.py
```diff
@@ -0,0 +1,155 @@
...
+            self._sink.write(record)
+        self._maybe_dump_debug(record, program, binary_state)
+
+    def _maybe_dump_debug(
+        self,
+        record: ChiselRecord,
+        program: Optional["ProgramState"],
+        binary_state: Optional["BinaryState"],
+    ) -> None:
+        # On the first failure for a given binary, dump the source MLIR and
```
mmilosevicTT: Ditto for docstring

tools/chisel/chisel/recorder.py
```diff
@@ -0,0 +1,155 @@
...
+            if record.program_name is None:
+                record.program_name = program.program_name
+            if record.program_index is None:
+                record.program_index = program.program_index
+        self.report.append(record)
+        if self._sink is not None:
+            self._sink.write(record)
+        self._maybe_dump_debug(record, program, binary_state)
+
+    def _maybe_dump_debug(
```
mmilosevicTT: I am not sure if maybe is required. We dump if we have all debug params configured, if not, we just cannot dump.

tools/chisel/chisel/recorder.py
```diff
@@ -0,0 +1,155 @@
...
+            mlir_path,
+            fb_path,
+            record.status.value,
+            record.op,
+            program.program_name,
+            program.program_index,
+        )
+
+
+def dump_debug_artifacts(
```
mmilosevicTT: if we prefix `maybe_dump_debug` with `_` then this one is private as well.

tools/chisel/chisel/callbacks.py
```diff
@@ -0,0 +1,172 @@
...
+    return iso_outs
+
+
+@chisel_safe
+def _default_pre_op(ctx: ChiselContext, config: ChiselOpConfig) -> None:
+    """Stash a host copy of every device input so POST can drive the golden."""
+    op = ctx.op
+    if config.no_golden:
+        ctx.write_record(
+            ChiselRecord(
```
mmilosevicTT, line 100: Should record include PCC = 1, as we skip op?

tools/chisel/chisel/callbacks.py
```diff
@@ -0,0 +1,172 @@
...
+    rt_op_context: OpContext,
+    *,
+    phase: Phase,
+) -> None:
+    """Dispatch the PRE or POST handler for the current op."""
+    ctx = get_instance()
+    with _op_callback(ctx, rt_binary, rt_program_context, rt_op_context, phase=phase):
+        config = ctx.get_op_config(ctx.op)
+        if phase is Phase.PRE:
+            ok = _default_pre_op(ctx, config)
```
mmilosevicTT: `ok` seems like a really poor naming for a variable

tools/chisel/chisel/callbacks.py
```diff
@@ -0,0 +1,172 @@
...
+from .ops import SSAName, get_op_inputs, get_op_outputs
+from .report import ChiselRecord, NoGoldenPayload, SkippedNumericsPayload
+from .safety import chisel_safe
+from .utils import get_op_asm, retrieve_tensor
+from .validators import check_numerics, check_shape_dtype
+
+logger = logging.getLogger("chisel")
+
+
+class Phase(Enum):
```
mmilosevicTT: Let's prefix this with CallbackPhase, it is really hard to keep track of variables that might have same name in the future. 

Other option is to call enums pre_callback and post_callback.

tools/chisel/chisel/callbacks.py
```diff
@@ -0,0 +1,172 @@
...
+    """Open callback scope (and, on PRE, op scope); close them symmetrically."""
+    program = ctx.begin_callback(rt_binary, rt_program_context, rt_op_context)
+    try:
+        if phase is Phase.PRE:
+            program.begin_op()
+        yield
+    finally:
+        if phase is Phase.POST:
+            program.end_op()
+        ctx.end_callback()
```
mmilosevicTT, line 59: Maybe I do not have enough info on this, but wouldn't we get a call from context for post_op separately? What I wanted to ask is following:
1. Why is POST phase not under try block?
2. Why is POST in finally phase?

mmilosevicTT (commented):
IMO great work on this. I believe this is a good starting point for the tool.

I left some comments and questions regarding the PR. Lmk when you go through them.

tools/chisel/chisel/validators.py
ndrakulicTT: Yeah, all numerics have either NUMERICS_FAIL or OK

tools/chisel/chisel/utils.py
ndrakulicTT: Good catch I will raise an error in that case

tools/chisel/chisel/callbacks.py
ndrakulicTT: I think that would just make it harder to later filter the report, this way you have separate the ops that you dont have implemented golden and ops that have good PCC

tools/chisel/chisel/callbacks.py
ndrakulicTT: So I needed two kind of function for handling the state:
1. Function that should be called before and after each op callback `ctx.begin_callback` and `ctx.end_callback`
2. Function that need to be fired once at begging of the pair of pre/post callback and at the end of the pair `program.begin_op`, `program.end_op`

tools/chisel/chisel/validators.py
ndrakulicTT: Create a separete `PCCConfig` inside the ChiselChecksConfig

tools/chisel/chisel/utils.py
ndrakulicTT: Good catch no need for clone `rt_data_ptr` is already memcpy from data on device so it should be good without it.

mmilosevicTT (approved):
LGTM

Synced offline.

jgrimTT:
This is just a nit, but are most/all chisel tests going to be golden checks? If so it seems to me that having the chisel test folder in `test/python/golden` makes more sense than `test/python`. We could move builder tests into `test/python/golden/builder`

ndrakulicTT:
Current tests are just chisel functionality tests, not intended for actual every day golden checking on whole CI. I think for now it is okay to stay separate.

PR #8429, opened 2026-05-14
runtime/debug: fix GIL-unsafe std::function copy in Hooks::get callbacks

### LINKED ISSUE #7835: GIL-Safety Fix for DebugHooks Callbacks
## Problem

`debug::Hooks` currently copies callback `std::function` objects by value,
which triggers python increase ref count in non-Python threads.

This surfaces when a Python process uses both `ttrt` and tt-xla that releases the GIL for its own threading (e.g. `torch_xla`, `jax`). Even trivial callbacks that never touch Python objects will segfault because the C++ side copies the `std::function` wrapper, and that copy increments the Python reference count **without holding the GIL** - which triggers the seg fault.

## Proposed Fix

Return callbacks by `const&` so no `std::function` copies occur after initial registration.

Dev-X25874:
### Ticket
Fixes #7835

### Problem description
`debug::Hooks::get()` accepted its four callback parameters as
`const std::optional<std::function> &` and then assigned them by value
into the static `Hooks` singleton:

```cpp
config.preOperatorCallback = preOperatorCallback;  // copies std::function
```

`OperationCallbackFn` / `ProgramCallbackFn` are `std::function` wrappers
that, when registered from Python, capture a pybind `nb::callable`.
Copying a `std::function` holding an `nb::callable` increments the Python
reference count. When this copy occurs on a non-Python thread (e.g. a
thread spawned by `torch_xla` or `jax` that has released the GIL), the
refcount bump happens without the GIL → **segfault**.

The mirror problem exists in `unregisterHooks()`: assigning `std::nullopt`
to the `mutable` members destructs the old `std::function`, decrementing
the Python refcount — also off the GIL in the same scenario.

### What's changed
Changed `Hooks::get()` in both the declaration (`debug.h`) and definition
(`debug.cpp`) to take the four callback parameters **by value** instead of
`const&`. The implementation then `std::move`s each into the singleton:

```cpp
config.preOperatorCallback = std::move(preOperatorCallback);
```

Moving a `std::function` transfers internal state via pointer swaps — no
Python object is touched, no refcount is bumped, GIL-irrelevant. The
macOS stub in `stubs_macos.cpp` is updated to match the new signature to
keep the ODR-clean build on that platform.

All no-argument call sites (`debug::Hooks::get()` in
`program_executor.cpp`) are completely unaffected — the default parameters
remain `std::nullopt` and no `std::function` is ever constructed there.

### Checklist


nsmithtt (approved):

ndrakulicTT:
Thanks for PR, I will check it out if it works as expected in `tt-xla` tomorrow and will merge it after that. Just please run the `pre-commit`.

Dev-X25874:
@ndrakulicTT run-tests (llmbox,tracy,18) appears to be a runner infrastructure failure — \Error: Config file is not valid JSON: Unexpected end of JSON input` — unrelated to this change. Could you re-trigger that job?`

Dev-X25874:
Hey @ndrakulicTT — just wanted to gently follow up on this. You mentioned 
you'd check it in tt-xla and merge after that. The PR has been open for 
4 days now and the failing check (run-tests llmbox,tracy,18) is a runner 
infrastructure failure unrelated to this change. Would really appreciate 
a re-trigger and merge when you get a chance! 🙏

ndrakulicTT:
I have tested it with tt-xla ` tests/op_by_op/test_tool_capability.py` just to be sure and everything seams to work.
Great work and thanks for the PR

Dev-X25874:
Thanks so much @ndrakulicTT! Really appreciate you taking the time to test it against tt-xla and following through on the merge. Glad everything checked out! 🙌

PR #8435, opened 2026-05-14
[TT-Lang] Integrate TT-Lang kernels into the StableHLO → TTIR → TTNN pipeline 

jackzhangTT:
### Ticket
[Link to Github Issue](https://github.com/tenstorrent/tt-xla/issues/4813#event-25777791477)

### Problem description
This is the first POC integration of tt-lang into the tt-xla compiler pipeline.

### What's changed
In the current flow, the user's tt-lang @operation kernel is captured at decoration time and stored in a Python kernel registry. The kernel is then emitted as an opaque stablehlo.custom_call in the graph and processed through the StableHLO -> TTIR -> TTNN pipeline without the kernel body, the body lives only in the Python registry, only metadata (kernel_id, arg_roles, etc.) is kept in the IR.

We deliberately defer invoking the tt-lang compiler until after the core TTIR -> TTNN conversion, because tt-lang needs the post-Shardy, shard-local shape information and the final TTNN layout encoding on each operand to compile correctly. This is implemented as two passes appended to the end of the TTIR -> TTNN pipeline:

1. `--ttnn-resolve-tt-lang-kernels` walks the module for all ttnn.tt_lang_ops and invokes the tt-lang compiler via an embedded Python interpreter (pybind11), passing the kernel body looked up from the registry plus the final operand types and grid. The compiler returns a CompiledTTNNKernel carrying everything needed to construct a ttnn.generic (kernel C++ source, CB configs, core ranges, kernel-tensor indices, runtime args, etc.). The pass serializes that to JSON and attaches it as the `kernel_artifact` attribute on the ttnn.tt_lang_op.

2. `--ttnn-lower-tt-lang-to-generic` runs immediately after, reads each resolved op's `kernel_artifact` attribute, and rewrites the ttnn.tt_lang_op into an equivalent ttnn.generic carrying a #ttnn.program descriptor. The kernels carry their C++ source inline (#ttnn.compute_kernel / #ttnn.read_kernel / #ttnn.write_kernel), and cb_configs become #ttnn.kernel_cb descriptors.

Once the lowering happens there are no tt_lang_ops left only ordinary ttnn.generic ops. The resulting ttnn.generic runs through the same TTNN-to-flatbuffer emission and runtime path as hand-written generic kernels.

Note: I decided to keep the TT-lang ops defined with DestinationStyleOpInterface rather than having them be purely in-place operations like update_cache, since you can use the SSA def-use chains for ordering. Whereas for in-place ops you rely solely on side effects.

**Tests added**
**StableHLO → TTIR** conversion (test/ttmlir/Conversion/StableHLOToTTIR/tt_lang_op.mlir):
- Single-output and multi-output tt_lang_op lowering from stablehlo.custom_call with frontend attributes.

**TTIR → TTNN lowering** (test/ttmlir/Dialect/TTNN/tt_lang_op.mlir):
- Single-result and dual-result kernels; verifies attributes (arg_roles, grid_*, kernel_id) are forwarded from ttir.tt_lang_op to ttnn.tt_lang_op.

**Op verifiers** (test/ttmlir/Dialect/TTIR/tt_lang_op_verify.mlir, test/ttmlir/Dialect/TTNN/tt_lang_op_verify.mlir):
- Negative cases: zero "out" operands, arg_roles count vs operand count mismatch, "out" count vs result count mismatch — covered in both TTIR and TTNN verifiers.

**Kernel resolver pass** (test/ttmlir/Dialect/TTNN/Transforms/ttnn_resolve_tt_lang_kernels.mlir):
--ttnn-resolve-tt-lang-kernels is a no-op on a module with no tt_lang_op (@no_tt_lang_op), and an already-populated kernel_artifact is preserved rather than re-resolved (@prebaked_kernel_artifact_is_preserved).

**Lower-to-generic + flatbuffer emission** (test/ttmlir/Dialect/TTNN/tt_lang_op_flatbuffer.mlir):
--ttnn-lower-tt-lang-to-generic rewrites a ttnn.tt_lang_op with a populated kernel_artifact into a ttnn.generic carrying a #ttnn.program (inline kernel C++ source + CB descriptors); ttmlir-translate --ttnn-to-flatbuffer then confirms it emits through the standard generic path with no tt_lang_op-specific emitter code.

**End-to-end on hardware** (tt-xla: tests/torch/ops/test_tt_lang_kernel_e2e.py):
A real tt-lang @operation kernel compiled and executed through the PJRT plugin on n150 (StableHLO → TTIR → TTNN → resolve → lower-to-generic → flatbuffer → execute).


AleksKnezevic:
Awesome! Are we able to serialize the compiled kernels into something other than an int8 array?

jackzhangTT:
> Awesome! Are we able to serialize the compiled kernels into something other than an int8 array?

yes absolutely! I just used int8 array as a placeholder for now wasn't sure what specific type you preferred

AleksKnezevic:
> > Awesome! Are we able to serialize the compiled kernels into something other than an int8 array?
> 
> yes absolutely! I just used int8 array as a placeholder for now wasn't sure what specific type you preferred

@zoecarver you mentioned something about a serialization format when we last spoke?

include/ttmlir/Dialect/TTNN/IR/TTNNOps.td
```diff
@@ -4049,4 +4193,89 @@ def TTNN_ResetGlobalSemaphoreOp : TTNN_Op<"reset_global_semaphore", [OpModelExem
...
+    // ``noc_async_write_tile`` with the synthesized
+    // ``aligned_page_size = tile_h * tile_w * elem_size_bytes`` per
+    // operand, and the compute thread interprets each page as a 32x32
+    // tile (4 16x16 faces). If an operand reaches us in row-major
+    // layout the addresses still land inside the tensor but the bytes
+    // we read don't form a valid tile -- the kernel silently produces
+    // garbage. Reject row-major here so the layout pipeline inserts
+    // ``ttnn.to_layout`` for every operand and the bridge sees a
+    // uniform Tile world.
+    wa::TTNNOperandsWorkarounds getOperandsWorkarounds() {
```
sdjordjevicTT: By default, the compiler generates inputs and outputs in tile format, so it's unclear why we need to specify this through workarounds. This process is handled in the TTNNLayout pass. Workarounds typically involve adding temporary solutions that will be fixed later. If I understand correctly, this is how TT-Lang operates, and its behavior will not change in the future, right?

lib/Dialect/TTIR/IR/TTIROps.cpp
```diff
@@ -8386,4 +8335,68 @@ ::mlir::OpFoldResult mlir::tt::ttir::SubtractOp::fold(FoldAdaptor adaptor) {
...
+             << trimmed << "\".";
+    }
+  }
+
+  if (outCount != getResults().size()) {
+    return emitOpError("number of \"out\" roles (")
+           << outCount << ") must match number of results ("
+           << getResults().size() << ").";
+  }
+  // Mirror the TTNN verifier: at least one "out" operand is required.
```
sdjordjevicTT: I would argue about the comment, I am not sure that we are mirroring the TTNN verifier, we just want on the TTIR level to have that semantics for the TT lang op, hence I would remove the part:
`"Mirror the TTNN verifier"`

lib/Dialect/TTNN/Interfaces/TTNNOpModelInterface.cpp
```diff
@@ -4928,4 +5759,62 @@ TopKOp::getOpRuntime(const std::vector<TTNNLayoutAttr> &inputs,
...
+
+//===----------------------------------------------------------------------===//
+// TtLangOp - TTNN Op Model Interface
+//===----------------------------------------------------------------------===//
+
+// tt-lang kernels are opaque to the TTNN cost model: their bodies are not
+// authored in TTNN dialects and are not materialised until the tt-xla plugin
+// resolves them post-pipeline, so neither op constraints nor runtime can be
+// queried from inside tt-mlir.
+llvm::Expected<op_model::OpConstraints>
```
sdjordjevicTT: We have a trait for this in TTNN, so you don't need to specify anything here, if you attach that trait to the op:
`OpModelExempt`

include/ttmlir/Dialect/TTNN/IR/TTNNOps.td
jackzhangTT: I believe the `shouldForceInputRowMajor` function in TTNNLayout pass actually forces function arguments  marked with `ttcore::ArgumentType::Input` be row_major. I can change the name of the function if you like.

lib/Dialect/TTNN/Interfaces/TTNNOpModelInterface.cpp
jackzhangTT: Thanks! I updated with the new trait

svuckovicTT:
Hey @jackzhangTT, awesome PR, super happy to see tt-lang path coming to life in the compiler! :)

Got a couple questions:

> the PJRT plugin walks the module for all ttnn.tt_lang_ops invokes the tt-lang compiler

Does this mean that tt-mlir won't be able to compile and run examples with tt-lang kernels end-to-end, without tt-xla in the mix? Would it be possible to add a mode where we embed the python code in IR? I could imagine it being useful for testing purposes.

Do you see any hurdles in supporting the emitTTNN path, specifically in Python to start with? If we were able to reach the compiled kernel with the data defined in `TTNN_TtLangOp`, I think we'd be okay there. Happy to hop on a call sometime this week, and discuss this further!

Could you point me to the tt-xla side of the implementation?

jackzhangTT:
> svuckovicTT

Hi Sasha the xla implementation of things is here: https://github.com/tenstorrent/tt-xla/pull/4712

> Does this mean that tt-mlir won't be able to compile and run examples with tt-lang kernels end-to-end, without tt-xla in the mix? Would it be possible to add a mode where we embed the python code in IR? I could imagine it being useful for testing purposes.

The current POC design was to just see if we can get a full end to end run from the tt-xla compiler so as of right now yes tt-xla is needed in the mix. I have a branch I'm working on right now where I move the resolve logic to tt-mlir so it can become a bit more independent of tt-xla. Whether or not we can embed the python code in the IR is an on going discussion as is the design of the integration. @AleksKnezevic thoughts? Maybe we can setup another meeting again?

> Do you see any hurdles in supporting the emitTTNN path, specifically in Python to start with? If we were able to reach the compiled kernel with the data defined in TTNN_TtLangOp, I think we'd be okay there. Happy to hop on a call sometime this week, and discuss this further!

yes lets have a call about this I'll ping on on slack

include/ttmlir/Target/Utils/FuncOpToProgram.h
```diff
@@ -44,14 +44,11 @@ inline std::string getOpLocInfo(mlir::Operation *op) {
 
 inline Value getOperandThroughDPSOps(Value value) {
   auto *op = value.getDefiningOp();
-  if (!op) {
-    return value;
-  }
-  while (isa<DestinationStyleOpInterface>(op)) {
-    assert(op->getResults().size() == 1);
+  while (op && isa<DestinationStyleOpInterface>(op)) {
```
vmilosevic: :warning: **llvm\-prefer\-isa\-or\-dyn\-cast\-in\-conditionals** :warning:
isa\_and\_nonnull\<\> is preferred over an explicit test for null followed by calling isa\<\>
```suggestion
  while (isa_and_nonnull<DestinationStyleOpInterface>(op)) {
```

vmilosevic (commented):

jackzhangTT:
> > > Awesome! Are we able to serialize the compiled kernels into something other than an int8 array?
> > 
> > 
> > yes absolutely! I just used int8 array as a placeholder for now wasn't sure what specific type you preferred
> 
> @zoecarver you mentioned something about a serialization format when we last spoke?

I've changed the format to be a strAttr so its more readable

include/ttmlir/Target/Utils/FuncOpToProgram.h
jackzhangTT: I updated and dissolved the fix in the commit that introduced this change (f8ca523c9fc05ada7cd1fe8bf213d53df6be1eb1)

zoecarver:
> Awesome! Are we able to serialize the compiled kernels into something other than an int8 array?
yes absolutely! I just used int8 array as a placeholder for now wasn't sure what specific type you preferred
@zoecarver you mentioned something about a serialization format when we last spoke?
I've changed the format to be a strAttr so its more readable

We currently emit C++ files/strings for the kernel bodies. I think stringAttr makes sense, or maybe we could just point at the .hpp files that are emitted? 

tt-lang currently does not serialize beyond that, eg to flatbuffer.

lib/Conversion/TTIRToTTNN/TTIRToTTNN.cpp
```diff
@@ -784,6 +784,41 @@ class PagedFillCacheOpConversionPattern
 };
 } // namespace
 
+namespace {
+// Converts `ttir.tt_lang_op` to `ttnn.tt_lang_op`. The op is opaque to the
+// compiler: we forward all metadata attributes (`kernel_id`, `version_tag`,
```
zoecarver, line 806: tt-lang kernel decorator args I guess fall out natrually as those are the tt-lang compiler's responsibility, so args like `--ttl-strict-f32-acc`, `--ttl-fpu-binary-ops`, `--ttl-maximize-dst` and whatever codegen is produced is used here and there are no other args that need to be kept around? I think that works well, just want to confirm.

zoecarver:
@jackzhangTT how is the build of tt-lang integrated here? Do we just assume that the user has also eg pip installed tt-lang and it's present in the build? Or is there some infrastructure I missed that does the tt-lang build? If we assume it's the user's responsibility, how is this tested in CI?

lib/Conversion/TTIRToTTNN/TTIRToTTNN.cpp
jackzhangTT: yes your understanding is correct! I don't think we need to pipe any of the compiler args in as that should be taken care of by the tt-lang compiler.

jackzhangTT:
> > Awesome! Are we able to serialize the compiled kernels into something other than an int8 array?
> > yes absolutely! I just used int8 array as a placeholder for now wasn't sure what specific type you preferred
> > @zoecarver you mentioned something about a serialization format when we last spoke?
> > I've changed the format to be a strAttr so its more readable
> 
> We currently emit C++ files/strings for the kernel bodies. I think stringAttr makes sense, or maybe we could just point at the .hpp files that are emitted?
> 
> tt-lang currently does not serialize beyond that, eg to flatbuffer.

I think it's not a bad idea to point it to the file instead of embedding the contents

jackzhangTT:
> @jackzhangTT how is the build of tt-lang integrated here? Do we just assume that the user has also eg pip installed tt-lang and it's present in the build? Or is there some infrastructure I missed that does the tt-lang build? If we assume it's the user's responsibility, how is this tested in CI?

Yes currently we just assume that the user has pip installed tt-lang and it's present in the build I have not added any infra for it. There needs to be a follow up for adding it to CI

AleksKnezevic:
> > @jackzhangTT how is the build of tt-lang integrated here? Do we just assume that the user has also eg pip installed tt-lang and it's present in the build? Or is there some infrastructure I missed that does the tt-lang build? If we assume it's the user's responsibility, how is this tested in CI?
> 
> Yes currently we just assume that the user has pip installed tt-lang and it's present in the build I have not added any infra for it. There needs to be a follow up for adding it to CI

Is there any way to get this in CI with this PR? I know it's a huge PR to begin with, but it would go a long way in protecting what you're adding.

nsmithtt (approved):
Signing off on cmake changes

dlokeTT (approved):

sshonTT (approved):

lib/CMakeLists.txt
```diff
@@ -117,6 +117,27 @@ target_link_libraries(TTMLIRCompiler PUBLIC coverage_config)
 target_link_directories(TTMLIRCompiler PRIVATE ${TTMLIR_TOOLCHAIN_DIR}/lib)
 target_link_directories(TTMLIRCompiler PRIVATE ${TTMLIR_TOOLCHAIN_DIR}/lib64)
 
+# The tt-lang kernel resolver pass (TTNNResolveTtLangKernelsPython.cpp in
+# MLIRTTNNTransforms) calls into CPython via pybind11, so
+# MLIRTTNNTransforms emits unresolved `Py_*` symbols (libpython). The
+# link dep on `pybind11::embed` inside MLIRTTNNTransforms does NOT
+# reliably propagate through MLIR's `obj.${name}` aggregation to the
```
jameszianxuTT: Sorry, what does "Not reliably propagate" mean? This comment is a little had to understand. Are you saying that based on internal build mechanics the Python3 target may or may not be resolved (Correctly? or at all? sometimes?)

jameszianxuTT (approved):
Added a nit but otherwise this looks good. 

Not a fan of the blind calling into an assumed python interpreter instance with specific loaded modules in the same process, but we talked offline and it is what it is due to ttlang's minimal serialization requirement being a trace of arbitrarily nested python functions with an interpreter. I think the defensive guarding is good though.

I think a future, better integration would be a precompile step to legitimately serialize the ttlang user kernels, in whatever way they are nested, into some format that can be embedded in the IR. But based on discussion that would need to be scoped out. 

Thanks for organizing commits in this PR and making it easier to review.

include/ttmlir/Dialect/TTNN/IR/TTNNOps.td
hshahTT: If `shouldForceInputRowMajor` is the only place where inputs for this op are being set to RM then I think it makes more sense to add an exception for ttnn.tt_lang_op in that function (similar to how we have an exception for the [MeshShard](https://github.com/tenstorrent/tt-mlir/blob/c1283272c006629ec2c44a2191a8aea8505dd44a/lib/Dialect/TTNN/Transforms/TTNNLayout.cpp#L645-L650) op)

hshahTT: If you find a workaround is absolutely needed then I'd prefer you move the implementation to a new function here: lib/Dialect/TTNN/IR/TTNNWorkaroundsPass.cpp, and call that function from within this .td file.

lib/Dialect/TTNN/Transforms/TTNNResolveTtLangKernels.cpp
```diff
@@ -0,0 +1,133 @@
...
+// concrete vector. Empty input -> empty vector (the tt-xla-side
+// defaults to `[1]`). Whitespace/empty components are rejected so a
+// typo doesn't silently degrade to `[]`.
+mlir::FailureOr<std::vector<std::uint32_t>>
+parseMeshShape(llvm::StringRef s, mlir::Operation *errorAnchor) {
+  std::vector<std::uint32_t> out;
+  if (s.empty()) {
+    return out;
+  }
+  llvm::SmallVector<llvm::StringRef, 4> parts;
```
hshahTT: Why do we expect the mesh shape to have 4 parts?

lib/Target/TTNN/TTNNToFlatbuffer.cpp
```diff
@@ -3590,6 +3593,609 @@ createOp(FlatbufferObjectCache &cache, GenericOp op) {
       ::tt::target::ttnn::ProgramType::ProgramDescriptor, program.Union());
 }
 
+// ============================================================================
+// tt-lang kernel_artifact deserialiser
+// ============================================================================
+//
+// Schema is documented end-to-end in
+// `python_package/tt_torch/tt_lang.py::_serialize_compiled_operation`. The
```
hshahTT: Might be worth mentioning in this comment that this file path is within the tt-xla project, not this one

lib/Target/TTNN/TTNNToFlatbuffer.cpp
```diff
@@ -3590,6 +3593,609 @@ createOp(FlatbufferObjectCache &cache, GenericOp op) {
...
+  // count, but assigning IDs and core ranges needs the structured
+  // semaphore layout that a future schema bump will carry.
+  std::vector<::flatbuffers::Offset<::tt::target::ttnn::SemaphoreDescriptor>>
+      semaphores;
+
+  return ::tt::target::ttnn::CreateProgramDescriptorDirect(*cache.fbb, &kernels,
+                                                           &semaphores, &cbs);
+}
+
+// Lower `ttnn.tt_lang_op` to a `GenericOp` flatbuffer record. The op
```
hshahTT: Instead of lowering tt_lang_op to a GenericOp flatbuffer record, what do you think of lowering ttnn.tt_lang_op to ttnn.generic_op in an earlier ttnn pass? That way we won't have to make many changes in this file and it will simplify the overall logic.

Up to now, the semantics for this file has just been to lower a ttnn op into its own flatbuffer record; and all needed conversions were done before this file. It feels a bit odd to perform a conversion from one op to another here.

hshahTT (changes requested):
Overall looks good. The only thing I'm requesting a change on is the TTNN workaround for the Tiled layout for input and output tensors. Instead of a workaround we should just add an exemption in `shouldForceInputRowMajor` for this op.

In the case a workaround is absolutely needed, we should move the implementation to a function in the TTNNWorkaroundsPass.cpp and call that function in the TTNNOps.td file.

I also would like to have a discussion about adding a lowering from ttnn.tt_lang_op to ttnn.generic_op in an earlier TTNN only pass instead of doing that implicit lowering in TTNNToFlatbuffer. I won't block on that however, since it can be done in a follow-up PR if needed.

lib/Target/TTNN/TTNNToFlatbuffer.cpp
jackzhangTT: yea I think I didn't understand the semantics of the file well enough before creating this. I'm not opposed to creating an earlier pass to do the actual conversion but I think that can be in a follow up PR if necessary. Also I agree that having a work around is not the way to go here I think you're right I should have instead just modified the `shouldForceInputRowMajor` function

svuckovicTT: +1 on this. Would love to see this in `TTIRToTTNN` pass (or earlier). If there is no urgent P0 issue that this PR addresses, I would push back on mainlining this in current form until this logic is moved.

svuckovicTT (changes requested):

jackzhangTT:
@hshahTT @svuckovicTT I've moved the lowering of the tt lang ops to ttnn.generic in a seperate dedicated pass now can you guys take a look

include/ttmlir/Dialect/TTIR/IR/TTIROps.td
```diff
@@ -6678,4 +6678,72 @@ def TTIR_MoeComputeOp : TTIR_NamedOp<"moe_compute", [TTIR_CCL, AttrSizedOperandS
     let hasVerifier = 1;
 }
 
+def TTIR_TtLangOp : TTIR_Op<"tt_lang_op",
```
svuckovicTT:
```suggestion
def TTIR_TTLangOp : TTIR_Op<"tt_lang_op",
```

I believe this would be the correct casing style. There's also a couple other spots in this PR to update this, like pass names, etc.

svuckovicTT (approved):

include/ttmlir/Dialect/TTNN/Transforms/Passes.td
```diff
@@ -871,4 +871,94 @@ def TTNNGreedyMemoryLayoutPropagation : Pass<"ttnn-greedy-memory-layout-propagat
...
+      ...
+    }
+
+    Each kernel becomes a `#ttnn.compute_kernel` / `#ttnn.read_kernel` /
+    `#ttnn.write_kernel` attribute carrying the kernel's C++ source inline
+    (the `source` parameter), its CB-index / tensor-accessor compile-time
+    args, and its buffer-address common runtime args. `cb_configs` become
+    `#ttnn.kernel_cb` descriptors. The resulting `ttnn.generic` runs through
+    the same flatbuffer / runtime path as hand-written generic kernels, so
+    the TTNN-to-flatbuffer emitter needs no `tt_lang_op`-specific code.
```
hshahTT: Nit: everything after "The resulting `ttnn.generic` runs..." isn't needed

include/ttmlir/Dialect/TTNN/Transforms/Passes.td
```diff
@@ -871,4 +871,94 @@ def TTNNGreedyMemoryLayoutPropagation : Pass<"ttnn-greedy-memory-layout-propagat
...
+
+    The op is destination-passing style (`arg_roles == in* out+`), so the
+    `ttnn.generic`'s `inputs_and_outputs` operands are the `tt_lang_op`
+    operands in declaration order, and each `tt_lang_op` result is replaced
+    by its tied trailing "out" operand.
+
+    This pass must run after `--ttnn-resolve-tt-lang-kernels` and before
+    flatbuffer emission. A `ttnn.tt_lang_op` with an empty/unset
+    `kernel_artifact` is an error (the resolve pass must run first).
```
hshahTT: Nit: "(the resolve pass must run first)" isn't needed

lib/Dialect/TTNN/Transforms/TTNNLowerTTLangToGeneric.cpp
```diff
@@ -0,0 +1,481 @@
...
+    op.emitError("kernel_artifact root is not a JSON object.");
+    return {};
+  }
+
+  // Schema version gate. The emitter inserts tensor-accessor markers into
+  // every NOC kernel's compile-time args and the runtime expands each at
+  // launch time, so the artifact carries no pre-baked TensorAccessor
+  // values. Anything other than the current version is fatal so the build
+  // fails loudly on schema drift.
+  constexpr int64_t kFormatVersion = 1;
```
hshahTT: Nit: I feel like this should be renamed to `EXPECTED_FORMAT_VERSION` and moved to the top of the file to make it more visible in case we need to change the version in the future

hshahTT (approved):
Thx for the changes! Left a few nits but overall looks great to me

include/ttmlir/Dialect/TTNN/Transforms/Passes.td
jackzhangTT: I think I agree that 
```
    The op is destination-passing style (`arg_roles == in* out+`), so the
    `ttnn.generic`'s `inputs_and_outputs` operands are the `tt_lang_op`
    operands in declaration order, and each `tt_lang_op` result is replaced
    by its tied trailing "out" operand.
```
can be removed but I its good to keep the last bit about this pass needing to be ran after `ttnn-resolve-tt-lang-kernels`

PR #8438, opened 2026-05-14
Parameterize L1AddrPtrType on 8/16/32 bits and remove volatile

arichinsTT:
### Ticket
for tt-lang element r/w bf16 support

### Problem description
- Parameterize `L1AddrPtrType` with an `elementWidth` parameter (defaulting to 32) to support sub-32-bit L1 pointer access (i8/uint8 and i16/bf16)
- Remove the `volatile` qualifier from L1 pointer types in the TTKernel EmitC path -- it is unnecessary for L1 pointers and was adding noise to the generated code
- Add i8 (8-bit) and i16 (16-bit) support to the L1 pointer load/store infrastructure, enabling sub-word L1 memory access

### What's changed
**Dialect definitions:**
- `TTKernelOpsTypes.td`: Add `elementWidth` parameter to `L1AddrPtrType` with validation for widths 8, 16, and 32
- `TTKernelOps.td`: Rename `CastToL1PtrOp` mnemonic from `reinterpret_cast<volatile tt_l1_ptr uint32_t*>` to `reinterpret_cast<tt_l1_ptr uint32_t*>`; widen `StoreToL1Op` and `LoadFromL1Op` value types from `I32` to `AnyTypeOf<[I8, I16, I32]>`

**EmitC conversion:**
- `TTKernelToEmitC.cpp`: Add `getL1PtrOpaqueTypeName()` that maps `elementWidth` to the correct C type (`tt_l1_ptr uint8_t`, `tt_l1_ptr uint16_t`, `tt_l1_ptr uint32_t`) without `volatile`; add dedicated `TTKernelCastToL1PtrOpToEmitCOpRewriter` that emits the width-appropriate `reinterpret_cast`; update `StoreToL1Op` rewriter to derive the pointee type from the pointer rather than hardcoding

**D2M lowering:**
- `D2MToTTKernel.cpp`: Update all `CastToL1PtrOp` creation sites to explicitly pass the result type (`L1AddrPtrType::get(ctx, 32)`) since the op is now parameterized

**CAPI:**
- `TTKernelTypes.cpp`: Pass explicit `elementWidth=32` to `L1AddrPtrType::get()`

**Tests:**
- Updated all 6 test files to remove `volatile` from op names and CHECK lines
- Added i8 and i16 test cases for `L1AddrPtrType<8>`/`<16>`, `load_from_l1` (i8/i16), and `store_to_l1` (i8/i16) in both dialect roundtrip and EmitC conversion tests

### Checklist


nsmithtt (approved):
Maybe we should add support for I8 while we're at it?

zoecarver (approved):

lib/Conversion/TTKernelToEmitC/TTKernelToEmitC.cpp
```diff
@@ -211,6 +211,17 @@ static std::string ensureCBDeclaration(Value cb, Operation *useOp,
   return cbName;
 }
 
+static StringRef getL1PtrOpaqueTypeName(unsigned elementWidth) {
+  switch (elementWidth) {
+  case 16:
+    return "volatile tt_l1_ptr uint16_t";
```
zoecarver: Is there a reason we need volatile here? I know it was there before so I guess this could be a future improvement, but I don't really see why it's needed and it might block some optimization.

nsmithtt: Most uses in the metal/llk codebase have it defined this way. @nathan-TT is volatile needed for l1 pointers? Or is this just an unnecessary copy paste?

nathan-TT: most likely unnecessary copy-paste.

hkwonTT (approved):

tapspatel (approved):

PR #8439, opened 2026-05-14
[d2m] Handle multi-level synchronized scopes in SplitUnifiedThread

nsmithtt:
The previous algorithm assumed all SynchronizableOpInterface ops in a unified-thread generic sit at the same nesting level (a single parent), which it asserted. Some dsl generated matmul kernels violate that assumption: remote_loads live inside the K accumulator loop, while the remote_store of the result lives in the enclosing loop. Two synchronized scopes coexist, one per parent.

Drop the single-scope assertion and tighten the [start, end) expansion in two ways:

1. Stop at a sibling op already in `opsWithSynchronizableOps`. Such an op hosts its own synchronized scope at a deeper level (e.g. the inner accumulator loop next to the outer-loop remote_store); folding it in would nest sync regions.

2. Stop before pulling in any op whose result is consumed past the wrap. `wrapInSynchronizedRegion` erases every op in the range whose operand chain reaches a non-pure op, so any external user would dangle. The check has to be unconditional on purity: an op like arith.index_cast is itself Pure but is treated as non-pure by the wrap utility once arith.divsi (which can fault) enters its operand chain.

### DIFF
--- a/lib/Dialect/D2M/Transforms/SplitUnifiedThread.cpp
+++ b/lib/Dialect/D2M/Transforms/SplitUnifiedThread.cpp
@@ -143,7 +143,7 @@ LogicalResult wrapComputeInSynchronizedRegion(GenericOp genericOp,
     return failure();
   }
 
-  // Expand and merge compute regions until we hit a syncrhonizable op on both
+  // Expand and merge compute regions until we hit a synchronizable op on both
   // ends.
   //
   // Beyond hitting a SynchronizableOpInterface op (which bounds the wrap),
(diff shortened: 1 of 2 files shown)


lib/Dialect/D2M/Transforms/SplitUnifiedThread.cpp
```diff
@@ -286,6 +538,27 @@ insertCBOpsForCompute(Block *computeBlock, PatternRewriter &rewriter,
...
+      // remote_store outside it), placing wait/pop or reserve/push around
+      // `synchronizedOp` would emit N pairs against a single DMA op. Climb
+      // synchronizedOp's ancestors until we reach one whose parent block
+      // matches the partner's; that ancestor is the anchor for the CB ops.
+      auto anchorForPartner = [&](Operation *partner) -> Operation * {
+        Block *targetBlock = partner->getBlock();
+        Operation *anchor = synchronizedOp;
+        while (anchor->getBlock() != targetBlock) {
+          Operation *parent = anchor->getParentOp();
+          if (!parent) {
```
vtangTT: shouldn't we assert or fail here, cuz this branch means we walked all parents and didn't find a partner?

lib/Dialect/D2M/Transforms/SplitUnifiedThread.cpp
```diff
@@ -89,15 +94,17 @@ LogicalResult wrapComputeInSynchronizedRegion(GenericOp genericOp,
...
+  // (e.g. remote_loads in an accumulator loop bounding the matmul compute,
+  // while a remote_store in the enclosing loop bounds a separate compute
+  // region around the epilogue).
   DenseSet<Operation *> opsWithSynchronizableOps;
   genericOp.getRegion(0).walk([&](Operation *op) {
     if (dyn_cast<SynchronizableOpInterface>(op)) {
       opsWithSynchronizableOps.insert(op->getParentOp());
     }
   });
-  assert(opsWithSynchronizableOps.size() == 1 &&
```
vtangTT, line 99: should we just early return if no ops with sync region? 
```
if (opsWithSynchronizableOps.empty()) {
    return success(); 
}
```

vtangTT (commented):
is it possible make some of the comments less verbose 😅

vtangTT (approved):
@sohaibnadeemTT ptal as well 🙏

alexbaden:
Gentle ping... would be great to get this landed for our triton efforts.

lib/Dialect/D2M/Transforms/SplitUnifiedThread.cpp
```diff
@@ -166,6 +222,179 @@ LogicalResult wrapComputeInSynchronizedRegion(GenericOp genericOp,
     computeRegions.push_back({start, std::next(end)});
   }
 
+  // Merge overlapping compute regions and fixpoint-expand each so
```
sohaibnadeemTT, line 225: Not sure if this makes sense, I think acquire_dst and tile_add should be wrapped together into a single region?

PR #8447, opened 2026-05-15
Group norm decomposition

djstefanovicTT:
### Problem description
Group norm can have problematic configurations that cannot be always executed on TT-device when coming directly from composite ops.

### What's changed
Added decomposition to primitive ops if optimizer is not enabled or if it cannot be verified to work. Simplified some decomposition logic to use existing helpers instead of duplicating them.

### Checklist

### DIFF
--- a/test/ttmlir/Dialect/TTNN/Transforms/Decomposition/group_norm_decomposition.mlir
+++ b/test/ttmlir/Dialect/TTNN/Transforms/Decomposition/group_norm_decomposition.mlir
@@ -1,0 +1,73 @@
+// RUN: ttmlir-opt --ttcore-register-device --ttnn-decomposition -o %t %s
+// RUN: FileCheck %s --input-file=%t
+
+#dram = #ttnn.buffer_type<dram>
+#ttnn_layout = #ttnn.ttnn_layout<(d0, d1, d2, d3) -> (d0 * 9216 + d1 * 9216 + d2, d3), <1x1>, memref<73728x128xf32, #dram>, <interleaved>>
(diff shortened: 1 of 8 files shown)


test/ttmlir/Dialect/TTNN/Transforms/Decomposition/group_norm_decomposition.mlir
```diff
@@ -0,0 +1,73 @@
+// RUN: ttmlir-opt --ttcore-register-device --ttnn-decomposition -o %t %s
```
sdjordjevicTT, line 1: Do we have some builder tests? Just make sure that we are not regressing ones that are passing now, it would be great if we can run them still(optimizer isn't enabled in builder tests, hence this op will always be decomposed with this PR)

sdjordjevicTT (approved):

test/ttmlir/Dialect/TTNN/Transforms/Decomposition/group_norm_decomposition.mlir
djstefanovicTT: There are some generic group norm builder tests in `tt-mlir/test/python/golden/ttir_ops/normalization/test_normalization.py`.  I could add tests for the specific problematic case.

sdjordjevicTT: I was thinking of just marking that test with:
https://github.com/tenstorrent/tt-mlir/blob/0371fd5184c3bb394cd17e93ff908952f9c746e4/test/python/golden/ttir_ops/reduction/test_reduction.py#L265

PR #8451, opened 2026-05-15
[D2M] Explicate cross-type elementwise binary ops in `PredicateTypeAlignment`

### LINKED ISSUE #8193: [D2M][Optimizer] Allow `ttnn.gt` to be part of `d2m_subgraph`s
When trying to bring up `gpt-oss` through D2M + Optimizer, we encountered an issue in `insertDstRegisterAccess` pass:
```
loc("compare.327"): error: 'affine.store' op value to store must have the same type as memref element type
```
I've included the IR dump from `d2mallocate -> insertdstRegisterAccess` so we can see the result of the block factor workaround and also a repro of the issue in a single `d2m` subgraph (it is in `d2m_subgraph_24`).
This should be reproducible with 
```
ttmlir-opt --ttcore-register-device --d2m-insert-dst-register-access insertdstregisterpass_repro.mlir
```

[d2mallocate_to_insertdstregisteraccess_log.txt.zip](https://github.com/user-attachments/files/27216357/d2mallocate_to_insertdstregisteraccess_log.txt.zip)
[insertdstregisterpass_repro.txt](https://github.com/user-attachments/files/27216338/insertdstregisterpass_repro.txt)

Initial analysis of the issue:
- **Offending op:** `ttnn.gt(%arg0: si32, %arg1: si32) -> bf16 (rank-4 broadcast 1x1x1x128 vs 1x1x17x1 → 1x1x17x128).`
- **Problem:** In d2m it lowers to `tile_bcast (si32→bf16) + tile_sub + tile_gtz`, then `affine.store bf16` into an `si32-typed DST memref` → verifier error.
- **Root cause:** cross-type `d2m.tile_bcast` inside one generic; the pass assumes a uniform DST element type and doesn't insert `d2m.dst_reinterpret_cast` at the `si32→bf16` boundary.
- **Scope:** only this one generic in `@d2m_subgraph_24`; no other subgraph has a cross-type tile_bcast.

--------------

Since this was blocking the model bring up, we decidede to disallow `gt` to be part of a `d2m_subgraph`.
However, the op should be allowed to be part of subgraphs once this issue is fixed.

sgholamiTT:
### Ticket
Closes #8193 

### Problem description

GPT-OSS-20B bring-up via D2M + Optimizer hit a verifier failure in `d2m-insert-dst-register-access`:

```
loc("compare.327"): error: 'affine.store' op value to store must have the same type as memref element type
```

A TTNN binary op like `ttnn.gt(si32, si32) -> bf16` (also `ttnn.eq(si32, si32) -> bf16`) carries an implicit typecast inside the kernel. D2M's compute region requires a single uniform element type, so the lowering ends up writing a bf16 result into a si32-typed DST slot. Tracked as [#8193](https://github.com/tenstorrent/tt-mlir/issues/8193); the workaround in #8194 (commit `001f864`) just removed `ttnn.gt` from the `d2m_subgraph` allow-list.

### Root cause

`AlignElementwiseBinaryTypesPattern` in `PredicateTypeAlignment.cpp` only promoted i1 to a real element type when lhs/rhs/result agreed on a single non-i1 type. When they disagreed (e.g. `ttir.eq(si32, si32) -> bf16`) the pattern bailed and the cross-type op reached D2M unchanged. The pattern also rebuilt new tensor types by reusing the original `encoding`, leaving the `ttnn.ttnn_layout`'s memref element type stale for downstream passes that read it.

### Fix
`AlignElementwiseBinaryTypesPattern` now falls back to the result element type as the target when operand and result types disagree and the result is non-i1, inserting typecasts on the disagreeing operands. A new `buildTensorWithElementType` helper rebuilds the `ttnn.ttnn_layout` encoding (preserving tile wrapping) so the layout's memref element type stays in sync with the outer tensor element type; all patterns in the file route through it.

Before:

```mlir
%0 = "ttir.eq"(%arg0, %arg1) : (tensor<4x4xsi32>, tensor<4x4xsi32>) -> tensor<4x4xbf16>
```

After:

```mlir
%0 = "ttir.typecast"(%arg0) : (tensor<4x4xsi32>) -> tensor<4x4xbf16>
%1 = "ttir.typecast"(%arg1) : (tensor<4x4xsi32>) -> tensor<4x4xbf16>
%2 = "ttir.eq"(%0, %1) : (tensor<4x4xbf16>, tensor<4x4xbf16>) -> tensor<4x4xbf16>
```

### Reverts

Reverts `001f864341826fb224818fb797dcb538980d22fe` ("[D2M][Optimizer] Disallow `ttnn.gt` op to be part of d2m_subgraph"). `ttnn.gt` is back in the eltwise-binary allow-list of `TTNNCreateD2MSubgraphs.cpp` now that the underlying bug is fixed.

### Testing

New e2e snippets under `test/python/golden/mlir_snippets/models/gpt_oss_20b/`, wired into `test_fusion_with_optimizer.py`:

- `attention_mask_compare_decode.mlir` — `gt(si32, si32) -> bf16` + 2× `logical_and`, decode shape (`1x1x1x128`); the exact subgraph the workaround used to filter out.
- `attention_mask_compare_prefill.mlir` — same pattern, prefill shape (`1x1x17x128`); matches `d2m_subgraph_24` from the issue dump.
- `compare_eq_multiply.mlir` — `eq(si32, si32) -> bf16` then `multiply(bf16, bf16) -> bf16` with broadcasted shapes (`1x4x544` / `4x1x1` / `4x4x544`).

All 13 entries in `test_fusion_with_optimizer.py` pass on silicon (n150) with `--require-opmodel`. The lit suite (`check-ttmlir`) was extended with three cross-type cases in `predicate_type_alignment.mlir` and passes in full (1761 / 1761 applicable; 213 unsupported / 3 expectedly-failed are pre-existing).

### Checklist

- [x] New tests cover the regressions (lit + e2e snippets).
- [x] Existing tests still pass (`check-ttmlir` and `test_fusion_with_optimizer.py`).
- [x] Workaround commit `001f864` is reverted in this PR.


xanderchin (approved):

lib/Dialect/TTIR/Transforms/PredicateTypeAlignment.cpp
anuragsingh-tt, line 1: The comments added in this file are pretty verbose. A good rule of thumb from @nsmithtt:
```
Comments should be answering the questions why and how, but are often not needed for what.

Good comment (answers why an otherwise subtle line of code is required):
    if (factors[i] == 0) {
      // The "Broadcast" part of inverseAndBroadcastProjectedPermutation will 0
      // fill unparticipating dims.  Promote these to 1's so that we can
      // multiply by blocking factor.
      factors[i] = 1;
    }

Bad comment (it's restating what the coding is doing in english instead of c++):
    if (factors[i] == 0) {
      // Set the factor to 1.
      factors[i] = 1;
    }

That's not to say in all cases that answering what is a bad thing, sometimes there is some really complicated code that's hard to read and understand and a what comment could go a long way.
```
Let's update with this heuristic.

lib/Dialect/TTIR/Transforms/PredicateTypeAlignment.cpp
```diff
@@ -12,6 +15,32 @@ namespace mlir::tt::ttir {
...
 
 namespace {
+
+// Create a new tensor type with the requested scalar element type. If the
+// input tensor carries a TTNN layout encoding, the layout's memref element
+// type is also updated (preserving any tile wrapping) so the encoding stays
+// consistent with the outer tensor element type. Without this, downstream
+// passes that read the memref element type from the encoding (e.g. when
+// lowering to D2M tile ops) end up with stale types.
+mlir::RankedTensorType
```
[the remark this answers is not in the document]
vwellsTT: seems fair

lib/Dialect/TTIR/Transforms/PredicateTypeAlignment.cpp
```diff
@@ -91,10 +124,29 @@ struct AlignElementwiseBinaryTypesPattern : public mlir::RewritePattern {
...
+      // Operand and result element types disagree on a single non-i1 type.
+      // This happens when a TTNN binary op carries an implicit typecast,
+      // e.g. ttnn.eq(si32, si32) -> bf16 or ttnn.gt(si32, si32) -> bf16.
+      // TTNN performs the typecast inside the kernel; D2M requires a
+      // uniformly-typed compute region. Explicate the typecast at the TTIR
+      // level by aligning operands to the result type when it is non-i1.
+      // (When the result is i1, ComparisonResultTypePattern below promotes
+      // it to lhs's element type first, so the typesAgree branch above
+      // handles that case.)
+      if (resultElemType.isInteger(1)) {
```
vwellsTT: this seems redundant w/ check in the lambda

lib/Dialect/TTIR/Transforms/PredicateTypeAlignment.cpp
```diff
@@ -91,10 +124,29 @@ struct AlignElementwiseBinaryTypesPattern : public mlir::RewritePattern {
...
+        return rewriter.notifyMatchFailure(
+            op, "operand types disagree and result is i1");
+      }
+      targetElemType = resultElemType;
```
vwellsTT: confusing to me that this is also being set inside the lambda, I feel like that should just return true/false w/o capture if we're going to set this here anyway?

lib/Dialect/TTIR/Transforms/PredicateTypeAlignment.cpp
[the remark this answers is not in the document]
sgholamiTT: Fixed — routed `PropagateUnaryTensorManipulationResultElementTypePattern` through `buildTensorWithElementType` as well.

lib/Dialect/TTIR/Transforms/PredicateTypeAlignment.cpp
sgholamiTT: Trimmed the verbose blocks down to a single why/how sentence each.

lib/Dialect/TTIR/Transforms/PredicateTypeAlignment.cpp
sgholamiTT: Hmmm, kinda agree but not fully.
The check is functionally needed (without it the code would set targetElemType = i1 when lhs/rhs disagree and the result is i1), but it really did read as redundant because both checks were spelled with the same predicate.
Refactored the whole "find target" flow. There is no more capturing lambda — agreesWith(t, target) is a pure helper. The first non-i1 type is picked deterministically with a small loop, agreement is checked with that helper, and then the target is chosen by a three-way if / else if / else over (typesAgree, result non-i1, otherwise). The i1-result fallback is now its own branch instead of an inner-if inside !typesAgree, so the redundancy goes away.

lib/Dialect/TTIR/Transforms/PredicateTypeAlignment.cpp
sgholamiTT: Refactored — replaced the capturing accumulateTarget with a pure agreesWith helper and an explicit if/else if/else that assigns the target in one place.

dlokeTT:
Before:

```
%0 = "ttir.eq"(%arg0, %arg1) : (tensor<4x4xsi32>, tensor<4x4xsi32>) -> tensor<4x4xbf16>
```

After:

```
%0 = "ttir.typecast"(%arg0) : (tensor<4x4xsi32>) -> tensor<4x4xbf16>
%1 = "ttir.typecast"(%arg1) : (tensor<4x4xsi32>) -> tensor<4x4xbf16>
%2 = "ttir.eq"(%0, %1) : (tensor<4x4xbf16>, tensor<4x4xbf16>) -> tensor<4x4xbf16>
```

For this, wouldn't it be better/make more sense to typecast the output instead of typecasting both inputs?

sgholamiTT:
@dlokeTT Good question. Actually thought about this before writing the PR.
The cast direction is intentional: TTNN's `dtype` attribute on these binary ops is the **compute** `dtype` — `ttnn.eq dtype=bf16 (si32, si32) -> bf16` actually casts the operands and runs the op in `bf16` inside the kernel, so this PR is just hoisting that into IR. 
Output-cast would work for eq/compares but the same pattern also fires on e.g. `add(si32, si32) -> bf16`, where computing in `si32` first and then casting would overflow / give different results than TTNN does today.

lib/Dialect/TTIR/Transforms/PredicateTypeAlignment.cpp
vwellsTT: okay, this reads better to me now thanks

vwellsTT (approved):

anuragsingh-tt (approved):

tapspatel (approved):

PR #8458, opened 2026-05-17
[runtime] Add scalar tensor support for TTNN kernel args

anuragsingh-tt:
Adds TTNN runtime support for scalar values mirroring TTMetal path in #8231. 
- Scalars are stored as host `UINT32` TTNN tensors inside `TTNNTensorWrapper`
- Unpacking happens only in `getScalarKernelArgAndValidate`
- Includes TTNN compile time arg verifier fix
- Includes e2e test of scalar runtime support

### DIFF
--- a/runtime/include/tt/runtime/detail/ttnn/types/types.h
+++ b/runtime/include/tt/runtime/detail/ttnn/types/types.h
@@ -32,6 +32,9 @@ using GlobalSemaphoreMapIterator = typename GlobalSemaphoreMap::iterator;
 class TTNNTensorWrapper;
 using OnDestroyTensorCallback = std::function<void(TTNNTensorWrapper *)>;
 
+using TTNNTensorWrapperPtr = std::shared_ptr<TTNNTensorWrapper>;
+using TTNNTensor = std::variant<TTNNTensorWrapperPtr, std::uint32_t>;
+
 // Wrapper for ttnn::Tensor that contains
 // additional metadata specific to our ttnn runtime
 class TTNNTensorWrapper {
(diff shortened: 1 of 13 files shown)


runtime/include/tt/runtime/detail/ttnn/types/types.h
```diff
@@ -32,6 +32,9 @@ using GlobalSemaphoreMapIterator = typename GlobalSemaphoreMap::iterator;
 class TTNNTensorWrapper;
 using OnDestroyTensorCallback = std::function<void(TTNNTensorWrapper *)>;
 
+using TTNNTensorWrapperPtr = std::shared_ptr<TTNNTensorWrapper>;
+using TTNNTensor = std::variant<TTNNTensorWrapperPtr, std::uint32_t>;
```
anuragsingh-tt: I find `using` here untidy, the alternative is to have a small wrapper class. Open to thoughts.

nsmithtt (approved):
Looks good!

1. Would be good to have an e2e test that goes through the runtime, I think the ttmetal version has an example.
2. Would like @jameszianxuTT to sign off

runtime/lib/ttnn/operations/generic/generic_op.cpp
```diff
@@ -195,6 +195,18 @@ createKernelArgs(const ::tt::target::ttnn::KernelCoreArgs &args,
               .address();
       break;
     }
+    case ::tt::target::ttnn::KernelArgType::KernelArgScalar: {
```
jameszianxuTT, line 204: Hey @anuragsingh-tt, I think guarding every current and future TTNN runtime API with "is this a scalar" is going to be a maintenance burden and am trying to think if there are any other alternatives here that can avoid that, while also avoiding massive runtime refactoring.

I believe the reason you want to unify TTNNTensorWrapper and this scalar tensor under a TTNNTensor union is because the runtime machinery relies on tensors for their inputs (tt::runtime::submit / ProgramTensorPool) and that is a fair point.

I don't see the entire picture for how these scalar tensors will eventually be used at runtime, but is it possible that we pack the scalar inputs into real [1,]-shape runtime tensors and then unpack them here when they need to be placed into the coreArgs list?

anuragsingh-tt: Updated the PR.

jameszianxuTT (approved):
I'm happy to reconsider adding scalars to ttnn runtime as more of a first class entity in future and the first commit in PR is a good reference for the roadbumps we may hit doing so. 

However, since we only need scalars for a very small subset of possible operations today, I think this tensor-wrapped-scalar is a good way to go without introducing excess maintenance burden. Thanks for making changes!

jameszianxuTT:
Also, please update PR description before merging if this is the path that everyone can agree on :)

vwellsTT (approved):

Issue #8449, opened 2026-05-15
KV cache dtype conversion does not propagate in TP models


kdimicTT:
#### Description  
When `--experimental-kv-cache-dtype=bfp_bf8` is used with TP (tensor parallel) models, compilation fails with:                                               
                                                                                                                                                               
```
  'ttnn.fill_cache' op Cache and input tensors must have the same dtype.
  Got cache dtype = bf16, input dtype = bfp_bf8
```        
In TP models the KV cache function argument flows through `ttnn.mesh_shard` before reaching `fill_cache`. TTNNKVCacheDtypeConversion updated the function argument type and inserted a typecast on the input operand, but left the mesh_shard result type as bf16. Since fill_cache receives the mesh_shard result as its cache operand the dtype mismatch caused a validation failure.                                                                                                                                                   

Single-chip models were unaffected because fill_cache's cache operand is the function argument directly.

PR #8463, opened 2026-05-18
Fix KV cache dtype conversion for TP models 

kdimicTT:
### Ticket
#8449 

### Problem description
`TTNNKVCacheDtypeConversion` did not update intermediate values between the function argument and fill_cache's cache operand, causing a dtype mismatch when mesh_shard sat between them in TP models.

### What's changed
- Added `propagateDtypeTowardRoot` to `TTNNKVCacheDtypeConversion`, which walks backward from fill_cache's cache operand through its defining ops, updating each intermediate tensor type to the target dtype. This is invoked before the existing typecast insertion so the full operand chain is consistently typed before validation.             
- Added new lit test `fill_cache_bfp8_tp.mlir` that covers the TP path, verifying that the `mesh_shard` result is retyped to the target dtype

### DIFF
--- a/lib/Dialect/TTNN/Transforms/TTNNKVCacheDtypeConversion.cpp
+++ b/lib/Dialect/TTNN/Transforms/TTNNKVCacheDtypeConversion.cpp
@@ -126,6 +126,13 @@ public:
       return;
     }
 
+    // Only propagate through mesh_shard since it is the only dtype-transparent 
+    // op expected between a kv_cache block argument and a write op in TP 
+    // models.
+    if (!mlir::isa<MeshShardOp>(value.getDefiningOp())) {
+      return;
+    }
+
     mlir::Type originalElemType = tensorType.getElementType();
     for (Value operand : value.getDefiningOp()->getOperands()) {
       auto operandType = mlir::dyn_cast<RankedTensorType>(operand.getType());
--- a/lib/Dialect/TTNN/Transforms/TTNNKVCacheDtypeConversion.cpp
+++ b/lib/Dialect/TTNN/Transforms/TTNNKVCacheDtypeConversion.cpp
@@ -105,6 +105,36 @@ public:
         mlir::FunctionType::get(ctx, newArgTypes, newResultTypes));
   }
 
+  // Walk backward from `value`, updating the element dtype of every value
+  // along the cache path until reaching the target dtype or a block argument.
+  static void propagateDtypeTowardRoot(Value value,
+                                       ttcore::DataType targetDtype) {
+    auto tensorType = mlir::dyn_cast<RankedTensorType>(value.getType());
+    if (!tensorType) {
+      return;
+    }
+
+    auto newType =
+        ttnn::utils::RankedTensorTypeFactory::create(tensorType, targetDtype);
+    if (value.getType() == newType) {
+      return;
+    }
+
+    value.setType(newType);
+
+    if (mlir::isa<BlockArgument>(value)) {
+      return;
+    }
+
+    mlir::Type originalElemType = tensorType.getElementType();
+    for (Value operand : value.getDefiningOp()->getOperands()) {
+      auto operandType = mlir::dyn_cast<RankedTensorType>(operand.getType());
+      if (operandType && operandType.getElementType() == originalElemType) {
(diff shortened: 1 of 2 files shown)


lib/Dialect/TTNN/Transforms/TTNNKVCacheDtypeConversion.cpp
```diff
@@ -105,6 +105,36 @@ class TTNNKVCacheDtypeConversionPass
         mlir::FunctionType::get(ctx, newArgTypes, newResultTypes));
   }
 
+  // Walk backward from `value`, updating the element dtype of every value
+  // along the cache path until reaching the target dtype or a block argument.
+  static void propagateDtypeTowardRoot(Value value,
```
dgolubovicTT, line 151: I would restrict propagation only through subset of operations we actually expect to have on this cache path. But if we can't propagate through chain of operations from cache argument to update/fill cache we shouldn't edit argument type nor the fill/update type...

kdimicTT: In the new commit I restricted it to only MeshShardOp, which is the only op expected on this path. I will update the comment above this method also.

lib/Dialect/TTNN/Transforms/TTNNKVCacheDtypeConversion.cpp
```diff
@@ -154,6 +184,7 @@ class TTNNKVCacheDtypeConversionPass
         llvm::TypeSwitch<Operation *>(op)
             .Case<FillCacheOp, UpdateCacheOp, PagedFillCacheOp>(
                 [&](auto cacheOp) {
+                  propagateDtypeTowardRoot(cacheOp.getCache(), dtype);
```
dgolubovicTT: So first we are calling `changeCacheArgTypes` and then we are propagating dtype `propagateDtypeTowardRoot`. What happens if we can't propagate dtype through a chain of ops?

lib/Dialect/TTNN/Transforms/TTNNKVCacheDtypeConversion.cpp
```diff
@@ -29,6 +29,39 @@ class TTNNKVCacheDtypeConversionPass
...
+      return false;
+    }
+
+    mlir::Type elemType = tensorType.getElementType();
+    for (Value operand : defOp->getOperands()) {
+      auto opType = mlir::dyn_cast<RankedTensorType>(operand.getType());
+      if (opType && opType.getElementType() == elemType) {
+        return canPropagateDtype(operand);
+      }
+    }
```
dmilinkovicTT: When we have multiple operands, we can theoretically reach multiple block arguments.

Instead, I think it would be better to perform this walk as a linked-list traversal, while ensuring that the ops on the path have only one operand.

lib/Dialect/TTNN/Transforms/TTNNKVCacheDtypeConversion.cpp
```diff
@@ -105,6 +136,37 @@ class TTNNKVCacheDtypeConversionPass
         mlir::FunctionType::get(ctx, newArgTypes, newResultTypes));
   }
 
+  // Walk backward from `value` through the chain, updating each value's element
+  // dtype. Only call this after canPropagateDtype confirms the chain is fully
+  // traversable.
+  static void propagateDtypeTowardRoot(Value value,
```
dmilinkovicTT, line 151: `canPropagateDtype` and `propagateDtypeTowardRoot` are essentially the same, with the only difference that the later one performs mutation.

I think we should try to extract the walk into a common helper.

lib/Dialect/TTNN/Transforms/TTNNKVCacheDtypeConversion.cpp
kdimicTT: I agree. Enforced to ensure there is exactly one `RankedTensorType` operand, rather than one operand total, since MeshShardOp also has a `device` operand alongside `input`. The non-tensor operand is invisible to the walk. Also tightened the termination condition so it must end at a block argument labeled as `kv_cache`.

dmilinkovicTT (approved):

PR #8466, opened 2026-05-18
Add DRAM-bandwidth based perf target to TTNNCollectPerfMetrics

odjuricicTT:
### Problem description
In order to evaluate forge model performance at scale we need to be able
to calculate top perf estimates / targets automatically.

This is intentionally the minimal slice: DRAM only, no compute FLOPs, no
op-graph traversal. It is meant to be used for LLM decode, which is the
DRAM-bandwidth-bound regime — the DRAM read time of weights + KV cache
per step is the dominant term, so a DRAM-only ceiling tracks the
measured t/s/u closely.

It's implemented as part of `TTNNCollectPerfMetrics` and the result is
appended to the existing stats the pass produces. Emitted JSON shape:

```json
"perf_targets": {
  "arch": "wormhole_b0",
  "num_chips": 1,
  "dram_bandwidth_bytes_per_sec": 288000000000,
  "params": {
    "count": 60821536,
    "memory_bytes": 121643072,
    "memory_bytes_bfp8": 62722209
  },
  "kv_cache": {
    "count": 4194304,
    "memory_bytes": 8388608,
    "memory_bytes_bfp8": 4325376
  },
  "roofline": {
    "dram_time_ms": 0.233,
    "top_perf_time_ms": 0.349,
    "top_perf_samples_per_sec": 2863.6
  }
}
```

### Calculation outline

Emits a theoretical performance ceiling under the assumption that every
forward pass is bottlenecked by reading weights and KV cache from DRAM.
The result lands in the `perf_targets` section of the existing
`TTNNCollectPerfMetrics` JSON output.

#### Process
1. Pick the entry point: the unique public forward-device function.
   Hard fail if not exactly one, if `ttcore.system_desc` is missing, or
   if the arch (currently Quasar) has no calibrated bandwidth.
2. Walk *only the forward function's argument list* (no op-graph walk).
   Each tensor arg is bucketed via existing ttcore attributes:
   - `ttcore.kv_cache` present       → KV cache bytes
   - `ttcore.argument_type != Input` → parameter bytes (Parameter / Constant / Default)
   - Input args                      → ignored (activations).
3. Normalize both totals to BFP8-equivalent storage
   (1056 B / 32×32 tile = 1.03125 B / scalar), independent of how the IR
   has encoded the arg dtype.
4. `dram_time_ms = total_bytes / dram_peak_bandwidth`,
   `top_perf_time_ms = dram_time_ms`,
   `top_perf_samples_per_sec = 1000 / top_perf_time_ms`.

#### Assumptions
- DRAM-bound regime. Compute (FLOPs vs TFLOPS), L1/SRAM, dispatch,
  kernel-launch overhead, host transfers, and inter-chip traffic are
  all ignored.
- All weights assumed to be BFP8.
- Relies on the front end labelling Parameter / Constant / kv_cache on
  args correctly.
- Peak per-arch DRAM bandwidth, hardcoded:
  - Wormhole B0 = 288 GB/s (12-channel GDDR6 aggregate)
  - Blackhole  = 512 GB/s (placeholder until measured)
  - Quasar     = uncalibrated (hard fail)

  Not pulled from `ChipDescAttr`; not derated for attainable / memory
  controller efficiency / access pattern; Galaxy effective bandwidth
  differs.
- Single-chip only. Multi-chip system_desc is a soft skip.
- Strict roofline, alpha = 1.0. `top_perf_time_ms = dram_time_ms`
  directly — no safety factor, no min(dram, compute) blend. The emitted
  number is a pure theoretical ceiling for consumers to compare
  measurements against.
- Static shapes only. Dynamic-shape args contribute zero bytes.
- One sample == one invocation. Consumers interpret accordingly
  (one token for decode, one image for vision).

#### Failure modes
- Missing `ttcore.system_desc`     → hard fail
- More than one outer forward func → hard fail
- `Arch::Quasar`                   → hard fail (no calibrated bandwidth)
- Multi-chip system_desc           → soft skip (no perf_targets emission)

### Checklist
- [x] New LIT tests cover the plain-element-type path
      (`perf_metrics_dram_target.mlir`) and the TileType path
      (`perf_metrics_dram_target_bfp8_tile.mlir`).
- [x] Smoke-tested on `llama_3_2_1b_decoder`: params count 60.8M
      (plausible for a 1-layer slice of a 1.2B model), DRAM time
      ~0.23 ms, top perf ~2864 samples/s.


lib/Dialect/TTNN/Transforms/TTNNCollectPerfMetrics.cpp
```diff
@@ -76,6 +78,71 @@ struct AggregatedMetrics {
   }
 };
 
+// First-cut DRAM-bandwidth roofline. Counts weights / KV cache from the
+// forward function arguments only (no op-graph walk) and divides by the
+// per-arch DRAM bandwidth. Top-perf time multiplies by 1.5x to approximate
+// the ~67% utilization tt-metal kernels hit in practice. Calibrated on the
```
odjuricicTT: 67% utilization that e2e models hit in practise.

lib/Dialect/TTNN/Transforms/TTNNCollectPerfMetrics.cpp
```diff
@@ -76,6 +78,71 @@ struct AggregatedMetrics {
...
+  uint64_t kvCacheCount = 0;
+  uint64_t kvCacheMemoryBytes = 0;
+  uint64_t kvCacheMemoryBytesBfp8 = 0;
+
+  double dramRooflineTimeMs = 0.0;
+  double topPerfTimeMs = 0.0;
+  double topPerfSamplesPerSec = 0.0;
+};
+
+inline uint64_t getScalarVolume(mlir::RankedTensorType tt) {
```
odjuricicTT: Does this already exist somewhere?

lib/Dialect/TTNN/Transforms/TTNNCollectPerfMetrics.cpp
```diff
@@ -262,6 +466,14 @@ class TTNNCollectPerfMetrics
         }
       }
 
+      // Single-chip DRAM-bandwidth perf target estimate. Computed once on the
+      // first matching function — if there are multiple forwards in the
+      // module we only score the first.
+      if (!perfTargetsComputed) {
+        perfTargets = computePerfTargets(module, funcOp);
+        perfTargetsComputed = true;
```
odjuricicTT: This only computes the targets for the first funcOp in this loop. We want to compute targets for the main outer level function. Is there a better and more stable way of indetifying it instead of taking the first funcOp?

odjuricicTT:
## DRAM roofline vs measured decode throughput (n150, single-chip)

This table compares the DRAM-bound performance ceiling emitted by this
PR (`perf_targets.roofline.top_perf_samples_per_sec`) against measured
n150 single-chip decode throughput for 19 LLMs. The estimate is the
strict roofline (α = 1.0): `dram_time_ms = (params + KV cache, BFP8) /
288 GB/s` and `top_perf_samples_per_sec = 1000 / dram_time_ms`. The
estimate is a *theoretical ceiling* — real workloads should approach but
not exceed it; consumers can apply their own utilization expectation
downstream.

Numbers from CI run [26226197146](https://github.com/tenstorrent/tt-xla/actions/runs/26226197146)
(workflow `Performance Benchmark`, `mlir_override=ce9b0c61424c5d849b8706fae9440162dbdcc06b`).

| Model | Actual Top Perf (t/s/u) | num_params | param_GB | Est α=1.0 (t/s/u) | (est−target)/target |
| :---- | :----: | :----: | :----: | :----: | :----: |
| Qwen 3 8B       | 21  | 8,190,735,424 | 7.867 | 32.88  | +56.6% |
| Llama 3.1 8B    | 21  | 8,030,261,312 | 7.712 | 33.65  | +60.3% |
| Ministral 8B    | 21  | 8,019,808,320 | 7.702 | 33.56  | +59.8% |
| Qwen 2.5 7B     | 24  | 7,615,616,576 | 7.314 | 36.11  | +50.5% |
| Falcon3 7B      | 24  | 7,455,550,592 | 7.161 | 36.31  | +51.3% |
| Mistral 7B      | 24  | 7,248,023,616 | 6.961 | 37.15  | +54.8% |
| Qwen 3 4B       | 42  | 4,411,424,320 | 4.237 | 59.25  | +41.1% |
| Llama 3.2 3B    | 56  | 3,606,752,320 | 3.464 | 72.70  | +29.8% |
| Qwen 2.5 3B     | 56  | 3,397,103,680 | 3.263 | 80.42  | +43.6% |
| Falcon3 3B      | 60  | 3,227,655,296 | 3.100 | 81.85  | +36.4% |
| Phi-2           | 60  | 2,779,683,856 | 2.670 | 80.93  | +34.9% |
| Qwen 3 1.7B     | 100 | 2,031,739,968 | 1.951 | 123.21 | +23.2% |
| Qwen 2.5 1.5B   | 112 | 1,777,088,064 | 1.707 | 152.13 | +35.8% |
| Falcon3 1B      | 167 | 1,669,408,896 | 1.603 | 153.41 | −8.1%  |
| Llama 3.2 1B    | 168 | 1,498,482,720 | 1.439 | 178.38 | +6.2%  |
| Phi-1.5         | 117 | 1,418,270,736 | 1.362 | 153.37 | +31.1% |
| Phi-1           | 117 | 1,418,270,736 | 1.362 | 153.37 | +31.1% |
| Qwen 3 0.6B     | 280 |   751,632,448 | 0.722 | 283.09 | +1.1%  |
| Qwen 2.5 0.5B   | 340 |   630,167,456 | 0.605 | 426.15 | +25.3% |

lib/Dialect/TTNN/Transforms/TTNNCollectPerfMetrics.cpp
```diff
@@ -76,6 +78,90 @@ struct AggregatedMetrics {
...
+  double dramRooflineTimeMs = 0.0;
+  double topPerfTimeMs = 0.0;
+  double topPerfSamplesPerSec = 0.0;
+};
+
+// Bytes per scalar element. Caller must ensure elementType is a plain
+// scalar (non-TileType); TTNN forward args reach this pass with scalar
+// element type and TTNNLayoutAttr encoding carrying the tile/dtype layout.
+inline double getBytesPerScalarElement(mlir::Type elementType) {
+  return static_cast<double>(elementType.getIntOrFloatBitWidth()) / 8.0;
```
mtopalovicTT: You don't have to use double here. Return unsigned. Same below :D

lib/Dialect/TTNN/Transforms/TTNNCollectPerfMetrics.cpp
```diff
@@ -76,6 +78,90 @@ struct AggregatedMetrics {
...
+}
+
+inline uint64_t getTensorMemoryBytes(mlir::RankedTensorType tt) {
+  if (!tt.hasStaticShape()) {
+    return 0;
+  }
+  double bytes = static_cast<double>(tt.getNumElements()) *
+                 getBytesPerScalarElement(tt.getElementType());
+  return static_cast<uint64_t>(bytes + 0.5);
```
mtopalovicTT: What is 0.5 for?

lib/Dialect/TTNN/Transforms/TTNNCollectPerfMetrics.cpp
```diff
@@ -76,6 +78,90 @@ struct AggregatedMetrics {
...
+//     emitted number is a pure theoretical ceiling for consumers to
+//     compare measurements against.
+//   - Static shapes only. Dynamic-shape args contribute zero bytes;
+//     today's decode entry points are fully static.
+//   - One sample == one invocation. `samples_per_sec` assumes a single
+//     forward call yields a single sample (one token for decode, one
+//     image for vision); consumers must interpret accordingly.
+//===----------------------------------------------------------------------===//
+struct PerfTargets {
+  std::string arch;
```
mtopalovicTT: Can you use  ttcore::Arch enum instad of string? :D

lib/Dialect/TTNN/Transforms/TTNNCollectPerfMetrics.cpp
```diff
@@ -176,6 +262,155 @@ class TTNNCollectPerfMetrics
...
+      }
+      // TTNN forward args carry a scalar element type with TTNNLayoutAttr
+      // encoding. A TileType element would mean the outer shape is in
+      // tile counts, breaking the scalarVol/byte math below.
+      assert(!mlir::isa<ttcore::TileType>(tensorType.getElementType()) &&
+             "TTNNCollectPerfMetrics: unexpected TileType element on "
+             "forward arg; expected scalar element + TTNNLayoutAttr "
+             "encoding.");
+      uint64_t scalarVol =
+          tensorType.hasStaticShape()
```
mtopalovicTT: We don't expect non static shapes. Whole compiler assumes that there are no dynamic shapes. Assert would be preferable if you really want to check it.

lib/Dialect/TTNN/Transforms/TTNNCollectPerfMetrics.cpp
```diff
@@ -176,6 +262,155 @@ class TTNNCollectPerfMetrics
...
+    auto bytesAtBfp8 = [](uint64_t count) {
+      return static_cast<uint64_t>(
+          static_cast<double>(count) * kBfp8BytesPerElement + 0.5);
+    };
+    perfTargets.paramMemoryBytesBfp8 = bytesAtBfp8(perfTargets.paramCount);
+    perfTargets.kvCacheMemoryBytesBfp8 = bytesAtBfp8(perfTargets.kvCacheCount);
+
+    uint64_t totalDramBytes =
+        perfTargets.paramMemoryBytesBfp8 + perfTargets.kvCacheMemoryBytesBfp8;
+    if (perfTargets.dramBandwidthBytesPerSec > 0) {
```
mtopalovicTT: I would assert here also.

mtopalovicTT (commented):
Is this calculation based on assumption that we would transfer whole kv cache and weights on every decode and thats why delta between est and actual is high?

lib/Dialect/TTNN/Transforms/TTNNCollectPerfMetrics.cpp
odjuricicTT: Stale round up

odjuricicTT:
> Is this calculation based on assumption that we would transfer whole kv cache and weights on every decode and thats why delta between est and actual is high?

This assumption is true for most weights and KV cache. Not for embedding weight tho, only part of these need to be read from DRAM and that is part of the delta. Another thing is that 288 GB/s dram bandwith is the hardware spec and not whan ops can achieve in practise.

@mtopalovicTT

mtopalovicTT (approved):

PR #8476, opened 2026-05-19
[CI] Install ttmlir_runtime lib

nsumrakTT:
### Problem description
_ttmlir_runtime library is built but not installed in python_packages so it is not available for use

### What's changed
added install step for ttmlir_runtime

### Checklist


runtime/python/CMakeLists.txt
```diff
@@ -96,6 +96,11 @@ if (TTMLIR_ENABLE_RUNTIME)
   )
   target_sources(_ttmlir_runtime PRIVATE $<TARGET_OBJECTS:_mlir_runtime_utils>)
   set_property(TARGET _mlir_runtime_utils PROPERTY CXX_STANDARD 20)
+
+  install(FILES $<TARGET_FILE:_ttmlir_runtime>
+    DESTINATION ${TTMLIR_PYTHON_PACKAGES_DIR}
+    COMPONENT SharedLib
+  )
```
jameszianxuTT, line 103: Sorry, can we have some context here on why this change is necessary? Who will be consuming the ttmlir runtime as a python package?

nsumrakTT: It is needed for chisel and other local debug tools.

jameszianxuTT (approved):
Approving, thanks for context. Please check if the copilot comment makes sense

PR #8485, opened 2026-05-19
[D2M] Add ScheduleDMA Support for Quasar DMA Architecture

### LINKED ISSUE #8506: [D2M] Add Support For Quasar DMA Architecture

arminaleTT:
### Ticket
Closes #8506 

### Problem description
- Quasar: 6 datamovement processors, 1 NoC. No preferred processor.
- WH/BH: 2 datamovement processors, 2 NoCs.

`ScheduleDMA` pass needs to be updated to support both cases.

### What's changed
- Refactor the `ScheduleDMA` pass and the `ThreadAttr` to centre on assigning processors to datamovement threads, not NoCs
- Modified the `ScheduleDMA` pass to support 2 or 6 datamovement processors
- Updated IR tests with the new `ThreadAttr`
- Defend in lowering passes when encountering Quasar system desc or Quasar artifacts like 6 datamovement processors

### Checklist


anuragsingh-tt (approved):

lib/Dialect/D2M/Transforms/ScheduleDMA.cpp
```diff
@@ -245,34 +277,41 @@ class D2MScheduleDMARewriter : public OpRewritePattern<GenericOp> {
...
+            writesDRAM ? ttcore::NocIndex::Noc1 : ttcore::NocIndex::Noc0);
+      } else {
+        processorIndex = 0;
+      }
+      generic.setThreadsAttr(rewriter.getArrayAttr({
+          rewriter.getAttr<ThreadAttr>(ThreadType::Datamovement, nullptr,
+                                       processorIndex),
+          generic.getThreadsAttr().getValue()[1],
+      }));
       return failure();
```
vtangTT, line 297: this should return success() no..?

lib/Dialect/D2M/Transforms/ScheduleDMA.cpp
vtangTT, line 94: for qsr we should prob relax this assertion right? not necessary in this PR just a comment

vtangTT (approved):
minor comments, otherwise lgtm ty 👍

wenbinlyuTT:
To me the `etNoCForProcessorIndex` and `getProcessorIndexForTwoNoCArch` helper and all the NoC<->Core translation feels more complicated than needed. I remember we associate BRISC with NoC1 and NCRISC with NoC0 for performance and that is set in stone for BH & WH correct? If so I think maybe we should just get rid of the concept of NoC index in the D2M dialect, and simply leave a note in `ThreadAttr` saying that the mapping is hard coded by the TTKernel dialect & D2M runtime.

nsmithtt (approved):

lib/Dialect/D2M/Transforms/ScheduleDMA.cpp
arminaleTT: the failure() means "no op matched/was rewritten". I'm pretty sure changing it to success() would cause an infinite loop unless we also somehow modify the IR so it doesn't match again

vtangTT: ah yes i remember getting confused about this before too ..

arminaleTT:
> To me the `etNoCForProcessorIndex` and `getProcessorIndexForTwoNoCArch` helper and all the NoC<->Core translation feels more complicated than needed. I remember we associate BRISC with NoC1 and NCRISC with NoC0 for performance and that is set in stone for BH & WH correct? If so I think maybe we should just get rid of the concept of NoC index in the D2M dialect, and simply leave a note in `ThreadAttr` saying that the mapping is hard coded by the TTKernel dialect & D2M runtime.

NoCIndex is a `TTCore` enum and we do need to program it for data movement kernels and fabric, so I don't want to just remove it without further investigation. Otherwise it is a fair point. I'll get rid of those helper functions and hard code the mappings

wenbinlyuTT (approved):

PR #8486, opened 2026-05-19
[d2m] mask op bufferization fix

dlokeTT:
### Problem description
d2m mask op lowering issue during bufferization because the inputs and output operands can bufferize differently depending if it comes from a view_layout or an empty

### What's changed
- Create an intermediate view_layout if bufferized input and output types don't match
- Added lit tests

### Checklist

### DIFF
--- a/lib/Dialect/D2M/IR/D2MOps.cpp
+++ b/lib/Dialect/D2M/IR/D2MOps.cpp
@@ -765,9 +765,24 @@ MaskOp::bufferize(mlir::RewriterBase &rewriter,
     return maybeOutput;
   }
 
-  rewriter.create<MaskOp>(getLoc(), maybeOutput->getType(), *maybeInput,
-                          *maybeOutput, getLogicalShapeAttr(),
-                          getFillValueAttr());
+  Value input = *maybeInput;
+  if (input.getType() != maybeOutput->getType()) {
+    auto inputType = mlir::cast<MemRefType>(input.getType());
+    auto outputType = mlir::cast<MemRefType>(maybeOutput->getType());
+    assert(inputType.hasStaticShape() && outputType.hasStaticShape());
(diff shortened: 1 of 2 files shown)


lib/Dialect/D2M/IR/D2MOps.cpp
```diff
@@ -765,9 +765,14 @@ MaskOp::bufferize(mlir::RewriterBase &rewriter,
     return maybeOutput;
   }
 
-  rewriter.create<MaskOp>(getLoc(), maybeOutput->getType(), *maybeInput,
-                          *maybeOutput, getLogicalShapeAttr(),
-                          getFillValueAttr());
+  Value input = *maybeInput;
+  if (input.getType() != maybeOutput->getType()) {
+    input =
```
vwellsTT, line 780: a view cannot resolve arbitrary type differences, only shape mismatches; thus, we should check for that explicitly imo

dlokeTT: updated check to be more stringent on shape mismatch only

anuragsingh-tt: > updated check to be more stringent on shape mismatch only

Is the volume check accomplishing this? E.g. 2x8 -> 4x4? What is the expected behavior there?

dlokeTT: Yeah, missed that. Made check even more stringent 😓

vwellsTT: I still don't like the new check 😆 
Some of these things feel like they should be asserts instead--we should always have both types, they should always be static, etc.  If there is some mismatch that's NOT a shape mismatch a view can handle, we ought to error out as well imo (though debatably I guess the MaskOp verifier could be responsible for that maybe?)

vwellsTT (approved):
thanks

anuragsingh-tt (approved):

PR #8488, opened 2026-05-20
[chisel] program-level golden accumulation

### LINKED ISSUE #7842: Chisel refactor Step 2: Single Program Flow
Add cross-op golden tensor chaining within a program. Golden outputs persist in `TensorPool` and feed into subsequent ops, so the golden execution path diverges naturally from the device path. This reveals how errors compound across a program, not just per-op.

Expands the slim `ChiselContext` from Step 1 #7841 into the full `ChiselContext`/`BinaryState`/`ProgramState` hierarchy. Adds program-level callbacks (`preProgram`/`postProgram`) and `TensorPool`. The `execute_golden()` core function gains a pool-aware wrapper (`execute_golden_from_pool()`) that pulls inputs from the golden pool and stores outputs back.

ndrakulicTT:
### Ticket
Closes #7842 

### Summary

Adds accumulated-mode golden checking to chisel: every PCC comparison now runs against a golden derived from a chain of prior op goldens (program-scoped), rather than only against an isolation golden recomputed from device inputs each op. Accumulated mode catches errors that isolation hides — e.g., a wrong layout or quantization upstream that the next op's isolation golden re-derives from "corrected" device inputs.

Isolation and accumulation modes are independently togglable via `ChiselChecksConfig(isolation=..., accumulation=...)`; by default chisel emits accumulation only. Pass `chisel.session(checks_config=ChiselChecksConfig(isolation=True))` to also run isolation.

### What's new

- **`NumericsMode`** (`isolated` | `accumulated`) tags every `NumericsPayload` / `SkippedNumericsPayload`, so downstream consumers can tell which golden a record was scored against.
- **Golden pool** lives on `ProgramState` as `_golden_tensor_pool: Dict[SSAName, GoldenMapTensor]`. It persists across ops within a program; reset per program.
- **Promotion**: in `_default_pre_op`, any input SSA not yet in the pool is seeded with the current device tensor. Emits a `golden_promoted` record (status `GOLDEN_PROMOTED`). Ideally only function args ever get promoted - intermediates promotions are an audit signal that some producer is `no_golden` or there is some other problem with ops before.
- **Eviction**: `ttnn.DeallocateOp` now has a custom `pre_op` (`_deallocate_pre_op`) that pops each input SSA from the pool and emits a `golden_evicted` record (`GOLDEN_EVICTED`).
- **Per-op handler overrides**: `ChiselOpConfig` gains `pre_op` / `post_op` callables. `run_op_callback` calls the override if set, else the default handler.
- **Renames** for clarity:
  - `ChiselOpConfig.skip_isolated_pcc` → `skip_pcc` (now skips PCC for both modes — shape/dtype still checked).

### Defaults

- `ChiselChecksConfig(isolation=False, accumulation=True)`. Accumulation is the cheaper and stricter signal for end-to-end runs; isolation is opt-in for debugging.

### DIFF
--- a/tools/chisel/chisel/callbacks.py
+++ b/tools/chisel/chisel/callbacks.py
@@ -22,10 +22,20 @@ from golden import GoldenMapTensor
 
 from .context import ChiselContext, get_instance
 from .exceptions import IrRuntimeMismatch
-from .executor import build_role_keyed_inputs, execute_golden
+from .executor import (
+    build_role_keyed_inputs,
+    execute_golden,
+    execute_golden_from_pool,
+)
 from .op_configs import ChiselOpConfig
 from .ops import SSAName, get_op_inputs, get_op_outputs
-from .report import ChiselRecord, NoGoldenPayload, SkippedNumericsPayload
+from .report import (
+    ChiselRecord,
+    GoldenPromotedPayload,
+    NoGoldenPayload,
+    NumericsMode,
+    SkippedNumericsPayload,
+)
 from .safety import chisel_safe
 from .utils import get_op_asm, retrieve_tensor
 from .validators import check_numerics, check_shape_dtype
@@ -80,20 +90,16 @@ def _validate_and_retrieve_tensor(
 
 def _run_isolation_golden(
     op: OpView,
-    mlir_outputs: List[Value],
     asm_state,
     ssa_inputs: Dict[SSAName, GoldenMapTensor],
 ) -> List[GoldenMapTensor]:
     role_inputs = build_role_keyed_inputs(op, ssa_inputs, asm_state)
-    iso_outs = execute_golden(op, role_inputs)
-    for mlir_output, iso_out in zip(mlir_outputs, iso_outs, strict=True):
-        check_shape_dtype(op, "mlir_vs_golden", mlir_output, iso_out)
-    return iso_outs
+    return execute_golden(op, role_inputs)
 
 
 @chisel_safe
 def _default_pre_op(ctx: ChiselContext, config: ChiselOpConfig) -> None:
-    """Stash a host copy of every device input so POST can drive the golden."""
+    """Stash host copies of device inputs and seed function args into the pool."""
     op = ctx.op
     if config.no_golden:
         ctx.write_record(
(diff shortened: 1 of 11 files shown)


tools/chisel/chisel/op_configs.py
```diff
@@ -3,38 +3,50 @@
...
 class ChiselOpConfig:
-    # no_golden: do not run the isolation golden; emit a NoGoldenPayload
-    # record per op so the report shows the op was visited but skipped.
-    # skip_isolated_pcc: run the isolation golden (shape/dtype still checked)
-    # but skip the PCC comparison.
+    """Per-op chisel behavior overrides.
+
+    no_golden: skip golden calculation; emit a NoGoldenPayload.
+    skip_pcc: run golden, check shape/dtype, but skip PCC checks.
+    pre_op / post_op: override default handlers; must be @chisel_safe.
```
mmilosevicTT: nit:
```suggestion
    pre_op / post_op: overrides for default handlers; must be @chisel_safe.
```

tools/chisel/chisel/callbacks.py
```diff
@@ -107,47 +113,107 @@ def _default_pre_op(ctx: ChiselContext, config: ChiselOpConfig) -> None:
 
     _assert_op_matches_runtime(ctx)
     asm_state = ctx.asm_state
+    pool = ctx.golden_tensor_pool
 
     mlir_op_inputs = get_op_inputs(op)
     for mlir_input, rt_tensor_ref in zip(mlir_op_inputs, ctx.input_refs, strict=True):
         # TODO(ndrakulic): Right now we are pulling input device tensors to host potentially multiple times
```
mmilosevicTT, line 110: Do we still need this TODO?

tools/chisel/chisel/callbacks.py
```diff
@@ -107,47 +113,107 @@ def _default_pre_op(ctx: ChiselContext, config: ChiselOpConfig) -> None:
...
-    iso_outs = _run_isolation_golden(op, mlir_op_outputs, asm_state, ctx.stashed_inputs)
+    if ctx.checks_config.isolation:
+        iso_outs = _run_isolation_golden(op, asm_state, ctx.stashed_inputs)
+    else:
+        iso_outs = [None] * len(mlir_op_outputs)
+
+    if ctx.checks_config.accumulation:
+        accum_outs = execute_golden_from_pool(op, ctx.golden_tensor_pool, asm_state)
+    else:
+        accum_outs = [None] * len(mlir_op_outputs)
```
mmilosevicTT, line 176: Why don't we have just `outs`? We are splitting here then zipping in the for just under. But the zip actually only uses one of these, since config cannot be both at the same time.

mmilosevicTT (approved):
LGTM

tools/chisel/chisel/callbacks.py
ndrakulicTT: Config can be both at same time, it would take longer but you can have both metric for isolated and accumulated PCC without problem

tools/chisel/chisel/callbacks.py
ndrakulicTT: Yeah, I will have one of the following PRs, to introduce device_tensor_pool, to store tensors in post_op so we dont need to fetch them in pre_op if they are not function arguments.

tools/chisel/chisel/callbacks.py
mmilosevicTT: Mhm, ok. Could we then just reorder the inputs in one of the methods (either execute_golden_from_pool or _run_isolation_golden) just to be in the same order?

Also could we make a helper method for this, as this seems like a repeated code. Just pass the method we want to run with parameters we want to use

ndrakulicTT: Yeah good catched, removed the `_run_isolation_golden` and moved it to `executor.py` as `execute_golden_with_ssa_inputs` as `execute_golden_from_pool` can also utlize that function. Left the logic for creating None list as is, it didnt fit into abstracted function properly

mmilosevicTT (approved):
LGTM

PR #8489, opened 2026-05-20
[Optimizer:Matmul] Matmul program config fixes

rpavlovicTT:
### Changes

1. MatmulProgramConfig: guard 2D mcast on grid > 1x1.

2. MatmulRules: filter sharded output hints when no MatmulProgramConfig would be emitted (batched matmuls).

### DIFF
--- a/lib/Dialect/TTNN/Analysis/MatmulProgramConfig.cpp
+++ b/lib/Dialect/TTNN/Analysis/MatmulProgramConfig.cpp
@@ -278,9 +278,21 @@ generateMatmulProgramConfig(Operation *op, TTNNLayoutAttr outputLayout) {
   }
 
   if (outputMemLayout == TensorMemoryLayout::BlockSharded) {
-    return generateMatmul2DProgramConfig(ctx, Mt, Nt, Kt, outputLayout,
-                                         fusedActivation, maxSubblockSize,
-                                         fuseBatch);
+    // 2D mcast requires both grid dims > 1. On a degenerate grid (one dim == 1)
+    // fall back to 1D mcast: tt-metal treats BlockSharded + 1D-grid as an
+    // intentional alias for WidthSharded/HeightSharded (see
+    // ttnn/cpp/ttnn/operations/matmul/device/matmul_device_operation.cpp
+    // ~L197), and the 1D mcast direction must follow which grid dim is
+    // degenerate, not outputMemLayout (otherwise mcast_in0 picks the wrong
+    // direction).
+    auto [gridX, gridY] = utils::getPhysicalGridDimensions(outputLayout);
+    if (gridX > 1 && gridY > 1) {
+      return generateMatmul2DProgramConfig(ctx, Mt, Nt, Kt, outputLayout,
+                                           fusedActivation, maxSubblockSize,
+                                           fuseBatch);
+    }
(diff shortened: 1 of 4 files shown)


bmalesevicTT (approved):
Could you add a bit more context in the PR description for these two cases: where they were hit, how they manifested (accuracy issue, runtime crash, metal assert, etc.)? That would help future maintenance, especially across tt-metal uplifts.
Also I suppose there is a way to address these directly in tt-metal, to file them as issues and eventually remove the special casings from here?

rpavlovicTT:
> Could you add a bit more context in the PR description for these two cases: where they were hit, how they manifested (accuracy issue, runtime crash, metal assert, etc.)? That would help future maintenance, especially across tt-metal uplifts. Also I suppose there is a way to address these directly in tt-metal, to file them as issues and eventually remove the special casings from here?

Apologies for auto merging it. Hopefully comments in the code provide some context.  
As for tt-metal, possibly we can guard such cases there with extra validation. Today obviously we can pass program config that is misconfigured and causes a hang like the case with GPT OSS I was observing. Therefore we must be careful with its generation in tt-mlir, but there are always edge cases like these. Long term we should rely on tt-metal generating program config for us.

bmalesevicTT:
> > Could you add a bit more context in the PR description for these two cases: where they were hit, how they manifested (accuracy issue, runtime crash, metal assert, etc.)? That would help future maintenance, especially across tt-metal uplifts. Also I suppose there is a way to address these directly in tt-metal, to file them as issues and eventually remove the special casings from here?
> 
> Apologies for auto merging it. Hopefully comments in the code provide some context. As for tt-metal, possibly we can guard such cases there with extra validation. Today obviously we can pass program config that is misconfigured and causes a hang like the case with GPT OSS I was observing. Therefore we must be careful with its generation in tt-mlir, but there are always edge cases like these. Long term we should rely on tt-metal generating program config for us.

Thanks, the code comments do explain the divergence in tt-metal behavior vs. compiler-produced IR. What I was asking for was the concrete observed examples of these two problematic cases: for each one, which model hit it and what was the observed failure mode? From your reply, it’s still not clear whether both were GPT-OSS hangs, or only one

rpavlovicTT:
> > > Could you add a bit more context in the PR description for these two cases: where they were hit, how they manifested (accuracy issue, runtime crash, metal assert, etc.)? That would help future maintenance, especially across tt-metal uplifts. Also I suppose there is a way to address these directly in tt-metal, to file them as issues and eventually remove the special casings from here?
> > 
> > 
> > Apologies for auto merging it. Hopefully comments in the code provide some context. As for tt-metal, possibly we can guard such cases there with extra validation. Today obviously we can pass program config that is misconfigured and causes a hang like the case with GPT OSS I was observing. Therefore we must be careful with its generation in tt-mlir, but there are always edge cases like these. Long term we should rely on tt-metal generating program config for us.
> 
> Thanks, the code comments do explain the divergence in tt-metal behavior vs. compiler-produced IR. What I was asking for was the concrete observed examples of these two problematic cases: for each one, which model hit it and what was the observed failure mode? From your reply, it’s still not clear whether both were GPT-OSS hangs, or only one

Both were causing hangs in gpt oss 20B on LB.
- batched matmul has specific program config rules which we don't follow in compiler, so we have to discard sharded outputs there. Luckily those mammals are not frequent. 
- creating MM2DMcast program config hung the chip when compute grid was [1, 1].

PR #8491, opened 2026-05-20
Add ttir/ttnn composite op and TTNNResolveComposites pass

mtopalovicTT:
Add ttir/ttnn composite op and TTNNResolveComposites pass
Composite op carries operations from upstream dialects that lack a typed
op in TTIR but may have one in a downstream dialect.  It also captures
fused patterns in TTIR where backend executability is uncertain.

TTNNResolveComposites attempts to promote a composite to its typed TTNN
op (validated through OpModel); when promotion is not possible the
decomposition body is inlined as a fallback.


lib/Dialect/TTNN/Transforms/TTNNResolveComposites.cpp
```diff
@@ -0,0 +1,232 @@
...
+  void runOnOperation() final {
+    registerBuiltinComposites();
+
+    ModuleOp moduleOp = getOperation();
+    llvm::DenseSet<func::FuncOp> decompositionFuncsToDelete;
+    SmallVector<CompositeOp> compositeOps;
+
+    moduleOp.walk([&](CompositeOp op) { compositeOps.push_back(op); });
+
+    for (CompositeOp compositeOp : compositeOps) {
```
dmilinkovicTT: nit: since we're separately tracking ops to delete, we could remove `compositeOps` and move loop body to walk lambda

lib/Dialect/TTNN/Transforms/TTNNResolveComposites.cpp
```diff
@@ -0,0 +1,232 @@
...
+// Try to create the typed op and validate it. Returns the created op on
+// success, or nullptr if verification/validation fails.
+// Requires OpModel — without it, always returns nullptr (forcing
+// decomposition).
+static Operation *tryCreateTypedOp(CompositeOp compositeOp,
+                                   OpBuilder &builder) {
+#ifndef TTMLIR_ENABLE_OPMODEL
+  (void)compositeOp;
+  (void)builder;
+  return nullptr;
```
dmilinkovicTT: Should we maybe use `OpValidator` here instead of `OpModel`, like we do for the fusing validation?

lib/Dialect/TTNN/Transforms/TTNNResolveComposites.cpp
```diff
@@ -0,0 +1,232 @@
...
+static void registerBuiltinComposites() {
+  auto &registry = getCompositeRegistry();
+  if (!registry.empty()) {
+    return;
+  }
+
+  registry["topk_router_gpt"] = [](CompositeOp compositeOp,
+                                   OpBuilder &builder) -> Operation * {
+    if (compositeOp.getInputs().size() != 3) {
+      return nullptr;
```
dmilinkovicTT: If a composite op fails any of the checks specified here, would it make more sense to assert-and-fail instead of silently fail?

After all, a composite carrying the identifier but violating the contract is a red flag which should be explicitly analyzed, instead of just falling back to inlining.

lib/Dialect/TTNN/Pipelines/TTNNPipelines.cpp
```diff
@@ -186,6 +186,9 @@ void createTTNNPipelineLoweringPasses(OpPassManager &pm,
   pm.addPass(createTTNNLayout());
   // Add pass to convert TTIR to TTNN.
   pm.addPass(createConvertTTIRToTTNNPass());
+  // Resolve composite ops into typed ops (with OpModel) or inline their
+  // decompositions.
+  pm.addPass(createTTNNResolveComposites());
```
dmilinkovicTT, line 191: Would it make more sense to move the logic of this pass into `ConvertTTIRToTTNNPass`?

On one hand, it would clutter the conversion pass with a non-trivial lowering pattern.

But on the other hand, we could remove `ttnn.composite` op from the dialect, and prevent any further misuses which could come from adding logic (passes) between TTIR -> TTNN conversion and composites resolution pass.

mtopalovicTT: Yea sure that makes more sense.

include/ttmlir/Dialect/TTIR/IR/TTIROps.td
```diff
@@ -6500,4 +6500,42 @@ def TTIR_MoeExpertTokenRemapOp : TTIR_NamedOp<"moe_expert_token_remap"> {
     let hasVerifier = 1;
 }
 
+//===----------------------------------------------------------------------===//
+// TTIR Composite op
+//===----------------------------------------------------------------------===//
+
+def TTIR_CompositeOp : TTIR_NamedOp<"composite", [TTCore_NonCacheableTrait]> {
```
sdjordjevicTT: Maybe we should create a TTCore composite op, so we don't have two definitions in TTIR/TTNN, as they seem the same?

lib/Dialect/TTNN/Pipelines/TTNNPipelines.cpp
mtopalovicTT: Actually that would be a bit trickier than I initially thought. Since library that compiles validation source is linking against library that contains all conversion logic i.e `MLIRTTNNTransforms` -> `TTMLIRTTIRToTTNN`. I would have to extract validation to separate library which would be included by both of these libs in order to achieve this.

mtopalovicTT: Actually this is also not possible because OpValidator.cpp includes Passes.h for workarounds and validation and fallback, so this TU is basically tightly coupled with MLIRTTNNTransforms....

include/ttmlir/Dialect/TTIR/IR/TTIROps.td
mtopalovicTT: Let me think it through. The only problem that is obvious to me is layouting pass matches ops which implement TTIROpsInterface and inserts to_layout ops if necessary...

lib/Dialect/TTNN/Pipelines/TTNNPipelines.cpp
sdjordjevicTT: I was thinking about this as well. Maybe it is better to stay in the realm of TTNN. To remove TTNN composite and potentially TTIR composite, I was referring to converting the composite to the TTCore op.

include/ttmlir/Dialect/TTIR/IR/TTIROps.td
sdjordjevicTT: And what about just passing the TTIR composite down in the TTNN dialect?

mtopalovicTT: Got it working with ttcore.

sdjordjevicTT: Nice!

sdjordjevicTT (approved):
Nice! Thanks!

PR #8492, opened 2026-05-20
[StableHLO] Fix sharded reshape compilation failure

### LINKED ISSUE #3643: Reshape ops with dimension splits sometimes fail when input is sharded
### Description

When a stablehlo.reshape op splits/merges dimensions on a sharded tensor, the StableHLO pipeline sometimes fails with a shape verification error. This happens because Shardy's conservative propagation cannot propagate sharding through the reshape's dimension split/merge, leaving the output unsharded. Later, `UpdateGlobalToLocalShapes` converts the input to its local (per-device) shape but leaves the output at its global shape, causing an element count mismatch that fails the StableHLO reshape verifier, e.g.:
```
loc("reshape.30"): error: number of output elements (30720) doesn't match expected number of elements (7680)
```

The failure chain involves multiple passes in the StableHLO pipeline:

1. `RegisterCustomShardingRule` - Reshape has no custom rule; Shardy's built-in rule is used. The built-in rule (`/opt/ttmlir-toolchain/src/shardy/.../op_sharding_rule_registry.cc:995`) correctly decomposes the reshape into factors (e.g., `([i, jk]) → ([l, j, k])` for a dimension split.

2. `UserPriorityPropagation` - Runs with `conservativePropagation = true`. This was intentionally set in https://github.com/tenstorrent/tt-mlir/commit/7a1f4689887ead0b3abfda6aea1688f7aa5b3dcb. Conservative mode blocks sub-axis propagation (`basic_factor_propagation.cc:125`). When the mesh axis size doesn't evenly divide the leftmost factor of the split (e.g., axis=4 doesn't divide factor j=6), Shardy cannot place the sharding on any output dimension and leaves the reshape output unsharded.

3. `InsertExplicitReshards` - Calls `getShardings(op->getResults())` which returns an empty array when the output has no sharding annotation. This causes `getMesh()` to return `nullopt`-> no all_gather is inserted. (Note: this part I'm not 100% sure it is true, but it doesn't matter whether all_gather is inserted or not, because input gets annotated and output doesn't, so the pipeline will still fail during `UpdateGlobalToLocalShapes` (step 5)).

4. `WrapUnderManualComputation` - Wraps everything in `sdy.manual_computation`.

5. `UpdateGlobalToLocalShapes` - Converts shapes inside the manual computation block. The reshape's input gets its type updated to the local shape (e.g., `1x7680`), but the output has no sharding annotation so it stays at the global shape (`1x6x5120`). The StableHLO verifier then rejects the reshape: e.g. 7680 != 30720.

### Example

- Reshape: `tensor<1x30720xbf16> → tensor<1x6x5120xbf16>`
- Mesh: `<["_axis_0"=4]>` (4 devices)
- Input sharding: dim 1 sharded on `"_axis_0"`

The built-in reshape rule produces:
```
([i, jk]) → ([l, j, k])  where i=1, j=6, k=5120, l=1
```

Factor `j=6` is shared between input dim 1 and output dim 1.
Factor `k=5120` is shared between input dim 1 and output dim 2.

Input dim 1 has axis `"_axis_0"=4`. This maps to the combined factor `jk=30720`.
To propagate to the output, Shardy would need to place axis 4 on one of the sub-factors:
- Factor `j=6`: 6 is not divisible by 4 -> blocked
- Factor `k=5120`: 5120 is divisible by 4, but placing it here requires a sub-axis split

With `conservativePropagation=true`, sub-axis splits are blocked (`basic_factor_propagation.cc:125`), so b no output dimension gets sharded.

#### What happens in the following passes?

| Pipeline Stage | Input shape | Output shape | Status |
|---|---|---|---|
| After propagation | 1x30720 (sharded on dim 1) | 1x6x5120 (unsharded) | Inconsistent but still global shapes |
| After `InsertExplicitReshards` | 1x30720 | 1x6x5120 | Skipped -> no all_gather |
| After `UpdateGlobalToLocalShapes` | 1x7680 | 1x6x5120 (unchanged) | Compilation fails 7680 != 30720 |

### Expected behavior

The pipeline should insert an all_gather before the reshape.

### Other ops

LLM-assisted investigation into other stablehlo ops suggests reshape is probably the only operation that may hit this issue.

### LINKED ISSUE #4767: [Krea Realtime 14B] Transformer - error: number of output elements (92160) doesn't match expected number of elements (23040)
### Describe the bug

- Transformer fails with `error: number of output elements (92160) doesn't match expected number of elements (23040)` - [may19_krea_transformer_4_chips.log](https://github.com/user-attachments/files/28005737/may19_krea_transformer_4_chips.log)
- Issue originates from [flatten](https://huggingface.co/krea/krea-realtime-video/blob/main/transformer/causal_model.py#L1085) - [may19_reshape_shard_issue.log](https://github.com/user-attachments/files/28005783/may19_reshape_shard_issue.log), [test_reshape_shard.py](https://github.com/user-attachments/files/28005784/test_reshape_shard.py)

### Repro

```
git checkout kkannan/may19_krea_transformer_reshape_shard_bug
git submodule update --init --recursive

export TT_VISIBLE_DEVICES=0,1
pytest tests/torch/models/krea_realtime/test_transformer.py -svv
pytest tests/torch/ops/test_reshape_shard.py -svv
```

kamalrajkannan78:
### Ticket

- closes https://github.com/tenstorrent/tt-xla/issues/3643
- closes https://github.com/tenstorrent/tt-xla/issues/4767

### Problem description

- Shardy's `UserPriorityPropagation` (run with `conservativePropagation=true`) cannot place an output sharding when a `stablehlo.reshape` splits a sharded flat dim into `(N, D)` where `N` is indivisible by the shard count. 
- The reshape result is left unannotated; `UpdateGlobalToLocalShapes` then resizes the operand to per-shard but keeps the result global, so the StableHLO verifier rejects it (e.g. `92160 vs 23040` for `(3, 30720) -> (3, 6, 5120)` on 4 chips).

### What's changed

- New `insert-reshards-for-unpropagated-ops` pass, slotted after `ReplicateNonSplittableConstants` and before `InsertExplicitReshards`. For every `stablehlo.reshape` whose result Shardy left without an `sdy.sharding`, it inserts `sdy.reshard <@mesh, [{}, ...]>` for each genuinely sharded operand.
- The reshard lowers to `all_gather` via `ReshardToCollectives`, so the consumer sees replicated operands and per-shard shapes match.
- Consolidated helpers into `ShardyUtils` (`getClosedReplicatedTensorSdyShardingAttr`, mesh-aware overload of `isFullyReplicatedTensor`, `createIfMissing` parameter on `getOperandShardingAttr`) and deduped the closed-replicated builder shared with `ReplicateNonSplittableConstants`.

### Checklist

- [x] Verify the changes through local testing in N150

### Logs

- [jun1_reshape_shard_before_fix.log](https://github.com/user-attachments/files/28470406/jun1_reshape_shard_before_fix.log)
- [jun1_reshape_shard_after_fix.log](https://github.com/user-attachments/files/28470405/jun1_reshape_shard_after_fix.log)
- [jun1_krea_transformer_before_fix.log](https://github.com/user-attachments/files/28470404/jun1_krea_transformer_before_fix.log)
- [jun1_krea_transformer_after_fix.log](https://github.com/user-attachments/files/28470398/jun1_krea_transformer_after_fix.log)

### DIFF
--- a/lib/Dialect/StableHLO/Transforms/InsertReshardsForUnpropagatedOps.cpp
+++ b/lib/Dialect/StableHLO/Transforms/InsertReshardsForUnpropagatedOps.cpp
@@ -1,0 +1,107 @@
+// SPDX-FileCopyrightText: (c) 2026 Tenstorrent AI ULC
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#include "ttmlir/Dialect/StableHLO/Transforms/Passes.h"
+#include "ttmlir/Dialect/StableHLO/Utils/ShardyUtils.h"
+
+#include "shardy/dialect/sdy/ir/dialect.h"
+#include "stablehlo/dialect/StablehloOps.h"
+
+namespace mlir::tt::stablehlo {
+#define GEN_PASS_DEF_INSERTRESHARDSFORUNPROPAGATEDOPSPASS
+#include "ttmlir/Dialect/StableHLO/Transforms/Passes.h.inc"
+
+static bool anyResultHasSharding(mlir::Operation *op) {
+  auto spv = op->getAttrOfType<mlir::sdy::TensorShardingPerValueAttr>(
+      mlir::sdy::TensorShardingAttr::name);
+  return spv && !spv.getShardings().empty();
+}
+
+static void replicateOperandViaReshard(mlir::OpBuilder &builder,
+                                       mlir::Operation *op,
+                                       mlir::OpOperand &operand,
+                                       mlir::StringAttr meshName) {
+  auto tensorType =
+      mlir::dyn_cast<mlir::RankedTensorType>(operand.get().getType());
+  if (!tensorType) {
+    return;
+  }
+
+  mlir::sdy::TensorShardingAttr replicated =
+      shardy_utils::getClosedReplicatedTensorSdyShardingAttr(
+          builder.getContext(), meshName, tensorType.getRank());
+
+  builder.setInsertionPoint(op);
+  auto reshard = builder.create<mlir::sdy::ReshardOp>(
+      op->getLoc(), operand.get(), replicated);
+  operand.set(reshard.getResult());
+}
+
+// For ops whose results were left unsharded by Shardy propagation, reshard
+// any sharded operand to fully replicated. The reshard later lowers to an
+// all_gather, so the consumer sees replicated operands and per-shard shapes
+// match in UpdateGlobalToLocalShapes. See tt-xla#3643.
+class InsertReshardsForUnpropagatedOpsPass
+    : public impl::InsertReshardsForUnpropagatedOpsPassBase<
+          InsertReshardsForUnpropagatedOpsPass> {
+public:
+  using impl::InsertReshardsForUnpropagatedOpsPassBase<
+      InsertReshardsForUnpropagatedOpsPass>::
+      InsertReshardsForUnpropagatedOpsPassBase;
+
+  void runOnOperation() final {
+    mlir::ModuleOp rootModule = getOperation();
+    MLIRContext *context = rootModule.getContext();
+
+    if (shardy_utils::isGraphSolved(rootModule)) {
+      return;
+    }
+
+    llvm::SmallVector<mlir::sdy::MeshOp> meshOps =
+        shardy_utils::getMeshOps(rootModule);
+    if (meshOps.empty()) {
+      return;
+    }
+    if (meshOps.size() > 1) {
+      rootModule.emitError("Pass currently only supports a single shardy mesh "
+                           "op in the module");
+      signalPassFailure();
+      return;
+    }
+    mlir::sdy::MeshOp globalMeshOp = meshOps[0];
+    mlir::StringAttr meshName = globalMeshOp.getSymNameAttr();
+
+    mlir::OpBuilder builder(context);
+
+    rootModule.walk([&](mlir::Operation *op) {
+      // Restricted to stablehlo.reshape — the only op known to hit this
+      // propagation gap. Other ops have custom sharding rules registered.
+      if (!mlir::isa<mlir::stablehlo::ReshapeOp>(op)) {
+        return;
+      }
+      if (op->getParentOfType<mlir::sdy::ManualComputationOp>()) {
+        return;
+      }
+      if (anyResultHasSharding(op)) {
+        return;
+      }
+
+      for (mlir::OpOperand &operand : op->getOpOperands()) {
+        if (!mlir::isa<mlir::RankedTensorType>(operand.get().getType())) {
+          continue;
+        }
+        mlir::sdy::TensorShardingAttr operandSharding =
+            shardy_utils::getOperandShardingAttr(operand, globalMeshOp,
+                                                 /*createIfMissing=*/false);
+        if (!operandSharding || shardy_utils::isFullyReplicatedTensor(
+                                    operandSharding, globalMeshOp)) {
+          continue;
+        }
+        replicateOperandViaReshard(builder, op, operand, meshName);
+      }
+    });
+  }
+};
+
+} // namespace mlir::tt::stablehlo
--- a/include/ttmlir/Dialect/StableHLO/Transforms/Passes.td
+++ b/include/ttmlir/Dialect/StableHLO/Transforms/Passes.td
@@ -555,6 +555,20 @@ def InsertExplicitReshardsPass : Pass<"insert-explicit-reshards", "::mlir::Modul
   ];
 }
 
+def InsertReshardsForUnpropagatedOpsPass : Pass<"insert-reshards-for-unpropagated-ops", "::mlir::ModuleOp">
+{
+  let summary = "Insert sdy.reshard to replicate operands of ops Shardy left unsharded.";
+  let description = [{
+    For ops whose results have no sdy.sharding after propagation, reshard
+    any sharded operand to fully replicated. Closes a gap that breaks
+    UpdateGlobalToLocalShapes.
+  }];
+
+  let dependentDialects = [
+    "::mlir::sdy::SdyDialect"
+  ];
+}
+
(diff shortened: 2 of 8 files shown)


saiarthiraguram (approved):

acicovicTT:
Is it possible to solve this without adding a whole new pass? It seems like a lot of overhead for an edge case that affects only one op.

kamalrajkannan78:
> Is it possible to solve this without adding a whole new pass? It seems like a lot of overhead for an edge case that affects only one op.

@acicovicTT — quick context first: this approach was the planned fix that @sdjukicTT documented [here](https://github.com/tenstorrent/tt-xla/issues/3643#issuecomment-4251634870) (Apr 15), and I [confirmed](https://github.com/tenstorrent/tt-xla/issues/3643#issuecomment-4486628416) it was still current with @sdjukicTT  on May 19 before starting ([reply](https://github.com/tenstorrent/tt-xla/issues/3643#issuecomment-4486709262)).

On the "no new pass" angle — the fix has to run after [UserPriorityPropagation](https://github.com/tenstorrent/tt-mlir/blob/69eb692924c5cdec8974692ca9c2b2b375bbe847/lib/Dialect/StableHLO/Pipelines/StableHLOPipelines.cpp#L92) and before [WrapUnderManualComputation](https://github.com/tenstorrent/tt-mlir/blob/69eb692924c5cdec8974692ca9c2b2b375bbe847/lib/Dialect/StableHLO/Pipelines/StableHLOPipelines.cpp#L113), so the only existing pass it could fold into is [ReplicateNonSplittableConstantsPass](https://github.com/tenstorrent/tt-mlir/blob/69eb692924c5cdec8974692ca9c2b2b375bbe847/lib/Dialect/StableHLO/Pipelines/StableHLOPipelines.cpp#L100). That's mechanically possible — same runOnOperation + rootModule.walk shape — but it would mean putting reshape sharding logic inside a pass named for constants. Future debugging won't grep for it there.

The new pass is ~50 lines of logic + ~30 lines of pass boilerplate. The 50 lines exist in either approach — only the 30 are the "new pass" tax, and they buy a clearly-named, grep-able place for this fix. Happy to fold into ReplicateNonSplittableConstants if you'd prefer that trade — let me know.

lib/Dialect/StableHLO/Transforms/InsertReshardsForUnpropagatedOps.cpp
```diff
@@ -0,0 +1,206 @@
...
+#include "shardy/dialect/sdy/ir/dialect.h"
+#include "stablehlo/dialect/StablehloOps.h"
+
+namespace mlir::tt::stablehlo {
+#define GEN_PASS_DEF_INSERTRESHARDSFORUNPROPAGATEDOPSPASS
+#include "ttmlir/Dialect/StableHLO/Transforms/Passes.h.inc"
+
+static mlir::sdy::TensorShardingAttr
+buildReplicatedSharding(MLIRContext *context, mlir::StringAttr meshName,
+                        int64_t rank) {
```
sdjukicTT: Can `buildReplicatedSharding` be moved to `ShardyUtils`, next to `getDefaultTensorSdyShardingAttr` which does a similar thing? `ReplicateNonSplittableConstants.cpp:111-120` has this same logic, we could deduplicate it.

lib/Dialect/StableHLO/Transforms/InsertReshardsForUnpropagatedOps.cpp
```diff
@@ -0,0 +1,206 @@
...
+}
+
+static bool anyResultHasSharding(mlir::Operation *op) {
+  auto spv = op->getAttrOfType<mlir::sdy::TensorShardingPerValueAttr>(
+      mlir::sdy::TensorShardingAttr::name);
+  return spv && !spv.getShardings().empty();
+}
+
+// Stricter than shardy_utils::isFullyReplicatedTensor: also returns true when
+// the annotation lists axes that all resolve to unit-sized mesh dims.
```
sdjukicTT: Can this be folded into `isFullyReplicatedTensor`?

lib/Dialect/StableHLO/Transforms/InsertReshardsForUnpropagatedOps.cpp
```diff
@@ -0,0 +1,206 @@
...
+      }
+      if (axisSize > 1) {
+        return false;
+      }
+    }
+  }
+  return true;
+}
+
+static bool insideManualComputation(mlir::Operation *op) {
```
sdjukicTT: Can this method be removed and inlined at call site with:
`if (op->getParentOfType<mlir::sdy::ManualComputationOp>())`?
Matches how parent checks are written elsewhere

lib/Dialect/StableHLO/Transforms/InsertReshardsForUnpropagatedOps.cpp
```diff
@@ -0,0 +1,206 @@
...
+    if (mlir::isa<mlir::sdy::ManualComputationOp>(p)) {
+      return true;
+    }
+  }
+  return false;
+}
+
+// Read-only equivalent of shardy_utils::getOperandShardingAttr. The latter
+// mutates the IR by adding default replicated annotations to unannotated
+// func args, which corrupts otherwise single-device side-graphs.
```
sdjukicTT: Can `shardy_utils::getOperandShardingAttr` be made read-only instead, or changed to take a `createIfMissing` parameter? Then `readOperandShardingNoMutate` could be dropped

sdjukicTT (commented):
I think it's ok for this to be a separate pass.
Few suggestions on moving helpers to `ShardyUtils` and modifying and reusing existing `ShardyUtils` helpers where possible, to make `InsertReshardsForUnpropagatedOps.cpp` shorter and cleaner.

sdjukicTT (approved):

PR #8499, opened 2026-05-20
[StableHLOToTTIR] Avoid silent bf16 rounding for integer-typed gather

sonalibaskaran2499:
### Ticket
[Link to Github Issue](https://github.com/tenstorrent/tt-metal/issues/44735)

### Problem description
Integer-typed `stablehlo.gather` was being lowered to `ttir.embedding` by `StableHLOGatherToEmbeddingPattern`. The TTNN embedding kernel only accepts a BFLOAT16 weight (`TT_FATAL(weights.dtype() == DataType::BFLOAT16, ...)` in `embedding_device_operation.cpp`), so the operand workaround in `createEmbeddingOpOperandsWorkarounds` unconditionally casts the weight operand to bf16. bf16 has only a 7-bit mantissa, so integer values in the 1024–2048 range step by 8 — values like 1900 round to 1904 and 2204 rounds to 2208. The cast happens after the device-level dtype is committed, so the rounding is invisible to higher layers.

This surfaced in Qwen 2.5 VL's vision encoder:
```
cu_seqlens = torch.repeat_interleave(
grid_thw[:, 1] * grid_thw[:, 2], grid_thw[:, 0]
).cumsum(dim=0, dtype=torch.int32)
```
which lowers to a `stablehlo.gather` over an i64 operand. The silently
rounded `cu_seqlens` then broke the downstream `torch.split(qkv, lengths.tolist(), dim=2)`:
`RuntimeError: split_with_sizes expects split_sizes to sum exactly to 2204(input tensor's size at dimension 2), but got split_sizes=[2208]`

### What's changed
1. **`StableHLOGatherToEmbeddingPattern::checkBasicLegality`** now early-exits with `notifyMatchFailure` when the operand element type is an integer. Integer gathers never reach the lossy bf16-cast workaround.

2. **`StableHLOGatherToGatherDimPattern`** is extended to also cover the non-batched integer case. Previously it required both operand and start_indices batching dims to be present. The check is replaced with:
   - "batching-dim presence must match between operand and start_indices"
   - "if neither has batching dims, only fire when the operand is integer-typed" — non-batched bf16/float gathers continue to take the embedding pattern → `ttnn.embedding` path, which is the faster lowering for vocab lookups.
The rest of the pattern (reshape `start_indices` to drop `index_vector_dim`, optionally unsqueeze + broadcast to match output rank, emit `ttir.gather`) is unchanged and reused for both flavors.

`ttir.gather` lowers to `ttnn.gather` directly with no bf16 cast, so integer precision is preserved end-to-end.

### Checklist

### DIFF
--- a/lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
+++ b/lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
@@ -4668,6 +4668,20 @@ class StableHLOGatherToEmbeddingPattern
     auto sliceSizes = srcOp.getSliceSizes();
     auto startIndexMap = dimensionNumbers.getStartIndexMap();
 
+    // ttir.embedding casts its weight to bf16, whose 8-bit mantissa
+    // exactly represents integers only in [-256, 256] (e.g. 1900 -> 1904).
+    // Wider integer / index operands silently round, so reject them here.
+    if (auto intTy = mlir::dyn_cast<mlir::IntegerType>(
+            srcOp.getOperand().getType().getElementType())) {
+      if (intTy.getWidth() > 8) {
+        return rewriter.notifyMatchFailure(
+            srcOp, "unsupported integer operand element type");
+      }
+    } else if (mlir::isa<mlir::IndexType>(
+                   srcOp.getOperand().getType().getElementType())) {
+      return rewriter.notifyMatchFailure(srcOp, "index-typed operand");
+    }
+
     // Check if start indices tensor isn't 1D when we are indexing multiple
     // dimensions because of matmul restrictions.
     if (startIndexMap.size() > 1 && startIndicesShape.size() == 1) {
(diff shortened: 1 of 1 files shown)


anuragsingh-tt (changes requested):
The comments added in these file are pretty verbose. A good rule of thumb from @nsmithtt:

```
Comments should be answering the questions why and how, but are often not needed for what.

Good comment (answers why an otherwise subtle line of code is required):
    if (factors[i] == 0) {
      // The "Broadcast" part of inverseAndBroadcastProjectedPermutation will 0
      // fill unparticipating dims.  Promote these to 1's so that we can
      // multiply by blocking factor.
      factors[i] = 1;
    }

Bad comment (it's restating what the coding is doing in english instead of c++):
    if (factors[i] == 0) {
      // Set the factor to 1.
      factors[i] = 1;
    }

That's not to say in all cases that answering what is a bad thing, sometimes there is some really complicated code that's hard to read and understand and a what comment could go a long way.
```

Let's update with this heuristic.

lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
```diff
@@ -4668,6 +4668,43 @@ class StableHLOGatherToEmbeddingPattern
...
+    // safe to round-trip through bf16. Refuse only the wider integer types
+    // (i16/u16/i32/u32/i64/u64/index) so the gather is routed through a
+    // non-lossy lowering (ttir.gather / slice / concat) or fails compilation
+    // loudly instead of returning wrong values.
+    if (auto intTy = mlir::dyn_cast<mlir::IntegerType>(
+            srcOp.getOperand().getType().getElementType())) {
+      if (intTy.getWidth() > 8) {
+        return rewriter.notifyMatchFailure(
+            srcOp,
+            "Refusing to lower wide integer-typed gather (bitwidth > 8) to "
```
anuragsingh-tt: My preference is to not bubble up workarounds in error messages like this.

sonalibaskaran2499: @anuragsingh-tt I have addressed your review comments.

lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
```diff
@@ -5419,7 +5443,85 @@ class StableHLOGatherToGatherDimPattern
       }
     }
 
-    // Step 3: Create gather_dim with the appropriate dimension.
+    // Workarounds for two tt-metal ttnn.gather bugs that fire when an
```
anuragsingh-tt, line 5444: Do we have tt-metal issues to track these workarounds? [If so, please link them here as per the guidelines. ](https://docs.tenstorrent.com/tt-mlir/coding-guidelines.html#comments)

anuragsingh-tt (approved):

lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
sonalibaskaran2499: @anuragsingh-tt Yes I have filed tt-metal issue. They're now linked inline in the workaround comment.

PR #8501, opened 2026-05-20
Fix #8496 - incorrect arg conversion in SHLO->TTIR conversion for ttnn.sparse_matmul

### LINKED ISSUE #8496: [EmitPy] sparse_matmul emits nnz=0, which the runtime rejects
Sub-issue of #8494.

## Context

Hit while bringing up codegen_py for the DeepSeek V3.2 decode graph on a 4×8 Blackhole galaxy. Reproduces against the standalone `tt-alchemist generate-python` driver on the captured `ttnn_<ts>.mlir` snapshot — no tt-xla / hardware required. See parent for the full emit recipe and attached IR.

## Evidence

Emitted call (three occurrences in this graph):

```python
ttnn.sparse_matmul(..., nnz=0, ...)
```

`sparse_matmul_device_operation.cpp:121`:

```cpp
TT_FATAL(
    operation_attributes.nnz.value_or(1) > 0,
    "nnz ({}) must be greater than 0", operation_attributes.nnz.value());
```

`nnz=0` from Python becomes `optional<uint32_t>{0}`, whose `.value_or(1)` evaluates to 0 — the fatal fires. Omitting the kwarg yields `nullopt`, which the runtime treats as "infer from the sparsity tensor at runtime", matching the semantics of MLIR's `nnz = 0 : i64`.

## Proposed fix

When MLIR's `nnz` is 0, emit either a positive value or omit the kwarg entirely; do not emit `nnz=0`.

svuckovicTT:
### Ticket
Fixes #8496

### Problem description
Arg `nnz` of `ttnn.sparse_matmul` op was incorrectly set to `0` and then hotfixed in runtime to be `std::nullopt`. This caused issues with emitpy.

### What's changed
IR is the source of truth now. Fix SHLO->TTIR conversion, added a `nnz > 0` verifier case for the TTNN op, properly set value in flatbuffer.

Added a negative test and fixed existing tests to not include `nnz = 0`.

Edit: add verifier for TTIR op as well, fix tests.

### Checklist


lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
```diff
@@ -8935,7 +8935,10 @@ class StableHLOToTTIRSparseMatmulOpConversionPattern
     BoolAttr isInputASparseAttr = rewriter.getBoolAttr(isInputASparse);
     BoolAttr isInputBSparseAttr = rewriter.getBoolAttr(isInputBSparse);
 
-    // Parse nnz attribute (optional)
+    // Parse nnz attribute (optional).
+    // In SHLO world, nnz=0 means "infer from sparsity tensor at runtime"
+    // In TTIR world (and TTNN), nnz=0 means "this matmul is dense"
```
sdjordjevicTT, line 8940: Let's not introduce the TTNN limitation on the TTIR level. Let's accept 0 in TTIR, and handle this during TTIR to TTNN conversion, where the limitation is actually introduced.

mtopalovicTT (approved):

lib/Conversion/StableHLOToTTIR/StableHLOToTTIRPatterns.cpp
svuckovicTT: I gave this some thought initially when implementing it this way, and these were my conclusions:
`nnz=0` does carry information (tensor is all-zeroes) - I could see a scenario where this would be useful, e.g. use only the default expert and none of the sparse ones. TTIR should be able to express this as well. The fact that SHLO doesn't seems wrong.

Thoughts?

sdjordjevicTT: Makes sense, thanks for explanation!

sdjordjevicTT (approved):

acolicTT (approved):

acicovicTT (approved):

PR #8503, opened 2026-05-20
[d2m] embedding dram support

dlokeTT:
### Problem description
Need to support dram as input and output memspace for embedding op

### What's changed
- `TTIRToD2M` embedding now supports DRAM defaults for input/output memspace
- Keeps indices in L1 for the actual d2m.embedding
- `d2m.indexed_row_copy` now accepts a DRAM source and dest memref for embedding table
- `D2MToTTKernel` generates correct noc address based on source memref and accounts for noc alignment reqs
- Fix tiled to untiled layout shape handing in lower to layout
-  Added builder and lit tests
### Checklist

### DIFF
--- a/lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
+++ b/lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
@@ -2113,21 +2113,33 @@ static Value getCollapsedOutputRow(OpBuilder &rewriter, Location loc,
   return row;
 }
 
-static Value buildMappedL1NocAddr(OpBuilder &rewriter, Location loc, Value base,
-                                  AffineMap memoryMap,
-                                  ValueRange logicalIndices,
-                                  ttcore::ChipDescAttr chipDesc) {
+static Value buildMappedNocAddr(OpBuilder &rewriter, Location loc, Value base,
+                                AffineMap memoryMap, ValueRange logicalIndices,
+                                ttcore::ChipDescAttr chipDesc,
+                                ttcore::MemorySpace memorySpace) {
   SmallVector<Value> mappedIndices =
       d2m::utils::applyMap(rewriter, loc, memoryMap, logicalIndices, true);
   return buildNocAddress(rewriter, loc, base, mappedIndices, chipDesc,
-                         ttcore::MemorySpace::DeviceL1);
+                         memorySpace);
 }
 
-static Value loadI32FromL1PacketThroughScratch(OpBuilder &rewriter,
-                                               Location loc, Value scratchCb,
-                                               Value srcNocAddr, Value laneI32,
-                                               Value transferSizeBytes,
-                                               Value onePage) {
+static int64_t getNocReadAlignmentBytes(ttcore::ChipDescAttr chipDesc,
+                                        ttcore::MemorySpace memorySpace) {
+  switch (memorySpace) {
+  case ttcore::MemorySpace::DeviceL1:
+    return chipDesc.getNocL1AddressAlignBytes();
+  case ttcore::MemorySpace::DeviceDRAM:
+    return chipDesc.getNocDRAMAddressAlignBytes();
+  default:
+    llvm_unreachable("Unsupported indexed row copy memory space");
+  }
+}
+
+static Value loadI32FromNocPacketThroughScratch(OpBuilder &rewriter,
+                                                Location loc, Value scratchCb,
+                                                Value srcNocAddr, Value laneI32,
+                                                Value transferSizeBytes,
+                                                Value onePage) {
   rewriter.create<ttkernel::CBReserveBackOp>(loc, scratchCb, onePage);
   Value writePtr = rewriter.create<ttkernel::GetWritePtrOp>(loc, scratchCb);
   rewriter.create<ttkernel::NocAsyncReadOp>(loc, srcNocAddr, writePtr,
@@ -2144,10 +2156,10 @@ static Value loadI32FromL1PacketThroughScratch(OpBuilder &rewriter,
   return value;
 }
 
-static void copyL1ToL1ThroughScratch(OpBuilder &rewriter, Location loc,
-                                     Value scratchCb, Value srcNocAddr,
-                                     Value dstNocAddr, Value transferSizeBytes,
-                                     Value onePage) {
+static void copyNocToNocThroughScratch(OpBuilder &rewriter, Location loc,
+                                       Value scratchCb, Value srcNocAddr,
+                                       Value dstNocAddr,
+                                       Value transferSizeBytes, Value onePage) {
   rewriter.create<ttkernel::CBReserveBackOp>(loc, scratchCb, onePage);
   Value writePtr = rewriter.create<ttkernel::GetWritePtrOp>(loc, scratchCb);
   rewriter.create<ttkernel::NocAsyncReadOp>(loc, srcNocAddr, writePtr,
@@ -2392,12 +2404,17 @@ public:
     auto srcType = mlir::cast<MemRefType>(op.getSrc().getType());
     auto dstType = mlir::cast<MemRefType>(op.getDst().getType());
 
+    ttcore::MemorySpace srcMemorySpace = ttcore::getMemorySpace(srcType);
     if (ttcore::getMemorySpace(indicesType) != ttcore::MemorySpace::DeviceL1 ||
-        ttcore::getMemorySpace(srcType) != ttcore::MemorySpace::DeviceL1 ||
         ttcore::getMemorySpace(dstType) != ttcore::MemorySpace::DeviceL1) {
       return rewriter.notifyMatchFailure(
-          op, "indexed row copy currently supports L1 indices, source, and "
+          op, "indexed row copy currently supports L1 indices and "
               "destination");
+    }
+    if (srcMemorySpace != ttcore::MemorySpace::DeviceL1 &&
+        srcMemorySpace != ttcore::MemorySpace::DeviceDRAM) {
+      return rewriter.notifyMatchFailure(
+          op, "indexed row copy currently supports L1 or DRAM source");
     }
     if (!hasCollapsed2DShard(indicesType) || !hasCollapsed2DShard(srcType) ||
         !hasCollapsed2DShard(dstType)) {
@@ -2427,20 +2444,31 @@ public:
     AffineMap dstMemoryMap =
         d2m::utils::getMemoryMap(device, op.getDst(), true);
 
-    constexpr int64_t kIndexTransferSizeBytes = 16;
-    constexpr int64_t kMaxRowTransferSizeBytes = 16;
+    int64_t indexElementSizeBytes =
+        ttcore::getElementSizeBytes(indicesType.getElementType());
+    int64_t indexTransferSizeBytes =
+        getNocReadAlignmentBytes(chipDesc, ttcore::MemorySpace::DeviceL1);
+    TT_assert(indexTransferSizeBytes > 0);
+    TT_assert(indexTransferSizeBytes % indexElementSizeBytes == 0);
+    int64_t indexElementsPerTransfer =
+        indexTransferSizeBytes / indexElementSizeBytes;
+
     int64_t rowElementSizeBytes =
         ttcore::getElementSizeBytes(srcType.getElementType());
-    int64_t rowElementsPerTransfer =
-        kMaxRowTransferSizeBytes / rowElementSizeBytes;
+    int64_t rowTransferSizeBytes =
+        getNocReadAlignmentBytes(chipDesc, srcMemorySpace);
+    TT_assert(rowTransferSizeBytes > 0);
+    TT_assert(rowTransferSizeBytes % rowElementSizeBytes == 0);
+    int64_t rowElementsPerTransfer = rowTransferSizeBytes / rowElementSizeBytes;
 
     Value onePage = i32(rewriter, loc, 1);
-    Value indexTransferSizeBytes = i32(rewriter, loc, kIndexTransferSizeBytes);
-    Value rowElementSizeBytesValue = i32(rewriter, loc, rowElementSizeBytes);
-    Value maxRowElementsPerTransfer =
+    Value indexTransferSizeBytesValue =
+        i32(rewriter, loc, indexTransferSizeBytes);
+    Value rowTransferSizeBytesValue = i32(rewriter, loc, rowTransferSizeBytes);
+    Value rowElementsPerTransferValue =
         index(rewriter, loc, rowElementsPerTransfer);
-    Value fourIndex = index(rewriter, loc, 4);
-    Value rowStep = index(rewriter, loc, rowElementsPerTransfer);
+    Value indexElementsPerTransferValue =
+        index(rewriter, loc, indexElementsPerTransfer);
     Value oneIndex = index(rewriter, loc, 1);
     Value myY = rewriter.create<ttkernel::MyLogicalYOp>(loc);
     Value myX = rewriter.create<ttkernel::MyLogicalXOp>(loc);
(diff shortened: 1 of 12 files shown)


lib/Dialect/D2M/Transforms/LowerToLayout/Plan.cpp
```diff
@@ -241,9 +241,14 @@ modifyDeviceType(MLIRContext *ctx, RankedTensorType baseType,
...
 
   ArrayRef<int64_t> tileShape;
-  if (mlir::isa<ttcore::TileType>(elementType)) {
-    tileShape =
-        newTileShape.value_or(ttcore::getTensorTileShapeOrEmpty(baseType));
+  if (auto tileType = mlir::dyn_cast<ttcore::TileType>(elementType)) {
+    if (newTileShape) {
+      tileShape = *newTileShape;
+    } else if (ttcore::isTiled(baseType)) {
+      tileShape = ttcore::getTensorTileShape(baseType);
```
wenbinlyuTT, line 269: For tiled-to-untiled change we end up in this branch correct? But this seems to assign a non-empty tile shape? The code is working so I guess I'm misunderstanding something here.

lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
```diff
@@ -3270,16 +3264,60 @@ class D2MEmbeddingOpRewriter : public OpConversionPattern<ttir::EmbeddingOp>,
...
+      } else {
+        layout = ttcore::MetalLayoutAttr::get(
+            rewriter.getContext(), logicalShape, memSpace,
+            ttcore::TensorMemoryLayout::Sharded);
+      }
+
+      llvm::SmallVector<int64_t> tileShape;
+      llvm::SmallVector<int64_t> unshardedShape =
+          layout.getPhysicalShape(tileShape);
+      llvm::SmallVector<int64_t> simpleGrid(unshardedShape.size(), 1);
```
wenbinlyuTT: Could you help me understand why is this a column and why its height is the rank of the physical device shape?

lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
```diff
@@ -3270,16 +3264,60 @@ class D2MEmbeddingOpRewriter : public OpConversionPattern<ttir::EmbeddingOp>,
...
+        memorySpaces[0] == ttcore::MemorySpace::DeviceL1
+            ? boundaryInputs[0]
+            : createUntiledLayoutOp(boundaryInputs[0], indicesType,
+                                    ttcore::MemorySpace::DeviceL1);
+    SmallVector<Value> inputs{indicesInput, boundaryInputs[1]};
     SmallVector<Value> origOutputs =
         createDpsOutputs(loc, rewriter, {resultType});
     SmallVector<Value> outputs{
-        createOptimalLayoutOp(origOutputs[0], memorySpaces[1],
+        createOptimalLayoutOp(origOutputs[0], ttcore::MemorySpace::DeviceL1,
```
wenbinlyuTT: A comment here explaining why the actual embedding output must be in L1 would be helpful.

lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
```diff
@@ -2113,21 +2113,33 @@ static Value getCollapsedOutputRow(OpBuilder &rewriter, Location loc,
...
-                         ttcore::MemorySpace::DeviceL1);
+                         memorySpace);
 }
 
-static Value loadI32FromL1PacketThroughScratch(OpBuilder &rewriter,
-                                               Location loc, Value scratchCb,
-                                               Value srcNocAddr, Value laneI32,
-                                               Value transferSizeBytes,
-                                               Value onePage) {
+static int64_t getNocReadAlignmentBytes(ttcore::ChipDescAttr chipDesc,
```
wenbinlyuTT: This function and the various use of its output below should be promoted as util functions like `getNocElementAlignmentL1()`, we will need more and more of these queries.

lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
```diff
@@ -2480,59 +2508,48 @@ class D2MIndexedRowCopyRewriter
...
           rewriter, loc, indicesShape, i, getCollapsed2DPhysicalRows(dstType));
-      auto columnLoop =
-          rewriter.create<scf::ForOp>(loc, startColumn, endColumn, rowStep);
+      auto columnLoop = rewriter.create<scf::ForOp>(
+          loc, startColumn, endColumn, rowElementsPerTransferValue);
       {
         OpBuilder::InsertionGuard columnGuard(rewriter);
         rewriter.setInsertionPointToStart(columnLoop.getBody());
         Value j = columnLoop.getInductionVar();
-        Value remainingElements =
```
vtangTT, line 2511: why don't we need to handle tail elements anymore?

lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
```diff
@@ -2392,13 +2404,18 @@ class D2MIndexedRowCopyRewriter
...
+    ttcore::MemorySpace srcMemorySpace = ttcore::getMemorySpace(srcType);
     if (ttcore::getMemorySpace(indicesType) != ttcore::MemorySpace::DeviceL1 ||
-        ttcore::getMemorySpace(srcType) != ttcore::MemorySpace::DeviceL1 ||
         ttcore::getMemorySpace(dstType) != ttcore::MemorySpace::DeviceL1) {
       return rewriter.notifyMatchFailure(
-          op, "indexed row copy currently supports L1 indices, source, and "
+          op, "indexed row copy currently supports L1 indices and "
               "destination");
     }
+    if (srcMemorySpace != ttcore::MemorySpace::DeviceL1 &&
```
vtangTT: maybe should merge this check into the one above? smthn like: `indexed row copy only supports L1/DRAM src w/ L1 dst`

lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
```diff
@@ -3270,16 +3264,60 @@ class D2MEmbeddingOpRewriter : public OpConversionPattern<ttir::EmbeddingOp>,
     }
 
     Location loc = op.getLoc();
-    SmallVector<Value> inputs{
+    auto createUntiledLayoutOp = [&](Value value, RankedTensorType logicalType,
```
vtangTT: can we extend `createOptimalLayoutOp` instead?

lib/Dialect/D2M/Utils/Utils.cpp
```diff
@@ -556,6 +556,16 @@ getGridMapsFromVirtualGridMapping(Value val, ArrayRef<int64_t> gridShape) {
   if (gridFwdMap.getNumDims() != rank || gridFwdMap.getNumResults() != 3) {
     return std::nullopt;
   }
+
+  SmallVector<int64_t> physicalGridShape =
+      ttmlir::utils::evalShape(gridFwdMap, gridShape);
+  ArrayRef<int64_t> physicalWorkerGridShape(physicalGridShape);
+  if (physicalGridShape.size() != 3 ||
```
vtangTT: isn't this already checked on L556?

test/python/golden/d2m/test_dram_ops.py
```diff
@@ -347,3 +348,68 @@ def test_dram_reduction(
         atol=_reduction_atol(reduce_type, shape, dim_arg, dtype),
         pcc=0.99,
     )
+
+
+DRAM_EMBEDDING_CASES = [
```
vtangTT, line 353: should we test case where we aren't aligned?

lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
dlokeTT: Tail elements are handled by layout padding before this lowering. The TTKernel loop copies fixed size noc packets across the shard, so we're fine as long as the shard width is noc aligned. I added an invariant checking this as well to address copilots comment. 

We still don't handle case where the read is non noc aligned, that needs to be handled separately since we need to load that into l1 and then have the risc do it which is a lot more work. We'll eventually need this for decode for full llms but for the embedding model I'm looking at, we don't need it so I left it to get minimal changes in.

Especially since this all might get reworked later.

lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
dlokeTT: done, also added support for dst to be straight to dram. put more info in reply to Wenbin

lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
dlokeTT: yea, updated

lib/Dialect/D2M/Utils/Utils.cpp
dlokeTT: oops yea, removed

test/python/golden/d2m/test_dram_ops.py
dlokeTT: added

lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
dlokeTT: Yeah, the variable name was a bit misleading here. This isn't describing a column of the physical device grid. simpleGrid is the placeholder tensor grid used by the initial ToLayoutOp; its length is the rank of the physical tensor shape returned by layout.getPhysicalShape(...), and each entry is 1 because real grid selection happens later in D2MGridSelection. For ND tensors we still need placeholder virtual-grid maps attached so downstream code has a valid mapping before grid selection rewrites it. I updated the helper comment to make that clearer and variable naming.

lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
dlokeTT: It technically doesn't. Noc can write to dram but I did it this way since those tensors are usually used by compute later and I thought it made sense to do write to l1 and then remote store to dram. But I think it actually makes more sense to let allocator pick this memspace later on so reworked to support any memspace.

lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
dlokeTT: done

lib/Conversion/D2MToTTKernel/D2MToTTKernel.cpp
```diff
@@ -2480,59 +2508,48 @@ class D2MIndexedRowCopyRewriter
...
 
       Value srcRowIndex = rewriter.create<arith::IndexCastOp>(
           loc, rewriter.getIndexType(), indexValue);
       Value outputRow = getCollapsedOutputRow(
           rewriter, loc, indicesShape, i, getCollapsed2DPhysicalRows(dstType));
-      auto columnLoop =
-          rewriter.create<scf::ForOp>(loc, startColumn, endColumn, rowStep);
+      auto columnLoop = rewriter.create<scf::ForOp>(
+          loc, startColumn, endColumn, rowElementsPerTransferValue);
       {
```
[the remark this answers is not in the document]
dlokeTT: added checks for this

lib/Dialect/D2M/Transforms/LowerToLayout/Plan.cpp
dlokeTT: For tiled-to-untiled, we should not enter the TileType branch. In that path elementType is the new scalar element type from newElementType, so dyn_cast<TileType>(elementType) fails and tileShape remains empty. The branch only applies when the target element type is still tiled. So a non-empty tile shape there means we’re constructing a tiled target layout, not the untiled side of the conversion.

dlokeTT: doesnt seem to be needed after adding output to dram directly support so removing

dlokeTT: nvm, its needed

wenbinlyuTT (approved):
Looks pretty clean now!

vtangTT (approved):

lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
```diff
@@ -359,9 +360,12 @@ class D2MNamedRewriterCommon {
...
 
     auto tensorType = mlir::cast<mlir::RankedTensorType>(value.getType());
-    ArrayRef<int64_t> logicalShape = tensorType.getShape();
+    if (!logicalType) {
+      logicalType = tensorType;
+    }
+    ArrayRef<int64_t> logicalShape = logicalType.getShape();
 
-    Type elementType = tensorType.getElementType();
+    Type elementType = logicalType.getElementType();
```
vwellsTT, line 370: not following these changes, why do you want logicalType != tensorType case?

lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
```diff
@@ -3270,12 +3269,19 @@ class D2MEmbeddingOpRewriter : public OpConversionPattern<ttir::EmbeddingOp>,
...
+    SmallVector<Value> boundaryInputs{
         createOptimalLayoutOp(op.getInput(), memorySpaces[0],
                               /*tiled=*/false, /*noCollapse=*/false, rewriter),
         createOptimalLayoutOp(op.getWeight(), memorySpaces[0],
                               /*tiled=*/false, /*noCollapse=*/false, rewriter)};
-
+    Value indicesInput =
+        memorySpaces[0] == ttcore::MemorySpace::DeviceL1
+            ? boundaryInputs[0]
+            : createOptimalLayoutOp(
```
vwellsTT, line 3283: can we just create this inline earlier vs replacing on the fly here?

lib/Dialect/D2M/Transforms/LowerToLayout/Plan.cpp
vwellsTT: not following why this needed, pls explain

lib/Dialect/D2M/Utils/Utils.cpp
```diff
@@ -556,6 +556,15 @@ getGridMapsFromVirtualGridMapping(Value val, ArrayRef<int64_t> gridShape) {
   if (gridFwdMap.getNumDims() != rank || gridFwdMap.getNumResults() != 3) {
     return std::nullopt;
   }
+
+  SmallVector<int64_t> physicalGridShape =
```
vwellsTT, line 560: this block of change for sure needs a comment, not following it

lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
dlokeTT: This is needed when we re-layout an already layedout value but want the new layout to preserve the original logical tensor shape. In the DRAM-input embedding path, %indices is first converted to the boundary input memspace, which may produce a tensor whose type shape is the physical/layouted shape. Then the actual embedding kernel needs indices in L1, so we insert a second to_layout from that boundary value to L1. If that second call used value.getType() directly, it would treat the physical tensor shape as the logical shape. Passing indicesType keeps the L1 staging layout’s logical shape equal to the original TTIR indices shape.

lib/Conversion/TTIRToD2M/TTIRToD2M.cpp
dlokeTT: yea done,

lib/Dialect/D2M/Utils/Utils.cpp
dlokeTT: added but basically this validates that the virtual grid forward map actually preserves worker count when collapsed back to a physical 2d worker grid. The earlier check only verified map arity (rank dims and 3 results). it didn't prove the mapped physical worker shape covered the same number of workers as the original virtual grid which would drop or duplicate workers

lib/Dialect/D2M/Transforms/LowerToLayout/Plan.cpp
dlokeTT: If the result is untiled, we should not pass a tile shape. This change makes sure we only set tileShape when the result is actually tiled. Before, the code could accidentally reuse the input’s tile shape, which would inflate the shape causing a mismatch

lib/Dialect/D2M/Utils/Utils.cpp
vwellsTT: makes sense, thanks

vwellsTT (approved):

PR #8518, opened 2026-05-20
OpTPathParity tests

pljubisavljevicTT:
### Ticket
Link to Github Issue

### Problem description
The way `TTNNOpInvokeLib` is currently implemented, it's easy to introduce bugs when building the NativeTable type.

### What's changed
Added regression tests that check the NativeTable type built with `buildOpTFromMLIR`(OpModel) matches the one built with `createOp + UnPackTo` (runtime). This way, if the two paths ever go out of sync, the tests will catch it.

### Checklist

### DIFF
--- a/test/unittests/MLIRAttrToFBNative/OpTPathParity.cpp
+++ b/test/unittests/MLIRAttrToFBNative/OpTPathParity.cpp
@@ -1,0 +1,1376 @@
+// SPDX-FileCopyrightText: (c) 2026 Tenstorrent AI ULC
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#include "OpTPathParity.h"
+
+#ifdef TTMLIR_ENABLE_OPMODEL
+#include "ttmlir/OpInvoke/TTNN/utils/utils.h"
+#include "ttmlir/OpModel/TTNN/NativeFromMLIR.h"
+#include "ttmlir/OpModel/TTNN/TTNNOpModel.h"
+#include "ttmlir/Target/TTNN/TTNNToFlatbuffer.h"
+
+void compareOutputTensorRefT(
+    const std::unique_ptr<::tt::target::ttnn::TensorRefT> &opNativeOpModelOut,
+    const std::unique_ptr<::tt::target::ttnn::TensorRefT> &opNativeFBOut) {
+  if (!opNativeOpModelOut && !opNativeFBOut) {
+    return;
+  }
+  ASSERT_TRUE(opNativeOpModelOut && opNativeFBOut);
+  EXPECT_EQ(ttnn_op_invoke::operations::utils::getTensorRefMemoryConfig(
+                *opNativeOpModelOut),
+            ttnn_op_invoke::operations::utils::getTensorRefMemoryConfig(
+                *opNativeFBOut));
+  EXPECT_EQ(ttnn_op_invoke::operations::utils::getDataType(*opNativeOpModelOut),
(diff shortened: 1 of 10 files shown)


test/unittests/MLIRAttrToFBNative/OpTPathParity.h
```diff
@@ -0,0 +1,117 @@
...
+
+template <typename OpT>
+inline mlir::tt::ttnn::TTNNLayoutAttr resolveOutputLayout(OpT op) {
+  return mlir::cast<mlir::tt::ttnn::TTNNLayoutAttr>(
+      mlir::cast<mlir::RankedTensorType>(op.getResult().getType())
+          .getEncoding());
+}
+
+inline void
+resetOutputTensorRefT(std::unique_ptr<::tt::target::ttnn::TensorRefT> &out) {
```
rpavlovicTT: Instead of resetting all these fields, can we reset whole out tensor and then compare only `memory_config` and `dtype`, as: 

```cpp

  // Drop the whole TensorRef from the OpT comparison — the two build paths
  // populate it differently (thin vs serialized) and that's legitimate.
  opTOpModel.out.reset();
  opTFB.out.reset();
  EXPECT_EQ(opTOpModel, opTFB);   // compares all scalars/configs, no out noise

  // Compare `out` through exactly the functions the lib consumes it with:
  EXPECT_EQ(getTensorRefMemoryConfig(*originalOpModelOut),
            getTensorRefMemoryConfig(*originalFBOut));
  EXPECT_EQ(getDataType(*originalOpModelOut), getDataType(*originalFBOut));
```

test/unittests/MLIRAttrToFBNative/OpTPathParity.cpp
```diff
@@ -0,0 +1,1510 @@
...
+createOp(::mlir::tt::FlatbufferObjectCache &cache, MatmulOp op);
+} // namespace mlir::tt::ttnn
+
+namespace mlir::tt::ttnn::op_model {
+#ifdef TTMLIR_ENABLE_OPMODEL
+::tt::target::ttnn::MatmulOpT buildMatmulOpTFromMLIR(
+    bool transposeA, bool transposeB, std::optional<llvm::StringRef> activation,
+    std::optional<mlir::Attribute> programConfigAttr,
+    std::optional<DeviceComputeKernelConfigAttr> computeKernelConfig,
+    TTNNLayoutAttr outputLayout);
```
rpavlovicTT: Can we avoid forward declarations like this one? Instead we can move all of them in a single header that is included in OpModel and in the test here.

test/unittests/MLIRAttrToFBNative/OpTPathParity.cpp
```diff
@@ -0,0 +1,1510 @@
...
+INSTANTIATE_TEST_SUITE_P(MatmulOpTPathParityTest, MatmulOpTPathParityTest,
+                         ::testing::ValuesIn(matmulOpList));
+
+//===----------------------------------------------------------------------===//
+// Conv2dOpTPathParity
+//===----------------------------------------------------------------------===//
+
+namespace mlir::tt::ttnn {
+::flatbuffers::Offset<::tt::target::ttnn::Conv2dOp>
+createOp(::mlir::tt::FlatbufferObjectCache &cache, Conv2dOp op);
```
rpavlovicTT: Same comment as for buildOpFromMlir function forward declarations.

PR #8524, opened 2026-05-20
Fix stale sys.modules cache in PythonModelRunner::loadModule

amilovanovicTT:
### Ticket
N/A
### Problem description
When a PyTorch model compiled with the EmitPy codegen option has a graph break, dynamo emits one FX subgraph per region, and the runner produces one codegen artifact per subgraph under `<export_path>/graph_N/`, each containing a file named `main.py`. At execution time, `PythonModelRunner` adds the graph's directory to `sys.path` and imports `"main"`. Because Python's import machinery caches modules by name in the process‑wide `sys.modules`, the second graph's import main hits the cache from the first graph and silently returns graph_0's module instead of loading graph_1's source from disk. The runtime then calls `graph_0.forward` with graph_1's input list, which can raise `IndexError: list index out of range`. 
### What's changed
`PythonModelRunner::loadModule` now evicts stale entries from `sys.modules` before importing the requested module. After eviction, the normal import proceeds and is forced to read the fresh sources from disk.

### Checklist

### DIFF
--- a/tools/tt-alchemist/csrc/python_runner/python_runner.cpp
+++ b/tools/tt-alchemist/csrc/python_runner/python_runner.cpp
@@ -8,6 +8,9 @@
 
 #include <Python.h>
 
+#include <filesystem>
+#include <system_error>
+
 #pragma clang diagnostic push
 #pragma clang diagnostic ignored "-Wcast-qual"
 #pragma clang diagnostic ignored "-Wglobal-constructors"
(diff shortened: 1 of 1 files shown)


tools/tt-alchemist/csrc/python_runner/python_runner.cpp
```diff
@@ -135,9 +138,75 @@ void PythonModelRunner::addToSysPath(const std::string &path) {
...
+      std::error_code statEc;
+      if (entry.is_regular_file(statEc) && !statEc &&
+          entry.path().extension() == ".py") {
+        std::string stem = entry.path().stem().string();
+        if (stem != moduleName) {
+          tryDeleteDictKey(sysModulesRaw, stem.c_str());
+        }
+      }
+      dirIt.increment(ec);
+    }
```
svuckovicTT, line 207: Not sure I'm following what's happening here - why is
```py
  tryDeleteDictKey(sysModulesRaw, moduleName.c_str());
```
not enough?

amilovanovicTT: This is to ensure that the other .py files apart from `main.py` in the generated solution (e.g., utils.py) are properly evicted as well before their new import, since those files are always named the same regardless of which `<execute_path>/graph_N` folder they belong to, and some of those files might differ per graph (e.g., `consteval.py` when file split is turned on).

svuckovicTT (approved):

PR #8526, opened 2026-05-21
[L1SpillManagement] Remove ttnn.output_l1_usage IR attribute and address review comments

rpavlovicTT:
## Summary

- Drop the temporary `ttnn.output_l1_usage` IR attribute. L1/DRAM distinction in `L1SpillManagement` is now derived directly from the result type's `TTNNLayoutAttr` via `hasL1BufferType()`. The cached L1-size fallback was a no-op (attribute was only set when `result.outputL1Usage > 0`, which `validate()` already returns).
- Raise outer `maxLegalLayouts` default 8 → 64 to match the inner pass default. This ensures the reshard candidate generator gets enough material to populate all three sharding-type buckets (block / height / width) up to the `maxReshardCandidatesPerType=4` cap (=12 total).
- Address remaining review comments on #7686: drop "Belady" naming in favor of "farthest-last-use eviction", rename test files (`belady_spill` → `l1_dram_demotion`, `beam_spill_with_reshards` → `l1_dram_demotion_fork`), simplify 5 helper signatures (`ensureFitsL1`, `handleNoFit`, `handleFragmentation`, `handleOOM` lose `attrOutputL1Usage` param), remove redundant initial `validate()` in `handleOOM`, assert instead of silent-continue on null/OOB result layouts, add debug log on L1 reshard insertion, and use `demoteToDram` (not `spillToDram`) in stage-3 OOM since `evictUntil` already drained the live set.

### DIFF
--- a/lib/Dialect/TTNN/Analysis/L1SpillManagement.cpp
+++ b/lib/Dialect/TTNN/Analysis/L1SpillManagement.cpp
@@ -364,12 +364,10 @@ void L1SpillManagement<MemoryTracker>::applyOutputConfig(
     }
 
     size_t ri = opResult.getResultNumber();
-    TTNNLayoutAttr resultLayout = (ri < result.actualOutputLayouts.size())
-                                      ? result.actualOutputLayouts[ri]
-                                      : chosenLayout;
-    if (!resultLayout) {
-      continue;
-    }
+    assert(ri < result.actualOutputLayouts.size() &&
+           "validation result has fewer output layouts than tensor results");
+    TTNNLayoutAttr resultLayout = result.actualOutputLayouts[ri];
+    assert(resultLayout && "result layout is null for tensor result");
 
     llvm::ArrayRef<int64_t> tensorShape = tensorType.getShape();
 
@@ -396,15 +394,6 @@ void L1SpillManagement<MemoryTracker>::applyOutputConfig(
     ttcore::DataTypeAttr newDataTypeAttr =
         ttcore::DataTypeAttr::get(op->getContext(), chosenLayout.getDataType());
     dtypeOp.setDtypeAttr(newDataTypeAttr);
-  }
-
-  // Update L1 usage attribute.
-  if (chosenLayout.hasL1BufferType() && result.outputL1Usage > 0) {
-    OpBuilder builder(op->getContext());
-    op->setAttr("ttnn.output_l1_usage",
-                builder.getI64IntegerAttr(result.outputL1Usage));
-  } else {
-    op->removeAttr("ttnn.output_l1_usage");
   }
 }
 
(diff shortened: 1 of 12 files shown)


mtopalovicTT (approved):

lib/Dialect/TTNN/Analysis/L1SpillManagement.cpp
```diff
@@ -609,31 +598,32 @@ uint64_t L1SpillManagement<MemoryTracker>::ensureFitsL1(
...
+  });
 
   if (result.isSuccess()) {
-    uint64_t l1Size =
-        result.outputL1Usage > 0 ? result.outputL1Usage : opL1Usage;
-    l1Size = ensureFitsL1(op, pos, data, opL1Usage, result.cbPeakUsage, l1Size);
+    uint64_t l1Size = result.outputL1Usage;
+    if (l1Size > 0) {
+      l1Size = ensureFitsL1(op, pos, data, result.cbPeakUsage, l1Size);
+    }
```
bmalesevicTT, line 631: Won't this still miss checking evictForDramCBGrowth? That check should be added as else branch here

lib/Dialect/TTNN/Analysis/L1SpillManagement.cpp
```diff
@@ -1193,7 +1186,7 @@ void L1SpillManagement<MemoryTracker>::run() {
       continue;
     }
```
bmalesevicTT: I am still confused by this path. Could you please explain how spilling outputs helps corrupted input, and how we even end up in a situation where input is in L1 after this pass while it wasn't before?

lib/Dialect/TTNN/Analysis/L1SpillManagement.cpp
```diff
@@ -645,13 +635,13 @@ void L1SpillManagement<MemoryTracker>::handleOOM(
...
+    // itself is too large for the configured cap. Demote its output to DRAM.
+    // No evictForDramCBGrowth needed: evictUntil already drained the live set.
     observer_->onSelfSpill(op, pos);
     TTMLIR_DEBUG(ttmlir::LogComponent::GreedyOptimizer,
-                 "    SPILL SELF: op exceeds budget alone");
-    for (auto r : tensorResults) {
-      spillToDram(r);
-    }
+                 "    DEMOTE SELF: op exceeds budget alone");
+    demoteToDram(op);
```
bmalesevicTT, line 649: This makes much more sense then spilling and I would keep it as fallback, but I am still a bit curious when we would hit this case. If we evict everything, than the op should be guaranteed to fit because it was validated in isolation in MemoryLayoutPropagation? The only case I see this happening is if the op has its own forced L1 inputs, so their reshards are still live in memory and can't be evicted, is that it?

bmalesevicTT (commented):
Nice, thanks for addressing concerns! Few questions still unclear from the last review

lib/Dialect/TTNN/Analysis/L1SpillManagement.cpp
rpavlovicTT: Not sure if I ever hit this case, but I had to protect against it.

lib/Dialect/TTNN/Analysis/L1SpillManagement.cpp
rpavlovicTT: Good point, we should demote here as well and leave it like that.

lib/Dialect/TTNN/Analysis/L1SpillManagement.cpp
rpavlovicTT: We can do it, I did not think of it, but yeah we could miss it in some super degenerate case. I can't really think of such case, but it won't hurt if we guard against it.

bmalesevicTT (approved):

PR #8527, opened 2026-05-21
[chisel] Add support for retrieving multi-device tensors

### LINKED ISSUE #7840: Multi-Device Tensor Access via retrieve_tensor_from_pool
## Description

Update the existing `retrieve_tensor_from_pool` runtime API to return a map of logical device ID -> per-device host tensor shard, instead of a single merged host tensor. Currently, the C++ implementation (`runtime/lib/ttnn/runtime.cpp:2032`) already calls `toHost()` which returns
`std::vector<Tensor>`, but then fatals if `hostTensors.size() != 1`. The fix is to return the full set of shards keyed by logical device ID, following the same pattern already used by `getOpOutputTensor` (`runtime/lib/ttnn/runtime.cpp:958`) which returns `std::unordered_map<std::uint32_t, Tensor>`. This enables Chisel debug hooks to inspect or override individual device shards (e.g., tracing per-device shapes, replacing output on specific devices).

## Current Behavior

- `retrieveTensorFromPool` returns `std::optional<Tensor>` (single tensor)
- Calls `toHost()` → gets `std::vector<Tensor>`
- `LOG_FATAL` if vector has more than one element (multi-device not supported)
- Python binding returns `Optional[tt.runtime.Tensor]`

## Proposed Change

Change the return type to `std::unordered_map<std::uint32_t, Tensor>` where keys are logical device IDs (indices into the `toHost()` vector, matching the row-major ordering convention documented in `getOpOutputTensor`).

mmilosevicTT:
### Ticket
Closes https://github.com/tenstorrent/tt-mlir/issues/7840

### Problem description
Currently `retrieve_tensor_from_pool` is only working for singlechip. This blocks Chisel from running on multichip graphs.

### What's changed
Since `to_host` is exposed in python, we can only return Tensor object for `retrieve_tensor_from_pool` and then later on decide what to do with this. This unifies the implementation for single-chip and multi-chip and does not create special cases that returning vector would introduce.

### Checklist


tools/chisel/chisel/ops.py
```diff
@@ -118,6 +118,21 @@ def get_asm_state(self) -> AsmState:
...
+        Returns `(1, 1)` when the attribute is absent (single-chip programs).
+        When multiple meshes are declared, the first one is used.
+        """
+        for named_attr in self.module.operation.attributes:
+            if named_attr.name != "ttcore.meshes":
+                continue
+            meshes = ttcore.ir.MeshesAttr.maybe_downcast(named_attr.attr)
+            if meshes is None or not meshes.meshes:
+                continue
+            return tuple(int(d) for d in meshes.meshes[0].shape)
```
ndrakulicTT, line 139: Hmm in what cases are there multiple meshes? Could we raise error or emit record that there is multiple shapes?
Maybe we should check attribute of `ttnn.get_device` instead as it is scoped on function, but not sure what is best solution here

tools/chisel/chisel/utils.py
```diff
@@ -29,16 +30,40 @@ def get_torch_tensor(tensor: Tensor) -> torch.Tensor:
...
 ) -> GoldenMapTensor:
-    device_tensor = tt_runtime.retrieve_tensor_from_pool(
-        rt_program_context, rt_tensor_ref
-    )
-    if device_tensor is None:
+    """Pull the pool entry for `rt_tensor_ref` and wrap it as a GoldenMapTensor.
+
+    The runtime returns one host tensor per device shard (single entry for
+    single-device tensors). Shards are keyed 0..N-1 to match the convention
+    used elsewhere (see tools/builder/base/builder_runtime.py:create_tensor),
```
ndrakulicTT: This comment is bit confusing, not sure what should be seen in `builder_runtime.py:create_tensor`

tools/chisel/chisel/ops.py
mmilosevicTT: Added a check if more meshes are present. For now we always just pick first one in all places.

jgrimTT (approved):
Builder LGTM, thanks!

nsumrakTT (approved):

pilkicTT (approved):

ndrakulicTT (approved):
LGTM

Issue #8479, opened 2026-05-19
PagedUpdateCacheOp produces wrong outputs since #8098


mvasiljevicTT:
## Bug

Since bb1deb417c (#8098), `PagedUpdateCacheOp` writes some users' fill values to the wrong cores on Wormhole. The affected users get garbage decode output.

**The regression is invisible to the existing perf benchmark** — its PCC check only validates batch index 0. The bug only shows up when PCC is computed for **every user** in the batch. On Llama 3.2 1B Instruct (b=32, n150), the per-user check gives:

```
decode PCC bad (<0.94): users 16, 20, 24, 28 -> ~0.0     <- caused by #8098
                      + users 3, 7, 11, 15, 19, 23, 27, 31 (pre-existing, separate)
```

Bisected to bb1deb417c. Root cause: the new `PagedUpdateCacheRuleBook::getInputLayoutFilter` only requires "L1 HeightSharded" but does **not** pin the grid. `OpModel` beam search can then pick a HeightSharded layout whose physical grid does not match the `{numUsers, 1} -> physical` affine map the runtime kernel expects, and per-user data lands on the wrong cores.

## Stopgap

Branches that always run `PagedUpdateCacheOpRewritePattern` (the old workaround) as a backstop:
- tt-mlir: [`mvasiljevic/fix-paged-update-cache-grid-regression`](https://github.com/tenstorrent/tt-mlir/tree/mvasiljevic/fix-paged-update-cache-grid-regression) (`72cc60465`)
- tt-xla: [`mvasiljevic/fix-paged-update-cache-grid-regression`](https://github.com/tenstorrent/tt-xla/tree/mvasiljevic/fix-paged-update-cache-grid-regression) (`c4b3d7f42`)

Verified end-to-end on `manual-benchmark.yml`:
- First run (all tests, `sh-runner=true` — n150 jobs landed on `tt-ubuntu-2204-n150-stable`): https://github.com/tenstorrent/tt-xla/actions/runs/26096646958
- Re-run (LLM-only, `sh-runner=false` — n150 jobs land on dedicated `n150-perf` for apples-to-apples vs baseline): https://github.com/tenstorrent/tt-xla/actions/runs/26099205589

Llama 3.2 1B n150 (from first run):
```
Prefill PCC verification passed with PCC=0.997850  <- correctness restored
First decode PCC verification passed with PCC=0.997706
Previous: sps=68.892715
Current:  sps=55.328705
[...]

PR #8542, opened 2026-05-21
[Optimizer] Fix PagedUpdateCache sharding rules

bmalesevicTT:
### Ticket
#8479

### Problem description
 PagedUpdateCacheRuleBook::getInputLayoutFilter previously accepted any L1 height-sharded layout for the fill-value (operand 1). The tt-metal kernel silently produces PCC-degraded outputs unless the input is height-sharded on the exact {numUsers, 1} virtual grid (one user per row, single column), so the optimizer was free to pick a layout that compiles cleanly but corrupts results.
This operand's grid and layout are already forced via workaround at optimization_level < 2. We want to avoid that workaround as it causes other problems on optimization_level = 2.
### What's changed
- OpRuleBook: added an input-tensor-type-aware getInputLayoutFilter(RankedTensorType inputTensorType, unsigned operandIdx) virtual that defaults to delegating to the existing (unsigned operandIdx) overload — every rule book that doesn't need the tensor type is untouched.
- PagedUpdateCacheRuleBook: overrides the new overload; captures numUsers = input.shape[1] (batch size) and tightens the filter to require gridShape == {numUsers, 1} on top of L1 height-sharded.
- MemoryLayoutPropagation: caller now forwards the operand tensor type to the rule book.

tt-metal fix addressed at: tenstorrent/tt-metal#44923.

### Checklist
- [x] Check most printed outputs for Users1+ are correct, without perf degradation or new errors:
  - Results on base tt-mlir commit: https://github.com/tenstorrent/tt-xla/actions/runs/26226350407 
  - Results on this branch: https://github.com/tenstorrent/tt-xla/actions/runs/26219348710
  - Both actually seemed to have the ticket solved
- [x] Made a branch with reverted commit [[Optimizer] Fix L1 sharding non-determinism: stable sort + ordered maps in beam and reshard candidate generation ](https://github.com/tenstorrent/tt-mlir/commit/fe1f20899a4f9f6447254b6ec0fb95489a61b13f) (which accidentally picks the right grid for now) to prove this fix fixes user outputs 16, 20, 24, 28 if the heuristics were to change again:
   - without fix sharding rules: https://github.com/tenstorrent/tt-xla/actions/runs/26280793115
   - with fix sharding rules: https://github.com/tenstorrent/tt-xla/actions/runs/26292149358/job/77397404892

### DIFF
--- a/.github/test_scripts/op_model.sh
+++ b/.github/test_scripts/op_model.sh
@@ -13,6 +13,8 @@ echo "Running op-model test Interface"
 $BUILD_DIR/test/unittests/OpModel/TTNN/Op/TestOpModelInterface
 echo "Running op-model test MockDevice"
 $BUILD_DIR/test/unittests/OpModel/TTNN/Lib/TestOpModelLibMockDevice
+echo "Running op-model tripwire tests"
+$BUILD_DIR/test/unittests/OpModel/TTNN/Lib/TestOpModelLibTripwires
 
 echo
 echo "Run Optimizer Models Perf Tests"
--- a/lib/Dialect/TTNN/Analysis/OpRules/TransformerRules.cpp
+++ b/lib/Dialect/TTNN/Analysis/OpRules/TransformerRules.cpp
@@ -105,8 +105,8 @@ PagedUpdateCacheRuleBook::getInputLayoutFilter(Operation *op,
                                                unsigned operandIdx) const {
   // Operand 1 (fill value) must be L1 height-sharded on a {numUsers, 1}
   // virtual grid.  Any other grid causes PCC degradation;
-  // TODO(): relax when tt-metal enforces this and raises an error for
-  // PCC-degrading layouts.
+  // https://github.com/tenstorrent/tt-metal/issues/44923 relax when tt-metal
+  // enforces this and raises an error for PCC-degrading layouts.
   if (operandIdx == 1) {
     int64_t numUsers =
         mlir::cast<RankedTensorType>(op->getOperand(1).getType()).getShape()[1];
--- a/lib/Dialect/TTNN/Analysis/OpRules/TransformerRules.cpp
+++ b/lib/Dialect/TTNN/Analysis/OpRules/TransformerRules.cpp
@@ -103,10 +103,18 @@ OutputHints SplitQKVRuleBook::getOutputHints(
 LayoutFilterFn
 PagedUpdateCacheRuleBook::getInputLayoutFilter(RankedTensorType inputType,
                                                unsigned operandIdx) const {
-  // Operand 1 (fill value) must be L1 height-sharded on a {numUsers, 1}
-  // virtual grid.  Any other grid causes PCC degradation;
-  // https://github.com/tenstorrent/tt-metal/issues/44923 relax when tt-metal
-  // enforces this and raises an error for PCC-degrading layouts.
+  // Operand 1 (fill value): L1 height-sharded with virtual grid
+  // {numUsers, 1}, where numUsers = input1.shape[1].  Any other grid
+  // silently produced PCC=0 for upper users
+  // (https://github.com/tenstorrent/tt-metal/issues/44923).  HS in tt-mlir
+  // IR is pinned to {M, 1}, so this is the canonical (and only) HS grid
+  // shape with num_cores == numUsers.
+  //
+  // TODO(bmalesevic): once tt-metal PR #45016 is uplifted, metal enforces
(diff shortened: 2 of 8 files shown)


lib/Dialect/TTNN/Analysis/OpRules/TransformerRules.cpp
```diff
@@ -99,15 +101,22 @@ OutputHints SplitQKVRuleBook::getOutputHints(
 //===----------------------------------------------------------------------===//
 
 LayoutFilterFn
-PagedUpdateCacheRuleBook::getInputLayoutFilter(unsigned operandIdx) const {
-  // Operand 1 (fill value) must be L1 height-sharded.
-  // Reject interleaved and all other sharding types so the beam search
-  // is forced to explore HeightSharded reshard candidates.
+PagedUpdateCacheRuleBook::getInputLayoutFilter(Operation *op,
+                                               unsigned operandIdx) const {
```
rpavlovicTT: I think it would be cleaner if we had

```
getInputLayoutFilter(RankedTensorType inputTensorType, size_t operandIdx)
```

Makes sense to generate filters based on input tensor & index. Passing op pointer seems redundant.

lib/Dialect/TTNN/Analysis/OpRules/TransformerRules.cpp
```diff
@@ -99,15 +101,22 @@ OutputHints SplitQKVRuleBook::getOutputHints(
...
+  // https://github.com/tenstorrent/tt-metal/issues/44923 relax when tt-metal
+  // enforces this and raises an error for PCC-degrading layouts.
   if (operandIdx == 1) {
-    return [](TTNNLayoutAttr layout) -> bool {
+    int64_t numUsers =
+        mlir::cast<RankedTensorType>(op->getOperand(1).getType()).getShape()[1];
+    return [numUsers](TTNNLayoutAttr layout) -> bool {
       auto ml = layout.getMemLayout();
+      auto gridShape = layout.getGridShape();
       return layout.hasL1BufferType() && ml &&
```
rpavlovicTT: You can simplify return statement, we know `ml` is not null (you can add an assert). Also assert `gridShape.size() == 2` and just check whether shape is correct.

lib/Dialect/TTNN/Analysis/OpRules/TransformerRules.cpp
bmalesevicTT: True, corrected now

lib/Dialect/TTNN/Analysis/OpRules/TransformerRules.cpp
bmalesevicTT: Okay, I tried to change as little as possible compared to existing, simplified now

test/unittests/OpModel/TTNN/Lib/TestOpModelLib.cpp
```diff
@@ -5602,6 +5602,47 @@ TEST_F(OpModelTest, PagedUpdateCacheOpWithoutPageTable) {
...
+// grid silently produced PCC=0 for the upper users.
+//
+// tt-mlir's PagedUpdateCacheRuleBook still pins to {numUsers, 1} because
+// tt-metal does not reject other grids
+// (https://github.com/tenstorrent/tt-metal/issues/44923).  This test
+// constructs a non-{numUsers, 1} grid and asserts that OpModel currently
+// accepts it.  When tt-metal adds the assert, this EXPECT will start
+// failing, which is the signal to flip it to EXPECT_FALSE and relax the
+// rule.
+TEST_F(OpModelTest, PagedUpdateCacheOpWrongGridTripwire) {
```
rpavlovicTT, line 5617: Make this a standalone test in a new build target. Reason is faster compile of the test, and easier removal later. Don't forget to add it in the CI script `tt-mlir/.github/test_scripts/op_model.sh`

nsumrakTT (approved):

.github/test_scripts/op_model.sh
```diff
@@ -13,6 +13,8 @@ echo "Running op-model test Interface"
 $BUILD_DIR/test/unittests/OpModel/TTNN/Op/TestOpModelInterface
 echo "Running op-model test MockDevice"
 $BUILD_DIR/test/unittests/OpModel/TTNN/Lib/TestOpModelLibMockDevice
+echo "Running op-model tripwire tests"
```
rpavlovicTT, line 16: I like the name "tripwire" :laughing:

rpavlovicTT (approved):

vvukomanTT (approved):

PR #8546, opened 2026-05-21
Fix unresolved materialization when view_layout is bypassed in d2m.spatial

### LINKED ISSUE #8483: ConvertD2MToTTMetal fails with unresolved materialization when view_layout is bypassed inside d2m.spatial
### Summary

`convert-d2m-to-ttmetal` can fail when a `d2m.generic` inside `d2m.spatial` consumes a value produced by `d2m.view_layout`.

In this flow, generic lowering bypasses `d2m.view_layout` and wires the underlying memref directly into `ttmetal.enqueue_program`.  
`d2m.view_layout` is then erased.  
This can leave `d2m.spatial` operand/type expectations out of sync with the already-rewritten nested ops.

As a result, conversion inserts `builtin.unrealized_conversion_cast` to bridge a memref type mismatch (for example, shard-style memref vs view-typed memref), and that cast remains live, causing full conversion to fail.

### Observed Behavior

- `convert-d2m-to-ttmetal` fails with unresolved materialization.
- A live `builtin.unrealized_conversion_cast` remains after conversion.
- Source and target memref types differ by layout/view semantics.

### Expected Behavior

Lowering should keep `d2m.spatial` operands and region argument types consistent with nested rewritten memref types, so no live `builtin.unrealized_conversion_cast` remains and full conversion succeeds.

### Suspected Root Cause

This appears to come from the two-phase lowering flow:

1. Phase 1 rewrites nested `d2m.generic` and `d2m.view_layout` while keeping `d2m.spatial` legal.
2. Phase 2 rewrites `d2m.spatial`.

Because view bypassing happens in phase 1, nested ops may already use underlying memrefs while `d2m.spatial` still carries view-based type expectations, creating an unresolved type bridge.

### Conceptual Repro Pattern

1. A value passes through `d2m.view_layout`.
2. A `d2m.generic` inside `d2m.spatial` consumes that value.
3. Generic lowering replaces the operand with the pre-view underlying memref.
4. `d2m.view_layout` is removed.
5. `d2m.spatial` still expects the old view-typed operand contract.
6. Conversion inserts a cast and later fails due to unresolved materialization.

hkwonTT:
### Ticket
closes #8483

### Problem description
convert-d2m-to-ttmetal could fail in d2m.spatial when d2m.generic used a value from d2m.view_layout.

During lowering, d2m.generic could bypass d2m.view_layout and use the underlying memref directly.
This could make d2m.spatial operand types and inner op types inconsistent, leaving a live builtin.unrealized_conversion_cast and causing conversion failure.

### What's changed
 - During view_layout lowering, d2m.spatial inputs are aligned with the canonical underlying memref path.
 - This update happens before removing d2m.view_layout, so d2m.spatial does not keep stale view-based operand. expectations.
 - Added a regression test for the d2m.spatial + d2m.view_layout pattern to ensure this path remains stable.

### Checklist


lib/Conversion/D2MToTTMetal/D2MToTTMetal.cpp
```diff
@@ -529,8 +529,20 @@ class D2MViewLayoutRewriter : public OpConversionPattern<d2m::ViewLayoutOp> {
...
                   ConversionPatternRewriter &rewriter) const final {
+    Value sourceInput = op.getInput();
+
+    // When d2m.spatial consumes the view result, pre-update that use to the
+    // view input so the spatial operand type tracks the underlying memref type.
+    for (OpOperand &use :
+         llvm::make_early_inc_range(op.getResult().getUses())) {
+      if (mlir::isa<d2m::SpatialOp>(use.getOwner())) {
+        use.set(sourceInput);
+        continue;
```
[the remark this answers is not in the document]
hkwonTT: Updateda as suggested.

lib/Conversion/D2MToTTMetal/D2MToTTMetal.cpp
```diff
@@ -534,8 +534,20 @@ class D2MViewLayoutRewriter : public OpConversionPattern<d2m::ViewLayoutOp> {
...
+      }
+    }
+
     // Erase views.
-    rewriter.replaceOp(op, adaptor.getInput());
+    rewriter.replaceOp(op, sourceInput);
```
vtangTT, line 550: correct me if i'm wrong, wouldn't changing this from `adaptor.getInput` to `op.getInput` fix this problem? idt we need the loop above?

hkwonTT: Nope. it does not fix the problem. The code above(L541-L547) fix the problem.

Copilot review pointed out I need to use adaptor.getInput() instead of op.getInput(). I fixed it before, but it seems that the update was dropped while rebasing hotfix PR. Thanks!

vtangTT (approved):

PR #8551, opened 2026-05-21
[TTIRToTTIRDecomposition] Decompose integer prod to avoid bf16 downcast

sonalibaskaran2499:
### Ticket
[Link to Github Issue](https://github.com/tenstorrent/tt-metal/issues/44942)

### Problem description
ttir.prod lowers to ttnn.prod, which runs on tile math units in bf16 and silently downcasts integer operands. Because bf16 has a step of 16 in the range [2048, 4096], an integer reduction like 1 * 38 * 58 = 2204 rounds to 2208, corrupting integer shape arithmetic. This surfaced in Qwen2.5-VL, QWEN3_5/multimodal where image_grid_thw.prod(-1) is used to compute token counts and a wrong integer here propagates into downstream shape logic, breaking the model.

### What's changed
Keep integer prods in the integer domain by decomposing them at the TTIR level instead of lowering to ttnn.prod.

- Added **ReductionProdIntegerDecomposePattern** in `TTIRToTTIRDecomposition.cpp` for an integer ttir.prod over a single dim, it emits a ttir.slice_static per index along that dim and multiplies them with ttir.multiply, optionally reshaping when keep_dim == false. Empty reductions are deferred so the existing lowering materializes the prod identity.

- Multi-dim integer prods are handled by composition: the existing ReductionProdPattern splits them into a chain of single-dim ttir.prods, which this new pattern then rewrites.

- Updated the dynamic legality check in `TTIRToTTIRDecompositionPass.cpp` so single-dim integer ttir.prod is illegal and the new pattern fires.

**Impact:** Integer ttir.prod now produces bit-exact integer results; the float path through ttnn.prod is unchanged. Unblocks Qwen2.5-VL's image_grid_thw.prod(-1) and any other model relying on integer shape reductions.

### Checklist

### DIFF
--- a/lib/Conversion/TTIRToTTIRDecomposition/TTIRToTTIRDecomposition.cpp
+++ b/lib/Conversion/TTIRToTTIRDecomposition/TTIRToTTIRDecomposition.cpp
@@ -1409,6 +1409,90 @@ public:
 };
 } // namespace
 
+// tt-metal's ttnn.prod runs on tile math units in bf16 and silently downcasts
+// integer operands: e.g. 1*38*58 = 2204 rounds to 2208 (bf16 step is 16 in
+// [2048, 4096]), corrupting integer shape arithmetic such as Qwen2.5-VL's
+// `image_grid_thw.prod(-1)`. Stay in the integer domain by emitting per-index
+// slices and multiplying them.
+//
+// Multi-dim integer prods are first split into single-dim ttir.ProdOps by
+// ReductionProdPattern, which this pattern then rewrites.
+namespace {
+struct ReductionProdIntegerDecomposePattern
+    : public OpConversionPattern<ttir::ProdOp> {
+public:
+  using OpConversionPattern<ttir::ProdOp>::OpConversionPattern;
+
+  LogicalResult
+  matchAndRewrite(ttir::ProdOp op, OpAdaptor adaptor,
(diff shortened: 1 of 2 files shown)


dmilinkovicTT (changes requested):
Since this is an issue with a TTNN op, we should perform this decomposition in `TTNNWorkarounds`, and file a tt-metal issue.

sonalibaskaran2499:
Thanks for the feedback @dmilinkovicTT. I will fix this error in TTNNWorkarounds.

sonalibaskaran2499:
Closing this PR as this will be fixed in TTNNWorkarounds.

sdjordjevicTT:
@sonalibaskaran2499 TTNN Workarounds are located in tt-mlir: 
https://github.com/tenstorrent/tt-mlir/tree/main/lib/Dialect/TTNN/Transforms/Workarounds/Decomposition

Additionally, please attach the tt-metal issue in the workaround comment that references it. Once the issue is resolved, we can remove the workaround.

sonalibaskaran2499:
Thanks for the clarification @sdjordjevicTT . I’ll move the decomposition to the TTNN Workarounds in tt-mlir and attach the corresponding tt-metal issue reference in the workaround comment so it can be cleaned up once the issue is resolved.

PR #8555, opened 2026-05-22
Remove DType

### LINKED ISSUE #8378: Remove `dtype` as op attribute in TTNN

djstefanovicTT:
### Ticket
closes https://github.com/tenstorrent/tt-mlir/issues/8378

### Problem description
`dtype` attribute was redundant as informations it contains can be found directly in layout. This created a situation where updating one place needed to be aware of the other and introduced unnecessary complications.

### What's changed
Removed `dtype` attribute from TTNN IR.
Updated all relevant call sites.
Created helper getDtypeAttr that create this attribute on demand from layout encoding.
Created helper hasDtypeChange which checks if the op changes data type.
Tensor layout is now the single source of truth regarding data type.

### Checklist


include/ttmlir/Dialect/TTNN/Interfaces/TTNNTensorSpecInterface.td
```diff
@@ -7,35 +7,25 @@
 
 include "mlir/IR/OpBase.td"
 
-// This interface is used to specify TTNN ops that have defined DtypeAttr
+// This interface is used to specify TTNN ops whose output data type can be
+// queried. The dtype is derived from the first result's TTNNLayoutAttr encoding.
```
dmilinkovicTT, line 11: Instead of deriving the dtype from the first result, I think it's better to assert that the op has exactly one result (possibly as a compile-time assertion), and derive dtype from the only result.

sdjordjevicTT (approved):

lib/Dialect/TTNN/Interfaces/TTNNOpModelInterface.cpp
```diff
@@ -1674,9 +1702,16 @@ BitcastConvertOp::getOpConstraints(const std::vector<TTNNLayoutAttr> &inputs,
   ASSIGN_OR_RETURN(ttcore::GridAttr deviceGrid,
                    detail::getValidatedDeviceGrid(getOperation()));
 
+  ttcore::DataTypeAttr dtype =
+      detail::resolveOutputDtype(getOperation(), opConfig.outputLayout);
+  if (!dtype) {
+    return issueErrorForGetOpConstraints(
+        getOperation(), detail::ReasonForLackOfSupport::MissingMetalDefinition);
+  }
```
sgholamiTT: Hmm I don't think this would be the right semantic for this purpose. Maybe `assert(dtype);` ? 
wdyt @rpavlovicTT ?

jgrimTT (approved):
Builder changes look good

arminaleTT (approved):

vtangTT (approved):

dmilinkovicTT (approved):

lib/Dialect/TTNN/Interfaces/TTNNOpModelInterface.cpp
rpavlovicTT: I agree with Saber. This is incorrect behaviour. Error is thrown only if we don't support querying such op. In this case you should assert.

lib/Dialect/TTNN/Interfaces/TTNNOpModelInterface.cpp
```diff
@@ -1748,9 +1804,20 @@ ToLayoutOp::getOpConstraints(const std::vector<TTNNLayoutAttr> &inputs,
...
+  // Only signal a dtype conversion to the op model when this to_layout is
+  // genuinely changing dtype. Use the resolved (candidate or intrinsic)
+  // output dtype, and compare it against the input dtype.
+  std::optional<ttcore::DataType> outputDtype = std::nullopt;
+  if (ttcore::DataTypeAttr dtype =
+          detail::resolveOutputDtype(getOperation(), opConfig.outputLayout)) {
+    if (dtype.getValue() != inputs[0].getDataType()) {
+      outputDtype = dtype.getValue();
+    }
+  }
```
rpavlovicTT, line 1804: Does this change behaviour compared to old way of passing dType for this op?

djstefanovicTT: It has some differences. Since type attribute does not exist any more, it validates now with dtype attribute that the optimizer is trying to probe, and it passes null opt if it is the same as input type. Helper `resolveOutputDtype` is added because of this.

rpavlovicTT: Got it. Looks good.

include/ttmlir/Conversion/Passes.td
```diff
@@ -396,8 +396,8 @@ def ConvertD2MToTTNN: Pass<"convert-d2m-to-ttnn", "::mlir::ModuleOp">{
       %output_l1 = ttir.ttnn_metal_layout_cast
 
     Output IR (post canonicalization):
-      %1 = "ttnn.to_memory_config"(%arg0) <{memory_config = #ttnn.memory_config<#l1, <block_sharded>, #ttnn.shard_spec<<[#ttnn.core_range<(0,0), (0,0)>]>, <32x32>, <row_major>, <physical>>>}> : (tensor<32x32xf32, #ttnn_layout>) -> tensor<32x32xf32, #ttnn_layout1>
-      %2 = "ttnn.empty"(%0) <{dtype = #ttcore.supportedDataTypes<f32>, layout = #ttnn.layout<tile>, memory_config = #ttnn.memory_config<#l1, <block_sharded>, #ttnn.shard_spec<<[#ttnn.core_range<(0,0), (0,0)>]>, <32x32>, <row_major>, <physical>>>, shape = #ttnn.shape<32x32>}> : (!ttnn.device) -> tensor<32x32xf32, #ttnn_layout1>
+      %1 = "ttnn.to_memory_config"(%arg0) <{layout = #ttnn.layout<tile>, #ttnn.shard_spec<<[#ttnn.core_range<(0,0), (0,0)>]>, <32x32>, <row_major>, <physical>>>}> : (tensor<32x32xf32, #ttnn_layout>) -> tensor<32x32xf32, #ttnn_layout1>
```
azecevic000: `to_memory_config` doesn't have attributes anymore, so it should be `%1 = "ttnn.to_memory_config"(%arg0) : (...) -> ...`

include/ttmlir/Conversion/Passes.td
```diff
@@ -396,8 +396,8 @@ def ConvertD2MToTTNN: Pass<"convert-d2m-to-ttnn", "::mlir::ModuleOp">{
...
+      %2 = "ttnn.empty"(%0) <{layout = #ttnn.layout<tile>, #ttnn.shard_spec<<[#ttnn.core_range<(0,0), (0,0)>]>, <32x32>, <row_major>, <physical>>>, shape = #ttnn.shape<32x32>}> : (!ttnn.device) -> tensor<32x32xf32, #ttnn_layout1>
```
azecevic000: for `empty` only `layout` persists, `shard_spec` is part of `memory_config` so it isn't in IR anymore

include/ttmlir/Dialect/TTNN/Interfaces/TTNNTensorSpecInterface.h
```diff
@@ -37,39 +37,56 @@ inline MemoryConfigAttr getMemoryConfigFromResult(mlir::Operation *op) {
...
-             << "output tensor layout data type "
-             << DataTypeEnumToString(outputLayoutAttr.getDataType())
-             << " must match output data type attribute "
-             << DataTypeEnumToString(outputDTypeAttr.getValue());
-    }
+  // Fall back to deriving from the tensor's element type.
+  std::optional<ttcore::DataType> dataType =
+      ttcore::elementTypeToDataTypeImpl(tensor.getElementType());
+  if (!dataType) {
+    return nullptr;
```
azecevic000: Not sure if it's a good idea to silently fail here, convention should be that if a value has a tensor type, we should always be able to derive the dtype.

include/ttmlir/Dialect/TTNN/Interfaces/TTNNTensorSpecInterface.td
azecevic000: If we go with this approach we should also provide `getDtypeAttr(idx)` API. IMO we shouldn't do `static_assert` for a sole reason that `TTNN_DtypeOpInterface` should be part of the topmost tablegen class, adding in to each op is error-prone. Runtime `assert` is more pragmatic approach in this case. `getDtypeAttr()` can assert that there is exatly one result and `getDtypeAttr(idx)` asserts that `idx < numResults`.

lib/Target/TTNN/TTNNToFlatbuffer.cpp
```diff
@@ -376,7 +378,10 @@ createOp(FlatbufferObjectCache &cache, FullOp op) {
   } else {
     llvm_unreachable("fill value must be float or integer");
   }
-  auto dtype = toFlatbuffer(cache, op.getDtype());
+  ::flatbuffers::Optional<::tt::target::DataType> dtype;
+  if (auto dtypeAttr = op.getDtypeAttr()) {
```
azecevic000: It makes sense to do `dtype = toFlatbuffer(cache, op.getDtypeAttr())` unconditionally, maybe even add `assert(dtype)` because we expect that dtype is always present. The same applies to other ops in this file.

include/ttmlir/Dialect/TTNN/Interfaces/TTNNTensorSpecInterface.td
```diff
@@ -7,35 +7,38 @@
...
       /*desc=*/[{
+        Returns the output data type as a DataTypeAttr, derived from the
+        targeted result's TTNNLayoutAttr encoding.
         }],
       /*retTy=*/"mlir::tt::ttcore::DataTypeAttr",
-      /*methodName=*/"getDtypeAttr",
-      /*args=*/(ins),
-      /*methodBody=*/[{
-        return $_op.getDtypeAttr();
+      /*methodName=*/"getDtypeFromResult",
```
azecevic000, line 21: IMO `getDtypeAttr` make more sense, it's a similar pattern as `getResult()`/`getResult(idx)` that MLIR already uses.

azecevicTT (approved):

vwellsTT (approved):

PR #8566, opened 2026-05-22
Hoist common reshape before slices

### LINKED ISSUE #8557: OOM caused by reshapes;  reshapes attached to multiple slice ops + tile-dim aligned, which enforces big memory footprint
There is a problematic pattern in video gen decoders that keeps large non tile aligned tensors and slices them. Each slice is then reshaped to tile aligned shape and worked on. But until all slices are processed original tensor lives in DRAM, which eventually causes OOM crash.

Since reshapes of each slice are equivalent, a reshape of the original tensor can be done. This way large non tile aligned tensor does not occupy dram space for long, and overall number of reshape ops is reduced.

djstefanovicTT:
### Ticket
closes https://github.com/tenstorrent/tt-mlir/issues/8557

### Problem description
Large non tile aligned tensors are living too long in DRAM and can cause OOM.

### What's changed
Added a pattern that moves equivalent reshape ops from every slice of a common tensor directly to it.

Before:

x -> slice -> reshape
   |
   -> slice -> reshape
   ...
   |
   -> slice -> reshape
   
 After:
 
 x -> reshape -> slice
                        |
                        -> slice
                        ...
                        |
                        -> slice
### Checklist

### DIFF
--- a/lib/Dialect/TTNN/Transforms/TTNNMemoryManagement.cpp
+++ b/lib/Dialect/TTNN/Transforms/TTNNMemoryManagement.cpp
@@ -23,6 +23,20 @@ static SmallVector<int32_t> toI32Vec(ArrayAttr attrs) {
   return result;
 }
 
+// Rewrites any negative slice index into its non-negative equivalent.
+static void normalizeSliceIndices(SmallVectorImpl<int32_t> &begins,
+                                  SmallVectorImpl<int32_t> &ends,
+                                  ArrayRef<int64_t> inputShape) {
+  for (size_t d = 0; d < inputShape.size(); ++d) {
+    if (begins[d] < 0) {
+      begins[d] += static_cast<int32_t>(inputShape[d]);
+    }
+    if (ends[d] < 0) {
+      ends[d] += static_cast<int32_t>(inputShape[d]);
+    }
+  }
+}
+
 static bool allUsersAreSlice(Operation *op) {
   return llvm::all_of(op->getUsers(), [](Operation *user) {
     return isa<ttnn::SliceStaticOp>(user);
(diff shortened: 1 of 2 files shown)


lib/Dialect/TTNN/Transforms/TTNNMemoryManagement.cpp
```diff
@@ -677,16 +867,17 @@ class TTNNMemoryManagement
...
-        PropagateSliceThroughPermute, PropagateSliceThroughReshape,
-        PropagateSliceThroughRepeat, PropagateSliceThroughEltwise<ttnn::AddOp>,
-        PropagateSliceThroughEltwise<ttnn::MultiplyOp>,
-        PropagateSliceThroughEltwise<ttnn::SubtractOp>,
-        PropagateSliceThroughEltwise<ttnn::DivideOp>, RepeatReshapeAdjusting,
-        ReshapeElementwiseAdjusting<ttnn::AddOp>,
-        ReshapeElementwiseAdjusting<ttnn::MultiplyOp>,
-        ReshapeElementwiseAdjusting<ttnn::SubtractOp>,
-        ReshapeElementwiseAdjusting<ttnn::DivideOp>>(&getContext());
+    patterns
```
sdjordjevicTT, line 892: This pattern seems beneficial to run all the time, right? I would mark this pass to run always in the TTNN pipeline, but selectively select paterns that should be run always. What do you think?

djstefanovicTT: I think that making the pass on by default and guarding problematic patterns that can have performance degradation with existing flag would be ok.

sdjordjevicTT: Yes, that sounds great!

sdjordjevicTT (approved):

Issue #8515, opened 2026-05-20
[L1SpillMgmt] Freshly inserted reshards missing from ScheduleData


rpavlovicTT:
## Summary

When the L1 spill management pass evicts a tensor and calls `insertReshardForConsumer()`, it inserts a new `ToMemoryConfigOp` directly into the IR.  These freshly inserted reshard ops are **never registered in `ScheduleData`**, which means:

1. **Future reshards** (consumer position > current sweep position) are absent from `scheduleData.schedule` and `positionMap`.  Subsequent ops that use the reshard's result cannot look up its schedule position, so the pass cannot emit correct L1 deallocation events for the reshard's output, and Belady eviction cannot account for its liveness interval.

2. **Past reshards** (consumer position < current sweep position — e.g. inserted for an already-processed op during a `handleOOM` cascade) are absent from the address snapshots.  If a later eviction triggers snapshot/replay, the replay loop skips the reshard's allocation event entirely, producing incorrect free-list state and potentially wrong addresses for all subsequent live tensors.

## Reproduction

Minimal scenario:

1. Ops scheduled as `[A, B, C]` where A's result lives across B and C.
2. At op C, OOM causes eviction of A → spill to DRAM.
3. A reshard is inserted for consumer C: IR becomes `[A, B, ReshardA, C]`.
4. `ScheduleData.positionMap` still maps `{A→0, B→1, C→2}`.  `ReshardA` has no entry.
5. If a second eviction occurs later, snapshot replay at position 2 doesn't replay `ReshardA`'s allocation, so the reconstructed tracker state is wrong.

## Root cause

`buildScheduleData()` is called once before the sweep begins and is never updated when new ops are inserted into the IR.  `insertReshardForConsumer()` mutates the IR but has no path back to `ScheduleData` or `addressSnapshots`.

Relevant code:
- `L1SpillManagement::buildScheduleData()` — `lib/Dialect/TTNN/Analysis/L1SpillManagement.cpp`
- `L1SpillManagement::insertReshardForConsumer()` — same file
- `SumL1MemoryTracker::addressSnapshots` and `l1EventLog` — `include/ttmlir/Dialect/TTNN/Analysis/L1SpillManagement.h`

## Fix

After `insertReshardForConsumer()` inserts a reshard op at position `insertPos` relative to `currentPos`:

**Case A — future reshard (`insertPos >= currentPos`):**
- Splice the new op into `scheduleData.schedule` at the correct index.
[...]

Issue #8514, opened 2026-05-20
[L1SpillMgmt] Add unit test infra for L1 spill management pass


rpavlovicTT:
## Summary

The L1 spill management pass (`L1SpillManagement.cpp`) contains several interacting mechanisms — memory tracker, event-log snapshots, Belady eviction, reshard insertion, spill-to-DRAM, and the revalidation cascade — but there is currently no way to unit-test these in isolation.

All existing coverage comes through full model compilation lit tests, which makes it hard to:
- Reproduce a specific failure case with a minimal graph
- Assert on a particular sequence of evictions or reshard insertions
- Verify snapshot/replay correctness after a targeted eviction
- Test the memory tracker's free-list and address aliasing logic directly

## Desired infra

Create a C++ unit-test fixture (gtest or llvm-lit based) that allows a test to:

1. **Build a custom op graph** with an arbitrary DAG of `ttnn` ops (e.g. `MatmulOp`, `AddOp`, `ToLayoutOp`, `ReshardOp`) using the MLIR C++ API without going through the full optimizer pipeline.
2. **Assign explicit layouts** to every op result — both L1-sharded and DRAM — so a test fully controls the memory footprint of each live tensor.
3. **Run the pass directly** on the constructed module, supplying a mock system descriptor with configurable L1 budget per core.
4. **Inspect pass output** and assert on:
   - Which tensors were spilled to DRAM (`ToMemoryConfigOp` insertion)
   - Which reshards were inserted and at what schedule positions
   - Final `ScheduleData` contents (positionMap, deathSchedule)
   - `SumL1MemoryTracker` state (occupied bytes, free-list)
   - Snapshot/replay log entries via the `L1SpillObserver` interface

## Why this matters

The pass interacts with OpModel validation, address replay, and MLIR IR mutation in ways that are very hard to exercise via lit tests alone.  A targeted graph-builder fixture would let us:
- Reproduce the schedule-data staleness bug (see companion issue) in a two-line test
- Catch regressions in Belady eviction ordering
- Validate CB-fragmentation cushion behaviour with exact L1 budgets

## Suggested approach

- Add a `test/unittests/Dialect/TTNN/L1SpillManagementTest.cpp` file.
- Provide a `L1SpillTestBuilder` helper that wraps `mlir::OpBuilder` and exposes `addMatmul()`, `addEltwiseBinary()`, `setLayout()`, `setL1Budget()`, etc.
[...]

PR #8567, opened 2026-05-22
[Optimizer] Fix address simulation to correctly track inserted reshards

bmalesevicTT:
### Ticket
[XLA issue Falcon3-7B L1 OOM fragmentation
](https://github.com/tenstorrent/tt-xla/issues/4394)
https://github.com/tenstorrent/tt-mlir/issues/8515

## Summary

- Fix `L1SpillManagement` address simulation to correctly track reshards
  inserted for already-processed consumers in the event log, preventing
  stuck address slots and stale CB-overlap checks
- Add `insertReshardIntoSchedule` so reshards inserted for not-yet-processed
  consumers are handled naturally by the forward sweep
- Rewind the sweep when a reshard is inserted for the op currently being
  processed, so the reshard and the op are processed in the correct order
- Guard inserted reshards from being selected as eviction victims, except in
  `handleFragmentation`'s sibling-operand eviction path (to align all siblings
  to DRAM)

## Changes

**`insertEventIntoLog(pos, event)`**
Inserts an `L1Event` at an arbitrary position in `l1EventLog` and shifts all
`allocEventIndex` entries and `addressSnapshots` keys `≥ pos` by 1, preserving
both index invariants.

**Past-consumer reshards**
When `evictValue` inserts a reshard for an already-processed consumer, it now
also inserts a `kAlloc` event just before the consumer's first alloc in the log
and a `kDealloc` just after its last. Since the victim is always scheduled
before its consumers, the immediately-following `markEvictedAndRebuild` replay
always passes through both events and refreshes the consumer's alloc snapshots.
Every subsequent replay then:
- Allocates the reshard's slot before the consumer (correct addresses).
- Frees it after — if the consumer is itself later evicted, its kAlloc is
  skipped but the reshard kDealloc still fires and reclaims the slot.

Past-consumer reshards are not added to `allocEventIndex` so they can never be selected as
eviction victims.

**Future-consumer reshards**
`insertReshardIntoSchedule` splices the reshard op into `data.schedule` at the
consumer's position and updates `positionMap` and `deathSchedule` so the forward
sweep processes it like any other L1 op, logging its kAlloc/kDealloc naturally.
Its per-core L1 size is taken from validating the inserted reshard. We don't
shift `lastUsePositions` for existing ops, because that map only drives 
farthest-last-use eviction heuristic and insertions preserve relative ordering;
the reshard's own `lastUsePositions` value is irrelevant since it is
purposefully never chosen for eviction.

**Reshard for the op currently being processed (schedule rewind)**
A reshard can also be inserted for the op being processed *right now* (its
operand was just spilled and the op hard-requires the L1 layout).
`insertReshardIntoSchedule` then splices the reshard at the current sweep position, shifting the op forward. `run()` detects this (`data.schedule[pos] !=
op`), rewinds `pos` by one and continues, so the sweep processes the inserted
reshard(s) and then reprocesses the op (now reading the reshard's L1 output). A
single `--pos` suffices regardless of how many reshards were inserted — they
land contiguously from the original position and the loop walks forward through
all of them.

**Reshard eviction guard**
Future-consumer reshards live in `liveValues`, so `evictFarthestUse` skips them
(they exist to feed their consumer). The exception is `handleFragmentation`'s
sibling-operand spill: when an op (e.g. concat) needs all operands in DRAM to be
valid, every L1 operand — inserted reshards included — is spilled. `evictValue`
is passed `skipReshardConsumer=op` so it spills *without* re-resharding that op
back to L1 (which validation would otherwise trigger mid-loop, seeing still-mixed
inputs and defeating the homogeneous spill). The op's other past/future
consumers are still restored.

**Known limitation**
Past consumers with DRAM-only outputs have no `allocEventIndex` anchor, so the
reshard's slot cannot be bracketed in the log. The reshard is still inserted in
the IR; the address simulation is slightly optimistic around that consumer's
execution window not to have CB-overlap due to input address change. Documented
in source comments.

### Checklist
- [x] Falcon3-7B passes without OOM L1 fragmentation on CI
- [x] No significant regressions on other perf benchmark tests (the other "improvements" are not really improvements but random fails)
  - on base commit: https://github.com/tenstorrent/tt-xla/actions/runs/26399822651
  - this change on top: https://github.com/tenstorrent/tt-xla/actions/runs/26401785210
image

- [x] Same result on newer version of base + changes on branch:
  - on base commit: https://github.com/tenstorrent/tt-xla/actions/runs/26826927936
  - this change on top: https://github.com/tenstorrent/tt-xla/actions/runs/26820846632
 
### Future related work
- Add tests through new test infra when added by https://github.com/tenstorrent/tt-mlir/issues/8514

### DIFF
--- a/include/ttmlir/Dialect/TTNN/Analysis/L1SpillManagement.h
+++ b/include/ttmlir/Dialect/TTNN/Analysis/L1SpillManagement.h
@@ -328,8 +328,7 @@ private:
 
   /// OOM recovery: evict farthest-last-use tensors or demote self to DRAM.
   void handleOOM(Operation *op, int64_t pos,
-                 llvm::ArrayRef<OpResult> tensorResults,
-                 ScheduleData &data,
+                 llvm::ArrayRef<OpResult> tensorResults, ScheduleData &data,
                  std::function<void(uint64_t)> addResultsToLiveSet);
 
   /// Evict all live L1 tensors. Used when encountering ops without OpModel
@@ -353,10 +352,14 @@ private:
   bool evictUntil(int64_t pos, ScheduleData &data,
                   std::function<bool()> shouldStop);
 
-  /// Evict a specific live value: spill to DRAM, update tracker, insert
-  /// reshards for already-processed consumers. Used by evictUntil and by
-  /// sibling-operand eviction paths.
-  void evictValue(Value victim, int64_t pos, ScheduleData &data);
+  /// Evict a specific live value: spill to DRAM, update tracker, and insert
+  /// reshards for consumers that still require the L1 layout. Sibling-operand
+  /// eviction passes skipReshardConsumer=op: it spills all operands of that op
+  /// to give it homogeneous DRAM inputs, so resharding one back to L1 would
+  /// defeat the purpose. The op's other (past/future) consumers are still
+  /// restored.
+  void evictValue(Value victim, int64_t pos, ScheduleData &data,
+                  Operation *skipReshardConsumer = nullptr);
 
   /// Insert a ToMemoryConfigOp before an already-processed consumer to
   /// convert the DRAM spill output back to the consumer's expected L1 layout.
--- a/include/ttmlir/Dialect/TTNN/Analysis/L1SpillManagement.h
+++ b/include/ttmlir/Dialect/TTNN/Analysis/L1SpillManagement.h
@@ -353,10 +353,9 @@ private:
                   std::function<bool()> shouldStop);
 
   /// Evict a specific live value: spill to DRAM, update tracker, and insert
-  /// reshards for consumers that still require the L1 layout. Sibling-operand
-  /// eviction passes skipReshardConsumer=op: it spills all operands of that op
-  /// to give it homogeneous DRAM inputs, so resharding one back to L1 would
-  /// defeat the purpose. The op's other (past/future) consumers are still
+  /// reshards for consumers that still require the L1 layout. If
+  /// skipReshardConsumer is non-null, that one consumer is left reading the
+  /// DRAM spill (no reshard inserted for it); all other consumers are still
   /// restored.
   void evictValue(Value victim, int64_t pos, ScheduleData &data,
                   Operation *skipReshardConsumer = nullptr);
--- a/include/ttmlir/Dialect/TTNN/Analysis/L1SpillManagement.h
+++ b/include/ttmlir/Dialect/TTNN/Analysis/L1SpillManagement.h
@@ -227,6 +227,13 @@ private:
   /// selected as eviction victims.
   llvm::DenseMap<Value, size_t> allocEventIndex;
 
+  /// Results of reshards inserted by insertReshardIntoSchedule (future
+  /// consumers). These must never be selected as eviction victims — evicting
+  /// them defeats the purpose of the reshard. A DenseSet rather than checking
+  /// isa<ToMemoryConfigOp> avoids falsely blocking regular to_memory_config
+  /// ops from MemoryLayoutPropagation, which ARE valid eviction candidates.
+  llvm::DenseSet<Value> insertedReshardValues;
+
   /// Mark victim's alloc/dealloc events as skipped, restore the snapshot
   /// taken before victim's alloc, then replay all subsequent non-skipped
   /// events to rebuild a consistent address-simulation state. Reshard
--- a/include/ttmlir/Dialect/TTNN/Analysis/L1SpillManagement.h
+++ b/include/ttmlir/Dialect/TTNN/Analysis/L1SpillManagement.h
@@ -201,6 +201,9 @@ private:
   /// Event log entry for address reconstruction. Records every L1 allocation
   /// and deallocation in schedule order so that eviction can replay the full
   /// history (including dead tensors) to compute accurate addresses.
+  /// Reshards inserted for past consumers also get kAlloc/kDealloc entries
+  /// (via insertEventIntoLog) so future replays account for their transient
+  /// slot without needing a separate injection map.
   struct L1Event {
     enum Kind { kAlloc, kDealloc };
     Kind kind;
@@ -213,17 +216,26 @@ private:
   llvm::SmallVector<L1Event> l1EventLog;
 
   /// Snapshots of tracker state taken before each alloc event, keyed by
-  /// event-log index. Used as starting points for replay after eviction.
+  /// event-log index. Invariant: addressSnapshots[i] = tracker state
+  /// immediately before event i fires. Maintained by insertEventIntoLog.
+  /// Used as starting points for replay after eviction.
   llvm::DenseMap<size_t, typename MemoryTracker::Snapshot> addressSnapshots;
 
-  /// Maps each tensor Value to its alloc event index in l1EventLog.
-  /// Provides O(1) lookup in markEvictedAndRebuild instead of linear scan.
+  /// Maps each live-tensor Value to its alloc event index in l1EventLog.
+  /// O(1) restore-point lookup in markEvictedAndRebuild. NOT populated for
+  /// reshard events for past consumers — irrelevant as reshards are never
+  /// selected as eviction victims.
   llvm::DenseMap<Value, size_t> allocEventIndex;
 
-  /// Mark all events for a tensor as skipped and rebuild from its alloc
-  /// snapshot. Updates snapshots during replay so future evictions start
-  /// from accurate state.
+  /// Mark victim's alloc/dealloc events as skipped, restore the snapshot
+  /// taken before victim's alloc, then replay all subsequent non-skipped
+  /// events to rebuild a consistent address-simulation state. Reshard
+  /// kAlloc/kDealloc pairs in the log are replayed naturally.
   void markEvictedAndRebuild(Value victim);
+
+  /// Insert event at pos in l1EventLog and shift all allocEventIndex entries
+  /// and addressSnapshots keys >= pos by 1, preserving the index invariants.
+  void insertEventIntoLog(size_t pos, L1Event event);
 
   /// Extract OpConfig from op's current IR state (result type + op-specific
   /// attrs like Conv2dConfig, MatmulProgramConfig).
@@ -271,6 +283,13 @@ private:
   /// Build schedule, last-use positions, death schedule, and position map.
   ScheduleData buildScheduleData();
 
+  /// Insert a reshard op into the schedule at consumerPos (shifting the
+  /// consumer and all later ops by +1), updating positionMap, lastUsePositions,
+  /// and deathSchedule so the forward sweep processes the reshard naturally.
+  void insertReshardIntoSchedule(Operation *reshardOp, Value reshardResult,
+                                 uint64_t reshardSizePerCore,
+                                 int64_t consumerPos, ScheduleData &data);
+
   /// Remove result tensors whose last use was the previous position.
   void processDeadTensors(int64_t pos, const ScheduleData &data);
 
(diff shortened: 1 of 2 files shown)


lib/Dialect/TTNN/Analysis/L1SpillManagement.cpp
```diff
@@ -784,16 +844,133 @@ void L1SpillManagement<MemoryTracker>::evictValue(Value victim, int64_t pos,
...
+                                             {L1Event::kAlloc, reshardResult,
+                                              reshardSizePerCore, false}});
+            }
+          } else {
+            // Insert the reshard into the schedule so the forward sweep
+            // processes it naturally, adding it to liveValues and the tracker.
+            insertReshardIntoSchedule(reshardResult.getDefiningOp(),
+                                      reshardResult, reshardSizePerCore,
+                                      posIt->second, data);
+          }
```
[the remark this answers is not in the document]
rpavlovicTT: This is a good point and I think we should re-read the schedule if the current pos was affected.

lib/Dialect/TTNN/Analysis/L1SpillManagement.cpp
```diff
@@ -338,6 +338,14 @@ Value L1SpillManagement<MemoryTracker>::evictFarthestUse() {
       continue;
     }
 
+    // Skip reshards we inserted for future consumers — evicting them defeats
+    // their purpose. Regular ToMemoryConfigOp outputs from
+    // MemoryLayoutPropagation are NOT in this set and remain valid eviction
+    // candidates.
```
rpavlovicTT: Can we keep comments bit more concise. Claude generally likes to be super verbose. Here for example it is sufficient to say `Skip reshards we inserted for future consumers — evicting them defeats their purpose.`. Mentioning other ops does not bring much to reader at this point.

lib/Dialect/TTNN/Analysis/L1SpillManagement.cpp
```diff
@@ -686,7 +718,11 @@ void L1SpillManagement<MemoryTracker>::markEvictedAndRebuild(Value victim) {
   memoryTracker.restoreSnapshot(snapIt->second);
 
   // Replay events from the alloc point forward, skipping evicted tensors.
-  // Update snapshots during replay so future evictions see accurate state.
+  // Reshard kAlloc/kDealloc pairs that evictValue inserted into the log are
+  // replayed here like any other event — the reshard is allocated before the
+  // consumer's alloc (correct address for the consumer) and freed after. When
+  // the consumer's own kAlloc events are skipped (consumer later evicted), the
+  // reshard kDealloc still fires and reclaims the slot, so it is never stuck.
```
rpavlovicTT: This comment change is unecessary to me. Code below did not change at all. Yes, mechanism is affected by new events but this comment just pollutes this clean function. We have other places and comments that explain reshard insertions.

lib/Dialect/TTNN/Analysis/L1SpillManagement.cpp
```diff
@@ -784,16 +844,133 @@ void L1SpillManagement<MemoryTracker>::evictValue(Value victim, int64_t pos,
       if (needsReshard) {
```
rpavlovicTT:
```cpp
if (!needsReshard) {continue;}
<rest of code>
```

lib/Dialect/TTNN/Analysis/L1SpillManagement.cpp
```diff
@@ -734,19 +770,42 @@ void L1SpillManagement<MemoryTracker>::evictValue(Value victim, int64_t pos,
...
+  // Collect all insertions first; apply largest-position-first so that
+  // inserting later events does not shift the positions of earlier ones.
+  struct TransientInsertion {
+    size_t pos;
+    L1Event event;
+  };
+  llvm::SmallVector<TransientInsertion> transientInsertions;
+
   if (originalL1Layout.hasL1BufferType()) {
+    uint64_t reshardSizePerCore = utils::getPerCoreL1Usage(
```
rpavlovicTT: More precise is to use result of reshard validation to get aligned size per core.

lib/Dialect/TTNN/Analysis/L1SpillManagement.cpp
```diff
@@ -930,9 +1107,15 @@ uint64_t L1SpillManagement<MemoryTracker>::handleFragmentation(
...
     llvm::SmallVector<Value> toEvict;
     for (Value operand : op->getOperands()) {
-      if (liveValues.count(operand)) {
-        toEvict.push_back(operand);
+      if (!liveValues.count(operand)) {
+        continue;
+      }
+      // Never evict an inserted reshard — it was placed here specifically
+      // to feed this op. Regular to_memory_config ops are NOT protected.
+      if (insertedReshardValues.count(operand)) {
```
rpavlovicTT: I am not sure if this is correct. This case of concat op is an example where we need to spill all operands together so the op is valid. In this case inserted reshard should be overriden, otherwise op will never be valid.

lib/Dialect/TTNN/Analysis/L1SpillManagement.cpp
```diff
@@ -784,16 +844,133 @@ void L1SpillManagement<MemoryTracker>::evictValue(Value victim, int64_t pos,
...
+    Operation *reshardOp, Value reshardResult, uint64_t reshardSizePerCore,
+    int64_t consumerPos, ScheduleData &data) {
+  // Register so evictFarthestUse and sibling eviction guards know not to
+  // evict this value — it exists specifically to feed its consumer.
+  insertedReshardValues.insert(reshardResult);
+
+  // Annotate the reshard so run()'s main loop knows its L1 output size.
+  // MemoryLayoutPropagation didn't have inserted ops, so we must set this
+  // manually.
+  reshardOp->setAttr("ttnn.output_l1_usage",
```
rpavlovicTT: To be removed once you rebase after I check in https://github.com/tenstorrent/tt-mlir/pull/8526

lib/Dialect/TTNN/Analysis/L1SpillManagement.cpp
```diff
@@ -784,16 +844,133 @@ void L1SpillManagement<MemoryTracker>::evictValue(Value victim, int64_t pos,
...
+  }
+  data.positionMap[reshardOp] = consumerPos;
+
+  // Shift lastUsePositions for all values whose last use is now past the
+  // insertion point, so processDeadTensors frees them at the right time.
+  for (auto &[val, lastUse] : data.lastUsePositions) {
+    if (lastUse >= consumerPos) {
+      ++lastUse;
+    }
+  }
```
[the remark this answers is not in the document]
bmalesevicTT: Not common, but valid discrepancy. Addressed now by using 2 coordinate spaces, one for all ops in data.lastUsePositions/liveSet (same from the start, unaffected by reshards inserted into schedule, to keep relative ordering which is important for eviction strategy) and the second coordinate space for the inserted reshards (which are contained in data.lastUsePositions/liveSet but never chosen for eviction, hence their values there are irrelevant). Explained in code comments as well.

lib/Dialect/TTNN/Analysis/L1SpillManagement.cpp
```diff
@@ -629,7 +636,9 @@ void L1SpillManagement<MemoryTracker>::handleOOM(
       // tensors that would collide with the op's CB region.
       evictForDramCBGrowth(op, pos, data);
     }
-    if (l1Size > 0) {
+    // Skip allocation if eviction inserted a reshard for `op` and shifted it
+    // past `pos`; run()'s sweep reprocesses the reshard and then `op`.
+    if (l1Size > 0 && data.schedule[pos] == op) {
```
rpavlovicTT: Looks to me that if eviction inserted reshard for op and shifted schedule, we can just bail out immediately. So do something like. 
```cpp
if (shifted) {return}
if (l1Size > 0) { ... }
```

lib/Dialect/TTNN/Analysis/L1SpillManagement.cpp
```diff
@@ -1108,6 +1301,13 @@ void L1SpillManagement<MemoryTracker>::run() {
       for (auto r : tensorResults) {
         Value val = r;
         auto luIt = data.lastUsePositions.find(val);
+        // Every scheduled op (original or inserted reshard) gets an entry at
+        // build time or in insertReshardIntoSchedule. A miss would give `val` a
+        // shifted-space priority via the fallback below, corrupting
+        // farthest-last-use ordering — assert loudly; the fallback only keeps
+        // release builds defined.
```
rpavlovicTT: Unnecessary comment, assert is fine

include/ttmlir/Dialect/TTNN/Analysis/L1SpillManagement.h
```diff
@@ -285,53 +311,55 @@ class L1SpillManagement {
...
-  /// sibling-operand eviction paths.
-  void evictValue(Value victim, int64_t pos, const ScheduleData &data);
+  /// Evict a specific live value: spill to DRAM, update tracker, and insert
+  /// reshards for consumers that still require the L1 layout. Sibling-operand
+  /// eviction passes skipReshardConsumer=op: it spills all operands of that op
+  /// to give it homogeneous DRAM inputs, so resharding one back to L1 would
+  /// defeat the purpose. The op's other (past/future) consumers are still
+  /// restored.
+  void evictValue(Value victim, int64_t pos, ScheduleData &data,
+                  Operation *skipReshardConsumer = nullptr);
```
rpavlovicTT: This comment should explain how param `skipReshardConsumer` affects `evictValue`, not necessarily which path passes it and why.

lib/Dialect/TTNN/Analysis/L1SpillManagement.cpp
```diff
@@ -1161,6 +1361,13 @@ void L1SpillManagement<MemoryTracker>::run() {
                    ttmlir::opToString(op), result.cbPeakUsage, l1Size);
 
       l1Size = ensureFitsL1(op, pos, data, result.cbPeakUsage, l1Size);
+      // Eviction may have inserted a reshard for `op` itself, shifting `op`
+      // past `pos`. Rewind to the first inserted reshard (always at `pos`); the
+      // loop walks forward through any others, then reprocesses `op`.
```
rpavlovicTT: I think it could happen that we insert more than 1 reshard for this op. For example in the case of binary or concat op where each operand gets resharded from DRAM due to spill. In this case current op is pushed for more than 1 place in the schedule, right?

lib/Dialect/TTNN/Analysis/L1SpillManagement.cpp
```diff
@@ -1161,6 +1361,13 @@ void L1SpillManagement<MemoryTracker>::run() {
...
+      if (data.schedule[pos] != op) {
+        --pos;
+        continue;
+      }
```
rpavlovicTT: You should wrap this in a lambda that checks whether rewind is necessary, moves `pos` to correct position and returns true so that you know you should `continue`. Reuse below.

lib/Dialect/TTNN/Analysis/L1SpillManagement.cpp
bmalesevicTT: We can insert more reshards but that doesn't affect our rewind count. If the op was at pos=5, and we insert 3 reshards for its inputs, those will land at schedule indexes 5,6,7 and op will be index 8. If there were no reshards inserted, next op to process would be pos=6, but since we want the rewind we go --pos, so we get to process pos=5 again("the first inserted reshard (always at `pos`)"). 
This way we will eventually process all reshards and the op itself, the loop `pos < static_cast<int64_t>(data.schedule.size())` will adjust to the now larger schedule.

rpavlovicTT: Awesome, thanks for explaining!

rpavlovicTT (approved):

PR #8571, opened 2026-05-22
[D2M] GridSelection abstraction cleanup

vwellsTT:
### Ticket
NA

### Problem description
In D2M, the way we assign grids is still a bit dirty; we split into a separate analysis + transform phase, but for certain ops like composite views, we still recompute shapes during transformation; furthermore, for regular cases, we often reconcile conflicting shapes across operands at transform-time too.

### What's changed
Enforce a stricter logical structure where we compute optimal grid per operand, including any reconciliation, during analysis, and transform is that much more mechanical.  This updates some padding logic a bit too, etc.

### Checklist


lib/Dialect/D2M/Analysis/GridAnalysis.cpp
```diff
@@ -42,6 +44,56 @@ static int64_t findLargestCommonFactor(int64_t maxFactor,
   return 1;
 }
 
+static int64_t getVolume(ArrayRef<int64_t> shape) {
```
wenbinlyuTT: Nit: We already have `ttmlir::utils::volume()`.

lib/Dialect/D2M/Transforms/GridSelection.cpp
```diff
@@ -418,108 +433,88 @@ static void applyTTNNTensorUpdate(const OperandGridInfo &info,
...
+    const EffectiveTargetGridRange &effectiveTargetGridRange,
+    OpBuilder &builder) {
+  auto inputType = mlir::cast<RankedTensorType>(input.getType());
+  auto inputLayout =
+      mlir::dyn_cast<ttcore::MetalLayoutAttr>(inputType.getEncoding());
+  if (!inputLayout) {
+    return input;
+  }
+
+  if (Value rewrittenProducer = rewriteSingleUseCompositeInputProducer(
```
wenbinlyuTT, line 484: Nit: it would be helpful to add the comment about why we care about single use inputs back here.

wenbinlyuTT (approved):
Composite view changes LGTM, only two nits.

PR #8572, opened 2026-05-22
[d2m-jit] Add Python pattern-rewrite framework for TTIR -> d2m lowering

nsmithtt:
Lets users define MLIR pattern rewrites in plain Python that match TTIR (or any) ops and replace them with d2m-jit kernel-template subgraphs. Adding a new pattern is a matter of dropping a new .py file next to the driver — no C++ rebuild required.

Surface (all re-exported via d2m_jit.api):

    @d2m.pattern(root=ttir.MatmulOp, benefit=5)
    def lower_matmul(op, rewriter):
        a = d2m.from_value(op.operands[0])
        b = d2m.from_value(op.operands[1])
        out = d2m.empty(d2m.infer_layout(op.result, grid_shape=[4, 4]))
        my_fused_kernel(a, b, out, grid=(4, 4))
        return d2m.from_device(out)

    # DAG match: predicate gates the rewrite.
    @d2m.pattern(root=ttir.ExpOp, match=lambda op: ...)
    def fuse_add_exp(op, rewriter): ...

    d2m.apply_patterns(module)        # Python driver, in-process
    # OR
    ttmlir-opt --d2m-python-rewrites=module-path=foo.py input.mlir

Implementation summary:

* PDL bootstrap (lib/RegisterAll.cpp). Registers pdl::PDLDialect and pdl_interp::PDLInterpDialect so textual pdl.pattern modules can be parsed via mlir.ir.Module.parse and consumed by mlir.rewrite.PDLModule from Python.

* Scope abstraction (tools/d2m-jit/_src/builder.py). Lazy-emission helpers (to_layout, empty, view_layout, _emit_kernel_generic, LazyTensor._resolve) now target a swappable build context via _get_scope() + a thread-local push/pop. The existing _Builder singleton is one concrete scope; the new RewriteScope plugs in a PatternRewriter's ctx/loc/IP so kernel emission lands at the matched op's site. RewriteScope.add_scalar_input emits an arith.constant of index type at the rewriter's IP (lazy mode adds a func arg; rewrite mode can't and uses a host-scope constant). to_host() refuses to run from a non-lazy scope.

* Framework (tools/d2m-jit/_src/rewrite.py). Each @d2m.pattern lowers to one pdl.pattern stub:

```
pdl.pattern @<name> : benefit(N) {
  %operands = pdl.operands
  %types = pdl.types
  %op = pdl.operation \"<root>\"(%operands : !pdl.range<value>)
                                 -> (%types : !pdl.range<type>)
  (pdl.apply_native_constraint \"<match>\"(%op : !pdl.operation))?
  pdl.rewrite %op {
    %v = pdl.apply_native_rewrite \"<dispatch>\"(%op : !pdl.operation) : !pdl.value
    pdl.replace %op with (%v : !pdl.value)
  }
}
```

  Variadic operands/types so one template handles ops of any arity. The dispatcher auto-pushes a RewriteScope before invoking the user function. Predicate-gated matching uses PDL constraints (which can fail cleanly) rather than rewrite-side bail-out (which would crash the greedy driver — see findings below).

  Helpers:
   - from_value(ir.Value, layout=None): wrap an SSA value as LazyTensor; layout defaults to infer_layout(value).
   - infer_layout(value_or_type, **overrides): safe defaults (logical / untiled / single-cell, mem_space=l1). Patterns override liberally.
   - from_device(lt): emit the d2m.to_layout that materialises a device LazyTensor back to a plain host tensor type — the typical last line of a rewrite body so the returned Value matches the matched op's result type.

* Two example pattern modules (tools/d2m-jit/patterns/).
   - eltwise_exp_to_kernel.py: single-op match (ttir.exp) -> fused d2m kernel that calls d2m.tile_exp via remote_load/remote_store.
   - eltwise_add_exp_to_kernel.py: DAG match (ttir.add -> ttir.exp) -> one fused kernel computing (a + b).exp() at the tile level. Benefit 20 outranks the single-eltwise pattern when both apply.

* Standalone C++ pass (lib/Dialect/D2M/Transforms/PythonRewritesPass.cpp, Passes.td entry). ttmlir-opt --d2m-python-rewrites=module-path=... embeds CPython, loads user pattern .py files at run time, and rewrites the host ModuleOp.
   - Uses the raw CPython C-API (not nanobind/pybind11) because MLIRD2MTransforms inherits LLVM's -fno-rtti / -fno-exceptions.
   - Does a TEXT round-trip across the boundary rather than sharing MlirModule pointers: ttmlir-opt links its own static MLIR while ttmlir's Python bindings ship libTTMLIRPythonCAPI.so with a separate MLIR runtime, so capsule pointers can't safely cross (verified empirically). The pass prints the module to a string, calls d2m_jit._src.rewrite.apply_patterns_text, re-parses the result back in the host context, and splices over the original module body. Slow and lossy on locations; a real fix is shared MLIR (LLVM_BUILD_LLVM_DYLIB), tracked separately.
   - Gated on TTMLIR_ENABLE_D2M_JIT: when off, the .cpp compiles as a fail-fast stub so ttmlir-opt --help still advertises the flag with a clean diagnostic.

* Tests.
   - test/d2m-jit/lit/eltwise_exp_pattern.py: parses a ttir.exp module, runs apply_patterns in-process, FileCheck'd for d2m.generic + d2m.tile_exp + the host-typed return.
   - test/d2m-jit/lit/eltwise_add_exp_pattern.py: same for the DAG case. Positive case CHECKs d2m.tile_add and d2m.tile_exp share a single d2m.generic. Negative case (exp with no upstream add) CHECKs the ttir.exp survives untouched.
   - test/d2m-jit/lit/python_rewrites_pass.py: drives the C++ pass via subprocess (ttmlir-opt --d2m-python-rewrites=...) on the same TTIR input and CHECKs the same output IR.
   - test/d2m-jit/test_pattern_eltwise.py: on-device pytest, exercises the bundled kernels directly through the lazy d2m-jit path with PCC vs torch. Combined with the lit tests this gives end-to-end coverage (lit proves the rewrite produces IR X; pytest proves IR X runs correctly on silicon).
   - %ttmlir_tools lit substitution added so the subprocess test can locate ttmlir-opt portably.

Three non-obvious findings worth flagging:

* PDL native rewrites only dispatch when the pdl.rewrite body uses pdl.apply_native_rewrite. The pdl.rewrite %op with \"name\"(args) external-form syntax parses and verifies but does not call the function registered via register_rewrite_function.

* Operands must be bound (pdl.operand or pdl.operands) for the bytecode interpreter to dispatch the rewrite. Without an operand binding the rewrite block is silently skipped — no error, no fire.

* The greedy pattern rewriter does not support recoverable failures inside pdl.apply_native_rewrite. Returning failure from the rewrite crashes with \"pattern rewriter doesn't support recovery\". That's why bail-out lives in match= (PDL constraint, which IS allowed to fail) rather than as a return None from the rewrite body.

### DIFF
--- a/lib/Dialect/D2M/Transforms/PythonRewritesPass.cpp
+++ b/lib/Dialect/D2M/Transforms/PythonRewritesPass.cpp
@@ -1,0 +1,279 @@
+// SPDX-FileCopyrightText: (c) 2026 Tenstorrent AI ULC
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#include "ttmlir/Dialect/D2M/Transforms/Passes.h"
+
+#include "mlir/IR/AsmState.h"
+#include "mlir/IR/Block.h"
+#include "mlir/IR/BuiltinOps.h"
+#include "mlir/Parser/Parser.h"
+#include "mlir/Pass/Pass.h"
+
+#include "llvm/Support/raw_ostream.h"
+
+namespace mlir::tt::d2m {
+#define GEN_PASS_DEF_D2MPYTHONREWRITES
+#include "ttmlir/Dialect/D2M/Transforms/Passes.h.inc"
+} // namespace mlir::tt::d2m
+
+#ifdef TTMLIR_ENABLE_D2M_JIT
+
+// We use the raw CPython C-API rather than nanobind/pybind11 here because
+// MLIRD2MTransforms inherits LLVM's -fno-rtti / -fno-exceptions, which
+// nanobind doesn't support.
+//
+// The pass does NOT cross C++/Python via a shared MlirModule pointer —
+// ttmlir-opt is statically linked against MLIR while the ttmlir Python
+// bindings ship their own copy of MLIR (libTTMLIRPythonCAPI.so). The
+// two MLIR runtimes can't share opaque storage (StringAttr uniquer
+// etc.), so we fall back to a text round-trip:
+//
+//   1. C++: print the host ModuleOp to a string.
+//   2. Python: parse the string in *its* ir.Context, run apply_patterns,
+//      print the rewritten module back to a string.
+//   3. C++: parse the result string into the host context as a new
+//      module, then swap its body into the original ModuleOp.
+
+#include <Python.h>
+
+#include <atomic>
+#include <string>
+
+namespace mlir::tt::d2m {
+namespace {
+
+// RAII wrapper for Py_DECREF on a PyObject*.
+class PyRef {
+public:
+  PyRef() = default;
+  explicit PyRef(PyObject *obj) : obj(obj) {}
+  PyRef(const PyRef &) = delete;
+  PyRef &operator=(const PyRef &) = delete;
+  PyRef(PyRef &&other) noexcept : obj(other.obj) { other.obj = nullptr; }
+  PyRef &operator=(PyRef &&other) noexcept {
+    if (this != &other) {
+      reset();
+      obj = other.obj;
+      other.obj = nullptr;
+    }
+    return *this;
+  }
+  ~PyRef() { reset(); }
+
+  void reset(PyObject *newObj = nullptr) {
+    Py_XDECREF(obj);
+    obj = newObj;
+  }
+  PyObject *get() const { return obj; }
+  explicit operator bool() const { return obj != nullptr; }
+
+private:
+  PyObject *obj = nullptr;
+};
+
+// RAII GIL acquire/release. Cheaper than thread-state save/restore.
+class PyGIL {
+public:
+  PyGIL() : state(PyGILState_Ensure()) {}
+  PyGIL(const PyGIL &) = delete;
+  PyGIL &operator=(const PyGIL &) = delete;
+  ~PyGIL() { PyGILState_Release(state); }
+
+private:
+  PyGILState_STATE state;
+};
+
+// Ensure Python is initialized exactly once per process.
+static void ensurePythonInitialized() {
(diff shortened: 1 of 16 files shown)


lib/Dialect/D2M/Transforms/PythonRewritesPass.cpp
```diff
@@ -0,0 +1,280 @@
...
+// ttmlir-opt is statically linked against MLIR while the ttmlir Python
+// bindings ship their own copy of MLIR (libTTMLIRPythonCAPI.so). The
+// two MLIR runtimes can't share opaque storage (StringAttr uniquer
+// etc.), so we fall back to a text round-trip:
+//
+//   1. C++: print the host ModuleOp to a string.
+//   2. Python: parse the string in *its* ir.Context, run apply_patterns,
+//      print the rewritten module back to a string.
+//   3. C++: parse the result string into the host context as a new
+//      module, then swap its body into the original ModuleOp.
```
vwellsTT, line 36: This seems unfortunate, but I'm not sure I have a better suggestion or anything

vwellsTT (approved):

lib/Dialect/D2M/Transforms/PythonRewritesPass.cpp
nsmithtt: My plan is to uplift LLVM which I think should have a better mechanism for this.  If not we should explore again ways to build the project to make this easier 😅

PR #8574, opened 2026-05-22
TTIR to TTNN: gate SDPA decode dispatch on Sk % 32 == 0

sdjukicTT:
### Problem description
`ScaledDotProductAttentionOpConversionPattern::shouldUseDecode` in
`lib/Conversion/TTIRToTTNN/TTIRToTTNN.cpp` lowered any `Sq == 1` TTIR SDPA
to `ttnn.scaled_dot_product_attention_decode`. The decode kernel's
internal `k_chunk_size = min(512, largest_pow2_divisor(Sk))` must be a
multiple of 32 (equivalent to `Sk % 32 == 0`), or tt-metal `TT_FATAL`s at
runtime in
`ttnn/.../sdpa_decode/sdpa_decode.cpp:64-67`
(`"Chunk size must be multiple of 32, ..."`). This is an intentional
kernel design — decode is for tile-aligned KV caches — so the fix lives
in the compiler.

### What's changed
`shouldUseDecode` also requires `Sk % 32 == 0`. Anything else falls
through to `lowerToSDPAOp` and is handled by the generic prefill
`ttnn.scaled_dot_product_attention`, which supports arbitrary `Sk`.

### Checklist
`test/ttmlir/Dialect/TTNN/transformer/scaled_dot_product_attention.mlir`
  (`Sq=1`, `Sk=16` → prefill, not decode).

### DIFF
--- a/lib/Conversion/TTIRToTTNN/TTIRToTTNN.cpp
+++ b/lib/Conversion/TTIRToTTNN/TTIRToTTNN.cpp
@@ -3362,11 +3362,15 @@ private:
   // [B, 1, Hq, Sk].
   static constexpr int64_t kDecodeMaskNumHeadsDim = 2;
 
-  // Determine if the decode op should be used based on query sequence length.
-  // SDPA decode is optimized for autoregressive decoding where seq_len == 1.
+  // Decode kernel requires Sq == 1 and Sk % 32 == 0 (k_chunk_size constraint);
+  // otherwise fall back to the generic prefill op.
   bool shouldUseDecode(ttir::ScaledDotProductAttentionOp op) const {
     auto queryType = mlir::cast<RankedTensorType>(op.getQuery().getType());
-    return queryType.getDimSize(kSeqLenDim) == 1;
+    if (queryType.getDimSize(kSeqLenDim) != 1) {
+      return false;
+    }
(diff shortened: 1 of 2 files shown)


jserbedzijaTT:
We should create a workaround that lowers sdpa decode into the regular sdpa op instead of hardcoding the logic here. Can you also open a tt-metal issue for this?

sdjordjevicTT:
> We should create a workaround that lowers sdpa decode into the regular sdpa op instead of hardcoding the logic here. Can you also open a tt-metal issue for this?

+1

sdjukicTT:
After consideration, I propose keeping this as is.

`shouldUseDecode` picks which TTNN op to emit (decode vs generic), and the two differ in operand layout - decode permutes/broadcasts Q, mask, and result. That's a lowering decision, so it belongs in conversion.

The SDPA workaround patterns all keep the op type fixed and just patch operands. Moving this there would mean either always emitting decode and reconstructing the generic op by inverting the permutes (fragile, and lossy - decode drops `sliding_window_size`), or always emitting generic and doing the decode layout transform inside workarounds (which is the actual lowering, not a fix-up). The `Sk % 32` clause is workaround-flavored knowledge, but it gates an op swap, not an operand patch. Also it is an intentional kernel design (decode is for tile-aligned KV caches), not a bug we are expecting to get fixed.

